Issue #1182, opened 2021-06-28
Clip names


AnitaBats:
### User Story
As a user of Audacity, I would like to have individual clip names, and be able to rename clips.

### Description
Currently, there is no concept of clip names in Audacity, only track names. Individual clip names are very useful when organising a project, or separating between copies or takes. The new affordance design has a space for clip names top right.

The design at https://github.com/audacity/audacity/issues/1258 describes how to rename clips as a user.

**Generated/default  clip names**

When recording to a track, the auto-generated clip name should be derived from the track name by default: trackname#XX, where #XX is the number of the recording/take. This can be changed to anything afterwards

When importing an audio file to a track, the clip name should be the file name. This can be changed to anything afterwards

### Rules for auto-naming split/joined clips and copies

Suggested rules for auto-clips naming:
- New clip recording: "{track_name} #{take_number} {channel}"
- Merge clips: "{first_clip_name}"
- Split: "{clip_name} {index}"
- Import: "{track_name}"
- Copy: "{original_clip_name} copy {number}"
- Copy/Paste selection(part of a clip): "{original_clip_name} samples [range]" 

 suggested scheme may produce results like: "Clip 1 1" "Clip 1 2" "Clip 2" when splitting "Clip" into "Clip 1" and "Clip 2" and then splitting "Clip 1" into "Clip 1 1" and "Clip 1 2", similar would happen with copy-pasting a clip or it's parts. Is it ok, or is there some better scheme for such cases? 
 
 ### Additions
 
- basic RTL appearance

vsverchinsky:
Est.: 16hrs

vsverchinsky:
Suggested rules for auto-clips naming:
- New clip recording: "{track_name} #{take_number} {channel}"
- Merge clips: "{first_clip_name}"
- Split: "{clip_name} {index}"
- Import: "{track_name}"
- Copy: "{original_clip_name} copy {number}"
- Copy/Paste selection(part of a clip): "{original_clip_name} samples [range]" 
@jounih suggested scheme may produce results like: "Clip 1 1" "Clip 1 2" "Clip 2" when splitting "Clip" into "Clip 1" and "Clip 2" and then splitting "Clip 1" into "Clip 1 1" and "Clip 1 2", similar would happen with copy-pasting a clip or it's parts. Is it ok, or there some better scheme for such cases?

petersampsonaudacity:
See also Discussion #1499 
>Thought-piece: labels for merged clips #1499

Issue #1324, opened 2021-07-19
Clip names: right to left languages


jounih:
Clip names should also support RTL settings - the position of the title should be mirrored

jounih:
![Clip names@2x](https://user-images.githubusercontent.com/10109867/126167060-98b35627-b47e-4d28-98e8-10b67f31fc33.jpg)

AnitaBats:
Design is ready, can be moved to implementation.

AnitaBats:
Done in scope of #1533 task. Needed QA when #1533 is done

PR #1533, opened 2021-08-25
Clip names

vsverchinsky:
Resolves: #1182 #1324

### DIFF
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -647,7 +647,16 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
             // Less than or equal, not just less than, to ensure a clip boundary.
             // when append recording.
             if (endTime <= t0) {
-               pending->CreateClip(t0);
+               auto newName = [&]() {
+                  for (auto i = 1;; ++i)
+                  {
+                     //i18n-hint a numerical suffix added to distinguish otherwise like-named clips when new record started
+                     auto name = XC("%s #%d", "clip name template").Format(pending->GetName(), i).Translation();
+                     if (pending->FindClipByName(name) == nullptr)
+                        return name;
+                  }
+               }();
+               pending->CreateClip(t0, newName);
             }
             transportTracks.captureTracks.push_back(pending);
          }
--- a/src/TrackArtist.cpp
+++ b/src/TrackArtist.cpp
@@ -54,6 +54,7 @@ audio tracks.
 #include "Decibels.h"
 #include "prefs/TracksPrefs.h"
 
+#include <wx/app.h>
 #include <wx/dc.h>
 
 //Thickness of the clip frame outline, shown when clip is dragged
@@ -239,6 +240,41 @@ void TrackArt::DrawNegativeOffsetTrackArrows(
                 rect.x + 6, rect.y + rect.height - 12);
 }
 
+wxString TrackArt::TruncateText(wxDC& dc, const wxString& text, const int maxWidth)
+{
+   static const wxString ellipsis = "\u2026";
+
+   if (dc.GetTextExtent(text).GetWidth() <= maxWidth)
+       return text;
+
+   auto left = 0;
+   //no need to check text + '...'
+   auto right = static_cast<int>(text.Length() - 2);
+
+   while (left <= right)
+   {
+      auto middle = (left + right) / 2;
+      auto str = text.SubString(0, middle).Trim() + ellipsis;
+      auto strWidth = dc.GetTextExtent(str).GetWidth();
+      if (strWidth < maxWidth)
+         //if left == right (== middle), then exit loop 
+         //with right equals to the last knwon index for which 
+         //strWidth < maxWidth
+         left = middle + 1;
+      else if (strWidth > maxWidth)
+         //if right == left (== middle), then exit loop with
+         //right equals to (left - 1), which is the last known
+         //index for which (strWidth < maxWidth) or -1
+         right = middle - 1;
+      else
+         return str;
+   }
+   if (right >= 0)
+      return text.SubString(0, right).Trim() + ellipsis;
+
+   return wxEmptyString;
+}
+
 
 #ifdef USE_MIDI
 #endif // USE_MIDI
@@ -263,7 +299,7 @@ void TrackArtist::UpdatePrefs()
    SetColours(0);
 }
 
-void TrackArt::DrawClipAffordance(wxDC& dc, const wxRect& rect, bool highlight, bool selected)
+void TrackArt::DrawClipAffordance(wxDC& dc, const wxRect& rect, const wxString& title, bool highlight, bool selected)
 {
    if (selected)
    {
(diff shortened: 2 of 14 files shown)


Paul-Licameli (commented):
First commit is good

src/TrackArtist.cpp
```diff
@@ -239,6 +240,35 @@ void TrackArt::DrawNegativeOffsetTrackArrows(
...
+      if (strWidth < maxWidth)
+         left = middle + 1;
+      else if (strWidth > maxWidth)
+         right = middle - 1;
+      else
+         return str;
+   }
+   auto index = std::min(left, right);
+   if (index >= 0)
+      return text.SubString(0, index).Trim() + ellipsis;
```
Paul-Licameli, line 267: This might be simplified and clarified by using this standard function https://en.cppreference.com/w/cpp/algorithm/partition_point

src/WaveTrack.cpp
```diff
@@ -90,6 +90,9 @@ Track::LinkType ToLinkType(int value)
    return static_cast<Track::LinkType>(value);
 }
 
+constexpr const char* SplitLeftClipPostfix{ " 1" };
+constexpr const char* SplitRightClipPostfix{ " 2" };
```
Paul-Licameli, line 94: Again I would make a format that substitutes the clip name, if I really do this.

Is this split-clip naming convention good?  I wonder.  Repeated splits might pile up " 1 1 1" ... repeatedly and be undesirable.  But I'm not ruling on that design question.

Paul-Licameli (changes requested):

Paul-Licameli:
All is well in the last commit

vsverchinsky:
Just rebase, nothing fixed yet

src/WaveTrack.h
```diff
@@ -572,6 +574,9 @@ class AUDACITY_DLL_API WaveTrack final : public PlayableTrack {
    ConstIntervals GetIntervals() const override;
    Intervals GetIntervals() override;
 
+   //! Returns nullptr if clip with such name was not found
+   const WaveClip* FindClipByName(const wxString& name) const;
+
```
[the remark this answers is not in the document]
vsverchinsky: It will work too

src/ProjectAudioManager.cpp
```diff
@@ -644,7 +644,15 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
             // Less than or equal, not just less than, to ensure a clip boundary.
             // when append recording.
             if (endTime <= t0) {
-               pending->CreateClip(t0);
+               auto newName = [&]() {
+                  for (auto i = 1;; ++i)
+                  {
+                     auto name = pending->GetName() + wxString::Format(" #%i", i);
```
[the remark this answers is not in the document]
vsverchinsky: Will fix that, thanks!

src/WaveTrack.cpp
vsverchinsky: I raised same question in https://github.com/audacity/audacity/issues/1182 there were no objections then

vsverchinsky: When making copies of a copy it's not unusual to get names like "filename - copy - copy - ...", but maybe there is a better approach for clip naming, I don't know.

src/WaveTrack.h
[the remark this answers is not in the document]
vsverchinsky: >Are non-unique clip names possible
Yes, but this method is used to check clip existence

src/TrackArtist.cpp
vsverchinsky: But i don't see how to use predicate in that case, if I understand you correctly you suggest this:
`std::partition_point(text.begin(), text.end(), [](auto char) { return /*how to get length?*/ }`

Paul-Licameli: I think you could take the char by reference and use its address.  A bit of a cheat though.

Paul-Licameli: If you don’t do that, I would prefer you clarify the loop by stating invariant conditions.  I think the invariant is: left - 1 is an allowed length for trimming the string or -1, and right +1 is a disallowed length.

If the loop terminates, the loop condition is false, so the min is right.  More, you can prove right == left - 1, and therefore a good length or -1 because the invariant is preserved.

The lack of comments a out the invariant and the unnecessary min makes me question how carefully you reasoned.

src/WaveTrack.cpp
```diff
@@ -1116,17 +1155,18 @@ void WaveTrack::HandleClear(double t0, double t1,
                   // Delete in the middle of the clip...we actually create two
                   // NEW clips out of the left and right halves...
 
-                  // left
-                  clipsToAdd.push_back
-                     ( std::make_unique<WaveClip>( *clip, mpFactory, true ) );
-                  clipsToAdd.back()->Clear(t0, clip->GetEndTime());
+                  auto leftClip = std::make_unique<WaveClip>(*clip, mpFactory, true);
+                  //i18n-hint Name assigned to the first clip on split
```
Paul-Licameli, line 1159: Maybe say “when splitting a wave track”

src/WaveTrack.cpp
```diff
@@ -1116,17 +1155,18 @@ void WaveTrack::HandleClear(double t0, double t1,
...
+                  clipsToAdd.push_back(std::move(leftClip));
 
-                  // right
-                  clipsToAdd.push_back
-                     ( std::make_unique<WaveClip>( *clip, mpFactory, true ) );
-                  WaveClip *const right = clipsToAdd.back().get();
-                  right->Clear(clip->GetStartTime(), t1);
-                  right->Offset(t1 - clip->GetStartTime());
+                  auto rightClip = std::make_unique<WaveClip>(*clip, mpFactory, true);
+                  //i18n-hint Name assigned to the second clip on split
```
Paul-Licameli, line 1165: Ditto

src/WaveTrack.cpp
```diff
@@ -1116,17 +1155,18 @@ void WaveTrack::HandleClear(double t0, double t1,
...
+                  rightClip->SetName(XC("%s 2", "clip name template").Format(leftClip->GetName()).Translation());
```
Paul-Licameli, line 1166: Does this test correctly?  I think this would crash because you use leftClip after moving from it.

Paul-Licameli (changes requested):
Null pointer dereference?

Paul-Licameli:
The i18n changes are good, I suggest better comments about the logic of the binary search, but there is a possible crash bug in split-cut of tracks which was introduced in a later force-push.  I think you write rightClip where you meant leftClip.

Fix that, and the comments, test split-cut (Edit > Remove Special > Split Cut), rebase and force-push, and I will approve.

src/WaveTrack.cpp
vsverchinsky: Typo, thanks!

src/TrackArtist.cpp
vsverchinsky: >I think you could take the char by reference and use its address. A bit of a cheat though.

I would avoid making suggestions about internal representation of strings in the wxString, though it's been said that it uses std::basic_string as a container, encoding is another problem

>so the min is right

Right, thanks!

Paul-Licameli (approved):

Issue #1288, opened 2021-07-13
Macros run on project are inconsistent in their use of Export commands


petersampsonaudacity:
**Describe the bug**
Macros run on project are inconsistent in their use of Export commands.
a) if the project is named then the macros-output directory is used (location as specified in Directories preferences)
b) if the project is un-named the the Export Audio dialog is shown to the user.

The bug here is the inconsistency in behavior (hence P2)

Plus the opacity involved in named versus un-named - a project acquires a name if:
a) the user saves the project
b) the user imports an audio file into the project as the first audio they acquire
This will be a subtlety lost on most users.

**To Reproduce**
To be provided

**Expected behavior**
When running a Macro on a project, at the Export Audio step in the Macro the Export Audio dialog should always be shown to the user,  regardless of whether the project is named or un-named

**Screenshots**
Not applicable.

**Additional information (please complete the following information):**
 - All OS
 - Version 3.0.2 (and alphas for 3.0.3 and 3.0.4)

**Additional context**
Apart from the consistency aspect - the presentation of the export dialog to the user is a good indication that the Macro has run properly.

See also GitHub Issue #1285

petersampsonaudacity:
I have updated the Manual with the _(erroneous)_ behavior as-is in 3.0.3 as I suspect this will not get fixed for 3.0.3
https://alphamanual.audacityteam.org/man/Apply_Macro#Project
and
https://alphamanual.audacityteam.org/man/Directories_Preferences

Paul-Licameli:
@MalevolentStrix interested?

MalevolentStrix:
Sure, will start working on it in a day

Paul-Licameli:
@petersampsonaudacity @MalevolentStrix Isn't there an existing "inconsistency" if you apply export commands directly with the menus?  There is always prompting when the project is not saved.  There is not prompting otherwise.  But that is like the "inconsistency" of File > Save.

That "inconsistency" between saved and unsaved projects, doing a command directly, will remain after this fix.  The fix will only change the applying of a macro directly that has an export command.  Then that applying-directly will not behave like direct use of a menu command (it will prompt always), which you might also call "inconsistency."

I don't know how to stipulate the best program behavior, really.  But, I will review the code that I see here supposing the intent is correct.

@MalevolentStrix the first thing you should do is change the pull request comment so the first line i

Resolves: #1288

Paul-Licameli:
[...]

PR #1541, opened 2021-08-26
Macros run on project are inconsistent in their use of Export commands #1288

CelestialZeus:
Resolves: https://github.com/audacity/audacity/issues/1288

Add a new variable mApplyToProject to keep track of whether the Macro being applied is on a project or not and adding the same as a condition to the prompt in FileMenus.cpp.

ezgif com-gif-maker

- [X] I made sure the code compiles on my machine
- [X] I made sure there are no unnecessary changes in the code
- [X] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [X] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"

### DIFF
--- a/src/BatchProcessDialog.cpp
+++ b/src/BatchProcessDialog.cpp
@@ -1303,9 +1303,12 @@ void MacrosWindow::OnDown(wxCommandEvent & WXUNUSED(event))
 
 void MacrosWindow::OnApplyToProject(wxCommandEvent & event)
 {
+   AudacityProject *project = &mProject;
+   project->mApplyToProject = true;
    if( !SaveChanges() )
       return;
    ApplyMacroDialog::OnApplyToProject( event );
+   project->mApplyToProject = false;
 }
 
 void MacrosWindow::OnApplyToFiles(wxCommandEvent & event)
--- a/src/Project.h
+++ b/src/Project.h
@@ -157,6 +157,7 @@ private:
  public:
    bool mbBusyImporting{ false }; // used to fix bug 584
    int mBatchMode{ 0 };// 0 means not, >0 means in batch mode.
+   bool mApplyToProject{ false };// turns true when enters body of OnApplyToProject and to false when it reaches the end.
 
  private:
    wxWeakRef< wxFrame > mFrame{};
--- a/src/menus/FileMenus.cpp
+++ b/src/menus/FileMenus.cpp
@@ -52,6 +52,7 @@ void DoExport(AudacityProject &project, const FileExtension &format)
 
    // Prompt for file name and/or extension?
    bool bPromptingRequired = !project.mBatchMode ||
+                             project.mApplyToProject||
                              projectName.empty() ||
                              format.empty();
 


src/BatchProcessDialog.cpp
```diff
@@ -1303,9 +1303,12 @@ void MacrosWindow::OnDown(wxCommandEvent & WXUNUSED(event))
 
 void MacrosWindow::OnApplyToProject(wxCommandEvent & event)
 {
+   AudacityProject *project = &mProject;
+   project->mApplyToProject = true;
```
Paul-Licameli, line 1307: You did NOT follow the example of MacroCommands::ApplyCommandInBatchMode in the use of finally() to reset the flag back to false.

There is a very important difference between that, and what you wrote here.  With finally() the restoring of the state is guaranteed even if the applying of the command throws and exception.  Not so as written here.

src/Project.h
```diff
@@ -157,6 +157,7 @@ class AUDACITY_DLL_API AudacityProject final
  public:
    bool mbBusyImporting{ false }; // used to fix bug 584
    int mBatchMode{ 0 };// 0 means not, >0 means in batch mode.
+   bool mApplyToProject{ false };// turns true when enters body of OnApplyToProject and to false when it reaches the end.
```
Paul-Licameli, line 160: I don't like introducing this public member to the top-level Project object, just to solve this very special problem.  (The other two members here are not good examples to follow.  They are a few legacies in need of cleanup.)

Better that the new state be in the ProjectFileIO object.  It should not be a public member variable, but a private variable whose state can be got with an accessor.

src/BatchProcessDialog.cpp
Paul-Licameli: If you instead use ProjectFileIO::Get(mProject) to get the object, then you might need to set the flag in it and set it again, with a setting method paired with the new accessor method.

But don't to that.  Do a more complicated but better thing.  See class ValueRestorer.  Give ProjectFileIO a method like

```
ValueRestorer<bool> ProjectFileIO::ApplyingToProjectScope()
```

and at point of use

```
auto scope = ProjectFileIO::ApplyingToProjectScope();
```

And then you don't need a finally(), but scope serves the purpose of guaranteeing cleanup when it is destroyed.

The reason for this trouble?  Because making a new variable with a getter and a setter is little better than just making a public variable.  But instead you can constrain the changes of state, so that users of ProjectFileIO cannot to otherwise but change the variable in this safer way with guaranteed cleanup.

Paul-Licameli: If you don't know the C++ programming pattern called "RAII," please look it up and learn it!  What I am recommending here falls under the wider patter.   It is applied, not to recover memory or another resource, but rather to guarantee that temporary changes that a function makes to its environment (the state of ProjectFileIO here) are guaranteed to be undone no matter how the function may exit.

Paul-Licameli (approved):

src/BatchProcessDialog.cpp
```diff
@@ -1303,9 +1303,12 @@ void MacrosWindow::OnDown(wxCommandEvent & WXUNUSED(event))
 
 void MacrosWindow::OnApplyToProject(wxCommandEvent & event)
 {
+   AudacityProject *project = &mProject;
```
Paul-Licameli, line 1306: Mistakenly approved, revoked it.  See preceding requests for changes.

Paul-Licameli (changes requested):

Paul-Licameli:
Your branch also needs rebasing onto more recent master and a force-push.  GitHub reports conflict in Project.h.  But I think you should rewrite it anyway with no changes in Project.h.

Issue #1153, opened 2021-06-25
Extract lib-project


Paul-Licameli:
- [x] #1163 
- [ ] #1153

PR #1586, opened 2021-08-31
Extract lib project

Paul-Licameli:
Resolves: #1153 

Platform neutral library defining the AudacityProject object and associated ProjectStatus (which just holds some strings)

There is also some fixing of some possible null pointer dereferences of the global active project -- which might be null on macOS.


libraries/lib-project/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+]]#
+
+set( SOURCES
+   Project.cpp
+   Project.h
+   ProjectStatus.cpp
+   ProjectStatus.h
+)
+set( LIBRARIES
+   lib-registries-interface
```
crsib, line 19: I think all dependencies should be private here

src/commands/ScriptCommandRelay.cpp
```diff
@@ -57,6 +58,8 @@ static int ExecCommand(wxString *pIn, wxString *pOut, bool fromMain)
       // Wait for and retrieve the response
       *pOut = builder.GetResponse();
    }
+   else
+      *pOut = wxString{};
```
crsib, line 62: This seems like a change of behavior that needs to be tested accurately.

libraries/lib-project/CMakeLists.txt
Paul-Licameli: Why?

Transitive dependencies simplify many other CMakeLists.txt files and the generated graph of dependencies.

I make the exception only for wxBase for certain other reasons.

My intention (although I haven’t yet reimplemented it with all the other build system changes) was that some libraries would use only the wxBase subset, and even be forced to compile with only a subset of that — by predefining on the command line the inclusion guard symbols of the headers I wanted to disallow, such as wx/app.h (in wxBase but using the event loop, so not allowed for platform neutrality)

In that case only, making the wxBase library a private dependency was needed to avoid breaking compilation of higher level things that are allowed to use those headers.

src/ProjectFileIO.cpp
```diff
@@ -283,6 +284,24 @@ ProjectFileIO::ProjectFileIO(AudacityProject &project)
    mTemporary = true;
 
    UpdatePrefs();
+
+   // Make sure there is plenty of space for Sqlite files
+   wxLongLong freeSpace = 0;
```
crsib, line 289: I'm not sure why this change is a part of this pull request.

Why 100 Mb? Why 100 mb is written that way? 100 * 1024 * 1024 expresses it better.

Why `AbbreviatePath` is used? It is confusing on non Windows platforms and not entirely correct on Windows as well. NTFS is much more complicated than just drive letters.

Also I have strong opinion that disk space/writability errors should allow users to easily change the problematic location, which is already implemeted in Audacity anyway. The current message is unhelpfull, confusing and sometimes even wrong.

libraries/lib-project/CMakeLists.txt
Paul-Licameli: That is: wxBase was meant to be a CMake target enforcing this restricted view of the wxBase library to the really platform neutral part (no event loop) so I can be sure libraries using it are really neutral.

crsib (changes requested):
I do not understand why the size check is in the list of changes and I do not agree with the implementation of the check.

libraries/lib-project/CMakeLists.txt
crsib: For the sake of the correct dependency graph, I really think that every dependency should be listed explicitly, not implicitly by some other dependency. The linker will not drop such dependencies. The compiler will have to deal with the larger set of include directories. So in the end the dependency graph won't reflect the actual state of the dependencies.

To clarify my case: consider a library, that depends on `lib-project`. It should explicitly list the dependency on `lib-registries-interface` if it uses the library API internally, rather than hide this dependency somewhere in the middle of the graph. This way it will be clear, which library actually depends on `lib-registries-interface`.

src/ProjectFileIO.cpp
crsib: Ok, I see now why it is a part of the request, I don't agree with implementation still. We should open an issue about improving this error.

crsib (approved):

libraries/lib-project/CMakeLists.txt
Paul-Licameli: I simply disagree.  I want my generated graphs to have paths of dependency of A on B, which is good enough, but not a clutter of too many crossing edges making them too difficult to understand.

Please do not insist on this before approving the commit.

src/commands/ScriptCommandRelay.cpp
Paul-Licameli: I have not tested it but this only involves mod-script-pipe which I believe could crash without this change in case of macOS with the application still running but not project window open.

The unchecked assumption that GetActiveProject returns non-null is not a safe assumption and I recall removing it in other case.

src/ProjectFileIO.cpp
Paul-Licameli: You understand the need to cut and paste the code that existed to some other place to break dependency.  I'm not changing the implementation.

src/commands/ScriptCommandRelay.cpp
Paul-Licameli: A review of the uses of GetActiveProject() isn't strictly in the scope of the problem of this library extraction.  Still I did it incidentally because I recall other bugs with use of that functon, and I changed the return type to a weak pointer to be sure there not not dangling non-null pointers.

I have really disliked GetActiveProject() and in the past eliminated many uses of it in favor instead of a project reference passed in CommandContext.

Paul-Licameli: Still GetActiveProject() will require use in a few places where CommandContext is itself constructed.

PR #1594, opened 2021-09-01
Some audio io refactoring

Paul-Licameli:
Resolves: #1580

Some refactoring tasks in AudioIO, before disentangling the MIDI playback parts

### DIFF
--- a/src/AudioIO.cpp
+++ b/src/AudioIO.cpp
@@ -2899,9 +2899,6 @@ void AudioIO::FillBuffers()
       // by not always trying to process tiny chunks, eating the
       // CPU unnecessarily.
       //
-      // The exception is if we're at the end of the selected
-      // region - then we should just fill the buffer.
-      //
       // May produce a larger amount when initially priming the buffer, or
       // perhaps again later in play to avoid underfilling the queue and falling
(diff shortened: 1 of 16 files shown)


src/AudioIO.cpp
```diff
@@ -3812,7 +3828,7 @@ int audacityAudioCallback(const void *inputBuffer, void *outputBuffer,
 {
    auto gAudioIO = AudioIO::Get();
    return gAudioIO->AudioCallback(
-      inputBuffer, outputBuffer, framesPerBuffer,
+      (constSamplePtr)inputBuffer, (float*)outputBuffer, framesPerBuffer,
```
crsib, line 3831: static_casts maybe?

crsib (approved):

src/AudioIO.cpp
Paul-Licameli: Yes.

I don't like the type aliases samplePtr and constSamplePtr, for char* and const char*, which often travel together with a sampleFormat telling you how to interpret the data.  Better to put the pointer and the format in a class.  But doing that cleanup, which will eliminate still more casts, is for later.

Issue #1258, opened 2021-07-08
Renaming clips


jounih:
### User Story
As a developer, I would like to know the UX details for implementing clip names and renaming.

### Description

Currently there is no concept of clip names in Audacity, only track names. Individual clip names are very useful when organising a project, or separating between copies or takes.

The new affordance UX has a space for a clip name top left, below are the designs which allow the user to change the clip name.

For keyboard navigation Tabbing should select a clip, and "Enter" could edit the clip name.

Audio/Midi clips:

![Clip names@2x](https://user-images.githubusercontent.com/10109867/126309380-68a8db34-691a-4eb7-a6ca-4efd0807d102.jpg)

Label clips:

Described here https://github.com/audacity/audacity/issues/1195

### Accessibility

Keyboard shortcuts for selecting and editing audio/midi clip names and label clip names should be identical. Tab to select/forward cycle selection, enter to edit currently selected.

Furthermore, they should both respect the accessibility preference for editing the name in a dialog: Tracks->Track Behaviours-> Use dialog for the name of a new label (Also this should be renamed to "Use dialog for the name of a new label or clip"

DavidBailes:
Re: keyboard navigation:

1. Clip selection would be better than clip focus for renaming. For example, in file managers, if the focused and selected items are different and you press F2, then the name of the selected item is opened for editing. In addition, using the selected item (or items?) allows for the possibility of easily being able to rename multiple clips if that is ever required. If selection is used, then using tab is no longer appropriate.

2. However (!), proper selection of clips if not yet implemented. If this is still the case when renaming clips is implemented, as an alternative to the selected clip, you could use the clip which contains the cursor in the focused track. There already are commands for selecting (time range) next/previous clip, and commands for moving the cursor to next/previous clip boundary.

3. I'm assuming that a custom text box will be used for editing the clip name. Unfortunately, on Windows and using the wxWidgets accessibility framework this can't be made accessible for users of screen readers. So there will need to be the option of using a standard text box in a dialog box. Such an option is already present in Audacity for entering the name of a label when it's created. (Preferences-> Tracks Behaviors -> Use dialog for the name of a new label.)

4. Enter toggles the selectedness of the track, so another shortcut would have to be used for the rename clip command.

Paul-Licameli:
Can this design really be considered done, @AnitaBats ?  David's points are unaddressed.

jounih:
Will be reviewing David's feedback today @Paul-Licameli

[...]

PR #1614, opened 2021-09-03
Wave Clip rename

vsverchinsky:
Resolves: #1258 


src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
```diff
@@ -109,6 +110,9 @@ void NoteTrackAffordanceControls::Draw(TrackPanelDrawingContext& context, const
                 py >= clipRect.GetTop() && py <= clipRect.GetBottom());
 
         context.dc.SetClippingRegion(rect);
+        context.dc.SetTextBackground(wxTransparentColor);
+        context.dc.SetTextForeground(theTheme.Colour(clrClipNameText));
+        context.dc.SetFont(wxFont(wxFontInfo()));
```
Paul-Licameli, line 115: Only changing note track affordance text colors?

src/tracks/ui/TextEditDelegate.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
Paul-Licameli, line 1: Add header comments as in other headers describing the purpose of the file, and use
@file and @brief comments for Doxygen.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -20,61 +21,183 @@
...
-    : CommonTrackCell(pTrack)
+    : CommonTrackCell(pTrack), mClipNameFont(wxFont(wxFontInfo()))
 {
+    if (auto waveTrack = dynamic_cast<WaveTrack*>(pTrack.get()))
+    {
+        //Unbind is unnecessary since wave track is 
+        //destroyed along with affordance cell
+        waveTrack->Bind(EVT_WAVETRACK_SELECTION,
+            &WaveTrackAffordanceControls::OnTrackChanged,
+            this);
```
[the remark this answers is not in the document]
Paul-Licameli: Unbind() is also unnecessary when the event "sink" object in the Bind() call also inherits wxEvtHandler.  Then when it is destroyed, the "source" wxEvtHandler is also notified.

(I'm not certain, but it may in fact be enough for the sink to inhert wxTrackable, a base class of wxEvtHandler.)

Can the text edit helper inherit wxEvtHandler and bind itself to the TrackList?  Then it can destroy itself in response to the event.

Paul-Licameli (changes requested):
I raise questions about the new event passing.  I think the new event type is unnecessary and should not be added.  Another implementation can be done with existing events.

This is not a complete review, but you can begin with these reqeusted changes.

src/tracks/ui/TextEditDelegate.h
```diff
@@ -0,0 +1,20 @@
...
+
+class wxString;
+class wxPoint;
+
+class AudacityProject;
+
+//Used as a delegate for TextEditHelper
+class TextEditDelegate
```
Paul-Licameli, line 9: Perhaps "TextEditPolicy".  This looks like the "Policy" pattern, also called "Strategy," where an object supplies a set of callbacks to a reusable framework.

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
+/*!********************************************************************
+*
+ Audacity: A Digital Audio Editor
+
+ TextEditHelper.h
+
+ Vitaly Sverchinsky
+
+ The major part of the logic is extracted from LabelTrackView.cpp and
+ LabelTextHandle.cpp
```
Paul-Licameli, line 10: This logic was DUPLICATED from LabelTrackView and LabelTextHandle.

But why duplicate, when we might really reuse?

Can it be given a better name like "TextEditingCell" and could it first be split out and used in LabelTrackView?

Paul-Licameli (changes requested):
More code reuse, less duplication?

Paul-Licameli (changes requested):
More code reuse, less duplication?

src/tracks/ui/TextEditDelegate.h
```diff
@@ -0,0 +1,20 @@
...
+{
+public:
+
+    virtual ~TextEditDelegate() = default;
```
Paul-Licameli, line 13: I recommend:  Combine this header with TextEditHelper.h, because the classes cooperate closely.  Neither is useful without the other.

Then, define the destructor of TextEditDelegate as non-inline.

If a class has virtual functions, it is better that at least one of them is defined non-inline, and the destructor can be the one if all the others are pure virtuals.  (Actually pure virtual functions cal also be given definitions, but that's another story.)

It is better because then it can be decided at compile time, in what single place, the virtual function table of this base class will be compiled.

Whereas otherwise, the compiler must duplicate the virtual function table definition, and the linker needs to remove all but one duplicate.

So avoid the bit of extra work for the linker.

Paul-Licameli (changes requested):
Advice about file structure and non-inline virtual destructor.

src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
vsverchinsky: In new commit history this commit contains also wave track colors changes

src/WaveTrack.cpp
```diff
@@ -270,6 +272,18 @@ void WaveTrack::SetPanFromChannelType()
...
+{
+   bool selected = GetSelected();
+   PlayableTrack::SetSelected(s);
+   if (selected != GetSelected()) {
+      TrackListEvent evt {
+         EVT_WAVETRACK_SELECTION, shared_from_this(), {}
+      };
+      ProcessEvent(evt);
+   }
+}
```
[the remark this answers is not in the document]
vsverchinsky: `WaveClipAffordanceControls` now inherits `wxEvtHandler` and this fixes issue with event processing and objects lifetime

src/tracks/ui/TextEditDelegate.h
vsverchinsky: No, it's not a policy. It could be a `TextEditListener` or replaced with wxEvent dispatch

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: It's not a cell because it does not define TrackPanel separation as other cells do.
There is another PR which addresses an issue with code duplication

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
```diff
@@ -15,12 +15,14 @@
...
+class TrackPanelResizeHandle;
+class WaveTrackAffordanceHandle;
 
 class AUDACITY_DLL_API WaveTrackAffordanceControls : public CommonTrackCell
 {
     std::weak_ptr<WaveClip> mFocusClip;
-    std::weak_ptr<AffordanceHandle> mAffordanceHandle;
-    std::weak_ptr<UIHandle> mResizeHandle;
+    std::weak_ptr<WaveTrackAffordanceHandle> mAffordanceHandle;
+    std::weak_ptr<TrackPanelResizeHandle> mResizeHandle;
```
Paul-Licameli, line 40: I see the purpose of this change and the commit that introduces it is to fix a "slicing problem" in AssignUIHandlePtr, where an object of derived type may be assigned to an object of base type, losing information.  Yes?

I don't know why I didn't see this earlier in review!

I suggest this additional change (add it to the label refactoring branch maybe):

In AssignUIHandlePtr, in the else-branch, add

```
      wxASSERT(typeid(*ptr) == typeid(*pNew));
```
You may also need
```
#include <typeinfo>
```

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -33,13 +33,6 @@ class WaveTrackAffordanceHandle final : public AffordanceHandle
 public:
     WaveTrackAffordanceHandle(const std::shared_ptr<Track>& track) : AffordanceHandle(track) { }
 
-    static UIHandlePtr HitAnywhere(std::weak_ptr<AffordanceHandle>& holder, const std::shared_ptr<Track>& pTrack)
-    {
-        auto result = std::static_pointer_cast<AffordanceHandle>(std::make_shared<WaveTrackAffordanceHandle>(pTrack));
-        result = AssignUIHandlePtr(holder, result);
```
Paul-Licameli, line 38: This change is good, but incomplete!

Similar code with the same slicing error is still not corrected in NoteTrackAffordanceControls.

Paul-Licameli (changes requested):

src/tracks/ui/TextEditHelper.h
```diff
@@ -0,0 +1,83 @@
...
+    wxFont mFont;
+    wxRect mBBox;
+
+    wxColor mTextSelectionColor;
+    wxColor mTextColor;
+
+    bool mRightDragging{ false };
+    int mOffset{ 0 };
+    int mInitialCursorPos{ 0 };
+    int mCurrentCursorPos{ 0 };
```
Paul-Licameli, line 40: Can you state in comments any intended invariant relations among these values?

Is it meant to be invariant that `mInitialCursorPos <= mCurrentCursorPos` ?

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+
+std::pair<int, int> TextEditHelper::GetSelection() const
+{
+    return std::make_pair(mInitialCursorPos, mCurrentCursorPos);
+}
+
+void TextEditHelper::SetSelection(int from, int to)
+{
+    mInitialCursorPos = from;
+    mCurrentCursorPos = to;
```
Paul-Licameli, line 82: Should you comment before this function,
```
@pre from <= to
```

Or swap the values before the assignment to make the condition hold?

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
```diff
@@ -38,15 +57,28 @@ class AUDACITY_DLL_API WaveTrackAffordanceControls : public CommonTrackCell
...
         AudacityProject* project) override;
 
+    unsigned Char
+    (wxKeyEvent& event, ViewInfo& viewInfo, wxWindow* pParent,
+        AudacityProject* project) override;
+
+    virtual void OnTextEditFinished(AudacityProject* project, const wxString& text);
+    virtual void OnTextEditCancelled(AudacityProject* project);
+    virtual void OnTextModified(AudacityProject* project, const wxString& text);
+    virtual void OnTextContextMenu(AudacityProject* project, const wxPoint& position);
```
Paul-Licameli, line 75: Please use `override` not `virtual`.

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+
+    int offsetX{ 0 };
+    if (mOffset > 0)
+    {
+        offsetX = dc.GetTextExtent(mText.Left(mOffset)).GetWidth();
+    }
+
+    if (wxTheApp->GetLayoutDirection() == wxLayout_RightToLeft)
+    {
+        if (index >= length)
```
Paul-Licameli, line 528: Should this be index <= 0?

Paul-Licameli (changes requested):

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+                // Actually this might be right or middle down
+                mRightDragging = true;
+            }
+
+#if defined(__WXGTK__) && (HAVE_GTK)
+            if (evt.MiddleDown()) {
+                // Paste text, making a NEW label if none is selected.
+                wxTheClipboard->UsePrimarySelection(true);
+                view.PasteSelectedText(project, newSel.t0(), newSel.t1());
+                wxTheClipboard->UsePrimarySelection(false);
```
Paul-Licameli, line 303: I don't know if it is important, but I think this if-block needs to move down, outside one more } , to correct the analogy with LabelTextHandle::HandleTextClick:  Do it whether bounding box contains the position, or not.

Paul-Licameli (changes requested):

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+    if (mCurrentCursorPos == mInitialCursorPos)
+    {
+        dc.SetPen(mTextColor);
+        auto top = rect.GetTop() + (rect.GetHeight() - cursorHeight) / 2;
+        dc.DrawLine(curPosX, top, curPosX, top + cursorHeight);
+    }
+
+    dc.DestroyClippingRegion();
+    if (restoreClippingRegion)
+        dc.SetClippingRegion(prevClippingRegion);
```
Paul-Licameli, line 391: My preference is to put cleanup lines like the above three inside
```
auto cleanup = finally([&]{ ... });
```
And put the declaration of cleanup higher, just after the first SetClippingRegion().

This is ad-hoc RAII.  `finally` generates a class that performs the lambda in its destructor, and binds an instance to `cleanup`.

This makes it easier to be confident that a long function isn't making a mistake somewhere, with an early exit that forgets to restore the context.  Or, in case of exceptions.

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+        auto leftBound = rect.GetLeft();
+        auto rightBound = rect.GetRight() + 1;
+        GetCharPositionX(mCurrentCursorPos, &curPosX);
+
+        if ((!rtl && curPosX >= rightBound) || (rtl && curPosX < leftBound))
+        {
+            while (mOffset < maxOffset)
+            {
+                GetCharPositionX(mCurrentCursorPos, &curPosX);
+                if (curPosX < rightBound && curPosX >= leftBound)
```
Paul-Licameli, line 349: Is this condition correct in case of right-to-left layout?

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+                    break;
+                ++mOffset;
+            }
+        }
+        if ((!rtl && curPosX < leftBound) || (rtl && curPosX >= rightBound))
+        {
+            while (mOffset > 0)
+            {
+                GetCharPositionX(mCurrentCursorPos, &curPosX);
+                if (curPosX >= leftBound && curPosX < rightBound)
```
Paul-Licameli, line 359: ditto ?

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+                    break;
+                --mOffset;
```
Paul-Licameli, line 361: I think mOffset needs comments about what it really is.  I'm not sure I understand it.

src/tracks/ui/TextEditHelper.cpp
```diff
@@ -0,0 +1,666 @@
...
+            }
+        }
+    }
+
+    if (mCurrentCursorPos != mInitialCursorPos)
+    {
+        auto left = 0;
+        auto right = 0;
+        GetCharPositionX(std::min(mCurrentCursorPos, mInitialCursorPos), &left);
+        GetCharPositionX(std::max(mCurrentCursorPos, mInitialCursorPos), &right);
```
Paul-Licameli, line 371: Are the above two lines right?  Isn't current >= initial always?

Is what you really want to do:  get char positions for initial and current, as left and right, but then swap in case of RTL layout?  Because, swap is needed in that case and that case only?

Did you really try all of this out with Arabic?  Putting breakpoints at every test of layout, and making sure all are hit?

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -52,6 +57,32 @@ class WaveTrackAffordanceHandle final : public AffordanceHandle
     }
 };
 
+class SetWaveClipNameCommand : public AudacityCommand
```
Paul-Licameli, line 99: I see you made this command by analogy with SetTrackNameCommand in CommonTrackControls.cpp, and like that command class, it isn't meant to be registered with CommandManager.

It is commented there, that the same effect can be achieved with other commands available to macro programming.

Is that so for change of clip name?

No.  So I suggest: the existing SetClipCommand class needs to be extended with another parameter for setting the name, by means of macro programming.

I know that wasn't part of the specification of this project, but it is something that should be done for completeness: exposing all changes of properties we can to macro programmers too.

This could be a separat PR.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -52,6 +57,32 @@ class WaveTrackAffordanceHandle final : public AffordanceHandle
...
+{
+public:
+    static const ComponentInterfaceSymbol Symbol;
+
+    ComponentInterfaceSymbol GetSymbol() override
+    {
+        return Symbol;
+    }
+    void PopulateOrExchange(ShuttleGui& S) override
```
Paul-Licameli, line 108: Is this your first experience writing some ShuttleGui code?

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -145,6 +176,28 @@ void WaveTrackAffordanceControls::Draw(TrackPanelDrawingContext& context, const
...
+
+        SetWaveClipNameCommand Command;
+        auto oldName = clip->GetName();
+        Command.mName = oldName;
+        auto result = Command.PromptUser(&GetProjectFrame(*project));
+        if (result && Command.mName != oldName)
+        {
+            clip->SetName(Command.mName);
+            ProjectHistory::Get(*project).PushState(XO("Modified Clip Name"),
+                XO("Clip Name Edit"), UndoPush::CONSOLIDATE);
```
Paul-Licameli: Whenever you add XOs, make sure Jouni agrees with the exact words.

src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
```diff
@@ -1098,6 +1098,33 @@ auto WaveTrackView::GetSubViews(const wxRect* rect) -> Refinement
    return results;
 }
 
+unsigned WaveTrackView::CaptureKey(wxKeyEvent& event, ViewInfo& viewInfo, wxWindow* pParent, AudacityProject* project)
+{
+    if (auto affordance = GetAffordanceControls())
```
Paul-Licameli, line 1103: This function, and the two following, may cause. Return key, while the mouse is in a wave track, to edit a clip name.

Was that part of the design?

I think the more correct way to do call-through:  call through to the affordance, but then check event.GetSkipped().  If it is true, meaning the affordance did not take its first-chance to capture the key, then call through to the base class in case it wants to capture.

Take the bitwise OR of the return from the two captureKey calls, as the return value of the function.

It may not make a difference now, because nothing really happens in the base class methods, but don't assume that will remain so.

Make similar fixes in the next two functions.

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -220,6 +220,9 @@ unsigned WaveTrackAffordanceControls::KeyDown(wxKeyEvent& event, ViewInfo& viewI
     auto keyCode = event.GetKeyCode();
     switch (keyCode)
     {
+    case WXK_TAB:
+        SelectNextClip(viewInfo, project);
+        break;
```
Paul-Licameli: Tab key navigation didn't work when I tried it, but the simple missing piece, is a change of WaveTrackAffordanceControls::CaptureKey so it does not skip the event for tab key.

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
Paul-Licameli: Do we also want to implement leftward navigation among clips for Shift + TAB, as is done for labels?

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -290,3 +473,12 @@ bool WaveTrackAffordanceControls::StartEditSelectedClipName(ViewInfo& viewInfo,
     }
     return false;
 }
+
+std::shared_ptr<TextEditHelper> WaveTrackAffordanceControls::MakeTextEditHelper(const wxString& text)
+{
+    auto helper = std::make_unique<TextEditHelper>(shared_from_this(), text, mClipNameFont);
```
Paul-Licameli, line 479: make_shared ?

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -182,16 +296,28 @@ bool WaveTrackAffordanceControls::StartEditClipName(AudacityProject* project)
...
+                ProjectHistory::Get(*project).PushState(XO("Modified Clip Name"),
+                    XO("Clip Name Edit"), UndoPush::CONSOLIDATE);
+
+                return true;
+            }
+        }
+        else
+        {
+            mTextEditHelper = MakeTextEditHelper(clip->GetName());
+            mFocusClip = lock;
```
Paul-Licameli, line 320: That's a strange line... I think it does nothing

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -83,15 +122,71 @@ class SetWaveClipNameCommand : public AudacityCommand
 const ComponentInterfaceSymbol SetWaveClipNameCommand::Symbol
 { XO("Set Wave Clip Name") };
 
+//Handle which is used to mouse events to TextEditHelper
```
Paul-Licameli, line 125: "used to send"

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -83,15 +122,71 @@ class SetWaveClipNameCommand : public AudacityCommand
...
+    {
+        static auto ibeamCursor =
+            ::MakeCursor(wxCURSOR_IBEAM, IBeamCursorXpm, 17, 16);
+        return {
+           XO("Click and drag to select text"),
+           ibeamCursor.get()
+        };
+    }
+
+    Result Release(const TrackPanelMouseEvent& event, AudacityProject* project, wxWindow*)
```
Paul-Licameli, line 159: override

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -83,15 +122,71 @@ class SetWaveClipNameCommand : public AudacityCommand
...
+    {
+        if (mHelper->OnRelease(event.event, project))
+            return RefreshCode::RefreshCell;
+        return RefreshCode::RefreshNone;
+    }
+
+    Result Cancel(AudacityProject* project)
```
Paul-Licameli, line 166: override

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
```diff
@@ -11,18 +11,37 @@
...
+    public TextEditDelegate,
+    public wxEvtHandler,
+    public std::enable_shared_from_this<WaveTrackAffordanceControls>
 {
     std::weak_ptr<WaveClip> mFocusClip;
     std::weak_ptr<WaveTrackAffordanceHandle> mAffordanceHandle;
     std::weak_ptr<TrackPanelResizeHandle> mResizeHandle;
+    std::weak_ptr<WaveClipTitleEditHandle> mTitleEditHandle;
+
+    std::shared_ptr<TextEditHelper> mTextEditHelper;
```
Paul-Licameli, line 43: Now I understand why TextEditHelper has a weak pointer to its delegate.  If it were shared, we would have a cycle of shared pointers which is a bad thing.

The helper is owned by its delegate and not the opposite.

Perhaps comments in those text edit classes should clarify this point of lifetime management?

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -26,33 +26,72 @@
...
+            return RefreshCode::RefreshNone;
         }
+        else
+            return AffordanceHandle::Click(event, project);
+    }
+
+    Result Release(const TrackPanelMouseEvent& event, AudacityProject* project, wxWindow* parent) override
+    {
+        if (mDoubleClickFlag)
+            return RefreshCode::RefreshNone;
```
Paul-Licameli, line 75: You can simplify.  Double Click can OR the return value with Cancelled and then you don't need mDoubleClickFlag.  See the much repeated comment "Do not start a drag" in the source tree for other examples.

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -83,15 +122,71 @@ class SetWaveClipNameCommand : public AudacityCommand
...
+        return RefreshCode::RefreshAll;
+    }
+};
+
 WaveTrackAffordanceControls::WaveTrackAffordanceControls(const std::shared_ptr<Track>& pTrack)
     : CommonTrackCell(pTrack), mClipNameFont(wxFont(wxFontInfo()))
 {
+    if (auto trackList = pTrack->GetOwner())
+    {
+        trackList->Bind(EVT_TRACKLIST_SELECTION_CHANGE,
```
Paul-Licameli, line 182: Now that you bind to the track list, the function OnTrackChanged will take its effect whenever any track changes it selected state.  Is that the intention, or should it check only for changes of the relevant wave track?

Paul-Licameli:
This round of review is now COMPLETE!  The overall approach is sound, but I flagged many details that need attention.  None of these is difficult by itself.

src/tracks/ui/TextEditHelper.h
vsverchinsky: Not neccessary that `mInitialCursorPos <= mCurrentCursorPos` it depends on direction in which selection is moving: from left to right or right to left. Variables have the same meaning as they do in LabelTrackView

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: Yes, it should, thanks!

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: Agreed, thanks!

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: Yep, don't work as expected with RTL layout and LTR language

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: Will add a comment

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: >Are the above two lines right? Isn't current >= initial always?

Not always, depends on direction of cursor movement

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
vsverchinsky: Sort of

src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
vsverchinsky: >Was that part of the design?

Yes, though editing will be invoked only if there is a "selected" clip on that track.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
vsverchinsky: Yes, it's intentional, thought was not specified by design document, so I made the behaviour similar to the behaviour in the label track, whenever it was possible (that was specified)

src/tracks/ui/TextEditHelper.cpp
Paul-Licameli: But I thought mCurrentCursorPos and mInitialCursorPos are logical positions in a string of characters, not x coordinates on a screen.  If they are not that,  what are they?

src/tracks/ui/TextEditDelegate.h
vsverchinsky: > Combine this header with TextEditHelper.h, because the classes cooperate closely

They do, but it's not neccessary that owner of the TextEditHelper and TextEditDelegate are the same object, and in the latter case it's better to keep it in a separate file

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
vsverchinsky: Thanks! fixed

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: I'm not sure that undoing clipping region in the case of exception would be a safe operation

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
vsverchinsky: Fixed, tabbing works for me on Windows, but didn't test it on other platforms.
Also added a Shift+TAB, thanks

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
vsverchinsky: Thanks!

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: Yes, they are a positions in a string. mInitialCursorPos is not moving and initialized when interaction is started, then, when user moves cursor either with mouse (holding a left button), or with Shift+Arrow Left/Right mCurrentCursorPos is used to track the "current" cusror position

src/tracks/ui/TextEditDelegate.h
Paul-Licameli: I still like non-inline virtual destructors (really, virtual anything should never be inlined, because it will usually never be inlined).

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
Paul-Licameli: This suggestion wasn't implemented yet

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -211,26 +315,86 @@ unsigned WaveTrackAffordanceControls::CaptureKey(wxKeyEvent& event, ViewInfo& vi
 {
     const auto keyCode = event.GetKeyCode();
     if (!(keyCode == WXK_RETURN || keyCode == WXK_NUMPAD_ENTER || keyCode == WXK_TAB))
+    if (!mTextEditHelper && !(keyCode == WXK_RETURN || keyCode == WXK_NUMPAD_ENTER || keyCode == WXK_TAB))
```
Paul-Licameli, line 318: I think you meant to replace the line before this new line.  Not to add a new line.  But the code is correct!  Just not simplified

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -228,6 +231,54 @@ unsigned WaveTrackAffordanceControls::KeyDown(wxKeyEvent& event, ViewInfo& viewI
     return RefreshCode::RefreshCell;
 }
 
+namespace {
+    template<typename Iter, typename Comp>
+    WaveClip* NextClipLooped(ViewInfo& viewInfo, Iter begin, Iter end, Comp comp)
```
Paul-Licameli, line 400: Nice use of a template to capture the common logic of the forward and backward cases!

Instead of `Comp comp` I would write `Pred pred` because that argument is a unary predicate, not a binary comparator.  But that's only variable names.

src/tracks/ui/TextEditHelper.cpp
vsverchinsky: Fixed with wxDCClipper

Paul-Licameli (approved):

PR #1635, opened 2021-09-07
Update BUILDING.md

thynus:
error:
CMake Error: Could not create named generator XCode

correction:
cmake -GXcode -T buildsystem=1 -Daudacity_use_mad="off" -Daudacity_use_id3tag=off ../audacity

Recommended:


Paul-Licameli:
Thank you for pointing this out.

I am closing this and instead I opened #1652 which makes other needed changes to BUILDING.md.  It should also say you need a C++17 compiler, not 14.

Issue #1628, opened 2021-09-06
VSTEffect samplePos should be a value in number of samples, not in seconds


Niko-Sound:
**Describe the bug**
VSTEffect mTimeInfo.samplePos should be a value in number of samples, not in seconds

In the file "aeffectx.h" from the Steinberg vst 2.4 SDK, it is written:
```

struct VstTimeInfo
{
	double samplePos;				///< current Position in audio samples (always valid)
```

In the file VSTEffect.cpp, the value is computed in seconds:
```

mTimeInfo.samplePos += ((double) blockLen / mTimeInfo.sampleRate);

```
It should be:
```

mTimeInfo.samplePos += blockLen;

```

**To Reproduce**
1. Inside a VST plugin, get the time info  with a callback using "audioMasterGetTime"
2. Get the samplePos from the time info

**Expected behavior**
The samplePos should be in samples, not in seconds

**Additional information (please complete the following information):**
 - OS: Ubuntu 20.10
 - Version Audacity 3.0.4

LWinterberg:
This line has been there since forever (since we no longer use svn anyway), and we're not aware of VST2 problems arising from this, so I'm marking this P4. 

(If you want to make a PR that fixes this, feel free to!)

Niko-Sound:
Pull request done! *(https://github.com/audacity/audacity/pull/1636)*

Some additional information: the VST plugin used for testing is a kind of scrolling spectrogram viewer, with a scrolling time axis at the bottom. It has been tested on a certain number of DAWs, getting the sample pos from the DAW in order to synchronize the time axis. For all these other tests, the sample pos unit is in number of samples.

LWinterberg:
Cool, cool! I'll try to get someone who actually knows C++ to review this. 

While you're at it, are you aware of other bugs/spec violations of our VST2 implementation that we maybe haven't discovered yet? I'm asking because we're planning to add proper real-time effects support for 3.1 (see also #992)

Paul-Licameli:
No!  3.2 is the target for that.

Niko-Sound:
Thank you :)

No, no particular problem detected after having tested with more than 20 VST plugins (with some of them having GUI resize features). I find that your VST support is great! And your plan to make real-time stackable effects looks awesome!

Just an idea, if it is not already there, it could be interesting to feed the plugins with silence when not playing. In order to manage the kind of plugins that capture some sound, edit it and play it, but all this inside the plugin, then finally render the edited result to the host.

[...]

PR #1636, opened 2021-09-07
Fix VST time info samplePos units

Niko-Sound:
Resolves: *(https://github.com/audacity/audacity/issues/1628)*

*VST time info samplePos should be a value in number of samples, not in seconds*

Recommended:


Paul-Licameli (approved):
A simple one liner.  Though finding the fix may not have been easy.  Thank you

PR #1637, opened 2021-09-08
Disentangle midi play from audio io

Paul-Licameli:
Resolves: #1581

Next refactoring steps for AudioIO.

All of the logic specific to MIDI playback is removed from AudioIO.cpp, and invoked instead through an abstract base class AudioIOExt.  A new, separate file MIDIPlay.cpp registers its extension policy with AudioIO.

These steps shrink the big file AudioIO.cpp from 4797 to 3596 lines, and are also preparations for putting Note track into its own optional extension module.

The unnecessary separate thread for MIDI playback is removed for simplicity, and some callbacks happen directly in the Portaudio thread.  That means some memory allocations may happen on that thread, when MIDI plays!  (See PrepareMidiIterator().) But future refactorings will correct that.

The dependency of lib-audio-devices on portmidi is severed.

A few other changes remove some dependencies on NoteTrack, and use Settings objects from Prefs.cpp to eliminate some duplicated literals and direct use of gPrefs in MIDIPlay.cpp.

Recommended:


Paul-Licameli:
Now rebased, with build fixes, and also removing use of deprecated std::iterator.

src/AudioIOExt.h
```diff
@@ -0,0 +1,60 @@
...
+   //! Typically statically constructed
+   struct AUDIO_DEVICES_API RegisteredFactory{
+      explicit RegisteredFactory(Factory factory);
+      ~RegisteredFactory();
+   };
+
+   virtual ~AudioIOExt();
+
+   // Formerly in AudioIoCallback
+   virtual void ComputeOtherTimings(double rate,
```
crsib, line 45: Maybe we should something else in place of *Other*? Like ExtensionStream or whatever?

crsib (approved):
MidiCode is much easier to understand now!

Issue #1632, opened 2021-09-07
Journal fails to record presses of toolbar buttons and sliders


Paul-Licameli:
QA should test the following with journals enabled, and starting up Audacity four different ways:
- [ ] toolbars in default positions.
- [ ] transport toolbar (play, record, etc) docked at the bottom.
- [ ] transport toolbar un-docked (free floating).
- [ ] transport and one other toolbar un-docked.

After starting, press, for instance, the Pause key.  (Not play or record, which require other work to be replayed in journals.  See #1591, #1592)

Observe a message that journal recording fails.

Expected: it should succeed and the resulting journal should replay successfully.

PR #1646, opened 2021-09-09
Journal toolbar buttons and sliders

Paul-Licameli:
Resolves: #1632

Attempting to record a journal, with presses of buttons or dragging of sliders in the toolbars, will no longer fail with an error message, no matter whether the toolbars are docked above or below or are free-floating.

Recommended:

### DIFF
--- a/src/ProjectWindow.h
+++ b/src/ProjectWindow.h
@@ -21,6 +21,7 @@ class Track;
 
 class wxScrollBar;
 class wxPanel;
+class wxPanelWrapper;
 
 class ProjectWindow;
 void InitProjectWindow( ProjectWindow &window );
@@ -53,8 +54,8 @@ public:
    void SetIsBeingDeleted() { mIsDeleting = true; }
 
    wxWindow *GetMainPage() { return mMainPage; }
-   wxPanel *GetMainPanel() { return mMainPanel; }
-   wxPanel *GetTopPanel() { return mTopPanel; }
+   wxPanel *GetMainPanel();
+   wxPanel *GetTopPanel();
 
    void UpdateStatusWidths();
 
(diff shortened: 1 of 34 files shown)


src/ProjectWindow.h
```diff
@@ -53,8 +54,8 @@ class AUDACITY_DLL_API ProjectWindow final : public ProjectWindowBase
    void SetIsBeingDeleted() { mIsDeleting = true; }
 
    wxWindow *GetMainPage() { return mMainPage; }
-   wxPanel *GetMainPanel() { return mMainPanel; }
-   wxPanel *GetTopPanel() { return mTopPanel; }
+   wxPanel *GetMainPanel();
```
ksoze95, line 57: Could these methods, GetMainPanel and GetTopPanel be made const?

src/export/ExportFFmpegDialogs.cpp
```diff
@@ -195,9 +195,13 @@ const std::vector< int > AC3BitRateValues{
 
 const int ExportFFmpegAC3Options::iAC3SampleRates[] = { 32000, 44100, 48000, 0 };
 
-ExportFFmpegAC3Options::ExportFFmpegAC3Options(wxWindow *parent, int WXUNUSED(format))
+ExportFFmpegAC3Options::ExportFFmpegAC3Options(
```
ksoze95, line 198: This is beyond the scope of this PR.

All the constructors for the various ExportFFmpeg options, apart from the ExportFFmpegCustomOptions, are identical.  Could these be refactored to use an inheriting constructor, say ExportFFmpegBase, with ExportFFmpegCustomOptions' mFormat and mCodec being default initialised.  However, you will need to implement a form of the template method pattern, maybe via CRTP, to call the derived class' PopulateOrExchange.

ksoze95 (changes requested):
ProjectWindow's GetMainPanel and GetTopPanel methods should be made const.

PR #1650, opened 2021-09-09
Context menu for picks in wave tracks between clips

Paul-Licameli:
These changes are very simple, after previous context menu work.

Right-click between clips brings up a small context menu with Paste and Mute/Unmute.

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
@@ -835,7 +835,11 @@ std::vector<ComponentInterfaceSymbol> WaveTrackSubView::GetMenuItems(
          // { L"", XO("Rename clip...") },
       };
    else
-      return {};
+      return {
+         { L"Paste", XO("Paste")  },
+         {},
+         { L"TrackMute", XO("Mute/Unmute Track") },
+      };
 }
 
 WaveTrackView &WaveTrackView::Get( WaveTrack &track )


petersampsonaudacity:
@Paul-Licameli  - surely the context menu that you get in the gray space between clips should be the same as the one you get when you right-click on a clip.  

The commands are not acting on the current mouse position, but rather on the selection or cursor position - thus the full set of commands is surely appropriate in both use-cases.

SteveDaulton:
Along the same lines as Peter's comment, it seems that the "clip context menu" actually functions as a "track panel context menu", but with "hot zones" where there are clips.

If I have multiple tracks with multiple clips in each, I would expect a "**clip** context menu" to  relate to the specific clip from which it was launched, just as a "**track** context menus" (track dropdown menu) relates to the track from which it was launched. That's not what happens - I can make a selection in any clip of any track, open the "clip context menu" from any clip in any track, and act on the selected clip.

What would  happen if "Clip name" were added to the clip context menu? Would that act on the clip which was right clicked, or on the clip(s) that are selected?

Surely the whole point of a "context menu" is the context from which it was launched.

Paul-Licameli:
> Along the same lines as Peter's comment, it seems that the "clip context menu" actually functions as a "track panel context menu", but with "hot zones" where there are clips.
> 
> If I have multiple tracks with multiple clips in each, I would expect a "**clip** context menu" to relate to the specific clip from which it was launched, just as a "**track** context menus" (track dropdown menu) relates to the track from which it was launched. That's not what happens - I can make a selection in any clip of any track, open the "clip context menu" from any clip in any track, and act on the selected clip.
> 
> What would happen if "Clip name" were added to the clip context menu? Would that act on the clip which was right clicked, or on the clip(s) that are selected?
> 
> Surely the whole point of a "context menu" is the context from which it was launched.

Yep, to all that.  The notion of selected or focused clip doesn’t really exist.  I think designers would like you to copy the clip as a unit, including the clip name.  It doesn’t work this way yet.

The design did ask for a renaming item in this clip context menu.  This does not exist yet because the UI for clip renaming, also reachable with left clicks in the name, is not past code review yet.

vsverchinsky (approved):

Issue #1624, opened 2021-09-05
Batch processing interrupted if sample rate too high for MP3 Export


SteveDaulton:
**Describe the bug**
When batch processing, if the export format is MP3 and the sample rate of the file is too high for the selected MP3 options, or greater than 48 kHz, then batch processing is interrupted - prompting the user to manually enter a lower sample rate.

**To Reproduce**
1. Apply "MP3 Conversion" macro to "Files" and select a file that has a sample rate > 48 kHz.
2. Observe  that the batch processing stops to ask for a lower sample rate.

**Expected behavior**
Audacity to automatically select an appropriate sample rate when MP3 encoding as  a batch process.

**Screenshots**
![prompt](https://user-images.githubusercontent.com/11681097/132141728-b0a071b8-c303-4fa4-8b40-4a398e0bed6e.png)

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version audacity-linux-3.0.4-x86_64.AppImage

**Additional context**
In most cases, the "suggested" sample rate of 8000 Hz is the worst possible option. Better to default to the highest supported sample rate rather than the lowest.

SteveDaulton:
There is now an experimental plug-in that overcomes this limitation (https://forum.audacityteam.org/viewtopic.php?p=434518#p434518) but it would be better if the solution was built in rather than requiring an additional plug-in.

PR #1656, opened 2021-09-10
Fixed issue #1624

pgirikishore:
Resolves: *(direct link to the issue)*https://github.com/audacity/audacity/issues/1624

The below solution was discussed and agreed upon by @SteveDaulton.
For a simple solution to the problem of batch processing being interrupted:
1. If the target bit rate is > 32 kbps and project rate <= 48 kHz, use the project rate (as now).
2. If the target bit rate is> 32 kbps and the project rate is > 48 kHz, use **48 kHz**.
3. If the target bit rate <= 32 kbps and project rate >  24 kHz, use **24 kHz**.
4. If the target bit rate <= 32 kbps and project rate <= 24 kHz, use the project rate (as now).

- [x] There are no behaviour changes unnecessary for the stated purpose of the PR

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes of behaviour


SteveDaulton:
I've not yet tried building this.
Is the user still prompted in cases 2 and 3 when manually exporting? I think they should be, as the "automatic" sample rate selection may not be optimal in all cases (but definitely much better than now for batch processing).

pgirikishore:
I was not aware that the user should still be prompted to select the sample rate in-case of manual exporting. Currently, the user is not prompted. I will look into that and fix that as well.

Paul-Licameli:
The change of behavior should depend on the batch mode flag stored in AudacityProject.  When it is false, do not change behavior.

src/export/ExportMP3.cpp
```diff
@@ -1859,7 +1858,17 @@ ProgressResult ExportMP3::Export(AudacityProject *project,
    // Verify sample rate
    if (!make_iterator_range( sampRates ).contains( rate ) ||
       (rate < lowrate) || (rate > highrate)) {
-      rate = AskResample(bitrate, rate, lowrate, highrate);
+       if (project->mBatch) {
```
Paul-Licameli, line 1861: Does this compile?  The member is mBatchMode.

That would correct the first version of your fix, for the interactive case, in which behavior should not change.

pgirikishore: Apologies. I think I might have accidentally changed it before pushing the code. I will fix it.

pgirikishore:
Hey @Paul-Licameli, Please review this PR when you have the time.

Paul-Licameli:
You merged audacity/master into your branch -- which we don't recommend.  To keep up to date, it's better to fetch master, rebase your branch onto it, and force-push (git push -f) the branch to your fork.

But in this case, the changes are so small, I would rather you just use interactive rebase (git rebase -i) to "squash" all into one small commit.

Use multiple commits if there is complicated transformation of the code that should be broke into reviewable steps (not so here); but don't leave multiple commits that are a record of your corrections of earlier attempts that were not correct.

src/export/ExportMP3.cpp
```diff
@@ -94,6 +94,7 @@
 #include "../widgets/AudacityMessageBox.h"
 #include "../widgets/ProgressDialog.h"
 #include "wxFileNameWrapper.h"
+#include "../../libraries/lib-project/Project.h"
```
Paul-Licameli, line 97: Don't write that, simply write
```
#include "Project.h"
```

src/export/ExportMP3.cpp
```diff
@@ -1859,7 +1859,17 @@ ProgressResult ExportMP3::Export(AudacityProject *project,
    // Verify sample rate
    if (!make_iterator_range( sampRates ).contains( rate ) ||
       (rate < lowrate) || (rate > highrate)) {
-      rate = AskResample(bitrate, rate, lowrate, highrate);
+       if (project->mBatchMode) {
+           if (bitrate > 32) {
+               rate = 48000;
+           }
+           else if (bitrate <= 32) {
+               rate = 240000;
```
Paul-Licameli, line 1867: You meant 24000?

Paul-Licameli: Is the intention to default to `highrate` appropriate for the bit rate?  Then say so in a comment.

Paul-Licameli (changes requested):
I'm done with the review.  There is an error and should be some comments.

src/export/ExportMP3.cpp
SteveDaulton: Perhaps clear without resorting to comments like this?
```
   // Verify sample rate
   if (!make_iterator_range( sampRates ).contains( rate ) ||
      (rate < lowrate) || (rate > highrate)) {
         if (project->mBatchMode) {
            if (rate == 0) {  // Maybe possible from macro command?
               return ProgressResult::Cancelled;
            }
            else if (rate < lowrate) {
               rate = lowrate;
            }
            else {
               rate = highrate;
            }
         }
         else {
            rate = AskResample(bitrate, rate, lowrate, highrate);
            if (rate == 0) {
               return ProgressResult::Cancelled;
         }
      }
   }

```

SteveDaulton: Thinking about it a bit more, I think it's possible to set an invalid MP3 bitrate via a macro, so to cover all possibilities we should ensure that macros also handle any *illegal* bitrate and not just rate == 0.

SteveDaulton: Yes it is possible for a macro to set an invalid sample rate, so something like...
```
   // Verify sample rate
   if (!make_iterator_range( sampRates ).contains( rate ) ||
      (rate < lowrate) || (rate > highrate)) {
         if (project->mBatchMode) {
            // Force valid sample rate in macros.
            if (!make_iterator_range( sampRates ).contains( rate )) {
               auto const bestRateIt = std::lower_bound(sampRates.begin(),
                                                         sampRates.end(), rate);
               rate = (bestRateIt == sampRates.end()) ? highrate : *bestRateIt;
               // wxLogDebug("Best Rate: %d", rate);
            }
            if (rate < lowrate) {
               rate = lowrate;
            }
            else if (rate > highrate) {
               rate = highrate;
            }
         }
         else {
            rate = AskResample(bitrate, rate, lowrate, highrate);
            if (rate == 0) {
               return ProgressResult::Cancelled;
         }
      }
   }

```

Paul-Licameli:
Hello @pgirikishore , can you please do as I requested above:  squash all change into a single commit, rebase onto recent master (always rebase:  do NOT merge master into your work), then force-push?

PR #1667, opened 2021-09-12
Issue 1642 Fix by rdb

SteveDaulton:
Fixes special case where the time warp function maps positive output times
to negative input times (negative input times are now handled by filling
with zeros).
Fixes improper handling of Nyquist's logical stop times and terminate times
which have to get mapped through the time warp function to the output.
Also the cmupv code was using long for sample count instead of int64_t.

Recommended:


SteveDaulton:
Closing this pull request due to Windows CI build error.
Will create a new pull request.

PR #1671, opened 2021-09-13
Delete todo.txt

LWinterberg:
I hope this doesn't get me in trouble with the archeological society, but I think after 9 years without change, we really can get rid of this one :)


SteveDaulton:
I agree that this file can / should be removed as it is so out of date. However, there's a  few points that are still relevant that should be noted elsewhere if not already.  These features are still requested by users, and I believe they would all be useful:

* A master gain control
* A bars/beats time ruler and snap-to bars/beats.
* improve reverb
* Smart Mix (like Quick Mix but guarantees no clipping)
* discontiguous selection
* Icon set (mostly  Linux)

I also not that this point is a close cousin of https://github.com/audacity/audacity/discussions/1686
* [JKC] L&R <-> Average+Difference display. This way you can see 
what a stereo track gets mixed down to as mono, and the stereo 
content separately

LWinterberg:
I've put some of the things into places where I can find them again. "improve reverb" probably will get fixed when we look over it again for the real time stuff.

Issue #1642, opened 2021-09-08
Nyquist: Intermittent failure of PHASEVOCODER


SteveDaulton:
**Describe the bug**
When running the Nyquist PHASEVOVODER function, the returned audio will sometimes be corrupt.
The corruption is evident from either a huge peak in the waveform at the beginning of the processed audio, or the presence of NaNs in the processed audio.

**To Reproduce**

Short Test, but failure is rare:
1. Apply this code via the Nyquist Prompt effect:
```
(let* ((dur (get-duration 1))
       (map (abs-env (pwlv 0 (* 2.0 dur) dur))))
  (phasevocoder *track* map))
```

Long Test, which usually fails:
1. Install this Macro:
[vocoder-test.txt](https://github.com/audacity/audacity/files/7129205/vocoder-test.txt)
2. Generate 1 second of silence
3. Duplicate the track 4 times so that there are 16 tracks total
4. Collapse all tracks so that they can be seen in the Audacity window
5. Select All
6. Apply the Macro

**Expected behavior**

Short Test: 2 seconds of silence.

Long Test: 16 tracks, each containing 1 second of silence.

**Screenshots**
Actual result from Long Test:
![Screenshot_2021-09-08_14-29-05](https://user-images.githubusercontent.com/11681097/132520528-57a2f32b-491f-46f0-8570-5224d3eff6a1.png)

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version: Audacity 3.1.0-alpha-20210908 Commit Id: 1c2a91

**Additional context**

SteveDaulton:
Additional screenshot from another run of "Long Test":
![Fullscreen000](https://user-images.githubusercontent.com/11681097/132535123-34398292-695f-4f61-b7f8-99cde5c1cc3f.png)

Paul-Licameli:
Are you going to debug this yourself, @SteveDaulton ?

Paul-Licameli:
Is this vocoder.ny or some other effect that is not shipped?

SteveDaulton:
> Are you going to debug this yourself, @SteveDaulton ?

It's a tough  one, especially as I don't speak C.
I've discussed it with rdb and he's looking into it, but it has proved elusive so far. The steps described here have enabled rdb to reproduce the issue (he had not been able to reproduce it in standalone Nyquist).

It's this phasevocoder: https://www.cs.cmu.edu/~rbd/doc/nyquist/part8.html#index551

[...]

PR #1672, opened 2021-09-13
Issue 1642 Fix by rdb

SteveDaulton:
Apply upstream update to Nyquist library.
Fixes special case where the time warp function maps positive output times
to negative input times (negative input times are now handled by filling
with zeros).
Fixes improper handling of Nyquist's logical stop times and terminate times
which have to get mapped through the time warp function to the output.

https://github.com/audacity/audacity/issues/1642

Applied upstream update to Nyquist library

Recommended:

### DIFF
--- a/lib-src/libnyquist/nyquist/cmupv/src/cmupv.c
+++ b/lib-src/libnyquist/nyquist/cmupv/src/cmupv.c
@@ -129,8 +129,8 @@ typedef enum {
 
 struct position  // each element in the structure array
 {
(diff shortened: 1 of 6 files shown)


crsib (approved):

Paul-Licameli:
I see an approval, but I think a few unnecessary changes to non-source code files are in here.

SteveDaulton:
> I see an approval, but I think a few unnecessary changes to non-source code files are in here.

This upstream change resolves the problem. What would be the benefit of diverging away from upstream?

Issue #404, opened 2019-12-20
Scrollbars in dark mode


locness3:
**Is your feature request related to a problem? Please describe.**
Scrollbars in dark interface themes aren't dark colored an stay in the system colors.

**Describe the solution you'd like**
I'd like the scrollbars to be themed with the rest of the interface.

**Additional context**
macOS :
![Capture d’écran 2019-12-20 à 15 20 57](https://user-images.githubusercontent.com/37651007/71260931-60f68500-233c-11ea-9efe-0f03a7323e30.png)

JamesCrook:
Thanks for creating an issue.  Yes, we would like this too.  At the moment it looks like we would need to draw the scrollbars ourselves to get the desired result.

locness3:
Yeah, I understand.

Issue #1296, opened 2021-07-14
Many parts of Dark theme Audacity are not really "dark"


petersampsonaudacity:
**Describe the bug**
Many parts of Dark theme Audacity are not really "dark" - in particular
1. Error messages
2. Preferences dialogs
3. Menu bar
4. Device toolbar
5. horizontal and vertical toolbars
6. Selection toolbar
7. Drop-down menus
8. History window
9. Plug-in Manager
10. Log file window
11. dialogs for Effects, Generators and Analyzers
12. The "affordances" drag bars for tracks and clips introduced in 3.0.4
13. there may be more ...

**To Reproduce**
1. Edi> Preferences>Interface
2. In Theme, select "Dark"
3. Observe: many parts are not really dark'

**Expected behavior**
A more consistent darkness

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/125619744-cef4a7e5-927d-4949-908b-411a04ad0607.png)

![image](https://user-images.githubusercontent.com/50205984/125619938-23715020-2c63-47e4-a8ce-7643c004b36c.png)

**Additional information (please complete the following information):**
 - All OS
 - All Audacity versions since Themes were introduced

**Additional context**
Also applies to High Contrast theme.

petersampsonaudacity:
Another example was previously logged on Bugzilla 1717
Bug 1717 - ENH: On opening new project in Dark or High Contrast theme, area to left of horizontal scroll bar is filled with Light theme color
https://bugzilla.audacityteam.org/show_bug.cgi?id=1717

![image](https://user-images.githubusercontent.com/50205984/147676480-7d6877e0-5952-4ad4-acab-bdabebce0aae.png)

This also occurs if you set dark theme, exit Audacity and then relaunch.

_Sometimes during the use of Audacity in Dark theme that area can turn black - but I cannot find repeatable steps to reproduce this transition._

petersampsonaudacity:
Flagging @Tantacrul as a UX/GUI issue

LWinterberg:
this is a consequence of wxWidgets using system-native elements for many things. (with the little area in the latest comment possibly being a bug). There's very little to be done here until we switch to a system that actually allows us to touch these buttons and design them ourselves (eg Qt), hence why I marked this as known issue.

zag2me:
Is there any news on this or a roadmap? 
I've just been trying to roll out 3.13 to our school departments, but our music department mentioned that the interface looks like its from the 1990's :)
[...]

PR #1674, opened 2021-09-14
Opt in for macOS dark mode support

mikaoj:
Resolves: https://github.com/audacity/audacity/issues/404

A bunch of UI components doesn't support dark mode on macOS. Like window frame, scrollbars, dropdowns, settings, etc.
This fixes it by opting in for dark mode support.
[NSRequiresAquaSystemAppearance](https://developer.apple.com/documentation/appkit/nsappearancecustomization/choosing_a_specific_appearance_for_your_macos_app) should be set to false to support dark mode when targeting macOS before 10.14. 

This is how Audacity looks today with dark mode enabled on macOS
audacity_currently

And this is how it looks with this fix
audacity_dark_mode

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/MacOSXBundleInfo.plist.in
+++ b/cmake-proxies/cmake-modules/MacOSXBundleInfo.plist.in
@@ -230,7 +230,7 @@
 	<key>NSHighResolutionCapable</key>
 	<true/>
    <key>NSRequiresAquaSystemAppearance</key>
-   <true/>
+   <false/>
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
The same is true on Windows I know - (and on Linux, I believe).

I logged this bug back in July:
 Many parts of Dark theme Audacity are not really "dark" #1296

LWinterberg:
I cannot believe that the fix for this was just this. Awesome find!

crsib (approved):

Issue #998, opened 2021-06-17
Context menu rework


AnitaBats:
### User Story
As a user, I would like to access some frequent functions directly in context, to save time.

### Description

Currently audio tracks don't have any context menus (right click menus) in Audacity. In most DAWs the most often used actions are also exposed through context menus (in addition to main menus and keyboard shortcuts). This saves users time and helps new users find commonly used actions quickly in context.

We would like to add a handful of often used actions as context menus when right clicking on an audio track, or in the empty space. We can add more actions over time, but this would be a good starter set.

Context menu over audio track:

![1 0 audacity3 menu on audio clip@2x](https://user-images.githubusercontent.com/10109867/125430957-fc5d9057-668b-438b-9dc3-79572facbc2b.jpg)

Content menu outside audio track:

![1 0 audacity3 menu on empty@2x](https://user-images.githubusercontent.com/10109867/125431792-e3063672-5db1-41c4-8365-ad4c7fb10331.jpg)

Discussion at #867 

### Scope, Restrictions and Сourner-cases

- in release 3.1 new context menu is applicable only to Mono and Stereo tracks
- context menu on blank space is different to the context menu on audio tracks (see the screenshots above, please)
- time tracks and label tracks are not in the scope of this task in release 3.1. Time and label tracks are still the same.

jounih:
There is a design here - https://github.com/audacity/audacity/discussions/867

I have a few suggestions there for items to go in the context menus, but we need to decide on a full list together

Penikov:
@jounih Jouni, could you, please, point what functions/actions are required in the new context menu?

AnitaBats:
Ready for development

gera-gas:
**Description of task decomposition.**
All estimation times have two borders: this is not 100% precision time frame, this is the approximate time window for solving task.

1. Add context menu
1.1 Create event handler for right mouse click in `ProjectManager` (this class already support mouseevent `ProjectWindow::OnMouseEvent`, but seems necessary passed up right click event additionally).
1.2 Create context  menu by `wxMenu` and bind (connect) handler for popup context menu callback.
Estimate: 2 - 3 hrs.

2. Call methods from context menu
2.1 Call exist methods from context menu callback, like `OnNewWaveTrack(const CommandContext &context)` for create a mono track or `void OnPaste(const CommandContext &context)`, etc, but it should doing carefully for support right `CommandContext` content.
2.2 Final behavior tests (2,4) hrs.
[...]

PR #1678, opened 2021-09-15
Clip renaming context menu item

Paul-Licameli:
Resolves: #1660

This makes #998 and #1614 more complete

Add an item to the wave clip context menu for renaming.

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
@@ -25,10 +25,12 @@ Paul Licameli split from TrackPanel.cpp
 #include "../../../../../images/Cursors.h"
 #include "../../../../AllThemeResources.h"
 
+#include "../../../../commands/CommandContext.h"
 #include "../../../../HitTestResult.h"
 #include "../../../../ProjectHistory.h"
 #include "../../../../RefreshCode.h"
 #include "../../../../TrackArtist.h"
+#include "../../../../TrackPanel.h"
 #include "../../../../TrackPanelDrawingContext.h"
 #include "../../../../TrackPanelMouseEvent.h"
 #include "../../../../TrackPanelResizeHandle.h"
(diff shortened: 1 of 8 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
```diff
@@ -824,6 +826,24 @@ auto WaveTrackSubView::GetMenuItems(
...
+      if (pTrack) {
+         auto &view = TrackView::Get(*pTrack);
+         auto pAffordance = view.GetAffordanceControls();
+         if ( auto affordanceControl =
+            std::dynamic_pointer_cast<WaveTrackAffordanceControls>(
+               pAffordance ) ) {
+            // Begin editing of the label, either with a dialog or directly
+            // in the track panel.
+            auto &project = context.project;
+            affordanceControl->StartEditSelectedClipName( project,
```
vsverchinsky, line 839: Suggest adding `WaveTrackAffordanceControls::StartEditClipNameAt(AudacityProject&, double time)`  instead, or use events

Paul-Licameli: I don't understand the suggestion to use events.

The new test argument makes the function very general, but maybe you think it is too general.  If you don't, you might object that the generalized function should at least be renamed, and I would agree.

vsverchinsky: It's not about generality. Yes, function name now became at bit controversial.

vsverchinsky: Also, right channel does not always have affordance cell, may be in that case call should be redirected to the view of the leader track?

Paul-Licameli: > Also, right channel does not always have affordance cell, may be in that case call should be redirected to the view of the leader track?

That is a good point.

Paul-Licameli: If events are used, which object would emit the event and which would react to it?

Paul-Licameli:
I have changed the function name and also check other channels.

I still do not understand why event notifications should be used here.  You used them so that changed of track selection state could be detected by TextEditHelper, and that was proper for avoidance of dependency cycles (TrackList should not depend on the code that reacts to the event).  I don't see a problem with cycles in this commit that needs this technique to break them.

Paul-Licameli:
I also added the neglected change of undo history.

Paul-Licameli:
The problem of pushing the undo history stack isn't solved, so I mark this as a draft until I do that.

Paul-Licameli:
I don't recall why I thought undo history was sometimes not pushed -- maybe I was confused.

But it can be unnecessarily pushed -- if you just hit Enter with no changes to the text.  I changed that case.  It should affect behavior of clicking on the name as well as the new context menu item.

Esc key works as you would expect, discarding any changes you may have made without hitting Enter.

vsverchinsky (approved):

Issue #1401, opened 2021-08-02
"Smart" clips: UI


vsverchinsky:
Implement UI changes related to "smart" clips.
Scope of work:
* Add new (stub) UIHandle for clip resizing ~1hrs
* Resizing with new UIHandle: ~4hrs
* Update selection behaviour within clip: ~2hrs

Details https://github.com/audacity/audacity/issues/994

Issue #1400, opened 2021-08-02
"Smart" clips: internals


vsverchinsky:
Implement internals related to smart clips.
Scope of work:
* Introduce WaveClip state for trimming; update related code: clip movement within track, overlaping with other tracks~12hrs
* Handle clip copying: ~8hrs

Details https://github.com/audacity/audacity/issues/994

PR #1679, opened 2021-09-15
Smart clips

vsverchinsky:
Resolves: #1400
Resolves: #1401

Work in progress

Recommended:

### DIFF
--- a/src/WaveClip.cpp
+++ b/src/WaveClip.cpp
@@ -140,7 +140,9 @@ WaveClip::WaveClip(const WaveClip& orig,
    // current sample block factory, because we might be copying
    // from one project to another
 
-   mOffset = orig.mOffset;
+   mSequenceOffset = orig.mSequenceOffset;
+   mTrimLeft = orig.mTrimLeft;
+   mTrimRight = orig.mTrimRight;
    mRate = orig.mRate;
    mColourIndex = orig.mColourIndex;
    mSequence = std::make_unique<Sequence>(*orig.mSequence, factory);
@@ -168,7 +170,8 @@ WaveClip::WaveClip(const WaveClip& orig,
 {
    // Copy only a range of the other WaveClip
 
-   mOffset = orig.mOffset;
+   mSequenceOffset = orig.mSequenceOffset;
+   
    mRate = orig.mRate;
    mColourIndex = orig.mColourIndex;
 
@@ -178,17 +181,15 @@ WaveClip::WaveClip(const WaveClip& orig,
 
    mIsPlaceholder = orig.GetIsPlaceholder();
 
-   sampleCount s0, s1;
-
-   orig.TimeToSamplesClip(t0, &s0);
-   orig.TimeToSamplesClip(t1, &s1);
+   auto s0 = orig.TimeToSequenceSamples(t0);
+   auto s1 = orig.TimeToSequenceSamples(t1);
 
    mSequence = orig.mSequence->Copy(factory, s0, s1);
 
    mEnvelope = std::make_unique<Envelope>(
       *orig.mEnvelope,
-      mOffset + s0.as_double()/mRate,
-      mOffset + s1.as_double()/mRate
+      GetSequenceStartTime() + s0.as_double()/mRate,
+      GetSequenceStartTime() + s1.as_double()/mRate
    );
 
    if ( copyCutlines )
@@ -196,12 +197,12 @@ WaveClip::WaveClip(const WaveClip& orig,
       for (const auto &ppClip : orig.mCutLines)
       {
          const WaveClip* clip = ppClip.get();
-         double cutlinePosition = orig.mOffset + clip->GetOffset();
+         double cutlinePosition = orig.GetSequenceStartTime() + clip->GetSequenceStartTime();
          if (cutlinePosition >= t0 && cutlinePosition <= t1)
          {
             auto newCutLine =
                std::make_unique< WaveClip >( *clip, factory, true );
-            newCutLine->SetOffset( cutlinePosition - t0 );
+            newCutLine->SetSequenceStartTime( cutlinePosition - t0 );
             mCutLines.push_back(std::move(newCutLine));
          }
       }
@@ -212,17 +213,10 @@ WaveClip::~WaveClip()
 {
 }
 
-/*! @excsafety{No-fail} */
-void WaveClip::SetOffset(double offset)
-{
-    mOffset = offset;
-    mEnvelope->SetOffset(mOffset);
-}
-
 bool WaveClip::GetSamples(samplePtr buffer, sampleFormat format,
                    sampleCount start, size_t len, bool mayThrow) const
 {
-   return mSequence->Get(buffer, format, start, len, mayThrow);
+   return mSequence->Get(buffer, format, start + TimeToSamples(mTrimLeft), len, mayThrow);
 }
 
 /*! @excsafety{Strong} */
@@ -230,7 +224,7 @@ void WaveClip::SetSamples(constSamplePtr buffer, sampleFormat format,
                    sampleCount start, size_t len)
 {
    // use Strong-guarantee
-   mSequence->SetSamples(buffer, format, start, len);
+   mSequence->SetSamples(buffer, format, start + TimeToSamples(mTrimLeft), len);
 
    // use No-fail-guarantee
    MarkChanged();
@@ -245,71 +239,6 @@ const BlockArray* WaveClip::GetSequenceBlockArray() const
 {
    return &mSequence->GetBlockArray();
 }
-
-double WaveClip::GetStartTime() const
-{
-   // JS: mOffset is the minimum value and it is returned; no clipping to 0
-   return mOffset;
-}
-
-double WaveClip::GetEndTime() const
-{
-   auto numSamples = mSequence->GetNumSamples();
-
-   double maxLen = mOffset + (numSamples+mAppendBufferLen).as_double()/mRate;
-   // JS: calculated value is not the length;
-   // it is a maximum value and can be negative; no clipping to 0
-
-   return maxLen;
-}
-
-sampleCount WaveClip::GetStartSample() const
-{
-   return sampleCount( floor(mOffset * mRate + 0.5) );
-}
-
-sampleCount WaveClip::GetEndSample() const
-{
-   return GetStartSample() + mSequence->GetNumSamples();
-}
-
-sampleCount WaveClip::GetNumSamples() const
-{
-   return mSequence->GetNumSamples();
-}
-
-// Bug 2288 allowed overlapping clips.
-// This was a classic fencepost error.
-// We are within the clip if start < t <= end.
-// Note that BeforeClip and AfterClip must be consistent 
-// with this definition.
-bool WaveClip::WithinClip(double t) const
-{
-   auto ts = (sampleCount)floor(t * mRate + 0.5);
-   return ts > GetStartSample() && ts < GetEndSample() + mAppendBufferLen;
-}
-
-bool WaveClip::BeforeClip(double t) const
-{
-   auto ts = (sampleCount)floor(t * mRate + 0.5);
-   return ts <= GetStartSample();
-}
-
-bool WaveClip::AfterClip(double t) const
-{
-   auto ts = (sampleCount)floor(t * mRate + 0.5);
-   return ts >= GetEndSample() + mAppendBufferLen;
-}
-
-// A sample at time t could be in the clip, but 
-// a clip start at time t still be from a clip 
-// not overlapping this one, with this test.
-bool WaveClip::IsClipStartAfterClip(double t) const
-{
-   auto ts = (sampleCount)floor(t * mRate + 0.5);
-   return ts >= GetEndSample() + mAppendBufferLen;
-}
-
 
 ///Delete the wave cache - force redraw.  Thread-safe
 void WaveClip::ClearWaveCache()
@@ -384,6 +313,8 @@ fillWhere(std::vector<sampleCount> &where, size_t len, double bias, double corre
 bool WaveClip::GetWaveDisplay(WaveDisplay &display, double t0,
                                double pixelsPerSecond) const
 {
+   t0 += GetTrimLeft();
+
    const bool allocated = (display.where != 0);
 
    const size_t numPixels = (int)display.width;
@@ -1013,11 +944,17 @@ bool WaveClip::GetSpectrogram(WaveTrackCache &waveTrackCache,
                               size_t numPixels,
                               double t0, double pixelsPerSecond) const
 {
+   t0 += GetTrimLeft();
+
    const WaveTrack *const track = waveTrackCache.GetTrack().get();
    const SpectrogramSettings &settings = track->GetSpectrogramSettings();
 
+   //Trim offset comparison failure forces spectrogram cache rebuild 
+   //and skip copying "unchanged" data after clip border was trimmed.
    bool match =
       mSpecCache &&
+      mSpecCache->leftTrim == GetTrimLeft() &&
+      mSpecCache->rightTrim == GetTrimRight() &&
       mSpecCache->len > 0 &&
       mSpecCache->Matches
       (mDirty, pixelsPerSecond, settings, mRate);
@@ -1070,6 +1007,8 @@ bool WaveClip::GetSpectrogram(WaveTrackCache &waveTrackCache,
 
    // Resize the cache, keep the contents unchanged.
    mSpecCache->Grow(numPixels, settings, pixelsPerSecond, t0);
+   mSpecCache->leftTrim = GetTrimLeft();
+   mSpecCache->rightTrim = GetTrimRight();
    auto nBins = settings.NBins();
 
    // Optimization: if the old cache is good and overlaps
@@ -1108,8 +1047,8 @@ bool WaveClip::GetSpectrogram(WaveTrackCache &waveTrackCache,
 
    mSpecCache->Populate
       (settings, waveTrackCache, copyBegin, copyEnd, numPixels,
-       mSequence->GetNumSamples(),
-       mOffset, mRate, pixelsPerSecond);
+       GetSequenceSamplesCount(),
+       GetSequenceStartTime(), mRate, pixelsPerSecond);
 
    mSpecCache->dirty = mDirty;
    spectrogram = &mSpecCache->freq[0];
@@ -1133,10 +1072,8 @@ std::pair<float, float> WaveClip::GetMinMax(
    if (t0 == t1)
       return{ 0.f, 0.f };
 
-   sampleCount s0, s1;
-
-   TimeToSamplesClip(t0, &s0);
-   TimeToSamplesClip(t1, &s1);
+   auto s0 = TimeToSequenceSamples(t0);
+   auto s1 = TimeToSequenceSamples(t1);
 
    return mSequence->GetMinMax(s0, s1-s0, mayThrow);
 }
@@ -1152,10 +1089,8 @@ float WaveClip::GetRMS(double t0, double t1, bool mayThrow) const
    if (t0 == t1)
       return 0.f;
 
-   sampleCount s0, s1;
-
-   TimeToSamplesClip(t0, &s0);
-   TimeToSamplesClip(t1, &s1);
+   auto s0 = TimeToSequenceSamples(t0);
+   auto s1 = TimeToSequenceSamples(t1);
 
    return mSequence->GetRMS(s0, s1-s0, mayThrow);
 }
@@ -1177,16 +1112,6 @@ void WaveClip::UpdateEnvelopeTrackLen()
    auto len = (mSequence->GetNumSamples().as_double()) / mRate;
    if (len != mEnvelope->GetTrackLen())
       mEnvelope->SetTrackLen(len, 1.0 / GetRate());
-}
-
-void WaveClip::TimeToSamplesClip(double t0, sampleCount *s0) const
-{
-   if (t0 < mOffset)
-      *s0 = 0;
-   else if (t0 > mOffset + mSequence->GetNumSamples().as_double()/mRate)
-      *s0 = mSequence->GetNumSamples();
-   else
-      *s0 = sampleCount( floor(((t0 - mOffset) * mRate) + 0.5) );
 }
 
 /*! @excsafety{Strong} */
@@ -1311,7 +1236,21 @@ bool WaveClip::HandleXMLTag(const wxChar *tag, const wxChar **attrs)
             if (!XMLValueChecker::IsGoodString(strValue) ||
                   !Internat::CompatibleToDouble(strValue, &dblValue))
                return false;
-            SetOffset(dblValue);
+            SetSequenceStartTime(dblValue);
+         }
+         else if (!wxStrcmp(attr, wxT("trimLeft")))
+         {
+            if (!XMLValueChecker::IsGoodString(strValue) ||
+               !Internat::CompatibleToDouble(strValue, &dblValue))
+               return false;
+            SetTrimLeft(dblValue);
+         }
+         else if (!wxStrcmp(attr, wxT("trimRight")))
+         {
+            if (!XMLValueChecker::IsGoodString(strValue) ||
+               !Internat::CompatibleToDouble(strValue, &dblValue))
+               return false;
+            SetTrimRight(dblValue);
          }
          else if (!wxStrcmp(attr, wxT("name")))
          {
@@ -1360,7 +1299,9 @@ void WaveClip::WriteXML(XMLWriter &xmlFile) const
 // may throw
 {
    xmlFile.StartTag(wxT("waveclip"));
-   xmlFile.WriteAttr(wxT("offset"), mOffset, 8);
+   xmlFile.WriteAttr(wxT("offset"), mSequenceOffset, 8);
+   xmlFile.WriteAttr(wxT("trimLeft"), mTrimLeft, 8);
+   xmlFile.WriteAttr(wxT("trimRight"), mTrimRight, 8);
    xmlFile.WriteAttr(wxT("name"), mName);
    xmlFile.WriteAttr(wxT("colorindex"), mColourIndex );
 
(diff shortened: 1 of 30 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -71,13 +77,49 @@ HitTestPreview WaveClipTrimHandle::Preview(const TrackPanelMouseState& mouseStat
...
+(const TrackPanelMouseEvent& event, AudacityProject* project)
 {
-    return RefreshCode::RefreshNone;
+    const auto newX = event.event.GetX();
+    const auto dx = newX - mDragStartX;
+
+    auto& viewInfo = ViewInfo::Get(*project);
+
+    
+    auto t = viewInfo.PositionToTime(viewInfo.TimeToPosition(mInitialEdgeTime, event.rect.x) + dx, event.rect.x);
```
Paul-Licameli, line 98: This is correct use of PositionToTime and TimeToPosition.   You understatnd that you don't just mutliply dx by some scaling factor.  The assumption of uniform samples per pixel should not be made.  Future projects might change that.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -89,5 +131,15 @@ UIHandle::Result WaveClipTrimHandle::Release
...
+    {
+    case Border::Left:
+        mClip->TrimLeftTo(mInitialEdgeTime);
+        break;
+    case Border::Right:
+        mClip->TrimRightTo(mInitialEdgeTime);
+        break;
+    default: break;
+    };
+    return RefreshCode::RefreshCell;
```
Paul-Licameli, line 144: The Cancel procedure is good.

The Release is still empty.  It will require pushing an undo history item.

src/tracks/playabletrack/wavetrack/ui/WaveTrackView.h
```diff
@@ -61,6 +61,7 @@ class AUDACITY_DLL_API WaveTrackSubView : public CommonTrackView
    std::weak_ptr<UIHandle> mResizeHandle;
    std::weak_ptr<UIHandle> mAdjustHandle;
    std::weak_ptr<UIHandle> mRearrangeHandle;
+   std::weak_ptr<UIHandle> mClipTrimHandle;
```
Paul-Licameli: std::weak_ptr<UIHandle> instead of some subclass raises the danger of the slicing problem, which you correct in a later commit.

But look at all those old bad examples before it that I wrote!  These should be corrected too.

src/WaveClip.cpp
```diff
@@ -140,7 +140,9 @@ WaveClip::WaveClip(const WaveClip& orig,
    // current sample block factory, because we might be copying
    // from one project to another
 
-   mOffset = orig.mOffset;
+   mSequenceOffset = orig.mSequenceOffset;
```
Paul-Licameli, line 143: I observe that this second commit in the draft PR does not build.

In the final PR, that should be corrected, and I would like this internals commit to be first in the sequence.  Or it might be a PR by itself.

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -261,7 +261,7 @@ void FormatMenuTable::OnFormatChange(wxCommandEvent & event)
 
    sampleCount totalSamples{ 0 };
    for (const auto& channel : TrackList::Channels(pTrack))
-      totalSamples += channel->GetNumSamples();
+      totalSamples += channel->GetSequenceSamplesCount();
```
Paul-Licameli, line 264: The unusual case where a call to one of the *Sequence* functions is needed not *Play*.  Comment it.

src/tracks/playabletrack/wavetrack/ui/WaveTrackView.cpp
```diff
@@ -40,9 +40,7 @@ Paul Licameli split from TrackPanel.cpp
 #include "../../../../TrackInfo.h"
 
 #include "WaveTrackAffordanceControls.h"
-
-namespace {
```
Paul-Licameli, line 44: I would slightly prefer not removing the anonymous namespace over can, making symbol tables just a little smaller.  This would require forward-declaring incomplete types in namespace { ... } in the .h file.

src/effects/nyquist/Nyquist.cpp
```diff
@@ -1239,8 +1239,8 @@ bool NyquistEffect::ProcessOne()
          for (size_t i=0; i<ca.size(); i++) {
             if (i < 1000) {
                clips += wxString::Format(wxT("(list (float %s) (float %s))"),
-                                         Internat::ToString(ca[i]->GetStartTime()),
-                                         Internat::ToString(ca[i]->GetEndTime()));
+                                         Internat::ToString(ca[i]->GetPlayStartTime()),
```
Paul-Licameli, line 1242: This change is correct, but discuss with Steve Daulton whether new information about the size of the hidden parts of the clip should also be passed to Nyquist.

Understand what happens here: C++ composes a string that is passed to the Lisp interpreter, and that sets up variables that are available to programming in Lisp.

src/effects/Reverse.cpp
```diff
@@ -182,12 +182,12 @@ bool EffectReverse::ProcessOneWave(int count, WaveTrack * track, sampleCount sta
...
             if(i+1 < clipArray.size()) // update currentEnd if there is a clip to process next
             {
-               auto nextClipStart = clipArray[i+1]->GetStartSample();
+               auto nextClipStart = clipArray[i+1]->GetPlayStartSample();
                currentEnd = currentEnd - (clipEnd - clipStart) - (nextClipStart - clipEnd);
             }
 
             revClips.push_back(track->RemoveAndReturnClip(clip)); // detach the clip from track
-            revClips.back()->SetOffset(track->LongSamplesToTime(track->TimeToLongSamples(offsetStartTime))); // align time to a sample and set offset
+            revClips.back()->SetPlayStartTime(track->LongSamplesToTime(track->TimeToLongSamples(offsetStartTime))); // align time to a sample and set offset
```
Paul-Licameli, line 190: As I review the necessary changes in effects, this occurs to me.  What is expected if effects are applied to clips with hidden parts?  If you do nothing else, then the hidden parts remain unchanged.  Perhaps that is expected behavior.  Or perhaps in the case of the Reverse effect, it would be surprising behavior.  I’m not sure whether you should do anything extra.

This is one of the few places where there was a call to SetOffset.  Other effects did not use it.

src/commands/GetInfoCommand.cpp
```diff
@@ -518,8 +518,8 @@ bool GetInfoCommand::SendClips(const CommandContext &context)
          for (WaveClip * pClip : ptrs) {
             context.StartStruct();
             context.AddItem((double)i, "track");
-            context.AddItem(pClip->GetStartTime(), "start");
-            context.AddItem(pClip->GetEndTime(), "end");
+            context.AddItem(pClip->GetPlayStartTime(), "start");
+            context.AddItem(pClip->GetPlayEndTime(), "end");
```
Paul-Licameli, line 522: As with Nyquist, so perhaps too for macro programming: give more information about the size of the hidden parts of the wave for sophisticated users.  Only information:  no interface for changing it.

src/commands/SetClipCommand.cpp
```diff
@@ -95,7 +95,7 @@ bool SetClipCommand::ApplyInner( const CommandContext &, Track * t )
                pClip->SetColourIndex(mColour);
             // No validation of overlap yet.  We assume the user is sensible!
```
Paul-Licameli, line 96: Notice this comment.  The non-overlapping condition for clips can be violated with macro programming!  Not our responsibility to fix that yet I think.

src/commands/SetClipCommand.cpp
```diff
@@ -95,7 +95,7 @@ bool SetClipCommand::ApplyInner( const CommandContext &, Track * t )
...
             if( bHasT0 )
-               pClip->SetOffset(mT0);
+               pClip->SetPlayStartTime(mT0);
```
Paul-Licameli, line 98: Another unusual SetOffset call.

src/WaveTrack.h
```diff
@@ -609,6 +610,8 @@ class AUDACITY_DLL_API WaveTrack final : public PlayableTrack {
    // Protected methods
    //
 
+   void PasteWaveTrack(double t0, const WaveTrack* other);
```
Paul-Licameli, line 613: I notice that this is a protected method.  I notice also that WaveTrack is a final class, so the distinction of protected and private is not useful.  Maybe the class should be simplified, only to have public and private methods.

src/WaveTrack.h
```diff
@@ -148,7 +148,8 @@ class AUDACITY_DLL_API WaveTrack final : public PlayableTrack {
    int GetWaveColorIndex() const { return mWaveColorIndex; };
    void SetWaveColorIndex(int colorIndex);
 
-   sampleCount GetNumSamples() const;
+   sampleCount GetPlaySamplesCount() const;
+   sampleCount GetSequenceSamplesCount() const;
```
Paul-Licameli, line 152: I see that WaveTrack::GetNumSamples becomes two public methods, each used in only one other place, and those other uses are correct.

I like the renaming changes here and in WaveClip -- *Sample* versus *Play* names -- and they force you to recompile many places and consider carefully which renaming is the correct one.  I am more confident in your work when I see that you did this.  Should the name changes be permanent?  I say, yes.

Much less has changed in the interface of WaveTrack.h, than for WaveClip.h.  It is a little surprising -- so to review carefully I must also look at what else is in WaveTrack.h that did not change, and be convinced that it is correct not to change them.

I notice that WaveTrack::GetStartTime() and GetEndTime() did not change names.  But as you know, that is because it is an override of a method of Track.h, which declares virtual GetStartTime() and GetEndTime() but doesn't comment what they are supposed to mean.  So, I understand this exception to renaming -- you don't want to make the name changes in the other overrides.

Still that means I must review all uses of GetStartTime and GetEndTime to be sure.

src/menus/ClipMenus.cpp
```diff
@@ -110,10 +110,10 @@ double AdjustForFindingStartTimes(
 {
    auto q = std::find_if(clips.begin(), clips.end(),
       [&] (const WaveClip* const& clip) {
-         return clip->GetEndTime() == time; });
```
Paul-Licameli, line 113: Comments above this function, and above the next function, still mention the old method names `GetStartTime, GetEndTime`.  Update the comments.

src/WaveTrack.h
Paul-Licameli: I'm doing that review, and I notice many GetStartTime and GetEndTime calls in effects, which I just assume correct -- that destructively applied affects will only apply to the un-hidden parts of tracks.  As I said, that might be surprising in some cases and a debatable point of design, but I will interpret the design for now as not demanding it.

I notice GetOffset() and SetOffset() are two other virtual functions in the base class Track that were not renamed, so I must also review uses of those to be sure of completeness.

src/WaveClip.cpp
```diff
@@ -1772,8 +1700,8 @@ void WaveClip::Resample(int rate, ProgressDialog *progress)
 // be exactly equal due to rounding errors.
 bool WaveClip::SharesBoundaryWithNextClip(const WaveClip* next) const
```
Paul-Licameli, line 1701: The comment preceding this function also needs update to mention new method names

src/WaveTrack.h
Paul-Licameli: Now I have done that review.  I think that the legacy in class Track needs a clean-up.  It's not clear what the distinction of "start time" versus "offset" is supposed to mean, or what purpose the mOffset member variable serves.  I think there should be a clean-up project to eliminate the "Offset" functions.

This is not to be part of the present pull request.  It is for one of us to write in the stabilization period and for the other one to review.

src/WaveTrack.cpp
```diff
@@ -420,6 +422,159 @@ const WaveClip* WaveTrack::FindClipByName(const wxString& name) const
    return nullptr;
 }
 
+void WaveTrack::PasteWaveTrack(double t0, const WaveTrack* other)
```
Paul-Licameli, line 425: Please insert this new member function immediately before WaveTrack::Paste().  That may make the commit more easily reviewable because there will be more common lines, and less in the diff view.

src/WaveTrack.cpp
```diff
@@ -73,8 +73,10 @@ bool AreAligned(const WaveClipPointers& a, const WaveClipPointers& b)
       return false;
 
    const auto compare = [](const WaveClip* a, const WaveClip* b) {
-      return a->GetStartTime() == b->GetStartTime() &&
-         a->GetNumSamples() == b->GetNumSamples();
+      return a->GetPlayStartTime() == b->GetPlayStartTime() &&
+         a->GetSequenceStartTime() == b->GetSequenceStartTime() &&
+         a->GetPlayEndTime() == b->GetPlayEndTime() &&
+         a->GetSequenceEndTime() == b->GetSequenceEndTime();
```
Paul-Licameli, line 79: Buried in all these changes is one very interesting thing here:  the need for a more strict definition of correspondence of clips.

A detail like this deserves comments, both in the code here, and also in the git commit comments.

I'd like to encourage you to write more comments, of both kinds, about such changes of behavior.

src/WaveTrack.cpp
```diff
@@ -625,28 +790,18 @@ void WaveTrack::Trim (double t0, double t1)
...
    for (const auto &clip : mClips)
    {
-      //Find the first clip greater than the offset.
-      //If we end up clipping the entire track, this is useful.
-      if(firstGreaterOffset < 0 &&
-            clip->GetStartTime() >= t0)
-         firstGreaterOffset = clip->GetStartTime();
-
-      if(t1 > clip->GetStartTime() && t1 < clip->GetEndTime())
+      if(t1 > clip->GetPlayStartTime() && t1 < clip->GetPlayEndTime())
```
Paul-Licameli, line 796: The function `WaveTrack::Trim` has another interesting non-trivial change of behavior, like the test for correspondence, which is worthy of a comment.  Instead of deleting material, as before, now we hide the material in the smart clip but do not forget it.

I looked for where this function is used.  One place is in EffectSoundTouch -- but it always applies to a newly generated temporary track.  The other implements the Trim editing command, in EditMenus.cpp.  There is a comment there, no longer appropriate, which should be changed:

```
//Delete the section before the left selector
```

src/WaveTrack.cpp
```diff
@@ -696,27 +851,27 @@ Track::Holder WaveTrack::Copy(double t0, double t1, bool forClipboard) const
...
+                 .Format(clip->GetName(), clip_t0 - clip->GetPlayStartTime(), clip_t1 - clip->GetPlayStartTime())
                  .Translation());
 
          //wxPrintf("copy: clip_t0=%f, clip_t1=%f\n", clip_t0, clip_t1);
 
          newClip->Offset(-t0);
-         if (newClip->GetOffset() < 0)
-            newClip->SetOffset(0);
+         if (newClip->GetPlayStartTime() < 0)
+            newClip->SetPlayStartTime(0);
```
Paul-Licameli, line 874: A definite mistake here.  WaveClip::SetOffset used to affect the envelope and the sequence.  But WaveClip::SetPlayStartTime now affects only the sequence.

src/WaveClip.cpp
```diff
@@ -1789,3 +1717,233 @@ const wxString& WaveClip::GetName() const
...
+}
+
+double WaveClip::GetPlayStartTime() const noexcept
+{
+    return mSequenceOffset + mTrimLeft;
+}
+
+void WaveClip::SetPlayStartTime(double time)
+{
+    mSequenceOffset = time - mTrimLeft;
```
Paul-Licameli, line 1806: If this function instead called through to WaveClip::SetSequenceStartTime(), which does change the envelope offset, then the bug last mentioned would be fixed.

Paul-Licameli (changes requested):
I found a bug in updating of envelopes.

src/WaveTrack.cpp
```diff
@@ -983,18 +1138,19 @@ void WaveTrack::ClearAndPaste(double t0, // Start of time to clear
             double st;
             double et;
 
-            st = clip->GetStartTime();
-            et = clip->GetEndTime();
+            st = clip->GetPlayStartTime();
+            et = clip->GetPlayEndTime();
```
Paul-Licameli, line 1142: I think cut line operations should only refer to sequence start and end times.  The mOffset of a cut line was always interpreted as relative to the mOffset of the containing clip.

I think there is not a real bug because the trim left value of a cut line never becomes nonzero, but then it is simpler not to rely on that.

src/WaveTrack.cpp
```diff
@@ -983,18 +1138,19 @@ void WaveTrack::ClearAndPaste(double t0, // Start of time to clear
...
 
             // Scan the cuts for any that live within this clip
             for (auto it = cuts.begin(); it != cuts.end();) {
                WaveClip *cut = it->get();
-               double cs = cut->GetOffset();
+               //cutlines in this array were orphaned previousely
```
Paul-Licameli, line 1147: “Previously”

src/WaveTrack.cpp
```diff
@@ -983,18 +1138,19 @@ void WaveTrack::ClearAndPaste(double t0, // Start of time to clear
...
+               double cs = cut->GetSequenceStartTime();
 
                // Offset the cut from the start of the clip and add it to
                // this clips cutlines.
                if (cs >= st && cs <= et) {
-                  cut->SetOffset(warper->Warp(cs) - st);
+                  cut->SetSequenceStartTime(warper->Warp(cs) - st);
```
Paul-Licameli, line 1153: I thought there might be a bug with cut lines, but I read all again carefully and no longer think so.  Good job

src/WaveTrack.cpp
```diff
@@ -2442,12 +2432,20 @@ void WaveTrack::UpdateLocationsCache() const
...
+         if (prev && fabs(prev->GetPlayEndTime() -
+                          clip->GetPlayStartTime()) < WAVETRACK_MERGE_POINT_TOLERANCE)
             ++num;
 
+         /*auto& cutlines = clip->GetCutLines();
+         for (auto& cut : cutlines)
+         {
+             auto cutlinePosition = clip->GetSequenceStartTime() + cut->GetSequenceStartTime();
+             if (clip->WithinPlayRegion(cutlinePosition))
+                 ++num;
```
Paul-Licameli, line 2447: Some doubtful code you dont need now?

src/WaveClip.cpp
vsverchinsky: Thanks!

src/WaveTrack.cpp
vsverchinsky: Yes, but now I think it's not that important to precisely count only visible cutlines here, since it's used only as an estimation for cache preallocation size and assertion at the end.  Removing it

src/WaveTrack.cpp
vsverchinsky: Thanks!

src/WaveTrack.cpp
vsverchinsky: Updated both commit message and function comment

Paul-Licameli:
I see this is rebased onto more recent master now.

I like the more detailed explanation in commit comments.

In commit comments, if you write more than one line, then always make the second line blank, and after it, use line breaks at about 80 columns.

src/WaveClip.cpp
```diff
@@ -1789,3 +1712,233 @@ const wxString& WaveClip::GetName() const
...
+}
+
+sampleCount WaveClip::GetSequenceStartSample() const
+{
+    return TimeToSamples(mSequenceOffset);
+}
+
+sampleCount WaveClip::GetSequenceEndSample() const
+{
+    return GetSequenceStartSample() + mSequence->GetNumSamples();
```
Paul-Licameli, line 1901: Also not accounting for buffer length, inconsistently with `GetSequenceEndTime()`

But that was also so in removed `GetEndSample`

src/WaveClip.cpp
```diff
@@ -1789,3 +1712,233 @@ const wxString& WaveClip::GetName() const
...
+
+// Bug 2288 allowed overlapping clips.
+// This was a classic fencepost error.
+// We are within the clip if start < t <= end.
+// Note that BeforeClip and AfterClip must be consistent 
+// with this definition.
+bool WaveClip::WithinPlayRegion(double t) const
+{
+    auto ts = (sampleCount)floor(t * mRate + 0.5);
+    return ts > GetPlayStartSample() && ts < GetPlayEndSample() + mAppendBufferLen;
```
Paul-Licameli, line 1917: This copies the comment and the logic of `WaveClip::WithinClip()` but it looks strange.  The fix for 2288 changed `<=` to `<`.  But should it have changed `>` to `>=` instead?

src/WaveClip.cpp
```diff
@@ -1789,3 +1712,233 @@ const wxString& WaveClip::GetName() const
...
+}
+
+bool WaveClip::BeforePlayStartTime(double t) const
+{
+    auto ts = (sampleCount)floor(t * mRate + 0.5);
+    return ts <= GetPlayStartSample();
```
Paul-Licameli, line 1923: And this copies `BeforeClip()` and the comment said there must be consistency.  But this looks like a very old off by one error.

src/WaveClip.cpp
```diff
@@ -1789,3 +1712,233 @@ const wxString& WaveClip::GetName() const
...
+}
+
+bool WaveClip::AfterPlayEndTime(double t) const
+{
+    auto ts = TimeToSamples(t);
+    return ts >= GetPlayEndSample() + mAppendBufferLen;
```
Paul-Licameli, line 1929: And the fix for 2288 changed '>=' to '>' but here you do correct the coresponding function.

Maybe you reintroduce the bug 2288, which had its questionable fix.

Paul-Licameli (commented):
Some observations about inconsistent use of append buffer length and possible off by one errors and rounding.  But I'm not recommending changes.

There may be a lot of subtle problems like this already existing and we might leave it until stabilization to study and try to fix them.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -0,0 +1,226 @@
...
+            std::vector<std::shared_ptr<WaveClip>>{ clip };
+        return AssignUIHandlePtr(holder,
+            std::make_shared<WaveClipTrimHandle>(trimRange, clips, targetBorder)
+        );
+    };
+
+    auto& zoomInfo = ViewInfo::Get(*pProject);
+    //Test left and right boundaries of each clip
+    //and determine the maximum offsets allowed for trimming,
+    //which are constrainted either by other clips, own length,
```
Paul-Licameli, line 85: "constrained"

src/tracks/playabletrack/wavetrack/ui/WaveTrackView.h
```diff
@@ -61,10 +71,11 @@ class AUDACITY_DLL_API WaveTrackSubView : public CommonTrackView
...
 
 private:
-   std::weak_ptr<UIHandle> mCloseHandle;
-   std::weak_ptr<UIHandle> mResizeHandle;
-   std::weak_ptr<UIHandle> mAdjustHandle;
-   std::weak_ptr<UIHandle> mRearrangeHandle;
+   std::weak_ptr<SubViewCloseHandle> mCloseHandle;
+   std::weak_ptr<TrackPanelResizeHandle> mResizeHandle;
+   std::weak_ptr<SubViewAdjustHandle> mAdjustHandle;
+   std::weak_ptr<SubViewRearrangeHandle> mRearrangeHandle;
```
Paul-Licameli, line 77: Thank you for fixing my embarrassing slicing mistakes!

src/WaveTrack.cpp
Paul-Licameli: Thanks for that

src/WaveClip.h
```diff
@@ -203,7 +205,9 @@ class AUDACITY_DLL_API WaveClip final : public XMLTagHandler
 
    // Always gives non-negative answer, not more than sample sequence length
    // even if t0 really falls outside that range
-   void TimeToSamplesClip(double t0, sampleCount *s0) const;
+   sampleCount TimeToSequenceSamples(double t) const;
+   sampleCount ToSequenceSamples(sampleCount s) const;
+   //void TimeToSamplesClip(double t0, sampleCount *s0) const;
```
Paul-Licameli, line 210: Remove unused function in comment

src/WaveClip.h
```diff
@@ -216,24 +220,46 @@ class AUDACITY_DLL_API WaveClip final : public XMLTagHandler
...
+   void SetTrimRight(double trim);
+   double GetTrimRight() const noexcept;
+
+   //! Moves play start position by deltaTime
+   void TrimLeft(double deltaTime);
+   //! Moves play end position by deltaTime
+   void TrimRight(double deltaTime);
+
+   void TrimLeftTo(double to);
+   void TrimRightTo(double to);
```
Paul-Licameli, line 252: Comment that these set the trimming to a given absolute time (if that is in bounds)

src/WaveClip.h
```diff
@@ -216,24 +220,46 @@ class AUDACITY_DLL_API WaveClip final : public XMLTagHandler
...
+
+   sampleCount GetPlayStartSample() const;
+   sampleCount GetPlayEndSample() const;
+   sampleCount GetPlaySamplesCount() const;
+
+   void SetTrimLeft(double trim);
+   double GetTrimLeft() const noexcept;
+
+   void SetTrimRight(double trim);
+   double GetTrimRight() const noexcept;
```
Paul-Licameli, line 244: Comment that these four get and set the non-negative length of hidden material at each end

src/WaveClip.h
```diff
@@ -357,12 +383,26 @@ class AUDACITY_DLL_API WaveClip final : public XMLTagHandler
    void SetName(const wxString& name);
    const wxString& GetName() const;
 
+   sampleCount TimeToSamples(double time) const noexcept;
+   double SamplesToTime(sampleCount s) const noexcept;
+
+   void SetSilence(sampleCount offset, sampleCount length);
```
Paul-Licameli, line 389: Comment that offset is relative to playable start

src/WaveClip.cpp
```diff
@@ -1789,3 +1712,233 @@ const wxString& WaveClip::GetName() const
 {
    return mName;
 }
+
+void WaveClip::ClearSequence(double t0, double t1)
```
Paul-Licameli, line 1716: Move this function, perhaps, so it is just after WaveClip::Clear, so the diff of the two versions is a bit smaller, more corresponding unchanged lines.

src/WaveClip.cpp
```diff
@@ -1380,51 +1321,69 @@ void WaveClip::Paste(double t0, const WaveClip* other)
...
+      newCutlines.push_back(std::move(cutlineCopy));
    }
 
-   sampleCount s0;
-   TimeToSamplesClip(t0, &s0);
+   sampleCount s0 = TimeToSequenceSamples(t0);
 
    // Assume Strong-guarantee from Sequence::Paste
-   mSequence->Paste(s0, pastedClip->mSequence.get());
+   mSequence->Paste(s0, newClip->mSequence.get());
```
Paul-Licameli, line 1379: There were three cases above:  pasting exactly at the play start time; exactly at play end; or other.

I think that in the first of these cases, the two lines above are incorrect.  Because ClearSequence() has removed some intial segment of samples, you need to paste at sample 0.

And I think you need to reassign mSequenceStartTime correctly in this case.

Am I mistaken?

Paul-Licameli (changes requested):
I think there are errors in one of the paths through pasting of WaveClip

src/WaveClip.cpp
```diff
@@ -1789,3 +1712,233 @@ const wxString& WaveClip::GetName() const
...
+{
+    return sampleCount(floor(time * mRate + 0.5));
+}
+
+double WaveClip::SamplesToTime(sampleCount s) const noexcept
+{
+    return s.as_double() / mRate;
+}
+
+void WaveClip::SetSilence(sampleCount offset, sampleCount length)
```
Paul-Licameli, line 1783: A comment should clarify the offset is relative to play start

src/WaveTrack.cpp
```diff
@@ -533,12 +538,22 @@ void WaveTrack::SetWaveColorIndex(int colorIndex)
...
+{
+    sampleCount result{ 0 };
+
+    for (const auto& clip : mClips)
+        result += clip->GetPlaySamplesCount();
+
+    return result;
+}
+
+sampleCount WaveTrack::GetSequenceSamplesCount() const
```
Paul-Licameli, line 551: comment should carify that cut lines are not counted

src/WaveClip.cpp
```diff
@@ -1789,3 +1712,233 @@ const wxString& WaveClip::GetName() const
...
+{
+    GetSequence()->SetSilence(sampleCount(floor(GetTrimLeft() * GetRate())) + offset, length);
```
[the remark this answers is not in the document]
vsverchinsky: Yes, it should, thanks!

src/WaveClip.cpp
```diff
@@ -1433,8 +1392,12 @@ void WaveClip::Paste(double t0, const WaveClip* other)
 /*! @excsafety{Strong} */
 void WaveClip::InsertSilence( double t, double len, double *pEnvelopeValue )
 {
-   sampleCount s0;
-   TimeToSamplesClip(t, &s0);
+   if (t == GetPlayStartTime() && t > GetSequenceStartTime())
+      ClearSequence(GetSequenceStartTime(), t);
+   else if (t == GetPlayEndTime() && t < GetSequenceEndTime())
+      ClearSequence(t, GetSequenceEndTime());
```
Paul-Licameli, line 1398: I don't understand the four lines above.

This function isn't doing what the name suggests -- just inserting silence -- but it may also discard the trimmed samples.

Maybe there is a reason for that, can you explain?  Should there be a different function for that purpose?

But is it also a bug in that case, that the trim-left and trim-right values in the clip are not reset to 0?

Paul-Licameli: I think most paths to WaveClip::InsertSilence involve newly created empty tracks, so it doesn't matter, but an exception is that WaveTrack::Join calls WaveClip::AppendSilence which calls this.

Did you insert those lines only to handle this special case?

Paul-Licameli (changes requested):
Explanation at least in comments, if you can prove there isn't a mistake.

src/WaveTrack.cpp
```diff
@@ -1158,14 +1162,13 @@ void WaveTrack::HandleClear(double t0, double t1,
...
-                  leftClip->Clear(t0, clip->GetEndTime());
+                  leftClip->TrimRight(clip->GetPlayEndTime() - t0);
                   clipsToAdd.push_back(std::move(leftClip));
 
                   auto rightClip = std::make_unique<WaveClip>(*clip, mpFactory, true);
                   //i18n-hint Name assigned to the second clip when splitting a wave track
                   rightClip->SetName(XC("%s 2", "clip name template").Format(rightClip->GetName()).Translation());
-                  rightClip->Clear(clip->GetStartTime(), t1);
-                  rightClip->Offset(t1 - clip->GetStartTime());
+                  rightClip->SetTrimLeft(t1 - rightClip->GetPlayStartTime());
```
Paul-Licameli, line 1171: I think TrimLeft was intended, not SetTrimLeft:  else, clearing the middle of a clip, which already had some left trimming, will give wrong results for the right hand new clip.

Am I correct?

Paul-Licameli (changes requested):

src/WaveClip.cpp
```diff
@@ -1567,20 +1493,16 @@ void WaveClip::ClearAndAddCutLine(double t0, double t1)
...
+   auto s1 = TimeToSequenceSamples(t1);
 
    // use Weak-guarantee
    GetSequence()->Delete(s0, s1-s0);
 
    // Collapse envelope
    auto sampleTime = 1.0 / GetRate();
    GetEnvelope()->CollapseRegion( t0, t1, sampleTime );
-   if (t0 < GetStartTime())
-      Offset(-(GetStartTime() - t0));
```
Paul-Licameli, line 1582: I was wondering about these unexplained deleted lines... but I think we can prove that they never happened because of the lines in WaveTrack::HandleClear():

```
         if (!clip->BeforePlayStartTime(t1) && !clip->AfterPlayEndTime(t0) &&
               (clip->BeforePlayStartTime(t0) || clip->AfterPlayEndTime(t1)))
         {
            addCutLines = false;
            break;
         }
```

So, okay

src/WaveTrack.cpp
```diff
@@ -1255,201 +1258,189 @@ void WaveTrack::SyncLockAdjust(double oldT1, double newT1)
...
-               // If clips are immovable we also allow prepending to clips
-               if (clip->WithinClip(t0) ||
-                     TimeToLongSamples(t0) == clip->GetStartSample())
-               {
-                  insideClip = clip.get();
-                  break;
-               }
+                // If clips are immovable we also allow prepending to clips
+                if (clip->WithinPlayRegion(t0) ||
+                    TimeToLongSamples(t0) == clip->GetSequenceStartSample())
```
Paul-Licameli, line 1348: Should this be GetPlayStartSample()?

Paul-Licameli (changes requested):

Paul-Licameli:
At last, I consider this review complete!

src/WaveClip.cpp
vsverchinsky: There is also another comment saying:
` One and only one of the following is true for a given t (unless the clip has zero length -- then BeforeClip() and AfterClip() can both be true). Within() is true if the time is substantially within the clip`. 
I think this is more about the definition of 'within'. Changing the implementation will require a careful review of the usage of each function. And I think that changing '<=' to '<' was anyway correct, because 'GetPlayEndSample()' is the index of the sample following the last one, so 'within' includes the last sample, but does not include the first sample.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
vsverchinsky: Thanks!

src/WaveClip.cpp
vsverchinsky: If I understand you correctly, then I think that the answer to the first question is: 
`sampleCount s0 = TimeToSequenceSamples(t0);`
`s0` should be 0 for given `t0` after `ClearSequence(GetSequenceStartTime(), t0);`
But I doubt that pasting in front of the clip is possible at all, taking into account implementation of the `WithinPlayRegion`

src/WaveClip.cpp
vsverchinsky: The reason was that I considered `InserSilence(t0, length, envelope)` to be equivalent to `Paste(t0, silenceClip)`. And `Paste` discards the hidden part of the destination clip in case if the clip is prepended or appended. And I thought that this was the safest approach in these cases, but maybe you are right and hidden data should not be discarded (`Paste` should not discard too)  and `WaveTrack::Join` should handle this case instead.

src/WaveTrack.cpp
vsverchinsky: Yes, thanks!

src/WaveTrack.cpp
vsverchinsky: Hm, yes, and this makes possible the case with prepending data in `Paste`. Thanks!

Paul-Licameli (approved):

Issue #1681, opened 2021-09-15
Position of New Label dialog is inconvenient


DavidBailes:
**Describe the bug**
The new label dialog, used for entering the name of a new label, is always at the far right hand side of the dialog. The problem occurs both when adding a label at selection, and when adding label at playback position.

**To Reproduce**
1. Enable the "use dialog for name of new label" check box in the track behaviour page of preferences.
2. Press Ctrl+B. The new label dialog appears at the far right of the Audacity window, irrespective of the cursor position.
3. The user has to unnecessarily move their gaze.

![image](https://user-images.githubusercontent.com/11887021/133456553-2be4cd99-52cc-41de-98af-5d43c65611d2.png)

**Expected behavior**
The horizontal position of the New label dialog should be such that it's close to cursor position, and where the new label will be created. In Audacity 3.0.4, and previous versions which had this dialog, the initial horizontal position of the cursor in the name text box in the New Label dialog is roughly the same as the horizontal position of the cursor.
Using Audacity 3.0.4:
![image](https://user-images.githubusercontent.com/11887021/133456115-3da5328e-5810-47b0-9349-38e96f83cd51.png)

**Additional information (please complete the following information):**
Tested on Windows 10

DavidBailes:
Not sure that this should be labelled as enhancement - existing desirable behaviour has been broken. Isn't this a bug?

LWinterberg:
ah, sorry, I didn't read the bit that said that it used to work differently.

petersampsonaudacity:
Testing on W10 with audacity-win-3.1.0-alpha-20210916+cb1931f-64bit

With Audacity at full screen and using Ctrl+M to create a label at playback part of the edit/create label dialog is off-screen:
![image](https://user-images.githubusercontent.com/50205984/133922878-61158f82-e7c1-4866-8354-ac3ace795f84.png)


If I reduce the scree size the fialg is outside the Audacity window:
![image](https://user-images.githubusercontent.com/50205984/133922978-c90f7bdf-aadc-4144-8bd9-43c6afc26203.png)

petersampsonaudacity:
I black-box tested this change of @DavidBailes on W10 with Audacity_Windows_32bit_1253483697_affc336 

With both full screen window and reduced window this now looks much better - a good improvement.

a) The dialog with Ctrl+B is popped under the cursor position
b) The dialog with Ctrl+M is popped under the playhead/recordhead position
c) In no case does the dialog overflow the window area.

PR #1690, opened 2021-09-20
Issue 1681: Position of New Label dialog is inconvenient

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/1681

If the "use dialog for name of new label" check box is checked in the track behaviour page of preferences, then when adding a label at either selection or playback position, a dialog is used for entering the name of the new label. For convenience the horizontal position of the dialog should be such that the initial horizontal position of the cursor in the edit box is roughly the same as the horizontal position of the cursor or playback position respectively.

This behaviour was broken by commit 8f8ec41.
In this commit in LabelTrackView::DialogForLabelName(), the two arguments passed to viewInfo.TimeToPosition() are in wrong way round.

The fix: swap the order of these arguments.

Recommended:


DavidBailes:
Fix does not work in all circumstances

PR #1691, opened 2021-09-20
Issue 1681: Position of New Label dialog is inconvenient

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/1681

If the "use dialog for name of new label" check box is checked in the track behaviour page of preferences, then when adding a label at either selection or playback position, a dialog is used for entering the name of the new label. For convenience the horizontal position of the dialog should be such that the initial horizontal position of the cursor in the edit box is roughly the same as the horizontal position of the cursor or playback position respectively.

This behaviour was broken by commit 8f8ec41.
In LabelTrackView::DialogForLabelName(), there was the change from:
   position.x += viewInfo.GetLabelWidth()
      + std::max(0, static_cast<int>(viewInfo.TimeToPosition(region.t0())))
      -40;

to:
   position.x +=
      + std::max(0, static_cast<int>(viewInfo.TimeToPosition(
         viewInfo.GetLeftOffset(), region.t0())))
      - 39;

The change that was causing the most problem is that the arguments passed to viewInfo.TimeToPosition() are the wrong way round. However in addition the original code guarded against the position of the cursor being negative - in that case it was treated as zero (this prevented the dialog from going off the left of Audacity's main window).

The fix: revert to the original structure of the statement.

Recommended:

### DIFF
--- a/src/tracks/labeltrack/ui/LabelTrackView.cpp
+++ b/src/tracks/labeltrack/ui/LabelTrackView.cpp
@@ -2255,9 +2255,8 @@ int LabelTrackView::DialogForLabelName(
       trackPanel.FindTrackRect( trackFocus.Get() ).GetBottomLeft();
    // The start of the text in the text box will be roughly in line with the label's position
    // if it's a point label, or the start of its region if it's a region label.
-   position.x +=
-      + std::max(0, static_cast<int>(viewInfo.TimeToPosition(
-         viewInfo.GetLeftOffset(), region.t0())))
+   position.x += viewInfo.GetLeftOffset()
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
I black-box tested this change of @DavidBailes on W10 with Audacity_Windows_32bit_1253483697_affc336 

With both full screen window and reduced window this now looks much better - a good improvement.

a) The dialog with Ctrl+B is popped under the cursor position

a) The dialog with Ctrl+M is popped under the playhead/recordhead position

In no case does the dialog overflow the window area.

**_BUT note that this did not build on Mac_**

DavidBailes:
Could this be reviewed and merged?
It has been tested by Peter.

Paul-Licameli (approved):

PR #1693, opened 2021-09-20
Lower CPU usage by doing less display updates

mikaoj:
**TL;DR:** Lowers CPU usage from 40-50% to 10-15% while recording.

Currently Audacity consumes ~90% CPU when recording on my 2019 Macbook Pro with 2,3 GHz 8-Core Intel Core i9 running macOS 11.6. That will kick the fans into jet engine mode which will be audible if you have the computer in the same room.
One trick is to toggle off "update display while playing" but it will still use 40-50%.

By removing this code (and having update display while playing toggled off) I'm able to get it to 10-15%.
I haven't noticed any artifacts or components not updating correctly, so to me this forced update seems to be done needlessly. But I'm expecting to be told otherwise in this PR :)

Recommended:

### DIFF
--- a/src/TrackPanel.cpp
+++ b/src/TrackPanel.cpp
@@ -287,7 +287,6 @@ TrackPanel::TrackPanel(wxWindow * parent, wxWindowID id,
 
    mTrackArtist = std::make_unique<TrackArtist>( this );
 
-   mTimeCount = 0;
    mTimer.parent = this;
    // Timer is started after the window is visible
    ProjectWindow::Get( *GetProject() ).Bind(wxEVT_IDLE,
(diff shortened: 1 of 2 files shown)


crsib:
Thank you a lot for your contribution! We have plans to fix the rendering performance, so we can keep the visual feedback from the recording (and make it smooth!) for 3.2

mikaoj:
> so we can keep the visual feedback from the recording

Out of curiosity so I learn something about the code base. Which visual feedback did this PR break? :)

crsib:
Well, waveform wasn't redrawn while recording unless the horizontal scrolling started :)

Issue #1279, opened 2021-07-12
Help>Check for Updates - does not actually check for updates


petersampsonaudacity:
**Describe the bug**

**To Reproduce**
1. Go to Help>Check for Updates
2. Observe: you are taken to the Update page of the Audacity Website
https://www.audacityteam.org/download/?from_ver=3.0.3&Time=Jul9202117:10:00

**Expected behavior**
Given that Audacity now has internet enabled update checking - check for updates should actually perform a proper update check - not just take the user to a wen page where they must do the checking.

This command should temporarily over-ride the setting a user may have in Application preferences for Update Checking to be "off" - BUT it should **_NOT_** change that setting to be on.  It is a perfectly valid use-case  that a user may wish to make a one-off on-demand check without having auto-checking turned "on"

**Screenshots**
Not applicable.

**Additional information (please complete the following information):**
 - Tested on W10 with latest 3.0.3 alpha:  Audacity_Windows_64bit_1015901279_01da348 
 - But applies to all OS

**Additional context**
None.

gera-gas:
**Fixes way:**
1. Make the opt-in of `Help>Check for Updates` item by `audacity_has_updates_check` CMake option.
2. Call `UpdateManager::Get().GetUpdates(true)` from `Help>Check for Updates` to checking new version (download and installing) if update option was enabled.

gera-gas:
Task required: 6-8 hrs

Paul-Licameli:
@LWinterberg Remember manual updates for this?

petersampsonaudacity:
Tested on W10 with audacity-win-3.1.0-alpha-20210924+97ca83e-64bit

This now looks fine

![image](https://user-images.githubusercontent.com/50205984/134725795-b5908333-7462-4424-aa8a-7f8f4353038f.png)

Penikov:
Well, there are three ways to check updates now. 
- automatically (once per day, raises the message if updates are available )
- help-check for updates (that is @petersampsonaudacity mentioned)
- the link on the welcome screen that leads a user to the website
![изображение](https://user-images.githubusercontent.com/25851359/135604138-fd009ebd-7816-4448-8513-3e7044354806.png)


All three are working good now.

PR #1697, opened 2021-09-21
Help_menu-check_for_update corrections

gera-gas:
Resolves: #1279

Recommended:

Dialog with "No updates available" information, for user notification in this case:

### DIFF
--- a/src/menus/HelpMenus.cpp
+++ b/src/menus/HelpMenus.cpp
@@ -449,7 +449,7 @@ void OnMenuTree(const CommandContext &context)
 #if defined(HAVE_UPDATES_CHECK)
 void OnCheckForUpdates(const CommandContext &WXUNUSED(context))
 {
-    UpdateManager::GetInstance().GetUpdates(false);
+    UpdateManager::GetInstance().GetUpdates(false, false);
 }
 #endif
 
--- a/src/menus/HelpMenus.cpp
+++ b/src/menus/HelpMenus.cpp
@@ -582,7 +582,7 @@ BaseItemSharedPtr HelpMenu()
       ,
 #endif
          // DA: Does not fully support update checking.
-   #if !EXPERIMENTAL_DA && defined(HAVE_UPDATES_CHECK)
+   #if !defined(EXPERIMENTAL_DA) && defined(HAVE_UPDATES_CHECK)
          Command( wxT("Updates"), XXO("&Check for Updates..."),
             FN(OnCheckForUpdates),
             AlwaysEnabledFlag ),
--- a/src/menus/HelpMenus.cpp
+++ b/src/menus/HelpMenus.cpp
@@ -28,6 +28,10 @@
 #include "../widgets/AudacityMessageBox.h"
 #include "../widgets/HelpSystem.h"
 
+#if defined(HAVE_UPDATES_CHECK)
+#include "update/UpdateManager.h"
+#endif
+
 // private helper classes and functions
 namespace {
 
@@ -445,7 +449,7 @@ void OnMenuTree(const CommandContext &context)
 #if defined(HAVE_UPDATES_CHECK)
 void OnCheckForUpdates(const CommandContext &WXUNUSED(context))
 {
-   ::OpenInDefaultBrowser( VerCheckUrl());
+    UpdateManager::GetInstance().GetUpdates(false);
 }
 #endif
 
--- a/src/menus/HelpMenus.cpp
+++ b/src/menus/HelpMenus.cpp
@@ -442,10 +442,12 @@ void OnMenuTree(const CommandContext &context)
       Verbatim("Menu Tree"), wxT("menutree.txt"), true );
 }
 
+#if defined(HAVE_UPDATES_CHECK)
 void OnCheckForUpdates(const CommandContext &WXUNUSED(context))
 {
    ::OpenInDefaultBrowser( VerCheckUrl());
 }
+#endif
 
 void OnAbout(const CommandContext &context)
 {
@@ -575,9 +577,8 @@ BaseItemSharedPtr HelpMenu()
 #else
       ,
 #endif
-
          // DA: Does not fully support update checking.
-   #ifndef EXPERIMENTAL_DA
+   #if !EXPERIMENTAL_DA && defined(HAVE_UPDATES_CHECK)
          Command( wxT("Updates"), XXO("&Check for Updates..."),
             FN(OnCheckForUpdates),
             AlwaysEnabledFlag ),
--- a/src/update/NoUpdatesAvailableDialog.cpp
+++ b/src/update/NoUpdatesAvailableDialog.cpp
@@ -21,7 +21,7 @@
 
 static const auto title =
 /* i18n-hint: Title of the dialog no updates available. */
-XO("No updates available");
+XO("No Updates Available");
 
 BEGIN_EVENT_TABLE(NoUpdatesAvailableDialog, wxDialogWrapper)
     EVT_BUTTON(wxID_OK, NoUpdatesAvailableDialog::OnOk)
@@ -56,7 +56,7 @@ NoUpdatesAvailableDialog::NoUpdatesAvailableDialog(wxWindow* parent)
                XO("If you want to change your preference for automatic updates checking, you can find it in %s."));
 
             preferencesMessage.FormatLink(
-               /* i18n-hint: Hyperlink title that opens Preferences dialog on Application page and endian part of "... you can find it in %s" string. */
+               /* i18n-hint: Hyperlink title that opens Preferences dialog on Application page and is substituted into "... you can find it in %s." string. */
                wxT("%s"), XO("Preferences > Application"), [parent, this]() {
                   EndModal(wxID_OK);
 
--- a/src/update/NoUpdatesAvailableDialog.cpp
+++ b/src/update/NoUpdatesAvailableDialog.cpp
@@ -2,7 +2,8 @@
 
   Audacity: A Digital Audio Editor
 
-  NoUpdatesAvailableDialog.cpp
+  @file NoUpdatesAvailableDialog.cpp
+  @brief Define a dialog with Information about no updates available.
 
   Anton Gerasimov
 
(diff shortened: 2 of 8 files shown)


src/menus/HelpMenus.cpp
```diff
@@ -574,9 +580,8 @@ BaseItemSharedPtr HelpMenu()
 #else
       ,
 #endif
-
          // DA: Does not fully support update checking.
-   #ifndef EXPERIMENTAL_DA
+   #if !EXPERIMENTAL_DA && defined(HAVE_UPDATES_CHECK)
```
Paul-Licameli, line 584: Please be consistent with other tests of EXPERIMENTAL flags:  check whether defined,
don't check the value.
```
#if !defined(EXPERIMENTAL_DA) && defined(HAVE_UPDATES_CHECK)
```

src/update/NoUpdatesAvailableDialog.cpp
```diff
@@ -0,0 +1,95 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  NoUpdatesAvailableDialog.cpp
```
Paul-Licameli, line 5:
```
@file NoUpdatesAvailableDialog.cpp
```
This helps Doxygen generate more information

src/update/NoUpdatesAvailableDialog.h
```diff
@@ -0,0 +1,26 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  NoUpdatesAvailableDialog.h
```
Paul-Licameli, line 5: Add @file here too
also, in the header, a line
```
@brief <your short explanation of the purpose of the file>
```

src/update/NoUpdatesAvailableDialog.h
```diff
@@ -0,0 +1,26 @@
...
+  Anton Gerasimov
+
+**********************************************************************/
+
+#pragma once
+
+	
+#include "widgets/wxPanelWrapper.h"
+
+/// Information dialog about no updates available, that allows to navigate to settings quickly
```
Paul-Licameli, line 16: Doxygen comment here, that's good.

Doxygen recognizes a few different formattings, and we are not wholly consistent but I prefer `//!` or `/*! ... */` comments

Paul-Licameli (changes requested):
This is not a complete review.  There are some minor matters of comment and preprocessor conventions.

src/update/NoUpdatesAvailableDialog.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "ShuttleGui.h"
+
+#include "prefs/PrefsDialog.h"
+#include "ui/AccessibleLinksFormatter.h"
+
+#include <wx/stattext.h>
+
+static const auto title =
+/* i18n-hint: Title of the dialog no updates available. */
+XO("No updates available");
```
Paul-Licameli, line 24: Maybe this should capitalize every word?

src/update/NoUpdatesAvailableDialog.cpp
```diff
@@ -0,0 +1,95 @@
...
+            wxFont font = titleCtrl->GetFont().MakeLarger().MakeBold();
+
+            titleCtrl->SetFont(font);
+
+            AccessibleLinksFormatter preferencesMessage(
+               /* i18n-hint: %s is replaced with 'Preferences > Application'. */
+               XO("You are on the latest version of Audacity. Consider turning %s on in Preferences, if you want to be informed automatically about new versions."));
+
+            preferencesMessage.FormatLink(
+               /* i18n-hint: Hyperlink title that opens Preferences dialog on Directories page. */
```
Paul-Licameli, line 59: i18n-hint should mention that this is the object of "Consider turning on..."

So that the translator for a language like Russian, with cases, will know which grammatical ending to use.

Paul-Licameli (changes requested):

src/update/UpdateManager.cpp
```diff
@@ -247,6 +248,10 @@ void UpdateManager::GetUpdates(bool ignoreNetworkErrors)
 #if UPDATE_LOCAL_TESTING == 0
         else // mVersionPatch.version > CurrentBuildVersion()
         {
+            gAudioIO->CallAfterRecording([] {
+                NoUpdatesAvailableDialog(nullptr).ShowModal();
```
Paul-Licameli, line 252: This dialog contains a suggestion to turn on update checking in preferences.  But maybe that suggestion should be conditional on `configurableNotification == false` ?  Should it be part of the dialog only when you use the Help menu, and not if it opens automatically?

src/update/NoUpdatesAvailableDialog.cpp
```diff
@@ -0,0 +1,96 @@
...
+         {
+            wxStaticText* titleCtrl = S.AddVariableText(title, false, 0, 500);
+
+            wxFont font = titleCtrl->GetFont().MakeLarger().MakeBold();
+
+            titleCtrl->SetFont(font);
+
+            AccessibleLinksFormatter preferencesMessage(
+               /* i18n-hint: %s is replaced with 'Preferences > Application'. */
+               XO("You are on the latest version of Audacity. Consider turning %s on in Preferences, if you want to be informed automatically about new versions."));
```
Paul-Licameli, line 56: Maybe the message should only contain the first sentence, in the case that this dialog opens in the case of automatic update checking that the user has already enabled.

So you could check the state of`DefaultUpdatesCheckingFlag` here.

Paul-Licameli:
My review is complete.  There are a few small changes to do.

src/update/NoUpdatesAvailableDialog.cpp
gera-gas: In current design it should without every word capitalization.

src/update/NoUpdatesAvailableDialog.cpp
gera-gas: I think we should keep of design, because in case when user forgot current update preferences state, it easy and fast way to move to update preferences and check it.

src/update/NoUpdatesAvailableDialog.cpp
Paul-Licameli: Designs can be mistaken.

src/update/NoUpdatesAvailableDialog.cpp
Paul-Licameli: No, consider.  If we can tell the preference is enabled, it is silly to suggest that the user enables it.  If this code executes because if automatic checking, not the help menu item, then the preference must have been on.

gera-gas: But what you think about case when user forget current update settings state? Just sometimes, I use the same fast link in other applications, that I long time don't using and want to do (just not rememder where update settings was described), I use fast way to check new version and check update settings.

Or course this is may only my case, but may be someone using this way like as me, for fast redirecting to settings. What you think about this?

src/update/NoUpdatesAvailableDialog.cpp
gera-gas: Okay, I will write Jouni to rechecking it now.

src/update/NoUpdatesAvailableDialog.cpp
Paul-Licameli: Are you saying the reminder should suggest turning the preference on or off?  Instead of what the words say now.

gera-gas: Yes

src/update/NoUpdatesAvailableDialog.cpp
jounih: Title Case vs Sentence case for titles... Americans like the former, in the UK the latter is more common. If Audacity already has an established title style using Title Case, then let's go with that

src/update/NoUpdatesAvailableDialog.cpp
jounih: Either:

A) Let's do what Paul suggests, ie. check if the user has automatic updates on, and show a version of this screen without the text reference to turning automatic updates on
OR

B) we could change that text to be more generic, and show the same text for everyone, whether or not they have automatic updates enabled. ie. "If you want to turn your preference for automatic updates on or off, you can find it in Preferences (link)"

src/update/NoUpdatesAvailableDialog.cpp
Paul-Licameli: @petersampsonaudacity can you remind us where "Connie" wrote down her guidelines?

petersampsonaudacity: > 
> 
> @petersampsonaudacity can you remind us where "Connie" wrote down her guidelines?

https://alphamanual.audacityteam.org/man/Consistency#Capitalization_of_page_title_and_section_headings

jounih: thanks @petersampsonaudacity !

So sentence case it is for titles

petersampsonaudacity: > thanks @petersampsonaudacity !
> 
> So sentence case it is for titles

Well really the tanks for that actually go to Gale Andrews (RIP)

Paul-Licameli: Ah, so this was really enforced for code and in fact the opposite convention, "Title Case" not "Sentence case", actually prevails, as you can see in manual images, or by running the program, or by scanning code for `wxDialogWrapper` and seeing the strings passed to its constructor.

So I say "Title Case" is right, to be consistent with other dialogs.  Consistency with manual pages is the wrong criterion.

Paul-Licameli: Examples that exist now
```
VST Effect Options
Audio Unit Effect Options
Import Audio Unit Presets
LADSPA Effect Options
LV2 Effect Settings
Select Preset
Save Preset
Manage Curves List
Command Output
Export Multiple
Import Raw Data
Audacity Preferences
Label Track Font
Set Rate
Recording Meter Options
Playback Meter Options
Audacity Configuration Error
Automatic Crash Recovery
Select Command
Locate FFmpeg
Compact Project
Sound Activated Record
Welcome to Audacity!
Edit Genres
Audacity Timer Record
```

counterexamples
```
Invalid sample rate
Configure custom FFmpeg options
FFmpeg not found
Edit Labels

```

petersampsonaudacity: > So I say "Title Case" is right, to be consistent with other dialogs. Consistency with manual pages is the wrong criterion.

It's your call now _(yours and @Tantacrul 's anyway)_ - Gale-the-Enforcer is no longer with us.

For me the important thing would be consistency
a) The app should be consistent within itself
b) the manual should be consistent within itself
c) the two can be different if that is what you want

Paul-Licameli: I propose:
* this new dialog follows the convention of most others, which is "Title Case"
* Connie should say explicitly that "Sentence case" applies to manual pages but the opposite applies to title dialogs
* Open an issue to fix the few inconsistent dialogs

Paul-Licameli (approved):

petersampsonaudacity:
Tested on W10 with audacity-win-3.1.0-alpha-20210924+97ca83e-64bit

This now looks fine

image

Issue #1571, opened 2021-08-30
The project file is being corrupt if it's opened while Audacity is running.


Penikov:
**Describe the bug**
The project file is being corrupt if it's opened while Audacity is running.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run Audacity
2. Open some project
3. Press Cmd + O and open another project

**Actual behaviour**
Audacity corrupts the second project

**Expected behaviour**
Another project is opened successfully

**Screenshots**
https://www.srut.de/audacity/audacity-3040-problem--opening-a-file-via-keyboard-shortcut-corrupts-file-irreversibly.mov

**Additional information (please complete the following information):**
 - OS:  Mac OS Big Sur
 - Version Audacity 3.0.4

JohnColket:
This issue has been reported by at least four(4) different users on the forum on the following two threads:
"aup-3 files are empty when they are opened":  https://forum.audacityteam.org/viewtopic.php?f=47&t=119330&sid=18e2f302da2d3b1fd5bad5a0c59c4542
and "Opening a file via keyboard shortcut corrupts project file irreversibly": https://forum.audacityteam.org/viewtopic.php?f=47&t=119956
The four reporting users are: Baum123, Carpezaldo, lilouch, and mathpoli.

petersampsonaudacity:
I cannot reproduce this on W10 with 3.1.0 alpha audacity-win-3.1.0-alpha-20210831+6d3dd0c-64bit

Both projects open fine - no cirruption that I can detect

mathpoli:
Hi, may I point you to what I have found out and posted on the Audacity forum?
https://forum.audacityteam.org/viewtopic.php?f=47&t=119956&start=20

mathpoli:
SUMMARY:
So far I have tested on these iMacs and had on all the bug:

– 1x iMac (Retina 5K, 27 Inch, Late 2015), Intel Core i5, with macOS 10.15.7 (internal Fusion Drive): Audacity v3.0.4.0, Audacity v3.0.3.0, Audacity v3.0.2.0
same iMac with macOS Big Sur (external USB drive): Audacity v3.0.4.0
same iMac with macOS 10.14.6 (external USB drive): Audacity v3.0.4.0

– 2x iMac (Retina 5K, 27 Inch, Late 2015), Intel Core i7, with macOS 10.15.7 (internal SSD, first partition), Audacity v3.0.4.0
[...]

PR #1701, opened 2021-09-22
Fix bug 1571

gera-gas:
Convert to C locale decimal point for stable parsing. In `Cmd+O` user case, `endptr` (into `wxString::ToDouble`) point to localized decimal point, that means it can't recognize separator char, therefore in result `ToDouble()` return false (`!*endptr`).

Resolves: #1571 

Recommended:

### DIFF
--- a/libraries/lib-strings/Internat.cpp
+++ b/libraries/lib-strings/Internat.cpp
@@ -135,9 +135,9 @@ bool Internat::CompatibleToDouble(const wxString& stringToConvert, double* resul
 {
    // Regardless of the locale, always respect comma _and_ point
    wxString s = stringToConvert;
-   s.Replace(wxT(","), wxString(GetDecimalSeparator()));
-   s.Replace(wxT("."), wxString(GetDecimalSeparator()));
-   return s.ToDouble(result);
+   // Convert to C locale decimal point for stable parsing.
+   s.Replace(wxString(GetDecimalSeparator()), wxT("."));
+   return s.ToCDouble(result);
 }
 
 double Internat::CompatibleToDouble(const wxString& stringToConvert)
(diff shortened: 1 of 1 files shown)


crsib (approved):

Paul-Licameli:
I’m not satisfied yet that this bug is explained.

Paul-Licameli:
I’m satisfied with the explanation now, but I want to explore uses of the changed function and be sure there are no unintended consequences.

libraries/lib-strings/Internat.cpp
```diff
@@ -135,9 +135,9 @@ bool Internat::CompatibleToDouble(const wxString& stringToConvert, double* resul
 {
    // Regardless of the locale, always respect comma _and_ point
    wxString s = stringToConvert;
-   s.Replace(wxT(","), wxString(GetDecimalSeparator()));
-   s.Replace(wxT("."), wxString(GetDecimalSeparator()));
-   return s.ToDouble(result);
+   // Convert to C locale decimal point for stable parsing.
+   s.Replace(wxString(GetDecimalSeparator()), wxT("."));
```
Paul-Licameli, line 139: Better to preserve old behavior by writing this replacement instead:
```
s.Replace(wxT(","), wxT("."))
```
So even when `GetDecimalSeparator()` is not `","`, a string with comman for decimal point still produces the same result.

This matters for certain uses of this function that are not part of serialization of files, such as the interpretation of Nyquist programs.  The old behavior in fact allows either . or , for decimal point, no matter what locale is set.  Don't change that.

Paul-Licameli (changes requested):
Should preserve certain old behavior, with a different substitution.

libraries/lib-strings/Internat.cpp
gera-gas: I think we should keep both replacement, like this:
```cpp
s.Replace(wxT(","), wxT("."));
s.Replace(wxString(GetDecimalSeparator()), wxT("."));
```
What you think?

Paul-Licameli (approved):

Issue #1613, opened 2021-09-02
stereo track clip handles don't affect both channels when grabbing sample 1 


LWinterberg:
**Steps to reproduce**

1. Have a stereo track
2. Zoom right in and to to the start to the track
3. Drag the stereo track using the affordances above the very first sample (rounded, so anything up to "sample 1.5" works)
4. The left channel will move independently from the right channel

**Additional information (please complete the following information):**
 - OS: Windows 11
 - Version Audacity 3.1.0-alpha-20210901

**Additional context**
(will provide a video soon)

DavidBailes:
Testing using the move tool, rather than the drag bar, the last time this worked correctly was 2.4.2.
This bug also affects the commands time shift left/right which can be used by keyboard users to move clips. A user may well use these commands with the cursor positioned at the start of the clip.

Issue #1689, opened 2021-09-20
Clip visuals become unpredictable on sample-level zooms


LWinterberg:
**Describe the bug**
Instead of looking like clips, the clips seemingly extend over empty space, terminate twice, and more. Hovering over the actual clip won't highlight the actual clip but the wrong side of it. 

**To Reproduce**
1. Have a clip
2. Zoom way in (to the sample level and then 1-4 steps further still

**Expected behavior**
They remain looking like clips

**Screenshots**
![image](https://user-images.githubusercontent.com/87814144/133983919-134552b1-eb3d-4d85-b028-c19b881f491c.png)
![image](https://user-images.githubusercontent.com/87814144/133983988-ea7c9325-2d01-4888-a8df-d858688131e1.png)
![image](https://user-images.githubusercontent.com/87814144/133984109-52137933-f751-4945-ac5d-045e7fa2f749.png)
![image](https://user-images.githubusercontent.com/87814144/133984288-d014341d-6baf-4a88-8948-b02deeda9b0b.png)

**Additional information (please complete the following information):**
 - OS: Windows 11
 - Version Audacity 3.1.0-alpha-20210916

LWinterberg:
![image](https://user-images.githubusercontent.com/87814144/133985328-b1ebf3fb-e2da-469c-808a-d239d2f4801c.png)
for completeness: It also happens at the end of the clips. Some levels are fine, some are wrong-way-round

Penikov:
@LWinterberg Looks like this is a win11-specific bug. I've tried to reproduce this on Win10 and haven't got the same glitch:

![изображение](https://user-images.githubusercontent.com/25851359/133986003-265e2eb9-4394-4f8b-bb95-0cbded722812.png)

Penikov:
@LWinterberg Looks like this is a win11-specific bug. I've tried to reproduce this on Win10 and haven't got the same glitch:

vsverchinsky:
> @LWinterberg Looks like this is a win11-specific bug. I've tried to reproduce this on Win10 and haven't got the same glitch:
> 
> ![изображение](https://user-images.githubusercontent.com/25851359/133986003-265e2eb9-4394-4f8b-bb95-0cbded722812.png)

I was able to reproduce it on win10, seems like drawing needs some workaround, quite simple though

Penikov:
@vsverchinsky Yes, you right )
But I meant this one )
[...]

PR #1711, opened 2021-09-23
Wave time shift tool hit testing fix

vsverchinsky:
Resolves: #1613 

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
@@ -21,8 +21,19 @@ public:
    HitTestResult HitTest(
       double time, const ViewInfo &viewInfo, HitTestParams* ) override
    {
-      auto pClip = mpTrack->GetClipAtTime( time );
-
+      auto pClip = [&]() {
+         auto ts = mpTrack->TimeToLongSamples(time);
+         
+         for (auto clip : mpTrack->GetClips())
+         {
+            const auto c0 = mpTrack->TimeToLongSamples(clip->GetStartTime());
+            const auto c1 = mpTrack->TimeToLongSamples(clip->GetEndTime());
+            if (ts >= c0 && ts < c1)
+               return clip;
+         }
+         return std::shared_ptr<WaveClip>{};
+      }();
+      
       if (!pClip)
          return HitTestResult::Miss;
 
@@ -38,7 +49,7 @@ public:
       UnfixIntervals( [&](const auto &interval){
          return
             static_cast<WaveTrack::IntervalData*>(interval.Extra())
-               ->GetClip().get() == pClip;
+               ->GetClip() == pClip;
       } );
       
       return HitTestResult::Intervals;
@@ -52,8 +63,11 @@ public:
          auto data =
             static_cast<WaveTrack::IntervalData*>( myInterval.Extra() );
          auto clip = data->GetClip().get();
-         return !(clip->IsClipStartAfterClip(interval.Start()) ||
-            clip->BeforeClip(interval.End()));
+         const auto c0 = mpTrack->TimeToLongSamples(clip->GetStartTime());
+         const auto c1 = mpTrack->TimeToLongSamples(clip->GetEndTime());
+         return 
+             mpTrack->TimeToLongSamples(interval.Start()) < c1 && 
+             mpTrack->TimeToLongSamples(interval.End()) >= c0;
       });
    }
 


src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
```diff
@@ -21,8 +21,22 @@ class WaveTrackShifter final : public TrackShifter {
...
+         auto ts = mpTrack->TimeToLongSamples(time);
+         
+         for (auto clip : mpTrack->GetClips())
+         {
+            const auto c0 = mpTrack->TimeToLongSamples(clip->GetStartTime());
+            //count clip length until the beginning of the last sample, to exclude
+            //overlapping (see TimeToLongSamples implementation) with the first sample
+            //of the next clip
+            const auto c1 = mpTrack->TimeToLongSamples(clip->GetEndTime() - 1.0 / mpTrack->GetRate());
+            if (ts >= c0 && ts <= c1)
```
Paul-Licameli, line 34: ... or more simply, just use < not <= and don't adjust c1

src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
```diff
@@ -52,8 +66,12 @@ class WaveTrackShifter final : public TrackShifter {
          auto data =
             static_cast<WaveTrack::IntervalData*>( myInterval.Extra() );
          auto clip = data->GetClip().get();
-         return !(clip->IsClipStartAfterClip(interval.Start()) ||
```
Paul-Licameli, line 55: This member function of WaveClip becomes unused.  Remove it?

Paul-Licameli (changes requested):
Some simplifications possible;  also I want a better understanding why the bug happened before.

src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
```diff
@@ -52,8 +66,12 @@ class WaveTrackShifter final : public TrackShifter {
...
-            clip->BeforeClip(interval.End()));
+         const auto c0 = mpTrack->TimeToLongSamples(clip->GetStartTime());
+         //see comment for HitTest routine
+         const auto c1 = mpTrack->TimeToLongSamples(clip->GetEndTime() - 1.0 / mpTrack->GetRate());
+         return 
+             mpTrack->TimeToLongSamples(interval.Start()) <= c1 && 
+             mpTrack->TimeToLongSamples(interval.End()) >= c0;
```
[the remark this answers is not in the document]
Paul-Licameli: Now I understand!  WaveClip::BeforeClip() is really the faulty function.  It contains <= where it should have < .

Here, you just avoid the use of BeforeClip().

There are uses of BeforeClip() elsewhere, though, which should be reexamined, and may reveal more bugs like this one.

Paul-Licameli:
When was the bug introduced?  @DavidBailes said after 2.4.2.  Likely in work I did in the commit range c60bf1c9945b362d7fa2e675c22e674b1f005eea..583d347ae05dcc8b981f3716a7c7e3e414f06fc2 where I did a lot of work to make a polymorphic track shifter abstraction, so that TimeShiftHandle would not be build-dependent on subclasses of Track.

But where, more exactly, in there?

Paul-Licameli:
Commit 9d52100f5cd00864652932a469588578f20d1b16 is where the old and more correct selection logic was deleted (see function `FindClipAtTime` which used `WaveClip::GetClipAtSample` which does not have the off-by-1 error).  But it may not be the first commit where the bug happened.  The function became unused, and was replaced with a different call, at an earlier place in this complicated sequence of commits transforming the time shifting code.

src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
vsverchinsky: Yep, thanks! It's a leftover I overlooked from the previous iteration where I used time borders instead of samples

src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
vsverchinsky: It's removed in smart clips PR

Paul-Licameli: ok

src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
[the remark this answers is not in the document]
vsverchinsky: >Even if it is not different, I agree it's good to write similar tests in the two places in this file.

Yes, it is the root of the issue - clip borders were determined differently in clicked track and other channels. Another issue is that visual borders of the clip do not correspond to a actual clip borders:
image
Here selection shows the actual length of the clip, but frame is drawn with offset and a bit shorter, to prevent frame overlap with samples and overlap between adjacent clips if there is no gap between them. I think that the drawing is done wrong https://github.com/audacity/audacity/issues/1689#issuecomment-926428948 The solution might be to draw frame to exactly match the beginning and the end of a clip, but move "dot" sample by half of a sample.

vsverchinsky:
> Commit [9d52100](https://github.com/audacity/audacity/commit/9d52100f5cd00864652932a469588578f20d1b16) is where the old and more correct selection logic was deleted (see function `FindClipAtTime` which used `WaveClip::GetClipAtSample` which does not have the off-by-1 error). But it may not be the first commit where the bug happened. The function became unused, and was replaced with a different call, at an earlier place in this complicated sequence of commits transforming the time shifting code.

I think that bug was introduced here https://github.com/audacity/audacity/commit/eb22892064b5f5b5aa7c8d26842428240504d663

src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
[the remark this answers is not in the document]
Paul-Licameli: How about exploring the uses of BeforeClip() for other off-by-one bugs that we have not yet identified?

Paul-Licameli (approved):

Issue #1427, opened 2021-08-04
Range selection does not work in affordance area


AnitaBats:

petersampsonaudacity:
Why would you expect it to?

When the cursor is over the Clip-handle drag-bar the cursor changes to the open hand which is the affordance to indicate that you can click and drag to move/drag  - NOT click and select.

vsverchinsky:
Sorry, a bad wording from my side. When selection tool is active it's not possible to start selection when cursor is over empty affordance area (where no clips at all)
![image](https://user-images.githubusercontent.com/20165230/128239078-5a596da0-d070-4012-970f-540dfc2c4ede.png)

petersampsonaudacity:
> When selection tool is active it's not possible to start selection when cursor is over empty affordance area (where no clips at all)

Ah I see what you mean - there is a visual cue to the user:
a) when the cursor is over the waveform area the cursor is the I-beam (default tool setting) - indicating selection is available
b) when you move the cursor into the space where the Clip-handle drag bars could be dragged the cursor changes to the default Widows broad white arrow - indicating selection is not possible from there.

Since all the space above the waveform for the Clip-handle drag bars is really only usable by the drag bars, what I would suggest is partitioning that part of the track off with a thin black line.  That would alos have the avdantage of a visual cue suggesting that the user can slide/drag the drag bars along horizontally.  See mockup:

![image](https://user-images.githubusercontent.com/50205984/128318380-3a22745d-f36e-4e3d-b91e-23c08feac98e.png)

@Tantacrul  and @jounih  - as the UX specialists on the team what are your thoughts on this please?

Tantacrul:
> @Tantacrul and @jounih - as the UX/GUI specialists on the team what are your thoughts on this please?

I would be careful about trying to make changes to try and solve this potential issue until we've tested it in the wild. It may be the case that users don't experience this problem.

If they do, I can think of a few interesting ideas but I won't try articulate them here.

petersampsonaudacity:
@Tantacrul I don't actually see a real problem here anyway - for me the visual cue of the change of cursor icon is sufficient.

As a user I wouldn't dream of trying to start a selection while my cursor was in that dead space without the I-beam selection cursor/tool showing.

I would be surprised if we ever get user complaints about this.

Tantacrul:
[...]

PR #1717, opened 2021-09-23
Makes selection tool available at affordance area

vsverchinsky:
Resolves: #1427

Recommended:

### DIFF
--- a/src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
+++ b/src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
@@ -13,6 +13,8 @@
 #include <wx/dc.h>
 
 #include "../../../ui/AffordanceHandle.h"
+#include "../../../ui/SelectHandle.h"
+#include "../../../ui/TrackView.h"
 #include "../../../../AllThemeResources.h"
 #include "../../../../AColor.h"
 #include "../../../../NoteTrack.h"
@@ -81,6 +83,17 @@ std::vector<UIHandlePtr> NoteTrackAffordanceControls::HitTest(const TrackPanelMo
         py >= headerRect.GetTop() && py <= headerRect.GetBottom())
     {
         results.push_back(NoteTrackAffordanceHandle::HitAnywhere(mAffordanceHandle, track));
+    }
+
+    const auto& settings = ProjectSettings::Get(*pProject);
+    const auto currentTool = settings.GetTool();
+    if (currentTool == ToolCodes::multiTool || currentTool == ToolCodes::selectTool)
+    {
+        results.push_back(
+            SelectHandle::HitTest(
+                mSelectHandle, state, pProject, std::static_pointer_cast<TrackView>(track->GetTrackView())
+            )
+        );
     }
 
     return results;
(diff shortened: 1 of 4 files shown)


src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
```diff
@@ -82,6 +84,19 @@ std::vector<UIHandlePtr> NoteTrackAffordanceControls::HitTest(const TrackPanelMo
...
+        const auto& settings = ProjectSettings::Get(*pProject);
+        const auto currentTool = settings.GetTool();
+        if (currentTool == ToolCodes::multiTool || currentTool == ToolCodes::selectTool)
+        {
+            results.push_back(
+                SelectHandle::HitTest(
+                    mSelectHandle, state, pProject, std::static_pointer_cast<TrackView>(track->GetTrackView())
+                )
+            );
+        }
```
Paul-Licameli, line 98: If instead you push this handle uncondtionally, then it will be possible to start a selection drag even if the mouse is inside a clip affordance -- if you hit Escape before the click and drag.

A minor advantage that few users may discover, but why not do it, for small extra effort.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -221,7 +223,22 @@ std::vector<UIHandlePtr> WaveTrackAffordanceControls::HitTest(const TrackPanelMo
                 break;
             }
         }
-        mFocusClip = focusClip;
+        if(focusClip.use_count() > 0)
```
Paul-Licameli, line 226: That test is strange... not simply `if (focusClip)` ?

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -221,7 +223,22 @@ std::vector<UIHandlePtr> WaveTrackAffordanceControls::HitTest(const TrackPanelMo
...
+            const auto& settings = ProjectSettings::Get(*pProject);
+            const auto currentTool = settings.GetTool();
+            if (currentTool == ToolCodes::multiTool || currentTool == ToolCodes::selectTool)
+            {
+                results.push_back(
+                    SelectHandle::HitTest(
+                        mSelectHandle, state, pProject, std::static_pointer_cast<TrackView>(track->GetTrackView())
+                    )
+                );
+            }
```
Paul-Licameli, line 240: Likewise here, you might push SelectHandle always, but sometimes it is not the first hit.

Paul-Licameli (changes requested):
This looks good.  I make a few small suggestions.

src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
vsverchinsky: Thanks! Will fix that. Also currently it is not possible to make selection in affrodances area while one of the clips is beeing edited, will fix that too

Paul-Licameli: I see the new round of changes.  How is this problem fixed?

Paul-Licameli (approved):

Paul-Licameli:
I approve the changes.  If you are not going to make more fixes (I wasn't sure from your comment), then you may rebase and merge this PR.

PR #1736, opened 2021-09-24
Fix clip frame drawing

vsverchinsky:
Resolves: #1689

Also fixes clipping usage

Recommended:

### DIFF
--- a/src/TrackArtist.cpp
+++ b/src/TrackArtist.cpp
@@ -301,20 +301,42 @@ void TrackArtist::UpdatePrefs()
 
 void TrackArt::DrawClipAffordance(wxDC& dc, const wxRect& rect, const wxString& title, bool highlight, bool selected)
 {
+   wxRect clipRect;
+   //Fix #1689: visual glitches appear on attempt to draw a rectangle
+   //larger than 0x7FFFFFF pixels wide (value was discovered
+   //by manual testing, and maybe depends on OS being used), but
+   //it's very unlikely that such huge rectangle will be ever fully visible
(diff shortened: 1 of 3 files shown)


src/TrackArtist.cpp
```diff
@@ -301,20 +301,42 @@ void TrackArtist::UpdatePrefs()
 
 void TrackArt::DrawClipAffordance(wxDC& dc, const wxRect& rect, const wxString& title, bool highlight, bool selected)
 {
+   wxRect clipRect;
+   //Fix #1689: visual glitches appear on attempt to draw a rectangle
+   //larger than 0x7FFFFFF pixels wide (value was discovered
+   //by manual testing, and maybe depends on OS beeing used), but
```
Paul-Licameli, line 307: "being"

src/TrackArtist.cpp
```diff
@@ -301,20 +301,42 @@ void TrackArtist::UpdatePrefs()
...
+   //it's very unlikely that such huge rectangle will be ever fully visible
+   //on the screen, so we can safely reduce it's size to be slightly larger than
```
Paul-Licameli, line 309: "its"

Paul-Licameli (approved):
Please fix comment typos.

src/TrackArtist.cpp
vsverchinsky: Thanks!

src/TrackArtist.cpp
vsverchinsky: Thanks!

Issue #1596, opened 2021-09-01
Import preset into EQ effects fail silently when more than 200 points


petersampsonaudacity:
**Describe the bug**
When importing a preset into Filter Curve EQ or Graphic EQ where there are more than 200 frequency / value points, only the first 200 points are imported. The remaining points are discarded (ignored) without warning. There should be an error message or warning rather than failing silently.

This limitation is not currently documented in the manual.

Additionally, 200 points does not seem very generous, perhaps that could be increased a bit (I suggest 512 points as a generous maximum number of points)

Logged originally on Bugzilla as P3 2658
https://bugzilla.audacityteam.org/show_bug.cgi?id=2658

**To Reproduce**
1. Import the attached preset file into Filter Curve EQ.
[preset.txt](https://github.com/audacity/audacity/files/7092174/preset.txt)

2. Observe that only the first 200 points are imported.
3. Observe that there is no error message or warning.

**Expected behavior**
The full set of EQ points should be loaded

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/131694706-6957ccbc-399a-491c-9bba-3c76e48016ba.png)

**Additional information (please complete the following information):**
 - OS: all
 - Version:  Audacity 3.0.4 and alpha 3.1.0

**Additional context**
Steve write un the Bugzilla thread
>I have noted in the manual that Filter Curve EQ supports a maximum of 200 control points.
I don't think it is necessary to document this limitation for Graphic EQ as it would be silly to define more than 200 points for a 31 band EQ.

witwald:
If one was trying to implement a 31-band graphic EQ, then over the 20Hz to 20kHz frequency range it would be quite beneficial to be able to use more than 200 points to define the response. If all bands were in use, then sometimes more than an average of 6-points per band would help to improve the definition of the curves, particularly those that involve large peaks or notches with high Q. 512 points seems like a reasonable upper limit, providing a nominal 16-points per band if needed.

PR #1741, opened 2021-09-25
Increased max FFT size in Plot Spectrum to enable higher resolution

witwald:
Resolves: Feature enhancement

The maximum FFT size in Plot Spectrum has been increased by a factor of 2 to 131072. This change has been made to allow a higher resolution FFT to be computed. This can be useful when assessing low frequency content of audio waveforms in music. This change gives much better than 1Hz resolution at a sampling frequency of 48kHz. At a sample rate of 192kHz, a 1.47Hz resolution is obtained, which is deemed acceptable.

Also, some minor typos have been corrected, and some of the code formatting adjusted for consistency.

Recommended:

### DIFF
--- a/src/FreqWindow.cpp
+++ b/src/FreqWindow.cpp
@@ -9,7 +9,7 @@
 *******************************************************************//**
 
 \class FrequencyPlotDialog
-\brief Displays a spectrum plot of the waveform.  Has options for
+\brief Displays a spectrum plot of the waveform. Has options for
 selecting parameters of the plot.
 
 Has a feature that finds peaks and reports their value as you move
@@ -18,7 +18,7 @@ the mouse around.
 *//****************************************************************//**
 
 \class FreqPlot
-\brief Works with FrequencyPlotDialog to dsplay a spectrum plot of the waveform.
+\brief Works with FrequencyPlotDialog to display a spectrum plot of the waveform.
 This class actually does the graph display.
 
 Has a feature that finds peaks and reports their value as you move
@@ -30,8 +30,6 @@ the mouse around.
   Salvo Ventura - November 2006
   Extended range check for additional FFT windows
 */
-
-
 
 #include "FreqWindow.h"
 
@@ -113,7 +111,6 @@ enum {
 #define FREQ_WINDOW_WIDTH 480
 #define FREQ_WINDOW_HEIGHT 330
 
-
 static const char * ZoomIn[] = {
 "16 16 6 1",
 " 	c None",
@@ -138,7 +135,6 @@ static const char * ZoomIn[] = {
 "+++@@           ",
 "@+@@            ",
 " @@             "};
-
 
 static const char * ZoomOut[] = {
 "16 16 6 1",
@@ -243,6 +239,7 @@ void FrequencyPlotDialog::Populate()
       Verbatim( "16384" ) ,
       Verbatim( "32768" ) ,
       Verbatim( "65536" ) ,
+      Verbatim( "131072" ) ,
    };
 
    TranslatableStrings funcChoices;
@@ -400,6 +397,7 @@ void FrequencyPlotDialog::Populate()
       // -------------------------------------------------------------------
       // ROW 3: Spacer
       // -------------------------------------------------------------------
+
       S.AddSpace(5);
       S.AddSpace(5);
       S.AddSpace(5);
@@ -469,7 +467,6 @@ void FrequencyPlotDialog::Populate()
       mExportButton = S.Id(FreqExportButtonID).AddButton(XXO("&Export..."));
 
       S.AddSpace(5);
-
 
       // ----------------------------------------------------------------
       // ROW 7: Function, Axix, Grids, Close
--- a/src/SpectrumAnalyst.cpp
+++ b/src/SpectrumAnalyst.cpp
@@ -22,8 +22,6 @@ and in the spectrogram spectral selection.
   Extended range check for additional FFT windows
 */
 
-
-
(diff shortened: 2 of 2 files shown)


SteveDaulton:
If this enhancement is used, there should probably be a similar enhancement to the Filter Curve EQ effect so that users can make practical use of the improved frequency analysis. However, note also: https://github.com/audacity/audacity/issues/1596

rbdannenberg:
Does anyone understand the numerical properties of the FFT implementation in use now? Mathematically, the FT and FFT are invertible, but in practice, especially using 32-bit floats, error can be significant.  E.g. it is common to average multiple FFTs rather than compute one larger FFT to estimate a spectral average. For this application, where output resolution might ultimately be limited by pixel size in a plot, it is probably OK, but where's the analysis? Would downsampling and using a shorter FFT accomplish the same goal?

witwald:
When performing an FFT analysis, it is common practice to use averaging to reduce the variance of the spectral estimates corresponding to each FFT bin. To be clear, we don't average the FFT as such, but the power spectrum, as we are considering the energy content of the signal. When displaying the spectrum using a logarithmic frequency scale, the low frequency end will not be particularly limited by pixel size. When analysing music signals, amongst many others, it is quite usual to use a logarithmic frequency scale in order to better display the energy content over many octaves. I don't envisage use of a 131072-point FFT  causing any numerical problems. Looking at the change in the displayed spectrum when going from a 65536-point to a 131072-point FFT shows the expected increase in resolution. In some of my test cases it became possible to resolve finer details than what were originally showing up in in the "smoothed" results presented the coarser FFT. When I've analysed some CD music tracks, it is surprising to see some quite well defined low-frequency energy at frequencies in the 5Hz to 10Hz range (from air conditioning noise?). Having more FFT points allows those low frequencies to be studied with greater resolving power should there be any distinct peaks present that show up after averaging many individual spectra in the time history that is being analysed.

rbdannenberg:
I think your answer is basically that you do not understand the numerical properties of the FFT implementation in use now and are more concerned with frequency resolution than numerical accuracy. I decided to look for myself and found SIAM J. Sci. COMPUT. Vol. 17, No. 5, pp. 1150-1166, September 1996 "ACCURACY OF THE DISCRETE FOURIER TRANSFORM AND THE FAST FOURIER TRANSFORM by SCHATZMAN. Figure 1 shows relative RMS at 128K and 32-bit float on RS6000 library to be about 0.01 or about 40dB SNR (!), but NCAR FFTPAK is between 10E-5 and 10E-6, or >100dB SNR. We do show power on a dB scale, so -40dB errors could be visible. I suspect our FFT is more accurate than the RS6K library that showed the large errors, so while FFT is generally accurate, error does grow with FFT size. A lot of people use data from Audacity in important work, including exporting spectral data. We don't want to produce sub-standard data.

witwald:
I did read the comments in the FFT code regarding potential loss of accuracy with longer FFTs. That's the reason I performed some testing, and I did not see any indication of accuracy problems. I fully take on board the comment that we need to maintain the SNR of the FFT when performing work. In fact, I am one of those people who would like to get accurate results. One important aspect is having appropriate frequency resolution available for the job at hand, as without it, it is not possible to do important work that involves accuracy of frequency estimates. FFTs that are too low in resolution are producing sub-par data, for me anyway.

As a further test, I created a 8.056640625 Hz sine wave with an amplitude of 1. As the signal is periodic in the rectangular window, there shouldn't be any side lobes. There weren't. The amplitude was 0.000000 dB at the peak, as expected, and the neighbouring FFT bins had values of -115.550842 dB and -115.155586 dB.

below is a representative excerpt from the exported data. I visually scanned it all, and I saw no evidence of any accuracy errors of the "–40 dB" type at all. The code appears to be working entirely as expected.

Frequency (Hz)	Level (dB)
0.366211 -163.002609
0.732422 -156.867752
1.098633 -153.317062
1.464844 -150.691299
1.831055 -148.582672
2.197266 -146.879913
2.563477 -145.193710
2.929688 -143.731705
3.295898 -142.347076
3.662109 -141.596878
4.028320 -139.696167
4.394531 -138.371826
4.760742 -137.013229
5.126953 -135.580856
5.493164 -134.070038
5.859375 -132.402420
6.225586 -130.517593
6.591797 -128.309143
6.958008 -125.548882
7.324219 -121.790947
7.690430 -115.550842
8.056641 0.000000
8.422852 -115.155586
8.789062 -120.997429
9.155273 -124.351471
9.521484 -126.708344
9.887695 -128.481918
10.253906 -129.925980
10.620117 -131.131287
10.986328 -132.160721
11.352539 -133.069000
11.718750 -133.871490
12.084961 -134.590683
12.451172 -135.249908
12.817383 -135.840912

Here is a plot of the FFT spectrum:
image

rbdannenberg:
Thanks for the reply. I don't know how to properly test an FFT implementation, but at least it passes a sanity check with flying colors.

witwald:
Are there any suggestions that anyone can offer as to how to progress this pull request? Would someone like to be a reviewer? Thanks for your input.

SteveDaulton:
Back in 2013, I submitted a patch to increase Plot Spectrum maximum size from 16384 to 65536. Choosing a maximum size will always be a matter of compromise. I was able to put forward an argument that was sufficiently compelling that the patch was accepted. I don't yet see a compelling argument for increasing it to 131072 (and not to say 524288 or 4194304).

Paul-Licameli:
Since @rbdannenberg approved the sanity check of data with 131072, I will merge this simple change.  Thank you.

Paul-Licameli (approved):

PR #1746, opened 2021-09-26
Increase max selection size for Plot Spectrum

witwald:
Resolves: Limitation on spectrum analysis due to small max selection size

It is worthwhile increasing the maximum size of the block of selected samples in Plot Spectrum. These proposed code changes now enable 46.60 minutes of a 48kHz sampled signal to be analyzed. The previous limit often prevented the analysis of a complete song.

In this proposal, the dynamically allocated memory for storing the selection is now being freed when the Plot Spectrum dialog is exited. This is better housekeeping, which is more important now that we can have a much larger selection.

A test track with 55 minutes of data at 48kHz sampling rate was created, and this triggered the appropriate warning message. The computed spectrum appeared to be correct.
 
Recommended:

### DIFF
--- a/src/FreqWindow.cpp
+++ b/src/FreqWindow.cpp
@@ -594,13 +594,16 @@ void FrequencyPlotDialog::GetAudio()
          auto start = track->TimeToLongSamples(selectedRegion.t0());
          auto end = track->TimeToLongSamples(selectedRegion.t1());
          auto dataLen = end - start;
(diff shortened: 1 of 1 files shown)


witwald:
Are there any suggestions that anyone can offer as to how to progress this pull request? Would someone be able to do a review? Thanks for your input.

Paul-Licameli (approved):
Looks good to me!

PR #294, opened 2018-08-25
Adjusted Hanning window to ensure it is periodic on observation interval

witwald:
When creating window functions for use with FFT analysis, it is necessary to ensure that the functions are periodic in the observation interval. In order to better visualise this, consider a short observation interval that consists of 8 samples. The corresponding window coefficients have been computed and are stored in the array in[] = in[0]...in[7]:
```
  Location:   in[0]  in[1]  in[2]  in[3]  in[4]  in[5]  in[6]  in[7]
     Value:    0.0    0.3    0.6    0.9    1.0    0.9    0.6    0.3
```
In order to ensure that Fourier periodicity is maintained, i.e. the windows coefficients are periodic in the observation interval, it is necessary to have performed the window computations so that:
```
   in[7] = in[1]
```
If the window coefficients do not have Fourier periodicity, then there will be errors in the computation of the Fourier coefficients. A simple test case is to create a 1024 point sinusoidal waveform, using a sampling frequency of exactly 1024 Hz. Choose a frequency of 128 Hz for this sine wave, and the waveform will be periodic in the FFT block. Taking the FFT of this signal will produce a single spike located at FFT bin 128, where FFT bin 0 is the DC component. If a correctly-formed Hanning window is
applied to this sinusoidal waveform, then there will be exactly three FFT that contain non-zero data values. For our example, the bins in question are bin numbers 127, 128, and 129. Furthermore, |bin 127| = |bin 129| = |bin 128|/2 (Harris, 1978).

Some useful references for windowing functions are as follows:

1. Fredric Harris. On the Use of Windows for Harmonic Analysis with
   Discrete Fourier Transforms. Proceedings IEEE, Vol 66, pp 51-83,
   January 1978.

2. Svend Gade and Henrik Herlufsen. Windows to FFT Analysis: Use of
   Weighting Functions in DFT/FFT Analysis (Part I and Part II).
   B&K Technical Review, Issue 3 and 4, 1987.

PR #1747, opened 2021-09-26
Modify cursor colour in Plot Spectrum to make it more visible

witwald:
Resolves: Cursor in Plot Spectrum difficult to see

The colour of the cursor in Plot Spectrum has been modified to make it more visible. It is now red in colour, making it much easier to see, especially when the cursor is close to or on top of vertical grid lines.

Recommended:

### DIFF
--- a/src/FreqWindow.cpp
+++ b/src/FreqWindow.cpp
@@ -931,7 +931,7 @@ void FrequencyPlotDialog::PlotPaint(wxPaintEvent & event)
       else
          px = (int)((bestpeak - xMin) * width / (xMax - xMin));
 
-      dc.SetPen(wxPen(wxColour(160,160,160), 1, wxPENSTYLE_SOLID));
+      dc.SetPen(wxPen(wxColour(255, 32, 32), 1, wxPENSTYLE_SOLID));
       AColor::Line(dc, r.x + 1 + px, r.y, r.x + 1 + px, r.y + r.height);
 
        // print out info about the cursor location


witwald:
Are there any suggestions that anyone can offer as to how to progress this pull request? Thanks for your input.

SteveDaulton:
> Are there any suggestions that anyone can offer as to how to progress this pull request? Thanks for your input.

How about just merging it?
It looks totally safe to me, and it does what it says. The "cursor" (vertical line at the closest peak to the mouse pointer) is almost invisible on my (Linux) machine with the current grey "cursor", and clearly visible when red. 

I don't like to rely on colour, as doing so can cause problems for colour blind users, but in this case the change from grey to red is no worse for colour blind users, and much more visible for the majority of users.

witwald:
I'm not sure how to action a merge at this stage. My screen says "Merging is blocked".
image

SteveDaulton:
> I'm not sure how to action a merge at this stage. My screen says "Merging is blocked".

You or I can't merge it, which is why I requested a review from two developers that can.
I'm guessing that the comment by AnitaBats means that  it is earmarked for "Sprint 7 - Release 3.1" (whatever that means).

LWinterberg:
Hello @witwald! I'm unfortunately not qualified to judge whether or not code is any good, but your contributions have been noted and their review (and probably: addition) is scheduled for next sprint, which should start either next week or the week after. I'm sorry this is taking so long, but I think you still should make it into the 3.1 release.

Paul-Licameli:
Hello Witwald, and welcome back to the conversation.  We have talked before: https://github.com/audacity/audacity/pull/294

Paul-Licameli:
The color doesn’t use the theming system — but neither did the old code, so the new is not worse on that score.

I’ll merge this.

I also mean to catch up in the other recent conversations you have opened.

I hope you have also noticed the recent development of a spectral editing brush tool and might be interested in testing it.

Penikov (commented):
Nice )

PR #1748, opened 2021-09-26
Increased resolution of frequency and dB readouts when plotting spectrum

witwald:
Resolves: Improves the limited resolution of dB and Hz readouts when plotting a spectrum

The dB and Hz readouts when plotting a spectrum and using the cursor are now displayed to 2 decimal places. This improves their usefulness, particularly when analysing low frequency audio signals.

Recommended:

### DIFF
--- a/src/FreqWindow.cpp
+++ b/src/FreqWindow.cpp
@@ -914,6 +914,7 @@ void FrequencyPlotDialog::PlotPaint(wxPaintEvent & event)
    }
 
    float xPos = xMin;
+   float yPos;
 
    // Find the peak nearest the cursor and plot it
    if ( r.Contains(mMouseX, mMouseY) & (mMouseX!=0) & (mMouseX!=r.width-1) ) {
@@ -934,7 +935,7 @@ void FrequencyPlotDialog::PlotPaint(wxPaintEvent & event)
       dc.SetPen(wxPen(wxColour(160,160,160), 1, wxPENSTYLE_SOLID));
       AColor::Line(dc, r.x + 1 + px, r.y, r.x + 1 + px, r.y + r.height);
 
-       // print out info about the cursor location
+      // Print out info about the cursor location
 
       float value;
 
@@ -950,14 +951,21 @@ void FrequencyPlotDialog::PlotPaint(wxPaintEvent & event)
       TranslatableString peak;
 
       if (mAlg == SpectrumAnalyst::Spectrum) {
+         // Determine the position of the cursor in the plot window.
+         float yRange = mYMax - mYMin;
+         float yTotal = yRange * (float)mZoomSlider->GetValue() / 100.0f;
+         int sTotal = yTotal * 100;
+         int sRange = yRange * 100;
+         int sPos = mPanScroller->GetThumbPosition() + ((mPanScroller->GetThumbSize() - sTotal) / 2);
+         float yMax = mYMax - (float)sPos / 100.0f;
+         float yValue = yMax - float(mMouseY - (r.y + 1)) / (float)r.height * yTotal;
+         // Determine the note corresponding to each frequency.
          auto xp = PitchName_Absolute(FreqToMIDInote(xPos));
          auto pp = PitchName_Absolute(FreqToMIDInote(bestpeak));
-         /* i18n-hint: The %d's are replaced by numbers, the %s by musical notes, e.g. A#*/
-         cursor = XO("%d Hz (%s) = %d dB")
-            .Format( (int)(xPos + 0.5), xp, (int)(value + 0.5));
-         /* i18n-hint: The %d's are replaced by numbers, the %s by musical notes, e.g. A#*/
-         peak = XO("%d Hz (%s) = %.1f dB")
-            .Format( (int)(bestpeak + 0.5), pp, bestValue );
+         /* i18n-hint: The %f's are replaced by numbers, the %s by musical notes, e.g. A# */
+         cursor = XO("%.2f Hz (%s) = %.2f dB").Format(xPos, xp, yValue);
+         /* i18n-hint: The %f's are replaced by numbers, the %s by musical notes, e.g. A# */
+         peak = XO("%.2f Hz (%s) = %.2f dB").Format(bestpeak, pp, bestValue);
       } else if (xPos > 0.0 && bestpeak > 0.0) {
          auto xp = PitchName_Absolute(FreqToMIDInote(1.0 / xPos));
          auto pp = PitchName_Absolute(FreqToMIDInote(1.0 / bestpeak));
@@ -977,7 +985,6 @@ void FrequencyPlotDialog::PlotPaint(wxPaintEvent & event)
       mCursorText->SetValue(wxT(""));
       mPeakText->SetValue(wxT(""));
    }
-
 
    // Outline the graph
    dc.SetPen(*wxBLACK_PEN);
--- a/src/FreqWindow.cpp
+++ b/src/FreqWindow.cpp
@@ -914,7 +914,6 @@ void FrequencyPlotDialog::PlotPaint(wxPaintEvent & event)
    }
 
    float xPos = xMin;
-   float yPos;
 
    // Find the peak nearest the cursor and plot it
    if ( r.Contains(mMouseX, mMouseY) & (mMouseX!=0) & (mMouseX!=r.width-1) ) {
@@ -951,19 +950,11 @@ void FrequencyPlotDialog::PlotPaint(wxPaintEvent & event)
       TranslatableString peak;
 
       if (mAlg == SpectrumAnalyst::Spectrum) {
-         // Determine the position of the cursor in the plot window.
-         float yRange = mYMax - mYMin;
-         float yTotal = yRange * (float)mZoomSlider->GetValue() / 100.0f;
-         int sTotal = yTotal * 100;
-         int sRange = yRange * 100;
-         int sPos = mPanScroller->GetThumbPosition() + ((mPanScroller->GetThumbSize() - sTotal) / 2);
-         float yMax = mYMax - (float)sPos / 100.0f;
-         float yValue = yMax - float(mMouseY - (r.y + 1)) / (float)r.height * yTotal;
          // Determine the note corresponding to each frequency.
          auto xp = PitchName_Absolute(FreqToMIDInote(xPos));
          auto pp = PitchName_Absolute(FreqToMIDInote(bestpeak));
          /* i18n-hint: The %f's are replaced by numbers, the %s by musical notes, e.g. A# */
-         cursor = XO("%.2f Hz (%s) = %.2f dB").Format(xPos, xp, yValue);
+         cursor = XO("%.2f Hz (%s) = %.2f dB").Format(xPos, xp, value);
          /* i18n-hint: The %f's are replaced by numbers, the %s by musical notes, e.g. A# */
          peak = XO("%.2f Hz (%s) = %.2f dB").Format(bestpeak, pp, bestValue);
       } else if (xPos > 0.0 && bestpeak > 0.0) {


SteveDaulton:
How accurate are the dB and Hz measurements? We don't want to mislead users  into thinking that the measurements are more accurate than they really  are.

witwald:
The dB measurements and the Hz measurements can be accurate to fractions of a Hz and fractions of a dB. For example, analysing a sine wave that is periodic in the sampling block will produce a signal with two side lobes that are exactly –6.02 dB relative to the amplitude at the signal frequency. This exactly matches the mathematically derived result using signal processing theory. The frequency bins may need to be accurately determined for post processing, and they are clearly defined to a better than integer precision over much of the frequency range. For some test signals, the amplitude of the sine wave of amplitude 1 comes out as 0.00000 dB in the exported file, exactly expected. The frequencies of the FFT bins are simply multiples of (sampling rate)/(number of FFT points), and so are exactly determined from a simple exact formula. Outputting the results to 2 decimal places will often not display the full available precision.

SteveDaulton:
What about frequencies that lie between frequency bins (and for real world  sounds there will be many)?

A very simple example:
Generate a 0.5 amplitude sine wave at 432 Hz.
Plot Spectrum, with Size = 256
Displayed result peak: 428 Hz, -5.1 dB

Two decimal places would be misleading in this case.

witwald:
It wouldn't be misleading. It would be more accurate. The rest of the energy in that original sine wave has leaked into the other frequency bins. The FFT calculation yields the correct answer, it's just not being displayed as fully as it could be. If you set up an appropriate test signal, the answers can be displayed to 2 decimal places in the dB, which tells the user that things are being computed correctly. For example, some signals will produce side lobes that are 6.02 dB below the main signal component, as the Hann window produces three Fourier coefficients whose amplitudes are 1/4, 1/2 and 1/4, with the 1/2 corresponding to the frequency of the signal. The computations are correct, its just the display of the answers that lacks a bit of precision, which comes in handy sometimes.

SteveDaulton:
> It wouldn't be misleading. It would be more accurate.

I expect that most users would interpret the highlighted result below to mean that the frequency peak is around 442.2 Hz, and around 5.4 dB:

spectrum

The actual frequency being analyzed is 432 Hz with a peak amplitude of -6 dB.
The displayed peak frequency has an error of more than 10 Hz, and an amplitude error of more than 0.5 dB, which imo is **much** more than expected from a 2 decimal place measurement.

rbdannenberg:
There are 2 things being discussed here. One is that the short-time FT is not entirely intuitive, as illustrated by frequencies that are not harmonics of the 1/window-size. These inharmonic frequencies, or frequencies other than the bin center frequencies, are represented in the STFT as the sum of multiple harmonics. Windowing, although generally a good thing, also introduces artifacts. The second thing is the numerical accuracy of the FFT implementation. This is entirely different. It makes me nervous to hear claims that "things are being computed correctly" -- computing correctly does not mean there are no numerical errors. (I'll repeat that I suspect these errors are not a big issue; it would be nice to do some evaluation though, especially as FFT sizes are getting bigger, where error is expected to grow.)

witwald:
> It makes me nervous to hear claims that "things are being computed correctly" -- computing correctly does not mean there are no numerical errors. 

We are not talking about an analytical answer. It would be entirely possible to say exactly the same thing about any function within Audacity, as it relies on the underlying precision of the arithmetic, at the very least. When I say "computing correctly", I mean the implementation is correct to a reasonable level of precision, as defined by single-precision or double-precision arithmetic, without any surprising errors or numerical instabilities. Why is outputting the numbers to greater number of decimal places seem to have become such an issue. It is possible to demonstrate that the existing implementation will output 6.020600 dB as 6 dB, instead of my proposed more accurate value of 6.02 dB. Yet it seems there may be a claim being made that 6 is more accurate than 6.02 in this instance? I am confused by this.

> (I'll repeat that I suspect these errors are not a big issue; it would be nice to do some evaluation though, especially as FFT sizes are getting bigger, where error is expected to grow.)

Maybe this will help give you more confidence in the results, which certainly seem to be quite accurate when compared to analytical solutions. The paper that you referred to earlier states that: "Our own conclusions concerning accuracy are that the FFT is remarkably stable, when implemented properly. The FFT is vastly more accurate than the slow DFT."

I have done some tests, and shown the results. Computational noise of –120 dB re a signal amplitude of 1 does not seem to be an issue. Theoretically, we might expect 20*log10(1e-07) = –140 dB, which is what we are getting as in terms of spectral leakage into some of the FFT bins. That seems entirely reasonable, and not swamped by any obvious errors. The FFT implementation seems quite good, and certainly gives accurate results in the tests that I have run, to the limits of the analytical answers  when represented by 32-bit floating point numbers.

witwald:
> I expect that most users would interpret the highlighted result below to mean that the frequency peak is around 442.2 Hz, and around 5.4 dB:
> 
> spectrum
> 
> The actual frequency being analyzed is 432 Hz with a peak amplitude of -6 dB. The displayed peak frequency has an error of more than 10 Hz, and an amplitude error of more than 0.5 dB, which imo is **much** more than expected from a 2 decimal place measurement.

This is not a problem in the display of the "computed" value to 2 decimal places. Keep in mind that the plotting does quite a bit of processing behind the scenes.

I went and duplicated you sine wave test. Sine wave amplitude of 0.5, frequency 432 Hz, sampling rate 44.1 kHz, FFT size 1024 points, which I assume is correct? There is absolutely nothing wrong with using 2 decimal places to display the results. The interpretation being made of those results does not change the fact that the actual values can be **displayed more accurately**. 

Here are the computed spectral values, as exported:

```
Frequency (Hz)	Level (dB)
0.000000	-96.566528
43.066406	-95.441795
86.132812	-92.652672
129.199219	-88.954201
172.265625	-84.580986
215.332031	-79.421295
258.398438	-73.125145
301.464844	-64.937370
344.531250	-52.794247
387.597656	-12.440313
430.664062	-6.025982
473.730469	-11.633661
516.796875	-50.712643
559.863281	-62.710781
602.929688	-70.506653
```

As can be seen, the spectral resolution is 43.07 Hz, so energy associated with that 432 Hz sine wave falls in the FFT bin centred on a frequency of 430.66 Hz. There is also some leakage into the adjacent frequency bins due to the Hann window that was used. Because the Hann window "spreads" the energy in the single-tone signal, we get an answer that requires interpretation in order to be made sense of.

If I use a rectangular window instead, I get:
image

The spectral values are:

```
Frequency (Hz)	Level (dB)
0.000000	-53.334827
43.066406	-53.202808
86.132812	-52.804195
129.199219	-52.129616
172.265625	-51.157543
215.332031	-49.843521
258.398438	-48.098545
301.464844	-45.737946
344.531250	-42.326851
387.597656	-36.466614
430.664062	-6.034704
473.730469	-35.907745
516.796875	-42.030930
559.863281	-45.552238
602.929688	-48.020317
```

We can see that the side lobes have dropped in value quite a bit, which is expected, as the sine wave falls very close in frequency to that of its FFT bin.

What I believe that I've demonstrated is that displaying the dB and Hz values to 2 decimal places will not cause problems. 

But what you have discovered is that "peak picking" jumps to the wrong peak, which is something that I too have noticed. Of course, I think that you'll agree that that has nothing to do with displaying results to 2 decimal places. In fact, the Hann window has produced a smooth fall-off either side of the peak by approximately 6 dB, but the plot display shows 2 distinct peaks instead of one! See below for a zoomed in view:
image

SteveDaulton:
> This is not a problem in the display of the "computed" value to 2 decimal places.

It's not my decision, but I disagree.

Taking a step back:

If I was told that the distance between two geographic locations was 23 km, then it would be reasonable for me to assume that the actual distance is about 23 km - it may be 22.6 km or 23.4 km, but I would expect it to be close to 23 km.

If I was told that the distance between two locations was 23434 m, then it would be reasonable for me to expect that the actual distance will be close to 23434 m - it may actually be 23433 m or 23435, but I certainly would not expect it to be 22600 m.

Showing the user a figure with two decimal places suggests that the measurement is accurate to around two decimal places. As shown in my example, the actual amplitude error is more than 0.5 dB. It is misleading to suggest that it is accurate to two decimal places when in fact it may not even be accurate to +/- 0.5 dB.

rbdannenberg:
I have to agree with witwald here -- Steve's complaint is that the spectrum does not directly reflect the actual amplitude of an arbitrary sinusoid (which in the short time of an FFT window, is not really defined -- windowing is a form of modulation, and we know that modulation changes frequency content), just as it does not reflect the exact frequency. It's a valid concern that precision of the FFT can mislead users to think that they have more precise knowledge of the underlying signal, but we should weigh this against the value of precision for those who understand the numbers more deeply.

SteveDaulton:
Those that understand the numbers more deeply can refer to the exported data. That really is more accurate.
I'm not convinced that the interpolated data is accurate to 2 decimal places.

rbdannenberg:
Yes, Steve, I agree. I was confused by the conversation referring to accuracy of FFT and "computed values", but as you point out, the displayed value in question is an estimate. Do we even know the algorithm? There's a lot of literature on interpolation, and it's pretty confusing. A common approach is cubic interpolation from the 3 nearest bins, and there's some interesting math why this is a reasonable interpolation strategy, but other papers claiming it can be improved upon. In any case, I agree the exported FFT numbers are well understood (except for the error bounds), probably accurate, and useful at higher resolution. On the other hand, the *estimated* peak in the display is not easy to put error bounds on, and I don't see any advantage to printing lots of digits.

SteveDaulton:
> Do we even know the algorithm?

When the bins are more than 1px apart, it uses cubic interpolation, which I agree is reasonable even though imperfect.

Something that does look weird to me is that when the bins are less than 1px apart, the drawn amplitude can be *much* lower than the reported amplitude, which seems to be due to averaging of the bin values. This can make it impossible to see narrow peaks, even though they are clearly shown in the numeric data. That is a problem in cases such as trying to locate high frequency interference (which is not  uncommon with USB microphones).
On the other hand, very narrow frequency notches can be clearly seen in the graph.
Personally I would  have opted for clearly visible peaks, as I believe that is the more common use case.

witwald:
>Personally I would have opted for clearly visible peaks, as I believe that is the more common use case.
I would tend to agree with that. I would like to look at the plotting of the spectrum in the future, with a view to potentially addressing some aspects of the display processing.

I personally have found some aspects of the spectrum plotting to get in the way of interpreting the measurements. That was one impetus for increasing the displayed numeric output to 2 decimal places, as it makes inspection of results easier in many situations. That is the reason behind the plot – to not have to go and look at the data in the exported spectrum file.

SteveDaulton:
A bit off-topic, but it would also be nice to have a sane way to manually set the y axis zoom and scroll. The automatic zoom can be very unhelpful at times.

Paul-Licameli:
> > I expect that most users would interpret the highlighted result below to mean that the frequency peak is around 442.2 Hz, and around 5.4 dB:
> > spectrum
> > The actual frequency being analyzed is 432 Hz with a peak amplitude of -6 dB. The displayed peak frequency has an error of more than 10 Hz, and an amplitude error of more than 0.5 dB, which imo is **much** more than expected from a 2 decimal place measurement.
> 
> This is not a problem in the display of the "computed" value to 2 decimal places. Keep in mind that the plotting does quite a bit of processing behind the scenes.
> 
> I went and duplicated your sine wave test. Sine wave amplitude of 0.5, frequency 432 Hz, sampling rate 44.1 kHz, FFT size 1024 points, which I assume is correct? There is absolutely nothing wrong with using 2 decimal places to display the results. The interpretation being made of those results does not change the fact that the actual values can be **displayed more accurately**.
> 
> Here are the computed spectral values, as exported:
> 
> ```
> Frequency (Hz)	Level (dB)
> 0.000000	-96.566528
> 43.066406	-95.441795
> 86.132812	-92.652672
> 129.199219	-88.954201
> 172.265625	-84.580986
> 215.332031	-79.421295
> 258.398438	-73.125145
> 301.464844	-64.937370
> 344.531250	-52.794247
> 387.597656	-12.440313
> 430.664062	-6.025982
> 473.730469	-11.633661
> 516.796875	-50.712643
> 559.863281	-62.710781
> 602.929688	-70.506653
> ```
> 
> As can be seen, the spectral resolution is 43.07 Hz, so energy associated with that 432 Hz sine wave falls in the FFT bin centred on a frequency of 430.66 Hz. There is also some leakage into the adjacent frequency bins due to the Hann window that was used. Because the Hann window "spreads" the energy in the single-tone signal, we get an answer that requires interpretation in order to be made sense of.
> 
> If I use a rectangular window instead, I get:
> 
> image
> 
> The spectral values are:
> 
> ```
> Frequency (Hz)	Level (dB)
> 0.000000	-53.334827
> 43.066406	-53.202808
> 86.132812	-52.804195
> 129.199219	-52.129616
> 172.265625	-51.157543
> 215.332031	-49.843521
> 258.398438	-48.098545
> 301.464844	-45.737946
> 344.531250	-42.326851
> 387.597656	-36.466614
> 430.664062	-6.034704
> 473.730469	-35.907745
> 516.796875	-42.030930
> 559.863281	-45.552238
> 602.929688	-48.020317
> ```
> 
> We can see that the side lobes have dropped in value quite a bit, which is expected, as the sine wave falls very close in frequency to that of its FFT bin.
> 
> What I believe that I've demonstrated is that displaying the dB and Hz values to 2 decimal places will not cause problems.
> 
> But what you have discovered is that the "peak picking" jumps to a false peak, which is something that I too have noticed, but have **deliberately** chosen to ignore for the time being, as it's another issue entirely. Of course, I think that you'll agree that that has nothing to do with displaying results to 2 decimal places.
> 
> From an inspection of the spectrum values, it can in fact be seen that the Hann window has produced a smooth fall-off either side of the peak, with the values dropping by approximately 6 dB, **but the plot display shows 2 distinct peaks instead of one**! See below for a zoomed in view:
> 
> image

The double peak in the lower picture is a weird thing I would invite you to debug, and also the repetition of -6 dB in the scale.

PR #1753, opened 2021-09-26
Theme system rewrite part 1

Paul-Licameli:
Resolves: 

* fixes a minor bug with theme colors
* some tidying of code
* some elimination of unused code and files
* open-endedness of the set of predefined themes, so that a module might add one; not a hard-coded list of four

Recommended:


JamesCrook (approved):
This is clearly successfully moving things in the direction you want them to go.  The enums are slightly undercommented for 'big picture', but that was so before, so not a fault of this PR.  So I approve this PR.

I think the code changes are premature without a clear design plan for how to handle both 'double open endedness' and  'migration to Qt'.  The theming is one feature where you'd surely want to use 'the same system' when a Qt version is made.  The 'double open endedness' is a generic problem of plug-ins, not just of theming, in my view, so it's worth thinking about the problem as a generic problem, even if solved as a special case.

Paul-Licameli:
Thank you for approval, but I see I have to fix build failures first.  I did build and run on Mac and verify that theme switching works as before.

So, no objections about all the deleted things.

I see all the more graphical uses of Qt as a more distant problem.  I have an implementation strategy for Qt migration that must first go through a “boring” phase of migrating menus, dialogs, command manager, the macro capability, etc.  Even so, if we get to swapping Qt bitmap classes for wxWidgets equivalents, I see that as the lesser problem for theming, and independent of the problems of larger scale structuring for extensibility.

PR #1756, opened 2021-09-27
Loop play button

Paul-Licameli:
Resolves: #1755 

Add a new toggle button to the Control toolbar.

Recommended:


src/AudacityApp.cpp
```diff
@@ -384,6 +384,11 @@ void PopulatePreferences()
       gPrefs->Write(wxT("/GUI/Toolbars/Time/Show"),1);
    }
 
+   if (std::pair{ vMajor, vMinor } < std::pair{ 3, 1 } ) {
```
gera-gas, line 387: I am understood what you do, but I am not understand why we should to do this?

Paul-Licameli: Run Audacity 3.0.4 or earlier and exit.  It makes audacity.cfg which remembers preferences and other settings.

Then run my branch, but without the line above.

You will see that the control toolbar is too narrow to show the new button completely.

This line fixes that.

src/toolbars/ControlToolBar.cpp
```diff
@@ -428,6 +438,9 @@ void ControlToolBar::ReCreateButtons()
       mRecord->PushDown();
    }
 
+   if (loopDown)
```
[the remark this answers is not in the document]
Paul-Licameli: "Check on recording" and do what, then?

No.  The point is to reproduce the up/down state of the button as it was before the button was destroyed and recreated.

Recording affects the enabled state of the button -- which is separate from the up/down state.

gera-gas (approved):

Issue #1751, opened 2021-09-26
Mac: Latest 3.1.0 alpha fails to see my installed FFmpeg library


petersampsonaudacity:
**Describe the bug**
The latest 3.1.0 alpha  Audacity_macOS_Intel_1270895276_97ca83e  fails to see my installed FFmpeg library

Launching 3.0.4 and importing *.MP4 works fine still
Thus this is a regression on 3.0.4

**To Reproduce**
1. Install latest alpha  Audacity_macOS_Intel_1270895276_97ca83e  on Mac
2. Launch
3. Try to import an MP4 audio file
4. Observe: Error message re FFmpeg

**Expected behavior**
a) no warning message
b) *.MP4 files imported correctly

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/134815226-223470e7-5b2a-4be5-a6db-fea3ae844e26.png)

**Additional information (please complete the following information):**
 - OS: macOS 11.5.2 Big Sur
 - Version Audacity 3.1.0

**Additional context**
This does not happen with audacity-win-3.1.0-alpha-20210924+97ca83e-64bit on my W10 laptop.

Note that there appears to be no error message re missing FFmpeg when launching at Step 2
**UPDATE:**  on subsequent launch I did get an "FFmpeg startup failed" message.

petersampsonaudacity:
Note the white text on black for the error message - see Issue #1752

PR #1758, opened 2021-09-27
Fix for FFmpeg loading on macOS

crsib:
Resolves: #1751 

On macOS - the dynamic library loader behaves differently.
All DYLD_* environment variables are read only during the application startup. This PR"emulates" the behavior of other OS by manually trying to load every library from every directory listed.

Recommended:

### DIFF
--- a/libraries/lib-ffmpeg-support/FFmpegFunctions.cpp
+++ b/libraries/lib-ffmpeg-support/FFmpegFunctions.cpp
@@ -163,21 +163,15 @@ struct FFmpegFunctions::Private final
       if (path.empty())
          return nullptr;
 
-      std::shared_ptr<wxDynamicLibrary> library =
-         std::make_shared<wxDynamicLibrary>();
-
-      if (!library->Load(wxFileNameFromPath(path)))
-         return nullptr;
-
-      return library;
+      return LoadLibrary(wxFileNameFromPath(path));
    }
 
    bool Load(FFmpegFunctions& functions, const wxString& path)
    {
       // We start by loading AVFormat
-      AVFormatLibrary = std::make_shared<wxDynamicLibrary>();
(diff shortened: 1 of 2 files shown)


libraries/lib-ffmpeg-support/FFmpegFunctions.cpp
```diff
@@ -226,6 +220,29 @@ struct FFmpegFunctions::Private final
 
       return true;
    }
+
+   std::shared_ptr<wxDynamicLibrary> LoadLibrary(const wxString& libraryName) const
+   {
+#if defined(__WXMAC__)
+      // On macOS dyld reads environment only shen application starts.
```
Paul-Licameli, line 227: "when"

libraries/lib-ffmpeg-support/FFmpegFunctions.cpp
```diff
@@ -163,21 +163,15 @@ struct FFmpegFunctions::Private final
...
          return nullptr;
 
-      std::shared_ptr<wxDynamicLibrary> library =
-         std::make_shared<wxDynamicLibrary>();
-
-      if (!library->Load(wxFileNameFromPath(path)))
-         return nullptr;
-
-      return library;
+      return LoadLibrary(wxFileNameFromPath(path));;
```
Paul-Licameli, line 166: `;;` ?

Paul-Licameli (approved):
Just two minor typos

libraries/lib-ffmpeg-support/FFmpegFunctions.cpp
crsib: Whoa, that was stupid of me. I rely to much on IDE to highlight typos :) CLion failed me here

Issue #1411, opened 2021-08-02
Clip-handles (affordances) make the Time Shift tool redundant


petersampsonaudacity:
**Describe the bug**
The new Clip-handles (affordances) make the Time Shift tool redundant.

It is much easier to use the Clip-handles to reposition clips that to use the Time Shift tool, which involves:
a) changing to the Time Shift tool in the Tools toolbar
b) moving the clip(s)
c) returning to the Tools toolbar to revert to the previous tool - normally the Select tool

**To Reproduce**
1. Use the Clip-handles to reposition clip(s)
2. Observe that it works without changing the currently active tool in Tools toolbar (default Select tool)

**Expected behavior**
Removal of the Time Shift tool

**Screenshots**
Not applicable

**Additional information (please complete the following information):**
 - OS: All
 - Version Audacity 3.0.3/3.1.0
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

**Additional context**
It would be good if the Tools toolbar could be removed altogether with alternative ways to effect
a) envelopes
b) draw repairs
c) and certainly zooming

Removing the Tools toolbar and its associated shortcuts would remove one of the frequently recurring bear-traps in Audacity - whereby users accidentally press one of the shortcut function keys and then find themselves in a tool mode that they don't understand and not knowing how to return to the default (and most useful) Selection tool.

See: https://wiki.audacityteam.org/wiki/ToDo:_Unstick_%27Stuck_In_a_Mode%27#beartrap
>Wiki>Stuck in a Mode - Bear-Traps

Tantacrul:
I know we call this 'affordances' but I've always thought of this as the 'replace timeshift tool job'.

I wholeheartedly support this suggestion and I want to subscribe to the PSampson newletter!

Paul-Licameli:
Before you do this, make sure you understand all the existing behavior, involving sync-lock, and interactions with selection, and so on.

For instance is there a way to move all labels of one track together (as there is to move all clips)?

petersampsonaudacity:
> Before you do this, make sure you understand all the existing behavior, involving sync-lock, and interactions with selection, and so on.
> 
> For instance is there a way to move all labels of one track together (as there is to move all clips)?

Yes, all labels in a track can be moved in sync simply by clicking in the gray space of a label track and dragging left or right.

[...]

PR #1764, opened 2021-09-28
Time shift button gone from Tools toolbar, and...

Paul-Licameli:
... reuse the vacated slot in the theme definition for the brush tool.

That tool is not yet eliminated from the toolbar.  We can see how the brush
image looks in each theme.

Function key strokes reassigned.

Resolves: #1411 
Resolves: #1765

Recommended:


Paul-Licameli:
Closing this.  Instead, #1779 subsumes it now.

PR #1787, opened 2021-09-30
Fix the build on Linux when system libraries are used

crsib:
Fixes a problematic `get_package_interface` which was modifying a global list `LIBRARIES`.

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/AudacityFunctions.cmake
+++ b/cmake-proxies/cmake-modules/AudacityFunctions.cmake
@@ -51,22 +51,29 @@ function( set_dir_folder dir folder)
 endfunction()
 
 # Helper to retrieve the settings returned from pkg_check_modules()
-macro( get_package_interface package )
-   set( INCLUDES
+function( get_package_interface package target )
+   set( package_includes
       ${${package}_INCLUDE_DIRS}
    )
 
-   set( LINKDIRS
+   set( package_linkdirs
       ${${package}_LIBDIR}
    )
 
    # We resolve the full path of each library to ensure the
    # correct one is referenced while linking
+   set( package_libraries )
    foreach( lib ${${package}_LIBRARIES} )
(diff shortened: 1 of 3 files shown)


cmake-proxies/cmake-modules/AudacityFunctions.cmake
```diff
@@ -51,22 +51,28 @@ function( set_dir_folder dir folder)
...
+      find_library( LIB_${lib} ${lib} HINTS ${package_linkdirs} )
+      list( APPEND package_libraries ${LIB_${lib}} )
    endforeach()
-endmacro()
+
+   # And add it to our target
+   target_include_directories( ${target} INTERFACE ${package_includes} )
+   target_link_libraries( ${target} INTERFACE ${package_libraries} )
+
+   message(STATUS "Inteface ${target}:\n\tinclude: ${includes}\n\tLibraries: ${LIBRARIES}")
```
Paul-Licameli, line 74: “Interface”

Paul-Licameli (approved):
Two suggestions

cmake-proxies/cmake-modules/AudacityFunctions.cmake
```diff
@@ -51,22 +51,28 @@ function( set_dir_folder dir folder)
...
       ${${package}_INCLUDE_DIRS}
    )
 
-   set( LINKDIRS
+   set( package_linkdirs
       ${${package}_LIBDIR}
    )
 
    # We resolve the full path of each library to ensure the
    # correct one is referenced while linking
```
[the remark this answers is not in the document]
crsib: Will do!

cmake-proxies/cmake-modules/AudacityFunctions.cmake
crsib: Fixed

Issue #1651, opened 2021-09-09
Generate silence creates a clip with no clip name (in the Clip-handle)


petersampsonaudacity:
**Describe the bug**
Generate silence creates a clip with no clip name (in the Clip-handle),

This is inconsistent with all the other Generate commands

**To Reproduce**
1. Generate Silence
2. Observe: Clip-handle drag-bar affirdance shows no clip name

**Expected behavior**
The clip of Silence should be named

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Windows 10 - but assume all
 - Version Audacity 3.1.0 alpha

**Additional context**
Consistencty in the GUI is important

This has a further deleterious impact on:
 Generate silence at the beginning of a clip creates an inconsistent, unnecessary, split #1653

PR #1793, opened 2021-09-30
Fix auto clip naming

vsverchinsky:
Resolves: #1651 

Also fixes wrong name assigned to a clip when recording to a new track.

Recommended:

### DIFF
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -792,6 +792,16 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
 
    bool appendRecord = !tracks.captureTracks.empty();
 
+   auto makeNewClipName = [&](WaveTrack* track) {
+       for (auto i = 1;; ++i)
+       {
+           //i18n-hint a numerical suffix added to distinguish otherwise like-named clips when new record started
+           auto name = XC("%s #%d", "clip name template").Format(track->GetName(), i).Translation();
+           if (track->FindClipByName(name) == nullptr)
+               return name;
+       }
+   };
+
    {
       if (appendRecord) {
          // Append recording:
@@ -826,16 +836,7 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
             // Less than or equal, not just less than, to ensure a clip boundary.
             // when append recording.
             if (endTime <= t0) {
-               auto newName = [&]() {
-                  for (auto i = 1;; ++i)
-                  {
-                     //i18n-hint a numerical suffix added to distinguish otherwise like-named clips when new record started
-                     auto name = XC("%s #%d", "clip name template").Format(pending->GetName(), i).Translation();
-                     if (pending->FindClipByName(name) == nullptr)
-                        return name;
-                  }
-               }();
-               pending->CreateClip(t0, newName);
+               pending->CreateClip(t0, makeNewClipName(pending.get()));
             }
             transportTracks.captureTracks.push_back(pending);
          }
@@ -909,6 +910,8 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
             else {
                newTrack->SetName(baseTrackName + wxT("_") + nameSuffix);
             }
+            //create a new clip with a proper name before recording is started
+            newTrack->CreateClip(.0, makeNewClipName(newTrack.get()));
 
             TrackList::Get( *p ).RegisterPendingNewTrack( newTrack );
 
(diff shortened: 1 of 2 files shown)


src/ProjectAudioManager.cpp
```diff
@@ -726,6 +727,8 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
             else {
                newTrack->SetName(baseTrackName + wxT("_") + nameSuffix);
             }
+            //create a new clip with a proper name before recording is started
+            newTrack->CreateClip(.0, makeNewClipName(newTrack.get()));
```
Paul-Licameli, line 731: No complaints about the second commit.

src/WaveTrack.cpp
```diff
@@ -48,6 +48,7 @@ from the project that will own the track.
 
 #include "Project.h"
 #include "ProjectRate.h"
+#include "Clipboard.h"
```
Paul-Licameli, line 51: Although it doesn’t make a cycle, this dependency on Clipboard feels wrong to me.

src/WaveTrack.cpp
```diff
@@ -1260,6 +1261,9 @@ void WaveTrack::Paste(double t0, const Track *src)
 {
    bool editClipCanMove = GetEditClipsCanMove();
 
+   auto pastingFromTempTrack = !src->GetOwner();
+   auto pastingFromClipboard = src->GetOwner() && src->GetOwner().get() == &Clipboard::Get().GetTracks();
```
Paul-Licameli, line 1265: There is code here for detection of the origin of the clip.  But can there be another fix that sets the clip names right when they are made?

Could you instead correct the names in the OnCut and OnCopy functions in EditMenus.cpp?

src/WaveTrack.cpp
```diff
@@ -1408,7 +1412,11 @@ void WaveTrack::Paste(double t0, const Track *src)
             newClip->Resample(mRate);
             newClip->Offset(t0);
             newClip->MarkChanged();
-            newClip->SetName(MakeClipCopyName(clip->GetName()));
+            if (pastingFromClipboard)
+               newClip->SetName(MakeClipCopyName(clip->GetName()));
+            else if (pastingFromTempTrack)
+               newClip->SetName(MakeNewClipName());
+            //else { leave clip name as is, track orginates from another track list }
```
Paul-Licameli, line 1419: These changes go beyond fixing the referenced issue.  If there was agreement about new naming, I don’t know where it is.

Paul-Licameli (changes requested):
I’m not satisfied with the first commit.  I think there is an alternative that might need more code but seems more elegant to me.  It avoids too the questionable new #include.

src/WaveTrack.cpp
vsverchinsky: Empty string is valid clip name, so I guess we cannot guess why it's empty

src/WaveTrack.cpp
vsverchinsky: No, the initial design document didn't consider generated clips at all.

src/WaveTrack.cpp
```diff
@@ -1408,7 +1412,11 @@ void WaveTrack::Paste(double t0, const Track *src)
             newClip->Resample(mRate);
             newClip->Offset(t0);
             newClip->MarkChanged();
-            newClip->SetName(MakeClipCopyName(clip->GetName()));
+            if (pastingFromClipboard)
+               newClip->SetName(MakeClipCopyName(clip->GetName()));
+            else if (pastingFromTempTrack)
+               newClip->SetName(MakeNewClipName());
```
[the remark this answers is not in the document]
vsverchinsky: Hm, Silence is not the only generator which produces clips with invalid names, other generators are affected too (but they create clips a bit differently)

src/WaveTrack.cpp
Paul-Licameli: Perhaps then, a clip could store a std::optional<wxString> !  Or otherwise invent a setter method for validity of the name.

src/WaveTrack.cpp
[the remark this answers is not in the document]
Paul-Licameli: How do you generate clips with bad names with other generators?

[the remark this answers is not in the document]
Paul-Licameli: I see, MakeClipCopyName generates names containing "copy" when it should not.

You want that only when pasting clips that really come from the clipboard -- not for clips that are also pasted into a track because effects are implemented that way.

So:  this does go beyond fixing the issue as stated.  You need to distinguish clips from the clipboard so that only those get "copy" in their names.

I would still rather let clips carry some name-is-valid flag in them.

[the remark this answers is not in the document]
vsverchinsky: Then I think it can be simplified: if the track has the owner then use the 'copy' name pattern, if not - use the 'new' name pattern.

Issue #1201, opened 2021-06-30
AppImage fails to load FFmpeg


SteveDaulton:
**Describe the bug**
Self built Audacity 3.0.2 loads FFmpeg correctly.
audacity-linux-3.0.3-RC2-x86_64.AppImage fails to load FFmpeg (regression against 3.0.2)

**To Reproduce**
1. Launch audacity-linux-3.0.3-RC2-x86_64.AppImage
2. Observe error message:
`FFmpeg was configured in Preferences and successfully loaded before, 
but this time Audacity failed to load it at startup. 
You may want to go back to Preferences > Libraries and re-configure it.`

**Expected behavior**
FFmpeg will be loaded.

petersampsonaudacity:
That is almost certainly because you have 32-bit FFmpeg installed and you are running 64-bit Audacity.

I had similar on Windows until I installed  64-bit FFmpeg from Buanzo's new page on his site.
https://lame.buanzo.org/ffmpeg64audacity.php

Does the Linux FFmpeg page need updating?
https://alphamanual.audacityteam.org/man/Installing_and_updating_Audacity_on_Linux#linff

SteveDaulton:
I'm on 64-bit Linux, so I've been using 64-bit builds of Audacity and FFmpeg for years.

crsib:
@SteveDaulton can you please provide some insight on your setup?

I've managed to compile FFMPEG using a modified script from Leland and after setting up `DT_RUNPATH` everything works like a charm for me. I will share the script after a bit of polishing.

The question is, what we do for this release anyway. 

For the next release I plan:

* Upgrade the FFMPEG (once again, yes), so system FFMPEG can be used on Linux
* Provide a set of scripts on how to build the "correct" version, so it can be easily buildable by users.

@AnitaBats @Tantacrul FYI

SteveDaulton:
My system is pretty standard Xubuntu 20.04 LTS.
FFmpeg library installed from official Ubuntu repository.

I prefer to use the system FFmpeg library as it is a dependency for many other audio apps.
The usual  way to build Audacity 2.x with FFmpeg support was to disable dynamic loading during 'configure' - this has worked without any problems for years.

Self built 3.0.4-alpha-20210628 also loads FFmpeg without any problems.

FFmpeg version: F(58.29.100),C(58.54.100),U(56.31.100)

crsib:
Lemme have a quick look into the loader code

crsib:
[...]

PR #1797, opened 2021-09-30
Fix for AppImage issues, related to conflicts with system libraries

crsib:
Resolves: #1201 (FFmpeg failed to load because of `libcairo` being packaged)

Resolves a recent crash on Arch/Manjaro due to a conflict between bundled `gmodule`  and system `glib2`. 

The PR implements an approach, used by MuseScore, that allows defining a set of libraries that we prefer to be system.

Recommended:

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -195,7 +195,7 @@ This option implies `-Daudacity_obey_system_dependencies=On` and disables `local
 
 ### Disabling pre-built binaries downloads for Conan
 
-It is possible to force Conan to build all the dependencies from the source code without using the pre-built binaries. To do so, please pass `-Daudaicity_conan_allow_prebuilt_binaries=Off` to CMake during the configration. This option will trigger rebuild every
+It is possible to force Conan to build all the dependencies from the source code without using the pre-built binaries. To do so, please pass `-Daudaicity_conan_allow_prebuilt_binaries=Off` to CMake during the configuration. This option will trigger rebuild every
 time CMake configuration changes.
 
 ### Reducing Conan cache size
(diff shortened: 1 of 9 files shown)


BUILDING.md
```diff
@@ -188,14 +188,19 @@ $ docker run --rm -v ${pwd}:/audacity/audacity/ -v ${pwd}/../build/linux-system:
 
 To find system packages, we rely on `pkg-config`. There are several packages that have broken `*.pc` or do not use `pkg-config` at all. For the docker image - we handle this issue by installing the correct [`pc` files](linux/build-environment/pkgconfig/).
 
+### Disabling AppImage support
+
+AppImage requires an additional executable called `findlib`. This executable will be installed to `bin`. If this behavior is not desired, it can be disabled
```
imciner2, line 193: Does this default to `audacity_package_appimage=yes`? If so, why is it defaulting the build chain to assuming the appimage is the build instead of making the appimage build opt-in to this (which would require far fewer changes by downstream packagers)?

shoogle: It's one change for packagers :)

But I agree that `findlib` shouldn't get built by default, or perhaps it could be built by default but not installed to `bin`? My preference would be for it to get moved to `bin` in [package_appimage.cmake](https://github.com/audacity/audacity/blob/de2213ea2c794a294c1f242a75d5295512d9a73c/linux/package_appimage.cmake#L15-L17), which only runs if CPack is called, so no new CMake options would be required. This means somebody can compile Audacity and later decide to build the AppImage without having to reconfigure.

imciner2:
Yes, thanks for changing that!

Issue #1459, opened 2021-08-12
Set the DB page size using pragma, not with compile time option.


crsib:
The DB page size should be set when creating the database, not as a compile-time option. 

Refs #1457

gera-gas:
Tasks required: 2-4 hrs

PR #1799, opened 2021-10-01
Set the page_size for aup3 files using PRAGMA

crsib:
This change ensures that page size will be 64k even if system SQLite3 was used (mostly relevant for Linux)

Resolves: #1459

Recommended:

### DIFF
--- a/src/DBConnection.cpp
+++ b/src/DBConnection.cpp
@@ -24,8 +24,17 @@ Paul Licameli -- split from ProjectFileIO.cpp
 #include "wxFileNameWrapper.h"
 #include "SentryHelper.h"
 
+#define AUDACITY_PROJECT_PAGE_SIZE 65536
+
+#define xstr(a) str(a)
+#define str(a) #a
+
+static const char* PageSizeConfig =
+   "PRAGMA <schema>.page_size = " xstr(AUDACITY_PROJECT_PAGE_SIZE) ";"
+   "VACUUM;";
+
 // Configuration to provide "safe" connections
-static const char *SafeConfig =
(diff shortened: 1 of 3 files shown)


src/DBConnection.cpp
```diff
@@ -343,6 +361,23 @@ int DBConnection::FastMode(const char *schema /* = "main" */)
...
+int DBConnection::SetPageSize(const char* schema)
+{
+   // First of all - let's check if the database is empty.
+   // Otherwise, VACUUM can take a significant amount of time.
+   // VACUUM is required to force SQLite3 to change the page size.
+   // This function will be the first called on the connection,
+   // so if DB is empty we can assume that journal was not
+   // set to WAL yet.
+   int rc = sqlite3_exec(
+      mDB, "SELECT 1 FROM project LIMIT 1;", nullptr, nullptr, nullptr);
```
Paul-Licameli, line 373: I haven't seen this technique before... not SELECT * ?

crsib: `SELECT 1 ...` means that the engine doesn't have to read any data. This is the fastest possible query in this case because we only care that the `project` table has a row in it.

src/DBConnection.cpp
```diff
@@ -343,6 +361,23 @@ int DBConnection::FastMode(const char *schema /* = "main" */)
...
+
+   if (rc == SQLITE_OK)
+      return SQLITE_OK; // Project table exists, to late to VACUUM now
```
Paul-Licameli, line 376: "too"

Paul-Licameli (approved):

Issue #1705, opened 2021-09-22
Playback looping


gera-gas:
Mentioned: #1260

 - [ ] Add looping range GUI for Timeline bar.
 - [ ] Add mouse event handler for looping range GUI.
 - [ ] Remove "quick play" preferences like a checkboxes.

PR #1801, opened 2021-10-01
Playback looping draft implementation.

gera-gas:
Resolves: #1705 

Draft implementation.

Recommended:

### DIFF
--- a/src/AdornedRulerPanel.cpp
+++ b/src/AdornedRulerPanel.cpp
@@ -1223,10 +1223,12 @@ void AdornedRulerPanel::OnPaint(wxPaintEvent & WXUNUSED(evt))
       DoDrawSelection(&backDC);
    }
 
+#if 0
    if (PlaybackLooping::GetInstance().IsProcessing())
    {
        mPlaybackLoopingCell->DrawLoopingRange(&backDC, viewInfo);
    }
+#endif
 
    DoDrawMarks(&backDC, true);
 
@@ -2349,6 +2351,13 @@ void AdornedRulerPanel::UpdateStatusMessage( const TranslatableString &message )
 
 void AdornedRulerPanel::CreateOverlays()
 {
+    if (!mPlaybackLoopingOverlay)
+    {
+        mPlaybackLoopingOverlay = std::make_shared<PlaybackLoopingIndication>(mInner);
+        this->AddOverlay(mPlaybackLoopingOverlay);
+    }
+
+#if 0
    if (!mOverlay) {
       mOverlay =
          std::make_shared<QuickPlayIndicatorOverlay>( mProject );
@@ -2361,6 +2370,7 @@ void AdornedRulerPanel::CreateOverlays()
          pCellularPanel->AddOverlay( mOverlay );
       this->AddOverlay( mOverlay->mPartner );
    }
+#endif
 }
 
 void AdornedRulerPanel::TogglePinnedHead()
--- a/src/AdornedRulerPanel.h
+++ b/src/AdornedRulerPanel.h
@@ -235,6 +235,7 @@ private:
 
    //PlaybackLooping mPlaybackLooping;
    std::shared_ptr<PlaybackLoopingCell> mPlaybackLoopingCell;
+   std::shared_ptr<PlaybackLoopingIndication> mPlaybackLoopingOverlay;
 };
 
 #endif //define __AUDACITY_ADORNED_RULER_PANEL__
--- a/src/PlaybackLooping.cpp
+++ b/src/PlaybackLooping.cpp
@@ -19,8 +19,13 @@
 #include "TrackPanelMouseEvent.h"
 #include "SelectUtilities.h"
 #include "Track.h"
+#include "ProjectWindow.h"
+#include "CellularPanel.h"
+#include "AColor.h"
+#include "tracks/ui/TrackView.h"
 
 #include <wx/event.h>
+#include <wx/graphics.h>
 
 PlaybackLooping::PlaybackLooping()
 {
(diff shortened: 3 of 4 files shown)


src/AdornedRulerPanel.cpp
```diff
@@ -1213,9 +1214,20 @@ void AdornedRulerPanel::OnPaint(wxPaintEvent & WXUNUSED(evt))
...
+      );
+#endif
       DoDrawSelection(&backDC);
    }
 
+   if (PlaybackLooping::GetInstance().IsProcessing())
+   {
+       mPlaybackLoopingCell->DrawLoopingRange(&backDC, viewInfo);
+   }
+
```
Paul-Licameli, line 1230: TrackPanel uses CellularPanel::Draw which leaves the painting to the Draw() overrides of cell and handle objects.  It would be preferable but perhaps not needed to rewrite AdornedRulerPanel to be like it.

src/AdornedRulerPanel.h
```diff
@@ -15,6 +15,7 @@
 #include "widgets/Ruler.h" // member variable
 #include "Prefs.h"
 #include "ViewInfo.h" // for PlayRegion
+#include "PlaybackLooping.h"
```
Paul-Licameli, line 18: You should not need to include a header inside another header just to declare a shared_ptr.  A forward declaration of an incomplete type would be better.

src/PlaybackLooping.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <wx/event.h>
+
+PlaybackLooping::PlaybackLooping()
+{
+	;
+}
+
+PlaybackLooping& PlaybackLooping::GetInstance()
+{
+	static PlaybackLooping instance;
```
Paul-Licameli, line 32: Why a static instance?  Why not per-project state?

src/PlaybackLooping.h
```diff
@@ -0,0 +1,101 @@
...
+        kPreview,
+        kCancel
+    };
+
+    bool LoopingProcessor(AudacityProject* pProject, wxCoord& mousePosX, ProcessingEvents processingEvents);
+
+    bool IsProcessing() const;
+
+    PlayRegion mOldPlayRegion;
+    PlayRegion mLoopedPlayRegion;
```
Paul-Licameli, line 55: Why are there new play region objects?  I expected changes in the display of the existing play region object.  I expected replacement of the double headed red arrow that shows the locked play region.  Instead I see both when I build.  And the new rectangle only displays, but doesn't affect what the play button does.

src/PlaybackLooping.cpp
```diff
@@ -0,0 +1,208 @@
...
+PlaybackLoopingHandle::PlaybackLoopingHandle()
+{
+	;
+}
+
+UIHandle::Result PlaybackLoopingHandle::Click (const TrackPanelMouseEvent& event, AudacityProject* pProject)
+{
+	if (event.event.LeftIsDown())
+	{
+		if (!PlaybackLooping::GetInstance().LoopingProcessor(pProject, event.event.m_x, PlaybackLooping::ProcessingEvents::kClick))
```
Paul-Licameli, line 109: It is the purpose of a UIHandle object to hold temporary state needed only for the duration of a click-drag-release.  There should not be a static object serving as a state machine.

A UIHandle changes some more permanent object ouside of itself in response to clicks and drags.  That would be the play region.  But I also said the play region should not be a global but per-project.

If I open two Audaicty projects, the new red rectangles do not change independently as they should.

src/PlaybackLooping.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+	auto start = playRegion.GetStart();
+	auto end = playRegion.GetEnd();
+
+	return RefreshCode::RefreshCell;
+}
+
+UIHandle::Result PlaybackLoopingHandle::Cancel(AudacityProject* pProject)
+{
+	return RefreshCode::RefreshCell;
```
Paul-Licameli, line 164: The purpose of a Cancel method is to implement the ESC key, which should restore all state (here, the play region) to what it was before the click.  This is a must-do.

src/PlaybackLooping.h
```diff
@@ -0,0 +1,101 @@
...
+    bool mLoopingEnabled{ false };
+    bool IsLoopingEnabled() const;
+    void EnableLooping(bool enable);
+
+    enum class ProcessingStates
+    {
+        kIdle,
+        kEnterLooping,
+        kLooping,
+        kDragging
```
Paul-Licameli, line 36: The "state machine" of older AdornedRulerPanel code, left-over from before the reimplementation with CellularPanel -- is not a good model to follow.

You should not need this enumeration.  CellularPanel keeps track of state transitions between unclicked, clicked, dragging, button-up, and a UIHandle just needs to supply callbacks for those.

src/PlaybackLooping.cpp
```diff
@@ -0,0 +1,208 @@
...
+			auto loopingPos = viewInfo.PositionToTime(mousePosX, viewInfo.GetLeftOffset(), false);
+			mLoopedPlayRegion.SetStart(loopingPos);
+
+			mProcessingState = ProcessingStates::kEnterLooping;
+		}
+		break;
+
+	case ProcessingStates::kEnterLooping:
+		if (processingEvents == ProcessingEvents::kDrag)
+		{
```
Paul-Licameli, line 78: This incomplete state machine (which as I said, you should not need) is broken.  It never enters the drag state.

I can't actually drag the play region in this build.

I thought the intention also was different dragging behavior, in case of clicking inside the play region (to slide it as a whole), or at an edge (to stretch or shorten it).  None of that yet here.

src/PlaybackLooping.cpp
gera-gas: I want to isolate all playback looping logic (toggle-on, etc) into PlaybackLooping service, therefore static allow me using this service instance from toolbar button handlers.

Paul-Licameli: I do not accept this answer.

Paul-Licameli:
Your implementation is taking a wrong approach.

Address the objections I raised.

Do not make new PlayRegion objects.  Draw and modify the one that exists.

Do not use an Overlay object.  Adapt the existing OnPaint, which already draws a selection rectangle.  Make that draw the play region instead.  Eliminate the two-headed arrow.

Do not use a global object.  If you have two projects open, their play regions should be independent.

Do not make a state machine.  Other examples of UIHandle do not do this.  CellularPanel is the state machine, as I explained.  Just capture some information in Click, update it in Drag, and revert changes in Cancel.

Issue #1460, opened 2021-08-12
Do not fail if `SQLITE_ENABLE_DBPAGE_VTAB` is disabled


crsib:
References #1457 

Audacity should gracefully check if the `sqlite_dbpage` table is present and use a different estimation method if it is not.

gera-gas:
Tasks required: 4-6 hrs

PR #1802, opened 2021-10-01
Remove the requirement for the SQLITE_ENABLE_DBPAGE_VTAB

crsib:
Resolves: #1460

This PR introduces a different approach to estimating the SampleBlock size on disk. It does not account for the SQLite3 internal structures size so well, but the difference is relatively negligible. 

We use this estimation to decide if we need to compact the database on disk, as we do not allow database vacuuming (for performance reasons?). The performance of the heuristic can be further improved, as `used_blocks`/`total_blocks` should work just fine. However, this would require more changes to the code.

`ProjectFileIO::GetTotalUsage()` and `ProjectFileIO::GetCurrentUsage()` are also used in now disabled compaction dialog. However using `GetTotalUsage()` was probably not the idea anyway (because it never accounted for the project and autosave tables). `GetCurrentUsage()` still gives a very good estimate of how much disk space is used.

Recommended:

### DIFF
--- a/src/ProjectFileIO.cpp
+++ b/src/ProjectFileIO.cpp
@@ -33,6 +33,7 @@ Paul Licameli split from AudacityProject.cpp
 #include "wxFileNameWrapper.h"
 #include "XMLFileReader.h"
 #include "SentryHelper.h"
+#include "MemoryX.h"`
 
 // Don't change this unless the file format changes
 // in an irrevocable way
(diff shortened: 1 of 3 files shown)


src/ProjectFileIO.cpp
```diff
@@ -2364,219 +2365,80 @@ int64_t ProjectFileIO::GetTotalUsage()
...
          }
+      });
 
-         // Copy the page content to the stack frame
-         memcpy(&pg.data,
-                sqlite3_column_blob(stmt, 0),
-                sqlite3_column_bytes(stmt, 0));
-
-         // And retrieve the total number of cells within it
-         pg.numCells = get2(&pg.data[3]);
```
Paul-Licameli, line 2446: The functions get2, get4, get_varint become unused and might be deleted.

Paul-Licameli (approved):
One suggestion, to delete more stuff that we don't need any more.

src/ProjectFileIO.cpp
crsib: Good catch!

PR #1837, opened 2021-10-06
Fix inaccurate rhythmtrack.ny

tnzw:
Resolves: none (cannot find any issue about this)

When you choose tempo 89.071, 4 beats per bar for 10 minutes
there is a small delay between real-life click and audacity click
(losing around 2ms every 9 minutes).
Tested with different kind of click generator (like FL Studio)
every one have the same accuracy except Audacity.

Recommended:

NB : The `(for-range ...)` I use is because `(loop ...)` just crash my Audacity (3.0.4 on Windows 10). Perhaps there is a better way to make a dumb loop. Also, I don't know lisp language at all 😉 I just learnt a bit nyquist to fix this plug-in. The `(for-range ...)` allows around 1000 recursion on my machine (10 times less if I remove all `(when ...)` lines. I need it because `(simrep (i bars) ...)` crashes if `bars` is greater than 90. Thanks!

### DIFF
--- a/plug-ins/rhythmtrack.ny
+++ b/plug-ins/rhythmtrack.ny
@@ -43,6 +43,24 @@ $control low (_ "MIDI pitch of weak beat") int (_ "18 - 116") 80 18 116
 
 ;; Helper functions:
 
+; cannot make (loop for ... from ... to ... do ...) to work...
+(defun for-range (start stop step fn)
+  (when (< start stop) (for-range-rec start 0 stop step fn)))
+(defun for-range-rec (start i stop step fn)  ; There is a recursion limit (like around 100 recusions max).
+  (setf for-range-rec-v (funcall fn (+ start (* i step))))
+  (when (< (+ start (* (+ i 1) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 1) step)))))
+  (when (< (+ start (* (+ i 2) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 2) step)))))
+  (when (< (+ start (* (+ i 3) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 3) step)))))
+  (when (< (+ start (* (+ i 4) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 4) step)))))
+  (when (< (+ start (* (+ i 5) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 5) step)))))
+  (when (< (+ start (* (+ i 6) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 6) step)))))
+  (when (< (+ start (* (+ i 7) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 7) step)))))
+  (when (< (+ start (* (+ i 8) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 8) step)))))
+  (when (< (+ start (* (+ i 9) step)) stop) (setf for-range-rec-v (funcall fn (+ start (* (+ i 9) step)))))
+  (when (< (+ start (* (+ i 10) step)) stop) (setf for-range-rec-v (for-range-rec start (+ i 10) stop step fn)))
+  for-range-rec-v)
+
+
 (defun round-up (x)
   (if (> x (truncate x))
     (truncate (1+ x))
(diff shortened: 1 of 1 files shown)


LWinterberg:
@SteveDaulton I believe you are our nyquist wizard (nyquizard?) around here

SteveDaulton:
The current Rhythm Track effect has been written to use memory efficiently, which it allows it to generate very long click tracks reasonably quickly. It does this by generating one bar, then repeating that bar the required number of times. A limitation of this approach is that if the bar length is not an exact number of samples, then it is rounded to an exact number of samples.

Example:
For a tempo of 199.999 bpm, each beat has a duration of 0.500004167 seconds 
= 22050.183751531 samples @ 44.1 kHz
= 88200.735006125 samples per bar at 4 beats per bar
So if the first bar starts at time = 0, then the second bar starts at time = 88201 samples (rounded to nearest sample) rather than 88200.735 samples.
In this case there is a rounding error of around 0.0003%

The current effect is sample accurate if a bar length is an exact number of samples, so there is a maximum cumulative error of 0.5 samples per bar. For a typical bar length of around 2 seconds, and a typical sample rate of 44.1 kHz, that's less than 0.0006%

How accurate are sound card clocks?
A quick Google search brings up this page, http://www.leapsecond.com/pages/sound-1pps/ . While not being very scientific, their result of 44 ppm seems to be in the right ballpark (0.0044%).

So in my opinion, the current effect is sufficiently accurate for virtually all practical purposes (and the accuracy can be improved by changing the sample rate).

Can the accuracy be improved? Yes.
Is it worth the effort? Possibly not.

Moving onto the proposed patch.
10 bars, 199.99 bpm, 4 beats per bar, 44.1 kHz
That is 40 beats, with 22050.183751531 samples per beat = 882007 samples (to the nearest sample).
With the proposed patch, 10 bars is indeed 882007 samples, whereas the current version is 882010 samples, so yes it is more accurate.
However, given that we are talking about "perfect accuracy", it is worth noting that each click is still aligned to an integer sample position, so individual clicks still deviate (very very slightly) from the exact time position unless the beat length is an exact number of samples.

The biggest problem with this proposed patch is that it fails when generating long tracks. Try generating a long track (say 1 hour) and it fails:
`error: argument stack overflow`

Additional note:
A "FOR" loop in Nyquist is done with "DO" (https://www.audacity-forum.de/download/edgar/nyquist/nyquist-doc/xlisp/xlisp-ref/xlisp-ref-093.htm)
There is also "DO*" (https://www.audacity-forum.de/download/edgar/nyquist/nyquist-doc/xlisp/xlisp-ref/xlisp-ref-094.htm)

tnzw:
Thanks the feedback!

To me, I expect to have mathematical "perfect" timing has I am working on a computer. (i.e. `for i in range(bars): at_abs = offset + i * beatlen` instead of `for i in range(bars): cue(measure)`)

Not talking about high sample rates, 2ms in 9m is too much and mathematical timing (+/- 1/44100) is good enough for my needs. The commit should be renamed `increase rhythmtrack.ny accuracy`…?

> The biggest problem with this proposed patch is that it fails when generating long tracks. Try generating a long track (say 1 hour) and it fails

Indeed, the `(do …)` should fix it, thanks for the links!!

> One possible approach to achieving "perfect" accuracy would be for the plug-in to generate the click track at a sample rate that provides an exact number of samples per bar, then resampling back to the track rate. […]

If we want a perfect-perfect accuracy, then I don't have time to think (or write) about a solution. But if the proposed improvement is ok, then I'll take the time to use `do` to fix the stack overflow.

> […] You could make two pre-computed measures: one shorter and and one longer […]

Good idea! Also, I won't take time to implement it, but sure thing it'll be high performance code and as accurate as my patch.

LWinterberg:
Fwiw: The current solution works well enough for me, but clearly doesn't for you. I think that just improving on it where it works for you may run into exactly the same issue, where someone else will find the more accurate version not accurate enough for their needs. 

Because of this, I think that having it fully accurate is the more useful approach to put work in. 

Also, as an aside: rhythm tracks really are more of a stop gap solution to Audacity not having a proper metronome. It is planned for a far-future Audacity to have a proper "musical mode" (#1788) which does all the things musicians need. Now, I definitely won't ask you to make all of it yourself, but if you happen to be more fluent in C++ than in Nyquist, maybe starting work on a metronome feature could be way more fruitful for you.

tnzw:
> Because of this, I think that having it fully accurate is the more useful approach to put work in.

Right, I understand.

> […] but if you happen to be more fluent in C++ than in Nyquist, maybe starting work on a metronome feature could be way more fruitful for you.

Unfortunately not fluent in C++. As a musician I'm more user than developer, else I'm used to code in Python and Javascript.

Well, eager to see the musical mode!

I think PR is ready to be closed 😉 Thanks for the review!

LWinterberg:
Thank you for the contribution!

Issue #1710, opened 2021-09-23
The use of the shortcut Enter for renaming introduces duplicate shortcuts


DavidBailes:
**Describe the bug**
The shortcut Enter has been hardcoded into Audacity for renaming labels and clips. It has precedence over any command which have been assigned that shortcut. By default Enter is assigned to the command to toggle the selectedness of the focus track, but the user could have assigned it to any command. 

**To Reproduce**
Assuming default shortcuts, either full or standard.
Steps to reproduce the behavior with an audio track:
1. Create an audio track with some content.
2. Select a clip.
3. While the clip is selected, the user can no longer use the shortcut Enter for toggling the selectedness of of the track. Pressing Enter opens the name of the clip for editing.

Note that if a clip is not selected, then pressing Enter does nothing. I assume that's just a bug in the implementation.

Steps to reproduce the behavior with a label track:
1. Create a label track containing some labels.
2. Select a label.
3. While the label is selected, the user can no longer use the shortcut Enter for toggling the selectedness of the track.

Note a label can remain selected even if the cursor/selection has been changed since the label was selected.

**Expected behavior**
When a track is the focus, the user should always be able to execute the command to which Enter has been assigned.

**Additional information (please complete the following information):**
Tested on Windows 10

DavidBailes:
https://github.com/audacity/audacity/pull/1857 was not a satisfactory fix.

crsib:
> then I'm not sure our designers would like it

I think Jouni was open to a different shortcut. 

>New Esc key behavior for label tracks is acceptable? 

The problem is, that this behavior is not implemented for clips, so it is only a half fix

PR #1857, opened 2021-10-07
Issue 1710 compromise

Paul-Licameli:
Resolves: #1710

This does not resolve all as suggested in the issue description, but may offer satisfactory workarounds.

For label track behavior:

The ESC key already removes the insertion cursor if there is one.  But now, if there no insertion cursor but there is a label navigated to by TAB, then ESC clears the navigation state.  This allows ESC and then RETURN to change the selected state of the label track.

For wave track behavior:

The RETURN (or number pad ENTER) key behavior will remain hard-coded.  It is also hard-coded for label track text editing, about which there isn't a complaint.

"Note that if a clip is not selected, then pressing Enter does nothing. I assume that's just a bug in the implementation."  This is fixed.  The RETURN key will be intercepted for editing only in case the selection coincides with the extent of a clip.

With this change, you can change the selection such as with J or left arrow in default key bindings so it becomes a point.  Then RETURN toggles the selected state of the wave track.

You may be unsatisfied that restoring the time selection to what it was before requires further TAB key presses.  Maybe further improvement is possible, and you, @DavidBailes , might suggest another commit to accomplish that.

Recommended:

### DIFF
--- a/src/tracks/labeltrack/ui/LabelTrackView.cpp
+++ b/src/tracks/labeltrack/ui/LabelTrackView.cpp
@@ -1509,6 +1509,7 @@ bool LabelTrackView::DoKeyDown(
    const auto pTrack = FindLabelTrack();
    const auto &mLabels = pTrack->GetLabels();
    if (IsValidIndex(mTextEditIndex, project)) {
+      // Do label text changes
       auto labelStruct = mLabels[mTextEditIndex];
       auto &title = labelStruct.title;
       wxUniChar wchar;
@@ -1676,8 +1677,12 @@ bool LabelTrackView::DoKeyDown(
    }
    else
    {
+      // Do navigation
       switch (keyCode) {
 
+      case WXK_ESCAPE:
+          mNavigationIndex = -1;
+          break;
       case WXK_TAB:
       case WXK_NUMPAD_TAB:
          if (!mLabels.empty()) {


DavidBailes:
> Resolves: #1710
> 
> This does not resolve all as suggested in the issue description, but may offer satisfactory workarounds.

This does not offer satisfactory workarounds.
I am at a loss as to how this issue has been handled. This was the initial design:
https://github.com/audacity/audacity/issues/1258
As you can see I raised the issue of Enter being a duplicate on 9 Jul, and there was a reply from @jounih:
"@DavidBailes makes sense to have customisable commands for selecting clips and editing names rather than hardcode these shortcuts. @vsverchinsky can you do this please?"
And here we are on 8 Oct with a duplicate shortcut.

Have the designers said  that there has be a duplicate shortcut?
Why can't this be fixed properly with a command and a user assignable shortcut like it's done in the rest of Audacity?
I really don't understand what is going on here.

crsib:
Shortcuts are generally context-sensitive. I agree, that ideally this should be implemented as a configurable shortcut. Workaround Paul proposes solves the problem of context switching. And it is (almost, see below) satisfactory in case we won't be able to implement the configurable shortcut in time for 3.1.

However, after testing I found out that it is not possible to deselect a clip with an Esc key and I think this should be fixed to @Paul-Licameli @Penikov (cc @AnitaBats)

crsib:
And in any case, I believe Esc should remove the selection from the clip @Paul-Licameli @Tantacrul @vsverchinsky

Paul-Licameli:
The notion of selectedness (or focus) of a clip, distinct from a time selection coindident with the clip boundaries, needs to become a real thing in the code but it needs thought and work and no short fix is going to do it now.

I have raised the idea that the set of keystroke bindings should not be a single global one, as now, but might more usefully by contextual.  That is, Return might be customized for one thing in general (changing track selection) but a different set of customizations could apply and override the meaning of Return when a clip or label or other thing is focused (starting the editing of a name).

That's a nice idea that has occurred to me before, still worth considering, but it might take a lot of elaboration to make it work, both in restructuring our CommandManager and in making a convenient way to customize the keystrokes in Preferences.  And it presupposes that the notion of clip focus is properly implemented first.

Paul-Licameli:
This pull request was approved and merged.  It has been judged insufficient for the issue.  But I ask again whether anyone thinks it contained anything wrong that should be reverted.  No one has said so.

PR #1865, opened 2021-10-08
Fix text edit shortcuts

vsverchinsky:
Resolves: #1833

Also "Select All" command is available when label is edited

Recommended:

### DIFF
--- a/src/menus/EditMenus.cpp
+++ b/src/menus/EditMenus.cpp
@@ -25,6 +25,7 @@
 #include "../export/Export.h"
 #include "../prefs/PrefsDialog.h"
 #include "../tracks/labeltrack/ui/LabelTrackView.h"
+#include "../tracks/playabletrack/wavetrack/ui/WaveTrackView.h"
 #include "../widgets/AudacityMessageBox.h"
 
 // private helper classes and functions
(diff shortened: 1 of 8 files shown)


Paul-Licameli:
Done reviewing

vsverchinsky:
> Actually, DoPasteText() used to handle the case of insertion of text at the edit cursor of a label, always correctly pushing the undo stack. But this PR causes a definite regression of that behavior, preempting DoPasteText().

Yet another thing I overlooked, but the fix seems to be quite simple: move project history modification code to `LabelTrackView::PasteText` and remove it from `OnPasteNewLabel`, `LabelTrackView::OnContextMenu` (though history item text will change) and `DoPasteText` becomes unneeded, so it can be removed. Also, I've noticed that there is some GTK-specific code that does not modify undo history as well
```
if (evt.MiddleDown()) {
         // Paste text, making a NEW label if none is selected.
         wxTheClipboard->UsePrimarySelection(true);
         view.PasteSelectedText(project, newSel.t0(), newSel.t1());
         wxTheClipboard->UsePrimarySelection(false);
      }
```
And it will be fixed too if the history modification code is moved to `LabelTrackView::PasteText`

Do you think it would be better now to avoid changing existing code for labels, and add some workarounds to handle text shortcuts in Wave Track?

Issue #1834, opened 2021-10-05
[Clip Renaming] Double-clicking to rename a track should automatically mark the entire clip name


crsib:

Penikov:
At the first double-click, it works well, but next double-clicks do nothing. 
So this task is done, I'll add a new bug:

https://user-images.githubusercontent.com/25851359/137305900-a31875f0-9ebc-4e91-8c32-3cc539b7aaf0.mp4


https://user-images.githubusercontent.com/25851359/137305904-bac599fb-7850-4e5b-8acd-a2a65d1f83d6.mp4

PR #1874, opened 2021-10-08
Fix clip name editing

vsverchinsky:
Resolves: #1835
Resolves: #1834 

Recommended:


petersampsonaudacity:
Tested on W10 with Audacity_Windows_64bit_1321978882_0f475f9  - branch build by @vsverchinsky 

This fix by Vitaly looks good - the Clip name is highlighted (for editing) when you double-click on the clip.

Issue #1578, opened 2021-08-31
A warning should be issued  to the user when using Reset Configuration - with the option to abort


petersampsonaudacity:
**Is your feature request related to a problem? Please describe.**
This enhancement request arises from a user post on the Forum:
https://forum.audacityteam.org/viewtopic.php?f=70&t=120059&p=433996
Why occurs the Configuration Reset without confirmation / warning?

Currently the **Tools>Reset Configuration** command opetes immediately with
a) no warning od what will happen
b) no ability to undo
So the user that accidentally uses thiswill find their configuration changed back to factory settings - and with no chance to **Undo**.

**Describe the solution you'd like**
When the user invoke the **Tools>Reset Configuration** command a warning dialog should be issued:

a) Similar or same text as in the intro div in the Manual
![image](https://user-images.githubusercontent.com/50205984/131464806-d51e3730-87b3-43bf-bf45-c1f535985344.png)

b) has an "Are you sure" Yes/No - offering the chance to abort

c) Has a checkbox for "Do not show this message again"

d) has a "?" help button linked to the Manual  page
https://manual.audacityteam.org/man/reset_configuration.html

**Describe alternatives you've considered**
Undo to be provided for this command

**Additional context**
I have assigned this to P2 as the suer may have a complex set of confiration and Tol=olbar settings which could take a lot of work to restore by manual **resets.**

JamesCrook:
The experienced teacher who asked for this feature (because he has to go round the classroom resetting Audacities got into strange states by the students) specifically asked for a reset that did not ask 'are you sure?'.

The "don't ask again" is pointless, as that is a setting and should be reset too.  Unless you make an exception for it.

SteveDaulton:
I would support the proposal to open a dialog window rather than immediately resetting (some) settings, if the dialog served a useful purpose, such as indicating what it does, or better, providing some additional options (such as resetting effects, and / or backing up preferences).

Personally I think the "risk" of the command as it currently is, is very small - no more than a minor inconvenience if clicked accidentally.

petersampsonaudacity:
> The experienced teacher who asked for this feature (because he has to go round the classroom resetting Audacities got into strange states by the students) specifically asked for a reset that did not ask 'are you sure?'.

He or she was by no means the first - this was a long-standing request from users and indeed the Forum elves who otherwise had to tell people how to void all or parts of their Audacity settings folder.

[...]

PR #1878, opened 2021-10-09
Add a Reset Configuration Dialog that warns the user as well as give options for resetting different preferences individually.

malevolentstrix:
image

Resolves: #1578 

This will add a dialog box that warns the user and gives the option to individually reset preferences that are usually reset by users.

Recommended:


SteveDaulton:
Testing on Xubuntu 20.04 with 1920x1080 display, there's a display problem:
prefs

The window CAN be resized to fit the entire text, but it defaults to the "too small" size each time it opens.

SteveDaulton:
While not *directly* related to this pull request, I  think it would be better for the default set of keyboard shortcuts to be the "Full" set. (Perhaps rename the "standard" set as "common shortcuts only" or similar.

SteveDaulton:
More often than not, I would want to be selective about which prefs to reset.

Perhaps "Reset Configuration" should remember it's last used settings, (or perhaps that might be a bit weird as that would be one setting that must not be reset). 
If it does NOT remember what the user did last, then maybe a better default would be nothing selected. (It already has the instruction: "Select the configurations you want to reset"). That would allow users to directly select one or more options without having to deselect "All configurations" first.

SteveDaulton:
Other than the minor issues that I've noted, this patch works well for me on Xubuntu 20.04. I like it a lot :+1:

MalevolentStrix:
> Other than the minor issues that I've noted, this patch works well for me on Xubuntu 20.04. I like it a lot +1

I will remove the default checking of the boxes, make the keyboard options in a separate static text group, and will do something to get the window size correct as you mentioned. 

Thanks a lot for the review @SteveDaulton  :slightly_smiling_face:

MalevolentStrix:
>The window CAN be resized to fit the entire text, but it defaults to the "too small" size each time it opens.

@SteveDaulton Can you check if the window is still smaller than the content as before? I couldn't find a proper way to set it universally, I tested it on Windows and Ubuntu 20.04 and it worked fine.

LWinterberg:

Also, sorry for dropping the ball on this one - we should've merged it when it still was fresh.

Issue #1780, opened 2021-09-29
A label can be created by typing regardless of option setting


DavidBailes:
**Describe the bug**
If a label is selected, then you can type to create a label, even if the "Type to create a label" option is disabled.

**To Reproduce**
1. Disable the Type to create a label option in the track behaviors page of preferences
2. Create a label track with one or more labels
3. Tab to a label to select it (or click on a region label)
4. Type a letter.
5. A new label is created.

**Expected behavior**
If "Type to create a label" is disabled, then typing never creates a label.

**Fixing this bug:**
Once this issue has been fixed: https://github.com/audacity/audacity/issues/1710,
then I think the fix is straightforward:
In LabelTrackView::DoCaptureKey(), the lines:
if (IsValidIndex(mTextEditIndex, project) || IsValidIndex(mNavigationIndex, project)) {
      if (IsGoodLabelEditKey(event)) {
         return true;
      }
   }
should be changed to:
In LabelTrackView::DoCaptureKey(), the lines:
if (IsValidIndex(mTextEditIndex, project)) {
      if (IsGoodLabelEditKey(event)) {
         return true;
      }
   }

If this change is made before #1710 is fixed, then the Enter key will no longer be captured for the hardcoded use of Enter to edit the label name. The code would have to be modified to capture this, but it shouldn't be hardcoded, so best to fix #1710 first.

petersampsonaudacity:
And it creates a second label in the same range of the region label.

DavidBailes:
Example of the effects of this bug:
With "type to create a label" disabled:
1. Select a label.
2. Press spacebar for playback.
3. Press P for Pause. Rather than pause a new label is created starting with P.

PR #1881, opened 2021-10-11
issue #1780: An unwanted label can be created by typing

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/1780

If a label is selected, then you can type to create a label, even if the "Type to create a label" option is disabled.

This bug was introduced in commit: a0ad72d
I presume that in the following line in LabelTrackView::DoCaptureKey:
if (IsValidIndex(mTextEditIndex, project) || IsValidIndex(mNavigationIndex, project)) {,

IsValidIndex(mNavigationIndex, project) was included to let the hardcoded Enter key be captured. This was not an appropriate way of doing so - the Enter key should have been handled individually, just like the tab key. This introduced the current bug.

Fix: Given that use of the hardcoded Enter key is shortly to be removed, then IsValidIndex(mNavigationIndex, project) can be removed.

Recommended:

### DIFF
--- a/src/tracks/labeltrack/ui/LabelTrackView.cpp
+++ b/src/tracks/labeltrack/ui/LabelTrackView.cpp
@@ -1352,7 +1352,7 @@ bool LabelTrackView::DoCaptureKey(
        !mLabels.empty())
       return true;
 
-   if (IsValidIndex(mTextEditIndex, project) || IsValidIndex(mNavigationIndex, project)) {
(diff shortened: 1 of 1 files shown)


DavidBailes:
When I wrote this, my understanding was the the use of the hardcoded Enter key was shortly to be removed.
It's now uncertain whether this is going to be case.
Hopefully this will be decided in the near future, and if necessary I, or someone else can amend this fix so that any hardcoded keys for renaming are captured.

Issue #1842, opened 2021-10-06
Toggle Focused Track ("Return" or "NUMPAD_ENTER") fails


SteveDaulton:
**Describe the bug**
Pressing the "Enter" or "NUMPAD_ENTER" does not toggle selection of the focused track.

**To Reproduce**
1. Generate a "Chirp"
2. "Ctrl + D" (duplicate)
3. "Enter" to deselect the first track
4. The first track is still selected.

**Expected behavior**
The "Toggle Focused Track" keybindings should toggle the selection of the focused track.

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version: 3.1.0-alpha-20211006

**Additional context**
This is a very important command when working with multiple tracks.

Paul-Licameli:
#1857 did mitigate the problem:  it occurs now only when the selection coincides with the extent of a clip, not always.

A fuller solution that we will do:  Remove the hard coded Return key so Start name editing.  (But not for finishing it.)

That will solve this issue as stated, but then leave @DavidBailes unsatisfied that there is any keyboard accessibility for clip name editing.

That will deserve a different issue, and I don’t promise a fix for that in 3.1.0.

I think it would be a good idea to allow various hierarchical levels of focus (none, track, or labels or clips within the track) and overload keystroke bindings depending on the focus, so that Return would be free to repurpose in this context, instead of using some other less mnemonic key (and running out of keys).

But we can’t do that yet.

I think this is exposing some of the limitations of our system of focus, navigation, and keystroke binding.  Watch for a new Gitub Issue from me about my ideas for a better system.

SteveDaulton:
> #1857 did mitigate the problem: it occurs now only when the selection coincides with the extent of a clip, not always.

Testing with 3.1.0-alpha-20211007

> **To Reproduce** 
> Steps to reproduce the behavior:
> 
>     1. Generate a "Chirp"
>     2. "Ctrl + D" (duplicate)
>     3. "Enter" to deselect the first track
>     4. The first track is still selected.

I see no mitigation.

Paul-Licameli:
> > #1857 did mitigate the problem: it occurs now only when the selection coincides with the extent of a clip, not always.
> 
> Testing with 3.1.0-alpha-20211007
> 
> > **To Reproduce**
> > Steps to reproduce the behavior:
> > ```
> > 1. Generate a "Chirp"
> > 2. "Ctrl + D" (duplicate)
> > 3. "Enter" to deselect the first track
[...]

PR #1898, opened 2021-10-12
Customizable Ctrl+F2 not hard-coded Return to edit clip names...

Paul-Licameli:
... This inserts a menu item under Edit.

Also share code with the clip context menu

Resolves: #1842 

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
@@ -14,6 +14,10 @@
 #include <wx/frame.h>
 
 #include "AllThemeResources.h"
+#include "../../../../commands/CommandContext.h"
+#include "../../../../commands/CommandFlag.h"
+#include "../../../../commands/CommandFunctors.h"
+#include "../../../../commands/CommandManager.h"
 #include "../../../../TrackPanelMouseEvent.h"
 #include "../../../../TrackArtist.h"
 #include "../../../../TrackPanelDrawingContext.h"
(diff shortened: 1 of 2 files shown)


DavidBailes:
Thanks for this, but it only partially resolves https://github.com/audacity/audacity/issues/1710. As stated in the steps to reproduce, the problem occurs in label tracks as well as audio tracks.

DavidBailes:
> I have said and will say again that I think that what we need (what will take more development than a bug fix) is an improved system of hierarchical foci and context sensitive assignment of shortcuts to actions:
> 
> * Develop a proper notion of nested foci (Track panel, track, clip or label)
> * Have easy keystroke navigation among them,  not just tabbing among peers but up- and down-hierarchy. (like what I can do with VoiceOver on Mac, but built-in to Audacity even without screen readers)
> * Let Return and other keys "scope" differently depending on what is focused.
> * Advantage is not running out of scarce keystrokes as more functionality is added, or having to use very un-mnemonic ones or modifier key chords, just to avoid conflicts in a single flat scope.
> 
> Your opinions of that?

I think you mentioned somewhere else that you were going to create an issue/discussion for this. I think that would be a better way of getting feedback, rather than on this pull request.

DavidBailes:
@Tantacrul Could you please confirm whether or not you think the shortcuts for renaming a "selected" clip and a selected label should be the same?

DavidBailes:
> I pushed another commit. Try it.

I tried the commit and still think it's confusing.
1. I still think that it's not at all intuitive that a keystroke that is used for changing the selectedness of a track may open a label name of editing if the label is selected.
2. A keyboard user can quite easily not know that a label is selected - either because it is off screen, or the user is using a screen reader.
3. In your commit, when a user pressed Enter to finish an edit of a label, this now deselects the label. I don't think a user would expect that.

> I want a compromise that satisfies you and us in short time.

I think we need to know whether the designers want the same command and default shortcut for renaming selected clips and labels. (And as I've said, if they do, I'm happy for Ctrl+F2 to be hardcoded for labels as a temporary measure if time is short).

Issue #1752, opened 2021-09-26
Mac: many cosmetic issues and Selection toolbar issues with latest 3.1.0 alpha


petersampsonaudacity:
**Describe the bug**
There are many cosmetic issues with the the latest 3.1.0 alpha Audacity_macOS_Intel_1270895276_97ca83e 

There are also issues withe the Selection Toolbar where the Project Rate and Snap To fields are blank and cannot be set (or observed) 

**To Reproduce**
1. Install latest alpha Audacity_macOS_Intel_1270895276_97ca83e  on Mac
2. Launch
3.  Observe Scroll bars for the Audacity Window are black (not gray)
4. Observe: Window header is  black (not gray)
5. Observe: The bottom bar (including the Status bar) is black (not gray)
6. Observe: welcome mat message has black header not gray
6. Observe: in the Selection Toolbar where the Project Rate and Snap To fields are blank and cannot be set (or observed) 

**Expected behavior**
a) same graphics as in 3.0.4 with gray not black
b) active and filled Project Rate and Snap To fields in the Selection toolbar

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/134815763-fafdc8a9-6765-40fb-bd24-4f637825fcf8.png)

**Additional information (please complete the following information):**
 - OS: macOS 11.5.2 Big Sur
 - Version Audacity 3.1.0

**Additional context**
This is a regression on 3.0.4

This does not happen on Windows with the latest alpha:  audacity-win-3.1.0-alpha-20210924+97ca83e-64bit

petersampsonaudacity:
Preferences come up white on black:
![image](https://user-images.githubusercontent.com/50205984/134816235-24022d02-f85f-4e70-94a7-d1601ca8d0f7.png)

petersampsonaudacity:
Mnues come up white text on black too:
![image](https://user-images.githubusercontent.com/50205984/134816391-23e2a2db-2cd9-4c9c-a789-d72a2dd5b06a.png)

petersampsonaudacity:
So do the dropdown menus for the time format controls in the Selection toolbar and the Timer toolbar

Paul-Licameli:
See what happens in ` ThemeBase::LoadTheme`

petersampsonaudacity:
My bad - Apple had "kindly" turned on a fashionable dark theme for me with the recents11.6 macOS upgrade.

When I switch it to light all is fine.

[...]

PR #1914, opened 2021-10-13
Allow the application appearance to match the theme on macOS

crsib:
Currently, this is implemented for macOS users only.
Windows has not yet published the API, Linux requires GTK3 and we build against GTK2 currently.

Resolves: #1752

Recommended:


LWinterberg:
@ohquait you may want to ask in the [forum](https://forum.audacityteam.org/viewforum.php?f=19) or the [dev discord](https://audacityteam.org/devserver) about this, instead of some random pull request that already is merged :)

Issue #1739, opened 2021-09-25
LOF files do not support Unicode (or Extended ASCII)


petersampsonaudacity:
**Describe the bug**
LOF files do not support Unicode (or Extended ASCII)

LOF files are currently limited to standard / lower ASCII characters only in file names. This prevents file names such as "Reinyö.wav" being used, even though the name is valid on all supported platforms.

Also, the error message "Audacity did not recognize the type of the file..." is not as helpful as it could be, and this limitation is not currently documented in the manual.

**To Reproduce**
1. Create an audio file containing characters that are not in the lower ASCII character table. Example: "®±µ».wav"
2. Create a LOF file containing the file created in step 1.
3. Import the LOF file into Audacity.
4. Observe error: "Audacity did not recognize the type of the file..."

**Expected behavior**
LOF files should support Unicode (or Extended ASCII)

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: all OS
 - Version all Audacity - originally reported against 3.0.0
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

**Additional context**
"lower ASCII":
https://www.computerhope.com/jargon/a/ascii.htm#standard-ascii-table)

petersampsonaudacity:
@SteveDaulton wrote to me on email:

>We've never really made much of a deal of LOF files.
I like the idea, and they are potentially useful, but it seems to me to be an underdeveloped feature.

>Perhaps better than logging this as a P3 bug, would be a feature request for "playlist" support, using an established format such as M3U or  PLS rather than LOF.

>M3U: https://en.wikipedia.org/wiki/M3U
PLS: https://en.wikipedia.org/wiki/PLS_(file_format)

>An interesting aspect of this is that a multi-track project could be exported as separate WAV files plus a playlist file - could be useful for backups, and for collaborative projects.

PR #1916, opened 2021-10-13
Fix bug with LOF unicode using.

gera-gas:
Resolves: #1739 

Recommended:

### DIFF
--- a/src/import/ImportLOF.cpp
+++ b/src/import/ImportLOF.cpp
@@ -180,25 +180,71 @@ TranslatableString LOFImportPlugin::GetPluginFormatDescription()
 std::unique_ptr<ImportFileHandle> LOFImportPlugin::Open(
    const FilePath &filename, AudacityProject *pProject)
 {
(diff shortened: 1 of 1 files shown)


SteveDaulton:
> Maybe you need this to guess the encoding:

Or maybe there's no need to check, other than ensuring that it isn't a huge file. wxTextFile seems to be pretty robust wrt opening files that don't contain text. If the file isn't readable as text, the import will fail, so we just need to tell the user that it failed.

It's unfortunate that LOF files were not defined to have a standard header of some sort so that we could easily recognize them. Something to keep in mind  if the LOF format is ever extended. As it is, about the only thing that  we know a LOF file *must* have is at least one occurrence of the word "file" as the first non-whitespace word on a line.

gera-gas:
@SteveDaulton I add new checking control and test in on ASCII, UTF-8 (with and without BOM) and UTF-16 (with and without BOM) too, on my PC all work fine.

PR #1920, opened 2021-10-13
Try to import as Raw Data when Audacity did not recognize the type of the file(Drag file into audacity case)

cjw1115:

I'm on windows audio processing related work, and there is a lot of raw audio data is dumped in processing pipe line.
When I want to import those raw data files into audacity, I have to go to file-import-raw data, but I prefer to just drag it into audacity window to parse raw data. so I request this pull-request.

Recommended:

### DIFF
--- a/src/import/Import.h
+++ b/src/import/Import.h
@@ -165,10 +165,13 @@ public:
     * Allocates NEW ExtImportItem, fills it with default data
     * and returns a pointer to it.
     */
-    std::unique_ptr<ExtImportItem> CreateDefaultImportItem();
+   std::unique_ptr<ExtImportItem> CreateDefaultImportItem();
 
-   // if false, the import failed and errorMessage will be set.
-   bool Import( AudacityProject &project,
+   /**
+    * if first tuple item is false, the import failed and errorMessage will be set.
+    * if second tuple item is true, means format is not supported, can try to import it as raw data.
+    */
+   std::tuple<bool, bool> Import( AudacityProject &project,
               const FilePath &fName,
               WaveTrackFactory *trackFactory,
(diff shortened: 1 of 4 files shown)


src/import/Import.h
```diff
@@ -165,10 +165,13 @@ class AUDACITY_DLL_API Importer {
...
-    std::unique_ptr<ExtImportItem> CreateDefaultImportItem();
+   std::unique_ptr<ExtImportItem> CreateDefaultImportItem();
 
-   // if false, the import failed and errorMessage will be set.
-   bool Import( AudacityProject &project,
+   /**
+    * if first tuple item is false, the import failed and errorMessage will be set.
+    * if second tuple item is true, means format is not supported, can try to import it as raw data.
+    */
+   std::tuple<bool, bool> Import( AudacityProject &project,
```
Paul-Licameli, line 174: Rather than return a tuple, define an enumeration of three possible return values.

Paul-Licameli (changes requested):

Paul-Licameli:
Hello @cjw1115 , if you are still interested in having this pull request merged into the present development branch, then please rebase onto a recent master version, squash the commits into one, and force-push the branch.

I say squash because I see commits that correct errors in previous commits.  I'd rather have that cleaned up.  Not every PR should be done as one commit, but this one is not large enough to require multiple commits for clarity.

cjw1115:
@Paul-Licameli Sure, let me update it

Paul-Licameli:
That's good.  Now, please make the suggested other source code changes.  An enumeration of three possibly return types would be better than a tuple.

LWinterberg:

Issue #1884, opened 2021-10-11
Smart Clips. There is not allowed to exit the clip name editing mode on the right channel by the Tab key  


Penikov:
**Describe the bug**
In the case when a user tries to exit the name editing mode by pressing the Tab key, nothing happens with the right channel.

**To Reproduce**
Steps to reproduce the behaviour:
1. Add stereo track and a clip to it
2. Split the track
3. Rename the clip on the right channel
4. Make stereo again
5. Edit the right-channel clip name
6. press the Tab key

**Expected behaviour**
The user can exit the editing mode with the Tab.

**Screenshots**

https://user-images.githubusercontent.com/25851359/136783448-fe36bd9d-c4a7-4519-a94f-5e8c8502765d.mp4

**Additional information (please complete the following information):**
 - OS: Windows 10
 - Version: Audacity 3.1.0-alpha-20211011

PR #1938, opened 2021-10-15
Fix tab usage in wavetrack

vsverchinsky:
Resolves: #1884 

Recommended:


Paul-Licameli:
My review is complete.  I don’t approve yet.  I’m not sure how I tell GitHub to retract my mistaken approval.

Paul-Licameli:
I approve, understanding that another issue will be opened for the problem of the tab key cycle for the right channel in case clips do not align.

Issue #1822, opened 2021-10-05
Loop play crashes when reaching the end of locked Timeline region


petersampsonaudacity:
**Describe the bug**
Loop play crashes when reaching the end of locked Timeline region.

This is a regression on 3.0.5

**To Reproduce**
1. get some audio
2. select a small section for looping
3. Lock the Timeline range of the selection
3.1 either with the new loop button in the Transport toolbar, or
3.2 using the "Lock Play Region" from the Timeline's right-click context menu
4. Hold shift down
5. Observe Play button changes to Loop-play button
6. Press that button (or press the space-bar)  to initiate loop play
7. Observe: when the end of the locked loop-play region is reached Audacity crashes

**Expected behavior**
On reaching the end of the locked loop-play region Audacity should return to the beginning of the locked loop-play region and play the loop.

**Screenshots**

**Additional information (please complete the following information):**
 - Windows 10 - but assume all OS
 - Version [e.g. Audacity 3.1.0 master alpha:  Audacity_Windows_64bit_1305813991_5b09637 

**Additional context**
@Tantacrul and @jounih 
I can't for the life of me figure out the point of the new loop button in the Transport toolbar:
![image](https://user-images.githubusercontent.com/50205984/135988671-5ab813a0-d67a-4040-8f4a-f6646d47ce30.png)

a) For a start it has no hovertext help to indicate its usage (unlike all the other buttons in the transport toolbar.

b) all it appears to do is lock and unlock the timeline play region - thus just duplicating the long standing "Lock Play Region" from the Timeline's right-click context menu - and thus an unnecessary duplication.

c) from the look of it, plus the fact that it is in the Transport toolbar, one would intuit (I certainly did anyway) that pressing it would cause loop play to start and stop.

d) Worse still when you hold down the Shift key you get two identical looking buttons in the Transport toolbar - but they both do different things:
![image](https://user-images.githubusercontent.com/50205984/135990128-dc92b28f-3a58-47a2-979e-8f7afa4676da.png)

petersampsonaudacity:
I cannot reproduce this crash when I testing with the latest 3.1.0 alpha from today audacity-win-3.1.0-alpha-20211005+39b3048-64bit

PR #1965, opened 2021-10-19
Fix crash on looping play seen on Windows...

Paul-Licameli:
Resolves: #1822

... A recurrence of bugs 1887 and 2025.  These worked around a bug in the soxr
library.  Recent rearrangements of playback engine sometimes failed to do the
workaround.

Recommended:

### DIFF
--- a/src/Mix.cpp
+++ b/src/Mix.cpp
@@ -725,6 +725,9 @@ double Mixer::MixGetCurrentTime()
    return mTime;
 }
 
+#if 0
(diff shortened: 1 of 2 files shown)


Paul-Licameli:
> However, it happens without looping as well. Looking into the fix makes me not so sure, that this will help. We need to wait for the AppImage to be built first, as the bug is much easier to catch in Release build

If the call stack doesn't involve the soxr library, it's something else

Issue #1860, opened 2021-10-07
Smart Clips. A clip may collapse and become invisible if a user drags clip borders too much


Penikov:
**Describe the bug**
If the user works at sample-level scales, all goes well. The clip has borders and is still visible even when the 'virtual range' of a clip changed to one sample.

But when the user changes track length at a lower scale, clip borders tend to disappear. At some scales the clip becomes invisible.

![изображение](https://user-images.githubusercontent.com/25851359/136449462-3b1f57ca-4330-4756-b919-2ecf9442b28d.png)

![изображение](https://user-images.githubusercontent.com/25851359/136449539-a455cd30-0ec8-40da-be82-b3ca73990891.png)

How it works in dynamic:

https://user-images.githubusercontent.com/25851359/136450715-8257c138-fc23-4f5c-a132-eda320fab5fc.mp4

**Steps To Reproduce**

1. Make a clip
2. Set a track scale to any "ordinary" level. 
3. Pull any clip border to the opposite one until the clip becomes collapsed.

**Expected behaviour**

There is not a simple design issue, but I think the same behaviour is appropriate to Audacity:

https://user-images.githubusercontent.com/25851359/136452000-f6cab997-a29d-43b9-83c0-accb50879093.mp4

You see, the UI in the video has some bugs too, but this concept of the clip shrinking looks good to me.

 - Version Audacity 3.1.0-alpha-20211007

Penikov:
@Tantacrul what do you think?

Tantacrul:
Completely agree with expected behaviour.

PR #1981, opened 2021-10-20
Clips visibility

vsverchinsky:
Resolves: #1860 

Recommended:


Paul-Licameli:
Please do an interactive rebase to permute "Unified hit testing routines for Wave clips" to the start.  Then every commit in the sequence will compile.

PR #2038, opened 2021-10-28
Image compiler

Paul-Licameli:
Resolves: Changes of colors and images being much harder than it should be

Automate the generation of the *ThemeAsCeeCode.h headers from .png files and Colors.txt files.

Build a small executable that shares lib-theme and executes at build time to regenerate the header files when they are out of date with the images.

Thus, we can just check in the images.

Also a recovered history of the images implied in the history of the *ThemeAsCeeCode.h files, for reference.

Recommended:


JamesCrook (approved):
Well done on getting this to work.
Nice that it does not require any additional external dependencies (or, shudder, checking in an exe)

Paul-Licameli:
> Well done on getting this to work.
> Nice that it does not require any additional external dependencies (or, shudder, checking in an exe)

Some testing on Windows today tells me I’m not home yet.  The four themes are not available.  The static objects in lib-theme-resources that register the four themes never get constructed.  Either that library should have an initializer function called at startup, or it should be a module instead of a library, or just moved back into the main executable.

I will decide on one of those things and force push at least once more.

Paul-Licameli:
> > Well done on getting this to work.
> > Nice that it does not require any additional external dependencies (or, shudder, checking in an exe)
> 
> Some testing on Windows today tells me I’m not home yet. The four themes are not available. The static objects in lib-theme-resources that register the four themes never get constructed. Either that library should have an initializer function called at startup, or it should be a module instead of a library, or just moved back into the main executable.
> 
> I will decide on one of those things and force push at least once more.

There is a new little .exe of course, but in source form, built and run as part of the procedure for building the application, and then even deleted after use, although that deletion doesn’t always work.  Rerunning Audacity in the VS debugger rebuilds the image compiler (but that takes very little time) and doesn’t destroy it (but that takes only tens of kilobytes).

Paul-Licameli:
I expect one more chapter in the story of theme system rewrite.  That will allow the breaking-up of AllThemeResources.h into multiple pieces, and the invocation of the image compiler correspondingly multiple times in the build.  The set of theme resources, as I have said, should be open-ended.  It should become a true registry.  The definition of images required only for the Edit Toolbar, to take an example, should be closely associated with the code implementing the toolbar, and not in the low level Theme library.

Paul-Licameli:
I made one more fix, in the first commit in the sequence, and verified that theme switching on Windows isn't broken now.

I think it's really really ready for merge now!

crsib:
The interesting question is how we can help the themes creators. Probably we should rethink the system from the ground up.

src/CMakeLists.txt
```diff
@@ -1292,6 +1292,18 @@ else()
    target_link_libraries(findlib ${CMAKE_DL_LIBS})
 endif()
 
+# Common to all platforms: remove the imageCompiler program so it isn't needlessly
+# copied into program bundles
+if (TARGET image-compiler)
```
crsib, line 1297: I think we should just put this executable in a different location. This way it won't be installed on Windows/macOS and there will be no need to rebuild it constantly.

Paul-Licameli:
> The interesting question is how we can help the themes creators. Probably we should rethink the system from the ground up.

The how-to is here:  https://wiki.audacityteam.org/wiki/Theming_HowTo#Image_Cache

Note  `Create a folder called "Theme" (without the quotes) in Audacity's folder for application data then place your modified .png file in there.`

I think it would be more convenient for creators of themes to have another option, to put the un-bundled .png files inside a certain folder instead.  Then, Audacity can look for those instead of the monolithic .png file.

Each way of defining a custom theme has advantages and disadvantages, but we can give the choice.

I have not yet done the work to realize this, but it is would not be a "ground-up" rewrite of things, I think.

@JamesCrook @LWinterberg 

For the task at hand, merging this into master:  is your objection blocking or not, @crsib?

crsib:
> is your objection blocking or not

Probably let's just move the image compile to a different location, so there is no need to delete it. On the other hand, this is probably not the most straightforward task on Windows

Paul-Licameli:
> > is your objection blocking or not
> 
> Probably let's just move the image compile to a different location, so there is no need to delete it. On the other hand, this is probably not the most straightforward task on Windows

Do you remember all the irritating fuss there is to make Audacity properly find its dynamic libraries, on each of the platforms?  There was similar irritation to get the image compiler program to run, and the easiest shortcut solution was to put it in that folder.

Paul-Licameli:
> > > is your objection blocking or not
> > 
> > 
> > Probably let's just move the image compile to a different location, so there is no need to delete it. On the other hand, this is probably not the most straightforward task on Windows
> 
> Do you remember all the irritating fuss there is to make Audacity properly find its dynamic libraries, on each of the platforms? There was similar irritation to get the image compiler program to run, and the easiest shortcut solution was to put it in that folder.

I was debugging my image compiler on Windows and got it to work.  Now I'm trying it out on Mac, and I find in debug build at least that there are crashes on startup of Audacity, after modifying images and rebuilding.  A message on stderr says the dynamic library loader can't find lib-exceptions!  I have to do clean rebuild.

That's weird and might be an annoyance for developers.  I don't know yet how to fix it.  But this branch does pass the CI builds, which are from clean state.

crsib:
I can help you with that, if you don't mind!

Paul-Licameli:
> I can help you with that, if you don't mind!

Debug build crashes this way, even restarting the program in the debugger, with no changes in sources or images.

crsib (approved):

Issue #2061, opened 2021-11-01
Using the Time shift left/right commands crashes Audacity


DavidBailes:
**Describe the bug**
Using the Time shift left/right commands crashes Audacity

**To Reproduce**
1. Create an audio track containing a clip
2. Extra > Edit > Time shift left/right
3. Audacity crashes

**Expected behavior**
A time shift rather than a crash

**Additional information (please complete the following information):**
- Windows 10
- Audacity 3.1.0

**Additional context**
Reported on the Audacity4Blind list

buaabyl:
Maybe the problem is the format of config file changed?

When I upgrade from previous version today, UI crash...

## details:

I am using v2.3.3, and upgrade to 3.1.0 and 3.1.3 beta1.

1. open audacity
2. drag audio file in
3. move the Horizontal scrollbar: mouse roll wheel + hold shift-key
4. the UI hang (no response).

After I delete `C:/Users/ww/AppData/Roaming/audacity`, 
Reinstall 3.1.3 beta1, everything works fine.

PR #2063, opened 2021-11-01
Fixes clip moving with Time Shift command

crsib:
Resolves: #2061 

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
@@ -25,9 +25,23 @@ public:
       auto pClip = [&]() {
          for (auto clip : mpTrack->GetClips())
          {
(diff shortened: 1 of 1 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
```diff
@@ -25,9 +25,23 @@ class WaveTrackShifter final : public TrackShifter {
...
+                      *clip, viewInfo, params->rect,
+                      { params->xx, params->yy }))
+                  return clip;
+            }
+            else
+            {
+               // WithinPlayRegion misses first sample, which breaks moving
+               // "selected" clip. Probable WithinPlayRegion should be fixed
+               // instead?
+               if (clip->GetPlayStartTime() <= time && time < clip->GetPlayEndTime())
```
vsverchinsky, line 40: The condition for comparing clip boundaries used earlier can be found here: https://github.com/audacity/audacity/commit/69681f84d5f2943e6a5c1361bb3f0a10dcf2de24
but I do not see much difference, since the original condition has already changed

Issue #2051, opened 2021-10-29
Working with a large number of clips became sluggish


gpapp:
Affects version: 3.1

Steps to reproduce:
 - Import an hour long clip
 - trim quiet times with a noise gate (to make sure there is plenty of data to discard)
 - detach at silences

Expected result:
 - clips on each track are separated, several hundred of them are present
 - Smaller file sizes (after compacting the project)

Actual result 
 - crash when trying to process 2 one hour tracks
 - when the tracks are processed separately, they take about 5-10 times longer than before (which was slow to start with) 
 - any attempt to move a track takes minutes
 - closing the file and compacting takes minutes
 - opening the resulting file takes minutes

I'm using a workstation with 32GB RAM, and most of it is unused, so that's not an issue.

With version 3.0.x and before, this worked great, the clips created were moved freely. The file sizes decreased greatly, and editing was easy.

**UPD** by Penikov:

The crash issue has been added in separated bug: #2072

LWinterberg:
Can you try to do this one more time and look at your project file size, and watch your hard drive utilization when editing?

Audacity 3.1 supports non-destructive trimming of clips. Detach at silences creates a full copy of the track for each clip, so you in theory can undo it. I fear this causes files so large, your hard drive becomes a massive bottleneck.

gpapp:
It doesn't really do anything with my disk (SSD), it only hogs the processor (one of the four Xeons cores).

I understand the concept of non-destructive trimming in general, however, what you described is the complete opposite, of what I expect to do when detaching at silences. The whole idea of doing so, was to remove the useless and void data from the tracks permanently, so I could deal with sounds that are actually meaningful. Why would anyone want to keep copies of null data, really?

LWinterberg:
Yeah, it's silly. This definitely is an oversight.

gpapp:
Thanks for the confirmation, I was really worried I got it wrong somehow. 

Good news: If I'm opening a project coming from the earlier version, with so many fragments, there is no slugishness. I was worried that it is the number of clips, that slows it down.

This however serves as a test to see how the new non-destructive trimming adversely affects performance. I guess it is a good indicator, that there may be a room for improvement in the future.

However I would like to mention that I love the idea of this new way of trimming, as it was always an issue, to get some cuts just right!

Thank you all for the great work, that is required to keep this magnificent project alive!

Penikov:
@gpapp Thank you for your report!

PR #2065, opened 2021-11-01
Improve the performance of project serialization by ~10x

crsib:
Originally, wxMemoryBuffer was used while generating binary XML data. wxMemoryBuffer grows only by 1Kb, which results in extremely low performance of project serialization (few seconds on Ryzen 5800x for a 100Mb XML file).

This commit introduces a MemoryStream class tuned for high performance and low memory fragmentation.

This issue is related to #2051 but is definitely not enough to fix it.

Recommended:

### DIFF
--- a/libraries/lib-utility/CMakeLists.txt
+++ b/libraries/lib-utility/CMakeLists.txt
@@ -20,6 +20,8 @@ set( SOURCES
    MemoryX.h
    ModuleConstants.cpp
    ModuleConstants.h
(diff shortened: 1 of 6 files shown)


vsverchinsky (approved):
Great improvement overall, though MemoryStream might seem to be a very generic title for solving quite a specific issue, isn't it? I'd suggest moving it under the ProjectSerializer scope.

Issue #1462, opened 2021-08-12
PortMixer uses deprecated APIs on macOS


crsib:
PortMixer uses long-deprecated `AudioDeviceGetPropertyInfo` in its CoreAudio implementation

PR #2070, opened 2021-11-01
Remove calls to deprecated CoreAudio API

l-tms:
Remove calls to deprecated CoreAudio API

Resolves: audacity/audacity#1462

I changed all the deprecated CoreAudio API calls to use the current active ones.

Recommended:

### DIFF
--- a/lib-src/portmixer/src/px_mac_coreaudio.c
+++ b/lib-src/portmixer/src/px_mac_coreaudio.c
@@ -68,7 +68,6 @@ int OpenMixer_Mac_CoreAudio(px_mixer *Px, int index)
    PxInfo   *info;
    OSStatus err;
    UInt32   outSize;
-   UInt32   inID;
    int      i;
 
    if (!initialize(Px)) {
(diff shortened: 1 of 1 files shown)


lib-src/portmixer/src/px_mac_coreaudio.c
```diff
@@ -523,14 +543,20 @@ static PxVolume get_play_through(px_mixer *Px)
    OSStatus err;
    UInt32   outSize;
    UInt32   flag;
+   AudioObjectPropertyAddress inAddress = {
+      kAudioDevicePropertyPlayThru,
+      // should we use ...ScopePlayThrough, ...ScopeGlobal or ...ScopeInput instead?
```
crsib, line 548: Gosh I live Apple documentation :-)

https://developer.apple.com/documentation/coreaudio/1494464-anonymous/kaudioobjectpropertyscopeplaythrough?language=objc

This is actually an excellent question of which scopes to use here. Have you found a way to check this under the debugger?

l-tms: I checked in the debugger the output for all new calls I introduced, comparing them to the old call version, but for this I never managed to get back anything useful. 

Even with the old version I got back `OSStatus 2003332927 kAudioHardwareUnknownPropertyError` every time, with both the MacBook builtin and external soundcard I have at home, so I assumed I was just missing hardware that would support this.

I used `ScopeOutput` in the end as the original code was also using that (`IS_OUTPUT`)

crsib: I think that we need to use PlayThrough scope

Some other guys do at least xD

https://github.com/itsthejb/HotPants/blob/master/CoreAudio/HAL/HPBase/HP_Device.cpp

l-tms: fair enough, I updated this line to use `kAudioDevicePropertyScopePlayThrough` :)

crsib (approved):

crsib:
Thank you a lot!

Issue #2072, opened 2021-11-02
Detach at silences. Audacity crashes with clips that trimmed and overlapped


Penikov:
**Describe the bug**
Audacity crashes with clips that trimmed and overlapped

**To Reproduce**
Steps to reproduce the behaviour:
1. Generate a clip
2. Select the clip this way:
![изображение](https://user-images.githubusercontent.com/25851359/139837434-e322f4db-622b-4796-a1f4-f6504a67fce3.png)

3. Apply Noise gate effect to the selection
4. Try to apply Edit-Clip boundaries - Detach at silences

**Expected behaviour**
All goes well, nothing crashes. 

**Actual behaviour:**
Audacity crashes when you try to apply Detach after Noise gate effect

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Windows, Mac (just raises a message about internal error)
 - Version [e.g. Audacity 3.1 +
 
**Additional context**
**UPD** by Penikov:
Just a little of details:
In addition, the bug is reproducing on Windows: Audacity crashes when I tried to apply Detach Silence _after_ the Noise Gate effect
and on Mac Audacity raises an error:

![изображение](https://user-images.githubusercontent.com/25851359/139723984-64013fc8-2470-4b7a-9818-7830b7f1bfa3.png)

I think the root cause of poor performance is duplicating the whole selection in every clip on the track. 

In the video, I trying to show how Noise Gate works with many clips. 

https://user-images.githubusercontent.com/25851359/139725360-a3ef5ef2-c075-49c4-a8af-0472c50a9278.mp4

It is an overwhelming overhead. Also, it crashes audacity when Detach Silence tries to operate with trimmed clips.

And the combination of Noise Gate and Detwch Silence can crash Audacity even with one single clip in a project.

relates to #2051

PR #2074, opened 2021-11-02
Fixes double free with InconsistencyException on Windows

crsib:
C++ standard actually requires an exception to be copyable.  http://eel.is/c++draft/except.throw#5

However, it only affects MSVC implementation of the exceptions, so it was unnoticed for quite some time

This PR is only a partial fix for the symptoms of #2072 

Recommended:

### DIFF
--- a/libraries/lib-exceptions/InconsistencyException.h
+++ b/libraries/lib-exceptions/InconsistencyException.h
@@ -37,8 +37,8 @@ public:
        , func { fn }, file { f }, line { l }
    {}
 
-   InconsistencyException(InconsistencyException&& that)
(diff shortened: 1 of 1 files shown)


Paul-Licameli:
See this commit?  I neglected this class!  You complete my work.  4a56af43aa4833ba6f6ee96911797170cebf3dd6

Paul-Licameli (approved):

Paul-Licameli:
And a link for your convenience to discussion of that other bug, which I have all forgotton by now.
https://bugzilla.audacityteam.org/show_bug.cgi?id=2442

PR #2087, opened 2021-11-03
Deserialization performance

crsib:
A first attempt to improve the loading times for large projects. This improves both performance and memory usage by around 2x.

46 Mb binary produces 75 Mb XML stream, which requires around 90 Mb of RAM and 5 seconds to load. Previously, RAM overhead was ~380 Mb, the loading time was 11 seconds.

The peak memory usage is now 221 Mb down from ~580. Audacity uses 134 Mb right after the project is loaded.

XML processing is now responsible for 78% of that time:
image

Half of the time is spent in "StartElement"
image

26% of the time is spent on constructing wxString objects.

Recommended:

### DIFF
--- a/libraries/lib-string-utils/ToChars.h
+++ b/libraries/lib-string-utils/ToChars.h
@@ -1,0 +1,35 @@
+/*!********************************************************************
+
+ Audacity: A Digital Audio Editor
+
+ @file ToChars.h
+ @brief Declare functions to convert numeric types to string representation.
+
+ Dmitry Vedenko
+ **********************************************************************/
+
+#pragma once
+
+#include <system_error>
+
+//! Result of the conversion, similar to std::to_chars_result
+struct STRING_UTILS_API ToCharsResult final
+{
+   char* ptr; //! A pointer to the last updated character, or last on failure
+   std::errc ec; //! std::errc::value_too_large on failure, std::errc() on success
+};
+
+//! Convert a single precision floating point number to a string, always uses the dot as decimal
+STRING_UTILS_API ToCharsResult ToChars(
+   char* buffer, char* last, float value,
+   int digitsAfterDecimalPoint = -1) noexcept;
+
+//! Convert a double precision floating point number to a string, always uses the dot as decimal
+STRING_UTILS_API ToCharsResult ToChars(
+   char* buffer, char* last, double value,
+   int digitsAfterDecimalPoint = -1) noexcept;
+
+//! Convert a signed 64 bit integer value to string
+STRING_UTILS_API ToCharsResult ToChars(char* buffer, char* last, long long value) noexcept;
+//! Convert a unsigned 64 bit integer value to string
+STRING_UTILS_API ToCharsResult ToChars(char* buffer, char* last, unsigned long long value) noexcept;
--- a/libraries/lib-utility/MemoryStream.h
+++ b/libraries/lib-utility/MemoryStream.h
@@ -15,13 +15,15 @@
 #include <list>
 #include <vector>
 
+#include "MemoryX.h"
+
 /*!
  * @brief A low overhead memory stream with O(1) append, low heap fragmentation and a linear memory view.
  *
  * wxMemoryBuffer always appends 1Kb to the end of the buffer, causing severe performance issues
- * and significant heap fragmentation. There is no possibility to controll the increment value.
+ * and significant heap fragmentation. There is no possibility to control the increment value.
  *
- * std::vector doubles it's memory size which can be problematic for large projects as well.
+ * std::vector doubles its memory size which can be problematic for large projects as well.
  * Not as bad as wxMemoryBuffer though.
  * 
  */
@@ -29,6 +31,29 @@ class UTILITY_API MemoryStream final
 {
 public:
    using StreamData = std::vector<uint8_t>;
+   using StreamChunk = std::pair<const void*, size_t>;
+
+private:
+   static constexpr size_t ChunkSize =
+      1024 * 1024 - // 1Mb
+      2 * sizeof(void*) - // account for the list node pointers
+      sizeof(size_t); // account for the bytes used member
+
+   struct Chunk final
+   {
+      std::array<uint8_t, ChunkSize> Data;
+      size_t BytesUsed { 0 };
+
+      // Returns data size left to append
+      size_t Append(StreamChunk& dataView);
+   };
+
+   using ChunksList = std::list<Chunk>;
+
+public:
+
+   MemoryStream() = default;
+   MemoryStream(MemoryStream&&) = default;
 
    void Clear();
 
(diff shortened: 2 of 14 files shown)


libraries/lib-xml/XMLWriter.cpp
```diff
@@ -30,7 +30,10 @@ the general functionality for creating XML in UTF8 encoding.
 #include <wx/ffile.h>
 #include <wx/intl.h>
 
-#include <string.h>
+#include <cstring>
+#include <charconv>
```
Paul-Licameli, line 34: As you now know, <charconv> is incomplete for macOS and not even known in the Linux build.  Sorry, you must find an alternative to to_chars.

Paul-Licameli (changes requested):

libraries/lib-utility/MemoryStream.h
```diff
@@ -29,6 +31,26 @@ class UTILITY_API MemoryStream final
 {
 public:
    using StreamData = std::vector<uint8_t>;
+   using StreamChunk = std::pair<const void*, size_t>;
+
+private:
+   static constexpr size_t ChunkSize = 1024 * 1024;
```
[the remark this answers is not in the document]
crsib: Yeah, makes total sense, thank you!

libraries/lib-xml/XMLWriter.cpp
crsib: I'm painfully aware of it now. This is really sad that a very simple part of C++17 is not implemented fully in 2021

libraries/lib-utility/MemoryStream.h
```diff
@@ -15,20 +15,45 @@
...
 {
 public:
    using StreamData = std::vector<uint8_t>;
+   using StreamChunk = std::pair<const void*, size_t>;
+
+private:
+   static constexpr size_t ChunkSize =
+      1024 * 1024 - // 1Mb
+      2 * sizeof(void*) - // account for the list node pointers
+      sizeof(size_t); // account for the bytes used member
```
Paul-Licameli, line 40: Maybe subtract a little more for operator new overhead.  Make a guess what that is.

libraries/lib-string-utils/ToChars.h
```diff
@@ -0,0 +1,35 @@
...
+
+#pragma once
+
+#include <system_error>
+
+//! Result of the conversion, similar to std::to_chars_result
+struct STRING_UTILS_API ToCharsResult final
+{
+   char* ptr; //! A pointer to the last updated character, or last on failure
+   std::errc ec; //! std::errc::value_too_larg on failure, std::errc() on success
```
Paul-Licameli, line 19: Large

libraries/lib-utility/MemoryStream.h
crsib: I saw no measurable difference here, to be honest. If CRT decides that it will allocate page size aligned amount of memory - we will have an overhead of 4K (less than 1% percent overhead in this case) and will be likely able to reuse it. If CRT fallbacks to some `mmap` kind of API - there is no overhead that we can control here. Controlling the overhead from std::allocator is more feasible, but I'm not sure if we really need such control. (And the simplest way is to simply use a handwritten allocator)

Anyway, the current bottleneck is very far away from this place both for serialization and deserialization. A better fix would be to perform incremental IO directly from the database, but SQLite3 interface doesn't really have an incremental write API. We can avoid memory "linearization" though because `sqlite3_bind_zeroblob` is a very fast API (not actually committing any pages, rather just setting a propper header)

Paul-Licameli: Shrug.  Whatever easy little help we can get.  Yes, measure first.

src/ProjectSerializer.cpp
```diff
@@ -394,45 +395,45 @@ wxString ProjectSerializer::Decode(const wxMemoryBuffer &buffer)
    mIds.clear();
 
    struct Error{}; // exception type for short-range try/catch
-   auto Lookup = [&mIds]( UShort id ) -> const wxString &
+   auto Lookup = [&mIds]( UShort id ) -> std::string_view
```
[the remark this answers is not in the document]
Paul-Licameli: FT_Push and FT_Pop never happen, in fact!

Paul-Licameli:
I have reviewed everything but the XMLUtf8BufferWriter class and have no objections to those parts.

I assume the ToChars implementation is just a great cut and paste of trusted code.  Or are there any differences from your source I should know about?

crsib:
> I assume the ToChars implementation is just a great cut and paste of trusted code. Or are there any differences from your source I should know about?

The only major difference is that I have added buffer size checks to Grisu2, which always blindly assumed that "the buffer is big enough"

Paul-Licameli:
You don't yet use sqlite3_bind_zeroblob.  Do you mean to?

This would allow skipping MemoryStream::GetData() which is one huge allocation and memory movement?

Why is the blob handle interface "not really" the incremental write we need?

Are there other opportunities for performance improvementh you have seen but not yet implemented in this PR?

libraries/lib-xml/XMLWriter.cpp
```diff
@@ -428,3 +432,183 @@ void XMLStringWriter::Write(const wxString &data)
...
+   // Well, that maintains the original behavior
+   WriteAttr(name, static_cast<long long>(value));
+}
+
+void XMLUtf8BufferWriter::WriteAttr(
+   const std::string_view& name, float value, int digits /*= -1*/)
+{
+   constexpr size_t bufferSize = std::numeric_limits<float>::max_digits10 +
+                                 5 + // No constexpr log2 yet! example - e-308
+                                 3; // Dot, sing an 0 separator
```
Paul-Licameli, line 523: Sign

libraries/lib-xml/XMLWriter.cpp
```diff
@@ -428,3 +432,183 @@ void XMLStringWriter::Write(const wxString &data)
...
+
+   WriteAttr(name, std::string_view(buffer, result.ptr - buffer));
+}
+
+void XMLUtf8BufferWriter::WriteAttr(
+   const std::string_view& name, double value, int digits /*= -1*/)
+{
+   constexpr size_t bufferSize = std::numeric_limits<double>::max_digits10 +
+                                 5 + // No constexpr log2 yet!
+                                 3;  // Dot, sing an 0 separator
```
Paul-Licameli, line 540: Sign

libraries/lib-xml/XMLWriter.cpp
```diff
@@ -428,3 +432,183 @@ void XMLStringWriter::Write(const wxString &data)
...
+   if (mInTag)
+   {
+      Write(">");
+      mInTag = false;
+   }
+
+   WriteEscaped(value);
+}
+
+void XMLUtf8BufferWriter::WriteSubTree(const std::string_view& value)
```
Paul-Licameli, line 563: I think this isn’t used

Paul-Licameli (approved):
Comment typos

Paul-Licameli:
Approved, let’s move

libraries/lib-xml/XMLWriter.cpp
crsib: I've just copied the "Writer" interface, but I agree that this can be dropped.

Paul-Licameli:
https://github.com/audacity/audacity/pull/2087#discussion_r745510411

crsib:
>But we can know the blob size before we write it?
>We still capture data in chunks and move it into the blob

That is what I plan to do in  a next PR

Issue #1889, opened 2021-10-11
Clip names corrupted with some effects


robmcmullen:
**Describe the bug**
Using some effects (Noise Gate, Noise Reduction) changes the clip name to match the previous clip. Noise Gate also changes the name of the following clip

**To Reproduce**
1. Need sample file with multiple cips
2. Click on clip name to highlight entire clip
3. Choose effect from menu (Noise Gate, Noise Reduction)

**Expected behavior**
Selected clip should not change name

**Screenshots**

Noise reduction: changes the name of the selected clip to the name of the previous clip
![noise-reduction](https://user-images.githubusercontent.com/896656/136821336-f132eef0-c0a3-4351-a82a-92a65c1169e8.gif)

Noise gate: change the name of the selected clip to the name of the previous clip **and** changes the name of the following clip to the name of the first clip
![noise-gate](https://user-images.githubusercontent.com/896656/136821519-a6475db7-3ffa-477f-983a-e4d4c4008721.gif)

**Additional information (please complete the following information):**
 - OS: Linux Mint 20.2
 - Version: Audacity 3.1.0-alpha-20211010

PR #2088, opened 2021-11-03
Issue 1889: fix clip name corrupt

### LINKED ISSUE #2051: Working with a large number of clips became sluggish

vsverchinsky:
Resolves: #1889

Also partly resolves #2051: applying the Nyquist effect to a range that contained splits resulted in multiple copies of the same data received from the effect being created but with different offsets and trims, and, at the same time, old trimmed data was lost. This fix is aimed to resolve both issues: preserve old trimmed data at split points and eliminate data duplication.

Recommended:

### DIFF
--- a/src/WaveTrack.cpp
+++ b/src/WaveTrack.cpp
@@ -40,6 +40,7 @@ from the project that will own the track.
 #include <float.h>
 #include <math.h>
 #include <algorithm>
+#include <optional>
 
 #include "float_cast.h"
 
@@ -836,6 +837,34 @@ void WaveTrack::SetWaveformSettings(std::unique_ptr<WaveformSettings> &&pSetting
    }
 }
 
+namespace {
+   
+   //Internal structure, which is supposed to contain
+   //data related to clip boundaries, and used during 
+   //ClearAndPaste to restore old splits after new
+   //data is pasted
+   struct SplitInfo
+   {
+      //Time, where boundary is located
+      double time;
+      //Contains trimmed data, which should be re-appended to
+      //the clip to the left from the boundary, may be null
+      std::shared_ptr<WaveClip> left;
+      //Contains trimmed data, which should be re-appended to
+      //the clip to the right from the boundary, may be null
+      std::shared_ptr<WaveClip> right;
+      //Contains clip name next to the left from the boundary,
+      //if present, that needs to be re-assigned to the matching
+      //clip after split
+      std::optional<wxString> leftClipName;
+      //Contains clip name next to the right from the boundary,
+      //if present, that needs to be re-assigned to the matching
+      //clip after split
+      std::optional<wxString> rightClipName;
+   };
+
+}
+
 //
 // ClearAndPaste() is a specialized version of HandleClear()
 // followed by Paste() and is used mostly by effects that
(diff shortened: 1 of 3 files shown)


src/WaveTrack.cpp
```diff
@@ -981,9 +1035,73 @@ void WaveTrack::ClearAndPaste(double t0, // Start of time to clear
...
+         for (const auto& split: splits) {
+            auto at = LongSamplesToTime(TimeToLongSamples(warper->Warp(split.time)));
+            for (const auto& clip : GetClips())
+            {
+               if (clip->WithinPlayRegion(at))//strictly inside
+               {
+                  auto newClip = std::make_unique<WaveClip>(*clip, mpFactory, true);
+
+                  clip->ClearRight(at);
+                  newClip->ClearLeft(at);
```
Paul-Licameli, line 1073: I think the two lines above should be conditional.  Clear old trim data, only if there is something to attach.  Otherwise, you make this behave unlike WaveClip::Split which preserves all sample data in each new clip at the split.

Am I wrong?

Paul-Licameli (changes requested):

src/WaveTrack.cpp
```diff
@@ -981,9 +1035,73 @@ void WaveTrack::ClearAndPaste(double t0, // Start of time to clear
       // Restore cut/split lines
       if (preserve) {
 
-         // Restore the split lines, transforming the position appropriately
-         for (const auto split: splits) {
-            SplitAt(warper->Warp(split));
+         auto attachLeft = [](WaveClip* target, WaveClip* src)
```
Paul-Licameli, line 1038: Making the second clip const would be nice, and also in the other lambda following

Paul-Licameli:
I'm done with this review

src/WaveTrack.cpp
vsverchinsky: I don't think you are wrong, because this question is more about desired behavior. WaveTrack::Split creates a new clip copy with other trimmings and offset values, being used within generators/automated actions they can lead to an enormous amount of duplicates, which is not maintainable (right now). I think it's reasonable to preserve original tracks data hidden behind those splits, but not to introduce new hidden data.

Paul-Licameli (approved):
The question about whether trimmed data is eliminated in too many cases may be deferred.

PR #2106, opened 2021-11-09
Blob streaming

crsib:
This PR uses blob_open/blob_write APIs to avoid a potentially huge memory copy that happens in `MemoryStream::GetData`

Recommended:

### DIFF
--- a/src/ProjectFileIO.cpp
+++ b/src/ProjectFileIO.cpp
@@ -12,6 +12,8 @@ Paul Licameli split from AudacityProject.cpp
 
 #include <atomic>
 #include <sqlite3.h>
+#include <optional>
+
 #include <wx/app.h>
 #include <wx/crt.h>
 #include <wx/frame.h>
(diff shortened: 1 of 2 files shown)


src/ProjectFileIO.cpp
```diff
@@ -1757,38 +1835,121 @@ bool ProjectFileIO::WriteDoc(const char *table,
...
+      wxString::Format("SELECT ROWID FROM %s.%s WHERE id = 1;", schema, table);
+
+   if (!Query(
+          rowIDSql,
+          [&rowID](int cols, char** vals, char** names)
+          {
+             if (cols != 1)
+                return -1;
+
+             rowID = std::stoll(vals[0]);
```
Paul-Licameli, line 1892: Should rowID be long long?

src/ProjectFileIO.cpp
```diff
@@ -211,6 +213,80 @@ class SQLiteIniter
...
+{
+public:
+   static std::optional<SQLiteBlobStream> Open(
+      sqlite3* db, const char* schema, const char* table, const char* column,
+      int64_t rowID, bool readOnly) noexcept
+   {
+      if (db == nullptr)
+         return {};
+
+      sqlite3_blob* blob;
```
Paul-Licameli, line 226: I would initialize that to nullptr...

src/ProjectFileIO.cpp
```diff
@@ -211,6 +213,80 @@ class SQLiteIniter
...
+
+      const int rc = sqlite3_blob_open(
+         db, schema, table, column, rowID, readOnly ? 0 : 1, &blob);
```
Paul-Licameli, line 229: ... just in case SQLITE_MISUSE is the return value, see the docs https://www.sqlite.org/c3ref/blob_open.html

src/ProjectFileIO.cpp
```diff
@@ -1723,17 +1799,19 @@ bool ProjectFileIO::WriteDoc(const char *table,
                              const char *schema /* = "main" */)
 {
    auto db = DB();
+
+   TransactionScope transaction(GetConnection(), "UpdateProject");
```
Paul-Licameli, line 1803: You do this to avoid an autocommit for each chunk?  You discovered the need for it or you knew?  This page isn't explicit about it https://www.sqlite.org/c3ref/blob_write.html but autocommit is mentioned for blob close.

You might comment.

Paul-Licameli (approved):
Approved, with a few suggestions

src/ProjectFileIO.cpp
crsib: It's `int64_t` in SQLite. Usually, it's an alias to `long long` anyway. At least until we will have 128 bit `long long` :-)

src/ProjectFileIO.cpp
crsib: Won't hurt for sure!

src/ProjectFileIO.cpp
crsib: It's really confusing to be fair. Is it safe to close a misused blob?

>This means that, provided the API is not misused, it is always safe to call sqlite3_blob_close() on *ppBlob after this function it returns.

Looking into the source code - there is a single case where ppBlob will not be initialized and MISUSE is returned.

src/ProjectFileIO.cpp
crsib: I want to make sure that the project is written atomically. It was guaranteed previously that the project is saved atomically, so I kept this behavior.

Issue #2043, opened 2021-10-28
Audacity Custom Theme not working(Version 3.1.0)


AzathothWAR:
After updating Audacity to the latest version 28/10/2021 , 3.1.0. 
Custom theme has stopped working , tried to reinstall and redoing but still no luck.
I tried the "Audcaity Clean Dark Theme" from : https://forum.audacityteam.org/viewtopic.php?t=97814
as well as from a tutorial video on youtube : https://www.youtube.com/watch?v=mE0qkHSxgMM 
Theme link : https://forum.audacityteam.org/viewtopic.php?t=100222&__cf_chl_jschl_tk__=d3ca8372eaa5660055577df8732672e3d935cf9e-1591467249-0-AQndHCEcrmrUHt5tmiPYvIw3ZJJpIUVUaXta-qPSVzYLd3R4T2NAN6P4GLvkDEyUY4uJ_rHLy0h3DNMGvvXWTIaJFQsEVCOZ2H5rsJLBpTJZgZWE1-TEExIsUSliRjO-MoGoEFCqUcDJZ3wRay1fMWvVR6JmzeEWm8vhQVK4kvW8g67qf4qxQjp9sLXAJPgLXYG7CBH_e-UB-N-fRa2-R0c8mJvfz9ZTV0L2WaZYIYCGnRNoppZSa5yqecR3iogTArIBYEEsE2RFtdPLzuTdGWDJ0bxeAx4jlWTkt7tiEVGT9lgZlhjcaiGTvvEQbC_MfdhPLLIwawffy0-XiGRk7hfiRS30kH1vIaLDizVJseK-

I've made sure to follow the steps accordingly to make sure I have not made any mistakes in installing the custom theme.. 

Below are screenshots as to what it looks like while "custom theme" is selected.

*NOTE*
Only Custom themes aren't working , the other ones such as 'Classic' , 'Light' , 'Dark' , 'High Contrast' has NO issue.

![image](https://user-images.githubusercontent.com/93334561/139279931-2f15db68-2e73-42bc-b488-cae8e6f728c4.png)

![image](https://user-images.githubusercontent.com/93334561/139279968-a827e06a-d3cb-44b4-92f9-5564ceeac380.png)

And here is a screenshot of whats in my theme folder 
![image](https://user-images.githubusercontent.com/93334561/139280608-232a7ad3-35cf-43d9-815a-0800d96be943.png)

GhostShadow0316:
[...]

PR #2107, opened 2021-11-10
Fix ability to load custom theme, which was lost at commit b3cd7a57

Paul-Licameli:
Resolves: #2043

Recent theme rewrites mistakenly eliminated the ability to load custom themes from an image cache file.  One-line fix.

Recommended:


crsib (approved):

crsib:
That was easy indeed. Still, I think old themes wont work correctly

PR #2110, opened 2021-11-10
Define a substitute for std::visit.  Thanks, macOS :-(

Paul-Licameli:
Resolves: A development annoyance preventing some more elegant C++

std::variant and std::visit aren't really here yet for macOS while our minimum isn't up to 10.13.

But I won't let that stop me...

Recommended:


crsib (approved):

Paul-Licameli:
Thanks for approval.

To be really clever, I could implement a jump table.

"Exercise for the reader" as the math books say.

PR #2115, opened 2021-11-11
Effect cleanup part 1

Paul-Licameli:
Resolves: Some legacy code made less confusing

The loading and storing of persistent effect settings becomes less repetitious, with the use of variants and visitors, and merging of two sets of functions (for "private" and "shared" configurations) into one.

Class Effect has one fewer abstract base class, and defines fewer overrides, as we eliminate class ConfigClientInterface whose purpose wasn't clear.

Instead, plug-ins like VST effects can just use PluginManager more directly.

Two other abstract classes lifted out of libraries/lib-components (home of may interfaces with unclear purposes).  More may follow.

Recommended:


vsverchinsky (approved):

vsverchinsky:
The only thing I want to note about using std::varinat in that context is that extension of the type list may lead to a necessity to change the existing function bodies (`PluginManager::GetConfigValue`, `PluginManager::SetConfigValue`)

Paul-Licameli:
> The only thing I want to note about using std::varinat in that context is that extension of the type list may lead to a necessity to change the existing function bodies (`PluginManager::GetConfigValue`, `PluginManager::SetConfigValue`)

PluginManager::SetConfigValue just needs a generic lambda, and it compiles calls to different overloads of wxConfigBase::Write() (see confbase.h in wxWidgets sources), depending on what type `auto` is bound to.  It is very nice to see how this works to eliminate a lot of duplicated code.

See at line 311 of confbase that you can overload wxToString globally for new kinds of values -- so this generic lambda might still work unchanged in future, even with custom types, provided you add wxToString for the new type.  See Identifier.h for an example where I did this already.

PluginManager::GetConfigValue is a little more complicated, because I was being very conservative about preserving the special logic that existed for float, but a closer reading of wxWidgets documentation might prove that this caution is not needed.

See line 204 of confbase.h for use of wxFromString to customize Read() too for other types, and see Identifier.h again for an example.  So again, defining that with the new type will allow the generic lambda to continue working unchanged.

PR #2118, opened 2021-11-11
Effect cleanup part 2

Paul-Licameli:
Resolves: Restructure more legacy code related to effects

Better separation of responsibilities between the common "host" dialog in EffectUI.cpp, and the "client" plug-ins that only decorate a panel on the dialog.  The "client" can tell the host whether the dialog should be modal, but leave it to the host to create and destroy the dialog.

Also remove some special member variables of Effect that existed only for the use of Nyquist.  Instead the public member function iterfaces allow Nyquist to specify the special case behavior.

Recommended:

### DIFF
--- a/src/effects/NoiseReduction.cpp
+++ b/src/effects/NoiseReduction.cpp
@@ -446,7 +446,12 @@ bool EffectNoiseReduction::CheckWhetherSkipEffect()
    return false;
 }
 
-bool EffectNoiseReduction::ShowInterface(
+//! An override still here for historical reasons, ignoring the factory
+/*! We would like to make this effect behave more like others, but it does have
+ its unusual two-pass nature.  First choose and analyze an example of noise,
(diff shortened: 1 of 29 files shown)


src/effects/NoiseReduction.cpp
```diff
@@ -474,13 +474,16 @@ bool EffectNoiseReduction::Settings::PromptUser
    dlog.CentreOnParent();
    dlog.ShowModal();
 
-   if (dlog.GetReturnCode() == 0)
+   const auto returnCode = dlog.GetReturnCode();
+   if (!returnCode)
       return false;
```
vsverchinsky, line 479: I believe it should be int now

vsverchinsky (approved):

src/effects/NoiseReduction.cpp
Paul-Licameli: Thank you

Issue #2119, opened 2021-11-11
Data loss on joining clips


SteveDaulton:
**Describe the bug**
Joining clips in a track may cause data loss

**To Reproduce**
1. Generate 30 second chirp
2. Split at around 5 seconds
3. Split at around 20 seconds
4. Select across the second split (from around 18 to 22 seconds
5. Join the clips ("Ctrl + J")
6. Observe data loss in first clip.

**Expected behavior**
The selected split to join and no data loss

**Screenshots**

![First Track000](https://user-images.githubusercontent.com/11681097/141360530-1076abd9-895b-4e30-a687-8a825c15bd7a.png)

![First Track001](https://user-images.githubusercontent.com/11681097/141360583-42d05252-9c8b-4c5f-96d1-e1ce5986360e.png)

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04 and Windows 10
 - Version: Audacity 3.1.0 and 3.1.1

**Additional context**
Hot fix requested asap. On a large project the user may not notice the bug until it is too late to recover.

Penikov:
Confirm.

But there is a workaround with splitting with double-click on clip borders

DavidBailes:
> But there is a workaround with joining with double-click on clip borders

But not for keyboard users.

vsverchinsky:
It's not lost but rather hidden behind the new clip... it's still a bug anyway

Paul-Licameli:
> It's not lost but rather hidden behind the new clip... it's still a bug anyway

Hidden behind— overlapping?  Clips should never overlap.

PR #2123, opened 2021-11-12
issue #2119: Data loss on joining clips

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/2119

Joining clips using the Join command (Ctrl+J) can cause data in preceeding clips to be silenced.

Fix: In WaveTrack::Join(), the initial value of t should be PlayStartTime of the first clip, not SequenceStartTime.

Resolves: https://github.com/audacity/audacity/issues/2119

Recommended:

### DIFF
--- a/src/WaveTrack.cpp
+++ b/src/WaveTrack.cpp
@@ -1708,9 +1708,10 @@ void WaveTrack::Join(double t0, double t1)
    if( clipsToDelete.size() == 0 )
       return;
 
-   auto t = clipsToDelete[0]->GetSequenceStartTime();
+   auto t = clipsToDelete[0]->GetPlayStartTime();
    //preserve left trim data if any
-   newClip = CreateClip(t, clipsToDelete[0]->GetName());
+   newClip = CreateClip(clipsToDelete[0]->GetSequenceStartTime(),
+      clipsToDelete[0]->GetName());
    
    for (const auto &clip : clipsToDelete)
    {


src/WaveTrack.cpp
```diff
@@ -1708,9 +1708,10 @@ void WaveTrack::Join(double t0, double t1)
    if( clipsToDelete.size() == 0 )
       return;
 
-   auto t = clipsToDelete[0]->GetSequenceStartTime();
+   auto t = clipsToDelete[0]->GetPlayStartTime();
    //preserve left trim data if any
-   newClip = CreateClip(t, clipsToDelete[0]->GetName());
+   newClip = CreateClip(clipsToDelete[0]->GetSequenceStartTime(),
```
vsverchinsky, line 1713: A new clip has zero length, `WaveClip::Paste` then makes some checks to decide which data would be discarded, and in that case, the outcome depends on the order in which these cases are handled. So I guess I need to handle another special case in `WaveClip::Paste` so that pasted clip is fully preserved when the target clip is empty. Now if I make a selection strictly inside one clip, and then call join, then the hidden data on the right is lost.

vsverchinsky (approved):

src/WaveTrack.cpp
DavidBailes: > So I guess I need to handle another special case in `WaveClip::Paste` so that pasted clip is fully preserved when the target clip is empty. Now if I make a selection strictly inside one clip, and then call join, then the hidden data on the right is lost.

An alternative fix for the reported bug, would have been to do something similar to void WaveTrack::MergeClips().
That is, not to delete clipsToDelete[0], but to use that as the "newClip", and paste the rest (+silence where necessary). I think this would avoid pasting into an empty clip.

Thanks for approving the  commit!

Issue #2134, opened 2021-11-14
Encoding issue when using non-latin characters


gpapp:
**Describe the bug**
When opening a file with a label track made with earlier version (3.0.5), that contains non-latin-1 characters, the characters are lost with 3.1.1.

**To Reproduce**
1. Create a label track in 3.0.5
2. Create a label with "Árvíztűrő tükörfúrógép" as a text (all Hungarian diacritics) 
3. Open the file in 3.1.1
4. The label now says "rvztr tkrfrgp"

**Expected behavior**
Label should not change

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Windows 10
 - Version 3.1.1

gpapp:
Probably relates to #2132

Issue #2132, opened 2021-11-14
Project fails to open when clip handle contains multi-byte character


SteveDaulton:
**Describe the bug**
When a clip name contains a multi-byte character, the project cannot be opened

**To Reproduce**
1. Add an audio clip and name it 大
2. Save the project and quit
3. Try to open the project
4. Observe the error:

`{
    "timestamp": 1636894463,
    "event_id": "a95d4dd2ec54495d88f331960ec7827f",
    "platform": "native",
    "release": "audacity@3.2.0",
    "contexts": {
        "os": {
            "type": "os",
            "name": "Unix",
            "version": "5.11.0"
        }
    },
    "exception": {
        "values": [
            {
                "type": "Error_Opening_Project",
                "value": "Unable to parse project information.",
                "mechanism": {
                    "type": "runtime_error",
                    "handled": false,
                    "data": {
                        "log": "12:51:39: Audacity <path>-alpha-20211112\n12:51:39: sqlite3 message: (1) no such table: project in \"SELECT 1 FROM project LIMIT 1;\"\n12:54:09: sqlite3 message: (1) no such table: project in \"SELECT 1 FROM project LIMIT 1;\"\n12:54:23: ParseString error: Error: not well-formed (invalid token) at line 3\n===begin===(null)\n===end===\n12:54:23: DBConnection SetError\n\tErrorCode: 0\n\tLastError: Unable to parse project information.\n\tLibraryError: Error: not well-formed (invalid token) at line 3\n"
                    }
                }
            }
        ]
    }
}`

**Expected behavior**

**Screenshots**

**Additional information (please complete the following information):**
 - OS: [e.g. Mac High Sierra, Ubuntu 18.04, Windows 10]
 - Version [e.g. Audacity 2.5.1]
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

**Additional context**

Penikov:
@SteveDaulton Could you tell, please, what Audacity version has been crashing in your case?

SteveDaulton:
> @SteveDaulton Could you tell, please, what Audacity version has been crashing in your case?

"platform": "native", "release": "audacity@3.2.0"

PR #2135, opened 2021-11-15
Fixes the check for the valid UTF8 symbols

crsib:
Resolves: #2132
Resolves: #2134

Recommended:

### DIFF
--- a/libraries/lib-xml/XMLWriter.cpp
+++ b/libraries/lib-xml/XMLWriter.cpp
@@ -583,7 +583,7 @@ MemoryStream XMLUtf8BufferWriter::ConsumeResult()
 
 void XMLUtf8BufferWriter::WriteEscaped(const std::string_view& value)
 {
-   for (char c : value)
+   for (auto c : value)
    {
       switch (c)
       {
@@ -607,7 +607,7 @@ void XMLUtf8BufferWriter::WriteEscaped(const std::string_view& value)
          Write("&gt;");
          break;
       default:
-         if (c > 0x1F || charXMLCompatiblity[c] != 0)
+         if (static_cast<uint8_t>(c) > 0x1F || charXMLCompatiblity[c] != 0)
             mStream.AppendByte(c);
       }
    }


Paul-Licameli:
This looks harmless.  I don't immediately understand why this fixes the bug.  Is it because char is signed?  But the signed-unsigned comparison would have coerced the negative to a large unsigned value.

More seriously I should ask why all the logic in XMLWriterx::XMLEsc was not reproduced in XMLUtf8BufferWriter::WriteEscaped.  I should have asked that question with the earlier PR.

crsib:
>But the signed-unsigned comparison would have coerced the negative to a large unsigned value.

This is a signed-signed conversion, unfortunately.

> More seriously I should ask why all the logic in XMLWriterx::XMLEsc was not reproduced in

The problem is that it was mostly reproduced, just with the exception of surrogate pairs handling.

Paul-Licameli:
> > But the signed-unsigned comparison would have coerced the negative to a large unsigned value.
> 
> This is a signed-signed conversion, unfortunately.

DUH!  I see 0x... and assume unsigned, but there wasn't ...u after it.  Of course.

> 
> > More seriously I should ask why all the logic in XMLWriterx::XMLEsc was not reproduced in
> 
> The problem is that it was mostly reproduced, just with the exception of surrogate pairs handling.

Paul-Licameli:
> > But the signed-unsigned comparison would have coerced the negative to a large unsigned value.
> 
> This is a signed-signed conversion, unfortunately.
> 
> > More seriously I should ask why all the logic in XMLWriterx::XMLEsc was not reproduced in
> 
> The problem is that it was mostly reproduced, just with the exception of surrogate pairs handling.

And we know the string has been converted to utf8 here 

```
void XMLUtf8BufferWriter::WriteAttr(const std::string_view& name, const Identifier& value) 
{
   const wxScopedCharBuffer utf8Value = value.GET().utf8_str();

   WriteAttr(name, { utf8Value.data(), utf8Value.length() });
}
```

And that makes the surrogate handling unnecessary?  It is ok to leave all the utf8 encodings un-escap3ed?

Issue #2113, opened 2021-11-11
Smart Clips. In some cases Audacity crashes at exporting


Penikov:
**Describe the bug**
Audacity crashes in some cases when a user tries to export clips from a projects been impacted by the bug #2051 

**Additional information (please complete the following information):**
 - OS: Windows 10, 11
 - Version 3.1; 3.1.1]

PR #2136, opened 2021-11-15
Fixes issue with track length mismatch after resample

vsverchinsky:
It could happen so that after resampling number of samples contained in `GetPlayEndTime() - GetPlayStartTime()`
didn't match to number of samples in `GetPlayEndSample() - GetPlayStartSample()`

Resolves: #2113 

Recommended:

### DIFF
--- a/src/WaveClip.cpp
+++ b/src/WaveClip.cpp
@@ -1808,7 +1808,7 @@ sampleCount WaveClip::GetSequenceSamplesCount() const
 
 double WaveClip::GetPlayStartTime() const noexcept
 {
-    return mSequenceOffset + mTrimLeft;
+    return mSequenceOffset + SamplesToTime(TimeToSamples(mTrimLeft));
 }
 
 void WaveClip::SetPlayStartTime(double time)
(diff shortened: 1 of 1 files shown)


Paul-Licameli:
The small changes look harmless, but I don't understand why they fix a crash.

Would it be better to apply the small corrections in mTrimLeft and mTrimRight within WaveClip::Resample?

vsverchinsky:
> The small changes look harmless, but I don't understand why they fix a crash.

Because the amount of samples returned by `WaveClip::TimeToSamples(trim)` depends on current sample rate. When applying resampling it may happen so that trim value does not exactly match the beginning of the sample any more. And the bug comes from difference in the way number of samples is calculated: `WaveClip::GetPlaySamplesCount` adjusts trimmings to the sample rate, and `WaveClip::GetPlayEndTime()` and others - do not, so `WaveClip::TimeToSamples(WaveClip::GetPlayEndTime() - WaveClip::GetPlayStartTime())` may not be equal to `WaveClip::GetPlaySamplesCount`, sometimes leading to an OOR.

> Would it be better to apply the small corrections in mTrimLeft and mTrimRight within WaveClip::Resample?

It may work in that particular case, but I don't see why it's better. Functions mentioned above still return different values, which should be fixed anyway, so it becomes unnecessary to fix `Resample`. Also, resample(newSampleRate)->resample(originalSampleRate) may change original trimming values, while with this fix preserves them.

PR #2152, opened 2021-11-16
Fixes detach at silences by trimming unnecessary silence #2312

gpapp:
HandleSplit in 3.x.x is using the new Trim features, instead of Clear. When using the noise gate and detach silence for tracks, this results in full copies of the orginal track, with enormous files.

When HandleSplit is modified to use Clear, the operation falls back to the intended way.

As a side effect, when using this modifications, there are no unintentional copies of invisible data are created, e.g. spliting a clip actually splits, does not create copies.

Recommended:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -171,3 +171,4 @@ appimagetool/*
 .build.arm64/
 .conan/
 .debug/
+/out/build/x64-Debug
--- a/src/WaveTrack.cpp
+++ b/src/WaveTrack.cpp
@@ -1288,7 +1288,7 @@ void WaveTrack::HandleClear(double t0, double t1,
                   // guarantee, and might modify another clip
                   clipsToDelete.push_back( clip.get() );
                   auto newClip = std::make_unique<WaveClip>( *clip, mpFactory, true );
-                  newClip->TrimLeft(t1 - clip->GetPlayStartTime());
+                  newClip->ClearLeft(t1);
                   clipsToAdd.push_back( std::move( newClip ) );
                }
                else if (clip->AfterPlayEndTime(t1)) {
@@ -1298,7 +1298,7 @@ void WaveTrack::HandleClear(double t0, double t1,
                   // guarantee, and might modify another clip
                   clipsToDelete.push_back( clip.get() );
                   auto newClip = std::make_unique<WaveClip>( *clip, mpFactory, true );
-                  newClip->TrimRight(clip->GetPlayEndTime() - t0);
+                  newClip->ClearRight(t0);
 
                   clipsToAdd.push_back( std::move( newClip ) );
                }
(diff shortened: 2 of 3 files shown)


LWinterberg:
Fwiw: #2051 has become a much larger project to increase performance in general (with #2065, #2087 and #2121) going much beyond the initial detach at silences issue.

gpapp:
Reworked the modification to only affect the Detach silence function

LWinterberg:
looks good, thanks!

src/WaveTrack.cpp
```diff
@@ -1694,12 +1695,59 @@ void WaveTrack::Disjoin(double t0, double t1)
...
+   for (unsigned int i = 0; i < clipRegions.size(); i++)
+   {
+      const ClipRegion& region = clipRegions.at(i);
+      // There is no segment to be kept on the left
+      if (region.clip->GetPlayStartSample() == region.start)
+      {
+         // the region spans to the end of the clip, so we can just delete it
+         if (region.clip->GetPlayEndSample() == region.end) {
+            WaveClipHolder toRemove = RemoveAndReturnClip(region.clip.get());
+            toRemove.~shared_ptr();
```
ksoze95, line 1714: No need to explicitly call the destructor.  The destructor will be called when toRemove goes out of scope.

src/WaveTrack.cpp
```diff
@@ -1694,12 +1695,59 @@ void WaveTrack::Disjoin(double t0, double t1)
...
+
+      // lookahead to see if we are done with this clip
+      if ((i == clipRegions.size() - 1) || (region.clip!=clipRegions.at(i+1).clip)) {
+         // the region spans to the end of the clip, so we can just delete it
+         if (region.clip->GetPlayEndSample() != region.end) {
+            newClip = CopyToNewClip(region.clip, region.end, region.clip->GetPlayEndSample());
+         }
+         //remove the remainder of the clip
+         WaveClipHolder toRemove = RemoveAndReturnClip(region.clip.get());
+         toRemove.~shared_ptr();
```
ksoze95, line 1729: Same as above.  Not needed.

src/WaveTrack.cpp
gpapp: Removed explicit destructor call

src/WaveTrack.cpp
gpapp: Removed explicit destructor call

vsverchinsky (changes requested):
New clips should reference existing blocks instead of duplicating samples.

.gitignore
```diff
@@ -171,3 +171,4 @@ appimagetool/*
 .build.arm64/
 .conan/
 .debug/
+/out/build/x64-Debug
```
Paul-Licameli, line 174: This is extraneous to the fix

LWinterberg:

Issue #2147, opened 2021-11-16
[Sentry] Audacity crashes when releasing the clip handle


crsib:
```
OS Version: Windows 10.0.19043 (None)
Report Version: 104

Crashed Thread: 14900

Application Specific Information:
Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ

Thread 14900 Crashed:
0   *************************\Audacity.exe0x7ff7bdbf614e      [inlined] std::_Ptr_base<T>::_Construct_from_weak (memory:1322)
1   *************************\Audacity.exe0x7ff7bdbf614e      [inlined] std::weak_ptr<T>::lock (memory:3050)
2   *************************\Audacity.exe0x7ff7bdbf614e      [inlined] Track::GetOwner (Track.h:379)
3   *************************\Audacity.exe0x7ff7bdbf614e      TrackList::Channels<T> (Track.h:1487)
4   *************************\Audacity.exe0x7ff7bdbfa1d4      SelectionState::SelectTrack (SelectionState.cpp:57)
5   *************************\Audacity.exe0x7ff7bdf0a9fc      AffordanceHandle::Release (AffordanceHandle.cpp:80)
6   *************************\Audacity.exe0x7ff7bdb3d5f3      CellularPanel::OnMouseEvent (CellularPanel.cpp:814)
7   *************************\wxbase313u_vc_x64_custom.dll0x7ffc0ca114a3      wxAppConsoleBase::CallEventHandler (appbase.cpp:669)
8   *************************\wxbase313u_vc_x64_custom.dll0x7ffc0cb3c733      wxEvtHandler::ProcessEventIfMatchesId (event.cpp:1415)
9   *************************\wxbase313u_vc_x64_custom.dll0x7ffc0cb3c2ab      wxEventHashTable::HandleEvent (event.cpp:1023)
10  *************************\wxbase313u_vc_x64_custom.dll0x7ffc0cb3dbcc      wxEvtHandler::TryHereOnly (event.cpp:1612)
11  *************************\wxbase313u_vc_x64_custom.dll0x7ffc0cb3c7a4      wxEvtHandler::ProcessEventLocally (event.cpp:1545)
12  *************************\wxbase313u_vc_x64_custom.dll0x7ffc0cb3c65b      wxEvtHandler::ProcessEvent (event.cpp:1518)
[...]

PR #2160, opened 2021-11-18
Issue 2147: fix affordances crash

vsverchinsky:
Resolves: #2147 

Please, pay attention to the possibly blocking behaviour

Recommended:

### DIFF
--- a/src/AudioIO.cpp
+++ b/src/AudioIO.cpp
@@ -1462,8 +1462,9 @@ void AudioIO::StopStream()
 
       while( mAudioThreadShouldCallTrackBufferExchangeOnce )
       {
-         // LLL:  Experienced recursive yield here...once.
-         wxTheApp->Yield(true); // Pass true for onlyIfNeeded to avoid recursive call error.
+         //FIXME: Seems like this block of the UI thread isn't bounded,
+         //but we cannot allow event handlers to see incompletely terminated
+         //AudioIO state with wxYield (or similar functions)
          wxMilliSleep( 50 );
       }
 
--- a/src/tracks/ui/AffordanceHandle.cpp
+++ b/src/tracks/ui/AffordanceHandle.cpp
@@ -16,9 +16,7 @@
 #include "ViewInfo.h"
 #include "../../SelectionState.h"
 #include "../../TrackPanelMouseEvent.h"
-#include "../../WaveClip.h"
 #include "../../Track.h"
-#include "../../WaveTrack.h"
 #include "../../../images/Cursors.h"
 
 HitTestPreview AffordanceHandle::HitPreview(const AudacityProject*, bool unsafe, bool moving)
@@ -69,17 +67,17 @@ UIHandle::Result AffordanceHandle::Click(const TrackPanelMouseEvent& evt, Audaci
 UIHandle::Result AffordanceHandle::Release(const TrackPanelMouseEvent& event, AudacityProject* pProject, wxWindow* pParent)
 {
     auto result = TimeShiftHandle::Release(event, pProject, pParent);
-    //Clip was not moved
-    if (!TimeShiftHandle::WasMoved())
+    if (!WasMoved())
     {
-        const auto track = TrackList::Get(*pProject).Lock<Track>(GetTrack());
-
-        auto& selectionState = SelectionState::Get(*pProject);
         auto& trackList = TrackList::Get(*pProject);
-        selectionState.SelectNone(trackList);
-        selectionState.SelectTrack(*track, true, true);
+        if(const auto track = trackList.Lock<Track>(GetTrack()))
+        {
+            auto& selectionState = SelectionState::Get(*pProject);
+            selectionState.SelectNone(trackList);
+            selectionState.SelectTrack(*track, true, true);
 
-        result |= SelectAt(event, pProject);
+            result |= SelectAt(event, pProject);
+        }
     }
     return result;
 }


src/tracks/ui/AffordanceHandle.cpp
```diff
@@ -69,17 +67,17 @@ UIHandle::Result AffordanceHandle::Click(const TrackPanelMouseEvent& evt, Audaci
...
-    if (!TimeShiftHandle::WasMoved())
+    if (!WasMoved())
     {
-        const auto track = TrackList::Get(*pProject).Lock<Track>(GetTrack());
-
-        auto& selectionState = SelectionState::Get(*pProject);
         auto& trackList = TrackList::Get(*pProject);
-        selectionState.SelectNone(trackList);
-        selectionState.SelectTrack(*track, true, true);
+        if(const auto track = TrackList::Get(*pProject).Lock<Track>(GetTrack()))
```
Paul-Licameli, line 73: You might eliminate a repetition of `TrackList::Get(*pProject)`

src/AudioIO.cpp
```diff
@@ -1462,8 +1462,9 @@ void AudioIO::StopStream()
 
       while( mAudioThreadShouldCallTrackBufferExchangeOnce )
       {
-         // LLL:  Experienced recursive yield here...once.
-         wxTheApp->Yield(true); // Pass true for onlyIfNeeded to avoid recursive call error.
+         //FIXME: Seems like this block of the UI thread isn't bounded,
+         //but we cannot allow event handlers to see incompletely terminated
+         //AudioIO state with wxYield (or similar functions)
```
Paul-Licameli, line 1467: Study the function `AudioThread::ExitCode AudioThread::Entry()` for the thread that must make the change of state that un-blocks this main thread.  @crsib has become more familiar with this thread, after reviewing other PRs of mine that refactor the functions called in it.

I strongly doubt that this other thread ever fails to make progress now, or to change the condition that the main thread is waiting on here.  If it ever did, then the loop above would have effectively become the new alternative event-loop of Audacity, making the program still responsive to some inputs, and yet the program would be stuck in a strange state, unable to stop a play or recording and start another one.  But I recollect no such bugs.

The Yield() call does nothing to assist this other thread in making progress, which is why I favor removing this ancient and unexplained piece of code.

Paul-Licameli: And, I am very aware that this bit of inter-thread communication with a volatile bool variable is not correct.  Leave it to me to rewrite it better, as part of my sweep of all the bad old stuff in the audio engine that I want to complete in the coming release.

This looks like the sort of thing that should be done the modern way with a std::condition_variable, and a companion std::mutex.  And it would not be bad to do so, because the other thread is not the low-latency thread.  We don't need extra effort to avoid memory allocations or mutex locking in it.

Paul-Licameli (approved):
Approved, with one small suggestion for simplification.

I would like @crsib to comment, even if only to confirm he has read all of my comments.

src/AudioIO.cpp
Paul-Licameli: The Yield() seems only to have been a source of bugs, because it causes some event callbacks to be invoked while still in the middle of other, un-finished event callbacks, causing surprising things to happen because the callbacks were not written in anticipation of this.

The Yield() call is NOT a yielding of the thread to other threads (like `std::this_thread::yield()`) but rather allows more event dispatches to happen while still in the main thread, as a sort of cooperative multi-tasking.

The original author of this Yield() thread (some time before the 2010 event horizon of this git repository) might just have been confused, and nobody has yet paused long enough to recognize that confusion.  Or maybe they really were trying to compensate for bugs in ancient versions of AudioThread, but these bugs were fixed in better ways by other means.

Paul-Licameli: And you have my great respect for working backwards from a crashing call stack to this Yield() as the ulterior cause of it.  It is very non-obvious.

src/tracks/ui/AffordanceHandle.cpp
vsverchinsky: Sure, thanks!

src/AudioIO.cpp
vsverchinsky: Thank you!

crsib: I agree that the code in StopStream is quite broken, and the check there doesn't work.

What bothers me here - is there a chance that the app will hang in the new loop? The previous version effectively created a new "run loop," which is also wrong but more user-friendly.

Should we think of a better synchronization here now? Like using atomics for mStreamToken and waiting on a proper condition before we exit. I can easily imagine a case where `mAudioThreadShouldCallTrackBufferExchangeOnce` is true till the end of the world. Should we just wait on mStreamToken?

crsib: We've debugged with Vitaly, probably it wont hang :)

Paul-Licameli: Sure, this needs rewriting.  In practice, the Audio thread doesn't take long to finish its work.

We could use a std condition variable and wait_for if we are worried about that contingency, but I don't know how the main program could recover if the Audio thread is not making progress.  Kill the thread?  Leaving things in what undefined state?  I think there would be nothing to do but put up an error dialog, then quit the program after the acknowledgment.

But it is not a known problem in my experience.

I'm speculating that the ancient code yielding to the event dispatcher MIGHT have helped twenty years ago, with slower drives recording the samples, to allow refreshes of windows to happen, in case the delay of the other thread was noticeable.  Maybe it prevented visual ugliness.

The problem is yeilding to all events indiscriminately.

crsib: After even more discussion with Vitaly, I think we should probably just use `std::this_thread::yield();` instead of both of the instructions.

crsib: But this code is fine as well!

crsib (approved):

src/AudioIO.cpp
Paul-Licameli: So you think a busy-wait, but with a yield, would be preferable to the repeated sleeps?  I have had that thought too.

petersampsonaudacity:
I tested this with @vsverchinsky  Vitaly's branch build: audacity-win-3.1.3-alpha-20211119+d998117-64bit

Testing on W10 I get no crash with this build - I do get the crash with 3.1.2

Issue #2004, opened 2021-10-22
Incorrect error message in Spectral Multi-Tool


SteveDaulton:
**Describe the bug**
The "Spectral Multi-Tool" may return an incorrect error message

**To Reproduce**
1. Make a spectral selection from less than 10 Hz (but more than 0  Hz), up to the top of the spectral display.
2. Apply "Spectral Multi-Tool"
3. Note that when the lower frequency is set and the upper frequency is not set, the "Spectral Multi-Tool" acts as a low-pass filter.
4. Observe the error message:
```
Error.
Notch filter parameters cannot be applied.
Try increasing the low frequency bound
or reduce the filter 'Width'.
```

**Expected behavior**
The error message should not refer to "Notch filter" and the "width" is meaningless for a low-pass filter.

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version Audacity 3.0.2

**Additional context**
The "steps to reproduce" are difficult to perform due to https://github.com/audacity/audacity/issues/1921

vivekKr24:
I would like to work on this issue.
Please guide me through.

LWinterberg:
@vivekKr24 My guess would be that you need to change this line:

https://github.com/audacity/audacity/blob/95a7fc6362729594565204b6db927b2e025d20c0/plug-ins/SpectralEditMulti.ny#L53

Feel free to ask on audacityteam.org/devserver if you have more questions

PR #2185, opened 2021-11-25
Error message shows up for Notch Filter instead on Lowpass filter 

vivekKr24:
Resolves: #2004

The condition now checks for valid Notch Filter Parameters.

Previously, it was not checking if upper frequency was set or not, resulting in an error message describing the "Notch filter", even when the tool was supposed to act like a Lowpass Filter (upper frequency was not set)
If the upper frequency f1 was undefined (not set), the SpectralEditMulti tool must work like a Lowpass Filter with very low 'lower frequency' `(f0 < 10)`,  which had similar effects compared to the 'Low pass Filter' plug-in on my test audio file. (Hence no problems here)
We need not modify the error message as we don't have to handle the case talked above, and only throw error at a problematic (low central frequency and high bandwith) "Notch Filter" as mentioned in the comments at line 49.

Adding a condition that f1 is not undefined (tool should now work like a notch filter) will fix the problem (Error will only be shown when the case described in comments at line 49 arrives). I have tried the steps for reproducing the problems mentioned in the issue and it is no longer occuring.

Recommended:

### DIFF
--- a/plug-ins/SpectralEditMulti.ny
+++ b/plug-ins/SpectralEditMulti.ny
@@ -48,7 +48,7 @@ $copyright (_ "Released under terms of the GNU General Public License version 2"
                   p-err f0)))
       ;; Biqud filter fails if centre frequency is very low and bandwidth very high.
       ;; 'Magic numbers' 10 Hz and 10 octaves are experimental.
-      ((and f0 f1 (< f0 10) (> bw 10))
+      ((and f0 (< f0 10) bw (> bw 10))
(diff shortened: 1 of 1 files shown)


SteveDaulton (commented):
Works for me, though I think the logic would be clearer as:
`((and f0 (< f0 10) bw (> bw 10))`

vivekKr24:
I have made the suggested changes.

SteveDaulton (approved):
Looks good to me.

Issue #1151, opened 2021-06-25
Extract lib-track


Paul-Licameli:
- [x] #1153 
- [x] #1158
- [ ] #1151

PR #2186, opened 2021-11-26
Extract lib track

Paul-Licameli:
Resolves: #1151

Extract a library for the Track abstract base class and for TrackList, which contains no mention of any concrete sub-classes of Track, not even as incomplete types.

The set of concrete Track sub-types known to the application should become open-ended, perhaps even extensible with plug-in modules.

Recommended:


crsib (approved):
A very good example of "sequence of easily reviewable commits", thank you, Paul!

Paul-Licameli:
> A very good example of "sequence of easily reviewable commits", thank you, Paul!

Is that sarcasm?  This was t easy to write.

crsib:
Absolutely sincere

PR #2187, opened 2021-11-26
Fix some build errors on linux Debian.

carlo-bramini:
I tried to build audacity on linux Debian and I got two errors.
This PR fixes them.


crsib:
The fix is pushed directly to master. This is unclear if we can accept such change without CLA; trivial changes are usually typos or so. Thank you for pointing out the problem!

carlo-bramini:
> The fix is pushed directly to master. This is unclear if we can accept such change without CLA; trivial changes are usually typos or so. Thank you for pointing out the problem!

Please excuse me, while we are on the topic, I just compiled the items in the form for CLA at this page:
https://www.audacityteam.org/cla/
But it is not clear to me if I completed it or not: I arrived at the end of the form and I checked the last question "I certify that the above information is accurate to the best of my knowledge" but I have not found a "submit" button or something like that.
Is the CLA completed by just writing the replies to the questions or there are other actions to do?
Thank you.

crsib:
Whoa, that is odd, we will recheck it (@LWinterberg)

LWinterberg:
the iframe got cut off; you now should be able to see a glorious purple send button again.

carlo-bramini:
> the iframe got cut off; you now should be able to see a glorious purple send button again.

"Send" button now visible and pressed, thank you.

PR #2201, opened 2021-11-29
Add missing includes for GCC 11.1.0

Mortal:
On Arch Linux, I was unable to build the AUR package for audacity-git without these changes.

  - Not required for trivial changes (added 2 missing includes)

Recommended:
  - There are two independent fixes, one per commit, but both are required in order to compile and run on my machine.


crsib:
The fix is pushed directly to master. This is unclear if we can accept such change without CLA; trivial changes are usually typos or so. Thank you for pointing out the problem!

PR #816, opened 2021-04-20
Minor build warning fixes

imciner2:
Minor changes to fix build warnings for `-Wreorder` and `-Wpessimize-move` on GCC 10 (seen when building the Fedora package, since these flags get turned on in their build system).

The first commit just rearranges the initialization list for classes so that the elements are in the same order as the class members (since that is how the initialization is done anyway).

The second commit removes unneeded `std::move` calls from return statements that would prevent compilers from perform any move/copy elision optimizations on the values being returned (https://developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c/).

PR #2212, opened 2021-12-01
Fix warnings found during Fedora package build

imciner2:
In the process of updating the Fedora package to 3.1.2 I noticed a lot of warnings. Many were harmless unused variable warnings, but there were others that are easy to fix and that do provide better compatibility/code quality. Specifically, the following are fixed:

* Replace the deprecated `std::uncaught_exception` with `std::uncaught_exceptions`.
`std::uncaught_exception` was deprecated in C++17 and removed in C++20 and replaced with `std::uncaught_exceptions` starting in C++17. Since the Audacity code is C++17 minimum, this was a simple change in the two places it is used.

* Cleanup initializer list ordering.
This makes the initializer list actually follow the same order as the order the variables will be initialized in, since in C++ they are initialized in the order they are declared in the class instead of in the list. These warnings can be found using the `-Wreorder` flag.
As part of this there was a small change to move a variable initialization out of the conditional compilation. That variable is not declared in the conditional compilation, so it was just being unset by default instead of getting a value.

* Don't force the use of `std::move` on return values since these will prevent the compiler form performing move/copy ellision optimization (developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c).

* Don't create a copy of an object when iterating over it.

* Fix a violation of the one definition rule by de-anonymizing a few classes so that the main `WaveTrackSubView` will have only one definition across all the compilation units.
Previously it would have a different definition in each compilation unit it was in because the `weak_ptr` on anonymous classes prevented that.
  

Recommended:

Note, this now replaces https://github.com/audacity/audacity/pull/816.

### DIFF
--- a/libraries/lib-exceptions/AudacityException.h
+++ b/libraries/lib-exceptions/AudacityException.h
@@ -214,7 +214,7 @@ R GuardedCall(
          // At this point, e is the "current" exception, but not "uncaught"
          // unless it was rethrown by handler.  handler might also throw some
          // other exception object.
-         if (!std::uncaught_exception()) {
+         if (std::uncaught_exceptions() == 0) {
             auto pException = std::current_exception(); // This points to e
             AudacityException::EnqueueAction(
                pException, std::move(delayedHandler));
--- a/src/FreqWindow.cpp
+++ b/src/FreqWindow.cpp
@@ -190,8 +190,8 @@ FrequencyPlotDialog::FrequencyPlotDialog(wxWindow * parent, wxWindowID id,
                            const wxPoint & pos)
 :  wxDialogWrapper(parent, id, title, pos, wxDefaultSize,
             wxDEFAULT_DIALOG_STYLE | wxRESIZE_BORDER | wxMAXIMIZE_BOX),
-   mAnalyst(std::make_unique<SpectrumAnalyst>())
-,  mProject{ &project }
+   mProject{ &project }
+,  mAnalyst(std::make_unique<SpectrumAnalyst>())
 {
    SetName();
 
--- a/src/Printing.cpp
+++ b/src/Printing.cpp
@@ -48,8 +48,8 @@ class AudacityPrintout final : public wxPrintout
    AudacityPrintout(wxString title,
                     TrackList *tracks, TrackPanel &panel):
       wxPrintout(title),
-      mTracks(tracks)
-      , mPanel(panel)
+        mPanel(panel)
+      , mTracks(tracks)
    {
    }
    bool OnPrintPage(int page);
--- a/src/ProjectSettings.cpp
+++ b/src/ProjectSettings.cpp
@@ -59,10 +59,6 @@ ProjectSettings::ProjectSettings(AudacityProject &project)
       NumericConverter::TIME,
       gPrefs->Read(wxT("/SelectionFormat"), wxT("")))
 }
-, mAudioTimeFormat{ NumericTextCtrl::LookupFormat(
-   NumericConverter::TIME,
-   gPrefs->Read(wxT("/AudioTimeFormat"), wxT("hh:mm:ss")))
-}
 , mFrequencySelectionFormatName{ NumericTextCtrl::LookupFormat(
    NumericConverter::FREQUENCY,
    gPrefs->Read(wxT("/FrequencySelectionFormatName"), wxT("")) )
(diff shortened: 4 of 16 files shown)


libraries/lib-exceptions/AudacityException.h
```diff
@@ -214,7 +214,7 @@ R GuardedCall(
          // At this point, e is the "current" exception, but not "uncaught"
          // unless it was rethrown by handler.  handler might also throw some
          // other exception object.
-         if (!std::uncaught_exception()) {
+         if (std::uncaught_exceptions() == 0) {
```
LWinterberg, line 217: Are you sure about this `== 0` here? wouldn't `!std::uncaught_exceptions()` be the normal way of doing this (as was the case before)? (same for the ImportAUP further down)

imciner2: `std::uncaught_exception()` returned a boolean value that was true if there were exceptions whereas `std::uncaught_exceptions()` returns the number of exceptions, so to be proper with the types it needs to be compared against an integer. The prior call was checking for when no exceptions were pending (it wanted `std::uncaught_exception` to return false), so that will be when `std::uncaught_exceptions` returns 0.

LWinterberg: I see.

imciner2:
Interesting, the macOS failure is:
```
In file included from /Users/runner/work/audacity/audacity/libraries/lib-exceptions/AudacityException.cpp:12:
/Users/runner/work/audacity/audacity/libraries/lib-exceptions/AudacityException.h:217:19: error: 'uncaught_exceptions' is unavailable: introduced in macOS 10.12
         if (std::uncaught_exceptions() == 0) {
                  ^
In file included from /Users/runner/work/audacity/audacity/libraries/lib-exceptions/AudacityException.cpp:12:
In file included from /Users/runner/work/audacity/audacity/libraries/lib-exceptions/AudacityException.h:14:
In file included from /Users/runner/work/audacity/audacity/libraries/lib-utility/MemoryX.h:5:
In file included from /Applications/Xcode_13.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.0.sdk/usr/include/c++/v1/iterator:419:
In file included from /Applications/Xcode_13.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.0.sdk/usr/include/c++/v1/__functional_base:15:
In file included from /Applications/Xcode_13.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.0.sdk/usr/include/c++/v1/typeinfo:61:
/Applications/Xcode_13.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.0.sdk/usr/include/c++/v1/exception:134:63: note: 'uncaught_exceptions' has been explicitly marked unavailable here
_LIBCPP_FUNC_VIS _LIBCPP_AVAILABILITY_UNCAUGHT_EXCEPTIONS int uncaught_exceptions() _NOEXCEPT;
                                                              ^
1 error generated.
```
but the downloads page here https://www.audacityteam.org/download/mac/ implies 10.13 should be the minimum version, so we should be safe using this function since it was in 10.12+. So it looks like the minimum macOS version needs to be bumped in the configure flags.

crsib:
>but the downloads page here https://www.audacityteam.org/download/mac/

We need better wording there :-) Currently, the minimal version is 10.9, but I don't think that this is tested much now.

imciner2:
Ok, so I guess the full C++17 isn't supported then :(. I can go ahead and remove that one commit from this patch set if you want. (IMO supporting anything older than 10.12 is difficult for macOS, and even 10.9 is pushing it because now wxWidgets is going 10.10 as their minimum).

Paul-Licameli:
> Ok, so I guess the full C++17 isn't supported then :(. I can go ahead and remove that one commit from this patch set if you want. (IMO supporting anything older than 10.12 is difficult for macOS, and even 10.9 is pushing it because now wxWidgets is going 10.10 as their minimum).

Hello Ian,

I reviewed the changes, and with the exception of the first commit, which breaks our macOS build, they are all good.

Please rebase onto recent master and force-push this branch again, minus that commit, and I will approve and merge it.

Thank you!

libraries/lib-exceptions/AudacityException.h
Paul-Licameli, line 217: A build-breaker, as you know, until we make a commitment to raising the minimum macOS version.  So please omit this commit from the branch.

Paul-Licameli (changes requested):
Build breaker because C++17 library is incomplete for older macOS

Paul-Licameli:
I rebased this branch, removing the one commit that breaks the macOS build, and merged it.

Thank you for the fixes.

Issue #1146, opened 2021-06-25
Extract lib-sample-track


Paul-Licameli:
- [x] #1159
- [ ]  #1151 
- [ ] remaining

PR #2219, opened 2021-12-02
Extract lib sample track

Paul-Licameli:
Resolves: #1146

Extract abstract base classes SampleTrack and (extending that) WritableSampleTrack, from WaveTrack.

Mixer and WaveTrackCache are rewritten in terms of SampleTrack only.

AudioIO uses Mixer, and other parts of it for recording will be rewritable in terms of WritableSampleTrack.

Thus this library extraction is a step toward an extraction of AudioIO into a higher library, which will not depend on the
implementation details of WaveTrack's division into WaveClips that use Sequence, or its persistency using sqlite.

SampleTrack may be a useful abstraction in effect stacks, or perhaps in more general graphs.

Recommended:


src/tracks/playabletrack/wavetrack/ui/SpectrumView.cpp
```diff
@@ -333,7 +332,8 @@ void DrawClipSpectrum(TrackPanelDrawingContext &context,
       return;
    }
 
-   const WaveTrack *const track = waveTrackCache.GetTrack().get();
+   const auto track =
+      static_cast<const WaveTrack*>(waveTrackCache.GetTrack().get());
```
crsib, line 337: Is it really safe?

Paul-Licameli: I thought you would flag that.

If we assume no other concrete subclasses of SampleTrack exist, yes.

Will that be a safe future assumption? ... Maybe not.

Paul-Licameli: Would it be nice to make SpectrumView depend only on SampleTrack, not WaveTrack?  Yes.

What stops it here?  The need for the track to define GetSpectrogramSettings and GetSpectrumBounds.

I have other plans, later in the big branch, to move those things into ClientData attachments to the SampleTrack, but that is a bit beside the point of the present library extraction.

Paul-Licameli: I could make it a dynamic_cast, and return early if it is null.  That means if we attempt spectrogram display of some other kind of hypothetical future concrete SampleTrack, it would appear as a blank rectangle, but not be a crashing bug.

Paul-Licameli: Do you not object to the other static_cast in WaveClip.cpp?  (Those things only ever belong to WaveTracks.)

Have you reviewed all else and found no problems?

Paul-Licameli: SampleTrack does not expose the notion of division of the sample sequence into clips.  It might need that for generalizing SpectrumView in future.  It is not yet needed for purposes of extracting AudioIO so it does not depend on WaveTrack.

Paul-Licameli: I force-pushed a fix for your complaint.

crsib (approved):

Paul-Licameli:
Thanks!

Which one should I do next?  AudioIO or WaveTrack?

Issue #1470, opened 2021-08-13
Add support for WavPack


crsib:
Add support for WavPack importing and exporting. 

WavePak is an open, patent-free codec supporting lossless and hybrid compression. 

WavPak supports 8, 16, 24, and 32-bit integer PCM; 32-bit float PCM.

The compression performance is relatively high - 78 MB/s for 32-bit float, which makes it to be a possible candidate for the persistence layer of Audacity.

LWinterberg:
@PratyayAmrit75 Sure, go ahead! I'll keep you assigned for a week or so, so you can get started. Please report back with your progress - you may want to [make a draft pull request](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) when you get it in a state where the idea of your implementation becomes apparent, but it doesn't quite work yet.

LWinterberg:
@PratyayAmrit75 you may want to ask in the discord server: 
https://discord.gg/N3XKxzTrq3

Subhra264:
Hello, want to work on this issue...

LWinterberg:
Hello @Subhra264! Do you know where to start?

Subhra264:
> Hello @Subhra264! Do you know where to start?

Yeah, I hope so,. Here's what I need to do( I think)---
1. Somehow add the libwavpack library
2. For wavpack import, create a ImportWavPack.cpp file next to ImportMP3.cpp
3. Write code to import wavpack following other ImportX.cpp files
4. Similar steps for ExportWavpack.cpp

LWinterberg:
Alrighty! I'll assign it to you for a bit; if you've got any questions, feel free to ask on our [discord](https://audacityteam.org/devserver)

Subhra264:
Looks like, now in a position to open a pull request.

>  you may want to [make a draft pull request](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) when you get it in a state where the idea of your implementation becomes apparent, but it doesn't quite work yet.

 So I need to create a Work In Progress Pull Request right?

PR #2224, opened 2021-12-03
Integrate WavPack library to export/import wavpack files

Subhra264:
Resolves: #1470 

This Pull Request is created to integrate wavpack with audacity so that users can import/export wavpack files.

### Export:
1. By default, hybrid mode is not selected and the wavpack library creates a pure lossless "wv" file. If hybrid mode is enabled and the "created correction file" option is not selected then it will create a single lossy "wv" file.
Export wavpack image 1

2.  Enabled "create correction file", so it will create a lossy "wv" file along with its "wvc" correction file.
Export wavpack image 2

3. Added the value for "Artist Name" tag.
Export wavpack image 3

4. 
Export wavpack image 4

5. 
Export wavpack image 5 (with wvc)

### Import
1. 
Import wavpack image 1(with wvc)

2. 
Import wavpack image 2(with wvc)

3.
Import wavpack image 3(with wvc)

4. Exported tags are present in the imported audio
Import wavpack image 4(with wvc)

Also, the exported "wv" file successfully opens with the "VLC Media Player".

Recommended:

### DIFF
--- a/src/export/ExportWavPack.cpp
+++ b/src/export/ExportWavPack.cpp
@@ -1,0 +1,65 @@
+/**********************************************************************
+
+Audacity: A Digital Audio Editor
+
+ExportWavPack.cpp
+
+Subhradeep Chakraborty
+
+This program is distributed under the GNU General Public License, version 2.
+A copy of this license is included with this source.
+
+Based on ExportOGG.cpp by:
+Joshua Haberman
+
+**********************************************************************/
+
+#ifdef USE_WAVPACK
+
+#include "Export.h"
+#include "../ShuttleGui.h"
+#include "../widgets/ProgressDialog.h"
+#include "wxFileNameWrapper.h"
+#include "../Tags.h"
+#include "Mix.h"
+
+
+#include <wavpack.h>
+
+
+class ExportWavPackOptions final : public wxPanelWrapper
+{
+public:
+
+   ExportWavPackOptions(wxWindow *parent, int format);
+   virtual ~ExportWavPackOptions();
+
+   void PopulateOrExchange(ShuttleGui & S);
+   bool TransferDataToWindow() override;
+   bool TransferDataFromWindow() override;
+
+};
+
+class ExportWavPack final : public ExportPlugin
+{
+public:
+   
+   ExportWavPack();
+
+   void OptionsCreate(ShuttleGui &S, int foramt) override;
+
+   ProgressResult Export(AudacityProject *project,
+               std::unique_ptr<ProgressDialog> &pDialog,
+               unsigned channels,
+               const wxFileNameWrapper &fName,
+               bool selectedOnly,
+               double t0,
+               double t1,
+               MixerSpec *mixerSpec = NULL,
+               const Tags *metadata = NULL,
+               int subformat = 0) override;
+   
+
+};
+
+#endif
--- a/src/export/ExportWavPack.cpp
+++ b/src/export/ExportWavPack.cpp
@@ -18,21 +18,24 @@
 #ifdef USE_WAVPACK
 
 #include "Export.h"
+#include "wxFileNameWrapper.h"
 #include "FileIO.h"
+#include "Prefs.h"
+#include "Mix.h"
 
 #include <wavpack.h>
 #include <wx/log.h>
 #include <wx/checkbox.h>
 #include <wx/choice.h>
+#include <wx/stream.h>
 
 #include "../ShuttleGui.h"
 #include "../ProjectSettings.h"
 #include "../widgets/AudacityMessageBox.h"
 #include "../widgets/ProgressDialog.h"
-#include "wxFileNameWrapper.h"
+#include "Track.h"
+#include "ProjectRate.h"
 #include "../Tags.h"
-#include "Mix.h"
-#include "Prefs.h"
 
 //---------------------------------------------------------------------------
 // ExportWavPackOptions
@@ -148,6 +151,9 @@ void ExportWavPackOptions::PopulateOrExchange(ShuttleGui & S)
    bool hybridMode = false;
    bool createCorrectionFile = false;
 
+   gPrefs->Read(wxT("/FileFormats/WavPackHybridMode"), &hybridMode, 0);
+   gPrefs->Read(wxT("/FileFormats/WavPackCreateCorrectionFile"), &createCorrectionFile, 0);
+
    S.StartVerticalLay();
    {
       S.StartHorizontalLay(wxEXPAND);
@@ -224,7 +230,7 @@ public:
 
    ExportWavPack();
 
-   void OptionsCreate(ShuttleGui &S, int foramt) override;
+   void OptionsCreate(ShuttleGui &S, int format) override;
 
    ProgressResult Export(AudacityProject *project,
                std::unique_ptr<ProgressDialog> &pDialog,
@@ -262,11 +268,16 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
                        const Tags *metadata,
                        int WXUNUSED(subformat))
 {
-   WavpackConfig config;
-   WriteId outWvFile, outWvcFile;
+   WavpackConfig config = {0};
+   WriteId outWvFile = {NULL}, outWvcFile = {NULL};
    outWvFile.file = std::make_unique< FileIO >(fName, FileIO::Output);
 
-   double rate = ProjectSettings::Get( *project ).GetRate();
+   if (!outWvFile.file.get()->IsOpened()) {
+      AudacityMessageBox( XO("Unable to open target file for writing") );
+      return ProgressResult::Cancelled;
+   }
+   
+   double rate = ProjectRate::Get( *project ).GetRate();
    const auto &tracks = TrackList::Get( *project );
 
    int quality = gPrefs->Read(wxT("/FileFormats/WavPackEncodeQuality"), 1);
@@ -295,6 +306,7 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
 
       if (createCorrectionFile) {
          config.flags |= CONFIG_CREATE_WVC;
+         outWvcFile.file = std::make_unique< FileIO >(fName.GetFullPath().Append("c"), FileIO::Output);
       }
    }
 
@@ -307,7 +319,8 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
 // Samples to write per run
 #define SAMPLES_PER_RUN 8192u
 
-   ArrayOf<int32_t> wavpackBuffer{ SAMPLES_PER_RUN * numChannels };
+   uint32_t bufferSize = SAMPLES_PER_RUN * numChannels;
+   ArrayOf<int32_t> wavpackBuffer{ bufferSize };
    auto updateResult = ProgressResult::Success;
    {
       auto mixer = CreateMixer(tracks, selectionOnly,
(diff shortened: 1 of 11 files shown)


LWinterberg:
Hello @Subhra264, are you planning to continue work on this?

Subhra264:
Hello, sorry, I know that I am late( I had no idea about audio channels, samples, sample rate etc. so had to study it first), but I am still working on it..

The `ImportWavPack.cpp` file is almost ready, except the `ImportWavPackFileHandle::Import` function which is partly complete. For example I pushed all the `channels` to `outTracks`. But it's taking time because I am not sure what to do with the `samplesRead` decoded with `WavpackUnpackSamples` as  in  the `ImportFLAC.cpp` file, I saw that channels were directly pushed to `outTracks`,  but in `ImportOgg.cpp` file, the read samples were appended to channels.
Also in `Import.cpp` and `ProjectFileManager.cpp`(I forgot the exact name but it was something similar), it looked like only those `tracks` were used.
Basically, I am confused that if only channels created with `WaveTrackFactory` and `sampleRate` are enough or I need to somehow append the decoded samples to the channels.

Hopefully I will be able complete in 1 or 2 days.
Thanks...

crsib:
Hi @Subhra264 

Sorry for the delay! Can you please rebase onto the latest master, squash you changes into a single commit and force push?

Paul-Licameli:
> Hi @Subhra264
> 
> Sorry for the delay! Can you please rebase onto the latest master, squash you changes into a single commit and force push?

I would like to see a rebase too, but not necessarily a squash to one commit, if the intermediate steps are really useful for review of the new feature in sensible incremental steps, and not just corrections of mistakes.

Subhra264:
> Sorry for the delay!

No Issues! :)
>  but not necessarily a squash to one commit, if the intermediate steps are really useful for review of the new feature in sensible incremental steps, and not just corrections of mistakes.

yeah, actually, I added new commits after the previous commit is bug-free ( I mean, after resolving the bugs I encountered related to that particular feature/implementation, so may not be "bug-free" ). So, each commit mostly represents a new feature/file. For example, the first commit is about integrating wavpack and creating a file containing all the necessary class details, the second commit implements all the `ImportPlugin` and `ImportFileHandle` class methods, and the third commit is to read tags, etc.

Subhra264:
> I have crafted a commit, that should resolve your CI issues https://github.com/crsib/audacity/commit/2a353fd357f6cebb1a9cda61be33a9afe3edfb36

All checks passed! Thank you!

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,431 @@
+/**********************************************************************
+
+   Audacity: A Digital Audio Editor
+
+   ExportWavPack.cpp
+
+   Subhradeep Chakraborty
+
+   This program is distributed under the GNU General Public License, version 2.
```
crsib, line 9: Please update to GPL2-or-later.

Or even better, please add `/*  SPDX-License-Identifier: GPL-2.0-or-later */` to the top of both Export and Import

Subhra264: done

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+#define SAMPLES_PER_RUN 8192u
+
+   uint32_t bufferSize = SAMPLES_PER_RUN * numChannels;
+   ArrayOf<int32_t> wavpackBuffer{ bufferSize };
+   auto updateResult = ProgressResult::Success;
+   {
+      auto mixer = CreateMixer(tracks, selectionOnly,
+         t0, t1,
+         numChannels, SAMPLES_PER_RUN, true,
+         rate, int16Sample, mixerSpec);
```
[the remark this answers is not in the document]
crsib: (I believe, that Audacity will return you an array of 32 ints, if you request sample24 as an output format for the mixer. Am I right, @Paul-Licameli?)

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+      while (updateResult == ProgressResult::Success) {
+         auto samplesThisRun = mixer->Process(SAMPLES_PER_RUN);
+
+         if (samplesThisRun == 0)
+            break;
+
+         char *mixed = (char *)(const char*)mixer->GetBuffer();
+         for (decltype(samplesThisRun) j = 0; j < samplesThisRun; j++) {
+            for (size_t i = 0; i < numChannels; i++) {
```
crsib, line 344: This cycle should be adopted to the different sample formats

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+               int32_t value = *mixed++ & 0xff;
+               value += *mixed++ << 8;
+               wavpackBuffer[j*numChannels + i] = value;
+            }
+         }
+
+         if (!WavpackPackSamples(wpc, wavpackBuffer.get(), samplesThisRun))
+            updateResult = ProgressResult::Cancelled;
```
crsib, line 352: This can be `return ProgressResult::Failed` now, as memory managment is handled by C++ now

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+         if (updateResult == ProgressResult::Success)
+            updateResult =
+               progress.Update(mixer->MixGetCurrentTime() - t0, t1 - t0);
+      }
+   }
+
+   if (!WavpackFlushSamples(wpc)) {
+      updateResult = ProgressResult::Cancelled;
```
crsib, line 361: `return ProgressResult::Failed`?

And I think that `BasicUI::ShowErrorDialog` should be used to notify the user about the error in places where codec has failed. 

(Please, see ExportPCM for example)

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+      wxString n;
+      for (const auto &pair : metadata->GetRange()) {
+         n = pair.first;
+         const auto &v = pair.second;
+         if (n == TAG_YEAR) {
+            n = wxT("DATE");
+         }
+         WavpackAppendTagItem(wpc,
+                              (char *) (const char *) n.mb_str(wxConvUTF8),
```
crsib, line 374: Both conversions are redundant

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+                              (char *) (const char *) v.mb_str(wxConvUTF8),
```
crsib, line 375: Ditto

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+                              (int) v.length());
+      }
+
+      if (!WavpackWriteTag(wpc)) {
+         // Not sure what to do
+      }
+   }
+
+   WavpackCloseFile(wpc);
```
crsib, line 384: No longer needed!

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+   }
+
+   return updateResult;
+}
+
+// Based on the implementation of write_block in dbry/WavPack
+// src: https://github.com/dbry/WavPack/blob/master/cli/wavpack.c
+int ExportWavPack::WriteBlock(void *id, void *data, int32_t length)
+{
+   WriteId *outId = (WriteId *) id;
```
crsib, line 398: `static_cast`

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+   uint32_t bcount;
```
crsib, line 400: Unused!

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+   if (outId->error)
+      return FALSE;
```
crsib, line 403: `return false;`

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+   if (outId && outId->file && outId->file.get() && data && length) {
```
crsib, line 405: This function can be rewritten with less nesting involved!

```
int ExportWavPack::WriteBlock(void *id, void *data, int32_t length)
{
    if (id == nullptr || data == nullptr || length == 0)
        return true; // This is considered to be success in wavpack.c reference code

    WriteId *outId = static_cast<WriteId*>(id);

    if (!outId->file)
        // This does not match the wavpack.c but in our case if file is nullptr - 
        // the stream error has occured
        return false; 

    if (!outId->file->Write(data, length).IsOk()) {
        outId->file.reset();
        return false;
    }

    outId->bytesWritten += length;

    if (outId->firstBlockSize == 0)
        outId->firstBlockSize = length;

    return true;
}
```

Also, this allows to remove `error` from the WriteId structure!

crsib (changes requested):
When I export from Audacity - the resulting file seems to be missing some metadata: VLC does not show the file length and does not allow seeking. Everything works fine, if I generate wv file using the official CLI.

Looks like you miss `WavpackUpdateNumSamples` call in the end.

This concludes the review of the ExportWavPack.cpp, review for the importer will follow.

crsib (changes requested):
When I export from Audacity - the resulting file seems to be missing some metadata: VLC does not show the file length and does not allow seeking. Everything works fine, if I generate wv file using the official CLI.

Looks like you miss `WavpackUpdateNumSamples` call in the end.

This concludes the review of the ExportWavPack.cpp, review for the importer will follow.

Subhra264:
> When I export from Audacity - the resulting file seems to be missing some metadata: VLC does not show the file length and does not allow seeking. Everything works fine, if I generate wv file using the official CLI.
> 
> Looks like you miss `WavpackUpdateNumSamples` call in the end.

I have replaced `FileIO` with `wxFile` as seeking to the first block of the file was required. I have tested the wavpack files generated from audacity with VLC media player. Now, it seems to work correctly as it is now allowing seeking and the audio length is also shown. Also, I have addressed the changes you asked me to make. The latest commit is now ready for another review.

src/export/ExportWavPack.cpp
```diff
@@ -371,21 +413,39 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
...
-
    if ( !outWvFile.file.get()->Close()
       || ( outWvcFile.file && outWvcFile.file.get() && !outWvcFile.file.get()->Close())) {
-      updateResult = ProgressResult::Cancelled;
+      return ProgressResult::Failed;
+   }
+
+   // wxFile::Create opens the file with only write access
+   // So, need to open the file again with both read and write access
+   if (!outWvFile.file->Open(fName.GetFullPath(), wxFile::read_write)) {
```
crsib, line 434: For simplicity we can cache that block in memory, but this will do too. Very good!

src/export/ExportWavPack.cpp
```diff
@@ -425,6 +486,4 @@ void ExportWavPack::OptionsCreate(ShuttleGui &S, int format)
 
 static Exporter::RegisteredExportPlugin sRegisteredPlugin{ "WavPack",
    []{ return std::make_unique< ExportWavPack >(); }
-};
-
-#endif
+};
```
crsib, line 489: We generally prefer to have an empty new line at the end of the file

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,489 @@
...
+{
+   return true;
+}
+
+bool ExportWavPackOptions::TransferDataFromWindow()
+{
+   ShuttleGui S(this, eIsSavingToPrefs);
+   PopulateOrExchange(S);
+
+   CreateCorrectionFileSetting.Write( mCreateCorrectionFile->GetValue() );
```
crsib, line 217: Hmm, I'm not sure why this is needed. 

And I think we need to flush settings here

crsib (commented):
Export looks fine now, I will review import now!

Subhra264:
> Export looks fine now, I will review import now!

Ok then, I will commit and push the remaining ExportWavPack changes along with ImportWavPack changes.

src/import/ImportWavPack.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+            // Get the actual length of the item key at this index i
+            itemLen = WavpackGetTagItemIndexed(mWavPackContext, i, NULL, 0);
+            item = (char *)malloc(itemLen + 1);
+            WavpackGetTagItemIndexed(mWavPackContext, i, item, itemLen + 1);
+            name = UTF8CTOWX(item);
+
+            // Get the actual length of the value for this item key
+            valueLen = WavpackGetTagItem(mWavPackContext, item, NULL, 0);
+            itemValue = (char *)malloc(valueLen + 1);
```
crsib, line 241: Ditto

crsib (changes requested):
Decoding loop need some love :-)

And some minor are changes will be welcome!

src/export/ExportWavPack.cpp
```diff
@@ -0,0 +1,502 @@
...
+            for (decltype(samplesThisRun) j = 0; j < samplesThisRun; j++) {
+               for (size_t i = 0; i < numChannels; i++) {
+                  wavpackBuffer[j*numChannels + i] = *mixed++;
+               }
+            }
+         } else {
+            const float *mixed = reinterpret_cast<const float*>(mixer->GetBuffer());
+            for (decltype(samplesThisRun) j = 0; j < samplesThisRun; j++) {
+               for (size_t i = 0; i < numChannels; i++) {
+                  int64_t intValue = (*mixed++) * (1u << 31);
```
Subhra264, line 391: In the `ExportWavPack.cpp` file, I think there were some issues with float 32 bit export. Earlier, I just passed 32 bit int values for float sample export. So, I think while importing it was getting imported as 32 bit int (before adding special cases for float sample import as you suggested). Though it was ok while importing in audacity ( wtih the previous code before applying your suggestions for import), I tried to play the 32 bit exported wavpack file using the VLC media player and it sounded like corrupted. Now, when I converted the `mixed` data into `float` and then into `int32_t` while exporting as 32 bit float, both VLC media and audacity (After applying your suggestions) opened and played the 32 bit float files correctly.

crsib: I think you should set `config->float_norm_exp: = 127;` to enable FP mode in Wavpack. In any case - you should consult `WavpackGetMode` to check, if you need to convert the values.

`1u << 31` - `std::numeric_limits<int32_t>::max()` please.

crsib: (Value `127` is an exponent  bias for IEEE754 32-bit float values. We have 8 bits for exponents, so the bias is `2^7 - 1`)

crsib (changes requested):
Great work! Just a few more issue, but I think that they are easy to resolve.

src/export/ExportWavPack.cpp
```diff
@@ -388,7 +389,7 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
             const float *mixed = reinterpret_cast<const float*>(mixer->GetBuffer());
             for (decltype(samplesThisRun) j = 0; j < samplesThisRun; j++) {
                for (size_t i = 0; i < numChannels; i++) {
-                  int64_t intValue = (*mixed++) * (1u << 31);
+                  int64_t intValue = (*mixed++) * (std::numeric_limits<int32_t>::max());
```
crsib, line 392: Minor, but I expect a compiler warning here. Let's add `static_cast<int64_t>` here. Sorry I've missed that before...

src/import/ImportWavPack.cpp
```diff
@@ -0,0 +1,298 @@
...
+   {
+      const uint32_t bufferSize = mNumChannels * SAMPLES_TO_READ;
+      ArrayOf<int32_t> wavpackBuffer{ bufferSize };
+      uint32_t samplesRead = 0;
+
+      do {
+         samplesRead = WavpackUnpackSamples(mWavPackContext, wavpackBuffer.get(), SAMPLES_TO_READ);
+
+         if (mFormat == int16Sample) {
+            ArrayOf<int16_t> tempBuffer{ bufferSize };
```
crsib, line 187: Not that important here, but is is best to avoid memory allocations in such loops. Allocations can be quite costly.

crsib (changes requested):
A one very minor step left!

crsib (approved):

src/export/ExportWavPack.cpp
wthrajat: @crsib I am curious to know why :D

Issue #2225, opened 2021-12-03
WIN32: cannot configure portmixer.


carlo-bramini:
Hello, I tried to configure Audacity for building a native build for Windows, but I got this error:

```
CMake Error at /home/carlo/audacity/lib-src/portmixer/cmake/CheckPaWinMME_GetStreamInputHandle/CMakeLists.txt:16 (find_package):
  By not providing "FindPortAudio.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "PortAudio", but CMake did not find one.

  Could not find a package configuration file provided by "PortAudio" with
  any of the following names:

    PortAudioConfig.cmake
    portaudio-config.cmake

  Add the installation prefix of "PortAudio" to CMAKE_PREFIX_PATH or set
  "PortAudio_DIR" to a directory containing one of the above files.  If
  "PortAudio" provides a separate development package or SDK, be sure it has
  been installed.


CMake Error at lib-src/portmixer/CMakeLists.txt:26 (try_compile):
  Failed to configure test project build system.
```
At the moment, according to the suggestion received from CMake, I added this line:

```
set( CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../../../cmake-proxies/cmake-modules" )
```
to the `CMakeLists.txt` into `lib-src\portmixer\cmake\CheckPaWinDS_GetDeviceGUID` and `lib-src\portmixer\cmake\CheckPaWinMME_GetStreamInputHandle` and the problem seems solved. What do you think?

crsib:
Worth trying for sure, although it is odd that it has popped up, as CI surely works fine :-(

carlo-bramini:
> Worth trying for sure, although it is odd that it has popped up, as CI works fine :-(

This is the current code:
```
set( MODULE_DIR "${CMAKE_BINARY_DIR}/../../.." )

set( CMAKE_MODULE_PATH 
    ${MODULE_DIR}
    ${CMAKE_MODULE_PATH}
)
```
To tell you the truth, I don't understand it.

LWinterberg:
The build system has changed since this was made, try again. Closing bug as stale.

PR #2235, opened 2021-12-05
WIN32: fix configuration of portmixer

carlo-bramini:
Fix issue #2225


crsib:
Hi Carlo! Sorry for bothering - can you please rebase onto the latest master?

carlo-bramini:
I'm not able to understand why detection of PortAudio fails with CMake-Gui and conan with MSVC (but now works with bash and without conan), since I just added an extra path to `CMAKE_MODULE_PATH`, sorry.

Issue #2239, opened 2021-12-05
Crash when cancelling Nyquist Prompt


SteveDaulton:
**Describe the bug**
When cancelling the Nyquist Prompt, Audacity may crash

**To Reproduce**
1. Generate a tone
2. Open the Nyquist Prompt and enter the code:

;control test "Test" float "" 5 0 10
(print test)

3. Click the "OK" button
4. Click the "Cancel" button
5. Observe crash

**Expected behavior**
Not crash

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version: Audacity 3.2.0-alpha-20211205

**Additional context**
This is a regression

Paul-Licameli:
I introduced this bug.  It happens only in the development branch.  See the linked PR for the fix.

PR #2240, opened 2021-12-05
Fix Issue #2239: crash cancelling Nyquist prompt...

Paul-Licameli:
... To reproduce:

* Generate a tone
* Open the Nyquist Prompt and enter the code:

```
;control test "Test" float "" 5 0 10
(print test)
```

* Click the "OK" button
* Click the "Cancel" button
* Observe crash

Bug was introduced at b1e988e4f50ff19050d00c8c5685d43a9fc8a62b
The nulling-out of a pointer to the Effect in the Close routine was deleted,
but the destructor is still called later.

To fix it, don't unnecessarily repeat in the destructor of the modal dialog
(which may be delayed until an event loop) what was already done when closing
it.

Resolves: #2239

Recommended:

### DIFF
--- a/src/effects/EffectUI.cpp
+++ b/src/effects/EffectUI.cpp
@@ -757,10 +757,7 @@ EffectUIHost::EffectUIHost(wxWindow *parent,
 
 EffectUIHost::~EffectUIHost()
 {
-   CleanupRealtime();
-   if (mNeedsResume)
-      Resume();
-   mClient.CloseUI();
+   wxASSERT(mClosed);
 }
 
 // ============================================================================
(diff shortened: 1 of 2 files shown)


src/effects/EffectUI.cpp
```diff
@@ -757,10 +757,7 @@ EffectUIHost::EffectUIHost(wxWindow *parent,
 
 EffectUIHost::~EffectUIHost()
 {
-   CleanupRealtime();
-   if (mNeedsResume)
-      Resume();
-   mClient.CloseUI();
+   wxASSERT(mClosed);
```
vsverchinsky, line 760: `mClosed` is a new member field, which seems to be used only in debug builds, since `wxASSERT` macro does nothing in release builds. Couldn't the definition of the variable be made conditional too?

vsverchinsky (approved):

src/effects/EffectUI.cpp
Paul-Licameli: As you wish

Issue #2114, opened 2021-11-11
Audacity freezes when it been closed with opened Logging window


Penikov:
**Describe the bug**
Audacity freezes when it been closed with opened Logging window

**To Reproduce**
1. Run Audacity
2. Open Help-Diagnostics-Show log
3. Close Audacity
4. Close logging dialog
5. See Audacity freezed process in the System Monitor

**Expected behavior**
Audacity closes and opens normally

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Linux, Windows
 - Version: Audacity 3.1 +

PR #2245, opened 2021-12-06
Destroys the log when user quits Audacity

crsib:
Resolves: #2114 

Recommended:


Paul-Licameli (approved):
This looks simple.

Paul-Licameli:
Should this crash fix go to a release branch?

crsib:
That makes sense, looks like the bug was introduced in 3.1.0

Issue #2226, opened 2021-12-03
Joining multiple clips sometimes causes data loss


jserquera:
When trying to join many audio blocks at once, the result is not well done. I mean, using the Edit menu (I have the Spanish version: Recortar limites-> Unir).

You have to go one by one.

I think it happens in all Windows and Mac versions....

LWinterberg:
In the latest 3.1.3 build, having clips like this:
![image](https://user-images.githubusercontent.com/87814144/144589079-7b82e858-6759-4dec-a217-8fa333f03431.png)

causes Select all followed by  Edit > Clip Boundaries > Join to make this
![image](https://user-images.githubusercontent.com/87814144/144589171-f7e0746d-cb05-44bb-be22-9a3a92f32a3a.png)

Which seems correct to me. 

@jserquera can you provide more information? What exactly is the problem? What Audacity version are you using? Can you maybe send a sample of the AUP3 file (you'll need to ZIP it first)?

DavidBailes:
Steps to reproduce:
1. Split an audio clip into three clips
2. Move the clips so that there are spaces between the clips.
3. Select all the audio (Ctrl+A)
4. Join the clips (Ctrl+J)
5. Observe that the space between the second and third clips is removed, rather than being replaced with silence.

LWinterberg:
following your STR results in the following clip: 
![image](https://user-images.githubusercontent.com/87814144/144596497-ad51abde-7b21-4078-a81f-04efafc54361.png)

I still don't understand what the problem is.

jserquera:
Attached is a project that gives me the error
[error.zip](https://github.com/audacity/audacity/files/7649327/error.zip)

DavidBailes:
Using commit d2afbf (3.1.3 beta 1):
clips selected:
![image](https://user-images.githubusercontent.com/11887021/144599805-82c2c9f0-9243-408d-be7e-dd2d8bb6f8f8.png)

clips joined:
![image](https://user-images.githubusercontent.com/11887021/144599869-0fa3f782-719d-4cd0-9fb6-b583906df02e.png)

LWinterberg:
Ah. I'm on commit 9ca1f56 / [CMake Build #3257](../actions/runs/1530570708) so it might be fixed in the final 3.1.3 after all.

jserquera:
Did you try with the project I sent?

SteveDaulton:
[...]

PR #2246, opened 2021-12-07
Issue #2226: Joining multiple clips sometimes time-shifts them

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/2226

Steps to reproduce:

1. Split an audio clip into three clips
2. Move the clips so that there are spaces between the clips.
3. Select all the audio (Ctrl+A)
4. Join the clips (Ctrl+J)
5. Observe that clips are not joined correctly.

Problem:
In  WaveClip::InsertSilence(), the following two lines:

else if (t == GetPlayEndTime() && t < GetSequenceEndTime())
      ClearSequence(t, GetSequenceEndTime());

After ClearSequence is called to remove the data to the end of the sequence, then there is not call to update mTrimRight, as is done in WaveClip::ClearRight(). So when more data is pasted, it's pasted in the wrong place.

Fix:
Update mTrimeRight.

Notes:
1. I don't know why the bug was not always reproduceable.
2. There appears to be a similar issue with the previous two lines:

   if (t == GetPlayStartTime() && t > GetSequenceStartTime())
      ClearSequence(GetSequenceStartTime(), t);

Compare with WaveClip::ClearLeft().
However I have left that unfixed, both because it doesn't affect this bug, and also because I couldn't work out when that call to ClearSequence would get called, and so I couldn't test any fix.

Recommended:

### DIFF
--- a/src/WaveClip.cpp
+++ b/src/WaveClip.cpp
@@ -1382,8 +1382,10 @@ void WaveClip::InsertSilence( double t, double len, double *pEnvelopeValue )
 {
    if (t == GetPlayStartTime() && t > GetSequenceStartTime())
       ClearSequence(GetSequenceStartTime(), t);
-   else if (t == GetPlayEndTime() && t < GetSequenceEndTime())
+   else if (t == GetPlayEndTime() && t < GetSequenceEndTime()) {
       ClearSequence(t, GetSequenceEndTime());
+      SetTrimRight(.0);
+   }
 
    auto s0 = TimeToSequenceSamples(t);
    auto slen = (sampleCount)floor(len * mRate + 0.5);


vsverchinsky (approved):

vsverchinsky:
I also tried to find when the code path for the left border is executed and didn't find any too. 
Thank you!

LWinterberg:
@vsverchinsky this is now in master - do we also want this in release-3.1.3?

vsverchinsky:
> @vsverchinsky this is now in master - do we also want this in release-3.1.3?

@LWinterberg yep, I've cherry-picked it into 3.1.3:
https://github.com/audacity/audacity/commit/c47d856be55359fd9abb91676203b708215927aa

Issue #2199, opened 2021-11-29
Mac Clip names can't be renamed at sync-locked tracks on macOS


Penikov:
**Describe the bug**
Clip names couldn't be renamed at sync-locked tracks on Mac OS

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a couple of tracks with clips

2. Turn On [Menu]-[Tracks]-[Sync Lock Tracks]

4. Try to rename the clip

If you click the 'cancel' button, Audacity will crash. This is another issue: #2198

**Expected behaviour**

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Mac OS Monterey
 - Version: Audacity 3.1.2+

Issue #2198, opened 2021-11-29
Clip Names. Audacity crashes when a user clicks 'Cancel' at clip renaming dialog


Penikov:
**Describe the bug**

 Audacity crashes when a user clicks 'Cancel' at clip renaming dialog

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a couple of tracks with clips

2. Turn On [Menu]-[Tracks]-[Sync Lock Tracks]

3. Go to [Preferences]-[Tracks]-[Tracks Behaviors] turn on the tickbox "use the dialogue for the name of new label"
![изображение](https://user-images.githubusercontent.com/25851359/143918281-72de7439-b329-4779-8a07-53c80a05a69c.png)

4. Open the 'Set wave clip name' dialog with Double-click
5. Click the 'Cancel' button and close the dialog

**Additional information (please complete the following information):**
 - OS: Windows 10, Mac OS Monterey
 - Version: Audacity 3.1.2 +

**Additional context**

Steps to reproduce on Windows:  are the same, except the last step. 
The last step on Windows should be:
5. Click the 'Close' button 
![изображение](https://user-images.githubusercontent.com/25851359/144010663-48e8b1ce-fd55-4820-8573-b9c9e34d1ccd.png)

relates to #2199

More details: https://youtu.be/trb0yqqw96Q?t=340

AnitaBats:
reproducible on Windows as well

PR #2249, opened 2021-12-08
Issue 2198: fix clip rename

vsverchinsky:
Resolves: #2198
Resolves: #2199

Interaction with modal dialog causes Cancel event to be dispatched before dialog itself is finished (easily reproducible in MacOS). Then ProjectHistory::Rollback in TimeShiftHandle::Cancel replaces all tracks in the current track list, so that track captured by TimeShiftHandle becomes orphaned. Another issue is that returning RefreshCode::Cancelled from WaveTrackAffordanceHandle::Click also causes ProjectHistory::Rollback and as a result tracks changes made to the clip name reverted. 

Recommended:


Paul-Licameli (approved):
Please remember when you merge to copy useful comments from the top of the thread into the text entry, before you hit the green button.

Issue #2178, opened 2021-11-23
Create different dialogs for crashes and errors in Audacity


AnitaBats:

crsib:
![image](https://user-images.githubusercontent.com/2660628/145037311-0e3028ee-5e3b-4619-91cd-037ad2dec188.png)

Tantacrul:
So we want to do an error report even though the file opens OK (in the warning shared by Leo)?

I'm wondering if we might place a simpler dialog before this error report. Something to the effect: "Hey, your file is fine but we've noticed a small error. (send error report)". Then if the user chooses, we open the error report dialog.

Any other suggestions?

crsib:
For starters, I think we need to hide "Problem details" under the cut.

For the case from Leo - we should change the error type.

Generally, we should revise all error messages in Audacity. They are extremely unhelpful. Our second most often error right now is literally "Error opening recording device. Error code: 0 Success." 

![image](https://user-images.githubusercontent.com/2660628/145040332-0d011bf5-a44d-4109-9f54-89a0a0f56c28.png)

crsib:
@Tantacrul 

![image](https://user-images.githubusercontent.com/2660628/145376972-f3be4a0b-e73c-4da3-961d-bb9fd17d777b.png)
![image](https://user-images.githubusercontent.com/2660628/145377007-ba29f373-7cc0-4072-8586-f1e96c712c03.png)

Tantacrul:
As far as a UX improvement for error dialogs go, it's great. No comment.

One thing: regarding the issue in the title of this page - about people thinking mistaking warnings for crash errors - will this help solve that issue?

How would it differ from a warning message?

Tantacrul:
In general, warnings have the (!) yellow triangle and errors have the red (!) circle. I think they're system icons, right?

crsib:
This issue was more about error vs. crash to be true :-) An icon is from the system, yes.

PR #2251, opened 2021-12-09
Error Dialog UI

crsib:
Resolves: #2178 

This PR fixes two issues:

1. ErrorReportDialog is now smaller and looks different from the crash reporter.
2. "Audio is muted" is considered to be a user error, i. e. it no longer can be reported to Sentry

Recommended:


Paul-Licameli (approved):
More deep dives into wxWidgets, oh joy

Issue #2066, opened 2021-11-01
Boundary snap guides do not appear when resizing clips


SteveDaulton:
**Describe the bug**
Moving clips to align with clips / labels in other tracks is aided by "boundary snap guides" (https://manual.audacityteam.org/man/boundary_snap_guides.html). Boundary snap guides do not appear when resizing clips, making it extremely difficult to precisely resize a clip to align with a clip in another track or with a label.

**To Reproduce**
1. Generate about 30 seconds of audio in one track, and  about 15 seconds of audio in another track
2. Drag the left edge of the first track to precisely align with the end of the first track
3. Observe how difficult it is to do precisely.

**Expected behavior**
Boundary snap guides to work when resizing clips

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version Audacity 3.1.0 AppImage

PR #2261, opened 2021-12-10
Issue 2066: clip trim snapping

vsverchinsky:
Resolves: #2066 

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
@@ -13,6 +13,9 @@
 #include "RefreshCode.h"
 
 
+#include "../../../../TrackArtist.h"
+#include "../../../../Snap.h"
+#include "../../../../TrackPanelDrawingContext.h"
 #include "../../../images/Cursors.h"
 #include "WaveClip.h"
 #include "WaveTrack.h"
@@ -39,15 +42,19 @@ namespace {
     }
 }
 
-WaveClipTrimHandle::ClipTrimPolicy::~ClipTrimPolicy() { }
+WaveClipTrimHandle::ClipTrimPolicy::~ClipTrimPolicy() = default;
 
 class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTrimPolicy
 {
+   std::shared_ptr<WaveTrack> mTrack;
    std::vector<std::shared_ptr<WaveClip>> mClips;
    double mInitialBorderPosition{};
    int mDragStartX{ };
    std::pair<double, double> mRange;
    bool mAdjustingLeftBorder;
+
+   std::unique_ptr<SnapManager> mSnapManager;
+   SnapResults mSnap;
 
    void TrimTo(double t)
    {
@@ -64,9 +71,60 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
       }
    }
 
+   //Search for a good snap points among all tracks, including
+   //the one to which adjusted clip belongs to, but not counting
+   //the borders of adjusted clip
+   static SnapPointArray FindSnapPoints(
+      const WaveTrack* currentTrack, 
+      WaveClip* adjustedClip,
+      const std::pair<double, double> range)
+   {
+      SnapPointArray result;
+
+      auto addSnapPoint = [&](double t, const Track* track)
+      {
+         if(t > range.second || t < range.first)
+            return;
+
+         for(const auto& snapPoint : result)
+            if(snapPoint.t == t)
+               return;
+         result.emplace_back(t, track);
+      };
+
+      if(const auto trackList = currentTrack->GetOwner())
+      {
+         for(const auto track : trackList->Any())
+         {
+            const auto isSameTrack = (track == currentTrack) ||
+               (track->GetLinkType() == Track::LinkType::Aligned && *trackList->FindLeader(currentTrack) == track) ||
+               (currentTrack->GetLinkType() == Track::LinkType::Aligned && *trackList->FindLeader(track) == currentTrack);
+            for(const auto& interval : track->GetIntervals())
+            {
+               if(isSameTrack)
+               {
+                  auto waveTrackIntervalData = dynamic_cast<WaveTrack::IntervalData*>(interval.Extra());
+                  if(waveTrackIntervalData->GetClip().get() == adjustedClip)
+                  //exclude boundaries of the adjusted clip
+                     continue;
+               }
+               addSnapPoint(interval.Start(), track);
+               if(interval.Start() != interval.End())
+                  addSnapPoint(interval.End(), track);
+            }
+         }
+      }
+      return result;
+   }
+
 public:
-   AdjustBorder(WaveTrack* track, std::shared_ptr<WaveClip>& clip, bool left)
-      : mAdjustingLeftBorder(left)
+   AdjustBorder(
+      const std::shared_ptr<WaveTrack>& track, 
+      const std::shared_ptr<WaveClip>& clip, 
+      bool leftBorder, 
+      const ZoomInfo& zoomInfo)
+      : mTrack(track),
+        mAdjustingLeftBorder(leftBorder)
    {
       auto clips = track->GetClips();
 
@@ -74,19 +132,16 @@ public:
 
       if (track->IsAlignedWithLeader() || track->GetLinkType() == Track::LinkType::Aligned)
          //find clips in other channels which are also should be trimmed
-         mClips = FindClipsInChannels(clip->GetPlayStartTime(), clip->GetPlayEndTime(), track);
+         mClips = FindClipsInChannels(clip->GetPlayStartTime(), clip->GetPlayEndTime(), track.get());
       else
          mClips.push_back(clip);
 
       if (mAdjustingLeftBorder)
       {
-         auto left = -std::numeric_limits<double>::infinity();
-         for (auto& other : clips)
+         auto left = clip->GetSequenceStartTime();
+         for (const auto& other : clips)
             if (other->GetPlayStartTime() < clip->GetPlayStartTime() && other->GetPlayEndTime() > left)
-            {
-               auto maxOffset = clip->SamplesToTime(clip->TimeToSamples(clip->GetPlayStartTime() - other->GetPlayEndTime()));
-               left = clip->GetPlayStartTime() - maxOffset;
-            }
+               left = other->GetPlayEndTime();
          //not less than 1 sample length
          mRange = std::make_pair(left, clip->GetPlayEndTime() - 1.0 / clip->GetRate());
 
@@ -94,17 +149,22 @@ public:
       }
       else
       {
-         auto right = std::numeric_limits<double>::infinity();
-         for (auto& other : clips)
+         auto right = clip->GetSequenceEndTime();
+         for (const auto& other : clips)
             if (other->GetPlayStartTime() > clip->GetPlayStartTime() && other->GetPlayStartTime() < right)
-            {
-               auto maxOffset = clip->SamplesToTime(clip->TimeToSamples(other->GetPlayStartTime() - clip->GetPlayEndTime()));
-               right = clip->GetPlayEndTime() + maxOffset;
-            }
+               right = other->GetPlayStartTime();
          //not less than 1 sample length
          mRange = std::make_pair(clip->GetPlayStartTime() + 1.0 / clip->GetRate(), right);
 
          mInitialBorderPosition = mClips[0]->GetPlayEndTime();
+      }
+
+      if(const auto trackList = track->GetOwner())
+      {
+         mSnapManager = std::make_unique<SnapManager>(
+            *trackList->GetOwner(),
+            FindSnapPoints(track.get(), clip.get(), mRange),
+            zoomInfo);
       }
    }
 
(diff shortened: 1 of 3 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.h
```diff
@@ -29,20 +29,31 @@ class WaveClipTrimHandle : public UIHandle
...
+       virtual void Draw(
+           TrackPanelDrawingContext &context, 
+           const wxRect &rect, 
+           unsigned iPass);
+
+       virtual wxRect DrawingArea(
+           TrackPanelDrawingContext&, 
+           const wxRect &rect, 
+           const wxRect &panelRect, 
+           unsigned iPass);
```
Paul-Licameli, line 45: One of the C++ Core Guidelines I like is that each virtual function declaration should be marked with exactly one of `virtual` `override` or `final` as appropriate, and for these two it should be `override`.

Paul-Licameli: Oh, I see I misunderstood the context.  This is not overriding UIHandle functions.  `virtual` is right.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -64,48 +71,105 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
...
+         result.emplace_back(t, track);
+      };
+
+      if(const auto trackList = currentTrack->GetOwner())
+      {
+         for(const auto track : trackList->Any())
+         {
+            if(track == currentTrack ||
+               track->GetLinkType() == Track::LinkType::Aligned && *trackList->FindLeader(currentTrack) == track ||
+               currentTrack->GetLinkType() == Track::LinkType::Aligned && *trackList->FindLeader(track) == currentTrack)
```
Paul-Licameli, line 101: Clang gives me a warning about '&&' within '||'.  Add parentheses to make the intent unmistakable.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -64,48 +71,105 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
...
+            {
+               //serching on the same track is a special case
```
Paul-Licameli, line 103: Same track, or aligned partner track

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -64,48 +71,105 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
...
+               for(const auto& clip : static_cast<const WaveTrack*>(track)->GetClips())
```
Paul-Licameli, line 104: I would prefer to iterate the intervals, as in the other branch, not the clips, for uniformity, and in case WaveTrack might add some other kind of interval in future (who knows?)

You can static_cast the Extra() pointer of each interval to WaveTrack::IntervalData* and then test for the clip you want to exclude.

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -64,48 +71,105 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
...
+         }
+      }
+      return result;
+   }
+
 public:
-   AdjustBorder(WaveTrack* track, std::shared_ptr<WaveClip>& clip, bool left)
-      : mAdjustingLeftBorder(left)
+   AdjustBorder(std::shared_ptr<WaveTrack>& track, std::shared_ptr<WaveClip>& clip, bool leftBorder, const ZoomInfo& zoomInfo)
+      : mTrack(track),
```
Paul-Licameli, line 130: The two shared_ptr arguments could be const references, and so they should, to aid the human reader.  Instead I read this then stopped to see whether `track` and `clip` were being reassigned for some reason.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -64,48 +71,105 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
...
          //not less than 1 sample length
          mRange = std::make_pair(left, clip->GetPlayEndTime() - 1.0 / clip->GetRate());
 
          mInitialBorderPosition = mClips[0]->GetPlayStartTime();
       }
       else
       {
-         auto right = std::numeric_limits<double>::infinity();
+         auto right = clip->GetSequenceEndTime();
          for (auto& other : clips)
```
Paul-Licameli, line 157: A const auto & here for consistency

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -268,11 +372,23 @@ HitTestPreview WaveClipTrimHandle::HitPreview(const AudacityProject*, bool unsaf
...
+
+wxRect WaveClipTrimHandle::ClipTrimPolicy::DrawingArea(TrackPanelDrawingContext&, const wxRect&, const wxRect&, unsigned)
 {
+   return {};
 }
 
-UIHandlePtr WaveClipTrimHandle::HitAnywhere(std::weak_ptr<WaveClipTrimHandle>& holder, WaveTrack* waveTrack, const AudacityProject* pProject, const TrackPanelMouseState& state)
+UIHandlePtr WaveClipTrimHandle::HitAnywhere(
+   std::weak_ptr<WaveClipTrimHandle>& holder, 
+   std::shared_ptr<WaveTrack>& waveTrack,
```
Paul-Licameli, line 389: Another argument might be const reference

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.h
```diff
@@ -52,15 +63,26 @@ class WaveClipTrimHandle : public UIHandle
...
+    // TrackPanelDrawable implementation
+
+    void Draw(TrackPanelDrawingContext &context,
+        const wxRect &rect,
+        unsigned iPass ) override;
+
+    wxRect DrawingArea(TrackPanelDrawingContext&,
+        const wxRect &rect,
+        const wxRect &panelRect,
+        unsigned iPass) override;
```
Paul-Licameli, line 87: See how SelectHandle overrides Enter, HasEscape, and Escape member functions.  These allow you to hit the Esc key and dismiss the yellow snap lines, enabling a fine, un-snapped adjustment, if that is what the user really wants.  The same might be done here.

Maybe some common base class, derived from UIHandle, should be defined?  Hm.

Paul-Licameli: You may note that I could have done the extra work I suggested in the UIHandle subclasses recently added to AdornedRulerPanel, but I didn't.

This is not mandatory extra work.  But I should reconsider how the pattern might be made easier and reusable across UIHandle classes that snap.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -268,11 +372,23 @@ HitTestPreview WaveClipTrimHandle::HitPreview(const AudacityProject*, bool unsaf
...
+   : mClipTrimPolicy{ std::move(clipTrimPolicy) }
+{
+
+}
+
+void WaveClipTrimHandle::ClipTrimPolicy::Draw(TrackPanelDrawingContext&, const wxRect&, unsigned) { }
+
+wxRect WaveClipTrimHandle::ClipTrimPolicy::DrawingArea(TrackPanelDrawingContext&, const wxRect&, const wxRect&, unsigned)
 {
+   return {};
```
Paul-Licameli, line 384: See TrackPanelDrawable::DrawingArea:  default behavior should not be to return an empty wxRect, but to return the second argument unchanged.

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
vsverchinsky: Funny, the tool I use to analyze a code suggests quite opposite: it says that parenthesizes are redundant :) But I usually tend to put them too. Thank you!

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
vsverchinsky: Didn't think of that. Thank you!

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.h
vsverchinsky: > See how SelectHandle overrides Enter, HasEscape, and Escape member functions. These allow you to hit the Esc key and dismiss the yellow snap lines, enabling a fine, un-snapped adjustment, if that is what the user really wants. The same might be done here.
> 
> Maybe some common base class, derived from UIHandle, should be defined? Hm.

If I try to hit escape when using a select handle it just resets the current selection, with a debugger attached I see that both `::Escape` and `::Cancel` member functions are triggered, did I misunderstood something?

Paul-Licameli: You can hit escape to dismiss the yellow snap line before the click.  It seems you can't dismiss the yellow snap line while dragging.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -64,48 +71,101 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
...
+         for(const auto track : trackList->Any())
+         {
+            const auto isSameTrack = (track == currentTrack) ||
+               (track->GetLinkType() == Track::LinkType::Aligned && *trackList->FindLeader(currentTrack) == track) ||
+               (currentTrack->GetLinkType() == Track::LinkType::Aligned && *trackList->FindLeader(track) == currentTrack);
+            for(const auto& interval : track->GetIntervals())
+            {
+               if(isSameTrack)
+               {
+                  auto waveTrackIntervalData = dynamic_cast<WaveTrack::IntervalData*>(interval.Extra());
```
Paul-Licameli, line 106: Just a discussion...

I would prefer `static_cast` myself here...  But actually, I'd rather see `std::any_cast`.  The Extra information in IntervalData looks like a place I could rewrite things with `std::any` now that we use C++17.  I think `any_cast` makes a simpler check than `dynamic_cast`.  Just equality of a statically determined `typeid` of the template parameter with the dynamically determined `typeid` in the `any`.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -268,11 +368,23 @@ HitTestPreview WaveClipTrimHandle::HitPreview(const AudacityProject*, bool unsaf
...
+   : mClipTrimPolicy{ std::move(clipTrimPolicy) }
+{
+
+}
+
+void WaveClipTrimHandle::ClipTrimPolicy::Draw(TrackPanelDrawingContext&, const wxRect&, unsigned) { }
+
+wxRect WaveClipTrimHandle::ClipTrimPolicy::DrawingArea(TrackPanelDrawingContext&, const wxRect&, const wxRect&, unsigned)
 {
+   return {};
```
Paul-Licameli, line 380: Still not resolved

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.h
vsverchinsky: Does not seem to work as well, I still get snap lines during the selection

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
crsib: I've never seen much performance impact from using `dynamic_cast`. It can be slow in tight loops with deep hierarchies, but that's not the case here.

vsverchinsky: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Pro-type-downcast

I agree that sometimes we can justify using `static_cast` instead of `dynamic_cast` if we want to get as much performance as possible, and we do guarantee that the specified type is the only possible option here. But one of the reasons why this part of the code was rewritten is that we may want to add more interval types to a wave track in the future, wasn't it?

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
```diff
@@ -118,18 +178,40 @@ class WaveClipTrimHandle::AdjustBorder final : public WaveClipTrimHandle::ClipTr
       return false;
    }
```
Paul-Licameli, line 179: A general comment: I'm reviewing changes in the class AdjustBorders.  The other subclass of ClipTrimPolicy is AdjustBetweenBorders.  That class is not constructed anywhere.  Is there a mistake?

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
vsverchinsky: Sorry, will fix that, thanks!

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
vsverchinsky: It's not a mistake, AdjustBetweenBorders moves borders of both clips if they are close enough, and it was my understanding of how things should happen, and turned out that I was wrong. But I was told that we may need this in the future.

Paul-Licameli (approved):
Youl will fix the small matter about the rectangle, and then there is nothing else to block this.

src/tracks/playabletrack/wavetrack/ui/WaveClipTrimHandle.cpp
crsib: `std::any` has costs attached. It's not a `zero-cost` abstraction in general, although compilers may implement a fast-pass for small objects.

Paul-Licameli:
Please rebase it once more, then create a merge commit.

vsverchinsky:
> Please rebase it once more, then create a merge commit.

Rebased, but I need an approve for this PR to merge

Paul-Licameli:
> > Please rebase it once more, then create a merge commit.
> 
> Rebased, but I need an approve for this PR to merge

I'm sorry, I thought I had done that already.

Paul-Licameli (approved):
Approved!

crsib:
> > > Please rebase it once more, then create a merge commit.
> > 
> > 
> > Rebased, but I need an approve for this PR to merge
> 
> I'm sorry, I thought I had done that already.

That's my fault! I will probably remove the "push invalidates review" rule from branch protection rules

Paul-Licameli:
> > > > Please rebase it once more, then create a merge commit.
> > > 
> > > 
> > > Rebased, but I need an approve for this PR to merge
> > 
> > 
> > I'm sorry, I thought I had done that already.
> 
> That's my fault! I will probably remove the "push invalidates review" rule from branch protection rules

Was it that, or was it a dismissal of my stale review by Vitaly, above?

crsib:
Vitaly pushed and that invalidated the review

PR #2267, opened 2021-12-10
Fixes an unexpected performance regression in SpectrumView

crsib:
After merging the spectrum brush branch - a new hot path was accidentally added to spectrum rendering.

1. std::round is expensive as is. The implementation is complex and designed to handle different situations.
2. By mistake the following chain was introduced: round(float)->int->double->round(double)->int

The new implementation uses `lrintf`. By default, it rounds to nearest, and ties are rounded to even. This does not match the behavior `std::round` entirely, but given that it only slightly affects the visuals, that shouldn't be a problem

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/SpectrumView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/SpectrumView.cpp
@@ -32,6 +32,8 @@ Paul Licameli split from WaveTrackView.cpp
 #include <wx/dcmemory.h>
 #include <wx/graphics.h>
 
+#include "float_cast.h"
+
 class BrushHandle;
 class SpectralData;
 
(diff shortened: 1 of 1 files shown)


Paul-Licameli:
The original std::round by Edward is replaced by lrintf which should make a difference only with exact values at the halves, and I don't care how the difference affects the display.

Rounding values that were already rounded was a supid mistake of my own!

Paul-Licameli (approved):

crsib:
>Rounding values that were already rounded was a supid mistake of my own!

What is stupid is how `std::round` is defined for the integers:

>A set of overloads or a function template accepting an argument of any integral type. Equivalent to 2), 6), or 10), respectively (the argument is cast to double).

I would expect it to be noop

PR #2283, opened 2021-12-13
List explicit dependencies of Audacity on our libraries...

Paul-Licameli:
... Don't make it automatically dependent on all of them.

This has two advantages:

It will allow us to define optional libraries, usable by modules but not
necessary to build a reduced Audacity without the modules.

It cleans up the generated dependency graph, omitting some redundant edges,
if we allow Audacity to inherit transitive dependencies from the public link
interfaces of the library targets.

Recommended:


src/CMakeLists.txt
```diff
@@ -1341,6 +1341,24 @@ if(HAVE_CLOCK_GETTIME)
...
+# A sub-sequence of what is in libraries/CMakeLists.txt :
+   lib-string-utils-interface
+   lib-uuid-interface
+   lib-screen-geometry-interface
+   lib-project-rate-interface
+   lib-ffmpeg-support-interface
+   lib-theme-resources-interface
+   lib-track-interface
+   lib-transactions-interface
+   lib-network-manager-interface
```
crsib, line 1354: I think it is worth adding a comment, that this list is purely for building the graph.

Also, this looks very fragile. It is not documented, I do expect it to break with the next library added.

Paul-Licameli: If "break" means the build fails, then you add a line there to fix the build, before you push, and maybe remove older lines that are unnecessary because they are now transitive dependencies.  That is what I intend.

What is wrong with that?

crsib: The build won't fail and that is the problem. Or do I miss something?

Paul-Licameli: You think the build won't fail when it should fail?

But (trust me) the build does fail in my monster branch when I extract a library but neglect to update this list, because Audacity tries to `#include` a header file from the new library, which is not found.

That build breakage is "good," early fragility that forces a fix at compile time (even before link time), not "bad," late fragility that lets the program build and run but then fail at startup time.

Did you also read the PR comment too, explaining the advantages I seek from this.

crsib: Well then I really missed something. I will try to figure out tomorrow

Paul-Licameli: The libraries in the list are given to `target_link_libraries` which affects more than the link step:  it affects also the `-I` options at compile time.

Paul-Licameli: It will make sense in the morning, I trust.  Спокойной ночи.

crsib: I will recheck, it is not clear to me how the build works with networking disabled in this case.

crsib: Doesn't build, so I was right!

```
LINK : fatal error LNK1104: cannot open file 'lib-network-manager-interface.lib' [C:\devel\projects\audacity\Paul\tie\b
uild\src\Audacity.vcxproj]```

Comment (Paul-Licameli) [reply src/CMakeLists.txt:1354]: That will be easily fixed by adding one to the list of direct dependencies.  But didn’t my CI run complete?  Are you using the same CMake options as in the CI build?

Even if I can and do fix the PR with one line, why is it a reason not to approve this PR, that the build might break because one line is missing?

Comment (Paul-Licameli) [reply src/CMakeLists.txt:1354]: https://github.com/audacity/audacity/actions/runs/1574761410

That build succeeded yesterday.

Comment (Paul-Licameli) [reply src/CMakeLists.txt]: The CI builds do turn on networking, but I also needed to make the build work if you do not enable it, as must have been so for your local builds but not mine.  There is another commit in the branch.  I will squash and merge if you approve.

To see why I want the change, look at this generated graph with the change:
modules dot

Comment (Paul-Licameli) [reply src/CMakeLists.txt]: Compare without the change.  See a lot of redundant edges that don't change the transitive closure and make it harder to understand.

modules dot

Comment (Paul-Licameli) [reply src/CMakeLists.txt]: Compare without:
modules dot

Comment (crsib) [reply src/CMakeLists.txt]: I build with default CMake options which are different from CI options. It is required that Audacity can be built without networking (and it was your call actually to disable networking by default).

Comment (Paul-Licameli) [reply src/CMakeLists.txt]: There are fewer edges in the first picture, but the transitive closure is the same.  I think the first can be easier to understand.  The difference will be more pronounced as more libraries are extracted.

Comment (Paul-Licameli) [reply src/CMakeLists.txt]: > I build with default CMake options which are different from CI options. It is required that Audacity can be built without networking (and it was your call actually to disable networking by default).

Indeed, but I often make local builds with everything enabled to make sure I don't break things that CI will build.

This was easily fixed with the second commit in this branch.

File: src/CMakeLists.txt [hunk 2]
```diff
@@ -1341,6 +1341,29 @@ if(HAVE_CLOCK_GETTIME)
...
+)
+
+if ( ${_OPT}has_networking )
+   list( APPEND AUDACITY_LIBRARIES
+      lib-network-manager-interface
+
+      # This library depends on lib-network-manager
+      # If Sentry reporting is disabled, an INTERFACE library
+      # will be defined
+      lib-sentry-reporting
```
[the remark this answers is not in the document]
Paul-Licameli: Did you observe a build failure?  I omitted the `-interface` suffix.  That suffix is not necessary for builds, but adding it might sometimes break the build.

I could leave off `-interface` from everything in this list.  In fact I might have no good reason to add it.

I wrote the CMake stuff that generates the extra `-interface` alias targets because of some difficulties getting the modules, not the libraries, to build and link on Windows.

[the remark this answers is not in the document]
Paul-Licameli: > Did you observe a build failure? I omitted the `-interface` suffix. That suffix is not necessary for builds, but adding it might sometimes break the build.
> 
> I could leave off `-interface` from everything in this list. In fact I might have no good reason to add it.
> 
> I wrote the CMake stuff that generates the extra `-interface` alias targets because of some difficulties getting the modules, not the libraries, to build and link on Windows.

In other words are you implying that adding `-interface` may be necessary for the build when lib-sentry-reporting-interface is defined?  I think not.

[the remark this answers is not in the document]
crsib: I do not imply anything about interfaces. I state the fact that lib-sentry-reporting always generates a target. If reporting is disabled - it generates a header-only library. This library is still required to build Audacity.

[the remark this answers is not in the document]
Paul-Licameli: So I don’t understand your objection to that line.  That target IS always present so will not cause a build error.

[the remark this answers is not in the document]
crsib: It is in `if ( ${_OPT}has_networking )` block, which is wrong.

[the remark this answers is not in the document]
Paul-Licameli: You are correct.

Paul-Licameli:
Is it better now?

I try to keep the libraries in the same sequence as in libraries/CMakeLists.txt

crsib (approved):

Issue #2149, opened 2021-11-16
Play-at-Speed has reverted to being non-dynamic


petersampsonaudacity:
**Describe the bug**
The Play-at-Speed has reverted to being non-dynamic.  

@JamesCrook  upgraded this several releases back (2.3.0) so that it was dynamic - i.e. while PAS was playing moving the slider would cause the play sped to change while PAS was playing.

This is now broken in 3.1.0 and 3.1.1 - and is a regression on 3.0.5.

**To Reproduce**
1. Get some audio
2.  click the Play-at-Speed button
3. move the slider during PAS playback
4. Observe the speed of play does not change.

**Expected behavior**
Play speed for PAS to change dynamically.

On this page it the Manual it says it does:
https://manual.audacityteam.org/man/play_at_speed_toolbar.html
where it clearly states:
>Play-at-Speed playback can be adjusted dynamically with the slider with playback is in progress.

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: Windows 10 - but assume all OS (@SteveDaulton tested on Linux and it fails there too)
 - Version: Audacity 3.1.0/3.1.1

**Additional context**
Assigned P1 due to the regression.

Originally reported by a user in this Forum post:
https://forum.audacityteam.org/viewtopic.php?f=46&t=121594&p=439466

petersampsonaudacity:
I assume this is likely to have been caused by the playback changes made for the new looping in 3.1.0

petersampsonaudacity:
Play-at-Speed loop play is also similarly non-dynamic

DavidBailes:
Tested with Audacity 3.1.1 on Windows 10, preferences->Playback->"Vari-speed play" checked.

1. Click the play-at-speed button, then adjusting the speed slider has no immediate effect.
2. Shift+Click the play-at-speed button, then adjusting the slider has an immediate effect.

Using the commands on the Extra->Play-at-speed submenu:
1.  Play-at-speed, then adjusting the speed slider has no immediate effect.
2. Play-at-speed once, then adjusting the speed slider has an immediate effect.

Fixing the bug should include fixing the commands on the Extra->Play-at-speed submenu.

LWinterberg:
Downgrading to P2 - has a workaround and doesn't corrupt projects

Paul-Licameli:
The solution to this may be more difficult than I thought.  Can we reevaluate its priority for 3.1.3?  Can it wait for 3.2?  I think this is a less discoverable problem.

[...]

PR #2288, opened 2021-12-15
Issue2149 Fix play-at-speed button

Paul-Licameli:
Resolves: #2149

Issue2149: Play-at-speed button will work again, when Shift (once-only play) not pressed....

... But now it must also respect loop play bounds and changes of bounds.

This requires total reimplementation of how it works, without reuse of scrubbing code.

The case of Shift + Click needs an additional fix, to come.

Recommended:

### DIFF
--- a/libraries/lib-audio-devices/AudioIOBase.h
+++ b/libraries/lib-audio-devices/AudioIOBase.h
@@ -70,10 +70,12 @@ struct AudioIOStartStreamOptions
    // The return value is a number of milliseconds to sleep before calling again
    std::function< unsigned long() > playbackStreamPrimer;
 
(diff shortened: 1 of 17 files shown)


crsib (approved):
The first part looks good!

Paul-Licameli:
Two more commits.

But -- now I find that the case of Shift + Click on the button is still not right.

So one more to come.

crsib (dismissed):
Still good, but looks like there is a conflict now

Paul-Licameli:
> Still good, but looks like there is a conflict now

There was a conflict because the base branch was master not release.  It wasn't real.

crsib (approved):

Issue #2292, opened 2021-12-16
Dock icon is duplicated on elementaryOS when run as flatpak


stan-janssen:
**Describe the bug**
When audacity is run as flatpak on elementaryOS, the dock icon gets duplicated, because the Dock cannot associate Audacity's main window with the flatpak application we just launched.

This can be easily solved by explicitly specifying the StartupWMClass in the `.desktop` file, for which I have a [pull request](https://github.com/audacity/audacity/pull/2293) available.

**To Reproduce**
1. Install audacity from flathub on elementaryOS
2. Launch audacity
3. Observe that a second icon in the dock appears next to the just-launched audacity icon. This second icon persists while the first disappears after a few seconds.

**Expected behavior**
The Audacity dock icon is  stable and is associated with the running application.

**Screenshots**

![audacity](https://user-images.githubusercontent.com/12174852/146329875-78fe8823-764b-4d6d-b39b-2417811055f5.gif)

**Additional information (please complete the following information):**
 - OS: elementaryOS 6
 - Version 3.1.2 (flathub)

sempasha:
Same problem with [qBittorrent](https://www.flathub.org/apps/details/org.qbittorrent.qBittorrent)

PR #2293, opened 2021-12-16
Solve a problem where the dock icon gets duplicated in the flatpak builds

stan-janssen:
Resolves: [Dock icon is duplicated on elementaryOS when run as flatpak](https://github.com/audacity/audacity/issues/2292)

By explicitly specifying the StartupWMClass, the elementaryOS Dock is now able to associate Audacity's window with the flatpak-launcher icon so that it does not spawn a second icon for Audacity.

I also contributed this [as a patch](https://github.com/flathub/org.audacityteam.Audacity/pull/87) to the flathub maintainer, but since there are no negative effects of explicitly stating the StartupWMClass in the .desktop file, it would be nice to have this in the upstream code as well.

Recommended:

### DIFF
--- a/src/audacity.desktop.in
+++ b/src/audacity.desktop.in
@@ -46,6 +46,7 @@ Comment[zh_CN]=录音和编辑音频文件
 Comment[zh_TW]=錄音和編輯音訊檔案
 
 Icon=@AUDACITY_NAME@
+StartupWMClass=Audacity
 
 Type=Application
 Categories=AudioVideo;Audio;AudioVideoEditing;


LWinterberg:
Hello. While we're not opposed to this change per se, we currently don't have the capacity to test whether this fix has unintended consequences across both many different distros and different packages (our own AppImage, flatpak, apt/dnf packages, etc.). Should we have too much time for testing in a future development cycle, we'll add it, but it might take a while.

LWinterberg (approved):
no issues found in testing

Issue #2282, opened 2021-12-13
Dynamic library path does not change unless using "Browse..." dialog


wildwestrom:
**Is your feature request related to a problem? Please describe.**
There's a link to download a `.pkg` file containing dynamic libraries required for audacity to work. These are outdated, and probably the reason why I can't import various formats supported by ffmpeg (opus for example). 
When I try to point to the path `$HOMEBREW_PREFIX/Cellar/ffmpeg/4.4.1/lib/*.dylib` nothing happens, it uses the same path as before.

**Describe the solution you'd like**
I'd much rather use the libraries provided by the homebrew package. This way I can use the latest stable or nightly release of ffmpeg.

**Describe alternatives you've considered**
I thought to try compiling the libraries myself, but I don't know how to do it in such a way that Audacity will accept the libs. It looks like it's expecting these three:
`ffmpeg.55.64bit.dylib ffmpeg_utils.52.64bit.dylib ffmpeg_codecs.55.64bit.dylib`
Homebrew creates these:
```
main@MacBook ~/h/C/f/4/lib> ls
libavformat.58.76.100.dylib  libpostproc.55.dylib
libavcodec.58.134.100.dylib  libavformat.58.dylib         libpostproc.a
libavcodec.58.dylib          libavformat.a                libpostproc.dylib
libavcodec.a                 libavformat.dylib            libswresample.3.9.100.dylib
libavcodec.dylib             libavresample.4.0.0.dylib    libswresample.3.dylib
libavdevice.58.13.100.dylib  libavresample.4.dylib        libswresample.a
libavdevice.58.dylib         libavresample.a              libswresample.dylib
libavdevice.a                libavresample.dylib          libswscale.5.9.100.dylib
libavdevice.dylib            libavutil.56.70.100.dylib    libswscale.5.dylib
libavfilter.7.110.100.dylib  libavutil.56.dylib           libswscale.a
libavfilter.7.dylib          libavutil.a                  libswscale.dylib
libavfilter.a                libavutil.dylib
libavfilter.dylib            libpostproc.55.9.100.dylib
```
It seems Audacity is hardcoded to only accept certain filenames: https://github.com/audacity/audacity/blob/0b3ba4fd7075ff96c59b04538cefe132056713cd/src/FFmpeg.cpp#L180-L192
[...]

PR #2297, opened 2021-12-16
Fixes the behavior of Locate FFmpeg dialog. (Bug #2282)

crsib:
The meaning of the "Location of " field became confusing after the support for multiple FFmpeg versions was added. 

This commit changes the way how the content of the dialog is interpreted. 
Both full paths to avformat and to the directory containing avformat are now valid.

Additionally, it fixes the way Audacity checks that FFmpeg works. Audacity will check if it can load the library instead of checking that the file exists.

Resolves: #2282

Recommended:

### DIFF
--- a/src/FFmpeg.cpp
+++ b/src/FFmpeg.cpp
@@ -116,7 +116,7 @@ public:
    FindFFmpegDialog(wxWindow *parent, const wxString &path, const wxString &name)
        : wxDialogWrapper(parent, wxID_ANY, XO("Locate FFmpeg"))
        , mName(name)
-       , mFullPath(path, name)
+       , mFullPath(path, {})
    {
       SetName();
 
@@ -183,7 +183,7 @@ public:
 #   if defined(__WXMSW__)
          { XO("Only avformat.dll"), { wxT("avformat-*.dll") } },
 #   elif defined(__WXMAC__)
-         { XO("Only ffmpeg.*.dylib"), { wxT("ffmpeg.*.dylib") } },
+         { XO("Only ffmpeg.*.dylib"), { wxT("ffmpeg.*.dylib"), wxT("libavformat.*.dylib") } },
 #   else
          { XO("Only libavformat.so"), { wxT("libavformat.so.*") } },
 #   endif
@@ -222,7 +222,12 @@ public:
 
    void UpdatePath()
    {
-      mFullPath = mPathText->GetValue();
+      const wxString path = mPathText->GetValue();
+
+      if (wxDirExists(path))
+         mFullPath = wxFileName(path, {}, wxPATH_NATIVE);
+      else
+         mFullPath = mPathText->GetValue();
    }
 
    wxString GetLibPath()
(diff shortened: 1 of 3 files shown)


src/FFmpeg.cpp
```diff
@@ -116,7 +116,7 @@ class FindFFmpegDialog final : public wxDialogWrapper
    FindFFmpegDialog(wxWindow *parent, const wxString &path, const wxString &name)
        : wxDialogWrapper(parent, wxID_ANY, XO("Locate FFmpeg"))
        , mName(name)
-       , mFullPath(path, name)
+       , mFullPath(path, {})
```
Paul-Licameli, line 119: Reviewing the consequences of this change to mFullPath, I see the later call to SelectFile is affected.  Should you pass name there as the fourth argument, to fill in a field by default?  A small convenience for the user?

src/FFmpeg.cpp
```diff
@@ -331,16 +336,35 @@ bool FindFFmpegLibs(wxWindow* parent)
...
+      path = fileName.GetPath();
+
    wxLogMessage(wxT("User-specified path = '%s'"), path);
 
-   if (!::wxFileExists(path)) {
-      wxLogError(wxT("User-specified file does not exist. Failed to find FFmpeg libraries."));
+   wxString prevPath;
+   const bool hasUserPath = AVFormatPath.Read(&prevPath);
+
+   AVFormatPath.Write(path);
```
Paul-Licameli, line 349: To be very careful I might try-catch around what follows to guarantee that side effect on Preferences is undone, attempting a strong guarantee...

Or do it somehow with finally

Paul-Licameli (approved):
Approved with a couple of suggestions.

src/FFmpeg.cpp
crsib: I removed it on purpose. It no longer makes sense because it is a wildcard. On top of that, at least on windows, I observed that having a name messes up if you have more than one filter entry

Paul-Licameli: Ok

src/FFmpeg.cpp
```diff
@@ -302,22 +307,53 @@ BEGIN_EVENT_TABLE(FFmpegNotFoundDialog, wxDialogWrapper)
...
+   }
+
+   void CommitValue() noexcept
+   {
+      mRevertToOld = false;
+   }
+
+   wxString mOldValue;
+   bool mHasOldValue;
+   bool mRevertToOld { false };
```
Paul-Licameli, line 339: I see what you are doing, but mRevertToOld must be constructed as true by default.

Should the destructor catch exceptions?

Paul-Licameli (changes requested):

src/FFmpeg.cpp
Paul-Licameli: Maybe this class wants to be generalized as a utility in Prefs.h.  It would want a template parameter for the Settings object.  But, another day.

crsib: >I see what you are doing, but mRevertToOld must be constructed as true by default.

My bad

>Should the destructor catch exceptions?

I'm not sure why and which? Is there something good we can do?

Paul-Licameli (commented):
Wrong initial value

Paul-Licameli (approved):
That was quick, thanks

src/FFmpeg.cpp
crsib: Maybe this class wants to be generalized as a utility in Prefs.h

We definitely should

Issue #2298, opened 2021-12-16
Fix licenses in packages


AnitaBats:
context: 
* #2142

PR #2300, opened 2021-12-16
Updates LICENSE.txt

LWinterberg:
Resolves: https://github.com/audacity/audacity/issues/2298

License change as per #2142

Further added full copies of all relevant licenses (CC-by, GPLv2, GPLv3)

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,14 +1,8 @@
-Note: 
-Audacity is distributed under the terms of the GNU GPLv2 or later.
-This includes everything in the source code distribution
-except code in the lib-src directory.  Code in the lib-src
-directory may be released under a different license
-(which is GPL-compatible).  For a summary, see the
-README.txt file, and for specific details, see the license
-information inside each subdirectory of lib-src.
-
-Audacity binaries with VST3 support are distributed under the terms
-of the GNU GPLv3. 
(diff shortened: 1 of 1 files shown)


LICENSE.txt
```diff
@@ -1,16 +1,25 @@
...
 directory may be released under a different license
 (which is GPL-compatible).  For a summary, see the
 README.txt file, and for specific details, see the license
 information inside each subdirectory of lib-src.
 
-Documentation is distributed under the Creative Commons license (CC BY 3.0)
-Please see here for that:
-http://creativecommons.org/licenses/by/3.0/legalcode
+Audacity binaries with VST3 support are distributed under the terms
+of the GNU GPLv3.
```
shoogle, line 11: The safest way to word all the above is probably:

> Audacity is released under the GNU General Public License version 3 (GPLv3). Individual source files may be available under other licenses as specified in those files or an accompanying file. In particular, many source files are available under GPL version 2 (GPLv2) or (at your option) any later version, and this is the default license used where no other license is specified.

This replaces everything about binaries, VST3, and the lib-src directory. The new statements below about documentation and full license texts can remain.

shoogle (changes requested):

shoogle (approved):

PR #2310, opened 2021-12-20
Issue2149 again

Paul-Licameli:
Resolves: #2149

Complete the fix for issue 2149:  play-at-speed should also work with a loop region, and you should be able to adust the speed or the region or both during play, without the play head moving incorrectly

Recommended:


crsib (approved):

crsib (changes requested):
Please fix the ubuntu build issue (missing #include <atomic>)! Otherwise I approve

Paul-Licameli:
I added <atomic> and merged this, sorry if I didn't wait for the formal reapproval

PR #2332, opened 2021-12-22
Observers

Paul-Licameli:
Add an implementation of Observer pattern to lib-utility; eliminate all custom wxEvent classes now in libraries in favor of this simpler alternative.

The older pattern was to use wxEvtHandler as the publisher, `Bind()` to subscribe, usually wxEvtHandler as the subscriber too (or else `Unbind()` needed an explicit call), and wxEvent subclasses for messages.  (Event handling is not necessarily done by the event loop, but can be invoked directly on the object publishing the event.)

"Platform neutrality" doesn't strictly require this, becuase wxEvtHandler is in wxBase, but the examples show that it really is more convenient.  Any structure or even any scalar type can be the message sent by a publisher, without the bother of overriding wxEvent::Clone() in a new class, or making the subscriber inherit wxEvtHandler too for automatic cleanup.  (Instead, the class Observer::Subscription acts as a smart pointer for the connection to a publisher; the handle can also expire before it is destroyed, like a std::weak_ptr, in case the Publisher is destroyed first.)

However the class wxCommandEvent is in wxCore not wxBase, so cannot be used as a base class of custom events in libraries.  Therefore some rewriting must be done so that lib-audioio and others can be extracted.  Do this preliminary now, so that those rewrites can also use the nicer interface.

Recommended:

### DIFF
--- a/libraries/lib-audio-devices/DeviceChange.cpp
+++ b/libraries/lib-audio-devices/DeviceChange.cpp
@@ -15,7 +15,7 @@
 
 #include "DeviceChange.h"
 
-
+#include "BasicUI.h"
 
 #if defined(EXPERIMENTAL_DEVICE_CHANGE_HANDLER)
 
@@ -24,9 +24,6 @@
 #include <wx/module.h>
 #include <wx/timer.h>
 #include <wx/thread.h>
-
-DECLARE_LOCAL_EVENT_TYPE(EVT_DEVICE_CHANGE, -1);
-DEFINE_EVENT_TYPE(EVT_DEVICE_CHANGE);
 
 #if defined(__WXMSW__)
 
@@ -121,10 +118,9 @@ public:
       if (mEnabled)
       {
          mEnabled = false;
-         wxMutexGuiEnter();
-         wxCommandEvent e(EVT_DEVICE_CHANGE);
-         mHandler->AddPendingEvent(e);
-         wxMutexGuiLeave();
+         BasicUI::CallAfter([This]{
+            This->mHandler->Publish(DeviceChangeMessage::Change);
+         });
       }
 
       return S_OK;
@@ -135,7 +131,7 @@ public:
       return S_OK;
    }
 
-   bool SetHandler(wxEvtHandler *handler)
+   bool SetHandler(DeviceChangeMessagePublisher *handler)
    {
       mHandler = handler;
 
@@ -160,7 +156,7 @@ public:
    }
 
 private:
-   wxEvtHandler *mHandler;
+   DeviceChangeMessagePublisher *mHandler;
    bool mEnabled;
    ULONG mRefCnt;
    IMMDeviceEnumerator *mEnumerator;
@@ -194,7 +190,7 @@ public:
    }
 
    // IUnknown implementation
-   bool SetHandler(wxEvtHandler *handler)
+   bool SetHandler(DeviceChangeMessagePublisher *handler)
    {
       mHandler = handler;
 
@@ -253,10 +249,9 @@ public:
                if (This->mEnabled)
                {
                   This->mEnabled = false;
-                  wxMutexGuiEnter();
-                  wxCommandEvent e(EVT_DEVICE_CHANGE);
-                  This->mHandler->AddPendingEvent(e);
-                  wxMutexGuiLeave();
+                  BasicUI::CallAfter([This]{
+                     This->mHandler->Publish(DeviceChangeMessage::Change);
+                  });
                }
          
                udev_device_unref(dev);
@@ -270,7 +265,7 @@ public:
    }
 
 private:
-   wxEvtHandler *mHandler;
+   DeviceChangeMessagePublisher *mHandler;
    bool mEnabled;
    pthread_t mThread;
 };
@@ -308,7 +303,7 @@ public:
    }
 
    // IUnknown implementation
-   bool SetHandler(wxEvtHandler *handler)
+   bool SetHandler(DeviceChangeMessagePublisher *handler)
    {
       mHandler = handler;
 
@@ -322,6 +317,7 @@ public:
                                                   &property_address,
                                                   DeviceChangeListener::Listener,
                                                   this) == 0;
+      return true;
    }
 
    void Enable(bool enable)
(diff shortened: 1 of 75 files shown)


libraries/lib-utility/Observer.h
```diff
@@ -0,0 +1,231 @@
...
+class Publisher;
+
+//! Default message type for Publisher
+/*!
+ This has a virtual destructor so that it can be the operand of dynamic_cast,
+ but that is not a requirement for Publisher's template argument.
+ */
+struct UTILITY_API Message {
+public:
+   ~Message();
```
crsib, line 32: The comment above claims the destructor to be virtual

libraries/lib-audio-devices/DeviceChange.cpp
```diff
@@ -348,8 +348,9 @@ class DeviceChangeListener final : public DeviceChangeInterface
          {
             This->mEnabled = false;
             wxMutexGuiEnter();
-            wxCommandEvent e(EVT_DEVICE_CHANGE);
-            This->mHandler->AddPendingEvent(e);
+            BasicUI::CallAfter([This]{
+               This->mHandler->Publish(DeviceChangeMessage::Change);
```
vsverchinsky, line 352: Is there a guarantee that `This` still exists when lambda is invoked?

src/TrackPanel.cpp
```diff
@@ -429,8 +429,7 @@ void TrackPanel::OnTimer(wxTimerEvent& )
 
    // Notify listeners for timer ticks
    {
-      wxCommandEvent e(EVT_TRACK_PANEL_TIMER);
-      p->ProcessEvent(e);
+      ProjectWindow::Get( *GetProject() ).GetPlaybackScroller().Publish({});
```
vsverchinsky, line 426: `window` variable can be used instead

libraries/lib-screen-geometry/ViewInfo.cpp
```diff
@@ -146,23 +134,10 @@ bool NotifyingSelectedRegion::setF1(double f, bool maySwap)
 
 void NotifyingSelectedRegion::Notify( bool delayed )
 {
-   SelectedRegionEvent evt{ EVT_SELECTED_REGION_CHANGE, this };
-   if ( delayed )
-      QueueEvent( evt.Clone() );
+   if (delayed)
+      BasicUI::CallAfter([this]{ Publish({}); });
```
vsverchinsky, line 138: Same question here: is it guaranteed that the `NotifyingSelectedRegion` instance exists when lambda is called?

libraries/lib-audio-devices/DeviceChange.cpp
```diff
@@ -348,8 +348,9 @@ class DeviceChangeListener final : public DeviceChangeInterface
          {
             This->mEnabled = false;
             wxMutexGuiEnter();
```
crsib, line 350: Do we still need this mutex? Looks like we now run the code strictly on the UI thread

src/AdornedRulerPanel.cpp
```diff
@@ -1295,20 +1295,16 @@ AdornedRulerPanel::AdornedRulerPanel(AudacityProject* project,
    wxToolTip::Enable(true);
 #endif
 
-   wxTheApp->Bind(EVT_AUDIOIO_CAPTURE,
-                     &AdornedRulerPanel::OnAudioStartStop,
-                     this);
-   wxTheApp->Bind(EVT_AUDIOIO_PLAYBACK,
-                     &AdornedRulerPanel::OnAudioStartStop,
-                     this);
+   mSubscription1 = AudioIO::Get()
```
crsib, line 1298: Maybe some more descriptive name? Like `mOnAudioStartStopSubscription`

crsib (dismissed):
Approved with some comments!

Paul-Licameli:
> Approved with some comments!

Did you have fun reading that template?  Do you like the simplifications?

Paul-Licameli:
> Maybe some more descriptive name? Like `mOnAudioStartStopSubscription`

If you wish, @crsib, but those Subscription handles aren't mentioned elsewhere after construction.  They are just there for RAII in the containing class's destructor.

libraries/lib-audio-devices/DeviceChange.cpp
Paul-Licameli: > Do we still need this mutex? Looks like we now run the code strictly on the UI thread

Firstly, EXPERIMENTAL_DEVICE_CHANGE_HANDLER is off by default.  I was being complete, trying to maintain compilability in case it is on, but I don't know if we will turn it on.

Secondly, I realize now that I only kept macOS compilable and had to do more.

Thirdly, no, the calls to publish the event happen on different threads, obviously with GTK where pthreads is used to spawn own.

Fourthly though, yes, even on other threads, wxMutexGuiEnter is not necessary to protect calls to wxEvtHandler::CallAfter (what is wrapped inside BasicUI::CallAfter), according to the documentation of CallAfter itself, as safe to call from any thread.

Paul-Licameli: So, I will remove the calls, as you suggested, and also fix the build of other platforms in case this EXPERIMENTAL is on.

libraries/lib-audio-devices/DeviceChange.cpp
Paul-Licameli: > 

Yes.  The DeviceChangeListener (whatever the OS-specific version of it) is held by a unique_ptr DeviceChangeHandler::mListener.  You can easily verify it is not destroyed until the DeviceChangeHandler is destroyed, and DeviceChangeHandler is only ever created as the base class of DeviceManager, and that class has a static single instance.

Anyway this is compiled only when one of the EXPERIMENTAL flags is switched.

src/TrackPanel.cpp
Paul-Licameli: Yes, thank you

libraries/lib-utility/Observer.h
Paul-Licameli: will fix

libraries/lib-screen-geometry/ViewInfo.cpp
Paul-Licameli: I feel less confident.  I'll rewrite with a weak pointer.

Issue #1142, opened 2021-06-25
Extract lib-wave-track


Paul-Licameli:
- [x] #1146
- [ ]  #1143 
- [ ] remaining

PR #2340, opened 2021-12-23
Extract wave track part 1

Paul-Licameli:
Partly resolves: #1142

Some of the dependency breaking moves needed before this next library extraction.

lib-wave-track will not be needed to extract the audio engine, but it will be needed to extract the handlers of Audacity's file format.

Recommended:


src/WaveClip.cpp
```diff
@@ -1670,7 +1669,7 @@ void WaveClip::SetRate(int rate)
 }
 
 /*! @excsafety{Strong} */
-void WaveClip::Resample(int rate, ProgressDialog *progress)
+void WaveClip::Resample(int rate, BasicUI::ProgressDialog *progress)
```
vsverchinsky, line 773: Didn't notice such dependency earlier. Would be nice to remove UI dependencies from WaveClip, but it's not in the scope of this PR anyway

src/WaveTrack.cpp
```diff
@@ -145,7 +159,7 @@ WaveTrack::WaveTrack( const SampleBlockFactoryPtr &pFactory,
    mOldGain[0] = 0.0;
    mOldGain[1] = 0.0;
    mWaveColorIndex = 0;
-   SetDefaultName(TracksPrefs::GetDefaultAudioTrackNamePreference());
+   SetDefaultName(GetDefaultAudioTrackNamePreference());
```
vsverchinsky, line 161: I cannot understand what's this "default name" for, and why do we need to keep a default name copy in each instance of the Track. It brings dependencies that seem unrelated to the WaveTrack or any other *Track.

src/WaveTrack.cpp
```diff
@@ -2851,3 +2850,18 @@ StringSetting AudioTrackNameSetting{
    // Computed default value depends on chosen language
    []{ return DefaultName.Translation(); }
 };
+
+// Bug 825 is essentially that SyncLock requires EditClipsCanMove.
+// SyncLock needs rethinking, but meanwhile this function
+// fixes the issues of Bug 825 by allowing clips to move when in
+// SyncLock.
+bool GetEditClipsCanMove()
```
vsverchinsky, line 2849: Now there is no dependency on `TrackBehaviorsPrefs` but dependency on preferences looks bad to me as well

src/Sequence.cpp
```diff
@@ -962,13 +962,16 @@ void Sequence::WriteXML(XMLWriter &xmlFile) const
          // editing operation that caused this, not fixing
          // the problem but moving the point of detection earlier if we
          // find a reproducible case.
+         using namespace BasicUI;
          auto sMsg =
             XO("Sequence has block file exceeding maximum %s samples per block.\nTruncating to this maximum length.")
                .Format( Internat::ToString(((wxLongLong)mMaxSamples).ToDouble(), 0) );
-         AudacityMessageBox(
+         ShowMessageBox(
```
vsverchinsky, line 969: Out of scope, yet accessing UI from here isn't a good thing

vsverchinsky (approved):

src/WaveClip.cpp
Paul-Licameli: Yes, I thought of just adding a `std::function<void()>` argument for progress notifications, so not even BasicUI is mentioned here.  That might look better.  Still this change is enough to make things platform neutral and keep library dependencies acyclic, and anyway there will be transitive dependency on lib-basic-ui because we use lib-exceptions.

src/WaveTrack.cpp
Paul-Licameli: TrackName::GetDefaultName() is a little weird, and I didn't put it there.  I think researching the code history may trace it back to accessibility fixes: see where GetDefaultName() is used, to tell whether the name is defaulted.  But the default name for audio tracks is not hard-coded, but changeable by the user in preferences.

I'm not fixing all the existing weirdness but just breaking dependency cycles.

src/WaveTrack.cpp
Paul-Licameli: I disagree.

I think it is not bad for the global Settings objects (`BoolSetting`, `IntSetting`, etc.) to exist in whatever lowest level libraries need to use them.  See lib-project-rate/QualitySettings.cpp for the precedent.

The user interface for changing those settings, using Preference dialog pages, can be supplied at a higher level of the dependency graph that is not independent of UI.  That is what the file `TracksBehaviorsPrefs` does.

The two preference paths mentioned in GetEditClipsCanMove() really should be re-done as Settings objects to eliminate the duplication of the string literal constant for that path in scattered places.

You can see that is what I did by introducing `AudioTrackNameSetting` in another commit, and it is used in only a few places.

But I didn't do it for "/GUI/SyncLockTracks" and "/GUI/EditClipCanMove" because of the large number of places that would have to change, making this PR overhwelming with too many other changes not related to the build time dependency breaking needed to make the library.

If I do make another PR to introduce those global Settings variables, will you review that?

src/Sequence.cpp
Paul-Licameli: I agree, but as my commit comment said, this is a piece of defensive programming against a very unlikely file corruption.

It's harmless now.

PR #2343, opened 2021-12-24
Modernization

Paul-Licameli:
Use more C++14 and C++17 library types

Recommended:


src/TrackPanel.h
```diff
@@ -53,9 +53,7 @@ struct TrackPanelDrawingContext;
 
 enum class UndoPush : unsigned char;
 
-enum {
-   kTimerInterval = 50, // milliseconds
-};
+static constexpr auto  kTimerInterval = std::chrono::milliseconds{50};
```
vsverchinsky, line 56: Feels like it won't harm to put duration_cast to ensure that values passed to wx* or other functions that accept not typed units work correctly even in case if someone decides to change units in the definition, now it's not stated that I'm obliged to use same units.

vsverchinsky (approved):
Except compilation issues caused by missing `#include` everything else seems to be ok

Paul-Licameli:
I will fix the build, and may use duration_cast as suggested, and will re-request review.

Paul-Licameli:
> I will fix the build, and may use duration_cast as suggested, and will re-request review.

As I promised.

Please review the last three, new commits.  Nothing has changed in the earlier commits except for rebasing and supplying needed include directives.

PR #2368, opened 2022-01-01
Simplify the macro definitions in AttachedVirtualFunction...

Paul-Licameli:
... see commit 48217d5a which introduced them.  Partly they simplify some
difficult syntax, but partly they worked around linkage problems on Windows.

Now I know a better way to avoid the problems:  see this reference
https://docs.microsoft.com/en-us/cpp/cpp/defining-inline-cpp-functions-with-dllexport-and-dllimport?view=msvc-170

It is not contradictory for a function to be dllexport and also inline.

So use dllexport unconditionally for the inline base class constructor and for
GetRegistry(), no matter whether compiling the target declaring the base, or
another target attaching an override.

The article implies that the linker will still coalesce the duplicated
definitions of static local variables in the function bodies, which is important
for correctness.

Recommended:

### DIFF
--- a/libraries/lib-project/Project.cpp
+++ b/libraries/lib-project/Project.cpp
@@ -109,6 +109,3 @@ void AudacityProject::SetInitialImportPath(const FilePath &path)
       mInitialImportPath = path;
    }
 }
-
-// Generate the needed, linkable registry functions
-DEFINE_XML_METHOD_REGISTRY( ProjectFileIORegistry );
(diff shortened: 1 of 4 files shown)


Paul-Licameli:
С Новым годом, Дима!

Paul-Licameli:
Experience with #2483 teaches that the approach in this PR is wrong, but it can also still be corrected with the CRTP idiom.  I'm making this a draft rather than closing it.

Issue #681, opened 2020-10-04
Too many placeholders in strings making it harder to translate


scootergrisen:
I don't like this way the following string was changed from a single string:

    Audacity is a free program written by a worldwide team of [[https://www.audacityteam.org/about/credits|volunteers]]. Audacity is [[https://www.audacityteam.org/download|available]] for Windows, Mac, and GNU/Linux (and other Unix-like systems).

To multiple strings using %s placeholders:

    %s is a free program written by a worldwide team of %s. %s is %s for Windows, Mac, and GNU/Linux (and other Unix-like systems).
    volunteers
    available

It makes it harder to read the string for translators and having to read comments about which strings go where which was unnessasary before.

Maybe something like this is better:

    %s is a free program written by a worldwide team of [[%s|volunteers]]. %s is [[%s|available]] for Windows, Mac, and GNU/Linux (and other Unix-like systems).

There is still the disadvantage that using %s for all the placeholders seems to give the translator no way to decide which order they go in which may or may not be wanted in some languages.

https://github.com/audacity/audacity/blob/master/src/AboutDialog.cpp#L332-L333

Same goes for the string:

    If you find a bug or have a suggestion for us, please write, in English, to our [[https://forum.audacityteam.org/|forum]]. For help, view the tips and tricks on our [[https://wiki.audacityteam.org/|wiki]] or visit our [[https://forum.audacityteam.org/|forum]].

That was changed to:

    If you find a bug or have a suggestion for us, please write, in English, to our %s. For help, view the tips and tricks on our %s or visit our %s.
    forum
    wiki
    forum

https://github.com/audacity/audacity/blob/master/src/AboutDialog.cpp#L350-L351

petersampsonaudacity:
This needs changing anyway as Audacity is no longer "written by a worldwide team of volunteers", so this is no longer true.

Audacity is now written by a professional team of developers and others employed by Muse, with occasional contributions from external volunteers.

PR #2373, opened 2022-01-04
remove unrendered strings

LWinterberg:
Resolves: https://github.com/audacity/audacity/issues/681

basically, I'm just getting some substitutions back into the string and also updating to the current state of affairs

Recommended:

### DIFF
--- a/src/AboutDialog.cpp
+++ b/src/AboutDialog.cpp
@@ -338,35 +338,6 @@ void AboutDialog::PopulateAudacityPage( ShuttleGui & S )
 {
    CreateCreditsList();
 
-   auto par1Str =
-// DA: Says that it is a customised version.
-#ifdef EXPERIMENTAL_DA
-      wxT(
-"Audacity, which this is a customised version of, is a free program written by a worldwide team of [[https://www.audacityteam.org/about/credits|volunteers]]. \
-Audacity is [[https://www.audacityteam.org/download|available]] for Windows, Mac, and GNU/Linux (and other Unix-like systems).")
-#else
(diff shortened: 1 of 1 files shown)


src/AboutDialog.cpp
```diff
@@ -348,39 +348,21 @@ Audacity is [[https://www.audacityteam.org/download|available]] for Windows, Mac
 /* Do the i18n of a string with markup carefully with hints.
  (Remember languages with cases.) */
       XO(
-/* i18n-hint: First and third %s will be the program's name,
-  second %s will be "volunteers", fourth "available" */
-"%s is a free program written by a worldwide team of %s. \
-%s is %s for Windows, Mac, and GNU/Linux (and other Unix-like systems).")
+/* i18n-hint: %s substitutes to the program name. Do not change the links (ie the left side of the [[ | ]] ) */
+"%s is free, open source software, developed by [[https://mu.se|Muse Group]] with support of the [[https://www.audacityteam.org/about/credits|Free/Open Source community]].  \
+%s is [[https://www.audacityteam.org/download|available]] for Windows, MacOS and GNU/Linux.")
```
Paul-Licameli, line 353: Do not remove "excessive" substitutions.

I wrote those complications in for a very good reason:  we don't want confusing markup to be in the translation catalog, and prone to errors made by translators.

Therefore the complication of substituting a translated word into a marked-up link, and then substitution of that into a bigger sentence.

Paul-Licameli (changes requested):
"Excessive" substitutions are there for a reason.  Don't undo them.

Paul-Licameli (approved):
Less code!

Issue #2359, opened 2021-12-30
MIDI playback starts from beginning regardless of cursor position


petersampsonaudacity:
**Describe the bug**
MIDI playback starts from beginning of the MIDI track regardless of cursor position and then plays only for the duration of the time from cursor position to end of track,

Happens both with normal Play and with Timeline Quickplay.

This is a regression on 3.0.5

**To Reproduce**
1. Import a MIDI track
2. position the cursor partway though the MIDI track
3. Press Play
4. Observe: Audacity plays from the beginning of the MIDI track ignoring the cursor position - _but the visuals are correct_
5. Click in the Timeline partly through the time-span of the MIDI track
6. Observe: Audacity Quickplays from the beginning of the MIDI track ignoring the cursor position - _but the visuals are correct_

**Expected behavior**
Playback should start from where the cursor is either in the track for Play or in the Timeline for Quickplay

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: W10 and macOS 12.0.1 Monterey - so assume all OS
 - Version: Audacity 3.1.3-3.1.0

**Additional context**
Scrubbing is not affected as you cannot scrub MIDI tracks in Audacity.

_Note that we have had no user complaints about this since the 3.1.x series was released, which indicates that MIDI playback is not commonly used,_

LWinterberg:
I wonder if @Paul-Licameli's big fix for #2081 resolves this.

Pokechu22:
I'm going to note that I ran into this a few days ago and was planning on reporting it when I had more time - it's also worth noting that playback works correctly when you have a selection; it just fails when you have an offset cursor position but no selection.

petersampsonaudacity:
Report today from a user on the Forum
https://forum.audacityteam.org/viewtopic.php?f=46&t=122606&p=442838

Penikov:
Confirm in 3.2

petersampsonaudacity:
See also:   Audacity hangs on Stop when playing MIDI that has notes before T=0 #2539

petersampsonaudacity:
Still the case on W10 with 3-2-0 Beta

But note that this only happens if you have a point selection.

If you have a range selection the the range is properly played.

petersampsonaudacity:
See also:  Adjustment of loop play bounds is ignored for MIDI playback #2081

LWinterberg:
[...]

Issue #2081, opened 2021-11-03
Adjustment of loop play bounds is ignored for MIDI playback


Paul-Licameli:
**Describe the bug**
Adjustment of loop play bounds is ignored for MIDI playback

**To Reproduce**
1. Import some MIDI
2. Make an audio track -- such as, generating a rhythm track as a metrononome that is in time with the MIDI.
3. Play without looping and observe the synchronization
4. Play with looping and adjust the loop play bounds
5. Observe that MIDI will quickly get out of time with the rhythm track and not play the parts 

**Expected behavior**
MIDI play remains in sync with looping play of the audio

petersampsonaudacity:
See also:   MIDI playback starts from beginning regardless of cursor position #2359

PR #2378, opened 2022-01-04
Midi looping, speed play, cut preview

Paul-Licameli:
Resolves: #2359
Resolves: #2081

Fix the cooperation of MIDI with the adjustable looping region, and also make it work with the speed slider and cut preview.

These kinds of play should now remain synchronized with any audio tracks that also play simultaneously.

Change of speed means MIDI events like note-on and -off are more rapid, but pitch will be the same.

Still to do:  make MIDI work with scrubbing too.

Recommended:


pietro68 (approved):
Quite honestly this code is beyond me and I would need quite some time to understand what is going on here.
I did some QA in the sense that I tried looping and speed changes on a midi file and it seems to work. 

One thing I noticed (out of the scope of this PR) is that although I selected my laptop speakers as playback device, a wave track will be played through them like it should, but the midi file gets played through my wireless headphones. This happens also in the current released 3.1.3

LWinterberg:
This PR has been sitting in approved for quite some time now, can it be merged?

Paul-Licameli:
> This PR has been sitting in approved for quite some time now, can it be merged?

Would likely need a difficult rebasing and re-review for 3.3

Which I would like to see happen

PR #2383, opened 2022-01-05
de-flag DarkAudacity 

LWinterberg:
The idea of this PR is to simplify the DarkAudacity parts found in Audacity today so they're either in or not (=just in the darkaudacity branch), rather than being carried around behind a build flag. 

Namely: 
- [x] Remove DA-specific branding and other bits which don't really affect anything (mostly the part 1 commit)
- [x] remove DA-code which may affect some things (part 2)
- [x] Make a decision on which bits of DA's new defaults and changes stay and which ones get removed (I've already chosen some, but I'll need to talk to the team on what we want for the rest) >> #2385
- [x] Updates some dates on the way
- [x] ~~include Harry Potter spoilers in the branch name~~

Recommended:


LWinterberg:
Okay, I think I'm done here. The thing I'm most insecure about is PlayableTrackControls.cpp; I'm not sure if I got everything which was only referenced by DarkAudacity out of there (or if I removed too much).

LWinterberg:
yeah, I'm not comfortable attempting to resolve that particular conflict.

Paul-Licameli:
Why close this?  I think it is a good cleanup though it has not been a high priority for review.

PR #2422, opened 2022-01-13
Global hooks

Paul-Licameli:
A class template utility to capture a pattern that has been repeated several times, and will be again, soon for extraction of
AudioIO and later to make the improved ShuttleGui system toolkit-neutral.

It generates global variables that each hold a std::function, are safely accessible from anywhere at static initialization time,
and support scopes that swap functions in and out of the variable.

This can be seen as an easy generator of small "facades" that each have just one entry point.  It's not a technique to
overuse, but often valuable when a dependency must be broken.

All present uses of the technique are rewritten with this class, and that makes it easy to find where all the hooks are with a
scan of the source code text.

There is a more general base class to generate global variables of other types besides std::function specializations.  This
too can be useful to simplify a much repeated pattern, eliminating the need to write a function each time with a local static
variable.  No direct uses yet of this base class.

Recommended:


libraries/lib-utility/GlobalVariable.h
```diff
@@ -0,0 +1,167 @@
...
+
+private:
+   //! Use static functions only.  Don't directly construct this.
+   GlobalVariable() = delete;
+
+   //! Generate the static variable
+#ifdef _WIN32
+   __declspec(dllexport)
+#endif
+   static mutable_type &Instance()
```
vsverchinsky, line 119: The name `Instance` looks a bit misleading to me, considering that return type isn't a `GlobalVariable<>`

src/commands/ScreenshotCommand.cpp
```diff
@@ -107,10 +107,7 @@ ScreenshotCommand::ScreenshotCommand()
    mbBringToTop=true;
    mIgnore=NULL;
    
-   static std::once_flag flag;
-   std::call_once( flag, []{
-      ::SetVetoDialogHook( MayCapture );
-   });
+   static VetoDialogHook::Scope scope{ MayCapture };
```
vsverchinsky, line 110: Since `Scope` keeps the previous value it may hold some unnecessary data longer than one may expect, also the previous value will be restored when the static object is destroyed.

libraries/lib-utility/GlobalVariable.h
```diff
@@ -0,0 +1,167 @@
...
+   template<typename... Arguments>
+   static result_type Call(Arguments &&...arguments)
+   {
+      auto &fn = GlobalHook::Get();
+      if (fn)
+         return fn(std::forward<Arguments>(arguments)...);
+      else if constexpr (std::is_void_v<result_type>)
+         return;
+      else
+         return result_type{};
```
vsverchinsky, line 163: There is no way to know whether the function was actually invoked, in many cases `result_type{}` could be a valid value

vsverchinsky: Maybe it makes sense to change the interface so that it becomes mandatory to initialize the hook with some meaningful value?

vsverchinsky:
In some cases the hooks are never changed, is it necessary to keep them then?

libraries/lib-utility/GlobalVariable.h
Paul-Licameli: But this is a private member function.  Users of the class use it only indirectly by calling `Get()` or using `Initializer`.

src/commands/ScreenshotCommand.cpp
Paul-Licameli: I defined `Scope` to work generally with `GlobalVariable` specializations other than for functions.

In this case, the default-constructed empty `std::function` (which I expect is implemented as only a null pointer) is all that gets moved into the `scope` object, so it's only "wasting" one statically allocated pointer.

I allow `GlobalVariable` specialization to support non-scoped reassignments, but not by default.  I want the RAII idiom to apply be default.

I'm not sure if you are questioning whether the destructor of `scope` might come after the destruction of the local static variable in `Instance`, using that variable after destruction?  But I asked myself that question, and found a reassuring answer here:

https://en.cppreference.com/w/cpp/utility/program/exit

"if a function-local (block-scope) static object was initialized during construction of a subobject of a class or array, it is only destroyed after all subobjects of that class or all elements of that array were destroyed."

(And "subobject" is defined here https://en.cppreference.com/w/cpp/language/object to include member objects.)

That describes `instance` in GlobalVariable.h, which will be constructed (if not already constructed) when `scope` initializes its member `m_previous`.  Therefore `instance` will be destroyed, safely, only after `scope`.

Notice that I also provide `Initializer` for other cases where this guarantee does not apply, but this case doesn't need it for safety.

libraries/lib-utility/GlobalVariable.h
Paul-Licameli: This is all as I intend, to simplify the usage.  The hook can be called with `Call` without the complication of using `Get` and a null check first, provided that the function type returns void or a default-constructible type.  I think this convenience is good in most cases.

If in a particular case it is important to know whether any function was installed, the return type might be a pointer or a `std::optional`, while it is expected that an explicitly installed function will return non-null.  Then the caller can test the return value.

If there is a particular case in which it it is not good, then the template argument of `GlobalHook` can be a function returning a non-default-constructible type, and `Call` will fail to compile, compelling the more verbose usage instead.

Paul-Licameli:
Have I answered your doubts sufficiently?

I have force-pushed the branch again onto more recent master, but with no changes.

vsverchinsky (approved):

Issue #2098, opened 2021-11-07
AudioIO::FillPlayBuffers needs resequencing


Paul-Licameli:
AudioIoCallback::DrainInputBuffers (in low latency thread) may call NewDefaultPlaybackPolicy::Done which calls NewDefaultPlaybackPolicy::RevertToOldDefault which uses schedule.mTimeQueue.GetLastTime().

PlaybackSchedule::TimeQueue::SetLastTime() is called by the TrackBufferExchange thread in NewDefaultPlaybackPolicy::RepositionPlayback, in AudioIO::FillPlayBuffers.

Comments for class PlaybackSchedule::TimeQueue explain that it does not have its own atomic variables for synchronization but depends on those in RingBuffer.

Those atomics are updated with release ordering in RingBuffer::Put which also is called by AudioIO::FillPlayBuffers().

The RepositionPlayback call FOLLOWS the call to RingBuffer::Put().

Therefore, this is incorrect synchronization!

A long issue comment here, to justify what will be a very small commit.  But please, follow my logic, dear reviewer.

_Originally posted by @Paul-Licameli in https://github.com/audacity/audacity/issues/2098#issuecomment-962531551_

Paul-Licameli:
AudioIoCallback::DrainInputBuffers (in low latency thread) may call NewDefaultPlaybackPolicy::Done which calls NewDefaultPlaybackPolicy::RevertToOldDefault which uses schedule.mTimeQueue.GetLastTime().

PlaybackSchedule::TimeQueue::SetLastTime() is called by the TrackBufferExchange thread in NewDefaultPlaybackPolicy::RepositionPlayback, in AudioIO::FillPlayBuffers.

Comments for class PlaybackSchedule::TimeQueue explain that it does not have its own atomic variables for synchronization but depends on those in RingBuffer.

Those atomics are updated with release ordering in RingBuffer::Put which also is called by AudioIO::FillPlayBuffers().

The RepositionPlayback call FOLLOWS the call to RingBuffer::Put().

Therefore, this is incorrect synchronization!

A long issue comment here, to justify what will be a very small commit.  But please, follow my logic, dear reviewer.

Issue #2082, opened 2021-11-03
Real-time effect calculation on producer not consumer side of queue


Paul-Licameli:
There is some existing support for real-time effects.

It applies calculations to buffers of samples in the low-latency portAudio thread, and may call foreign code in VST plug-ins.  It is unknown if such code might allocate free store.

It may call into built-in Audacity effects too, where we can see all the code.  Still it constrains that code to avoid free store.

Better to figure out a restructuring that applies effect calculations on the procuder side of the RingBuffer, in the TrackExchangeBuffer thread, which must keep up sufficient throughput but is not subject to the constraints of real-time and low latency.

Paul-Licameli:
This issue and #2098 are co-involved and may need to be fixed together

PR #2428, opened 2022-01-14
Producer side effect calculation

Paul-Licameli:
Resolves: #2082 
Resolves: #2098

Manage the RingBuffers for playback differently.  Write track data into them, then transform in-place as needed for real-time
effects, then "flush" to make all visible to the consumer thread.  Do not do effect calculations on the consumer side, where
there are low-latency constraints and as little work as possible should be done in each pass.

Recommended:


src/effects/EffectUI.cpp
```diff
@@ -1158,10 +1138,16 @@ void EffectUIHost::LoadUserPresets()
 
 void EffectUIHost::InitializeRealtime()
 {
-   if (mSupportsRealtime && !mInitialized)
-   {
-      RealtimeEffectManager::Get(*mProject).RealtimeAddEffect(mEffect);
-      
+   if (mSupportsRealtime && !mInitialized) {
+      mpState =
```
pietro68, line 1142: this part is affected by the same problem we have seen in PR 2500 i.e. mpState will be returned null - this because the code inside AudioIO::AddState was probably cherry-picked from Realtime-effect-manager-rewrite-part3. 
I can not test how effects work until that part will be fixed - so please merge future fixes from Realtime-effect-manager-rewrite-part3 into this and I can try again.

Paul-Licameli:
@pietro68 if you fetch the updated version of this branch, the problem you mention should no recur.

This branch is based on "Realtime-effect-manager-rewrite-part-3".   That other branch has been updated with a fix, and now this branch is rebased onto the fixed version.

pietro68 (approved):
My ususal simple test passed, and I also tested that no additional latency happens when monitoring or recording.

PR #2441, opened 2022-01-17
Restrictions on the view of wxWidgets through the wxBase target...

Paul-Licameli:
... First, add some utilities to find and store argc and argv.

Then, rewrite main() and WinMain() to use them.  This removes the dependency
on wx/app.h from PluginHost.cpp.  That is needed before what comes next.

Finally, wxBase is no longer an alias target.  It is an interface target to wxWidgets
that adds some preprocessor definitions that restrict at compile time what
parts of wxWidgets may be used, to general wxBase utilities only.

No graphical interface is allowed, but also none of the main event loop.  So in
fact only a subset of wxBase is good to use, because even that includes an event
loop usable in console programs.  Restrict the use of certain headers by
pre-including their include guards, so that an attempt to include them skips
the definitions they contain and makes compilation fail.

The library targets that depend on wxBase must make it PRIVATE, for otherwise
the restrictions on use of wxWidgets would be propagated to users of those
libraries that do not need to be toolkit-neutral and break their compilation.

Restoring what was defined before at 1f71ef4 but was soon lost at f66381b with
the conan-izing.

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/AudacityFunctions.cmake
+++ b/cmake-proxies/cmake-modules/AudacityFunctions.cmake
@@ -732,3 +732,29 @@ function( addlib dir name symbol required check )
    endif()
 endfunction()
 
+# Copy named properties from one target to another
+function(copy_target_properties
+  src  # target
+  dest # target
+  # prop1 prop2...
+)
+   foreach(property ${ARGN})
+      get_target_property(value ${src} ${property})
+      if(value)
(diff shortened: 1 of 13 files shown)


crsib:
I think we should rebase it on latest master and see, if Linux packages build succeeds

Paul-Licameli:
Oh, this forgotten PR.  Okay, rebased now.  But this isn't urgent.

crsib (approved):

Issue #2084, opened 2021-11-03
Rewrite EffectManager and related legacy code


Paul-Licameli:
The legacy code that manages the set of known effects, their presets, and realtime effect state is confusing and not well documented as to the purpose of the various abstract class interfaces.

Rewrite it better for maintenance.

There is work in progress at this branch, which I last worked on in July.  (It includes some restructuring of Nyquist too, but that is not necessary for release 3.2.)  I need to revisit it, refresh my memory of the problems, and turn into a review-ready PR.

https://github.com/Paul-Licameli/audacity/tree/effect-manager-rewrite

PR #2444, opened 2022-01-18
Realtime effect manager rewrite part 1

Paul-Licameli:
Preliminaries for issues #2082, #2084

Removal of two EXPERIMENTAL flags (one turned on, the other abandoned), a source code file split, a new data structure
RealtimeEffectList associated with the project and with each track (but doing nothing yet), some RAII and modernization.

Scoping objects for the processing of realtime and for suspension of it.

Recommended:


src/effects/lv2/LV2Effect.cpp
```diff
@@ -1408,8 +1409,9 @@ size_t LV2Effect::RealtimeProcess(int group, float **inbuf, float **outbuf, size
    return numSamples;
 }
 
-bool LV2Effect::RealtimeProcessEnd()
+bool LV2Effect::RealtimeProcessEnd() noexcept
 {
+return GuardedCall<bool>([&]{
```
vsverchinsky, line 1414: `R GuardedCall(const F1 &body, const F2 &handler = F2::Default(), std::function<void(AudacityException*)> delayedHandler = DefaultDelayedHandlerAction{})`
Don't you think that the last parameter makes this call not completely exception-safe?

src/effects/RealtimeEffectManager.cpp
```diff
@@ -225,23 +215,20 @@ void RealtimeEffectManager::RealtimeProcessStart()
...
 
 //
 // This will be called in a different thread than the main GUI thread.
 //
 size_t RealtimeEffectManager::RealtimeProcess(int group, unsigned chans, float **buffers, size_t numSamples)
 {
-   // Protect ourselves from the main thread
-   mRealtimeLock.Enter();
+   // Protect...
+   std::lock_guard<std::mutex> guard(mLock);
```
vsverchinsky, line 237: Shouldn't we avoid such things in the realtime processing thread?

src/effects/lv2/LV2Effect.cpp
Paul-Licameli: Do you mean that delayedHander should be noexcept?  But as the name suggests, its execution is delayed, and always in the main thread, invoked from the event loop, where exceptions escaping from it would also be caught by the top-level exception handler.  It is not invoked in the catch block.

The second argument is the non-delayed handler and the documentation is explicit that this function may re-throw its argument, or throw another exception.

src/effects/RealtimeEffectManager.cpp
Paul-Licameli: That would be good.  Reexamination of why that lock is there, and maybe elimination of it, should be attempted.  But two answers:

It's not the purpose of the pull request to fix this.  It's not worse than before, in which there was also locking.

Also the realtime effect processing will be moved (Issue #2082) from the "real-time" thread to the other worker thread that sends messages across the RingBuffer.  That thread needs to maintain the throughput but has more tolerance for variance in time to process batches.  Locking a mutex or allocating memory are things less urgent to remove from that thread.

src/effects/lv2/LV2Effect.cpp
vsverchinsky: Not exactly, `LV2Effect::RealtimeProcessEnd()` is now `noexcept` but I think that `std::function` constructor does not provide such guarantees

Paul-Licameli: I see what you mean.

From https://en.cppreference.com/w/cpp/utility/functional/function/function:  "5,10) Does not throw if f is a function pointer or a std::reference_wrapper, otherwise may throw std::bad_alloc or any exception thrown by the copy constructor of the stored callable object."

So, please see the added commit, where I make some effort to avoid potentially throwing construction of std::function.

I also add a lot of `noexcept` in AudacityException.h.  The proof that exceptions really don't happen is still not complete, though, without diving into `AudacityException::EnqueueAction` and from there into `BasicUI::CallAfter` to be sure no other kinds of exceptions escape besides `std::bad_alloc`.

Still there is an opinion of some that you should implement `https://en.cppreference.com/w/cpp/memory/new/set_new_handler` if you have a way to do it, and then ignore `std::bad_alloc`, because memory exhaustion is an unrecoverable situation which is going to lead to a `std::terminate` anyway.

src/effects/EffectUI.h
```diff
@@ -14,101 +14,13 @@
 #ifndef __AUDACITY_EFFECTUI_H__
 #define __AUDACITY_EFFECTUI_H__
 
-#include <wx/bitmap.h> // member variables
```
vsverchinsky, line 17: I believe it's still required by `EffectUIHost`, and causes compilation issues on Windows platform

vsverchinsky (approved):
Approved, but compilation issues need to be fixed before the branch is merged

Issue #2372, opened 2022-01-03
`(return "something")` done twice in the Nyquist prompt hangs Audacity


werame:
**Describe the bug**
Nyquist Prompt `(return "something)` hangs Audacity when done twice

**To Reproduce**
1. Open a Nyquist Prompt.
2. Type
```
(return "something")
```
3. And click "OK".
4. Repeat steps 2, 3, 4. 
5. Audacity hangs.

**Expected behavior**
As at step 3:  the dialog is just quietly dismissed.

**Screenshots**
![image](https://user-images.githubusercontent.com/97036286/147941010-a6bdfd48-8a33-4ab0-9959-aeaacf4679ac.png)

**Additional information (please complete the following information):**
 - OS: Windows 10
 - Version Audacity 3.1.3

**Alternative reproducer (clicking Debug instead of OK hangs on empty track as well)**
Audacity doesn't hang when running the exact same snippet of Nyquist code given above if either there's no track selected or even if the mono track is empty tough. I.e. if at step 1 above one does just (the clicking equivalent equivalent of) NewMonoTrack, then no hang occurs, but nothing is returned to the user via a dialog box either.

However, "scenario 2", one can make it hang even with a empty track if instead of "OK" one clicks Debug in the Nyquist Prompt dialog, i.e. the following sequence also hangs Audacity:

1. Have an empty mono track selected or no track selected, in which case Audacity will auto-generate an empty mono one when one enters the Nyquist Prompt.
2. Open a Nyquist Prompt.
3. Type
```
(return "something")
```
4. And click "Debug" (instead of OK, as in the previous scenario).
5. Audacity hangs.

**Additional context**
Also, it doesn't matter if it's the top-level function that's being returned, the following also hangs in the circumstances described

```
(defun foobar (x)
    (return "something")) ; also hangs

(foobar 10)
```

On the other hand, a proper use of `return-from` instead of `return` works as one may expect

```
(defun foobar (x)
    (return-from foobar "something")) ; ok, no hang

(foobar 10) ;  "something" returned in dialog box
```

[...]

PR #2459, opened 2022-01-20
Xlisp fixes

Paul-Licameli:
Resolves: #2372

It was possible to hang Nyquist with certain inputs to the Nyquist prompt.

This was fun... I had to dive into the XLisp interpreter code.  This changes lib-src, and should be considered by @rbdannenberg for upstreaming to Nyquist.

Recommended:

### DIFF
--- a/lib-src/libnyquist/nyquist/xlisp/xlread.c
+++ b/lib-src/libnyquist/nyquist/xlisp/xlread.c
@@ -185,13 +185,14 @@ int xlload(const char *fname, int vflag, int pflag)
 
     /* read, evaluate and possibly print each expression in the file */
     xlbegin(&cntxt,CF_ERROR,s_true);
-    if (_setjmp(cntxt.c_jmpbuf))
+    if (_setjmp(cntxt.c_jmpbuf)) {
         sts = FALSE;
         #ifdef DEBUG_INPUT
             if (read_by_xlisp) {
 		fprintf(read_by_xlisp, ";;;;xlload: catch longjump, back to %s\n", fullname);
             }
         #endif
+    }
     else {
         #ifdef DEBUG_INPUT
             if (read_by_xlisp) {
@@ -277,14 +278,14 @@ int xlread(LVAL fptr, LVAL *pval, int rflag)
     int sts;
 
     /* read an expression */
-    while ((sts = readone(fptr,pval)) == FALSE)
+    while ((sts = readone(fptr,pval)) == FALSE) {
 #ifdef DEBUG_INPUT
     if (debug_input_fp) {
         int c = getc(debug_input_fp);
         ungetc(c, debug_input_fp);
     }
 #endif
-        ;
+    }
 
     /* return status */
     return (sts == EOF ? FALSE : TRUE);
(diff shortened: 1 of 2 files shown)


rbdannenberg:
Very interesting -- this may take a few days, since it's surprising this could have been lurking in Xlisp for decades (or maybe there's a branch like Xlispstat where it was detected and fixed alreaady). But I'll definitely look into it. Thanks.

Paul-Licameli:
> Very interesting -- this may take a few days, since it's surprising this could have been lurking in Xlisp for decades (or maybe there's a branch like Xlispstat where it was detected and fixed alreaady). But I'll definitely look into it. Thanks.

It is not XLisp proper but Nyquist’s additions to it that had the bug.  See the commit comments.

rbdannenberg (changes requested):
Very interesting! The xlisp.c code does come from original xlisp, although it was split to allow re-entering the interpreter, and I believe it's correct without the added xlend. The reason is that you can only get to top_level by executing xlrestore, which loads the entire heap from a file and sets xlcontext to NULL. This is all disabled in Nyquist because Nyquist allows for "externals" and there's no implementation for archiving and restoring added datatypes, so this code is unreachable. If you call TOP-LEVEL, you unwind to a context with a CF_TOPLEVEL flag, which takes you to some other place.

The rest looks right. I would consider setting xlcontext to NULL when reentering the interpreter, although with xlend() in nyx.c, it shouldn't be necessary.

Paul-Licameli:
> Very interesting! The xlisp.c code does come from original xlisp, although it was split to allow re-entering the interpreter, and I believe it's correct without the added xlend. The reason is that you can only get to top_level by executing xlrestore, which loads the entire heap from a file and sets xlcontext to NULL. This is all disabled in Nyquist because Nyquist allows for "externals" and there's no implementation for archiving and restoring added datatypes, so this code is unreachable. If you call TOP-LEVEL, you unwind to a context with a CF_TOPLEVEL flag, which takes you to some other place.
> 
> The rest looks right. I would consider setting xlcontext to NULL when reentering the interpreter, although with xlend() in nyx.c, it shouldn't be necessary.

I do not see a function called `xlrestore` in Audacity's version of the xlisp library, only a function called `sound_xlrestore`.

Are you sure this unusual bug isn't present even in the original?  That evaluating `(return)` once is ok (`xlfail` will be invoked) but the second time might hang the interpreter?  But it is unnecessary for Audacity so I can omit that fix.

Paul-Licameli:
I have amended the pull request.  Please approve it @rbdannenberg if there are no further objections.

rbdannenberg (approved):

rbdannenberg:
Sorry, it's xrestore() rather than xlrestore() (in xldmem.c).  It looks like when I added external types, I planned for xrestore() to work, but then decided sound types were too messy to save and restore, so sound_xlrestore() is where the restore implementation would go, but it should never get called. The XLisp RESTORE function is not defined unless SAVERESTORE is defined. 

I tried various returns in command line Nyquist just to make sure, and it seems safe:
```
> (return)
error: no target for RETURN
1> (return)
error: no target for RETURN
2> (top-level)
[ back to top level ]
> (return)
error: no target for RETURN
1> (setf *breakenable* nil)
NIL
1> (top-level)
[ back to top level ]
> (return)
error: no target for RETURN
> (return)
error: no target for RETURN
> 
```

Issue #2437, opened 2022-01-16
Better support for stereo plugins


werame:
**Is your feature request related to a problem? Please describe.**

Right now if you write a stereo generator e.g. in Nyquist you're faced with the following usability issue: if you have no track selected a mono will be inserted, and your stereo plugin will fail. 

**Describe the solution you'd like**

The easiest thing would be for a plug-in to declare how many channels it wants to generate, like say, at leas support some syntax like

```
;type generate stereo
```

**Additional context**

Although I first thought this would be an easy hack to add to Nyquist.cpp, and I even wrote the parser part, it turns out not to be the easy at all because the decision what/whether to insert is done in the massive, monolithic method of the Effect base namely DoEffect.

And there's no single object in Audacity that represents a stereo track, WaveTrack is mono. Inserting a stereo track via the menu does this little song and dance:

```
   auto left = tracks.Add( trackFactory.NewWaveTrack( defaultFormat, rate ) );
   left->SetSelected(true);

   auto right = tracks.Add( trackFactory.NewWaveTrack( defaultFormat, rate ) );
   right->SetSelected(true);

   tracks.MakeMultiChannelTrack(*left, 2, true);

   ProjectHistory::Get( project )
      .PushState(XO("Created new stereo audio track"), XO("New Track"));

   TrackFocus::Get(project).Set(left);
   left->EnsureVisible();
```

But DoEffect assumes newTrack is one object, not a pair, including in cleanup. So it will require fairly substantial rewrite to handle "atomic" insertion and cleanup of more than one track.

So I'll leave this to people who designed this class affair to sort it out how to implement this relatively trivial feature.

werame:
Or scratch that simple idea of predeclaring return type as too inflexible as well. Come to think of it, Audacity is capable of inserting the right resource in one case: when a process plugin returns a label track. So, instead of preinserting any track at all for generate, it should simply do it after the plugin returns a Lisp object, which isn't a data track yet, just a generator for one, aka Nyquist SOUND object--that's how Nyquist works. 

So the logic for type generate should probably rather be:

1. run the Nyquist code, see what it returns.
2. If the number of channels in the SOUND object matches the selected track, just insert in there.
3. If not, pop up a dialog telling the user about the mismatching and asking if they'd rather have the result inserted in a new track.

[...]

PR #2460, opened 2022-01-21
Better support for Nyquist stereo plugins (generators and mono->stereo effects)

werame:
Resolves: #2437

## Use case A: stereo generator on empty project

On a empty project run a generator plugin that returns stereo. As there aren't any default stereo ones that come with Audacity, write e.g. at the Nyquist prompt.

```(vector (hzosc 100) (hzosc 900))```

Observe that you get the error "Nyquist returned too many audio channels" in dialog box.

This PR fixes this error, allowing stereo output on the auto-generated mono track by first creating "overflow" tracks and then deleting the empty, auto-generated track. All 3 commits included are basically needed for this use case.

Why I chose this solution is that it also enable a few related use cases.

## Use case B: stereo generator attempting to insert on mono track

On an empty project first create some mono track. Then, after pacing the insertion cursor in the that track, try the same command as in use case A. Normally you'd get an error. With this PR you'll get dialog box that asks if you want to keep the result in a new track.

image

If you choose "Cancel" you get nothing (i.e. the old behavior). If you choose OK, it works as in use case A.

## Use case C: mono->stereo effect, not needing pre-duplication (anymore)

On an empty project first create some mono track that has some (preferably visible) waveform, e.g. use the built-in Tone gen. Then select some region of that track type the following at the Nyquist prompt 

```(vector (scale 0.3 *track*) (scale 0.7 *track*))```

Normally, again, you'd get an error. With this PR instead it will work as in use case B: you'll be prompted if you want the result returned in a new track. (A workaround without this PR is to to duplicate the track first and apply some mono effects separately to these copies, assuming the effects don't have intricate inter-channel data dependency.) 

Use cases B and C only really need the first two patches in PR. But as the joke goes: allocating memory is easy, deleting it without crashes is a bit harder. As "stress test" derived from case A:

### Case D (really A followed by B)

Do use case A, then delete the result from the project and do use case B. Audacity shouldn't crash. Why is this worth mentioning? Because the NyquestEffect object is a singleton, so its state needs to be properly reset between plugin runs.

Recommended:

### CODE REVIEW

**Comment by werame** [comment]
So this bit is visible here as to avoid possibly duplicated efforts/comments on the same matter in different venues, I'm copying this conversation I had with @SteveDaulton the forum:

>> Just wondering why you limited it to two channels. (A use case for supporting more than 2 channels could be splitting a track into multiple frequency bands).
>
> Indeed, but there are a lot of hardcoded two-channel arrays in Nyquist.h (just search for "[2]" [in the file](https://github.com/audacity/audacity/blob/master/src/effects/nyquist/Nyquist.h#L258)). And I'm not responsible for those. (Also nyx.c isn't a lot better in that regard.) It will take a fair bit of work to make it support more.
>
>> Also, when returning 2 channels when one channel is selected, the returned audio is at the start of the track rather than aligned to the selection. Is that intentional (why?) or do you intend to change that?
>
>I can add that as a feature later, even during this PR maybe. It did occur to me already one may want that alignment, but it wasn't immediately obvious to me how to code it so it happens. I'm not using ClearAndPaste like the usual ("overwriting") fx code path does because I use the "hidden" tracks (outputTracks) that were already being created in Nyquist.cpp, they were just not being added to the TrackList. And those outputTracks are zero aligned. I'm not sure how to time shift them. I might need to create new ones and ClearAndPaste into the latter, since ClearAndPaste does support offsets, but I'm hoping there's a simpler method.

The (lack of) alignment is not an issue in use case A (as far as I can tell), but it is in the other ones.

**Comment by werame** [comment]
@SteveDaulton I fixed the track alignment issue (see commit right above). Thankfully, it was easy to do without making any further track copies.

**Comment by werame** [comment]
I've rebased this PR past the ~100 commits made by Paul in the past two weeks. 🤖 It still works for me.

**Comment by SteveDaulton** [comment]
It still works here too (and nice to see the audio returned within the temporal selection).

I still think that if the limitation of "returning the same number of channels as selected" is removed, it's a bit disappointing that it is only being extended to a maximum of 2 channels. I can think of several use cases for allowing an arbitrary number of channels (such as splitting audio into Bass, Mid, Treble).

**Comment by werame** [comment]
@SteveDaulton 

Expecting the first PR from a new contributor to solve all the technical debt like

```
(master)$ git grep -P '(?<!tokens)\[2\]' -- src/effects/nyquist/
src/effects/nyquist/Nyquist.cpp:1569:   std::shared_ptr<WaveTrack> outputTrack[2];
src/effects/nyquist/Nyquist.h:258:   WaveTrack         *mCurTrack[2];
src/effects/nyquist/Nyquist.h:259:   sampleCount       mCurStart[2];
src/effects/nyquist/Nyquist.h:273:   Buffer            mCurBuffer[2];
src/effects/nyquist/Nyquist.h:274:   sampleCount       mCurBufferStart[2];
src/effects/nyquist/Nyquist.h:275:   size_t            mCurBufferLen[2];
src/effects/nyquist/Nyquist.h:277:   WaveTrack        *mOutputTrack[2];
```

and all the use sites of those vars, is a bit much, IMHO. Especially since changing those to be dynamically allocated is not trivial in the context of a singleton object where a single re-initialization oversight [can lead to crashes](https://github.com/audacity/audacity/commit/5e390e2a08c8a5fe597d5e8e6f2e151648fc6ccf) in "did X then did Y scenarios".

Please be aware that even solving this little wart of not being able to add a stereo gen to an empty project (or while not having a track selected) took about 80 lines of code (albeit including comments), mostly additions, due to other technical debts in Nyquist.cpp and Effect.cpp. And insofar nobody from the core devs has stepped up yet to even do a code review of this PR's changes.

Please do add a separate GH issue for Nyquist-in-Audacity to support >2 multi-channel though.

SteveDaulton:
> Expecting the first PR from a new contributor to solve all the technical debt

I'm not "expecting" werame to solve everything. I mentioned the temporal alignment because I thought it would make the new feature more complete - it turned out not to be too much trouble, and imo it's a big improvement over the initial pull request.

Similarly, extending to an arbitrary number of tracks would, imo, be another big improvement and would make this feature feel complete. I don't know how hard that will be, but we already handle adding multiple tracks at the same time when recording multiple tracks, so perhaps the same can be done here (?)

I agree that a code review from one of the core devs is overdue.

LWinterberg:

Issue #2458, opened 2022-01-20
Persistency of settings for real-time effect stack(s)


Paul-Licameli:
There exists now a global realtime effect "stack," though it never gets more than one effect at a time.

There will be per-track effect stacks in 3.2, but a global effect stack will also be useful.

There will be need to save the user's choices of effects and their settings to the Audacity project and re-load them.

This issue tracks the implementaiton of persistency, which will be testable for the global effects even while the interface for per track effects is still waiting.

PR #2466, opened 2022-01-22
Realtime effect manager rewrite part 2

Paul-Licameli:
Resolves: #2458 

Merge parts of the work of @lllucius that reorganize the realtime effects into separate per-project and per-track stacks, and
include the persistency of effect settings in the .aup3 project file.

The user interface to access realtime effects is temporarily removed in the intermediate commits, then restored in the last
one, but this restored version is only provisional pending further release 3.2 developments.  It still applies only one effect
per-project, while that effect's non-modal dialog is open.

As mentioned in other code and commit comments, still unresolved problems are
* allowing independent settings for multiple occurrences of an effect in stacks
* undo and redo of changes to effect stacks

Recommended:

### DIFF
--- a/src/effects/RealtimeEffectManager.cpp
+++ b/src/effects/RealtimeEffectManager.cpp
@@ -167,18 +167,6 @@ void RealtimeEffectManager::RealtimeSuspend()
       state->RealtimeSuspend();
 }
 
-void RealtimeEffectManager::RealtimeSuspendOne( EffectProcessor &effect )
-{
-   auto begin = mStates.begin(), end = mStates.end();
-   auto found = std::find_if( begin, end,
-      [&effect]( const decltype( mStates )::value_type &state ){
-         return state && &state->GetEffect() == &effect;
-      }
-   );
-   if ( found != end )
-      (*found)->RealtimeSuspend();
-}
-
 void RealtimeEffectManager::RealtimeResume() noexcept
 {
    // Protect...
@@ -194,18 +182,6 @@ void RealtimeEffectManager::RealtimeResume() noexcept
 
    // And we should too
    mSuspended = false;
-}
-
-void RealtimeEffectManager::RealtimeResumeOne( EffectProcessor &effect )
-{
-   auto begin = mStates.begin(), end = mStates.end();
-   auto found = std::find_if( begin, end,
-      [&effect]( const decltype( mStates )::value_type &state ){
-         return state && &state->GetEffect() == &effect;
-      }
-   );
-   if ( found != end )
-      (*found)->RealtimeResume();
 }
 
 //
--- a/src/effects/RealtimeEffectManager.cpp
+++ b/src/effects/RealtimeEffectManager.cpp
@@ -56,51 +56,6 @@ bool RealtimeEffectManager::RealtimeIsSuspended() const noexcept
    return mSuspended;
 }
 
-void RealtimeEffectManager::RealtimeAddEffect(EffectProcessor &effect)
-{
-   // Block RealtimeProcess()
-   SuspensionScope scope{ &mProject };
-
-   // Add to list of active effects
-   mStates.emplace_back( std::make_unique< RealtimeEffectState >( effect ) );
-   auto &state = mStates.back();
-
-   // Initialize effect if realtime is already active
-   if (mActive)
-   {
-      // Initialize realtime processing
-      effect.RealtimeInitialize();
-
-      // Add the required processors
-      for (size_t i = 0, cnt = mRealtimeChans.size(); i < cnt; i++)
-      {
-         state->RealtimeAddProcessor(i, mRealtimeChans[i], mRealtimeRates[i]);
-      }
-   }
-}
-
-void RealtimeEffectManager::RealtimeRemoveEffect(EffectProcessor &effect)
-{
-   // Block RealtimeProcess()
-   SuspensionScope scope{ &mProject };
-
-   if (mActive)
-   {
-      // Cleanup realtime processing
-      effect.RealtimeFinalize();
-   }
-      
-   // Remove from list of active effects
-   auto end = mStates.end();
-   auto found = std::find_if( mStates.begin(), end,
-      [&](const decltype(mStates)::value_type &state){
-         return &state->GetEffect() == &effect;
-      }
-   );
-   if (found != end)
-      mStates.erase(found);
-}
-
 void RealtimeEffectManager::RealtimeInitialize(double rate)
 {
    // The audio thread should not be running yet, but protect anyway
--- a/src/effects/RealtimeEffectManager.cpp
+++ b/src/effects/RealtimeEffectManager.cpp
@@ -10,6 +10,7 @@
 
 
 #include "RealtimeEffectManager.h"
+#include "RealtimeEffectList.h"
 #include "RealtimeEffectState.h"
 
 #include "EffectInterface.h"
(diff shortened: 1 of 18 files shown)


Paul-Licameli:
I force pushed an updated branch, only to rebase onto recent master and fix the Windows build with another `#include`.

src/effects/RealtimeEffectManager.cpp
```diff
@@ -274,6 +276,55 @@ void RealtimeEffectManager::VisitAll(StateVisitor func)
...
+RealtimeEffectManager::AddState(Track *pTrack, const PluginID & id)
+{
+   auto pLeader = pTrack ? *TrackList::Channels(pTrack).begin() : nullptr;
+   RealtimeEffectList &states = pLeader
+      ? RealtimeEffectList::Get(*pLeader)
+      : RealtimeEffectList::Get(mProject);
+
+   SuspensionScope scope{ &mProject };
+   // Protect...
+   std::lock_guard<std::mutex> guard(mLock);
```
pietro68, line 289: mLock here will be locked/unlocked three times:

1) locked and then released as a result of the call to ::Suspend() in the suspension scope constuctor
2) locked by `guard`, then released by its destructor when return is reached
3) locked and then released as a result of a call to Resume() in the supension scope destructor

There is potentially unprotected time between a release and a subsequent lock. I wonder if all of these operations should better be under only one lock. One solution could be to have non-locking versions of Suspend() and Release()

src/effects/RealtimeEffectManager.cpp
```diff
@@ -56,51 +56,6 @@ bool RealtimeEffectManager::RealtimeIsSuspended() const noexcept
    return mSuspended;
```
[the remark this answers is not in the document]
Paul-Licameli: Better, let's remove the function because it is used nowhere.

The function as written would not be reliable, because the caller would have no guarantee that the condition it reports remains the same, unless holding the mutax lock.  But then the function itself would deadlock, unless `std::recursive_mutex` were used instead.

It reminds me of: https://en.cppreference.com/w/cpp/memory/weak_ptr/expired

"This function is inherently racy if the managed object is shared among threads. In particular, a false result may become stale before it can be used. A true result is reliable."

Yet in our case we can't even say that one of the results remains reliable.

src/effects/RealtimeEffectManager.cpp
Paul-Licameli: Very good.  I noticed the same too.

This is in a function that runs in the main thread.  The functions that can run in another thread are `ProcessStart`, `Process`, `ProcessEnd`.  These are the calls that really need mutual exclusion from the main thread functions, so that the contents of the `RealtimeEffectList`s are stable for that thread.

Each of those functions checks first whether processing is suspended, and they do nothing if so.

I don't think there is any incorrectness in the avoidance of races, but there is probably needless inefficiency, and a lock-free alternative should be sought.

That is not yet in the intended scope of any of these pull requests.

pietro68 (dismissed):

src/effects/RealtimeEffectState.cpp
```diff
@@ -144,8 +138,6 @@ bool RealtimeEffectState::AddTrack(int group, unsigned chans, float rate)
 
       // Add a NEW processor
       mEffect->RealtimeAddProcessor(gchans, rate);
-
-      // Bump to next processor
       mCurrentProcessor++;
```
pietro68, line 122: I am not sure that this scheme of things will work when having multiple effects (multiple RealtimeEffectState-s) referencing the same EffectProcessor.
Let's imagine a mono track having two instances of BassAndTreble on it. `mCurrentProcessor` is basically an index referencing the "slaves" created in BassAndTreble::RealtimeAddProcessor. I understand it like this: during this method, `mCurrentProcessor` will take a value of 0 and possibly also 1, if the track is stereo - because in that case, two slaves are needed, one per channel. The only instance of BassAndTreble will create two slaves, and this instance of RealtimeEffectState will reference them with indexes 0 and 1 - all good so far.

But what will happen when the second instance of RealtimeEffectState will reach this method? it will make the only instance of BassAndTreble create two more slaves, but it will reference them again with indexes 0 and 1, instead of 2 and 3.

This potential erratic behavior can not be proven now because AFAIK the GUI still does not allow us to place more effects on the same track; moreover, this way of managing per-instance states is going to be rewritten anyway, so I do not know if it is useful to correct this problem. 

If we need to, my proposed solution is simply that `::RealtimeAddProcessor` would return the index of the created slave, and we use that instead of the increased `mCurrentProcessor`

pietro68: Anyway - adding effects per-track is not supported by the GUI so this bug has no chance to manifest itself - plus, we will do a rewrite of how states are stored, probably before the supporting GUI code will be written; therefore, I will mark this conversation as resolved.

src/effects/RealtimeEffectManager.cpp
```diff
@@ -65,25 +66,31 @@ void RealtimeEffectManager::Initialize(double rate)
...
-void RealtimeEffectManager::AddTrack(int group, unsigned chans, float rate)
+void RealtimeEffectManager::AddTrack(Track *track, unsigned chans, float rate)
 {
-   VisitGroup(nullptr, [&](RealtimeEffectState &state, bool){
-      state.AddTrack(group, chans, rate);
-   });
-
-   mChans.push_back(chans);
-   mRates.push_back(rate);
+   auto leader = *track->GetOwner()->FindLeader(track);
```
pietro68, line 85: Here is the problem which I outlined during our conversation:

If I have a project with two tracks, the Visitor function will make the RealtimeEffectState be initialized a second time, thereby forgetting the indexes in the EffectProcessor slaves; this means that when play is pressed, both tracks will be processed by the same slave.

Paul-Licameli: My recent force push should fix this:  please confirm.

pietro68: I built PR 2500 and tested it with my usual test (two mono tracks and BassTreble used in realtime).
There seems to be a defect there which prevents me to test the fix (I will place a comment there to describe it)
Anyway, I have noticed that the code changed so much, that my initial comment is not applicable anymore, so I will press resolve and being this the last issue, I will approve this PR.

Paul-Licameli:
Merge parts of the work of @lllucius that reorganize the realtime effects into separate per-project and per-track stacks, and
include the persistency of effect settings in the .aup3 project file.

The user interface to access realtime effects is temporarily removed in the intermediate commits, then restored in the last
one, but this restored version is only provisional pending further release 3.2 developments. It still applies only one effect
per-project, while that effect's non-modal dialog is open.

As mentioned in other code and commit comments, still unresolved problems are

allowing independent settings for multiple occurrences of an effect in stacks
undo and redo of changes to effect stacks

Paul-Licameli:
Merge parts of the work of @lllucius that reorganize the realtime effects into separate per-project and per-track stacks, and
include the persistency of effect settings in the .aup3 project file.

The user interface to access realtime effects is temporarily removed in the intermediate commits, then restored in the last
one, but this restored version is only provisional pending further release 3.2 developments. It still applies only one effect
per-project, while that effect's non-modal dialog is open.

As mentioned in other code and commit comments, still unresolved problems are

allowing independent settings for multiple occurrences of an effect in stacks
undo and redo of changes to effect stacks

Issue #2472, opened 2022-01-25
AUP project fails to open in 3.x if selection is before zero


SteveDaulton:
**Describe the bug**
Opening an old AUP project fails if the project was save with a selection before zero.

**To Reproduce**
1. Create a project in Audacity 2.x with a selection extending before zero.
2. Save the project.
3. Attempt to open the project in Audacity 3.x
4. See error: 
```
Couldn't import the project:

Could not load file: "name-of-file.aup"
```

**Expected behavior**
Selections before zero are valid, so the project should open in Audacity 3.x

**Additional information (please complete the following information):**
 - OS: Any
 - Version: Audacity 3.x

**Additional context**
Bug reported on the forum.

PR #2473, opened 2022-01-25
Fix bug 2472

SteveDaulton:
Selection before zero is legal in both Audacity 2.x and 3.x

Resolves: https://github.com/audacity/audacity/issues/2472

*Don't error on negative selection times*

Recommended:


Paul-Licameli (approved):
Simple enough

Issue #2474, opened 2022-01-25
New track control panel button: effects


AnitaBats:
@Higherself1133 please add details to this task.

Higherself1133:
<img width="322" alt="Screen Shot 2022-01-26 at 10 20 44 AM" src="https://user-images.githubusercontent.com/97682925/151136488-4ea7ee02-7756-475a-b701-2e32bc47e13e.png">
<img width="443" alt="Screen Shot 2022-01-26 at 10 20 59 AM" src="https://user-images.githubusercontent.com/97682925/151136495-5a5fb6c7-623b-4cfd-8105-f36da502e70a.png">

Figma Link: https://www.figma.com/file/odPgVM0lWc6MdsJWxpgPJl/Audacity?node-id=5822%3A58269

DavidBailes:
Please could these figma diagrams also always include a specification for how keyboard users can execute a task - not just a specification for mouse users.

PR #2480, opened 2022-01-26
Track effects button

Paul-Licameli:
Resolves: #2474

Add per-track Effects buttons, not yet functional

This depends on PR #2466

Recommended:

### DIFF
--- a/src/tracks/playabletrack/ui/PlayableTrackControls.cpp
+++ b/src/tracks/playabletrack/ui/PlayableTrackControls.cpp
@@ -18,6 +18,7 @@ Paul Licameli split from TrackInfo.cpp
 #include "../../../TrackInfo.h"
 #include "../../../TrackPanelDrawingContext.h"
 #include "ViewInfo.h"
+#include "../../../effects/RealtimeEffectManager.h"
 
 #include <wx/dc.h>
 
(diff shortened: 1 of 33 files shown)


src/tracks/playabletrack/ui/PlayableTrackControls.cpp
```diff
@@ -158,6 +187,24 @@ void MuteAndSoloDrawFunction
...
+
+   wxRect bev = rect;
+
+   GetEffectsHorizontalBounds( rect, bev );
+   {
+      auto target = dynamic_cast<EffectsButtonHandle*>( context.target.get() );
+      bool hit = target && target->GetTrack().get() == pTrack;
+      bool captured = hit && target->IsClicked();
+      bool down = captured && bev.Contains( context.lastState.GetPosition());
+      EffectsDrawFunction( dc, bev, pTrack, down, captured, hit );
```
vsverchinsky, line 205: There are two things that isn't implemented in this PR: 
>Once an effect is applied, the ‘Add Effects’ button changes to a differently coloured ‘Effects’ button. This will help the user distinguish between tracks that have effects applied and tracks that don’t.

So text should change from 'Add Effect' to 'Effects'. and button color as well

Paul-Licameli:
Closing this, now that @vsverchinsky has incorporated it into a larger prototype.

Issue #2483, opened 2022-01-26
Master build crashes at startup


Penikov:
**Describe the bug**

One of the recent PRs brokes down master branch so Audacity crashes at startup

**To Reproduce**

1. Download the latest master build
2. Run Audacity

**Additional information (please complete the following information):**
 - OS: Windows 10

`>	
        lib-transactions.dll!TransactionScope::Commit() Line 58	C++
 	Audacity.exe!UndoManager::RemoveStates(unsigned __int64 begin, unsigned __int64 end) Line 239	C++
 	Audacity.exe!UndoManager::ClearStates() Line 252	C++
 	Audacity.exe!ProjectHistory::InitialState() Line 52	C++
 	Audacity.exe!ProjectManager::New() Line 389	C++
 	Audacity.exe!AudacityApp::InitPart2() Line 1548	C++
 	Audacity.exe!AudacityApp::OnInit() Line 1384	C++
 	wxbase313ud_vc_x64_custom.dll!wxAppConsoleBase::CallOnInit() Line 93	C++
 	wxbase313ud_vc_x64_custom.dll!wxEntryReal(int & argc, wchar_t * * argv) Line 491	C++
 	wxbase313ud_vc_x64_custom.dll!wxEntry(int & argc, wchar_t * * argv) Line 184	C++
 	wxmsw313ud_core_vc_x64_custom.dll!wxEntry(HINSTANCE__ * hInstance, HINSTANCE__ * __formal, char * __formal, int nCmdShow) Line 306	C++
 	Audacity.exe!WinMain(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, char * __formal, int nCmdShow) Line 799	C++`

werame:
Probably one of the last 15 commits or so. It worked a couple of days ago.

crsib:
I've checked the builds from your branch and observed no issue

PR #2486, opened 2022-01-26
Fix GlobalHook implementation for Windows ...

Paul-Licameli:
... Now it works just like subclasses of ClientData::Site :

Use GlobalHook by deriving a struct from it, and give the correct visibility
to the struct, which will guarantee generation of unique non-inline functions in
one library (or the executable), with the needed static local variables.

The first template parameter is thus now another "CRTP" parameter.

Resolves: #2483

Recommended:

### DIFF
--- a/libraries/lib-sample-track/Mix.h
+++ b/libraries/lib-sample-track/Mix.h
@@ -63,8 +63,9 @@ class SAMPLE_TRACK_API Mixer {
     {
     public:
        //! Hook function for default time warp
-       using DefaultWarp =
-          GlobalHook<WarpOptions, const BoundedEnvelope*(const TrackList&)>;
+       struct SAMPLE_TRACK_API DefaultWarp : GlobalHook<DefaultWarp,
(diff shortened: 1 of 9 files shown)


Penikov:
I've tried to run a WinX64 build (artifact: audacity-win-3.2.0-alpha-20220126+4121821-x64-msvc2019) 
Unfortunately, this build has crashed on my PC at startup.

Penikov:
I've tried to run a WinX64 build (artifact: audacity-win-3.2.0-alpha-20220126+4121821-x64-msvc2019)
Unfortunately, this build has crashed on my PC at startup.

crsib (approved):

Paul-Licameli:
> I've tried to run a WinX64 build (artifact: audacity-win-3.2.0-alpha-20220126+4121821-x64-msvc2019)
> Unfortunately, this build has crashed on my PC at startup.

Is the call stack known to be the same?

Issue #2210, opened 2021-12-01
Error configuring when using system wxWidgets and no libjpeg-turbo-devel package


imciner2:
**Describe the bug**

The CMake configure step is always searching for the installed libjpeg-turbo library development files, even when set to use the system wxWidgets (which is already compiled).

**To Reproduce**
Run CMake with the following options on an install that doesn't have libjepg-turbo.
```
    -DAUDACITY_BUILD_LEVEL:STRING=2 \
    -Daudacity_conan_enabled=Off \
    -Daudacity_has_networking=Off \
    -Daudacity_has_crashreports=Off \
    -Daudacity_has_updates_check=Off \
    -Daudacity_has_sentry_reporting=Off \
    -Daudacity_lib_preference=system \
    -Daudacity_obey_system_dependencies=On \
    -Daudacity_use_wxwidgets=system \
    -Daudacity_use_uuid=system \
    -Daudacity_use_sqlite=system \
    -Daudacity_use_libsndfile=system \
    -Daudacity_use_soxr=system \
    -Daudacity_use_lame=system \
    -Daudacity_use_twolame=system \
    -Daudacity_use_libflac=system \
    -Daudacity_use_ladspa=on \
    -Daudacity_use_libvorbis=system \
    -Daudacity_use_libid3tag=system \
    -Daudacity_use_expat=system \
    -Daudacity_use_soundtouch=system \
    -Daudacity_use_vamp=system \
    -Daudacity_use_lv2=system \
    -Daudacity_use_portaudio=system \
    -Daudacity_use_midi=system \
    -Daudacity_use_libogg=system \
    -Daudacity_use_ffmpeg=off \
    -Daudacity_use_portsmf=local \
    -Daudacity_use_sbsms=local \
```
See the following error:
```
-- Checking for module 'zlib >= 1.2.11'
--   Found zlib , version 1.2.11
-- Using 'ZLIB' system library
-- Interface ZLIB::ZLIB:
	include: 
	Libraries: 
-- Checking for module 'expat >= 2.1.0'
--   Found expat , version 2.4.1
-- Using 'expat' system library
-- Interface expat::expat:
	include: 
	Libraries: 
CMake Error at cmake-proxies/cmake-modules/AudacityDependencies.cmake:308 (message):
  Error: JPEG is required.

  Please install it with using command like:

  	$ sudo apt install libjpeg-turbo8-dev
Call Stack (most recent call first):
[...]

PR #2492, opened 2022-01-28
Only check libpng and libjpeg existence when using local wxWidgets

crsib:
Resolves: #2210 

Recommended:


cmake-proxies/CMakeLists.txt
```diff
@@ -81,6 +79,19 @@ add_conan_lib(
       libjpeg-turbo:SIMD=False
 )
 
+# If we are building using Conan provided version of
+# wxWidgets, but against system libraries - ensure that required
+# Librries are installed.
```
vsverchinsky, line 84: typo: *Libraries

PR #2500, opened 2022-01-31
Realtime effect manager rewrite part 3

Paul-Licameli:
Depends on #2466

Partly Resolves: #2084

Further rewrites of the use of RealtimeEffectManager so that there are RAII scope objects wherever begin and end methods
of the manager must be called in complementary pairs.

A change for safe shutdown of processing that was first identified by Leland Lucius.  noexcept specifiers on steps that must
happen inside destructors.

Recommended:

### DIFF
--- a/src/AudioIO.cpp
+++ b/src/AudioIO.cpp
@@ -143,6 +143,27 @@ AudioIO *AudioIO::Get()
 }
 
 struct AudioIoCallback::TransportState {
+   TransportState(std::weak_ptr<AudacityProject> wOwningProject,
+      const WaveTrackArray &playbackTracks,
+      unsigned numPlaybackChannels, double rate)
+   {
+      if (auto pOwningProject = wOwningProject.lock();
+          pOwningProject && numPlaybackChannels > 0) {
+         // Setup for realtime playback at the rate of the realtime
+         // stream, not the rate of the track.
+         mpRealtimeInitialization.emplace(move(wOwningProject), rate);
+         // The following adds a new effect processor for each logical track.
+         for (size_t i = 0, cnt = playbackTracks.size(); i < cnt;) {
+            auto vt = playbackTracks[i].get();
+            unsigned chanCnt = TrackList::Channels(vt).size();
+            i += chanCnt; // Visit leaders only
+            mpRealtimeInitialization->AddTrack(
+               vt, std::min(numPlaybackChannels, chanCnt), rate);
+         }
+      }
+   }
+
+   std::optional<RealtimeEffects::InitializationScope> mpRealtimeInitialization;
 };
 
 // static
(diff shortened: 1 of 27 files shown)


src/AudioIO.cpp
```diff
@@ -380,6 +404,27 @@ AudioIO::~AudioIO()
    mThread.reset();
 }
 
+RealtimeEffectState *AudioIO::AddState(Track *pTrack, const PluginID & id)
+{
+   if (mpState && mpState->pRealtimeScope)
```
pietro68, line 409: My usual test: two mono tracks and an EffBassTreble, then hit play, place breakpoint here.
When it breaks here, mpState is reported as empty, so no addition of EffectStates will be made. This will result in the tracks not being processed by the effect.

src/AudioIO.h
```diff
@@ -345,6 +350,10 @@ class AUDACITY_DLL_API AudioIoCallback /* not final */
    RecordingSchedule mRecordingSchedule{};
    PlaybackSchedule mPlaybackSchedule;
 
+   struct TransportState;
+   //! Holds some state for duration of playback or recording
+   std::unique_ptr<TransportState> mpState;
```
pietro68, line 355: Could we rename this to mpTransportState? when I first encountered this member in the code, I got a bit confused because now "state" default-maps in my mind to RealtimeEffectState

src/effects/EffectUI.cpp
```diff
@@ -1158,10 +1138,16 @@ void EffectUIHost::LoadUserPresets()
 
 void EffectUIHost::InitializeRealtime()
 {
-   if (mSupportsRealtime && !mInitialized)
-   {
-      RealtimeEffectManager::Get(*mProject).RealtimeAddEffect(mEffect);
-      
+   if (mSupportsRealtime && !mInitialized) {
+      mpState =
```
pietro68, line 1142: As I commented earlier, at the moment mpState will be returned as nullptr in the moment I instantiate a BassTreble. 
There is no check here for (mpState == nullptr), I think it should be added

src/AudioIO.cpp
pietro68: I think I found the reason for this bug: mpState is not initialized until AudioIO::StartStream is reached, but that happens in response to pressing the play button, while the code above using mpState is triggered when instantiating an effect.

src/AudioIO.cpp
```diff
@@ -142,6 +142,30 @@ AudioIO *AudioIO::Get()
...
+            auto vt = playbackTracks[i].get();
+            unsigned chanCnt = TrackList::Channels(vt).size();
+            i += chanCnt; // Visit leaders only
+            pRealtimeScope->AddTrack(
+               vt, std::min(numPlaybackChannels, chanCnt), rate);
+         }
+      }
+   }
+
+   std::optional<RealtimeEffects::InitializationScope> pRealtimeScope;
```
pietro68, line 166: could we rename this to pInitializationScope? (mpInitializationScope, actually) just to reduce the confusion in my head :-) 

anyway I see the point of this now, which is to make sure that any SuspensionScope or ProcessingScope will be bracketed by an InitializationScope, and have this guarantee at compile time.

src/AudioIO.cpp
Paul-Licameli: AddState needs to suspend conditionally, only in case playback is happening.  That was my mistake.

src/AudioIO.h
Paul-Licameli: ok

src/AudioIO.cpp
Paul-Licameli: ok

pietro68 (dismissed):
Not really sure if I should "Request Changes" - I guess the defect I pointed out in the first conversation is being taken care of (or it will be shortly), and its solution will come up in a new PR - so I will select "Approve" instead.

Paul-Licameli:
> Not really sure if I should "Request Changes" - I guess the defect I pointed out in the first conversation is being taken care of (or it will be shortly), and its solution will come up in a new PR - so I will select "Approve" instead.

No, my intention was to address the problems you found in this same PR, and now I have pushed again.  Please review again.

src/effects/EffectUI.cpp
```diff
@@ -1139,8 +1139,8 @@ void EffectUIHost::LoadUserPresets()
 void EffectUIHost::InitializeRealtime()
 {
    if (mSupportsRealtime && !mInitialized) {
-      mpState = RealtimeEffectManager::Get(*mProject)
-         .AddState(nullptr, PluginManager::GetID(&mEffect));
+      mpState =
```
pietro68, line 1142: test: open project with one track, instantiate a BassTreble before ever pressing play.

Again mpState will be returned nullptr, but this time because of a different reason:

the call to GetOwningProject() inside AudioIO::AddState will return nullptr.
this happens because the owning project is not set until `AudioIO::StartPortAudioStream` is reached, but that happens only after pressing play.

pietro68: Please watch out: even if I press play (and then stop) to set the owning project, and only then I instantiate a BassTreble, `AudioIO::AddState` will still return nullptr, but this time because `mpTransportState `will be null.

Paul-Licameli: > Please watch out: even if I press play (and then stop) to set the owning project, and only then I instantiate a BassTreble, `AudioIO::AddState` will still return nullptr, but this time because `mpTransportState `will be null.

Let's try this once more.  I made changes to fix this.

Paul-Licameli:
All right, one MORE time!  Please see the last two commits.  All else is unchanged.

pietro68 (approved):

Issue #2094, opened 2021-11-06
Use atomics properly in audio playback


Paul-Licameli:
**Describe the bug**
There are many places where the low-latency thread reads data written by other threads, or writes data for the others to read, and the writes may happen during play or record.

These should be reviewed and rewritten properly where needed with atomic variables, to avoid undefined behavior.  

Fetches and stores of those atomics should be rewritten with appropriate non-default memory orderings (less strict than sequential consistency) to avoid performance penalties.  Whether these can be simple relaxed orderings -- or whether acquire and release orderings, or fences, are necessary -- needs careful consideration in each case.

- [x] Remaining inappropriate uses of the `volatile` keyword should be removed.
- [x] Mute and solo states of Wave and Note tracks
- [x] Wave track pan and gain
- [x] AudioIOBase::mPaused
- [x] NoteTrack::mVisibleChannels
- [x] NoteTrack::mVelocity
- [x] AudioIoCallback::mMixerOutputVol

Also listing some non-examples, where the low-latency thread does read what other threads write, but all changes happen outside of active play or record and are properly synchronized; or there is initialization before play, but writes and re-reads only during play thereafter.  These should be commented.

- [ ] Channel type of a WaveTrack (presumably not editable during playback)
- [x] WaveTrack::mOldGain
- [x] AudioIOBase::mStreamToken (why is it `volatile`?)
- [x] AudioIOBase::mRate
- [x] AudioIoCallback::mListener
- [x] AudioIoCallback::mPauseRec
- [x] AudioIoCallback::mbMicroFades
- [x] AudioIoCallback::mPlaybackTracks
- [x] AudioIoCallback::mSoftwarePlaythrough
- [x] AudioIoCallback::mNumPlaybackChannels
- [x] AudioIoCallback::mNumCaptureChannels
- [x] AudioIoCallback::mDetectDropouts
- [x] AudioIoCallback::mDetectUpstreamDropouts
- [x] RecordingSchedule::mLatencyCorrection

Other more difficult things

* See also #2095
* Are the Envelope objects in Time tracks guaranteed not to be modified by the main thread?  (MIDIPlay may use PlaybackSchedule::RealDuration() which may use envelopes)

PR #2511, opened 2022-02-04
Correct atomics in playback

Paul-Licameli:
Resolves: #2094

Use atomics with relaxed memory order wherever the worker threads for audio playback and recording might read a value
that the main thread can concurrently change, so that all is properly done without "undefined behavior."

Remove all inappropriate `volatile` qualifiers, with one exception (addrressed in #2503).

Comment variables that are set by the main thread but remain unchanged during playback and recording, and so do not
require atomics.

Recommended:

### DIFF
--- a/libraries/lib-audio-devices/AudioIOBase.h
+++ b/libraries/lib-audio-devices/AudioIOBase.h
@@ -252,6 +252,7 @@ protected:
    int                 mStreamToken{ 0 };
 
    /// Audio playback rate in samples per second
+   /*! Read by worker threads but unchanging during playback */
    double              mRate;
 
    PaStream           *mPortStreamV19;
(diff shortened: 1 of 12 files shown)


libraries/lib-audio-devices/AudioIOBase.h
```diff
@@ -245,7 +245,7 @@ class AUDIO_DEVICES_API AudioIOBase /* not final */
    std::weak_ptr<AudacityProject> mOwningProject;
 
    /// True if audio playback is paused
-   bool                mPaused;
+   std::atomic<bool>   mPaused{ false };
```
pietro68, line 248: Compiling fails on windows because of missing #include <atomic> in this header. I added it (only locally) and could move on testing.

Paul-Licameli: Fixed in the rebased branch, no other changes

pietro68 (approved):

PR #2514, opened 2022-02-04
Effect cleanup part 4

Paul-Licameli:
Partly Resolves: #2084

Make the interfaace EffectUIHostInterface almost sufficient for the needs of EffectManager, so it has less dependency on
fat class Effect.

The remaining exception is the use of Effect::Startup, to be addressed later.

Recommended:

### DIFF
--- a/src/effects/EffectUI.h
+++ b/src/effects/EffectUI.h
@@ -95,7 +95,7 @@ private:
 
    AudacityProject &mProject;
    wxWindow *mParent;
-   EffectUIHostInterface &mEffect;
+   EffectUIHostInterface &mEffectUIHost;
    EffectUIClientInterface &mClient;
    RealtimeEffectState *mpState{ nullptr };
 
--- a/src/effects/EffectUI.h
+++ b/src/effects/EffectUI.h
@@ -33,7 +33,6 @@ struct AudioIOEvent;
 
 class AudacityCommand;
 class AudacityProject;
-class Effect;
 class RealtimeEffectState;
 
 class wxCheckBox;
@@ -45,7 +44,7 @@ public:
    // constructors and destructors
    EffectUIHost(wxWindow *parent,
                 AudacityProject &project,
-                Effect &effect,
+                EffectUIHostInterface &effect,
                 EffectUIClientInterface &client);
    virtual ~EffectUIHost();
 
@@ -96,7 +95,7 @@ private:
 
    AudacityProject &mProject;
    wxWindow *mParent;
-   Effect &mEffect;
+   EffectUIHostInterface &mEffect;
    EffectUIClientInterface &mClient;
    RealtimeEffectState *mpState{ nullptr };
 
@@ -147,7 +146,7 @@ class CommandContext;
 namespace  EffectUI {
 
    AUDACITY_DLL_API
-   wxDialog *DialogFactory( wxWindow &parent, EffectHostInterface &host,
+   wxDialog *DialogFactory( wxWindow &parent, EffectUIHostInterface &host,
       EffectUIClientInterface &client);
 
    /** Run an effect given the plugin ID */
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -93,10 +93,6 @@ public:
    //! Whether the effect sorts "above the line" in the menus
    virtual bool IsDefault() = 0;
 
-   // This will go away when all Effects have been updated to the new
-   // interface.
-   virtual bool IsLegacy() = 0;
-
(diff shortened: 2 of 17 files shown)


src/AudioIO.cpp
```diff
@@ -1849,6 +1891,74 @@ void AudioIO::FillPlayBuffers()
...
+            // Such as when a mono track is processed for stereo play?
+            // Then supply some non-null fake input buffers, because the
+            // various ProcessBlock overrides of effects may crash without it.
+            // But it would be good to find the fixes to make this unnecessary.
+            float **scratch = &mScratchPointers[mNumPlaybackChannels + 1];
+            while (iChannel < mNumPlaybackChannels)
+               pointers[iChannel++] = *scratch++;
+
+            if (len)
+               pScope->Process(vt, &pointers[0], mScratchPointers.data(), len);
```
pietro68, line 1958: A Debug build will assert here, reporting that pScope is empty. Unfortunately this bug can not be reproduced through a definite sequence of steps, so it is probably due to synchronization issues.
In general I found this happens when stopping playback, or when moving the playback position while playing by clicking on the ruler, however it might happen at the first stop/click or after a number of them.

pietro68, line 1958: A Debug build will assert here, reporting that pScope is empty. Unfortunately this bug can not be reproduced through a definite sequence of steps, so it is probably due to synchronization issues.
In general I found this happens when stopping playback, or when moving the playback position while playing by clicking on the ruler, however it might happen at the first stop/click or after a number of them.

pietro68, line 1958: A Debug build will assert here, reporting that pScope is empty. Unfortunately this bug can not be reproduced through a definite sequence of steps, so it is probably due to synchronization issues.
In general I found this happens when stopping playback, or when moving the playback position while playing by clicking on the ruler, however it might happen at the first stop/click or after a number of them.

pietro68, line 1958: A Debug build will assert here, reporting that pScope is empty. Unfortunately this bug can not be reproduced through a definite sequence of steps, so it is probably due to synchronization issues.
In general I found this happens when stopping playback, or when moving the playback position while playing by clicking on the ruler, however it might happen at the first stop/click or after a number of them.

pietro68, line 1958: A Debug build will assert here, reporting that pScope is empty. Unfortunately this bug can not be reproduced through a definite sequence of steps, so it is probably due to synchronization issues.
In general I found this happens when stopping playback, or when moving the playback position while playing by clicking on the ruler, however it might happen at the first stop/click or after a number of them.

pietro68, line 1958: A Debug build will assert here, reporting that pScope is empty. Unfortunately this bug can not be reproduced through a definite sequence of steps, so it is probably due to synchronization issues.
In general I found this happens when stopping playback, or when moving the playback position while playing by clicking on the ruler, however it might happen at the first stop/click or after a number of them.

pietro68: The commit which introduced this issue was `27cbce8` (I can not seem to comment directly there - github reports a page not found)

src/effects/EffectUI.h
```diff
@@ -96,7 +95,7 @@ class EffectUIHost final : public wxDialogWrapper
 
    AudacityProject &mProject;
    wxWindow *mParent;
-   Effect &mEffect;
+   EffectUIHostInterface &mEffect;
```
pietro68, line 98: It would help me read and understand this code some weeks from now, if we could rename `mEffect` to `mEffectUIHost` or `mEffectUIHostInterface`

src/effects/RealtimeEffectManager.cpp
```diff
@@ -161,8 +161,8 @@ void RealtimeEffectManager::ProcessStart()
 // This will be called in a different thread than the main GUI thread.
 //
 size_t RealtimeEffectManager::Process(Track *track,
-                                      float **buffers,
-                                      size_t numSamples)
+   float *const *buffers, float *const *scratch,
+   size_t numSamples)
```
pietro68, line 165: in my simple test (have one mono track, instantiate BassTreble eff, press play) ::Process will exit early at line 173 because `mSuspended `will be true.

src/AudioIO.cpp
pietro68: I see this has been fixed by checking `if (len && pScope)` so now when pScope is empty, effect processing will be skipped.

My question now is: pScope being sometimes empty is something that we can consider normal?

pietro68:
Just as a note to myself for later: I tested 70555d3 with BassTreble and it works.

Issue #2409, opened 2022-01-11
Proposal Auto Detect Audio Device Changes


petersampsonaudacity:
related: #2570

**Describe the bug**

When a new device is plugged in '''after''' Audacity has started, Audacity does not detect it.
1. You have to explicitly request a rescan with **Transport > Rescan Audio Devices**
2. This is not just a USB problem.
3. Some headphone-or-microphone jacks detect plug-in or removal of a device.

_Note that **Transport > Rescan Audio Devices** appears not to be readily "discoverable" by many users who post on the Forum._

This was originally logged back in 201 as Bugzilla Issue 41  https://bugzilla.audacityteam.org/show_bug.cgi?id=41
This issue was in bugzilla as [http://bugzilla.audacityteam.org/show_bug.cgi?id=41 Bug 41]
and then converted to a Wiki Proposal:
https://wiki.audacityteam.org/wiki/Proposal_Auto_Detect_Audio_Device_Changes

**To Reproduce**
1. Launch Audacity
2. plug in a USB audio device (soundcard or mic)
3. Try to select it the Device Toolbar
4. Observe: it is not present in the list

**Expected behavior**
Rescan devices when the relevant dialog is opened, if they have not been scanned in the last 5 seconds.

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: all
 - Version: all

**Additional context**
We get countless posts on the Forum regarding this ...

From the original Bugzilla entry
>Adding or removing an external audio device while Audacity is open will not be automatically reflected in the list in Device Toolbar or Devices Preferences. On Mac, unplugging then replugging an external device will give "error while opening sound device" (on Linux, pressing "Record" a second time will remove the error). '''Workaround:''' Use Transport > Rescan Audio Devices.

jserquera:
I would like to mention that this problem is really a mess in the classroom. I use Audacity, great software, in my classes, and solving this problem would be great.

Paul-Licameli:
This pull request includes maintenance of experimental code to to just this, which was written years ago but not enabled in releases.

https://github.com/audacity/audacity/pull/6204

avm:
Can we add a "Rescan..." option into every device menu until this is solved? The experimental code to auto-rescan, written years ago, may have to mature for many more years, but a "Rescan..." button in a discoverable place can't hurt (since we already have that button in a non-discoverable place, so presumably it works well enough).

ianalarina:
[...]

PR #2517, opened 2022-02-06
Proposal Auto Detect Audio Device Changes #2409

malevolentstrix:
Trigger Device Rescans whenever a user moves his mouse to the Audacity Device Toolbar or usethe Selection window via the keyboard shortcut or tries to move the mouse inside the Edit>Preferences>Devices
Dialog for device choices. Use Bind() with wxEVT_ENTER_WINDOW mouse event for the DeviceToolbar window and
DevicePrefs window.

Resolves: https://github.com/audacity/audacity/issues/2409

Recommended:

### DIFF
--- a/src/prefs/DevicePrefs.cpp
+++ b/src/prefs/DevicePrefs.cpp
@@ -83,6 +83,8 @@ void DevicePrefs::Populate()
 {
    // First any pre-processing for constructing the GUI.
    GetNamesAndLabels();
+   
+   Bind(wxEVT_ENTER_WINDOW, &DevicePrefs::OnToolbarEntry, this);
 
    // Get current setting for devices
    mPlayDevice = AudioIOPlaybackDevice.Read();
@@ -207,6 +209,11 @@ void DevicePrefs::PopulateOrExchange(ShuttleGui & S)
    S.EndStatic();
    S.EndScroller();
 
+}
+
+void DevicePrefs::OnToolbarEntry(wxMouseEvent &event)
+{
+   DeviceManager::Instance()->Rescan();
 }
 
 void DevicePrefs::OnHost(wxCommandEvent & e)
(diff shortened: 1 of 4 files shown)


petersampsonaudacity:
I tested this on W10 with audacity-win-3.2.0-alpha-20220206+0c9726d-x64-msvc2022  from:
https://github.com/audacity/audacity/actions/runs/1802664482

1) Launch Audacity
2) Plug in my Edirol UA-1EX
3) Click on output device in Device toolbar
4) Observe the UA-1EX is available for output
5) Click on input device in Device toolbar
6) Observe the UA-1EX is **_NOT_** available for output

6 above is actually a Windows issue with my USB device - when I plug in the USB-1EX Windows seen the device, it has it enabled for output but **_disabled_** for input (I have fallen foul of this and stumbled in the past_.  This was **_not_** the case with earlier versions of Windows from 15 or so years ago when I was using the UA-1EX to digitize ln LP collection - then just plugging it in made it available for both input and output

Once I tell Windows W10 to enable the UZ-1EX for input then plugging and unplugging the device makes it unavailable and available in the Device toolbar for both input and output - so my conclusion is that this fix works.

LWinterberg:
@Paul-Licameli didn't you test this the other day, too?

iamjithinjohn:
Just realized that the fix doesn't work properly because the function for rescanning can only run when playing and recording are stopped. So in this case, if I play any audio and move the mouse above the toolbar, the audio gets stopped. I'll try to make it to when the user clicks on the choice list.

Paul-Licameli:
> @Paul-Licameli didn't you test this the other day, too?

I made a PR that enabled other old experimental code, which I didn't write, to do it automatically, and I left it for others to test.

Jithin thought of a very different approach however which might also work.

petersampsonaudacity:
> Just realized that the fix doesn't work properly because the function for rescanning can only run when playing and recording are stopped. So in this case, if I play any audio and move the mouse above the toolbar, the audio gets stopped.

I tested this this today on W10 with audacity-win-3.2.0-alpha-20220206+0c9726d-x64-msvc2022 from:
https://github.com/audacity/audacity/actions/runs/1802664482

I confirm that this is indeed the case with this branch build

Paul-Licameli:
Compare PR https://github.com/audacity/audacity/pull/2457

LWinterberg:

Issue #2534, opened 2022-02-08
MessageBuffer improvements


Paul-Licameli:
Extract MessageBuffer from PlaybackSchedule as a reusable utility for lock-free inter-thread communication.

PR #2535, opened 2022-02-08
Message buffer utility

Paul-Licameli:
Resolves: #2534 

Extract MessageBuffer from PlaybackSchedule as a reusable utility for lock-free inter-thread communication.

Recommended:


libraries/lib-utility/MessageBuffer.h
```diff
@@ -0,0 +1,112 @@
...
+   NonInterfering<UpdateSlot> mSlots[2];
+
+   using Control = unsigned char;
+   //! Bit fields
+   enum : Control {
+      SlotToRead = 0x1, //!< Next slot to read, unless switched
+      SwitchSlot = 0x2, //!< Set only by writer; true only while not writing
+      ReadInProgress = 0x4, //!< Changed only by reader thread
+   };
+   static size_t ChooseReaderSlot(Control control)
```
crsib, line 36: A very complex function to understand. I think it is worth commenting on its meaning or rewriting it without bit magic.

Also, the choice of types is confusing as well.

pietro68 (approved):
I see the point of having a templated write, so that we have a unique method to write either const references or rvalue references. Fine by me.

Issue #2478, opened 2022-01-25
Add unit testing framework, integrate with CI


AnitaBats:

Paul-Licameli:
Can we have more explanation of this issue?  When it's done, what happen?  Will there be at least one trivial example of a unit test that is regularly exercised?

crsib:
There will be a non-trivial example based on `lib-string-utils` which will run on CI for every build

PR #2536, opened 2022-02-08
Unit testing

crsib:
Resolves: #2478 

This PR introduces a framework for automated testing in Audacity. It is now possible to define the following types of tests:

### Unit Tests

`add_unit_test(NAME name SOURCES file1 ... LIBRARIES lib1 ...)`

This function creates an executable called `${name}-test` from the sources files (`${file1}`, ...). and linked to libraries (`${lib1}`, ...).

Catch2 is linked implicitly. There is no need to define a main function in the unit test.

### Journal Tests

`add_journal_test(journal_file)`

Adds a test, that runs Audacity with the journal `${journal_file}`. The test name is based on the name component of the `${journal_file}`.

---

Both unit and journal tests are registered with the CTest framework. This PR implements unit tests for FromChars set of functions and a journal test that checks that we can run Audacity at all. 

To be able to run journal tests Audacity now suppresses a few modal dialogs, shown during the startup:

* update permission dialog,
* auto-recovery dialog,
* extension registration on Windows.

For libraries, if the `libraries/(name)/tests` directory is present - CMake will include the unit tests for the library automatically. 

The `Build Audacity` workflow will run both unit and journal tests for each build. In the future, journal tests can be separated into a different nightly workflow. 

### Examples

Adding a unit test:

```
add_unit_test(
   NAME
      lib-string-utils
   SOURCES
      FromCharsTests.cpp
   LIBRARIES
      lib-string-utils
)
```

Adding a journal test:

```
add_journal_test( "journal_sanity.txt" )
```

Running tests:

```
$ ctest -C Debug
```

---------------------------------------------------------------------------------------------------------------

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/AudacityTesting.cmake
+++ b/cmake-proxies/cmake-modules/AudacityTesting.cmake
@@ -1,0 +1,144 @@
+if( ${_OPT}has_tests )
+
+   set( TESTS_DIR "${CMAKE_BINARY_DIR}/tests" )
+
+   if( CMAKE_CONFIGURATION_TYPES )
+      set( TESTS_DEST_DIR "${CMAKE_BINARY_DIR}/tests/${CMAKE_CFG_INTDIR}" )
+   else()
+      set( TESTS_DEST_DIR "${CMAKE_BINARY_DIR}/tests/${CMAKE_BUILD_TYPE}" )
+   endif()
+
+   # Install Catch2 framework
+
+   add_conan_lib(
+      Catch2
+      catch2/2.13.8
+      REQUIRED
+      ALWAYS_ALLOW_CONAN_FALLBACK
+   )
+
+   # Setup default CTest arguments when running from IDE
+
+   set(CMAKE_CTEST_ARGUMENTS "--output-on-failure;--verbose;${CMAKE_CTEST_ARGUMENTS}")
+
+   enable_testing()
+
+   #[[
+      add_unit_test(NAME name SOURCES file1 ... LIBRARIES lib1 ...)
+
+      Creates an executable called ${name}-test from the source files ${file1}, ... and linked
+      to libraries ${lib1}, ... Catch2 is linked implicitly.
+
+      Create a CTest test called ${name}, that expects 0 code on success
+   ]]
+   function( add_unit_test )
+      cmake_parse_arguments(
+         ADD_UNIT_TEST # Prefix
+         "" # Options
+         "NAME" # One value keywords
+         "SOURCES;LIBRARIES"
+         ${ARGN}
+      )
+
+      if( NOT ADD_UNIT_TEST_NAME )
+         message( FATAL_ERROR "Missing required NAME parameter for the add_unit_test")
+      endif()
+
+      set( test_executable_name "${ADD_UNIT_TEST_NAME}-test" )
+
+      # Create test executable
+
+      add_executable( ${test_executable_name} ${ADD_UNIT_TEST_SOURCES} "${CMAKE_SOURCE_DIR}/tests/Catch2Main.cpp")
+      target_link_libraries( ${test_executable_name} ${ADD_UNIT_TEST_LIBRARIES} Catch2::Catch2 )
+
+      set_target_properties(
+         ${test_executable_name}
+         PROPERTIES
+            FOLDER "tests" # for IDE organization
+            RUNTIME_OUTPUT_DIRECTORY ${TESTS_DIR}
+            BUILD_RPATH ${_SHARED_PROXY_PATH}
+            # Allow running tests from Visual Studio by setting up the proper PATH
+            VS_DEBUGGER_ENVIRONMENT "PATH=${_SHARED_PROXY_PATH};%PATH%"
+      )
+
+      # Register unit test with CTest
+
+      add_test(
+         NAME
+            ${ADD_UNIT_TEST_NAME}
+         COMMAND
+            ${test_executable_name}
+      )
+
+      set_tests_properties(
+         ${ADD_UNIT_TEST_NAME}
+         PROPERTIES
+            LABELS "unit_tests"
+      )
+
+      if( WIN32 )
+         # On Windows, set the PATH so it points to the DLL location
+         # This is required to avoid invoking CopyLibs.
+
+         # CTest expects that ENVIRONMENT is a CMake list.
+         # Escape ';' so PATH is handled correctly
+         string(REPLACE ";" "\\;" escaped_path "$ENV{PATH}")
+
+         set_tests_properties(
+            ${ADD_UNIT_TEST_NAME}
+            PROPERTIES
+               ENVIRONMENT "PATH=$<SHELL_PATH:${_SHARED_PROXY_BASE_PATH}/$<CONFIG>>\\;${escaped_path}"
+         )
+      elseif( APPLE )
+         # We target an old version of macOS, disable std::uncaught_exceptions
+         target_compile_definitions( ${test_executable_name} PRIVATE CATCH_CONFIG_NO_CPP17_UNCAUGHT_EXCEPTIONS )
+         # Similar to Windows, but uses DYLD_FALLBACK_LIBRARY_PATH istead of PATH
+         set_tests_properties(
+            ${ADD_UNIT_TEST_NAME}
+            PROPERTIES
+               ENVIRONMENT "DYLD_FALLBACK_LIBRARY_PATH=$<SHELL_PATH:${_SHARED_PROXY_BASE_PATH}/$<CONFIG>>"
+         )
+      endif()
+   endfunction()
+
+   #[[
+      add_journal_test(journal_file)
+
+      Adds a test, that runs Audacity with the journal ${journal_file}.
+
+      Test name is based on the name component of the ${journal_file}
+   ]]
+   function( add_journal_test journal_file )
+      get_filename_component(test_name ${journal_file} NAME_WE)
+
+      if( APPLE )
+         # On macOS CMake will generate a placeholder that CTest fails to handle correctly,
+         # so we have to setup the path manually
+         set( audacity_target "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$<CONFIG>/Audacity.app/Contents/MacOS/Audacity" )
+      else()
+         set( audacity_target "$<TARGET_FILE:Audacity>" )
+      endif()
+
+      # Adds a test that invokes Audacity with the given journal file
+
+      add_test(
+         NAME
+            ${test_name}
+         COMMAND
+            ${audacity_target} --journal ${journal_file}
+      )
+
+      set_tests_properties(
+         ${ADD_UNIT_TEST_NAME}
+         PROPERTIES
+            LABELS "journal_tests"
+      )
+   endfunction()
+else()
+   # Just a placeholder for the cases unit testing is disabled
+   function(add_unit_test)
+   endfunction()
+
+   function( add_journal_test journal_file )
+   endfunction()
+endif()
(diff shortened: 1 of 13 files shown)


libraries/lib-string-utils/tests/FromCharsTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <catch2/catch.hpp>
+
+#include <string>
+#include <string_view>
+#include <type_traits>
+
+#include "FromChars.h"
+
+template <typename T>
+void TestFromChars(std::string_view input, T expextedValue, std::errc errc = {})
```
Paul-Licameli, line 20: "expectedValue"

libraries/lib-string-utils/tests/FromCharsTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+      std::to_string(std::numeric_limits<TestType>::min()) + "0",
+      std::numeric_limits<TestType>::min(), std::errc::result_out_of_range);
+
+   TestFromChars<TestType>(
+      std::to_string(std::numeric_limits<TestType>::max()) + "0",
+      std::numeric_limits<TestType>::max(), std::errc::result_out_of_range);
+}
+
+TEMPLATE_TEST_CASE(
+   "FromChars/unsigned integeres", "", unsigned short, unsigned int,
```
Paul-Licameli, line 70: "integers"

Paul-Licameli:
Wow!  This is wonderful.  Thanks for figuring all of this out.

I have not approved all yet -- I want to read and understand all of AudacityTesting.cmake.

Very likely those CMake functions will need enhancement with more options.  Journal tests may need to run with a certain prior environment setup, such as having a particular audacity.cfg file already in place at startup, and then the environment would need cleanup after the test.  CTest gives you the means to do such things by defining setup and teardown commands as "test" too in a certain dependency graph.  We may also want to associate labels with test so we can filter them and run subsets.

But all that can be future work and no reason to delay this.

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -0,0 +1,143 @@
+if( ${_OPT}has_tests )
+   if( CMAKE_CONFIGURATION_TYPES )
+      set( TESTS_DIR "${CMAKE_BINARY_DIR}/tests" )
+      set( TESTS_DEST_DIR "${CMAKE_BINARY_DIR}/tests/${CMAKE_CFG_INTDIR}" )
+   else()
+      set( TESTS_DIR "${CMAKE_BINARY_DIR}/tests" )
```
Paul-Licameli, line 6: You might lift the common assignment of `TEST_DIR` before the `if()`

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -0,0 +1,143 @@
...
+   # Setup default CTest arguments when running from IDE
+
+   set(CMAKE_CTEST_ARGUMENTS "--output-on-failure;--verbose;${CMAKE_CTEST_ARGUMENTS}")
+
+   enable_testing()
+
+   #[[
+      add_unit_test(NAME name SOURCES file1 ... LIBRARIES lib1 ...)
+
+      Creates an executable called ${name}-test from the sources files ${file1}, ... and linked
```
Paul-Licameli, line 28: "source files"

Paul-Licameli: It may be useful to add `PROPERTIES prop1 ...` and pass those to: https://cmake.org/cmake/help/latest/command/set_tests_properties.html?highlight=set_tests_properties

Paul-Licameli: And here are all the possible properties https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#test-properties

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -0,0 +1,143 @@
...
+      to libraries ${lib1}, ... Catch2 is linked implicitly.
+
+      Create a CTest test called ${name}, that expects 0 code on success
```
Paul-Licameli, line 31: ... unless one uses one of the properties:
* FAIL_REGULAR_EXPRESSION
* PASS_REGULAR_EXPRESSION
* SKIP_REGULAR_EXPRESSION
* SKIP_RETURN_CODE
* WILL_FAIL

cmake-proxies/cmake-modules/AudacityTesting.cmake
Paul-Licameli: Maybe the function will have other keyword arguments to abbreviate other combinations of properties.  Don't ask me yet, what exactly.  Experience will tell what is useful.

cmake-proxies/cmake-modules/AudacityTesting.cmake
crsib: I have a very strong opinion that unit test should return 0 on success

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -0,0 +1,143 @@
...
+      if( NOT ADD_UNIT_TEST_NAME )
+         message( FATAL_ERROR "Missing required NAME parameter for the add_unit_test")
+      endif()
+
+      set( test_executable_name "${ADD_UNIT_TEST_NAME}-test" )
+
+      # Create test executable
+
+      add_executable( ${test_executable_name} ${ADD_UNIT_TEST_SOURCES} "${CMAKE_SOURCE_DIR}/tests/Catch2Main.cpp")
+      target_link_libraries( ${test_executable_name} ${ADD_UNIT_TEST_LIBRARIES} Catch2::Catch2 )
```
Paul-Licameli, line 51: Catch2 needs to link something?  I thought it is just one amalgamated header file.  Or is that precisely what this line does, just adding in include path.

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -0,0 +1,143 @@
...
+
+      set_target_properties(
+         ${test_executable_name}
+         PROPERTIES
+            FOLDER "tests" # for IDE organization
+            RUNTIME_OUTPUT_DIRECTORY ${TESTS_DIR}
+            BUILD_RPATH ${_SHARED_PROXY_PATH}
+            # Allow running tests from Visual Studio by setting up the propper PATH
```
Paul-Licameli, line 59: "proper"

cmake-proxies/cmake-modules/AudacityTesting.cmake
crsib: "Modern" CMake uses `target_link_libraries` for header-only libraries as well.

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -0,0 +1,143 @@
...
+         )
+      elseif( APPLE )
+         # We target an old version of macOS, disable std::uncaught_exceptions
+         target_compile_definitions( ${test_executable_name} PRIVATE CATCH_CONFIG_NO_CPP17_UNCAUGHT_EXCEPTIONS )
+         # Similar to Windows, but uses DYLD_FALLBACK_LIBRARY_PATH istead of PATH
+         set_tests_properties(
+            ${ADD_UNIT_TEST_NAME}
+            PROPERTIES
+               ENVIRONMENT "DYLD_FALLBACK_LIBRARY_PATH=$<SHELL_PATH:${_SHARED_PROXY_BASE_PATH}/$<CONFIG>>"
+         )
```
Paul-Licameli, line 99: I wonder if these fixes you figured out for dynamic libraries should be reused with image-compiler too.  Or have you done that already?

Maybe a common cmake function for such steps then.

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -0,0 +1,143 @@
...
+      endif()
+
+      # Adds a test that invokes Audacity with the given journal file
+
+      add_test(
+         NAME
+            ${test_name}
+         COMMAND
+            ${audacity_target} --journal ${journal_file}
+      )
```
Paul-Licameli, line 128: Again maybe use a common low function for add_test, and consider how to pass other command arguments and options into it.

cmake-proxies/cmake-modules/AudacityTesting.cmake
crsib: They can be reused, but the image-compiler is a) rarely built, b) does not test the functionality of a library. For these reasons, it is left as-is for now.

Paul-Licameli (approved):
I have finished my review.
 
Just some spelling mistakes in comments in cmake scripts that should be fixed immediately.  Otherwise, suggestions for improvements, but nothing that needs to be done before merging this.

Paul-Licameli:
I'm trying it in the Xcode build but I don't yet see how to build the tests.

Paul-Licameli:
> 

Oh forget that!  The directory tests was a bit of junk just on my own file system, from my own experiements with tests.

crsib:
This looks to be your local issue, lib-strings doesn't have tests directory. 

I think this is leftover from the time you explored CTest

Paul-Licameli (approved):
Reapproval

Paul-Licameli:
OK now I see tests in the Xcode browser.  Waiting to build and run them.

Paul-Licameli:
There are all those CI build failures however

crsib:
Gosh Python word is bad :-( 

https://github.com/aws/aws-sam-cli/issues/3661

Paul-Licameli:
"Error: The process '/Users/runner/hostedtoolcache/Python/3.10.2/x64/bin/conan' failed with exit code 1"

I was glancing quickly at that, and misread the word "headache" in there

crsib:
Some third-party library broke half of the world again...
And I had issues with it when I was doing offline Linux builds as well. It is simply improperly packaged.

Paul-Licameli:
I have built the tests in XCode, and confirmed that build fails if I edit the test program to have a syntax error.

Now I added `REQUIRE(false)` and I want to observe the consequences of a test failure.  How?

crsib:
There are different ways:

1. You can select a specific test as a run target on XCode (if you want to debug the failure)
2. You can build RUN_TESTS

Paul-Licameli:
The journal sanity test makes a lot of spam, and so will every journal test:
```
2: 06:47:42: Debug: Warning: command doesn't exist: 'Cut'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Copy'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Paste'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Trim'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Silence'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Undo'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Redo'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomIn'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomOut'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomToggle'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomSel'
2: 06:47:42: Debug: Warning: command doesn't exist: 'FitInWindow'
2: 06:47:42: Debug: Unrecognized accel key 'NUMPAD_ENTER', accel string ignored.
```

Let's have an issue to clean that up

cmake-proxies/cmake-modules/AudacityTesting.cmake
Paul-Licameli: Got it

crsib:
Well, `add_journal_test` will be modified accordingly then :)

Like we can say `add_journal_test( fail.txt EXPECT_FAIL )` or similar.

PR #2537, opened 2022-02-09
Independent effect states

Paul-Licameli:
Resolves: #2538

EffectDefinitionInterface has alternative const methods using externalized settings objects: factory, save, load; but default implementations that just call-through to old functions, violating const-correctness and changing the single state as before.

EffectManager maintains a default settings object with each effect.

Each RealtimeEffectState has its own distinct settings object.

However no effect yet implements really independent settings.  The intent is to migrate existing effects to the new system
incrementally and phase out the old interface.  This branch accomplishes some of the necessary rewriting of other code in
terms of the new interface.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -97,6 +97,46 @@ bool EffectDefinitionInterfaceEx::LoadSettings(
       return false;
 }
 
+bool EffectDefinitionInterfaceEx::LoadUserPreset(
+   const RegistryPath & name, Settings &settings) const
+{
+   if (auto pEffect = FindMe(settings))
+      // Call through to old interface
+      return pEffect->LoadUserPreset(name);
+   else
+      return false;
+}
+
+bool EffectDefinitionInterfaceEx::SaveUserPreset(
+   const RegistryPath & name, const Settings &settings) const
+{
+   if (auto pEffect = FindMe(settings))
+      // Call through to old interface
+      return pEffect->SaveUserPreset(name);
+   else
+      return false;
+}
+
+bool EffectDefinitionInterfaceEx::LoadFactoryPreset(
+   int id, Settings &settings) const
+{
+   if (auto pEffect = FindMe(settings))
+      // Call through to old interface
+      return pEffect->LoadFactoryPreset(id);
+   else
+      return false;
+}
+
+bool EffectDefinitionInterfaceEx::LoadFactoryDefaults(
+   Settings &settings) const
+{
+   if (auto pEffect = FindMe(settings))
+      // Call through to old interface
+      return pEffect->LoadFactoryDefaults();
+   else
+      return false;
+}
+
 EffectDefinitionInterfaceEx *
 EffectDefinitionInterfaceEx::FindMe(const Settings &settings) const
 {
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -56,6 +56,42 @@ bool EffectDefinitionInterface::IsHiddenFromMenus()
    return false;
 }
 
+auto EffectDefinitionInterfaceEx::MakeSettings() const -> Settings
+{
+   // Temporary default implementation just saves self
+   // Cast away const! Capture pointer to self
+   return Settings( const_cast<EffectDefinitionInterfaceEx*>(this) );
+}
+
+bool EffectDefinitionInterfaceEx::SaveSettings(
+   const Settings &settings, CommandParameters & parms) const
+{
+   if (auto pEffect = FindMe(settings))
+      // Call through to old interface
+      return pEffect->GetAutomationParameters(parms);
+   else
+      return false;
+}
+
+bool EffectDefinitionInterfaceEx::LoadSettings(
+   CommandParameters & parms, Settings &settings) const
+{
+   if (auto pEffect = FindMe(settings))
+      // Call through to old interface
+      return pEffect->SetAutomationParameters(parms);
+   else
+      return false;
+}
+
+EffectDefinitionInterfaceEx *
+EffectDefinitionInterfaceEx::FindMe(const Settings &settings) const
+{
+   if (auto ppEffect = settings.cast<EffectDefinitionInterfaceEx*>();
+       ppEffect && *ppEffect == this)
+      return *ppEffect;
+   return nullptr;
+}
+
 //bool EffectDefinitionInterface::DefineParams(ShuttleParams & S)
 //{
 //   return false;
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -1,4 +1,4 @@
-/**********************************************************************
+/*!********************************************************************
 
   Audacity: A Digital Audio Editor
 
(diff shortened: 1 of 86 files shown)


libraries/lib-components/EffectInterface.h
```diff
@@ -131,10 +138,69 @@ class COMPONENTS_API EffectDefinitionInterface  /* not final */ : public Compone
...
+{
+public:
+   /*! @name old-settings-interface
+    Old interface for saving and loading non-externalized settings
+    */
+   //! @{
+   //! Save current settings into parms
+   virtual bool GetAutomationParameters(CommandParameters & parms) = 0;
+   //! Change settings to those stored in parms
+   virtual bool SetAutomationParameters(CommandParameters & parms) = 0;
```
pietro68, line 185: Could we have `const CommandParameters&`  instead? I assume the intent of this method is to only read parms

vsverchinsky:
There seems to be a bug in gcc 9.3 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90415 which is used to compile "Ubuntu 20.04 AppImage (x86_64)" target that prevents the code from being compiled

libraries/lib-components/EffectInterface.h
```diff
@@ -70,7 +71,8 @@ struct EffectSettings : std::any{ using any::any; };
 //! another thread; to be used in the main thread, only.
 /*! Updates are communicated atomically both ways.  The address of Get() should
  not be relied on as unchanging between calls. */
-class COMPONENTS_API EffectSettingsAccess {
+class COMPONENTS_API EffectSettingsAccess
```
vsverchinsky, line 74: Can be squashed with "Independent effect states" commit

src/EffectHostInterface.h
```diff
@@ -49,12 +50,15 @@ class AUDACITY_DLL_API EffectHostInterface
 class wxDialog;
 class wxWindow;
 class EffectUIClientInterface;
-
+class EffectSettingsAccess;
+using EffectSettingsAccessPtr = std::shared_ptr<EffectSettingsAccess>;
```
vsverchinsky, line 54: Suggest moving this alias to `EffectUIHostInterface` scope

libraries/lib-components/EffectInterface.cpp
```diff
@@ -56,6 +56,42 @@ bool EffectDefinitionInterface::IsHiddenFromMenus()
    return false;
 }
 
+auto EffectDefinitionInterfaceEx::MakeSettings() const -> Settings
+{
+   // Temporary default implementation just saves self
+   // Cast away const! Capture pointer to self
+   return Settings( const_cast<EffectDefinitionInterfaceEx*>(this) );
```
vsverchinsky, line 77: Settings object always refer to a current state, so `RealtimeEffectState` too. This is for another PR, right?

src/effects/EffectManager.cpp
```diff
@@ -730,13 +730,39 @@ void EffectManager::SetBatchProcessing(const PluginID & ID, bool start)
...
+
+std::pair<EffectUIHostInterface *, EffectSettings *>
+EffectManager::GetEffectAndDefaultSettings(const PluginID & ID)
+{
+   auto &results = DoGetEffect(ID);
+   if (results.effect) {
+      auto &settings = results.settings;
+      if (!settings.has_value())
+         // Create on demand
+         settings = results.effect->GetDefinition().MakeSettings();
```
vsverchinsky, line 754: I noticed that `RealtimeEffectState` uses `MakeSettings` directly to create it's own state (unique, supposedly), then the question is why do we need this function? I've searched for the uses of this function and turned out that it's indirectly used by `EffectManager::PromptUser` and `EffectUI::DoEffect` which, I suppose, should be parametrized with settings as some other functions in this PR. This two functions, in turn, are used in many places where 'default settings' may be supplied.

src/effects/EffectUI.cpp
```diff
@@ -882,16 +883,24 @@ void EffectUIHost::OnUserPreset(wxCommandEvent & evt)
 {
    int preset = evt.GetId() - kUserPresetsID;
    
-   mEffectUIHost.GetDefinition()
-      .LoadUserPreset(mEffectUIHost.GetUserPresetsGroup(mUserPresets[preset]));
+   // Make mutable copy
+   auto settings = mpAccess->Get();
```
vsverchinsky, line 887: After object copied it's contents will be discarded by `LoadUserPreset`

src/effects/RealtimeEffectState.cpp
```diff
@@ -11,8 +11,126 @@
...
+      explicit ToMainSlot(const EffectSettings &settings)
+         // Copy std::any
+         : mSettings{ settings }
+      {}
+      ToMainSlot &operator=(ToMainSlot &&) = default;
+
+      // Worker thread writes the slot
+      ToMainSlot& operator=(EffectAndSettings &&arg) {
+         // This happens during MessageBuffer's busying of the slot
+         arg.effect.CopySettingsContents(arg.settings, mSettings);
```
vsverchinsky, line 68: I'm still not sure that what `EffectSettings` object represents here. Is it supposed to hold values for every parameter that could be changed? If so, isn't it very expensive then?

libraries/lib-components/EffectInterface.h
```diff
@@ -131,10 +140,73 @@ class COMPONENTS_API EffectDefinitionInterface  /* not final */ : public Compone
...
+    */
+   virtual bool SaveSettings(
+      const Settings &settings, CommandParameters & parms) const = 0;
+
+   //! Restore settings from keys and values
+   /*!
+    @return true on success
+    */
+   virtual bool LoadSettings(
+      CommandParameters & parms, Settings &settings) const = 0;
```
pietro68, line 192: This is meant to copy/map a CommandParameters to a Settings (at least that's how I understand it)
Should it not be a `const CommandParameters&` then? or at least, this is the dual of `SaveSettings`, and the ref to Settings is const there

pietro68 (approved):
I think that at this point (and especially thanks to the live explanations) I can get a general idea of what this changes are going to achieve, and how.

libraries/lib-components/EffectInterface.cpp
Paul-Licameli: I'm not sure I understand the question.

The Settings object has now way to hold a reference to a RealtimeEffectState object.  But, each RealtimeEffectState object will hold an instance of Settings.  Maybe you meant that.

Besides which, the global registry of effects, EffectsManager, will hold Settings.  That is for purposes of applying effects in the old, destuctive way.  Audacity now has a memory of the last settings you used for each effect, and applies the effect without prompting you again with a dialog, if you use the "Repeat Last Effect" command.  Also those settings appear again if you do reopen the dialog for another destructive application.

So if the Effect object becomes stateless, EffectManager must store one more Settings with the effect in the registry.

src/effects/EffectManager.cpp
Paul-Licameli: If you were reviewing my commits sequentially, then you may written this before seeing that two later, I do that.  PromptUser is one function in which the default settings are retrieved from EffectManager.  The other is EffectUI::DialogFactory and DoEffect uses a pointer to that function.

src/effects/EffectUI.cpp
Paul-Licameli: But observe that `mpAccess->Set(std::move(settings))` follows.

src/EffectHostInterface.h
Paul-Licameli: Good idea

src/effects/EffectUI.cpp
Paul-Licameli: The reading, modifying, and rewriting of EffectSettings stored elsewhere is done this long way using EffectSettingsAccess.  The last commit in this branch introduces another implementation of that abstract class, which handles inter-thread communication of changes of settings.

src/effects/RealtimeEffectState.cpp
Paul-Licameli: That is the idea.

For each of our built-in effects, that will only be a few numbers and booleans, and `CopySettingsContents` would be overriden just to copy them, rather than copying the std::any which implies memory allocation.

libraries/lib-components/EffectInterface.h
Paul-Licameli: You are right, but when I try to do that, there is some strange code in Nyquist.cpp that is mutating the parameters sometimes.  Probably unnecessarily, and I will figure out a fix.  I have issue #2578 to address these const correctness issues later.

PR #2554, opened 2022-02-14
Transportstate fix - part 1

pietro68:
Alone this simple moving resolved the issue, because:
in all instances where mpTransportState was found to be empty during ::TransformBuffers, the call stack
showed that the call was originated in response to a request to process "once", and never in response to process "LoopRunning".

In ::StopStream, a call to process "once" is always requested, because `if (mStreamToken > 0)` will always
be satisfied - I can not seem to reproduce a case when it is not (but somebody else might know better!)

As long as `if (mStreamToken > 0)` always happens, there is no point in introducing what we discussed earlier,
(i.e. a new atomic and a waiting loop on that in order to signal that "LoopRunning" processing is really stopped)
it would be an unnecessary complication, because as the atomic to process "once" is later set to true and then
waited for until it is "false", this will always happen after the last processing coming from "LoopRunning".

However, if there are ever cases when `if (mStreamToken > 0)` is false, then we need this extra atomic, to wait
to be sure that "LoopRunning" processing came to a stop, before resetting mpTransportState. I have all this extra
things ready into a stash anyway - first we must decide if this extra complication is necessary.

Recommended:

### DIFF
--- a/src/AudioIO.cpp
+++ b/src/AudioIO.cpp
@@ -999,10 +999,10 @@ int AudioIO::StartStream(const TransportTracks &tracks,
    // audio thread call TrackBufferExchange here makes the code more predictable, since
    // TrackBufferExchange will ALWAYS get called from the Audio thread.
    mAudioThreadShouldCallTrackBufferExchangeOnce
-      .store(true, std::memory_order_relaxed);
+      .store(true, std::memory_order_release);
 
    while( mAudioThreadShouldCallTrackBufferExchangeOnce
-      .load(std::memory_order_relaxed)) {
+      .load(std::memory_order_acquire)) {
       using namespace std::chrono;
       auto interval = 50ms;
       if (options.playbackStreamPrimer) {
@@ -1751,11 +1751,11 @@ AudioThread::ExitCode AudioThread::Entry()
       gAudioIO->mAudioThreadTrackBufferExchangeLoopActive
          .store(true, std::memory_order_relaxed);
       if( gAudioIO->mAudioThreadShouldCallTrackBufferExchangeOnce
-         .load(std::memory_order_relaxed) )
+         .load(std::memory_order_acquire) )
       {
          gAudioIO->TrackBufferExchange();
          gAudioIO->mAudioThreadShouldCallTrackBufferExchangeOnce
-            .store(false, std::memory_order_relaxed);
+            .store(false, std::memory_order_release);
       }
       else if( gAudioIO->mAudioThreadTrackBufferExchangeLoopRunning
          .load(std::memory_order_relaxed))
@@ -3138,9 +3138,9 @@ int AudioIoCallback::CallbackDoSeek()
 
    // Reload the ring buffers
    mAudioThreadShouldCallTrackBufferExchangeOnce
-      .store(true, std::memory_order_relaxed);
+      .store(true, std::memory_order_release);
    while( mAudioThreadShouldCallTrackBufferExchangeOnce
-      .load(std::memory_order_relaxed) )
+      .load(std::memory_order_acquire) )
    {
       using namespace std::chrono;
       std::this_thread::sleep_for(50ms);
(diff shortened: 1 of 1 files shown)


Paul-Licameli:
"In ::StopStream, a call to process "once" is always requested, because if (mStreamToken > 0) will always
be satisfied - I can not seem to reproduce a case when it is not (but somebody else might know better!)"

The input toolbar, with the small microphone image, says "Click to start monitoring".  If you do that, the meter responds to sound, but there is no playback or recording.  If you click again, you reach this test and it is false.

In that case TrackBufferExchange is not doing anything and there is no need to send it a message to stop.

src/AudioIO.cpp
```diff
@@ -1554,6 +1552,12 @@ void AudioIO::StopStream()
       }
    }
 
+
+   // No longer need effects processing. This must be done after the stream is stopped
+   // to prevent the callback from being invoked after the effects are finalized.
+   mpTransportState.reset();
```
Paul-Licameli, line 1558: This line moves very much later, so you must ask whether reodering it with any of the things it passed over has any important consequences.  I think not.  Still, I would not move it so very late.

Paul-Licameli: I suggest:  split the `if (mStreamToken > 0) {` block, that is close it earlier, then test the condition again.  In between, destroy state.

The block `if (mPlaybackTracks.size() > 0)` is harmless if it is outside the `if (mStreamToken > 0) {` block, and now destroying of state is next to it.

And that is good:  now the cleanup steps, duplicated here and in `AudioIO::StartStreamCleanup`, are next to resetting of mpTransportState in both places.

That means the other cleanup I would like to see -- moving the repeated steps into the destructor of mpTransportState, eliminating duplication -- can be done as a next step in a later PR.

Paul-Licameli: The elimination. of duplicate code can be a later pull request.

Do the things I suggested though, and I'm not yet satisfied.

Use stronger memory ordering, acquire and release, for `mAudioThreadShouldCallTrackBufferExchangeOnce` before I consider this fix correct.  But maybe the other atomic variables are not needed yet.

Paul-Licameli (changes requested):
Maybe we can do a simplifer fix than I had described last week, but not as simple as this pull request stands now.

src/AudioIO.cpp
pietro68: Alright, suggestions implemented in next commit

src/AudioIO.cpp
```diff
@@ -1463,31 +1463,39 @@ void AudioIO::StopStream()
       // call TrackBufferExchange one last time (it normally would not do so since
       // Pa_GetStreamActive() would now return false
       mAudioThreadShouldCallTrackBufferExchangeOnce
-         .store(true, std::memory_order_relaxed);
-      while( mAudioThreadShouldCallTrackBufferExchangeOnce
-         .load(std::memory_order_relaxed) )
+         .store(true, std::memory_order_release);
+      while (mAudioThreadShouldCallTrackBufferExchangeOnce
+         .load(std::memory_order_acquire))
```
Paul-Licameli, line 1468: This is still incomplete without changing the uses of this atomic in `AudioThread::Entry` too, with more than
relaxed ordering.

Lock-free synchronization with an atomic variable requires both communicating threads to use the right memory orderings.

Paul-Licameli (changes requested):
The synchronization is still incomplete

pietro68:
Alright - apologies for creating two PRs for the same problem, I did some renaming to make things less confusing. 

We now have this PR (2554 - renamed to Transportate fix part 1) and 2557 (renamed to Transportate fix part 2). If you are ok with it, I would move the requests/discussion threads still open here, to 2557 (Transportate fix part 2) and continue there only.

Paul-Licameli (approved):

PR #2559, opened 2022-02-15
Replace call to a deprecated WINAPI function in VST3Effect

vsverchinsky:
An obsolete `SHGetFolderPathW` was used to retrieve system folder path, replaced with `SHGetKnownFolderPath`, also fixes wrong `#include`

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -45,7 +45,7 @@
 #include "VST3OptionsDialog.h"
 
 #ifdef __WXMSW__
-#include <ShlObj_core.h>
+#include <shlobj.h>
 #elif __WXGTK__
 #include "internal/x11/SocketWindow.h"
 #endif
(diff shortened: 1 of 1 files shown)


pietro68:
I get a different set of error messages now:

```
effects\VST3\VST3Effect.cpp(104,28): error C2065: 'FOLDERID_ProgramData': undeclared identifier
effects\VST3\VST3Effect.cpp(104,50): error C2065: 'KF_FLAG_DEFAULT': undeclared identifier
effects\VST3\VST3Effect.cpp(104,7): error C3861: 'SHGetKnownFolderPath': identifier not found
```

crsib:
@pietro68 do you have Windows SDK installed? I do not see any issues on CI or locally

pietro68:
I am not sure about what is going on at this point. Yes, it looks like I do not have Windows SDK installed (see attached picture) but at the same time Vitaly suggested to me a fix which worked i.e. I can build even without the SDK. Is the SDK a requirement, and if yes, what version and we should add it as a requirement to the public page instructing how to build Audacity
Capture
?

pietro68:
I seem to have Windows 10 SDK version 10.0.19041.0, but I also seem to not have both older and newer versions of it, at least according to my installation details (see attached picture) - anyway, Vitaly suggested a different header #include that solved the problem for me.

Do we need to list a specific SDK version on the how to build page?
Capture

pietro68 (approved):
Works for me now, thanks. Please see my comment on the conversation.

crsib:
You actually do have SDK installed (the second line in you screenshot)

Issue #2543, opened 2022-02-10
WaveTrack naming template


vsverchinsky:
The name of the WaveTrack should be generated according to this template:
`[DefaultTrackName] [N]`
Where `N` is minimum possible positive number so that name is unique.

Est.: ~2hrs

petersampsonaudacity:
I've just been playing around with @vsverchinsky  Vitaly's branch build for Issue #2543 re track naming. 
audacity-win-3.2.0-alpha-20220215+2e0317d-x64-msvc2022  

One little residual I note is that if you:
a) don't have Track name overlays turned on, or
b) don't have enough of a clip showing to show the full clip name
then the nice new track name differentiation is lost.  
Less of a problem with the clip-name IRL - but the track name in the Track Control Panel remains an issue.

![image](https://user-images.githubusercontent.com/50205984/154237673-5917be74-de3e-48f8-b9b4-455e48514e1d.png)

A simple solution would be to change the track nomenclature to:
Audio 1
Audio 2
etc.

vsverchinsky:
@petersampsonaudacity I've changed the default name to "Audio", now it should be fine, there was also changes that affects the text composed for screen readers - now the template is always "Track [N]: [track name]"

petersampsonaudacity:
Hi @vsverchinsky  Vitaly, 

testing on W10 with your branch audacity-win-3.2.0-alpha-20220216+0beecb3-x64-msvc2022 
that looks good to me.  👍 

There is residual issue where Label tracks are always labelled Label Track.  Perhaps, for diferentiation, they should be labelled:
Labels 1
Labels 2
etc.

![image](https://user-images.githubusercontent.com/50205984/154472415-45a79eea-a5e2-499a-aeff-4c8e3f7c2834.png)

_Note that the label tracks do not have clip-handle drag-bars so the is no name display that way - only if you turn on "Show track name as overlay" in Tracks preferences._

petersampsonaudacity:
**Time Track** is fine as it as - as there can only ever be one time track in a project.

petersampsonaudacity:
@LWinterberg this does cause an interesting problem for the manual Leo as there are many, many, places, in the manual where the labelling in the tracks in the Track Control Panel is "**Audio Track**" and "**Label Track**"

petersampsonaudacity:
> @petersampsonaudacity As you suggested I also changed the default name of the label tracks, 

[...]

PR #2561, opened 2022-02-15
Issue 2543: wave track naming

vsverchinsky:
Resolves: #2543 

Note that the logic of the name assignment behind track copying wasn't changed at all (copy will always receive the name of the original track).

Recommended:

### DIFF
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -66,7 +66,6 @@ void Track::Init(const Track &orig)
 {
    mId = orig.mId;
 
-   mDefaultName = orig.mDefaultName;
    mName = orig.mName;
 
    mSelected = orig.mSelected;
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -450,6 +450,27 @@ TrackList::~TrackList()
    Clear(false);
 }
 
+wxString TrackList::MakeUniqueTrackName(const wxString& baseTrackName) const
+{
+   int n = 1;
+   while(true)
+   {
+      auto name = wxString::Format("%s %d", baseTrackName, n++);
+
+      bool found {false};
+      for(const auto track : Any())
+      {
+         if(track->GetName() == name)
+         {
+            found = true;
+            break;
+         }
+      }
+      if(!found)
+         return name;
+   }
+}
+
 void TrackList::RecalcPositions(TrackNodePointer node)
 {
    if ( isNull( node ) )
--- a/src/DropoutDetector.cpp
+++ b/src/DropoutDetector.cpp
@@ -24,16 +24,14 @@ static AudacityProject::AttachedObjects::RegisteredFactory sKey {
       [&project](RecordingDropoutEvent &evt){
          evt.Skip();
          // Make a track with labels for recording errors
-         auto uTrack = std::make_shared<LabelTrack>();
-         auto pTrack = uTrack.get();
          auto &tracks = TrackList::Get( project );
-         tracks.Add( uTrack );
+
          /* i18n-hint:  A name given to a track, appearing as its menu button.
           The translation should be short or else it will not display well.
           At most, about 11 Latin characters.
           Dropout is a loss of a short sequence of audio sample data from the
           recording */
-         pTrack->SetName(_("Dropouts"));
+         auto pTrack = LabelTrack::Create(tracks, tracks.MakeUniqueTrackName(_("Dropouts")));
          long counter = 1;
          for (auto &interval : evt.intervals)
             pTrack->AddLabel(
--- a/src/LabelTrack.cpp
+++ b/src/LabelTrack.cpp
@@ -55,6 +55,11 @@ static ProjectFileIORegistry::ObjectReaderEntry readerEntry{
    LabelTrack::New
 };
 
+wxString LabelTrack::GetDefaultName()
+{
+   return _("Labels");
+}
+
 LabelTrack *LabelTrack::New( AudacityProject &project )
 {
    auto &tracks = TrackList::Get( project );
@@ -63,12 +68,24 @@ LabelTrack *LabelTrack::New( AudacityProject &project )
    return result;
 }
 
+LabelTrack* LabelTrack::Create(TrackList& trackList, const wxString& name)
+{
+   auto track = std::make_shared<LabelTrack>();
+   track->SetName(name);
+   trackList.Add(track);
+   return track.get();
+}
+
+LabelTrack* LabelTrack::Create(TrackList& trackList)
+{
+   return Create(trackList, trackList.MakeUniqueTrackName(GetDefaultName()));
+}
+
 LabelTrack::LabelTrack():
    Track(),
    mClipLen(0.0),
    miLastLabel(-1)
 {
-   SetName(_("Label Track"));
 }
 
 LabelTrack::LabelTrack(const LabelTrack &orig) :
--- a/src/LabelTrack.cpp
+++ b/src/LabelTrack.cpp
@@ -68,8 +68,7 @@ LabelTrack::LabelTrack():
    mClipLen(0.0),
    miLastLabel(-1)
 {
-   SetDefaultName(_("Label Track"));
-   SetName(GetDefaultName());
+   SetName(_("Label Track"));
 }
 
 LabelTrack::LabelTrack(const LabelTrack &orig) :
--- a/src/LabelTrack.h
+++ b/src/LabelTrack.h
@@ -88,8 +88,22 @@ class AUDACITY_DLL_API LabelTrack final
    , public wxEvtHandler
 {
  public:
+   static wxString GetDefaultName();
+
    // Construct and also build all attachments
    static LabelTrack *New(AudacityProject &project);
+
+   /**
+    * \brief Create a new LabelTrack with specified @p name and append it to the @p trackList
+    * \return New LabelTrack with custom name
+    */
+   static LabelTrack* Create(TrackList& trackList, const wxString& name);
+
+   /**
+    * \brief Create a new LabelTrack with unique default name and append it to the @p trackList
+    * \return New LabelTrack with unique default name
+    */
+   static LabelTrack* Create(TrackList& trackList);
 
    LabelTrack();
    LabelTrack(const LabelTrack &orig);
--- a/src/TrackPanelAx.cpp
+++ b/src/TrackPanelAx.cpp
@@ -36,8 +36,6 @@
 #include "LabelTrack.h"
 #include "NoteTrack.h"
 #include "TimeTrack.h"
-#include "WaveTrack.h"
-
 
 wxDEFINE_EVENT(EVT_TRACK_FOCUS_CHANGE, wxCommandEvent);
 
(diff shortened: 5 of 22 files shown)


libraries/lib-track/Track.cpp
```diff
@@ -450,6 +450,28 @@ TrackList::~TrackList()
    Clear(false);
 }
 
+wxString TrackList::MakeUniqueTrackName(const wxString& baseTrackName) const
+{
+   auto name = baseTrackName;
```
vsverchinsky, line 455: Unnecessary assignment

src/TrackPanelAx.cpp
```diff
@@ -368,33 +363,8 @@ wxAccStatus TrackPanelAx::GetName( int childId, wxString* name )
...
-#ifdef USE_MIDI
-                ,
-               [&](const NoteTrack *) {
-                  /* i18n-hint: This is for screen reader software and indicates that
-                     this is a Note track.*/
-                  name->Append( wxT(" ") + wxString(_("Note Track")));
-               }
-#endif
-            );
+            name->Printf(_("Track %d"), TrackNum(t));
```
Paul-Licameli, line 366: Do not delete the i18n_hint comment that helps the translators

src/TrackPanelAx.cpp
```diff
@@ -368,33 +363,8 @@ wxAccStatus TrackPanelAx::GetName( int childId, wxString* name )
...
-                  /* i18n-hint: This is for screen reader software and indicates that
-                     this is a Time track.*/
-                  name->Append( wxT(" ") + wxString(_("Time Track")));
-               }
-#ifdef USE_MIDI
-                ,
-               [&](const NoteTrack *) {
-                  /* i18n-hint: This is for screen reader software and indicates that
-                     this is a Note track.*/
-                  name->Append( wxT(" ") + wxString(_("Note Track")));
```
Paul-Licameli, line 394: Why was it necessary for resolving this issue, to delete this logic that appends "Label Track", "Note Track", etc. to `name`?

This name is NOT the name you will see on the screen.  It is a name that the screen reader will speak.

@DavidBailes may not agree with this change.

You are removing the Default Name (I didn't like it either) so the suffix "Label Track", etc. so now you are prepending "Track #n" always.  I think that is all right.

Paul-Licameli (changes requested):

libraries/lib-track/Track.cpp
Paul-Licameli: You review yourself?  :-)

src/import/ImportPlugin.cpp
```diff
@@ -69,5 +69,5 @@ sampleFormat ImportFileHandle::ChooseFormat(sampleFormat effectiveFormat)
 std::shared_ptr<WaveTrack> ImportFileHandle::NewWaveTrack(
    WaveTrackFactory &trackFactory, sampleFormat effectiveFormat, double rate)
 {
-   return trackFactory.NewWaveTrack(ChooseFormat(effectiveFormat), rate);
+   return trackFactory.Create(ChooseFormat(effectiveFormat), rate);
```
Paul-Licameli, line 72: Certain import file formats might contain multiple channels that become multiple tracks.  Should you consider giving them unique names?  But in the usual case of one track, I understand that you would not want to add the numeral.

src/WaveTrack.h
```diff
@@ -623,14 +623,23 @@ class AUDACITY_DLL_API WaveTrackFactory final
...
+    */
+   std::shared_ptr<WaveTrack> Create();
+
+   /**
+    * \brief Creates an empty WaveTrack with custom sample format and custom rate
+    * \param format Desired sample format
+    * \param rate Desired sample rate
+    * \return Orphaned WaveTrack
+    */
+   std::shared_ptr<WaveTrack> Create(sampleFormat format, double rate);
```
Paul-Licameli, line 638: I understand that you changed the factory to force you to fix compilation of all uses of it, and so consider at each place how to set the track name.

How can we guarantee that future calls to these functions will also have names set?

Should you at least comment here that the caller should consider setting a unique track name?

Paul-Licameli:
I'm done reviewing.

Please reconsider what happens to the audible track name for the screen reader.

Otherwise nothing is wrong with this, but I don't know from reading the issue, who specified or agreed to this change of behavior.  Please tell.

src/WaveTrack.h
vsverchinsky: This function doesn't set the name. The reason is that the track with random name isn't better than track without the name, when the intention is to have a track with a certain name. Another reason is that there are many cases when track is created to hold some temprorary data and then inserted into another track. I'll improve the comment by adding "an unnamed empty track".

src/import/ImportPlugin.cpp
vsverchinsky: >Should you consider giving them unique names?

If I understand your question correctly, then I think `ProjectFileManager::AddImportedTracks` assigns unique names to the channels.

>But in the usual case of one track, I understand that you would not want to add the numeral.

Yes.

src/TrackPanelAx.cpp
vsverchinsky: I've found multiple instances, of the same string in different places and with different comments (or without them), and didn't find an approriate one for this case, but missed that one: `/* i18n-hint: The %d is replaced by the number of the track.*/`
Thanks.

libraries/lib-track/Track.cpp
vsverchinsky: Yes, I need to improve my practice of self-reviewing)

src/TrackPanelAx.cpp
Paul-Licameli: If a translatable string literal is repeated exactly in more than one place, and without a context string, then it is enough the one occurrencde has a comment.

src/effects/nyquist/Nyquist.cpp
```diff
@@ -1530,8 +1530,10 @@ bool NyquistEffect::ProcessOne()
       unsigned int l;
       auto ltrack = * mOutputTracks->Any< LabelTrack >().begin();
       if (!ltrack) {
+         auto newTrack = std::make_shared<LabelTrack>();
+         newTrack->SetName(inputTracks()->MakeUniqueTrackName(LabelTrack::GetDefaultName()));
```
Paul-Licameli, line 1534: The repetition of making a track, then setting its default name with a unique number, is repeated in a few places.  I would suggest a static member function of LabelTrack to simplify this repetition.

src/LabelTrack.cpp
```diff
@@ -55,6 +55,11 @@ static ProjectFileIORegistry::ObjectReaderEntry readerEntry{
    LabelTrack::New
 };
 
+wxString LabelTrack::GetDefaultName()
+{
+   return _("Labels");
```
Paul-Licameli, line 60: Please see the struct Track::ClassTypeInfo and learn a bit about how it is used.

Maybe it should be extended with a field containing the default name for a track type, so there is a uniform way to specify it for Label and Wave and other types of tracks.

If for some reason there is need for polymorphism in getting "the default name for a track type" but unspecified which type that is.

But this generalization is for now a solution looking for a problem.  Don't do it, but think a little about it.

Paul-Licameli:
I approve the new changes that I see, with one suggestion for eliminating some repetition.

But I'm not sure what is decided about the right thing to do for the screen reader, or if that is meant to be treated in another issue.  I don't see a change in that part since the last discussion.

Was this supposed to be review ready?

src/effects/nyquist/Nyquist.cpp
vsverchinsky: Makes sense, thanks

vsverchinsky:
Eleminated code repetitions, and improved text for screen readers

DavidBailes:
> and improved text for screen readers

Thanks for removing the initial "Track" from the accessibility name - that's much better.
However, the accessibility name for label tracks needs improving, as it contains too much repetition, which can be very annoying.
Using 3.1.3 or earlier, the acc name of a label track with a default name would be something like "Track 4 Label track"
Using your branch the acc name of a newly created label track is something like "4 Labels1 Label track". In this case "Label track" does not need appending, as the name of the track makes it clear that it contains labels.
Using your branch, if a project is opened which was created with 3.1.3 or earlier, and there are existing label tracks, then the acc name of one of these label tracks is something like "4 Label track Label track". Again, "Label track" does not need appending.
So the appending of "Label track" should be conditional. Something like "Label track" is appended except in the cases where the track name contains the string "label" or the string "labels" (not sure if the second case has to be included, I'm not sure how some non-english languages handle plurals).

vsverchinsky:
>So the appending of "Label track" should be conditional. Something like "Label track" is appended except in the cases where the track name contains the string "label" or the string "labels" (not sure if the second case has to be included, I'm not sure how some non-english languages handle plurals).

@DavidBailes 
We can conditionally append "Label Track" to the name if it does not contain default name as its part.
"Labels 1" => "Labels 1"
"My Label" => "My Label Label Track"
"Labels Voice" => "Labels Voice"
Is it ok? Considering plural forms might overcomplicate things right now I believe.

Paul-Licameli:
> > Is it ok? Considering plural forms might overcomplicate things right now I believe.
> 
> It's just about OK, but it would be better if label tracks created using 3.1.3 and earlier were not read as something like "4 Label Track Label Track". I'm not sure why you think it will overcomplicate things. The additional rule could just be if the name is "Label Track", then "Label Track" is not appended.

To avoid overcomplication and implement something that works for all languages -- I would say scan for a singular string only, using the translation of the old internationalized strings formerly appended to the names if the migration from older Audacity versions matters.

This would miss "Labels".  There could be some special English only logic for that.  Doing plurals in every language would be hard.

src/DropoutDetector.cpp
```diff
@@ -24,16 +24,14 @@ static AudacityProject::AttachedObjects::RegisteredFactory sKey {
...
          auto &tracks = TrackList::Get( project );
-         tracks.Add( uTrack );
+
          /* i18n-hint:  A name given to a track, appearing as its menu button.
           The translation should be short or else it will not display well.
           At most, about 11 Latin characters.
           Dropout is a loss of a short sequence of audio sample data from the
           recording */
-         pTrack->SetName(_("Dropouts"));
+         auto pTrack = LabelTrack::Create(tracks, _("Dropouts"));
```
Paul-Licameli, line 34: The latest changes to simplify the repetitions lost the use of MakeUniqueTrackName on pahts like this where the argument is explicit.  Tell me you meant that, or else fix it.

Paul-Licameli (changes requested):
Lost unique naming?

src/LabelTrack.h
```diff
@@ -88,9 +88,23 @@ class AUDACITY_DLL_API LabelTrack final
...
+    * \brief Creates a new LabelTrack with specified @p name and appends it to the @p trackList
+    * \return New LabelTrack with custom name
+    */
+   static LabelTrack* Create(TrackList& trackList, const wxString& name);
+
+   /**
+    * \brief Create a new LabelTrack with default name and appends it to the @p trackList
+    * \return New LabelTrack with default name
+    */
+   static LabelTrack* Create(TrackList& trackList);
```
Paul-Licameli, line 106: Comment the two functions above about the intention to add unique names, unless that is caller's responsibility.  Or maybe add a boolean to control whether there is unique numbering.

src/TrackPanelAx.cpp
```diff
@@ -363,74 +361,61 @@ wxAccStatus TrackPanelAx::GetName( int childId, wxString* name )
...
+               this track is selected.*/
+            name->Append( wxT(" ") + wxString(_( " Selected" )) );
+         }
+         if (SyncLock::IsSyncLockSelected(t.get()))
+         {
+            /* i18n-hint: This is for screen reader software and indicates that
+               this track is shown with a sync-locked icon.*/
+            // The absence of a dash between Sync and Locked is deliberate -
+            // if present, Jaws reads it as "dash".
+            name->Append( wxT(" ") + wxString(_( " Sync Locked" )) );
```
Paul-Licameli, line 418: Nothing wrong with these latest changes, but I understand there is still debate and it's not yet final.

If a virtual function can be added to Track so that the dynamic_casts and #include of Note, Wave, and Label tracks can go way, that would be nice.

Eliminating WaveTrack.h is not as important -- I don't plan to segregate WaveTrack from the Audacity monolith in an add-on.  WaveTracks are just too much involved with everything else.

src/LabelTrack.h
vsverchinsky: Only one with default name assigns unique name to the track

src/DropoutDetector.cpp
vsverchinsky: Thanks

DavidBailes:
The accessibility name of a label track is now fine - for both current and previous versions of the default track name, "Label Track" is not appended. Thanks.

src/LabelTrack.h
```diff
@@ -88,9 +88,23 @@ class AUDACITY_DLL_API LabelTrack final
    , public wxEvtHandler
 {
  public:
+   static wxString GetDefaultName();
+
    // Construct and also build all attachments
    static LabelTrack *New(AudacityProject &project);
 
+   /**
+    * \brief Create a new LabelTrack with specified @p name and append it to the @p trackList
```
Paul-Licameli, line 97: "appends"

src/LabelTrack.h
```diff
@@ -88,9 +88,23 @@ class AUDACITY_DLL_API LabelTrack final
...
+    * \return New LabelTrack with custom name
+    */
+   static LabelTrack* Create(TrackList& trackList, const wxString& name);
+
+   /**
+    * \brief Create a new LabelTrack with unique default name and append it to the @p trackList
```
Paul-Licameli, line 103: "appends"

Paul-Licameli (approved):

Issue #2475, opened 2022-01-25
Add sidebar and make sure it shows and hides: effects


AnitaBats:
@Higherself1133 could you please populate the description?

thirdxlabs:
Here's the structure for the Realtime Effects panel. When there are no effects added, + graphic and onboarding text would be there. The onboarding graphic would be updated soon (some UI polishing work needs to be there)

Panel opens upon clicking the button "Add Effects" on the audio track. User can close the panel by clicking the x button on top right.


<img width="853" alt="Screen Shot 2022-01-28 at 2 10 45 PM" src="https://user-images.githubusercontent.com/72258608/151552558-72535795-c5e5-469e-bc0d-82146262ccef.png">

For now, the proposal is to use the key 'E' to open and close the realtime effects panel.

Higherself1133:
Here's the structure for the Realtime Effects panel. When there are no effects added, + graphic and onboarding text would be there. The onboarding graphic would be updated soon (some UI polishing work needs to be there).

<img width="734" alt="Screen Shot 2022-01-28 at 2 03 23 PM" src="https://user-images.githubusercontent.com/97682925/151552975-dda9eb66-9154-40e1-88cc-81a11329e37d.png">


Panel opens upon clicking the button "Add Effects" on the audio track. User can close the panel by clicking the x button on top right.

<img width="853" alt="Screen Shot 2022-01-28 at 2 10 45 PM" src="https://user-images.githubusercontent.com/97682925/151553020-3acbb076-ff75-4055-8ade-6ceb95eed934.png">


For now, the proposal is to use the Keyboard shortcut 'E' to open and close the realtime effects panel.

PR #2565, opened 2022-02-17
Effects side panel

vsverchinsky:
Resolves: #2475
Resolves: #2474

Adds an effects panel container, and applies neccessary project window window structure changes, and "Effects" button to the wave tracks that opens panel. There is nothing yet in the panel itself, the example effects UI prototype, which uses that pannel added can be found here: https://github.com/vsverchinsky/audacity/tree/i2474-effects-button

Recommended:

### DIFF
--- a/libraries/lib-theme/AColor.cpp
+++ b/libraries/lib-theme/AColor.cpp
@@ -76,6 +76,84 @@ wxBrush AColor::spareBrush;
 wxPen AColor::uglyPen;
 wxBrush AColor::uglyBrush;
 
+namespace
+{
+   //Simplified variation of nine patch scale drawing
+   //https://en.wikipedia.org/wiki/9-slice_scaling
+   //Bitmap and rect expected to have at least 3px in both directions
+   void DrawNinePatch(wxDC& dc, wxBitmap& bitmap, const wxRect& r)
+   {
+      wxMemoryDC memDC;
+      memDC.SelectObject(bitmap);
+
+      // image slices
+
+      const auto uw0 = bitmap.GetWidth() / 2;
+      const auto uw1 = 1;
+      const auto uw2 = bitmap.GetWidth() - uw0 - uw1;
+
+      const auto vh0 = bitmap.GetHeight() / 2;
+      const auto vh1 = 1;
+      const auto vh2 = bitmap.GetHeight() - vh1 - vh0;
+
+      const auto u0 = 0;
+      const auto u1 = uw0;
+      const auto u2 = uw0 + uw1;
+
+      const auto v0 = 0;
+      const auto v1 = vh0;
+      const auto v2 = vh0 + vh1;
+
+      //Button geometry
+
+      const auto xw0 = std::min(uw0, r.width / 2);
+      const auto xw2 = std::min(uw2, r.width / 2);
+      const auto xw1 = r.width - xw0 - xw2;
+
+      const auto yh0 = std::min(vh0, r.height / 2);
+      const auto yh2 = std::min(vh2, r.height / 2);
+      const auto yh1 = r.height - yh0 - yh2;
+
+      const auto x0 = r.x;
+      const auto x1 = r.x + xw0;
+      const auto x2 = r.x + xw0 + xw1;
+
+      const auto y0 = r.y;
+      const auto y1 = r.y + yh0;
+      const auto y2 = r.y + yh0 + yh1;
+
+      dc.StretchBlit(x0, y0, xw0, yh0, &memDC, u0, v0, uw0, vh0, wxCOPY, true);
+      dc.StretchBlit(x1, y0, xw1, yh0, &memDC, u1, v0, uw1, vh0, wxCOPY, true);
+      dc.StretchBlit(x2, y0, xw2, yh0, &memDC, u2, v0, uw2, vh0, wxCOPY, true);
+
+      dc.StretchBlit(x0, y1, xw0, yh1, &memDC, u0, v1, uw0, vh1, wxCOPY, true);
+      dc.StretchBlit(x1, y1, xw1, yh1, &memDC, u1, v1, uw1, vh1, wxCOPY, true);
+      dc.StretchBlit(x2, y1, xw2, yh1, &memDC, u2, v1, uw2, vh1, wxCOPY, true);
+
+      dc.StretchBlit(x0, y2, xw0, yh2, &memDC, u0, v2, uw0, vh2, wxCOPY, true);
+      dc.StretchBlit(x1, y2, xw1, yh2, &memDC, u1, v2, uw1, vh2, wxCOPY, true);
+      dc.StretchBlit(x2, y2, xw2, yh2, &memDC, u2, v2, uw2, vh2, wxCOPY, true);
+   }
+
+   int GetButtonImageIndex(bool up, bool selected, bool highlight)
+   {
+      // There are eight button states in the TCP.
+      // A theme might not differentiate among them all.  That's up to 
+      // the theme designer.
+      //   Button highlighted (i.e. hovered over) or not.
+      //   Track selected or not
+      //   Button up or down.
+      // Highlight in most themes is lighter than not highlighted.
+      if ( highlight && selected)
+         return up ? bmpHiliteUpButtonExpandSel : bmpHiliteButtonExpandSel;
+      if ( highlight )
+         return up ? bmpHiliteUpButtonExpand : bmpHiliteButtonExpand;
+      if( selected )
+         return up ? bmpUpButtonExpandSel : bmpDownButtonExpandSel;
+      return up ? bmpUpButtonExpand : bmpDownButtonExpand;
+   }
+}
+
 //
 // Draw an upward or downward pointing arrow.
 //
@@ -204,31 +282,22 @@ void AColor::Bevel(wxDC & dc, bool up, const wxRect & r)
    AColor::Line(dc, r.x, r.y + r.height, r.x + r.width, r.y + r.height);
 }
 
-void AColor::Bevel2
-(wxDC & dc, bool up, const wxRect & r, bool bSel, bool bHighlight)
+void AColor::ButtonStretch(wxDC& dc, bool up, const wxRect& r, bool selected, bool highlight)
 {
-   int index = 0;
-   // There are eight button states in the TCP.
-   // A theme might not differentiate between them all.  That's up to 
-   // the theme designer.
-   //   Button highlighted (i.e. hovered over) or not.
-   //   Track selected or not
-   //   Button up or down.
-   // Highlight in most themes is lighter than not highlighted.
-   if ( bHighlight && bSel)
-      index = up ? bmpHiliteUpButtonExpandSel : bmpHiliteButtonExpandSel;
-   else if ( bHighlight )
-      index = up ? bmpHiliteUpButtonExpand : bmpHiliteButtonExpand;
-   else if( bSel )
-      index = up ? bmpUpButtonExpandSel : bmpDownButtonExpandSel;
-   else
-      index = up ? bmpUpButtonExpand : bmpDownButtonExpand;
+   DrawNinePatch(
+      dc,
+      theTheme.Bitmap(GetButtonImageIndex(up, selected, highlight)),
+      r
+   );
+}
 
-   wxBitmap & Bmp = theTheme.Bitmap( index );
+void AColor::Bevel2(wxDC & dc, bool up, const wxRect & r, bool bSel, bool bHighlight)
+{
+   auto& Bmp = theTheme.Bitmap( GetButtonImageIndex(up, bSel, bHighlight) );
    wxMemoryDC memDC;
    memDC.SelectObject(Bmp);
-   int h = wxMin( r.height, Bmp.GetHeight() );
 
+   int h = std::min(r.height, Bmp.GetHeight());
 
    dc.Blit( r.x,r.y,r.width/2, h, &memDC, 0, 0, wxCOPY, true );
    int r2 = r.width - r.width/2;
--- a/src/ProjectManager.cpp
+++ b/src/ProjectManager.cpp
@@ -183,10 +183,6 @@ void ProjectManager::SaveWindowSize()
    sbWindowRectAlreadySaved = true;
 }
 
-#ifdef EXPERIMENTAL_NOTEBOOK
-   extern void AddPages(   AudacityProject * pProj, GuiFactory & Factory,  wxNotebook  * pNotebook );
-#endif
-
 void InitProjectWindow( ProjectWindow &window )
 {
    auto &project = window.GetProject();
@@ -249,14 +245,13 @@ void InitProjectWindow( ProjectWindow &window )
    // irrespective of the order in which they were created.
    ToolManager::Get(project).GetTopDock()->MoveBeforeInTabOrder(&ruler);
 
-   const auto pPage = window.GetMainPage();
 
    wxBoxSizer *bs;
    {
       auto ubs = std::make_unique<wxBoxSizer>(wxVERTICAL);
       bs = ubs.get();
       bs->Add(topPanel, 0, wxEXPAND | wxALIGN_TOP);
-      bs->Add(pPage, 1, wxEXPAND);
+      bs->Add(window.GetContainerWindow(), 1, wxEXPAND);
       bs->Add( ToolManager::Get( project ).GetBotDock(), 0, wxEXPAND );
       window.SetAutoLayout(true);
       window.SetSizer(ubs.release());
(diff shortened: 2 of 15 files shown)


src/ProjectWindow.h
```diff
@@ -52,9 +53,28 @@ class AUDACITY_DLL_API ProjectWindow final : public ProjectWindowBase
    bool IsBeingDeleted() const { return mIsDeleting; }
    void SetIsBeingDeleted() { mIsDeleting = true; }
 
-   wxWindow *GetMainPage() { return mMainPage; }
-   wxPanel *GetMainPanel() { return mMainPanel; }
-   wxPanel *GetTopPanel() { return mTopPanel; }
+   /**
+    * \brief Effect window contains list off effects assigned to
+    * a selected track.
+    * \return Pointer to a effects side-panel window (not null)
```
Paul-Licameli, line 59: "an"

src/ProjectWindow.h
```diff
@@ -52,9 +53,28 @@ class AUDACITY_DLL_API ProjectWindow final : public ProjectWindowBase
...
+   wxWindow* GetEffectsWindow() noexcept;
+   /**
+    * \brief Track list window is the parent container for TrackPanel
+    * \return Pointer to a track list window (not null)
+    */
+   wxWindow* GetTrackListWindow() noexcept;
+   /**
+    * \brief Container is a parent window for both effects panel and
+    * track list windows
+    * \return Pointer to a track list window (not null)
```
Paul-Licameli, line 70: Wrong comment

src/TrackPanel.cpp
```diff
@@ -350,18 +350,13 @@ AudacityProject * TrackPanel::GetProject() const
...
-   wxASSERT( pWind );
-   pWind = pWind->GetParent(); //Notebook
-   wxASSERT( pWind );
-#endif
-   pWind = pWind->GetParent(); //MainPanel
-   wxASSERT( pWind );
-   pWind = pWind->GetParent(); //ProjectWindow
-   wxASSERT( pWind );
-   return &static_cast<ProjectWindow*>( pWind )->GetProject();
+   auto pWind=(wxWindow*)pConstWind;
```
Paul-Licameli, line 353: Maybe define pWind, then assign it to pConstWind, and get rid of the ugly C-style pointer cast

src/TrackPanel.cpp
```diff
@@ -350,18 +350,13 @@ AudacityProject * TrackPanel::GetProject() const
    //Do it in two stages in case 'this' is not a wxWindow.
    //when the compiler will flag the error.
    wxWindow const * const pConstWind = this;
-   wxWindow * pWind=(wxWindow*)pConstWind;
-#ifdef EXPERIMENTAL_NOTEBOOK
```
Paul-Licameli, line 354: Let's get rid of the rest of EXPERIMENTAL_NOTEBOOK in the source tree.  That includes the line in Experimental.cmake.  The rest of the source code isn't available to turn it on again even if we wanted to.

src/TrackPanel.cpp
```diff
@@ -350,18 +350,13 @@ AudacityProject * TrackPanel::GetProject() const
...
-   pWind = pWind->GetParent(); //Notebook
-   wxASSERT( pWind );
-#endif
-   pWind = pWind->GetParent(); //MainPanel
-   wxASSERT( pWind );
-   pWind = pWind->GetParent(); //ProjectWindow
-   wxASSERT( pWind );
-   return &static_cast<ProjectWindow*>( pWind )->GetProject();
+   auto pWind=(wxWindow*)pConstWind;
+   while(pWind = pWind->GetParent())
```
Paul-Licameli, line 354: Put extra parentheses around this condition to suppress warnings (from clang at least).  Or rewrite as a for loop.

src/ProjectManager.cpp
```diff
@@ -249,14 +249,13 @@ void InitProjectWindow( ProjectWindow &window )
...
 
-   const auto pPage = window.GetMainPage();
 
    wxBoxSizer *bs;
    {
       auto ubs = std::make_unique<wxBoxSizer>(wxVERTICAL);
       bs = ubs.get();
       bs->Add(topPanel, 0, wxEXPAND | wxALIGN_TOP);
-      bs->Add(pPage, 1, wxEXPAND);
+      bs->Add(window.GetContainerWindow(), 1, wxEXPAND);
```
Paul-Licameli, line 258: Notice that the bottom toolbar docking area is part of the right side of the container window, therefore it will be narrower when the effects panel shows, rather than limiting the vertical extent of that panel.

Not right or wrong -- just saying we understand the consequence.

src/ProjectManager.cpp
```diff
@@ -270,9 +269,9 @@ void InitProjectWindow( ProjectWindow &window )
    //      will be given the focus even if we try to SetFocus().  By
    //      making the TrackPanel that first window, we resolve several
    //      keyboard focus problems.
-   pPage->MoveBeforeInTabOrder(topPanel);
+   window.GetContainerWindow()->MoveAfterInTabOrder(topPanel);
 
-   bs = (wxBoxSizer *)pPage->GetSizer();
+   bs = (wxBoxSizer *)window.GetTrackListWindow()->GetSizer();
```
Paul-Licameli, line 274: Clean this up as a modern `static_cast` while you are visiting the neighborhood.

src/ProjectManager.cpp
```diff
@@ -270,9 +269,9 @@ void InitProjectWindow( ProjectWindow &window )
    //      will be given the focus even if we try to SetFocus().  By
    //      making the TrackPanel that first window, we resolve several
    //      keyboard focus problems.
-   pPage->MoveBeforeInTabOrder(topPanel);
+   window.GetContainerWindow()->MoveAfterInTabOrder(topPanel);
```
Paul-Licameli, line 272: Tab order reminds me that we may need to check out how tabbing now works in the presence of the new window.  @DavidBailes might have opinions.

src/ProjectManager.cpp
```diff
@@ -308,16 +307,8 @@ void InitProjectWindow( ProjectWindow &window )
...
-
-#ifdef EXPERIMENTAL_NOTEBOOK
-   AddPages(this, Factory, pNotebook);
-#endif
-
-   auto mainPanel = window.GetMainPanel();
-
-   mainPanel->Layout();
+   window.GetTrackListWindow()->SetAutoLayout(true);
+   window.GetTrackListWindow()->Layout();
```
Paul-Licameli, line 311: It would look a little nicer to me, to use a variable to eliminate the repeated calls to `window.GetTrackListWindow()` and `window.GetContainerWindow()`

src/ProjectWindow.cpp
```diff
@@ -39,6 +39,7 @@ Paul Licameli split from AudacityProject.cpp
 #include <wx/display.h>
 #include <wx/scrolbar.h>
 #include <wx/sizer.h>
+#include <wx/splitter.h>
```
Paul-Licameli, line 42: The only new #include in this PR, and it's a wxWidgets header.  No new dependencies among Audacity's own source code files.  That is good.

src/ProjectWindow.cpp
```diff
@@ -595,45 +596,35 @@ ProjectWindow::ProjectWindow(wxWindow * parent, wxWindowID id,
...
-   wxWindow    * pPage;
+   auto container = safenew wxSplitterWindow(this, wxID_ANY,
+      wxDefaultPosition,
+      wxDefaultSize,
+      wxNO_BORDER | wxSP_LIVE_UPDATE);
+   container->Bind(wxEVT_SPLITTER_DOUBLECLICKED, [](wxSplitterEvent& event){
+      //"The default behaviour is to unsplit the window"
+      event.Veto();//do noting instead
+   });
+   container->SetBackgroundColour(theTheme.Colour(clrLight));
```
Paul-Licameli, line 607: What happens if the user changes the theme?  Do you need to add to `ProjectWindow::OnThemeChange` to reset this color?  Or, is the background color in fact not seen because everything else in the window hides it?

src/ProjectWindow.cpp
```diff
@@ -595,45 +596,35 @@ ProjectWindow::ProjectWindow(wxWindow * parent, wxWindowID id,
...
-   //pPage->SetBackgroundColour( theTheme.Colour( clrDark ));
-#endif
-   pPage->SetLayoutDirection(wxLayout_LeftToRight);
+   mTrackListWindow->SetSizer( safenew wxBoxSizer(wxVERTICAL) );
+   mTrackListWindow->SetLabel("Main Panel");// Not localized.
+   mTrackListWindow->SetLayoutDirection(wxLayout_LeftToRight);
+
+   mEffectsWindow = safenew wxWindow(mContainerWindow, wxID_ANY, wxDefaultPosition, wxSize(250, 20));
+   mEffectsWindow->Hide();
+   mEffectsWindow->SetBackgroundColour(theTheme.Colour(clrMedium));
```
Paul-Licameli, line 620: Another background color you may need to update with theme changes

src/ProjectWindow.cpp
```diff
@@ -1839,6 +1852,27 @@ void ProjectWindow::DoZoomFit()
...
+      auto effectsWindowSize = mEffectsWindow->GetSize();
+      mEffectsWindow->Show();
+      mContainerWindow->SplitVertically(mEffectsWindow, mTrackListWindow, effectsWindowSize.GetWidth());
+   }
+}
+
+void ProjectWindow::HideEffectsPanel()
+{
+   mEffectsWindow->Hide();
+   mContainerWindow->Unsplit(mEffectsWindow);
```
Paul-Licameli, line 1873: Is Unsplitting enough without the call to Hide?

src/ProjectWindow.cpp
```diff
@@ -1839,6 +1852,27 @@ void ProjectWindow::DoZoomFit()
...
+
+   if(track == nullptr)
+      return;
+
+   if(!mEffectsWindow->IsShown())
+   {
+      //Restore previous effects window size
+      auto effectsWindowSize = mEffectsWindow->GetSize();
+      mEffectsWindow->Show();
+      mContainerWindow->SplitVertically(mEffectsWindow, mTrackListWindow, effectsWindowSize.GetWidth());
```
Paul-Licameli, line 1866: Is the explicit Show() needed?

src/ProjectWindow.cpp
Paul-Licameli: What happens if you try to drag the separator?  Is that not wanted, and do you need to Bind and Veto another event?

Paul-Licameli:
I'm done with a round of review.  I await your answers.

Paul-Licameli:
See this, about difference between `SetBackgroundColour` and `SetOwnBackgroundColour` https://docs.wxwidgets.org/trunk/classwx_window.html#a37219df52734626e23401fd83b25d8a0

I build with modification to show the effecs panel, and I can confirm that the background color updates correctly with theme change.

The separator is draggable.  There is a limiting left position, but it can be dragged right as far as the edge of the window, crowding out the track panel.

src/ProjectWindow.h
vsverchinsky: Thanks, will fix

src/ProjectWindow.h
vsverchinsky: Thanks, will fix

src/TrackPanel.cpp
vsverchinsky: Makes sense, will fix

src/TrackPanel.cpp
vsverchinsky: Will do

src/ProjectManager.cpp
vsverchinsky: Not sure that understand that, bottom toolbar docking area is a sibbling of the container window, isn't it?

src/ProjectManager.cpp
vsverchinsky: Ok

src/ProjectManager.cpp
vsverchinsky: Ok

src/ProjectWindow.cpp
vsverchinsky: >What happens if the user changes the theme? Do you need to add to ProjectWindow::OnThemeChange to reset this color? Or, is the background color in fact not seen because everything else in the window hides it?

In my prototype branch I had changed background color in response to the theme event, but there was no visible effect, because of the child windows. I mistakenly committed this line.

>What happens if you try to drag the separator? Is that not wanted, and do you need to Bind and Veto another event?

If I understand you correctly you saying that we don't need the separator to be draggable?

src/ProjectWindow.cpp
vsverchinsky: Turned out - yes, thanks

src/ProjectWindow.cpp
Paul-Licameli: I ask whether it is intended.

maybe a limit on rightward dragging is needed.

src/ProjectManager.cpp
Paul-Licameli: You are right, I misread that

Paul-Licameli:
I informally approve the changes.  Do you wish to merge this PR now (I would recommend it) or wait to complete it with a button for showing and hiding the sidebar?  If you wait, then change the top comment to un-link it from the issue, which it does not completely resolve.

vsverchinsky:
@Paul-Licameli I've added 3 new commits with "Effects" track panel button.

src/tracks/playabletrack/ui/PlayableTrackControls.cpp
```diff
@@ -98,6 +110,23 @@ void MuteOrSoloDrawFunction
...
 
+void EffectsDrawFunction
+( wxDC *dc, const wxRect &bev, const Track *pTrack, bool down, 
+  bool sel, bool hit )
+{   
+   wxCoord textWidth, textHeight;
+   wxString str = XO("Effects").Translation();
+
+   bool selected = pTrack ? pTrack->GetSelected() : true;
+   auto pt = dynamic_cast<const PlayableTrack *>(pTrack);
```
Paul-Licameli, line 121: unused variable

libraries/lib-theme/AColor.cpp
```diff
@@ -192,43 +270,39 @@ void AColor::Bevel(wxDC & dc, bool up, const wxRect & r)
...
 
-   wxBitmap & Bmp = theTheme.Bitmap( index );
+void AColor::Bevel2(wxDC & dc, bool up, const wxRect & r, bool bSel, bool bHighlight)
+{
+   auto& Bmp = theTheme.Bitmap( GetButtonImageIndex(up, bSel, bHighlight) );
    wxMemoryDC memDC;
    memDC.SelectObject(Bmp);
-   int h = wxMin( r.height, Bmp.GetHeight() );
 
+   int h = wxMin( r.height, Bmp.GetHeight() );
```
Paul-Licameli, line 305: Feel free to clean up wxMin, wxMax with std::min, std::max instead

libraries/lib-theme/AColor.cpp
```diff
@@ -76,6 +76,84 @@ wxBrush AColor::spareBrush;
...
+
+      dc.StretchBlit(x0, y2, xw0, yh2, &memDC, u0, v2, uw0, vh2, wxCOPY, true);
+      dc.StretchBlit(x1, y2, xw1, yh2, &memDC, u1, v2, uw1, vh2, wxCOPY, true);
+      dc.StretchBlit(x2, y2, xw2, yh2, &memDC, u2, v2, uw2, vh2, wxCOPY, true);
+   }
+
+   int GetButtonImageIndex(bool up, bool selected, bool highlight)
+   {
+      // There are eight button states in the TCP.
+      // A theme might not differentiate between them all.  That's up to
```
Paul-Licameli, line 141: "among" instead of "between" is better English

Paul-Licameli: I know you just moved this comment, but fix it

libraries/lib-theme/AColor.cpp
```diff
@@ -192,43 +270,39 @@ void AColor::Bevel(wxDC & dc, bool up, const wxRect & r)
...
 
-   AColor::Line(dc, r.x, r.y, r.x + r.width, r.y);
-   AColor::Line(dc, r.x, r.y, r.x, r.y + r.height);
+   const auto left = r.x;
+   const auto right = r.x + r.width - 1;
+   const auto top = r.y;
+   const auto bottom = r.y + r.height - 1;
+
+   AColor::Line(dc, left, top, right, top);
+   AColor::Line(dc, left, top, left, bottom);
```
Paul-Licameli, line 279: Did you discover off by one mistakes in this function that you are also fixing?  A commit comment should say so.

Should you fix it?  Is it making small but noticeable changes in other drawing that wasnt' specified?

Paul-Licameli (changes requested):
Are there small, unintended changes in appearance elsewhere?  This is changing one function more widely used than just in the track control panel.

libraries/lib-theme/AColor.cpp
Paul-Licameli: You add no calls to Bevel.  Is it used in drawing functions that were changed?

Rather than change the more widely used function, adjust the rectangle that you pass into the function in the one relevant place.

Paul-Licameli: But I think there are no such uses that you need to adjust.

Paul-Licameli (approved):

Issue #2090, opened 2021-11-05
Misbehavior of Sound Activated Recording (bad inter-thread communication for pausing)


Paul-Licameli:
**Describe the bug**
My concern is to clean up improper inter-thread communication in playback and recording.

Review of the code discovered one possible symptom, described here.  This might not be considered high priority from QA's point of view, but it is one way I found to demonstrate bad behavior.

**To Reproduce**
1. In Device preferences, set latency correction (which defaults to a negative value) to zero, or closer to zero (lesser absolute value).
2. In Recording preferences, enable Sound Activation, with a high threshold (0 dB).
3. Press Record, wait, press Stop, and don't make any loud noises.
4. A small length of sound is recorded.

**Expected behavior**
Nothing is recorded, if you don't make any sound above the threshold.  (If you clap your hands, you would expect to see recording start and stop again.  Or with more reasonable threshold, it might record only when you speak and not when you pause.)

**Implementation notes**
The low-latency thread calls AudioIOListener::OnSoundActivationThreshold, which is overridden in ProjectAudioManager, and that calls CallAfter, entering critical sections and allocating some free store -- bad things to do on this thread!

Then the main thread, after some delay, calls ProjectAudioManager::Pause() which eventually calls AudioIOBase::SetPaused(), and finally, that causes later calls to AudioIoCallback::AudioCallback on the low latency thread to discard input.  This timing window allows the bug to happen as described, though it could be hidden at the usual default latency correction setting in preferences, which causes a third thread (TrackBufferExchange) to do its own discarding of some initial sample data.

Instead, the callback OnSoundActivationThreshold should be eliminated, AudioIoCallback should set the paused state directly in response to sound activation, and main thread code should poll for this change of state in idle time, as might be needed to update button appearances.

PR #2567, opened 2022-02-17
Issue with sound activated recording

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2090

*see original issue for motivations*

Recommended:

### DIFF
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -71,9 +71,14 @@ ProjectAudioManager::ProjectAudioManager( AudacityProject &project )
       registerStatusWidthFunction{ StatusWidthFunction };
    project.Bind( EVT_CHECKPOINT_FAILURE,
       &ProjectAudioManager::OnCheckpointFailure, this );
+
+   wxTheApp->Bind(wxEVT_IDLE, &ProjectAudioManager::OnIdle, this);
 }
 
-ProjectAudioManager::~ProjectAudioManager() = default;
+ProjectAudioManager::~ProjectAudioManager()
+{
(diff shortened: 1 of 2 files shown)


pietro68:
After these changes, I tried the steps in the linked issue and I can not reproduce the bug anymore.
Normal sound-activated recording (i.e. with a normal threshold) still seems to work fine, so previous functionality is preserved.

src/toolbars/ControlToolBar.cpp
```diff
@@ -615,7 +615,10 @@ void ControlToolBar::OnIdle(wxIdleEvent & event)
       return;
 
    auto &projectAudioManager = ProjectAudioManager::Get( mProject );
-   if ( projectAudioManager.Paused() )
+
+   bool isPaused = AudioIO::Get()->IsPaused();
+
+   if ( isPaused )
```
Paul-Licameli, line 621: This fix is not conservative enough.  The state changes in ProjectAudioManager no longer happen.  All the checks in that function do not happen.

I say, do not change ControlTooBar.  Rather, change ProjectAudioManager.  Make it inherit wxEvtHandler.  Then let it receive idle events from the application, similar to this other line in SpectralDataDialog.cpp:

```
   wxTheApp->Bind(wxEVT_IDLE, &SpectralDataDialogWorker::OnIdle, this);
```

Then let the event handler do all that `OnSoundActivationThreshold` did.  (Except delaying the call to Pause() using CallAfter will not be needed, as you will surely already be on the main thread and in the idle time of the event loop.)

Paul-Licameli (changes requested):

pietro68:
Hopefully what you will see in my last commit is what you meant in your suggestion; the problem still persist though (see commit message for details).
I will now try something different on top of this commit.

Issue #2580, opened 2022-02-21
TypedAny utility class template


Paul-Licameli:
Define a class template generating wrappers of std::any, making type distinctions.

Will be used by issue #2538, and may be re-used in the future.

Helps to solve the Ubuntu build errors that were blocking the pull request #2537 for that issue.

Issue #2538, opened 2022-02-09
Interface for externalized effect settings


Paul-Licameli:
The present system has only one mutable and global object for each effect, storing settings. This system should be phased out.

EffectDefinitionInterface will have alternative const methods using externalized settings objects: factory, save, load (and default implementations that just call-through to old functions).

EffectManager will store a settings object with each effect, rewritten in terms of the new interface.

Paul-Licameli:
The present system has only one mutable and global object for each effect, storing settings.  This system should be phased out.

EffectDefinitionInterface will have alternative const methods using externalized settings objects: factory, save, load (and default implementations that just call-through to old functions).

EffectManager will store a settings object with each effect, rewritten in terms of the new interface.

PR #2581, opened 2022-02-21
Define utility class template TypedAny: see code comments

Paul-Licameli:
Resolves: #2580

Define a class template generating wrappers of std::any, making type distinctions.

Will be used by issue https://github.com/audacity/audacity/issues/2538, and may be re-used in the future.

Helps to solve the Ubuntu build errors that were blocking the pull request https://github.com/audacity/audacity/pull/2537 for that issue.


crsib (approved):

pietro68 (approved):
I see now the purpose of this wrapper.

Issue #2577, opened 2022-02-21
EffectSettings passed into destructive effect processing routines too


Paul-Licameli:
This is needed so that, on either the destructive or nondestructive paths, the various real-time effects can each have a uniform implementation of transformation of sample streams block by block.

PR #2582, opened 2022-02-22
Settings in destructive effects

Paul-Licameli:
Resolves: #2577

The last batch of changes passed around EffectSettings into enough places to implement the inter-thread communication of
changing settings.  This batch passes EffectSettings into yet more places, including those on the path of destructive
effect processing (which does not have the same inter-thread problems), and to low levels of block processing that are
common to destructive and real-time processing.

Still nothing is ever in EffectSettings but a pointer to the EffectDefinitionInterfaceEx object.

Also still to do is break inheritance of EffectProcessor from EffectDefinitionInterface, and instead let EffectDefinitionInterface be a factory of EffectProcessor objects.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -353,14 +353,14 @@ public:
    virtual size_t ProcessBlock(
       const float *const *inBlock, float *const *outBlock, size_t blockLen) = 0;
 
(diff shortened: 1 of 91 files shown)


pietro68 (approved):
I can see that the theme of this PR is to pass Settings in all places where they are needed - so all good for me.

vsverchinsky (approved):

PR #2585, opened 2022-02-23
Externalize effect event handling

Paul-Licameli:
Resolves: #2584

Recommended:


libraries/lib-components/EffectInterface.cpp
```diff
@@ -160,4 +160,21 @@ EffectDefinitionInterfaceEx::FindMe(const Settings &settings) const
...
+EffectUIValidator::~EffectUIValidator() = default;
+
+DefaultEffectUIValidator::DefaultEffectUIValidator(
+   EffectUIClientInterface &effect)
+   : mEffect{effect}
+{}
+
+DefaultEffectUIValidator::~DefaultEffectUIValidator()
+{
+   mEffect.CloseUI();
```
pietro68, line 172: I am guessing that the purpose of this class is to have something which we can query to check whether all numeric values in the text fields of the UI of an effect are in the valid range (I count only text fields - because I assume that the possible range of motion of knobs and sliders is already bounded to a valid range).

What I'd not expect here, is that when this validation object gets destroyed, the User Interface is closed too (as in, the dialog will disappear). Are not these two things unrelated?

libraries/lib-components/EffectInterface.h
```diff
@@ -433,17 +433,29 @@ class COMPONENTS_API EffectUIClientInterface /* not final */
...
    //! Adds controls to a panel that is given as the parent window of `S`
-   virtual bool PopulateUI(ShuttleGui &S, EffectSettingsAccess &access) = 0;
+   /*!
+    @param S interface for adding controls to a panel in a dialog
+    @param access guaranteed to have a lifetime containing that of the returned
+    object
 
-   virtual bool ValidateUI() = 0;
-   virtual bool CloseUI() = 0;
+    @return null for failure; else an object invoked to retrieve values of UI
```
pietro68, line 441: I do not understand this comment: "return... an object invoked to retrieve values of UI controls" 
I look at the definition of EffectUIValidator and all I see there is `virtual bool Validate() = 0;` - is there a plan to extend that interface later on?

libraries/lib-components/EffectInterface.cpp
Paul-Licameli: You are correct about the purpose of the sole virtual function of the class.

See the existing overrides of EffectUIClientInterface::CloseUI() which still had to be called somewhere after this reorganization.  It does not cause the dialog to disappear -- rather, it happens during `EffectUIHost::OnClose` which is preparing the dialog for destruction.

It is a place where certain cleanup tasks must happen (like popping the event handler on a window), complementary to the setup tasks that are done in `EffectUIClientInterface::PopulateUI` which is the factory for this object.

The object exists to do more than satisfy the virtual function interface:  it also sets itself up to handle user interface events, and update the EffectSettings object that is accessible through the `access` argument that was given to `PopulateUI`.

libraries/lib-components/EffectInterface.h
Paul-Licameli: When I said "retrieve" I mean the object is expected to respond to the `Validate` message by examining all controls and updating its state.  I did not mean that it returns anything to the caller besides the success code.  I could clarify the comments.

libraries/lib-components/CMakeLists.txt
```diff
@@ -6,9 +6,9 @@ for storing as a configuration file value) with a user-visible translatable
 string.  It serves to "name" various things.
 
 CommandParameters, for write and reading key-value pairs to and from strings.
-]]#
+]]
```
Paul-Licameli, line 9: foobar

Paul-Licameli: haha

pietro68 (approved):
I think I am starting to understand RealtimeEffectState::AccessState so I can close this now.

PR #2594, opened 2022-02-25
I2090 second attempt

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2090

This solution consists of running the bare minimum set of code to fix the bug, while preserving existing behavior.

Whenever the sound-activation threshold is crossed (in either direction), all things that must be done in response to that are now all done in the paThread, without deferring anything to the main thread. So far this seems to be the only way to fix the issue - and what is executed in the paThread in response to the crossing is just read/writes of bool variables, without allocations or locks etc.

Changes are presented in a set of commits, to show step-by-step how existing functionality/state-changes are preserved, despite the final code being simpler than before; actually, even the first commit is enough to solve the issue, but I wanted to go on and simplify the code as much as possible.
In case of approval then I would squash commits together.

Recommended:

### DIFF
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -535,7 +535,7 @@ void ProjectAudioManager::Stop(bool stopStream /* = true*/)
       gAudioIO->AILADisable();
    #endif
 
-   projectAudioManager.SetPaused( false );
+   projectAudioManager.SetPausedOff();
    //Make sure you tell gAudioIO to unpause
    gAudioIO->SetPaused( false );
 
@@ -559,19 +559,6 @@ void ProjectAudioManager::Stop(bool stopStream /* = true*/)
       toolbar->EnableDisableButtons();
 }
 
-void ProjectAudioManager::Pause()
-{
-   auto &projectAudioManager = *this;
-   bool canStop = projectAudioManager.CanStopAudioStream();
-
-   if ( !canStop ) {
-      auto gAudioIO = AudioIO::Get();
-      gAudioIO->SetPaused(!gAudioIO->IsPaused());
-   }
-   else {
-      OnPause();
-   }
-}
 
 WaveTrackArray ProjectAudioManager::ChooseExistingRecordingTracks(
    AudacityProject &proj, bool selectedOnly, double targetRate)
@@ -983,7 +970,7 @@ void ProjectAudioManager::OnPause()
    }
 
    bool paused = !projectAudioManager.Paused();
-   projectAudioManager.SetPaused( paused );
+   TogglePaused();
 
    auto gAudioIO = AudioIO::Get();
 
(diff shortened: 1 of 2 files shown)


src/ProjectAudioManager.cpp
```diff
@@ -1089,10 +1076,18 @@ void ProjectAudioManager::OnCommitRecording()
 
 void ProjectAudioManager::OnSoundActivationThreshold()
 {
-   auto &project = mProject;
    auto gAudioIO = AudioIO::Get();
-   if ( gAudioIO && &project == gAudioIO->GetOwningProject().get() ) {
```
Paul-Licameli, line 1094: The simplification lost the restriction, testing GetOwningProject.  What justifies the removal of that condition?

src/ProjectAudioManager.cpp
```diff
@@ -1089,10 +1076,18 @@ void ProjectAudioManager::OnCommitRecording()
...
-      wxTheApp->CallAfter( [this]{ Pause(); } );
+   if (gAudioIO)
+   {
+      bool canStop =  CanStopAudioStream();
+
+      gAudioIO->SetPaused(!gAudioIO->IsPaused());
+
+      if (canStop)
+      {
+         // this is what ::OnPause really boils down to, when we consider that we are not scrubbing
```
Paul-Licameli, line 1088: What "proves" that you aren't scrubbing?  I'll tell you:  This function is only reached if recording is happening, and recording is mutually exclusive with scrubbing.  So comment.

src/ProjectAudioManager.cpp
```diff
@@ -1089,10 +1076,18 @@ void ProjectAudioManager::OnCommitRecording()
...
+         SetPaused( !Paused() );
```
Paul-Licameli, line 1089: This calls SetPaused on ProjectAudioManager.  It is only changing the boolean `mPaused`.

Strictly speaking, this means the boolean ought to be an atomic, tested and updated with relaxed memory ordering, because it is the non-main thread changing it here, but it is tested elsewhere in the main thread.

src/ProjectAudioManager.cpp
Paul-Licameli: I mean, comment that reasoning here.  But also, comment that `OnSoundActivationThreshold` is called only during recording, in AudioIOListener.h.  Make the assumptions you rely on an explicit part of the contract for that interface (even if only informally as comments).

Paul-Licameli:
Overall, you took a different approach from what I expected, but it is acceptable, with the one exception of my first comment above, where you made an un-conservative change that wasn't explained.

Make changes and comments as I suggested, and just squash all to one commit -- this is small enough.

src/ProjectAudioManager.cpp
pietro68: Good point. Nothing justifies it, it was my mistake. I will reinstate the check.

src/ProjectAudioManager.cpp
pietro68: I was assuming the same (i.e. that when recording, one can not scrub) but it is good to have a confirmation by you and yes, better write that in a comment.

src/ProjectAudioManager.cpp
pietro68: Completely right, I did not see this potential problem! I will do as you advice.

pietro68: I made the mPaused member an atomic, but in order to be able to toggle it atomically, I had to make it an int; please have a look at my latest commit and let me know what do you think.

Paul-Licameli:
Added commits are good -- but ProjectAudioManager::Paused() now fetches the atomic with implied std::memory_order_seq_cst which could just be relaxed.

Please make that one change and squash all to one commit, then merge this.

Paul-Licameli (approved):

Issue #2593, opened 2022-02-25
Note Track has an Effect button in TCP


petersampsonaudacity:
**Describe the bug**
With latest 3.2,0 alpha Note Tracks have an Effect button in their Track Control Panels - **_BUT_** you can't apply effects to a MIDI Note track

**To Reproduce**
1. Install latest alpha master and run it
2. Import a MIDI file
3. Observe the TCP has an **Effect** button

**Expected behavior**
No **Effect** button for Note tracks

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/155756438-60cc1875-fd10-4b66-bed3-ff59e599b2cc.png)

**Additional information (please complete the following information):**
 - OS: Windows 10 - but assume all OS
 - Version: latest alpha master:   audacity-win-3.2.0-alpha-20220225+9edc800-x64-msvc2022

**Additional context**
_I note too that pressing the Effect button in the TCP of an audio track just creates empty white space to the left of the track and shifts the track rightwards and there is now way of reverting from that - I assume that this is probably Work-In-Progress towards stack-able per-track effects._

Flagging @vsverchinsky as Vitaly made this alpha build

petersampsonaudacity:
Tested on W10 with @vsverchinsky  Vitaly's branch build:  audacity-win-3.2.0-alpha-20220301+8208b60-x64-msvc2022 

Looks fine now in Vitaly's branch:   
![image](https://user-images.githubusercontent.com/50205984/156173712-f069213e-50a1-4333-8685-931ea5a7c53b.png)

petersampsonaudacity:
Tested on W10 with audacity-win-3.2.0-alpha-20220322+b1dc193-x64-msvc2022

Looks fine now
a) imported Note track has **_no_** Effects button
a) imported audio track **_has_** an Effects button
a) recorded audio track **_has_** an Effects button

PR #2601, opened 2022-03-01
Remove "Effects" button from track panel

vsverchinsky:
Resolves: #2593

Recommended:


src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -1077,7 +1077,7 @@ using TCPLine = TrackInfo::TCPLine;
 static const struct WaveTrackTCPLines
    : TCPLines { WaveTrackTCPLines() {
    (TCPLines&)*this =
-      PlayableTrackControls::StaticTCPLines();
+      PlayableTrackControls::StaticWaveTCPLines();
```
Paul-Licameli, line 1080: No surprises

Paul-Licameli (approved):

Issue #2578, opened 2022-02-21
const-correctness of effect-related methods


Paul-Licameli:

Paul-Licameli:
All subtasks are now done!

PR #2619, opened 2022-03-03
point 1 of i2578 - const correctness

pietro68:
Resolves: point 1 of https://github.com/audacity/audacity/issues/2578

Adding const-ness to methods of ComponentsInterface had very extended ripple effects - hundreds of files had to be changed.
Changes were trivial (just adding "const" in the right places), except one case in which I had to declare a member as mutable - but without changing functionality.

Recommended:

### DIFF
--- a/src/commands/AudacityCommand.cpp
+++ b/src/commands/AudacityCommand.cpp
@@ -64,7 +64,7 @@ AudacityCommand::~AudacityCommand()
 
 
 PluginPath AudacityCommand::GetPath() const {        return BUILTIN_GENERIC_COMMAND_PREFIX + GetSymbol().Internal(); }
-VendorSymbol AudacityCommand::GetVendor(){      return XO("Audacity");}
+VendorSymbol AudacityCommand::GetVendor() const {      return XO("Audacity");}
 wxString AudacityCommand::GetVersion(){     return AUDACITY_VERSION_STRING;}
 
 
--- a/src/commands/AudacityCommand.cpp
+++ b/src/commands/AudacityCommand.cpp
@@ -65,7 +65,7 @@ AudacityCommand::~AudacityCommand()
 
 PluginPath AudacityCommand::GetPath() const {        return BUILTIN_GENERIC_COMMAND_PREFIX + GetSymbol().Internal(); }
 VendorSymbol AudacityCommand::GetVendor() const {      return XO("Audacity");}
-wxString AudacityCommand::GetVersion(){     return AUDACITY_VERSION_STRING;}
+wxString AudacityCommand::GetVersion()  const {     return AUDACITY_VERSION_STRING;}
 
 
 bool AudacityCommand::Init(){
--- a/src/commands/AudacityCommand.cpp
+++ b/src/commands/AudacityCommand.cpp
@@ -63,7 +63,7 @@ AudacityCommand::~AudacityCommand()
 }
 
 
-PluginPath AudacityCommand::GetPath(){        return BUILTIN_GENERIC_COMMAND_PREFIX + GetSymbol().Internal(); }
+PluginPath AudacityCommand::GetPath() const {        return BUILTIN_GENERIC_COMMAND_PREFIX + GetSymbol().Internal(); }
 VendorSymbol AudacityCommand::GetVendor(){      return XO("Audacity");}
 wxString AudacityCommand::GetVersion(){     return AUDACITY_VERSION_STRING;}
 
--- a/src/commands/CommandType.h
+++ b/src/commands/CommandType.h
@@ -44,13 +44,12 @@ class wxString;
 class AUDACITY_DLL_API OldStyleCommandType : public AudacityCommand
 {
 private:
-   ComponentInterfaceSymbol mSymbol;
    std::optional<CommandSignature> mSignature;
 
 public:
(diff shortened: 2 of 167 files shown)


src/commands/CommandType.h
```diff
@@ -44,21 +44,21 @@ class wxString;
 class AUDACITY_DLL_API OldStyleCommandType : public AudacityCommand
 {
 private:
-   ComponentInterfaceSymbol mSymbol;
+   mutable ComponentInterfaceSymbol mSymbol;
```
Paul-Licameli, line 47: The first commit in this pull request fails to build for me.  In this case, these classes are the problem:

AudioUnitEffects
AudioUnitEffectsModule
ApplicationPrefs

The last of these is not macOS specific -- and that explains why your PR is failing to build on all platforms.

The last of these is conditionally compiled, only if you enable check-for-updates in your build.  It is off by default for anyone who downloads and builds Audacity.  It is on for our Continuous Integration (CI) builds.

To turn it on for yourself, you must re-configure the project, setting certain CMake variables.  Re-configure just once, and those variable changes will be cached and reused with each build.

You can fix it using the cmake-gui program if you like, or do this just from a command line by passing `-Dname=value` arguments to cmake.

You need
audacity_has_updates_check=YES
audacity_has_networking=YES

Paul-Licameli (changes requested):

src/commands/CommandType.h
Paul-Licameli, line 47: There is a misspelling in the first commit comment.

There is another use of mutable, which is not in a very important class.  Still, when you do use mutable or const_cast to cheat strict const correctness, you should either think about how to eliminate that, or write a comment to explain why it is defensible.

What is a good excuse for mutable?
* you are caching some computation, and you can prove that the cache is never invalidated by later changes of the data that the result depends on.
* the computation may be expensive.

More examination of this case tells me the first is correct but the second may not be.

Can you simply eliminate the mutable variable?  It is easy in this case.  The member variable mSymbol is used for nothing else but to impement `OldStyleCommandType::GetSymbol`.

Could you move the assignment of mSymbol to some earlier place in initialization of the `OldStyleCommandType` object?  (As I suggested for the other mutable in the other pull request.)

Actually, no, not in this case!

The only obvious place to try that is in the constructor of `OldStyleCommandType` -- but when you are still constructing a base class object, the virtual function overrides defined by the derived class (here, `BuildName` is the function) cannot be called yet.  That is a C++ language rule.

I wrote a lot about an unimportant bit of code but the principles should be remembered for other cases.

So, either comment about mutable mSymbol, or just remove it.  I say remove it.

Paul-Licameli (changes requested):
Another mutable to rewrite or at least to comment about

Paul-Licameli:
All the automatic checks now pass.  Can you now make an interactive rebase to apply these fix ups earlier in the commit sequence?

src/commands/CommandType.h
pietro68: Thanks, that worked

Paul-Licameli:
> I squashed the two OSx fix commits into one, and squashed another fix commit, but I can not move the fix commits any earlier than where they are now (after my last rebase-forcepush), unless I split commit `fix OSx build` in five distinct pieces and squash these with other commits, and likewise split/squash `fix ApplicationPrefs` in two pieces. Should I do this? These commits are all going to be squashed together in the end anyway before the merge, or?

If you can make the extra effort so that each commit in the sequence remains buildable on MacOS, I would prefer that.

mic you do an interactive rebase, change each “pick” to “edit” and then you can stop at each commit, make the small fixes, and continue rebasing.

src/commands/AudacityCommand.cpp
```diff
@@ -63,7 +63,7 @@ AudacityCommand::~AudacityCommand()
 }
 
 
-PluginPath AudacityCommand::GetPath(){        return BUILTIN_GENERIC_COMMAND_PREFIX + GetSymbol().Internal(); }
+PluginPath AudacityCommand::GetPath() const {        return BUILTIN_GENERIC_COMMAND_PREFIX + GetSymbol().Internal(); }
```
Paul-Licameli, line 66: I notice now that the declarations in this class are missing the override specifier.  It would be nice to add one commit that supplies that, but not essential.

src/prefs/PrefsPanel.h
```diff
@@ -102,7 +102,7 @@ class AUDACITY_DLL_API PrefsPanel /* not final */
    virtual bool Commit() = 0; // used to be called "Apply"
 
 
-   virtual PluginPath GetPath();
```
Paul-Licameli, line 105: “override” is missing from these function declarations.  It would be nice to add a commit that supplies it.  Not essential though for this issue.

Paul-Licameli (approved):

Paul-Licameli:
I approve the changes.  Just these suggestions:

the last commit can be combined into the first with a “fix up.”

the comment of the first commit has a spelling mistake.

a new commit might add “override” as suggested above.

Now merge it.  Remember:  first, fetch latest master and rebase onto that.  Then, force-push.  Then, use the merge button on this page, choosing first from the menu, “create merge commit.”  Then before you confirm, you can edit the commit comment for the merge, giving an explanation of the whole.

PR #2621, opened 2022-03-03
point 2 of i2578

pietro68:
Resolves: point 2 of https://github.com/audacity/audacity/issues/2578

Again, trivial but extensive work. One member had to be made mutable.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -135,7 +135,7 @@ public:
    virtual bool SupportsAutomation() const = 0;
 
    //! Whether the effect dialog should have a Debug button; default, always false.
-   virtual bool EnablesDebug();
+   virtual bool EnablesDebug() const;
 
    //! Name of a page in the Audacity alpha manual, default is empty
    virtual ManualPageID ManualPage();
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -114,7 +114,7 @@ public:
    virtual ~EffectDefinitionInterface();
 
    //! Type determines how it behaves.
-   virtual EffectType GetType() = 0;
+   virtual EffectType GetType() const = 0;
 
    //! Determines which menu it appears in; default same as GetType().
    virtual EffectType GetClassification();
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -120,7 +120,7 @@ public:
    virtual EffectType GetClassification() const;
 
    //! Report identifier and user-visible name of the effect protocol
-   virtual EffectFamilySymbol GetFamily() = 0;
+   virtual EffectFamilySymbol GetFamily() const = 0;
 
    //! Whether the effect needs a dialog for entry of settings
    virtual bool IsInteractive() = 0;
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -123,7 +123,7 @@ public:
    virtual EffectFamilySymbol GetFamily() const = 0;
 
    //! Whether the effect needs a dialog for entry of settings
-   virtual bool IsInteractive() = 0;
+   virtual bool IsInteractive() const = 0;
 
    //! Whether the effect sorts "above the line" in the menus
    virtual bool IsDefault() = 0;
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -132,7 +132,7 @@ public:
    virtual bool SupportsRealtime() const = 0;
 
    //! Whether the effect has any automatable controls.
-   virtual bool SupportsAutomation() = 0;
+   virtual bool SupportsAutomation() const = 0;
 
    //! Whether the effect dialog should have a Debug button; default, always false.
    virtual bool EnablesDebug();
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -129,7 +129,7 @@ public:
    virtual bool IsDefault() const = 0;
 
    //! Whether the effect supports realtime previewing (while audio is playing).
-   virtual bool SupportsRealtime() = 0;
+   virtual bool SupportsRealtime() const = 0;
 
    //! Whether the effect has any automatable controls.
    virtual bool SupportsAutomation() = 0;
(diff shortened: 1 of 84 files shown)


libraries/lib-components/EffectInterface.h
```diff
@@ -114,37 +114,37 @@ class COMPONENTS_API EffectDefinitionInterface  /* not final */ : public Compone
    virtual ~EffectDefinitionInterface();
 
    //! Type determines how it behaves.
-   virtual EffectType GetType() = 0;
+   virtual EffectType GetType() const = 0;
```
Paul-Licameli, line 117: This PR fails to compile on macOS.  AudioUnitEffect is compiled only for that operating system, so a Windows build could succeed while the necessary changes to AudioUnitEffect were not done to make its methods const.

Even if you don't have a Mac available, it should not be difficult to amend each commit to build for Mac by using a test editor to fix AudioUnitEffect at each step.

Paul-Licameli (changes requested):

libraries/lib-components/EffectInterface.h
```diff
@@ -114,37 +114,37 @@ class COMPONENTS_API EffectDefinitionInterface  /* not final */ : public Compone
...
-   virtual bool IsDefault() = 0;
+   virtual bool IsDefault() const = 0;
 
    //! Whether the effect supports realtime previewing (while audio is playing).
-   virtual bool SupportsRealtime() = 0;
+   virtual bool SupportsRealtime() const = 0;
 
    //! Whether the effect has any automatable controls.
-   virtual bool SupportsAutomation() = 0;
+   virtual bool SupportsAutomation() const = 0;
```
Paul-Licameli, line 135: The commit that adds `const` to `SupportsAutomation` has an incorrect commit comment.

src/effects/nyquist/Nyquist.h
```diff
@@ -240,7 +240,7 @@ class AUDACITY_DLL_API NyquistEffect final : public Effect
    TranslatableString mCopyright;
    wxString          mManPage;   // ONLY use if a help page exists in the manual.
    wxString          mHelpFile;
-   bool              mHelpFileExists;
+   mutable bool      mHelpFileExists;
```
Paul-Licameli, line 243: This is the one place where you added mutable.

Paul-Licameli: It is mutable because it is reassigned in `NyquistEffect::HelpPage`.  But it would be better to find a fix that does not rely on mutable.  Instead, mHelpFileExists should be assigned at some earlier place inside a non-const function that initializes the NyquistEffect.

What other things does the correct value of mHelpFileExists depend on?  When are those things assigned and no longer changed, so that you could move the computation of mHelpFileExists correctly to an earlier place that is not too early?

Paul-Licameli (changes requested):
I'd like to avoid using mutable by moving the assignment to that variable

libraries/lib-components/EffectInterface.h
pietro68: right, corrected

libraries/lib-components/EffectInterface.h
pietro68: Alright, fixed now

src/effects/nyquist/Nyquist.h
pietro68: I moved the intialization of mHelpFileExists at the end of the constructor, please see commit 3aa08b2

pietro68:
I will do an interactive rebase so to redistribute the OSX fixes to keep every commit building, like I did for PR 2619

src/effects/nyquist/Nyquist.cpp
```diff
@@ -195,6 +195,8 @@ NyquistEffect::NyquistEffect(const wxString &fName)
 
    if (!mOK && mInitError.empty())
       mInitError = XO("Ill-formed Nyquist plug-in header");
+
+   mHelpFileExists = CheckHelpPage().first;
```
Paul-Licameli, line 199: What other member variables does CheckHelpPage depend on?  Answer: mHelpFile

Where is mHelpFile reassigned?  Answer: NyquistEffect::Parse and NyquistEffect::ParseProgram

Are these functions called only in the constructor of NyquistEffect?  Answer: no (explore the call graph)

Therefore, the end of the constructor is not the correct place to assign the variable.

But, the end of ParseProgram would be the correct place.  Because Parse is called only inside ParseProgram.  So that whenever mHelpFile might change, the update of mHelpFileExists will be correct.

src/effects/nyquist/Nyquist.cpp
```diff
@@ -241,48 +243,55 @@ TranslatableString NyquistEffect::GetDescription()
...
       }
    }
-   return wxEmptyString;
+   return { false, wxEmptyString };
+}
+
+
+FilePath NyquistEffect::HelpPage() const
+{
+   return CheckHelpPage().second;
```
Paul-Licameli, line 272: You invoke CheckHelpPage() again, but I think you could rely instead on the result of it being pre-computed.  Call the function in one place only and make another member variable that stores the second value.

Paul-Licameli (changes requested):

src/effects/nyquist/Nyquist.cpp
pietro68: Good catch - I was making the wrong assumption that `ParseFile `(which calls `ParseProgram`) was called in the constructor only, but it is called in ::`Init `too.

src/effects/nyquist/Nyquist.cpp
pietro68: ok, done

src/effects/nyquist/Nyquist.cpp
pietro68: done

src/effects/audiounits/AudioUnitEffect.h
```diff
@@ -77,8 +77,8 @@ class AudioUnitEffect : public wxEvtHandler,
 
    // EffectProcessor implementation
 
-   unsigned GetAudioInCount() override;
-   unsigned GetAudioOutCount() override;
+   unsigned GetAudioInCount() const override;
+   unsigned GetAudioOutCount() const override;
```
Paul-Licameli, line 81: One remaining small defect of this pull request is that the first commit fails to build on MacOS but the second commit corrects that.  The declaration of GetAudioInCount does not match the definition in the first commit.

I trust you to fix that as you rebase into latest master.

Paul-Licameli (approved):
One suggestion

src/effects/audiounits/AudioUnitEffect.h
pietro68: That's right! thanks, corrected.

Issue #2531, opened 2022-02-08
Move rendering code away from WaveformView. WaveformView should only supply data to the renderer.


AnitaBats:
* Implement software waveform rendering. 
* Implement new cache for waveform imagery (reusable later by spectrum view)

crsib:
Original case:
[rendering_test.zip](https://github.com/audacity/audacity/files/8033480/rendering_test.zip)

![image](https://user-images.githubusercontent.com/2660628/153221742-bdd63726-9df5-4cf4-b5c4-dbb8d8ba86d4.png)

PR #2635, opened 2022-03-05
Waveform Rendering

crsib:
Resolves: #2531 

The goals of this PR are to separate waveform rendering logic into a library and to provide a base for the protentional hardware accelerated approach. 

## Approach to data caching 

Data is cached in blocks with 256 columns each. Each block is identified by a pair of `(zoomLevel, firstSample)`. Each clip has its own set of caches. 

Dividing data into blocks helps to avoid unnecessary copying of data, which is expensive on the high-resolution displays. On top of that, we can have hierarchies of caches. 

Recommended:


libraries/lib-graphics/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+#[[
+A library responsible for custom rendering in Audacity
+]]#
```
Paul-Licameli, line 3: The trailing # isn't needed to bracket a long comment

libraries/lib-graphics/FrameStatistics.cpp
```diff
@@ -0,0 +1,130 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
```
Paul-Licameli, line 1: Is this a comment convention we should do in all files?

There should be more regularity but I'm not sure what exactly it should be.

libraries/lib-graphics/FrameStatistics.cpp
```diff
@@ -0,0 +1,130 @@
...
+/**********************************************************************
```
Paul-Licameli, line 2: Start with `/*!` so Doxygen properly sees it

libraries/lib-graphics/FrameStatistics.cpp
crsib: This is a standardized approach, so yes, I would like it see more of it!

libraries/lib-graphics/FrameStatistics.h
```diff
@@ -0,0 +1,92 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  FrameStatistics.h
+
+  Dmitry Vedenko
```
Paul-Licameli, line 8: Again `/*!` and use `@file` before the file name, and add a lest a `@brief` line about the purpose of the header.

Much more Doxygenation might be added but that much is the minimum I would like so there is some summary on pages like this one https://doxy.audacityteam.org/dir_68267d1309a1af8e8297ef4c3efbcdba.html

I'm less concerned to add `@file` to .cpp files, because they are not interfaces.  It's generally understood that they implement whatever the companion .h file is about.

src/AudacityApp.cpp
```diff
@@ -502,6 +503,7 @@ static void QuitAudacity(bool bForce)
    // termination when exiting is requested
    #if !defined(__WXMAC__)
    LogWindow::Destroy();
+   FrameStatisticsDialog::Destroy();
```
Paul-Licameli, line 506: So the FrameStatisticsDialog is a per-application non-modal dialog (not per-project)?

src/FrameStatisticsDialog.cpp
```diff
@@ -0,0 +1,149 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/**********************************************************************
```
Paul-Licameli, line 2: Ditto doxygen

src/FrameStatisticsDialog.cpp
```diff
@@ -0,0 +1,149 @@
...
+   };
+
+   Section mSections[size_t(FrameStatistics::SectionID::Count)];
+
+   Observer::Subscription mStatisticsUpdated;
+
+   std::chrono::steady_clock::time_point mLastUpdate;
+};
+
+Destroy_ptr<Dialog> sDialog;
```
Paul-Licameli, line 127: I wish to deprecate Destroy_ptr in favor of: https://docs.wxwidgets.org/3.0/classwx_window_ptr_3_01_t_01_4.html

I didn't know of that when I reinvented Destroy_ptr.

libraries/lib-graphics/PixelSampleMapper.h
```diff
@@ -0,0 +1,69 @@
...
+
+      double mInitialValue {};
+      double mSamplesPerPixel {};
+
+      sampleCount operator()(uint32_t column) const noexcept;
+
+      explicit operator bool() const noexcept;
+   };
+   // GCC 9.3.0 fails horribly if you do not initialize variant explicitly here
+   std::variant<LinearMapper, CustomMapper> mMapper { LinearMapper {} };
```
Paul-Licameli, line 68: The commit that introduces this variant didn't compile for me with clang either.  I fixed it for myself by initializing with one argument `std::in_place_type<LinearMapper>`.  I expect the built fix-ups will be squashed with interactive rebasing in a future version of this PR.

libraries/lib-utility/MemoryX.h
```diff
@@ -574,4 +574,16 @@ auto Visit(Visitor &&vis, Variant &&var)
...
+
+template <typename LType, typename RType>
+auto RoundUp(LType numerator, RType denominator) noexcept
+{
+   static_assert(std::is_integral_v<LType>);
+   static_assert(std::is_integral_v<RType>);
+
+   const auto result = numerator / denominator;
+
+   return result * denominator == numerator ? result : result + 1;
```
Paul-Licameli, line 586: https://en.cppreference.com/w/cpp/language/operator_arithmetic

If the denominator is zero, there is UB not an exception, so `noexcept` is strictly speaking correct.

If the values are not both positivie, what should happen?  (Division must truncate toward zero in C++11 and later; so if this case matters, then it is already "rounded up" and adding 1 is wrong.)

If those cases aren't supposed to happen, add a doxygen `@pre` that numerator is nonnegative and denominator is positive.  (Precondition means "it's callers responsibility to exclude those cases, not mine to handle them")

Ignoring the sign question, I would simplify:

`return (numerator + denominator - 1) / denominator;`

libraries/lib-graphics/GraphicsDataCache.h
```diff
@@ -0,0 +1,327 @@
...
+   double mSampleRate {}; // DV: Why do we use double for sample rate? I don't know
+
+   // The max width of the request processed in pixels
+   int64_t mMaxWidth {};
+   // This value is incremented on every lookup
+   uint64_t mCacheAccessIndex {};
+   // A multiplier used to control the cache size
+   int32_t mCacheSizeMultiplier { 4 };
+
+   template <typename CacheElementType>
```
Paul-Licameli, line 130: I get a clang build error:

```
GraphicsDataCache.h:213:23: error: declaration of 'CacheElementType' shadows template parameter
   template <typename CacheElementType>
```

Easily fixed by using another name (like just T)

crsib: Build errors are fixed in later commits. Given that I have to build on 3 different systems I will postpone force pushes to the much later times

libraries/lib-utility/MemoryX.h
crsib: Great catch, thank you!

libraries/lib-graphics/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+   waveform/WaveDataCache.cpp
+   waveform/WaveDataCache.h
+)
+set( LIBRARIES
+   PUBLIC
+      lib-utility-interface
+   PRIVATE
+      lib-math-interface
+      lib-screen-geometry-interface
+      wxwidgets::base
```
Paul-Licameli, line 26: At the later commit that adds more dependencies, I would simplify:

```
set( LIBRARIES
       lib-math-interface
       lib-screen-geometry-interface
PRIVATE
       wxwidgets::base
)
```

The transitive dependency on lib-utility-interface is now implied, so the generated graph of dependencies can have one less arc.  Consistently with other libraries, only the dependency on wxBase should be private.  Libraries that depend on lib-graphics could themselves simplify their CMakeLists.txt by using transitive dependency on lib-math and lib-screen-geometry.

libraries/lib-graphics/PixelSampleMapper.h
Paul-Licameli: I see a commit "Fixes GCC build" adding this line but followed by another that fixes Clang again.

Using `in_place_type` instead is enough to fix Clang in the first commit, without including SampleCount.h.

If you are curious about why the incomlete type was a problem... I looked a bit at the error-novel for clang and I think there is a metaprogram deducing the correct alternative of the variant to initialize from the constructor arguments, which leads to the expansion of a conditional `noexcept` specifier, and somehow the operator () of this type is used in an unevaluated context.

But use `in_place_type` instead and that metaprogram isn't needed to resolve the overloaded constructor.

So, avoiding one more #include inside a header is nice if you can do it, plus there is in-place construction instead of a copy of a LinearMapper.  Small gains.

libraries/lib-utility/MemoryX.h
crsib: Your version is faster, but does not consider the possible integer overflow. I think I will just add Unsafe to the function name, though :-)

libraries/lib-graphics/GraphicsDataCache.h
```diff
@@ -0,0 +1,327 @@
...
+      }
+
+      if (element == nullptr)
+         return nullptr;
+
+      if (mInitializer)
+      {
+         if (!mInitializer(key, *element))
+         {
+            DisposeElement(element);
```
vsverchinsky, line 293: No `noexcept` guarantees

LWinterberg:
Some quick tests: 

* "Generate Silence" is *really* silent now: 
image
* The scrub ruler isn't granted space and is overlaying the tracks and z-fighting it 
image
* The entire project area is now flickering slightly on my HDR-capable/OLED monitor but not my LCD. Fwiw: This is also the case for Ableton Live, but is not the case for Audacity 3.1.3
* The entire thing still seems to be locked to 20Hz

you probably know half of these, but I thought I'd note them down anyway :)

crsib:
>The entire thing still seems to be locked to 20Hz

There was an issue with how is selection handled, for example. And there is still lots of stuff to be fixed in that domain.

>The entire project area is now flickering slightly 

Well, that is something we will have to investigate later. Too bad I have no HDR display available

PR #2638, opened 2022-03-07
More uses of effect settings

Paul-Licameli:
Resolves: #2629

Pass EffectSettings into still more functions, ultimately to allow elimination of the special setting Effect::mDuration which is common to many effects.  And that will allow elimination of the confusing class EffectHostInterface whose purpose isn't clear.

Recommended:

### DIFF
--- a/src/effects/EffectUI.cpp
+++ b/src/effects/EffectUI.cpp
@@ -886,6 +886,7 @@ void EffectUIHost::OnUserPreset(wxCommandEvent & evt)
    auto settings = mpAccess->Get();
    mEffectUIHost.GetDefinition().LoadUserPreset(
       mEffectUIHost.GetUserPresetsGroup(mUserPresets[preset]), settings);
+   TransferDataToWindow();
    // Communicate change of settings
    mpAccess->Set(std::move(settings));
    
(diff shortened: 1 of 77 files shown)


pietro68 (approved):

src/effects/EffectUI.cpp
```diff
@@ -554,12 +554,10 @@ void EffectUIHost::OnApply(wxCommandEvent & evt)
       return;
    }
    
-   // This will take care of calling TransferDataFromWindow() for an effect.
-   if (!mEffectUIHost.GetDefinition().SaveUserPreset(
-      mEffectUIHost.GetCurrentSettingsGroup(), mpAccess->Get()))
-   {
+   if (!TransferDataFromWindow() ||
```
pietro68, line 563: If the call to `TransferDataFromWindow()` fails, the call to `SaveUserPreset ` will still be tried? I would not expect that

src/effects/EffectUI.cpp
```diff
@@ -554,12 +554,10 @@ void EffectUIHost::OnApply(wxCommandEvent & evt)
...
+       !mEffectUIHost.GetDefinition().SaveUserPreset(
+         mEffectUIHost.GetCurrentSettingsGroup(), mpAccess->Get()))
       return;
```
pietro68, line 566: Just a general consideration about these methods: they return void, so if TransferDataFromWindow fails, the error is not propagated to the caller.

pietro68 (approved):
I think I have enough understanding of the code now. 

I only note that many uses of TransferData(To|From) do not check their return value, or if they do, they do not propagate failure up in the caller tree. But of course this is because the calling methods return void - so this has to do with more general architecture and is out of the scope of the current work.

src/effects/EffectUI.cpp
Paul-Licameli: TransferDataFromWindow may fail because a number was typed that is out of the allowed range, for example.  It may alert the user with an error dialog.

PR #2647, opened 2022-03-08
point 3 of I2578

pietro68:
Resolves: point 3 of https://github.com/audacity/audacity/issues/2578

This is blocked by a problem in NyquistEffect; I left the temporary fix to it in a separate commit, just to make it easier to see where it is. I need advice to find a proper solution for this.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -259,7 +259,7 @@ public:
    //! Save current settings into parms
    virtual bool GetAutomationParameters(CommandParameters & parms) = 0;
    //! Change settings to those stored in parms
-   virtual bool SetAutomationParameters(CommandParameters & parms) = 0;
+   virtual bool SetAutomationParameters(const CommandParameters & parms) = 0;
 
    //! Change settings to a user-named preset
    virtual bool LoadUserPreset(const RegistryPath & name) = 0;
--- a/src/effects/Amplify.cpp
+++ b/src/effects/Amplify.cpp
@@ -149,7 +149,7 @@ bool EffectAmplify::GetAutomationParameters(CommandParameters & parms)
    return true;
 }
 
-bool EffectAmplify::SetAutomationParameters(CommandParameters & parms)
+bool EffectAmplify::SetAutomationParameters(const CommandParameters & parms)
 {
    ReadAndVerifyFloat(Ratio);
    mRatio = Ratio;
--- a/src/effects/Amplify.h
+++ b/src/effects/Amplify.h
@@ -41,7 +41,7 @@ public:
 
    EffectType GetType() const override;
    bool GetAutomationParameters(CommandParameters & parms) override;
-   bool SetAutomationParameters(CommandParameters & parms) override;
+   bool SetAutomationParameters(const CommandParameters & parms) override;
    bool LoadFactoryDefaults() override;
 
    // EffectProcessor implementation
--- a/src/effects/AutoDuck.cpp
+++ b/src/effects/AutoDuck.cpp
@@ -155,7 +155,7 @@ bool EffectAutoDuck::GetAutomationParameters(CommandParameters & parms)
    return true;
 }
 
-bool EffectAutoDuck::SetAutomationParameters(CommandParameters & parms)
+bool EffectAutoDuck::SetAutomationParameters(const CommandParameters & parms)
 {
    ReadAndVerifyDouble(DuckAmountDb);
    ReadAndVerifyDouble(InnerFadeDownLen);
--- a/src/effects/AutoDuck.h
+++ b/src/effects/AutoDuck.h
@@ -38,7 +38,7 @@ public:
 
    EffectType GetType() const override;
    bool GetAutomationParameters(CommandParameters & parms) override;
-   bool SetAutomationParameters(CommandParameters & parms) override;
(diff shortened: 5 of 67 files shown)


pietro68:
I fixed the osX build and squashed all in a single commit, including the const_cast fix for NyquistEffect.

Paul-Licameli:
I'm not satisfied with the const_cast in NyquistEffect.cpp.  How can you really fix the problem revealed by this change in argument type, instead of just hiding it?  const_cast is not to be used too liberally.

`CommandParameters::SetParameters` is called in one other place in Nyquist.cpp, and it is on a temporary `CommandParameters` object.

I think you should do the same along this path in `NyquistEffect::SetAutomationParameters`.

Do you understand all that is happening here?  I doubt it (because I didn't either, being unfamiliar with it).  I'm looking closer, and I think there is another bug in Nyquist revealed here, but not important to fix now.

The path is taken only when using the Nyquist Prompt from the Tools menu, which allows you to enter what would otherwise be stored as text in a .ny file.  I may explain more detail later, or see if I can define an Issue with steps to reproduce demonstrating the problem for a later fix.

pietro68:
I do not understand what is happening here. A sequence of steps to trigger that branch would help.

Are you suggesting that I fix things in this way:

```
CommandParameters dummyParms;
dummyParms.SetParameters(mParameters);
```

Basically, this is equivalent to doing nothing. Is that what happens in the original code too? I mean, the input `CommandParameters&` would not be used after the call, or?

Paul-Licameli:
> I do not understand what is happening here. A sequence of steps to trigger that branch would help.
> 
> Are you suggesting that I fix things in this way:
> 
> ```
> CommandParameters dummyParms;
> dummyParms.SetParameters(mParameters);
> ```
> 
> Basically, this is equivalent to doing nothing. Is that what happens in the original code too? I mean, the input `CommandParameters&` would not be used (read) after the call, or?

To figure out how to exercise this... I discovered a bug in recent work, and so I pushed another commit that fixes it onto your branch.  (GitHub allows this pushing onto other people's branches in the remote!)

This is a very rarely used piece of code.  You can try making some sound and selecting it, then Tools > Nyquist prompt... and then enter this:

```
(mult 0.5 *track*)
```

That interprets a little bit of Lisp to modify the sound.  But try it again like this:

```
;control  gain “gain” real “” 0.5 0.0 1.0
(mult gain *track*)
```

And the first magic comment line means a second dialog will open with a "gain" slider, a vlue called `gain` will be determined by that, and used as the factor.

Paul-Licameli:
> I confirm, I do not understand what is happening here! A sequence of steps to trigger that branch would help.
> 
> Are you suggesting that I fix things in this way:
> 
> ```
> CommandParameters dummyParms;
> dummyParms.SetParameters(mParameters);
> ```
> 
> Basically, this is equivalent to doing nothing. Is that what happens in the original code too? I mean, the input `CommandParameters&` would not be used (read) after the call, or?

No.  The line
`parms.Read(KEY_Parameters, &mParameters, wxEmptyString);`
sets mParameters from the input `parms` and then mParameters is reinterpreted as a nested config file that was stored as the value of `KEY_Parameters` in the outer config.

HOWEVER -- I think this very rarely exercised recursive path was never correctly implemented.  The reading done here in `SetAutomationParameters` needs to cooperate with `GetAutomationParameters`, and I think you will discover that this code never stores non-empty mParameters.

The mParameters should be remembering the controls created by the magic `;control` comments in the Nyquist prompt, but I think it may be failing to do that.

The Nyquist system for making Audacity programming is fascinating but complicated.  Debugging all this is not of very high value right now.

pietro68:
I placed a breakpoint at the start of `NyquistEffect::SetAutomationParameters`, then tried both nyquist commands you suggested; they worked correctly on the selected audio, but the breakpoint was never reached.

I am not sure that I know how to go on now; const_cast or something else?

Paul-Licameli:
> I placed a breakpoint at the start of `NyquistEffect::SetAutomationParameters`, then tried both nyquist commands you suggested; they worked correctly on the selected audio, but the breakpoint was never reached.
> 
> I am not sure that I know how to go on now; const_cast or something else?

You should be able to visit `NyquistEffect::SetAutomationParameters` and enter the block `if (mIsPrompt)`, yes?  Then read mParameters.

But -- mParameters is always empty so you don't reach the changed lines.

It is always empty because, I believe, mParameters is never first saved by `GetAutomationParameters`.

pietro68:
Alright, I found something about reaching the method. But before I expose what I found, I would like to understand the general direction: should I try to fix this constness problem in the proper way, or can we go with the const_cast for now, and maybe some comments?

So this is what I found: if I place a breakpoint at the start of NyquistEffect::SetAutomationParameters, I can reach it, but only on the first time that I select Tools | Nyquist prompt...  and not on subsequent tries.

Also, the breakpoint is reached two times in succession, apparently with the same parameters. And yes, in both occurrences `mParameters` is empty, so we never reach the const_cast thing.

The breakpoint is hit twice, because in `Effect::SetAutomationParametersFromString`, `DefineParams(S)` is called twice.

Paul-Licameli (approved):
This final version looks good.  I would just like the third commit squashed with the first one.

Remember:  fetch latest audacity, rebase, force-push, and merge with the choice to make a merge commit.  Before confirming, you may edit the message box in the GitHub page to make a more detailed commit comment for the merge.

PR #2652, opened 2022-03-09
point 4 of I2578

pietro68:
Resolves: point 4 of https://github.com/audacity/audacity/issues/2578

This is split in two commits - the latter only to highlight a change that may or may not be a temporary solution.

In EffectManager::GetCommandDefinition, a non-const pointer to a ComponentInterface is defined. That can not be turned in a const pointer, because later DefineParams is called on it, and one of the requisites of the original issue (at point 1) is that the latter stays non-const (not that it would be easy to make it const, anyway) - hence the const_cast.

Recommended:

### DIFF
--- a/libraries/lib-components/ComponentInterface.h
+++ b/libraries/lib-components/ComponentInterface.h
@@ -82,7 +82,7 @@ public:
    TranslatableString GetName() const;
 
    // Parameters, if defined.  false means no defined parameters.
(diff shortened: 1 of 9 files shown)


Paul-Licameli:
Please squash the two commits.

The const_cast to allow a call to DefineParams is something I will accept for now, because fixing that is known to be a large and difficult thing.  So please comment to that effect.

Can you also make `GetDefinition` a const member function?

Paul-Licameli (approved):
All good

PR #2655, opened 2022-03-09
Eliminate Effect::mDurationFormat; New event handling for DTMF generator

Paul-Licameli:
Resolves: #2673
Resolves: #2674

To eliminate EffectHostInterface and state variables in Effect that it accessed and modified has involved some unusual twists and turns...  The dependencies among the necessary tasks are becoming surprising on closer look.  This big PR gives
necessary solutions to various sub-sub-tasks.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -414,7 +414,7 @@ public:
    /*!
     @return true only if panel settings are acceptable
     */
-   virtual bool Validate() = 0;
+   virtual bool ValidateUI() = 0;
 };
 
 /*************************************************************************************//**
@@ -435,7 +435,8 @@ public:
    DefaultEffectUIValidator(
       EffectUIClientInterface &effect, EffectSettingsAccess &access);
    ~DefaultEffectUIValidator() override;
-   bool Validate() override;
+   //! Calls mEffect.ValidateUI()
+   bool ValidateUI() override;
 private:
    EffectUIClientInterface &mEffect;
    EffectSettingsAccess &mAccess;
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -64,9 +64,20 @@ typedef enum EffectType : int
 
 using EffectFamilySymbol = ComponentInterfaceSymbol;
 
+//! Non-polymorphic package of settings values common to many effects
+class EffectSettingsExtra final {
+};
+
 //! Externalized state of a plug-in
 struct EffectSettings : audacity::TypedAny<EffectSettings> {
    using TypedAny::TypedAny;
+   EffectSettingsExtra extra;
+
+   void swap(EffectSettings &other)
+   {
+      TypedAny::swap(other);
+      std::swap(extra, other.extra);
+   }
 };
 
 //! Interface for accessing an EffectSettings that may change asynchronously in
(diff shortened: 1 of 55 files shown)


src/effects/Effect.cpp
```diff
@@ -1327,7 +1321,7 @@ bool Effect::ProcessTrack(EffectSettings &settings,
    decltype(mBufferSize) outputBufferCnt = 0;
    bool cleared = false;
 
-   auto chans = std::min<unsigned>(mNumAudioOut, mNumChannels);
+   auto chans = std::min<unsigned>(GetAudioOutCount(), mNumChannels);
```
pietro68, line 1288: Could `numAudioOut` be used instead?

libraries/lib-components/EffectInterface.h
```diff
@@ -64,9 +64,20 @@ typedef enum EffectType : int
 
 using EffectFamilySymbol = ComponentInterfaceSymbol;
 
+//! Non-polymorphic package of settings values common to many effects
+class EffectSettingsExtra final {
+};
```
pietro68, line 76: Is this a placeholder that will be filled later on?

pietro68: The next commit answered my question.

src/effects/DtmfGen.cpp
```diff
@@ -568,33 +578,41 @@ bool EffectDtmf::MakeDtmfTone(float *buffer, size_t len, float fs, wxChar tone,
...
+   settings.Recalculate(*this);
    UpdateUI();
 }
 
 void EffectDtmf::OnDuration(wxCommandEvent & WXUNUSED(evt))
 {
+   auto &settings = mSettings;
    SetDuration(mDtmfDurationT->GetValue());
-   Recalculate();
+   settings.Recalculate(*this);
```
pietro68, line 608: Can we not call this on mSettings directly?

pietro68 (approved):
I looked mainly at the last 6 commits (as told to) and I think I understand what goes on there. I only added some minor observations.

src/effects/DtmfGen.cpp
Paul-Licameli: I anticipate that the means of communicating the settings into many of these functions might change might change in the future.  Binding a reference to the settings at the top will avoid the need to rewrite many lines again later.  I did that consistently, even in some short functions.

Paul-Licameli:
Pietro, please review it again.

It is rebased onto recent master.  The most recent four commits are new.  In the one before those "Define and use Settings for EffectDtmf, but not yet externalized..." I only made variable name changes.  I use `dtmfSettings` instead of `settings` in many places.  I may want the name `settings` in the same scope later for something else.

src/effects/EffectUI.cpp
```diff
@@ -207,7 +207,12 @@ EffectUIHost::~EffectUIHost()
 bool EffectUIHost::TransferDataToWindow()
 {
    // Transfer-to takes const reference to settings
-   return mEffectUIHost.TransferDataToWindow(mpAccess->Get());
+   return mEffectUIHost.TransferDataToWindow(mpAccess->Get()) &&
+      //! Do other apperance updates
+      mpValidator->UpdateUI() &&
+      //! Do valiators
```
pietro68, line 213: typo (?)

pietro68 (approved):
All understood now!

src/effects/EffectUI.cpp
Paul-Licameli: thanks

PR #2666, opened 2022-03-11
More const correctness

Paul-Licameli:
Make more functions related to effect settings management const.  These functions save and export settings, so
logically they don't change the state of the effect instance and should be const.

One const_cast where the notorious DefineParams is used, and should be cleaned up later but with difficulty.  Some other
forgivable const_casts, which comments about them.

Recommended:


src/effects/EffectManager.cpp
```diff
@@ -233,12 +233,15 @@ bool EffectManager::SupportsAutomation(const PluginID & ID)
    return false;
 }
 
+// This function is used only in the macro programming user interface
 wxString EffectManager::GetEffectParameters(const PluginID & ID)
 {
-   if (auto effect = GetEffect(ID)) {
+   auto pair = GetEffectAndDefaultSettings(ID);
+   if (auto effect = pair.first) {
+      assert(pair.second);
```
pietro68, line 241: This defends us from dereferencing a nullptr in a debug build, but not in a release

Issue #2609, opened 2022-03-01
Show the dialog about the plugins that didn’t load


AnitaBats:

dozzzzer:
When enabling a faulty plug-in, Audacity freezes.
Find the above mentioned plug-in in the attached. After plug0ins are scanned, just go to Effects > Manage Plug-ins, find that VST and click [Enable]. Then submit by clicking [OK]. After that Audacity hangs.
![image](https://user-images.githubusercontent.com/70063970/169881125-ec0f1dea-f04d-4a8e-b105-774fbbdc9cd7.png)

[adelay.zip](https://github.com/audacity/audacity/files/8757169/adelay.zip)

LWinterberg:
this feels like expected behavior to me. We disable a plugin for you because we can tell there's a problem with it with our automatic tests. However, it may be that our automatic tests aren't accurate and that it fails because it expects something to happen which gets suppressed in our test. In this case you're given the option to re-enable it here, and if it crashes that's kinda on you.

vsverchinsky:
> When enabling a faulty plug-in, Audacity freezes.

Agreed with Leonard.
We run out-of-process plugin validation to check if it's safe to run plugins, and disable them by default if it's not safe

Issue #2607, opened 2022-03-01
Check plugins on startup


AnitaBats:

dozzzzer:
@vsverchinsky 

- [ ] after a VST plug-in is enabled, it has no identifier displayed next to its full path in the Plug-in Manager window:

<img width="844" alt="image" src="https://user-images.githubusercontent.com/70063970/170029455-4b3a3150-c84f-4eb3-b153-0d3596c03cd0.png">

Only VST3 plug-ins have it.

crsib:
Why do you consider that it is a bug?

crsib:
And it is not even something new. Here is how 3.1.3 handles LV2 for example

![image](https://user-images.githubusercontent.com/2660628/170033025-dfa46317-d8fa-4c47-9e6e-b8994a42b76f.png)

PR #2668, opened 2022-03-11
Issue 2607: plugin validation

vsverchinsky:
Resolves: #2607
Resolves: #2609 
Resolves: #2608 

Recommended:

### DIFF
--- a/libraries/lib-module-manager/AsyncPluginValidator.cpp
+++ b/libraries/lib-module-manager/AsyncPluginValidator.cpp
@@ -1,0 +1,237 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file AsyncPluginValidation.cpp
+
+  @author Vitaly Sverchinsky
+
+  Part of lib-module-manager library
+
+**********************************************************************/
+
+#include "AsyncPluginValidator.h"
+
+#include <optional>
+#include <mutex>
+
+#include "BasicUI.h"
+#include "IPCChannel.h"
+#include "IPCServer.h"
+#include "spinlock.h"
+#include "XMLTagHandler.h"
+
+#include "PluginIPCUtils.h"
+#include "PluginDescriptor.h"
+#include "PluginHost.h"
+#include "XMLFileReader.h"
+
+AsyncPluginValidator::Delegate::~Delegate() = default;
+
+class AsyncPluginValidator::Impl final : 
+   public IPCChannelStatusCallback,
+   public std::enable_shared_from_this<Impl>
+{
+   //Variables below are accessed from any thread
+
+   IPCChannel* mChannel{nullptr};
+   std::optional<wxString> mRequest;
+
+   //The main reason to use spinlock instead of std::mutex here
+   //is that in most cases there will be no attempts for simultaneous
+   //data access (except OnConnect/Disconnect cases), because of
+   //protocol design
+   spinlock mSync;
+
+   //Variable below is accessed only from the main/UI thread
+
+   Delegate* mDelegate{nullptr};
+   std::unique_ptr<IPCServer> mServer;
+
+   //Variables below is accessed only from worker threads
+
+   detail::InputMessageReader mMessageReader;
+
+   //called on main thread only!
+   void StartHost()
+   {
+      auto server = std::make_unique<IPCServer>(*this);
+      if(!PluginHost::Start())
+         mDelegate->OnInternalError("Cannot start plugin host");
+      else
+         mServer = std::move(server);
+   }
+   
+   void HandleInternalError(const wxString& msg) noexcept
+   {
+      try
+      {
+         BasicUI::CallAfter([wptr = weak_from_this(), msg]
+         {
+            if(auto self = wptr.lock(); self && self->mDelegate != nullptr)
+               self->mDelegate->OnInternalError(msg);
+         });
+      }
+      catch(...)
+      {
+         //no way to report about a problem, though shouldn't be the case...
+      }
+   }
+
+   void HandleResult(detail::PluginValidationResult&& result) noexcept
+   {
+      try
+      {
+         BasicUI::CallAfter([wptr = weak_from_this(), result = detail::PluginValidationResult { result }]
+         {
+            if(auto self = wptr.lock())
+            {
+               if(self->mDelegate == nullptr)
+                  return;
+
+               //Release the current value to keep state invariant
+               //Caller is free to use Validate now
+               std::optional<wxString> request;
+               {
+                  std::lock_guard lck_sync(self->mSync);
+                  self->mRequest.swap(request);
+               }
+
+               if(!request.has_value())
+               {
+                  //Invalid state error
+                  self->mDelegate->OnInternalError(result.GetErrorMessage());
+                  return;
+               }
+
+               if(result.IsValid())
+               {
+                  for(auto& desc : result.GetDescriptors())
+                     self->mDelegate->OnPluginFound(PluginDescriptor { desc });
+               }
+               else
+               {
+                  wxString providerId;
+                  wxString pluginPath;
+                  detail::ParseRequestString(*request, providerId, pluginPath);
+
+                  //Even if plugin validation has failed or no effects were discovered,
+                  //create an entry for it but make it disabled by default
+                  PluginID ID = providerId + wxT("_") + pluginPath;
+                  PluginDescriptor pluginDescriptor;
+                  pluginDescriptor.SetPluginType(PluginTypeStub);
+                  pluginDescriptor.SetID(ID);
+                  pluginDescriptor.SetProviderID(providerId);
+                  pluginDescriptor.SetPath(pluginPath);
+                  pluginDescriptor.SetEnabled(false);
+                  pluginDescriptor.SetValid(false);
+
+                  self->mDelegate->OnPluginFound(pluginDescriptor);
+               }
+               self->mDelegate->OnValidationFinished();
+            }
+         });
+      }
+      catch(...)
+      {
+         ///no way to report about a problem, though shouldn't be the case...
+      }
+   }
+
+public:
+
+   Impl(Impl&) = delete;
+   Impl(Impl&&) = delete;
+   Impl& operator=(Impl&) = delete;
+   Impl& operator=(Impl&&) = delete;
+
+   Impl(Delegate& delegate) : mDelegate(&delegate) { }
+
+   ~Impl() override
+   {
+      //important to reset delegate before IPCChannelStatusCallback::OnDisconnect
+      //is called by server
+      mDelegate = nullptr;
+   }
+
+   void OnConnect(IPCChannel& channel) noexcept override
+   {
+      std::lock_guard lck(mSync);
+
+      mChannel = &channel;
+      if(mRequest)
+      {
+         try
+         {
+            detail::PutMessage(channel, *mRequest);
+         }
+         catch(...)
+         {
+            HandleInternalError("Can't send message to host");
+         }
+      }
+   }
+
+   void OnDisconnect() noexcept override
+   {
+      {
+         std::lock_guard lck(mSync);
+         mChannel = nullptr;
+      }
+      detail::PluginValidationResult result;
+      result.SetError("Disconnect");
+      HandleResult(std::move(result));
+   }
+
+   void OnConnectionError() noexcept override
+   {
+      HandleInternalError("Can't connect");
+   }
+
+   void OnDataAvailable(const void* data, size_t size) noexcept override
+   {
+      try
+      {
+         mMessageReader.ConsumeBytes(data, size);
+         if(mMessageReader.CanPop())
+         {
+            detail::PluginValidationResult result;
+            XMLFileReader xmlReader;
+            xmlReader.ParseString(&result, mMessageReader.Pop());
+
+            HandleResult(std::move(result));
+         }
+      }
+      catch(...)
+      {
+         HandleInternalError("Can't process response from the host");
+      }
+   }
+
+   void Validate(const wxString& providerId, const wxString& pluginPath)
+   {
+      std::lock_guard lck(mSync);
+
+      //one request at a time
+      assert(!mRequest.has_value());
+
+      mRequest = detail::MakeRequestString(providerId, pluginPath);
+      if(mChannel)
+         detail::PutMessage(*mChannel, *mRequest);
+      else
+         //create host process on demand
+         StartHost();
+   }
+};
+
+AsyncPluginValidator::AsyncPluginValidator(Delegate& delegate)
+{
+   mImpl = std::make_unique<Impl>(delegate);
+}
+
+AsyncPluginValidator::~AsyncPluginValidator() = default;
+
+void AsyncPluginValidator::Validate(const wxString& providerId, const wxString& pluginPath)
+{
+   mImpl->Validate(providerId, pluginPath);
+}
--- a/libraries/lib-module-manager/AsyncPluginValidator.h
+++ b/libraries/lib-module-manager/AsyncPluginValidator.h
@@ -1,0 +1,66 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file AsyncPluginValidator.h
+
+  @author Vitaly Sverchinsky
+
+  Part of lib-module-manager library
+
+**********************************************************************/
+
+#pragma once
+
+#include <wx/string.h>
+#include <memory>
+
+class PluginDescriptor;
+
+
+/**
+ * \brief Starts and communicates with a dedicated process to perform
+ * plugin validation. Once instantiated, client should call
+ * AsyncPluginValidation::Validate for plugin that needs to be validated.
+ * When done, AsyncPluginValidation will notify caller via Delegate on the
+ * UI thread (requires event loop). After Delegate::OnValidationFinished
+ * is called procedure can be repeated with another plugin id.
+ */
+class MODULE_MANAGER_API AsyncPluginValidator final
+{
+   class Impl;
+   std::shared_ptr<Impl> mImpl;
+public:
+   /**
+    * \brief Used to talk back to calling side
+    */
+   class MODULE_MANAGER_API Delegate
+   {
+   public:
+      virtual ~Delegate();
+
+      ///Called for each plugin instance found inside module
+      virtual void OnPluginFound(const PluginDescriptor& plugin) = 0;
+      ///Called when module processing finished
+      virtual void OnValidationFinished() = 0;
+      ///Called on error, further processing is not possible.
+      virtual void OnInternalError(const wxString& msg) = 0;
+   };
+
+   AsyncPluginValidator(AsyncPluginValidator&) = delete;
+   AsyncPluginValidator(AsyncPluginValidator&&) = delete;
+   AsyncPluginValidator& operator=(AsyncPluginValidator&) = delete;
+   AsyncPluginValidator& operator=(AsyncPluginValidator&&) = delete;
+
+   explicit AsyncPluginValidator(Delegate& delegate);
+   ~AsyncPluginValidator();
+
+   /**
+    * \brief Each call to Validate should result in appropriate call
+    * OnValidationFinished, until then it's not allowed to call this
+    * method again.
+    * \param providerId ID of the provider that should be used for validation
+    * \param pluginPath path to the plugin module
+    */
+   void Validate(const wxString& providerId, const wxString& pluginPath);
+};
--- a/libraries/lib-module-manager/PluginHost.cpp
+++ b/libraries/lib-module-manager/PluginHost.cpp
@@ -1,0 +1,220 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file PluginHost.cpp
+
+  @author Vitaly Sverchinsky
+
+  Part of lib-module-manager library
+
+**********************************************************************/
+
+#include "PluginHost.h"
+
+#include <wx/app.h>
+#include <wx/log.h>
+#include <wx/module.h>
+#include <wx/process.h>
+
+#include "BasicUI.h"
+#include "PathList.h"
+#include "FileNames.h"
+#include "ModuleManager.h"
+#include "IPCClient.h"
+#include "PlatformCompatibility.h"
+#include "PluginManager.h"
+
+namespace
+{
+   //Attempts to instantiate plugin module and put plugin descriptors into result
+   void Discover(detail::PluginValidationResult& result, const wxString& providerId, const wxString& pluginPath)
+   {
+      try
+      {
+         if(auto provider = ModuleManager::Get().CreateProviderInstance(providerId, wxEmptyString))
+         {
+            TranslatableString errorMessage{};
+
+            provider->DiscoverPluginsAtPath(pluginPath, errorMessage, [&result](PluginProvider *provider, ComponentInterface *ident)
+            {
+               //Workaround: use DefaultRegistrationCallback to create all descriptors for us
+               //and then put a copy into result
+               auto id = PluginManager::DefaultRegistrationCallback(provider, ident);
+               if(const auto desc = PluginManager::Get().GetPlugin(id))
+                  result.Add(PluginDescriptor { *desc });
+               return id;
+            });
+            if(!errorMessage.empty())
+               result.SetError(errorMessage.Debug());
+         }
+         else
+            result.SetError("provider not found");
+      }
+      catch(...)
+      {
+         result.SetError("unknown error");
+      }
+   }
+}
+
+PluginHost::PluginHost()
+{
+   FileNames::InitializePathList();
+
+   wxFileName configFileName{ FileNames::Configuration() };
+   auto pConfig = std::make_unique<FileConfig>(
+      AppName, wxEmptyString, configFileName.GetFullPath(),
+      wxEmptyString, wxCONFIG_USE_LOCAL_FILE);
+   pConfig->Init();
+   InitPreferences(std::move(pConfig));
+
+   auto& moduleManager = ModuleManager::Get();
+   moduleManager.Initialize();
+   moduleManager.DiscoverProviders();
+
+   mClient = std::make_unique<IPCClient>(*this);
+}
+
+void PluginHost::OnConnect(IPCChannel& channel) noexcept
+{
+   std::lock_guard lck(mSync);
+   mChannel = &channel;
+}
+
+void PluginHost::OnDisconnect() noexcept
+{
+   Stop();
+}
+
+void PluginHost::OnConnectionError() noexcept
+{
+   Stop();
+}
+
+void PluginHost::OnDataAvailable(const void* data, size_t size) noexcept
+{
+   try
+   {
+      mInputMessageReader.ConsumeBytes(data, size);
+      if(mInputMessageReader.CanPop())
+      {
+         {
+            std::lock_guard lck(mSync);
+            assert(!mRequest);
+            mRequest = mInputMessageReader.Pop();
+         }
+         mRequestCondition.notify_one();
+      }
+   }
+   catch(...)
+   {
+      Stop();
+   }
+}
+
+bool PluginHost::Serve()
+{
+   std::unique_lock lck(mSync);
+   mRequestCondition.wait(lck, [this]{ return !mRunning || mRequest.has_value(); });
+
+   if(!mRunning)
+      return false;
+
+   if(mRequest)
+   {
+      std::optional<wxString> request;
+      mRequest.swap(request);
+
+      lck.unlock();
+      
+      wxString providerId;
+      wxString pluginPath;
+      detail::PluginValidationResult result;
+      if(detail::ParseRequestString(*request, providerId, pluginPath))
+         Discover(result, providerId, pluginPath);
+      else
+         result.SetError("malformed request string");
+
+      XMLStringWriter xmlWriter;
+      result.WriteXML(xmlWriter);
+
+      lck.lock();
+      if(mChannel)
+         detail::PutMessage(*mChannel, xmlWriter);
+   }
+
+   return true;
+}
+
+
+void PluginHost::Stop() noexcept
+{
+   try
+   {
+      {
+         std::lock_guard lck(mSync);//may throw
+         mRunning = false;
+         mChannel = nullptr;
+      }
+   }
+   catch(...)
+   {
+      //If something went wrong with mutex locking we'll try to
+      //awake main thread if it's blocked on condition variable.
+      //Attempt to relock the mutex there should throw as well(?)...
+      //which, in turn, will result in std::terminate being called
+   }
+   mRequestCondition.notify_one();
+}
+
+bool PluginHost::Start()
+{
+   const auto cmd = wxString::Format("\"%s\" %s", PlatformCompatibility::GetExecutablePath(), PluginHost::HostArgument);
+
+   auto process = std::make_unique<wxProcess>();
+   process->Detach();
+   if(wxExecute(cmd, wxEXEC_ASYNC, process.get()) != 0)
+   {
+      //process will delete itself upon termination
+      process.release();
+      return true;
+   }
+   return false;
+}
+
+bool PluginHost::IsHostProcess()
+{
+   return wxTheApp && wxTheApp->argc >= 2 && wxStrcmp(wxTheApp->argv[1], HostArgument) == 0;
+}
+
+class PluginHostModule final :
+   public wxModule
+{
+public:
+   DECLARE_DYNAMIC_CLASS(PluginHostModule)
+
+   bool OnInit() override
+   {
+      if(PluginHost::IsHostProcess())
+      {
+         //log messages will appear in a separate window
+         //redirect to log file later
+         wxLog::EnableLogging(false);
+
+         //Handle requests...
+         PluginHost host;
+         while(host.Serve()) { }
+         //...and terminate app
+         return false;
+      }
+      //do noting if current process isn't a host process
+      return true;
+   }
+   
+   void OnExit() override
+   {
+
+   }
+};
+IMPLEMENT_DYNAMIC_CLASS(PluginHostModule, wxModule);
--- a/libraries/lib-module-manager/PluginHost.h
+++ b/libraries/lib-module-manager/PluginHost.h
@@ -1,0 +1,70 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file PluginHost.h
+
+  @author Vitaly Sverchinsky
+
+  Part of lib-module-manager library
+
+**********************************************************************/
+
+#pragma once
+
+#include "IPCChannel.h"
+#include <memory>
+#include <mutex>
+#include <optional>
+#include <condition_variable>
+#include <wx/string.h>
+
+#include "PluginIPCUtils.h"
+
+class IPCClient;
+
+/**
+ * \brief Internal class, processes plugin validation requests
+ * from the main app. Request is a simple string formatted by
+ * detail::MakeRequestString. After connection is established
+ * host starts to wait for a request from server. Once request
+ * is successfully processed host sends a reply. Host is capable
+ * to handle only one request at a time, so it's not allowed to
+ * send another request until host hasn't finish processing
+ * previous request.
+ */
+class PluginHost final : public IPCChannelStatusCallback
+{
+   static constexpr auto HostArgument =  "--host";
+
+   std::unique_ptr<IPCClient> mClient;
+   IPCChannel* mChannel{nullptr};
+   detail::InputMessageReader mInputMessageReader;
+   std::mutex mSync;
+   std::condition_variable mRequestCondition;
+   std::optional<wxString> mRequest;
+
+   bool mRunning{true};
+
+   void Stop() noexcept;
+
+public:
+   /**
+    * \brief Attempts to start a host application (should be called from
+    * the main application)
+    * \return true if host has started successfully
+    */
+   static bool Start();
+
+   ///Returns true if current process is considered to be a plugin host process
+   static bool IsHostProcess();
+
+   PluginHost();
+
+   void OnConnect(IPCChannel& channel) noexcept override;
+   void OnDisconnect() noexcept override;
+   void OnConnectionError() noexcept override;
+   void OnDataAvailable(const void* data, size_t size) noexcept override;
+
+   bool Serve();
+};
--- a/libraries/lib-module-manager/PluginIPCUtils.cpp
+++ b/libraries/lib-module-manager/PluginIPCUtils.cpp
@@ -1,0 +1,159 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file PluginIPCUtils.cpp
+
+  @author Vitaly Sverchinsky
+  
+  Part of lib-module-manager library.
+
+**********************************************************************/
+
+#include "PluginIPCUtils.h"
+
+#include <cstring>
+
+#include "PluginDescriptor.h"
+
+#include "IPCChannel.h"
+#include "XMLWriter.h"
+
+using namespace detail;
+
+namespace {
+   using HeaderBlock = size_t;
+   constexpr auto HeaderBlockSize = sizeof(HeaderBlock);
+
+   constexpr auto NodePlugin = "Plugin";
+   constexpr auto NodeError = "Error";
+   constexpr auto AttrErrorMessage = "msg";
+}
+
+bool detail::ParseRequestString(const wxString& req, wxString& providerId, wxString& pluginPath)
+{
+   auto strings = wxSplit(req, ';');
+   if(strings.size() == 2)
+   {
+      providerId = strings[0];
+      pluginPath = strings[1];
+      return true;
+   }
+   return false;
+}
+
+wxString detail::MakeRequestString(const wxString& providerId, const wxString& pluginPath)
+{
+   return wxJoin(wxArrayStringEx {providerId, pluginPath}, ';');
+}
+
+void detail::PutMessage(IPCChannel& channel, const wxString& value)
+{
+   auto utf8 = value.ToUTF8();
+   const HeaderBlock length = utf8.length();
+   channel.Send(&length, HeaderBlockSize);
+   channel.Send(utf8.data(), length);
+}
+
+void InputMessageReader::ConsumeBytes(const void* bytes, size_t length)
+{
+   auto from = mBuffer.size();
+   mBuffer.resize(from + length);
+   std::memcpy(&mBuffer[from], bytes, length);
+}
+
+bool InputMessageReader::CanPop() const noexcept
+{
+   if(mBuffer.size() >= HeaderBlockSize)
+   {
+      auto expectedSize = *reinterpret_cast<const HeaderBlock*>(mBuffer.data());
+      return mBuffer.size() >= expectedSize + HeaderBlockSize;
+   }
+   return false;
+}
+
+wxString InputMessageReader::Pop()
+{
+   wxString message;
+   const auto length = *reinterpret_cast<const HeaderBlock*>(mBuffer.data());
+   assert(mBuffer.size() >= length + HeaderBlockSize);
+   if(length > 0)
+      message = wxString::FromUTF8(mBuffer.data() + HeaderBlockSize, length);
+   mBuffer.erase(mBuffer.begin(), mBuffer.begin() + HeaderBlockSize + length);
+   return message;
+}
+
+bool PluginValidationResult::IsValid() const noexcept
+{
+   return !mHasError && !mDescriptors.empty();
+}
+
+bool PluginValidationResult::HasError() const noexcept
+{
+   return mHasError;
+}
+
+const wxString& PluginValidationResult::GetErrorMessage() const noexcept
+{
+   return mErrorMessage;
+}
+
+void PluginValidationResult::Add(PluginDescriptor&& desc)
+{
+   mDescriptors.push_back(std::move(desc));
+}
+
+void PluginValidationResult::SetError(const wxString& msg)
+{
+   mHasError = true;
+   mErrorMessage = msg;
+}
+
+const std::vector<PluginDescriptor>& PluginValidationResult::GetDescriptors() const noexcept
+{
+   return mDescriptors;
+}
+
+bool PluginValidationResult::HandleXMLTag(const std::string_view& tag, const AttributesList& attrs)
+{
+   if(tag == NodeError)
+   {
+      mHasError = true;
+      for(auto& p : attrs)
+      {
+         auto key = wxString(p.first.data(), p.first.length());
+         auto& attr = p.second;
+         if(key == AttrErrorMessage)
+            mErrorMessage = attr.ToWString();
+      }
+   }
+   return true;
+}
+void PluginValidationResult::HandleXMLEndTag(const std::string_view&) { }
+
+XMLTagHandler* PluginValidationResult::HandleXMLChild(const std::string_view& tag)
+{
+   if(tag == PluginDescriptor::XMLNodeName)
+   {
+      mDescriptors.resize(mDescriptors.size() + 1);
+      return &mDescriptors.back();
+   }
+   return nullptr;
+}
+
+void PluginValidationResult::WriteXML(XMLWriter& writer) const
+{
+   if(mHasError)
+   {
+      writer.StartTag(NodeError);
+      writer.WriteAttr(AttrErrorMessage, mErrorMessage);
+      writer.EndTag(NodeError);
+   }
+   if(!mDescriptors.empty())
+   {
+      writer.StartTag(NodePlugin);
+      for(auto& desc : mDescriptors)
+         desc.WriteXML(writer);
+      writer.EndTag(NodePlugin);
+   }
+}
--- a/libraries/lib-module-manager/PluginIPCUtils.h
+++ b/libraries/lib-module-manager/PluginIPCUtils.h
@@ -1,0 +1,85 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file PluginIPCUtils.h
+
+  @author Vitaly Sverchinsky
+
+  @brief Provides a set of internal commonly used functions/structures
+
+  Part of lib-module-manager library.
+
+**********************************************************************/
+
+#pragma once
+
+#include <vector>
+#include <wx/string.h>
+#include "XMLTagHandler.h"
+
+class IPCChannel;
+class PluginDescriptor;
+class XMLWriter;
+
+namespace detail
+{
+   ///Helper function that extracts plugin provider id and plugin path from the item string
+   ///return true if item string was successfully parsed
+   bool ParseRequestString(const wxString& req, wxString& providerId, wxString& pluginPath);
+
+   ///Helper function that builds item string from given plugin provider id and plugin path strings
+   ///return Item string that can be passed as argument to wxConnection::Request
+   wxString MakeRequestString(const wxString& providerId, const wxString& pluginPath);
+
+   ///Writes the length of the string and string bytes into the channel.
+   ///Message can be then extracted with InputMessageReader
+   void PutMessage(IPCChannel& channel, const wxString& value);
+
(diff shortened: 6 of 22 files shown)


src/effects/VST/VSTEffect.cpp
```diff
@@ -166,41 +166,8 @@ DECLARE_BUILTIN_PROVIDER(VSTBuiltin);
 /// Auto created at program start up, this initialises VST.
 ///
 ///////////////////////////////////////////////////////////////////////////////
-class VSTSubEntry final : public wxModule
-{
-public:
-   bool OnInit()
-   {
-      // Have we been started to check a plugin?
-      if (wxTheApp && wxTheApp->argc == 3 && wxStrcmp(wxTheApp->argv[1], VSTCMDKEY) == 0)
```
Paul-Licameli, line 175: Unused constant VSTCMDKEY no longer needs a definition

src/effects/VST/VSTEffect.cpp
```diff
@@ -166,41 +166,8 @@ DECLARE_BUILTIN_PROVIDER(VSTBuiltin);
...
-
-   DECLARE_DYNAMIC_CLASS(VSTSubEntry)
-};
-IMPLEMENT_DYNAMIC_CLASS(VSTSubEntry, wxModule);
-
-//----------------------------------------------------------------------------
-// VSTSubProcess
-//----------------------------------------------------------------------------
-#define OUTPUTKEY wxT("<VSTLOADCHK>-")
 enum InfoKeys
```
Paul-Licameli, line 171: Remove this enum with VSTSubProcess

src/effects/EffectManager.cpp
```diff
@@ -111,6 +111,9 @@ TranslatableString EffectManager::GetCommandName(const PluginID & ID)
 
 TranslatableString EffectManager::GetEffectFamilyName(const PluginID & ID)
 {
+   if(auto description = PluginManager::Get().GetPlugin(ID))
+      return TranslatableString { description->GetEffectFamily(), {} };
+
```
Paul-Licameli, line 116: My recent changes in EffectManager may make the benefits of these changes less but still not zero.  After my changes, instance initialization is avoided, but still not the plugin initialization.

libraries/lib-module-manager/PluginManager.h
```diff
@@ -174,9 +37,45 @@ typedef wxArrayString PluginIDs;
...
 {
 public:
+   class MODULE_MANAGER_API PluginCheckDelegate
+   {
+   public:
+      virtual ~PluginCheckDelegate();
+
+      virtual void OnPluginValidationStarted(const wxString& path, int index, int count) = 0;
+      virtual void OnPluginValidationFinished() = 0;
+      virtual void OnPluginValidationResult(const wxString& path, bool valid) = 0;
```
Paul-Licameli, line 69: I don't believe in using `noexcept` too widely, but functions meant to be called inside a `finally` and therefore inside a destructor should have it.

Paul-Licameli: That applies to `OnPluginValidationFinished `

dozzzzer:
Some issues were observed when testing the solution.

- #2808
- #2809
- #2810

libraries/lib-ipc/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+#[[
+]]#
```
Paul-Licameli, line 2: Second '#' isn't needed.

Add a few words about the general purpos of these files.

libraries/lib-ipc/IPCChannel.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <cstddef>
+
+/**
+ * \brief Interface for sending data from client to server or vise versa,
```
Paul-Licameli, line 6: "vice versa"

libraries/lib-ipc/IPCChannel.h
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+   virtual ~IPCChannel();
+   /**
+    * \brief Write data to the channel
+    * \param bytes Pointer to the data
+    * \param length Number of bytes to send
+    */
+   virtual void Send(const void* bytes, size_t length) = 0;
+};
```
Paul-Licameli, line 19: When you need an abstract base class with just one virtual function besides the destructor, an alternative is just to generate the type you need from std::function:

```
using IPCChanel = std::function< void(const void *bytes, size_t length) >;
```

It's your choice whether you prefer that suggestion here.

libraries/lib-ipc/IPCChannel.h
```diff
@@ -0,0 +1,49 @@
...
+
+   /**
+    * \brief Called when connection attempt fails.
+    */
+   virtual void OnConnectionError() = 0;
+   /**
+    * \brief Called when connection established.
+    * \param channel Using this channel client or server can send data to the other side.
+    */
+   virtual void OnConnect(IPCChannel* channel) = 0;
```
Paul-Licameli, line 37: Is the argument never null?

If so, I would prefer to use a reference to say what I mean with types, leaving no doubt in the reader's mind about that, while needing no extra comments either.

libraries/lib-ipc/IPCClient.h
```diff
@@ -0,0 +1,25 @@
...
+   class Impl;
+   std::unique_ptr<Impl> mImpl;
+public:
+   IPCClient();
+   ~IPCClient();
+
+   /**
+    * \brief Attempts to connect to a server. It is expected that server already exists.
+    * \param id TCP socket port number.
+    * \param callback See descriotion of the callback parameter in IPCServer::CreateChannel.
```
Paul-Licameli, line 22: "description"

libraries/lib-ipc/internal/spinlock.h
```diff
@@ -0,0 +1,26 @@
...
+class spinlock
+{
+   std::atomic_flag flag = ATOMIC_FLAG_INIT;
+public:
+   ~spinlock() { }
+
+   void lock()
+   {
+      for(unsigned i = 0; flag.test_and_set(std::memory_order_acquire); ++i)
+         if(i & 1) std::this_thread::yield();
```
Paul-Licameli, line 19: This yields only every 2 failed tests.  I'm curious, why 2, not 1, or some other number.  Did some article recommend that?

libraries/lib-ipc/internal/socket_guard.h
```diff
@@ -0,0 +1,26 @@
...
+
+class socket_guard final
+{
+   SOCKET mSocket{INVALID_SOCKET};
+public:
+   socket_guard(const socket_guard&) = delete;
+   socket_guard& operator=(const socket_guard&) = delete;
+   socket_guard& operator=(socket_guard&&) = delete;
+
+   socket_guard(SOCKET sock) noexcept : mSocket(sock) { }
```
Paul-Licameli, line 13: `explicit` ?

libraries/lib-ipc/internal/ipc-types.h
```diff
@@ -0,0 +1,18 @@
...
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <unistd.h>
+#include <poll.h>
+
+#define INVALID_SOCKET -1
+#define SOCKET_ERROR -1
+#define SOCKET int
+#define CLOSE_SOCKET close
+#define NFDS(x) (x + 1)
```
Paul-Licameli, line 17: Would a more careful definition be
```
#define NFDS(x) ((x) + 1)
```

libraries/lib-ipc/internal/socket_guard.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "ipc-types.h"
+
+class socket_guard final
```
Paul-Licameli, line 5: So I read the class to understand that it's a move-only object for RAII cleanup when a socket is opened.

Comment that before the class.  Perhaps too, "unique_socket_guard" would be a name that suggests that it is move-only.

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+class BufferedIPCChannel final : public IPCChannel
+{
+   spinlock mSync;
+
+   IPCChannelStatusCallback* mCallback;
+
+   SOCKET mSocket;
+   
+   std::vector<char> mOutputBuffer;
```
Paul-Licameli, line 23: Perhaps comment
`// TODO C++20: use std::byte`

libraries/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ set( LIBRARIES
    lib-preferences
    lib-math
    lib-files
+   lib-ipc
```
Paul-Licameli, line 16: This line should be added in the same commit that defines the source code files in lib-pic.  Not later.

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,38 @@
...
+   
+   std::vector<char> mInputBuffer;
```
Paul-Licameli, line 25: comment likewise

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,38 @@
...
+   
+public:
+   //Channel constructor. Does not own objects passed as arguments
+   BufferedIPCChannel(SOCKET socket, IPCChannelStatusCallback* callback);
```
Paul-Licameli, line 29: Can the second argument be a reference?

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * \brief Implementation of IPCChannel that uses intermediate
+ * buffer for data exchange between client and server.
+ */
+class BufferedIPCChannel final : public IPCChannel
+{
+   spinlock mSync;
+
+   IPCChannelStatusCallback* mCallback;
```
Paul-Licameli, line 19: Can this be a const reference or a `*const`?

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,38 @@
...
+
+   SOCKET mSocket;
```
Paul-Licameli, line 21: Can this member be const?
It's always good to add const when you can to help the reader understand what parts never vary after construction.

libraries/lib-ipc/internal/BufferedIPCChannel.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <mutex>
+#include <cstring>
+
+BufferedIPCChannel::BufferedIPCChannel(SOCKET socket, IPCChannelStatusCallback* callback)
+   : mCallback(callback), mSocket(socket)
+{
+   mInputBuffer.resize(2048);
+   mOutputBuffer.reserve(2048);
+
+   callback->OnConnect(this);
```
Paul-Licameli, line 12: Indeed this assumes callback isn't null, so I would rather have a reference argument then

cmake-proxies/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ endif()
 
 add_conan_lib(
    wxWidgets
-   wxwidgets/3.1.3.2-audacity
+   wxwidgets/3.1.3.3-audacity
```
Paul-Licameli, line 60: What feature in wxWidgets was needed?  Say in the commit comment

libraries/lib-xml/XMLWriter.cpp
```diff
@@ -459,11 +459,17 @@ void XMLUtf8BufferWriter::EndTag(const std::string_view& name)
    }
 }
 
+void XMLUtf8BufferWriter::WriteAttr(const std::string_view& name, const wxString& value)
+{
+   const auto utf8 = value.ToUTF8();
```
Paul-Licameli, line 464: ToUTF8 used here, utf8_str used instead in the following function -- made me check in the wxWidgets headers to be sure they were not really different.

Why not simply use utf8_str as in the existing function making the differences between the two functions less?

libraries/lib-module-manager/PluginDescriptor.h
```diff
@@ -0,0 +1,157 @@
...
+   PluginTypeModule=1<<5,
+} PluginType;
+
+class PluginDescriptorXMLTagHandler;
+
+// TODO:  Convert this to multiple derived classes
+//! Represents either a PluginProvider or a loaded plug-in and caches some
+//! information about it
+class MODULE_MANAGER_API PluginDescriptor
+{
```
Paul-Licameli, line 39: While you are touching this class, please do these things too:

- Use in-class member initializers so that the body of the constructor shrinks to nothing.
- Make sure no other members of non-class types are uninitialized.
- Remove the friend declaration, because long-distance friendships are bad for understanding code.  Make Load() and Set() public so you can.

libraries/lib-module-manager/PluginManager.h
```diff
@@ -20,148 +20,11 @@
 
 #include "EffectInterface.h"
 #include "PluginInterface.h"
```
Paul-Licameli, line 22: This include becomes redundant when PluginDescriptor.h is separated

libraries/lib-ipc/IPCServer.h
```diff
@@ -0,0 +1,35 @@
...
+    * \brief Creates channel and starts listening for the incoming connection.
+    * \param id Basically is a tcp port number used to establish a connection
+    * \param callback In case of successful connection callback should exist
+    * until IPCChannelStatusCallback::OnDisconnect or
+    * IPCChannelStatusCallback::OnConnectionError called or server instance
+    * destroyed.
+    * \return true if channel was successfully created and waiting for incoming
+    * connection
+    */
+   bool CreateChannel(int id, IPCChannelStatusCallback* callback);
```
Paul-Licameli, line 34: - callback is assumed non-null inside the function -- therefore, make it a reference argument
- id is a "naked" int.  I would write a small ChannelId class wrapping the int and preventing implicit conversions to and from int.  Probably it would need operators == and != and maybe < (or just <=> if we had C++20)

libraries/lib-ipc/IPCServer.cpp
```diff
@@ -0,0 +1,116 @@
...
+      addrin.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+      addrin.sin_port = htons(static_cast<u_short>(id));
+
+      if(bind(serverfd, reinterpret_cast<const sockaddr*>(&addrin), sizeof(addrin)) == SOCKET_ERROR)
+         return false;
+
+      if(listen(serverfd, 1) == SOCKET_ERROR)
+         return false;
+
+      mWorker = std::make_unique<std::thread>([this, id, callback, serverfd, guard = std::move(serverfd_guard)]()
```
Paul-Licameli, line 46: If channel becomes a reference instead of pointer, remember to pass `std::ref(channel)` to avoid the copy constructor.

libraries/lib-ipc/IPCServer.cpp
```diff
@@ -0,0 +1,116 @@
...
+      auto serverfd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+      if(serverfd == INVALID_SOCKET)
+         return false;
+
+      socket_guard serverfd_guard { serverfd };
+
+      sockaddr_in addrin {};
+      addrin.sin_family = AF_INET;
+      addrin.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+      addrin.sin_port = htons(static_cast<u_short>(id));
```
Paul-Licameli, line 38: Is there no "un-binding" needed later for INET?

libraries/lib-ipc/IPCServer.cpp
```diff
@@ -0,0 +1,116 @@
...
+            if(!mAlive)
+               //do not notify
+               return;
+            if(res == 0)
+            {
+               std::this_thread::yield();
+               continue;
+            }
+            if (res == 1)
+               clientfd = accept(serverfd, nullptr, nullptr);
```
Paul-Licameli, line 75: `man poll` tells me that -1 is a possible return value, indicating an error, stored in `errno` (which is a macro, not really a variable, on my system:  I suppose they addressed the thread safety problem).

So please make the check more exhaustive.

libraries/lib-ipc/IPCClient.cpp
```diff
@@ -0,0 +1,77 @@
...
+   }
+};
+
+IPCClient::IPCClient()
+{
+#ifdef _WIN32
+   WSADATA wsaData;
+   auto result = WSAStartup(MAKEWORD(2, 2), &wsaData);
+   if (result == NO_ERROR)
+      mImpl = std::make_unique<Impl>();
```
Paul-Licameli, line 65: On Windows this is a possibly-failing constructor but there is no indication to the caller of the failure.  Either you need a boolean method to test for success, or should throw an exception.

Same applies to the server, on Windows only.

libraries/lib-ipc/IPCClient.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <stdexcept>
+
+#include "internal/ipc-types.h"
+#include "internal/socket_guard.h"
+#include "internal/BufferedIPCChannel.h"
+
+class IPCClient::Impl
+{
+   std::atomic<bool> mAlive{true};
+   std::unique_ptr<std::thread> mWorker;
```
Paul-Licameli, line 15: ditto question use of `unique_ptr`

libraries/lib-ipc/IPCClient.cpp
Paul-Licameli: ... Oh I see,  but then CreateChannel() can indicate error.  OK then.

libraries/lib-ipc/IPCServer.h
```diff
@@ -0,0 +1,35 @@
...
+
+class IPCChannel;
+class IPCChannelStatusCallback;
+
+/**
+ * \brief Simple TCP socket based ipc server.
+ * Currently, implementation allows only one channel per
+ * instance. IPCChannelStatusCallback methods are always called
+ * from the worker thread, though IPCChannel instance created
+ * by this implementation is thread safe (considering lifetime limitations)
```
Paul-Licameli, line 13: The IPCServer object itself is not thread safe -- I mean `CreateChannel` could race to assign `Impl::mWorker` if two threads called it.  Maybe that is worth a comment.

libraries/lib-ipc/IPCClient.h
```diff
@@ -0,0 +1,25 @@
...
+    */
+   void CreateChannel(int id, IPCChannelStatusCallback* callback);
```
Paul-Licameli, line 24: As with server, this initialization is not itself a thread safe function

libraries/lib-ipc/IPCClient.cpp
```diff
@@ -0,0 +1,77 @@
...
+   {
+      mAlive = false;
+      if(mWorker)
+         mWorker->join();
+   }
+
+   void CreateChannel(int id, IPCChannelStatusCallback* callback)
+   {
+      if(mWorker)
+         throw std::runtime_error("cannot handle more than one channel");
```
Paul-Licameli, line 28: Return void and throw, or return boolean for error?  You made inconsistent choices for client and server.  And client, unlike server, makes no error if mImpl failed initialization on Windows.

I think it is eaiser to make both return boolean and not throw.

libraries/lib-ipc/IPCClient.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+      mWorker = std::make_unique<std::thread>([this, id, callback]()
```
Paul-Licameli, line 30: Ditto use of a reference for the callback, and `std::ref`

libraries/lib-ipc/IPCServer.cpp
Paul-Licameli: Argument `id` is not used

libraries/lib-ipc/IPCChannel.h
Paul-Licameli: In fact the unique use of OnConnect in BufferedIPCChannel.cpp is always passing non-null (`this`).

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -577,6 +441,29 @@ PluginManager::~PluginManager()
    Terminate();
 }
 
+void PluginManager::InitializePlugins()
```
Paul-Licameli, line 444: Too much is happening in the commit "PluginValidator interface" with too few code and commit comments.  I am having trouble understanding it.

The second line of the commit comment says "Code path that is used with "bFast" flag in CheckForUpdates was extracted into InitializePlugins function" and that explains this function.

It might help to separate a commit that does only this part.  Perhaps "RevalidatePlugins" is also a clearer name for what this does.

libraries/lib-module-manager/PluginManager.h
```diff
@@ -174,9 +37,46 @@ typedef wxArrayString PluginIDs;
 
 class PluginRegistrationDialog;
 
+class MODULE_MANAGER_API PluginValidator
```
Paul-Licameli, line 40: Four new abstract classes are introduced in the commit "PluginValidator interface" but I'm not sure what they really do, or how they are meant to be used, from reading their names alone and from the lack of comments.  I must study other code to understand.

src/PluginHost.cpp
```diff
@@ -0,0 +1,585 @@
...
+   bool OnInit() override
+   {
+      if(wxTheApp && wxTheApp->argc >= 2 && wxStrcmp(wxTheApp->argv[1], HostArgument) == 0)
+      {
+         //log messages will appear in a separate window
+         //redirect to log file later
+         wxLog::EnableLogging(false);
+
+         PluginHost host;
+         while(host.Serve()) { }
```
Paul-Licameli, line 568: I am beginning to understand this new file.  If you are here, you are in the main thread of a child process.  The Serve() loop detects requests and sends responses, while the overridden functions do work on another thread.

Nothing ever blocks inside Serve(), which makes it a busy loop, in the admittedly unlikely event that it handles its requests faster than the responses arrive.  Whereas `OnDataAvailable` contains an assertion that is not provable from what I read here, and migh fail with unusual timing, unless there is more I don't understand yet about the main process.

I would be more satisfied to see:

- a std::mutex and associated std::condition_variable.
- Serve() waits https://en.cppreference.com/w/cpp/thread/condition_variable/wait for the condition: `!running || (request && channel)`
- OnConnectionError, OnDisconnect, an OnDataAvailable may make the condition true, so they call https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one
- Serve() releases the mutex only after moving from mRequest.
- OnDataAvailable holds the mutex while assigning to mRequest
  - After taking the mutex, it also waits on the condition variable for `!mRequest`, which really "proves" it
  - OnDisconnect, which can make the condition true, also notifies the variable

Paul-Licameli (changes requested):

src/PluginHost.cpp
```diff
@@ -0,0 +1,585 @@
...
+                  xmlWriter.StartTag(NodePluginDescriptor);
+                  xmlWriter.WriteAttr(AttrID, desc->GetID());
+                  xmlWriter.WriteAttr(AttrType, static_cast<int>(desc->GetPluginType()));
+                  xmlWriter.WriteAttr(AttrEnabled, true);
+                  xmlWriter.WriteAttr(AttrValid, true);
+                  xmlWriter.WriteAttr(AttrProviderID, desc->GetProviderID());
+                  xmlWriter.WriteAttr(AttrPath, desc->GetPath());
+                  xmlWriter.WriteAttr(AttrName, desc->GetSymbol().Internal());
+                  xmlWriter.WriteAttr(AttrVendor, desc->GetVendor());
+                  xmlWriter.WriteAttr(AttrVersion, desc->GetVendor());
```
Paul-Licameli, line 208: should be `desc->GetUntranslatedVersion()`

Paul-Licameli (changes requested):

libraries/lib-preferences/FileConfig.cpp
```diff
@@ -37,7 +37,7 @@ FileConfig::FileConfig(const wxString& appName,
    mLocalFilename(localFilename),
    mGlobalFilename(globalFilename),
    mStyle(style),
-   mConv(conv),
+   mConv(conv.Clone()),//pass to Init() instead?
```
Paul-Licameli, line 40: Commit comment misspelled "instance"

libraries/lib-module-manager/PluginHost.h
```diff
@@ -0,0 +1,64 @@
+/**********************************************************************
```
Paul-Licameli, line 1: Commit comment misspelled "stability"

libraries/lib-module-manager/PluginIPCUtils.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include <list>
+#include <wx/string.h>
+#include "XMLTagHandler.h"
+
+class IPCChannel;
+class PluginDescriptor;
+class XMLWriter;
+
+namespace details
```
Paul-Licameli, line 25: Is the convention `details` or `detail`?  I used `detail`

libraries/lib-module-manager/PluginIPCUtils.h
```diff
@@ -0,0 +1,84 @@
...
+{
+   //Helper function that extracts plugin provider id and plugin path from the item string
+   //return true if item string was successfully parsed
+   bool ParseRequestString(const wxString& req, wxString& providerId, wxString& pluginPath);
+
+   //Helper function that build item string from given plugin provider id and plugin path strings
```
Paul-Licameli, line 31: "builds"

libraries/lib-module-manager/PluginIPCUtils.h
```diff
@@ -0,0 +1,84 @@
...
+#include <wx/string.h>
+#include "XMLTagHandler.h"
+
+class IPCChannel;
+class PluginDescriptor;
+class XMLWriter;
+
+namespace details
+{
+   //Helper function that extracts plugin provider id and plugin path from the item string
```
Paul-Licameli, line 27: All these comments might use Doxygen format
`//! ...`
or
`
/*!
...
*/
`

libraries/lib-module-manager/PluginIPCUtils.cpp
```diff
@@ -0,0 +1,157 @@
...
+      return mBuffer.size() >= expectedSize + HeaderBlockSize;
+   }
+   return false;
+}
+
+wxString InputMessageReader::Pop()
+{
+   wxString message;
+   const auto length = *reinterpret_cast<const HeaderBlock*>(mBuffer.data());
+   assert(mBuffer.size() >= length + HeaderBlockSize);
```
Paul-Licameli, line 77: An assertion without proof -- but sufficient "proof" for me would be that you document in the header file,

```
//! @pre `CanPop()`
```

So the burden of further proof is properly on the caller.

libraries/lib-module-manager/PluginIPCUtils.h
```diff
@@ -0,0 +1,84 @@
...
+   //as a message
+   class PluginValidationResult final : public XMLTagHandler
+   {
+      std::list<PluginDescriptor> mDescriptors;
+      wxString mErrorMessage;
+      bool mHasError{false};
+   public:
+
+      //Result considered to be valid if error flag isn't set or
+      //it does not contain any plugin descriptor.
```
Paul-Licameli, line 66: This comment disagrees with the implementation, and I think you want to change the comment.

Result is valid if and only if there is no error and there is at least one descriptor.

libraries/lib-module-manager/PluginIPCUtils.cpp
```diff
@@ -0,0 +1,157 @@
...
+}
+
+const wxString& PluginValidationResult::GetErrorMessage() const noexcept
+{
+   return mErrorMessage;
+}
+
+void PluginValidationResult::Add(PluginDescriptor&& desc)
+{
+   mDescriptors.push_back(desc);
```
Paul-Licameli, line 101: Why not `std::move(desc)`?  You need it because `desc` alone still is an lvalue reference even though the argument is declared with `&&`.

And that might allow vector instead of list for the container, but maybe there is a reason for list I don't understand yet, some requirement of non-relocation of the memory.

libraries/lib-module-manager/PluginIPCUtils.cpp
```diff
@@ -0,0 +1,157 @@
...
+bool PluginValidationResult::HandleXMLTag(const std::string_view& tag, const AttributesList& attrs)
+{
+   if(tag == NodeError)
+   {
+      mHasError = true;
+      for(auto& p : attrs)
+      {
+         auto key = wxString(p.first.data(), p.first.length());
+         auto& attr = p.second;
+         if(key == "msg")
```
Paul-Licameli, line 124: Duplicated string literal

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+            {
+               //Workaround: use DefaultRegistrationCallback to create all descriptors for us
+               //and then put a copy into result
+               auto id = PluginManager::DefaultRegistrationCallback(provider, ident);
+               if(const auto desc = PluginManager::Get().GetPlugin(id))
+                  result.Add(PluginDescriptor { *desc });
+               return id;
+            });
+            if(!errorMessage.empty())
+               result.SetError(errorMessage.Translation());
```
Paul-Licameli, line 49: Now I'm wondering about the proper internationalization of error messages.  Does `Translation` really work, or has the host process not intiailized translations before this point is reached?  If it doesn't, maybe `Translation()` can be applied to the msgid on the host side, if the `TranslatableString` is a simple one without format arguments.

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+         }
+         else
+            result.SetError("provider not found");
+      }
+      catch(...)
+      {
+         result.SetError("unknown error");
```
Paul-Licameli, line 56: And, two new non-internationalized messages.

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+PluginHost::PluginHost()
+{
+   FileNames::InitializePathList();
+
+   wxFileName configFileName{ FileNames::Configuration() };
+   auto pConfig = std::make_unique<FileConfig>(
+      AppName, wxEmptyString, configFileName.GetFullPath(),
+      wxEmptyString, wxCONFIG_USE_LOCAL_FILE);
+   pConfig->Init();
+   InitPreferences(std::move(pConfig));
```
Paul-Licameli, line 70: In AudacityApp.cpp, `PopulatePreferences` follows `InitPreferences` and that is where translation is initialized.

That function, or pieces of it, should be moved out of AudacityApp.cpp and demoted into one of the lib-s, so it can be used here too, resolving the doubts about `Translation()` I mentioned before.

Paul-Licameli (changes requested):

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+      {
+         std::lock_guard lck(mSync);//may throw
+         mRunning = false;
+         mChannel = nullptr;
+      }
+   }
+   catch(...)
+   {
+      //If something went wrong with mutex locking we'll try to
+      //awake main thread if it's blocked on conditinal variable.
```
Paul-Licameli, line 164: "condition"

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+      lck.unlock();
+      
+      wxString providerId;
+      wxString pluginPath;
+      details::PluginValidationResult result;
+      if(details::ParseRequestString(*request, providerId, pluginPath))
+         Discover(result, providerId, pluginPath);
+      else
+         result.SetError("malformed request string");
```
Paul-Licameli, line 137: Another non-translated string

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+   {
+      std::optional<wxString> request;
+      mRequest.swap(request);
+
+      lck.unlock();
+      
+      wxString providerId;
+      wxString pluginPath;
+      details::PluginValidationResult result;
+      if(details::ParseRequestString(*request, providerId, pluginPath))
```
Paul-Licameli, line 134: This implies that this server expects messages as formatted by `MakeRequestString`.  Then that should be commented in the header file.

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+void PluginHost::OnDataAvailable(const void* data, size_t size) noexcept
+{
+   try
+   {
+      mInputMessageReader.ConsumeBytes(data, size);
+      if(mInputMessageReader.CanPop())
+      {
+         {
+            std::lock_guard lck(mSync);
+            assert(!mRequest);
```
Paul-Licameli, line 104: An assertion without proof.  To fix it, improve documentation to explain that a request must be sent only after responses of all previous requests have been received by the client -- otherwise, requests may be lost:  they don't queue up.

That makes this class more understandable without reading the code that uses the class -- and that code doesn't exist yet, at the commit introducing this.

libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+   auto process = std::make_unique<wxProcess>();
+   process->Detach();
+   if(wxExecute(cmd, wxEXEC_ASYNC, process.get()) != 0)
+   {
+      //process will delete itself upon termination
+      process.release();
+      return true;
+   }
+   return false;
```
Paul-Licameli, line 183: Maybe it won’t be difficult to build a small executable that links only this library and its dependencies.  Avoid wxModule and wasteful loading of another Audacity with all of its static initializations.

Not essential to making this PR.

libraries/lib-module-manager/AsyncPluginValidator.h
```diff
@@ -0,0 +1,64 @@
...
+#pragma once
+
+#include <wx/string.h>
+#include <memory>
+
+class PluginDescriptor;
+
+
+/**
+ * \brief Used to perform plugin validation in a dedicated process.
```
Paul-Licameli, line 22: Maybe less misleading to write, "Starts and communicates with a dedicated process to perform plugin validation".

But the member functions of it are available to call in the main process.

libraries/lib-module-manager/AsyncPluginValidator.cpp
```diff
@@ -0,0 +1,237 @@
...
+   Impl& operator=(Impl&) = delete;
+   Impl& operator=(Impl&&) = delete;
+
+   Impl(Delegate& delegate) : mDelegate(&delegate) { }
+
+   ~Impl() override
+   {
+      //important to reset delegate before IPCChannelStatusCallback::OnDisconnect
+      //is called by server
+      mDelegate = nullptr;
```
Paul-Licameli, line 154: A thread join happens at the closing brace, so callbacks to members may happen before that completes, making mDelegate null matters while joining.  Am I right?  It may help to explain a little more.

libraries/lib-module-manager/AsyncPluginValidator.cpp
```diff
@@ -0,0 +1,237 @@
...
+            {
+               if(mDelegate == nullptr)
+                  return;
+
+               //Release the current value to keep state invariant
+               //Caller is free to use Validate now
+               std::optional<wxString> request;
+               {
+                  std::lock_guard lck_sync(mSync);
+                  mRequest.swap(request);
```
Paul-Licameli, line 97: If you really want to minimize the time inside the lock_guard:  Note that swap of an optional with another empty wone does perform a move-construction, see https://en.cppreference.com/w/cpp/utility/optional/swap ...

BUT for wxString that performs a string copy anyway, because wxWidgets is an old library not defining move constructors.

You can improve this using `optional<std::string>` or you can use the member function `wxString::swap` on *mRequest.  (Then `mRequest.has_value()` is still true, it has an empty string in it; so `mRequest.reset()` would be needed.)

Paul-Licameli: You also used `optional<wxString>` in PluginHost and might change it.

libraries/lib-module-manager/AsyncPluginValidator.cpp
```diff
@@ -0,0 +1,237 @@
...
+         mServer = std::move(server);
+   }
+   
+   void HandleInternalError(const wxString& msg) noexcept
+   {
+      try
+      {
+         BasicUI::CallAfter([this, self = weak_from_this(), msg]
+         {
+            if(auto lck = self.lock() && mDelegate != nullptr)
```
Paul-Licameli, line 71: What you wrote is equivalent to
```
bool lck = (!self.expired() && mDelegate)
```

but does not lock the weak pointer for the intended scope!  It should be
```
if (auto lck = self.lock(); lck && mDelegate)
```

Paul-Licameli (changes requested):

libraries/lib-module-manager/AsyncPluginValidator.h
```diff
@@ -0,0 +1,64 @@
...
+    * \brief Used to talk back to calling side
+    */
+   class MODULE_MANAGER_API Delegate
+   {
+   public:
+      virtual ~Delegate();
+
+      ///Called for each plugin instance found inside module
+      virtual void OnPluginFound(const PluginDescriptor& plugin) = 0;
+      ///Called when module processing finished
```
Paul-Licameli, line 42: This means, handling of one request is complete, and another may be sent.

It is very pleasing to see only one virtual function interface here, and the simplified implementation without the state machine and a switch.  I find it more understandable than the first draft.

libraries/lib-module-manager/AsyncPluginValidator.cpp
Paul-Licameli: I would also write the lambda capture without this, only with self to access this.  And that would have caused a compilation error after the incorrect if, which might make you fix it correctly.

src/PluginStartupRegistration.cpp
```diff
@@ -0,0 +1,81 @@
+/**********************************************************************
```
Paul-Licameli, line 1: The sixth commit, which introduces this file, fails to compile.  It requires the overload of `RegisterPlugin` added at the next commit.

Paul-Licameli (changes requested):

src/PluginRegistrationDialog.cpp
```diff
@@ -34,6 +34,8 @@
 #if wxUSE_ACCESSIBILITY
 #include "widgets/WindowAccessible.h"
 
+#define DISABLE_STATE_NEW
```
Paul-Licameli, line 37: The constants `STATE_New` and `ID_ShowNew` might also exist only when conditionally compiled.  That would make it more obvious that the changes in this commit are complete and correct.

That would mean one of the event bindings in the table is also eliminated, and one case of a switch statement, and one if-block in `SetState`.

libraries/lib-module-manager/PluginManager.h
```diff
@@ -147,7 +149,12 @@ class MODULE_MANAGER_API PluginManager final : public PluginManagerInterface
    const ComponentInterfaceSymbol & GetSymbol(const PluginID & ID);
    ComponentInterface *Load(const PluginID & ID);
 
-   void CheckForUpdates(bool bFast = false);
+   /**
+    * \brief Ensures that all currently registered plugins are still exist
```
Paul-Licameli, line 153: "plugins still exist"

libraries/lib-module-manager/AsyncPluginValidator.h
```diff
@@ -0,0 +1,64 @@
...
+ * AsyncPluginValidation will notify caller via Delegate.
+ */
+class MODULE_MANAGER_API AsyncPluginValidator final
+{
+   class Impl;
+   std::shared_ptr<Impl> mImpl;
+public:
+   /**
+    * \brief Used to talk back to calling side
+    */
```
Paul-Licameli, line 34: Comments should explain that the callbacks occur on the main thread when the event loop idles or is yielded to.

src/PluginStartupRegistration.h
```diff
@@ -0,0 +1,53 @@
...
+
+//Helper class that passes plugins provided in constructor
+//to plugin validator, then "good" plugins are registered in
+//PluginManager. 
+class PluginStartupRegistration final :
+   public AsyncPluginValidator::Delegate
+{
+   std::unique_ptr<BasicUI::ProgressDialog> mDialog;
+   std::unique_ptr<AsyncPluginValidator> mValidator;
+   std::vector<std::pair<wxString, wxString>> mPlugins;
```
Paul-Licameli, line 31: It's good to make members of new classes `const` where possible as an aid to understanding.  This is one member that could be const.

src/PluginStartupRegistration.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+
+PluginStartupRegistration::PluginStartupRegistration(std::vector<std::pair<wxString, wxString>> plugins) : mPlugins(std::move(plugins))
+{
+   mValidator = std::make_unique<AsyncPluginValidator>(*this);
+}
+
+void PluginStartupRegistration::OnInternalError(const wxString& error)
+{
+   wxLogError(error);
```
Paul-Licameli, line 30: Is this where errors from PluginHost are used at last?  Then fixing the setup of locale in the other process is unimportant after all, and literal English string errors are all right.  But then, where `Translation()` was used, make it `Debug()`.

src/PluginStartupRegistration.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const std::vector<wxString>& PluginStartupRegistration::GetFailedPlugins() const noexcept
+{
+   return mFailedPlugins;
+}
+
+void PluginStartupRegistration::Run()
+{
+   mDialog = BasicUI::MakeProgress(XO("Searching for plugins"), XO(""));
```
Paul-Licameli, line 54: This could be just a local variable

src/PluginStartupRegistration.cpp
```diff
@@ -0,0 +1,81 @@
...
+   ProcessNext();
+   while(mValidator)
+   {
+      const auto message = TranslatableString { mPlugins[mNumChecked].second, { } };
+      //Update UI
+      if(mDialog->Poll(mNumChecked, mPlugins.size(), message) != BasicUI::ProgressResult::Success)
+         Stop();
+      //Process pending events
+      wxTheApp->Yield();
```
Paul-Licameli, line 63: `Poll` already did a `Yield` too

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -964,93 +992,6 @@ void PluginManager::SaveGroup(FileConfig *pRegistry, PluginType type)
    return;
 }
 
-// If bFast is true, do not do a full check.  Just check the ones
-// that are quick to check.  Currently (Feb 2017) just Nyquist
-// and built-ins.
-void PluginManager::CheckForUpdates(bool bFast)
```
Paul-Licameli, line 970: If the definition of `CheckPluginUpdates` is moved here, it makes the commit easier to review.  More black lines, fewer red and green.

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -333,6 +338,29 @@ PluginManager::~PluginManager()
...
+
+   // Check all known plugins to ensure they are still valid.
+   for (auto &pair : mRegisteredPlugins) {
+      auto &plug = pair.second;
+      const wxString & plugPath = plug.GetPath();
+      PluginType plugType = plug.GetPluginType();
+      
+      if (plugType != PluginTypeNone && plugType != PluginTypeStub && plugType != PluginTypeModule)
+      {
+         plug.SetValid(mm.IsPluginValid(plug.GetProviderID(), plugPath, true));
```
Paul-Licameli, line 353: Moving the definition makes it clearer that false became true in this line.  A more complete check of validity is done.

Still now, I wonder about this.  What happens if a plug-in is upgraded to a newer version reinstalled at the same path?  Should that be detected and cause another load again in the sandbox?

Nyquist.cpp contains checks for advance of the modification date of the plug-in file, making it easier to try changes while you develop a plug-in.  But those are just interpreted source code.

Maybe recording of time stamps is a future enhancement to do.

src/AudacityApp.cpp
```diff
@@ -400,6 +406,74 @@ void PopulatePreferences()
    gPrefs->Flush();
 }
 
+wxDialog* MakePluginsCheckResultsDialog(const std::vector<wxString>& failed)
+{
+   auto dialog = safenew wxDialogWrapper(nullptr, wxID_ANY, XO("New Plugins"));
```
Paul-Licameli, line 411: Does this leak?  A wxWindow is managed when it has a non-null parent, but this non-modal dialog does not!
But I see it does not leak because all the button event handlers cause it to destroy itself.  Very well, mention that here, to justify `safenew` (which is something like an assertion -- something needing a proof when you see it.)

src/AudacityApp.cpp
```diff
@@ -400,6 +406,74 @@ void PopulatePreferences()
...
+   dialog->SetSize(635, 414);
+   auto layout = std::make_unique<wxBoxSizer>(wxVERTICAL);
```
Paul-Licameli, line 413: You do a lot with explicit sizers and control creations and so on.  Are you not comfortable enough with ShuttleGui?  All right if not, but a rewrite of this should be in the future to make all of this desirable plug-in checking platform neutral too.

Besides which:  perhaps this dialog belongs in its own small source file.

Paul-Licameli:
This round of review is complete.  I am satisfied with the much improved large scale structure of interactions.  There are a few must-fix errors.  Resolve them and this can be approved.

libraries/lib-module-manager/PluginIPCUtils.h
vsverchinsky: Depends on project I believe, do we have such convention?

libraries/lib-module-manager/PluginIPCUtils.cpp
vsverchinsky: >Why not std::move(desc)? You need it to avoid copying, because desc alone still is an lvalue reference even though the argument is declared with &&.

Right, thanks!

>And that might allow vector instead of list for the container, but maybe there is a reason for list I don't understand yet, some requirement of non-relocation of the memory.

There is no such requirement, `PluginDescriptor` will be copied later, when added to `PluginManager`

libraries/lib-module-manager/PluginHost.cpp
vsverchinsky: Hm, didn't think of it, thanks. It's hard to answer such questions without a complete understanding of requirement. Though we can initialize i18n on the host side too, but then there is a question of language preference change. So, I don't have a complete answer, not use i18n in messages is a possible solution I think.

libraries/lib-module-manager/PluginHost.cpp
vsverchinsky: I was thinking of it too, and there is a downside - the size of application will increase, don't know how much though.

libraries/lib-module-manager/AsyncPluginValidator.cpp
vsverchinsky: Yes, it's correct

libraries/lib-module-manager/AsyncPluginValidator.cpp
vsverchinsky: Thanks! `this` is needed for class-member access

src/PluginStartupRegistration.cpp
vsverchinsky: Fixed

libraries/lib-module-manager/PluginManager.h
vsverchinsky: Thanks!

src/PluginStartupRegistration.cpp
vsverchinsky: Wasn't aware of that, thanks!

src/PluginStartupRegistration.cpp
vsverchinsky: It will cost some time because from what I see it doesn't yield each `Poll`. In my case the difference was 2-4 seconds.

libraries/lib-module-manager/PluginManager.cpp
vsverchinsky: >Still now, I wonder about this. What happens if a plug-in is upgraded to a newer version reinstalled at the same path? Should that be detected and cause another load again in the sandbox?

I'd say yes. But for this we need to track module file timestamp (if there is a file)

src/AudacityApp.cpp
vsverchinsky: >Are you not comfortable enough with ShuttleGui?

I'm still not very comfortable with wxWidgets)

>Besides which: perhaps this dialog belongs in its own small source file.

Probably, you are right

Paul-Licameli:
> Hm, didn't think of it, thanks. It's hard to answer such questions without a complete understanding of requirement. Though we can initialize i18n on the host side too, but then there is a question of language preference change. So, I don't have a complete answer, not use i18n in messages is a possible solution I think.

Language change only happens if the user uses the Preferences dialog, which can't happen during application initialization.  I wouldn't worry.

Anyway I see you take the acceptably simple way out -- you just use `Debug`.  Still let's remember this discussion.  Lowering the initialization of locale into some other place in lib- might not be needed yet but lookds easy and like the right thing to do in principle.

More generally, many miscellaneous things happening in AudacityApp.cpp could use reexamination for movement elsewhere.

Paul-Licameli:
> Hm, didn't think of it, thanks. It's hard to answer such questions without a complete understanding of requirement. Though we can initialize i18n on the host side too, but then there is a question of language preference change. So, I don't have a complete answer, not use i18n in messages is a possible solution I think.

Experiment will tell!

If ever my separation of modules (not libraries) proceeds, which allows pieces of Audacity functionality to be loaded only selectively -- then it might also be easy to suppress the loading of modules at startup time when in the plugin host process,
and that would reduce the memory footprint and static initialization code.

libraries/lib-module-manager/AsyncPluginValidator.cpp
Paul-Licameli: You didn't take this suggestion yet in the lastest force-push.  Did you not understand it?

Paul-Licameli:
> Thanks! `this` is needed for class-member access

But my suggestion was to access members through `lck->` instead of implicitly with captured `this`

See `UndoManager::EnqueueMessage` for an example I wrote

Paul-Licameli:
> It will cost some time because from what I see it doesn't yield each `Poll`. In my case the difference was 2-4 seconds.

In that case the other `Yield` could be kept for better performance, with a comment explaining it.  It's an easy tuning change to make later if we like.

Paul-Licameli:
> > Still now, I wonder about this. What happens if a plug-in is upgraded to a newer version reinstalled at the same path? Should that be detected and cause another load again in the sandbox?
> 
> I'd say yes. But for this we need to track module file timestamp (if there is a file)

New issue opened.  (Not yet a necessity for 3.2.)

Paul-Licameli:
> > Are you not comfortable enough with ShuttleGui?
> 
> I'm still not very comfortable with wxWidgets)
> 
> > Besides which: perhaps this dialog belongs in its own small source file.
> 
> Probably, you are right

ShuttleGui was invented to let us avoid worrying about many details in the use of wxWidgets.  

As I explained, it didn't go far enough for open-endedness, platform neutrality, declarativity, and recombination with generalized variables for holding control values.  But, it's a start.

Anyway @ksoze95 might take on the task of rewriting this small dialog with ShuttleGui soon.

Paul-Licameli (approved):

Paul-Licameli:
The one conversation I left un-resolved above is the detail about swapping of strings out of `std::optional` in a more efficient way.  Unimportant.

Let's merge this!  A very satisfying achievement!

Issue #2446, opened 2022-01-18
Some text input boxes for Macro commands are far too small


SteveDaulton:
**Is your feature request related to a problem? Please describe.**
Try using the "Export2:" command in a Macro. Note that a fully qualified file name is required in the very small text box

**Describe the solution you'd like**
A larger text box

**Describe alternatives you've considered**
Type the fully qualified file name in a plain text editor, then copy and paste into the text box.

**Additional context**
This applies to several Macro commands, including Export2, Open Project, Save Project,...

![export2](https://user-images.githubusercontent.com/11681097/149920659-93be1848-ef93-402c-8886-06a824cadb19.png)

Haoranxia:
Hey, I would like to take a look at this issue.

Maninder-Singh2217:
Hi, I would like to work on this issue.

Maninder-Singh2217:
Hi @LWinterberg, I was looking in the code and using same shuttleGUI class that is used in Preferences:Directories menu, was looking forward to expand this text box, by providing a browse button to select folder and then separately providing a line edit for specifying the filename.
Is this suitable as for the desired outcome?

![image](https://user-images.githubusercontent.com/43217891/167250468-45129241-4757-4751-9f37-a3da001e67fb.png)

LWinterberg:
looks good to me

Maninder-Singh2217:
Ok, will work on it

PR #2671, opened 2022-03-13
Made filename inputbox bigger for Export2 macro - issue:2446

Haoranxia:
Resolves: [#2446
](https://github.com/audacity/audacity/issues/2446)
*(Increased inputbox size for the Export2 macro. Same solution can be applied to other macros such as "OpenProject2" and "Import2". If this fix looks acceptable then I will be happy to apply it to any other macros that have this problem too :))*

Recommended:

### DIFF
--- a/src/commands/ImportExportCommands.cpp
+++ b/src/commands/ImportExportCommands.cpp
@@ -84,7 +84,7 @@ void ExportCommand::PopulateOrExchange(ShuttleGui & S)
 
    S.StartMultiColumn(2, wxALIGN_CENTER);
    {
-      S.TieTextBox(XXO("File Name:"),mFileName);
+      S.TieTextBox(XXO("File Name:"), mFileName, mFileName.length());
       S.TieTextBox(XXO("Number of Channels:"),mnChannels);
    }
    S.EndMultiColumn();


src/commands/ImportExportCommands.cpp
```diff
@@ -84,7 +84,7 @@ void ExportCommand::PopulateOrExchange(ShuttleGui & S)
 
    S.StartMultiColumn(2, wxALIGN_CENTER);
    {
-      S.TieTextBox(XXO("File Name:"),mFileName);
+      S.TieTextBox(XXO("File Name:"), mFileName, mFileName.length());
```
LWinterberg, line 87: this fix kinda works, but runs into some issues: 
1. the size of the window is dependent on the last exported file. For example, if I export D:\A.wav and open the dialog again, it's as short as ever:
image
Whereas if I just had an AAA.....wav, it is somewhat longer: 
image
2. using any sort of `string.length()` will give you a value that never works for variable-width font. III and WWW are both length 3, but one is obviously much wider than the other. 

I think a better fix for this issue could be either one of these: 

* A text box that's relative to the window size, and the window being expandable to whatever width you need it to be. 
* A "browse" button next to the text field that lets you select the file in the file browser, similar to what the normal import/export dialog windows would do. You may want to take a look at how the ffmpeg library locator did it, as it also can display a path. 
image

LWinterberg (changes requested):

Haoranxia:
Okay, thanks for the feedback! I will look into an alternative method.

src/commands/ImportExportCommands.cpp
Haoranxia: Hey @LWinterberg ,

So I tried looking into the 'Browse' button solution but got stuck at a certain point. I was able to create a 'Browse' button at a location similar to the `Locate FFmpeg` window.

However, I am unsure how to connect the button to a window that allows you to select a file. I have looked into the FFmpeg class and have tried to mimic the logic there. However I think I do not understand the codebase to a good extent yet.

From what I understand you define an `EVENT_TABLE` and then link an event to some ID. At the same time the button is linked to that ID too. I basically followed what was done in the `FFmpeg` file. 
However I have trouble figuring the eventhandling part of the code. So I have the following:

image

image

An `EVENT_TABLE` where I link the button to an `onBrowse` function that tries to mimic the same behaviour as what is defined in ``FFmpeg`. However when I try this I ran into some issues. 
- the `onBrowse` in `FFmpeg` uses a `FileSelector` and I can't seem to satisfy all the parameters required in the class `ImportExportCommands` which I am working in.

I also think that my approach could be entirely wrong due to my limited knowledge about the codebase. I was not really able to find any sources to understand things better. Maybe you could point me into the right direction or point me towards some sources I could take a look at?

LWinterberg: I'm unfortunately not a programmer myself, but maybe some of the following is helpful: 
1. There's a code documentation thing on https://doxy.audacityteam.org
2. if you install Graphviz (as described in the BUILDING.md), you'll get images showing you how various libraries depend on each other. 

Tell me whether that's been helpful, if not i can ask the team tomorrow

Haoranxia: Thanks. I think these will be very helpful. I'll let you know if I have any specific questions.

co18332:
Hi @Haoranxia @LWinterberg, the required form of export 2 widget is still not fulfilled right? Can I work on this?

Haoranxia:
Hey,

Yeah please feel free to work on it. I have stumbled onto busy times so I haven't been able to do much.

Kind regards,
Haoran

________________________________

PR #2672, opened 2022-03-13
Delete unnecessary comments from the AudacityApp.cpp file

NafieAmrani:
Resolves: Cluttered comments removal

*(While trying to understand the codebase I found several sections with seemingly un-useful comments. I decided to remove them and make a PR to get started with the project.)*

Recommended:

### DIFF
--- a/src/AudacityApp.cpp
+++ b/src/AudacityApp.cpp
@@ -479,8 +479,6 @@ static void QuitAudacity(bool bForce)
 /*end+*/
    {
       if (AllProjects{}.size())
-         // PRL:  Always did at least once before close might be vetoed
-         // though I don't know why that is important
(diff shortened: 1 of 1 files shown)


LWinterberg (approved):
unless @Paul-Licameli thinks that his old chat with james should remain

Paul-Licameli:
> unless @Paul-Licameli thinks that his old chat with james should remain

@LWinterberg  The presumption SHOULD be AGAINST a pull request like this.

If there are comments with initials by somebody, and they explain something non-obvious, leave them.

Unless you PROVE that the comments are wrong, or that whatever they are cautioning about doesn't really apply.

Just calling someone else's old comments "Seemingly un-useful" is -- weak.

PR #2677, opened 2022-03-14
Rewrite commands

Paul-Licameli:
Resolves: #2676

Recommended:

### DIFF
--- a/src/Shuttle.h
+++ b/src/Shuttle.h
@@ -53,30 +53,41 @@ public:
 
 class CommandParameters;
 /**************************************************************************//**
-\brief Shuttle that deals with parameters.  This is a base class with lots of
+\brief Visitor of effect or command parameters.  This is a base class with lots of
 virtual functions that do nothing by default.
 Unrelated to class Shuttle.
 ********************************************************************************/
-class AUDACITY_DLL_API ShuttleParams /* not final */
+class AUDACITY_DLL_API SettingsVisitor /* not final */
 {
 public:
    wxString mParams;
-   bool *pOptionalFlag;
-   CommandParameters * mpEap;
-   ShuttleParams() { mpEap = NULL; pOptionalFlag = NULL; }
-   virtual ~ShuttleParams() {}
+   bool *pOptionalFlag{};
+   CommandParameters * mpEap{};
+
+   SettingsVisitor() {}
+   virtual ~SettingsVisitor();
+
    bool ShouldSet();
-   virtual ShuttleParams & Optional( bool & WXUNUSED(var) ){ pOptionalFlag = NULL;return *this;};
-   virtual ShuttleParams & OptionalY( bool & var ){ return Optional( var );};
-   virtual ShuttleParams & OptionalN( bool & var ){ return Optional( var );};
-   virtual void Define( bool & var,     const wxChar * key, const bool vdefault, const bool vmin=false, const bool vmax=false, const bool vscl=false );
-   virtual void Define( size_t & var,   const wxChar * key, const int vdefault, const int vmin=0, const int vmax=100000, const int vscl=1 );
-   virtual void Define( int & var,      const wxChar * key, const int vdefault, const int vmin=0, const int vmax=100000, const int vscl=1 );
-   virtual void Define( float & var,    const wxChar * key, const float vdefault, const float vmin, const float vmax, const float vscl=1.0f );
-   virtual void Define( double & var,   const wxChar * key, const float vdefault, const float vmin, const float vmax, const float vscl=1.0f );
-   virtual void Define( double & var,   const wxChar * key, const double vdefault, const double vmin, const double vmax, const double vscl=1.0f );
-   virtual void Define( wxString &var, const wxChar * key, const wxString vdefault, const wxString vmin = {}, const wxString vmax = {}, const wxString vscl = {} );
-   virtual void DefineEnum( int &var, const wxChar * key, const int vdefault,
+   virtual SettingsVisitor & Optional( bool & var );
+   virtual SettingsVisitor & OptionalY( bool & var );
+   virtual SettingsVisitor & OptionalN( bool & var );
+   virtual void Define( bool & var, const wxChar * key, bool vdefault,
+      bool vmin = false, bool vmax = false, bool vscl = false );
+   virtual void Define( size_t & var, const wxChar * key, int vdefault,
+      int vmin = 0, int vmax = 100000, int vscl = 1 );
+   virtual void Define( int & var, const wxChar * key, int vdefault,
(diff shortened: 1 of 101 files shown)


src/ShuttleAutomation.cpp
```diff
@@ -181,3 +181,72 @@ void ShuttleSetAutomation::DefineEnum( int &var, const wxChar * key, const int v
    if( bWrite && bOK)
       var = temp;
 }
+
```
pietro68, line 184: In the previous commit, this renaming happened:  ShuttleParams ---> SettingsVisitor

This class is called ShuttleDefaults and used to inherit from ShuttleParams - now called SettingsVisitor.
By analogy, this class should now be called SettingsDefaults, or SettingsVisitorDefaults maybe?

src/Shuttle.h
```diff
@@ -56,41 +58,53 @@ class CommandParameters;
 \brief Shuttle that deals with parameters.  This is a base class with lots of
```
pietro68, line 58: Update comment "Shuttle"?

ksoze95 (approved):

src/commands/AudacityCommand.h
```diff
@@ -120,6 +120,9 @@ class AUDACITY_DLL_API AudacityCommand /* not final */ : public wxEvtHandler,
    //! Parameters, if defined.  false means no defined parameters.
```
pietro68, line 120: Are these comments outdated now? if you will rewrite them, please could you write in which case the Const visitor version is used, vs. the non-const version. 
My guess is, the non-const version is used when we want the visitor to read and write the settings, while the other is to be used when the visitor only needs to read them. An example use-case in the comments would be helpful :-)

src/Shuttle.h
```diff
@@ -56,41 +58,53 @@ class CommandParameters;
...
+@tparam Const if true, then visited settings are not modifiable.
 ********************************************************************************/
-class AUDACITY_DLL_API SettingsVisitor /* not final */
+template<bool Const>
+class SettingsVisitorBase /* not final */
 {
 public:
+   template<typename T> using Ref = std::conditional_t<Const, const T&, T&>;
+   template<typename T> using Arg = std::conditional_t<Const, T, T&>;
+
```
pietro68, line 70: I am a bit confused by the purpouse of Arg<>. 
I would have expected it to be defined like `std::conditional_t<Const, const T, T>;` - but as it is now, it switches passing an arg either by copy or by nonconst-reference; I find it strange that the bool Const in one case (Ref) effectively switches constness, and in the other (Arg) switches between copy/reference.

pietro68 (changes requested):
Please have a look at my comments/questions

src/ShuttleAutomation.cpp
Paul-Licameli: I could rename things too, but I thought that might make the PR more overwhelming.  But if you don't mind it, I will add a renaming commit.

Paul-Licameli: I think subclasses of abstract visitors should be named for what they do.  This one should be called "SettingsLoader".  I would rename

ShuttleSetAutomation -> SettingsSaver
ShuttleGetDefinition -> SettingsDescriber
ShuttleDefaults -> SettingsResetter

src/Shuttle.h
Paul-Licameli: Yes I should, and also comments in ShuttleAutomation.h and EffectAutomationParameters.h and ShuttleGetDefinition.h

src/Shuttle.h
```diff
@@ -56,41 +58,53 @@ class CommandParameters;
...
-   bool *pOptionalFlag{};
+   std::conditional_t<Const, const bool, bool> *pOptionalFlag{};
    CommandParameters * mpEap{};
 
-   SettingsVisitor() {}
-   virtual ~SettingsVisitor();
+   SettingsVisitorBase() {}
+   virtual ~SettingsVisitorBase();
 
    bool ShouldSet();
```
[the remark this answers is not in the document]
Paul-Licameli: Some subclasses reassign values in the effect objects.  Those are the non-const SettingsVisitors.  Both type aliases expand to non-const references for both.  They are bound to members of the effects when they are called.

Some subclasses only examine values, and these must become the ConstSettingVisitors in later commits.  They must take other than non-const references.  I just pass by value, except in the case of wxString-valued fields.  Then it's const reference just for efficiency.

src/commands/AudacityCommand.h
Paul-Licameli: You rightly noticed that the ConstSettingsVisitor class is defined but not yet used in this pull request.

It will become the base class for ShuttleGetDefinition and ShuttleGetAutomation, and the NullShuttle in EffectManager.cpp will be of that class.  Those are the visitors that do not modify the effect or command objects.

Other work must happen before I do that.  Effects, as well as commands, must have two overloads of VisitSettings so they can also accept ConstSettingsVisitor.  This PR accomplishes that only for commands.  A later PR will do it for effects.

Paul-Licameli: > Are these comments outdated now? (because they reference "Parameters") in any case I do not find them much helpful so I'd ask you to correct and expand them a bit.
> 
> Please could you also comment in which case the Const visitor version is used, vs. the non-const version. An example use-case would be helpful.

Parameters, settings... interchangeable words I think.

Basically, there is just a success code.  The base class AudacityCommand doesn't know how to visit any settings.

src/Shuttle.h
Paul-Licameli: I explained why the second alternative must be `T&` to allow non-const visitors to mutate the effect they are visiting.

The difference between Ref and Arg for const visitors is just about the efficiency of pass by value versus reference.

Making the first alternative of arg `Const T` doesn't really change the meaning of the function declarations that use it.  `const` qualifying a by-value argument in a declaration really makes no difference to how the calling code compiles.  It makes a difference only in the function definition, to treat the argument as an immutable local variable.

src/Shuttle.h
pietro68: Thanks

src/Shuttle.h
pietro68: Alright, I see it now in this way: Ref and Arg are both ways to statically switch between passing an argument with a no-change guarantee, or not. 

Arg differs by Ref in this way: the no-change-guarantee is implemented with pass-by-value. Use one or the other, depending on what is being passed.

pietro68 (approved):

src/ShuttleAutomation.cpp
pietro68: Fine for doing that on another PR

Paul-Licameli:
I have merged this, with differences only in improved comments from what was reviewed.

PR #2680, opened 2022-03-14
Simplify effect parameters

Paul-Licameli:
Resolves: #2679 

Still a very drafty draft, based on the waiting PR #2676 

I invite the start of reviews now.  The fourth commit is big.

Recommended:

### DIFF
--- a/src/Shuttle.h
+++ b/src/Shuttle.h
@@ -18,7 +18,12 @@
 class ComponentInterfaceSymbol;
 class WrappedType;
 
-template<typename Type> struct EffectParameter {
+template<
+   typename Structure, //!< Structure holding the parameter
+   typename Member, //!< Often the same as Type
+   typename Type //!< Type of the given default value
+> struct EffectParameter {
+   Member Structure::*const mem{}; //!< Member holding the parameter
    const wxChar *const key{}; //!< Identifier in configuration file
    const Type def{};          //!< Default value
    const Type min{};          //!< Minimum value
@@ -33,20 +38,26 @@ template<typename Type> struct EffectParameter {
 
 // Deduction guides
 // Type of def chooses the parameter type; others just need to be convertible
-template<typename Type, typename... Args>
-EffectParameter(const wxChar *key, const Type &def, Args...)
-   -> EffectParameter<Type>;
+template<typename Structure, typename Member,
+   typename Type, typename... Args>
+EffectParameter(Member Structure::*const mem,
+   const wxChar *key, const Type &def, Args...)
+      -> EffectParameter<Structure, Member, Type>;
 // Deduce string type from string literal
-template<typename Char, size_t N, typename... Args>
-EffectParameter(const wxChar *key, const Char (&def)[N], Args...)
-   -> EffectParameter<wxString>;
+template<typename Structure, typename Member,
+   typename Char, size_t N, typename... Args>
+EffectParameter(Member Structure::*const mem,
+   const wxChar *key, const Char (&def)[N], Args...)
+      -> EffectParameter<Structure, Member, wxString>;
 
-struct EnumParameter : EffectParameter<int>
+template<typename Structure, typename Member>
+struct EnumParameter : EffectParameter<Structure, Member, int>
 {
-   EnumParameter(
+   EnumParameter(Member Structure::*const mem,
       const wxChar *key, int def, int min, int max, int scale,
       const EnumValueSymbol *symbols_, size_t nSymbols_ )
-      : EffectParameter{ key, def, min, max, scale, {} }
+      : EffectParameter<Structure, Member, int>{
+         mem, key, def, min, max, scale, {} }
       , symbols{ symbols_ }
       , nSymbols{ nSymbols_ }
    {}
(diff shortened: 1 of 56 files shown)


src/effects/FindClipping.cpp
```diff
@@ -96,8 +96,8 @@ bool EffectFindClipping::GetAutomationParameters(CommandParameters & parms) cons
 
 bool EffectFindClipping::SetAutomationParameters(const CommandParameters & parms)
 {
-   ReadAndVerifyInt(Start);
-   ReadAndVerifyInt(Stop);
+   ReadParam(Start);
```
pietro68: Now that `ReadParam` resolves types at compile time, we have no more need for the many `ReadAndVerifyInt`, `ReadAndVerifyBool `etc.  - however, in the process we lost the verb "verify" in the function name , which is what ReadParam actually does. So, could we rename `ReadParam` to `ReadAndVerify`?

src/effects/Effect.h
```diff
@@ -133,6 +138,11 @@ class AUDACITY_DLL_API Effect /* not final */ : public wxEvtHandler,
    size_t SetBlockSize(size_t maxBlockSize) override;
    size_t GetBlockSize() const override;
 
+   // VisitSettings(), GetAutomationParameters(), and SetAutomationParameters()
+   // use the functions of EffectParameterMethods.  By default, this function
+   // defines an empty list of parameters.
+   virtual const EffectParameterMethods &Parameters() const;
+
```
pietro68, line 145: Proposal: change the name of this method to GetEffectParameterMethods

src/effects/Equalization.h
```diff
@@ -274,6 +276,9 @@ class EffectEqualization : public Effect,
    wxBoxSizer *szrM;
 #endif
 
+   CapturedParameters<EffectEqualization> mParameters;
```
pietro68, line 278: What is this member for? we already have the static capture `parameters` in `EffectEqualization::Parameters`

src/ShuttleAutomation.h
```diff
@@ -11,85 +11,293 @@
...
+    and returning a pointer to something that holds the parameters
+ @tparam Parameters a pack of non-type template parameters, whose types are
+    specializations of the class template EffectParameter
+
+ TODO in C++20: could use simply `auto ...Parameters` if they are `constexpr`
+*/
+template <typename EffectType, const auto &...Parameters>
+class CapturedParameters : public EffectParameterMethods {
+public:
+   static constexpr auto *fetcher = &EffectType::FetchParameters;
```
pietro68, line 56: On this line, using msvc2019 I get:
`error C3535: cannot deduce type for 'auto *' from 'int'`

src/Shuttle.h
```diff
@@ -56,34 +107,51 @@ class CommandParameters;
...
+      float vmin, float vmax, float vscl = 1.0f );
+   virtual void Define( Arg<double> var, const wxChar * key, double vdefault,
+      double vmin, double vmax, double vscl = 1.0f );
+   virtual void Define( Ref<wxString> var, const wxChar * key,
+      wxString vdefault,
+      wxString vmin = {}, wxString vmax = {},
+      wxString vscl = {} );
+   virtual void DefineEnum( Arg<int> var, const wxChar * key, int vdefault,
       const EnumValueSymbol strings[], size_t nStrings );
 };
```
pietro68, line 150: On these two I get this warnings on msvc2019:

`warning C4910: 'SettingsVisitorBase<false>': '__declspec(dllexport)' and 'extern' are incompatible on an explicit instantiation`

Not sure how serious this warning is though: I can not get to the link stage because I get a compile error (which I noted elsewhere)

src/ShuttleAutomation.h
```diff
@@ -11,85 +11,293 @@
...
+      Value temp;
+      if (!parms.ReadAndVerify(param.key, &temp, param.def,
+         param.min, param.max))
+         return false;
+      structure.*(param.mem) = temp;
+      return true;
+   }
+   // More specific overload for enumeration parameters
+   template< typename Member >
+   static bool SetOne(Effect &effect, Params &structure,
```
pietro68, line 186: as above

src/effects/FindClipping.cpp
Paul-Licameli: The commit "Remove unused macros and the cache in EffectParameter" ultimately eliminates these macros.  This is just one simplifying step in my transformation.

src/effects/Equalization.h
Paul-Licameli: Good find.  That was left over from an earlier version of this pull request, in which the parameters object still needed to capture `this` and could not be a static.  I left the same in TruncSilence.h.

src/effects/FindClipping.cpp
pietro68: ok

src/ShuttleAutomation.h
```diff
@@ -11,85 +11,293 @@
...
+      // Serialize one enumeration variable as a string identifier, not a number
+      parms.Write(
+         param.key, param.symbols[ structure.*(param.mem) ].Internal() );
+   }
+   static void DoGet(const Params &structure, CommandParameters &parms) {
+      (GetOne(structure, parms, Parameters), ...);
+   }
+
+   template< typename Member, typename Type, typename Value >
+   static bool SetOne(Effect &effect, Params &structure,
```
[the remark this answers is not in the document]
Paul-Licameli: Good catch

Paul-Licameli:
I have updated the branch, fixing the unused parameter Pietro pointed out.

I also did a self-review of "Use CapturedParameters to simplify some effects..." and fixed some details, where there are post-set functions used in the construction of parameters.

pietro68 (approved):
I think I understood the code, save from a few bits in the CapturedParameters section - but that code has changed a bit since this review was opened, so I am going to study that again.

Issue #2684, opened 2022-03-16
EffectSettings required in still more functions


Paul-Licameli:
Making things properly const-correct, and the externalizing of state, are inherently connected problems.

EffectSettings as state must be passed into still more functions, so that all the functions that should be const become so.

This will eliminate most of the interface in the transitional class EffectDefinitionInterfaceEx.

PR #2685, opened 2022-03-16
Even more effect settings

Paul-Licameli:
Resolves: #2684

Recommended:


pietro68 (approved):
I had a look at the last 6 commits (i.e. the actual new ones) and I got a sense of what they do.

Issue #2614, opened 2022-03-03
Selection indication on the time bar does not correspond to selection


Penikov:
**Describe the bug**
![image](https://user-images.githubusercontent.com/25851359/156559629-ef8bf200-7100-4457-ac05-519772ace0c2.png)

**To Reproduce**
1. Generate a clip
2. Click the Add Effect button
3. Select a random region on a track

**Expected behavior**
The timeline should be bound to track area

**Additional information (please complete the following information):**
 - OS: All
 - Version: Audacity 3.2.0-alpha-20220302

petersampsonaudacity:
Note that the Timeline is out of kilter too.

In the image above the audio appears to start at ~4.3 seconds

petersampsonaudacity:
This may be related to:
 With 3.2.0 the track area can be dragged totally off screen #2620

petersampsonaudacity:
Testing on W11 with @vsverchinsky  Vita;'s branch build  audacity-win-3.2.0-alpha-20220316+f36d06b-x64-msvc2022 

This is looking much better

a) works fine at default size of the ne (nlank as of now) effect pane
b) the pinned playhead works fine too
![image](https://user-images.githubusercontent.com/50205984/158771047-4b464833-cd3d-4b76-86d1-5f7b87bd7f0b.png)


c) and works with increased size effects pane
![image](https://user-images.githubusercontent.com/50205984/158771268-6f87f649-4cba-4088-8351-6eecbca2f5dd.png)

petersampsonaudacity:
Testing on W10 with latest alpha master audacity-win-3.2.0-alpha-20220412+17bfc14-x64-msvc2022

Looks good to go to me

dozzzzer:
Testing on MacOS 11.6, looking good. Moving to 'Done'.

PR #2686, opened 2022-03-16
Issue 2614: AdornedRulerPanel position fix

vsverchinsky:
Resolves: #2614

AdornedRulerPanel was reparented to the track panel list so that when effects panel is shown ruler properly changes it's position

Recommended:

### DIFF
--- a/src/ProjectManager.cpp
+++ b/src/ProjectManager.cpp
@@ -224,11 +224,6 @@ void InitProjectWindow( ProjectWindow &window )
    ToolManager::Get( project ).LayoutToolBars();
 
    //
-   // Create the horizontal ruler
-   //
-   auto &ruler = AdornedRulerPanel::Get( project );
-
-   //
    // Create the TrackPanel and the scrollbars
    //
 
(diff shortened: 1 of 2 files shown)


src/ProjectManager.cpp
```diff
@@ -237,15 +232,9 @@ void InitProjectWindow( ProjectWindow &window )
    {
       auto ubs = std::make_unique<wxBoxSizer>(wxVERTICAL);
       ubs->Add( ToolManager::Get( project ).GetTopDock(), 0, wxEXPAND | wxALIGN_TOP );
-      ubs->Add(&ruler, 0, wxEXPAND);
       topPanel->SetSizer(ubs.release());
    }
```
Paul-Licameli, line 236: Now that this vertical sizer only contains the top tool dock panel, there is no need for it.  The tool panel top dock might be parented directly to `topPanel`, simplifying.

Paul-Licameli: I'm not sure what subtle changes in keystroke navigation that might bring.

Paul-Licameli (approved):

Issue #2569, opened 2022-02-18
Understandable journal test results:  Eliminate Audacity's spam on stdout and stderr


Paul-Licameli:
I tried running the new unit + functional testing system.  This includes a run of the simplest possible journal file, which just starts Audacity and exits.

The journal sanity test makes a lot of spam, and so will every journal test:
```
2: 06:47:42: Debug: Warning: command doesn't exist: 'Cut'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Copy'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Paste'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Trim'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Silence'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Undo'
2: 06:47:42: Debug: Warning: command doesn't exist: 'Redo'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomIn'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomOut'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomToggle'
2: 06:47:42: Debug: Warning: command doesn't exist: 'ZoomSel'
2: 06:47:42: Debug: Warning: command doesn't exist: 'FitInWindow'
2: 06:47:42: Debug: Unrecognized accel key 'NUMPAD_ENTER', accel string ignored.
```

Let's have an issue to clean that up

_Originally posted by @Paul-Licameli in https://github.com/audacity/audacity/issues/2536#issuecomment-1044399155_

PR #2687, opened 2022-03-17
Eliminate Audacity's spam on stdout and stderr

ksoze95:
Resolves: #2569

Recommended:

### DIFF
--- a/src/Menus.cpp
+++ b/src/Menus.cpp
@@ -46,6 +46,7 @@
 
 #include <wx/menu.h>
 #include <wx/windowptr.h>
+#include <wx/log.h>
 
 MenuCreator::MenuCreator()
 {
@@ -439,6 +440,8 @@ void MenuCreator::CreateMenusAndCommands(AudacityProject &project)
 void MenuManager::Visit( ToolbarMenuVisitor &visitor )
 {
    static const auto menuTree = MenuTable::Items( MenuPathStart );
+
+   wxLogNull nolog;
    Registry::Visit( visitor, menuTree.get(), &sRegistry() );
 }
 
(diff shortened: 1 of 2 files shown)


src/ProjectManager.cpp
```diff
@@ -216,6 +216,10 @@ void InitProjectWindow( ProjectWindow &window )
    // Near as I can tell, this is only a problem under Windows.
    //
 
+   //
+   // Create the Menus and Commands
+   //
+   MenuManager::Get( project ).CreateMenusAndCommands( project );
```
Paul-Licameli, line 222: This is a simple change,  but I would phrase the commit comment to describe the program’s behavior after the change,  not before.

Paul-Licameli: Also add a comment in code, why the ordering of these two calls matters.

Paul-Licameli (changes requested):
Comments and other code movement

src/Menus.cpp
```diff
@@ -46,6 +46,7 @@
 
 #include <wx/menu.h>
 #include <wx/windowptr.h>
+#include <wx/log.h>
```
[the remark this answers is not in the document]
ksoze95: :)

src/ProjectManager.cpp
ksoze95: I have moved function call back to ProjectManager::New, just before the InitProjectWindow.

Paul-Licameli (approved):

Paul-Licameli (changes requested):

Paul-Licameli (approved):

PR #2691, opened 2022-03-18
I2590 recording lost

pietro68:
Resolves: (https://github.com/audacity/audacity/issues/2590)

The first commit will be removed before merge, and its purpouse is only to make the bug happen much more often, so that the fix can be tested more reliably.

The second commit is the actual fix.

With the above mentioned enhancer of bug rate turned on, A/B tests were done like so:

1) add a mono track
2) add a label track
3) hit record button
4) click on the label track, press a key (e.g. "t") and *keep it pressed*
5) release the key when either the bug happens, or after 12 seconds of it not happening.
6) delete both tracks

Repeat the above steps 10 times without the fix, then again 10 times with the fix.

Without the fix: the bug always happened, between 2 and 4 seconds from the start.
With the fix:    the bug never happened.

Recommended:


src/ProjectFileIO.cpp
```diff
@@ -1895,7 +1895,13 @@ bool ProjectFileIO::WriteDoc(const char *table,
 {
    auto db = DB();
 
-   TransactionScope transaction(mProject, "UpdateProject");
+   // This method is called:
+   // - periodically when recording, in the AudioIO Thread (i.e. DrainBuffers)
+   // - when keystroking a label,    in the Main    Thread
```
Paul-Licameli, line 1900: Comment should really say, it's called from the main thread whenever there is any editing operation.  Most of those are disabled during play and recording but addition of labels is one example that is not (but that's not the exhaustive list of cases).

libraries/lib-transactions/TransactionScope.cpp
```diff
@@ -43,6 +48,9 @@ TransactionScope::~TransactionScope()
          wxLogMessage("Transaction active at scope destruction");
       }
    }
+
+   if (mpMutex)
+      mpMutex->unlock();
```
Paul-Licameli, line 53: You found a way to make the bug occur more frequently.  I speculated that the fix for it is to make transactions on the database effectively single-threaded (serialized).  You achieved that this way, and observed that the bug could not be reproduced where before it had been easy to reproduce (given the change of buffer size).

That is all good -- however, I fear other problems might result from this commit.

Either thread trying to use a TransactionScope might end up blocking for the duration of the other thread's TransactionScope.  If the Audio thread blocks for the main thread, that might be unacceptable.  Who knows how long the transaction commit on the main thread might take, in general?  It might be unusually long sometimes and so cause drop-outs in recording.

My first suggestion for fixing this bug was instead that you use BasicUI::CallAfter to let the Audio thread describe the work to do in a lambda, whose body is later executed on the main thread.

There is still some implied mutex locking and unlocking inside wxWidgets, in the management of that queue of work items to do in idle time, and some memory allocation too.  But the duration of that lock would only last for the pushing of the queue item.  The TransactionScope objects would only ever exist in the main thread, so the mutex you added to the class would not be needed.

This fix may require more lines of code -- you may need to explore the call graph to identify all possible paths to construction of TransactionScope that could happen in the Audio thread.

@crsib anything to add?

Paul-Licameli (changes requested):

libraries/lib-transactions/TransactionScope.cpp
pietro68: Apologies, I overlooked that suggestion you wrote on the issue.

Alright, so I will try your suggestion then. This will mean, every auto-save of the wave while is being recorded, will happen in the main thread instead of the AudioIO thread. 

Are we sure that the implications of this will not be problematic? like for instance, if the user records 1.5 hours of audio (as the user reporting the problem says he did), when the auto-saving happens, it will happen in the main thread, therefore blocking user input for all the time it needs to save (for instance, the stop button won't work during all the time it takes)

Caveat: I am just trying to imagine what could happen in situations like that, but I am a bit clueless really, maybe the wave saving is done incrementally, in that case the main thread will be blocked for the time it takes to save 6 seconds of audio, not 1.5 hours of it.

pietro68:
I reverted my fix and coded the fix suggested by Paul. I repeated the tests in the description and I still got a failure rate of 30%.

I found out that the cause was a TransactionScope enclosing the AutoSave (which also uses a TransactionScope); the former runs in the AudioThread, the latter (after coding the fix) runs in the MainThread. This caused sometimes a failure when both TransactionScopes came to an end and ran concurrently the closing SQL statement (releasing a savepoint).

The outer TransactionScope was found to be actually useless (*) so I deleted it; then I repeated the tests, and I got a 0% failure rate.

(*) as far as I can see - I step-debugged the code to some depth and I found no uses of the DB between the start and the end of the outer TransactionScope, except the uses which are already protected by the inner TransactionScope.

libraries/lib-transactions/TransactionScope.cpp
DavidBailes: > Are we sure that the implications of this will not be problematic? like for instance, if the user records 1.5 hours of audio (as the user reporting the problem says he did), when the auto-saving happens, it will happen in the main thread, therefore blocking user input for all the time it needs to save (for instance, the stop button won't work during all the time it takes)

Am I right in thinking that user input may now be blocked during auto saves? So for, example, if the user presses the letter X to stop and set the cursor position, then this may be delayed by an auto save? If so, this is not at all satisfactory or keyboard users who use such commands for positioning the cursor, and start/end of the selection.

DavidBailes: I had a look at the changes, and I don't think they affect playback, so if this is true, then my concerns expressed in my previous comment are not valid.

Paul-Licameli (approved):

Issue #2646, opened 2022-03-08
Move Effect::mDuration to EffectSettings


Paul-Licameli:
Move one special effect setting, that is common to many, into EffectSettings, eliminating some state from Effect.

Prepares for the elimination of EffectHostInterface.

Issue #2630, opened 2022-03-04
Eliminate confusing effect management things Effect::Startup, EffectHostInterface, and Effect::mClient


Paul-Licameli:
Eliminate EffectHostInterface,  Effect::Startup, Effect::mClient -- just what were they for?  It was hard to describe.

Instead, the third party effect classes will simply all inherit the same Effect class.

PR #2698, opened 2022-03-21
Simplify effect initialization

Paul-Licameli:
Resolves: #2630
Resolves: #2646

Clear up confusing things in management of effects:  the host/client distinction that was hard to explain.  All effects,
including third-party, implement all of the "host" and the "client" interfaces.  Third party effects don't need a "host"
shell Effect to forward certain method calls to them.

To achieve this, required also resolving the confusions about management of the duration setting of generators.  That is
now part of the non-polymorphic part of EffectSettings.

Recommended:


pietro68 (approved):
All fine by me

Issue #2576, opened 2022-02-21
DTMF effect converted to use the new interface for stateless effects


Paul-Licameli:
To serve as a first example, than other built-in effects converted by analogy.

Formerly mentioned BassTreble in the title, but that overspecified what we needed.  It looks more likely that EffectDtmf will need to be ther first complete example.  (Actually a generator, not strictly an effect.)

PR #2707, opened 2022-03-22
Rewrite dtmf

Paul-Licameli:
Resolves: #2576

Make DTMF the first effect (generator, to be exact) that externalizes its settings and state.

Recommended:

### DIFF
--- a/src/effects/DtmfGen.cpp
+++ b/src/effects/DtmfGen.cpp
@@ -32,12 +32,13 @@
 namespace {
 wxString AllSymbols();
 }
+
 const EffectParameterMethods& EffectDtmf::Parameters() const
 {
    static CapturedParameters<EffectDtmf,
       Sequence, DutyCycle, Amplitude
    > parameters{
-      [](EffectDtmf &, EffectSettings &es, Settings &s, bool updating){
+      [](EffectDtmf &, EffectSettings &es, DtmfSettings &s, bool updating){
          if (updating) {
             if (s.dtmfSequence.find_first_not_of(AllSymbols())
                 != wxString::npos)
@@ -88,7 +89,6 @@ namespace{ BuiltinEffectsModule::Registration< EffectDtmf > reg; }
 
 EffectDtmf::EffectDtmf()
 {
-   Parameters().Reset(*this);
 }
 
 EffectDtmf::~EffectDtmf()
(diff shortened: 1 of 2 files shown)


src/effects/DtmfGen.h
```diff
@@ -94,7 +111,6 @@ class EffectDtmf final : public Effect
    struct Validator;
```
pietro68, line 132: Github will not allow me to comment on line 102 so I need to do it here: 
On msvc2019 I get this error:

`error C2397: conversion from 'size_t' to 'int' requires a narrowing conversion`

Paul-Licameli:
I have force-pushed it to be up to date with #2698 and to fix the Linux and Windows build failures.

src/effects/DtmfGen.cpp
```diff
@@ -386,6 +415,14 @@ EffectDtmf::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess &access)
    return result;
 }
 
+std::shared_ptr<EffectInstance>
+EffectDtmf::MakeInstance(EffectSettings &settings)
+{
+   // TODO: don't use Effect::mT0 and Effect::mSampleRate, but use an
```
pietro68, line 421: When I try to generate a sequence, I get an access violation. I traced the root cause to this point.

When this point is reached, mSampleRate is 0.0 (as defaulted in class `Effect` and never changed then) and that is sent to the created instance.

That, in turn, will make `numSamplesSequence`, `numSamplesTone`, `numSamplesSilence` be all 0 when they are calculated in `EffectDtmf::Instance::ProcessInitialize`, and that in turn will trigger a violation access in EffectDtmf::Instance::ProcessBlock (because the correct indexing of the string containing the sequence depends on those values being right).

Paul-Licameli: Yes.

Please see the latest force push, in which the Instance class overrides SetSampleRate().

And now this PR depends on not one but two other pending PRs !

pietro68: I see this has been fixed in later commits, ok

pietro68 (approved):
I got a sense of what the changes are about - mostly about defining new classes, changing the inheritance lines of the existing ones, moving some methods, and most importantly having the concept of EffectInstance in place.

Paul-Licameli:
> I got a sense of what the changes are about - mostly about defining new classes, changing the inheritance lines of the existing ones, moving some methods, and most importantly having the concept of EffectInstance in place.

Did you exercise all the new code in the debugger too?

pietro68:
Not all of it, I only tried the DtmfGen on some breakpoints. I will do a deeper exploration in the debugger then.

Issue #2284, opened 2021-12-14
The Journaling Tool. Audacity crashes at replaying the comand of a clip generating


Penikov:
**Is your feature request related to a problem? Please describe.**
The Journaling tool hasn't been able to journal and replay the clip-generating commands such as Chirp, DTFM tones, Noise, Silence,     Tone.

**Describe the solution you'd like**
I would like to have a feature of journaling and automated replaying for generating
In the current version, Audacity just crashes in an attempt of replaying this type of command.

dozzzzer:
I ran repeatedly into the same issue: when playing a journal.txt file, Audacity freezes.
STR:

1. Open Audactiy with the "Write Journal" option enabled
2. Generate > DTMF tones... > OK
3. Quit and confirm
4. Launch Audacity via Terminal with the path to the journal file

Act. result: Audacity launches, creates a blank track and freezes
Exp.: Audacity launches, creates DTMF tones track and quits

OS: MacOS Big Sur 11.6
Audacity version: [audacity-macOS-3.2.0-alpha-20220329+defe745-universal](https://github.com/audacity/audacity/suites/5847072124/artifacts/196742830
[journal.txt](https://github.com/audacity/audacity/files/8380006/journal.txt)

https://user-images.githubusercontent.com/70063970/160824247-0e5e8cba-bf34-4e55-9144-68932cba8bce.mp4

)

PR #2720, opened 2022-03-22
Fixed journal playback for clip-generating commands.

ksoze95:
Resolves: #2284

Recommended:

### DIFF
--- a/src/effects/Effect.cpp
+++ b/src/effects/Effect.cpp
@@ -473,7 +473,7 @@ int Effect::ShowHostInterface(wxWindow &parent,
 
    // Let the client show the dialog and decide whether to keep it open
    auto result = client->ShowClientInterface(parent, *mHostUIDialog, forceModal);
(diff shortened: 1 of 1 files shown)


Paul-Licameli (approved):
A simple null check.

But did you also look further into the failure to clean up the single instance checker?

It would still be desirable to fix that.  In future, if journal playback should crash for any other reason, it should not leave behind a dirty environment.

crsib:
>But did you also look further into the failure to clean up the single instance checker?

This is an old known issue with the same internal cause as #2114. Audacity closes all windows, but actually continues to run.

Paul-Licameli:
@crsib @ksoze95 I approved this one-line null pointer check but didn't try a rerun.

Is the issue "fixed" only to the extent of not crashing on journal replay -- or is it a complete fix, meaning the journal replays to the end and Audacity exits with a successful return code to the shell?

PR #2730, opened 2022-03-25
Define effect instance

Paul-Licameli:
Resolves: #2729

Next step of restructuring for stackable effects.  Define interfaces to make the distinction between a plug-in, which should
become stateless, from the instance(s) of it which may hold state.

Recommended:


src/effects/PerTrackEffect.h
```diff
@@ -50,11 +50,9 @@ class PerTrackEffect : public Effect
       FloatBuffers &inBuffer,
       FloatBuffers &outBuffer,
       ArrayOf< float * > &inBufPos,
-      ArrayOf< float *> &outBufPos);
+      ArrayOf< float *> &outBufPos, size_t bufferSize, unsigned mNumChannels);
```
pietro68, line 100: Trivial: numChannels instead of mNumChannels

pietro68:
I checked that the Nyquist prompt example worked.

I found that all realtime effects would trigger an assertion when used, and that this happens because a member of `RealtimeEffectState`, specifically `EffectProcessor *mEffect;` was not initialized to nullptr; however, later commits happening on branch Realtime-effects-use-EffectInstance remove this member and replaces it with something else, so this bug will be superseded. 

Still maybe it would be better to apply this small fix before merging, so to leave realtime effects working on master.

pietro68 (approved):
I think I understand the changes; before merging, please read my last comment in the conversation.

PR #2731, opened 2022-03-25
update credits and debian/copyright

LWinterberg:
Updates the credits to the current state of affairs another time, and also attempts to stop the .deb from showing up as "proprietary"

### DIFF
--- a/help/audacity.appdata.xml
+++ b/help/audacity.appdata.xml
@@ -3,12 +3,12 @@
   <developer_name>Muse Group</developer_name>
   <id>org.audacityteam.Audacity</id>
   <launchable type="desktop-id">audacity.desktop</launchable>
-  <project_license>GPL-2.0</project_license>
+  <project_license>GPL-3.0</project_license>
   <metadata_license>CC0-1.0</metadata_license>
   <name>Audacity</name>
   <summary>Record and edit audio files</summary>
   <description>
-    <p>Audacity® is a free, easy-to-use, multi-track audio editor and recorder for Windows, Mac OS X, GNU/Linux and other operating systems. The interface is translated into many languages.</p>
+    <p>Audacity® is a free, easy-to-use, multi-track audio editor and recorder for Windows, macOS, GNU/Linux and other operating systems. The interface is translated into many languages.</p>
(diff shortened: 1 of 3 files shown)


LWinterberg:
TIL that github.com's "rebase and merge" is different to a rebase. will fix that another time. 

Also, it seemed to not help with anything anyway:

Screenshot from 2022-04-12 08-11-15

Not sure what exactly the problem here is. or why it just says "Windows." I'll also investigate that another time

crsib:
Can this be a bug in GNOME? https://gitlab.gnome.org/GNOME/gnome-software/-/issues/833

crsib:
https://gitlab.gnome.org/GNOME/gnome-software/-/issues/875

LWinterberg:
yeah, looks like it's a gnome bug. thanks for digging that up!

shoogle (approved):

help/audacity.appdata.xml
```diff
@@ -3,12 +3,12 @@
   <developer_name>Muse Group</developer_name>
   <id>org.audacityteam.Audacity</id>
   <launchable type="desktop-id">audacity.desktop</launchable>
-  <project_license>GPL-2.0</project_license>
+  <project_license>GPL-3.0</project_license>
```
imciner2, line 6: Note that the appdata license field must use an SPDX identifier according to its spec, and this identifier was deprecated in SPDX version 3.0. It is replaced with either `GPL-3.0-only` or `GPL-3.0-or-later` depending on which you mean. If this field parses to something that isn't a valid SPDX identifier, then it can cause programs to think the app is proprietary.

crsib: @LWinterberg can you please check?

shoogle: @imciner2, well spotted! The plain GPL SPDX identifiers were indeed deprecated as they were [considered ambiguous](https://www.gnu.org/licenses/identify-licenses-clearly.html). We need to use `GPL-3.0-only` due to the inclusion of VST3, even though most of our code is compatible with `GPL-2.0-or-later`. @LWinterberg, sorry I missed this in my review.

Issue #2588, opened 2022-02-24
Make the "Warning" about saving project less alarming


SteveDaulton:
**Is your feature request related to a problem? Please describe.**
Some new users see the "[Warning](https://manual.audacityteam.org/man/warnings_preferences.html)" about saving projects as an error, and get frustrated and confused that they can't save a project without getting this "error".

**Describe the solution you'd like**
The message about saving projects was added because many users were saving projects when they really wanted to export an audio file. The explanatory  messages went a long way towards solving that problem, but it alarms some users that interpret the "warning" as an "error" when really it is just "information".

**Describe alternatives you've considered**

**Additional context**

Suggest changing this:
![warning](https://user-images.githubusercontent.com/11681097/155507485-198bd5ea-e35e-4cc1-bdcb-1d673bbb5132.png)

to this (or similar):
![info](https://user-images.githubusercontent.com/11681097/155507580-92416176-e42c-4036-90dc-be4a7989b638.png)

shubhangi013:
I'd like to work on this issue!

LWinterberg:
go ahead, @shubhangi013!

CoderHuman:
I'd also like to work on this issue.

LWinterberg:
go ahead as well! @CoderHuman 

This issue seems to be just renaming a string (or perhaps exchanging a wxWidgets element?), so having a look through the code to understand what happens where is beneficial to everyone looking to contribute :)

SteveDaulton:
> This issue seems to be just renaming a string

A bit more than that. The "Warning" string in the title bar is hard coded into the WarningDialog class.
See: [/src/widgets/Warning.cpp](https://github.com/audacity/audacity/blob/master/src/widgets/Warning.cpp)

CoderHuman:
> go ahead as well! @CoderHuman
> 
> This issue seems to be just renaming a string (or perhaps exchanging a wxWidgets element?), so having a look through the code to understand what happens where is beneficial to everyone looking to contribute :)

Thanks! I have been trying to find a way around this and have understood the codebase from a bird's eye view. 

My approach would be to replace the "Warning" string using a compatible string variable and check if the internalDialogName variable is equal to "FirstProjectSave". If found true, I will set the string variable to "Information". 

But like @SteveDaulton said, the "Warning" string in the title bar is hard coded and I cannot find a way to replace it using a string variable. 

I would appreciate both of your (@LWinterberg and @SteveDaulton) help in my following questions:

[...]

PR #2732, opened 2022-03-25
Changed string part from warning to information

Dhriti03:

Recommended:

### DIFF
--- a/src/widgets/Warning.cpp
+++ b/src/widgets/Warning.cpp
@@ -52,14 +52,14 @@ END_EVENT_TABLE()
 
 const TranslatableString &DefaultWarningFooter()
 {
-   static auto result = XXO("Don't show this warning again");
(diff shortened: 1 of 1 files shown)


LWinterberg:
Is this supposed to resolve #2588? If so, it changes the text near the checkbox but not the window title.

Dhriti03:
> Is this supposed to resolve #2588? If so, it changes the text near the checkbox but not the window title.

So isn't it like the line number 55 was dedicated for the text near the  checkbox and line 62 for the window title

LWinterberg:
Ah, sorry, I misread. What you're doing here is you're changing *all* warnings to say Information. However, the bug is only one specific case.

Dhriti03:
So which means I need to have a soft code for warning part to be converted to information with a condition statement  and having a string variable representing that string of warning is it ....wrt the class ...........please correct me with this ..

Issue #2716, opened 2022-03-22
Clean up top navigation bar


AnitaBats:
@LWinterberg could you please add a few words as a description for this task? 

@ksoze95 please contact @LWinterberg for designs and spec details

LWinterberg:
![image](https://user-images.githubusercontent.com/87814144/159689812-16d1553b-6076-4923-88cc-445ce96a03a2.png)

Legend: 
* Delete stuff crossed out in red
* Hide stuff (=disable by default) crossed out in pink
* Move stuff as described by green 
* Optional: Move stuff as described by yellow.

petersampsonaudacity:
### Upgrade from 3.1.3 to 3.2.0
There is an oddity if you upgrade from 3.1.3 (with the old toolbar layout and size) to 3.2.0.  The Window stays the old size with a very odd layout - and note that the Device toolbar still remains showing and is not hidden - and the Play-at-Speed toolbar remains in the top tooldock:
![image](https://user-images.githubusercontent.com/50205984/161114448-61e3e199-01dd-4aa9-aa2b-67d807bc91a1.png)

### After Reset Configuration
After doing a Reset configuration the new layout (sort of) obtains - but is cramped up as the new Window width is not implemented by the reset:
![image](https://user-images.githubusercontent.com/50205984/161115144-bf9012af-6fbc-413d-9c48-b22dd17a9947.png)


### Widen the window
If you then click&drag on the right to enlarge the window only then di you get the intended layout - the one you get with a virgin install of 3,2,0
![image](https://user-images.githubusercontent.com/50205984/161115575-1469dfff-33f6-4adf-9313-c020cf65957a.png)

dozzzzer:
Re-opened. @ksoze95 please see the comment above.
In addition, please move **the playback meter toolbar** to the right as per @LWinterberg 's comment above:

<img width="1033" alt="image" src="https://user-images.githubusercontent.com/70063970/161264415-e053a6c7-5e8b-49c9-a347-f88fe6fbe42c.png">

LWinterberg:
playback meter was optional to this task and has been spun off into #2757

petersampsonaudacity:
There is an underlying issue which maybe a separate issue in its own right:

**Tools > Reset Configuration** does **_not_** reset the Audacity window size to default size - perhaps it should,

Do we need a separate issue to be logged for this?

dozzzzer:
> Do we need a separate issue to be logged for this?
[...]

Issue #2717, opened 2022-03-22
Audio setup menu rework


AnitaBats:
@LWinterberg could you please add a few words as a description for this task?

@ksoze95 please contact @LWinterberg for designs and spec details

LWinterberg:
This is the audio settings dropdown. 

![image](https://user-images.githubusercontent.com/87814144/159690376-0da9387f-a69a-4ad0-a16c-883469a69599.png)

DavidBailes:
@LWinterberg .  Will the user interactions with this dropdown be the same as a standard dropdown menu?

LWinterberg:
Yes-ish. Ideally it should stay open even after clicking on any action, and only close when you click outside it or click the "Audio Settings..." entry.

DavidBailes:
It doesn't look like a wxWidgets menu, because it contains headings for the various sections. And if it were to stay open after clicking an item, then it really isn't a menu. In addition, I don't know how it could be made accessible for users of screen readers, as headings in menus aren't part of the msaa api.
Why not just have a dialog instead - then users can change as many things as they want before closing it, and the accessibility would be straigtforward.

DavidBailes:
> How does the current version deal with the devices toolbar and the devices settings?

Here's NVDA's speech output when tabbing along the device toolbar:

Audio Host  combo box  Windows WASAPI  collapsed
Recording Device  combo box  Microphone (USB Audio Device)  collapsed
Recording Channels  combo box  1 (Mono) Recording Channel  collapsed
Playback Device  combo box  Speakers (Realtek High Definition Audio)  collapsed

And tabbing through device settings in Preferences:

Interface  grouping
Host:  combo box  Windows WASAPI  collapsed  Alt+h
Playback  grouping
Device:  combo box  Speakers (Realtek High Definition Audio)  collapsed  Alt+d
Recording  grouping
Device:  combo box  Microphone (USB Audio Device)  collapsed  Alt+v
Channels:  combo box  1 (Mono)  collapsed  Alt+n

On Windows you can start and stop the Narrator screen reader by pressing Ctrl+Windows Key + Enter, if you want to check this out for yourself. There are minor differences between Jaws, NVDA, and Narrator.

PR #2744, opened 2022-03-29
Top navigation bar cleanup

ksoze95:
Resolves: #2716

Recommended:


vsverchinsky (approved):

Paul-Licameli (changes requested):
I'm trying to build each commit.

At the second commit, it built, but running Audacity, I noticed the wrong images were shown in the edit toolbar.  The old Cut image was on the button that still correctly did Trim, and so on.

The CMake Dependencies are not set up so that the *AsCeeCode.h files are regenerated when images are only deleted.

Deleteing the generated *AsCeeCode.h files and rebuilding does fix it.

How can the rebuilding be made automatic without requiring the developer to do that?

I think I have an answer, which I may give by pushing a commit onto this branch.  I suggest you then interactively rebase, permuting that commit to the start of the sequence.

Paul-Licameli:
I confirmed that my added commit works to do what I want, when it is at the appropriate place in the sequence of commits.

Please fetch this commit, and do interactive rebase to put it first in the sequence of commits, then force-push this branch again.

Paul-Licameli:
Otherwise, this PR looks good.

I would only question whether changing the default width of Audacity (when started with a fresh audacity.cfg files), to 1200 is acceptable to designers.

Also I need View > Toolbars > Reset Toolbars if I restart Audacity without a fresh audacity.cfg in order to see the desired changes of layout.  Old users of Audacity migrating to the new version might observe the same.

Hm, what's the fix for that?

LWinterberg:
> I would only question whether changing the default width of Audacity (when started with a fresh audacity.cfg files), to 1200 is acceptable to designers.

It is, because:

* #2717 and the vertical volume meter may require further changes to the default width, so this is an acceptable temporary solution until we know just how wide things need to be
* Even if this temporary change becomes a permanent one, I think the only somewhat common desktop resolutions smaller than 1200 would be 800x600 and 1024x768 - and both of those are quite vintage. 
* And even if someone finds out their screen is too small for the new default, they will be able to just resize it.

> Also I need View > Toolbars > Reset Toolbars if I restart Audacity without a fresh audacity.cfg in order to see the desired changes of layout. Old users of Audacity migrating to the new version might observe the same. Hm, what's the fix for that?

This is indeed tricky. The Audacity installer has a toggle to reset configuration, though that of course is quite nuclear. Maybe there should be another toggle, checked by default, saying "switch to new default layout" which is essentially the View > Toolbars > Reset Toolbars?

Paul-Licameli (approved):

PR #2747, opened 2022-03-30
Build before the fix + extra modification that triggers the bug more often

pietro68:

Recommended:


Paul-Licameli:
This pull request is not meant for merge, right?   It was only for making temporary builds available to QA.

You could do this with no pull request.  Just push a branch to your fork, find the build of it on the Actions tab, and send that link to others.

pietro68:
Yes, no merging intended - I did not know the build would be done also without a PR, good to know thanks!

PR #2749, opened 2022-03-30
changed warning to  information

Dhriti03:

Recommended:

### DIFF
--- a/src/widgets/Warning.cpp
+++ b/src/widgets/Warning.cpp
@@ -34,6 +34,7 @@ class WarningDialog final : public wxDialogWrapper
  public:
    // constructors and destructors
    WarningDialog(wxWindow *parent,
+                 char message1= ("Information") ,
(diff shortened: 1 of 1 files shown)


src/widgets/Warning.cpp
```diff
@@ -52,14 +53,14 @@ END_EVENT_TABLE()
 
 const TranslatableString &DefaultWarningFooter()
 {
-   static auto result = XXO("Don't show this warning again");
+   static auto result = XXO("Don't show this " + message1 + "again");
```
LWinterberg, line 56: avoid chopping up strings, they're a pain to translate. I believe this way would be better:
`wxString::Format( XXO("Don't show this %s again"), message1);`
see also https://github.com/audacity/audacity/wiki/Strings-&-translatable-code

LWinterberg (changes requested):

PR #2751, opened 2022-03-30
Changed line no. 55 & 56 with  wxString format

Dhriti03:

Recommended:

### DIFF
--- a/src/widgets/Warning.cpp
+++ b/src/widgets/Warning.cpp
@@ -34,6 +34,7 @@ class WarningDialog final : public wxDialogWrapper
  public:
    // constructors and destructors
    WarningDialog(wxWindow *parent,
+                 char message1= ("Information") ,
                  const TranslatableString &message,
                  const TranslatableString &footer,
                  bool showCancelButton);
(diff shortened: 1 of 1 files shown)


Dhriti03:
Now y is it giving an error ...........Please can u help me out with this @LWinterberg

LWinterberg:
from the linux output:

```
/home/runner/work/audacity/audacity/src/widgets/Warning.cpp:37:23: note:   no known conversion for argument 2 from ‘const TranslatableString’ to ‘char’
   37 |                  char message1= ("Information") ,
      |                  ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
```
I think the issue is some of the following: 
* You are using char instead of TranslatableString
* &DefaultWarningFooter() is now not returning anything

You may find some help from the doxygen documentation ([warning dialog](https://doxy.audacityteam.org/class_warning_dialog.html), [translatable string](https://doxy.audacityteam.org/class_translatable_string.html)) or by just googling it. I'm afraid I can't help you much since I'm not a programmer. 

You may also want to compile Audacity locally before submitting a PR.

Paul-Licameli:
Remember some languages have declension of nouns.  The string “information” by itself does not tell the translator that it will be the object of the verb in the other string.

i would just use two strings that are the complete sentences with this one word difference.

Dhriti03:
> Remember some languages have declension of nouns. The string “information” by itself does not tell the translator that it will be the object of the verb in the other string.
> 
> i would just use two strings that are the complete sentences with this one word difference.
Can you please give an example for this

Dhriti03:
Also where can we ask doubts related to these contributions please can you let me know

LWinterberg:
> Can you please give an example for this

```
"don't show this information again"
"don't show this warning again"
```

> Also where can we ask doubts related to these contributions please can you let me know

https://audacityteam.org/devserver

anyway: This issue got solved by #2756. No need to continue with this one.

PR #2753, opened 2022-03-31
Realtime effects use effect instance

Paul-Licameli:
Resolves: #2754

All is in place for different occurrences of an effect in (the same, or different) effect stacks can have independent state.

But work remains to rewrite each subclass of effect so it is really stateless and really makes independent instances.

Recommended:


pietro68:
These commits are the first 8 commits in https://github.com/audacity/audacity/pull/2707 which I have already approved? so I could approve this PR too.

pietro68 (approved):
Looks like all the commits in this PR are contained in https://github.com/audacity/audacity/pull/2707 too, which I already reviewed - so I can approve this too.

PR #2756, opened 2022-04-01
Make the "Warning" about saving project less alarming - issue:2588

kuouu:
Resolves: #2588

Add boolean parameter to WarningDialog to check if the title should be "Information" or not.

Recommended:


LWinterberg:
image
fix looks like it's working.

Pokechu22:
You might also want to change the line with `SetIcon(wxArtProvider::GetIcon(wxART_WARNING, wxART_MESSAGE_BOX));` to use `wxART_INFORMATION` in this case.

erickuo5124:
> You might also want to change the line with `SetIcon(wxArtProvider::GetIcon(wxART_WARNING, wxART_MESSAGE_BOX));` to use `wxART_INFORMATION` in this case.

Fixed. 
But the icon in WarningDialog seem not appear in macOS, so I can not check if it can correctly show up.

PR #2759, opened 2022-04-02
Stateless effect utilities

Paul-Licameli:
Resolves: #2758 

Utility classes to simplify the transition of built-in effects to statelessness.

Classes StatefulEffectBase and StatefulPerTrackEffect encapsulate the reliance of EffectInstance classes on their effects to
hold state.  They are used by default for all the built-ins.  But the dependencies of individual effects on these classes can be
broken.  EffectDtmf will be the first example.

Recommended:


pietro68:
Looks like all the commits in this PR are contained in https://github.com/audacity/audacity/pull/2707 too, which I already reviewed? I could then approve this too.

Issue #2725, opened 2022-03-24
Vocal Reduction Remove Vocals (mono) bug


junaayd123:
**Describe the bug**
So in some songs, mostly mp3's, there is a bug when removing vocals (mono). It, no matter what, removes the default ranges and strength and converts to mono. In other words, it will remove the default preset ranges and strength, and no matter what, its not changing the sound if I change the ranges.

**To Reproduce**
note: make sure you use an mp3
1. Go to 'Vocal Reduction and Isolation'
2. Click on 'Remove Vocals to Mono'
3. Change ranges and strength
4. Vocal remove
5. Try to use a different range and see if it changes the sound.

**Expected behavior**
I expected it to do the same vocal removing and backing stuff, like with Remove Vocals, and then make it mono.

**Screenshots**

**Additional information (please complete the following information):**
 - OS: macOS Monterey
 - Version: Audacity 3.1.3
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

**Additional context**

MridulTi:
It does change the sound but not to the extent that 'Remove vocals' does

MridulTi:
Yes, you are right I can also reproduce the issue. 
The issue is 'Remove Vocal to mono' won't change the sound if the ranges are changed too. It works fine when ranges are not changed though.

MridulTi:
@LWinterberg  can you label this accordingly

john-g-h-doe:
Some additional info:
- "Strength" has no effect for the method "Remove Vocals: to mono"
And this is normal, for this algorithm there is no "Strength" parameter.
- "Low Cut for Vocals" and "High Cut for Vocals" have an effect
e.g. put both to 1Hz => vocals not removed, then put "Low Cut" to 24000Hz and "High Cut" to 1Hz => vocals removed again.

Will follow a pull request that adds the "Strength" parameter to the "Remove Vocals: to mono" method, so it will behave similarly to the "Remove Vocals" method.

SteveDaulton:
This is not a bug. The behaviour is as intended and as described in the manual: https://manual.audacityteam.org/man/vocal_reduction_and_isolation.html

LWinterberg:
I don't quite understand what the issue is here. 

@junaayd123 @MridulTi @john-g-h-doe Aside from that it's somewhat silly (but probably technically necessary) to show a slider that doesn't do anything - what do you expect this slider to do?

[...]

PR #2760, opened 2022-04-03
fix bug #2725: Vocal Reduction Remove Vocals (mono) bug

john-g-h-doe:
Resolves: https://github.com/audacity/audacity/issues/2725

Add the "Strength" parameter to the "Remove Vocals: to mono" method, so it will behave similarly to the "Remove Vocals" method.

Recommended:


SteveDaulton:
I'm not keen on changing the behaviour of a long established effect. The change proposed here is incompatible with the current effect, and any macros that include this effect will be broken by this change.

PR #2766, opened 2022-04-05
Extract lib history

Paul-Licameli:
Resolves: #1129

Recommended:


src/Lyrics.h
```diff
@@ -102,8 +101,8 @@ class LyricsPanel final
    void SetLyricsStyle(const LyricsStyle newLyricsStyle);
 
    void Update(double t);
-   void UpdateLyrics(wxEvent &e);
-   void UpdateLyrics();
+   void UpdateLyrics(struct UndoRedoMessage);
```
ksoze95, line 104: The struct keyword is not needed.

src/Menus.h
```diff
@@ -113,8 +114,9 @@ class AUDACITY_DLL_API MenuManager final
    void TellUserWhyDisallowed(const TranslatableString & Name, CommandFlag flagsGot,
       CommandFlag flagsRequired);
 
-   void OnUndoRedo( wxCommandEvent &evt );
+   void OnUndoRedo(struct UndoRedoMessage);
```
ksoze95, line 117: Same a in Lyrics.h, struct keyword is not needed.

libraries/lib-project-history/UndoManager.cpp
```diff
@@ -25,27 +25,37 @@ UndoManager
...
-wxDEFINE_EVENT(EVT_UNDO_PURGE, wxCommandEvent);
+namespace {
+   using Savers = std::vector<UndoRedoExtensionRegistry::Saver>;
+   static Savers &GetSavers()
+   {
+      static Savers theSavers;
+      return theSavers;
+   }
+
+   UndoState::Extensions GetExtensions(AudacityProject &project)
```
ksoze95, line 45: This method could be made const.

src/TrackPanel.h
```diff
@@ -100,7 +100,7 @@ class AUDACITY_DLL_API TrackPanel final
    void OnProjectSettingsChange(wxCommandEvent &event);
    void OnTrackFocusChange( wxCommandEvent &event );
 
-   void OnUndoReset( wxCommandEvent &event );
+   void OnUndoReset(struct UndoRedoMessage);
```
ksoze95, line 103: Same a in Lyrics.h, struct keyword is not needed.

libraries/lib-project-history/ProjectHistory.h
```diff
@@ -12,23 +12,34 @@ Paul Licameli split from ProjectManager.h
...
+
    static ProjectHistory &Get( AudacityProject &project );
    static const ProjectHistory &Get( const AudacityProject &project );
 
    explicit ProjectHistory( AudacityProject &project )
       : mProject{ project }
    {}
-   ProjectHistory( const ProjectHistory & ) PROHIBITED;
-   ProjectHistory &operator=( const ProjectHistory & ) PROHIBITED;
+   ProjectHistory( const ProjectHistory & ) = delete;
```
ksoze95, line 41: Does the move constructor and move assignment operator need to be deleted as well?

src/SqliteSampleBlock.cpp
```diff
@@ -1008,6 +1011,57 @@ void SqliteSampleBlock::CalcSummary(Sizes sizes)
...
+      &seen);
+   }, begin, end );
+   return mayDelete.size();
+}
+
+void SqliteSampleBlockFactory::OnBeginPurge(size_t begin, size_t end)
+{
+   // Install a callback function that updates a progress indicator
+   using namespace BasicUI;
+   auto pDialog = std::shared_ptr<ProgressDialog>{ MakeProgress(
```
ksoze95, line 1050: As discuss, this fails to build when running under Visual Studio 2019 IDE.

Had to change to std::shared_ptr<BasicUI::ProgressDialog> for it to compile.

ksoze95 (changes requested):
The Linux builds seems to have failed.

Need to fix SqliteSampleBlock.cpp so that it can be built with Visual Studio 2019 from within the IDE.

Paul-Licameli:
> 

I think the Linux build failures are spurious.  Dmitry told me to ignore similar in a recent PR.

src/SqliteSampleBlock.cpp
ksoze95: Removing the ProgressDialog.h, from AudacityHeaders.h did the trick.

ksoze95: Removing the ProgressDialog.h, from AudacityHeaders.h did the trick.

src/Lyrics.h
Paul-Licameli: Incorrect, actually.  Try removing `struct`.  It won't compile.

It will compile without `struct` only if a previous line in the header has
```
struct UndoRedoMessage;
```

As I wrote it, it is an example of an "elaborated type" acting as a forward declaration introducing the type name into the compilation unit.

src/Menus.h
Paul-Licameli: Also incorrect for the same reason.

libraries/lib-project-history/ProjectHistory.h
Paul-Licameli: Good idea maybe but incidental to the purpose of this PR

src/Lyrics.h
ksoze95: I see.

ksoze95 (approved):

PR #2772, opened 2022-04-07
Make echo stateless

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2755

Step-by-step conversion of the existing Echo effect into a stateless effect.
The step numbers mentioned in the commit messages, refer to steps in a [guide I am writing](https://github.com/audacity/audacity/files/8441661/how.to.make.effects.stateless.txt)

Recommended:
NO - only the last commit preserves original behavior, the other commits will build, but the existing behavior is not preserved there.

### DIFF
--- a/src/effects/Echo.cpp
+++ b/src/effects/Echo.cpp
@@ -21,6 +21,9 @@
 
 
 #include "Echo.h"
+
+#if 0
+
 #include "LoadEffects.h"
 
 #include <wx/intl.h>
@@ -158,3 +161,5 @@ EffectEcho::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess &)
    S.EndMultiColumn();
    return nullptr;
 }
+
+#endif
--- a/src/effects/Echo.cpp
+++ b/src/effects/Echo.cpp
@@ -232,12 +232,25 @@ void EffectEcho::Validator::PopulateOrExchange(ShuttleGui& S, const EffectSettin
 }
 
 
-#if 0
-
-std::unique_ptr<EffectUIValidator>
-EffectEcho::PopulateOrExchange(ShuttleGui& S, EffectSettingsAccess&)
+// have do-nothing implementations for the time being
+bool EffectEcho::Validator::ValidateUI()
 {
-   // TODO later
+   return true;
 }
 
-#endif
+bool EffectEcho::Validator::UpdateUI()
+{
+   return true;
+}
+
+
+std::unique_ptr<EffectUIValidator>
+EffectEcho::PopulateOrExchange(ShuttleGui& S, EffectSettingsAccess& access)
+{
+   auto& settings = access.Get();
+   auto& myEffSettings = GetSettings(settings);
+   auto result = std::make_unique<Validator>(*this, access, myEffSettings);
+   result->PopulateOrExchange(S, settings, mProjectRate);
+   return result;   
+}
+
--- a/src/effects/Echo.cpp
+++ b/src/effects/Echo.cpp
@@ -50,7 +50,7 @@ EffectEcho::MakeInstance(EffectSettings& settings) const
 }
 
 
-#if 0
+
 
 #include "LoadEffects.h"
 
@@ -75,7 +75,7 @@ namespace{ BuiltinEffectsModule::Registration< EffectEcho > reg; }
 
 EffectEcho::EffectEcho()
 {
-   Parameters().Reset(*this);
+   //Parameters().Reset(*this);
    SetLinearEffectFlag(true);
 }
 
@@ -116,6 +116,8 @@ unsigned EffectEcho::GetAudioOutCount() const
 {
    return 1;
 }
+
+#if 0
 
 bool EffectEcho::ProcessInitialize(
    EffectSettings &, sampleCount, ChannelNames)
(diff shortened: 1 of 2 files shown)


Paul-Licameli:
I don't like it that the intermediate steps fail to build and run.

I would prefer a modified sequence (of four steps), even if this transformation changes certain lines of code more than once, and even if the final result is the same.  Verify that the Echo effect continues to work unsurprisingly at each step.  That includes persistency of the last used settings -- memory of the last used values within the session, also if you exit and restart the program.  This will demonstrate a better understanding now and if the code is ever reexamined much later.

- Define EffectEchoSettings (with default member initializers)
  - Base class is `EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>` and `FetchParameters` is removed
  - REMOVE the members of the effect that go into the settings.  REPLACE instead with `EffectEchoSettings mSettings;` in the effect and then do what is needed to fix compilation and running:
  - Update pointers-to-members in EffectParameter declarations
  - Insert sufficient `auto& echoSettings = mSettings;` (which changes later) and `echoSettings.` (which won't need more change) to fix compilation
  - Remove call to Reset() in constructor
  - ADD THIS LINE in EffectEcho::PopulateOrExhchange:  `echoSettings = GetSettings(access.Get());`.  And why?  Because in this transitional state, MakeSettings() does have a non-default definition (generated by the template EffectWithSettings), so EffectManager IS storing externalized settings and NOT reading and writing the copy in the Effect object.  So settings must be copied into the effect when the dialog opens, if persistency of the last-used Echo settings is to remain correct.  That is, so that "Repeat last effect" has the correct memory of settings, or if you exit and restart the program.

- Define validator
  - add `struct Validator;` to the Effect class, and class definition in the .cpp with mSettings; and member functions
  - The mSettings; member is EffectEchoSettings & transitionally -- NOT yet independently stored values.  The comment in the constructor should not say "Copy Settings" because it's not yet really copied.  You only bind a reference.
  - The constructor argument is non-const reference
  - The Effect's PopulateOrExchange has the line `auto& myEffSettings = mSettings;`, transitionally
  - The Validator's `PopulateOrExchange` need not take rate as an argument -- that was needed for DTMF but will not be for every effect
  - Give complete `UpdateUI` and `ValidateUI` implementations
  - Preserve the assignment of EffectEchoSettings still in EffectEcho::PopulateOrExchange
 
- Define Instance class
  - In the effect header, `struct Instance;` and declare `MakeInstance` and remove state data members and `Process*` functions
  - Define the Instance struct completely in the .cpp file
  - Insert `Instance::` and replace `Effect::` where needed
  - The Finalize() function can be simplified.  It doesn't need redundant calls like reset() to free memory.  The Instance is destroyed soon after use (unlike the case when the state was in the effect).
  - Understand that the Instance object IS now using externalized settings, which it does not modify, and the framework supplies them

- Real statelessness of EffectEcho
  - Make the base `EffectWithSettings<EffectEchoSettings, PerTrackEffect>` (now that MakeInstance has its override, we don't use StatefulPerTrackEffect)
  - delete EffectEcho::mSettings; and other needed changes:
  - Fix EffectEcho::PopulateOrExchange, which no longer assigns the contents of myEffSettings inside access.Get()
  - Validator changes its member to non-reference and its constructor argument to const reference

pietro68:
I am having a problem with the last sub-bullet point of the first bullet point "ADD THIS LINE in EffectEcho::PopulateOrExhchange.."

You wrote "echoSettings = GetSettings(access.Get());" now, I suppose you meant, at the beginning of the method, and then the method will reference &echoSettings.delay etc.

What is missing is the type declaration.

`auto echoSettings = GetSettings(access.Get());`  builds, but clearly will not work
`auto& echoSettings = GetSettings(access.Get());` build breaks at the usage points e.g. &echoSettings.delay 

The best I could come up with, is this:

```
mSettings = GetSettings(access.Get());
auto& echoSettings = mSettings;

```
Builds & works, but the last set settings will not be remembered when restarting Audacity. 
I guess I need a way to send them back to the framework, after I close the dialog by pressing OK, how do I do that?

Paul-Licameli:
> I am having a problem with the last sub-bullet point of the first bullet point "ADD THIS LINE in EffectEcho::PopulateOrExhchange.."
> 
> You wrote "echoSettings = GetSettings(access.Get());" now, I suppose you meant, at the beginning of the method, and then the method will reference &echoSettings.delay etc.
> 
> What is missing is the type declaration.
> 
> `auto echoSettings = GetSettings(access.Get());` builds, but clearly will not work `auto& echoSettings = GetSettings(access.Get());` build breaks at the usage points e.g. &echoSettings.delay
> 
> The best I could come up with, is this:
> 
> ```
> mSettings = GetSettings(access.Get());
> auto& echoSettings = mSettings;
> ```

I intended an assignment to echoSettings, AFTER you declarate it as a reference to mSettings.  So your last rewrite was it.

Paul-Licameli:
> 
> Builds & works, but the last set settings will not be remembered when restarting Audacity. I guess I need a way to send them back to the framework, after I close the dialog by pressing OK, how do I do that?

Put a breakpoint in Effect::SaveSettings to understand what happens.

Paul-Licameli:
> > Builds & works, but the last set settings will not be remembered when restarting Audacity. I guess I need a way to send them back to the framework, after I close the dialog by pressing OK, how do I do that?
> 
> Put a breakpoint in Effect::SaveSettings to understand what happens.

Actually...  now I think I understand why the persistency of effect settings will be broken at step 1, but repaired when you implement the validator again at step 2.

I could write the general fix for step 1, but it may not be worth the effort.

PR #2777, opened 2022-04-08
Don't link against lib-ffmpeg-support-interface when ffmpeg is disabled

imciner2:
The lib-ffmpeg-support-interface can be disabled (in which case no library is created with that name). So we shouldn't try to link the application against that library when no ffmpeg support is included.

I found this by trying to compile the master branch with the option `-Daudacity_use_ffmpeg=off`.

Recommended:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1441,12 +1441,17 @@ set( AUDACITY_LIBRARIES
    lib-uuid-interface
    lib-screen-geometry-interface
    lib-project-rate-interface
-   lib-ffmpeg-support-interface
    lib-theme-resources-interface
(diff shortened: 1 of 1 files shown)


imciner2:
@Paul-Licameli, Dmitry in the Discord chat said this is probably something you need to review.

Paul-Licameli:
No problems.

Paul-Licameli (approved):

Paul-Licameli:
Except there are failures in the Linux CI builds but the details in the logs seem unrelated to your changes. @crsib  ?

imciner2:
Yea, I think those CI errors are also appearing in other PRs as well.

Issue #2034, opened 2021-10-27
Audacity should not warn on closing empty projects


win93:
**Describe the bug**
The following message: "Save project before closing? / If saved, the project will have no tracks." Is kind of silly to me. Is there some actual user case where I would want to save an empty audacity project? This dialog message is annoying enough to spend the time to write this issue, especially when I don't fully understand the reason for having the dialog in the first place.

**To Reproduce**
1. Make a new track (ctrl+shift+n or cmd+shift+n)
2. Delete that track
3. Close the window (ctrl+w or cmd+w)
4. Receive screencapped dialog box

**Expected behavior**

I expect the application to close without prompting.

**Screenshots**
<img width="534" alt="image" src="https://user-images.githubusercontent.com/3725496/139122595-72a36b18-78ff-4ade-bb9e-c9cb5189ffd5.png">

**Additional information:**
 - OS: MacOS Catalina
 - Audacity Version 3.0.4

Connorbub:
Has this issue been assigned? And if not, could I look into it?

Pokechu22:
I don't think anyone else is currently working on it, so go ahead.

SteveDaulton:
> Is there some actual user case where I would want to save an empty audacity project? 

Yes there is.
When a project has never been saved, the project files are written in Audacity's temp folder. For regular users of Audacity that have a lot of Audacity projects stored to disk, it can be handy to save the (empty) project before doing anything else. This allows the user to decide where the project will be written.

For example, before beginning a very long recording, the user may want to save the empty project to their computer's secondary drive as that has a lot more free space than the primary (SSD) drive.

Also, once the project has been saved, saving can be done with "Ctrl + S", and no need to enter a target location as that has already been set by the initial save.

Warning about saving an empty project is a good idea in cases where there is undo data as this could prevent accidental deletion of a project. Remember that Audacity has a large community of blind users that can't see if the project has tracks.

> I suppose instead of fully removing it, it also would make sense to show a checkbox "don't show this again"

I think that would be the best solution as it still provides helpful information to those that need it, and removes the annoyance for those that don't.

Pokechu22:
For what it's worth, no prompt appears when closing an empty project if no changes have been made (just as with opening a project and not making any changes and then closing it).  The prompt only appears if undo data exists.

More interestingly, if you have an empty project, add a track to it, and then _undo_ adding that track (in comparison to deleting the track), then no prompt to save appears.

LWinterberg:
[...]

PR #2778, opened 2022-04-08
Fix issue #2034: warning when closing empty projects

connoryo:
Resolves: [https://github.com/audacity/audacity/issues/2034](https://github.com/audacity/audacity/issues/2034)

Added a "do not show this message again" box to the dialog popup when a user tries to close an empty project. I accomplished this by switching the box from an AudacityMessageBox to a wxRichMessageDialog, which allows for the use of a checkbox. This will hopefully avoid user frustration when they want to close an empty project, without getting rid of the dialog entirely so that users with a valid reason to save an empty project will remain satisfied.

Recommended:


Connorbub:
I wasn't sure if changing the type of the dialog from AudacityMessageBox to a more generic wxRichMessageDialog was a good idea for consistency's sake, but it does indeed do what we want it to. Let me know if it would be better to use something else instead.

Issue #2780, opened 2022-04-08
Effect Amplify: "New Peak Amplitude" has garbage value at startup


pietro68:
**Describe the bug**
When using an Amplify effect, the "New Peak Amplitude" field will show a garbage value

**To Reproduce**
1. Start Audacity
2. create a noise track, select a bit of it
3. Effect | Amplify...

**Actual behavior**
field "New Peak Amplitude" will show a garbage value, i.e. a value which has no connection with the peak of the selected audio and the amplification value

**Expected behavior**
the above field should show what would be the new peak value if the selected amplification would be applied (hint: move the slider slightly, and bring it back to its initial position, to see what the value should be)

**Screenshots**
![Capture](https://user-images.githubusercontent.com/16491307/162422866-470bde5e-f9bf-4a70-93f5-9030d0386598.PNG)

**Additional information (please complete the following information):**
 - OS: Win10
 - Version: happens on master, commit hash 793cd28

petersampsonaudacity:
@pietro68   this is a duplicate of #2700

SteveDaulton:
The proposed fix does not reinstate the behaviour prior to this bug. Issue #2700 describes how the Amplify effect behaved in all previous versions of Audacity.

petersampsonaudacity:
> The proposed fix does not reinstate the behaviour prior to this bug. Issue #2700 describes how the Amplify effect behaved in all previous versions of Audacity.

See the Audacity Manual for the required behavior:
https://manual.audacityteam.org/man/amplify.html

where it clearly says:
>When the Amplify dialog first appears it shows in the Amplification (dB) box the amplification amount needed to create a New Peak Amplitude of 0 dB, thus maximizing the volume of the selected track. If this is your goal you can just click the OK button.
>
>Note that when applied to a single stereo track or to multiple tracks, Amplify shows the amplification needed to maximize the volume of the loudest track or channel and will apply that amount of amplification to all selected tracks and/or channels. In this way, Amplify preserves the relative volumes of the selected tracks and/or channels. This is different from the Normalize effect which can be optionally set to maximize the volume of each track or channel independently.

dozzzzer:
Tested.

- Win 10 x64
- [audacity-win-3.2.0-alpha-20220412+77aea2d-x64-msvc2022](https://github.com/audacity/audacity/suites/6087523642/artifacts/211013563)

Issue #2700, opened 2022-03-21
Amplify fails to default offer amplification to 0db


petersampsonaudacity:
**Describe the bug**
Amplify fails to default offer amplification to 0db

This is a regression on 3.1.3 - and a regression on the previous alpha master that I have:
audacity-win-3.2.0-alpha-20220302+b7708d1-x64-msvc2022

**To Reproduce**
1. Purge the Audacity settings folder (to get OOTB factory settings)
2. Launch latest 3.2.0 alpha master
3. Generate default 30 minute chirp - default max amplitude 0.8
4. Effect > Amplify
5. **Observe: amplify offers to amplify by -  -0.915 dB**
6. Press OK to apply the effect at default setting
7. **Observe: the negative amplification is applied**

**Expected behavior**
On invoking Amplify the default amplification offered by the effect should be that which is required to raise the maximum signal level to 0dB _(i.e. maximum volume)_

**Screenshots**
### with the latest 3.2.0 alpha master
![image](https://user-images.githubusercontent.com/50205984/159270891-314d88de-5052-48ed-96ec-460725410e6f.png)

### With 3.1.3
![image](https://user-images.githubusercontent.com/50205984/159271409-6fb2a5b1-f934-4064-8a0a-a5fc8eac556c.png)

**Additional information (please complete the following information):**
 - OS: Windows 10 - but also reported on Linux by @SteveDaulton  - so assume all OS
 - Version: latest alpha 3,2,0 master - audacity-win-3.2.0-alpha-20220321+a994066-x64-msvc2022

**Additional context**
This makes it very hard almost impossible for the user to achieve full maximum volume 0dB with the Amplify effect.

See also:
Normalize OK button is grayed out (on first use) #2701

petersampsonaudacity:
Testing on W10 with latest alpha master audacity-win-3.2.0-alpha-20220412+eb33420-x64-msvc2022

I confirm this is fixed  :-)

Paul-Licameli:
The problems for Normalize and Amplify will not bisect to the same place.  The Amplify one is more recent.

PR #2782, opened 2022-04-08
Fix for garbage value for New Peak Value

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2780
Resolves: #2700 

The fix was rather easy.

Recommended:

### DIFF
--- a/src/effects/Amplify.cpp
+++ b/src/effects/Amplify.cpp
@@ -229,6 +229,9 @@ EffectAmplify::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess &)
       }
    }
 
+   // At this point mNewPeak is still uninitialized; this will intialize it
+   ClampRatio();
+
    S.AddSpace(0, 5);
 
    S.StartVerticalLay(0);
--- a/src/effects/Amplify.h
+++ b/src/effects/Amplify.h
@@ -73,13 +73,13 @@ private:
    void CheckClip();
 
 private:
-   double mPeak;
+   double mPeak      = 1.0;
 
-   double mRatio;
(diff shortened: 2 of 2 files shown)


src/effects/Amplify.cpp
```diff
@@ -229,6 +229,13 @@ EffectAmplify::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess &)
       }
    }
 
+   // Determine the new peak, because at this stage it still has a garbage value;
+   // also, we can not use mRatio to do it (as done in other parts of the code)
+   // because at this stage mRatio is not yet reflecting the slider/textinput value
+   // (which is linked to mAmp)
+   //
+   mNewPeak = LINEAR_TO_DB(DB_TO_LINEAR(mAmp) * mPeak);
```
Paul-Licameli, line 237: 1. Can you verify that commit 7f63376534dc645b1c4b78a669c6ca4266975093 introduced the bug, which did not exist before?
2. Can you fix it instead by calling ClampRatio() to avoid duplication of the logic?
3. Also make an in-class initialized of mNewPeak to 0, and for other members while you can (mRatioClip, mPeak).  Better to initialize members of built-in number types always, and better to do that with member initializers, not assignments inside the constructor.  In-class initializers are best for assisting the human reader of the code to be confident that it is right.

Paul-Licameli (changes requested):

src/effects/Amplify.cpp
pietro68: 1. yes, verified

Paul-Licameli (approved):

Issue #2781, opened 2022-04-08
Effect Amplify: amplification factor not saved


pietro68:
**Describe the bug**
When using the Amplify effect, the last used amplification factor will not be remembered - the first time the effect is used in a fresh-started Audacity, it will always be equal to -0.915

@Paul-Licameli writes:  Also the default state of the "Allow Clipping" checkbox is on, which is an unintended change from 3.1.3

**To Reproduce**
1. Start Audacity
2. generate a noise track, select it
3. Effects | Amplify
4. choose some amplification factor using the slider (e.g. -6)
5. close Audacity
6. repeat steps 1-3

**Actual behavior**
The shown amplification factor will be -0.915

**Expected behavior**
The shown amplification factor should be what you chose at step 4

**Additional information (please complete the following information):**
 - OS: Win10
 - Version: master build on commit 793cd28
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

petersampsonaudacity:
> When using the Amplify effect, the last used amplification factor will not be remembered - 

That is intentional - each Time the Amplify effect is used it is supposed to calculate the amount of amplification required to take the maximum peak up to 0dB


>the first time the effect is used in a fresh-started Audacity, it will always be equal to -0.915

This is a regression bug that @SteveDaulton logged already (I'll try to locate it)

PR #2783, opened 2022-04-08
Fix for amplification factor not saved

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2781

Please see commit message for details.

Recommended:


Paul-Licameli:
#3529 now has a more correct fix for this regression

PR #2795, opened 2022-04-11
Reset configuration

ksoze95:
Resolves: #2716

Reset Configuration menu option now also resets the Audacity window to the default size.

Recommended:

### DIFF
--- a/src/ProjectManager.cpp
+++ b/src/ProjectManager.cpp
@@ -57,6 +57,23 @@ Paul Licameli split from AudacityProject.cpp
 #include "../images/AudacityLogoAlpha.xpm"
 #endif
 
+namespace {
+   void SaveWindowPreferences(const wxRect& windowRect, const wxRect& normalRect,
+                              bool isMaximized, bool isIconized) {
+      gPrefs->Write(ProjectWindow::Preference_X, windowRect.GetX());
+      gPrefs->Write(ProjectWindow::Preference_Y, windowRect.GetY());
+      gPrefs->Write(ProjectWindow::Preference_Width, windowRect.GetWidth());
+      gPrefs->Write(ProjectWindow::Preference_Height, windowRect.GetHeight());
+      gPrefs->Write(ProjectWindow::Preference_Maximized, isMaximized);
+      gPrefs->Write(ProjectWindow::Preference_Normal_X, normalRect.GetX());
+      gPrefs->Write(ProjectWindow::Preference_Normal_Y, normalRect.GetY());
+      gPrefs->Write(ProjectWindow::Preference_Normal_Width, normalRect.GetWidth());
+      gPrefs->Write(ProjectWindow::Preference_Normal_Height, normalRect.GetHeight());
+      gPrefs->Write(ProjectWindow::Preference_Iconized, isIconized);
+      gPrefs->Flush();
+   }
+}
+
 const int AudacityProjectTimerID = 5200;
 
 static AudacityProject::AttachedObjects::RegisteredFactory sProjectManagerKey {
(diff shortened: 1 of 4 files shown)


petersampsonaudacity:
On W10 I tested @ksoze95 's branch build:  audacity-win-3.2.0-alpha-20220411+3629298-x64-msvc2022 

Reset Configuration now appears to nicely reset the Window size to default in this branch build

src/ProjectManager.cpp
```diff
@@ -124,62 +124,50 @@ void ProjectManager::SaveWindowSize()
...
-         gPrefs->Write(wxT("/Window/Y"), defWndRect.GetY());
-         gPrefs->Write(wxT("/Window/Width"), defWndRect.GetWidth());
-         gPrefs->Write(wxT("/Window/Height"), defWndRect.GetHeight());
-         gPrefs->Write(wxT("/Window/Maximized"), FALSE);
-         gPrefs->Write(wxT("/Window/Normal_X"), defWndRect.GetX());
-         gPrefs->Write(wxT("/Window/Normal_Y"), defWndRect.GetY());
-         gPrefs->Write(wxT("/Window/Normal_Width"), defWndRect.GetWidth());
-         gPrefs->Write(wxT("/Window/Normal_Height"), defWndRect.GetHeight());
-         gPrefs->Write(wxT("/Window/Iconized"), FALSE);
+         SaveWindowPreferences(defWndRect, defWndRect, FALSE, FALSE);
```
Paul-Licameli, line 168: I like the elimination of duplication.

Let's be modern and use `true` and `false`.

src/ProjectManager.cpp
```diff
@@ -124,62 +124,50 @@ void ProjectManager::SaveWindowSize()
...
          foundIconizedProject =  TRUE;
-
    }
-   if (validWindowForSaveWindowSize)
-   {
-      wxRect windowRect = validProject->GetRect();
-      wxRect normalRect = validProject->GetNormalizedWindowState();
-      bool wndMaximized = validProject->IsMaximized();
+
+   auto SaveWindowPreferences = [](const auto& windowRect, const auto& normalRect, auto isMaximized, auto isIconized) {
```
Paul-Licameli, line 129: I would prefer not to use auto for argument types.

src/menus/PluginMenus.cpp
```diff
@@ -416,6 +416,8 @@ void OnResetConfig(const CommandContext &context)
    gPrefs->Write("/SelectionToolbarMode", 0);
    gPrefs->Flush();
    DoReloadPreferences(project);
+
+   ProjectWindow::OnResetWindow(context);
```
Paul-Licameli, line 420: Second commit looks good.

src/ProjectManager.cpp
ksoze95: I have swapped out TRUE/FALSE for true/false, and removed duplications of the string literals "/Window/X", etc.

src/ProjectManager.cpp
ksoze95: I have replaced the auto parameter types with actual types.

Change the SaveWindowPreferences lambda to a simple function instead.

Paul-Licameli (changes requested):
The duplication of string literals is fixed now.  I would squash the first two commits into one.

But wait!  Here is another good cleanup to do here too:

Define global StringSetting and BoolSetting objects (see existing examples), eliminating several direct uses of gPrefs.

Because, I also want all those uses to disappear ultimately.  This PR is an excuse to accomplish a piece of that.

Paul-Licameli (approved):

Paul-Licameli:
The code all looks good now with the use of settings objects!

The only small thing to correct is that there are some misspelled words in the second commit comment.

PR #2801, opened 2022-04-11
Make echo stateless 2

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2755

Second attempt, after it was advised in https://github.com/audacity/audacity/pull/2772 to organize changes in 4 commits, all building and not changing behavior

Recommended:

### DIFF
--- a/src/effects/Echo.cpp
+++ b/src/effects/Echo.cpp
@@ -42,6 +42,41 @@ const ComponentInterfaceSymbol EffectEcho::Symbol
 
 namespace{ BuiltinEffectsModule::Registration< EffectEcho > reg; }
 
+
+struct EffectEcho::Instance
+   : public PerTrackEffect::Instance
+   , public EffectInstanceWithBlockSize
+   , public EffectInstanceWithSampleRate
+
+{
+   Instance(const PerTrackEffect& effect)
+      : PerTrackEffect::Instance{ effect }
+   {}
+
+   bool ProcessInitialize(EffectSettings& settings,
+      sampleCount totalLen, ChannelNames chanMap) override;
+
+   size_t ProcessBlock(EffectSettings& settings,
+      const float* const* inBlock, float* const* outBlock, size_t blockLen)  override;
+
+   bool ProcessFinalize(void) override;
+
+   Floats history;
+   size_t histPos;
+   size_t histLen;
+};
+
+
+
+std::shared_ptr<EffectInstance>
+EffectEcho::MakeInstance(EffectSettings&) const
+{
+   return std::make_shared<Instance>(*this);
+}
+
+
+
+
 EffectEcho::EffectEcho()
 {
    SetLinearEffectFlag(true);
@@ -85,10 +120,13 @@ unsigned EffectEcho::GetAudioOutCount() const
    return 1;
 }
 
-bool EffectEcho::ProcessInitialize(
-   EffectSettings &, sampleCount, ChannelNames)
+bool EffectEcho::Instance::ProcessInitialize(
+   EffectSettings&, sampleCount, ChannelNames)
 {
-   auto& echoSettings = mSettings;
+   // temporary - in the final step this will be replaced by
+   // auto& echoSettings = GetSettings(settings);
+   //
+   auto& echoSettings = static_cast<const EffectEcho&>(mProcessor).mSettings;
 
    if (echoSettings.delay == 0.0)
    {
@@ -108,23 +146,25 @@ bool EffectEcho::ProcessInitialize(
       history.reinit(histLen, true);
    }
    catch ( const std::bad_alloc& ) {
-      Effect::MessageBox( XO("Requested value exceeds memory capacity.") );
+      mProcessor.MessageBox( XO("Requested value exceeds memory capacity.") );
       return false;
    }
 
    return history != NULL;
 }
 
-bool EffectEcho::ProcessFinalize()
+bool EffectEcho::Instance::ProcessFinalize()
 {
-   history.reset();
    return true;
 }
 
-size_t EffectEcho::ProcessBlock(EffectSettings &,
+size_t EffectEcho::Instance::ProcessBlock(EffectSettings&,
    const float *const *inBlock, float *const *outBlock, size_t blockLen)
 {
-   auto& echoSettings = mSettings;
+   // temporary - in the final step this will be replaced by
+   // auto& echoSettings = GetSettings(settings);
+   //
+   auto& echoSettings = static_cast<const EffectEcho&>(mProcessor).mSettings;
 
    const float *ibuf = inBlock[0];
    float *obuf = outBlock[0];
--- a/src/effects/Echo.cpp
+++ b/src/effects/Echo.cpp
@@ -141,6 +141,31 @@ size_t EffectEcho::ProcessBlock(EffectSettings &,
    return blockLen;
 }
 
+
+
+struct EffectEcho::Validator
+   : DefaultEffectUIValidator
+{
+   Validator(EffectUIClientInterface& effect,
+      EffectSettingsAccess& access, EffectEchoSettings& settings)
+      : DefaultEffectUIValidator{ effect, access }
+      , mSettings{ settings }
+   {}
+   virtual ~Validator() = default;
+
+   Effect& GetEffect() const { return static_cast<Effect&>(mEffect); }
+
+   bool ValidateUI() override;
+   bool UpdateUI() override;
+
+   void PopulateOrExchange(ShuttleGui& S);
+
+   EffectEchoSettings& mSettings;
+};
+
+
+
+
 std::unique_ptr<EffectUIValidator>
 EffectEcho::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess& access)
 {
@@ -163,3 +188,33 @@ EffectEcho::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess& access)
    S.EndMultiColumn();
    return nullptr;
 }
+
+
+bool EffectEcho::Validator::ValidateUI()
+{
+   mAccess.ModifySettings
+   (
+      [this](EffectSettings& settings)
+   {
+      // pass back the modified settings to the MessageBuffer
+
+      // TODO uncomment at last step
+      //EffectEcho::GetSettings(settings) = mSettings;
+   }
+   );
+
+   return true;
+}
+
+
+bool EffectEcho::Validator::UpdateUI()
+{
+   // get the settings from the MessageBuffer and write them to our local copy
+   const auto& settings = mAccess.Get();
+
+   // TODO uncomment at last step
+   //mSettings = GetSettings(settings);
+
+   return true;
+}
+
--- a/src/effects/Echo.cpp
+++ b/src/effects/Echo.cpp
@@ -121,13 +121,10 @@ unsigned EffectEcho::GetAudioOutCount() const
 }
 
 bool EffectEcho::Instance::ProcessInitialize(
-   EffectSettings&, sampleCount, ChannelNames)
+   EffectSettings& settings, sampleCount, ChannelNames)
 {
-   // temporary - in the final step this will be replaced by
-   // auto& echoSettings = GetSettings(settings);
-   //
-   auto& echoSettings = static_cast<const EffectEcho&>(mProcessor).mSettings;
-
+   auto& echoSettings = GetSettings(settings);
+  
    if (echoSettings.delay == 0.0)
    {
       return false;
@@ -158,14 +155,11 @@ bool EffectEcho::Instance::ProcessFinalize()
    return true;
 }
 
-size_t EffectEcho::Instance::ProcessBlock(EffectSettings&,
+size_t EffectEcho::Instance::ProcessBlock(EffectSettings& settings,
    const float *const *inBlock, float *const *outBlock, size_t blockLen)
 {
-   // temporary - in the final step this will be replaced by
-   // auto& echoSettings = GetSettings(settings);
-   //
-   auto& echoSettings = static_cast<const EffectEcho&>(mProcessor).mSettings;
-
+   auto& echoSettings = GetSettings(settings);
+   
    const float *ibuf = inBlock[0];
    float *obuf = outBlock[0];
 
@@ -187,7 +181,7 @@ struct EffectEcho::Validator
    : DefaultEffectUIValidator
 {
    Validator(EffectUIClientInterface& effect,
-      EffectSettingsAccess& access, EffectEchoSettings& settings)
+      EffectSettingsAccess& access, const EffectEchoSettings& settings)
       : DefaultEffectUIValidator{ effect, access }
       , mSettings{ settings }
    {}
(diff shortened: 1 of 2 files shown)


src/effects/Echo.cpp
```diff
@@ -196,16 +196,25 @@ struct EffectEcho::Validator
 
    void PopulateOrExchange(ShuttleGui& S, const EffectSettings& settings);
 
-   EffectEchoSettings& mSettings;
+   EffectEchoSettings mSettings;
```
Paul-Licameli, line 199: The steps aren't exactly as I described before, but that is all right.

Did you verify that each step in the transformation preserves the persistency of settings?

Maybe the second and fourth steps could be combined as the third step of three.

Because you never construct the Validator in the second commit -- it isn't necessary to make mSettings a reference and then change it to a value in the last step.

Paul-Licameli (changes requested):
Just one unnecessary line and function argument that can be removed

src/effects/Echo.cpp
```diff
@@ -196,16 +196,25 @@ struct EffectEcho::Validator
...
+   auto result = std::make_unique<Validator>(*this, access, myEffSettings);
+   result->PopulateOrExchange(S, settings);
+   return result;
+}
+
+
+void EffectEcho::Validator::PopulateOrExchange(ShuttleGui & S, const EffectSettings& settings)
 {
-   mSettings = GetSettings(access.Get());
+   mSettings = GetSettings(settings);
```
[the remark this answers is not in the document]
pietro68: That's right! the correction of this will be visible in the next force-push.

pietro68:
I pushed some of the requested changes. I still have to understand why settings are not saved and recalled in commits 1, 2 & 3

src/effects/Echo.cpp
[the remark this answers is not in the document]
Paul-Licameli: I still see the unnecessary assignment to mSettings.

Paul-Licameli:
> I pushed some of the requested changes. I still have to fully understand why settings are not saved and recalled in commits 1, 2 & 3.
> 
> So far I understood this: in those commits, the base class of Echo is temporarily set to `EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>`
> 
> I put a breakpoint in void `ShuttleGetAutomation::Define`; before and after those 3 commits, I see there that the values coming in, are the actual values set in the gui; with either of those 3 commits, I do not (they'd be the last saved ones)
> 
> Maybe this has to do with FetchParameters returning something wrong when the base class is `EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>`?

Let's be satisfied with persistency of settings in the final commit, even if it fails in the intermediate commits.

I think the reason for failure is that MakeSettings() is redefined when you use the base class EffectWithSettings, so EffectManager is storing a real settings object instead of an empty `any`.  Then there are calls to SaveSettings and LoadSettings which really use that object.  But while EffectEcho is not yet really stateless, it is still using its own internal copy of the settings instead of the one stored in EffectManager.

pietro68:
I see now, thanks for the explanation. 
What do you think is left to do here? would you like me to reduce the number of steps from 4 to 3, like you hinted before?

src/effects/Echo.cpp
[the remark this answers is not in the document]
pietro68: Did you look at the latest force-push?

there is an assignment `mSettings = GetSettings(access.Get());` in `EffectEcho::PopulateOrExchange` only intermediate steps, but I guess that one is needed.

in the last step there is no such assignment, in either  `EffectEcho::PopulateOrExchange` or `EffectEcho::Validator::PopulateOrExchange`

pietro68:
The last push you see is after I rebased on branch More-stateless-effect-utilities, in order to try using EffectWithSettingsBase to preserve settings in intermediate transformation steps (i.e. 1-2-3)

But what I found out, is that to preserve settings, I do not need that class. I just needed to correct FetchParameters to return the right address of the settings, and leave it there for all 3 intermediate steps. I verified it works.

Paul-Licameli:
I'm not persuaded that the transitional steps are really all correct.  They may run the effect without crashing, but I think that at step 3, the effect does not apply the correct values.

I could approve the PR anyway if the final result is correct, but I would prefer that you understand the intermediate steps well enough to know why bugs do or don't exist in them.

Paul-Licameli:
The third step does not apply settings correctly.

It might also be useful to use the Manage button to save a named preset.  Then verify that it is correctly retrieved when you load it, in each intermediate step.

To make the intermediate steps strictly correct, you should not use EffectSettings arguments of the Instance, and should store a reference not a copy of settings in the Validator, until the last commit changes that.  You should not use EffectWithSettings until the last commit, where you remove the mSettings member of the effect.

Paul-Licameli:
My transformation of Ladspa to statelessness does as I described, with other work too so there are more commits.  I also don’t need EffectWithSettingsBase at any point.

Paul-Licameli:
This also means the places where the Validator uses GetSettings are not added until the last commit.

pietro68:
I confirm that step 3 does not apply the chosen settings. I will edit each commit now, following these directives:

a) for steps 1-3, inherit from StatefulPerTrackEffect
b) for step 4, inherit from EffectWithSettings<EffectEchoSettings, PerTrackEffect>
c) in steps 1-3, have a mSettings as a member of Echo, and use exclusively that one
d) in steps 1-3 in the Validator, use a reference to the the settings, not a copy, but then in step 4 turn it to a copy

Is the above what you meant?

It is not really clear to me how in step 3 I can access the Effect's mSettings member from within the instance - with a downcast to EffectEcho of the instance's mProcessor member?

pietro68:
Some changes to the last agreed plan:

step 2 could not inherit `StatefulPerTrackEffect` but had to inherit `EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>` - otherwise Validator implementations have no access to `GetSettings()`

step 3 could not inherit `StatefulPerTrackEffect` but had to inherit `EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>` - otherwise there are no calls to the instance's Process*** methods

TESTS
- preservation of last used settings: all step pass

- switching between user presets: 
  steps 1-2-3 pass, 
  step 4 shows a bug: when having two user presets A and B in the list, clicking once on one of them is not enough to change the values, the preset must be clicked twice

pietro68:
I could reproduce the bug I noticed with the two user presets with DTMFgen too, so I conclude the cause is in the framework. I am going to open an issue to track that.

Paul-Licameli:
> Some changes to the last agreed plan:
> 
> step 2 could not inherit `StatefulPerTrackEffect` but had to inherit `EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>` - otherwise Validator implementations have no access to `GetSettings()`

No.

GetSettings() should not be used until the last step.  I think I said that already.  The parts of the Validator code that use GetSettings should wait until step 4 to be added.


> 
> step 3 could not inherit `StatefulPerTrackEffect` but had to inherit `EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>` - otherwise there are no calls to the instance's Process*** methods
> 

No.  Inheritance of the class changes only in the last step.


> TESTS
> 
> * preservation of last used settings: all steps pass
> * switching between user presets:
>   steps 1-2-3 pass,
>   step 4 shows a **bug**: when having two user presets A and B in the list, clicking once on one of them is not enough to change the values, the preset must be clicked twice.
> 
> Is this bug due to an error in assembling final step 4, or is the bug in the framework? I will have to inspect this tomorrow but if you have a hint, it is welcome!

pietro68:
Alright, I can make step 2 still inherit `StatefulPerTrackEffect` if I comment out the calls to GetSettings, not a big problem there.

But... to make step3 inherit StatefulPerTrackEffect - I would have to leave ProcessInitialize, ProcessFinalize and ProcessBlock still belong to the effect and not to the instance. That's ok actually, it is just a little more work postponed to step 4.

I found the preset switching bug also in DTMFGen, I opened this issue for it: https://github.com/audacity/audacity/issues/2822

src/effects/Echo.cpp
```diff
@@ -141,6 +141,31 @@ size_t EffectEcho::ProcessBlock(EffectSettings &,
...
+   virtual ~Validator() = default;
+
+   Effect& GetEffect() const { return static_cast<Effect&>(mEffect); }
+
+   bool ValidateUI() override;
+   bool UpdateUI() override;
+
+   void PopulateOrExchange(ShuttleGui& S);
+
+   EffectEchoSettings mSettings;
```
Paul-Licameli, line 199: I reverse my previous comment at this line.

Do NOT put the validator into its final form at this commit.  Make mSettings a reference at this step.  Change it to no reference only at the last step.

src/effects/Echo.cpp
```diff
@@ -163,3 +188,29 @@ EffectEcho::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess& access)
...
+
+
+bool EffectEcho::Validator::ValidateUI()
+{
+   mAccess.ModifySettings
+   (
+      [this](EffectSettings& settings)
+   {
+      // pass back the modified settings to the MessageBuffer
+      EffectEcho::GetSettings(settings) = mSettings;
```
Paul-Licameli, line 244: Do NOT add these lines in the second commit.  Only at the last commit.

src/effects/Echo.cpp
```diff
@@ -163,3 +188,29 @@ EffectEcho::PopulateOrExchange(ShuttleGui & S, EffectSettingsAccess& access)
...
+
+   return true;
+}
+
+
+bool EffectEcho::Validator::UpdateUI()
+{
+   // get the settings from the MessageBuffer and write them to our local copy
+   const auto& settings = mAccess.Get();
+   mSettings = GetSettings(settings);
```
Paul-Licameli, line 256: Again add this only at the last commit.

GetSettings should not be available until then.  The point where the echo settings move out of the effect and I to EffectSettings is only then.

src/effects/Echo.cpp
```diff
@@ -42,6 +42,43 @@ const ComponentInterfaceSymbol EffectEcho::Symbol
...
+
+struct EffectEcho::Instance
+   : public PerTrackEffect::Instance
+   , public EffectInstanceWithBlockSize
+{
+   Instance(const PerTrackEffect& effect)
+      : PerTrackEffect::Instance{ effect }
+   {}
+
+   void SetSampleRate(double rate) override { mSampleRate = rate; }
```
Paul-Licameli, line 54: Use EffectInstanceWithSampleRate to simplify.  Remove the member function and variable for mSampleRate.  Just inherit them.

src/effects/Echo.cpp
```diff
@@ -85,10 +122,13 @@ unsigned EffectEcho::GetAudioOutCount() const
    return 1;
 }
 
-bool EffectEcho::ProcessInitialize(
-   EffectSettings &, sampleCount, ChannelNames)
+bool EffectEcho::Instance::ProcessInitialize(
+   EffectSettings& settings, sampleCount, ChannelNames)
```
Paul-Licameli, line 126: Do NOT name and the settings argument in step 3.  Add that only at step 4.

And step 3, static_cast mProcessor to EffectEcho and use its mSettings.  At step 4, remove those lines, as you must to compile without the mSettings member.  Only then use the arguments.

src/effects/Echo.cpp
```diff
@@ -85,10 +122,13 @@ unsigned EffectEcho::GetAudioOutCount() const
...
 {
-   auto& echoSettings = mSettings;
+   // temporary - in the final step this will be replaced by
+   // auto& echoSettings = GetSettings(settings);
+   //
+   auto& echoSettings = static_cast<const EffectEcho&>(mProcessor).mSettings;
```
Paul-Licameli: Ok this is now what I expect for step 3.

src/effects/Echo.cpp
```diff
@@ -108,23 +148,25 @@ bool EffectEcho::ProcessInitialize(
...
 
-size_t EffectEcho::ProcessBlock(EffectSettings &,
+size_t EffectEcho::Instance::ProcessBlock(EffectSettings& settings,
    const float *const *inBlock, float *const *outBlock, size_t blockLen)
 {
-   auto& echoSettings = mSettings;
+   // temporary - in the final step this will be replaced by
+   // auto& echoSettings = GetSettings(settings);
+   //
+   auto& echoSettings = static_cast<const EffectEcho&>(mProcessor).mSettings;
```
Paul-Licameli: Ok just that the argument name settings isn’t needed at this step

src/effects/Echo.h
```diff
@@ -30,12 +30,9 @@ struct EffectEchoSettings
    double decay{ decayDefault };
 };
 
-class EffectEcho final : public EffectWithSettings<EffectEchoSettings, StatefulPerTrackEffect>
+class EffectEcho final : public EffectWithSettings<EffectEchoSettings, PerTrackEffect>
```
Paul-Licameli, line 33: The base class should change only at this step.

EffectWithSettingsBase won’t be used, so you can interactively rebase, removing the commit from me.

This step causes a MakeSettings member to be defined and called.  You must understand that it is not correct to use GetSettings until this point, because the generated MakeSettings should not be used either before this.  Only when the mSettings member of the effect is removed, must these other changes be made too, to keep all steps of the transformation correct.

Paul-Licameli:
> I could reproduce the bug I noticed with the two user presets with DTMFgen too, so I conclude the cause is in the framework. I opened issue #2822 for that.

Are you sure the problem is in the framework?  Could it be a bug that already existed in 3.1?

DavidBailes:
> Are you sure the problem is in the framework? Could it be a bug that already existed in 3.1?

The bug isn't present in 3.1.3

pietro68:
Oops I was wrong when I said step 3 can not inherit StatefulPerTrackEffect - it can!
I will now read your other comments and act on them.

src/effects/Echo.cpp
pietro68: ok, done

src/effects/Echo.cpp
pietro68: I commented it out + put a TODO comment "uncomment at final step"

src/effects/Echo.cpp
pietro68: I commented it out + put a TODO comment "uncomment at final step"

src/effects/Echo.cpp
pietro68: ok, done

src/effects/Echo.cpp
```diff
@@ -42,6 +42,43 @@ const ComponentInterfaceSymbol EffectEcho::Symbol
...
+
+   Floats history;
+   size_t histPos;
+   size_t histLen;
+};
+
+
+
+std::shared_ptr<EffectInstance>
+EffectEcho::MakeInstance(EffectSettings& settings) const
```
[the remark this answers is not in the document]
pietro68: done

src/effects/Echo.cpp
pietro68: ok, done

src/effects/Echo.cpp
pietro68: ok, removed

Paul-Licameli:
> > Are you sure the problem is in the framework? Could it be a bug that already existed in 3.1?
> 
> The bug isn't present in 3.1.3

Can you @pietro68  or @DavidBailes write up an issue with exact steps please?

src/effects/Echo.h
pietro68: Alright, commit with EffectWithSettingsBase removed.
The other point is understood now.

pietro68:
> > > Are you sure the problem is in the framework? Could it be a bug that already existed in 3.1?
> > 
> > 
> > The bug isn't present in 3.1.3
> 
> Can you @pietro68 or @DavidBailes write up an issue with exact steps please?

https://github.com/audacity/audacity/issues/2822

pietro68:
I think I addressed the last round of requests; all 4 steps tested ok, with the exception of the bug happening at step 4, also happening in the DTMFgenerator - see issue at https://github.com/audacity/audacity/issues/2822

Paul-Licameli (approved):

Paul-Licameli:
At last this sequence looks like a transformation that is correct at every step.  I will merge it because you will be away for a few days.  I hope you can write a revised outline describing the procedure for transforming an effect, which you can repeat with other effects.

pietro68:
Thanks for the merge! I already attached the revised outline in my previous comment here, but I think I will add it to the repository - a simple PR will follow for that.

Issue #2811, opened 2022-04-12
More utilities for transition of effects to statelessness


Paul-Licameli:
Another utility to define GetSettings() but preserve the persistency of effect settings in transitional commits.

Another utility classes to simplify the definition of Instance classes of stateless effects.

PR #2812, opened 2022-04-12
More stateless effect utilities

Paul-Licameli:
Resolves: #2811

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -166,6 +166,13 @@ size_t EffectInstanceWithBlockSize::SetBlockSize(size_t maxBlockSize)
    return (mBlockSize = maxBlockSize);
 }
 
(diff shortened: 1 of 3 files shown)


Paul-Licameli:
> I approve the code, but as you will see from my comment [here](https://github.com/audacity/audacity/pull/2801#issuecomment-1097752082) I do not really need the new EffectWithSettingsBase in order to achieve preservation of settings at every step. What shall we do then with this PR?

You could still use EffectInstanceWithSampleRate to simplify the instance class.  Just inherit that, and omit the declarations of mSampleRate and SetSampleRate().

Is the rest of the pull request really correct?  I'm not convinced yet.

Paul-Licameli:
Please approve this pull request.  I will merge only the first commit.  I use EffectInstanceWithSampleRate to rewrite Ladspa effects but I don’t use the other commit.

pietro68 (approved):
Changes are fine with me.

Issue #2816, opened 2022-04-13
Split New fails with multiple partial clip selections


petersampsonaudacity:
**Describe the bug**
**Edit >Audio Clips > Split New** fails with multiple partial clip selections.

This is a regression on 3.0.5 ad all earlier introduced in 3.1.0

While this is easy to see in my simple chirp example, in real like with real audio iy can be a lot harder to see what's going on and thus may result in a loss of data for unsuspecting and unobservant users.

@LWinterberg hence I believe this warrants a P2 rating

**To Reproduce**
1. Generate a 30 second clip
2. Split the clip at 10 and 15 seconds => 3 clips
3. select the middle clip - the 10-15 seconds
4. delete that clip
5. Select from 5-20 seconds - _see screenshot 1_
6. Edit >Audio Clips > Split New
7. **Observe: the 5-10 seconds is the right length but looks to be the 15-20 seconds portion - and the second portion is a flat line filling the space** _see screenshot 2_
See Screenshot 3 for expected result

It also fails if the two clips are touching - see screenshot 4

**Actual behavior**
Tell what actually happened after taking the steps outlined in the previous section.

**Expected behavior**

**Screenshots**
### 1) partial selection of two separated clips
![image](https://user-images.githubusercontent.com/50205984/163197132-6e6fd5c0-c182-457a-8003-ad210d8052c5.png)

### 2) Step 7 - after the Split New we get a corrupted split
![image](https://user-images.githubusercontent.com/50205984/163198071-3aa3551b-1e39-4537-a507-ab1fe335fbc4.png)

### 3) Expected result (achieved with 3.0.5)
![image](https://user-images.githubusercontent.com/50205984/163199783-52285993-185c-4792-b795-13e7ea503cdc.png)

### 4) It also fails if the two clips are touching
![image](https://user-images.githubusercontent.com/50205984/163200438-75422394-c1ed-43be-8870-5c143287b95f.png)

**Additional information (please complete the following information):**
 - OS: Windows 10 - but assume all OS
 - Version [Audacity  3.1.0 - 3.2.0 alpha including current 3.1.3

**Additional context**
**Edit >Audio Clips > Split New**  works fine if I
a) just select from 5-10 seconds, or
b) just select from 15-20 seconds

petersampsonaudacity:
It works properly if I select the whole of the second clip:
![image](https://user-images.githubusercontent.com/50205984/163201448-9b5644df-cb97-4df6-b1fe-1748bdb237f2.png)


[...]

PR #2820, opened 2022-04-13
Fix issue #2816: Split New fails with multiple partial clip selections

Connorbub:
Resolves: [https://github.com/audacity/audacity/issues/2816](https://github.com/audacity/audacity/issues/2816)

This changes the way the offset for the new clip is calculated when splitting to a new track. This fixes the issue with clips not being copied, but does not add a silent placeholder track.

Recommended:


Connorbub:
disregard, set up the PR improperly

PR #2821, opened 2022-04-13
Fix issue #2816: Partial clips are not copied correctly during Split New

Connorbub:
Resolves: [https://github.com/audacity/audacity/issues/2816](https://github.com/audacity/audacity/issues/2816)

Changes: Recalculates the offset for each clip during a copy.

Motivation: Ensures that no data is lost in the Split New process; also ensures that the new clips match the original clips during a copy.

Recommended:


Connorbub:
This is mostly rolled into https://github.com/audacity/audacity/pull/2847, so there's really no reason to keep this older PR anymore.

Issue #2683, opened 2022-03-15
Add a linear-scale dB vertical ruler


LWinterberg:
Currently, there is an option between linear scale (from 1 to -1) and a fairly well-hidden logarithmic scale (from -0dB to -60dB by default). This causes confusion for beginners as they look at the waveform and peak at 0.5, thinking they're way too low - while in reality that's equivalent to -6dB, so already bordering the "red" area. 

Using the logarithmic dB scale by default comes with a set of problems on its own (see https://github.com/audacity/audacity/issues/2399#issuecomment-1009916140), so using it as default is not an option. 

However, using a linear dB scale is a well-established idea, using Audition as example: 
![image](https://user-images.githubusercontent.com/87814144/158396992-59a9c024-bc5f-4289-81b3-2edefa812680.png)

relevant files:
https://github.com/audacity/audacity/blob/master/src/widgets/Ruler.cpp
https://github.com/audacity/audacity/tree/master/src/tracks/playabletrack/wavetrack/ui

SteveDaulton:
Yes please :)

micpap25:
I'll work on resolving this issue today.

micpap25:
I began working a little bit ago; this might take longer than expected given that the codebase is built on having exactly two types of scales for the Waveform Ruler.

micpap25:
To add a linear db scale, I'm at least going to have to rename variables in the vast majority of the waveform-drawing functions throughout the code, so I'm going to wait for a response before I go ahead.

The code consistently makes the distinction as "This is either linear or decibel" and so I also don't want to move forward until I fully understand which behaviors a linear db scale should borrow from linear integer, and which it should borrow from logarithmic db.

LWinterberg:
The waveform is identical to a linear waveform (ie Audacity's default, going from 1 to -1). The only thing that changes is the ruler on the side, which instead of eg {1, 0.5, 0, 0.5, 1} would read {0, -6, -inf, -6, 0}

micpap25:
I appreciate the advice in regards to focusing on the Ruler class. I will try to minimize my changes to other classes; I still will have to add the option to select the linear decibel ruler, which is found in the Waveform parts of the code.

I've been busy last night and this morning; I'll do my best to complete this by Friday night.

micpap25:
Status update: I've finished all the code for selecting the Linear (dB) ruler. I just need to create the ruler now. 
[...]

PR #2828, opened 2022-04-16
Implement Linear DB Scale

micpap25:
Resolves: #2683

Creates a third scale for the waveform ruler, which measures in decibels referencing the linear scale. This scale can be selected in either the context menu or preferences menu. By doing this, users will be able to view their waveforms in decibels without distorting them through the logarithmic dB scale. 

Recommended:

### DIFF
--- a/src/commands/SetTrackInfoCommand.cpp
+++ b/src/commands/SetTrackInfoCommand.cpp
@@ -287,7 +287,8 @@ static const EnumValueSymbol kColourStrings[nColours] =
 enum kScaleTypes
 {
    kLinear,
-   kDb,
+   kLogarithmic,
+   kLinearDb,
    nScaleTypes
 };
 
@@ -296,7 +297,8 @@ static const EnumValueSymbol kScaleTypeStrings[nScaleTypes] =
    // These are acceptable dual purpose internal/visible names
    { XO("Linear") },
    /* i18n-hint: abbreviates decibels */
-   { XO("dB") },
+   { XO("Logarithmic (dB)") },
+   { XO("Linear (dB)")}
 };
 
 enum kZoomTypes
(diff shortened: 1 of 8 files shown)


petersampsonaudacity:
Tested on W10- with branch build https://github.com/audacity/audacity/actions/runs/2176996124

This branch build looks good to me for this new view:

image

image

micpap25:
@SteveDaulton @petersampsonaudacity Would you say this can be merged in its current state and I can come back to it in the future to implement a "2.0" version of it which marks integer decibel values at the default view?

petersampsonaudacity:
> @SteveDaulton @petersampsonaudacity Would you say this can be merged in its current state and I can come back to it in the future to implement a "2.0" version of it which marks integer decibel values?

Not my call any more - but my 2c worth is that it looks most odd with the fractional decimal numbering.  But you don't see that at normal zoom levels.

SteveDaulton:
> @SteveDaulton @petersampsonaudacity Would you say this can be merged in its current state and I can come back to it in the future to implement a "2.0" version of it which marks integer decibel values?

Not my call either, but I think that the tick mark positions is an essential feature for dB view. There's a risk that inexperienced users may guess that the ticks represent 1 dB intervals or  some such, whereas for experienced users it just looks wrong.
I do appreciate that this is a difficult feature to implement having looked into it myself some years ago (before being distracted by a more pressing issue). I was rather surprised to see it marked as a good first issue.

micpap25:
I think my current solution is the best I can do with the current state of the Ruler class. I believe the custom and truly nonlinear ruler will have to be implemented for a Linear DB ruler that doesn't rely on current Linear infrastructure.

micpap25:
Hi; what's the status on this PR? I'm not entirely sure what the process for merging is so I just want to make sure everything is OK.

LWinterberg:
Everything's okay, the devs are just busy with the next release :) 

In the meantime, I'd encourage you to look into the "2.0" version of it :)

Paul-Licameli:
This is not the correct PR with Michael’s latest fixes

Paul-Licameli:
Michael's active branch is at https://github.com/micpap25/audacity/tree/post-gsoc-fixes-1

Paul-Licameli:
#4393 supersedes this pull request

Issue #2659, opened 2022-03-10
Group Effects by what they do


LWinterberg:
Proposed grouping

# For effects

-   Add Realtime Effect (#2661)
-   Add/Remove Plugins
-   Repeat last effect 
-   ~~Often used effects (#2660)~~
*  -------
-   Volume and compression
    -   Amplify
    -   Compressor
    -   Limiter
    -   Normalize
    -   Loudness Normalization
    -   Auto Duck
-   Fading 
    -   Fade in
    -   Fade out
    -   Studio Fade out
    -   Adjustable fade
    -   Crossfade Clips
    -   Crossfade Tracks
-   Pitch and Tempo
    -   Change Pitch
    -   Change Speed
    -   Change Tempo
    -   Paulstretch
    -   Sliding Stretch
    -   Reverse
-   EQ and Filters
    -   Bass and treble
    -   Graphic EQ
    -   Filter curve EQ
    -   High-pass filter
    -   Low-pass filter
    -   Notch Filter
-   Noise Removal and Repair
    -   Click removal
    -   noise reduction
    -   noise gate
    -   repair
    -   Clip fix
-   Delay and Reverb
    -   Echo
    -   Reverb
    -   Delay
-   Distortion and Modulation
    -   Tremolo
    -   Distortion
    -   Wahwah
    -   Phaser
    -   Vocoder
-   Special 
    -   Repeat
    -   Invert [RENAME effect to INVERT POLARITY]
    -   Truncate Silence 
    -   Vocal reduction and isolation
    -   vocal remover
-   Spectral tools
    -   Spectral Delete
    -   Spectral edit multi tool
    -   Spectral edit parametric EQ
    -   Spectral edit shelves
*   -----
-   [any external plugins, whether it be nyquist, VST or LV2, arranged as thus:]
-   Author/Publisher
    -   Effect, sorted alphabetically by name

(sidenotes: the first 3 items in special and the entirety of "fades" really should be editing tools, not effects. Also, truncate silence is very similar to the "detach at silence" in the edit menu)

# for Generators and Analyzers

* Add/Remove Plugins
* ~~[Often used Generators/Analyzers - only if generate/analyze-type plugins are active]~~
* ----
* Audacity's default generators/analyzers, alphabetically
* ----
* [any external plugins, arranged as thus:
* Author/Publisher
    * Generator/Analyzer, sorted alphabetically

petersampsonaudacity:
[...]

PR #2831, opened 2022-04-18
Issue 2659: effects menu sorting

vsverchinsky:
Resolves: #2659 

Recommended:

### DIFF
--- a/locale/res.its
+++ b/locale/res.its
@@ -1,0 +1,5 @@
+<?xml version="1.0"?>
+<its:rules xmlns:its="http://www.w3.org/2005/11/its" version="1.0">
+    <!-- translate textual content of all elements -->
+    <its:translateRule selector="//*" translate="yes"/>
+</its:rules>
--- a/locale/update_po_files.sh
+++ b/locale/update_po_files.sh
@@ -33,6 +33,23 @@ xargs xgettext \
 --package-version='3.0.3' \
 --msgid-bugs-address="audacity-translation@lists.sourceforge.net" \
 --add-location=file -L Lisp -j -o audacity.pot 
+echo ";; Adding resource files to audacity.pot"
+for path in ../res ; do 
+   find $path -name \*.xml 
+done | \
+sed -E 's/\.\.\///g' |\
+xargs xgettext \
+--its=res.its \
+--no-wrap \
+--default-domain=audacity \
+--directory=.. \
+--add-location=file  \
+--copyright-holder='Audacity Team' \
+--package-name="audacity" \
+--package-version='3.0.3' \
+--msgid-bugs-address="audacity-translation@lists.sourceforge.net" \
+-j -o audacity.pot 
+
 if test "${AUDACITY_ONLY_POT:-}" = 'y'; then
     return 0
 fi
--- a/res/effects_menu_defaults.xml
+++ b/res/effects_menu_defaults.xml
@@ -1,0 +1,96 @@
+<EffectMenuList>
+    <Group>
+        <!-- Effects menu group name; audio dynamics compression, not data compression -->
+        <Name>Volume and Compression</Name>
+        <Effects>
+            <Effect>Amplify</Effect>
+            <Effect>Compressor</Effect>
+            <Effect>Limiter</Effect>
+            <Effect>Normalize</Effect>
+            <Effect>Loudness Normalization</Effect>
+            <Effect>Auto Duck</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <!-- Effects menu group name -->
+        <Name>Fading</Name>
+        <Effects>
+            <Effect>Fade In</Effect>
+            <Effect>Fade Out</Effect>
+            <Effect>Studio Fade Out</Effect>
+            <Effect>Adjustable Fade</Effect>
+            <Effect>Crossfade Clips</Effect>
+            <Effect>Crossfade Tracks</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>Pitch and Tempo</Name>
+        <Effects>
+            <Effect>Change Pitch</Effect>
+            <Effect>Change Speed</Effect>
+            <Effect>Change Tempo</Effect>
+            <Effect>Paulstretch</Effect>
+            <Effect>Sliding Stretch</Effect>
+            <Effect>Reverse</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>EQ and Filters</Name>
+        <Effects>
+            <Effect>Bass and Treble</Effect>
+            <Effect>Graphic EQ</Effect>
+            <Effect>Filter Curve EQ</Effect>
+            <Effect>High-Pass Filter</Effect>
+            <Effect>Low-Pass Filter</Effect>
+            <Effect>Notch Filter</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>Noise Removal and Repair</Name>
+        <Effects>
+            <Effect>Click Removal</Effect>
+            <Effect>Noise Reduction</Effect>
+            <Effect>Noise Gate</Effect>
+            <Effect>Repair</Effect>
+            <Effect>Clip Fix</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>Delay and Reverb</Name>
+        <Effects>
+            <Effect>Echo</Effect>
+            <Effect>Reverb</Effect>
+            <Effect>Delay</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>Distortion and Modulation</Name>
+        <Effects>
+            <Effect>Tremolo</Effect>
+            <Effect>Distortion</Effect>
+            <Effect>Wahwah</Effect>
+            <Effect>Phaser</Effect>
+            <Effect>Vocoder</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <!-- Effects menu group name -->
+        <Name>Special</Name>
+        <Effects>
+            <Effect>Repeat</Effect>
+            <Effect>Invert</Effect>
+            <Effect>Truncate Silence</Effect>
+            <Effect>Vocal Reduction and Isolation</Effect>
+            <Effect>Vocal Remover</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>Spectral Tools</Name>
+        <Effects>
+            <Effect>Spectral Delete</Effect>
+            <Effect>Spectral edit multi tool</Effect>
+            <Effect>Spectral edit parametric EQ</Effect>
+            <Effect>Spectral edit shelves</Effect>
+        </Effects>
+    </Group>
+</EffectMenuList>
--- a/src/menus/PluginMenus.cpp
+++ b/src/menus/PluginMenus.cpp
@@ -32,6 +32,213 @@
 // private helper classes and functions
 namespace {
 
+enum class GroupBy
+{
+   Publisher,
+   Type
+};
+
+enum class SortBy
+{
+   Name,
+   PublisherName,
+   TypeName
+};
+   
+// Some weird special case stuff just for Noise Reduction so that there is
+// more informative help
+CommandFlag FixBatchFlags(CommandFlag batchflags, const PluginDescriptor* plug)
+{
+   if ( plug->GetSymbol().Msgid() == XO( "Noise Reduction" ) )
+      return ( batchflags | NoiseReductionTimeSelectedFlag() ) & ~TimeSelectedFlag();
+   return batchflags;
+}
+
+   
+bool IsEnabledPlugin(const PluginDescriptor* plug)
+{
+   if( PluginManager::Get().IsPluginLoaded(plug->GetID()) && EffectManager::Get().IsHidden(plug->GetID()) )
+         return false;
+   if ( !plug->IsEnabled() ){
+      return false;// don't add to menus!
+   }
+   return true;
+}
+
+bool IsDefaultPlugin(const PluginDescriptor* plug)
+{
+   if (plug->IsEffectDefault()
+#ifdef EXPERIMENTAL_DA
+      // Move Nyquist prompt into nyquist group.
+      && (plug->GetSymbol() !=
+            ComponentInterfaceSymbol("Nyquist Effects Prompt"))
+      && (plug->GetSymbol() != ComponentInterfaceSymbol("Nyquist Tools Prompt"))
+      && (plug->GetSymbol() != ComponentInterfaceSymbol(NYQUIST_PROMPT_ID))
+#endif
+      )
+      return true;
+   return false;
+}
+
+
+// Forward-declared function has its definition below with OnEffect in view
+void AddEffectMenuItemGroup(
+   MenuTable::BaseItemPtrs &table,
+   const TranslatableStrings & names,
+   const PluginIDs & plugs,
+   const std::vector<CommandFlag> & flags,
+   bool useSubgroups);
+
+   
+void AddGroupedEffectMenuItems(
+   MenuTable::BaseItemPtrs &table,
+   std::vector<const PluginDescriptor*> & plugs,
+   CommandFlag batchflags,
+   CommandFlag realflags,
+   GroupBy groupBy,
+   bool useSubgroups)
+{
+   TranslatableString last;
+   TranslatableString current;
+
+   size_t pluginCnt = plugs.size();
+
+   TranslatableStrings groupNames;
+   PluginIDs groupPlugs;
+   std::vector<CommandFlag> groupFlags;
+
+   for (size_t i = 0; i < pluginCnt; i++)
+   {
+      const PluginDescriptor *plug = plugs[i];
+
+      auto name = plug->GetSymbol().Msgid();
+
+      if (plug->IsEffectInteractive())
+         name += XO("...");
+      
+      if (groupBy == GroupBy::Publisher/*wxT("groupby:publisher")*/)
+      {
+         current = EffectManager::Get().GetVendorName(plug->GetID());
+         if (current.empty())
+         {
+            current = XO("Unknown");
+         }
+      }
+      else if (groupBy == GroupBy::Type /*wxT("groupby:type")*/)
+      {
+         current = EffectManager::Get().GetEffectFamilyName(plug->GetID());
+         if (current.empty())
+         {
+            current = XO("Unknown");
+         }
+      }
+
+      if (current != last)
+      {
+         using namespace MenuTable;
+         BaseItemPtrs temp;
+         bool bInSubmenu = !last.empty() && (groupNames.size() > 1);
+
+         AddEffectMenuItemGroup(temp,
+            groupNames,
+            groupPlugs, groupFlags, useSubgroups);
+
+         table.push_back( MenuOrItems( wxEmptyString,
+            ( bInSubmenu ? last : TranslatableString{} ), std::move( temp )
+         ) );
+
+         groupNames.clear();
+         groupPlugs.clear();
+         groupFlags.clear();
+         last = current;
+      }
+
+      groupNames.push_back( name );
+      groupPlugs.push_back(plug->GetID());
+      groupFlags.push_back(
+         plug->IsEffectRealtime() ? realflags : FixBatchFlags( batchflags, plug ) );
+   }
+
+   if (groupNames.size() > 0)
+   {
+      using namespace MenuTable;
+      BaseItemPtrs temp;
+      bool bInSubmenu = groupNames.size() > 1;
+
+      AddEffectMenuItemGroup(temp,
+         groupNames, groupPlugs, groupFlags, useSubgroups);
+
+      table.push_back( MenuOrItems( wxEmptyString,
+         ( bInSubmenu ? current : TranslatableString{} ), std::move( temp )
+      ) );
+   }
+}
+
+void AddSortedEffectMenuItems(
+   MenuTable::BaseItemPtrs &table,
+   std::vector<const PluginDescriptor*> & plugs,
+   CommandFlag batchflags,
+   CommandFlag realflags,
+   SortBy sortBy,
+   bool useSubgroups)
+{
+   size_t pluginCnt = plugs.size();
+
+   TranslatableStrings groupNames;
+   PluginIDs groupPlugs;
+   std::vector<CommandFlag> groupFlags;
+
+   for (size_t i = 0; i < pluginCnt; i++)
+   {
+      const PluginDescriptor *plug = plugs[i];
+
+      auto name = plug->GetSymbol().Msgid();
+
+      if (plug->IsEffectInteractive())
+         name += XO("...");
+
+      TranslatableString group;
+      if (sortBy == SortBy::PublisherName/* wxT("sortby:publisher:name")*/)
+      {
+         group = EffectManager::Get().GetVendorName(plug->GetID());
+      }
+      else if (sortBy == SortBy::TypeName /*wxT("sortby:type:name")*/)
+      {
+         group = EffectManager::Get().GetEffectFamilyName(plug->GetID());
+      }
+
+      if (plug->IsEffectDefault())
+      {
+         group = {};
+      }
+
+      groupNames.push_back(
+         group.empty()
+            ? name
+            : XO("%s: %s").Format( group, name )
+      );
+
+      groupPlugs.push_back(plug->GetID());
+      groupFlags.push_back(
+         plug->IsEffectRealtime() ? realflags : FixBatchFlags( batchflags, plug ) );
+   }
+
+   if (groupNames.size() > 0)
+   {
+      AddEffectMenuItemGroup(
+         table, groupNames, groupPlugs, groupFlags, useSubgroups);
+   }
+}
+
+struct MenuSectionBuilder
+{
+   std::vector<const PluginDescriptor*> plugins;
+
+   std::function<bool(const PluginDescriptor*)> filter;
+   std::function<bool(const PluginDescriptor*, const PluginDescriptor*)> compare;
+   std::function<void(MenuTable::BaseItemPtrs&, std::vector<const PluginDescriptor*>&)> add;
+};
+
 AttachedWindows::RegisteredFactory sMacrosWindowKey{
    []( AudacityProject &parent ) -> wxWeakRef< wxWindow > {
       auto &window = ProjectWindow::Get( parent );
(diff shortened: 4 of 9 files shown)


src/menus/PluginMenus.cpp
```diff
@@ -32,6 +32,107 @@
...
+using MenuList = std::vector<std::pair<TranslatableString, std::vector<TranslatableString>>>;
+
+const MenuList& PreferredEffectsMenu()
+{
+   static MenuList items =
+   {
+      {
+         XO("Volume and compression"),
+         {
+            XO("Amplify"),
```
Paul-Licameli, line 44: I don't like the closed-endedness -- the violation of the "O", open-closed principle, of SOLID:  "A design should be open to extension but closed to modification"

In other words, if we ever add a new effect that falls into one of these categories, then it won't be self-contained in one new implementation file, but instead this file will need modification to make the menu sorting continue to work as desired.

And we are also depending on long-distance coincidences of string literal values, between this file and each of the many effect implementation files, which is хрупкость.

So I suggest:

- A new enumeration EffectCategory, defined in Effect.h
- A table mapping EffectCategory to EnumValueSymbol (which pairs an identifier and TranslatableString), defined in Effect.cpp and accessible by a static member function
- MenuList is instead `std::vector<std::pair<EffectCategory, std::vector<TranslatableString>>>` and gets TranslatableString from that table
- A virtual function of Effect called GetEffectCategory, appropriately overridden in all the effects that matter
- Some of these effects are not built-ins, but Nyquist plug-ins. So NyquistEffect needs an override of GetEffectCategory that doesn't just return a constant, but computes an answer too.
  - Add a case in NyquistEffect::Parse to recognize a new special Lisp comment `$category`, which does reverse-lookup of the table of categories, by the Identifier (not by the TranslatableString), to find the enum value
  - Modify the plug-ins/*.ny files as needed to identify their categories
  - Ping @SteveDaulton when you do this, to review the changes at least in Nyquist.cpp, and recommend how to update documentation of Nyquist
- PreferredEffectsMenu does not return a static value, but computes its answer instead, by iterating the plug-ins and querying GetEffectCategory
- Maybe it can cache that computed answer, but invalidation of the cache would need to be done by the plugin manager dialog, or if the user changes the Language preference; see class PreferenceListener

Do all of this, and I will be less dissatisfied... but then the set of categories and the sorting of the categories themselves will still be closed-ended, and in a further future I might want to seek a fix for that too involving a registration of categories.

Paul-Licameli (changes requested):
Closed-endedness and magic string literals are problems.  See previous comment.

src/menus/PluginMenus.cpp
vsverchinsky: 1. EffectCategory is very likely to be extended/changed.
2. Not each effect has should have effect category assigned, effects can be added or removed from that list.
3. MenuList can be changed at any time, without modification of the effects source code, it's a resource object.
Considering that, I don't agree that my implementation violates "O"

src/menus/PluginMenus.cpp
```diff
@@ -32,6 +32,383 @@
...
+            XO("Wahwah"),
+            XO("Phaser"),
+            XO("Vocoder")
+         }
+      },
+      {
+         XO("Special"),
+         {
+            XO("Repeat"),
+            XO("Invert"),
```
LWinterberg, line 117: rename the `Invert` effect to `Invert Polarity`, please.

Paul-Licameli: Why?

src/menus/PluginMenus.cpp
```diff
@@ -32,6 +32,383 @@
...
 namespace {
 
+using MenuList = std::vector<std::pair<TranslatableString, std::vector<TranslatableString>>>;
+
+const MenuList& PreferredEffectsMenu()
+{
+   static MenuList items =
+   {
+      {
+         XO("Volume and compression"),
```
LWinterberg, line 42: Ah, sorry, you copied my typo here. It should be "Volume and Compression" with a big C

src/menus/PluginMenus.cpp
```diff
@@ -32,6 +32,383 @@
...
+            XO("Vocal Remover")
+         }
+      },
+      {
+         XO("Spectral tools"),
+         {
+            XO("Spectral Delete"),
+            XO("Spectral edit multi tool"),
+            XO("Spectral edit parametric EQ"),
+            XO("Spectral edit shelves")
```
LWinterberg, line 129: Same with these guys: 
* Spectral Tools
* Spectral Delete
* Spectral Edit Multi Tool
* Spectral Edit Parametric EQ
* Spectral Edit Shelves

src/menus/PluginMenus.cpp
Paul-Licameli: If it is agreed to make that name change, the correct change in Invert.cpp is:

```
const ComponentInterfaceSymbol EffectInvert::Symbol
{ L"Invert", XO("Invert Polarity") };
```

because in case anyone uses the Invert effect in a macro, this will avoid breaking that macro when they migrate to a newer version of Audacity.  Only the localized name of the effect will be changed, not the English identifier that macro programmers use.

src/menus/PluginMenus.cpp
vsverchinsky: @LWinterberg as Paul said in his comment above these are "identifiers", not names, but "Spectral Tools" is a group name and I'll fix that

Paul-Licameli: I VETO this change request as part of the same issue.

Make this change request a different, minor issue.

Changing the name while preserving macro compatibility turns out not to be as easy for Nyquist effects as it would be for Invert.  It needs more thought.

src/menus/PluginMenus.cpp
Paul-Licameli: Your implementation violates open ended ness for reasons I described.  You must change PluginMenus.cpp if in future a new categorized effect is added, rather than having all necessary new code self contained in the effect’s implementation.  I do not accept this.

Paul-Licameli: > 1. EffectCategory is very likely to be extended/changed.

How does that make me wrong?

It is something I admitted: that even if you do my steps, it is not a complete fix for the problem of closed ended ness.  Still the steps I described are necessary if not sufficient.

> 2. Not each effect should have effect category assigned, effects can be added or removed from that list.

Do you mean there should be another enumeration value for effects of undefined category?  That is easy enough.

“Effects can be added or removed from the list.”  Do you mean as future development, or do you mean that the composition of the lists can be changed at runtime?

> 3. MenuList can be changed at any time, without modification of the effects source code, it's a resource object.
> 

Do you mean that as a future development, we change our decision about the categorization of an individual effect?  But it is much more proper, if some source code must change, that title is the source code of that effect only rather than PluginMenus.cpp.

> Considering that, I don't agree that my implementation violates "O". Ideally, there could be an external file describing the stucture of that menu, which can be changed without any code modification.

Paul-Licameli: "MenuList can be changed at any time, without modification of the effects source code, it's a resource object.
Does “resource object” mean that the categorization of individual effects should not be hard oder in them, but somehow specified in an external file, like the configuration, and the user would have some way to change its contents, and the program would do effect sorting according to an interpretation of that file? But this implementation does none of that."

So I do t understand what you mean by “resource object.”

Paul-Licameli: Sorry I edited your original comment by mistake.  I didn't mean to delete it.  But I think I quoted all of it in my replies.

Paul-Licameli: "Maybe it can cache that computed answer, but invalidation of the cache would need to be done by the plugin manager dialog, or if the user changes the Language preference; see class PreferenceListener"

To clarify I mention this really to suggest you don't bother with it:  Do recalculate the list each time you need to repopulate the effects menus.

Paul-Licameli: > > How does that make me wrong?
> 
> "A design should be open to extension but closed to modification"

That doesn't answer my question.

> 
> > it is not a complete fix for the problem of closed ended ness
> 
> Please, let do not pollute interfaces with more constants that has some meaning for about 30 effects, it has nothing to do with OC.

What "interface" is polluted?  Effect itself?  Well it's going through lots of necessary changes this release.  But it is quite simple to make effects un-categorized with a default implementation of the virtual function, and so most effects will need no modification for this.  Only the few that are categorized.

> 
> > Do you mean there should be another enumeration value for effects of undefined category?
> 
> Why should prebundled effect, which isn't present in that menu, should be categorized as "Undefined"? Or why should any other external effect have category which will never be used?

This really isn't different in principle than IsHiddenFromMenus.

> 
> > Do you mean as future development, or do you mean that the composition of the lists can be changed at runtime?
> 
> During development ofcourse, and this is in fact points to the possibility to violate OC in future.

We are failing to understand "OC" to mean the same thing.

If there is a new effect, its categorization should be determined by the code of that effect alone.

The file that makes effect menus should be generalized, and should not make special mentions of any individual effects.  The interface offered by abstract Effect is extended now to be sufficient for the needs of the plugin menus.

Do you see there is special weird stuff in PluginMenus about noise reduction and the Nyquist prompt?  True.  But that is bad legacy, not good precedent.

> 
> > Do you mean that as a future development, we change our decision about the categorization of an individual effect? But it is much more proper, if some source code must change, that is the source code of that effect only rather than PluginMenus.cpp.
> 
> Yes, we can add or remove effects to/from that list, even now there are prebundled effects that are not present in that list, the set of groups can change as well. Question: there are groups like "Distortion and Modulation", "EQ and Filters", should there be 4 emum values or only 2? What if effect provides both Distortion and EQ? It's not a problem from the implementation point of view, and it's not a bad idea in general, but there is no use for it right now.

As I said, I am admitting to incompleteness of this solution for open endedness, but I do not accept your argument that therefore, let this PR stand and make no attempt to fix it.

Effects should have some virtual function reporting their category as some kind of value, and PluginMenus.cpp should use that value and not have a hard-coded list of effect names.

Start by making that value an enumeration.  Though this enumeration of categories is closed ended, the set of effects in any one category is not closed ended.

Let me see the redraft.  Then we can discuss the further step of changing the value type from enumeration to some other dynamically determined set of cateogires.

> 
> > Does “resource object” mean that the categorization of individual effects should not be hard oder in them, but somehow specified in an external file, like the configuration, and the user would have some way to change its contents, and the program would do effect sorting according to an interpretation of that file? But this implementation does none of that.
> 
> There is a set of effects that should be put in a certaing menu groups, there is no need to "categorize" all effects. MenuLists can be put into separate cpp file (like AllThemeResources), or even external xml (json, txt,...) file. Not sure that we need to provide a way for user to change that list, but surely we can do that. In fact we can extend format with rules and filters and provide a way to make this menus very customizable, and your idea with effects category will make it more flexible for the user.

vsverchinsky: >That doesn't answer my question.

What question then? Using enum for entity like effect category is very likely to violate OC (until we agree that it will never change). So I don't see why your impementation violating OC is preferrable over my implementation which "violates OC" (which is not, because you don't have to change effect to change menu ordering, and you don't have to change anything in the menu to add a new effect, though you may want to change that too, but the problem here is that you cannot answer to the question "how should I change that menu?")

>If there is a new effect, its categorization should be determined by the code of that effect alone.

Effect placement in the menu should be determined by the menu. It can use properties of the effect to do that and using effect ID for that purpose isn't worse than using any other property of the effect. 

>The file that makes effect menus should be generalized, and should not make special mentions of any individual effects.

Effect ID and effect category are both properties of the same entity, there is no difference between them in sence that they both could to be used as a filter parameters.

This task does not specify that we should assign category to each effect, and build menu so that all effects grouped by these categories. This task requires to put **certain** effects into certain submenus, that's why there are references to the individual effects in the menu description, you cannot solve such task using generalization.

src/menus/PluginMenus.cpp
SteveDaulton: > Polarity



> rename the `Invert` effect to `Invert Polarity`, please.

As Paul wrote: "why?". What other kinds of waveform inversion are there? Yes there is such a thing as "spectral inversion" (which Audacity does not have), but if that's the distinction that you want to make then I don't think that "invert polarity" does it as well as "invert waveform".

Regardless of which wording is decided, please avoid changing the command name unless absolutely necessary as doing so will break custom macros that use the command. As Paul wrote: "Only the localized name of the effect" should be changed.

src/menus/PluginMenus.cpp
Paul-Licameli: > > That doesn't answer my question.
> 
> What question then? Using enum for entity like effect category is very likely to violate OC (until we agree that it will never change). So I don't see why your impementation violating OC is preferrable over my implementation which "violates OC" (which is not, because you don't have to change effect to change menu ordering, and you don't have to change anything in the menu to add a new effect, though you may want to change that too, but the problem here is that you cannot answer to the question "how should I change that menu?")

One more time:  I admit the incompleteness of my suggestions, but still say it is a better organization than this first attempt.  If you are actually arguing to go beyond my suggestions and figure out open-endedness of categories, not just the members of categories -- great, do so!

If not, still I want some of the extra effort done now.

But the rest of what you say is unpersuasive to me.  Saying you only satisfy the issue in a looks-like-it-works way in the black box view is not acceptable.

The decision to categorize an individual effect one way or another way should be encoded in the implementation of that effect.  That is the open-ended solution in regard to the enlargement of the set of effects.

Ideally effects should be plug-ins, not baked into the executable.  In fact this is realized in the case of Nyquist effects, though we don't do this yet for effects in C++.

At least for Nyquist effects, a Nyquist programmer should have the ability to put a new effect into the sub-menu of their choice, just by writing appropriate code in the .ny file, without needing special cooperation from a re-compiled Audacity executable.

Paul-Licameli:
Vitaly and I are disagreeing about the proposed implementation.  In this first attempt, the tree that Leo specified is essentially replicated in one place in the source code.

I don't find that acceptable.  Instead, I say category information should be associated with each individual effect in the code that defines that effect.

The code that builds the effect menu should iterate all effects and gather their category information in a generalized way according to some new extension of the effect interface, then sort them without special hardcoded information about any one effect.

I have pointed out that cetainly in the case of Nyquist effects, some other Nyquist programmer wanting to extend Audacity should have means to specify in the .ny file how to categorize that effect, without rebuilding Audacity.  I say build into Audacity now the code that interprets a new kind of special Nyquist comment, then use that comment in .ny files.

Paul-Licameli:
I think extra effort is needed for open-endedness.  If you really do want to go beyond my incomplete suggestion, I would encourage it.  The value returned by the new virtual function of Effect would not be of an enumeration type.  It would be a ComponentInterfaceSymbol pairing an Identifier and a TranslatableString.  Effect.h could have extern declaration of some ComponentInterfaceSymbol constants commonly used, including the ones specified in the issue.

But PluginMenus.cpp would not require the returned values to be from this limited set of choices.

What I described would not provide a way to order the categories extrinsically.  But if @LWinterberg agrees that a rule like major sort by the localization of the category name is acceptable, then that's enough.

src/menus/PluginMenus.cpp
vsverchinsky: >but still say it is a better organization

I say that they are orthogonal to each other.

>looks-like-it-works

What exactly does not work here?

>At least for Nyquist effects, a Nyquist programmer should have the ability to put a new effect into the sub-menu of their choice, just by writing appropriate code in the .ny file, without needing special cooperation from a re-compiled Audacity executable.

It's opposite to what is specified in the task:
>[any external plugins, whether it be nyquist, VST or LV2, arranged as thus:]
Author/Publisher
Effect, sorted alphabetically by name

Paul-Licameli: > > but still say it is a better organization
> 
> I say that they are orthogonal to each other.

I don't understand what is orthogonal to what.

> 
> > looks-like-it-works
> 
> What exactly does not work here?

Nothing.  BUT what I say from point of view of architecture is, it's not enough that it merely looks like it works.  It shoudl be properly designed for extension.

> 
> > At least for Nyquist effects, a Nyquist programmer should have the ability to put a new effect into the sub-menu of their choice, just by writing appropriate code in the .ny file, without needing special cooperation from a re-compiled Audacity executable.
> 
> It's opposite to what is specified in the task:
> 
> > [any external plugins, whether it be nyquist, VST or LV2, arranged as thus:]
> > Author/Publisher
> > Effect, sorted alphabetically by name

Then @LWinterberg should respecify the issue.

We can specify the interface of Nyquist programming as we can't for the other protocols.  We can and should make the effort to give Nyquist programmers the ability to categorize effects as they choose.

Paul-Licameli: I think @SteveDaulton would agree with the last point.

src/menus/PluginMenus.cpp
```diff
@@ -32,6 +32,217 @@
...
+};
+   
+// Some weird special case stuff just for Noise Reduction so that there is
+// more informative help
+CommandFlag FixBatchFlags(CommandFlag batchflags, const PluginDescriptor* plug)
+{
+   if ( plug->GetSymbol().Msgid() == XO( "Noise Reduction" ) )
+         return
+            ( batchflags | NoiseReductionTimeSelectedFlag() ) & ~TimeSelectedFlag();
+      return batchflags;
```
Paul-Licameli, line 150: Indentation?

src/menus/PluginMenus.cpp
```diff
@@ -28,10 +28,105 @@
...
+
+      GroupHandler(std::pair<TranslatableString, std::vector<TranslatableString>>& group) : group(group) { }
+
+      bool HandleXMLTag(const std::string_view& tag, const AttributesList& attrs) override { return true; }
+      void HandleXMLContent(const std::string_view& text) override { textContent = text; }
+      void HandleXMLEndTag(const std::string_view& tag) override
+      {
+         if(tag == "Name")
+         {
+            group.first = TranslatableString { *textContent, { } };
```
Paul-Licameli, line 83: Add a null check on `textContent`

src/menus/PluginMenus.cpp
```diff
@@ -28,10 +28,381 @@
...
+{
+   EffectsMenuGroups result;
+   EffectsMenuGroupsHandler handler(result);
+
+   XMLFileReader reader;
+   reader.Parse(&handler, path);
+   return result;
+}
+
+enum class GroupBy
```
Paul-Licameli, line 130: The first commit, introducing this enum and much else, is a very pleasing refactoring.

locale/update_po_files.sh
```diff
@@ -33,6 +33,24 @@ xargs xgettext \
...
+xargs xgettext \
+--its=res.its \
+--no-wrap \
+--default-domain=audacity \
+--directory=.. \
+--add-comments=" i18n" \
+--add-location=file  \
+--copyright-holder='Audacity Team' \
+--package-name="audacity" \
+--package-version='3.0.3' \
```
Paul-Licameli, line 50: Someone sould make the repeated '3.0.3' in this script an argument or take it from an environment variable and make it more current.  Not for this PR.

All it does it make a comment in the generated .pot file out of date.

locale/res.its
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0"?>
+<its:rules xmlns:its="http://www.w3.org/2005/11/its" version="1.0">
+    <!-- translate textual content of all elements -->
+    <its:translateRule selector="//*" translate="yes"/>
```
Paul-Licameli, line 4: I understand this simple rule extracts all xml content as strings to translate, but only the <Name> strings really need it, as strings not already part of the program.

locale/update_po_files.sh
```diff
@@ -33,6 +33,24 @@ xargs xgettext \
...
+for path in ../res ; do 
+   find $path -name \*.xml 
+done | \
+sed -E 's/\.\.\///g' |\
+xargs xgettext \
+--its=res.its \
+--no-wrap \
+--default-domain=audacity \
+--directory=.. \
+--add-comments=" i18n" \
```
Paul-Licameli, line 46: I tested comment extraction and it works:  add to the .xml file
```
<!-- audio dynamics compression, not data compression -->
```

before the "Volume and Compression" line, ran the script, confirmed that it was extracted.

It seems I don't need "i18n".  The option to xgettext may be unnecessary.

Include at least one example of a comment in the final version.

res/effects_menu_defaults.xml
```diff
@@ -0,0 +1,93 @@
...
+            <Effect>Fade In</Effect>
+            <Effect>Fade Out</Effect>
+            <Effect>Studio Fade Out</Effect>
+            <Effect>Adjustable Fade</Effect>
+            <Effect>Crossfade Clips</Effect>
+            <Effect>Crossfade Tracks</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>Pitch and tempo</Name>
```
Paul-Licameli, line 25: Which capitalization convention should we use?  This is not like "Volume and Compression".  Should it have "Tempo"?

res/effects_menu_defaults.xml
```diff
@@ -0,0 +1,93 @@
...
+            <Effect>Change Pitch</Effect>
+            <Effect>Change Speed</Effect>
+            <Effect>Change Tempo</Effect>
+            <Effect>Paulstretch</Effect>
+            <Effect>Sliding Stretch</Effect>
+            <Effect>Reverse</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>EQ and filters</Name>
```
Paul-Licameli, line 36: caps

res/effects_menu_defaults.xml
```diff
@@ -0,0 +1,93 @@
...
+        <Effects>
+            <Effect>Repeat</Effect>
+            <Effect>Invert</Effect>
+            <Effect>Truncate Silence</Effect>
+            <Effect>Vocal Reduction and Isolation</Effect>
+            <Effect>Vocal Remover</Effect>
+        </Effects>
+    </Group>
+    <Group>
+        <Name>Spectral tools</Name>
```
Paul-Licameli, line 85: caps

res/effects_menu_defaults.xml
```diff
@@ -0,0 +1,93 @@
...
+        <Effects>
+            <Effect>Spectral Delete</Effect>
+            <Effect>Spectral edit multi tool</Effect>
+            <Effect>Spectral edit parametric EQ</Effect>
+            <Effect>Spectral edit shelves</Effect>
```
Paul-Licameli, line 90: Maybe these few effect names must be fixed, here and in the .ny files

src/menus/PluginMenus.cpp
```diff
@@ -28,10 +28,381 @@
...
+};
+
+EffectsMenuGroups LoadEffectsMenuGroups(const wxString& path)
+{
+   EffectsMenuGroups result;
+   EffectsMenuGroupsHandler handler(result);
+
+   XMLFileReader reader;
+   reader.Parse(&handler, path);
+   return result;
```
Paul-Licameli, line 127: This function nicely encapsulates the decisions about default grouping and sorting of effects.  To define some different policy in a later version should change only what happens in here.

Maybe a static variable can hold the results of the parse so that the .xml file need not be read more than once each session.

But didn't we agree to sort effect names in each group by their translations?  Easily added here.

Did we also want to sort the groups by their translations?

Re-sorting of names in the statically duration results might be repeated in case the user changes languages and the menus must be rebuilt.

Paul-Licameli: Maybe we only want to sort each list by translation, but keep the names ordered as in the resource file.  Whatever the decision, it is easy to change it just here.

Paul-Licameli: For me on Mac, the sorting doesn't work.

The file is at
/Users/paullicameli/Github/audacity2/build/bin/Debug/Audacity.app/Contents/res/effects_menu_defaults.xml

The program looks for a file at
/Users/paullicameli/Github/audacity2/build/bin/Debug/Audacity.app/Contents/MacOS/res/effects_menu_defaults.xml

Paul-Licameli: I moved the .xml to the expected place in the tree and was then able to test the changes and verify that switching grouping criteria in Effects preferences does as expected.

Paul-Licameli:
I'm done with this round of review.  There are some things to fix but the basic design is good.

vsverchinsky:
@petersampsonaudacity In the recent PR update I've also attemted to fix the issue metioned in the discrord channel about bundled effects being grouped by publisher, as well as other issues metioned by @Paul-Licameli in this thread earlier

src/menus/PluginMenus.cpp
```diff
@@ -144,231 +523,201 @@ bool CompareEffectsByType(const PluginDescriptor *a, const PluginDescriptor *b)
...
-         }
+         return AddGroupedEffectMenuItems(items, plugins, batchflags, realflags, groupBy, useSubgroups);
+      };
+   };
 
-         groupNames.push_back( name );
-         groupPlugs.push_back(plug->GetID());
-         groupFlags.push_back(
-            plug->IsEffectRealtime() ? realflags : getBatchFlags( plug ) );
+   auto DefaultFilter = [](auto plug) { return IsEnabledPlugin(plug) && IsDefaultPlugin(plug); };
```
Paul-Licameli, line 557: I see the new changes with `DefaultFilter` and how `IsDefaultPlugin` changed.

No objections.

Paul-Licameli (approved):
I approve the changes.

I suggest one more commit to define a sort comparator for TranslatableString in just one place.

I suggest @petersampsonaudacity should open another issue for what he observed in the case that multiple versions of Audacity are installed.

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch build audacity-win-3.2.0-alpha-20220609+831286a-x64

1)  following an upgrade from 3.1.3 to this branch build there is no longer a "Steve Daulton" menu entry in the Analyze menu - but note that the two Steve Daulton plug-ins are duplicated in the Analyze menu:
image


2) following an upgrade from 3.1.3 to this branch build the Tools menu still has a "Steve Daulton" menu entry - and here too the sub-menu items are duplicated:
image

petersampsonaudacity:
> I suggest @petersampsonaudacity should open another issue for what he observed in the case that multiple versions of Audacity are installed.

@Paul-Licameli  -  I could do that but at the moment this is only a developer branch build it is not in master yet.   Normally we only log bugs that are in the master builds.  

But if you really want I can break with "tradition" and log this issue against @vsverchinsky 's branch build, please advise.

vsverchinsky:
>primarily it's an issue for anyone with a single version of Audacity upgrading from an earlier Audacity version to 3.2.0

@petersampsonaudacity it happens even if there is only a single instance of Audacity installed at the same location as the older version?

petersampsonaudacity:
> > primarily it's an issue for anyone with a single version of Audacity upgrading from an earlier Audacity version to 3.2.0
> 
> @petersampsonaudacity it happens even if there is only a single instance of Audacity installed at the same location as the older version?

@vsverchinsky  - so I tested on my wife's PC simulating an upgrade from 3.1.3 to your 3.2.0 branch and all looks good when I do that. For the "normal" upgrading user with a single Audacity version all seems fine.

1. The existing and only version of Audacity on that PC was 3/1/3 installed in the default location C:\Program Files\Audacity
2. I upgraded to 3.1.3 - using the EXE
3. then installed your branch build (ZIP not EXE) 
4. Observe: the menus:

### Effect menu
image

### Analyze menu
image

### Generate menu
image

### Tools Menu
image

Paul-Licameli:
Peter removed our last doubts.  I merged the branch, although I broke my usual rule that branches should be rebased first onto latest master.  Vitaly is taking time off for two weeks and I didn't want this to wait.

PR #2832, opened 2022-04-19
added guide

pietro68:
Only a formal PR to add the guide to make an effect stateless, which mirrors what was done in https://github.com/audacity/audacity/pull/2801

Recommended:


src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
+4 STEPS TO MAKE AN EFFECT STATELESS
```
Paul-Licameli, line 1: If you think adding this memo to the source code tree will be a helpful reference to you and others, why not.  We should aim to delete this again when all effects have been migrated.

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+
+1) "Define EffectEchoSettings"
```
Paul-Licameli, line 3: Of course this guide should say "Let use use EffectEcho as an example" and it might refer to commit ids that made this example transformation.

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+
+   In header file:
+       before the effect definition, define a struct EffectEchoSettings containing the settings.
+
+       Base class remains: StatefulPerTrackEffect
```
Paul-Licameli, line 8: We may transform other effects that are not per-track effects but will be made stateless too

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+        bool EffectEcho::Validator::ValidateUI()
+        {
+           mAccess.ModifySettings
+           (
+              [this](EffectSettings& settings)
+           {
+              // pass back the modified settings to the MessageBuffer
+
+              // TODO uncomment at last step
+              //EffectEcho::GetSettings(settings) = mSettings;
```
Paul-Licameli, line 64: I did say, avoid real use of `GetSettings` until the last step.

Another possibility, which I am pursuing in my transformation of AudioUnit effects (a big recent PR that is still a draft and doesn't yet do the validator), is that I MAY put all the needed calls to GetSettings() in early, BUT I do not yet use EffectWithSettings<> and the definition of GetSettings() that it generates.  Instead, I define GetSettings() to ignore its argument and return mSettings in the effect object.  At the last step of real statelessness, I can delete the definition of GetSettings() but leave the calls.

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+
+
+        std::shared_ptr<EffectInstance>
+        EffectEcho::MakeInstance(EffectSettings&) const
+        {
+           return std::make_shared<Instance>(*this);
+        }
+        
+        
+    Make the 3 Effect::Process*** methods belong to the instance by:
```
Paul-Licameli, line 133: For some effects, this will be more than 3.  See the example of LadspaEffect::Instance in the waiting PR.

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+        auto& echoSettings = static_cast<const EffectEcho&>(mProcessor).mSettings;
+
+    
+
+
+4) "Real Statelessness"
+
+    In header file:
+    
+    - make the effect inherit  EffectWithSettings<EffectEchoSettings, PerTrackEffect>
```
Paul-Licameli, line 151: Or the second template argument may be just Effect for those that are not per-track.  No examples of this exist yet.

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+    
+    - remove the FetchParameters method
+    
+    - remove mSettings member
```
Paul-Licameli, line 155: Also remove the definition of GetSettings() if you took the alternative I described above

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+
+
+    In cpp file:
+    
+    - go to processInitialize, ProcessBlock, ProcessFinalize; remove the downcast and enable the commented out GetSettings
```
Paul-Licameli, line 160: ... And RealtimeInitialize, etc.

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+    
+    - go to Validator::ValidateUI & Validator::UpdateUI, remove the commenting out on the commented lines
+    
+    - add const to arg EffectEchoSettings& in the Validator constructor
```
Paul-Licameli, line 164: Or (as in LadspaEffect::Validator) you might need only access to get the settings to be copied:

```
   Validator(EffectUIClientInterface &effect,
      EffectSettingsAccess &access, double sampleRate, EffectType type)
      : EffectUIValidator{ effect, access }
/* other member initializations */
      // Copy settings
      , mSampleRate{ sampleRate }
      , mType{ type }
      , mSettings{ GetSettings(access.Get()) }
   {}
```

src/effects/How to make an effect stateless.txt
```diff
@@ -0,0 +1,189 @@
...
+        std::unique_ptr<EffectUIValidator>
+        EffectEcho::PopulateOrExchange(ShuttleGui& S, EffectSettingsAccess& access)
+        {
+           auto& settings = access.Get();
+           auto& myEffSettings = GetSettings(settings);
+           auto result = std::make_unique<Validator>(*this, access, myEffSettings);
+           result->PopulateOrExchange(S);
+           return result;
+        }
+
```
Paul-Licameli, line 183: There may be a fifth step:  see the commit I recently pushed to the branch for ladspa effects.

`MakeSettings` and `CopySettingsContents` may require explicit definitions instead of those generated by the template EffectWithSettings.

That does not apply to built-in effects where the settings structure simply contains several scalar fields and no variable sized containers.

For Ladspa, you can see that MakeSettings() instead allocates a vector of values, whose size is not known at compilation time but depends on other information that the effect is queried for.  So default construction of the settings, as in the generated MakeSettings(), will not do.

Then CopySettingsContents must also be overridden, not for correctness but for efficiency.  The override is supposed to avoid memory allocations when the main thread calls it, so it should not use the copy constructors of the embedded std::vector but instead rewrite an existing vector without changing its capacity.  The assumption will be made that the destination settings containers are already correctly sized, because MakeSettings did that, or else full copies done in the main thread (where allocations are allowed) already did that.

Other third party effect families like AudioUnits may have Settings structures containing maps from strings to values, instead of vectors, but the problems will be similar.

Paul-Licameli (approved):
There is no reason to block a merge of this text file that will not affect the build at all.

I point out many qualifications I could make to the procedure.  Update your document in response, or not, as you please.

pietro68:
Thanks for the feedback, I integrated it.

Issue #2834, opened 2022-04-19
Fix `Testing eqality to 'None'`warnings mentioned in LGTM report 


luzpaz:
**Describe the bug**
> When you compare an object to None, use is rather than ==. None is a singleton object, comparing using == invokes the __eq__ method on the object in question, which may be slower than identity comparison. Comparing to None using the is operator is also easier for other programmers to read.
>
> Recommendation: Replace `==` with `is`.

Ref: https://lgtm.com/rules/7900090/

**To Reproduce**
n/a

**Actual behavior**
n/a

**Expected behavior**
n/a

**Screenshots**
n/a

**Additional information (please complete the following information):**
 - OS: all
 - Version: HEAD

**Additional context**

Audacity LGTM report: https://lgtm.com/projects/g/audacity/audacity/

dozzzzer:
This cannot be changed since it belongs to a 3rd party library. We're also planning to move the code outside the Audacity source tree.

PR #2835, opened 2022-04-19
Fix LGTM 'Testing equality to None' alerts

luzpaz:
Resolves: https://github.com/audacity/audacity/issues/2834

Testing whether an object is `'None'` using the `==` operator is inefficient and potentially incorrect.

Recommended:


luzpaz:
Not sure if these are 3rd party libs or not. But I sent the PR nonetheless. Sorry for the noise if they are.

luzpaz:
Per @dozzzzer in https://github.com/audacity/audacity/issues/2834#issuecomment-1102805279
> This cannot be changed since it belongs to a 3rd party library. We're also planning to move the code outside the Audacity source tree.

Again, sorry for the noise

PR #2843, opened 2022-04-20
Audio unit cleanup 1

Paul-Licameli:
Resolves: #2840

In AudioUnit effects, clarify the confusions about "Plug-in" versus "Instance" initialization.

Besides that first commit, much other code cleanup matters, like modernization with custom smart pointer usage
(and some memory leak fixes), but nothing very important structurally.

Recommended:


src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -502,19 +502,14 @@ class AudioUnitEffectOptionsDialog final : public wxDialogWrapper
 {
 public:
    AudioUnitEffectOptionsDialog(
-      wxWindow * parent, EffectDefinitionInterface &effect);
+      wxWindow * parent, bool &useLatencey, wxString &uiType);
```
pietro68, line 492: typo "Latencey"

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -973,8 +959,10 @@ bool AudioUnitEffect::SupportsAutomation() const
 
 bool AudioUnitEffect::CreateAudioUnit()
 {
-   auto result = AudioComponentInstanceNew(mComponent, &mUnit);
-   return (!result && mUnit != nullptr);
+   AudioUnit unit{};
+   auto result = AudioComponentInstanceNew(mComponent, &unit);
+   mUnit.reset(unit);
```
pietro68, line 951: I wonder if it would make sense to check `result ` before resetting the smart pointer?

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -1043,34 +1043,54 @@ AudioUnitEffect::MakeInstance(EffectSettings &settings) const
...
 AudioUnitEffect::DoMakeInstance(EffectSettings &settings)
 {
-   OSStatus result;
-
    if (mMaster)
-      // Do common steps
-      InitializePlugin();
-
-   if (!mMaster) {
+      // This is a slave
```
pietro68, line 990: Not an object of this PR, but: I find the naming of this member `mMaster` very counterintuitive. The comment on its definition says `// non-NULL if a slave`, and the two lines here just confirm this.

pietro68 (approved):
I understand the changes, except for commit "Rewrite smart pointer for Apple Core resources" - if it is worth, we could have a live chat to explain it.

Paul-Licameli:
> I understand the changes, except for commit "Rewrite smart pointer for Apple Core resources" - if it is worth, we could have a live chat to explain it.

Ah yes, I full wanted to invite @ksoze95 to review that one commit.  I think he will enjoy all the C++ technicalities in it.

src/effects/audiounits/AudioUnitEffect.cpp
Paul-Licameli: I will do that when I merge it!

Issue #2841, opened 2022-04-20
Reduce much verbosity and repetition in AudioUnit code


Paul-Licameli:

dozzzzer:
@Paul-Licameli I'm getting a crash when trying to open up an AU plug-in.
OS: MacOS Monterey
Version: [audacity-macOS-3.2.0-alpha-20220511+d54f67f-universal](https://github.com/audacity/audacity/suites/6466694314/artifacts/238943396)


https://user-images.githubusercontent.com/70063970/168112815-f2813f09-bd2d-49be-a8c7-272856319995.mp4

Crash Report:
[Audacity_dbgrpt-28294-20220512T173139.zip](https://github.com/audacity/audacity/files/8680557/Audacity_dbgrpt-28294-20220512T173139.zip)

AU Plugin used:
https://www.audiopluginsforfree.com/tdr-nova/

dozzzzer:
> Did you enable your AU plug-ins in another build? I am failing to enable new AUs in master now.

Yes, I enabled them in an earlier version.

PR #2844, opened 2022-04-20
Audio unit cleanup 2

Paul-Licameli:
Resolves: #2841

Much code in AudioUnits calling into the SDK was too repetitious.

Define many utilities to factor out repeated patterns and greatly shorten AudioUnit.cpp, making many individual functions
more easily comprehensible.

Recommended:

### DIFF
--- a/src/effects/audiounits/AudioUnitEffect.cpp
+++ b/src/effects/audiounits/AudioUnitEffect.cpp
@@ -1105,11 +1105,10 @@ size_t AudioUnitEffect::GetTailSize() const
 bool AudioUnitEffect::ProcessInitialize(
    EffectSettings &, sampleCount, ChannelNames chanMap)
 {
-   mInputList.reinit(mAudioIns);
-   mInputList[0].mNumberBuffers = mAudioIns;
-   
-   mOutputList.reinit(mAudioOuts);
-   mOutputList[0].mNumberBuffers = mAudioOuts;
+   mInputList =
+      PackedArrayAllocateCount<AudioBufferList>(mAudioIns)(mAudioIns);
+   mOutputList =
+      PackedArrayAllocateCount<AudioBufferList>(mAudioOuts)(mAudioOuts);
 
    memset(&mTimeStamp, 0, sizeof(AudioTimeStamp));
    mTimeStamp.mSampleTime = 0; // This is a double-precision number that should
@@ -1140,26 +1139,24 @@ bool AudioUnitEffect::ProcessFinalize()
 {
    mOutputList.reset();
    mInputList.reset();
-
    return true;
 }
 
 size_t AudioUnitEffect::ProcessBlock(EffectSettings &,
    const float *const *inBlock, float *const *outBlock, size_t blockLen)
 {
-   for (size_t i = 0; i < mAudioIns; i++)
-   {
-      mInputList[0].mBuffers[i].mNumberChannels = 1;
-      mInputList[0].mBuffers[i].mData = const_cast<float*>(inBlock[i]);
-      mInputList[0].mBuffers[i].mDataByteSize = sizeof(float) * blockLen;
-   }
+   // mAudioIns and mAudioOuts don't change after plugin initialization,
+   // so ProcessInitialize() made sufficient allocations
+   assert(PackedArrayCount(mInputList) >= mAudioIns);
+   for (size_t i = 0; i < mAudioIns; ++i)
+      mInputList[i] = { 1, static_cast<UInt32>(sizeof(float) * blockLen),
+         const_cast<float*>(inBlock[i]) };
 
-   for (size_t i = 0; i < mAudioOuts; i++)
-   {
-      mOutputList[0].mBuffers[i].mNumberChannels = 1;
-      mOutputList[0].mBuffers[i].mData = outBlock[i];
-      mOutputList[0].mBuffers[i].mDataByteSize = sizeof(float) * blockLen;
-   }
+   // See previous comment
+   assert(PackedArrayCount(mOutputList) >= mAudioOuts);
+   for (size_t i = 0; i < mAudioOuts; ++i)
+      mOutputList[i] = { 1, static_cast<UInt32>(sizeof(float) * blockLen),
+         outBlock[i] };
 
    AudioUnitRenderActionFlags flags = 0;
    OSStatus result;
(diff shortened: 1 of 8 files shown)


src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -1911,13 +1844,7 @@ bool AudioUnitEffect::SavePreset(const RegistryPath & group) const
...
    // Now retrieve the preset content
    CF_ptr<CFPropertyListRef> content;
-   UInt32 size = sizeof(content);
-   AudioUnitGetProperty(mUnit.get(),
-                        kAudioUnitProperty_ClassInfo,
-                        kAudioUnitScope_Global,
-                        0,
-                        &content,
-                        &size);
+   GetFixedSizeProperty(kAudioUnitProperty_ClassInfo, content);
```
pietro68: maybe we could add a check on the return value?

src/effects/audiounits/AudioUnitUtils.h
```diff
@@ -15,6 +15,23 @@
...
 namespace AudioUnitUtils {
+   //! Type-erased function to get an AudioUnit property of fixed size
+   OSStatus GetFixedSizePropertyPtr(AudioUnit unit, AudioUnitPropertyID inID,
+      void *pProperty, UInt32 size, AudioUnitScope inScope,
+      AudioUnitElement inElement);
+
+   //! Get an AudioUnit property of deduced type and fixed size,
+   //! supplying most often used values as defaults for scope and element
+   template<typename T>
+   OSStatus GetFixedSizeProperty(AudioUnit unit, AudioUnitPropertyID inID,
```
pietro68, line 27: I like these helpers, and I was wondering whether instead of returning an OSStatus (which I understand is 0 for success) they could just return a bool - this because the if statements which use these helpers, at first sight display an "inverted logic" which causes fatigue for the reader of the code.

Paul-Licameli:
@ksoze95 may like to review "Define smart pointer utility PackedArrayPtr"

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -1152,26 +1151,21 @@ bool AudioUnitEffect::ProcessFinalize()
...
-   for (size_t i = 0; i < mAudioIns; i++)
-   {
-      mInputList[0].mBuffers[i].mNumberChannels = 1;
-      mInputList[0].mBuffers[i].mData = const_cast<float*>(inBlock[i]);
-      mInputList[0].mBuffers[i].mDataByteSize = sizeof(float) * blockLen;
-   }
+   assert(PackedArrayCount(mInputList) >= mAudioIns);
+   for (size_t i = 0; i < mAudioIns; ++i)
+      mInputList[i] = { 1, static_cast<UInt32>(sizeof(float) * blockLen),
+         const_cast<float*>(inBlock[i]) };
```
pietro68, line 1164: I might be wrong but... this change is not equivalent? the previous code loops on the mBuffers[i] of mInputList[0]; the new code loops on mInputLists instead, always assigning only the first mBuffer.

pietro68: Besides, I would keep the original assigning by field, instead of the aggregate assignment. Easier to read and more resilient to future changes in the struct.

src/effects/audiounits/AudioUnitUtils.h
Paul-Licameli: It seems some inconsistent inverted logic is unavoidable so long as we call other functions in the AudioUnits sdk that return a 0 value for success.

In principle OSStatus could distinguish different kinds of errors with different values, but funny I can't find any enumeration of those values in the header files, just a typedef of OSStatus as SInt32.  🤷

src/effects/audiounits/AudioUnitEffect.cpp
Paul-Licameli: Yes, I will rewrite the commit for getting fixed sized properties to check return statuses always, here and elsewhere

src/effects/audiounits/AudioUnitEffect.cpp
Paul-Licameli: I hope you are reading my commit comments as well as code comments.  I explained in one:

"Note that AudioUnitGetProperty can retrieve less of the data than are available
    from the plug-in if it doesn't first check with AudioUnitGetPropertyInfo and
    it passes a small size.
    
    The SDK appears to do this to allow versioning of the structures while not
    breaking code written to prior versions."

For that reason, I suspect that what you suggest is an unnecessary caution:  Apple would be smart enough not to do
that, breaking Audacity and any other audio program getting reinstalled on upgraded versions of the operating system.

But then I shouldn't read their minds either.  So I will insert another commit in this branch that adds some more
C++ wrappers around the SDK structures, in the header `AudioUnitUtils.h`, and I will amend some other commits to
use those wrappers rather than directly using the SDK.  Wait for it...

Paul-Licameli: > I assume that the change in what is being indexed here, is the correction of the formal error mentioned in the commit message.

The old version was subscripting `AudioBufferList::mBuffers` (`CoreAudioBaseTypes.h`) after it had allocated it "wrong" though harmlessly wrong (a bit of overallocation)

```
struct AudioBufferList
{
    UInt32      mNumberBuffers;
    AudioBuffer mBuffers[1]; // this is a variable length array of mNumberBuffers elements
};
```

The "proper" thing to do is not to allocate an array of `AudioBufferList` structures, but to do the things that `PackedArrayAllocateCount` does:  make space for one `AudioBufferList ` plus sufficient for extra `AudioBuffer` objects, whose number is known only after first calling `AudioUnitGetPropertyInfo`.

Paul-Licameli: So maybe I'm just being pedantic fixing what never caused a bug, but, it seemed to be a bit of old confusion of thought that is now better expressed.

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -1152,26 +1151,21 @@ bool AudioUnitEffect::ProcessFinalize()
...
-   for (size_t i = 0; i < mAudioOuts; i++)
-   {
-      mOutputList[0].mBuffers[i].mNumberChannels = 1;
-      mOutputList[0].mBuffers[i].mData = outBlock[i];
-      mOutputList[0].mBuffers[i].mDataByteSize = sizeof(float) * blockLen;
-   }
+   assert(PackedArrayCount(mOutputList) >= mAudioOuts);
+   for (size_t i = 0; i < mAudioOuts; ++i)
+      mOutputList[i] = { 1, static_cast<UInt32>(sizeof(float) * blockLen),
+         outBlock[i] };
```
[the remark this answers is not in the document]
Paul-Licameli: ditto

pietro68 (approved):
I think I understand the changes now, nothing else to note besides my comments

Issue #2846, opened 2022-04-20
Split New can create a spurious unwanted additional silent clip


petersampsonaudacity:
**Describe the bug**
If you select a lop or part of a clip and extend the selection rightwards beyond the clip (but not so as to abut or overlap the next adjacent clip) the Split New causes a spurious unwanted additional silent clip to be created. 

And note the silent clip is in-named.

**To Reproduce**
1. Create a 30 second chirp
2. Split the clip at 15 seconds
3. Move the second clip rightwards by 15 seconds
4. Select from 10-20 seconds
5. Edit > Audio Clips > Split New
6. Observe: a spurious unwanted silent clip is produced from 15-20 seconds
7. Observe: the clip has no clip-name

**Actual behavior**
A spurious unwanted silent clip is produced

**Expected behavior**
No spurious unwanted silent clip should be produced

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/164269396-95a4d167-385e-4c6e-88ff-d9fbfa56f477.png)

**Additional information (please complete the following information):**
 - OS: Windows 10 - but assume all OS
 - Version 3.2.0 latest master audacity-win-3.2.0-alpha-20220420+97d0c34-x64-msvc2022

**Additional context**
1) The first set of steps, above - blank space on the right appears to be a regression o 1.3.6 (or maybe 1.3.7-1.3.9 as I don't have those) - fails on 1.3.10 and later.

2) the second set of steps, below - blank space on the left appears to have been the case eve since Split New was introduced

petersampsonaudacity:
Selecting with blank space on the left of the clip produces a strand spurious result too.

1. Create a 30 second chirp
2. Split the clip at 15 seconds
3. Move the second clip rightwards by 15 seconds
4. Select from 25-30 seconds
5. Edit > Audio Clips > Split New
6. Observe: The 30-35 second split clip is moved to 25-30 seconds
7. 
![image](https://user-images.githubusercontent.com/50205984/164276953-1daac31e-b61c-4d79-b668-9fd15c841e0f.png)

petersampsonaudacity:
See also:   Split New fails with multiple partial clip selections #2816

Connorbub:
So just to be certain, the expected behavior is that we will never create a silent clip? If that's the case, then I believe the PR above should take care of that, as well as the issue with the end clip not lining up properly.

Connorbub:
For reference, here's the new behavior on the test case above:
[...]

PR #2847, opened 2022-04-20
Fix issue #2846 by removing the addition of a silent clip when copying clips

Connorbub:
Resolves: [https://github.com/audacity/audacity/issues/2846](https://github.com/audacity/audacity/issues/2846)

This changes the way copy (and by extension, split new) works in two ways:
1. No longer adds an unwanted silent clip
2. Fixes some lingering issues with copying clips over to the right time in the new track

Recommended:

### DIFF
--- a/src/WaveTrack.cpp
+++ b/src/WaveTrack.cpp
@@ -751,8 +751,7 @@ Track::Holder WaveTrack::Copy(double t0, double t1, bool forClipboard) const
       }
       else if (t1 > clip->GetPlayStartTime() && t0 < clip->GetPlayEndTime())
       {
-         // Clip is affected by command, but the entire clip is not contained
-         // in the selection
+         // Clip is affected by command
          //wxPrintf("copy: clip %i is affected by command\n", (int)clip);
 
          const double clip_t0 = std::max(t0, clip->GetPlayStartTime());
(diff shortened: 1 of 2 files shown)


petersampsonaudacity:
Tested on W10 with @Connorbub latest branch build  audacity-win-3.2.0-alpha-20220422+026e1ca-x64-msvc2022 
and
Tested on Macbook Pro Monterey 12.3.1 with @Connorbub latest branch build  audacity-macOS-3.2.0-alpha-20220422+026e1ca-x86_64 

All seems fine now with this latest branch build - _**ALL**_ use cases previously examined in the #2846 bug thread thread now work as expected.

There are no unwanted silences, no shiftings of audio position and no disappearing audio.

See:   Split New can create a spurious unwanted additional silent clip #2846 

@crsib @LWinterberg @dozzzzer   - from a user-test perspective this looks good to go.

src/WaveTrack.cpp
```diff
@@ -763,9 +764,19 @@ Track::Holder WaveTrack::Copy(double t0, double t1, bool forClipboard) const
...
-         if (newClip->GetPlayStartTime() < 0)
+         // If we are in the case with an audio clip on the right side with blank space to its left,
+         // we need to calculate the offset slightly differently:
+         //  |    [-----CLIP--|--]             [-|----CLIP----]      |
+         // t0                t1       VS.       t0                  t1
+         //  |----|                            |-|
+         //  OFFSET = clip_t0 - t0            OFFSET = 0
+         if (t0 < clip->GetPlayStartTime() && t1 < clip->GetPlayEndTime()) {
+            wxASSERT(clip_t0 - t0 >= 0); //debugging
+            newClip->SetPlayStartTime(clip_t0-t0); //not sure why Offset doesn't work for this; may want to modify
```
vsverchinsky, line 775: `Offset` moves clip beginning on `delta` seconds. Not sure which argument You were trying to pass to it but `Offset(-t0)` should work, and as I said it actually yields same result as `SetPlayStartTime(clip_t0 - t0)` in that case. So what was the case where `Offset` didn't work? If it didn't work for wrong arguments then please revert those changes because they do not change anything

src/WaveTrack.cpp
```diff
@@ -763,9 +764,19 @@ Track::Holder WaveTrack::Copy(double t0, double t1, bool forClipboard) const
...
 
-         newClip->Offset(-t0);
-         if (newClip->GetPlayStartTime() < 0)
+         // If we are in the case with an audio clip on the right side with blank space to its left,
+         // we need to calculate the offset slightly differently:
+         //  |    [-----CLIP--|--]             [-|----CLIP----]      |
+         // t0                t1       VS.       t0                  t1
+         //  |----|                            |-|
+         //  OFFSET = clip_t0 - t0            OFFSET = 0
+         if (t0 < clip->GetPlayStartTime() && t1 < clip->GetPlayEndTime()) {
```
vsverchinsky, line 773: `t1 < clip->GetPlayEndTime()` condition is redundant. Both `t1 >= clip->GetPlayEndTime()` and `t0 < clip->GetPlayStartTime()` isn't possible in that if-branch

vsverchinsky (changes requested):
@Connorbub 
Changes in `EditMenus.cpp` only are good and seem to be enough to fix the issue. `offset` variable become useless and should be removed. 
Changes in `WaveTrack.cpp` do not change the behavior and should be reverted.
Also please squash all commits in this PR into one.
Thank you!

vsverchinsky:
Closing in favor of #4294 that contains squashed commits from original PR

Issue #2489, opened 2022-01-28
Does not load FFmpeg's libavformat.so.58.76.100


lfam:
**Describe the bug**
I am using Audacity 3.1.3 as [packaged](https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/audio.scm?id=f870977e6ac5a37f1e42af7e97c78c1e28dadfc8#n724) by GNU Guix (this is Linux), and I am having trouble using FFmpeg. I am trying with FFmpeg 4.4.1. But, I cannot make it work.

I know that Audacity stopped linking to FFmpeg in recent releases, so I am attempting to load it.

I go through the menus `Edit -> Preferences -> Libraries -> Locate...`, browse to the correct location, and select the correct path, `/gnu/store/31fag7lys7fa4rxb0cbmdc43qh26lnsh-ffmpeg-4.4.1/lib/libavformat.so.58.76.100`. This is the location of the library on Guix.

However, Audacity still does not make use of it, and when I return to the `Libraries` menu, it says `FFmpeg Library Version: FFmpeg library not found`. Also, I check `audacity.cfg`, and it shows that the library's location was not registered:

```
[FFmpeg]
Enabled=0
FFmpegLibPath=
```

I can manually edit `audacity.cfg` to contain these lines:

```
[FFmpeg]
Enabled=1
FFmpegLibPath=/gnu/store/31fag7lys7fa4rxb0cbmdc43qh26lnsh-ffmpeg-4.4.1/lib/libavformat.so.58.76.100
```
But, when starting Audacity, it shows me the dialog `FFmpeg was configured in Preferences and successfully loaded before, but this time Audacity failed to load it at startup.`

For that case, I launched Audacity in strace and noticed that it does successfully stat `libavformat.so`:

```
[pid 51608] newfstatat(AT_FDCWD, "/gnu/store/31fag7lys7fa4rxb0cbmdc43qh26lnsh-ffmpeg-4.4.1/lib/libavformat.so.58.76.100", {st_mode=S_IFREG|0555, st_size=2527096, ...}, 0) = 0
[pid 51608] openat(AT_FDCWD, "/gnu/store/ny3b7qlnkhxaxvswsnnvddp2y45irfi5-audacity-3.1.3/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 16
[pid 51608] newfstatat(16, "", {st_mode=S_IFREG|0444, st_size=43264, ...}, AT_EMPTY_PATH) = 0
[...]

PR #2852, opened 2022-04-22
Make wahwah stateless

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2858

Wahwah was chosen as the next one to make stateless because unlike EffectEcho, it has explicitly declared widgets (and is relatively simple). Explicit widgets require more transformation operations, which were added to the guide in a new section in step 4.

Recommended:

### DIFF
--- a/src/effects/How to make an effect stateless.txt
+++ b/src/effects/How to make an effect stateless.txt
@@ -17,11 +17,17 @@ Let use use EffectEcho as an example.
        Modify FetchParameters: it must now return the address of mSettings, not of the effect
     
        Update pointers-to-members in EffectParameter declarations
+
+       you can remove forward declarations like:
+         class wxCheckBox;
+         class wxSlider;
+         class wxSpinCtrl;
+
     
    In cpp file:
        Insert sufficient auto& echoSettings = mSettings; (which changes later) and echoSettings. (which won't need more change) to fix compilation
         
-       Remove call to Reset() in constructor
+       Remove call to Parameters().Reset(*this); in constructor
         
     
 2) "Define validator"
--- a/src/effects/How to make an effect stateless.txt
+++ b/src/effects/How to make an effect stateless.txt
@@ -126,9 +126,24 @@ Let use use EffectEcho as an example.
            size_t ProcessBlock(EffectSettings& settings,
               const float* const* inBlock, float* const* outBlock, size_t blockLen)  override;
 
-           bool ProcessFinalize(void) override;
+           bool ProcessFinalize(void) override; // not every effect needs this
+           
+           /* you will need these for realtime-capable effects
+           
+           bool RealtimeInitialize(EffectSettings& settings) override;
 
-           << MEMBERS WHICH REPRESENT STATE which you commented out earlier >>
+           bool RealtimeAddProcessor(EffectSettings& settings,
+              unsigned numChannels, float sampleRate) override;
+
+           bool RealtimeFinalize(EffectSettings& settings) noexcept override;
+
+           size_t RealtimeProcess(int group, EffectSettings& settings,
+              const float* const* inbuf, float* const* outbuf, size_t numSamples)
+              override;
+           
+           */
+
+           // << MEMBERS WHICH REPRESENT STATE which you commented out earlier >>
         };
 
 
@@ -143,7 +158,11 @@ Let use use EffectEcho as an example.
     
     - prepending ::Instance where needed
 
-    - adding this at the beginning (may not be needed in all of them)
+    - replacing the lines you added in step 1:
+    
+        auto& echoSettings = mSettings;
+            
+      with this 
     
         // temporary - in the final step this will be replaced by
         // auto& echoSettings = GetSettings(settings);
--- a/src/effects/How to make an effect stateless.txt
+++ b/src/effects/How to make an effect stateless.txt
@@ -18,64 +18,80 @@ Let use use EffectEcho as an example.
     
        Update pointers-to-members in EffectParameter declarations
 
-       you can remove forward declarations like:
-         class wxCheckBox;
-         class wxSlider;
-         class wxSpinCtrl;
+       
 
     
    In cpp file:
        Insert sufficient auto& echoSettings = mSettings; (which changes later) and echoSettings. (which won't need more change) to fix compilation
         
-       Remove call to Parameters().Reset(*this); in constructor
+       Go to the constructor and remove call to
+
+         Parameters().Reset(*this);
+
+       As the initializations it did, now happen in the constructor of EffectEchoSettings because we use in-struct initializers with the same default values as in the EffectParmeters.
         
     
 2) "Define validator"
 
+
     In header file:
-    
-    add line "struct Validator;" in public section
+
+      - you can remove forward declarations like:
+         class wxCheckBox;
+         class wxSlider;
+         class wxSpinCtrl;
+
+      - add line "struct Validator;" in public section
+
+      - if the following is declared, remove it:
+         bool TransferDataToWindow(const EffectSettings &settings) override;
+
 
 
     In cpp file:
-    add class definition in the .cpp with mSettings:
+
+       add class definition in the .cpp with mSettings:
     
-        struct EffectEcho::Validator
-           : DefaultEffectUIValidator
-        {
-           Validator(EffectUIClientInterface& effect,
-              EffectSettingsAccess& access, EffectEchoSettings& settings)
-              : DefaultEffectUIValidator{ effect, access }
-              , mSettings{ settings }
-           {}
-           virtual ~Validator() = default;
-
-           Effect& GetEffect() const { return static_cast<Effect&>(mEffect); }
-
-           bool ValidateUI() override;
-           bool UpdateUI() override;
-
-           void PopulateOrExchange(ShuttleGui& S);
-
-           EffectEchoSettings& mSettings;
-        };
-
-
-        bool EffectEcho::Validator::ValidateUI()
-        {
-           mAccess.ModifySettings
-           (
-              [this](EffectSettings& settings)
+           struct EffectEcho::Validator
+              : EffectUIValidator
            {
-              // pass back the modified settings to the MessageBuffer
+              Validator(EffectUIClientInterface& effect,
+                 EffectSettingsAccess& access, EffectEchoSettings& settings)
+                 : EffectUIValidator{ effect, access }
+                 , mSettings{ settings }
+              {}
+              virtual ~Validator() = default;
 
-              // TODO uncomment at last step
-              //EffectEcho::GetSettings(settings) = mSettings;
+              Effect& GetEffect() const { return static_cast<Effect&>(mEffect); }
+
+              bool ValidateUI() override;
+              bool UpdateUI() override;
+
+              void PopulateOrExchange(ShuttleGui& S);
+
+              EffectEchoSettings& mSettings;
+           };
+
+
+           bool EffectEcho::Validator::ValidateUI()
+           {
+              mAccess.ModifySettings
+              (
+                 [this](EffectSettings& settings)
+              {
+                 // pass back the modified settings to the MessageBuffer
+
+                 // TODO uncomment at last step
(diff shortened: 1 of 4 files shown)


src/effects/How to make an effect stateless.txt
```diff
@@ -21,7 +21,7 @@ Let use use EffectEcho as an example.
    In cpp file:
        Insert sufficient auto& echoSettings = mSettings; (which changes later) and echoSettings. (which won't need more change) to fix compilation
         
-       Remove call to Reset() in constructor
+       Remove call to Parameters().Reset(*this); in constructor
```
Paul-Licameli, line 24: What makes it correct at this step to remove this call -- is that the initializations it formerly did now happen in the constructor of `EffectEchoSettings` because you use in-struct initializers with the same default values as in the `EffectParmeters`.

So that is worth mention too in the guide.

src/effects/Wahwah.h
```diff
@@ -37,11 +37,42 @@ class EffectWahwahState
...
+    mDepth - Wah depth
+    mRes - Resonance
+    mFreqOfs - Wah frequency offset
+    mOutGain - output gain
+
+      !!!!!!!!!!!!! IMPORTANT!!!!!!!!! :
+      mDepth and mFreqOfs should be from 0(min) to 1(max) !
+      mRes should be greater than 0 !  */
+
+   static constexpr double mFreqDefault = 1.5;
```
Paul-Licameli, line 55: We aren't strict about naming conventions.  But the "m" prefix suggests a nonstatic data member, while these are just constants.  I would simply use no prefix.

src/effects/Wahwah.cpp
```diff
@@ -60,24 +51,144 @@ const ComponentInterfaceSymbol EffectWahwah::Symbol
...
-    EVT_TEXT(ID_Freq, EffectWahwah::OnFreqText)
-    EVT_TEXT(ID_Phase, EffectWahwah::OnPhaseText)
-    EVT_TEXT(ID_Depth, EffectWahwah::OnDepthText)
-    EVT_TEXT(ID_Res, EffectWahwah::OnResonanceText)
-    EVT_TEXT(ID_FreqOfs, EffectWahwah::OnFreqOffText)
-    EVT_TEXT(ID_OutGain, EffectWahwah::OnGainText)
-END_EVENT_TABLE();
+
+struct EffectWahwah::Validator
+   : DefaultEffectUIValidator
```
Paul-Licameli, line 56: As I explained Wednesday, this should use EffectUIValidator as its base class now because it overrides ValidateUI for itself.  DefaultEffectUIValidator now exists only to default that function.  BindTo is available just from EffectUIValidator.

src/effects/Wahwah.cpp
```diff
@@ -60,24 +51,144 @@ const ComponentInterfaceSymbol EffectWahwah::Symbol
...
+   return true;
+}
+
+struct EffectWahwah::Instance
+   : public PerTrackEffect::Instance
+   , public EffectInstanceWithBlockSize
+   , public EffectInstanceWithSampleRate
+
+{
+   Instance(const PerTrackEffect& effect)
```
Paul-Licameli, line 149: Maybe add "explicit"

src/effects/Wahwah.cpp
```diff
@@ -60,24 +51,144 @@ const ComponentInterfaceSymbol EffectWahwah::Symbol
 
 namespace{ BuiltinEffectsModule::Registration< EffectWahwah > reg; }
 
-BEGIN_EVENT_TABLE(EffectWahwah, wxEvtHandler)
```
Paul-Licameli, line 63: The elimination of this table, and the use of BindTo calls instead, could be done at step 2 of the transformation.

src/effects/Wahwah.cpp
```diff
@@ -60,24 +51,144 @@ const ComponentInterfaceSymbol EffectWahwah::Symbol
...
-END_EVENT_TABLE();
+
+struct EffectWahwah::Validator
+   : DefaultEffectUIValidator
+{
+   Validator(Effect& effect,
+      EffectSettingsAccess& access, const EffectWahwahSettings& settings)
+      : DefaultEffectUIValidator{ effect, access }
+      , mSettings{ settings }
+      , mActualEffect{ effect }
```
Paul-Licameli, line 62: I think this member is unnecessary. You can use the inherited mEffect and `static_cast` it and write an inline convenience function.  In fact you did that already at step 2.

src/effects/Wahwah.h
```diff
@@ -60,97 +91,27 @@ class EffectWahwah final : public StatefulPerTrackEffect
...
-   void OnResonanceSlider(wxCommandEvent & evt);
-   void OnFreqOffSlider(wxCommandEvent & evt);
-   void OnGainSlider(wxCommandEvent & evt);
-
-   void OnFreqText(wxCommandEvent & evt);
-   void OnPhaseText(wxCommandEvent & evt);
-   void OnDepthText(wxCommandEvent & evt);
-   void OnResonanceText(wxCommandEvent & evt);
-   void OnFreqOffText(wxCommandEvent & evt);
-   void OnGainText(wxCommandEvent & evt);
```
Paul-Licameli, line 101: The moving of event handlers to the Validator should be part of step 2

src/effects/Wahwah.h
```diff
@@ -60,97 +91,27 @@ class EffectWahwah final : public StatefulPerTrackEffect
...
-   mRes should be greater than 0 !  */
-
-   double mFreq;
-   double mPhase;
-   int mDepth;
-   double mRes;
-   int mFreqOfs;
-   double mOutGain;
-
-   wxTextCtrl *mFreqT;
```
Paul-Licameli, line 125: The moving of pointers to controls into the Validator should be done at step 2

src/effects/Wahwah.cpp
```diff
@@ -60,24 +51,144 @@ const ComponentInterfaceSymbol EffectWahwah::Symbol
...
+
+
+bool EffectWahwah::Validator::ValidateUI()
+{
+   mAccess.ModifySettings
+   (
+      [this](EffectSettings& settings)
+      {
+         // pass back the modified settings to the MessageBuffer
+         EffectWahwah::GetSettings(settings) = mSettings;
```
Paul-Licameli, line 116: The `EffectWahWah::` qualifier can be omitted for brevity.  Static members of EffectWahWah are in scope in member functions of its embedded struct Validator.  Likewise for Instance.

src/effects/Wahwah.cpp
```diff
@@ -345,107 +471,138 @@ size_t EffectWahwah::InstanceProcess(EffectSettings &settings,
...
 {
-   mFreq = (double) evt.GetInt() / Freq.scale;
+   auto& ms = mSettings;
+
+   ms.mFreq = (double) evt.GetInt() / Freq.scale;
    mFreqT->GetValidator()->TransferToWindow();
-   EnableApply(mUIParent->Validate());
+
+   // TODO code this into a single method of class Effect, call it e.g. "EnableApplyFromValidate"
+   mActualEffect.EnableApply(mActualEffect.GetUIParent()->Validate());
```
Paul-Licameli, line 482: This is incomplete.  Every one of the `::On` functions that handles an event and changes mSettings needs to send that change into the message buffers.  So add a call to ValidateUI at the end of each.

In future, I hope Kit and I will merge a lot of work that will simplify the defintion of effect dialogs and make these event handler functions unnecessary: instead PopulateOrExchange can encode a complete and short description of how values are copied and how one control can be updated to be consistent with another (as with these slider and text pairs).  But, that's future.

Paul-Licameli (changes requested):
I'm done reviewing.  The last comment is important for correctness, when issue #2489 is fixed, or else WahWah (but not Ladspa) effects will still suffer the bug described in that issue.

There are other suggestions, importantly to do more transformations in step 2 where they really belong, rather than 4.

Paul-Licameli:
Notice that I call ValidateUI at the end of every event handler of LadspaEffect::Instance

Paul-Licameli:
I made a new issue #2858 of the checkbox item for this and linked it to this PR

src/effects/Wahwah.cpp
pietro68: Ok I missed that. Interestingly, the effect works correctly (applying the right amounts and remembering them) in destructive mode even without that final call to ValidateUI(), but I guess the call would be needed when using it in realtime mode?

pietro68:
I applied your feedback in per-steps commits for your convenience, and will squash them later, so that in the end we'll have only 5 commits: the 4 steps and the update of the guide.

pietro68:
One final thing I will then do in an interactive rebase, is to move the things that could have been done in step 2 from step 4 to step 2.

Paul-Licameli:
The recent changes all look good.  It remains to squash it.

src/effects/How to make an effect stateless.txt
```diff
@@ -35,7 +39,7 @@ Let use use EffectEcho as an example.
     add class definition in the .cpp with mSettings:
     
         struct EffectEcho::Validator
-           : DefaultEffectUIValidator
+           : EffectUIValidator
```
Paul-Licameli, line 42: The initialization of the base class below also changes the class name

Paul-Licameli:
You will find that not all effects have an override of `TransferDataToWindow`.  Some my have `TransferDataFromWindow` which this effect did not.  If the latter function is defined, its steps are moved into ValidateUI() in step 2 when you define that function.

Paul-Licameli:
I hope you understand that when I say certain changes should really belong in step 2 not step 4 of the transformation, it is because I want step 4 really to be only the essential steps of doing what its commit comment says -- switching the effect to real statelessness and externalized settings.

src/effects/How to make an effect stateless.txt
pietro68: oops that's right

pietro68:
> I hope you understand that when I say certain changes should really belong in step 2 not step 4 of the transformation, it is because I want step 4 really to be only the essential steps of doing what its commit comment says -- switching the effect to real statelessness and externalized settings.

yes, I will do that in an interactive rebase and then force-push here

pietro68:
> You will find that not all effects have an override of `TransferDataToWindow`. Some my have `TransferDataFromWindow` which this effect did not. If the latter function is defined, its steps are moved into ValidateUI() in step 2 when you define that function.

info now added to guide, at step 2

Paul-Licameli:
I see the changes.  The final state of the code is good.

I notice now that even more lines might be removed from the header file.  The forward declarations of the wxControl subclasses is no longer needed.

I hate to ask for another round of review, but step 4 still changes more lines than I would if preparing this branch for a review, and line 2 still changes too few.

Step 2 declares the event handlers for the Validator, but does not define them (by moving them out of the Effect class) until 4.

Paul-Licameli:
Step 3 can also delete some declarations of initialization functions that move into Instance.

pietro68:
> Step 3 can also delete some declarations of initialization functions that move into Instance.

I found the following two and removed them: 

`void InstanceInit(EffectWahwahState & data, float sampleRate);`

```
size_t InstanceProcess(EffectSettings &settings, EffectWahwahState & data,
      const float *const *inBlock, float *const *outBlock, size_t blockLen);

```

pietro68:
> I notice now that even more lines might be removed from the header file. The forward declarations of the wxControl subclasses is no longer needed.

ok, done

> Step 2 declares the event handlers for the Validator, but does not define them (by moving them out of the Effect class) until 4.

The thing is, by postponing the definition of the handlers to step 4, I thought the diffs can show better what needs to be done - i.e. only inserting ::Validator to the start of the handlers, plus a few small changes - without having duplicate code. But I also see a reason to specify as much of Validator as possible in the step dedicated to it, so I will proceed with that.

Paul-Licameli (changes requested):
Now the steps 1, 2, and 3 fail to build.

Paul-Licameli (approved):

Paul-Licameli:
One remaining small suggestion:  At step 3, the new declaration of `MakeInstance` should include `override`, which will suppress a compiler warning at least with clang on macOS.

Issue #2863, opened 2022-04-27
Debug-only crash opening files containing stereo tracks


Paul-Licameli:
- Start debug build of Audacity
- Tracks > Add New > Stereo Track
- Force-quit
- Restart
- Choose Recover

Alternatively:  simply save, close, and reload a project.

A crash results, inside an `assert`, which (IMPROPERLY FOR AN ASSERT!) is trying to make a side-effect to repair the inconsistency.

The bug won't happen in normal release builds, because `assert` compiles to nothing.

So two things to fix:
- Don't make a side effect
- Don't falsely detect inconsistency when opening a file and checking consistency of a TrackList that is not yet completely rebuilt

This problem began at commit 93be3aeb

Paul-Licameli:
@crsib

PR #2864, opened 2022-04-27
Rewrite track link consistency checking...

Paul-Licameli:
... Make a non-mutating version that only checks without doing repairs, safe
to use in a debug assertion.

Don't violate the assertion incorrectly when the reloading of the list of tracks
is in progress.

Fixes error from 93be3aeb that could crash debug build when opening a project.

All other uses of LinkConsistencyCheck() predated that commit, and are changed
to call the new name LinkConsistencyFix().

Resolves: #2863

Recommended:

### DIFF
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -150,7 +150,7 @@ void Track::SetIndex(int index)
    mIndex = index;
 }
 
-void Track::SetLinkType(LinkType linkType)
+void Track::SetLinkType(LinkType linkType, bool completeList)
 {
    auto pList = mList.lock();
    if (pList && !pList->mPendingUpdates.empty()) {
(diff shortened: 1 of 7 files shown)


crsib (approved):
More complex fix than I expected. Also looks like that `completeList == false` implies `Track::LinkConsistencyFix() == true`, but that is fine

Paul-Licameli:
> More complex fix than I expected. Also looks like that `completeList == false` implies `Track::LinkConsistencyFix() == true`, but that is fine

That is the intention:  a weaker consistency check is appropriate, when setting the link type of WaveTrack during deserialization, while the complete track list is not yet known.

Assertions should surely not have side effects!  Shame on me for forgetting that, or forgetting that the ConsistencyCheck in fact had side effects.

Issue #2845, opened 2022-04-20
Add&Remove effects from the sidebar


AnitaBats:

DavidBailes:
Is there a design specification for this which includes the keyboard user interaction?

DavidBailes:
@LWinterberg : Is there a design specification for this which includes the keyboard user interaction?

dozzzzer:
Reopened: a few issue were discovered

### **Issue 1: An effect called from sidebar UI doesn't apply if playback is stopped.**

STR:

- generate a track/load an audio clip
- click [Add Effects]
- select an effect
- play the clip

Actual result: no changes to the audio (see the video attachment); the effect applies _only_ when selected during playback
Expected result: the audio is changed according to a selected effect

https://user-images.githubusercontent.com/70063970/170982832-0f286b69-a972-4948-99cc-0de95cb22079.mp4

### **Issue 2: Audio Unit modules of some plug-ins aren't populated to the list of effects in the stackable effects UI**

Compare: 3 instances (AU, VST, VST3) of the effect when called from 'Effect...' menu:
<img width="1231" alt="image" src="https://user-images.githubusercontent.com/70063970/170983086-f31314c8-e094-4124-ba49-3b723a820aae.png">

vs only 2 instances of the same effect when called via stackable effects UI:
<img width="412" alt="image" src="https://user-images.githubusercontent.com/70063970/170983171-e636c348-bc0f-4a3c-8fad-2c20df9e7592.png">

dozzzzer:
**### Issue 4 (minor): Hover state of the "Add new effect" button is missing**
<img width="269" alt="image" src="https://user-images.githubusercontent.com/70063970/171145417-d340a036-ccc3-46f7-9ebd-51b6d0a9760a.png">

Paul-Licameli:
I think these various new problems could have separate new issue cards referred to by a checklist

LWinterberg:
issues mentioned above have been split out to new bugs (3113~4), closing

PR #2867, opened 2022-04-28
Issue 2845: effects stack UI

vsverchinsky:
Resolves: #2845 

ToDo: On/Off toggle on the whole stack and individual effects does not work as it is not implemented in the underlying effects stack model.
Setting for individual effects is expected to work for effects that do support per instance settings

- [*] I signed [CLA](https://www.audacityteam.org/cla/)
- [*] The title of the pull request describes an issue it addresses
- [*] If changes are extensive, then there is a sequence of easily reviewable commits
- [*] Each commit's message describes its purpose and effects
- [*] There are no behavior changes unnecessary for the stated purpose of the PR

Recommended:


libraries/lib-theme/AllThemeResources.h
```diff
@@ -374,3 +386,7 @@ from there.  Audacity will look for a file called "Pause.png".
 
    DEFINE_COLOUR( clrClipNameText,          wxColour(0, 0, 0),       wxT("ClipNameText") );
    DEFINE_COLOUR( clrClipNameTextSelection, wxColour(183, 202, 226), wxT("ClipNameTextSelection") );
+
+   DEFINE_COLOUR( clrDropHintHighlight, wxColour( 71, 140, 247 ), wxT("DropHintHighlight") );
+   DEFINE_COLOUR( clrEffectListItemBackground, wxColour( 60, 60, 60 ), wxT("EffectListItemBackground"));
+   DEFINE_COLOUR( clrEffectListItemBorder, wxColour( 204, 204, 204 ), wxT("EffectListItemBorder"));
```
Paul-Licameli, line 392: This changes the format for the monolithic "sprite sheet" for dropping in a custom theme.  After merging this we must remember to give @LWinterberg the updated sheet to put in the Wiki page.  If more changes to this file are expected during this release cycle, then that can wait until they are all done.

src/widgets/ThemedWrappers.h
```diff
@@ -0,0 +1,114 @@
+/*!********************************************************************
+
+   Audacity: A Digital Audio Editor
+
+   @file ThemedWrappers.h
+
+   @author Vitaly Sverchinsky
+
+   A set of classes which are simplify theme support
```
Paul-Licameli, line 9: "which simplify"

src/widgets/ThemedWrappers.h
```diff
@@ -0,0 +1,114 @@
...
+   
+**********************************************************************/
+
+#pragma once
+
+#include <wx/app.h>
+#include "prefs/ThemePrefs.h"
```
Paul-Licameli, line 16: You include this for `EVT_THEME_CHANGE` and not `PrefsPanel`?

As you might have guessed, that is a usage of wxEvent that I would rather simplify with Observer.

In fact, I have a commit waiting in my big branch of module extractions doing just that.

I should cherry-pick it, and push it onto your branch, and let you merge it with this branch.

src/widgets/ThemedWrappers.h
Paul-Licameli: Perhaps say "which decorate given base classes with simplified theme support"

This is sort of a "decorator pattern" only it inserts behavior with inheritance, not object composition.

src/effects/RealtimeEffectList.cpp
```diff
@@ -104,11 +104,37 @@ void RealtimeEffectList::RemoveState(RealtimeEffectState &state)
       mStates.erase(found);
 }
 
+size_t RealtimeEffectList::GetStatesCount() const noexcept
+{
+   return mStates.size();
+}
+
+RealtimeEffectState* RealtimeEffectList::GetStateAt(size_t index) noexcept
```
Paul-Licameli, line 112: I'm more hesitant about using `noexcept` liberally.  I saw a CppCon talk by John Lakos describing reasons for caution.  I need to review it again.  I mostly use it only when I know something must be called inside a destructor.

I look at this and I see it doesn't have a bounds check on the array access.

Paul-Licameli: Given that the index is in bounds (might say it with a `@pre` comment), should this return a reference to a guaranteed non-null object?  Perhaps.  But maybe warn too about invalidation in case states are added or removed.

src/effects/RealtimeEffectList.cpp
```diff
@@ -104,11 +104,37 @@ void RealtimeEffectList::RemoveState(RealtimeEffectState &state)
...
+   if(fromIndex != toIndex)
+   {
+      auto effectState = std::move(mStates[fromIndex]);
+      mStates.erase(mStates.begin() + fromIndex);
+      
+      mStates.emplace(
+         mStates.begin() + (toIndex < fromIndex ? toIndex : toIndex - 1),
+         std::move(effectState)
+      );
+   }
```
Paul-Licameli, line 135: Use `std::rotate` to simplify this

Paul-Licameli: And becuase this does all that `Swap` can do, and more -- then remove `Swap`

src/widgets/RealtimeEffectPanel.h
```diff
@@ -0,0 +1,44 @@
...
+
+   @file RealtimeEffectPanel.cpp
+
+   @author Vitaly Sverchinsky
+   
+**********************************************************************/
+
+#pragma once
+
+#include <wx/scrolwin.h>
```
Paul-Licameli, line 13: Failing linux build wants (at least) `#include <memory>`

src/widgets/RealtimeEffectPanel.h
```diff
@@ -0,0 +1,44 @@
...
+   RealtimeEffectPanel(wxWindow *parent,
+                wxWindowID id,
+                const wxPoint& pos = wxDefaultPosition,
+                const wxSize& size = wxDefaultSize,
+                long style = 0,
+                const wxString& name = wxPanelNameStr);
+
+   /**
+    * \brief Shows effects from the effect stack of the track
+    * \param track Pointer to the existing track, or null
```
Paul-Licameli, line 41: Good comment, saying what they types don't alone imply

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
+/*!********************************************************************
+
+   Audacity: A Digital Audio Editor
+
+   @file RealtimeEffectPanel.cpp
+
+   @author Vitaly Sverchinsky
+   
+**********************************************************************/
```
Paul-Licameli, line 9: Please put this new file pair directly into `src` not `src/widgets`

The sub-folder division of `src` was never really ever done on any consistent principle, whereas the coming modularization will only move files again.

The seeming intention of `src/widgets` is for reusable, and reused, windows like `Ruler` but this looks like it will only ever have one use, closely associated with ProjectWindow.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+
+#include "RealtimeEffectPanel.h"
+
+#include <wx/sizer.h>
+#include <wx/button.h>
```
Paul-Licameli, line 14: `#include <wx/dcclient.h>` needed to build non-Windows

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+   //initial and final element position inside wxSizer (if present)
+   class MovableControlEvent final : public wxCommandEvent
+   {
+      int mSourceIndex{-1};
+      int mTargetIndex{-1};
+   public:
+      MovableControlEvent(wxEventType eventType, int winid = 0)
+         : wxCommandEvent(eventType, winid) { }
+
+      MovableControlEvent(const MovableControlEvent& other)
```
Paul-Licameli, line 57: Just default it?

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+      int mTargetIndex { -1 };
+      int mSourceIndex { -1 };
+   public:
+
+      MovableControl(wxWindow* parent,
+                   wxWindowID id,
+                   const wxPoint& pos = wxDefaultPosition,
+                   const wxSize& size = wxDefaultSize,
+                   long style = wxTAB_TRAVERSAL | wxNO_BORDER,
+                   const wxString& name = wxPanelNameStr)
```
Paul-Licameli, line 97: `name` is not used

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+
+         SetSizer(sizer.release());
+      }
+
+      void SetEffect(const std::shared_ptr<Track>& track, int effectIndex)
+      {
+         mTrack = track;
+
+         auto& effects = RealtimeEffectList::Get(*track);
+         assert (effectIndex >= 0 && effectIndex < effects.GetStatesCount());
```
Paul-Licameli, line 342: Proof?

If none, make a `@pre` comment to license it, but then examine the calls to this function , where the burden of proof is then shifted.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+         
+         Destroy();
+      }
+      
+      void OnOptionsClicked(wxCommandEvent& event)
+      {
+         if(mEffectState == nullptr)
+            return;//not initialized
+         
+         auto effectPlugin = EffectManager::Get().GetEffect(mEffectState->GetID());
```
Paul-Licameli, line 368: Do you need a null check on the return value?

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+         auto access = mEffectState->GetAccess();
+            
+         access->ModifySettings([&](EffectSettings& settings)
+         {
+            auto project = mTrack->GetOwner()->GetOwner();
+            auto &trackFactory = WaveTrackFactory::Get( *project );
+            auto &selectedRegion = ViewInfo::Get( *project ).selectedRegion;
+            auto rate = ProjectRate::Get(*project).GetRate();
+            auto &window = ProjectWindow::Get( *project );
+            auto flags = static_cast<unsigned>(EffectManager::kNone);
```
Paul-Licameli, line 379: I’d like to remove the need for this static_cast.   Not a fault of this PR

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+
+            effectPlugin->DoEffect(settings,
+               rate,
+               &*mTrack->GetOwner(),
```
Paul-Licameli, line 383: Assuming not null again as at 374

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+            || (a->GetVendor() == b->GetVendor() && a->GetSymbol().Translation() < b->GetSymbol().Translation());
+      };
+
+      for(auto& effect : pluginManager.EffectsOfType(EffectTypeProcess))
+      {
+         if(!effect.IsEffectRealtime() || !effect.IsEnabled())
+            continue;
+
+         auto it = std::lower_bound(effects.begin(), effects.end(), &effect, compareEffects);
+         effects.insert(it, &effect);
```
Paul-Licameli, line 456: Quadratic insertion sort instead of n log n std::sort? Shrug

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+
+         if(currentSubMenuName != effect.GetVendor())
+         {
+            if(currentSubMenu)
+            {
+               currentSubMenu->Bind(wxEVT_MENU, submenuEventHandler);
+               menu.AppendSubMenu(currentSubMenu.release(), currentSubMenuName);
+            }
+            currentSubMenuName = effect.GetVendor();
+            currentSubMenu = std::make_unique<wxMenu>();
```
Paul-Licameli, line 479: I imagine different sorting and grouping criteria might be future demands.  Do what now?  Nothing but think about it

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+
+      menu.Append(wxID_MORE, XO("Get more effects...").Translation());
+
+      menu.Bind(wxEVT_MENU, [&](wxCommandEvent& event)
+      {
+         if(event.GetId() == wxID_REMOVE)
+            currentEffectControl->RemoveFromList();
+         else if(event.GetId() == wxID_MORE)
+         {
+            //TODO: implement
```
Paul-Licameli, line 496: And?

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+         }
+      });
+
+      if(parent->PopupMenu(&menu) && selectedEffectIndex != -1)
+         return effects[selectedEffectIndex]->GetID();
```
Paul-Licameli, line 501: Can the return type of this function be Identifier then, not naked wxString?

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+                     const wxPoint& pos = wxDefaultPosition,
+                     const wxSize& size = wxDefaultSize,
+                     long style = wxScrolledWindowStyle,
+                     const wxString& name = wxPanelNameStr)
+      : wxScrolledWindow(parent, winid, pos, size, style, name)
+   {
+      #ifdef __WXMSW__
+      //Fixes flickering on redraw
+      SetDoubleBuffered(true);
+   #endif
```
Paul-Licameli, line 525: Indent

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+               {
+                  mTargetIndex = targetIndex;
+                  PostDragEvent(parent, EVT_MOVABLE_CONTROL_DRAG_POSITION);
+               }
+            }
+         }
+      }
+   };
+
+   wxDEFINE_EVENT(EVT_EFFECT_ITEM_REMOVE, wxCommandEvent);
```
Paul-Licameli, line 265: An alternative possibility:  Don't define this event.  Don't post it from RealtimeEffectControl.

Instead, make the RealtimeEffectList s `Publisher` of messages when its item list change -- analogous to what TrackList does.  (Don't delay events with CallAfter as TrackList does for legacy reasons.  Let listeners CallAfter to delay their reactions as they need.)

Then the RealtimeEffectsWindow subscribes.

This way of doing it is more consistent with existing designs and would allow other subscribers if we ever need them, which won't need to cooperate specially with RealtimeEffectControl.

Paul-Licameli: I like this solution better for the possible future expansions.  But you can do just enough for preset needs:  let the list publish only deletions.  We might generalize for additions and permutations later.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+
+      SetSizer(rootSizer.release());
+
+      mEffectListContainer->Bind(EVT_EFFECT_ITEM_REMOVE, [this](wxCommandEvent&)
+      {
+         wxWindowUpdateLocker freeze(this);
+         Layout();
+      });
+
+      Bind(EVT_MOVABLE_CONTROL_DRAG_STARTED, [line](const MovableControlEvent& event)
```
Paul-Licameli, line 556: For these other event types, closer coupling of this window and the items is proper.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+
+   void OnAddEffectClicked(const wxCommandEvent& event)
+   {
+      auto effectID = ShowSelectEffectMenu(dynamic_cast<wxWindow*>(event.GetEventObject()));
+      if(!effectID.empty())
+      {
+         auto& effects = RealtimeEffectList::Get(*mTrack);
+         if(auto effectState = effects.AddState(effectID))
+         {
+            assert(effects.GetStatesCount() > 0);//If effect added, then there should be at least one effect on the stack
```
Paul-Licameli, line 629: Good, now also state it as a postcondition comment for `AddState`

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+   };
+
+   wxDEFINE_EVENT(EVT_EFFECT_ITEM_REMOVE, wxCommandEvent);
+
+   //UI control that represents individual effect from the effect list
+   class RealtimeEffectControl : public MovableControl
+   {
+      std::shared_ptr<Track> mTrack;
+      //int mEffectIndex{-1};
+      RealtimeEffectState* mEffectState{nullptr};
```
Paul-Licameli, line 272: I see now that there is a serious problem in this design.

There should be undo and redo of changes to the effect list.  This is not yet part of this pull request.  I anticipated the need at commit 5da76c6eae0aceb1823be43bcb9009aa20ead755 with some support.

If this window persists through pushes and pops of the undo stack, then storing this pointer is the wrong thing.  It might point into a version of the state that is in undo history but not the present project state, and it might become a dangling pointer if redo states are destroyed after an undo and then other edits.

Should the parent window just rebuild itself entirely whenever the undo stack is pushed or popped?  Maybe that's easiest to do.  Just subscribe to events from UndoManager.

In fact, you don't have a choice.  If you can't predict at all how the composition of the RealtimeEffectList changes with undo or redo, you just have to read it all and rebuild.

Okay, that eliminates the dangling pointer danger.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+      {
+         mTrack = track;
+
+         auto& effects = RealtimeEffectList::Get(*track);
+         assert (effectIndex >= 0 && effectIndex < effects.GetStatesCount());
+
+         mEffectState = effects.GetStateAt(effectIndex);
+         auto desc = PluginManager::Get().GetPlugin(mEffectState->GetID());
+
+         mChangeButton->SetLabel(desc->GetSymbol().Translation());
```
Paul-Licameli, line 347: I see uses of `Translation` here and elsewhere when building buttons.  That means one of these windows should be a `PrefsListener` and recreate those labels, in case the user changes the language in Interface preferences.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+      }
+
+      void RemoveFromList()
+      {
+         if(mEffectState != nullptr)
+         {
+            auto& effects = RealtimeEffectList::Get(*mTrack);
+            effects.RemoveState(*mEffectState);
```
Paul-Licameli, line 355: This side-effect on the effect list should cause a push of the Undo history.  See my previous comment mentioning Undo/Redo matters.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+         }
+      });
+      Bind(EVT_MOVABLE_CONTROL_DRAG_FINISHED, [this, line](const MovableControlEvent& event)
+      {
+         auto& effectList = RealtimeEffectList::Get(*mTrack);
+         if(event.GetSourceIndex() != event.GetTargetIndex())
+         {
+            assert(event.GetSourceIndex() >= 0 && event.GetSourceIndex() < effectList.GetStatesCount());
+            assert(event.GetTargetIndex() >= 0 && event.GetTargetIndex() <= effectList.GetStatesCount());
+            effectList.Reorder(event.GetSourceIndex(), event.GetTargetIndex());
```
Paul-Licameli, line 593: This too needs an Undo push.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+   }
+
+   void OnAddEffectClicked(const wxCommandEvent& event)
+   {
+      auto effectID = ShowSelectEffectMenu(dynamic_cast<wxWindow*>(event.GetEventObject()));
+      if(!effectID.empty())
+      {
+         auto& effects = RealtimeEffectList::Get(*mTrack);
+         if(auto effectState = effects.AddState(effectID))
+         {
```
Paul-Licameli, line 628: And here, another Undo push.

src/widgets/RealtimeEffectPanel.cpp
Paul-Licameli: What I said there:  the list window needs to subscribe to the UndoManager so it can rebuild after undo and redo commands that might change the current list of effects, to avoid pointers into the wrong version of the RealtimeEffectList.

Paul-Licameli (changes requested):
Window needs rebuilding for change of language or undo and redo; changes of the effect list need to push new Undo history states.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+   auto header = safenew ThemedWindowWrapper<wxWindow>(this, wxID_ANY);
+   header->SetBackgroundColorIndex(clrMedium);
+   {
+      auto hSizer = std::make_unique<wxBoxSizer>(wxHORIZONTAL);
+      auto toggleEffects = safenew ThemedButtonWrapper<wxToggleButton>(header, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize, wxNO_BORDER);
+      toggleEffects->SetBitmapLabelIndex(bmpEffectOff);
+      toggleEffects->SetBitmapPressedIndex(bmpEffectOn);
+      toggleEffects->SetBackgroundColorIndex(clrMedium);
+      toggleEffects->SetMinSize(toggleEffects->GetBitmapLabel().GetSize());
+      mToggleEffects = toggleEffects;
```
Paul-Licameli, line 663: I see no callback bound to this toggle button.  Maybe bind an empty lambda with one more explicit TODO comment, like other TODOs in this commit.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,711 @@
...
+         access->ModifySettings([&](EffectSettings& settings)
+         {
+            auto project = mTrack->GetOwner()->GetOwner();
+            auto &trackFactory = WaveTrackFactory::Get( *project );
+            auto &selectedRegion = ViewInfo::Get( *project ).selectedRegion;
+            auto rate = ProjectRate::Get(*project).GetRate();
+            auto &window = ProjectWindow::Get( *project );
+            auto flags = static_cast<unsigned>(EffectManager::kNone);
+
+            effectPlugin->DoEffect(settings,
```
Paul-Licameli, line 381: Still more Undo stack pushes are needed, for changes of settings in an effect stack.

But this doesn't work -- as you may have known from testing.  I think it was something you needed to discuss with me, but didn't find the opportunity?

It doesn't work because for realtime capable effects, a non-modal dialog is opened, and ModifySettings exits before the dialog closes.

Maybe then, the individual effect has other ModifySettings calls in its event handlers, so it sort-of works.  But not as you might expect from reading this code, and it won't interact correctly with undo and redo.

How to fix this then?  See the example of `EffectManager::PromptUser` and `MacroCommands::PromptForParamsFor` which calls it, while inside this scope object's lifetime:

```
   auto cleanup = EffectManager::Get().SetBatchProcessing(ID);
```

That forces effect dialogs to open modally, even for realtime effects, if you are editing a macro, which is how you can step through those functions and understand.

But they change the global EffectSettings for the effect, which is of course not the right thing here.

Paul-Licameli (changes requested):
One more Undo history push is needed, and we also need to force the realtime effect dialogs to be modal

src/widgets/RealtimeEffectPanel.cpp
Paul-Licameli: Another question to ask is whether all these things requiring pushes and pops of the undo stack are safe to do while playback is in progress.

I think we surely want that -- one hope is that you listen as you change settings of a single effect, or even change the composition of the effect list.  But now I doubt whether all is set up correctly in the audio engine for that, both to avoid dangling pointers and crashes, and to communicate changes of the sequence of effects properly inter-thread.

The limited thing that I know was anticipated by Leland's work, is the powering on and off of individual effects, which is what the suspending and resuming of states is for.

And the point of the EffectSettingsAccess class I developed is to communicate changes of settings of an individual effect correctly -- what this piece of code here will do.

But adding, deleting, and permuting effects?  Or changing the whole stack at one with undo and redo?

I must think.

src/ProjectWindow.cpp
```diff
@@ -1882,9 +1893,25 @@ void ProjectWindow::DoZoomFit()
...
+   wxWindowUpdateLocker freeze(this);
+
+   mEffectsTrackSelectionChanged = 
+      track->GetOwner()->Subscribe([this](const TrackListEvent &event)
+      {
+         if(event.mType == TrackListEvent::SELECTION_CHANGE)
+         {
+            auto track = event.mpTrack.lock();
+            mEffectsWindow->SetTrack(track && track->IsSelected() ? track : nullptr);
+         }
```
Paul-Licameli, line 1910: I see what this subscription is doing and I understand why you need something like it.

I question whether this is exactly what we will want.  First, sometimes multiple tracks are selected, as when you select all.  This might, wastefully, visit this function once for each track, rebuilding the window many times, and finally leaving it as for the last selected track.

Would it be better to scan the track list for a unique selected track -- and empty the effect list if there is no such track?

But then -- as I pointed out long ago, critiquing the design proposals -- there is a distinction between selection, which is non-exclusive, and focus (indicated by the yellow border), which is.

I think it makes more sense to show the effect stack for the focused track, which means you should listen for what is now the `EVT_TRACK_FOCUS_CHANGE` event (but, like theme change, ANOTHER thing I want to rewrite with Observer, and I also have a waiting commit for that, which I can cherry-pick and push onto your branch).

@DavidBailes will surely want to test this new feature and be satisfied with keystroke navigation, and again, the notion of a "focused" object, rather than a "selected" one, is in general relevant to the targeting of commands given by keystroke.  See for instance `OnTrackMenu` in `TrackMenus.cpp` which handles Shift+M in default key bindings.  That keyboard shortcut opens the track drop-down menu for the (unique) *focused*, not selected, track.

Paul-Licameli:
THIS COMPLETES ROUND ONE OF REVIEW!

Paul-Licameli:
I have pushed the two promised commits onto this branch.  If you rebase and force-push, perhaps they could be moved to the start of the commit sequence so you don't add and old style event handling code for the theme, only to change it again.

src/ProjectWindow.cpp
DavidBailes: > @DavidBailes will surely want to test this new feature and be satisfied with keystroke navigation

Is there a current design specification for the effects panel which includes the keyboard navigation? I have asked this question on https://github.com/audacity/audacity/issues/2845, but have not yet received a reply.

Concerning whether the effects panel show the effects stack for the focused or selected track. The normal pattern is for a side panel to show some property of a selected item in a list. However, I don't think it fits the current case. A user might want to look at the effect stack without changing which track or track was selected, and each audio track having its own effect button implies this. Note however, that when a user presses an effect button, then I presume the intention is that the focus is automatically moved to the effects panel. This means that the track is not the focus and won't have a yellow outline. Are the designers happy with that?

Concerning the list of effects in the Effects panel. When the figma designs of this were originally shared by Jouni, the keyboard navigation of the list of effects, including the controls in each item, was specified as just a series of tabs. I commented on discord at the time that it would be much better if the keyboard navigation was appropriate for a list. Unfortunately, nobody replied to that comment. It is still my view that the list of effects should have keyboard navigation which is appropriate for a list where each list item contains a number of controls, and in addition, the list is read as being a list by users of screen readers. Given that these requirements have implications for how the code for the list is structured, first writing it so that works only with mouse interaction, and then ensuring that it works properly with keyboard interaction and screen readers will end up involving a lot more work than if it was written with all types of interaction from the start.

Paul-Licameli: Also possible:  the sidebar changes only if you press Effects on another track.  (Or delete that track)

The Effects button of that track only is painted as down.

Paul-Licameli:
A third suggested commit added to the branch

src/effects/RealtimeEffectList.cpp
vsverchinsky: Added a commetary and replaced pointer with reference type

src/effects/RealtimeEffectList.cpp
vsverchinsky: I'll skip this, as you've already done that

src/widgets/RealtimeEffectPanel.cpp
vsverchinsky: There is no link yet)

src/widgets/RealtimeEffectPanel.cpp
vsverchinsky: It is not much sense in doing so: `MovableControl` is a wxWidgets based component

Paul-Licameli: I do not understand your response.  I think you did not understand my objection.

Changes in the RealtimeEffectList — a UI neutral internal entity— should be published with our UI neutral Publisher class, as is also done for changes of the TrackList.  Then this new UI works more indirectly:  one drag and drop changes the RealtimrEffectList object, then it publishes its message, then the other control reacts.  But then it is also possible for other things that need to be kept consistent with the real-time effect list to react also.

I do not want to use Qt signals and slots outside of a UI layer of code built on a toolkit neutral layer.

src/effects/RealtimeEffectList.h
```diff
@@ -54,12 +74,25 @@ class RealtimeEffectList final
    //! Apply the function to all states sequentially.
    void Visit(StateVisitor func);
 
-   //! Returns null if the id is nonempty but no such effect was found
+   //! Returns null if no such effect was found.
+   //! Sends message Insert message on success.
```
Paul-Licameli, line 78: You repeated "message"

src/effects/RealtimeEffectList.h
```diff
@@ -54,12 +74,25 @@ class RealtimeEffectList final
...
    RealtimeEffectState *AddState(const PluginID &id);
+   //! On success sends Remove message.
    void RemoveState(RealtimeEffectState &state);
-   void Swap(size_t index1, size_t index2);
 
-   using States = std::vector<std::unique_ptr<RealtimeEffectState>>;
+   //! Returns total amount of effects in this list
```
Paul-Licameli, line 83: "number" would be a better word than "amount" (for a count-noun, not a mass-noun)

src/effects/RealtimeEffectList.cpp
```diff
@@ -85,14 +85,20 @@ void RealtimeEffectList::Visit(StateVisitor func)
 RealtimeEffectState *RealtimeEffectList::AddState(const PluginID &id)
 {
    auto pState = std::make_unique<RealtimeEffectState>(id);
-   if (id.empty() || pState->GetEffect() != nullptr) {
+   if (pState->GetEffect() != nullptr) {
```
Paul-Licameli, line 88: You lost the conjunct `id.empty()` but it is important not to return nullptr in that case.  It will happen when reloading a `RealtimeEffectList` from a file.  The state is constructed first, then the id is read from the file and assigned to it.

Paul-Licameli (changes requested):

src/effects/RealtimeEffectList.h
```diff
@@ -9,18 +9,34 @@
...
+   enum class Type
+   {
+      Insert,///<New effect item was added to the list
+      Remove,///<Effect item was removed from the list
+      Move ///<Item position has changed, new position will be stored as size_t value in extra
+   };
+   Type type;
+   size_t index;
+
+   std::any extra;///< depends on message type
```
Paul-Licameli, line 37: Maybe `std::variant` is the more proper thing when the alternatives are not open-ended but instead all enumerated here :shrug:  To my mind, `std::any` is more appropriate where there is open-endedness (if, say, RealtimeEffectListMessage was the root of a class hierarchy)

Or simply make it a numerical field that is ignored for the other message types.

src/widgets/ThemedWrappers.h
```diff
@@ -0,0 +1,141 @@
...
+      if (message.appearance)
+      {
+         if(mBackgroundColorIndex != -1)
+            WindowBase::SetBackgroundColour(theTheme.Colour(mBackgroundColorIndex));
+         if(mForegroundColorIndex != -1)
+            WindowBase::SetForegroundColour(theTheme.Colour(mForegroundColorIndex));
+         WindowBase::Refresh();
+      }
+      if(mTranslatableLabel)
+         WindowBase::SetLabel(mTranslatableLabel->Translation());
```
Paul-Licameli, line 70: You need to update the text of labels and buttons for changes of language -- and deriving from `PrefsListener` is how that is done.

It is not sufficient to update the labels when the theme changes -- and I don't understand why it is necessary then either, but maybe I don't know.

Ah, but is it that the dismissing of the Preferences dialog coincidentally sends theme change and PrefsListener messages together, always?

So maybe this appears to work when you test change of language, but still I don't think it is done properly.  And who knows, we might invent a different UI for language change some day that doesn't have this coincidence.

Paul-Licameli (changes requested):

src/widgets/ThemedWrappers.h
Paul-Licameli: See what was done in commit 20ace31

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,813 @@
...
+         const auto movedItem = sizer->GetItem(msg.index);
+
+         const auto proportion = movedItem->GetProportion();
+         const auto flag = movedItem->GetFlag();
+         const auto border = movedItem->GetBorder();
+
+         auto window = movedItem->GetWindow();
+         
+         sizer->Remove(msg.index);
+         sizer->Insert(std::any_cast<size_t>(msg.extra), window, proportion, flag, border);
```
Paul-Licameli, line 637: This breaks the macOS build.  Only the pointer-valued overload of `any_cast` is available, with our minimum targeted version.

src/widgets/ThemedWrappers.h
vsverchinsky: >Ah, but is it that the dismissing of the Preferences dialog coincidentally sends theme change and PrefsListener messages together, always?

I didn't think it was a coincidence, as I understand making a change to a single parameter in a settings group ("interface" or "effects") updates values for all parameters in that group

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,813 @@
...
+
+         auto& effectList = RealtimeEffectList::Get(*mTrack);
+         auto from = event.GetSourceIndex();
+         auto to = event.GetSourceIndex() < event.GetTargetIndex() ? event.GetTargetIndex() - 1 : event.GetTargetIndex();
+         if(from != to)
+         {
+            effectList.MoveEffect(from, to);
+            ProjectHistory::Get(*project).PushState(
+               //i18n-hint: undo history, first parameter - track name
+               XO("'%s' effects reorder %d->%d").Format(mTrack->GetName(), from, to),
```
Paul-Licameli, line 608: i18n-hint comment does not match the string which has three format directives.

Paul-Licameli: Should you add 1 to from and to?  Users may understand 1-based indices better.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,813 @@
...
+               //i18n-hint: undo history record
+               XO("Realtime effect reorder"), UndoPush::CONSOLIDATE);
+         }
+         else
+         {
+            wxWindowUpdateLocker freeze(this);
+            Layout();
+         }
+      });
+      SetScrollRate(0, 20);
```
Paul-Licameli, line 618: This line isn't properly part of the commit for undo history so it might be put in an earlier commit

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,813 @@
...
+         return;
+      
+      if(auto state = AudioIO::Get()->AddState(*project, &*mTrack, effectID))
+      {
+         auto effect = state->GetEffect();
+         ProjectHistory::Get( *project ).PushState(
+            //i18n-hint: undo history, first parameter - realtime effect name, second - track name
+            XO("'%s' added to the '%s' effect stack").Format(effect->GetName(), mTrack->GetName()),
+            //i18n-hint: undo history record
+            XO("Add realtime effect"));
```
Paul-Licameli, line 723: The changes for undo and redo look good, though incomplete.  Replacement of one effect with another is not yet implemented and will also push an undo state.  Changing on/off states will use `ModifyState` not `PushState`.

Changing settings after clicking OK the settings dialog should also push an undo state.

The details of the short and long user visible messages for undo and redo may disagree with the specifictaion, but maybe the specification will be corrected.

None of what I just wrote here will block this pull request.

src/widgets/RealtimeEffectPanel.h
```diff
@@ -0,0 +1,48 @@
+/*!********************************************************************
```
Paul-Licameli, line 1: This pair of files is still in `src/widgets` and I would rather just put it under `src` pending other folder reorganizations later.

src/ProjectWindow.cpp
```diff
@@ -1880,11 +1893,32 @@ void ProjectWindow::DoZoomFit()
...
 {
+   mEffectsTrackSelectionChanged.Reset();
+
    if(track == nullptr)
       return;
 
+   wxWindowUpdateLocker freeze(this);
+
+   mEffectsTrackSelectionChanged = 
+      track->GetOwner()->Subscribe([this](const TrackListEvent &event)
```
Paul-Licameli, line 1906: This still subscribes to selection change efents, but we have agreed that is should instead be focus changes.

Also I suggest that the effects window might subscribe itself rather than relying on `ProjectWindow` to make the changes.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,813 @@
...
+         dropHintLine->Show();
+         dropHintLine->Raise();
+         //Make it zero-sized until first EVT_MOVABLE_CONTROL_DRAG_POSITION is triggered
+         dropHintLine->SetSize({0, 0});
+         if(auto window = dynamic_cast<wxWindow*>(event.GetEventObject()))
+            window->Raise();
+      });
+      Bind(EVT_MOVABLE_CONTROL_DRAG_POSITION, [this, dropHintLine](const MovableControlEvent& event)
+      {
+         constexpr auto DropHintLineHeihgh { 3 };//px
```
Paul-Licameli, line 566: "DropHintLineHeight"

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,813 @@
...
+               access);
+         });
+      }
+
+      void OnChangeButtonClicked(wxCommandEvent& event)
+      {
+         if(mEffectState == nullptr)
+            return;//not initialized
+
+         ShowSelectEffectMenu(mChangeButton, this);
```
Paul-Licameli, line 398: I see that this call pops a menu and returns a plugin id which is ignored.  So it's still to do:  substitute one effect for another in the list, and also push an undo item.

This PR might be merged with this known incompleteness.  (Should have a TODO comment)

src/ProjectWindow.cpp
Paul-Licameli: When the button is pressed to change the track whose effect list is shown -- then the track focus should also be changed to that track.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,813 @@
...
+            InsertEffectRow(i, effects.GetStateAt(i));
+      }
+      mAddEffect->Enable(!!mTrack);
+      Layout();
+   }
+
+   void OnUndoRedoMessage(const UndoRedoMessage& msg)
+   {
+      if(msg.type == UndoRedoMessage::UndoOrRedo || msg.type == UndoRedoMessage::Reset)
+         ReloadEffectsList();
```
Paul-Licameli, line 699: I tested this, and Undo/Redo did not have the desired effect, and I figured out why.

`mTrack` is a shared pointer, so there isn't a danger of a dangling pointer.  However, it must be reassigned when the undo state is pushed or popped, so that it points at a track in the current version of the track list.

See `Track::GetId()` for an identifier of a track across undo/redo states.  This window needs to store that ID, then find that ID again with each undo and redo.

It may also be necessary to `Refresh()` after `SetTrack()`.

Paul-Licameli (changes requested):

Paul-Licameli:
In fact `TrackList::FindById` is the convenient thing to use on undo or redo.  Use `Track::GetId` to find the value to store in the window.

src/widgets/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,814 @@
...
+
+      mEffectListItemMovedSubscription.Reset();
+      mUndoManagerSubscription.Reset();
+      
+      mTrack = track;
+      ReloadEffectsList();
+
+      if(track)
+      {
+         auto& effects = RealtimeEffectList::Get(*mTrack);
```
Paul-Licameli, line 676: Here and in `ReloadEffectList`, please check that the track is a WaveTrack before getting the `RealtimeEffectList`, just to avoid creation-on-demand of a useless, empty list in case the track is of non-wave type.

src/ProjectWindow.cpp
vsverchinsky: Hm, seems like I've lost part of my changes during commit history rewrite

src/widgets/RealtimeEffectPanel.cpp
vsverchinsky: Thanks!

vsverchinsky: But turns out this is not neccessary. `ProjectWindow` gets notification from `TrackFocus` about changed focus, and given that focus now points to a new track instance `RealtimeEffectPanel` updates anyway

src/widgets/RealtimeEffectPanel.cpp
vsverchinsky: Shouldn't be the case as if we do not show "effects" button for other types of tracks

Paul-Licameli: But changing track focus using the up and down arrow keys is another way to change the contents of the sidebar.

vsverchinsky: Didn't think of that, thanks

src/ThemedWrappers.h
```diff
@@ -0,0 +1,154 @@
...
+   void SetTranslatableLabel(TranslatableString label)
+   {
+      mTranslatableLabel = std::move(label);
+      WindowBase::SetLabel(mTranslatableLabel->Translation());
+   }
+
+   void UpdatePrefs() override
+   {
+      if(mTranslatableLabel)
+         WindowBase::SetLabel(mTranslatableLabel->Translation());
```
Paul-Licameli, line 64: Should it also call through to the base class’s UpdatePrefs?

I think yes in general, but it might call through to a pure virtual.  Which might be given a default do nothing implementation.  (Did you know that a pure virtual can have a definition but also remain pure?  Yes it can, so subclasses must define it to be constructible, while qualified calls to the function are also allowed.)

src/ThemedWrappers.h
```diff
@@ -0,0 +1,154 @@
...
+   void SetTranslatableLabel(TranslatableString label)
+   {
+      mTranslatableLabel = std::move(label);
+      ButtonBase::SetLabel(mTranslatableLabel->Translation());
+   }
+
+   void UpdatePrefs() override
+   {
+      if(mTranslatableLabel)
+         ButtonBase::SetLabel(mTranslatableLabel->Translation());
```
Paul-Licameli, line 134: Ditto.

Preference changes might be written to send messages like theme changes, but it was implemented this way with virtual functions before Observer existed.  That rewrite would eliminate the problem of missed call throughs but would be the bigger work now.

src/RealtimeEffectPanel.h
```diff
@@ -0,0 +1,48 @@
...
+
+class wxToggleButton;
+class wxStaticText;
+
+class RealtimeEffectListWindow;
+class AudacityProject;
+
+/**
+ * \brief UI Panel that displays realtime effects from the effect stack of
+ * individual track, provides controls for accessing effect settings,
```
Paul-Licameli, line 26: "an individual"

src/TrackPanel.cpp
```diff
@@ -312,14 +314,24 @@ TrackPanel::TrackPanel(wxWindow * parent, wxWindowID id,
...
       .Subscribe(*this, &TrackPanel::OnUndoReset);
 
    mAudioIOSubscription =
       AudioIO::Get()->Subscribe(*this, &TrackPanel::OnAudioIO);
+
+   mRealtimeEffectManagerSubscription = RealtimeEffectManager::Get(*theProject)
+      .Subscribe([this](const RealtimeEffectManagerMessage& msg)
+      {
+         if(!msg.track)
+            Refresh();
```
Paul-Licameli, line 330: I understand you need to repaint a track if effects are inserted or removed in its stack, because the button may need to change its appearance when the effect list becomes empty or becomes nonempty.

I do not see why you need to refresh the whole panel when there are changes in the global list of effects.

No harm would be done, but a needless repainting.

src/ThemedWrappers.h
```diff
@@ -0,0 +1,154 @@
+/*!********************************************************************
```
Paul-Licameli, line 1: You also took this header file out of `src/widgets`

Actually I don't mind this file being in that folder, because it is more general-purpose and reusable.

I only wanted `RealtimeEffectPanel.h` to move out of it.

But this file can stay in `src` too

src/ProjectWindow.cpp
```diff
@@ -677,7 +690,19 @@ ProjectWindow::ProjectWindow(wxWindow * parent, wxWindowID id,
...
+   mThemeChangeSubscription =
+      theTheme.Subscribe(*this, &ProjectWindow::OnThemeChange);
+
+   mFocusChangeSubscription = TrackFocus::Get(project)
+      .Subscribe([this](const TrackFocusChangeMessage& msg) {
+         if(mEffectsWindow->IsShown())
+         {
+            auto& project = GetProject();
+            auto& trackFocus = TrackFocus::Get(project);
+            ShowEffectsPanel(project, trackFocus.Get());
```
Paul-Licameli, line 702: Will that line un-collapse the effects panel if it was hidden, whenever I use the up and down arrow keys?  Maybe not desirable if so.

src/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,796 @@
...
+         dropHintLine->Show();
+         dropHintLine->Raise();
+         //Make it zero-sized until first EVT_MOVABLE_CONTROL_DRAG_POSITION is triggered
+         dropHintLine->SetSize({0, 0});
+         if(auto window = dynamic_cast<wxWindow*>(event.GetEventObject()))
+            window->Raise();
+      });
+      Bind(EVT_MOVABLE_CONTROL_DRAG_POSITION, [this, dropHintLine](const MovableControlEvent& event)
+      {
+         constexpr auto DropHintLineHeihgt { 3 };//px
```
Paul-Licameli, line 568: Still misspelled

src/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,796 @@
...
+      {
+         auto& effects = RealtimeEffectList::Get(*mTrack);
+         mEffectListItemMovedSubscription = effects.Subscribe(*this, &RealtimeEffectListWindow::OnEffectListItemChange);
+      }
+   }
+
+   void ReloadEffectsList()
+   {
+      wxWindowUpdateLocker freeze(this);
+      //delete items that was added to the sizer
```
Paul-Licameli, line 677: were

Paul-Licameli:
If the responses to undo and redo appear to work now, without any use of `TrackId` or any subscription to undo manager changes, it is because you rely on coindicental track focus change messages -- those send from EditMenus.cpp, right?

I fell a bit uncomfortable with that.  It seems a bit fragile, and it might be preferable that this panel should express its dependency on the current undo state directly, even if it requires a little more work.

You are depending on things coded in EditMenus.cpp for this window to work correctly, but the src/menus files should be tops of the dependency graph -- they depend on other things, but nothing depends on them.  This isn't "dependency" in the really hard sense of link time, but dependency in a weaker sense.

I might not insist on this fix to get this merged and testable.  But I'm not wholly satisfied without it.

src/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,796 @@
...
+      }
+
+      void RemoveFromList()
+      {
+         if(mEffectState == nullptr)
+            return;
+         
+         auto effectName = mEffectState->GetEffect()->GetName();
+         AudioIO::Get()->RemoveState(*mProject, &*mTrack, *mEffectState);
+         ProjectHistory::Get(*mProject).PushState(
```
Paul-Licameli, line 349: I don't understand why you changed mProject to a `wxWeakRef` but still assume it is non-null anyway.

Is there any reason to think the project will ever be destroyed before the associated effects window?  Did your experiments suggest that is so?

See `ProjectManager::OnCloseWindow` where there is some care about the sequence of destruction of things.  But note the ProjectWindow is destroyed there (which should destroy its descendants), just before the last line which destroys the Project object.

So I think there is no need at all for wxWeakRef for safety.

src/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,796 @@
...
+         if(mEffectState == nullptr)
+            return;//not initialized
+         
+         const auto ID = mEffectState->GetID();
+         const auto effectPlugin = EffectManager::Get().GetEffect(ID);
+         if(effectPlugin == nullptr)
+         {
+            ///TODO: effect is not avaialble
+            ///should it be removed from the list on interaction or
+            ///during project load/plugin manager list modification?
```
Paul-Licameli, line 368: See `RealtimeEffectList::HandleXMLEndTag` for some answer to your question.  Yes, there is a check on success of `GetEffect` at the time the project is reopened, which we can assume, I think, follows all plugin initializations at startup.  And I think there is then no way to insert an effect from a plug-in that does not exist.

But can a plug-in ever be un-loaded while the session continues?  I don't think that ever happens.  If it did, perhaps then more uses of Observer and subscriptions might let the RealtimeEffectLists update themselves, and they would
 further propagate `Publish`-ed messages to these windows.

src/RealtimeEffectPanel.cpp
```diff
@@ -0,0 +1,796 @@
...
+               }
+            }
+         }
+      }
+   };
+
+   //UI control that represents individual effect from the effect list
+   class RealtimeEffectControl : public MovableControl
+   {
+      wxWeakRef<AudacityProject> mProject;
```
Paul-Licameli, line 265: ditto

Paul-Licameli:
One more thing -- did you resolve the point, that a click on the "[Add] Effects" button should reset the track focus?  I don't see where that omission was fixed.

src/RealtimeEffectPanel.cpp
vsverchinsky: >I don't understand why you changed mProject to a wxWeakRef but still assume it is non-null anyway.

Right, it does not make sense

>See ProjectManager::OnCloseWindow where there is some care about the sequence of destruction of things.

`wxWindow::Destroy` does not destroy window immediately, but I'm not sure about exact order of things happening after that.

Paul-Licameli: "wxWidgets delays actual deletion of the window until all events have been processed, since otherwise there is the danger that events will be sent to a non-existent window."

https://docs.wxwidgets.org/3.0/overview_windowdeletion.html

So maybe your precaution does make sense in event handlers, but it was not fully realized without adding the null checks

src/RealtimeEffectPanel.cpp
vsverchinsky: >See RealtimeEffectList::HandleXMLEndTag for some answer to your question

But it happens silently, without any warnings, isn't it?

>And I think there is then no way to insert an effect from a plug-in that does not exist.

If there was a warning, user could have decided to close the project, install plugin and reopen this project again. Also, it's not necessary to remove `RealtimeEffectState` if plugin which it refers to is missing. There might be something that indicates a faulty plugin in the stack. But this is more UI/UX question I guess.

Paul-Licameli (approved):
All concerns I raised are either addressed, or are remaining problems for me.

Let's merge it!

Issue #2861, opened 2022-04-27
Audacity hangs when monitoring is stopped


crsib:
**Describe the bug**

Audacity hangs when monitoring is stopped, or when record or play is pressed while monitoring

**To Reproduce bug #1 - at monitoring stop**
1. Start Audacity
2. Click on Input Meter Panel to start monitoring
3. Click on Input Meter Panel to stop monitoring

**To Reproduce bug #2 - at record pressed**
1. Start Audacity
2. Click on Input Meter Panel to start monitoring
3. Click on record

**To Reproduce bug #3 - at play pressed**
1. Start Audacity
2. have a track ready to play something
3. Click on Input Meter Panel to start monitoring
4. Click on play

**Actual behavior**
Audacity hangs

**Expected behavior**
Monitoring is stopped + Audacity does not hang

**Screenshots**
If applicable, add screenshots/video to help explain your problem.

**Additional information (please complete the following information):**
Happens in `master` branch

**Additional context**
Main thread waits for audio thread to stop in `AudioIoCallback::WaitForAudioThreadStopped()`

pietro68:
This happens since commit `de00e9f` done by me, so I will take on the issue.

petersampsonaudacity:
Confirmed on W10 with latest alpha master build.

This is a regression on 3.1.3

petersampsonaudacity:
Testing on W10 with @pietro68 latest branch audacity-win-3.2.0-alpha-20220428+8747f0d-x64-msvc2022 

This now no longer hangs Audacity with this branch build

petersampsonaudacity:
Testing on W10 with latest alpha master audacity-win-3.2.0-alpha-20220503+ba1664f-x64-msvc2022

I confirm this is fixed on Windows

dozzzzer:
Tested on Win10 and audacity-win-3.2.0-alpha-20220502+b3c2a0d-x64

PR #2868, opened 2022-04-28
Audacity hangs when turning monitoring off

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2861 

The root of the problem is: 
1) turning monitoring on does NOT call `::StartStream`, but 
2) turning monitoring off does call `::StopStream`

(1) means: when monitoring is on, the else branch in `AudioThread::Entry` loop is executed - so the transition of `lastState`can never happen, and when `::StopStream` is called in response to stop monitoring, it will wait forever for the transition.

I understand that the solution I propose, might not be the cleanest one; I think that to have a real clean solution, we'd have to reorganize the code in AudioIO a bit better - for instance, we might want to have a branch in the AudioThread::Entry loop, for the case when monitoring is on.

Recommended:

### DIFF
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -937,6 +937,9 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
          gAudioIO->AILAInitialize();
       #endif
 
+      if (gAudioIO->IsMonitoring())
+         gAudioIO->StopMonitoring();
+
       int token = gAudioIO->StartStream(transportTracks, t0, t1, t1, options);
 
       success = (token != 0);
--- a/src/widgets/MeterPanel.cpp
+++ b/src/widgets/MeterPanel.cpp
@@ -1882,7 +1882,7 @@ void MeterPanel::StartMonitoring()
 
    auto gAudioIO = AudioIO::Get();
    if (gAudioIO->IsMonitoring()){
-      gAudioIO->StopStream();
+      gAudioIO->StopMonitoring();
    } 
 
    if (start && !gAudioIO->IsBusy()){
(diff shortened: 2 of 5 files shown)


petersampsonaudacity:
Testing on W10 with @pietro68 latest branch audacity-win-3.2.0-alpha-20220428+8747f0d-x64-msvc2022 

This now no longer hangs Audacity

Paul-Licameli:
Please update the description of the linked issue to include the other bug for pressing Record, so QA will verify that is fixed too.

src/ProjectAudioManager.cpp
```diff
@@ -937,6 +937,9 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
          gAudioIO->AILAInitialize();
       #endif
 
+      if (gAudioIO->IsMonitoring())
+         gAudioIO->StopMonitoring();
+
```
Paul-Licameli, line 942: Is there a similar problem still to fix if you start play while monitoring?

src/widgets/MeterPanel.cpp
```diff
@@ -1897,13 +1897,6 @@ void MeterPanel::StartMonitoring()
    }
 }
 
-void MeterPanel::StopMonitoring(){
-   mMonitoring = false;
-   auto gAudioIO = AudioIO::Get();
-   if (gAudioIO->IsMonitoring()){
-      gAudioIO->StopStream();
-   } 
-}
```
Paul-Licameli, line 1906: I think this deletion was a change for debugging purposes that you don’t want to merge.  The stopping level of monitoring with a click on the meter is disabled and not restored again.  Do I misunderstand something?

Paul-Licameli (changes requested):

src/widgets/MeterPanel.cpp
pietro68: I deleted this because it was not used and it is confusing. What gets called in response to either starting or stopping monitoring, is always MeterPanel::StartMonitoring.

src/ProjectAudioManager.cpp
pietro68: Good point! it happens also when starting to play, both in the latest master and on this branch at this point, so I will have to fix also that case. I am updating the issue description to include this case.

pietro68:
Apologies, but I realized that a much cleaner fix could be done, please find the PR for it at https://github.com/audacity/audacity/pull/2883

pietro68:
Closing this PR, because I could code a cleaner solution in PR https://github.com/audacity/audacity/pull/2883 which was approved and merged.

PR #2879, opened 2022-04-29
lib-ipc

vsverchinsky:

Recommended:

### DIFF
--- a/libraries/lib-ipc/IPCChannel.h
+++ b/libraries/lib-ipc/IPCChannel.h
@@ -1,0 +1,61 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file IPCChannel.h
+
+  @author Vitaly Sverchinsky
+
+  Part of lib-ipc library
+
+**********************************************************************/
+
+#pragma once
+
+#include <cstddef>
+
+/**
+ * \brief Interface for sending data from client to server or vice versa,
+ * complemented by IPCChannelStatusCallback
+ */
+class IPC_API IPCChannel
+{
+public:
+   virtual ~IPCChannel();
+   /**
+    * \brief Write data to the channel
+    * \param bytes Pointer to the data
+    * \param length Number of bytes to send
+    */
+   virtual void Send(const void* bytes, size_t length) = 0;
+};
+
+/**
+ * \brief Interface for listening connection status changes
+ */
+class IPC_API IPCChannelStatusCallback
+{
+public:
+   virtual ~IPCChannelStatusCallback();
+
+   /**
+    * \brief Called when connection attempt fails.
+    */
+   virtual void OnConnectionError() noexcept = 0;
+   /**
+    * \brief Called when connection established.
+    * \param channel Using this channel client or server can send data to the other side.
+    */
+   virtual void OnConnect(IPCChannel& channel) noexcept = 0;
+   /**
+    * \brief Invalidates IPCChannel passed as argument in OnConnect.
+    */
+   virtual void OnDisconnect() noexcept = 0;
+   /**
+    * \brief Called when data chunk received as a result of a preceding call to IPCChannel::Send.
+    * Generally, data pointer should not be accessed outside this method, copied if necessary.
+    * \param data Pointer to the chunk
+    * \param size Size of the chunk
+    */
+   virtual void OnDataAvailable(const void* data, size_t size) noexcept = 0;
+};
--- a/libraries/lib-ipc/IPCServer.cpp
+++ b/libraries/lib-ipc/IPCServer.cpp
@@ -1,0 +1,135 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file IPCServer.cpp
+
+  @author Vitaly Sverchinsky
+
+  Part of lib-ipc library
+
+**********************************************************************/
+
+#include "IPCServer.h"
+#include "IPCChannel.h"
+
+#include <thread>
+#include <mutex>
+#include <stdexcept>
+
+#include "internal/ipc-types.h"
+#include "internal/socket_guard.h"
+#include "internal/BufferedIPCChannel.h"
+
+class IPCServer::Impl
+{
+   bool mTryConnect{true};
+   std::mutex mSync;
+   std::unique_ptr<BufferedIPCChannel> mChannel;
+   std::unique_ptr<std::thread> mConnectionRoutine;
+
+   socket_guard mListenSocket;
+public:
+
+   Impl(IPCChannelStatusCallback& callback)
+   {
+      mListenSocket = socket_guard { socket(AF_INET, SOCK_STREAM, IPPROTO_TCP) };
+      if(!mListenSocket)
+         throw std::runtime_error("cannot create socket");
+      
+      sockaddr_in addrin {};
+      addrin.sin_family = AF_INET;
+      addrin.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+      addrin.sin_port = htons(static_cast<u_short>(IPC_TCP_CONNECTION_PORT));
+
+      if(bind(*mListenSocket, reinterpret_cast<const sockaddr*>(&addrin), sizeof(addrin)) == SOCKET_ERROR)
+         throw std::runtime_error("socket bind error");
+
+      if(listen(*mListenSocket, 1) == SOCKET_ERROR)
+         throw std::runtime_error("socket listen error");
+
+      mChannel = std::make_unique<BufferedIPCChannel>();
+      mConnectionRoutine = std::make_unique<std::thread>([this, &callback]
+      {
+         socket_guard connfd;
+         
+         while(true)
+         {
+            {
+               //combine flag check and internal state initialization within a single lock...
+
+               std::lock_guard lck {mSync};
+               if(!mTryConnect)
+                  return;
+
+               if(connfd)
+               {
+                  mListenSocket.reset();//do not need that any more
+                  try
+                  {
+                     mChannel->StartConversation(connfd.release(), callback);
+                  }
+                  catch(...)
+                  {
+                     callback.OnConnectionError();
+                  }
+                  //initialization finished, we can leave now
+                  break;
+               }
+            }
+
+            fd_set readfds, exceptfds;
+            FD_ZERO(&readfds);
+            FD_ZERO(&exceptfds);
+            FD_SET(*mListenSocket, &readfds);
+            FD_SET(*mListenSocket, &exceptfds);
+
+            auto ret = select(NFDS(*mListenSocket), &readfds, nullptr, &exceptfds, nullptr);
+            if(ret == 1)
+            {
+               connfd = socket_guard { accept(*mListenSocket, nullptr, nullptr) };
+               if(!connfd)
+               {
+                  callback.OnConnectionError();
+                  break;
+               }
+               //connection created, finish initialization during next loop iteration under guarded section
+            }
+            else//SOCKET_ERROR
+            {
+               callback.OnConnectionError();
+               break;
+            }
+         }
+      });
+
+   }
+
+   ~Impl()
+   {
+      {
+         std::lock_guard lck{mSync};
+         mTryConnect = false;
+         //this will also interrupt select in connection thread
+         mListenSocket.reset();
+         mChannel.reset();
+      }
+      if(mConnectionRoutine)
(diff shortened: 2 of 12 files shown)


libraries/lib-ipc/IPCChannel.h
```diff
@@ -0,0 +1,49 @@
...
+
+   /**
+    * \brief Called when connection attempt fails.
+    */
+   virtual void OnConnectionError() noexcept = 0;
+   /**
+    * \brief Called when connection established.
+    * \param channel Using this channel client or server can send data to the other side.
+    */
+   virtual void OnConnect(IPCChannel* channel) noexcept = 0;
```
Paul-Licameli, line 37: If I wrote this I would still prefer a reference argument to make clear it is never null

libraries/lib-ipc/internal/socket_guard.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
+
+#include <cassert>
+#include "ipc-types.h"
+
+class socket_guard
```
Paul-Licameli, line 6: A brief class comment a out the purpose?  But not hard to learn from reading it.

libraries/lib-ipc/internal/socket_guard.h
```diff
@@ -0,0 +1,63 @@
...
+   {
+      return mSocket != INVALID_SOCKET;
+   }
+
+   socket_guard& operator=(socket_guard&& rhs) noexcept
+   {
+      if(this == &rhs)
+         return * this;
+
+      assert(mSocket == INVALID_SOCKET || rhs.mSocket != mSocket);
```
Paul-Licameli, line 31: This asserts “the socket was uniquely guarded”, very informally.

I would make that a `@pre` comment.

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,57 @@
+#pragma once
+
+#include <mutex>
+#include <condition_variable>
```
Paul-Licameli, line 4: Hello, `condition_variable`.  This review is getting very interesting.

libraries/lib-ipc/internal/socket_guard.h
Paul-Licameli: Less urgent while this remains “internal”

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,57 @@
...
+{
+   static constexpr int DefaultOutputBufferCapacity  { 2048 };
+   static constexpr int DefaultInputBufferSize { 2048 };
+
+   bool mAlive {true};
+   std::mutex mSocketSync;
+   std::condition_variable mSendCondition;
+
+   std::unique_ptr<std::thread> mRecvRoutine;
+   std::unique_ptr<std::thread> mSendRoutine;
```
Paul-Licameli, line 28: `unique_ptr`s to Threads again... I still say it is an unnecessary indirection and you should test with `joinable` whether a thread is really there.

libraries/lib-ipc/internal/BufferedIPCChannel.h
```diff
@@ -0,0 +1,57 @@
...
+   /**
+    * \brief Thread-safe
+    */
+   void Send(const void* bytes, size_t length) override;
+
+   /**
+    * \brief Allowed to be called only once during object lifetime.
+    * Takes ownership over a socket. Callback should be guaranteed to be alive
+    * between IPCChannelStatusCallback::OnConnect and IPCChannelStatusCallback::OnDisconnect,
+    * and will be accessed from multiple threads.
```
Paul-Licameli, line 52: Very good statement of contract.

libraries/lib-ipc/IPCServer.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <memory>
+
+class IPCChannel;
+class IPCChannelStatusCallback;
+
+/**
+ * \brief Simple TCP socket based ipc server. When created
+ * server starts listen for incoming connection (see IPCClient).
```
Paul-Licameli, line 10: “Starts to listen”

libraries/lib-ipc/internal/BufferedIPCChannel.cpp
```diff
@@ -0,0 +1,158 @@
...
+BufferedIPCChannel::BufferedIPCChannel()
+{
+   mOutputBuffer.reserve(DefaultOutputBufferCapacity);
+}
+
+BufferedIPCChannel::~BufferedIPCChannel()
+{
+   if(mSocket != INVALID_SOCKET)
+      //interrupts any blocking calls
+      CLOSE_SOCKET(mSocket);
```
Paul-Licameli, line 18: Good.  Does the operating system also give us any guarantees when the process holding one side of the socket dies abnormally?  I’m sure it should, but please point me to a reference.  I think we rely on that, so Audacity does not hang if PluginHost chokes. — Have we actually tested with a corrupt plugin?  Now that I can build swh Ladspa for Putin’s fro lm source, I could easily modify the source to make bad plug-ins.

vsverchinsky: >Have we actually tested with a corrupt plugin?

I did test it, though on windows only.

>Does the operating system also give us any guarantees when the process holding one side of the socket dies abnormally?

For [posix-compliant](https://pubs.opengroup.org/onlinepubs/9699919799/functions/_Exit.html#tag_16_01) exit:
```
Process termination caused by any reason shall have the following consequences:
...
All of the file descriptors, directory streams, conversion descriptors, and message catalog descriptors open in the calling process shall be closed.
```

libraries/lib-ipc/IPCChannel.h
vsverchinsky: Presence of reference itself does not guarantee nothing.

Paul-Licameli: It communicates clearly that channel is never null which is in fact true in usage but not clear just from reading this header.  You took my other suggestions to change a pointer to a reference.

Classes should have clearly expressed contracts in header files and one way to make things clear to those with a good knowledge of C++ is to use types in understood conventional ways.

Paul-Licameli: Not having this clearly stated, every override of OnConnect instead has the responsibility to null-check the argument.

Paul-Licameli:
I will soon complete a review.  So far I have made only non-blocking suggestions.

libraries/lib-ipc/IPCChannel.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <cstddef>
+
+/**
+ * \brief Interface for sending data from client to server or vise versa,
```
Paul-Licameli, line 6: "vice versa"

libraries/lib-ipc/IPCServer.h
```diff
@@ -0,0 +1,30 @@
...
+{
+   class Impl;
+   std::unique_ptr<Impl> mImpl;
+public:
+   /**
+    * \brief May fail with exception or with call to
+    * IPCChannelStatusCallback::OnConnectionError.
+    * Callback should be guaranteed to be alive
+    * until either IPCChannelStatusCallback::OnDisconnect
+    * or IPCChannelStatusCallback::OnConnectionError is called.
```
Paul-Licameli, line 22: Good, this explains that an exception may be thrown.  The other pull request that uses IPCServer now needs try-catch blocks.

Not clear perhaps from this explanation is that a call to `OnConnectionError` happens on another thread that is created during a successful construction.

libraries/lib-ipc/IPCServer.cpp
```diff
@@ -0,0 +1,123 @@
...
+               }
+            }
+
+            fd_set readfds, exceptfds;
+            FD_ZERO(&readfds);
+            FD_ZERO(&exceptfds);
+            FD_SET(*mListenSocket, &readfds);
+            FD_SET(*mListenSocket, &exceptfds);
+
+            auto ret = select(NFDS(*mListenSocket), &readfds, nullptr, &exceptfds, 0);
```
Paul-Licameli, line 75: Now selecting with null for timeout which will block for availability.  Good, we want that.  But again I question what select adds in case there is only one file descriptor?  Maybe this can be simplified just by calling `accept` which according to the man page, blocks if the socket was not explicitly constructed as non-blocking.

libraries/lib-ipc/IPCServer.cpp
```diff
@@ -0,0 +1,123 @@
...
+      });
+
+   }
+
+   ~Impl()
+   {
+      {
+         std::lock_guard lck{mSync};
+         mTryConnect = false;
+         //this will also interrupt select in connection thread
```
Paul-Licameli, line 101: Would it interrupt `accept` too with the suggested rewrite?  But not important to do that.

libraries/lib-ipc/internal/BufferedIPCChannel.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+void BufferedIPCChannel::Send(const void* bytes, size_t length)
+{
+   std::lock_guard lck(mSocketSync);
+
+   auto offset = mOutputBuffer.size();
+   mOutputBuffer.resize(offset + length);
+   memcpy(mOutputBuffer.data() + offset, bytes, length);
+
+   mSendCondition.notify_one();
```
Paul-Licameli, line 34: There are two uses of `notify_one`.  It is explained here that the notifier does not need to lock the mutex and it might be slightly bad for performance if it does so. https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one

Paul-Licameli: In this case mutual exclusion is needed for other reasons of course.  Still the appending to the buffer must be excluded from the consuming of it.

libraries/lib-ipc/internal/BufferedIPCChannel.cpp
```diff
@@ -0,0 +1,158 @@
...
+      secondaryOutputBuffer.reserve(DefaultOutputBufferCapacity);
+      
+      while(true)
+      {
+         std::unique_lock lck(mSocketSync);
+         mSendCondition.wait(lck, [this]{ return !mAlive || !mOutputBuffer.empty(); });
+
+         if(!mAlive)
+         {
+            lck.unlock();
```
Paul-Licameli, line 58: Don't need that line

libraries/lib-ipc/internal/BufferedIPCChannel.cpp
```diff
@@ -0,0 +1,158 @@
...
+                     secondaryOutputBuffer.data() + offset,
+                     static_cast<int>(secondaryOutputBuffer.size()) - offset,
+                     0);
+                  if(ret > 0)
+                     offset += ret;
+                  else if (ret == SOCKET_ERROR)
+                     break;//error
+               }
+               else
+                  break;//error
```
Paul-Licameli, line 91: If either break is reached, there is no indication of the error to the user.   Should the callback be captured in this lambda too, should `OnDisconnect` be called?  Should the channel shut itself down?

vsverchinsky: `OnDisconnect` is called from receiving thread, there is no need to understand the reason of disconnection right now

libraries/lib-ipc/internal/BufferedIPCChannel.cpp
```diff
@@ -0,0 +1,158 @@
...
+               auto err = WSAGetLastError();
+               if(err != WSAEWOULDBLOCK && err != EAGAIN)
+                  break;
+#else
+               if(errno != EWOULDBLOCK && errno != EAGAIN)
+                  break;
+#endif
+            }
+            else
+               break;//closed by remote(0), or error(SOCKET_ERROR)
```
Paul-Licameli, line 150: Actually that path is only for closed by remote

libraries/lib-ipc/internal/BufferedIPCChannel.cpp
Paul-Licameli: Are we certain every error seen by a sending thread involves an error in the receiving thread in the other process?  Errors can happen many ways.

It looks like there is some possiblity of sent data being lost but the user of the channel is never informed of it.  Maybe error handling could be more complete.

Paul-Licameli (approved):

PR #2883, opened 2022-05-02
added new state for AudioThread loop

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2861

This is a cleaner (compared to former PR https://github.com/audacity/audacity/pull/2868) fix to all 3 bugs listed in the linked issue. 

Recommended:

### DIFF
--- a/src/AudioIO.cpp
+++ b/src/AudioIO.cpp
@@ -1738,7 +1738,7 @@ double AudioIO::GetStreamTime()
 
 AudioThread::ExitCode AudioThread::Entry()
 {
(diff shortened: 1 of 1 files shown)


Paul-Licameli (approved):
Simple and effective.

But now the worker thread tests `IsMonitoring` and that depends on two members `mPortStreamV19` and `mStreamToken` which are not atomic.  Strictly speaking this meets the definition of a "data race".

But I will not insist, now, on review and correction of all uses of those member variables.

It might merit another issue for cleanup.

Issue #2884, opened 2022-05-02
The new "Rename Clip" command is poorly positioned in the Edit menu structure - would be better in Audio Clips submenu


petersampsonaudacity:
**Describe the bug**
The new "Rename Clip" command is poorly positioned in the Edit menu structure, especially so since for 3.2.0 the old menu item for "Clip Boundaries" has been renamed to "Audio Clips".

With that change surely it would be better for the "Rename Clip" menu item to be in the submenu "Audio Clip" as that submenu deals with actions on Clips.

**To Reproduce**
1. Click on the Edit menu
2. Observe: " Rename Clip" is in this menu -see image 1
3. Click on "Audio Clips"
4. Observe:  all entries are actions on a Clip or Clips

**Actual behavior**
"Rename Clip" in in the top level Edit menu

**Expected behavior**
"Rename Clip"  to be in the submenu for "Audio Clips"

**Screenshots**
### 1) Edit Menu
![image](https://user-images.githubusercontent.com/50205984/166227217-b45826e9-9127-44b4-ad6b-a3812d2a08b8.png)

### 2) Audio Clips sub-menu
![image](https://user-images.githubusercontent.com/50205984/166227386-cb0c6f6d-2c77-4513-8d2f-e5b104049d28.png)

**Additional information (please complete the following information):**
 - OS: Windows 10 - but assume all OS
 - Version: 3.2.0

**Additional context**
I countered this as I have started work  _(planning only at present - the updating is tricksy)_  on updating the 3.2.0 alpha Manual for the recent changes to the Edit menu for this release.  The more I looked at it the more I thought that Rename Clip would be much better in the submenu for Audio Clips.

One also has to question just why this new menu item for Rename Clips has been added when there is a perfectly good mechanism in the right-click context menu for a click.

Tagging @Tantacrul and @LWinterberg as a design issue.

petersampsonaudacity:
I will hold off working on updating this part of the Manual until this is resolved one way or the other, adopting or rejecting this proposal.

LWinterberg:
Good idea. This only emerged because I renamed Clip Boundaries to Audio Clips :)

DavidBailes:
When this command was introduced there was the intention at some point to generalize it so that it could be used for renaming either audio clips or labels. If that is still the intension, then that may well affect what is considered the best placement of the command in the menus.

petersampsonaudacity:
[...]

PR #2885, opened 2022-05-02
move Rename Clip into the Audio Clips menu

LWinterberg:
Resolves: #2884 

or, well, that'd be the plan anyway


LWinterberg:
I was kinda hoping for this to sort itself in here, but apparently I misunderstand what the code does. 

 https://github.com/audacity/audacity/blob/master/src/menus/EditMenus.cpp#L1159

petersampsonaudacity:
BUMP @LWinterberg

Issue #2842, opened 2022-04-20
Settings, validator, instance classes, and statelessness of AudioUnit effects


Paul-Licameli:
- [x] #2931
- [x] #2933
- [x] #2935
- [x] #2939
- [x] #2943

Paul-Licameli:
all done

PR #2895, opened 2022-05-05
Stateless audio unit effect

Paul-Licameli:
Resolves: #2842

Settings, Validator, Instance, and statelessness for AudioUnit effects

Recommended:

### DIFF
--- a/src/effects/audiounits/AudioUnitEffect.cpp
+++ b/src/effects/audiounits/AudioUnitEffect.cpp
@@ -1105,11 +1105,10 @@ size_t AudioUnitEffect::GetTailSize() const
 bool AudioUnitEffect::ProcessInitialize(
    EffectSettings &, sampleCount, ChannelNames chanMap)
 {
-   mInputList.reinit(mAudioIns);
-   mInputList[0].mNumberBuffers = mAudioIns;
-   
-   mOutputList.reinit(mAudioOuts);
-   mOutputList[0].mNumberBuffers = mAudioOuts;
+   mInputList =
+      PackedArrayAllocateCount<AudioBufferList>(mAudioIns)(mAudioIns);
+   mOutputList =
+      PackedArrayAllocateCount<AudioBufferList>(mAudioOuts)(mAudioOuts);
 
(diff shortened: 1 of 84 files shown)


src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -1831,8 +1861,12 @@ TranslatableString AudioUnitEffect::Export(const wxString & path) const
 }
 
 std::pair<CF_ptr<CFDataRef>, TranslatableString>
-AudioUnitEffect::MakeBlob(const wxCFStringRef &cfname, bool binary) const
+AudioUnitEffect::MakeBlob(const AudioUnitEffectSettings &settings,
+   const wxCFStringRef &cfname, bool binary) const
 {
+   // Update state of the unit from settings
+   StoreSettings(settings);
+
```
pietro68, line 1837: The commit message says "...And use it in SavePreset" but it is actually used only here (?)

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -1184,10 +1184,8 @@ bool AudioUnitEffect::RealtimeAddProcessor(
    slave->SetBlockSize(mBlockSize);
    slave->SetSampleRate(sampleRate);
 
-   if (!CopyParameters(mUnit.get(), slave->mUnit.get()))
-   {
+   if (!slave->StoreSettings(GetSettings(settings)))
```
pietro68, line 1205: I may be missing something here but: is this really equivalent to what happened before? this sets the slave with input parameter `settings` - these in principle could be different from the settings currently in mUnit

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -967,6 +979,18 @@ bool AudioUnitEffect::InitializePlugin()
    return true;
 }
 
+struct AudioUnitEffect::Validator : DefaultEffectUIValidator {
```
pietro68, line 992: Remember to use base class EffectUIValidator :-)

PR #2901, opened 2022-05-06
Plugin validation preliminary work

vsverchinsky:

Recommended:


libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -373,14 +373,7 @@ void PluginManager::Initialize(FileConfigFactory factory)
       // Allow the module to auto-register children
       module->AutoRegisterPlugins(*this);
    }
-
-   // And finally check for updates
-#ifndef EXPERIMENTAL_EFFECT_MANAGEMENT
    CheckForUpdates();
-#else
-   const bool kFast = true;
-   CheckForUpdates( kFast );
```
Paul-Licameli, line 626: To make this commit have no behavior changes, the other conditional compilation branch should be kept.

libraries/lib-xml/XMLFileReader.cpp
```diff
@@ -208,7 +208,18 @@ void XMLFileReader::startElement(void *userData, const char *name,
    }
 
    if (XMLTagHandler *& handler = handlers.back()) {
-      if (!handler->ReadXMLTag(name, atts)) {
```
Paul-Licameli, line 211: Commit comment has a spelling mistake

src/effects/EffectManager.h
```diff
@@ -81,13 +81,6 @@ class AUDACITY_DLL_API EffectManager
    EffectManager();
    virtual ~EffectManager();
 
-   //! Here solely for the purpose of Nyquist Workbench until a better solution is devised.
-   /** Register an effect so it can be executed.
-     uEffect is expected to be a self-hosting Nyquist effect */
-   const PluginID & RegisterEffect(std::unique_ptr<EffectPlugin> uEffect);
-   //! Used only by Nyquist Workbench module
-   void UnregisterEffect(const PluginID & ID);
```
Paul-Licameli, line 89: This commit will break compilation of Nyquist Workbench.  Although we are not maintaining that now, I don’t want to abandon all hope of salvaging interesting ideas from that in the future.

Try to keep it at least compilable if not runnable.  Keep the unregister function.  Let the work bench call the other overload of RegisterPlugin passing null for provider.  PluginDescriptor::Load should allow empty providerID without error.

(Also the simlification of Load properly belongs to this commit, not the previous one.)

libraries/lib-module-manager/PluginDescriptor.h
```diff
@@ -12,6 +12,8 @@
 #include <wx/string.h>
 
 #include "PluginInterface.h"
+#include "XMLTagHandler.h"
+#include "XMLWriter.h"
```
Paul-Licameli, line 16: This header only needs a forward declaration of type `XMLWriter` and should avoid the nested header

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -389,26 +394,18 @@ void PluginManager::Initialize(FileConfigFactory factory)
...
 
-   // Now get rid of the modules
-   iter = mPlugins.begin();
-   while (iter != mPlugins.end())
-   {
-      mPlugins.erase(iter++);
-   }
+   // Now get rid of others
+   mRegisteredPlugins.clear();
+   mLoadedInterfaces.clear();
```
Paul-Licameli, line 392: My last remark for this round of review:  I understand these changes of PluginManager and PluginDescriptor but I’m waiting to see how they are used in the next rebasing of the other work for plugin validation.

The suggestion in old comments that PluginDescriptor might become a root of a class hierarchy— this didn’t happen here and I suppose it will not, or else I would have seen it in these preliminaries.  No need yet for that other idea.

libraries/lib-module-manager/PluginManager.cpp
vsverchinsky: Thanks!

src/effects/EffectManager.h
vsverchinsky: Probably I misunderstood what you said, so I'll revert those changes (except changes in `PluginManager::Load`), and squash this commit with the previous one

libraries/lib-xml/XMLTagHandler.cpp
```diff
@@ -88,23 +88,6 @@ bool XMLValueChecker::IsGoodPathString(const FilePath &str)
...
-      const char* name = *attrs++;
-      const char* value = *attrs++;
-
-      mCurrentTagAttributes.emplace_back(
-         std::string_view(name), XMLAttributeValueView(std::string_view(value)));
-   }
-
-   bool result = HandleXMLTag(tag, mCurrentTagAttributes);
-
-   return result;
```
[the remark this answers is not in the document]
vsverchinsky: The main purpose wasn't the performance improvement. XMLTagHandler is usually inherited by other types and thus `mCurrentTagAttributes` become a part of objects of those types, but it used only during xml parsing.

libraries/lib-module-manager/PluginDescriptor.h
```diff
@@ -114,41 +112,29 @@ class MODULE_MANAGER_API PluginDescriptor
...
-   // Load a plug-in, or locate a PluginProvider
-   ComponentInterface *Load();
-   void Set(std::unique_ptr<ComponentInterface> instance);
 
    // Common
 
-   // Among other purposes, PluginDescriptor acts as the resource handle,
-   // or smart pointer, to a resource created in a plugin library, and is responsible
-   // for a cleanup of this pointer.
-   std::unique_ptr<ComponentInterface> muInstance; // may be null for a module
```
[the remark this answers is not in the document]
vsverchinsky: It was one of the reasons, yes. Also PluginDescriptor now inherits XMLTagHandler, which could make state maintenance a bit more complicated

libraries/lib-module-manager/PluginManager.cpp
vsverchinsky: Alternative might be use attached objects instead of building a hierachy

Paul-Licameli:
> The main purpose wasn't the performance improvement. XMLTagHandler is usually inherited by other types and thus `mCurrentTagAttributes` become a part of objects of those types, but it used only during xml parsing.

I see: space savings are the real motivation.  That's good too.

Paul-Licameli:
`PluginDescriptor::IsEffectLegacy` and `PluginDescriptor::mEffectLegacy` are now removable in this branch.

Paul-Licameli (approved):
One small suggestion to remove some unused code.  Nyquist workbench will still compile.

Please rebase, force push, and merge with a merge commit.  Please do this soon, because I need to base the other needed framework changes I mentioned today (those blocking further development of statelessness) on these prior changes to PluginManager and PluginDescriptor.

Issue #2900, opened 2022-05-06
Steps to reproduce #2137 now cause division-by-zero crash


Paul-Licameli:
Steps to reproduce #2137 now cause division-by-zero crash, when mono tracks are joined to stereo and show no wave at all, and then you mouse-over the track, while *in multi-tool mode* (F6 key)

_Originally posted by @Paul-Licameli in https://github.com/audacity/audacity/issues/2137#issuecomment-1119655833_

Crashing line is in EnvelopeHandle.cpp where `CountourSpacing` is zero:

`   distance = abs((yDisplace % ContourSpacing) - ContourSpacing / 2);`

PR #2904, opened 2022-05-08
Issue2900: fix modulus-by-zero crash mousing over collapsed track...

Paul-Licameli:
... with no waveform visible because there are drag handles in both channels
of a stereo track make by joining two mono tracks

Resolves: #2900

Recommended:


pietro68 (approved):
I stepped-debug both unfixed and fixed, I think the fix is ok

Issue #2909, opened 2022-05-10
When the Export Multiple dialog opens on Windows, the first control is not the focus


DavidBailes:
**Describe the bug**
When the Export Multiple dialog opens on Windows, the first control is not the focus

**To Reproduce**
1. Open the Export Multiple dialog

**Actual behavior**
One of the controls in the Options section is the focus.

**Expected behavior**
The first control in the dialog, the Folder text box, is the focus.

**Additional information (please complete the following information):**
 - OS: Windows 10. The bug does not occur on macOS or Xubuntu
 - Version 3.1.3

**Additional context**
This bug is a regression wrt Audacity 2.4.1
The bug can be confusing for users of screen readers who may assume that the initial focus is the first control, and so not find some of the controls in this dialog.

petersampsonaudacity:
Tested @DavidBailes  branch audacity-win-3.2.0-alpha-20220510+97a6dc2-x64-msvc2022 on W10

This looks to be fixed David's changes, with initial focus on the folder location for export.

PR #2910, opened 2022-05-10
Issue #2909: Export Multiple dialog: focus issue on Windows

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/2909

Problem:
When the Export Multiple dialog opens on Windows, the first control is not the focus.

This bug was introduced between Audacity 2.4.1 and 2.4.2. During this period there were no significant changes to the code of this dialog. However the version of wxWidgets changed from 3.1.1 to 3.1.3, so this is probably a wxWidgets bug.

Fix:
Explicitly setting the first control to be the focus, as a work around for the assumed wxWidgets bug.

Recommended:

### DIFF
--- a/src/export/ExportMultiple.cpp
+++ b/src/export/ExportMultiple.cpp
@@ -219,6 +219,13 @@ int ExportMultipleDialog::ShowModal()
 
    EnableControls();
 
+   // This is a work around for issue #2909, and ensures that
+   // when the dialog opens, the first control is the focus.
+   // The work around is only needed on Windows.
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
Tested @DavidBailes  branch audacity-win-3.2.0-alpha-20220510+97a6dc2-x64-msvc2022 on W10

This looks to be fixed David's changes, with initial focus on the folder location for export.

Paul-Licameli (approved):

PR #2922, opened 2022-05-12
Make reverb stateless

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2920

Recommended:

### DIFF
--- a/src/effects/Reverb.cpp
+++ b/src/effects/Reverb.cpp
@@ -31,20 +31,6 @@
 #include "Reverb_libSoX.h"
 
 
-enum 
-{
-   ID_RoomSize = 10000,
-   ID_PreDelay,
-   ID_Reverberance,
-   ID_HfDamping,
-   ID_ToneLow,
-   ID_ToneHigh,
-   ID_WetGain,
-   ID_DryGain,
-   ID_StereoWidth,
-   ID_WetOnly
-};
-
 const EffectParameterMethods& EffectReverb::Parameters() const
 {
    static CapturedParameters<EffectReverb,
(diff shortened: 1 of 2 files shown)


src/effects/Reverb.cpp
```diff
@@ -297,23 +352,37 @@ bool EffectReverb::DoLoadFactoryPreset(int id)
...
+
+void EffectReverb::Validator::PopulateOrExchange(ShuttleGui & S)
 {
    S.AddSpace(0, 5);
 
    S.StartMultiColumn(3, wxEXPAND);
    {
       S.SetStretchyCol(2);
 
+      wxEVT_TEXT;
```
Paul-Licameli, line 408: Has no effect

src/effects/Reverb.cpp
```diff
@@ -367,38 +439,23 @@ bool EffectReverb::TransferDataToWindow(const EffectSettings &)
...
-   rs.mHfDamping = mHfDampingS->GetValue();
-   rs.mToneLow = mToneLowS->GetValue();
-   rs.mToneHigh = mToneHighS->GetValue();
-   rs.mWetGain = mWetGainS->GetValue();
-   rs.mDryGain = mDryGainS->GetValue();
-   rs.mStereoWidth = mStereoWidthS->GetValue();
-   rs.mWetOnly = mWetOnlyC->GetValue();
-
-   return true;
-}
```
Paul-Licameli, line 378: I see a deleted function above but I don’t see the ValidateUI that should replace it in the second step.

src/effects/Reverb.cpp
```diff
@@ -166,6 +166,36 @@ bool EffectReverb::Validator::ValidateUI()
...
+      sampleCount totalLen, ChannelNames chanMap) override;
+
+   size_t ProcessBlock(EffectSettings& settings,
+      const float* const* inBlock, float* const* outBlock, size_t blockLen)  override;
+
+   bool ProcessFinalize(void) override; // not every effect needs this
+
+
+   unsigned mNumChans {};
+   Reverb_priv_t *mP;
```
Paul-Licameli, line 188: An in class initializer for the pointer

src/effects/Reverb.cpp
```diff
@@ -383,20 +376,21 @@ bool EffectReverb::DoLoadFactoryPreset(int id)
       return false;
    }
 
-   mSettings = FactoryPresets[id].preset;
+   // TODO - ::LoadFactoryPreset should belong to the instance,
+   // but the framework is not ready for that yet
+   // 
+   // mSettings = FactoryPresets[id].preset;
```
Paul-Licameli, line 382: The left side of the assignment simply changes to `GetSettings(settings)` and you no longer need Const_cast and a Do... function.

That eliminates all const_cast from this source file, a good thing to mention in the commit comment.

src/effects/Reverb.cpp
pietro68: Hmm no, I do not delete TransferDataFromWindow here, I only replace all occurrences of mParams with local reference rs.

pietro68: In step 2, you can see that the contents of this method has moved to ValidateUI

src/effects/Reverb.cpp
pietro68: good point, will do

src/effects/Reverb.cpp
pietro68: Ok - I was wrong, the framework was not missing something, rather I did not fully understand it. I did how you suggest above (actually - before reading your comment; good that I came to realize it by myself!) and I verified that presets are correctly loaded.

src/effects/Reverb.cpp
Paul-Licameli: I see it now.  I overlooked it.

If you don't copy and paste the function to a different place in the file, it could help make the commit more easily reviewable.

Issue #2923, opened 2022-05-13
Make EffectReverb realtime capable


pietro68:
For simplicity, this will be done first to the original, stateful effect; later it will be done on the stateless effect too.

Ryan-Miller-Audacity:
After discussion with Sergey @dozzzzer , this ticket is not fully testable until completion of #2954 as only then will we be able to see actual realtime effect playback. So while it was tested above, this will not suffice to fully meet the acceptance criteria for this ticket.

Ryan-Miller-Audacity:
This ticket is being superseded by #3155 which denotes much further work needed for reverb to become fully realtime usable.

PR #2927, opened 2022-05-13
Make reverb realtime

pietro68:
Resolves: part of https://github.com/audacity/audacity/issues/2923

The realtime transformation was performed on the stateful version for sake of simplicty: this is my first attempt to make an effect realtime, so I wanted as few additional issues as possible. Anyway, once this PR is approved, applying the same changes to the stateless version should be straightforward.

Recommended:

### DIFF
--- a/src/effects/How to make an effect realtime.txt
+++ b/src/effects/How to make an effect realtime.txt
@@ -1,0 +1,100 @@
+HOW TO MAKE AN EFFECT REALTIME CAPABLE
+====================================================================================================
+
+HEADER (.h)
+
+Add these declarations:
+
+public:
+
+   bool SupportsRealtime() const override;
+
+   bool RealtimeInitialize(EffectSettings &settings) override;
+   
+   bool RealtimeAddProcessor(EffectSettings &settings,
+      unsigned numChannels, float sampleRate) override;
+      
+   bool RealtimeFinalize(EffectSettings &settings) noexcept override;
+   
+   size_t RealtimeProcess(int group,  EffectSettings &settings,
+                          const float *const *inbuf, float *const *outbuf, size_t numSamples) override;
+
+private:
+
+   EffectDistortionState              mMaster;
+   std::vector<EffectDistortionState> mSlaves;
+   
+   void InstanceInit(EffectDistortionState & data, float sampleRate);
+   
+   // the above will be called by both: 
+   // - the destructive process init (::ProcessInitialize)
+   // - the realtime process init (::RealtimeAddProcessor)
+   
+   size_t InstanceProcess(EffectSettings& settings,
+                          EffectDistortionState& data,
+                          const float *const *inBlock, float *const *outBlock, size_t blockLen);
+   
+   // the above will be called by both: 
+   // - the destructive process (::ProcessBlock)
+   // - the realtime process    (::InstanceProcess)
+   
+   
+   
+====================================================================================================
+
+IMPL (.cpp)
+
+Add these implementations:
+
+    bool EffectDistortion::SupportsRealtime() const
+    {
+       return true;
+    }
+
+then something along the lines of:
+
+    bool EffectDistortion::RealtimeInitialize(EffectSettings &)
+    {
+       SetBlockSize(512);
+
+       mSlaves.clear();
+
+       return true;
+    }
+
+    bool EffectDistortion::RealtimeAddProcessor(
+       EffectSettings &, unsigned, float sampleRate)
+    {
+       EffectDistortionState slave;
+
+       InstanceInit(slave, sampleRate);
+
+       mSlaves.push_back(slave);
+
+       return true;
+    }
+
+    bool EffectDistortion::RealtimeFinalize(EffectSettings &) noexcept
+    {
+       mSlaves.clear();
+
+       return true;
+    }
+
+    size_t EffectDistortion::RealtimeProcess(int group, EffectSettings &settings,
+       const float *const *inbuf, float *const *outbuf, size_t numSamples)
+    {
+       return InstanceProcess(settings, mSlaves[group], inbuf, outbuf, numSamples);
+    }
+
+
+
+
+
+
+
+
+
+   
+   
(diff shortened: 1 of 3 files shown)


src/effects/How to make an effect realtime.txt
```diff
@@ -0,0 +1,100 @@
...
+      
+   bool RealtimeFinalize(EffectSettings &settings) noexcept override;
+   
+   size_t RealtimeProcess(int group,  EffectSettings &settings,
+                          const float *const *inbuf, float *const *outbuf, size_t numSamples) override;
+
+private:
+
+   EffectDistortionState              mMaster;
+   std::vector<EffectDistortionState> mSlaves;
```
Paul-Licameli, line 25: The aray of slaves, strictly, isn't needed until the step 2 commit.

src/effects/Reverb.cpp
```diff
@@ -183,31 +190,43 @@ bool EffectReverb::ProcessInitialize(
...
    return true;
 }
 
 bool EffectReverb::ProcessFinalize()
 {
-   for (unsigned int i = 0; i < mNumChans; i++)
+   return InstanceDeinit(mMaster);
+}
+
+bool EffectReverb::InstanceDeinit(ReverbState& state)
```
Paul-Licameli, line 204: Why is this function here? -- Because, this file is one of the few places missed by the systematic cleanup of naked news and deletes in favor of smart pointers that I did tyears ago.

Let's fix this with a more modern smart pointer while disturbing this file.

src/effects/Reverb.cpp
```diff
@@ -158,21 +158,28 @@ unsigned EffectReverb::GetAudioOutCount() const
...
+   if (    (chanMap && chanMap[0] != ChannelNameEOL && chanMap[1] == ChannelNameFrontRight)
+        || forceStereo )
    {
       isStereo = true;
-      mNumChans = 2;
+      state.mNumChans = 2;
    }
 
-   mP = (Reverb_priv_t *) calloc(sizeof(*mP), mNumChans);
+   state.mP = (Reverb_priv_t *) calloc(sizeof(*state.mP), state.mNumChans);
```
Paul-Licameli, line 178: Define something like this (you don't need my comments in it):
```
struct Reverb_priv_ex : Reverb_priv_t {
   Reverb_priv_ex() : Reverb_priv_t{} {} // this does zero-filling of the new structure, as calloc would
   ~Reverb_priv_ex{ reverb_delete(reverb); } };
```

Then:

```
state.mP = std::make_unique<Reverb_priv_ex[]>(state.mNumChans)
```

Make appropriate change of type of mP so this assignment works.

src/effects/Reverb.cpp
```diff
@@ -183,31 +190,43 @@ bool EffectReverb::ProcessInitialize(
...
+bool EffectReverb::InstanceDeinit(ReverbState& state)
+{
+   for (unsigned int i = 0; i < state.mNumChans; i++)
    {
-      reverb_delete(&mP[i].reverb);
+      reverb_delete(&state.mP[i].reverb);
    }
 
-   free(mP);
+   free(state.mP);
```
Paul-Licameli, line 211: Then the above becomes unnecessary (and, wrongly double-freeing!) for avoidance of memory leaks.  The smart pointer does all.

Paul-Licameli (changes requested):
Rewrite this old code better with smart pointers

Paul-Licameli:
Do the smart pointer cleanup while we are disturbing this old code; otherwise, all looks good.

Paul-Licameli:
Actually:  it wasn't all good, because as it is now written, it will leak memory, because InstanceDeinit is not called to balance all of the InstanceInit calls.

But, do as I suggest with smart pointers, and all will be good with fewer lines of code.

src/effects/Reverb.cpp
pietro68: I'll be happy to bring this code into the early '10s...

src/effects/Reverb.cpp
pietro68: Done; although it had to be `reverb_delete(&reverb)`

A consequence of using the unique_ptr in the State, was that when pushing back a slave to the slaves vector, I had to use `push_back( std::move(slave) )` instead of a simple `push_back( slave )`

src/effects/Reverb.cpp
pietro68: The code above that line had to be removed too, because the reverb_delete now happens in the destructor of `Reverb_priv_ex`

src/effects/Reverb.cpp
```diff
@@ -198,18 +209,6 @@ bool EffectReverb::InstanceInit(ReverbState& state, ChannelNames chanMap, bool f
...
-
-bool EffectReverb::InstanceDeinit(ReverbState& state)
-{
-   for (unsigned int i = 0; i < state.mNumChans; i++)
-   {
-      reverb_delete(&state.mP[i].reverb);
-   }
-
-   free(state.mP);
-
```
Paul-Licameli, line 200: I think you still need `state.mP.reset()` to free memory as soon as possible?

Paul-Licameli:
Now I see that this branch isn't based on the other that made Revert stateless, and there is non-trivial conflict resolution to do.

This would have been better done as a continuation of the other branch.

src/effects/Reverb.cpp
Paul-Licameli: Rebase this branch onto the other for statelessness, and it won't matter I think, because the Instance is soon destroyed when effect processing is finished.

Paul-Licameli (changes requested):
Redo it, based on the branch for statelessness, which will cause many conflicts for resolution.

There is no point in further review before that is done.

pietro68:
My reason to start this on the stateful version, was to have the easiest possible environment to understand what needs to be done - next time I will surely try stateless+realtime in one go. New PR as requested at https://github.com/audacity/audacity/pull/2938

PR #2932, opened 2022-05-15
Split au

Paul-Licameli:
Resolves: #2931

Preliminary splitting of AudioUnitEffect.cpp into multiple source files

Recommended:


pietro68 (approved):
I see this PR is about introducing a namespace and breaking up AudioUnitEffect.* in different files, this is all fine by me.

Issue #2935, opened 2022-05-16
Define Validator for AudioUnitEffect


Paul-Licameli:
Define appropriate Validator class for AudioUnit effects

PR #2936, opened 2022-05-16
Au validator

Paul-Licameli:
Resolves: #2935

Recommended:

### DIFF
--- a/src/effects/audiounits/AudioUnitEffect.cpp
+++ b/src/effects/audiounits/AudioUnitEffect.cpp
@@ -307,25 +307,34 @@ bool AudioUnitEffect::InitializePlugin()
    return true;
 }
 
-class AudioUnitValidator : public DefaultEffectUIValidator {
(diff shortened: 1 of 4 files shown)


src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -286,6 +307,28 @@ bool AudioUnitEffect::InitializePlugin()
    return true;
 }
 
+class AudioUnitValidator : public DefaultEffectUIValidator {
```
pietro68: Not sure if I remember this correctly, anyway: we said we should use EffectUIValidator directly?

Paul-Licameli: That is transitional and changes again in a later commit

pietro68 (approved):
Tried this branch with the usual AUs, all fine; changes are ok for me, and I take them as a guide for the corresponding step for VST

PR #2940, opened 2022-05-16
Au instance

Paul-Licameli:
Resolves: #2939
Partly resolves: #3069

Define the Instance class for AudioUnit effects.

Incidentally, ouput monitoring as with AUDymamicProcessing works again in the non-modal dialog (ALTHOUGH, it is actually reflecting only the first track, not the mix).

However, I anticipate that after full statelessness, and use of such effects in the per-track effect stacks -- there will remain other problems with slow rate of update.

Recommended:

### DIFF
--- a/src/effects/audiounits/AudioUnitEffect.cpp
+++ b/src/effects/audiounits/AudioUnitEffect.cpp
@@ -233,7 +233,6 @@ bool AudioUnitEffect::InitializeInstance()
       return false;
 
    mSampleRate = 44100;
-   GetChannelCounts();
    SetRateAndChannels();
 
    // Retrieve the desired number of frames per slice
@@ -270,6 +269,8 @@ bool AudioUnitEffect::InitializePlugin()
    // It's not just for implementing EffectInstance.  AudioUnits is unlike other
    // third party effect families that distinguish the notions of plug-in and
    // instance.
+
+   GetChannelCounts();
 
    // When AudioUnitEffect implements its own proper Instance class, this
    // should call CreateAudioUnit() directly and not do the rest of
--- a/src/effects/audiounits/AudioUnitEffect.cpp
+++ b/src/effects/audiounits/AudioUnitEffect.cpp
@@ -99,7 +99,6 @@ AudioUnitEffect::AudioUnitEffect(const PluginPath & path,
    , mPath{ path }
    , mName{ name.AfterFirst(wxT(':')).Trim(true).Trim(false) }
    , mVendor{ name.BeforeFirst(wxT(':')).Trim(true).Trim(false) }
-   , mMaster{ master }
 {
 }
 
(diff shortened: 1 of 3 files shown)


pietro68:
I tested it and ran into an assertion. Generate a noise track, instantiate a AUGraphicEQ, select the track and click apply, this happens:

Screenshot 2022-05-26 at 10 10 46

Paul-Licameli:
The problem was that my changes did not preserve the behavior of SetBlockSize.   Please review one new commit at the end of the rebased branch.  Previous commits have not changed.

Paul-Licameli:
And one more small commit.  What was an assertion, should not be an assertion.  It's not provable.  The error should instead be propagated.

pietro68:
I verified the problem I reported above, does not happen anymore.

However, I noticed another one - please try the following:

1) new project, generate noise in mono track
2) instantiate a AUGraphicEQ, select 10 bands config, bring the 1khz band to full down (-20 dB)
3) start play
4) you will hear that the white noise is unchanged. But as soon as you give a slight nudge to the 1khz slider, the effect will be heard
5) if you press stop and repeat from (3), you will see it keeps happening

Honestly I do not know if this is a regression, compared to the previous step - it might have happened also there.

Paul-Licameli:
> I verified the problem I reported above, does not happen anymore.
> 
> However, I noticed another one - please try the following:
> 
> 1. new project, generate noise in mono track
> 2. instantiate a AUGraphicEQ, select 10 bands config, bring the 1khz band to full down (-20 dB)
> 3. start play
> 4. you will hear that the white noise is unchanged. But as soon as you give a slight nudge to the 1khz slider, the effect will be heard
> 5. if you press stop and repeat from (3), you will see it keeps happening
> 
> Honestly I do not know if this is a regression, compared to the previous step - it might have happened also there.

Does this same behavior happen in current master?  Then this does not indicate a bug in this pull request, but some other general bug requiring study and fix.

pietro68:
> Does this same behavior happen in current master? Then this does not indicate a bug in this pull request, but some other general bug requiring study and fix.

Tried it on master, but now I have a strange problem which prevents me from verifying the behavior. The problem is: when I try to press the play button, I only hear a system "bleep" telling me that the whole audacity window is disabled and won't receive/process user clicks, as long as an AU plugin window is displayed. I wonder what changed in the meantime - maybe some local settings, or an OSx update?

Anyway, this PR has been waiting too long so I will approve it and figure out later why the play button is disabled for me, and then verify whether the behavior is already present in master - it probably is, but if not I will bisect and open an issue.

pietro68 (dismissed):
EDIT
Changes are understood and fine for me; the two problems I reported above, came in from other merges. When testing this PR was still possible (because #3011 had yet to happen) I could observe no regressions, so I would say this PR could be merged.

Paul-Licameli:
However, you have also identified issue #3012 which does indicate an unintended consequence of previous AudioUnit refactorings.  I think I need to understand that, fix it, and rebase this PR onto that fix.

When I fix it, it might also imply another checkbox for the analogous work in VST and VST3.

pietro68:
I tested the head of this branch (1ee092f3f) and it is still not possible to press the play button in the main transport panel; this also means that I can not test the bug about forgotten slider positions.

What are the last 4 commits supposed to fix?

Paul-Licameli:
> I tested the head of this branch ([1ee092f](https://github.com/audacity/audacity/commit/1ee092f3fa16f4fe82d9c554291b1d0c6aa6fdad)) and it is still not possible to press the play button in the main transport panel; this also means that I can not test the bug about forgotten slider positions.
> 
> What are the last 4 commits supposed to fix?


> I tested the head of this branch ([1ee092f](https://github.com/audacity/audacity/commit/1ee092f3fa16f4fe82d9c554291b1d0c6aa6fdad)) and it is still not possible to press the play button in the main transport panel; this also means that I can not test the bug about forgotten slider positions.
> 
> What are the last 4 commits supposed to fix?

This branch was not yet rebased onto commit 27ad8e9833788fdc9e8d9f3438488d756096e66b which is why you could not test it.

Now I have rebased it.  Please satisfy yourself with testing.

Paul-Licameli:
> What are the last 4 commits supposed to fix?

I don't understand the question.  The last 4 are the only 4.  The third does most of the work of defining the Instance class, and the first does a necessary step before that, which moves a bit of code that doesn't belong properly to the instance.

The second commit is just a bit of stylistic simplification in some of the related code, and the fourth corrects a misuse of assertion that should instead an error return path.  I came across that incidentally in my own testing, but I forget how.

pietro68:
> I don't understand the question. The last 4 are the only 4
Pardon me, I was a bit confused yesterday by all this testing on different branches.

I tested the branch head `98236d27c` and I found a regression, compared to the start of the branch at `07ed40e50` :

Generate some noise, select some region of it, instantiate an AUGraphicEQ, 10 bands, bring all the way down bands from 1Khz onwards, click apply. Expected: the selected region should become much lower in volume Observed: the selected region stays unaffected.

Paul-Licameli:
> > I don't understand the question. The last 4 are the only 4
> 
> Pardon me, I was a bit confused yesterday by all this testing on different branches.
> 
> I tested the branch head `98236d27c` and I found a regression, compared to the start of the branch at `07ed40e50` :
> 
> Generate some noise, select some region of it, instantiate an AUGraphicEQ, 10 bands, bring all the way down bands from 1Khz onwards, click apply. Expected: the selected region should become much lower in volume Observed: the selected region stays unaffected.

Did you attempt any debugging beyond that?

I might begin with a breakpoint in AudioUnitInstance::ProcessBlock

Paul-Licameli:
Never mind, I think I understand now

Paul-Licameli:
@pietro68 Review once more!

The fix for this was smaller than I thought it would be, and specific only to AudioUnit effects, not requiring anything for other effects that are now stateless.

But other effect families may need to make the analogous step.

See what happens to the function ProcessInitialize().

pietro68:
I tested `444324f `- the problem I observed when using the effect destructively was solved, but I noted a regression not present in the base of the branch `d18bf6b` which is: when using AGraphicEQ in realtime, changes to the sliders are not applied. One must stop and restart play to hear the changes.

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -203,57 +202,107 @@ bool AudioUnitEffect::SupportsAutomation() const
...
+   OSStatus Render(AudioUnitRenderActionFlags *inActionFlags,
+      const AudioTimeStamp *inTimeStamp,
+      UInt32 inBusNumber, UInt32 inNumFrames, AudioBufferList *ioData);
+
+   bool BypassEffect(bool bypass);
+
+private:
+   //! Whether the master instance is now allocated to a group number
+   bool mRecruited{ false };
+   std::vector<std::unique_ptr<AudioUnitInstance>> mSlaves;
```
pietro68, line 259: I am a bit surprised to see that an Instance can have slaves. I thought the main point of having instances was to replace the slaves in the effect, externalizing them - so that what was once a slave, would now be an instance (and the first instance would be the "master")  maybe I am missing something here.

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -479,9 +535,12 @@ size_t AudioUnitEffect::GetTailSize() const
 }
 #endif
 
-bool AudioUnitEffect::ProcessInitialize(
-   EffectSettings &, sampleCount, ChannelNames chanMap)
+bool AudioUnitInstance::ProcessInitialize(
+   EffectSettings &settings, sampleCount, ChannelNames chanMap)
```
pietro68, line 539: This is more about the framework than about this PR, anyway: I think the passed EffectSettings reference should be const? or could in principle the settings be changed by the initialization?

pietro68 (approved):
I verified that all bugs previously reported by me on the parent issue were solved.

I understand the code; I only left a couple of comments for a later discussion, they are not really request for changes in this PR.

src/effects/audiounits/AudioUnitEffect.cpp
Paul-Licameli: I thought earlier we would need to externalize spaces as objects but decided against it.

Wahwah and Reverb have slaves implemented differently, not as occurrences of the instance class.

It doesn’t matter.  RealtimeAddProcessor and Process and Finalize are sufficient abstract interface on just one object.

src/effects/audiounits/AudioUnitEffect.cpp
Paul-Licameli: Ladspa and Lv2 have the notion of an output control port that the instance may change during processing.  So parts of EffectSettings may in fact be modifiable, and the changes settings in general need to be copied back to main thread.

Remember that: it will be relevant to other bug fixes soon.

PR #2941, opened 2022-05-16
Audio setup rework

ksoze95:

Recommended:


petersampsonaudacity:
Tested on macOS 12.3.1 Monterey with @ksoze95 branch build.

I see the new cogwheel "Audacity Audio Setup Toolbar" - can't test the options on Mac as there are no options to switch to.

I note that the Device Toolbar has been removed entirely (replace by the new toolbar).  I assume that this is intentional?

petersampsonaudacity:
> > I note that the Device Toolbar has been removed entirely (replaced by the new toolbar in the View>Toolbars menu). I assume that this is intentional?
> 
> No, this was a miscommunication. the device toolbar is supposed to be just hidden, not completely removed.

So that would give the user **_three_** ways to make the same settings - that seems a little OTT.  

I can see that you might want to keep it hidden by default just so that users used to having a Device Toolbar can still have one - but it is somewhat redundant.  The only advantage that it has is that (when enabled) it always shows the user what the basic devices settings are.

Tantacrul:
Perhaps we're being overly cautious. I'm worried about all those tutorials that show people how to record from their laptop, etc. I'll need to **really call this out** in our release video.

petersampsonaudacity:
> Perhaps we're being overly cautious. I'm worried about all those tutorials that show people how to record from their laptop, etc. I'll need to **really call this out** in our release video.

But since you'll be hiding the Device Toolbar by default a lot of those users will still be fretting ...

SteveDaulton:
I see that this is still listed as "work in progress", but in case it's helpful, it's not working correctly for me.
If the "Output Device" and/or "Input Device" are set to non-default settings in the main "Cog wheel" menu, then setting the devices in either of "Cog > Audio Settings" or "Preferences > devices" have no effect.

Testing with the Ubuntu 20.04 AppImage on Xubuntu 20.04.

petersampsonaudacity:
Testing on W10 with @ksoze95 latest branch build

I) the interaction between the Cogwheel Audio setup and the Devices Preferences is not working for the input and output devices

2) that interaction seems to work OK for the Host - and for Buffer Length and Latency Compensation

3) a nomenclature **_inconsistency_** has been introduced
The first entry on the Cogwheel menu is called **Driver Type** - this has always been **Host** in the App and documentation
image

In Preferences it is **Host**

petersampsonaudacity:
Making a device setting in the Cogwheel does not stick - it reverts to default
1. purge Audacity settings folder (for OOTB default factory settings)
2. Launch latest Cogwheel branch audacity-win-3.2.0-alpha-20220516+465d3ee-x64-msvc2022
3. Observe in both Cogwheel and Pref s the default playback an recording devices match and are the default
4. In Cogwheel change the recording device to be the non-default **Stereo Mix**  YMMV
5. Exit the Cogwheel
6. Click again the Cogwheel
7. Note that the Stereo Mix has not stuck and the input device has reverted to the default

Tantacrul:
@ksoze95

- Can you make all the changes to strings that @petersampsonaudacity has raised (basically returning to what Audacity already uses)?

petersampsonaudacity:
Testing on W10 with the latest branch build:  audacity-win-3.2.0-alpha-20220517+5d5623f-x64-msvc2022 

1) I find that if I make a change to a non-default device using Cogwheel>Audio settings dialog then the change does not stick and stays at default device.

2) But If I use Cogwheel > Input Device then to make then change than it douse stick - AND it shows as changed in Prefs

3) I still find that changes I make in Devices for the I/O devices still do not stick and revert/stay defaults

Tantacrul:
> @Tantacrul  if the Device Toolbar is to be retained (as you said earlier) then perhaps you can add a menu item to the Cogwheel menu for showing/hiding Device Toolbar ?
> 
> 
> 
> That would be  less cumbersome than going to View>Toolbars - and should be more "discoverable".

I would say we should get it functioning properly first and then test it a bit before doing that. I have no idea how much of an impact this change may make.

src/prefs/DevicePrefs.cpp
```diff
@@ -415,12 +415,16 @@ bool DevicePrefs::Commit()
    return true;
 }
 
-namespace{
-PrefsPanel::Registration sAttachment{ "Device",
-   [](wxWindow *parent, wxWindowID winid, AudacityProject *)
-   {
+PrefsPanel::Factory DevicePrefsFactory()
+{
+   return [](wxWindow *parent, wxWindowID winid, AudacityProject *) {
```
Paul-Licameli, line 420: You might make this simpler than the corresponding code in SpectrumPrefs.  There, I needed a non-empty lambda capture.  Here, you don't.  So just declare instead

```
PrefsPanel *DevicePrefsFactory(wxWindow *parent, wxWindowID winid, AudacityProject *)
```

src/toolbars/DeviceToolBar.cpp
```diff
@@ -96,6 +139,34 @@ const DeviceToolBar &DeviceToolBar::Get( const AudacityProject &project )
...
+#ifdef __WXMAC__
+      6
+#else
+      12
+#endif
+   };
+   r->SetFocusRect(r->GetClientRect().Deflate(deflation, deflation));
+
+   return r;
+}
```
Paul-Licameli, line 168: This looks like duplication from ControlToolBar.cpp.  Instead, let's lower the repeated function into ToolBar as an overload of MakeButton.

src/toolbars/DeviceToolBar.cpp
```diff
@@ -351,42 +452,44 @@ void DeviceToolBar::EnableDisableButtons()
...
+
+      // Some have a second
+      switch (iWinID)
+      {
+      case ID_OPEN_AUDIO_SETUP_BUTTON:
+         break;
+      }
+      ToolBar::SetButtonToolTip(
+         mProject, *pCtrl, commands.data(), commands.size());
+   }
```
Paul-Licameli, line 488: I see more imitation of ControlToolBar.cpp here.  Do you understand what this is doing?

For ControlToolBar, it looks up `name` in the CommandManager to determine the keystroke shortcut, if any, that is associated with that name, and then formats the hover text for the button to include that shortcut key in the description.

Here, no key will be found because "Open Audio Setup" is not the name of any command.

But then, perhaps it should be registered as a named command so it can indeed be bound to a keystroke if the user chooses in Preferences -- even if we do not decide to give it a key binding by default.

@DavidBailes would be interested I'm sure in keystroke accessibility of the new menu under the cogwheel.

src/toolbars/DeviceToolBar.cpp
```diff
@@ -724,9 +907,9 @@ void DeviceToolBar::ShowComboDialog(wxChoice *combo, const TranslatableString &t
    {
       S.StartHorizontalLay(wxCENTER, false);
       {
-         c = S.AddChoice( Verbatim( combo->GetName() ),
+         c = S.AddChoice( Verbatim( prompt.Translation() ),
```
Paul-Licameli, line 910: `Verbatim(prompt.Translation())` should simplify just to `prompt`

src/toolbars/DeviceToolBar.cpp
```diff
@@ -724,9 +907,9 @@ void DeviceToolBar::ShowComboDialog(wxChoice *combo, const TranslatableString &t
...
             transform_container<TranslatableStrings>( inputSources, Verbatim ),
-            combo->GetSelection());
+            TranslatableString( inputSource, {} ) );
```
Paul-Licameli, line 912: This should be `Verbatim(inputSource)`

Paul-Licameli:
Done with my first round of review of this draft

ksoze95:
> The same happens if I set either the input device or output device to be the non-deafults: Mic

It is not specific to this PR.  It appears to be broken on master as well.

Tantacrul:
> It is not specific to this PR. It appears to be broken on master as well.

OK, understood. Well, we should fix it before release separately then. Thanks!

petersampsonaudacity:
> > It is not specific to this PR. It appears to be broken on master as well.
> 
> OK, understood. Well, we should fix it before release separately then. Thanks!

Confirmed on W10 with latest alpha 3.2.0 master:  audacity-win-3.2.0-alpha-20220516+465d3ee-x64-msvc2022

So does that need a new GitHub issue logging for it ?

petersampsonaudacity:
@Tantacrul - might it not have been simpler if Cogwheel>Audio Settings...  had just invoked Devices Preferences rather than creating a whole new dialog for it, since it duplicates everything that Devices Prefs does?  

Or is the plan to remove Devices Preferences from the Prefs settings?

Tantacrul:
> @Tantacrul - might it not have been simpler if **Cogwheel>Audio Settings...** had just invoked **Devices Preferences** rather than creating a whole new dialog for it, since that new dialog duplicates everything that Devices Prefs does?
> 
> Or is the plan to remove Devices Preferences from the Prefs settings in favor of the new cogwheel dialog?

Let's discuss all that later. And probably best discussed in Discord too to avoid making poor Kit and Paul's head spin :)

petersampsonaudacity:
> I will log a new GitHub issue logging for this.

### Update
Done:  **Devices Prefs ignores user settings for changes of I/O devices** #2948

petersampsonaudacity:
> Let's discuss all that later. And probably best discussed in Discord too to avoid making poor Kit and Paul's head spin :)

So do you want to kick that off @Tantacrul or do you want me to do that?

petersampsonaudacity:
Can I please suggest that the Cogwheel Toolbar is moved across leftwards so that it is adjacent to the Transport Toolbar (as it relates to what goes on in the Transport Toolbar:
image

petersampsonaudacity:
Oh that's a shame I've already gotten to like Mr, Coggy ...
Folk are very used to seeing cogwheels for setup stuff.

And yes I like that new layout look

Tantacrul:
Giving it a name like '**mr Coggy**' is a wonderful tactic to make me feel guilty about changing him.

Paul-Licameli:
> Giving him a name like '**Mr Coggy**' is a wonderful tactic to make me feel guilty about changing him.

Would you rather have Mr Clippy?

petersampsonaudacity:
> Giving him a name like '**Mr Coggy**' is a wonderful tactic to make me feel guilty about changing him.

I'm lovin' Mr Coggy - my Mystery-Shopper does too

We both feel that he gives an immediate modern idiom twist to Audacity's old-fashioned GUI.

It's a knowing twist that says "yeah we know the GUI looks ancient - but we can fix it!" - a pointer to the future the GUI/GUIs yet to come ,,,

SteveDaulton:
"Audio Setup" is much less ambiguous than a cog icon.

While the gear / cog icon has become familiar to most in relation to "settings", there's no indication that it has anything to do with _audio_ settings until you actually use it. In terms of visual  language, the gear / cog icon is vague.

petersampsonaudacity:
> While the gear / cog icon has become familiar to most in relation to "settings", there's no indication that it has anything to do with _audio_ settings until you actually use it. In terms of visual language, the gear / cog icon is vague.

Perhaps then we could consider retaining Mr Coggy but re-purposing it as a more "discoverable" way of invoking Preferences.

In that role it could perhaps become a single height button.

Tantacrul:
> > Giving him a name like '**Mr Coggy**' is a wonderful tactic to make me feel guilty about changing him.
> 
> Would you rather have Mr Clippy?

"It looks like you're trying to record a podcast! Want help with that?"

src/toolbars/AudioSetupToolBar.cpp
```diff
@@ -0,0 +1,840 @@
...
+*******************************************************************//*!
+
+\class AudioSetupToolBar
+\brief A toolbar to allow easier changing of input and output devices .
+
+*//*******************************************************************/
+
+
+
+#include "AudioSetupToolBar.h"
```
Paul-Licameli, line 16: When you make a new .cpp/.h file pair, the .cpp should include the .h before any other inclusion, to be sure the .h file is really self-contained.

And you did that!  Was it accident or were you aware of this rule?

src/toolbars/AudioSetupToolBar.h
```diff
@@ -0,0 +1,97 @@
...
+class wxCommandEvent;
+class wxBoxSizer;
+class wxSize;
+class wxPoint;
+class wxMenu;
+class wxString;
+struct DeviceSourceMap;
+
+class AButton;
+class AudacityProject;
```
Paul-Licameli, line 30: Maybe you copied and pasted.  I like to keep the #include directives and forward declarations minimal.  I think you can eliminate:

<vector>
class wxCommandEvent;
class wxBoxSizer;
class wxSize;
class wxPoint;
class wxMenu;
class wxString;
class AButton;
class AudacityProject;

src/toolbars/AudioSetupToolBar.h
```diff
@@ -0,0 +1,97 @@
...
+
+class AudioSetupToolBar final : public ToolBar {
+
+ public:
+
+   AudioSetupToolBar( AudacityProject &project );
```
Paul-Licameli, line 36: `explicit` ?

src/toolbars/AudioSetupToolBar.cpp
```diff
@@ -0,0 +1,840 @@
...
+static RegisteredToolbarFactory factory{ AudioSetupBarID,
+   []( AudacityProject &project ){
+      return ToolBar::Holder{ safenew AudioSetupToolBar{ project } };
+   }
+};
+
+namespace {
+AttachedToolBarMenuItem sAttachment{
+   /* i18n-hint: Clicking this menu item shows the toolbar
+      that manages devices */
```
Paul-Licameli, line 836: i18n (internationalization) hint comments are significant.  They get extracted into translation catalog template files and read by human experts in foreign languages.

Don't just copy-past this from DeviceToolBar.cpp.  Adapt it meaningfully.

src/toolbars/AudioSetupToolBar.cpp
```diff
@@ -0,0 +1,840 @@
...
+#include "../prefs/DevicePrefs.h"
+#include "../widgets/AButton.h"
+#include "../widgets/AudacityMessageBox.h"
+#include "../widgets/BasicMenu.h"
+#include "../widgets/Grabber.h"
+#include "../widgets/wxWidgetsWindowPlacement.h"
+
+#if wxUSE_ACCESSIBILITY
+#include "../widgets/WindowAccessible.h"
+#endif
```
Paul-Licameli, line 66: Add `#include <wx/log.h>`

You can eliminate
#include <wx/wxprec.h>
#include <wx/setup.h>
#include <wx/app.h>
#include <wx/choice.h>
#include <wx/event.h>
#include <wx/intl.h>
#include <wx/menuitem.h>
#include <wx/settings.h>
#include <wx/statbmp.h>
#include <wx/stattext.h>
#include <wx/string.h>
#include "../TrackPanel.h"
#include "AColor.h"
#include "ImageManipulation.h"
#include "Prefs.h"
#include "Project.h"
#include "../ShuttleGui.h"
#include "../widgets/Grabber.h"
#include "../widgets/AudacityMessageBox.h"
#include "../widgets/Grabber.h"
#include "../widgets/WindowAccessible.h"

src/toolbars/AudioSetupToolBar.cpp
```diff
@@ -0,0 +1,840 @@
...
+END_EVENT_TABLE()
+
+//Standard constructor
+AudioSetupToolBar::AudioSetupToolBar( AudacityProject &project )
+: ToolBar( project, AudioSetupBarID, XO("Audio Setup"), wxT("Audio Setup") )
+{
+   mSubscription = DeviceManager::Instance()->Subscribe(
+      *this, &AudioSetupToolBar::OnRescannedDevices );
+
+   mSizer = nullptr;
```
Paul-Licameli, line 117: Don't assign, when you can use a member initializer.
But don't use a member initializer, when you can use an in-class initializer!

src/toolbars/AudioSetupToolBar.cpp
```diff
@@ -0,0 +1,840 @@
...
+#if wxUSE_TOOLTIPS
+   RegenerateTooltips();
+   wxToolTip::Enable(true);
+   wxToolTip::SetDelay(1000);
+#endif
+
+   // Set default order and mode
+   ArrangeButtons();
+
+   // Set default order and mode
```
Paul-Licameli, line 172: Duplicated comment is unclear here

petersampsonaudacity:
Testing on macOS 12.3.1 Monterey with @ksoze95 Kit's latest build:  audacity-macOS-3.2.0-alpha-20220525+4bdb8f8-x86_64 
https://github.com/audacity/audacity/actions/runs/2383883708

I cannot test non-default changes on Mac as I only have a single option for the I/O devices and Host.

**_BUT_** I note that we still have the regression residual where the default channels is now **1 (mono)** and not **2 (stereo)**

And RIP Mr Coggy - I'm missing him already

petersampsonaudacity:
Testing on W10 with @ksoze95 Kit's latest branch build:  audacity-win-3.2.0-alpha-20220526+15d701b-x64

We still have the regression residual where the default channels is now **1 (mono)** and not **2 (stereo)**.

And now with the Device toolbar hidden/disabled by default it is not immediately obvious that this is the case.

Tantacrul:
@petersampsonaudacity - since this is a separate issue that was not introduced in this PR, @ksoze95 won't be fixing it here. Just so you know!

petersampsonaudacity:
> @Tantacrul since this is a separate issue that was not introduced in this PR, @ksoze95 won't be fixing it here. Just so you know!

But testing with that latest master alpha this is not the case - with that the default is **still 2 (stereo)**

DavidBailes:
The items on the main audio setup menu (host, playback device, etc) should have access keys.

petersampsonaudacity:
> The items on the main audio setup menu (host, playback device, etc) should have access keys.

@DavidBailes they appear to have these when I test @ksoze95 's latest brabch build on W11:  audacity-win-3.2.0-alpha-20220601+f8b50f1-x64

petersampsonaudacity:
> Lack of access keys: image

@DavidBailes Oh I see what you mean now - I was looking at the dialoh=g not the menu:
image

libraries/lib-audio-devices/AudioIOBase.h
```diff
@@ -329,6 +329,7 @@ extern AUDIO_DEVICES_API StringSetting AudioIOHost;
 extern AUDIO_DEVICES_API DoubleSetting AudioIOLatencyCorrection;
 extern AUDIO_DEVICES_API DoubleSetting AudioIOLatencyDuration;
 extern AUDIO_DEVICES_API StringSetting AudioIOPlaybackDevice;
+extern AUDIO_DEVICES_API StringSetting AudioIOPlaybackSource;
```
Paul-Licameli, line 332: The commit that introduced this Setting object (the third commit) did not remove all uses of the path, although that did happen in a later commit.

I like a sequence of commits that make a sensible stepwise transformation of the source code tree.  I might make a small commit just to introduce this setting and all the rewrites.  Then proceed to the next step.

src/toolbars/ToolBar.cpp
```diff
@@ -869,6 +881,44 @@ AButton * ToolBar::MakeButton(wxWindow *parent,
    return button;
 }
 
+// This is a convenience function that allows for button creation in
+// MakeButtons() with fewer arguments
+/// @param parent            Parent window for the button.
+/// @param eEnabledUp        Background for when button is Up.
+/// @param eEnabledDown      Background for when button is Down.
+/// @param eHilite           Background for when button is Hilit.
```
Paul-Licameli, line 889: There is no such parameter!

src/toolbars/ControlToolBar.cpp
```diff
@@ -150,33 +150,6 @@ void ControlToolBar::Create(wxWindow * parent)
    UpdatePrefs();
 }
 
-// This is a convenience function that allows for button creation in
-// MakeButtons() with fewer arguments
-AButton *ControlToolBar::MakeButton(ControlToolBar *pBar,
```
Paul-Licameli, line 155: A separate commit just to move this function into Toolbar.cpp might be nice too.

src/toolbars/AudioSetupToolBar.h
```diff
@@ -0,0 +1,91 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  AudioSetupToolBar.h
```
Paul-Licameli, line 5: One commit adds this file (but does much else besides), another corrects the comment.

I prefer to see a squashed sequence of commits that doesn't leave a record of mistakes and corrections.

src/toolbars/ControlToolBar.cpp
Paul-Licameli: Actually though... you don't really use this overload in the final commit

src/toolbars/AudioSetupToolBar.cpp
```diff
@@ -0,0 +1,856 @@
...
+      {
+         gAudioIO->StopStream();
+         while (gAudioIO->IsBusy()) {
+            using namespace std::chrono;
+            std::this_thread::sleep_for(100ms);
+         }
+      }
+
+      if (audioSettingsChosen) {
+         if (auto project = ::GetActiveProject().lock()) {
```
Paul-Licameli, line 823: New uses of `GetActiveProject` should be avoided!  It dependes on a global variable to identify the project.

But this toolbar was constructed with a reference to the project, which was passed to the bass class ToolBar.

Give ToolBar a const accessor to return AudacityProject& and use that.

Paul-Licameli (changes requested):

src/toolbars/AudioSetupToolBar.cpp
Paul-Licameli: Actually, it's a projecte member, so just use `mProject` directly.

ksoze95: Done.

src/toolbars/AudioSetupToolBar.h
ksoze95: Commit squashed.

Issue #2943, opened 2022-05-17
Statelessness of AudioUnitEffect


Paul-Licameli:

pietro68:
> I can't reproduce the symptom you describe in 2. Is there some more exact step to reproduce you did not mention?

First let me say that actually only 2. is a regression, compared to the start of the branch at `e46d75408`.
This is how I reproduce it:

1) generate some noise
2) instantiate an AUGraphicEQ in the sidebar
3) press play
4) click the cogwheel
5) try moving one of the band sliders

Paul-Licameli:
I'm still not seeing those symptoms.  Maybe this depends on the operating system version.

Paul-Licameli:
> Observed on `42da8c561`
> 
> There are 2 problems when using an AU in the sidebar:
> 
> 1. the AU gui dialog seems to be modal - once brought up by clicking the cogwheel, I can not click the play button in the menu bar. There is no play button in the AU container either.
> 2. even when pressing the play button and then the cogwheel, this happens: as I try to move one of the sliders of AUGraphicEQ, it keeps flickering between the mouse pointer and the initial position; at mouse release, it goes back to the initial position. I can not hear changes in the sound anyway, as the slider is dragged.

Call for help!  Who can reproduce point 2 in this build, which I can’t now?  Does it depend on .cfg files?

Paul-Licameli:
@petersampsonaudacity @Ryan-Miller-Audacity @dozzzzer

dozzzzer:
I've managed to reproduce the issue. In fact there's two of them:

- the flickering issue with the sliders not changing their position on mouse release, sticking to the zero position. It only happens when number of bands is set to 10

STR:

1. remove pluginsettings.cfg and launch Audacity
2. load/generate an audio clip
3. open side panel and select "Apple > AUGraphic EQ"
4. start playback
5. click the gear button in order to bring up the effect's window
6. change the bands amount to "10"
7. try to pull the frequency sliders up and down

Act. result: the sliders jump back to zero position upon mouse release

- the inability to switch between 10/31 bands is only observed when a) playback is active and b) switching from 31 to 10, not the other way around

STR:
1. same as above, remove pluginsettings.cfg and launch Audacity
2. load/generate an audio clip
3. open side panel and select "Apple > AUGraphic EQ"
4. click the gear button
5. in the opened effect's window set number of bands to 10 > confirm by clicking [OK] button 
6. start playback
7. click the gear icon again
9. switch to 31 bands
[...]

PR #2944, opened 2022-05-17
Au statelessness

Paul-Licameli:
Resolves: #2943

Make AudioUnit effects stateless; also some more source file splitting

Recommended:


Paul-Licameli:
I have found a fix for the problems @pietro68 observed with the graphic EQ effect.  Please review again.

pietro68 (approved):
Verified that the last defect I reported was fixed (but please see my last comment on the issue); I think I understand the code and all is fine there for me.

Issue #2948, opened 2022-05-18
Devices Prefs ignores user settings for changes of I/O devices


petersampsonaudacity:
**Describe the bug**
Devices Preferences ignores user settings for changes of input (recording) and output (playback) devices.

**To Reproduce**
1. Clear the Audacity settings folder to get OOTB default factory settings
2. launch 3.2.0
3. Edit > Preferences > Devices
4. Change either the recording or playback device to be a non-default
5. OK to exit prefs
6. Edit > Preferences > Devices
7. **Observe:** the change(s) made at step 4 were ignored by Audacity - and the devices remain as default

**Actual behavior**
Devices Preferences ignores user settings for changes of input (recording) and output (playback) devices.

**Expected behavior**
Devices Preferences should implement  the user's settings for changes of input (recording) and output (playback) devices.

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: W10 - but assume all OS
 - Version audacity-win-3.2.0-alpha-20220516+465d3ee-x64-msvc2022

**Additional context**
This is a regression on 3.1.3 where this works properly as expected.

If you enable the Device Toolbar before Step 4 you will observe that after Step 4 the devices changes made at step 4 do not show in Device Toolbar.

**_BUT_** curiously the reverse is not true, if you make changes in the Device Toolbar these appear to properly ripple through to the Devices Preferences

dozzzzer:
Confirm the bug on MacOS Monterey: the changes made to the Device Preferences menu do not retain.

petersampsonaudacity:
Testing on W10 with @ksoze95 branch build:  audacity-win-3.2.0-alpha-20220519+190cd9e-x64-msvc2022 
https://github.com/audacity/audacity/actions/runs/2352069132

This now looks to work properly again with the fix in this branch build

Paul-Licameli:
> Testing on W10 with @ksoze95 branch build: audacity-win-3.2.0-alpha-20220519+190cd9e-x64-msvc2022 https://github.com/audacity/audacity/actions/runs/2352069132
> 
> This now looks to work properly again with the fix in this branch build

I have reviewed Kit's fix and I'm not satisfied.  It is an effective special case fix for these symptoms, but there is a more general case problem that is the fault of code I wrote (and Kit reviewed!).  I'll discuss with him, and you can wait for the alternative.

petersampsonaudacity:
[...]

PR #2952, opened 2022-05-19
Bugfix: #2948.  Persist Devices preference changes.

ksoze95:
Resolves: #2948 

Recommended:

### DIFF
--- a/src/prefs/DevicePrefs.cpp
+++ b/src/prefs/DevicePrefs.cpp
@@ -395,10 +395,8 @@ bool DevicePrefs::Commit()
       map = (DeviceSourceMap *) mPlay->GetClientData(
             mPlay->GetSelection());
    }
-   if (map) {
+   if (map)
       AudioIOPlaybackDevice.Write(map->deviceString);
-      AudioIOPlaybackDevice.Commit();
-   }
 
    map = NULL;
    if (mRecord->GetCount() > 0) {
@@ -412,11 +410,6 @@ bool DevicePrefs::Commit()
       else
          AudioIORecordingSource.Reset();
       AudioIORecordChannels.Write(mChannels->GetSelection() + 1);
-
-      AudioIORecordingDevice.Commit();
-      AudioIORecordingSourceIndex.Commit();
-      AudioIORecordingSource.Commit();
-      AudioIORecordChannels.Commit();
    }
 
    return true;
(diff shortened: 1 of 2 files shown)


src/prefs/DevicePrefs.cpp
```diff
@@ -410,6 +412,11 @@ bool DevicePrefs::Commit()
       else
          AudioIORecordingSource.Reset();
       AudioIORecordChannels.Write(mChannels->GetSelection() + 1);
+
+      AudioIORecordingDevice.Commit();
+      AudioIORecordingSourceIndex.Commit();
+      AudioIORecordingSource.Commit();
+      AudioIORecordChannels.Commit();
```
Paul-Licameli, line 419: I am not willing to accept this very special case fix of what might be an error of commit d9a873a that introduced SettingsTransaction.  It was not intended that individual settings variables should require commits like this.

Do you instead have an explanation why the line `mTransaction->Commit();` in PrefsDialog.cpp fails to have the effect it should?

Can you investigate, verify that d9a873a is the first bad commit in 3.2 showing this bug?

Paul-Licameli (changes requested):

Paul-Licameli:
Kit:  I was bothered enough that I debugged this a bit myself.

`AudioIOPlaybackDevice` is written in `DevicePrefs::Commit` and the later `mTransaction->Commit()` in `PrefsDialog::OnOk` should persist the setting but doesn't -- why not?

Because in between, there is re-reading of `AudioIOPlaybackDevice` that destroys the un-committed change, and that's a more general bug in Prefs.h that should be fixed.

See that there are two overloads of `ReadWithDefault`.  One of them is correct in case the settings object already has a value, but the other is incorrect.

Paul-Licameli (approved):
Very simple!

petersampsonaudacity:
Testing on W10 with @ksoze95 's latest branch build:  audacity-win-3.2.0-alpha-20220522+2d8ee49-x64-msvc2022 

This is **_not_** working properly:


See  Devices Prefs ignores user settings for changes of I/O devices #2948  for my test results

Paul-Licameli:
I'm closing this PR and writing a different fix that addresses more fundamental errors in the Preferences system and doesn't contain special case code to write one of the settings.

Paul-Licameli:
I'm closing this PR and writing a different fix that addresses more fundamental errors in the Preferences system and doesn't contain special case code to write one of the settings.

PR #2953, opened 2022-05-19
Plugin validation fixes

vsverchinsky:
Resolves: 

1. Startup plugin validation freeze
2. "New" radio button was available if `wxUSE_ACCESSIBILITY` was not defined

Recommended:

### DIFF
--- a/libraries/lib-module-manager/AsyncPluginValidator.cpp
+++ b/libraries/lib-module-manager/AsyncPluginValidator.cpp
@@ -57,9 +57,8 @@ class AsyncPluginValidator::Impl final :
    {
       auto server = std::make_unique<IPCServer>(*this);
       if(!PluginHost::Start())
-         mDelegate->OnInternalError("Cannot start plugin host");
-      else
-         mServer = std::move(server);
+         throw std::runtime_error("cannot start plugin host process");
+      mServer = std::move(server);
    }
    
    void HandleInternalError(const wxString& msg) noexcept
--- a/src/PluginStartupRegistration.cpp
+++ b/src/PluginStartupRegistration.cpp
@@ -18,6 +18,7 @@
 #include "PluginManager.h"
 #include "PluginDescriptor.h"
 #include "widgets/ProgressDialog.h"
+#include "widgets/wxWidgetsWindowPlacement.h"
 
 
 PluginStartupRegistration::PluginStartupRegistration(std::vector<std::pair<wxString, wxString>> pluginsToProcess)
@@ -28,8 +29,7 @@ PluginStartupRegistration::PluginStartupRegistration(std::vector<std::pair<wxStr
 
 void PluginStartupRegistration::OnInternalError(const wxString& error)
 {
-   wxLogError(error);
-   Stop();
+   StopWithError(error);
 }
 
 void PluginStartupRegistration::OnPluginFound(const PluginDescriptor& desc)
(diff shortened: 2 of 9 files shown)


libraries/lib-module-manager/AsyncPluginValidator.cpp
```diff
@@ -57,9 +57,8 @@ class AsyncPluginValidator::Impl final :
    {
       auto server = std::make_unique<IPCServer>(*this);
       if(!PluginHost::Start())
-         mDelegate->OnInternalError("Cannot start plugin host");
-      else
-         mServer = std::move(server);
+         throw std::runtime_error("cannot start plugin host process");
```
Paul-Licameli, line 60: This possibly thrown exception is not caught anywhere and might terminate the program.  One remedy would be, to move initialization of PluginStartupRegistration::mValidator out of the constructor, and into the try-catch block in Run().

src/PluginStartupRegistration.cpp
```diff
@@ -52,23 +53,32 @@ const std::vector<wxString>& PluginStartupRegistration::GetFailedPlugins() const
...
+         //delivery, but ProgressDialog::Poll implementation does not call
+         //wxApp::Yield each time, which may result in too long CPU stalls
+         if(wxTheApp->HasPendingEvents())
+            wxTheApp->Yield();
+         else
+            //Seems like we do have no events yet, save CPU cycles then
+            std::this_thread::sleep_for(std::chrono::milliseconds { 10 });
+      }
+   }
+   catch(std::exception& e)
```
Paul-Licameli, line 76: Was this catch meant for the new throw?  But it does not enclose it.

Paul-Licameli (changes requested):
try/catch is misplaced for the new exception

libraries/lib-module-manager/AsyncPluginValidator.cpp
vsverchinsky: StartHost isn't called from the constructor

src/PluginStartupRegistration.cpp
vsverchinsky: I don't see any other places from where StartHost could be called

Paul-Licameli:
I was mistaken to trace the call graph of the constructor instead of StartHost.  However, is this an example of a call stack that could leave a potential exception uncaught, in one of the worker threads of `BufferedIPCChannel`?

```
Impl::StartHost
Impl::Validate
AsyncPluginValidator::Validate
PluginStartupRegistration::ProcessNext
PluginStartupRegistration::OnValidationFinished
AsyncPluginValidator::Impl::HandleResult
AsyncPluginValidator::Impl::OnDataAvailable or OnDisconnect
thread function spawned in BufferedIPCChannel::StartConversation
```

So, I think one more try-catch is needed to be sure exceptions from `Validate` are stopped.  It could perhaps be in `OnValidationFinished`

vsverchinsky:
> So, I think one more try-catch is needed to be sure exceptions from `Validate` are stopped. It could perhaps be in `OnValidationFinished`

Where? All calls to AsyncPluginValidator::Delegate are dispatched to the main thread

Paul-Licameli:
> > So, I think one more try-catch is needed to be sure exceptions from `Validate` are stopped. It could perhaps be in `OnValidationFinished`
> 
> Where? All calls to AsyncPluginValidator::Delegate are dispatched to the main thread

Do you mean the function `AsyncPluginValidator::Validate` ?
I see it is called by `PluginStartupRegistration::ProcessNext`, and that is called in two places, but only one is inside a try block.  The other is in `PluginStartupRegistration::OnValidationFinished`

Following the call stack up from there -- I was mistaken, what calls that is a lambda that the other thread enqueues for the main thread to do, using CallAfter.  Still:  in the main thread, there is this other place needing a try-catch to be sure the exception can't terminate the program.

vsverchinsky:
> > > So, I think one more try-catch is needed to be sure exceptions from `Validate` are stopped. It could perhaps be in `OnValidationFinished`
> > 
> > 
> > Where? All calls to AsyncPluginValidator::Delegate are dispatched to the main thread
> 
> Do you mean the function `AsyncPluginValidator::Validate` ? I see it is called by `PluginStartupRegistration::ProcessNext`, and that is called in two places, but only one is inside a try block. The other is in `PluginStartupRegistration::OnValidationFinished`
> 
> Following the call stack up from there -- I was mistaken, what calls that is a lambda that the other thread enqueues for the main thread to do, using CallAfter. Still: in the main thread, there is this other place needing a try-catch to be sure the exception can't terminate the program.

The main thread execution happens inside `while` loop

Paul-Licameli (approved):

Paul-Licameli:
I see and approve the latest rewrite, which makes one narrow-scope try-catch block at a lower level, sufficient to catch all the exceptions.

Now I ask you to review the suggested extra commit from me and merge it, if it helps to avoid traps in the debug build.

You may also notice -- the added use of `Translation()` suggests that plain-UI dialogs may fail to update a localized string in case the user happens to change the language preference.  That is true; that is a common problem with plain VST and Ladspa dialogs; that could also be fixed as in commit 20ace31; but, it is an unlikely use case and is beside the point of the fix required just now.

PR #2957, opened 2022-05-20
Stateless vst3 settings

pietro68:
Resolves: part of https://github.com/audacity/audacity/issues/2969

I am still trying to find my way around the code, so this PR will be a draft.
I defined settings, and operations to store/fetch them, somehow like it was done for AudioUnits.

All points where settings are potentially sent back to the framework are covered, i.e. settings will be fetched and sent back to the framework. This is only one half of the work of course, the next half would be to cover all spots where settings enter from the framework and should be used.

In the process of doing the above, I think I found what looks like a pre-existing bug (probably resulting by copy-paste, please expand the context around the diff to see why I think so) and fixed it in a specific commit - please confirm that either it was indeed a bug, or if not, we'd have to talk about it.

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -343,6 +343,68 @@ bool VST3Effect::SupportsAutomation() const
    return false;
 }
 
+
+bool VST3Effect::FetchSettings(VST3EffectSettings& settings) const
+{
+   // partly copied from ::SaveSettings + some adaptations made
+
+   if (mEditController == nullptr)
+      return false;
+
+   using namespace Steinberg;
+
+   auto theCount = mEditController->getParameterCount();
+
+   for (int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
+   {
+      Vst::ParameterInfo parameterInfo{ };
+      if (mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+      {
+         if (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
+         {
+
+            Steinberg::Vst::ParamID    key = parameterInfo.id;
+            Steinberg::Vst::ParamValue value = mEditController->getParamNormalized(parameterInfo.id);
+
+            settings.mValues[key] = value;
+         }
+      }
+   }
+
+   return true;
+}
+
+
+bool VST3Effect::StoreSettings(const VST3EffectSettings& settings) const
+{
+   // partly copied from ::LoadSettings + some adaptations made
+
+   using namespace Steinberg;
+
+   if (mComponentHandler == nullptr)
+      return false;
+
+   for (const auto& item : settings.mValues)
+   {
+      const Steinberg::Vst::ParamID& id    = item.first;
+      const Vst::ParamValue&         value = item.second;
+
+      if (mComponentHandler->beginEdit(id) == kResultOk)
+      {
+         auto cleanup = finally([&] {
+            mComponentHandler->endEdit(id);
+         });
+         mComponentHandler->performEdit(id, value);
+      }
+      mEditController->setParamNormalized(id, value);
+   }
+
+   return true;
+}
+
+
+
+
 bool VST3Effect::SaveSettings(
    const EffectSettings &, CommandParameters & parms) const
 {
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -343,92 +343,117 @@ bool VST3Effect::SupportsAutomation() const
    return false;
 }
 
-
-bool VST3Effect::FetchSettings(VST3EffectSettings& settings) const
+struct VST3Effect::ParameterInfo
 {
-   // partly copied from ::SaveSettings + some adaptations made
+   Steinberg::Vst::ParameterInfo mVstParamInfo;
+};
 
+
+bool VST3Effect::ForEachParameter(ParameterVisitor visitor) const
+{
    if (mEditController == nullptr)
       return false;
 
-   using namespace Steinberg;
-
-   auto theCount = mEditController->getParameterCount();
-
    for (int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
    {
-      Vst::ParameterInfo parameterInfo{ };
-      if (mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+      ParameterInfo pi;
+
+      Steinberg::Vst::ParameterInfo vstParameterInfo{ };
+      if (mEditController->getParameterInfo(i, vstParameterInfo) == Steinberg::kResultOk)
       {
-         if (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
+         if (vstParameterInfo.flags & Steinberg::Vst::ParameterInfo::kCanAutomate)
          {
-
-            Steinberg::Vst::ParamID    key = parameterInfo.id;
-            Steinberg::Vst::ParamValue value = mEditController->getParamNormalized(parameterInfo.id);
-
-            settings.mValues[key] = value;
+            if (!visitor(ParameterInfo{ vstParameterInfo } ))
+               return false;
          }
+         else
+         {
+            return false;
+         }
+      }
+      else
+      {
+         return false;
       }
    }
 
    return true;
+}
+
+
+bool VST3Effect::FetchSettings(VST3EffectSettings& settings) const
+{
+   return ForEachParameter(
+
+      [&](const ParameterInfo& pi)
+      {
+        const auto id = pi.mVstParamInfo.id;
+        settings.mValues[id] = mEditController->getParamNormalized(id);
+        return true;
+      }
+   );
 }
 
 
 bool VST3Effect::StoreSettings(const VST3EffectSettings& settings) const
 {
-   // partly copied from ::LoadSettings + some adaptations made
+   return ForEachParameter(
 
-   using namespace Steinberg;
-
-   if (mComponentHandler == nullptr)
-      return false;
-
-   for (const auto& item : settings.mValues)
-   {
-      const Steinberg::Vst::ParamID& id    = item.first;
-      const Vst::ParamValue&         value = item.second;
-
-      if (mComponentHandler->beginEdit(id) == kResultOk)
+      [&](const ParameterInfo& pi)
       {
-         auto cleanup = finally([&] {
-            mComponentHandler->endEdit(id);
-         });
-         mComponentHandler->performEdit(id, value);
+         if (mComponentHandler == nullptr)
+            return false;
+
+         const auto id = pi.mVstParamInfo.id;
+
+         auto itr = settings.mValues.find(id);
+         if (itr != settings.mValues.end())
+         {
+            const auto& value = itr->second;
+
+            // set the value to the handler at the id
+            if (mComponentHandler->beginEdit(id) == Steinberg::kResultOk)
+            {
+               auto cleanup = finally([&] {
+                  mComponentHandler->endEdit(id);
+               });
+               mComponentHandler->performEdit(id, value);
+            }
+            else
+            {
+               return false;
+            }
+            mEditController->setParamNormalized(id, value);
+
+         }
+         else
+         {
+            return false;
+         }
+
+         return true;
       }
-      mEditController->setParamNormalized(id, value);
-   }
+   );
 
-   return true;
 }
-
-
 
 
 bool VST3Effect::SaveSettings(
    const EffectSettings &, CommandParameters & parms) const
 {
-   if(mEditController == nullptr)
-      return false;
+   return ForEachParameter(
 
-   using namespace Steinberg;
-   
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+      [&](const ParameterInfo& pi)
       {
-         if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
-         {
-            parms.Write(
-               VST3Utils::MakeAutomationParameterKey(parameterInfo),
-               mEditController->getParamNormalized(parameterInfo.id)
-            );
-         }
-      }
-   }
+         parms.Write(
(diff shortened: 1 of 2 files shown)


src/effects/VST3/VST3Effect.h
```diff
@@ -38,6 +38,12 @@ namespace Steinberg
 
 class ParameterChangesProvider;
 
+
+struct VST3EffectSettings
+{
+   std::unordered_map<Steinberg::Vst::ParamID, Steinberg::Vst::ParamValue> mValues;
```
Paul-Licameli, line 44: I need to add `#include <unordered_map>` to build it

src/effects/VST3/VST3Effect.cpp
```diff
@@ -416,7 +490,7 @@ bool VST3Effect::LoadUserPreset(
    if(mEffectComponent->setState(processorState) != kResultOk)
       return false;
 
-   if(PluginSettings::HasConfigValue(*this, PluginSettings::Private, name, wxT("ProcessorState")))
+   if(PluginSettings::HasConfigValue(*this, PluginSettings::Private, name, wxT("ControllerState")))
```
Paul-Licameli, line 493: This looks like a correct fix to me.  @vsverchinsky ?

But let's agree to go farther: write
```
constextpr auto ProcessorState = L"ProcessorState";
```
and eliminate the repetition of the string literal, using the named constant instead in three places.  Do likewise for the other string constant.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -339,6 +339,68 @@ bool VST3Effect::SupportsAutomation() const
...
+bool VST3Effect::StoreSettings(const VST3EffectSettings& settings) const
+{
+   // partly copied from ::LoadSettings + some adaptations made
+
+   using namespace Steinberg;
+
+   if (mComponentHandler == nullptr)
+      return false;
+
+   for (const auto& item : settings.mValues)
```
Paul-Licameli, line 383: These functions to fetch settings from the VST3 handle into our own map, and store them back, are a beginning toward statelessness by analogy with the transformatio I am making for AudioUnits.

For AudioUnits, the fetching and storing both perform a loop over the settings as defined by the handle.  StoreSettings doesn't iterate our own hash map, but looks up in it.

Also in AudioUnits, the details of how to iterate the handle's metadata for settings is written just once in a higher-order function called `ForEachParameter`, and that function takes a lambda for the loop body.

I suggest you make similar changes.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -365,6 +427,14 @@ bool VST3Effect::SaveSettings(
    return true;
 }
 
+
+EffectSettings VST3Effect::MakeSettings() const
+{
+   auto result = EffectSettings::Make<VST3EffectSettings>();
+   return result;
+}
```
Paul-Licameli, line 435: If you define `MakeSettings`, also define an efficient `CopySettingsContents`.

But make those definitions only very late, when all else is ready for the final switch to statelessness.  So, not in this first stage PR.

src/effects/VST3/VST3Effect.h
```diff
@@ -179,4 +181,12 @@ class VST3Effect final : public StatefulPerTrackEffect
    bool FetchSettings(VST3EffectSettings& settings) const;
 
    bool StoreSettings(const VST3EffectSettings& settings) const;
+
+   static inline VST3EffectSettings& GetSettings(EffectSettings& settings)
```
Paul-Licameli, line 185: Do not make `GetSettings` a static function, until the last step to real statelessness.

Make it a non-static member function of  VST3Effect that ignores its argument and uses a VST3EffectSettings member of the VST3Effect, which is still stateful.

This would parallel what I did with AudioUnits.

This allows some changes to use `GetSettings` where that will be needed, while not yet changing real behavior, when the transformation to statelessness is still incomplete.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -395,11 +465,15 @@ bool VST3Effect::LoadSettings(
       } while(parms.GetNextEntry(key, index));
    }
 
+   VST3EffectSettings specificSettings;
+   FetchSettings(specificSettings);
+   GetSettings(settings) = specificSettings;
```
Paul-Licameli, line 470: Here and elsewhere, you might write just one line
`FetchSettings(GetSettings(settings))`

Paul-Licameli: But then, also, FetchSettings should clear the map before iterating the parameters and reinserting them.

Paul-Licameli: `FetchSettings` is called in all necessary places, good

Paul-Licameli:
I have made some suggestions for change.  Consider this round one of review.

I see this draft pull request is incomplete.  `StoreSettings` is defined but not yet used.

To complete the analogy with the first phase of statelessness for AudioUnits, we should also see SaveSettings and LoadSettings use only the intermediate VST3EffectSettings.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -478,7 +478,7 @@ bool VST3Effect::LoadUserPreset(
    if(mEffectComponent->setState(processorState) != kResultOk)
       return false;
 
-   if(PluginSettings::HasConfigValue(*this, PluginSettings::Private, name, wxT("ProcessorState")))
+   if(PluginSettings::HasConfigValue(*this, PluginSettings::Private, name, wxT("ControllerState")))
```
vsverchinsky: Looks like a bug indeed

src/effects/VST3/VST3Effect.cpp
```diff
@@ -339,6 +339,68 @@ bool VST3Effect::SupportsAutomation() const
...
+      Vst::ParameterInfo parameterInfo{ };
+      if (mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+      {
+         if (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
+         {
+
+            Steinberg::Vst::ParamID    key = parameterInfo.id;
+            Steinberg::Vst::ParamValue value = mEditController->getParamNormalized(parameterInfo.id);
+
+            settings.mValues[key] = value;
```
vsverchinsky: Is that possible to store pointer to the controller instead?

src/effects/VST3/VST3Effect.cpp
```diff
@@ -339,6 +339,68 @@ bool VST3Effect::SupportsAutomation() const
...
+   using namespace Steinberg;
+
+   auto theCount = mEditController->getParameterCount();
+
+   for (int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
+   {
+      Vst::ParameterInfo parameterInfo{ };
+      if (mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+      {
+         if (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
```
vsverchinsky: Parameters stored by `SaveSettings` are used by macroses in Audacity and are meant to be read by human, I'm not sure if that was correct to skip other types of parameters, but this was an attempt to restrict the amount of parameters (there could be thousands of them) to a set of paramters that are useful and meaningful

src/effects/VST3/VST3Effect.cpp
pietro68: ok thanks - I wonder if we should correct this in a separate, trivial PR so that it goes in master soon?

src/effects/VST3/VST3Effect.cpp
vsverchinsky: Yes, instead of storing parameter values in `settings` use `settings` object to access underlying `EditController`

src/effects/VST3/VST3Effect.cpp
vsverchinsky: Agree, @Paul-Licameli

Paul-Licameli: Agreed, write it, also eliminating the repetition of string literals.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -339,32 +339,129 @@ bool VST3Effect::SupportsAutomation() const
    return false;
 }
 
-bool VST3Effect::SaveSettings(
-   const EffectSettings &, CommandParameters & parms) const
+struct VST3Effect::ParameterInfo
 {
-   if(mEditController == nullptr)
+   Steinberg::Vst::ParameterInfo mVstParamInfo;
+};
```
Paul-Licameli, line 345: Maybe this `struct` should just simplify to a `using` alias

src/effects/VST3/VST3Effect.cpp
```diff
@@ -339,32 +339,129 @@ bool VST3Effect::SupportsAutomation() const
...
       {
-         if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
+         if (vstParameterInfo.flags & Steinberg::Vst::ParameterInfo::kCanAutomate)
          {
-            parms.Write(
-               VST3Utils::MakeAutomationParameterKey(parameterInfo),
-               mEditController->getParamNormalized(parameterInfo.id)
-            );
+            if (!visitor(ParameterInfo{ vstParameterInfo } ))
+               return false;
```
Paul-Licameli, line 363: This is not analogous to what I did for AudioUnits:  the visitor function can return false, not to cause `ForEachParameter` to fail, but only to cause it to stop visiting other parameters.

In fact `ForEachParameter` for AudioUnits ended up as a void-returning function, and yours might too.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -339,32 +339,129 @@ bool VST3Effect::SupportsAutomation() const
...
          }
+         else
+         {
+            return false;
+         }
+      }
+      else
+      {
+         return false;
```
Paul-Licameli, line 372: As @vsverchinsky explained, skip over certain parameters that don't automate; but don't stop the visit or return a failure code.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -339,32 +339,129 @@ bool VST3Effect::SupportsAutomation() const
...
-bool VST3Effect::SaveSettings(
-   const EffectSettings &, CommandParameters & parms) const
+struct VST3Effect::ParameterInfo
 {
-   if(mEditController == nullptr)
+   Steinberg::Vst::ParameterInfo mVstParamInfo;
+};
+
+
+bool VST3Effect::ForEachParameter(ParameterVisitor visitor) const
```
Paul-Licameli, line 348: This function could also be used to simplify `SupportsAutomation` and `LoadFactoryDefaults`

As for `LoadSettings`, change it to use `ForEachParameter`, and not iterating over `parms`.  Instead call `parms.Read` as in `AudioUnitEffect::LoadSettings`

src/effects/VST3/VST3Effect.cpp
Paul-Licameli: Can we assume that the set of parameters, and the kCanAutomate flag for each parameter, is unchanging during the lifetime of the loaded VST3 plug-in?  And independent of any instance state?  I would assume so.

When you write CopySettingsContents analogous to that for AudioUnits (not yet merged, but in the last of the sequence of PRs for AU), you may be making such an assumption.

src/effects/VST3/VST3Effect.cpp
Paul-Licameli: I'm missing some context.  These lines are no longer in the last commit of this branch.

Are you suggesting that a pointer to `EditController` should be carried around in the `EffectSettings` object?  I don't like that.  `EffectSettings` should be thought of as a "frozen" description of instance state, not using any third-party library handles.  Thus our fetching and storing functions to synchronize frozen with live state, one way or the other.

The `EffectSettings` is something we can pass across inter-thread boundaries, and also copy from one live instance to the other by means of fetching from one and storing to the other.

Maybe it is not the most efficient way for every effect protocol, but it seems to me to have the generality that we need that would work with all of them in one unifying framework.

Paul-Licameli:
None of the boxes in https://github.com/audacity/audacity/issues/2969 are yet ticked off, and some of these small commits are fixes of previous ones.

I think it's time for an interative rebase onto latest master, and also that the sequence of commits should address the checkboxes in the suggested sequence.

And as discussed, let's make a small commit before any others, to fix the error in registry paths that Pietro identified.

src/effects/VST3/VST3Effect.h
pietro68: now added in commit `created ForEachParameters and used in 3 places`

src/effects/VST3/VST3Effect.h
pietro68: done in commit `GetSettings turned to temporary version`

src/effects/VST3/VST3Effect.cpp
pietro68: done in commit `simplified parameterInfo`

src/effects/VST3/VST3Effect.cpp
pietro68: done in commit `correct the behavior of ForEachParameter...`

src/effects/VST3/VST3Effect.cpp
pietro68: I am not sure about using it for `SupportsAutomation`, because `ForEachParameter` will always stop at the first parameter where the visitor fails, and return true - while `SupportsAutomation` stops at the first parameter which can be automated, returning true, but if none of the parameters can be automated, it will return false.

I think I'd need a different version of `ForEachParameter`, which will return true at the first successful visit and false if no successful visits happened (it could be named something like VerifyAtLeastOneParameter)

pietro68: As for `LoadSettings`, I need to have a talk about that first

pietro68: I prepared a version of `LoadSettings` that uses `ForEachParameter`, but it is deactivated (under #if 0) in favor of the existing version - this because I can not find a way to reach this method (I tried using macros, as hinted in our conversation, but the method called when playing back the macro will not take the else branch that calls LoadSettings). When I can rebase on Vitaly's recent work for stacks, I will be able to reach this method and test if the new version works.

pietro68:
This PR has been superseded by https://github.com/audacity/audacity/pull/3009 too, so I am closing it.

Issue #2955, opened 2022-05-20
No warning message when failed to enable a VST in Plug-in Manager


dozzzzer:
**Describe the bug**
For some VSTs, when Audacity fails to enable them, there's no warning message shown.

**To Reproduce**
1. Install a VST that will fail to load (e.g. https://plugins4free.com/plugin/447/)
2. Delete pluginregistry.cfg and run Audacity
3. Click [Continue] within Incompatible Plug-ins window
4. Go to Effect > Manage Plug-Ins 
5. Find that VST (marked as "Disabled"), select it and click [Enable]
6. click [OK]

**Actual behavior**
The plugin wasn't enabled; no warning message was shown either.

**Expected behavior**
If the VST has failed to initialize, there's a proper error message shown.

**Screenshots**
An example of the error message pop-up:
<img width="292" alt="image" src="https://user-images.githubusercontent.com/70063970/169549454-9c670017-3126-466f-a7e5-de3a0d05109b.png">

**Additional information (please complete the following information):**
 - OS:  Mac Monterey
 - Version: [audacity-macOS-3.2.0-alpha-20220518+a4b3f8e-universal](https://github.com/audacity/audacity/suites/6556680726/artifacts/245137100)

Was discovered during the testing of #2608.

PR #2973, opened 2022-05-24
Issue 2955: vst fixes

vsverchinsky:
Resolves: #2955 

Fix: invalid plugin path formatting for plugins which contain only one plugin (regression)
Fix: error message was not assigned if plugin module loading has failed (regression)

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -360,25 +360,27 @@ unsigned VSTEffectsModule::DiscoverPluginsAtPath(
    const PluginPath & path, TranslatableString &errMsg,
    const RegistrationCallback &callback)
 {
+
    VSTEffect effect(path);
    if(effect.InitializePlugin())
    {
-      const auto effectIDs = effect.GetEffectIDs();
-      if(!effectIDs.empty())
-      {
-         for(auto id : effectIDs)
-         {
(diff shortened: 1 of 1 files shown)


Paul-Licameli (approved):
I undestand why these changes make sense.  Are they fixing only what the issue describes?  Or were there also examples of VST 2 effects that incorrectly failed to load -- and now they do load?

When a failure message is correct, is this new translatable string actually seen by the user?  My recollection of review of plugin validation at startup, is that in fact these messages only go to the log in English.

vsverchinsky:
> Are they fixing only what the issue describes?

Yes

>Or were there also examples of VST 2 effects that incorrectly failed to load -- and now they do load? When a failure message is correct, is this new translatable string actually seen by the user?

Faulty plugin didn't load in both cases, but without this fix there was no error message when you attempted to enable such plugin manually via "Add/Remove plugins..."

Paul-Licameli:
> > Are they fixing only what the issue describes?
> 
> Yes
> 
> > Or were there also examples of VST 2 effects that incorrectly failed to load -- and now they do load? When a failure message is correct, is this new translatable string actually seen by the user?
> 
> Faulty plugin didn't load in both cases, but without this fix there was no error message when you attempted to enable such plugin manually via "Add/Remove plugins..."

I don't fully understand the chain of causes that now leads to a proper error message -- I'm satisfied if you say it works.

PR #2975, opened 2022-05-25
Bug in vst3 LoadUserPreset

pietro68:
Resolves: a bug which was found while converting vst3 to be stateless

A wrong string key was used. While replacing with the right one, centralized constexpr definition of the string keys were also done.

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -53,6 +53,10 @@
 
 namespace {
 
+   // define some shared registry keys
+   constexpr auto processorStateKey  = wxT("ProcessorState");
+   constexpr auto controllerStateKey = wxT("ControllerState");
+
 Steinberg::Vst::IHostApplication& LocalContext()
 {
    static Steinberg::Vst::HostApplication localContext;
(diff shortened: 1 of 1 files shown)


src/effects/VST3/VST3Effect.cpp
```diff
@@ -406,20 +406,23 @@ bool VST3Effect::LoadUserPreset(
...
    auto processorState = PresetsBufferStream::fromString(processorStateStr);
    if(mEffectComponent->setState(processorState) != kResultOk)
       return false;
 
-   if(PluginSettings::HasConfigValue(*this, PluginSettings::Private, name, wxT("ProcessorState")))
+   if(PluginSettings::HasConfigValue(*this, PluginSettings::Private, name, controllerStateKey))
    {
       wxString controllerStateStr;
-      if(!GetConfig(*this, PluginSettings::Private, name, wxT("ControllerState"), controllerStateStr, wxEmptyString))
+      if(!GetConfig(*this, PluginSettings::Private, name, controllerStateKey, controllerStateStr, wxEmptyString))
```
Paul-Licameli, line 425: The elimination of duplicated string constants is incomplete.  The new names should also be used in `SaveUserPreset`.  That can help increase confidence that there is not a mismatch between the save procedure, which left the data in the config file for the load procedure to interpret here.

Paul-Licameli (changes requested):

Paul-Licameli (approved):

Paul-Licameli:
Please squash and merge all as one small commit.

vsverchinsky (approved):

Issue #2985, opened 2022-05-26
Effect settings changes in the stackable effects sidebar don't stick


Paul-Licameli:
It is important to test this both while playing and not playing.

While playing, the sound effects may not be heard:  that's a different issue #2986

- make an audio track
- Add effects; add (e.g.) Reverb
- click gearwheel; make some changes of settings; OK
- Observe: a new undo state in history (as should be)
- Click gearwheel again

Expect to see the settings you last entered; instead, you still see the defaults.  BUT IF INSTEAD:

- ditto
- ditto
- click gearwheel; make some changes of settings; Esc to cancel
- click gearwheel

THEN -- be sure the change does NOT stick.

When the change really sticks, also expect that you can undo, click the wheel, see defaults; redo, click to gear, see changes.

Issue #2871, opened 2022-04-29
Safety and correctness of effect stack edits while playing back


Paul-Licameli:
- [ ] #3027

Another question to ask is whether all these things requiring pushes and pops of the undo stack are safe to do while playback is in progress.

I think we surely want that -- one hope is that you listen as you change settings of a single effect, or even change the composition of the effect list.  But now I doubt whether all is set up correctly in the audio engine for that, both to avoid dangling pointers and crashes, and to communicate changes of the sequence of effects properly inter-thread.

The limited thing that I know was anticipated by Leland's work, is the powering on and off of individual effects, which is what the suspending and resuming of states is for.

And the point of the EffectSettingsAccess class I developed is to communicate changes of settings of an individual effect correctly -- what this piece of code here will do.

But adding, deleting, and permuting effects?  Or changing the whole stack at one with undo and redo?

I must think.

_Originally posted by @Paul-Licameli in https://github.com/audacity/audacity/pull/2867#discussion_r861433554_

dozzzzer:
Tested on MacOS 12.3.1. and Windows 10.

PR #2988, opened 2022-05-26
Make settings changes stick

Paul-Licameli:
Resolves: #2985
Resolves: part of #2871

Make changes of per-track effect states from the sidebar really stick.

Also addressing some, but not all, known race conditions in managing the lists of realtime effect states:  races involving
the collection of states in the list, but not yet, certain races involving members of those states and of the RelatimeEffectList.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -80,9 +80,16 @@ public:
    //! @return value is not negative
    double GetDuration() const { return mDuration; }
    void SetDuration(double value) { mDuration = std::max(0.0, value); }
+
+   //! Versioning counter for detecting echo from worker thread;
+   //! it does not need a large range of values
+   using Counter = unsigned char;
+   Counter GetCounter() const { return mCounter; }
+   void SetCounter(Counter value) { mCounter = value; }
 private:
    NumericFormatSymbol mDurationFormat{};
    double mDuration{}; //!< @invariant non-negative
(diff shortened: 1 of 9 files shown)


libraries/lib-components/EffectInterface.h
```diff
@@ -80,9 +80,16 @@ class COMPONENTS_API EffectSettingsExtra final {
...
+
+   //! Versioning counter for detecting echo from worker thread;
+   //! it does not need a large range of values
+   using Counter = unsigned char;
+   Counter GetCounter() const { return mCounter; }
+   void SetCounter(Counter value) { mCounter = value; }
 private:
    NumericFormatSymbol mDurationFormat{};
    double mDuration{}; //!< @invariant non-negative
+   Counter mCounter{ 0 };
```
vsverchinsky, line 92: The assumption is that it's unlikely that there will be 255 calls to `Set` between two `Get`'s?

vsverchinsky (approved):
Probably I need to catch up on overall progress in EffectSettings to completely understand this PR

libraries/lib-components/EffectInterface.h
Paul-Licameli: Yes:  the point of this commit is something like a request/response alternation between the main and worker thread.

Paul-Licameli: The counter might even be just one bit

Issue #2555, opened 2022-02-14
VST3 plain gui: serious problems with a number of VST3 plugins


Paul-Licameli:
@DavidBailes wrote:

I've had a look at the plain gui, and have found serious problems using a couple of the free plugins from here:
https://www.voxengo.com/group/free-vst-plugin-download/

I downloaded and installed both the vst and vst3 versions of the Mastering Graphic Equalizer, and the Tube Amplifier.
For the VST versions, the existing plain gui was OK.
However for the VST3 versions the plain gui is definitely not OK:
1. Values not in the right range - for example in the Graphic equalizer, the sliders in the normal gui have ranges -12 to 12, but the equivalent spin control in the plain gui has range 0 to 1.
2. A wxSpinCtrlDouble can be used for a property which has discrete values. In the normal view of the tube amplifier, there's a mode which can be set to be mode 1 or mode 2. In the plain gui, there's a wxSpinCtrlDouble for the mode, with a range 0 to 1.0

Checked the VST3 versions in Reaper, and the plain gui version, which is just a series of sliders and values, does not have the first isssue above - the values are in the correct range.

There are also accessibility isssues, but it probably best to address them after above problems have been sorted out.

_Originally posted by @DavidBailes in https://github.com/audacity/audacity/issues/2295#issuecomment-1038980633_

DavidBailes:
Here's another vst3 plugin, for which the plain gui has serious problems (ie it's unusable):
https://www.kvraudio.com/product/valhallafreqecho-by-valhalla-dsp
I installed both the vst and vst3 versions:

Audacity's normal gui for the vst3 version:
![image](https://user-images.githubusercontent.com/11887021/154081648-bf6f3594-b8db-49ac-b9cd-5c1c2c7e3f1a.png)

Audacity's plain gui for the vst3 version:
![image](https://user-images.githubusercontent.com/11887021/154081893-a07f5c12-938b-4e7a-99b3-3e11c8c81e74.png)

Audacity's plain gui for the vst version:
![image](https://user-images.githubusercontent.com/11887021/154082093-f936e90a-cba6-47d4-aa2a-289308e05e85.png)


Reaper's plain gui of the vst3 version:
![image](https://user-images.githubusercontent.com/11887021/154081159-4a06a03c-69bf-4f7d-9c6b-a16d5f64b330.png)

DavidBailes:
[...]

PR #2990, opened 2022-05-26
Issue 2555: vst3 plain UI fixes

vsverchinsky:
Resolves: #2555

Also fixes issue with plain UI not being updated after user preset change

Recommended:


src/effects/VST3/VST3ParametersWindow.h
```diff
@@ -0,0 +1,58 @@
...
+class wxStaticText;
+class VST3ParameterControl;
+
+class VST3ParametersWindow
+   : public wxScrolledWindow
+{
+   std::unordered_map<Steinberg::Vst::ParamID, VST3ParameterControl*> mControls;
+   std::unordered_map<Steinberg::Vst::ParamID, VST3ParameterControl*> mLabels;
+   Steinberg::IPtr<Steinberg::Vst::IEditController> mEditController;
+   Steinberg::IPtr<Steinberg::Vst::IComponentHandler> mComponentHandler;
```
Paul-Licameli, line 37: Can these two members be const?

src/effects/VST3/VST3Utils.cpp
```diff
@@ -52,124 +52,145 @@ bool VST3Utils::ParsePluginPath(const wxString& pluginPath, wxString* modulePath
...
+                     );
+                     choice->AppendString(VST3Utils::ToWxString(displayValue));
+                  }
+                  
+                  choice->SetSelection(
+                     static_cast<int>(editController.normalizedParamToPlain(parameterInfo.id, currentNormalizedValue))
+                  );
+                  choice->Bind(wxEVT_CHOICE, [this, id = parameterInfo.id](const wxCommandEvent& event) {
+                     const auto comboBox = static_cast<wxComboBox*>(event.GetEventObject());
+                     UpdateParameter(id, comboBox->GetSelection());
```
Paul-Licameli: The first commit introduces the new window class but also changes the combo to choice, and introduces an error, because the `static_cast` above did not change the class.

Perhaps make one commit to change combo to choice first so that change isn't hidden among all the other diff for this commit.

Paul-Licameli (changes requested):

src/effects/VST3/VST3ParametersWindow.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+#include "VST3ParametersWindow.h"
+
+#include <pluginterfaces/vst/ivsteditcontroller.h>
+
+#include <wx/sizer.h>
+#include <wx/stattext.h>
+#include <wx/checkbox.h>
+#include <wx/choice.h>
+#include <wx/combobox.h>
```
Paul-Licameli, line 22: Can that #include go away?

src/effects/VST3/VST3Utils.cpp
```diff
@@ -124,64 +118,101 @@ namespace
...
+                  0,
+                  MaxValue);
+               const auto unitsStr = VST3Utils::ToWxString(parameterInfo.units);
+               const auto valueText = safenew wxStaticText(this, wxID_ANY, wxEmptyString);
+               UpdateParameterValueText(valueText, parameterInfo.id, currentNormalizedValue, unitsStr);
+
+               slider->Bind(wxEVT_SLIDER, [this, id = parameterInfo.id, valueText, unitsStr](wxCommandEvent& event) {
+                  const auto slider = static_cast<wxSlider*>(event.GetEventObject());
+                  const auto value = static_cast<Vst::ParamValue>(slider->GetValue()) / static_cast<Vst::ParamValue>(MaxValue);
+                  UpdateParameter(id, value);
```
Paul-Licameli: This is the correct normalization of `value` before `UpdateParameter` gets it?

src/effects/VST3/VST3ParametersWindow.cpp
```diff
@@ -0,0 +1,187 @@
+/**********************************************************************
```
Paul-Licameli, line 1: The commit that introduces this file makes a dependency cycle between it and VST3Utils.h

The easy solution, I think, is to move `BuildPlainUI` into the new file.  Then the new file depends one-way on VST3Utils.

src/effects/VST3/VST3ParametersWindow.cpp
```diff
@@ -25,6 +25,166 @@
 #include "MemoryX.h"
 #include "VST3Utils.h"
 
+
+//Interface between wx control and IEditController parameter.
+class VST3ParameterControl
```
Paul-Licameli, line 30: The commit that introduces this class fails to build on macOS.  It just needs in the header file:

#include <unordered_map>

src/effects/VST3/VST3ParametersWindow.cpp
```diff
@@ -101,87 +265,86 @@ void VST3ParametersWindow::Fill(Steinberg::Vst::IEditController& editController,
...
-void VST3ParametersWindow::UpdateParameterValueText(wxStaticText* text, Steinberg::Vst::ParamID id, Steinberg::Vst::ParamValue normalizedValue, const wxString& units)
+void VST3ParametersWindow::ReloadParameters()
+{
+   for(auto& p : mControls)
+      p.second->UpdateValue(*mEditController);
+   for(auto& p : mLabels)
+      p.second->UpdateValue(*mEditController);
+}
+
+void VST3ParametersWindow::UpdateParameter(Steinberg::Vst::ParamID paramId)
```
Paul-Licameli, line 316: This and the previous functions do similar things (that, to all parameters, and this, to one parameter only).  Should they have more similar names?

src/effects/VST3/VST3ParametersWindow.cpp
```diff
@@ -101,87 +265,86 @@ void VST3ParametersWindow::Fill(Steinberg::Vst::IEditController& editController,
...
-         choice->SetSelection(
-            static_cast<int>(editController.normalizedParamToPlain(parameterInfo.id, currentNormalizedValue))
-         );
-         choice->Bind(wxEVT_CHOICE, [this, id = parameterInfo.id](const wxCommandEvent& event) {
-            const auto comboBox = static_cast<wxComboBox*>(event.GetEventObject());
-            UpdateParameter(id, mEditController->plainParamToNormalized(id, comboBox->GetSelection()));
-         });
-         sizer->Add(choice, 0, wxEXPAND);
+         list->UpdateValue(editController);
+         list->Bind(wxEVT_CHOICE, &VST3ParametersWindow::OnParameterValueChanged, this);
```
Paul-Licameli, line 271: I see now that the mistake with combo boxes is fixed within the PR.  Still it would be better not to introduce it at all in the sequence of code transformations.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -403,6 +404,9 @@ bool VST3Effect::LoadSettings(
       } while(parms.GetNextEntry(key, index));
    }
 
+   if(mPlainUI != nullptr)
+      mPlainUI->ReloadParameters();
```
Paul-Licameli, line 408: I see that in the last commit of this pull request, `ReloadParameters` gets some uses but `UpdateParameter` is still not used anywhere.

Paul-Licameli: These four new calls to `ReloadParameter` will become just one call inside `TransferDataToWindow` when Pietro rebases his work for VST3 effect settings onto this branch.

I want the various settings-loading functions not to have the added responsibility to update UI but that's not a requirement to approve this pull request.

Paul-Licameli:
This completes my review.  There are no large problems.

src/effects/VST3/VST3ParametersWindow.cpp
vsverchinsky: I don't think so, `UpdateParameter` and `ReloadParmeters` have different semantics: `UpdateParameter` is to be used when individual parameters are changing during regular plugin usage, `ReloadParmeters` is to be used when plugin state is reset and we don't know which

src/effects/VST3/VST3Effect.cpp
```diff
@@ -882,7 +894,7 @@ std::unique_ptr<EffectUIValidator> VST3Effect::PopulateUI(ShuttleGui& S,
          auto vSizer = std::make_unique<wxBoxSizer>(wxVERTICAL);
          auto controlsRoot = safenew wxWindow(parent, wxID_ANY);
          if(!LoadVSTUI(controlsRoot))
-            VST3Utils::BuildPlainUI(controlsRoot, mEditController, mComponentHandler);
+            mPlainUI = VST3ParametersWindow::Setup(*controlsRoot, *mEditController, *mComponentHandler);
```
Paul-Licameli, line 897: I will approve these changes, but there is one remainning suggestion:

When you deference pointers, it's an implied assertion that the pointer is not null.  Better to prove it (with
assertions or invariants) or have null checks.

`controlsRoot` is obviously not null in this context.  Maybe the others need null checks.

Ditto, at the other use of `Setup`

Paul-Licameli (approved):
All I asked for earlier is now resolved, but maybe a few null checks should be added.

Issue #2983, opened 2022-05-26
Crash with effects stack GUI


petersampsonaudacity:
**Describe the bug**

**To Reproduce**
1. Get some audio
2. Click on the TCP "Add effects" button
3. Click on "Add new effect"
4. Choose one of the Audacity effects
5. Click on the effect button
6. In the dropdown click "No Effect"
7. Observe: Audacity crashes

I clicked on the button to send the crash report.

**Actual behavior**
Audacity crashes

**Expected behavior**
No crash

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/170450258-b4d3cdac-4951-45d4-930f-7c5c231ac671.png)

![image](https://user-images.githubusercontent.com/50205984/170449927-8a9b7080-cc93-4420-989e-02fb157a7f9a.png)

**Additional information (please complete the following information):**
 - OS: Windows 10
 - Version: audacity-win-3.2.0-alpha-20220525+8e51ac7-x64-msvc2022

**Additional context**
I am also puzzled that I get different versions of the button at Step 2, sometimes is says **Add effects** and other times it says **Effects** - see screenshots above

Also I note that if I recover the crashed project and go to the effects stack I note that the effect added at Steps 3-4 is now there twice in the stack.

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's branch build   
audacity-win-3.2.0-alpha-20220526+3404ea3-x64-msvc2022 

There is now no crash with this branch build at Step 6 - instead the effect is removed from the effect stack

vsverchinsky:
@petersampsonaudacity Thanks!

> instead the effect is removed from the effect stack
 
 As per design (if I understood everything correctly)

petersampsonaudacity:
And works fine with latest alpha master audacity-win-3.2.0-alpha-20220527+8a7fbf5-x64-msvc2022

dozzzzer:
Verified on MacOS Monterey.

PR #2991, opened 2022-05-26
Issue 2983: Fixes crash after removing single effect in the stack

vsverchinsky:
Resolves: #2983 

Recommended:


petersampsonaudacity:
Testing on W10 with @vsverchinsky 's branch build   
audacity-win-3.2.0-alpha-20220526+3404ea3-x64-msvc2022 

There is now no crash with this branch build at Step 6 - instead the effect is removed from the effect stack

Paul-Licameli (approved):

Issue #2969, opened 2022-05-23
define the settings struct, to capture VST3 parameters


Paul-Licameli:
- [x] (like 213ecd3) Extract a base class of the effect, holding the handles (effect component and edit controller) that are needed to iterate parameters and extract preset state (this new base will be re-used also with the Instance later)
- [x] (like 908c16a) Make a higher-order function for the visiting of parameter values, hiding details of how to iterate them (and use the function)
- [x] (like 3cce7fd7) import/export steps lowered into the base class
- [x] define the structure, `FetchSettings` method, `StoreSettings` method
- [x] (like part of 218c56e) Define TransferDataToWindow (to become UpdateUI of a validator later), calling StoreSettings; remove parts of LoadSettings that duplicate this
- [x] Remove `VST3Effect::LoadFactoryDefaults` (let it default, as with other effect families, to the loading of a special user preset as defined in class `Effect`)
- [x] (like rest of 218c56e) Rewrite `LoadSettings` and `SaveSettings` not using the handles, and insert needed calls to `FetchSettings`
- [x] define MakeSettings (like in commit e0b969fe39f4fb1b242f2b584fcd52fc210b65f3)
- [x] (like 5199894) Rewrite export and saving of presets, using `StoreSettings` on the "scratch" handles in the Effect object
- [ ] all of the above steps approved in PR and merged

Issue #2606, opened 2022-03-01
Convert VST3 effects to new stateless system


AnitaBats:
- [x] #2969
- [ ] #3313
- [ ] #3306
- [ ] #3307
- [ ] #3308

pietro68:
- [ ] define the settings struct, to capture VST3 parameters
- [ ] write FetchSettings method, to get settings from the plugin instance
- [ ] write StoreSettings method, to set settings to the plugin instance

Paul-Licameli:
I recall these non-obvious details were discovered in conversation with @vsverchinsky:

- `ComponentHandler` objects may be created and destroyed with opening and closing of dialogs (don't need long life), so are part of Validator
- The `performEdit` override happens during GUI event callbacks, and is where EffectSettingsAccess should be used
- The duplicate inter-thread communication mechanism (uses of `.exchange`) will be removed, and the `Access` class in `RealtimeEffectState.cpp` will handle that instead
- `VST3Effect::LoadSettings` should not need a `ComponentHandler` just to retrieve values from the config and put them into `EffectSettings`
- Our subclass of `IParameterChanges` needs to be reimplemented "somehow" (details not yet clear) to detect only the difference between before-and-after versions of EffectSettings, and send that to the VST libraries
- `VST3Effect::ProcessBlock`, also now using `getPendingChanges`, is only on the path of single-threaded, destructive effect processing, without concurrent changes of settings; therefore, it might be simplified (use a different `IParameterChanges` subclass?)

That's my brain dump of details.  These might become sub-sub-tasks in broken-out issues later.  Some related to the Validator; others, to "Elimination of duplicate inter-thread-communication" which I might make a main item in the checkbox list.

vsverchinsky:
> ComponentHandler objects may be created and destroyed with opening and closing of dialogs (don't need long life), so are part of Validator

Theoretically yes, it is said that IComponentHandle is resposible `transfer of parameter editing to component (processor) via host` and , but at the same time it is marked as `mandatory` [https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IComponentHandler.html#details](url)
Maybe UI is not always the reason for parameters to change?

>Our subclass of IParameterChanges needs to be reimplemented "somehow" (details not yet clear) to detect only the difference between before-and-after versions of EffectSettings, and send that to the VST libraries

[...]

PR #2992, opened 2022-05-26
Define settings vst3 i2969

pietro68:

Implemented bullet points of the above mentioned issue; one of the bullet points was expanded to 3 commits, otherwise there is a 1:1 bullet point/commit ratio, and the commit message is the same as the bullet point.

(please note: the last bullet point is still to be committed)

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -1217,3 +1217,26 @@ void VST3Effect::ReloadUserOptions()
 
    SetBlockSize(mUserBlockSize);
 }
+
+bool VST3Wrapper::ForEachParameter(ParameterVisitor visitor) const
+{
+   if (mEditController == nullptr)
+      return false;
+
+   for (int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
+   {
+      ParameterInfo pi;
+
+      Steinberg::Vst::ParameterInfo vstParameterInfo{ };
+      if (mEditController->getParameterInfo(i, vstParameterInfo) == Steinberg::kResultOk)
+      {
+         if (vstParameterInfo.flags & Steinberg::Vst::ParameterInfo::kCanAutomate)
+         {
+            if (!visitor(ParameterInfo{ vstParameterInfo }))
+               break;
+         }
+      }
+   }
+
+   return true;
+}
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -350,27 +350,18 @@ bool VST3Effect::SupportsAutomation() const
 bool VST3Effect::SaveSettings(
    const EffectSettings &, CommandParameters & parms) const
 {
-   if(mEditController == nullptr)
-      return false;
-
-   using namespace Steinberg;
-   
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+   return ForEachParameter
+   (
+      [&](const ParameterInfo& pi)
       {
-         if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
-         {
-            parms.Write(
-               VST3Utils::MakeAutomationParameterKey(parameterInfo),
-               mEditController->getParamNormalized(parameterInfo.id)
-            );
-         }
-      }
-   }
+         parms.Write(
+            VST3Utils::MakeAutomationParameterKey(pi),
+            mEditController->getParamNormalized(pi.id)
+         );
 
-   return true;
+         return true;
+      }
+   );
 }
 
 bool VST3Effect::LoadSettings(
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -1262,3 +1262,63 @@ bool VST3Wrapper::SavePreset(Steinberg::IBStream* fileStream, const Steinberg::F
       mEditController.get()
    );
 }
+
+
+bool VST3Wrapper::FetchSettings(VST3EffectSettings& settings) const
+{
+   settings.mValues.clear();
+
+   return ForEachParameter(
+
+      [&](const ParameterInfo& pi)
+   {
+      const auto id = pi.id;
+      settings.mValues[id] = mEditController->getParamNormalized(id);
+      return true;
+   }
+   );
+}
+
+
+bool VST3Wrapper::StoreSettings(const VST3EffectSettings& settings) const
+{
+   return ForEachParameter(
+
+      [&](const ParameterInfo& pi)
+      {
+         if (mComponentHandler == nullptr)
+            return false;
+
+         const auto id = pi.id;
+
+         auto itr = settings.mValues.find(id);
+         if (itr != settings.mValues.end())
+         {
+            const auto& value = itr->second;
+
+            // set the value to the handler at the id
+            if (mComponentHandler->beginEdit(id) == Steinberg::kResultOk)
+            {
+               auto cleanup = finally([&] {
+                  mComponentHandler->endEdit(id);
+               });
+               mComponentHandler->performEdit(id, value);
+            }
+            else
+            {
+               return false;
+            }
+            mEditController->setParamNormalized(id, value);
+
+         }
+         else
+         {
+            return false;
+         }
+
+         return true;
+      }
+   );
+
+}
+
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -492,33 +492,6 @@ bool VST3Effect::LoadFactoryPreset(int id, EffectSettings &) const
    return true;
 }
 
-bool VST3Effect::LoadFactoryDefaults(EffectSettings &) const
-{
-   using namespace Steinberg;
-   if(mComponentHandler == nullptr)
-      return false;
-
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
-      {
-         if(parameterInfo.flags & Vst::ParameterInfo::kIsReadOnly)
-            continue;
-
-         if(mComponentHandler->beginEdit(parameterInfo.id) == kResultOk)
-         {
-            auto cleanup = finally([&]{
-               mComponentHandler->endEdit(parameterInfo.id);
-            });
-            mComponentHandler->performEdit(parameterInfo.id, parameterInfo.defaultNormalizedValue);
-         }
-         mEditController->setParamNormalized(parameterInfo.id, parameterInfo.defaultNormalizedValue);
-      }
-   }
-
-   return true;
-}
 
 namespace
 {
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -427,6 +427,8 @@ bool VST3Effect::LoadUserPreset(
    }
    SyncParameters(settings);
 
+   FetchSettings(GetSettings(settings));
+
    return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/effects/VST3/VST3Effect.cpp
```diff
@@ -1217,3 +1183,128 @@ void VST3Effect::ReloadUserOptions()
...
+
+
+bool VST3Wrapper::ForEachParameter(ParameterVisitor visitor) const
+{
+   if (mEditController == nullptr)
+      return false;
+
+   for (int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
+   {
+      ParameterInfo pi;
```
Paul-Licameli, line 1208: unused variable

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1217,3 +1183,128 @@ void VST3Effect::ReloadUserOptions()
...
+   {
+      return true;
+   }
+   
+   return false;
+}
+
+
+
+bool VST3Wrapper::ForEachParameter(ParameterVisitor visitor) const
```
Paul-Licameli, line 1201: In the commit that introduces this function, don't just define it, but also insert uses of it in all possible places.

`getParameterCount` should be called directly in this function only.

I count four functions to rewrite in VST3Effect.cpp and one more in VST3Utils.cpp.  There is a function `FillParameters` which could be changed to take a reference to a wrapper instead of EditController.

This also implies the VST3Wrapper struct should be declared and defined in VST3Utils.h/.cpp, so that there is not a dependency cycle between those and VST3Effect.

src/effects/VST3/VST3Effect.h
```diff
@@ -38,17 +39,72 @@ namespace Steinberg
...
+};
+
+struct VST3Wrapper
+{
+   // For the time being, here we have members that are needed
+   // to iterate parameters and extract preset state
+   //
+   Steinberg::IPtr<Steinberg::Vst::IEditController> mEditController;
+   Steinberg::IPtr<Steinberg::Vst::IComponent>      mEffectComponent;
+   Steinberg::IPtr<internal::ComponentHandler>      mComponentHandler;
```
Paul-Licameli, line 55: See my comments here: https://github.com/audacity/audacity/issues/2606

I determined that the uses of component handler do not belong in the fetching and storing of settings.  Instead, these are (some of?) the steps that need to move into `TransferDataToWindow` and ultimately to the Validator.

So I think the wrapper should only have the first two of these handles.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1217,3 +1183,128 @@ void VST3Effect::ReloadUserOptions()
...
+}
+
+
+bool VST3Wrapper::StoreSettings(const VST3EffectSettings& settings) const
+{
+   return ForEachParameter(
+
+      [&](const ParameterInfo& pi)
+      {
+         if (mComponentHandler == nullptr)
```
Paul-Licameli, line 1275: The component handler operations should not be in this function; but `setParamNormalized` needs to be called, so there should be a null check on the edit controller.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1217,3 +1183,128 @@ void VST3Effect::ReloadUserOptions()
 
    SetBlockSize(mUserBlockSize);
 }
+
+
+bool VST3Effect::TransferDataToWindow(const EffectSettings& settings)
+{
+   // Update parameter values in the plugin
+   if (StoreSettings(GetSettings(settings)))
```
Paul-Licameli, line 1191: The component handler operations need to be lifted out of `StoreSettings` and called directly from this function.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -481,45 +473,23 @@ RegistryPaths VST3Effect::GetFactoryPresets() const
...
       // To do: externalize state so const_cast isn't needed
-      return const_cast<VST3Effect*>(this)->LoadPreset(filename.GetFullPath());
+      const bool res = const_cast<VST3Effect*>(this)->LoadPreset(filename.GetFullPath());
+      FetchSettings(GetSettings(settings));
+      return res;
    }
-   return true;
-}
-
-bool VST3Effect::LoadFactoryDefaults(EffectSettings &) const
```
Paul-Licameli, line 495: In the comment of the commit that removes this function, add more explanation.  Quote the rest of my bullet point.

src/effects/VST3/VST3Effect.cpp
Paul-Licameli: Recall also: it may be better to fill a vector of ParameterInfo to iterate, only once, say in the contructor of `VST3Wrapper`.  Then the for-each function just iterates the vector.

When the Instance class is introduced, construct its wrapper to share -- not reinitialize -- the vector of ParameterInfo that is in the EffectPlugin.  That would imply other changes in the constructor.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -350,58 +350,48 @@ bool VST3Effect::SupportsAutomation() const
...
-   if(mEditController == nullptr)
-      return false;
-
-   using namespace Steinberg;
-   
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+   return ForEachParameter
```
Paul-Licameli, line 353: Rewriting `SaveSettings` to use `ForEachParameter` should happen, as I said, in step 2.  Not this later step.

What is meant to change in this later step is the elimination of the use of the handles (here the editController).  Instead, EffectSettings is used, to find the already fetched values.

And because there is an assumption that values are fetched, this step shouldn't really be separated into three.  The insertion of all the necessary calls to `FetchSettings` should happen in the same commit -- otherwise this step in the transformation introduces an incorrectness, though fixed in a later step.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -350,58 +350,48 @@ bool VST3Effect::SupportsAutomation() const
...
-      do
+
+   ForEachParameter
+   (
+      [&](const ParameterInfo& pi)
       {
-         Steinberg::Vst::ParamID id;
-         Vst::ParamValue value;
-         if(VST3Utils::ParseAutomationParameterKey(key, id) && parms.Read(key, &value))
+         const auto id = pi.id;
```
Paul-Licameli, line 379: Again:  there should be a first transformation of this function to use `ForEachParameter`.  That means the iteration over `parms` instead becomes a lookup.

`MakeAutomationParameterKey` is used in `SaveSettings` to map `ParameterInfo` to the string key value used in the config file.  This function should call the same, but then to read, not write, that key in the config.  The function `ParseAutomationParameterKey` becomes unused and can be removed.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -350,58 +350,48 @@ bool VST3Effect::SupportsAutomation() const
...
-            }
-            mEditController->setParamNormalized(id, value);
+            auto cleanup = finally([&] {
+               mComponentHandler->endEdit(id);
+            });
+            mComponentHandler->performEdit(id, value);
          }
-      } while(parms.GetNextEntry(key, index));
-   }
+         mEditController->setParamNormalized(id, value);
```
Paul-Licameli, line 389: The second step of the transformation of this function is meant to eliminate all the uses of the handles, but you have not done that yet.

When you define `StoreSettings`, it introduces a duplication of the use of the edit controller -- but don't eliminate this call yet.

When you define `TransferDataToWindow`, you eliminate the duplicate lines that use the component handler, which move to that function.  And because `TransferDataToWindow` inserts a call to `StoreSettings`, then at this step you can eliminate the duplicate use of the edit controller here.  You should be able to verify with testing of an effect, that there is no loss of response of the effect to changes of the dials.

So then perhaps all those prior steps will leave nothing to do in `LoadSettings` at this late step.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -437,6 +427,8 @@ bool VST3Effect::LoadUserPreset(
    }
    SyncParameters(settings);
 
+   FetchSettings(GetSettings(settings));
```
Paul-Licameli, line 430: Reviewing the last commit reminds me of `SyncParameters` which happens before this `FetchSettings`.  I think the `SyncParameters` call is another thing that needs to move into `TransferDataToWindow`.

Paul-Licameli: Indeed, `SyncParameters` uses the component handler (the handle that will become the Instance object, I think).  So that call doesn't belong in these functions that are supposed to manipulate settings.  It is not one of the "scratch" handles needed to import and export.

So `SyncParameters` does need to move somewhere else.  I'm not sure `TransferDataToWindow` is the right place, but let's begin by making that transformation in this PR at the earlier step.  Maybe we reconsider that later.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -481,45 +473,23 @@ RegistryPaths VST3Effect::GetFactoryPresets() const
...
-         {
-            auto cleanup = finally([&]{
-               mComponentHandler->endEdit(parameterInfo.id);
-            });
-            mComponentHandler->performEdit(parameterInfo.id, parameterInfo.defaultNormalizedValue);
-         }
-         mEditController->setParamNormalized(parameterInfo.id, parameterInfo.defaultNormalizedValue);
-      }
-   }
+   FetchSettings(GetSettings(settings));
```
Paul-Licameli, line 487: The two calls to `FetchSettings` in `LoadFactoryPreset` might be simplified to one.  Just return `res` when done.

Paul-Licameli, line 487: The second call to `FetchSettings` isn't needed -- it's on an error path in which no side-effect happened on the handles, so there is no change to fetch.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -481,45 +473,23 @@ RegistryPaths VST3Effect::GetFactoryPresets() const
...
-bool VST3Effect::LoadFactoryPreset(int id, EffectSettings &) const
+bool VST3Effect::LoadFactoryPreset(int id, EffectSettings& settings) const
 {
    if(id >= 0 && id < mFactoryPresets.size())
    {
       auto filename = wxFileName(GetFactoryPresetsPath(mEffectClassInfo), mFactoryPresets[id] + ".vstpreset");
       // To do: externalize state so const_cast isn't needed
-      return const_cast<VST3Effect*>(this)->LoadPreset(filename.GetFullPath());
+      const bool res = const_cast<VST3Effect*>(this)->LoadPreset(filename.GetFullPath());
+      FetchSettings(GetSettings(settings));
```
Paul-Licameli, line 483: I think two new calls to `FetchSettings` can simplify to one, if you lower it into `LoadPreset`, which will cover the import path too.

Paul-Licameli:
This completes my first round of review.

Paul-Licameli:
See also this comment https://github.com/audacity/audacity/pull/2990/commits/38b3c3835866d7aed6810f7bc9e61fcc469f0694#r883707852

src/effects/VST3/VST3Effect.cpp
pietro68: Unfortunately I can not fill the vector of ParameterInfo in the VST3Wrapper constructor, because I need to reference `mEditController `and that gets ready only at the end of `VST3Effect `constructor. So I had to opt for an ::`Init `method instead. Later I could try to rework the `VST3Effect `contsructor, to move the stuff which initializes `mEditController `to the Wrapper.

src/effects/VST3/VST3Effect.cpp
pietro68: ok

src/effects/VST3/VST3Effect.cpp
Paul-Licameli: Then, some kind of create-once-on-demand function in the new base class can return the vector of ParameterInfo, and that function could fail if called to early, and the for-each function should check for that error and pass it along.

src/effects/VST3/VST3Effect.cpp
pietro68: I see, however I find it strange that on the error path the function returns true

pietro68:
The requested changes are such that I thought it would be easier to start a new PR at https://github.com/audacity/audacity/pull/3009 - so I am closing this one.

Issue #2997, opened 2022-05-27
AU PeakLimiter effect is not displaying interface


Ryan-Miller-Audacity:
**Describe the bug**
In testing of #2934, it was noted that the AU PeakLimiter effect is no longer displaying correctly. When opened, the popup for the effect is displayed, but the actual interface (to make changes) is not; it is just a blank popup. 

**To Reproduce**
1. Go to https://github.com/audacity/audacity/actions/runs/2363414860 and retrieve the relevant macOS build from the artifacts
2. Open audacity and add some audio (import, generate chirp etc.)
3. From the effect dropdown at the top, select AUPeakLimiter
4. Note that the popup displays but without the effect's interface

**Actual behavior**
<img width="1440" alt="Screen Shot 2022-05-27 at 7 44 37 AM (2)" src="https://user-images.githubusercontent.com/88568755/170693107-19f38653-c3e3-4180-afbe-276848da5821.png">

**Expected behavior**
<img width="1440" alt="Screen Shot 2022-05-27 at 7 43 52 AM (2)" src="https://user-images.githubusercontent.com/88568755/170692970-d8eccaf3-41bc-4984-a103-73bda15e3e89.png">

**Screenshots**
See above.

**Additional information (please complete the following information):**
 - OS: macOS Monterey 12.3.1
 - Version: Audacity 3.2.0-alpha-20220521 commit ID: da7cfe [Related Merge](https://github.com/audacity/audacity/actions/runs/2363414860)
 - (if relevant) Sound Device: N/A

**Additional context**
N/A

Update:  verify this test case too:

```
   //       To see the problem, comment out the setFrameSize calls below, create/save
   //       a preset for the AUDelay effect with "Invert Feedback" checked, restore
   //       factory defaults and then reselect the save preset.  The display will not
   //       update properly.  But, resize the window and it does.
```
Note that:  proper updating means not only checking the box, but also changing the vertical scale of the graph so it goes from -100% to +100%.

Paul-Licameli:
The problem is only that the initial size of the dialog is wrong.  The dials appear if you drag the lower edge down.

Paul-Licameli:
There is another relevant old piece of code.  @Ryan-Miller-Audacity note this test case:

```
   //       To see the problem, comment out the setFrameSize calls below, create/save
   //       a preset for the AUDelay effect with "Invert Feedback" checked, restore
   //       factory defaults and then reselect the save preset.  The display will not
[...]

PR #2999, opened 2022-05-28
Fix intial size of AU PeakLimiter while also not breaking the fix...

Paul-Licameli:
... described in the old comments from LLL

Resolves: #2997

Recommended:


pietro68:
I verified this fix works, and the code change is fine for me.

pietro68 (approved):
I verified this fix works, and the code change is fine for me.

Issue #2993, opened 2022-05-26
Importing AU effect presets results in error


Ryan-Miller-Audacity:
**Describe the bug**
When testing changes from #2934, it was noted that attempting to import an exported AU plugin preset does not properly load the preset and gives the user the following error message: 
<img width="1440" alt="Screen Shot 2022-05-25 at 11 03 05 AM (2)" src="https://user-images.githubusercontent.com/88568755/170530726-b31206ab-ba2a-48f6-9e2b-d8e2ac67fc36.png">

**To Reproduce**
1. Go to https://github.com/audacity/audacity/actions/runs/2363414860 and retrieve the relevant macOS build from the artifacts
2. Open Audacity and generate a sound clip (import some audio, generate a chirp etc) 
3. From the effects menu select an AudioUnit effect 
4. Edit the effect so that it is different than the default of the plugin
5. Click the hamburger menu
6. Select "Export"
7. Give the preset a unique name and save
8. Reset the plugin back to the factory preset - default (this just allows you to see a difference _if_ the preset did import correctly
9. Click the hamburger menu, select "import", and select the file you created previously
10. Note the error prompt displayed.

**Actual behavior**
Attempting to import a preset for an AudioUnit Effect plugin results in an error prompt.

**Expected behavior**
The preset should be correctly imported and the changes made to the AU's interface based on the preset. 

**Screenshots**
See above.

**Additional information (please complete the following information):**
 - OS: MacOS Monterey 12.3.1
 - Version: Audacity 3.2.0-alpha-20220521 commit ID: da7cfe [Related Merge](https://github.com/audacity/audacity/actions/runs/2363414860)
 - (if relevant) Sound Device: N/A

**Additional context**
N/A

dozzzzer:
Tested on MacOS 12.3.1

PR #3000, opened 2022-05-29
Fix importing of AU presets, a one-liner; c5b7594 was to blame

Paul-Licameli:
Resolves: #2993

Recommended:


pietro68 (approved):
One-liner fix looks fine, tried the branch and the steps and I could see the crash happens no more.

Paul-Licameli:
> One-liner fix looks fine, tried the branch and the steps and I could see the crash happens no more.

Crash?  It wasn’t a crash, I thought it was just a failure with an error message.

Issue #2898, opened 2022-05-06
Mix and Render 


SteveDaulton:
**Describe the bug**
When applied twice to a stereo track, the right channel is mixed into the left channel.

**To Reproduce**
1. Generate a (mono) 30 second Chirp
2. Generate a (mono) 30 second DTMF tone
3. Join the two tracks into a stereo track
4. Select the track
5. Mix & Render (workaround for bug #2137)
6. Observe that nothing appears to happen
7. Add some envelope points
8. Mix and Render (workaround for bug #1500)

**Actual behavior**
At step 6, nothing appears to happen.
On step 8, the right channel is mixed into the left channel.

**Expected behavior**
At step 6, the "stereo-like" track with drag bars in both channels is expected to become a proper stereo track with just one drag bar at the top.
On step 8, the envelope points should be rendered without cross-channel mixing.

**Screenshots**
If applicable, add screenshots/video to help explain your problem.

![First Track000](https://user-images.githubusercontent.com/11681097/167127664-97708292-e6e4-4f2d-aa0b-0af103ae4b9d.png)

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version Audacity 3.2.0 f1b586
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

**Additional context**

petersampsonaudacity:
Conitmed on W10 with 3,2,0 alpha master audacity-win-3.2.0-alpha-20220505+f1b5865-x64-msvc2022

Testing shows this is s regression on 3.1.3

petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch build audacity-win-3.2.0-alpha-20220530+777bb3e-x64

This branch build appears to properly fix this issue on Windows

petersampsonaudacity:
Testing on W10 with latest alpha master:  audacity-win-3.2.0-alpha-20220530+41d501e-x64-msvc2022

I confirm that his now works as expected on Windows

PR #3006, opened 2022-05-30
Fix mix-and-render of unaligned stereo channels...

Paul-Licameli:
... so the result has aligned channels, as in version 3.1

Resolves: #2898

Recommended:

### DIFF
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -219,13 +219,14 @@ void Track::DoSetLinkType(LinkType linkType, bool completeList)
       // Becoming unlinked
       assert(mpGroupData);
       if (HasLinkedTrack()) {
-         // Make independent copy of group data in the partner, which should
-         // have had none
-         auto partner = GetLinkedTrack();
-         assert(!partner->mpGroupData);
-         partner->mpGroupData =
-            std::make_unique<ChannelGroupData>(*mpGroupData);
-         partner->mpGroupData->mLinkType = LinkType::None;
+         if (auto partner = GetLinkedTrack()) {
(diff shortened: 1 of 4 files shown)


petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch build audacity-win-3.2.0-alpha-20220530+777bb3e-x64

This branch build appears to properly fix this issue on Windows

libraries/lib-track/Track.cpp
```diff
@@ -219,13 +219,14 @@ void Track::DoSetLinkType(LinkType linkType, bool completeList)
...
       assert(mpGroupData);
       if (HasLinkedTrack()) {
-         // Make independent copy of group data in the partner, which should
-         // have had none
-         auto partner = GetLinkedTrack();
-         assert(!partner->mpGroupData);
-         partner->mpGroupData =
-            std::make_unique<ChannelGroupData>(*mpGroupData);
-         partner->mpGroupData->mLinkType = LinkType::None;
+         if (auto partner = GetLinkedTrack()) {
```
vsverchinsky, line 222: It is possible that `HasLinkedTrack` will succeed, but `GetLinkedTrack` still may return null pointer?

vsverchinsky (approved):

libraries/lib-track/Track.cpp
Paul-Licameli: Yes, that happens in this case.

A track was made as EmptyCopy of another, including the group data, and added to the end of the TrackList.  It was stereo group data but the right channel was not yet also added.  So in fact link consistency check criteria are temporarily violated.  But then the inconsistency is repaired by resetting the link type, but keeping any other group data.

PR #3009, opened 2022-05-31
Define settings vst3 i2969 2nd

pietro68:

Second iteration, addressing requests made in PR https://github.com/audacity/audacity/pull/2992

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -1199,3 +1199,47 @@ void VST3Effect::ReloadUserOptions()
 
    SetBlockSize(mUserBlockSize);
 }
+
+
+bool VST3Effect::TransferDataToWindow(const EffectSettings& settings)
+{
+   const auto vst3Settings = GetSettings(settings);
+
+   return ForEachParameter(
+
+      [&](const ParameterInfo& pi)
+      {
+         if (mComponentHandler == nullptr)
+            return false;
+
+         const auto id = pi.id;
+
+         auto itr = vst3Settings.mValues.find(id);
+         if (itr != vst3Settings.mValues.end())
+         {
+            const auto& value = itr->second;
+
+            // set the value to the handler at the id
+            if (mComponentHandler->beginEdit(id) == Steinberg::kResultOk)
+            {
+               auto cleanup = finally([&] {
+                  mComponentHandler->endEdit(id);
+               });
+               mComponentHandler->performEdit(id, value);
+            }
+            else
+            {
+               return false;
+            }
+         }
+         else
+         {
+            return false;
+         }
+
+         return true;
+      }
+
+   );
+
+}
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -329,22 +329,15 @@ bool VST3Effect::SupportsRealtime() const
 
 bool VST3Effect::SupportsAutomation() const
 {
-   if(mEditController == nullptr)
-      return false;
-
    using namespace Steinberg;
-   
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
-      {
-         if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
-            return true;
-      }
-   }
 
-   return false;
+   return AtLeastOne
+   (
+      [&](const ParameterInfo& parameterInfo)
+      {
+         return (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate);            
+      }
+   );
 }
 
 bool VST3Effect::SaveSettings(
@@ -354,23 +347,21 @@ bool VST3Effect::SaveSettings(
       return false;
 
    using namespace Steinberg;
-   
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+
+   return ForEachParameter
+   (
+      [&](const ParameterInfo& parameterInfo)
       {
-         if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
+	      if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
          {
             parms.Write(
                VST3Utils::MakeAutomationParameterKey(parameterInfo),
                mEditController->getParamNormalized(parameterInfo.id)
             );
-         }
+	      }
+         return true;
       }
-   }
-
-   return true;
+   );   
 }
 
 bool VST3Effect::LoadSettings(
@@ -378,30 +369,27 @@ bool VST3Effect::LoadSettings(
 {
    using namespace Steinberg;
 
-   if(mComponentHandler == nullptr)
+   if (mComponentHandler == nullptr)
       return false;
-   
-   long index { };
-   wxString key;
-   if(parms.GetFirstEntry(key, index))
-   {
-      do
+
+   ForEachParameter
+   (
+      [&](const ParameterInfo& pi)
       {
-         Steinberg::Vst::ParamID id;
-         Vst::ParamValue value;
-         if(VST3Utils::ParseAutomationParameterKey(key, id) && parms.Read(key, &value))
+         const auto id = pi.id;
+         const auto value = pi.defaultNormalizedValue;
+
+         if (mComponentHandler->beginEdit(id) == kResultOk)
          {
-            if(mComponentHandler->beginEdit(id) == kResultOk)
-            {
-               auto cleanup = finally([&]{
-                  mComponentHandler->endEdit(id);
-               });
-               mComponentHandler->performEdit(id, value);
-            }
-            mEditController->setParamNormalized(id, value);
+            auto cleanup = finally([&] {
+               mComponentHandler->endEdit(id);
+            });
+            mComponentHandler->performEdit(id, value);
          }
-      } while(parms.GetNextEntry(key, index));
-   }
+         mEditController->setParamNormalized(id, value);
+      return true;
+      }
+   );
 
    return true;
 }
@@ -498,13 +486,12 @@ bool VST3Effect::LoadFactoryDefaults(EffectSettings &) const
    if(mComponentHandler == nullptr)
       return false;
 
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
+   return ForEachParameter
+   (
+      [&](const ParameterInfo& parameterInfo)
       {
          if(parameterInfo.flags & Vst::ParameterInfo::kIsReadOnly)
-            continue;
+            return true;
 
          if(mComponentHandler->beginEdit(parameterInfo.id) == kResultOk)
          {
@@ -514,10 +501,11 @@ bool VST3Effect::LoadFactoryDefaults(EffectSettings &) const
             mComponentHandler->performEdit(parameterInfo.id, parameterInfo.defaultNormalizedValue);
          }
          mEditController->setParamNormalized(parameterInfo.id, parameterInfo.defaultNormalizedValue);
-      }
-   }
 
-   return true;
+         return true;
+      }
+   );
+
 }
 
 namespace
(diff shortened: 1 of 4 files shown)


Paul-Licameli:
There is one more checkbox in the list, not corresponding to a commit.

Paul-Licameli:
The PR is failing to build on all operating systems.  Are you sure you disabled precompiled headers?

src/effects/VST3/VST3Utils.cpp
```diff
@@ -238,3 +245,132 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
...
+      if (mEditController == nullptr)
+         return nullptr;
+
+      for (int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
+      {
+         Steinberg::Vst::ParameterInfo vstParameterInfo{ };
+         if (mEditController->getParameterInfo(i, vstParameterInfo) == Steinberg::kResultOk)
+         {
+            // Add all parameterinfos, regardless of whether they can be automated or not,
+            // because different flags of them will be checked by different methods.
```
Paul-Licameli, line 263: I think Vitaly mentioned that the set of parameters might really be very large?

If the filtering by parameter types is moved into `ForEachParameter`, then more repetition of code can be eliminated, and fewer parameters would need to be remembered here.

If you need to filter differently in one place, then VisitParameters might take another argument indicating how to filter the parameters, and its default value could select the usual case.

src/effects/VST3/VST3Utils.h
```diff
@@ -68,3 +77,70 @@ class VST3Utils final
...
+   // to iterate parameters and extract preset state
+   //
+   Steinberg::IPtr<Steinberg::Vst::IEditController> mEditController;
+   Steinberg::IPtr<Steinberg::Vst::IComponent>      mEffectComponent;
+
+   using ParameterInfo = Steinberg::Vst::ParameterInfo;
+
+   using ParameterVisitor = std::function< bool(const ParameterInfo& pi) >;
+
+   mutable std::vector<ParameterInfo> mParameterInfos;
```
Paul-Licameli, line 99: Explain in a comment that this is mutable because it is cached once on demand.

Or even better:  define a class ParameterCache.  Give the wrapper a `const std::shared_ptr` to a cache.  Put the mutable member in the cache class, and give it a `const` member function that gives a reference to this vector.

Give `VSTWrapper` a constructor taking `std::shared_ptr` to this cache.  The constructor calls `make_shared` if the argument is null, or else uses the given argument.

These suggestions anticipate the definition of an Instance, which can share the cache of parameters with its plug-in.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -238,3 +245,132 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
...
+   {
+      if ( ! visitor(parameterInfo) )
+         break;
+   }
+
+   return true;
+}
+
+
+bool VST3Wrapper::AtLeastOne(ParameterVisitor visitor) const
```
Paul-Licameli, line 289: Maybe you can call this `AnyOf` by analogy with `std::any_of`

Paul-Licameli (changes requested):

src/effects/VST3/VST3Utils.cpp
```diff
@@ -238,3 +245,132 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
...
+         auto itr = settings.mValues.find(parameterinfo.id);
+         if (itr != settings.mValues.end())
+         {
+            const auto& value = itr->second;
+
+            mEditController->setParamNormalized(parameterinfo.id, value);
+         }
+         else
+         {
+            return false;
```
Paul-Licameli, line 369: You cause failure if a parameter is known in the plug-in but not known in the mapping.  This is not like what I did for AudioUnits.  I think it's better to ignore this case and continue storing other parameters.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1217,3 +1175,47 @@ void VST3Effect::ReloadUserOptions()
 
    SetBlockSize(mUserBlockSize);
 }
+
+
+bool VST3Effect::TransferDataToWindow(const EffectSettings& settings)
```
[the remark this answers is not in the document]
Paul-Licameli: Remember also the calls to `mPlainUI->ReloadParameters();` that move into this function, when this PR is rebased onto @vsverchinsky 's #2990.  (But wait for that PR to be redrafted in response to the first round of review.)

src/effects/VST3/VST3Utils.cpp
```diff
@@ -55,7 +57,8 @@ namespace
 void FillParameters(
    wxWindow* parent,
    Steinberg::Vst::IEditController* editController,
-   Steinberg::Vst::IComponentHandler* handler)
+   Steinberg::Vst::IComponentHandler* handler,
+   const VST3Wrapper& wrapper)
```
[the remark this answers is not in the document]
pietro68: > The second commit fails to compile for me, unless I add to VST3Utils.h:
> 
> ```
> #include <vector>
> ```

Apologies, it was not yet clear to me that I should cmake-generate with pch off, when doing development.  When turning it off, I actually found that I'd also need `#include <functional>`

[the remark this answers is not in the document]
pietro68: > This commit adds an argument to FillParameters and BuildPlainUI. Instead, this could replace the editController argument, and that can be fetched from the wrapper.

good point, will do that

src/effects/VST3/VST3Utils.cpp
pietro68: I have seen that there are about 2000 parameters for the gain plugin which I am using in my tests - so yes, very large (and I can not fathom why a gain plugin needs to have 2k parameters... but that's another issue).

Scanning 2k of them each time is definitively overkill, so I'd implement what you suggest. However, I think that to make it worth, I'd need to have not just one vector<parameterinfo> as we have now, but a set of vector<parameterinfo> e.g. one for the params which are automatable, one for the readonly ones - otherwise if I had only one vector, we'd be back to square one i.e. I'd have to scan all 2000 items of it anyway.

src/effects/VST3/VST3Utils.cpp
pietro68: ok, done

src/effects/VST3/VST3Utils.cpp
Paul-Licameli: Note this from Steinberg's header file, ivsteditcontroller.h:

```
		kIsReadOnly		 = 1 << 1,	///< parameter cannot be changed from outside the plug-in (implies that kCanAutomate is NOT set)
...
		kIsHidden		 = 1 << 4,	///< parameter should be NOT displayed and cannot be changed from outside the plug-in 
									///< (implies that kCanAutomate is NOT set and kIsReadOnly is set) [SDK 3.7.0]

		kIsProgramChange = 1 << 15,	///< parameter is a program change (unitId gives info about associated unit 
									///< - see \ref vst3ProgramLists)
		kIsBypass		 = 1 << 16	///< special bypass parameter (only one allowed): plug-in can handle bypass
									///< (highly recommended to export a bypass parameter for effect plug-in)
```

So, check my logic please: I think we can always ignore the hidden parameters.  Because, see how `parameterInfo.flags` is used.

In the places where only the automatable parameters are examined -- that implies that the read-only are excluded.  In `SyncParameters`, all (and only) the read-onlies are excluded (explicitly).  Therefore, these visits exclude all the hidden.

In `FillParameters`, when building a plain UI -- we may examine some read-only parameters.  But, we exclude the hidden ones.

Therefore:  we always ignore the hidden.

Maybe you can easily verify with some temporary code, that most of these Amplify parameters are hidden?

src/effects/VST3/VST3Effect.cpp
```diff
@@ -341,10 +341,9 @@ bool VST3Effect::SupportsAutomation() const
 }
 
 bool VST3Effect::SaveSettings(
-   const EffectSettings &, CommandParameters & parms) const
+   const EffectSettings& settings, CommandParameters & parms) const
 {
-   if(mEditController == nullptr)
-      return false;
+   const VST3EffectSettings ownSettings = GetSettings(settings);
```
Paul-Licameli, line 346: Make `ownSettings` a reference and avoid a map copy

src/effects/VST3/VST3Effect.cpp
```diff
@@ -365,32 +364,32 @@ bool VST3Effect::SaveSettings(
...
 
 bool VST3Effect::LoadSettings(
-   const CommandParameters & parms, EffectSettings &settings) const
+   const CommandParameters& parms, EffectSettings& settings) const
 {
    using namespace Steinberg;
 
-   if (mComponentHandler == nullptr)
-      return false;
+   VST3EffectSettings ownSettings;
```
Paul-Licameli, line 371: Could write: `auto &ownSettings = GetSettings(settings); ownSettings.clear();` and eliminate the assignment a the end

src/effects/VST3/VST3Effect.cpp
```diff
@@ -365,32 +364,32 @@ bool VST3Effect::SaveSettings(
...
-         const auto value = pi.defaultNormalizedValue;
-
-         if (mComponentHandler->beginEdit(id) == kResultOk)
+         if (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
          {
-            auto cleanup = finally([&] {
-               mComponentHandler->endEdit(id);
-            });
-            mComponentHandler->performEdit(id, value);
+            const auto key = VST3Utils::MakeAutomationParameterKey(parameterInfo);
```
Paul-Licameli, line 379: When you correct the lookup into `parms`, also remove `ParseAutomationParameterKey` which becomes unused

src/effects/VST3/VST3Utils.cpp
pietro68: Apologies, it looks like the vst3 gain plugin I used for the tests is very unusual - it has 2049 parameters but only 2 of them are "real" and automatable, i.e. gain amount and on/off - and none of the 2049 params is hidden. 

 I tried with another gain plugin and it has only two parameters. So maybe we can leave things as they are?

src/effects/VST3/VST3Utils.h
pietro68: I tried what you suggest, but I came to a point where I had to use the mutable keyword anyway. 

The problem is, when it is time for constructing VST3Wrapper, mEditController is still null - if it was not, I could send it as an arg to the VST3Wrapper ctor, to be used to populate the cache - and there would no need to have a mutable something.

But the fact that the cache can be populated only when mEditController is not null, implies that the cache must be mutable.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -979,11 +942,7 @@ void VST3Effect::ExportPresets(const EffectSettings &) const
       return;
    }
 
-   if(!Vst::PresetFile::savePreset(
-      fileStream,
-      FUID::fromTUID (mEffectClassInfo.ID().data()),
-      mEffectComponent.get(),
-      mEditController.get()))
+   if ( ! VST3Wrapper::SavePreset(fileStream, FUID::fromTUID(mEffectClassInfo.ID().data())) )
```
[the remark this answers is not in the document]
pietro68: Actually, this duplication happened only for LoadPreset. I removed the extra unnecessary conversion in VST3Wrapper::LoadPreset and verified it still works.

src/effects/VST3/VST3Utils.h
Paul-Licameli: What I meant was, the member shared_ptr to the cache can be const, the cache object can be const, but then the members of the cache object need to be mutable.  There is still mutable, but hidden at a lower level.  Is that not what you had to do?

I recommend the shared_ptr so the plug-in and its instances can share just one cached definition of the set of parameters that is initialized only once, so we can be sure all visits are really over the same set of parameters.

I did something like this for AudioUnit, but the details did not require shared_ptr or mutable.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -238,3 +245,132 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
...
+      mEditController.get()
+   );
+}
+
+
+bool VST3Wrapper::FetchSettings(VST3EffectSettings& settings) const
+{
+   settings.mValues.clear();
+
+   return ForEachParameter(
```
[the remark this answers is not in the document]
pietro68: ok, done

src/effects/VST3/VST3Utils.cpp
pietro68: Alright, also done now

src/effects/VST3/VST3Effect.cpp
pietro68: ok

src/effects/VST3/VST3Effect.cpp
pietro68: good point, ok

src/effects/VST3/VST3Effect.cpp
pietro68: ok

pietro68:
Closing this, as a new PR at https://github.com/audacity/audacity/pull/3030 was opened to address the comments here.

Issue #3014, opened 2022-06-01
wxThread destruction is not reliably safe on Linux


Paul-Licameli:
This problem blocks merge of #2941

As reported by @crsib and @ksoze95, on Linux there is a crash on exit of the automated journal file test, for obscue timing-dependent reasons not obviously related to the changes in the PR itself.

Rather than disabling the journal test, let's fix and modernize the code, by abandoning use of `wxThread` for `std::thread`.

PR #3015, opened 2022-06-01
Use std thread

Paul-Licameli:
Resolves: #3014

Eliminate remaning uses of `wxThread`, using `std::thread` instead.

Changes in AudioIO.cpp are known to be necessary to unblock #2941 .  The other two (in LV2Effect and Scrubbing) are not known to be needed, but modernized here too.

Textual scan of the code will still show `wxThread` in Equalization48x, but that is a lot of experimental code that was never improved to be production ready.  I didn't bother.

Recommended:

### DIFF
--- a/src/tracks/ui/Scrubbing.h
+++ b/src/tracks/ui/Scrubbing.h
@@ -11,8 +11,7 @@ Paul Licameli split from TrackPanel.cpp
 #ifndef __AUDACITY_SCRUBBING__
 #define __AUDACITY_SCRUBBING__
 
-
-
+#include <thread>
 #include <vector>
 #include <wx/longlong.h>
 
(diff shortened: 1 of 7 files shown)


src/tracks/ui/Scrubbing.h
```diff
@@ -149,11 +148,17 @@ class AUDACITY_DLL_API Scrubber final
 private:
    void UpdatePrefs() override;
 
+#ifdef USE_SCRUB_THREAD
+   //! @pre `!mThread.joinable()`
+#endif
```
crsib, line 153: `#ifdef` looks a bit strange here. Probably ```//! @pre `!defined(USE_SCRUB_THREAD) || !mThread.joinable()` ``` would be better readable (and will be visible in docs too!)

crsib (approved):
I won't be able to check if it helps on Linux today, but I see no reason to hold this PR

PR #3016, opened 2022-06-01
Realtime effects UI fixes

vsverchinsky:

Recommended:

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -678,6 +678,15 @@ public:
       wxWindowUpdateLocker freeze(this);
       //delete items that were added to the sizer
       mEffectListContainer->GetSizer()->Clear(true);
+
+#ifdef __WXGTK__
+      //Workaround for GTK: if none children were added to the container,
+      //underlying gtk widget size will not be updated, resulting in widget
+      //overlap.
+      if(!mTrack || RealtimeEffectList::Get(*mTrack).GetStatesCount() == 0)
+         mEffectListContainer->Hide();
+#endif
+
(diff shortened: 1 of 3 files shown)


src/RealtimeEffectPanel.cpp
```diff
@@ -678,6 +664,15 @@ class RealtimeEffectListWindow : public wxScrolledWindow
       wxWindowUpdateLocker freeze(this);
       //delete items that were added to the sizer
       mEffectListContainer->GetSizer()->Clear(true);
+
+#ifdef __WXGTK__
+      //Workaround for GTK: if none children were added to the container,
```
Paul-Licameli, line 669: "no children"

src/RealtimeEffectPanel.cpp
```diff
@@ -267,7 +267,7 @@ namespace
       RealtimeEffectState* mEffectState{nullptr};
 
       ThemedButtonWrapper<wxButton>* mChangeButton{nullptr};
-      wxToggleButton* mEnableButton{nullptr};
+      wxButton* mEnableButton{nullptr};
```
Paul-Licameli, line 270: wxBitmapToggleButton exists.  Is it the right thing for you?

Paul-Licameli (approved):

src/RealtimeEffectPanel.cpp
vsverchinsky: Using `wx*ToggleButton` will simplify logic, but it looks noticeably different compared to `wx*Button` in GTK. Few lines of code should be added later to change icon when effect changes it's power state.

PR #3030, opened 2022-06-06
Define settings vst3 i2969 3rd

pietro68:

3rd iteration, trying to address comments on https://github.com/audacity/audacity/pull/3009

Recommended:


src/effects/VST3/VST3Utils.h
```diff
@@ -14,6 +14,10 @@
 
 #include <string>
 #include <pluginterfaces/vst/vsttypes.h>
+#include <pluginterfaces/vst/ivstaudioprocessor.h>
```
Paul-Licameli, line 17: The addition of this `include` of this header allows the deletion of the other two.

src/effects/VST3/VST3Utils.h
```diff
@@ -52,7 +56,73 @@ class VST3Utils final
...
+   //
+   Steinberg::IPtr<Steinberg::Vst::IEditController> mEditController;
+   Steinberg::IPtr<Steinberg::Vst::IComponent>      mEffectComponent;
+
+   using ParameterInfo = Steinberg::Vst::ParameterInfo;
+
+   using ParameterVisitor = std::function< bool(const ParameterInfo& pi) >;
+
+   // made mutable because it is cached once on demand
+   mutable std::vector<ParameterInfo> mParameterInfos;
```
Paul-Licameli, line 81: Here and elsewhere, I suggest you learn the conventions for Doxygen comments.  See : https://www.doxygen.nl/manual/docblocks.html

Simply adding `!` after `//` means a comment will also be extracted to the automatically generated web pages.  If you install the program, you can run `doxygen nographs.dox` to generate web pages on your own computer, then open `dox/html/index.html` to test how it looks.  I say this as a recommendation, not a requirement for merging this.

src/effects/VST3/VST3Utils.h
```diff
@@ -52,7 +56,73 @@ class VST3Utils final
...
+
+   // Visit parameters and keep doing so as long as visitor returns true.
+   //
+   // Return false if the parameters are not yet available
```
Paul-Licameli, line 85: For doxygen, you can write
```
/*!
   @return false if the parameters are not yet available
*/
```

There are a few alternatives of the comment syntax that doxygen also recognizes.

src/effects/VST3/VST3Utils.h
Paul-Licameli: Maybe this mutable member should become private.

src/effects/VST3/VST3Utils.h
```diff
@@ -52,7 +56,73 @@ class VST3Utils final
...
+   //
+   bool ForEachParameter(ParameterVisitor visitor) const;
+
+   // Stop visiting as soon as the visitor returns true, and return true in that case;
+   // otherwise if all visits returned false,  return false.
+   //
+   bool AnyOf(ParameterVisitor visitor) const;
+
+   const std::vector<ParameterInfo>* getParameterInfos() const;
```
Paul-Licameli, line 94: Pointer or reference as return type?  I would prefer reference.  It's a conventional way of saying, null isn't possible, and making it unnecessary to specify that with a `//! @post` comment.

But otherwise, seeing just this declaration, I as a possible caller of this function, would be left in doubt whether null checks on all the uses of this function should be done or not.

Paul-Licameli: Now I see the lines in the function:

```
     if (mEditController == nullptr)
         return nullptr;
```

suggesting that sometimes it's necessary to return null because the effect plugin can't be completely initialized yet.  Is that right?  Then, a comment in the header should explain.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -60,16 +63,133 @@ wxString VST3Utils::MakeAutomationParameterKey(const Steinberg::Vst::ParameterIn
...
-   unsigned long value { };
-   if(idStr.ToULong(&value))
+   if (mParameterInfos.size() == 0)
    {
-      paramId = static_cast<Steinberg::Vst::ParamID>(value);
-      return true;
+      if (mEditController == nullptr)
+         return nullptr;
+
+      for (int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
```
Paul-Licameli, line 74: I have learned to prefer `size_t` not `int` for loops like this.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -60,16 +63,133 @@ wxString VST3Utils::MakeAutomationParameterKey(const Steinberg::Vst::ParameterIn
...
+      {
+         Steinberg::Vst::ParameterInfo vstParameterInfo{ };
+         if (mEditController->getParameterInfo(i, vstParameterInfo) == Steinberg::kResultOk)
+         {
+            // Add all parameterinfos, regardless of whether they can be automated or not,
+            // because different flags of them will be checked by different methods.
+            mParameterInfos.push_back(vstParameterInfo);
```
Paul-Licameli, line 81: Not skipping the hidden parameters?  As you wish.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -331,82 +331,68 @@ bool VST3Effect::SupportsRealtime() const
...
-               mComponentHandler->performEdit(id, value);
+               ownSettings.mValues[parameterInfo.id] = value;;
             }
-            mEditController->setParamNormalized(id, value);
          }
-      } while(parms.GetNextEntry(key, index));
-   }
-
-   if(mPlainUI != nullptr)
-      mPlainUI->ReloadParameters();
```
Paul-Licameli, line 409: Strictly speaking, the deletion of these two lines ought to wait until the commit "Define TransferDataToWindow" so that each step if the transformation is correct, introducing no temporary known bugs.

src/effects/VST3/VST3Utils.h
```diff
@@ -52,7 +56,73 @@ class VST3Utils final
...
+   using ParameterVisitor = std::function< bool(const ParameterInfo& pi) >;
+
+   // made mutable because it is cached once on demand
+   mutable std::vector<ParameterInfo> mParameterInfos;
+
+   // Visit parameters and keep doing so as long as visitor returns true.
+   //
+   // Return false if the parameters are not yet available
+   //
+   bool ForEachParameter(ParameterVisitor visitor) const;
```
Paul-Licameli, line 87: The second commit, which introduces this function, might also rewrite `VST3ParametersWindow::VST3ParametersWindow`

so that the commit comment is correct:  only one call of `getParameterCount` remains

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1200,11 +1149,7 @@ bool VST3Effect::LoadPreset(const wxString& path)
       return false;
    }
 
-   if(!Vst::PresetFile::loadPreset(
-      fileStream,
-      FUID::fromTUID(mEffectClassInfo.ID().data()),
-      mEffectComponent.get(),
-      mEditController.get()))
+   if ( ! VST3Wrapper::LoadPreset(fileStream, FUID::fromTUID(mEffectClassInfo.ID().data())) )
```
Paul-Licameli, line 1152: I still have a feeling that putting `mEffectClassInfo` into the wrapper base class might make nicer simplifications of usage... but I don't insist

src/effects/VST3/VST3Utils.h
```diff
@@ -52,7 +56,73 @@ class VST3Utils final
...
-   //Attempts to extract a unique VST3 parameter id from the key.
-   //Returns true on success.
-   static bool ParseAutomationParameterKey(const wxString& key, Steinberg::Vst::ParamID& paramId);
 };
+
+struct EffectSettings;
+
+struct VST3EffectSettings
+{
+   std::unordered_map<Steinberg::Vst::ParamID, Steinberg::Vst::ParamValue> mValues;
```
Paul-Licameli, line 65: The commit that introduces this line needs `#include <unordered_map>` to compile on macOS

src/effects/VST3/VST3Utils.cpp
```diff
@@ -11,12 +11,15 @@
 **********************************************************************/
 
 #include "VST3Utils.h"
+#include "MemoryX.h"
+#include "EffectInterface.h"
```
Paul-Licameli, line 15: Or this?

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1233,3 +1177,54 @@ void VST3Effect::ReloadUserOptions()
...
+            return false;
+         }
+
+         return true;
+      }
+
+   );
+
+   EffectSettings dummySettings;
+   SyncParameters(dummySettings);
```
Paul-Licameli, line 1224: This on the other hand is here because a call to `SyncParameters` is deleted from `LoadUserPreset`.  But it uses a dummy settings, which is strange; but that is because `SyncParameters` takes a non-const reference to settings, as it must in other contexts.

So I think the visit above should be deleted, and these two lines should be retained, but some comment added to explain that this is only transitional, pending the definition of the Instance class.

We need to discuss with @vsverchinsky .

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1233,3 +1177,54 @@ void VST3Effect::ReloadUserOptions()
...
+
+   if (mPlainUI != nullptr)
+      mPlainUI->ReloadParameters();
```
Paul-Licameli, line 1227: These lines are certainly good ones that will remain in the final version.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -331,82 +331,68 @@ bool VST3Effect::SupportsRealtime() const
...
+   (
+      [&](const ParameterInfo& parameterInfo)
       {
-         if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
+         if (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
          {
             parms.Write(
                VST3Utils::MakeAutomationParameterKey(parameterInfo),
-               mEditController->getParamNormalized(parameterInfo.id)
+               ownSettings.mValues.at(parameterInfo.id)
```
Paul-Licameli, line 363: `at` might allow brevity but it can throw an exception that isn't handled anywhere.  I would rather use `find` and a test against `end` and skip parameters that are not found.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -331,82 +331,68 @@ bool VST3Effect::SupportsRealtime() const
...
+   ownSettings.mValues.clear();
+
+   ForEachParameter
+   (
+      [&](const ParameterInfo& parameterInfo)
       {
-         Steinberg::Vst::ParamID id;
-         Vst::ParamValue value;
-         if(VST3Utils::ParseAutomationParameterKey(key, id) && parms.Read(key, &value))
+         if (parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
```
Paul-Licameli, line 383: This line is added at the last commit, to correspond with what `SaveSettings` does.  But strictly it should be added earlier, in the commit "define the structure, FetchSettings method, StoreSettings method"

src/effects/VST3/VST3Effect.cpp
```diff
@@ -331,82 +331,68 @@ bool VST3Effect::SupportsRealtime() const
...
+            const wxString key = VST3Utils::MakeAutomationParameterKey(parameterInfo);
+            Vst::ParamValue value;
+
+            if (parms.Read(VST3Utils::MakeAutomationParameterKey(parameterInfo), &value))
             {
-               auto cleanup = finally([&]{
-                  mComponentHandler->endEdit(id);
-               });
-               mComponentHandler->performEdit(id, value);
+               ownSettings.mValues[parameterInfo.id] = value;;
```
Paul-Licameli, line 390: extra `;`

Paul-Licameli:
This completes a round of review

src/effects/VST3/VST3Utils.h
pietro68: hmm I could not get rid of the #include <string> though

src/effects/VST3/VST3Utils.h
pietro68: ok, done

src/effects/VST3/VST3Utils.cpp
pietro68: Right, better skip them, done.

src/effects/VST3/VST3Utils.cpp
pietro68: ok, done

src/effects/VST3/VST3Utils.h
pietro68: I take a chance now to refresh my rusty Doxygen knowledge...

src/effects/VST3/VST3Utils.h
pietro68: That's right, it is for the case when mEditController could not be initialized.
Added comment for it.

Paul-Licameli:
`SaveSettings` and `LoadSettings` should not use `ForEachParameter` after all, but iterate over the map (when saving) or, as in Vitaly's original, iterate over the config file (when loading).

I'm reversing what I said before, but in fact approving a point you raised a few weeks ago.

src/effects/VST3/VST3Effect.cpp
pietro68: I forgot about that, now done.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -11,12 +11,15 @@
 **********************************************************************/
 
 #include "VST3Utils.h"
+#include "MemoryX.h"
```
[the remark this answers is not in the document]
pietro68: oops, actually no, it is a leftover from an attempt to find missing headers. Will remove it.

src/effects/VST3/VST3Utils.cpp
pietro68: same as above

src/effects/VST3/VST3Utils.h
pietro68: thanks, added

src/effects/VST3/VST3Effect.cpp
pietro68: I will do an edit on that commit

src/effects/VST3/VST3Effect.cpp
pietro68: ok

src/effects/VST3/VST3Effect.cpp
```diff
@@ -361,10 +361,13 @@ bool VST3Effect::SaveSettings(
...
-               ownSettings.mValues.at(parameterInfo.id)
-            );
+            const auto itr = ownSettings.mValues.find(parameterInfo.id);
+
+            if (itr != ownSettings.mValues.end())
+            {
+               parms.Write(VST3Utils::MakeAutomationParameterKey(parameterInfo),
+                           ownSettings.mValues.at(parameterInfo.id));
+            }
          }
```
pietro68, line 369: woops I just realized that I am searching the map two times - instead of .at() I should have used the iterator directly.

pietro68, line 369: woops I just realized that I am searching the map two times - instead of .at() I should have used the iterator directly.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1233,3 +1177,54 @@ void VST3Effect::ReloadUserOptions()
 
    SetBlockSize(mUserBlockSize);
 }
+
+
+bool VST3Effect::TransferDataToWindow(const EffectSettings& settings)
+{
+   const auto vst3Settings = GetSettings(settings);
+
+   const bool scanOK = ForEachParameter(
```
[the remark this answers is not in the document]
pietro68: I think I must rewrite this method, because it is not doing what is supposed to, i.e. take the input settings and move sliders and knobs on the plugin GUI to reflect their values. 

What it is doing instead, is only passing on the settings to the audio processor (through the ComponentHandler).

Paul-Licameli:
To exercise `SaveSettings` and `LoadSettings`, I think you can use the button that says "Manage" (for plain UI) or shows the "hamburger" icon (for fancy UI) and save a user preset, then load the same again.

Or you can modify your build so that `SupportsRealtime` is true for VST3 -- not expecting all else to work, but it will allow you to make an Audacity project with an effect stack in a track, save it, then reopen it.

pietro68:
closing this PR and moving to #3070

PR #3031, opened 2022-06-06
Adds verbose logging for Journal tests on CI

crsib:
It was problematic to determine the reason why the sanity test has failed,
especially on Windows. 

CTest framework does not allow to get the exit code of the application (or there is no easy way to retrieve it at lease).

To simplify the test run analysis the following was added to the journaling system:

* Journal writes every step and sync exception into `journallog.txt` with aggressive flushing
* On Windows - test run is wrapped into a PowerShell script, that outputs the exit code and collects the logs of the test run. It is unclear, if a similar script is needed on *nix systems

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/AudacityTesting.cmake
+++ b/cmake-proxies/cmake-modules/AudacityTesting.cmake
@@ -101,6 +101,8 @@ if( ${_OPT}has_tests )
       endif()
    endfunction()
 
+   set( JOURNAL_TEST_TIMEOUT_SECONDS 180 )
+
    #[[
       add_journal_test(journal_file)
 
@@ -115,6 +117,14 @@ if( ${_OPT}has_tests )
          # On macOS CMake will generate a placeholder that CTest fails to handle correctly,
          # so we have to setup the path manually
          set( audacity_target "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$<CONFIG>/Audacity.app/Contents/MacOS/Audacity" )
+      elseif (WIN32)
+         set( audacity_target
+            powershell
+               -ExecutionPolicy Bypass
+               -File "${CMAKE_SOURCE_DIR}/tests/journals/test_runner.ps1"
+               "$<TARGET_FILE:Audacity>"
+               --timeout ${JOURNAL_TEST_TIMEOUT_SECONDS}
+         )
       else()
          set( audacity_target "$<TARGET_FILE:Audacity>" )
       endif()
@@ -132,7 +142,7 @@ if( ${_OPT}has_tests )
          ${test_name}
          PROPERTIES
             LABELS "journal_tests"
-            TIMEOUT 180
+            TIMEOUT ${JOURNAL_TEST_TIMEOUT_SECONDS}
       )
    endfunction()
 else()
--- a/src/Journal.cpp
+++ b/src/Journal.cpp
@@ -20,9 +20,14 @@
 #include <algorithm>
 #include <wx/app.h>
 #include <wx/filename.h>
+#include <wx/ffile.h>
+
+#include <string>
+#include <string_view>
 
 #include "MemoryX.h"
 #include "Prefs.h"
+#include "FileNames.h"
 
 namespace Journal {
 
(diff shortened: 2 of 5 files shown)


tests/journals/test_runner.ps1
```diff
@@ -0,0 +1,47 @@
...
+Write-Host "Journal File: $journalFile"
+
+if (-not(Test-Path -Path $journalFile -PathType Leaf)) {
+    Write-Host "Journal file does not exist"
+    exit 1
+}
+
+$process = [Diagnostics.Process]::Start("$audacityExecutable", "--journal $journalFile")
+
+$completedInTime = $process.WaitForExit(180000)
```
Paul-Licameli, line 14: What is the time unit for this number?

Should this timeout be given as an argument and made to agree with whatever timeout CTest is using?

src/widgets/BasicMenu.cpp
```diff
@@ -199,7 +199,9 @@ void ReplayPopup( wxMenu *theMenu )
    }
 
    // Replay did not find all as expected
-   throw Journal::SyncException{};
+   throw Journal::SyncException(wxString::Format(
+      "PopuMenu has failed to invoke %s",
```
Paul-Licameli, line 203: spelling

src/widgets/BasicMenu.cpp
```diff
@@ -199,7 +199,9 @@ void ReplayPopup( wxMenu *theMenu )
...
+      wxJoin(fields, ',').ToStdString().c_str()));
```
Paul-Licameli, line 204: I think `wxJoin(fields, '.')` is all youy need

crsib: wxWidgets uses C va_format, it can only be used with `const char*`

src/Journal.cpp
```diff
@@ -38,11 +43,89 @@ int sLineNumber = -1;
 
 BoolSetting JournalEnabled{ L"/Journal/Enabled", false };
 
+class JournalLogger final
+{
+public:
+   JournalLogger()
+   {
+      wxFileName logFile(FileNames::DataDir(), wxT("journallog.txt"));
```
Paul-Licameli, line 51: Let's avoid new `wxT` and just use the L prefix

src/Journal.cpp
```diff
@@ -38,11 +43,89 @@ int sLineNumber = -1;
...
+}
+
+template<typename T>
+std::string ToString(const T& arg)
+{
+   return std::to_string(arg);
+}
+
+template<typename... Args>
+void Log( std::string_view message, const Args... args )
```
Paul-Licameli, line 87: Maybe `Args && ... args`

src/Journal.cpp
```diff
@@ -38,11 +43,89 @@ int sLineNumber = -1;
...
+{
+   if (message.empty())
+      return;
+
+   constexpr auto n = sizeof...(Args);
+   
+   std::string strings[n];
+   std::size_t i = 0;
+   ((strings[i++] = ToString(args)), ...);
```
Paul-Licameli, line 96: Or maybe perfect forwarding of `args` doesn't matter.

Meanwhile, fold expressions are fun!

Paul-Licameli: `std::string strings[] = { ToString(args), ... };`
Does that compile?

src/Journal.cpp
crsib: Yeah, should be const ref, missed that ;-(

src/Journal.h
```diff
@@ -86,7 +86,7 @@ namespace Journal
    //\brief thrown when playback of a journal doesn't match the recording
    class SyncException : public AudacityException {
    public:
-      SyncException();
+      explicit SyncException(const wxString& message);
```
Paul-Licameli, line 89: I see no nullary ctor any more, so this forced rewrite of all uses.  That's good.  I would just like a comment explaning that the message is for the side logging purposes only and is not written to the output journal.

cmake-proxies/cmake-modules/AudacityTesting.cmake
```diff
@@ -115,6 +115,12 @@ if( ${_OPT}has_tests )
          # On macOS CMake will generate a placeholder that CTest fails to handle correctly,
          # so we have to setup the path manually
          set( audacity_target "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$<CONFIG>/Audacity.app/Contents/MacOS/Audacity" )
+      elseif (WIN32)
+         set( audacity_target
+            powershell
+               -ExecutionPolicy Bypass
+               -File "${CMAKE_SOURCE_DIR}/tests/journals/test_runner.ps1"
+               "$<TARGET_FILE:Audacity>" )
```
Paul-Licameli, line 123: Maybe you will want to do all of this wrapping only conditionally and turn it off once you have debugged the present problems.

Paul-Licameli (approved):

cmake-proxies/cmake-modules/AudacityTesting.cmake
crsib: I really want to have for every run. It doesn't produce to much output in successful runs anyway.

src/Journal.cpp
Paul-Licameli: I was suggesting perfect forwarding which means `&&` but not `const`

Not (necessarily) rvalue references because `Args` is a template parameter.

src/widgets/BasicMenu.cpp
Paul-Licameli: Oho yes.  I had `TranslatableString` on my mind.

cmake-proxies/cmake-modules/AudacityTesting.cmake
Paul-Licameli: As you wish

Issue #3032, opened 2022-06-06
Sidebar effect UI should not push and pop temporary RealtimeEffectState


Paul-Licameli:
This will be a necessary but not sufficient step in making the settings of effects in the sidebar really transform the sound as expected.

It is also a necessary step before #3012 can be fixed.

The code paths for reaching the effect dialog from the side bar and from the main menus need to differ.  This is not all of the fix for #2987 either, but it is needed to unblock some testing to verify #3012

PR #3033, opened 2022-06-06
Effect UI host changes

Paul-Licameli:
Resolves: #3032

Recommended:


src/effects/RealtimeEffectState.h
```diff
@@ -94,11 +109,11 @@ class RealtimeEffectState : public XMLTagHandler
...
 
-   struct Access;
-   struct AccessState;
-   std::shared_ptr<AccessState> mpAccessState; // Destroy before mSettings
-   std::weak_ptr<EffectSettingsAccess> mwAccess;
-
+   // This must not be reset to nullptr while a worker thread is running.
+   // In fact it is never yet reset to nullptr, before destruction.
+   // Destroy before mSettings:
+   AtomicUniquePointer<AccessState> mpAccessState{ nullptr };
```
crsib, line 118: I think `shared_ptr` is the only way to get some guarantees about the object lifetime in MT environments.  Unique-style pointer here isn't really much different in the sense of the tread safety here than the regular unique_ptr,

vsverchinsky (approved):

src/effects/RealtimeEffectState.h
Paul-Licameli: The point of this class was atomicity and RAII.

Atomicity becauase std::atomic alone has a trivial destructor.  I want a non-trivial one, which assumes (as is proper with any object) other synchronization preventing a race between destruction and other use.

I want atomicity in the assignment of the pointer after construction.  The relase memory ordering guarantees that the 
other thread using acquire won't see the pointer value before it can correctly fetch the contents from that pointer.

Issue #3012, opened 2022-05-31
AudioUnits: sliders positions are forgotten


pietro68:
**Describe the bug**
Since commit `6a5645` (on master), AudioUnit effects seem to ignore UI sliders positions, until a nudge to them is given while playing.

**To Reproduce**
0. Delete pluginsettings.cfg before you start Audacity
1. generate a mono track with noise
2. instantiate an AudioUnit AUGraphicEQ
3. select 10 bands configuration, press the button to reset all sliders to 0dB
4. bring down the 1khz slider completely
5. start playback (using the main transport toolbar)

**Actual behavior**
The output is unaffected, however as soon as you move the 1Khz slider even slightly, it starts to be affected.
When pressing stop and then play again, the output will be again unaffected, until the 1Khz slider is again nudged.

**Expected behavior**
Output should be affected by the slider position, without having to nudge it while playing.

**Additional information (please complete the following information):**
 - OS: MacOS 
 - Version: local build done on commit # in description

**Additional context**
Update from Paul: The bug happens if you start with empty PluginSettings.cfg.  But if you make a destructive application of the effect (no matter what settings), then thereafter the bug doesn't happen, and neither does it happen after a restrart.

Ryan-Miller-Audacity:
Testing Completed.

Platform Information:
* Operating Systems: MacOS Monterey 12.3.1

Main Testing Completed:
* Bug as described above has been fixed. 
* AUGraphicEQ stil lfunctions as expected
* Slider changes properly effect playback when utilizing the main transport bar
* Other AU Effects are still working as expected as well
 
Bugs:
* No bugs reported.

PR #3034, opened 2022-06-06
Fix audio units

Paul-Licameli:
Resolves: #3012

Recommended:

### DIFF
--- a/src/effects/audiounits/AudioUnitEffect.cpp
+++ b/src/effects/audiounits/AudioUnitEffect.cpp
@@ -658,6 +658,17 @@ int AudioUnitEffect::ShowClientInterface(
    return dialog.ShowModal();
 }
 
+EffectSettings AudioUnitEffect::MakeSettings() const
+{
+   auto result = StatefulPerTrackEffect::MakeSettings();
+   // Cause initial population of the map stored in the stateful effect
+   if (!mInitialFetchDone) {
+      FetchSettings(GetSettings(result));
+      mInitialFetchDone = true;
+   }
+   return result;
+}
+
 bool AudioUnitEffect::SaveSettings(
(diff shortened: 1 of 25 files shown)


Paul-Licameli:
I have un-drafted this after some significant changes.

src/effects/audiounits/AudioUnitEffect.cpp
```diff
@@ -658,6 +658,17 @@ int AudioUnitEffect::ShowClientInterface(
...
+{
+   auto result = StatefulPerTrackEffect::MakeSettings();
+   // Cause initial population of the map stored in the stateful effect
+   if (!mInitialFetchDone) {
+      FetchSettings(GetSettings(result));
+      mInitialFetchDone = true;
+   }
+   return result;
+}
+
```
pietro68, line 673: `mInitialFetchDone` is a member variable, which controls if local variable `result` will be fetched or not. This looks suspicious. First call to `MakeSettings` will give back an actual fetched result, but following ones will not? or am I missing something?

Paul-Licameli: There is not real statelessness yet.  There is only one real AudioUnitEffectSettings (for each effect).  One time only for that effect, fetch settings so that the map is populated; but give back a dummy EffectSettings `result` that doesn't really contain the state.

Paul-Licameli: Do this side effect before the first `StoreSettings` happens, otherwise `StoreSettings` fails to do anything.

As I recall, your bug reproduced only when pluginsettings.cfg was clean.  But if the graphic EQ had been applied destructively in a previous run of Audacity, then it was not clean.

pietro68 (approved):
I confirmed that the issue is now fixed, and I understand the changes.

PR #3037, opened 2022-06-06
Adds missing preprocessor definitions to the testing frameworks

crsib:
Duplicates some of the preprocessor definitions used to build Audacity and libraries. Adds `UNIT_TEST` preprocessor definition available only for unit tests.

Additionally, fixes some formatting mess on `AudacityFunctions.cmake`

Recommended:


Paul-Licameli:
This commit looks simple enough, when I ignore whitespace changes.

And yet I don't like duplication of literals that are not just coincidentally alike, whether that's C++ or CMake code.  So I would suggest a little extra work to extract a common function callable by `audacity_append_common_compiler_options` and `add_unit_test`.

A suggestion only.

Paul-Licameli (approved):

PR #3041, opened 2022-06-07
Adds basic rendering statistics

crsib:
Add a set of helpers to simplify profiling of the TrackPanel painting. The dialog is only available in alpha builds.

Recommended:

### DIFF
--- a/libraries/lib-graphics/FrameStatistics.cpp
+++ b/libraries/lib-graphics/FrameStatistics.cpp
@@ -1,0 +1,132 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  FrameStatistics.cpp
+
+  Dmitry Vedenko
+
+**********************************************************************/
+
+#include "FrameStatistics.h"
+
+#include <algorithm>
+#include <numeric>
+
+namespace
+{
+FrameStatistics& GetInstance() noexcept
+{
+   static FrameStatistics frameStatistics;
+   return frameStatistics;
+}
+}
+
+FrameStatistics::Stopwatch::~Stopwatch() noexcept
+{
+   GetInstance().AddEvent(mSection, FrameStatistics::Clock::now() - mStart);
+}
+
+FrameStatistics::Stopwatch::Stopwatch(SectionID section) noexcept
+    : mSection(section)
+    , mStart(FrameStatistics::Clock::now())
+{
+}
+
+FrameStatistics::Duration
+FrameStatistics::Section::GetLastDuration() const noexcept
+{
+   return mLastDuration;
+}
+
+FrameStatistics::Duration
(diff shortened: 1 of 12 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -353,6 +355,9 @@ void DrawMinMaxRMS(
    long pixAnimOffset = (long)fabs((double)(wxDateTime::Now().GetTicks() * -10)) +
       wxDateTime::Now().GetMillisecond() / 100; //10 pixels a second
 
+   const auto ms = wxDateTime::Now().GetMillisecond();
+   const auto ticks = (long)fabs((double)(wxDateTime::Now().GetTicks() * -10));
```
vsverchinsky, line 359: unused?

vsverchinsky (approved):

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
crsib: It will be removed in the next PR anyway, but not quite clear how it ended up here anyway

src/FrameStatisticsDialog.cpp
```diff
@@ -0,0 +1,171 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  FrameStatisticsDialog.h
```
Paul-Licameli, line 6: I'm late to this party but... wrong filename here

src/FrameStatisticsDialog.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+      bool Dirty { true };
+   };
+
+   Section mSections[size_t(FrameStatistics::SectionID::Count)];
+
+   Observer::Subscription mStatisticsUpdated;
+};
+
+Destroy_ptr<Dialog> sDialog;
```
Paul-Licameli, line 149: These days, I instead use https://docs.wxwidgets.org/3.0/classwx_window_ptr_3_01_t_01_4.html

libraries/lib-graphics/FrameStatistics.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+   mNextIndex = (mNextIndex + 1) % KERNEL_SIZE;
+
+   if (mKernelItems < KERNEL_SIZE)
+      ++mKernelItems;
+
+   mAvgDuration =
+      std::accumulate(
+         mFilteringKernel, mFilteringKernel + mKernelItems, Duration {}) /
+      mKernelItems;
```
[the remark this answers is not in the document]
Paul-Licameli: That could make the update independent of the denominator and allow you a larger denominator

Paul-Licameli:
That's all I have to say.

PR #3043, opened 2022-06-07
Updated layout of the structures used in Audacity waveform painting code

crsib:
Audacity uses `WaveDisplay` class to organize data required to draw waveforms on screen.

Originally, it was structured as 5 arrays:

* `where` - array of integers, that represent the first sample of each column
* `bl` - block index, which was only used with on-demand rendering (fun fact - the animation code never worked!)
* `min`, `max`, `rms` - extrema and rms values calculated over all samples composing the column

This PR:

* Removes `bl` as on-demand functionality is not present in Audacity 3
* Removes `where` as it can be calculated in constant time
* Converts structure-of-arrays to array-of-structures to improve the cache locality

While it doesn't provide much performance improvement on its own, it lays the fundamentals for the future PRs.

Recommended:
- [?] Each commit compiles and runs on my machine without known undesirable changes of behavior

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
@@ -62,7 +62,7 @@ struct MinMaxSumsq
 }
 
 bool GetWaveDisplay(const Sequence &sequence,
-   float *min, float *max, float *rms, int* bl,
+   float *min, float *max, float *rms,
    size_t len, const sampleCount *where)
 {
    wxASSERT(len > 0);
@@ -141,8 +141,6 @@ bool GetWaveDisplay(const Sequence &sequence,
          : (samplesPerPixel >= 256) ? 256
          : 1;
 
-      int blockStatus = b;
-
       // How many samples or triples are needed?
 
       const size_t startPosition =
@@ -159,7 +157,6 @@ bool GetWaveDisplay(const Sequence &sequence,
          // Do some defense against this case anyway
          while (pixel < nextPixel) {
             min[pixel] = max[pixel] = rms[pixel] = 0;
-            bl[pixel] = blockStatus;//MC
             ++pixel;
          }
          continue;
@@ -243,7 +240,6 @@ bool GetWaveDisplay(const Sequence &sequence,
          // Assign results
          std::fill(&min[pixel], &min[pixelX], values.min);
          std::fill(&max[pixel], &max[pixelX], values.max);
-         std::fill(&bl[pixel], &bl[pixelX], blockStatus);
          std::fill(&rms[pixel], &rms[pixelX], (float)sqrt(values.sumsq / rmsDenom));
 
          pixel = pixelX;
--- a/src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
@@ -63,10 +63,10 @@ struct MinMaxSumsq
 }
 
 bool GetWaveDisplay(
-   const Sequence& sequence, const sampleCount* where, WaveDisplayColumn* columns, size_t len)
+   const Sequence& sequence, const PixelSampleMapper& mapper, WaveDisplayColumn* columns, size_t len)
 {
    wxASSERT(len > 0);
-   const auto s0 = std::max(sampleCount(0), where[0]);
+   const auto s0 = std::max(sampleCount(0), mapper.GetFirstSample(0));
    const auto numSamples = sequence.GetNumSamples();
    if (s0 >= numSamples)
       // None of the samples asked for are in range. Abandon.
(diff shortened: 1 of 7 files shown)


src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.h
```diff
@@ -16,14 +16,14 @@ class Sequence;
 class sampleCount;
 
 // where is input, assumed to be nondecreasing, and its size is len + 1.
-// min, max, rms, bl are outputs, and their lengths are len.
+// min, max, rms, and their lengths are len.
```
Paul-Licameli, line 20: Preserve “are outputs”

src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
```diff
@@ -156,7 +156,10 @@ bool GetWaveDisplay(const Sequence &sequence,
          wxASSERT(false);
          // Do some defense against this case anyway
          while (pixel < nextPixel) {
-            min[pixel] = max[pixel] = rms[pixel] = 0;
+            columns[pixel].min = {};
+            columns[pixel].max = {};
+            columns[pixel].rms = {};
```
Paul-Licameli, line 161: I would write `columns[pixel] = {};` and leave it to https://en.cppreference.com/w/cpp/language/value_initialization

src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
```diff
@@ -196,11 +199,11 @@ bool GetWaveDisplay(const Sequence &sequence,
...
             MinMaxSumsq values(temp.get(), diff, divisor);
             const int lastPixel = pixel - 1;
-            float &lastMin = min[lastPixel];
+            float &lastMin = columns[lastPixel].min;
             lastMin = std::min(lastMin, values.min);
-            float &lastMax = max[lastPixel];
+            float &lastMax = columns[lastPixel].max;
             lastMax = std::max(lastMax, values.max);
-            float &lastRms = rms[lastPixel];
+            float &lastRms = columns[lastPixel].rms;
```
Paul-Licameli, line 206: I would precede all this with `const auto &column = columns[lastPixel];` just to eliminate the need for me to read the repeated subscripting.  But maybe compilers are smart enough to eliminate the implied repeated code.  (Calling Mr. Godbolt?)

src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
```diff
@@ -238,10 +241,12 @@ bool GetWaveDisplay(const Sequence &sequence,
          MinMaxSumsq values(pv, std::max(0, rmsDenom), divisor);
 
          // Assign results
-         std::fill(&min[pixel], &min[pixelX], values.min);
-         std::fill(&max[pixel], &max[pixelX], values.max);
-         std::fill(&rms[pixel], &rms[pixelX], (float)sqrt(values.sumsq / rmsDenom));
+         WaveDisplayColumn column = { values.min, values.max,
+                                      (float)sqrt(values.sumsq / rmsDenom) };
```
Paul-Licameli, line 245: `sqrtf`?  And one less C-cast

src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
```diff
@@ -238,10 +241,12 @@ bool GetWaveDisplay(const Sequence &sequence,
...
 
+         for (auto col = pixel; col < pixelX; ++col)
+            columns[col] = column;
```
Paul-Licameli, line 248: Don't like `std::fill`?

src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
crsib: Nope, sqrt is intentional. Although I dunno how I ended up with C cast here, will fix that.

src/tracks/playabletrack/wavetrack/ui/WaveformCache.cpp
```diff
@@ -11,23 +11,15 @@
...
-      , start(-1)
-      , pps(0)
-      , rate(-1)
-      , where(0)
-      , min(0)
-      , max(0)
-      , rms(0)
-   {
-   }
+   WaveCache() = default;
```
Paul-Licameli, line 22: The initialization of `pps` changed from 0 to -1.  Does it matter?

src/tracks/playabletrack/wavetrack/ui/WaveformCache.cpp
```diff
@@ -228,9 +205,9 @@ bool WaveClipWaveformCache::GetWaveDisplay(
                   sumsq += val * val;
                }
 
-               min[i] = theMin;
-               max[i] = theMax;
-               rms[i] = (float)sqrt(sumsq / len);
+               columns[i].min = theMin;
+               columns[i].max = theMax;
+               columns[i].rms = (float)sqrt(sumsq / len);
```
Paul-Licameli, line 211: or `columns[i] = { theMin, theMax, sqrtf(sumsq / len) };`

src/tracks/playabletrack/wavetrack/ui/GetWaveDisplay.cpp
crsib: Yep, it was carried from the original code

src/tracks/playabletrack/wavetrack/ui/WaveformCache.cpp
crsib: I don't think so, but let's revert for consistency

crsib:
I have addressed your feedback regarding the first 2 commits

ksoze95 (approved):

PR #3047, opened 2022-06-08
VST3 fixes

vsverchinsky:
Resolves: #2555 
Resolves: #3224 

Fixes crash in `IPlugView`, when plugin does not increment reference count of the `IPlugFrame`
Fixes text label values update in plain UI, when plugin does not update internal state in `IEditController:setParamNormalized`

Recommended:

### DIFF
--- a/src/effects/VST3/AudacityVst3HostApplication.cpp
+++ b/src/effects/VST3/AudacityVst3HostApplication.cpp
@@ -1,0 +1,67 @@
+#include "AudacityVst3HostApplication.h"
+
+#include <public.sdk/source/vst/hosting/hostclasses.h>
+#include <public.sdk/source/vst/utility/stringconvert.h>
+
+
+AudacityVst3HostApplication::AudacityVst3HostApplication()
+{
+   FUNKNOWN_CTOR
+}
+
+AudacityVst3HostApplication::~AudacityVst3HostApplication()
+{
+   FUNKNOWN_DTOR
+}
+
+AudacityVst3HostApplication& AudacityVst3HostApplication::Get()
+{
+   static AudacityVst3HostApplication instance;
+   return instance;
+}
+
+
+Steinberg::tresult PLUGIN_API AudacityVst3HostApplication::queryInterface (const char* _iid, void** obj)
+{
+	QUERY_INTERFACE (_iid, obj, FUnknown::iid, IHostApplication)
+	QUERY_INTERFACE (_iid, obj, IHostApplication::iid, IHostApplication)
+
+	*obj = nullptr;
+	return Steinberg::kResultFalse;
+}
+
+Steinberg::uint32 PLUGIN_API AudacityVst3HostApplication::addRef ()
+{
+	return 1;
+}
+
+Steinberg::uint32 PLUGIN_API AudacityVst3HostApplication::release ()
+{
+	return 1;
+}
+
+Steinberg::tresult PLUGIN_API AudacityVst3HostApplication::getName(Steinberg::Vst::String128 name)
+{
+   return VST3::StringConvert::convert ("Audacity VST3 host application", name) ?
+      Steinberg::kResultTrue : Steinberg::kInternalError;
+}
+
+Steinberg::tresult PLUGIN_API AudacityVst3HostApplication::createInstance(Steinberg::TUID cid, Steinberg::TUID _iid, void** obj)
+{
+   using namespace Steinberg;
+
+   FUID classID (FUID::fromTUID (cid));
+	FUID interfaceID (FUID::fromTUID (_iid));
+	if (classID == Vst::IMessage::iid && interfaceID == Vst::IMessage::iid)
+	{
+		*obj = safenew Vst::HostMessage;
+		return kResultTrue;
+	}
+   else if (classID == Vst::IAttributeList::iid && interfaceID == Vst::IAttributeList::iid)
+	{
+		*obj = safenew Vst::HostAttributeList;
+		return kResultTrue;
+	}
+	*obj = nullptr;
+	return kResultFalse;
+}
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -48,6 +48,7 @@
 #include "internal/x11/SocketWindow.h"
 #endif
 
+#include "AudacityVst3HostApplication.h"
 #include "ConfigInterface.h"
 #include "VST3ParametersWindow.h"
 
@@ -57,11 +58,6 @@ namespace {
    constexpr auto processorStateKey  = wxT("ProcessorState");
    constexpr auto controllerStateKey = wxT("ControllerState");
 
-Steinberg::Vst::IHostApplication& LocalContext()
-{
-   static Steinberg::Vst::HostApplication localContext;
-   return localContext;
-}
 
 
 void ActivateDefaultBuses(Steinberg::Vst::IComponent* component, const Steinberg::Vst::MediaType mediaType, const Steinberg::Vst::BusDirection direction, bool state)
@@ -127,7 +123,7 @@ void VST3Effect::Initialize()
    const auto& pluginFactory = mModule->getFactory();
    if(auto effectComponent = pluginFactory.createInstance<Vst::IComponent>(mEffectClassInfo.ID()))
    {
-      if(effectComponent->initialize(&LocalContext()) == kResultOk)
+      if(effectComponent->initialize(&AudacityVst3HostApplication::Get()) == kResultOk)
       {
          if(auto audioProcessor = FUnknownPtr<Vst::IAudioProcessor>(effectComponent))
          {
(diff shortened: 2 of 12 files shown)


Paul-Licameli (approved):
I suggest you correct spelling of one function name to `RegisterParameterControl`.

I approve these changes, but I am disturbed by what this tells us about some VST3 plug-ins.

If I understand this, it means that `getParamNormalized` returns values we can trust if the plug-in is in its initial state or just after it has loaded a preset, and also if its native UI is used, but that if we set a parameter directly in the edit controller, then we can't always expect to read that parameter back again.

This might invalidate some of the assumptions I had been making about the `FetchParameters` and `StoreParameters` functions I asked @pietro68 to write.

Fetch of parameters (from one edit controller), then store of the parameters might work to change audio processing state as expected.  But store followed by fetch might not give back the same values as were stored, correct?

We may need to examine flow of control to understand where we ever fetch after storing and test such paths more carefully.

vsverchinsky:
>If I understand this, it means that getParamNormalized returns values we can trust if the plug-in is in its initial state or just after it has loaded a preset, and also if its native UI is used, but that if we set a parameter directly in the edit controller, then we can't always expect to read that parameter back again.

In case of iZotope `getParamNormalized` always returns 0 and `setParamNormalized` doesn't change anything, but when I tried to change parameters in the GUI these changes were delivered to the `IComponentHandler` as expected so that the host was able to pass them further to the `IAudioProcessor`. I changed plain UI logic so that it does not rely on the `IEditController` implementation correctness.

I guess that "capturing" parameter changes which are delivered to the `IComponentHandler` instead of (re-)reading the `IEditController` state will loose requirements to the plugin implementation correctness

>Fetch of parameters (from one edit controller), then store of the parameters might work to change audio processing state as expected. But store followed by fetch might not give back the same values as were stored, correct?

The only two ways to change the processing state I'm aware of are pass changes to `IAudioProcessor::process` or use `IComponent::setState`, `IEditController::setParamNormalized` is only expected to change GUI state and update internal state of the controller itself.

Paul-Licameli:
> > If I understand this, it means that getParamNormalized returns values we can trust if the plug-in is in its initial state or just after it has loaded a preset, and also if its native UI is used, but that if we set a parameter directly in the edit controller, then we can't always expect to read that parameter back again.
> 
> In case of iZotope `getParamNormalized` always returns 0 and `setParamNormalized` doesn't change anything, but when I tried to change parameters in the GUI these changes were delivered to the `IComponentHandler` as expected so that the host was able to pass them further to the `IAudioProcessor`. I changed plain UI logic so that it does not rely on the `IEditController` implementation correctness.
> 
> I guess that "capturing" parameter changes which are delivered to the `IComponentHandler` instead of (re-)reading the `IEditController` state will loose requirements to the plugin implementation correctness
> 

Did you mean "lose" correctness (bad) or "loosen" requirements on the plug-ins (good)?

So could it be, our `FetchSettings` implementation must use `IComponentHandler` ?

But as I understand, the component handler intercepts *transitions* of settings.  There is still the need to discover the initial *states* of them.  Which, as I said, appearts to be reliable.

> > Fetch of parameters (from one edit controller), then store of the parameters might work to change audio processing state as expected. But store followed by fetch might not give back the same values as were stored, correct?
> 
> The only two ways to change the processing state I'm aware of are pass changes to `IAudioProcessor::process` or use `IComponent::setState`, `IEditController::setParamNormalized` is only expected to change GUI state and update internal state of the controller itself.

Paul-Licameli:
The third commit fails to build on macOS:

```
/Users/paullicameli/Github/audacity2/src/effects/VST3/AudacityVst3HostApplication.cpp:62:18: error: calling a private constructor of class 'Steinberg::Vst::HostAttributeList'
                *obj = safenew Vst::HostAttributeList;
```

Paul-Licameli:
> The third commit fails to build on macOS:
> 
> ```
> /Users/paullicameli/Github/audacity2/src/effects/VST3/AudacityVst3HostApplication.cpp:62:18: error: calling a private constructor of class 'Steinberg::Vst::HostAttributeList'
>                 *obj = safenew Vst::HostAttributeList;
> ```

Strange, it passed CI, but I can't build it.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -199,46 +195,32 @@ VST3Effect::VST3Effect(
...
+         editController = pluginFactory.createInstance<Vst::IEditController>(VST3::UID(controllerCID));
       else
-         wxLogMessage("Cannot create edit controller: failed read controller Class ID from the effect component");
+         throw std::runtime_error("Failed to instantiate edit controller");
    }
-   else
-      mEditController = editController;
+   
+   mEditController = editController;
+   mEditController->initialize(&AudacityVst3HostApplication::Get());
```
Paul-Licameli, line 364: There is now an assumption that editController is not null and initialize succeeds.  Why is it allowable now to remove the checks that the old version had?

src/effects/VST3/memorystream.cpp
```diff
@@ -0,0 +1,319 @@
+//-----------------------------------------------------------------------------
+// Project     : SDK Core
```
Paul-Licameli, line 2: I will skip review of these files, which I assume are copied from a trusted third party library without modifications.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -57,24 +59,170 @@ namespace {
    constexpr auto processorStateKey  = wxT("ProcessorState");
    constexpr auto controllerStateKey = wxT("ControllerState");
 
-Steinberg::Vst::IHostApplication& LocalContext()
+size_t VST3ProcessBlock(
```
Paul-Licameli, line 62: I cut this function and pasted it just before `VST3Effect::ProcessBlock` so that there are fewer difference lines when viewed with whitespace changes ignored.

This makes the commit more easily reviewable.  I suggest you do that before merging it.

Paul-Licameli:
Maybe the last commit can be split into a few

Paul-Licameli (changes requested):
Missing if or what?

Paul-Licameli:
My last comment for this round of review:  I see calls to `getBusInfo` were checked for OK error return before, but not now.  Should you check errors, return boolean from `ActivateMainAudioBuses`, and propagate from `SetupProcessing`?

Should you check returns too from `activateBus` ?

I also wonder whether a de-activation complementary to the activation should be done somewhere else, but if that's an omission, it is not a new omission with this PR.

src/effects/VST3/AudacityVst3HostApplication.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+Steinberg::uint32 PLUGIN_API AudacityVst3HostApplication::release ()
+{
+	return 1;
+}
+
+Steinberg::tresult PLUGIN_API AudacityVst3HostApplication::getName(Steinberg::Vst::String128 name)
+{
+   return VST3::StringConvert::convert ("Audacity VST3 host application", name) ?
```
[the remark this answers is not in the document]
vsverchinsky: It's not displayed anywhere, some plugins might want to get it for display purposes (in theory), but I haven't seen it yet. No comments on that matter in SDK documentation.

src/effects/VST3/VST3Utils.h
```diff
@@ -73,6 +73,9 @@ class PresetsBufferStream : public Steinberg::Vst::BufferStream
 struct EffectSettings;
 struct VST3EffectSettings
 {
+   //VS: is it necessary to convert state to string?
+   //Alternative might be to store IPtr<Vst::IBStream> instead
```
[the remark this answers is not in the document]
vsverchinsky: Ok

src/effects/VST3/VST3Effect.cpp
```diff
@@ -688,7 +744,11 @@ bool VST3Effect::RealtimeAddProcessor(
    {
       auto effect = std::make_unique<VST3Effect>(*this);
       effect->mSetup.processMode = Vst::kRealtime;
-      effect->mSetup.sampleRate = sampleRate;
+      effect->mSetup.sampleRate != sampleRate;
```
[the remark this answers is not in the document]
vsverchinsky: Thanks! Yes, it's a typo, should be `effect->mSetup.sampleRate = sampleRate`

vsverchinsky:
> My last comment for this round of review: I see calls to `getBusInfo` were checked for OK error return before, but not now. Should you check errors, return boolean from `ActivateMainAudioBuses`, and propagate from `SetupProcessing`?
> 
> Should you check returns too from `activateBus` ?

It looks like all I've been overconscious with all those checks returned by almost each sdk method. It's very unexpected for a plugin to reject main bus activation, though I might mistaken, considering the number of plugins out there

> I also wonder whether a de-activation complementary to the activation should be done somewhere else, but if that's an omission, it is not a new omission with this PR.

There's no need to deactivate bus, but as I understand that it's expected that host call `activateBus` each time after it calls `setupProcessing`

In the latest update I also moved `canProcessSampleSize` to `VST3Effect::Initialize` - there is no need to check it more than once

src/effects/VST3/VST3Effect.cpp
```diff
@@ -687,6 +758,10 @@ bool VST3Effect::RealtimeAddProcessor(
       auto effect = std::make_unique<VST3Effect>(*this);
       effect->mSetup.processMode = Vst::kRealtime;
       effect->mSetup.sampleRate = sampleRate;
+      //IAudioProcessor should be cofigured for a different processing mode
```
Paul-Licameli, line 761: "configured"

src/effects/VST3/VST3Effect.cpp
```diff
@@ -671,7 +735,14 @@ size_t VST3Effect::ProcessBlock(EffectSettings &,
 
 bool VST3Effect::RealtimeInitialize(EffectSettings &settings, double sampleRate)
 {
-   mSetup.sampleRate = sampleRate;
+   if(mSetup.sampleRate != sampleRate)
+   {
+      auto setup = mSetup;
+      setup.sampleRate = sampleRate;
+      if(!SetupProcessing(*mEffectComponent.get(), setup))
```
Paul-Licameli, line 742: In this place, you are not careful, as with other calls to `SetupProcessing` to avoid side effects on mSetup in case of failure.  Should that change?

There are also two calls in constructors of VST3Effect but I think I understand that it is not necessary there.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -687,6 +758,10 @@ bool VST3Effect::RealtimeAddProcessor(
       auto effect = std::make_unique<VST3Effect>(*this);
       effect->mSetup.processMode = Vst::kRealtime;
       effect->mSetup.sampleRate = sampleRate;
+      //IAudioProcessor should be cofigured for a different processing mode
+      if(!SetupProcessing(*effect->mEffectComponent.get(), effect->mSetup))
+         return false;
```
Paul-Licameli, line 763: In this place, you are not careful, as with other calls to `SetupProcessing` to avoid side effects on mSetup in case of failure.  Should that change?

There are also two calls in constructors of VST3Effect but I think I understand that it is not necessary there.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -192,54 +246,53 @@ VST3Effect::VST3Effect(
...
+   : VST3Wrapper(std::move(module), std::move(effectClassInfo))
 {
    using namespace Steinberg;
    Initialize();
 
+   //defaults
+   mSetup.processMode = Vst::kOffline;
+   mSetup.symbolicSampleSize = Vst::kSample32;
+   mSetup.maxSamplesPerBlock = mUserBlockSize;
+   mSetup.sampleRate = 44100.0;
```
Paul-Licameli, line 258: These steps are done only in one of two constructors.  Is that right?

Paul-Licameli (approved):

src/effects/VST3/VST3Effect.cpp
vsverchinsky: If realtime effect instance cannot accept configuration than it's completely discarded and `RealtimeAddProcessor` reports failure

src/effects/VST3/VST3Effect.cpp
vsverchinsky: Copy constructor copies 'master' effect's setup and neccessary modifications are applied later in `VST3Effect::RealtimeAddProcessor`

Issue #1607, opened 2021-09-02
Nyquist effects overflow at > 2^31 samples


petersampsonaudacity:
**Describe the bug**
Nyquist effects may not work if used on extremely long selections''' containing more than 2^31 samples (just over 13.5 hours at 44100 Hz).

Originally logged on Bugzilla as P3 439 in 2011

**To Reproduce**
Example 1.
Make a selection of 14 hours duration at 441000 Hz and enter either of the following commands into the Nyquist Prompt:
(get-duration 1)
(print len)

Example 2.
Generate 14 hours of silence at 44.1 kHz then run
(s-max s (const 0.5))

Example 3.
Just run delay.ny on a 14 hour 44100 Hz tone.

Example 4. 
Evaluate (snd-play (s-rest (* 15 60 60))). SND-PLAY doesn't actually play anything; it's there for benchmarking audio computation and all it does is retrieve and discard samples, so it's much faster than writing samples to disk. This expression crashes Nyquist directly.

**Expected behavior**
Nyquist effects should work if used on extremely long selections.

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/131844511-c0dd7fe3-7433-4f1d-a282-e500809f8cfd.png)

**Additional information (please complete the following information):**
 - OS: all Os
 - Version: originally reported in 2011 - but remain extant ten years later with 3.0.4 and alpha 3.1.0
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

**Additional context**
Roger wrote on Bugzilla
I think the problem here is internal to Nyquist. Does anyone actually apply effects to 14+ hours of 44.1K audio, or is this just a what-if scenario? An ideal fix would probably be to bump the sample counts to 64 bits in Nyquist, but I think the most practical thing is to compute when overflow is going to occur and just stop and raise an error in Nyquist to avoid crashing. I will look at this on the Nyquist (upstream) side.

Steve replied
>Surprisingly we do see people wanting to process very long recordings for some very unlikely reasons, for example (an actual example from the Audacity forum) for marine biology research (hydrophone recordings of whales).  I'd be happy to see this demoted to a "won't fix" if it were not for the crashing (which I guess would be very frustrating if you'd been waiting for a 14 hour file to import).  If the crashing can be prevented, then the workaround of processing shorter sections is both simple and obvious. The problem at the moment is that it is not possible to write checks into Nyquist plug-ins against this, because the checks also fail at 2^31 samples.

[...]

PR #3051, opened 2022-06-09
bug 1607 - ensure that NYQ_MAX_LEN is big enough on all supported pla…

SteveDaulton:
bug 1607 - ensure that NYQ_MAX_LEN is big enough on all supported platforms

Resolves: #1607 

Define NYQ_MAX_LEN as LLONG_MAX as this is the same on all supported platforms, whereas LONG_MAX isn't.

Recommended:

### DIFF
--- a/src/effects/nyquist/Nyquist.cpp
+++ b/src/effects/nyquist/Nyquist.cpp
@@ -109,8 +109,9 @@ enum
    ID_FILE = 15000
 };
 
-// Protect Nyquist from selections greater than 2^31 samples (bug 439)
-#define NYQ_MAX_LEN (std::numeric_limits<long>::max())
+// mMaxLen is still required for some plug-ins even though bug 439 appears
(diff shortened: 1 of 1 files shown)


src/effects/nyquist/Nyquist.cpp
```diff
@@ -913,20 +914,7 @@ bool NyquistEffect::Process(EffectInstance &, EffectSettings &settings)
...
-                     .Format((long)NYQ_MAX_LEN, hours);
-               Effect::MessageBox(
-                  message,
-                  wxOK | wxCENTRE,
-                  XO("Nyquist Error") );
-               if (!mProjectChanged)
-                  em.SetSkipStateFlag(true);
-               return false;
-            }
+            wxASSERT(mCurLen <= NYQ_MAX_LEN);
```
crsib, line 917: This assertion is really a no-op

Issue #3058, opened 2022-06-11
Some effect previews stall


Paul-Licameli:
- make some sound, select all
- open a realtime-capable effect, for example Reverb, from the drop-down
- Hit the "Start Playback" button of the effect dialog
- Hit Apply

Observe:  a stall of a few seconds before playback stops and the effect applies destructively (this might vary with operating system?)
Expect: prompt stop before effect applies

dozzzzer:
Verified on MacOS 12.3.1. and Win 10 x64.

PR #3059, opened 2022-06-11
Don't stall when Apply button of effect dialog stops play...

Paul-Licameli:
... Swap the main and side of EffectSettingsAccessTee, so that Set() uses
the SimpleEffectSettingsAccess -- not the complicated accessor for the temporary
realtime effect state; the one that waits for an "echo"

Resolves: #3058

Recommended:


pietro68 (approved):
I confirm that the fix works; the code change is simple but to be honest it implies some knowledge that I do not have - we can have a talk about it later.

Issue #3061, opened 2022-06-11
Reverb "wet only" checkbox applies to play in progress only when dialog closes


Paul-Licameli:
Update:  Actually, this is a problem very specific to just this one checkbox.  No such problem with the "dry gain" slider, for instance.

- Generate a click track, open the side bar, add Reverb
- Press play and listen
- While playing, press the gear wheel, and check WetOnly
- Press OK

Observe:  the sound really did change according to the settings, but only after OK
Expect: sound changes prompty when you check the box
Should also expect:  if you then Cancel, sound changes back to original settings

Paul-Licameli:
I notice that I can repeat the procedure of opening the dialog, changing settings, and closing during one play-back.  That is some reassurance.  But there is the unexplained #3060

pietro68:
I tried to reproduce the bug on a build of master, at f84dbec - by the time I press OK, an unhandled exception happens:

![image](https://user-images.githubusercontent.com/16491307/173291408-8cf201da-5b43-45ba-8c72-56d0dff2da13.png)

PR #3062, opened 2022-06-11
"Wet only" checkbox of Reverb works in real-time

Paul-Licameli:
Resolves: #3061

Recommended:

### DIFF
--- a/src/effects/Reverb.cpp
+++ b/src/effects/Reverb.cpp
@@ -149,6 +149,8 @@ struct EffectReverb::Validator
 
 #undef SpinSliderHandlers
 
+   void OnCheckbox(wxCommandEvent &evt);
+
 };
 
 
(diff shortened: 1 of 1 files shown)


pietro68:
I could not reproduce the original issue, however I found a problem while testing the Reverb on its own - as soon as I click the `Wet Only` checkbox, a stack overflow happens - by looking at the stack, it looks like an infinite recursion is happening:

image

Paul-Licameli:
If you think my pr #3059 fixed the infinite recursion (I never saw this problem on Mac), then please reevaluate this PR, which I rebased onto the other after merging that to master.

Paul-Licameli:
If you see that infinite recursion again, find me some stack levels if you can, lower on the page than all those repetitions.

pietro68 (approved):
Verified it works and understood code changes.

Issue #3050, opened 2022-06-09
Clicking effect name in sidebar to replace


Paul-Licameli:
Clicking name of an effect in the sidebar should bring up a menu to replace it with another.

The change should also be undo-able.

petersampsonaudacity:
Testing on W10 with latest 3.2.0 alpha master audacity-win-3.2.0-alpha-20220608+a2b9986-x64-msvc2022

Clicking name of an effect in the sidebar **_does_** bring bring up a menu - bit it does not make the replavement if to click on an alternate effect.

![image](https://user-images.githubusercontent.com/50205984/172894375-e603f79d-e6e5-4d47-a7a6-05ff8a05506b.png)

dozzzzer:
Verified on Windows 11 and MacOS Monterey.

petersampsonaudacity:
Testing on W10 with latest alpha master:  audacity-win-3.2.0-alpha-20220704+679d6b6-x64-msvc2022

I confirm this now works on W10

PR #3064, opened 2022-06-13
Effect replacement in sidebar

Paul-Licameli:
Resolves: #3050
Resolves: #3092 

Implement the replacement of a chosen effect in the sidebar by the menu button.

Also fix minor omissions in recent UI changes.  Allow the text in the sidebar to wrap,
and update the most recent undo state for changes of power button state.

Recommended:

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -708,6 +708,21 @@ public:
 
    void OnEffectListItemChange(const RealtimeEffectListMessage& msg)
    {
+      const auto insertItem = [this, &msg](){
+         auto& effects = RealtimeEffectList::Get(*mTrack);
+         InsertEffectRow(msg.srcIndex, effects.GetStateAt(msg.srcIndex));
+      };
+      const auto removeItem = [this, &msg](){
+         auto sizer = mEffectListContainer->GetSizer();
+         auto item = sizer->GetItem(msg.srcIndex)->GetWindow();
+         item->Destroy();
+#ifdef __WXGTK__
+         // See comment in ReloadEffectsList
+         if(sizer->IsEmpty())
+            mEffectListContainer->Hide();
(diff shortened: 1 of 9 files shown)


petersampsonaudacity:
Tested on W10 with @Paul-Licameli 's branch:  audacity-win-3.2.0-alpha-20220613+4f0e7cc-x64

The Effect in the Effects Stack can be changed now by clicking on an existinf effect in the stack and selecting a different one - but I only see two to choose from

src/RealtimeEffectPanel.cpp
```diff
@@ -450,11 +458,40 @@ namespace
...
+         auto &em = RealtimeEffectManager::Get(*mProject);
+         auto oIndex = em.FindState(&*mTrack, mEffectState);
+         if (!oIndex)
+            return;
+
+         auto oldName = GetEffectName();
+         auto &project = *mProject;
+         auto trackName = mTrack->GetName();
+         if (auto state = AudioIO::Get()
+            ->ReplaceState(*mProject, &*mTrack, *oIndex, effectID)
```
ksoze95, line 479: Can replace *mProject argument with project instead.

src/effects/RealtimeEffectManager.cpp
```diff
@@ -265,89 +243,112 @@ void RealtimeEffectManager::AllListsLock::Reset()
...
+{
+   auto [pLeader, states] = FindStates(mProject, pTrack);
+   auto pOldState = states.GetStateAt(index);
+   if (!pOldState)
+      return nullptr;
+   auto pNewState = MakeNewState(pScope, pTrack, id);
+   if (!pNewState)
+      return nullptr;
+
+   // Only now add the completed state to the list, under a lock guard
```
ksoze95, line 317: Shouldn't this comment mention replacing rather than adding.

Issue #3021, opened 2022-06-03
vst3 plain gui: accessibility of check boxes


DavidBailes:
**Describe the bug**
The mouse interaction and focus of check boxes in the vst3 plain gui is non-standard, and the names of the check boxes are not read by screen readers.

**To Reproduce**
vst3 plugin used in the examples: https://valhalladsp.com/shop/delay/valhalla-freq-echo/

Steps to reproduce the non-standard focus:
1. Open a vst3 effect set to show the plain gui, and tab to a checkbox.

**Actual behavior**
The dotted outline indicating the focus is in the wrong place and contains no text:
![image](https://user-images.githubusercontent.com/11887021/171832430-1ddb0826-e33b-49b4-9b35-056f1cf86091.png)

**Expected behavior**
This is check box which is the focus, taken from preferences. The outline indicating the focus surrounds the the name of the check box.
![image](https://user-images.githubusercontent.com/11887021/171832753-bbea1e86-1f2b-4fdc-94f9-069fef734255.png)

Steps to reproduce the non-standard mouse interaction:
1. Open a vst3 effect set to show the plain gui.
2. Click the name of a check box to change its state. For users who find it difficult to position a mouse accurately it's useful to be able to change the state of a check box by clicking the name as well as the box, as this provides a bigger target.

**Actual behavior**
Nothing happens. (You can change the state by clicking within the dotted outline which contains no text, but this is non-standard and the user would have to realise that is the case.)

**Expected behavior**
The state of the check box is changed.

Steps to reproduce the name of the text box not being read by screen readers:
1. On Windows, start the Narrator screen reader (Ctrl+Windows key+Enter).
2. Open a vst3 effect set to show the plain gui.
3. Tab to a check box.

**Actual behavior**
The name of the check box is not read.

**Expected behavior**
The name of the check box is read.

**Additional information (please complete the following information):**
 - OS: Windows 10.
 - Version Audacity 3.2.0 alpha

**Additional context**
For standard check boxes in wxWidgets, where the name is to the right of the check box, then Windows automatically provides the name of the check box to screen readers. So it is only if the check box remains as it is that code will have to written to provide an accessibility name for check boxes.

DavidBailes:
@dozzzzer . Given that this bug includes the name of a control not being read by screen readers, I think that the priority should be P2 rather than P3. And this would be in line with your P2 setting for: https://github.com/audacity/audacity/issues/3039.

DavidBailes:
[...]

PR #3066, opened 2022-06-13
issue #3021: vst3 plain gui: accessibility of check boxes

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/3021

Problem:
The mouse interaction and focus of check boxes in the vst3 plain gui is non-standard, and the names of the check boxes are not read by screen readers.

This is caused by creating a wxCheckBox with an empty label, and adding the text separately.

Fix:
Use a wxCheckBox with a label, and with the wxALIGN_RIGHT style so that the label is to the left of the check box which is the desired layout in this case.

Recommended:


DavidBailes:
Closing and will open new pull request having fixed the conflicts.

PR #3070, opened 2022-06-14
Define settings vst3 i2969 4th

pietro68:

After some reviews and discussions we decided that settings should not hold maps of IDs and normalized values, but rather the whole "state" of the processor and the controller, as created by the VST3 SDK.

This implies such large ripple effects in the code, that this new PR had to be opened.

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -327,66 +327,50 @@ bool VST3Effect::SupportsAutomation() const
 }
 
 bool VST3Effect::SaveSettings(
-   const EffectSettings &, CommandParameters & parms) const
+   const EffectSettings& settings, CommandParameters& parms) const
 {
-   if(mEditController == nullptr)
-      return false;
+   const auto& vst3Settings = GetSettings(settings);
 
-   using namespace Steinberg;
-   
-   for(int i = 0, count = mEditController->getParameterCount(); i < count; ++i)
-   {
-      Vst::ParameterInfo parameterInfo { };
-      if(mEditController->getParameterInfo(i, parameterInfo) == kResultOk)
-      {
-         if(parameterInfo.flags & Vst::ParameterInfo::kCanAutomate)
-         {
-            parms.Write(
-               VST3Utils::MakeAutomationParameterKey(parameterInfo),
-               mEditController->getParamNormalized(parameterInfo.id)
-            );
-         }
-      }
-   }
+   if (vst3Settings.mProcessorStateStr)
+      parms.Write( processorStateKey,  wxString( *vst3Settings.mProcessorStateStr )  );
+
+   if (vst3Settings.mControllerStateStr)
+      parms.Write( controllerStateKey, wxString( *vst3Settings.mControllerStateStr) );
 
    return true;
 }
 
 bool VST3Effect::LoadSettings(
-   const CommandParameters & parms, EffectSettings &settings) const
+   const CommandParameters& parms, EffectSettings& settings) const
 {
-   using namespace Steinberg;
+   auto& vst3Settings = GetSettings(settings);
 
-   if(mComponentHandler == nullptr)
-      return false;
-   
-   long index { };
-   wxString key;
-   if(parms.GetFirstEntry(key, index))
+   wxString tmpString;
+
+   if (parms.Read(processorStateKey, &tmpString))
    {
-      do
-      {
-         Steinberg::Vst::ParamID id;
-         Vst::ParamValue value;
-         if(VST3Utils::ParseAutomationParameterKey(key, id) && parms.Read(key, &value))
-         {
-            if(mComponentHandler->beginEdit(id) == kResultOk)
-            {
-               auto cleanup = finally([&]{
-                  mComponentHandler->endEdit(id);
-               });
-               mComponentHandler->performEdit(id, value);
-            }
-            mEditController->setParamNormalized(id, value);
-         }
-      } while(parms.GetNextEntry(key, index));
-   }   
+      vst3Settings.mProcessorStateStr = tmpString;
+   }
+   else
+   {
+      vst3Settings.mProcessorStateStr = std::nullopt;
+   }
 
+
+   if (parms.Read(controllerStateKey, &tmpString))
+   {
+      vst3Settings.mControllerStateStr = tmpString;
+   }
+   else
+   {
+      vst3Settings.mControllerStateStr = std::nullopt;
+   }
+   
    return true;
 }
 
 bool VST3Effect::LoadUserPreset(
-   const RegistryPath& name, EffectSettings &settings) const
+   const RegistryPath& name, EffectSettings& settings) const
 {
    using namespace Steinberg;
 
@@ -414,6 +398,8 @@ bool VST3Effect::LoadUserPreset(
          mEditController->setState(controllerState) != kResultOk)
          return false;
    }
+
+   FetchSettings(GetSettings(settings));
 
    return true;
 }
@@ -459,13 +445,13 @@ RegistryPaths VST3Effect::GetFactoryPresets() const
    return mFactoryPresets;
 }
 
-bool VST3Effect::LoadFactoryPreset(int id, EffectSettings &) const
+bool VST3Effect::LoadFactoryPreset(int id, EffectSettings& settings) const
 {
    if(id >= 0 && id < mFactoryPresets.size())
    {
       auto filename = wxFileName(GetFactoryPresetsPath(mEffectClassInfo), mFactoryPresets[id] + ".vstpreset");
       // To do: externalize state so const_cast isn't needed
-      return const_cast<VST3Effect*>(this)->LoadPreset(filename.GetFullPath());
+      return const_cast<VST3Effect*>(this)->LoadPreset(filename.GetFullPath(), settings);
    }
    return true;
 }
@@ -871,6 +857,9 @@ bool VST3Effect::ValidateUI(EffectSettings &settings)
 {
    if (mDuration != nullptr)
       settings.extra.SetDuration(mDuration->GetValue());
+
+   FetchSettings(GetSettings(settings));
+
    return true;
 }
 
@@ -938,7 +927,7 @@ void VST3Effect::ExportPresets(const EffectSettings &) const
    }
 }
 
-void VST3Effect::ImportPresets(EffectSettings &)
+void VST3Effect::ImportPresets(EffectSettings& settings)
 {
    using namespace Steinberg;
 
@@ -956,7 +945,7 @@ void VST3Effect::ImportPresets(EffectSettings &)
    if(path.empty())
       return;
 
-   LoadPreset(path);
+   LoadPreset(path, settings);
 }
 
 bool VST3Effect::HasOptions()
@@ -1112,7 +1101,7 @@ void VST3Effect::SyncParameters() const
    }
 }
 
-bool VST3Effect::LoadPreset(const wxString& path)
+bool VST3Effect::LoadPreset(const wxString& path, EffectSettings& settings)
 {
    using namespace Steinberg;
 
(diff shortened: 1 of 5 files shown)


src/effects/VST3/VST3Utils.h
```diff
@@ -56,3 +57,25 @@ class VST3Utils final
...
+   // otherwise the destruction of mEditController and mEffectComponent would trigger a memory fault.
+   std::shared_ptr<VST3::Hosting::Module> mModule;
+
+   const VST3::Hosting::ClassInfo mEffectClassInfo;
+
+   // For the time being, here we have only members that are needed
+   // to iterate parameters and extract preset state
+   //
+   Steinberg::IPtr<Steinberg::Vst::IEditController> mEditController;
+   Steinberg::IPtr<Steinberg::Vst::IComponent>      mEffectComponent;
```
Paul-Licameli, line 99: Maybe these two lines should be switched, just to be sure that their sequence remains the same as it was in VST3Effect before, so the sequence of their destruction is the same.

Or maybe it makes not difference.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -73,3 +76,70 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
...
+   if (!vst3settings.mProcessorStateStr)
+      return false;
+
+   auto processorState = PresetsBufferStream::fromString(vst3settings.mProcessorStateStr.value());
+   if (mEffectComponent->setState(processorState) != kResultOk)
+      return false;
+
+   if (vst3settings.mControllerStateStr)
+   {
+      auto controllerState = PresetsBufferStream::fromString(vst3settings.mControllerStateStr.value());
```
Paul-Licameli: Remove the two uses of `std::optional::value` above and use the `*` operator instead.

It breaks the Mac build, as do certain other C++17 library functions that can throw new kinds of exceptions.  But on Mac they can't build and link unless a certain minimum version of the operating system is targeted, above our present minimum.

The difference with `*` is that is just gives undefined behavior if there is no value, not an exception.

But you have the proper null (or should I say nullopt) checks before use so it is safe.

Paul-Licameli (changes requested):

src/effects/VST3/VST3Utils.h
```diff
@@ -15,6 +15,9 @@
 #include <pluginterfaces/vst/ivstaudioprocessor.h>
 #include <string>
 #include <public.sdk/source/vst/hosting/module.h>
+#include <public.sdk/source/vst/vstpresetfile.h>
```
Paul-Licameli, line 18: This inclusion at the second commit lets you remove the header in one other place.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -73,3 +76,70 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
...
 
 }
+
+
+
+Steinberg::IPtr<PresetsBufferStream> PresetsBufferStream::fromString(const wxString& str)
+{
+   Steinberg::Buffer buffer(str.size() / 4 * 3);
+   auto len = Base64::Decode(str, buffer);
+   assert(len <= buffer.getSize());
```
Paul-Licameli, line 86: You didn't write this assertion -- you only move a line that existed in another file.

But read this uncommented assertion and ask, what's the proof?

The proof should ideally be deduced from a postcondition stated explicitly in a comment in class Base64.

And size_t would be a better return type for `Decode`.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -10,14 +10,17 @@
 
 **********************************************************************/
 
+
 #include "VST3Utils.h"
+#include "Base64.h"
```
Paul-Licameli, line 15: Again you add an #include and should check whether you can remove the same file elsewhere.  And you can.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -73,3 +76,70 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
...
+
+
+wxString PresetsBufferStream::toString() const
+{
+   return Base64::Encode(mBuffer, mBuffer.getSize());
+}
+
+
+
+bool VST3Wrapper::FetchSettings(VST3EffectSettings& vst3Settings) const
```
Paul-Licameli, line 102: `FetchSettings` is given settings that are not assumed to be initially empty.  Be sure to assign nullopt along paths that fail to get state from the plug-in.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -329,64 +329,36 @@ bool VST3Effect::SupportsAutomation() const
...
-               mEditController->getParamNormalized(parameterInfo.id)
-            );
-         }
-      }
-   }
+   if (vst3Settings.mProcessorStateStr)
+      parms.Write( processorStateKey,  wxString( vst3Settings.mProcessorStateStr.value() )  );
+
+   if (vst3Settings.mControllerStateStr)
+      parms.Write( controllerStateKey, wxString( vst3Settings.mControllerStateStr.value() ) );
```
Paul-Licameli: `value` again twice needs to be rewritten

src/effects/VST3/VST3Effect.cpp
```diff
@@ -329,64 +329,36 @@ bool VST3Effect::SupportsAutomation() const
...
 
    return true;
 }
 
 bool VST3Effect::LoadSettings(
-   const CommandParameters & parms, EffectSettings &settings) const
+   const CommandParameters& parms, EffectSettings& settings) const
```
Paul-Licameli, line 346: `LoadSettings`, like `FetchSettings`, should make the given settings reflect exactly what is in the instance or the config file.  So, add other branches that assign nullopt when the configuration does not contain a string.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -329,64 +329,36 @@ bool VST3Effect::SupportsAutomation() const
 }
 
 bool VST3Effect::SaveSettings(
-   const EffectSettings &, CommandParameters & parms) const
+   const EffectSettings& settings, CommandParameters& parms) const
```
Paul-Licameli, line 332: It is documented that `SaveSettings` may assume `parms` is empty.  This is unlike `LoadSettings` which cannot assume `settings` begins empty.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -128,13 +128,13 @@ bool VST3Wrapper::StoreSettings(const VST3EffectSettings& vst3settings) const
...
 
-   auto processorState = PresetsBufferStream::fromString(vst3settings.mProcessorStateStr.value());
+   auto processorState = PresetsBufferStream::fromString(*vst3settings.mProcessorStateStr);
    if (mEffectComponent->setState(processorState) != kResultOk)
       return false;
 
    if (vst3settings.mControllerStateStr)
    {
-      auto controllerState = PresetsBufferStream::fromString(vst3settings.mControllerStateStr.value());
+      auto controllerState = PresetsBufferStream::fromString(*vst3settings.mControllerStateStr);
```
Paul-Licameli, line 137: You anticipated me.  I see this now, taking the commits in sequence.  Good:  but of course rebase interactively and squash the fixes onto earlier commits before this is merged.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -942,11 +942,7 @@ void VST3Effect::ExportPresets(const EffectSettings &) const
       return;
    }
 
-   if(!Vst::PresetFile::savePreset(
-      fileStream,
-      FUID::fromTUID (mEffectClassInfo.ID().data()),
-      mEffectComponent.get(),
-      mEditController.get()))
+   if (!VST3Wrapper::SavePreset(fileStream, FUID::fromTUID(mEffectClassInfo.ID().data())))
```
Paul-Licameli, line 914: The wrapper base class already hass `mEffectClassInfo` so this call and the call to `LoadPreset` can pass just one argument.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1129,6 +1129,8 @@ bool VST3Effect::LoadPreset(const wxString& path)
    if(mPlainUI != nullptr)
       mPlainUI->ReloadParameters();
 
+   FetchSettings(GetSettings(settings));
```
Paul-Licameli, line 1132: The commit that inserts the `FetchSettings` calls is good.  It should be squashed with the earlier commit that rewrites `LoadSettings` and `SaveSettings`.  That commit isn't complete without the calls to `FetchSettings` because `SaveSettings` is assuming that the fetch was already done, and is not correct without it.

Paul-Licameli:
That's a round of review.  I understand this is still incomplete.  Two other boxes at least still to check off.

src/effects/VST3/VST3Utils.h
```diff
@@ -12,8 +12,9 @@
 
 #pragma once
 
+#include <pluginterfaces/vst/ivstaudioprocessor.h>
```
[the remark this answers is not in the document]
pietro68: ok, done

src/effects/VST3/VST3Utils.h
pietro68: done

src/effects/VST3/VST3Utils.cpp
pietro68: done

src/effects/VST3/VST3Utils.cpp
pietro68: ok, done

src/effects/VST3/VST3Utils.cpp
pietro68: ok, done

src/effects/VST3/VST3Utils.h
pietro68: Right, done

src/effects/VST3/VST3Effect.cpp
pietro68: ok, done

src/effects/VST3/VST3Effect.cpp
pietro68: I assume that by this you mean that in `LoadSettings `I should add `else `branches to assign nullopt in case something could not be read, as you comment later.

src/effects/VST3/VST3Effect.cpp
pietro68: ok, done

src/effects/VST3/VST3Utils.cpp
Paul-Licameli: The second commit in the sequence is still not buildable.

Rebase and squash the fix onto the second commit.

Paul-Licameli:
The commit that inserts calls to `FetchSettings` should also insert one at the end of `LoadUserPreset`.

I overlooked that in earlier rounds of review.

src/effects/VST3/VST3Utils.cpp
pietro68: I interpret your comment in this way:

1) remove the assertion, and write a comment in class Base64 that the returned value should be <= buffer.getSize()
2) size_t len = static_cast<size_t>( Base64::Decode(str, buffer) );

or do you mean
2) rewrite Base64::Decode so that it returns a size_t instead of an int ?

Paul-Licameli:
The latest version of the branch still has these omissions:

- The commit that introduces call to `FetchSettings` still needs a call in `LoadUserPreset`.
- The `Define MakeSettings` commit is only adding one boolen data member.  (There was a more complete commit at 3af23e931425a532e790924ab8fcd8ae936aa661 in an earlier rebasing.  Something got lost.)

Paul-Licameli (approved):

Issue #2162, opened 2021-11-19
Punch and roll: tracks are not played in sync.


DavidBailes:
**Describe the bug**
Punch and roll: tracks are not played in sync.

**To Reproduce**
1. Record yourself counting to ten.
2. Duplicate the track.
3. In the first track, click to position the cursor about half way through.
4. With overdub enabled, press shift + D for punch and roll.
5. Playback starts at the pre roll point in track 1, but at the cursor in track 2. The playback of the two tracks is out of sync.

**Expected behavior**
The playback of the tracks should be in sync.

**Additional information (please complete the following information):**
- Win 10
- Audacity 3.1.2

**Additional context**
This is a regression wrt 3.0.5.
Reported on the Audacity4Blind list.

Paul-Licameli:
Thank you for finding this.

It's not good, however it won't be a problem for a narrator just laying down one mono track.

DavidBailes:
A report of the bug on the forum yesterday:
https://forum.audacityteam.org/viewtopic.php?f=46&t=121734&sid=d2bd32f5b1266a84b0908f51b58c9c6b

SteveDaulton:
and another report on the forum here: https://forum.audacityteam.org/viewtopic.php?f=70&t=121679

jserquera:
I teach punch recording and my students cannot work with this bug. Next semester I will have to point the new students to this download page:

https://github.com/audacity/audacity/releases/tag/Audacity-3.1.0 

Which is an inconvenience.

Higher priority, please.

petersampsonaudacity:
Yet another user has complained about this on the Forum today:
https://forum.audacityteam.org/viewtopic.php?f=70&t=121679&p=442336#p442336

This was logged as an issue against 3.1.2 back in mid-November and was not fixed for 3.1.3.  It would seem to be a serious problem particularly for readers of audiobooks and those making podcasts - I really think this should be addressed for the 3.2.0 release @crsib @Paul-Licameli

petersampsonaudacity:
And another user complaint today from the Forum:
https://forum.audacityteam.org/viewtopic.php?f=47&t=122453&p=442372

petersampsonaudacity:
@Penikov @LWinterberg @crsib @Paul-Licameli   -  another complaint about this today on the Forum:

https://forum.audacityteam.org/viewtopic.php?f=70&t=121679

Is there any chance this bug can be addressed and fixed for the upcoming 3.2.0 ?

[...]

PR #3071, opened 2022-06-14
issue #2162 Punch and roll: tracks are not played in sync

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/2162

I think the problem started in master with commit:30b7882e107c32

Fix:
If there are any preroll tracks, set the start of all the playback tracks to the same value.

Recommended:


petersampsonaudacity:
Testing on W10 with @DavidBailes  branch build:  audacity-win-3.2.0-alpha-20220614+564d2a6-x64

David's branch build appears to fix this problem.

Paul-Licameli (approved):

Issue #3029, opened 2022-06-05
Huffman Data Overrun (v3.1.3), Redux


pavakhov:
.**Describe the bug**
Audacity refuses to open mp3 files, reporting "Import Failed", or "Huffman data overrun".  Every other player/editor tested opened/edited the files without any hint of a problem.  *Just Audacity*

This is still an issue, with v.3.1.3.  (Alpha build is old, from July 2021 v3.0.3):
 Madplay (which also uses the libmad library) reports "Huffman data overrun" on Frame0 or Frame2 of an mp3 file created from an Icecast stream using ffmpeg.  This has an impact for staff reviewing/editing of hourly airchecks recorded at a radio station.

**To Reproduce**
1. I can send a link to a problematic file.
2. Open/Import Audio mp3 file
3. Error dialog box opens, refusing to open a perfectly fine mp3 file.

**Actual behavior**
File is not opened

**Expected behavior**
Open mp3 file with or without a warning.

**Screenshots**
This is known behaviour, righteously identified as desired since libmad reports the error.  

**Additional information (please complete the following information):**
 - OS:Windows 10, Debian Bookworm, OS x 
 - Audacity V. 3.1.3 (and v2.4.2 etc...)

**Additional context**

petersampsonaudacity:
Possibly the same as (or a close cousin of) this previously logged issue: 
Audacity doesn't recognize a mp3 file even though it's a mp3 file. #2634

pavakhov:
Hi - thanks for the response.  I looked into this a bit more, and the following from the 'madplay' man page, I think, is relevant: 

> error: frame #: bad main_data_begin pointer
>               This  message  can occur while decoding a Layer III stream that has been cut or spliced without preserving its bit reservoir. The affected frame cannot be properly decoded,
>               but will be used to help restore the bit reservoir for following frames.
> 
Since I am creating an mp3 file from a continuous stream, the bit reservoir may be lost.  Same thing can happen with ffmpeg and the copy codec:
> ffmpeg -i in_file.mp3 -acodec copy -ss $start_time -t 1 out_file.mp3
Using 'madplay -v' on the result often results in a 'Huffman data overrun' error.

If the original file, 'in_file.mp3' is encoded without a bit reservoir:
> lame -b 128 --nores  in_file.wav nores_file.mp3
Then I never see this error when trimming an mp3 file loss-lessly as above with ffmpeg

[...]

Issue #2634, opened 2022-03-05
Audacity doesn't recognize a mp3 file even though it's a mp3 file.


julxploitrr:
**Describe the bug**
When I import a MP3 file, it shows me that: Audacity did not recognize the type of the file.

**To Reproduce**
1: Installing FFmpeg.

**Expected behavior**
Error messages showing up saying 'Audacity did not recognize the type of the file'

**Screenshots**
![image](https://user-images.githubusercontent.com/100502245/156877713-614d4466-ed86-4674-b8d5-0a80562979b9.png)

**Additional information (please complete the following information):**
 - OS: Windows 10
 - Version: 3.1.3

**Additional context**
This is a major issue happening to a lot of other people as well.

Penikov:
Could not reproduce:

![image](https://user-images.githubusercontent.com/25851359/156877958-d04659fa-3705-4fe6-8f28-ec22a0569902.png)
![image](https://user-images.githubusercontent.com/25851359/156877982-5bb50aeb-7da0-4903-9f7c-af8fb4beae9c.png)
![image](https://user-images.githubusercontent.com/25851359/156878129-d18eba20-7d2e-48f0-86e2-299b08fb7ab0.png)
![image](https://user-images.githubusercontent.com/25851359/156878149-09857385-4271-4134-b912-f8ca43a2bf72.png)

Anyway, @julxploitrr could you show your file .mp3 specification? 
And could you check is does Audacity 'see' ffmpeg libs?

julxploitrr:
Hey, I'm very busy this week and can't do the following. This is a large bug a lot of other people are talking about, check audacity forums for more information about this bug as I can't, thank you.

LWinterberg:
What I've found so far that this happens with malformed files coming off some YouTube Downloader services.

junaayd123:
audacity can’t recognize the file because it’s partially corrupt. Did it play when you double clicked the file?

petersampsonaudacity:
Testing on W11 with 3.1.3 and latest 3.2.0 alpha master i _**can**_ reproduce this

Regarding the problemmatic MP3 imports, I have two MP3 files here on my Zurich laptop that were reported as problemmatic the last time I was over here in Switzerland.

Neither of them can be opened/imported in Audacity - you get an error message:
![image](https://user-images.githubusercontent.com/50205984/157266152-1dcdec57-68e1-4a83-8417-150bc3b677f6.png)


But:
1. Reaper loads them and plays them
[...]

Issue #2236, opened 2021-12-05
ImportMP3: migrate from 'libmad' to 'libmpg123'


carlo-bramini:
Hello,
I was wondering if it could be evaluated the chance to migrate from 'libmad' to 'libmpg123', when using the feature of importing MP2/MP3/MPA files. By reading several tests made available on the net, 'libmad' seems to be an outdated piece of software, less accurate and much slower than 'libmpg123', which provides also decent support to modern build tools and optimized code for various CPUs.
At the time of writing, the entire support of this feature is self-contained into `src\import\ImportMP3.cpp`.
Could it be accepted this enhancement (in my opinion, it is)?
Thank you.

LWinterberg:
* https://www.wezm.net/technical/2008/04/mp3-decoder-libraries-compared/ has mpg123 come out 2-3x faster than mad
* mad is GPLv2, mpg123 is LGPLv2.1 
* mad got its last update in 2004, mpg123 is getting ~monthly releases (though most of the changes seem to be to non-mp3 import things, so I wouldn't say it's massively unstable because of it)
* (I wonder if mpg123 many other tools would be useful for us as well)

I think I would agree that mpg123 is an enhancement

LWinterberg:
bumping priority - "Audacity unable to open mp3" is the # 1 request we get on the forum, often accompanied with a "but it works fine in VLC". [VLC uses mpg123](https://code.videolan.org/videolan/vlc/-/blob/master/modules/codec/mpg123.c) since 3.0, released 2018, so I have very strong hopes that switching will help us, too.

petersampsonaudacity:
See:   Audacity doesn't recognize a mp3 file even though it's a mp3 file. #2634

PR #3079, opened 2022-06-15
MP3 import based on libmpg123

crsib:
Resolves: #2236 
Resolves: #2634 
Resolves: #3029

Audacity uses `libmad` to import MP3 files. It seems that this library struggles to play some allegedly valid mp3 files.
Similar behavior is observed when using `madplay`.

This PR introduces an alternative MP3 importer, based on `libmpg123`, which is known to be used by major software, such as VLC.

Recommended:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -706,7 +706,7 @@ list( APPEND SOURCES
       >
 
       $<$<BOOL:${USE_LIBMAD}>:
-         import/ImportMP3.cpp
+         import/ImportMP3_MAD.cpp
       >
 
       $<$<AND:$<BOOL:${USE_LIBOGG}>,$<BOOL:${USE_LIBVORBIS}>>:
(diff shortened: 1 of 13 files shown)


ksoze95 (approved):

src/CMakeLists.txt
```diff
@@ -706,7 +706,7 @@ list( APPEND SOURCES
       >
 
       $<$<BOOL:${USE_LIBMAD}>:
-         import/ImportMP3.cpp
+         import/ImportMP3_MAD.cpp
```
Paul-Licameli, line 709: The old file name is still mentioned in comments in itself and in Import.cpp.

crsib: I will check!

crsib: Anyway, I have restructured commits to improve it

PR #3081, opened 2022-06-15
Lv2 smart pointers

Paul-Licameli:
Resolves: #2977

Simplify lv2 code with more smart pointers; fix at least one leak and one dangling pointer

Recommended:


crsib (approved):
Carefully going through the commits I do not observe any undesired side effects.

Tremendous work!

PR #3086, opened 2022-06-16
Use std::uncaught_exceptions where possible

crsib:
`std::uncaught_exception` was deprecated in C++17, however it is only available on macOS 10.12 or later.

Recommended:

### DIFF
--- a/libraries/lib-exceptions/AudacityException.h
+++ b/libraries/lib-exceptions/AudacityException.h
@@ -173,10 +173,10 @@ inline SimpleGuard< void > MakeSimpleGuard() noexcept { return {}; }
 
 /*!
   Executes a given function (typically a lamba), in any thread.
- 
+
   If there is any exception, can invoke another given function as handler, which may rethrow that or
   another exception, but usually just returns the value for the GuardedCall.
- 
+
   If AudacityException is handled, then it queues up a delayed handler action for execution later in
   the event loop at idle time, on the main thread; typically this informs the user of the error.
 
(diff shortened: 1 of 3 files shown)


libraries/lib-exceptions/AudacityException.h
```diff
@@ -226,7 +228,7 @@ noexcept(
          // At this point, e is the "current" exception, but not "uncaught"
          // unless it was rethrown by handler.  handler might also throw some
          // other exception object.
-         if (!std::uncaught_exception()) {
+         if (!HasUncaughtExceptions()) {
```
Paul-Licameli, line 231: If `uncaught_exceptions` is available, then a more correct test would be this:

- save its value before the `finally`
- test here that its current value is not more than the saved

Paul-Licameli (changes requested):

libraries/lib-exceptions/AudacityException.h
crsib: Fixed and force-pushed

libraries/lib-exceptions/AudacityException.h
```diff
@@ -218,15 +218,21 @@ noexcept(
...
       noexcept(noexcept(
          std::function<void(AudacityException*)>{
             std::move(delayedHandler)} )) {
          // At this point, e is the "current" exception, but not "uncaught"
          // unless it was rethrown by handler.  handler might also throw some
          // other exception object.
+      #ifdef UNCAUGHT_EXCEPTIONS_UNAVAILABLE
          if (!std::uncaught_exception()) {
+      #else         
+         if (uncaughtExceptionsCount >= std::uncaught_exceptions()) {
```
Paul-Licameli, line 234: I think `==` is a sufficient test -- `uncaught_exceptions()` might have become more, but I don't know how it could become less!

But this is not incorrect so let it stand.

Paul-Licameli (approved):

libraries/lib-exceptions/AudacityException.h
crsib: Yeah, I was thinking to do `==`, but `>=` is even broader condition

PR #3087, opened 2022-06-16
Lv2 features list

Paul-Licameli:
Factor a base class from LV2Effect for options and features...

... This may become a shared base class with the LV2 Instance class
when that exists.

This begins to break up the too-big LV2Effect.cpp file and legalize it.

Recommended:


crsib (approved):
Looks good

PR #831, opened 2021-05-03
Use XDG dirs by default on Linux

serebit:
# Notice

The contents of this PR are now licensed under Creative Commons Zero v1.0 Universal, and have been released into the public domain. You may adapt this patch as you wish for any purpose.

# Original Text

This PR is an updated version of #632, which required a rebase against v3.0.x-based Audacity master.

This addresses [issue 2201](https://bugzilla.audacityteam.org/show_bug.cgi?id=2201) on the Bugzilla tracker.

The new behavior is only applicable to Linux. If the `~/.audacity-data` folder doesn't exist (i.e. on a new installation of Audacity) the proper XDG-compliant directories will be used for saving data (configuration files go in `~/.config/audacity`, data files go in `~/.local/share/audacity`). On existing installations of Audacity where `~/.audacity-data` already exists, that directory will continue to be used. This allows for potential downgrade compatibility should Audacity drop the legacy behavior in the future, and doesn't disrupt existing installations.

PR #632, opened 2020-07-27
Use XDG-compliant dirs by default on Unix platforms

serebit:
This addresses [issue 2201](https://bugzilla.audacityteam.org/show_bug.cgi?id=2201) on the Bugzilla tracker.

The new behavior is only applicable to Unix platforms. If the `~/.audacity-data` folder doesn't exist (i.e. on a new installation of Audacity) the proper XDG-compliant directories will be used for saving data (configuration files go in `~/.config/audacity`, data files go in `~/.local/share/audacity`). On existing installations of Audacity where `~/.audacity-data` already exists, that directory will continue to be used. This allows for potential downgrade compatibility should Audacity drop the legacy behavior in the future, and doesn't disrupt existing installations.

PR #3088, opened 2022-06-16
Use XDG dirs by default on Linux

### LINKED ISSUE #453: Data folder placement: use XDG dirs on Linux
On Linux the Audacity data folder is on:
~/.audacity-data

Most likely it should be:
~/.config/audacity

RuRo:
This PR is an updated version of #831 which itself is an updated version of #632

Fixes #453 and resolves: https://bugzilla.audacityteam.org/show_bug.cgi?id=2201

Recommended:

### DIFF
--- a/libraries/lib-files/FileNames.cpp
+++ b/libraries/lib-files/FileNames.cpp
@@ -30,6 +30,7 @@ used throughout Audacity into this one place.
 #include <wx/filename.h>
 #include <wx/intl.h>
 #include <wx/stdpaths.h>
+#include <wx/utils.h>
 #include "BasicUI.h"
 #include "Prefs.h"
 #include "Internat.h"
@@ -44,8 +45,6 @@ used throughout Audacity into this one place.
 #if defined(__WXMSW__)
 #include <windows.h>
 #endif
-
-static wxString gDataDir;
 
 const FileNames::FileType
      FileNames::AllFiles{ XO("All files"), { wxT("") } }
@@ -224,13 +223,67 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
    return dir;
 }
 
-FilePath FileNames::DataDir()
+namespace // Implementation for user directories
 {
-   // LLL:  Wouldn't you know that as of WX 2.6.2, there is a conflict
-   //       between wxStandardPaths and wxConfig under Linux.  The latter
-   //       creates a normal file as "$HOME/.audacity", while the former
-   //       expects the ".audacity" portion to be a directory.
-   if (gDataDir.empty())
+enum class DirTarget
+{
+   Cache,
+   Config,
+   Data,
+   State,
+   _targetCount,
+};
+
+static FilePath gTargetDirs[size_t(DirTarget::_targetCount)] = {};
+
+#if defined(__WXGTK__)
+struct XDGDirConfig final
+{
+   wxString dirEnvVar;
+   FilePath dirDefault;
+};
+
+const XDGDirConfig gXDGUnixDirs[] = {
+   {wxT("XDG_CACHE_HOME"),  wxT("/.cache")      },
+   {wxT("XDG_CONFIG_HOME"), wxT("/.config")     },
+   {wxT("XDG_DATA_HOME"),   wxT("/.local/share")},
+   {wxT("XDG_STATE_HOME"),  wxT("/.local/state")}
+};
+
+static_assert(
+   sizeof(gXDGUnixDirs) / sizeof(*gXDGUnixDirs) == size_t(DirTarget::_targetCount),
+   "Not all DirTarget cases were implemented!"
+);
+
+FilePath GetXDGTargetDir(DirTarget target)
+{
+   static const auto oldUnixDataDir = wxFileName::GetHomeDir() + wxT("/.audacity-data");
+   static const auto oldUnixDataDirExists = wxDirExists(oldUnixDataDir);
+   // Compatibility: Use old user data dir folder, if it already exists
+   if (oldUnixDataDirExists)
+      return oldUnixDataDir;
+
+   // see if the XDG_*_HOME env var is defined. if it is, use its value.
+   // if it isn't, use the default XDG-specified value.
+   wxString newDir;
+   const auto [dirEnvVar, dirDefault] = gXDGUnixDirs[size_t(target)];
+   if (!wxGetEnv(dirEnvVar, &newDir) || newDir.empty())
+      newDir = wxFileName::GetHomeDir() + dirDefault;
+
+#ifdef AUDACITY_NAME
+   newDir = newDir + wxT("/" AUDACITY_NAME);
+#else
+   newDir = newDir + wxT("/audacity");
+#endif
+
+   return newDir;
+}
+#endif
+
+FilePath GetUserTargetDir(DirTarget target)
+{
+   auto& dir = gTargetDirs[size_t(target)];
+   if (dir.empty())
    {
       // If there is a directory "Portable Settings" relative to the
       // executable's EXE file, the prefs are stored in there, otherwise
@@ -248,19 +301,27 @@ FilePath FileNames::DataDir()
       if (::wxDirExists(portablePrefsPath.GetFullPath()))
       {
          // Use "Portable Settings" folder
-         gDataDir = portablePrefsPath.GetFullPath();
+         dir = portablePrefsPath.GetFullPath();
       } else
       {
+#if defined(__WXGTK__)
+         // Use XDG Base Directory compliant folders
+         wxString newDir(GetXDGTargetDir(target));
+#else
          // Use OS-provided user data dir folder
-         wxString dataDir( LowerCaseAppNameInPath( wxStandardPaths::Get().GetUserDataDir() ));
-#if defined( __WXGTK__ )
-         dataDir = dataDir + wxT("-data");
+         wxString newDir( FileNames::LowerCaseAppNameInPath( wxStandardPaths::Get().GetUserDataDir() ));
 #endif
-         gDataDir = FileNames::MkDir(dataDir);
+         dir = FileNames::MkDir(newDir);
       }
    }
-   return gDataDir;
+   return dir;
 }
+} // End of implementation for user directories
+
+FilePath FileNames::CacheDir() { return GetUserTargetDir(DirTarget::Cache); }
+FilePath FileNames::ConfigDir() { return GetUserTargetDir(DirTarget::Config); }
+FilePath FileNames::DataDir() { return GetUserTargetDir(DirTarget::Data); }
+FilePath FileNames::StateDir() { return GetUserTargetDir(DirTarget::State); }
 
 FilePath FileNames::ResourcesDir(){
    wxString resourcesDir( LowerCaseAppNameInPath( wxStandardPaths::Get().GetResourcesDir() ));
@@ -315,17 +376,17 @@ FilePath FileNames::PlugInDir()
 
 FilePath FileNames::Configuration()
 {
-   return wxFileName( DataDir(), wxT("audacity.cfg") ).GetFullPath();
+   return wxFileName( ConfigDir(), wxT("audacity.cfg") ).GetFullPath();
 }
 
 FilePath FileNames::PluginRegistry()
 {
-   return wxFileName( DataDir(), wxT("pluginregistry.cfg") ).GetFullPath();
+   return wxFileName( ConfigDir(), wxT("pluginregistry.cfg") ).GetFullPath();
 }
 
 FilePath FileNames::PluginSettings()
 {
-   return wxFileName( DataDir(), wxT("pluginsettings.cfg") ).GetFullPath();
+   return wxFileName( ConfigDir(), wxT("pluginsettings.cfg") ).GetFullPath();
 }
 
 FilePath FileNames::BaseDir()
--- a/libraries/lib-files/FileNames.h
+++ b/libraries/lib-files/FileNames.h
@@ -109,12 +109,28 @@ namespace FileNames
       FilePaths &otherNames, wxFileName &newName);
 
    FILES_API wxString LowerCaseAppNameInPath( const wxString & dirIn);
+
+   /** \brief Audacity user cache directory
+    *
+    * Where audacity keeps its cache squirreled away, by default ~/.cache/audacity/
+    * on Unix, Application Data/Audacity on windows system */
+   FILES_API FilePath CacheDir();
+   /** \brief Audacity user config directory
+    *
+    * Where audacity keeps its settigns squirreled away, by default ~/.config/audacity/
+    * on Unix, Application Data/Audacity on windows system */
+   FILES_API FilePath ConfigDir();
    /** \brief Audacity user data directory
     *
-    * Where audacity keeps its settings and other user data squirreled away,
-    * by default ~/.audacity-data/ on Unix, Application Data/Audacity on
-    * windows system */
+    * Where audacity keeps its user data squirreled away, by default ~/.local/share/audacity/
+    * on Unix, Application Data/Audacity on windows system */
    FILES_API FilePath DataDir();
+   /** \brief Audacity user state directory
+    *
+    * Where audacity keeps its user state squirreled away, by default ~/.local/state/audacity/
+    * on Unix, Application Data/Audacity on windows system */
+   FILES_API FilePath StateDir();
+
    FILES_API FilePath ResourcesDir();
    FILES_API FilePath HtmlHelpDir();
    FILES_API FilePath HtmlHelpIndexFile(bool quick);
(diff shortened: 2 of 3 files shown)


RuRo:
@LWinterberg @crsib is there anything left on my side, that is blocking this PR or are we just waiting for someone to review this?

LWinterberg:
I think the main blocker for this was the weekend :) 

The review is already inside our kanban board, so crsib will look at it soon

libraries/lib-files/PathList.cpp
```diff
@@ -20,6 +20,9 @@ void FileNames::InitializePathList()
    auto &standardPaths = wxStandardPaths::Get();
    const auto programPath = standardPaths.GetExecutablePath();
 
+   // encouraged by wxwidgets
+   standardPaths.SetFileLayout(wxStandardPaths::FileLayout::FileLayout_XDG);
```
crsib, line 24: Probably it should go to `#if defined(__WXGTK__)` as well

src/AboutDialog.cpp
```diff
@@ -612,7 +612,10 @@ void AboutDialog::PopulateInformationPage( ShuttleGui & S )
 #endif
 
    // Location of settings
-   AddBuildinfoRow(&informationStr, XO("Settings folder:"), \
+   AddBuildinfoRow(&informationStr, XO("Config folder:"), \
```
crsib, line 615: I think we could leave "Settings folder" here. What do you think, @Paul-Licameli?

libraries/lib-files/PathList.cpp
RuRo: Will do.

src/AboutDialog.cpp
RuRo: Okay.

libraries/lib-files/FileNames.cpp
```diff
@@ -224,8 +230,48 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
...
+         gConfigDir = portablePrefsPath.GetFullPath();
+#if defined(__WXGTK__)
+      } else if (::wxDirExists(gOldUnixDataDir))
+      {
+         // Use old user data dir folder
+         gConfigDir = gOldUnixDataDir;
+#endif
+      } else {
+         // Use OS-provided user data dir folder
+         wxString configDir(wxStandardPaths::Get().GetUserConfigDir() + wxT("/audacity"));
```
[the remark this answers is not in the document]
RuRo: I am not sure, if I follow? `GetUserDataDir()` will return `~/.local/share`, not `~/.config`. +1 on using `AUDACITY_NAME` instead of a hardcoded string.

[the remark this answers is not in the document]
crsib: Docs are confusing: `Only use this method if you have a single configuration file to put in this directory, otherwise [GetUserDataDir()](https://docs.wxwidgets.org/3.1/classwx_standard_paths.html#a5b9d1b1addc3e4ce30ccb3817cbbe19b) is more appropriate as the latter adds appinfo to the path, unlike this function.` https://docs.wxwidgets.org/3.1/classwx_standard_paths.html#a0561d70f7da648e10b25bf9679309da3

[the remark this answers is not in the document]
RuRo: I agree, that the docs are confusing. I am currently rewriting the implementation based on your `enum DirTarget` idea.

Actually the `wxwidgets` `GetUserWhateverDir` methods don't map to corresponding `XDG_WHATEVER_HOME` directories, so I decided to just implement the correct XDG Base Directory logic for unix and keep the windows case exactly as it currently is.

More info later. Btw, are we sure that `__WXGTK__` is the correct preprocessor macro to distinguish unix? I just kept it from the original PR, but I am not 100% sure.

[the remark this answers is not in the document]
crsib: > are we sure that __WXGTK__ 

We use it for now, at least, so it's good to go!

RuRo:
@crsib PTAL. I've implemented the 4 XDG Base Directories. Currently, all 4 directories default back to `~/.audacity-data` if it already exists and to `GetUserDataDir` on Windows.

Also, I completely removed the `FileLayout_XDG` logic since it would only affect `Config` and we would still need to implement the XDG logic manually for `Cache`, `Data` and `State`, at which point there is no reason to use `GetUserConfigDir` only for `Config`.

For now, `Cache` and `State` aren't used anywhere. If you have any suggestions regarding which other old `DataDir()` calls should be replaced with `Cache`, `Config` or `State`, feel free to point them out. I am not too familiar with the various audacity `xml` files, so I don't know, which of them should go to `Config` or `Data` or `State` etc.

RuRo:
Here are some files/directories that currently still use `DataDir()`:
- lastlog.txt
- genres.txt
- journal.txt
- journallog.txt
- EQCurves.xml
- EQDefaultCurves.xml
- EQBackup.xml
- ffmpeg_presets.xml
- Chains
- Macros
- NRP
- Plug-Ins
- Theme

I might have missed some.

RuRo:
Oops. I don't have a Windows/Mac machine, so I missed a typo. @crsib sorry, can you approve the CI/CD again.

RuRo:
@imciner2 I've refactored the code to use `glib` and `get_user_*_dir()`, but it seems that `glib` is normally not used for building `libraries/` so I had to manually add `glib` to `libraries/lib-files/CMakeLists.txt`. I am not sure, if I did that right (I just copy-pasted that line from `src/CMakeLists.txt`).

libraries/lib-files/FileNames.cpp
```diff
@@ -26,6 +26,7 @@ used throughout Audacity into this one place.
 
 #include <memory>
 
+#include <glib.h>
```
crsib, line 29: Looks like glib is included unconditionally.

libraries/lib-files/FileNames.cpp
```diff
@@ -224,13 +232,48 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
...
+FilePath &gTargetDir(DirTarget target)
+{
+   switch (target)
+   {
+      case DirTarget::Cache: return gCacheDir;
+      case DirTarget::Config: return gConfigDir;
+      case DirTarget::Data: return gDataDir;
+      case DirTarget::State: return gStateDir;
+   }
+   wxASSERT(!"Not all DirTarget cases were implemented.");
```
crsib, line 252: Let's use c-assert. We are moving away from wxASSERT

libraries/lib-files/FileNames.cpp
```diff
@@ -224,13 +232,48 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
...
+{
+   switch (target)
+   {
+      case DirTarget::Cache: return g_get_user_cache_dir();
+      case DirTarget::Config: return g_get_user_config_dir();
+      case DirTarget::Data: return g_get_user_data_dir();
+      case DirTarget::State: return g_get_user_state_dir();
+   }
+   wxASSERT(!"Not all DirTarget cases were implemented.");
+   abort();
```
crsib, line 266: Ditto

libraries/lib-files/FileNames.cpp
```diff
@@ -224,13 +232,48 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
...
+}
+
+FilePath userTargetDir(DirTarget target)
 {
+   auto dir = gTargetDir(target);
```
crsib, line 271: You return non const reference from gTargetDir, but here C++ deduce the type as `FilePath`. Is is desired?

libraries/lib-files/FileNames.cpp
```diff
@@ -245,23 +288,45 @@ FilePath FileNames::DataDir()
...
       } else
       {
          // Use OS-provided user data dir folder
-         wxString dataDir( LowerCaseAppNameInPath( wxStandardPaths::Get().GetUserDataDir() ));
-#if defined( __WXGTK__ )
-         dataDir = dataDir + wxT("-data");
+         wxString newDir( FileNames::LowerCaseAppNameInPath( wxStandardPaths::Get().GetUserDataDir() ));
 #endif
-         gDataDir = FileNames::MkDir(dataDir);
+         dir = FileNames::MkDir(newDir);
```
crsib, line 319: Looks like you want overwrite the value, so you need `auto& dir = gTargetDir(target);`

crsib (changes requested):
Glib is included and used unconditionally which is definitely a blocker. Some other issues as well.

crsib:
Linux build has failed as well!

RuRo:
> Linux build has failed as well!

Yeah. It seems that CI uses an older GLib version, that doesn't have `g_get_user_state_dir`. It seems to be only available since 2.71.1. What do you think we should do:

1) Update required GLib version
2) Remove `FileNames::StateDir` and `State` from `DirTarget`
3) Don't use GLib at all and revert to manually checking the environment variables

crsib:
We will not upgrade GLib version on CI any time soon, as it will break AppImage for users on older GNU/Linux versions.

Otherwise it's your call

RuRo:
@imciner2 requested to use GLib
@crsib requested to add `State`

Please decide, which request should be reverted. :sweat_smile:
I would generally prefer to just remove the GLib dependency and revert to manual env lookup, but this is a weak preference.

crsib:
>I would generally prefer to just remove the GLib dependency and revert to manual env lookup,

Me too, to be fair.

imciner2:
If you really want to use the state directory, then more work would be needed probably. The XDG state directory was only added to the spec last year, so there may be issues with the base directories not existing on older systems, so you need to ensure the entire directory path exists, not just the Audacity subdirectory.

> We will not upgrade GLib version on CI any time soon, as it will break AppImage for users on older GNU/Linux versions.

Doesn't appimage bundle its dependencies into the image?

RuRo:
> ... there may be issues with the base directories not existing on older systems, so you need to ensure the entire directory path exists, not just the Audacity subdirectory. ...

Currently, the code calls `FileNames::MkDir` that uses `wxFileName::Mkdir` with `wxPATH_MKDIR_FULL`
```
If the flags contain wxPATH_MKDIR_FULL flag, try to create each directory in the path and also don't return an error if the target directory already exists.
```

So that shouldn't be an issue.

RuRo:
So, @crsib @imciner2 what's the conclusion here? Is there anything else I can do to get this PR accepted?

crsib:
I'll do a hopefully final round today

libraries/lib-files/FileNames.cpp
```diff
@@ -45,7 +47,14 @@ used throughout Audacity into this one place.
 #include <windows.h>
 #endif
 
+#if defined(__WXGTK__)
+static wxString gOldUnixDataDir;
```
crsib, line 51: I think we can do something like

```
namespace
{
auto GetOldUserDataDir()
{
    static const auto dir = wxFileName::GetHomeDir() + wxT("/.audacity-data");
    return dir;
}

auto OldUserDataDirExists()
{
    static const auto exists = wxDirExists(GetOldUserDataDir());
    return exists;
}
}
```

This will simplify code below

libraries/lib-files/FileNames.cpp
```diff
@@ -45,7 +47,14 @@ used throughout Audacity into this one place.
...
+#endif
+
+static wxString gConfigDir;
```
crsib, line 54: I think we can have an array here by adding `Count` to  `DirTarget`:

```
enum class DirTarget
{
    Cache,
    Config,
    Data,
    State,

     Count
};

wxString directories[size_t(DirTarget::Count)];
```

crsib: (Sorry for the formatting, GitHub aint a good text editor xD)

libraries/lib-files/FileNames.cpp
```diff
@@ -224,13 +233,69 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
...
-FilePath FileNames::DataDir()
+enum class DirTarget
 {
+    Cache,
+    Config,
+    Data,
+    State
+};
+
+FilePath &gTargetDir(DirTarget target)
```
crsib, line 244: Can be done as
```
struct XDGDirConfig final
{
   wxString envName;
   wxString defaultDir;
};

XDGDirConfig xdgConfig[] = {
  { L"XDG_CACHE_HOME", L"/.cache" },
...
};

static_assert(sizeof(xdgConfig) / sizeof(xdgConfig[0] == size_t(DirTarget::Count)

wxString GetXDGDir(DirTarget target)
{
   const auto& dirConfig = xdgConfig[size_t(target)];
 
   wxString dir;
   // see if the XDG_*_HOME env var is defined. if it is, use its value.
   // if it isn't, use the default XDG-specified value.
   if ( !wxGetEnv(dirConfig.envName, &dir) || dir.empty() )
        dir = wxFileName::GetHomeDir() +  dirConfig.defaultDir;

  return dir;
}
```

This way we force the compiler to do all the checks necessary. Given that the enum is private, there is not much need left for any runtime checks.

libraries/lib-files/FileNames.cpp
```diff
@@ -224,13 +233,69 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
...
+      case DirTarget::State: return wxT("/.local/state");
+   }
+   throw SimpleMessageBoxException{
+      ExceptionType::Internal,
+      XO("Not all DirTarget cases were implemented."),
+   };
+}
+#endif
+
+FilePath userTargetDir(DirTarget target)
```
crsib, line 291: Let's name it `GetUserTargetDir` for consistency please!

libraries/lib-files/FileNames.cpp
```diff
@@ -224,13 +233,69 @@ wxString FileNames::LowerCaseAppNameInPath( const wxString & dirIn){
...
+{
+   auto& dir = gTargetDir(target);
```
crsib, line 293: This becomes `auto& dir = directories[size_t(target)];`

libraries/lib-files/FileNames.cpp
```diff
@@ -245,23 +310,50 @@ FilePath FileNames::DataDir()
 #endif
       wxFileName portablePrefsPath(exePath.GetPath(), wxT("Portable Settings"));
 
+#if defined(__WXGTK__)
+      if (gOldUnixDataDir.empty())
```
crsib, line 314: This goes away

libraries/lib-files/FileNames.cpp
```diff
@@ -245,23 +310,50 @@ FilePath FileNames::DataDir()
...
+         gOldUnixDataDir = wxFileName::GetHomeDir() + wxT("/.audacity-data");
+#endif
+
       if (::wxDirExists(portablePrefsPath.GetFullPath()))
       {
          // Use "Portable Settings" folder
-         gDataDir = portablePrefsPath.GetFullPath();
+         dir = portablePrefsPath.GetFullPath();
+#if defined(__WXGTK__)
+      } else if (::wxDirExists(gOldUnixDataDir))
```
crsib, line 323: `if (OldUserDataDirExists())...`

libraries/lib-files/FileNames.h
```diff
@@ -109,12 +109,28 @@ namespace FileNames
       FilePaths &otherNames, wxFileName &newName);
 
    FILES_API wxString LowerCaseAppNameInPath( const wxString & dirIn);
+
+   /** \brief Audacity user cache directory
+    *
+    * Where audacity keeps its cache squirreled away, by default ~/.cache/audacity/
+    * on Unix, Application Data/Audacity on windows system */
+   FILES_API FilePath CacheDir();
```
crsib, line 117: Let's make all 4 functions to be const!

crsib (changes requested):
A small suggested rewrite :-)

libraries/lib-files/FileNames.cpp
RuRo: Actually, I think I'll just extract all the XDG-specific logic into a separate function and then just inline these 2 functions into it.

libraries/lib-files/FileNames.cpp
RuRo: Makes sense. Would `_targetCount` be an acceptable name? I would like to somehow distinguish that `Count` is a sentinel, not a valid Target.

libraries/lib-files/FileNames.cpp
RuRo: How about
```c++
using XDGDirConfig = std::tuple<...>;
```
Also I think the `xdgConfig` array should be `const`.

crsib: std::pair would do too, but I prefer descriptive names :-)

libraries/lib-files/FileNames.h
RuRo: What do you mean? This is a namespace, not a class.

crsib: Oh shoot, ignore me than

RuRo:
@crsib can you approve the CI again, please.

crsib (approved):

Issue #2970, opened 2022-05-23
define the settings struct, to capture VST2 parameters


Paul-Licameli:
- [x] 1. (like 213ecd3) Extract a base class of the effect, holding the handles (effect component and edit controller) that are needed to iterate parameters and extract preset state (this new base will be re-used also with the Instance later)
- [x] 2. (like 908c16a) Make a higher-order function for the visiting of parameter values, hiding details of how to iterate them
- [x] 3. (like 3cce7fd7) import/export steps lowered into the base class
- [x] 4. define the structure, `FetchSettings` method, `StoreSettings` method
- [x] 5. (like part of 218c56e) Define TransferDataToWindow (to become UpdateUI of a validator later), calling `StoreSettings`; remove parts of LoadSettings that duplicate this
- [x] 6. (like rest of 218c56e) Rewrite `LoadSettings`, `LoadParameters`, `SaveSettings`, `SaveParameters` not using the handles, and insert needed calls to `FetchSettings`
- [x] 7. (like 5199894) Rewrite export and saving of presets, using `StoreSettings` on the "scratch" handles in the Effect object

PR #3089, opened 2022-06-16
Define settings vst2

pietro68:
Resolves: #2970

Just a draft to see if I am heading the right way; some commits will need to be rearranged.

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.h
+++ b/src/effects/VST/VSTEffect.h
@@ -73,6 +73,19 @@ typedef SInt16 CFBundleRefNum;
 #endif
 #endif
 
+struct VSTEffectWrapper : public VSTEffectLink
+{
+   AEffect* mAEffect = nullptr;
+
+   intptr_t callDispatcher(int opcode, int index,
+      intptr_t value, void* ptr, float opt) override;
+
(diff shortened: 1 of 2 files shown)


src/effects/VST/VSTEffect.h
```diff
@@ -73,6 +73,17 @@ typedef SInt16 CFBundleRefNum;
 #endif
 #endif
 
+struct VSTEffectSettings
+{
+   std::unordered_map<wxString, double> mParamsMap;
```
Paul-Licameli: Remember `#include <unordered_map>` in the commit that introduces this line, so the macOS build isn't broken.

src/effects/VST/VSTEffect.h
```diff
@@ -73,6 +73,11 @@ typedef SInt16 CFBundleRefNum;
 #endif
 #endif
 
+struct VSTEffectWrapper
+{
+   AEffect* mAEffect = NULL;
```
Paul-Licameli, line 101: nullptr is more modern

src/effects/VST/VSTEffect.h
pietro68: ok, done

Issue #2978, opened 2022-05-25
EffectSettings for lv2


Paul-Licameli:
Define and use a structure describing externalized state of an LV2 effect, although the effect will just contain an instance of it and not really be stateless yet.

Paul-Licameli:
Only partially resolved by the first PR

PR #3096, opened 2022-06-17
Lv2 ports

Paul-Licameli:
Partly Resolves: #2978

Separate immutable descriptions of the various kinds of LV2 ports
from structures that hold associated state.

Just one of these is the EffectSettings, and as with Ladspa, it is just a
vector<float>.  Except -- will also transiently carries Lilv_state after loading a preset.

The other state structures will later belong to instance or validator objects.

Also some source file splitting for levelization.

The misleadingly named "LV2EffectSettingsDialog" is also renamed and moved into a separate file. 

Recommended:


src/effects/lv2/LV2Effect.cpp
```diff
@@ -1395,6 +1395,8 @@ std::unique_ptr<LV2Wrapper> LV2Effect::InitInstance(float sampleRate)
    wrapper->SetBlockSize();
    wrapper->SetSampleRate();
 
+   static float blackHole;
```
crsib, line 1284: Well, that's a name

src/effects/lv2/LV2Effect.cpp
```diff
@@ -99,12 +102,13 @@ BEGIN_EVENT_TABLE(LV2EffectMeter, wxWindow)
...
 END_EVENT_TABLE()
 
-LV2EffectMeter::LV2EffectMeter(wxWindow *parent, const LV2ControlPortPtr port)
-:  wxWindow(parent, wxID_ANY, wxDefaultPosition, wxDefaultSize, wxDEFAULT_CONTROL_BORDER),
-   mControlPort(port)
+LV2EffectMeter::LV2EffectMeter(
+   wxWindow *parent, const LV2ControlPortPtr port, const float &value
+)  : wxWindow(parent, wxID_ANY, wxDefaultPosition, wxDefaultSize,
+      wxDEFAULT_CONTROL_BORDER), mControlPort(port)
+   , mValue{ value }
```
crsib: We can drop `const float& value` I think and just initialize `mValue{ port->mVal }` probably.

I've spent some time to convince myself that the lifetime of the pointers (references) match.  But I haven't seen next commits, probably they justify the need for the third argument.

crsib: Yep, they do justify

crsib (approved):
I do enjoy the direction LV2 is going

src/effects/lv2/LV2Effect.cpp
```diff
@@ -1755,7 +1755,7 @@ bool LV2Effect::BuildPlain(EffectSettingsAccess &access)
          for (auto & p : mGroupMap[label]) { auto & port = mControlPorts[p];
             auto labelText = port->mName;
             if (!port->mUnits.empty())
-               labelText += wxT(" (") + port->mUnits + wxT(")");
+               labelText += XO("(%s)").Format(port->mUnits).Translation();
```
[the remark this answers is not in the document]
Paul-Licameli: The string XO(“%s”) occurs elsewhere with i18n comments.

Even punctuation may need to localize.  For instance the Chinese locales prefer slightly different parenthesis characters that are more rounded.

src/effects/lv2/LV2Effect.cpp
Paul-Licameli: Haha

Paul-Licameli: I could have called it `bitBucket`

src/effects/lv2/LV2Effect.cpp
Paul-Licameli: But, as you must understand by now, this leads up to removal of `mVal` from the port structure so it too is an immutable

src/effects/lv2/LV2Effect.h
```diff
@@ -227,6 +227,15 @@ class LV2ControlPort : public LV2Port
    std::vector<double> mScaleValues;
    wxArrayString mScaleLabels;
 
+   //! Map a real number to one of the scale points
+   size_t Discretize(float value) const {
```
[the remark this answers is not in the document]
Paul-Licameli: I did that in the final version

Issue #2727, opened 2022-03-25
attempting to locate FFMPEG is broken


LWinterberg:
STR:

1. Go to Preferences > Libraries
2. click the locate FFMPEG button
3. when asked whether you want to override the automatic detection, click yes
4. find the avformat DLL

![image](https://user-images.githubusercontent.com/87814144/160080514-4e58f6a3-49e8-465f-8e4e-ee2cb1f1a7bc.png)

OS: Windows 11
Regression on 3.1.3
Using 6c64dc455 - I have not attempted further bisection

abcasada:
Workaround - install v3.0.5, then find the DLL, then reinstall latest version (3.1.3 as of now), and it should still have the FFmpeg library.

Other old versions may work also. Verified on Windows 10.

abcasada:
Update to previous workaround - a few relaunches later, v3.1.3 failed to load FFmpeg. Going back to v3.0.5, FFmpeg is found again. Hopefully v3.0.5 won't lose it.

patheticcockroach:
For me, ffmpeg doesn't work in 3.0.5 either. I'm stuck with version 2.4.2 for the moment. In 3.0.5 and 3.1.3, when I locate the DLL manually, Audacity just seems to ignore it.
Tried with both avformat-55.dll and avformat-59.dll, on Windows 10.

crsib:
This will be fixed in 3.2

crsib:
@Ryan-Miller-Audacity I think that you've verified this already. It should be fully fixed now, all the code required is in master

dozzzzer:
Verified on Windows 10 x64

Ryan-Miller-Audacity:
> @Ryan-Miller-Audacity I think that you've verified this already. It should be fully fixed now, all the code required is in master

I did check this with #3097 and double checked this again with the most recent master branch and verified it's working as expected.

PR #3097, opened 2022-06-17
Allows to nest SettingScope

crsib:
Fixes part of: #2727

`FindFFmpegLibs` created nested setting transactions when invoked from the Preferences dialog.  This case seems to be a perfect example where nesting SettingTransaction (and SettingScope) is needed.

This PR allows to nest SettingScope.

Recommended:

### DIFF
--- a/libraries/lib-preferences/Prefs.cpp
+++ b/libraries/lib-preferences/Prefs.cpp
@@ -233,49 +233,76 @@ void FinishPreferences()
    }
 }
 
-SettingScope *SettingScope::sCurrent = nullptr;
+namespace
+{
+std::vector<SettingScope*> sScopes;
+}
 
 SettingScope::SettingScope()
 {
-   if ( sCurrent )
-      // nesting of transactions is not supported
-      wxASSERT( false );
-   else
-      sCurrent = this;
+   sScopes.push_back(this);
 }
 
 SettingScope::~SettingScope() noexcept
 {
-   if ( sCurrent == this ) {
-      if ( !mCommitted )
-         for ( auto pSetting : mPending )
-            pSetting->Rollback();
-      sCurrent = nullptr;
-   }
+   // Settings can be scoped only on stack
+   // so it should be safe to assume that sScopes.top() == this;
+   assert(!sScopes.empty() && sScopes.back() == this);
+
+   if (sScopes.empty() || sScopes.back() != this)
+      return;
+
+   if (!mCommitted)
+      for (auto pSetting : mPending)
+         pSetting->Rollback();
+
+   sScopes.pop_back();
 }
 
 // static
 auto SettingScope::Add( TransactionalSettingBase &setting ) -> AddResult
 {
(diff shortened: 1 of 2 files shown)


libraries/lib-preferences/Prefs.cpp
```diff
@@ -233,49 +233,76 @@ void FinishPreferences()
    }
 }
 
-SettingScope *SettingScope::sCurrent = nullptr;
+namespace
+{
+thread_local std::vector<SettingScope*> sScopes;
```
Paul-Licameli, line 238: Why bother with `thread_local`?

We only need this now for the main thread, and if we ever really wanted to make it work correctly with multiple threads, there would be more work to make it behave right, like checking for the disjointness of the sets of settings objects updated by concurrent scopes, making commit fail otherwise for the loser of the race, writing uses of the transaction scope to detect failure and retry, etc.

Paul-Licameli: There isn't even any protection against races to update a single Settings object even without this "transaction" system.

So really then, there should be a comment in the header that transations (and in fact all writes to settings) are intended for main thread only.  UB if you try otherwise.

crsib: Let's drop it, no problems

Paul-Licameli:
Otherwise this looks good and I approve informally.

But I want to be sure @ksoze95 has also read this and understood.

libraries/lib-preferences/Prefs.h
```diff
@@ -271,19 +277,44 @@ class Setting : public CachingSettingBase< T >
...
+      // This can be only called from within the transaction
+      assert(!this->mPreviousValues.empty());
+
+      if (this->mPreviousValues.empty())
+         return false;
+      
+      const auto result = this->mPreviousValues.size() > 1 || DoWrite();
+      mPreviousValues.pop_back();
+
+      return result;
```
ksoze95, line 305: Currently this method will always return true.

ksoze95 (changes requested):

libraries/lib-preferences/Prefs.h
```diff
@@ -271,19 +277,44 @@ class Setting : public CachingSettingBase< T >
...
+
+   bool Commit() override
+   {
+      // This can be only called from within the transaction
+      assert(!this->mPreviousValues.empty());
+
+      if (this->mPreviousValues.empty())
+         return false;
+      
+      const auto result = this->mPreviousValues.size() > 1 || DoWrite();
```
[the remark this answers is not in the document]
crsib: Please not that there is a **strict** inequality here.

If there is only one item - we write. If there is more then one - we do not write.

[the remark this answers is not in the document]
ksoze95: Opps, I mistook the statement to have read size() > 0.   Apologies.

ksoze95 (approved):

PR #3109, opened 2022-06-19
Pass encoding for pipe open

VasiliiGeneralov:
Allow users to pass encoding for pipes to work with non-ASCII data, e.g.
Cyrillic fonts

Signed-off-by: Vasilii Generalov <begulme@gmail.com>

Recommended:


crsib (approved):
The default behavior seems to be kept, so this PR is good to go

PR #3115, opened 2022-06-20
[PORTAUDIO] Fix error when USE_PORTMIXER is zero 

carlo-bramini:
I configured with `audacity_use_portmixer=Off` and I got many errors because `PaDeviceInfo` is undefined.
This happens because `portaudio.h` is not included by anyone if `USE_PORTMIXER` is undefined or zero.
This patch fixes this bug, by including `portaudio.h` before the test on portmixer.

Recommended:


libraries/lib-audio-devices/AudioIOBase.cpp
```diff
@@ -18,6 +18,8 @@ Paul Licameli split from AudioIO.cpp
 #include "Meter.h"
 #include "Prefs.h"
 
+#include "portaudio.h"
```
crsib, line 21: Can we use angle brackets please? `portaudio` is external library (unlike portmixer, which will be kept as a part of Audacity sources)

carlo-bramini: Actually, I included it without angle brackets because, into all other source files of Audacity, `portaudio.h` is included in that way, as you can see after a quick search. Like here, for example:

https://github.com/audacity/audacity/blob/6cfef1edc582d57f4b3b796cbafc68db50d2ab7b/src/AudioIO.cpp#L88

So, when I wrote the patch, I didn't care to much and I made the same thing as it has been made elsewhere.
Perhaps, there is a reason for that...

crsib (approved):

PR #3120, opened 2022-06-21
[CORE] Fix build error: images/Cursors.h not found

carlo-bramini:
I got this error when compiling Audacity:

```
src/tracks/labeltrack/ui/LabelGlyphHandle.cpp:24:10: fatal error: ../../../images/Cursors.h: No such file or directory
   24 | #include "../../../images/Cursors.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Actually, the "../../.." points to "src" directory but the "image" directory is one step higher, as you can see at the time of writing.
For fixing the build, I had to add an extra ".." for reaching the right level.

Recommended:


carlo-bramini:
> I'm really curios why such errors only appear under Cygwin. Each commit to master produces 12 builds using different compilers and OSes without any issue.

Probably, there is the chance that it does not depend on CYGWIN.
May it depend by the configuration options to CMake?
I turned OFF all options that I could from the ones that are listed here:

https://github.com/audacity/audacity/blob/master/CMAKE_OPTIONS.md

Perhaps, it may be possible that compiling a bare-metal executable of Audacity could help to find some issues that are not directly visible when building those 12 executables. For example, I got some troubles when I removed portmixer and now fixed by some PR... at the moment, I'm still getting other errors elsewhere after other packages have been disabled but unfortunately I have still not found the reasons for them.

PR #3122, opened 2022-06-21
Lv2 port states

Paul-Licameli:
Next steps in refactoring lv2.

Separate other state, besides the EffectSettings, into other structures.
Gather initialization details into constructors.  Move all that into
LV2Ports.cpp.

Later the structures will need to belong to the Instance and Validator
objects, moving out of the stateless Effect object.

Recommended:


crsib (approved):

Paul-Licameli:
I fixed the build errors, and rebased.

Issue #2493, opened 2022-01-28
3.2.0 alpha mac universal build opens in German on my Mac


petersampsonaudacity:
**Describe the bug**
3.2.0 alpha mac universal build opens in German on my Mac

**To Reproduce**
1. Install mac universal  audacity-macOS-3.2.0-alpha-20220127+4a69e09-universal 
2. clear the audacity settings folder to OOTB factory settings
3. launch 3.1.3
4. Observe: Audacity is in English
5. Quit Audacity
6. launch 3.2.0  -  audacity-macOS-3.2.0-alpha-20220127+4a69e09-universal 
7. Observe: Audacity is in German
8. Audacity > Einstellungen > Oberfläche
9. Observe:  Sprache: states "System"

**Expected behavior**
Audacity to open in may native locale for the Mac which is English

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: macOS 12.1 Monterey
 - Version:  audacity-macOS-3.2.0-alpha-20220127+4a69e09-universal 

**Additional context**
_I am currently trying to learn German with Duolingo - but the technical terms in Audacity auf Deutsch are a bit of a fast-track for me_  ;-)

crsib:
Is there a chance that you have per-app setting in macOS somehow (System Preferences -> Language & Regions -> Apps)?

crsib:
(And that setting, well, sticks. I removed it weeks ago and Russian is still returned as a system language)

petersampsonaudacity:
> Is there a chance that you have per-app setting in macOS somehow (System Preferences -> Language & Regions -> Apps)?

I don't think so - because if I run 3.1.3 (or earlier) after the German 3.2.0 3.1.3 spins up properly in English.

petersampsonaudacity:
> Is there a chance that you have per-app setting in macOS somehow (System Preferences -> Language & Regions -> Apps)?

Under Language & Regions -> General it shows Region: United Kingdom
and Preferred languages:  English - English (UK)- Primary

Language & Regions -> Apps  -  I have no language settings, all fields are blank

Penikov:
These are Language and Region preferences on my macOS 12.01, but the universal build demonstrates Russian localization instead of English

<img width="972" alt="Screenshot 2022-01-28 at 21 43 21" src="https://user-images.githubusercontent.com/25851359/151596873-41bd65e4-0eda-400d-90f3-8296ef4be5b2.png">



And this is what x64 build shows
[...]

PR #3124, opened 2022-06-21
A new way to read user locale on macOS

crsib:
Resolves: #2493 

It was observed, that for some undiscovered reason macOS can different `System Default` languages based  on application alone.

This issue affects only the builds made by `audacity/audacity` CI. Comparing the builds shows, that there is no metadata difference. Three binaries are different, but it seems that the reasons is that Audacity build is not reproducible.

As a workaround the system preferences are read, which seem to provide a consistent value. Apple used a similar approach internally, but they claim that the performance is not good enough for tight loops. This is clearly not the case for Audacity

Screenshot 2022-06-22 at 17 54 07
Screenshot 2022-06-22 at 17 53 53

Recommended:

### DIFF
--- a/libraries/lib-strings/Languages.cpp
+++ b/libraries/lib-strings/Languages.cpp
@@ -74,7 +74,7 @@ static bool TranslationExists(const FilePaths &pathList, wxString code)
 }
 
 #ifdef __WXMAC__
-#include <CoreFoundation/CFLocale.h>
+#include <CoreFoundation/CFPreferences.h>
 #include <wx/osx/core/cfstring.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


libraries/lib-strings/Languages.cpp
```diff
@@ -74,7 +74,7 @@ static bool TranslationExists(const FilePaths &pathList, wxString code)
 }
 
 #ifdef __WXMAC__
-#include <CoreFoundation/CFLocale.h>
+#include <CoreFoundation/CFPreferences.h>
```
Paul-Licameli, line 77: In the commit comment, were you beginning to write some more explanation?

scripts/build/macOS/build_universal.sh
```diff
@@ -0,0 +1,38 @@
...
+buildDirectory=$1
+buildType=$2
+
+scriptLocation=$(dirname "$(readlink -f "$0")")
+sourceLocation=$(readlink -f "$scriptLocation/../../..")
+
+cmake -B "${buildDirectory}/x64" -S "${sourceLocation}" -DCMAKE_BUILD_TYPE="${buildType}" -DCMAKE_CONFIGURATION_TYPES="${buildType}" -DMACOS_ARCHITECTURE=x86_64 -G Xcode
+cmake --build "${buildDirectory}/x64" --config "${buildType}"
+
+cmake -B "${buildDirectory}/arm64" -S "${sourceLocation}" -DCMAKE_BUILD_TYPE="${buildType}" -DCMAKE_CONFIGURATION_TYPES="${buildType}" -DMACOS_ARCHITECTURE=arm64 -G Xcode -DIMAGE_COMPILER_EXECUTABLE="${buildDirectory}/x64/utils/${buildType}/image-compiler"
```
Paul-Licameli, line 14: Why was the `-DIMAGE_COMPILER_EXECUTABLE` not given on line 11 too?

Paul-Licameli (approved):
Mostly it's "I trust you" for the .sh file.  I don't understand why it is a related part of this bug fix.

I last disturbed those lines in Languages.cpp, but so long ago!  New changes look all right and do preserve the prior fix I commented about -- trying again as needed without country code.

scripts/build/macOS/build_universal.sh
crsib: Line 11 builds image-compiler, line 14 does not

crsib:
`.sh` file is a helper script to build universal builds locally, which is not a most straightforward process

Issue #2779, opened 2022-04-08
Combine the Mixer toolbar with the meter toolbars


LWinterberg:
![image](https://user-images.githubusercontent.com/87814144/162411523-13074e0d-6847-4ec8-9563-8d1ebf545ae6.png)

This code already is in DarkAudacity (and perhaps as `EXPERIMENTAL_METER_SLIDERS` in Audacity itself?)

Compared to the DarkAudacity implementation, there should be some changes: 

- [ ] Keep the numbering on the meter
- [ ] Keep using the gradient style, don't use the LED or RMS style
- [ ] Change the sliders to be logarithmic relative to the meter range. That is to say, if Preferences > Interface specifies a range of -60dB, the slider should go from -60dB to 0dB. However, if the slider is dragged to the left beyond the meter area, the volume should be set to -inf. 
- [ ] Add accessibility as specified in figma
- [ ] OPTIONAL: Make the slider respond to the meter options (dB (default) vs linear).

petersampsonaudacity:
OOH, yes please _(lets us free up some tooldock real estate)_ 

@JamesCrook  was always intending to migrate this to normal Audacity at some stage but never got around to it.  

IIRC I think Gale Andrews (RIP) blocked this particular enhancement when we worked on moving many of James' changes for DarkAudacity (including Themes) into the mainstream.

petersampsonaudacity:
But hopefully you will do it retaining the current default of Gradient meters rather than the older RMS meters that @JamesCrook  had as his default (and as show in @LWinterberg  Leo's first post in this thread

petersampsonaudacity:
As part of that change 0 @JamesCrook also added some rather nice digital metering that gave the meters a more modern look

![image](https://user-images.githubusercontent.com/50205984/162467284-bdd76891-179c-42ac-9f5f-103dfe7fa5c1.png)

I presume the code for that also exists in DarkAudacity - so this too should be quick and easy to implement - any chance that this can make it into 3.2.0 >

petersampsonaudacity:
Note that @JamesCrook  removed the numbers on the metering for DarkAudacity - the Team consensus at the time was that we all felt that the numbers were useful and should remain.

DavidBailes:
> Compared to the DarkAudacity implementation, there should be some changes:

another necessary change is that these new meters need to to be made accessible for users of screen readers - they are not in the version in dark audacity.

My suggested way of doing this is follows.
[...]

PR #3133, opened 2022-06-24
Combined mixer toolbar with meter toolbars

ksoze95:
Resolves: #2779 

Recommended:


DavidBailes:
It currently isn't possible to tab to the slider.
Also, note my comment on the original issue: https://github.com/audacity/audacity/issues/2779#issuecomment-1093835971
I think it would work better for screen readers if the meter control, that is the recording/playback icon which acts as a button, and the slider are siblings. The alternative, that the recording/playback button contains a slider doesn't make sense.
I think that your current scheme is that the meter contains the slider. If I was able to tab to the slider, I could check out how this is read by the various screen readers on Windows.

src/widgets/ASlider.cpp
```diff
@@ -480,12 +480,17 @@ LWSlider::LWSlider(wxWindow * parent,
                      float stepValue,
                      bool canUseShift,
                      int style,
+                     bool showlabels /* = true */,
```
Paul-Licameli, line 483: Misspelled “without” in the commit comment

ksoze95: Done.  Corrected spelling mistake.

src/widgets/MeterPanel.cpp
```diff
@@ -341,6 +342,17 @@ MeterPanel::MeterPanel(AudacityProject *project,
    mRuler.SetLabelEdges(true);
    //mRuler.SetTickColour( wxColour( 0,0,255 ) );
 
+   mSlider = std::make_unique<LWSlider>(this, XO(""),
```
Paul-Licameli, line 345: There are MeterPanels also in the Mixer board, which you can open using the View menu.  It makes a control for each audio track, and a meter panel with isInput = false.

Did you test this too?

I don't see unintended changes in the Mixer board yet resulting from this change.  No slider is drawn over that meter.  I don't understand why not, but it is as it should be.

src/widgets/MeterPanel.cpp
```diff
@@ -671,65 +716,16 @@ void MeterPanel::OnPaint(wxPaintEvent & WXUNUSED(event))
    }
 #endif
 
-   // Let the user know they can click to start monitoring
-   if( mIsInput && !mActive )
+   if (mStyle != MixerTrackCluster)
```
Paul-Licameli, line 719: Now I see it.  All the painting of the slider is skipped for mixer tracks.

Maybe then mSlider should remain null in case of mixer board clusters.  Avoid making a useless LWSlider.

src/toolbars/MeterToolBar.cpp
```diff
@@ -72,6 +72,19 @@ MeterToolBar::~MeterToolBar()
...
+{
+   return MeterToolBars{
+      Get(project, PlayMeterBarID),
+      Get(project, RecordMeterBarID)
+   };
+}
+
+const MeterToolBars MeterToolBar::GetToolBars(const AudacityProject &project)
+{
+   return GetToolBars( const_cast<AudacityProject&>( project ));
```
Paul-Licameli, line 85: This use of reference wrappers is nice.

However this function exposes a vector of non-const reference (wrappers) to toolbars and only makes the returned vector const.

Not your intention I think:  so I think you must define this return type:

```
using ConstMeterToolBars = std::vector< std::reference_wrapper<Const MeterToolBar> >;
```

... and not call through to the other function.

Paul-Licameli: However, you don't even use the const overload, I think

Paul-Licameli (approved):

PR #3134, opened 2022-06-24
Dont set sample rate

Paul-Licameli:
The next step in LV2 refactoring led to a long digression into other effects.

The EffectInstance abstract class had a virtual function for change of sample
rate, but whether it was really properly done for Ladspa and LV2 effects was
doubtful.

This transformation eliminates the need for that virtual function and ultimately
allows some simplifications of lv2 effect support.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -391,7 +391,7 @@ public:
     Other member functions related to realtime return true or zero, but will not
     be called, unless a derived class overrides RealtimeInitialize.
     */
-   virtual bool RealtimeInitialize(EffectSettings &settings);
+   virtual bool RealtimeInitialize(EffectSettings &settings, double sampleRate);
 
    /*!
     @return success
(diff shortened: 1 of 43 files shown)


libraries/lib-components/EffectInterface.h
```diff
@@ -391,7 +391,7 @@ class COMPONENTS_API EffectInstance
     Other member functions related to realtime return true or zero, but will not
     be called, unless a derived class overrides RealtimeInitialize.
     */
-   virtual bool RealtimeInitialize(EffectSettings &settings);
+   virtual bool RealtimeInitialize(EffectSettings &settings, double sampleRate);
```
crsib, line 392: I always wondered why we use double for sample rate. I've never seen non integer rates

crsib (approved):

Paul-Licameli:
The next step in LV2 refactoring led to a long digression into other effects.

The EffectInstance abstract class had a virtual function for change of sample
rate, but whether it was really properly done for Ladspa and LV2 effects was
doubtful.

This transformation eliminates the need for that virtual function and ultimately
allows some simplifications of lv2 effect support.

libraries/lib-components/EffectInterface.h
Paul-Licameli: The AudioUnit library uses double, who knows why.

PR #3136, opened 2022-06-25
Update AboutDialog.cpp

LWinterberg:


src/AboutDialog.cpp
```diff
@@ -223,6 +224,7 @@ void AboutDialog::CreateCreditsList()
    AddCredit(wxT("Mark Phillips"), developerFormat, roleContributor);
    AddCredit(wxT("Andr\u00E9 Pinto"), developerFormat, roleContributor);
    AddCredit(wxT("Jean Claude Risset"), composerFormat, roleContributor);
+   AddCredit(wxT("RuRo"), developerFormat, roleContributor);
```
Paul-Licameli, line 227: Are we okay with putting funny nicknames in the credits (instead of real names or at least pseudonyms)?  Does it look professional enough?  Should such handles be "quoted" Just asking an opinion @Tantacrul

Paul-Licameli: And do these individuals know they are being credited and have they agreed to being identified exactly so?

Paul-Licameli: If the answers are yes to all of that, ask me about someone who deserves an overdue credit too, though never identified to us by a real name.

SteveDaulton (commented):
Although not directly about this pull request, "Jean Claude Risset" is not directly a contributor to Audacity. I suspect that his name was erroneously added to the credits because he is mentioned here: https://github.com/audacity/audacity/blob/5d10d6f2d1d636a1b267c5dcc45cb3da0421f01f/plug-ins/rissetdrum.ny#L13

Paul-Licameli:
I am pleased to see the long overdue credit added for @Pokechu22 .

crsib (approved):

PR #3138, opened 2022-06-25
[WINDOWS] Add support to GNUC in addition to MSVC 

carlo-bramini:
I tried to build with MinGW-w64 and I got this error:

```
i686-w64-mingw32-g++: fatal error: cannot specify ‘-o’ with ‘-c’, ‘-S’ or ‘-E’ with multiple files
compilation terminated.
```

This happens because this line assumes to include configwin.h only to MSVC with /FI option:

https://github.com/audacity/audacity/blob/9000d4cb37dafa8d122bb696a80d56872623fa29/cmake-proxies/cmake-modules/AudacityFunctions.cmake#L230

Attached patch allows to recognize also GNUC and it fixed the build.

Recommended:


cmake-proxies/cmake-modules/AudacityFunctions.cmake
```diff
@@ -227,7 +227,10 @@ function( audacity_append_common_compiler_options var use_pch )
          PRIVATE
             # include the correct config file; give absolute path to it, so
             # that this works whether in src, modules, libraries
-            $<$<PLATFORM_ID:Windows>:/FI${CMAKE_BINARY_DIR}/src/private/configwin.h>
+            $<$<PLATFORM_ID:Windows>:
+               $<$<CXX_COMPILER_ID:MSVC>:/FI${CMAKE_BINARY_DIR}/src/private/configwin.h>
```
crsib, line 231: I think `$<IF>` makes more sense here. This way we separate MSVC and GCC/CLANG

carlo-bramini: > I think `$<IF>` makes more sense here. This way we separate MSVC and GCC/CLANG

Excuse me... I'm sorry but I have not understood very well how I should use the `$<IF>` expression here.

crsib: Something like

```
$<IF:$<CXX_COMPILER_ID:MSVC>, 
    /FI${CMAKE_BINARY_DIR}/src/private/configwin.h,
    -include ${CMAKE_BINARY_DIR}/src/private/configwin.h
>
```

crsib (approved):

Rossmaxx:
hi i am trying to get audacium (a fork of audacity) compiled on mingw and saw 2 of your prs. if you don't mind, can you please help me.

here's my pr on audacium : https://github.com/Audacium/audacium/pull/119

PR #3145, opened 2022-06-26
Lv2 wrapper

Paul-Licameli:
LV2Wrapper class factored from LV2Effect.cpp, and with more immutable members.

The several files in src/effects/lv2 now have this tidy dependency graph:

Screen Shot 2022-06-26 at 10 36 07 AM

Recommended:


src/effects/lv2/LV2Effect.h
```diff
@@ -339,16 +336,26 @@ class LV2Effect final : public LV2FeaturesList
...
-   //! May spawn a thread
-   LV2Wrapper(const LV2FeaturesList &featuresList, int latencyPort,
+   //! Factory
+   static std::unique_ptr<LV2Wrapper> Create(
+      const LV2FeaturesList &featuresList,
+      const LV2Ports &ports, LV2PortStates &portStates,
+      const LV2EffectSettings &settings, float sampleRate, bool useOutput);
+
+   //! Constructor may spawn a thread
+   LV2Wrapper(CreateToken&&,
```
crsib: I wonder if making the constructor private will do the trick as well

crsib (approved):

src/effects/lv2/LV2Effect.h
Paul-Licameli: Making the constructor private breaks make_shared or make_unique which have to invoke the constructor.

See guideline C.50 for an example of the trick: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-factory

crsib: Makes sense!

Issue #3013, opened 2022-06-01
Scrub ruler and play indicator not displaying correctly


Paul-Licameli:
This problem bisects to commit 5835ae5d87f195ce0e2e8b985586b2ed45c4e1ca which was a preparation for the side bar interface.

- Start program, make some sound
- View > Toolbars > Scrub Toolbar (on)
- Press the third button of the bar, the toggle that shows and hides the ruler
- Play sound; observer the green play head line; stop
- Repeat last, with toggle button up

Observe:  The bottom of the ruler changes appearance slightly, but the scrub zone is only a few pixels high, and the scrub cursor isn't available just from mouse movement over it.

Observe:  The top of the green play head line is too low

Alternative steps:  Don't show the Scrub Toolbar, but use Transport > Scrubbing sub-menu to toggle it

Paul-Licameli:
A second screen shot on macOS, with the toggle down for the scrub ruler.  There is a subtle one or two pixel change in the height of the ruler only.  I can't position my mouse anywhere for the hit test result that changes the play head to the green double-headed scrubbing arrow.  I can do scrubbing if I use the toolbar buttons.

<img width="200" alt="Screen Shot 2022-06-06 at 9 47 06 AM" src="https://user-images.githubusercontent.com/11670369/172174006-5c56c506-8fce-4b8b-9e58-67ecf326e5ff.png">

Paul-Licameli:
Compare with how it appears when I build commit ec5f272803ab4c524256569a37e6d8fb7c476187, one before the blamed commit.  As it also appears in 3.1.3.

<img width="270" alt="Screen Shot 2022-06-06 at 10 04 08 AM" src="https://user-images.githubusercontent.com/11670369/172176763-2d85ae94-2224-4cb0-8bf8-1aa1803b8537.png">
<img width="362" alt="Screen Shot 2022-06-06 at 10 03 55 AM" src="https://user-images.githubusercontent.com/11670369/172176759-c9d9cf44-350a-474c-a7db-60d4fd015c5b.png">

petersampsonaudacity:
> First screen shot, on macOS, with the scrub ruler off. Observe that the green line does not extend to the top of the white area.

@Paul-Licameli  - I reported similar in a different thread some days ago.  And similar happens when recording. 

_I'll try to find that report - but it may have been on Discord_

Paul-Licameli:
I discovered the problem incidentally when rewriting our uses of threads.

[...]

PR #3161, opened 2022-06-29
issue #3013: Scrub ruler not displaying correctly

DavidBailes:
Partially Resolves: https://github.com/audacity/audacity/issues/3013

This is a partial fix for issue 3013, which also includes the play indicator not displaying correctly.

Fix:
Now that the timeline has been reparented by the main window, the size event only needs to be sent to the timeline's parent, not grandparent - no other windows are affected by the resize.

Recommended:


Paul-Licameli:
If it's only partial, what part is not fixed?

PR #3167, opened 2022-07-01
Immutable lv2 features lists

Paul-Licameli:
Another detour along the journey to lv2 statelessness...

... Don't recycle one array of stateful Features in the effect object, when making the user
interface -- instead, make an immutable, throw-away list of features each time there is 
a new dialog; and don't questionably expose a pointer to a temporary to the foreign code.

All that was actually itself just a digression, to enable the next:

Fix some hidden shared state still hiding in the mutable sample rate in the Effect object,
moving it instead into per-LV2Wrapper Options and Features.

So that eliminates one piece of shared state in LV2Instance, but it remains also to eliminate
mMaster and mPortStates from the Effect object.

Recommended:

### DIFF
--- a/src/effects/lv2/LV2Effect.cpp
+++ b/src/effects/lv2/LV2Effect.cpp
@@ -831,7 +831,9 @@ bool LV2Effect::BuildFancy(LV2Validator &validator,
    const auto uinode = lilv_ui_get_uri(ui);
    lilv_world_load_resource(gWorld, uinode);
    UIHandler &handler = *this;
-   auto &features = validator.mUIFeatures.emplace(mFeatures, handler, uinode);
+   auto &features = validator.mUIFeatures.emplace(
+      mFeatures, handler, uinode, &instance,
+      (uiType == node_ExternalUI) ? nullptr : mParent);
    if (!features.mOk)
       return false;
 
@@ -840,18 +842,12 @@ bool LV2Effect::BuildFancy(LV2Validator &validator,
       containerType = LV2_EXTERNAL_UI__Widget;
    else {
       containerType = nativeType;
-      features.mFeatures[features.mParentFeature].data = mParent->GetHandle();
 #if defined(__WXGTK__)
       // Make sure the parent has a window
       if (!gtk_widget_get_window(GTK_WIDGET(mParent->m_wxwindow)))
          gtk_widget_realize(GTK_WIDGET(mParent->m_wxwindow));
 #endif
    }
-
-   features.mFeatures[features.mInstanceAccessFeature].data =
-      lilv_instance_get_handle(&instance);
-   features.mExtensionDataFeature =
-      { lilv_instance_get_descriptor(&instance)->extension_data };
 
    // Set before creating the UI instance so the initial size (if any) can be captured
    mNativeWinInitialSize = wxDefaultSize;
--- a/src/effects/lv2/LV2Effect.cpp
+++ b/src/effects/lv2/LV2Effect.cpp
@@ -84,6 +84,14 @@ LV2Instance::~LV2Instance()
       .mMaster.reset();
 }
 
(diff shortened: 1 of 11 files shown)


src/effects/lv2/LV2Effect.cpp
```diff
@@ -835,8 +835,9 @@ bool LV2Effect::BuildFancy(LV2Validator &validator,
 
    const auto uinode = lilv_ui_get_uri(ui);
    lilv_world_load_resource(gWorld, uinode);
-   auto &features =
-      validator.mUIFeatures.emplace(mFeatures, (UIHandler&)*this, uinode);
+   auto &features = validator.mUIFeatures.emplace(
+      mFeatures, (UIHandler&)*this, uinode, &instance,
```
crsib: I understand, that it is not a goal, but C-cast?

crsib (changes requested):
Linux build is broken

src/effects/lv2/LV2FeaturesList.cpp
```diff
@@ -178,9 +186,11 @@ bool LV2FeaturesList::CheckFeatures(const LilvNode *subject, bool required)
             /* Supported but handled in LV2Wrapper */
          }
          else if (required) {
-            const auto end = mFeatures.end();
-            supported = (end != std::find_if(mFeatures.begin(), end,
-               [&](auto &feature){ return strcmp(feature.URI, uri) == 0; }));
+            auto features = GetFeaturePointers();
```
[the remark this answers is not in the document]
Paul-Licameli: GetFeaturePointers() builds a vector of pointers, as you may recall from reviewing an earlier PR.

I could write
`const auto &features = ...`

and it would also compile, and the reviewer might not then have to pause and think about it.

src/effects/lv2/LV2UIFeaturesList.h
```diff
@@ -0,0 +1,47 @@
...
+
+#if USE_LV2
+
+#include "lv2_external_ui.h"
+#include <suil/suil.h>
+
+struct LV2UIFeaturesList final {
+   //! Abstraction of host services that a plug-ins native UI needs
+   struct UIHandler {
+      ~UIHandler();
```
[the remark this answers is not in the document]
Paul-Licameli: Oh for shame

src/effects/lv2/LV2Effect.cpp
Paul-Licameli: That is a cast to base class, up-cast not down-, and needed because of private inheritance of `UIHander`.  Else the up-cast is attempted inside a function of `std::optional` where there isn't access.

src/effects/lv2/LV2FeaturesList.cpp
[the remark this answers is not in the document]
Paul-Licameli: As for threads, this should be done only in construction and initializations of the plug-in, or its instances, or its validators, in the main thread

src/effects/lv2/LV2Effect.cpp
```diff
@@ -906,33 +893,22 @@ bool LV2Effect::BuildFancy(
...
       lilv_node_as_uri(lilv_ui_get_uri(ui)), lilv_node_as_uri(uiType),
       bundlePath.get(), binaryPath.get(),
-      mUIFeatures.GetFeaturePointers().data()));
+      features.GetFeaturePointers().data()));
 
    // Bail if the instance (no compatible UI) couldn't be created
-   if (!mSuilInstance)
-   {
-#if defined(__WXMSW__)
-      SetDllDirectory(NULL);
```
crsib, line 918: I wonder if this code had side effects

crsib: But I see that it is invoked in a better way now!

crsib (approved):

PR #3168, opened 2022-07-01
Issue 3013: Fixes AdornedRulerPanel resize event sent to the wrong parent

vsverchinsky:
Resolves: #3013 

Fixes second part of the issue with the ruler panel

Recommended:


crsib:
Complete duplicate of #3161

Issue #3102, opened 2022-06-17
Make realtime effects settings non-modal


LWinterberg:
A user must be able to have multiple effects settings open, change the settings and jump around their project to get things just right. As such, the realtime effects settings windows cannot be modal. 

Cases to test, in which the user performs other editing actions, which should cause the dialog to close:

- track is deleted
- Effect on the track is deleted or replaced
- Undo or redo causes the track or the effect to disappear

(The first and third of these are blocked during playback, but there is not necessarily playback while the non-modal dialog is open.

The second should be safe during playback, but, test it!)

Case to test, which should cause an update of the dialog:

- undo or redo changes the effect settings

Design questions about cases where you don’t hit the apply button:

- If I drag sliders, undo, redo — should I get the same slider settings back, though there is no distinct undo history item made for this?  (Which would be similar to the treatment of the power button)
- Ditto in case of drag, redo, undo
- Ditto, if I drag sliders, do some unrelated undo-able edit, undo, redo.

Other cases to test:

- a repeated press of the cogwheel while the dialog is open should do nothing (not open a duplicate dialog)
- if we decide to disallow the dialog during recording, it must be closed when recording begins
- If we permit the dialog during recording (maybe you are overdubbing?), the undo/redo history should keep applied settings changes separate from, and previous to, the recording, even though parts of the recording happen before the hits of the apply button.  In other words, to undo the “Record” item should undo all of it, not just a late part of it.

Paul-Licameli:
@LWinterberg @Tantacrul @Ryan-Miller-Audacity @dozzzzer I updated the description of this issue with the non obvious requirements and I don’t know if this is exhaustive yet.

I agree that non-modality and multiple simultaneous dialogs would be desirable but it will not be very simple.

The previously existing real time effect dialog is in fact non-modal but did not have all of these difficulties because it only managed a one-deep stack of effects, associated with the whole project and not with any one track, and did not require undo and redo of changes.  And even then, I’m not confident that it was ever thoroughly tested and bug free.

LWinterberg:
Aye. Let's have it out-of-scope for this release and look at it for a later release.

Tantacrul:
I think we need to just test the build when it is ready for user testing and then figure out if non-modal dialogs are essential or not. My guess is that they won't be absolutely essential but we may get really bad feedback that changes our mind.

crsib:
I don't think that the dialog should handle Undo/Redo. As a user, I want Undo to work even if the dialog is closed.

[...]

Issue #3090, opened 2022-06-16
Effect plugin window beautifications


LWinterberg:
Dev checklist:

* If `!mwState.expired()`: Power button is visible, no buttons on the right
* Otherwise: No power button:
  * **Apply** is always shown
  * **Preview** is shown if `mSupportsRealtime == true`  
* **Presets & Settings** is always shown

---------------------  

essentially a redesign of the existing play controls on plugins. 
<details><summary>see design files</summary>

![9 0](https://user-images.githubusercontent.com/87814144/174152014-4e24668e-eaf0-4e07-965c-b212a58ebac0.png)
![10 0](https://user-images.githubusercontent.com/87814144/174152035-d851dd76-cc06-48c0-b53c-639571bb49ce.png)
![11 0](https://user-images.githubusercontent.com/87814144/174152042-41c8bce8-7c92-4060-8611-d1041966aeb4.png)

</details>

Note: In addition to what's shown in the images, there needs to be a play/stop button

LWinterberg:
Why are they modal? They weren't modal in 3.1.3. And the designs show the user doing stuff with the UI while the effect settings are open.

Paul-Licameli:
> Why are they modal? They weren't modal in 3.1.3. And the designs show the user doing stuff with the UI while the effect settings are open.

Non-modal dialogs are inherently harder to do right.

Addressing this issue, putting the decorations in place on the dialog, may acutally be the easier way.

LWinterberg:
Okay. Spun off #3102 for later, and added this to the kanban board.

DavidBailes:
@LWinterberg 
1. Which types of plugins are going to be affected by these changes? Does this include Audacity's built-in plugins?
2. For VST, LV2 etc which don't have their graphical user interface enabled, are there going to be any changes? For example:
![image](https://user-images.githubusercontent.com/11887021/177281312-9d76d040-fec5-4442-8808-618149a3cd6a.png)

Paul-Licameli:
I observe that when opening the dialog from the sidebar, there is no button to close it other than the top bar decorations added by the operating system, shown here as macOS style.

Assuming the dialog can be so closed -- or also with the ESC key -- then there is no way to distinguish between keeping the changes you have tried, and cancelling them so the settings revert as they were before you opened it.

I think that's a design omission.  Do you?  @jessjwilliamson @LWinterberg @Tantacrul

crsib:
[...]

PR #3176, opened 2022-07-04
Non modal setting dialogs for stacked effects

crsib:
Resolves: #3102
Resolves: #3366 
Resolves: #3367

This PR makes setting dialogs to be non modal, so Audacity interface can be interactive while the dialog is open.

Key points:

1. One dialog can be opened per one effect instance:
    1. Dialog is closed when the effect instance is removed
    2. Dialog is closed when the track is removed
    3. Dialog survives "destructive" operations on the track, such as recording, using generators and applying destructive effects
    4. Dialog title reflects the name of a track
    5. Dialog title does not (yet) reflect the effect placement inside the stack or any additional parameters (for example - Ardour shows like `Track 1: Distortion [VST3]`). Current design currently lacks this features (@Tantacrul @LWinterberg FYI)
2. This PR does not address #3090. `Apply` behavior is left as is, as it will be removed as a part of #3090
3. Changing effect parameters is not reflected in the Undo Stack. This should be a separate task, and it should be a responsibility of each individual effect subclass to implement such functionality: we need to reflect the changed parameter in the undo text, like "Undo Change Gain". It has been agreed that this can be solved outside the Audacity 3.2 scope. For reference - Garage Band and Motu Performer offer no such functionality and it is unreliable in Ableton.

Recommended:

### DIFF
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -1307,7 +1307,9 @@ struct TrackListEvent
       ADDITION,
 
       //! Posted when a track has been deleted from a tracklist. Also posted when one track replaces another
-      /*! mpTrack points to the first track after the deletion, if there is one. */
+      /*! mpTrack points to the removed track. It is expected, that track is valid during the event.
+       *! mExtra is 1 if the track is being replaced by another track, 0 otherwise.
+       */
       DELETION,
    };
 
@@ -1709,7 +1711,7 @@ private:
    void DataEvent( const std::shared_ptr<Track> &pTrack, int code );
    void EnsureVisibleEvent(
       const std::shared_ptr<Track> &pTrack, bool modifyState );
-   void DeletionEvent(TrackNodePointer node = {});
+   void DeletionEvent(std::weak_ptr<Track> node, bool duringReplace);
    void AdditionEvent(TrackNodePointer node);
    void ResizingEvent(TrackNodePointer node);
 
--- a/src/EffectPlugin.h
+++ b/src/EffectPlugin.h
@@ -6,7 +6,7 @@
 
    Paul Licameli
    split from EffectInterface.h
-   
+
 **********************************************************************/
 
 #ifndef __AUDACITY_EFFECTPLUGIN_H__
@@ -83,6 +83,14 @@ public:
       wxWindow &parent, const EffectDialogFactory &factory,
       std::shared_ptr<EffectInstance> &pInstance, EffectSettingsAccess &access,
       bool forceModal = false) = 0;
+
+   //! Returns the EffectUIClientInterface instance for this effect
+   /*!
+    * Usually returns self. May return nullptr. EffectPlugin is responsible for the lifetime of the
+    * returned instance.
+    * @return EffectUIClientInterface object or nullptr, if the effect does not implement the interface.
+    */
+   virtual EffectUIClientInterface* GetEffectUIClientInterface() = 0;
 
    virtual void Preview(EffectSettingsAccess &access, bool dryOnly) = 0;
    virtual bool SaveSettingsAsString(
--- a/src/ProjectWindow.cpp
+++ b/src/ProjectWindow.cpp
@@ -640,7 +640,7 @@ ProjectWindow::ProjectWindow(wxWindow * parent, wxWindowID id,
 
    mContainerWindow->Initialize(mTrackListWindow);
 
-   auto effectsPanel = safenew ThemedWindowWrapper<RealtimeEffectPanel>(mContainerWindow, wxID_ANY);
+   auto effectsPanel = safenew ThemedWindowWrapper<RealtimeEffectPanel>(mProject, mContainerWindow, wxID_ANY);
    effectsPanel->SetBackgroundColorIndex(clrMedium);
    effectsPanel->Hide();//initially hidden
    effectsPanel->Bind(wxEVT_CLOSE_WINDOW, [this](wxCloseEvent&)
--- a/src/ProjectWindow.cpp
+++ b/src/ProjectWindow.cpp
@@ -699,7 +699,7 @@ ProjectWindow::ProjectWindow(wxWindow * parent, wxWindowID id,
          {
             auto& project = GetProject();
             auto& trackFocus = TrackFocus::Get(project);
-            ShowEffectsPanel(project, trackFocus.Get());
+            ShowEffectsPanel(trackFocus.Get());
          }
       });
 
@@ -1906,7 +1906,7 @@ void ProjectWindow::DoZoomFit()
    window.TP_ScrollWindow(start);
 }
 
-void ProjectWindow::ShowEffectsPanel(AudacityProject& project, Track* track)
+void ProjectWindow::ShowEffectsPanel(Track* track)
 {
    if(track == nullptr)
    {
@@ -1916,7 +1916,7 @@ void ProjectWindow::ShowEffectsPanel(AudacityProject& project, Track* track)
 
    wxWindowUpdateLocker freeze(this);
 
-   mEffectsWindow->SetTrack(project, track->shared_from_this());
+   mEffectsWindow->SetTrack(track->shared_from_this());
 
    if(mContainerWindow->GetWindow1() != mEffectsWindow)
    {
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -43,6 +43,29 @@
 
 namespace
 {
+   void UpdateRealtimeEffectUIData(Track& track)
+   {
+      auto& effects = RealtimeEffectList::Get(track);
+
+      effects.Visit(
+         [&track](auto& effectState, bool)
+         {
+            auto& ui = RealtimeEffectStateUI::Get(effectState);
+            ui.UpdateTrackData(track);
+         });
+   }
+
+   void HideRealtimeUIForTrack(Track& track)
+   {
+      auto& effects = RealtimeEffectList::Get(track);
+
+      effects.Visit(
+         [&track](auto& effectState, bool)
+         {
+            auto& ui = RealtimeEffectStateUI::Get(effectState);
+            ui.Hide();
+         });
+   }
    //fwd
    class RealtimeEffectControl;
    PluginID ShowSelectEffectMenu(wxWindow* parent, RealtimeEffectControl* currentEffectControl = nullptr);
@@ -379,6 +402,9 @@ namespace
          if(mProject == nullptr || mEffectState == nullptr)
             return;
 
+         auto& ui = RealtimeEffectStateUI::Get(*mEffectState);
+         ui.Hide();
+
          auto effectName = GetEffectName();
          //After AudioIO::RemoveState call this will be destroyed
          auto project = mProject.get();
@@ -412,6 +438,8 @@ namespace
          }
 
          auto& effectStateUI = RealtimeEffectStateUI::Get(*mEffectState);
+
+         effectStateUI.UpdateTrackData(*mTrack);
          effectStateUI.Toggle( *mProject );
       }
 
@@ -797,10 +825,7 @@ public:
          mEffectListItemMovedSubscription = effects.Subscribe(
             *this, &RealtimeEffectListWindow::OnEffectListItemChange);
 
-         effects.Visit([track](auto& effect, bool) {
-            auto& ui = RealtimeEffectStateUI::Get(effect);
-            ui.UpdateTrackData(*track);
-         });
+         UpdateRealtimeEffectUIData(*track);
       }
    }
 
@@ -877,9 +902,12 @@ public:
    }
 };
 
-RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const wxPoint& pos, const wxSize& size,
+RealtimeEffectPanel::RealtimeEffectPanel(
+   AudacityProject& project, wxWindow* parent, wxWindowID id, const wxPoint& pos,
+   const wxSize& size,
    long style, const wxString& name)
       : wxWindow(parent, id, pos, size, style, name)
+      , mProject(project)
 {
    auto vSizer = std::make_unique<wxBoxSizer>(wxVERTICAL);
 
@@ -901,8 +929,8 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const 
             mEffectList->EnableEffects(!wasEnabled);
          }
          pButton->SetBitmapIndex(wasEnabled ? bmpEffectOff : bmpEffectOn);
-         if (mProject)
-            ProjectHistory::Get(*mProject).ModifyState(false);
+         
+         ProjectHistory::Get(mProject).ModifyState(false);
       });
 
       hSizer->Add(toggleEffects, 0, wxSTRETCH_NOT | wxALIGN_CENTER | wxLEFT, 5);
@@ -943,6 +971,35 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const 
    mEffectList = effectList;
 
    SetSizerAndFit(vSizer.release());
+
+   mTrackListChanged = TrackList::Get(mProject).Subscribe([this](const TrackListEvent& evt) {
+         auto track = evt.mpTrack.lock();
+         auto waveTrack = dynamic_cast<WaveTrack*>(track.get());
+
+         if (waveTrack == nullptr)
+            return;
+         
+         switch (evt.mType)
+         {
+         case TrackListEvent::TRACK_DATA_CHANGE:
+            if (mCurrentTrack.lock() == track)
+               mTrackTitle->SetLabel(track->GetName());
+            UpdateRealtimeEffectUIData(*waveTrack);
+            break;
+         case TrackListEvent::DELETION:
+            if (evt.mExtra == 0)
+               HideRealtimeUIForTrack(*waveTrack);
+            break;
+         case TrackListEvent::ADDITION:
+            // Addition can be fired as a part of "replace" event.
+            // Calling UpdateRealtimeEffectUIData is mostly no-op,
+            // it will just create a new State and Access for it.
+            UpdateRealtimeEffectUIData(*waveTrack);
+            break;
+         default:
+            break;
+         }
+   });
 }
 
 void RealtimeEffectPanel::SetTrack(AudacityProject& project, const std::shared_ptr<Track>& track)
@@ -958,7 +1015,8 @@ void RealtimeEffectPanel::SetTrack(AudacityProject& project, const std::shared_p
             : bmpEffectOff
       );
       mEffectList->SetTrack(project, track);
-      mProject = &project;
+
+      mCurrentTrack = track;
    }
    else
       ResetTrack();
@@ -970,5 +1028,5 @@ void RealtimeEffectPanel::ResetTrack()
    mToggleEffects->SetBitmapIndex(bmpEffectOff);
    mToggleEffects->Enable(false);
    mEffectList->ResetTrack();
-   mProject = nullptr;
+   mCurrentTrack.reset();
 }
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -1075,7 +1075,7 @@ RealtimeEffectPanel::RealtimeEffectPanel(
       });
 }
 
-void RealtimeEffectPanel::SetTrack(AudacityProject& project, const std::shared_ptr<Track>& track)
+void RealtimeEffectPanel::SetTrack(const std::shared_ptr<Track>& track)
 {
    //Avoid creation-on-demand of a useless, empty list in case the track is of non-wave type.
    if(track && dynamic_cast<WaveTrack*>(&*track) != nullptr)
@@ -1087,7 +1087,7 @@ void RealtimeEffectPanel::SetTrack(AudacityProject& project, const std::shared_p
             ? bmpEffectOn
             : bmpEffectOff
       );
-      mEffectList->SetTrack(project, track);
+      mEffectList->SetTrack(mProject, track);
 
       mCurrentTrack = track;
    }
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -41,6 +41,7 @@
 #include "effects/RealtimeEffectState.h"
 #include "effects/RealtimeEffectStateUI.h"
 #include "UndoManager.h"
+#include "Prefs.h"
 
 namespace
 {
@@ -65,6 +66,20 @@ namespace
    void HideRealtimeUIForTrack(Track& track)
    {
       VisitRealtimeEffectStateUIs(track, [](auto& ui) { ui.Hide(); });
+   }
+
+   void ReopenRealtimeEffectUIData(AudacityProject& project, Track& track)
+   {
+      VisitRealtimeEffectStateUIs(
+         track,
+         [&](auto& ui)
+         {
+            if (ui.IsShown())
+            {
+               ui.Hide();
+               ui.Show(project);
+            }
+         });
    }
    //fwd
    class RealtimeEffectControl;
@@ -902,12 +917,30 @@ public:
    }
 };
 
+
+struct RealtimeEffectPanel::PrefsListenerHelper : PrefsListener
+{
+   AudacityProject& mProject;
+
+   explicit PrefsListenerHelper(AudacityProject& project)
+       : mProject { project }
+   {}
+
+   void UpdatePrefs() override
+   {
+      auto& trackList = TrackList::Get(mProject);
+      for (auto waveTrack : trackList.Any<WaveTrack>())
+         ReopenRealtimeEffectUIData(mProject, *waveTrack);
+   }
+};
+
 RealtimeEffectPanel::RealtimeEffectPanel(
    AudacityProject& project, wxWindow* parent, wxWindowID id, const wxPoint& pos,
    const wxSize& size,
    long style, const wxString& name)
       : wxWindow(parent, id, pos, size, style, name)
       , mProject(project)
+      , mPrefsListenerHelper(std::make_unique<PrefsListenerHelper>(project))
 {
    auto vSizer = std::make_unique<wxBoxSizer>(wxVERTICAL);
 
@@ -1073,6 +1106,10 @@ RealtimeEffectPanel::RealtimeEffectPanel(
 
          mPotentiallyRemovedTracks.clear();
       });
+}
+
+RealtimeEffectPanel::~RealtimeEffectPanel()
+{
 }
 
 void RealtimeEffectPanel::SetTrack(const std::shared_ptr<Track>& track)
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -40,31 +40,31 @@
 #include "effects/RealtimeEffectList.h"
 #include "effects/RealtimeEffectState.h"
 #include "effects/RealtimeEffectStateUI.h"
+#include "UndoManager.h"
 
 namespace
 {
-   void UpdateRealtimeEffectUIData(Track& track)
+   template <typename Visitor>
+   void VisitRealtimeEffectStateUIs(Track& track, Visitor&& visitor)
    {
       auto& effects = RealtimeEffectList::Get(track);
-
       effects.Visit(
-         [&track](auto& effectState, bool)
+         [visitor](auto& effectState, bool)
          {
             auto& ui = RealtimeEffectStateUI::Get(effectState);
-            ui.UpdateTrackData(track);
+            visitor(ui);
          });
+   }
+
+   void UpdateRealtimeEffectUIData(Track& track)
+   {
+      VisitRealtimeEffectStateUIs(
+         track, [&](auto& ui) { ui.UpdateTrackData(track); });
    }
 
    void HideRealtimeUIForTrack(Track& track)
    {
-      auto& effects = RealtimeEffectList::Get(track);
-
-      effects.Visit(
-         [&track](auto& effectState, bool)
-         {
-            auto& ui = RealtimeEffectStateUI::Get(effectState);
-            ui.Hide();
-         });
+      VisitRealtimeEffectStateUIs(track, [](auto& ui) { ui.Hide(); });
    }
    //fwd
    class RealtimeEffectControl;
@@ -988,7 +988,7 @@ RealtimeEffectPanel::RealtimeEffectPanel(
             break;
          case TrackListEvent::DELETION:
             if (evt.mExtra == 0)
-               HideRealtimeUIForTrack(*waveTrack);
+               mPotentiallyRemovedTracks.push_back(track);
             break;
          case TrackListEvent::ADDITION:
             // Addition can be fired as a part of "replace" event.
@@ -1000,6 +1000,79 @@ RealtimeEffectPanel::RealtimeEffectPanel(
             break;
          }
    });
+
+   mUndoSubscription = UndoManager::Get(mProject).Subscribe(
+      [this](UndoRedoMessage message)
+      {
+         if (
+            message.type == UndoRedoMessage::Type::Purge ||
+            message.type == UndoRedoMessage::Type::BeginPurge ||
+            message.type == UndoRedoMessage::Type::EndPurge)
+            return;
+
+         auto& trackList = TrackList::Get(mProject);
+
+         // Realtime effect UI is only updated on Undo or Redo
+         auto waveTracks = trackList.Any<WaveTrack>();
+         
+         if (
+            message.type == UndoRedoMessage::Type::UndoOrRedo ||
+            message.type == UndoRedoMessage::Type::Reset)
+         {
+            for (auto waveTrack : waveTracks)
+               UpdateRealtimeEffectUIData(*waveTrack);
+         }
+
+         // But mPotentiallyRemovedTracks processing happens as fast as possible.
+         // This event is fired right after the track is deleted, so we do not
+         // hold the strong reference to the track much longer than need.
+         if (mPotentiallyRemovedTracks.empty())
+            return;
+
+         // Collect RealtimeEffectUIs that are currently shown
+         // for the potentially removed tracks
+         std::vector<RealtimeEffectStateUI*> shownUIs;
+         
+         for (auto track : mPotentiallyRemovedTracks)
+         {
+            // By construction, track cannot be null
+            assert(track != nullptr);
+
+            VisitRealtimeEffectStateUIs(
+               *track,
+               [&shownUIs](auto& ui)
+               {
+                  if (ui.IsShown())
+                     shownUIs.push_back(&ui);
+               });
+         }
+
+         // For every UI shown - check if the corresponding state
+         // is reachable from the current track list.
+         for (auto effectUI : shownUIs)
+         {
+            bool reachable = false;
+            
+            for (auto track : waveTracks)
+            {               
+               VisitRealtimeEffectStateUIs(
+                  *track,
+                  [effectUI, &reachable](auto& ui)
+                  {
+                     if (effectUI == &ui)
+                        reachable = true;
+                  });
+
+               if (reachable)
+                  break;
+            }
+
+            if (!reachable)
+               effectUI->Hide();
+         }
+
+         mPotentiallyRemovedTracks.clear();
+      });
 }
 
 void RealtimeEffectPanel::SetTrack(AudacityProject& project, const std::shared_ptr<Track>& track)
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -5,7 +5,7 @@
    @file RealtimeEffectPanel.cpp
 
    @author Vitaly Sverchinsky
-   
+
 **********************************************************************/
 
 #include "RealtimeEffectPanel.h"
@@ -39,34 +39,10 @@
 #include "effects/EffectManager.h"
 #include "effects/RealtimeEffectList.h"
 #include "effects/RealtimeEffectState.h"
+#include "effects/RealtimeEffectStateUI.h"
 
 namespace
 {
-   auto DialogFactory(
-      const std::shared_ptr<RealtimeEffectState> &pEffectState
-   ){
-      return [=](wxWindow &parent,
-         EffectPlugin &host, EffectUIClientInterface &client,
-         EffectSettingsAccess &access
-      ) -> DialogFactoryResults {
-         // Make sure there is an associated project, whose lifetime will
-         // govern the lifetime of the dialog
-         if (auto project = FindProjectFromWindow(&parent)) {
-            std::shared_ptr<EffectInstance> pInstance;
-            if (Destroy_ptr<EffectUIHost> dlg{ safenew EffectUIHost{
-                  &parent, *project, host, client, pInstance, access,
-                  pEffectState } }
-               ; dlg->Initialize()
-            ){
-               auto pValidator = dlg->GetValidator();
-               // release() is safe because parent will own it
-               return { dlg.release(), pInstance, pValidator };
-            }
-         }
-         return {};
-      };
-   }
-
    //fwd
    class RealtimeEffectControl;
    PluginID ShowSelectEffectMenu(wxWindow* parent, RealtimeEffectControl* currentEffectControl = nullptr);
@@ -111,7 +87,7 @@ namespace
    public:
       MovableControlEvent(wxEventType eventType, int winid = 0)
          : wxCommandEvent(eventType, winid) { }
-      
+
       void SetSourceIndex(int index) noexcept { mSourceIndex = index; }
       int GetSourceIndex() const noexcept { return mSourceIndex; }
 
@@ -162,9 +138,9 @@ namespace
          event.SetEventObject(this);
          wxPostEvent(target, event);
       }
-      
+
    private:
-      
+
       void OnMouseCaptureLost(wxMouseCaptureLostEvent& event)
       {
          if(mDragging)
@@ -194,12 +170,12 @@ namespace
             return;
 
          mSourceIndex = mTargetIndex = -1;
-         
+
          CaptureMouse();
 
          mInitialPosition = evt.GetPosition();
          mDragging=true;
-         
+
          if(auto sizer = parent->GetSizer())
          {
             for(size_t i = 0, count = sizer->GetItemCount(); i < count; ++i)
@@ -218,12 +194,12 @@ namespace
       {
          if(!mDragging)
             return;
-         
+
          ReleaseMouse();
-         
+
          DragFinished();
       }
-      
+
       void OnMove(wxMouseEvent& evt)
       {
          if(!mDragging)
@@ -241,7 +217,7 @@ namespace
             if(boxSizer->GetOrientation() == wxVERTICAL)
             {
                auto targetIndex = mSourceIndex;
-               
+
                //assuming that items are ordered from top to bottom (i > j <=> y(i) > y(j))
                //compare wxSizerItem position with the current MovableControl position!
                if(GetPosition().y < boxSizer->GetItem(mSourceIndex)->GetPosition().y)
@@ -250,7 +226,7 @@ namespace
                   for(int i = 0; i < mSourceIndex; ++i)
                   {
                      const auto item = boxSizer->GetItem(i);
-                     
+
                      if(GetRect().GetTop() <= item->GetPosition().y + item->GetSize().y / 2)
                      {
                         targetIndex = i;
@@ -342,7 +318,7 @@ namespace
          optionsButton->SetBitmapCurrentIndex(bmpEffectSettingsHover);
          optionsButton->SetBackgroundColorIndex(clrEffectListItemBackground);
          optionsButton->Bind(wxEVT_BUTTON, &RealtimeEffectControl::OnOptionsClicked, this);
-         
+
          auto dragArea = safenew wxStaticBitmap(this, wxID_ANY, theTheme.Bitmap(bmpDragArea));
          dragArea->Disable();
          sizer->Add(dragArea, 0, wxLEFT | wxCENTER, 5);
@@ -361,7 +337,7 @@ namespace
          auto desc = PluginManager::Get().GetPlugin(ID);
          return desc;
       }
-      
+
       //! @pre `mEffectState != nullptr`
       TranslatableString GetEffectName() const
       {
@@ -402,7 +378,7 @@ namespace
       {
          if(mProject == nullptr || mEffectState == nullptr)
             return;
-         
+
          auto effectName = GetEffectName();
          //After AudioIO::RemoveState call this will be destroyed
          auto project = mProject.get();
@@ -420,51 +396,23 @@ namespace
             XO("Remove %s").Format(effectName)
          );
       }
-      
+
       void OnOptionsClicked(wxCommandEvent& event)
       {
          if(mProject == nullptr || mEffectState == nullptr)
             return;//not initialized
-         
+
          const auto ID = mEffectState->GetID();
          const auto effectPlugin = EffectManager::Get().GetEffect(ID);
+
          if(effectPlugin == nullptr)
          {
-            ///TODO: effect is not avaialble
+            ///TODO: effect is not available
             return;
          }
-         auto access = mEffectState->GetAccess();
-         // Copy settings
-         auto initialSettings = access->Get();
-         auto cleanup = EffectManager::Get().SetBatchProcessing(ID);
 
-         std::shared_ptr<EffectInstance> pInstance;
-
-         // Like the call in EffectManager::PromptUser, but access causes
-         // the necessary inter-thread communication of settings changes
-         // if play is in progress
-         bool changed = effectPlugin->ShowHostInterface(
-            ProjectWindow::Get( *mProject), DialogFactory(mEffectState),
-            pInstance, *access, true );
-
-         if (changed)
-         {
-            auto effectName = GetEffectName();
-            ProjectHistory::Get(*mProject).PushState(
-               /*! i18n-hint: undo history record
-                first parameter - realtime effect name
-                second parameter - track name
-                */
-               XO("Changed %s in %s").Format(effectName, mTrack->GetName()),
-               /*! i18n-hint: undo history record
-                first parameter - realtime effect name */
-               XO("Change %s").Format(effectName)
-         );
-         }
-         else
-            // Dialog was cancelled.
-            // Reverse any temporary changes made to the state
-            access->Set(std::move(initialSettings));
+         auto& effectStateUI = RealtimeEffectStateUI::Get(*mEffectState);
+         effectStateUI.Toggle( *mProject );
       }
 
       void OnChangeButtonClicked(wxCommandEvent& event)
@@ -522,7 +470,7 @@ namespace
          dc.SetBrush(theTheme.Colour(clrEffectListItemBorder));
          dc.DrawLine(rect.GetBottomLeft(), rect.GetBottomRight());
       }
-      
+
    };
 
    static wxString GetSafeVendor(const PluginDescriptor& descriptor)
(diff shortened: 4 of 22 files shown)


libraries/lib-track/Track.h
```diff
@@ -1307,7 +1307,9 @@ struct TrackListEvent
       ADDITION,
 
       //! Posted when a track has been deleted from a tracklist. Also posted when one track replaces another
-      /*! mpTrack points to the first track after the deletion, if there is one. */
+      /*! mpTrack points to the removed track. It is expected, that track is valid during the event.
```
Paul-Licameli, line 1310: This commit is ok except I think this comment should be clearer.  What is “valid”?

say maybe instead, “mpTrack points to the removed track, if some other shared pointer is preventing its deletion”

crsib: Hmm, I'm trying to say that track pointer is valid during the event in any case

Paul-Licameli:
On MacOS:

1. Record and count to ten aloud.
2. Open sidebar and add AUDynamicsProcessor (an example with an output meter)
3. Play and observe that the output metering does not work

Paul-Licameli (dismissed):
See previous comment.  "Something must be done" so that output metering works.

That isn't very specific of me, but I do have ideas about how to do it.

I think it will require some major changes of approach, and abandonment of many of these proposed changes.

I may develop a first commit that a revised branch can be based on.

But the changes for track list events are so far an approved, necessary part of the fix.

crsib:
I'm not sure if I follow you. This PR provides a fix for a major UI issue, I do not see how is it related to output metering.

crsib:
BTW following you steps I observe that output metering works (very slow, yes)

Paul-Licameli:
> BTW following you steps I observe that output metering works (very slow, yes)

I said it works, but slowly, in master yet not at all in your branch.  Are you agreeing with all that?

crsib:
I understand the reason of the problem and thinking of a solution now

Paul-Licameli:
General idea:  some parts of RealtimeEffect state that are not now copied, should be put in a structure managed by a `shared_ptr` which is copied.  This includes the pointer to the instance.  The complications with ClientData and EffectSettingsAccessTee may all be unnecessary.

But I must look at it closer to make this precise and see if it all makes sense.

crsib:
>General idea: some parts of RealtimeEffect state that are not now copied

Yep, exactly the way I plan approaching it

>The complications with ClientData

I still like the idea of UI to be attached separately, but this can be moved to be a part of RealtimeEffectState itself . The UI controller will be much more simple though.

> EffectSettingsAccessTee 

Yep, that won't be needed, but I think keeping it "public" does not harm.

The main question is still about the Undo/Redo stacks. Right now "copying" instance has a side effect of preserving the effect settings in the undo stack. I'm not sure if it is a desired behavior provided that Undo/Redo for parameters is postponed  and is not implement is several other DAWs at all. Without Undo/Redo preservation I do not see much complexity required to fix the issue you brought up

src/effects/NoiseReduction.cpp
```diff
@@ -694,6 +694,10 @@ bool EffectNoiseReduction::Process(EffectInstance &, EffectSettings &)
    this->ReplaceProcessedTracks(bGoodResult);
    return bGoodResult;
 }
+EffectUIClientInterface* EffectNoiseReduction::GetEffectUIClientInterface()
+{
+   return nullptr;
+}
```
Paul-Licameli, line 700: I don't understand why NoiseReduction required special treatment.  Can you explain?

This new function needs to be called only from RealtimeEffectStateUI, and Noise Reduction is not one of the realtime capable effects, so I don't understand how you can ever reach this override.

Paul-Licameli:
Not yet a complete review.  I pushed a suggested nicer looking way (to me anyway) to implement "Do not set Effect to be a controller..."

crsib:
Your commit introduces a new dependency for the `lib-components`, it now requires access to wxWindow. I'm still checking if it works at all.

crsib:
So is it really worth introducing a new dependency just to remove on `if`? I say it is up to you.

src/effects/NoiseReduction.cpp
crsib: As you know, I do not trust in "something is not called now" :-) The reason `GetEffectUIClientInterface` is there for symmetry. I had some reasoning why this was not done for `NyquistEffect`, but I don't remember it.

Paul-Licameli:
> So is it really worth introducing a new dependency just to remove on `if`? I say it is up to you.

Oh my, how did I overlook that?  Good point: if lib-components should only use wxBase, then the DefaultEffectUIValidator subclass should also be moved somewhere else, into Audacity.

However, actually lib-components and EffectUIClientInterface is still a somewhat confused and confusing piece of the legacy code.  There used to be several other badly named interface classes with unclear purpose that I figured out how to eliminate in this release.  EffectUIClientInterface is still one of those unsatisfying ones I haven't figured out how to clean up just yet.

crsib:
To be fair, at some point we will need to have some façade for the event handlers anyway. So probably we can allow linking to `wxwidgets::core` here for now

>EffectUIClientInterface 

I am really confused about it, yes :-)

Paul-Licameli:
You mention Nyquist effects too now, which are also never realtime capable, but I don't see special treatment of them in this branch.

Nyquist and NoiseReduction are also the only effects that override ShowHostInterface specially, and one of your commits deletes a call to ShowHostInterface.

So is this part of the explanation?

crsib:
Looking on `int Effect::ShowClientInterface` I think the reason is that NoiseReduction behaves very differently from other effects, including Nyquist. I have no strong feelings about that change and I'm happy to remove it.

Paul-Licameli:
I will make another PR about the popping of event handlers, and you can rebase onto it if it is approved and merged.

It will also move DefaultEffectUIValidator to Audacity.

crsib:
Hope I will see it soon, this PR will require vast amount of testing.

Paul-Licameli:
See #3276

crsib:
Well, that was really quick, thank you, Paul!

Paul-Licameli:
The proper complementary pushing and popping was a detail I recall having doubts about, and I'm glad to have reviewed and settled it to my satisfaction now.

src/RealtimeEffectPanel.cpp
```diff
@@ -445,7 +445,7 @@ namespace
          // if play is in progress
          bool changed = effectPlugin->ShowHostInterface(
             ProjectWindow::Get( *mProject), DialogFactory(mEffectState),
-            pInstance, *access, true );
+            pInstance, *access, false );
```
Paul-Licameli: Maybe the commit that does this should logically be last in the sequence, because all else is in place not to cause bugs in non-modality only then

crsib: Agreed!

src/effects/RealtimeEffectState.cpp
```diff
@@ -213,20 +213,25 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
...
    SetID(id);
 }
 
 RealtimeEffectState::RealtimeEffectState(const RealtimeEffectState &other)
-  : mID{ other.mID }
+  : mEffectStateExtensions(other.mEffectStateExtensions)
+  , mID{ other.mID }
   , mPlugin{ other.mPlugin }
   , mMainSettings{ other.mMainSettings }
+  , mwInstance { other.mwInstance }
```
Paul-Licameli, line 226: This member should move earlier in the header file, before `mMainSettings` (and the initializers ordered appropriately) just so that the `@name` section comments are correct.

src/effects/RealtimeEffectStateUI.h
```diff
@@ -0,0 +1,51 @@
...
+   static RealtimeEffectStateUI &Get(RealtimeEffectState &state);
+   static const RealtimeEffectStateUI &Get(const RealtimeEffectState &state);
+
+   PointerType Clone() const override;
+
+private:
+   void SetActiveState(RealtimeEffectState& state); 
+   
+   class SharedState;
+   std::shared_ptr<SharedState> mSharedState;
```
Paul-Licameli, line 50: RealtimeEffectState has a `Site` which has this thing in a slot which has a `shared_ptr` to the good stuff.

I might have arranged it differently:  RealtimeEffectState has a `shared_ptr` to a SharedState structure that it declares, as a `Site`; then this would be a simple attachment and would not need to be cloneable.

Either would work.  What other attchments to states might be needed in future?  I don't know yet.  The latter way of doing it would make it impossible for them to be un-shared.

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,266 @@
...
+{
+public:
+   ~SharedState()
+   {
+      Hide();
+   }
+
+   void SetActiveState(RealtimeEffectState& state)
+   {
+      mCurrentState = state.shared_from_this();
```
Paul-Licameli, line 44: At first I feared a leaky `shared_ptr` cycle but then I saw `mCurrentState` is a `weak_ptr`.  Good, so use `weak_from_this()` instead to make that more obvious.

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,266 @@
...
+   void UpdateTitle(const Track* track = nullptr)
+   {
+      if (track != nullptr)
+         mTrackName = track->GetName();
+
+      // Lazily evaluate the name of the effect
+      if (mEffectName.empty())
+      {
+         auto currentState = mCurrentState.lock();
+         const auto ID = currentState->GetID();
```
Paul-Licameli, line 146: Should there be a null check on `currentState`?

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+      EffectUIClientInterface* client =
+         effectPlugin->GetEffectUIClientInterface();
+
+      // Effect has no client interface
+      if (client == nullptr)
+         return;
+
+      // mUIHost will acquire the correct instance
+      // and will manage it's lifetime
```
Paul-Licameli, line 101: "its"

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,266 @@
...
+      Destroy_ptr<EffectUIHost> dlg(safenew EffectUIHost(
+         &projectWindow, project, *effectPlugin, *client, pInstance, *access,
+         currentState));
+
+      if (!dlg->Initialize())
+         return;
+
+      // pInstance is updated in EffectUIHost constructor  
+      mCurrentInstance = pInstance;
+      // Dialog is owned by it's parent now!
```
Paul-Licameli, line 118: its

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+      UpdateTitle();
+
+      client->ShowClientInterface(
+         projectWindow, *mEffectUIHost, mEffectUIHost->GetValidator(), false);
+
+      // The dialog was modal? That shouldn't have happened
+      if (mEffectUIHost == nullptr || !mEffectUIHost->IsShown())
+      {
+         assert(mEffectUIHost != nullptr && mEffectUIHost->IsShown());
```
Paul-Licameli, line 129: or just `assert(false)`

crsib: Ironically, there was assert(false), but at some point I have decided to make it more explicit

src/effects/RealtimeEffectStateUI.h
Paul-Licameli: The `shared_ptr` is only lazily initialized.
I think this means that different versions of the state in the undo history my or may not shared state.  Would that have bad consequences?

- make a new state
- do some other edit
- click the cogwheel

Then the current track list points to the shared state but the pushed undo state does not.  Does this mean I could undo, click the wheel again, and have two non-modal dialogs open at once?  Or is there something I haven't seen yet that destroys the dialog whenever you undo and redo?  (But then why have this shared state)?

But swap the last two steps, and then there is sharing.

If I am right that there could be a bug, then the shared_ptr should instead be assigned eagerly whenever a new state is made, then always copied.

So either rearrange the storage as I suggested, or used `BuildAll` in the constructor of `RealtimeEffectState` and have `RealtimeEffectStateUI` alway make the shared state.

src/RealtimeEffectPanel.cpp
```diff
@@ -420,51 +396,23 @@ namespace
...
-               XO("Changed %s in %s").Format(effectName, mTrack->GetName()),
-               /*! i18n-hint: undo history record
-                first parameter - realtime effect name */
-               XO("Change %s").Format(effectName)
-         );
-         }
-         else
-            // Dialog was cancelled.
-            // Reverse any temporary changes made to the state
-            access->Set(std::move(initialSettings));
```
Paul-Licameli, line 467: The deleted code for a modal dialog restored state as it was in case the dialog is cancelled or the Esc key is hit, and it pushes an undo history item for change of state if the dialog is OK'd.

I don't see where this commit preserves that behavior by other means.

src/RealtimeEffectPanel.h
```diff
@@ -35,9 +36,15 @@ class RealtimeEffectPanel : public wxWindow
    ThemedButtonWrapper<wxBitmapButton>* mToggleEffects{nullptr};
    wxStaticText* mTrackTitle {nullptr};
    RealtimeEffectListWindow* mEffectList{nullptr};
-   wxWeakRef<AudacityProject> mProject;
+   AudacityProject& mProject;
```
Paul-Licameli, line 39: Replacing a weak pointer with a reference might look unsafe, but I think it is a safe enough assumption that this window will be destroyed only before the associated project.

src/RealtimeEffectPanel.cpp
crsib: As it was mentioned in PR description - the Undo/Redo is intentionally left unimplemented for the effect parameters. And for the reference - there are several DAWs that do not implement it all as well and at least for Ableton there were problems observed with the implementation (i. e. Undo  restored the parameter value to some unexpected state).

If and when we decide to implement the undo/redo for stacked effect parameters - we will have to create an item for each parameter user changes (see Ableton as the reference). 

The potential problem I expect now is that Audacity project won't get "dirty" on parameter change, but I would rather fix it in a different PR.

src/RealtimeEffectPanel.cpp
```diff
@@ -379,6 +402,9 @@ namespace
          if(mProject == nullptr || mEffectState == nullptr)
             return;
 
+         auto& ui = RealtimeEffectStateUI::Get(*mEffectState);
+         ui.Hide();
```
Paul-Licameli, line 406: When the user deletes an effect from the stack, hide the dialog.

When the user replaces one effect with another in the stack, that too should hide the dialog.  But I don't see a corresponding change in `OnEffectListItemChange`.  That is the function reached through an Observer callback.

src/RealtimeEffectPanel.cpp
```diff
@@ -412,6 +438,8 @@ namespace
          }
 
          auto& effectStateUI = RealtimeEffectStateUI::Get(*mEffectState);
+
+         effectStateUI.UpdateTrackData(*mTrack);
```
Paul-Licameli, line 442: Does this line properly belong to the earlier commit "Adds a class to manage stacked effects settings UI" ?

crsib: I would say no

src/RealtimeEffectPanel.cpp
crsib: Will fix

src/RealtimeEffectPanel.h
crsib: RealtimeEffectPanel is a child of ProjectWindow

src/RealtimeEffectPanel.cpp
```diff
@@ -926,6 +954,35 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const
...
+         switch (evt.mType)
+         {
+         case TrackListEvent::TRACK_DATA_CHANGE:
+            if (mCurrentTrack.lock() == track)
+               mTrackTitle->SetLabel(track->GetName());
+            UpdateRealtimeEffectUIData(*waveTrack);
+            break;
+         case TrackListEvent::DELETION:
+            if (evt.mExtra == 0)
+               HideRealtimeUIForTrack(*waveTrack);
```
Paul-Licameli, line 974: Why is this not done in case of replacements (as indicated by evt.mExtra)?  I think I do understand that this can happen if you do a destructive effect application via other UI, or you finish recording, if I look for uses of `TrackList::Replace`.  The commit comment gave some explanation, and this code should too.

But then in the case of replacement, shouldn't there be another branch to call `UpdateRealtimeEffectUIData`?  The Track object is replaced with another one which is now the current state of the track in the project.

src/RealtimeEffectPanel.cpp
```diff
@@ -926,6 +954,35 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const
...
+            break;
+         case TrackListEvent::ADDITION:
+            // Addition can be fired as a part of "replace" event.
+            // Calling UpdateRealtimeEffectUIData is mostly no-op,
+            // it will just create a new State and Access for it.
+            UpdateRealtimeEffectUIData(*waveTrack);
```
Paul-Licameli, line 980: Aha, here it is then.  You are relying on the DELETION and ADDITION events both arriving, which is part of the contract in Track.h.  And you don't rely on the sequence of their arrival.

src/RealtimeEffectPanel.cpp
Paul-Licameli: Was it agreed then with designers that the formerly specified undo/redo behavior is now abandoned?

Do we open another issue for the future effort to figure out the problem?

If there will not be undo/redo states for the OK button, still I think something could be done:  the OK button should call `ModifyState` -- what also happens for the power button changes.  Then the change of settings becomes, not another state, but fused with whatever is the latest undo state.  This is similar to what is done when you simply change the selection, and makes for argubly less surprising behavior, if after OK-ing the dialog, you undo and redo again, expecting the settings to be unchanged.

Besides all that, there is also the reasonable expectation that if you cancel the dialog, tentative changes of settings should be cancelled and old state restored.  But nothing does that now.

src/RealtimeEffectPanel.h
Paul-Licameli: Right

src/RealtimeEffectPanel.cpp
Paul-Licameli: Not sure I understood why.

Paul-Licameli:
This completes a round of review of the lines of code changed, but now I need to reread the bullet points in the issue and consider whether they were all addressed.

src/RealtimeEffectPanel.cpp
crsib: >Was it agreed then with designers that the formerly specified undo/redo behavior is now abandoned

I've talked with @Tantacrul on that. Can I  see the previous specification please? AFAIK the dialog specified has no Ok/Cancel buttons whatsoever. 

>the OK button

There will be no OK button

crsib: Please see #3090 for the design. Stackable effects are really different from what happens on destructive path so there are different assumptions about the behavior.

Unfortunately we have a chicken-egg problem here: we cannot reasonably implement #3090  before non modality is achieved and without it there seems to be a confusion about the behavior of the buttons that will be removed from the dialog.

crsib:
Please consider the *expected* design, not the one currently implemented for the effect UI dialog.

src/RealtimeEffectPanel.cpp
Paul-Licameli: So then I looked at slide 8.0 (macOS examples) and that is so.

But is that an omission in the design?  How is the user supposed to close the dialog?  If it's only with Esc or the close button, then there is no way to distinguish between committing and cancelling the changes.

Paul-Licameli: I have commented at #3090

crsib: This matches the behavior for stacked effects in other DAWs though. In most of the cases I saw (including Garage Band) settings are always committed (on every change) and software provides no way to undo them. Ableton tracks each and every parameter change, but every change is committed immediately as well. 

image

Paul-Licameli: So maybe we have an opportunity to do things a bit better than those.

src/RealtimeEffectPanel.cpp
crsib: Turned out that this commit is not needed at all

src/RealtimeEffectPanel.cpp
crsib: Turned out that removing wasn't handled properly either

crsib:
I have fixed the issues found during the previous run

src/effects/RealtimeEffectList.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "RealtimeEffectList.h"
 #include "RealtimeEffectState.h"
+#include "RealtimeEffectStateUI.h"
```
Paul-Licameli, line 11: This `#include` is new since my last round of review, and unfortunately makes bad depenency cycles.

This is from a "before" picture made by scripts/graph.pl:
Screen Shot 2022-07-28 at 4 36 18 PM

Bad -- cycles crept in this release that we should fix later.  But let's not make them worse:


Screen Shot 2022-07-28 at 4 33 14 PM

Remove the one `#include` and this bubble still grows by one, including the new file RealtimeEffectStateUI.  I'm not sure which is the edge to cut so that this doesn't happen.  But at least this does not tie AudioUI and the other Realtime* files.  My unmerged library extractions in progress would move those files down into lib-audio-io.

And how to do the rest of the code change to break the dependency is easy to see: the two functions of `RealtimeEffectList` that you modify already to `Publish` so you can make subscriptions elsewhere to handle this.

Paul-Licameli (changes requested):
Avoid compilation dependency cycle

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,255 @@
...
+   wxWeakRef<EffectUIHost> mEffectUIHost;
+
+   TranslatableString mEffectName;
+   wxString mTrackName;
+};
+
+RealtimeEffectStateUI::RealtimeEffectStateUI()
+    : mSharedState(std::make_shared<RealtimeEffectStateUI::SharedState>())
+{
+}
```
Paul-Licameli, line 196: You followed one of the suggestions for changing the management of lifetimes of the shared state.  Is it all correct?

- generate noise, open sidebar, add any effect, but do NOT yet click the gear
- Do something undo-able
- undo
- click the gear
- redo
- click the gear again

This was a sequence that I feared might erroneously leave two non-modal dialogs opened connected to the same effect state.

This doesn't happen.  Yet, the reasons why it doesn't happen seem a little fragile to me.

It doesn't happen because `ProjectHistory::PopState` fires many track deletion, then addition, events in these lines:

```
   dstTracks.Clear();
   for (auto t : tracks->Any())
      dstTracks.Add(t->Duplicate());
```

And the Subscription you added calls `HideRealtimeUIForTrack` -- with the consequence, that ANY undo or redo makes the non-modal dialog disappear, I think? -- which might be considered unacceptable to QA.

Now, the "lazy" creation of `RealtimeEffectStateUI` almost always, whenever you cause a push or pop of the Undo stack -- can mean there is little point in avoiding doing it sooner, eagerly.  And the way to do that, as I suggested, is to have `RealtimeEffectState` call `mEffectStateExtensions.BuildAll()` in the constructor.  That would defeat my doubts.

Or it would also do so, if you do the other thing:  have `RealtimeEffectState` hold a `shared_ptr` to its `Extensions` and copy that pointer in the copy ctor.  Then `BuildAll` won't be needed.  And not needing `BuildAll` might mean economy in case `Extensions` ever needs to host some other attachment.

Paul-Licameli:
It's not true that edits that push the Undo stack make the non-modal dialog disappear, but it seems any Undo or Redo makes it disappear even if "the same" track with "the same" state still exists.  Is that acceptable to design and QA?  I don't know.  I won't block the merge for that, but I do want the dependency cycle problem addressed.

Paul-Licameli:
Break dependency of `ProjectWindow` on `RealtimeEffectPanel` (a dependency made earlier when the panel was introduced, not by you) and that cycle shrinks, not to one, but not worse than in 3.1.3.

How to do that?  It needs some manoeuvres with `GlobalHook` probably, that I might figure out in stabilization or for patch release.

So don't you worry about that one small addition to the cycle up in the UI layer.  But please do unblock the extraction of lib-audio-io.

src/effects/RealtimeEffectList.cpp
crsib: Shoot, my bad! Will be fixed for sure!

Paul-Licameli:
Some thoughts:

If mMainSettings is moved into the shared state, then there is no need for NonInterfering and the related templates to avoid false sharing, which simplifies the macOS build problem.

But I think it should not be moved into the shared state.  If you keep copying it as before, then it will be possible for a state's settings to be different at different points of Undo history.

Even if there are not distinct Undo/Redo items just for changes of settings, it may be an improvement to call ModifyState whenever the non-modal dialog closes.  This is done when you change the power button states.  Observe how it works.  Make some edit, and then toggle power.  Undo and redo.  What do you expect for power button state?  What you get is how you last left the button, before the undo; not the button as it was when the Undo state was first pushed on the stack.

ModifyState is also used for things like changes of the selected times.

Paul-Licameli:
That is separate from the lifetime problems that must be addressed to avoid the crashes in idle time for lv2 effects.

Paul-Licameli:
Maybe do the opposite, though:  Put worker settings into shared state.  Undo history has versions of main settings, which is copied by the copy constructor of the state, but doesn't need multiple worker settings just taking up space.  Then no need for `SharedNonInterfering` or `NonInterfering`.

Paul-Licameli:
Should `mID` and `mPlugin` remain in shared state?  Yes.  There is no need to reassign them after a state is made.  Replacement of one effect with another is implemented as construction of a new `RealtimeEffectState` with its own different, new shared state.

crsib:
I'm not sure if I like your idea on project history. As a user - this is not a behavior I would expect. I close a non modal dialog, press Ctrl-Z, and everything is reverted without my knowledge about that. Please, let's agree that changes to effect settings do not affect project history for now, we will return to it after 3.2.

crsib:
My proposal stands - I think that all of the state should be shared. I just need your confirmation, because you have designed this system and better understand possible problems that will occur due to this change

Paul-Licameli:
> I'm not sure if I like your idea on project history. As a user - this is not a behavior I would expect. I close a non modal dialog, press Ctrl-Z, and everything is reverted without my knowledge about that. Please, let's agree that changes to effect settings do not affect project history for now, we will return to it after 3.2.

This is a question to put to design, really not for either of us to decide unless the implementation difficulties are simply too great, but I think they would not be. @jessjwilliamson @LWinterberg @Tantacrul 

I did the power buttons deliberately as they are with Leo's agreement.

However the matter can be ignored for the purpose of this pull request and revisited as a separate issue.

Paul-Licameli:
> My proposal stands - I think that all of the state should be shared. I just need your confirmation, because you have designed this system and better understand possible problems that will occur due to this change

I missed where exactly you proposed that ALL state be shared -- but that would include worker settings, unlike in the version I reviewed.

Yes, I think that may be right -- excepting perhaps the main settings -- and could solve the assertion violation that happened  for lv2 idle.

So the undo/redo history will surely include the changes of composition of the list -- adding, removing, replacing, permuting the states.  But there would be no more weak pointer in the EffectSettingsAccess object that could expire because of ReplaceTrack.  Instead it could hold another shared_ptr to SharedState.

I'm going to review again all that exists in the present master version of RealtimeEffectState to be sure this would work.

Paul-Licameli:
If main and worker settings remain in the same structure, whether the shared or the unshared state -- then you should use `NonInterfering`.

crsib:
> power buttons 

Power buttons are different:

1. You revert your last action. This is not true for the rest of the settings. 
2. You see the result. In this case - the action is completely hidden from user. 

Consider that you change settings in you IDE and they are reverted when you press Ctrl-Z after closing the window.

The case that effect settings bypass the project history has been already discussed a lot. Please let's not spend more time on it.

Paul-Licameli:
> > power buttons
> 
> Power buttons are different:
> 
> 1. You revert your last action. This is not true for the rest of the settings.
> 2. You see the result. In this case - the action is completely hidden from user.
> 
> Consider that you change settings in you IDE and they are reverted when you press Ctrl-Z after closing the window.
> 
> The case that effect settings bypass the project history has been already discussed a lot. Please let's not spend more time on it.

I'm not convinced that the issue has been settled knowingly by designers.  There was a specification for undo/redo history strings from setting changes in the original design.  I think a *different* issue card could be opened for this an reconsidered in 3.3.

However the easy thing for now is not to address the issue (even if not forgetting the issue either), and we will agree to do the easiest thing for now, so that we do get the non-modality.

Paul-Licameli:
> > > power buttons
> > 
> > 
> > Power buttons are different:
> > 
> > 1. You revert your last action. This is not true for the rest of the settings.
> > 2. You see the result. In this case - the action is completely hidden from user.
> > 
> > Consider that you change settings in you IDE and they are reverted when you press Ctrl-Z after closing the window.
> > The case that effect settings bypass the project history has been already discussed a lot. Please let's not spend more time on it.
> 
> I'm not convinced that the issue has been settled knowingly by designers. There was a specification for undo/redo history strings from setting changes in the original design. I think a _different_ issue card could be opened for this an reconsidered in 3.3.
> 
> However the easy thing for now is not to address the issue (even if not forgetting the issue either), and we will agree to do the easiest thing for now, so that we do get the non-modality.

This is *not* the other question of the Esc key, which was knowingly settled in favor of doing nothing about it.

Tantacrul:
I would say it is likely that any reference to Undo/Redo in the design was not properly considered. Probably a throwaway line - although I can't find it myself.

src/EffectPlugin.h
```diff
@@ -84,6 +84,15 @@ class AUDACITY_DLL_API EffectPlugin
       std::shared_ptr<EffectInstance> &pInstance, EffectSettingsAccess &access,
       bool forceModal = false) = 0;
 
+   //! Returns the EffectUIClientInterface instance for this effect
+   /*!
+    * Usually returns self, but in case of NoiseReduction, it returns
```
Paul-Licameli, line 89: The comment is not correct -- NoiseReduction does not override the function.  And I don't think it needs to because it is not eligible for effect stacks.

src/effects/RealtimeEffectList.cpp
```diff
@@ -25,7 +26,7 @@ std::unique_ptr<ClientData::Cloneable<>> RealtimeEffectList::Clone() const
 {
    auto result = std::make_unique<RealtimeEffectList>();
    for (auto &pState : mStates)
-      result->mStates.push_back(RealtimeEffectState::make_shared(*pState));
+      result->mStates.push_back(pState);
```
Paul-Licameli, line 29: This change is new since I last reviewed, and I would describe this oppositely -- what was a deep copy is now shallow.

This has the consequence that the state is only transient state.  There is no longer any undoing or redoin of changes of settings, but only of the composition of the effect list.

src/effects/RealtimeEffectList.h
```diff
@@ -25,18 +25,19 @@ class RealtimeEffectState;
...
    enum class Type
    {
-      Insert,///<New effect item was added to the list at srcIndex position
-      Replace,///<Effect item was replaced with a new item at srcIndex position
-      Remove,///<Effect item was removed from the list at srcIndex position
-      Move ///<Item position has changed, from srcIndex to dstIndex
+      Insert,///<New effect item was added to the list at srcIndex position. affectedState is a new state
+      Replace,///<Effect item was replaced with a new item at srcIndex position. affectedState is an old state.
+      Remove,///<Effect item was removed from the list at srcIndex position. affectedState is removed state.
+      Move ///<Item position has changed, from srcIndex to dstIndex.  affectedState is the moved state
```
Paul-Licameli, line 35: This comment is not consistent with `RealtimeEffectList::MoveEffect` which leaves the pointer in the message null

src/effects/RealtimeEffectList.cpp
crsib: Yeah, you are right, my bad

Paul-Licameli (commented):
Unnecessary `#include`

src/effects/RealtimeEffectList.cpp
Paul-Licameli, line 11: The `#include` that made the bad dependency cycle is still here, but not needed to build now so easily deleted.

src/effects/RealtimeEffectState.h
```diff
@@ -28,14 +30,15 @@ class RealtimeEffectState
...
+   , public ClientData::Site<RealtimeEffectState>
 {
 public:
    struct AUDACITY_DLL_API EffectFactory : GlobalHook<EffectFactory,
       const EffectInstanceFactory *(const PluginID &)
    >{};
 
    explicit RealtimeEffectState(const PluginID & id);
-   RealtimeEffectState(const RealtimeEffectState &other);
+   RealtimeEffectState(const RealtimeEffectState &other) = delete;
```
Paul-Licameli, line 41: Deletion of the copy constructor might be done in the commit "RealtimeEffectList performs shallow copy for states"

src/RealtimeEffectPanel.cpp
```diff
@@ -987,6 +971,35 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const
...
+         auto track = evt.mpTrack.lock();
+         auto waveTrack = dynamic_cast<WaveTrack*>(track.get());
+
+         if (waveTrack == nullptr)
+            return;
+         
+         switch (evt.mType)
+         {
+         case TrackListEvent::TRACK_DATA_CHANGE:
+            if (mCurrentTrack.lock() == track)
```
Paul-Licameli, line 985: I think this may fail to fire when intended, in case `mCurrentTrack` expired because the track was replaced because some other edit was done to it that caused TrackList::Replace

src/RealtimeEffectPanel.cpp
```diff
@@ -987,6 +971,35 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const
...
+               mTrackTitle->SetLabel(track->GetName());
+            UpdateRealtimeEffectUIData(*waveTrack);
+            break;
+         case TrackListEvent::DELETION:
+            if (evt.mExtra == 0)
+               HideRealtimeUIForTrack(*waveTrack);
```
Paul-Licameli, line 991: The problems of dangling pointers to states or parts of states are surely fixed now, so no fear.

Maybe though, this event handler closes the dialog too often because it isn't discriminating about which track.

But to do that right might require a shared pointer to the interesting track, not weak.

And then there would be the problem of invalidating that shared pointer, and updating it to find the corresponding track in a different undo state.

Because I observe that Undo or Redo causes the dialog to disappear, because of this case of the switch.  Maybe that is acceptable for now.  Maybe testing will result in some request to change the design.

src/RealtimeEffectPanel.cpp
crsib: I have checked and mCurrentTrack is always consistent

src/RealtimeEffectPanel.cpp
crsib: I do not see your point. `mExtra == 0` means that the track is removed from the project as per contract. Please, provide an STR of a behavior you think is wrong.

crsib: Seems that undo is calling Clear, readding the duplicate back later

src/effects/RealtimeEffectList.cpp
```diff
@@ -90,7 +90,8 @@ RealtimeEffectList::AddState(std::shared_ptr<RealtimeEffectState> pState)
       Publisher<RealtimeEffectListMessage>::Publish({
          RealtimeEffectListMessage::Type::Insert,
          mStates.size() - 1,
-         { }
+         { },
+         pState
```
Paul-Licameli, line 94: I should have noticed this last round -- `pState` is moved-from therefore null.  Nothing in non-modality depends on this but the events should anyway behave as advertised.

src/effects/RealtimeEffectState.cpp
```diff
@@ -213,20 +213,16 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
...
 };
 
-RealtimeEffectState::RealtimeEffectState(const PluginID & id)
+RealtimeEffectState::RealtimeEffectState(const PluginID& id)
 {
    SetID(id);
+   BuildAll();
 }
 
-RealtimeEffectState::RealtimeEffectState(const RealtimeEffectState &other)
```
Paul-Licameli, line 221: The commit that removes this copy constructor should also remove the comment in the header file, `Members that are copied for undo and redo`

src/RealtimeEffectPanel.cpp
```diff
@@ -420,51 +396,23 @@ namespace
...
          if(mProject == nullptr || mEffectState == nullptr)
             return;//not initialized
-         
+
          const auto ID = mEffectState->GetID();
          const auto effectPlugin = EffectManager::Get().GetEffect(ID);
+
          if(effectPlugin == nullptr)
          {
             ///TODO: effect is not avaialble
```
Paul-Licameli, line 436: spelling

src/RealtimeEffectPanel.cpp
```diff
@@ -823,27 +774,33 @@ class RealtimeEffectListWindow : public wxScrolledWindow
...
 
-      if(track)
+      if (track)
       {
          auto& effects = RealtimeEffectList::Get(*mTrack);
-         mEffectListItemMovedSubscription = effects.Subscribe(*this, &RealtimeEffectListWindow::OnEffectListItemChange);
+         mEffectListItemMovedSubscription = effects.Subscribe(
+            *this, &RealtimeEffectListWindow::OnEffectListItemChange);
+
+         effects.Visit([track](auto& effect, bool) {
```
Paul-Licameli: The variable name `effect` might be less confusing as `state`

src/effects/RealtimeEffectState.h
```diff
@@ -28,14 +30,15 @@ class RealtimeEffectState
    : public XMLTagHandler
    , public std::enable_shared_from_this<RealtimeEffectState>
    , public SharedNonInterfering<RealtimeEffectState>
+   , public ClientData::Site<RealtimeEffectState>
```
Paul-Licameli, line 33: Wow, that's CRTP x 3 !

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+RealtimeEffectStateUI::~RealtimeEffectStateUI()
+{
+   Hide();
+}
+
+bool RealtimeEffectStateUI::IsShown() const noexcept
+{
+   return mEffectUIHost != nullptr;
```
Paul-Licameli, line 42: Can the non-modal dialog exist but be hidden?  Then this function name might mislead.

Paul-Licameli: But, no, I see.

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+void RealtimeEffectStateUI::Toggle(AudacityProject& project)
+{
+   if(IsShown())
+      Hide();
+   else
+      Show(project);
+}
+
+void RealtimeEffectStateUI::UpdateTrackData(const Track& track)
```
Paul-Licameli, line 119: We come here when rebuilding the side bar.

But can there be edits of the track name that don't get reflected in the dialog title?

A complete solution might be done with subscription to undo/redo events.  Which I suggest as a bug fix after this PR, not a blocking problem.

Paul-Licameli: To be REALLY complete... it would also subcribe also to changes of preferences, in case the effect name is a built-in that can be retranslated for changed of locale.

src/effects/RealtimeEffectStateUI.h
```diff
@@ -0,0 +1,57 @@
...
+
+  RealtimeEffectStateUI.h
+
+  Dmitry Vedenko
+
+**********************************************************************/
+#pragma once
+
+// wx/weakref.h misses this include
+#include <type_traits>
```
Paul-Licameli, line 14: Maybe that #include isn't needed

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,160 @@
...
+   if (effectPlugin == nullptr)
+      return;
+
+   EffectUIClientInterface* client = effectPlugin->GetEffectUIClientInterface();
+
+   // Effect has no client interface
+   if (client == nullptr)
+      return;
+
+   // mUIHost will acquire the correct instance
```
Paul-Licameli, line 66: copy-pasted comment refers to a nonexistent variable

Paul-Licameli: Comment is trying to explain why `release()` doesn't leak below.  It's because the dialog has a non null parent.

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+   // The dialog was modal? That shouldn't have happened
+   if (mEffectUIHost == nullptr || !mEffectUIHost->IsShown())
+   {
+      assert(false);
+      mEffectUIHost = {};
+   }
+
+   mProjectWindowDestroyedSubscription = projectWindow.Subscribe(
+      [this](ProjectWindowDestroyedMessage) { Hide(); });
```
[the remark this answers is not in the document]
Paul-Licameli: Did you intend to hide a non-modal dialog but not destroy it?

The closing of the project window does cause the dialog to be destroyed, because it sets its parent pointer to the ProjectWindow.

There are other non-modal dialogs associated with a project -- like Tools > Macros... -- and nothing special is needed to ensure that they are cleaned up.

src/RealtimeEffectPanel.cpp
Paul-Licameli: I have also

src/effects/RealtimeEffectStateUI.cpp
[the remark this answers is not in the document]
crsib: It fixes an assert in `EffectUIHost::~EffectUIHost()`. Is this assert important? You tell me.

src/effects/RealtimeEffectState.cpp
```diff
@@ -213,20 +213,16 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
    std::weak_ptr<RealtimeEffectState> mwState;
 };
 
-RealtimeEffectState::RealtimeEffectState(const PluginID & id)
+RealtimeEffectState::RealtimeEffectState(const PluginID& id)
 {
    SetID(id);
+   BuildAll();
```
Paul-Licameli, line 219: I think BuildAll is now unnecessary for correct lifetimes of `RealtimeEffectStateUI`, now that the state objects are shared across Undo history states.

Undo/Redo still closes the window.  If there is an enhancement to avoid that, remember that test case should be:
- Add an effect to a stack but don't open settings yet
- make some undoable edit
- undo
- click the gear
- redo (here, the sidebar must reconnect to another track)
- click the gear again
And the second click should hide the dialog, not open a second dialog.

Paul-Licameli (approved):
Approved, at last!  Today's comments include some suggestions about names and comments and a fix for TrackList events, but none of that should block this.

I will open some other issues about possible further bug fixes.

src/effects/RealtimeEffectStateUI.cpp
crsib: I had naïve expectation that TRACK_DATA_CHANGE will be fired, because, well, track data has changed. But you are right, it is not

crsib:
>I will open some other issues about possible further bug fixes.

Please don't

src/effects/RealtimeEffectList.cpp
crsib: Ironically, MSVC now has a diagnostic for that
image

src/effects/RealtimeEffectStateUI.cpp
crsib: Well, that means it's a bad comment anyway :-) It should be about `EffectInstance` lifetime

crsib: It is no longer needed on the other hand, the (weak) pointer to EffectInstance is no longer stored

src/effects/RealtimeEffectState.cpp
crsib: Let's keep it for now, probably, it does no harm either

Paul-Licameli: If any more ClientData are attached to the RealtimeEffectState, they too would be built non-lazily.

Well we don't care about that now.

src/effects/RealtimeEffectStateUI.cpp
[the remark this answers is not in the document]
Paul-Licameli: Okay, and maybe that old assertion is an inappropriate one.  But let it all stand.

src/RealtimeEffectPanel.cpp
```diff
@@ -1000,6 +1001,37 @@ RealtimeEffectPanel::RealtimeEffectPanel(
...
          }
    });
+
+   mUndoSubscription = UndoManager::Get(mProject).Subscribe(
+      [this](UndoRedoMessage message)
+      {
+         auto& trackList = TrackList::Get(mProject);
+
+         // Realtime effect UI is only updated on Undo or Redo
+         if (message.type == UndoRedoMessage::Type::UndoOrRedo)
```
Paul-Licameli, line 1030: Also for the `Reset`?  You can make that happe with the History dialog

src/RealtimeEffectPanel.cpp
```diff
@@ -1000,6 +1001,37 @@ RealtimeEffectPanel::RealtimeEffectPanel(
...
+            return;
+         
+         for (auto track : mPotentiallyRemovedTracks)
+         {
+            // By construction, track cannot be null
+            assert(track != nullptr);
+            auto matchingTrack = trackList.FindById(track->GetId());
+
+            if (matchingTrack == nullptr)
+               HideRealtimeUIForTrack(*track);
```
Paul-Licameli, line 1049: This requires that #3365 be merged first, so that the dialog is closed when it must be.

src/RealtimeEffectPanel.cpp
crsib: Will do

src/RealtimeEffectPanel.cpp
crsib: Yep, right now it is  kind of no-op

src/ProjectWindow.cpp
```diff
@@ -699,7 +699,7 @@ ProjectWindow::ProjectWindow(wxWindow * parent, wxWindowID id,
          {
             auto& project = GetProject();
             auto& trackFocus = TrackFocus::Get(project);
-            ShowEffectsPanel(project, trackFocus.Get());
+            ShowEffectsPanel(trackFocus.Get());
```
Paul-Licameli, line 702: This commit breaks the build but the next and final commit fixes it.

src/RealtimeEffectPanel.cpp
```diff
@@ -903,12 +904,30 @@ class RealtimeEffectListWindow : public wxScrolledWindow
    }
 };
 
+
+struct RealtimeEffectPanel::PrefsListenerHelper : PrefsListener
```
Paul-Licameli, line 908: This commit is insufficient to update langauges in (e.g.) the dialog for Reverb.  It only fixes the dialog title, not the static texts and label buttons.  Destroying and recreating the dialog may do it.  That is I think `RealtimeEffectStateUI::Hide` then `RealtimeEffectStateUI::Show`

src/RealtimeEffectPanel.cpp
Paul-Licameli: Correct that, it is rather, still closing the dialog unnecessarily.

We want the dialog to close exactly when it should.

Paul-Licameli: Now I'm trying to make it work by cherry-picking my fix onto this branch, and it doesn't, revealing an incompleteness of my proposed fix.  So watch for an amendment of #3365 !

Paul-Licameli: Okay, I ameneded #3365, which you should review first.

But still not enough:  if I cherry-pick it onto your branch, I must make one more change:

Your message handler must do nothing messages of type `Purge`.

Now I observe that undo and redo close the dialog only exactly as needed.

So, are we done yet, debugging undo/redo?

No, I think there is one more problem.  Wait for more.

Paul-Licameli:
Remaining problem (at least):

- generate noise
- add an effect to the stack
- Click the gear
- Undo

Expect:  the dialog should disappear, because the effect is no longer in the track's stack, even though the track still exists

Paul-Licameli:
Maybe this is a solution for all the  undo problems, which won't need TrackId after all:

- intercept the track deletions that happen when popping undo state, as you do now, and remember those tracks.
- Then the event comes from the Undo manager, after the current TrackList in the project is repopulated.
- Find all RealtimeEffectStateUI that are showing, and "reachable" from the old tracks.  Meaning, of course, you can visit the states in their RealtimeEffectLists and find them.
- Find all that are reachable from the new list of tracks.
- Close the UIs in the first set and not the second.

Then this is enough to hide UIs either in case Undo removes the track, or Undo leaves (some state of) the track in the project but the track no longer has the RealtimeEffectState any more.

crsib:
All proposed changes are implemented, @Paul-Licameli

src/RealtimeEffectPanel.cpp
```diff
@@ -987,9 +1004,113 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const
...
+               UpdateRealtimeEffectUIData(*waveTrack);
+         }
+
+         // But mPotentiallyRemovedTracks processing happens as fast as possible.
+         // This event is fired right after the track is deleted, so we do not
+         // hold the strong reference to the track much longer than need.
+         if (mPotentiallyRemovedTracks.empty())
+            return;
+
+         // Collect ReltimeEffectUIs that are currently shown
```
Paul-Licameli, line 1063: Spelling

src/RealtimeEffectPanel.cpp
```diff
@@ -41,6 +41,7 @@
 #include "effects/RealtimeEffectState.h"
 #include "effects/RealtimeEffectStateUI.h"
 #include "UndoManager.h"
+#include "Prefs.h"
```
Paul-Licameli, line 44: The last commit in the sequence is still correcting a build error of the previous commit.  I'd like both commits to be compilable.

src/RealtimeEffectPanel.cpp
```diff
@@ -1000,6 +1000,77 @@ RealtimeEffectPanel::RealtimeEffectPanel(
...
+            message.type == UndoRedoMessage::Type::BeginPurge ||
+            message.type == UndoRedoMessage::Type::EndPurge)
+            return;
+
+         auto& trackList = TrackList::Get(mProject);
+
+         // Realtime effect UI is only updated on Undo or Redo
+         auto waveTracks = trackList.Any<WaveTrack>();
+         
+         if (message.type == UndoRedoMessage::Type::UndoOrRedo)
```
Paul-Licameli, line 1051: You don't unpdate UI for the Reset message, which can happen if you choose a state using the History dialog, and so the title of a dialog might be incorrect in that case.  However that case will correctly fall through to the check for reachability of states below.

src/RealtimeEffectPanel.cpp
crsib: Well, rebase error again. Rebasing takes like half a time on this branch now

Paul-Licameli: I merged other approved things to master lately, so please also rebase onto those.

Paul-Licameli (approved):
Formally re-approved now.  Just make one more behavior change, the minor fix for the Reset message.

Issue #2982, opened 2022-05-25
User is told that AU and LADSPA plug-ins are "incompatible"


petersampsonaudacity:
**Describe the bug**
The Apple-provided AU plug-ins appear to fail to load on Mac (the user is told they are "incompatible") - you get the message shown in the screenshot.
This message is a regression on 3.1.3 - and is likely to worry Mac users

However if you go to the Plug-in Manager you can see all the AU effects there but they are disabled:
a) they can be enabled
b) then they can successfully be used

**To Reproduce**
1. Purge the Audacity settings folder
2. Launch 3.2.0
3. Observe: AU plug-ins fail to load (see screenshot)
4. Plug-in Manger
5. Observe: AU plug-ins marked as Disabled
6. Enable AU plug-ins
7. Use AU plug-ins
8. Observe: AU plugins available for use and they work

**Actual behavior**
AU plug-ins fail to load

**Expected behavior**
AU plug-ins should load

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/170320440-9f4dab09-90dc-4ee0-98cc-33eb7ad9a1b9.png)

**Additional information (please complete the following information):**
 - OS: macOS  Monterey 12.3.1
 - Version:  audacity-macOS-3.2.0-alpha-20220525+65a2ea0-x86_64 

**Additional context**
Note that in 3.1.3 and earlier the AU effects appear in the "New" category" and not the "Disabled" category.

Update from @Paul-Licameli : I have the swh package of LADSPA plug-ins built from source on my Mac, and those plug-ins have the same problems, and I predict they will also fail on other operating systems.  I test Audacity at commit 8a7fbf5cc9b50ed973b0eff167dd9402c63a0534.

There is a common bug for these two effect protocols, which, I predict, won't affect lv2, VST2, or VST3, though other unrelated bugs might also be found with those.

dozzzzer:
I confirm this on Mac Monterey 12.3.1

Paul-Licameli:
@pietro68 remind us please, where are the readilyt downloaded binaries for swh plug-ins for Windows?

Paul-Licameli:
`std::bad_function_call` is thrown in the PluginHost process, where `PluginManager::GetSettings` tries to call `sFactory`, but `PluginManager::Initialize` was not yet done.

[...]

PR #3177, opened 2022-07-04
Issue 2982: plugin validation fix

vsverchinsky:
Resolves: #2982

Fixes incorrect plugin registration when multiple providers associate themselves with same module path

Recommended:

### DIFF
--- a/src/PluginStartupRegistration.cpp
+++ b/src/PluginStartupRegistration.cpp
@@ -21,9 +21,10 @@
 #include "widgets/wxWidgetsWindowPlacement.h"
 
 
-PluginStartupRegistration::PluginStartupRegistration(std::vector<std::pair<wxString, wxString>> pluginsToProcess)
-   : mPluginsToProcess(std::move(pluginsToProcess))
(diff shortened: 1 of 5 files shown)


src/PluginStartupRegistration.cpp
```diff
@@ -34,20 +35,44 @@ void PluginStartupRegistration::OnInternalError(const wxString& error)
 
 void PluginStartupRegistration::OnPluginFound(const PluginDescriptor& desc)
 {
-   PluginManager::Get().RegisterPlugin(PluginDescriptor { desc });
-   if(!desc.IsValid())
```
Paul-Licameli, line 38: There is no longer a check of IsValid().  The condition for considering a plug-in failed is instead that all descriptors for the path are stubs.

I had to study the code again to be convinced that the condition for considering a plug-in failed, in case of only one provider, is equivalent to the old criterion.  But, I believe it now.

Plugin destriptors deserialized from the IPC pipe are stubs if and only if they are marked invalid.  Otherwise they were successfully registered in PluginManager in the other process and marked valid.

Paul-Licameli (approved):

Issue #3076, opened 2022-06-15
Effects menu entries


LWinterberg:
- [x] Rename "Add/Remove Plug-Ins" to "Plugin Manager". NB: This may need to be done for the Generate/Analyze/Tools menu, too.
- [x] Add new option "Add Realtime Effects" which opens the effects panel for the currently focussed track.

Paul-Licameli:
Note that the first task affects also the Generate, Analyze, and Tools menus too.
The second might affect only Effects, but there are also some plug-ins that are analyzers, not properly effects.  For instance: https://www.maat.digital/gon/

Paul-Licameli:
Question @LWinterberg :  Were you assuming the new menu items should end with "..." as the old ones did?  Just changing words?  After all it is the convention meaning that a dialog will open up.

But @ksoze95 took your specification literally and left off the "..."

LWinterberg:
Microsoft recommends: 

> This doesn't mean you should use an ellipsis whenever an action displays another window only when additional information is required to perform the action. For example, the commands About, Advanced, Help, Options, Properties, and Settings must display another window when clicked, but don't require additional information from the user. Therefore they don't need ellipses.

-- https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#using-ellipses

IMHO for "Add/Remove Plug-Ins..." the ellipsis was necessary as the action requires more input from the user, but for "Plugin Manager" it is not - clicking opens the manager.

ksoze95:
I jumped the gun and reintroduced the "..." 🤦  Will fix up my commit again.

LWinterberg:
Fwiw, I just noticed that this is inconsistent with some other menu items (like Preferences...) - I'll bring the other entries in like with this later.

petersampsonaudacity:
@LWinterberg for Consistency in Audacity the "plugin" should be hyphenated i.e. the menu entry should be "Plug-in Manager"

See Connie:  https://alphamanual.audacityteam.org/man/Consistency_in_wording_and_punctuation_in_Audacity

And note that "Plug-ins" is hyphenated in the Plug-in Manager dialog.   Perhaps the dialog should be re-titled to "Plug-in Manager"
![image](https://user-images.githubusercontent.com/50205984/178113186-c13c5e43-00e6-4b91-beeb-8c9c9861b28e.png)


Do you want to re-open this or should I raise a new issue for this.

LWinterberg:
[...]

PR #3178, opened 2022-07-05
Effects menu entries

ksoze95:
Resolves: #3076

Recommended:

### DIFF
--- a/src/menus/PluginMenus.cpp
+++ b/src/menus/PluginMenus.cpp
@@ -16,6 +16,7 @@
 #include "../ProjectSelectionManager.h"
 #include "../toolbars/ToolManager.h"
 #include "../Screenshot.h"
+#include "../TrackPanelAx.h"
 #include "TempDirectory.h"
 #include "UndoManager.h"
 #include "../commands/CommandContext.h"
@@ -435,6 +436,15 @@ void DoManagePluginsMenu(AudacityProject &project)
       MenuCreator::RebuildAllMenuBars();
 }
 
+
+void DoShowRealtimeEffectsSidePanel(AudacityProject &project)
+{
+   auto &trackFocus = TrackFocus::Get(project);
+   auto &projectWindow = ProjectWindow::Get(project);
+
+   projectWindow.ShowEffectsPanel(project, trackFocus.Get());
+}
+
 bool CompareEffectsByName(const PluginDescriptor *a, const PluginDescriptor *b)
 {
    return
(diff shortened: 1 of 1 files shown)


locale/af.po
```diff
@@ -13739,7 +13739,7 @@ msgid "&Generate"
 msgstr "&Genereer"
 
 #: src/menus/PluginMenus.cpp
-msgid "Add / Remove Plug-ins..."
+msgid "Plugin Manager"
```
vsverchinsky, line 13742: @Paul-Licameli is it necessary to update po files each time new msgid added or removed from the sources?

Paul-Licameli, line 13742: The second commit in this sequence is quite unnecessary!  Remove it.

As part of the release process, we will automatically update the .po files using the GNU gettext family of tools (msgfmt, etc.)

Paul-Licameli (changes requested):

src/menus/PluginMenus.cpp
```diff
@@ -1215,7 +1215,7 @@ BaseItemSharedPtr GenerateMenu()
    ( FinderScope{ findCommandHandler },
    Menu( wxT("Generate"), XXO("&Generate"),
       Section( "Manage",
-         Command( wxT("ManageGenerators"), XXO("Add / Remove Plug-ins..."),
+         Command( wxT("ManageGenerators"), XXO("Plugin Manager"),
```
Paul-Licameli, line 1234: The first commit is simple but I think "..." is not supposed to be omitted from the new texts.  See the conversation at the Issue page.

src/menus/PluginMenus.cpp
```diff
@@ -803,6 +813,12 @@ void OnManageEffects(const CommandContext &context)
    DoManagePluginsMenu(project);
 }
 
+void OnAddRealtimeEffects(const CommandContext& context)
+{
+   auto& project = context.project;
+   DoShowRealtimeEffectsSidePanel(project);
+}
```
Paul-Licameli, line 820: No objections to the third commit.

Paul-Licameli:
That's all!

locale/af.po
ksoze95: Commit dropped.

src/menus/PluginMenus.cpp
ksoze95: I have reintroduced the "..."

ksoze95: I jumped the gun.  I have now removed "..." due to @LWinterberg comment on the issue page.

Paul-Licameli (approved):

PR #3180, opened 2022-07-05
One more validator virtual

Paul-Licameli:
Some preliminaries for progress with lv2 refactoring,
making more widespread chang in other effect families.

Recommended:


crsib (approved):
This PR breaks #3176, I expect that there will be quite an amount of work to fix it after it is merged

PR #3183, opened 2022-07-06
Define settings vst2 - 2nd

pietro68:
Resolves: https://github.com/audacity/audacity/issues/2970

Because of difficulties to perform an interactive rebase on the first created branch, this new branch/PR was created.

Commit messages start with a number, which references the step# in the original issue they belong to. This was done because some of the steps (3 & 6) were too complex to have them done in one commit (also to make reviewing easier); but all commits starting with the same number can be easily squashed together later on.

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1201,18 +1201,17 @@ int VSTEffect::ShowClientInterface(
    return mDialog->ShowModal();
 }
 
-bool VSTEffect::SaveSettings(
-   const EffectSettings &, CommandParameters & parms) const
-{
-   for (int i = 0; i < mAEffect->numParams; i++)
-   {
-      wxString name = GetString(effGetParamName, i);
-      if (name.empty())
-      {
-         name.Printf(wxT("parm_%d"), i);
-      }
 
-      float value = callGetParameter(i);
+
+bool VSTEffect::SaveSettings(const EffectSettings& settings, CommandParameters& parms) const
+{
+   const VSTEffectSettings& vstSettings = GetSettings(settings);
+
+   for (const auto& item : vstSettings.mParamsMap)
+   {
+      const auto& name  = item.first;
+      const auto& value = item.second;
+
       if (!parms.Write(name, value))
       {
          return false;
@@ -1222,35 +1221,35 @@ bool VSTEffect::SaveSettings(
    return true;
 }
 
-bool VSTEffect::LoadSettings(
-   const CommandParameters & parms, EffectSettings &settings) const
+
+bool VSTEffect::LoadSettings(const CommandParameters& parms, EffectSettings& settings) const
 {
-   constCallDispatcher(effBeginSetProgram, 0, 0, NULL, 0.0);
-   for (int i = 0; i < mAEffect->numParams; i++)
+   VSTEffectSettings& vstSettings = GetSettings(settings);
+   vstSettings.mParamsMap.clear();
+
+   long index{};
+   wxString key;
+   float value = 0.0f;
+   if (parms.GetFirstEntry(key, index))
    {
-      wxString name = GetString(effGetParamName, i);
-      if (name.empty())
+      do
       {
-         name.Printf(wxT("parm_%d"), i);
-      }
+         if (parms.Read(key, value))
+            vstSettings.mParamsMap[key] = value;
+         else
+            return false;
 
-      double d = 0.0;
-      if (!parms.Read(name, &d))
-      {
-         return false;
-      }
-
-      if (d >= -1.0 && d <= 1.0)
-      {
-         const_cast<VSTEffect*>(this)->callSetParameter(i, d);
-         for (const auto &slave : mSlaves)
-            slave->callSetParameter(i, d);
-      }
+      } while (parms.GetNextEntry(key, index));
    }
-   constCallDispatcher(effEndSetProgram, 0, 0, NULL, 0.0);
+
+   vstSettings.mChunk     = std::nullopt;
+   vstSettings.mVersion   = VSTEffectWrapper::mVstVersion;
+   vstSettings.mUniqueID  = VSTEffectWrapper::mAEffect->uniqueID;
+   vstSettings.mNumParams = VSTEffectWrapper::mAEffect->numParams;
 
    return true;
 }
+
 
 bool VSTEffect::LoadUserPreset(
    const RegistryPath & name, EffectSettings &settings) const
@@ -1909,6 +1908,10 @@ bool VSTEffect::SaveParameters(
       mAEffect->version);
    SetConfig(*this, PluginSettings::Private, group, wxT("Elements"),
       mAEffect->numParams);
+
+
+   CommandParameters eap2;
+   SaveSettings(settings, eap2);
 
    if (mAEffect->flags & effFlagsProgramChunks)
    {
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -2159,9 +2159,9 @@ void VSTEffect::callSetParameter(int index, float value)
 }
 
 
-bool VSTEffectWrapper::callSetParameterB(int index, float value)
+bool VSTEffectWrapper::callSetParameterB(int index, float value) const
 {
-   if (mVstVersion == 0 || callDispatcher(effCanBeAutomated, 0, index, NULL, 0.0))
+   if (mVstVersion == 0 || constCallDispatcher(effCanBeAutomated, 0, index, NULL, 0.0))
    {
       mAEffect->setParameter(mAEffect, index, value);
       return true;
@@ -3462,6 +3462,95 @@ void VSTEffectWrapper::ForEachParameter(ParameterVisitor visitor) const
       if (!visitor(pi))
          break;
    }
+}
+
+
+bool VSTEffectWrapper::FetchSettings(VSTEffectSettings& vst3settings) const
+{
+   // Get the fallback ID-value parameters
+   ForEachParameter
+   (
+      [&](const ParameterInfo& pi)
+      {
+         float val = callGetParameter(pi.mID);
+         vst3settings.mParamsMap[pi.mName] = val;
+         return true;
+      }
+   );
+
+   // These are here to be checked against for compatibility later
+   vst3settings.mVersion   = mAEffect->version;
+   vst3settings.mUniqueID  = mAEffect->uniqueID;
+   vst3settings.mNumParams = mAEffect->numParams;
+
+   // Get the chunk (if supported)
+   vst3settings.mChunk = std::nullopt;
+   if (mAEffect->flags & effFlagsProgramChunks)
+   {
+      void* chunk = NULL;
+      int clen = (int)constCallDispatcher(effGetChunk, 1, 0, &chunk, 0.0);
+      if (clen > 0)
+      {
+         vst3settings.mChunk = Base64::Encode(chunk, clen);
+      }
+   }
+
+   return true;
+}
+
+
+bool VSTEffectWrapper::StoreSettings(const VSTEffectSettings& vst3settings)
+{
+   // First, make sure settings are compatibile with the plugin
+   if ((vst3settings.mUniqueID  != mAEffect->uniqueID)   ||
+       (vst3settings.mVersion   != mAEffect->version)    ||
+       (vst3settings.mNumParams != mAEffect->numParams)      )
+   {
+      return false;
+   }
+
+
+   // Try using the chunk first (if available)
+   if (vst3settings.mChunk)
+   {
+      ArrayOf<char> buf{ vst3settings.mChunk->length() / 4 * 3 };
+
+      int len = Base64::Decode(*vst3settings.mChunk, buf.get());
+      if (len)
+      {
+         VstPatchChunkInfo info = { 1, mAEffect->uniqueID, mAEffect->version, mAEffect->numParams, "" };
+
+         callSetChunkB(true, len, buf.get(), &info);
+         return true;
+      }         
+   }
+
+
+   // Chunk unavailable / decoding it did not work? fall back to param ID-value pairs
+
+   constCallDispatcher(effBeginSetProgram, 0, 0, NULL, 0.0);
+
+   ForEachParameter
+   (
+      [&](const ParameterInfo& pi)
+      {
+         const auto itr = vst3settings.mParamsMap.find(pi.mName);
+         if (itr != vst3settings.mParamsMap.end())
+         {
+            const float& value = itr->second;
+
+            if (value >= -1.0 && value <= 1.0)
+            {
+               callSetParameterB(pi.mID, value);
+            }
+         }
+         return true;
+      }
+   );
+   
+   constCallDispatcher(effEndSetProgram, 0, 0, NULL, 0.0);
+
+   return false;
 }
 
 
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -793,25 +793,7 @@ VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
 :  mPath(path),
    mMaster(master)
 {
-   mModule = NULL;
-   mAEffect = NULL;
-
    mTimer = std::make_unique<VSTEffectTimer>(this);
-   mTimerGuard = 0;
-
-   mInteractive = false;
-   mAudioIns = 0;
-   mAudioOuts = 0;
-   mMidiIns = 0;
-   mMidiOuts = 0;
-   mBlockSize = mUserBlockSize = 8192;
-   mBufferDelay = 0;
-   mProcessLevel = 1;         // in GUI thread
-   mHasPower = false;
-   mWantsIdle = false;
-   mWantsEditIdle = false;
-   mUseLatency = true;
-   mReady = false;
 
    memset(&mTimeInfo, 0, sizeof(mTimeInfo));
    mTimeInfo.samplePos = 0.0;
@@ -821,11 +803,6 @@ VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
    mTimeInfo.timeSigNumerator = 4;
    mTimeInfo.timeSigDenominator = 4;
    mTimeInfo.flags = kVstTempoValid | kVstNanosValid;
-
-   // UI
-   
-   mGui = false;
-   mContainer = NULL;
 
    // If we're a slave then go ahead a load immediately
    if (mMaster)
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -3014,7 +3014,7 @@ void VSTEffectWrapper::SaveFXB(const wxFileName & fn) const
    return;
 }
 
-void VSTEffect::SaveFXP(const wxFileName & fn) const
+void VSTEffectWrapper::SaveFXP(const wxFileName & fn) const
 {
    // Create/Open the file
    const wxString fullPath{ fn.GetFullPath() };
@@ -3025,7 +3025,7 @@ void VSTEffect::SaveFXP(const wxFileName & fn) const
          XO("Could not open file: \"%s\"").Format( fullPath ),
          XO("Error Saving VST Presets"),
          wxOK | wxCENTRE,
-         mParent);
+         nullptr);
       return;
    }
 
@@ -3042,7 +3042,7 @@ void VSTEffect::SaveFXP(const wxFileName & fn) const
          XO("Error writing to file: \"%s\"").Format( fullPath ),
          XO("Error Saving VST Presets"),
          wxOK | wxCENTRE,
-         mParent);
+         nullptr);
    }
 
    f.Close();
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -827,7 +827,7 @@ PluginPath VSTEffect::GetPath() const
 
 ComponentInterfaceSymbol VSTEffect::GetSymbol() const
 {
-   return mName;
+   return VSTEffectWrapper::GetSymbol();
 }
 
 VendorSymbol VSTEffect::GetVendor() const
@@ -2110,7 +2110,7 @@ wxString VSTEffectWrapper::GetString(int opcode, int index) const
    return str;
 }
 
-void VSTEffect::SetString(int opcode, const wxString & str, int index)
+void VSTEffectWrapper::SetString(int opcode, const wxString & str, int index)
 {
    char buf[256];
    strcpy(buf, str.Left(255).ToUTF8());
@@ -2149,14 +2149,27 @@ float VSTEffectWrapper::callGetParameter(int index) const
 
 void VSTEffect::callSetParameter(int index, float value)
 {
+   bool setOK = callSetParameterB(index, value);
+
+   if (setOK)
+   {
+      for (const auto& slave : mSlaves)
+         slave->callSetParameterB(index, value);
+   }
+}
+
+
+bool VSTEffectWrapper::callSetParameterB(int index, float value)
+{
    if (mVstVersion == 0 || callDispatcher(effCanBeAutomated, 0, index, NULL, 0.0))
    {
       mAEffect->setParameter(mAEffect, index, value);
-
-      for (const auto &slave : mSlaves)
-         slave->callSetParameter(index, value);
+      return true;
    }
+   return false;
 }
+
+
 
 void VSTEffect::callSetProgram(int index)
 {
@@ -2169,7 +2182,7 @@ void VSTEffect::callSetProgram(int index)
    callDispatcher(effEndSetProgram, 0, 0, NULL, 0.0);
 }
 
-void VSTEffect::callSetChunk(bool isPgm, int len, void *buf)
+void VSTEffectWrapper::callSetChunkB(bool isPgm, int len, void *buf)
 {
    VstPatchChunkInfo info;
 
@@ -2179,10 +2192,10 @@ void VSTEffect::callSetChunk(bool isPgm, int len, void *buf)
    info.pluginVersion = mAEffect->version;
    info.numElements = isPgm ? mAEffect->numParams : mAEffect->numPrograms;
 
-   callSetChunk(isPgm, len, buf, &info);
+   callSetChunkB(isPgm, len, buf, &info);
 }
 
-void VSTEffect::callSetChunk(bool isPgm, int len, void *buf, VstPatchChunkInfo *info)
+void VSTEffectWrapper::callSetChunkB(bool isPgm, int len, void *buf, VstPatchChunkInfo *info)
 {
    if (isPgm)
    {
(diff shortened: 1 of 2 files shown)


src/effects/VST/VSTEffect.h
```diff
@@ -73,6 +76,136 @@ typedef SInt16 CFBundleRefNum;
...
+   std::optional<wxString> mChunk;
+
+   // Fallback data used when the chunk is not available.
+   std::unordered_map<wxString, double> mParamsMap;
+};
+
+
+struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
+{
+   AEffect* mAEffect = nullptr;
```
Paul-Licameli, line 104: I notice there is an assignment of this member in the constructor of VSTEffect that is no longer necessary.

I suggest, you make a new preliminary commit, and rebase your branch onto that:

First rewrite the constructor of VSTEffect to eliminate as many assignments to members, as possible.  Instead, use in-class initializations where possible, otherwise member initializers in the constructor.

src/effects/VST/VSTEffect.cpp
```diff
@@ -2187,6 +2262,17 @@ void VSTEffect::callSetParameter(int index, float value)
    }
 }
 
+
+void VSTEffectWrapper::callSetParameterB(int index, float value)
```
Paul-Licameli, line 2266: You wrote the functions to set chunks so that you avoided duplications:  the old functions invoke the `B` functions for common steps.

Do likewise for setting the parameter.  Make the return from this function boolean if necessary to enable that rewrite.

src/effects/VST/VSTEffect.h
```diff
@@ -73,6 +76,136 @@ typedef SInt16 CFBundleRefNum;
...
+   static inline VST3EffectSettings& GetSettings(EffectSettings& settings)
+   {
+      auto pSettings = settings.cast<VST3EffectSettings>();
+      assert(pSettings);
+      return *pSettings;
+   }
+   */
+
+   // These are here because they are used by the import/export methods
+   wxWindow* mParent;
```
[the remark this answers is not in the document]
pietro68: ok, done

src/effects/VST/VSTEffect.h
```diff
@@ -73,6 +76,136 @@ typedef SInt16 CFBundleRefNum;
...
+   bool HandleXMLTag(const std::string_view& tag, const AttributesList& attrs) override;
+   void HandleXMLEndTag(const std::string_view& tag) override;
+   void HandleXMLContent(const std::string_view& content) override;
+   XMLTagHandler* HandleXMLChild(const std::string_view& tag) override;
+
+   void SetString(int opcode, const wxString& str, int index = 0);
+
+   ComponentInterfaceSymbol GetSymbolB() const
+   {
+      return XO("VST Effects");
```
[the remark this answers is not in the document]
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
pietro68: I do not understand: I already wrote ::callSetParameterB. Do you mean, I should have it return a boolean and check the value where called?

The reason to write the "B" versions of callSetChunk, callSetParameter, callSetProgram was this: they are needed by other methods in the Wrapper, but the original versions also loop on mSlaves, and I guess the Wrapper should not be aware of them. The "B" versions are basically copies of the originals, with the loop on mSlaves removed.

src/effects/VST/VSTEffect.h
```diff
@@ -342,24 +342,24 @@ class VSTEffect final
...
 
    VstTimeInfo mTimeInfo;
 
-   bool mUseLatency;
-   int mBufferDelay;
+   bool mUseLatency{true};
+   int mBufferDelay{0};
 
-   unsigned mBlockSize;
+   unsigned mBlockSize{8192};
```
Paul-Licameli, line 467: This first commit is good.

Do you know that you could even write
```
unsigned mBlockSize{ mUserBlockSize };
```

The in-class initializer doesn't have to be a constant expression, and it can refer to earlier declared members.  That would more exactly preserve the mening of the assignment

```
mBlockSize = mUserBlockSize = 8192;
```

src/effects/VST/VSTEffect.h
```diff
@@ -101,6 +105,33 @@ struct VSTEffectWrapper : public VSTEffectLink
...
+   bool LoadXML(const wxFileName& fn);
+   bool HandleXMLTag(const std::string_view& tag, const AttributesList& attrs) override;
+   void HandleXMLEndTag(const std::string_view& tag) override;
+   void HandleXMLContent(const std::string_view& content) override;
+   XMLTagHandler* HandleXMLChild(const std::string_view& tag) override;
+
+   void SetString(int opcode, const wxString& str, int index = 0);
+
+   ComponentInterfaceSymbol GetSymbolB() const
+   {
```
Paul-Licameli, line 189: Now that mName is in the wrapper, I suggest:
- make this function non-inline
- simply call it `GetSymbol`
- Redefine `VSTEffect::GetSymbol` (a virtual function override) as the call-through to `VSTEffectWrapper::GetSymbol`

src/effects/VST/VSTEffect.cpp
```diff
@@ -3463,4 +3463,94 @@ void VSTEffectWrapper::ForEachParameter(ParameterVisitor visitor) const
...
+         callSetChunkB(true, len, buf.get(), &info);
+         return true;
+      }         
+   }
+
+
+   // Chunk unavailable / decoding it did not work? fall back to param ID-value pairs
+   const auto vstVersion = constCallDispatcher(effGetVstVersion, 0, 0, NULL, 0);
+
+   ForEachParameter
```
Paul-Licameli, line 3605: This visit of parameters has these differences from `VSTEffect::LoadSettings`:
- The bracketing dispatches of `effBeginSetProgram` and `effEndSetProgram` are not done
- The bounds check on `value` between +/- 1 are not done

Is it important to preserve the bounds check?  I honestly don't know.  Preserve it, as real code or as commented out, and experiment may later decide if we really need it.

`LoadSettings` as I now see, and as you probably saw too, was doing `callSetParameter` on self and then again on slaves:  which was unnecessary, because the `callSetParameter` call on self was already recursive.

The begin and end of set program were only done on master.  Was it right or wrong not to do it to on slaves?  I don't know.  But let's assume, it can be done on every handle for every instance, and the calls should therefore be part `StoreSettings.`

In practice, multiple slaves will happen only when you use the old non-modal dialog from the Effects menu, which will eventually not be part of 3.2 anyway.  But we may in future want to reimplement something like it for a master bus of effects.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3463,4 +3463,94 @@ void VSTEffectWrapper::ForEachParameter(ParameterVisitor visitor) const
 }
 
 
+bool VSTEffectWrapper::FetchSettings(VSTEffectSettings& vst3settings) const
```
Paul-Licameli, line 3541: Comment in the commit that `FetchSettings` adapts `SaveSettings` and `SaveParameters` but storing in-memory in EffectSettings, not in a config file.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3551,6 +3551,15 @@ bool VSTEffectWrapper::StoreSettings(const VSTEffectSettings& vst3settings) cons
    return false;
 }
 
+bool VSTEffect::TransferDataToWindow(const EffectSettings& settings)
+{
+   if (!StoreSettings(GetSettings(settings)))
+      return false;
+
+   // TODO (?): update of the GUI controls
```
Paul-Licameli, line 3634: The commit that introduces this function should also call `RefreshParameters()` which is a step done in the existing functions that load user and factory presets and that import.

Maybe you did this in a later commit I haven't reviewed yet?  But you can do it in this early commit to make the function complete.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3561,5 +3561,16 @@ bool VSTEffect::TransferDataToWindow(const EffectSettings& settings)
...
 
+EffectSettings VSTEffect::MakeSettings() const
+{
+   auto result = StatefulPerTrackEffect::MakeSettings();
+   // Cause initial population of the map stored in the stateful effect
+   if (!mInitialFetchDone) {
+      FetchSettings(GetSettings(result));
+      mInitialFetchDone = true;
+   }
+   return result;
```
Paul-Licameli, line 3647: This `MakeSettings` is good.  To be very fussy about the correct logical sequence of commits, I would say do this one before defining `TransferDataToWindow`, just so that the lines

```
   if (!StoreSettings(GetSettings(settings)))
      return false;
```

... are sure NOT to return false because `mSettings` has not yet been initialized.

src/effects/VST/VSTEffect.cpp
Paul-Licameli: Now I have looked at the whole PR and I see this wasn't done.  The task breakdown said

"Define TransferDataToWindow (to become UpdateUI of a validator later), calling StoreSettings; remove parts of LoadSettings that duplicate this"

So, to make that step complete, add the `RefreshParameters()` call and also remove the call in three places, all in the same commit.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1411,6 +1411,8 @@ bool VSTEffect::ValidateUI(EffectSettings &settings)
    if (GetType() == EffectTypeGenerate)
       settings.extra.SetDuration(mDuration->GetValue());
 
+   FetchSettings(GetSettings(settings));
```
Paul-Licameli, line 1414: The commit that inserts calls to `FetchSettings` should also do it in `LoadFactoryPreset`

src/effects/VST/VSTEffect.cpp
```diff
@@ -1258,6 +1260,49 @@ bool VSTEffect::LoadSettings(
...
+   {
+      do
+      {
+         if (parms.Read(key, value))
+            vstSettings.mParamsMap[key] = value;
+         else
+            return false;
+
+      } while (parms.GetNextEntry(key, index));
+   }
```
Paul-Licameli, line 1300: This new definition of `LoadSettings` should set up all members of `vstSettings` from config file information, not just the parameter map.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1477,6 +1477,8 @@ void VSTEffect::ExportPresets(const EffectSettings &) const
       return;
    }
 
+   StoreSettings(GetSettings(settings));
```
Paul-Licameli, line 1480: In the last commit, which adds two calls to `StoreSettings`, check the return value and return early if it is false.  In `SaveParameters`, do that before any other steps changing the config file.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1979,17 +1979,21 @@ bool VSTEffect::LoadParameters(
...
-   SetConfig(*this, PluginSettings::Private, group, wxT("Version"),
-      mAEffect->version);
-   SetConfig(*this, PluginSettings::Private, group, wxT("Elements"),
-      mAEffect->numParams);
+   const auto& vstSettings = GetSettings(settings);
 
-   if (mAEffect->flags & effFlagsProgramChunks)
+   SetConfig(*this, PluginSettings::Private, group, wxT("UniqueID"), vstSettings.mUniqueID );
+   SetConfig(*this, PluginSettings::Private, group, wxT("Version"),  vstSettings.mVersion  );
+   SetConfig(*this, PluginSettings::Private, group, wxT("Elements"), vstSettings.mNumParams);
```
Paul-Licameli, line 1996: In the commit that adds these lines, "Rewrite SaveParameters not using the handles" -- you store values from vstSettings and not from mAeffect, which is what I meant.  Hiding all mention of mAeffect in this function isn't what I meant.  It's not wrong to factor out `SupportsChunk`, but not what I meant.

But the last commit in the sequence (after this) should also, as I said, do `StoreSettings` first, anyway, which includes the check that the values in the two agree.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1911,22 +1911,35 @@ std::vector<int> VSTEffect::GetEffectIDs()
...
+   VstPatchChunkInfo info = { 1, mAEffect->uniqueID, mAEffect->version, mAEffect->numParams, "" };
+   return info;
+}
+
+bool VSTEffectWrapper::IsCompatible(const VstPatchChunkInfo& info) const
+{
+   return  (info.pluginUniqueID == mAEffect->uniqueID) &&
+           (info.pluginVersion  == mAEffect->version) &&
+           (info.numElements    == mAEffect->numParams);
+}
```
Paul-Licameli, line 1928: The commit adding these two functions, "Rewrite LoadParameters not using the handles", is just hiding mention of mAeffect in LoadParameters while not removing dependency on its state.  It's not what I meant.

However, I must read and understand all that you commented in "Rewritten LoadSettings / SaveSettings".

Actually eliminating use of mAEffect might not be possible, as I think you are saying:  instead, this handle is really needed as a "scratch" as with AudioUnits.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1911,22 +1911,35 @@ std::vector<int> VSTEffect::GetEffectIDs()
...
+
 bool VSTEffect::LoadParameters(
    const RegistryPath & group, EffectSettings &settings)
 {
    wxString value;
 
-   VstPatchChunkInfo info = {1, mAEffect->uniqueID, mAEffect->version, mAEffect->numParams, ""};
+   auto info = GetChunkInfo();
+
```
Paul-Licameli, line 1936: The commit that rewrites this function can also make it and `StoreCommandParameters` const, and that will allow you to remove `DoLoadUserPreset` and a `const_cast`

src/effects/VST/VSTEffect.cpp
```diff
@@ -1207,8 +1207,9 @@ int VSTEffect::ShowClientInterface(
    return mDialog->ShowModal();
 }
 
-bool VSTEffect::SaveSettings(
-   const EffectSettings &, CommandParameters & parms) const
+
+// This used to be ::SaveSettings
```
Paul-Licameli, line 1211: Okay, commit 6.2, the most difficult one:

I'm not convinced by what you write in the commit comment, about the need to keep the old functions.  And, the new LoadSettings/SaveSettings are not using the same "parm_%d" convention to name the keys (sometimes), which means it might break compatibility with old config files.

`StoreCommandParameters` is used only by `LoadParameters`.  If it takes the "chunk" alternative, then it is using the mAEffect handle, as a "scratch" from which it then fetches settings.  That's ok.

If that function does not take the early exit path loaing a chunk -- then instead a string value for key "Parameters" is loaded, and reinterpreted as a config file of key-value pairs.  I don't see why this could not be given to the new `LoadSettings` directly.  In which case, you do not rely on a "scratch" handle and so should NOT use `FetchSettings`.

`FetchCommandParameters` is used only by `SaveParameters`.  That function first calls `StoreSettings` to change the scratch handle state, and then maybe exits early if chunks are supported.  But if they are not, then again, I don't understand why the direct translation of settings to key-value pairs in `eap` can't be done.

Paul-Licameli:
That's enough for this round of review!

src/effects/VST/VSTEffect.h
pietro68: This would look better - done, although I had to use a static cast:

`unsigned mBlockSize{ static_cast<unsigned>(mUserBlockSize) };`

src/effects/VST/VSTEffect.h
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
```diff
@@ -3463,4 +3463,94 @@ void VSTEffectWrapper::ForEachParameter(ParameterVisitor visitor) const
...
+         VstPatchChunkInfo info = { 1, mAEffect->uniqueID, mAEffect->version, mAEffect->numParams, "" };
+
+         callSetChunkB(true, len, buf.get(), &info);
+         return true;
+      }         
+   }
+
+
+   // Chunk unavailable / decoding it did not work? fall back to param ID-value pairs
+   const auto vstVersion = constCallDispatcher(effGetVstVersion, 0, 0, NULL, 0);
```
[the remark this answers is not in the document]
pietro68: Alright, done. And yes, probably a good part if not all of ::Load will need to move to the wrapper.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3463,4 +3463,94 @@ void VSTEffectWrapper::ForEachParameter(ParameterVisitor visitor) const
...
+         vst3settings.mChunk = Base64::Encode(chunk, clen);
+      }
+   }
+
+   return true;
+}
+
+
+bool VSTEffectWrapper::StoreSettings(const VSTEffectSettings& vst3settings)
+{
```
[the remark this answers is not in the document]
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
pietro68: You are right, I forgot this. Now done.

src/effects/VST/VSTEffect.cpp
pietro68: Good point, done

src/effects/VST/VSTEffect.cpp
pietro68: Good catch, now done

src/effects/VST/VSTEffect.cpp
pietro68: This is what I added now, I hope it is what you meant?

 ```
   vstSettings.mChunk     = std::nullopt;
   vstSettings.mVersion   = VSTEffectWrapper::mVstVersion;
   vstSettings.mUniqueID  = VSTEffectWrapper::mAEffect->uniqueID;
   vstSettings.mNumParams = VSTEffectWrapper::mAEffect->numParams;
```

File: src/effects/VST/VSTEffect.cpp (hunk 13 above)
Comment (pietro68) [reply src/effects/VST/VSTEffect.cpp:1480]: That's right, done.

File: src/effects/VST/VSTEffect.cpp (hunk 7 above)
Comment (pietro68) [reply src/effects/VST/VSTEffect.cpp:3605]: I added the bounds checking and the  `effBeginSetProgram` / `effEndSetProgram` bracketing. I need a plugin which does not support chunks to test this, but since LoadSettings did this, it looks ok if StoreSettings also does it.

File: src/effects/VST/VSTEffect.h (hunk 5 above)
Comment (Paul-Licameli) [reply src/effects/VST/VSTEffect.h:467]: Maybe mBlockSize should just be size_t.

Or you can use the = syntax not {} .  There is a difference.  {} disallows narrowing conversions.

File: src/effects/VST/VSTEffect.cpp (hunk 12 above)
Comment (Paul-Licameli) [reply src/effects/VST/VSTEffect.cpp:1300]: Yes

File: src/effects/VST/VSTEffect.cpp (hunk 14 above)
Comment (pietro68) [reply src/effects/VST/VSTEffect.cpp:1996]: If it is allowed to reference mAEffect for that case, then I am happy to revert `SupportsChunk`

File: src/effects/VST/VSTEffect.cpp (hunk 16 above)
Comment (pietro68) [reply src/effects/VST/VSTEffect.cpp:1936]: ok, done

File: src/effects/VST/VSTEffect.cpp (hunk 17 above)
Comment (pietro68) [reply src/effects/VST/VSTEffect.cpp:1211]: I interpret your comment like so: in `LoadParameters`, instead of calling `StoreCommandParameters` and then `FetchSettings`, call `LoadSettings` instead. Then there would be no need for StoreCommandParameters to exist.

But the thing is: the new LoadSettings only translates to settings the CommandParameters that it is given, without writing them to the handle. I thought this was the semantics of LoadSettings we agreed on - is it not? it should be only a translator, not a translator + a setter. 

Because I found the existing LoadSettings not to conform to these agreed semantics (it was also writing to the handle), I renamed it to StoreCommandParameters and wrote a new LoadSettings which does the translation only.

But I could also revert it all and just have the original LoadSettings which translates and sets to the handle.  No problem for me, the code would also be simpler. I would only have the concern that the semantics of LoadSettings vary across effect families.

File: src/effects/VST/VSTEffect.h (hunk 5 above)
Comment (pietro68) [reply src/effects/VST/VSTEffect.h]: I opted to make mBlockSize a size_t

File: src/effects/VST/VSTEffect.cpp (hunk 15 above)
Comment (pietro68) [reply src/effects/VST/VSTEffect.cpp:1933]: Let's try to see it from this angle: what LoadParameters has to do, is to read some stuff from the config, set it to a plugin handle (done in the call to callSetChunkB) and fetch the settings from the plugin handle. This handle is mAEffect, a member of the wrapper. Methods of VSTEffect do not operate on mAEffect directly, but through methods of VSTEffectWrapper. Is this wrong?

File: src/effects/VST/VSTEffect.cpp [hunk 20]
```diff
@@ -928,10 +905,10 @@ bool VSTEffect::IsDefault() const
 auto VSTEffect::RealtimeSupport() const -> RealtimeSince
 {
    // TODO reenable after achieving statelessness
-   return RealtimeSince::Never;
-//   return GetType() == EffectTypeProcess
-//      ? RealtimeSince::Always
-//      : RealtimeSince::Never;
+   //return RealtimeSince::Never;
```
Paul-Licameli, line 908: Just remove that commented line

Paul-Licameli:
Still incomplete in the commit that defines `TransferDataToWindow`: remove `RefreshParameters` from `LoadUserPreset` `DoLoadFactoryPreset` and `ImportPresets`

Paul-Licameli:
> Let's try to see it from this angle: what `LoadParameters` has to do, is to read some stuff from the config, set it to a plugin handle (done in the call to `callSetChunkB`) and fetch the settings from the plugin handle. This handle is `mAEffect`, a member of the wrapper. Methods of `VSTEffect` do not operate on `mAEffect` directly, but through methods of `VSTEffectWrapper`. Is this wrong?

Not wrong, but not necessary,  just to hide mention of the member mAEffect inside the function VSTEffect::LoadParameters.

Right, and necessary, to use mAEffect somewhere if a "chunk" must be reinterpreted as a settings object.  It's then a scratch handle.

But, I think that if LoadParameters takes the other path that does not use the chunks, then it is not necessary to use a scratch handle, not necessary to define `StoreCommandParameters` or then to use `FetchSettings`.  On that path, a direct conversion can be done from strings stored in the config file into values in the map in EffectSettings.

Paul-Licameli:
> I interpret your comment like so: in `LoadParameters`, instead of calling `StoreCommandParameters` and then `FetchSettings`, just call `LoadSettings`. Then there would be no need for `StoreCommandParameters` to exist.
> 
> But the thing is: the new `LoadSettings` only translates to settings the `CommandParameters` that it is given, without writing them to the handle. I thought this was the semantics of `LoadSettings` we agreed on - is it not? it should be only a translator, not a translator + a setter.
> 

It should be a translator, but sometimes needs a scratch handle, as with the paht that uses chunks.  But not the rest.

You wrote `StoreCommandParameters` which uses a handle but unnecessarily.  Don't add the function, but just adapt LoadSettings to interpret `parms` and update a map.

> Because I found the existing `LoadSettings` not to conform to these agreed semantics (it was also writing to the handle), I renamed it to `StoreCommandParameters` and wrote a new `LoadSettings` which does the translation only.
> 

There is no need to retain `StoreCommandParameters`.  Change `LoadSettings` to avoid use of a handle and iterate over the config file.  Well you did that already.  But also, do not change the convention previously used to name the keys associated with the values.

> But I could also revert it all, i.e. just have the original `LoadSettings` which translates and sets to the handle, and then I could use it as you suggest. No problem for me, the code would also be simpler. I would only have the concern that the semantics of `LoadSettings` would then vary across effect families.

Maybe what you misunderstand is that `LoadParameters`, which is used by `LoadUserPreset`, is also intended as a translator only.  Side effect on a scratch handle for the interpretation of chunks is unavoidable as part of interpretation.  But on other paths, it is not important to keep whatever side effects used to happen.

src/effects/VST/VSTEffect.cpp
pietro68: done

pietro68:
> Still incomplete in the commit that defines `TransferDataToWindow`: remove `RefreshParameters` from `LoadUserPreset` `DoLoadFactoryPreset` and `ImportPresets`

right, done

pietro68:
> Maybe what you misunderstand is that `LoadParameters`, which is used by `LoadUserPreset`, is also intended as a translator only. 

I surely misunderstood that! I have now fixed things in this way.

pietro68:
Regarding the incompatibility with keys like "parm_%d": 
that kind of key is used when for some reason the handle will not be able to return proper parameter names (e.g. "gain").
I fixed this by having this code 

```
wxString name = GetString(effGetParamName, i);
if (name.empty())
{
   name.Printf(wxT("parm_%d"), i);
}
```
both in ::ForEachParameter and in ::LoadSettings.

Paul-Licameli (changes requested):
I see the latest round of changes.

One thing I'm still unsatisfied with:  The commit "6.2 Rewritten LoadSettings / SaveSettings..." needs an update of the commit comment, and `SaveSettings` and `LoadSettings` should remain as in the previous iterations of the branch:  Don't iterate mAEffect, don't do the side effects of `constCallDispatcher` or `callSetParameter`.  Do:  iterate over the map in `const EffectSettings` or over the key/value pairs in `const CommandParameters`.

Paul-Licameli:
Commit 7c469cdd604c3b44a6b8252ce8562da7d9689519 had almost acceptable versions of `SaveSettings` and `LoadSettings` -- except for the problem of preserving the formatting of the keys.  `SaveSettings` from that commit can now be used verbatim, because the change in `ForEachParameter` corrected the keys.

pietro68:
> Commit [7c469cd](https://github.com/audacity/audacity/commit/7c469cdd604c3b44a6b8252ce8562da7d9689519) had almost acceptable versions of `SaveSettings` and `LoadSettings` -- except for the problem of preserving the formatting of the keys. `SaveSettings` from that commit can now be used verbatim, because the change in `ForEachParameter` corrected the keys.

I modified commit 6.2 in the way I think you specified - basically SaveSettings and LoadSettings are the same as in the commit you pointed out, belonging to the previous iteration. The switching of the key to the parm_%d in case the handle can not provide alphabetic names for keys is now done exclusively in ForEachParameter - I guess this is right?

Paul-Licameli:
> > Commit [7c469cd](https://github.com/audacity/audacity/commit/7c469cdd604c3b44a6b8252ce8562da7d9689519) had almost acceptable versions of `SaveSettings` and `LoadSettings` -- except for the problem of preserving the formatting of the keys. `SaveSettings` from that commit can now be used verbatim, because the change in `ForEachParameter` corrected the keys.
> 
> I modified commit 6.2 in the way I think you specified - basically `SaveSettings` and `LoadSettings` are the same as in the commit you pointed out, belonging to the previous iteration. The switching of the key to the parm_%d in case the handle can not provide alphabetic names for keys is now done exclusively in `ForEachParameter` - I guess this is right?
> 
> btw, you said:
> 
> > don't do the side effects of `constCallDispatcher` or `callSetParameter`
> 
> But in that commit, neither LoadSettings or SaveSettings were calling them? or maybe I am missing something

True, the previous version did not do that.

Issue #2987, opened 2022-05-26
Effect dialog accessed from main Effects menu always modal and destructive only


Paul-Licameli:
As agreed with design team:  the old user interface for accessing realtime effects -- as a project-wide not per-track list, and limited to one effect only -- will be suppressed in favor of the new sidebar.

Old behavior was that the effects dialog is modal and can only apply effects destructively (though there is also the Preview button to calculate results and let you listen first before applying) -- but with exceptions for realtime-capable effects, in which case, the dialog is non-modal and does not exclude other editing of the project.

New behavior will make all effect dialogs modal when accessed this way.

Renabling a master effects list is future enhancement.

Must make sure:
- [ ] realtime capable effects, accessed from menus, are disabled during playback like other effects

dozzzzer:
During the playback, many plugins of different families appear to be available, though they cannot be selected of course. This looks pretty confusing and needs to be fixed.

<img width="400" alt="image" src="https://user-images.githubusercontent.com/70063970/186372015-2d70a985-c47c-4328-89c3-ff067a9129a9.png">
<img width="400" alt="image" src="https://user-images.githubusercontent.com/70063970/186372105-a5a07f3f-5b1c-4370-99bc-6f2ad4bf8ab5.png">
<img width="400" alt="image" src="https://user-images.githubusercontent.com/70063970/186372215-df9a2d23-745a-46a8-9817-d7d4a736074c.png">

PR #3187, opened 2022-07-06
Effect settings dialog is always modal for destructive workflows

crsib:
Resolves: #2987

Recommended:

### DIFF
--- a/src/effects/EffectBase.cpp
+++ b/src/effects/EffectBase.cpp
@@ -166,7 +166,7 @@ bool EffectBase::DoEffect(EffectSettings &settings, double projectRate,
    if ( pParent && dialogFactory && pAccess &&
       IsInteractive()) {
       if (!ShowHostInterface(
-         *pParent, dialogFactory, pInstance, *pAccess, IsBatchProcessing() ) )
+         *pParent, dialogFactory, pInstance, *pAccess, true ) )
          return false;
       else if (!pInstance)
          return false;
--- a/src/menus/PluginMenus.cpp
+++ b/src/menus/PluginMenus.cpp
@@ -279,8 +279,7 @@ void AddGroupedEffectMenuItems(
 
       groupNames.push_back( name );
       groupPlugs.push_back(plug->GetID());
-      groupFlags.push_back(
-         plug->IsEffectRealtime() ? realflags : FixBatchFlags( batchflags, plug ) );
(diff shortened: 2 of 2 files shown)


vsverchinsky (approved):

Paul-Licameli (changes requested):
I request that the merge be delayed.

Nothing wrong with the changes.

Paul-Licameli:
The reason to delay merge is that, while other effect families are making the transition to statelessness, the old interface for realtime effects should also remain available to QA while testing.

In case refactoring introduces errors, testing with the old user interface may help us understand if the regression in behavior is properly part of the refactoring of one effect family, or is instead some bug remaining in the general framework.

Paul-Licameli:
> I'm not sure if I understand your point. The interface is not changed (preview button is still in place) and I only found out that the dialog is not modal two weeks ago, I have always used `Preview` button

The Preview button may be available, but if it treats the effect like any other non-realtime effect, then it will pre-compute some sound to play, and not respond to control changes while playing.

That should be the behavior of the dialog in this branch (is it?) but is not the behavior I would like to keep available to QA.

Ryan-Miller-Audacity:
> @Ryan-Miller-Audacity @dozzzzer any comments on that? Do you use play button in such case?

I would say in general the play button is where I go when dealing with anything realtime.

 However, with the described usefulness of being able to determine if there is a regression in behavior or a bug in the general framework, would it not be possible to go ahead and do the merge, but keep a separate branch for any discovered bugs to be verified in to determine if it is a regression or not? Or would that require a lot of upkeep? IMO, it's best to have QA taking place how the end user will actually end up seeing/using it as much as possible, but using Paul's plan as a additional tool could be helpful for better determining bugs in relation to regression etc.

Essentially, i'm asking for a "best of both worlds" scenario here.

crsib:
Should this be merged now @Paul-Licameli?

Paul-Licameli:
> Should this be merged now @Paul-Licameli?

I still want to wait.

crsib:
@Paul-Licameli any reason to wait longer? It should be merged before alpha anyway

Paul-Licameli (approved):

Paul-Licameli:
VST3 statelessness is done and VST is not must-have for alpha, so poyekhali!

Issue #3185, opened 2022-07-06
Audacity freezes when importing AC3 and AMR files on FFMPEG 5.0


LWinterberg:
This affects FFMPEG 5.0 and 5.0.1, but not 4.4.1 or 2.2.2.

This affects AC3, AMR, but not M4A (aac), opus, and wma. 

Tested with a 30s noise sample. Exporting works just fine. I did not yet test even more obscure formats (which can be exported via the "custom ffmpeg export" option - cc @Ryan-Miller-Audacity )

LWinterberg:
retested with 5.0.1, everything imports just fine now :)

PR #3189, opened 2022-07-06
Fixes `read_packet` implementation for FFmpeg 4.4+

crsib:
Starting from some unknown moment FFmpeg (version 58 at least) requires returning `AVERROR_EOF` to mark the end of stream. 

In all versions FFmpeg accepted negative return values to indicate that error has occurred.

Resolves: #3185 

Recommended:


libraries/lib-ffmpeg-support/wrappers/AVIOContextWrapper.cpp
```diff
@@ -85,8 +85,8 @@ int AVIOContextWrapper::FileRead(void* opaque, uint8_t* buf, int size)
 {
    AVIOContextWrapper* wrapper = static_cast<AVIOContextWrapper*>(opaque);
    if (!(wrapper && wrapper->mpFile))
```
vsverchinsky, line 87: `wrapper->mpFile` seems to be redundant (dupicated in `Read`)

vsverchinsky (approved):

PR #3217, opened 2022-07-12
Activating the menu "Add Realtime Effects" now toggles the Realtime effects side panel.

ksoze95:
Resolves: #3215 

Recommended:

### DIFF
--- a/src/ProjectWindow.cpp
+++ b/src/ProjectWindow.cpp
@@ -1942,6 +1942,11 @@ void ProjectWindow::HideEffectsPanel()
    Layout();
 }
 
+bool ProjectWindow::IsEffectsPanelShown()
+{
+   return mEffectsWindow->IsShown();
+}
+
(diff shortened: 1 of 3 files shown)


src/ProjectWindow.cpp
```diff
@@ -1942,6 +1942,12 @@ void ProjectWindow::HideEffectsPanel()
    Layout();
 }
 
+bool ProjectWindow::IsEffectsPanelShown()
+{
+   wxWindowUpdateLocker freeze(this);
```
vsverchinsky, line 1947: What this line is for?

ksoze95: I was not sure if it was needed as mEffectsWindows is passed to mContainerWindow's SplitVertically in both ShowEffectsPanel and HideEffectsPanel.

I will removed the lock.

vsverchinsky (approved):

Issue #3221, opened 2022-07-12
Stacked effects with latency de-synchronize playback with other tracks


Paul-Licameli:
UPDATED alternative steps:
- Generate rhythm track
- Put (for instance AUGraphicEQ) in the stack
- mix-and-render to new track
- apply Invert effect to the new track
- play tracks together
- expect silence

Steps to reproduce:
- Generate some silence
- At about 0.5 seconds, zoom in close and pull one sample up to +1 with the pencil tool
- Duplicate the track
- Listen
- Use the menus to apply some realtime-capable effect destructively, which has nonzero latency (I tried with AUDynamicsProcessor which has 256 sample delay, about 5 milliseconds at 44.1 kHz; also AM pitch shifter of swh ladspa plug-ins, with the size slider at maximum, which makes 2048 sample delay, 46 ms), to the first track only
- Listen: it still sounds the same
- Undo the destructive effect
- But now put the same effect with same settings into the stack of the first track
- Listen:  it's not the same:  the first track's click is later than the other track, and it's humanly detectable even at 5 ms

Comparison with the destructively applied effect proves that it is not the effect computation itself causing what you hear at the third listen.  The code for destructive effect application accounts for latency in a way that the realtime playback does not.

I haven't yet found an example of an effect with longer latency that might make the problem more obvious.  For AUDynamicsProcessor, the 256 sample latency seems to be invariant, so if the tracks instead have the lowest available sample rate of 800 Hz, then the difference is about 30 ms and the change in sound becomes more obvious.

Some ladspa effects in the swh package have latency and might be experimented with too:
am_pitchshift_1433.xml -- I tried this with the "size" parameter at maximum, which makes a 2048 sample latency

bode_shifter_1431.xml
bode_shifter_cv_1432.xml
dj_eq_1901.xml
fast_lookahead_limiter_1913.xml
gsm_1215.xml
hilbert_1440.xml
imp_1199.xml
latency_1914.xml
mbeq_1197.xml
pitch_scale_1193.xml
pitch_scale_1194.xml

DavidBailes:
> I haven't yet found an example of an effect with longer latency that might make the problem more obvious

https://www.voxengo.com/product/latencydelay/

- with a delay of 0 entered in the interface, this will leave the delay introduced by the effect to be 10000 samples

pietro68:
[...]

Issue #3150, opened 2022-06-27
Realtime effects do not apply when Mixing-and-Rendering or exporting audio


dozzzzer:
**Describe the bug**
Realtime effects do not apply when Mixing-and-Rendering or exporting audio

**To Reproduce**
1. Generate 30 seconds DTMF Tones
2. Add Reverb as a real time effect (default settings)
3. Play - the reverb is audible
4. Export as WAV

OR

1. Generate 30 seconds DTMF Tones
2. Add Reverb as a real time effect (default settings)
3. Play - the reverb is audible
4. Mix and Render to New Track

**Actual behavior**
The exported file has no reverb / There is no reverb on the rendered track

**Expected behavior**
The exported file has reverb applied to it / There is reverb on the rendered track

**Additional information (please complete the following information):**
 - OS: any
 - Version: latest master

PR #3225, opened 2022-07-12
Per track effect refactor 1

Paul-Licameli:
Some preliminaries for addressing issues #3221 (latency) and #3150 (rendering)

Commits ec8ff49fe, 733cb1e4, f6e4fe993 added many arguments to
PerTrackEffect::ProcessTrack() as many unnecessary member variables became
stack variables instead for only the duration of the calling function; this exposed
much of the state ProcessTrack() was using.

This branch begins an encapsulation of this state into some objects of new classes
which will leave ProcessTrack() as a much shorter function with fewer arguments.

It also makes GetLatency() a const member function of effect instances, as it should
be.

Recommended:


pietro68 (approved):
Code changes are understood and fine for me.

Issue #3209, opened 2022-07-11
Stackable effects: all buttons should fit to the default width of the sidebar


dozzzzer:
**Describe the bug**
When adding an effect with a long title, it doesn't fit the default minimal width of the sidebar, which, in turn, makes gear icon hide and calculates the width of the 'Add effect' button wrongly.

**To Reproduce**
1. Add an effect with a long title in the stackable effects sidebar (e.g. AUMultibandCompressor) 
2. observe 

**Actual behavior**
1. gear button isn't visble
2. effect's button as well as 'Add effect' button are partially hidden
<img width="409" alt="image" src="https://user-images.githubusercontent.com/70063970/178274142-86df23ba-dbd8-4bab-be60-324e0c8fe343.png">

**Expected behavior**
1. both effect's button and gear button fit the default width if the sidebar; if the title of the effect is too long, it should be truncated with an ellipsis
2. 'Add effect' button is fully visible too

petersampsonaudacity:
@LWinterberg  and @Tantacrul 

How "discoverable" is it that you can click the join between the Effects Stack panel and the tracks panel to enlarge the Effects Stack panel ?

LWinterberg:
not massively, hence why it's in https://github.com/audacity/audacity/issues/3195 as an improvement

Issue #3078, opened 2022-06-15
Make the effects panel accessible


LWinterberg:
# Shortcut

Users should be able to hide / show the ‘Realtime effects’ panel by pressing the ‘e’ key (without modifier). Screenreader reads: “Realtime effects for `track name`”

Pressing `e` again or `ESC` closes the panel and re-focuses on the track. 

# Keyboard Navigation

Tab cycles through the elements (black outline), in the order:
1. Panel header (reads title) 
2. Panel power button
3. Close effect panel 
4. Effects slot 1 (whole slot)
5. Effect Power Button
6. Effect Dropdown
7. Effect Settings
8. Effects slot 2

...
n-1. Add New Effect
n. (the text info about realtime effects)
n+1. cycles back to the top (Panel header)

Further, Up/Down can be used to navigate between effect slots. Right is equivalent to Down, Left is equivalent to Up. 

<details><summary>examples</summary>

![image](https://user-images.githubusercontent.com/87814144/175000497-08e4c0f0-51b4-47ee-93da-6f85331f71a5.png)
![image](https://user-images.githubusercontent.com/87814144/175000527-f0fbe59c-e20c-436d-887c-ce2137776585.png)
![image](https://user-images.githubusercontent.com/87814144/175000553-71191427-c5fe-4a41-bba1-29e9fe51fbdc.png)

</details>

# Reordering Effects

When an effect slot is selected, pressing Alt+Down/Right moves it down, Alt+Left/Up moves it up. 

# Selecting Items
* The bypass can be activated/deactivated with Enter. 
* The dropdown can be opened with Enter
* The settings can be opened with Enter. 

# Outlines
The normal wx focus outline is used. 
<details><summary>examples</summary>

![image](https://user-images.githubusercontent.com/87814144/173855660-26ba77ec-69cb-49d5-8079-38588aa11e24.png)
![image](https://user-images.githubusercontent.com/87814144/173855699-b27eb445-aaf3-4fe5-a777-66c5a6121069.png)
![image](https://user-images.githubusercontent.com/87814144/173855757-75dfe340-1e12-4d7b-888f-b07eeb93eb8d.png)

</details>

DavidBailes:
I have previously suggested that it would be better if the list of effects in the effects panel had a keyboard interaction appropriate for a list, and was read by screen readers as a list. I appreciate that wxWidgets does not supply any list classes where each list item contains a number of controls, and so a custom list would have to be constructed.
[...]

PR #3229, opened 2022-07-13
Issue 3078: realtime effects UI keyboard navigation & accessibility

vsverchinsky:
Resolves: #3078 
Resolves: #3209

Effects stack UI can now handle navigation keys and available for screen readers (though reordering messages and toggle button state changes aren't delivered)
Changes in keyboard navigation: 'E' key opens effects panel for a currently focused track; effects panel also added to panel navigation loop sequence (Ctrl+F6, Ctrl+Shift+F6)
Fixes toggling AButton withih enter key

Recommended:


src/RealtimeEffectPanel.cpp
```diff
@@ -160,10 +361,63 @@ namespace
          event.SetSourceIndex(mSourceIndex);
          event.SetTargetIndex(mTargetIndex);
          event.SetEventObject(this);
-         wxPostEvent(target, event);
+         target->GetEventHandler()->ProcessEvent(event);
       }
-      
+
+      int FindIndexInParent()
```
ksoze95, line 367: Could this method be made const?

vsverchinsky: Seems so, thanks! I'll fix that

Paul-Licameli:
Will you soon fix build failures?

src/widgets/ImageRoll.h
```diff
@@ -69,32 +69,4 @@ class AUDACITY_DLL_API ImageRoll
    wxSize       mMaxSize;
 };
 
-// A very simple class that just display an ImageRoll that doesn't
```
ksoze95, line 72: I think both ImageRoll files could be removed as there are no longer any references to them.

src/ThemedWrappers.h
```diff
@@ -161,10 +171,103 @@ class ThemedButtonWrapper final :
...
+      }
+   }
+
+private:
+
+   void OnThemeChange(ThemeChangeMessage message)
+   {
+      if (message.appearance)
+      {
+         for(auto& p : mSets)
```
ksoze95, line 253: Could this be changed to a 'const auto&'?

src/RealtimeEffectPanel.cpp
```diff
@@ -760,19 +1107,29 @@ class RealtimeEffectListWindow : public wxScrolledWindow
...
-      const auto removeItem = [this, &msg](){
-         auto sizer = mEffectListContainer->GetSizer();
-         auto item = sizer->GetItem(msg.srcIndex)->GetWindow();
-         item->Destroy();
-#ifdef __WXGTK__
-         // See comment in ReloadEffectsList
+      const auto removeItem = [&](){
+         auto window = sizer->GetItem(msg.srcIndex)->GetWindow();
+         sizer->Remove(msg.srcIndex);
+         wxTheApp->CallAfter([ref = wxWeakRef { window }] {
```
ksoze95, line 1120: Missing <wx/app.h> include for reference to wxTheApp

src/widgets/ImageRoll.h
Paul-Licameli: I disagree.  Understand whether there are good ideas there worth revisiting before you say that.

And I have never myself studied these files closely enough to answer that.

src/RealtimeEffectPanel.cpp
```diff
@@ -842,23 +1204,32 @@ class RealtimeEffectListWindow : public wxScrolledWindow
...
-#ifdef __WXGTK__
-      //Workaround for GTK: Underlying GTK widget does not update
-      //its size when wxWindow size is set to zero
+      
       if(!mTrack || RealtimeEffectList::Get(*mTrack).GetStatesCount() == 0)
          mEffectListContainer->Hide();
-#endif
-
+      
+      auto isEmpty{true};
```
ksoze95, line 1217: Possibly change to 'const auto'.  Just being pedantic.

src/widgets/ImageRoll.h
Paul-Licameli: ImageRoll.h is used by AButton.  ExpandingToolBar.* are the unused files.  I suggest don't remove them, but maybe don't modify them except minimally to keep them compiling?

src/menus/NavigationMenus.cpp
```diff
@@ -29,57 +29,46 @@ void NextOrPrevFrame(AudacityProject &project, bool forward)
...
+      ProjectWindow::Get(project).GetEffectsWindow(),
+      ToolManager::Get( project ).GetBotDock(),
    };
 
    auto FindAncestor = [&]() {
       wxWindow *pWindow = wxWindow::FindFocus();
-      unsigned index = rotationSize;
-      while ( pWindow &&
-              (rotationSize == (index = IndexOf(pWindow) ) ) )
+      unsigned index = seq.size();
```
ksoze95, line 42: This line can be removed as index is not referenced.

src/menus/NavigationMenus.cpp
```diff
@@ -29,57 +29,46 @@ void NextOrPrevFrame(AudacityProject &project, bool forward)
...
+      while (pWindow)
+      {
+         auto it = std::find(seq.begin(), seq.end(), pWindow);
```
ksoze95, line 45: I think begin/end could be replaced with cbegin/cend.

src/menus/NavigationMenus.cpp
```diff
@@ -29,57 +29,46 @@ void NextOrPrevFrame(AudacityProject &project, bool forward)
...
    auto idx2 = idx;
-   auto increment = (forward ? 1 : rotationSize - 1);
-
-   while( idx != (idx2 = (idx2 + increment) % rotationSize) ) {
-      wxWindow *toFocus = begin[idx2];
-      bool bIsAnEmptyDock=false;
-      if( idx2 != 1 )
-         bIsAnEmptyDock = ((idx2==0) ? toolManager.GetTopDock() : botDock)->
-         GetChildren().GetCount() < 1;
+   const auto increment = (forward ? 1 : static_cast<int>(seq.size()) - 1);
```
ksoze95, line 58: How about adding a pre-condition assert check before this line, an assert(seq.size() != 0)?  If seq.size() is 0 and forward is false then increment will become a very large number as size_t is an unsigned number.

src/ThemedWrappers.h
```diff
@@ -105,6 +109,12 @@ class ThemedButtonWrapper final :
       return mBitmapIndex;
    }
 
+   void SetBitmapFocusIndex(int index)
+   {
```
ksoze95, line 113: Does this function need to have out of bounds check, especially before calling theTheme.Bitmap with an invalid index?

src/RealtimeEffectPanel.cpp
vsverchinsky: Can't be made const, it changes value later

src/menus/NavigationMenus.cpp
vsverchinsky: Right, thanks!

src/menus/NavigationMenus.cpp
vsverchinsky: It can't be zero-sized, though it may contain nullptrs

src/ThemedWrappers.h
vsverchinsky: There is a check inside theTheme.Bitmap, but I think it's not really neccessary since all bitmap indices have names in the code

src/RealtimeEffectPanel.cpp
ksoze95: Ah, I missed that.

ksoze95 (approved):

src/widgets/AButton.cpp
```diff
@@ -390,10 +385,8 @@ void AButton::OnPaint(wxPaintEvent & WXUNUSED(event))
 
    mImages[mAlternateIdx].mArr[buttonState].Draw(dc, GetClientRect());
 
-   if( this == wxWindow::FindFocus() )
-   {
+   if(HasFocus())
```
Paul-Licameli, line 392: This change makes a difference for windows that override `GetMainWindowOfCompositeControl`.  What are such windows?

Paul-Licameli: Combos and list controls I see, but no such are used in the sidebar.

src/widgets/AButton.h
```diff
@@ -13,36 +13,50 @@
 #define __AUDACITY_BUTTON__
 
 #include <vector>
+#include <array>
+#include <memory>
 
 #include <wx/setup.h> // for wxUSE_* macros
 #include <wx/window.h> // to inherit
 
-#include "ImageRoll.h" // member variable
```
Paul-Licameli, line 20: This deletion requires you insert the `#include` in `ExpandingToolBar.cpp` so that the commit "AButton refactoring" builds

libraries/lib-theme/AColor.h
```diff
@@ -50,6 +50,7 @@ class THEME_API AColor {
    static void Bevel(wxDC & dc, bool up, const wxRect & r);
    static void Bevel2
       (wxDC & dc, bool up, const wxRect & r, bool bSel=false, bool bHighlight = false);
+   static void DrawHStretch(wxDC& dc, const wxRect& rect, wxBitmap& bitmap);
```
Paul-Licameli, line 53: Comment, please

libraries/lib-theme/AColor.cpp
```diff
@@ -305,6 +305,23 @@ void AColor::Bevel2(wxDC & dc, bool up, const wxRect & r, bool bSel, bool bHighl
...
+
+   const auto sh = bitmap.GetHeight();
+   const auto dh = rect.height;
+   const auto w0 = std::min((bitmap.GetWidth() - 1) / 2, rect.width / 2);
+   const auto dx0 = rect.x;
+   const auto dx1 = rect.x + w0;
+   const auto dx2 = rect.x + rect.width - 1 - w0;
+
+   dc.StretchBlit(dx0, rect.y, w0, dh, &srcDC, 0, 0, w0, sh);
+   dc.StretchBlit(dx1, rect.y, rect.width - w0 * 2, dh, &srcDC, w0, 0, 1, sh);
```
Paul-Licameli, line 321: Off by one mistake -- the third argument is not `dx1 - dx0`

Paul-Licameli: And dx2 - dx1 may be zero or negative, in which case, skip it

src/RealtimeEffectPanel.cpp
```diff
@@ -290,8 +288,8 @@ namespace
       std::shared_ptr<RealtimeEffectState> mEffectState;
       std::shared_ptr<EffectSettingsAccess> mSettingsAccess;
 
-      ThemedButtonWrapper<wxButton>* mChangeButton{nullptr};
-      ThemedButtonWrapper<wxBitmapButton>* mEnableButton{nullptr};
+      ThemedAButtonWrapper<AButton>* mChangeButton{nullptr};
```
Paul-Licameli, line 603: The commit "RealtimeEffectPanel wx*Button replaced with AButton" doesn't build.  The next commit repairs that.  Make both build.

src/widgets/AButton.cpp
Paul-Licameli: You don't add any overrides in later commits.

But this change is harmless.

src/menus/NavigationMenus.cpp
```diff
@@ -29,57 +29,45 @@ void NextOrPrevFrame(AudacityProject &project, bool forward)
...
       return;
 
    auto idx2 = idx;
-   auto increment = (forward ? 1 : rotationSize - 1);
-
-   while( idx != (idx2 = (idx2 + increment) % rotationSize) ) {
-      wxWindow *toFocus = begin[idx2];
-      bool bIsAnEmptyDock=false;
-      if( idx2 != 1 )
-         bIsAnEmptyDock = ((idx2==0) ? toolManager.GetTopDock() : botDock)->
```
Paul-Licameli, line 73: See: https://bugzilla.audacityteam.org/show_bug.cgi?id=1564

I think this change may break that old bug fix, on Windows though not on Mac.  The first window in `seq` is not a `ToolDock` but the parent of one.  Drag all toolbars off the top dock so they float, and test it.

Paul-Licameli: See also #3419

src/widgets/ImageRoll.cpp
```diff
@@ -421,41 +421,3 @@ void ImageRoll::Draw(wxDC &dc, wxRect rect, int WXUNUSED(logicalFunc))
 
    } // switch
 }
-
-BEGIN_EVENT_TABLE(ImageRollPanel, wxPanelWrapper)
```
Paul-Licameli, line 425: The changes in commit "ImageRollPanel usage replaced with wxPanelWrapper" are not necessary for this issue.  Not harmful either, as they remove and change two unused classes.  Why did you think you needed this?

src/RealtimeEffectPanel.cpp
```diff
@@ -987,6 +987,8 @@ RealtimeEffectPanel::RealtimeEffectPanel(wxWindow* parent, wxWindowID id, const
    mEffectList = effectList;
 
    SetSizerAndFit(vSizer.release());
+
+   Bind(wxEVT_CHAR_HOOK, &RealtimeEffectPanel::OnCharHook, this);
```
Paul-Licameli, line 1364: It seems that almost everywhere else the program handles `WXK_ESCAPE`, it's the key-down, not CHAR event.  Does this matter?  I don't know.  If it works, ok

src/RealtimeEffectPanel.cpp
```diff
@@ -290,8 +288,8 @@ namespace
       std::shared_ptr<RealtimeEffectState> mEffectState;
       std::shared_ptr<EffectSettingsAccess> mSettingsAccess;
 
-      ThemedButtonWrapper<wxButton>* mChangeButton{nullptr};
-      ThemedButtonWrapper<wxBitmapButton>* mEnableButton{nullptr};
+      ThemedAButtonWrapper<AButton>* mChangeButton{nullptr};
+      AButton* mEnableButton{nullptr};
```
Paul-Licameli, line 604: The power buttons are just `AButton` and not themable.  Does this matter?  Are the images really the same in all themes?

src/RealtimeEffectPanel.cpp
```diff
@@ -290,8 +288,8 @@ namespace
       std::shared_ptr<RealtimeEffectState> mEffectState;
       std::shared_ptr<EffectSettingsAccess> mSettingsAccess;
 
-      ThemedButtonWrapper<wxButton>* mChangeButton{nullptr};
```
Paul-Licameli, line 293: The template `ThemedButtonWrapper` is now unused at commit "RealtimeEffectPanel wx*Button replaced with AButton" while the template `ThemedAButtonWrapper` is only used with AButton as the template argument.

I wonder if the templates are really needed then, but maybe you think there will be future uses.

src/RealtimeEffectPanel.cpp
```diff
@@ -310,45 +308,49 @@ namespace
          auto sizer = std::make_unique<wxBoxSizer>(wxHORIZONTAL);
 
          //On/off button
-         auto enableButton = safenew ThemedButtonWrapper<wxBitmapButton>(this, wxID_ANY, wxBitmap{}, wxDefaultPosition, wxDefaultSize, wxNO_BORDER);
-         enableButton->SetBitmapIndex(bmpEffectOn);
+         auto enableButton = safenew ThemedAButtonWrapper<AButton>(this);
```
Paul-Licameli, line 634: What was the major advantage in this replacement of `wxButton` with `AButton`?

More states that can be associated with changed images?

More convenient control of the button state?

src/RealtimeEffectPanel.cpp
```diff
@@ -293,10 +294,63 @@ namespace
          event.SetSourceIndex(mSourceIndex);
          event.SetTargetIndex(mTargetIndex);
          event.SetEventObject(this);
-         wxPostEvent(target, event);
+         target->GetEventHandler()->ProcessEvent(event);
```
Paul-Licameli, line 365: Perhaps then the name of this function should change to `ProcessDragEvent` because it no longer delays the dispatch.

src/RealtimeEffectPanel.cpp
```diff
@@ -511,31 +511,31 @@ namespace
             }
          });
 
-         //Central button with effect name
-         const auto changeButton = safenew ThemedAButtonWrapper<AButton>(this, wxID_ANY);
-         changeButton->SetImageIndices(0,
+         //Central button with effect name, show settings
+         const auto optionsButton = safenew ThemedAButtonWrapper<AButton>(this, wxID_ANY);
+         optionsButton->SetImageIndices(0,
```
Paul-Licameli, line 654: I was not aware that this interchange of behavior of the two buttons was specified, but it is certainly correctly implemented.

src/ProjectWindow.cpp
```diff
@@ -641,6 +641,7 @@ ProjectWindow::ProjectWindow(wxWindow * parent, wxWindowID id,
    mContainerWindow->Initialize(mTrackListWindow);
 
    auto effectsPanel = safenew ThemedWindowWrapper<RealtimeEffectPanel>(mContainerWindow, wxID_ANY);
+   effectsPanel->SetName(_("Realtime effects"));
```
Paul-Licameli, line 644: This long-distance coincidence of literals here and in RealtimeEffectsPanel.cpp is unfortunate.  Maybe a named constant.  Also the mismatch of capitalization would make two strings in the translation catalog.

The _ macro does a run-time translation.  If the language preference changes, is this name reset?  No.  That means the screen reader might read it in the wrong language.  Important?  Maybe not enough to fix, but comment it.

src/RealtimeEffectPanel.cpp
```diff
@@ -866,9 +944,12 @@ class RealtimeEffectListWindow : public wxScrolledWindow
       mAddEffectHint = addEffectHint;
 
       //TODO: set link
-      auto addEffectTutorialLink = safenew ThemedWindowWrapper<wxHyperlinkCtrl>(this, wxID_ANY, "x", "https://www.audacityteam.org");
+      auto addEffectTutorialLink = safenew ThemedWindowWrapper<wxHyperlinkCtrl>(this, wxID_ANY, _("Watch video"), "https://www.audacityteam.org");
```
Paul-Licameli: I see that the template that became unused earlier now is reused.

I see _ again but the template adds the handler for preference changes that updates the title.

The handler for preference change in that template, and also in the `ThemedButtonWrapper`, should also update the window name with the title, for the screen reader.

src/RealtimeEffectPanel.cpp
```diff
@@ -1213,6 +1301,8 @@ void RealtimeEffectPanel::SetTrack(AudacityProject& project, const std::shared_p
          : mToggleEffects->PopUp();
       mEffectList->SetTrack(project, track);
       mProject = &project;
+      //i18n-hint: argument - track name
+      mHeader->SetName(wxString::Format(_("Realtime effects for %s"), track->GetName()));
```
Paul-Licameli, line 1380: If language is changed in preferences, this line is not visited to update the name.

Paul-Licameli:
I'm done!

Paul-Licameli (changes requested):
> See: https://bugzilla.audacityteam.org/show_bug.cgi?id=1564
> 
> I think this change may break that old bug fix, on Windows though not on Mac. The first window in `seq` is not a `ToolDock` but the parent of one. Drag all toolbars off the top dock so they float, and test it.

In my opinion the regression above is the one suggestion I made that should be fixed before this is merged.

Making all commits buildable, fixing the off-by-one in AColor.cpp, and commenting that function, would be nice but inessential.

Making all the updates in case of change of language in preferences is not essential.

libraries/lib-theme/AColor.cpp
vsverchinsky: Right, but `dx2` calculation is wrong

vsverchinsky: Right, but `dx2` calculation is wrong

src/menus/NavigationMenus.cpp
vsverchinsky: Thanks, will fix that

vsverchinsky: @Paul-Licameli I tested it in current master (without my changes) as you said and both top and bottom panels are focusable, so even if I revert this check in my branch it wouldn't fix the issue.  Also tried after configuration reset, both docks aren't empty, though they are look empty

src/widgets/ImageRoll.cpp
vsverchinsky: For same reason we are trying to avoid dependency cycles, because aliases complicates the code

src/RealtimeEffectPanel.cpp
vsverchinsky: Key down event delivered to the window that has focus. wxDialog handles WXK_ESCAPE same way

src/RealtimeEffectPanel.cpp
vsverchinsky: It is themable, but stored as a `AButton*`

src/RealtimeEffectPanel.cpp
vsverchinsky: Native buttons are not focusable on mac until such behaviour explicitly enabled in the preferences. Native buttons do not support text elision, appearance can't be customized... many reasons actually

src/ProjectWindow.cpp
vsverchinsky: >If the language preference changes, is this name reset? No. That means the screen reader might read it in the wrong language. Important?

As I concerned it works same way everyway, don't see why this case is special

src/menus/NavigationMenus.cpp
Paul-Licameli: I don't understand.  Top or bottom panel is meant to be skipped in the Ctrl-f6 cycle if there are no toolbars on the dock.

Are you saying that there was already a regression in this behavior, in master, for Windows?

If so, then don't try to fix the regression if your PR is not introducing it.  There should instead be an issue which I think should not be fixed for alpha.

Paul-Licameli:
I see you have force-pushed some changes, but this is not yet rebased onto recent master, and I understand there will be conflicts to resolve.

src/menus/NavigationMenus.cpp
Paul-Licameli: Now I see and understand your rewrite.

Paul-Licameli (approved):

Issue #3234, opened 2022-07-14
The nodes inside the Auto Duck effect cannot be dragged


dozzzzer:
**Describe the bug**
The nodes inside the Auto Duck effect cannot be dragged

**To Reproduce**
1. Add an audio track as Track 1
2. Add Track 2
3. Select Track 1 and go to Effect > Auto Duck
4. Try to drag the nodes on the graph
<img width="400" alt="image" src="https://user-images.githubusercontent.com/70063970/179011105-56394301-75bc-4923-9b3b-2cda679738a6.png">

**Actual behavior**
The nodes can no longer be dragged

**Expected behavior**
The nodes can be dragged

**Additional context**

- a regression against 3.2.0 deaa3f 
- the effect can be set up by entering the desired values into the inputs

dozzzzer:
It was decided to reduce priority to **P2** since it is actually possible to interact with the effect by entering values into the input fields which makes it somewhat useable.

PR #3238, opened 2022-07-14
Auto Duck - Fix Drag And Drop

adamdalechurch:
Resolves: https://github.com/audacity/audacity/issues/3234

Impetus:
Box Values Not Updated on drag.

Recommended:
- [X ] Each commit compiles and runs on my machine without known undesirable changes of behavior

### DIFF
--- a/src/effects/AutoDuck.cpp
+++ b/src/effects/AutoDuck.cpp
@@ -823,22 +823,27 @@ void EffectAutoDuck::Panel::OnMotion(wxMouseEvent & evt)
          case outerFadeDown:
             newValue = ((double)(FADE_DOWN_START - evt.GetX())) / FADE_SCALE;
             mEffect->mOuterFadeDownLen = std::clamp<double>(newValue, OuterFadeDownLen.min, OuterFadeDownLen.max);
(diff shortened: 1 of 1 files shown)


Paul-Licameli (changes requested):
Did you determine the commit that caused this regression?  I did and it's 652588d268b3c4e7abea303a92d3943a9c1e0a0f

This can be replaced with a simpler two line fix.

Restore the deleted lines

```
  if (!mUIParent->TransferDataToWindow())
      return false;
```
just before the call tp `DoTransferDataToWindow()`

adamdalechurch:
No sorry, new here.

Wouldn't know where to begin to do that.

Guess I should just delete this PR?

P.s. Love Audacity, thanks for doing great work on it.

Paul-Licameli:
> No sorry, new here.
> 
> Wouldn't know where to begin to do that
> 
> Guess I should just delete this PR?
> 
> P.s. Love Audacity, thanks for doing great work on it.

Are you familiar with `git bisect`?  Look it up:  it's one way to discover where the commit history introduced a bug, if you
know one point where the bug does happen (current master) and one where it did not (last release, the tag Audacity-3.1.3).

I encourage you to make this pull request again as I suggested, verifying that it really fixes things (I didn't, I only deduced that from reading code, but I'm fallible).

adamdalechurch:
I will do that. Thanks.

Issue #3214, opened 2022-07-12
LV2 plugins don't monitor signal level


dozzzzer:
**Describe the bug**
LV2 metering plugins don't monitor input signal levels.

**To Reproduce**
1. Download and install Meter Collection from x42 (https://x42-plugins.com/x42/x42-meters)
2. Try to call various meter plugins either from the real-time sidebar or destructively (e.g. DR-14, True-Peak  + RMS)
3. Check the signal levels

**Actual behavior**
Metering simply doesn't work as there's no reflection of the input signal levels

**Expected behavior**
The metering shows correct values and updates at the proper rate.

**Screenshots**
<img width="666" alt="image" src="https://user-images.githubusercontent.com/70063970/178461810-3310ba73-01c5-4ac8-b6fc-45f3cab85657.png">

https://user-images.githubusercontent.com/70063970/178461785-31756027-44ff-4b85-ae7d-383734202a3e.mp4

**Additional information (please complete the following information):**
 - OS: MacOS 12.2.1
 - latest master

dozzzzer:
Verified on MacOS 12.2.1 and Windows 11 x64.

Issue #3240, opened 2022-07-15
LV2 Output Monitoring does not work in plain UI


crsib:
**Describe the bug**
LV2 Output Monitoring does not work in plain UI

It was agreed with @Tantacrul that this issue is lower priority, so it is split from #3214

PR #3241, opened 2022-07-15
LV2 output monitoring

crsib:
Resolves: #3214 

This PR fixes the output monitoring for LV2 and LADSPA effect families, which are similarly implemented. Unlike AudioUnits - LV2/LADSPA have output ports, that are used to update the UI. 

Audacity implementation maps both input and output ports into the `EffectSettings` instance. During the realtime processing,
Audacity uses double buffered channels to transfer settings data between main and worker threads. In order not to overwrite
output values during this process `SettingsCopyDirection` argument is added to `CopySettingsContents`. This argument is
ignored now by all effects except LV2/LADSPA.

The next issue that was observed - it was not possible to sync `RealtimeEffectState::Access` and `RealtimeEffectState`. The sync now can be performed explicitly by calling `Flush`. It is not possible to sync implicitly, as it breaks effects UI. (That was observed first with AU and fixed by @Paul-Licameli)

`Flush` is called by the LV2Validator inside idle handler. LADSPA implementation is different and it seems that it doesn't require Flush call (but if LADSPA would work properly with non modal dialogs is a question).

This PR does not provide output monitoring for plain UI, see #3240

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -239,6 +239,15 @@ public:
    virtual bool IsHiddenFromMenus() const;
 };
 
+//! Direction in which settings are copied
+enum class SettingsCopyDirection
+{
+   //! Main thread settings replicated to the worker thread
+   MainToWorker,
+   //! Worker thread settings replicated to main thread
+   WorkerToMain
+};
+
 /*************************************************************************************//**
 
 \class EffectSettingsManager
@@ -277,10 +286,11 @@ public:
 
     @param src settings to copy from
     @param dst settings to copy into
+    @param copyDirection direction in which copy is performed
     @return success
     */
    virtual bool CopySettingsContents(
-      const EffectSettings &src, EffectSettings &dst) const;
+      const EffectSettings &src, EffectSettings &dst, SettingsCopyDirection copyDirection) const;
 
    //! Store settings as keys and values
    /*!
--- a/src/effects/RealtimeEffectState.cpp
+++ b/src/effects/RealtimeEffectState.cpp
@@ -189,7 +189,6 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
    void Flush() override {
       if (auto pState = mwState.lock()) {
          if (auto pAccessState = pState->GetAccessState()) {
-            auto &lastSettings = pAccessState->mLastSettings;
             const EffectSettings *pResult{};
             while (!(pResult = FlushAttempt(*pAccessState))) {
                // Wait for progress of audio thread
@@ -197,6 +196,7 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
                std::this_thread::sleep_for(50ms);
             }
             pState->mMainSettings.Set(*pResult); // Update the state
+            pAccessState->mLastSettings = *pResult;
          }
       }
    }
--- a/src/effects/RealtimeEffectState.cpp
+++ b/src/effects/RealtimeEffectState.cpp
@@ -78,7 +78,9 @@ public:
       // Worker thread writes the slot
       ToMainSlot& operator=(EffectAndSettings &&arg) {
          // This happens during MessageBuffer's busying of the slot
-         arg.effect.CopySettingsContents(arg.settings, mSettings);
+         arg.effect.CopySettingsContents(
+            arg.settings, mSettings,
+            SettingsCopyDirection::WorkerToMain);
          mSettings.extra = arg.settings.extra;
          return *this;
       }
@@ -111,7 +113,9 @@ public:
       struct Reader { Reader(FromMainSlot &&slot,
          const EffectSettingsManager &effect, EffectSettings &settings) {
             // This happens during MessageBuffer's busying of the slot
-            effect.CopySettingsContents(slot.mSettings, settings);
+            effect.CopySettingsContents(
+               slot.mSettings, settings,
+               SettingsCopyDirection::MainToWorker);
             settings.extra = slot.mSettings.extra;
       } };
 
--- a/src/effects/ladspa/LadspaEffect.cpp
+++ b/src/effects/ladspa/LadspaEffect.cpp
@@ -121,7 +121,7 @@ EffectSettings LadspaEffect::MakeSettings() const
 }
 
 bool LadspaEffect::CopySettingsContents(
-   const EffectSettings &src, EffectSettings &dst) const
+   const EffectSettings &src, EffectSettings &dst, SettingsCopyDirection copyDirection) const
 {
    // Do not use the copy constructor of std::vector.  Do an in-place rewrite
    // of the destination vector, which will not allocate memory if dstControls
--- a/src/effects/ladspa/LadspaEffect.cpp
+++ b/src/effects/ladspa/LadspaEffect.cpp
@@ -126,10 +126,33 @@ bool LadspaEffect::CopySettingsContents(
    // Do not use the copy constructor of std::vector.  Do an in-place rewrite
    // of the destination vector, which will not allocate memory if dstControls
    // began with sufficient capacity.
+   const auto portCount = mData->PortCount;
+   
    auto &srcControls = GetSettings(src).controls;
    auto &dstControls = GetSettings(dst).controls;
-   dstControls.resize(0);
-   copy(srcControls.begin(), srcControls.end(), back_inserter(dstControls));
+
+   assert(srcControls.size() == portCount);
+   assert(dstControls.size() == portCount);
+
+   const auto portValuesCount =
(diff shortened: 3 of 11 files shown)


crsib:
I observe some odd behavior in release builds (the do not affect this specific branch, so please review). Looks like the last commit needs to be updated

crsib:
I have dropped commits that reuse the first processor for the UI. I do not expect that it will affect much of the plugins, as LV2 docs suggest against using effect instance from the UI: http://lv2plug.in/ns/ext/instance-access 

The reason why the commits were dropped is that it is possible that playback and project rates are different. As it seem, LV2 does not allow changing the sample rate of an instance. Consider the following scenario:

1. Effect is opened from the effect menu. An instance is created with the project's sample rate.
2. Preview button is pressed: we need to recreate the instance with the playback sample rate. However, this instance is already referenced by the UI.

libraries/lib-components/EffectInterface.h
```diff
@@ -277,10 +286,11 @@ class COMPONENTS_API EffectSettingsManager  /* not final */
 
     @param src settings to copy from
     @param dst settings to copy into
+    @param copyDirection direction in which copy is performed
     @return success
-    */
+    */
```
Paul-Licameli, line 291: white space

src/effects/lv2/LV2Effect.cpp
```diff
@@ -191,20 +191,40 @@ EffectSettings LV2Effect::MakeSettings() const
...
 {
    auto &srcControls = GetSettings(src).values;
    auto &dstControls = GetSettings(dst).values;
 
+   const auto &controlPorts = mPorts.mControlPorts;
+   const auto portsCount = controlPorts.size();
    // Do not use the copy constructor of std::vector.  Do an in-place rewrite
    // of the destination vector, which will not allocate memory if dstControls
    // began with sufficient capacity.
    // And that will be try if dstControls originated with MakeSettings() or a
```
Paul-Licameli, line 205: Fix this comment, should say "true" not "try"

src/effects/lv2/LV2Effect.cpp
```diff
@@ -191,20 +191,40 @@ EffectSettings LV2Effect::MakeSettings() const
...
+      return false;
+
+   const auto copyInputs = copyDirection == SettingsCopyDirection::MainToWorker;
+   
+   size_t portIndex {};
+
+   for (auto& port : controlPorts)
+   {
+      if (port->mIsInput == copyInputs)
+         dstControls[portIndex] = srcControls[portIndex];
```
Paul-Licameli, line 224: This loop copies input ports if and only if direction is MainToWorker, and output ports, if and only if, the opposite.

But maybe sufficient to fix the bug is:  change behavior only for MainToWorker, but leave it unchanged when copying worker to main again.  Which changes behavior in fewer circumstances.

src/effects/ladspa/LadspaEffect.cpp
```diff
@@ -121,15 +121,38 @@ EffectSettings LadspaEffect::MakeSettings() const
...
+
+   for (unsigned long p = 0; p < portCount; ++p)
+   {
+      LADSPA_PortDescriptor d = mData->PortDescriptors[p];
+      
+      if (!(LADSPA_IS_PORT_CONTROL(d)))
+         continue;
+
+      if (copyInputs == LADSPA_IS_PORT_INPUT(d))
+         dstControls[p] = srcControls[p];
```
Paul-Licameli, line 153: Ditto for Ladspa as for LV2

src/effects/RealtimeEffectState.cpp
```diff
@@ -185,14 +189,14 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
...
          if (auto pAccessState = pState->GetAccessState()) {
-            auto &lastSettings = pAccessState->mLastSettings;
             const EffectSettings *pResult{};
             while (!(pResult = FlushAttempt(*pAccessState))) {
                // Wait for progress of audio thread
                using namespace std::chrono;
                std::this_thread::sleep_for(50ms);
             }
             pState->mMainSettings.Set(*pResult); // Update the state
+            pAccessState->mLastSettings = *pResult;
```
Paul-Licameli, line 199: Thanks for finding the error in my Access class.  I think I suggested making the fix in `FlushAttempt` instead.  That might update the main-thread version of output port values a little more frequently, maybe improving update rate seen in theater?  But either way is sufficient for `Flush().`

crsib: After testing the code it turned out, that it results in problems when changing the settings from the effect UI, which you have mentioned to be the case for AU previously.

crsib: Probably it will be fixed when CopySettings will be fixed according to your proposal as well

Paul-Licameli:
I pushed a suggested extra commit for incidental cleanup of OnIdle

Paul-Licameli (changes requested):
What I suggested for CopySettingsContents in lv2 and Ladspa is certainly needed.  Stickiness of settings, if they are changed during playback, is broken by this PR.

Paul-Licameli:
> I have pushed the changes you requested. I still strongly disagree with the commit you have pushed - I see little evidence that it is needed at all and I am sure that it should not be a part of this PR. However, I do not want to keep this PR any longer than needed.

I'm just guarding against possible UB (calling a null function pointer) and I don't see any harm.

Paul-Licameli (approved):

Paul-Licameli:
Include or remove my suggested commit as you prefer

crsib:
>I'm just guarding against possible UB (calling a null function pointer) and I don't see any harm.

I do agree that we should not trust the LV2 developers, but I feel that the chances that effect declares `LV2_UI__idleInterface` but does not implement a single function of the interface sound miniscule. 

On top of that, according to the docs:

----

```struct LV2UI_Show_Interface```
```UI Show Interface (LV2_UI__showInterface)```

UIs can provide this interface to show and hide a window, which allows them to function in hosts unable to embed their widget. This allows any UI to provide a fallback for embedding that works in any host.

If used:

* The host MUST use [LV2UI_Idle_Interface](https://lv2plug.in/doc/html/group__ui.html#structLV2UI__Idle__Interface) to drive the UI.
* The UI MUST return non-zero from [LV2UI_Idle_Interface::idle()](https://lv2plug.in/doc/html/group__ui.html#a72d526f42aeb2b9106eec7bdcde77d4d) when it has been closed.
* If idle() returns non-zero, the host MUST call [hide()](https://lv2plug.in/doc/html/group__ui.html#abc65fe3322349d5a3aadacfba90afbbe) and stop calling idle(). It MAY later call [show()](https://lv2plug.in/doc/html/group__ui.html#ac5f8e21bb1914304b16058dcfb3411c8) then resume calling idle().

----

So do we remove UB (and dereferencing `nullptr` is not quite an UB unless you are writing the kernel) or do we add more by violating the LV2 expectations from the host?

crsib:
I've split you commit into #3270

Paul-Licameli:
> > I'm just guarding against possible UB (calling a null function pointer) and I don't see any harm.
> 
> I do agree that we should not trust the LV2 developers, but I feel that the chances that effect declares `LV2_UI__idleInterface` but does not implement a single function of the interface sound miniscule.
> 
> On top of that, according to the docs:
> 
> `struct LV2UI_Show_Interface` `UI Show Interface (LV2_UI__showInterface)`
> 
> UIs can provide this interface to show and hide a window, which allows them to function in hosts unable to embed their widget. This allows any UI to provide a fallback for embedding that works in any host.
> 
> If used:
> 
> * The host MUST use [LV2UI_Idle_Interface](https://lv2plug.in/doc/html/group__ui.html#structLV2UI__Idle__Interface) to drive the UI.
> * The UI MUST return non-zero from [LV2UI_Idle_Interface::idle()](https://lv2plug.in/doc/html/group__ui.html#a72d526f42aeb2b9106eec7bdcde77d4d) when it has been closed.
> * If idle() returns non-zero, the host MUST call [hide()](https://lv2plug.in/doc/html/group__ui.html#abc65fe3322349d5a3aadacfba90afbbe) and stop calling idle(). It MAY later call [show()](https://lv2plug.in/doc/html/group__ui.html#ac5f8e21bb1914304b16058dcfb3411c8) then resume calling idle().
> 
> So do we remove UB (and dereferencing `nullptr` is not quite an UB unless you are writing the kernel) or do we add more by violating the LV2 expectations from the host?

LV2 appears to do a good job of stating precise contracts, and you know I'm putting more emphasis on programming by contract too.

But still: there is the big but bounded base of trusted source code we control, which allows us to elide checks of preconditions in release, and yet there is also foreign code we can't always trust to honor contracts.

Issue #3195, opened 2022-07-07
toolbar/effects panel UI prettifications


LWinterberg:
- [x] Icons - #3142
- [x] sliders - #3248
- [ ] Button backgrounds
- [ ] rescaling transport toolbar buttons to 36x36 on all platforms
- [ ] aligning realtime effects x button and playhead options button
- [ ] adding a divider line between realtime effects panel and project area, as well as between sidebar header and items
- [ ] widening the realtime effects panel default size slightly
- [ ] switching behavior of gear and name in realtime effects panel

PR #3248, opened 2022-07-18
sliders

LWinterberg:
resolves part 2 of #3195


petersampsonaudacity:
Those look better than the blue rings @LWinterberg 

Maybe they could be a little taller (not full height of the meter) width looks OK I think


But if you enlarge the meter then the slider looks far too small as it does not enlarge with the meter:
image

Tantacrul:
I think these are looking a lot better than before. I'll let you and @jessjwilliamson confer a bit more. If you think this is what we should go with, you have my blessing!

Tantacrul (approved):

jessjwilliamson (approved):

petersampsonaudacity:
Testing on W10 with latest alpha master:  audacity-win-3.2.0-alpha-20220719+f084112-x64-msvc2022


The updated now slightly taller sliders look much better I think - and look better too in an enlarged meter.
Looks good to go to me.

Issue #1577, opened 2021-08-30
Scripting: "SetLabel:" fails when adding > 100 labels


SteveDaulton:
**Describe the bug**
When creating > 100 labels and adding label text, the  SetLabel: command fails after 100 labels.

**To Reproduce**
Run this script in the Nyquist Prompt:

```
;type tool
(setf number-of-labels 105)

(dotimes (i number-of-labels "Done")
  (aud-do (format nil "Select: Start=~s End=~s" i i))
  (aud-do "AddLabel:")
  (aud-do (format nil "SetLabel: Label=~s Text=~s" i i)))
```
Observe the error after 100 labels:
```
Set Label: Could not  load settings below. Default settings will be used.
Label=101 Text=101

```

**Expected behavior**
104 labels at 1 second intervals, numbered 0 to 103.

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version Audacity 3.1.0 and 3.0.2.

**Additional context**
Adding labels with scripting commands is slow.
Compare with running:
```
;type analyze
(setf number-of-labels 105)

(let (labels)
  (dotimes (i number-of-labels labels)
    (push (list i i (format nil "~a" i)) labels)))

```

SteveDaulton:
Another report of the same issue: https://forum.audacityteam.org/viewtopic.php?p=435189#p435189

SteveDaulton:
> I have had this exact issue, this time when using pipeclient.py as a CLI. Any way that I can help solve it?

Change line 43 in /src/SetLabelCommand.cpp
from:
`   S.Define(    mLabelIndex,                            wxT("Label"), 0, 0, 100 );`
to:
`   S.Define(    mLabelIndex,                            wxT("Label"), 0, 0, 10000 );`

Then rebuild Audacity.

PR #3249, opened 2022-07-18
Fix Issue 1577

SteveDaulton:
Resolves: Resolves: https://github.com/audacity/audacity/issues/1577

Increase maximum label number from 100 to 10000, which should be enough for practical purposes.

Recommended:

### DIFF
--- a/src/commands/SetLabelCommand.cpp
+++ b/src/commands/SetLabelCommand.cpp
@@ -40,7 +40,7 @@ SetLabelCommand::SetLabelCommand()
 
 template<bool Const>
 bool SetLabelCommand::VisitSettings( SettingsVisitorBase<Const> & S ){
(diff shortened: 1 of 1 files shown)


crsib:
I wonder if we need any kind of limit here at all, but this looks harmless enough

crsib (approved):

SteveDaulton:
> I wonder if we need any kind of limit here at all

I don't know why, but "SetLabel:" is rather slow. When there's a large number of labels, Audacity becomes unresponsive (there's no progress bar / cancel button) until it has finished (imo there should be, but that's another story). If there was no limit, then Audacity could be unresponsive for days / weeks / months.

Imo, 10,000 labels is a generous number that should satisfy virtually all practical cases, though it may take half an hour to complete.

PR #3257, opened 2022-07-19
Polymorphic ruler version 3

Paul-Licameli:

Recommended:


src/widgets/Ruler.h
```diff
@@ -62,6 +63,13 @@ class AUDACITY_DLL_API Ruler {
    // (at the center of the pixel, in both cases)
    void SetRange(double min, double max, double hiddenMin, double hiddenMax);
 
+   // Set the kind of updater the ruler will use
+   // (Linear, Logarithmic, Custom, etc.)
+   void SetUpdater(std::unique_ptr<Updater> pUpdater);
+
+   // An overload to replace SetUseZoomInfo
+   void SetUpdater(std::unique_ptr<Updater> pUpdater, int leftOffset);
```
Paul-Licameli, line 63: I see two overloads of `SetUpdater`.  Can they be collapsed into one, by use of a default argument for leftOffset?

Yes:  you might pass `int*` not `int` and default the argument to `nullptr`

But better than that -- avoiding funny looking function declarations that might leave the reviewer with doubts about  lifetime questions -- you can make the intent even clearer by passing `std::optional<int>` around, by-value.

src/AdornedRulerPanel.cpp
```diff
@@ -1282,9 +1282,9 @@ AdornedRulerPanel::AdornedRulerPanel(AudacityProject* project,
 
    mOuter = GetClientRect();
 
-   mRuler.SetUpdater( std::make_unique<LinearUpdater>( mRuler, mViewInfo ), mLeftOffset );
+   mRuler.SetUpdater( std::make_unique<LinearUpdater>( mViewInfo ), mLeftOffset );
```
Paul-Licameli, line 1285: I see that the second commit corrects several things I found doubtful in the first commit.

Good to correct them:  but better not to introduce them though at any point in the history.

Maybe the first two commits should be squashed (after rebbasing on a separate small commit for AudacityHeaders.h)

But then, make the separate, and prior, commit to define RulerStruct with all necessary changes in Ruler.cpp.  And then, rebase all onto that.

This will present a more reviewable sequence of changes.  Ruler.cpp will make a lot of necessary, "noisy" changes in a prior commit that is mostly just the noise.

Then the later commit separates the "noise" from the more interesting structural change.

src/widgets/Updater.cpp
```diff
@@ -20,8 +20,29 @@
 *//******************************************************************/
 
 #include "Updater.h"
+#include <wx/font.h>
+#include <wx/dc.h>
 
-Updater::TickSizes::TickSizes(double UPP, int orientation, Ruler::RulerFormat format, bool log)
+RulerStruct::RulerStruct()
+{
+   mMin = mHiddenMin = 0.0;
```
Paul-Licameli, line 32: The second commit introduces `RulerStruct` and this constructor.

This constructor has many assignments in it.

It's better C++ style (since 1998) to prefer member initialization where possible over assignments in the constructor body, for instance:

```
RulerStruct::RulerStruct()
   : mMin(0.0)
{
```

But since C++11, it's even better to prefer the "uniform initialization" syntax (curly braces) where possible:

```
RulerStruct::RulerStruct()
   : mMin{ 0.0 }
{
```

-- among other things, this disallows implicit narrowing convertions (as from float to integer).

But, also in C++11, and better still where it applies:  use in-class initialization.  That is, in Updater.h, declare the member as:

```
double mMin = 0.0;
```

or better

```
double mMin{ 0.0 };
```

or, even:

```
double mMin{};
```

Because 0 is the default value for all kinds of number types.

Any of those is better than
```
double mMin;
```

which implies no initialization.

When I review a class to understand it better, I want to start with the header file.  If I see in-class initializers, I don't have to read the constructor definitions to have confidence that members are never left uninitialized by mistake.  Rather I know that they default reasonably.  And then when I do read the constructor, it's a shorter constructor.

Only when there is no reasonable, argument-independent default for a member -- then, leave it without an initializer.  But then, I know that I must read the constructor body to understand fully what the member is for.

src/AdornedRulerPanel.cpp
```diff
@@ -1282,9 +1282,9 @@ AdornedRulerPanel::AdornedRulerPanel(AudacityProject* project,
...
    mRuler.SetLabelEdges( false );
-   mRuler.SetFormat( Ruler::TimeFormat );
+   mRuler.SetFormat( TimeFormat );
```
Paul-Licameli, line 1287: I confirm that at the second commit, you do not introduce dependency cycles among sourc code files.  👍

src/widgets/Updater.cpp
```diff
@@ -661,6 +661,9 @@ bool Updater::TickCustom(wxDC& dc, int labelIdx, wxFont font,
 }
 
 // Theoretically this helps
+
+Updater::~Updater() {}
+
```
Paul-Licameli, line 647: The third commit corrects the omission of the virtual destructor.  Good:  but, logically this should be squashed with the same commit that introduces 

```
std::unique_ptr<Updater> mpUpdater;
```

in Ruler.h, so the known mistake of missing default constructors is never even introduced into the commit history.

Besides that, I recommend:
- Define all of the destructors in this commit as non-inline, not only the one for Updater
- You can use the C++11 alternative syntax: `Updater::~Updater() = default`
- Also, now that I think of this too:  when you introduce the subclasses of `Updater`, make them `final` if possible, e.g. `struct LinearUpdater final : public Updater { ...` this means: you do not intend further subclassing of those classes.

What are my reasons for these practices?
- A linker has to put the virtual function table for each class in some unique place, and often implemented to follow this rule: the .cpp file definining the first non-inline virtual function generates the object file in which the table will live.  But, if there is non non-inline virtual function, then the compiler has fall back on generating multiple copies of the virtual function table that the linker just eliminates later.  Moral:  be nice to your linker, and avoid in-line virtuals.
- in-line optimization is inherently incompatible with virtual function dispatch anyway, so what do it ? ...
- except: when a class is `final`, and the declared object is of the leaf-class type, then the compiler can know that it really can to do the inline optimization, and may be smart enough to in-line default destructors.

src/widgets/RulerPanel.h
```diff
@@ -0,0 +1,82 @@
...
+  RulerPanel.h
+
+  Dominic Mazzoni
+
+**********************************************************************/
+
+#ifndef __AUDACITY_RULER_PANEL__
+#define __AUDACITY_RULER_PANEL__
+
+#include "Ruler.h"
```
Paul-Licameli, line 14: The fourth commit splits RulerPanel from Ruler.

Observe that including `Ruler.h` in `RulerPanel.h` is unavoidable, because there is a nested `Ruler` object.

Ok then -- but that means the transitive inclusion of `Ruler.h` happens whever `RulerPanel.h` is included.

There are four such inclusions, including that in `RulerPanel.cpp`.  I suggest then you simplify the lists of `#inlude` directives in each place.

src/widgets/Ruler.cpp
```diff
@@ -702,18 +702,3 @@ void Ruler::SetCustomMinorLabels(
    }
 }
 #endif
-
-void Updater::Label::Draw(wxDC&dc, bool twoTone, wxColour c) const
```
Paul-Licameli: The fifth commit moves this function.

I'd rather just squash that into the commit that introduces class `Updater`.  It's not recommended to have the non-inline definitions of a class's member functions in more than one file: so, avoid putting things into this state to begin with.

src/widgets/Updater.cpp
```diff
@@ -641,44 +641,6 @@ bool Updater::Tick(wxDC& dc,
    return !rect.IsEmpty();
 }
 
-bool Updater::TickCustom(wxDC& dc, int labelIdx, wxFont font,
-   // in/out:
-   TickOutputs outputs,
-   const RulerStruct& context) const
-{
```
Paul-Licameli: The last commit, too, might as well be squashed onto the commit that introduces Updater.

src/widgets/Updater.h
```diff
@@ -0,0 +1,84 @@
+﻿/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  Updater.h
+
+  Dominic Mazzoni
```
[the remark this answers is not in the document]
micpap25: I will change it to RulerUpdater and I'll update the credits when I do. No rush to change it.

src/widgets/Updater.h
```diff
@@ -0,0 +1,84 @@
...
+      int          mDigits;
+
+      TickSizes(double UPP, int orientation, Ruler::RulerFormat format, bool log);
+
+      TranslatableString LabelString(
+         double d, Ruler::RulerFormat format, const TranslatableString& units)
+         const;
+   };
+
+   double ComputeWarpedLength(const Envelope& env, double t0, double t1) const
```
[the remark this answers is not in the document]
micpap25: I will look through it

src/widgets/Ruler.h
micpap25: I have never heard of `std::optional<>`. I will look into it.

src/widgets/Ruler.cpp
```diff
@@ -108,14 +85,17 @@ Ruler::Ruler()
...
-
-   mCustom = false;
    mbMinor = true;
 
    mTwoTone = false;
 
    mUseZoomInfo = NULL;
+
+   // This part in particular needs inspection, not giving an error
+   // But is this corret? And should it be set to NULL or nullptr if a default
```
Paul-Licameli, line 95: typo: "correct"

src/widgets/Ruler.h
```diff
@@ -152,143 +144,43 @@ class AUDACITY_DLL_API Ruler {
...
    bool         mHasSetSpacing;
-   bool         mLabelEdges;
-   RulerFormat  mFormat;
-   bool         mLog;
-   bool         mFlip;
-   bool         mCustom;
    bool         mbMinor;
-   TranslatableString mUnits;
    bool         mTwoTone;
    const ZoomInfo *mUseZoomInfo;
```
Paul-Licameli, line 183: Reviewing the first again, I see the `Ruler::mUseZoomInfo` becomes unused at the first commit.  Reassigned in `Ruler::SetUpdater` but not used elsewhere for anything.  So eliminate it.

src/widgets/Updater.h
```diff
@@ -0,0 +1,144 @@
...
+
+   double mDbMirrorValue{ 0.0 };
+
+
+   mutable std::unique_ptr<Fonts> mpFonts;
+   TranslatableString mUnits;
+
+   int mLeftOffset;
+
+   NumberScale mNumberScale;
```
Paul-Licameli, line 61: Reviewing the commit "Remove circular dependencies..." which moves fields into the new type `RulerStruct` -- I see that the member `mLeftOffset` is never initialized.

A later commit removes the constructor in favor of in-class initializers, which is all right, but still does not correct this.

The remaining fields without in-class initializers are all right to remain so, because they are not primitive types.  So their default constructors are implicitly invoked.  You could write `{}` to initialize each but it isn't needed and makes no difference in compiled code.  I omit such initializers myself in such cases.

src/widgets/Ruler.h
```diff
@@ -12,6 +12,7 @@
 #define __AUDACITY_RULER__
 
 #include "wxPanelWrapper.h" // to inherit
+#include "Updater.h" // to inherit
```
Paul-Licameli, line 15: The comment is incorrect about the purpose of the inclusion

src/widgets/Ruler.cpp
```diff
@@ -67,6 +52,11 @@ array of Ruler::Label.
 #include "Theme.h"
 #include "ViewInfo.h"
 
+#include "Updater.h"
+// Need to include to set default
+#include "LinearUpdater.h"
+#include "LogarithmicUpdater.h"
```
Paul-Licameli, line 58: The commit that separates RulerPanel.cpp can also remove this `#include` which becomes unnecessary.

src/commands/SetTrackInfoCommand.cpp
```diff
@@ -286,17 +286,20 @@ static const EnumValueSymbol kColourStrings[nColours] =
...
+   kLogarithmicDb,
+   kLinearDb,
    nScaleTypes
 };
 
 static const EnumValueSymbol kScaleTypeStrings[nScaleTypes] =
 {
    // These are acceptable dual purpose internal/visible names
-   { XO("Linear") },
+   { XO("Linear (amp)") },
```
Paul-Licameli, line 298: This changed line should be preceded by an `i18n-hint` comment explaining that "amp" abbreviates "amplitude".

Those comments matter because `xgettext` extracts them to `locale/audacity.pot` when we update that file, at least once in each development cycle.  Then the comments are copied to the several `locale/*.po` files, and serve as clues for the volunteer translators who update the string catalogs for other languages.

Paul-Licameli: See `scripts/update_po_files.sh` if you are curious about that process.

Issue #3206, opened 2022-07-11
mpg123 reports a fail as success, causing FFMPEG-imports to fail silently if FFMPEG is not present


LWinterberg:
"fail silently" meaning that you just get put in a state as if you never attempted to import anything.

Log: 
```
13:22:28: Audacity 3.2.0-alpha-20220711
13:22:28: sqlite3 message: (1) no such table: project in "SELECT 1 FROM project LIMIT 1;"
13:22:29: Error: Failed to load shared library 'avformat-59.dll' (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:29: Error: Failed to load avformat-59.dll (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:29: Error: Failed to load shared library 'avformat-58.dll' (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:29: Error: Failed to load avformat-58.dll (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:29: Error: Failed to load shared library 'avformat-57.dll' (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:29: Error: Failed to load avformat-57.dll (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:29: Error: Failed to load shared library 'avformat-55.dll' (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:29: Error: Failed to load avformat-55.dll (error 126: Das angegebene Modul wurde nicht gefunden.)
// ^all of the above say "module not found"

13:22:45: File name is C:\Users\mc\Documents\Audacity\Rick Astley - Never Gonna Give You Up.m4a
13:22:45: Mime type is *
13:22:45: Opening with libav
13:22:45: Error: Failed to load shared library 'avformat-59.dll' (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:45: Error: Failed to load avformat-59.dll (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:45: Error: Failed to load shared library 'avformat-58.dll' (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:45: Error: Failed to load avformat-58.dll (error 126: Das angegebene Modul wurde nicht gefunden.)
13:22:45: Error: Failed to load shared library 'avformat-57.dll' (error 126: Das angegebene Modul wurde nicht gefunden.)
[...]

PR #3265, opened 2022-07-20
Fixes over optimistic libmpg123 behavior

crsib:
Resolves: #3206 

This PR attempts to:
1. Improve the detection of the MP3 files on `Open`. This is can be tricky as MP3 has no container header.
2. Prevent Frankenstein MP3 files from importing. Such files are extremely rare. Probably FFMPEG based importer will be able to handle them
3. Fix the case, where `Open` succeeded, but `Import` failed. This can happen surprisingly often, because it is hard to detect MP3 files

Recommended:

### DIFF
--- a/src/import/ImportMP3_MPG123.cpp
+++ b/src/import/ImportMP3_MPG123.cpp
@@ -133,6 +133,8 @@ public:
    ByteCount GetFileUncompressedBytes() override;
    ProgressResult Import(WaveTrackFactory *trackFactory, TrackHolders &outTracks, Tags *tags) override;
 
+   bool SetupOutputFormat();
+
    void ReadTags(Tags* tags);
 
    wxInt32 GetStreamCount() override;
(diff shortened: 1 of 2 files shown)


src/import/ImportMP3_MPG123.cpp
```diff
@@ -374,6 +348,31 @@ ProgressResult MP3ImportFileHandle::Import(WaveTrackFactory *trackFactory,
    return mUpdateResult;
 }
 
+bool MP3ImportFileHandle::SetupOutputFormat()
+{
+   long rate;
+   int channels;
+   int encoding = MPG123_ENC_FLOAT_32;
+   mpg123_getformat(mHandle, &rate, &channels, &encoding);
```
vsverchinsky, line 356: This funcition has a return value and can fail, is that a redundant check?

vsverchinsky (approved):

src/import/ImportMP3_MPG123.cpp
crsib: We call mpg123_getformat after parsing the first frame, so it will not fail as per the source and docs.

Issue #3267, opened 2022-07-20
Many swh (Ladspa and lv2) plug-ins don't persist correctly


Paul-Licameli:
In fact, any effect with an author name containing <> characters fails to persist.

- Install swh Ladspa plug-ins
- generate some sound in a new project
- add any effect, for which the author name includes an email address in `<angle brackets>`, to the track in the sidebar
- save the project; close it; reopen it; reopen the sidebar
- Observe: the effect is disabled and the cog wheel is gray

dozzzzer:
Verified on Windows 11.

Several plugins seem to not have any controls to adjust, so for them no dialog window can be initiated when clicking the cog wheel button.

These are:
- dc_remove_1207.xml
- declip_1195.xml
- inv_1429.xml
- matrix_st_ms_1420.xml
- split_1406.xml
- surround_encoder_1401.xml
- wave_terrain_1412.xml
- zm1_1428.xml

PR #3268, opened 2022-07-20
Fix sidebar persistency: don't doubly escape strings for XML...

Paul-Licameli:
... See XMLWriter::WriteAttr -- it already applies one layer of escapes.

Revert d2ee5908a; more, make the function private to discourage misuse.

Resolves: #3267

Recommended:


libraries/lib-xml/XMLWriter.h
```diff
@@ -56,9 +56,10 @@ class XML_API XMLWriter /* not final */ {
 
    virtual void Write(const wxString &data) = 0;
 
+ private:
    // Escape a string, replacing certain characters with their
    // XML encoding, i.e. '<' becomes '&lt;'
-   static wxString XMLEsc(const wxString & s);
+   wxString XMLEsc(const wxString & s);
```
crsib, line 62: I'm not sure if I understand why static was removed. I do not see any changes to the function, so I see little reason why it is no longer static or at least why is it not const.

Paul-Licameli: It isn't static in 3.1.  Static was inserted by @lllucius in the prototype he left us only to implement new effect stacks, and I cherry-picked that part months ago.  Now with better understanding, I see it as unnecessary for correct implementation.

crsib: You miss my point. Does this function requires `this`? Does this function modifies `this`? At least it should be `const`, but I would prefer it to be `static`. In this case it can be removed from the class altogether.

Paul-Licameli: Okay, I'll leave it as static, but I think it should be private too.

crsib: I have no objections on `private`

crsib:
I can approve this PR as I see nothing wrong with it, unless you want specifically Kit to review it :-)

crsib (approved):

ksoze95 (approved):

Issue #3192, opened 2022-07-07
define the instance class for VST2 effects


Paul-Licameli:
The Instance might be done before the Validator.  That is not how things progressed with built-ins.  But if you look at the dependency graphs for the source files implementing AudioUnits or LV2, you see that the validator depends on the instance, not the opposite.

So really the Instance is the more basic object, which could be extracted from existing code first, although it may also be necessary to define the Validator partially so that Instance can be defined completely.

Suggested outline of steps, by analogy with the recent LV2 transformation

- [x] 1. (As needed, perhaps not) Definine a minimal Validator class (see 72f3c002)
- [x] 2. Define minimal Instance class (8ec84fa7)
- [x] 3. Define temporary overloads of `GetSettings` in scope of the Instance (e6d6ed95)
- [x] 4. Instance doesn't inherit `StatefulPerTrackEffect::Instance`; move `ProcessInitialize`, `ProcessBlock`, `ProcessFinalize`, `GetLatency` from Effect to Instance class (ffbec4a8); and effect inherits `PerTrackEffect` and `StatefulEffectBase` but not `StatefulPerTrackEffect`
- [ ] 5. Instance doesn't inherit `StatefulEffectBase::Instance`; move other necessary overrides into Instance class; make slaves instances (76028f62); effect no longer inherits `StatefulEffectBase`
- [x] 6. (Only needed if output metering doesn't work:  Instance re-uses itself, not a slave, for group number 0 in RealtimeProcess (3447e35d))

pietro68:
Looks like output metering already works, so I will cross out number 6.

PR #3271, opened 2022-07-21
Define instance vst2

pietro68:
Resolves: part of https://github.com/audacity/audacity/issues/3192

Before even going through the steps suggested at the issue above, it seemed necessary to bring some important functionality in the wrapper class: the vst callback function, and the part of the original ::Load method which makes sense to be executed when a new instance will be created. Take into consideration that VstEffectInstance will inherit from VstEffectWrapper.

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1784,7 +1784,7 @@ bool VSTEffect::LoadCommon()
 }
 
 
-bool VSTEffect::Load()
+bool VSTEffectWrapper::Load()
 {
    bool success = false;
 
@@ -1807,7 +1807,7 @@ bool VSTEffect::Load()
       // Note:  Some hosts use "user" and some use "ptr2/resvd2".  It might
       //        be worthwhile to check if user is NULL before using it and
       //        then falling back to "ptr2/resvd2".
-      mAEffect->ptr2 = static_cast<VSTEffectWrapper*>(this);
+      mAEffect->ptr2 = this;
 
       // Give the plugin an initial sample rate and blocksize
       callDispatcher(effSetSampleRate, 0, 0, NULL, 48000.0);
@@ -1817,7 +1817,13 @@ bool VSTEffect::Load()
       // I've found one plugin (SoundHack +morphfilter) that will
       // crash Audacity when saving the initial default parameters
       // with this.
-      callSetProgram(0);
+      //
+      // The following 3 calls to callDispatcher are equivalent to callSetProgram(0),
+      // only done on the mAEFfect handle and not on the slaves too, if present
+      //
+      callDispatcher(effBeginSetProgram, 0, 0, NULL, 0.0);
+      callDispatcher(effSetProgram,      0, 0, NULL, 0.0);
+      callDispatcher(effEndSetProgram,   0, 0, NULL, 0.0);
 
       // Pretty confident that we're good to go
       success = true;
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -802,10 +802,11 @@ VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
    mTimeInfo.timeSigDenominator = 4;
    mTimeInfo.flags = kVstTempoValid | kVstNanosValid;
 
-   // If we're a slave then go ahead a load immediately
+   // If we're a slave then go ahead and load immediately
    if (mMaster)
    {
-      Load();
+      if ( ! Load() )
+         Unload();
    }
 }
 
@@ -918,7 +919,8 @@ bool VSTEffect::InitializePlugin()
 {
    if (!mAEffect)
    {
-      Load();
+      if ( ! Load() )
+         Unload();
    }
 
    if (!mAEffect)
@@ -1819,11 +1821,6 @@ bool VSTEffect::Load()
 
       // Pretty confident that we're good to go
       success = true;
-   }
-
-   if (!success)
-   {
-      Unload();
    }
 
    return success;
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -802,6 +802,10 @@ VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
    mTimeInfo.timeSigDenominator = 4;
    mTimeInfo.flags = kVstTempoValid | kVstNanosValid;
 
+   // Get the constant data that are shared by all instances
+   // (e.g. entry point in the dll, vst version, no. of channels etc.)
+   LoadCommon();
+
    // If we're a slave then go ahead and load immediately
    if (mMaster)
    {
@@ -1782,13 +1786,9 @@ bool VSTEffect::LoadCommon()
 
 bool VSTEffect::Load()
 {
-   // TODO: move this call out of here, so that it is executed only once
-   if ( ! LoadCommon() )
-      return false;
-
    bool success = false;
 
-   // Initialize the plugin
+   // Create the handle to the plugin, also setting the vst callback function
    try
    {
       mAEffect = mPluginMain(VSTEffectWrapper::AudioMaster);
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1693,13 +1693,94 @@ bool VSTEffect::LoadCommon()
 
 #endif
 
-   return true;
+   AEffect* tempHandle;
+
+   try
+   {
+      tempHandle = mPluginMain(VSTEffectWrapper::AudioMaster);
+   }
+   catch (...)
+   {
+      wxLogMessage(wxT("VST plugin initialization failed\n"));
+      return false;
+   }
+
+   bool success = false;
+
+   // Ask the plugin to identify itself...might be needed for older plugins
+   callDispatcher(tempHandle, effIdentify, 0, 0, NULL, 0);
+
+   // Open a temporary plugin handle, to get some info on the plugin
+   callDispatcher(tempHandle, effOpen, 0, 0, NULL, 0.0);
+
+   // Get the VST version the plugin understands
+   mVstVersion = callDispatcher(tempHandle, effGetVstVersion, 0, 0, NULL, 0);
+
+   // Ensure that it looks like a plugin and can deal with ProcessReplacing
+   // calls.  Also exclude synths for now.
+   if (   tempHandle->magic == kEffectMagic &&
+        !(tempHandle->flags & effFlagsIsSynth) &&
+          tempHandle->flags & effFlagsCanReplacing)
+   {
+      if (mVstVersion >= 2)
+      {
+         mName = GetStringFromHandle(effGetEffectName, tempHandle);
+         if (mName.length() == 0)
+         {
+            mName = GetStringFromHandle(effGetProductString, tempHandle);
+         }
+      }
+      if (mName.length() == 0)
+      {
+         mName = wxFileName{ mRealPath }.GetName();
+      }
+
+      if (mVstVersion >= 2)
+      {
+         mVendor = GetStringFromHandle(effGetVendorString, tempHandle);
+         mVersion = wxINT32_SWAP_ON_LE(callDispatcher(tempHandle, effGetVendorVersion, 0, 0, NULL, 0));
+      }
+      if (mVersion == 0)
+      {
+         mVersion = wxINT32_SWAP_ON_LE(tempHandle->version);
+      }
+
+      if (tempHandle->flags & effFlagsHasEditor || tempHandle->numParams != 0)
+      {
+         mInteractive = true;
+      }
+
+      mAudioIns = tempHandle->numInputs;
+      mAudioOuts = tempHandle->numOutputs;
+
+      mMidiIns = 0;
+      mMidiOuts = 0;
+
+      // Check to see if parameters can be automated.  This isn't a guarantee
+      // since it could be that the effect simply doesn't support the opcode.
+      mAutomatable = false;
+      for (int i = 0; i < tempHandle->numParams; i++)
+      {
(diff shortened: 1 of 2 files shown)


src/effects/VST/VSTEffect.h
```diff
@@ -199,6 +204,33 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+   static intptr_t AudioMaster(AEffect *effect,
+                               int32_t opcode,
+                               int32_t index,
+                               intptr_t value,
+                               void * ptr,
+                               float opt);
+
+   // These are needed to move the AudioMaster callback from VSTEffect to this struct
+   //
+   static intptr_t mCurrentEffectID;
```
Paul-Licameli, line 217: I see why this moved member ever needed to be static.  Make it non-static and initialize it with `{}`

src/effects/VST/VSTEffect.cpp
```diff
@@ -603,18 +603,18 @@ BEGIN_EVENT_TABLE(VSTEffect, wxEvtHandler)
...
-                                float opt)
+intptr_t VSTEffectWrapper::AudioMaster(AEffect * effect,
+                                       int32_t opcode,
+                                       int32_t index,
+                                       intptr_t value,
+                                       void * ptr,
+                                       float opt)
 {
-   VSTEffect *vst = (effect ? (VSTEffect *) effect->ptr2 : NULL);
+   VSTEffectWrapper *vst = (effect ? (VSTEffectWrapper *) effect->ptr2 : NULL);
```
Paul-Licameli, line 617: Modernize with `static_cast` and `nullptr`

src/effects/VST/VSTEffect.cpp
```diff
@@ -1547,15 +1553,14 @@ void VSTEffect::ShowOptions()
...
+// Load the VST things that can be shared among all instances
+//------------------------------------------------------------------------------
+bool VSTEffect::LoadCommon()
 {
-   vstPluginMain pluginMain;
-   bool success = false;
-
    long effectID = 0;
-   wxString realPath = mPath.BeforeFirst(wxT(';'));
+   mRealPath = mPath.BeforeFirst(wxT(';'));
```
Paul-Licameli, line 1561: Second commit needs to change `realPath` also in the mac-only lines below

src/effects/VST/VSTEffect.cpp
```diff
@@ -1631,20 +1636,20 @@ bool VSTEffect::Load()
...
          return false;
 
       // Bail if it wasn't successful
       if (!lib->IsLoaded())
          return false;
 
       // Try to find the entry point, while suppressing error messages
-      pluginMain = (vstPluginMain) lib->GetSymbol(wxT("VSTPluginMain"));
-      if (pluginMain == NULL)
+      mPluginMain = (vstPluginMain)lib->GetSymbol(wxT("VSTPluginMain"));
```
Paul-Licameli, line 1648: Mac build also requires `mPluginMain` in lines above

Paul-Licameli: Linux compilation will also need a fix in the second commit.

src/effects/VST/VSTEffect.h
```diff
@@ -199,6 +204,33 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+   virtual void UpdateDisplay() = 0;
+   virtual VstTimeInfo* GetTimeInfo() = 0;
+   virtual void SetBufferDelay(int samples) = 0;
+   virtual float GetSampleRate() = 0;
+   virtual int GetProcessLevel() = 0;
+   virtual void SizeWindow(int w, int h) = 0;
+   virtual void Automate(int index, float value) = 0;
+
+   // Loaded once when getting common VST things;
+   // placed here becaue they will be reused when an instance is created
```
Paul-Licameli, line 228: "because"

src/effects/VST/VSTEffect.h
```diff
@@ -199,6 +204,33 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+   //
+   vstPluginMain mPluginMain;   // entry point in the plugin dll
```
Paul-Licameli, line 230: This is just a "C" pointer to function -- so initialize it to null

src/effects/VST/VSTEffect.cpp
```diff
@@ -1631,20 +1636,20 @@ bool VSTEffect::Load()
...
+      if (mPluginMain == NULL)
       {
-         pluginMain = (vstPluginMain) lib->GetSymbol(wxT("main"));
-         if (pluginMain == NULL)
+         mPluginMain = (vstPluginMain)lib->GetSymbol(wxT("main"));
```
Paul-Licameli, line 1651: You can modernize more casts and NULLs

src/effects/VST/VSTEffect.cpp
```diff
@@ -2147,6 +2193,16 @@ intptr_t VSTEffectWrapper::callDispatcher(int opcode,
    return mAEffect->dispatcher(mAEffect, opcode, index, value, ptr, opt);
 }
 
+intptr_t VSTEffectWrapper::callDispatcher(
+   AEffect* handle,
+   int opcode, int index, intptr_t value, void* ptr, float opt)
+{
+   // Needed since we might be in the dispatcher when the timer pops
+   wxCRIT_SECT_LOCKER(locker, mDispatcherLock);
```
Paul-Licameli, line 2201: If this function is redefined as `static`, which I recommend, then `mDispatcherLock` must become an argument, passed by reference.

And you can also modernize by using `std::mutex` instead of the wx synchronization classes.

Paul-Licameli: Here's a better idea for splitting Load, phase 2.

Define a small structure holding mAEffect and the mutex.  Make `callDispatcher` a member function of it that locks the mutex.  Make its destructor call the dispatcher with `effClose`, replacing the two places where you do that explicitly.

You can also change the existing `mAEffect` member to be of this structure type, and define an `operator ->`, which will let you avoid changing many lines of code.

src/effects/VST/VSTEffect.cpp
```diff
@@ -2131,6 +2160,23 @@ wxString VSTEffectWrapper::GetString(int opcode, int index) const
    return str;
 }
 
+wxString VSTEffectWrapper::GetStringFromHandle(int opcode, AEffect* handle) const
```
Paul-Licameli, line 2163: `GetStringFromHandle` can become a static, nonmember function if `handle` is of the new structure type.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1695,13 +1695,94 @@ bool VSTEffect::LoadCommon()
...
    return true;
 }
 
 
 bool VSTEffect::Load()
 {
-   // TODO: move this out of here, so that it can be called only once
+   // TODO: move this call out of here, so that it is executed only once
    if ( ! LoadCommon() )
       return false;
```
Paul-Licameli: Here's another idea.  Change the return type of `LoadCommon` to be the new structure.  Give it an `explicit operator bool()` returning false when it contains a null pointer, and return `{}` for failure.

Define move construction and assignment for the new type, to transfer the pointer to the left hand side and leave it null on the right.

Do all of this -- and the loading of the plug-in can avoid destroying one AEffect only to build another one at once.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1695,13 +1695,94 @@ bool VSTEffect::LoadCommon()
...
+         if (callDispatcher(tempHandle, effCanBeAutomated, 0, i, NULL, 0.0))
+         {
+            mAutomatable = true;
+            break;
+         }
+      }
+   }
+   else
+   {
+      return false;
```
Paul-Licameli, line 1779: This was an error path that did not balance `effOpen` and `effClose`.  See how the smart pointer structure would make this easy to ensure.

src/effects/VST/VSTEffect.cpp
```diff
@@ -920,7 +925,8 @@ bool VSTEffect::InitializePlugin()
 {
    if (!mAEffect)
    {
-      Load();
+      if ( ! Load() )
+         Unload();
```
Paul-Licameli, line 929: This may improve error checking, causing the function to return false of load failes for some other reason after constructing `mAEffect` -- though I see no such reasons now.

I reviewed where else this function is called and I see `VSTEffectsModule::LoadPlugin` does not check the status.  So change that function to check and return null for failure.

src/effects/VST/VSTEffect.cpp
```diff
@@ -804,10 +804,15 @@ VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
...
+   // Get the constant data that are shared by all instances
+   // (e.g. entry point in the dll, vst version, no. of channels etc.)
+   LoadCommon();
+
+   // If we're a slave then go ahead and load immediately
    if (mMaster)
    {
-      Load();
+      if ( ! Load() )
+         Unload();
```
Paul-Licameli, line 815: So, construction of a `VSTEffect` may leave it in an unusable state if load fails.  Where should that failure be checked?  Where `make_unique<VSTEffect>` happens, and one place is `LoadPlugin`, just mentioned, the other in `VSTEffect::RealtimeAddProcessor`.

I don't yet see if you add a check there and return false, but I'm still marching through this commit sequence.

src/effects/VST/VSTEffect.cpp
```diff
@@ -804,10 +804,15 @@ VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
    mTimeInfo.timeSigDenominator = 4;
    mTimeInfo.flags = kVstTempoValid | kVstNanosValid;
 
-   // If we're a slave then go ahead a load immediately
+   // Get the constant data that are shared by all instances
+   // (e.g. entry point in the dll, vst version, no. of channels etc.)
+   LoadCommon();
```
Paul-Licameli, line 809: When `LoadCommon` is lifted out of `Load`, then you might `move` its return value into a new argument of `Load` to transfer ownership of the handle.

src/effects/VST/VSTEffect.cpp
Paul-Licameli: Now I see that `Load` is not always preceded by `LoadCommon` in the later commits.
But that can't possibly work, can it?  `Load` assumes `mPluginMain` to be inititalized to non-null, but `LoadCommon` is where that happens.

Paul-Licameli:
I have examined every commit in the sequence and I'm almost done with review of this draft.  Now, general comments.

src/effects/VST/VSTEffect.cpp
pietro68: ok, done

src/effects/VST/VSTEffect.h
pietro68: ok

Paul-Licameli:
Moving `Load` without `LoadCommon` into the wrapper seems wrong.  Why were you trying to do it?  I think the answer is that you wanted to avoid moving some data members into the wrapper.  But I don't think it is necessary to avoid that.

The wrapper class contains an AEffect handle, and that is right.  Each instance needs a different handle, and there is another scratch handle for the plug-in itself.

The later part of `LoadCommon` assigns mVstVersion, mName, mVendor, mInteractive, mAudioIns, mAudioOuts, mMidiIns, mMidiOuts, and mAutomatable.  These things are all "metadata" about the plug-in itself, not state of instances of plug-ins.

The first two are already in the wrapper class, but now you skip the assignment of them for the plug-in -- another reason this PR is in error.  For the rest, there is no reason why they can't be in the wrapper too, like mName.

What about the early part of LoadCommon?

It assigns mRealPath, mCurrentEffectID, mPluginMain in the wrapper, and mPath, mBundleRef, mResource, mModule in the effect.  The latter things are (platform specific) resource handles that keep the shared library of foreign code loaded into the process for as long as necessary.

Again, either you miss the assignment of things that were already in the wrapper, or else, you avoid moving things into the wrapper that are not part of Instance state.

So, I simply don't see your purpose of the splitting of Load().

Keep this work:  there could be this benefit of this reexamination of Load(), though I don't think it is necessary for VST2 statelessness.

The things assigned inside LoadCommon() really need to be assigned only once.  They could be moved into a structure managed by a `shared_ptr` and that pointer can be in the Wrapper.  The Effect object does the work of LoadCommon only once.  When the Effect makes an Instance, or the Instance makes a slave, then the shared pointer can simply be copied, and the discovery of things inside LoadCommon would not need to be repeated.

src/effects/VST/VSTEffect.cpp
pietro68: right, done

src/effects/VST/VSTEffect.cpp
pietro68: right, done

src/effects/VST/VSTEffect.h
pietro68: oops - done

src/effects/VST/VSTEffect.h
pietro68: ok

src/effects/VST/VSTEffect.h
Paul-Licameli: Can’t you also rewrite the case in AudioMaster to use `vst->mCurrentEffectID`?

pietro68: That's right, then it is possible - done.

src/effects/VST/VSTEffect.cpp
```diff
@@ -2147,6 +2193,16 @@ intptr_t VSTEffectWrapper::callDispatcher(int opcode,
    return mAEffect->dispatcher(mAEffect, opcode, index, value, ptr, opt);
```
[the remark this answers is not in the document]
pietro68: ok, done

PR #3276, opened 2022-07-22
Pop event handlers

Paul-Licameli:
Popping of event handlers on the stack of the effect dialog is done in destructors
of DefaultEffectUIValidator (as was already done for LV2 validators).

The validator makes sense then as a cleanup object holding a responsibility.

All uses of PushEventHandler in the entire program are easily reviewed, to be sure the change in Effect.cpp (which pushes
event handlers less often) is not leaving any existing PopEventHandler or RemoveEventHandler calls unbalanced.

Recommended:


src/effects/Effect.h
```diff
@@ -25,24 +25,33 @@ class WaveTrack;
...
-/*!
+/*
+ Also pops the even handler stack of a window, if given to the contructor
 
  This is a transitional class; it should be eliminated when all effect classes
  define their own associated subclasses of EffectUIValidator, which can hold
  state only for the lifetime of a dialog, so the effect object need not hold it
 */
-class DefaultEffectUIValidator
+class COMPONENTS_API DefaultEffectUIValidator
```
crsib, line 35: `COMPONENTS_API` seems to be wrong here. I would've expect `AUDACITY_API`

crsib: Yep, it breaks Windows build

crsib (changes requested):
Please replace `COMPONENTS_API ` with `AUDACITY_API` (or remove it at all)

src/effects/Effect.h
Paul-Licameli: Oh heck

crsib (approved):

PR #3278, opened 2022-07-22
Define instance vst2 2nd

pietro68:
Resolves: https://github.com/audacity/audacity/issues/3192

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -948,7 +948,8 @@ std::shared_ptr<EffectInstance> VSTEffect::DoMakeInstance()
    GetConfig(*this, PluginSettings::Shared, wxT("Options"),
       wxT("UseLatency"), mUseLatency, true);
    mBlockSize = mUserBlockSize;
-   return std::make_shared<Instance>(*this);
+
+   return std::make_shared<VSTEffectInstance>(*this);
 }
 
 unsigned VSTEffect::GetAudioInCount() const
@@ -1298,7 +1299,7 @@ bool VSTEffect::DoLoadFactoryPreset(int id)
 // ============================================================================
 
 std::unique_ptr<EffectUIValidator> VSTEffect::PopulateUI(ShuttleGui &S,
-   EffectInstance &, EffectSettingsAccess &access)
+   EffectInstance& instance, EffectSettingsAccess &access)
 {
    auto parent = S.GetParent();
    mDialog = static_cast<wxDialog *>(wxGetTopLevelParent(parent));
@@ -1330,7 +1331,7 @@ std::unique_ptr<EffectUIValidator> VSTEffect::PopulateUI(ShuttleGui &S,
    auto pParent = S.GetParent();
    pParent->PushEventHandler(this);
 
-   return std::make_unique<VSTEffectValidator>(*this, access, pParent);
+   return std::make_unique<VSTEffectValidator>(dynamic_cast<VSTEffectInstance&>(instance), *this, access, pParent);
 }
 
 bool VSTEffect::IsGraphicalUI()
@@ -3594,6 +3595,18 @@ EffectSettings VSTEffect::MakeSettings() const
 
 VSTEffectValidator::~VSTEffectValidator() = default;
 
+
+VSTEffectValidator::VSTEffectValidator
+(
+   VSTEffectInstance&       instance,
+   EffectUIClientInterface& effect,
+   EffectSettingsAccess&    access,
+   wxWindow*                pParent
+)
+   : DefaultEffectUIValidator(effect, access, pParent),
+     mInstance(instance)
+{}
+
 // Default, do-nothing implementations of virtuals in VSTEffectWrapper
 void VSTEffectWrapper::NeedIdle()
 {   
@@ -3629,6 +3642,12 @@ void VSTEffectWrapper::SizeWindow(int w, int h)
 
 void VSTEffectWrapper::Automate(int index, float value)
 {
+}
+
+
+VSTEffectInstance::~VSTEffectInstance()
+{
+   
 }
 
 
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -603,14 +603,14 @@ END_EVENT_TABLE()
 
 typedef AEffect *(*vstPluginMain)(audioMasterCallback audioMaster);
 
-intptr_t VSTEffectWrapper::AudioMaster(AEffect * effect,
+intptr_t VSTInstanceBase::AudioMaster( AEffect * effect,
                                        int32_t opcode,
                                        int32_t index,
                                        intptr_t value,
                                        void * ptr,
                                        float opt)
 {
-   VSTEffectWrapper *vst = (effect ? static_cast<VSTEffectWrapper*>(effect->ptr2) : nullptr);
+   VSTInstanceBase* vst = (effect ? static_cast<VSTInstanceBase*>(effect->ptr2) : nullptr);
 
    // Handles operations during initialization...before VSTEffect has had a
    // chance to set its instance pointer.
@@ -786,7 +786,7 @@ void VSTEffect::ResourceHandle::reset()
 #endif
 
 VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
-:  VSTEffectWrapper(path),
+:  VSTInstanceBase(path),
    mMaster(master)
 {
    mTimer = std::make_unique<VSTEffectTimer>(this);
@@ -948,8 +948,8 @@ std::shared_ptr<EffectInstance> VSTEffect::DoMakeInstance()
    GetConfig(*this, PluginSettings::Shared, wxT("Options"),
       wxT("UseLatency"), mUseLatency, true);
    mBlockSize = mUserBlockSize;
-
-   return std::make_shared<VSTEffectInstance>(*this);
+ 
+   return std::make_shared<VSTEffectInstance>(*this, mPath, mBlockSize, mUserBlockSize, mUseLatency);
 }
 
 unsigned VSTEffect::GetAudioInCount() const
@@ -973,25 +973,27 @@ size_t VSTEffect::GetBlockSize() const
    return mBlockSize;
 }
 
-sampleCount VSTEffect::GetLatency() const
+sampleCount VSTEffectInstance::GetLatency(const EffectSettings& settings, double sampleRate) const
 {
    if (mUseLatency)
       return mBufferDelay;
    return 0;
 }
 
-bool VSTEffect::IsReady()
+bool VSTEffectInstance::IsReady()
 {
    return mReady;
 }
 
-bool VSTEffect::ProcessInitialize(
-   EffectSettings &, double sampleRate, ChannelNames)
+bool VSTEffectInstance::ProcessInitialize(
+   EffectSettings& settings, double sampleRate, ChannelNames)
 {
+   StoreSettings(GetSettings(settings));
+
    return DoProcessInitialize(sampleRate);
 }
 
-bool VSTEffect::DoProcessInitialize(double sampleRate)
+bool VSTInstanceBase::DoProcessInitialize(double sampleRate)
 {
    // Initialize time info
    memset(&mTimeInfo, 0, sizeof(mTimeInfo));
@@ -1016,7 +1018,7 @@ bool VSTEffect::DoProcessInitialize(double sampleRate)
    return true;
 }
 
-bool VSTEffect::ProcessFinalize() noexcept
+bool VSTInstanceBase::DoProcessFinalize() noexcept
 {
 return GuardedCall<bool>([&]{
    mReady = false;
@@ -1027,7 +1029,14 @@ return GuardedCall<bool>([&]{
 });
 }
 
-size_t VSTEffect::ProcessBlock(EffectSettings &,
+
+bool VSTEffectInstance::ProcessFinalize() noexcept
+{
+   return VSTInstanceBase::DoProcessFinalize();
+}
+
+
+size_t VSTEffectInstance::ProcessBlock(EffectSettings &,
    const float *const *inBlock, float *const *outBlock, size_t blockLen)
 {
    // Only call the effect if there's something to do...some do not like zero-length block
@@ -1043,45 +1052,20 @@ size_t VSTEffect::ProcessBlock(EffectSettings &,
    return blockLen;
 }
 
-unsigned VSTEffect::GetChannelCount()
-{
-   return mNumChannels;
-}
-
-void VSTEffect::SetChannelCount(unsigned numChannels)
-{
-   mNumChannels = numChannels;
-}
 
 bool VSTEffect::RealtimeInitialize(EffectSettings &settings, double sampleRate)
 {
-   return ProcessInitialize(settings, sampleRate, nullptr);
+   return DoProcessInitialize(sampleRate);
 }
 
 bool VSTEffect::RealtimeAddProcessor(
    EffectSettings &settings, unsigned numChannels, float sampleRate)
 {
-   auto slave = std::make_unique<VSTEffect>(mPath, this);
+   auto slave = std::make_unique<VSTEffectInstance>(*this, mPath, mBlockSize, mUserBlockSize, mUseLatency);
 
    slave->SetBlockSize(mBlockSize);
-   slave->SetChannelCount(numChannels);
 
-   int clen = 0;
-   if (mAEffect->flags & effFlagsProgramChunks) {
-      void *chunk = NULL;
-      clen = (int) callDispatcher(effGetChunk, 1, 0, &chunk, 0.0); // get master's chunk, for the program only
-      if (clen != 0)
-         slave->callSetChunk(true, clen, chunk); // copy state to slave, for the program only
-   }
-
-   if (clen == 0) {
-      callDispatcher(effBeginSetProgram, 0, 0, NULL, 0.0);
-      for (int i = 0; i < mAEffect->numParams; i++)
-         slave->callSetParameter(i, callGetParameter(i));
-      callDispatcher(effEndSetProgram, 0, 0, NULL, 0.0);
-   }
-
-   if (!slave->ProcessInitialize(settings, sampleRate, nullptr))
+   if (!slave->ProcessInitialize(settings, sampleRate, ChannelNames()))
       return false;
 
    mSlaves.emplace_back(move(slave));
@@ -1092,10 +1076,10 @@ bool VSTEffect::RealtimeFinalize(EffectSettings &) noexcept
 {
 return GuardedCall<bool>([&]{
    for (const auto &slave : mSlaves)
-      slave->ProcessFinalize();
+      slave->DoProcessFinalize();
    mSlaves.clear();
 
-   return ProcessFinalize();
+   return DoProcessFinalize();
 });
 }
 
@@ -1165,16 +1149,18 @@ bool VSTEffect::RealtimeProcessEnd(EffectSettings &) noexcept
 /// all provide the information (kn0ck0ut is one).
 ///
 int VSTEffect::ShowClientInterface(
-   wxWindow &parent, wxDialog &dialog, EffectUIValidator *, bool forceModal)
+   wxWindow &parent, wxDialog &dialog, EffectUIValidator* validator, bool forceModal)
 {
    //   mProcessLevel = 1;      // in GUI thread
 
-   if (!IsReady())
+   VSTEffectValidator* vstValidator = static_cast<VSTEffectValidator*>(validator);
+
+   if (! vstValidator->GetInstance().IsReady() )
    {
       // Set some defaults since some VSTs need them...these will be reset when
       // normal or realtime processing begins
       mBlockSize = 8192;
-      DoProcessInitialize(mProjectRate);
+      vstValidator->GetInstance().DoProcessInitialize(mProjectRate);
    }
 
    // Remember the dialog with a weak pointer, but don't control its lifetime
@@ -1321,7 +1307,7 @@ std::unique_ptr<EffectUIValidator> VSTEffect::PopulateUI(ShuttleGui &S,
    // Build the appropriate dialog type
    if (mGui)
    {
-      BuildFancy();
+      BuildFancy(instance);
    }
    else
    {
@@ -1687,7 +1673,7 @@ bool VSTEffectWrapper::Load()
    // Initialize the plugin
    try
    {
-      mAEffect = pluginMain(VSTEffectWrapper::AudioMaster);
+      mAEffect = pluginMain(VSTInstanceBase::AudioMaster);
    }
    catch (...)
    {
@@ -1812,6 +1798,22 @@ void VSTEffect::Unload()
 }
 
 
+void VSTEffectInstance::Unload()
+{
+   if (mAEffect)
+   {
+      // Turn the power off
+      PowerOff();
+
+      // Finally, close the plugin
+      callDispatcher(effClose, 0, 0, NULL, 0.0);
+      mAEffect = NULL;
+   }
+
+   //ResetModuleAndHandle();
+}
+
+
 void VSTEffectWrapper::ResetModuleAndHandle()
 {
    if (mModule)
@@ -1889,7 +1891,7 @@ bool VSTEffect::LoadParameters(
       int len = Base64::Decode(value, buf.get());
       if (len)
       {
-         callSetChunkB(true, len, buf.get(), &info);
+         callSetChunk(true, len, buf.get(), &info);
          FetchSettings(GetSettings(settings));
       }
 
@@ -1997,23 +1999,23 @@ void VSTEffect::NeedEditIdle(bool state)
    mTimer->Start(100);
 }
 
-VstTimeInfo *VSTEffect::GetTimeInfo()
+VstTimeInfo* VSTInstanceBase::GetTimeInfo()
 {
    mTimeInfo.nanoSeconds = wxGetUTCTimeMillis().ToDouble();
    return &mTimeInfo;
 }
 
-float VSTEffect::GetSampleRate()
+float VSTInstanceBase::GetSampleRate()
 {
    return mTimeInfo.sampleRate;
 }
 
-int VSTEffect::GetProcessLevel()
+int VSTInstanceBase::GetProcessLevel()
 {
    return mProcessLevel;
 }
 
-void VSTEffect::PowerOn()
+void VSTInstanceBase::PowerOn()
 {
    if (!mHasPower)
    {
@@ -2031,7 +2033,7 @@ void VSTEffect::PowerOn()
    }
 }
 
-void VSTEffect::PowerOff()
+void VSTInstanceBase::PowerOff()
 {
    if (mHasPower)
    {
@@ -2085,12 +2087,14 @@ void VSTEffect::Automate(int index, float value)
    }
 
    for (const auto &slave : mSlaves)
-      slave->callSetParameter(index, value);
+      slave->callSetParameterB(index, value);
+
+   FetchSettings(mSettings);
 
    return;
 }
 
-void VSTEffect::SetBufferDelay(int samples)
+void VSTInstanceBase::SetBufferDelay(int samples)
 {
    // We do not support negative delay
    if (samples >= 0 && mUseLatency)
@@ -2148,7 +2152,7 @@ intptr_t VSTEffectWrapper::constCallDispatcher(int opcode,
       ->callDispatcher(opcode, index, value, ptr, opt);
 }
 
-void VSTEffect::callProcessReplacing(const float *const *inputs,
+void VSTInstanceBase::callProcessReplacing(const float *const *inputs,
    float *const *outputs, int sampleframes)
 {
    mAEffect->processReplacing(mAEffect,
@@ -2161,7 +2165,7 @@ float VSTEffectWrapper::callGetParameter(int index) const
    return mAEffect->getParameter(mAEffect, index);
 }
 
-void VSTEffect::callSetParameter(int index, float value)
+void VSTInstanceBase::callSetParameter(int index, float value)
 {
    bool setOK = callSetParameterB(index, value);
 
@@ -2191,7 +2195,7 @@ void VSTEffect::callSetProgram(int index)
 
    callDispatcher(effSetProgram, 0, index, NULL, 0.0);
    for (const auto &slave : mSlaves)
-      slave->callSetProgram(index);
+      slave->callSetProgramB(index);
 
    callDispatcher(effEndSetProgram, 0, 0, NULL, 0.0);
 }
@@ -2207,7 +2211,7 @@ void VSTEffectWrapper::callSetProgramB(int index)
 }
 
 
-void VSTEffectWrapper::callSetChunkB(bool isPgm, int len, void *buf)
+void VSTEffectWrapper::callSetChunk(bool isPgm, int len, void *buf)
 {
    VstPatchChunkInfo info;
 
@@ -2217,10 +2221,10 @@ void VSTEffectWrapper::callSetChunkB(bool isPgm, int len, void *buf)
    info.pluginVersion = mAEffect->version;
    info.numElements = isPgm ? mAEffect->numParams : mAEffect->numPrograms;
 
-   callSetChunkB(isPgm, len, buf, &info);
+   callSetChunk(isPgm, len, buf, &info);
 }
 
-void VSTEffectWrapper::callSetChunkB(bool isPgm, int len, void *buf, VstPatchChunkInfo *info) const
+void VSTEffectWrapper::callSetChunk(bool isPgm, int len, void *buf, VstPatchChunkInfo *info) const
 {
    if (isPgm)
    {
@@ -2244,18 +2248,6 @@ void VSTEffectWrapper::callSetChunkB(bool isPgm, int len, void *buf, VstPatchChu
    constCallDispatcher(effEndSetProgram, 0, 0, NULL, 0.0);
 }
 
-void VSTEffect::callSetChunk(bool isPgm, int len, void* buf)
-{
-   callSetChunkB(isPgm, len, buf);
-}
-
-void VSTEffect::callSetChunk(bool isPgm, int len, void* buf, VstPatchChunkInfo* info)
-{
-   callSetChunkB(isPgm, len, buf, info);
-
-   for (const auto& slave : mSlaves)
-      slave->callSetChunkB(isPgm, len, buf, info);
-}
 
 void VSTEffect::RemoveHandler()
 {
@@ -2279,10 +2271,12 @@ static void OnSize(wxSizeEvent & evt)
    }
 }
 
-void VSTEffect::BuildFancy()
+void VSTEffect::BuildFancy(EffectInstance& instance)
 {
+   auto& vstEffInstance = dynamic_cast<VSTEffectInstance&>(instance);
+
    // Turn the power on...some effects need this when the editor is open
-   PowerOn();
+   vstEffInstance.PowerOn();
 
    auto control = Destroy_ptr<VSTControl>{ safenew VSTControl };
    if (!control)
@@ -2681,7 +2675,7 @@ bool VSTEffectWrapper::LoadFXB(const wxFileName & fn)
          }
 
          // Set the entire bank in one shot
-         callSetChunkB(false, size, &iptr[40], &info);
+         callSetChunk(false, size, &iptr[40], &info);
 
          // Success
          ret = true;
@@ -2887,7 +2881,7 @@ bool VSTEffectWrapper::LoadFXProgram(unsigned char **bptr, ssize_t & len, int in
       // Set the entire program in one shot
       if (!dryrun)
       {
-         callSetChunkB(true, size, &iptr[15], &info);
+         callSetChunk(true, size, &iptr[15], &info);
       }
 
       // Update in case we're loading an "FxBk" format bank file
@@ -3403,7 +3397,7 @@ void VSTEffectWrapper::HandleXMLEndTag(const std::string_view& tag)
          int len = Base64::Decode(mChunk, buf.get());
          if (len)
          {
-            callSetChunkB(true, len, buf.get(), &mXMLInfo);
+            callSetChunk(true, len, buf.get(), &mXMLInfo);
          }
 
          mChunk.clear();
@@ -3534,7 +3528,7 @@ bool VSTEffectWrapper::StoreSettings(const VSTEffectSettings& vstSettings) const
       {
          VstPatchChunkInfo info = { 1, mAEffect->uniqueID, mAEffect->version, mAEffect->numParams, "" };
 
-         callSetChunkB(true, len, buf.get(), &info);
+         callSetChunk(true, len, buf.get(), &info);
          return true;
       }         
    }
@@ -3584,7 +3578,7 @@ ComponentInterfaceSymbol VSTEffectWrapper::GetSymbol() const
 
 EffectSettings VSTEffect::MakeSettings() const
 {
-   auto result = StatefulPerTrackEffect::MakeSettings();
+   auto result = PerTrackEffect::MakeSettings();
    // Cause initial population of the map stored in the stateful effect
    if (!mInitialFetchDone) {
       FetchSettings(GetSettings(result));
@@ -3607,47 +3601,74 @@ VSTEffectValidator::VSTEffectValidator
      mInstance(instance)
 {}
 
+
+VSTEffectInstance& VSTEffectValidator::GetInstance()
+{
+   return mInstance;
+}
+
+
 // Default, do-nothing implementations of virtuals in VSTEffectWrapper
-void VSTEffectWrapper::NeedIdle()
+void VSTInstanceBase::NeedIdle()
 {   
 }
 
-void VSTEffectWrapper::UpdateDisplay()
+void VSTInstanceBase::UpdateDisplay()
 {
 }
 
-VstTimeInfo* VSTEffectWrapper::GetTimeInfo()
-{
-   return nullptr;
-}
-
-void VSTEffectWrapper::SetBufferDelay(int samples)
+void VSTInstanceBase::SizeWindow(int w, int h)
 {
 }
 
-float VSTEffectWrapper::GetSampleRate()
+void VSTInstanceBase::Automate(int index, float value)
 {
-   return 44100.0f;
+   callSetParameter(index, value);
 }
 
-int VSTEffectWrapper::GetProcessLevel()
-{
-   //return mProcessLevel;
-   return 1;
-}
 
-void VSTEffectWrapper::SizeWindow(int w, int h)
-{
-}
+VSTEffectInstance::VSTEffectInstance
+(
+   TempClass&        effect,
+   const PluginPath& path,
+   size_t            blockSize,
+   size_t            userBlockSize,
+   bool              useLatency
+)
 
-void VSTEffectWrapper::Automate(int index, float value)
+   : StatefulEffectBase::Instance(effect)
+   , PerTrackEffect::Instance(effect)
+   , VSTInstanceBase(path)
 {
+   // what also happens in the effect ctor
+   //
+   memset(&mTimeInfo, 0, sizeof(mTimeInfo));
+   mTimeInfo.samplePos = 0.0;
+   mTimeInfo.sampleRate = 44100.0;  // this is a bogus value, but it's only for the display
+   mTimeInfo.nanoSeconds = wxGetUTCTimeMillis().ToDouble();
+   mTimeInfo.tempo = 120.0;
+   mTimeInfo.timeSigNumerator = 4;
+   mTimeInfo.timeSigDenominator = 4;
+   mTimeInfo.flags = kVstTempoValid | kVstNanosValid;
+
+   mBlockSize = blockSize;
+   mUserBlockSize = userBlockSize;
+   mUseLatency = useLatency;
+
+   Load();
 }
 
 
 VSTEffectInstance::~VSTEffectInstance()
 {
-   
+   Unload();
+}
+
+
+bool VSTEffect::Process(EffectInstance& instance, EffectSettings& settings)
+{
+   // Call through to a non-virtual function
+   return PerTrackEffect::Process(instance, settings);
 }
 
 
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1329,7 +1329,8 @@ std::unique_ptr<EffectUIValidator> VSTEffect::PopulateUI(ShuttleGui &S,
 
    auto pParent = S.GetParent();
    pParent->PushEventHandler(this);
-   return std::make_unique<DefaultEffectUIValidator>(*this, access, pParent);
+
+   return std::make_unique<VSTEffectValidator>(*this, access, pParent);
 }
 
 bool VSTEffect::IsGraphicalUI()
@@ -3591,6 +3592,7 @@ EffectSettings VSTEffect::MakeSettings() const
    return result;
 }
 
+VSTEffectValidator::~VSTEffectValidator() = default;
 
 // Default, do-nothing implementations of virtuals in VSTEffectWrapper
 void VSTEffectWrapper::NeedIdle()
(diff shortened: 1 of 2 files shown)


src/effects/VST/VSTEffect.h
```diff
@@ -101,6 +101,10 @@ struct VSTEffectSettings
 
 struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
 {
+   VSTEffectWrapper(const PluginPath& path)
```
Paul-Licameli, line 104: Most one-argument constructors should be `explicit`

src/effects/VST/VSTEffect.h
```diff
@@ -218,6 +222,34 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+   int      mVersion;
+   bool     mInteractive{ false };
+   unsigned mAudioIns{ 0 };
+   unsigned mAudioOuts{ 0 };
+   int      mMidiIns{ 0 };
+   int      mMidiOuts{ 0 };
+   bool     mAutomatable;
+
+   virtual void callSetProgram(int index) = 0;
+   virtual void Unload() = 0;
```
Paul-Licameli, line 259: I see you add two virtual functions to the wrapper when you moved Load() into it.

I understand virtual `Unload` lets wrapper avoid mention of `mDialog`.

But I think the last part of `Unload`:

```
   if (mModule)
   {
#if defined(__WXMAC__)
      mResource.reset();
      mBundleRef.reset();
#endif

      mModule.reset();
      mAEffect = NULL;
   }
```

should be the responsibility of Wrapper, balancing what it does in load.  Make those lines a non-virtual function; call it in the destructor of LV2Wrapper, and also on the failure path of `Load()`.

And I think you defined virtual `callSetProgram` to avoid dependending on `mSlaves`.  But `Load()` only needs `callSetProgram` to initialize a plug-in or an instance which will not yet have any slaves.  So I think you don't need this function, and can instead let `Load` call `callSetProgramB`

src/effects/VST/VSTEffect.h
```diff
@@ -563,4 +563,12 @@ class VSTEffectsModule final : public PluginProvider
       LoadPlugin(const PluginPath & path) override;
 };
 
+
+class VSTEffectValidator final : public DefaultEffectUIValidator
+{
+public:
+    VSTEffectValidator(EffectUIClientInterface& effect, EffectSettingsAccess& access);
```
Paul-Licameli, line 586: Might simplify it with an inherited constructor:
`using DefaultEffectUIValidator::DefaultEffectUIValidator `

src/effects/VST/VSTEffect.cpp
```diff
@@ -3609,6 +3609,13 @@ VSTEffectValidator::VSTEffectValidator(EffectUIClientInterface& effect,
 VSTEffectValidator::~VSTEffectValidator() = default;
 
 
+VSTEffectInstance::VSTEffectInstance(StatefulPerTrackEffect& effect)
```
Paul-Licameli, line 3610: The commit that stubs `ProcessInitialize` and others is incomplete.  My fault for giving an incomplete description for step 4.

`ProcessFinalize` should also be moved to the instance, and you will know all is correct when the effect also inherits `PerTrackEffect` not `StatefulPerTrackEffect`.

Please figure out step 4 as a whole and squash all the 4.x commits.

src/effects/VST/VSTEffect.cpp
```diff
@@ -987,9 +987,7 @@ size_t VSTEffect::GetBlockSize() const
 
 sampleCount VSTEffect::GetLatency() const
 {
-   if (mUseLatency)
-      return mBufferDelay;
-   return 0;
+   return GetLatencyB();
```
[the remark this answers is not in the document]
Paul-Licameli: `VSTEffectInstance` will also need to define all the other pure virtuals that were added to `VSTEffectWrapper`.

Some of the new functions should be empty stubs, others should have the lines from the corresponding VSTEffect functions moved into them, and data members should be moved too as needed to compile it.

Or a third possibility is that some members should be moved into VSTWrapper and the functions become non-virtual.

Which of the new functions of `VSTEffectInstance` should be empty?  I think NeedIdle, UpdateDisplay, SizeWindow all serve UI purposes only so should be empty.

What about the rest of them?  I'm not decided yet.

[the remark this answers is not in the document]
Paul-Licameli: The virtual functions with non-void return are `GetTimeInfo` and `GetSampleRate`.  If we have a vst handle for the plug-in distinct from the instances, will it need to report meaningful values out of `AudioMaster` or will those cases never be invoked?

I'm not sure of the answer.  But I suggest this, and let experiment see if there is a problem:

Make default implementations of these non-pure virtuals to return nullptr at 44100.

Make overrides in the Instance class.

Move the member mTimeInfo to the instance.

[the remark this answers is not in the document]
Paul-Licameli: It's unclear what `GetProcessLevel` is supposed to report to the plug-in, but the legacy code in fact never returned anything but 1 because mProcessLevel is never changed.

I say move mProcessLevel into the Wrapper class and provide a default implementation to the non-pure virtual.  Maybe we will figure out the need for an override later.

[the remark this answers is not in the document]
Paul-Licameli: I think the `void`-returning `SetBufferDelay` and `Automate` are meaningful only for instances.  Make default implementations that do nothing.  Override them in the instance.

src/effects/VST/VSTEffect.h
```diff
@@ -199,6 +210,65 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+
+
+   bool mUseLatency{ true };
+   int  mBufferDelay{ 0 };
+
+   sampleCount GetLatencyB() const
+   {
+      if (mUseLatency)
+         return mBufferDelay;
+      return 0;
```
Paul-Licameli, line 269: Last comment for this round of review!

This definition of `GetLatencyB` seems weird to me.  I think you do it because the `SetBufferDelay` virtual function must be defined for all Wrappers.  But I say again, make it non-pure and let it do nothing in the base class.  Let the Instance override it properly.  Then, that leaves no need to move the two data members into the wrapper.  They belong in the Instance only.

src/effects/VST/VSTEffect.h
```diff
@@ -101,6 +101,17 @@ struct VSTEffectSettings
 
 struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
 {
+   VSTEffectWrapper(const PluginPath& path)
+      : mPath(path)
+   {}
+
+   // This constructor and SetPath are meant to be used when creating a VSTEffectInstance
+   VSTEffectWrapper() {}
+   void SetPath(const PluginPath& path)
```
Paul-Licameli, line 110: Ah, one more comment.  Honestly the last.

I don't think this function is necessary.

src/effects/VST/VSTEffect.h
pietro68: ok, done

src/effects/VST/VSTEffect.h
pietro68: ok, done

src/effects/VST/VSTEffect.h
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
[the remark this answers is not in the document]
pietro68: Automate also references mMaster, so to have it a member of the instance, I had to move it to the wrapper.

I think I implemented all your suggestions here? please check commit "4.3 Implement some virtuals in the instance"

src/effects/VST/VSTEffect.h
pietro68: That's right, I could add the PluginPath as an arg to the Instance constructor. Done in commit 4.4

src/effects/VST/VSTEffect.cpp
```diff
@@ -1822,6 +1821,28 @@ void VSTEffect::Unload()
       mAEffect = NULL;
    }
 
+   ResetModuleAndHandle();
```
Paul-Licameli, line 1824: This call within `Unload` is unnecessary.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1822,6 +1821,28 @@ void VSTEffect::Unload()
...
+   {
+      // Turn the power off
+      PowerOff();
+
+      // Finally, close the plugin
+      callDispatcher(effClose, 0, 0, NULL, 0.0);
+      mAEffect = NULL;
+   }
+
+   ResetModuleAndHandle();
```
Paul-Licameli, line 1840: This too is unnecessary.

src/effects/VST/VSTEffect.h
pietro68: I commited part of your suggestion in 4.5. `mUseLatency` can not be moved to the instance, because it is referenced by `VSTEffect::ShowOptions` - any idea on how to solve that?

src/effects/VST/VSTEffect.cpp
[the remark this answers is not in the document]
Paul-Licameli: You did not remove `GetLatencyB` or undo the move of the two data members `mUseLatency`, `mBufferDelay`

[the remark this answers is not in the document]
Paul-Licameli: If you examine how mMaster is used, it's really just a boolean that is true when an instance (a VSTEffect) is NOT the master.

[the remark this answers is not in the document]
Paul-Licameli: You need mMaster in Automate, but the wrapper has a no-op override of Audomate, so there should be no need to put mMaster in the Wrapper.  You also moved mTimeInfo into the Wrapper but it should be in the Instane.

src/effects/VST/VSTEffect.h
Paul-Licameli: Just duplicate that boolean in the instance.  Make it a constructor argument.

src/effects/VST/VSTEffect.cpp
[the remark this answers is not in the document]
pietro68: > You need mMaster in Automate, but the wrapper has a no-op override of Automate, so there should be no need to put mMaster in the Wrapper.

but mMaster is set and used in 3 different VSTEffect methods - that's why I put it in the wrapper. I probably could rework the code in a way that VSTEffect does not need it anymore, but I'd leave that for later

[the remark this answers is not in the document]
pietro68: > You also moved mTimeInfo into the Wrapper but it should be in the Instance.

`mTimeInfo` is again written/used in different methods of `VSTEffect`. Should I figure out how to move these uses of it from `VSTEffect` to `VSTEffectInstance` as part of step 4?

src/effects/VST/VSTEffect.h
pietro68: In commit 4.6 I moved `mUseLatency` and `mBufferDelay` in the instance, leaving a copy of `mUseLatency` in the Effect class which is passed to the instance at construction time.

src/effects/VST/VSTEffect.cpp
pietro68: ok, removed

src/effects/VST/VSTEffect.cpp
pietro68: ok, removed

pietro68:
Please note: "effect inherits PerTrackEffect not StatefulPerTrackEffect" is specified originally as part of step 4; but once I changed that inheritance, the code would not build until I implemented the most parts of step 5. - this is why it is part of step 5.

src/effects/VST/VSTEffect.h
```diff
@@ -288,6 +288,38 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
    int mProcessLevel{ 1 };  // in GUI thread
 
    VSTEffect* mMaster;     // non-NULL if a slave
+
+#if defined(__WXMAC__)
```
Paul-Licameli, line 292: Please do an interactive rebase, squashing this commit onto the second one, so that all commits are buildable.

pietro68: ok, now done and force-pushed

pietro68: oops I should delete the "build fix for OSx" commit - it's only whitespaces now

Paul-Licameli:
"removed unnecessary calls to ResetModuleAndHandle" might be squashed earlier

Paul-Licameli:
> Please note: "effect inherits PerTrackEffect not StatefulPerTrackEffect" is specified originally as part of step 4; but once I changed that inheritance, the code would not build until I implemented the most parts of step 5. - this is why it is part of step 5.

The problem, as I explained before, was that I omitted before part of what needed to be done in step 4.  That is, move `ProcessFinalize` also into the Instance.

Your step 5.1 simply deletes that function from the Effect but does not restore the equivalent into the Instance.  That may be losing an important side effect.

So I can't accept this pull request yet.

Paul-Licameli (changes requested):
This still isn't right and the fault is partly mine for not stating the steps in the issue clearly enough.  Please carefully re-read the updated steps 4 and 5.

For step 4: ProcessFinalize must also move into the instance.  The Effect stops inheriting StatefulPerTrackEffect but inherits TWO bases instead:  PerTrackEffect and StatefulEffectBase (which are also the direct bases of StatefulPerTrackEffect).  Just squash all the 4.x steps into one;  do not define ProcessBlock in the Wrapper because you don't need to shared it with the Effect.  Do not move callProcessReplacing into the Wrapper.

For step 5:  It is here that the Effect loses its other base class StatefulEffectBase, and other virtual functions of the effect for realtime move into the instance.

pietro68:
I redid step 4. from scratch - hopefully it is not too complicated, I could not split it into subparts.

src/effects/VST/VSTEffect.h
```diff
@@ -101,6 +101,15 @@ struct VSTEffectSettings
 
 struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
 {
+   explicit VSTEffectWrapper(const PluginPath& path)
+      : mPath(path)
+   {}
+
+   ~VSTEffectWrapper()
```
Paul-Licameli, line 108: I just noticed in review today, the omission of `virtual` from this destructor, though there are other virtual functions.

It might look nicer to add it.  It's not in fact important, so long as there are never yet any unique or shared pointers to a VSTEffectWrapper.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1341,7 +1355,8 @@ std::unique_ptr<EffectUIValidator> VSTEffect::PopulateUI(ShuttleGui &S,
 
    auto pParent = S.GetParent();
    pParent->PushEventHandler(this);
-   return std::make_unique<DefaultEffectUIValidator>(*this, access, pParent);
+
+   return std::make_unique<VSTEffectValidator>(*this, access, pParent);
```
Paul-Licameli, line 1359: The minimal validator class doesn't yet appear to be necessary for the VST2 transformation, unlike the case for AudioUnits and VST3.  But no harm in keeping the commit that adds it.

Paul-Licameli:
I wrote earlier "Also in step 4, the PowerOn() in BuildFancy() should be changed to call it on the Instance, not on this.
Pass the Instance into BuildFancy() from PopulateUI()."

The latest push doesn't do this any more.  Please restore that change.

src/effects/VST/VSTEffect.h
```diff
@@ -199,6 +208,135 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+                               intptr_t value,
+                               void * ptr,
+                               float opt);
+
+   // These are needed to move the AudioMaster callback from VSTEffect to this struct
+   //
+   intptr_t mCurrentEffectID {};
+
+   bool mWantsIdle{ false };
+   std::unique_ptr<VSTEffectTimer> mTimer;
```
Paul-Licameli, line 224: It is not necessary to move these two into the Wrapper and I think they will eventually be in the Instance.

src/effects/VST/VSTEffect.h
```diff
@@ -212,15 +350,24 @@ using VSTEffectArray = std::vector < std::unique_ptr<VSTEffect> > ;
 DECLARE_LOCAL_EVENT_TYPE(EVT_SIZEWINDOW, -1);
 DECLARE_LOCAL_EVENT_TYPE(EVT_UPDATEDISPLAY, -1);
 
+
+// Temporary class allowing an intermediate phase in the transformation.
+// It is needed because the instance constructor needs something which is
+// both a PerTrackEffect and a StatefulEffectBase
+class TempClass : public PerTrackEffect
```
Paul-Licameli, line 363: The existing `StatefulEffect` class might also serve the purpose

src/effects/VST/VSTEffect.h
```diff
@@ -212,15 +350,24 @@ using VSTEffectArray = std::vector < std::unique_ptr<VSTEffect> > ;
...
+{};
+
 ///////////////////////////////////////////////////////////////////////////////
 ///
 /// VSTEffect is an Audacity effect that forwards actual
 /// audio processing via a VSTEffectLink
 ///
 ///////////////////////////////////////////////////////////////////////////////
 class VSTEffect final
-   : public StatefulPerTrackEffect
```
Paul-Licameli, line 222: The `#include` at top of the file can now be `PerTrackEffect.h`, not needing the stateful subclass

src/effects/VST/VSTEffect.cpp
```diff
@@ -952,7 +950,8 @@ std::shared_ptr<EffectInstance> VSTEffect::DoMakeInstance()
    GetConfig(*this, PluginSettings::Shared, wxT("Options"),
       wxT("UseLatency"), mUseLatency, true);
    mBlockSize = mUserBlockSize;
-   return std::make_shared<Instance>(*this);
+ 
+   return std::make_shared<VSTEffectInstance>(*this, mPath);
```
Paul-Licameli, line 945: An earlier version of this PR copied mUseLatency into the Instance in its constructor.  That change should remain.

What I should have said too:  the block size, and the user block size, should also be passed into the constructor of the instance.

Also the array of slaves used to be passed as a reference into the constructor.  Why no longer?  Do that too.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1038,7 +1037,14 @@ bool VSTEffect::ProcessFinalize()
    return true;
 }
 
-size_t VSTEffect::ProcessBlock(EffectSettings &,
+
+bool VSTEffectInstance::ProcessFinalize()
+{
+   return VSTEffectWrapper::DoProcessFinalize();
```
Paul-Licameli, line 1036: I don't like `DoProcessFinalize`, `DoProcessInitialize`, or `ProcessBlockB`.  Too much is stil being done in the wrapper class, and therefore the Effect still has more state that it should.  The purpose of this PR should be to remove as much effect state as possible, identifying the parts that should be per-instance state.

And I think the reason you thought you needed this is:  you let the slaves remain VSTEffect objects, and did not change the type of the slaves to be VSTEffectInstance.  But you can do this.

The array of slaves remains temporarily in the Effect, until step 5 is done.  But the virtual functions needed in the slaves are only `ProcessFinalize` and `ProcessBlock`, now in the Instance class, and other functions already in the Wrapper.

Paul-Licameli (changes requested):

src/effects/VST/VSTEffect.h
```diff
@@ -199,6 +208,141 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+   intptr_t mCurrentEffectID {};
+
+   bool mWantsIdle{ false };
+   std::unique_ptr<VSTEffectTimer> mTimer;
+
+   virtual void NeedIdle() = 0;
+   virtual void UpdateDisplay() = 0;
+   virtual void SizeWindow(int w, int h) = 0;
+
+   virtual VstTimeInfo* GetTimeInfo()
```
Paul-Licameli, line 230: I would prefer to have these virtual functions defined non-inline, with comments inside the class about what they do (which is mostly nothing)

src/effects/VST/VSTEffect.h
```diff
@@ -264,19 +417,17 @@ class VSTEffect final
...
    bool IsReady();
-   bool ProcessInitialize(EffectSettings &settings, double sampleRate,
-      sampleCount totalLen, ChannelNames chanMap) override;
-   bool DoProcessInitialize(double sampleRate);
-   bool ProcessFinalize() override;
-   size_t ProcessBlock(EffectSettings &settings,
-      const float *const *inBlock, float *const *outBlock, size_t blockLen)
-      override;
+
+   bool Process(EffectInstance& instance, EffectSettings& settings)
```
Paul-Licameli, line 425: This function should be marked `override` and its definition should be non-inline

src/effects/VST/VSTEffect.cpp
```diff
@@ -952,7 +941,8 @@ std::shared_ptr<EffectInstance> VSTEffect::DoMakeInstance()
    GetConfig(*this, PluginSettings::Shared, wxT("Options"),
```
Paul-Licameli, line 941: If all is done right, `DoMakeInstance` and the `const_cast` should no longer be needed.
`mUseLatency` and `mUserBlockSize` don't need to be member variables of the effect.  They can become local variables of this function.  But then they must be passed to the constructo of Instance which stores them.

But this may need to wait until step 5.

src/effects/VST/VSTEffect.h
pietro68: Correct. Now moved back in the effect.

src/effects/VST/VSTEffect.h
pietro68: Hmm... StatefulEffect does not inherit PerTrackEffect. I gave it a go anyway, and I got compiler errors. Never mind, this is only a temporary class that should disappear at point 5.

src/effects/VST/VSTEffect.h
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
pietro68: mUseLatency, block size, user block size are now passed into the constructor of the instance and saved there.

I can not remember ever passing the array of slaves to the instance? and I wonder why I should do it now. I converted slaves to be instances. Why would a slave/instance need to know the other slaves/instances?

src/effects/VST/VSTEffect.cpp
pietro68: I now implemented the "middle class" VstInstanceBase as we discussed, please have a look there.

src/effects/VST/VSTEffect.h
pietro68: I made them inlinenow.

src/effects/VST/VSTEffect.h
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
pietro68: I now pass them to the Instance constructor. As you said, later in step 5. I could remove them from the effect.

pietro68:
> I wrote earlier "Also in step 4, the PowerOn() in BuildFancy() should be changed to call it on the Instance, not on this. Pass the Instance into BuildFancy() from PopulateUI()."
> 
> The latest push doesn't do this any more. Please restore that change.

ok, restored

src/effects/VST/VSTEffect.cpp
pietro68: Well actually at this point I need it, because I pass the instance it belongs to in its constructor, to retrieve it later in ::ShowClientInterface

src/effects/VST/VSTEffect.cpp
Paul-Licameli: I do not yet see a force-push with that update

pietro68:
All of what we discussed yesterday is now in the last push I did, commit 4.

Paul-Licameli:
The latest push looks good, but it awaits completion with step 5.  A few comments:

- `VSTEffectValidator::mIntance` might be private.
- I was surprised to see `AudioMaster` moved into `VSTInstanceBase`.  I thought at first, that's wrong, it should remain in the Wrapper, and the Validator may need to use it too.  But your comments suggest, we will split the `AudioMaster` function into multiple functions, and pass different functions to `pluginMain`.  I now think that's a good alternative solution too.
- `ProcessBlockB ` doesn't need to be separated from `ProcessBlock ` as a member function of `VSTInstanceBase`, now that slaves are also Instances.  (But not so, yet, for the initialize and finalize functions.)
- Where is `VSTEffectInstance::mNumChannels` used?  I can't find it.  You might just delete that, and the functions that get and set it.
- `callSetChunk` (which has two overloads) is no longer called anywhere!  (And the four-argument overload in fact became unused already at commit 2df527f6d72c92c5cf5c8cc679ac6122a03c4649). So, just delete those, and rename the `callSetChunkB` functions as `callSetChunk`.
- I think `VSTInstanceBase::Automate` is wrong... I think the `VSTInstance` needs to override it, and call `callSetParameter` (not B), and so, the Instance needs, transitionally, a reference to the array of slaves.
- Why is `mTimeInfo` still in the Effect class?  Only the Instance and the slaves, also Instances, need it.  Once you address the point about `ProcessBlockB`, this will become clearer.  Remove it from the Effect.

Please do an interactive rebase addressing all of these, force push it, and then proceed to step 5.

pietro68:
> * Why is `mTimeInfo` still in the Effect class?  Only the Instance and the slaves, also Instances, need it.  Once you address the point about `ProcessBlockB`, this will become clearer.  Remove it from the Effect.

`mTimeInfo` is actually in the InstanceBase class, not the Effect class; it is referenced by:
- `GetTimeInfo()`  called from the vst callback
- `GetSampleRate()` called from the vst callback
- `VSTEffect::VSTEffect` 

As long as the vst callback is called for the effect too, I can not move `mTimeInfo` in the Instance only. I actually wonder if I can move it there even after I split the vst callback in two parts (one for the effect, one for the instance) because maybe the callback for the effect will still need to call `GetTimeInfo` and/or `GetSampleRate`? this because the vst GUI might want to draw something that depends on the samplerate (think of a band equalizer for instance)

Paul-Licameli:
> > * Why is `mTimeInfo` still in the Effect class?  Only the Instance and the slaves, also Instances, need it.  Once you address the point about `ProcessBlockB`, this will become clearer.  Remove it from the Effect.
> 
> `mTimeInfo` is actually in the InstanceBase class, not the Effect class; it is referenced by:
> 
> * `GetTimeInfo()`  called from the vst callback
> * `GetSampleRate()` called from the vst callback
> * `VSTEffect::VSTEffect`
> 
> As long as the vst callback is called for the effect too, I can not move `mTimeInfo` in the Instance only. I actually wonder if I can move it there even after I split the vst callback in two parts (one for the effect, one for the instance) because maybe the callback for the effect will still need to call `GetTimeInfo` and/or `GetSampleRate`? this because the vst GUI might want to draw something that depends on the samplerate (think of a band equalizer for instance)

I thought we decided it is not needed in the Effect. If for GetTimeInfo and GetSampleRate, the effect can simply inherit the trivial implementations, then mTimeInfo can belong to Instance only.

Paul-Licameli:
> > > * Why is `mTimeInfo` still in the Effect class?  Only the Instance and the slaves, also Instances, need it.  Once you address the point about `ProcessBlockB`, this will become clearer.  Remove it from the Effect.
> > 
> > 
> > `mTimeInfo` is actually in the InstanceBase class, not the Effect class; it is referenced by:
> > 
> > * `GetTimeInfo()`  called from the vst callback
> > * `GetSampleRate()` called from the vst callback
> > * `VSTEffect::VSTEffect`
> > 
> > As long as the vst callback is called for the effect too, I can not move `mTimeInfo` in the Instance only. I actually wonder if I can move it there even after I split the vst callback in two parts (one for the effect, one for the instance) because maybe the callback for the effect will still need to call `GetTimeInfo` and/or `GetSampleRate`? this because the vst GUI might want to draw something that depends on the samplerate (think of a band equalizer for instance)
> 
> I thought we decided it is not needed in the Effect. If for GetTimeInfo and GetSampleRate, the effect can simply inherit the trivial implementations, then mTimeInfo can belong to Instance only.

If experiment shows that those cases in AudioMaster are not reached for the Effect object, but only for the Instance, then
the simplification of moving mTimeInfo out of Effect could be done and would be desirable.

But we might leave that until we do such experiment.

pietro68:
> * I think `VSTInstanceBase::Automate` is wrong... I think the `VSTInstance` needs to override it, and call `callSetParameter` (not B), and so, the Instance needs, transitionally, a reference to the array of slaves.

I'd like to think out loud a bit now. Automate is called from the vst callback, sometimes on the Effect, sometimes on the Instance. I put some breakpoints and I observed this:

1) when instantiating an effect 
--> Automate is called on the effect, nothing happens because mMaster is null and mSlaves is empty

2) when moving a slider on the effect gui --> as above

3) when pressing play 
--> Automate is called on the instance

4) when clicking "apply" (i.e. destructive processing) 
--> Automate is called first on the effect (mMaster=null mSlaves={}) and then on the instance

5) when moving a slider while playing 
--> Automate is called on the effect, mMaster=null but this time mSlaves has one item

My conclusions (but I might be wrong): 

a) we need both overrides of Automate (for Effect and Instance), and when I will split the callback, both of them will need to call Automate when requested to. 

b) mMaster can be removed (it is always nullptr in my tries above)

c) I could do as you say, i.e. pass the array of slaves to the instance and have Instance::Automate call callSetParameter (not B) on them, but, is it really needed? without this, everything still works and it is a temporary thing anyway

pietro68:
> If experiment shows that those cases in AudioMaster are not reached for the Effect object, but only for the Instance, then the simplification of moving mTimeInfo out of Effect could be done and would be desirable.
> 
> But we might leave that until we do such experiment.

Did a little experiment: using a simple plugin like gain, `GetTimeInfo()` is never reached, neither for the effect nor for the instance. But already using an equalizer plugin, at least `GetTimeInfo()` is reached, and for the effect only. 
So I would say, at least `GetTimeInfo()` should remain for the effect, and if we want to play it safe, also` GetSampleRate()`.

pietro68:
A note about my commits for step 5. :

I said earlier (and wrote in code comments) that I would split the vst callback function in two parts, one for the instance and one for the effect.

Since then I changed my mind and opted for a more conservative solution, which is: keep the vst callback in the wrapper and have it call virtual methods. I think doing so is cleaner and poses less risks of breaking things.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3599,4 +3597,42 @@ EffectSettings VSTEffect::MakeSettings() const
...
+{
+}
+
+float VSTEffectWrapper::GetSampleRate()
+{
+   return 44100.0f;
+}
+
+int VSTEffectWrapper::GetProcessLevel()
+{
```
Paul-Licameli, line 3624: This is the one virtual function of `VSTEffectWrapper` that we can define in the base and then not override again, or so we think now.

So the first commit might also remove `VSTEffect::GetProcessLevel()`

src/effects/VST/VSTEffect.cpp
```diff
@@ -3607,6 +3608,18 @@ EffectSettings VSTEffect::MakeSettings() const
 
 VSTEffectValidator::~VSTEffectValidator() = default;
 
+
+VSTEffectValidator::VSTEffectValidator
+(
+   VSTEffectInstance&       instance,
+   EffectUIClientInterface& effect,
+   EffectSettingsAccess&    access,
+   wxWindow*                pParent
```
Paul-Licameli, line 3589: The commit that defines the Instance adds a reference to Instance in the Validator.  It also adds three members that are not needed.  They only duplicate protected members of the base class.

src/effects/VST/VSTEffect.h
```diff
@@ -287,29 +270,103 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+   bool mReady{ false };
+
+   int mBufferDelay{ 0 };
+
+   // The vst callback is currently called both for the effect and for any instance.
+   // After we'll have instances handle realtime processing too (step 5.) then we'll
+   // be able to move the callback and the method it calls into the instance class.
+   //
+   // Actually: the callback will have to be split in two - one part as the callback
+   // for the instances, and the remaining part as the callback for the effect
```
Paul-Licameli: Is this part of the comment still correct?

src/effects/VST/VSTEffect.cpp
```diff
@@ -1087,11 +1086,11 @@ bool VSTEffect::RealtimeAddProcessor(
    if (clen == 0) {
       callDispatcher(effBeginSetProgram, 0, 0, NULL, 0.0);
       for (int i = 0; i < mAEffect->numParams; i++)
-         slave->callSetParameter(i, callGetParameter(i));
+         slave->callSetParameterB(i, callGetParameter(i));
```
Paul-Licameli, line 1072: I did not mention this before, but if you look for the analogous uses of `StoreSettings` on AudioUnit instances, one of them happens here.  In step 4 you might make a similar change here too, eliminating the loop over parameters, and the beginnin and end of set-program, which would happen inside `StoreSettings`.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3620,47 +3627,74 @@ VSTEffectValidator::VSTEffectValidator
...
 {
-   return 44100.0f;
 }
 
-int VSTEffectWrapper::GetProcessLevel()
+void VSTInstanceBase::Automate(int index, float value)
 {
-   //return mProcessLevel;
-   return 1;
+   callSetParameterB(index, value);
```
Paul-Licameli, line 3630: Automate() is changed to use `callSetParameterB`, losing the loop over slaves.  That seems like a wrong change in side-effects.

You should call `callSetParameter`.   But that doesn't exist in `VSTInstanceBase`.

The fix should then be:  move Automate() to the Instance class in step 4, which is what the header file says should happen.

Also move `callSetParameter` (which includes the loop) into Instance, out of the Effect object.

But then `VSTEffect::OnSlider` won't compile.  But the solution for that will be, the Validator object will call `callSetParameter` on the Instance.  That function will ultimately be in the validator.

So make this temporary fix here in step 4:  Let the Effect retain a pointer to the last constructed Validator.  Assume there is never more than one Valdator at once and that its lifetime will include any call to `OnSlider`, which will happen only while a non-fancy, plain UI dialog is open.

src/effects/VST/VSTEffect.cpp
```diff
@@ -2204,7 +2221,7 @@ void VSTEffect::callSetProgram(int index)
 
    callDispatcher(effSetProgram, 0, index, NULL, 0.0);
    for (const auto &slave : mSlaves)
-      slave->callSetProgram(index);
+      slave->callSetProgramB(index);
```
Paul-Licameli: Can the distinction between `callSetProgram` and `callSetProgramB` now be eliminated?  I think so, but it needs some longer explanation from me.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1037,7 +1039,14 @@ return GuardedCall<bool>([&]{
 });
 }
 
-size_t VSTEffect::ProcessBlock(EffectSettings &,
+
+bool VSTEffectInstance::ProcessFinalize() noexcept
+{
+   return VSTInstanceBase::DoProcessFinalize();
```
Paul-Licameli: I think there is no longer any need to make `DoProcessFinalize` a separate function.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1128,21 +1128,31 @@ bool VSTEffect::RealtimeResume()
...
    wxASSERT(numSamples <= mBlockSize);
    if (group >= mSlaves.size())
       return 0;
-   return mSlaves[group]->ProcessBlock(settings, inbuf, outbuf, numSamples);
+
+   {
+      // If we assume that the user might be moving knobs during realtime processing and wants
+      // to hear how the sound changes, we must do this then
+
+      auto guard = std::lock_guard{ GetEffect().mSettingsMutex };
```
Paul-Licameli, line 1130: Do not do this.

We don't want to do mutex locking and unlocking in a worker thread.

We will trust `settings` to be a copy of settings, properly updated in an atomic fashion, by the `EffectSettingsAccess` object.

Yes, strictly speaking you are right that a data race on the `VSTEffectSettings` stored in the effect, is possible while the transition to statelessness is not complete, and you might remove this mutex locking then.

But I think that is excessive caution.  I didn't transform AudioUnit effects with similar caution and I don't recommend you do it for VST.

src/effects/VST/VSTEffect.cpp
```diff
@@ -2191,18 +2188,14 @@ float VSTEffectWrapper::callGetParameter(int index) const
...
+
 void VSTEffect::callSetParameter(int index, float value)
 {
    bool setOK = callSetParameterB(index, value);
-
-   if (setOK)
-   {
-      for (const auto& slave : mSlaves)
-         slave->callSetParameterB(index, value);
-   }
```
Paul-Licameli, line 2177: Your step 5.1 removes these lines.  I already explained how it should be done in step 4, eliminating the need for two versions of the set-parameter function.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3650,20 +3641,31 @@ void VSTInstanceBase::SizeWindow(int w, int h)
...
+   // Because we come here when a control on the effect's GUI is moved,
+   // we must update the temporary Effect-owned settings - users might
+   // want to hear what happens when they move a knob.
+   //
+   {
+      // we need a mutex because FetchSettings writes mSettings in the main
+      // thread, but the storing of the settings passed to ::RealtimeProcess
+      // happens in the worker thread
+      auto guard = std::lock_guard{ mSettingsMutex };
+      FetchSettings(mSettings);
```
Paul-Licameli, line 3641: You made these changes in 5.1.  I don't like the mutex.

Does experiment show that this is called on the main thread only in response to slider changes?

I need the answer to that, before I can recommend the correct thing  to do.

src/effects/VST/VSTEffect.h
```diff
@@ -605,6 +582,27 @@ class VSTEffectInstance final : public StatefulEffectBase::Instance,
    
    bool ProcessFinalize() noexcept override;
 
+
+   size_t SetBlockSize(size_t maxBlockSize) override;
+   size_t GetBlockSize() const override;
+
+   // this not?
+   //bool Process(EffectInstance& instance, EffectSettings& settings) override;
```
Paul-Licameli: There is nothing in the instance now corresponding to `Process`.  Instead you trust `PerTrackEffect` to do destructive effect processing in its `Process`, calling the `ProcessBlock` function of the instance.  And after other recent refactorings, that happens in member functions of an `EffectStage` object.

src/effects/VST/VSTEffect.cpp
```diff
@@ -785,9 +785,8 @@ void VSTEffect::ResourceHandle::reset()
 }
 #endif
 
-VSTEffect::VSTEffect(const PluginPath & path, VSTEffect *master)
-:  VSTInstanceBase(path),
-   mMaster(master)
+VSTEffect::VSTEffect(const PluginPath & path)
+:  VSTInstanceBase(path)
```
Paul-Licameli: I think step 5.3 makes more sense squashed with step 4

src/effects/VST/VSTEffect.cpp
```diff
@@ -1182,7 +1177,7 @@ int VSTEffect::ShowClientInterface(
       // Set some defaults since some VSTs need them...these will be reset when
       // normal or realtime processing begins
       mBlockSize = 8192;
-      DoProcessInitialize(mProjectRate);
+      vstValidator->GetInstance().DoProcessInitialize(mProjectRate);
```
Paul-Licameli, line 1181: This change might be squashed onto step 4, making 4 more correct

src/effects/VST/VSTEffect.cpp
pietro68: Removing it at this stage is not so simple, because DoProcessFinalize is in VSTInstanceBase and it is used by VSTEffect; but VSTEffect does not know VSTEffectInstance... we'd need to wait until step 5

src/effects/VST/VSTEffect.cpp
pietro68: OK done that in commit 4.2 (yet to be pushed at this moment) - you will see that to make it work I also had to add a call to `FetchSettings` inside `Automate`.

src/effects/VST/VSTEffect.cpp
pietro68: The short answer is: not, when doing realtime processing, at least at this stage (I mean, step 4)

Why? I tried the following: 
1) place breakpoint in CallSetProgram
2) choose a factory preset: when we reach CallSetProgram, mSlaves is empty

So if we did only destructive processing, then we could go with callSetProgramB only (i.e. the version which ignores slaves)

but:
if I start realtime processing and then choose a factory preset, when I reach callSetProgram, mSlaves is not empty - it has one item and I guess it is the item doing the realtime processing.

src/effects/VST/VSTEffect.cpp
pietro68: I was wrong when I said that Automate has to belong to the instance. 
It must indeed belong to the Effect, because changes in sliders are intercepted by the Effect vst callback.

That being said, yes it is still wrong that Automate() loses the loop over slaves. In commit 4.3 I fix that by having Automate use callSetParameter and that required that mInstances move to the instancebase class.

src/effects/VST/VSTEffect.cpp
pietro68: As I told you during our talk, without introducing this mutex, I always ran into a crash when I moved a slider while processing realtime. This is because mSettings was written in the main thread (when moving the slider), while it was read in the worker thread. 
But I will add a comment about the mutex being a temporary thing until the validator is fully implemented.

src/effects/VST/VSTEffect.cpp
pietro68: Please see my comment above https://github.com/audacity/audacity/pull/3278#discussion_r946051875

Paul-Licameli:
> I was wrong when I said that Automate has to belong to the instance. It must indeed belong to the Effect, because changes in sliders are intercepted by the Effect vst callback.
> 
> That being said, yes it is still wrong that Automate() loses the loop over slaves. In commit 4.3 I fix that by having Automate use callSetParameter and that required that mInstances move to the instancebase class.

I'm still not satisfied.  As I explain:  Both in Automate(), and in OnSlider() (which will both ultimately go to the Validator class), callSetParameter should be invoked, but NOT in `this`, but instead on an Instance object.

That is the Instance stored in the minimal Validator class now; and, until the Validator is complete, the Effect will keep a pointer to the active Validator.

src/effects/VST/VSTEffect.cpp
pietro68: Ok, done in commit 4.5

pietro68:
> I'm still not satisfied. As I explain: Both in Automate(), and in OnSlider() (which will both ultimately go to the Validator class), callSetParameter should be invoked, but NOT in `this`, but instead on an Instance object.
> 
> That is the Instance stored in the minimal Validator class now; and, until the Validator is complete, the Effect will keep a pointer to the active Validator.

I can not do this because:

1) `::Automate` comes in two versions:

- `VSTInstanceBase::Automate`  which has no access to members in the Effect
- `VSTEffect::Automate` it is actually calling callsetparameterB


2) I tried doing it in `::OnSlider`, but then it breaks this functionality:  as I move sliders of the plain GUI while realtime processing, I do not hear changes in the sound anymore.

src/effects/VST/VSTEffect.h
pietro68: ok

src/effects/VST/VSTEffect.cpp
```diff
@@ -3594,6 +3595,18 @@ EffectSettings VSTEffect::MakeSettings() const
 
 VSTEffectValidator::~VSTEffectValidator() = default;
 
+
+VSTEffectValidator::VSTEffectValidator
+(
+   VSTEffectInstance&       instance,
+   EffectUIClientInterface& effect,
+   EffectSettingsAccess&    access,
+   wxWindow*                pParent
```
Paul-Licameli, line 3562: I suggest it again:  at step 2, you don't need `effect`, `access`, `pParent` members because you can just use the protected members of the base classes of `VSTEffectValidator.`

Paul-Licameli:
I am reviewing the latest force-push and so far I am satisfied with the 4.x steps.  They can be squashed into one in your next force-push.

src/effects/VST/VSTEffect.h
```diff
@@ -150,6 +151,8 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
    bool IsCompatible(const VstPatchChunkInfo&) const;
 
    VSTEffectSettings mSettings;  // temporary, until the effect is really stateless
+   std::mutex mSettingsMutex;    // to avoid read/write races on mSettings - this is needed temporarily
+                                 // and will be removed when the Validato will be implemented
```
Paul-Licameli: "Validator"

src/effects/VST/VSTEffect.cpp
```diff
@@ -1154,6 +1165,7 @@ int VSTEffect::ShowClientInterface(
    //   mProcessLevel = 1;      // in GUI thread
 
    VSTEffectValidator* vstValidator = static_cast<VSTEffectValidator*>(validator);
+   mValidator = vstValidator;
```
Paul-Licameli, line 1150: Step 5.1 adds this assignment, but never uses the member, and never reassigns it back to null.  But I'm still progressing through the 5.x commits, which may be squashed into one if I approve them.

src/effects/VST/VSTEffect.h
pietro68: ok

src/effects/VST/VSTEffect.cpp
pietro68: Correct. I introduced `mValidator` as a way to call `callSetParameter` on the Validator's instance, in `::OnSlider`, as you previously suggested (or so I understood). Then I found out that things work also without that call (and I left the code there commented out) so I could delete `mValidator` now. Shall I do it?

src/effects/VST/VSTEffect.cpp
```diff
@@ -374,7 +374,7 @@ unsigned VSTEffectsModule::DiscoverPluginsAtPath(
          //Subsequent VSTEffect::Load may seem like overhead, but we need
          //to initialize EffectDefinitionInterface part, which includes
          //properly formatted plugin path
-         VSTEffect subeffect(wxString::Format("%s;%d", path, id), &effect);
+         VSTEffect subeffect(wxString::Format("%s;%d", path, id));
```
Paul-Licameli, line 377: In step 5.3, add:
`subeffect.Load()`
to preserve the behavior.

Paul-Licameli (changes requested):

src/effects/VST/VSTEffect.cpp
```diff
@@ -2013,7 +2009,7 @@ int VSTInstanceBase::GetProcessLevel()
    return mProcessLevel;
 }
 
-void VSTInstanceBase::PowerOn()
+void VSTEffectWrapper::PowerOn()
```
Paul-Licameli, line 2009: In step 5.4, think `PowerOn` and `PowerOff` should not move to Wrapper.  They are only for the Instance.

Power comes ON, only in an instance.  Note `vstEffInstance.PowerOn()` -- it's not done on This.

But there are calls to `PowerOff` done on the Effect.

The right thing is to change those.

Where `CloseUI` does `PowerOff`, do it instead to the Instance stored in the Validator.  Which is why `VSTEffect::mValidator` finds a use after all.

Als in `CloseUI`, reassign `mValidator` to null when done.

The `PowerOff()` in the unloading of the effect should just be deleted.

Paul-Licameli (changes requested):

src/effects/VST/VSTEffect.cpp
pietro68: Good catch! ok

Paul-Licameli:
Please rebase onto latest master, squash steps 4 into one, and squash steps 5 into another commit.  Address in step 5 the two reqeusts for change.  Then, this will be approved, and you will move on to the Validator.

Paul-Licameli:
Less important, `mUseLatency` might also move from Wrapper to Instance

Paul-Licameli (approved):
Steps 4.x and 5.x might be squashed into just two commits before merging.  All else looks good now.

Issue #3286, opened 2022-07-25
Vamp plug-ins: some plug-ins cannot be loaded


vsverchinsky:
Vamp plug-ins could not be loaded if the name of the plug-in contains '/' symbol.
(Like "bbc-vamp-plugins:bbc-speechmusic-segmenter/Speech/Music segmenter: Detection function" from [this](https://www.vamp-plugins.org/pack.html) pack.)

1. Install plugin pack using link above
2. Run Audacity and note that some vamp plugins are reported to be incompatible (skip this step if version of Audacity is lower than 3.2)
3. Open plugin manager and try to enable one of them

**Actual behavior**
Observe: "could not load library" message

**Expected behavior**
Plugin not reported as incompatible and loads without an issue

dozzzzer:
I've managed to reproduce this.
In addition to 'Detection function', 
`bbc-vamp-plugins:bbc-speechmusic-segmenter/Speech/Music segmenter: Segmentation`

has also failed to load.

Issue #2965, opened 2022-05-23
Uninstalled Nyquist plug-in listed as "Disabled"


SteveDaulton:
**Describe the bug**
When a Nyquist plug-in is removed (deleted) from the user's Plug-Ins folder and Audacity rebooted, the plug-in still appears in the plug-in manager as "Disabled". This is potentially very confusing: If the plug-in was uninstalled a long time previously, there is no indication of why enabling the plug-in fails. Worse still, when attempting to enable the missing effect, the plug-in manager says that it is Enabled (it isn't).

**To Reproduce**
1. Install a custom Nyquist plug-in
2. Enable the plug-in
3. Close Audacity
4. Uninstall (delete) the plug-in from the user's Plug-Ins folder
5. Restart Audacity
6. Open the plug-in manager
7. Observe that the effect is listed as "Disabled"
8. Enable the plug-in
9. Observe that the effect is now listed as "Enabled"
10. Close the plug-in manager
11. Observe that the plug-in does not appear in Audacity's menus
12. Open the plug-in manager
13. Observe that the plug-in is still listed as "**Enabled**"!

**Actual behavior**
Plug-ins may be listed as Enabled or Disabled when in fact they are totally absent.

**Expected behavior**
The plug-in manager to accurately report on which plug-ins are available, and whether they are enabled or disabled.

**Screenshots**
If applicable, add screenshots/video to help explain your problem.

**Additional information (please complete the following information):**
 - OS: Ubuntu 20.04
 - Version [e.g. Audacity 2.5.1] 3.2.0 6a5645 

**Additional context**

Ryan-Miller-Audacity:
I confirmed this bug in current public release Audacity 3.1.3 on macOS Monterey 12.3.1.

vsverchinsky:
> I confirmed this bug in current public release Audacity 3.1.3 on macOS Monterey 12.3.1.
> 
> Note that this doesn't appear to only effect Nyquist plugins as the title/description implies and appears to be more generalized to perhaps plugins in general. I tested this as well with a Blue Cat VST effect plug in and noted the same effects.

@Ryan-Miller-Audacity I've [created](https://github.com/audacity/audacity/pull/3288) a PR which addresses that issue for each plugin type we currently support

PR #3288, opened 2022-07-25
Issue 2965: plugin management improvements

vsverchinsky:
Resolves: #2965
Resolves: #3286

Not only Nyquist plug-ins are removed from the registry, but all other types of plug-ins are checked for existence as well.

Recommended:

### DIFF
--- a/libraries/lib-components/PluginProvider.h
+++ b/libraries/lib-components/PluginProvider.h
@@ -155,8 +155,6 @@ public:
       const RegistrationCallback &callback )
          = 0;
 
-   /*! @return true if the plug-in is still valid, otherwise false. */
-   virtual bool IsPluginValid(const PluginPath & path, bool bFast) = 0;
    /**
     * \brief Performs plugin/module existence check, still plugin may fail to load.
     * Implementation should avoid loading plugins during this check.
--- a/libraries/lib-components/PluginProvider.h
+++ b/libraries/lib-components/PluginProvider.h
@@ -171,7 +171,7 @@ public:
 // be declared static so as not to interfere with other providers during link.
 // ----------------------------------------------------------------------------
 #define DECLARE_PROVIDER_ENTRY(name)                  \
-static PluginProvider * name()
+static std::unique_ptr<PluginProvider> name()
 
 // ----------------------------------------------------------------------------
 // This will create a class and instance that will register the provider entry
@@ -203,11 +203,11 @@ static name name ## _instance;
 DECLARE_BUILTIN_PROVIDER_BASE(name)                   \
 void name::Register()                                 \
 {                                                     \
-   RegisterProvider(AudacityModule);                  \
+   RegisterProviderFactory(AudacityModule);                  \
 }                                                     \
 void name::Unregister()                               \
 {                                                     \
-   UnregisterProvider(AudacityModule);                \
+   UnregisterProviderFactory(AudacityModule);                \
 }
 
 #endif // __AUDACITY_MODULEINTERFACE_H__
--- a/libraries/lib-components/PluginProvider.h
+++ b/libraries/lib-components/PluginProvider.h
@@ -157,6 +157,13 @@ public:
 
    /*! @return true if the plug-in is still valid, otherwise false. */
    virtual bool IsPluginValid(const PluginPath & path, bool bFast) = 0;
+   /**
+    * \brief Performs plugin/module existence check, still plugin may fail to load.
+    * Implementation should avoid loading plugins during this check.
+    * \param path Internal plugin path/ID discovered via DiscoverPluginsAtPath
+    * or module path returned by FindModulePaths
+    */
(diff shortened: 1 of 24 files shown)


libraries/lib-components/PluginProvider.h
```diff
@@ -171,7 +171,7 @@ class COMPONENTS_API PluginProvider  /* not final */
 // be declared static so as not to interfere with other providers during link.
 // ----------------------------------------------------------------------------
 #define DECLARE_PROVIDER_ENTRY(name)                  \
-static PluginProvider * name()
+static std::unique_ptr<PluginProvider> name()
```
Paul-Licameli, line 179: First commit fails to build on macOS
Fix with
```
#include "PluginProvider.h"
```

in AudioUnitWrapper.cpp

libraries/lib-module-manager/ModuleManager.h
```diff
@@ -58,15 +58,28 @@ class Module
    fnModuleDispatch mDispatch;
 };
 
-struct PluginProviderDeleter {
-   void operator ()(PluginProvider *pInterface) const;
-};
+class MODULE_MANAGER_API PluginProviderUniqueHandle final
```
Paul-Licameli, line 61: I think this could be more simplify defined just as `std::unique_ptr<PluginProvider, PluginProviderDeleter>` and just supply that deleter class

libraries/lib-components/PluginProvider.h
Paul-Licameli: Also need a `make_unique` in AudioUnitEffectsModule.cpp as for the other factories.

Also:  commit comment has a spelling mistake

libraries/lib-components/PluginProvider.h
```diff
@@ -157,6 +157,13 @@ class COMPONENTS_API PluginProvider  /* not final */
 
    /*! @return true if the plug-in is still valid, otherwise false. */
    virtual bool IsPluginValid(const PluginPath & path, bool bFast) = 0;
+   /**
+    * \brief Performs plugin/module existence check, still plugin may fail to load.
+    * Implementation should avoid loading plugins during this check.
+    * \param path Internal plugin path/ID discovered via DiscoverPluginsAtPath
+    * or module path returned by FindModulePaths
+    */
+   virtual bool CheckPluginExist(const PluginPath& path) const = 0;
```
Paul-Licameli, line 164: `CheckPluginExists` might be a better name.  This commit fails to build on macOS because `AudioUnitEffectsModule` needs to define an override.

libraries/lib-components/PluginProvider.h
vsverchinsky: > First commit fails to build on macOS
> Also need a make_unique in AudioUnitEffectsModule.cpp as for the other factories.

Sorry, I completely forgot to fix that, will do soon

libraries/lib-module-manager/ModuleManager.h
vsverchinsky: I've attempted to replace `safenew` and `delete` with just std::make_unique<>. Though the first commit isn't neccessary part of this PR and I remove it

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -1162,43 +1163,16 @@ std::map<wxString, std::vector<wxString>> PluginManager::CheckPluginUpdates()
...
-      else if (plugType != PluginTypeStub)
+   for(auto& [id, provider] : moduleManager.Providers())
+   {
+      const auto paths = provider->FindModulePaths(*this);
+      for(const auto& path : paths)
       {
-         plug.SetValid(mm.IsPluginValid(plug.GetProviderID(), plugPath, false));
-         if (!plug.IsValid())
-            plug.SetEnabled(false);
+         const auto modulePath = path.BeforeFirst(';');
```
Paul-Licameli, line 1173: Not new in this commit -- but PluginManager is doing some interpretation of the paths returned by `FindModulePaths`.  So the comment in PluginProvider.h about `FindModulePaths` is not correct.  It is expected paths can be partly interpreted in a way not specific to the provider, at least when there is a semicolon.

libraries/lib-components/PluginProvider.h
```diff
@@ -155,8 +155,6 @@ class COMPONENTS_API PluginProvider  /* not final */
       const RegistrationCallback &callback )
          = 0;
 
-   /*! @return true if the plug-in is still valid, otherwise false. */
-   virtual bool IsPluginValid(const PluginPath & path, bool bFast) = 0;
```
Paul-Licameli, line 159: And again, AudioUnitEffectsModule must update too, to keep this commit buildable

Paul-Licameli:
This completes a round of review

Paul-Licameli:
"refactoring" misspelled in first commit comment

Paul-Licameli (approved):

PR #3301, opened 2022-07-28
Lib audio graph

Paul-Licameli:
Classes recently factored out of PerTrackEffect are moved to other files,
most of them in a new library, lib-audio-graph.

And one more class, AudioGraph::Task, implements the loop of PerTrackEffect::ProcessTrack.

Recommended:


crsib (approved):
It appears that <cstring> is missing from AudioGraphBuffers.cpp

Please, fix it before merging

PR #3317, opened 2022-07-29
Effect plugin window beautifications

ksoze95:
Resolves: #3090 

Recommended:


src/effects/EffectUI.cpp
```diff
@@ -309,15 +309,24 @@ int EffectUIHost::ShowModal()
    mApplyBtn->Destroy();
    mApplyBtn = apply;
    mApplyBtn->SetDefault();
-   mApplyBtn->SetLabel(wxGetStockLabel(wxID_OK, 0));
+
+   if (mIsBatch)
+      mApplyBtn->SetLabel(wxGetStockLabel(wxID_OK, 0));
+   else if (mEffectUIHost.GetDefinition().GetType() == EffectTypeGenerate)
+      mApplyBtn->SetLabel(XO("Generate").Translation());
```
crsib: I would like to see `XXO("&Generate")`. This way this string will be reused.

src/effects/EffectUI.cpp
```diff
@@ -309,15 +309,24 @@ int EffectUIHost::ShowModal()
    mApplyBtn->Destroy();
    mApplyBtn = apply;
    mApplyBtn->SetDefault();
-   mApplyBtn->SetLabel(wxGetStockLabel(wxID_OK, 0));
+
+   if (mIsBatch)
```
crsib, line 302: Probably a wrong condition?

src/effects/EffectUI.cpp
```diff
@@ -309,15 +309,24 @@ int EffectUIHost::ShowModal()
...
+   else if (mEffectUIHost.GetDefinition().GetType() == EffectTypeGenerate)
+      mApplyBtn->SetLabel(XO("Generate").Translation());
+
    mCloseBtn->SetLabel(wxGetStockLabel(wxID_CANCEL, 0));
 #else
-   mApplyBtn->SetLabel(wxGetStockLabel(wxID_OK));
+   if (mIsBatch)
+      mApplyBtn->SetLabel(wxGetStockLabel(wxID_OK));
+   else if (mEffectUIHost.GetDefinition().GetType() == EffectTypeGenerate)
+      mApplyBtn->SetLabel(XO("Generate").Translation());
```
crsib, line 312: Ditto

src/effects/EffectUI.cpp
```diff
@@ -357,73 +344,46 @@ wxPanel *EffectUIHost::BuildButtonBar(wxWindow *parent, bool graphicalUI)
...
-            {
-               mPlayToggleBtn = S.Id( kPlayID )
-                  .ToolTip(XO("Start and stop playback"))
-                  .AddButton( XXO("Start &Playback"),
-                              wxALIGN_CENTER | wxTOP | wxBOTTOM );
-            }
-            else if (
-               (mEffectUIHost.GetDefinition().GetType() != EffectTypeAnalyze) &&
-               (mEffectUIHost.GetDefinition().GetType() != EffectTypeTool) )
+            if (mwState.expired())
```
crsib, line 367: Wrong condition? After the #3391 mwState should not be used.

src/effects/EffectUI.cpp
```diff
@@ -357,73 +344,46 @@ wxPanel *EffectUIHost::BuildButtonBar(wxWindow *parent, bool graphicalUI)
...
             {
-               mPlayToggleBtn = S.Id( kPlayID )
-                  .ToolTip(XO("Preview effect"))
-                  .AddButton( XXO("&Preview"),
+               S.AddSpace( 5, 5 );
+
+               mPlayToggleBtn = S.Id(kPlayID)
+                  .ToolTip(XO("Start and stop preview"))
+                  .AddButton( XXO("Start &Preview"),
```
crsib, line 373: I have checked all the design sources - button is always called "Preview". Also, it should be "grouped" with the "Apply", right now it leans to the left (i. e. it is near the "Presets" button)

src/effects/RealtimeEffectState.h
```diff
@@ -87,6 +91,9 @@ class RealtimeEffectState
    //! Main thread cleans up playback
    bool Finalize() noexcept;
 
+   //! Notify all subscribers of a state change
+   void PublishStateUpdate(RealtimeEffectStateChange state);
```
crsib, line 95: You can do `using Observer::Publisher<RealtimeEffectStateChange>::Publish`, but I don't yet understand why this function is public

src/RealtimeEffectPanel.cpp
```diff
@@ -458,6 +469,25 @@ namespace
          effectStateUI.Toggle( *mProject );
       }
 
+      void OnEnableButtonClicked(wxCommandEvent& event)
+      {
+         auto pButton = static_cast<ThemedButtonWrapper<wxBitmapButton>*>(mEnableButton);
+         auto index = pButton->GetBitmapIndex();
+         bool wasEnabled = (index == bmpEffectOn);
+
+         if (mSettingsAccess) {
```
crsib, line 478: I think this should become a `RealtimeEffectState::SetActive`. That specific method should also Publish the message

crsib (changes requested):
* Opening fancy UI from the RT panel crashes Audacity
* Using Beat Finder crashes
* Preview button is in a wrong place
* Some other feedback in comments

Also I want a design review to be completed first. "Non fancy" UI looks quite odd, especially with built-in generators. Probably we need a horizontal line to visually separate the buttons from effect settings.

src/effects/EffectUI.cpp
LWinterberg: Agree on calling it "preview" (changing to "stop preview" after it's been pressed). Regarding the grouping, it apparently was difficult and martin didn't mind it ungrouped iirc.

crsib: Well, I know how to do such grouping :-)

src/effects/RealtimeEffectState.h
ksoze95: Its also called inside EffectUIHost::OnEnable to signal back to the realtime side panel that there has been a state change.

src/effects/EffectUI.cpp
ksoze95: Done.

src/effects/EffectUI.cpp
ksoze95: Changed. to XXO("&Generate").

src/effects/RealtimeEffectState.h
ksoze95: Done.

src/RealtimeEffectPanel.cpp
ksoze95: Done.

src/effects/EffectUI.cpp
ksoze95: Changed to use mpTempProjectState instead, plus tacked on a few other condition checks as well.

src/effects/RealtimeEffectState.cpp
```diff
@@ -545,6 +545,20 @@ bool RealtimeEffectState::IsActive() const noexcept
    return mWorkerSettings.extra.GetActive();
 }
 
+void RealtimeEffectState::SetActive(bool active)
+{
+   if (auto access = GetAccess()) {
```
vsverchinsky, line 550: `GetAccess` does not return empty pointer

src/effects/RealtimeEffectState.h
```diff
@@ -20,17 +20,21 @@
...
 #include "GlobalVariable.h"
 #include "MemoryX.h"
+#include "Observer.h"
 #include "PluginProvider.h" // for PluginID
 #include "XMLTagHandler.h"
 
 class EffectSettingsAccess;
 class Track;
 
+enum class RealtimeEffectStateChange : char { EffectOff, EffectOn };
```
crsib, line 30: Do we really need `char` here?

src/effects/EffectUI.cpp
```diff
@@ -1278,17 +1113,18 @@ std::shared_ptr<EffectInstance> EffectUIHost::InitializeInstance()
...
    return result;
 }
 
 void EffectUIHost::CleanupRealtime()
 {
-   mSubscription.Reset();
+   mAudioIOSubscription.Reset();
 
    if (mSupportsRealtime && mInitialized) {
       if (mpTempProjectState) {
```
crsib, line 1125: That not yet created method here please as well

crsib (changes requested):
There is still some feedback to addressed left. We can waive that enable button bug for now, I don't want it to block this PR.

(But we need to log that bug)

src/effects/EffectUI.cpp
ksoze95: Done.

src/effects/EffectUI.cpp
```diff
@@ -355,111 +339,44 @@ wxPanel *EffectUIHost::BuildButtonBar(wxWindow *parent, bool graphicalUI)
...
       S.SetBorder( margin );
 
-      if (!mIsGUI)
-      {
-         mMenuBtn = S.Id( kMenuID )
-            .ToolTip(XO("Manage presets and options"))
-            .AddButton( XXO("&Manage"), wxALIGN_CENTER | wxTOP | wxBOTTOM );
-      }
-      else
+      if (!mpTempProjectState && mSupportsRealtime)
```
crsib, line 342: `IsOpenedFromEffectPanel`? Minor though

src/effects/EffectUI.cpp
```diff
@@ -355,111 +339,44 @@ wxPanel *EffectUIHost::BuildButtonBar(wxWindow *parent, bool graphicalUI)
...
-                  .ToolTip(XO("Skip forward"))
-                  .Name(XO("Skip &Forward"))
-                  .AddBitmapButton( CreateBitmap(
-                     effect_ffwd_xpm, true, true) );
-               mFFwdBtn->SetBitmapDisabled(
-                  CreateBitmap(effect_ffwd_disabled_xpm, true, false));
-               mFFwdBtn->SetBitmapPressed(CreateBitmap(effect_ffwd_xpm, false, true));
-            }
-
             S.AddSpace( 5, 5 );
```
crsib, line 374: It appears that both branches are identical now

crsib (approved):
I think I can approve it now, with a couple minorish problems. Please rebase the branch on master first

PR #3327, opened 2022-07-31
Rewrite mixer

Paul-Licameli:
Preliminary cleanups of the Mixer class in lib-sample-track,
before we fix the rendering of effects by reuse of the new AudioGraph
classes, which wil be separated from PerTrackEffect.cpp.

Recommended:


crsib (approved):

Paul-Licameli:
Preliminary cleanups of the Mixer class in lib-sample-track,
before we fix the rendering of effects by reuse of the new AudioGraph
classes, which have been separated from PerTrackEffect.cpp.

PR #3330, opened 2022-08-02
several WavPack import/export fixes

dbry:
Hi Audacity team!

First I would like to thank the Audacity team, and particularly @Subhra264, for integrating WavPack into Audacity. Great work!

Unfortunately at the time I was not aware that this was ongoing because I found out about it the day it was committed to master. I'm sure that if I'd been aware I could have helped it along and perhaps saved some headaches. But in any event, I have been testing it now for the past week or so and have discovered and and fixed some minor issues and put together these eight commits.

Probably the easiest way to review them is to look at the individual commits. Most of them were pretty trivial to find (like files from the [official WavPack test suite](http://www.rarewares.org/wavpack/test_suite.zip) that simply would not load), and while I'm happy to go into more detail, these should all be pretty self-explanatory. One issue, however, requires a little more explanation.

The hybrid mode of WavPack was configured with a bitrate from 16 - 384 `kbps` (kbits per second). The problem with this is that this fixed range does not take into account the sample rate and number of channels (which directly affect the bitrates available), and therefore presents sometimes inappropriate values. For example, with 24/96 stereo audio the minimum bitrate that the hybrid mode will work at is around 430 kbps, and 6-channel, 48-kHz audio will only go down to 650 kbps. So for these common audio formats the entire range of presented bitrates is too low, which essentially makes the feature useless.

The solution I came up with is to present the bitrate in `bits per sample` instead, ranging from 2.2 (the minimum the hybrid mode will operate) to 8.0 (the practical upper limit which is always audibly transparent). I believe that this makes the feature usable while still presenting something easily understood by the user.

Please let me know if I can explain any of these changes more fully, and thanks again for this contribution!

Recommended:

### DIFF
--- a/src/export/ExportWavPack.cpp
+++ b/src/export/ExportWavPack.cpp
@@ -80,6 +80,7 @@ ExportWavPackOptions::~ExportWavPackOptions()
 
 const TranslatableStrings ExportQualityNames{
    XO("Low Quality (Fast)") ,
+   XO("Normal Quality") ,
    XO("High Quality (Slow)") ,
    XO("Very High Quality (Slowest)") ,
 };
@@ -88,6 +89,7 @@ const std::vector< int > ExportQualityValues{
    0,
    1,
    2,
+   3,
 };
 
 namespace {
@@ -315,10 +317,10 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
 
    if (quality == 0) {
       config.flags |= CONFIG_FAST_FLAG;
-   } else if (quality == 1) {
+   } else if (quality == 2) {
       config.flags |= CONFIG_HIGH_FLAG;
-   } else {
-      config.flags |= CONFIG_VERY_HIGH_FLAG;
+   } else if (quality == 3) {
+      config.flags |= CONFIG_HIGH_FLAG | CONFIG_VERY_HIGH_FLAG;
    }
 
    if (hybridMode) {
--- a/src/export/ExportWavPack.cpp
+++ b/src/export/ExportWavPack.cpp
@@ -421,7 +421,7 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
          WavpackAppendTagItem(wpc,
                               n.mb_str(wxConvUTF8),
                               v.mb_str(wxConvUTF8),
-                              static_cast<int>( v.length() ));
+                              static_cast<int>( strlen(v.mb_str(wxConvUTF8)) ));
       }
 
       if (!WavpackWriteTag(wpc)) {
--- a/src/export/ExportWavPack.cpp
+++ b/src/export/ExportWavPack.cpp
@@ -310,10 +310,16 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
 
    config.num_channels = numChannels;
    config.sample_rate = rate;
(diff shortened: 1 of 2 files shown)


src/export/ExportWavPack.cpp
```diff
@@ -429,7 +406,7 @@ ProgressResult ExportWavPack::Export(AudacityProject *project,
          WavpackAppendTagItem(wpc,
                               n.mb_str(wxConvUTF8),
                               v.mb_str(wxConvUTF8),
-                              static_cast<int>( v.length() ));
+                              static_cast<int>( strlen(v.mb_str(wxConvUTF8)) ));
```
crsib, line 409: Great catch!

src/import/ImportWavPack.cpp
```diff
@@ -104,11 +105,12 @@ TranslatableString WavPackImportPlugin::GetPluginFormatDescription()
...
 {
    char errMessage[100]; // To hold possible error message
-   int flags = OPEN_WVC | OPEN_FILE_UTF8 | OPEN_TAGS;
+   int flags = OPEN_WVC | OPEN_FILE_UTF8 | OPEN_TAGS | OPEN_DSD_AS_PCM | OPEN_NORMALIZE;
    WavpackContext *wavpackContext = WavpackOpenFileInput(filename, errMessage, flags, 0);
    
    if (!wavpackContext) {
       // Some error occured(e.g. File not found or is invalid)
+      wxLogDebug("WavpackOpenFileInput() failed on file %s, error = %s", filename, errMessage);
       return nullptr;
```
crsib, line 114: Probably `wxLogError` is a better idea. `wxLogDebug` does not generate any user visible output really.

src/import/ImportWavPack.cpp
```diff
@@ -251,12 +257,14 @@ ProgressResult WavPackImportFileHandle::Import(WaveTrackFactory *trackFactory, T
             itemLen = WavpackGetTagItemIndexed(mWavPackContext, i, NULL, 0);
             std::string item (itemLen + 1, '\0');
             WavpackGetTagItemIndexed(mWavPackContext, i, item.data(), itemLen + 1);
+            item.resize(itemLen);           // remove terminating NULL from std::string
```
crsib, line 260: I'm not sure why do is this needed. This is safe here, though, so nothing critical

crsib (approved):

src/import/ImportWavPack.cpp
dbry: I didn't really intend this for the user, especially because this gets triggered any time the WavPack importer is invoked to check whether some random file is importable. However, it might be useful for debugging if `libwavpack` refuses to open a file that it _should_ be able to.

src/import/ImportWavPack.cpp
dbry: I'm not a regular C++ programmer and just discovered that std::strings can include terminating NULLs. The issue this caused was that when the tag name had a NULL at the end it would no longer match the same string without a NULL. So, for example, this line did not work as intended:

https://github.com/audacity/audacity/blob/2c3c9d9919b778bb27cc94d99ece211d9686c12f/src/TagsEditor.cpp#L462

dbry:
@crsib Thanks for the prompt review and merge! I'll continue testing, but I think this is pretty good at this point.

crsib:
@dbry thank you a lot for your help!

PR #3334, opened 2022-08-03
Mixer source

Paul-Licameli:
Last refactoring step before really making rendering of effects work.

Extract an AudioGraph::Source subclass from Mixer, which will be adapted with EffectStage.

Recommended:


libraries/lib-sample-track/Mix.h
```diff
@@ -271,5 +253,86 @@ class SAMPLE_TRACK_API Mixer {
...
+ reference to its base class.
+ */
+class MixerSource final : public AudioGraph::Source {
+public:
+   using TimesAndSpeed = MixerOptions::TimesAndSpeed;
+
+   /*!
+    @pre `pTimesAndSpeed != nullptr`
+    */
+   MixerSource(const SampleTrack &leader, size_t i,
```
crsib: Generally I'm not a fan of such a long arguments lists,

libraries/lib-sample-track/Mix.cpp
```diff
@@ -210,8 +210,10 @@ double ComputeWarpFactor(const Envelope &env, double t0, double t1)
 }
 
 size_t MixerSource::MixVariableRates(
-   size_t ii, const size_t maxOut, float &floatBuffer)
+   size_t j, const size_t maxOut, float &floatBuffer)
```
crsib: `j` is as bad as `ii` to be fair.

libraries/lib-sample-track/Mix.cpp
```diff
@@ -210,8 +210,10 @@ double ComputeWarpFactor(const Envelope &env, double t0, double t1)
...
 {
+   const auto ii = i + j;
```
crsib: In the context of this review I have no clue what is i, j and ii.

libraries/lib-sample-track/Mix.cpp
```diff
@@ -571,13 +589,7 @@ size_t Mixer::Process(const size_t maxToProcess)
...
       auto increment = finally([&]{ i += nInChannels; });
 
-      MixerSource source{ *leader, i, mRate,
-         mResampleParameters.mbVariableRates, mEnvelope, mMayThrow
-         , mTimesAndSpeed
-         , mInputTrack, mSamplePos, mSampleQueue, mQueueStart
-         , mQueueLen, mResample, mEnvValues
-      };
-      auto oResult = source.Acquire(mFloatBuffers, maxToProcess);
+      auto oResult = pSource++ ->Acquire(mFloatBuffers, maxToProcess);
```
crsib: Looks quite awkward to be fair

crsib (approved):

libraries/lib-sample-track/Mix.h
Paul-Licameli: Of course not, and the purpose of later commits is to shorten them after that rough separation

libraries/lib-sample-track/Mix.cpp
```diff
@@ -34,33 +34,23 @@
...
-   {
-      wxASSERT(false);
-      maxSpeed = 0;
-   }
-   if (minSpeed > maxSpeed)
-   {
-      wxASSERT(false);
-      std::swap(minSpeed, maxSpeed);
-   }
+   assert(min >= 0);
```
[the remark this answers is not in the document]
Paul-Licameli: minSpeed and maxSpeed were reassigned, but not the arguments min, max.  I preserve the assertions, and make explicit the precondition in the header, while also making this struct an immutable -- therefore, it doesn't need private members or friends.  The constructor establishes the invariant property of the structure which won't be violated thereafter, barring dirty tricks with casts.

Issue #3323, opened 2022-07-30
Calf Reverb crashes (regression)


SteveDaulton:
**Describe the bug**
When attempting to apply Calf  Reverb as a destructive effect, Audacity crashes.

**To Reproduce**
1. Generate a tone
2. Apply Calf Reverb
3. Crash

**Actual behavior**
Crash

**Expected behavior**
To apply Reverb without crashing

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version audacity-linux-3.2.0-alpha-20220729+af44406-x64.AppImage

**Additional context**
Regression against Audacity 3.1.3

dozzzzer:
I was able to reproduce this on Ubuntu 20.04

Ryan-Miller-Audacity:
@crsib Testing this in latest master (https://github.com/audacity/audacity/actions/runs/2799554924) in my Linux VM I am still observing a crash when opening the Calf Reverb plugin via the destructive effects menu. I note the same behavior in a previous master release as well as on Audacity 3.1.3. I am wondering if this is more so an issue with the set up of my linux VM? 

Here is a link to the crash in sentry: https://sentry.audacityteam.org/share/issue/a12dfd276cf64d01929fc2971db27fcd/

crsib:
That is different from a previous crash and I cannot reproduce it. This looks to be some stack overflow inside the GTK, so probably this is related to your Linux distro. I've been testing in regular (almost) version of 20.04.

Ryan-Miller-Audacity:
With me being unable to successfully test this due to the crash I experience noted above (which appears to be unrelated to this issue and has to do with my Linux VM), @dozzzzer can you take a look at this and see if you can determine that it is successfully fixed?

dozzzzer:
Looking good on Ubuntu 20.04!

PR #3336, opened 2022-08-04
Fixes uninitalized mBuffer in LV2AtomPortState

crsib:
Fixes uninitalized mBuffer in LV2AtomPortState
`LV2Wrapper::Create` performs `ReceiveFromInstance` for every atom port. However, mBuffer contents was never initialized before that. 

`zix_ring_write` expects that the first two DWORDS are well defined, not
doing so resulted in UB.

Resolves: #3323 

Recommended:


Paul-Licameli (approved):

Paul-Licameli:
Thank you for figuring that out

PR #3337, opened 2022-08-04
Unify resource loading location

RuRo:
Resolves: #2659 (specifically [this comment](https://github.com/audacity/audacity/issues/2659#issuecomment-1152921419))

Before this PR the `res/effects_menu_defaults.xml` file was searched relative to the `audacity` executable and wasn't even shipped on Linux. Additionally, the `presets/EQDefaultCurves.xml` had custom logic to get it installed in the right place.

Both the `presets` and the `res` directories didn't have any files in them other than these 2 files. I think, that both of these files fall under the broad category of "resources" and should ideally be all installed in a unified way into `${prefix}/share/audacity` on linux, `audacity.app/Contents/Resources` on Mac and alongside the executable on Windows. We can then uniformly retreive them using `FileNames::ResourcesDir()`.

In the future, to add a new resource file, all you will need to do would be to add it to `resources` and add it to this list in `src/CMakeLists.txt`:
```cmake
# Define resources
#
list( APPEND RESOURCES
   ../resources/EQDefaultCurves.xml
   ../resources/EffectsMenuDefaults.xml
)
```

**Disclaimer**: I don't have Windows/Mac machines, so this PR is currently only tested on Linux (Arch). We should probably get someone to verify, that the resources are bundled (and subsequently found) correctly in all the supported distribution formats. A simple way to verify this is to check if the `Effects` menu properly groups the built-in effects:

<details><summary>Right</summary>

Screenshot_20220804_230316

</details>

<details><summary>Wrong</summary>

Screenshot_20220804_231027

</details>

---

Recommended:


vsverchinsky (approved):

LWinterberg:
This change broke the Fedora package build apparently. 

https://github.com/audacity/audacity/actions/runs/2868753432

``` 
error: Installed (but unpackaged) file(s) found:

   /usr/share/audacity/EffectsMenuDefaults.xml

    File not found: /home/user/rpmbuild/BUILDROOT/audacity-3.2.0.alpha.r7d3fd92-0.fc34.x86_64/usr/share/audacity/help
RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/share/audacity/EffectsMenuDefaults.xml
```

RuRo:
No idea about how rpm packaging works. It looks like you need to declare the installed files explicitly in the `%files` directive?

Regarding #3386, I have mentioned in the Disclaimer that this should be tested on Mac/Windows before merging, but I guess that was accidentally overlooked. Either the windows build doesn't install the resources, or the location is different from what I've expected, and the code can no longer find them. Unfortunately, I can't help you with debugging this on Windows as I only run Linux.

Issue #3313, opened 2022-07-29
Rewrite VST3 inter-thread communication using EffectSettingsAccess


Paul-Licameli:
Main thread uses `EffectSettingsAccess::Set` to send parameter changes to the worker thread, in reimplemented `ComponentHandler` subclass.
VST3 effect settings may need to carry extra information.
Worker thread receives the messages in `RealtimeProcessStart`.

Issue #3308, opened 2022-07-29
Real statelessness for VST3


Paul-Licameli:
Some steps (and analogous commits)

- [ ] Really stateless (squash following into one commit when done) (014836dd)
  - [ ] Define appropriate `MakeSettings`, and a `CopySettingsContents` that does not allocate memory
  - [ ] `RealtimeProcessStart` stores settings into Instances
  - [ ] `mSettings` member of Effect is removed, and `GetSettings` is rewritten as a static function really using its argument

Issue #3307, opened 2022-07-29
define the Validator class for VST3


Paul-Licameli:
Things to do (not necessarily in this sequence), with some analogous older commits

- [ ] Define minimal Instance class -- holding references to Instance and EffectSettingsAccess (43d9887c)
- [ ] Validator can fetch settings from, store to, Instance (And stores on construction; if needed to initally populate dialog values correctly) (7d1c0c6f)
- [ ] Effect's `TransferDataToWindow` becomes Validator's `UpdateUI` override; `ValidateUI` moves into Validator, uses `mAccess.ModifySettings`, fetches settings from Instance
- [ ] Plain UI controls, event handlers moved into Validator; all use ModifySettings, maybe just invoking ValidateUI (no analogy with AudioUnits)
- [ ] Native UI event and idle callbacks to host are handled by validator; parameter value changes invoke `mAccess.ModifySettings` (bae543a9)

Issue #3306, opened 2022-07-29
define the Instance class for VST3


Paul-Licameli:
The Instance might be done before the Validator.  That is not how things progressed with built-ins.  But if you look at the dependency graphs for the source files implementing AudioUnits or LV2, you see that the validator depends on the instance, not the opposite.

So really the Instance is the more basic object, which could be extracted from existing code first, although it may also be necessary to define the Validator partially so that Instance can be defined completely.

Suggested outline of steps, by analogy with the recent LV2 transformation

- [x] 1. (As needed, perhaps not) Definine a minimal Validator class (see 72f3c002)
- [x] 2. Define minimal Instance class (8ec84fa7)
- [x] 3. Define temporary overloads of `GetSettings` in scope of the Instance (e6d6ed95)
- [x] 4. Instance doesn't inherit `StatefulPerTrackEffect::Instance`; move `ProcessInitialize`, `ProcessBlock`, `ProcessFinalize`, `GetLatency` from Effect to Instance class (ffbec4a8); and effect inherits `PerTrackEffect` and `StatefulEffectBase` but not `StatefulPerTrackEffect`
- [ ] 5. Instance doesn't inherit `StatefulEffectBase::Instance`; move other necessary overrides into Instance class; make slaves instances (76028f62); effect no longer inherits `StatefulEffectBase`
- [ ] 6. (Only needed if output metering doesn't work:  Instance re-uses itself, not a slave, for group number 0 in RealtimeProcess (3447e35d))

Paul-Licameli:
The sequence of events that happens in response to `LoadUserPreset`:

- it is called, and calls `setState`, updating an EffectComponent in the Effect object
- `FetchSettings` happens at the end of the function, pulling information from the component into `VST3EffectSettings`
- Control returns to `EffectUIHost::OnUserPreset` which, as now, calls `TransferDataToWindow`
- That calls the Effect's `TransferDataToWindow` and then the Validator's `UpdateUI`
- `VST3Effect::TransferDataToWindow` (or, after the next task, `VST3Validator::UpdateUI`) calls `StoreSettings` to retrieve information from `VST3Settings`, and store it into (another) EffectComponent

For now there is in fact only one EffectComponent.  But what I think will happen, is that the Effect and each Instance will have a disctinct EffectComponent.

For transformation of some of the effect families, like AudioUnits, it wasn't really possible to do the EffectInstance class completely without also doing the Validator class at least partially, and I think that is the case here.

[...]

PR #3338, opened 2022-08-04
I3306 vst3 instance

vsverchinsky:
Resolves: #3306
Resolves: #3313
Resolves: #3307 
Resolves: #3308

Workaround for UI display can't handle multiple non-modal dialogs for same effect.

Recommended:


src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+#include "AudacityVst3HostApplication.h"
+
+#include <pluginterfaces/vst/ivsteditcontroller.h>
+
+#include "memorystream.h"
+
+#include "VST3Utils.h"
+#include "internal/ConnectionProxy.h"
+#include "internal/ComponentHandler.h"
```
Paul-Licameli, line 17: First commit needs this to build:
`#include <wx/string.h>`

src/effects/VST3/VST3Wrapper.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
+#include <pluginterfaces/base/smartpointer.h>
+#include <pluginterfaces/vst/ivstaudioprocessor.h>
+#include <public.sdk/source/vst/hosting/module.h>
+#include "internal/ComponentHandler.h"
```
Paul-Licameli, line 6: First commit needs `#include <optional>`

src/effects/VST3/VST3Effect.cpp
```diff
@@ -181,18 +181,41 @@ bool VST3Effect::SupportsAutomation() const
    return true;
 }
 
-bool VST3Effect::SaveSettings(const EffectSettings&, CommandParameters& parms) const
+bool VST3Effect::SaveSettings(const EffectSettings& settings, CommandParameters& parms) const
```
Paul-Licameli, line 163: The commit "Attempt to address..." does not compile, but the following commit repairs that

libraries/lib-components/EffectInterface.h
```diff
@@ -408,6 +408,15 @@ class COMPONENTS_API EffectInstance
...
 
+   //! How many input buffers to allocate at once
+   /*!
+    If the effect ALWAYS processes channels independently, this can return 1
+    */
+   virtual unsigned GetAudioInCount() const = 0;
+
+   //! How many output buffers to allocate at once
+   virtual unsigned GetAudioOutCount() const = 0;
+
```
Paul-Licameli, line 419: The commit that moves this virtual function to the Instance class is surprising.  I thought the number of channels in and out is a property of a plug-in independent of its instances.  What problem led to this change?

vsverchinsky: In VST3 this property belongs to the instance. Also the number of IO channels is configurable

Paul-Licameli: To understand that, I need more explanation, or an example of a plug-in.

I'm not comfortable with this change.  I need to reexamine all the uses of these virtual functions to be sure they make sense.

Alternatively, if the number of channels in and out can be described as a function of EffectSettings, independently of an Instance state, that might make more sense.

I can make modifications of the branch so that it can build without this commit.  That is, I have not discovered where you need to call these functions as part of the implementation of other overridden virtual functions in the instance.

Paul-Licameli: I'm reconsidering this, and I think you do make a change for the better.  I have noticed some confusion about the meaning of these virtual functions.

But ultimately, `EffectInstance::ProcessBlock` can be reached (for destructive effects) or `EffectInstance::RealtimeProcess` and those functions are not given a count of channels.  Instead the instance assumes that it "knows" the right number of buffers in and out.

I think I want to cherry-pick this commit, and do some other things relating to these virtual functions, and propse a PR to be merged first, and then the rest of your work can rebase onto that.

Paul-Licameli: However there should be some kind of contract expaining when it is permissible to call these virtual functions on an instance.

Where does the change of VST3 instance state happen, that determines the result of `CountChannels`?  It is not at construction time of the Instance, because no information is given to `MakeInstance`.  It must be one of the later functions that pass Settings, yes?

vsverchinsky: Sorry, I mean that it's possible in VST3 to change channel arrangement dynamically, but we don't(yet) support this feature.
https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Change+History/3.0.0/Multiple+Dynamic+IO.html?highlight=channel#audio-bus-channel-configuration

Issue #3112, opened 2022-06-20
An audible "tail" can be heard when switching an effect off and then back on


dozzzzer:
- [x] AudioUnit (e. g. AUMatrixReverb)
- [x] Ladspa (e.g. GVerb from swh plug-ins)
- [ ] LV2 (e.g. any of the Dragonfly reverbs)
- [ ] VST (not yet testable)
- [ ] VST3 (not yet testable)

@Paul-Licameli 
When the effect is turned off during the playback, it leaves some "residual" tail of audio.
It can be properly observed on a DTMF tones track where signal and silence are alternating.
So the trick is to start playback having the effect turned on and then turn it off in the middle of the playback of the tone:

https://user-images.githubusercontent.com/70063970/174297861-4c81bcc0-4d38-423c-b22a-c96fb186e8c7.mp4

And then, when clicking the toggle again during the silence part, you can hear an audible "tail" while there should be silence.

_Originally posted by @dozzzzer in https://github.com/audacity/audacity/issues/3060#issuecomment-1158821687_

Ryan-Miller-Audacity:
FWIW, I was able to reproduce as described above as well.

crsib:
Does it only affects reverb @dozzzzer?

crsib:
@Paul-Licameli FYI^

dozzzzer:
The above-mentioned issue can still be observed when using the built-in Reverb plugin.

Paul-Licameli:
> The above-mentioned issue can still be observed when using the built-in Reverb plugin.

Reverb will not be available in the side bar this release.  Please check the boxes in the description for the effect families for which you can demonstrate the fix.

dozzzzer:
For an LV2 effect (e.g. Dragonfly Reverb) the issue still remains:

https://user-images.githubusercontent.com/70063970/185418379-73709b36-f00e-4ed6-9710-243648b6b66e.mp4

Paul-Licameli:
@dozzzzer yes, I observe that with Dragonfly Plate, but also observe that Dragonfly Hall now works as expected after my fix.

I think you should check off the box if you can demonstrate that at least one effect of the family was not clearing the tail before my fix, but now, it does.  That means our host side gets its part of the protocol right.  But if it fails still with some individual effects -- then I think we can blame those effets for not being correctly implemented.

dozzzzer:
@Paul-Licameli 
[...]

PR #3349, opened 2022-08-08
Flush delay buffers

Paul-Licameli:
Resolves: #3112 (except for VST2)

Fix realtime play of Ladspa, Lv2, and AudioUnit effects, so they properly clear delay buffers when you power an effect or
effect list off.

I believe VST3 required no fix here, but the right fix for VST2 isn't known to me yet.

Recommended:


crsib (approved):

Paul-Licameli:
Fix realtime play of Ladspa, Lv2, and AudioUnit effects, so they properly clear delay buffers when you power an effect or
effect list off.

I believe VST3 required no fix here, but the right fix for VST2 isn't known to me yet.  Testing those should wait until they are made stateless.

PR #3354, opened 2022-08-09
Fix shadows

Paul-Licameli:
Change clang compiler options so that variables shadowing others in surrounding scope are disallowed.

Make other source code changes needed to fix the build.

Commits divide the fixes for review among different areas of responsibility but will be squashed into one when approved.

This may help catch more errors at development time.

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
@@ -1066,7 +1066,7 @@ void WaveformView::Draw(
    WaveTrackSubView::Draw( context, rect, iPass );
 }
 
-static const WaveTrackSubViews::RegisteredFactory key{
+static const WaveTrackSubViews::RegisteredFactory myey{
    []( WaveTrackView &view ){
       return std::make_shared< WaveformView >( view );
    }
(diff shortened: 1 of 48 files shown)


crsib:
I would like to see it to be enabled for all compilers

pietro68:
I approve the commit for shadowings for pietro68 - shall we do it so that each dev reviews his own section?

crsib:
I will probably push a commit on top though, that will enable this behavior for all the compilers.

crsib:
I have pushed a proposed chage

crsib:
After more careful consideration I have decided not to enable `-Werror=shadow` for GCC right now. It appears that GCC has stricter rules than Clang and MSVC and much more code changes are required. Probably to much fuss for now

src/widgets/Grid.cpp
```diff
@@ -551,45 +551,48 @@ void Grid::OnEditorShown(wxGridEvent &event)
...
-                  return;
+                  wxTextDataObject data;
+                  if (wxClipboard::Get()->GetData(data))
+                  {
+                     SetCellValue(crow, ccol, data.GetText());
+                     return;
+                  }
                }
             }
+            break;
```
ksoze95, line 594: I think this break should be move to the next level out, so that it is consistent with the other case statement blocks.

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -1066,7 +1066,7 @@ void WaveformView::Draw(
    WaveTrackSubView::Draw( context, rect, iPass );
 }
 
-static const WaveTrackSubViews::RegisteredFactory key{
+static const WaveTrackSubViews::RegisteredFactory myey{
```
ksoze95, line 1069: Not that it really matters, but should this have been changed to myKey.

Issue #3315, opened 2022-07-29
Real-time preview for some effects sound different to applied effect on mono tracks


SteveDaulton:
**Describe the bug**
For some real-time preview effects, the preview sounds very different from the applied effect when applied to a mono track.

See updates from Paul below

**To Reproduce**
1. Preview the LADSPA Freeverb effect (available in the Ubuntu repository) on a mono track.
2. Apply
3. Close the effect

**Actual behavior**
The preview sound is about twice as loud as the track after applying.

**Expected behavior**
The applied effect to sound the same as the preview.

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version Audacity 3.2.0 AppImage

**Additional context**
I think the clue is in Freeverb being a stereo effect, and the problem occurring with mono tracks.

@Paul-Licameli update:  Try this again now that #3022 is fixed.  There is still a problem.
- Generate DTMF
- Duplicate the track
- Add any of several Reverb effects to the stack of the first track (AUMatrixReverb for example)
- Apply the same effect destructively the the second track with the same settings
- Apply the Invert effect to the second track
- Play
Expect: silence (because the second track should exactly negate the first)
Observe:  still some sound

dozzzzer:
I've managed to observe ~1 db difference, which is insignificant, when using Freeverb on a mono track:
<img width="1236" alt="image" src="https://user-images.githubusercontent.com/70063970/182153150-bf657ab2-fbe8-4f40-ae8a-141def5c7184.png">

dozzzzer:
When the **'wet level'** parameter is above zero I indeed see a fixed ~6 dB difference regardless of the wet level.

Paul-Licameli:
I think this bug is not new in 3.2.  Can you confirm that it happened also in 3.1.3?

SteveDaulton:
> I think this bug is not new in 3.2. Can you confirm that it happened also in 3.1.3?

I can't test Freeverb in Audacity 3.1.3 because Audacity 3.1.3 can't find it (not listed in the Plug-in Manager).

Paul-Licameli:
> > I think this bug is not new in 3.2. Can you confirm that it happened also in 3.1.3?
> 
> I can't test Freeverb in Audacity 3.1.3 because Audacity 3.1.3 can't find it (not listed in the Plug-in Manager).

Do you know if any of the swh plug-ins can demonstrate the same?

Paul-Licameli:
[...]

PR #3355, opened 2022-08-09
Fix mono reverbs

Paul-Licameli:
Resolves: #3315 

Make sure the inputs given to realtime play of a mono track are the same as given to destructive effect application and
rendering.

This makes a difference with reverb effects, which do not process channels independently, but rather have each output
channel affected by each input channel.

Recommended:


pietro68:
> The fix was to change the calls to alloca, substituting `nBuffers` for `chans`. Please fetch the branch and try it again.

I confirm that it works now.

pietro68 (approved):
Verified fix works and code changes are ok for me.

Issue #3356, opened 2022-08-09
Rendering a track with effects should result in a cleared effect stack


Paul-Licameli:
- Get some sound
- Put one or more effects in its stack
- listen
- Mix and render
- listen

Expect:  hear the same thing
Observe:  you may hear something different, because the rendered track still has the realtime effects, so the intended effect stack is applied twice.  If you open the effect stack and disable all effects -- then you would hear the same.

This matters when you
- [ ] Mix and render
- [ ] Mix and render to a new track
- [ ] Mix stereo down to mono

PR #3357, opened 2022-08-09
Clear effect stacks after Mix Down to Mono or (either) Mix-and-Render

Paul-Licameli:
Resolves: #3356

Mix Stereo Down To Mono, Mix And Render (to same or to new track)
now leave the resulting effect stacks empty

Recommended:

### DIFF
--- a/src/effects/RealtimeEffectList.cpp
+++ b/src/effects/RealtimeEffectList.cpp
@@ -148,6 +148,19 @@ void RealtimeEffectList::RemoveState(
    }
 }
 
+void RealtimeEffectList::Clear()
+{
+   decltype(mStates) temp;
+   
+   // Swap an empty list in as a whole, not removing one at a time
+   // Lock for only a short time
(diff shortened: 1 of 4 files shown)


src/effects/RealtimeEffectList.cpp
```diff
@@ -148,6 +148,19 @@ void RealtimeEffectList::RemoveState(
...
+{
+   decltype(mStates) temp;
+   
+   // Swap an empty list in as a whole, not removing one at a time
+   // Lock for only a short time
+   (LockGuard{ mLock }, swap(temp, mStates));
+
+   for (auto index = temp.size(); index--;)
+      Publisher<RealtimeEffectListMessage>::Publish({
+         RealtimeEffectListMessage::Type::Remove, index, { } });
```
crsib, line 161: This conflicts with #3176. Not a blocker for either PRs, but worth to be noted

crsib (approved):

Issue #2721, opened 2022-03-23
Wrong playing state after playing generated tone


pietro68:
**Describe the bug**
Generate a tone then play it; when its end is reached, it looks as if the play state is still "playing" (as far as the play/stop buttons are concerned)

**To Reproduce**
1) Start Audacity, no project loaded
2) Generate | DTMF tones...  (or silence, noise - any kind of generation)
3) press OK, tone will be generated
4) press Play button, let the generated tone be played until the end

**Expected behavior**
the play state is stopped, i.e. the play button is not higlighted

**Actual behavior**
the play state is "playing", i.e. the play button is higlighted as if the clip is still being played, and one must press the stop button to make the highlight go away.

**Additional information:**
 - OS: [Windows 10]
 - Version: this happens on master branch, since at least commit ab7a1a9 (but possibly also earlier than that - I just stopped backtracking at that commit)

LWinterberg:
may be fixed together with #2617

PR #3358, opened 2022-08-09
Play of non-point selection with no loop region should stop at end...

Paul-Licameli:
... This fixes a regression from 3.1.3, and is another example of a silly bug
that would have been caught if shadowed variables were flagged as errors.

Resolves: #2721

Recommended:


Paul-Licameli:
This fix is not sufficient yet.
It appears that if the selection is short, the bug persists.
It appears that the shortest duration to play that does work at the default sample rate is exactly 2 s + 44 samples, which is 2.001 s rounded to the nearest sample.

Paul-Licameli:
This amended PR should now do it even for short selections.

pietro68 (approved):
Verified and understood.

Issue #3279, opened 2022-07-22
Manual rescan is required to reveal some plugins in the Plugin Manager


dozzzzer:
**Describe the bug**
Some plugins are shipped through a standalone app used to install, manage, and update plugins and licenses (e.g. Waves plugins are shipped through Waves Central these days). When installing plugins using this app, Audacity only sees the plugins that were installed initially and doesn't see the consequent ones unless the `pluginregistry.cfg` is removed.

**To Reproduce**
1. Install Waves Central
2. Select a plugin in it and install
3. Run Audacity and see it being enabled
4. Install one or more plugins in Waves Central
5. Re-run Audacity to see if it picks them up

**Actual behavior**
No newly added plugins are added to Plugin Manager

**Expected behavior**
After discussing with @crsib it was agreed that the best possible option here would be to add an option to manually rescan the plugins by adding the 'Rescan' button into the Plugin Manager dialog.

dozzzzer:
The only question left is the placement of the button within Plugin Manager.
@Tantacrul what do you reckon: left or right? Or anywhere else maybe?

<img width="500" alt="image" src="https://user-images.githubusercontent.com/70063970/180471511-3c4c9815-3b47-47d9-838d-34767b3f605e.png">

Tantacrul:
I would have thought on the left, simply due to space. You don't want to overwhelm poor 'OK' and 'Cancel' 😄

crsib:
I think this would be a good task for Kit, but we will see how it goes

dozzzzer:
Verified

PR #3359, opened 2022-08-09
Added Rescan option to the PluginRegistrationDialog.

ksoze95:
Resolves: #3279 

Recommended:


libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -1143,6 +1143,19 @@ ComponentInterface *PluginManager::Load(const PluginID & ID)
...
 
+void PluginManager::ClearEffectPlugins()
+{
+   for (auto it = mRegisteredPlugins.cbegin(); it != mRegisteredPlugins.cend(); ++it)
+   {
+      auto& desc = it->second;
+      if (desc.GetPluginType() == PluginTypeEffect)
+      {
+         mLoadedInterfaces.erase(desc.GetID());
+         it = mRegisteredPlugins.erase(it);
```
Paul-Licameli, line 1154: Incorrect loop.  Increment the iterator or erase it, not both.  Or else you skip some elements.  May crash if you erase the last element.

Paul-Licameli (changes requested):

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -1143,6 +1143,19 @@ ComponentInterface *PluginManager::Load(const PluginID & ID)
    return nullptr;
 }
 
+void PluginManager::ClearEffectPlugins()
+{
+   for (auto it = mRegisteredPlugins.cbegin(); it != mRegisteredPlugins.cend(); ++it)
+   {
+      auto& desc = it->second;
+      if (desc.GetPluginType() == PluginTypeEffect)
```
vsverchinsky, line 1151: Maybe you need to erase plugins with type `PluginTypeStub` as well

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -1143,6 +1143,19 @@ ComponentInterface *PluginManager::Load(const PluginID & ID)
...
+      {
+         mLoadedInterfaces.erase(desc.GetID());
```
vsverchinsky, line 1153: Probably I was wrong when I said that effect should be erased from the `mLoadedInterfaces`. It may cause problems with effects being used in the project since they are unique ptrs. Removing description of the loaded plugin isn't good thing as well because there are code pathes that request plugin information from the `PluginManager`. Placing `PluginDescription` of the loaded plugins into special container in `PluginManager::ClearEffectPlugins` so that later calls to `PluginManager::GetPlugin`... could work

vsverchinsky (approved):

libraries/lib-module-manager/PluginManager.cpp
ksoze95: Done.  Moved iterator increment to the body of the for loop, and when not erasing.

Issue #3345, opened 2022-08-07
Crash when running Nyquist Prompt with code that has widgets


SteveDaulton:
**Describe the bug**
Running any code in the Nyquist Prompt that includes a ";control" widget crashes Audacity

**To Reproduce**
Minimal example - apply the following code in the Nyquist Prompt to a selection in an audio track:
```
;control dummy "Dummy control" int "" 5 0 10
(noise)
```

**Actual behavior**
Audacity crashes.

**Expected behavior**
White noise returned to the selected region

**Screenshots**
If applicable, add screenshots/video to help explain your problem.

**Additional information (please complete the following information):**
 - OS: Any
 - Version Audacity 3.2.0 alpha

**Additional context**
Regression.
`../source_subfolder/src/common/wincmn.cpp(1492): assert ""firstHandler->GetPreviousHandler() == __null"" failed in PopEventHandler(): the first handler of the wxWindow stack should have no previous handlers set`

Ryan-Miller-Audacity:
I was able to reproduce this as described. 

Here is a link to the sentry report for the crash as well: https://sentry.audacityteam.org/share/issue/56dc4b54c3024710903a5aade0bf4149/

Paul-Licameli:
I have made an alternative fix to Dmitry's, which is surprisingly simple, though I took some wrong turns finding it.

PR #3363, opened 2022-08-11
Fixes Nyquist Prompt for effects, producing UI

crsib:
`NyquistEffect` violates the assumption, that Effect outlives Validator in case the code from Nyquist Prompt generates another dialog. In this case, UI is shown for a stack allocated `NyquistEffect`, which is destroyed before the `EffectUI` destructor is called.

Resolves: #3345 

Recommended:

### DIFF
--- a/src/effects/nyquist/Nyquist.cpp
+++ b/src/effects/nyquist/Nyquist.cpp
@@ -119,6 +119,68 @@ static const wxChar *KEY_Parameters = wxT("Parameters");
 
 ///////////////////////////////////////////////////////////////////////////////
 //
+// NyquistEffectUIValidator
+//
+///////////////////////////////////////////////////////////////////////////////
+struct NyquistEffect::SharedValidatorState final
+{
+   explicit SharedValidatorState(NyquistEffect& effect)
+       : mEffect(&effect), mParent(mEffect->GetUIParent())
+   {
+      mParent->PushEventHandler(&effect);
+   }
+
+   ~SharedValidatorState()
+   {
+      Reset();
+   }
+
+   void Reset()
+   {
+      if (mEffect == nullptr)
+         return;
+
+      if (mParent)
+         mParent->PopEventHandler();
+
+      mEffect = nullptr;
+   }
+
+   NyquistEffect* mEffect;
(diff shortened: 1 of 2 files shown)


src/effects/nyquist/Nyquist.cpp
```diff
@@ -117,6 +117,68 @@ enum
...
+//
+// NyquistEffectUIValidator
+//
+///////////////////////////////////////////////////////////////////////////////
+struct NyquistEffect::SharedValidatorState final
+{
+   explicit SharedValidatorState(NyquistEffect& effect)
+       : mEffect(&effect), mParent(mEffect->GetUIParent())
+   {
+      mParent->PushEventHandler(&effect);
```
Paul-Licameli, line 130: Check null parent on construction?

src/effects/nyquist/Nyquist.cpp
```diff
@@ -117,6 +117,68 @@ enum
...
+   }
+
+   ~SharedValidatorState()
+   {
+      Reset();
+   }
+
+   void Reset()
+   {
+      if (mEffect == nullptr)
```
Paul-Licameli, line 140: If the constructor takes a reference to the effect, how can mEffect by null?

src/effects/nyquist/Nyquist.cpp
```diff
@@ -117,6 +117,68 @@ enum
...
+         return;
+
+      if (mParent)
+         mParent->PopEventHandler();
+
+      mEffect = nullptr;
+   }
+
+   NyquistEffect* mEffect;
+   wxWeakRef<wxWindow> mParent;
```
Paul-Licameli, line 150: Make the members const?  Or hold a reference not pointer to effect?

crsib: I specifically use pointer so I can nullify it on Reset.

src/effects/nyquist/Nyquist.cpp
crsib: Because I need to ensure that Reset is only called once.

Paul-Licameli: Oh yes, not just in destruction.

PR #3365, opened 2022-08-12
TrackId really can identify tracks across undo states...

Paul-Licameli:
... It wasn't broken for its other intended purposes, but had not been used for
this purpose before.

Resolves: assists development of non-modality of effect dialogs

Recommended:


Paul-Licameli:
As mentioned here, this PR may not be needed after all, to solve problems with non-modality of the effect dialogs.  If you acree, @crsib, close this PR.

https://github.com/audacity/audacity/pull/3176#issuecomment-1214169676

Paul-Licameli:
Not necessary now to make non-modal effect UIs work.  But I'll remember this idea which might be useful in future.

PR #3370, opened 2022-08-13
Issue3345 very simply

Paul-Licameli:
Resolves: #3345

Fix crash running Nyquist Prompt with code that has widgets

Recommended:

### DIFF
--- a/libraries/lib-utility/MemoryX.h
+++ b/libraries/lib-utility/MemoryX.h
@@ -170,19 +170,27 @@ using Destroy_ptr = std::unique_ptr<T, Destroyer<T>>;
 
 // Construct this from any copyable function object, such as a lambda
 template <typename F>
-struct Final_action {
-   Final_action(F f) : clean( f ) {}
-   ~Final_action() { clean(); }
+struct Finally {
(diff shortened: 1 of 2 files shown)


crsib:
There difference between two fixes that my approach only affects Nyquist, your approach has an unknown set of potential side effects from forcing run loop iteration. I can confirm that right now it works on Windows, I haven't check that it works on Linux. Will it break later? I don't know.

I will approve this PR. Please choose between the two considering the consequences from  yielding.

crsib (approved):

Paul-Licameli:
> There difference between two fixes that my approach only affects Nyquist, your approach has an unknown set of potential side effects from forcing run loop iteration. I can confirm that right now it works on Windows, I haven't check that it works on Linux. Will it break later? I don't know.
> 
> I will approve this PR. Please choose between the two considering the consequences from yielding.

I don't fear bad consequences of yielding, which still happens only in very unusual usage of Nyquist, and much prefer the simplicity of this solution as I hope you do too.

crsib:
For the record - I don't agree with you, but as I said - it is your call

PR #3373, opened 2022-08-15
Wavpack enhancements

dbry:
I found a few straggling issues with the WavPack import/export, and I also added a couple enhancements.

These changes should all be simple and straightforward except one. I noticed during my testing that once I had enabled the "correction" files, I was unable to turn them back off in the dialog. This was not obvious at first because the setting _appeared_ to be off (the checkbox was unchecked), but the correction file was still generated. This was persistent and the only way to clear the setting for real was to shutdown and restart Audacity.

I never could figure out why this was happening. Either the `TieCheckBox` functionality is broken (unlikely) or it was simply not being used correctly in this case. I looked at other instances and couldn't find another example of `TieCheckBox` being used the exact same way, so I decided to just convert it to a toggle event like the "hybrid" checkbox (which was working correctly). This fixed the issue and I have not seen any regression after extensive testing.

Thanks again for your consideration and including WavPack in Audacity! 

Recommended:

### DIFF
--- a/src/export/ExportWavPack.cpp
+++ b/src/export/ExportWavPack.cpp
@@ -39,6 +39,7 @@
 //---------------------------------------------------------------------------
 
 #define ID_HYBRID_MODE 9000
+#define ID_CREATE_WVC  9001
 
 class ExportWavPackOptions final : public wxPanelWrapper
 {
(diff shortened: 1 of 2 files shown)


crsib (approved):
That is fantastic, thank you, David!

PR #3374, opened 2022-08-15
Channel counts

Paul-Licameli:
Contributes to VST3 statelessness

Recommended:

### DIFF
--- a/src/effects/Reverb.cpp
+++ b/src/effects/Reverb.cpp
@@ -237,6 +237,16 @@ struct EffectReverb::Instance
       return InstanceProcess(settings, mSlaves[group], inbuf, outbuf, numSamples);
    }
 
+   unsigned GetAudioOutCount() const override
+   {
+      return 2;
+   }
+
+   unsigned GetAudioInCount() const override
+   {
+      return 2;
+   }
+
    bool InstanceInit(EffectSettings& settings, double sampleRate,
       EffectReverbState& data, ChannelNames chanMap, bool forceStereo);
 
@@ -287,16 +297,6 @@ ManualPageID EffectReverb::ManualPage() const
 EffectType EffectReverb::GetType() const
 {
    return EffectTypeProcess;
-}
-
-unsigned EffectReverb::GetAudioInCount() const
-{
-   return 2;
-}
-
-unsigned EffectReverb::GetAudioOutCount() const
-{
-   return 2;
 }
 
 auto EffectReverb::RealtimeSupport() const -> RealtimeSince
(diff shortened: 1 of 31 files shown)


src/effects/Reverb.cpp
```diff
@@ -317,8 +318,9 @@ bool EffectReverb::Instance::ProcessInitialize(EffectSettings& settings,
 bool EffectReverb::Instance::InstanceInit(EffectSettings& settings,
    double sampleRate, EffectReverbState& state,
    ChannelNames chanMap, bool forceStereo)
-{   
-   auto& rs = GetSettings(settings);   
+{
+   auto& rs = GetSettings(settings);
+   mWidth = rs.mStereoWidth ? 2 : 1;
```
pietro68, line 324: I am not 100% sure but this looks incorrect to me: `mStereoWidth` is a continuous controller in the range 0-100, to enable gradual application of stereo separation. 

What happens here is that as soon as the user wants at least 1% of stereo width, then we tell the calling code to treat this effect as a 2-in, 2-out effect.

I would rather assign `mWidth` in the if structure following the original assignment:

```
bool isStereo = false;
   state.mNumChans = 1;
   if (    (chanMap && chanMap[0] != ChannelNameEOL && chanMap[1] == ChannelNameFrontRight)
        || forceStereo )
   {
      isStereo = true;
      state.mNumChans = 2;
      **mWidth = 2;**
   }
  else
  {
    **mWidth = 1;**
  }

```
And actually I would rename `mWidth` to `mNumChannels` or similar

Paul-Licameli: Besides this point -- do you understand the rest and have no objections?  If so, then I want to merge this branch, maybe minus one commit, to unblock other work.

Paul-Licameli: Please observe:  it is only when initializing, either for destructive or realtime, that this makes a difference.  And, we will not make Reverb realtime in 3.2.  But at least for destructive processing, this is restoring logic that was in 3.1.3 and earlier but got lost in this release, and I'm not sure it was correct to lose that.

See the earlier commit that I link:  this is just restoring logic as it was.

pietro68: I see, so: besides this point, which is actually restoring logic as it was, I have no objections so I will approve the PR.

Maybe you could just add a comment for later, writing that this pre-existing logic is probably wrong, especially citing the case when the instance starts with 0% width and then the user might want to increase it (in realtime processing), but then the effect will remain in mono.

pietro68 (approved):

Paul-Licameli:
I made some changes in response to @pietro68 's suggestions.  Please review again

pietro68 (approved):
Changes are ok for me; regarding the existing width logic, we'll need to fix it sometime later.

vsverchinsky (approved):

PR #3375, opened 2022-08-15
Attempt to simplify EffectSettings inter-thread communication

vsverchinsky:
During inter-thread communication we use special function CopyEffectConents which is supposed to perform non-allocating data copying, though not all data that is to be stored in EffectsSettings can be copied that way. Copying such EffectSettings back to main thread may overwrite.
mLastSettings removed as they seem to be redundant

Recommended:


Paul-Licameli:
@vsverchinsky can this PR be closed now?

PR #3384, opened 2022-08-17
Inter thread communication change

Paul-Licameli:
Contributes to VST3 statelessness

The receipt of the "echo" of settings changes by the main thread is generalized with a virtual function of EffectInstance,
which may be overridden to avoid copying everything.

Recommended:


libraries/lib-components/EffectInterface.h
```diff
@@ -481,6 +474,13 @@ class COMPONENTS_API EffectInstance
    //! Function that has not yet found a use
    //! Correct definitions of it will likely depend on settings and state
    virtual size_t GetTailSize() const;
+
+   //! Main thread receives updates to settings from a processing thread
+   /*!
+    Default implementation simply assigns by copy, not move
+    This might be overridden to copy contents only selectively
+    */
```
crsib, line 482: Why do you require R-vaue here in case you expect copy assignment?

Paul-Licameli: Allowing more freedom to overriding methods if it is needed

crsib: You force very harsh contract on the users of this code: they have to provide R-value to this function.

Paul-Licameli:
These changes will be incorporated in #3396

PR #3388, opened 2022-08-17
Fix free store allocation that was hiding in EffectSettingsAccess...

Paul-Licameli:
... This is the simplest fix to write for now.

Note that the get and set of duration format happen only in the main thread, so
no harm in simply sharing the state.

In the next version we should rewrite the inter-thread communication more
thoroughly and separate concerns better, so that the structure holding
persistent settings may be different from those carrying messages between
threads.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -73,9 +73,9 @@ class COMPONENTS_API EffectSettingsExtra final {
 public:
    static const RegistryPath &DurationKey();
    const NumericFormatSymbol& GetDurationFormat() const
-      { return mDurationFormat; }
+      { return *mpDurationFormat; }
    void SetDurationFormat(const NumericFormatSymbol &durationFormat)
-      { mDurationFormat = durationFormat; }
+      { *mpDurationFormat = durationFormat; }
 
    //! @return value is not negative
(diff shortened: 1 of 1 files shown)


crsib (commented):
I believe that now SetDurationFormat is called only during the destructive processing. So now we are going from 0 allocations to 1 allocation.

Are there plans to change it? Can we do lazy initialization at least?

Paul-Licameli:
> I believe that SetDurationFormat is called only during the destructive processing. So now we are going from 0 allocations to 1 allocation.
> 
> Are there plans to change it? Can we do lazy initialization at least?

What I’m fixing is the copying of EffectSettingsExtra causing copy construction of a string and a TranslatableString in the Numeric format symbol.

crsib:
Given that string is empty it is zero cost. Proposed solution, on the other hands, always allocates

Paul-Licameli:
The EffectSettings that is modified by SetDurationFormat in destructive processing still outlasts that, stored in EffectSettingsManager.  So you must ask, does it ever get copied afterward from there into a RealtimeEffectState? .... Actually no, because GetDefaultSettings is not used to initialize EffectSettings in RealtimeEffectStates.

Hm, so little point in this change then.

Issue #3364, opened 2022-08-12
Run Benchmark tool fails


Paul-Licameli:
- Tools > Run Benchmark (this is a little known regression test built into the program itself)
- Press Run
- Observe failure message

Problem bisects to commit

33318a421544fcadbf753271500d3bb5b238ba99

petersampsonaudacity:
Confirmed on W11 with latest alpha master  -  audacity-win-3.2.0-alpha-20220812+fd9bd23-x64-msvc2022

This is a regression on 3.1.3

vsverchinsky:
Turned out there are few more issues introduced by that commit:
1) Incorrect selection region after cutline expansion
2) Fixes insertion into the clip
3) Fixes issue with replacing part of wave track with generated data (and special cases with cutlines and preserving split lines)

PR #3390, opened 2022-08-17
Issue 3364: clip fixes

vsverchinsky:
Resolves: #3364

Completes changes started in #2818

Recommended:

### DIFF
--- a/src/Benchmark.cpp
+++ b/src/Benchmark.cpp
@@ -428,11 +428,11 @@ void BenchmarkDialog::OnRun( wxCommandEvent & WXUNUSED(event))
    // as we're about to do).
    t->GetEndTime();
 
-   if (t->GetClipByIndex(0)->GetSequence()->GetNumSamples() != nChunks * chunkSize) {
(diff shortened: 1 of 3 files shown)


Paul-Licameli:
Approved conditional on the re-running of jobs completing.

I would modify the first commit comment to mention the previous commit id 33318a421544fcadbf753271500d3bb5b238ba99, not just the issue number.

I see some other incidental things involving wave clips are fixed here, beyond the minimum to fix the benchmark.

I see that `GetSequence()->GetNumSamples()` did not mean what it used to mean, so the benchmark test is changed.

But that raises a question:  did other uses of GetNumSamples() become incorrect?  Should they be reviewed to for some errors that have not been noticed?

`GetWaveDisplay` and `WaveClipWaveformCache::GetWaveDisplay` are two places where that function is used outside of the internals of WaveClip implementation.

Paul-Licameli (approved):

vsverchinsky:
> But that raises a question: did other uses of GetNumSamples() become incorrect? Should they be reviewed to for some errors that have not been noticed?

I've checked the use of `WaveClip::GetSequence()`, `Sequence::GetNumSamples` and they seem to be fine.

Issue #3389, opened 2022-08-17
The effect can still be heard on the track even after undo-ing it


dozzzzer:
### Bug description

After applying an effect destructively and canceling it via "Edit > Undo" menu, it still can be heard as if no Undo action was made.

### Steps to reproduce

1. load some audio
2. apply an effect (e.g. Reverb)
3. Undo via "Edit > undo"
4. play the audio again

### Expected behavior

The original sound (before application of the effect) is played

### Actual behavior

The sound has the effect applied to it as if it was never undone:

https://user-images.githubusercontent.com/70063970/185210309-17d50c58-80e3-4da4-9918-e2a866d8b241.mp4

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

Discovered while testing #3315 

I believe the regression was introduced by the PR #3349

crsib:
STR can be simpler
* Generate chirp
* Open any RT capable effect from the Effects menu
* Close it
* Observe

PR #3391, opened 2022-08-17
Fixes RT effects processing on destructive path

crsib:
Strong reference to the created temporary instance is stored in order to fulfil preconditions required by EffectUIHost::CleanupRealtime()

Resolves: #3389

Recommended:

### DIFF
--- a/src/effects/EffectUI.cpp
+++ b/src/effects/EffectUI.cpp
@@ -1242,7 +1242,7 @@ std::shared_ptr<EffectInstance> EffectUIHost::InitializeInstance()
 
    if (mSupportsRealtime && !mInitialized) {
       if (!priorState)
-         mpState =
+         mwState = mpState = mpTempProjectState =
             AudioIO::Get()->AddState(mProject, nullptr, GetID(mEffectUIHost));
       if (mpState) {
          // Find the right instance to connect to the dialog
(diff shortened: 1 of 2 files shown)


Paul-Licameli:
Approved conditionally on test completion.

So I understand there was a "leak" of states in the old global one-deep effect stack, which this fixes.

A few suggestions for code cleanup.

src/effects/EffectUI.cpp
```diff
@@ -1284,14 +1284,12 @@ std::shared_ptr<EffectInstance> EffectUIHost::InitializeInstance()
...
-   if (mSupportsRealtime && mInitialized) {
-      auto mpState = mwState.lock();
 
-      if (noPriorState && mpState) {
-         AudioIO::Get()->RemoveState(mProject, nullptr, mpState);
-         mpState.reset();
+   if (mSupportsRealtime && mInitialized) {
+      if (mpTempProjectState) {
+         AudioIO::Get()->RemoveState(mProject, nullptr, mpTempProjectState);
+         mpTempProjectState.reset();
```
Paul-Licameli, line 1292: This makes the member variable `mwState` not really used so it could be removed from the class.

I think only the inner `if` condition is enough and `mInitialized` can be removed from the class.

crsib: Let's do it separately sometime soon please

crsib:
@Paul-Licameli run succeeded

src/effects/EffectUI.cpp
Paul-Licameli: ok

Paul-Licameli (approved):

Issue #3222, opened 2022-07-12
"Closing" an LV2 effect interface causes crash


Ryan-Miller-Audacity:
**Describe the bug**
1. Attempting to close an LV2 plugin via the "Cancel"/"Close" button for realtime/destructive effects result in a crash. 

2. Attempting to confirm an LV2 plugin via the "OK" button for realtime effects results in a crash. (This crash does not occur when clicking apply on a destructive effect panel).

The two items mentioned above create similar (if not the same) sentry crash reports which can be viewed here: https://sentry.audacityteam.org/share/issue/dfefb4f328b24978b4e69cccfb3cce3f/

Note that this issue is different than #3211 in that these include the interface properly displaying within the plugin window and appear to produce different error reports as well. 

**To Reproduce**
Steps to reproduce the behavior from Item 1:
1. Install an LV2 effect like http://x42-plugins.com/x42/x42-limiter
2. Load/generate some audio
3. Select the audio and open the LV2 plugin via the realtime panel or the effects dropdown
4. Press the gear icon on the realtime effects panel to open the interface (the dropdown will open it on its own)
5. Click the cancel/close button
6. Audacity Crashes

Steps to reproduce the behavior from Item 2: 
1. Install an LV2 effect like http://x42-plugins.com/x42/x42-limiter
2. Load/generate some audio
3. Select the audio and open the LV2 plugin via the realtime panel or the effects dropdown
4. Press the gear icon on the realtime effects panel to open the interface 
5. Click the "OK" button (changing plugin sliders itsn't necessary)
6. Audacity Crashes

**Actual behavior**
1. Audacity crashes after clicking the cancel/close button on the LV2 effect interface.
2. Audacity crashes after clicking the "OK" button on an LV2 effect loaded from the sidebar.

**Expected behavior**
1. The effect window closes without crashing.
2. The effect window closes without crashing.

**Additional information (please complete the following information):**
 - OS: MacOS Monterey 12.3.1 and Windows 10
 - Version: latest alpha 3.2.0

crsib:
I think #3220 fixes this as well

crsib:
I think I can no longer reproduce this issue

Ryan-Miller-Audacity:
[...]

PR #3393, opened 2022-08-18
Fix-crash-on-closing-LV2

Paul-Licameli:
Fix issue 3222 for macOS ...
    
 ... delay the destruction of an obj-C object

Resolves: #3222

Recommended:


crsib:
It still doesn't build. I hope I will have enough time to try updating LV2 SDK past alpha 1

Paul-Licameli:
Let's try that once more...

Paul-Licameli:
And once more, and rebased onto recent master

crsib (approved):

PR #3396, opened 2022-08-18
Issue 2606: vst3 statelessness

vsverchinsky:
Resolves: #2606
Resolves: #3313 
Resolves: #3306
Resolves: #3307
Resolves: #3308

Recommended:


src/effects/VST3/VST3Utils.cpp
```diff
@@ -81,10 +81,8 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
 
 Steinberg::IPtr<PresetsBufferStream> PresetsBufferStream::fromString(const wxString& str)
 {
-   Steinberg::Buffer buffer(str.size() / 4 * 3);
-   auto len = Base64::Decode(str, buffer);
-   assert(len <= buffer.getSize());
-   buffer.setSize(len);
+   auto data = str.To8BitData();
+   Steinberg::Buffer buffer(data, data.length());
```
Paul-Licameli, line 85: This commit is not strictly needed for statelessness.

You must have found a case where this appeared to fix persistency of settings in the configuration file.

The analogous code for VST2, which you originally adapted to VST3, still uses Base64 encoding and to change that would have some small risk of breaking compatibility of someone's settings configuration or macros if they migrate to 3.2.

What was the original reason Base64 encoding was used at all?  I'm not certain but I think it was a way of converting binary data into something that could be stored in a config file with proper escapes and no loss of data.  Is that right, and is it a reason why you still might want to figure out some other way to keep Base64 encoding working?

src/effects/VST3/VST3Effect.cpp
```diff
@@ -38,27 +38,21 @@
 
 #include "ShuttleGui.h"
 
-#include "memorystream.h"
 #include "VST3Utils.h"
 #include "VST3ParametersWindow.h"
 #include "VST3OptionsDialog.h"
+#include "VST3Wrapper.h"*
```
Paul-Licameli: Extra character in `#include`

src/effects/VST3/VST3Utils.cpp
```diff
@@ -94,86 +94,3 @@ wxString PresetsBufferStream::toString() const
 {
    return wxString::From8BitData(mBuffer.str8(), mBuffer.getFillSize());
 }
-
-
-
-bool VST3Wrapper::FetchSettings(VST3EffectSettings& vst3Settings) const
```
Paul-Licameli, line 102: The meanings of `FetchSettings` and `StoreSettings` got reversed, and are now inconsistent with terminology in AudioUnits.  There, `FetchSettings` is fetching state *from* an AudioUnit handle *into* an EffectSettings package of data.  `StoreSettings` does the opposite.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+Steinberg::int32 VST3Wrapper::GetLatencySamples() const
+{
+   return mAudioProcessor->getLatencySamples();
+}
+
+EffectSettings VST3Wrapper::MakeSettings()
+{
+   auto settings = EffectSettings::Make<VST3EffectSettings>();
+   auto vst3setting = GetSettings(settings);
```
Paul-Licameli, line 582: unused local variable

src/effects/VST3/VST3Effect.cpp
```diff
@@ -179,122 +159,34 @@ wxString GetFactoryPresetsPath(const VST3::Hosting::ClassInfo& effectClassInfo)
...
-   mAudioProcessor = audioProcessor;
-}
-
 EffectFamilySymbol VST3Effect::GetFamilySymbol()
 {
    return XO("VST3");
 }
 
-VST3Effect::VST3Effect(const VST3Effect& other)
-   : VST3Wrapper(other.mModule, other.mEffectClassInfo)
```
Paul-Licameli, line 212: Copy construction was eliminated, so perhaps you may also declare the copy with explicit `= delete`

src/effects/VST3/VST3Effect.cpp
```diff
@@ -967,19 +777,15 @@ void VST3Effect::ExportPresets(const EffectSettings& settings) const
       );
       return;
    }
-
-   if (!StoreSettings(GetSettings(settings)))
-      return;
```
Paul-Licameli, line 991: The elimination of `StoreSettings` leaves the argument `settings` unused in this function, at least so I see at the fifth commit in the sequence.  So this "breaks" that function, but I presume you fix it in later commits I have not yet reviewed.  I understand from other commit comments that you put back a local wrapper object in the function for this purpose.

Strictly then this commit sequence isn't preserving intended behavior all the way through but I won't insist on that now.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -1235,24 +1011,11 @@ void VST3Effect::ReloadUserOptions()
 
 EffectSettings VST3Effect::MakeSettings() const
 {
-   auto result = StatefulPerTrackEffect::MakeSettings();
-   // Cause initial population of the map stored in the stateful effect
-   if (!mInitialFetchDone) {
-      FetchSettings(GetSettings(result));
-      mInitialFetchDone = true;
-   }
-   return result;
```
Paul-Licameli, line 1265: Last things to say about the fifth commit.
When I transformed other effect families, the move to actual statelessness came late.  Until a late commit, `GetSettings` was a non-static member function, ignoring its argument, and really using information in the stateful Effect object.

All the places where `GetSettings` had to be called were figured out first; and only then did I make a change so that the `EffectSettings` objects really carried independent information around and were not dummies.

You make this change earlier, so `EffectSettings` objects really carry data, making it possible that buggy behavior exists in the intermedate steps of this transformation because of inconsistencies about where the correct data really are.  I already mentioned the possible wrong behavior of exporting.

It's not how I would write a very careful transoformation.  In case bugs still exist after everything and I want to bisect to understand where the bug was introduced, the other caution I took with intermediate steps might help me find out where the mistake was really introduced.

I'm not insisting now on rewriting this pull request that way.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -38,27 +38,21 @@
...
 
 #ifdef __WXMSW__
 #include <shlobj.h>
 #elif __WXGTK__
 #include "internal/x11/SocketWindow.h"
 #endif
 
-#include "AudacityVst3HostApplication.h"
 #include "ConfigInterface.h"
-#include "VST3ParametersWindow.h"
```
Paul-Licameli, line 54: I think it is the fifth commit that introduces the build failure on Linux.  It seems "Always" is a macro defined somewhere in one of the inclusions.

I was able to fix Linux builds of my cherry-picks of your branch by moving the definition of `VST3Effect::RealtimeSupport` earlier in the file, before other `#include` directives.  Another thing I didn't try was just to insert `#undef Always` before the function.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -176,17 +176,11 @@ VST3Effect::VST3Effect(
...
-   mWrapper->mSetup.maxSamplesPerBlock = mUserBlockSize;
-   mWrapper->mSetup.sampleRate = 44100.0;
-
-   if(!SetupProcessing(*mWrapper->mEffectComponent, mWrapper->mSetup))
-      //ProcessInitiaize should attempt to change that...
-      mWrapper->mSetup.sampleRate = .0;
+   mWrapper->mSetup.processMode = Steinberg::Vst::kOffline;
+   mWrapper->mSetup.symbolicSampleSize = Steinberg::Vst::kSample32;
+   mWrapper->mSetup.maxSamplesPerBlock = mProcessingBlockSize;
+   mWrapper->mSetup.sampleRate = mProjectRate;
```
Paul-Licameli: Just use literal `44100.0` as a default, as before.  Don't use `EffectBase::mProjectRate`.  That is another piece of state in the effect objects I need to eliminate in a future release.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -109,8 +111,120 @@ bool SetupProcessing(Steinberg::Vst::IComponent& component, Steinberg::Vst::Proc
    return false;
 }
 
+class InputParameterChanges final : public Steinberg::Vst::IParameterChanges
+{
+   const Steinberg::int32 mParameterCount;
+   const SingleInputParameterValue* mParameterQueues;
```
Paul-Licameli, line 117: Could this instead be `SingleInputParameterValue *const mParameterQueues;` so you can remove the `const_cast` below?

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -302,6 +419,131 @@ void VST3Wrapper::Finalize()
...
+{
+   const auto& vst3settings = GetSettings(settings);
+   for(auto& p : vst3settings.parameterChanges)
+   {
+      auto it = std::find_if(mParameters.begin(), mParameters.end(), [&p](const auto& v) { return v.first == p.first; });
+      if(it != mParameters.end())
+         it->second = p.second;
+      else
+         mParameters.push_back(p);
+   }
```
Paul-Licameli, line 437: Should this function clear the map in the settings? -- maybe not if it needs to be done in a worker thread.

You don't (yet) use it in any worker thread (at the eighth commit of ten).

Paul-Licameli: At this commit I see accumulation of changes in the map in `performEdit` but clearing nowhere.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -447,6 +347,7 @@ bool VST3Effect::RealtimeAddProcessor(
       auto wrapper = std::make_unique<VST3Wrapper>(*mModule, mEffectClassInfo.ID());
       if(!wrapper->Initialize(settings, sampleRate, Vst::kRealtime, mProcessingBlockSize))
          throw std::runtime_error { "VST3 realtime initialization failed" };
+      wrapper->ConsumeChanges(settings);
```
Paul-Licameli: This is the only place where the processors get changed settings information, and it only happens before playback.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -487,12 +387,8 @@ bool VST3Effect::RealtimeResume()
    return true;
 }
 
-bool VST3Effect::RealtimeProcessStart(EffectSettings &)
+bool VST3Effect::RealtimeProcessStart(EffectSettings& settings)
 {
-   assert(mPendingChanges == nullptr);
-   
-   //Same parameter changes are used among all of the realtime processors
-   mPendingChanges = mWrapper->mComponentHandler->getPendingChanges();
```
Paul-Licameli: Something is needed to replace the deleted lines, of course, so that there can be continuing adjustment of settings during play.  So it seems you would need more `ConsumeSettings` but in a worker thread.

I don't yet see the complete transformation of the code in later commits, but at this commit it is surely incomplete.

Paul-Licameli: Note the `RealtimeProcessStart` is the only place where the worker thread should check for changes of state communicated by the main thread.  Not in `RealtimeProcess` or `RealtimeProcessEnd`.

Paul-Licameli: Maybe the `EffectSettings` could contain a map from parameter ids to `optional<>` values, so the consuming thread could "clear" the map by reassigning `nullopt` everywhere, but not doing any free-store deallocation.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -630,3 +630,26 @@ void VST3Wrapper::SaveUserPreset(const EffectDefinitionInterface& effect, const
...
+      //on the other side on swap.
+      GetSettings(dst).parameterChanges.clear();
+      //Don't allocate in worker
+      std::swap(
+         GetSettings(*const_cast<EffectSettings*>(&src)).parameterChanges,
+         GetSettings(dst).parameterChanges
+      );
+   }
+   else
+      GetSettings(dst).parameterChanges.clear();
```
Paul-Licameli, line 659: Moving on the the ninth commit.  As I already mentioned, you should avoid either allocating or freeing and that means, not clearing a `std::map` either.  So I suggest again, a map from keys to optional values and just reassignment of null opt instead of clearing.

Paul-Licameli: `CopySettingsContents` in either direction is always done on the worker thread.

src/effects/VST3/VST3Instance.h
```diff
@@ -0,0 +1,74 @@
+#pragma once
+
+#include <public.sdk/source/vst/utility/uid.h>
+
+#include "effects/StatefulPerTrackEffect.h"
```
Paul-Licameli, line 5: `PerTrackEffect.h` is enough, don't need its subclass

src/effects/VST3/VST3Effect.h
```diff
@@ -47,19 +51,12 @@ class VST3Effect final : public StatefulPerTrackEffect
    //Used if graphical plugin interface is disabled in the settings, or not provided by the plugin
    VST3ParametersWindow* mPlainUI { nullptr };
 
-   std::vector<std::unique_ptr<VST3Wrapper>> mRealtimeGroupProcessors;
```
Paul-Licameli: This commit loses the vector of processors in the effect, but does not restore that array elsewhere in the instance.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -242,170 +237,6 @@ bool VST3Effect::LoadFactoryPreset(int id, EffectSettings& settings) const
...
-{
-   return mProcessingBlockSize;
-}
-
-sampleCount VST3Effect::GetLatency() const
-{
-   if(mUseLatency)
-   {
-      if(!mRealtimeGroupProcessors.empty())
-         return mRealtimeGroupProcessors[0]->GetLatencySamples();
```
Paul-Licameli: This deleted line has no corresponding one in the new function for the Instance.  But I see from other functions that you assume latency is available at initialization time and does not change, so you cache the value.

Ok.

But I wonder if varying latency, with varying settings, really is a possible thing we will need to handle later.  You know that fixing the handling of latency in realtime play is a problem I need to solve, maybe in a patch release.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -242,170 +237,6 @@ bool VST3Effect::LoadFactoryPreset(int id, EffectSettings& settings) const
...
-   }
-   return { 0u };
-}
-
-bool VST3Effect::ProcessInitialize(
-   EffectSettings &settings, double sampleRate, ChannelNames)
-{
-   if(mWrapper->Initialize(settings, sampleRate, Steinberg::Vst::kOffline, mProcessingBlockSize))
-   {
-      mWrapper->ConsumeChanges(settings);
```
Paul-Licameli: The new `ProcessInitialize` no longer calls `ConsumeChanges`.  Does destructive processing correctly apply the changes to settings made in the dialog?  How is that communicated?  I must see the tenth commit I suppose.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -242,170 +237,6 @@ bool VST3Effect::LoadFactoryPreset(int id, EffectSettings& settings) const
...
-   EffectSettings &settings, unsigned numChannels, float sampleRate)
-{
-   using namespace Steinberg;
-
-   try
-   {
-      auto wrapper = std::make_unique<VST3Wrapper>(*mModule, mEffectClassInfo.ID());
-      if(!wrapper->Initialize(settings, sampleRate, Vst::kRealtime, mProcessingBlockSize))
-         throw std::runtime_error { "VST3 realtime initialization failed" };
-      wrapper->ConsumeChanges(settings);
```
Paul-Licameli: This logic moves to `RealtimeInitialize` while the new `RealtimeAddProcessor` does nothing.  The `ConsumeChanges` is lost so I am again unsure that the processor really begins in the correct state.

src/effects/VST3/VST3Instance.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+bool VST3Instance::RealtimeProcessEnd(EffectSettings& settings) noexcept
+{
+   return true;
+}
+
+bool VST3Instance::RealtimeProcessStart(EffectSettings& settings)
+{
+   mWrapper->ConsumeChanges(settings);
```
Paul-Licameli, line 90: This was the added `ConsumeChanges` call I was expecting to see -- good, but again, I don't understand why it is right to remove the other calls, and it also needs rewrite to avoid deallocations.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -562,8 +397,11 @@ void VST3Effect::ExportPresets(const EffectSettings& settings) const
       );
       return;
    }
-   
-   if (!mWrapper->SavePreset(fileStream))
+
+   auto wrapper = std::make_unique<VST3Wrapper>(*mModule, mEffectClassInfo.ID());
+   wrapper->FetchSettings(settings);
```
Paul-Licameli, line 320: Ninth commit re-introduces use of the `settings` argument, good

src/effects/VST3/VST3Effect.h
```diff
@@ -47,19 +51,12 @@ class VST3Effect final : public StatefulPerTrackEffect
...
+   std::shared_ptr<VST3Instance> mCurrentDisplayInstance;
```
Paul-Licameli: I expect that this new member of the Effect moves to the Validator next

src/effects/VST3/VST3Instance.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+void VST3Instance::ReloadUserOptions()
+{
+   // Reinitialize configuration settings
+   int userBlockSize;
+   GetConfig(mProcessor, PluginSettings::Shared, wxT("Options"),
+      wxT("BufferSize"), userBlockSize, 8192);
+   mUserBlockSize = std::max( 1, userBlockSize );
+   GetConfig(mProcessor, PluginSettings::Shared, wxT("Options"),
+      wxT("UseLatency"), mUseLatency, true);
```
Paul-Licameli, line 185: There is a long-distance coincidence of repeated string literals here and in `VST3OptionsDialog.cpp`.  It would be nice to clean that up, but it's not necessary now

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -312,6 +312,11 @@ VST3Wrapper::~VST3Wrapper()
       mEffectComponent->terminate();
 }
 
+bool VST3Wrapper::IsActive() const noexcept
+{
+   return mActive;
```
Paul-Licameli, line 317: I saw this new function and thought, is it safe to use only on the main thread?  Do changes of `mActive` happen only in the main thread?  Is `mActive` privately used only on the main thread?

Yes, yes, and yes, which is good.  So please just add a comment in the header file that it's for main thread usage only.

src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -0,0 +1,205 @@
...
+   mPlainUI = nullptr;
+   mParent = nullptr;
+   if(mPlugView)
+   {
+      mPlugView->setFrame(nullptr);
+      mPlugView->removed();
+      mPlugView = nullptr;
+      mPlugFrame = nullptr;
+   }
+}
```
Paul-Licameli, line 165: I think this should call through to `EffectUIValidator::OnClose` so the effect doesn't hold on to `mParent` either

src/effects/VST3/VST3Effect.cpp
```diff
@@ -253,110 +252,29 @@ int VST3Effect::ShowClientInterface(wxWindow& parent, wxDialog& dialog,
...
-   {
-      mPlugView->setFrame(nullptr);
-      mPlugView->removed();
-      mPlugView = nullptr;
-      mPlugFrame = nullptr;
-   }
-
-   if(mCurrentDisplayInstance)
-   {
-      mCurrentDisplayInstance->GetWrapper().EndParameterEdit();
```
Paul-Licameli: This call got lost in the Validator.  The function is no longer called anywhere.

src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+bool VST3UIValidator::UpdateUI()
+{
+   if(!mWrapper.IsActive())
+      mWrapper.FetchSettings(mAccess.Get());
```
Paul-Licameli, line 170: My understanding of these lines is that they update both the edit controller and the effect component after you change a preset.

The first, presumably, causes graphical updates.  The other affects the state of the processor, but this might be happening during processing.

Are we confident of the thread safety of this?

If not -- then `RealtimeProcessStart` in the Instance would have to complete the change of state of the effect component.  Would it be correct just to do a complete state change there, not using the pending changes interface?

Paul-Licameli: This also raises the question of the thread safety of the calls that may precede `UpdateUI` and those are `LoadFactoryDefaults`, `LoadFactoryPreset`, `ImportPresets`, `LoadUserPreset`.  Are they all safe to do while another thread is processing?

src/effects/VST3/VST3Effect.cpp
```diff
@@ -253,110 +252,29 @@ int VST3Effect::ShowClientInterface(wxWindow& parent, wxDialog& dialog,
...
-
-bool VST3Effect::ValidateUI(EffectSettings &settings)
-{
-   if (mDuration != nullptr)
-      settings.extra.SetDuration(mDuration->GetValue());
-
-   assert(mCurrentDisplayInstance);
-
-   mCurrentDisplayInstance->GetWrapper().StoreSettings(settings);
-   mCurrentDisplayInstance->GetWrapper().FlushSettings(settings);
```
Paul-Licameli: The sequence of those two calls is changed in the Validator now, and I think that is correct

src/effects/VST3/VST3Effect.cpp
```diff
@@ -253,110 +252,29 @@ int VST3Effect::ShowClientInterface(wxWindow& parent, wxDialog& dialog,
...
-   }
-   else if(!LoadVSTUI(*wrapper.mEditController, parent))
-   {
-      mPlainUI = VST3ParametersWindow::Setup(
-         *parent,
-         *wrapper.mEditController,
-         *wrapper.mComponentHandler
-      );
-   }
-   wrapper.BeginParameterEdit(access);
```
Paul-Licameli: This call also got lost in the transformation.

I think this means that `ComponentHandler::perforEdit` will never do anything, so changes of dials will not be communicated in realtime to the processor.

src/effects/VST3/VST3Effect.cpp
vsverchinsky: mProjectRate value is used to initialize generators, like in another effects, but without explicit sample rate specification

Paul-Licameli: Later commits eliminated these lines, so it doesn't matter.

`mSetup` is now, properly, part of Instance state, not the effect, and is initialized with the correct rate.

src/effects/VST3/VST3Utils.cpp
vsverchinsky: >The analogous code for VST2, which you originally adapted to VST3, still uses Base64 encoding and to change that would have some small risk of breaking compatibility of someone's settings configuration or macros if they migrate to 3.2.

I haven't yet figured out what exactly causes data corruption. May be it's in Base64::Encode or Base64::Decode or may be there are some implicit charset conversion which wasn't taken into account.

>What was the original reason Base64 encoding was used at all? I'm not certain but I think it was a way of converting binary data into something that could be stored in a config file with proper escapes and no loss of data.

Correct

src/effects/VST3/VST3Effect.cpp
vsverchinsky: Using StoreSettings on VST3Effect instance could change it's state, which isn't correct behaviour. Considering that there is a single instance and the copy of settings that passed to this function was more likely fetched from same instance it shouldn't harm?

src/effects/VST3/VST3Effect.cpp
vsverchinsky: Name collision caused by indirect include of X.h, fixed by moving includes to another file

src/effects/VST3/VST3Effect.cpp
vsverchinsky: >Something is needed to replace the deleted lines, of course, so that there can be continuing adjustment of settings during play.

Not neccessarily. VST3Effect doesn't yet (at this step) support realtime processing and doesn't support parameter changes "on the fly". So at this point it's enough to call ConsumeSettings once, just before processing has started.

>consuming thread could "clear" the map by reassigning nullopt everywhere, but not doing any free-store deallocation.

parameterChanges used to be a vector... but as I said in the previous comment, I'll move it to AssignSettings.

src/effects/VST3/VST3Effect.cpp
vsverchinsky: In case of VST3 plugin should explicitly ask host to restart (via setActive) effect component  by sending IComponentHanler::restartComponent(kLatencyChanged).

src/effects/VST3/VST3Instance.cpp
vsverchinsky: 1. There is only one instance of VST3Wrapper which owned by VST3Instance (no per-track sub effect initialization)
2. There is only one point where actual parameter change transfer before processing could occur - EffectInstance::RealtimeProcessStart
3. In case of destructive processing instance state updated via IComponent::setState with settings stored by VST3UIValidator::ValidateUI

src/effects/VST3/VST3Effect.cpp
vsverchinsky: And also it creates a helper effect instance

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: >Yes, yes, and yes, which is good. So please just add a comment in the header file that it's for main thread usage only.

The only exception in the interface of this class is a Process function, otherwise it's not thread-safe

src/effects/VST3/VST3Effect.cpp
vsverchinsky: Yep, thanks, fixed that

src/effects/VST3/VST3UIValidator.cpp
vsverchinsky: Thanks

src/effects/VST3/VST3UIValidator.cpp
vsverchinsky: >Are we confident of the thread safety of this?

It's safe to call setState/getState during processing, but this doesn't happen during processing `if(!isActive)`

>Would it be correct just to do a complete state change there, not using the pending changes interface?

I would object to that since `RealtimeProcessStart` is getting called for each block, I guess it's not how it supposed to work. Correct thing would be to use setState once (or when preset has changed), send changes via `IComponentHandler` and then save state at least once (in our case - when UI is closed)

src/effects/VST3/VST3Effect.cpp
Paul-Licameli: The point, however, is to make ExportPresets work independently of any instance.  And you achieve that in the later commit with a temporary Wrapper.

src/effects/VST3/VST3Effect.cpp
Paul-Licameli: I see you added the build fix, but maybe it would be better done earlier in the commit sequence to keep all commits buildable.

src/effects/VST3/internal/x11/SocketWindow.h
```diff
@@ -12,9 +12,8 @@
 
 #pragma once
 
-#include <gdk/gdk.h>
-#include <gtk/gtk.h>
-#include <gdk/gdkx.h>
+#include <glib.h>
```
Paul-Licameli, line 15: The build fixes might be done earlier in the commit sequence to keep all commits buildable.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+void VST3Wrapper::CopySettingsContents(const EffectSettings& src, EffectSettings& dst, SettingsCopyDirection copyDirection)
+{
+   if(copyDirection == SettingsCopyDirection::MainToWorker)
+   {
+      //Don't allocate in worker
+      std::swap(
+         GetSettings(*const_cast<EffectSettings*>(&src)).parameterChanges,
+         GetSettings(dst).parameterChanges
+      );
```
Paul-Licameli, line 651: You have simply deleted the clearing of the `std::map` instead of the more complicated suggstion I made to use a map of unchanging size with `std::optional` values in it.
```suggestion
      );
```
```suggestion
      );
```
```suggestion
      );
```
```suggestion
      );
```

There is also a swapping of map contents.

I wonder whether this will reall work right.  It would be that the old contents swapped away might later be incorrectly swapped in again, if `CopySettingsContents` is called again before the main thread sends an updated message.

So I think the right thing to do, which might fix some bugs with real time adjustment of dials, is as I suggested.  More explicitly:

- `MakeSettings` makes a map from each parameter id to `nullopt`
- `performEdit` emplaces values
- `CopySettingsContents` copies each value and assigns `nullopt` to the source (note, std::move from the optional will NOT make the source null opt)

If you are experiencing some bugs with adjustment of sliders, then try this instead.

src/effects/VST3/internal/x11/SocketWindow.h
vsverchinsky: it doesn't build for you?

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: The more complicated suggestion will not work because:
a) `MakeSettings` can't make such map because to do so it would need to instantiate `IEditController`
b) Assigning `nullopt` or clearing the map turned out to be pointless, considering that the real purpose is to clear setting on the main thread side (so that subsequent `Get` see it empty) and not the intermediate objects, which could be achieved only by using `Flush`

For now I guess it's not a big problem that main thread remebers latest values of all parameters that have changed

> It would be that the old contents swapped away might later be incorrectly swapped in again, if CopySettingsContents is called again before the main thread sends an updated message.

I thought that counter prevents that, isn't it?

Paul-Licameli:
> The new `ProcessInitialize` no longer calls `ConsumeChanges`. Does destructive processing correctly apply the changes to settings made in the dialog? How is that communicated? I must see the tenth commit I suppose.

I did stepping through in the debugger and now I understand that `FlushSettings` and then `StoreSettings` in `ValidateUI` puts complete state into `EffectSettings`, then `ProcessInitialize` causes `FetchSettings` to complete the copy of state correctly into another instance object.

src/effects/VST3/VST3Wrapper.cpp
Paul-Licameli: > > Yes, yes, and yes, which is good. So please just add a comment in the header file that it's for main thread usage only.
> 
> The only exception in the interface of this class is a Process function, otherwise it's not thread-safe

I don't understand that answer.  There is no use of `mActive` but on the main thread.  Correct?

Anyway this unresolved point is only about a comment.

Paul-Licameli:
I just noticed one `TODO` can be easily fixed:  `VST3Effect::LoadPreset` can now be `const` and the ugly `const_cast` at line 232 can be removed.

src/effects/VST3/VST3Effect.cpp
Paul-Licameli: I consider this resolved but I leave it open so you read and understand my previous comment.

Paul-Licameli:
I'm almost approving this commit.  Please see all the comments above that I did not yet mark resolved.

I would like to have some explanation of why Base64 conversion was not working as expected.

Also see the points about UpdateUI.  I did some experiments, exporting a custom preset, then trying to switch between defaults and importing my preset, using the sidebar dialog, during play and while not playing.

This does not yet work properly.

We might decide it's good enough for alpha release without this, but you are right that we must "send changes via IComponentHandler".

Paul-Licameli:
Besides those two important points, there is the suggestion that a map to optionals might be better.

I can demonstrate a problem that I think would be fixed by this.

With Valhalla Supermassive, I use default settings, click "Mix" and then, during play, click and drag that dial, then release.

I can get the dial to fluctuate between two different values, sometimes very different values, and hear the playback switching between them.

Paul-Licameli:
Really then, that is three important points.  There are other recent points I made that are only suggestions.

Paul-Licameli:
Second commit comment has a spelling error in "redundant"
Third commit "Moving some of components into VST3Wrapper" now fails to build!

src/effects/VST3/internal/x11/SocketWindow.h
Paul-Licameli: I didn't really try building on Linux.  I may be mistaken.  All commits were building for me on macOS -- but not after the latest rebasing.

Paul-Licameli:
Another commit comment misspells "actual"

Paul-Licameli:
Your new last commit makes a much simpler fix for the problem of fluctuating settings.  Now I feel silly that I didn't think of it!

Paul-Licameli (approved):
I would like you to make one last interactive rebase that fixes the commit that doesn't build because it refers to a member of `VST3EffectSettings` that doesn't exist.  Also fix commit comments.

Otherwise, I'm convinced you fixed two out of the three major problems I mentioned.

But I think there will be remaining problems with use of factory presets or imported presets.  These can be described in a new issue that we can fix later.

PR #3415, opened 2022-08-23
Allows to set border for InvisiblePanel

crsib:

Recommended:


Paul-Licameli (approved):
Changing arguments of a not-yet-used function, looks harmless

Issue #3309, opened 2022-07-29
define the Validator class for VST2 effects


Paul-Licameli:
Things to do (not necessarily in this sequence), with some analogous older commits

- [x] 0. Define minimal Instance class -- holding references to Instance and EffectSettingsAccess (43d9887c)
- [x] 1. Validator can fetch settings from, store to, Instance (And stores on construction; if needed to initally populate dialog values correctly) (7d1c0c6f)
- [x] 2. Effect's `TransferDataToWindow` becomes Validator's `UpdateUI` override; `ValidateUI` moves into Validator, uses `mAccess.ModifySettings`, fetches settings from Instance
- [x] 3. Plain UI controls, event handlers moved into Validator; all use ModifySettings, maybe just invoking ValidateUI (no analogy with AudioUnits)
  - [ ] EVENT_TABLE macros refer to validator class, not effect **[PM: it was deleted]**
  - [ ] DECLARE_EVENT_TABLE() needed in Validator class but deleted from Effect **[PM: deleted from Effect, but not moved to Validator - no need to do it]**
  - [ ] Pushing and popping of the event handler are not needed when the event table is no more **[PM: actually, Pushing is still needed, please see comment in the code]**
- [x] 4. Native UI event and idle callbacks to host are handled by validator; parameter value changes invoke `mAccess.ModifySettings` (bae543a9)
  - [x] for VST2, this means certain cases of `AudioMaster`

Somewhere in this sequence, the mutex should be eliminated -- instead, `OnSlider` and `Automate` will use `EffectSettingsAccess::ModifySettings`.

The `VSTEffectTimer` will be part of the Validator not the Effect.

It may be that the Validator requires its own Wrapper to dispatch the Automate messages.  Or it may be that the Instance's Wrapper will be sent the messages, but on the main thread, and the Validator should actually handle them.  This isn't clear, and experiment should decide it.

But the Effect's Wrapper should not dispatch any Automate messages.  It should be possible for two dialogs to be open at once for two instances of the same Effect.  Any state information should be duplicated and separated, not unique in the Effect.

PR #3420, opened 2022-08-23
vst2 Validator

pietro68:
Resolves: https://github.com/audacity/audacity/issues/3309

Recommended:


src/effects/VST/VSTEffect.h
```diff
@@ -282,6 +282,24 @@ struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
...
+   ArrayOf<wxStaticText*> mNames;
+   ArrayOf<wxSlider*> mSliders;
+   ArrayOf<wxStaticText*> mDisplays;
+   ArrayOf<wxStaticText*> mLabels;
+   NumericTextCtrl* mDuration;
+   VSTControl* mControl;
+   void NeedEditIdle(bool state);
+      bool mWantsEditIdle{ false };
+      std::unique_ptr<VSTEffectTimer> mTimer;
+   void RefreshParameters(int skip = -1) const;
```
Paul-Licameli, line 301: These should not be in the Wrapper class.  Instead move all from the effect to the validator -- except that mDialog and mParent are duplicated in Effect and Validator.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1299,20 +1299,30 @@ std::unique_ptr<EffectUIValidator> VSTEffect::PopulateUI(ShuttleGui &S,
...
+   auto pParent = S.GetParent();
+   pParent->PushEventHandler(this);
+
+   auto validator = std::make_unique<VSTEffectValidator>(dynamic_cast<VSTEffectInstance&>(instance), *this, access, pParent);
+
+   // These could be moved in the validator constructor
+   validator->Load();
+   validator->mParent = mParent;
+   validator->mTimer = std::make_unique<VSTEffectTimer>(this);
+   validator->mDialog = static_cast<wxDialog*>(wxGetTopLevelParent(parent));
```
Paul-Licameli, line 1311: These three assignments can be done instead in the constructor of the validator

src/effects/VST/VSTEffect.cpp
```diff
@@ -1332,6 +1342,10 @@ bool VSTEffect::ValidateUI(EffectSettings &settings)
 
 bool VSTEffect::CloseUI()
 {
+   // TODO: now that the UI is owned/managed by the validator,
```
Paul-Licameli, line 1345: ValidateUI, CloseUI, OnTimer, NeedIdle, NeedEditIdle, RefreshParameters, OnSizeWindow, OnSlider all move to Validator

TransferDataToWindow becomes UpdateUI of the Validator

src/effects/VST/VSTEffect.h
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
pietro68: Done (actually, by using already passed arg `parent`) - and I also moved there the call to `Load()`.

src/effects/VST/VSTEffect.cpp
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
```diff
@@ -1308,6 +1308,10 @@ std::unique_ptr<EffectUIValidator> VSTEffect::PopulateUI(ShuttleGui &S,
    }
 
    auto pParent = S.GetParent();
+
+   // Looks like pushing the event handler is still needed, otherwise
+   // an assert will pop up when closing the gui:
+   // "xwWindowsBase::PopEventHandler(): cannot pop the wxWindows itself"
```
Paul-Licameli, line 1315: The solution for this:  stop inheriting `DefaultEffectUIValidator`, just inherit `EffectUIValidator`

Which means you no longer have the overrides of `ValidateUI`, `IsGraphicalUI`, `Disconnect` working for you.

But the first is already overridden in your class by the time we reach this commit, `IsGraphicalUI` is trivial to do, and `Disconnect` is unimportant (you could override it if you like, to zero out the parent and dialog pointers)

src/effects/VST/VSTEffect.cpp
```diff
@@ -1758,24 +1758,11 @@ bool VSTEffectWrapper::Load()
...
 
-void VSTEffect::Unload()
-{
-   if (mAEffect)
-   {
-      // Finally, close the plugin
-      callDispatcher(effClose, 0, 0, NULL, 0.0);
-      mAEffect = NULL;
-   }
-}
```
Paul-Licameli, line 1789: I think the correct thing to do in the "Unify Unload" commit is call Unload form the destructor of VSTEffectWrapper and remove the calls in the other destructors.

But note that this causes it to happen for the Validator too, which is also a wrapper.

As this commit stands, a Validator would have a Load at construction time, not paired with an Unload, and that might mean that you are leaking resources.

Paul-Licameli:
I'm looking over the sum of changes in the squashed sequence of commits, and there are few things to addres.

- Remove mParentFE and just use null instead for the parent, wherever VSTEffect member functions need to make a modal dialog.
- That means VSTEffect::SizeWindow should become a virtual function override in VSTEffectValidator instead.  Leave VSTEffect then with a no-op override.

Paul-Licameli:
`FetchSetingsFromInstance` may be inappropriate for VST.  The analogy with AudioUnits may be the wrong one to follow.

I'm looking at `VSTEffectValidator::ValidateUI()` and wondering what's the next step.  Remember how this override is implemented for validators of the few built-in effects that became stateless.

The validator has its own complete copy of settings.  These are simply assigned to the argument of the lambda, after the special treatment of duration.

This may be what you do in the step to statelessness, removing the need for the lock_guard in at least one place.

I think UpdateUI would stop using `StoreSettingsToInstance` while `StoreSettings` would update the Validator's copy of settings.

src/effects/VST/VSTEffect.cpp
pietro68: thanks for the tip, done

src/effects/VST/VSTEffect.cpp
pietro68: Good point! done.

pietro68:
> I'm looking over the sum of changes in the squashed sequence of commits, and there are few things to addres.
> 
>     * Remove mParentFE and just use null instead for the parent, wherever VSTEffect member functions need to make a modal dialog.
> 
>     * That means VSTEffect::SizeWindow should become a virtual function override in VSTEffectValidator instead.  Leave VSTEffect then with a no-op override.

also done now

src/effects/VST/VSTEffect.cpp
```diff
@@ -789,7 +782,7 @@ void VSTEffect::ResourceHandle::reset()
 VSTEffect::VSTEffect(const PluginPath & path)
 :  VSTEffectWrapper(path)
 {
-   mTimer = std::make_unique<VSTEffectTimer>(this);
+   //mTimer = std::make_unique<VSTEffectTimer>(this);
```
Paul-Licameli, line 785: You can delete the comment

pietro68: ok

pietro68:
> I'm looking at `VSTEffectValidator::ValidateUI()` and wondering what's the next step. Remember how this override is implemented for validators of the few built-in effects that became stateless.

I created a new PR based on this one, with one new commit at the end, which does final statelessness - #3541

src/effects/VST/VSTEffect.cpp
```diff
@@ -911,6 +911,139 @@ bool VSTEffect::InitializePlugin()
...
+
+   return true;
+}
+
+
+bool VSTEffectWrapper::CopySettingsContents(const VSTEffectSettings& src,
+                                                  VSTEffectSettings& dst) const
+{
+   return TransferSettingsContents(
+      const_cast<VSTEffectSettings&>(src), dst, /*move =*/ true, /*merge =*/ false);
```
Paul-Licameli, line 982: CopySettingsContents should not move from src!  src is not declared rvalue reference, so a function should not surprisingly mutate it.

(But ultimately CopySettingsContents will not be needed)

src/effects/VST/VSTEffect.cpp
```diff
@@ -911,6 +911,139 @@ bool VSTEffect::InitializePlugin()
...
+      {
+         src.mChunk = std::nullopt;
+      }
+   }
+   else if(!doMerge)
+   {
+      dst.mChunk = std::nullopt;
+   }
+   
+
```
Paul-Licameli, line 943: All the stuff above with chunks and the rest does not belong in the Message object.

VST3EffectSettings needs a base class, or a member, which contains only the map.

TransferSettingsContents should only be concerned with two maps.

src/effects/VST/VSTEffect.cpp
```diff
@@ -911,6 +911,139 @@ bool VSTEffect::InitializePlugin()
...
+      }
+      else if (!doMerge)
+      {
+         // source has no value, and we do not want to merge (i.e. keep any
+         // dest value that might be there) - delete the destination value
+         dstValue = std::nullopt;
+      }
+
+      dstIter++;
+   }
```
Paul-Licameli, line 972: This loop looks correct.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3415,15 +3548,22 @@ void VSTEffectWrapper::ForEachParameter(ParameterVisitor visitor) const
...
-         vstSettings.mParamsMap[pi.mName] = val;
+         if (doFetch)
+         {
+            float val = callGetParameter(pi.mID);
+            vstSettings.mParamsMap[pi.mName] = val;
+         }
+         else
+         {
+            vstSettings.mParamsMap[pi.mName] = std::nullopt;
+         }
```
Paul-Licameli, line 3566: This too looks all right.

However there is no remembering of the mID value from ParameterInfo.

I think the Message may need to have that when the values are ultimately used in RealtimeProcessStart, so you can do callSetParameter.

So maybe it will be easiest to change the data type of the map to something like `std::optional<std::pair<int, double>>`

src/effects/VST/VSTEffect.cpp
pietro68: Whoops! definitively a copy-paste error there. Thanks.

src/effects/VST/VSTEffect.cpp
pietro68: Well, VST3EffectSettings already has a member that (in a way) contains only the map - it's the map itself. I could just remove from this method all lines that reference what is not the map. Would that be fine?

Paul-Licameli: ok, the map itself can be the argument type passed to these functions.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3631,7 +3634,7 @@ bool VSTEffectWrapper::StoreSettings(const VSTEffectSettings& vstSettings) const
          const auto itr = vstSettings.mParamsMap.find(pi.mName);
          if (itr != vstSettings.mParamsMap.end())
          {
-            const float& value = *itr->second;
+            const float& value = itr->second->second;
```
Paul-Licameli, line 3637: The assumption that iter->second is not nullopt needs to be commented now, and the explanation is that FetchSettings passed the default true as the second argument, whenever the EffectSettings are made rather than EffectMessage.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1344,15 +1344,18 @@ bool VSTEffect::LoadSettings(const CommandParameters& parms, EffectSettings& set
...
    {
       do
       {
          if (parms.Read(key, value))
-            vstSettings.mParamsMap[key] = value;
+            vstSettings.mParamsMap[key] = { intKey, value };
          else
             return false;
 
+         intKey++;
```
Paul-Licameli, line 1357: This doesn't look right.

What makes you sure this assignment of integer keys is correct?

Paul-Licameli: Actually it doesn't matter because those integers will never be used when part of EffectSettings, but only when part of a Message.

In fact, maybe the Message should not use a map at all, but instead a vector indexed by small integers?

Yes, in fact, maybe the analogy with AudioUnits is misleading.  Paramer IDs for a VST effect will be some small initial segment of integers of size `mAEffect->numParams` and you only need a vector of optional<float> in the Message.

It didn't work like this for AudioUnits because parameter IDs could be a sparse sequence of larger numbers, so a map made more sense there.

PR #3422, opened 2022-08-24
Allow supplying custom Progress Dialogs to Export

crsib:

Recommended:

### DIFF
--- a/src/export/Export.cpp
+++ b/src/export/Export.cpp
@@ -469,7 +469,8 @@ bool Exporter::Process(bool selectedOnly, double t0, double t1)
    }
 
    // Export the tracks
-   bool success = ExportTracks();
+   std::unique_ptr<BasicUI::ProgressDialog> pDialog;
+   bool success = ExportTracks(pDialog);
 
    // Get rid of mixerspec
    mMixerSpec.reset();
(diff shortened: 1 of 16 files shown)


src/widgets/wxWidgetsBasicUI.cpp
```diff
@@ -155,7 +155,7 @@ wxWidgetsBasicUI::DoMessageBox(
 }
 
 namespace {
-struct MyProgressDialog : ::ProgressDialog, BasicUI::ProgressDialog {
+struct MyProgressDialog : ::ProgressDialog {
```
Paul-Licameli, line 158: The first commit can now simply eliminate this class and use `::ProgressDialog` directly.

src/export/Export.cpp
```diff
@@ -249,19 +249,24 @@ std::unique_ptr<Mixer> ExportPlugin::CreateMixer(const TrackList &tracks,
...
 {
    if (!pDialog)
       pDialog = std::make_unique<ProgressDialog>( title, message );
-   else {
-      pDialog->SetTitle( title );
-      pDialog->SetMessage( message );
-      pDialog->Reinit();
+   else
+   {
+      if (auto pd = dynamic_cast<ProgressDialog*>(pDialog.get()))
```
Paul-Licameli, line 259: Rather than this `dynamic_cast`, the abstract class `BasicUI::ProgressDialog` can have two more virtual functions.  `SetTitle` has an obvious purpose.  `Reinit` should be commented as resetting progress to zero and resetting any cancelled or stopped state.

src/widgets/wxWidgetsBasicUI.cpp
crsib: Yeah, makes sense. Surprisingly, MSVC was building as it was, so I've changed this line very late and haven't though that MyProgressDialog is no longer needed at all

Paul-Licameli (approved):
Take my comments as non-blocking suggestions only.

src/export/Export.cpp
crsib: `SetTitle` comes from wxWidgets. I can call it SetDialogTitle though

Paul-Licameli:
The rebasing looks good to me

PR #3426, opened 2022-08-25
Effect latency preliminaries

Paul-Licameli:
First steps toward latency correction in realtime effects, change no behavior yet

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -174,6 +174,12 @@ void EffectInstance::AssignSettings(EffectSettings &dst, EffectSettings &&src)
    dst = src;
 }
 
+auto EffectInstance::GetLatency(const EffectSettings &, double) const
+   -> SampleCount
+{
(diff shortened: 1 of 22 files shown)


crsib:
While looking into this PR I found a P1 in VST3 in master as it seems :-(
image

Paul-Licameli:
> While looking into this PR I found a P1 in VST3 in master as it seems :-(
> image

You may have discovered and example of a bug I suspected might happen, and I asked @Ryan-Miller-Audacity to find an example with free plug-ins.  And I think, @vsverchinsky , this example is why the implementation of VST3 statelesness can't just skip the problem of "slave" processors.

Paul-Licameli:
@crsib besides this incidental discovery of a bug in VST3, do I have an approval of these preliminaries for the latency problem?

crsib:
Not yet, sorry. I stuck on the first commit around 1 am, found a bug and decided that it was quite enough for one day xD

crsib (approved):

PR #3430, opened 2022-08-25
LabelTrackView: Fix layout bugs

bjj:
Fixes several issues with label track layout that are especially visible
at larger font sizes:

* Text aligned too low inside containing box, with descenders escaping
  the box or even overlapping the label bar.
* Text boxes overlapping the label bar above.
* Empty text boxes drawing at the top edge of the track when the vertical
  size of the track is too small to display any part of the label.
* Icons/glyphs (the handles on the ends of the label bar) not being
  properly centered on the label bar.
* Rounding errors causing visual instability when moving between even/odd
  font sizes.
* Layout not responding properly to adjustments of the constants that drive
  the layout (sizes, padding, etc).

One deliberate visual change:

* LabelBarHeight was even while the icons/glyphs were odd height, making
  it impossible to perfectly center them. Reduced LabelBarHeight by 1.

Recommended:

### DIFF
--- a/src/tracks/labeltrack/ui/LabelTrackView.cpp
+++ b/src/tracks/labeltrack/ui/LabelTrackView.cpp
@@ -425,6 +425,14 @@ void LabelTrackView::ComputeTextPosition(const wxRect & r, int index) const
    labelStruct.xText = xText;
 }
 
+// All things that contribute to one row:
+// The complete text height (including padding), the margin between (-TextFrameYOffset) and
+// the LabelBarHeight (icon is centered on labelbar and does not affect height)
+int LabelTrackView::GetLabelRowHeight()
+{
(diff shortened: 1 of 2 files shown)


bjj:
Added another commit to match the click hitbox to the UI. The old hitbox (at least the vertical component) was not directly related to text position, so there's some argument that it could have been a design choice, but I suspect it was based on a much older layout where text was vertically aligned within the label.

LWinterberg:
Hello! We just [switched master to 4.0.0](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444) which makes this PR incompatible and/or obsolete as we're switching UI framework and reorganizing large parts of the codebase. This includes an entirely new label track!

You're very welcome to make these kinds of contributions in the future though, after things stabilize a little bit.

Issue #3394, opened 2022-08-18
Quick audio sharing


LWinterberg:
Add a quick audio sharing feature [as previously announced](https://forum.audacityteam.org/viewtopic.php?f=68&t=126279).

PR #3432, opened 2022-08-25
Audio sharing support

crsib:
Resolves: #3394 

Implements a quick audio sharing feature [as previously announced](https://forum.audacityteam.org/viewtopic.php?f=68&t=126279).

Recommended:


libraries/lib-cloud-audiocom/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+#[[
+A library, that allows uploading to audio.com sharing service
+]]#
```
Paul-Licameli, line 3: Don’t need second ‘#’

libraries/lib-cloud-audiocom/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+   ServiceConfig.cpp
+   ServiceConfig.h
+   UploadService.cpp
+   UploadService.h
+   UserService.cpp
+   UserService.h
+)
+
+
+set ( LIBRARIES PRIVATE
```
Paul-Licameli, line 22: I prefer PRIVATE only before wxBase, otherwise allowing transitive dependencies

libraries/lib-cloud-audiocom/OAuthService.h
```diff
@@ -0,0 +1,113 @@
...
+public:
+   //! Indicates, that service has a valid access token, i. e. that the user is authorized.
+   bool HasAccessToken() const;
+   //! Indicates, that the service has a possibly valid refresh token, which can be used
+   //! to authorize the user
+   bool HasRefreshToken() const;
+
+   //! Attempt to authorize the user.
+   /*
+      If `HasAccessToken() || !HasRefreshToken()` invokes completedHandler synchronosly,
```
Paul-Licameli, line 49: Synchronously

libraries/lib-cloud-audiocom/OAuthService.h
```diff
@@ -0,0 +1,113 @@
...
+   void ValidateAuth(std::function<void(std::string_view)> completedHandler);
+
+   //! Handle the OAuth callback
+   /*
+      This method supports the following URLs:
+      - audacity://link?authorization_code=code
+      - audacity://link?token=refresh_token
+      - audacity://link?username=user&password=password
+
+      This method will attempt to perform OAuth2 authorization and invoke completedHadler (if valid).
```
Paul-Licameli, line 66: Completedhandler

libraries/lib-cloud-audiocom/CMakeLists.txt
crsib: Yeah, I keep copying it :-(

libraries/lib-cloud-audiocom/ServiceConfig.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+AudioFormat ServiceConfig::GetPreferredAudioFormat() const
+{
+   return AudioFormat::WAVPACK;
+}
+
+const ServiceConfig& GetServiceConfig()
+{
+   static ServiceConfig config;
+   return config;
```
Paul-Licameli, line 66: A singleton object instead of a namespace for all these functions, but it is stateless, for now at least.  Will that need to change?  Will there ever be more than one ServiceConfig?

libraries/lib-cloud-audiocom/ServiceConfig.cpp
```diff
@@ -0,0 +1,69 @@
...
+   std::string_view audioID, std::string_view token) const
+{
+   return "http://audio.com/audacity/upload?audioId=" + std::string(audioID) +
+          "&token=" + std::string(token) +
+          "&clientId=" + std::string(GetOAuthClientID());
+}
+
+std::chrono::milliseconds ServiceConfig::GetProgressCallbackTimeout() const
+{
+   return std::chrono::seconds(3);
```
Paul-Licameli, line 55: Or to be cuter
```
using std::chrono;
return 3s;
```

libraries/lib-cloud-audiocom/ServiceConfig.cpp
```diff
@@ -0,0 +1,69 @@
...
+   return "https://audio.com/auth/sign-in/success";
+}
+
+std::string ServiceConfig::GetAPIUrl(std::string_view apiURI) const
+{
+   return std::string(GetAPIEndpoint()) + std::string(apiURI);
+}
+
+std::string ServiceConfig::GetFinishUploadPage(
+   std::string_view audioID, std::string_view token) const
```
Paul-Licameli, line 46: I like seeing `string_view` and wish we could have C++20 soon, so we can also use `std::span` to pass ranges of constant data, instead of `const std::vector<T>&`, without some irrelevant allocator type as a template parameter.

libraries/lib-cloud-audiocom/ServiceConfig.cpp
```diff
@@ -0,0 +1,69 @@
...
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  ServiceConfig.cpp
+
+  Dmitry Vedenko
+
+**********************************************************************/
+#include "ServiceConfig.h"
```
Paul-Licameli, line 11: I have build lib-cloud-audiocom and I want to understand all that is in it, and I begin here at the bottom of the dependency graph of files.

libraries/lib-cloud-audiocom/OAuthService.cpp
```diff
@@ -0,0 +1,370 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  UploadService.cpp
```
Paul-Licameli, line 6: Incorrect comment

libraries/lib-cloud-audiocom/OAuthService.cpp
```diff
@@ -0,0 +1,370 @@
...
+}
+
+bool IsPrefixed(std::string_view hay, std::string_view prefix)
+{
+   if (hay.length() < prefix.length())
+      return false;
+
+   return std::equal(
+      hay.begin(), hay.begin() + prefix.length(), prefix.begin(), prefix.end(),
+      [](auto a, auto b) { return std::tolower(a) == b; });
```
Paul-Licameli, line 76: Perhaps overload 3 might express it more clearly: https://en.cppreference.com/w/cpp/algorithm/mismatch

libraries/lib-cloud-audiocom/OAuthService.h
```diff
@@ -0,0 +1,113 @@
...
+#include <mutex>
+
+#include "Observer.h"
+
+namespace cloud::audiocom
+{
+class ServiceConfig;
+
+//! Message that is sent when authorization state changes. This message can be
+//! sent from any thread.
```
Paul-Licameli, line 25: That comment above makes me nervous.  `Observer.h` says `intended for single-threaded use only`.  Should you use `CallAfter` wherever you publish, and change the comment to say, "Messages are received in the main thread in idle time" ?

crsib: No, I'd rather stop using Observer.

Paul-Licameli: Or, should there be mutual exclusion around all `Publish` calls?  What might stil happen is that a Subscription is destroyed concurrently with a publication.  That might not have bad consequences, but I would have to reexamine Observer.h to be sure of that.

libraries/lib-cloud-audiocom/OAuthService.cpp
```diff
@@ -0,0 +1,370 @@
...
+   rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+   document.Accept(writer);
+
+   DoAuthorise(
+      config, { buffer.GetString(), buffer.GetSize() },
+      std::move(completedHandler));
+}
+
+void OAuthService::AuthoriseRefreshToken(
+   const ServiceConfig& config, std::string_view refreshToken,
```
Paul-Licameli, line 209: Shadowing of the static `refreshToken`

libraries/lib-cloud-audiocom/OAuthService.cpp
```diff
@@ -0,0 +1,370 @@
...
+#include "UrlDecode.h"
+
+#include "BasicUI.h"
+
+namespace cloud::audiocom
+{
+namespace
+{
+
+StringSetting refreshToken { L"/cloud/audiocom/refreshToken", "" };
```
Paul-Licameli, line 38: If it is stipulated in the header file as a precondition that all functions that may use this global are to be called in the main thread only then there is no need for mutexes to guard it and you can avoid the recursive mutex.  I think that is true from inspection of the entire PR, but not obvious.  The functions would be `ValidateAuth`, `HasRefreshToken`, `UnlinkAccount`

libraries/lib-cloud-audiocom/OAuthService.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+   DoAuthorise(
+      config, { buffer.GetString(), buffer.GetSize() },
+      std::move(completedHandler));
+}
+
+bool OAuthService::HasAccessToken() const
+{
+   std::lock_guard<std::mutex> lock(mMutex);
+   return !mAccessToken.empty() && Clock::now() < mTokenExpirationTime;
```
Paul-Licameli, line 276: This is equivalent to simply `!GetAccessToken().empty()`

libraries/lib-cloud-audiocom/OAuthService.cpp
```diff
@@ -0,0 +1,370 @@
...
+      args = nextArg == std::string_view::npos ? "" : args.substr(nextArg + 1);
+
+      if (IsPrefixed(arg, usernamePrefix))
+         username = arg.substr(usernamePrefix.length());
+      else if (IsPrefixed(arg, passwordPrefix))
+         password = arg.substr(passwordPrefix.length());
+      else if (IsPrefixed(arg, tokenPrefix))
+         token = arg.substr(tokenPrefix.length());
+      else if (IsPrefixed(arg, authorizationCodePrefix))
+         authorizationCode = arg.substr(authorizationCodePrefix.length());
```
Paul-Licameli, line 137: Do you want to defend against a string the defines one of the arguments more than once?

libraries/lib-cloud-audiocom/UploadService.h
```diff
@@ -0,0 +1,98 @@
...
+      //! Upload was aborted by the user
+      Aborted,
+      //! Specified file is not found
+      FileNotFound,
+      //! Authorization is required
+      Unauthorized,
+      //! audio.com has failed to understand what Audacity wants
+      InvalidData,
+      //! Audacity has failed to understand audio.com response
+      UnexpetedResponse,
```
Paul-Licameli, line 41: `UnexpectedResponse`

libraries/lib-cloud-audiocom/UploadService.h
```diff
@@ -0,0 +1,98 @@
...
+   std::string errorMessage;
+
+   //! URL to open in browser to finish up the anonymous upload
+   std::string finishUploadURL;
+   //! "Slug" to be used for sharable URL construction
+   std::string audioSlug;
+};
+
+//! Class used to track the upload operation
+class CLOUD_AUDIOCOM_API UploadPromise
```
Paul-Licameli, line 59: I think this is more like a future than a promise?

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  OAuthService.cpp
```
Paul-Licameli, line 6: incorrect comment

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
...
+std::string_view DeduceMimeType(const wxString& ext)
+{
+   if (ext == "wv")
+      return "audio/x-wavpack";
+   else if (ext == "flac")
+      return "audio/x-flac";
+   else if (ext == "mp3")
+      return "audio/mpeg";
+   else
+      return "audio/x-wav";
```
Paul-Licameli, line 45: Four types recognized here, but only three in the enumeration in ServiceConfig.h?
I notice now, that enum is only for the return of `AudioFormat GetPreferredAudioFormat` and it's always the same.  Did you forget to define an `EnumSetting` to save it as a preference?

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
...
+    std::enable_shared_from_this<UploadOperation>
+{
+   UploadOperation(
+      const ServiceConfig& serviceConfig, wxString fileName,
+      wxString projectName,
+      UploadService::CompletedCallback completedCallback,
+      UploadService::ProgressCallback progressCallback)
+       : mServiceConfig(serviceConfig)
+       , mFileName(std::move(fileName))
+       , mProjectName(std::move(projectName))
```
Paul-Licameli, line 112: `move` doesn't help with `wxString`, but no harm.

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
...
+       , mCompletedCallback(std::move(completedCallback))
+       , mProgressCallback(std::move(progressCallback))
+   {
+   }
+
+   const ServiceConfig& mServiceConfig;
+   
+   wxString mFileName;
+   wxString mProjectName;
```
Paul-Licameli, line 121: These two members can be `const`

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+   UploadService::CompletedCallback mCompletedCallback;
+   UploadService::ProgressCallback mProgressCallback;
```
Paul-Licameli, line 124: Above four members can be `const`

libraries/lib-cloud-audiocom/UploadService.h
```diff
@@ -0,0 +1,98 @@
...
+public:
+   virtual ~UploadPromise();
+
+   //! Returns true if the upload is finished
+   virtual bool IsCompleted() = 0;
+   //! Abort the upload, if running
+   virtual void Abort() = 0;
+   //! Abort the upload, if running,
+   //! notify audio.com that the uploaded file is no longer needed otherwise 
+   virtual void DiscardResult() = 0;
```
Paul-Licameli, line 70: What should happen in case this "promise" (or future) is destroyed before completion, without any calls to `Abort` or `DiscardResult` ?  It could be that `UploadOperation` still occupies some resources unnecessarily because `mAborted` isn't true.

Should `UploadPromise` call `Abort()` in its detructor if not completed?  What are you crazy?  That would be a pure virtual function call!  The base class can't do that...

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
...
+   rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+   document.Accept(writer);
+
+   return std::string(buffer.GetString());
+}
+
+
+// This class will capture itself inside the request handlers
+// by a strong reference. This way we ensure that it outlives all
+// the outstanding requests.
```
Paul-Licameli, line 100: ... so instead, the destructor of the concrete class could guarantee that one of completion, abort, or discard has happened, and then the lambdas should capture weak, not strong, references to self.

I think this would make better RAII guarantees on the main thread.  The "leak" in case there is no explicit abort on the main thread isn't until program exit, but it could mean the other thread works longer than needed before stopping.

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
...
+      
+      auto response =
+         NetworkManager::GetInstance().doPost(request, std::move(form));
+
+      mActiveResponse = response;
+      
+      response->setRequestFinishedCallback(
+         [response, sharedThis = shared_from_this(), this](auto)
+         {
+            HandleS3UploadCompleted(response);
```
Paul-Licameli, line 318: What does "S3" abbreviate?

crsib: Well, it kinda means S3 (or Simple Storage Service, but I doubt anyone calls it that). It's a somewhat standard protocol for object storages. 

https://aws.amazon.com/s3/

libraries/lib-cloud-audiocom/UploadService.cpp
crsib: This will complicate the code. Abort is not synchronous and I don't see much reason to change that (although it is achievable). Passing weak pointers will additionally increase complexity of the implementation.

Right now Promise (or future, or operation) outlives the dialog at most until lubcurl invokes any of it's callbacks. Making Abort synchronous would basically just delay the dialog closing for that specific time. I do not, however, like the idea for blocking UI thread for such reasons.

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+         for (auto it = fields.MemberBegin(); it != fields.MemberEnd(); ++it)
+            form->Add(it->name.GetString(), it->value.GetString());
+      }
+
+      const auto fileField =
+         document.HasMember("field") ? document["field"].GetString() : "file";
+
+      const wxFileName name { mFileName };
+      form->AddFile(fileField, DeduceMimeType(name.GetExt()), name);
```
Paul-Licameli, line 277: I checked to see what would happen if a nonexisted file was given.

`AddFile` throws a `SimpleMessageBoxException`.

But we get here on a worker thread, right?  Is it safe to do that?  There will be a `CallAfter` that pops up the dialog, no problem, but is there a `catch` above us here?

If not, should there be a `GuardedCall`?

Do we need to guard any other calls into `lib-network-manager`?

crsib: The file existence is checked on the main thread first. But I will improve safety here anyway.

libraries/lib-cloud-audiocom/UploadService.cpp
Paul-Licameli: Did I misunderstand the intended context of use of this, then?

Is there going to be a non-modal dialog in Audacity giving an interface that lets the user abort the operation?  Or will that be on a web page only?

libraries/lib-cloud-audiocom/UploadService.cpp
Paul-Licameli: It's good for a class at a lower level not to make too-strong assumptions, or else, to make those assumptions explicit in comments in the header.

libraries/lib-cloud-audiocom/UploadService.cpp
crsib: There is a modal dialog in Audacity that calls Abort when closing, if the promise was not fulfilled.

libraries/lib-cloud-audiocom/UploadService.cpp
crsib: This is private class. The file existence is checked by a public "factory" class before the object of this class is  created. There is a subtle chance that the file will be deleted while we are waiting to get here, so I will improve the exception guarantees at this specific point.

crsib: This class is not even listed in header, to be fair.

libraries/lib-cloud-audiocom/UserService.cpp
```diff
@@ -0,0 +1,229 @@
...
+   BasicUI::CallAfter(
+      [this]()
+      {
+         userName.Write({});
+         displayName.Write({});
+         avatarEtag.Write({});
+
+         gPrefs->Flush();
+
+         Publish({});
```
Paul-Licameli, line 125: This is the safe, main-thread publishing done my UserService.

libraries/lib-cloud-audiocom/UserService.cpp
```diff
@@ -0,0 +1,229 @@
...
+{
+   auto& oauthService = GetOAuthService();
+
+   if (!oauthService.HasAccessToken())
+      return;
+
+   using namespace audacity::network_manager;
+
+   Request request(GetServiceConfig().GetAPIUrl("/me"));
+   request.setHeader("Authorization", std::string(oauthService.GetAccessToken()));
```
Paul-Licameli, line 71: Perhaps the string literals given to `setHeader` should be named constants to avoid repetition of them within this new library.

Paul-Licameli:
My review of `lib-cloud-audiocom` and earlier commits is done.

src/cloud/audiocom/LinkAccountDialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+   {
+      s.StartInvisiblePanel(16);
+      {
+         s.SetBorder(0);
+         s.AddFixedText(XO("Enter token to link your account"));
+
+         s.AddSpace(0, 4, 0);
+
+         mToken = s.AddTextBox(TranslatableString {}, {}, 60);
+         mToken->SetName(XO("Token").Translation());
```
Paul-Licameli, line 46: You understand that `SetName` affects audible screen reader output?

crsib: Yes, that is the reason for SetName here

src/export/ExportWavPack.cpp
```diff
@@ -95,27 +95,31 @@ const std::vector< int > ExportQualityValues{
...
-   XO("24 bit") ,
-   XO("32 bit float ") ,
-};
-
-const std::vector< int > ExportBitDepthValues{
-   16,
-   24,
-   32,
-};
+extern IntSetting QualitySetting;
```
Paul-Licameli, line 99: This commit reminds me that I turn off wavpack in my local build because for some reason I never got it to work.  `wavpack/wavpack.h` can't be found.

Paul-Licameli: If I choose `system` not `local` for `audacity_use_wavpack` it doesn't even configure.

crsib: I'm not sure why it should. Do you have a system version of wavpack installed?

>This commit reminds me that I turn off wavpack in my local build because for some reason I never got it to work. wavpack/wavpack.h can't be found.

I was sure that we have overcame this problem by dropping your Conan cache.

src/cloud/audiocom/ShareAudioDialog.cpp
```diff
@@ -0,0 +1,842 @@
...
+extern ChoiceSetting FLACBitDepth;
+extern ChoiceSetting FLACLevel;
+#endif
+
+#ifdef USE_WAVPACK
+extern IntSetting QualitySetting;
+extern IntSetting BitrateSetting;
+extern IntSetting BitDepthSetting;
+extern BoolSetting HybridModeSetting;
+#endif
```
Paul-Licameli, line 59: Let it stand for alpha, but I don't like these `extern`s and I hope you don't either.  It makes the dialog dependent on src/export files but the dependency is not made explicit in an `include` directive.  Those export files are header-less because they are supposed to be highest-level things in the dependency graph, and candidates for easy removal as modules, if we can ever return to the extraction refactorings.

I'm sure you or I will think up something in the future to rewrite the necessary parts of this file as a registry that the exporters can plug into.  Making possible tinier builds of Audacity that only know Wav exporting for instance.

crsib: I don't like it, but I don't think that it is feasible to rewrite the whole Export framework. It is highly coupled with the UI and provides pretty much no sane way to configure the exporting plugin.

src/cloud/audiocom/ShareAudioDialog.cpp
```diff
@@ -0,0 +1,842 @@
...
+      return FileExtension { "wv" };
+   }
+};
+#endif
+
+// We use WAV only as a fall back (unless config asks for it explicitly)
+std::unique_ptr<ExportHelper>
+GetExportHelper(AudacityProject& project)
+{
+   const auto preferredFormat = GetServiceConfig().GetPreferredAudioFormat();
```
Paul-Licameli, line 234: Here it is... that preference that I noted before, is not a preference.  It's always WAV.  No memory of a user choice.  Must that be fixed for release?

crsib: I should be always wavpack (if available), I will tipple check that. There are no plans to give the user control of exporting right now, so this is not a bug.

crsib: And it is part of service *configuration*, it is not a preference.

src/cloud/audiocom/ShareAudioDialog.cpp
```diff
@@ -0,0 +1,842 @@
...
+      unsigned long long numerator, unsigned long long denominator,
+      const TranslatableString&) override
+   {
+      mParent.UpdateProgress(numerator, denominator);
+
+      const auto now = Clock::now();
+
+      // Exporter polls in the main thread. To make the dialog responsive
+      // periodic yielding is required
+      if ((now - mLastYield > std::chrono::milliseconds(50)) || (numerator == denominator))
```
Paul-Licameli, line 291: You want a busy-wait when numerator equals denominator?

crsib: This is the behavior from the original progress bar.

src/cloud/audiocom/ShareAudioDialog.cpp
```diff
@@ -0,0 +1,842 @@
...
+
+   if (current == 0)
+      return;
+
+   if (current > total)
+      current = total;
+
+   if (mLastProgressValue != current)
+   {
+      constexpr int scale = 10000;
```
Paul-Licameli, line 590: Is this the right power of 10?

crsib: This value is unrelated to the original progress dialog. I can use virtually any number greater that 1.

10000 is probably an overkill, but it doesn't affect the result much. Rescaling is needed because wxGauge does not allow 64 bit values.

.github/workflows/build.yml
```diff
@@ -100,7 +100,10 @@ env:
   BUILD_TYPE: ${{ (inputs && inputs.build_type) || (github.event.inputs && github.event.inputs.build_type) || 'RelWithDebInfo' }}
   CONFIGURATION_TYPES: ${{ (inputs && inputs.configuration_types) || (github.event.inputs && github.event.inputs.configuration_types) || 'RelWithDebInfo' }}
 
-  CONFIGURE_CMAKE_OPTIONS: ${{ (inputs && inputs.configure_cmake_options) || (github.event.inputs && github.event.inputs.configure_cmake_options) || '' }}
+  CONFIGURE_CMAKE_OPTIONS: |
+    -Daudacity_has_url_schemes_support=On
+    -Daudacity_has_audiocom_upload=On
+    ${{ (inputs && inputs.configure_cmake_options) || (github.event.inputs && github.event.inputs.configure_cmake_options) || '' }}
```
Paul-Licameli, line 106: I'm not competent in this language yet

src/cloud/ShareAudioToolbar.cpp
```diff
@@ -0,0 +1,245 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  ShareAudioToolbar.cpp
```
Paul-Licameli, line 6: What was the reasoning to put this in `src/cloud` but the rest in `src/cloud/audiocom`?

libraries/lib-cloud-audiocom/OAuthService.h
crsib: I have reviewed the usages, it can be moved to main thread for now.

Paul-Licameli:
Done

libraries/lib-cloud-audiocom/OAuthService.cpp
crsib: Nope

libraries/lib-cloud-audiocom/UploadService.cpp
crsib: Only the first two. I want to be able to reset the handlers, so the associated resources can be released as well

src/cloud/ShareAudioToolbar.cpp
Paul-Licameli: I’m not asking for a change.

libraries/lib-cloud-audiocom/OAuthService.cpp
```diff
@@ -0,0 +1,380 @@
...
+{
+   using namespace audacity::network_manager;
+
+   Request request(config.GetAPIUrl("/auth/token"));
+
+   request.setHeader(
+      common_headers::ContentType, common_content_types::ApplicationJson);
+   
+   request.setHeader(
+      common_headers::Accept, common_content_types::ApplicationJson);
```
Paul-Licameli, line 307: This line is new since the last push.  Is it correct?

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,531 @@
...
+   {      
+      using namespace audacity::network_manager;
+
+      Request request(mServiceConfig.GetAPIUrl("/audio"));
+         
+      request.setHeader(
+         common_headers::ContentType, common_content_types::ApplicationJson);
+      
+      request.setHeader(
+         common_headers::Accept, common_content_types::ApplicationJson);
```
Paul-Licameli, line 209: New too

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -0,0 +1,531 @@
...
+      {
+         mLastProgressReportTime = now;
+
+         using namespace audacity::network_manager;
+         Request request(mProgressUrl);
+         
+         request.setHeader(
+            common_headers::ContentType, common_content_types::ApplicationJson);
+         request.setHeader(
+            common_headers::Accept, common_content_types::ApplicationJson);
```
Paul-Licameli, line 372: Also new... okay I trust this is a pattern done for a good reason

src/cloud/audiocom/ShareAudioDialog.cpp
Paul-Licameli: It's still true in the latest push that the return value of this function never changes.  Is it your intent to check availability of WavPack?

I think I know how to do that.  Wait for me to push a PR onto your PR.

Paul-Licameli (changes requested):
One lock_guard is needed, I think, but all else is good and resolves all my questions

Paul-Licameli:
Another omission you might fix:  make a `AttachedToolBarMenuItem` in ShareAudioToolbar.cpp analogous with other toolbars, so that there is a new View > Toolbars menu item.

Paul-Licameli (approved):

crsib:
There will be additional changes added

Paul-Licameli:
> There will be additional changes added

OK, I'll merge my two little PRs rather than yield to you now

libraries/lib-cloud-upload/CloudExporterPlugin.h
```diff
@@ -0,0 +1,43 @@
...
+**********************************************************************/
+#pragma once
+
+#include <wx/string.h>
+
+#include "Identifier.h"
+
+namespace cloud
+{
+//! Helper interface, that allows to setup the desired export format on the ExportPluging.
```
Paul-Licameli, line 19: ExportPlugin

libraries/lib-cloud-upload/CloudExporterPlugin.h
```diff
@@ -0,0 +1,43 @@
...
+/*!
+ * Audacity exporting framework relies heavily on using the global state.
+ * This interface attempts to overcome this problem by allowing to setup the desired
+ * export format on the ExportPluging before exporting.
```
Paul-Licameli, line 23: ExportPlugin

Paul-Licameli (approved):

Paul-Licameli:
The hidden dependency is broken, so mod-pcm etc. might still be extracted -- good.

The configuration of preferred types is stored in lib-cloud-audiocom and not in the UI layer.  OK.

Otherwise I'm not sure I understand the other objections to my draft solution, or why there is an essential difference.

Why would it be so wrong in principle to combine the two abstract interfaces if in future the ExportPlugin interface could be demoted into a library too?  The concrete implementations of ExportPlugin and associated CloudExporterPlugin end up side-by-side anyway.

And it's not even necessary, for cycle breaking, to demote the new CloudExporterPlugin interface into a library.  lib-cloud-audiocom depends on lib-cloud-upload -- but really only for MimeTypesList which is just a few type aliases.

I don't know quite what you mean about the existing exporter framework being so difficult to rewrite. It is true that it has its own "statefulness" problem much like effects.  Each particular exporter has a set of global variables for preferred settings.  Is there some great need for a menu of "Presets" for each exporter?  Would we store those in std::anys? ...

Finally my foolish question, why can't any file format you like be uploaded to cloud, then downloaded again to another Audacity executable that can interpret the same format?  Why can't we have open-endedness here?  I don't understand the technical limitation that the server would impose againt it.

crsib:
> Why would it be so wrong in principle to combine the two abstract interfaces

Well, this is wrong ... in principle. For example, it breaks ISP, which is I in SOLID. I do not like the implementation I proposed either, but provided the current framework it will have to stay for now. Hopefully, it would be removed one day.

 ExportPlugin should know nothing about the possibility of audio uploads. Uploading is an optional feature. Uploading depends on exporting, not vice versa. I can easily see how we need to extend those interfaces based on user testing, I do not want to change ExportPlugin interface to do that. You are correct, that implementations reside close to the plugins now, but that is not because I like it there. There is simply no other feasible options right now.

> std::anys

To be fair I had never seen usage of `any` that solved problems instead of introducing ones. Scripting language binding, commonly referred as a `good example`, are not using `any`. Scripting languages have a very well defined set of types and bindings must know how to map the specific host language type into a scripting language type. `any` has virtually no use there. Python has a stronger type system than C, so to speak.

Speaking of exporter configuration I am afraid that FFmpeg approach is the "nessecary evil". That is - we have a string identifier mapping a well defined variant, with values checked by a plugin for validity. Plugin, on the other hand, provides a list of supported options with the constraints on them, so the UI can be built.

Using `any` as a configuration will not allow interface clients to supply configuration to the plugin unless we link against a specific plugin. However, I like the presets idea very much.

> why can't any file format you like be uploaded to cloud

audio.com is an audio sharing service. That is - it needs to know how to play and how to transcode the uploaded audio .

Paul-Licameli:
I'm not sure I understood all of the previous.  std::any is surely a dangerous escape into typelessness, unlike std::variant, and not to be overused, but it does allow open-endedness as variant does not.  I think it is the right tool for the job in the system of effects.

There are "producers" of anys (Effects doing MakeSettings) and "consumers" that can interpret them (instances) but in between a custodian that "warehouses" them (RealtimeEffectState).  The middleman needs to belong to a lower level of the dependency graph, and not need to know a closed-ended variant of possibilities, which would violate "D".  The middleman's job is to guarantee only that it gives the right anys back to the right parties that ask to use them and know how to interpret them.

crsib:
> but it does allow open-endedness

I do not agree, but I though we agreed to disagree here :-)

> I think it is the right tool for the job in the system of effects.

There are significant limitations that will make implementing certain features expensive or not possible. On top of that, it makes debugging hard. So we are not on the same page here as well. (Think of any feature that needs a list of parameters!)

 > which would violate "D"

I'm not sure how D will be violated, but L is violated right now. D, as in dependency inversion, is very strict on L. Otherwise it is impossible to make the system safe.

> The middleman needs

I am not convinced yet that this middleman itself is needed and there is definitely no need for it in exporters :-) Using similar approach for exporters will not solve the issue we are discussing right now.

Paul-Licameli:
I don’t understand what alternative you would suggest for the system of effects.

EffectSettings may have been stretched for too many purposes that should be separated — yes.  The inter-thread communication could be simpler.  I have ideas for that I would like to do next release.

Still see the need for Real-time Effect State to store some polymorphic data for the related stateless Effect object to interpret and serialize and deserialize.

crsib:
>  stateless Effect object to interpret and serialize and deserialize

This just means, that stateless Effect is not serializable, the statefull EffectInstance is. 

> I don’t understand what alternative you would suggest for the system of effects.

Quite a complex topic to discuss on a train using a phone. Can we discuss it sometimes later please?

Issue #3427, opened 2022-08-25
VST3: The effect's controls are reset to default when repeatedly opening up its dialog window


dozzzzer:
### Bug description

When instantiating an effect in Realtime Effects panel, changing its settings, closing it and re-opening it, its parameters get reset to default.

### Steps to reproduce

1. load some audio
2. open up Realtime Effects panel
3. select an effect, click the cog wheel button
4. change its parameters and close it
5. click the cog wheel button again
6. observe

### Expected behavior

The effect's parameters are reset to their default state.

### Actual behavior

The previously set parameters persist.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

The issue occurs only with VST3 plugins.

PR #3436, opened 2022-08-26
Fixes number of problems with VST3 settings

vsverchinsky:
Resolves: #3427
Resolves: #3437 

VST3Wrapper now stores latest parameter values as a part of it's state that, presumably, would help avoid issues with consistency since `IComponent::setState` are applied immediately while parameter changes aren't applied until consumed on the worker side. Though for this reason it is possible that there could be audiable transition when preset is applied (see `VST3UIValidator::UpdateUI`) not caused by absense of interpolation.

Recommended:

### DIFF
--- a/src/effects/VST3/VST3UIValidator.cpp
+++ b/src/effects/VST3/VST3UIValidator.cpp
@@ -143,7 +143,7 @@ bool VST3UIValidator::ValidateUI()
    mAccess.ModifySettings([&](EffectSettings &settings){
       if (mDuration != nullptr)
          settings.extra.SetDuration(mDuration->GetValue());
-      mWrapper.FlushSettings(settings);
+      mWrapper.FlushParameters(settings);
       mWrapper.StoreSettings(settings);
    });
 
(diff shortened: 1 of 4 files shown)


crsib:
1. Generate chirp
2. Add Berzerk Distortion Mono
3. Open UI using cogwheel
4. Press play
5. Observe nullopt dereference

After opening UI I observe that settings have 40 changes, but `vst3settings->processorState == nullopt`

Without opening UI I observe no problems

src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -176,6 +186,9 @@ bool VST3UIValidator::UpdateUI()
    if (mPlainUI != nullptr)
       mPlainUI->ReloadParameters();
 
+   //Write to main...
+   mAccess.Flush();
```
Paul-Licameli, line 190: Is this flush needed to fix the bug?

It used to be that there was no Flush, and every Get did what Flush does.  I thought this caused too slow updating of meters, at least for AudioUnits, because of frequent UpdateUI calls. so I made Flush a different thing called less often.

crsib: Flush is called frequently in LV2 now without observable performance hit. Looking into debugger I can say it is called more frequently than UpdateUI does.

vsverchinsky: In VST3 UpdateUI called only on preset change

crsib:
@Paul-Licameli can you please have a look at this?

crsib:
I have checked and see not issues, but you are the expert in this area, so I'd like to know your opinion as well. Please merge, if you are ok with this PR

src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -166,16 +166,31 @@ void VST3UIValidator::OnClose()
 
    mWrapper.EndParameterEdit();
 
+   //Make sure all previous changes has been processed by the worker
+   mAccess.Flush();
```
Paul-Licameli, line 170: Maybe  not needed with idle time flushes now

crsib: Disagree. This flush happens at a very specific time. Removing it may result in rare random bugs

Paul-Licameli (approved):

Issue #3416, opened 2022-08-23
Ensure correct effects are displayed in RT panel


crsib:
- [ ] Audacity Reverb is not available
- [ ] Audacity Wahwah is not available
- [ ] Generators (such as LV2) are not available
- [ ] VST is not available (if VST is not ready by alpha)

PR #3446, opened 2022-08-26
Updates the RealtimeSupport for certain effects

crsib:
Resolves: #3416 

Recommended:

### DIFF
--- a/src/effects/Reverb.cpp
+++ b/src/effects/Reverb.cpp
@@ -302,7 +302,7 @@ EffectType EffectReverb::GetType() const
 
 auto EffectReverb::RealtimeSupport() const -> RealtimeSince
 {
-   return RealtimeSince::Since_3_2;
+   return RealtimeSince::Never;
 }
 
 static size_t BLOCK = 16384;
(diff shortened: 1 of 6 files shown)


Paul-Licameli:
I approve these simple changes, but are they sufficient?  What happens if you
- Remove pluginregistry.cfg
- Start Audacity 3.1.3
- Exit
- Open Audacity 3.2. alpha

Does it still correctly exclude non-processing plug-ins from the sidebar?  If not, then users upgrading to 3.2 won't have the intended behavior.

crsib:
Hmm, probably you are right and it should be `Since_3_2`. This will cause problems when downgrading, but we already have them.

vsverchinsky (approved):

Paul-Licameli:
Can you give exmples of plug-ins that are not processors?

Paul-Licameli:
Here's one: https://www.maat.digital/gon/ in VST and AU versions, but VST is off the table for 3.2, and our AU code actually doesn't correctly report whether effects are processors!  (More about that another time)

crsib:
Plenty of generators, at least. 

https://soundsmag.com/top-generator-vst-plugins/

I have several LADSPA generators as well from the SWH

Issue #3448, opened 2022-08-26
Persistency of AU effect settings in sidebar is broken again


Paul-Licameli:
### Bug description

It worked at commit 6f6501c7c28c0a5f72149ed0f89274ad45adfbd7 but has regressed since then, and the problem bisects to commit 32a863b142738b5acd1e98f786ad7d3190cd2341

### Steps to reproduce

1. Generate noise, open sidebar
2. Add an AU effect, for instance AUGraphicEQ
3. Change some settings, for instance drag 1K slider to the top
4. Close and reopen dialog, observe that in-session memory of settings is OK

Then do either (but not both) of these:

- Play -- hear WRONG results; then Stop, play again -- hear RIGHT results

or

- Save new project, close, reopen, reopen sidebar, reopen effect settings; see WRONG settings

### Expected behavior

Hear the right thing, or see the same settings as you saved

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

### Additional context

This does NOT happen for LV2 effects, so it's not general across effect families.

VST3 persistency might have bugs too, which I did not check, considering other problems with basic stickiness of settings after closing the dialog.

The bisection is rather surprising.  The commit that broke AU persistency touches no code special to AU.

Paul-Licameli:
The button formerly called "OK" became "Apply" at commit 32a863b142738b5acd1e98f786ad7d3190cd2341 and changes behavior:  it applies the effect destructively whereas OK did not do that at the previous commit.

If you Apply and then Undo at commit 32a863b142738b5acd1e98f786ad7d3190cd2341, then save -- it "works."  But the "Close" button does not persist the effect.

crsib:
The commit you've mentioned was written for the case when none of the buttons are present. The was the whole idea behind the PR to be fair. The omission of full scale AU testing is unfortunate, but Mac is not my main platform, so I haven't put enough testing time in it, while still checking the basics to work.

PR #3449, opened 2022-08-26
A single Flush() to fix persistency of all effect families...

Paul-Licameli:
... Not just AU.

Resolves: #3448

Recommended:


src/effects/EffectUI.cpp
```diff
@@ -508,6 +508,13 @@ void EffectUIHost::OnClose(wxCloseEvent & WXUNUSED(evt))
    DoCancel();
    CleanupRealtime();
 
+   // Flush settings when the dialog closes.  But please note -- this makes
+   // close always commit settings changes.  If it is decided to distinguish
+   // a cancellation of the dialog from a commit, then this call should move
+   // somewhere else.
+   if (mpAccess)
```
crsib, line 515: Should we distinguish between realtime and destructive processing here? Probably we should.

Paul-Licameli: I think not.  If you step through it, you will see that for destructive application, an EffectSettingsAccessTee flushes to a SimpleEffectSettings access -- which is no-op -- and to the side, an EffectSettingsAccess connected to a temporary RealtimEffectState which we don't need anyway.

Paul-Licameli: However, I may want to redraft this anyway.  It may be that one `Flush()` in an idle time handler of the `EffectUIHost` is all we need.

Paul-Licameli:
Alternative fix; and more steps to reproduce in the issue.

crsib (approved):
I like this solution most

PR #3453, opened 2022-08-28
Zoom toggle

bjj:
The zoom toggle function (shift-Z) did not have good viewport positioning logic when the selection span was much larger than zero.

The first commit fixes a sign error in `ProjectWindow::Zoom` which caused zoom toggling to center on a region one selection width to the left of the actual selection.

However, with that fixed, the behavior still felt bad compared to other forms of zoom, so the second patch changes `OnZoomToggle` to use the much richer `ZoomInByFactor` and `ZoomOutByFactor` logic as appropriate. This make zoom toggling feel "right" to me.

Recommended:

### DIFF
--- a/src/ProjectWindow.cpp
+++ b/src/ProjectWindow.cpp
@@ -1636,7 +1636,7 @@ void ProjectWindow::Zoom(double level)
    float t0 = viewInfo.selectedRegion.t0();
    float t1 = viewInfo.selectedRegion.t1();
    float tAvailable = viewInfo.GetScreenEndTime() - viewInfo.h;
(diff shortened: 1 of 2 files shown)


LWinterberg:
Hello! We just [switched master to 4.0.0](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444) which makes this PR incompatible and/or obsolete as we're switching UI framework and reorganizing large parts of the codebase. This includes new track and zooming behavior.

You're very welcome to make these kinds of contributions in the future though, after things stabilize a little bit.

Issue #3469, opened 2022-08-29
Effect's settings get reset when clicking Play and having effect's dialog window open


dozzzzer:
### Steps to reproduce

0. Install [Klone](https://bedroomproducersblog.com/2021/11/29/the-klone-fazertone/) plugin
1. Open it from the RT effects panel
2. Change its parameters and don't close the dialog window
3. Click Play
4. Observe

### Expected behavior

No changes to the effect's controls

### Actual behavior

The controls are getting reset to the default state

### Audacity Version

other

### Operating system

Other

Issue #3431, opened 2022-08-25
Realtime Effects: (VST3) when loading factory defaults and playing the track, the plugin loads the last saved user preset


dozzzzer:
### Steps to reproduce

1. Load some audio
2. In the Realtime Effects panel select a plugin (VST3) and open its settings
3. Make some changes to the effect's parameters and save it as a user preset
4. Load the effect's defaults
5. Start playback
6. Observe

### Expected behavior

The effect's parameters remain the way they were set just before the playback started = factory defaults.

### Actual behavior

The effect's parameters revert to their previous state, i.e. after the user preset was saved.

https://user-images.githubusercontent.com/70063970/186736054-e2a590c2-d55c-47c8-98e4-e3f4ce47acef.mp4

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

### Additional context

Built-ins/LADSPA/VST aren't affected by this issue.

Paul-Licameli:
@dozzzzer AudioUnits?

dozzzzer:
@Paul-Licameli ...and VST3s

Paul-Licameli:
Likely there will be separate fixes for VST3 and AU, therefore I duplicatd this issue at https://github.com/audacity/audacity/issues/3441

AnitaBats:
@dozzzzer please re-test against mentioned PR, should not be reproducible any longer

dozzzzer:
Verified on MacOS & Windows. Indeed all good now, closing the ticket.

vsverchinsky:
PR isn't merged yet

Ryan-Miller-Audacity:
This will be briefly tested in Alpha now that it has been merged as well. Unassigning @dozzzzer and assigning myself with the other merged tickets of #3475.

PR #3475, opened 2022-08-30
VST3 settings fixes

vsverchinsky:
Resolves: #3469
Resolves: #3431
Resolves: #3480 
Resolves: #3530

More fixes for VST3 statelessness

Recommended:

### DIFF
--- a/src/effects/VST3/VST3UIValidator.cpp
+++ b/src/effects/VST3/VST3UIValidator.cpp
@@ -181,10 +181,8 @@ void VST3UIValidator::OnClose()
 
 bool VST3UIValidator::UpdateUI()
 {
-   mWrapper.EndParameterEdit();
-   mWrapper.FetchSettings(mAccess.Get());
-   mWrapper.BeginParameterEdit(mAccess);
-
+   mAccess.ModifySettings([&](EffectSettings& settings) { mWrapper.FetchSettings(settings); });
+   
    if (mPlainUI != nullptr)
       mPlainUI->ReloadParameters();
 
--- a/src/effects/RealtimeEffectState.cpp
+++ b/src/effects/RealtimeEffectState.cpp
@@ -99,6 +99,8 @@ public:
          Reader(ToMainSlot &&slot, SettingsAndCounter &settings,
             const std::shared_ptr<EffectInstance> pInstance
          ) {
+            if(settings.counter == slot.mSettings.counter)
+               return;
             settings.counter = slot.mSettings.counter;
             if (pInstance)
                pInstance->AssignSettings(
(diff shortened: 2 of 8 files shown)


src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -181,10 +181,8 @@ void VST3UIValidator::OnClose()
 
 bool VST3UIValidator::UpdateUI()
 {
-   mWrapper.EndParameterEdit();
-   mWrapper.FetchSettings(mAccess.Get());
-   mWrapper.BeginParameterEdit(mAccess);
-
+   mAccess.ModifySettings([&](EffectSettings& settings) { mWrapper.FetchSettings(settings); });
```
Paul-Licameli: For all other effect families, ValidateUI() and UI event handlers use ModifySettings to start the inter-thread message to the running instance, but UpdateUI() is supposed to move data the other way -- take changes of instance state, and update the appearance of a dialog to reflect that; not to change instance state.

But here UpdateUI is doing something to the wrapper immediately in this thread, but ALSO starting the inter-thread communication that later causes ConsumeChanges().

Is all of that really needed to fix the bug?

Or, was ModifySettings done for inessential reasons, because now FetchSettings takes a non-const argument?

See the suggested extra commit I pushed and please tell me if it is a simplification that preserves the fixes, and squash it, or tell me it's definitely wrong and force-push the branch to remove it.

Paul-Licameli (approved):
If my suggested simplifications preserve the bug fixes, I would like you to squash them.  If they do not, then omit them.

I approve the rest

Paul-Licameli:
Linux build didn't succeed yet.  I pushed another one line to fix it.

src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -181,7 +181,7 @@ void VST3UIValidator::OnClose()
 
 bool VST3UIValidator::UpdateUI()
 {
-   mAccess.ModifySettings([&](EffectSettings& settings) { mWrapper.FetchSettings(settings); });
+   mWrapper.FetchSettings(mAccess.Get());
```
vsverchinsky, line 184: The problem is that some plugins use `IComponent::performEdit` to deliver the changes to the host (that should be followed by `IAudioProcessor::process`, for that I need `ModifySettings`) and do not update internal processing state during `IAudioProcessor::setState`

vsverchinsky:
> If my suggested simplifications preserve the bug fixes, I would like you to squash them. If they do not, then omit them.
> 
> I approve the rest

Thanks, but I will delete the first commit, previous comment explains that

Paul-Licameli:
Approved, but, Git falsely credits me as author of the commit!  I don't want to take undue credit.

Paul-Licameli (approved):

PR #3486, opened 2022-08-31
Fixes to theme authoring issues

bjj:
While working on theme-related stuff I ran into some issues that made it harder than necessary.

Recommended:


LWinterberg:
You're probably getting tired of these emails by now, but yes: Audacity 4.0 also will  have a new theming system!

PR #3497, opened 2022-08-31
update video link

LWinterberg:
This updates the what's new link and also the link in the realtime effect sidepanel. (the video thumbnail comes another time)

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -1005,8 +1005,7 @@ public:
       addEffectHint->SetForegroundColorIndex(clrTrackPanelText);
       mAddEffectHint = addEffectHint;
 
-      //TODO: set link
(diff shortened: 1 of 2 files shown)


src/RealtimeEffectPanel.cpp
```diff
@@ -1006,7 +1006,7 @@ class RealtimeEffectListWindow : public wxScrolledWindow
       mAddEffectHint = addEffectHint;
 
       //TODO: set link
-      auto addEffectTutorialLink = safenew ThemedWindowWrapper<wxHyperlinkCtrl>(this, wxID_ANY, _("Watch video"), "https://www.audacityteam.org", wxDefaultPosition, wxDefaultSize, wxHL_ALIGN_LEFT | wxHL_CONTEXTMENU);
+      auto addEffectTutorialLink = safenew ThemedWindowWrapper<wxHyperlinkCtrl>(this, wxID_ANY, _("Watch video"), "https://www.audacityteam.org/3.2.0-video", wxDefaultPosition, wxDefaultSize, wxHL_ALIGN_LEFT | wxHL_CONTEXTMENU);
```
crsib, line 1009: Hmm, probably we should have a different link (probably linking to the same page for now).

This will allow us to add RT specific tutorial, if we choose to do so.

LWinterberg: that's a good idea

PR #3512, opened 2022-09-01
Open ended uploaders

Paul-Licameli:

Fix hidden dependency in new uploading code.

This doesn't solve the problem of WavPack support being possibly absent in the deployment, but could be adapted to do something in that case.

Recommended:


src/cloud/audiocom/ShareAudioDialog.cpp
```diff
@@ -69,105 +69,93 @@ const wxSize avatarSize = { 32, 32 };
...
-
-   explicit ExportHelper(AudacityProject& project)
-       : mProject { project }
-   {
-   }
+   virtual ~ExportHelper();
+
+   //! May modify global settings affecting export;
+   //! not responsible to revert changes
+   virtual void SetupUploadFormat() = 0;
```
crsib: This interface was never intended for public use. It was solving very small and closed problem.

libraries/lib-cloud-audiocom/ServiceConfig.cpp
```diff
@@ -55,11 +55,6 @@ std::chrono::milliseconds ServiceConfig::GetProgressCallbackTimeout() const
    return std::chrono::seconds(3);
 }
 
-AudioFormat ServiceConfig::GetPreferredAudioFormat() const
```
crsib: I strongly disagree with this change. You lift responsibility from the business logic into the UI.

libraries/lib-cloud-audiocom/ServiceConfig.h
```diff
@@ -16,14 +16,6 @@
 
 namespace cloud::audiocom
 {
-//! Preferred audio format for the upload
-enum class AudioFormat
```
crsib: This can go, although I do not see a real reason.

libraries/lib-cloud-audiocom/ServiceConfig.h
```diff
@@ -42,8 +34,6 @@ class CLOUD_AUDIOCOM_API ServiceConfig final
    std::string GetFinishUploadPage(std::string_view audioID, std::string_view token) const;
    //! Timeout between progress callbacks
    std::chrono::milliseconds GetProgressCallbackTimeout() const;
-   //! Preferred audio format
-   AudioFormat GetPreferredAudioFormat() const;
```
crsib: But this should not

src/export/Export.h
```diff
@@ -108,6 +108,17 @@ class AUDACITY_DLL_API ExportPlugin /* not final */
     * of channels in exported file. -1 for unspecified */
    virtual int SetNumExportChannels() { return -1; }
 
+   // TODO:  the next three virtuals might take arguments, as does
+   // GetFormat()
+
+   //! May modify global settings affecting export;
+   //! not responsible to revert changes
+   virtual void SetupUploadFormat();
```
crsib, line 116: I strongly disagree with this change. `ExportPlugin` has nothing to do with uploading.

src/export/ExportWavPack.cpp
```diff
@@ -495,6 +499,24 @@ void ExportWavPack::OptionsCreate(ShuttleGui &S, int format)
    S.AddWindow( safenew ExportWavPackOptions{ S.GetParent(), format } );
 }
 
+void ExportWavPack::SetupUploadFormat()
```
crsib, line 498: This breaks compilation.

crsib (commented):
I do not agree with the proposed changes. 

1. Service configuration is moved to UI for no reason. 
2. The initial interface was never intended to be public. It is not enough for the general case.
3. I strongly disagree that ExportPlugin should provide any interface related to uploading.

I know of a fix that will satisfy you, without violating cases listed above. But really, there are more important issues to focus on right now.

Paul-Licameli:
Please understand that the main point is to break a dependency that would interfere with some of the module extractions I have had prepared for a long time, and that I hope we can begin to merge in another release.

Each of the several Export plug-ins was ready for extraction out of the monolithic executable as a highest-level node, but this PR would interfere with that for PCM, WavPack, and FLAC.

I thought the uploading logic shouldn't have dependencies on any particular close-ended list of formats but would be better designed with open-endedness, allowing the choice of format to be specified by a string identifier.

If that's all wrong, and you want a closed-ended list -- then to break up the monolith, the uploading would have to become an even higher level module dependent on three of the exporters, which would then need to be not header-less.

Would that in fact be the better large scale organization?  If so, then I would simply make new headers for those three exporters instead of the extern declarations.

crsib:
I think exporting framework needs a rework similar to the effects framework. I do not see much point in extracting it out as is. 

If the framework would provide an open-ended way to configure the export - this problem wouldn't exist at all. Using the global state does not feel to be a right choice at all.

I do not agree, that for the sake of breaking dependencies we should make ExportPlugin interface heavier than it is now. More over, I think those are not enough to handle the generic case. Even the different approach I will soon present to you won't be completely sufficient.

The list of formats supported by a 3d party service is not our choice, this is not something really extensible on our side.

Paul-Licameli:
> I think exporting framework needs a rework similar to the effects framework. I do not see much point in extracting it out as is.
> 
> If the framework would provide an open-ended way to configure the export - this problem wouldn't exist at all. Using the global state does not feel to be a right choice at all.
> 
> I do not agree, that for the sake of breaking dependencies we should make ExportPlugin interface heavier than it is now. More over, I think those are not enough to handle the generic case. Even the different approach I will soon present to you won't be completely sufficient.
> 
> The list of formats supported by a 3d party service is not our choice, this is not something really extensible on our side.

I need to understand more.  Meanwhile if my attempt at breaking the hidden dependency is unacceptable, let's not make the dependency break a necessity for now.  I approve your PR without it.

Issue #3483, opened 2022-08-30
realtime effect dialogs: playback cannot be started or stopped using the keyboard within the dialog


DavidBailes:
### Bug description

realtime effect dialogs: playback cannot be started or stopped using the keyboard within the dialog - you have to make the main window the focus to be able to do that.

### Steps to reproduce

1. Open a realtime effect.
2.  Try to start/stop playback without leaving the dialog.

### Expected behavior

You can.

### Actual behavior

You can't.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

This is a regression.

LWinterberg:
I would expect for `Space` to not toggle the power button but always play.

DavidBailes:
> I would expect for `Space` to not toggle the power button but always play.

Are you suggesting this behaviour for the power button, but not for any other controls in the dialog?

LWinterberg:
no, for everything - I'd expect it to always play no matter what's selected

dozzzzer:
Verified on macOS and Windows.

PR #3539, opened 2022-09-05
Enables CommandManagers keystroke handling in realtime effect UI

vsverchinsky:
Resolves: #3483

Recommended:

### DIFF
--- a/src/commands/CommandManager.cpp
+++ b/src/commands/CommandManager.cpp
@@ -171,12 +171,13 @@ struct CommandListEntry
    bool useStrictFlags{ false };
 };
 
-NonKeystrokeInterceptingWindow::~NonKeystrokeInterceptingWindow()
-{
-}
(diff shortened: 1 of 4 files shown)


src/commands/CommandManager.cpp
```diff
@@ -171,12 +171,13 @@ struct CommandListEntry
...
-{
-}
+NonKeystrokeInterceptingWindow::~NonKeystrokeInterceptingWindow() = default;
 
-TopLevelKeystrokeHandlingWindow::~TopLevelKeystrokeHandlingWindow()
+TopLevelKeystrokeHandlingWindow::~TopLevelKeystrokeHandlingWindow() = default;
+
+bool TopLevelKeystrokeHandlingWindow::HandleCommandKeystrokes()
 {
+   return false;
```
Paul-Licameli, line 180: Should the default return true, to preserve old behavior except in cases where you override it?

vsverchinsky: Right, thanks!

Issue #3489, opened 2022-08-31
"Presets & settings" button in effect dialogs does not have an access key


DavidBailes:
### Steps to reproduce

Open an effect dialog

### Expected behavior

Presets & settings button has an access key

### Actual behavior

Presets & settings button does not have an access key

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

The Presets & settings button replaces the Manage button in previous versions of Audacity.
The Manage button had the access key M. So once an access key is assigned (which won't be M!), then all the built-in effects will need to be checked to see if that access key is used for another control, and if so access key or keys reassigned so that access keys remain unique in a dialog.

LWinterberg:
Hello! We just [switched development to Audacity 4](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444). We've redone, or plan on redoing, various things from scratch, including the entire UI, and the part this bug is about. As such, I'm closing this bug as it will be fixed there and we won't be fixing it for the final releases of Audacity 3.

PR #3540, opened 2022-09-05
Adds access key to the presets & settings button

LWinterberg:
(previously "manage"). Moves existing access keys on s to new places. Also resolves a conflict on p in the process.

Resolves: #3489

### DIFF
--- a/src/effects/BassTreble.cpp
+++ b/src/effects/BassTreble.cpp
@@ -195,7 +195,7 @@ std::unique_ptr<EffectUIValidator> EffectBassTreble::PopulateOrExchange(
             .Name(XO("Bass (dB):"))
             .Validator<FloatingPointValidator<double>>(
                1, &mBass, NumValidatorStyle::DEFAULT, Bass.min, Bass.max)
-            .AddTextBox(XXO("Ba&ss (dB):"), L"", 10);
+            .AddTextBox(XXO("&Bass (dB):"), L"", 10);
 
          mBassS = S.Id(ID_Bass)
             .Name(XO("Bass"))
--- a/src/effects/ChangePitch.cpp
+++ b/src/effects/ChangePitch.cpp
@@ -292,7 +292,7 @@ std::unique_ptr<EffectUIValidator> EffectChangePitch::PopulateOrExchange(
                   2, &m_dSemitonesChange,
                   NumValidatorStyle::TWO_TRAILING_ZEROES
                )
-               .AddTextBox(XXO("&Semitones (half-steps):"), wxT(""), 12);
+               .AddTextBox(XXO("S&emitones (half-steps):"), wxT(""), 12);
          }
          S.EndHorizontalLay();
       }
--- a/src/effects/ChangeSpeed.cpp
+++ b/src/effects/ChangeSpeed.cpp
@@ -268,7 +268,7 @@ std::unique_ptr<EffectUIValidator> EffectChangeSpeed::PopulateOrExchange(
                3, &mMultiplier,
                NumValidatorStyle::THREE_TRAILING_ZEROES,
                Percentage.min / 100.0, ((Percentage.max / 100.0) + 1) )
-            .AddTextBox(XXO("&Speed Multiplier:"), L"", 12);
+            .AddTextBox(XXO("Speed &Multiplier:"), L"", 12);
 
          mpTextCtrl_PercentChange = S.Id(ID_PercentChange)
             .Validator<FloatingPointValidator<double>>(
--- a/src/effects/ClickRemoval.cpp
+++ b/src/effects/ClickRemoval.cpp
@@ -302,7 +302,7 @@ std::unique_ptr<EffectUIValidator> EffectClickRemoval::PopulateOrExchange(
       mWidthT = S.Id(ID_Width)
          .Validator<IntegerValidator<int>>(
             &mClickWidth, NumValidatorStyle::DEFAULT, Width.min, Width.max)
-         .AddTextBox(XXO("Max &Spike Width (higher is more sensitive):"),
+         .AddTextBox(XXO("&Max Spike Width (higher is more sensitive):"),
                      wxT(""),
                      10);
 
--- a/src/effects/Contrast.cpp
+++ b/src/effects/Contrast.cpp
@@ -305,7 +305,7 @@ ContrastDialog::ContrastDialog(wxWindow * parent, wxWindowID id,
          S.Name(XO("Background end time"))
             .AddWindow(mBackgroundEndT);
 
-         m_pButton_UseCurrentB = S.Id(ID_BUTTON_USECURRENTB).AddButton(XXO("Mea&sure selection"));
+         m_pButton_UseCurrentB = S.Id(ID_BUTTON_USECURRENTB).AddButton(XXO("&Measure selection"));
          mBackgroundRMSText = S.Id(ID_BACKGROUNDDB_TEXT)
             .ConnectRoot(wxEVT_KEY_DOWN,
                          &ContrastDialog::OnChar)
--- a/src/effects/DtmfGen.cpp
+++ b/src/effects/DtmfGen.cpp
@@ -351,7 +351,7 @@ void EffectDtmf::Validator::PopulateOrExchange(ShuttleGui & S,
             vldDtmf.SetIncludes(wxArrayString(WXSIZEOF(kSymbols), kSymbols));
             return vldDtmf;
          })
-         .AddTextBox(XXO("DTMF &sequence:"), wxT(""), 10);
+         .AddTextBox(XXO("DT&MF sequence:"), wxT(""), 10);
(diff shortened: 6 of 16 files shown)


crsib (commented):
I can see lot's of changes to the old strings. Is there a reason to use `s` for `Presets & Settings`?

Please keep in mind that any change to XO/XXO string means that it needs to be retranslated, potentially it will cause translations to be missing for not-so-actively updated languages

LWinterberg:
the previous access key was m, which no longer appears in the new strings (unless we do it like the japanese localization and do `"Presets && settings (&M)`), so a new letter needed to be chosen. P is taken by preview, and all other letters have collisions with access keys inside the effects. Technically, `I` has the fewest collisions, but access keys are to be remembered after all, so I chose the start of the second word - "alt+s for settings".

It is possible to allow the collision anyway, ie to put the access key on s while keeping the other strings on s, too. If this is chosen, it would work as thus: 

1. open "change pitch"
2. press alt+s
3. the focus jumps to the semitone controls
4. press alt+s again
5. presets and settings open
6. close menu
7. press alt+s again
8. presets and settings open

Generally, in case of collision, if there is a matching access key inside the form (ie the effect settings) after the current focus, the access key jumps to it. If the focus is on or after the last instance of a matching access key, it presses the button instead, but the  buttons generally give back the focus to where it was previously. Note that you don't really get trapped here, as tab navigation (or using other access keys) gets you to where you want to be anyway.

(Aside: Import/export remove the focus altogether, making keyboard navigation impossible afaict. I'll make a separate bug on that later)

----

Any of these options are easy enough to implement/change the implementation to it. 

* Keep the existing access key and rename the Presets and settings button to "Presets & Settings (M)"
* Use S as the access key and resolve collisions (this PR)
* Use I as the access key and resolve collisions (this would require 4 additional string changes, compared to 15 here)
* Use [S,I, or any of the other letters found in presets & settings] as the access key but allow the collisions (this would be just a single string change, inside the button itself)

After this investigation I think I prefer the latter option, S as the access key while allowing collisions. What do you think? @crsib @DavidBailes @Tantacrul

DavidBailes:
On Tue, 6 Sept 2022 at 10:05, Dmitry Vedenko ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
>
> I can see lot's of changes to the old strings. Is there a reason to use s
> for Presets & Settings?
>
> Please keep in mind that any change to XO/XXO string means that it needs
> to be retranslated, potentially it will cause translations to be missing
> for not-so-actively updated languages
>
Why can't two strings which differ only in the position of an & be treated
as being the same for the purposes of whether a new translation is needed?

DavidBailes:
On Tue, 6 Sept 2022 at 11:26, LWinterberg ***@***.***> wrote:

> the previous access key was m, which no longer appears in the new strings
> (unless we do it like the japanese localization and do "Presets &&
> settings (&M)), so a new letter needed to be chosen. P is taken by
> preview, and all other letters have collisions with access keys inside the
> effects. Technically, I has the fewest collisions, but access keys are to
> be remembered after all, so I chose the start of the second word - "alt+s
> for settings".
>
> It is possible to allow the collision anyway, ie to put the access key on
> s while keeping the other strings on s, too. If this is chosen, it would
> work as thus:
>
>    1. open "change pitch"
>    2. press alt+s
>    3. the focus jumps to the semitone controls
>    4. press alt+s again
>    5. presets and settings open
>    6. close menu
>    7. press alt+s again
>    8. presets and settings open
>
> Generally, in case of collision, if there is a matching access key inside
> the form (ie the effect settings) after the current focus, the access key
> jumps to it. If the focus is on or after the last instance of a matching
> access key, it presses the button instead, but the buttons generally give
> back the focus to where it was previously. Note that you don't really get
> trapped here, as tab navigation (or using other access keys) gets you to
> where you want to be anyway.
>
> (Aside: Import/export remove the focus altogether, making keyboard
> navigation impossible afaict. I'll make a separate bug on that later)
> ------------------------------
>
> Any of these options are easy enough to implement/change the
> implementation to it.
>
>    - Keep the existing access key and rename the Presets and settings
>    button to "Presets & Settings (M)"
>    - Use S as the access key and resolve collisions (this PR)
>    - Use I as the access key and resolve collisions (this would require 4
>    additional string changes, compared to 15 here)
>    - Use [S,I, or any of the other letters found in presets & settings]
>    as the access key but allow the collisions (this would be just a single
>    string change, inside the button itself)
>
> After this investigation I think I prefer the latter option, S as the
> access key while allowing collisions. What do you think? @crsib
> <https://github.com/crsib> @DavidBailes <https://github.com/DavidBailes>
> @Tantacrul <https://github.com/Tantacrul>
>
Allowing collisions is my least favourite option - it significantly
degrades the usability of access keys when you have to press the relevant
keystroke a different number of times depending on the position of the
focus.

As for the other options, I'd prefer to be certain of whether new
translations are required when the access key is changed before choosing.

crsib:
> I'd prefer to be certain of whether new translations are required when the access key is changed before choosing.

I am quite certain that new translations will be required. @Paul-Licameli can you confirm it please?

DavidBailes:
On Tue, 6 Sept 2022 at 13:53, Dmitry Vedenko ***@***.***>
wrote:

> I'd prefer to be certain of whether new translations are required when the
> access key is changed before choosing.
>
> I am quite certain that new translations will be required. @Paul-Licameli
> <https://github.com/Paul-Licameli> can you confirm it please?
>

But I don't understand why!

crsib:
And even if it does - the translations likely need to change the accelerator key as well. 

On top of that, I think that gettext ignoring & in keys will break complex interfaces, so I don't think that they've considered such functionality.

DavidBailes:
On Tue, 6 Sept 2022 at 15:08, Dmitry Vedenko ***@***.***>
wrote:

> But I don't understand why!
>
> & is a part of msgid, which is essentially a key into the translation.
> Changing the location of & changes the key.
>
> A real world example of our translation file:
>
> #: src/prefs/KeyConfigPrefs.cpp
> msgid "&Tree"
> msgstr "&Arborescence"
>
> I do not see enough evidence yet that gettext can handle that only & has
> changed
>
thanks for the explanation, I hadn't realized that this was handled by
third party code, I thought that it might be something that could be
tweaked by us.

DavidBailes:
On Tue, 6 Sept 2022 at 15:12, Dmitry Vedenko ***@***.***>
wrote:

> And even if it does - the translations likely need to change the
> accelerator key as well.
>
but not necessarily for the label for that particular control. If the
"Manage" button is changed to a "Presets & settings" button, then in a
translation, depending on the translations of "Manage" and "Presets &
settings", the access key for that button may or may not have to be
changed. If it's a new access key, then whether it creates a duplicate
depends on the access keys assigned to the translated strings of the other
controls in the dialog. So the correction of duplicates in the translation
doesn't, in general,  involve having to change the access keys in the names
of the same controls that have to be changed in the original english.

For the "not-so-actively updated languages", then if "Manage" is changed to
"Presets & settings", then, as I understand it, the name of this important
control which appears in all effect dialogs may go untranslated, so there's
a problem regardless of the translation problem which you have raised. So
the translation is going to have to be updated anyway, if it's going to
remain usable.

On top of that, I think that gettext ignoring & in keys will break complex

Tantacrul:
Hmm. This is a bit of a pickle with not much time at our disposal (and not much time for translations) - but loads of time after we get the release out. 

Since the Presets and Settings button can already be accessed, we should hold off setting an access key now, which will simplify things to allow us to focus on our remaining audio issues.

I think we bump this to 3.2.x

DavidBailes:
On Wed, 7 Sept 2022 at 10:51, Tantacrul ***@***.***> wrote:

> Hmm. This is a bit of a pickle with not much time at our disposal (and not
> much time for translations) - but loads of time after we get the release
> out.
>
> Since the Presets and Settings button can already be accessed, we should
> hold off setting an access key now, which will simplify things to allow us
> to focus on our remaining audio issues.
>
> I think we bump this to 3.2.x
>
An alternative would be to keep the name of the button as "Manage" until
3.2.x.  Then, no translations of "Presets and Settings" are needed until
3.2.x

Tantacrul:
Yeah. That makes sense. I'm reticent since the general UX benefit of calling it 'Presets and Settings' is high, since it is much more descriptive and quite important for real-time FX.

I feel it slightly outweighs the access key issue. Only because there is already a way to access the button.

Paul-Licameli:
> > But I don't understand why!
> 
> `&` is a part of `msgid`, which is essentially a key into the translation. Changing the location of `&` changes the key.
> 
> A real world example of our translation file:
> 
> ```
> #: src/prefs/KeyConfigPrefs.cpp
> msgid "&Tree"
> msgstr "&Arborescence"
> ```
> 
> I do not see enough evidence yet that gettext can handle that only `&` location has changed

What @crsib says about get text is correct

But the translation update can help the translator at least by detecting near-matching msgIds, keeping the old translations in comments, and suggesting them for the new strings.

Issue #3310, opened 2022-07-29
Real statelessness for VST2


Paul-Licameli:
Some steps (and analogous commits)

- [ ] Really stateless (squash following into one commit when done) (014836dd)
  - [x] `unordered_map` in Settings may need to become `map` for predictable iteration order in CopySettingsContents
  - [x] Define appropriate `MakeSettings`, and a `CopySettingsContents` that does not allocate memory
  - [x] `RealtimeProcessStart` stores settings into Instances
  - [x] `mSettings` member of Effect is removed, and `GetSettings` is rewritten as a static function really using its argument

PR #3541, opened 2022-09-05
Vst2 real statelessness

pietro68:
Resolves: https://github.com/audacity/audacity/issues/3310

Recommended:


Paul-Licameli:
This is failing to build on CI

pietro68:
> Should I consider this a draft? It's not finished until `VST::mSettings` and the mutex are eliminated.

Oops this is really weird! I am sure I deleted them and it built at least on Windows - something must have gone wrong while rebasing. I will turn this to draft, fix, and reopen it

src/effects/VST/VSTEffect.cpp
```diff
@@ -1193,6 +1187,48 @@ bool VSTEffect::SaveSettings(const EffectSettings& settings, CommandParameters&
...
+   for (auto& [key, value] : srcMap)
+   {
+      assert(dstIter != dstEnd);
+      auto& [dstKey, dstValue] = *dstIter;
+      assert(dstKey == key);
+
+      dstValue = value;
+
+      dstIter++;
+   }
```
Paul-Licameli, line 1227: One more assertion after the loop, that dstIter is the end iter, would complete the assertion of correspondence of keys

Paul-Licameli:
A few things to say:

First have you tested output monitoring?  If it doesn't work, maybe a commit analogous to b5e1872220042ffc5d676982805b4cb8082acb8a will do it:  don't make a slave the first time, but use self.

Secondly, your testing must have shown that copying of the chunk is necessary for correct working.  Very well:  but then it surely is doing free store allocation.  This might cause drop-outs in playing on a slow machine.  And maybe it never will on your machine.

I say leave that efficienty problem un-fixed.  I may propose changes to fix it.

src/effects/VST/VSTEffect.cpp
pietro68: ok, done

pietro68:
The last two commits introduced a bug - to reproduce it, have a vst2 plugin in the stack, press play, move a slider, press stop and then press play again - this happens:

image

Paul-Licameli:
Omit my commits then; fix metering.

When metering works, return to them perhaps and debug further.

Paul-Licameli:
You know I don't like this implementation.  Revert it and attempt this instead, see if it works:

VSTEffectWrapper has a base class VSTEffectUIWrapper and VSTEffectValidator inherits that base, only.  Move just enough members of VSTEffectWrapper down into that base so VSTEffectValidator recompiles.

But do NOT move mAEffect !  Where the other Validator methods now use information from mAEffect, try to pass that instead into the constructor of the Validator, taking values from instance.

Validator does not need to Load() or Unload() then.  It doesn't need StoreSettings().

Now, the Instance class can hold a pointer to a VSTEffectUIWrapper -- not to the more complete Validator class, so that breaks dependency cycles.  Its NeedIdle, SizeWindow, Automate overrides delegate to that wrapper, if the pointer is not null, else they do nothing.

Instance exposes a method to set its pointer.  Validator uses that method, setting itself at the pointer when constructed, and clearing that pointer when it is destroyed.

Do you understand?

One less mAEffect must be created.  The same mAEffect object must do all the work in the various branches of AudioMaster.

This eliminates the waste of multiple mAEffect objects and of the exta buffers, extra copies, and extra callSetReplacing.

PR #3542, opened 2022-09-05
Separate nyquist prompt

Paul-Licameli:
Transformation of NyquistEffect so that the special prompt effect is a subclass

Recommended:


Paul-Licameli:
@rbdannenberg This pull request begins the factoring of Nyquist.cpp into smaller pieces, far above the level of the Nyquist library itsefl.  I don't need your detailed review but you may be curious to follow this part of the rewriting too.

src/effects/nyquist/Nyquist.h
```diff
@@ -25,6 +25,11 @@ class wxTextCtrl;
 
 #define NYQUISTEFFECTS_VERSION wxT("1.0.0.0")
 
+// Protect Nyquist from selections greater than 2^31 samples (bug 439)
+#define NYQ_MAX_LEN (std::numeric_limits<long>::max())
```
SteveDaulton, line 29: As I said 11 days ago, I don't agree with perpetuating [known bugs with known fixes](https://github.com/audacity/audacity/issues/1607) into new code.

Paul-Licameli: Steve, I will try to give that long neglected issue some attention.  But meantime, this proposed pull request addresses other problems, and is cutting and pasting the existing code in a way that does not make this issue worse.  Not better, but no worse.

So please consider this pull request on its own and be convinced that it is not causing any regressions.

Issue #3544, opened 2022-09-05
Debug button of Nyquist Prompt does nothing


Paul-Licameli:
### Steps to reproduce

As reported by @SteveDaulton 

Make a selection and enter this code into the Nyquist Prompt:
(format t "Hello World")
(hzosc 440)
Then click the Debug button.
The Debug window does not open. 

### Expected behavior

Debug window opens

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

PR #3545, opened 2022-09-05
Fix Debug button of Nyquist workbench...

Paul-Licameli:
... Commit b221829 was at fault.

Do not change the ids associated with the buttons.  Other parts of the code
rely on those special values.

Resolves: #3544 

Recommended:

### DIFF
--- a/src/effects/EffectUI.cpp
+++ b/src/effects/EffectUI.cpp
@@ -127,8 +127,6 @@ static const int kDeletePresetDummyID = 20007;
 static const int kMenuID = 20100;
 static const int kEnableID = 20101;
 static const int kPlayID = 20102;
-static const int kApplyID = 20103;
-static const int kDebugID = 20104;
 static const int kPlaybackID = 20105;
 static const int kCaptureID = 20106;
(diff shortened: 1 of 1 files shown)


crsib (approved):

SteveDaulton:
There's a subtle change in behaviour from previous versions.

Previously, clicking the Debug button in the Nyquist Prompt would open the debug window after running Nyquist code in the Nyquist Prompt, and clicking the Debug button in a "Nyquist  Worker" (a GUI plug-in launched from the Nyquist Prompt) would open the debug window after running the code in the Nyquist Worker effect.

With this patch, the Debug button in the Nyquist Worker effect has no effect. To see the Debug window from a Nyquist Worker effect it is now necessary to click the Debug button in the Nyquist Prompt.

This "fix" is certainly an improvement, though it seems less intuitive than the old behaviour, and it's undocumented. I'd expect the Debug button in the Nyquist Worker effect to control the debug window for the Nyquist Worker effect.

(Tested on Xubuntu 20.04, with https://github.com/audacity/audacity/pull/3560 also applied.)

PR #3552, opened 2022-09-06
Effect latencies

Paul-Licameli:
Resolves: #3221 

Recommended:

### DIFF
--- a/src/AudioIO.cpp
+++ b/src/AudioIO.cpp
@@ -1795,21 +1795,21 @@ void AudioIO::FillPlayBuffers()
    if (mNumPlaybackChannels == 0)
       return;
 
-   // Though extremely unlikely, it is possible that some buffers
-   // will have more samples available than others.  This could happen
-   // if we hit this code during the PortAudio callback.  To keep
-   // things simple, we only write as much data as is vacant in
+   // It is possible that some buffers will have more samples available than
+   // others.  This could happen if we hit this code during the PortAudio
+   // callback.  Also, if in a previous pass, unequal numbers of samples were
+   // discarded from ring buffers for differing latencies.
+
+   // To keep things simple, we write no more data than is vacant in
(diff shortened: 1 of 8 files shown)


src/effects/RealtimeEffectManager.cpp
```diff
@@ -129,7 +129,7 @@ size_t RealtimeEffectManager::Process(bool suspended, Track &track,
    // Can be suspended because of the audio stream being paused or because effects
```
pietro68, line 129: So the meaning of the return value has changed; but it is difficult for me to understand what it is now. Number of processed samples? a comment in the prototype would help

src/AudioIO.cpp
```diff
@@ -1795,22 +1795,22 @@ void AudioIO::FillPlayBuffers()
...
+   // It is possible that some buffers will have more samples available than
+   // others.  This could happen if we hit this code during the PortAudio
+   // callback.  Also, if in a previous pass, unequal numbers of samples were
+   // discarded from ring buffers for differing latencies.
+
+   // To keep things simple, we write no more data than is vacant in
    // ALL buffers, and advance the global time by that much.
    auto nAvailable = GetCommonlyFreePlayback();
 
    // Don't fill the buffers at all unless we can do the
```
pietro68, line 1807: maybe "the" should be deleted?

src/effects/RealtimeEffectManager.cpp
Paul-Licameli: Exactly.  The comments in header files are updated.

The sum of latencies of the effects must be discarded but the state object is responsible to report the count, only, not do the discarding.

pietro68 (approved):
I can not honestly say that I fully understood the changes, but I get a sense of them - I do not know this part of the code well enough. Anyway, I verified that the fix works.

Issue #3543, opened 2022-09-05
No Visible Scrollbar in Plugin Manager


hellocatfood:
### Bug description

I'm using the Audacity-3.2.0-alpha-1 appimage on Ubuntu 22.04 using the Adwaita theme with a screen resolution of 3840x2400. When I go to Effect > Plugin Manager a long list of plugins appear but there is no visible scrollbar.

![Screenshot from 2022-09-05 17-33-05](https://user-images.githubusercontent.com/151947/188490436-fb415d17-e5fb-49c3-abca-436d752767c9.png)

However, I can scroll through the plugins by dragging the small area between the plugin list and the Plugin Manager Window

https://user-images.githubusercontent.com/151947/188493557-11ede2bd-61c2-47a4-96f9-75668e0edd51.mp4

### Steps to reproduce

1. Go to Effect > Plugin Manager
2. Try scrolling the window

### Expected behavior

Being able to scroll the window using a visible scrollbar

### Audacity Version

current master / alpha / nightly build

### Operating system

Linux

LWinterberg:
Do you have scrollbars elsewhere? because - 

* #2937 
* #2222

hellocatfood:
> (also, for completeness: cannot repro on windows)

Just tried on a clean Ubuntu 22.04 install via a VM and still getting no scrollbars

![Screenshot from 2022-09-06 11-25-41](https://user-images.githubusercontent.com/151947/188679313-14904166-a46a-4d14-9a1b-0d950af975b2.png)

Also yes no scrollbars elsewhere either

![Screenshot from 2022-09-06 11-29-52](https://user-images.githubusercontent.com/151947/188679389-88353509-8e9f-4655-81a3-a816f91a282a.png)

In the above screenshot both bits of audio extend past the visible area

LWinterberg:
Fixed by #3571

Issue #2937, opened 2022-05-16
Linux: Missing horizontal scrollbar


SteveDaulton:
**Describe the bug**
The horizontal scrollbar is missing

**To Reproduce**
1. Generate a 30 second tone
2. Zoom in

**Actual behavior**
Although part of the project is off-screen, there is no horizontal scrollbar

**Expected behavior**
A horizontal scrollbar to appear at the bottom of the tracks window when part of the project is off-screen.

**Screenshots**
If applicable, add screenshots/video to help explain your problem.

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version Audacity 3.2.0 2b5fa3

**Additional context**
Regression against Audacity 3.2.0 deaa3f

petersampsonaudacity:
Testing on W10 with audacity-win-3.2.0-alpha-20220513+2b5fa3c-x64-msvc2022

This works fine on Windows:
![image](https://user-images.githubusercontent.com/50205984/168580781-9cd22cbd-7ae4-42d7-a9ca-4ffd84ce1b55.png)

petersampsonaudacity:
Testing on macOS 12.3.1 Monterey with 
audacity-macOS-3.2.0-alpha-20220513+2b5fa3c-x86_64 

Works fine on Mac too.

LWinterberg:
duplicate of #2222

SteveDaulton:
> duplicate of #2222

It's the same symptoms, but not an exact duplicate, because the version of Audacity in #2222 worked correctly on Xubuntu 20.04.

Same machine, same OS, two different versions of Audacity - one works and the other doesn't.

wchouser3:
I can confirm - the horizontal scrollbar doesn't appear on the Flatpak, the Appimage or if you compile from source (Linux, that is) 
However, if you unmaximize, then re-maximize the scrollbar appears. Explain that one. I'm using Linux Mint 21, which is using the forked version of Mutter - under the X protocol. perhaps that has something to do with it.

spicy-bah:
I found a workaround for this. In preferences/interface, check *show scrub ruler* under *timeline*. That adds a few extra lines of pixels to the timeline, but now the horizontal scrollbar will show each time I open Audacity. 
Tested this with version 3.2.4 on Q4OS (Debian) and version 3.2.3 on KDE Neon, using the AppImage on both systems.

Accidentaccio:
[...]

Issue #2624, opened 2022-03-04
Text on Input and output drop-down lists go out of frames


Charlie-Ramirez-Animation-Studios-de-MX:
**Describe the bug**
in the audacity 3.1.3 GUI the text in the drop down windows for server selection, audio input and output, pops out of the boxes or hides/overlaps when trying to resize making it impossible to move them out of position unless they expand up to a very large size and there is another free area of ​​the same size
**To Reproduce**
1. On a Debian 11 Computer with a KDE Plasma desktop
2.  Download or install audacity 1.3.1 in any of its available versions [Flatpak and/or Appimage package].

3. Open the app.
You will notice that the dropdown lists one of them is not visible and the others are truncated.

4. If you click on them they continue to work, but it is complicated to use since they tend to overlap one list with another [Trying to move them from position is impossible, trying it will open the audio servers tab (due to the overlap)]
5. Try changing the size to a very large one The lists should be moderately aligned or accommodated enough to be able to see the text and drag its position

**Expected behavior**
Texts and click areas should be kept "centered" (or at least visible text with left alignment) and auto-fit-hide to bounds based on the size of the dropdowns as in the case of previous versions

**Screenshots**
### For Flatpak:
![image](https://user-images.githubusercontent.com/64709446/156678183-7e88c06c-2737-4db6-9730-403b0fb0c73a.png)
![image](https://user-images.githubusercontent.com/64709446/156678438-6b5be48b-c89c-4f37-b17a-b3593d0df986.png)
### On Appimage [The problem is more tolerable [does not cause overlapping on click] however the dropdown icons have disappeared]
![image](https://user-images.githubusercontent.com/64709446/156678968-dc4cd712-f7a4-4a9f-b586-6f067682984a.png)
![image](https://user-images.githubusercontent.com/64709446/156679144-7b3a72d0-6a0f-46da-b784-598b7ee0502f.png)

**Additional information (please complete the following information):**
 - OS: Debian 11 x64 (KDE Plasma Desktop) -Testing Packages Enabled
 ### -Flatpak
 - Version [Audacity 3.1.1 - Tested on Flatpak and Appimage Ver.]
 Fecha de compilación: Jan 17 2022 Commit Id: ff6d67 of Wed Dec 22 16:35:36 2021 +0100
Tipo de compilación: CMake Compilación de distribución (nivel de debug 1), 64 bits
Compilador: GCC 11.2.0
[...]

Issue #2279, opened 2021-12-13
Icons missing from dialog windows


SteveDaulton:
**Describe the bug**
Icons (such as "Alert", "Error",...) are missing from pop-up message dialog screens

**To Reproduce**
1. Select a mono track
2. "Effect menu > Auto Duck"
3. Observe the generic icon placeholder  in the pop-up error message.

**Expected behavior**
The correct icon to be displayed

**Screenshots**

![error-message](https://user-images.githubusercontent.com/11681097/145820962-2b857b56-3b6f-41dd-b972-e35d96a79a31.png)

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version: 3.2.0-alpha-20211210

**Additional context**
Regression.

SteveDaulton:
May be related to https://github.com/audacity/audacity/issues/2270

LWinterberg:
I can reproduce this with 3.1.3 on Ubuntu 21.10

nadrolinux:
For me both icons and widgets like a check boxes, radio buttons, sliders, standard buttons etc. are missing. As you can see at attached screenshot I can't even see which options are enabled. I tried different Ubuntu and Audacity themes, but nothing helped. My platform: Audacity 3.1.3 (AppImage) + Ubuntu 21.10. I had the same problem with Audacity 3.1.2 (AppImage). Flatpak version looks good, however Flatpak version has problem with Pipewire, so at now all versions are broken for me.

![audacity_broken_widgets](https://user-images.githubusercontent.com/11942352/147294619-793c5736-6a4a-494c-b718-ef369d703aa8.png)

LunaGNUisance:
I've tried the Audacity 3.1.3 Appimage on Arch Linux/GNOME 41.2/X11 with the same results as @nadrolinux . The lack of a visible scrollbar makes it very hard to use. 
![appimage-widgets-missing-1](https://user-images.githubusercontent.com/78894455/148326787-c41f6fbc-bf51-4a77-ad7d-d7606db1700d.png)

I tried a few AppImages and the issue was present across all of them. I didn't bother with the Flatpak version.

I then tried to build it using [this AUR package,](https://aur.archlinux.org/packages/audacity-gtk4/) which failed early due to a conan issue.

[...]

PR #3571, opened 2022-09-07
Fixes GTK packaging on Linux

crsib:

Recommended:

### DIFF
--- a/linux/create_appimage.sh
+++ b/linux/create_appimage.sh
@@ -69,6 +69,8 @@ if create_path "linuxdeploy"; then
 (
     cd "linuxdeploy"
     download_linuxdeploy_component linuxdeploy continuous
+    wget -q https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh
+    chmod +x linuxdeploy-plugin-gtk.sh
 )
 fi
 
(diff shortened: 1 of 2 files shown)


LWinterberg:
While this fix seems to not apply the GTK theme properly (maybe it's some sort of fallback/default? Certainly isn't adwita-dark), Testing across the board of linux UI bugs, 

* #2279 is fixed
image
* #3543 is fixed
image
* #2222 and
* #2937 aren't really fixed: The scrollbar functionality appears invisibly as soon as its needed, but the visuals only update when the window is resized. This also applies the other way round, if you have scrollbars and then delete your tracks, white bars will stay around the edge where the scrollbars used to be.
* #2270 is fixed. While there is a single icon shown as an x, I believe this to be a case where my icon pack doesn't specify that icon, rather than something Audacity related. 
image
* #2624 is mostly fixed
image
* #1800 and 
* #1456 aren't really fixed - even at 100%, text gets cut off ever so slightly, compared to other dropdowns. At 200% it completely runs off (see below)
image
* #1709 is fixed for me.
* #503 does not change (not that I expected that it would). Here's an example of Audacity inside a properly behaved app. 
image
* #1536 appears fixed at 100%, at least I don't get the bugs listed inside it as examples ("some elements are malformed, notably the Device Toolbar appears to be empty, and text in the track's control panel is tiny.")
* #1609 is fixed (probably: never was broken) on 100%
image


I think that's all of them. I don't know how many of them should be marked as resolved by this PR :)

crsib:
Whoa, that is quite a lot of feedback, thank you!

crsib:
I thin #1536 is about GTK3 build, we are still building GTK2. So not fixed, I think

vsverchinsky (approved):

Issue #3576, opened 2022-09-08
The rescan button re-enables user-disabled plugins


LWinterberg:
Testing with the alpha-3.2.0 branch, the "Rescan" button does not only rescan for plug-ins, it also automatically enables every plug-in that it finds on my computer, thus re-enabling the hundreds of plug-ins that I've previously spent time disabling.

Not only that, but after pressing "Rescan", when I see that it wants to enable hundreds of plug-ins that I don't want enabled, it then fails to honour the "Cancel" button:
1. Ensure that you have some custom effects installed on your computer.
2. Launch Audacity
3. Disable all custom effects
4. Close and restart Audacity
5. Check and see that you have only the standard set of plug-ins.
6. Launch "Manage Plugins"
7. Click the "Rescan" button
8. (optional) Curse
9. Press the "Cancel" button
10. Observe that the custom effects are NOT listed in the menus
11. Quit Audacity and relaunch
12. Observe that the custom effects ARE now listed in the menus.

Now I have to go through the effects again and disable all of those that I disabled previously.
I seriously hope that this is a bug and not the intended behaviour.

_Originally posted by @SteveDaulton in https://github.com/audacity/audacity/issues/2967#issuecomment-1240726550_

----

Expected behavior:
* Plugins which exist on-disk but were not previously in the list get enabled.
  This should also apply to Nyquist effects.
* Plugins which are in the list but are disabled don't get touched
* Plugins which are in the list but no longer exist on-disk get removed

vsverchinsky:
@LWinterberg could you please clarify what should happen if user decides to cancel plugin scanning?

LWinterberg:
> Right now if I cancel it I would see only those that were found so far, and when I rerun Audacity it starts scanning from where it was canceled.

this sounds correct

> Also what about "Incompatible plugin(s)" dialog?

Good question, let me think

SteveDaulton:
Regarding cancelling things, Paul previously did a lot of work on making cancel work the same way for nearly everything. Cancel stops the current operation and winds back to the previous state (so you are back to where you were before starting the thing that was cancelled). Imo this is what most users expect "Cancel" to do. Perhaps worth having a word PRL about this.

LWinterberg:
> Cancel stops the current operation and winds back to the previous state (so you are back to where you were before starting the thing that was cancelled). Imo this is what most users expect "Cancel" to do.

[...]

PR #3586, opened 2022-09-09
Ensured that disabled plugins are still disabled after a manual rescan.

ksoze95:
Resolves: #3576

Recommended:

### DIFF
--- a/src/PluginRegistrationDialog.cpp
+++ b/src/PluginRegistrationDialog.cpp
@@ -874,7 +874,20 @@ void PluginRegistrationDialog::OnRescan(wxCommandEvent& WXUNUSED(evt))
    mEffects->Update();
 
    wxTheApp->CallAfter([this] {
+      PluginPaths disabledPlugins;
       auto& pm = PluginManager::Get();
+
+      // Record list of plugins that are currently disabled
+      for (auto& plug : pm.AllPlugins())
+      {
+         PluginType plugType = plug.GetPluginType();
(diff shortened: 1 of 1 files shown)


src/PluginRegistrationDialog.cpp
```diff
@@ -615,6 +615,9 @@ void PluginRegistrationDialog::PopulateItemsList(PluginManager& pm)
          continue;
 
       const auto& path = plug.GetPath();
+      if (make_iterator_range(previouslyDisabledPlugins).contains(path))
```
vsverchinsky, line 618: Suggest moving this logic to lambda, and do not change function signature

vsverchinsky (approved):

src/PluginRegistrationDialog.cpp
```diff
@@ -884,6 +897,20 @@ void PluginRegistrationDialog::OnRescan(wxCommandEvent& WXUNUSED(evt))
...
 
+      // Disable all plugins which were previously disabled
+      for (auto& plug : pm.AllPlugins())
+      {
+         PluginType plugType = plug.GetPluginType();
+         if (plugType != PluginTypeEffect && plugType != PluginTypeStub)
+            continue;
+
+         const auto& path = plug.GetPath();
+         if (make_iterator_range(disabledPlugins).contains(path))
```
Paul-Licameli, line 908: Correct but maybe ineffecient -- I would have looked up in a hash map instead of doing a linear scan of a vector.

Issue #3565, opened 2022-09-07
[LADSPA] GVerb crashes Mix&Render when stacked on a stereo track


dozzzzer:
### Plugin

GVerb

### Plugin Format

LADSPA

### Steps to reproduce

1. load some stereo track
2. in the sidebar select GVerb plugin
3. click Share Audio
4. Click Continue

### Expected behavior

Upload started

### Actual behavior

Crash: https://sentry.audacityteam.org/share/issue/cd2edceef5f94944ae9c99ce5f577bc5/

### Audacity Version

current master / alpha / nightly build

### Operating system

Any

----------------

It appears, that GVerb declares a single input and two outputs. This confuses Mix&Render..

crsib:
I appears that stereo track is required

crsib:
And it really affects Mix&Render/Export, not the sharing feature itself. Output format is irrelevant.

Ryan-Miller-Audacity:
Tested and verified fixed in latest alpha branch.

PR #3587, opened 2022-09-09
Issue 3565: GVerb crashes in effect stack of stereo track...

Paul-Licameli:
... This simple fix corrects the crash.  Whether the resulting output is correct is a matter of definition.

GVerb takes one channel in and always writes two channels out.  When rendering the right channel then, supply an extra throw-away buffer.

What happens in the result is that the left output, only, of each channel is retained.

What really should be done?  It might make sense to mix the two pairs of output channels down to one pair, losing none of the output.

That is outside the scope of this simple fix.

Resolves: #3565

Recommended:

### DIFF
--- a/libraries/lib-sample-track/Mix.cpp
+++ b/libraries/lib-sample-track/Mix.cpp
@@ -94,7 +94,9 @@ Mixer::Mixer(Inputs inputs,
 
    // PRL:  Bug2536: see other comments below for the last, padding argument
    // TODO: more-than-two-channels
-   , mFloatBuffers{ 2, mBufferSize, 1, 1 }
+   // Issue 3565 workaround:  allocate one extra buffer when applying a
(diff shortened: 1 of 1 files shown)


libraries/lib-sample-track/Mix.cpp
```diff
@@ -238,7 +240,7 @@ size_t Mixer::Process(const size_t maxToProcess)
 
    Clear();
    // TODO: more-than-two-channels
-   auto maxChannels = mFloatBuffers.Channels();
+   auto maxChannels = std::max(2u, mFloatBuffers.Channels());
```
crsib, line 243: Probably it makes sense to have `maxChannels` as constexpr somewhere, and reuse it in both places?

crsib (approved):
Good enough as a quick fix, but definitely not good enough in the long run.

Issue #3580, opened 2022-09-08
AutoDuck effect can be opened in a single track project


SteveDaulton:
### Bug description

The AutoDuck effect requires that the project contains at least two audio tracks, and that a non-selected track exists below the selected track(s). The correct behavior, as seen in previous versions of Audacity, is that when these conditions are not met, an error dialog is displayed and the effect does not open.

In the current 3.2.0-alpha, the error dialog is displayed,  but when that message is dismissed, the effect opens as if it can be applied.

### Steps to reproduce

1. Create a project with one audio track
2. Select the audio track
3. "Effect menu > AutoDuck"
4. Observe the error message:
`Auto Duck needs a control track which must be placed below the selected track(s).`
5. Click "OK"

### Expected behavior

The error dialog to close (nothing more, nothing less).

### Actual behavior

The error dialog closes, and the AutoDuck GUI opens

### Audacity Version

current master / alpha / nightly build

### Operating system

Linux

Ryan-Miller-Audacity:
This was tested and verified in latest Beta branch. 

PR #3589

PR #3589, opened 2022-09-09
When EffectInstance::Init() fails, don't proceed with effect dialog

Paul-Licameli:
Resolves: #3580

Recommended:

### DIFF
--- a/src/effects/EffectUI.h
+++ b/src/effects/EffectUI.h
@@ -66,6 +66,12 @@ public:
 
    bool HandleCommandKeystrokes() override;
 
+   void SetClosed() {
+#if wxDEBUG_LEVEL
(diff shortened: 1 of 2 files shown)


src/effects/EffectUI.h
```diff
@@ -66,6 +66,8 @@ class EffectUIHost final
 
    bool HandleCommandKeystrokes() override;
 
+   void SetClosed() { mClosed = true; }
```
crsib, line 69: mClosed is conditionally compiled under `#if wxDEBUG_LEVEL`

crsib (changes requested):
I expect that this PR will fail to compile. Please see mClosed definition.

Issue #3581, opened 2022-09-08
No changes to the sound/sound stuttering when adding some VST3/AU plugins in the sidebar during playback


dozzzzer:
### Bug description

The issue was observed on both VST3 and AU plugin families.

### Steps to reproduce

0. Install [this plugin](https://www.kvraudio.com/product/orilriver-by-denis-tihanov/downloads) or [this one](https://www.izotope.com/en/products/vinyl.html) or [even  OTT by Xfer](https://xferrecords.com/freeware)
1. Launch Audacity, load an audio track
2. Start playback
3. Open RT panel
4. Select the plugin from step 0
5. Observe
6. Open the plugin's GUI
7. Observe

### Expected behavior

The sound changes accordingly

### Actual behavior

The sound doesn't change at all; then the sound stutters (when opening GUI)

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

crsib:
https://user-images.githubusercontent.com/2660628/189369959-502f3ed2-a108-4feb-9901-a35809a31888.mp4

Potentially important note - if UI is opened when the track is silent - no sound will be produced at all while UI is open

Paul-Licameli:
At least in the case of Xfer OTT, the plug-in does not appear to implement bypassing, instead returning an error to Audacity.

This call is made when the instance initializes, and also when toggling power buttons.

dozzzzer:
Verified. The issue now seems to be fixed, however, I'm experiencing crash with iZotope Vinyl on macOS, for which I've logged a separate ticket: #3607

PR #3597, opened 2022-09-09
Ignore return value of AudioUnitInstance::BypassEffect...

Paul-Licameli:
... This allows Xfer OTT to work, and also to power on and off

Resolves: #3581 
(At least for Xfer OTT on macOS)

Recommended:


src/effects/audiounits/AudioUnitInstance.cpp
```diff
@@ -135,7 +135,8 @@ bool AudioUnitInstance::ProcessInitialize(EffectSettings &settings,
       return false;
 
    if (!BypassEffect(false))
```
crsib, line 137: Hmm, looks like if's can be removed, not critical

crsib (approved):

PR #3599, opened 2022-09-10
Additional folders added to .gitignore to enable easier building with VS 2022

witwald:
Resolves: The inability to build Audacity using Visual Studio 2022 when the build binaries are located in a folder outside of the audacity repository.

To build Audacity under Windows using Visual Studio 2022, the binaries need to be built inside a folder within the local audacity repository. Adding "/audacity-binaries" and "/.vs" to the .gitignore file is required to stop changes therein being monitored by Git. Here, "/audacity-binaries" is the target name of the folder for the binaries when setting up using CMake GUI.

Recommended:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,6 +140,8 @@ win/.vs/
 win/packages
 win/.nuget
 win-alt-build
+/audacity-binaries
+/.vs
 
 # All those help files
 help/manual*


crsib:
Is `audacity-binaries` a name that MSVC chooses? Sounds a bit too specific...

PR #3604, opened 2022-09-12
VST3 plugin initialization fix

vsverchinsky:
Resolves: #3581

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Wrapper.cpp
+++ b/src/effects/VST3/VST3Wrapper.cpp
@@ -499,12 +499,11 @@ bool VST3Wrapper::Initialize(EffectSettings& settings, Steinberg::Vst::SampleRat
 
    if(mEffectComponent->setActive(true) == kResultOk)
    {
-      if(mAudioProcessor->setProcessing(true) == kResultOk)
-      {
-         //...make sure they are delivered to the processor
-         ConsumeChanges(settings);
-         return true;
-      }
+      //Omit the return value check - some plugins return unexpected values (see bug #3581)
+      mAudioProcessor->setProcessing(true);
+      //...make sure they are delivered to the processor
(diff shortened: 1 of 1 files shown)


src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -499,12 +499,11 @@ bool VST3Wrapper::Initialize(EffectSettings& settings, Steinberg::Vst::SampleRat
...
    if(mEffectComponent->setActive(true) == kResultOk)
    {
-      if(mAudioProcessor->setProcessing(true) == kResultOk)
-      {
-         //...make sure they are delivered to the processor
-         ConsumeChanges(settings);
-         return true;
-      }
+      //Omit the return value check - some plugins return unexpected values (see bug #3581)
+      mAudioProcessor->setProcessing(true);
```
Paul-Licameli, line 503: Is this intended by analogy with 1b095c80960e994152f8b4538bec85ca4abf9001 to fix Xfer Ott and other effects?

Unlike in the AudioUnits case, there was only this one place where the return of `setProcessing` (like `Bypass()`) was checked.

Did you also make sure powering the effect on and off during play works as expected?

Paul-Licameli (approved):

crsib (approved):

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: >Did you also make sure powering the effect on and off during play works as expected?

Yes, it works.

>Is this intended by analogy with https://github.com/audacity/audacity/commit/1b095c80960e994152f8b4538bec85ca4abf9001 to fix Xfer Ott and other effects?

It worked before, so I did a bisecting

Issue #2482, opened 2022-01-26
Metadata Editor may have a minimum height greater than the display height


SteveDaulton:
**Describe the bug**
The Metadata Editor appears to have a minimum height set on open to fit the contents. If there is a lot of metadata, this may be greater than the display height

**To Reproduce**
1. Open the metadata editor
2. Add 20+ custom tags
3. Click the "Set Default" button
4. Close and relaunch Audacity
5. Open the Metadata Editor
6. Observe that the minimum height is very large.

**Expected behavior**
The Metadata Editor to always have a minimum height <= about 600 px so that the buttons at the bottom are always available.

**Screenshots**
![Screenshot_2022-01-26_16-05-48](https://user-images.githubusercontent.com/11681097/151200247-16116535-03c0-4630-a95c-0ba7ac3796ae.png)

**Additional information (please complete the following information):**
 - OS: Reported on Windows 10, confirmed on Xubumtu 20.04
 - Version: Audacity 3.x
 - (if relevant) Sound Device [e.g. Windows WASAPI, USB Sound card]

**Additional context**

Penikov:
Tested on Ubuntu 20.04
Looks good even at a low resolution like 1024x768

![image](https://user-images.githubusercontent.com/25851359/151335372-58ad6291-abcd-41bc-bda9-b0ab91007754.png)

petersampsonaudacity:
I tested this on W10 with latest master build audacity-win-3.2.0-alpha-20220127+4a69e09-x64-msvc2019
Which is the build that allegedly fixes this issue

I added 47 extra tags A-Z and 1-21

a) Observe: by default the bottom row of buttons (and the bottom right resizer are not visible

b) dragging the Metadata dialog as far up the screen as possible (use the header at the top of dialog)
Observe@ this reveals the second row of buttons but fails to show the triangle resizer


![image](https://user-images.githubusercontent.com/50205984/151546654-2398d85b-d9ab-49ad-8643-0cd2bfe0d9a4.png)

![image](https://user-images.githubusercontent.com/50205984/151547578-b242ca3d-5d38-4999-a975-d71a19862617.png)

@Penikov Pavel I therefore think this issue need to be reopened

SteveDaulton:
> a) Observe: by default the bottom row of buttons (and the bottom right resizer are not visible

The fix is based on the original design of showing as much data as possible.


> b) dragging the Metadata dialog as far up the screen as possible (use the header at the top of dialog)
[...]

PR #3609, opened 2022-09-13
metadata editor: set max height to prevent taskbar overlap

LWinterberg:
Resolves: #2482 

While #2484 set a reasonable min-height and makes the window resizable once it overlaps with the taskbar, it still would allow for the default window size to go all the way from the top to the taskbar. Here I shave off another 100px so that a user always has access to the entire thing without resizing (even though they still may need to drag it upwards to have it not overlap). 

NB: This intentionally targets master; I'm not sure how this handles HiDPI and dual screens with different sizes, and I don't want to hold up this release with a P4 fix

### DIFF
--- a/src/TagsEditor.cpp
+++ b/src/TagsEditor.cpp
@@ -243,7 +243,8 @@ TagsEditorDialog::TagsEditorDialog(wxWindow * parent,
    Fit();
    Center();
    wxSize sz = GetSize();
-   SetSizeHints(sz.x, std::min(sz.y, 600));
+   SetSizeHints(sz.x, std::min(sz.y, 600), 
+      wxDefaultCoord, wxDisplay().GetGeometry().GetHeight() - 100);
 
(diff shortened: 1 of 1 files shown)


src/TagsEditor.cpp
```diff
@@ -243,7 +243,7 @@ TagsEditorDialog::TagsEditorDialog(wxWindow * parent,
    Fit();
    Center();
    wxSize sz = GetSize();
-   SetSizeHints(sz.x, std::min(sz.y, 600));
+   SetSizeHints(sz.x, std::min(sz.y, 600),-1,wxDisplay().GetGeometry().GetHeight()-100);
```
Paul-Licameli, line 246: OK for its effects

Stylistically, I'd prefer the name `wxDefaultCoord` to the magic number -1, spaces after each comma and around binray `-`, and a line break to keep it in 80 columns.

Paul-Licameli (approved):

Issue #3485, opened 2022-08-30
Crash when hitting Apply during preview of a destructive effect (LV2/VST3)


dozzzzer:
### Steps to reproduce

1. Install an LV2/VST3 plugin (should be realtime-capable)
2. Load some audio
3. Go to Effect menu > select that effect
4. Click **Preview**
5. Click **Apply**

### Expected behavior

There's a popup with the progress bar displayed preventing the user from clicking outside. Thus, clicking Apply during preview isn't possible at all.

### Actual behavior

Crash: https://sentry.audacityteam.org/share/issue/04cf9671c1ed4368ba96f71ae7faabc4/

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

Ryan-Miller-Audacity:
Tested on both Windows and MAC with VST3, AU, LV2 plugin families and there is no longer a crash when destructively applying an effect during preview playback.

PR #3613, opened 2022-09-13
Fixes Apply button during the preview

crsib:
Resolves: #3485

Recommended:

### DIFF
--- a/src/effects/EffectUI.cpp
+++ b/src/effects/EffectUI.cpp
@@ -501,6 +501,9 @@ void EffectUIHost::OnPaint(wxPaintEvent & WXUNUSED(evt))
 
 void EffectUIHost::OnClose(wxCloseEvent & WXUNUSED(evt))
 {
+   if (mPlaying)
+      StopPlayback();
+
    DoCancel();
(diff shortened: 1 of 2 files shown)


src/effects/EffectUI.cpp
```diff
@@ -501,6 +501,9 @@ void EffectUIHost::OnPaint(wxPaintEvent & WXUNUSED(evt))
 
 void EffectUIHost::OnClose(wxCloseEvent & WXUNUSED(evt))
 {
+   if (mPlaying)
```
ksoze95, line 504: Conditional check is not needed as StopPlayback returns early if mPlaying is not set.

src/effects/EffectUI.cpp
```diff
@@ -530,6 +533,9 @@ void EffectUIHost::OnApply(wxCommandEvent & evt)
    {
       return;
    }
+
+   if (mPlaying)
```
ksoze95, line 537: Same as above.  Conditional check is not needed.

ksoze95 (approved):

Issue #3627, opened 2022-09-15
Input & Playback metering toolbars: UI enhancements


dozzzzer:
### Your idea

- [x] make metering toolbars slightly longer
- [ ] vertically stack meters by default (so they stay stacked when widening the window)

### 1. Make the width of the metering toolbars slightly larger

The default width of those two meters is too narrow and only fits 12dB increments. Let's make it a tad wider so that the increment will become 6 dB:

<img width="317" alt="image" src="https://user-images.githubusercontent.com/70063970/190448242-4d62a0c4-3717-4737-a625-9968691fbb0b.png">

### 2. Keep the arrangement of both toolbars the same way as after config reset, i.e. one above the other

When resetting the config by **Tools > Reset Configuration**, the toolbars get placed one above the other. Double-click Audacity window to maximize - and you'll get them placed side by side:

<img width="631" alt="image" src="https://user-images.githubusercontent.com/70063970/190448737-732b7c0b-fed0-4506-84fa-0e1f90e72d1c.png">

Let's keep the placement the same way as it is initially, i.e. one above another: input meter on top, playback meter below.

### Problem to be solved

Oddly looking UI

### Additional context

Confirmed with @Tantacrul

petersampsonaudacity:
@Tantacrul A big +1 for @dozzzzer 's   point no. 1 regarding the default with of the meters.

Especially as we now have blank space in the upper tooldock to the right of the meters:
![image](https://user-images.githubusercontent.com/50205984/190578814-26e7b713-cee4-408d-84ab-873737e38193.png)

Tantacrul:
@petersampsonaudacity - yes, I've bumped it to this release.

dozzzzer:
Verified. The default size of the meters is now showing 6 dB increments.
Point no. 2 from the description is still to be fixed though.

petersampsonaudacity:
@LWinterberg @dozzzzer 

Surely this was fixed for 3.2 already - the current 3.2.11 at default settings shows:
![image](https://user-images.githubusercontent.com/50205984/196903554-7aee3f17-beab-42ea-bbdd-46d844942dfb.png)

LWinterberg:
part 2 - vertically stack meters by default (so they stay stacked when widening the window) - is not yet fixed. They're side-by-side and get stacked on the default windowed size, but unstack once you go fullsscreen. Goal is to have them be stacked even in fullscreen.

petersampsonaudacity:
[...]

PR #3629, opened 2022-09-16
tweak to default width of meter bar and project window

LWinterberg:
Resolves: #3627
part 1 of it anyway.

looks like this to me: 
image

### DIFF
--- a/src/ProjectWindow.cpp
+++ b/src/ProjectWindow.cpp
@@ -51,9 +51,9 @@ Paul Licameli split from AudacityProject.cpp
 namespace
 {
 #ifdef HAS_AUDIOCOM_UPLOAD
-   constexpr int DEFAULT_WINDOW_WIDTH = 1260;
+   constexpr int DEFAULT_WINDOW_WIDTH = 1120;
 #else
-   constexpr int DEFAULT_WINDOW_WIDTH = 1200;
+   constexpr int DEFAULT_WINDOW_WIDTH = 1060;
 #endif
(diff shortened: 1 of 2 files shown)


petersampsonaudacity:
@LWinterberg 

I think you have made the slightly too large Leo - they now pop down in a third row in the top tooldock in your branch build.

image

petersampsonaudacity:
Ahh, I note that you have also made the Audacity default window narrower (which I like).

a) on my PC with your branch build the window is 21.7 cm wide

b) on my PC with 3-2-0 Beta it is 3cm wider at 24.7 cm wide

So rather than reducing the new size of your meters maybe you just need to widen the default window by a small amount.

LWinterberg:
oh, interesting: the default window widths are 1120 and 1060 (depending on whether the share audio button is present), but the actual width is 1106 for the audio sharing button version. I wonder where we lose the 14px

petersampsonaudacity:
@LWinterberg  testing your latest branch - that's much better Leo - nice  :-))

image

petersampsonaudacity:
I assume this is gonna make it into 3.2.0 ?

petersampsonaudacity:
@LWinterberg @crsib @Tantacrul 

Will this get pulled into the 3.2.0 release - looks much better like this with Leo's changes?

crsib (approved):

petersampsonaudacity:
Testing on W10 with latest 2.3.0 build (today's)

Looking good now in the release build
image

Issue #1818, opened 2021-10-04
The slider setting value for AAC/MP4 export is not honored by Audacity - and the slider cannot make precise settings


petersampsonaudacity:
**Describe the bug**
a) The slider setting value for AAC/MP4 export is not honored by Audacity.

b) Plus it is very hard (perhaps impossble) to set precise values - I found it impossible to set it to the "standard" rates of 192kbps and 256kbps

**To Reproduce**
1. Get some stereo audio
2. File > Export > Export Audio
3. choose M4A(AAC) Files FFmpeg
4. Set the Quality slider to 320kbps
5. Click "Save" to make the export
6. Examine the file (on W10 I used Windows Explorer)
7. Observe the file is not 320kbps (mine was 465kbps YMMV)

**Expected behavior**
a)  Audacity should export at the rate the user selects
b) The slider should allow for more precise setting (making it wider may help_
c) the user should be able to set a precise kbps value by typing it in _(cf. MP3 export and mic/speaker sliders)_

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: tested on Windows 10 - but assume all
 - Version: Audacity 3.0.5 and 3.1.0

**Additional context**
1) Note that I am using the "recommended" FFmpeg the 64-bit Win downloaded from Buanzo's site
    This of course may change with an updated FFmpeg version ...

2) This does not happen with MP3 export in Audacity - the precise value is used for the exported file

3) Note that MP3 export uses a picking list and not a slider for the kbps setting - so we have an **_inconsistency_** issue here

4) The AACs that I have were made in iTunes,  
4.1) I export fro m Audacity as 16-bit WAV files
4.2) I import those WAV files into iTunes
4.3) I use iTunes to convert these WAVS to AAC files choosing 256kbps
4.4) I observe that iTunes converts these to 256kbps AAC files:
![image](https://user-images.githubusercontent.com/50205984/135894858-26143f29-2805-41b1-b0a5-12e6ef5c3f6a.png)

petersampsonaudacity:
This was reported today on the Forum:
https://forum.audacityteam.org/viewtopic.php?f=46&t=120701&p=436404

It looks to be the same issue as this one logged in Bugzilla, which I shall close in favor of this GitHub issue
https://bugzilla.audacityteam.org/show_bug.cgi?id=2161
Bug 2161 - Enh: Same quality setting and range used for mono and stereo M4A (AAC).

petersampsonaudacity:
[...]

PR #3634, opened 2022-09-16
Ffmpeg aac m4a export

ksoze95:
Resolves: #1818

Recommended:

### DIFF
--- a/src/export/ExportFFmpegDialogs.cpp
+++ b/src/export/ExportFFmpegDialogs.cpp
@@ -280,9 +280,9 @@ void ExportFFmpegAACOptions::PopulateOrExchange(ShuttleGui & S)
          S.StartMultiColumn(2, wxCENTER);
          {
             S.SetStretchyCol(1);
-            S.Prop(1).TieSlider(
+            S.TieSpinCtrl(
                XXO("Quality (kbps):"),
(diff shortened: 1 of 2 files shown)


ksoze95:
Replaced the slider with a spinctrl.  The control is constraint to values between 98-320 only via the up and down arrows.  The textbox is currently not validated yet, so you can enter a value outside this range.  However when you press save, it will constrain the desired target kps to the nearest value inside the valid range.

A later PR will be raised to have validation for the textbox as well.

src/export/ExportFFmpegDialogs.cpp
```diff
@@ -280,9 +280,9 @@ void ExportFFmpegAACOptions::PopulateOrExchange(ShuttleGui & S)
          S.StartMultiColumn(2, wxCENTER);
          {
             S.SetStretchyCol(1);
-            S.Prop(1).TieSlider(
-               XXO("Quality (kbps):"),
-               {wxT("/FileFormats/AACQuality"), 160},320, 98);
+            S.TieSpinCtrl(
+               XXO("Quality (kbps)"),
```
crsib, line 284: Is changing this line is necessary?

ksoze95: Nope.  Forgot to undo change.  Will fix it up now.

PR #3639, opened 2022-09-17
Plugin host process startup fix

vsverchinsky:

Recommended:

### DIFF
--- a/libraries/lib-module-manager/PluginHost.h
+++ b/libraries/lib-module-manager/PluginHost.h
@@ -54,12 +54,12 @@ public:
     * the main application)
     * \return true if host has started successfully
     */
-   static bool Start();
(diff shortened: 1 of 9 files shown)


libraries/lib-module-manager/PluginHost.h
```diff
@@ -54,12 +54,12 @@ class PluginHost final : public IPCChannelStatusCallback
...
     * \return true if host has started successfully
     */
-   static bool Start();
+   static bool Start(int connectPort);
 
    ///Returns true if current process is considered to be a plugin host process
    static bool IsHostProcess();
 
-   PluginHost();
+   PluginHost(int connectPort);
```
crsib, line 62: Please mark `explicit`

crsib (approved):

Issue #3210, opened 2022-07-11
Cannot instantiate an LV2 effect on MacOS


dozzzzer:
**Describe the bug**
'Cannot instantiate effect' error message is observed when calling an LV2 effect on Mac M1.

**To Reproduce**
1. install an LV2 effect (e.g. [Dragonfly Reverb](https://michaelwillis.github.io/dragonfly-reverb/))
2. launch Audacity and load some audio
3. try to call the plugin either from the real-time effects sidebar or destructively

**Actual behavior**
<img width="304" alt="image" src="https://user-images.githubusercontent.com/70063970/178330328-1caa9b6f-dc08-4102-96b0-15ccd040a2c4.png">

**Expected behavior**
The effect opens up and functions properly

**Additional information (please complete the following information):**
 - OS: MacOS 12.2.1, M1 Pro chip
 - Version: latest alpha build

Note: **after** it fails to initialize destructively, playback in a normal way isn't possible: the output sound is just essentially noise.

https://user-images.githubusercontent.com/70063970/178332996-64d27427-4667-42ec-afd5-b6736e046413.mp4

crsib:
This bug is related to an issue within plugin discovery. Turns out, that before a real instance of an effect is created LV2 does not load any shared libraries at all. `LV2Effect::MakeInstance` does not actually instantiate the effect.

(@Paul-Licameli @vsverchinsky)

dozzzzer:
Verified

PR #3646, opened 2022-09-19
Plugin validation fixes

vsverchinsky:
Resolves: #3210

Recommended:

### DIFF
--- a/libraries/lib-module-manager/PluginHost.cpp
+++ b/libraries/lib-module-manager/PluginHost.cpp
@@ -35,14 +35,18 @@ namespace
          if(auto provider = ModuleManager::Get().CreateProviderInstance(providerId, wxEmptyString))
          {
             TranslatableString errorMessage{};
-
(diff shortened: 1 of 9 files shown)


libraries/lib-module-manager/PluginHost.cpp
```diff
@@ -35,14 +35,18 @@ namespace
...
+            auto validator = provider->MakeValidator();
+            provider->DiscoverPluginsAtPath(pluginPath, errorMessage, [&](PluginProvider *provider, ComponentInterface *ident)
             {
                //Workaround: use DefaultRegistrationCallback to create all descriptors for us
                //and then put a copy into result
                auto id = PluginManager::DefaultRegistrationCallback(provider, ident);
                if(const auto desc = PluginManager::Get().GetPlugin(id))
+               {
+                  if(validator)
+                     validator->Validate(*ident);
```
Paul-Licameli, line 47: This line is in a necessary catch block, but I would also define Validate to return bool, so that the failure paths in the overrides don't need to throw more exceptions.

This opinion isn't a blocker for this PR.

Paul-Licameli (approved):

Issue #3520, opened 2022-09-02
Parameters of some plugins are preserved when closing and re-opening them destructively


dozzzzer:
### Bug description

When calling a plugin via the main Effect/Generate/Analyze menu, changing its parameters, closing the dialog window, and then re-opening it back again, the state of the parameters is preserved.

### Steps to reproduce

1. load and select some audio
2. open **Bass and Treble** effect from the main Effect menu
3. change the effect's controls
4. press Esc key
5. re-open the effect
6. check the effect's controls state

### Expected behavior

The controls are at that exact state as they were when the effect's dialog window was called in step #2, i.e. before the changes were made.

### Actual behavior

The effect's controls state was preserved after the dialog window was closed

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

### Additional context

Full list of the plugins which preserve the state of the parameters:
<details>
Effects:

- Bass and Treble
- Auto Duck
- Change Pitch
- Change Speed
- Change Tempo
- Classic Filters
- Click Removal
- Compressor
- Distortion
- Filter Curve EQ
- Graphic EQ
- Loudness Normalization
- Normalize
- Paulstretch
- Phaser
- Sliding Stretch
- Truncate Silence

Generators:
- Chirp
- Tone

Analyzers:
- Plot Spectrum
</details>

Paul-Licameli:
Plot Spectrum has little in common with effects and generators.  A general fix won't do for that one.

This is old behavior in Audacity, not a regression.

But in 3.2, the effects that we have made "stateless" will do the correct thing.  The built-ins not on your list include the stateless ones.  Echo, Reverb, Wahwah, DTMF tones.

Paul-Licameli:
Note too (and it's an old bug):  If you "Repeat last effect" (Ctrl+R) then the settings last displayed before the Esc key are applied.  Not the settings last done with a press of Apply.

Paul-Licameli:
> Once applied, it starts showing the _previously applied values_ every time you open it, not the ones set before the dialog window was closed.

And that describes the expected behavior.  The problem is to make it so also when the effect has never yet been applied.

dozzzzer:
The only plugin which still seems to be preserving the state of parameters when re-opened is Compressor. The rest are in order.

dozzzzer:
Verified

PR #3648, opened 2022-09-20
Fix effect dialog esc key

Paul-Licameli:
Resolves: #3520

Recommended:


crsib:
Hmm, have you intended to open it against `release-3.2.0`?

Paul-Licameli:
Fixed

Why is there also a distinct alpha branch?

crsib:
I will delete it

PR #3656, opened 2022-09-20
Fix compressor presets...

Paul-Licameli:
... Capture correct pointers-to-members.  Mistake entered at commit 4dcc5ee

Resolves: #3520

Recommended:


crsib (approved):

Paul-Licameli:
> And my usual question - is the branch right?

Yes.

But now it seems that if I make a PR first on master, and then correct the base branch -- that causes the linking of the PR with the issue card that I want to see!

Paul-Licameli:
... but still I must move the issue card by hand to the right column.

Issue #3658, opened 2022-09-21
Audacity hangs when you switch a VST3 plugin in the stack during playback with the plugin interface open


Ryan-Miller-Audacity:
### Bug description

Switching a VST3 plugin in the stack while playback is active and the interface is open can cause Audacity to hang.

https://user-images.githubusercontent.com/88568755/191496843-8844b1ac-6e8b-4cf2-9492-4f845bfd1315.mp4

### Steps to reproduce

1. Open Audacity and grab/make some sound
2. Add a VST3 effect to the Realtime stack
3. Open the plugin interface
4. Start playback 
5. Change the plugin in the stack to a different plugin (with the interface still open) 
6. Observe hang

### Expected behavior

Plugin shoud be changed and Audacity doesn't hang

### Actual behavior

Audacity hangs when attempting to change the plugin during playback with the plugin interface open

### Audacity Version

latest beta

### Operating system

Windows, macOS, Linux

### Additional context

@Paul-Licameli noted the following: 

![image](https://user-images.githubusercontent.com/88568755/191498087-d7709423-971b-4f64-b176-abf3a080fd91.png)

PR #3659, opened 2022-09-21
Fix freeze

Paul-Licameli:
Resolves: #3658

Recommended:

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -1181,9 +1181,12 @@ public:
       {
          insertItem();
       }
-      else if(msg.type == RealtimeEffectListMessage::Type::Replace)
+      else if(msg.type == RealtimeEffectListMessage::Type::WillReplace)
       {
          removeItem();
+      }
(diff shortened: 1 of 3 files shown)


src/RealtimeEffectPanel.cpp
```diff
@@ -1181,9 +1181,12 @@ class RealtimeEffectListWindow : public wxScrolledWindow
       {
          insertItem();
       }
-      else if(msg.type == RealtimeEffectListMessage::Type::Replace)
+      else if(msg.type == RealtimeEffectListMessage::Type::WillReplace)
```
vsverchinsky, line 1184: Not sure if it matters, but `Insert` and `Replace` should fix that as well...

vsverchinsky (approved):

src/RealtimeEffectPanel.cpp
Paul-Licameli: As it is, the publisher supplies a little more information, distinguishing types, which in this case isn't needed, but perhaps it will be useful later.

PR #3661, opened 2022-09-21
cmake: Use mp3lame pkg-config for lame

bdrung:
The lame project does not provide a pkg-config file for libmp3lame. lame 3.100-4 in Debian/Ubuntu ships a `mp3lame.pc` file in `libmp3lame-dev`.

Correct the pkg-config name to `mp3lame` to find the pkg-config file.

Recommended:

### DIFF
--- a/cmake-proxies/CMakeLists.txt
+++ b/cmake-proxies/CMakeLists.txt
@@ -105,7 +105,7 @@ add_conan_lib(
    libmp3lame/3.100
    REQUIRED
    INTERFACE_NAME libmp3lame::libmp3lame
-   PKG_CONFIG "lame >= 3.100"
+   PKG_CONFIG "mp3lame >= 3.100"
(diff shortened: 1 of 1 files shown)


crsib:
I can observe that Arch packages `lame.pc`: https://archlinux.org/packages/extra/x86_64/lame/files/

In any case we use `find_package` to locate libmp3lame as it seems there is no portable way to use `pkg-config` for it.

So this fix is harmless, but simply makes Ubuntu builds to use `pkg-config` instead of Arch.

bdrung:
Okay. Since Arch provides `lame.pc` I need to change the patch to look for both: `lame.pc` and `mp3lame.pc`.

bdrung:
I looked into searching for both pkg-config names, but that is non-trivial. So I decided to change the pkg-config name in the Debian/Ubuntu package to `lame.pc` to match the Arch Linux name.

Issue #3447, opened 2022-08-26
Auto-save does not recover effect settings changes


Paul-Licameli:
### Steps to reproduce

1. Make a track with an effect in its stack
2. Open the dialog, make changes, close the dialog
3. Force-quit Audacity
4. Restart and recover the project
5. Open the dialog again

### Expected behavior

See settings as you last left them.  But instead, settings remain as they were when the dialog first opened.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

I would not promise a fix in case the crash or force-quit happens while the dialog is open.

To fix the sequence as described, I thinkn a call to AutoSave when the dialog closes will be enough.

Paul-Licameli:
@crsib

crsib:
I am aware of this bug, it is not a blocker for alpha I think.

Paul-Licameli:
Is there a duplicate card for it already?

dozzzzer:
@vsverchinsky reopened: the fix doesn't seem to work for VST3s on macOS. All good with AUs though.

dozzzzer:
Verified on macOS and Windows

dozzzzer:
Reopened: the issue still occurs (macOS, build: https://github.com/audacity/audacity/actions/runs/3104289656)

dozzzzer:
@Paul-Licameli LV2 effect's parameters aren't saved when re-starting Audacity after force quit.

OS: macOS
Plugin: https://x42-plugins.com/x42/mac/x42-compressor-v0.6.2.pkg

Issue #3663, opened 2022-09-21
Data loss possible when realtime effect dialog closes


Paul-Licameli:
### Bug description

Project can save and then fail to reopen!

Updated STR:  you don't need play in progress to get the misbehavior.  Just have a sidebar dialog open when you close the project.  Needn't quit the program either.

### Steps to reproduce

1. Generate tone
2. Add an effect to its stack, e.g. Dragonfly HallReverb
(3. Play all in a loop)
4. Open effect settings (during play)
5. Ctrl+Q to quit or Ctrl+W to close
6. Say "yes" to saving the project
7. (Restart Audacity and) reopen the saved project
8. See message that project was not properly saved and was recovered from backup.  There are no tracks!

### Expected behavior

Keep the track that I saved!

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

Paul-Licameli:
At least on a macOS debug build with the effect I named, I can reproduce this regularly.

dozzzzer:
Verified

PR #3664, opened 2022-09-21
Fix data loss

Paul-Licameli:
Resolves: #3447

Recommended:

### DIFF
--- a/src/effects/RealtimeEffectStateUI.cpp
+++ b/src/effects/RealtimeEffectStateUI.cpp
@@ -28,6 +28,10 @@ const RealtimeEffectState::RegisteredFactory realtimeEffectStateUIFactory { [](a
 } // namespace
 
 
+BEGIN_EVENT_TABLE(RealtimeEffectStateUI, wxEvtHandler)
+   EVT_CLOSE(RealtimeEffectStateUI::OnClose)
+END_EVENT_TABLE()
+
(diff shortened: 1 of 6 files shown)


src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -156,3 +166,7 @@ void RealtimeEffectStateUI::UpdateTitle()
    mEffectUIHost->SetName(title);
 }
 
+void RealtimeEffectStateUI::AutoSave(AudacityProject &project)
+{
+   ProjectHistory::AutoSave::Call(project);
```
crsib, line 171: I still observe this call during the playback. I had an impression that this is wrong?

Paul-Licameli: Playback?  Which steps to reproduce are you referring to?

crsib: Hmm, looks like I've missed something. So this change is really about moving ProjectWindowDestroyedMessage?

Paul-Licameli: Updated steps to reproduce at https://github.com/audacity/audacity/issues/3663 explain that playback was not a necessary part

crsib (approved):

Issue #3289, opened 2022-07-25
3.1.3: Private libraries are not treated as private


dvzrv:
**Describe the bug**
When trying to package 3.1.3 on Arch Linux I ran into the issue, that the built private libraries (ending up in `/usr/lib/audacity/` on the target system) are linked against in the `audacity` executable in a global manner (as if they were located in `/usr/lib/`).
Without an override to [ldconfig](https://man.archlinux.org/man/ldconfig.8.en) it is not possible to run audacity without it crashing due to being unable to find the required shared object files.

**To Reproduce**
```
  cmake -B build -S $_name-$pkgver-Source \
      -DCMAKE_BUILD_TYPE=None \
      -DCMAKE_INSTALL_PREFIX=/usr \
      -DCMAKE_SKIP_RPATH=ON \
      -Daudacity_conan_enabled=OFF \
      -Daudacity_has_networking=OFF \
      -Daudacity_has_crashreports=OFF \
      -Daudacity_has_updates_check=OFF \
      -Daudacity_has_sentry_reporting=OFF \
      -Daudacity_lib_preference=system \
      -Daudacity_obey_system_dependencies=ON \
      -Wno-dev
  cmake --build build
  ctest --test-dir build --output-on-failure
  DESTDIR="$pkgdir" cmake --install build
  rm -rv "$pkgdir/usr/$pkgname"
```

**Actual behavior**
Audacity fails to start with
```
audacity: error while loading shared libraries: lib-screen-geometry.so: cannot open shared object file: No such file or directory
```

**Expected behavior**
Audacity starts successfully.

**Screenshots**
If applicable, add screenshots/video to help explain your problem.

**Additional information (please complete the following information):**
 - OS: Arch Linux
 - Version 3.1.3

**Additional context**
An ldconfig override should not be necessary, if shared libs are installed in the public location (e.g. `/usr/lib/`).
To prevent name conflicts it would make sense to provide an audacity specific namespace there (e.g. prefix with `audacity_`).

crsib:
I believe that correct RPATH is set during the `install` phase. But probably that was done after 3.1.3

dvzrv:
I'm also explicitly using `-DCMAKE_SKIP_RPATH=ON ` because it would create an invalid RPATH that points at the build directory.

crsib:
In master rpath is set to `$ORIGIN/../lib/audacity`

dvzrv:
[...]

PR #3671, opened 2022-09-22
Fix rpath for private libraries on Linux

bdrung:
Installing audacity on Linux will produce private libraries that have `RUNPATH` set to the build directory instead of the installation directory. https://github.com/audacity/audacity/pull/3302 implemented a workaround. Instead of letting cmake write a wrong `RUNPATH` into the installed libraries and fixing `RUNPATH` with a lengthy script afterwards, better configure cmake to create a correct `RUNPATH` in the first place.

The root cause is that the library directory is copied to the installation directory without touching the libraries. The cmake wiki says in RPATH handling caveats [1]: "Since install-side RPATH tweaking is an operation that is done by target-specific installation handling, any target that should have its install RPATH changed (e.g. to `CMAKE_INSTALL_RPATH`) needs to end up in the installation via an `install(TARGETS ...)` signature and not via directory-based copying."

So replace `install(DIRECTORY ...)` by individual `install(TARGETS ...)` for the libraries and modules. Then cmake will replace the `RUNPATH` to `$ORIGIN/../lib/audacity`, which is still incorrect. Therefore set `INSTALL_RPATH` explicitly.

Fixes: https://github.com/audacity/audacity/issues/3289
[1] https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#caveats

Recommended:


crsib:
Are you sure that this fixes is needed? Can you please provide more details?

bdrung:
Yes, since https://github.com/audacity/audacity/pull/3302 was just a workaround instead of a real fix. This merge request reverts https://github.com/audacity/audacity/pull/3302 and fixes the cmake configuration.

crsib:
Why that was a workaround and not a fix?

crsib:
From you PR I observe that you install unconditionally, so I expect problems with it on platforms other than Linux. Have you checked on them?

bdrung:
I consider it a workaround to let cmake install libraries with wrong `RUNPATH` and run a 59 line long script (named `fix_rpath.cmake`) afterwards to fix it.

bdrung:
Thanks for the pointer. I changed the install command to only be included on Linux (using the same if-condition that were applied previously).

crsib:
So it appears that your fix breaks the Linux build when it is not using system only libraries. And now I remember why I had to hack-around the issue in the first place.

bdrung:
The solution from this merge request should also work for AppImage. I haven't found time yet to debug the failure there.

crsib:
It will not work for the case when Conan is enabled. It works for "system only" builds. This means that we either need to

* vendor all the dependencies back,
* or drop older Linux distributions,
* or manually build and install all the packages. 
 
We do not have plans for either of the options right now nor in foreseeable future.

bdrung:
Why won't it work when Conan is enabled?

PR #3692, opened 2022-09-26
Make phaser stateless

pietro68:
Resolves: #3689 

Recommended:

### DIFF
--- a/src/effects/Phaser.cpp
+++ b/src/effects/Phaser.cpp
@@ -47,7 +47,9 @@ const EffectParameterMethods& EffectPhaser::Parameters() const
    static CapturedParameters<EffectPhaser,
       Stages, DryWet, Freq, Phase, Depth, Feedback, OutGain
    > parameters{
-      [](EffectPhaser &, EffectSettings &, EffectPhaser &e, bool updating){
+
+      [](EffectPhaser &, EffectSettings &, EffectPhaserSettings &e, bool updating)
+      {
          if (updating)
             e.mStages &= ~1; // must be even, but don't complain about it
          return true;
@@ -90,7 +92,6 @@ END_EVENT_TABLE()
 
 EffectPhaser::EffectPhaser()
 {
-   Parameters().Reset(*this);
    SetLinearEffectFlag(true);
 }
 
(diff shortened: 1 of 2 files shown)


src/effects/Phaser.h
```diff
@@ -91,21 +91,21 @@ class EffectPhaser final : public StatefulPerTrackEffect
...
+      return 0;
+   }
+
+   // Temporarily needed
+   size_t RealtimeProcess(size_t group, EffectSettings& settings,
+      const float* const* inbuf, float* const* outbuf, size_t numSamples)
+      override
+   {
+      return 0;
+   }
```
Paul-Licameli: I don't like the unexplained "Temporarily needed" functions.

Instead, change the base class to PerTrackEffect, then you don't need these two overrides.

Also, declare `MakeInstance` as override.

src/effects/Phaser.cpp
```diff
@@ -130,6 +130,61 @@ struct EffectPhaser::Validator
...
+      float* const* outBlock,
+      size_t                 blockLen);
+
+   void Coefficients(double hz, double slope, double gain, double samplerate, int type,
+      double& a0, double& a1, double& a2, double& b0, double& b1, double& b2);
+
+   float DoFilter(EffectPhaserState& data, float in);
+
+   EffectPhaserState mMaster;
+   std::vector<EffectPhaserState> mSlaves;
```
Paul-Licameli, line 176: Commenting on the whole PR, it appears to be a correct transformation.  But I want to add a requirement.

Make the objects in this array the same class as the instance -- as is done already for AudioUnits and VST2.

When that works, do the same when you transform other built-in effects to statelessness.  Write PRs to do the same retroactively for the Reverb and Wahwah transformations.

If time allows, make PRs to do the same for Ladspa and LV2.

The reason?

I don't like the duplicated interface for processing EffectInterface.h.  There is RealtimeProcess but also ProcessBlock.  I wabt them unified so the interface is simpler.

That means eliminating the group argument of realtime process.  More about that later:  but making every array of slaves an array of Instances is the beginning.  Then the "array of slaves" is exposed to RealtimeEffectState, not an internal detail of each Instance class.  The RealtimeEffectState will dispatch to the correct Instance object.

src/effects/Phaser.h
pietro68: Ok, integrated that in step 3

pietro68: Ok, both integrated into step 3 now.

Paul-Licameli (approved):
All looks good.

Adapt the rest of the real-time capable built-ins to this same pattern in other pull requests.

Those may soon include Distortion and Echo.

Issue #3484, opened 2022-08-30
Unwanted beep when either the add or replace effect button is pressed using the keyboard.


DavidBailes:
### Bug description

In the realtime effects panel, there's an unwanted beep when either the add or replace effect button is pressed using the keyboard.

### Steps to reproduce

1. Open the realtime effects panel.
2.  Tab to the add effect button, or the replace effect button for an effect which has already been added.
3.  Press Enter to press the button.

### Expected behavior

A menu opens, without any beeps.

### Actual behavior

There is a beep sound indicating some issue with associated command, but the menu opens.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

Could be (very) annoying.

dozzzzer:
I was able to reproduce it on Windows 11, it plays the system "Asterisk" sound when I press enter having the "Add Effect" button in focus.

dozzzzer:
Verified

Issue #3230, opened 2022-07-14
Allow keyboard focus on the combined meter/mixer sliders


ksoze95:
https://github.com/audacity/audacity/issues/2779#issuecomment-1093835971

dozzzzer:
Verified on macOS and Windows

Issue #3690, opened 2022-09-26
The scale for the recording volume slider is badly wrong


DavidBailes:
### Bug description

On default settings, the range of the recording volume slider is 60db.
This slider maps exactly onto the Windows volume slider for the device. A pointed out by "guzmann" on discord, the range of the Windows volume slider depends on the device. For example, the microphone in my webcam has a range of 24db, which is quite different from the range of 60db shown in Audacity.
I think it's very confusing for users that the scale is so inaccurate.

### Steps to reproduce

Change the recording volume slider by x db

### Expected behavior

The recorded volume reduces by x db

### Actual behavior

The recorded volume reduces by y db, where x and y are quite different.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Before the mixer and meter toolbars were merged, the recording volume slider had a scale 0 to 1.0, which is the same scale which is used by the Windows recording volume slider, by default.

Ryan-Miller-Audacity:
I agree it's quite confusing that the slider can be set in such a manner being effected by windows settings; however, I believe the real answer is just decoupling the windows recording setting and Audacity recording setting so that Audacity can independently set it's own recording values without interference from windows settings (as you would see in many common DAWs). We have talked about this in design meetings a few times. 

/cc @Tantacrul @LWinterberg

crsib:
I think users won't be happy with such a change at all.

Tantacrul:
I wonder about that. However, I think we can possibly get out of jail for free by providing a default 'Same as system' option, as well as the ability to manually adjust your mic settings.

Both Discord and Zoom do this and it's rather nice.

crsib:
On Windows Zoom adjusts system level for both input and output. Discord does the opposite, which is fun :-)

In any case - the problem is that after we have merged meter and slider - it is (is it?) expected that slider and meter are using the same scale, which is not the case currently. And while we can make it work for the output, it is not quite possible to do the same for the input now.

Tantacrul:
Well, we do need to discuss 'per track' input monitoring anyway, which will kill the recording meter on the top bar completely. I assert that this will solve all our problems 👀

DavidBailes:
> And while we can make it work for the output

Testing using Reaper with a loopback input, then at least on Windows, the scale of the output slider appears to be correct.

SteveDaulton:
Seems to be a similar issue to https://github.com/audacity/audacity/issues/3322

[...]

PR #3696, opened 2022-09-27
Meter toolbar navigation & accessibility

vsverchinsky:
Resolves: #3230
Resolves: #3810 
Resolves: #3484

Recommended:


DavidBailes:
I'm not sure what the current design spec is for this, but what is read by screen readers is different from my initial suggestion: https://github.com/audacity/audacity/issues/2779#issuecomment-1093835971
What I was expecting was that when you tabbed to the button containing the playback/recording icon and the slider, screen readers would read something like
"record meter peak -30db button", and 
"recording volume slider -4db",
respectively.

However, what gets read by nvda when tabbing to the button and slider is:
"Record  button", and
"Record Meter  Peak -60 dB  slider  46"
respectively.
I think the current implementation is less clear and more confusing compared to what I was expecting.

There are also some issues to do which how the slider is read when its value is changed, but these are secondary to the above issue.

DavidBailes:
@LWinterberg . Given that the scale for the recording volume slider presented visually is wrong (https://github.com/audacity/audacity/issues/3690),
I think it would be better if the values of the volume slider read by screen readers were in the range 0 to 100, rather than incorrect ones in the range -60db to 0db. I think that would be far less confusing. What do you think?

Tantacrul:
I agree with the suggestion that we move from dB to %.

Let's use 0-100 and then we can provide more precise meters in the mixer (which we will be promoting and improving soon)

vsverchinsky:
@DavidBailes @Tantacrul I've updated a PR, can you please tell if it's alright now?

Tantacrul:
My days! I am really struggling to figure out how to tab to the record volume slider (using MacOS) in order to change it and hear a screenreader 😦.

Is it mainly a Windows thing? Need tutorial!

Apart from that two quick points:

When changing the slider I can see a colon before the %. So it says **:77%**
image

When I hover over the 'Add effect' button, Voice Over (MacOS) calls it a 'Taste'. (!) Are we aware of this strangeness? I guess This should be raised as a separate issue. I was just curious if anyone had any thoughts.
image

src/toolbars/MeterToolBar.cpp
```diff
@@ -187,7 +188,7 @@ void MeterToolBar::Populate()
    if( mWhichMeters & kWithPlayMeter ){
       mPlaySetupButton = safenew AButton(this);
       mPlaySetupButton->SetLabel({});
-      mPlaySetupButton->SetName(_("Playback"));
+      mPlaySetupButton->SetName(wxString::Format(_("Playback meter peak %d db"), -DecibelScaleCutoff.GetDefault()));
```
Paul-Licameli, line 191: Did you confirm that the buttons reformat after the value is changed and the Preferences dialog is dismissed?

If that doesn't work, then I maybe commit 3598fe4 added enough calls to Invalidate().

DavidBailes:
> @DavidBailes @Tantacrul I've updated a PR, can you please tell if it's alright now?

Just to let you know that I've started to have a look at this. It clearly isn't complete, and I'll add detailed comments soon.

src/toolbars/MeterToolBar.cpp
```diff
@@ -146,43 +144,87 @@ void MeterToolBar::ReCreateButtons()
...
       /* i18n-hint: (noun) The meter that shows the loudness of the audio being recorded.*/
       mRecordMeter->SetName( XO("Record Meter"));
       /* i18n-hint: (noun) The meter that shows the loudness of the audio being recorded.
        This is the name used in screen reader software, where having 'Meter' first
        apparently is helpful to partially sighted people.  */
       mRecordMeter->SetLabel( XO("Meter-Record") );
-      mSizer->Add( mRecordMeter, wxGBPosition( 0, 0 ), wxDefaultSpan, wxEXPAND );
    }
 
    if( mWhichMeters & kWithPlayMeter ){
```
Paul-Licameli, line 190: Does mWhichMeters still need to be a bitwise or of non-exclusive choices?

There was old, old code for an alternative combined play & record meter toolbar in one, instead of the separate ones.  I forget when that was removed, but it appears the program is still constructing this third, combined meter toolbar, and yet never showing it and giving you no user interface to turn it on!

src/widgets/ASlider.cpp
```diff
@@ -530,6 +530,7 @@ LWSlider::LWSlider(wxWindow *parent,
       speed = 0.5;
       break;
    case FRAC_SLIDER:
+   case PERCENT_SLIDER:
```
Paul-Licameli, line 533: The commit that adds PERCENT_SLIDER for the play and record meters is perhaps incomplete.

I can double-click on the slider thumb and get a dialog to enter a value.  But it must be a fraction from 0 to 1, not a percentage value.  Is that acceptable?  A design question @Tantacrul

Paul-Licameli: I also see that not all calls to SendUpdate() are followed by ShowTip()  -- the dialog has some of those calls.  Should ShowTip() be added?

Paul-Licameli: At this commit, at least in macOS, I do not yet get the hover texts over these two slider thumbs.  But I do get them for track pan and gain and the mixer board gain slider.  Maybe a later commit will fix this. I haven't built them all yet.

src/widgets/MeterPanel.cpp
```diff
@@ -346,11 +346,11 @@ MeterPanel::MeterPanel(AudacityProject *project,
       mSlider = std::make_unique<LWSlider>(this, XO(""),
          pos,
          size,
-         FRAC_SLIDER,
+         PERCENT_SLIDER,
```
Paul-Licameli, line 351: OK, after another commit that changes this default, I do see tooltips, on macOS -- but only if I click the thumb, not when I just hover the mouse over it.  This is different from the pan and gain sliders.

The text I see is like  ": 47%" suggesting something was supposed to be formatted before the ":" but is left empty.

I think that is mName being empty at these lines of ASlider.cpp:

```
      /* i18n-hint: An item name followed by a value, with appropriate separating punctuation */
      label = XO("%s: %s").Format( mName, val );
```

With pan and gain sliders, I see a hover text, but when I press the mouse, I see the same text, but it moves and is drawn in a bordered, rounded rectangle.  That's a little strange.

src/toolbars/MeterToolBar.cpp
```diff
@@ -183,6 +183,7 @@ void MeterToolBar::Populate()
        This is the name used in screen reader software, where having 'Meter' first
        apparently is helpful to partially sighted people.  */
       mRecordMeter->SetLabel( XO("Meter-Record") );
+      mRecordMeter->SetSliderName(XO("Recording Level"));
```
Paul-Licameli, line 187: Okay, at another commit, now I see the problems of previous commits corrected.  Names before the :, and a nice hover text with a rounded border.

In fact, now the old behavior of the pan and gain sliders looks worse!  There should be a minor issue I think to fix them to be consistent with the volume sliders.  Which is not part of the scope of this PR.

Paul-Licameli: It is still the case, however, that when double clicking the volume slider thumb to get a dialog, I'm still not entering a percentage value, but instead a fraction in the range 0.0 to 1.0.

src/toolbars/MeterToolBar.cpp
```diff
@@ -150,7 +151,7 @@ void MeterToolBar::Populate()
       //(maybe we should do it differently for Arabic language :-)  )
       mRecordSetupButton = safenew AButton(this);
       mRecordSetupButton->SetLabel({});
-      mRecordSetupButton->SetName(_("Record"));
+      mRecordSetupButton->SetName(wxString::Format(_("Record meter peak %d db"), -DecibelScaleCutoff.GetDefault()));
```
Paul-Licameli, line 155: Why GetDefault() and not just Read()?

Now if I change interface preferences so the bottom of scale is not -60 dB, the drawing updates for it correctly, but the name spoken for the button still says -60, inconsistently with the display.

Paul-Licameli: `DecibelScaleCutoff.GetDefault()` also occurs in the constructor of MeterPanel.  I put it there at commit 01a53e7da9b616d26d1bb49f381180f04438504f.  As a behavior-preserving code transformation, it was correct as I did it, eliminating the use of the constant ENV_DB_RANGE.  But was old behavior correct?  Perhaps not, and that other line should also be corrected to use Read().

Paul-Licameli:
I'm done with this round of review.

DavidBailes:
Some comments on the meter buttons. The names of these buttons read by screen readers does not get updated to reflect the updated peak, whether their is monitoring, and whether there has been clipping.
One way of fixing this would be as follows:
- derive a class from Abutton, say called MeterButton, and use it instead of AButton in this toolbar.
- In the MeterButton class override the function wxString GetName() const, and in this function return a string with the same contents as the string produced by MeterAx::GetName(). (MeterAx now represents a slider, so MeterAx::GetName() should be removed).

With respect to fully transferring the fix in commit https://github.com/audacity/audacity/commit/3d420e019ceb7aa0d0d86b51fb44a17680eb34c4,  once the accessibility names of the buttons are correct, I can check whether anything further needs doing about this. There may need to be a minor modification to AButtonAx to give it a silent mode.

DavidBailes:
> My days! I am really struggling to figure out how to tab to the record volume slider (using MacOS) in order to change it and hear a screenreader 😦.

Testing with commit 700256 on MacOS
- I can tab to the sliders
- After pressing VO+Shift+Down arrow to interact with one, I can use the arrow keys to change the value of the slider
- VoiceOver does not automatically read the new value of the slider when it is changed - don't think it's ever done this for sliders in Audacity.

> 
> When I hover over the 'Add effect' button, Voice Over (MacOS) calls it a 'Taste'. (!) Are we aware of this strangeness? I guess This should be raised as a separate issue. I was just curious if anyone had any thoughts. image

Fortunately, I can't reproduce this.

DavidBailes:
> Some comments on the meter buttons. The names of these buttons read by screen readers does not get updated to reflect the updated peak, whether their is monitoring, and whether there has been clipping.
> One way of fixing this would be as follows:

Another way of fixing it would be:
- Derive a class from AButtonAx, say MeterButtonAx, and set this as the accessible for the AButton in this toolbar.
- In the MeterButtonAx class override the function wxAccStatus AButtonAx::GetName, and in this function create a string with the same contents as the string created by MeterAx::GetName(). (MeterAx now represents a slider, so MeterAx::GetName() should be removed).

Again, there may need to be some minor fixes to fully transfer the fix in commit https://github.com/audacity/audacity/commit/3d420e019ceb7aa0d0d86b51fb44a17680eb34c4.

src/toolbars/MeterToolBar.cpp
vsverchinsky: Thanks, fixed

src/toolbars/MeterToolBar.cpp
vsverchinsky: Thanks, dialog now shows values in range 0..100

petersampsonaudacity:
@dozzzzer @LWinterberg 

Testing on W19 with @vsverchinsky Vitaly's branch:  audacity-win-3.2.2-alpha-20221018+915aba6-x64

In this branch build:

a) the sliders on the Meter Toolbars now have the value of the current position

b) the tool-tip nicely updates when the slider is moved

image

DavidBailes:
Comments on the meter buttons in the latest commit:
- if the meter is clipped, this is not read by a screen reader. In bool MeterPanel::IsClipping() const, it should be mBar[c].clipping, not mBar[c].isclipping, as in the old MeterAx::GetName().
- when appropriate, the screen reader should read "monitoring" or "active", but not both, as in the oldMeterAx::GetName().
- the accessibility name does not include the current peak level, as in the old MeterAx::GetName(). In MeterToolBar::Populate(), the accessibility names of the buttons should just be set to "record meter" and "playback meter", as the peak level should be included in MeterButtonAx::GetName().
- When the button is pressed by pressing Enter, there is an unwanted "beep" indicating an invalid action. See also https://github.com/audacity/audacity/issues/3484

petersampsonaudacity:
@vsverchinsky  and @crsib 

since Vitaly's branch build audacity-win-3.2.2-alpha-20221018+915aba6-x64
https://github.com/audacity/audacity/actions/runs/3272357779

at least fixes:    #3810  Hovertext tooltips for recording and playback sliders no longer show the values
and this is a **_very_** useful improvement for users,

is there any chance that we could pull this fix right now while the further important accessibility issues that @DavidBailes raises can be dealt with separately - leaving this PR thread and #3230 still open - while closing off #3810 ?

DavidBailes:
> * When the button is pressed by pressing Enter, there is an unwanted "beep" indicating an invalid action.

You may well be aware of this already, but there's a comment in void MeterPanel::OnKeyDown(wxKeyEvent &evt) about avoiding a beep.

vsverchinsky:
>the accessibility name does not include the current peak level, as in the old MeterAx::GetName(). In MeterToolBar::Populate(), the accessibility names of the buttons should just be set to "record meter" and "playback meter", as the peak level should be included in MeterButtonAx::GetName().

@DavidBailes Do I understand correctly that button name should include **peak hold** value and the slider should include **peak** value (`DecibelScaleCutoff`)?

DavidBailes:
> @DavidBailes Do I understand correctly that button name should include **peak hold** value and the slider should include **peak** value (`DecibelScaleCutoff`)?

- The slider should NOT include the peak value.

- The button name should include the peak hold value, as it did in the old MeterAx::GetName().  (And the order of phases that make up the name should remain the same.) Snippet of old code:

      float peak = 0.;
      bool clipped = false;
      for (unsigned int i = 0; i < m->mNumBars; i++)
      {
         peak = wxMax(peak, m->mBar[i].peakPeakHold);
         if (m->mBar[i].clipping)
            clipped = true;
      }

      if (m->mDB)
         *name += wxT(" ") + wxString::Format(_(" Peak %2.f dB"), (peak * m->mDBRange) - m->mDBRange);
      else
         *name += wxT(" ") + wxString::Format(_(" Peak %.2f "), peak);

      if (clipped)
         *name += wxT(" ") + _(" Clipped ");
   }

DavidBailes (approved):
The reading by screen readers of the meter toolbar is now fine.
Glad to see that you got rid of the unwanted "beep", and that your fix also fixes the issue: https://github.com/audacity/audacity/issues/3484

DavidBailes:
@Paul-Licameli . I know you're busy, but could you review this. I may want to submit a fix for another bug, and want to avoid conflicts when it's merged.

Paul-Licameli:
> @Paul-Licameli . I know you're busy, but could you review this. I may want to submit a fix for another bug, and want to avoid conflicts when it's merged.

Very nearly done

src/toolbars/MeterToolBar.cpp
```diff
@@ -40,6 +40,87 @@
...
+         *name = button->GetName();
+
+         if(const auto panel = mAssociatedMeterPanel.get())
+         {
+            // translations of strings such as " Monitoring " did not
+            // always retain the leading space. Therefore a space has
+            // been added to ensure at least one space, and stop
+            // words from being merged
+            if(panel->IsMonitoring())
+               *name += wxT(" ") + _(" Monitoring ");
```
Paul-Licameli, line 76: If you add the space yourself, for the special reasons explained in the comment, then do not include the spaces in the strings to be translated.

Paul-Licameli: This applies to five translatable strings in this function.

Paul-Licameli (approved):

petersampsonaudacity:
@crsib @vsverchinsky 

So now this has been approved by @Paul-Licameli  and @DavidBailes  - how does this get pulled so it can be tested in a 2.3.3 alpha or beta?

dozzzzer:
> @dozzzzer did yet complete testing it, right?

Not yet.

petersampsonaudacity:
Note that latest 3.3.0 alpha:  audacity-win-3.3.0-alpha-20221109+fe50013-x64-msvc2022

does **_not_** yet have thus merged in.

Looks good on latest 3.2.2.beta:  audacity-win-3.2.2-beta-20221109+f0bb347-x64-msvc2022

PR #3701, opened 2022-09-27
Drop UBUNTU_MENUPROXY=0 workaround

bdrung:
wxWidgets 2.6 did not show menus correct on the global menu applet in Unity in 2010 ([Ubuntu bug #662007](https://launchpad.net/bugs/662077)). The environment variable `UBUNTU_MENUPROXY` was set as workaround.

Audacity 3.2.0 with wxWidgets 3.2 on Ubuntu 22.10 (kinetic) works fine under the Unity desktop. Therefore drop the old workaround.

Recommended:


crsib:
Ubuntu 22.10 is not the only the Ubuntu version Audacity is expected to work on. Have you checked that your PR works on 18.04? Have you checked that AppImage works on 18.04? Does it break anything on Arch so you are trying to fix it?

bdrung:
I haven't tested other distributions or releases yet.

LWinterberg:
Hello! We just [switched master to 4.0.0](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444) which makes this PR incompatible and/or obsolete as we're switching UI framework and reorganizing large parts of the codebase.

Issue #2305, opened 2021-12-18
"-Daudacity_use_midi=off" seems to be ignored


juanitotc:
When building with "-Daudacity_use_midi=off", the build fails looking for portmidi.h with:

```
[ 64%] Building CXX object src/CMakeFiles/Audacity.dir/MIDIPlay.cpp.o
[...]

PR #3717, opened 2022-09-29
Fix setting audacity_use_midi=off

bdrung:
Resolves: #2305
When disabling MIDI support, cmake will fail:

```
$ cmake -B build -Daudacity_use_midi=off
[...]
CMake Error at cmake-proxies/CMakeLists.txt:314 (message):
  EXPERIMENTAL_MIDI_OUT requires USE_MIDI
```

Disable `EXPERIMENTAL_MIDI_OUT` in case MIDI support was explicitly disabled.

Recommended:

### DIFF
--- a/src/Experimental.cmake
+++ b/src/Experimental.cmake
@@ -107,23 +107,6 @@ set( EXPERIMENTAL_OPTIONS_LIST
    # Paul Licameli (PRL) 29 Nov 2014
    #IMPROVED_SEEKING
 
-   #MIDI_IN
-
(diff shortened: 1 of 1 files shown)


LWinterberg:
Hello! We just [switched master to 4.0.0](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444) which makes this PR incompatible and/or obsolete as we're switching UI framework and reorganizing large parts of the codebase.

Issue #3716, opened 2022-09-29
Freeze via left clicking play head bar and pushing space bar to play


Dontknowhow2code:
### Bug description

Clicking the play head area (highlighted in yellow) to play the track and pushing space to play the track at the same time causes audacity to crash.
It is easier to rapidly left click on the play head area whilst rapidly spamming the space bar.
![audacity crash](https://user-images.githubusercontent.com/81732647/192980107-1847e987-f399-46aa-aa54-7463e513abeb.png)

### Steps to reproduce

1. Open audacity
2. Open any track
3. Rapidly left click on the play head area repeatedly whilst rapidly pressing the space bar repeatedly.
4. Audacity crashes

### Expected behavior

Audacity shouldn't crash and should just play and pause over and over again.

### Actual behavior

Crash

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Log during crash:
> 21:26:51: Audacity 3.2.0
> 21:26:51: sqlite3 message: (1) no such table: project in "SELECT 1 FROM project LIMIT 1;"
> 21:26:51: Error: Failed to load shared library 'avformat-59.dll' (error 126: The specified module could not be found.)
> 21:26:51: Error: Failed to load avformat-59.dll (error 126: The specified module could not be found.)
> 21:26:51: Error: Failed to load shared library 'avformat-58.dll' (error 126: The specified module could not be found.)
> 21:26:51: Error: Failed to load avformat-58.dll (error 126: The specified module could not be found.)
> 21:26:51: Error: Failed to load shared library 'avformat-57.dll' (error 126: The specified module could not be found.)
> 21:26:51: Error: Failed to load avformat-57.dll (error 126: The specified module could not be found.)
> 21:26:51: Error: Failed to load shared library 'avformat-55.dll' (error 126: The specified module could not be found.)
> 21:26:51: Error: Failed to load avformat-55.dll (error 126: The specified module could not be found.)

device info txt generated by audacity:
[deviceinfo.txt](https://github.com/audacity/audacity/files/9672666/deviceinfo.txt)

computer info:
Dell Inspiron 3505
CPU: AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx 2.10 GHz
8GB RAM
x64-based processor

screen recording of crash:
https://user-images.githubusercontent.com/81732647/192988667-76d3dbce-e7b0-4886-99f1-78e6cd516b87.mp4

[...]

Issue #3699, opened 2022-09-27
Audacity freezes when quickly starting and stopping playback


LWinterberg:
### Steps to reproduce

1. have a track
2. mash spacebar to send many start and stop instructions
3. freeze

### Expected behavior

When more inputs are sent than Audacity can process, additional inputs should be discarded. 

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

I've seen two users report this bug on socials and forum, so it's at least slightly frequently encountered. 

When I pause Audacity in the VSCode debugger after it hits the freeze, I end up with the following call stack:

<details>
<summary>view call stack</summary>

![image](https://user-images.githubusercontent.com/87814144/192497504-bd931f9b-2541-4577-98a6-c28e512dc2fc.png)

```
ntdll.dll!00007ffd30924204() (Unknown Source:0)
ntdll.dll!00007ffd308db873() (Unknown Source:0)
KernelBase.dll!00007ffd2dd9b391() (Unknown Source:0)
msvcp140d.dll!00007ffc91302ede() (Unknown Source:0)
Audacity.exe!std::this_thread::sleep_until<std::chrono::steady_clock,std::chrono::duration<__int64,std::ratio<1,1000000000>>>(const std::chrono::time_point<std::chrono::steady_clock,std::chrono::duration<__int64,std::ratio<1,1000000000>>> & _Abs_time) Line 201 (c:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.33.31629\include\thread:201)
Audacity.exe!std::this_thread::sleep_for<__int64,std::ratio<1,1000>>(const std::chrono::duration<__int64,std::ratio<1,1000>> & _Rel_time) Line 207 (c:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.33.31629\include\thread:207)
Audacity.exe!AudioIoCallback::WaitForAudioThreadStopped() Line 3246 (c:\Users\mc\Documents\GitHub\audacity\src\AudioIO.cpp:3246)
Audacity.exe!AudioIO::StopStream() Line 1408 (c:\Users\mc\Documents\GitHub\audacity\src\AudioIO.cpp:1408)
Audacity.exe!ProjectAudioManager::Stop(bool stopStream) Line 531 (c:\Users\mc\Documents\GitHub\audacity\src\ProjectAudioManager.cpp:531)
[...]

PR #3721, opened 2022-09-29
Wait for AudioThread to ACK in StartStream

crsib:
Resolves: #3699 
Resolves: https://github.com/audacity/audacity/issues/3716

When rapidly mashing the spacebar, there is a chance that AudioThread will 
never change the internal state to `State::eLoopRunning`.

In this scenario, `StopStream` will wait indefinitely for the AudioThread to ACK.

Waiting in `StartStream` ensures, that AudioThread has entered an expected 
state before the `StopStream` is called.

Recommended:

### DIFF
--- a/src/AudioIO.cpp
+++ b/src/AudioIO.cpp
@@ -1028,6 +1028,9 @@ int AudioIO::StartStream(const TransportTracks &tracks,
       Publish({ pOwningProject.get(), AudioIOEvent::CAPTURE, true });
 
    commit = true;
+
+   WaitForAudioThreadStarted();
+
    return mStreamToken;
 }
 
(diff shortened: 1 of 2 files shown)


LWinterberg:
I can no longer reproduce either freeze in this build

pietro68 (commented):
This fix looks correct to me, thanks

src/AudioIO.cpp
```diff
@@ -3211,7 +3214,7 @@ auto AudioIoCallback::AudioIOExtIterator::operator *() const -> AudioIOExt &
 
 void AudioIoCallback::StartAudioThread()
 {
-   mAudioThreadTrackBufferExchangeLoopRunning.store(true, std::memory_order_release);
+    mAudioThreadTrackBufferExchangeLoopRunning.store(true, std::memory_order_release);
```
Paul-Licameli, line 3217: Indent changed.

I see StartAudioThreadAndWait was introduced at 00a8f2a68b1418b2f5979a7f33dc8092eb63b95a but never used.  And that contains the only other call to WaitForAudioThreadStarted.  That was an oversight in my earlier review.  Sorry.

The starting of the stream and the waiting for the ack are split, which isn't wrong.  So remove the unused function.

StopAudioThreadAndWait is unused too.

Reconsidering correctness of memory orderings, I think too that to be very correct, mAudioThreadTrackBufferExchangeLoopActive should be stored with release by AudioThread and tested with acquire by the main thread.  But maybe no difference as compiled for Intel.

Paul-Licameli: I just made a careful re-read of all of AudioIO::AudioThread.  Nothing more to say.

Paul-Licameli (approved):
OK with some suggestions

PR #3727, opened 2022-09-30
Wahwah: slaves are now instances, not states

pietro68:
Resolves: request from Paul to make slaves instances.

Recommended:


Paul-Licameli (approved):

pietro68:
for QA: I created a branch in which this effect is realtime-enabled, at https://github.com/audacity/audacity/pull/3756 you should be able to find builds there in a few minutes.

PR #3736, opened 2022-10-01
Fix data race

Paul-Licameli:
Resolves: possible data races

The recent fixes for Ladspa and LV2 effect metering introduced data races, but this fixes
them, and will next allow simplifications that preserve both metering and stickiness
of settings, while eliminating the copying of EffectSettings from the worker thread to the
main.

Recommended:


Paul-Licameli:
Superseded by #3762

Issue #3743, opened 2022-10-02
Maintainability of effect inter-thread communication, part one


Paul-Licameli:
### Bug description

Too many needless complications accumulated in the implementation of stackable effects.  We need to stop and reorganize this, so that we can also add VST2 statelessness without a lot of hacks as were done for VST3.

### Steps to reproduce

Nothing wrong now in the observable behavior

### Expected behavior

After merging, verify workings of Ladspa, LV2, VST3, AudioUnits effects, especially the stickiness of settings

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

PR #3744, opened 2022-10-02
Eliminate one settings copy

Paul-Licameli:
Resolves: #3743

Simplify inter-thread communications for stackable effects: 

One of two copyings of EffectSettings is eliminated -- the one from worker thread to main.

It was implemented in the first place for the needs of Ladspa and LV2 effects, to implement
metering updates correctly without data races.

But now EffectOutputs serves this purpose, so one copy of EffectSettings can be removed,
and with it various other complications that only worked around the undesirable consequences
of the overwrite of EffectSettings for other effect families.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -180,12 +180,6 @@ size_t EffectInstance::GetTailSize() const
    return 0;
 }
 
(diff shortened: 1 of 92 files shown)


Paul-Licameli:
Superseded by #3763

Issue #3745, opened 2022-10-02
Maintainability of effect inter-thread communication, part two


Paul-Licameli:
### Bug description

See #3742

This issue tracks the second of two parts of clean-up of the general stackable effect framework.  But then work for particular effect families may follow:  AU, VST3, and a new VST2 statelessness built on these better foundations.

LV2, Ladspa and built-in effects might also be simplified, but they did not get as difficult as the others.

### Steps to reproduce

Nothing wrong now in the observable behavior

### Expected behavior

After merging, verify workings of Ladspa, LV2, VST3, AudioUnits effects, especially the stickiness of settings

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

pietro68:
I verified that stickiness for vst2 works, with the first versions of the PRs. Still have to try the second versions.

dozzzzer:
LV2 effects (at least the ones made by x42) are now capable of input metering at a good refresh rate both on macOS and Windows.

Paul-Licameli:
> I verified that stickiness for vst2 works, with the first versions of the PRs. Still have to try the second versions.

Did you mean VST3?

Real stickiness also includes persistency when you save and reopen the project.  Did you test that?

dozzzzer:
RT application and stickiness of the settings work fine. Verified on LADSPA, VST3, AU, LV2 plugins on macOS, the build from the PR https://github.com/audacity/audacity/pull/3765

Ryan-Miller-Audacity:
Yes RT application, stickiness and persistence of settings as well as metering for LV2 and LADSPA plugins appear working well. Tested with Build from PR #3765 on both Windows and Linux.

PR #3746, opened 2022-10-02
Effect message

Paul-Licameli:
Resolves: #3745

The previous PR eliminated one copy of settings contents, from worker to main thread,
which became unnecessary after Ladspa and LV2 were rewritten to communicate back with
EffectOutputs instead.

This PR does not eliminate the copy of settings from main to worker, but instead is analogous
to the earlier step for Ladspa/LV2: first introduce another structure besides EffectSettings that
can carry information, and after that shift responsibility to it.

So EffectInstance::Message is now a std::any that is meant to contain only a change of settings;
furthermore, the MessageBuffer can know how to accumulate changes that are still not consumed
by the worker thread.

This alternative communication channel is not yet used by any particular effects.

The intention is that AudioUnits will be rewritten first to use it, in the next PR, and so serve as an
analogy for LV2 to be made stateless.

The rewriting of other effect families and built-ins may be done later.  When it is complete, then
CopySettingsContents overrides can all be eliminated as instead MoveMessageContents and
MoveOutputsContents overrides will serve the purpose.

Then EffectSettings can serve the more limited purpose of serializing and deserializing settings, not
the triple purpose it used to have.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -180,12 +180,6 @@ size_t EffectInstance::GetTailSize() const
    return 0;
 }
 
-void EffectInstance::AssignSettings(EffectSettings &dst, EffectSettings &&src)
-   const
-{
-   dst = src;
-}
-
 auto EffectInstance::GetLatency(const EffectSettings &, double) const
    -> SampleCount
 {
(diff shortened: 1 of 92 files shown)


Paul-Licameli:
Superseded by #3765

Issue #3694, opened 2022-09-26
Audacity crashes when applying mono Berzerk to mono track destructively


crsib:
### Bug description

Unhandled exception at 0x00007FF9A0873CE7 (WavesLib1_14.0.83_Win64.dll) in Audacity.exe: 0xC000041D: An unhandled exception was encountered during a user callback.

It appears to be a stack overflow due to infinite recursion:

Fragment of stack:

```
 	WavesLib1_14.0.83_Win64.dll!00007ff9a0873ce7()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a0874e3e()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a0872e2e()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a0874614()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a08882d2()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a088529b()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a087dbfa()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a07a0021()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a079f76f()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a0755fce()	Unknown
 	Berzerk Distortion.dll!00007ff9a78f5299()	Unknown
 	Berzerk Distortion.dll!00007ff9a78fa9be()	Unknown
 	Berzerk Distortion.dll!00007ff9a78fa824()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a078a362()	Unknown
 	WavesLib1_14.0.83_Win64.dll!00007ff9a078a6ac()	Unknown
 	Berzerk Distortion.dll!00007ff9a78ffb36()	Unknown
 	WaveShell1-VST3 14.0_x64.vst3!00007ff9a0ce3940()	Unknown
 	WaveShell1-VST3 14.0_x64.vst3!00007ff9a0ce935f()	Unknown
>	Audacity.exe!`anonymous namespace'::SetupProcessing(Steinberg::Vst::IComponent & component, Steinberg::Vst::ProcessSetup & setup) Line 164	C++
 	Audacity.exe!VST3Wrapper::FlushParameters(EffectSettings & settings) Line 606	C++
[...]

PR #3750, opened 2022-10-03
Fixes infinite recursion/idle freeze in VST3Wrapper

vsverchinsky:
Resolves: #3694 

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Wrapper.cpp
+++ b/src/effects/VST3/VST3Wrapper.cpp
@@ -244,14 +244,25 @@ public:
    {
       assert(mStateChangeSettings != nullptr);
       if(!mParametersCache.empty())
-      {
-         auto& vst3settings = GetSettings(*mStateChangeSettings);
-         for(auto& p : mParametersCache)
-            vst3settings.parameterChanges[p.first] = p.second;
-         mParametersCache.clear();
-         ++vst3settings.changesCounter;
-      }
+         FlushCache(*mStateChangeSettings);
       mStateChangeSettings = nullptr;
+   }
+
+   void FlushCache(EffectSettings& settings)
+   {
+      if(mParametersCache.empty())
+         return;
+
+      auto& vst3settings = GetSettings(settings);
+      for(auto& p : mParametersCache)
+         vst3settings.parameterChanges[p.first] = p.second;
+      mParametersCache.clear();
+      ++vst3settings.changesCounter;
+   }
+
+   void ResetCache()
+   {
+      mParametersCache.clear();
    }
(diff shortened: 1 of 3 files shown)


src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -529,28 +533,26 @@ bool VST3Wrapper::Initialize(EffectSettings& settings, Steinberg::Vst::SampleRat
...
+   FetchSettings(settings);
+
    if(mEffectComponent->setActive(true) == kResultOk)
    {
-      //Omit the return value check - some plugins return unexpected values (see bug #3581)
-      mAudioProcessor->setProcessing(true);
-      //...make sure they are delivered to the processor
-      ConsumeChanges(settings);
-      return true;
+      if(mAudioProcessor->setProcessing(true) != kResultFalse)
```
Paul-Licameli, line 546: One suggestion, if setProcessing fails after setActive succeeds, should the failure path setActive back to false?

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -602,14 +604,19 @@ void VST3Wrapper::FlushParameters(EffectSettings& settings)
...
+
       SetupProcessing(*mEffectComponent, mSetup);
       mActive = true;
       if(mEffectComponent->setActive(true) == Steinberg::kResultOk)
       {
          ConsumeChanges(settings);
-         mAudioProcessor->setProcessing(true);
-         Process(nullptr, nullptr, 0);
-         mAudioProcessor->setProcessing(false);
+         if(mAudioProcessor->setProcessing(true) != Steinberg::kResultFalse)
```
Paul-Licameli, line 615: Likewise here

Paul-Licameli (approved):
This class is still difficult to understand after discussing it and reading it.  Some difficulties are inherent, as with the need to `FlushParameters`, to compensate certain plug-in implementations.  But I think some of the complication can evaporate after my other changes are merged.

See for instance how the AudioUnit event handler changes:

https://github.com/audacity/audacity/pull/3753/commits/2877e4a6d31ad7c1ee47b99cdf4410074d8954de

The separation of concerns between one structure that reflects current settings state, and another that collects changes to be communicated inter-thread, makes the event callback branch-less.

Likewise I hope performEdit, the corresponding piece of VST3 implementation, will simplify and not depend on whether the plug-in is active, and BeginStateChange and EndStateChange will not be needed.

I hope the analogies among the several effect families will be more direct and understandable and easier for anyone else to maintain.

Issue #3729, opened 2022-10-01
Output metering doesn't work for Ladspa and LV2 plain UI


Paul-Licameli:
### Steps to reproduce

1. Generate DTMF tones with default settings
2. Put Fast Lookahead Limiter on it from the swh ladspa plugins; open its dialog; change Limit to -10 dB; keep it open
4. Put a metering LV2 plug-in, like x42-dpl Digital Peak Limiter Mono, on it, and open its dialog
5. Under Presets & settings > Options, un-check "Enable graphical interface"; close the dialog; open it again
6. Drag the dialogs to be non-overlapping so you can watch both
7. Play

### Expected behavior

See the latency meters in each dialog painted partly blue, and unvarying.

See the other meters also painted blue, and varying.

### Actual behavior

See only a narrow white line for each fo the meters.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

PR #3761, opened 2022-10-05
Separate effect outputs v2

Paul-Licameli:
Resolves: #3729

This redraft of #3735 avoids using std::any and also adds more const-correctness of arguments.

Recommended:


Paul-Licameli:
Latest force-push only changed commit comments.

Paul-Licameli:
Another push fixes one foolish mistake where I called `reserve` but meant `resize` in `LV2Effect::MakeOutputs`

src/effects/ladspa/LadspaEffect.cpp
```diff
@@ -1639,8 +1648,15 @@ LADSPA_Handle LadspaEffect::InitInstance(
...
       LADSPA_PortDescriptor d = mData->PortDescriptors[p];
-      if (LADSPA_IS_PORT_CONTROL(d))
-         mData->connect_port(handle, p, &controls[p]);
+      if (LADSPA_IS_PORT_CONTROL(d)) {
+         if (LADSPA_IS_PORT_INPUT(d))
+            mData->connect_port(handle, p, &controls[p]);
+         else {
+            static LADSPA_Data sink;
+            mData->connect_port(handle, p,
+               pOutputs ? &pOutputs->controls[p] : &sink);
```
vsverchinsky, line 1659: Correct me if I'm wrong, but effect writes `LadspaEffectOutputs` on the processing thread, here you bind control directly to the location of these outputs, what prevents data races in that case?

Paul-Licameli: You are right.  I explained the bug I introduce knowingly here but then correct in the second PR.

The two should be merged as a whole -- or perhapsw the whole sequence of four should be -- but I thought the first PR was big enough already.

Issue #3737, opened 2022-10-01
Ladspa and LV2 metering has a data race


Paul-Licameli:
### Bug description

Pull request #3735 to fix issue #3729 will be incomplete because it introduces data races.

Fixes for that are known.

It's not all one pull request so that reviewers can take all if it in stages.  So this is an issue that the second PR can link.

### Steps to reproduce

Play a track with an LV2 or Ladspa realtime effect that has output metering.  Listen and watch.

You probably will notice nothing.  However, analysis of the code shows that there is a subtle data race which is technically C++ undefined behavior.

As a result, maybe the sound and the visual output will not be as well synchronized as they should be.

### Expected behavior

All that QA should do to verify this fix is that LV2 and Ladspa effect metering aren't obviously broken.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

crsib:
Should this be closed @Paul-Licameli ?

PR #3762, opened 2022-10-05
Fix data race v2

Paul-Licameli:
Resolves: #3737

This redraft of #3736 without std::any lets one EffectOutputs reassign itself from another without the help of an Effect object.

Recommended:


Paul-Licameli:
This force-push adds overrides of EffectValidator::Disconnect for Ladspa and LV2 as precaution against dangling pointers (though I'm not certain whether there were such bugs before)

pietro68 (approved):
Code changes understood. I repeated tests with Ladspa fast lookahead limiter; metering, stickiness and serialization work.

Issue #3769, opened 2022-10-06
Crash editing Amplify (and some other effects) macro parameters


Paul-Licameli:
### Bug description

Reported to me by Larry Hudson who has for years taught classes in the use of Audacity for voiceover.

### Steps to reproduce

1. Tools > Macro Manager
2. Now in left pane, enter "My macro", or select existing macro name
3. Insert in right pane
4. Click Amplify in command list
5. Click "Edit Parameters" near the top
Crash!

But a known workaround:

0.  Select some sound, open Amplify once from the Effect menu, Esc key

Thereafter, editing the macro parameter is possible, so long as you do not close the Audacity project window in which you opened Amplify

### Expected behavior

No crash

### Actual behavior

a) _No response_  in some cases

b) in other cases one gets the ET-phone-home dialog

### Audacity Version

current master / alpha / nightly build

3.2.1

### Operating system

Windows, macOS, Linux

### Additional context

This bisects to 35581fa4b81753303a45b9229d89d8f2138c76e8 and does not affect all effects.

I don't know yet the complete list of effects that suffer.  Specific Amplify code wasn't changed.  It was instead things in the framework.  But Nyquist effects (example, Low-pass filter) do not have the problem.

Another rescue for this user could be to write a simple Nyquist effect to substitute for amplify and share it.

petersampsonaudacity:
Testing on W10 with 3.2.1   both **Graphic EQ** and **Filter Curve EQ** cause a crash too  -   I got the ET-phone-home-message and pressed the OK to send it on

petersampsonaudacity:
And **Autoduck** does the same

And two of the Vamp plugins:  **Bar and Beat Tracker: Bars**  and  **Bar and Beat Tracker: Beats**

Paul-Licameli:
Suspect effects include those that override the `Init` function

So these should be checked:
Nyquist (but negative, I checked)
Vamp
Amplify
AutoDuck
ChangeSpeed
ChangeTempo
Equalization
Classic Filters

petersampsonaudacity:
@Paul-Licameli @crsib @LWinterberg 

Testing with Paul's Branch build:  audacity-win-3.2.2-alpha-20221006+73eb7de-x64
**Fix crash when editing macro parameters of Amplify and others... Build Audacity #2740**

The following now do _**NOT**_ crash Audacity at Step of of the STR:
1. Amplify
2. Auto Duck
3. Classic Filters
4. Spectral Edit Shelves
[...]

PR #3772, opened 2022-10-06
Fix crash when editing macro parameters of Amplify and others...

Paul-Licameli:
... Commit 35581fa4 caused calls to Effect::Init that had not happened before. Some of these overrides assumed the setting of EffectBase::mTracks was done in DoEffect and got null pointer crashes.

Fix this by supplying an empty TrackList.

Also clean up mTracks in the effect object after DoEffect to avoid a dangling pointer.

Resolves: #3769 

Recommended:


ksoze95 (approved):

Paul-Licameli:
@ksoze95 @dozzzzer Please re-review and re-test, that the EQ effects are also fixed now

ksoze95 (approved):

Issue #3719, opened 2022-09-29
Crash when trying to use a deleted VST plugin before a Rescan


petersampsonaudacity:
### Bug description

If you delete VST plugins while Audacity is open and then try to use them, then Audacity will crash.

Does not occur if you exit Audacity and relaunch as a Rescan is forced on startup

### Steps to reproduce

1. Install some Steinberg VST plugins (I had Voxengo Elephant, BLOCKFISH compressor, FLOORFISH de-esser & SPITFISH de-esser
2. Observe they are installed in C:\Program Files (x86)\Steinberg\VST
3. Launch Audacity 3.2.0
4. click on the Effect menu
5. Observe: the VST effects are shown at the bottom of te list
6. get some audio
7. apply one of those VST effects to the audio
8. Observe: it works
9. Delete the VST dll files from C:\Program Files (x86)\Steinberg\VST
10. click on the Effect menu
11. Observe: the VST effects are still showing
12. try applying one of those VST effects to the audio
13. Observe: Audacity crashes

_I sent the ET-phone-home crash report_

### Expected behavior

No Crash

### Actual behavior

Audacity crashes

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

If after the crash you relaunch Audacity, all is well as the launch forces a rescan and the deleted VSTs are no longer shown in the Effect menu.

The workaround is to do a Rescan after Step 9 where the VST dll files are deleted

LWinterberg:
I can reproduce this. To me, the expected behavior would be an error message á la

> This plugin could not be loaded. It may have been deleted.

petersampsonaudacity:
@LWinterberg  -  perhaps an error message could advise the user to do a Rescan.

dozzzzer:
Verified on macOS and Windows

PR #3801, opened 2022-10-12
Disallow adding a destructive, or realtime, effect which has been deleted since application start up.

ksoze95:
Resolves: #3719 #3817 

Recommended:

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -25,6 +25,7 @@
 #include "Theme.h"
 #include "AllThemeResources.h"
 #include "AudioIO.h"
+#include "BasicUI.h"
 #include "Observer.h"
 #include "PluginManager.h"
 #include "Project.h"
@@ -1272,6 +1273,19 @@ public:
       const auto effectID = ShowSelectEffectMenu(dynamic_cast<wxWindow*>(event.GetEventObject()));
       if(effectID.empty())
          return;
+
+      auto plug = PluginManager::Get().GetPlugin(effectID);
+      if(!plug)
+         return;
+
+      if(!PluginManager::IsPluginAvailable(*plug)) {
+         BasicUI::ShowMessageBox(
+            XO("This plugin could not be loaded.\nIt may have been deleted."),
+            BasicUI::MessageBoxOptions()
+               .Caption(XO("Plugin Error")));
+
+         return;
+      }
 
       if(auto state = AudioIO::Get()->AddState(*mProject, &*mTrack, effectID))
       {
--- a/src/effects/EffectUI.cpp
+++ b/src/effects/EffectUI.cpp
@@ -16,6 +16,7 @@
 
 #include "AllThemeResources.h"
 #include "widgets/BasicMenu.h"
+#include "BasicUI.h"
 #include "ConfigInterface.h"
 #include "EffectManager.h"
 #include "PluginManager.h"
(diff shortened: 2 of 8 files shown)


src/RealtimeEffectPanel.cpp
```diff
@@ -1270,6 +1271,15 @@ class RealtimeEffectListWindow : public wxScrolledWindow
       if(effectID.empty())
          return;
 
+      auto plug = PluginManager::Get().GetPlugin(effectID);
+      if(!plug)
+         return;
+
+      if(!PluginManager::IsPluginAvailable(*plug)) {
+         wxMessageBox("This plugin could not be loaded.\nIt may have been deleted.", "Plugin Error");
```
vsverchinsky, line 1279: Please make these strings translatable

src/effects/EffectUI.cpp
```diff
@@ -1231,6 +1232,12 @@ DialogFactoryResults EffectUI::DialogFactory(wxWindow &parent,
    if (!plug)
       return false;
 
+   if (!PluginManager::IsPluginAvailable(*plug))
+   {
+      wxMessageBox("This plugin could not be loaded.\nIt may have been deleted.", "Plugin Error");
```
vsverchinsky, line 1237: Same

vsverchinsky (changes requested):
Dialog message text needs translation

src/RealtimeEffectPanel.cpp
ksoze95: Done.

src/effects/EffectUI.cpp
ksoze95: Done.

vsverchinsky (approved):

src/RealtimeEffectPanel.cpp
```diff
@@ -1270,6 +1270,15 @@ class RealtimeEffectListWindow : public wxScrolledWindow
       if(effectID.empty())
          return;
 
+      auto plug = PluginManager::Get().GetPlugin(effectID);
+      if(!plug)
+         return;
+
+      if(!PluginManager::IsPluginAvailable(*plug)) {
+         wxMessageBox(_("This plugin could not be loaded.\nIt may have been deleted."), _("Plugin Error"));
```
Paul-Licameli, line 1278: Use ::AudacityMessageBox

src/effects/EffectUI.cpp
```diff
@@ -1231,6 +1231,12 @@ DialogFactoryResults EffectUI::DialogFactory(wxWindow &parent,
    if (!plug)
       return false;
 
+   if (!PluginManager::IsPluginAvailable(*plug))
+   {
+      wxMessageBox(_("This plugin could not be loaded.\nIt may have been deleted."), _("Plugin Error"));
```
Paul-Licameli, line 1236: Use ::AudacityMessageBox

Paul-Licameli (changes requested):

src/effects/EffectUI.cpp
Paul-Licameli: Or better, use BasicUI::ShowMessageBox in both places.

src/RealtimeEffectPanel.cpp
ksoze95: Replaced wxMessageBox with BasicUI::ShowMessageBox as suggested below.

src/menus/FileMenus.cpp
```diff
@@ -184,6 +186,41 @@ void OnOpen(const CommandContext &context )
...
+            }
+         });
+   }
+
+   if (unavailablePlugins > 0)
+   {
+      BasicUI::ShowMessageBox(
+         XO("This project contains some plugins that cannot be found on the system."),
+         BasicUI::MessageBoxOptions()
+            .Caption(XO("Project Error")));
```
LWinterberg, line 222: Let's do
> Window title: Missing Plugins
> This project contains some realtime effect plugins that cannot be found on this system.
> The project may sound different than intended. [Learn more](https://audacityteam.org/errors#missing-plugins)
> [OK]

with the "learn more" having an accessible link to https://audacityteam.org/errors#missing-plugins

LWinterberg (approved):

src/effects/EffectUI.cpp
```diff
@@ -1243,6 +1244,16 @@ DialogFactoryResults EffectUI::DialogFactory(wxWindow &parent,
    if (!plug)
       return false;
 
+   if (!PluginManager::IsPluginAvailable(*plug))
```
Paul-Licameli, line 1247: Should this message be shown also in case `plug` is null?

Paul-Licameli (approved):
A few suggestions to improve it, but nothing that should block it.

PR #3804, opened 2022-10-13
Registry changes

Paul-Licameli:
Resolves: #4172

Must do this first before extracting modules for exporters

Recommended:

### DIFF
--- a/libraries/lib-registries/Registry.cpp
+++ b/libraries/lib-registries/Registry.cpp
@@ -16,11 +16,11 @@ Paul Licameli split from Menus.cpp
 
 #include "BasicUI.h"
 
+using namespace Registry;
+using namespace Registry::detail;
+
 namespace {
 struct ItemOrdering;
-
-using namespace Registry;
-using namespace detail;
 
 //! Used only internally
 struct PlaceHolder : GroupItemBase {
@@ -169,8 +169,6 @@ void CollectItem(CollectedItems &collection,
    }
 }
 
-using Path = std::vector< Identifier >;
-
    std::unordered_set< wxString > sBadPaths;
    void BadPath(
      const TranslatableString &format, const wxString &key, const Identifier &name )
@@ -209,10 +207,12 @@ XO("Plug-in items at %s specify conflicting placements"),
          key, name);
    }
 
+using Path = VisitBase::Path;
+
 struct ItemOrdering {
    wxString key;
 
-   ItemOrdering( const Path &path )
+   ItemOrdering(const Path &path)
    {
       // The set of path names determines only an unordered tree.
       // We want an ordering of the tree that is stable across runs.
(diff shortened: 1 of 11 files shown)


libraries/lib-registries/Registry.cpp
```diff
@@ -642,12 +744,11 @@ void VisitItems(
...
          wxString newValue;
-         for ( const auto &item : newCollection.items ) {
-            const auto &name = item.visitNow->name;
+         for ( const auto &name : itemOrdering.ordering ) {
             if ( !name.empty() )
                newValue += newValue.empty()
-                  ? name.GET()
-                  : ',' + name.GET();
+                  ? name
+                  : ',' + name   ;
```
ksoze95, line 751: Trailing spaces after name are not needed.

src/effects/LoadEffects.cpp
```diff
@@ -42,6 +42,20 @@ void BuiltinEffectsModule::DoRegistration(
    Entry::Registry().emplace_back( Entry{ name, factory, excluded } );
 }
 
+void BuiltinEffectsModule::UndoRegistration(
+   const ComponentInterfaceSymbol &name )
+{
+   auto &items = Entry::Registry();
+   auto end = items.end();
+   auto iter = std::find_if( items.begin(), end, [&](auto &entry){
```
ksoze95, line 50: The lambda should be changed to 'const auto&'

src/commands/CommandFlag.h
```diff
@@ -123,4 +126,6 @@ struct AUDACITY_DLL_API RegisteredMenuItemEnabler{
 // Unnecessary #include to indicate otherwise hidden link dependencies
 #include "Menus.h"
 
+static ReservedCommandFlag::Init sInitReservedCommandFlags;
```
ksoze95, line 129: Could this be a potential problem?  As there could be possible multiple instances of these being created, each belonging to different translation units.

libraries/lib-registries/Registry.h
```diff
@@ -26,25 +26,41 @@ namespace Registry {
...
       Identifier name;
 
       OrderingHint() {}
-      OrderingHint( Type type_, const wxString &name_ = {} )
-         : type(type_), name(name_) {}
+      OrderingHint( Type type,
+         const wxString &name = {}, ConflictResolutionPolicy policy = Error )
+         : type{ type }, name{ name }, policy{ policy } {}
 
+      //! The conflict resoution policy is not significant in equality
```
ksoze95, line 54: Spelling mistake on resolution

libraries/lib-registries/Registry.h
```diff
@@ -26,25 +26,41 @@ namespace Registry {
...
+      //! comparison
       bool operator == ( const OrderingHint &other ) const
       { return name == other.name && type == other.type; }
 
+      //! The conflict resoution policy is not significant in ordering
```
ksoze95, line 59: Spelling mistake on resolution

Issue #3233, opened 2022-07-14
Redundant blank space around some plugin windows (AU, VST, VST3)


dozzzzer:
**Describe the bug**
There's redundant blank space around some VST/VST3/AU plugins which makes the dialog window somewhat larger than it should.

**To Reproduce**
1. Select audio, go to Effect
2. Select a VST/VST3/AU plugin, e.g. Valhalla Room, Kilhearts plugins

**Actual behavior**
There's some excessive space above and under the plugin as well as on the left/ride sides of it:
<img width="600" alt="image" src="https://user-images.githubusercontent.com/70063970/179004436-30263a02-f853-4483-a6d0-f1261ed0a2e1.png">

**Expected behavior**
The plugin window doesn't have any additional blank space around.

**Additional information (please complete the following information):**
 - OS: MacOS 12.2.1, Windows 10
 - latest master build

LWinterberg:
This doesn't appear for me with the new design on Windows 11: 
![image](https://user-images.githubusercontent.com/87814144/191721256-9c10e407-4f72-4c4d-914e-fba11e82b791.png)

That said: The grabber that's supposed to resize the plugin window doesn't actually resize the plugin window and just resizes the internal thing. The actual window cannot be resized
![image](https://user-images.githubusercontent.com/87814144/191721425-74fa5e9e-b367-4870-8917-3e2b6a21a945.png)

vsverchinsky:
@dozzzzer @Ryan-Miller-Audacity Please note that plugin dialog resize behaviour on Linux and MacOS (for any type of plugin) can't be fixed with reasonable efforts, the problem is that system dialogs have their own resize handle (which sometimes overlap plugins own resize handle), and that can't be changed on these systems. That means that there is no way to completely avoid blank spaces during resize. #3805 removes blank spaces for VST3 when you open the plugin UI and also disables system resize handle on Windows platform.

dozzzzer:
Verified. This is now fixed according to @vsverchinsky 's comment. 
No blank space anymore on VST3s on Windows with the ability to properly resize using the plugin's handle:

![image](https://user-images.githubusercontent.com/70063970/196169678-dc71b97c-7536-4763-aa10-c9b24d7dbb13.png)

PR #3805, opened 2022-10-13
VST3 better native UI size handling 

vsverchinsky:
Resolves: #3233

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Effect.cpp
+++ b/src/effects/VST3/VST3Effect.cpp
@@ -236,14 +236,17 @@ bool VST3Effect::LoadFactoryPreset(int id, EffectSettings& settings) const
 }
 
 int VST3Effect::ShowClientInterface(wxWindow& parent, wxDialog& dialog,
-   EffectUIValidator *, bool forceModal)
+   EffectUIValidator *validator, bool forceModal)
 {
-   if(!IsGraphicalUI())
-   {
-      //Restrict resize of the "plain" dialog
-      dialog.SetMaxSize(dialog.GetSize());
-      dialog.SetMinSize(dialog.GetSize());
-   }
+#ifdef __WXMSW__
+   if(validator->IsGraphicalUI())
+      //Not all platforms support window style change.
+      //Plugins that support resizing provide their own handles,
+      //which may overlap with system handle. Not all plugins
+      //support free sizing (e.g. fixed steps or fixed ratio)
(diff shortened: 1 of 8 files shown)


src/effects/VST3/internal/PlugFrame.cpp
```diff
@@ -23,26 +24,32 @@ internal::PlugFrame::~PlugFrame()
...
 {
    if(auto window = mWindow.get())
    {
-      auto fixed = view->canResize() != Steinberg::kResultTrue;
-      UpdateWindowSize(window, {newSize->getWidth(), newSize->getHeight()}, fixed);
+      auto topWindow = wxGetTopLevelParent(window);
+
+      wxWindowUpdateLocker windowUpdateLocker(topWindow);
+
+      window->SetInitialSize({newSize->getWidth(), newSize->getHeight() });
```
Paul-Licameli, line 41: A deleted comment said
```
           //No need to accommodate the off-by-one error with Steinberg::ViewRect
           //as we do it with wxRect
```

Are you sure there are no implications of off-by-one errors here in the rewrite?

Paul-Licameli (approved):
I am not expert in the details of X11 UI code but I have a general understanding that certain settings of window sizes need to happen earlier.  No serious objections, but one suggestion for clarifying the use of bitwise operations.

src/effects/VST3/VST3Effect.cpp
```diff
@@ -236,14 +236,11 @@ bool VST3Effect::LoadFactoryPreset(int id, EffectSettings& settings) const
...
+   EffectUIValidator *validator, bool forceModal)
 {
-   if(!IsGraphicalUI())
-   {
-      //Restrict resize of the "plain" dialog
-      dialog.SetMaxSize(dialog.GetSize());
-      dialog.SetMinSize(dialog.GetSize());
-   }
+   if(validator->IsGraphicalUI())
+      dialog.SetWindowStyle(dialog.GetWindowStyle() ^ (wxRESIZE_BORDER | wxMAXIMIZE_BOX));
```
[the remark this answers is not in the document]
vsverchinsky: Yes, it's not a correct expression, thanks!

src/effects/VST3/internal/PlugFrame.cpp
vsverchinsky: Where is that error?

Issue #3830, opened 2022-10-20
Preliminaries for VST2 statelessness


Paul-Licameli:
Need ways to communite EffectInstance::Message to other threads in more cases: importing effect settings or switching among factory or user presets.

Don't need more changes in RealtimeEffectState.cpp.

QA:  should do sanity checks on effects of each family in the sidebar

PR #3831, opened 2022-10-20
More effect messages

Paul-Licameli:
Resolves: #3830

optional<unique_ptr<EffectSettingsAccess::Message>>, not boolean, returned now from
more virtual functions.

The overrides return {} for failure and { nullptr } for success.

Later some overrides may return a non-empty optional with a non-null unique_ptr in it which
will then be sent across threads.

Recommended:


Paul-Licameli:
#3886 supersedes this

PR #3832, opened 2022-10-20
Vst2 fix metering

pietro68:
Resolves: metering for vst2

Recommended:


Paul-Licameli:
Before anything, please rebase this onto latest release-3.2.2 so the number of commits isn't 74!

src/effects/VST/VSTEffect.cpp
```diff
@@ -3790,9 +3790,15 @@ void VSTEffectWrapper::SizeWindow(int w, int h)
 
 void VSTEffectWrapper::Automate(int index, float value)
 {
+   // This is now called only for the instance;
+   // TODO (?) we'd need to copy here part of what is done in VSTEffectValidator::Automate
+   // - we need to send back to the framework the new value for the parameter
```
Paul-Licameli, line 3791: The commit that adds this comment is making a promise to fix things in a later commit.  I'm still reviewing to see whether that happens.

src/effects/VST/VSTEffect.h
```diff
@@ -100,7 +100,14 @@ struct VSTEffectSettings
 };
 
 
-struct VSTEffectWrapper : public VSTEffectLink, public XMLTagHandler
+struct VSTEffectUIWrapper
+{
+   virtual void NeedIdle();
+   virtual void SizeWindow(int w, int h);
```
Paul-Licameli, line 106: I think this interface is insufficient.  I think `NeedEditIdle()` and `Automate()` must be part of it too.

Paul-Licameli:
Another omission so far -- not specific to metering, but needed for a proper implementation of the validator -- is to rewrite the OnSlider event handler for plain UI.

It should contain a call to `ModifySettings` which makes a message from `i` and the slider value and returns that.

It should not do `callSetParameter` directly on the instance.  Let that happen in `RealtimeProcessStart` on the other thread.

src/effects/VST/VSTEffect.cpp
pietro68: ok, done now

src/effects/VST/VSTEffect.h
pietro68: now done

pietro68:
> 



> Another omission so far -- not specific to metering, but needed for a proper implementation of the validator -- is to rewrite the OnSlider event handler for plain UI.
> 
> It should contain a call to `ModifySettings` which makes a message from `i` and the slider value and returns that.
> 
> It should not do `callSetParameter` directly on the instance. Let that happen in `RealtimeProcessStart` on the other thread.

Your request makes sense, but I think that to do that I'd need a scratch instance and call callSetParameter on it, to be able to correctly update the plainUI.

Why:
Imagine bringing up the GUI and tweaking the "Enable" slider on DeeGain - either when doing destructive processing or when doing realtime processing at play stopped.

As I do this, I expect to see the enable value switch between "On" and "Off", and those strings come from an mAEffect. They are hardcoded in there, and depend on the value which is set to the mAEffect.

Paul-Licameli:
Superseded by #3886 which contains all the same code but a simplified commit history

Issue #3670, opened 2022-09-22
User isn't asked to save changes when tweaking parameters of the effect and then quitting


dozzzzer:
### Steps to reproduce

1. open an existing project which has an RT effect added to the track
2. open the RT effect's dialog
3. tweak some parameters
4. Cmd + Q
5. observe

### Expected behavior

The user is asked to save the changes (Yes/No)

### Actual behavior

The project window is closed right away without asking the user to save the changes

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

Ryan-Miller-Audacity:
Confirmed that users are now asked to save project when attempting to quick exit Audacity after making tweaks to effect parameters. This was tested on both MacOS and Windows 10 with all RT effect families.

PR #3836, opened 2022-10-24
Adds ability to notify about the parameter changes for RT capable effects

crsib:
Resolves: #3670 

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -52,6 +52,8 @@
 #include <type_traits>
 #include <wx/event.h>
 
+#include "Observer.h"
+
 class ShuttleGui;
 template<bool Const> class SettingsVisitorBase;
 using SettingsVisitor = SettingsVisitorBase<false>;
@@ -627,6 +629,18 @@ public:
 
 };
 
+/*************************************************************************************/ /**
+
+ \class EffectSettingChanged
+
+ \brief Message sent by validator when a setting is changed by a user
+
+ *******************************************************************************************/
+struct COMPONENTS_API EffectSettingChanged final
+{
+   size_t index { size_t(-1) };
+   float newValue {};
+};
 /*************************************************************************************//**
 
 \class EffectUIValidator
@@ -635,6 +649,7 @@ public:
 
 *******************************************************************************************/
 class COMPONENTS_API EffectUIValidator /* not final */
+    : public Observer::Publisher<EffectSettingChanged>
 {
 public:
    EffectUIValidator(
--- a/src/effects/RealtimeEffectStateUI.cpp
+++ b/src/effects/RealtimeEffectStateUI.cpp
@@ -108,6 +108,9 @@ void RealtimeEffectStateUI::Show(AudacityProject& project)
          // project state is destroyed
          Hide(&project);
       });
+
+   mParameterChangedSubscription = mEffectUIHost->GetValidator()->Subscribe(
+      [this](auto) { ProjectHistory::Get(*mpProject).ModifyState(false); });
 }
 
 void RealtimeEffectStateUI::Hide(AudacityProject* project)
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -2321,6 +2321,30 @@ void VSTEffectInstance::SizeWindow(int w, int h)
    }
 }
 
+void VSTEffectValidator::NotifyParameterChanged(int index, float value)
+{
+   const auto& settings = VSTEffectWrapper::GetSettings(mAccess.Get());
+
+   GetInstance().ForEachParameter(
+      [index, value, &settings, this](const auto& pi)
+      {
+         if (pi.mID != index)
+            return true;
+
+         auto it = settings.mParamsMap.find(pi.mName);
+
+         // For consistency with other plugin families
+         constexpr float epsilon = 1.0e-5f; 
+
+         if (
+            it == settings.mParamsMap.end() || !it->second.has_value() ||
+            std::abs(*it->second - value) > epsilon)
+            Publish(EffectSettingChanged { size_t(index), value });
+
+         return false;
+      });
+}
+
 void VSTEffectValidator::OnIdle(wxIdleEvent& evt)
 {
    evt.Skip();
@@ -2772,6 +2796,7 @@ void VSTEffectValidator::OnSlider(wxCommandEvent & evt)
    int i = s->GetId() - ID_Sliders;
    float value = s->GetValue() / 1000.0;
 
+   NotifyParameterChanged(i, value);
    // Send changed settings (only) to the worker thread
    mAccess.ModifySettings([&](EffectSettings&) {
       auto result = GetInstance().MakeMessage(i, value);
@@ -3940,6 +3965,7 @@ void VSTEffectInstance::Automate(int index, float value)
 
 void VSTEffectValidator::Automate(int index, float value)
 {
+   NotifyParameterChanged(index, value);
    // Send changed settings (only) to the worker thread
    mAccess.ModifySettings([&](EffectSettings&) {
       auto result = GetInstance().MakeMessage(index, value);
--- a/src/effects/VST3/VST3UIValidator.cpp
+++ b/src/effects/VST3/VST3UIValidator.cpp
@@ -54,6 +54,12 @@ VST3UIValidator::VST3UIValidator(wxWindow* parent, VST3Wrapper& wrapper, EffectB
          *mWrapper.mComponentHandler
       );
    }
+
+   mWrapper.ParamChangedHandler =
+      [this](Steinberg::Vst::ParamID id, Steinberg::Vst::ParamValue value) {
+         Publish({ static_cast<size_t>(id), static_cast<float>(value) });
+      };
+   
    mWrapper.BeginParameterEdit(mAccess);
 
    Bind(wxEVT_IDLE, &VST3UIValidator::OnIdle, this);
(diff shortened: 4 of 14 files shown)


libraries/lib-components/EffectInterface.h
```diff
@@ -615,6 +617,18 @@ class COMPONENTS_API EffectInstanceFactory
...
+/*************************************************************************************/ /**
+
+ \class EffectSettingChanged
+
+ \brief Message sent by validator when a setting is changed by a user
+
+ *******************************************************************************************/
+struct COMPONENTS_API EffectSettingChanged final
+{
+   size_t index { size_t(-1) };
```
Paul-Licameli, line 629: I might write `~size_t{}` instead

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -108,6 +108,9 @@ void RealtimeEffectStateUI::Show(AudacityProject& project)
          // project state is destroyed
          Hide(&project);
       });
+
+   mParameterChangedSubscription = mEffectUIHost->GetValidator()->Subscribe(
+      [this](auto) { ProjectHistory::Get(*mpProject).ModifyState(false); });
```
Paul-Licameli, line 113: This subscriber doesn't use the value inside the message.  Is there supposed to be another subscriber?

src/effects/ladspa/LadspaEffect.cpp
```diff
@@ -1803,3 +1802,16 @@ void LadspaEffect::Validator::RefreshControls()
...
    }
 }
+
+void LadspaEffect::Validator::UpdateControl(int index, float value)
+{
+   auto& controls = mSettings.controls;
+
+   assert(index < static_cast<int>(controls.size()));
+
+   if (std::abs(controls[index] - value) < std::numeric_limits<float>::epsilon())
```
Paul-Licameli, line 1812: Is this tolerance test necessary, and if so, is it correct?

"Returns the machine epsilon, that is, the difference between 1.0 and the next value representable by the floating-point type T."

But possibly some control is scaled to an interval of small fractional values.

All the Ladspa sliders are graded to 1000 steps.  Maybe the epsilon should be 1/1000 times the difference of maximum and minimum values of the control.

src/effects/lv2/LV2Validator.cpp
```diff
@@ -705,12 +705,24 @@ void LV2Validator::SetSlider(
    ctrl.slider->SetValue(lrintf((val - lo) / (hi - lo) * 1000.0));
 }
 
+void LV2Validator::UpdateControlPortValue(
+   LV2EffectSettings& settings, size_t controlPortIndex, float value)
+{
+   const auto currentValue = settings.values[controlPortIndex];
+
+   if (std::abs(currentValue - value) < std::numeric_limits<float>::epsilon())
```
Paul-Licameli, line 713: Similar concerns here.  Some controls are discrete -- toggles or choices.  Maybe the right epsilon for the control should be an argument of the function.

src/effects/ladspa/LadspaEffect.cpp
Paul-Licameli: Or rather, make that 1/2000 -- one-half of the interval.

src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -163,6 +169,8 @@ void VST3UIValidator::OnClose()
    //Make sure that new state has been written to the caches...
    mAccess.Flush();
 
+   mWrapper.ParamChangedHandler = {};
```
Paul-Licameli, line 172: This line defends against a dangling pointer.  I would duplicate it also in the destructor of validator, just to be very certain it is never missed in exceptional situations.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -222,6 +222,10 @@ class ComponentHandler : public Steinberg::Vst::IComponentHandler
    
    EffectSettings* mStateChangeSettings {nullptr};
    std::map<Steinberg::Vst::ParamID, Steinberg::Vst::ParamValue> mParametersCache;
+   std::map<Steinberg::Vst::ParamID, Steinberg::Vst::ParamValue> mCurrentParamValues;
+
+   bool mInStateChange { false };
```
Paul-Licameli, line 227: I think this variable isn't needed, and you can just use `mStateChangeSettings != nullptr`

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -265,13 +292,34 @@ class ComponentHandler : public Steinberg::Vst::IComponentHandler
...
+      Steinberg::Vst::ParamID id, Steinberg::Vst::ParamValue valueNormalized)
+   {
+      auto it = mCurrentParamValues.find(id);
+
+      if (it == mCurrentParamValues.end())
+         return;
+
+      constexpr Steinberg::Vst::ParamValue epsilon = 1e-6;
+
+      if (std::abs(it->second - valueNormalized) < epsilon)
```
Paul-Licameli, line 305: Another epsilon.  I don't know how to find the range of values for it.

src/effects/audiounits/AudioUnitValidator.cpp
```diff
@@ -204,6 +204,23 @@ void AudioUnitValidator::EventListener(const AudioUnitEvent *inEvent,
...
+
+   constexpr AudioUnitParameterValue epsilon = 1e-6;
+
+   auto it = mParameterValues.find(inEvent->mArgument.mParameter.mParameterID);
+
+   // When the UI is opened - EventListener is called for each parameter
+   // with the current value.
+   if (it == mParameterValues.end())
+      mParameterValues.insert(std::make_pair(inEvent->mArgument.mParameter.mParameterID, inParameterValue));
+   else if (std::abs(it->second - inParameterValue) > epsilon)
```
Paul-Licameli, line 219: `AudioUnitWrapper::ParameterInfo::mInfo`, as supplied to the lambda argument of `ForEachParameter`, has members minValue, maxValue

Paul-Licameli:
I'm done reviewing this.

I understand it is incomplete, pending some solution for VST 2 statelessness.

The built-in effects that are real-time capable will also need changes in 3.3.

src/effects/ladspa/LadspaEffect.cpp
crsib: After careful consideration I don't think that using sliders step as a delta is correct. For LADSPA machine epsilon appears to be a good approximation, but for consistency I will define it to match the one LV2. For LV2 epsilon should be larger (but still - it is unrelated to amount of steps we define for sliders), because we allow text input in the plain UI (and the epsilon is only used in plain UI). 

For VST3 I know plugins, that require epsilon around 10^-6

src/effects/RealtimeEffectStateUI.cpp
Paul-Licameli: Still unresolved question?

src/effects/ladspa/LadspaEffect.cpp
```diff
@@ -1731,13 +1739,13 @@ void LadspaEffect::Validator::OnSlider(wxCommandEvent & evt)
       str = Internat::ToDisplayString(val);
 
    mFields[p]->SetValue(str);
-   controls[p] = val;
+   // Use half of the range as the tolerance level
```
Paul-Licameli, line 1742: Wrong comment

src/effects/RealtimeEffectStateUI.cpp
crsib: I don't know what to say. I can make the structure empty if this will make you happier, I don't care

src/effects/VST/VSTEffect.cpp
```diff
@@ -2304,6 +2304,30 @@ void VSTEffectInstance::SizeWindow(int w, int h)
...
+            return true;
+
+         auto it = settings.mParamsMap.find(pi.mName);
+
+         // For consistency with other plugin families
+         constexpr float epsilon = 1.0e-5f; 
+
+         if (
+            it == settings.mParamsMap.end() || !it->second.has_value() ||
+            std::abs(it->second->second - value) > epsilon)
```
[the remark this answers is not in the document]
crsib: I am confused why it was building locally

src/effects/ladspa/LadspaEffect.cpp
```diff
@@ -1209,6 +1214,10 @@ struct LadspaEffect::Validator : EffectUIValidator {
    void OnTextCtrl(wxCommandEvent & evt);
    void RefreshControls();
 
+   void UpdateControl(int index, float value, float epsilon);
+   void
+   UpdateControls(LadspaEffectSettings& dst, const LadspaEffectSettings& src);
```
Paul-Licameli, line 1219: `dst` is now unused

src/effects/lv2/LV2Validator.cpp
```diff
@@ -842,7 +866,8 @@ void LV2Validator::OnIdle(wxIdleEvent &evt)
  */
                &value);
             state.mLst = value;
-         }
+            Publish({ size_t(port->mIndex), value });
```
Paul-Licameli, line 869: Might publish unnecessarily for output port changes

Paul-Licameli: Maybe `suil_port_write` is the function to change, not the idle handler?

src/effects/VST/VSTEffect.cpp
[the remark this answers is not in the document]
Paul-Licameli: You had not rebased yet onto latest?

Paul-Licameli:
Done, no more objections.

Fix the build and ask for the last round (we hope)

Paul-Licameli (approved):

Issue #3838, opened 2022-10-24
Plugin scan dialog improvements


vsverchinsky:
1. Add `&Skip` button next to plugin name label
2. Convert plugin name label to ~multiline~ **elapsized** label, so that dialog does not grow horizontally and stays fixed 500px in width. 
3. Remove 'Remaining Rime'

UPD: no way to reliably wrap text (with breaks in the middle of the word) in wx

Ryan-Miller-Audacity:
Tested via the build from PR #3850. Tested both initial scan as well as rescan.

1. Skip button is now included in plugin scan window and properly skips over plugins as clicked.
2. Plugin names are truncated properly and keeps the window display at a consistent 500px in width.
3. Remaining time counter has been removed from the dialog; elapsed time remains and works as expected.

Tested on Windows 10 and MacOS covering all plugin families.

SteveDaulton:
> 1. Skip button is now included in plugin scan window and properly skips over plugins as clicked.


Testing with audacity-linux-3.2.2-x64.AppImage

Where is the "skip button"? Is it that line in the bottom right corner? It does not appear to be clickable.

![Screenshot_2022-12-09_16-35-34](https://user-images.githubusercontent.com/11681097/206749560-e309e7d8-9934-4f1c-915c-186bb2a62085.png)

Update: Yes it is clickable if the mouse pointer is in exactly the right place (it took a few attempts to activate it).

DavidBailes:
> Where is the "skip button"? Is it that line in the bottom right corner? It does not appear to be clickable.

Here's dialog on Windows:
![image](https://user-images.githubusercontent.com/11887021/206751121-01248225-b4fa-45a8-8838-57726e042883.png)

(Skip button it top right)

PR #3850, opened 2022-10-26
Plugin scan dialog improvements

vsverchinsky:
Resolves: #3838

Recommended:

### DIFF
--- a/src/PluginStartupRegistration.cpp
+++ b/src/PluginStartupRegistration.cpp
@@ -35,17 +35,29 @@ void PluginStartupRegistration::OnInternalError(const wxString& error)
 
 void PluginStartupRegistration::OnPluginFound(const PluginDescriptor& desc)
 {
-   //Multiple providers can report same module paths
-   if(desc.GetPluginType() == PluginTypeStub)
-      //do not register until all associated providers have tried to load the module
+   if(!mValidProviderFound)
+      mFailedPluginsCache.clear();
+
+   mValidProviderFound = true;
+   if(!desc.IsValid())
       mFailedPluginsCache.push_back(desc);
-   else
-   {
-      mValidProviderFound = true;
-      if(!desc.IsValid())
-         mFailedPluginsCache.push_back(desc);
-      PluginManager::Get().RegisterPlugin(PluginDescriptor { desc });
-   }
+   PluginManager::Get().RegisterPlugin(PluginDescriptor { desc });
+}
+
+void PluginStartupRegistration::OnPluginValidationFailed(const wxString& providerId, const wxString& path)
+{
+   PluginID ID = providerId + wxT("_") + path;
+   PluginDescriptor pluginDescriptor;
+   pluginDescriptor.SetPluginType(PluginTypeStub);
+   pluginDescriptor.SetID(ID);
+   pluginDescriptor.SetProviderID(providerId);
+   pluginDescriptor.SetPath(path);
+   pluginDescriptor.SetEnabled(false);
+   pluginDescriptor.SetValid(false);
+
+   //Multiple providers can report same module paths
+   //do not register until all associated providers have tried to load the module
+   mFailedPluginsCache.push_back(std::move(pluginDescriptor));
 }
 
 void PluginStartupRegistration::OnValidationFinished()
--- a/src/PluginStartupRegistration.cpp
+++ b/src/PluginStartupRegistration.cpp
@@ -14,18 +14,123 @@
 
 #include <wx/log.h>
 #include <wx/app.h>
+#include <wx/stattext.h>
+#include <wx/gauge.h>
+#include <wx/button.h>
+#include <wx/timer.h>
+#include <wx/sizer.h>
 
 #include "PluginManager.h"
 #include "PluginDescriptor.h"
-#include "widgets/ProgressDialog.h"
-#include "widgets/wxWidgetsWindowPlacement.h"
+#include "widgets/wxPanelWrapper.h"
 
+namespace
+{
+   enum {
+      OnPluginScanTimeout = wxID_HIGHEST + 1,
+   };
+
+   class PluginScanDialog : public wxDialogWrapper
+   {
+      wxStaticText* mText{nullptr};
+      wxStaticText* mElapsed{nullptr};
+      wxGauge* mProgress{nullptr};
+      std::chrono::system_clock::time_point mStartTime;
+   public:
+      
+      PluginScanDialog(
+         wxWindow* parent,
+         wxWindowID winid,
+         const TranslatableString& title)
+         : wxDialogWrapper(parent, winid, title)
+      {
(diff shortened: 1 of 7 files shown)


src/PluginStartupRegistration.cpp
```diff
@@ -14,12 +14,118 @@
...
+      std::chrono::system_clock::time_point mStartTime;
+   public:
+      
+      PluginScanDialog(
+         wxWindow* parent,
+         wxWindowID winid,
+         const TranslatableString& title)
+         : wxDialogWrapper(parent, winid, title)
+      {
+         auto rootSizer = std::make_unique<wxBoxSizer>(wxVERTICAL);
```
Paul-Licameli, line 47: If the schedule were more relaxed, I would ask you to learn the use of ShuttleGui rather than making so many direct calls to wxSizer methods.  But the task of rewriting this will wait until another release and perhaps be someone else's problem.

src/PluginStartupRegistration.h
```diff
@@ -34,13 +32,19 @@ class PluginStartupRegistration final :
    bool mValidProviderFound{false};
    std::vector<wxString> mFailedPluginsPaths;
    std::vector<PluginDescriptor> mFailedPluginsCache;
+   wxDialogWrapper* mScanDialog{nullptr};
+   wxTimer* mTimeoutTimer{nullptr};
+   std::chrono::system_clock::duration mTimeout;
+   std::chrono::system_clock::time_point mRequestStartTime;
```
Paul-Licameli, line 38: I would add `{}` to the two members above

src/PluginStartupRegistration.cpp
```diff
@@ -100,31 +208,68 @@ const std::vector<wxString>& PluginStartupRegistration::GetFailedPluginsPaths()
...
-      //Update UI
-      if(dialog->Poll(mCurrentPluginIndex, mPluginsToProcess.size(), message) != BasicUI::ProgressResult::Success)
-         Stop();
-      //AsyncPluginValidator uses event loop for internal message
-      //delivery, but ProgressDialog::Poll implementation does not call
-      //wxApp::Yield each time, which may result in too long CPU stalls
-      if(wxTheApp->HasPendingEvents())
-         wxTheApp->Yield();
+      mTimeoutTimer = nullptr;
+      mScanDialog = nullptr;
```
Paul-Licameli, line 222: Maybe wxWeakRef<> pointer members can allow simplification

src/PluginStartupRegistration.cpp
```diff
@@ -100,31 +208,68 @@ const std::vector<wxString>& PluginStartupRegistration::GetFailedPluginsPaths()
...
+      if(evt.GetId() == wxID_IGNORE)
+         Skip();
+   });
+   dialog.Bind(wxEVT_TIMER, [this](wxTimerEvent& evt) {
+      if(evt.GetId() == OnPluginScanTimeout)
+      {
+         if(mValidator->InactiveSince() < mRequestStartTime)
+            Skip();
+         //else
+         //   wxMessageBox("Please check for plugin popups!");
```
Paul-Licameli, line 236: What's the activity check really doing if there is no else branch?  But I'm still reviewing the second commit, so let me see.

Paul-Licameli: I understand better now, I see where mRequestStartTime is assigned.

src/PluginStartupRegistration.cpp
```diff
@@ -100,31 +208,68 @@ const std::vector<wxString>& PluginStartupRegistration::GetFailedPluginsPaths()
...
+
+   if(!mValidProviderFound)
+   {
+      for(;mCurrentPluginProviderIndex < mPluginsToProcess[mCurrentPluginIndex].second.size(); ++mCurrentPluginProviderIndex)
+         OnPluginValidationFailed(
+            mPluginsToProcess[mCurrentPluginIndex].second[mCurrentPluginProviderIndex],
+            mPluginsToProcess[mCurrentPluginIndex].first);
+      mCurrentPluginProviderIndex = mPluginsToProcess[mCurrentPluginIndex].second.size() - 1;
+   }
+   OnValidationFinished();
```
Paul-Licameli, line 272: This function is the most difficult thing to understand in the entire pull request.  I don't think I do, yet.

It should be commented that the `OnValidationFinished` call supplies what would otherwise be an idle-time callback if the Skip button was not pressed.

src/PluginStartupRegistration.cpp
```diff
@@ -100,31 +208,68 @@ const std::vector<wxString>& PluginStartupRegistration::GetFailedPluginsPaths()
...
+void PluginStartupRegistration::Skip()
+{
+   mValidator->SetDelegate(nullptr);
+   std::thread([validator = std::shared_ptr<AsyncPluginValidator>(std::move(mValidator))]{ }).detach();
+   mValidator = std::make_unique<AsyncPluginValidator>(*this);
+
+   if(!mValidProviderFound)
+   {
+      for(;mCurrentPluginProviderIndex < mPluginsToProcess[mCurrentPluginIndex].second.size(); ++mCurrentPluginProviderIndex)
+         OnPluginValidationFailed(
```
Paul-Licameli, line 267: I see that this call, also, simulates the asynchronous messages that are skipped.  So please comment that too.

Paul-Licameli:
Done with this round.  Nothing seriously wrong, but please comment some things and consider the other suggestions.

src/PluginStartupRegistration.cpp
```diff
@@ -14,12 +14,118 @@
...
+         }
+      }
+      
+      void OnIdle(wxIdleEvent& evt)
+      {
+         evt.Skip();
+         Update();
+      }
+
+      void Update()
```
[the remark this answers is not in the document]
vsverchinsky: Didn't notice that, thanks!

src/PluginStartupRegistration.cpp
```diff
@@ -88,31 +207,75 @@ const std::vector<wxString>& PluginStartupRegistration::GetFailedPluginsPaths()
...
+      // Note: create stub `PluginDescriptors` for each associated provider
+      for(;mCurrentPluginProviderIndex < mPluginsToProcess[mCurrentPluginIndex].second.size(); ++mCurrentPluginProviderIndex)
+         OnPluginValidationFailed(
+            mPluginsToProcess[mCurrentPluginIndex].second[mCurrentPluginProviderIndex],
+            mPluginsToProcess[mCurrentPluginIndex].first);
+      mCurrentPluginProviderIndex = mPluginsToProcess[mCurrentPluginIndex].second.size() - 1;
+   }
+   //else
+   //    Don't assume that `OnValidationFinished()` and `OnPluginFound()`
+   //    aren't deferred in within run loop
```
Paul-Licameli, line 276: "deferred within"

Paul-Licameli (approved):

PR #3870, opened 2022-10-28
Open ended toolbars

Paul-Licameli:
Resolves:

To prepare for extraction of individual toolbars into separate modules, leaving the
general toolbar system itself open-ended: eliminate the enumeration of toolbar types.

There is still mention of toolbar types, but only by string identifiers describing the default
configuration, and it is allowed for the lookup of one of the names to fail.

There was already a registration of toobar types but not yet complete open-endedness.
This isn't really complete yet either -- ideally toolbars should register their own preferred
placements with hints, as do the menu items.

But review is invited for the existing commits which are not likely to change.

Recommended:

### DIFF
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -552,7 +552,8 @@ void ProjectAudioManager::Stop(bool stopStream /* = true*/)
       }
    }
 
-   const auto toolbar = ToolManager::Get( *project ).GetToolBar(ScrubbingBarID);
+   // To do: eliminate this, use an event instead
+   const auto toolbar = ToolManager::Get( *project ).GetToolBar(wxT("Scrub"));
    if (toolbar)
       toolbar->EnableDisableButtons();
 }
--- a/src/cloud/ShareAudioToolbar.cpp
+++ b/src/cloud/ShareAudioToolbar.cpp
@@ -35,8 +35,13 @@ IMPLEMENT_CLASS(cloud::ShareAudioToolbar, ToolBar);
 
 namespace cloud
 {
+Identifier ShareAudioToolbar::ID()
+{
+   return wxT("Share Audio");
+}
+
 ShareAudioToolbar::ShareAudioToolbar(AudacityProject& project)
-    : ToolBar(project, ShareAudioBarID, XO("Share Audio"), wxT("Share Audio"))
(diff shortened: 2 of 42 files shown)


src/cloud/ShareAudioToolbar.cpp
```diff
@@ -35,8 +35,13 @@ IMPLEMENT_CLASS(cloud::ShareAudioToolbar, ToolBar);
 
 namespace cloud
 {
+Identifier ShareAudioToolbar::ID()
```
ksoze95, line 38: Could this be made const

src/toolbars/ToolManager.h
```diff
@@ -72,15 +72,15 @@ class AUDACITY_DLL_API ToolManager final
 
    void LayoutToolBars();
 
-   bool IsDocked( int type );
+   bool IsDocked( Identifier type );
```
ksoze95, line 75: Could we make this and IsVisible const

src/ProjectAudioManager.cpp
```diff
@@ -552,7 +552,8 @@ void ProjectAudioManager::Stop(bool stopStream /* = true*/)
       }
    }
 
-   const auto toolbar = ToolManager::Get( *project ).GetToolBar(ScrubbingBarID);
+   // To do: eliminate this, use an event instead
+   const auto toolbar = ToolManager::Get( *project ).GetToolBar(wxT("Scrub"));
```
ksoze95, line 556: Should the GetToolBar call ScrubberToolBar::ID() instead of wxT("Scrub")?

Paul-Licameli: No:  that would make an undesirable dependency cycle.

The Scrubbing toolbar (and each other toolbar) should be easily extractible as a module.  Such a cycle would prohibit that.

Lookups by string, which may fail, are a weaker sort of dependency.

Notice the null check on toolbar and also my new comment.

A lot of old scrubbing code (which I wrote, years ago) makes some difficult tangling in the core application that I must still figure out.

src/toolbars/ToolManager.h
Paul-Licameli: Done

ksoze95 (approved):

dozzzzer:
Verified

PR #3871, opened 2022-10-28
Effect modules

Paul-Licameli:
Resolves: #1131
Resolves: #1119
Resolves: #1118
Resolves: #1117
Resolves: #1116
Resolves: #1115
Resolves: #1098

Seven modules for support of effect plug-in families and for the several built-in effects

The Audacity executable has five fewer dependencies on third-party libararies, as seen in this diagram:

image

Recommended:


vsverchinsky (approved):
Now there are two modules containing builtin effects: `mod-soundtouch` and `mod-effect-classes`, the difference is that effects from `mod-soundtouch` require optional third-party libraries. Other modules bring effect families support. So my suggestion would be to put all builtin effects into a single `mod-effect-classes` module and conditionally include effects depending on whether required feature/option/library is enabled.

I don't see much problems with `StereoToMono` being an effect, though it could be non-optional

crsib:
I haven't reviewed yet, please do not merge

Paul-Licameli:
> I haven't reviewed yet, please do not merge

I'm surely not hurrying this

scripts/build/macOS/fixup_libs.py
```diff
@@ -6,6 +6,20 @@
 import os
 import shutil
 
+def find_target(path):
```
crsib, line 9: Even though I am not going to approve the first commit at all, I want to stress that naming is very important.

Why is called `find`? It does not perform any kind of searching.

Paul-Licameli: Then call it `parse_path` if you like.

Given a path, deduce the target name.

Issue #3869, opened 2022-10-28
Screen refresh rate during play while VST2 dialogs are open should be acceptable


Paul-Licameli:
### Steps to reproduce

1. Play a track while the dialog for a stacked VST 2 effect is open (either plain or fancy UI)
2. Move sliders while playing
3. Observe movement of the green play head

### Expected behavior

In all cases, no glitches in the sound, and smooth movement of the play head

### Actual behavior

I have not observed any audio glitches in the VST 2 branch as it now stands (commit 8388e12f66)

Refresh seems good while the dialog is just idle

But refresh is still bad while you move sliders

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

dozzzzer:
Verified

PR #3882, opened 2022-10-31
Screen refresh rate improve by using condition_variable

pietro68:
Resolves: #3869 

Recommended:

### DIFF
--- a/src/effects/RealtimeEffectState.cpp
+++ b/src/effects/RealtimeEffectState.cpp
@@ -18,6 +18,7 @@
 
 #include <chrono>
 #include <thread>
+#include <condition_variable>
 
 //! Mediator of two-way inter-thread communication of changes of settings
 class RealtimeEffectState::AccessState : public NonInterferingBase {
@@ -66,9 +67,15 @@ public:
       mChannelFromMain.Read<FromMainSlot::Reader>(mEffect, mState);
    }
    void WorkerWrite() {
-      // Worker thread avoids memory allocation.
-      mChannelToMain.Write(CounterAndOutputs{
-         mState.mWorkerSettings.counter, mState.mOutputs.get() });
+
+      {
+         std::unique_lock lk(mLockForCV);
+
+         // Worker thread avoids memory allocation.
+         mChannelToMain.Write(CounterAndOutputs{
+            mState.mWorkerSettings.counter, mState.mOutputs.get() });
+      }
+      mCV.notify_one();
    }
 
    struct ToMainSlot {
@@ -154,6 +161,9 @@ public:
    SettingsAndCounter mLastSettings;
 
    MessageBuffer<ToMainSlot> mChannelToMain;
+
+   std::mutex mLockForCV;
+   std::condition_variable mCV;
 };
 
 //! Main thread's interface to inter-thread communication of changes of settings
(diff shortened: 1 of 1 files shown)


pietro68:
It is only the last two commits - I can not seem to base this PR on vst2-fix-metering!

Paul-Licameli:
Please rebase the last two commits only directly onto the release-3.2.2 branch.

pietro68:
> Please rebase the last two commits only directly onto the release-3.2.2 branch.

Will do. And of course there was a missing header <condition_variable> on gcc builds which Visual Studio did not care about.

Paul-Licameli:
Can you confirm that this really appears to correct the performance problem?

src/effects/RealtimeEffectState.cpp
```diff
@@ -69,9 +69,14 @@ class RealtimeEffectState::AccessState : public NonInterferingBase {
...
    }
    void WorkerWrite() {
+
+      std::unique_lock lk(mLockForCV);
+
       // Worker thread avoids memory allocation.
       mChannelToMain.Write(CounterAndOutputs{
          mState.mWorkerSettings.counter, mState.mOutputs.get() });
+
+      mCV.notify_one();
```
Paul-Licameli, line 79: As it says here https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one, it is not necessary to keep the mutex locked while signalling, and might even be worse for performancde.  It needs to be held for the duration of changes that make the condition the waiter waits on true.

So you might insert `{` before the declaration of `lk` and `}` before you call `noitfy_one`.

src/effects/RealtimeEffectState.cpp
```diff
@@ -179,11 +181,19 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
...
+   */
+
    const EffectSettings &Get() override {
       if (auto pState = mwState.lock()) {
          if (auto pAccessState = pState->GetAccessState()) {
             if (pAccessState->mState.mInitialized)
-               FlushAttempt(*pAccessState); // try once, ignore success
+            {
+               // try once
+               assert(pAccessState->mState.mInitialized);
```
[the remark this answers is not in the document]
pietro68: yes, I had to remove it when I rebased on release-3.2.2

src/effects/RealtimeEffectState.cpp
pietro68: However, I based my choice on what this page https://embeddedartistry.com/blog/2022/01/10/remember-to-lock-around-all-stdcondition_variable-variables/ says - search for this text: "However, it seems that the [C](https://embeddedartistry.com/fieldmanual-terms/c/)++ Core Guidelines discussions on GitHub prefer [this general guideline](https://github.com/isocpp/CppCoreGuidelines/issues/925#issuecomment-302269850):"

pietro68:
> Can you confirm that this really appears to correct the performance problem?

To me it looks like when I move the sliders while playing, the play line advances much smoother now, but still not 100% smooth. Anyway, it has improved. This on Windows, I still have to try it on OSx.

src/effects/RealtimeEffectState.cpp
pietro68: but of course I will give a try to your suggestion

crsib:
macOS has a pretty decent profiler. Why guess?

Paul-Licameli:
> macOS has a pretty decent profiler. Why guess?

Why guess, when you can deduce?

Paul-Licameli (approved):
The total change might be small enough to squash into just one commit now

Issue #3887, opened 2022-11-02
VST2: Change structures for settings and messages


pietro68:
Some plugins give problems when enumerating parameters by strings. Enumerate them by numerical ID instead.

Replace the current:
std::map<wxString, std::optional<std::pair<int,double> > > mParamsMap;

With two members:

- a `std::unordered_map<wxString, optional<double>>` for **saving and loading of settings**
- a  `vector<optional<double>>` for **messages**        

With the above, `CopySettingsContents` can simply be a no-op now

Paul-Licameli:
Closing as a duplicate because the changes will be incorporated into #3900 which also fixes issues #3899 and #3855 with user-visible consequences

Issue #3855, opened 2022-10-26
VST2: In plain UI mode, when tweaking different sliders, both the current and the previous one are moved together


dozzzzer:
### Bug description

When in plain UI mode with the effect's power button turned off, dragging several sliders left and right one by one during playback causes the previously dragged slider to shift too.

### Steps to reproduce

1. Have some audio
2. Open a VST2 effect in the sidebar in plain UI mode
3. Turn the power button off for this effect
4. Start playback
5. Drag a slider left and right OR click within its range to make it jump to a certain value
6. Drag/click within another slider
7. Observe

### Expected behavior

Only one slider moves

### Actual behavior

Both sliders are moving: the one which you drag moves where you pointing, while the previous one jumps to an unpredicted value. If you keep dragging the same slider, everything works fine. Though, when you tweak each slider just once, you'll see the previous one jumping consistently.

https://user-images.githubusercontent.com/70063970/198054684-5593d79d-a517-49a4-b160-d859b8825344.mp4

### Audacity Version

other

### Operating system

macOS

dozzzzer:
With PR #3900, the issue can no longer be reproduced.
Verified on macOS and Windows 10.
@Paul-Licameli @pietro68

Ryan-Miller-Audacity:
I was able to confirm the same, sliders move independently of each other in VST Plain GUI in build from PR #3900. Tested on Windows 10.

PR #3891, opened 2022-11-03
New definition of message and map in settings

pietro68:
Resolves: #3887 
Resolves: #3855

I tested it with DeeGain and it seems to work ok.

Update: @Paul-Licameli includes an alternative fix for #3855 that doesn't have data races

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1504,24 +1504,6 @@ bool VSTEffect::LoadSettings(const CommandParameters& parms, EffectSettings& set
    return true;
 }
 
-OptionalMessage VSTEffect::LoadUserPreset(
-   const RegistryPath & name, EffectSettings &settings) const
-{
-   if (!LoadParameters(name, settings))
-   {
-      return {};
-   }
-
-   return MakeMessageFS( VSTEffectInstance::GetSettings(settings) );
-}
-
-
-bool VSTEffect::SaveUserPreset(
-   const RegistryPath & name, const EffectSettings &settings) const
-{
-   return SaveParameters(name, settings);
-}
-
 RegistryPaths VSTEffect::GetFactoryPresets() const
 {
    RegistryPaths progs;
@@ -2106,7 +2088,7 @@ bool VSTEffectWrapper::IsCompatible(const VstPatchChunkInfo& info) const
            (info.numElements    == mAEffect->numParams);
 }
 
-OptionalMessage VSTEffect::LoadParameters(
+OptionalMessage VSTEffect::LoadUserPreset(
    const RegistryPath & group, EffectSettings &settings) const
 {
    wxString value;
@@ -2165,7 +2147,7 @@ OptionalMessage VSTEffect::LoadParameters(
 }
 
 
-bool VSTEffect::SaveParameters(
+bool VSTEffect::SaveUserPreset(
    const RegistryPath & group, const EffectSettings &settings) const
 {
    const auto& vstSettings = GetSettings(settings);
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1005,9 +1005,21 @@ namespace
 {
    struct VSTEffectMessage : EffectInstance::Message
    {
-      explicit VSTEffectMessage(VSTEffectSettings settings)
-         : settings{ std::move(settings) }
-      {}
+      typedef std::vector<std::optional<double> > ParamVector;
+
+      // Make a message from a chunk and ID-value pairs
+      explicit VSTEffectMessage(const std::vector<char>& chunk, const ParamVector& params)
+         : mChunk(chunk),
+           mParamsVec(params)
+      {
+      }
+
+      // Make a message from a single parameter
+      explicit VSTEffectMessage(int id, double value, size_t numParams)
+      {
+         mParamsVec.resize(numParams, std::nullopt);
+         mParamsVec[id] = value;
+      }
 
       ~VSTEffectMessage() override;
 
@@ -1015,7 +1027,10 @@ namespace
       void Assign(Message&& src) override;
       void Merge(Message&& src) override;
 
-      VSTEffectSettings settings;
+
+      std::vector<char> mChunk;
+      ParamVector       mParamsVec;
+
    };
 }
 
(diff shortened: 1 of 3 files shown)


Paul-Licameli:
Change the base branch of the commit please.

Only the last commit is new; it's understood this is based on vst2-statelessness while that awaits approval and merge.

src/effects/VST/VSTEffect.h
```diff
@@ -97,7 +97,7 @@ struct VSTEffectSettings
    std::vector<char> mChunk;
 
    // Fallback data used when the chunk is not available.
-   std::map<wxString, std::optional<std::pair<int,double> > > mParamsMap;
+   std::unordered_map<wxString, std::optional<double> > mParamsMap;
```
Paul-Licameli, line 100: You change this back to `unordered_map` because the parallel iteration of two maps is no longer required so the iteration order of elements can be left unpredictable.

Change the line `#include <map>` to fix macOS compilation.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1005,17 +1005,32 @@ namespace
 {
    struct VSTEffectMessage : EffectInstance::Message
    {
-      explicit VSTEffectMessage(VSTEffectSettings settings)
-         : settings{ std::move(settings) }
-      {}
+      typedef std::vector<std::optional<double> > ParamVector;
```
Paul-Licameli, line 1008: Prefer `using` over `typedef`

src/effects/VST/VSTEffect.cpp
```diff
@@ -1026,60 +1041,91 @@ auto VSTEffectMessage::Clone() const -> std::unique_ptr<Message>
...
+
+   vstSrc.mChunk.resize(0);  // capacity will be preserved though
+
+   assert(mParamsVec.size() == vstSrc.mParamsVec.size());
+
+   for (size_t i = 0; i < mParamsVec.size(); i++)
+   {
+      if (chunkWasAssigned)
+      {
+         mParamsVec[i] = vstSrc.mParamsVec[i];
```
Paul-Licameli, line 1087: Comment that this may overwrite a value with nullopt

src/effects/VST/VSTEffect.cpp
```diff
@@ -1026,60 +1041,91 @@ auto VSTEffectMessage::Clone() const -> std::unique_ptr<Message>
...
-         {
-            settings.mParamsMap[pi.mName] = std::nullopt;
-         }
+         paramVector[pi.mID] = settings.mParamsMap[pi.mName];
          return true;
       }
    );
 
-   return std::make_unique<VSTEffectMessage>(std::move(settings));
+   return std::make_unique<VSTEffectMessage>(settings.mChunk, paramVector);
```
Paul-Licameli, line 1122: `move()` both arguments to avoid vectory copies

src/effects/VST/VSTEffect.cpp
```diff
@@ -1005,17 +1005,32 @@ namespace
 {
    struct VSTEffectMessage : EffectInstance::Message
    {
-      explicit VSTEffectMessage(VSTEffectSettings settings)
-         : settings{ std::move(settings) }
-      {}
+      typedef std::vector<std::optional<double> > ParamVector;
+
+      // Make a message from a chunk and ID-value pairs
+      explicit VSTEffectMessage(const std::vector<char>& chunk, const ParamVector& params)
```
Paul-Licameli, line 1011: Change `&` to `&&` twice

src/effects/VST/VSTEffect.cpp
```diff
@@ -1005,17 +1005,32 @@ namespace
...
+         : mChunk(chunk),
+           mParamsVec(params)
```
Paul-Licameli, line 1013: Then add `move(...)` twice, and this completes the elimination of vector copies

src/effects/VST/VSTEffect.cpp
```diff
@@ -1026,60 +1041,91 @@ auto VSTEffectMessage::Clone() const -> std::unique_ptr<Message>
...
 
-
-std::unique_ptr<EffectInstance::Message> VSTEffectInstance::MakeMessage(int id, double value) const
-{
-   VSTEffectSettings settings;
+   VSTEffectMessage::ParamVector paramVector;
+   paramVector.resize(settings.mParamsMap.size(), std::nullopt);
 
    ForEachParameter
    (
```
Paul-Licameli, line 1114: `FetchSettings` calls `ForEachParameter`.  Then you call `ForEachParameter` again.  And that means `GetString(effGetParamName, i)` is called `2 * mAEffect->numParams` times.

But the names are not needed, and the frequent calls of `GetString` might be a reason the main thread consumes too much CPU during slider movements.

So I suggest:  simply iterate an integer up to `mAEffect->numParams` and do `callGetParameter` directly.  It might help performance.

src/effects/VST/VSTEffect.cpp
```diff
@@ -3815,10 +3863,22 @@ VSTEffectValidator::~VSTEffectValidator()
...
-std::unique_ptr<EffectInstance::Message> VSTEffectValidator::MakeMessage(
-   VSTEffectSettings &settings)
+std::unique_ptr<EffectInstance::Message>
+VSTEffectWrapper::MakeMessageFS(VSTEffectSettings &settings) const
 {
-   return std::make_unique<VSTEffectMessage>(settings);
+   VSTEffectMessage::ParamVector paramVector;
+   paramVector.resize(settings.mParamsMap.size(), std::nullopt);
+   
+   ForEachParameter
```
Paul-Licameli, line 3872: This `ForEachParameter` can remain as it is.  It's not needed to optimize this to make the play head movement smooth while dragging the sliders.

src/effects/VST/VSTEffect.cpp
Paul-Licameli: You need a chunk too to make the message, so extract that part of `FetchSettings` as a function you can re-use here, with `doFetch == false` so that it only really reserves sufficient chunk space.

The purpose of the no-argument `MakeSettings()` is just initialization of Message objects with vectors of sufficient size.

Paul-Licameli:
That's the end of this round of review.

pietro68:
> Change the base branch of the commit please.
> 
> Only the last commit is new; it's understood this is based on vst2-statelessness while that awaits approval and merge.

I am not sure that I know how/can do this. Github lets me switch only to branches that are in the origin and  vst2-statelessness is not.

src/effects/VST/VSTEffect.h
pietro68: ok

src/effects/VST/VSTEffect.cpp
pietro68: ok

src/effects/VST/VSTEffect.cpp
pietro68: ok

src/effects/VST/VSTEffect.cpp
pietro68: OK, and I left the const modifier, not sure if it makes a difference but it does not hurt

src/effects/VST/VSTEffect.cpp
pietro68: right, ok

src/effects/VST/VSTEffect.cpp
pietro68: However, this code does not get called in response to a slider being moved, but only about 2 times when the effect is instantiated in the track, and then when its gui is opened. 

Anyway, I agree it is not so clean like it is now. I will do this too.

src/effects/VST/VSTEffect.cpp
pietro68: ok

src/effects/VST/VSTEffect.cpp
pietro68: However (again): the original code (previous to these changes) did not get the chunk **and** did not get the single values - so I should do the same here (which is consistent: either I get both kinds of info, or I get none), and for this I just need to remove the call to ForEachParam, which also solves the original objection.

src/effects/VST/VSTEffect.cpp
Paul-Licameli: Remove the const modifiers

src/effects/VST/VSTEffect.cpp
Paul-Licameli: But a purpose of the call is to reserve sufficient space, to avoid memory allocations and dellocations later.

So if the loop is entirely eliminated, you should still make a vector of sufficient `nullopt`s.  That just requires the number of parameters.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1104,22 +1104,15 @@ void VSTEffectMessage::Merge(Message && src)
...
    paramVector.resize(settings.mParamsMap.size(), std::nullopt);
 
-   ForEachParameter
-   (
-      [&](const ParameterInfo& pi)
-      {
-         paramVector[pi.mID] = settings.mParamsMap[pi.mName];
-         return true;
-      }
-   );
```
Paul-Licameli: As I explained, this is still incorrect:  don't just delete, but replace with a sizing of paramVector to `mAEffect->numParams` and filled with nullopts.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1005,12 +1005,12 @@ namespace
...
    struct VSTEffectMessage : EffectInstance::Message
    {
-      typedef std::vector<std::optional<double> > ParamVector;
+      using ParamVector = std::vector<std::optional<double> >;
 
       // Make a message from a chunk and ID-value pairs
-      explicit VSTEffectMessage(const std::vector<char>& chunk, const ParamVector& params)
-         : mChunk(chunk),
-           mParamsVec(params)
+      explicit VSTEffectMessage(const std::vector<char>&& chunk, const ParamVector&& params)
```
Paul-Licameli, line 1011: Remove the `const` qualifiers from the rvalue reference parameters.

Rvalue reference to const is possible in C++ but it's weird.

src/effects/VST/VSTEffect.cpp
```diff
@@ -1104,22 +1104,15 @@ void VSTEffectMessage::Merge(Message && src)
...
 std::unique_ptr<EffectInstance::Message> VSTEffectInstance::MakeMessage() const
 {
+   // The purpose here is just to allocate vectors (chunk and paramVector)
+   // with sufficient size, not to get the values too
    VSTEffectSettings settings;
-   FetchSettings(settings, false);
+   FetchSettings(settings, /* doFetch = */ false);
 
    VSTEffectMessage::ParamVector paramVector;
    paramVector.resize(settings.mParamsMap.size(), std::nullopt);
```
Paul-Licameli, line 1113: Better simply `mAEffect->numParams` for the vector size

src/effects/VST/VSTEffect.cpp
pietro68: ok, done

src/effects/VST/VSTEffect.cpp
pietro68: ok

Paul-Licameli:
See PR #3900 instead which now includes these changes.

Issue #3899, opened 2022-11-07
VST2: Previous settings are restored after loading a preset and pressing play


pietro68:

pietro68:
Apologies: I originally wrote that this happens with DeeGain too, but it does happen only on branch i-3887, where the change in the definition of the message was implemented. It does not happen with DeeGain on release-3.2.2 (after vst2-statelessness was merged into it)

Paul-Licameli:
I prepared first a user preset called LOUD for DeeGain (VST 2 version) with the slider at maximum.

I run my fix at #3900.  I do the steps you describe.  It works as expected.  Same too when I change to "Factory defaults" (which are not a true factory preset, there are none of those).

I do the same at the previous commit, 7616b89f9ea38f31a64631b36de3b727dd7d4248.  But it's also as expected.

I checkout your i-3887.  I rebase its last five commits onto 7616b89f9ea38f31a64631b36de3b727dd7d4248.  I DO reproduce the problem.

I cherry-pick my #3900 onto the branch, and resolve a conflict.  I repeat.  The problem is fixed.

Therefore I conclude that my fix for #3855 introduced this other real problem that you identified, so now I have fixed my own fix.

pietro68:
I confirm that what you committed in branch Issue-3899 fixed the regression that emerged, specifically for DeeGain, in my branch i-3877. So I will approve that PR.

Now there are two things to take care of:

1) after I rebased i-3877 on Issue-3899, both branches have the same content. So we'd have to merge in release only one of the two.

2) this issue reports a problem that happens also for BlueCat Chorus - but it happens also before either of the two branches above were started, and it still happens now. I would say, I could edit the description of this issue to refer to DeeGain only and not to BC Chorus anymore; and what is affecting BC Chorus, would still be descripted by previous issue #3842

Ryan-Miller-Audacity:
Confirmed via PR #3900 that this issue is no longer occurring within that branch. 
Tested on Windows 10.

Paul-Licameli:
I updated the title to be very specific

dozzzzer:
Verified.

PR #3900, opened 2022-11-08
Issue 3899

Paul-Licameli:
Resolves: #3899
Resolves: #3855

Fix play after switching to a user preset in VST effects

Recommended:

### DIFF
--- a/src/effects/RealtimeEffectState.cpp
+++ b/src/effects/RealtimeEffectState.cpp
@@ -211,8 +211,10 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
                // Other thread isn't processing.
                // Let the instance consume the message directly.
                if (auto pInstance = pState->mwInstance.lock()) {
+                  auto &stateSettings = pState->mMainSettings.settings;
+                  stateSettings = std::move(settings);
                   EffectInstance::MessagePackage package{
-                     pState->mWorkerSettings.settings, pMessage.get()
+                     stateSettings, pMessage.get()
                   };
                   pInstance->RealtimeProcessStart(package);
                   return;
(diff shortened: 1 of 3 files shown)


pietro68:
I tried this with BlueCat Chorus and the two originally reported problems still persist. Maybe what happened is that you tested it using DeeGain, but I was wrong when I wrote that DeeGain is affected by this (I just corrected the original issue).

Paul-Licameli:
> I tried this with BlueCat Chorus and the two originally reported problems still persist. Maybe what happened is that you tested it using DeeGain, but I was wrong when I wrote that DeeGain is affected by this (I just corrected the original issue).
> 
> I tried some debug printing in VSTEffectInstance::ProcessInitialize, printing the settings that come in (that is what gets called when pressing play). With both this branch and the release branch (on which this branch is based) what gets printed are the settings for what was in effect before the loading of the preset; and this both in the chunk and in the params map.
> 
> In other words, it looks as if the loading of a user preset is not remembered by the framework.

You did not specify which branch you tested when you reported the problem.

I think the problem with DeeGain is real at the previous commit 7616b89f9ea38f31a64631b36de3b727dd7d4248 and is fixed by this commit.

pietro68:
> You did not specify which branch you tested when you reported the problem.

I was testing branch i-3877

> I think the problem with DeeGain is real at the previous commit [7616b89](https://github.com/audacity/audacity/commit/7616b89f9ea38f31a64631b36de3b727dd7d4248) and is fixed by this commit.

I confirm that the commit `Fix start of play of VST after loading preset...` fixed the problem reported in #3899, but for DeeGain only. The problem persists with BlueCat Chorus.

pietro68 (commented):
Changes understood and fine with me. Please disregard my above comment about BlueCat Chorus, let's keep it out of the scope of this PR.

pietro68 (approved):

PR #3901, opened 2022-11-08
Fixes Xcode >= 14 build

crsib:
Xcode >= 14 no longer allows to disable code signing. This limitation is overcame by enabling "Sign to Run Locally by Xcode"

> This PR does not immediately change the behavior of CI builds, as CI uses Xcode 13

Recommended:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1429,6 +1429,22 @@ elseif( CMAKE_SYSTEM_NAME MATCHES "Darwin" )
          ${CMAKE_COMMAND} -E copy $<TARGET_FILE:Wrapper> "${_EXEDIR}/Wrapper"
       POST_BUILD
    )
+
+   # Because the wrapper is copied into the bundle, we need to
+   # re-sing the bundle as the last post build step.
+   # This is only required when using the Xcode generator with
+   # the  Xcode >= 14.
+   if (XCODE_VERSION VERSION_GREATER_EQUAL 14)
+      add_custom_command(
+         TARGET
+            ${TARGET}
+         COMMAND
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -1429,6 +1429,22 @@ elseif( CMAKE_SYSTEM_NAME MATCHES "Darwin" )
...
+   # Because the wrapper is copied into the bundle, we need to
+   # we need to resing the bundle as the last post build step.
+   # This is only required when using the Xcode generator with
+   # the  Xcode >= 14.
+   if (XCODE_VERSION VERSION_GREATER_EQUAL 14)
+      add_custom_command(
+         TARGET
+            ${TARGET}
+         COMMAND
+            ${CMAKE_COMMAND} -E echo "Fixing Xcode 14 code singning issues"
```
Paul-Licameli, line 1442: "signing"

src/CMakeLists.txt
```diff
@@ -1429,6 +1429,22 @@ elseif( CMAKE_SYSTEM_NAME MATCHES "Darwin" )
          ${CMAKE_COMMAND} -E copy $<TARGET_FILE:Wrapper> "${_EXEDIR}/Wrapper"
       POST_BUILD
    )
+
+   # Because the wrapper is copied into the bundle, we need to
+   # we need to resing the bundle as the last post build step.
```
Paul-Licameli, line 1434: "we need to" 2x

vsverchinsky, line 1434: Extra "we need to"

vsverchinsky (approved):

src/CMakeLists.txt
Paul-Licameli: "re-sign"
(not "resign" which means something else)

Issue #3470, opened 2022-08-29
Accessibility of the power toggle buttons in the realtime effects panel


DavidBailes:
### Bug description

The state of the power toggle button is not accessible at all using Windows' built in screen reader, Narrator, and the accessibility using Jaws or NVDA is not good enough.

So for for users of Narrator, the realtime effects panel is unusable, and for users of Jaws and NVDA the usability it isn't good.

### Steps to reproduce

1. Open the realtime effects panel.
2. Tab to a power toggle button.
3. Press the toggle button.
4. Tab away from it and then back to it to read it again in the changed state.

### Expected behavior

When a toggle button becomes the focus, the state of the button should always be read. Screen readers on Windows differ in the exact words used, but NVDA reads either pressed or not pressed.

When a toggle button is pressed, the screen reader reads the new state, for example pressed or not pressed.

### Actual behavior

When the power toggle button becomes the focus:
a. Using NVDA or Jaws, the state is only read if the button is "on".
b. Using Narrator, the state is not read.

When the power toggle button is pressed, none of the screen readers reads the new state.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

A suggested fix, would be to use a custom accessible object for the toggle button, such that:
1. Its role is static text, so that the screen reader does not read any role.
2. The accessible name contains the name, role, and state: for example "power button pressed"
3. When the button is pressed, a name change event is created, and so the screen reader would say something like "power button not pressed".

Note: The actual control should not be derived from a wxButton, and this leads to problems with both Jaws and NVDA, for different reasons (available on request).

LWinterberg:
tested your fix, seems to work now with Narrator

PR #3908, opened 2022-11-10
issue 3370: accessibility of toggle buttons in RT effects panel

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/3470

Problem:
Narrator doesn't read the pressed state at all, and Jaws and NVDA don't read the toggle buttons as expectd.

Fix:
In MSAA, there isn't such a thing as a toggle button. So imitate one by including the role and the state in the name, and creating a name change event when the state changes.

Recommended:


DavidBailes:
didn't mean to merge into master!

Issue #3842, opened 2022-10-24
VST2: Effect's parameters get reset to the default values when playback starts or when re-opening the effect's window


dozzzzer:
### Steps to reproduce

1. Open up a VST2 effect from the sidebar (e.g. BlueCat Chorus or [TAL Reverb 2](https://plugins4free.com/plugin/683/) )
2. Move one of the sliders to some value (e.g. Gain or Wet to minimum)
3. Load a preset where the slider you moved in the step above is set to a different value 
4. Start playback, or close/reopen the GUI

### Expected behavior

The effect's parameters remain unchanged

### Actual behavior

The parameters get reset to what was set at step 2. 

### Audacity Version

Win x64, macOS universal builds [from here](https://github.com/audacity/audacity/actions/runs/3312593613)

### Operating system

macOS, Windows

dozzzzer:
@pietro68 FYI, the exact same behavior shows the destructive application of the effects: open an effect, tweak some controls, start preview, stop preview, start again - the controls get reset to the defaults.

dozzzzer:
Also, please ignore my comment:
> When using effect's presets, the parameters stick when playing the track/re-opening the UI. The described defect can only be observed in the case of tweaking the parameters by hand.

After further investigation, it turned out that some effects (e.g. Fresh Air by Slate Digital) have their controls reset too when using presets.

pietro68:
These are actually two separate bugs (considering repro step 4.) and they actually also happen using DeeGain. Debug print statements showed that:

a) when pressing play, VSTEffectInstance::ProcessInitialize will be called ant the passed settings will contain an outdated chunk and map of parameters

b) when hiding the gui and reopening, at some point VSTEffectInstance::RealtimeProcessStart will be called and the passed message will contain an outdated chunk and ParamsVector

Ryan-Miller-Audacity:
Commit: https://github.com/audacity/audacity/commit/852a4b4de6cce99d549a33e974a83d38c923206a
GH job: https://github.com/audacity/audacity/actions/runs/3429017268
OS: Windows 10

Plugin: ValhallaDelay
This issue exists: No

Plugin: TAL Reverb-2
This issue exists: YES

Plugin: MCompressor by Melda Productions
This issue exists: YES

Plugin: MBandPass by Melda Productions
This issue exists: YES

Plugin: Chandler GAV19T by Plugin Alliance
This issue exists: YES

Plugin: Valhalla Supermssive
This issue exists: NO

[...]

PR #3914, opened 2022-11-10
Fix to issue 3842

pietro68:
Resolves: #3842 

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1316,8 +1316,31 @@ bool VSTEffectInstance::RealtimeProcessStart(MessagePackage& package)
       const auto len = chunk.size();
       const auto data = chunk.data();
       callSetChunk(true, len, data, &info);
+
+      // Emulate what happens for plugins like DeeGain,
+      // i.e. after setting the chunk, the vst callback is called
+      // with message audioMasterAutomate for all parameters.
+      //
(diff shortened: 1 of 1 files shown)


src/effects/VST/VSTEffect.cpp
```diff
@@ -1316,9 +1316,32 @@ bool VSTEffectInstance::RealtimeProcessStart(MessagePackage& package)
...
+      {
          slave->callSetChunk(true, len, data, &info);
 
+         // Emulation of DeeGain-type plugins for the slaves too
+         for (size_t i = 0; i < message.mParamsVec.size(); i++)
+         {
+            if (message.mParamsVec[i])
+               slave->Automate(i, *message.mParamsVec[i]);
+         }
+      }
```
Paul-Licameli, line 1343: Great idea!  I’m guessing however that it is enough to do this to self only, and not needed for slaves.  Can you confirm?  If I am right, then make the minimal needed change.

Paul-Licameli: Also be sure to test the changing of presets while playing and while not.

I haven't tried but I think this may fail to fix it during play, because VSTEffectInstance::Automate required a check that it runs on the main thread, and otherwise does nothing.  That was necessary for me to add, to avoid deadlocks.

So:  the final fix may require you to wrap the call in a lambda as argument to `BasicUI::CallAfter`.

PR #3918, opened 2022-11-11
Build system overhaul

crsib:
**This PR introduces breaking changes to the build system**
**The old build directory must be removed!**

This commit addresses the following problems of the build system:

* Shared libraries management is very complex - libraries are copied as a post build step to the desired location.
  This easily breaks, if library modification does not result in the executable file relink. *macOS* still requires a custom
  build step to set the  portable library names inside a bundle. This step will be dropped after Conan 2.0 is released.
* conan-cmake likely will not gain the support for Conan 2.0
* Dependency managements is quite complex and knows too much about Conan
* At the same time - the approach taken is not configurable enough to properly support Qt
* Instead of using `find_package` for all the dependencies, we sometimes use `pkg-config`. 
  This further complicates the CMake code.

In this PR:

* `conan-cmake` is removed
* `AudacityDependencies` now generates the list of dependencies based on `DependenciesList.cmake`
* Added a much simplified `audacity_find_package` function
* Conan recipe is now written in Python. This allows much more customization to the `imports` section
* Much of the Conan invocation logic  is moved to a python script. This is much more readable and maintainable than the
  CMake script
* Missing Find* modules are added

Recommended:


scripts/build/macOS/fix_bundle.py
```diff
@@ -0,0 +1,121 @@
...
+
+def macos_collect_dependencies(file):
+    result = []
+    dylib_id = get_dylib_id(file)
+
+    with subprocess.Popen(['otool', '-L', file], stdout=subprocess.PIPE) as p:
+        lines = [line.decode('utf-8').strip() for line in p.stdout.readlines()]
+        for line in lines:
+            match = re.match(r'(.*)\s+\(', line)
+            if match:
```
ksoze95, line 43: I think you should reduce cyclomatic complexity by returning early rather than constant nesting.

scripts/build/macOS/fix_bundle.py
```diff
@@ -0,0 +1,121 @@
...
+        install_name_tool.append(rpath)
+
+    install_name_tool.append(file)
+
+    print(' '.join(install_name_tool))
+    try:
+        subprocess.check_call(install_name_tool)
+    except subprocess.CalledProcessError as err:
+        print(f'=========\ninstall_name_tool failed with code {err.returncode}\n\tstdout: {err.stdout}\n\tstderr: {err.stderr}\n=========')
+        if err.returncode == -9:
```
ksoze95, line 78: I think you should reverse the logic of this condition and raise the exception instead, then remove else and left indent the rest of the code.

linux/packages/ubuntu-20.04/debian/rules
```diff
@@ -14,14 +14,16 @@ override_dh_auto_configure:
 	dh_auto_configure -- \
 	-DCMAKE_BUILD_TYPE=Release \
 	-Daudacity_conan_allow_prebuilt_binaries=no \
-   -D audacity_conan_force_build_dependencies=yes \
+    -Daudacity_conan_force_build_dependencies=yes \
```
ksoze95, line 17: Could we replace these preceding spaces with tabs so that they are consistent with the other -D defines?  Or replace all the other with same amount of spaces.

crsib:
@ksoze95 I've pushed the fixes in a separate commit. Can you please review?

ksoze95 (approved):

vsverchinsky (approved):

cmake-proxies/cmake-modules/DependenciesList.cmake
```diff
@@ -0,0 +1,83 @@
...
+
+# Conan expects Python boolean, CMake will convert values to
+# ON/OFF
+
+if( ${_OPT}has_asio_support )
+   list(APPEND conan_package_options "use_asio=True")
+endif()
+
+if(UNIX AND NOT APPLE)
+   find_package(JACK QUIET)
```
imciner2, line 49: Is there a reason JACK is handled differently here than the other packages (e.g. using standard `find_package` instead of the new `audacity_find_package`)?

linux/packages/fedora34/audacity.spec
```diff
@@ -84,10 +84,6 @@ supports PulseAudio, OSS and ALSA under Linux.
 %cmake \
     -D CMAKE_BUILD_TYPE=Release \
     -D audacity_conan_enabled=Off \
-    -D audacity_conan_allow_prebuilt_binaries=no \
-    -D audacity_conan_force_build_dependencies=yes \
-    -D audacity_lib_preference=system \
```
imciner2, line 89: Why did you remove the Fedora preference for system libraries in this CI spec? Keeping this preference is very useful for matching the downstream builds we do on the Fedora infrastructure and repos because we prefer system libraries when possible (and only do local if truly needed).

src/CMakeLists.txt
```diff
@@ -1639,24 +1606,21 @@ else()
...
    else()
 
-      install( TARGETS ${TARGET} RUNTIME )
+      install( TARGETS ${TARGET}
+               RUNTIME
+               RESOURCE DESTINATION "${_PKGDATA}" )
 
       install( DIRECTORY "${_DEST}/${_LIBDIR}/"
                DESTINATION "${_LIBDIR}"
                USE_SOURCE_PERMISSIONS
```
imciner2, line 1615: Is there a reason you need to keep the source permissions? In the Fedora build we actually want the libraries to be executable, so permissions of 755 would be appropriate. I haven't looked into why it isn't getting the 755 permissions at build time, but this has been an issue in the past because without the executable bit, the debug info extraction doesn't run and there are no symbols for any of the libraries.

imciner2 (commented):
Does this changeset also cleanup the rpath handling on Linux builds? In the past we have seen the rpath pose (e.g. https://github.com/audacity/audacity/issues/3289, https://github.com/audacity/audacity/pull/3671). I haven't had a chance to run this through the Fedora build system yet (and I have actually been a bit busy so I haven't had a chance to try 3.2 with it yet), but in the past we have had to strip the rpath from the private libraries and modules (https://src.fedoraproject.org/rpms/audacity/blob/rawhide/f/audacity.spec#_206).

cmake-proxies/cmake-modules/DependenciesList.cmake
crsib: Yes, we never build Jack. This keeps the old behavior actually.

linux/packages/fedora34/audacity.spec
crsib: `audacity_conan_enabled=Off` implies that system libraries are used for all the Conan managed dependencies. However, you are correct, that this changes the behavior for LV2, so I should put this line back

src/CMakeLists.txt
crsib: I'm pretty sure that without it will copy the files with 0666 permissions. This should not be an issue for Fedora though - without Conan this step is effectively an no-op, as all the libraries are installed in the previous step.

vsverchinsky (approved):

PR #3921, opened 2022-11-11
Better fix for issue #3842

pietro68:
Resolves: #3842

Maybe this fix is good enough to be merged; if not, take it as a proof of concept showing what was the core problem and how it can be fixed.

In a nutshell: when loading a preset, what happened was that the call to `Flush()` in `EffectUIHost::OnIdle` would make `mMainSettings` (which just took the values from the loaded preset) be overwritten by `mLastSettings` (which was set by the last knob move which happened before the preset loading). I proved this by commenting out the call to Flush() in ::OnIdle - bug disappeared. 

But we probably do not want to remove the call to `Flush` in `OnIdle`, so I resorted to this: force mLastSettings to be set to the just assigned (from preset) mMainSettings, before any call to ::OnIdle comes after the preset load. That's what you will see in this fix.

You might ask yourself: why did DeeGain not suffer from this? (and bluecat in my previous fix where I emulated DeeGain's behavior?) it probably happened that those real/emulated calls to `::Automate` that would happen in response to setting the chunk, would make the preset's settings be copied to `mLastSettings` **before** the first call to `::OnIdle` would come - so `mMainSettings` would then get the correct values.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.h
+++ b/libraries/lib-components/EffectInterface.h
@@ -209,6 +209,8 @@ public:
       auto result = std::forward<Function>(function)(settings);
       this->Set(std::move(settings), std::move(result));
    }
+
+   virtual void SetLastSettingsEqualToMainSettings() {};
 };
 
 //! Implementation of EffectSettings for cases where there is only one thread.
--- a/src/effects/EffectUI.cpp
+++ b/src/effects/EffectUI.cpp
@@ -826,6 +826,9 @@ void EffectUIHost::OnUserPreset(wxCommandEvent & evt)
       return mEffectUIHost.GetDefinition().LoadUserPreset(
          UserPresetsGroup(mUserPresets[preset]), settings).value_or(nullptr);
    });
+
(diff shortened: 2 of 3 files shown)


Paul-Licameli (changes requested):
It's good that you discovered the fix.

But I don't accept adding the new virtual function.

The function just does what `Flush()` is already meant to do.  So, simply add two calls to `Flush()` and confirm that the bug is still fixed.

pietro68:
> It's good that you discovered the fix.
> 
> But I don't accept adding the new virtual function.
> 
> The function just does what `Flush()` is already meant to do. So, simply add two calls to `Flush()` and confirm that the bug is still fixed.

I do not think so? the function I added does the opposite of `Flush()` (at least this is how I understand it by reading it: Flush() assigns `mLastSettings` to `mMainSettings`?).  I tried replacing the call to `SetLastSettingsEqualToMainSettings `with Flush() and the bug would not be fixed anymore.

Paul-Licameli:
Introduces a hang when changing presets while the dialog is open

But see #3929 now

PR #3928, opened 2022-11-13
Update cs.po

pafri:
Czech translation update. Fixed e-mail in header.

Recommended:

### DIFF
--- a/locale/cs.po
+++ b/locale/cs.po
@@ -1,12 +1,12 @@
 # Header entry was created by Lokalize.
 #
-# Pavel Fric <pavelfric@seznam.cz>, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021.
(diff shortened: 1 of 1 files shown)


pafri:
Hi,

could someone make the review, for the file to be able to go further.

Thanks.

LWinterberg:
@pafri Looks like your changes break Audacity, it doesn't compile. It seems like it breaks sometime during or shortly after looking at the nyquist effects.

pafri:
Also, what does it mean. Should something (specific strings) remain in English, but which. Or some quotation in source file is missing?

pafri:
I would find that command to check the .po file first. It could show the exact problematic row.

pafri:
Hi, after the fix in some strings, I´ve made the new request: see "Update cs.po - second attemt #4090". So this failed attempt shoul be closed and the #4090 could be approved:

"All checks have passed
11 successful checks"

Issue #3942, opened 2022-11-15
Some Valhalla VST2 Plugins crash Audacity when starting playback with interface open


Ryan-Miller-Audacity:
### Bug description

This is a new issue that occurs in the most recent release build. With either ValhallaSuperMassive or ValhallaDelay (VST2) in the stack, if you open the interface and then press play, Audacity crashes. 

This does not occur with other Valhalla plugins, only the two mentioned above, and I haven't found any other examples of this issue in my plugin library so far.

This issue occurs in this build: https://github.com/audacity/audacity/actions/runs/3465950077
This issue is a regression on this build (where the issue did not occur): https://github.com/audacity/audacity/actions/runs/3461897323

This appears to only happen for ValhallaDelay on MacOS, but both ValhallaDelay and ValhallaSupermassive on Windows.

### Steps to reproduce

0. Delete pluginsettings.cfg
1. Get Valhalla Supermassive/Delay plugins from https://valhalladsp.com/plugins/
2. Open Audacity and grab some audio
3. Add Valhalla Supermassive or Valhalla Delay to the RT stack
4. Click the plugin to open its interface
5. Press the play button
6. Audacity crashes

### Expected behavior

Playback works normally; no crash

### Actual behavior

Audacity crashes

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

Ryan-Miller-Audacity:
FYI @pietro68 @Paul-Licameli

Paul-Licameli:
I tried this with Supermassive (the free version) and failed to reproduce it.

Paul-Licameli:
Aha, I can reproduce with Delay.

Paul-Licameli:
I added a missing step 0 to reproduce the problem reliably.

Paul-Licameli:
Commit c75bfef6b8b71c09334dc2a260e8e8056bd1472a is the bad one

Paul-Licameli:
Actually I have seem the problem one commit previous, at ac953c12ba61eb43e4d63791c1f6c79aaa0e55f9

There seems to be some intermittency about it

Paul-Licameli:
Try the fix! https://github.com/audacity/audacity/actions/runs/3473602583 @dozzzzer @Ryan-Miller-Audacity

dozzzzer:
The fix works for me both on macOS and Windows, no crash observed.

Ryan-Miller-Audacity:
[...]

PR #3945, opened 2022-11-15
Fix dangling reference

Paul-Licameli:
Resolves: #3942

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1177,7 +1177,13 @@ bool VSTEffectInstance::IsReady()
 bool VSTEffectInstance::ProcessInitialize(
    EffectSettings& settings, double sampleRate, ChannelNames)
 {
-   StoreSettings(GetSettings(settings));
+   // Issue 3942: Copy the contents of settings first.
+   // settings may refer to what is in the RealtimeEffectState, but that might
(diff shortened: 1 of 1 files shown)


pietro68 (approved):
Code changes and reason for them understood.

Paul-Licameli:
Added another commit with similar precautions for VST3 and requested another review.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -447,9 +448,19 @@ void VST3Wrapper::FetchSettings(EffectSettings& settings)
       auto cleanup = finally([&] { componentHandler->EndStateChange(); });
 
       //Restore state
-      const auto* vst3settings = &GetSettings(settings);
-      if(!vst3settings->processorState.has_value())
-         vst3settings = &GetSettings(mDefaultSettings);
+      // See Issue 3942
+      // Callbacks to performEdit might happen inside setState that call
+      // ModifySettings to reassign the given settings, but the settings may be
```
vsverchinsky, line 453: `BeginStateChange` prevents `ModifySettings` from being called in the same callstack with `setState`

Paul-Licameli: You are right.

Do you recall why you wrote BeginStateChange?

Were you observing exactly the sort of dangling reference problem I destribed?

Issue #3752, opened 2022-10-03
Waves StudioRack responsiveness (Simplify AudioUnit implementation)


Paul-Licameli:
### Bug description

Waves Studio Rack takes unusually long to open its dialog or update after a change of preset

### Expected behavior

Verify after merge of PR that AudioUnit effects in stacks aren't broken and stickiness works

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

### Additional context

Maybe this isn't strictly needed for 3.2.x.

But this is also meant to serve as an analogy for implementation of VST2 statelessness.

PR #3953, opened 2022-11-17
Audio unit messages

Paul-Licameli:
Resolves: #3752

Recommended:


src/effects/audiounits/AudioUnitValidator.cpp
```diff
@@ -205,10 +208,10 @@ void AudioUnitValidator::EventListener(const AudioUnitEvent *inEvent,
          // Reassign settings, so that there is "stickiness" when dialog is
```
pietro68: I think that you could remove the call to ModifySettings and leave the lambda body there, even at this stage? this because the passed EffectSettings arg is not used anymore in the lambda.

pietro68 (approved):
Code changes understood and fine by me.

Issue #3839, opened 2022-10-24
VST3 sidebar effect persistency is broken


Paul-Licameli:
### Bug description

Update:  This may have been an erroneous issue report.  Steps to reproduce are in doubt.  The defect may have existed only in some intermediate builds of the VST2 statelessness branch.

### Steps to reproduce

1. New project; generate noise; put a VST3 effect in the stack, like https://www.dotec-audio.com/deegain.html
2. Open dialog, change settings to non-default (like dragging the gain to minimum)
3. Save the project, close it, reopen it
4. Reopen the effect dialog

### Expected behavior

See the settings you saved

### Actual behavior

See default settings again

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows, macOS, Linux

Ryan-Miller-Audacity:
I retested this as part of the "Retest after VST merged" board and am seeing effect persistency for VST3 effects working as expected. If @dozzzzer / @Paul-Licameli agree we can likely close this and remove from the board(s).

dozzzzer:
In the [latest release build](https://github.com/audacity/audacity/actions/runs/3471420722) I'm seeing that effect persistency is broken when using **Deegain VST3** module following the steps below:

1. Launch Audacity
2. In the sidebar select Deegain (VST3), open it
2. Pull the gain slider all the way down
3. Cmd (Ctrl) + S (not closing the Deegain's dialog)
4. Cmd (Ctrl) + W to close the project
5. Cmd (Ctrl) + O to reopen the just saved project
6. Open the Deegain's dialog
7. Check the gain value

Act. res.: "0.0"
Exp. res.: "-20.0"

Note: the issue was observed both on macOS and Windows. No other modules (VST, AU) are showing this effect.

Paul-Licameli:
#3840 is now rebased onto latest release, including all other bug fixes.  It may be the medicine we need now.

dozzzzer:
Verified

vsverchinsky:
Another STR:
1. Open Audacity (with empty project), create a track, add deegain to the stack
2. Open effect UI and move gain slider to the minimum
3. **Close effect UI**
4. Save the project
5. Open effect UI once again and move gain slider to the maximum
6. Now **close the project**
7. Open same project and then open effect UI

Observe: gain slider is in the minumum position 

@dozzzzer

dozzzzer:
Indeed, following the steps above the problem occurs again

[...]

PR #3954, opened 2022-11-17
VST3 persitency fix

vsverchinsky:
Resolves: #3839

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Wrapper.cpp
+++ b/src/effects/VST3/VST3Wrapper.cpp
@@ -266,8 +266,7 @@ public:
    void EndStateChange()
    {
       assert(mStateChangeSettings != nullptr);
-      if(!mParametersCache.empty())
-         FlushCache(*mStateChangeSettings);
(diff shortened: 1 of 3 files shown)


src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -653,13 +647,21 @@ void VST3Wrapper::ConsumeChanges(const EffectSettings& settings)
...
       auto componentHandler = static_cast<ComponentHandler*>(mComponentHandler.get());
       componentHandler->FlushCache(settings);
 
+      const auto doProcessing = !GetSettings(settings).parameterChanges.empty();
+
+      if(hasChanges != nullptr)
+         *hasChanges = doProcessing;
+
+      if(!doProcessing)
+         return;
```
Paul-Licameli, line 663: I see a new early exit, which also eliminates some unnecessary work, besides fixing the issue described.  That's good

Paul-Licameli (approved):

Issue #3843, opened 2022-10-24
VST2: Dry signal during first-time playback with the effect added in the sidebar and its GUI opened


dozzzzer:
### Steps to reproduce

1. Cmd (Ctrl) + N to create a new project
2. Have some audio
3. In the sidebar, select [TAL Reverb 2](https://plugins4free.com/plugin/683/)
4. Open its dialog
5. Load the Dull Plate preset
6. Play the track
7. Stop
8. Now play the track again

### Expected behavior

The effect is applied correspondingly.

### Actual behavior

The output signal is dry when hitting Play for the first time after showing the UI of the effect.
When playing the track again, the output signal turns to wet.

https://user-images.githubusercontent.com/70063970/197574786-f93c411c-e98e-40af-b8c6-71d101bbfa37.mp4

### Audacity Version

The build [from here](https://github.com/audacity/audacity/actions/runs/3312593613)

### Operating system

macOS, Windows 11

pietro68:
So far it looks like this is specific to TAL-reverb, I can not reproduce it using BlueCat Chorus

dozzzzer:
Commit: 852a4b4de6cce99d549a33e974a83d38c923206a
GH job: https://github.com/audacity/audacity/actions/runs/3429017268
OS: macOS 13.0

Plugin: Renegate by Auburn Sounds
The issue exists: No

Plugin: Deegain
The issue exists: No

Plugin: BassDrive by Tonelib
The issue exists: No

Plugin: BC Chorus 4
The issue exists: No

Plugin: Valhalla Supermassive
The issue exists: No

Plugin: TAL Reverb-2
The issue exists: Yes

@Paul-Licameli @pietro68

Ryan-Miller-Audacity:
My findings are the same as others. The issue only seems to occur for TAL Reverb 2, even testing other plugins from TAL did not show this issue.

Commit: [https://github.com/audacity/audacity/commit/852a4b4de6cce99d549a33e974a83d38c923206a 852a4b4]
GH job: https://github.com/audacity/audacity/actions/runs/3429017268
OS: Windows 10

Plugin: MReverb by Melda Productions
The issue exists: No

Plugin: SPL Mo-Verb Plus by Plugin Alliance
The issue exists: No

Plugin: TAL Dub III  by Togu Audio Line (TAL)
The issue exists: No

Plugin TAL-Chorus-LX by Togu Audio Line (TAL)
The issue exists: No

[...]

PR #3955, opened 2022-11-18
Issue 3843

Paul-Licameli:
Resolves: #3843

Do initial powering-on of VST Effect Instance whenever constructed, not only later when dialog opens.

This makes TAL reverb correctly store the desired settings, so a change of preset is heard correctly
at the first playback.

Also, not strictly needed to fix the issue: remove unused functions, and don't mutate the settings that are
the source of complete messages.

Recommended:


pietro68 (approved):
Code changes understood and fine by me.

PR #3956, opened 2022-11-18
VST3 UI close crash fix

vsverchinsky:
STR:
1. Create track
2. Add VST3 effect to the stack
3. Open effect UI
4. Close effect UI

Observe: crash

Recommended:

### DIFF
--- a/src/effects/EffectUI.h
+++ b/src/effects/EffectUI.h
@@ -126,7 +126,6 @@ private:
    std::weak_ptr<RealtimeEffectState> mwState{};
    // Temporary state used for destructive processing
    std::shared_ptr<RealtimeEffectState> mpTempProjectState {};
-   std::unique_ptr<EffectUIValidator> mpValidator;
 
    RegistryPaths mUserPresets;
    bool mInitialized{ false };
@@ -162,6 +161,8 @@ private:
 
    const std::shared_ptr<EffectInstance> mpInstance;
    const EffectOutputs *const mpOutputs;
+
+   std::unique_ptr<EffectUIValidator> mpValidator;
 
    DECLARE_EVENT_TABLE()
 };


src/effects/VST3/VST3UIValidator.cpp
```diff
@@ -69,7 +69,6 @@ VST3UIValidator::VST3UIValidator(wxWindow* parent, VST3Wrapper& wrapper, EffectB
 
 VST3UIValidator::~VST3UIValidator()
 {
-   mWrapper.ParamChangedHandler = {};
```
Paul-Licameli, line 72: mWrapper was a dangling reference?  How did that happen?  I thought the lifetime of the validator was guaranteed to be nested inside that of the instance.

This may be an easy fix, but then the destructor isn't complementing what happened in the constructor.  I want to understand why this line doesn't work.

Paul-Licameli: I can't reproduce the problem at e3d0b1ddf9a2363ff3ab65fa23e21e6289fd857d

vsverchinsky: > mWrapper was a dangling reference?

Seem so, I'm experiencing the crash on Windows, 100% reproducible

Paul-Licameli: Try this alternative fix:  Move the member `EffectUIHost::mpValidator` after `mpInstance`.

If that fix works, I prefer it.

vsverchinsky: Yes it works, thanks!

Paul-Licameli (approved):

Ryan-Miller-Audacity:
Unfortunately I wasn't able to reproduce the initial reported behavior in the previous build; however, I tested the [fix build](https://github.com/audacity/audacity/actions/runs/3498681045) on both Windows and MacOS with various VST3 effect plugins and observed no crashes around opening and closing (realtime and destructively) them.

Issue #3902, opened 2022-11-08
Slow screen refresh while dragging VST2 effect sliders or dials


Paul-Licameli:
### Steps to reproduce

1. Generate sound
2. Put DeeGain in the stack (for instance), VST, VST3, or AudioUnit version
(3. Open the dialog and options, and change to the plain UI option, close, reopen; or, don't)
4. Turn computer volume down so you are not distracted
5. Play, while dragging the slider; and also stop dragging the slider
6. Watch the smoothness of updating of the green play head line

### Expected behavior

The smoothness of the screen refresh is the same whether dragging the slider or not

### Actual behavior

For VST2, slightly slower refresh when dragging the slider, in plain or native UI.

For other versions, it's OK

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

dozzzzer:
Verified on macOS and Windows 10

PR #3961, opened 2022-11-19
Vst refresh rate

Paul-Licameli:
Resolves: #3902

Fix slow refresh rate while dragging sliders during playback of VST2 effects.

Mainly it works by not fetching settings from the instance in idle time, while still
presrving stickiness and persistency of sliders by other means.

Recommended:

### DIFF
--- a/libraries/lib-components/EffectInterface.cpp
+++ b/libraries/lib-components/EffectInterface.cpp
@@ -34,6 +34,10 @@ void SimpleEffectSettingsAccess::Set(EffectSettings &&settings,
    mSettings = std::move(settings);
 }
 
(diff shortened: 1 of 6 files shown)


Paul-Licameli:
See also #3904

pietro68:
Repeated test matrix for top commit (fdc6b69) and all cases pass, for DeeGain/BlueCat Chorus/TAL reverb

[vst2-tests.pdf](https://github.com/audacity/audacity/files/10055280/vst2-tests.pdf)

pietro68 (approved):
Changes understood and tested ok.

Issue #3889, opened 2022-11-02
StudioRack by Waves (AU): presets get reset to defaults when re-opening the plugin's dialog


dozzzzer:
### Plugin

StudioRack by Waves

### Plugin Format

Audio Unit

### Steps to reproduce

1. add the plugin from the sidebar
2. open its dialog
3. select a preset
4. close the dialog and re-open
5. observe

@Paul-Licameli : we don't even have stickiness; once that is fixed, persistency too (saving, closing project, reloading) should
be required of a fix.

Note that there are also two ways to change the preset:  Audacity's own manage menu, and the choice controls (or arrow
buttons) in the native UI.  Test both.

### Expected behavior

The selected preset remain in place

### Actual behavior

The plugin's parameters got reset to the defaults, no preset is selected

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

macOS

### Architecture

Apple Silicon (M1, M2, ...)

Paul-Licameli:
@dozzzzer I have downloaded this and tried the mono version on a mono track of white noise with default settings.

I get a choppy "staccato" effect when I play it.  Do you observe the same?

Maybe that merits an issue separately.

Ryan-Miller-Audacity:
Testing AU effects in this build including presets (from both audacity's menu as well as interface menus), stickieness, persistency, realtime vs destructive etc. appears to be working correctly. Presets are properly loaded and maintained for a variety of plugins via both Audacity menu and plugin menu. 

Stickieness and persistency for the specific plugin above appears to be working as expected as well and the interface (preset) is no longer reset when closing and opening the plugin interface.

As Paul noted in his PR, "However some things in that complicated interface do not yet stick, including macro button
assignments." I observed this as well during testing, but noted the other portions of the plugin maintaining properly.

Tested on MacOS Monterey 12.5. with build: https://github.com/audacity/audacity/actions/runs/3515924195

Ryan-Miller-Audacity:
Verified using https://github.com/audacity/audacity/actions/runs/3531807253 on MacOS.

Further checking of studiorack shows improvement to the plugin. Presets are loaded properly and are not reset upon opening/closing the interface or the app. Presets are working properly from both the audacity menu as well as the built in.

I did not notice any residual issue of changing presets during playback causing any issues, this worked as expected.

PR #3967, opened 2022-11-21
Audio unit presets

Paul-Licameli:
Resolves: #3889

This extends #3953

Settings for AudioUnits includes the last chosen factory preset, plus other parameter
changes accumulated after that.

This fixes stickiness and persistency of choices of factory presets for the Waves Studio
Rack plug-in.

However some things in that complicated interface do not yet stick, including maro button
assignments.

Recommended:


pietro68 (approved):
Verified the fix and code changes are fine for me.

Issue #3915, opened 2022-11-10
VST2: Melda plugins cause hang/crash when you change preset during playback


Ryan-Miller-Audacity:
### Bug description

If you attempt to change the preset on a Melda Plugin (officially tested with MCompressor and MBandPass) while playback is active, Audacity will hang and eventually have to be force closed.

These seems to be reproducible with any of the Melda Plugin suite. VST2 and Windows only. This issue is not reproducible with VST3 or audiounit version of the plugins on Windows or MacOS.

### Steps to reproduce

1. Install the Melda Plugin Suite (including VST2 versions): https://www.meldaproduction.com/MFreeFXBundle
2. Open Audacity, generate some sound and add a Melda plugin to the effects stack
3. Open the Plugin interface make some adjustments and save a user preset
4. Start playback
5. Use the presets menu to recall the defaults
6. Audacity hangs

The same is true if you have the defaults active and attempt to load a user preset.

### Expected behavior

Plugin defaults/preset is loaded during playback without hang/crash.

### Actual behavior

Audacity hangs when attempting to load a plugin preset during playback.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

pietro68:
What version of Melda did you use? 15.01 or 16.01?

Ryan-Miller-Audacity:
> What version of Melda did you use? 15.01 or 16.01?

16.01 is the currently installed version on my machine, so that's what I used. I do still want to uninstall those and test this one as well as some of the other Melda issues to confirm if they happen with only one or both versions of this.

pietro68:
I tried both with 16.01 and 15.01.
With 16.01 I would experience an app freeze every time, with 15.01 only like 25% of the times.
With both versions I can see a lot of exceptions being printed in the output window in visual studio, coming from the plugin itself.

I tried to step-debug; the freeze does not happen when executing code conatined in our vst2 host code. It looks like the control comes back to the app's event loop (after executing the load preset code) and from there the freeze will happen.

Paul-Licameli:
To understand a freeze, you may need to interrupt and examine simultaneous call stacks of threads.

Is the main thread waiting on a mutex or condition variable?

pietro68:
> To understand a freeze, you may need to interrupt and examine simultaneous call stacks of threads.
> 
> Is the main thread waiting on a mutex or condition variable?

[...]

PR #3970, opened 2022-11-21
Fix for freeze when loading a preset with Melda Plugins

pietro68:
Resolves: #3915

I know this is not a proper fix - it works though, and it unlocks many plugins.

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1,4 +1,4 @@
-/**********************************************************************
+ /**********************************************************************
 
   Audacity: A Digital Audio Editor
 
@@ -1236,7 +1236,14 @@ bool VSTEffectInstance::RealtimeProcessStart(MessagePackage& package)
    auto& message = static_cast<VSTEffectMessage&>(*package.pMessage);
 
    auto &chunk = message.mChunk;
-   if (!chunk.empty()) {
+
+   // Do not set the chunk if the plugin is by Melda and we are called
+   // in the audio thread - otherwise the whole app will freeze.
+   const bool IsMeldaPlugin = (mVendor == "MeldaProduction");
+   const bool IsAudioThread = (mMainThreadId != std::this_thread::get_id());
+   
+   if (!chunk.empty() && !(IsMeldaPlugin && IsAudioThread) )
+   {
       // Apply the chunk first
 
       VstPatchChunkInfo info = {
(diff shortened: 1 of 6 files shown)


src/effects/VST/VSTEffect.cpp
```diff
@@ -1236,7 +1236,14 @@ bool VSTEffectInstance::RealtimeProcessStart(MessagePackage& package)
...
 
    auto &chunk = message.mChunk;
-   if (!chunk.empty()) {
+
+   // Do not set the chunk if the plugin is by Melda and we are called
+   // in the audio thread - otherwise the whole app will freeze.
+   const bool IsMeldaPlugin = (mVendor == "MeldaProduction");
+   const bool IsAudioThread = (mMainThreadId != std::this_thread::get_id());
+   
+   if (!chunk.empty() && !(IsMeldaPlugin && IsAudioThread) )
```
Paul-Licameli, line 1245: I don’t like this hack.

changing between presets while playing and in plain UI — I predict, will stop working.

Did you look for possibilities of deadlock on a mutex it critical section?  Did you determine whether callSetChunk calls back to AudioMaster?

Our understanding of the problem is not good enough.

Paul-Licameli: Another possibility:  power the effect off if it was on.  Set the chunk.  Power it on again if it was on.

Make no test of the vendor.  See if TAL reverb still works correctly that way.

pietro68: You are correct, changing presets while playing and in plain UI stops working with this - the sound will be updated but not the slider positions.

I determined that callSetChunk does call back to AudioMaster, but it does so in an unrepeatable way, please see my comments on the issue for details.

I did not look for possibilities of a deadlock on a mutex in a critical section, honestly I have no idea how to do that? callSetChunk places multiple calls to  ::callDispatcher, and inside there there is a lock on a critical section (implemented using `wxCRIT_SECT_LOCKER`), but afaics that critical section is only there.

Paul-Licameli: Deadlock might be obvious if you interrupt in the debugger while the program is frozen and view all effect stacks.  See which threads are waiting on sync objects or perhaps spinning in a wait loop.

Paul-Licameli: Another possibility.  The scope of mutual exclusions may need to be wider.  The critical section may need to be held across the duration of all the dispatcher calls done as steps of setting the chunk.

To avoid the thread deadlocking itself, change the wxCRIT object to std::recursive_mutex

pietro68: Update: I inserted print statements before ("lock") and after ("release") the locking of the critical section. When the freeze happens, I see a lock but no release (i.e. I do not see two consecutive locks, that's what I was searching for)

pietro68: I also tried the poweroff - setchunk - poweon suggestion, did not work.

PR #3971, opened 2022-11-22
Issue 3915 attempt

### LINKED ISSUE #3915: VST2: Melda plugins cause hang/crash when you change preset during playback

pietro68:
Failed attempt to resolve #3915

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1236,6 +1236,8 @@ bool VSTEffectInstance::RealtimeResume()
 
 bool VSTEffectInstance::RealtimeProcessStart(MessagePackage& package)
 {
+   recm.lock();
+
    if (!package.pMessage)
       return true;
 
@@ -1313,6 +1315,8 @@ size_t VSTEffectInstance::RealtimeProcess(size_t group, EffectSettings &settings
 
 bool VSTEffectInstance::RealtimeProcessEnd(EffectSettings &) noexcept
 {
+   recm.unlock();
+
    return true;
 }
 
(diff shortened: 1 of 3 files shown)


src/effects/VST/VSTEffect.cpp
```diff
@@ -2347,7 +2351,10 @@ intptr_t VSTEffectWrapper::callDispatcher(int opcode,
                                    int index, intptr_t value, void *ptr, float opt)
 {
    // Needed since we might be in the dispatcher when the timer pops
-   wxCRIT_SECT_LOCKER(locker, mDispatcherLock);
+   //wxCRIT_SECT_LOCKER(locker, mDispatcherLock);
+
+   std::lock_guard<std::recursive_mutex> lk(recm);
```
Paul-Licameli, line 2356: I think you can simplify in C++17 with CTAD -- you can leave out the angle brackets

PR #3974, opened 2022-11-23
Vst2 getprocesslevel

pietro68:
Resolves: failed attempt to solve 3915, but we might want to keep these changes anyway

Recommended:


src/effects/VST/VSTEffect.h
```diff
@@ -403,6 +402,8 @@ class VSTEffect final
 
    EffectSettings MakeSettings() const override;
 
+   virtual int GetProcessLevel() override;
```
Paul-Licameli, line 405: Preferred style is to omit virtual declaring an override

Issue #3883, opened 2022-11-01
Label loses focus when typing on Mac OS Ventura


nelclem:
### Bug description

When typing text into a label, it loses focus after the first character is pressed.
If the next character correlates to a keyboard shortcut, that action is invoked.

### Steps to reproduce

1. "Ctrl + b" to create label.
2. Type.
3. When the first character is pressed, the focus is no longer on the label text.

### Expected behavior

For label text to remain in focus until Enter is pressed.

### Actual behavior

Focus goes away from label text area.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

macOS

dozzzzer:
I was able to reproduce the bug on my machine (macOS Ventura)

billw58:
This seems to be a symptom of a wider problem of the Audacity project window losing focus.
Audacity 3.2.1, macOS 13.0.1, tested on 2019 iMac (x86) and MacBook Air (M1)
Steps:
New project
Generate audio
After closing the generate dialog, project window does not have focus (window widgets at upper left are greyed out, no yellow focus border on generated track)
Click in track to get focus back
Select the track and apply the Fade In effect
As above, project window does not have focus

doctoratlantis:
Same issue - you can sort of work around it by right-clicking on the label track and select the edit dialog.  However, this makes the process of quickly making notes in the software very tedious.

M-D-M:
Can confirm, this is happening to me, as well.   

And not just the first click loses focus, but the second click as well.  It takes three clicks total to type into a label.

dozzzzer:
Verified on macOS Ventura

PR #3975, opened 2022-11-23
Delay/Avoid db block purge dialog creation

vsverchinsky:
Resolves: #3883

It's a general improvement. 
Dialog show caused focus lose on `TrackView`, but dialog lifetime might be too short for it to be presented to the user.

Recommended:


Paul-Licameli (approved):
Add comments in the code explaining the motivation for the change.

Nothing wrong with the changes.

PR #3976, opened 2022-11-23
Fix freezing melda plugins

Paul-Licameli:
Resolves: #3915

Melda plug-ins appear to need event dispatch to continue on the main thread, for
any other thread to make progress while setting the chunk.  (No, returning a more
proper Process Level value did not help).

Not sure whether to blame the plug-in, but this workaround may do it.

But I have some doubts about other consequences of yields to the event loop.  It
can cause some surprising unexpected recursions to our own event handlers.

Recommended:


pietro68:
I tried this attempt to fix, but unfortunately it does not work, I still get hangs when loading a preset while playing.

Issue #3853, opened 2022-10-26
Opening Melda VST plugins during playback causes hang


Ryan-Miller-Audacity:
### Bug description

Attempting to open a Melda Productions VST while playback is active causes Audacity to hang.

I did not observe this behavior with any other VSTs outside of the Melda Productions group and it seems to effect any of the medla productions plugins (tested with MReverb, MComb, MCompressor etc.). There is no sentry crash dialog either. Audacity hangs until you click exit and you get a non-responsive dialog. 

I was able to reproduce this behavior on Windows, but not MacOS. VST3 versions of the plugins work as expected.

### Steps to reproduce

1. Install the Melda Productions free bundle from https://www.meldaproduction.com/MFreeFXBundle (be sure to install the VST version)
2. Open Audacity and grab some sound
3. (I find it easiest to edit preferences and only allow VSTs to avoid confusion in the realtime menu) Add a Melda Productions VST plugin to the RT stack
4. Start Playback
5. Click the plugin to open its interface
6. Audacity hangs

### Expected behavior

Melda VST interface opens and displays without hanging during playback.

### Actual behavior

Attempting to open Melda VST interface during playback causes hanging.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

Ryan-Miller-Audacity:
@Paul-Licameli 

After retesting on the most recent VST2 build (Build #2823 https://github.com/audacity/audacity/actions/runs/3346486868) I am still able to reproduce this on Windows with various Melda Plugins (MComb, MEqualizer, MReverb etc.); however, Testing on Mac, I was unable to reproduce this crash.

pietro68:
I can not reproduce this bug with version 15.01 of the Melda plugins, but I can reproduce it with version 16 (i.e. the latest available)

pietro68:
I tried it with the latest available version (16.01) - plugins of this version seem to be unstable, for instance when equalizer is loaded (even before playing or showing the gui) many exceptions are thrown:

![image](https://user-images.githubusercontent.com/16491307/199001731-ff669f3c-21b1-4e7c-8110-8be8c583aabc.png)

pietro68:
Some interesting information: I tried Melda MEqualizer and MCompressor on windows, on release branch commit ca981b6.

1. MCompressor will never hang
2. MEqualizer will hang
[...]

PR #3979, opened 2022-11-24
Issue #3853 - opening Melda vst2 plugins while playing causes hang

pietro68:
Resolves: #3853 

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -2353,7 +2353,8 @@ intptr_t VSTEffectWrapper::callDispatcher(int opcode,
                                    int index, intptr_t value, void *ptr, float opt)
 {
    // Needed since we might be in the dispatcher when the timer pops
-   wxCRIT_SECT_LOCKER(locker, mDispatcherLock);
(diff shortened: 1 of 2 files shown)


Paul-Licameli:
I am very skeptical that this PR would work while my #3976 failed to fix it.  That PR was based on the same commit as this — so that is not an uncontrolled variable.

I am very skeptical that just changing the type of synchronization object makes a difference in resolving the deadlock.

I do not know why std::mutexwould work but std::recursive_mutex would not work to resolve a deadlock.

Please double check your testing of the use case with and without the changes.  Be sure you are really hitting a breakpoint in VSTEffect::MakeInstance.

pietro68:
I am really confused. I thought I tried it successfully with MEqualizer yesterday, but I tried it again today and the bug is still there. On the contrary, Sergey tells me that he can not reproduce the bug even using MEqualizer, but maybe different factors are at play (me using Windows10, him using Windows11 - or maybe different audio drivers?) 

Anyway, if it does not work for me anymore, then I will close this PR.

> I do not know why std::mutexwould work but std::recursive_mutex would not work to resolve a deadlock
however please note: when we tried that, it was for a different bug i.e. when loading a preset during play

Issue #3256, opened 2022-07-19
'Audacity quit unexpectedly' when discovering a plugin that requires authorization


dozzzzer:
**Describe the bug**
'Audacity quit unexpectedly' when discovering a plugin that requires authorization (e.g. Soothe2 by Oeksound)

**To Reproduce**
1. Install a demo version of a plugin and let the trial expire (e.g. https://oeksound.com/plugins/soothe2/)
2. Purge audacity's pluginregistry.cfg file
3. Run Audacity
4. Observe 

**Actual behavior**
'Audacity quit unexpectedly' error message:

https://user-images.githubusercontent.com/70063970/179743655-a433d173-1a5c-475b-80c0-43b333bf328b.mp4

**Expected behavior**
No 'Audacity quit unexpectedly' error message

**Additional information (please complete the following information):**
 - OS: MacOS 12.2.1
 - latest master

crsib:
I raise the priority to P2. In no situation Audacity should crash during the plugin discovery, that is the reason why we perform plugin validation in a separate process. 

This potentially prevents users from using Audacity at all.

crsib:
After conversation with @vsverchinsky demoting it back to P4

crsib:
Or not :-)

dozzzzer:
Verified on macOS and Windows 11

theirongiant82:
A bunch of izotope plugins generated the following error during analysis: "Can't get core function pointer."

theirongiant82:
Audacity was able to scan or skip the UAD plugins, but many of them are listed twice.

crsib:
> I think the arm64 build is hosed. 

Apple does not allow running unsigned code on ARM when it detects that the application was downloaded. With the universal build x86_64 slice is run. `xattr -cr Audacity.app` will allow you to run the unsigned ARM build.

PR #3980, opened 2022-11-25
Plugin host startup fixes

### LINKED ISSUE #3802: Add a skip option + timeout to plugin validation
<details>
<summary>original report</summary>
### Your idea

Make VST startup scan optional; perhaps with an entry manually added to .cfg file.

### Problem to be solved

Audacity crashes or fails to startup up due to apparently non-conforming VSTs.

### Prior art

_No response_

### Additional context

Please read this this forum thread here: https://forum.audacityteam.org/viewtopic.php?p=456850#p456850
</details>

vsverchinsky:
Resolves: #3256
Resolves: #3802

Disables breakpad handler in plugin host process. Presumably that can fix #3802 on MacOS/Linux.
Hides plugin host process icon from dock on MacOS and prevents native crash report dialog to appear #3256.

Recommended:

### DIFF
--- a/libraries/lib-ipc/IPCClient.cpp
+++ b/libraries/lib-ipc/IPCClient.cpp
@@ -31,6 +31,12 @@ public:
       if(!fd)
          throw std::runtime_error("cannot create socket");
 
+#if defined(__unix__) || defined(__APPLE__)
+      auto fdFlags = fcntl(*fd, F_GETFD, 0);
+      if(fdFlags != -1)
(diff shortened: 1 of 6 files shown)


Paul-Licameli:
The changes in AudacityApp.cpp remind me that ideally the plugin host should be a different, smaller executable with its own main() that doesn't load so much other unnecessary code.  But that would be a future improvement.

crsib (approved):

Issue #3981, opened 2022-11-25
VST2: Plain UI does not update sliders when loading preset while playing


pietro68:
### Bug description

If you use the Plain UI with a vst2 effect, and load a preset while playing is on, the sliders will not be updated right.

### Steps to reproduce

1. have some audio track
2. in the effect stack, instantiate any vst2 effect (even the simplest one, DeeGain)
3. modify the options so to show the plain UI (you will need to close and reopen the gui, if you were not in plain UI mode)
4. Start playback
5. Use the presets menu to load a preset

### Expected behavior

Sliders on the plain UI are positioned according to the preset

### Actual behavior

Sliders on the plain UI will not move from their previous position.

A second selection of any preset, will then make the sliders be updated to reflect the first preset selection.

This was observed on Windows but it very likely happens on MacOS too, it is a logical bug in the control flow.

It happens since a long time, at least since merge 7616b89 in the release branch.

Ryan-Miller-Audacity:
Confirmed the fix in Build from PR #3982 on both Windows and MacOS. Sliders are now updating when a preset is called both in and out of playback mode for user and factory presets. (FYI TDR Kotelnikov was great for testing this since it has so many presets). 

I did notice some issues with the VST2 version of the Valhalla plugins not responding to presets at all, but that seems to be independent of these changes/fix.

PR #3982, opened 2022-11-25
Issue 3981

pietro68:
Resolves: #3981

When a preset is loaded ::realtimeprocessStart will be called; however: 

- when the preset load is done **when playing is stopped**, it will be called in the main thread and then a call to UpdateUI() will follow; but

- when the preset load is done **when playing is on**, first we get the call to UpdateUI() and then the call to ::realtimeprocessStart, but the latter will be **in the audio thread** this time

So here we do this: take a "note" when the second case happens by means of an atomic bool, and act on it during idle time.

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -1234,6 +1234,12 @@ bool VSTEffectInstance::RealtimeResume()
    return true;
 }
 
+
+bool VSTEffectInstance::OnePresetWasLoadedWhilePlaying()
+{
+   return mPresetLoadedWhilePlaying.exchange(false);
+}
+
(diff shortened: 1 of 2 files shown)


Paul-Licameli:
This fails to build on Linux.  Needs
```
#include <atomic>
```
in VSTEffect.h

Paul-Licameli:
Otherwise it looks harmless.  I will approve after the commit is amended to fix Linux build, and force-pushed.

pietro68:
> Otherwise it looks harmless. I will approve after the commit is amended to fix Linux build, and force-pushed.

ok, done that

Paul-Licameli (approved):

PR #3990, opened 2022-11-28
Defer setchunk

pietro68:
Resolves: #3853
Resolves: #3915
Resolves: #3992
Resolves: #3994

Recommended:

### DIFF
--- a/src/effects/RealtimeEffectState.cpp
+++ b/src/effects/RealtimeEffectState.cpp
@@ -234,6 +234,7 @@ struct RealtimeEffectState::Access final : EffectSettingsAccess {
                      stateSettings, pMessage.get()
                   };
                   pInstance->RealtimeProcessStart(package);
+                  pInstance->RealtimeProcessEnd(stateSettings);
                   pAccessState->mLastSettings.settings = stateSettings;
                   return;
                }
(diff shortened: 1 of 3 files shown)


Paul-Licameli (changes requested):
I will just write general comments instead of at every line:

1) The changes of the instance wrapper should be excluded from happening during the sequence of calls to RealtimeProcessStart, RealtimeProcess, RealtimeProcessEnd.  Therefore, it's time for the fix that locks a mutex in Start and unlocks it in End.
2) Do not add another mutex for this purpose.  Instead change mDispatcherLock to `std::recursive_mutex`
3) Add a member function to VSTEffectInstance that decides whether to change how chunks are applied to this alternative.  Make it do the ugly test of the version string for Melda.  If the fix does not apply, the mutual exclusion isn't necessary so the extra lock and unlock might be skipped.  This might avoid unnecessarily punishing the responsiveness of the user interface for other effects.  And other behavior should not then change.
4) The passing of chunk data is very round-about now!   It can be simplified.  Don't wait until idle time.  Instead, apply the chunk (for Melda effects only, and with the mutual exclusion) in `MakeMessageFS`.  Do not skip any of the behavior already in that function.
5) Passing the chunk through the message buffers is still needed to keep the logic of merging and assigning messages correct.
6) For a MeldaEffect, `RealtimeProcessStart` must still check whether there is a chunk in the message, but then only resize it to 0.

Paul-Licameli (approved):

Paul-Licameli (approved):
Re-approved, but when this is merged, squash all into one commit because the first is incomplete and buggy without the second too.

Issue #3860, opened 2022-10-27
Enable VST3 program lists


AnitaBats:
Some plugins use VST3's Program Lists as a storage for factory presets. Such programs should be listed in `Presets & Settings -> Factory Presets` menu

Issue #3414, opened 2022-08-23
Factory presets not properly populated in hamburger menu for VST3


Ryan-Miller-Audacity:
### Bug description

For VST3 versions of plugins, the available factory presets are not being properly populated in the hamburger menu. This does work correctly for AUs. 

For Example Auburn Sounds Couture plugin (https://www.auburnsounds.com/products/Couture.html): 

Audiounit module: 
<img width="641" alt="Screen Shot 2022-08-23 at 7 55 10 AM" src="https://user-images.githubusercontent.com/88568755/186152601-3b595a9e-597e-4ade-a776-cfe6568b0930.png">

VST3 module: 
<img width="797" alt="Screen Shot 2022-08-23 at 7 55 36 AM" src="https://user-images.githubusercontent.com/88568755/186152643-ad7be580-9aec-4bfa-8651-893aed3b6930.png">

Note: Due to #3292, you may need to disable Audiounit family in the preferences to properly instantiate the VST3 version of the plugin.

### Steps to reproduce

1. Open Audacity and grab/generate some sound
2. Select the Audiounit version of a plugin that has both a VST3 and Audiounit version and includes factory presets outside the plugin interface (Valhalla Supermassive etc. include presets within their interface)
3. Select the hamburger menu and observe the full list of available factory presets.
4. Select the VST3 Version (you may need to disable Audiounits from the effects page within preferences) of the plugin used previously
5. Select the hamburger menu and observe that there are no options (other than default) listed under the factory presets menu.

### Expected behavior

Factory presets should be properly loaded within the hamburger menu for VST3 plugins.

### Actual behavior

Factory presets are not being properly loaded within the hamburger menu for VST3 plugins.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

Ryan-Miller-Audacity:
FYI @Paul-Licameli @vsverchinsky

Ryan-Miller-Audacity:
Listing as a P2 as a lot of people (myself included) turn to these factory presets to make using some more complicated plugins less of a headache and this would be a great inconvenience to the end user.

vsverchinsky:
It may be that factory presets in VST3 isn't same thing as in AU. For this specific plugin I didn't find any presets to be installed under any of expected presets locations (https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Locations+Format/Preset+Locations.html)
[...]

PR #3996, opened 2022-11-30
VST3 factory presets

vsverchinsky:
Resolves: #3860

Recommended:

### DIFF
--- a/src/effects/VST3/VST3Instance.cpp
+++ b/src/effects/VST3/VST3Instance.cpp
@@ -32,11 +32,11 @@ namespace
    }
 }
 
-VST3Instance::VST3Instance(const PerTrackEffect& effect, VST3::Hosting::Module& module, VST3::UID effectUID)
-   : Instance(effect), mEffectUID(effectUID)
+VST3Instance::VST3Instance(const PerTrackEffect& effect, VST3::Hosting::Module& module, const VST3::Hosting::ClassInfo& effectClassInfo)
+   : Instance(effect)
 {
    ReloadUserOptions();
-   mWrapper = std::make_unique<VST3Wrapper>(module, mEffectUID);
+   mWrapper = std::make_unique<VST3Wrapper>(module, effectClassInfo);
    mWrapper->InitializeComponents();
 }
 
@@ -58,7 +58,7 @@ bool VST3Instance::RealtimeAddProcessor(EffectSettings& settings,
    // Assign another instance with independent state to other processors
    auto &effect = static_cast<const PerTrackEffect&>(mProcessor);
    auto uProcessor =
-      std::make_unique<VST3Instance>(effect, mWrapper->GetModule(), mEffectUID);
+      std::make_unique<VST3Instance>(effect, mWrapper->GetModule(), mWrapper->GetEffectClassInfo());
    if (!uProcessor->RealtimeInitialize(settings, sampleRate))
       return false;
    mProcessors.push_back(move(uProcessor));
--- a/src/effects/VST3/VST3Instance.cpp
+++ b/src/effects/VST3/VST3Instance.cpp
@@ -37,6 +37,7 @@ VST3Instance::VST3Instance(const PerTrackEffect& effect, VST3::Hosting::Module& 
 {
    ReloadUserOptions();
    mWrapper = std::make_unique<VST3Wrapper>(module, mEffectUID);
+   mWrapper->InitializeComponents();
 }
 
 VST3Instance::~VST3Instance() = default;
@@ -148,7 +149,7 @@ bool VST3Instance::ProcessFinalize() noexcept
 
 bool VST3Instance::ProcessInitialize(EffectSettings &settings, double sampleRate, ChannelNames chanMap)
 {
-   if(mWrapper->Initialize(settings, sampleRate, Steinberg::Vst::kRealtime, mProcessingBlockSize))
+   if(mWrapper->Initialize(settings, sampleRate, Steinberg::Vst::kOffline, mProcessingBlockSize))
    {
       mInitialDelay = mWrapper->GetLatencySamples();
       return true;
--- a/src/effects/VST3/VST3Utils.cpp
+++ b/src/effects/VST3/VST3Utils.cpp
@@ -122,6 +122,27 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
    return false;
 }
 
+wxString VST3Utils::MakeFactoryPresetID(Steinberg::Vst::UnitID unitId, Steinberg::int32 programIndex)
+{
+   return wxString::Format("%d:%d",
+      static_cast<int>(unitId),
+      static_cast<int>(programIndex));
+}
+
+bool VST3Utils::ParseFactoryPresetID(const wxString& presetId, Steinberg::Vst::UnitID& unitId, Steinberg::int32& programIndex)
+{
+   auto parts = wxSplit(presetId, ':');
+   long nums[2]{};
+   if(parts.size() == 2 && parts[0].ToLong(&nums[0]) && parts[1].ToLong(&nums[1]))
+   {
+      unitId = static_cast<Steinberg::Vst::UnitID>(nums[0]);
+      programIndex = static_cast<Steinberg::int32>(nums[1]);
+      return true;
+   }
+   return false;
+}
+
+
 wxString VST3Utils::GetFactoryPresetsPath(const VST3::Hosting::ClassInfo& effectClassInfo)
 {
    return GetPresetsPath(
--- a/src/effects/VST3/VST3Utils.cpp
+++ b/src/effects/VST3/VST3Utils.cpp
@@ -15,11 +15,57 @@
 #include "Base64.h"
 
 #include <wx/string.h>
+#include <wx/filename.h>
 #include <wx/sizer.h>
 
 #include <pluginterfaces/vst/ivsteditcontroller.h>
 #include <pluginterfaces/vst/ivstparameterchanges.h>
+#include <wx/regex.h>
 
+#include "MemoryX.h"
+
+
+#ifdef __WXMSW__
+#include <shlobj.h>
+#endif
+
+namespace
+{
+   
+wxString GetFactoryPresetsBasePath()
+{
+#ifdef __WXMSW__
+   PWSTR commonFolderPath { nullptr };
+   auto cleanup = finally([&](){ CoTaskMemFree(commonFolderPath); });
+   if(SHGetKnownFolderPath(FOLDERID_ProgramData, KF_FLAG_DEFAULT , NULL, &commonFolderPath) == S_OK)
+      return wxString(commonFolderPath) + "\\VST3 Presets\\";
+   return {};
+#elif __WXMAC__
+   return wxString("Library/Audio/Presets/");
+#elif __WXGTK__
+   return wxString("/usr/local/share/vst3/presets/");
+#endif
+}
+
+wxString GetPresetsPath(const wxString& basePath, const VST3::Hosting::ClassInfo& effectClassInfo)
+{
+   wxRegEx fixName(R"([\\*?/:<>|])");
+   wxString companyName = wxString (effectClassInfo.vendor()).Trim();
+   wxString pluginName = wxString (effectClassInfo.name()).Trim();
+
+   fixName.ReplaceAll( &companyName, { "_" });
+   fixName.ReplaceAll( &pluginName, { "_" });
+
+   wxFileName result;
+   result.SetPath(basePath);
+   result.AppendDir(companyName);
+   result.AppendDir(pluginName);
+   auto path = result.GetPath();
+
+   return path;
+}
+
+}
 
 wxString VST3Utils::MakePluginPathString(const wxString& modulePath, const std::string& effectUIDString)
 {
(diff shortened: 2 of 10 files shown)


src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -240,12 +240,9 @@ class ComponentHandler : public Steinberg::Vst::IComponentHandler
...
          mWrapper.mEditController->getParameterInfo(i, info);
 
-         if (info.flags & (ParameterInfo::kIsHidden | ParameterInfo::kIsProgramChange))
+         if ((info.flags & ParameterInfo::kIsReadOnly) != 0)
             continue;
-
-         if ((info.flags & ParameterInfo::kCanAutomate)  == 0)
-            continue;
-
+
```
Paul-Licameli, line 275: It's not obvious locally in the code here, but I understand a consequence of this is that changes of factory preset will "dirty" the project so you are prompted to save it when you close.  Am I right?  Is there another reason for this?

src/effects/VST3/VST3Instance.cpp
```diff
@@ -153,7 +154,7 @@ bool VST3Instance::ProcessFinalize() noexcept
 
 bool VST3Instance::ProcessInitialize(EffectSettings &settings, double sampleRate, ChannelNames chanMap)
 {
-   if(mWrapper->Initialize(settings, sampleRate, Steinberg::Vst::kRealtime, mProcessingBlockSize))
+   if(mWrapper->Initialize(settings, sampleRate, Steinberg::Vst::kOffline, mProcessingBlockSize))
```
Paul-Licameli, line 157: I understand the second of the six commits, except for this unexplained change.  What does it accomplish?

src/effects/VST3/VST3Wrapper.h
```diff
@@ -114,6 +123,10 @@ class VST3Wrapper
    void BeginParameterEdit(EffectSettingsAccess& access);
    void EndParameterEdit();
 
+   //! Returns an array of factory preset ids.
+   //! Safe to call before `InitilizeComponents`
```
Paul-Licameli, line 127: "InitializeComponents"

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -544,28 +552,61 @@ void VST3Wrapper::StoreSettings(EffectSettings& settings) const
...
+   using namespace Steinberg;
+
+   auto fileStream = owned(Vst::FileStream::open(presetId.c_str(), "rb"));
+   if(!fileStream)
+      throw FileException(FileException::Cause::Open, presetId);
+
+   if(!LoadPresetFromStream(fileStream))
+   {
+      throw SimpleMessageBoxException(
+         ExceptionType::Internal,
```
Paul-Licameli, line 690: Is `Internal` the right type for this and the other new exception, not `BadEnvironment`?  If this failure occurs, do we want Sentry data about it or not?

`FileException` does use `Internal` but see the comment in its constructor explaining that as an unusual decision.

src/effects/VST3/VST3Utils.cpp
```diff
@@ -122,6 +122,27 @@ bool VST3Utils::ParseAutomationParameterKey(const wxString& key, Steinberg::Vst:
    return false;
 }
 
+wxString VST3Utils::MakeFactoryPresetID(Steinberg::Vst::UnitID unitId, Steinberg::int32 programIndex)
+{
+   return wxString::Format("%d:%d",
```
Paul-Licameli, line 127: Are you certain ':' will not occur in either of the substituted strings?  The inverse function assumes this.

If you doubt it, maybe the utility functions `wxSplit` and `wxJoin` can be used, to easily escape  certain characters.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -562,8 +563,14 @@ void VST3Wrapper::InitializeComponents()
       mEditController->setComponentState(&stateStream);
    }
 
-   mDefaultSettings = MakeSettings();
-   StoreSettings(mDefaultSettings);
+   {
+      auto cache = GetCache(mEffectClassInfo.ID());
+      if(!cache->defaultSettings.has_value())
```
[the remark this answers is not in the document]
Paul-Licameli: In fact `GetCache` is not now guaranteeing a non-null return, so please null check this.

Paul-Licameli (dismissed):

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -28,6 +28,32 @@ constexpr auto processorStateKey  = wxT("ProcessorState");
...
+      return &it->second;
+   return nullptr;
+}
+
+VST3PluginCache& CreateCache(const VST3::UID& effectUid)
+{
+   const auto key = effectUid.toString();
+   assert(sVST3PluginCache.find(key) == sVST3PluginCache.end());
+   const auto result = sVST3PluginCache.insert(std::pair { key, VST3PluginCache {} });
+   return result.first->second;
```
Paul-Licameli, line 55: Why not simply combine the two functions ?  Just let `GetCache` return a reference to `sVST3PluginCache[key]` and that will create and default-initialize a member that isn't there already.

Then the assertion isn't needed, and you know I don't like assertions without comments:  when I review an assertion,. I'm not satisfied until I also review all uses of the function.  Less burden of review without it.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -416,6 +442,66 @@ VST3Wrapper::VST3Wrapper(VST3::Hosting::Module& module, const VST3::Hosting::Cla
    mEditController = editController;
 
    mEditController->initialize(&AudacityVst3HostApplication::Get());
+
+   if(GetCache(mEffectClassInfo.ID()) == nullptr)
```
Paul-Licameli, line 447: To make my last suggestion work, defaulted `VST3PluginCache` must be distinguishable from one that was initialized as in the following code, so you can write another test here.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -416,6 +442,66 @@ VST3Wrapper::VST3Wrapper(VST3::Hosting::Module& module, const VST3::Hosting::Cla
...
+         Vst::ParameterInfo parameterInfo {};
+         if(mEditController->getParameterInfo(parameterIndex, parameterInfo) != kResultOk)
+            continue;
+
+         if(parameterInfo.unitId != Vst::kRootUnitId)
+            continue;
+         if((parameterInfo.flags & Vst::ParameterInfo::kIsProgramChange) == 0)
+            continue;
+
+         cache.rootUnitProgramChangeParameterID = parameterInfo.id;
```
Paul-Licameli, line 466: ... and I think this assignment makes that distinguishability.

src/effects/VST3/VST3Wrapper.cpp
```diff
@@ -564,7 +681,7 @@ void VST3Wrapper::LoadPreset(const wxString& presetId)
    {
       throw SimpleMessageBoxException(
          ExceptionType::Internal,
-         XO("Unable to apply VST3 preset file %s").Format(presetId),
+         XO("Unable to apply VST3 preset %s").Format(presetId),
```
Paul-Licameli, line 691: This is a change of a user facing string that I suppose you must tell design about, if there is any important opinion about string consistency.

Paul-Licameli:
I'm done reviewing.  Mostly good.  But I had some questions.

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: It is correct, though formally there could be other types of parameter that aren't automatable or hidden. For such parameters it is correct to "dirty" the project too I think.

src/effects/VST3/VST3Instance.cpp
vsverchinsky: Some VST3 plugins can offer better processig quality in `kOffline` mode. I think it used to be `kOffline` some time before refactoring, but I've made a typo

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: > Is Internal the right type for this and the other new exception, not BadEnvironment? If this failure occurs, do we want Sentry data about it or not?

Not sure how to answer the first question, but I don't think that it's a good idea to send preset loading failures to sentry. `LoadPresetFromStream(fileStream)` load preset file from the disc, and user may attempt to load preset that was genrated by older version of same plugin, or even by another plugin.

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: The answer is in VST3Wrapper constructor - cache initialization process. If `GetCache` and `CreateCahce` were combined I had to add an initialization state. Reference type might have convinced you that return value isn't nullptr (though that can be violated too), and the code is formally correct, but you must make sure that the state of returned object is correct too. I don't think that it worth changing.

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: Yes, and all other places should check that too. Not obvious what should happen if `GetCache` returns not initialized object

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: Right, thanks!

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: Yes, the meaning for `kNoProgramListId` is that plugin does not have program list parameter, which is fine.

src/effects/VST3/VST3Instance.cpp
Paul-Licameli: So then, will you revert that change or comment that you are keeping it?

src/effects/VST3/VST3Wrapper.cpp
Paul-Licameli: If there should not be Sentry reporting for this error, then Internal is the wrong choice.  See this other example:

```
         GuardedCall(
            [&message, rc] {
            throw SimpleMessageBoxException{ rc != SQLITE_FULL ? ExceptionType::Internal : ExceptionType::BadEnvironment,
               message, XO("Warning"), "Error:_Disk_full_or_not_writable" }; },
            SimpleGuard<void>{},
            [this](AudacityException * e) {
               // This executes in the main thread.
               if (mCallback)
                  mCallback();
               if (e)
                  e->DelayedHandlerAction();
            }
         );
```

If Sqlite fails but tells the program it's exhaustion of device space, then we don't need to collect a report.  Only if Sqlite fails for some other various reason do we want a report.

Paul-Licameli (changes requested):
Still I must request one change, to pass the correct exception type and avoid unnecessary error reports.

Besides that, just one spelling mistake in a comment, and resolve the question whether `kOffline` is right.

src/effects/VST3/VST3Instance.cpp
vsverchinsky: No, I think `kOffline` fits better for destructive processing. `RealtimeInitialize` uses `kRealtime` mode.

src/effects/VST3/VST3Wrapper.cpp
vsverchinsky: Reverted that line

DavidBailes:
@vsverchinsky . It's great to see that this is being worked on so that keyboard users can access the factory presets of vst3 plugins:
https://github.com/audacity/audacity/issues/3414

I've tested the plugin mentioned by that issue: https://www.auburnsounds.com/products/Couture.html, and can confirm that the factory presets are now available.
However, I've tested with TDR Kotelnikov  (https://www.tokyodawn.net/tdr-kotelnikov/), and TDR VOS SlickEQ (https://www.tokyodawn.net/tdr-vos-slickeq/), and the presets are still not available for these. (They are using Reaper.)

Paul-Licameli (approved):

Paul-Licameli:
All my questions are resolve now, BUT, the latest run failed to build for Ubuntu.

But those looked like spurious failures caused by intermittent network problems, so just started a rerun of the failed jobs.  If that passes, then this can be merge.

p0nce:
You can use Graillon free VST3 for VST3 program lists, we are actually getting support issues from Audacity users :) https://www.auburnsounds.com/products/Graillon.html

vsverchinsky:
> @vsverchinsky . It's great to see that this is being worked on so that keyboard users can access the factory presets of vst3 plugins: #3414
> 
> I've tested the plugin mentioned by that issue: https://www.auburnsounds.com/products/Couture.html, and can confirm that the factory presets are now available. However, I've also tested with TDR Kotelnikov (https://www.tokyodawn.net/tdr-kotelnikov/), and TDR VOS SlickEQ (https://www.tokyodawn.net/tdr-vos-slickeq/), and the presets are still not available for these. (They are using Reaper.)

@DavidBailes I've updated PR and plugins mentioned should also list factory programs in the "settings & presets" menu. Though one issue remains - there could be more than one menu entry named 'Defaults', and moreover they could be same, but I'm not sure what to do about it right now (the problem is that for some plugins that may be not true, I think)

DavidBailes:
> @DavidBailes I've updated PR and plugins mentioned should also list factory programs in the "settings & presets" menu. Though one issue remains - there could be more than one menu entry named 'Defaults', and moreover they could be same, but I'm not sure what to do about it right now (the problem is that for some plugins that may be not true, I think)

@vsverchinsky . Thanks, I can confirm that the problem with those two plugins is fixed.
The issue with entries for defaults and default is a minor issue (I notice that it also occurs for vst2 plugins).

Paul-Licameli (approved):
I have reviewed what is new since last review.  I have no comments.

Issue #3999, opened 2022-11-30
Change cursor to grab hand or four-way-move arrow on toolbar grabbers


abhi-kr-2100:
### Your idea

As @LWinterberg mentioned in [this comment](https://github.com/audacity/audacity/issues/3968#issuecomment-1328851658), the cursor should change to "a grab hand (or a 4-way move arrow) to indicate you can grab and move the thing."

### Problem to be solved

Change cursor appropriately to a grab hand or a 4-way-move-arrow when on grabbers.

abhi-kr-2100:
One other issue is the tooltip on toolbar grabbers. I agree with @petersampsonaudacity's [comment](https://github.com/audacity/audacity/issues/3968#issuecomment-1328868802) that having the tooltip might be useful. However, some indication of what action is possible should also be provided (originally said by @LWinterberg).

This issue was not fixed in PR #3968 as translations are not available, [as pointed out](https://github.com/audacity/audacity/issues/3968#issuecomment-1328823015) by @vsverchinsky.

I've the following compromise in mind:

* Translations are available for "X Toolbar", so we can remove the "Audacity" prefix (with some other minor changes).
* We can then append something like "(click and drag to move)". Translation is not available for this. As such, this part will appear in English in all locales. We can hide it in non-English (and non-Turkish and non-Hindi — I can translate those) versions of the app.
* This way, at least English users will benefit from the extra information.

petersampsonaudacity:
> Change cursor appropriately to a grab hand or a 4-way-move-arrow when on grabbers.

I recall that a while back @tantacrul stated that there would be a review of all the various cursors in Audacity as he found several inappropriate.  

_The one that always bugs me is the pointing gloved hand you get when hovering over the edge of a selection to facilitate expanding or contracting the selection._

![image](https://user-images.githubusercontent.com/50205984/205021585-c921dad5-96b4-446b-b3bf-17b3c8a0d7e5.png)

petersampsonaudacity:
> > Change cursor appropriately to a grab hand or a 4-way-move-arrow when on grabbers.
> 
> This sounds sensible to me - a better indicator of purpose.

Also for reference this too was logged a while back as  #2307

LWinterberg:
This is no longer relevant for Audacity 4

PR #4000, opened 2022-11-30
Fix issue #3999: Change cursor to four-way-arrows when on a grabber

abhi-kr-2100:
Signed-off-by: Abhishek Kumar <abhi.kr.2100@gmail.com>

Resolves: https://github.com/audacity/audacity/issues/3999

Changes cursor to four-way-arrows when on a grabber.

Recommended:


abhi-kr-2100:
I would've preferred a hand grab cursor as it seems more modern (and also more appropriately named). There doesn't seem to be a hand grab cursor in `wx/gdicmn.h`. Searching online told me that creating a custom cursor is possible and I would like to try it in the future if other contributors agree.

PR #4002, opened 2022-12-01
Vst modules

Paul-Licameli:
Resolves: #1131
Resolves: #1119
Resolves: #1118
Resolves: #1117
Resolves: #1116
Resolves: #1115
Resolves: #1098
Resolves: #1114
Resolves: #3874

Update:  combined with the Effect-modules branch.  That other PR is closed.  Rebased onto recent master, and conflicts resolved.  Build system changes, formerly part of this, are no longer needed.

Commits exclusive of the last three have already been reviewed and approved by @vsverchinsky and are unchanged besides conflict resolutions.

Seven modules for support of effect plug-in families and for the several built-in effects

The Audacity executable has five fewer dependencies on third-party libararies, as seen in this diagram:

image

Two more effects modules extracted and four more third-party libraries eliminated from Audacity's dependencies.

image

Recommended:


Paul-Licameli:
After Equalization refactoring was merged, this PR required a rebasing to resolve conflicts in the Eq* files only.

vsverchinsky (approved):
Approved with assumption that compilation error will be fixed (is this PR dependent on Equalization PR?). Though one suggestion. Now, when all effect plugin types are managed as modules, having two ways to enable/disable them is confusing, suggest to remove `Enable Effects` section from `Effects` in preferences.

Paul-Licameli:
> Approved with assumption that compilation error will be fixed (is this PR dependent on Equalization PR?). Though one suggestion. Now, when all effect plugin types are managed as modules, having two ways to enable/disable them is confusing, suggest to remove `Enable Effects` section from `Effects` in preferences.

I have rebased onto recent master, resolved a few simple conflicts in #include lines and in the visibility of RealtimeEffectList, fixed the Windows link, and force-pushed again.

I don't think we will follow your suggestion:  instead, I think the extension of the Modules preference page should be hidden in non-alpha builds.

The matter of module extraction needs a little discussion with the rest of the team, and again, I move this card back to review-in-progress until I get @crsib to agree to it as well.

Thanks for the review.

Issue #3998, opened 2022-11-30
"editing a clip can move other clips" setting bug


DavidBailes:
### Bug description

Once this setting is enabled, disabling it has no effect.

### Steps to reproduce

Start with default configuration:
1. Create an audio track containing two clips.
2. Delete part of the first clip. The second clip does not move as expected.
3. Enable  "editing a clip can move other clips" in track behaviors in preferences.
4. Delete part of the first clip. The second clip moves as expected.
5. Disable "editing a clip can move other clips" in track behaviors in preferences.
6. Delete part of the first clip.

### Expected behavior

The second clip does not move.

### Actual behavior

The second clip moves.

### Audacity Version

latest 3.2.2 beta  -  and 3.2.0 & 3.2.1

### Operating system

Windows

### Additional context

This is a regression wrt 3.1.3

DavidBailes:
Had a quick look at the code. This setting makes use of the class BoolSetting, which is also used by several other settings. I checked one of them - track behaviors > enable scrolling left of zero, and it has the same bug as "editing a clip can move other clips". So a number of other settings may be affected by the same bug.

Paul-Licameli:
See commit 3598fe4 and issue #3399

Paul-Licameli:
See the easy one-line fix at #4005 
Which is marked draft, until we know it's complete, addressing all checkboxes in all Preferendes.

Paul-Licameli:
I think #4005 is now the comprehensive fix.  I reviewed all calls to the Tie... functions taking Settings objects that are not Choice or Enum settings (those don't define Invalidate()).  One of these isn't a BoolSetting but IntSetting.

petersampsonaudacity:
Testing on W10 @Paul-Licameli 's draft branch:  audacity-win-3.2.2-alpha-20221201+e174945-x64

I confirm that this now works properly in Paul's branch for the setting for **Editing a clip can move other clips** in **Tracks Behaviors** preferences, where changes of setting are now properly honored.

I do not have time to exhaustively test _**all**_ of the checkbox settings in _**all**_ preferences.  

One cannot be properly tested until a subsequent release is made and that is **Check for updates** in **Application** preferences.

dozzzzer:
Verified on macOS and Windows 11.

petersampsonaudacity:
Testing on W10 with latest 3.2.2 beta:  audacity-win-3.2.2-beta-20221202+05c36ac-x64-msvc2022

[...]

PR #4005, opened 2022-12-01
Issue3998

Paul-Licameli:
Resolves: #3998

Recommended:

### DIFF
--- a/src/prefs/ApplicationPrefs.cpp
+++ b/src/prefs/ApplicationPrefs.cpp
@@ -107,6 +107,7 @@ bool ApplicationPrefs::Commit()
 {
    ShuttleGui S(this, eIsSavingToPrefs);
    PopulateOrExchange(S);
+   DefaultUpdatesCheckingFlag.Invalidate();
 
    return true;
 }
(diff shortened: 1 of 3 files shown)


petersampsonaudacity:
Testing on W10 @Paul-Licameli 's draft branch:  audacity-win-3.2.2-alpha-20221201+e174945-x64

I confirm that this now works properly for the setting for **Editing a clip can move other clips** in **Tracks Behaviors** preferences, where changes of setting are now properly honored.

I do not have time to exhaustively test _**all**_ of the checkbox settings in _**all**_ preferences.  

One cannot be properly tested until a subsequent release is made and that is **Check for updates** in **Application** preferences.

DavidBailes (approved):
Glad to see you moved the fix from 3.2.3 to 3.2.2, as the bug with the moving clips setting could be very confusing to users.
I can confirm that both settings "editing a clip can move other clips" and "enable scrolling left of zero" now work as expected.
There wasn't a problem with the default sample rate setting before this fix, and there isn't one after the fix.
I haven't had a look at the code in any detail, but the need for these calls to Invalidate() doesn't seem ideal.

Issue #4007, opened 2022-12-01
Xcode build of master branch strips debug symbols from modules and libraries


Paul-Licameli:
### Steps to reproduce

1. Be a developer
2. Build Audacity in Xcode
3. Put a breakpoint in code under /libraries (like ModuleManager::Get()) that you expect to hit at startup
4. Run the program

### Expected behavior

Hit the breakpoint

### Actual behavior

But you don't

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

### Additional context

The problem exists only in the master branch now, not release

PR #4008, opened 2022-12-01
Redo b04d70a so that debug builds in Xcode don't strip symbols...

Paul-Licameli:
... The configure-time variable CMAKE_BUILD_TYPE may be empty while the generate-time variable CONFIG is not.  Before commit b04d70a, that was the variable that governed whether stripping was skipped.

Resolves: #4007

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/AudacityFunctions.cmake
+++ b/cmake-proxies/cmake-modules/AudacityFunctions.cmake
@@ -469,11 +469,18 @@ function( audacity_module_fn NAME SOURCES IMPORT_TARGETS
    target_link_options( ${TARGET} PRIVATE ${LOPTS} )
    target_link_libraries( ${TARGET} PUBLIC ${LIBRARIES} )
 
-   if( NOT CMAKE_SYSTEM_NAME MATCHES "Windows" AND NOT CMAKE_BUILD_TYPE MATCHES "Debug|RelWithDebInfo" )
+   if( NOT CMAKE_SYSTEM_NAME MATCHES "Windows" )
+      # Generate-time boolean values must be "0" or "1",
+      # not "on", "off", etc. like configure-time booleans
+      if (CMAKE_BUILD_TYPE MATCHES "Debug|RelWithDebInfo")
+         set(nostrip 1)
+      else()
+         set(nostrip 0)
+      endif()
       add_custom_command(
          TARGET "${TARGET}"
          POST_BUILD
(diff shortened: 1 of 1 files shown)


cmake-proxies/cmake-modules/AudacityFunctions.cmake
```diff
@@ -469,11 +469,18 @@ function( audacity_module_fn NAME SOURCES IMPORT_TARGETS
...
+      if (CMAKE_BUILD_TYPE MATCHES "Debug|RelWithDebInfo")
+         set(nostrip 1)
+      else()
+         set(nostrip 0)
+      endif()
       add_custom_command(
          TARGET "${TARGET}"
          POST_BUILD
-         COMMAND strip -x $<TARGET_FILE:${TARGET}>
+         COMMAND "$<IF:$<OR:$<CONFIG:Debug>,${nostrip}>,echo,strip>" -x $<TARGET_FILE:${TARGET}>
```
crsib, line 483: This should contain `:$<CONFIG:Debug,RelWithDebInfo>` too for consistency sake.

But I would rather just remove this `if`. If a user wants to strip the resulting binaries - it can be done using "cmake --install --strip`. On CI - we perform strip manually as a part of the packaging process.

Paul-Licameli: I don't know all the build system changes you intend for later, but given what we have, debugging in Xcode is broken (in master only) and this change will unblock my debugging.  Can you approve, and revise it later?

That is, approve after I just add the one more disjunct you have suggested.

Paul-Licameli: The `if` may now be redundant with the complete packaging process -- yes.

But I added this `if` in the first place when I was experimenting with extractions, and also watching what effects it had on build sizes.  For that reason it can still be a development convenience when not doing the complete CI with packaging.

crsib: This change is inconsistent between single and multi config  generators, which is quite wrong. Please at least make it consistent by replacing $<CONFIG:Debug> with $<CONFIG:Debug,RelWithDebInfo>. 

>and also watching what effects it had on build sizes.

Audacity binary is never stripped so I expect incorrect results. In any case - it is preferred to use `cmake --install . --strip --prefix installed` to get the correct sizes.

crsib (approved):

Paul-Licameli:
Thanks.  We will sort out build details later I'm sure.

Issue #4009, opened 2022-12-01
VST2: Effect's settings do not stick when changed during playback


dozzzzer:
**Description:** the effect's settings do not stick when they are changed during playback

https://user-images.githubusercontent.com/70063970/205163483-2a86a1ac-076c-4c64-88b3-32f2c136a2cc.mp4

There are two slightly different ways to reproduce it.

**Scenario 1: with pauses between Play/Stop**

- have some audio
- add Deegain in the sidebar
- open its UI (the fader is at 0.0 dB)
- start playback by pressing Space
- move the fader down to -20.0 while playing the track
- stop the playback by pressing Space
- wait 1 second
- now play the track again for a few seconds
- stop it by pressing Space
- play again by pressing Space

**Actual result:** the slider is moved back to 0.0 dB

**Scenario 2: Play/Stop is done with no pauses**
- have some audio
- add Deegain in the sidebar
- open its UI (the fader is at 0.0 dB)
- start playback by pressing Space
- move the fader down to -20.0 while playing the track
- hit the Space key **twice** to stop and play the track respectively

**Actual result:** the slider is moved back to 0.0 dB
**Expected result:** the slider remains at -20.0 dB

dozzzzer:
UPD: the case described in **Scenario 1** is fixed by @Paul-Licameli 's PR in the build https://github.com/audacity/audacity/actions/runs/3596206050

dozzzzer:
Testing the build https://github.com/audacity/audacity/actions/runs/3601033675
The problem persists.

dozzzzer:
Scenario 2 is moved to #4014

PR #4011, opened 2022-12-02
Issue 4009

pietro68:
Resolves: #4009 

I can attest that this PR fixes Scenario 1 described in #4009, but I can not do the same for Scenario 2, because I could never reproduce it in the first place. Paul, I know you could reproduce Scenario 2 and you also wrote this fix, so please could you tell in the comments if this fixes also Scenario 2?

Recommended:


pietro68 (commented):
I understand the changes.

Paul-Licameli (approved):

Issue #3884, opened 2022-11-01
Toggle for sharing visibility


AnitaBats:

DilsonsPickles:
Figma project file can be found [here](https://www.figma.com/file/L6v5zVmxOMzs4buZM6ladI/10_WX---Share-audio?node-id=30%3A761).

DavidBailes:
> Figma project file can be found [here](https://www.figma.com/file/L6v5zVmxOMzs4buZM6ladI/10_WX---Share-audio?node-id=30%3A761).

Can this file be made viewable by anyone?

dozzzzer:
The share audio dialog doesn't remember the previously selected option, i.e. every time a user tries to share audio, they end up seeing the "Public" option selected by default, not the option they selected when uploading last time.

<img width="349" alt="image" src="https://user-images.githubusercontent.com/70063970/206421447-64f682ad-e763-4bad-ad1f-0cc0fa1dc71d.png">

DilsonsPickles:
It should yes, this was included in the spec document as seen below: 

![image](https://user-images.githubusercontent.com/111350316/206427842-ee4d94fe-546d-41a9-b3be-fb348666e9de.png)

crsib:
Sorry, missed that

PR #4013, opened 2022-12-02
Allow setting the visibility of the shared audio

crsib:
Resolves: #3884

Recommended:

### DIFF
--- a/libraries/lib-cloud-audiocom/UploadService.cpp
+++ b/libraries/lib-cloud-audiocom/UploadService.cpp
@@ -50,7 +50,8 @@ std::string_view DeduceMimeType(const wxString& ext)
       return "audio/x-wav";
 }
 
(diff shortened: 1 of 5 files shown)


vsverchinsky (approved):

DavidBailes:
The text of the radio buttons is not accessible for users of screen readers.
1. You are calling Name() with just part of the text. You need to call it with all of the text so that screen readers will read all of it, eg Name(XO("Public. Anyone will be able to listen to this audio."))
2. For screen readers to read any of this text, you'll need to set an accessible object which will pick it up e.g:
#if wxUSE_ACCESSIBILITY
   // so that name can be set on a standard control
   isPublic->SetAccessible(safenew WindowAccessible(isPublic));
#endif

crsib:
Thank you for your feedback, it is now addressed

vsverchinsky (approved):

PR #4030, opened 2022-12-06
Vst param id bounds check

Paul-Licameli:
Resolves: In search of an issue number

Defend against VST plug-ins that Automate an parameter ID that is out of the bounds
reported by the plug-in

Recommended:

### DIFF
--- a/src/effects/VST/VSTEffect.cpp
+++ b/src/effects/VST/VSTEffect.cpp
@@ -935,9 +935,10 @@ namespace
       // Make a message from a single parameter
       explicit VSTEffectMessage(int id, double value, size_t numParams)
       {
-         mParamsVec.resize(numParams, std::nullopt);
-         if (id < numParams)
-            mParamsVec[id] = value;
+         if (id >= numParams)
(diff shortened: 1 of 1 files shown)


crsib (approved):

src/effects/VST/VSTEffect.cpp
```diff
@@ -935,9 +935,10 @@ namespace
       // Make a message from a single parameter
       explicit VSTEffectMessage(int id, double value, size_t numParams)
       {
-         mParamsVec.resize(numParams, std::nullopt);
-         if (id < numParams)
-            mParamsVec[id] = value;
+         if (id >= numParams)
+            wxLogDebug(wxT("VST parameter id out of nominal bounds!"));
+         mParamsVec.resize(std::max<size_t>(id + 1, numParams), std::nullopt);
```
crsib, line 940: Looks like it potentially can allocate an arbitrary amount of memory

Paul-Licameli: Yes, why I would rather either consider it a bug in the plugin, or else represent the array of parameters sparsely.  The second option is more work.

PR #4032, opened 2022-12-06
Menu callback simplification

Paul-Licameli:

Eliminate verbosity in the definition of most menu items, which can call a stateless
nonmember function.  The prior system was general enough to use member functions
of a stateful static object, but that requires more boilerplate that only a few callbacks need.

Let's do this simplification before other restructurings that will cut and paste many menu
items, so that more of them are scattered and use registries to put them in their places.

Recommended:


Paul-Licameli:
> Quite welcome cleanup

All reviewed, so soon?  Mostly repetitious stuff after the interesting first commit.

I don't often use union.

I hope @ksoze95 looks this over too.  Expect much more cleanup related to menus.

crsib:
Please define "soon"

Issue #4004, opened 2022-12-01
Add prompt dialog when copying clips between projects


vsverchinsky:
- Add dialog asking user to choose option to either discard trimmed data or copy all contents.
 - Dialog should allow "remember my choice"
 - There should be an option in preferences that can change remembered user selection
 - Add copy progress dialog

Addresses use case issues reported in #3820

petersampsonaudacity:
Tested this on W10 with @vsverchinsky vVitaly's latest branch build:  audacity-win-3.3.0-alpha-20221206+9c91e83-x64-msvc2022
This works fine when you copy a clip from one project to another
![image](https://user-images.githubusercontent.com/50205984/206185599-6cb5beb9-e3d5-4762-a70e-4d772c11f0f9.png)

a) but can we **_please_** also have this dialog and option when copying clips within a project thereby helping to avoid the over-bloated projects that may users have complained about.

b) It is unclear to me whether or not the user's choice for "**Selected only**" or "**Whole**" is sticky and retained/remembered if they have ticked the "**Don't show me this again**" check-box.

c) if the user has checked the "**Don't show me this again**" check-box - then there appears to be no way for the user to set this dialog to be displayed again (except with a **Tools > reset Configuration** which may be way more drastic than the user requires).
_There appears to be no preference setting or command to control the display/hiding pf this dialog_

Flagging both these as UX/design issues to @LWinterberg @Tantacrul @DilsonsPickles

petersampsonaudacity:
UPDATE

The message looks good - but the copy fails to work

If I copy a short selection to another project with this build choosing "Selected only"

a) the length of the selection is OK for the copy

b) no audio is copied - just a selection 

![image](https://user-images.githubusercontent.com/50205984/206187822-1da8ac2c-50c7-4f2f-9a8e-5c7f94148b62.png)

The Copy works fine if I choose the "**Whole**" option

petersampsonaudacity:
Testing on W10 with @vsverchinsky Vital'ys latest branch:  audacity-win-3.2.2-alpha-20221207+12b213c-x64

a) I think the message is improved with the changes

b) with the **Paste selected audio only** selected  the audio **_now does get pasted_** and it is properly reduced to just the clip

c) BUT if I then check on **Remember my choice** it fails to do and and reverts to **Paste smart clip** - **_but_** now with no dialog displayed to show the reversion -
[...]

Issue #3820, opened 2022-10-17
Performance and file size issues when copying and pasting a small portion of a very large clip


JohnColket:
### Bug description (edited by @Tantacrul)

Audacity 3.2.1

1) Generate a 1 hour chirp
2) Select the very small portion of the audio and copy it
3) Paste it into a new project
4) Notice the entire original clip has been pasted, which massively increases the file size of the new project

See comments below from @Tantacrul for a proposed solution.

---

**Secondly**, there have been reports of large performance problems as a result of copying and pasting data working with large clips. These need to be investigated @Ryan-Miller-Audacity @dozzzzer.

mapisilafi:
Struggling with this too. It appeared after installing 3.2.1. Usually recording hours and then taking small snippets from here and there using copy-paste. It takes minutes to paste to a new file. During operation Disk usage at 100%, 400MB/s. Previously that operation took only few seconds.

petersampsonaudacity:
A close cousin of this is that the **Edit>Remove Special**  commands
a) Split Cut
b) Split Delete
c) Trim Audio
do _**NO LONGER**_  delete the removed audio.

Rather the remaining audio clips are turned into "smart clips" with the ostensibly removed audio still present as part of the clip(s) but hidden.  This new behavior is a regression on earlier Audacity, introduced in 3.1.0 - and will be a surprise to most users (if they notice it at all).

What it does do is to unnecessarily increase the size of a users project as several users have reported - and this is how we found this odd behavior.

Normally when one uses a delete, cut or trim command one expects the removed data to be actually removed (and in the case of cut placed on the clipboard).  Certainly that is the case when one uses Word or Excel, or other apps that I us.

It seems to me that if you do want to have such functionality it should be new commands which make it explicitly clear that data is not being removed but merely hidden.

AFAICT this hidden creation of "smart clips" is not documented or announced anywhere.

### Steps To Reproduce
1. generate a 30 second chirp
2. select 20-20 seconds
3. Edit > Remove Special > Spit Delete (or Trim Audio or Split Cut)
4. Drag at the top edge of the remaining clips
5. Observe: the supposedly deleted audio is revealed as the smart clip expands

Note too that with Spit Delete and Split Cut _**BOTH**_ of the remaining "smart clips" have the not just the full data that has allegedly been removed they also have the full 30 second clip that can be dragged out and more - so you then have not one but two copies of the full clip.

Flagging @Tantacrul and @LWinterberg as this is a UX/Design issue

[...]

PR #4033, opened 2022-12-07
Clip copy dialog

vsverchinsky:
Resolves: #4004
Resolves: #3820 

Recommended:

### DIFF
--- a/src/AudioPasteDialog.cpp
+++ b/src/AudioPasteDialog.cpp
@@ -1,0 +1,147 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file AudioPasteDialog.cpp
+
+  @author Vitaly Sverchinsky
+
+**********************************************************************/
+
+#include "AudioPasteDialog.h"
+
+#include <wx/sizer.h>
+#include <wx/checkbox.h>
+#include <wx/stattext.h>
+#include <wx/radiobut.h>
+
+#include "ShuttleGui.h"
+#include "prefs/TracksBehaviorsPrefs.h"
+#include "widgets/WindowAccessible.h"
+
+BEGIN_EVENT_TABLE(AudioPasteDialog, wxDialogWrapper)
+   EVT_BUTTON(wxID_OK, AudioPasteDialog::OnContinue)
+   EVT_BUTTON(wxID_CANCEL, AudioPasteDialog::OnCancel)
+END_EVENT_TABLE()
+
+namespace {
+
+const TranslatableString audioPasteModeText[] = {
+   XO("Smart clip.\n"
+      "The entire source clip will be pasted into your project, allowing you to access\ntrimmed audio data anytime."),
+   XO("Selected audio only.\n"
+      "Only the selected portion of the source clip will be pasted.")
+};
+
+//Setting is different from TracksBehaviorsAudioTrackPastePolicy
+//This is to remember the last mode used in the dialog
+ChoiceSetting AudioPasteDialogDefaultModeSetting = {
+   wxT("/GUI/AudioPasteDialogDefaultMode"),
+   {
+      ByColumns,
+      {{}, {}},
+      { wxT("Keep"), wxT("Discard") }
+   },
+   0, // "Keep"
+};
+
+};
+
+AudioPasteDialog::AudioPasteDialog(wxWindow* parent, wxULongLong nCopiedBytes, int winid)
+   : wxDialogWrapper(parent, winid, XO("Paste audio")), mCopiedBytesNum(nCopiedBytes)
+{
+   ShuttleGui S(this, eIsCreatingFromPrefs);
+   PopulateOrExchange(S);
+   
+   SetInitialSize({520, -1});
+   
+   Layout();
+   Fit();
+   Center();
+}
+
+void AudioPasteDialog::PopulateOrExchange(ShuttleGui &S)
+{
+   S.SetBorder(20);
+   S.AddFixedText(XO("How would you like to paste your audio?"));
+   S.AddSpace(10);
+   S.StartRadioButtonGroup(AudioPasteDialogDefaultModeSetting);
+   {
+      for(int i = 0, count =AudioPasteDialogDefaultModeSetting.GetSymbols().size();
+         i < count;
+         ++i)
+      {
+         S.StartHorizontalLay(wxALIGN_LEFT);
+         {
+            S.SetBorder(0);
+            S.AddSpace(20);
+            wxRadioButton* radioButton{};
+            S.StartVerticalLay(0);
+            {
+               radioButton =
+                  S.Name(audioPasteModeText[i]).TieRadioButton();
+               S.AddSpace(0, 0, 1);
+            }
+            S.EndVerticalLay();
+
+            S.AddVariableText(audioPasteModeText[i])->Bind(
+               wxEVT_LEFT_UP, [=](auto)
+               {
+                  radioButton->SetValue(true);
+               });
+   #if wxUSE_ACCESSIBILITY
+            safenew WindowAccessible(radioButton);
+   #endif
+            S.AddSpace(20, 0);
+         }
+         S.EndHorizontalLay();
+         if(i != count - 1)
+            S.AddSpace(0, 10);
+      }
+   }
+   S.EndRadioButtonGroup();
+   
+   S.SetBorder(20);
+   S.AddFixedText(
+      XO("Audio data is %s. Larger sizes will take longer to paste.")
+         .Format(wxFileName::GetHumanReadableSize(mCopiedBytesNum))
+   );
+   
+   
+   S.SetBorder(0);
+   S.StartHorizontalLay(wxEXPAND);
+   {
+      S.AddSpace(20, 0);
+      mDontShowAgain = S
+         .AddCheckBox(XO("Remember my choice and don't ask again"), false);
+      if(mDontShowAgain != nullptr)
+         mDontShowAgain->SetFont(wxFont(wxFontInfo().Bold()));
+      S.AddSpace(0, 0, 1);
+      S.Id(wxID_CANCEL).AddButton(XXO("Cancel"), wxALIGN_CENTER);
+      S.AddSpace(6, 0);
+      S.Id(wxID_OK).AddButton(XXO("Continue"), wxALIGN_CENTER, true);
+      S.AddSpace(20, 0);
+   }
+   S.EndHorizontalLay();
+   
+   S.AddSpace(0, 10);
+}
+
+void AudioPasteDialog::OnContinue(wxCommandEvent&)
+{
+   ShuttleGui S(this, eIsSavingToPrefs);
+   PopulateOrExchange(S);
+   
+   auto pasteMode = AudioPasteDialogDefaultModeSetting.Read();
+   if(mDontShowAgain->IsChecked())
+      TracksBehaviorsAudioTrackPastePolicy.Write(pasteMode);
+   
+   gPrefs->Flush();
+   
+   EndModal(pasteMode == wxT("Discard") ? DISCARD : KEEP);
+}
+
+void AudioPasteDialog::OnCancel(wxCommandEvent&)
+{
+   EndModal(wxID_CANCEL);
+}
--- a/src/AudioPasteDialog.h
+++ b/src/AudioPasteDialog.h
@@ -1,0 +1,38 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file AudioPasteDialog.h
+
+  @author Vitaly Sverchinsky
+
+**********************************************************************/
+
+#pragma once
+
+#include "widgets/wxPanelWrapper.h"
+
+class wxCheckBox;
+class ShuttleGui;
+
+class AudioPasteDialog final : public wxDialogWrapper
+{
+   wxCheckBox* mDontShowAgain { nullptr };
+   wxULongLong mCopiedBytesNum {};
+public:
+   
+   DECLARE_EVENT_TABLE()
+   
+public:
+   enum DialogResult {
+      KEEP = wxID_HIGHEST + 1,
+      DISCARD
+   };
+   
+   AudioPasteDialog(wxWindow* parent, wxULongLong nCopiedBytes, int winid = wxID_ANY);
+   
+private:
+   void PopulateOrExchange(ShuttleGui& S);
+   void OnContinue(wxCommandEvent&);
+   void OnCancel(wxCommandEvent&);
+};
--- a/src/menus/EditMenus.cpp
+++ b/src/menus/EditMenus.cpp
@@ -71,62 +71,52 @@ bool DoPasteText(AudacityProject &project)
    return false;
 }
 
-// Return true if nothing selected, regardless of paste result.
-// If nothing was selected, create and paste into NEW tracks.
+// Create and paste into NEW tracks.
+// Simplified version of DoPaste, used when there is no selection
+// on tracks
 // (This was formerly the second part of overly-long OnPaste.)
-bool DoPasteNothingSelected(AudacityProject &project)
+void DoPasteNothingSelected(AudacityProject &project, const TrackList& src, double t0, double t1)
 {
    auto &tracks = TrackList::Get( project );
    auto &selectedRegion = ViewInfo::Get( project ).selectedRegion;
    auto &viewInfo = ViewInfo::Get( project );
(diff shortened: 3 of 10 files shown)


petersampsonaudacity:
Tested this on W10 with @vsverchinsky vVitaly's latest branch build:  audacity-win-3.3.0-alpha-20221206+9c91e83-x64-msvc2022
This works fine when you copy a clip from one project to another
image

a) but can we **_please_** also have this dialog and option when copying clips within a project thereby helping to avoid the over-bloated projects that may users have complained about.

c) It is unclear to me whether or not the user's choice for "**Selected only**" or "**Whole**" is sticky and retained/remembered if they have ticked the "**Don't show me this again**" check-box.

Flagging both these as UX/design issues to @LWinterberg @Tantacrul @DilsonsPickles

petersampsonaudacity:
UPDATE

The message looks good - but the copy fails to work

If I copy a short selection to another project with this build choosing "Selected only"

a) the length of the selection is OK for the copy

b) no audio is copied - just a selection 

image

DavidBailes:
There are some issues for users of screen readers:
1. Concerning the text for the radio buttons using both in preferences and the dialog. For each of the radio buttons, there needs to be a full stop at the end of the first line, otherwise screen readers read it without a pause between the end of the first line and the start of the second. eg it should be:
Selected audio only.
Only the selected portion of the source clip will be pasted.

2. In preferences, the text "Paste audio from other Audacity project as" is not automatically read by screen readers. My suggestion is to have this text as the text used by StartStatic (ie the group box), replacing Pasted audio. It will then be read by screen readers.

3. The radio buttons and text box need access keys, eg, m, S and R.

src/AudioPasteDialog.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include "widgets/wxPanelWrapper.h"
+
+class wxCheckBox;
+class ShuttleGui;
+
+class AudioPasteDialog final : public wxDialogWrapper
+{
+   DECLARE_DYNAMIC_CLASS(WaveClipCopyWarningDialog)
```
Paul-Licameli, line 20: This macro isn't needed and the named class doesn't exist

src/AudioPasteDialog.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+IMPLEMENT_CLASS(AudioPasteDialog, wxDialogWrapper)
+
+namespace {
+
+const TranslatableString audioPasteModeText[] = {
+   XO("Smart clip\n"
+          "The entire source clip will be pasted into your project, allowing you to access it at anytime."),
+   XO("Selected audio only\n"
+          "Only the selected portion of the source clip will be pasted.")
```
Paul-Licameli, line 35: These string literals are repeated in TracksBehaviorsPrefs.cpp.  Eliminating duplication might be nice but in this case is not very important.  No bug would result if they were differently modified.  The strings are not used as identifiers.

src/menus/EditMenus.cpp
```diff
@@ -73,65 +79,173 @@ bool DoPasteText(AudacityProject &project)
...
 // (This was formerly the second part of overly-long OnPaste.)
-bool DoPasteNothingSelected(AudacityProject &project)
+void DoPasteNothingSelected(AudacityProject &project, const TrackList& src, double t0, double t1)
 {
    auto &tracks = TrackList::Get( project );
    auto &selectedRegion = ViewInfo::Get( project ).selectedRegion;
    auto &viewInfo = ViewInfo::Get( project );
    auto &window = ProjectWindow::Get( project );
+   
+   assert(!tracks.Selected());
```
Paul-Licameli, line 198: What is now asserted instead of checked, should be commented above the function too as the precondition it assumes.

src/menus/EditMenus.cpp
```diff
@@ -73,65 +79,173 @@ bool DoPasteText(AudacityProject &project)
...
+      }
+      result->Add(trackCopy);
+   }
+   return result;
+}
+
+// Create and paste into NEW tracks.
 // (This was formerly the second part of overly-long OnPaste.)
-bool DoPasteNothingSelected(AudacityProject &project)
+void DoPasteNothingSelected(AudacityProject &project, const TrackList& src, double t0, double t1)
```
Paul-Licameli, line 191: The third commit has no errors, but I don't yet see what code duplication is removed as the commit comment says.  Or rather, this is allowing something to be reused?

src/menus/EditMenus.cpp
```diff
@@ -73,65 +79,173 @@ bool DoPasteText(AudacityProject &project)
...
+            --dstChannelCount;
+         }
+         
+         while(dstChannelCount > 0)
+         {
+            f(**lastCopiedTrack);
+            ++dstTrack;
+            --dstChannelCount;
+         }
+      }
```
Paul-Licameli, line 126: Here, assume negation of the while-condition:
`dstTrack == dstTrackRange.end() || srcTrack == srcTrackRange.end()`

src/menus/EditMenus.cpp
```diff
@@ -73,65 +79,173 @@ bool DoPasteText(AudacityProject &project)
...
+      while(dstTrack != dstTrackRange.end())
```
Paul-Licameli, line 127: Therefore, `srcTrack == srcTrackRange.end()`

src/menus/EditMenus.cpp
```diff
@@ -73,65 +79,173 @@ bool DoPasteText(AudacityProject &project)
...
+      {
+         if((*dstTrack)->GetSelected())
+            f(**lastCopiedTrack);
+         ++dstTrack;
+      }
```
Paul-Licameli, line 132: But what about the other possibility, that destination tracks ran out before source tracks?  Don't you need to check that and visit extra source tracks?  You do that at the top of the function, only when there are zero destination tracks.

src/menus/EditMenus.cpp
```diff
@@ -73,65 +79,173 @@ bool DoPasteText(AudacityProject &project)
...
+            ++dstTrack;
+            continue;
+         }
+         
+         auto srcChannelCount = TrackList::Channels(*srcTrack).size();
+         auto dstChannelCount = TrackList::Channels(*dstTrack).size();
+                  
+         while(srcChannelCount > 0 && dstChannelCount > 0)
+         {
+            f(**srcTrack);
```
Paul-Licameli, line 111: Here and elsewhere that you call `f`, what proves that there is not a null dereference?  Track iterators are allowed to dereference an end iterator, getting a null, but then the second dereference might be unsafe.

The function would look better with a test for empty src and early exit.

src/menus/EditMenus.cpp
```diff
@@ -73,65 +79,173 @@ bool DoPasteText(AudacityProject &project)
...
+            f(**lastCopiedTrack);
+         ++dstTrack;
+      }
+   }
+}
+
+wxULongLong EstimateCopyBytesCount(const TrackList& src, const TrackList& dst)
+{
+   wxULongLong result{};
+   ForEachCopiedWaveTrack(src, dst, [&](const WaveTrack& waveTrack) {
```
Paul-Licameli, line 139: The algorithm in `ForEachCopiedWaveTrack` might visit one source track multiple times and count its space multiply in the estimate.

This isn't needed for copying and pasting within a project because of the sharing of data.

Rather than all this work, it suggests to me that cross-project copying and pasting needs a smarter algorithm to replicate the graph structure and pattern of sharing.

This could prevent explosion of file sizes in the first place without so much need to warn about it.

Paul-Licameli (changes requested):
The only must-fix matters I see are the corrections in `ForEachCopiedWaveTrack`

src/AudioPasteDialog.h
vsverchinsky: Thanks!

src/AudioPasteDialog.cpp
vsverchinsky: Agreed, though not very clear where exactly such strings could be placed

src/menus/EditMenus.cpp
vsverchinsky: Clipboard access and double checking on same condition (`!tracks.Selected()`) - `DoPasteNotingSelected` assumes that nothing is selected... but I agree that commit message isn't very clear, I'll change it to `Audio paste code simplification`

src/menus/EditMenus.cpp
vsverchinsky: When copying tracks we consider two cases when pasting:
1) There is no selection
2) Not empty region is selected
In the first case we copy **all tracks** from `src` (used in simplified paste method `DoPasteNothingSelected`). When selection isn't empty `N = min(src.size(), dst.size())` tracks are copied from `src`, plus the **last track** from `src` could be duplicated `M = dst.size() - N` times more, if `M > 0` (corresponds to a paste logic after `!tracks.Selected()` condition in `OnPaste`). In both cases `ForEachCopiedWaveTrack` visits tracks that are to be copied according to behaviour described above.
Does that answers your questions?

src/menus/EditMenus.cpp
vsverchinsky: Not sure that understand your question. `dst` is allowed to be empty, `src` is expected to be not empty (and it can't have null tracks, right?). I don't see where null pointer dereference could happen in that case. I can comment about preconditions, but it's not a publicly available function which's purpose is to simplify implementation of another function.

src/menus/EditMenus.cpp
vsverchinsky: >The algorithm in ForEachCopiedWaveTrack might visit one source track multiple times and count its space multiply in the estimate.

Also it's possible that different source tracks can reference same blocks

src/menus/EditMenus.cpp
Paul-Licameli: Yes, I accept this reasoning.  But then the new function should have a comment containing this explanation.  Reading the function in isolation made me question it.

src/menus/EditMenus.cpp
Paul-Licameli: Right, to prove no null dereference happens, you must assume src is not an empty track list, and that is guaranteed at the call site.

It's all happening within one file, yes, so I won't be pedantic about stating preconditions.

Paul-Licameli (approved):

petersampsonaudacity:
Testing on W1- with @vsverchinsky Vitaly's latest branch build for this:   

a)  the copy properly honor's the user's request to either copy the whole smart clip ot just the selection

b) the displayed data size does _**not**_ change when toggling between the two copy forms - I would expect the smart clip data size to be much larger than the simple selection size.

c) At default settings the dialog for Tracks Behaviors preferences is now not quite long enough for the new text:
image

vsverchinsky:
1 commit added that removes 'X' button from progress dialog when hide close and cancel button flags are used.
Accessibility improvements.

petersampsonaudacity:
Testing on @vsverchinsky 's latest branch build:  audacity-win-3.3.0-alpha-20230210+429a3c0-x64

I confirm that the Copy dialog no longer has the "X" to cancel the process

BUT consider this use case
1. Generate a five hour chirp to simulate a long recording project
2. select a three minute "song"
3. Ctrl+N to open a new project
4. Ctrl+V to paste the entire smart clip in (the default setting in the dialog)
5. Observe: this takes 2min 24secs on my machine with _**NO**_ longer any ability to interrupt this 

If you make a mistake and intended only to copy in just the selection (which takes less than a second) 0t's like watching paint dry as you wait !

Removing the "X" cancel as a "fix" seems a very kludgy approach and poor UX.

vsverchinsky:
> Testing on @vsverchinsky 's latest branch build: audacity-win-3.3.0-alpha-20230210+429a3c0-x64
> 
> I confirm that the Copy dialog no longer has the "X" to cancel the process
> 
> BUT consider this use case
> 
> 1. Generate a five hour chirp to simulate a long recording project
> 2. select a three minute "song"
> 3. Ctrl+N to open a new project
> 4. Ctrl+V to paste the entire smart clip in (the default setting in the dialog)
> 5. Observe: this takes 2min 24secs on my machine with _**NO**_ longer any ability to interrupt this
> 
> If you make a mistake and intended only to copy in just the selection (which takes less than a second) 0t's like watching paint dry as you wait !
> 
> Removing the "X" cancel as a "fix" seems a very kludgy approach and poor UX.

And it's been so before dialog was added. The major problem is that I can't suggest a solution that doesn't require deep changes in the code and those changes should be discussed with @crsib and @Paul-Licameli first, I guess

petersampsonaudacity:
> And it's been so before dialog was added. The major problem is that I can't suggest a solution that doesn't require deep changes in the code and those changes should be discussed with @crsib and @Paul-Licameli first, I guess

@vsverchinsky 

For me the lack of a **Cancel** button or **X** is a relatively minor residual.

I will just be **_very_** happy that users who want to copy a small selection from a large clip in one project to a different project will now be able to copy just the selected audio data and _**not**_ the whole shebang of the "smart clip".

So I'm _**really**_ hoping that this fix gets pulled into master for 3.3.0

Paul-Licameli:
> > Testing on @vsverchinsky 's latest branch build: audacity-win-3.3.0-alpha-20230210+429a3c0-x64
> > I confirm that the Copy dialog no longer has the "X" to cancel the process
> > BUT consider this use case
> > 
> > 1. Generate a five hour chirp to simulate a long recording project
> > 2. select a three minute "song"
> > 3. Ctrl+N to open a new project
> > 4. Ctrl+V to paste the entire smart clip in (the default setting in the dialog)
> > 5. Observe: this takes 2min 24secs on my machine with _**NO**_ longer any ability to interrupt this
> > 
> > If you make a mistake and intended only to copy in just the selection (which takes less than a second) 0t's like watching paint dry as you wait !
> > Removing the "X" cancel as a "fix" seems a very kludgy approach and poor UX.
> 
> And it's been so before dialog was added. The major problem is that I can't suggest a solution that doesn't require deep changes in the code and those changes should be discussed with @crsib and @Paul-Licameli first, I guess

See SqliteSampleBlockFactory::OnBeginPurge

Not directly applicable here, but `SampleBlock` has already exposed ways to hook some of its operations.

Perhaps we could temporarily inject a similar callback for creation of blocks, which could poll the cancel button of a progress dialog, then throw an exception.

src/AudioPasteDialog.cpp
```diff
@@ -0,0 +1,147 @@
...
+BEGIN_EVENT_TABLE(AudioPasteDialog, wxDialogWrapper)
+   EVT_BUTTON(wxID_OK, AudioPasteDialog::OnContinue)
+   EVT_BUTTON(wxID_CANCEL, AudioPasteDialog::OnCancel)
+END_EVENT_TABLE()
+
+namespace {
+
+const TranslatableString audioPasteModeText[] = {
+   XO("Smart clip.\n"
+      "The entire source clip will be pasted into your project, allowing you to access\ntrimmed audio data anytime."),
```
Paul-Licameli, line 31: I assume all visible design and wording changes were prescribed by someone else, but I think "any time" is more proper English usage.  Although Grammarly just told me it's considered acceptable as one word now.  @LWinterberg

petersampsonaudacity: Grammarly says:
>A century ago, it was standard to write any time as two words in all contexts. But it’s now perfectly acceptable to write anytime as one word when you’re using it as an [adverb](https://www.grammarly.com/blog/adverb/). However, some readers still consider it a casualism, so you may want to stick to the two-word version for extremely formal writing.

I'm minded to agree with @Paul-Licameli  the it would be better as "any time" with the space.

Paul-Licameli (approved):
I reviewed again, even without the rebasing onto master.

The conflicts to be resolved are all easy ones involving `#include` directives or visibility macros added to classes by the extraction of lib-wave-track.

The question about wording can easily be decided and fixed later.

petersampsonaudacity:
Testing with @vsverchinsky 's latest brabch build:  audacity-win-3.3.0-alpha-20230220+b954cc7-x64

I note that we still have the "anytime" in the clip copy dialog and in Tracks Behaviors prefs - and not "any time" or "at any time" as we (@Paul-Licameli ) discussed at the end of last week.


Also note that the Tracks Behaviors prefs dialog is till not long enough vertically:
image

JohnColket:
To all who worked on this issue - THANK YOU!  :smiley:

petersampsonaudacity:
I have updated the Tracks Behaviors preferences page with the new behavior
https://alphamanual.audacityteam.org/man/Tracks_Behaviors_Preferences#Pasted_audio

image

PR #4052, opened 2022-12-12
Rewrite audio setup menus

Paul-Licameli:

QA:  sanity checking that behavior of the Audio Setup toolbar has not changed

Use of wx/menu.h should be limited to just one place in the whole source tree, behind a
facade interface.

This rewrite of AudioSetupToolBar.cpp does not yet remove wx/menu.h but is a step toward
that.

The toolbar was using the checkmark sates of items of hidden wxMenu objects to track
the state of choices the user has made.  But that can be done instead with vectors of
strings and indices into them.

Recommended:

### DIFF
--- a/src/toolbars/AudioSetupToolBar.cpp
+++ b/src/toolbars/AudioSetupToolBar.cpp
@@ -37,12 +37,6 @@
 #include "../widgets/wxWidgetsWindowPlacement.h"
 
 namespace {
-   static constexpr int kHost = 15000;
-   static constexpr int kInput = 15200;
-   static constexpr int kInputChannels = 15400;
-   static constexpr int kOutput = 15600;
-   static constexpr int kAudioSettings = 15800;
-
    class ViewDeviceSettingsDialog final : public PrefsDialog
    {
    public:
@@ -116,7 +110,7 @@ void AudioSetupToolBar::DeinitChildren()
    mInput.reset();
    mOutput.reset();
    mInputChannels.reset();
-   mHost.reset();
+   mHost.Clear();
 }
 
 void AudioSetupToolBar::Populate()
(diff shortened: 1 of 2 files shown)


ksoze95 (dismissed):
There seems to be a regression with this refactoring.  Under windows, when I chose the audio device host as "Windows DirectSound" the Playback device menu item is disabled, however the "Audio Settings..." shows that the playback device is "Primary Sound Driver".

This regression seems to happen with commit 42ca39b5a

Could 4e3a7ab9e7f7040f0e2256f16e51ee89881fab4b be fixup/squash with an earlier commit so that the later commits also build under Windows.

Paul-Licameli:
> There seems to be a regression with this refactoring. Under windows, when I chose the audio device host as "Windows DirectSound" the Playback device menu item is disabled, however the "Audio Settings..." shows that the playback device is "Primary Sound Driver".
> 
> This regression seems to happen with commit [42ca39b](https://github.com/audacity/audacity/commit/42ca39b5abd50a8309a0b849dc5a8ae4b011ed4b)
> 
> Could [4e3a7ab](https://github.com/audacity/audacity/commit/4e3a7ab9e7f7040f0e2256f16e51ee89881fab4b) be fixup/squash with an earlier commit so that the later commits also build under Windows.

That is "Don't use hidden menu to track chosen output device"

Paul-Licameli:
@ksoze95 I squashed the Windows build fix as you asked.  Now the bad commit is at a different hash.

ksoze95:
Only tested on Windows, the latest commit seems to have made the "Playback Device" menu item behaviour worst.  Starting without a audacity.cfg file, the "Playback Device" menu option is always disabled, no matter what I choose as the host.

ksoze95 (approved):

Paul-Licameli:
I squashed the small bug fix commit into its proper place in the commit sequence.  There is no new net change in the source tree after the force push.

Paul-Licameli:
Rebased again onto more recent master but unchanged.

PR #4053, opened 2022-12-12
Toolkit neutral menus part 2

Paul-Licameli:

Further changes after #4017 define a sufficient toolkit neutral interface for most uses of
toolbar and context menus.

That facade is not yet written with abstract base classes and injection of the toolkit-specific
implementation at startup.

The intent is to eliminate all but one use of wx/menu.h.  This is almost done here.

But see #4052.  This pull request does not yet remove wx/menu.h from AudioSetupToolbar,
which has many other (but unnecessary) uses of menus that were added after these
commits were drafted two years ago.  That pull request (based directly on current
master, not on #4017) does other rewriting first.

There are also still uses of wxEVT_MENU_OPEN, wxEVT_MENU_CLOSE, and wxEVT_MENU
to remove (these are defined not in wx/menu.h but in wx/event.h)

Recommended:


crsib (changes requested):
After spending quite a lot of time going multiple times through this PR I am still very confused with naming used and with semantics behind those names. I think this is something that should be improved.

In the commit that introduces Handle it feels like some work is missing. `wxMenu` is used directly instead of Handle methods.

In general, I think that this is the case where we should separate the **definition** of menu and the toolkit specific code in two different layers instead of proposed `Handle` approach.

Consider pseudocode like

```
BasicMenu::Menu menu;

menu.Append(XO("No Effect"). []{})'
menu.AppendSeparator();
...

// wxMenu is only created here...
BasicMenu::Popup(menu);
```

I think it will simplify the lifetime management quite a bit.

Paul-Licameli:
Wait!  You are reviewing the part 2 PR before part 1!

Paul-Licameli:
I order the PRs in the column of the project board to show which I want reviewed sooner.

crsib:
I am reviewing both because Part 1 makes very little sense to me by itself. I wanted to see if further work will clarify my previous confusions - it did not.

Issue #4051, opened 2022-12-11
Make echo effect realtime


pietro68:
Echo built-in effect was transformed into stateless, but it is still missing realtime capabilities.

PR #4054, opened 2022-12-12
Make echo realtime

pietro68:
Resolves: #4051 

I had to change the type of member `history` from `Floats` to `std::vector<float>`, because of the call to `mSlaves.push_back(slave)` - that did not go well with `history` being a `Floats`.

Recommended:

### DIFF
--- a/src/effects/Echo.cpp
+++ b/src/effects/Echo.cpp
@@ -51,6 +51,9 @@ struct EffectEcho::Instance
       : PerTrackEffect::Instance{ effect }
    {}
 
+   bool InstanceInit(EffectSettings& settings, double sampleRate,
+      ChannelNames chanMap);  
+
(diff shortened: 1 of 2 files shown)


src/effects/Echo.cpp
```diff
@@ -134,14 +164,14 @@ bool EffectEcho::Instance::ProcessInitialize(
...
-      history.reinit(histLen, true);
+      history.resize(histLen, 0.0f);
    }
    catch ( const std::bad_alloc& ) {
       mProcessor.MessageBox( XO("Requested value exceeds memory capacity.") );
       return false;
    }
 
-   return history != NULL;
+   return history.size() > 0;
```
ksoze95, line 174: I think this should be replaced with ```return history.empty();``` instead.

Issue #3704, opened 2022-09-27
Can't rename clips


robindouglasjohnson:
### Bug description

Seems to be intermittent, but I'm unable to rename audio clips. Double-clicking the title of the click in the track just selects the clip. Right-clicking and selecting "Rename clip", or pressing ctrl+F2, both do nothing.

### Steps to reproduce

1. Get some audio
2.  split it into two clips
3. Separate the clips such that only the first clip is on screen
4. Try to rename the clip, either by double-clicking its title, or selecting it then either right-clicking and choosing "Rename clip" or pressing Ctrl+F2
5. Observe:  Nothing happens

 _@vsverchinsky Vitaly  I, Peter Sampson, edited the STRs following a reliable simple reproducible set pointed out later in this bug-thread by @DavidBailes_

### Expected behavior

You should be able to rename clips

### Actual behavior

I cannot rename clips

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Windows 11. Behaviour is intermittent but I can't see a pattern. Possibly once a project gets into this state (or beyond a certain size/number of clips?) it becomes permanently impossible to rename clips. I'm actually able to rename the clips in a new project I just created, but not in my existing one, both open at the same time.

**Workaround**
@DavidBailes points out in a comment below in this bug-thread - there is a workaround to be had by setting the **Tracks Behaviors** pref for  "**Use dialog for the name of a new label**" to be **on**.

Somewhat surprisingly this totally undocumented Easter-Egg trick works as for some unknown reason this setting also affects the behaviour of clip renaming forcing that to be done by dialog rather than in the clip-handle drag-bar:
a) this is undocumented in the Manual
b) it is totally unexpected behavior from the labeling of the preference, with no indication that it also acts on clips
c) maybe there should be separate new **Tracks Behaviors** pref for  "**Use dialog for editing clip names**" - for greater clarity of purpose,

LWinterberg:
I could reproduce something similar once - I was able to rename the clip but no change showed until I pressed Enter. Same thing for resizing clips, they resized just fine but their visuals didn't update until I let go

petersampsonaudacity:
Testing on W10 with 3.2.0 I cannot reproduce this

Ryan-Miller-Audacity:
[...]

PR #4056, opened 2022-12-12
Audio clip rename fix

vsverchinsky:
Resolves: #3704 

Recommended:

### DIFF
--- a/src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
+++ b/src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
@@ -127,7 +127,8 @@ void NoteTrackAffordanceControls::Draw(TrackPanelDrawingContext& context, const 
             context.dc.SetTextBackground(wxTransparentColor);
             context.dc.SetTextForeground(theTheme.Colour(clrClipNameText));
             context.dc.SetFont(wxFont(wxFontInfo()));
-            TrackArt::DrawClipAffordance(context.dc, clipRect, nt->GetName(), highlight, selected);
+            const auto titleRect = TrackArt::DrawClipAffordance(context.dc, clipRect, highlight, selected);
+            TrackArt::DrawClipTitle(context.dc, titleRect, nt->GetName());
         }
     }
 }
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
@@ -139,8 +139,16 @@ WaveTrackAffordanceControls::WaveTrackAffordanceControls(const std::shared_ptr<T
 {
     if (auto trackList = pTrack->GetOwner())
     {
-        mSubscription = trackList->Subscribe(
-            *this, &WaveTrackAffordanceControls::OnTrackChanged);
+        mTrackListEventSubscription = trackList->Subscribe(
+            *this, &WaveTrackAffordanceControls::OnTrackListEvent);
+        if(auto project = trackList->GetOwner())
+        {
+            auto& viewInfo = ViewInfo::Get(*project);
+            mSelectionChangeSubscription =
+                viewInfo.selectedRegion.Subscribe(
+                    *this,
+                    &WaveTrackAffordanceControls::OnSelectionChange);
+        }
     }
 }
 
@@ -447,10 +455,15 @@ void WaveTrackAffordanceControls::ResetClipNameEdit()
     mEditedClip.reset();
 }
 
-void WaveTrackAffordanceControls::OnTrackChanged(const TrackListEvent& evt)
+void WaveTrackAffordanceControls::OnTrackListEvent(const TrackListEvent& evt)
 {
     if (evt.mType == TrackListEvent::SELECTION_CHANGE)
        ExitTextEditing();
+}
+
+void WaveTrackAffordanceControls::OnSelectionChange(NotifyingSelectedRegionMessage)
+{
+    ExitTextEditing();
 }
 
 unsigned WaveTrackAffordanceControls::ExitTextEditing()
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
@@ -269,19 +269,19 @@ void WaveTrackAffordanceControls::Draw(TrackPanelDrawingContext& context, const 
                    continue;
                 }
 
-                auto selected = GetSelectedClip().lock() == clip;
-                auto highlight = selected || affordanceRect.Contains(px, py);
+                const auto selected = GetSelectedClip().lock() == clip;
+                const auto highlight = selected || affordanceRect.Contains(px, py);
+                const auto titleRect = TrackArt::DrawClipAffordance(context.dc, affordanceRect, highlight, selected);
                 if (mTextEditHelper && mEditedClip.lock() == clip)
                 {
-                    auto titleRect = TrackArt::DrawClipAffordance(context.dc, affordanceRect, wxEmptyString, highlight, selected);
-                    mTextEditHelper->Draw(context.dc, titleRect);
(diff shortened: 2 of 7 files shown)


DavidBailes (changes requested):
This  fix introduces slightly different bugs when interacting using the keyboard.
Issue 1:
1. place the cursor just before the first clip.
2. Press Alt + period to select the first clip.
3. Press Ctrl + F2 to edit the name. Type in a new name and press Enter.
4. Press Alt + period to select the second clip.
5. Press Ctrl + F2 to edit the name. The name is NOT opened for editing. But pressing Ctrl + F2 again does open it for editing.

Issue 2.
Having selected a clip using the keyboard pressing Ctrl + F2 repeatedly keeps toggling whether or not the name is open for editing. It shouldn't - once opened, Ctrl + F2 should not close it.

vsverchinsky:
Thanks! Though I can't reproduce issue 1

DavidBailes:
> Thanks! Though I can't reproduce issue 1

If it works for renaming the second clip, then press Alt + comma to select the first clip, and press ctrl + f2 and see if that works.
(And I've got only one clip visible at once, as in the steps to reproduce.)

Had a look at the new code:
After pressing Alt + comma/period, WaveTrackAffordanceControls::Draw() gets called, but mFocusClip has not been updated, and so:
if(mFocusClip.lock() == clip)
                        mFocusedClipNameVisible = !titleRect.IsEmpty();
does not correctly update mFocusedClipNameVisible.

So when you try to rename a clip, in WaveTrackAffordanceControls::StartEditClipName():
            if(!mFocusedClipNameVisible)
                //Clip name isn't visible, there is no point in showing editor then
                return false;
an incorrect value of mFocusedClipNameVisible can be used.

vsverchinsky:
> This looks a lot better, however it introduces a bug for keyboard users. Prior to the fix, when the option "Use dialog for the name of a new label" was enabled, there were no problems. This is now not the case, when the start of a clip is visible, but its name is not. For example:
> image
> When you use alt + period to select the second clip in the example, the name of the clip remains invisible, because the select next clip command only ensures that the start of the clip is visible (it was written before there were clip names). So when you press ctrl+f2 to edit the clip name, nothing happens regardless of the "use dialog for the name of a new label" setting.
> In the case when the dialog setting is enabled, this is a new bug, and I think needs to be fixed i this pull request. One solution would be to not check for visibility of the name when this setting is enabled, as was done prior to this fix. This would ensure that it always worked.
> In the case where the dialog setting is not enabled, this is an existing bug, and maybe it could be treated as a separate bug.

Hm, behavior change is a bug, but it may also be a question to a design team @LWinterberg

DavidBailes:
> Also there is another case to be considered when editing via dialog is enabled: select a clip by clicking on the clip's title and the scroll Track Panel so that selection is out of view, then press Ctrl+F2. Dialog appears though clip isn't visible

I would expect the dialog to open, regardless of whether the clip is visible. Examples in other programs:
1. In file explorer in Windows, select a file, scroll so that it's no longer visible, press Alt+Enter, the properties dialog opens.
2. In Reaper, select and item, scroll so that it's no longer visible, run a command which opens a dialog for the item, such as show media item source properties, or any of the renaming commands, and a dialog opens.

In Audacity, when the dialog option is not enabled, then the current design appears to be that if the clip is not visible, and the user presses Ctrl+F2, nothing happens. I think it would be a lot more user friendly if in that case pressing Ctrl+F2 caused the name of the clip to be scrolled into view and then name could be changed. See for example in File Explorer in Windows: select a file, scroll so that it's no longer visible, press F2 to rename it, the file is scrolled into view, and the file can be renamed.

Obviously, for users of screen readers, a command silently failing because the clip is not visible is very unhelpful. Currently screen readers need the dialog option enabled for the renaming to be accessible, but when Audacity moves to Qt that hopefully will no longer be the case.

DavidBailes (approved):
This now looks fine. If the dialog option is enabled, you can rename a clip regardless of the visibility of the clip name.
The behaviour when the dialog option is not enabled, and the clip name is not visible is a separate issue (and I think could be made more user friendly).

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
```diff
@@ -46,9 +46,9 @@ class AUDACITY_DLL_API WaveTrackAffordanceControls :
     std::shared_ptr<TextEditHelper> mTextEditHelper;
 
     wxFont mClipNameFont;
-
-    //Helper flag, checked when text editing is triggered (and dialog-edit option is disabled)
-    bool mClipNameVisible { false };
+   
+    //Helper contianer used to track clips names visibility
```
Paul-Licameli, line 51: "Container"

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -570,13 +561,13 @@ namespace {
 void OnEditClipName(const CommandContext &context)
 {
    auto &project = context.project;
-   const auto [pTrack, pClip] = SelectedClipOfFocusedTrack(project);
-   if (pTrack && pClip) {
-      if (auto pAffordance = FindAffordance(*pTrack)) {
-         pAffordance->StartEditNameOfMatchingClip(project,
-            [pClip = pClip](auto &clip){ return &clip == pClip; });
-         // Refresh so the cursor appears
```
Paul-Licameli, line 578: It would be nice to keep the comment

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -139,8 +139,14 @@ WaveTrackAffordanceControls::WaveTrackAffordanceControls(const std::shared_ptr<T
...
-        mSubscription = trackList->Subscribe(
-            *this, &WaveTrackAffordanceControls::OnTrackChanged);
+        if(auto project = trackList->GetOwner())
+        {
+            auto& viewInfo = ViewInfo::Get(*project);
+            mSelectionChangeSubscription =
+                viewInfo.selectedRegion.Subscribe(
+                    *this,
+                    &WaveTrackAffordanceControls::OnSelectionChange);
+        }
```
Paul-Licameli, line 149: This adds detection of changes in the selected interval of time, but removes detection of changes of the binary selection state of tracks.

Why don't we do both, with two subscriptions?

src/tracks/playabletrack/notetrack/ui/NoteTrackAffordanceControls.cpp
```diff
@@ -127,7 +127,8 @@ void NoteTrackAffordanceControls::Draw(TrackPanelDrawingContext& context, const
             context.dc.SetTextBackground(wxTransparentColor);
             context.dc.SetTextForeground(theTheme.Colour(clrClipNameText));
             context.dc.SetFont(wxFont(wxFontInfo()));
-            TrackArt::DrawClipAffordance(context.dc, clipRect, nt->GetName(), highlight, selected);
+            const auto titleRect = TrackArt::DrawClipAffordance(context.dc, clipRect, highlight, selected);
+            TrackArt::DrawClipTitle(context.dc, titleRect, nt->GetName());
```
Paul-Licameli, line 131: Should analogous changes be made for note tracks?  It might make sense but there would be less demand for such corrections.

vsverchinsky: Not sure if I understand your question correctly, but there is no midi clip name editing feature yet.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
vsverchinsky: Will do

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
vsverchinsky: Thanks!

Paul-Licameli (approved):

Issue #3155, opened 2022-06-28
Reverb can not work in realtime


pietro68:
The built-in reverb effect was recently converted to work in realtime mode, but I found that this conversion was not complete, and there is actually no way that it could be completed.

It was not complete because changes in the settings were not propagated to the reverb "core" i.e. what is in Reverb_libSoX.h, but most importantly, they can not be propagated there anyway - a rewrite of the reverb core would be needed.

This because in Reverb_libSoX.h provides an api which only lets clients declare how they want their reverb, but no api calls to modify its parameters, between process calls. Adding such api calls looks impossible without a rewrite of the core, because some entities are allocated with a size which depends on the pre-delay value.

I suggest that for a start we turn off realtime support for the reverb (it's a one liner: make ::SupportsRealtime return false) and then see if Reverb_libSoX can be rewritten to support parameter changes while processing - however this is a much longer task and it is also not easy to make it sound without artifacts as the parameters are changed.

@Paul-Licameli I examined Reverb_libSoX which is only 270 lines of code and depends on no other files.  I think the rewrite to allow realtime variation of parameters is achievable, though I don't know yet if artifacts would prove to be bad enough to require further effort.   Here is a checklist of subtasks that I see.

- [x] Add a function `reverb_clear` that assigns zeroes to `one_pole_t::i1` and `one_pole_t::o1`, and `filter_t::store`, and the buffer in `filter_t` from subscripts 0 to `size` -- for each of the two one-poles and the many filters.  Call it from the RealtimeSuspend override. Then demonstrate the bug like  #3112 does not happen.
- [x] Replace `reverb_create` with two functions, `reverb_allocate` and `reverb_init`.  The first simply allocates all structures, with the maximum size that could be needed for any settings -- that means as if `depth == 2` and `scale == 1.0`.  The second function fills in parameter values assuming sufficient allocation.  It also assigns the `size` members of the filters.  (The capacity of the allocated buffers, maybe larger than this, does not need to be recorded for deallocation to work correctly later.)  Note that some sizes depend on the sample rate r but of course this can be assumed constant during play.
- [x] Define `filter_resize` which moves the contents in the circular buffer in filter_t for change of size, assuming sufficient capacity.  (Only the first `size` items are occupied.  If `size` increases, add zeros for the least-recent values.  You may need to std::rotate or memcpy two ranges, and when done, have `ptr == buffer + size`)
[...]

PR #4060, opened 2022-12-12
Added function reverb_clear and connected it to power off

pietro68:
Resolves: first point of #3155 

**Note for QA**: what needs to be verified here, is that issue #3112 does not happen anymore with the built-in reverb.

Recommended:

### DIFF
--- a/src/effects/Reverb_libSoX.h
+++ b/src/effects/Reverb_libSoX.h
@@ -268,3 +268,37 @@ static void reverb_delete(reverb_t * p)
    fifo_delete(&p->input_fifo);
 }
 
+static void reverb_clear(reverb_t* p)
+{
+   for (size_t c = 0; c < 2; c++)
+   {
+      auto& chn = p->chan[c];
+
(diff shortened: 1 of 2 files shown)


src/effects/Reverb_libSoX.h
```diff
@@ -268,3 +268,34 @@ static void reverb_delete(reverb_t * p)
...
 }
 
+static void reverb_clear(reverb_t* p)
+{
+   for (size_t c = 0; c < 2; c++)
+   {
+      auto& chn = p->chan[c];
+
+      chn.one_pole->i1 = 0.0;
+      chn.one_pole->o1 = 0.0;
```
Paul-Licameli, line 278: There are two one_pole objects in an array.   Both must be cleared.

pietro68: good catch! now done

Paul-Licameli (approved):

dozzzzer:
#3112 cannot be reproduced in this build, verified both on macOS and Windows

Issue #4063, opened 2022-12-13
BassTreble, Phaser, Reverb, WahWah should be available in sidebar


Paul-Licameli:
These four should be among the built-in realtime capable effects, and available for testing soon.

Should also arrange that when switching from 3.3 (or later) to an earlier Audacity, then back again -- the effects are in the sidebar only when you run 3.3 + .

There are known problems with Reverb not responding to adjustment of controls, to be fixed later.

It is not expected that the other threee have similar problems, but if such are discovered, then open other issues.

petersampsonaudacity:
I tested this on w10 with @Paul-Licameli 's branch build:  https://github.com/audacity/audacity/actions/runs/3685700498.

The four Audacity RT effects seem to action in the Effects Stack 

As Paul requested in his pull request I tested subsequently launching 3.2.2 and 3.2.1 and there I observe that the four Audacity RT effects are _**not**_ available in the Effects Stack.

PR #4064, opened 2022-12-13
Phaser, Wahwah, BassTreble, and Reverb available in the sidebar...

Paul-Licameli:
... Though Reverb has known problems with adjustment of controls, to be fixed.

Also bump REGVERCUR so that you don't need to delete an old pluginregistry.cfg first, and also, the effects will not be in the sidebar if you then downgrade to Audacity 3.2.2 or earlier.

Also renamed one enumeration constant to be clearer about its meaning.

Resolves: #4063

Recommended:


pietro68 (approved):
Code changes ok for me, and tested

Issue #3723, opened 2022-09-30
GetInfo: Type=Clips does not return clip names


SteveDaulton:
### Bug description

The scripting / macro command "GetInfo: Type=Clips" only returns partial information about the audio clips. The clip labels / names are missing.

### Steps to reproduce

1. Create a project with several named clips
2. "Extra menu > Scriptables II > GetInfo"
3. Select "Clips" as the "Type"
4. Observer that the returned data does not show the clip names

### Expected behavior

Example:
```

[ 
  { "track":0, "start":0, "end":15.2758, "color":0, "name":"Audio Track" },
  { "track":0, "start":15.2758, "end":64.7525, "color":0, "name":"Clip 2"  },
  { "track":0, "start":64.7525, "end":93.3522, "color":0, "name":"Chorus"  },
  { "track":0, "start":93.3522, "end":120, "color":0, "name":"Instrumental 1"  } ]
```

### Actual behavior

```
[ 
  { "track":0, "start":0, "end":15.2758, "color":0 },
  { "track":0, "start":15.2758, "end":64.7525, "color":0 },
  { "track":0, "start":64.7525, "end":93.3522, "color":0 },
  { "track":0, "start":93.3522, "end":120, "color":0 } ]

```

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows, Linux

### Additional context

Same issue on all OS.

This issue relates to a user question on the Audacity forum.

DoraTrix:
Just a bump here that the PR merge for this (4067) seems to be held up by a lack of review (the original reviewer deleted their account, not sure if that impacts the original approval).

dozzzzer:
@DoraTrix On Windows, every single row of the JSON output is printed in a separate dialog, one by one:


https://github.com/audacity/audacity/assets/70063970/a389ed15-c9eb-4160-a0df-ac9c42c64e41

On macOS, it works properly:

<img width="320" alt="Screenshot 2023-11-14 at 15 03 38" src="https://github.com/audacity/audacity/assets/70063970/7cedbf92-ff9c-47bb-b0e8-9e3f775abd8b">

crsib:
I don't think that this is a problem with this PR, though

dozzzzer:
@crsib you're absolutely right, my bad. A separate ticket for that is https://github.com/audacity/audacity/issues/5615

PR #4067, opened 2022-12-14
fix bug Audacity#3723: Add clip name to GetInfo: Type=Clips

DoraTrix:
Resolves: [#3723](https://github.com/audacity/audacity/issues/3723)

Adds "name" to the struct for the GetInfo command with type "Clips", uses the existing GetName() function in WaveClip.

Recommended:

### DIFF
--- a/src/commands/GetInfoCommand.cpp
+++ b/src/commands/GetInfoCommand.cpp
@@ -528,6 +528,7 @@ bool GetInfoCommand::SendClips(const CommandContext &context)
             context.AddItem(pClip->GetPlayStartTime(), "start");
             context.AddItem(pClip->GetPlayEndTime(), "end");
             context.AddItem(pClip->GetColourIndex(), "color");
+            context.AddItem(pClip->GetName(), "name");
             context.EndStruct();
          }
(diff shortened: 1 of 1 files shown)


enteredsername (approved):

marcushultman:
Would be nice to have this land.

marcushultman:
@Paul-Licameli Sorry for unsolicited pinging, but it appears this PR has fallen off everyone's radar, and I was wondering what it would take to get this to land. Not sure if the PR author is up for amending fixes, but I'd be happy to lend a hand where possible.

DoraTrix:
As far as I can tell, it just needs an additional approving review?  It looks like the original reviewer deleted their account which might un-approve it in that sense - but the refresh button to request a new review isn't in the sidebar, so... 🤷‍♀️

marcushultman:
@crsib Would you mind taking a look?

crsib (approved):

DoraTrix:
_sigh_

@crsib I botched the rebase, so I just added a new commit to the fix branch with the change.

Issue #3693, opened 2022-09-26
Restore the Cut, Copy and Paste buttons as a new (hidden by default) toolbar


petersampsonaudacity:
### Your idea

Restore the Cut, Copy and Paste buttons as a new toolbar that would be 

a) hidden by default to keep the revised 3.2 GUI clean

b) available for users to make visible in the upper tooldock with the View>Toolbars command

**Oversize mockup**
![image](https://user-images.githubusercontent.com/50205984/192318204-a46bcc69-6697-4e3d-9966-3964f6386dc6.png)

**In situ mockup**
![image](https://user-images.githubusercontent.com/50205984/192319720-87b5fbca-b7cd-4995-b961-5fee22129631.png)

Some users are already complaining about the loss of these buttons on the Forum:
https://forum.audacityteam.org/viewtopic.php?f=46&t=127074&p=456464
and this is causing some of them to revert to older versions of Audacity just to get the buttons back _(which is not ideal)_

### Problem to be solved

1)  some users are complaining about the removal of the Cut, Copy and Past buttons ffrom the Edit toolbar

2) some of those users are thus choosing to revert to older versions of Audacity, just to get the buttons back

### Prior art

Microsoft apps that I use have such buttons _(in addition to the "standard" shortcuts)_ - e.g.:
1. Microsoft Office: Word and Excel
2. Microsoft Paint

### Additional context

See Forum threads regarding the loss of these buttons:
**(Audacity 3.2.0) Please Bring Back Cut, Copy, & Paste Buttons]**
https://forum.audacityteam.org/viewtopic.php?f=70&t=127081#p456403)#

And:
**audacity old version download**
https://forum.audacityteam.org/viewtopic.php?f=46&t=127074

petersampsonaudacity:
Flagging @Tantacrul as a UX issue

Tantacrul:
Oh dear, I've never been called out as a UX issue before 😭

petersampsonaudacity:
> Oh dear, I've never been called out as a UX issue before 😭

No - you're a UX champion who can help these button-pushers get their buttons back - they'd love you for it   ;-)

Tantacrul:
No matter the function, if you remove something, someone's workflow will be interrupted. However, in this case there are already three other ways one can cut, copy or paste. It's just not something we should burden the interface with.

petersampsonaudacity:
A comment from a user on the Forum 

@LWinterberg Leo wrote in this thread
[...]

PR #4071, opened 2022-12-16
Added new toolbar which contains the cut, copy and paste

ksoze95:
Resolves: #3693 

Provide cut, copy and paste functionality from a toolbar.

This PR have the following couple of issues.

1. The paste toolbar button is enabled on the application starts up, it should be disabled until a copy action has first been preformed.
2. The new toolbar should be hidden by default.  When ToolManager.cpp was modified to hide the toolbar, audio setup and both meters toolbars are moved to above all the other toolbars.
 

Recommended:


libraries/lib-theme/AllThemeResources.h
```diff
@@ -108,6 +108,14 @@ from there.  Audacity will look for a file called "Pause.png".
    DEFINE_IMAGE( bmpRedo, wxImage( 26, 24 ), wxT("Redo"));
    DEFINE_IMAGE( bmpRedoDisabled, wxImage( 26, 24 ), wxT("RedoDisabled"));
 
+   SET_THEME_FLAGS(  resFlagPaired  );
+   DEFINE_IMAGE( bmpCut, wxImage( 20, 20 ), wxT("Cut"));
+   DEFINE_IMAGE( bmpCutDisabled, wxImage( 20, 20 ), wxT("CutDisabled"));
+   DEFINE_IMAGE( bmpCopy, wxImage( 20, 20 ), wxT("Copy"));
+   DEFINE_IMAGE( bmpCopyDisabled, wxImage( 20, 20 ), wxT("CopyDisabled"));
+   DEFINE_IMAGE( bmpPaste, wxImage( 20, 20 ), wxT("Paste"));
+   DEFINE_IMAGE( bmpPasteDisabled, wxImage( 20, 20 ), wxT("PasteDisabled"));
```
Paul-Licameli, line 117: After this is merged, remember to tell @LWinterberg to update the "sprite sheet" in documentation for users who customize the theme.

src/toolbars/CutCopyPasteToolBar.cpp
```diff
@@ -0,0 +1,174 @@
...
+BEGIN_EVENT_TABLE( CutCopyPasteToolBar, ToolBar )
+   EVT_COMMAND_RANGE(TBCutID+first_TB_ID,
+                      TBCutID+first_TB_ID + TBNumButtons - 1,
+                      wxEVT_COMMAND_BUTTON_CLICKED,
+                      CutCopyPasteToolBar::OnButton )
+END_EVENT_TABLE()
+
+//Standard constructor
+CutCopyPasteToolBar::CutCopyPasteToolBar( AudacityProject &project )
+: ToolBar(project, CutCopyPasteBarID, XO("Cut/Copy/Paste"), wxT("Cut/Copy/Paste"))
```
Paul-Licameli, line 90: For the fourth argument -- an internal identifier the user doesn't see, which may be used as a key in the configuration file -- do not use the '/' character.  I think it might get misinterpreted as multiple config file path components.  Just omit those characters.

Paul-Licameli (changes requested):

src/toolbars/CutCopyPasteToolBar.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+   RegenerateTooltips();
+}
+
+void CutCopyPasteToolBar::UpdatePrefs()
+{
+   RegenerateTooltips();
+
+   // Set label to pull in language change
+   SetLabel(XO("Cut/Copy/Paste"));
```
Paul-Licameli, line 140: Make a literal instead of repeating the string constant

src/toolbars/CutCopyPasteToolBar.cpp
```diff
@@ -0,0 +1,174 @@
...
+  menus. The window can be embedded within a normal project
+  window, or within a ToolBarFrame.
+
+  All of the controls in this window were custom-written for
+  Audacity - they are not native controls on any platform -
+  however, it is intended that the images could be easily
+  replaced to allow "skinning" or just customization to
+  match the look and feel of each platform.
+
+*//*******************************************************************/
```
Paul-Licameli, line 28: This entire file is a cut-and-paste of EditToolBar with some modifications.

These two beg to be unified into a common base class, simply parameterized by a table of buttons and commands.

Such a simple, table driven toolbar might even become a useful toolkit neutral facade if it is done right.

None of this comment is a request for changes now.

src/toolbars/ToolBarButtons.cpp
```diff
@@ -0,0 +1,125 @@
+#include <utility>
```
Paul-Licameli, line 1: This new source file should have comments at top about its purpos

src/toolbars/CutCopyPasteToolBar.cpp
```diff
@@ -0,0 +1,174 @@
...
+static RegisteredToolbarFactory factory{ CutCopyPasteBarID,
+   []( AudacityProject &project ){
+      return ToolBar::Holder{ safenew CutCopyPasteToolBar{ project } }; }
+};
+
+#include "ToolManager.h"
+
+namespace {
+AttachedToolBarMenuItem sAttachment{
+   /* i18n-hint: Clicking this menu item shows the toolbar for editing */
```
Paul-Licameli, line 170: I suggest an OrderingHint to control the position of insertion of this menu item.  Did design specify it?  Maybe it should come after the edit toolbar, not at the bottom of the menu.

src/toolbars/ToolBarButtons.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+AButton* ToolBarButtons::CreateButton(teBmps eEnabledUp, teBmps eEnabledDown, teBmps eDisabled, int thisButtonId, const TranslatableString& label, bool toggle)
+{
+   AButton *&r = mButtons[thisButtonId];
+
+   r = ToolBar::MakeButton(mParent,
+      bmpRecoloredUpSmall, bmpRecoloredDownSmall, bmpRecoloredUpHiliteSmall, bmpRecoloredHiliteSmall,
+      eEnabledUp, eEnabledDown, eDisabled,
+      wxWindowID(thisButtonId + mFirstButtonId),
```
Paul-Licameli, line 61: Two integers are passed into this function, just so they can be added.

Why not simplify, passing in just the sum?

src/toolbars/ToolBarButtons.cpp
```diff
@@ -0,0 +1,125 @@
...
+   TBActEnableDisable = 2,
+};
+
+ToolBarButtons::ToolBarButtons(ToolBar *const parent, AudacityProject & project, ButtonList buttonList, int size, int firstButtonId)
+   : mParent(parent)
+   , mProject(project)
+   , mButtonList(std::move(buttonList))
+   , mFirstButtonId(firstButtonId)
+{
+   mButtons.resize(size);
```
Paul-Licameli, line 26: I prefer member initializations where possible instead of mutations inside the constructor body.

But beware, it should be
`mButtons(size)`

not the curly-bracket form!  Because `vector` has constructors taking `initializer_list` and you don't want overload resolution to one of those.

src/toolbars/ToolBarButtons.cpp
```diff
@@ -0,0 +1,125 @@
...
+enum {
+   TBActTooltips = 1,
+   TBActEnableDisable = 2,
+};
+
+ToolBarButtons::ToolBarButtons(ToolBar *const parent, AudacityProject & project, ButtonList buttonList, int size, int firstButtonId)
+   : mParent(parent)
+   , mProject(project)
+   , mButtonList(std::move(buttonList))
+   , mFirstButtonId(firstButtonId)
```
Paul-Licameli, line 24: The sequence of intializers should match the sequence of member declarations, because it is the latter that really determines the sequence of execution.

It is unimportant in this case but I think this is one of the warnings we should elevate to errors in our build.  In some cases there may be surprises.

src/toolbars/EditToolBar.cpp
```diff
@@ -59,6 +59,54 @@
...
+   { ETBZoomFitID,  wxT("FitInWindow"), XO("Fit project to width")  },
+
+   { ETBTrimID,     wxT("Trim"),        XO("Trim audio outside selection")  },
+   { ETBSilenceID,  wxT("Silence"),     XO("Silence audio selection")  },
+#ifdef OPTION_SYNC_LOCK_BUTTON
+   { ETBSyncLockID, wxT("SyncLock"),    XO("Sync-Lock Tracks")  },
+#endif
+   { ETBUndoID,     wxT("Undo"),        XO("Undo")  },
+   { ETBRedoID,     wxT("Redo"),        XO("Redo")  },
+};
```
Paul-Licameli, line 108: Please avoid unnecessary movements of functions and tables within files.  It makes review more difficult.  You could instead leave them where they were, so that the diff contains fewer lines.

src/toolbars/ToolBarButtons.cpp
Paul-Licameli: OK, a later commit changes this

src/toolbars/ToolBarButtons.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+   ::HandleTextualCommand( cm, mButtonList[id].commandName, context, flags, false );
+
+#if defined(__WXMAC__)
+   // Bug 2402
+   // LLL: It seems that on the Mac the IDLE events are processed
+   //      differently than on Windows/GTK and the AdornedRulerPanel's
+   //      OnPaint() method gets called sooner that expected. This is
+   //      evident when zooming from this toolbar only. When zooming from
+   //      the Menu or from keyboard ommand, the zooming works correctly.
```
Paul-Licameli, line 49: "command"

src/toolbars/EditToolBar.cpp
```diff
@@ -129,49 +173,49 @@ void EditToolBar::Populate()
    /* Buttons */
    // Tooltips match menu entries.
    // We previously had longer tooltips which were not more clear.
-   AddButton(this, bmpZoomIn, bmpZoomIn, bmpZoomInDisabled, ETBZoomInID,
+   AddButton(bmpZoomIn, bmpZoomIn, bmpZoomInDisabled, ETBZoomInID,
```
Paul-Licameli, line 176: I like the effort to unify repepetious code into a utility class.

It might go farther (in 3.3 development).

The `Populate()` routines might also become completely table-driven if `Entry` could indicate a separator and also what the initial enabled state should be.

The `_EVENT_TABLE` macros might be removed and instead `ToolBarButtons` could `Bind` an event handler to each button as it is made.  (The preferred thing in newer wxWidgets code)

It might even be not too difficult to devise a platform-neutral facade class to unify these two simple toolbars, and also use a registry of button entries so that an add-on might insert buttons.

src/toolbars/ToolBarButtons.h
```diff
@@ -0,0 +1,65 @@
...
+#include <functional>
+#include <vector>
+#include "ToolBar.h"
+
+class wxCommandEvent;
+
+class AudacityProject;
+class AButton;
+
+struct Entry {
```
Paul-Licameli, line 23: I don't like a simple bland name like `Entry` in the global namespace.  This might be a nested type in `ToolBarButtons`.

Paul-Licameli:
No more to say in this round of review.

src/toolbars/ToolBarButtons.cpp
crsib: I think we should stop that practice and stick just to SPDX license identifiers

src/toolbars/ToolBarButtons.cpp
crsib: I am not quite sure what do you mean

Paul-Licameli: I was reviewing the commits sequentially.  My comment applies to a step of the code transformation that is not the last.  The final version uses only one id argument.

src/toolbars/ToolBarButtons.cpp
```diff
@@ -0,0 +1,136 @@
...
```
Paul-Licameli, line 60: Still misspelled

src/toolbars/ToolManager.cpp
```diff
@@ -669,6 +673,9 @@ void ToolManager::Reset()
       }
 
    }
+
+   for (auto ndx : HideAfterReset)
+      Expose(ndx, false);
```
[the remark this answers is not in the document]
crsib: This is required, so we can properly setup the whole chain of neighbors.

src/toolbars/ToolBar.cpp
```diff
@@ -641,6 +646,12 @@ ToolDock *ToolBar::GetDock()
    return dynamic_cast<ToolDock*>(GetParent());
 }
 
+void ToolBar::SetPreferredNeighbors(ToolBarID left, ToolBarID top)
+{
+   mPreferredLeftNeighbor = left;
+   mPreferredTopNeighbor = top;
+}
```
Paul-Licameli, line 653: You rewrote my proposal with more generality.  Very well.

But as you might guess, I don't really like `DefaultConfigEntry` because it is a failure of open-endedness of the set of toolbars.  I would rather take the time fo make a new Registry that scatters the definition of the default configuration.

Each toolbar should decide where to insert itself.

For that, each toolbar's .cpp file would need to refer to another toolbar, but only the weak for of dependency that mentions a textual id, and if the toolbar module isn't present, there is no harm.

My pull request Open-ended-toolbars does what is needed to eliminate the centralized enumeration and identify toolbars by strings instead.  It doesn't yet go further to build up the default configuration by collecting registry items.

Paul-Licameli (approved):

Issue #3216, opened 2022-07-12
Deadlock when loading PhoenixVerb plugin in non-GUI mode


dozzzzer:
**Steps To Reproduce**

1. load a track
2. in the Effect menu select [PhoenixVerb plugin (version 6.0.1)](https://www.izotope.com/en/products/downloads/phoenixverb/release-notes.html) (VST3) 
3. disable graphical interface, close the plugin
4. load the plugin again

**Actual behavior**
Loading the plugin in plain GUI mode causes deadlock due to the huge (>1000) number of parameters to handle.

**Expected behavior**
The plugin should be blacklisted

**Additional information (please complete the following information):**

- OS: Windows 10
- Version: [audacity-win-3.2.0-alpha-20220629+48b9c64-x64-msvc2022](https://github.com/audacity/audacity/suites/7135653719/artifacts/283447296)

UPD: instead of blacklisting the plugin, UI loading time was improved by removing MIDI CC parameters from the list

dozzzzer:
Verified

PR #4075, opened 2022-12-16
Hide MIDI CC parameters from VST3's plain UI

vsverchinsky:
Resolves: #3216 

Recommended:

### DIFF
--- a/src/effects/VST3/VST3ParametersWindow.cpp
+++ b/src/effects/VST3/VST3ParametersWindow.cpp
@@ -255,6 +255,15 @@ VST3ParametersWindow::VST3ParametersWindow(wxWindow *parent,
       if((parameterInfo.flags & (Vst::ParameterInfo::kCanAutomate | Vst::ParameterInfo::kIsBypass | Vst::ParameterInfo::kIsReadOnly)) == 0)
          continue;
 
+      {
+         // Hide proxy parameters with name that starts with "MIDI CC "
+         // That prevents Plain UI from creating many useless controls
+         static_assert(sizeof(Steinberg::tchar) == sizeof(char16_t));
+         static const std::basic_string_view<tchar> MIDI_CC { reinterpret_cast<const tchar*>(u"MIDI CC ") };
+         if(std::basic_string_view<tchar>(parameterInfo.title).rfind(MIDI_CC, 0) == 0)
+            continue;
+      }
+      
       if (parameterInfo.stepCount != 1)      // not a toggle
          sizer->Add(safenew wxStaticText(
             this,


Paul-Licameli (changes requested):

src/effects/VST3/VST3ParametersWindow.cpp
```diff
@@ -255,6 +265,14 @@ VST3ParametersWindow::VST3ParametersWindow(wxWindow *parent,
       if((parameterInfo.flags & (Vst::ParameterInfo::kCanAutomate | Vst::ParameterInfo::kIsBypass | Vst::ParameterInfo::kIsReadOnly)) == 0)
          continue;
 
+      if(hideMIDICC)
+      {
+         static_assert(sizeof(Steinberg::tchar) == 2, "Make sure that UNICODE is defined in VST3");
+         static const std::u16string_view MIDI_CC { u"MIDI CC " };
+         if(std::u16string_view(parameterInfo.title).rfind(MIDI_CC, 0) == 0)
```
Paul-Licameli, line 272: Is this a reliable naming convention in all effects, or is this only a special case fix for one or a few problematic effects from one vendor?

Is there no other reliable way to decide this from the information in ParameterInfo?

src/effects/VST3/VST3OptionsDialog.cpp
```diff
@@ -100,6 +102,17 @@ void VST3OptionsDialog::PopulateOrExchange(ShuttleGui & S)
                           mUseGUI);
          }
          S.EndStatic();
+         S.StartStatic(XO(""));
+         {
+            S.AddVariableText(XO(
+"Don't list MIDI CC parameters in plain UI mode."
```
Paul-Licameli, line 108: What does "CC" abbreviate?  I still don't know.  Would an uninstructed user know?

src/effects/VST3/VST3OptionsDialog.cpp
```diff
@@ -100,6 +102,17 @@ void VST3OptionsDialog::PopulateOrExchange(ShuttleGui & S)
...
+" Reopen the effect for this to take effect.\n"
+"Warning: disabling this parameter in some cases"
+" may significantly increase plugin UI loading time."),
+               false, 0, 650);
+            S.TieCheckBox(XXO("Hide MIDI CC"), mHideMIDICC);
+         }
+         S.EndStatic();
```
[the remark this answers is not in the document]
vsverchinsky: Thanks!

src/effects/VST3/VST3ParametersWindow.cpp
vsverchinsky: MIDI Control Change.

> Is this a reliable naming convention in all effects, or is this only a special case fix for one or a few problematic effects from one vendor?

JUCE framework uses such naming template. 

> Is there no other reliable way to decide this from the information in ParameterInfo?

I've spent some time reading the documentation and Steinberg forums and didn't find one.

src/effects/VST3/VST3OptionsDialog.cpp
vsverchinsky: I'll put that into quotes, in that case "MIDI CC" relates to parameter title rather than a term.
On the other hand, considering that we don't support VSTi yet, it might be better remove that option at all. @LWinterberg

LWinterberg: I'd say you can safely hide all the MIDI stuff in plainUI mode for the time being, yeah.

Paul-Licameli (approved):

PR #4077, opened 2022-12-16
Issue 3155 - Reverb can not work in realtime

pietro68:
Resolves: #3155 

Recommended:

### DIFF
--- a/src/effects/Reverb.cpp
+++ b/src/effects/Reverb.cpp
@@ -202,10 +202,14 @@ struct EffectReverb::Instance
 
    // Realtime section
 
-   bool RealtimeInitialize(EffectSettings& settings, double) override
+   bool RealtimeInitialize(EffectSettings& settings, double sampleRate) override
    {
       SetBlockSize(512);
       mSlaves.clear();
+
+      mLastAppliedSettings = GetSettings(settings);
+      mLastSampleRate = sampleRate;
+
       return true;
    }
 
@@ -233,6 +237,26 @@ struct EffectReverb::Instance
    size_t RealtimeProcess(size_t group, EffectSettings& settings,
       const float* const* inbuf, float* const* outbuf, size_t numSamples) override
    {
+
+      const auto& incomingSettings = GetSettings(settings);
+      if ( !(incomingSettings == mLastAppliedSettings) )
+      {
+         for (auto& slave : mSlaves)
+         {
+            for (unsigned int i = 0; i < slave.mState.mNumChans; i++)
+            {
+               auto& reverbCore = slave.mState.mP[i].reverb;
+               const auto& is = incomingSettings;
+               reverb_init(&reverbCore, mLastSampleRate,
+                           is.mWetGain, is.mRoomSize, is.mReverberance, is.mHfDamping,
+                           is.mPreDelay, is.mStereoWidth, is.mToneLow, is.mToneHigh   );
+            }
+         }         
+
+         mLastAppliedSettings = incomingSettings;
+      }
+
+
       if (group >= mSlaves.size())
          return 0;
       return InstanceProcess(settings, mSlaves[group].mState, inbuf, outbuf, numSamples);
@@ -273,6 +297,9 @@ struct EffectReverb::Instance
    std::vector<EffectReverb::Instance> mSlaves;
 
    unsigned mChannels{ 2 };
+
+   EffectReverbSettings mLastAppliedSettings;
+   double mLastSampleRate{ 0 };
 };
 
 
@@ -590,3 +617,19 @@ void EffectReverb::Validator::OnCheckbox(wxCommandEvent &evt)
 }
 
 #undef SpinSliderHandlers
+
+bool operator==(const EffectReverbSettings& a, const EffectReverbSettings& b)
+{
+   // With C++20, all of this can be replaced by =default
+   return      (a.mRoomSize     == b.mRoomSize)
+            && (a.mPreDelay     == b.mPreDelay)
+            && (a.mReverberance == b.mReverberance)
+            && (a.mHfDamping    == b.mHfDamping)
+            && (a.mToneLow      == b.mToneLow)
+            && (a.mToneHigh     == b.mToneHigh)
+            && (a.mWetGain      == b.mWetGain)
+            && (a.mDryGain      == b.mDryGain)
+            && (a.mStereoWidth  == b.mStereoWidth)
+            && (a.mWetOnly      == b.mWetOnly);           
+}
+
--- a/src/effects/Reverb_libSoX.h
+++ b/src/effects/Reverb_libSoX.h
@@ -218,6 +218,42 @@ typedef struct {
    float * out[2];
 } reverb_t;
 
+
+static void reverb_allocate(reverb_t* p, size_t buffer_size, float** out)
+{
+   memset(p, 0, sizeof(*p));
+
+   // Input queue
+   fifo_create(&p->input_fifo, sizeof(float));
+
+   // Outputs
+   out[0] = lsx_zalloc(p->out[0], buffer_size);
+   out[1] = lsx_zalloc(p->out[1], buffer_size);
+}
+
+static void reverb_init
+(
+   reverb_t* p,
+   double sample_rate_Hz,
+   double wet_gain_dB,
+   double reverberance,
+   double hf_damping,
+   double pre_delay_ms
+)
+{
+   // Input queue 
+   size_t delay = pre_delay_ms / 1000 * sample_rate_Hz + .5;
+   memset(fifo_write(&p->input_fifo, delay, 0), 0, delay * sizeof(float));
+
+   // Feedback, Damping, Gain
+   double a = -1 / log(1 - /**/.3 /**/);           /* Set minimum feedback */
+   double b = 100 / (log(1 - /**/.98/**/) * a + 1);  /* Set maximum feedback */
+   p->feedback = 1 - exp((reverberance - b) / (a * b));
+   p->hf_damping = hf_damping / 100 * .3 + .2;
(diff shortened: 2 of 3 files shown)


src/effects/Reverb.cpp
```diff
@@ -233,6 +237,26 @@ struct EffectReverb::Instance
...
+               auto& reverbCore = slave.mState.mP[i].reverb;
+               const auto& is = incomingSettings;
+               reverb_init(&reverbCore, mLastSampleRate,
+                           is.mWetGain, is.mRoomSize, is.mReverberance, is.mHfDamping,
+                           is.mPreDelay, is.mStereoWidth, is.mToneLow, is.mToneHigh   );
+            }
+         }         
+
+         mLastAppliedSettings = incomingSettings;
+      }
```
Paul-Licameli, line 257: The block of code above might instead be inserted into a new override of RealtimeProcessStart, to be more consistent with how other effects receive notifications of changes of settings, and also later to allow this to be rewritten to use EffectMessage instead.

For now it is not a blocker to merge this.

src/effects/Reverb_libSoX.h
```diff
@@ -218,6 +268,174 @@ typedef struct {
...
+   // Input queue
+   fifo_create(&p->input_fifo, sizeof(float));
+
+   // Outputs
+   out[0] = lsx_zalloc(p->out[0], buffer_size);
+   out[1] = lsx_zalloc(p->out[1], buffer_size);
+
+   // Allpass & Comb filters
+   filter_array_allocate(p->chan + 0, sample_rate_Hz, 1.0, 0.0);
+   filter_array_allocate(p->chan + 1, sample_rate_Hz, 1.0, 1.0);
```
Paul-Licameli, line 394: Comment it here, that you pass the numbers that will cause the maximum possible allocation of the buffer just once, even if it is in excess of what is required by the settings passed in initialization after the allocation.

src/effects/Reverb_libSoX.h
```diff
@@ -218,6 +268,174 @@ typedef struct {
...
+   double reverberance,
+   double hf_damping,
+   double pre_delay_ms,
+   double stereo_depth,
+   double tone_low,       /* % */
+   double tone_high       /* % */
+)
+{
+   // Input queue
+   p->input_fifo.allocation = FIFO_MIN;
```
Paul-Licameli, line 412: Is it correct to clear the FIFO whenever you reinitialize?

Or should this be done only when you also use `filter_t_resize_resetting`?

Paul-Licameli: Or maybe instead, write a new function that overwrites the fifo with zeroes.

src/effects/Reverb_libSoX.h
```diff
@@ -218,6 +268,174 @@ typedef struct {
...
+   //        V
+   //  | C B A |        smaller size, and ptr would be at the new right edge; do nothing.
+   // 
+   //      v
+   //  | B A |          smaller size, and ptr would be beyond the new right edge: left-shift what is ahead of ptr,
+   //                   AND move ptr to size-1
+
+   assert(newSize > 0);
+
+   const int    sizeDiff = (newSize - p->size);
```
Paul-Licameli, line 314: DANGER -- mixing unsigned and signed arithmetic.

I think `newSize` and `p->size` which are both `size_t` should first be each cast to `ssize_t` before the subtraction,  which may make a negative value.

Paul-Licameli (changes requested):

src/effects/Reverb_libSoX.h
```diff
@@ -218,6 +268,174 @@ typedef struct {
...
+      }
+      else if (ptrPos == newSize - 1)
+      {
+         // sub-case: ptr is at the new edge - no shifting to do, and ptr can stay where it is
+      }
+      else
+      {
+         // sub-case: ptr would be beyond the new edge
+         // left-shift what is ahead of ptr and make ptr point to the new edge
+         size_t samplesBeyond = ptrPos - newSize + 1;
```
Paul-Licameli, line 350: unused variable

src/effects/Reverb_libSoX.h
```diff
@@ -148,8 +148,68 @@ typedef struct {
    one_pole_t one_pole[2];
 } filter_array_t;
 
+static void set_eq(filter_array_t* p, double rate,
```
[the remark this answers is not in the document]
pietro68: yes, done

src/effects/Reverb.cpp
```diff
@@ -319,7 +346,7 @@ EffectType EffectReverb::GetType() const
 
 auto EffectReverb::RealtimeSupport() const -> RealtimeSince
 {
-   return RealtimeSince::Never;
+   return RealtimeSince::Always;
```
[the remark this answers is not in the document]
pietro68: I have done the rebasing on it.

src/effects/Reverb_libSoX.h
```diff
@@ -218,6 +268,174 @@ typedef struct {
...
+{
+   // Choose your resize method.
+   //
+   // When moving the Room Size slider:
+   //
+   // - using resize_resetting, you will not hear previous reverberations,
+   //   but at the same time you will not hear sound artifacts
+   //
+   // - using resize_preserving, you will hear previous reverberations,
+   //   but sometimes you might hear sound artifacts
```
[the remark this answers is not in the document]
pietro68: Basically the story is: I started doing the more complicate thing (i.e. preserving as much as possible of what is in the comb/allpass filters) and that ended up having sound artifacts, whereas the simple solution had no artifacts (but then users will hear no reverberation as long as they move the slider).

About more elaborate ideas: we could cross-fade between an instance running with the last settings before a slider was moved, and an instance with the settings currently set by the sliders. But that requires double processing and further complications, I doubt that it would be worth.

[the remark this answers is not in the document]
pietro68: > Is the Room Size slider perhaps the only one for which it would be very importatn?

I think so.

src/effects/Reverb_libSoX.h
pietro68: done

src/effects/Reverb_libSoX.h
pietro68: Yes, I found out that if I did not do fifo_clear when I reinitialize, then the output would not sound right (specifically: when dragging the room size slider from 100% to 0% while playing, what I would get would not sound as if I would have started the reverb at 0% from scratch)

And it is true, there is no need to also set .allocation to FIFO_MIN - not doing that, can avoid later allocations. I changed that.

src/effects/Reverb_libSoX.h
```diff
@@ -218,6 +268,174 @@ typedef struct {
...
+   p->size = newSize;
+}
+
+
+// Resizes a filter, trying to keep as much of its history as possible.
+static void filter_t_resize_preserving(filter_t* p, size_t newSize)
+{
+   // Imagine we have this filter_t as input:
+   //
+   //                  ptr goes from right to left and when falling off the left side, it warps back to the right;
```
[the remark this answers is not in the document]
pietro68: I always get this mixed up.

src/effects/Reverb_libSoX.h
pietro68: Totally right - done it.

src/effects/Reverb_libSoX.h
pietro68: ok, removed

Paul-Licameli (approved):

Paul-Licameli (changes requested):
Wrong approach to fixing the crash; and more needed to ensure the calculations are the same as in the original

src/effects/Reverb_libSoX.h
```diff
@@ -203,13 +204,13 @@ static void filter_array_init(filter_array_t* p, double rate,
...
       size_t newSize = (size_t)(scale * r * (comb_lengths[i] + stereo_adjust * offset) + .5);
-      filter_t_resize(pcomb, newSize);
+      filter_t_resize(pcomb, std::min(newSize, pcomb->allocSize));
    }
    for (i = 0; i < array_length(allpass_lengths); ++i, offset = -offset)
    {
       filter_t* pallpass = &p->allpass[i];
       size_t newSize = (size_t)(r * (allpass_lengths[i] + stereo_adjust * offset) + .5);
-      filter_t_resize(pallpass, newSize);
+      filter_t_resize(pallpass, std::min(newSize, pallpass->allocSize));
```
Paul-Licameli, line 213: No.  Something is wrong here.  This change makes the crash go away but changes the calculations.

We thought we allocated sufficient capacity for the largest size required by any parameters.  The fact of the crash tells us our reasoning was wrong somewhere.  Look closer to see where.

Here it is:  in `filter_array_allocate` is this line (twice):
```
   for (i = 0; i < array_length(comb_lengths); ++i, offset = -offset)
```

So as `offset` (a double between 0 and 1) becomes larger, the size of odd-numbered comb and allpass filters becomes *smaller*.

The alternation of offset between positive and negative should be preserved in `filter_array_init` so it computes the same results as before, but should be removed from the allocation function, so that all the filters have sufficient capacity for the size that is computed for any stereo depth value.

(Or really it could alternate between the given offset value and 0 and still make sufficient capacity.)

Please make that change instead.

And make the other changes I suggested too, which are not necessary to fix the crash perhaps, but should be done to be sure the same results are computed.

That is:  `reverb_t` should remember whether `depth` was 0 when `reverb_init` was done, and test that, and not whether `p->out[i]` is null, in `reverb_process`.

pietro68: ok, all done now

Paul-Licameli (approved):

Issue #3807, opened 2022-10-14
After crashing and restart, project recovery won't recover the audio track


dozzzzer:
### Steps to reproduce

1. download & install the build [from here](https://github.com/audacity/audacity/actions/runs/3244638840) to be able to simulate a crash consistently
2. record some audio in Track 1
3. select and effect from the main Effect menu
4. observe the crash 
5. restart Audacity
6. click **Recover Selected**
7. observe

### Expected behavior

Audio track has been recovered

### Actual behavior

Audio track is lost

### Audacity Version

other

### Operating system

Windows, macOS

Paul-Licameli:
With a debug build from source, I failed to reproduce, at least at the first try.  Maybe there is intermittency or I need a relase build.

How reliably can you reproduce it?

dozzzzer:
@Paul-Licameli 
I can reliably reproduce it both on macOS and Windows with the only difference: on Mac you have to manually force quit the program when it freezes, while on Windows it crashes and closes immediately.
Try to follow the exact steps from the video I've attached.

Paul-Licameli:
Are you certain of the correct link to the build?

Paul-Licameli:
@dozzzzer I would like you to contrast behavior of two builds.  This one: https://github.com/audacity/audacity/actions/runs/3725302570

... reintroduces the crash bug that existed in application of destructive effects, so that you may restest the scenario and see whether the autosaved projects would still lose data.  (But I still fail to reproduce this.)

Paul-Licameli:
This build: https://github.com/audacity/audacity/actions/runs/3725317870

... will still reproduce that crash but contains a possible fix for autosave behavior.

Outcomes of testing may be:
- The autosave in the first build loses the track, but in the second, does not.  What we hope for.
- Both builds recover the project correctly.  We can no longer reproduce the original report.  There is no proof that my changes fixes anything.  Neither does it appear to do harm.  Code review may decide it is desirable anyway.
- Both builds fail to recover the project.  A disproof that my change is a fix.  But, again not harmless and still code review might recomment to merge it.

Paul-Licameli:
I see my problem now -- take step 2 literally!  Data can be lost ONLY in case recording was the last undoable thing before the crash.

dozzzzer:
Verified!

PR #4084, opened 2022-12-18
Fix unsafe autosave

Paul-Licameli:
Resolves: #3807

I still have not reproduced #3807.  This is a fix for a possible failure of autosave to keep audio track data -- which I have seen sometimes, but only in my own modified builds, and not in any commit of the master branch.

Recommended:


Paul-Licameli:
Rethinking this

PR #4085, opened 2022-12-19
Fix x42 crash on close

Paul-Licameli:
Resolves: #4096

Fix crash that happens on macOS when closing project window or program,
as with Command+W or Command+Q, while the non-modal effect dialog is
open, of an x42 plug-in.

See commit 32a6779, which compensated for memory management bugs in these
plug-ins before, but not this case.

This strange fix pushes some event handlers on the effect dialog and the main
project window, to control their sequence of destruction in a way that works
around the bug, but still also depending on the fix at 32a6779 to correct an
unbalanced extra CFRelease somewhere in the foreign code.

Code comments explain more about the simpler attempted fixes that were not
enough.

Recommended:

### DIFF
--- a/src/effects/RealtimeEffectStateUI.cpp
+++ b/src/effects/RealtimeEffectStateUI.cpp
@@ -30,7 +30,7 @@ const RealtimeEffectState::RegisteredFactory realtimeEffectStateUIFactory { [](a
 
 
 BEGIN_EVENT_TABLE(RealtimeEffectStateUI, wxEvtHandler)
-   EVT_CLOSE(RealtimeEffectStateUI::OnClose)
+   EVT_CLOSE(RealtimeEffectStateUI::OnCloseDialog)
 END_EVENT_TABLE()
 
 RealtimeEffectStateUI::RealtimeEffectStateUI(RealtimeEffectState& state)
@@ -109,6 +109,7 @@ void RealtimeEffectStateUI::Show(AudacityProject& project)
          // project state is destroyed
          Hide(&project);
       });
+   mSubscriber.PushHandler(project);
 
    mParameterChangedSubscription = mEffectUIHost->GetValidator()->Subscribe(
       [this](auto) { UndoManager::Get(*mpProject).MarkUnsaved(); });
@@ -123,6 +124,9 @@ void RealtimeEffectStateUI::Hide(AudacityProject* project)
       mEffectUIHost->Close();
       mEffectUIHost = {};
    }
+   if (mpProjectWindow)
+      // Closing the dialog normally, did not intercept a main window closing
+      mpProjectWindow->RemoveEventHandler(&mSubscriber);
 }
 
 void RealtimeEffectStateUI::Toggle(AudacityProject& project)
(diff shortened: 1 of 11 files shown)


src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -193,3 +198,36 @@ void RealtimeEffectStateUI::OnClose(wxCloseEvent & evt)
    if (next)
       next->ProcessEvent(evt);
 }
+
+void RealtimeEffectStateUI::ProjectWindowSubscriber::
+PushHandler(AudacityProject &project)
+{
+   auto& projectWindow = ProjectWindow::Get(project);
+   if (mState.mpProjectWindow == &projectWindow)
```
crsib, line 256: I am not a fan of empty `if` closes and it makes absolutely no sense to me here

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -193,3 +198,36 @@ void RealtimeEffectStateUI::OnClose(wxCloseEvent & evt)
...
+
+void RealtimeEffectStateUI::
+ProjectWindowSubscriber::InterceptCloseFrame(wxCloseEvent & evt)
+{
+   // Issue 4062, step 1:
+   // Intercept the closing of the project window once only; remove handler
+   auto next = GetNextHandler();
+   mState.mpProjectWindow->RemoveEventHandler(this);
+   mState.mpProjectWindow = nullptr;
+   if (true) {
```
crsib: Please remove this `if`

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -193,3 +198,36 @@ void RealtimeEffectStateUI::OnClose(wxCloseEvent & evt)
...
+}
+
+BEGIN_EVENT_TABLE(RealtimeEffectStateUI::ProjectWindowSubscriber, wxEvtHandler)
+   EVT_CLOSE(RealtimeEffectStateUI::ProjectWindowSubscriber::InterceptCloseFrame)
+END_EVENT_TABLE()
+
+void RealtimeEffectStateUI::
+ProjectWindowSubscriber::InterceptCloseFrame(wxCloseEvent & evt)
+{
+   // Issue 4062, step 1:
```
crsib, line 273: While I understand why `mpProjectWindow` is expected to be valid here, I would preffer to see a check or assertion before the first use

src/effects/RealtimeEffectStateUI.cpp
```diff
@@ -86,6 +99,22 @@ void RealtimeEffectStateUI::Show(AudacityProject& project)
...
+   // must happen after the LV2 validator is destroyed, and that must be
+   // done in idle time, outside of any yield to the event loop
+   mValidatorDestroyedSubscription =
+      mEffectUIHost->Subscribe([this](auto){
+         if (mDoClose) {
+            mDoClose = false;
+            // Closing the project during the destructor of another dialog makes
+            // me nervous -- so defer this closing event, too
+            if (mpProjectWindow)
+               DelayedCloseEvent(*mpProjectWindow, mCanVeto);
```
crsib, line 113: So chances are that several Close events are queued. Have you checked that this is safe? Should this be reflected in a message to QA?

PR #4091, opened 2022-12-21
Update Polish Translation - 3.2.3

Gootector:
Update Polish Translation - 3.2.3

Greetings,
Gootector

Recommended:

### DIFF
--- a/locale/pl.po
+++ b/locale/pl.po
@@ -6085,9 +6085,7 @@ msgstr "Publiczne"
 
 #: src/cloud/audiocom/ShareAudioDialog.cpp
 msgid "Anyone will be able to listen to this audio."
-msgstr ""
-"Każdy będzie mógł słuchać tego nagrania.\n"
-"\n"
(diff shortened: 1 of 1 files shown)


locale/pl.po
```diff
@@ -6083,23 +6081,21 @@ msgstr "&OK"
...
 msgid "Public"
-msgstr ""
+msgstr "Publiczne"
 
 #: src/cloud/audiocom/ShareAudioDialog.cpp
-#, fuzzy
 msgid "Anyone will be able to listen to this audio."
 msgstr ""
-"Nie będzie można odtwarzać/nagrywać dźwięku.\n"
+"Każdy będzie mógł słuchać tego nagrania.\n"
```
LWinterberg, line 6089: It seems like this is causing the compilaiton error: 

> D:/a/audacity/audacity/locale/pl.po:6088: 'msgid' and 'msgstr' entries do not both end with '\n'

Gootector: Done.

Paul-Licameli (approved):

Issue #3702, opened 2022-09-27
Convert Distortion to stateless


pietro68:

dozzzzer:
Verified

PR #4095, opened 2022-12-21
Make distortion stateless

pietro68:
Resolves: #3702

Recommended:

### DIFF
--- a/src/effects/Distortion.cpp
+++ b/src/effects/Distortion.cpp
@@ -22,7 +22,7 @@
 
 #include <cmath>
 #include <algorithm>
-#define _USE_MATH_DEFINES
+//#define _USE_MATH_DEFINES
 
 // Belt and braces
 #ifndef M_PI
@@ -67,7 +67,7 @@ const EffectParameterMethods& EffectDistortion::Parameters() const
       [](EffectDistortion &e, EffectSettings &settings, EffectDistortionSettings &p,
          bool updating) {
          if (!updating)
-            e.mThreshold = DB_TO_LINEAR(p.mThreshold_dB);
+            e.mMaster.mThreshold = DB_TO_LINEAR(p.mThreshold_dB);
          return true;
       }
    };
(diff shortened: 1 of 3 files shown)


src/effects/Distortion.h
```diff
@@ -239,20 +250,33 @@ class EffectDistortion final : public StatefulPerTrackEffect
...
+static constexpr EffectParameter NoiseFloor
+{ &EffectDistortionSettings::mNoiseFloor, L"Noise Floor",
+   EffectDistortionSettings::mDefaultNoiseFloor ,  -80.0,   -20.0,                1    };
+
+static constexpr EffectParameter Param1
+{ &EffectDistortionSettings::mParam1, L"Parameter 1",
+   EffectDistortionSettings::mDefaultParam1,    0.0,   100.0,                1    };
+
+static constexpr EffectParameter Param2
+{ &EffectDistortionSettings::mParam2, L"Parameter 2",    50.0,    0.0,   100.0,                1    };
```
Paul-Licameli, line 168: Replace 50.0 with the named constant `mDefaultParam2` consistently with the other parameters

src/effects/Distortion.cpp
```diff
@@ -277,17 +377,29 @@ OptionalMessage EffectDistortion::DoLoadFactoryPreset(int id)
...
+   // TODO
+//   auto& settings = access.Get();
+//   auto& myEffSettings = GetSettings(settings);
+   auto& myEffSettings = mSettings;
+   auto result = std::make_unique<Validator>(*this, access, myEffSettings);
+   result->PopulateOrExchange(S);
+   return result;
+}
+
+// STEP TODO: add all links to OnSomething handlers
```
Paul-Licameli: I don't understand the comment

src/effects/Distortion.cpp
```diff
@@ -296,14 +408,18 @@ std::unique_ptr<EffectUIValidator> EffectDistortion::PopulateOrExchange(
    {
       S.StartMultiColumn(4, wxCENTER);
       {
-         mTypeChoiceCtrl = S.Id(ID_Type)
+         mTypeChoiceCtrl = S
```
Paul-Licameli, line 488: ID_TYPE and other constants become unused, so may be removed from the header

Paul-Licameli: Oh, I see they are still used in `UpdateUIControls`.  That made me fear a bug -- that there was code still finding controls by id.  But no, there is a switch on ID that then just uses the pointers to controls saved in the validator.

So it is OK

Paul-Licameli: But ID_Type at least can go

src/effects/Distortion.cpp
```diff
@@ -133,27 +133,128 @@ const ComponentInterfaceSymbol EffectDistortion::Symbol
...
+
+
+bool EffectDistortion::Validator::ValidateUI()
+{
+   {
+      // This section was copied from the original
+      // EffectDistortion::TransferDataFromWindow
+      //
+      // However, the call to ->Validate would bring up an error dialog
+      // saying "Empty value"
```
Paul-Licameli, line 222: I'm not satisfied that this is not explained

Paul-Licameli: Maybe that can be debugged later

src/effects/Distortion.h
```diff
@@ -90,30 +95,24 @@ class EffectDistortion final : public StatefulPerTrackEffect
...
-      override;
-   bool RealtimeInitialize(EffectSettings &settings, double sampleRate)
-      override;
-   bool RealtimeAddProcessor(EffectSettings& settings, EffectOutputs *pOutputs,
-      unsigned numChannels, float sampleRate) override;
-   bool RealtimeFinalize(EffectSettings &settings) noexcept override;
-   size_t RealtimeProcess(size_t group,  EffectSettings &settings,
-      const float *const *inbuf, float *const *outbuf, size_t numSamples)
-      override;
+   // Need this dummy method, else it won't build
```
Paul-Licameli: But at this commit, as with BassTreble, you can change the effect base class to `PerTrackEffect` and also change the #include above.  Then this is not needed

src/effects/Distortion.cpp
```diff
@@ -66,8 +66,7 @@ const EffectParameterMethods& EffectDistortion::Parameters() const
    > parameters{
       [](EffectDistortion &e, EffectSettings &settings, EffectDistortionSettings &p,
          bool updating) {
-         if (!updating)
-            e.mMaster.mThreshold = DB_TO_LINEAR(p.mThreshold_dB);
+
```
Paul-Licameli, line 69: Now that the lambda's body shrinks to nothing, just remove it.  `parameters` needs no constructor arguments at all.

src/effects/Distortion.cpp
```diff
@@ -138,8 +137,10 @@ struct EffectDistortion::Validator
    : EffectUIValidator
 {
    Validator(EffectUIClientInterface& effect,
+             EffectDistortion::Instance& instance,
       EffectSettingsAccess& access, EffectDistortionSettings& settings)
       : EffectUIValidator{ effect, access }
+       , mInstance(instance)
```
Paul-Licameli, line 143: indentation

src/effects/Distortion.cpp
```diff
@@ -189,8 +189,8 @@ EffectType EffectDistortion::GetType() const
 auto EffectDistortion::RealtimeSupport() const -> RealtimeSince
 {
    // TODO reenable after achieving statelessness
```
Paul-Licameli, line 371: You can remove this TODO comment

Paul-Licameli (approved):
My suggestions for improvement are all minor.

src/effects/Distortion.h
pietro68: good catch, done

src/effects/Distortion.cpp
pietro68: It was deleted later, just a reminder for me to do something and I forgot it there.

src/effects/Distortion.cpp
pietro68: ok, removed

src/effects/Distortion.cpp
pietro68: ok, done

src/effects/Distortion.cpp
pietro68: ok

Issue #2196, opened 2021-11-28
Clip names lost when pasting between projects


SteveDaulton:
**Describe the bug**
Pasting clips from one project to another may cause clip names in the target project to be lost.

**To Reproduce**
1. Generate a Chirp
2. Split the chirp into two sections
3. Name the two clips "A" and "B"
4. Create a new project
5. Generate a tone
6. Split the tone into two clips
7. Name the clips "1" and "2"
8. Select all of the track from the first project and copy it
9. Paste the copied audio to the start of the second project
10. Observe that the names of the clips in the second project are now: "A", "B", "Audio Track", "Audio Track 1".

**Expected behavior**
At step 10, the clips names are expected to be "A", "B", "1", "2".

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Reported on Windows10. Confirmed on Linux.
 - Version Audacity 3.1.2 and 3.1.3 alpha

**Additional context**
Reported here: https://forum.audacityteam.org/viewtopic.php?p=440425#p440425

petersampsonaudacity:
Works fine for me on W10 with 3.1.2

![image](https://user-images.githubusercontent.com/50205984/143902777-046138fd-3ccc-4b58-860f-8c158b6841f4.png)

SteveDaulton:
> Works fine for me on W10 with 3.1.2
Sorry if it wasn't clear, but step 9 should be:

9. Paste the copied audio into the track that contains clips "1" and "2", so that you (expect to) get 4 clips in one track:  "A", "B", "1", "2".

Is that clearer?

petersampsonaudacity:
> > Works fine for me on W10 with 3.1.2
> 
> Sorry if it wasn't clear, but step 9 should be:
> 
>     9. Paste the copied audio into the track that contains clips "1" and "2", so that you (expect to) get 4 clips in one track:  "A", "B", "1", "2".
> 
> 
> Is that clearer?
 Still works for me on w10 with 3.1.2

![image](https://user-images.githubusercontent.com/50205984/144081515-9ae31bd6-0b27-43fd-9977-d8e64256f6cf.png)

petersampsonaudacity:
OIC - further testing shows the corruption if you paste in at T=0 i.e. before the existing audio in the second project

![image](https://user-images.githubusercontent.com/50205984/144082092-4eb7aab0-f0f3-4dff-be48-acdbba00e11e.png)

petersampsonaudacity:
[...]

PR #4099, opened 2022-12-22
Fix clip rename when pasting when "...move other clips" option enabled

vsverchinsky:
Resolves: #2196 

Fixes issue when original clip names are lost due to name transformations applied at the end of a `WaveTrack::Paste`

Recommended:

### DIFF
--- a/src/WaveTrack.cpp
+++ b/src/WaveTrack.cpp
@@ -1389,8 +1389,12 @@ void WaveTrack::SyncLockAdjust(double oldT1, double newT1)
       {
          // Check if clips can move
          if (EditClipsCanMove.Read()) {
-            auto tmp = Cut (oldT1, GetEndTime() + 1.0/GetRate());
-            Paste(newT1, tmp.get());
+            const auto offset = newT1 - oldT1;
+            for(const auto& clip : mClips)
+            {
+               if (clip->GetPlayStartTime() > oldT1 - (1.0 / mRate))
+                  clip->Offset(offset);
+            }
          }
          return;
       }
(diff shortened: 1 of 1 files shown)


SteveDaulton:
It works for me.
I can't comment on code correctness.

src/WaveTrack.cpp
```diff
@@ -1389,8 +1389,12 @@ void WaveTrack::SyncLockAdjust(double oldT1, double newT1)
       {
          // Check if clips can move
          if (EditClipsCanMove.Read()) {
-            auto tmp = Cut (oldT1, GetEndTime() + 1.0/GetRate());
-            Paste(newT1, tmp.get());
+            const auto offset = newT1 - oldT1;
+            for(const auto& clip : mClips)
+            {
+               if (clip->GetPlayStartTime() > oldT1 - (1.0 / mRate))
```
Paul-Licameli, line 1395: What is the reason for the correction by a sample duration?
Might it be better to substract one-half of that amount?

src/WaveTrack.cpp
```diff
@@ -1457,21 +1461,16 @@ void WaveTrack::PasteWaveTrack(double t0, const WaveTrack* other)
...
-            {
-                if (clip->GetPlayStartTime() > t0 - (1.0 / mRate))
-                    clip->Offset(insertDuration);
-            }
+        //else if there is a clip at t0 insert new clip inside it and ...
+       
+        // ... move everything to the right
+        for (const auto& clip : mClips)
+        {
+            if (clip->GetPlayStartTime() > t0 - (1.0 / mRate))
```
Paul-Licameli, line 1472: I could ask the same here.

But, I see this only preserves old logic, whether that treats the boundary condition right or wrong.

Maybe there is a bug to discover here, but not one introduced by this commit.

Paul-Licameli (approved):
Approved but with some questions about a possible old bug, not introduced by this PR

src/WaveTrack.cpp
vsverchinsky: Yes, correct, that should be one-half
Thanks!

vsverchinsky: Or, not... it is possible that there could be 1 sample-length clip that has a beginning at `oldT1 - 1/rate`, and at such clip should not be moved (this is guaranteed by `>`), but considering that clips aren't bound to be aligned to a track's sample grid, it is actually possible that some clip starts somewhere between `oldT1 - 1/rate` and `oldT1 - 0.5/rate` and such clip should be adjusted

Issue #3877, opened 2022-10-29
Weird behaviour in smart clips when changing track rate


SteveDaulton:
### Bug description

Changing the track sample rate may produce incorrect results when the track contains smart clips.

### Steps to reproduce

1. With the Project Rate set to 44100 (default), generate a 10 second tone at 1000 Hz.
2. Select the first 5 seconds
3. "Ctrl + D" (Duplicate)
4. Select the second track
5. From the track's context menu, change the track rate to 88200 Hz.

### Expected behavior

The second track to show 2.5 seconds of the tone with a frequency of 2000 Hz.

### Actual behavior

The clip in the second track disappears!

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Linux

nyanpasu64:
Worse yet Audacity can *crash* after changing sampling rate with clips:

- Open [bass-h23.aup3.zip](https://github.com/audacity/audacity/files/9915096/bass-h23.aup3.zip)
- Change the track's rate to 22050 or greater
- Play from the beginning

```
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff58a16b3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007ffff5851958 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff583b53d in __GI_abort () at abort.c:79
#4  0x00007ffff5ad2112 in std::__glibcxx_assert_fail(char const*, int, char const*, char const*) (file=<optimized out>, line=<optimized out>, function=<optimized out>, condition=<optimized out>) at /usr/src/debug/gcc/libstdc++-v3/src/c++11/debug.cc:60
#5  0x0000555555e20cb0 in Sequence::Get(int, char*, sampleFormat, sampleCount, unsigned long, bool) const ()
#6  0x0000555555e20d83 in Sequence::Get(char*, sampleFormat, sampleCount, unsigned long, bool) const ()
#7  0x0000555555ecf4a0 in WaveTrack::Get(char*, sampleFormat, sampleCount, unsigned long, fillFormat, bool, sampleCount*) const ()
[...]

PR #4104, opened 2022-12-23
Fix audio clip trim values after rate change

vsverchinsky:
Resolves: #3877

Recommended:

### DIFF
--- a/src/WaveClip.cpp
+++ b/src/WaveClip.cpp
@@ -731,7 +731,11 @@ void WaveClip::CloseLock()
 
 void WaveClip::SetRate(int rate)
 {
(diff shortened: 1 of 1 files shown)


src/WaveClip.cpp
```diff
@@ -763,7 +763,12 @@ void WaveClip::CloseLock()
 
 void WaveClip::SetRate(int rate)
 {
+   const auto ratio = mRate / static_cast<double>(rate);
```
Paul-Licameli, line 766: Unused variable

Paul-Licameli (approved):
Just one unused variable to delete

PR #4120, opened 2022-12-27
Fixes wrong speaker arrangement used for aux input buses

vsverchinsky:
Aux buses aren't supported yet, set speaker arrangement to kEmpty for both input and output aux bus types

Recommended:


crsib (approved):

AbdullahOB (approved):

dozzzzer:
Verified

Issue #3932, opened 2022-11-14
Classic Filters should be in the "EQ and Filters" section of the Effect menu


petersampsonaudacity:
### Bug description
**Classic Filters** should really belong in the "**EQ and Filters**" section of the **Effect** menu

Since 3.2.0  **Classic Filters** is now auto-enabled by default 

### Steps to reproduce
1. get some audio and select it
2. Effect menu
3. Observe: Classic filters sitting by itself at the bottom of the list and not in any functional grouping

### Expected behavior
![image](https://user-images.githubusercontent.com/50205984/201685090-ad53bcdd-1a1c-410d-ac78-0293634c40b8.png)

### Actual behavior
Classic filters sits by itself at the bottom of the list and not in any functional grouping.

### Audacity Version
3.2.1, 3.2.2 beta and 3.3.0 alpha

### Operating system
Windows 10

### Additional context
Flagging @Tantacrul as this is a design/UX issue

petersampsonaudacity:
Note that if you sort the effects by **Effect Name** then **Classic Filters** appears above the line along with the other shipped Audacity effects:
![image](https://user-images.githubusercontent.com/50205984/201685851-a78911a9-ac06-4566-aa91-433cc607c5bd.png)

petersampsonaudacity:
See also:   Classic Filters is now Disabled in 3.2.2 beta and 3.3.0 alpha #3972

harivamsi9:
hi, I would like to know if I could work on this issue?

Tantacrul:
> hi, I would like to know if I could work on this issue?

You have our blessing! Just let us know when you've figured it out and we'll check it :)

Thanks!

Ariel-M:
@Tantacrul Could I also try fixing this? This seems like a great first issue for getting started with contributing to the project.

LWinterberg:
https://github.com/audacity/audacity/blob/master/resources/EffectsMenuDefaults.xml is the file i think

petersampsonaudacity:
Tested on W10 with latest 3.3.0 alpha:  audacity-win-3.3.0-alpha-20221228+de3f25e-x64-msvc2022

works fine with:
a) a simulated virgin install of 3.3.0
b) an upgrade to 3.3.0 from 3.2.3
And it is properly **off** by default in both use-cases.

![image](https://user-images.githubusercontent.com/50205984/209933860-e90d5fa3-b15b-4c09-ae55-8b2273a6a4fb.png)


thanks @Ariel-M

PR #4122, opened 2022-12-27
added Classic Filters to EQ and Filters menu

Ariel-M:
Resolves: https://github.com/audacity/audacity/issues/3932

added Classic Filters to EQ and Filters menu

Recommended:


LWinterberg (approved):
changes understood and tested

Issue #4131, opened 2022-12-29
Add a Delete button to the Cut/Copy/Paste toolbar


petersampsonaudacity:
### Your idea

The new Cut/Copy/Paste toolbar toolbar has been implemented with an empty bottom right cell.

This is not following the original Figma design spec, by designer @DilsonsPickles 
>Here's the Figma link to the design specification: https://www.figma.com/file/uYdHVW7Hhkha51ZGGfC3RF/09_WX---Tools?node-id=328%3A4323
>
>Or [click here](https://www.figma.com/file/uYdHVW7Hhkha51ZGGfC3RF/09_WX---Tools?node-id=328%3A4323)

It also looks a tad ugly and unfinished with the empty cell.

It may have been removed die to a belief that **Cut** and **Delete** do the same thing making the Delete superfluous.  This is _**not**_ the case as Cut replaces the contents of the clipboard whereas Delete does not - and the user may want the carefully placed prior contents of the clipboard to be retained there.

### Problem to be solved

**As implemented in 3.2.3**
![image](https://user-images.githubusercontent.com/50205984/209966212-ddc0fb9f-b1e9-4d34-a39f-f6b76b311957.png)

**As specified by @DilsonsPickles in the Figma design** 
![image](https://user-images.githubusercontent.com/50205984/209966571-5e6aa167-3aaf-4324-bcbe-fb5b89903a3b.png)

### Prior art

**Paint3D**
Here's an example from Microsoft's Paint3D which has a similar toolbar - and that does have a Delete button in addition to Cut, Copy and Paste.

![image](https://user-images.githubusercontent.com/50205984/209966037-5b337206-9ee6-4404-ba7a-a9ec14808dd8.png)

### Additional context

See also:   Restore the Cut, Copy and Paste buttons as a new (hidden by default) toolbar #3693

abhi-kr-2100:
Is there a reason why one needs to request before viewing the design specification at https://www.figma.com/file/uYdHVW7Hhkha51ZGGfC3RF/09_WX---Tools?node-id=328%3A4323?
![image](https://user-images.githubusercontent.com/66527624/210158021-e7dcb0d7-65fd-47e2-a014-ec269c9ebfe3.png)

petersampsonaudacity:
Testing on W10 with @abhi-kr-2100 's   branch build:  audacity-win-3.3.0-alpha-20230101+c23548e-x64-msvc2022

This looks good to me in this branch build:

[...]

PR #4133, opened 2023-01-01
Add a delete button to the Cut/Copy/Paste toolbar

abhi-kr-2100:
Signed-off-by: Abhishek Kumar <abhi.kr.2100@gmail.com>

Resolves: #4131 

Add a delete button to the Cut/Copy/Paste toolbar

Recommended:


abhi-kr-2100:
While .png icons for cut, copy and paste exist inside [libraries/lib-theme-resources/classic/Components](https://github.com/audacity/audacity/tree/master/libraries/lib-theme-resources/classic/Components), `Delete.png` and `DeleteDisabled.png` need to be added.

LWinterberg:
The source for these icons can be found at https://github.com/musescore/MuseScore/blob/master/fonts/mscore/MusescoreIcon.ttf,  is the glyph in question (it's inside a private use area and won't render outside the musescore icon font)

you'll need to color them manually in an image editor though. Be aware that the four themes (classic, light, dark, high contrast) each need their own versions.

petersampsonaudacity:
Testing on W10 with @abhi-kr-2100 's   branch build:  audacity-win-3.3.0-alpha-20230101+c23548e-x64-msvc2022

This looks good to me in this branch build:
a) The Delete button does indeed delete the selection - and properly does _**not**_ replace the clipboard contents
b) the Cut button deletes the selection - and properly **_does_** replace the clipboard contents
c) the Delete button icon is as @DilsonsPickles specified in the Figma spec

Tantacrul:
I have one suggestion. When there is no audio selected, the delete button should be disabled.
image

There is no reason why we should need to show this message:
image

I'm aware that when you use a shortcut to delete, you see the above message. At some other point, we'll remove the message in this instance too.

Tantacrul:
One other little comment, although I'm not sure if it is relevant to this PR or not.

When I first opened this, the **cut, copy, paste** toolbar was moved to a second row under the main toolbar. Resetting my configuration fixed this. I only mention it to make sure we don't accidentally revert our changes that make the **cut, copy, paste** toolbar hidden by default, and when one does enable it, that it appears in the right place.

Thanks @abhi-kr-2100 !

petersampsonaudacity:
> I have one suggestion. When there is no audio selected, the delete button should be disabled.

@Tantacrul   Same then should also be true for Cut and Copy when there is nothing selected - and for Copy when there is nothing on the clipboard.

1. clear Audacity settings folder
2. launch the branch build
3. enable the Cut/Copy/Paste toolbar
4. Observe: the Paste button is bold/enabled - but there is nothing on the clipboard
5. import some audio
6. Observe: now the Cut, Copy and Delete buttons are bold/enabled too - though nothing is selected

BTW this erroneous button behavior is a regression introduced somewhere between 2.0.0 and 2.4.2 _(I don't have time to bisect right now, sorry - will do so later)._

crsib:
Enabled "Paste" is an older bug of Audacity. It was not fixed.

petersampsonaudacity:
> When I first opened this, the **cut, copy, paste** toolbar was moved to a second row under the main toolbar.

I don't see that

1) when I pseudo-upgrade from 3.2.2 to this branch and then enable Cut/Copy/Paste I see:
image


2) If I pseudo virgin install this branch and then enable Cut/Copy/Paste I see:
image

petersampsonaudacity:
> Enabled "Paste" is an older bug of Audacity. It was not fixed.

@crsib Is it a logged bug Dmitry?

crsib:
I have no idea. But the toolbar buttons reflect the state of the "Edit" menu.

Tantacrul:
> Enabled "Paste" is an older bug of Audacity. It was not fixed.

@petersampsonaudacity - is it a bug? I saw you mention it a week or so ago and I forgot to respond. The problem you identified, IIRC, was that if you copied text on your clipboard somewhere else, then the Paste button is active.

I'm wondering if that is wrong, bearing in mind you could paste that into a label?

petersampsonaudacity:
> > Enabled "Paste" is an older bug of Audacity. It was not fixed.
> 
> @petersampsonaudacity - is it a bug? I saw you mention it a week or so ago and I forgot to respond. The problem you identified, IIRC, was that if you copied text on your clipboard somewhere else, then the Paste button is active.

I no longer think this is entirely true

1) If I reboot my PC (so there is no data on my Windows clipboard) and then launch Audacity i observe that the Paste button is active/enabled even though there is no data to post.

2) The other thing to bear in mind is that there are two clipboards
a) the Windows clipboard that normally contains alphanumeric text
b) the Audacity clipboard that contains chunks of audio track or chunks of label track

The Paste command and button is a little odd in that it can:
i) Paste chunks of audio track  from the Audacity clipboard into audio tracks
ii) Paste chunks of label track  from the Audacity clipboard into label tracks
iii) but _(surprisingly, to me anyway)_ paste alphanumeric text from the Windows clipboard into a label

In contrast to iii) what the Paste command cannot do is to paste alphanumeric text from the Windows clipboard into other alphanumeric fields in Audacity:
iv) Directories prefs
v) Keyboard prefs
vi) Edit > Metadata
vii) Import and Export dialogs
For all these only the Windows shortcut **Ctrl+V** will work 

As with the Copy and Cut buttons issue (and the commands, as @crsib has pointed out) the Paste button also appears to be a regression created somewhere between 2.0.0 and 2.4.2.  I will try to find some time later today to test in which version the regression occurred and then I will log the bug.

@Tantacrul @petersampsonaudacity 
Meanwhile since this residual issue also affects Cut, Copy and Paste too. maybe we could just allow this pull request to be pulled into master for 3.3.0 - and deal with the residual for all buttons as a separate logged issue?

_That would enable me to then get on with updating the Reference Manual._

petersampsonaudacity:
Argh  -  now I'm totally mindblown ...

If I clear my Audacity settings folder and then launch @abhi-kr-2100 's branch and enable the Cut/Copy/Paste toolbar
I now see the Cut, Copy and Delete buttons alll grayed out

Tantacrul:
I think you might be right that if this PR is not introducing issues, then we should fix them separately. I'll leave it up to @crsib to determine if this is the case. Thanks @petersampsonaudacity !

ksoze95 (approved):

PR #4141, opened 2023-01-03
Parallelize the Equalization effect

Paul-Licameli:
When you EQ your hour long podcast, your extra cores should not be idle!

(This extends the previously reviewed but still unmerged PR #4016.)

Enjoy this present 🎅 

Recommended:

### DIFF
--- a/src/effects/Equalization.cpp
+++ b/src/effects/Equalization.cpp
@@ -41,6 +41,7 @@
 
 #include "WaveClip.h"
 #include "WaveTrack.h"
+#include <thread>
 
 const EffectParameterMethods& EffectEqualization::Parameters() const
 {
@@ -571,17 +572,45 @@ bool EffectEqualization::ProcessOne(int count, WaveTrack * t,
       return bLoopSuccess.load(order);
    });
 
+   // Divide among available processors, with a little oversubscription
+   // (If there is blocking on database operations, let some other thread resume
+   // calculations)
+   const auto nProcessors = std::thread::hardware_concurrency();
+   auto nTasks = nProcessors + 2;
+   const auto nBlocks = ((len + idealBlockLen - 1) / idealBlockLen);
+   if (nTasks > nBlocks)
+      nTasks = std::max<unsigned>(1, nBlocks.as_size_t());
    std::vector<EqualizationTask> tasks;
-   tasks.reserve(1);
-   tasks.emplace_back(mParameters, M, idealBlockLen, *t, poller, true, true);
+   tasks.reserve(nTasks);
+
+   // First task
+   tasks.emplace_back(
+      mParameters, M, idealBlockLen, *t, poller, true, nTasks-- == 1);
+   // More tasks
+   while (nTasks)
+      tasks.emplace_back(
+         mParameters, M, idealBlockLen, *t, poller, false, nTasks-- == 1);
+   nTasks = tasks.size();
+
    t->ConvertToSampleFormat( floatSample );
 
    std::vector<TaskRunner> runners;
-   runners.reserve(tasks.size());
+   runners.reserve(nTasks);
+
+   size_t ii = 0;
+   auto taskStart = start,
+      nextStart =
+         start + std::min(len, idealBlockLen * (nBlocks * ++ii / nTasks));
    for (auto &task : tasks) {
       auto &runner = runners.emplace_back(task, bLoopSuccess);
-      runner(start, len);
+      // If nBlocks/nProcessors has a fraction, this length calculation rounds
+      // sometimes up and sometimes down
+      runner(taskStart, nextStart - taskStart);
+      taskStart = nextStart;
+      nextStart =
+         start + std::min(len, idealBlockLen * (nBlocks * ++ii / nTasks));
    }
+   assert(nextStart == start + len);
    for (auto &runner : runners) {
       if (runner.pExc)
          std::rethrow_exception(runner.pExc);
--- a/src/effects/Equalization.cpp
+++ b/src/effects/Equalization.cpp
@@ -411,6 +411,9 @@ bool EffectEqualization::TransferDataToWindow(const EffectSettings &settings)
 }
 
 namespace {
+//! Function takes an increment of samples processed, returns true to continue
+using Poller = std::function<bool(sampleCount)>;
+
 struct EqualizationTask {
    EqualizationTask( size_t M, size_t idealBlockLen, WaveTrack &t )
       // Capacity for M - 1 additional right tail samples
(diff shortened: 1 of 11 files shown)


src/effects/Equalization.cpp
```diff
@@ -449,11 +459,16 @@ struct EqualizationTask {
...
    float *lastWindow{ window2.get() };
 
+   const WaveTrack &input;
+   Poller &poller;
+
    // create a NEW WaveTrack to hold all of the output,
    // including 'tails' each end
    std::shared_ptr<WaveTrack> output;
 
+   size_t M;
```
crsib, line 503: I understand that you are restructuring the old code, but I think that a comment would be very welcome here

crsib: I meant a comment in the source code, not on GitHub

crsib (approved):
I was quite surprised to observe, that UI thread is still used for the processing. 

No need to change it now, and it doesn't really the performance (I've checked). 

However, as a rule of thumb, we should not do any lengthy operations blocking UI thread.

Paul-Licameli:
> I was quite surprised to observe, that UI thread is still used for the processing.
> 
> No need to change it now, and it doesn't really the performance (I've checked).
> 
> However, as a rule of thumb, we should not do any lengthy operations blocking UI thread.

In the sequential version, the UI thread does everything.

In a parallel version, the UI thread still needs to poll the progress dialog for user cancellation.

If it has no other work to do, then it would need to sleep in a loop to avoid busy-waiting.  Or can you think of an alternative?

If the UI thread then sleeps to let other threads crunch numbers -- then to get polling of the button, wouldn't we need the UI thread no only to sleep and wake up, but also evict some other thread from a processor, only to test the button, sleep again, and impose another context switch to resume the worker?

For these reasons I thought it might actually be simpler and marginally better performance to leave it as I did it.  Correct me if I misunderstand.

Paul-Licameli:
I will soon rebase and force-push this onto more recent master.

I will add a commit with a comment explaining "M".

I must resolve some conflicts in WaveClip and Sequence at the commit
"Growing WaveTrack can retain a minimum buffer in memory until Flush()"

which resulted from the other project about dithering.

Paul-Licameli:
That is now done.  Nothing else changed.

But please notice, the last two commits were only unsuccessful guesses I made to try to get a better speedup for @petersampsonaudacity .

The commit using a splin lock is probably a good idea anyway, but I'm not confident the last one should be included.

Still an unresolved is what more intelligent choice of number of threads we should use.

dozzzzer:
Testing this PR, I've discovered some inaccuracies in calculations, which require further investigation.

In order to observe those, you'll need to do the following:

1. Generate a 30s  white noise clip and save it
2. Using any build of Audacity that doesn't have the parallelization EQ changes, import the noise clip
3. Apply **Graphic EQ** with certain parameters to it
4. Save the project as project A
5. Using the build from this PR (https://github.com/audacity/audacity/pull/4141), import the noise clip 
6. Save the current project as project B
7. Repeat step 2
8. Open up project A and copy the audio clip from it
9. Paste it into project B
10. Now that you have two clips in project B, select one of them and apply "Invert" to it
11. Cmd + A > Mix and Render
12. Switch the scale from Linear to dB
13. Zoom in the resulting audio clip at the very end of it

As a result, you'll get smth like this:

image

The portion of the clip of around 3800 samples long, which will amount to smth around -74 dB. 
The rest of the track, except this small area, shows **-INF dB** meaning that the two clips - the one produced by the application of the sequential EQ effect and the one made by the new parallelized EQ - are perfectly identical and cancel each other when one of them is inverted.

petersampsonaudacity:
@dozzzzer  

Are you exporting WAV or MP3?

If MP3 you would expect that as all MP3 exports pad a little at the beginning of the file.

petersampsonaudacity:
Testing on W10 with 3.2.4 versus @Paul-Licameli  Paul's latest branch build for this: audacity-win-3.3.0-alpha-20230207+1ea47ed-x64

Test project is 2 0m 44s of stereo audio recorded from a webcast _(i.e. a real-life typical large data-set for me)_

1)  Exporting as **MP3 256kbps constant Bitrate Stereo** using the Audacity command **File>Export>Export as MP3**
I get a slight speed improvement:
2.4.3    2min:06sec
Paul's   1min:52sec

2) Exporting as **WAV 16-bit PCM** I get a slightly better improvement
2.4.3    1min:03sec
Paul's   0min:35sec

3) The two exported MP3s are identical
    tested by importing both, iinverting one and mixing - result is a pure flat line

crsib:
> Are you exporting WAV or MP3?

I think @dozzzzer saved the project, no export was involved. Even exporting to WAV is a potentially destructive operation. 

With your second comment I feel that you are commenting in the wrong thread - this PR is about EQ effect, it is not related to Paul's latest efforts on exporting performance.

petersampsonaudacity:
>With your second comment I feel that you are commenting in the wrong thread - this PR is about EQ effect, it is not related to Paul's latest efforts on exporting performance.

@crsib   OOPS, sorry Dmitry  _<blushes deeply>_

I have transferred my export tests that Paul requested to the correct PR thread

Paul-Licameli:
I spoke with @dozzzzer about this.

I recommended going to Interface preferences and choosing the lowest value for cutoff of the dB scale, which made this inaudible anomaly at the end visible in waveform display.

I don't know why it is there and I should figure it out, but it is reassuring that all else in the long track was exactly zero.

@petersampsonaudacity as I also pointed out to @dozzzzer -- for these testing purposes, there should not be any lossy export, or saves to sample format other than float.  Otherwise quantization and dither noise would obscure what is happening.

src/effects/Equalization.cpp
```diff
@@ -411,40 +416,121 @@ bool EffectEqualization::TransferDataToWindow(const EffectSettings &settings)
...
 namespace {
-struct EqualizationTask {
-   EqualizationTask( size_t M, size_t idealBlockLen, WaveTrack &t )
-      : buffer{ idealBlockLen }
+//! Function takes an increment of samples processed, returns true to continue
+using Poller = std::function<bool(sampleCount)>;
+
+class EffectTask {
+public:
+   virtual ~EffectTask();
```
saintmatthieu, line 424: Why not directly here `virtual ~EffectTask() = default;` ?

Paul-Licameli: Virtual functions need to have non-inline bodies compiled somewhere.  The compiler might generate them even if you make them all inline, but then it has to eliminate duplicates at link time.

If you make at least one virtual function non-inline, the compiler can follow the rule that it generates the rest of the functions only in the same translation unit and avoid the extra work of removing the duplicates.  So I do this with destructors to be a little nicer to the compiler.

Though sometimes, a template is generating the class, and then you often must inline them anyway and rely on the elimination.

saintmatthieu (approved):
I really concentrated on the Equalization itself. Did a few tests, LGTM, to the exception of a few thingies I spot. Please consider addressing before merging.

dozzzzer:
Testing this PR now shows the issue remains present. I'm consistently getting about the same result as previously:

image

PR #4149, opened 2023-01-05
Scatter various menu items

Paul-Licameli:
Resolves: 

Some dependencies that block some future module extractions are broken;
to do that, more menu items are registered.

So this is mostly just boring cutting and pasting.

Recommended:


src/Printing.cpp
```diff
@@ -191,3 +195,39 @@ void HandlePrint(
...
+{
+   auto &project = context.project;
+   auto name = project.GetProjectName();
+   auto &tracks = TrackList::Get( project );
+   auto &window = GetProjectFrame( project );
+   HandlePrint(&window, name, &tracks, TrackPanel::Get( project ));
+}
+
+using namespace MenuTable;
+BaseItemSharedPtr PrintingItems()
```
ksoze95, line 216: I think could be made static

src/toolbars/TranscriptionToolBar.cpp
```diff
@@ -1065,3 +1065,110 @@ AttachedToolBarMenuItem sAttachment{
...
+      Menu( wxT("PlayAtSpeed"), XXO("&Play-at-Speed") ) };
+   return menu;
+}
+
+AttachedItem sAttachment2{
+   wxT("Optional/Extra/Part1"),
+   Shared( ExtraPlayAtSpeedMenu() )
+};
+
+BaseItemSharedPtr ExtraPlayAtSpeedItems()
```
ksoze95, line 1147: This submenu has moved from Extra->Play-At-Speed to now Extra->Play-At-Speed->Play-At-Speed.  Is this intentional?

Paul-Licameli: Actually this is not a mistake:  the sub-menu, and then the items within the sub-menu, are registered separatedly, each with a correct path, and when I run the program, the menu is correct.

The code might be simpler with one registration, but it is not wrong.

ksoze95: Apologies, it looks like that I might have cut and paste (with a bit of further fixup) incorrectly applied the code transformations for the commit on my local machine.

ksoze95 (approved):

Issue #4151, opened 2023-01-05
Turning an effect on and off doesn't make the project dirty


dozzzzer:
### Steps to reproduce

1. add an effect in the sidebar
2. save the project 
3. turn the effect on/off by clicking the blue power button
4. ctrl + w

### Expected behavior

The prompt to save the changes in the project pops up

### Actual behavior

The project is closed with no warning

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

Paul-Licameli:
#4153 is an easy fix for the problem as stated -- but if you press a power button twice for no net change, then the prompt to save when closing will also appear, though unnecessarily.

Should extra work be done to correct that?

dozzzzer:
> Should extra work be done to correct that?

I'd say it's not necessary

dozzzzer:
@Paul-Licameli 
the fix looks good to me, however, in the case when turning an effect on/off by opening its UI first, and then clicking the blue power button, it doesn't make the project dirty. We also need to fix that.

dozzzzer:
Tested

PR #4153, opened 2023-01-06
Power buttons dirty the project (even if only restoring prior state)

Paul-Licameli:
Resolves: #4151

A simple change so that pressing the power button of an effect list or a
single effect causes the nag to save the project when you close the window.

But this also causes the warning without need if you press the button twice,
toggling back to the original state.

Recommended:

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -358,6 +358,8 @@ namespace
          enableButton->Bind(wxEVT_BUTTON, [this](wxCommandEvent&) {
 
             mEffectState->SetActive(mEnableButton->IsDown());
+            if (mProject)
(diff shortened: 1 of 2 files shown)


ksoze95 (dismissed):
Looks good to me

Paul-Licameli:
Reviewing the conversation at the linked issue:  @dozzzzer told us why the fix wan't wrong, but incomplete.

So now another commit completes it.

locale/fr.po
saintmatthieu, line 1: I don't expect this file's changes to be intentional ? If not please revert before merge.

saintmatthieu (approved):

locale/fr.po
Paul-Licameli: I forgot to change the base branch of the PR to release.

crsib (approved):

Issue #4147, opened 2023-01-04
Better default settings dor Noise Reduction effect: 6, 6. 6


petersampsonaudacity:
### NR default
The default settings in Audacity for Noise Reduction are 12, 6, 3
![image](https://user-images.githubusercontent.com/50205984/210622295-6fdaf91b-b573-4735-81f8-b1ef787ba2e4.png)

### Suggested improved default
Greg Kozikowski (Koz), a long-standing Forum elf and ex Audacity Team member, often recommends 6, 6 ,6 as a better alternative
![image](https://user-images.githubusercontent.com/50205984/210622755-598c804a-12ff-4045-9862-f85139dd36f1.png)
 

### Problem to be solved

Koz is a professional sound engineer in Hollywood so I trust his recommendation in this regard very highly.

I have used Koz' settings many time to remove FM hiss from radio broadcasts and got excellent results.

Paul-Licameli:
"Noise Reduction of the Beast" !

But I thought what is appropriate might depend on one's recording setup.

This is a recommendation for voiceover work in particular?

petersampsonaudacity:
> This is a recommendation for voiceover work in particular?

I have recommended it to users on your FB Audacity Voicover group Paul - and they seemed satisfied.

Personally I've used it for FM music broadcasts.

KennyMacheka:
I'd like to take this up if no one else is?

RayAsh37:
Hi @petersampsonaudacity , @LWinterberg 
I have made the required changes and submitted a PR. Please review

petersampsonaudacity:
BTW 
**IF** the user makes changes to the NR presets in an earlier version and later in an earlier version **_manually_** resets the settings to 12,6,3
**THEN** when this branch is launched Audacity does _**not**_ overwrite the 12,6,3 with the new 6,6,6 default

_So presumably Audacity has some hidden cunning way of knowing the that the default has never been overwritten._

RayAsh37:
I believe it creates files storing all the user settings. Some other softwares that I have worked with (eg. Blender) have it where I can export the file and then install a new version and then read the file from the new version. But doing it manually is a bit cumbersome, so the auto-detection is pretty convenient.

petersampsonaudacity:
@LWinterberg  do I assume correctly, Leo, that this will also need a code review before it can be pulled into master ?

dozzzzer:
Verified

petersampsonaudacity:
[...]

PR #4155, opened 2023-01-07
updated default values for noise reduction effect

RayAsh37:
Made the required changes as mentioned in issue  #4147 to update the values from 12, 6, 3 to 6, 6, 6

#4147 

This is the requested change as mentioned in the Issue to change the default values for the Noise reduction effect

Recommended:

### DIFF
--- a/src/effects/NoiseReduction.cpp
+++ b/src/effects/NoiseReduction.cpp
@@ -525,10 +525,10 @@ bool EffectNoiseReduction::Settings::PrefsIO(bool read)
 
    static const PrefsTableEntry<Settings, double> doubleTable[] = {
          { &Settings::mNewSensitivity, wxT("Sensitivity"), 6.0 },
-         { &Settings::mNoiseGain, wxT("Gain"), 12.0 },
+         { &Settings::mNoiseGain, wxT("Gain"), 6.0 },
          { &Settings::mAttackTime, wxT("AttackTime"), 0.02 },
          { &Settings::mReleaseTime, wxT("ReleaseTime"), 0.10 },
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
Testing on W10 with @RayAsh37 's branch build:

1)  This works fine for a Virgin install of this (3.3.0) branch build.

2) When upgrading from an earlier version 
a) any settings that the user made and (expected to keep) in the earlier version use of Noise Reduction are properly _**not overwritten**_  by the new 6, 6, 6 default.
b) However 
**IF**  the user has not made any changes to the default settings in Noise Reduction when using it in earlier Audacity versions
**THEN** Audacity overwrites the old 12, 6, 3 default with the new 6, 6, 6  default
_Both a) and b) are quite cunning, as users who have always used Audacity's default will now get Audacity's new default - but users who have carefully changed settings to suit them will get to retain their previous settings._

So this looks functionally good to me.


@crsib @Paul-Licameli   @LWinterberg  do I assume correctly that this will also need a code review before it can be pulled into master ?

Paul-Licameli:
The code change is trivial.

It only needs a decision whether this is right to do.

Do we want our QA to do ear testing to decide whether they agree with the change?  Not merely to verify that the settings do change as specified.

Paul-Licameli (approved):

dozzzzer:
Verified

LWinterberg:
After a bit more testing and research, I found the softer approach of 6,6,6 slightly beneficial. Especially when ramping the reduction amount, the wider smoothing felt better, particularly on spoken content. I wasn't able to form a strong opinion either way for musical content.

Overall, I think 6,6,6 is a somewhat "safer" starting point for noise reduction than 12,6,3 is.

Issue #1099, opened 2021-06-25
Extract lib-tags


Paul-Licameli:
Extract a toolkit neutral library for audio file metadata.  Needed for import and export.

PR #4161, opened 2023-01-09
Extract lib-tags

Paul-Licameli:
Resolves: #1099

Extract a library for audio file metadata, which is used in import and export

Recommended:


crsib (approved):

Paul-Licameli:
Force pushed just to squash the two commits and rebase onto recent master; no other changes

dozzzzer:
Verified

Issue #3866, opened 2022-10-27
Effect grouping preferences should be reflected in the RT effects panel


dozzzzer:
### Your idea

Currently, whatever option you select for grouping the effects, it's only reflected in the main Effect menu. While in the RT effects panel, the effects are always grouped by author, no matter what is selected in the Preferences.

In order to remain consistent and ease up the navigation among the plugins, especially in cases when there're multiple different families of the same effect installed in the system, we need to reflect the grouping option set in Preferences > Effects menu in both places, including the RT effects panel.

<img width="776" alt="image" src="https://user-images.githubusercontent.com/70063970/198310664-08516fb0-512f-4399-abdc-1e374f1349b5.png">

### Problem to be solved

Troublesome navigation through the list of the RT effects.

petersampsonaudacity:
Surely this is a duplicate of:   Effect grouping preferences not reflected in realtime effects menu #3311

Ryan-Miller-Audacity:
During testing I am finding that it is taking a full close/reopen of Audacity for the preference change to update in the realtime effects menu. The change is applied to the destructive effects menu instantly. Seems this needs a little more refinement to work exactly correct.

dozzzzer:
@vsverchinsky 
When changing **Realtime effect organization** option to "Group by category" in Preferences, in the RT effects panel appear some effects that aren't supposed to be there, such as: Delay, Limiter, etc.

I believe this issue is coupled with another one: https://github.com/audacity/audacity/issues/3423

Having this issue discoverable in yet another place now - RT effects panel - I think it needs to be addressed first.

dozzzzer:
> When changing **Realtime effect organization** option to "Group by category" in Preferences, in the RT effects panel appear some effects that aren't supposed to be there,

This is now fixed in PR #4171

dozzzzer:
@vsverchinsky  
Also, testing PR #4171 shows that the list of publishers isn't sorted by name (A>Z):

<img width="429" alt="image" src="https://user-images.githubusercontent.com/70063970/222419362-f2e4f0d5-ab14-47af-a8c1-463e6599cad3.png">

dozzzzer:
Tested

PR #4171, opened 2023-01-11
Realtime effects menu grouping

vsverchinsky:
Resolves: #3866

Recommended:


Paul-Licameli (changes requested):
There are link failues on Windows to fix before merge, and there are conflicts in PluginMenus.cpp to resolve in rebasing

src/ListNavigationEnabled.h
```diff
@@ -0,0 +1,115 @@
...
+#include <wx/containr.h>
+
+/**
+ * \brief Changes default arrow navigation to behave more list- or table-like.
+ * Instead of searching focusable items among children first, list navigation
+ * searches for siblings when arrow key is pressed. Tab behaviour stays same.
+ * Requires wxWANT_CHARS style flag to be set
+ */
+template<class WindowBase>
+class ListNavigationEnabled : public wxNavigationEnabled<WindowBase>
```
Paul-Licameli, line 22: If I wrote this, I would do the extra work to define a .cpp file, so that (1) at least one .cpp file in the system includes the header file before any other, guaranteeing that the header is self-sufficient for compilation; and (2) making out-of-line functions doing most of the work that doesn't depend on the template type parameter, leaving only small template functions in the header to be inlined.

See for instance Observer, which separates an out-of-line, "type erased" part.

I don't insist that you follow the suggestion for this PR.

Paul-Licameli: It would also make sense to put this file in src/wrappers.  The resembles `wxPanelWrapper` for instance.  But I don't insist.  Library extractions might soon upset the folder organization anyway.  Although a library for src/wrappers would definitely be a non-toolkit-neutral part of the program until we figure out better abstractions.

src/CMakeLists.txt
```diff
@@ -170,6 +170,8 @@ list( APPEND SOURCES
       Legacy.cpp
       Legacy.h
       ListNavigationEnabled.h
+      ListNavigationPanel.cpp
+      ListNavigationPanel.h
```
Paul-Licameli, line 174: Ditto, suggestion about folder placement.  I don't insist.

src/CMakeLists.txt
```diff
@@ -186,6 +186,8 @@ list( APPEND SOURCES
       MixAndRender.h
       MixerBoard.cpp
       MixerBoard.h
+      MovableControl.cpp
+      MovableControl.h
```
Paul-Licameli, line 190: And this file.

But whatever the folders they are in,  it is good to see RealtimeEffectPanel shrink to its essentials.

src/RealtimeEffectPanel.cpp
```diff
@@ -138,52 +139,6 @@ namespace
       }
    };
 
-   //Alias for ListNavigationEnabled<wxWindow> which provides wxWidgets-style ctor
```
Paul-Licameli, line 242: This comment got lost in the cut and paste.

Paul-Licameli: And if restored, it should also be more complete, describing what the event handlers accomplish.

src/menus/MenuHelper.cpp
```diff
@@ -0,0 +1,782 @@
+#include "MenuHelper.h"
+#include "PluginManager.h"
+#include "effects/EffectManager.h"
+#include "CommonCommandFlags.h"
+#include "prefs/EffectsPrefs.h"
+#include "BatchCommands.h"
+#include "BatchProcessDialog.h"
```
Paul-Licameli, line 7: Do all these #include directives compile on all platforms?  I thought ../ would be needed in some of them.

src/CMakeLists.txt
```diff
@@ -769,6 +769,8 @@ list( APPEND SOURCES
       menus/FileMenus.cpp
       menus/HelpMenus.cpp
       menus/LabelMenus.cpp
+      menus/MenuHelper.cpp
+      menus/MenuHelper.h
```
Paul-Licameli, line 773: Maybe a more suggestive name?  PluginMenuBuilder?

libraries/lib-module-manager/PluginManager.cpp
```diff
@@ -525,6 +525,7 @@ bool PluginManager::DropFile(const wxString &fileName)
                   mRegisteredPlugins[id].SetEnabled(enable);
                // Make changes to enabled status persist:
                this->Save();
+               this->NotifyPluginsChanged();
```
Paul-Licameli, line 528: Not reachable with default Experimental.cmake but it's good you noticed it

Paul-Licameli: I see one use of Save() not followed by the publication, and that is in `PluginManager::InitializePlugins()`.  Is that right or wrong?

Paul-Licameli: Not wrong, I see, because the only subcriber is RealtimeEffectListWindow which cannot possibly be constructed earlier than that initialization.

src/RealtimeEffectPanel.cpp
```diff
@@ -730,8 +733,85 @@ class RealtimeEffectListWindow : public wxScrolledWindow
...
+      mPluginsChangedSubscription = PluginManager::Get().Subscribe([this](PluginsChangedMessage) { UpdateEffectMenuItems(); });
+      UpdateEffectMenuItems();
    }
 
+   std::optional<wxString> PickEffect(wxWindow* parent, const wxString& selectedEffectID)
+   {
+      wxMenu menu;
+      if(!selectedEffectID.empty())
+      {
+         //no need to handle language change since menu creates it's own event loop
```
Paul-Licameli, line 746: "its" not it's

src/RealtimeEffectPanel.cpp
```diff
@@ -822,6 +902,7 @@ class RealtimeEffectListWindow : public wxScrolledWindow
    void ResetTrack()
    {
       mEffectListItemMovedSubscription.Reset();
+      mPluginsChangedSubscription.Reset();
```
Paul-Licameli, line 905: The constructor makes subscriptions.

`ResetTrack()` destroys both subscriptions.

`SetTrack()` restores the old subscription to non-null, but not the new one.

Is that right or wrong?

src/RealtimeEffectPanel.cpp
```diff
@@ -47,13 +47,79 @@
...
+#include "Menus.h"
 
 #if wxUSE_ACCESSIBILITY
 #include "widgets/WindowAccessible.h"
 #endif
 
 namespace
 {
+   
+   class RealtimeEffectsMenuVisitor final : public MenuVisitor
```
Paul-Licameli, line 60: I am glad to see you figured out the usage of MenuVisitor!  I hope it did not seem unnecessarily difficult.  If you gained some more understanding of Registry.h and Registry.cpp that is good -- they need some re-work which will need careful review by someone.

src/RealtimeEffectPanel.cpp
```diff
@@ -730,8 +733,85 @@ class RealtimeEffectListWindow : public wxScrolledWindow
...
+   {
+      auto root = std::make_unique<MenuTable::MenuItem>(Identifier {}, TranslatableString {});
+      
+      auto analyzeItems = MenuHelper::PopulateEffectsMenu(
+         EffectTypeAnalyze,
+         {}, {},
+         nullptr,
+         [](const PluginDescriptor& desc) {
+            return desc.IsEffectRealtime();
+         }
```
Paul-Licameli, line 785: This repeated lambda might be a local variable defined once

src/RealtimeEffectPanel.cpp
```diff
@@ -730,8 +733,85 @@ class RealtimeEffectListWindow : public wxScrolledWindow
...
+               XO("Analyze"),
+               std::move(analyzeItems)
+            )
+         );
+         root->items.push_back(std::move(analyzeSection));
+      }
+   
+      auto processItems = MenuHelper::PopulateEffectsMenu(
+         EffectTypeProcess,
+         {}, {},
```
Paul-Licameli, line 803: I checked the function again to see what these two {}s are -- and I discovered, `realFlags` is passed around among many functions in MenuHelper.cpp but never used!

I don't know where in source code history it became unused -- but you have an opportunity to simplify with one more commit.

Paul-Licameli:
I have finished a round of review.

The Windows builds are still failing, so you must fix that.

The question above about `SetTrack` and `ResetTrack` are the only unresolved doubt I have.

Otherwise all else is just suggestion and so this is *almost* approved.

Paul-Licameli:
The windows build failure is because of a duplicate symbol:

```
wxmsw31u_core.lib(wxmsw313u_core_vc_custom.dll) : error LNK2005: "public: virtual __thiscall wxNavigationEnabled<class wxWindow>::~wxNavigationEnabled<class wxWindow>(void)" (??1?$wxNavigationEnabled@VwxWindow@@@@UAE@XZ) already defined in ListNavigationPanel.obj
```

... which I think you can fix just with one more `extern template` declaration at the second commit.

src/menus/MenuHelper.cpp
vsverchinsky: I believe so, yes. If I remember correctly `src` folder is added to include directories somewhere in cmake

src/CMakeLists.txt
vsverchinsky: It used to provide macros menu helpers too. And it doesn't build menus

src/RealtimeEffectPanel.cpp
vsverchinsky: `mPluginsChangedSubscription.Reset()` shouldn't be here, thanks!

src/RealtimeEffectPanel.cpp
vsverchinsky: Had to spend some time to figure out how it works, yes :)

src/RealtimeEffectPanel.cpp
vsverchinsky: Yep

src/RealtimeEffectPanel.cpp
```diff
@@ -1109,8 +733,88 @@ class RealtimeEffectListWindow : public wxScrolledWindow
...
+   }
+   
+   void UpdateEffectMenuItems()
+   {
+      auto root = std::make_unique<MenuTable::MenuItem>(Identifier {}, TranslatableString {});
+
+      static auto realtimeEffectPredicate = [](const PluginDescriptor& desc)
+      {
+         return desc.IsEffectRealtime();
+      };
```
Paul-Licameli, line 786: You made the suggested simplification to remove duplication.

Amusingly:  you could even simplify this more to:

```
      static auto realtimeEffectPredicate = &PluginDescriptor::IsEffectRealtime;
```

The implicit constructor of `std::function` used when you call `MenuHelper::PopulateEffectsMenu` works even when the template parameter is of type pointer-to-memner-function.

Paul-Licameli (approved):

vsverchinsky:
Branch was rebased and few new commits were added, starting from [Effects preferences strings change](https://github.com/audacity/audacity/pull/4171/commits/537cecaaeac188f2d514e73ecd12394d9dde180e)

src/menus/PluginMenus.cpp
```diff
@@ -385,7 +384,6 @@ BaseItemSharedPtr EffectMenu()
          { return Items( wxEmptyString, MenuHelper::PopulateEffectsMenu(
             EffectTypeProcess,
             AudioIONotBusyFlag() | TimeSelectedFlag() | WaveTracksSelectedFlag(),
-            IsRealtimeNotActiveFlag(),
```
Paul-Licameli: The definition of `IsRealtimeNotActiveFlag` is in this file.  It becomes unused, so may be deleted too.

src/ThemedWrappers.h
```diff
@@ -17,22 +17,28 @@
...
 #include "Theme.h"
 #include "Prefs.h"
 
+struct PrefsListenerDummy { };
+
 template<class WindowBase>
 class ThemedWindowWrapper final :
    public WindowBase,
-   public PrefsListener
+   public std::conditional_t<std::is_base_of_v<PrefsListener, WindowBase>, PrefsListenerDummy, PrefsListener>
```
Paul-Licameli, line 33: Maybe this is more elegant:

```
template<class WindowBase, class... OtherBases>
class ThemedWindowWrapper final :
   public WindowBase, public OtherBases...,
```

So you can simply use `PrefsListener` where needed as another template argument, and not mention that class name here.

src/ThemedWrappers.h
```diff
@@ -63,7 +69,7 @@ class ThemedWindowWrapper final :
 
    void UpdatePrefs() override
    {
-      PrefsListener::UpdatePrefs();
+      PrefsListenerBase::UpdatePrefs();
```
Paul-Licameli, line 72: Maybe you could use composition instead of inheritance:  a member of a type inheriting PrefsListener.  Then you would not need to play with type traits to pass through to the base.  Instead UpdatePrefs is directly invoked directly on this member and on the base class to do different unrelated things that can happen in either sequence.

src/menus/MenuHelper.cpp
```diff
@@ -156,48 +148,25 @@ void AddEffectMenuItemGroup(
          groupCnt--;
```
Paul-Licameli, line 148: This variable is not longer used, so remove it too

src/prefs/EffectsPrefs.cpp
```diff
@@ -124,6 +127,15 @@ void EffectsPrefs::PopulateOrExchange(ShuttleGui & S)
    }
    S.EndStatic();
 #endif
+
+   S.AddButton(XO("Open Plugin Manager"), wxALIGN_LEFT)->Bind(wxEVT_BUTTON, [this](auto) {
+      //Adding dependency on PluginRegistrationDialog, not good. Alternatively
```
Paul-Licameli, line 132: A wise comment.  I like it that you are aware of dependencies you add.

But I checked that this does not add a dependency cycle.

True, it may still be an undesirable dependency, but we can figure out some trick to break it if we need that later.

src/ThemedWrappers.h
Paul-Licameli: ... or maybe, even without composition, all you need is:
```
if constexpr(std::is_base_of_v<PrefsListener, WindowBase>)
   WindowBase::UpdatePrefs();
```

Paul-Licameli (approved):
I caught up at last with the new version of `AddGroupedEffectMenuItems` and I'm finished with this review.

Paul-Licameli (approved):

vsverchinsky:
> I note that the default setttings for the Effect Menu and the Realtime effects are different. Is this intentional, by design, or are they supposed to be the same by default?

Yes, they are different by design

PR #4173, opened 2023-01-12
Extract lib audio io

Paul-Licameli:
Resolves: #1136 

GUI Toolkit-neutral library for the audio engine; and a subset of that, just for the realtime effect management, can
be turned into another small library that it depends on.

Recommended:


Paul-Licameli:
A force push only to add, at the proper commit in the sequence, a missing #include

Paul-Licameli:
Sorry for the force push.  Only minimal changes to rebase onto latest master, resolve a simple conflict in include directories of AudioIO.cpp, and fix the Windows link with visibility of the structure `RealtimeEffectManager::AllListsLock`

src/AudioIO.h
```diff
@@ -76,12 +80,12 @@ struct AudioIOEvent {
 };
 
 struct TransportTracks {
-   WaveTrackArray playbackTracks;
-   WaveTrackArray captureTracks;
+   WritableSampleTrackArray playbackTracks;
```
crsib: I wonder why playback tracks should be writable

crsib (approved):
A very welcome PR! Please fix the link failure before merging.

src/AudioIO.h
Paul-Licameli: Later commit addressed that, yes?

crsib: Yes!

PR #4176, opened 2023-01-12
Log some statistics about the times when ProgressDialog was shown

crsib:
This is a minor change, that logs the time an operation with a ProgressDialog took.

Recommended:

### DIFF
--- a/src/widgets/ProgressDialog.cpp
+++ b/src/widgets/ProgressDialog.cpp
@@ -1043,6 +1043,13 @@ ProgressDialog::~ProgressDialog()
    if (mHadFocus && SearchForWindow(wxTopLevelWindows, mHadFocus)) {
       mHadFocus->SetFocus();
    }
+
+   using namespace std::chrono;
+   wxLogInfo(
+      "Operation '%s' took in %f seconds. Poll was called %d times and took %f seconds. Yield was called %d times.",
+      GetTitle(), mElapsedTime / 1000.0, mPollsCount,
+      duration_cast<duration<double>>(mPollTime).count(),
+      mYieldsCount);
 }
 
 void ProgressDialog::Init()
@@ -1073,6 +1080,7 @@ void ProgressDialog::Reinit()
    mStartTime = wxGetUTCTimeMillis().GetValue();
    mLastUpdate = mStartTime;
    mYieldTimer = mStartTime;
+   mElapsedTime = 0;
    mCancel = false;
    mStop = false;
 
@@ -1092,6 +1100,10 @@ void ProgressDialog::Reinit()
       button->Enable();
 
    wxDialogWrapper::Show(true);
+
+   mPollTime = {};
+   mPollsCount = {};
+   mYieldsCount = {};
 }
 
 void ProgressDialog::SetDialogTitle(const TranslatableString& title)
@@ -1328,6 +1340,13 @@ bool ProgressDialog::Create(const TranslatableString & title,
 ProgressResult ProgressDialog::Update(
    int value, const TranslatableString & message)
 {
+   using namespace std::chrono;
+   auto updatePollTime = finally([this, pollStart = high_resolution_clock::now()] {
+         mPollTime += high_resolution_clock::now() - pollStart;
+   });
+
+   mPollsCount++;
+   
    if (mCancel)
    {
       // for compatibility with old Update, that returned false on cancel
@@ -1339,9 +1358,9 @@ ProgressResult ProgressDialog::Update(
    }
 
    wxLongLong_t now = wxGetUTCTimeMillis().GetValue();
-   wxLongLong_t elapsed = now - mStartTime;
+   mElapsedTime = now - mStartTime;
 
-   if (elapsed < 500)
+   if (mElapsedTime < 500)
    {
       return ProgressResult::Success;
    }
@@ -1362,7 +1381,7 @@ ProgressResult ProgressDialog::Update(
       value = 1000;
    }
 
-   wxLongLong_t estimate = elapsed * 1000ll / value;
+   wxLongLong_t estimate = mElapsedTime * 1000ll / value;
    wxLongLong_t remains = (estimate + mStartTime) - now;
 
    SetMessage(message);
@@ -1378,7 +1397,7 @@ ProgressResult ProgressDialog::Update(
    if ((now - mLastUpdate > 1000) || (value == 1000))
    {
       if (m_bShowElapsedTime) {
-         wxTimeSpan tsElapsed(0, 0, 0, elapsed);
+         wxTimeSpan tsElapsed(0, 0, 0, mElapsedTime);
          mElapsed->SetLabel(tsElapsed.Format(wxT("%H:%M:%S")));
          mElapsed->SetName(mElapsed->GetLabel()); // fix for bug 577 (NVDA/Narrator screen readers do not read static text in dialogs)
          mElapsed->Update();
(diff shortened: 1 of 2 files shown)


src/widgets/ProgressDialog.cpp
```diff
@@ -1047,6 +1049,13 @@ ProgressDialog::~ProgressDialog()
    if (mHadFocus && SearchForWindow(wxTopLevelWindows, mHadFocus)) {
       mHadFocus->SetFocus();
    }
+
+   using namespace std::chrono;
+   wxLogInfo(
+      "Operation '%s' took in %f seconds. Poll was called %d times and took %f seconds. Yield was called %d times.",
```
Paul-Licameli, line 1055: Either say "took" or "finished in"

crsib: Shoot, my bad, changed the text sometime in the middle xD

It was `completed` initially, by I felt that this is not quite right, because this message will be shown for cancelled tasks as well

src/widgets/ProgressDialog.cpp
```diff
@@ -1413,6 +1434,7 @@ ProgressResult ProgressDialog::Update(
    // Nyquist effects call Update on every callback, but YieldFor is
    // quite slow on Linux / Mac, so don't call too frequently. (bug 1575)
    if ((now - mYieldTimer > 50) || (value == 1000)) {
+      mYieldsCount++;
```
Paul-Licameli, line 1437: Perhaps the total duration of yield time, within polling time, could be accumulated too?

Paul-Licameli (approved):
We have a willing performance tester of parallel EQ in @petersampsonaudacity .

Peter has his ways of timing things with macros, but this enhancement could help us get more precise information.

Can you instruct Peter in how to gather these numbers?

src/widgets/ProgressDialog.cpp
crsib: I think it will match precisely the Poll time, but this can be easily done (and might be quite educative!)

petersampsonaudacity:
> Peter has his ways of timing things with macros, but this enhancement could help us get more precise information.

@Paul-Licameli  more accurately @lllucius  (Leland) provided the tools (macro trace) in version 3.0.0 - he even made me a special 2.4.2 build post-release so that I could test AUP versus AUP3 timings.

So the timing tools are there in Audacity for anyone to use - I think it is an easter egg and not documented though ...

Paul-Licameli:
I think @crsib wants to try reducing the frequency of polling of the progress indicator dialog for button presses as one way to get even better performance, and this new change will provide more information about whether that is so.

petersampsonaudacity:
> We have a willing performance tester of parallel EQ in @petersampsonaudacity .

**_reasonably willing_** I should say (time permitting) - don't forget I'm not on the payroll, just a willing volunteer  ;-)

Paul-Licameli:
> > We have a willing performance tester of parallel EQ in @petersampsonaudacity .
> 
> **_reasonably willing_** I should say (time permitting) - don't forget I'm not on the payroll, just a willing volunteer ;-)

@petersampsonaudacity @crsib I've been thinking it over, and I have a different theory of how we might further improve performance for Peter's 2-core machine or even my 6 core machine.  (4 vs 12 logical processors, with hyper threading.)

I think the yielding of the main thread to the event loop has little effect on performance, contrary to Dmitry's speculations.

Rather we should try to saturate the disk utilization, because it is the contended resource that is the sequential bottleneck that sets the limit on the speedup that we can realize.

EQ was compute-bound when sequential.  More threads make it less compute-bound.  Still the disk isn't saturated, which means the task subdivision is still too coarse.  (Even on my MacBook, where I did get a several-fold speedup, not just tens of %.  Evidence described below.)

Oversubscription is in general a good idea -- more threads than available logical processors.  I simply added 2 to `std::thread::hardware_concurrency()` but that was just a poor guess.

Windows Performance Monitor or the Xcode debugger, even attached to a release build, can yield interesting data.  With Xcode I see the highest spike in disk utilization AFTER the EQ is done, which I think is the background thread doing checkpointing.  During the EQ calculation, disk utilization still didn't go that high in my branch.  Suggesting, time is still spent with all threads computing, not reading or writing.  So my oversubscription just wasn't enough!

Maybe multiplying by 2 would be better!

Maybe adding a control to the effect dialog, just in alpha builds, to vary the number of threads, would enable experiments to gather data to make a more intellligent decision about how much oversubscription is best.

Beyond some point I would expect adding threads would give diminishing and then even negative returns.

petersampsonaudacity:
It's a shame we don't have any **_usage monitoring_** to let the devs know which effects and other functions are most-used and in particular most-used on large projects.  ;-)

My key ones are Amplify and Export with my 2-2.25 hour weekly projects.

crsib:
> contrary to Dmitry's speculations.

Measurements shown that my speculations were wrong. That's why we do measurements!

> Rather we should try to saturate the disk utilization

I don't agree with this. I don't think that our goal is to "freeze" the user PC. What is also missing here is that you have a very fast NVMe drive in your MacBook. I can easily get 100% utilization of my HDD for extended periods of times simply by using Audacity. 

> Oversubscription is in general a good idea 

It is not. Measurement is. Lets dive in a bit. 

I have and 8C/16T Ryzen 5800x paired with a high end X570 chipset. I have tested with 3 different drives:

1. NVMe 2 Tb 980 Pro 
2. SATA3 SSD 4 Tb 870 Evo
3. SATA3 HDD 6 Tb IronWolf

On disk 1 I observe little under 4x improvement when 18 threads are used. Roughly  8 secs is reduced to 2 secs with 2 hour chirp. Using 6 threads yields the very similar result. Using 4 threads is a little bit slower. Speedup looks very good, but clearly - adding more threads probably won't really help. However, I have noticed that disk kicks in almost at the end of the calculation and is used intensively (even **before** checkpointing) after the dialog is closed. It appears, that we close the dialog **before** the result is written to the disk.

So I have generated 6 hour-long chirp. For this case I have only compared 18 threads vs 1 thread and I have measured roughly how long does takes to settle down all writes **after** the disk has started writing. This **includes** checkpointing, There is an impact on operations performed during the checkpointing. I wonder, why current approach for checkpointing was chosen instead of a write queue of some kind.

| Scenario     | CPU Time (s) | Disk time (s) |
| --- | :---: | :---: |
| Disk 1/18T | 21 | 36 |
| Disk 1/1T | 41 | 50 |
| Disk 2/18 T | 21 | 40 |
| Disk 2/1T | 41 | 60 |
| Disk 3/18T | 21 | 90 |
| Disk 3/1T | 42 | 95 |

During the processing Disk 1 is heavily underutilized, both in terms of write speed and IOPS. But there is only 2x speedup here, so it appears that we are **not limited** by the disk performance at all. What exactly is limiting the performance is a very good question, we will have to find it out later.  My current speculation is that we are limited by SQLite MT performance. This is just a speculation, likely it is wrong.

For the history, SATA SSD had 50% load on average **during** the processing, HDD was at 100%.

petersampsonaudacity:
> > Rather we should try to saturate the disk utilization
> 
> I don't agree with this. I don't think that our goal is to "freeze" the user PC.

I agree - not least because the user may be using the computer for other work too (with Audacity processing in the background).

_Note that when Dave Cutler (at DEC, Digital Equipment) was designing RSX, the design objective was not for the OS to get itself into a busy state, bur rather it's "desired" state was actually quiescence._

Paul-Licameli (approved):

PR #4179, opened 2023-01-13
Some effect cleanups

Paul-Licameli:

Some reorganization of effect code, before those that have an "Init()" might also become "PerTrackEffects"
(such as EQ and Noise Reduction), and thus possible to put into the sidebar.

UI-related abstract base classes are moved out of the lower-level libraries that should be concerned only with
calculations.

There is still some state in the Effect objects (even in so-called "stateless" effects) related to the overall context
of tracks and open windows.  This begins to eliminate some of that other state.

QA:  please sanity-check:

Rendering of effect stacks
Destructive application of built-ins and Ladspa effects (don't need many variations of settings)

Recommended:


src/effects/Effect.h
```diff
@@ -180,11 +180,13 @@ class AUDACITY_DLL_API Effect /* not final */
 
    static void IncEffectCounter(){ nEffectsDone++;}
 
-   bool EnableApply(bool enable = true);
+   static
+   bool EnableApply(wxWindow *parent, bool enable = true);
```
vsverchinsky, line 184: As a members those functions make sense here, but not much sense as a statics. I'd suggest to move them out from these class completely

vsverchinsky (commented):
Few suggestions, there is a build error that should be fixed

src/effects/Effect.h
```diff
@@ -304,12 +304,9 @@ class AUDACITY_DLL_API Effect /* not final */
    // UI
    //! This smart pointer tracks the lifetime of the dialog
    wxWeakRef<wxDialog> mHostUIDialog;
-   wxWindow       *mUIParent{};
-
-public:
-   wxWindow* GetUIParent() { return mUIParent; }
 
 private:
+   wxWindow       *mUIParent{};
```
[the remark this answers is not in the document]
Paul-Licameli: I want to eliminiate Effect::mUIParent.  This pull request does not complete that.  It does remove direct uses of it in subclasses, which instead shadow it.  Stateless subclasses store it in the Validator object instead of the effect, which is even better, eliminating some state.

To eliminaite it entirely requires rewriting of Effect::MessageBox and its many uses.  That is to be done later.

src/effects/Effect.h
Paul-Licameli: I don't know why it's so bad to make them statics, but as functions related specially to user interface, they might make more sense as static utility members of EffectUIValidator.

Paul-Licameli:
I have rebased onto recent master, which had no conflicts.

I added a commit to fix the Windows build, and another to answer the suggestion to move two static member functions out of Effect.

vsverchinsky (approved):

Ryan-Miller-Audacity:
Completed sanity checks of mix and rendering of effect stacks, exporting audio, and applying built in/Ladspa effects.

Issue #4190, opened 2023-01-16
Cutlines disappear when duplicating track with selection


vsverchinsky:
STR:
- Make sure cutlines option is enabled in preferences
- Generate audio clip
- Create cutline in the middle
- Select short region around the cutline
- Duplicate track (Ctrl+D)

Observe: duplicated track does not contain cutline (or it can be displaced)

![image](https://user-images.githubusercontent.com/20165230/212687841-ee8ec064-6bb3-4811-8ccb-4b49759532d4.png)

Ryan-Miller-Audacity:
Verified cutlines no longer disappear/displace when a track with cutlines is duplicated.

PR #4192, opened 2023-01-16
Preserve all cutlines when making a clip copy

vsverchinsky:
Resolves: #4190 

Recommended:

### DIFF
--- a/src/WaveClip.cpp
+++ b/src/WaveClip.cpp
@@ -100,19 +100,8 @@ WaveClip::WaveClip(const WaveClip& orig,
    mEnvelope = std::make_unique<Envelope>(*orig.mEnvelope);
 
    if ( copyCutlines )
-      // Copy cutline clips that fall in the range
-      for (const auto &ppClip : orig.mCutLines)
-      {
-         const WaveClip* clip = ppClip.get();
-         double cutlinePosition = orig.GetSequenceStartTime() + clip->GetSequenceStartTime();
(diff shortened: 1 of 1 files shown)


src/WaveClip.cpp
```diff
@@ -100,19 +100,8 @@ WaveClip::WaveClip(const WaveClip& orig,
...
-         if (cutlinePosition >= t0 && cutlinePosition <= t1)
-         {
-            auto newCutLine =
-               std::make_unique< WaveClip >( *clip, factory, true );
-            newCutLine->SetSequenceStartTime( cutlinePosition - t0 );
-            mCutLines.push_back(std::move(newCutLine));
-         }
-      }
+      for (const auto &cutline : orig.mCutLines)
+         mCutLines.push_back(std::make_unique<WaveClip>(*cutline, factory, true));
```
Paul-Licameli, line 104: So is it a consequence of this change:

I can make a clip; make a cutline; drag the smart clip boundary to hide the cutline; copy a part of the clip; paste to a new track; drag the boundary of the new clip -- and find the duplicated cutline too?

Paul-Licameli (approved):

src/WaveClip.cpp
vsverchinsky: For some reason cutlines aren't copied with regular copy-paste, but they should be copied when part of a clip that contains a cutline is duplicated. But in general what you said is correct.

PR #4199, opened 2023-01-16
Simplify module graph

Paul-Licameli:

Make fewer edges in the graphviz output in modules.dot.svg by allowing inherited
dependencies on wxBase from any library, or module, or Audacity and removing some
other unnecessary edges.

This will aid understanding as libraries and modules become more numerous.

Color what is not toolkit-neutral red, and what is, green.

Before:
modules dot

After:

modules dot

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/AudacityFunctions.cmake
+++ b/cmake-proxies/cmake-modules/AudacityFunctions.cmake
@@ -344,6 +344,8 @@ function( append_node_attributes var target )
    if( NOT "wxwidgets::wxwidgets" IN_LIST dependencies )
       # Toolkit neutral targets
       set( color "lightgreen" )
+      # Enforce usage of only a subset of wxBase that excludes the event loop
+      apply_wxbase_restrictions( ${target} )
    endif()
    string( APPEND "${var}" " style=filled fillcolor=${color}" )
    set( "${var}" "${${var}}" PARENT_SCOPE)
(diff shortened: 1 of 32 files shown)


cmake-proxies/cmake-modules/AudacityFunctions.cmake
```diff
@@ -338,6 +338,43 @@ function( canonicalize_node_name var node )
    set( "${var}" "${node}" PARENT_SCOPE )
 endfunction()
 
+function( append_node_attributes var target )
+   get_target_property( dependencies ${target} AUDACITY_GRAPH_DEPENDENCIES )
```
crsib, line 342: I wonder if `define_property` is needed. But probably not

crsib (approved):
This PR makes very hard to understand the library dependencies **from the code**. Relying on additional tools does not seem particularly useful to me. TBH, I've never done that personally besides the times you've shown me it during the call.

I don't have very strong opinion on this, so I will approve PR anyway.

Paul-Licameli:
> This PR makes very hard to understand the library dependencies **from the code**. Relying on additional tools does not seem particularly useful to me. TBH, I've never done that personally besides the times you've shown me it during the call.
> 
> I don't have very strong opinion on this, so I will approve PR anyway.

Thank you . Do you also agree to merge it at once without pausing for QA?

crsib:
> Do you also agree to merge it at once without pausing for QA?

Sure!

Issue #3492, opened 2022-08-31
Accessibility name of "power" toggle button in realtime effects dialog is "enable"


DavidBailes:
### Bug description

I'm assuming that these toggle buttons will be referred to as power toggle buttons in the documentation, and so the accessibility name of the toggle button in realtime effect dialogs should be power, rather than enable.

### Steps to reproduce

Open a realtime effects dialog

### Expected behavior

accessibility name of power toggle button should be "power"

### Actual behavior

It's "enable".

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

See also:
https://github.com/audacity/audacity/issues/3489
https://github.com/audacity/audacity/issues/3482

Ryan-Miller-Audacity:
I tested this and confirmed with NVDA screenreader on windows.

Paul-Licameli:
Will require a re-test to merge #4334

Ryan-Miller-Audacity:
Retested with PR #4334.

Issue #3482, opened 2022-08-30
The state of the power toggle buttons in realtime effects dialogs are not accessible for users of screen readers


DavidBailes:
### Bug description

The state of the power toggle buttons in realtime effects dialogs are not accessible for users of any of the screen readers on Windows.

### Steps to reproduce

1. Open a realtime effects dialog.
2.  Tab to the power toggle button.
3.  Press Enter to press the power toggle button.
4.  Tab away from the button and then back to it to read it in its changed state.

### Expected behavior

When a toggle button becomes the focus, the state of the button should always be read. Screen readers on Windows differ in the exact words used, but NVDA reads either pressed or not pressed.

When a toggle button is pressed, the screen reader reads the new state, for example pressed or not pressed.

### Actual behavior

When the power toggle button becomes the focus, all screen readers just read the name of the button - "Enable button"

When the power toggle button is pressed, all of the screen readers are silent.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

The power toggle button replaces an enable check box in previous versions of Audacity. The state of the check box was accessible for users of screen readers, and so this bug is a regression.

See related issue: https://github.com/audacity/audacity/issues/3470

DavidBailes:
@LWinterberg . Since I've submitted a pull request to fix this, can this issue be added to the 3.3 milestone?

Ryan-Miller-Audacity:
@DavidBailes I've tested this on Windows and confirmed that this is now working as the expected behavior describes above; however, my only concern being, the "Master" power button and the "individual" effect power buttons are labeled/read the exact same. I believe this could be confusing to Visually Impaired users. 

Should we change the "Master" effect power button (the power button that controls status for the entire track's effect stack) to read as "Master power button" or something of this sort?

DavidBailes:
> Should we change the "Master" effect power button (the power button that controls status for the entire track's effect stack) to read as "Master effect power button" or something of this sort?

I think "master" might be misunderstood to be something connected with a master bus (although there isn't one yet). An alternative might be "Track effects power button", or maybe just "Track power button", as it's in the realtime effects pane.

Ryan-Miller-Audacity:
I think "Track effects power button" would make sense. I'll move the related tickets for this pull request forward and we can create an alternate ticket for renaming that. As long as @LWinterberg thinks that naming makes sense?

Paul-Licameli:
Will require a re-test to merge #4334

[...]

PR #4202, opened 2023-01-18
Issue #3482: accessibility of power toggle buttons in dialogs

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/3482
Resolves: https://github.com/audacity/audacity/issues/3492

Problem: The state of the power toggle buttons in realtime effects dialogs are not accessible for users of screen readers.

Fix: A wxBitmapButton was previously used for the power toggle button. However it's not possible to imitate a toggle button for all the Windows screen readers using this control. So use an AButton instead. This is already used for the power buttons in the realtime effects panel, and has already been made to imitate a toggle button for all Windows screen readers.

Notes:
- Using an AButton means that it probably won't be easily possible to assign an access key to this button. But this is a minor issue compared to the state of the button not being accessible.
- The accessibility name of the toggle button was changed from Enable to Power to make it consistent with the power toggle buttons in the realtime effects panel.
- In EffectUIHost::InitializeInstance(), a line updating the state of the toggle button was removed, as it was immediately prior to a call to UpdateControls() which updates the state of the toggle button.

Recommended:

### DIFF
--- a/src/ShuttleGui.cpp
+++ b/src/ShuttleGui.cpp
@@ -120,6 +120,7 @@ for registering for changes.
 #include "widgets/ReadOnlyText.h"
 #include "widgets/wxPanelWrapper.h"
 #include "widgets/wxTextCtrlWrapper.h"
+#include "widgets/AButton.h"
 #include "AllThemeResources.h"
 
 #if wxUSE_ACCESSIBILITY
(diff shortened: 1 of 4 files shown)


DavidBailes:
Will resubmit after correcting error

PR #4209, opened 2023-01-19
improve zooming behavior

LWinterberg:
Zooming with the scrollwheel in Audacity got identified as being much too quick, resulting in easy overscrolling into either the sample level or the days-and-weeks level. This PR reduces the factor from 2x to 1.25x for mouse wheel scrolling. 

Partially reverts fc25ce093fcee957f920827fa7e052b35f3c6c72


LWinterberg:
The right-of-waveform check doesn't feel entirely sufficient; I wrote up a spec in #4212 instead

PR #4211, opened 2023-01-20
Extract lib wave track

Paul-Licameli:
Resolves: 

A toolkit-neutral library for the WaveTrack class and supporting classes.

But it still depends on a start-up dependency injection of the implementation
of block files.  That is, SqliteSampleBlock and SqliteSampleBlockFactory are
not in this library.  Only their abstract base classes.

So this library has no dependency on sqlite.

Notes for QA:  should test some of the GUI related to track displays, because this separates display from core logic.

- Display and clicks on clip boundaries and cutlines
- Adjustment of vertical bounds and scale type in vertical spectrogram ruler
- Varying spectrogram settings of individual tracks (NOT the use of the Preferences dialog which affects defaults)
- Waveform view vertical bounds and scale type changes

Recommended:


src/WaveTrack.cpp
```diff
@@ -201,13 +188,8 @@ void WaveTrack::Reinit(const WaveTrack &orig)
...
 
-   {
-      auto &settings = orig.mpWaveformSettings;
-      if (settings)
-         mpWaveformSettings = std::make_unique<WaveformSettings>(*settings);
-      else
-         mpWaveformSettings.reset();
-   }
+   // Copy attached data from orig.  Nullify data in this where orig had null.
+   ((Caches&)*this) = orig;
```
crsib, line 175: Why C-cast?

src/WaveTrack.cpp
```diff
@@ -216,12 +198,10 @@ void WaveTrack::Merge(const Track &orig)
...
       DoSetGain(wt.GetGain());
       DoSetPan(wt.GetPan());
-      mDisplayMin = wt.mDisplayMin;
-      mDisplayMax = wt.mDisplayMax;
       SetSpectrogramSettings(wt.mpSpectrumSettings
          ? std::make_unique<SpectrogramSettings>(*wt.mpSpectrumSettings) : nullptr);
-      SetWaveformSettings
-         (wt.mpWaveformSettings ? std::make_unique<WaveformSettings>(*wt.mpWaveformSettings) : nullptr);
+      // Copy attached data from orig.  Nullify data in this where orig had null.
+      ((Caches&)*this) = *pwt;
```
crsib, line 185: Why C-cast?

src/prefs/WaveformSettings.h
```diff
@@ -81,4 +89,34 @@ class AUDACITY_DLL_API WaveformSettings : public PrefsListener
    // Convenience
    bool isLinear() const { return stLinear == scaleType; }
 };
+
+class AUDACITY_DLL_API WaveformSettingsCache
```
crsib, line 93: Something feels really wrong with this class. Judging by comment - it allows changing application global state. Following the real post effects is really hard here too.

src/prefs/WaveformSettings.h
```diff
@@ -81,4 +89,34 @@ class AUDACITY_DLL_API WaveformSettings : public PrefsListener
...
+
+   ~WaveformSettingsCache() override;
+   PointerType Clone() const override;
+
+   int ZeroLevelYCoordinate(wxRect rect) const;
+
+   void GetDisplayBounds(float &min, float &max) const
+   { min = mDisplayMin; max = mDisplayMax; }
+
+   void SetDisplayBounds(float min, float max) const
```
crsib, line 112: Why setters are const?

src/prefs/WaveformSettings.h
```diff
@@ -81,4 +89,34 @@ class AUDACITY_DLL_API WaveformSettings : public PrefsListener
...
+   { mDisplayMin = min; mDisplayMax = max; }
+
+   float GetLastScaleType() const { return mLastScaleType; }
+
+   mutable float mDisplayMin = -1.0f, mDisplayMax = 1.0f;
```
crsib, line 117: Why the fields are public?

src/prefs/SpectrogramSettings.h
```diff
@@ -195,4 +200,26 @@ class AUDACITY_DLL_API SpectrogramSettings : public PrefsListener
 
 extern AUDACITY_DLL_API IntSetting SpectrumMaxFreq;
 
+class AUDACITY_DLL_API SpectrogramSettingsCache
```
crsib, line 203: All the comments from the previous class apply here. 

BTW, why cache? What does it cache exactly?

src/WaveTrackLocation.cpp
```diff
@@ -13,6 +13,8 @@ Paul Licameli -- split from WaveTrack.h
 #include "WaveTrack.h"
 #include "WaveClip.h"
 
+#include <math.h>
```
crsib, line 16: Please use C++ header: cmath

crsib (changes requested):
There are few places in this PR that need clarification

src/WaveTrack.cpp
Paul-Licameli: Invoke assignment only if the ClientData::Site base class to clone the attachments.  That conserves the effect of the deleted lines.

src/WaveTrack.cpp
Paul-Licameli: Ditto

src/WaveTrack.cpp
crsib: I do not understand what exactly do you mean and how it justifies C-cast.

Paul-Licameli: More explicitly `this->Caches::operator=(orig)` with no cast but I don’t find that prettier.

This is an intentional slicing assignment.

src/prefs/WaveformSettings.h
Paul-Licameli: Which comment do you mea?

This is extra state, but not of the application.  Each WaveTrack has its own independently changeable cache.

src/prefs/WaveformSettings.h
Paul-Licameli: See that the corresponding deleted member functions of WaveTrack were also like that.  The members were mutable.

Mutability makes some sense -- the graphics code wants to decorate a track, that it can examine but not change, with some extra memos only for its own use.

Instead of making those members of WaveTrackCache mutable, I could do this:

```
WaveformSettings &WaveformSettings::Get( const WaveTrack &track )
 {
+   return Get(const_cast<WaveTrack&>(track));
 }
```

-- that is, overload the static function accessing the cache, so it can take a const track but give back a non-const cache.

src/prefs/WaveformSettings.h
Paul-Licameli: No good reason here.  I can change it.

Though I do not agree that whenever there is a structure, you must encapsulate everything with get and set functions.  Some things are simply bags of data with no important consistency constraints, no invariants, to be established and preserved by member functions.  Such for instance is the case with many structure names ending in `...Options` that are just meant to shorten long lists of arguments.

src/prefs/SpectrogramSettings.h
Paul-Licameli: See what member variables are removed from WaveTrack.

They were all mutable and in that sense "caches."  Mutables aren't always nice but it seemed preferable to me before to make all those things mutables, rather than to require non-const tracks as arguments in drawing functions.  Logically, mere drawing should not be able to mutate the data model.  These extra things aren't really part of the data model.

Display and interaction code can take a const wave track that it is not allowed to mutate, but use ClientData to associate extra information with each track for its own needs.

src/WaveTrack.cpp
crsib: I don't find this explanation convincing. This should be solved differently. Definitely not with a C-cast. 

I don't quite understand, why `static_cast` or some form of implicit conversion is not used here. 

I am not going to approve the code in the way it is now done.

src/prefs/WaveformSettings.h
crsib: Specifically this comment is really confusing.  `//! Get either the global default settings, or the track's own if previously created`

src/prefs/WaveformSettings.h
crsib: I have a pretty strong opinion, that mixing POD-like objects and getters/setters is not a best idea. Also in most of the places `m` prefix in Audacity means some kind of encapsulation. It is frustrating to read and understand the code when naming conventions are violated.

src/prefs/SpectrogramSettings.h
crsib: I don't see how making variable mutable makes it a `cache`.

src/WaveTrack.cpp
Paul-Licameli: Okay then, will you be satisfied with:

`static_cast<Caches&>(*this) = orig;` ?

The reason C-cast didn't bother me is that this is an up-cast of the object to its base class, not the checked down-cast which is what `static_cast` is usually for with pointers and references.  In other words this also works with no casts:

```
Caches &caches = *this;
caches = orig;
```

In fact I even like that better.  No casts, just a reference binding.

Paul-Licameli: As to what the line actually DOES, never mind the style matters -- there is no complaint?  It does no more nor less than the deleted lines it replaces.

src/prefs/WaveformSettings.h
Paul-Licameli: That comment may be extraneous, yes.  By mistaken analogy with the Spectogram settings code where that comment makes more sense.

src/prefs/WaveformSettings.h
Paul-Licameli: Arguably this is not a POD because there should be an ordering constraint on mins and maxes.

However I have my equally strong opinion that one should not be slavish about the `m` convention.  Some PODs really are PODs.  A structure with no invariant conditions it needs to protect is just a bunch of stuff and not properly a class.

src/prefs/SpectrogramSettings.h
Paul-Licameli: "memo" then?  Not in the sense of memo-izing an expensive computation, but it is some piece of associated attached data.  Or, if this helps, I guess in LISP world it would be something like a property list entry that you can stick onto a symbol.

"attachment"?

"decoration"?

other suggestions welcome

src/prefs/WaveformSettings.h
Paul-Licameli: The fact that something does NOT use the `m` conventions should be its own recognized convention too.  That, this structure is meant to be just a POD.

src/WaveTrack.cpp
crsib: Implicit conversion, as with two lines you proposed, works just fine for me.

src/prefs/WaveformSettings.h
crsib: > Arguably this is not a POD because there should be an ordering constraint on mins and maxes.

Public access to members removes any constraints.

crsib (approved):

Paul-Licameli:
Force pushed only to rebase onto recent master while waiting for QA approval.

Ryan-Miller-Audacity:
QA verified on MacOs and Windows the following: 
* Display and clicks on clip boundaries and cutlines
* Adjustment of vertical bounds and scale type in vertical spectrogram ruler
* Varying spectrogram settings of individual tracks (NOT the use of the Preferences dialog which affects defaults)
* Waveform view vertical bounds and scale type changes

All appears to be working as expected when compared to previous versions.

PR #4222, opened 2023-01-23
Extract lib track selection

Paul-Licameli:
Includes all commits of #4211 which awaits QA approval and merge

This is another step in unblocking the extraction of a toolkit neutral
library for destructive effect application and the registry of effects.

A library for changes in binary selection states of tracks in a project,
and for project sync-lock state and the sync-lock grouping logic.

TrackFocus would make sense as a part of this too, but more untangling
work must be done first before that can be toolkit-neutral, not dependent
on wxWidgets' interface for accessibility.

Recommended:


Paul-Licameli:
Rebased without changes onto more recent master.

crsib (approved):

dozzzzer:
Tested on macOS Ventura and Windows 11

Paul-Licameli:
Rebased and resolved minor conflicts before merging, but I'm waiting for the present CI problems to be cleared.  The previous merge into master has not built successfully yet.

Issue #4212, opened 2023-01-20
Improve zooming behavior


LWinterberg:
Zooming with the scrollwheel in Audacity got identified as being much too quick and unrestricted, resulting in easy overscrolling into either the sample level or the days-and-weeks level.

Another issue is that zooming near the start of the track may accidentally cause vertical zooms instead, as reported in #5096.

- [x] Set the zoom factor for mousewheel scrolling to 4th-root of 2. The button zooms stay as-is. #4243 
- [x]  Limit the range accessible to the user to 2x the project length. (so for a 10 minute project, the range between 0:00:00 and 0:20:00 can be reached). 
- [x]  When scrolling with the wheel, always center zooming in and out on the mouse cursor, with two exceptions:  
  a) if zooming would reveal an area beyond the limit, the zoom center gets moved (this is the same behavior as observed currently at the 0:00:00 mark)  
  b) (nice-to-have) when the cursor is within 5px of a range selection edge or point selection, center the zoom on the edge/point instead and align the center with the cursor
- [ ] Add a zoom sensitivity preference
- [x] When changing tempo, adjust the zoom such that the current view stays visible despite taking up different amounts of time

For clarity, there currently are some rules in place when zooming with selections and one which attempts to keep at least 1px of project visible in the zoom window at all times. These rules can be removed. 

Initial design link: https://zpl.io/ewND3ZN

petersampsonaudacity:
Testing on W10 with @LWinterberg 's latest branch build:  audacity-win-3.3.0-alpha-20230119+04efd88-x64

1) the zoom factor for scroll zooming is now less aggressive with 1.25x scale factor rather than the 2x

2) However, the Zoom In/Out buttons in the Edit toolbar and the View >Zoom > Zoom In/Out both still work with the 2x zoom factor.  This is intentional and is unlikely to be changed.

Tantacrul:
I think point 3 and 4 need just a very simple visual to make sure the desired effect is clear.

Perhaps @DilsonsPickles can aid here?

petersampsonaudacity:
@Tantacrul @DilsonsPickles  Maybe there should be a limit to zoom out such that the audio is always visible.

This, for an example use case, is a thirty-second chirp after using the zoom out button 10 times:
![image](https://user-images.githubusercontent.com/50205984/214846693-8eb7be24-eed7-4e36-8c85-dcd0e125f333.png)

LWinterberg:
See point 2 @petersampsonaudacity

[...]

PR #4243, opened 2023-01-27
Reduce zoom factor when zooming with mouse wheel

LWinterberg:
Part 1 of #4212

### DIFF
--- a/src/ProjectWindow.cpp
+++ b/src/ProjectWindow.cpp
@@ -481,7 +481,7 @@ mutable -> unsigned {
 #endif
 
       wxCoord xTrackEnd = viewInfo.TimeToPosition( audioEndTime );
-      viewInfo.ZoomBy(pow(2.0, steps));
+      viewInfo.ZoomBy(pow(2.0, steps/4.0));
 
       double new_center_h = viewInfo.PositionToTime(xx, trackLeftEdge);
       viewInfo.h += (center_h - new_center_h);


src/ProjectWindow.cpp
```diff
@@ -481,7 +481,7 @@ mutable -> unsigned {
 #endif
 
       wxCoord xTrackEnd = viewInfo.TimeToPosition( audioEndTime );
-      viewInfo.ZoomBy(pow(2.0, steps));
+      viewInfo.ZoomBy(pow(pow(2.0, 1/4.0), steps));
```
Paul-Licameli, line 484: Or simply `pow(2.0, steps/4.0)`

Paul-Licameli (approved):

Paul-Licameli (approved):
Even better now.

LWinterberg:
I'd like a test from the @Tantacrul to see if this feels better to him

SteveDaulton:
This feels worse for me as it  makes zooming much slower. For example, to zoom from a 30 second view to sample level view requires three mouse scrolls with previous version, and 10 mouse scrolls (tedious) with this patch.

Personally I have no problem with the mouse zoom in previous versions, and due to the absence of feedback to the contrary, I suspect that the vast majority of users agree.

**If** mouse zoom is too coarse for some users (I guess that is the case for [LWinterberg](https://github.com/LWinterberg)), then it may be worth making this a user preference. Personally I'd opt for the same zoom scrolling as Audacity has had for the last decade.

Paul-Licameli:
> This feels worse for me as it makes zooming much slower. For example, to zoom from a 30 second view to sample level view requires three mouse scrolls with previous version, and 10 mouse scrolls (tedious) with this patch.
> 
> Personally I have no problem with the mouse zoom in previous versions, and due to the absence of feedback to the contrary, I suspect that the vast majority of users agree.
> 
> **If** mouse zoom is too coarse for some users (I guess that is the case for [LWinterberg](https://github.com/LWinterberg)), then it may be worth making this a user preference. Personally I'd opt for the same zoom scrolling as Audacity has had for the last decade.

A user preference is always a possibility.

Just lately I was using CTRL plus mouse wheel (or actually, pinch and spread on my touch pad instead of a wheel), while the pointer was in the vertical ruler of the spectrogram view.  Maybe little known, but you can pan (with Shift) and zoom vertical scales too.  And I the defaults were too coarse for my liking.

Paul-Licameli:
I wonder if we could even experiment with some kind of accelerating zoom?  Small steps (logarithmically that is) at first, but larger if the next mouse click arrives in less than some elapsed time?  Complicating the implementation of course.

Paul-Licameli:
Anyway if the multiplier should be some constant but not 2, I suggested it should be a root of 2 so you get the exact twofold multiplier with a few wheel clicks.  It’s now the fourth root, but maybe the cube root is a compromise.

LWinterberg:
https://github.com/audacity/audacity/actions/runs/4027212455 - 4th root (4 scroll ticks = 1 current scroll tick)
https://github.com/audacity/audacity/actions/runs/4038643540 - 3rd root (3 scroll ticks = 1 current scroll tick)

The current behavior is definitely too sensitive. It takes a single scroll to go from default zoom to half a day. With a free-spinning mousewheel, I can zoom from one extreme to the other faster than I can put my finger down to stop the scroll midway. With a trackpad, taking off the fingers from the pad can cause a tick (rarely two) of zooming, which throws off the zoom level by up to an order of magnitude. 

With the 3rd root, a single scroll goes from default zoom to about a song length. With a free-spinning wheel I now have hope of catching it at a reasonable level, and the trackpad errors are nowhere near as significant.

Tantacrul:
@LWinterberg - they're both quite good. I marginally prefer: https://github.com/audacity/audacity/actions/runs/4027212455

As for hiding this change behind a setting. I think that will only unfortunately do two undesirable things

- It won't solve the quite poor performance of the mouse wheel when zooming (responsiveness is part 1 of the issue)
- New settings just introduce more complexity that 95% of our audience will never discover

You can't make changes of any kind without _someone_ being unhappy with it.

SteveDaulton:
> As for hiding this change behind a setting.

Adding a preference does not in any way "hide" the change if you make your preferred setting the default. It just acknowledges that some users may prefer mouse zoom the way that it has been for the last decade.

> You can't make changes of any kind without _someone_ being unhappy with it.

Surely, when some users prefer a feature one way, and other users prefer it another, then it makes sense for it to be a user preference.

Tantacrul:
I think the simplest summary I can give is:

- I doubt too many people will object (or even notice) a slightly less responsive setting for zoom.
- If I'm wrong, we can create a setting for it
- Regardless, the revised version should be the default (unless my finger is completely off the pulse and most users totally hate it)

petersampsonaudacity:
Testing on W10 with @LWinterberg 's latest branch:  audacity-win-3.3.0-alpha-20230129+20dfd68-x64

I find myself liking the less sensitive mouse-wheel scrolling.

I do note that the old zoom factor of 2x still applies to the zoom in/out buttons and commands - and I like that this zoom factor is retained for the menu commands and Edit toolbar buttons.

petersampsonaudacity:
> This feels worse for me as it makes zooming much slower. For example, to zoom from a 30 second view to sample level view requires three mouse scrolls with previous version, and 10 mouse scrolls (tedious) with this patch.

This only takes me (with Leo's branch build) 5 mouse scrolls to go from 30secs to sample level view at full screen size - and only 4 mouse scrolls at default window size,  where a mouse scroll is a full push of the wheel i.e. mutiple mouse-wheel clicks and not a single click.

I also note that if I spin the mouse-wheel forward fast with Ctrl held down the it zooms straight to maximum zoom individual sample level.  Is this an easter-egg?

petersampsonaudacity:
I've just run this past my "Mystery-Shopper" - she too thinks the less aggressive mouse zoom is much better ...

Tantacrul:
We're getting dangerously close to unanimous agreement here. I feel weird.

Ryan-Miller-Audacity:
Tested on Windows and MacOS and confirmed that Zoom factor when zooming with mouse wheel has been reduced and features a much higher degree of usability for general users.

PR #4246, opened 2023-01-27
fix miscellaneous compilation issues for musl libc

nekopsykose:
fixes various issues compiling under musl libc:

- u_short isn't a standard typename and doesn't exist
- <sys/select.h> is needed for fd_ functions
- assigning to stdout/err is invalid, but freopen already does this so there is no need to

Recommended:

### DIFF
--- a/libraries/lib-ipc/IPCClient.cpp
+++ b/libraries/lib-ipc/IPCClient.cpp
@@ -13,6 +13,7 @@
 #include "IPCClient.h"
 #include "IPCChannel.h"
 
+#include <cstdint>
 #include <thread>
 #include <stdexcept>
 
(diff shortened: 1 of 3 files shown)


crsib:
Windows build has failed, there is no select.h under Windows

nekopsykose:
> Windows build has failed, there is no select.h under Windows  

right, good point. guarded with not _WIN32

crsib (approved):

libraries/lib-ipc/IPCChannel.h
```diff
@@ -13,6 +13,9 @@
 #pragma once
 
 #include <cstddef>
+#ifndef _WIN32
+#include <sys/select.h>
```
vsverchinsky, line 17: `IPCChannel.h` is a public interface file, please move this include to `internal/ipc-types.h`

vsverchinsky (approved):
Approved with one small suggestion

PR #4254, opened 2023-01-31
Update German from Transifex

Gootector:
Update German from Transifex

Greetings,
Gootector

Recommended:

### DIFF
--- a/locale/de.po
+++ b/locale/de.po
@@ -1,38 +1,26 @@
 # SOME DESCRIPTIVE TITLE.
-# Copyright (C) YEAR Audacity Team
+# Copyright (C) 2023 Audacity Team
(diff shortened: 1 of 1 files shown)


LWinterberg:
These commits only change typographically correct … to ASCII-compatible ... and remove some duplicate (?) credits from the file afaict. Were they causing trouble?

Gootector:
The merging into the master is your decision. I can create 1 commit PR, if this PR, with 4 commits, is some problem. Or I can close this PR and create a new, when there will be more changes (if there will be any changes).
I think that ordered credits are better than disordered and duplicated credits. But everyone has their own opinion 🤔
Greetings,
Gootector

PR #4257, opened 2023-02-01
Effect UI untangling

Paul-Licameli:

Many steps toward extraction of the ui-independent parts of destructive effect
processing into a library.

One minor change in observable behavior:  certain modal dialogs may open while
an effect dialog is open (from sidebar or from the main menus).  These may now be
positioned differently, no longer centered as before above the effect dialog.  Preserving
the old behavior would add unnecessary bother to the implementation.

Also removes some, but not all, remaining state still carried in Effect objects that we
have called "stateless" (enough for realtime processing purposes) and so makes
more functions const (but adds some more const_casts in stateful effects).

Also improves terminology.  EffectUIValidator is renamed EffectEditor and
EffectUIClientInterface becomes EffectUIServices.

EffectUIServices is an interface of const member functions only, with one exception,
and they take const EffectPlugin & as context.  Just as a stateless Effect is a factory
for Instances, so now EffectUIServices is a factory for Editors.

Concrete effect classes inherit both EffectUIServices and EffectPlugin but EffectPlugin
now has no UI related functions in it.  The EffectUIEditor interface of an effect is
discovered where needed by a dynamic_cast.  Passing the effect again as the EffectPlugin
argument allows EffectUIServices not to be based on EffectPlugin, and enables some
common implementations of its methods to query the effect via EffectPlugin for some
other needed information.

For QA:  test these things
- Preview play of any effect in the destructive effect dialog -- including Amplify and Sliding stretch
- Showing and hiding of non-modal effect dialog; exist of program while it still shows
- Modal dialogs for effect options (from the manage menu) and for import/export -- as noted, positioning may be different, but let's not care about that change.  (Ladspa, AU, VST, VST3, LV2, and at least one built-in)
- Thoroughly testing dialogs of Echo; the Amplify effect; and Normalize
- 

Recommended:


src/effects/EffectBase.h
```diff
@@ -37,10 +38,17 @@ class AUDACITY_DLL_API EffectBase /* not final */
 
    void SetTracks(TrackList *pTracks) { mTracks = pTracks; }
 
+   //! Called when Preview() starts, to allow temporary effect state changes
+   /*!
+    default returns a null
+    @return will undo its effects in its destructor before Preview() finishes
+    */
+   virtual std::any BeginPreview(const EffectSettings &settings);
```
crsib, line 44: If it is called only by Preview, why is it public? It feels that it should be protected instead.

crsib: Actually, this is a rare case where I don't disagree with `any` :-)

src/effects/Effect.cpp
```diff
@@ -155,18 +155,17 @@ int Effect::ShowClientInterface(wxWindow &parent, wxDialog &dialog,
    EffectUIValidator *, bool forceModal)
 {
    // Remember the dialog with a weak pointer, but don't control its lifetime
```
crsib: Comment appears to be out of date

src/effects/EffectBase.cpp
```diff
@@ -390,10 +391,6 @@ void EffectBase::Preview(EffectSettingsAccess &access, bool dryOnly)
             std::dynamic_pointer_cast<EffectInstanceEx>(MakeInstance())
          )
             pInstance->Init();
-
-      // In case any dialog control depends on mT1 or mDuration:
-      if ( mUIDialog )
```
crsib, line 388: I wonder why this branch is missing now

src/effects/EffectUI.cpp
```diff
@@ -755,8 +755,12 @@ void EffectUIHost::OnPlay(wxCommandEvent & WXUNUSED(evt))
       if (!TransferDataFromWindow())
          return;
       
-      mEffectUIHost.Preview(*mpAccess, false);
-      
+      auto updater = [this]{ TransferDataToWindow(); };
+      mEffectUIHost.Preview(*mpAccess, updater, false);
+      // After restoration of settings and effect state:
+      // In case any dialog control depends on mT1 or mDuration:
+      updater();
```
crsib, line 771: Oh, nvm question about the branch

src/effects/Effect.cpp
```diff
@@ -68,11 +68,6 @@ Effect::Effect()
 
 Effect::~Effect()
 {
-   // Destroying what is usually the unique Effect object of its subclass,
-   // which lasts until the end of the session.
-   // Maybe there is a non-modal realtime dialog still open.
-   if (mHostUIDialog)
```
crsib, line 74: I wonder if should revisit the modality of destructive dialogs though. It appears that we have broken a common workflow by making destructive application dialogs modal. 

This won't really matter when automation is introduced, so probably it doesn't really matters.

src/EffectPlugin.cpp
```diff
@@ -25,9 +25,9 @@ bool EffectInstanceEx::Init()
 }
 
 EffectUIValidator::EffectUIValidator(
-   EffectUIClientInterface &effect, EffectSettingsAccess &access)
-   : mEffect{effect}
-   , mAccess{access}
+   EffectUIServices &services, EffectSettingsAccess &access)
+   : mServices{ services }
```
crsib: Minor suggestion - let's call it mUIServices maybe?

src/effects/EffectBase.h
Paul-Licameli: Protected makes sense.

std::any is just type erasure for things that can copy and may have nontrivial destructors.

My type erased pipeline class also uses it.

src/effects/Effect.cpp
Paul-Licameli: I think I removed it at some late commit in this long sequence?

src/effects/EffectBase.cpp
Paul-Licameli: See the finally

src/EffectPlugin.cpp
Paul-Licameli: Ok

crsib (approved):
A good step towards toolkit neutrality for the effect.

Paul-Licameli:
Rebased and force pushed with no changes or conflicts

Paul-Licameli:
Rebased onto recent master, with conflicts resolved in CMakeLists.txt files and in include directives

Paul-Licameli:
Rebased without change onto more recent master.

dozzzzer:
@Paul-Licameli 
Moving back to To Do: loading a VST effect causes deadlock in both real-time and destructive application

STR:
1. Have some audio
2. Select a VST effect from the main Effect menu OR
3. Select a VST effect in the sidebar and try to open its dialog

Act. result: Audacity hangs completely

Paul-Licameli:
@crsib please see how the new commits first turn the stack overflow into a build failure, then fix VST effects.

Once these are understood I want to squash them earlier in the branch.

crsib (approved):

dozzzzer:
The issue above bisects to https://github.com/audacity/audacity/commit/95b7997ebb56e914d185cc3ed07e2dc55b7d5699
@Paul-Licameli

Paul-Licameli:
> The issue above bisects to [95b7997](https://github.com/audacity/audacity/commit/95b7997ebb56e914d185cc3ed07e2dc55b7d5699)
> @Paul-Licameli

If so, then this branch is not the cause of it, but an issue should be opened about this recent regression and fixed as soon as I can.

Paul-Licameli:
Rebased onto recent master to resolve small conflicts, but no other changes

dozzzzer:
Verified

Issue #2427, opened 2022-01-13
Linux: Playback stutters at start


SteveDaulton:
**Describe the bug**
The audio stutters when playback starts. This is a regression against 3.2.0-alpha-20211210

See
e4f215f Good
e2b1a2b Bad

**To Reproduce**
1. Generate a Chirp
2. Press Play
3. Hear a stutter as playback starts.

**Expected behavior**
Playback to start without stuttering.

**Screenshots**

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version : 3.2.0-alpha-20220113
 - (if relevant) Sound Device: ALSA default (internal sound card)

**Additional context**
Playback cursor does not move.

See also #2314, #2331

Bisected to: 81f74f7
Consider commit range: 81f74f72485c81dc90e21894270e9bea34e03a0a...fc458eb314d633b10bf0e688da6d269488bfef03

The commit fc458eb314d633b10bf0e688da6d269488bfef03 fixed similar issue #2314

Penikov:
Audacity 3.2. not only stutters but stucks at the playback on Win10

SteveDaulton:
In 3.2.0-alpha-20220115, the playback cursor moves with playback, but the stuttering still occurs.

AnitaBats:
@SteveDaulton could you please tell the bisection bracket?

SteveDaulton:
e4f215fa645dfb8a6f6aaedf53c134d9aba9e503 Good
e2b1a2be6b92661628bbb054f915bc50b211c020 Bad

SteveDaulton:
81f74f72485c81dc90e21894270e9bea34e03a0a is the first bad commit

werame:
I've been using my own builds from based on https://github.com/audacity/audacity/commit/fc587e67556a940a5d46eb0fe9fbdbfcbfc0b43e and they don't stutter. 

There is a strange issue when switching to asio as it takes a few seconds to "think about it" in the drop-down menu for the output type switching, but playback is ok otherwise.

SteveDaulton:
> I've been using my own builds from based on [fc587e6](https://github.com/audacity/audacity/commit/fc587e67556a940a5d46eb0fe9fbdbfcbfc0b43e) and they don't stutter.

On Linux?

werame:
Windows. No ASIO on Linux, afaik... On a quick test, all 4 interfaces (WASAPI, MME, DS, ASIO) are ok with that build. 

[...]

PR #4263, opened 2023-02-03
Take hardware latency into account when allocating buffers

crsib:
Resolves: #2427 

This PR uses the output latency reported by PortAudio as a hint about the ring buffer size and granularity. 

Recommended:

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -608,6 +608,12 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
       SetCaptureMeter( mOwningProject.lock(), options.captureMeter );
    }
 
+   const auto deviceInfo = usePlayback ?
+                              Pa_GetDeviceInfo(playbackParameters.device) :
+                              Pa_GetDeviceInfo(captureParameters.device);
+
+   mUsingAlsa = deviceInfo && deviceInfo->hostApi == paALSA;
+
    SetMeters();
 
 #ifdef USE_PORTMIXER
@@ -794,12 +800,6 @@ int AudioIO::StartStream(const TransportTracks &tracks,
          std::this_thread::sleep_for(50ms);
       }
    }
-
-#ifdef __WXGTK__
-   // Detect whether ALSA is the chosen host, and do the various involved MIDI
-   // timing compensations only then.
-   mUsingAlsa = (AudioIOHost.Read() == L"ALSA");
-#endif
 
    gPrefs->Read(wxT("/AudioIO/SWPlaythrough"), &mSoftwarePlaythrough, false);
    gPrefs->Read(wxT("/AudioIO/SoundActivatedRecord"), &mPauseRec, false);
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -626,7 +626,7 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
 #endif
 
    // July 2016 (Carsten and Uwe)
-   // BUG 193: Possibly tell portAudio to use 24 bit with DirectSound. 
+   // BUG 193: Possibly tell portAudio to use 24 bit with DirectSound.
    int  userData = 24;
    int* lpUserData = (captureFormat_saved == int24Sample) ? &userData : NULL;
 
@@ -650,6 +650,23 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
                                     paNoFlag,
                                     audacityAudioCallback, lpUserData );
       if (mLastPaError == paNoError) {
+         const auto stream = Pa_GetStreamInfo(mPortStreamV19);
+         // Use the reported latency as a hint about the hardware buffer size
+         // required for uninterrupted playback.
+         mHardwarePlaybackLatencyFrames = lrint(stream->outputLatency * mRate);         
+#ifdef __WXGTK__
+         // DV: When using ALSA PortAudio does not report the buffer size.
+         // Instead, it reports periodSize * (periodsCount - 1). It is impossible
+         // to retrieve periodSize or periodsCount either. By default PA sets
+         // periodsCount to 4. However it was observed, that PA reports back ~100msec
+         // latency and expects a buffer of ~200msecs on Audacity default settings
+         // which suggests that ALSA changes the periodsCount to suit its needs.
+         // 
+         // Why 3? 2 doesn't work for me, 3 does :-) So similar to PA - this
+         // is the value that works for author setup. 
+         if (mUsingAlsa)
+            mHardwarePlaybackLatencyFrames *= 3;
+#endif
          break;
       }
       wxLogDebug("Attempt %u to open capture stream failed with: %d", 1 + tries, mLastPaError);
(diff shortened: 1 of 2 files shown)


crsib:
@SteveDaulton can you please check that the issue is for you? I don't have physical Linux installation right now.

DavidBailes:
@crsib . This change affects the keyboard scrubbing using the u and i keys on Windows.
In previous verisons of Audacity you can quickly change the direction of the scrubbing:
1. hold down the i key to scrub forwards.
2. hold down the u key. scrub immediately changes direction.
3. release the i key.

With this change, in step 2, the scrubbing does not change direction. To change direction you have to release the i key, and then hold down the u key.

crsib:
Thanks, David, I will check that out!

DavidBailes:
> Thanks, David, I will check that out!

Thanks. Another problem is that changing the zoom level whilst either the u or i key is being held down no longer immediately changes the scrub speed.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -573,6 +573,8 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
          const PaHostApiInfo* hostInfo = Pa_GetHostApiInfo(playbackDeviceInfo->hostApi);
          bool isWASAPI = (hostInfo && hostInfo->type == paWASAPI);
          playbackParameters.suggestedLatency = isWASAPI ? 0.0 : latencyDuration/1000.0;
+
+         mUsingAlsa = (hostInfo && hostInfo->type == paALSA);
```
Paul-Licameli, line 577: This assignment happens only when `mSoftwarePlaythrough` is false.  I think it should be independent of that.

crsib: Hmm, great catch!

Paul-Licameli: It happens too only when playing, not when recording only (without overdub).

Is there any similar issue with mixed up timing at the start of recording with ALSA?  But I am aware of no complaint.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -644,6 +646,23 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
                                     paNoFlag,
                                     audacityAudioCallback, lpUserData );
       if (mLastPaError == paNoError) {
+         const auto stream = Pa_GetStreamInfo(mPortStreamV19);
+         // Use the reported latency as a hint about the hardware buffer size
+         // required for uninterrupted playback.
+         mHardwarePlaybackLatencyFrames = lrint(stream->outputLatency * mRate);
```
Paul-Licameli, line 652: Did you see also in MIDIPlay.cpp
```

   // We use audio latency to estimate how far ahead of DACS we are writing
   if (info) {
      // this is an initial guess, but for PA/Linux/ALSA it's wrong and will be
      // updated with a better value:
      mAudioOutLatency = info->outputLatency;
      mSystemMinusAudioTimePlusLatency += mAudioOutLatency;
   }
```

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -644,6 +646,23 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
...
+         // Use the reported latency as a hint about the hardware buffer size
+         // required for uninterrupted playback.
+         mHardwarePlaybackLatencyFrames = lrint(stream->outputLatency * mRate);         
+#ifdef __WXGTK__
+         // DV: When using ALSA PortAudio does not report the buffer size.
+         // Instead, it reports periodSize * (periodsCount - 1). It is impossible
+         // to retrieve periodSize or periodsCount either. By default PA sets
+         // periodsCount to 4. However it was observed, that PA reports back ~100msec
+         // latency and expects a buffer of ~200msecs on Audacity default settings
+         // which suggests that ALSA changes the periodsCount to suit it's needs.
```
Paul-Licameli, line 659: "its"

libraries/lib-audio-io/AudioIO.cpp
crsib: I saw that, but estimation there happens too late to fix this bug.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -1139,11 +1152,18 @@ bool AudioIO::AllocateBuffers(
       try
       {
          if( mNumPlaybackChannels > 0 ) {
-            // Allocate output buffers.  For every output track we allocate
-            // a ring buffer of ten seconds
+            // Allocate output buffers.
+            // Allow at least 2x of the buffer latency.
             auto playbackBufferSize =
-               (size_t)lrint(mRate * mPlaybackRingBufferSecs.count());
+               std::max((size_t)lrint(mRate * mPlaybackRingBufferSecs.count()), mHardwarePlaybackLatencyFrames * 2);
```
Paul-Licameli, line 1158: Does this line cause no change on Mac and Windows?

crsib: I have observed none, both usually report relatively low latency and report it correctly. This specific line cannot really hurt, there is huge inconsistency between `playbackBufferSize`  on different policies and I don't really think that provided ring buffer size matters that much if it is sufficient. 

`mPlaybackQueueMinimum` is different, but it must handle the reported output latency anyway. Simply using some random number is incorrect. Yet again - reported latency is usually lower than the one requested by playback policy. This PR does affect responsiveness of scrubbing on ALSA, but the bug itself makes Audacity **unusable** on ALSA, which is much worse.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -1166,6 +1186,14 @@ bool AudioIO::AllocateBuffers(
             mPlaybackQueueMinimum =
                std::min( mPlaybackQueueMinimum, playbackBufferSize );
 
+            // Limit the mPlaybackQueueMinimum to the hardware latency
+            mPlaybackQueueMinimum =
+               std::max(mPlaybackQueueMinimum, mHardwarePlaybackLatencyFrames);
+
+            // Make mPlaybackQueueMinimum a multiple of mPlaybackSamplesToCopy
+            mPlaybackQueueMinimum = mPlaybackSamplesToCopy *
+               ((mPlaybackQueueMinimum + mPlaybackSamplesToCopy - 1) / mPlaybackSamplesToCopy);
```
Paul-Licameli, line 1195: This last assignment may make `mPlaybackQueueMinimum` still larger than `mHardwarePlaybackLatencyFrames`

Is that what you want?  Or should it be

`mPlaybackQueueMinimum -= mPlaybackQueueMinimum % mPlaybackSamplesToCopy`

crsib: It should be `not less`, so larger definitely works, otherwise bug won't be fixed.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -1139,11 +1152,18 @@ bool AudioIO::AllocateBuffers(
...
             auto playbackBufferSize =
-               (size_t)lrint(mRate * mPlaybackRingBufferSecs.count());
+               std::max((size_t)lrint(mRate * mPlaybackRingBufferSecs.count()), mHardwarePlaybackLatencyFrames * 2);
+
+            // Make playbackBufferSize a multiple of mPlaybackSamplesToCopy
+            playbackBufferSize = mPlaybackSamplesToCopy *
+               ((playbackBufferSize + mPlaybackSamplesToCopy - 1) / mPlaybackSamplesToCopy);
 
+            // Adjust mPlaybackRingBufferSecs correspondingly
+            mPlaybackRingBufferSecs = PlaybackPolicy::Duration { playbackBufferSize / mRate };
```
Paul-Licameli, line 1165: Irrelevant to the bug fix, but I notice now that `mPlaybackRingBufferSecs` can be a local variable of this function only, eliminating a member variable

libraries/lib-audio-io/AudioIO.cpp
Paul-Licameli: Oh I was being silly

Paul-Licameli: Still, in two places, one thing is rounded to a multiple of another thing.  I don't yet understand the need.

Paul-Licameli (changes requested):
So the one thing to fix is `mUsingAlsa`

I don't feel like I understand all the intricacies but I see nothing wrong looiking and I trust the results of experiment.

crsib:
Thank you, @Paul-Licameli. I have updated the first commit detecting ALSA differently

PR #4266, opened 2023-02-03
Update french translation

Leaumonte:
Resolves: Some untranslated messages, fix or update of messages

To translate the new amazing features for real time effect for French people, like me ;-)

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes of behavior => only the last commit compiles sorry

### DIFF
--- a/locale/fr.po
+++ b/locale/fr.po
@@ -3571,7 +3571,7 @@ msgstr "Continuer"
 #: src/IncompatiblePluginsDialog.cpp
 #, c-format
 msgid "Audacity has found %d incompatible plugins which could not be loaded. We have disabled these plugins to avoid any stalling or crashes. If you would still like to attempt to use these plugins, you can enable them using \"Manage Plugins\". Otherwise, select \"Continue\"."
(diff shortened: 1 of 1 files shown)


locale/fr.po
```diff
@@ -45,7 +46,7 @@ msgstr "Erreur inconnue"
 
 #: crashreports/crashreporter/CrashReportApp.cpp
 msgid "Problem Report for Audacity"
-msgstr "Rapport de problème pour Audacity"
+msgstr ""
```
crsib, line 49: I wonder why this got empty :-)

Leaumonte: Because the guys that receive this email, doen't speak french ;-)

crsib: Well, we can use Google Translate. 

On top of that - Crash Reporter sends out a binary with a stack trace, no user input is allowed

crsib:
Something got wrong

```
fr.po:70: number of format specifications in 'msgid' and 'msgstr' does not match
```

Leaumonte:
> 

Sorry I didn't run the job check, only a local check...

It's done now 

https://github.com/Leaumonte/audacity/actions/runs/4165786772

I fix the two issues in locale/fr.po

> Something got wrong
> 
> ```
> fr.po:70: number of format specifications in 'msgid' and 'msgstr' does not match
> ```

LWinterberg (approved):

Issue #4272, opened 2023-02-03
Normalize tracks [in]dependently


JC3:
### Your idea

An additional checkbox in the Normalize settings dialog that reads "Normalize Tracks Independently".

This checkbox should be checked by default to match the current behavior.

When unchecked, all tracks should be normalized non-independently, i.e. to the maximum value of all of the selected tracks together.

*Interaction with "Normalize Stereo Channels Independently":*

| | Independent Stereo Checked | Independent Stereo Unchecked |
|-|-|-|
| Independent Track Checked | Normalize every channel of every track independently. (Current behavior) | Link stereo channels, normalize every track independently. (Current behavior) |
| Independent Track Unchecked | See note below. | Link stereo channels. The maximum value for stereo tracks is the combined max of both channels. The maximum value for mono tracks is just the max of the mono channel. Normalize every track to the max of these maximums. |

When "Normalize Stereo Channels Independently" is checked but "Normalize Tracks Independently" is unchecked, we have a choice of how to handle this. Options include, but are probably not limited to:

- Option 1: Disallow this combination: When "Normalize Tracks Independently" is unchecked, change "Normalize Stereo Channels Independently" to an unchecked, disabled state.
- Option 2: Use a weird behavior that I can't think of a use case for (is there one?): For stereo tracks, normalize the channels independently, and treat the maximum value of the track as a whole as the max of the two stereo channels. Then, use that when considering max track values for the final normalization.
- Option 3: Instead of adding a new check box, replace "Normalize Stereo Channels Independently" with a drop down containing three choices: "Normalize Tracks Independently", "Normalize Stereo Channels Independently", and "Normalize All Tracks Together" (or something to that effect), thus avoiding the combination entirely.

### Problem to be solved

Inability to link multiple non-stereo tracks when recording. For example, normalizing a multi-microphone setup where each microphone was gain-calibrated ahead of time (and thus the relative gains should not be changed by normalization).

The workaround of temporarily joining two mono tracks into a stereo track does not work when there are more than 2 tracks.

JC3:
As an aside, this would be less confusing to describe if "Normalize Stereo Channels Independently" was changed to "Link Stereo Channels" and inverted.

JC3:
@petersampsonaudacity Is there a unit testing framework buried in here somewhere?

[...]

PR #4276, opened 2023-02-05
Add support for linked track normalization

JC3:
Resolves: #4272 

Adds "Normalize Tracks Independently" check box to Normalize effect, checked by default.

When unchecked, the maximum values of all selected tracks will be considered together, and every selected track will be multiplied by the same multiplier. I.e. all tracks will be linked when normalizing.

Includes a number of supporting code changes, notably separating analysis step and processing step into two different loops.

i18n: This adds one new string that will need translated ("Normalize tracks independently").

See #4272 for discussion.
See commit log for details.

Test file: [normalize_test.aup3.gz](https://github.com/audacity/audacity/files/10609962/normalize_test.aup3.gz)

Test instructions:
- With "Normalize Tracks Independently" checked, select various combinations of tracks and observe results.
- With "Normalize Tracks Independently" unchecked, select the tracks that aren't already normalized (i.e. "80+0" and "90+0") and observe results.

Recommended:

### DIFF
--- a/src/effects/Normalize.cpp
+++ b/src/effects/Normalize.cpp
@@ -36,7 +36,7 @@
 const EffectParameterMethods& EffectNormalize::Parameters() const
 {
    static CapturedParameters<EffectNormalize,
-      PeakLevel, ApplyGain, RemoveDC, StereoInd
+      PeakLevel, ApplyGain, RemoveDC, StereoInd, TrackInd
    > parameters;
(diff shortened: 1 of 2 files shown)


JC3:
Discussion is at #4272.

JC3:
Resolution: PR is unnecessary because feature is redundant with Amplify on multiple tracks.

PR #4290, opened 2023-02-07
Pipeline parallelism for some exports

Paul-Licameli:

Recommended:

### DIFF
--- a/libraries/CMakeLists.txt
+++ b/libraries/CMakeLists.txt
@@ -58,6 +58,7 @@ set( LIBRARIES
    lib-mixer
    lib-channel
    lib-stretching-sequence
+   lib-concurrency
 )
 
 if ( ${_OPT}has_networking )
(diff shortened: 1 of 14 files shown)


Paul-Licameli:
> 1. Exporting as **MP3 256kbps constant Bitrate Stereo** using the Audacity command **File>Export>Export as MP3**
>    I get a slight speed improvement:
>    2.4.3    2min:06sec
>    Paul's   1min:52sec
> 2. Exporting as **WAV 16-bit PCM** I get a slightly better improvement
>    2.4.3    1min:03sec
>    Paul's   0min:35sec

Thanks, Peter.  This is just one trial.  What can you tell us about your computer's CPU and storage devices?  How many cores?

MP3 exporting cut the time by exactly 1/9 or 11.11%.
WAV improved by exactly 4/9 or 44.44% -- that looks really good

So preliminary data suggests then I find out where the number 9 is in the code, and try changing it to 10. 🤪 

More seriously some other things I have seen suggest that I can't improve MP3, because the compute-bound compression step, middle stage of 3, now occupies one thread and there is no obvious way to divide that into more pipeline stages.

But in the case of WAV, there are only two stages, and the downstream one includes the Soundfile library process and writes to disk, but it waits on the slower upstream stage, which is similar to the first stage for MP3.  This is Audacity's Mixer class at work and we can still tune that.

petersampsonaudacity:
>Thanks, Peter. This is just one trial. What can you tell us about your computer's CPU and storage devices? How many cores?

image


image

Issue #4289, opened 2023-02-07
The Macro Command "Set Clip" doesn't save parameter "Name"


JohnColket:
### Bug description

This issue is fully described in this forum post: https://forum.audacityteam.org/viewtopic.php?f=69&t=129092

Along with a description of the problem is a proposed solution, making this a GOOD FIRST ISSUE for aspiring developers.

### Steps to reproduce

See referenced forum post.

### Expected behavior

Name parameter should be saved as part of macro

### Actual behavior

Name parameter is not saved within macro

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

JohnColket:
> This is my first time contributing , where is the source code ?

The current alpha source code is here: https://github.com/audacity/audacity
You would be expected to submit pull request here: https://github.com/audacity/audacity/pulls
For help compiling and making pull requests, ask here: https://discord.gg/6km9SeUAdN

JohnColket:
Note that others may have already looked into this issue.  It may be best to verify that the issue is present in the current Alpha nightly builds before continuing. See: https://nightly.link/audacity/audacity/workflows/build/master

aymza:
Issue still present in Audacity 3.7.7: when applying multiple "SetClip: Name" with different names in a macro, all the clips are renamed with the same name.

PR #4293, opened 2023-02-07
Added "Name" parameter to Macro command "Set Clip" issue #4289

duckling69:
Resolves: #4289

Recommended:


Paul-Licameli (changes requested):
This is incomplete.  It may make an entry for the name in the dialog that prompts for macro parameters, but then the execution of the command does nothing with the input.

Paul-Licameli (changes requested):
This is incomplete.  It may make an entry for the name in the dialog that prompts for macro parameters, but then the execution of the command does nothing with the input.

Issue #1156, opened 2021-06-25
Extract lib-shuttlegui


Paul-Licameli:
- [x] #1157 
- [ ] #1366
- [ ] remaining

PR #4299, opened 2023-02-09
Extract lib shuttlegui

Paul-Licameli:
Resolves: #1156
Resolves: #1366

Extract a library that is not toolkit-neutral.  But this may lead to modules for Ladspa and Nyquist effects that do not depend on the executable.

A better replacement for ShuttleGui should be found that is a toolkit neutral facade awaiting its own back end reimplementation in Qt.  Then this would require rewrites of existing ShuttleGui uses but the correspondences should be straightforward.

Nodes of the graph of modules and libraries that depend on wxWidgets only through lib-shuttlegui are colored yellow.

Also two libraries that are toolkit neutral.  One, which lib-shuttlegui depends on and lib-effects also will depend on, is for visiting of effect or command parameters.

Another, not really related to the others but also in this PR as a dependency lib-effects will need:  a library for the choices of numeric formats in time/frequency controls, and the memory of per-project formats.

Recommended:


Paul-Licameli:
Sorry, the Linux build failed.  I insert a commit in the branch to fix that.  Try again...

libraries/lib-foo/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+]]#
+
+set( SOURCES
+   Foo.cpp
+)
+set( LIBRARIES
+   lib-shuttlegui-interface
+)
+audacity_library( lib-foo "${SOURCES}" "${LIBRARIES}"
```
crsib, line 24: I am a bit lost here :-)

libraries/lib-shuttlegui/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+#[[
+An embedded language for describing dialog layouts, and also alternative
```
crsib, line 2: I don't agree that ShuttleGui should be referred to as a language

libraries/lib-shuttlegui/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+visitations of the layout description to construct a dialog, or to shuttle
+values between its controls and program variables or preferences.
+
+Some good properties that this library does not have, but that the hoped for
```
crsib, line 6: I'm not sure, that commenting a possible future makes sense here. 

To be fair, after working much more time on Audacity since I proposed that we can do some replacement I don't think it would be feasible. We won't be able to keep any of UI related code either way.

libraries/lib-parameters/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+   ShuttlePrefs.cpp
+   ShuttlePrefs.h
+   WrappedType.cpp
+   WrappedType.h
+)
+set( LIBRARIES
+   lib-preferences-interface
+   wxBase
+)
+audacity_library( lib-parameters "${SOURCES}" "${LIBRARIES}"
```
crsib, line 20: Probably `lib-shuttle-parameters`? I would never expect to see Shuttle.h in a library called `parameters`

Paul-Licameli:
Rebased onto more recent master, with no changed in earlier commits, but two new commits.

crsib:
> Rebased onto more recent master, with no changed in earlier commits, but two new commits.

From the git perspective - every commit is new now, so I need to restart the review.

src/commands/CommandBuilder.cpp
```diff
@@ -82,6 +82,60 @@ void CommandBuilder::Success(const OldStyleCommandPointer &cmd)
    mValid = true;
 }
 
+namespace {
+// This class was formerly in Shuttle.cpp and inherited Shuttle but the
+// polymorphism wasn't really needed in the sole use of the class in this file
+struct ShuttleCli final // : public Shuttle
```
crsib, line 88: A very welcome change, thank you! :-)

libraries/lib-parameters/CMakeLists.txt
Paul-Licameli: "Shuttle" is some not very descriptive terminology from James.  I would call it "ParameterVisitor" if I wanted to fix terminology too instead of just moving files.

crsib: `lib-parameters` is a way to generic name in any case

libraries/lib-shuttlegui/CMakeLists.txt
Paul-Licameli: I have not yet given up that hope.

libraries/lib-shuttlegui/CMakeLists.txt
crsib: In case we decide to push for `Shuttle2` I don't think it will be a part of this specific library. I think we will have something like `lib-declarative-ui` or `lib-shuttle2` :-)

libraries/lib-parameters/CMakeLists.txt
Paul-Licameli: Macro commands and effects both use them.  Lib-effect-parameters would be too specific.  What is a good but not long name?  I don’t like the legacy term “shuttle”

libraries/lib-shuttlegui/CMakeLists.txt
Paul-Licameli: Yes, if there is a replacement, it will be in another library, and this one, deprecated

Paul-Licameli:
All right @crsib, in answer to your criticims, consider this a draft again.

I want to make a new, smaller PR only with the parts that I need for now, to take out a toolkit neutral lib-effects.

I will call the new library lib-command-parameters, taking "command" as the inclusive term for macro commands and effects.

PR #4304, opened 2023-02-11
Extract lib effects

Paul-Licameli:

A library for the ui-less part of effect processing

QA:  test some effects.
- [x] Do destructive application of built-in effects
- [x] Do rendering of built-in effects -- both with stackable, and with macros (wider choice available)
- [x] test at least one error message box, such as trying EQ on two tracks with different sample rates
- [x] Get progress dialogs on long tracks, and cancel

Recommended:


Paul-Licameli:
Rebased these four new commits onto the bug-fixed #4313, and fixed an error in the third commit.

crsib (approved):

dozzzzer:
Tested on Windows 11 & macOS Ventura

PR #4314, opened 2023-02-15
Break an effect dependency

Paul-Licameli:

Break dependency of Effect on LabelTrack, for which there is not (yet) a library.

Even if there is such a library later, better not to have this extra dependency.

Extract utilities from Effect that are only used by some Analyzers.

QA:  things to test include Vamp analyers and the Find Clipping analyzer.

Recommended:


vsverchinsky (approved):

dozzzzer:
Verified

Issue #3137, opened 2022-06-25
[NYQUIST] link error, nyq_reformat_aud_do_response() not found.


carlo-bramini:
I configured the build tree with the option `-Daudacity_use_nyquist=off` and I got this error when linking:

```
CMakeFiles/Audacity.dir/commands/ScriptCommandRelay.cpp.o: in function `ExecForLisp(char*)':
/home/Carlo/audacity/src/commands/ScriptCommandRelay.cpp:102: undefined reference to `nyq_reformat_aud_do_response(wxString const&)'
collect2: error: ld returned 1 exit status
```

It seems that all other calls to the functions of libnyquist are guarded by `#if USE_NYQUIST ... #endif`, so this may be the only forgotten point.

PR #4322, opened 2023-02-17
[NYQUIST] fix nyq_reformat_aud_do_response() not found. 

carlo-bramini:
Resolves: [*(direct link to the issue)*](https://github.com/audacity/audacity/issues/3137)

If NYQUIST is disabled when configuring Audacity, the build process hangs with an error because `nyq_reformat_aud_do_response()` is not found when linking, see issue #3137. Since `ExecForLisp()` is called only by `src\effects\nyquist\Nyquist.cpp` but this source is not compiled if Nyquist is disabled:

https://github.com/audacity/audacity/blob/ea8163f94e58899844a173c679f5a267b4252d1c/src/CMakeLists.txt#L637

it is worth to apply here the same test on `USE_NYQUIST` macro, as it has been already done elsewhere.

Recommended:


crsib (approved):

Paul-Licameli (changes requested):
Please cherry pick the last commit onto recent master and force push this branch with only that commit.  There are 14 extraneous merge commits before it.

crsib:
We can just squash it on merge

carlo-bramini:
> We can just squash it on merge

I would really appreciate that... for doing this quick fix, I used GitHub web editor and those extraneous commits are made by the "Sync fork" button on the web interface.

crsib:
I'll do that, no worries

crsib:
Given that this change is not visible to out QA team, I'll go ahead and merge it

Issue #4169, opened 2023-01-11
The realtime effect settings of a duplicated track are not independent


DavidBailes:
### Steps to reproduce

1. Create an audio track.
2. Add a realtime effect
3. Duplicate the audio in the track.
4. Open the dialog for the realtime effect for the original track, and change the value of a parameter.
5. Open the dialog for the realtime effect for the duplicated track, and observe the value of the parameter changed in the previous step

### Expected behavior

The values for the parameters for the two tracks should be different.

### Actual behavior

The values for the parameters for the two tracks are the same.

### Audacity Version

3.7

### Operating system

Windows, MacOS, Linux

Paul-Licameli:
> Testing against PR #4325 on Windows and macOS and the issue is fixed now.

Yes but there is a good objection from @crsib in the code review about other consequences of the change for undo and redo behavior.

chinakov:
Tested: 
- Custom VST and VST3 effects are independent
- Built-in effects are independent

PR #4325, opened 2023-02-17
Make realtime effect settings of duplicated track independent...

Paul-Licameli:
... Note that RealtimeEffectList::Clone() was correct as originally written when it held states by unique_ptr, but then the change to shared_ptr introduced this problem.

Resolves: #4169

Recommended:

### DIFF
--- a/libraries/lib-realtime-effects/RealtimeEffectList.cpp
+++ b/libraries/lib-realtime-effects/RealtimeEffectList.cpp
@@ -25,7 +25,7 @@ std::unique_ptr<ClientData::Cloneable<>> RealtimeEffectList::Clone() const
 {
    auto result = std::make_unique<RealtimeEffectList>();
    for (auto &pState : mStates)
-      result->mStates.push_back(pState);
+      result->mStates.push_back(pState->Clone());
    result->SetActive(this->IsActive());
    return result;
 }
--- a/libraries/lib-realtime-effects/RealtimeEffectState.cpp
+++ b/libraries/lib-realtime-effects/RealtimeEffectState.cpp
@@ -318,9 +318,14 @@ RealtimeEffectState::RealtimeEffectState(const PluginID& id)
    BuildAll();
 }
 
-RealtimeEffectState::~RealtimeEffectState()
-{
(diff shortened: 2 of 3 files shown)


libraries/lib-realtime-effects/RealtimeEffectList.cpp
```diff
@@ -25,7 +25,7 @@ std::unique_ptr<ClientData::Cloneable<>> RealtimeEffectList::Clone() const
 {
    auto result = std::make_unique<RealtimeEffectList>();
    for (auto &pState : mStates)
-      result->mStates.push_back(pState);
+      result->mStates.push_back(pState->Clone());
```
crsib, line 28: I wonder if it breaks the undo/redo stack. I remember that shallow copy was used there for a reason. @dozzzzer can you please check?

Paul-Licameli: A good point.

Define what "break" means exactly, but there may be some other visible change of behavior.  This code will also be visited when pushing undo items.

Undo items in memory may occupy more storage.

The state of sliders in dialogs, although not causing an undo push or pop when you change them directly, will be snap-shotted whenever the undo stack is pushed for other reasons, and restored to previous state when you undo.

I had raised the question whether this would be desired and intended undo/redo behavior or not.  It was decided, not.  But now, this change would make it so.

So there may be more work to pass more information about how `Clone()` is reached, and change its behavior only in the case you get there from the Duplicate editing command.

Paul-Licameli: Consider this PR just draft, then.  The good new is that the cause of the reported bug is known and easily fixable but for the other side effects of the fix.

crsib: As far as I remember the problem is that UI state is "pinned" to a specific `RealtimeEffectState`. So at some point UI modifies the state which is stored in the history, not the one used by the processor. 

This might be no longer a problem, as we've made a lot of changes on top. We need to regress all effect families carefully, specifically when using Undo/Redo with UI open.

PR #4336, opened 2023-02-20
Import/export preliminaries

vsverchinsky:
Preliminary work for import/export, extracted from #3879, mostly without changes except conflicts resolution. 

Recommended:


Paul-Licameli:
Can you explain why these cherry picks are important to unblock the rest of the work?

This PR mostly duplicates #3804 , which is a draft that I need to make final and submit to a long and difficult review.

vsverchinsky:
> Can you explain why these cherry picks are important to unblock the rest of the work?
> 
> This PR mostly duplicates #3804 , which is a draft that I need to make final and submit to a long and difficult review.

Preferences extraction and plugin un-registration will make other work a bit easier. And it will be much harder to resolve conflicts later.

PR #4343, opened 2023-02-21
Implemented an initial version of the tools toolbar in QML

ksoze95:

Recommended:


crsib (changes requested):
I think that on C++ side we should maintain the "code style" common to Audacity: that's it "CamelCase", not "camelCase".

Otherwise everything looks just fine

crsib (approved):

PR #4348, opened 2023-02-22
Windows linker fix for `StatefulEffect::MakeInstance`. 

hugofloresgarcia:
This is a temporary fix for a linker error we're running into on windows. When building `mod-deep-learning` the windows linker seems to be unable to link to the definition of `StatefulEffect::MakeInstance`if the function is defined in the `Effect.cpp` file. Moving the definition to the same place as the declaration (i.e. `Effect.h`) fixes the linker error, but the whole situation is fishy. The linker IS successfully linking the rest of Effect.h, but why not this specific function? 

Recommended:


hugofloresgarcia:
oops! just realized i meant to open this in my own fork of audacity, nevermind this....

PR #4352, opened 2023-02-23
Shelf Filter added

SteveDaulton:
Resolves: *Absence of a shelf filter in Audacity*

*Audacity has High-pass, Low-pass, and a Notch filter, but notable by its absence there is no shelf filter. This pull requests adds the missing filter. The filter can be set for High-shelf or Low-shelf.*

Recommended:

### DIFF
--- a/plug-ins/ShelfFilter.ny
+++ b/plug-ins/ShelfFilter.ny
@@ -1,0 +1,34 @@
+$nyquist plug-in
+$version 4
+$type process
+$preview linear
+$name (_ "Shelf Filter")
+$debugbutton disabled
+$author (_ "Steve Daulton")
+$release 2.4.0
+$copyright (_ "GNU General Public License v2.0")
+
+;; License: GPL v2
+;; http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+;;
+;; For information about writing and modifying Nyquist plug-ins:
+;; https://wiki.audacityteam.org/wiki/Nyquist_Plug-ins_Reference
+
+
+$control type (_ "Filter type") choice (("Low" (_ "Low-shelf"))
+                                        ("High" (_ "High-shelf"))) 0
+$control hz (_ "Frequency (Hz)") int "" 1000 10 10000
+$control gain (_ "Amount (dB)") int "" -6 -72 72
+
+
+(cond ((> hz (/ *sound-srate* 2))
+          (format nil (_ "Error.~%Frequency set too high for selected track.")))
+      ((> hz (/ *sound-srate* 2.1))  ;Handle edge case close to Nyquist frequency.
+          (setf *track* (force-srate (* 2 *sound-srate*) *track*))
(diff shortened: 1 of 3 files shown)


petersampsonaudacity:
Testing on W10 with @SteveDaulton 's branch build

The Shelf Filter appears in the Effects menu - _**BUT**_ it appears in the below the line scion and _**not**_ in the **EQ and Filters** section of the Effect menu:
image

SteveDaulton:
> _**BUT**_ it appears in the below the line section and _**not**_ in the **EQ and Filters**

In your `EffectsMenuDefaults.xml` file, you need to add `<Effect>Shelf Filter</Effect>` to whichever `<Group>` you want the effect to appear in.

When I built from source, the `EffectsMenuDefaults.xml` file was NOT copied into the build folder, so I'm guessing that menu sorting is a work  in progress, probably with a view to making it user configurable. The `EffectsMenuDefaults.xml` file was recently moved from `presets/EQDefaultCurves.xml` to `resources/EQDefaultCurves.xml`, which may be why it wasn't copied using the build steps given in `BUILDING.md`.

I agree that the `EQ and Filters` group would be a logical place to put this effect in the default grouping.

SteveDaulton:
> which may be why it wasn't copied using the build steps given in `BUILDING.md`,

I've asked Dmitry about this. He said: _"I think it is only copied on install"_

I've not set a default menu group - as it says in the pull request template: `There are no behavior changes unnecessary for the stated purpose of the PR`.

plug-ins/ShelfFilter.ny
```diff
@@ -0,0 +1,34 @@
+$nyquist plug-in
+$version 4
+$type process
+$preview linear
+$name (_ "Shelf Filter")
+$debugbutton disabled
+$author (_ "Steve Daulton")
+$release 2.4.0
```
Paul-Licameli, line 8: Is this line correct?

Paul-Licameli (approved):

plug-ins/ShelfFilter.ny
SteveDaulton: I'm not sure which line you are referring to, but lines 5 to 8 look fine.

The "release" version number refers to the plugin, not to the version of Audacity. It doesn't really matter what number is used so long as each subsequent update is higher than the previous. All of my plugins use the earliest Audacity version that I tested the plugin with as the initial "release" number, and then increments from that point. If there's any change to the plugin, the "release" number should be bumped.

Paul-Licameli: I meant line 8 and this satisfies me.

dozzzzer:
Tested on Windows and macOS

petersampsonaudacity:
I have pre-prepared the documentation for this - including the zebra-table on the EGATs page in the Manual.

Once this gets pulled all I need to do is switch to the new Zebra-table

Issue #2530, opened 2022-02-08
Overdubbing on Windows fails with WASAPI with default Recording device


petersampsonaudacity:
**Describe the bug**
Overdubbing on Windows fails with WASAPI host with default Recording device

**To Reproduce**
1. Clear Audacity settings folder (to ensure overdub is "on" and  Detect dropouts are both "on"
2.  Set host to WASAPI - leave Recording device as the default Mic
3.  Press Record and record a few seconds of audio
4. Observe: recording works fine
5. Press Stop
6. Press Shift+Record (to overdub record)
7. Observe: error message re. invalid sample rate (see Screenshot)

This does **_not_** fail if you set the recording device to be Loopback instead of the default mic.

This does **_not_** happen on W10 if the host is set to be either MME (default) or Windows DirectSound

**Expected behavior**
Since the first track records fine with WASAPI host (step 4)  I fully expect the overdub recording to work fine too (steps 6-7)

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/152988970-54aeaa56-c08e-44d7-9c23-c296f74261c7.png)

**Additional information (please complete the following information):**
 - OS: Windows 10  _(I will try to test later on Mac)_
 - Version: latest master alpha audacity-win-3.2.0-alpha-20220208+2f132fd-x64-msvc2022
 - Sound Device Windows WASAPI, onboard Realtek High Definition Audio(SST)  Sound card - default Mic input

**Additional context**
No problem on Mac.

This does not appear to be a regression this is the same all the way back to 2.1.0 where WASAPI with multiple input devices was added _(in late series 2.0.x WASAPI was first added but the only host available was loopback)_.

petersampsonaudacity:
Curiously you get the same error message if you try to record with WASAPI as the host and default mic input device - with 
a)  overdub turned off and 
b) Software Playthrough turned on

Penikov:
I tried to reproduce this bug recently. But overdub with Dropout detection works fine on my PC with WASAPI host in Audacity.
But I got the same error if one of the i/o devices has re-connected by Windows for some reason. In my case, this is a USB driver issue. 

Reloading Audacity or [Transport]-[Rescan Audio devices] always helps me 
![image](https://user-images.githubusercontent.com/25851359/153639370-249f8b0e-f8b7-4ad6-a89e-f2c757c4976d.png)

LWinterberg:
Say @petersampsonaudacity does this also occur for you when recording in mono?

[...]

PR #4353, opened 2023-02-23
Issue 2530: Special case added for WASAPI bidirectional audio streams

agguzman:
Resolves: https://github.com/audacity/audacity/issues/2530

Special case added to AudioIO::StartPortAudioStream() to solve Issue #2530 by enabling WASAPI automatic Sample Rate Conversion for the playback device. Automatic SRC is enabled only for bidirectional audio streams if there is no supported sample rate common to input and output devices.

Small fix in AudioIOBase::GetSupportedPlaybackRates() and GetSupportedCaptureRates() when a valid rate parameter is provided (different than 0).

Recommended:

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -82,6 +82,9 @@ time warp info and AudioIOListener and whether the playback is looped.
 #endif
 
 #include "portaudio.h"
+#ifdef __WXMSW__
+#include "pa_win_wasapi.h"
+#endif
 
 #if USE_PORTMIXER
 #include "portmixer.h"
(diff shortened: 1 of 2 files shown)


libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -85,6 +85,9 @@ time warp info and AudioIOListener and whether the playback is looped.
 #endif
 
 #include "portaudio.h"
+#ifdef __WXMSW__
```
Paul-Licameli, line 88: This correctly makes the `#include` directive conditional on the platform.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -539,6 +552,7 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
    bool usePlayback = false, useCapture = false;
    PaStreamParameters playbackParameters{};
    PaStreamParameters captureParameters{};
+   PaWasapiStreamInfo wasapiStreamInfo{};
```
Paul-Licameli, line 555: This type is not known except when compiling for Windows.  So `#ifdef __WXMSW__` and `endif` should surround this line too, and other lines below that do things with this variable.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -593,6 +621,14 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions &options,
       if( captureDeviceInfo == NULL )
          return false;
 
+      const PaHostApiInfo* hostInfo = Pa_GetHostApiInfo(captureDeviceInfo->hostApi);
+      bool isWASAPI = (hostInfo && hostInfo->type == paWASAPI);
```
Paul-Licameli, line 630: The constant `paWASAPI` is defined even when not compiling windows, so you will not need a change here to fix the build.

crsib (approved):

Issue #4201, opened 2023-01-17
In Noise Reduction effect, a Sensitivity setting of 0 disables the effect


billw58:
### Bug description

Since a sensitivity setting of 0 disables the effect, it should not be offered as an option.

### Steps to reproduce

1) Import audio appropriate for testing the NR effect (I used an acoustic guitar chord fading out into the vinyl noise)
2) Duplicate the track (for later comparison)
3) Select a noise-only region and Get Noise Profile
4) Select all of second track and apply Noise Reduction with settings;   Reduction: 12, Threshold: 0, Smoothing: 0
5) Select the noise region selected in step three and set that as the loop region
6) Solo each track back and forth while looping the noise region
Note that noise reduction is not applied
7) To confirm, select the noise region in each track and do Effect > Amplify and look at the amplification amount

### Expected behavior

Noise reduction should be applied.

### Actual behavior

Sensitivity = 0, no NR
Sensitivity = 1, 3 dB NR, audible artifacts
Sensitivity = 2, 8 dB NR, audible artifacts
Sensitivity = 3, 12 dB NR, almost no audible artifacts
Sensitivity =3, Smoothing = 3,  12 dB NR, no audible artifacts

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

macOS

LWinterberg:
Hm. there's a handful of effects which can be configured as a no-op: Amplify by 0dB, change pitch/speed/tempo by 0, EQ/basstreble, delay, maybe more. Though I guess it's a bit more opaque that this effect would be a no-op than in most other effects which also have this

SteveDaulton:
Setting "Sensitivity" to 0 also disables the effect.

Personally I don't find that "Threshold = 0 dB = no noise reduction" is particularly surprising. It seems logical to me in the same way that "Amplify by 0dB, change pitch/speed/tempo by 0, EQ/basstreble, delay" are no-ops, but setting "Sensitivity" to minimum being a no-op seems both surprising and totally pointless.

billw58:
Sorry, throughout this discussion I should have said "Sensitivity" instead of "Threshold". I was thinking about what the setting apparently did, which was raise the threshold (the border between noise and signal). I'll try to edit the title

Sririram:
is the issue solved?

Crazycoder122:
I would like to contribute in this issue. 

Also found the file (https://github.com/audacity/audacity/blob/master/src/effects/NoiseReduction.cpp)

[...]

PR #4354, opened 2023-02-23
Fix issue 4201

SteveDaulton:
When 'Sensitivity' is zero, the effect is unexpectedly a no-op. The solution is to limit the Sensitivity slider to a range that works. As the widget has (very reasonably) 2 decimal places, and a Sensitivity value of 0.01 has a very subtle effect, 0.01 is a good choice as the minimum Sensitivity.

Resolves: #4201 (https://github.com/audacity/audacity/issues/4201)*

*Set minimum Sensitivity to 0.01*

Recommended:

### DIFF
--- a/src/effects/NoiseReduction.cpp
+++ b/src/effects/NoiseReduction.cpp
@@ -1275,7 +1275,7 @@ const ControlInfo *controlInfo() {
          0.0, 48.0, 48, wxT("%d"), true,
          XXO("&Noise reduction (dB):"), XO("Noise reduction")),
          ControlInfo(&EffectNoiseReduction::Settings::mNewSensitivity,
-         0.0, 24.0, 48, wxT("%.2f"), false,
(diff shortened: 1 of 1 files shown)


SteveDaulton:
I'm closing this for now as I'm not convinced that Sensitivity settings below 1.0 are useful.
If anyone is interested in testing, I'd be interested to hear if anyone can provide a test case where Sensitivity < 1.0 is useful.

SteveDaulton:
After more extensive testing, I've found that while Sensitivity values in the range 3 to 6 tend to be best in most cases, there are some situations where very low values (even as low as 0.01) can produce better, less damaging results.
A Sensitivity setting of zero is always a waste of time as it causes the effect to do nothing.

I'm therefore reopening this pull request.

Paul-Licameli (approved):

PR #4363, opened 2023-02-27
Extract lib label track

Paul-Licameli:
Resolves: #5446

A toolkit neutral library for label tracks, including their import and export, but none of the graphics.

Also:  move mutable display related fields out of LabelStruct, like what was done in #4364 

Also:  fix two places that improperly mutated layout information during drawing.  Better to keep layout update separate from painting.

QA:  test label creation, editing, import and export;  test again, creation of labels for Find Clipping and Vamp analyzers

Test changes of label font propagate correctly

Test that updates of label box widths for change of text are correct, with various fonts

Confirm this old bug fix persist:  When a label is not a point, then its start and end verticals are at least one pixel distant, even zoomed out

Recommended:


crsib:
Does it block #1627? It waits for review for a really unreasonable amount of time now.

Paul-Licameli:
> Does it block #1627? It waits for review for a really unreasonable amount of time now.

I just tried to rebase the single commit in that PR onto this one, and there was only one easy conflict in the `#include` lines at top.

That one is itself not blocking anything.

vsverchinsky (approved):

Paul-Licameli:
@vsverchinsky thanks for the review, but more commits will be pushed onto this.  It's incomplete.

Paul-Licameli:
This is a draft until I review #1627 and see if I can resolve conflicts with that

PR #4364, opened 2023-02-27
Extract lib note track

Paul-Licameli:
Resolves: #5445

Toolkit neutral library for saving and loading MIDI data, and MIDI playback.

But not only the minimal work to enable the library to build:

Also moving mutable members of NoteTrack, used only in display, into ClientData attachments
managed in the proper UI layer, like what was done for WaveTrack at 7caf7b02.

QA: Basically, all things MIDI...
- [x] Import and export of MIDI
- [x] Display of MIDI (resizing track vertically, scrolling left and right, changing vertical ruler bounds)
- [x] Playback of MIDI (including mute, solo, and the buttons that mute individual channels -- left and right clicks)
- [x] Persistency of MIDI tracks in projects

Playback should simply work on Windows.

If you haven't yet, you might find SimpleSynth64 for Mac, packaged by our own old team member Leland Lucius, which will enable testing on that platform too.

Recommended:


src/NoteTrack.cpp
```diff
@@ -887,7 +887,7 @@ void NoteTrack::SetTopNote(int note)
    mTopNote = note;
 }
 
-void NoteTrack::SetNoteRange(int note1, int note2)
+void NoteTrack::SetNoteRange(int note1, int note2) const
```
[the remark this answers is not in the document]
Paul-Licameli: Yes!  See later commits that remove the mutable members.  I didn’t rest satisfied  with this.

vsverchinsky (approved):

cmake-proxies/portsmf/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 
-add_library( ${TARGET} STATIC )
+add_library( ${TARGET} OBJECT )
```
crsib, line 2: `STATIC` does not require `__declspec`, so I don't understand your motivation here. I don't think that the problem was identified correctly.

libraries/lib-registries/ClientData.h
```diff
@@ -45,14 +45,18 @@ template< typename Object > using BarePtr = Object*;
 template<
    typename Covariant = void, // CRTP derived class when not void
    template<typename> class Owner = UniquePtr
-> struct REGISTRIES_API Cloneable
+> struct
+#ifdef _WIN32
+__declspec(dllexport)
```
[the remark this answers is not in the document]
crsib: I am pretty sure that this is wrong. `__declspec(dllexport)` is not symmetrical, What happens here - every library that uses `Cloneable` now exports it.

Paul-Licameli:
https://github.com/audacity/audacity/pull/4364#discussion_r1133635250

Here is my reference for this.    https://docs.microsoft.com/en-us/cpp/cpp/defining-inline-cpp-functions-with-dllexport-and-dllimport?view=msvc-170
    
The linker knows how to remove the duplicate symbols and make only one definition.  In fact, a standard C++ linker that doesn't need the visibility annotations will do the same with template functions and importantly with any static variables defined inside those.  Much code using ClientData::Site already relies on that to be correct on Mac and Linux too.

Paul-Licameli:
Does @crsib wish to revoke the approval of this pull request?

I know I had to do some unusual things with libraries to make the link work.  One, using and OBJECT library.  Two, there are still the unusual `#include` paths to get allegro.h and I thought @crsib might want to propose a cleaner way to do that.

cmake-proxies/portsmf/CMakeLists.txt
Paul-Licameli: I recall the Windows build failing when I undid that (already a few years ago).  I could repeat the experiment and see if the build still fails without it.

crsib:
I don't see how it supports your reasoning. By unconditionally specifying `dllexport` you are forcing the symbol to be exported from every library that uses Cloneable. It is never imported and each library uses it's own instantiation. Linker cannot remove duplication when it happens in different link units.

Paul-Licameli:
https://github.com/audacity/audacity/actions/runs/4404023816

Watch that build to see whether it works with portsmf as a STATIC library

crsib:
You are marking the class as dllexport. Instead, MSDN says that you need to "explicitly instantiate the class template and mark this explicit instantiation with dllexport".

Paul-Licameli:
Yes, I should read the page to the end.  (Maybe the page was updated since years ago when I first had this problem?). It specifically mentions the CRTP pattern (doesn't call it that) in the last two code examples.

I know the build succeeded as I did it, but I can also try that alternative.  Put  the declspec on the derived class.

Paul-Licameli:
> https://github.com/audacity/audacity/actions/runs/4404023816
> 
> Watch that build to see whether it works with portsmf as a STATIC library

That build did succeed.  But will it be correct?  Now there are two staticlly linked copies of portsmf and I don't know if there are duplicated global variables to worry about?

Paul-Licameli:
> Resolves: _(direct link to the issue)_
> 
> Toolkit neutral library for saving and loading MIDI data, and MIDI playback.
> 
> But not only the minimal work to enable the library to build:
> 
> Also moving mutable members of NoteTrack, used only in display, into ClientData attachments managed in the proper UI layer, like what was done for WaveTrack at [7caf7b0](https://github.com/audacity/audacity/commit/7caf7b02239edff5b26e57d3798304c86472be1e).
> 
> * [x]  I signed [CLA](https://www.audacityteam.org/cla/)
> * [x]  The title of the pull request describes an issue it addresses
> * [x]  If changes are extensive, then there is a sequence of easily reviewable commits
> * [x]  Each commit's message describes its purpose and effects
> * [x]  There are no behavior changes unnecessary for the stated purpose of the PR
> 
> Recommended:
> 
> * [x]  Each commit compiles and runs on my machine without known undesirable changes of behavior

QA:  Test not only the display, but also persistency, and MIDI playback, for any regressions in behavior.

vsverchinsky (approved):
The major issue with `header-substitutes/allegro.h` is the need to apply path relative to the library or module where you reference it. Maybe we could move it to `lib-sr/portsmf/include/allegro_wrapped.h` or `cmake-proxies/portsmf/allegro_wrapped.h`?

Paul-Licameli:
Rebased onto recent master with some conflict resolution.

The problem with allegro.h still isn't solved satisfactorily.

I removed the commit that made portsmf an OBJECT library and the later reversion of that.

Paul-Licameli:
This old PR has been rebased onto recent master, and I made other changes.  I think I don't need any of the tricks with `__declspec` after all.  But I will know for certain when this build completes.

Paul-Licameli:
Please review.  The most recent build failures were only CI glitches.  Reruns succeeded.

Paul-Licameli:
Another force-push and conflict resolutions with recent master.

src/NoteTrack.cpp
```diff
@@ -906,10 +905,26 @@ bool NoteTrack::ExportMIDI(const wxString &f) const
...
 }
 
+EnumSetting<bool> NoteTrack::AllegroStyleSetting{
+   wxT("/FileFormats/AllegroStyleChoice"),
+   {
+      EnumValueSymbol{ wxT("Seconds"), XXO("&Seconds") },
+      /* i18n-hint: The music theory "beat" */
+      EnumValueSymbol{ wxT("Beats"), XXO("&Beats") },
+   },
+   0, // true
```
crsib, line 793: But is it ... true?

crsib (approved):

PR #4365, opened 2023-02-28
Break dependencies on ffmpeg and mp3

Paul-Licameli:

Some dependency breaking before other import and export refactoring into modules

QA:  Test that all works as before in the Libraries page of the Preferences dialog.

Verify that the error message shows as before, when you try to import a file type that requires FFmpeg, but that is not installed on the system.

Recommended:

### DIFF
--- a/libraries/lib-registries/Registry.h
+++ b/libraries/lib-registries/Registry.h
@@ -233,7 +233,24 @@ namespace Registry {
    // a GroupItem.
    REGISTRIES_API
    void RegisterItem( GroupItem &registry, const Placement &placement,
-      BaseItemPtr pItem );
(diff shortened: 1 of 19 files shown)


crsib (approved):
I'm not sure if we need MP3 panel, but let's keep the behavior for now

vsverchinsky (approved):

Paul-Licameli:
Rebased onto more recent master.  No conflicts, no changes.

Ryan-Miller-Audacity:
Confirmed on both windows and Mac.

Paul-Licameli:
Some dependency breaking before other import and export refactoring into modules

PR #4366, opened 2023-02-28
Position independent code for GNU compiles of some 3party libs...

Paul-Licameli:
... so that they can be linked into MODULE type libraries, when later we make extractions.

But following the example of 824ffe95, just do it for all platforms, though not needed to make modules on macOS or Windows.

(Also improve some CMake code to assign, not append list variables so that they aren't contaminated by the calling environment.)

Recommended:


crsib (approved):

Paul-Licameli:
Should this skip QA?

crsib:
I think so

PR #4369, opened 2023-02-28
Extract lib time track

Paul-Licameli:

Toolkit-neutral library for TimeTrack

Recommended:


src/tracks/timetrack/ui/TimeTrackView.cpp
```diff
@@ -156,9 +158,26 @@ void TimeTrackView::Draw(
...
    if ( iPass == TrackArtist::PassTracks ) {
+      const auto pTrack = FindTrack();
+      auto pList = pTrack->GetOwner();
+      if (!pList)
+         // Track isn't owned by a list.  Can't proceed!
+         return;
+      if (!pList)
+         // List isn't owned by a project.  Can't proceed!
+         // But this shouldn't happen when drawing it
+         return;
```
vsverchinsky, line 169: One of these checks is redundant, or maybe something is missing here?

src/tracks/timetrack/ui/TimeTrackView.cpp
```diff
@@ -156,9 +158,26 @@ void TimeTrackView::Draw(
...
+
+      auto pProject = pList->GetOwner();
+      auto &zoomInfo = ViewInfo::Get(*pProject);
+      LinearUpdater updater;
+      updater.SetData(&zoomInfo);
+      Ruler ruler{ updater, TimeFormat::Instance() };
+      ruler.SetLabelEdges(false);
```
vsverchinsky, line 176: I see that `Ruler` does some cacheing, wouldn't it be better to make `LinearUpdater` and `Ruler` a member variable of `TimeTrackView`?

src/tracks/timetrack/ui/TimeTrackView.cpp
Paul-Licameli: Thanks I think I meant `!pList->GetOwner()`

src/tracks/timetrack/ui/TimeTrackView.cpp
Paul-Licameli: I debated this with myself.  But I think this ruler isn't saving any important state and the calculations are not large.  I may change my mind.

Paul-Licameli:
I fixed the mistaken null check, and I have added a comment, justifying the use of a temporary stack allocated Ruler.  Please review again.

vsverchinsky (approved):

PR #4375, opened 2023-03-01
Fixes DebugBreak declaration in Windows builds

crsib:

Recommended:

### DIFF
--- a/libraries/lib-strings/Internat.h
+++ b/libraries/lib-strings/Internat.h
@@ -56,9 +56,7 @@ extern STRINGS_API const wxString& GetCustomSubstitution(const wxString& str1 );
 
    #ifdef __WXMSW__
 
-   // Eventually pulls in <windows.h> which indirectly defines DebugBreak(). Can't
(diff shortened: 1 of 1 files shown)


libraries/lib-strings/Internat.h
```diff
@@ -58,7 +58,7 @@ extern STRINGS_API const wxString& GetCustomSubstitution(const wxString& str1 );
 
    // Eventually pulls in <windows.h> which indirectly defines DebugBreak(). Can't
    // include <windows.h> directly since it then causes "MemoryX.h" to spew errors.
-   #include <wx/app.h>
+   extern "C" __declspec(dllimport) void __stdcall DebugBreak();
```
Paul-Licameli, line 61: Simple enough, but then, delete the comment that precedes.

crsib: Hmm, one sec

crsib: Force pushed without a comment

Paul-Licameli (approved):

Issue #3863, opened 2022-10-27
Import / Export refactoring


AnitaBats:
Goals:
- Move generic UI out of `ExportPlugin` implementations
- Extract import/export code into `lib-import-export`, put implementations into `mod-*`
- Export should not use global state

Refactoring steps:

- [x] Preliminary work: cleanup and simplify interfaces (move `ChooseFormat` and `NewWaveTrack`, remove `ExportPlugins::mFormatInfos` etc.).
- [x] Eliminate dependency on `ProgressDialog`.
- [x] Extract `ExportFileDialog` from `GetFilename`.
- [x] Move dialogs invocation to caller side (`ExportMixerDialog`, `TagsEditorDialog` etc.)
- [x] Remove `AudacityMessageBox` from exporters and importers.
- [x] Plugin expose a list of parameter descriptions with their current values.
- [x] Do export using parameters passed as argument.
- [x] UI uses new export parameters API, eliminate `OptionsCreate`.
- [x] Extract `lib-import-export`
- [x] Run export in a dedicated thread.

Paul-Licameli:
Can this issue have more explanation of the scope of the task?

I do not yet understand how this task relates the replatforming to Qt.  But I don't know enough about the thinking behind it.

crsib:
Import/Export has major problems:

* UI is **extremely** tied to the processing. For example, `Import` returns `ProgressResult` which is insane. Implementations do not handle it carefully at all.
* Not only the current framework is stateful - it uses hidden **global** state with no way to access the state without the hacks. This makes any usage of the framework besides the menu items to be overcomplicated and error prone. 

In general, this task is not about helping the Qt migration, it is about making Import/Export to be manageable and usable. However, it is a must for the Qt migration.

The detailed decomposition is expected quite soon.

ianalarina:
will be resolved in scope of #4317

PR #4384, opened 2023-03-03
Import/export refactoring

vsverchinsky:
Resolves: #3863

Recommended:


Paul-Licameli:
Excellent!  But still a draft.  What is missing?

vsverchinsky:
> Excellent! But still a draft. What is missing?

Export parameters should replace global prefs and break dependency on ShuttleGUI, direct invocations of message dialogs still happening during export process, import and export process is still happening in the main thread ... I'll update parent ticket so it would be easier to understand the progress.

Paul-Licameli:
Failing builds

src/export/Export.cpp
```diff
@@ -90,114 +90,15 @@ bool ExportPlugin::CheckFileName(wxFileName & WXUNUSED(filename), int WXUNUSED(f
...
-   bool isext = false;
-   for (int i = index; i < GetFormatCount(); i = GetFormatCount())
+   if(index >= 0 && index < GetFormatCount())
    {
-      const auto &defext = GetExtension(i);
-      const auto &defexts = GetExtensions(i);
-      int indofext = defexts.Index(ext, false);
-      if (defext.empty() || (indofext != wxNOT_FOUND))
-         isext = true;
+      auto formatInfo = GetFormatInfo(index);
```
Paul-Licameli: auto &  ?

src/export/Export.cpp
```diff
@@ -90,114 +90,15 @@ bool ExportPlugin::CheckFileName(wxFileName & WXUNUSED(filename), int WXUNUSED(f
...
-{
-   return mFormatInfos[index].mMaxChannels;
-}
-
-bool ExportPlugin::GetCanMetaData(int index)
-{
-   return mFormatInfos[index].mCanMetaData;
-}
-
 bool ExportPlugin::IsExtension(const FileExtension & ext, int index)
```
Paul-Licameli: `const` function?

src/export/Export.cpp
```diff
@@ -418,9 +319,11 @@ bool Exporter::Process(bool selectedOnly, double t0, double t1)
    if (!CheckMix()) {
       return false;
    }
+   
+   auto exportFormatInfo = mPlugins[mFormat]->GetFormatInfo(mSubFormat);
```
Paul-Licameli, line 144: Again, use `&`?

src/export/Export.cpp
```diff
@@ -478,7 +381,8 @@ bool Exporter::Process(
       ++i;
       for (int j = 0; j < pPlugin->GetFormatCount(); j++)
       {
-         if (pPlugin->GetFormat(j).IsSameAs(type, false))
+         auto formatInfo = pPlugin->GetFormatInfo(j);
```
Paul-Licameli, line 198: again

src/export/Export.cpp
```diff
@@ -603,9 +507,9 @@ bool Exporter::GetFilename()
          ++i;
          for (int j = 0; j < pPlugin->GetFormatCount(); j++)
          {
-            auto mask = pPlugin->GetMask(j);
-            fileTypes.insert( fileTypes.end(), mask.begin(), mask.end() );
-            if (mPlugins[i]->GetFormat(j) == defaultFormat) {
+            auto formatInfo = pPlugin->GetFormatInfo(j);
```
Paul-Licameli: again

src/export/Export.cpp
```diff
@@ -603,9 +507,9 @@ bool Exporter::GetFilename()
...
+            fileTypes.insert(fileTypes.end(), { formatInfo.mDescription, formatInfo.mExtensions });
```
Paul-Licameli: Correctness of this change assumes `SetMask` was nowhere used -- and that is true

src/export/ExportCL.cpp
```diff
@@ -343,15 +346,18 @@ class ExportCL final : public ExportPlugin
...
-   SetMaxChannels(255,0);
-   SetCanMetaData(false,0);
-   SetDescription(XO("(external program)"),0);
+   return 1;
+}
+
+FormatInfo ExportCL::GetFormatInfo(int) const
+{
+   return {
+      wxT("CL"), XO("(external program)"), {""}, {}, 255, false
```
Paul-Licameli: The `{}` here and elsewhere initializes `mMask` but that member is nowhere used, so simplify

src/export/ExportPCM.cpp
```diff
@@ -409,29 +408,44 @@ class ExportPCM final : public ExportPlugin
...
-   SetDescription(XO("Other uncompressed files"), selformat);
-   SetCanMetaData(true, selformat);
-   SetMaxChannels(255, selformat);
+      return {
+         sf_header_shortname(si.format),
+         XO("Other uncompressed files"),
+         { sf_header_extension(si.format) },
+         {},
+         static_cast<unsigned>(si.channels),
+         true
```
Paul-Licameli, line 440: Not the literals "LIBSNDFILE" and `255` as before.  Is this right?

Paul-Licameli: Commit comment should mention that you are correcting this case, not just refactoring

src/export/ExportPCM.cpp
```diff
@@ -409,29 +408,44 @@ class ExportPCM final : public ExportPlugin
...
-   }
+      SF_INFO si = {};
+
+      si.format = LoadOtherFormat() & SF_FORMAT_TYPEMASK;
+      si.format |= LoadEncoding(si.format);
+
+      for (si.channels = 1; sf_format_check(&si); si.channels++)
+      {
+         // just counting
+      }
```
Paul-Licameli, line 433: Did you neglect `--si.channels;` after the loop?

src/export/ExportCL.cpp
Paul-Licameli: ... And so you did in the second commit.  Just squash it.

src/export/Export.cpp
```diff
@@ -66,13 +47,13 @@
...
 #include "FileNames.h"
 #include "widgets/HelpSystem.h"
 #include "widgets/ProgressDialog.h"
 #include "wxFileNameWrapper.h"
 
+#include "ExportMixerDialog.h"
+
 //----------------------------------------------------------------------------
 // ExportPlugin
 //----------------------------------------------------------------------------
```
Paul-Licameli: Move (or just delete) the banner comments too, for the moved classes

Paul-Licameli: The commit comment can also mention ExportPlugin

src/export/ExportMixerPanel.h
```diff
@@ -0,0 +1,54 @@
...
+
+///\class ExportMixerPanel
+///\brief Panel that displays mixing for advanced mixing option.
+class ExportMixerPanel final : public wxPanelWrapper
+{
+public:
+   ExportMixerPanel( wxWindow *parent, wxWindowID id,
+         MixerOptions::Downmix *mixerSpec, wxArrayString trackNames,
+         const wxPoint& pos = wxDefaultPosition,
+         const wxSize& size = wxDefaultSize);
```
Paul-Licameli, line 31: Not obvious in this large commit -- this is not just copied and pasted, but also the type of the third argument changes.  Please mention that in the commit comment.

Paul-Licameli: Oh but I see now -- same type, just another alias.  So never mind.

src/import/ImportProgressListener.h
```diff
@@ -0,0 +1,41 @@
...
+   enum class ImportResult {
+      Success,
+      Error,
+      Cancelled
+   };
+   
+   virtual ~ImportProgressListener() = default;
+   
+   ///Called by Importer when it attempts to import file using registered ImportPlugin instance
+   ///Could be called more than once, but for each call there will be a complimentary call to OnImportResult
```
Paul-Licameli, line 30: should be "complementary"

src/import/ImportProgressListener.h
```diff
@@ -0,0 +1,41 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  ImportProgressListener.h
+
+  V. Sverchinsky
+
+*//****************************************************************/
```
Paul-Licameli, line 9: As a matter of form I'd prefer that there is also a .cpp including this so we are sure the header has all its compilation dependencies, and the destructor should have a non-inline definition there.

src/ProjectFileManager.cpp
```diff
@@ -1204,6 +1207,55 @@ bool ImportProject(AudacityProject &dest, const FilePath &fileName)
...
+         importFileHandle.SetStreamUsage(0,TRUE);
+      return true;
+   }
+   
+   void OnImportProgress(double progress) override
+   {
+      constexpr double ProgressSteps { 1000.0 };
+      if(!mProgressDialog)
+      {
+         wxFileName ff( mImportFileHandle->GetFilename() );
```
Paul-Licameli, line 1238: Should I worry about the safety of that `->`? Should there be a shared or weak pointer?  Or should there be comments in the base class header explaining the assumptions that derived classes may make about lifetimes of `ImportFileHandle`?

src/import/ImportWavPack.cpp
```diff
@@ -228,20 +233,26 @@ ProgressResult WavPackImportFileHandle::Import(WaveTrackFactory *trackFactory, T
...
    if (WavpackGetNumErrors(mWavPackContext))
       AudacityMessageBox( XO( "Encountered %d errors decoding WavPack file!" ).Format( WavpackGetNumErrors(mWavPackContext) ),
                           XO( "WavPack Importer" ), wxOK | wxICON_EXCLAMATION | wxCENTRE);
 
-   if (updateResult != ProgressResult::Stopped && updateResult != ProgressResult::Cancelled
-         && totalSamplesRead < mNumSamples)
-      updateResult = ProgressResult::Failed;
-
-   if (updateResult == ProgressResult::Failed || updateResult == ProgressResult::Cancelled)
-      return updateResult;
```
Paul-Licameli, line 243: In the old logic, there was a distinction between stopping and cancelling.  In case of stopping, there was no error reported, even if `totalSamplesRead < mNumSamples`.

In the new logic, this changes.

Paul-Licameli: Progress dialogs had a Stop button distinct from Cancel in the previous version, but with this commit they don't.

Is this change intended?

src/import/ImportFFmpeg.cpp
```diff
@@ -522,12 +522,11 @@ ProgressResult FFmpegImportFileHandle::Import(WaveTrackFactory *trackFactory,
    }
    // This is the heart of the importing process
    // The result of Import() to be returned. It will be something other than zero if user canceled or some error appears.
-   auto res = ProgressResult::Success;
-
+   auto hasError {false};
```
Paul-Licameli, line 525: This is an unusual way to write it.

Funny C++ fact:  if you change it to
`auto hasError = {false};`

then instead they type is deduced as `std::initializer_list<bool>` and compilation fails because that is not convertible too bool.

src/import/ImportOGG.cpp
```diff
@@ -336,19 +340,22 @@ ProgressResult OggImportFileHandle::Import(
 
          samplesSinceLastCallback += samplesRead;
          if (samplesSinceLastCallback > SAMPLES_PER_CALLBACK) {
-            updateResult = mProgress->Update(ov_time_tell(mVorbisFile.get()),
-               ov_time_total(mVorbisFile.get(), bitstream));
+            progressListener.OnImportProgress(ov_time_tell(mVorbisFile.get()) / ov_time_total(mVorbisFile.get(), bitstream));
```
Paul-Licameli, line 343: another denominator to check for 0

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  ExportFileDialog.cpp
+
+  Dominic Mazzon
```
Paul-Licameli, line 7: Mazzoni

src/export/ExportPCM.cpp
```diff
@@ -33,6 +33,7 @@
 #include "../widgets/ProgressDialog.h"
 #include "../widgets/wxWidgetsWindowPlacement.h"
 #include "wxFileNameWrapper.h"
+#include "ExportFileDialog.h"
```
Paul-Licameli, line 36: I thought this was unnecessary for the build, but now I see it's needed for `AUDACITY_FILE_SUFFIX_EVENT`

That is one place where I omitted to change wxEvents to Observer messages, but then this is a case of wxWidgets dialogs communicating to each other.

src/import/ImportPCM.cpp
```diff
@@ -373,20 +385,16 @@ ProgressResult PCMImportFileHandle::Import(WaveTrackFactory *trackFactory,
...
 
-         updateResult = mProgress->Update(
-            framescompleted.as_long_long(),
-            fileTotalFrames.as_long_long()
-         );
-         if (updateResult != ProgressResult::Success)
-            break;
-
-      } while (block > 0);
+         progressListener.OnImportProgress(framescompleted.as_double() / fileTotalFrames.as_double());
```
Paul-Licameli, line 388: And one more denominator.

Maybe `OnImportProgress` should take numerator and denominator as two arguments, and you can write the check for zero division in one place only.

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
...
+                             const wxString& defaultName,
+                             const wxString& defaultFormatName,
+                             const TranslatableString& title,
+                             const wxPoint& pos,
+                             const wxSize& sz,
+                             const wxString& name)
+{
+   wxString formatName = defaultFormatName;
+   if(defaultFormatName.empty())
+      formatName = gPrefs->Read(wxT("/Export/Format"),
```
Paul-Licameli, line 119: It would be good to make a StringSetting now and not repeat the literal path string

src/export/Export.cpp
```diff
@@ -800,7 +471,6 @@ bool Exporter::ProcessFromTimerRecording(bool selectedOnly,
    mFilename = fnFile;
    mFormat = iFormat;
    mSubFormat = iSubFormat;
-   mFilterIndex = iFilterIndex;
```
Paul-Licameli, line 1042: With this deletion, iFilterIndex becomes unused.

I expected then that Timer record would achieve this side effect otherwise, by a call to configure the exporter, but I don't see it.

Paul-Licameli: If that is correct, then `m_iAutoExportFilterIndex` is a useless member of `TimerRecordDialog`.

Paul-Licameli: And if that is removed, then `Exporter::GetAutoExportFilterIndex()` will be called nowhere.

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
...
+                                         wxT("WAV"));
+   
+   wxFileName filename;
+   //Bug 1304: Set a default path if none was given.  For Export.
+   filename.SetPath(FileNames::FindDefaultPath(FileNames::Operation::Export));
+   filename.SetFullName(defaultName);
```
Paul-Licameli, line 125: I think this has `SetFullName` where the old version had `SetName` in `Exporter::GetFileName()`.  Why the change?

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
...
+   else
+      filename.SetFullName(defaultName);
+   
+   if (!filename.HasExt())
+   {
+      auto defext = plugins[pluginIndex]->GetFormatInfo(formatIndex).mExtensions[0].Lower();
+      filename.SetExt(defext);
+   }
+   
+   wxTabTraversalWrapper<FileDialog>::Create(parent,
```
Paul-Licameli, line 96: Why not `FileDialogWrapper` as in the original?

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
...
+                                             filename.GetFullName(),
+                                             FileNames::FormatWildcard(fileTypes),
+                                             wxRESIZE_BORDER | wxFD_SAVE | wxFD_OVERWRITE_PROMPT,
+                                             pos, sz, name);
+   
+   Bind(wxEVT_FILECTRL_FILTERCHANGED, &ExportFileDialog::OnFilterChanged, this);
+   SetUserPaneCreator(CreateUserPaneCallback, (wxUIntPtr) this);
+}
+
+int ExportFileDialog::Export(wxWindow* parent,
```
Paul-Licameli, line 108: `Export` is a misleading name.  This function only chooses a path before export really happens.

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
...
+            auto prompt =
+               XO("Are you sure you want to export the file as \"%s\"?\n")
+                  .Format( filename.GetFullName() );
+
+            int action = AudacityMessageBox(
+               prompt,
+               XO("Warning"),
+               wxYES_NO | wxICON_EXCLAMATION);
+            if (action != wxYES) {
+               return false;
```
Paul-Licameli, line 196: I think this should be `continue` to conserve the old logic

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
...
+            }
+         }
+
+         filename.SetExt(defext);
+      }
+      
+      if (!plugins[pluginIndex]->CheckFileName(filename, formatIndex))
+      {
+         return false;
```
Paul-Licameli, line 205: `continue` here too

src/export/ExportFileDialog.cpp
```diff
@@ -0,0 +1,345 @@
...
+            wxYES_NO | wxICON_EXCLAMATION);
+         if (action != wxYES) {
+            return false;
+         }
+      }
+
+      if (filename.GetFullPath().length() >= 256) {
+         AudacityMessageBox(
+            XO( "Sorry, pathnames longer than 256 characters not supported.") );
+         return false;
```
Paul-Licameli, line 225: `continue` here too

Paul-Licameli:
This round of review is complete!

src/export/ExportUtils.h
```diff
@@ -36,5 +39,6 @@ class ExportUtils final
    static void InitProgress(std::unique_ptr<BasicUI::ProgressDialog> &pDialog,
          const wxFileNameWrapper &title, const TranslatableString &message);
 
+   static TrackIterRange<const WaveTrack> FindExportWaveTracks(const TrackList& tracks, bool selectedOnly);
```
Paul-Licameli, line 42: Do you understand the difference between `SampleTrack` and `WaveTrack`?  The first is an abstract base class of the second.  (And `WritableSampleTrack` is between them.)

I think these utilities would be better defined in terms of the more general interface.  We don't yet have other concrete subclasses of `SampleTrack` but it may be useful later.

`SampleTrack` does not expose the notion of a clip within the track, but I don't see that you require knowledge of subdivision into clips.

src/export/ExportUtils.cpp
```diff
@@ -59,3 +54,13 @@ void ExportUtils::InitProgress(std::unique_ptr<BasicUI::ProgressDialog> &pDialog
    return InitProgress(
       pDialog, Verbatim( title.GetName() ), message );
 }
+
+//TODO: used in many places in anticipation that Exporter yields same result, fix that
```
Paul-Licameli, line 58: I don't understand the comment.  What "same result?"

src/export/ExportUtils.cpp
```diff
@@ -59,3 +54,13 @@ void ExportUtils::InitProgress(std::unique_ptr<BasicUI::ProgressDialog> &pDialog
...
+TrackIterRange<const WaveTrack> ExportUtils::FindExportWaveTracks(const TrackList& tracks, bool selectedOnly)
+{
+   bool anySolo = !(( tracks.Any<const WaveTrack>() + &WaveTrack::GetSolo ).empty());
+
+   return tracks.Any<const WaveTrack>()
+            + ( selectedOnly ? &Track::IsSelected : &Track::Any  )
+            - ( anySolo ? &WaveTrack::GetNotSolo :  &WaveTrack::GetMute);
```
Paul-Licameli, line 65: `Track` which is the base of `SampleTrack` does expose selectedness.

But it is `PlayableTrack` that defines mute and solo and it is not a base of `SampleTrack`.

Yet you still might apply `Filter` to the track range to get a range of `SampleTrack`.

Maybe `PlayableTrack` should instead become a mix-on property of tracks that you discover with a `dynamic_cast`?  That's a bigger idea I should think about for future.

Paul-Licameli: Feel free to ignore all the suggestions about `SampleTrack` for a draft.  But remember the idea.

src/export/Export.cpp
```diff
@@ -385,22 +373,26 @@ bool Exporter::CheckMix(bool prompt /*= true*/ )
...
-                           exportedChannels,
-                           NULL,
-                           1,
-                           XO("Advanced Mixing Options"));
-      if (prompt) {
-         if (md.ShowModal() != wxID_OK) {
+      auto exportTracks = ExportUtils::FindExportWaveTracks(TrackList::Get( *mProject ), mSelectedOnly);
+      mMixerSpec = std::make_unique<MixerOptions::Downmix>(exportTracks.size(),
+                                                           std::clamp(exportedChannels,
+                                                                      // STF (April 2016): AMR (narrowband) and MP3 may export 1 channel.
```
Paul-Licameli: No preservation of the other old comments from JKC? (James)

Paul-Licameli:
That's all for the first new commit

src/export/ExportMP3.cpp
```diff
@@ -1757,15 +1758,6 @@ bool ExportMP3::CheckFileName(wxFileName & WXUNUSED(filename), int WXUNUSED(form
    return true;
 }
 
-int ExportMP3::SetNumExportChannels()
-{
-   bool mono;
-   gPrefs->Read(wxT("/FileFormats/MP3ForceMono"), &mono, 0);
```
Paul-Licameli, line 1755: If you change lines like this, also take the opportunity to eliminate one more repeated string literal path with a `BoolSetting` object.

src/TimerRecordDialog.cpp
```diff
@@ -591,9 +591,21 @@ int TimerRecordDialog::ExecutePostRecordActions(bool bWasStopped) {
 
    // Do Automatic Export?
    if (m_bAutoExportEnabled) {
+      const auto& tracks = TrackList::Get(mProject);
       Exporter e{ mProject };
+      
+      bool skipSilenceAtBeginning;
+      gPrefs->Read(wxT("/AudioFiles/SkipSilenceAtBeginning"),
```
Paul-Licameli, line 611: Another opportunity to use `BoolSetting`.

But be careful:  this one is used as an argument to `TieCheckBox`.

When that happens, you also need to use `Invalidate` in the `Commit()` member function of the preference panel.

I intend to remove that need for `Invalidate`, but it requires first using a named Setting object in every call of the `Tie...` functions, in Preferences, and also in Export dialogs, and also in SoundActivatedRecord.cpp

src/menus/FileMenus.cpp
```diff
@@ -34,6 +34,35 @@
...
+bool ExportWithPrompt(AudacityProject &project,
+                      Exporter &exporter,
+                      const FileExtension& format,
+                      double t0, double t1, bool selectedOnly)
+{
+   bool skipSilenceAtBeginning;
+   gPrefs->Read(wxT("/AudioFiles/SkipSilenceAtBeginning"),
+                                   &skipSilenceAtBeginning, false);
+   
+   if(!exporter.SetExportRange(t0, t1, false, skipSilenceAtBeginning))
```
Paul-Licameli, line 52: Did you mean to pass `selectedOnly` for the third argument?

src/TimerRecordDialog.cpp
```diff
@@ -591,9 +591,21 @@ int TimerRecordDialog::ExecutePostRecordActions(bool bWasStopped) {
...
+                                      &skipSilenceAtBeginning, false);
+      
+      if(!e.SetExportRange(0, tracks.GetEndTime(), false, skipSilenceAtBeginning))
+      {
+         //But there should be at least one recorded track...
+         ShowExportErrorDialog(
+            ":576", XO("All audio is muted."), XO("Warning"), false);
```
Paul-Licameli, line 618: You lifted this UI out of the testing logic,but now it duplicates a string literal in two places.  Is there a good file in which to put a single occurrence of the literal?

src/menus/FileMenus.cpp
```diff
@@ -34,6 +34,35 @@
 // private helper classes and functions
 namespace {
 
+bool ExportWithPrompt(AudacityProject &project,
+                      Exporter &exporter,
+                      const FileExtension& format,
+                      double t0, double t1, bool selectedOnly)
+{
```
Paul-Licameli, line 47: This new function shows the file dialog after the setting of export range, but previous to this commit, the two things were in the opposite order (as they remain for timer recording).

So there may be an unintended, though unimportant, change in behavior.  Such as giving the warning dialog before any file selection.

Better to change no behavior unless really meaning to.

src/export/ExportPCM.cpp
vsverchinsky: Thanks!

src/ProjectFileManager.cpp
vsverchinsky: Aren't comments for `ImportProgressListener` enough?

src/import/ImportWavPack.cpp
vsverchinsky: I think that the major distinction between them is that the data that has been been read by the time Stop button was pressed is preserved, but mpg123 importer seem to be an exception. Will update this commit so that this behavior is preserved.

src/import/ImportAUP.cpp
```diff
@@ -525,14 +526,19 @@ bool AUPImportFileHandle::Open()
    return false;
 }
 
+void AUPImportFileHandle::Cancel()
+{
```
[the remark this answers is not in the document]
vsverchinsky: Moved them to `ImportPluginHandleEx`, though I don't want to convert it's type to atomic in advance

src/ProjectFileManager.cpp
```diff
@@ -1204,6 +1207,55 @@ bool ImportProject(AudacityProject &dest, const FilePath &fileName)
...
+         auto title = XO("Importing %s").Format(  mImportFileHandle->GetFileDescription() );
+         mProgressDialog = BasicUI::MakeProgress(title, Verbatim(ff.GetFullName()), BasicUI::ProgressShowCancel);
+      }
+      auto result = mProgressDialog->Poll(progress * ProgressSteps, ProgressSteps);
```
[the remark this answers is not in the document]
vsverchinsky: Probably, but similarly I don't want to do it in advance

src/export/ExportFileDialog.cpp
vsverchinsky: This should not change the behavior, considering that currently `ExportFileDialog::Show` is always called with project name without extension, but in context of `ExportFileDialog` I think it make more sense to allow user to specify full file name.

src/menus/FileMenus.cpp
vsverchinsky: Yes, thanks!

src/TimerRecordDialog.cpp
vsverchinsky: No, but if I understand correctly this really shouldn't happen.

src/menus/FileMenus.cpp
vsverchinsky: The behavior is restored in one of the latest commits

src/export/Export.cpp
```diff
@@ -300,91 +289,6 @@ void Exporter::FixFilename()
...
-      if(prompt)
-      {
-         ExportMixerDialog md(exportTracks, mMixerSpec.get(),
-                              NULL,
-                              1,
-                              XO("Advanced Mixing Options"));
-         if(md.ShowModal() != wxID_OK)
-            return false;
-      }
-      mChannels = mMixerSpec->GetNumChannels();
```
Paul-Licameli, line 888: At the commit "Move `ExportMixerDialog` invocation and mix-related warnings", this line is deleted, and I don't see where it moves.

src/export/Export.cpp
```diff
@@ -485,6 +377,52 @@ bool Exporter::SetAutoExportOptions() {
...
+                                                                   // state the number of channels - so we assume there are always at least two.
+                                                                   // The downside is that if someone is exporting to a mono device, the dialog
+                                                                   // will allow them to output to two channels. Hmm.  We may need to revisit this.
+                                                                   // STF (April 2016): AMR (narrowband) and MP3 may export 1 channel.
+                                                                   1u,
+                                                                   MaxExportChannels));
+   return mMixerSpec.get();
+}
+
+Exporter::DownMixMode Exporter::SetUseStereoOrMonoOutput()
```
Paul-Licameli, line 387: The return value of this function is not (yet?) used at the commit that introduces the function.

src/export/Export.cpp
```diff
@@ -336,21 +321,6 @@ bool Exporter::ExportTracks(
    return success;
 }
 
-bool Exporter::ProcessFromTimerRecording()
```
Paul-Licameli: I'm pleased to see this function unified with `Process`.  It was an ugly violation of open-closed to see this as part of the Exporter interface.

Fix the spelling in the commit comment: "became"

src/ProjectFileManager.cpp
```diff
@@ -1211,8 +1211,15 @@ bool ImportProject(AudacityProject &dest, const FilePath &fileName)
 class ImportProgress final
    : public ImportProgressListener
 {
+   wxWeakRef<AudacityProject> mProject;
 public:
    
+   ImportProgress(AudacityProject& project)
```
Paul-Licameli, line 1217: `explicit`

libraries/lib-import-export/PlainExportOptionsEditor.cpp
```diff
@@ -0,0 +1,106 @@
...
+      else if(auto val = std::get_if<int>(&it->second))
+         config.Read(mConfigKeys[index], val);
+      else if(auto val = std::get_if<double>(&it->second))
+         config.Read(mConfigKeys[index], val);
+      else if(auto val = std::get_if<std::string>(&it->second))
+      {
+         wxString wxstr;
+         if(config.Read(mConfigKeys[index], &wxstr))
+            *val = wxstr.ToStdString();
+      }
```
Paul-Licameli, line 81: I could use my new utility like this, avoiding repetition, and requiring no rewriting in case there is ever a need
to add more variant alternatives that the generic lambda handles:

```
Visit(Variant::MakeVariant(
   [](auto &val) { config.Read(mConfigKeys[index], &val); },
   [](wxString &val) {
      wxString wxStr;
      if(config.Read(mConfigKeys[index], &wxstr))
         *val = wxstr.ToStdString();
   }
), it->second);
```

Similarly in other places where you use `get_if`

saintmatthieu:
I just tried to open a wav file through the File>Recent menu, which yielded 
image ; Maybe I'll find out that it's intended, but I'm skeptical.

vsverchinsky:
Closed in favor of https://github.com/audacity/audacity/pull/4668

PR #4385, opened 2023-03-03
Reorganize libraries

Paul-Licameli:

Almost no code moved into libraries, but reorganizations so that the graph of the present libraries makes more sense.

- Files removed from lib-screen-geometry, which becomes a leaf like lib-utility, not even using wxBase
- New higher level lib-time-frequency-selection can be the future home of snapping logic
- Invert former dependencies of lib-project-histories on track lists and time selection
- Move MixAndRender up the graph into lib-effects, so lib-wavetrack doesn't use lib-realtime-effects
- Remove an extra arc from the diagram and make some third party library dependencies private
- The one movement of code into libraries: serializers for realtime effect stacks are now in lib-effects (and comments there about indepencence of lib-wave-track and lib-realtime-effects are correct now)

QA:
- [x] Test undo/redo of track changes (including abandonment of redo items) also with attention to behavior of time and spectral selection
- [x] Check mix-and-render
- [x] Check persistency of effect stacks
- [x] Hit-testing of clip drags bars; clip left and right edges; samples (in the pencil tool mode)

Recommended:


crsib (approved):
Please, fix the Linux build first!

dozzzzer:
Tested on Windows and macOS

PR #4393, opened 2023-03-07
Linear d b ruler

Paul-Licameli:
Resolves: #2683

Part of the GSOC 2022 project.  Refactoring of ruler formatting code to support more cases; a new vertical waveform scale showing non-uniformly spaced dB values.

QA:  If the new feature is sound but details of spacing and numbering and formatting are imperfect, then open issues but do not block merge of this.

However what should prevent merge is detection of any unintended changes of behavior of the ruler code elsewhere, such as when windows resize, or when scale types change from linear to log for rulers supporting that.  The exhaustive list of those places:

- [x] Horizontal timeline
- [x] Frequency analysis window (horizontal and vertical)
- [x] File > Print draws a timeline
- [x] Vertical and horizontal rulers of Time track
- [x] Horizontal and vertical rulers in Compressor effect dialog
- [x] Horizontal and vertical rulers of Curve EQ
- [x] Horizontal and vertical rulers of Classic Filters effect
- [x] Vertical ruler of MIDI track (just check the sizing)
- [x] Spectrogram vertical ruler, with various scale choices
- [x] Waveform vertical ruler (of course) with the old scale choices
- [x] Meters in Mixer board and meter toolbars

Recommended:


dozzzzer:
Tested

PR #4395, opened 2023-03-07
A bit of fun with templates

Paul-Licameli:

Redo the fix at 6a6b6d41961e24dd493aed8a621367088d4f0f72 with new template conveniences that simplify it, and will simplify many repetitions of the pattern to come in more refactoring.

QA: test starting and stopping of playback, recording (including punch-and-roll), monitoring (click in the output meter), play-at-speed button, scrubbing

Recommended:

### DIFF
--- a/libraries/lib-audio-io/ProjectAudioIO.cpp
+++ b/libraries/lib-audio-io/ProjectAudioIO.cpp
@@ -16,9 +16,9 @@ Paul Licameli split from AudacityProject.cpp
 #include "ProjectRate.h"
 #include "Track.h"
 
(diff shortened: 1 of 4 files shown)


Paul-Licameli:
Rebased without change onto recent master

vsverchinsky (approved):

dozzzzer:
Tested. 

- Playback works as expected
- Recording (incl. Punch and Roll rec.) works as expected too
- Monitoring (click in the output meter), play-at-speed button, and scrubbing show now issues too

PR #4399, opened 2023-03-09
More in lib track selection

Paul-Licameli:
Resolves: #5448

The logic for getting and setting the focused track is lowered to libraries.

TrackPanelAx is now used only by TrackPanel and together they inject the extra toolkit-dependent interaction with screen readers.

The other places that used to use TrackPanel.h, use only TrackFocus.h now.

QA: 
- [x] yellow focus border of tracks draws, and moves with cursor arrow keys
- [ ] screen reader navigation to tracks and speaking of track names

Recommended:


Paul-Licameli:
Really ready for review now and rebased onto latest master

Paul-Licameli:
Force-pushed only a commit comment change

crsib (approved):

chinakov:
> screen reader navigation to tracks and speaking of track names

Yeah, there seems to be a problem with this. Unlike in the release version of Audacity 3.4.2, this build does not let the narrator 'narrate' the individual tracks added to Audacity. Rather, it only narrates the track pane.

Paul-Licameli:
Rebased to recent master, resolved conflicts, inserted two small fixup commits so the screen reader works

Paul-Licameli:
Rebased to recent master

crsib (dismissed):

Paul-Licameli:
I promised you two fixup commits to review, but somehow misplaced them and didn't really include the fixes for the problems Anton identified.

Paul-Licameli:
Now they are there.  Please review again.  It should take just a minute.

crsib (approved):

PR #4403, opened 2023-03-11
Additional ruler changes

Paul-Licameli:

Extends
- [x] #4394 

Three commits from @micpap25 that this reviewer was unsure of.  They would have some visible side effects on other rulers than the beats and measures.  Are they really part of the intended scope?

- Fix an off-by-one error (was it an error?) drawing ticks at extremes
- Add "dB" after each number on the old logarithmic vertical ruler
- Don't draw negative time.  (But in #4394 already, there are no negative beats.)

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveformVRulerControls.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveformVRulerControls.cpp
@@ -353,7 +353,7 @@ void WaveformVRulerControls::DoUpdateVRuler(
       }
    }
    else {
-      vruler->SetUnits({});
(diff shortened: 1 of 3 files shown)


crsib (approved):

Paul-Licameli:
I want @LWinterberg to approve too that these minor design changes are as intended

LWinterberg (approved):
lgtm

Paul-Licameli:
> lgtm

Even the second bullet?  I thought I recalled you did not want "dB" drawn repeatedly in the vertical ruler.

dozzzzer:
Tested

Issue #4408, opened 2023-03-13
Default time format in Selection toolbar (and Generators) changed to hh:mm:ss + hundredths


petersampsonaudacity:
### Bug description
In 3.2.5 the default time format for the selection toolbar was changed to **hh:mm:ss + hundredths** from **hh:mm:ss + milliseconds**.

This also affects the default time format for Generators.

**hh:mm:ss + milliseconds** has been the default since Audacity 1.3.0 through 3.2.4 - and is what is documented throughout in the manual as the default.

### Steps to reproduce
1. clear auddacity settings folder
2. run 3.2.4 or any earlier
3. Observe time format in the selection toolbar is **hh:mm:ss + milliseconds**
4. Generate > Chirp
5. Observe defaulttime format is **hh:mm:ss + milliseconds**
6. close Audacity
7. Clear the audacity settings folder
8. run 3.2.5 (or 3.3.0 alpha)
9. Observe time format in the selection toolbar is **hh:mm:ss + hundredths**
10. Generate > Chirp
11. Observe defaukt time format is **hh:mm:ss + hundredths**

### Expected behavior
I see no reason for the default format to have chanfed - and I see no indication of this in the Release Notes

### Actual behavior
Default Time format in the Selection Toolbar and Generators has changed in 3.2.5

### Audacity Version
3.2.5 and latest 3.3.0 alpha

### Operating system
Windows, macOS

### Additional context
Initially surfaced in testing against this Forum post:

https://forum.audacityteam.org/t/audacity-3-2-5-generate-behaviour-has-changed/71034

petersampsonaudacity:
Testing on W11 with @Kirktopode 's branch build:  audacity-win-3.3.0-alpha-20230315+3deb3e7-x64

This branch build fixes the issue with the default time formats for the Selection and the Generators restored to **hh:mm:ss + milliseconds**

petersampsonaudacity:
This persists on the latest branch alpha by @crsib audacity-win-3.3.0-alpha-20230317+745c285-x64

![image](https://user-images.githubusercontent.com/50205984/226166306-df6aea46-a562-4031-8808-a4b5abb48906.png)

dozzzzer:
Tested

PR #4413, opened 2023-03-14
fix default time format and TimeAndSampleFormat index

Kirktopode:
Resolves: #4408

It appears that at some point, an additional time format was added before the default time format (hh:mm:ss - milliseconds), without updating places where the index of the hh:mm:ss - milliseconds format is used. I updated the references to this index and to the index for TimeAndSampleFormat referenced in NumericConverter.cpp.

Recommended:

### DIFF
--- a/libraries/lib-numeric-formats/NumericConverter.cpp
+++ b/libraries/lib-numeric-formats/NumericConverter.cpp
@@ -527,9 +527,9 @@ static const BuiltinFormatString BandwidthConverterFormats_[] = {
 // ----------------------------------------------------------------------------
 //
 NumericFormatSymbol NumericConverter::DefaultSelectionFormat()
-{ return TimeConverterFormats_[4].name; }
-NumericFormatSymbol NumericConverter::TimeAndSampleFormat()
 { return TimeConverterFormats_[5].name; }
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
> It appears that at some point, an additional time format was added ...

Ah that would be the new **seconds + milliseconds**  -  this was intended for 3.3.0, but somehow inadvertently slipped into 3.2.5

petersampsonaudacity:
Testing on W11 with @Kirktopode 's branch build:  audacity-win-3.3.0-alpha-20230315+3deb3e7-x64

This branch build fixes the issue with the default time formats for the Selection and the Generators restored to **hh:mm:ss + milliseconds**

LWinterberg (approved):
A PR even I understand :)

petersampsonaudacity:
> A PR even I understand :)

Indeed yes, so maybe that could be reviewer and if OK pulled into main

 **fix default time format and TimeAndSampleFormat index** #4413

petersampsonaudacity:
It is possible that this issue is related to:

 **Reset Configuration does not use the proper time format for the Time toolbar** #4426

PR #4416, opened 2023-03-15
"Delay" effect: Add better quality pitch shifting

SteveDaulton:
Resolves: Poor quality pitch shift in Delay effect.

*(Nyquist now has a phase-vocoder which can provide much better quality pitch shifting than the old STK pitch shifter. 
This commit adds a "High-quality pitch shift" option, while retaining backward compatibility with the previous version.)*

Recommended:

### DIFF
--- a/plug-ins/delay.ny
+++ b/plug-ins/delay.ny
@@ -6,11 +6,11 @@ $name (_ "Delay")
 $manpage "Delay"
 $debugbutton false
 $author (_ "Steve Daulton")
-$release 2.3.1-1
+$release 2.4.2-1
 $copyright (_ "GNU General Public License v2.0")
 
 
-;; License: GPL v2
+;; License: GPL v2 or later.
 ;; http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 ;; based on 'Delay' by David R. Sky
 ;;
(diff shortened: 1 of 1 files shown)


plug-ins/delay.ny
SteveDaulton, line 1: The "force push" was to fix a typo in a comment.

Paul-Licameli (approved):

LWinterberg:
image

Unfortunately this crashes for me here

SteveDaulton:
> Unfortunately this crashes for me here

Yes that is unfortunate.
I assume that you are testing on Windows?

Does the problem happen every time?
(I've not been able to reproduce the crash, but I might only have tested in release builds of Audacity).

As the bug is in Audacity rather than in the plugin, have you logged the bug?

Paul-Licameli:
Does the first button make a call stack trace?

petersampsonaudacity:
Testing on W10 with Steve's branch:  audacity-win-3.3.0-alpha-20230316+c72c661-x64

I can't get this to crash - not with @LWinterberg 's setings above, nor with other settings that I try.

LWinterberg:
Ah, thank you David, you saved me a bunch of time - I was about to recompile this with different things turned on and off to figure out why peter can't repro it.

LWinterberg:
image
Testing with a release build and just a short tone impulse, the new option is absolutely an improvement.

petersampsonaudacity:
Pitch/Tempo looks rubbishy compared to the other too.

@LWinterberg  I would suggest that it could benefit users to change the default setting to **High-quality Pitch Shift**

The High Quality Pitch Shift seems to operate in a rapid enough manner

SteveDaulton:
I've now tested with a debug build on Linux, and I'm still not able to reproduce the issue. Maybe it's specific to Windows?

crsib:
You need to use some kind of a sanitizer, that prevents reads for uninitialized variables. It is unsafe, because in release builds such variables do contain random content.

Issue #4381, opened 2023-03-02
Add support for ffmpeg 6


marillat:
### Your idea

Would be nice if you can add support for ffmpeg 6

### Problem to be solved

Missing support for ffmpeg 6

Ryan-Miller-Audacity:
Verified support for FFMPEG 6 on both Windows and MacOS.

SakiPapa:
Audacity for macOS does not recognize ffmpeg 6.0 libraries

I am using Audacity 3.2.5 on macOS. (M1 PowerBook 13 Pro, Monterey 12.6.5, 16 GB memory)

My ffmpeg library (Arm version)  was installed by Homebrew.

Yesterday after I updated ffmpeg to version 5.1.2 version 6.0,  I got a warning that the library could not be found.

I confirmed that the ffmpeg libraries are installed in 

/opt/homebrew/Cellar/ffmpeg/6.0/.

I restored the old ffmpeg (5.1.2_6) from the TimeMachine backup back to 

/opt/homebrew/Cellar/ffmpeg/5.1.2_6 

and Audacity recognized the 5.1.2 library.

I think the macOS version of Audacity 3.2.5 (Apple Silicon) is not recognize ffmpeg 6.0.

The same result was occured with Audacity 3.3.0-beta1.

PR #4419, opened 2023-03-16
FFmpeg 6 support

crsib:
Resolves: #4381 

Recommended:


libraries/lib-ffmpeg-support/CMakeLists.txt
```diff
@@ -70,6 +70,13 @@ if (${_OPT}use_ffmpeg)
       impl/avformat/59/AVFormatImpl.cpp
       impl/avutil/57/AVUtilImpl.cpp
 
+      impl/ffmpeg-6.0.0-single-header.h
+      impl/avutil/58/avconfig.h
+      impl/avcodec/60/AVCodecIDLookup.cpp
+      impl/avcodec/60/AVCodecImpl.cpp
+      impl/avformat/60/AVFormatImpl.cpp
+      impl/avutil/58/AVUtilImpl.cpp
```
Paul-Licameli, line 78: I see a mix of "58" and "60" here which I assume is intended

libraries/lib-ffmpeg-support/impl/avcodec/60/AVCodecIDLookup.cpp
```diff
@@ -0,0 +1,37 @@
+extern "C"
+{
+#include "../../avutil/58/avconfig.h"
```
Paul-Licameli, line 3: 60 using 58 again, comment ?

libraries/lib-ffmpeg-support/impl/avformat/60/AVFormatImpl.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+  AVFormatImpl.cpp
+
+  Dmitry Vedenko
+
+**********************************************************************/
+
+extern "C"
+{
+#include "../../avutil/58/avconfig.h"
```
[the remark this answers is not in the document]
Paul-Licameli: Or maybe not? ... maybe this difference of number is just frequent in all the other versions too.  I'm comparing with 55 for instance.

libraries/lib-ffmpeg-support/impl/avcodec/60/AVCodecIDLookup.cpp
crsib: This files are literally copied, none of the original files have a comment about different versions between avutil and the rest of the libraries. 

The difference between versions is due to the difference inside the FFmpeg itself.

Paul-Licameli (approved):

PR #4432, opened 2023-03-20
Qt volume control

ksoze95:
Resolves: #4345

Recommended:


src/qt/uicomponents/MeterPanelHandler.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+qreal MeterPanelHandler::Value() const
+{
+   return m_value;
+}
+
+void MeterPanelHandler::SetValue(qreal value)
+{
+   if (value < 0.0) {
+      qDebug() << "MeterPanelHandler value cannot be less than 0";
```
crsib, line 17: Probably lets `assert` as well, so it fails clearly in Debug?

src/qt/uicomponents/VolumeControlHandler.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+qreal VolumeControlHandler::Volume() const
+{
+   return m_volume;
+}
+
+void VolumeControlHandler::SetVolume(qreal volume)
+{
+   if (volume < 0.0) {
+      qDebug() << "volume cannot be less than 0";
```
crsib, line 17: ditto

crsib (approved):

PR #4441, opened 2023-03-23
Fix handle leak on Linux and macOS

Quipyowert2:
If opening toFifo succeeds but fromFifo fails, the original code would leak the toFifo filehandle.

I found, via Cppcheck, that there are a few file handle leaks in Audacity. This PR fixes those leaks. In setup_preferences on macOS, if opening the file whose name is specified in `prefname` succeeds but a colon was not found in the filename, the file handle `pf` leaks when the function returns without closing it.

Cppcheck errors:
```
[D:/Linux_home/nathan/src/audacity/lib-src/libnyquist/nyquist/sys/mac/macstuff.c:213] (error) Resource leak: pf [resourceLeak]
[D:/Linux_home/nathan/src/audacity/modules/mod-script-pipe/PipeServer.cpp:164] (error) Resource leak: toFifo [resourceLeak]
```

Cppcheck version 2.10
Audacity version: 3.2.5-6ec6548af "A bit of fun with templates"

Recommended:

### DIFF
--- a/modules/scripting/mod-script-pipe/PipeServer.cpp
+++ b/modules/scripting/mod-script-pipe/PipeServer.cpp
@@ -161,6 +161,7 @@ void PipeServer()
    if (fromFifo == NULL)
    {
       perror("Unable to open fifo from server to script");
+      fclose(toFifo);
       return;
    }
 
--- a/lib-src/libnyquist/nyquist/sys/mac/macstuff.c
+++ b/lib-src/libnyquist/nyquist/sys/mac/macstuff.c
@@ -210,7 +210,10 @@ void setup_preferences(char *filename)
         pf = fopen((char *) prefname, "w");
         if (pf == NULL) return;
         cp = strrchr((char *) filename, ':');
-        if (cp == NULL) return;
+        if (cp == NULL) {
+            fclose(pf);
+            return;
+        }
         cp[1] = 0;
         /* now, filename is the path. If filename ends in runtime, this
          * is probably the standard nyquist runtime folder. We should put


Quipyowert2:
Oops, accidentally committed a mode change to this pull request. I've removed the mode change. Sorry.

Quipyowert2:
Added a short description for the handle bug on macOS.

Quipyowert2:
Fixed merge conflict due to moving modules/mod-script-pipe directory to modules/scripting. Rebased onto master and pushed.

modules/scripting/mod-script-pipe/PipeServer.cpp
```diff
@@ -161,6 +161,7 @@ void PipeServer()
    if (fromFifo == NULL)
```
GrigorasGabriel36, line 161: Just a quick suggestion: I noticed you're using NULL in the code. Since we're working with C++11 and beyond, it might be a good idea to switch to nullptr. It's more type-safe and helps avoid some subtle bugs that can come from using NULL, especially in template or overloaded function contexts. Plus, nullptr has its own type (std::nullptr_t), which makes it more explicit and clear. Just a small modern C++ improvement!
 There is allot of technical literature on this topic.
 
 1. **Effective Modern C++** by Scott Meyers  
2. **The C++ Programming Language (4th Edition)** by Bjarne Stroustrup  
3. **C++ Primer (5th Edition)** by Stanley B. Lippman, Josée Lajoie, and Barbara E. Moo  
4. **More Effective C++** by Scott Meyers

Quipyowert2: I'm at work right now but I can change it to nullptr when I get home.

Issue #4285, opened 2023-02-06
Beats & Measures: new dragging behaviour when snapping is turned on


Tantacrul:
### Your idea

When the user opts to turn on snapping, we require an additional behaviour: clips that are moved should _also_ snap to the selected unit of measurement. This is essential for Beats & Measures to function effectively. See the full [**Figma design here**](https://www.figma.com/file/SKDJypvy37EAauTPgwo2tL/07_WX---Beats-%26-measures?node-id=568%3A17189&t=UKmXqBrifEeSaFAe-1)

![image](https://user-images.githubusercontent.com/21022311/217015608-020cbb37-c83a-48db-9d74-60feb60c8772.png)

### Problem to be solved

This change will allow musical creation work to be much easier and predictable. It is an essential component of adding Beats and Measures functionality.

petersampsonaudacity:
Testing this on W10 with @crsib 's branch build for this:  https://github.com/audacity/audacity/actions/runs/4574323691

This works as @Tantacrul  intends for dragging clips when snapping is "on"

However it does _**NOT**_ affect the behaviour for**Tracks > Align Tracks** - is this intended desired behaviour?

Note that this was always the case for **Align Tracks** - see this clip from the 3.2.5 Manual:
![image](https://user-images.githubusercontent.com/50205984/229117358-23e6883b-34b3-492c-99cd-6ff39082f95f.png)

On this page:
https://manual.audacityteam.org/man/selection_toolbar.html

petersampsonaudacity:
I ask this mainly in the interests of getting the manual's documentation correct.

Tantacrul:
Musically speaking, it's not massively useful for align tracks to ignore snapping. I just tried out a few scenarios and found this:

- With snapping to a beat (or time value) on, snapping behaved as expected
- If I wanted to snap to a clip on another track, I was able to by switching the 'snap' checkbox off

It's hard to make a very confident judgement here, since it's really about user expectation. However, the way it is working in the build you referenced makes sense to me. It reduces the possibility of accidentally misaligning things.

I'd suggest we see how this performs and potentially create an option for the user to control snapping more finely in preferences.

dozzzzer:
Tested against #4453

PR #4453, opened 2023-03-27
Clip snap to time

vsverchinsky:
Resolves: #4285 

Also fixes wrong guidline positions when clips are moved between tracks

Currently adds to new commits on top of #4414

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -2292,6 +2292,7 @@ bool WaveTrack::CanInsertClip(
 {
    for (const auto &c : mClips)
    {
+      //VS: Do we need to take into account sample rate difference?
       double d1 = c->GetPlayStartTime() - (clip->GetPlayEndTime()+slideBy);
       double d2 = (clip->GetPlayStartTime()+slideBy) - c->GetPlayEndTime();
       if ( (d1<0) &&  (d2<0) )
@@ -2302,6 +2303,9 @@ bool WaveTrack::CanInsertClip(
          // move the clip at most once.  
          // We divide by 1000 rather than set to 0, to allow for 
          // a second 'micro move' that is really about rounding error.
+
+         // VS: clip could be moved more than once, moving it in opposite
+         // direction may reintroduce overlapping condition
          if( -d1 < tolerance ){
             // right edge of clip overlaps slightly.
             // slide clip left a small amount.
--- a/src/tracks/labeltrack/ui/LabelTrackShifter.cpp
+++ b/src/tracks/labeltrack/ui/LabelTrackShifter.cpp
@@ -134,12 +134,14 @@ public:
       return true;
    }
 
-   bool Attach( Intervals intervals ) override
+   bool Attach( Intervals intervals, double offset ) override
    {
       auto pTrack = mpTrack.get();
       std::for_each( intervals.rbegin(), intervals.rend(),
-         [this, pTrack](auto &interval){
+         [this, pTrack, offset](auto &interval){
             auto pData = static_cast<IntervalData*>( interval.Extra() );
+            if(offset != .0)
+               pData->region.move(offset);
             auto index = pTrack->AddLabel(pData->region, pData->title);
             // Recreate the simpler TrackInterval as would be reported by LabelTrack
             mMoving.emplace_back( pTrack->MakeInterval(index) );
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
@@ -153,7 +153,7 @@ public:
       return ok;
    }
 
-   bool Attach( Intervals intervals ) override
+   bool Attach( Intervals intervals, double offset ) override
    {
       for (auto &interval : intervals) {
          auto pData = static_cast<WaveTrack::IntervalData*>( interval.Extra() );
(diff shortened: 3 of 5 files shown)


src/tracks/labeltrack/ui/LabelTrackShifter.cpp
```diff
@@ -134,12 +134,14 @@ class LabelTrackShifter final : public TrackShifter {
...
 
-   bool Attach( Intervals intervals ) override
+   bool Attach( Intervals intervals, double offset ) override
    {
       auto pTrack = mpTrack.get();
       std::for_each( intervals.rbegin(), intervals.rend(),
-         [this, pTrack](auto &interval){
+         [this, pTrack, offset](auto &interval){
             auto pData = static_cast<IntervalData*>( interval.Extra() );
+            if(offset == .0)
```
Paul-Licameli, line 143: Do you mean `!-`?

src/tracks/ui/TimeShiftHandle.cpp
```diff
@@ -227,7 +227,7 @@ bool TrackShifter::AdjustFit(
    return false;
 }
 
-bool TrackShifter::Attach( Intervals )
+bool TrackShifter::Attach( Intervals, double offset )
```
Paul-Licameli, line 230: Might omit `offset`

src/tracks/labeltrack/ui/LabelTrackShifter.cpp
vsverchinsky: Yes, thanks!

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli, line 1: In commit comment “guidelines”

src/tracks/ui/TimeShiftHandle.cpp
```diff
@@ -565,71 +565,59 @@ UIHandle::Result TimeShiftHandle::Click
...
-            if (pInterval) {
-               clipLeft = pInterval->Start() + desiredSlideAmount;
-               clipRight = pInterval->End() + desiredSlideAmount;
-            }
-            else {
-               clipLeft = capturedTrack.GetStartTime() + desiredSlideAmount;
-               clipRight = capturedTrack.GetEndTime() + desiredSlideAmount;
-            }
+      auto track = state.mCapturedTrack.get();
+      double clipLeft, clipRight;
```
Paul-Licameli, line 576: Initializers please

src/tracks/playabletrack/wavetrack/ui/WaveTrackShifter.cpp
```diff
@@ -153,15 +153,24 @@ class WaveTrackShifter final : public TrackShifter {
...
          auto pData = static_cast<WaveTrack::IntervalData*>( interval.Extra() );
          auto pClip = pData->GetClip();
          if ( !mpTrack->AddClip( pClip ) )
             return false;
          mMigrated.insert( pClip.get() );
-         mMoving.emplace_back( std::move( interval ) );
+         if(offset == .0)
+            mMoving.emplace_back( std::move( interval ) );
+         else
+         {
```
Paul-Licameli, line 167: Were the side effects on interval caused by move in the first branch important?  This branch omits them

Paul-Licameli: I suppose not, now I see they belong to expiring `remover`

Paul-Licameli:
The older version adjusts `desiredSlideAmount` first to quantize, then for snapping, then for AdjustFit (small tolerances allowed by WaveClip, but depending also on zoom level -- time per pixel column), then for `AdjustOffsetSmaller`.

The new version adjusts `desiredSlideAmount` for quantization, then for AdjustFit, then for snapping, then for `AdjustOffsetSmaller`.

That was difficult to figure out, in all this logic I had forgotten about for long.

Maybe this makes a subtle change of behavior of clip moving in some cases?  But maybe this does not matter.

Paul-Licameli:
> The older version adjusts `desiredSlideAmount` first to quantize, then for snapping, then for AdjustFit (small tolerances allowed by WaveClip, but depending also on zoom level -- time per pixel column), then for `AdjustOffsetSmaller`.
> 
> The new version adjusts `desiredSlideAmount` for quantization, then for AdjustFit, then for snapping, then for `AdjustOffsetSmaller`.
> 
> That was difficult to figure out, in all this logic I had forgotten about for long.
> 
> Maybe this makes a subtle change of behavior of clip moving in some cases? But maybe this does not matter.

I'm thinking it through more carefully.

In the old version, I think`AdjustOffsetSmaller` does not matter in case of moving between tracks, because all the other checks for fit already adjust the offset, so it will do nothing.  Only in case of sliding horizonally within track, does this matter.  `AdjustFit` does all the protection against overlapping clips.

But with this change, snapping may make another change of horizontal position of clips after those checks!

But then, `AdjustOffsetSmaller` really does work to prevent overlap.

So really it's just the composition of a correct vertical move and a correct horizontal move.

We won't get overlapping clips.  We might devise some test case where a drag between clips is permitted in one version but not the other.  Because the check for fit might give different answers at the snapped and unsnapped x coordinates.

Paul-Licameli:
I tried this case:  make three clips in sequence that touch exactly.  Drag the middle one to another empty track below.  Then try to drag it back again into the gap that fits it exactly.

Actually it's hard to do in both versions, in neither one, just working simply.  It seems in the old version I must drag horizontally until snapping, then move vertically without any more lateral movement.

I'm not sure, maybe the new makes it slightly easier but still it doesn't always just work.

Maybe you can figure out an extra fix for this case.

But, as for exploring consequences, to find any reason to dealy this PR -- no, I haven't found one.

Paul-Licameli (approved):

Paul-Licameli:
I would just ask for some more comments in `Drag` making explicit what I said about the correctness of clip fit in all case.

Issue #4438, opened 2023-03-22
Enhancement to Nyquist stye guide


SteveDaulton:
### Your idea

It is generally considered good practice to avoid using global variables as far as possible, but when writing Nyquist plugins that have a GUI, global variables are unavoidable because widget variables have global scope.

This proposal is to introduce a convention that allows widget variables to be easily identified visually, and to treat widget variables as constants by typing widget variables in uppercase.

Old style:
`;control tempo "Enter the tempo" float "bpm" 120 30 300`

New style:
`;control TEMPO "Enter the tempo" float "bpm" 120 30 300`

### Problem to be solved

Nyquist plugin widget variables have global scope, but lack visual clues to indicate that this is the case. Typing widget variables in uppercase would improve readability, and along with a convention to treat widget variables as constants would help to avoid bugs.

### Prior art

Global constants in Python are written in uppercase by convention.

### Additional context

Nyquist variables are case insensitive, so changing widget variables to upper case, (rather than using \*earmuffs\*) avoids breaking backward compatibility. This is particularly relevant for users that may have macros that use Nyquist effects.

PR #4494, opened 2023-04-05
Issue #4438 Nyquist code style: Update plugins

SteveDaulton:
Resolves: #4438

*(Upper case widget "constants")*

Recommended:

What should be tested:

- [x] ShelfFilter
- [x] SpectralEditParametricEQ
- [x] SpectralEditShelves
- [x] Adjustable-fade
- [x] Beat
- [x] Clipfix
- [x] Crossfadetracks
- [x] Delay
- [x] Eq-xml-to-txt-converter
- [x] Equalabel
- [x] Highpass
- [x] label-sounds
- [x] Limiter
- [x] Lowpass
- [x] Noisegate
- [x] Notch
- [x] Niquist-plug-in-installer
- [x] Pluck
- [x] Rhythmtrack
- [x] RIssetdrum
- [x] Sample-data-export
- [x] Sample-data-import
- [x] Tremolo
- [x] Vocalrediso
- [x] Vocoder

### DIFF
--- a/plug-ins/crossfadetracks.ny
+++ b/plug-ins/crossfadetracks.ny
@@ -5,7 +5,7 @@ $name (_ "Crossfade Tracks")
 $debugbutton disabled
 $preview selection
 $author (_ "Steve Daulton")
-$release 3.0.4-1
+$release 3.0.4-2
 $copyright (_ "GNU General Public License v2.0 or later")
 
 ;; License: GPL v2+
@@ -14,23 +14,21 @@ $copyright (_ "GNU General Public License v2.0 or later")
 ;; For information about writing and modifying Nyquist plug-ins:
 ;; https://wiki.audacityteam.org/wiki/Nyquist_Plug-ins_Reference
 
-$control type (_ "Fade type") choice (
-   ("ConstantGain" (_ "Constant Gain"))
-   ("ConstantPower1" (_ "Constant Power 1"))
-   ("ConstantPower2" (_ "Constant Power 2"))
-   ("CustomCurve" (_ "Custom Curve"))
-) 0
-$control curve (_ "Custom curve") real "" 0 0 1
-$control direction (_ "Fade direction") choice (
-   (_ "Automatic")
-   ("OutIn" (_ "Alternating Out / In"))
-   ("InOut" (_ "Alternating In / Out"))
-) 0
+$control TYPE (_ "Fade TYPE") choice (
+    ("ConstantGain" (_ "Constant Gain"))
+    ("ConstantPower1" (_ "Constant Power 1"))
+    ("ConstantPower2" (_ "Constant Power 2"))
+    ("CustomCurve" (_ "Custom Curve"))) 0
+$control CURVE (_ "Custom curve") real "" 0 0 1
+$control DIRECTION (_ "Fade direction") choice (
+    (_ "Automatic")
+    ("OutIn" (_ "Alternating Out / In"))
+    ("InOut" (_ "Alternating In / Out"))) 0
 
 
-(defun crossfade (type dir curve)
+(defun crossfade ()
   (setf fade-out
-    (case dir
+    (case DIRECTION
       (0 (equal (guessdirection) 'OUT))   ; auto
       (1 (oddp (get '*track* 'index)))    ; fade out odd
       (T (evenp (get '*track* 'index))))) ; fade out even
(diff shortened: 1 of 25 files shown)


plug-ins/sample-data-import.ny
```diff
@@ -104,4 +104,4 @@ $control bad-data (_ "Invalid data handling") choice (("ThrowError" (_ "Throw Er
     (close stream)
     audio-out))
 
-(catch 'err (sound-from-file filename))
+(catch 'err (sound-from-file))
```
LWinterberg, line 107: why can `filename` be omitted here?

plug-ins/crossfadetracks.ny
```diff
@@ -80,6 +78,7 @@ audio clip, fade in, otherwise fade out."
       (setf out-dist (min out-dist (abs (- end (second (nth i clips)))))))
     (if (< in-dist out-dist) 'in 'out)))
 
+
 (if (< (length (get '*selection* 'tracks)) 2)
     (format nil (_ "Error.~%Select 2 (or more) tracks to crossfade."))
-    (crossfade type direction curve))
+    (crossfade))
```
LWinterberg, line 84: Why can these arguments now be omitted?

plug-ins/equalabel.ny
```diff
@@ -23,134 +23,145 @@ $copyright (_ "GNU General Public License v2.0 or later")
...
+
+(setf num-before-text (<= ZEROS 3))
+(setf zeropad (1+ (rem (1- ZEROS) 3)))
 
 ;; Analyze plug-ins may return text message per track but
 ;; we only want error messages once, and only one set of labels.
 (if (lasttrackp)
-    (make-labels)
-    "") ;No-op
+    (make-labels num-before-text zeropad)
```
LWinterberg, line 166: Why does this one gain arguments?

LWinterberg (approved):
I went through this and saw the promised style changes. I also saw some workarounds removed only relevant for old versions and some reshuffling of where things are declared.

What I don't understand is why some functions gain or lose arguments, seemingly out of the blue. I've marked some instances of that in the comments. I'm obviously no lisp developer, so I'm curious why that is.

plug-ins/sample-data-import.ny
SteveDaulton: (Nyquist is case insensitive)

> why can `filename` be omitted here?

Because `FILENAME` (previously `filename`) is a global variable, so can be read directly from the global scope. It is not being modified anywhere, so we are correctly treating it as a global constant, as indicated by the capitalisation of the name.

Previously, `filename` within the function `sound-from-file` redefined the variable name from the outer scope as a variable with local scope. This is generally considered bad practice as it can be unclear that `filename` within the  function is a different variable than `filename` from the outer scope. In the new code, it is explicit that we are using  `FILENAME` as a global constant.

plug-ins/equalabel.ny
SteveDaulton: > Why does this one gain arguments?

The function `make-labels` was previously defined as: `(defun make-labels (&aux labels)`
The `&aux labels`  parameter defines an auxiliary variable that is local within the function.

`num-before-text` was previously defined with global scope in the line:
```
(setf num-before-text (<= zeros 3))
```
It is generally better to avoid global variables unless absolutely necessary (as is the case with `;control` lines).
By passing `num-before-text` as an argument, we maintain encapsulation rather than littering the code unnecessarily with globals.

In the case of `zeropad`, it was even worse as not only were we previously using `zeros` with global scope, but we were also modifying it within the function `make-one-label`. In the new version, we explicitly pass the value of `zeropad` and maintain encapsulation, which not only improves readability, but is also a lot safer.

plug-ins/crossfadetracks.ny
SteveDaulton: > Why can these arguments now be omitted?

Because the function `crossfade` does not need them.
`TYPE`, `DIRECTION` and `CURVE` are all global variables. They have to be global because they are defined by the controls:
```
$control TYPE (_ "Fade TYPE") choice (
    ("ConstantGain" (_ "Constant Gain"))
    ("ConstantPower1" (_ "Constant Power 1"))
    ("ConstantPower2" (_ "Constant Power 2"))
    ("CustomCurve" (_ "Custom Curve"))) 0
$control CURVE (_ "Custom curve") real "" 0 0 1
$control DIRECTION (_ "Fade direction") choice (
    (_ "Automatic")
    ("OutIn" (_ "Alternating Out / In"))
    ("InOut" (_ "Alternating In / Out"))) 0
```

The point of this pull request is that since we **must** use global variables in Nyquist plug-ins that have controls, then it is safer and more readable to implement a convention where we treat them as global constants, and mark them as such by using upper case.

plug-ins/equalabel.ny
LWinterberg: thanks for the explanations!

dozzzzer:
@SteveDaulton please rebase this branch onto `master`

petersampsonaudacity:
@dozzzzer  

@SteveDaulton Steve is offline for a couple of weeks with no access to GitHub - he says he'll deal with it when he gets back home

chinakov:
I have recieved a list of all the .ny plugins that are affected by changes within this task. 
I will go through each one of them separately and verify that nothing got broken. 

image

chinakov:
Added checklist.

SteveDaulton:
> @SteveDaulton please rebase this branch onto `master`

Why does it require rebasing? GitHub says that there's no conflicts, so can't you just squash and merge?

PR #4511, opened 2023-04-07
v3.3.0: Update Italian Translations

mlocati:
Here's the 100% translated strings for Italian.

Is there a standard formatting approach for .po files?

I mean, I translated those strings with Transifex, and I downloaded them.
But I don'l like the formatting they use.

For example, here's a section of the file generated by Transifex:

```po
msgid "Cannot check mutual sample rates without both devices.\n"
msgstr ""
"Impossibile verificare le frequenze di campionamento reciproche senza "
"entrambi i dispositivi.\n"
```

- Why the source string is right after `msgid` and the translation is on its own line?
- I don't want to break lines at random points: I'd like to break lines when the string contains `\n`.

So I reformatted the .po file this way:

```sh
cat downloaded-from-transifex.po \
    | sed -z 's/"\n"//g'  \
    | sed -z 's/\\n/\\n"\n"/g'  \
    | sed '/^""$/d' \
    > fixed.po
```

Where:
- the first `sed` make it so that we have just one line for every `msgid`/`msgstr`
- the second `sed` splits the strings when they contain the `\n` string
- the third `sed` removes lines containing only `""`

For example, let's consider this string (as downloaded from Transifex):

```po
msgid ""
"Running two copies of Audacity simultaneously may cause\n"
"data loss or cause your system to crash.\n"
"\n"
msgstr ""
"L'esecuzione contemporanea di due copie di Audacity può causare\n"
"perdita di dati o un blocco del sistema.\n"
"\n"
```

With the first `sed` we have:

```po
msgid "Running two copies of Audacity simultaneously may cause\ndata loss or cause your system to crash.\n\n"
msgstr "L'esecuzione contemporanea di due copie di Audacity può causare\nperdita di dati o un blocco del sistema.\n\n"
```

With the second `sed` we have:

```po
msgid "Running two copies of Audacity simultaneously may cause\n"
"data loss or cause your system to crash.\n"
"\n"
""
msgstr "L'esecuzione contemporanea di due copie di Audacity può causare\n"
"perdita di dati o un blocco del sistema.\n"
"\n"
""
```

With the final `sed` we have:

```po
msgid "Running two copies of Audacity simultaneously may cause\n"
"data loss or cause your system to crash.\n"
"\n"
msgstr "L'esecuzione contemporanea di due copie di Audacity può causare\n"
"perdita di dati o un blocco del sistema.\n"
"\n"
```

But everything seems rather arbitrary... I mean, it's ok for me, but others may have a different preference about formatting.

What about adding some specifics of the format of the .po files (or, even better, a script)?

### DIFF
--- a/locale/it.po
+++ b/locale/it.po
@@ -2,25 +2,24 @@
 # Copyright (C) YEAR Audacity Team
 # This file is distributed under the same license as the audacity package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
-#
+# 
 # Translators:
-# Matteo O., 2022
-# Paval Shalamitski <yehekim@gmail.com>, 2022
-# Michele Locati <michele@locati.it>, 2022
-#
+# Matteo O., 2023
+# Paval Shalamitski <yehekim@gmail.com>, 2023
+# Michele Locati <michele@locati.it>, 2023
+# 
 #, fuzzy
 msgid ""
-msgstr ""
-"Project-Id-Version: audacity 3.0.3\n"
+msgstr "Project-Id-Version: audacity 3.0.3\n"
 "Report-Msgid-Bugs-To: audacity-translation@lists.sourceforge.net\n"
 "POT-Creation-Date: 2023-04-05 16:57+0300\n"
-"PO-Revision-Date: 2022-12-22 07:39+0000\n"
-"Last-Translator: Michele Locati <michele@locati.it>, 2022\n"
-"Language-Team: Italian (https://www.transifex.com/klyok/teams/690/it/)\n"
-"Language: it\n"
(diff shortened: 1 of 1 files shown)


crsib:
Hi, Michele! Thank you a lot for the translation!

We have no formatting preferences for PO files format. I think that they are rarely edited manually, so it doesn't feel that having formatting preferences is needed much.

mlocati:
@crsib do you mind if I submit another pull request that adds a script aimed to "normalize" the .po files?

I'd also make it so it's called by the `locale/update_po_files.sh` script (that way, I won't have to remember those `sed` commands, and any subsequent commit will contain only *actual* changes).

crsib:
Please do!

PR #4512, opened 2023-04-07
Fix registry usage

Paul-Licameli:
Resolves: Two questionable but minor problems found post-review

Recommended:

### DIFF
--- a/libraries/lib-numeric-formats/formatters/ParsedNumericConverterFormatter.cpp
+++ b/libraries/lib-numeric-formats/formatters/ParsedNumericConverterFormatter.cpp
@@ -1048,10 +1048,9 @@ Registry::BaseItemPtr MakeGroup (
 
    for (size_t index = 0; index < count; ++index)
    {
-      const auto functionIdentifier =
-         wxString::Format(L"%s_%zu", identifier.GET(), index);
-
       auto& formatString = formatStrings[index];
+
+      const auto functionIdentifier = formatString.name.Internal();
 
       group->items.push_back(std::make_unique<NumericConverterRegistryItem>(
          functionIdentifier, formatString.name,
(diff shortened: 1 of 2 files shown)


crsib:
I don't understand what this PR addresses.

Paul-Licameli:
> I don't understand what this PR addresses.

Are the commit comments enough explanation?

crsib:
The first one performs a copy of the registry for no reason, because we don't care about the ordering. 

The second one is very minor, but I don't see how it changes anything really. I think initially I haven't used the format string because I don't understand how registry sorting works. I don't really understand it even now.

Implementation worked for me, so I kept it. If we decide to allow user to sort the list (I don't see it coming at all) - it doesn't matter what is written to the config - format string, some ordered identifier or even GUID. 

So...
> Are the commit comments enough explanation?
No, not really.

crsib:
Oh, and yeah - the first commit crashes tests, so it definitely should not be merged.

Paul-Licameli:
> Oh, and yeah - the first commit crashes tests, so it definitely should not be merged.

I should look closer at it then.  Maybe fix the inconsistency the opposite way.  There is another call to Visit() in the file that could instead be the changed one.

crsib:
They are inconsistent because they are used to achieve different goals. Why this inconsistency is a problem? Rewriting `void SnapFunctionsRegistry::Visit(SnapRegistryVisitor& visitor)` will break UI, I won't approve it either.

Paul-Licameli:
> They are inconsistent because they are used to achieve different goals. Why this inconsistency is a problem? Rewriting `void SnapFunctionsRegistry::Visit(SnapRegistryVisitor& visitor)` will break UI, I won't approve it either.

Then wait to see the rest of a bigger PR that cleans up the Registry system to simplify usage.

Paul-Licameli:
Registry should have let you express your intent better and it should have been easier to figure it out.
- Type safety (no usage of BaseItemPtr)
- Less boilerplate code for `SnapRegistryItemRegistrator`, `SnapFunction`, `SnapFunctionGroup`, etc. -- easy generation from templates instead
- Easier definition of analogues of `Command`, `Menu`, etc. so you could express the un-scattered, ORDERED tree of built-in snap functions directly
- Therefore avoiding use of ordering hints except where really needed for scattered items

Issue #4454, opened 2023-03-28
Implement QML side panel


ksoze95:
Sub-tasks:

- [x] Modify meter panel to allow for vertical orientation and resizing
- [x] Implement custom dial
- [x] Implement custom slider
- [x] Implement sidebar
- [ ] Implement a tester component
- [x] Make a model item list view delegate
- [x] Add a way to dynamically add new track control to the sidebar at runtime
- [x] Implement custom option button which just deletes the current track control panel
- [ ] Add scrollview with custom themed scrollbar
- [x] Add '+ Add new track' button to the sidebar
- [x] Make the solo and mute buttons toggleable
- [ ] Add hover and active states to the track control panel
- [ ] Implement the correct behaviour for the solo and mute buttons, possibly follows MuseScore mixer behaviour

PR #4516, opened 2023-04-07
Qt sidebar

ksoze95:
Resolves: #4454

Recommended:


src/qt/EditToolbarHandler.h
```diff
@@ -1,5 +1,5 @@
 #ifndef EDIT_TOOLBAR_HANDLER_H
-#define EDIT_TOOLBAR_HANDELR_H
+#define EDIT_TOOLBAR_HANDLER_H
```
crsib, line 2: TBH I prefer `#pragma once` exactly for this reason

src/qt/uicomponents/MeterPanelHandler.cpp
```diff
@@ -13,15 +13,7 @@ qreal MeterPanelHandler::Value() const
...
-   if (value < 0.0) {
-      qDebug() << "MeterPanelHandler value cannot be less than 0";
-      return;
-   }
-
-   if (value > 1.0) {
-      qDebug() << "MeterPanelHandler value cannot be greater than 1";
-      return;
-   }
+   Q_ASSERT(value >= 0.0 && value <= 1.0);
```
crsib, line 16: 1. I would prefer the standard `assert`
2. Asserting should not replace the checks

src/qt/qml/main.qml
```diff
@@ -93,9 +93,17 @@ ApplicationWindow {
       id: toolsToolbar
```
crsib, line 93: I dunno why we need this, but no objection for now

src/qt/EditToolbarHandler.h
ksoze95: Replaced with pragma.

src/qt/uicomponents/MeterPanelHandler.cpp
ksoze95: Done.

src/qt/uicomponents/TrackType.h
```diff
@@ -0,0 +1,24 @@
+#ifndef TRACK_TYPE_H
+#define TRACK_TYPE_H
+
+#include <QtCore/QtCore>
+#include <QtQml/qqmlregistration.h>
+
+class TrackType
```
crsib, line 7: We will need to figure out how to make it open ended. No assumption should be made about the available track types...

crsib (approved):
I wonder if this should be squash merged

PR #4518, opened 2023-04-07
Update Corsican locale co.po for Audacity 3.3.0

Patriccollu:
Hello @crsib 

This is an update of **Corsican** localization for Audacity 3.3.0.

The three files on [Transifex](https://app.transifex.com/klyok/audacity/language/co/) are also 100% translated.

Best regards,
Patriccollu.


crsib:
Thank you a lot!

Issue #4427, opened 2023-03-19
Display oddity with Snapping toolbar when turning snapping off


petersampsonaudacity:
### Bug description

In the new **Snapping** toolbar, if you turn Snap on with a chosen snap format and then turn it if you are left with a residual blue background the size of the text of the now defunct Snap format.

### Steps to reproduce

1. Launch @crsib latest branch build for the updated lower tooldock
2. Observe: snapping is off - and the snap field has a grayed-out (presumably) **secconds**#
3. turn snapping on
4. Observe: seconds is no longer grayed out - and is active
5. Change snapping to anything else (I used samples)
6. turn snapping off
7. Observe, instead of seeing an expected grayed-out format as selected in step 5 - instead you get a blue block the size of the format selected in step 5

At step 2
![image](https://user-images.githubusercontent.com/50205984/226168819-8a09dc19-faf8-4cfe-8223-d07800837046.png)

At Step 4
![image](https://user-images.githubusercontent.com/50205984/226168973-41478f59-6360-4035-b121-670271fcf785.png)

At Step 7
![image](https://user-images.githubusercontent.com/50205984/226169120-5aac53e7-8a16-413b-a001-cc7375d56a96.png)

### Expected behavior
The Snapping field, when turned off, should show a grayed out default setting or tha last-used setting

### Actual behavior
See above in STRs

### Audacity Version
@crsib latest branch for updated lower tooldock:  audacity-win-3.3.0-alpha-20230317+745c285-x64

### Operating system
Windows 10 - but assume all OS

### Additional context
None

Ryan-Miller-Audacity:
I can reproduce this still within a newer version of the bottom bar ui changes: https://github.com/audacity/audacity/actions/runs/4574686550

petersampsonaudacity:
Testing on W10 with @crsib 's branch build:  audacity-win-3.3.0-alpha-20230410+1552625-x64

This issue appears to be fixed (and improved)
a) there is no longer the colorway deformation when turning snapping off
b) as an improvement the time format now remains as the user set it (rather than reverting to "seconds")

![image](https://user-images.githubusercontent.com/50205984/230882908-e6f7ff6e-5d8d-47bb-a5c6-006410b872eb.png)

dozzzzer:
Tested.
While testing, another minor related issue was found: https://github.com/audacity/audacity/issues/4533

PR #4523, opened 2023-04-10
Fixes display oddity when SnapTo combo is disabled

crsib:
Resolves: #4427

Recommended:


vsverchinsky (approved):

petersampsonaudacity:
Testing on W10 with @crsib 's branch build:  audacity-win-3.3.0-alpha-20230410+1552625-x64

This issue appears to be fixed (and improved)
a) there is no longer the colorway deformation when turning snapping off
b) as an improvement the time format now remains as the user set it (rather than reverting to "seconds")


image

Issue #4526, opened 2023-04-10
Change Pitch and Change Tempo effects crash with High Quality checkbox on


Paul-Licameli:
### Steps to reproduce

1. Generate sound
2. Open Change pitch or Change Tempo
3. Check the box at bottom of the dialog
4. Apply

### Expected behavior

Apply the effect

### Actual behavior

Crash!

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

See #4517 -- these are related

petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch:  audacity-win-3.3.0-alpha-20230410+5721047-x64

This now works properly with Paul's branch build, at step $ the effect is proper;y applied.  I tested with both Change Pitch and Change Tempo

Ryan-Miller-Audacity:
Tested and confirmed in Paul's branch.

petersampsonaudacity:
Testing on W10 with latest 3.3.0 Beta

This now works properly in the latest beta - at step 4 the effect is properly applied.  

I tested with both Change Pitch and Change Tempo

Issue #4517, opened 2023-04-07
Crash using Nyquist Prompt effect


SteveDaulton:
### Bug description

When applying the Noise Gate code via the Nyquist Prompt, Audacity crashes.

### Steps to reproduce

1. Generate a tone and apply a couple of fades (this provides a suitable signal for the effect to be clearly seen).
2. Apply Noise Gate (via the Effect menu) with Gate Threshold set to around -12 dB (other settings at default).
3. Observe that the effect works correctly.
4. Undo the Noise Gate effect.
5. Copy the code from noisegate.ny and paste into the Nyquist Prompt effect.
6. Apply with the same settings as used in step 2

### Expected behavior

The Noise Gate effect to be applied.

### Actual behavior

Audacity crashes.

### Audacity Version

current master / alpha / nightly build

### Operating system

Linux

### Additional context

Possibly related to https://github.com/audacity/audacity/issues/4498 though in this case the Debug window is not involved.

Regression against Audacity 3.2.4.

petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch:  audacity-win-3.3.0-alpha-20230410+5721047-x64

This now works properly with Paul's branch build, at step 6 the effect is properly applied.

Ryan-Miller-Audacity:
Tested and confirmed in Paul's branch.

petersampsonaudacity:
Testing on W10 with latest 3.3.0 Bet

This now works properly with the latest beta  -   at step 6 the effect is properly applied.

PR #4528, opened 2023-04-10
Effect delegation fixes

Paul-Licameli:
Resolves: #4517
Resolves: #4526

Recommended:

### DIFF
--- a/libraries/lib-effects/EffectBase.cpp
+++ b/libraries/lib-effects/EffectBase.cpp
@@ -164,7 +164,9 @@ bool EffectBase::DoEffect(EffectSettings &settings,
    std::shared_ptr<EffectInstance> pInstance;
 
    if (IsInteractive()) {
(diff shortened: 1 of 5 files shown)


libraries/lib-effects/EffectBase.cpp
```diff
@@ -164,7 +164,9 @@ bool EffectBase::DoEffect(EffectSettings &settings,
    std::shared_ptr<EffectInstance> pInstance;
 
    if (IsInteractive()) {
-      if (auto result = finder(settings))
+      if (!finder)
```
crsib, line 167: How could I reach this branch?

Paul-Licameli: Build this commit, not the last one, and do the steps to reproduce either issue.  It changes a crash to a quiet failure of the effects.

Issue #4501, opened 2023-04-06
Selection bar: with bar:beat:tick format, length is incorrect


DavidBailes:
### Steps to reproduce

1. Set the format of the numeric controls in the selection bar to bar:beat:tick
2. Choose an option in the selection bar that includes length, eg start and length.
3. Have no audio selected - ie there is a cursor.
4. Read the length of the selection

### Expected behavior

0 bar 0 beat 0 tick

### Actual behavior

1 bar 1 beat 1 tick

### Audacity Version

latest beta

### Operating system

Windows

petersampsonaudacity:
Testing on W10 with @crsib 's branch build:  audacity-win-3.3.0-alpha-20230410+431f64b-x64

This now works properly in Dmitry's branch build - see image below:

![image](https://user-images.githubusercontent.com/50205984/231109509-e1171746-9ddf-458a-ac44-81b0e2b7eb07.png)

dozzzzer:
Tested

PR #4529, opened 2023-04-10
Adds DURATION as the NumericConverter type

crsib:
Resolves: #4501 

Recommended:


libraries/lib-numeric-formats/formatters/ParsedNumericConverterFormatter.cpp
```diff
@@ -170,7 +170,7 @@ class ParsedNumericConverterFormatter final :
    {
       UpdateFormat();
 
-      if (mType == NumericConverterType_TIME)
+      if (IsTimeRelatedFormat())
```
Paul-Licameli, line 173: I understand this substitution, but then I wanted to check that it was done completely.

I see an omission in `ParsedNumericConverterFormatterFactory`.

libraries/lib-numeric-formats/formatters/BeatsNumericConverterFormatter.cpp
```diff
@@ -331,25 +335,39 @@ class BeatsNumericConverterFormatterFactory final :
...
+NumericConverterItemRegistrator beatsDuration {
+   Registry::Placement { {}, { Registry::OrderingHint::After, L"parsedDuration" } },
+   BuildBeatsGroup(false)
 };
 } // namespace
 
 std::unique_ptr<NumericConverterFormatter> CreateBeatsNumericConverterFormatter(
-   const FormatterContext& context, int fracPart /*= 0*/)
+   const FormatterContext& context, int fracPart /*= 0*/,
+   bool timeFormat /*= true*/)
```
Paul-Licameli, line 370: This function is called only from unit tests.  Is it easy and sensible to add some test cases?

libraries/lib-numeric-formats/NumericConverter.cpp
```diff
@@ -101,6 +101,19 @@ void NumericConverter::ControlsToValue()
                mInvalidValue;
 }
 
+bool NumericConverter::SetTypeAndFormatName (const NumericConverterType& type, const NumericFormatSymbol& formatName)
```
Paul-Licameli, line 104: Not really related to this PR, but I scanned for all occurences of `NumericConverter` and notice mentions of it in `ParsedNumericConverterFormatter.cpp` in comments, which seem no longer correct.

Paul-Licameli: Also the comment in `NumericConverter.h` refers to the wrong place to find an explanation of the mini language.

Paul-Licameli: And `NumericTextCtrl.h` duplicates that comment

src/widgets/NumericTextCtrl.cpp
```diff
@@ -235,6 +235,16 @@ void NumericTextCtrl::UpdateAutoFocus()
    }
 }
 
+bool NumericTextCtrl::SetTypeAndFormatName(const NumericConverterType& type, const NumericFormatSymbol& formatName)
+{
```
Paul-Licameli, line 239: This function shadows the base class function but does not override it.  Yet it does override other functions.

As you have mentioned, and I agree, ideally the control has-a converter, not is-a.  But, that's for future.

Paul-Licameli (changes requested):
See my first comment in this conversation.  Please fix it, or explain why it is not an error.

crsib:
I have pushed fixes, I will squash them with the first commit before merging

libraries/lib-numeric-formats/formatters/ParsedNumericConverterFormatter.cpp
```diff
@@ -1019,9 +1019,10 @@ class ParsedNumericConverterFormatterFactory final :
        : mType { std::move(type) }
        , mFormat { std::move(format) }
    {
-      // Only TIME with # in the format depends on sample rate now
-      mDependsOnSampleRate = mType != NumericConverterType_TIME ||
-                             mFormat.Debug().find(L'#') != wxString::npos;
+      // # in the format means that Sample Rate is used
+      // to calculate the values. Otherwise, Sample Rate is optional for
+      // the TIME and DURATION formats and ignored by the others.
+      mDependsOnSampleRate = mFormat.Debug().find(L'#') != wxString::npos;
```
Paul-Licameli, line 1025: But now this means that frequency and bandwidth converters depended on sample rate before, but now they do not.  Is that correct?

Paul-Licameli: Yes I see it is correct, now that I reexamined all uses of `mSampleRate`.  It really is only those formats contianing `#` that have any dependency on it.

Paul-Licameli (approved):

PR #4535, opened 2023-04-12
Variant utilities

Paul-Licameli:

Variant::VariantVisitor is a useful utility template for defining visitors of variants, uniting any
number of callables into one set of overload resolution candidates.

Variant::Visit() now simulates the one-visitor case of C++17 std::visit more exactly according to the standard.

This includes perfect forwarding of return values, the requirement that all alternatives return exactly the same type,
allowing the variant argument to be a proper subclass of a std::variant specialization, and sfinae to exclude other
calls to Visit from overload resolution.

Also, the performance guarantee of the standard, that dispatch is constant-time.

Recommended:


crsib (approved):
Pretty awesome!

crsib (changes requested):
I'll like the proposed changes, but GCC does not :-)

Paul-Licameli:
> I'll like the proposed changes, but GCC does not :-)

It may be GCC isn't doing CTAD correctly and a `make_visitor` function like `make_pair` may be needed after all

Paul-Licameli:
This idea needed much more work than I thought to get all the details really right.

Please review the whole again from the beginning.

There are also now many more compile-time tests (so we know the cases don't violate the `static_assert`s) and run-time tests of various combinations of values in the visitor, and constness and reference types of the arguments and the expected return value type.

libraries/lib-utility/Variant.h
```diff
@@ -31,57 +33,76 @@ struct VisitHelperReturn {
...
+   assert(pValue);
+   if constexpr (std::is_lvalue_reference_v<Variant>)
+      return std::invoke(std::forward<Visitor>(vis), (*pValue));
+   else
+      return std::invoke(std::forward<Visitor>(vis), std::move(*pValue));
+}
+
+//! Help to define Visit() below
+template <size_t Index, typename Visitor, typename Variant>
+auto TypeCheckedVisitHelperFunction(Visitor &&vis, Variant &&var)
```
crsib, line 67: I wonder if `decltype(auto)` is needed here as well

crsib: Or, sorry, I have missed the `->` arrow

libraries/lib-utility/Variant.h
```diff
@@ -101,18 +132,18 @@ VisitHelper(std::index_sequence<Indices...>, Visitor &&vis, Variant &&var)
  so let's use this even when not needed on the other platforms, instead of
  having too much conditional compilation
  */
-template <typename Visitor, typename Variant>
+template <typename Visitor, typename Variant,
+   typename VariantBase = detail::deduced_variant<Variant &&>>
```
crsib, line 136: Pretty cool!

crsib (approved):
Looks fantastic!

Issue #4470, opened 2023-03-30
The units of the timeline cannot be set using the keyboard


DavidBailes:
### Steps to reproduce

Using the keyboard, try to set the units of the timeline to beats and measures

### Expected behavior

It's possible

### Actual behavior

It's not possible

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

This can be done in Reaper, using a sub menu on the View menu.

petersampsonaudacity:
Tested this on W10 with latest 3.3.0 Beta and this works fine from the Extra menu.

I have updated the Manual for this.

PR #4537, opened 2023-04-12
Allow setting the ruler type from the menu

crsib:
Resolves: #4470 

Recommended:


petersampsonaudacity:
Tested on W10 with @crsib 's branch:  audacity-win-3.3.0-alpha-20230412+296cab1-x64

This works fine in Dmitry's branch - it is obtainable from the **Extra** menu:
image

vsverchinsky (approved):

PR #4540, opened 2023-04-13
Typelists

Paul-Licameli:

A small metprogramming library for typelist manipulations

Recommended:


Paul-Licameli:
I'm labelling this as "code quality" because it also contains a little bit of simplification of older code that isn't really needed for other refactoring

saintmatthieu:
It certainly is an interesting read. The example (static unit-tests) are helpful not only to test but also to understand the code.
I'm having a hard time figuring out the motivation behind `TypeList`s, though. @Paul-Licameli could you shed some light on this ?

Paul-Licameli:
Tell me if you find any errors, or think anything could be more clearly commented.

Code review is also for knowledge sharing.  You should understand what I did well enough, that you can explain back to me what I was thinking.

(Those remarks apply generally to code reviews.)

Meanwhile if you found this entertaining, try this.  https://github.com/audacity/audacity/pull/4535

Paul-Licameli:
These pull requests for new utilities are cherry picked from a growing branch that I have not yet shared.  It will clean up the registry system so it is more convenient to use, and also better type checked.

See how the type-erased Registry.cpp uses only four subclasses of BaseItem, assuming they are an exhaustive partition of cases.  But there is not now check of this assumption when groups are built.  I want to make a compile time check and will find TypeList convenient to express this in a `static_assert`.

But also, each different re-use of Registry for special purposes will define its own concrete subclasses.  For instance `CommandItem`, etc. just for the menu registry, which are not known types in Registry.h.  I also want a way to pass a type list to Registry.h as a template parameter, and so express more static checks of types.

TypeList will find other applications too.  There is yet another branch. https://github.com/audacity/audacity/pull/4290/files

I did some other metaprogramming there that could have been simplified if I had invented TypeList earlier.  I want to redraft that other one.

There is also some older stuff in Track.h, implementing `TypeSwitch` and `Visit` in a generic way so that Track.h does not depend on the subclasses of Track.  There are opportunities to use TypeList there, and maybe separate a general utility for such type switching and otherwise clean it up.

This pull request also contains one small application of TypeList in the second commit.

crsib:
When I first read `Modern C++ Design`, which introduces Type Lists I though "OMG, it is so cool, I need to use it right now". And you know - I failed. Some 15 years later I am still not convinced that Type Lists have a lot of practical applications. Unless, of course, you want to have a lot of free time during your day while your compiler tries to crunch. 

And looking through this PR I can see, that C++11 made type lists even more complex!  

The way it is used right now is the most insane and unreadable way to simply define `std::variant<wxString&, int&,...>` and `std::variant<const wxString&, const int&,...>`. I don't agree that this types should be used in the first place, but this is something of an ancient history. However the way it was done before was much shorter and easier to understand. 

Instead of relying  on standard library you imply that someone who runs into this code has a lot of understanding of some very Audacity specific implementation of Loki/Boost MPL. This complicates further Audacity development and I don't see any practical gain from it. 

In general, while I have pretty much nothing to say about the first commit (and we can even put it into the Audacity source tree), I have strong feelings about the second one.

Paul-Licameli:
> When I first read `Modern C++ Design`, which introduces Type Lists I though "OMG, it is so cool, I need to use it right now". And you know - I failed. Some 15 years later I am still not convinced that Type Lists have a lot of practical applications. Unless, of course, you want to have a lot of free time during your day while your compiler tries to crunch.
> 
> And looking through this PR I can see, that C++11 made type lists even more complex!
> 
> The way it is used right now is the most insane and unreadable way to simply define `std::variant<wxString&, int&,...>` and `std::variant<const wxString&, const int&,...>`. I don't agree that this types should be used in the first place, but this is something of an ancient history. However the way it was done before was much shorter and easier to understand.

Don't you know, std::variant cannot contain reference types?  Check that at cppreference.com, and look closer at PluginManager.h.  The type alias is actually for
```
std::variant<
   std::reference_wrapper<wxString>,
   std::reference_wrapper<int>,
   std::reference_wrapper<bool>,
   std::reference_wrapper<float>,
   std::reference_wrapper<double>
>
``` 
and another even longer type would describe the const version.

In PluginManager, using a metafunction to generate the type (whether implemented the old or the new way) reduces verbosity and improves clarity, if you just take the metafunction as given.  And indeed, maybe this metafunction ought to move into Variant.h for potential reuse.

What matters is parametrizing the type generation clearly by just mentioning the essentials, the contained types, and you could simply edit just that list of types if need arises.

Maybe this small example isn't compelling enough for applying TypeList, and the second commit is not essential to anything else I need to do.  I hope I will change your opinion when you see the larger applications of TypeList than this small example in Plugin manager.

There will be need to pass multiple packs of types around as template parameters, and rather than abuse std::tuple for that purpose I want something distinct.

> 
> Instead of relying on standard library you imply that someone who runs into this code has a lot of understanding of some very Audacity specific implementation of Loki/Boost MPL. This complicates further Audacity development and I don't see any practical gain from it.
> 

The standard library doesn't yet have an equivalent of what I do here, letting me do map, filter, and all the other familiar basic functional programming things.

One will not need "a lot" of understanding to use the TypeLists, guided by the examples in TypeList.cpp, without studying all the details of TypeList.h.  The notation in the examples is not really complicated.  If you are used to `std::conditional_t` and other `type_traits`, then this is not much stranger.

C++11 actually makes notation for a non-lazy type list easier, because you can just use a variable length parameter back, rather than something like `pair<int, pair<float, nil>>`.  I know from experience in the early 00s (when MPL and Loki were invented), when I also made my own typelists, that you had to define a Nil and then use iterated application of std::pair or similar.  It was uglier than this.

MPL, I understand, includes lazily generated typelists and the notion of a meta-iterator!  I'm not attempting that generality, but if you do then I think you do have to go back to the 1998 non-variadic style of lists as interated pairings of heads with tails.

> In general, while I have pretty much nothing to say about the first commit (and we can even put it into the Audacity source tree), I have strong feelings about the second one.

Even with the simplicity of variadics, always with a good library there may be complexity of implementation but the interface hides it.  The examples of use in `TypeList.cpp` are meant to look simple, really.  If you know a bit of Lisp, you should recognize this as like Lisp with just some notational change.

Paul-Licameli:
I mention `map` and `filter`, but also `apply` is a great convenience.  From TypeList.cpp:

```
static_assert(is_same_v<tuple<int, double>, Apply_t<tuple, Example>>);
```

In other words `Apply_t<tuple, Example>` neatly generates a type for you, and you can easily vary the first argument, `Apply_t<variant, Example>`.

So -- just pass a pack of types around as a single thing, and combine it ad hoc with any other variadic template.

See the old version of PluginManager.cpp -- where, without Apply, you must first define a primary template, then a partial specialization, to generate the type.   Apply lets you avoid repetition of that patter each time you do similar things.

crsib:
> Don't you know, std::variant cannot contain reference types?

Sure. I would push further and say that the need for variant of references is an indication of the problems with code. That is what I explicitly stated in my comment. 

> and another even longer type would describe the const version.

Which is much more clear BTW. However, I was comparing old vs new. I wasn't even mentioning the *explicit* way.

> reduces verbosity and improves clarity

I disagree. I've tried to read it - it is literally hard. So I fell back comparing the old approach to see if the new is equivalent. Surely I am not as brilliant as you are, but you have to live with me for now.

> One will not need "a lot" of understanding to use the TypeLists

I disagree. It is completely unclear what the code is doing.

>  If you know a bit of Lisp

So you expect every potential read of an open source code base to know Lisp?

> See the old version of PluginManager.h

I find the old version to be easier to understand. Verbosity is not something bad. Do not replace verbosity with complex magic just for the sake of reducing. The code should be easy to read and understand, not short.

Paul-Licameli:
> > Don't you know, std::variant cannot contain reference types?
> 
> Sure. I would push further and say that the need for variant of references is an indication of the problems with code. That is what I explicitly stated in my comment.

See commit 3acbcb4c2d8ada1e74991be79ecb3844568bcc0a where it began.  See how much repetitious code was eliminated with variants and visitors.  I prefer it this way.

> 
> > and another even longer type would describe the const version.
> 
> Which is much more clear BTW. However, I was comparing old vs new. I wasn't even mentioning the _explicit_ way.
> 
> > reduces verbosity and improves clarity
> 
> I disagree. I've tried to read it - it is literally hard. So I fell back comparing the old approach to see if the new is equivalent. Surely I am not as brilliant as you are, but you have to live with me for now.
> 
> > One will not need "a lot" of understanding to use the TypeLists
> 
> I disagree. It is completely unclear what the code is doing.
> 
> > If you know a bit of Lisp
> 
> So you expect every potential read of an open source code base to know Lisp?

Frankly, yes, everybody should have a little bit of exposure to the ideas basic list manipulation with one notation or another.  They don't need to read the whole Common Lisp HyperSpec.

And this is just a slightly different notation, for compile-time manipulation of types.  And I expect C++ programmers not to find compile time manipulation of types a strange idea.  Everybody uses `vector` and from there you advance.

> 
> > See the old version of PluginManager.h
> 
> I find the old version to be easier to understand. Verbosity is not something bad. Do not replace verbosity with complex magic just for the sake of reducing. The code should be easy to read and understand, not short.

Please see the new commit.

PluginManager.h just invokes a utility with a descriptive name.  You can read it, just trusting the implementation to do what it describes, to understand the intent.

It's just like using `std::binary_search`.  A descriptive name hiding details elsewhere that you don't need to study.

libraries/lib-utility/TypeListVisitor.h
```diff
@@ -0,0 +1,41 @@
...
+**********************************************************************/
+#ifndef __AUDACITY_TYPELIST_VISITOR__
+#define __AUDACITY_TYPELIST_VISITOR__
+
+#include "TypeList.h"
+#include <type_traits>
+#include <variant>
+
+namespace TypeListVisitor {
+using namespace TypeList; using namespace std;
```
crsib, line 19: I'd love to see `using namespace std` removed here, but no string feelings about it

crsib (approved):
With the latest two commits I can approve this PR, after some squashing.

crsib:
And provided that it builds xD

Paul-Licameli:
And if I don't even use TypeList.h anywhere, I think we can agree this one can bypass QA.  It's just then the new files plus the little movement of the bit of Obj-C++.

crsib:
> I will withhold the changes to PluginManager.h, saving them for a later pull request.

You can include them, with the VariantOfReferences_t approach. I am totally fine with that. 

Given that it only affects the way the variant with code I am good with bypassing QA even if you choose to keep PluginManager.h changes.

Paul-Licameli:
Also, the MaybeConst metafunction will be so freqently needed, I should add it to TypeList.h, and also the mapping of a given type list by it.  Then TypeListVisitor.h wouldn't mention `conditional_t` and it could look like

```
template<bool Const, typename TypeList> struct VariantOfReferences
   : Apply<variant, Map_t<Fn<reference_wrapper>, MapConst_t<Const, TypeList>>>
{};
```

PR #4545, opened 2023-04-14
change nyquist limit to 64 bit

LWinterberg:
Resolves: #1607 

based on the changes in #3051

### DIFF
--- a/src/effects/nyquist/Nyquist.cpp
+++ b/src/effects/nyquist/Nyquist.cpp
@@ -110,7 +110,7 @@ enum
 };
 
 // Protect Nyquist from selections greater than 2^31 samples (bug 439)
-#define NYQ_MAX_LEN (std::numeric_limits<long>::max())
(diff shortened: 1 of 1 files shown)


src/effects/nyquist/Nyquist.cpp
```diff
@@ -916,20 +916,7 @@ bool NyquistEffect::Process(EffectInstance &, EffectSettings &settings)
...
-                     .Format((long)NYQ_MAX_LEN, hours);
-               EffectUIServices::DoMessageBox(*this,
-                  message,
-                  wxOK | wxCENTRE,
-                  XO("Nyquist Error") );
-               if (!mProjectChanged)
-                  em.SetSkipStateFlag(true);
-               return false;
-            }
+            wxASSERT(mCurLen <= NYQ_MAX_LEN);
```
crsib, line 919: This assertion is a no-op really.

PR #4551, opened 2023-04-16
Anonymous items

Paul-Licameli:

Clarify one of the larger confusions about how the Registry system is meant to be used,
about so-called "Transparent" items.

Some other changes of terminology too.

Recommended:

### DIFF
--- a/libraries/lib-registries/Registry.cpp
+++ b/libraries/lib-registries/Registry.cpp
@@ -103,7 +103,7 @@ struct CollectedItems
       const BaseItemPtrs &toMerge, const OrderingHint &hint ) -> void;
 };
 
-// When a computed or shared item, or nameless grouping, specifies a hint and
+// When a computed or indirect item, or nameless grouping, specifies a hint and
 // the subordinate does not, propagate the hint.
 const OrderingHint &ChooseHint(BaseItem *delegate, const OrderingHint &hint)
 {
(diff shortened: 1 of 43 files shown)


crsib (approved):
Much better naming indeed!

crsib:
I think it can go straight to master

libraries/lib-registries/Registry.cpp
```diff
@@ -377,28 +391,30 @@ auto CollectedItems::MergeWithExistingItem(
...
             // Subordinate items from one of the groups will be merged in
             // another call to MergeItems at a lower level of path.
-            // Note, however, that at most one of the two should be other
-            // than a plain grouping item; if not, we must lose the extra
+            // Note, however, that at most one of the two should be a
+            // strongly ordered item; if not, we must lose the extra
             // information carried by one of them.
-            bool pCollectionGrouping = pCollectionGroup->Transparent();
-            auto pRegistryGrouping = pRegistryGroup->Transparent();
+            bool pCollectionGrouping =
```
saintmatthieu, line 398: Could be const

saintmatthieu, line 398: Why the prefix `p` ? I imagined this was reserved to pointers.

libraries/lib-registries/Registry.h
```diff
@@ -122,21 +124,34 @@ namespace Registry {
...
          : BaseItem{ internalName }, items{ std::move( items_ ) }
       {}
       GroupItem( const GroupItem& ) PROHIBITED;
       ~GroupItem() override = 0;
 
-      // Whether the item is non-significant for path naming
-      // when it also has an empty name
-      virtual bool Transparent() const = 0;
+      //! Choose treatment of the children of the group when merging trees
+      enum Ordering {
```
saintmatthieu, line 142: Have you considered using `enum class` ? One wouldn't get false positives searching the code for 
e.g. `Ordering::Strong` as opposed to just `Strong`, for example.

src/Menus.cpp
```diff
@@ -103,46 +103,50 @@ void MenuManager::UpdatePrefs()
    mStopIfWasPaused = true;  // not configurable for now, but could be later.
 }
 
-void MenuVisitor::BeginGroup( Registry::GroupItem &item, const Path &path )
+void MenuVisitor::BeginGroup( Registry::GroupItemBase &item, const Path &path )
```
saintmatthieu, line 106: It was confusing to find the declaration of this guy not in Menus.h but CommandManager.h

saintmatthieu (commented):
The code is unfortunately too dense for me to make sense of in a half a day. If more attention is requested from me to this PR please let me know, otherwise I might leave it at that and leave only these superficial comments.

src/Menus.cpp
Paul-Licameli: Yes:  Menus and CommandManager are tangled up badly, and there is another big pull request in progress to sort that out.

PR #4554, opened 2023-04-17
Qt time toolbar

ksoze95:
Resolves: #4346

Implemented initial version of the TimeControl for the QML/Qt port.

Recommended:


crsib (approved):
A good starting point, but a huge rework will be needed to integrate it with `lib-numeric-formats`

Issue #4556, opened 2023-04-17
Snapping: keyboard commands with default shortcuts left and right arrows etc don't snap to correct units


DavidBailes:
### Steps to reproduce

1. Create an audio track.
2. Set the format in the selection toolbar to seconds
3. Set snapping on, and set the combo box to milliseconds (or anything apart from seconds)
4. Use the keystrokes left/right arrow, shift + left/right arrow, ctrl+shift+left/right arrow.

### Expected behavior

Either:
1. the positions of the cursor or the start and end of selections are snapped to the value of the combo box in the snapping toolbar.
2. the positions are not snapped. (left and right arrow shortcuts in Reaper which move the cursor are not snapped when snapping is on)

### Actual behavior

the positions of the cursor or the start and end of selection are snapped to seconds, not the value of the combo box in the snapping toolbar.

### Audacity Version

latest beta

### Operating system

Windows

petersampsonaudacity:
Testing on W10 with latest 3.3.0 Beta

I confirm David's finding

a) snapping works based on the time format format in set in the snapping toolbar for manually placing the cursor with a mouse click

b) but with using the left and right arrows it looks to me as though snapping then is still based on the format set for time in the Selection toolbar (as indeed was the case in 3.2.5 and earlier.   

Surely the use of the arrows should snap to the same time format as when manually placing the cursor, i.e. the format set in the snapping toolbar.

petersampsonaudacity:
Testing on W10 with @crsib 's branch build for this fix:  audacity-win-3.3.0-alpha-20230417+90bbe68-x64

This issue appears to be properly fixed in Dmitry's branch build.

I tested:
a) left/right arrows  -  moves the cursor by the snapping format
b) Shift + left/right arrows  -  expands the selection by the snapping format
c) Ctrl + Shift + left/right arrows  -  shrinks the selection by the snapping format

petersampsonaudacity:
Re-tested again with @crsib 's very latest branch build:  audacity-win-3.3.0-alpha-20230418+026d33e-x64

All still works as expected wot this new branch build _(the one following @Paul-Licameli 's code review)_

dozzzzer:
Tested

PR #4564, opened 2023-04-17
Snapping when moving the cursor with keyboard

crsib:
Resolves: #4556

Recommended:

### DIFF
--- a/libraries/lib-snapping/SnapUtils.cpp
+++ b/libraries/lib-snapping/SnapUtils.cpp
@@ -15,8 +15,6 @@
 #include <cmath>
 #include <unordered_map>
 
-#include "ProjectRate.h"
-#include "ProjectTimeSignature.h"
 
 namespace
 {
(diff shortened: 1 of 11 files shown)


petersampsonaudacity:
Testing on W10 with @crsib 's branch build for this fix:  audacity-win-3.3.0-alpha-20230417+90bbe68-x64

This issue #4556 appears to be properly fixed in Dmitry's branch build.

I tested:
a) left/right arrows  -  moves the cursor by the snapping format
b) Shift + left/right arrows  -  expands the selection by the snapping format
c) Ctrl + Shift + left/right arrows  -  shrinks the selection by the snapping format

libraries/lib-snapping/SnapUtils.cpp
```diff
@@ -246,142 +250,100 @@ namespace
...
-      // DV: It was decided that for the time being,
-      // BPM sets the duration for quarter notes.
-      // For this reason, `cfg.timeSignature.second` is ignored
-      const auto fracDuration = quarterDuration * 4.0 / divisor;
-      const auto multiplier = 1.0 / fracDuration;
-
-      return SnapWithMultiplier(value, multiplier, nearest);
-   };
-}
+      if (!mMultiplierFunctor)
```
Paul-Licameli, line 311: Assertion in the constructor, and constancy of the member, should make the null check unnecessary

libraries/lib-snapping/SnapUtils.cpp
```diff
@@ -246,142 +250,100 @@ namespace
...
-      "cd", XO("CD frames"), false,
-      SnapFunction("cd_75_fps", XO("CDDA frames (75 fps)"), SnapToFrames(75.0)))
-};
+Registry::BaseItemPtr TimeInvariantSnapFunction(
+   const Identifier& functionId, const TranslatableString& label,
+   MultiplierFunctor functor)
+{
+   return std::make_unique<ProjectDependentMultiplierSnapItem>(
+      functionId, label, std::move(functor));
+}
```
Paul-Licameli, line 341: My latest PR "namespace-Callable" will conflict, but once this is merged into master, I will rebase, and this bit of new annoyance code will also simplify

src/menus/SelectMenus.cpp
```diff
@@ -166,26 +166,19 @@ void SeekWhenAudioActive(double seekStep, wxLongLong &lastSelectionAdjustment)
 double GridMove
 (AudacityProject &project, double t, int minPix)
 {
-   auto &formats = ProjectNumericFormats::Get(project);
```
Paul-Licameli, line 169: Good to see this; now, you can remove "ProjectNumericFormats.h" and "NumericConvert.h" from inclusions.

Paul-Licameli (approved):

libraries/lib-snapping/SnapUtils.cpp
crsib: We have very different view on assertions. There are very little cases where I would agree that assertion without a corresponding `if` is not a programming mistake.

src/menus/SelectMenus.cpp
crsib: done

libraries/lib-snapping/SnapUtils.cpp
Paul-Licameli: If that is your opinion, then make the other  constructor consistent.

Paul-Licameli: I meant to say, the other unchecked use of mMultiplierFunctor is inconsistent

crsib: Great catch, sorry for missing it in the first place

Paul-Licameli: I don’t like to review a new assertion that has zero comment with it.  The minimum here would be a ‘@pre’ at the declaration of the constructor.

PR #4572, opened 2023-04-18
Namespace callable

Paul-Licameli:

Move some existing utilities into a new file and namespace. 

Rename Variant::Visitor as Callable::OverloadSet.

Add unit tests.

Add UniqueMaker utility and tests.

Apply UniqueMaker to eliminate a lot of boilerplate code involving `...` and `std::forward`.

Recommended:


libraries/lib-numeric-formats/NumericConverterRegistry.h
```diff
@@ -90,24 +91,11 @@ struct NUMERIC_FORMATS_API NumericConverterRegistry final
...
-
-template <typename... Args>
-Registry::BaseItemPtr NumericConverterFormatterGroup(
-   const Identifier& groupId, const NumericConverterType& type,
-   Args&&... args)
-{
-   return std::make_unique<NumericConverterRegistryGroup>(
-      groupId, type, std::forward<Args>(args)...);
-}
+constexpr auto NumericConverterFormatterItem =
```
crsib, line 94: Neat!

crsib (approved):
I think this can go straight to master

Paul-Licameli:
> I think this can go straight to master

Recall, it will conflict with your recent changes, but the fixes for that will be simple.

Paul-Licameli:
@crsib please review again.

Changes needed after merging latest release into master were nontrivial.

You may or may not like the simplification of SnapUtils.cpp now.  You can tell me to remove those commits and let you do it yourself later.

libraries/lib-snapping/SnapUtils.h
```diff
@@ -87,12 +87,23 @@ struct SNAPPING_API SnapRegistryItem : public Registry::SingleItem
    virtual SnapResult SingleStep(const AudacityProject& project, double time, bool upwards) const = 0;
 };
 
-class ConstantMultiplierSnapItem final : public SnapRegistryItem
+using MultiplierFunctor = std::function<double(const AudacityProject&)>;
+
+class TimeInvariantSnapItem final : public SnapRegistryItem
```
crsib, line 92: I would much prefer the way it is done in previous commit. This still exposes details outside the translation unit (and in case we will want to add `SnapRegistryItem`s in different libraries - we will have to expose it outside the library too). However version with two classes express intention clearly and avoids (very small) storage and branching overhead.

I do not have any strong opinion on that though, so it is up to you.

crsib (approved):
I left some comments about the Snapping commits. Either way I do not have any strong feelings to request specific changes to this PR.

Paul-Licameli:
It was agreed that this can go straight to merge, so I will do that.

I will omit the changes in SnapUtils.cpp, which are not such obvious wins for simplification, and lead to compromised encapsulation.

I will let you decide how you would like to do it, after reading my examples.

A complete class definition does need to be in scope to use UniqueMaker.  Maybe you could use the `namespace detail` convention?

PR #4579, opened 2023-04-19
GCC13 support

crsib:

Recommended:

### DIFF
--- a/lib-src/libnyquist/nyquist/nyqstk/include/Stk.h
+++ b/lib-src/libnyquist/nyquist/nyqstk/include/Stk.h
@@ -46,7 +46,8 @@ typedef uint64_t uintmax_t;
 */
 #endif
 
-#include <stddef.h>
+#include <cstddef>
+#include <cstdint>
 #include <string>
 #include <iostream>
 #include <sstream>
@@ -355,10 +356,6 @@ typedef signed short SINT16;
 typedef signed int SINT32;
 typedef float FLOAT32;
 typedef double FLOAT64;
-// added by RBD
-#ifdef __linux__
-#define intptr_t long
-#endif
 
 // The default sampling rate.
 const StkFloat SRATE = 44100.0;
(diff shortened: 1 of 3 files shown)


lib-src/libnyquist/nyquist/nyqstk/include/Stk.h
```diff
@@ -355,10 +356,6 @@ typedef signed short SINT16;
 typedef signed int SINT32;
 typedef float FLOAT32;
 typedef double FLOAT64;
-// added by RBD
-#ifdef __linux__
-#define intptr_t long
-#endif
```
Paul-Licameli, line 361: Write to Roger, recommend this for upstream

libraries/lib-effects/LoadEffects.cpp
```diff
@@ -154,9 +154,9 @@ void BuiltinEffectsModule::AutoRegisterPlugins(PluginManagerInterface & pm)
       if (rediscoverAll ||
          !pm.IsPluginRegistered(path, &pair.second->name.Msgid())
       ){
-         DiscoverPluginsAtPath(path, ignoredErrMsg, [&](PluginProvider *provider, ComponentInterface *ident)
+         DiscoverPluginsAtPath(path, ignoredErrMsg, [&](PluginProvider *provider, ComponentInterface *ident) -> const PluginID&
          {
-            const auto pluginId = PluginManagerInterface::DefaultRegistrationCallback(provider, ident);
+            const auto& pluginId = PluginManagerInterface::DefaultRegistrationCallback(provider, ident);
```
Paul-Licameli, line 159: Really, the argument expected by DiscoverPluginsAtPath could be changed to a `std::function` returning void.  (But don't). I was surprised no crashed happend because of this error -- but the return value from the callback is always discarded, in all overrides of `DiscoverPluginsAtPath`.

Paul-Licameli (approved):

lib-src/libnyquist/nyquist/nyqstk/include/Stk.h
Paul-Licameli: Then it could be changed to
```
#if (defined __linux__) && !defined(__cplusplus)
```

Paul-Licameli:
> I think this go straight to merge, an I right @Paul-Licameli ?

Agreed

Issue #4612, opened 2023-04-26
custom project rate stays greyed out


LWinterberg:
When choosing "other" in the sample rate dropdowns, in the current version a text box right next to it should enable. This is not the case for the project rate, only for the default rate.
![image](https://user-images.githubusercontent.com/87814144/234560691-983c7310-5da0-46a6-be89-d5af9a3ebc9a.png)

LWinterberg:
![image](https://user-images.githubusercontent.com/87814144/234563799-9380ae86-cae0-4d16-94d9-323a96e8a918.png)
A workaround exists via Extra > Scriptables I > Set Project

dozzzzer:
Tested

PR #4615, opened 2023-04-26
Fixes the ability to set arbitrary project rate

crsib:
Resolves: #4612 

Recommended:


petersampsonaudacity:
Testing on W10 with @crsib 's branch build

This looks to be properly fixed in this branch build - I can now set both the **Project Sample Rate** and the **Default Sample Rate** to arbitrary values using the "**other**" settings.

image

Issue #4619, opened 2023-04-27
Play at speed slider: Using the keyboard it can be not straightforward to get back to a speed of 1.00


DavidBailes:
### Steps to reproduce

1. With the play at speed slider being the focus, and the initial speed 1.00.
2. Using either right/left arrow, or page up/down change the speed to either the min of max value, and then try to change it back to 1.00

### Expected behavior

You can easily change the speed back to 1.00

### Actual behavior

The closest you can get is 1.01

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

A work around for this is: Extra > Play-at-speed > Adjust playback speed, where you can set the speed to 1.00.

This is a minor bug, but was reported on the Audacity4Blind mailing list.

DavidBailes:
@LWinterberg Could the review of this fix be moved to 3.4. It's a simple safe fix.

crsib:
Honestly, we've discussed it and ... forgot about it. I have a strong opinion that step should be an integer, later divided by whatever divisor that fits. Floating point magic will always have unexpected corner cases.

DavidBailes:
> Honestly, we've discussed it and ... forgot about it. I have a strong opinion that step should be an integer, later divided by whatever divisor that fits. Floating point magic will always have unexpected corner cases.

I think that the issue is caused by the minimum play speed being 0.01 rather than 0.0. I don't think having integer steps would help.

crsib:
> being 0.01 

I don't think that this value is cut in stone. We can easily go with 1/60 (or 1/600, or whatever) as a minimum, so we have the slider ranger in [1, 120] range

crsib:
We can even have a special treatment of zero, remapping it to 0.01

DavidBailes:
> We can even have a special treatment of zero, remapping it to 0.01

I did think about that when I was writing my fix. From what I remember constants for range and increments are not in the same place, so it wasn't straightforward to write the special treatment that wouldn't break if someone changed one of the constants. I also thought that that my fix was good enough for most users, and certainly better than the existing code.
But if you think that the fix should be done better, I can close my PR.

PR #4620, opened 2023-04-27
Play at speed slider: Problem with step sizes for keyboard users.

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/4619

Problem:
Using the keyboard it can be not straightforward to get back to a speed of 1.00 after setting the slider to its minimum or maximum slider.

In void TranscriptionToolBar::Populate(), with the values set for options.Line() and options.Page(), and given how the step sizes are calculated in LWSlider::Increase() and LWSlider::Decrease(), the step sizes are 2.99/60, and 2.99/6.

Fix:
Change the values set for options.Line() and options.Page so that the step sizes are 3.00/60 and 3.00/6. Note: After this fix, a user will still encounter the same issue if they set the slider to the minimum value, and then try to return to 1.00, but they can work around this by setting the slider to the maximum value and then returning to 1.00. In addition most users will probably be using speeds of > 1.00.

Recommended:

### DIFF
--- a/src/toolbars/TranscriptionToolBar.cpp
+++ b/src/toolbars/TranscriptionToolBar.cpp
@@ -234,8 +234,8 @@ void TranscriptionToolBar::Populate()
       ASlider::Options{}
          .Style( SPEED_SLIDER )
          //  6 steps using page up/down, and 60 using arrow keys
-         .Line( 0.16667f )
-         .Page( 1.6667f )
+         .Line( 0.16722f )
+         .Page( 1.6722f )
    );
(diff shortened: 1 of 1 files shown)


saintmatthieu (approved):

LWinterberg:
@DavidBailes can you rebase this onto master to regenerate the builds? sorry for this massive delay

DavidBailes:
> can you rebase this onto master to regenerate the builds? sorry for this massive delay

Done.

PR #4623, opened 2023-04-27
Namespace composite

Paul-Licameli:

Utility class templates that can simplify some Registry code

Implement the Composite pattern of a container of nodes that is also of the node type.

Variadic and range constructors can take many components, also interposing type checks.

A template for easy definition of extensions that inherit the variadic and range constructors.

QA:  Nothing to do.  This introduces utilities but uses them nowhere yet but in unit tests.

Recommended:

### DIFF
--- a/libraries/lib-utility/Composite.h
+++ b/libraries/lib-utility/Composite.h
@@ -12,6 +12,8 @@
 #ifndef __AUDACITY_COMPOSITE__
 #define __AUDACITY_COMPOSITE__
 
+#include <cassert>
+#include <iterator>
 #include <type_traits>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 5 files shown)


Paul-Licameli:
Rebased to fix conflicts, and so I squashed the small build fix commits too

libraries/lib-utility/Composite.h
```diff
@@ -0,0 +1,66 @@
...
+//! back_inserter)
+/*!
+ @tparam Component common base class of nodes and leaves (and of Base itself),
+ which must have a virtual destructor
+ @tparam Item type of pointer to Components to be stored
+ @tparam ComponentArgs... passed to constructor of Component
+ */
+template<
+   typename Component,
+   typename Item,
```
crsib, line 34: So probably we can call it `ComponentPtr` or `PointerType`?

Paul-Licameli: Easily done, if you like, only affecting the template parameter locally.

Inside the class, it's aliased to `value_type`, named that specifically to make it work with `std::back_inserter`.

vsverchinsky (approved):

PR #4624, opened 2023-04-27
Simplify with composite

Paul-Licameli:

Depends on
- #4623

Use the Composite utility to simplify definitions of menu registry classes,
and simplify the registration of numeric format conversions and snap functions,
so they too can enjoy syntactic simplicity like menu descriptions.

Recommended:


Paul-Licameli:
I need to fix CI failures before this is reviewed

vsverchinsky (approved):

libraries/lib-numeric-formats/NumericConverterRegistry.h
```diff
@@ -38,21 +38,24 @@ using NumericConverterFormatterFactoryPtr =
...
-{
-   template <typename... Args>
-   NumericConverterRegistryGroup(
-      const Identifier& internalName, NumericConverterType _type,
-      Args&&... args)
-       : GroupItem { internalName, std::forward<Args>(args)... }
-       , type { std::move(_type) }
-   {
-   }
+struct NumericConverterRegistryGroupData {
```
crsib, line 40: I'm not quite sure what is the benefit of having one more type, I don't have any strong opinion though

libraries/lib-registries/Registry.h
```diff
@@ -300,7 +355,16 @@ namespace detail {
       Literal mRoot;
    };
 
-extern template struct GroupItem<>;
+#ifndef _WIN32
+// extern explicit instantiation to avoid link time de-duplication
+extern
```
crsib, line 360: There is [evidence](https://www.zverovich.net/2023/04/10/cxx20-modules-in-clang.html) that extern templates do not really work on Clang, the code is quite spooky on CL. Do we need them here? What is the measured effect of this?

crsib (approved):
I don't agree that this could be called a "simplification". This PR does some clean up. I have no objections. 

NumericTextCtrl should be tested carefully.

Paul-Licameli:
> I don't agree that this could be called a "simplification". This PR does some clean up. I have no objections.
> 
> NumericTextCtrl should be tested carefully.

I want to avoid rewriting variadic template constructors and std::forward a lot.

I want it to be easy to write more classes like those for menu trees, so it is easy to express other trees declaratively instead of building them up with procedural code.

If you review the other pull requests that follow, you can see registries also getting smarter about type checking the trees as they are constructed.

libraries/lib-registries/Registry.h
Paul-Licameli: I needed special code for Windows to link.  I might make it compile nothing otherwise.

dozzzzer:
Tested

PR #4634, opened 2023-04-29
portsmf/allegro.h: include <cstdint> for intXX_t

nekopsykose:
needed now with gcc13, not there by default

fixes

```
lib-src/portsmf/allegro.h:125:9: error: 'int64_t' does not name a type
  125 |         int64_t i;  // integer -- 64 bit for 32/64-bit compatibility
      |         ^~~~~~~
lib-src/portsmf/allegro.h:52:1: note: 'int64_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
   51 | #include <assert.h>
  +++ |+#include <cstdint>
```

Recommended:


crsib:
Hmm, I've built on Fedora few days ago without any issues

nekopsykose:
might be a slight difference in even what c++ headers are there when built against musl. this is specifically with gcc13 at 20230429 on a -linux-musl system.

even if it is that specific though, seems correct to include it for the sized ints anyway

crsib:
Absolutely

crsib (approved):

crsib:
Can you please rebase your branch on `release-3.3.2` and force push?

nekopsykose:
that then contains the version bump commits too, unless you mean to also target a different branch other than master; i assume this is what you want however

crsib:
Yeah, I just need a single commit with the fix on top of 3.3.2 branch :-)

crsib:
Awesome, thank you a lot!

PR #4641, opened 2023-05-01
Qt localisation

ksoze95:
Resolves: #4640

Recommended:


src/qt/ApplicationHandler.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <QtCore/QObject>
+#include <QtCore/QTranslator>
+#include <QtQml/qqmlregistration.h>
+
+class ApplicationHandler : public QObject
```
crsib, line 7: Should it be something like TranslationManager?

src/qt/ApplicationHandler.h
```diff
@@ -0,0 +1,21 @@
...
+{
+   Q_OBJECT
+   QML_ELEMENT
+
+public:
+   explicit ApplicationHandler(QObject *parent = nullptr);
+   virtual ~ApplicationHandler() = default;
+
+   Q_INVOKABLE void ChangeToEnglish();
```
crsib, line 16: Ideally it should be `ChangeLanguage(const QString& code)`

crsib (changes requested):
Please address the comments

ksoze95:
Rebased off latest audacity qt branch

src/qt/ApplicationHandler.h
ksoze95: Done

src/qt/ApplicationHandler.h
ksoze95: Done

crsib (approved):

ksoze95:
Rebased of latest qt branch

PR #4651, opened 2023-05-03
Registry type checking

Paul-Licameli:

QA:
- [x] Try some track context menu items
- [x] Sequence of items in the entire tree of pull-down menus is unchanged
- [x] Likewise, the tree in the left column of the Preferences dialog
- [x] Sequence of controls on the Library page of Preferences
- [x] Sequence of formats in timer controls of Selection toolbar
- [x] ditto snapping toolbar
- [x] sequence of choices for "Format" in File > Export Audio... dialog

Depends on
 - #4623
 - #4624 

Each Registry specifies an exhaustive list of its leaf and node item types in its Traits.

Registry trees are checked for type correctness as they are constructed or registered for later
merge, although this check can still be evaded by (deprecated) use of GroupItemBase::push_back
to build a tree by other means.

Visitors of registry items are allowed only const access to items.

This pull request is not mostly about simplification of the use of registries, but some of that
happened too.  See for instance how the definition of class `NumericConverterItemRegistrator`
simplifies to a type alias.

There is still more type checking to do, and significant simplifications to achieve, in registry
visiting, making use of the TypeSwitch utility to be extracted from Track.h (see #4650).

Recommended:

### DIFF
--- a/libraries/lib-import-export/ExportPluginRegistry.cpp
+++ b/libraries/lib-import-export/ExportPluginRegistry.cpp
@@ -65,10 +65,10 @@ void ExportPluginRegistry::Initialize()
       {
       }
 
-      void Visit( SingleItem &item, const Path &path ) override
+      void Visit(const SingleItem &item, const Path &path) override
       {
          mPlugins.emplace_back(
-            static_cast<ExportPluginRegistryItem&>( item ).mFactory() );
+            static_cast<const ExportPluginRegistryItem&>(item).mFactory() );
       }
 
       ExportPlugins& mPlugins;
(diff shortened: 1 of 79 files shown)


Paul-Licameli:
Rebased ont the fixed `simplify-with-Composite` branch.  Removed the failed fix for Windows build.  The branch just rebased on has a different fix which may work for this branch too.

Paul-Licameli:
Windows build still wasn't good, now trying something else, which is actually a good enhancement of Composite, not just a workaround

crsib (approved):
Quite a large PR

Paul-Licameli:
> Quite a large PR

Explain it to me.  It has been months since I wrote it…

Paul-Licameli:
But I hope it is nice that visitors can often be mere lambdas

Paul-Licameli:
Rebased, unchanged

Paul-Licameli:
And again

PR #4661, opened 2023-05-04
Fixes Snap and Time Signature toolbars appearance under wxGTK

crsib:

Recommended:

### DIFF
--- a/src/toolbars/TimeSignatureToolBar.cpp
+++ b/src/toolbars/TimeSignatureToolBar.cpp
@@ -61,7 +61,7 @@ TimeSignatureToolBar::TimeSignatureToolBar(AudacityProject& project)
             {
                if (mTempoControl)
                   mTempoControl->SetValue(settings.newTempo);
-               
+
                if (mUpperSignatureControl)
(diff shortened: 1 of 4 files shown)


src/toolbars/TimeSignatureToolBar.cpp
```diff
@@ -104,11 +104,16 @@ void TimeSignatureToolBar::Create(wxWindow* parent)
 
 void TimeSignatureToolBar::Populate()
 {
+#ifndef __WXGTK__
    const auto tempoSigSize = wxSize(60, -1);
    auto timeSigSize = wxSize(50, -1);
+#else
+   const auto tempoSigSize = wxSize(60, -1);
```
vsverchinsky, line 111: `tempoSigSize` values are same in both branches, is that correct?

vsverchinsky (approved):

src/toolbars/TimeSignatureToolBar.cpp
crsib: Yep

PR #4680, opened 2023-05-09
Stereo track restructuring

Paul-Licameli:
Resolves: #4681

Depends on
- #4754

Recommended:


libraries/lib-effects/PerTrackEffect.cpp
```diff
@@ -108,10 +108,7 @@ bool PerTrackEffect::ProcessPass(Instance &instance, EffectSettings &settings)
    };
 
    const bool multichannel = numAudioIn > 1;
-   auto range = multichannel
-      ? mOutputTracks->Leaders()
-      : mOutputTracks->Any();
-   range.VisitWhile( bGoodResult,
+   const auto &waveTrackVisitor =
```
saintmatthieu, line 111: Is this even allowed, a reference to a right value ?

Paul-Licameli: I wrote an unnecessary but also harmless `&` there which I will remove before committing.

There is a rule in C++ that when an initializer is bound to a local reference variable, then then lifetime of the object constructed on the right hand side is extended until the reference variable name goes out of scope.

Paul-Licameli:
This first attempt at implementation has been superseded by other approaches

PR #4684, opened 2023-05-10
More registry type checking

Paul-Licameli:
Resolves: #5436

Depends on:
- #4651

Simplify calls to visit registries.  (When only visiting leaf nodes of one type, the visitor can be just one lambda).

Eliminate many dynamic_casts.  Instead TypeSwitch hides all the casting.

This also makes better type safety.  GroupItemBase because a detail class that applications should not use.

QA:
- [x] All as for #4651 again
   - [x] Try some track context menu items
   - [x] Sequence of items in the entire tree of pull-down menus is unchanged
   - [x] Likewise, the tree in the left column of the Preferences dialog
   - [x] Sequence of controls on the Library page of Preferences
   - [x] Sequence of formats in timer controls of Selection toolbar
   - [x] ditto snapping toolbar
   - [x] sequence of choices for "Format" in File > Export Audio... dialog
- [x] Exercise at least one item of each context menu
   - [x] Common track drop-down items
   - [x] Special drop-down items of each of the four kinds of track
   - [x] Wave track color choices
   - [x] clip
   - [x] background area (adds tracks)
   - [x] Label box
   - [x] Vertical rulers of waveform, spectrum, and MIDI tracks
 

Recommended:

### DIFF
--- a/libraries/lib-import-export/ExportPluginRegistry.cpp
+++ b/libraries/lib-import-export/ExportPluginRegistry.cpp
@@ -59,24 +59,13 @@ void ExportPluginRegistry::Initialize()
       { {wxT(""), wxT("PCM,MP3,OGG,Opus,FLAC,WavPack,FFmpeg,MP2,CommandLine") } },
    };
 
-   struct CreatePluginsVisitor final : Visitor {
-      CreatePluginsVisitor(ExportPlugins& plugins)
-         : mPlugins(plugins)
-      {
-      }
-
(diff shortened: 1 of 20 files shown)


Paul-Licameli:
Rebased onto the updated dependency

crsib (approved):
This PR needs an extensive list of QA items

Paul-Licameli:
> This PR needs an extensive list of QA items

True

Thanks for reviewing.  Does your head hurt?

PR #4723, opened 2023-05-23
4717 relocate caching from sampletrackcache to some place below WaveClip

### LINKED ISSUE #4717: Relocate caching from SampleTrackCache to some place below WaveClip
### Your idea

Relocate caching from SampleTrackCache to SQLiteSampleBlock

### Problem to be solved

So far I've only come up with one reason for the SampleTrackCache, namely to limit the queries to SQLiteSampleBlock, which have a very large overhead.

However there are 3 problems with that :
1. StretchingSampleTrack is underneath SampleTrackCache, and when the latter pulls such a large amount of samples at once, the time-stretching processing time will cause a large outlier in the time it takes for SampleTrackCache::GetFloats to complete.
2. This propagates caching complexity rather high up in the stack. Why not doing it at the lowest level possible, namely in SQLiteSampleBlock?
3. Introducing time stretching makes the mapping from UI time to raw clip sample indices even more complex. I'd like to centralize this mapping logic and remove all sample index mentions from the WaveClip and WaveTrack API. SampleTrackCache uses these methods, though.

### Prior art

_No response_

### Additional context

I've already made experiments and measurements. In the figure below, I compare the time it takes for SampleTrackCache::GetFloats to execute without any caching (each GetFloats results in an SQL query), with caching as currently done, and with a quick-and-dirty caching (to be improved) I wrote in SQLiteSampleBlock. Observations :
* Caching does improve performance by an order of magnitude, reducing from an average of roughly 1ms down to 1us,
* The caching method still has outliers, namely the few calls where GetFloats results in an SQL query,
* caching in SQLiteSampleBlock works and provides equivalent performance, even roughly written.

GetFloats elapsed time

saintmatthieu:
Resolves: #4717 

Caching cannot be done above `WaveTrack` and `WaveClip` anymore, since its is done on the fly for large amount of samples (to reduce frequency of costly calls to SQLite DB) and time stretching cannot be done on such large chunks at once during playback without causing hick-ups in playback.

This also has the advantage of containing some of the complexity at a lower stack level.

Things to test, preferably on large project with many tracks and clips :
- [ ] playback also after changing track sample rate
- [ ] there is no noticeable CPU and memory usage increase
- [ ] looping and scrubbing still work
- [ ] Apply some effect (e.g. gain) to a part of a clip and play back
- [ ] Spectrum display
- [ ] Merge clips together


libraries/lib-wave-track/WaveClip.h
```diff
@@ -374,6 +370,9 @@ class WAVE_TRACK_API WaveClip final : public XMLTagHandler
    bool mIsPlaceholder { false };
 
 private:
+   //! Returns the index of the first sample of the underlying sequence
+   sampleCount GetSequenceStartSample() const;
```
Paul-Licameli, line 374: I'm starting the review in the middle, with the commit "Remove now obsolete API methods".  It makes the most widespread changes, by changing the interface of WaveTrack, and I want to understand that first as perhaps what motivates much else.

My first comment:  WaveClip::ToSequenceSamples becomes unused so could be removed.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -168,37 +162,25 @@ bool SpecCache::CalculateOneSpectrum
...
-      if (copy || !useBuffer)
-         useBuffer = scratch;
 
       if (autocorrelation) {
          // not reassignment, xx is surely within bounds.
          wxASSERT(xx >= 0);
          float *const results = &out[nBins * xx];
-         // This function does not mutate useBuffer
-         ComputeSpectrum(useBuffer, windowSizeSetting, windowSizeSetting,
+         // This function does not mutate adj
```
Paul-Licameli, line 182: Should the comment instead say `scratch`?

libraries/lib-wave-track/SequenceSampleCache.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
```
Paul-Licameli, line 1: Add banner comments

libraries/lib-wave-track/SequenceSampleCache.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include "SampleCount.h"
+#include "SampleFormat.h"
+
+#include <array>
+#include <functional>
+#include <optional>
+
+class SequenceSampleCache
```
Paul-Licameli, line 10: Add some Doxygen comments about the purpose of the class

libraries/lib-wave-track/SequenceSampleCache.h
```diff
@@ -0,0 +1,40 @@
...
+      const sampleCount sequenceStartSample;
+      const size_t numSamples;
+      SampleBuffer buffer;
+   };
+
+   const SourceSampleGetterFn mGetterFn;
+   const sampleCount mSequenceLength;
+   const size_t mMaxBlockSize;
+   std::array<std::optional<Block>, 2> mBlocks;
+   std::optional<size_t> mLastContainingBlockIndex;
```
Paul-Licameli, line 39: Very minor, but I think about shrinking the `sizeof` and `alignof` structs when I define them.  I might think to permute the members, putting `mMaxBlockSize` last.  And then forget about it.

libraries/lib-wave-track/SequenceSampleCache.cpp
```diff
@@ -0,0 +1,77 @@
+#include "SequenceSampleCache.h"
+
+#include <cassert>
+
+SequenceSampleCache::SequenceSampleCache(
+   SourceSampleGetterFn fn, sampleCount sequenceLength, size_t maxBlockSize)
+    : mGetterFn(std::move(fn))
+    , mMaxBlockSize(maxBlockSize)
+    , mSequenceLength(sequenceLength)
```
Paul-Licameli, line 9: The sequence of member initializers does not agree with the sequence that the class declares the memebrs.

It's legal but will compile to execute in the sequence of the members, not the sequence of the initializers seen here.

Some compilers warn about that.

In this case it doesn't matter because there are no side effects or ordering dependencies among the initializers.  But it is good habit always to make the sequences agree.

libraries/lib-wave-track/SequenceSampleCache.cpp
```diff
@@ -0,0 +1,77 @@
...
+         len - numSamplesToRead, mayThrow);
+   }
+}
+
+SequenceSampleCache::Block::Block(
+   size_t numSamples, sampleFormat format, sampleCount sequenceStartSample)
+    : format(format)
+    , sequenceStartSample(sequenceStartSample)
+    , numSamples(numSamples)
+    , buffer(numSamples, format)
```
Paul-Licameli, line 69: My preference is to use new uniform initialization syntax for members
```
: format{ format }
```
which has an advantage of disallowing narrowing conversions.  But one must know about a few dangers, where you must use parentheses to disambiguate.  Such as, constructing a vector and specifying an initial size instead of vector contents.

libraries/lib-wave-track/SequenceSampleCache.cpp
```diff
@@ -0,0 +1,77 @@
...
+    , mSequenceLength(sequenceLength)
+{
+}
+
+bool SequenceSampleCache::Get(
+   samplePtr buffer, sampleFormat format, sampleCount start, size_t len,
+   bool mayThrow)
+{
+   auto containingBlockIt = std::find_if(
+      mBlocks.begin(), mBlocks.end(),
```
Paul-Licameli, line 18: The terminology throws me a bit.  "Block" is already much used in Sequence itself with a different meaning, and there can be any number.  But "blocks" here are just two.  I suggest instead, "slots".

libraries/lib-wave-track/SequenceSampleCache.cpp
```diff
@@ -0,0 +1,77 @@
...
+            0u;
+      const auto numSamplesToCache =
+         std::min(sampleCount { mMaxBlockSize }, mSequenceLength - start)
+            .as_size_t();
+      auto& block = mBlocks[blockIndex];
+      block.emplace(numSamplesToCache, format, start);
+      if (!mGetterFn(
+             block->buffer.ptr(), format, start, numSamplesToCache, mayThrow))
+      {
+         assert(false);
```
Paul-Licameli, line 36: I think this misuses the assertion.  If it's part of the contract that the getter function cannot fail, then just give it a void return.  If failure is tolerable (and indeed you handle the case next), then don't assert.

And I would do the latter, knowing what I do about the history.  The `mayThrow` arguments ultimately cause exception propagation when true in case there is some sort of disk failure.  Exceptions lead to a top-level handler in the program that backs out the side effects of a failed operation and restores the last good state out of undo history.  See `AudacityApp::OnExceptionInMainLoop`.

But in some contexts (painting the screen) false is passed around instead to suppress exceptions, and it's expected to fill sample buffers with zero and give that to painting.

Failure conditions because of exogenous device problems or correupted files are contingencies to be handled, and it is not proper to assert that they never happen.

As I said, I don't like new assertions that have no accompanying comments anywhere, unless they are in the middle of the function and very easy to deduce as always correct in the context.

A new assertion should mean "This condition was provable at the time the assertion was written."

You should know, and persuade your reviewer too, that assertion failure is LOGICALLY impossible at the time you add the assertion.

I see assertions as a kind of documentation, not just there for their runtime effects in debug builds.

Now maybe later code modification breaks the proof because it is hard to maintain all the consistency.  That's when the runtime effect of assertions is also valuable.

libraries/lib-wave-track/SequenceSampleCache.cpp
```diff
@@ -0,0 +1,77 @@
...
+      mBlocks.begin(), mBlocks.end(),
+      [start](const std::optional<Block>& block) {
+         return block.has_value() && block->Contains(start);
+      });
+   if (containingBlockIt == mBlocks.end())
+   {
+      const auto blockIndex =
+         mLastContainingBlockIndex.has_value() ?
+            (*mLastContainingBlockIndex + 1) % mBlocks.size() :
+            0u;
```
Paul-Licameli, line 27: This is "LRU", isn't it?  Least-recently-used.

libraries/lib-wave-track/SequenceSampleCache.cpp
```diff
@@ -0,0 +1,77 @@
...
+      containingBlockIt = mBlocks.begin() + blockIndex;
+   }
+   const auto& block = *containingBlockIt;
+   const auto sampleOffset = (start - block->sequenceStartSample).as_size_t();
+   const auto numSamplesToRead =
+      std::min(len, block->numSamples - sampleOffset);
+   const auto numBytesToRead = numSamplesToRead * SAMPLE_SIZE(format);
+   const auto blockPtr =
+      block->buffer.ptr() + sampleOffset * SAMPLE_SIZE(format);
+   std::copy(blockPtr, blockPtr + numBytesToRead, buffer);
```
Paul-Licameli, line 51: Maybe `memcpy` instead, or maybe compilers are smart enough to make that optimization

libraries/lib-wave-track/SequenceSampleCache.cpp
```diff
@@ -0,0 +1,77 @@
...
+   {
+      return true;
+   }
+   else
+   {
+      return Get(
+         buffer + numBytesToRead, format, start + numSamplesToRead,
+         len - numSamplesToRead, mayThrow);
+   }
+}
```
Paul-Licameli, line 62: I have now read all of this replacement for `SampleTrackCache`.

That class had many more complications that I wrote long ago and reviewed again just today.  Did you read and understand all of it?  I don't ask to reproduce all of those, yet..

Like this cache, that one was meant to work best with a pattern of fetching from successive sample offsets.  Many small fetches then would not be punished with repeated database queries.

Unlike this, that one also allowed for efficient, sequentially backward fetches, as might also happen in scrubbing.

Also unlike this, that one had awareness of the underlying block boundaries within the Sequence and adjusted its caching pre-fetch sizes accordingly.  But that is lost with the simplifications here.

Removing the complications of best and ideal block sizes from WaveTrack and WaveClip interfaces still seems a good idea to me, cleaning up those abstractions, but now that the new cache can cooperate more intimately with Sequence, it should still implement similar logic to align its prefetches to the underlying Sequence blocks.

Also unlike this, SampleTrackCache maintained two buffers and could satisfy a request partly from each.  So suppose two buffers equal to the maximum block size.

So suppose only the maximum block size is exposed by WaveTrack as a preferred size for sequential fetches, and suppose an Effect does just that.  (Let's neglect time stretching).  Still the request might not align to the actual blocks, because of the history of cuts and pastes causing some under-sized blocks, or because some starting point was chosen without regard to alignment.

But maintaining two buffers would still reduce the number of database queries.  A request for the maximum block size might be filled partly by one slot, then partly by the other, but then also the remainder of the other slot is available to satisfy the next request without again invoking the getter function.

In other words -- without reproducing all complications of `SampleTrackCache`, still I would rewrite the above to loop, filling alternate slots, then copying to the out buffer, until satisfying the total request.

libraries/lib-wave-track/Sequence.cpp
```diff
@@ -1377,6 +1393,12 @@ bool Sequence::Append(
    return result;
 }
 
+void Sequence::InitPlaybackCache() {
+   mPlaybackCache = std::make_unique<SequenceSampleCache>(
+      std::bind(&Sequence::DoGet, this, _1, _2, _3, _4, _5), mNumSamples,
+      mMaxSamples);
+}
```
Paul-Licameli: As we discussed, a question is whether the cache is invalidated in sufficient places.

The cache depends on mNumSamples and mMaxSamples, so it should be invalidated at least when either of those changes.  (Though also when contents change).

I haven't studied the later commits, but even this one is incomplete by that criterion.

 `mMaxSamples` can change in `ConvertToSampleFormat`.  (Also `HandleXMLTag`, but that is deserialization of the structure from disk, so in fact just one initialization of the cache at the end of `HandleXMLEndTag`  is sufficient.)

`mNumSamples` may change where `AppendBlock` is called -- surprise!, the third argument is a reference.  Also in `Paste`, `Delete`.  Also `CommitChangesIfConsistent`, `AppendBlocksIfConsistent` but you covered those.

libraries/lib-wave-track/SequenceSampleCache.h
Paul-Licameli: ... or don't make a separate file?

You made this cache separate from Sequence and used a std::function callback -- were you thinking of avoiding dependency cycle?  Yes, it avoids the cycle between two translation units.

But now I think it could be quite proper for Sequence and its cache to cooperate intimately, each class using the other's interface directly, and one a friend of the other -- and so, put them all just in Sequence.*

Avoidance of dependency cycles is desirable in the graph if `#include`s relating translation units.  But it's not necessary at the level of closely cooperating classes, like containers and their iterators.

libraries/lib-wave-track/Sequence.cpp
```diff
@@ -82,11 +82,6 @@ size_t Sequence::GetMaxBlockSize() const
    return mMaxSamples;
 }
 
-size_t Sequence::GetIdealBlockSize() const
-{
-   return mMaxSamples;
-}
```
Paul-Licameli, line 85: Surely this functiion served little purpose

libraries/lib-wave-track/SequenceSampleCache.h
crsib: And the next thing is we spend time splitting classes into different headers/translation units :-)

libraries/lib-wave-track/Sequence.cpp
```diff
@@ -778,41 +773,6 @@ void Sequence::AppendBlock( SampleBlockFactory *pFactory, sampleFormat format,
...
 
-sampleCount Sequence::GetBlockStart(sampleCount position) const
-{
-   int b = FindBlock(position);
-   return mBlock[b].start;
-}
-
-size_t Sequence::GetBestBlockSize(sampleCount start) const
-{
-   // This method returns a nice number of samples you should try to grab in
```
Paul-Licameli, line 786: Finally I reach the commit that deletes the two functions above, and much else.

As I already explained, I don't like these deletions from Sequence.  I think some valuable old logic for alignment of cache fetches is lost.  I think these two functions at least should be retained, and they should be private, but the cache of the sequence should be a friend or a nested class.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -296,12 +296,9 @@ class WAVE_TRACK_API WaveTrack final : public WritableSampleTrack
    // and alignment for efficiency
    //
 
-   sampleCount GetBlockStart(sampleCount t) const override;
-
    // These return a nonnegative number of samples meant to size a memory buffer
-   size_t GetBestBlockSize(sampleCount t) const override;
    size_t GetMaxBlockSize() const override;
-   size_t GetIdealBlockSize();
+   size_t GetMaxBlockSize();
```
Paul-Licameli, line 296: Why are there now const and non-const overrides?  This suggests the query might mutate the track.

I think `GetIdealBlockSize` was erroneously non-const, and you just made a global search and replace?

libraries/lib-wave-track/WaveClip.h
Paul-Licameli: `ToSequenceSamples` can be removed, as now unused -- and then `GetSequenceStartSample` is also unused, so delete it too

src/SpectrumTransformer.cpp
```diff
@@ -357,16 +357,11 @@ bool TrackSpectrumTransformer::Process( const WindowProcessor &processor,
    bool bLoopSuccess = true;
    auto samplePos = start;
    while (bLoopSuccess && samplePos < start + len) {
-      //Get a blockSize of samples (smaller than the size of the buffer)
-      const auto blockSize = limitSampleBufferSize(
-         std::min(bufferSize, track->GetBestBlockSize(samplePos)),
-         start + len - samplePos);
```
Paul-Licameli, line 363: This change is incorrect!

Removing the use of `GetBestBlockSize` is all right, but not removing the call to `limitSampleBufferSize`, which makes sure that the last fetch does not exceed the intended range of the sequence.

Paul-Licameli (changes requested):

libraries/lib-wave-track/SequenceSampleCache.h
Paul-Licameli: > And the next thing is we spend time splitting classes into different headers/translation units :-)

@crsib We split lots of old legacy code that combined UI and internals carelessly and was generally written without regard to proper dependency cycle management.  This case isn't like that.

But are you thinking of some generalized abstract, polymorphic Sequence?  I think you hinted at that and I don't undertand you yet.

src/VoiceKey.cpp
```diff
@@ -864,19 +864,16 @@ double VoiceKey::TestEnergy (
 
    while(len > 0)
       {
-         //Figure out how much to grab
-         auto block = limitSampleBufferSize ( t.GetBestBlockSize(s), len );
```
Paul-Licameli, line 868: Also incorrect to remove `limitSampleBufferSize` here, though this is experimental code unreachable in the release build.

Paul-Licameli (changes requested):

libraries/lib-wave-track/CachingSequence.h
```diff
@@ -0,0 +1,101 @@
+#pragma once
+
+#include "Sequence.h"
+#include "SequenceInterface.h"
+#include "SequenceSampleCache.h"
+
+class WAVE_TRACK_API CachingSequence final : public SequenceInterface
```
Paul-Licameli, line 7: I don't like this solution to the invalidation problem.  Distinguishing abstract SequenceInterface, and making this decorator, seems to me, excessive complication for the problem.

I would rather, as I said, let the Sequence and its cache be closely cooperating classes in one translation unit, and just invalidate the cache in all mutating member functions of Sequence.  There would be little need to disturb other code outside of Sequence.

This commit does add a lazy creation-on-demand of the cache when getting from the sequence -- that part is good.

Putting sufficient invalidations into Sequence.cpp so that we can examine it in isolation and be confident of cache correctness -- that is good.

Still there might be the other problem of "eviction."  Reclaiming the memory resources of the cache when it's not good to occupy them.

The old caches were short-lived -- they disappeared after rendering a spectrogram, or after destrying a Mixer object.

The new caches are internal cursors of Sequences, and might be more numerous than before, as they become per-clip and not per-track, and are not local variables in appropriate scopes.

Sequence (and therefore WaveClip, and therefore WaveTrack) might therefore benefit by exposing an `InvalidateCaches` function that the application can decide to call in certain contexts.

True, it might seem like WaveTrack becomes a "leaky abstraction," but then it already also exposes functions to count and visit blocks for various other purposes, so adding this might not be too bad.

I would at least be sure to empty the caches of the duplicate tracks that are in the Undo stack -- both when creating Undo states, and also when duplicating tracks out of the Undo state.

Actually that might even require a pure virtual function in Track for cache invalidation.  See `TrackListRestorer` where the duplications happen.

Paul-Licameli:
And that almost completes a round of review...

... Just one thought I have now is that anticipatory pre-fetching of the next block by a background thread could be a further enhancement.  Just entertain that thought for now as a bonus project, that might speed up rendering for instance, but it may not be essential to get tolerable performance in realtime.

Paul-Licameli:
I've been thinking more about this.  I'm not convinced this will work.

In the old caching, the caches were external to the track, and spectrogram drawing or MixerSource controlled their lifetimes.  The invalidation problem did not exist, because in those contexts it could be assumed that the contents of the track would not be changing in the lifetime of the caches.

Furthermore, it was possible for different code to access the same wavetrack in different places, but different caches would exist to take advantage of the different patterns of locality.  For instance, a track might be playing or looping, while a changing display (such as, spectrogram with pinned play head, or a change of zoom during play) might require a different loop over the blocks concurrently.  And yes, concurrently -- there really were multiple threads involved.

But putting an internal cursor into the Sequence object has all the disadvantage of non-thread-safety; even if synchronized, different fetches from different threads would repeatedly miss the cache; and without a way to destroy the caches early, extra problem of invalidation exists.

So we need a very different solution that still allows the caches to be external to the Sequence, and yet, lets MixerSource control cache lifetimes although it is in lib-sample-track and can't depend on lib-wave-track where Sequence lives.

But how?  I can explain.

Paul-Licameli:
1. Sequence exposes a function that returns a cache object, and the caller determines its lifetime.
2. (If ever longer lifetimes are needed, such that invalidation becomes a problem, then Sequence can be a Publisher and the caches can subscribe.  But don't do that until it's really necessary, and then be sure there aren't multithreading problems.)
3. `Sequence::Get` (the overload not taking a first in argument) takes an optional pointer to cache defaulting to nullptr, and uses it if it isn't null.
4. This extra argument propagates up the stack in changes of arguments of WaveClip::GetSamples().
5. WaveTrack can construct multiple Sequence caches (one for each of its clips, storing them in a hash map from clip addresses) and call that single object a cache, and return it to the caller.  WaveTrack::Get(), if given a cache from outside, can pass the correct cache to the correct clip as needed.
6. SampleTrack can have a new pure virtual function, `MakeCache`, returning `X` , and its virtual function `Get` can take an optional pointer to `X`, defaulting to nullptr.  I just explained in 5 how these pure virtuals are overridden.
7. What type is `X`?  I think it's a good job for `std::any` -- that's just a polymorphic holder of anything that can be copied and destroyed.
8. Then construct the caches exactly where `SampleTrackCache` was constructed before, destroy them exactly where the old caches were destroyed, and whever the old caches were used, pass the new caches to `SampleTrack::Get()`.

Paul-Licameli:
> * Sequence exposes a function that returns a cache object, and the caller determines its lifetime.
> * (If ever longer lifetimes are needed, such that invalidation becomes a problem, then Sequence can be a Publisher and the caches can subscribe.  But don't do that until it's really necessary, and then be sure there aren't multithreading problems.)
> * `Sequence::Get` (the overload not taking a first int argument) takes an optional pointer to cache defaulting to nullptr, and uses it if it isn't null.
> * This extra argument propagates up the stack in changes of arguments of WaveClip::GetSamples().
> * WaveTrack can construct multiple Sequence caches (one for each of its clips, storing them in a hash map from clip addresses) and call that single object a cache, and return it to the caller.  WaveTrack::Get(), if given a pointer to a cache from outside, can pass the correct sub-cache to the correct clip as needed.
> * SampleTrack can have a new pure virtual function, `MakeCache`, returning `X` , and its virtual function `Get` can take an optional pointer to `X`, defaulting to nullptr.  I just explained in 5 how these pure virtuals are overridden.
> * What type is `X`?  I think it's a good job for `std::any` -- that's just a polymorphic holder of anything that can be copied and destroyed.  `WaveTrack::Get` uses `std::any_cast`.  (It must be the overload taking pointer, not reference to work in our macOS builds).
> * Then construct the caches exactly where `SampleTrackCache` was constructed before, destroy them exactly where the old caches were destroyed, and whever the old caches were used, pass the new caches to `SampleTrack::Get()`.

Paul-Licameli:
Also some afterthought about `WaveTrack::GetIdealBlockSize()`:  the side effect it might have had as a non-const function was unimportant, and is possible quite wrong in case the non-const function shadows the const one and becomes more widely used.  It could have a side effect of making a new, empty clip.

Don't do it.  Just delete that function.

And then, what purpose is still served by the old, and const, `GetMaxBlockSize` ?

It is, to hint a a good size for fetching some WaveTrack data for a sequential march through a long stretch of track duration, for instance in many of the effects.

Maybe it's just not important to have this function.  Those places should just allocate something like a megabyte.  If they shoudl later add caches for efficiency or pre-fetching (they aren't cached now, and would not be if you followed my suggestions above strictly), it wouldn't really bring any advantage to tune the effect's buffer size to match the internal buffers of the cache.

But there is no harm in keeping the function either, so let it remain and all the existing calls to it.

crsib:
> What type is X? I think it's a good job for std::any -- that's just a polymorphic holder of anything that can be copied and destroyed.

I will veto that and you know it well.

Paul-Licameli:
> > What type is X? I think it's a good job for std::any -- that's just a polymorphic holder of anything that can be copied and destroyed.
> 
> I will veto that and you know it well.

Explain your alternative.

Paul-Licameli:
std::any was used without objections here https://github.com/audacity/audacity/commit/3eb6f9861266ca8f75579c7862ae22b35bd5c719

crsib:
> the caches were external to the track, and spectrogram drawing or MixerSource controlled their lifetimes

But we are solving different problems here, aren't we?

> Sequence::Get (the overload not taking a first int argument) takes an optional pointer to cache, defaulting to nullptr, and uses it if it isn't null.

I see this as "excessive complication for the problem." Especially because it needs propagation `up the stack in changes of arguments of *`. I see it as a much more invasive approach.

> What type is X?

This is a huge red flag in your proposal. So it cannot be even expressed in type? The user needs to know what is safe to use to manipulate it or where to pass it? It cannot be verified by a compiler? A big, strong no. I had more trouble with `EffectSetting` than I asked already. I know that I was not alone. We've been talking through it for a very long time, and I have been largely ignored. So now I will be pushing hard against any such code, which is something I told you before too. JS gets type, Python gets type, and Audacity written in C++ erases types randomly because it's fun and because the committee decides to take `std::any` into the standard. (Still not the worst thing that slipped into the std lib, just look at `std::async`)   

If there are cases where you find per clip/per track caches to be required make an appropriate cache and READ using it, do not pass it to the track or clip. This is irrelevant to this problem anyway; we must efficiently satisfy around-the-border reads, which is only the problem with SQLite-based sample blocks. 

>  Reclaiming the memory resources of the cache when it's not good to occupy them.

Dropping the cache because the caller decided it no longer needs it is a very weird pattern. But again, it still can be done. You can have more than one cache. Your CPU has **three** just to satisfy reads.

> Explain your alternative.

Easy. Do not use `std::any`. I am really frustrated that I have to cover this again, but while I have never seen a good excuse to use `std::any` (that includes "what if you need a cache that can store anything by a key"), there are numerous problems it causes:

1. Compiler is no longer your friend. It is now the responsibility of the developer to validate types.
2. It violates LSP, and I have a very strong opinion that L is the single most important letter in SOLID. Without L it just makes no sense.
3. It makes debugging intrusive and hard.
4. It breaks all kinds of features that use the knowledge of the code model: code completion, refactoring, static analysis, etc.

> std::any was used without objections here [3eb6f98]

I was not happy, and I told you that during the review. However, that specific case was nothing more than a scope guard, the returned object was never meant to be used, so I allowed it. It could've been done differently and if I knew that it would be used as an argument for yet another `EffectSetting` I would've pushed to do it differently.

libraries/lib-wave-track/SequenceSampleCache.cpp
saintmatthieu: Missed that, thanks.

crsib:
On top of that - `std::any` is not polymorphic. It won't be possible to use different caching policies unless they are implemented within the same cache, so the proposed approach is closed for extension, forcing you to **modify** existing implementation if one needs to extend the behavior.

saintmatthieu:
@Paul-Licameli thanks for your constructive review.

I did think of multithreading, and yes, I wouldn't leave this code as-is given this possibility.

About the cache lifetime issue, I admit the scenario of a project with many small clips (which would happen if a user concatenates several clip instances of the same underlying data to make a looping track) would require far more memory than necessary.

I'm open to alternative approaches and am looking at yours with interest. Comments/questions coming up.

crsib:
> std::any is sufficient.

std::any is wrong. 

> The cache can be passed back to the object that produced it which then interprets it.

Why this complexity with passing random objects back and forth? There is a need for a cache with a specific lifetime? Create a cache, point it to the data source, and **read from the cache**. That's how caching works. Not by passing the cache back to the data source.

>  I foresee no further requirements.

If it was possible to foresee all the possibilities and requirements - software won't be needed. But this is not the case. "I do not see now" does not mean such a case won't occur tomorrow. Introducing rigid and overcomplicated code "just because" is not something that we need to continue to do.

> It addresses the present problems compatibly with library dependencies.

Please focus not only on breaking library dependencies. The code should be maintainable and readable first. `std::any` hides dependencies and result in bad code.

> I still don’t know what alternative is offered.

There is a cache already that suits your needs. It is orthogonal to what was needed for the abandoned graphics PR and to what stretching needs. So leave it this way. Do not try to invent a single cache to rule them all, this won't work.

In the end, you almost give the answer yourself. If you need a cache that returns you the samples for the given range - **explicitly** define the contract. Not with "comments", not with `std::any`. Explicitly.

Paul-Licameli:
I'm preparing a long comment, please wait.

Paul-Licameli:
Sorry for unintentional close.

Some study now convinces me that the really acceptable solution must first wait on some other restructuring I need to do (even without cache restructuring) in AudioIO.cpp.

saintmatthieu:
I realize that, in my draft above, each `WaveClip` holds on to one or more `SQLiteSampleBlock::BytePtr`, meaning many clips would mean lots of memory.

Trying again, this time moving the holders to `SampleTrack` implementation level. If the `SampleTrack` clients query small audio chunks at a time, which is typical and convenient, then there'd be typically one, sometimes two, strong references to `SQLiteSampleBlock::BytePtr` per track at all time, which would be totally acceptable memory wise. I think that'd address your concern, @Paul-Licameli, of
> putting a bound on the memory the cache may occupy at one time

```cpp
class SQLiteSampleBlock {
public:
  using BytePtr = shared_ptr<char[]>
  using ByteWPtr = weak_ptr<char[]>
  BytePtr Get();
}

class Sequence {
public:
  struct Cache {
    std::vector<BytePtr> sqliteCaches; // Contiguous audio samples
    const sampleCount start;
    const size_t len;
  }
  using CachePtr = shared_ptr<Cache>
  using CacheWPtr = weak_ptr<Cache>

  CachePtr Get(sampleCount start, size_t len) {
    CachePtr newCache;
    // !!!!!!!!!!!! change from previous draft !!!!!!!!!!!!
    // Put together newCache from sample blocks ...
    return newCache;
  }
}

class WaveClip {
public:
  Sequence::Cache Get(sampleCount start, size_t len) {
    return mSequence->Get(start, len);
  }
}

class WaveTrack {
public:
  Sequence::Cache Get(sampleCount start, size_t len) {
    // !!!!!!!!!!!! change from previous draft !!!!!!!!!!!!
    // Here do not release mCacheHolder until assembledCache is constructed with the BytePtrs it needs,
    // else that'd drop the ref count to 0 and SQLiteSampleBlock would need SQL-querying again.
    return mCacheHolder = assembledCache;
  }

private:
  Sequence::Cache mCacheHolder;
}

class StretchingSampleTrack {
public:
  Sequence::Cache Get(sampleCount start, size_t len) {
    // !!!!!!!!!!!! change from previous draft !!!!!!!!!!!!
    // Simiilarly to WaveTrack: do not let go of the strong reference before the one to be returned is created.
    return mCacheHolder = stretcherOutput;
  }

private:
  Sequence::Cache mCacheHolder;
}
```

It's not perfect that the clients of `Sequence` need knowledge of the caching mechanism to use it properly, there may be room for improvement there.

saintmatthieu:
> There is also still the problem of lifting the construction and lifetime management of these caches to an appropriate place, so that lib-sample-track and lib-audio-graph don't need any mention of the caches in their interfaces

Mention of the caches in their interface doesn't bother me so much. It would still be an improvement from what we currently have API wise because much easier to use, I expect.

saintmatthieu:
... we probably don't even have to mention "cache" in the returned audio containers. I'm more bothered (although not that much) by the clients having to make sure they do not drop a shared pointer before acquiring a new one. But maybe we can return a more intelligent object that takes care of this.

crsib:
I don't understand why we need a per-block sample cache.

Paul-Licameli:
> I don't understand why we need a per-block sample cache.

I doubted the idea too at first, but now I see advantages:  the cache aligns perfectly to block boundaries, and immutability means to need for invalidation logic.

Maybe the cache data are held by SeqBlock and not the SampleBlock itself.  If the mutable Sequence changes, a SeqBlock can be destroyed and free its cache.

Maybe Sequence can implement an LRU policy with a bound on total cache size (or block count) to evict some cache from memory and avoid too much memory usage.

Doing that MRU cache update with thread safety would present some difficulty but I think it could be overcome.

These ideas aren’t formed up yet.  We are exploring alternatives and need various ideas.

saintmatthieu:
@Paul-Licameli pointed out that this wasn't thread-safe. Looks like we need yet some rather different solution. We are evaluating ideas, but this one looks unlikely to be carried through. Declining.

PR #4724, opened 2023-05-23
Do not rely on `USE_VST` as VST is a library now

crsib:

Recommended:

### DIFF
--- a/libraries/lib-vst/VSTWrapper.h
+++ b/libraries/lib-vst/VSTWrapper.h
@@ -9,8 +9,6 @@
   Paul Licameli split from VSTEffect.h
 
 **********************************************************************/
-#if USE_VST
-
 #ifndef __AUDACITY_VST_WRAPPER__
 #define __AUDACITY_VST_WRAPPER__
 
@@ -329,5 +327,3 @@ struct VSTMessage : EffectInstance::Message
 };
 
 #endif
-
-#endif // USE_VST
--- a/libraries/lib-vst/VSTEffectBase.cpp
+++ b/libraries/lib-vst/VSTEffectBase.cpp
@@ -22,8 +22,6 @@
 #include "VSTEffectBase.h"
 #include "VSTInstance.h"
 
-#if USE_VST
-
 #include <wx/time.h>
 
 #include "Base64.h"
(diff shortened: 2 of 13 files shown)


libraries/lib-vst/VSTWrapper.h
```diff
@@ -329,5 +327,3 @@ struct VSTMessage : EffectInstance::Message
 };
 
 #endif
-
-#endif // USE_VST
```
Paul-Licameli, line 333: Is this complete?  `USE_VST` occurs in 15 files.

crsib: I have force pushed the updated version.

Paul-Licameli:
It's still incomplete.  USE_VST is still mentioned in src/AboutDialog.cpp and in src/CMakeLists.txt.

About the latter file -- my error, when I extracted lib-vst.  I made a mistake of not deleting lines no longer needed that refer to source files that no longer exist at paths under src/ .

crsib:
I disagree. I have left it in `src/AboutDialog.cpp` as there is no other way to appropriately fill the dialog. However, `AboutDialog` does not rely on `target_compile_definitions`, instead it relies on `config*.h`, which is forcibly included in all translation units by `audacity_append_common_compiler_options`.

Paul-Licameli:
> I disagree. I have left it in `src/AboutDialog.cpp` as there is no other way to appropriately fill the dialog. However, `AboutDialog` does not rely on `target_compile_definitions`, instead it relies on `config*.h`, which is forcibly included in all translation units by `audacity_append_common_compiler_options`.

OK, granted, but please clean up the nonexistent source file paths still in src/CMakeLists.txt, then I'll approve this.

Paul-Licameli:
> I disagree. I have left it in `src/AboutDialog.cpp` as there is no other way to appropriately fill the dialog. However, `AboutDialog` does not rely on `target_compile_definitions`, instead it relies on `config*.h`, which is forcibly included in all translation units by `audacity_append_common_compiler_options`.

OK -- and I was confused about src/effects/VST, I didn't move all of it to libraries

Paul-Licameli (approved):

Ryan-Miller-Audacity:
Tested.

Launch checks and quick VST2 sanity checks performed in MacOS, Windows, and Linux.

Issue #4720, opened 2023-05-22
Short and simple specification of how copy-paste "fits" the clipboard into the selected tracks


Paul-Licameli:
### Your idea

When some tracks have been cut or copied, and are then pasted (not pasting label or title text), and the set of selected tracks is not empty, succeed if and only if there is a "good fit" of ALL copied tracks into the selected tracks.

If successful, and sync lock is enabled, then in all cases, any track sync-locked with a changed track must be adjusted.

If not successful, fail with just one error message "Pasting one type of track into another is not allowed."

And this is a bit of pseudocode to define "Good fit" -- in short, it means, find a sub-sequence of selected tracks with types like those in the clipboard, or else fail.

A special rule will be applied:  when exactly one track is selected, consider `selected tracks` to include that track and all following ones. 

```
Correspondence := empty
ToCopy := clipboard
ToPaste := selected tracks
while ToCopy is not empty
   while ToPaste is not empty AND front(ToPaste) does not "fit into" front(ToCopy)
      pop_front ToPaste
   if ToPaste is empty
      fail
   push_back Correspondence, { front(ToCopy), front(ToPaste) }
   pop_front ToCopy
   pop_front ToPaste
```
And “fits into” means “of the same type" recognizing four distinct types (Mono, Stereo, Label, MIDI)

Or, it might instead be an asymmetrical relation, allowing mono to “fit into” stereo, replicating the sample data.

### Problem to be solved

Resolution of this question is needed to unblock refactoring for wide clips.  I want to refactor the present confused paste logic into something sensible before I go on.

### Prior art

As paste behaves now, the major decision tree is:

1. If there is a focused label or clip name, do pasting of text, not track data.
2. Otherwise if there is no selection, then paste all copied tracks into new tracks.
3. Otherwise... it's hard to describe all the cases simply.  The sequence of copied tracks may have types not corresponding to selected tracks, and some of those may be skipped. Sometimes paste fails with a message that "Pasting one type of track into another is not allowed." or that "Copying stereo audio into a mono track is not allowed."  Sometimes a copied track is multiply pasted while other copied tracks are not pasted at all!  (Examples will be added below)

See related issues #4714, #4718; also there may be sync-lock adjustment of unselected tracks.  

### Additional context

Documenting some strange cases in present behavior:

Case 1:
[...]

Issue #4718, opened 2023-05-22
Specify whether pasting mono into stereo is permitted


Paul-Licameli:
### Your idea

Pasting mono track data into a stereo track is now permitted, replicating it in both channels.

Do we wish to conserve this behavior?

See also #4714

### Problem to be solved

The legacy behavior of pasting from a mono wave track into another should be reexamined and respecified so that wide clip restructuring can proceed.

### Additional context

Present Audacity behavior:

1. Cut or copy from at least one wave track (possibly other tracks of other types too)
2. Paste into an equal number other selected tracks (maybe of anothe project), without correspondence of number of channels

In case a stereo track of the copy corresponds to a mono track in the attempted paste -- the operation fails with an error message, probably correctly so.

In case a mono track of the copy corresponds to a stereo track -- paste proceeds, replicating the mono data in each stereo track.

Is this legacy behavior to be preserved?  Or would it be preferable to disallow this too?

Ryan-Miller-Audacity:
Tested. This will wait for merge until the issue descrbied in the PR is discussed. Details can be found here: https://github.com/audacity/audacity/pull/4730#issuecomment-1579008868

Issue #4714, opened 2023-05-20
Redefine how Paste replicates the last copied wave track


Paul-Licameli:
### Your idea

Design should respecify the case of pasting a smaller number of wave tracks into a larger number of tracks.

### Problem to be solved

Audacity was written to replicate "the last" (maybe only) copied wave track when pasting into a greater number of selected tracks.

Certain existing edge case behavior of Paste may copy data from a stereo track right channel (only) into mono or stereo tracks, or paste a mono track into stereo tracks.

This behavior needs to be precisely re-specified for the present project to restructure wave clips as "wide."

Update: The first "yes" in this comment implies the answer to the question this issue raises https://github.com/audacity/audacity/issues/4720#issuecomment-1560194675

### Additional context

Observe present behavior of Audacity:

Case A:

1. Select (whole time range of) a stereo track with distinct left and right channel data
2. Copy (or cut)
3. Open new project
4. Add three empty mono tracks, then one stereo track; select all (Ctrl+A); paste
5. Observe error "Copying stereo audio into a mono track is not allowed" -- as it should be
6. Click drop-down of first track and "Make Stereo Track"; select all; paste
7. Observe pasting into the first stereo track, but also replication of the right channel into the mono track and each channel of the second stereo track (probably not wanted)

What would be correct?  I suggest:  don't paste anything into the mono, but replicate stereo into stereo

Case B:
1. Select (whole range of) a mono track
2.-4. as in case A
5. Observe the mono channel replicated in all five mono and stereo channels
6. Acceptable?  Or, should mono paste only into mono?

Paul-Licameli:
Assigning myself the implementation work, a necessary edge case fix in all the restructuring now happening.  Who wants to be assigned the design specification? @Tantacrul @LWinterberg @DilsonsPickles

Tantacrul:
I think @LWinterberg and @DilsonsPickles will tackle the question and we'll discuss it in our next meeting and then make a proposal to the team. Happy to hear suggestions too, obviously.

Ryan-Miller-Audacity:
Tested. This will wait for merge until the issue descrbied in the PR is discussed. Details can be found here: https://github.com/audacity/audacity/pull/4730#issuecomment-1579008868

PR #4730, opened 2023-05-24
Rewrite paste

Paul-Licameli:
Resolves: #4720
Resolves: #4714
Resolves: #4718

Rewrite the logic for pasting into tracks, fixing strange corner cases that sometimes pasted a copied track multiple times, sometimes never, and sometimes omitting needed sync-lock adjustments.

Recommended:

### DIFF
--- a/src/menus/EditMenus.cpp
+++ b/src/menus/EditMenus.cpp
@@ -42,7 +42,7 @@ void FinishCopy
       list.Add( dest );
 }
 
-// Handle text paste (into active label), if any. Return true if did paste.
+// Handle text paste. Return true if did paste.
 // (This was formerly the first part of overly-long OnPaste.)
 bool DoPasteText(AudacityProject &project)
 {
@@ -50,8 +50,8 @@ bool DoPasteText(AudacityProject &project)
    auto &selectedRegion = ViewInfo::Get( project ).selectedRegion;
    auto &window = ProjectWindow::Get( project );
 
-   for (auto pLabelTrack : tracks.Any<LabelTrack>())
-   {
+   // Paste into the active label (if any)
+   for (auto pLabelTrack : tracks.Any<LabelTrack>()) {
       // Does this track have an active label?
       if (LabelTrackView::Get( *pLabelTrack ).GetTextEditIndex(project) != -1) {
 
@@ -73,6 +73,18 @@ bool DoPasteText(AudacityProject &project)
          }
       }
    }
+
+   //Presumably, there might be not more than one track
+   //that expects text input
+   for (auto wt : tracks.Any<WaveTrack>()) {
+      auto& view = WaveTrackView::Get(*wt);
+      if (view.PasteText(project)) {
+         auto &trackPanel = TrackPanel::Get(project);
+         trackPanel.Refresh(false);
+         return true;
+      }
+   }
+
    return false;
 }
 
@@ -256,10 +268,6 @@ bool HasHiddenData(const TrackList& trackList)
    }
    return false;
 }
-
-}
-
-namespace {
 
 // Menu handler functions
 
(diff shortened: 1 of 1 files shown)


Paul-Licameli:
Rebased onto master, resolved conflicts with my other recent PR, and force-pushed.

One fewer commit because TrackList::NChannels is now already defined.

LWinterberg:
In playing around with this build, it felt sensible overall, though the error messages ended up more annoying than I had anticipated. Can we make it so that in cases where the error pops up currently, it instead creates a new track? In the case of multiple selected tracks, any "overflowing" pasted contents would go immediately after the last selected track, in case of a single selected track, it would go at the end of the project.

crsib (approved):

Ryan-Miller-Audacity:
Tested PR build and confirmed the new paste details as outlined in each of the attached tickets. Did some brief testing of the paste function as a whole and confirmed no new issues identified. Tested on Windows, MacOS, and Linux builds.

During testing I did notice one crash which has been logged in #4769. This issue is not due to these changes and is reproducible back to Audacity 3.3.0.

There is one use case that I'd like to get clarification on after talking with Paul. 

1) When I have a mono track and a stereo track, copy my stereo track, and attempt to paste by clicking within my mono track then pressing Ctrl+V, it doesn't paste within that mono track (to be expected); however, it does paste that data into my stereo track even though I do not have my stereo track selected at all. 

In the Issue ticket you mention "In case a stereo track of the copy corresponds to a mono track in the attempted paste -- the operation fails with an error message, probably correctly so", this message no longer displays which I believe would be the desired behavior in this case right? Certainly shouldn't have data being pasted where I am not current working in a seperate track.

You can see the behavior I'm describing pretty easily in this video with a couple of chirps.

[Untitled_ Jun 6, 2023 7_50 AM.webm](https://github.com/audacity/audacity/assets/88568755/f0bc3aa9-d2aa-4729-beee-b3ec869219ff)

After talking with Paul, it was noted that this behavior was discussed (Point 4 from #4720), but in use this just doesn't feel correct. Focusing and making a selection in a different track, but audio is being pasted into a completely uninvolved track from my actions seems like bad behavior in general.

@LWinterberg do you have any additional thoughts to add this? I mean this is largely an edge case since most people would not be attempting to paste stereo audio into a mono track in the first place; however, if discovered it would/could make a user think something is wrong. I would think that the selected track would return an error rather than placing the audio into the unrelated track (when a single track is selected). I would say this is fine behavior if the user uses a select all and then makes the attempt, but expect different results when isolating a single track.

LWinterberg:
> Focusing and making a selection in a different track, but audio is being pasted into a completely uninvolved track from my actions seems like bad behavior in general.

The ideal behavior would be to mix down the stereo track you have selected into the mono track and tell you about it, but I understand this is out of scope for the purpose of this refactor. And between the next-best options (giving you an error, or pasting into the nearest available slot), I prefer this version.

src/menus/EditMenus.cpp
```diff
@@ -616,209 +670,93 @@ void OnPaste(const CommandContext &context)
...
+   auto correspondence = FindCorrespondence(tracks, *srcTracks);
+   if (correspondence.empty()) {
+      if (tracks.SelectedLeaders().size() == 1)
+         AudacityMessageBox(XO(
+"There are not enough tracks after the selected track for the copied tracks.")
+      );
+      else
+         AudacityMessageBox(XO(
+            "There are not enough selected tracks for the copied tracks.")
+      );
```
LWinterberg, line 683: If we keep this message (and end up deferring #4734 until later), we probably should point out both the problem and the solution. So: 

`"The content you are trying to paste will span across more tracks than you currently have available. Add more tracks and try again."`

and

`"There are not enough tracks selected to accommodate your copied content. Select additional tracks and try again."`

Paul-Licameli:
> > Focusing and making a selection in a different track, but audio is being pasted into a completely uninvolved track from my actions seems like bad behavior in general.
> 
> The ideal behavior would be to mix down the stereo track you have selected into the mono track and tell you about it, but I understand this is out of scope for the purpose of this refactor. And between the next-best options (giving you an error, or pasting into the nearest available slot), I prefer this version.

In the general case, maybe a select a label track which is followed by a mono track, and I'm pasting one mono track.  Mix-down doesn't matter here.  In that example the question is whether to succeed quietly pasting into the following track or not.

Issue #4737, opened 2023-05-26
Stereo tracks should not allow different sample rates in the channels


Paul-Licameli:
### Bug description

Mono tracks with different rates can be joined into a stereo track, then split again, retaining their original rates.

This is incompatible with the restructuring needed for wide clips.  An attempt to join two such tracks should make an error message in 3.4.  In the unlikely event that a 3.3 project was saved with such a pathological stereo track, Audacity should refuse to open it.

If you really want to join two channels with mismatched rates, you should resample one or both to a common rate.

### Steps to reproduce

1. Generate some noise in a mono track with default settings; observe "44100 Hz" in the track control panel.
2. Generate another mono track of noise, and use Track > Resample to change it t9 48000 Hz; observe change in its panel.  (observe that it becomes "hairy" when you resample)
3. Select "Make stereo track" from the first track's menu and observe "44100 Hz" in the panel.  (But internally the right channel has not forgotten its original rate.)
4. Now "Split Stereo track" and observe that the second still has 48000 Hz.

But:  also observe that dragging clips into other tracks with different rates is not a problem.  For instance:
- Make a mono noise track at 44100 Hz
- Make new empty monon track; change to 48000 Hz
- Drag the clip from the first to the second
- Observe that it becomes "hairy" indicating that resampling really happened, as it should, to keep new clips in the track consistent in rate

Or you can do this:
- make a problematic track with empty left at 44100, empty right at 48000
- Make new stereo track of noise at 44100, 48000, or some other, higher rate
- Drag stereo clip into the first track
- Observe that respectively right only, left only, or both get "hairy" -- as expected, to keep consistency of clips in each channel

### Expected behavior

Step 3 should fail instead with an error message (what exactly?).

### Actual behavior

See discussion in Steps

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

petersampsonaudacity:
Fascinating catch @Paul-Licameli  - regression testing shows that this is _**NOT**_ a regression - this behavior has obtained all the way from Audacity 1.0.0 through to 3.3.2

I have no idea whether or not Dominic and Roger intended this to be the case or if it was just never blocked _(maybe it never occurred to them that folk might try to do this)_ ...

### Aside
[...]

PR #4739, opened 2023-05-26
Disallow mismatched sample rates

Paul-Licameli:
Resolves: #4737

Store sample rate nonredundantly for the stereo track pair.

Don't allow joining of mono tracks into stereo when rates differ. 
(Don't try to use one or the other rate, but let the user first resample.)

Disallow the opening of legacy files that might have been stored with such a mismatch.

Incidentally fix the failure to abort opening of files when inconsistency checks fail -- which was a bug introduced in 3.0.0

Recommended:


petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch build:  audacity-win-3.4.0-alpha-20230526+4a8eda0-x64

I confirm that the behavior reported in #4737 is now blocked:
image

Paul-Licameli:
> Testing on W10 with @Paul-Licameli 's branch build: audacity-win-3.4.0-alpha-20230526+4a8eda0-x64
> 
> I confirm that the behavior reported in #4737 is now blocked in this branch build: image
> 
> **Track Format** But note carefully that this fix does not block the creation of a stereo track from two single tracks each with a different **Sample Format**.
> 
> I tried with a 16-bit track and a 32-bit track - they could be merged into a single stereo track at 16-bit _(the 16-bit track was the upper track)._ And when split again they both reverted to their original bit formats 16-bit and 32-bit.

Thanks.  Yes, I should do other work to disallow mismatch of sample widths.

Did I do the right thing for rates?  An alternative to the prohibition might be a quiet resampling of the lesser rate to the greater.

Likewise for widths:  the narrower might quietly change to match the wider.  No dither would be needed, but storage requirements of the saved project might increase.

So should a join of channels with mismatched formats fail too with an error?

A third possibility (with either rates or channels) would be to ask the user.

petersampsonaudacity:
> Did I do the right thing for rates? An alternative to the prohibition might be a quiet resampling of the lesser rate to the greater.

Well I like it throwing the error - but I'm no expert in such matters.

_I have written to Dominic to as if he allowed this deliberately for some reason._


> Likewise for widths: the narrower might quietly change to match the wider. No dither would be needed, but storage requirements of the saved project might increase.

Folk already complain a lot about storage requirements increasing ...


> So should a join of channels with mismatched formats fail too with an error?

I would have thought so

 
> A third possibility (with either rates or channels) would be to ask the user.

That could also work - but would be complex if both rate and format were mismatched.


_**The preferred approach is really a decision for the Muse design team, deciding if they prefer the simple error(s) or a corrective dialog.**_  @LWinterberg @Tantacrul @DilsonsPickles

Tantacrul:
@LWinterberg @DilsonsPickles - can we investigate a little bit what we should do here to be as gracefully and friendly to the user as possible? I feel we should try to make a decision about this.

> Did I do the right thing for rates? An alternative to the prohibition might be a quiet resampling of the lesser rate to the greater.

A part of me thinks that we should just do this and not ask the user any questions.

crsib:
Resampling is generally lossy, I'm not sure that implicit resampling is a good idea

petersampsonaudacity:
> Resampling is generally lossy, I'm not sure that implicit resampling is a good idea

@crsib a VERY good point Dmitry

Paul-Licameli:
> @LWinterberg @DilsonsPickles - can we investigate a little bit what we should do here to be as gracefully and friendly to the user as possible? I feel we should try to make a decision about this.
> 
> > Did I do the right thing for rates? An alternative to the prohibition might be a quiet resampling of the lesser rate to the greater.
> 
> A part of me thinks that we should just do this and not ask the user any questions. However, there's going to be more to it than that methinks. A discussion for our design meeting, I feel!

Remember the user might attempt to make a stereo track interactively, or there might be files saved before 3.4 that the user tries to load into new Audacity.

Whatever the resolution (disallow, fix quietly, or ask), the work of detecting the condition is necessary and I put that effort in for the rates, but not yet for the formats.

Actually, format mismatch between channels, with no rate mismatch, might be a tolerable thing with all the data restructuring, not presenting significant trouble for the time stretching calculations we need to do, but it's all the same a weird thing to do.

petersampsonaudacity:
> ... or there might be files saved before 3.4 that the user tries to load into new Audacity.

A good point Paul.

So would you error trap that and offer to fix it by up-sampling to the higher of the two rates?

Tantacrul:
I'm coming to the conclusion that I cannot solve this with a single sentence :)

Thanks all. We'll need to give it some proper thought. Perhaps explicit prompts are the way to go.

Paul-Licameli:
> Resampling is generally lossy, I'm not sure that implicit resampling is a good idea

Is that really so?

If there is in fact no frequency content above 1/2 of the lesser sample rate (the lesser of two Nyquist frequencies), I think the Sampling Theorem tells us that theoretically you could down-sample from the from the higher rate to the lower, then up-sample again, without loss.

Practical calculations would truncate the tails of sinc functions though.

rbdannenberg:
There's no inherent loss, but calculation is expensive. An old Smith and Gossett paper did an error analysis and said you need something like 50 multiply/adds and 50 accurate estimates of the sinc function to get 16-bit resolution (so even more for 20 or 24 bits). You can estimate the sinc function with interpolated table lookup, but that's two memory accesses, a float to int conversion (usually this is expensive) and additional float operations (times 50 or however many points you use) so it's much worse than just 50 multiply/adds. I don't think there are any tricks to do fast interpolated lookups. On the other hand, I think macOS uses implicit sample rate conversion in core audio so that you can open  an output device at any sample rate even while the device is playing a stream at some other rate. I wouldn't be surprised if they cut corners on conversion, and I strongly believe software sampling synthesizers do not do such high-quality conversion since the worst case numerically (e.g. full amplitude at 20kHz) is not something you'll find in listenable audio, so you can normally get away with fewer points (i.e. a more truncated sync function). I don't know what's common practice, and people I've asked told me the details were trade secrets.

Paul-Licameli:
Thanks, Roger!

So, there is theory but then there is practice… and so quiet resampling is probably not a thing we should do and it isn’t what I did in this pull request.

crsib:
Not to mention that it would further increase the storage requirements!

saintmatthieu:
My guess on how most users would feel about the negative side effects of the various solutions :
* Quality loss: unaware, indifferent
* Memory increase: unaware, understanding
* Error pop-up: perplexed, annoyed, clueless
I therefore would favour the silent conversion (both sampling-rate and bit-depth wise). Whatever user is aware of the quality loss is probably tech savy and will find a way around it, or look a default behaviour she could change.

crsib:
> Memory increase: unaware, understanding

They are most certainly frustrated by the storage requirements, so definitely aware and not so understanding. Resampling a track will effectively keep the old track on your drive before you close Audacity (or even after, if you are unlucky), followed by a slow (and storage space demanding!) compaction. 

Trying to join two tracks with different sample rates sounds like a very rare corner case. I would much prefer to have a descriptive error message here, probably coupled with "Resample to xxx" button.

petersampsonaudacity:
> > Memory increase: unaware, understanding
> 
> They are most certainly frustrated by the storage requirements, so definitely aware and not so understanding.

Dmitry is right here :  we see many cases/complaints on the Forum (and here on GitHub) about the ballooning sizes of projects - so to add to this with hidden automatic resampling seem absolutely not the right thing to do.


>Trying to join two tracks with different sample rates sounds like a very rare corner case. I would much prefer to have a descriptive error message here, probably coupled with "Resample to xxx" button.

And in the light of Roger's comments _(and Dmitry's earlier comments in this thread)_ - I **_fully_** support Dmitry's approach here - which is basically the approach that Paul was originally pursuing with this PR "fiix".

crsib (approved):
I see no issues with the proposed solution. I think design issues could be solved in a follow up pull requests.

saintmatthieu:
> Resampling a track will effectively keep the old track on your drive

I'm aware, but if a user wants to do this, gets the error message and knows how to solve the problem, she will probably proceed with resampling manually, resulting in the same memory increase, no?

petersampsonaudacity:
> I see no issues with the proposed solution. I think design issues could be solved in a follow up pull requests.

A good decision.

But note that we also have to deal with mis-matched Formats - but @Paul-Licameli  is already aware of that.

And consideration needs to be given to Paul's earlier point about old projects already containing such stereo tracks with mis-matched rates and/or formats.

crsib:
> resulting in the same memory increase, no

But it won't be implicit. On top of that, the user may want to resample to a lower rate, not to a higher one. Or just to pan tracks accordingly (in case there are no plans to apply a stereo effect)

saintmatthieu:
I like the panning idea, and suggest that the error message one way or another indicates this workaround.

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -810,6 +810,15 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
    auto partner = static_cast< WaveTrack * >
       ( *tracks.Find( pTrack ).advance( 1 ) );
 
+   if (pTrack->GetRate() != partner->GetRate()) {
+      using namespace BasicUI;
+      ShowMessageBox(XO("Channels must have the same sample rate"),
```
saintmatthieu, line 815: This sentence will probably let the user think that she now has to convert one of the tracks to the sample rate of the other. As @crsib pointed out, an alternative worth considering would be panning - simpler, no memory usage increase, lossless.
- [] Please consider suggesting this in your sentence.

DavidBailes: There is also the danger that the user changes the sample rate of one of the tracks by using the rate sub menu on the track's menu, rather than Tracks > Resample.

petersampsonaudacity:
I've been discussing this with Dominic Mazzoni, the Godfather of Audacity.

His basic view
>Oh, that's definitely just a bug!   ...   I'm pretty sure I just never thought about that possibility.

petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch build the use case of "attempting to load a project (made previously) with mismatching rates in a stereo track".

I confirm that this is indeed blocked by Paul's branch build as @Ryan-Miller-Audacity reported in #4737 

But, the error message I get is:
image

This message is not helpful at all to the user:

**A) they would need to guess that they could manually fix this:**
1. open it in an earlier version of Audacity
2.  examine each stereo track in turn by splitting to mono, examining the TCP
3. resample requisite tracks to match sample rates
4. rejoin the mono tracks as stereo tracks
5. save the project
6. reopen project in latest Audacity

**B) Audacity offers no repair help**

Most folk are going to struggle to guess or even implement the A) manual fix.

So while we seem to have agreed we should just block such mismatched stereo track creation in 3.4.0 onwards with no automatic resampling - I think a _**good**_ case can be made for offering automatic resampling of channels in stereo tracks as part of the above error message.

Not many users are going to guess/intuit the step required for the A) manual fix - and I'm pretty sure we wouldn't want to put those manual steps in an error message.

Paul-Licameli:
Rebased onto recent master; one commit added to change error messages; otherwise unchanged.

crsib (approved):

Paul-Licameli:
Added one commit to fix the first of two problems described at https://github.com/audacity/audacity/issues/4737#issuecomment-1570799693

crsib (approved):

crsib:
Should I wait a bit for review this time?)

Paul-Licameli:
Two more commits.  One more review, please.

Addressing the other problem in Ryan's comment.

The work to make higher-level file consistency checks work properly, incidental to disallowing the sample rate mismatch -- it wasn't complete.  Thanks @Ryan-Miller-Audacity , and I don't agree with leaving the problem to be fixed later.

crsib:
> One more review, please.

Absolutely no problem with that :-) I was just joking a bit, sorry :-)

crsib (approved):

Paul-Licameli:
Sorry

DavidBailes (approved):
I agree with @crsib 's comment "I see no issues with the proposed solution. I think design issues could be solved in a follow up pull requests."
The comment that I made was to do with the design of the "error" message - I think that it could be more helpful, and point users towards Tracks > resample (rather than the rate sub menu on a track's context menu) , and mention @crsib 's suggesting of just using panning if stereo effects are not required.
So I'm quite happy to approve this pull request.

petersampsonaudacity:
>  I think that it could be more helpful, and point users towards Tracks > resample (rather than the rate sub menu on a track's context menu) ...

@DavidBailes  these are the revised error messages that @Tantacrul and Team discussed and agreed on - and are in @Paul-Licameli 's latest branch build

>So after discussing this with the team, we think this PR is good to go in terms of functionality.
>
>However, we need some copy improvements for the dialogs.
>
>Dialog 1: **Mono tracks must have the same sample rate in order to be combined into a stereo track**
>
>Dialog 2: **This project cannot be opened because it contains a stereo track with different sample rates. This functionality is no longer supported since the release of Audacity version 3.4**


Note that neither method of resampling is discussed here in these messages - it is up to the user tro decide whetherbtoi use te TCP dropdown or Resample Tracks from the menu, nor any other workaround like panning

And I think that is a good and right decision as otherwise the error messages become too long an overburdened and end up being TL;DR.

DavidBailes:
> @DavidBailes these are the revised error messages that @Tantacrul and Team discussed and agreed on
> and are in @Paul-Licameli 's latest branch build

OK.

petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch:  

**1. Making a mismatched stereo track**
This remains blocked by the revised error message:
image

**### 2. Spitting a stereo track**
Splitting to stereo and mono remain different with split to stereo producing panned L  & R tracks and spli to mono producing tw center-panned tracks

### 3. Project opening with a mismatched stereo track
A project made and saved in 3.3.2 with a stereo track with mis-matched sample rates cannot be opened in Paul's branch:
image

crsib (approved):

vsverchinsky (approved):

DavidBailes (approved):

PR #4743, opened 2023-05-29
Fixes the audio.com upload flow

crsib:
1. Track Title field is added
2. After the upload is completed URL is opened in the browser immediately

Recommended:

### DIFF
--- a/libraries/lib-cloud-audiocom/UploadService.cpp
+++ b/libraries/lib-cloud-audiocom/UploadService.cpp
@@ -206,6 +206,7 @@ struct AudiocomUploadOperation final :
 
    std::string mAudioID;
    std::string mUploadToken;
+   std::string mUserName;
 
    std::string mAudioSlug;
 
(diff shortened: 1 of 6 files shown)


dozzzzer:
Tested. A new issue was found: https://github.com/audacity/audacity/issues/4747

libraries/lib-cloud-audiocom/UploadService.cpp
```diff
@@ -579,7 +586,7 @@ UploadOperationHandle UploadService::Upload(
       mServiceConfig, fileName, projectName, isPublic,
       std::move(completedCallback), std::move(progressCallback));
 
-   mOAuthService.ValidateAuth([operation](std::string_view authToken)
+   mOAuthService.ValidateAuth([operation, this](std::string_view authToken)
```
vsverchinsky, line 589: `this` isn't used

vsverchinsky (approved):

PR #4754, opened 2023-05-31
Lift effect preview from libraries

Paul-Licameli:

A preliminary before lifting wave track caching scope out of lib-sample-track.

Move the previewing for Effects out of lib-effects, back up into Audacity.

This also makes lib-effects independent of lib-audio-io, which is nice.

(There may be a future "lib-effects-ui" for this layer, though it might not be toolkit neutral.)

- [x] QA:  Test playback, recording, effects preview from the destructive dialog.

Recommended:


crsib:
Please fix the build :-)

Paul-Licameli:
> Please fix the build :-)

Sorry, forgot to turn off precompiled headers locally.

It only needs one forward class declaration.

crsib (approved):

saintmatthieu:
So this PR removes the preview functionality from `lib-effects` because the preview functionality needs the creation of `TransportTracks`, which in turn need `TrackList`. So now `lib-effects` doesn't need `TrackList` for that purpose.
That's a couple of pieces of the puzzle, still missing the bigger picture. We can talk about it later. Don't wait for my approval.

dozzzzer:
Tested: playback, recording, effects preview from the destructive dialog

Paul-Licameli:
But it is also nice that lib-effects loses one dependency, on lib-audio-io.  I didn't like that dependency.  I really should have thought of all this extra work when lib-effects was extracted.  But that was itself already a large and difficult thing to do, requiring a lot of other tough untangling of some confused control flow involving UI with internals.

Issue #4344, opened 2023-02-21
Clip Copy dialog: the "Audio data" figure is misleading


petersampsonaudacity:
### Bug description

In the Clip Copy dialog introduced for 3.3.0 when copying a selection from one project to another a calculated figure is shown:
**Audio data is xx.x MB. Larger sizes will take longer to paste.**

This figure remains the same if the user changes from the default option of **copy to full clip** to just **copy the selection** (which is normally a _**lot**_ smaller.

### Steps to reproduce

1. Generate a 1 hour chirp
2. select a 5 minute chunk of audio
3. Ctrl+C to copy the "selection" to the clipboard
4. Ctrl+N to open a new project
5. Ctrl+V to paste into the new project
6. Observe: in the Clip Copy dialog the data size is reported as 605.6 MB with the default **Smart clip** setting
7. Select the **Selected audio only** radio button
8. Observe: the data size remains reported as 605.6 MB

### Expected behavior
Proper size for the data occupancy of only the selection to be reported when **copy the selection**  is active.

**Smart Clip**   -  605.5 MB is the expected size for a full 30 minute chirp
![image](https://user-images.githubusercontent.com/50205984/220390014-72faeb9b-281b-4c5c-adff-7abb0172fd37.png)

**Selection only**  -  50.5 MB is the expected size for a five minute selection
![image](https://user-images.githubusercontent.com/50205984/220392311-5a436845-2732-4bf0-8015-5848c68b6c88.png)

### Actual behavior
Misleading data size reported for the selection to be copied when **copy the selection**  is active.

**Smart Clip**   -  605.5 MB is the expected size for a full 30 minute chirp
![image](https://user-images.githubusercontent.com/50205984/220390036-d3d4c5a4-a1b5-4324-a9a3-c1966c38c75d.png)

**Selection only**  -  still showing as 605.5 MB and **_not_** the reduced 50.5 MB, the expected size for a five minute selection
![image](https://user-images.githubusercontent.com/50205984/220390527-5f7b940f-8e93-441c-88a9-9a994a814e0a.png)

### Audacity Version
3.3.0  audacity-win-3.3.0-alpha-20230220+55584ef-x64-msvc2022(2)

### Operating system
Windows 10 - but assume all OS

### Additional context
See:  Add prompt dialog when copying clips between projects #4004 
and:  Performance and file size issues when copying and pasting a small portion of a very large clip #3820

petersampsonaudacity:
[...]

PR #4757, opened 2023-06-01
clarify which option the MB size is referring to

LWinterberg:
Resolves: #4344


dozzzzer (approved):

petersampsonaudacity:
Testing on w10 with @LWinterberg 's branch:  audacity-win-3.4.0-alpha-20230601+fbe2798-x64

The message is now unambiguous, much better.

image

But as I noted earlier could be much improved, much more helpful to the user, if the size of the selection was also shown. 

This would help them in deciding which to paste.

Issue #4756, opened 2023-05-31
“Split Stereo” and “Split Stereo to Mono” now indistinguishable 


Paul-Licameli:
### Steps to reproduce

1. Generate a stereo track
2. Choose “Split Stereo Track” from track’s drop down 
3. Observe pan sliders
4. Undo
5. Choose “split stere to mono”
6. Observe pan sliders

### Expected behavior

Difference

### Actual behavior

Since only May 24 in alpha builds — no difference, both give you center panned results

(my fault for this regression)

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

So let’s decide:  the distinction mattered, and I should restore it, or it didn’t, and instead we just have one command

LWinterberg:
initial community feedback indicates that both are needed too, so go ahead

saintmatthieu:
I imagined that the implementation of splitting a stereo track would strive to preserve left-right balance, but in current live (and what Paul's PR currently intends to restore) it is not the case, left/right panning of a stereo track gets lost. For the use case you described, @LWinterberg, where one wants to make corrections to either channel separately, that'd be a nuisance.

If it were trivial, I'd suggest using this PR to change to this "lossless" behaviour, but there'd be the need to map pan values to gain values back and forth, which doesn't exist in the code yet AFAICS. Shouldn't be too hard to implement either, though.

PR #4759, opened 2023-06-01
Remove mono track types

Paul-Licameli:
Resolves: #4756

Remove another nuisance in the way of the restructuring for wide wave clips: a three-state distcintion among mono tracks.

Also makes lib-audio-graph not dependent on lib-track.

Besides fixing the recent regression issue 4756, it's not completely neutral behaviorally, as noted in commit comments.

The Track controls always say "Mono", not "Left" or "Right" when you make a hard pan in one direction.  Copy and paste of
hard-panned mono tracks no longer makes tracks with the legacy three-state set differently.

Phaser and WahWah might compute slightly different results for hard-right panned mono tracks, or
Loudness for hard left or right.

I don't care about those subtleties and I hope you don't either.

QA: besides the bug fix, exercise these things to be sure of no unintended changes of behavior
- Test export (especially "Export Multiple..."), mixing and rendering with some stacked effects in mono and stereo tracks
- Test all the above and also playback with changes of the pan slider
- Test the obscure Export Mixer dialog ("Use advanced mixing options" in Import/Export preferences)

Recommended:

### DIFF
--- a/libraries/lib-audio-graph/AudioGraphChannel.h
+++ b/libraries/lib-audio-graph/AudioGraphChannel.h
@@ -1,0 +1,49 @@
+/**********************************************************************
+ 
+ Audacity: A Digital Audio Editor
+ 
+ @file AudioGraphChannel.h
+
+ @brief Abstraction of a channel of a wide stream that knows whether it is mono,
+ left, or right
+ 
+ Paul Licameli
+ 
+ **********************************************************************/
+#ifndef __AUDACITY_AUDIO_GRAPH_CHANNEL__
+#define __AUDACITY_AUDIO_GRAPH_CHANNEL__
+
+namespace AudioGraph {
+//! Mutually exclusive channel classifications
+enum ChannelType : unsigned
+{
+   MonoChannel,
+   LeftChannel,
+   RightChannel,
+};
+
+struct AUDIO_GRAPH_API Channel {
+   virtual ~Channel();
+   //! Classify this channel
+   virtual ChannelType GetChannelType() const = 0;
+};
+
+//! Whether the channel is mono
+inline bool IsMono(const Channel &channel) {
+   return channel.GetChannelType() == MonoChannel;
+}
+
+//! Whether the channel may play through a left speaker
+inline bool PlaysLeft(const Channel &channel) {
+   const auto type = channel.GetChannelType();
+   return type == MonoChannel || type == LeftChannel;
+}
+//! Whether the channel may play through a right speaker
+inline bool PlaysRight(const Channel &channel) {
+   const auto type = channel.GetChannelType();
+   return type == MonoChannel || type == RightChannel;
+}
+
+}
+
+#endif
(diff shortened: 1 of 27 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -850,15 +791,16 @@ void WaveTrackMenuTable::SplitStereo(bool stereo)
...
    int totalHeight = 0;
    int nChannels = 0;
+
+   float pan = -1.0f;
    for (auto channel : channels) {
       auto &view = TrackView::Get( *channel );
-      if (stereo)
-         channel->SetPanFromChannelType();
+      if (stereo) {
+         channel->SetPan(pan);
```
saintmatthieu, line 801: This is a bug : the second iteration will pan **both** channels hard right. This is because channels of one track now share the same `GainAndPan` sample track attachment.

saintmatthieu (changes requested):

saintmatthieu (commented):
I guess this PR can be summarized as crossing two hurdles with one leap : fixing the regression, as well as factoring our the channel type from `Track` into `AudioGraph::Channel`. Is there something else?
(I'll ask you to remind me why it is a good thing that such methods are removed from `Track`.)

libraries/lib-audio-graph/AudioGraphChannel.h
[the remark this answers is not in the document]
Paul-Licameli: Let this stand.  The header comments explain what the functions do.  What is important is to reason that I preserve behavior in all places using the function instead of the enumeration.  That conservatism does require tests for IsMono moved first in some places.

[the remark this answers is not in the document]
Paul-Licameli: I found the old functions confusing because SetChannel stored a value retrieved by GetChannelIgnoringPan but not necessarily by the simply named GetChannel.

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
Paul-Licameli: Right, I should unlink channels first

Paul-Licameli:
Pan, gain, mute, and solo were notions that didn’t belong in lib-track.  They are not useful for all track subtypes.

Likewise the channel type.

lib-track is still messy for historical reasons.  It should be just about maintaining the list of tracks of heterogeneous types.

Another pull request will move the TypeSwitch metaprogramming into lib-utility.

saintmatthieu:
> Pan, gain, mute, and solo were notions that didn’t belong in lib-track. They are not useful for all track subtypes.
>
> Likewise the channel type. 

That's reminding me that not all tracks are sample tracks. Makes sense.

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
saintmatthieu: Works now.

saintmatthieu (approved):

Issue #2471, opened 2022-01-25
Audacity crashes on export from Macros to a yop level drive


petersampsonaudacity:
**Describe the bug**
The problem is that the exporting path for macros, which you can specify in menu Edit > Preferences > Directories
must be a folder and can't be directly the root of a partition, example F:\ it crashes!

With path: F:\macro it works and it creates automatically a subfolder named “macro-output”

Originally logged on the Forum by user Tyco72(2)
https://forum.audacityteam.org/viewtopic.php?f=46&t=122648&p=444407

**To Reproduce**
1. Edit > Preferences >Directories
2. Set the Macro output directory to be a top level drive ( say an exFAT formatted USB stick)
3. get some audio and select it
4. Tools > Macros
5. choose the MP3 Conversion Macro
6. Click on Apply Macro to Project
7. Observe: Audacity crashes on Export _(I sent the ET -phone-home report in the error dialog)_

**Expected behavior**
It should work with any valid path, in this case creating a a folder “macro-output” at the top level on the drive

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: Windows 10 - but assume all
 - Version Audacity 2.1.3 and 3.2.0 alpha

**Additional context**
I set the priority to P1 as:
a) Audacity should **_not_** crash,
b) the user does not receive ant helpful error message to inform them of the nature of the problem (see Forum thread)

In directories prefs - Open, Close, Import and Export directories worked as the top-level location without a folder and no crash

Even the temporary directory could be set to the exFAT USB stick and it works -_but I REALLY wouldn't recommend that IRL._

However it's not an entirely silly use case as my Windows PC has a 256GB SSD C: drive and on onboard spinning metal 1TB d:drive - so one might set the temporary directory to be on the D: drive - and note that is a directory "Session Data" on the D: drive and just a set of temporary files.

davidchocholaty:
Hi, I will try to take a look on it. @LWinterberg can you assign me please to this issue? Thank you in advance.

LWinterberg:
feel free to fix this! @davidchocholaty

davidchocholaty:
Hi, I'm so sorry for the later response. I tried to fix the issue for the GSoC project. However, I couldn't fix it and it wouldn't be possible to find time for it. I will unassign myself from this issue to let it for other devs that can fix it. Once so sorry for the later response.

petersampsonaudacity:
Still occurs on latest (today's) 3.2.0 release build:
[...]

PR #4767, opened 2023-06-05
Fixes crash when exporting to the root folder

vsverchinsky:
Resolves: #2471 

Recommended:

### DIFF
--- a/src/menus/FileMenus.cpp
+++ b/src/menus/FileMenus.cpp
@@ -75,7 +75,7 @@ void DoExport(AudacityProject &project, const FileExtension &format)
 
       // Append the "macro-output" directory to the path
       const wxString macroDir( "macro-output" );
-      if (fileName.GetDirs().back() != macroDir) {
+      if (fileName.GetDirs().empty() || fileName.GetDirs().back() != macroDir) {
          fileName.AppendDir(macroDir);
       }
 


crsib (approved):

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's branch build:   audacity-win-3.4.0-alpha-20230605+3df662a-x64

And using that branch it all works fine with both my on-board D:\drive a spinning metal 1TB disk and my external G:\ exFAT USB stick.

So Vitaly's branch looks to fix this issue (on Windows at least) I can try to test later on Mac - _it's a shame this didn't get tested in time to be pulled for the recent 3.3.3 release_

petersampsonaudacity:
Testing on macOS 12.6.3 Monterey with @vsverchinsky 's branch Universal Mac build

1) still fails with 3.3.3 0 BUT does bot crash on Mac - instead gives an error message:
image
which is nonsense as I can indeed write to it - the above image was transferred to my PC via that same USB stick


2.  On Mac too with Vitaly's branch Universal build all works fine 
I get  a macro-output folder created on Black 32 DT (the USB stick top level directory) - and the result of the MP£ Conversion macro is sent to that macro-output folder.


So Vitaly's fix for this works fine on Mac as well as Windows

dozzzzer:
The issue seems to be properly fixed now.

### Before the fix:

https://github.com/audacity/audacity/assets/70063970/bf95e0a4-d137-4ef8-9e5e-73dfac90a6d7

### After the fix:

https://github.com/audacity/audacity/assets/70063970/5da19a7d-617d-4408-87c3-ac212be5280f

PR #4771, opened 2023-06-06
SampleTrack caching refactoring: weak_ptr approach

### LINKED ISSUE #4717: Relocate caching from SampleTrackCache to some place below WaveClip

saintmatthieu:
Resolves: [4717](https://github.com/audacity/audacity/issues/4717)

Revisit caching strategy so as to allow `SampleTrack` clients to query small amount of samples at a time yet keeping the need for underlying SQLite queries low.

Further requirements:
- [x] cached data only held in memory the time it is needed (or not much longer)
- [x] is thread safe, and ideally allows concurrent reading from the **same** cache
- [x] any client of the same `WaveTrack` object can acquire data from anywhere in the track without affecting the risk of cache misses by other clients

Recommended:

Things to test:
- [x] spectrograms of stereo tracks, in which there is some obvious difference of content between the channels, to be sure the correct channel's stuff is drawn in each rectangle.
- [x] spectrum viewing, zoom in, zoom out, play around with it : it should feel at least as smooth as before.
- [x] find largest project that plays without hick-ups in previous Audacity version. Playback should also be hick-up-less with this version. (If it is, you may try to increase the number of tracks to see if this PR improves performance.)


libraries/lib-sample-track/SampleTrackStuff.h
```diff
@@ -0,0 +1,63 @@
...
+   // For now using float cos I don't care about other formats, but could be
+   // made generic.
+   Block(
+      size_t size, std::ofstream& ofs,
+      const std::chrono::time_point<std::chrono::steady_clock>& playStartTime);
+   ~Block();
+   std::vector<float> audio;
+
+private:
+   long long GetElapsedMs() const;
```
saintmatthieu, line 26: also for debugging

libraries/lib-project-file-io/SqliteSampleBlock.cpp
```diff
@@ -289,6 +298,35 @@ SampleBlockPtr SqliteSampleBlockFactory::DoCreateFromXML(
...
+   auto cache = mCache.lock();
+   if (cache)
+      return cache;
+
+   std::lock_guard<std::mutex> lock(mCacheMutex);
+   cache = mCache.lock();
+   if (cache)
+      return cache;
+
+   // For debugging - will be gone.
```
saintmatthieu, line 316: I wrote that debugging code to make sure that 
> cached data is released soon after last needed

Playback of a 20s mono track yielded for me
```
0 00000250D1D69EE0: Hello!
5315 00000250D1D68EE0: Hello!
5357 00000250D1D69EE0: Bye!
10765 00000250D1D68F60: Hello!
10808 00000250D1D68EE0: Bye!
16209 00000250D1D69A60: Hello!
16262 00000250D1D68F60: Bye!
20354 00000250D1D69A60: Bye!
```
Columns are `elapsed time (ms) / object address / ctor-dtor message` (See implementation of `LibSampleTrack::Block` for more detail).

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu, line 1: Sorry for the inadvertent formatting ...

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2064,6 +2064,37 @@ bool WaveTrack::Get(samplePtr buffer, sampleFormat format,
...
+   std::vector<sampleCount> blockOffsets(intersectingClips.size());
+   for (auto i = 0u; i < intersectingClips.size();++i)
+   {
+      const auto& clip = intersectingClips[i];
+      const auto clipT0 = std::max(t0 - clip->GetPlayStartTime(), 0.);
+      const sampleCount start { clipT0 * clip->GetRate() };
+      const sampleCount len { (t1 - clipT0) * clip->GetRate() };
+      auto newBlocks = clip->GrabFloats(start, len);
+      t0 += newBlocks.usefulLength.as_double() / GetRate();
+      blocks[i] = clip->GrabFloats(start, len);
```
saintmatthieu, line 2092: Oj, one `clip->GrabFloats(start, len);` too many ; useless but doesn't break anything.

libraries/lib-sample-track/MixerSource.cpp
saintmatthieu, line 1: I haven't tried this yet, but now we should be able to query not more than the amount of samples required by the playback buffer (2400 for me on Windows).

saintmatthieu:
If my reviewers think this is worthwhile pursuing, the next steps would include:
* Let all clients of `WaveTrack` adopt this new API,
* Remove `SampleTrackCache` and all related code,
* Consider unit-testing

libraries/lib-wave-track/SampleBlock.h
```diff
@@ -44,6 +45,8 @@ class MinMaxRMS
 class WAVE_TRACK_API SampleBlock
 {
 public:
+   virtual LibSampleTrack::BlockPtr Get() = 0;
```
Paul-Licameli, line 48: I would prefer a solution that does not require “polluting” the interface of SampleTrack with any mention of caching, but instead, lift the creation of caching scope up to places where WaveTrack is known, and we also construct the “decorator” tracks.

And that must wait on my other refactoring work, part of which I asked for review yesterday.

libraries/lib-project-file-io/SqliteSampleBlock.cpp
```diff
@@ -289,6 +298,35 @@ SampleBlockPtr SqliteSampleBlockFactory::DoCreateFromXML(
    return sb;
 }
 
+LibSampleTrack::BlockPtr SqliteSampleBlock::Get()
+{
+   assert(mSampleCount > 0);
+
+   // First try to acquire the cache without locking. If that doesn't succeed,
+   // lock and try again. Can't remember how this pattern is called.
```
Paul-Licameli, line 306: You are thinking of “double checked locking” but that is considered an anti pattern now because of our-of-order execution possibilities.

https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic

I think a proper solution must use these functions instead somewhere, but they are for shared pointers not weak.

Even the check of the weak pointer before the lock is not thread safe.  The standard library requires different shared and weak pointers to maintain the managed objects control block with thread safety, yes.  But races on the pointers themselves can cause undefined behavior.

So really you must put a lock scope around the weak pointer.  You might use the spin lock class that is already in out libraries.

You might also use two, more limited lock scopes by constructing the cache outside of the lock, and only then locking and assigning.

money thread might lose the race to assign non null, in which rare case, it can just release what it allocated.

libraries/lib-wave-track/SampleBlock.h
Paul-Licameli: I misplaced my comment— modifying this class in lib-wave-track is acceptable.  But the new types it uses should also be defined in this library and not in the lower one.

libraries/lib-sample-track/MixerSource.cpp
```diff
@@ -123,30 +124,23 @@ size_t MixerSource::MixVariableRates(
...
             if (backwards) {
-               auto results =
-                  cache.GetFloats(*pos - (getLen - 1), getLen, mMayThrow);
-               if (results)
-                  memcpy(&queue[*queueLen], results, sizeof(float) * getLen);
-               else
-                  memset(&queue[*queueLen], 0, sizeof(float) * getLen);
-
+               // Not sure this works, haven't tested.
+               timePos -= getLen / track->GetRate();
```
Paul-Licameli, line 128: No.

This will conflict seriously with other refactoring I want you to review which introduces the wide sample sequence abstraction into the mixer.

And I repeat, we should seek a solution where cache maintenance is entirely internal to lib-wave-track, and control of cache lifetime is upstairs in Audacity, and it is all completely non-intrusive in lib-sample-track.

libraries/lib-wave-track/Sequence.cpp
```diff
@@ -1114,6 +1114,23 @@ bool Sequence::Read(samplePtr buffer, sampleFormat format,
...
+   const auto sequenceOffset = start - GetBlockStart(start);
+   auto cursor = start;
+   while (cursor < start + length)
+   {
+      const auto b = FindBlock(cursor);
+      const SeqBlock& block = mBlock[b];
+      blocks.push_back(block.sb->Get());
+      cursor = block.start + block.sb->GetSampleCount();
+   }
+   return { std::move(blocks), sequenceOffset.as_size_t(), length };
```
Paul-Licameli, line 1131: This holds shared pointers to blocks, so there is no dangling pointer problem.   But this might construct a cache that becomes out of date with changing sequence contents.

Maybe it won’t matter in actual uses but comments should explain the limitation.

Or maybe a complete redesign of caching with have Sequence maintain more state and keep consistency.

libraries/lib-sample-track/MixerSource.cpp
saintmatthieu: > we should seek a solution where cache maintenance is entirely internal to lib-wave-track

Currently cache maintenance is in your to-be-introduced `lib-mixer`, through still persisting `SampleTrackCache`. So this proposal wouldn't be worse than what we have now (with respect to your concern).

Giving responsibility of cache maintenance to the consumers of `SampleTrack` (e.g. `MixerSource`, `SpectrogramCache`) makes the problems of eviction and cache miss avoidance easy. I cannot think of a solution that easy if this responsibility were to be that of `WaveTrack`.

libraries/lib-project-file-io/SqliteSampleBlock.cpp
crsib: I have looked into the spinlock implementation, and it calls `yield` every other time, so it's not better. I think we need extend it with a policy, that does not `yield` at all. Yielding make the spinlock at least as expensive as the `mutex`, it should be avoided in threads that have latency guarantees.

libraries/lib-sample-track/MixerSource.cpp
saintmatthieu: > This will conflict seriously with other refactoring I want you to review which introduces the wide sample sequence abstraction into the mixer.

I wouldn't mind so much rebasing my draft on top of your branch.

saintmatthieu: > we should seek a solution where cache maintenance is entirely internal to lib-wave-track

If that's of concern, we could define an abstract wrapper for `SampleTrack` (to become `WideSampleSequence`), whose implementation (in `lib-wave-track` or a new lib only `lib-wave-track` would depend on) is injected in consumers of ´WaveTrack`.

Paul-Licameli: > > we should seek a solution where cache maintenance is entirely internal to lib-wave-track
> 
> Currently cache maintenance is in your to-be-introduced `lib-mixer`, through still persisting `SampleTrackCache`. So this proposal wouldn't be worse than what we have now (with respect to your concern).
> 
> Giving responsibility of cache maintenance to the consumers of `SampleTrack` (e.g. `MixerSource`, `SpectrogramCache`) makes the problems of eviction and cache miss avoidance easy. I cannot think of a solution that easy if this responsibility were to be that of `WaveTrack`.

The not-worse organization should become a positively better one.

saintmatthieu:
@crsib @vsverchinsky @Paul-Licameli I rebased that branch onto Paul's `Mixer-does-not-use-Track` branch. Maybe there was a smarter way but, to keep the diff to my own changes, I opened a new PR from my branch onto Paul's. Here's the link : https://github.com/Paul-Licameli/audacity/pull/9 . @Paul-Licameli I guess you could add Dmitry and Vitaly as reviewers - I couldn't.

libraries/lib-sample-track/MixerSource.cpp
saintmatthieu: Maybe. If not I don't imagine this would get in the way of time stretching, meanwhile our current caching problem is.

libraries/lib-project-file-io/SqliteSampleBlock.cpp
saintmatthieu: > I think a proper solution must use these functions instead somewhere, but they are for shared pointers not weak.

You shared a link to a solution for the case where multiple threads access the _same shared_ptr object_. Here a new `shared_ptr` object is created each time - in which case "All member functions (including copy constructor and copy assignment) can be called by multiple threads on different instances of shared_ptr without additional synchronization". (https://en.cppreference.com/w/cpp/memory/shared_ptr)

libraries/lib-sample-track/SampleTrack.h
```diff
@@ -16,6 +16,7 @@ Paul Licameli split from WaveTrack.h
 #include "SampleCount.h"
 #include "SampleFormat.h"
 #include "WideSampleSequence.h"
+#include "SampleTrackStuff.h"
```
Paul-Licameli, line 19: Remove this unnecessary line; then, this revision of the pull request makes no changes in `lib-sample-track`, which is very good.

libraries/lib-mixer/SampleTrackStuff.h
```diff
@@ -0,0 +1,39 @@
...
+#include <vector>
+
+namespace LibSampleTrack
+{
+using BlockPtr = std::shared_ptr<std::vector<float>>;
+using BlockWPtr = std::weak_ptr<std::vector<float>>;
+
+struct MIXER_API ClipBlockSequence
+{
+   // adjacent sample blocks
```
Paul-Licameli, line 17: Just replacing `//` with `//!` on a brief comment line before a member variable allows
Doxygen to extract the comment.  Do that with each member of this structure and also explain `usefulLength`.

libraries/lib-mixer/SampleTrackStuff.h
```diff
@@ -0,0 +1,39 @@
...
+   std::vector<BlockPtr> blocks;
+
+   // FillBufferFromTrackBlockSequence needs these two. We keep them here for
+   // convenience.
+   size_t offsetIntoFirstBlock;
+   sampleCount usefulLength;
```
Paul-Licameli, line 23: I had to read Sequence.cpp to understand that `usefulLength` is a span of sample positions not necessarily covered by the blocks.  Better that I could have read a comment explaining that here.

libraries/lib-audio-io/AudioIO.h
```diff
@@ -78,7 +78,7 @@ struct AudioIOEvent {
 };
 
 struct AUDIO_IO_API TransportTracks final {
-   SampleTrackConstArray playbackTracks;
+   std::vector<std::shared_ptr<const WideSampleSequence>> playbackTracks;
```
Paul-Licameli, line 81: Slow down.

You are only setting yourself up for more conflict with stuff I have to do.

libraries/lib-mixer/MixerSource.h
```diff
@@ -112,6 +113,8 @@ class MixerSource final : public AudioGraph::Source {
 
    //! SampleTrackCache is the source of data
    std::unique_ptr<SampleTrackCache> mInputSequence;
+   std::vector<LibSampleTrack::TrackBlockSequence> mCacheHolders;
+   double mTimePos = 0.;
```
Paul-Licameli: This solution no longer intrudes in SampleTrack but still does intrude in lib-mixer.

I am not yet decided whether I like it.

libraries/lib-project-file-io/SqliteSampleBlock.cpp
Paul-Licameli: > What could be executed out of order here? According to https://en.cppreference.com/w/cpp/memory/weak_ptr/lock, the creation of the shared ptr is executed atomically, and all other variables before the mutex are local.

"Effectively returns expired() ? shared_ptr<T>() : shared_ptr<T>(*this), executed atomically."

Okay, I learned something -- `lock()` on a `std::weak_ptr` is thread safe.  My memory though is that you can't have unsynchronized usage of a `shared_ptr` object.

Paul-Licameli: There were more naive implementations of the double checked locking pattern in older days before C++11 that were later understood to be an anti-pattern.

Paul-Licameli: Other related and interesting C++ things to know are the thread safety guarantees in the language specification for the first time construction of static variables inside functions, and the related `std::once_flag` defined in `<mutex>`.

Paul-Licameli:
> @crsib @vsverchinsky @Paul-Licameli I rebased that branch onto Paul's `Mixer-does-not-use-Track` branch. Maybe there was a smarter way but, to keep the diff to my own changes, I opened a new PR from my branch onto Paul's. Here's the link : [Paul-Licameli#9](https://github.com/Paul-Licameli/audacity/pull/9) . @Paul-Licameli I guess you could add Dmitry and Vitaly as reviewers - I couldn't.

Don't make pull requests against my fork please.  That will only make more confusion.

Make all pull requests against audacity/audacity and we can have just one Conversation page like this one in which we find all related comments all have made.

I do not object to force pushes of this evolving project.

I still need to change the foundations for you to build on correctly.

Paul-Licameli:
General remarks about the new commit "CachingWideSampleSequence decorator".  I like it better now but still don't want to review it completely.  I still think I should continue preparing other foundational work for you to review, and you should rebase onto that, and only then should I do a complete review.

At first I was very displeased to see lots of change in AudioIO.* which will only conflict with the next refactoring I need to do, adding lots of changed lines of code, before I had completely understood the first commit.

But now I see the commit comment says that all the changes in MixerSource.cpp are reverted.  That may be good.  I had not yet understood all those changes but I thought it would be better to find a solution that does not change lib-mixer.

So really if I haven't understood the first commit fully, I can forget about it, but should instead look at the squash of the two commits.

There is now no need for `SampleTrackStuff` to be defined in lib-mixer.  It can instead be defined in Sequence.h.

That leaves lib-mixer almost unchanged, but not quite:  still you add `GetSampleTrack` to `WideSampleSequence`.

I don't like that -- lib-mixer mentioning a class defined in lib-sample-track, which is not one of its dependencies.  It compiles, because there is only a forward declaration of the class, but still it is questionable for lower level libraries to mention types defined in higher level libraries.

You make lib-audio-io accept the decorated `WideSampleSequence` objects constructed at higher levels without needing to know the implementation details.  That is goo.

But lib-audio-io still depends on `SampleTrack`, and therefore, this questionable workaround of `GetSampleTrack` is needed, to let it find that interface from `WideSampleSequence`.

Remember what I have said: really I want to make lib-audio-io have no dependency on lib-track at all.

Let me first complete that, so you can review it, then rebase this work onto that.  No changes to lib-audio-io should be needed after what I do.

saintmatthieu:
> Don't make pull requests against my fork please. That will only make more confusion.

I admit that it didn't prove helpful.

libraries/lib-wave-track/CachingWideSampleSequence.cpp
```diff
@@ -0,0 +1,128 @@
...
+   return mWaveTrack.WidestEffectiveFormat();
+}
+
+bool CachingWideSampleSequence::HasTrivialEnvelope() const
+{
+   return mWaveTrack.HasTrivialEnvelope();
+}
+
+void CachingWideSampleSequence::GetEnvelopeValues(
+   double* buffer, size_t bufferLen, double t0) const
```
Paul-Licameli, line 120: As we have discussed:  the decorator should also remap envelope values on the time axis, and also remap the end time, and not just pass through the calls.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2024,6 +2076,60 @@ bool WaveTrack::Get(samplePtr buffer, sampleFormat format,
    return result;
 }
 
+std::vector<LibSampleTrack::TrackBlockSequence> WaveTrack::GrabFloats(
+   size_t iChannel, size_t nBuffers, double& t0, sampleCount length) const
+{
+   const auto nChannels = NChannels();
+   assert(nBuffers - iChannel <= nChannels); // precondition
```
Paul-Licameli, line 2083: If you say it's a precondition, then int should be commented where the member function is declared in the header file, preferably with the `@pre` Doxygen comment convention

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: Yes, why not instead
```
blocks[i] = std::move(newBlocks);
```

libraries/lib-wave-track/CachingWideSampleSequence.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+bool CachingWideSampleSequence::Get(
+   size_t iChannel, size_t nBuffers, samplePtr buffers[], sampleFormat format,
+   sampleCount start, size_t len, fillFormat fill, bool mayThrow,
+   sampleCount* pNumWithinClips) const
+{
+   // For now leaving `GrabFloats` using t0 as query start time to minimize the
+   // diff.
+   auto t0 = mWaveTrack.LongSamplesToTime(start);
+   mCacheHolders = mWaveTrack.GrabFloats(iChannel, nBuffers, t0, len);
```
Paul-Licameli, line 87: This reassings `mCacheHolders`, recreating the complete cache stucture, each time `Get` is called.  Isn't this excessive?

Shouldn't you test whether the pointer is null first?

That work work assuming that sample sequence contents are unchanging during the lifetime of the decoartor.

libraries/lib-wave-track/Sequence.cpp
Paul-Licameli: I understand better how it works now.

Every time you call `Get` on the decorator, it goes back to the Sequence and finds the blocks that overlap the range.  The old cache is not yet destroyed when a new one is made.  Any blocks with a non-expired weak pointer to a vector of samples use that.

If the underlying Sequence has been changed between calls to `Get`, there is no invalidation problem.  The range of the request is just mapped to the new blocks.

saintmatthieu:
@crsib @vsverchinsky this PR is based on #4795. If you wish to review please only those commits with me face :D

saintmatthieu:
I compared some aspects of performance before and after this PR:
* Real-time performance : can play an Audacity project with as many as 1024 stereo, 146-second-long tracks without glitches, both before and after those changes.
* Memory usage : Audacity project with 4 stereo., 146-second-long tracks in spectrum display mode. Memory requirements decrease by 10% with this PR, from 231 to 212MB.

libraries/lib-wave-track/AudioSegmentSampleView.cpp
```diff
@@ -0,0 +1,34 @@
+#include "AudioSegmentSampleView.h"
+
+size_t
+AudioSegmentSampleView::Copy(float* buffer, size_t bufferSize) const
+{
+   size_t toWrite =
+      std::min(mLength.as_long_long(), static_cast<long long>(bufferSize));
```
Paul-Licameli, line 7: What is the proof that there is no loss from this narrowing conversion?

Please use `limitSampleBufferSize` which contains some assertions.

`sampleCount` is a wide integral type used for positions in a sample stream or durations, because you can overflow 32 bits at 44.1 kHz with only about 13.5 hours, unlikely but not impossible.

`size_t` is sufficient for buffer sizes that fit in memory.

libraries/lib-wave-track/AudioSegmentSampleView.cpp
```diff
@@ -0,0 +1,34 @@
...
+sampleCount AudioSegmentSampleView::GetSampleCount() const
+{
+   return mLength;
+}
+
+AudioSegmentSampleView::AudioSegmentSampleView(
+   std::vector<BlockSampleView> blockViews, size_t start, sampleCount length)
+    : mBlockViews(std::move(blockViews))
+    , mStart(start)
+    , mLength(length)
```
Paul-Licameli, line 32: I prefer to use C++11 universal initialization syntax  '{...}' where I can for member initializers.

libraries/lib-wave-track/AudioSegmentSampleView.cpp
Paul-Licameli: Only the lesser of `bufferSize` and `GetSampleCount` is written, instead of filling the buffer up to `bufferSize` with zeroes as needed.

That behavior is not obvious to guess from the un-commented header file.  So please comment.

libraries/lib-wave-track/AudioSegmentSampleView.h
```diff
@@ -0,0 +1,27 @@
...
+#include "SampleCount.h"
+
+#include <chrono>
+#include <fstream>
+#include <memory>
+#include <vector>
+
+using BlockSampleView = std::shared_ptr<std::vector<float>>;
+
+// An audio segment is either a whole clip or the silence between clips.
```
Paul-Licameli, line 12: Replace `//` with `//!` for Doxygen

libraries/lib-project-file-io/SqliteSampleBlock.cpp
Paul-Licameli: "Double checked locking".  Use that in the comment.

libraries/lib-project-file-io/SqliteSampleBlock.cpp
```diff
@@ -289,6 +298,31 @@ SampleBlockPtr SqliteSampleBlockFactory::DoCreateFromXML(
...
+   auto cache = mCache.lock();
+   if (cache)
+      return cache;
+
+   std::lock_guard<std::mutex> lock(mCacheMutex);
+   cache = mCache.lock();
```
Paul-Licameli, line 312: Please mention that `std::weak_ptr::lock()` guarantees atomicity, which is important to make this work without races.

libraries/lib-project-file-io/SqliteSampleBlock.cpp
```diff
@@ -289,6 +298,31 @@ SampleBlockPtr SqliteSampleBlockFactory::DoCreateFromXML(
...
+   if (cache)
+      return cache;
+
+   const auto newCache =
+      std::make_shared<std::vector<float>>(mSampleCount);
+   const auto cachedSize = DoGetSamples(
+      reinterpret_cast<samplePtr>(newCache->data()), floatSample, 0,
+      mSampleCount);
```
Paul-Licameli, line 320: Notice this `mSampleCount` correctly does NOT have type `sampleCount` but instead `size_t` because it is for an in-memory block.

libraries/lib-wave-track/Sequence.cpp
```diff
@@ -1114,6 +1114,24 @@ bool Sequence::Read(samplePtr buffer, sampleFormat format,
    return true;
 }
 
+AudioSegmentSampleView
+Sequence::GetFloatSampleView(sampleCount start, size_t length) const
+{
+   assert(start < mNumSamples);
+   length = std::min(sampleCount { length }, mNumSamples - start).as_size_t();
```
Paul-Licameli, line 1121: Again please use `limitSampleBufferSize`

Paul-Licameli, line 1121: Note `as_size_t` contains an assertion that you are not narrowing, so I like it better than the unchecked narrowing.

Paul-Licameli: `limitSampleBufferSize` might also express this

Paul-Licameli: That function also guards against the case that the `sampleCount` is negative.  That can happen because the type is also meant to expressed signed differences of positions in audio streams.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -136,6 +136,13 @@ WaveClip::~WaveClip()
 {
 }
 
+AudioSegmentSampleView
+WaveClip::GetSampleView(size_t ii, sampleCount start, size_t length) const
+{
+   assert(ii < GetWidth());
```
Paul-Licameli, line 142: I do not like new assertions with no accompanying comments to explain why they are justified.  Assertions should document assumptions you are making and explain them to the user.

I consider an assumption at the top of a function to be justified if the header file has a `@pre` Doxygen comment.

libraries/lib-wave-track/WaveClip.h
```diff
@@ -220,6 +221,14 @@ class WAVE_TRACK_API WaveClip final : public XMLTagHandler
    //! @returns Number of samples within t0 and t1 if t1 > t0, 0 otherwise
    sampleCount CountSamples(double t0, double t1) const;
 
+   /*
+    * @brief Request up to `length` samples. The actual number of samples
+    * available from the returned view is queried through
+    * `AudioSegmentSampleView::GetSampleCount()`.
+    */
```
Paul-Licameli, line 228: You need more commentary to explain `ii` and state the precondition on its legal values.

Then, it is understood that the burden of a guarantee of the correctness of the assumption lies not in the function but in the places that use it.

Paul-Licameli:
```
   /*!
    @param ii identifies the channel
    @pre `ii < GetWidth()`
    */
```

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2229,6 +2230,70 @@ bool WaveTrack::GetOne(samplePtr buffer,
...
+   const auto pOwner = GetOwner();
+   if (!pOwner) {
+      // GetSampleView called on a WaveTrack outside a project ??
+      // Not a complication I want to address, all the more that ongoing
+      // refactoring should rid us of this call.
+      assert(false);
+      return {};
```
Paul-Licameli, line 2226: But see `WaveTrack::Get`.  You can easily copy some of that to make this function more robust.

```
   const auto pOwner = GetOwner();
   if (!pOwner) {
      //! an un-owned track should have reported one channel only
      assert(nChannels == 1);
      nBuffers = std::min<size_t>(nBuffers, 1);
   }
```

Paul-Licameli: This is better than your not wholly justified `assert(false)`.  Instead it asserts a postcondition of `NBuffers()`.

Actually that's not an explicitly stated postcondition in the header, but that can and should be added easily.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2229,6 +2230,70 @@ bool WaveTrack::GetOne(samplePtr buffer,
...
+      const auto& clip = intersectingClips[i];
+      const auto clipStartTime = clip->GetPlayStartTime();
+      if (t0 < clipStartTime) {
+         // Insert silence
+         const auto numSamples = TimeToLongSamples(clipStartTime - t0);
+         std::vector<float> silence(numSamples.as_long_long());
+         std::vector<BlockSampleView> pointers = {
+            std::make_shared<std::vector<float>>(std::move(silence))
+         };
+         segments.push_back({ std::move(pointers), 0u, numSamples });
```
Paul-Licameli: It seems wasteful to allocate a lot of zeroes for this case.  With some more effort in `AudioSegmentSampleView` the silent case might be treated differently without the allocation.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2229,6 +2230,70 @@ bool WaveTrack::GetOne(samplePtr buffer,
...
+         t0 = clipStartTime;
+      }
+      const auto clipT0 = t0 - clipStartTime;
+      const auto clipS0 = TimeToLongSamples(clipT0);
+      const auto len = TimeToLongSamples(t1 - clipT0).as_size_t();
```
Paul-Licameli, line 2279: What's a proof that this is a save `as_size_t` ?  I think it can be deduced that `len` is not more than `length`.  An `assert` of that with some explanation of that deduction would be welcome.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -40,6 +41,8 @@ using WaveClipConstHolders = std::vector < std::shared_ptr< const WaveClip > >;
 using WaveClipPointers = std::vector < WaveClip* >;
 using WaveClipConstPointers = std::vector < const WaveClip* >;
 
+using TrackSampleView = std::vector<AudioSegmentSampleView>;
```
Paul-Licameli, line 44: Rename as `ChannelSampleView` ?

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2229,6 +2230,70 @@ bool WaveTrack::GetOne(samplePtr buffer,
...
+   std::vector<AudioSegmentSampleView> segments;
+   segments.reserve(intersectingClips.size());
+   for (auto i = 0u; i < intersectingClips.size();++i)
+   {
+      const auto& clip = intersectingClips[i];
+      const auto clipStartTime = clip->GetPlayStartTime();
+      if (t0 < clipStartTime) {
+         // Insert silence
+         const auto numSamples = TimeToLongSamples(clipStartTime - t0);
+         std::vector<float> silence(numSamples.as_long_long());
```
Paul-Licameli: That should be `as_size_t` and there should be some comments why that is not narrowing

libraries/lib-wave-track/Sequence.cpp
```diff
@@ -1114,6 +1114,24 @@ bool Sequence::Read(samplePtr buffer, sampleFormat format,
...
+   const auto sequenceOffset = start - GetBlockStart(start);
+   auto cursor = start;
+   while (cursor < start + length)
+   {
+      const auto b = FindBlock(cursor);
+      const SeqBlock& block = mBlock[b];
+      blockViews.push_back(block.sb->GetFloatSampleView());
+      cursor = block.start + block.sb->GetSampleCount();
+   }
+   return { std::move(blockViews), sequenceOffset.as_size_t(), length };
```
Paul-Licameli, line 1132: Comment why this doesn't narrow

Paul-Licameli (changes requested):

libraries/lib-wave-track/CachingPlayableSequence.h
```diff
@@ -0,0 +1,46 @@
...
+
+   // PlayableSequence
+   bool IsLeader() const override;
+   bool GetSolo() const override;
+   bool GetMute() const override;
+
+private:
+   const WaveTrack& mWaveTrack;
+   // Ok to have it mutable so long as it is used by one thread only.
+   // One per channel, one per track.
```
Paul-Licameli, line 41: I think the comment should simply say "One per channel"

libraries/lib-wave-track/CachingPlayableSequence.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "AudioIOSequences.h"
+#include "AudioSegmentSampleView.h"
+
+#include <memory>
+
+class WaveTrack;
+
+class WAVE_TRACK_API CachingPlayableSequence : public PlayableSequence
+{
```
Paul-Licameli, line 11: Some Doxygen commentary please about the purpose of the class.

libraries/lib-wave-track/CachingPlayableSequence.cpp
```diff
@@ -0,0 +1,115 @@
...
+         segment.GetSampleCount().as_long_long(),
+         static_cast<long long>(bufferSize - written));
+      written += segment.Copy(buffer, toWrite);
+   }
+   std::fill(buffer + written, buffer + bufferSize, 0.f);
+}
+} // namespace
+
+CachingPlayableSequence::CachingPlayableSequence(const WaveTrack& waveTrack)
+    : mWaveTrack(waveTrack)
```
Paul-Licameli, line 25: `{}` member initializer syntax?

libraries/lib-wave-track/CachingPlayableSequence.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+void FillBufferFromTrackBlockSequence(
+   const TrackSampleView& track, float* buffer, size_t bufferSize)
+{
+   size_t written = 0;
+   for (const auto& segment : track)
+   {
+      size_t toWrite = std::min(
+         segment.GetSampleCount().as_long_long(),
+         static_cast<long long>(bufferSize - written));
```
Paul-Licameli, line 17: Again, use `limitSampleBufferSize`

Paul-Licameli: You assume `bufferSize >= written` but what proves that?  Should there be a check?

libraries/lib-wave-track/CachingPlayableSequence.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+   return mWaveTrack.GetBestBlockSize(t);
+}
+
+bool CachingPlayableSequence::Get(
+   size_t iChannel, size_t nBuffers, samplePtr buffers[], sampleFormat format,
+   sampleCount start, size_t len, fillFormat fill, bool mayThrow,
+   sampleCount* pNumWithinClips) const
+{
+   mCacheHolders = mWaveTrack.GetSampleView(iChannel, nBuffers, start, len);
```
Paul-Licameli, line 44: You assume after this a post of `GetSampleView`:  `size() == nBuffers - iChannel`

Please make that explicit in header comments, and assert the post here

src/tracks/playabletrack/wavetrack/ui/SpectrumView.cpp
```diff
@@ -32,9 +32,10 @@ Paul Licameli split from WaveTrackView.cpp
 #include "ViewInfo.h"
 #include "WaveClip.h"
 #include "WaveTrack.h"
+#include "WideSampleSequence.h"
+#include "CachingPlayableSequence.h"
```
Paul-Licameli, line 36: You need only the second

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -374,6 +373,8 @@ void SpecCache::Populate
       const int lowerBoundX = jj == 0 ? 0 : copyEnd;
       const int upperBoundX = jj == 0 ? copyBegin : numPixels;
 
+// todo(mhodgkinson): I don't find an option to define _OPENMP anywhere. Is this
+// still of interest?
```
Paul-Licameli, line 377: I recall this was some non-team member's experiment that wasn't completed
See commit 8e1b084f01f97266eb843af72ec8d31045aa11fa

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.h
```diff
@@ -119,11 +118,11 @@ struct WaveClipSpectrumCache final : WaveClipListener
    void Invalidate() override; // NOFAIL-GUARANTEE
 
    /** Getting high-level data for screen display */
-   bool GetSpectrogram(const WaveClip &clip, SampleTrackCache &cache,
-                       const float *& spectrogram,
-                       const sampleCount *& where,
-                       size_t numPixels,
-                       double t0, double pixelsPerSecond);
+   bool GetSpectrogram(
+      const WaveClip& clip, const WideSampleSequence& sequence,
```
Paul-Licameli, line 122: Please comment that only the 0th channel of `sequence` is really used.

In the interim, this still works correctly for `WideSampleSequence` backed by a right channel track, which always ignores its partner.

Please comment in the PR that QA should test spectrograms of stereo tracks, in which there is some obvious difference of content between the channels, to be sure the correct channel's stuff is drawn in each rectangle.

libraries/lib-mixer/MixerSource.cpp
```diff
@@ -126,29 +124,22 @@ size_t MixerSource::MixVariableRates(
...
-               const auto result =
-                  (iChannel == 0) ? results.first : results.second;
-               const auto queue = mSampleQueue[iChannel].data();
-               const auto pFloat = floatBuffers[iChannel];
-               if (result)
-                  memcpy(&queue[queueLen], result, sizeof(float) * getLen);
-               else
-                  memset(&queue[queueLen], 0, sizeof(float) * getLen);
-            }
+            std::vector<float*> dst;
```
Paul-Licameli, line 127: Could use alloca for this small allocation

libraries/lib-wave-track/CachingPlayableSequence.h
saintmatthieu: Why on this particular class? Class comments are the exception in this repo, if there's a rule of thumb I'd like to know it (besides that's it's a cool class).

libraries/lib-wave-track/CachingWideSampleSequence.cpp
saintmatthieu: I don't understand this comment, I think you might be referring to the time-stretching decorator, which is another PR.

libraries/lib-wave-track/Sequence.cpp
saintmatthieu: `limitSampleBufferSize` is handy :thumbs-up:

libraries/lib-wave-track/WaveClip.h
saintmatthieu: Thanks.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: I'm addressing this reluctantly, I was hoping for no debate and time spending over this.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: You're right, that's not good.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: `as_size_t` already does the assertion, but I'm adding a comment.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2229,6 +2230,70 @@ bool WaveTrack::GetOne(samplePtr buffer,
...
+      intersectingClips.begin(), intersectingClips.end(),
+      [](const auto& a, const auto& b) {
+         return a->GetPlayStartTime() < b->GetPlayStartTime();
+      });
+   std::vector<AudioSegmentSampleView> segments;
+   segments.reserve(intersectingClips.size());
+   for (auto i = 0u; i < intersectingClips.size();++i)
+   {
+      const auto& clip = intersectingClips[i];
+      const auto clipStartTime = clip->GetPlayStartTime();
```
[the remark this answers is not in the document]
saintmatthieu: That'd be wrong.
Say we have one noisy clip starting at 0, the first second of which is trimmed. Now `start` is 0, and `length` is 2s.
In your proposal, the returned view would be 2s of noise. `WaveTrack` has to account for trimming ; the 1st second should be silence.

libraries/lib-wave-track/CachingPlayableSequence.h
Paul-Licameli: Comment it at least because the class is meant for use outside the library that defines it.  I don’t think you must comment every type defined in anonymous namespaces.

If many classes don’t have DOxygen, that’s not an excuse for improving practice now.

libraries/lib-wave-track/CachingPlayableSequence.h
saintmatthieu: Thank you.

libraries/lib-wave-track/CachingPlayableSequence.cpp
saintmatthieu: Right, I never took the habit for some reason.

libraries/lib-wave-track/CachingPlayableSequence.cpp
saintmatthieu: This is based on the assumption that `AudioSegmentSampleView::Copy` does what it says, namely not returning more than `toWrite` samples. A unit-test proves this, which I'd argue is enough ?

libraries/lib-wave-track/CachingPlayableSequence.cpp
```diff
@@ -15,7 +15,7 @@ void FillBufferFromTrackBlockSequence(
       size_t toWrite = std::min(
          segment.GetSampleCount().as_long_long(),
          static_cast<long long>(bufferSize - written));
-      written += segment.Copy(buffer, toWrite);
+      written += segment.Copy(buffer + written, toWrite);
```
Paul-Licameli, line 18: I understand the bug fix commit and expect you will squash it after complete review.

Paul-Licameli:
I have now completed a round of review up to commit 9703ec78ab4799dfc6031963590ee5266c54957b which was the branch tip yesterday.  Your edits for the next round have overlapped the tail end of my review.

libraries/lib-wave-track/CachingPlayableSequence.cpp
saintmatthieu: `nBuffers - iChannel`, really?
I have been interpreting those arguments as "give me `nBuffers` starting from index `iChannel`", which results in `size() == nBuffers`. What you are saying is "give me buffers from index `iChannel` to index `nBuffers` (exclusive)".
(Your implementation of `WaveTrack::Get` does neither, but I think it attempted the former interpretation. I'm attempting a fix accordingly.)

Paul-Licameli: Sorry, what I wrote was mistaken.

What I was thinking was, how to prove you don't index `buffers` out of bounds.

So what I should have said is, there is an unstated postcondition that `GetSampleView` always returns a vector of size `nBuffers`.

So it would be good to make that explicit in that function's header.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
saintmatthieu: Mmh, but this commit is just making use of it in some other file. In any case, if you're correct now one should care if I removed those ifs. Will append a tentative commit that does that.

libraries/lib-wave-track/CachingPlayableSequence.cpp
Paul-Licameli: Tests are just one line of defense ensuring code quality and raising our confidence.  I think they should be complemented with assertions in the program.  (I mean both `assert` and also the informal statement of pres and posts and invariants.)

When I write assertions it means I'm very serious about persuading the unfamiliar reader (who might be you, or future me) that the code is CORRECT, because I was working hard at the time to convince myself first and to make more explicit the assumptions that I relied on.

A test, though automated, is the weaker form of "well it seems to work".

True, human reasoning is still fallible and tests may still catch mistakes (perhaps because of future modifications by someone understanding the code imprefectly).

Still -- write contracts first, I think, and a well thought contract is then a guide for writing the test cases themselves.

Some correspondence between test cases and formal or informal assertions in the program is a good idea.

libraries/lib-wave-track/CachingPlayableSequence.cpp
saintmatthieu: Can do that.

libraries/lib-wave-track/CachingPlayableSequence.cpp
Paul-Licameli: You should also remember, though, that sometimes you must use `()`.  One example is initializing a vector with a size.  If the element type is implicitly constructible from an int or size_t, the `{}` syntax will invoke the wrong constructor.

This danger exists when a class has a constructor taking `std::initializer_list`.

saintmatthieu:
> I have now completed a round of review.

Thanks a lot!

libraries/lib-wave-track/WaveTrack.cpp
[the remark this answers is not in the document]
Paul-Licameli: Did you try this case?  I didn't, but I thought I deduced a problem.

`WaveClip::GetPlayStartTime()` returns `mSequenceOffset + SamplesToTime(TimeToSamples(mTrimLeft));` which is assigned to `clipStartTime`

Then we compute `clipS0` as so many samples into the PLAYABLE part of the clip (the unhidden part)

Then we pass that into `WaveClip::GetSampleView` as `start` and it is passed unchanged into `Sequence::GetFloatSampleView`

BUT that function is interpreting `start` as relative to the start of the sequence (which doesn't know about the trimming) and that includes the left trimmed part.  So doesn't this mean that the cache contains the wrong stuff, left of the correct place by a shift equal to the trim length?

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2126,17 +2109,19 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers, samplePtr buffers[],
    bool mayThrow, sampleCount * pNumWithinClips) const
 {
    const auto nChannels = NChannels();
-   assert(nBuffers - iChannel <= nChannels); // precondition
+   assert(iChannel + nBuffers <= nChannels); // precondition
```
saintmatthieu, line 2112: This fix attempt relates to https://github.com/audacity/audacity/pull/4771#discussion_r1239606798

libraries/lib-wave-track/CachingPlayableSequence.cpp
crsib: Numerous times, I have seen well-thought assertions fail on the first run.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: You saw my retraction of my mistaken comment of yesterday?

Paul-Licameli: I said I finished review, but then I wrote some more in answer to your answers.

libraries/lib-wave-track/CachingPlayableSequence.h
saintmatthieu: > If many classes don’t have DOxygen, that’s not an excuse for improving practice now.

That wasn't my intention, but I don't want to feel I have to comment every single class where not necessary.

The visibility-beyond-hosting-library sounds like a good rule of thumb to me.

libraries/lib-wave-track/CachingPlayableSequence.cpp
Paul-Licameli: Yes, we are fallible, but still we should value well thought assertions and make attempts at rigor.

As I said, I think contracts as a guide to what cases should be tested.  Or if you like the TDD approach of writing failing tests first, and then then changing the program to succeed, I think then you should say something in the contract too.

Tests might be "executable documentation" but I like documentation in the header itself, explaining all that is important in the interface in one place.

saintmatthieu: I actually found an alternative implementation, where it is obvious that `written < bufferSize`, and which does not unnecessarily iterate over segments after if `written` already `== bufferSize`.

But I'd have complied.

libraries/lib-wave-track/CachingPlayableSequence.cpp
saintmatthieu: Ok.
I already added a `@return` line to that header stating this. Is that good enough?

libraries/lib-wave-track/WaveTrack.cpp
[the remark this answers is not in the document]
saintmatthieu: Good catch!
But I'm fixing it in `WaveClip::GetSampleView`, accounting for `mTrimLeft` like `GetSamples` does.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: > You saw my retraction of my mistaken comment of yesterday?

I did, yes, but I still think this assertion is incorrect.

libraries/lib-wave-track/AudioSegmentSampleView.cpp
```diff
@@ -1,10 +1,47 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
```
Paul-Licameli, line 2: I didn't mention the file banners -- good that you remembered

libraries/lib-wave-track/AudioSegmentSampleView.h
```diff
@@ -9,19 +19,47 @@
...
 using BlockSampleView = std::shared_ptr<std::vector<float>>;
 
-// An audio segment is either a whole clip or the silence between clips.
-class WAVE_TRACK_API AudioSegmentSampleView
+//! An audio segment is either a whole clip or the silence between clips.
+class WAVE_TRACK_API AudioSegmentSampleView final
 {
 public:
+   /*
+    * @brief Constructs an `AudioSegmentSampleView` from `BlockSampleView`s.
```
Paul-Licameli, line 27: The grave accents are markdown that make text appear in typewriter font.  I often use it in `@pre` and `@post` when a pre/post-condition can be written formally in C++.

It's less important here -- Doxygen will recognize the class name and format it as an html link.  The grave accents cause the link to have typewriter font too.

"BlockSampleViews" however will fail to link because it's not the exact type alias name.

libraries/lib-wave-track/AudioSegmentSampleView.cpp
Paul-Licameli: You might use the `@file` and `@brief` Doxygen commands in this banner, so the summary html page of files has a comment about the purpose.

Here's the reference for all commands. https://www.doxygen.nl/manual/commands.html

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.h
```diff
@@ -119,11 +118,15 @@ struct WaveClipSpectrumCache final : WaveClipListener
    void Invalidate() override; // NOFAIL-GUARANTEE
 
    /** Getting high-level data for screen display */
-   bool GetSpectrogram(const WaveClip &clip, SampleTrackCache &cache,
-                       const float *& spectrogram,
-                       const sampleCount *& where,
-                       size_t numPixels,
-                       double t0, double pixelsPerSecond);
+   // From @Paul-Licameli:
```
Paul-Licameli, line 121: Just use my initials `PRL:` as in many other comments

Paul-Licameli:
> I actually found an alternative implementation, where it is obvious that `written < bufferSize`, and which does not unnecessarily iterate over segments after if `written` already `== bufferSize`.
> 
> But I'd have complied.

The proof relies on `AudioSegmentSampleView::Copy` returning not more than its `bufferSize` argument.
The comment on that function is sufficient, if you look for it.
It's enough.

Paul-Licameli:
> Comment it at least because the class is meant for use outside the library that defines it. I don’t think you must comment every type defined in anonymous namespaces.
> 
> If many classes don’t have DOxygen, that’s not an excuse for improving practice now.

Of course I meant to say... it's not and excuse for NOT improving practice!

Paul-Licameli:
> Good catch! But I'm fixing it in `WaveClip::GetSampleView`, accounting for `mTrimLeft` like `GetSamples` does.

A sufficient fix for a real bug; so now, please comment in the header that the origin of the argument for `WaveClip::GetFloatSampleView` is in fact the play start.

libraries/lib-wave-track/AudioSegmentSampleView.h
```diff
@@ -0,0 +1,65 @@
...
+    */
+   AudioSegmentSampleView(
+      std::vector<BlockSampleView> blockViews, size_t start,
+      sampleCount length);
+
+   /*
+    * @brief Constructs a silent `AudioSegmentSampleView`.
+    * @param length The number of "silence samples".
+    */
+   AudioSegmentSampleView(sampleCount length);
```
Paul-Licameli, line 43: This function is not intended to work with long lengths.  The argument should be `size_t`.  Where it is used in `WaveTrack::GetOneSampleView`, there is a use of `as_size_t` which could use some comments.

libraries/lib-wave-track/AudioSegmentSampleView.h
```diff
@@ -0,0 +1,65 @@
...
+{
+public:
+   /*
+    * @brief Constructs an `AudioSegmentSampleView` from `BlockSampleView`s.
+    * @param blockViews A continuous sequence of samples, segmented in blocks.
+    * @param start Index of the first sample in `blockViews` to look at.
+    * @param length The number of samples from `start` to look at.
+    * @pre `!blockViews.empty()`
+    * @pre `start < blockViews[0].size()`
+    * @pre `length < sum(sizes(blockViews)) - start`
```
Paul-Licameli, line 33: The last `@pre` is not quite executable as written, so if I wrote it, I would not put only variable names in grave accents and leave the statement informal.

libraries/lib-wave-track/tests/AudioSegmentSampleViewTest.cpp
```diff
@@ -0,0 +1,57 @@
...
+   {
+      AudioSegmentSampleView sut { 3 };
+      std::vector<float> buffer(4);
+      std::fill(buffer.begin(), buffer.end(), 1.f);
+      SECTION("and asked MORE values than it holds.")
+      {
+         REQUIRE(sut.Copy(buffer.data(), 4u) == 3u);
+         REQUIRE(buffer == std::vector<float> { 0.f, 0.f, 0.f, 1.f });
+      }
+      SECTION("and asked LESS values than it holds.")
```
Paul-Licameli, line 27: I like the improved test!

My last irritating trivial comment for today:  "FEWER" not "LESS" is better English according to the grammar books, but a lot of English speakers don't observe that usage.

Paul-Licameli (approved):

Paul-Licameli:
I make a number of suggestions in this round, to improve documentations and (in one place) proper argument types.  I left two boxes un-resolved from the previous round.

Clean it up for final merge if you like, but I none of this is important enough to delay the movement of the card right on the board.

libraries/lib-wave-track/AudioSegmentSampleView.h
saintmatthieu: Good that you noticed the `as_size_t` to in the end call the implicit `sampleCount` ctor.
It probably will never be used for more than `size_t` can represent, yet semantically this is an audio segment between two clips, which `sampleCount` can help realize. (Also, I'd have to change it in the other ctor for consistency.)

dozzzzer:
Tested.

Issue #4618, opened 2023-04-27
EQ effects do not resize properly 


petersampsonaudacity:
### Bug description
Both of the EQ effects now fail to resize properly - the window resizes, but the display no longer resizes with it,  This makes resizing the effect window useless.

This is a regression on 3,2.5 and earlier
![image](https://user-images.githubusercontent.com/50205984/234806656-7f6de4dc-a911-4ee2-8116-b39b7125853d.png)

### Steps to reproduce
1. launch 3.3.x
2. get some audio
3. select the audio
4. Effect > EQ and Filters > Filter Curve EQ (or Graphic EQ)
5. resize the effect window larger
6. Observe: the window enlarges, but the displayed graphic does not enlarge with it

### Expected behavior
The graphic image should resize with the resized effect window (it always used to)

### Actual behavior
The graphic image Fails resize with the resized effect window (it always used to)

### Audacity Version
3.3.0 and 3.3.1 beta

### Operating system
Windows - but assume all OS

### Additional context
Originally reported on the Forum by user CaptainDodgers:
https://forum.audacityteam.org/t/audacity-3-3-0-filter-curve-eq-ui-not-scaling/74502

SteveDaulton:
This issue also occurs on Linux.

petersampsonaudacity:
Still the case in the recently released 3.3.1

doom-hamster:
Glad someone already posted this issue. When the Curve EQ window is small - its hard to be precise, so i usually maximize it. 
How it was before 3.3.0 (maximized window on 2560x1080 display):

![изображение](https://github.com/audacity/audacity/assets/106200718/e032d1c4-9b4a-444c-bdc3-53143c58a6eb)

how it is now:

![изображение](https://github.com/audacity/audacity/assets/106200718/2c785a8e-3c50-4650-81f1-42b975c0bbe2)

petersampsonaudacity:
@LWinterberg @Tantacrul @DilsonsPickles 

This regression bug serious hampers the use of the Audacity EQ effects, it offers a poor UX to the users.

And note that EQ must be among the most commonly used effects

Is there any chance we can see this getting fixed for the next release please?

Ryan-Miller-Audacity:
Tested this issue with the build from PR #4776 and confirmed that built-in effects now properly scales/resizes based on user input. Tested Filter Curve EQ and Graphic EQ on Windows, MacOS, and Linux.

petersampsonaudacity:
[...]

PR #4776, opened 2023-06-07
Install the correct wxEvtHandler for built in Equalizer effects

crsib:
Resolves: #4618 

Recommended:


src/effects/EqualizationUI.cpp
```diff
@@ -38,6 +38,56 @@ BEGIN_EVENT_TABLE(EqualizationUI, wxEvtHandler)
 
 END_EVENT_TABLE()
 
+class EqualizationUIEditor : public EffectEditor, protected wxEvtHandler
```
Paul-Licameli, line 41: Might be in anonymous namespace

crsib: oops

crsib: Citing `DefaultEffectEditor` - `Inherit wxEvtHandler so that Un-Bind()-ing is automatic in the destructor`. I haven't tested that claim, I want minimal possible changes considering that we (possibly) want to target 3.3.3.

Paul-Licameli: I think it's unnecessary, because Bind() wasn't used, but let it stand if it is urgent and you have any doubt.

Paul-Licameli (approved):

Paul-Licameli:
Thanks for fixing the bug I introduced.  Do I need to wear a paper bag on my head tomorrow?

crsib:
I missed it during the code review, so in some way, I fixed my bug :-)

Issue #4792, opened 2023-06-13
Unintended behavior change in export


Paul-Licameli:
### Bug description

Un-panned mono tracks export wrongly as stereo

### Steps to reproduce

1. Generate some sound in a mono track; do not touch the pan slider
2. Export as WAV
3. Open the exported file
 
### Expected behavior

As in 3.3: a mono track

### Actual behavior

a stereo track; also a message that your tracks will be mixed down

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

Very recently introduced bug bisects to 255880d73e6b35b13d2c7217753dbc07cbea3712

PR #4793, opened 2023-06-13
Fix bug introduced at 255880d -- unpanned mono exporting as stereo...

Paul-Licameli:
... there were uninitialized member variables, now they are removed.

Resolves: #4792

Recommended:

### DIFF
--- a/src/export/Export.cpp
+++ b/src/export/Export.cpp
@@ -504,16 +504,20 @@ bool Exporter::ExamineTracks()
    double earliestBegin = mT1;
    double latestEnd = mT0;
 
-   auto &tracks = TrackList::Get( *mProject );
+   auto &tracks = TrackList::Get(*mProject);
 
    bool anySolo =
       !((tracks.Any<const WaveTrack>() + &WaveTrack::GetSolo).empty());
 
-   for (auto tr :
-         tracks.Any< const WaveTrack >()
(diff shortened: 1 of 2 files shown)


src/export/Export.cpp
```diff
@@ -799,19 +803,12 @@ bool Exporter::CheckMix(bool prompt /*= true*/ )
...
-      }
-      else {
-         mChannels = 1;
-      }
-      mChannels = std::min(mChannels,
-                           mPlugins[mFormat]->GetMaxChannels(mSubFormat));
-
-      auto numLeft =  mNumLeft + mNumMono;
-      auto numRight = mNumRight + mNumMono;
+      unsigned channels = mMono ? 1 : 2;
```
saintmatthieu, line 806: `const` please

src/export/Export.h
```diff
@@ -263,9 +263,8 @@ class  AUDACITY_DLL_API Exporter final : public wxEvtHandler
    int mFilterIndex;
    int mFormat;
    int mSubFormat;
-   int mNumSelected;
-   unsigned mNumLeft;
-   unsigned mNumRight;
+   int mNumSelected{};
```
saintmatthieu, line 266: So I suppose this will initialize `mNumSelected` to 0 and `mMono` to false? I prefer explicit assignments.

saintmatthieu (approved):

PR #4797, opened 2023-06-16
Wide sample sequence stretching decorator

### LINKED ISSUE #4791: WideSampleSequence with stretching capability
### Your idea

Write a `WideSampleSequence` decorator that can be injected in the `Mixer` and that stretches the samples of a standard `WaveTrack` using our recently introduced time-stretching library.

### Problem to be solved

This ticket will allow to get forward with productionizing the time-stretching PoC code. It should even allow QA to test this decorator by exporting and rendering tracks. The exported and rendered audio will not be stretched, since clips are not yet resizable, but the mechanics of this new code will still be put to use.

### Prior art

_No response_

### Additional context

_No response_

saintmatthieu:
Resolves: [#4791](https://github.com/audacity/audacity/issues/4791)

Introduces `StretchingSequence`, a `WideSampleSequence` decorator with stretching capability.

It decorates a `WaveTrack` object, which it has a const reference to. It accesses the track's clips to do time stretching at clip level, to be able to have different stretch ratios for different clips.

Based on still unmerged [#4746](https://github.com/audacity/audacity/pull/4746), which introduced `WideSampleSequence`.

Exporting as well as mix-and-render now use a `StretchingSequence` instance. Since clips cannot be stretched yet, the internal stretching algorithm will not yet be used, but all the mechanics above will, which should be tested.

Recommended:

Things to test (not exclusive):
- [x] mix-and-render and exporting, for a variety of tracks, made up of more than one clip, very short ones (a few samples) as well as long ones (more than 10 seconds), with silence in between. The result must be bit-identical to what it was before these changes. The time it takes must not have noticeably increased.
- [x] playback in various modes (normal, scrubbing, etc.)
- [x] code related to envelopes was modified, verify that that works
- [x] benchmarking (if you know what that is - I don't)
- [x] mute & solo


saintmatthieu:
As mentioned in the description, this PR is based on @Paul-Licameli 's current #4746; you may (should) ignore his commits.

Before going through the commits, I recommend taking a look at `./libraries/lib-stretching-sequence/readme.md`, which presents the rationale for the newly introduced methods in `WideSampleSequence` as well as the `lib-stretching-sequence` implementation.

Paul-Licameli:
I don't understand all but here are some things to change.

- The internal cursor, if really necessary, should be a hidden implementation detail of `StretchingSequence`.
   - `ResetCursor` is called (outside `StretchingSequence`) only in `MixAndRender.cpp` immediately after construction.  Elininate that, relying on assumptions about initial state.
   - `WaveTrack` has no need to override the new functions for the cursors.  Those overrides are unreachable.
- `lib-mixer` should not be changed at all.
   - The remaining change there after doing the previous is to add `ClipInterface`.  Instead put that in `lib-stretching-sequence`.  Let `lib-wave-track` depend on that -- in other words, inverting the dependency from what the first attempt did

saintmatthieu:
Thanks Paul for your timely review.
> The internal cursor, if really necessary, should be a hidden implementation detail of `StretchingSequence`
Although possible for this PR, I'd rather not. `MixerSource` will need to call this to change position or direction of play when looping (in `MixerSource::Reposition`) and scrubbing.

saintmatthieu:
> Let lib-wave-track depend on that

Works for me. Done.

saintmatthieu:
>  I think it will still be needed for painting code which will still need that pattern of "channel-major, clip-minor"

I still don't imagine that, after `SampleTrackCache` is removed, anything other than `MixerSource` will need a continuity of samples across clips. Waveform as well as spectrum painting look like they do (or at least could) access data from the clips directly. But I may be wrong, I don't mind removing the deprecation and later remove that method altogether if I turn out to be right. (If I'm right we might not need a `WaveTrack` at all anymore. Except for some `StretchingSequence` methods that only forward calls to `mWaveTrack`, `WaveTrack` isn't an interface `StretchingSequence` needs.)

saintmatthieu:
Yes, the mere action of reading out some output of the time stretcher modifies its state. (Hence `GetNext` is a non-const method.)
I explained the need for a cursor to the best of my ability in this [readme.md](https://github.com/audacity/audacity/pull/4797/files#diff-db4c5a9d199bbfe619baa64ed0283e6f7d3f7bd94d1232ca928640ff72b3dc84). I hope this helps convincing you.

saintmatthieu:
> I dislike any disturbance of lib-mixer interfaces. Is it possible to avoid that by letting the decorator instead deduce the need for a change of its cursor from the pattern of access in the Get calls?

It is possible, and this is how `StretchingSequence` implements legacy `Get`.

`GetNext` does the maintenance of the read position for the client, so this is done in one place, clients simply cannot use it wrongly and it can be tested once and for all (see call in `SECTION("ResetCursor", "reconstructs segment sequence when expected")`, StretchingSequenceTest.cpp).

For this reason, after `SampleTrackCache` is removed, I'd like `MixerSource` to use `GetNext`, even if it possibly still could use legacy `Get`, and one way or another have the possibility to indicate a change of direction.

My hope was that we could also remove legacy `Get`, but spectrum display code uses it. (I am skeptical that this is a good thing, though : that code draws spectra clip by clip, and so first asks for clip boundaries before querying the corresponding samples onto the track. I can't see why this wouldn't be possible by just asking samples from the clips themselves, but it's not the time for me to check.) Most pragmatic is, at least for now, to leave this API, too.

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
```diff
@@ -50,11 +50,9 @@ StaffPadTimeAndPitch::StaffPadTimeAndPitch(
 void StaffPadTimeAndPitch::GetSamples(float* const* output, size_t outputLen)
```
Paul-Licameli, line 50: My first comment on this PR.

I did not review the PR that added this library, so I must become familiar with that first.

I see `libraries/lib-time-and-pitch/CMakeLists.txt` lacks a long-form CMake comment at the top, which is the convention with other libraries.

Please supply that comment, about the general purpose of the library.

libraries/lib-mixer/WideSampleSequence.h
```diff
@@ -47,6 +48,25 @@ class MIXER_API WideSampleSequence
...
+   Attempts to fill `buffers` with the next available `numSamples`. The progress
+   direction is forward, except if specified otherwise in a prior call to
+   `ResetCursor`. The internal cursor will progress in that direction by
+   `numSamples`.
+   @param buffers output buffer
+   @param numChannels the number of channels in `buffers`. // todo(mhodgkinson) review handling of possible inconsistencies between the requested number of channels and that available in internal clips.
+   @param numSamples the number of samples queried
+   */
+   virtual bool
+   GetNext(float* buffers[], size_t numChannels, size_t numSamples) = 0;
```
Paul-Licameli, line 68: I see this header file changes in the second commit in the sequence, and nothing else changes in lib-mixer thereafter.

You know I dislike the notion of an internal, stateful cursor leaking out of the abstract interface, and I would think it more elegant instead to change the `Get` function to take a direction argument, and also to specify the reversal of samples is done in the result.

But I think I will not insist on this if the schedule presses.

libraries/lib-mixer/WideSampleSequence.h
```diff
@@ -47,6 +48,25 @@ class MIXER_API WideSampleSequence
...
+   */
+   virtual void ResetCursor(double t, PlaybackDirection playbackDirection) = 0;
+
+   /*!
+   Attempts to fill `buffers` with the next available `numSamples`. The progress
+   direction is forward, except if specified otherwise in a prior call to
+   `ResetCursor`. The internal cursor will progress in that direction by
+   `numSamples`.
+   @param buffers output buffer
+   @param numChannels the number of channels in `buffers`. // todo(mhodgkinson) review handling of possible inconsistencies between the requested number of channels and that available in internal clips.
```
Paul-Licameli, line 64: Some breaks in this very long comment line

libraries/lib-mixer/WideSampleSequence.h
```diff
@@ -80,13 +82,14 @@ class MIXER_API WideSampleSequence
     @copydetails SampleTrack::GetFloats()
     @param format sample format of the destination buffer
     */
-   virtual bool Get(size_t iChannel, size_t nBuffers, samplePtr buffers[],
-      sampleFormat format, sampleCount start, size_t len,
+   virtual bool Get(
+      size_t iChannel, size_t nBuffers, samplePtr buffers[],
+      sampleFormat format, sampleCount start, size_t len, bool backward,
```
Paul-Licameli, line 87: Good, now please add to the Doxygen comments above, explaining the argument, and the expectation that the outputs are reversed.

And likewise for `GetEnvelopeValues`

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2415,6 +2424,8 @@ void WaveTrack::GetEnvelopeValues(double *buffer, size_t bufferLen,
          clip->GetEnvelope()->GetValues(rbuf, rlen, rt0, tstep);
       }
    }
+   if (backwards)
+      std::reverse(buffer, buffer + bufferLen);
```
Paul-Licameli, line 2446: I see reversal of the envelope values.  But I was expecting to see `ReverseSamples` too, restoring the removed call in MixerSource.

I think this means the first commit breaks backward scrubbing, and that should not be.  I haven't yet reviewed later commits, which might make it work again, when the WaveTrack has its stretching decorator, but logically, shouldn't the proper place to apply reversal be here in WaveTrack?

Paul-Licameli (dismissed):

libraries/lib-stretching-sequence/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+set( SOURCES
```
Paul-Licameli, line 1: Library header comment, please

libraries/lib-stretching-sequence/readme.md
```diff
@@ -0,0 +1,76 @@
+# `StretchingSequence`
```
Paul-Licameli, line 1: If you really want to include this in the tree, the CMakeLists.txt comment can refer the reader to it for more explanation.

Paul-Licameli: I'm skipping the rest of this big commit for now, I will return and review it in detail last

libraries/lib-wave-track/WaveClip.h
```diff
@@ -226,12 +231,10 @@ class WAVE_TRACK_API WaveClip final : public XMLTagHandler
     * available from the returned view is queried through
     * `AudioSegmentSampleView::GetSampleCount()`.
     *
-    * @param ii identifies the channel
     * @param start index of first clip sample from play start
     * @pre `ii < GetWidth()`
     */
-   AudioSegmentSampleView
-   GetSampleView(size_t ii, sampleCount start, size_t length) const;
+   ChannelSampleViews GetSampleView(sampleCount start, size_t length) const override;
```
Paul-Licameli, line 237: If removing `ii` from the argument list is the right decision (I'm not decided), then you should also delete the `@pre` mentioning it

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2284,7 +2284,7 @@ ChannelSampleView WaveTrack::GetOneSampleView(
       //    t0 - clipT0 <= 0,
       // then `length >= len`. `as_size_t` won't narrow.
       const auto len = TimeToLongSamples(t1 - clipT0).as_size_t();
-      auto newSegment = clip->GetSampleView(0u, clipS0, len);
+      auto newSegment = clip->GetSampleView(clipS0, len)[0u];
```
Paul-Licameli, line 2305: This call no longer needs to satisfy the removed precondition, but must now not index the array out of bounds, and so `ClipInterface::GetSampleView` should have a `@post` or `@return` comment guaranteeing the size of the result equals the width.

And `ClipInterface::GetWidth()` should replicate the promise of `WideSampleSequence::NChannels`:

```
   //! A constant property
   /*!
    @post result: `result > 0`
    */
```

libraries/lib-wave-track/WaveClip.h
Paul-Licameli: However, I'm not yet sure I agree that `ii` should be removed as an argument.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -136,12 +136,14 @@ WaveClip::~WaveClip()
...
-   assert(ii < GetWidth());
-   return mSequences[ii]->GetFloatSampleView(
-      start + TimeToSamples(mTrimLeft), length);
+ChannelSampleViews
+WaveClip::GetSampleView(sampleCount start, size_t length) const
+{
+   ChannelSampleViews channels;
+   for (auto i = 0u; i < GetWidth(); ++i)
+      channels.push_back(mSequences[i]->GetFloatSampleView(
+         start + TimeToSamples(mTrimLeft), length));
```
Paul-Licameli, line 145: You are assuming here that all channels will be needed in the cache.  But remember what I explained about drawing code, including spectrogram, needing to stride "channel-major" but "clip-minor".

The caches for spectrogram drawing might now be doing wasteful duplicated work?  Creating a left and right side cache, but only using one side, destroying that cache, then repeating for the right channel?

.

libraries/lib-wave-track/WideClip.cpp
```diff
@@ -0,0 +1,55 @@
...
+  WideClip.cpp
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#include "WideClip.h"
+
+WideClip::WideClip(
+   std::shared_ptr<ClipInterface> left, std::shared_ptr<ClipInterface> right)
+    : mChannels { left, right }
```
Paul-Licameli, line 15: Might `move` the shared pointers

libraries/lib-wave-track/WideClip.h
```diff
@@ -0,0 +1,49 @@
...
+// A provisional class we can get rid of when Audacity actually uses wide clips.
+// In the meantime, `WaveTrack::GetClipInterfaces()` must construct these
+// querying clips of the other channel tracks if any. If that's not clear please
+// take a look into the `GetClipInterface` implementation.
+class WideClip : public ClipInterface
+{
+public:
+   /*
+    * @pre `left` and `right` are equal in sample rate, play start time, play
+    * end time and stretch ratio.
```
Paul-Licameli, line 26: I think you really mean: 
```
`left` is not null, and `right` is null, or else etc.
```

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1985,6 +1984,25 @@ bool WaveTrack::CloseLock() noexcept
...
+   assert(pOwner);
+   ClipConstHolders wideClips;
+   wideClips.reserve(mClips.size());
+   for (auto clipIndex = 0u; clipIndex < mClips.size(); ++clipIndex)
+   {
+      const auto leftClip = mClips[clipIndex];
+      const auto rightClip =
+         NChannels() == 2u && pOwner ?
+            (*++pOwner->Find<const WaveTrack>(this))->mClips[clipIndex] :
+            nullptr;
```
Paul-Licameli, line 2000: You handle un-owned tracks anyway, so weaken the `@pre`.

Now remember too, that clips are not necessarily stored, sorted by start time.  But are you assuming that where you use the results of this function?

libraries/lib-effects/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ set( LIBRARIES
    lib-command-parameters-interface
    lib-numeric-formats-interface
    lib-realtime-effects
+   lib-stretching-sequence-interface
```
Paul-Licameli, line 25: This line, and the one in src/CMakeLists.txt, are unnecessary for the build, and I would prefer you omit them so that the generated library graph does not have unnecessary edges.

Paul-Licameli:
Now reconsidering the long commit.

libraries/lib-wave-track/WaveClip.cpp
Paul-Licameli: Oh, never mind.  The caching for spectrum view will not include the stretching decorator.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: The stretching decorator re-implements ::Get, where the clip samples are reversed before being passed to the stretching algorithm. As a result scrubbing does work when injected that decorator, but you're right, `WaveTrack::Get` should reverse the samples, even if no other code has use for it. `WaveTrack::GetSampleView` isn't a virtual function and `backwards` could/should be removed from its arguments, leaving `CachingPlayableSequence::Get` do the reversal.

libraries/lib-stretching-sequence/readme.md
saintmatthieu: I have to update this, thanks for reminding.

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: Not the stretching decorator, but `CachingPlayableSequence` also uses `WaveClip::GetSampleView`.
Currently it calls it on two separate tracks each with mono clips, so there's no unnecessary work. When we do use wide clips, though, there'll be only one stereo track, `WaveTrack::GetSampleView` will call `GetSampleView` on clips, and yes, doing unnecessary work if it stays at that.
Looks like I indeed need to preserve the `ii` argument ...

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: Ok for the `@pre`.
The stretching decorator sorts the clips this method returns, so we're good.

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
saintmatthieu: Done.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: In the interests of progress I will not block the PR if this is left unfinished, and scrubbing works in the end result, but you are aware of a need to come back and clean it up.

Really you should make sure that the sample reversal is part of the stated contract of `WideSampleSequence` and that all overrides of `Get` do it correctly -- whether it's the undecorated or the decorated WaveTrack.  Maybe some overrides of `Get` won't be directly called in 3.4, but you should not assume it will always be so.  The classes should make sense in isolation and in test harnesses, fulfilling the base class contract.  That's what the L of SOLID is all about.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: Then perhaps the lack of promise of sortedness should be mentioned in a header comment.

libraries/lib-wave-track/WaveClip.cpp
Paul-Licameli: If you just make a `// TODO` note to revisit the efficiency problem in a later PR, I would accept that.

libraries/lib-stretching-sequence/AudioSegment.h
```diff
@@ -0,0 +1,22 @@
...
+
+#pragma once
+
+#include "SampleCount.h"
+
+class AudioSegment
+{
+public:
+   virtual ~AudioSegment() = default;
+   virtual size_t GetFloats(float* const* buffers, size_t numSamples) = 0;
```
Paul-Licameli, line 20: This interface looks strange -- I don't see what tells it how many `buffers` are there.

libraries/lib-stretching-sequence/AudioSegment.h
```diff
@@ -0,0 +1,22 @@
...
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+
+#pragma once
+
+#include "SampleCount.h"
+
+class AudioSegment
```
Paul-Licameli, line 16: Class comments?

libraries/lib-stretching-sequence/AudioSegment.h
```diff
@@ -0,0 +1,22 @@
...
+{
+public:
+   virtual ~AudioSegment() = default;
```
Paul-Licameli, line 19: I would define `AudioSegment.cpp` just to define the destructor non-inline -- so that the linker can avoid de-duplication of the definition of the destructor and the virtual function table of the class; also so that "AudioSegment.h" is included in at least one translation unit before any other header, and we can be sure the header does not require any earlier #include for it to compile.

libraries/lib-stretching-sequence/AudioSegmentFactory.h
```diff
@@ -0,0 +1,40 @@
...
+
+class ClipInterface;
+using ClipConstHolders = std::vector<std::shared_ptr<const ClipInterface>>;
+
+class STRETCHING_SEQUENCE_API AudioSegmentFactory final :
+    public AudioSegmentFactoryInterface
+{
+public:
+   AudioSegmentFactory(
+      size_t sampleRate, size_t numChannels, const ClipConstHolders& clips);
```
Paul-Licameli, line 24: `size_t` seems a strange choice of type to describe a sample rate.  It's not sizing something in memory.

libraries/lib-stretching-sequence/ClipInterface.cpp
```diff
@@ -0,0 +1,3 @@
+#include "ClipInterface.h"
+
+ClipInterface::~ClipInterface() = default;
```
Paul-Licameli, line 3: You did make a tiny translation unit for this one

libraries/lib-stretching-sequence/SilenceSegment.cpp
```diff
@@ -0,0 +1,37 @@
...
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#include "SilenceSegment.h"
+
+#include <algorithm>
+
+SilenceSegment::SilenceSegment(size_t numChannels, sampleCount numSamples)
+    : mNumChannels(numChannels)
+    , mNumRemainingSamples(numSamples)
```
Paul-Licameli, line 17: Uniform member initialization syntax, here and elsehwere, maybe?

libraries/lib-stretching-sequence/SilenceSegment.cpp
```diff
@@ -0,0 +1,37 @@
...
+{
+   const size_t numSamplesToProduce =
+      std::min<long long>(mNumRemainingSamples.as_long_long(), numSamples);
+   for (auto i = 0u; i < mNumChannels; ++i)
+   {
+      auto buffer = buffers[i];
+      std::fill(buffer, buffer + numSamplesToProduce, 0.f);
+   }
+   mNumRemainingSamples -= numSamplesToProduce;
+   return numSamplesToProduce;
```
Paul-Licameli, line 31: Only now do I understand the meaning of the return value of `GetFloats`.  But AudioSegment.h, lacking comments, didn't help me.

Please write some informal description of the contract in that header file.

If it were me, I would even add a pure virtual `size_t GetWidth() const` just so that I could use it to state a precondition on the size of `buffers` and call it just in not-NDEBUG builds to assert the precondition.

libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -0,0 +1,116 @@
...
+         mTotalNumSamplesToProduce - mTotalNumSamplesProduced)
+         .as_size_t();
+   mStretcher->GetSamples(buffers, numSamplesToProduce);
+   mTotalNumSamplesProduced += numSamplesToProduce;
+   return numSamplesToProduce;
+}
+
+bool ClipSegment::Empty() const
+{
+   return mTotalNumSamplesProduced == mTotalNumSamplesToProduce;
```
Paul-Licameli, line 79: Maybe that should be `>=` just to be sure there is eventual emptiness even in the presence of an off-by-few error

libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -0,0 +1,116 @@
...
+         clip.GetWidth(), *this, GetStretchingParameters(clip)))
+{
+}
+
+size_t ClipSegment::GetFloats(float* const* buffers, size_t numSamples)
+{
+   const auto numSamplesToProduce =
+      std::min(
+         sampleCount { numSamples },
+         mTotalNumSamplesToProduce - mTotalNumSamplesProduced)
```
Paul-Licameli, line 70: Maybe `limitSampleBufferSize`; and are you sure that subtraction never underflows?

libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -0,0 +1,116 @@
...
+         .as_size_t();
+   mStretcher->GetSamples(buffers, numSamplesToProduce);
+   mTotalNumSamplesProduced += numSamplesToProduce;
```
Paul-Licameli, line 73: Maybe `mTotalNumSamplesProduced = std::min(mTotalNumSamplesToProduce, mTotalNumSamplesProduced + numSamplesToProduce)`

libraries/lib-stretching-sequence/ClipSegment.cpp
Paul-Licameli: > Maybe `limitSampleBufferSize`; and are you sure that subtraction never underflows?

libraries/lib-stretching-sequence/AudioSegment.h
saintmatthieu: I don't know of evidence that the compilation time gain is worth the manual effort. The early-include verification... yeah, all right.

libraries/lib-stretching-sequence/AudioSegment.h
saintmatthieu: The effective number of channels is known in the context instances are used by both implementation and user. But you're right, looking back that's unusual. I'll pass a vector reference instead.

libraries/lib-stretching-sequence/AudioSegmentFactory.h
saintmatthieu: Never thought of `size_t` that way ... Will use just int.

libraries/lib-stretching-sequence/ClipInterface.cpp
saintmatthieu: Because I initially added this in `lib-mixer` and wanted to be consistent with the code around me :D

libraries/lib-stretching-sequence/ClipSegment.cpp
saintmatthieu: Similarly to above, `mTotalNumSamplesProduced` simply can't be larger than `mTotalNumSamplesToProduce`. I wouldn't want the reader to start doubting this.

libraries/lib-stretching-sequence/AudioSegment.h
saintmatthieu: Oops : I added `AudioSegment.cpp` and moved the dtor definition there, and now I'm getting unresolved external symbol. I can't see why, so at least for now I'll leave it with the inline definition.

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: I'm addressing it.

libraries/lib-mixer/WideSampleSequence.h
```diff
@@ -63,30 +63,34 @@ class MIXER_API WideSampleSequence
...
+         start, len, backwards, fill, mayThrow, pNumWithinClips);
    }
 
    //! Retrieve samples of one of the channels from a sequence in a specified
    //! format
    /*!
     @copydetails SampleTrack::GetFloats()
     @param format sample format of the destination buffer
+    @param backward retrieves samples from `start` to `start + len` if false,
+       else from `start` to `start - len` in reverse order.
```
Paul-Licameli, line 84: I understand the "from" is inclusive and the "to" is exclusive, but maybe you want to say that explicitly.

Paul-Licameli, line 84: `start` is an inclusive bound for the usual forward fetch and the other bound is exclusive.

But as I recall how mixer works, `start` is exclusive and the other inclusive for backwards fetch.

Worth commenting.

libraries/lib-mixer/WideSampleSequence.h
```diff
@@ -126,8 +130,12 @@ class MIXER_API WideSampleSequence
 
    //! Fetch envelope values corresponding to uniformly separated sample times
    //! starting at the given time
-   virtual void GetEnvelopeValues(double *buffer, size_t bufferLen,
-                         double t0) const = 0;
+   /*!
+    @param backwards if true, fetch values in reverse order, from `t0` to
+       `t0 - bufferLen / rate`
```
Paul-Licameli, line 135: Likewise, clarify inclusive and exclusive bounds.

libraries/lib-wave-track/CachingPlayableSequence.cpp
```diff
@@ -47,11 +47,12 @@ float CachingPlayableSequence::GetChannelGain(int channel) const
...
    size_t iChannel, size_t nBuffers, samplePtr buffers[], sampleFormat format,
-   sampleCount start, size_t len, fillFormat fill, bool mayThrow,
-   sampleCount* pNumWithinClips) const
+   sampleCount start, size_t len, bool backwards, fillFormat fill,
+   bool mayThrow, sampleCount* pNumWithinClips) const
 {
    assert(iChannel + nBuffers <= mWaveTrack.NChannels());
-   mCacheHolders = mWaveTrack.GetSampleView(iChannel, nBuffers, start, len);
+   mCacheHolders =
+      mWaveTrack.GetSampleView(iChannel, nBuffers, start, len, backwards);
```
Paul-Licameli, line 55: I see `backwards` is used to construct the cache, but is ignored when using it.  Samples are not taken from the end and reversed.

So this implementation of `CachingPlayableSequence` is incomplete, and so in the first commit, backwards scrubbing won't work right.

But I see that later commits stop using this class, and in the remaining uses for spectrogram, there is never any backwards fetch.

So for now just `assert(!backwards)` and comment about it.

libraries/lib-stretching-sequence/AudioSegment.h
```diff
@@ -0,0 +1,39 @@
...
+public:
+   virtual ~AudioSegment();
+
+   /**
+    * @brief Fills `buffers` with as many as `numSamples` or the number of
+    * remaining samples, whichever is smaller.
+    * @return The number of samples actually provided.
+    */
+   virtual size_t
+   GetFloats(std::vector<float*>& buffers, size_t numSamples) = 0;
```
Paul-Licameli, line 33: You changed `float *const *` to `std::vector<float*>&` so there is information about the number of buffers.

This is all right.  But `std::vector` carries the irrelevancy about the allocator type in its second template argument, and this doesn't allow the caller simply to stack-allocate the pointers if desired.

This is why I wanted `std::span` for such uses, but that is C++20.  However it might be easily imitated in a new utility.

It's also why my preference, lacking `span`, is to pass two arguments, the array of pointers and its size.

Now should there be an explanation about which buffers are filled?  If the caller happens to pass more buffers then the width, should you expect zeroes in the extra buffers or unspecified?

This abstraction still has no interface to query the width.

None of the above is a blocking criticism.

libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -0,0 +1,117 @@
...
+         clip, durationToDiscard) }
+    , mPlaybackDirection { direction }
+    , mStretcher { std::make_unique<StaffPadTimeAndPitch>(
+         clip.GetWidth(), *this, GetStretchingParameters(clip)) }
+{
+}
+
+size_t ClipSegment::GetFloats(std::vector<float*>& buffers, size_t numSamples)
+{
+   assert(buffers.size() == mClip.GetWidth());
```
Paul-Licameli, line 67: Here, an override of a pure virtual function is adding an assertion at the top -- that's an unstated, stricter precondition than the base class specified.

That's bad form (violating the L principle) -- so again, I think `GetWidth` should also be a pure virtual of the base class, so that the base class can formally state the precondition.  And that would answer the question about excess buffers -- they are just not allowed.

libraries/lib-stretching-sequence/tests/StretchingSequenceIntegrationTest.cpp
```diff
@@ -0,0 +1,79 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
```
Paul-Licameli, line 1: This test file fails to compile in the commit that introduces it.  The Audacity executable compiles.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1987,6 +1986,24 @@ bool WaveTrack::CloseLock() noexcept
    return true;
 }
 
+ClipConstHolders WaveTrack::GetClipInterfaces() const
+{
+   assert(IsLeader());
```
Paul-Licameli, line 1991: Again -- if I see a new assertion at the top of a function, I want to see it commented too as an explicit precondition.

The commit introducing this function repairs all the previous build problems.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1987,6 +1986,24 @@ bool WaveTrack::CloseLock() noexcept
...
+   ClipConstHolders wideClips;
+   wideClips.reserve(mClips.size());
+   for (auto clipIndex = 0u; clipIndex < mClips.size(); ++clipIndex)
+   {
+      const auto leftClip = mClips[clipIndex];
+      const auto rightClip =
+         NChannels() == 2u && pOwner ?
+            (*++pOwner->Find<const WaveTrack>(this))->mClips[clipIndex] :
+            nullptr;
+      wideClips.emplace_back(std::make_shared<WideClip>(leftClip, rightClip));
```
Paul-Licameli, line 2002: The informally stated (and not `assert`ed) precondition of the constructor of `WaveClip` can be violated if you make a track with misaligned clips.

But of course it doesn't matter much.  WideClip is provisional only until I make real wide WaveClips and misaligned clips become impossible.

src/CMakeLists.txt
```diff
@@ -1425,6 +1425,7 @@ set( AUDACITY_LIBRARIES
    lib-time-track-interface
    lib-snapping-interface
    lib-audio-io-interface
+   lib-stretching-sequence-interface
```
Paul-Licameli, line 1428: Still an unnecessary arc in the modules graph

tests/WavFileIO.cpp
```diff
@@ -28,9 +28,10 @@ bool WavFileIO::Read(
       std::cerr << "libsndfile could not read "s + inputPath << std::endl;
       return false;
    }
-   const auto numFramesToRead =
-      upTo.has_value() ? static_cast<int>(upTo->count() * sfInfo.samplerate) :
-                         sfInfo.frames;
+   const auto upToSamples = static_cast<int>(upTo->count() * sfInfo.samplerate);
+   const auto numFramesToRead = upTo.has_value() ?
+                                   std::min<int>(sfInfo.frames, upToSamples) :
+                                   sfInfo.frames;
```
Paul-Licameli, line 34: This last commit could be squashed on one of the earlier ones?

libraries/lib-wave-track/WaveClip.h
```diff
@@ -176,14 +181,14 @@ class WAVE_TRACK_API WaveClip final : public XMLTagHandler
...
+   double GetPlayStartTime() const noexcept override;
    void SetPlayStartTime(double time);
 
-   double GetPlayEndTime() const;
+   double GetPlayEndTime() const override;
 
    sampleCount GetPlayStartSample() const;
    sampleCount GetPlayEndSample() const;
-   sampleCount GetPlaySamplesCount() const;
+   sampleCount GetPlaySamplesCount() const override;
```
Paul-Licameli, line 191: Still missing one more `override`, for `GetRate`

libraries/lib-wave-track/CachingPlayableSequence.cpp
Paul-Licameli: I'm actually recommending an L principle violation with that assertion!  But I understand the fix as not worth the effort now

Paul-Licameli:
Okay, I see my critiques of yesterday mostly resolved to my satisfaction; I reiterated two still unresolved; I added more comments about what was rewritten.

I am glad to see you did the work to accomodate the "channel-major" access pattern to caches as needed.

Intermediate commits are not all compilable and free of known bugs (the backwards scrubbing).  I won't hold you to the rule.

I see nothing yet to delay approval, with just suggestions that you can implement as you wish with a rebase.

I still must finish a complete review of the big commit, then likely will approve it.

libraries/lib-stretching-sequence/AudioSegment.h
```diff
@@ -0,0 +1,39 @@
...
+ */
+class STRETCHING_SEQUENCE_API AudioSegment
+{
+public:
+   virtual ~AudioSegment();
+
+   /**
+    * @brief Fills `buffers` with as many as `numSamples` or the number of
+    * remaining samples, whichever is smaller.
+    * @return The number of samples actually provided.
```
Paul-Licameli, line 30: I would clarify, "provided in each buffer"

libraries/lib-stretching-sequence/tests/StretchingSequenceIntegrationTest.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  StretchingSequenceIntegrationTest.cpp
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#include "AudioContainer.h"
+#include "AudioContainerHelper.h"
+#include "AudioSegmentFactory.h"
+#include "MockAudioSegmentFactory.h"
```
Paul-Licameli, line 14: That #include at least is unnecessary, maybe some others too

libraries/lib-stretching-sequence/tests/AudioContainerHelper.h
```diff
@@ -0,0 +1,27 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  TestWaveTrackMaker.cpp
```
Paul-Licameli, line 6: Wrong comment

libraries/lib-stretching-sequence/tests/StretchingSequenceTest.cpp
```diff
@@ -0,0 +1,193 @@
...
+               AudioContainerHelper::GetData(output).data(), start, len,
+               backwards);
+            REQUIRE(output.channelVectors[0] == expected);
+            start -= 2;
+         }
+      }
+
+      SECTION("reconstructs segment sequence when expected")
+      {
+         auto factory = new MockAudioSegmentFactory;
```
Paul-Licameli, line 76: This is only test code, but I don't like naked `new`.  You could `make_unique` and then `move` it.

libraries/lib-stretching-sequence/tests/MockPlayableSequence.h
```diff
@@ -0,0 +1,99 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  MockPlayableSequence.cpp
```
Paul-Licameli, line 6: You mean `.h`

libraries/lib-stretching-sequence/tests/AudioContainerHelper.h
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include "AudioContainer.h"
+
+class AudioContainerHelper
+{
+public:
+   template <typename T = float>
+   static std::vector<T*>
+   GetData(const AudioContainer& container, size_t offset = 0u)
```
Paul-Licameli, line 20: You need `#include <cstddef>` to define `size_t` and fix Linux builds.

libraries/lib-stretching-sequence/tests/StretchingSequenceTest.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+TEST_CASE("StretchingSequence with real audio")
+{
+   const auto filenameStem = "FifeAndDrumsStereo"s;
+   const auto inputPath =
+      std::string(CMAKE_SOURCE_DIR) + "/tests/samples/" + filenameStem + ".wav";
+
+   std::optional<std::string> outputDir;
+   // Uncomment this if you want to look at the output locally.
+   // outputDir = "C:/Users/saint/Downloads/StaffPadTimeAndPitchTestOut";
```
Paul-Licameli, line 124: That's a path on your computer only!

libraries/lib-stretching-sequence/tests/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+      StretchingSequenceTest.cpp
+      StretchingSequenceIntegrationTest.cpp
+      WaveClipSegmentTest.cpp
+      TestWaveClipMaker.cpp
+      TestWaveClipMaker.h
+      TestWaveTrackMaker.cpp
+      TestWaveTrackMaker.h
+   MOCK_PREFS
+   MOCK_AUDIO
+   WAV_FILE_IO
```
Paul-Licameli, line 27: I think you do need MOCK_PREFS but why MOCK_AUDIO?

libraries/lib-stretching-sequence/tests/FloatVectorClip.h
```diff
@@ -0,0 +1,60 @@
...
+   {
+      return playStartTime + GetPlayDuration();
+   }
+
+   double GetStretchRatio() const override
+   {
+      return stretchRatio;
+   }
+
+   void ShiftTo(double t);
```
Paul-Licameli, line 49: Declared, nowhere defined

libraries/lib-stretching-sequence/tests/MockSampleBlock.h
```diff
@@ -0,0 +1,51 @@
...
+   bool
+   GetSummary64k(float* dest, size_t frameoffset, size_t numframes) override;
+
+   size_t GetSpaceUsage() const override;
+
+   void SaveXML(XMLWriter&) override;
+
+   size_t DoGetSamples(
+      samplePtr dest, sampleFormat destformat, size_t sampleoffset,
+      size_t numsamples);
```
Paul-Licameli, line 40: add `override` please

libraries/lib-stretching-sequence/tests/TestWaveClipMaker.h
```diff
@@ -0,0 +1,45 @@
...
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#pragma once
+
+#include "AudioContainer.h"
+#include "SampleFormat.h"
+#include "WaveClip.h"
+#include "WaveTrack.h"
```
Paul-Licameli, line 16: `lib-stretching-sequence` does not depend on `lib-wave-track`, but the opposite.

However the test executable depends on `lib-wave-track`.

The build system let you do this, but I think I don't like it and should consider it a cycle.

Instead, you could lift all testing files that depend on WaveTrack and WaveClip into a new `lib-wave-track/test` directory.

That may mean `lib-wave-track/test` might need to used some of the mocks defined in `lib-stretching-sequence/test`.

Does the test build system allow that dependency?  @crsib ?

libraries/lib-stretching-sequence/ClipSegment.h
```diff
@@ -0,0 +1,48 @@
...
+#include "TimeAndPitchInterface.h"
+
+#include <memory>
+
+class ClipInterface;
+using ChannelSampleViews = std::vector<AudioSegmentSampleView>;
+
+class STRETCHING_SEQUENCE_API ClipSegment final :
+    public AudioSegment,
+    public TimeAndPitchSource
```
Paul-Licameli, line 25: I think the second base could be `private`

libraries/lib-stretching-sequence/StretchingSequence.h
```diff
@@ -0,0 +1,83 @@
...
+{
+public:
+   static std::shared_ptr<StretchingSequence>
+   Create(const PlayableSequence&, const ClipConstHolders& clips);
+   static std::shared_ptr<StretchingSequence>
+   Create(const PlayableSequence&, const ClipHolders& clips);
+
+   StretchingSequence(
+      const PlayableSequence&, int sampleRate, size_t numChannels,
+      std::unique_ptr<AudioSegmentFactoryInterface>);
```
Paul-Licameli, line 37: If the constructor is not meant for public use, but instead only the `Create` factories -- maybe use the `CreateToken` trick as in `EffectStage`.  The constructor has to be public, but `Create` needs make-shared so `std` functions have to call the constructor.  But the trick is a way to restrict the usage without the `std` functions needing to be `friend`s.

libraries/lib-stretching-sequence/tests/TestWaveClipMaker.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+**********************************************************************/
+#include "TestWaveClipMaker.h"
+#include "AudioContainerHelper.h"
+#include "MockSampleBlockFactory.h"
+
+TestWaveClipMaker::TestWaveClipMaker(
+   int sampleRate, SampleBlockFactoryPtr factory)
+    : mSampleRate(sampleRate)
+    , mFactory(factory)
```
Paul-Licameli, line 18: More uniform initializations might be done for this and other test classes.

Paul-Licameli (approved):
I'm done!  Build fixes for Linux are the only must-do.

Read my more recent comments.  I trust you with minor modifications to comments as you see fit, and also the suggested changes to fix the builds of intermediate commits.

libraries/lib-stretching-sequence/tests/AudioContainerHelper.h
saintmatthieu: Thank you!

libraries/lib-stretching-sequence/tests/FloatVectorClip.h
saintmatthieu: Removed - thanks!

libraries/lib-stretching-sequence/tests/CMakeLists.txt
saintmatthieu: `AudacityProject::Create()` leads to an audio-io setup in my `TestWaveTrackMaker`. Not sure if/how it could be avoided.

libraries/lib-stretching-sequence/tests/StretchingSequenceIntegrationTest.cpp
saintmatthieu: Makes sense, there must be modifications in `WaveTrack` needed for this integration test that I added in follow-up commits.

libraries/lib-stretching-sequence/tests/StretchingSequenceTest.cpp
saintmatthieu: I'd have refer to `factory` after the move. It's feasible but more complicated.

libraries/lib-stretching-sequence/tests/TestWaveClipMaker.h
saintmatthieu: I don't think we have a cyclic dependency. In fact, if I could remove `lib-stretching-sequence` from the list of libraries needed by this test executable and it would still compile. It's just one unnecessary arrow in the dependency graph ;)

libraries/lib-stretching-sequence/AudioSegment.h
saintmatthieu: If we're picky, yes, we'll want `GetWidth` and specifications about `buffers.size() > GetWidth`.
`AudioSegment`s are used in `StretchingSequence`, which, like all `PlayableSequence`s in Audacity as of now, has either only mono or stereo clips. This is why a need for `GetWidth` wasn't felt in the first place.
I suggest we leave it at that for now, and address it properly when/if we ever support tracks of mixed channel counts.

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "ClipInterface.h"
+#include "ClipSegment.h"
+#include "SilenceSegment.h"
+
+#include <algorithm>
+
+using ClipConstHolder = std::shared_ptr<const ClipInterface>;
+
+AudioSegmentFactory::AudioSegmentFactory(
+   int sampleRate, int numChannels, const ClipConstHolders& clips)
```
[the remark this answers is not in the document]
saintmatthieu: I wish C++ had a strictly-positive-integer type ...

libraries/lib-stretching-sequence/ClipSegment.cpp
saintmatthieu: Oh well, you win :D

libraries/lib-stretching-sequence/ClipSegment.h
saintmatthieu: Could it ? Checking ...
... no, because it passes itself as `TimePitchAndSource` to `mTimeAndPitch`.

libraries/lib-stretching-sequence/StretchingSequence.h
saintmatthieu: It's ok, it can be used publicly, it's just that clients usually don't want to bother about injecting a factory and are happy with the default one. One test case, though, needs it.
But that's a nice trick, never thought about it.

libraries/lib-wave-track/CachingPlayableSequence.cpp
saintmatthieu: I agree. It'd be hard to test, too. Spot on, though.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: > WaveTrack::Get should reverse the samples

I meant by that that I was going to address it. It's done now.

src/CMakeLists.txt
saintmatthieu: addressed

saintmatthieu:
Comments addressed in squashed commits. Hoping for a green build and QA round ...

libraries/lib-stretching-sequence/tests/TestWaveClipMaker.h
Paul-Licameli: But I think the test directory for a library should be responsible to test only that library.

If a test executable involves other libraries higher in the graph, that argues for moving those tests.

dozzzzer:
Moving back to `To Do` @saintmatthieu for two reasons:

- **Wrong mixing-and-rendering behavior**: when duplicating a piece of a mono clip, then making these two mono tracks into a stereo track, mixing-and-rendering it, the resulting track's right channel has wrong data. Compare:

https://github.com/audacity/audacity/assets/70063970/b454a7c5-7523-4dea-83ef-e33757c77c0d

vs recent `master`:

https://github.com/audacity/audacity/assets/70063970/b6b2f59c-22cb-4961-859b-7d71f815f424

- **crash when duplicating a tiny portion of a mono track** surrounded by silence > merging two mono tracks into a stereo track > mixing-and-rendering:

https://github.com/audacity/audacity/assets/70063970/edaf37ef-d73f-4278-9d3c-552c6877e323

[Audacity_dbgrpt-20167-20230705T132231.zip](https://github.com/audacity/audacity/files/11957375/Audacity_dbgrpt-20167-20230705T132231.zip)

Paul-Licameli (approved):

Issue #4800, opened 2023-06-16
Added initial theming support to the Qt application


ksoze95:
Add support for theming.  Implement both dark and light themes.

PR #4802, opened 2023-06-16
Added theming support to Qt application

ksoze95:
Resolves: #4800 

The hover highlight over the track control, on the side bar, does not behave correctly when in dark mode.  Raised #4801 to track this issue.

Recommended:


src/qt/uicomponents/ApplicationConfiguration.cpp
```diff
@@ -1,7 +1,99 @@
+#include <algorithm>
 #include <QFontDatabase>
 #include <QGuiApplication>
+#include <QMap>
+#include <QString>
+#include <QVariant>
 #include "ApplicationConfiguration.h"
 
+namespace
```
crsib, line 9: I would much prefer to see themes loaded from the QML, not from C++. We want custom themes, so this must be supported anyway.

ksoze95: Done.

vsverchinsky: Is there a reason not to use [QPalette](https://doc.qt.io/qt-6.2/qpalette.html#details)?

Issue #4808, opened 2023-06-19
Dropping a clip from another track with different sample rate doesn't work


saintmatthieu:
### Bug description

Audacity allows dropping a clip of sample rate x onto a track of sample rate y. (It does so by resampling.) Looks like this got broken, possibly with #4739.

### Steps to reproduce

1. Create track A with some audio.
2. Create track B, set its rate to something other than track A's.
3. Drag and drop A's clip onto B.

UPDATE:  Also test:
1. Create a stereo track.
2. Save project.
3. Close.
4. Reopen the saved project.

### Expected behavior

Clip dragged from A onto B is playable from B.

And successful opening of files.

### Actual behavior

We first get the following assert: 
![image](https://github.com/audacity/audacity/assets/22740106/2d356bdb-7f70-46cb-8c31-ce31405f5a8d)
If I choose to ignore the assert and continue, both tracks are now empty.

And opening of files fails.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

petersampsonaudacity:
@saintmatthieu 

I'm not sure I agree with your expected behavior
>Clip dragged from A onto B is playable from B.

I am strongly minded to think that a clip should NOT be allowed to be dragged into a track with a different sample rate - and should be blocked with an error message.  
Just as we now block a stereo track with channels with different rates 

Or am I missing something

saintmatthieu:
Thanks David,
I understand better now:
You don't get the assert because that's a debug assert and the published builds are release builds.
My track B is actually also not empty, but my 1-second, 44.1kHz noise segment became over 13mn long :
![image](https://github.com/audacity/audacity/assets/22740106/cd14ec12-1b0a-4e6e-bb7a-07897908e972)

saintmatthieu:
Thanks @dozzzzer too !
Observe as well that the displayed track rates go bezirk. (In my screenshot above this looks over 7 billion Hz/sec)

saintmatthieu:
Yes indeed. @Paul-Licameli also mentioned that behavior in his bug description of https://github.com/audacity/audacity/pull/4739.

saintmatthieu:
Sorry @petersampsonaudacity , I called you David - my apologies :D

petersampsonaudacity:
Works fine in 3.3.3 too - so this is a regression on 3.3.3

petersampsonaudacity:
[...]

PR #4823, opened 2023-06-21
Fix bug dragging a clip into a track with a different rate

Paul-Licameli:
Resolves: #4808

Recommended:


petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch build:  audacity-win-3.4.0-alpha-20230621+140ef2c-x64

I confirm that this now works fine once again with this branch build on Windows 10

saintmatthieu (approved):
This obviously fixes this bug, but I still have a bad feeling about `mLegacyRate` also used as argument to `SetClipRates` without this greater-than-zero check. Please review that place and consider adding this check there too.

Paul-Licameli:
> `mLegacyRate`

That would be a one line fix and I will do it

Paul-Licameli:
Rebased onto recent master, and added the suggested additional check for zero.

PR #4827, opened 2023-06-22
NumericTextControl component for qt version

vsverchinsky:
Contains few commits that do some necessary refactoring.

Depends on:
 - #5046

Recommended:

### DIFF
--- a/src/qt/numeric-formats/CMakeLists.txt
+++ b/src/qt/numeric-formats/CMakeLists.txt
@@ -1,0 +1,28 @@
+set(MODULE_NAME numeric-formats)
+
+set(MODULE_SOURCES
+   NumericConverterFormatterAdapter.cpp
+   NumericConverterFormatterAdapter.h
+   NumericConverterFormatterAdapterFactory.cpp
+   NumericConverterFormatterAdapterFactory.h
+)
+
+set(MODULE_QML_SOURCES
+   qml/NumericTextControl.qml
+)
+
+audacity_qml_module(${MODULE_NAME}
+   URI "Audacity.NumericFormats"
+   VERSION 1.0
+   STATIC
+   DEPENDENCIES
+      QtQuick
+      Audacity.UiComponents
+   QML_FILES
+      ${MODULE_QML_SOURCES}
+   SOURCES
+      ${MODULE_SOURCES}
+   LIBRARIES
+      lib-qt-init-interface
+      lib-numeric-formats-interface
+)
--- a/src/qt/CMakeLists.txt
+++ b/src/qt/CMakeLists.txt
@@ -23,6 +23,7 @@ qt_add_resources(QT_RESOURCES
 
 add_subdirectory(uicomponents)
 add_subdirectory(uithemes)
+add_subdirectory(numeric-formats)
 
 set(QT_TRANSLATION_SOURCES
    translations/de.ts
@@ -59,6 +60,7 @@ audacity_qml_module(
       QtQuick
       Audacity.UiComponents
       Audacity.UiThemes
+      Audacity.NumericFormats
    QML_FILES
       ${QML_SOURCES}
    SOURCES
@@ -71,8 +73,11 @@ audacity_qml_module(
       KDAB::kddockwidgets
       uicomponentsplugin
       uithemesplugin
+      numeric-formatsplugin
       lib-qt-init
       lib-project
+      lib-audio-io
+      lib-numeric-formats
 )
 
 set_target_properties(
--- a/src/qt/main.cpp
+++ b/src/qt/main.cpp
@@ -8,7 +8,9 @@
 #include <QQuickWindow>
 #include <QWindow>
 
+#include "AudioIO.h"
 #include "BasicSettings.h"
+#include "Prefs.h"
 #include "Project.h"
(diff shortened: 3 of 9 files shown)


libraries/lib-preferences/FileConfig.h
```diff
@@ -12,39 +12,39 @@
...
-class PREFERENCES_API FileConfig : public wxConfigBase
+#include <wx/confbase.h>
+
+///@brief Wraps wxConfigBase adding dirty flag that prevents extensive
+///redundant writes when wxConfigBase::Flush() is called. Replicates
+///wxConfigBase interface.
+///Requires another wxConfigBase instance that is used as an actual
+///storage (e.g. all write/reads are delegated to that object)
+//TODO: wxFileConfig already has dirty flag
+class PREFERENCES_API FileConfig final : public wxConfigBase
```
crsib, line 23: This looks like a proper hotfix, but I would instead push it further. 

1. I would like it to use Audacity specific interface. Much smaller one, only providing the functionality Audacity needs. `SetPath` is redundant as well.
2. I don't quite get the need for `SetImpl`, especially because it expects to implement the same interface. I think something like `GlobalHook` should be used to initialize `gPrefs` (or anything else really).

libraries/lib-project/CMakeLists.txt
```diff
@@ -19,10 +19,23 @@ set( SOURCES
...
    ProjectStatus.h
 )
+
 set( LIBRARIES
    lib-registries-interface
    lib-xml-interface
 )
+
+
+if(${_OPT}enable_experimental_qt_support)
```
crsib, line 29: This should go to a `lib-project-qml` or `lib-qml-support` or whatever. `lib-project` should remain green for all build configurations

src/qt/numeric-formats/qml/TimecodeControl.qml
```diff
@@ -0,0 +1,348 @@
+import QtQuick
```
crsib, line 1: Generally, it's not a TimecodeControl, it's really a NumericTextControl. It is suitable right now to display Time, Duration, Frequency and Bandwith

src/qt/numeric-formats/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+set(MODULE_NAME numeric-formats)
```
crsib, line 1: This raises a good question about where to put Qt-related modules. Personally I would just put in under `modules`, but with `qt-` as prefix

crsib (approved):
I do want to see a different approach to how Audacity works with config, but I would rather make it a task for the Audacity `master` first.

libraries/lib-preferences/FileConfig.h
vsverchinsky: > I don't quite get the need for SetImpl, especially because it expects to implement the same interface

It used to create `wxFileConfig` inside `FileConfig` which was performing all the actual job. Now it injected instead. This refactoring made it clear that the only purpose of `FileConfig` is prevent redundant disk writes, which is actually what `wxFileConfig` do as well...

libraries/lib-registries/ClientData.h
```diff
@@ -255,7 +255,7 @@ class Site
     Usually agrees with the size() of each site unless some registrations happened later
     than some Site's construction.
     */
-   static size_t slots() { return GetFactories().mObject.size(); }
+   static size_t numFactories() { return GetFactories().mObject.size(); }
```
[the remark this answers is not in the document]
vsverchinsky: 🤷 I thought that it would be better to preserve original code style

vsverchinsky:
Moved `ProjectQMLEnvironment` to `lib-project-qml`
`TimecodeControl` -> `NumericTextControl`
MacOS compilation fixes

src/qt/numeric-formats/CMakeLists.txt
crsib: I still feel that this library should be under `libraries`, not under `src`.

Probably under `libraries/qt`.

Paul-Licameli:
Will this be merged soon into the Qt branch?

But I don't want to accumulate divergence in libraries/ between that and master if that can be avoided.

Can there be a separate pull request against master, and after merge it, merge master into the qt branch, then do the rest of this PR?

src/qt/numeric-formats/CMakeLists.txt
Paul-Licameli: > This raises a good question about where to put Qt-related modules. Personally I would just put in under `modules`, but with `qt-` as prefix

Note that I have begin using a `lib-wx-` naming convention.

crsib:
I don't see any libraries in this PR that should go into the master, so I strongly disagree. The only thing is the FileConfig, which should be done differently later.

Paul-Licameli:
> I don't see any libraries in this PR that should go into the master, so I strongly disagree. The only thing is the FileConfig, which should be done differently later.

I see one new library special to qml, but many changes to existing libraries.  True, mostly minor things, changing `#include` directives.  Still I would rather not accumulate divergences.  Are all the little changes incidental to fixing FileConfig?  Then I hope to see a PR for just that.

ksoze95 (approved):

Issue #4769, opened 2023-06-06
Crash when pasting stereo data into project with Mono track preceding stereo track with all tracks selected


Ryan-Miller-Audacity:
### Bug description

When pasting stereo data into a project with a leading mono track before a stereo track, if the user selects all (rather than specifically selecting the stereo track) Audacity will crash when you attempt to paste.

https://github.com/audacity/audacity/assets/88568755/4c2843fa-c20f-4dd4-ac34-49cc9139fa62

### Steps to reproduce

1) Open Audacity and generate some audio in a stereo track
2) Copy the stereo track data (Ctrl+A, Ctrl+C)
3) Ctrl + N
4) and create a mono track and a stereo track
5) Ctrl+A to select all tracks
6) Ctrl+V to paste

### Expected behavior

Audacity does not crash and with the copy/paste changes in PR #4730 the data would be correctly pasted into the stereo track (skipping the mono track when all tracks are selected)

### Actual behavior

Audacity crashes 

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

There is a rather obvious workaround for this which is to select the track that you are intending to paste into rather than using Ctrl+A to select all. 

This was discovered while testing the paste updates in PR #4730, but was not introduced as part of those changes. I tracked it back to being introduced in Audacity 3.3.0. Versions previous to that give an error of "Coping stereo audio into a mono track is not allowed." when attempting these steps. Versions 3.3.0 or later produce the crash.

Sentry Report: https://sentry.audacityteam.org/share/issue/a0a69ecaca714f8a8fee9505ef37f41b/

saintmatthieu:
Still happens on current master. Call stack:
```
Audacity.exe!std::vector<std::shared_ptr<WaveClip const>,std::allocator<std::shared_ptr<WaveClip const>>>::_Unchecked_begin() Line 1914 (c:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\include\vector:1914)
Audacity.exe!``anonymous namespace'::EstimateCopiedBlocks'::`2'::<lambda_1>::operator()(const WaveTrack & waveTrack) Line 173 (c:\Users\saint\git\github\saintmatthieu\audacity2\src\menus\EditMenus.cpp:173)
[...]

PR #4833, opened 2023-06-23
Track copy is all-or-nothing now, so simplify estimation of copy

Paul-Licameli:
Resolves: #4769

Recommended:

### DIFF
--- a/src/menus/EditMenus.cpp
+++ b/src/menus/EditMenus.cpp
@@ -88,92 +88,27 @@ bool DoPasteText(AudacityProject &project)
    return false;
 }
 
-/*
-Track copy helper function. 
-When copying tracks we consider two cases when pasting:
-1. There is no selection
-2. Not empty region is selected
-In the first case we copy all tracks from src (used in simplified paste method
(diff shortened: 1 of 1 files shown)


src/menus/EditMenus.cpp
```diff
@@ -89,75 +89,19 @@ bool DoPasteText(AudacityProject &project)
...
-   }
+   for (auto waveTrack : src.Any<const WaveTrack>())
+      f(*waveTrack);
 }
 
 wxULongLong EstimateCopyBytesCount(const TrackList& src, const TrackList& dst)
 {
    wxULongLong result{};
-   ForEachCopiedWaveTrack(src, dst, [&](const WaveTrack& waveTrack) {
+   ForEachCopiedWaveTrack(src, [&](const WaveTrack& waveTrack) {
```
vsverchinsky, line 104: could be just:
`for (auto waveTrack : src.Any<const WaveTrack>())`

vsverchinsky (dismissed):

vsverchinsky (approved):

Issue #4849, opened 2023-06-28
Have clips know project tempo


saintmatthieu:
The stretching of clips depends also on project tempo and project tempo changes. Clips (and tracks) must know project tempo at all times, which they currently don't.

dozzzzer:
Tested:

- For non-audio tracks changing the project tempo results in the track's horizontal scaling staying in sync with the beats-and-measure ruler
- Audio tracks still stay in sync with the time ruler
- Changing project tempo for all track types making sure it is correct
- Copy/paste between the projects
- Undo/Redo
- Save, close and reopen projects

PR #4853, opened 2023-06-28
Have clips know project tempo

saintmatthieu:
Resolves: #4849 

The stretching of clips depends also on project tempo and project tempo changes. Clips (and tracks) must know project tempo at all times, which they currently don't.

Recommended:


saintmatthieu:
This PR is based on currently-in-QA https://github.com/audacity/audacity/pull/4797

saintmatthieu:
Message of second commit:
> lib-numeric-format implements `ProjectTempoPublisher` and offers a `ProjectTempoListenerInterface`.
Whoever implements the listener must be able to use the `TrackList` to make sure all tracks of a project get the update, AND to make sure newly added tracks also get the project tempo set.
Since lib-track and lib-numeric-format do not depend on one another, I have the source code itself implement the listener.

My first approach during the early PoC days was to have `lib-wave-track` depend on `lib-numeric-format`. It was a request from @Paul-Licameli that I avoid this dependency and we came up with this solution. I still am not convinced that it is the wisest choice, though, since it comes at the cost of rather lots of synchronization code. Also, at it stands I implement `ProjectTempoListenerInterface` in `src` itself, which means more work for the Qt migration I suppose?

libraries/lib-numeric-formats/ProjectTempoListenerInterface.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <memory>
+
+class NUMERIC_FORMATS_API ProjectTempoListenerInterface
+{
+public:
+   virtual ~ProjectTempoListenerInterface();
+   virtual void SetProjectTempo(double newTempo) = 0;
```
Paul-Licameli, line 9: When you need an abstract interface with just one interesting virtual function -- just make an alias for a `std::function<void(double)>`

src/ProjectManager.cpp
```diff
@@ -92,9 +93,11 @@ const ProjectManager &ProjectManager::Get( const AudacityProject &project )
    return Get( const_cast< AudacityProject & >( project ) );
 }
 
-ProjectManager::ProjectManager( AudacityProject &project )
-   : mProject{ project }
-   , mTimer{ std::make_unique<wxTimer>(this, AudacityProjectTimerID) }
+ProjectManager::ProjectManager(AudacityProject& project)
+    : mProject { project }
+    , mTempoListener { project, TrackList::Get(project) }
+    , mTempoPublisher { project, mTempoListener }
```
Paul-Licameli, line 99: Can the listener and publisher be new attached objects on the project, so that no intrusion in ProjectManager is necessary?

src/ProjectTempoListener.cpp
```diff
@@ -0,0 +1,27 @@
+#include "ProjectTempoListener.h"
+#include "Project.h"
+#include "ProjectTimeSignature.h"
+#include "WaveTrack.h"
+
+ProjectTempoListener::ProjectTempoListener(
```
Paul-Licameli, line 6: I think the listener and publisher classes should just be fused into one.

Paul-Licameli:
Or put this another way -- can you make a bigger PR?  Show me how `WaveClip::GetStretchRatio` becomes a real function and not just the stub returning 1.0 always.

Then I'd say, change it to take default tempo as an argument, figure out all other necessary changes in `lib-stretching-sequence` to supply that argument, and then working up the call stack, you would arrive at the uses of `StretchingSequence::Create`.

Then maybe all this effort with publishers and listeners won't be needed, and we will have one less consistency invariant in the data structures that we would need to worry about preserving.

src/ProjectManager.cpp
saintmatthieu: Sounds like a use case for an attachment, indeed : these blokes are just here to live, no one manipulates them ...
The attachment should still be defined in `src`, though, the only possible get-together place for `ProjectTimeSignature` and `TrackList` without modifying the dependency graph.

src/ProjectTempoListener.cpp
saintmatthieu: But that'd mean moving the implementation of the publisher down in `src`, meaning more code to re-implement when migrating to Qt.

saintmatthieu:
The stretch ratio (not only clip-specific, but that which takes project tempo into account too) isn't only needed during playback, but also for painting. And we must care to look at up-to-date project tempo values, too.
That said, playback assumes constant project tempo (we want to disable the possibility of tempo change during playback), so your suggestion would probably work in this case.
As for the painting, it's all in `src`, so yeah, looks like an up-to-date project tempo value can be queried for each painting pass.
I'll try it out.

crsib:
Constant tempo during the playback is not the design choice, we just had no time to make the tempo and time signature variable.

And even with tempo being constant for the whole project duration, I am not sure that changing the tempo during the playback is disabled now or should be disabled in the future.

I haven't looked through the PR yet, so I don't know why there is a need for another publisher.

saintmatthieu:
> Constant tempo during the playback is not the design choice, we just had no time to make the tempo and time signature variable.

I can't think of anything UX-wise that should prevent tempo change during playback, but even with more time it seems to me that the implementation cost is not worth the additional convenience for the user. In any case, yes, we don't have time in this iteration. I believe this was mentioned during stand-up, there seemed to be a consensus that tempo change could be disabled during playback, and no one mentioned it should be done on the longer term. You'd be the first :)

Paul-Licameli:
Tempo change during playback might raise the problem of proper updates without race conditions when a worker thread needs to query the tempo.

Maybe `GetStrerchRatio` needs a context type that will contain a constant double in a first version but a virtual function to get the varying tempo in future and the override could use atomics.

Paul-Licameli:
See also the TrackArtist structure and uses of TrackArtist::Get.  That's how other context information can be got into drawing routines.

See TrackPanel::DrawTracks where things are set up in this context.

saintmatthieu:
Trying the project-tempo-passed-as-argument-when-needed approach ... There's a difficulty, that could be worth tackling.

`WaveClip` boundary getters and setters (`GetPlayStartTime`, `SetSequenceStartTime`, etc) use _seconds_. Now consider situations such as this one :
1. User moves clip, which updates its boundary member variables.
2. User changes tempo.
3.  Painting wants to paint, which requires clip to report its time boundaries. The correct value would be `boundaryInSeconds * tempoWhenBoundaryWasSet / currentTempo`. Where do we get `tempoWhenBoundaryWasSet` ?

Possible approaches:
* Publisher approach: Like I've been doing so far: each project tempo change is published to clip, which modifies its boundaries accordingly. This requires this PR.
* Beat approach: Clip boundaries are expressed in beats, as well as all its public methods.

| | Pros | Cons |
|--- | ------------- | ------------- |
| **Publisher** | * Proven to work<br/>* likely fastest delivery | * need to maintain up-to-date values<br/>* rounding errors might build up (each tempo change implies a `*=` operation on those time values)  |
| **Beat** | * clips don't worry about tempo changes<br/>* consistent with UI : project tempo changes modify position in time view, not B&M view -> might have unforeseen positive side-effect | * clients must know project tempo when asking these values (may not be a bad thing)<br/>* compatibility |

**Explanation about the compatibility problem of the beat approach** :
Old Audacity projects store boundary-related clip member variables in seconds. To convert them in beats, the project tempo must be known. This information, available somewhere in the project file, will have to be communicated to the clip.
Backward compatibility (opening 3.4+ projects with older versions), if deemed important, could also be achieved by saving time values, in which case project tempo at save time will also have to be communicated.

I have a goodish feeling about the beat approach. I'll give it a try, at least until new feedback arrives.

Paul-Licameli:
> Trying the project-tempo-passed-as-argument-when-needed approach ... There's a difficulty, that could be worth tackling.
> 
> `WaveClip` boundary getters and setters (`GetPlayStartTime`, `SetSequenceStartTime`, etc) use _seconds_. Now consider situations such as this one :
> 
> 1. User moves clip, which updates its boundary member variables.
> 2. User changes tempo.
> 3. Painting wants to paint, which requires clip to report its time boundaries. The correct value would be `boundaryInSeconds * tempoWhenBoundaryWasSet / currentTempo`. Where do we get `tempoWhenBoundaryWasSet` ?
> 
> Possible approaches:
> 
> * Publisher approach: Like I've been doing so far: each project tempo change is published to clip, which modifies its boundaries accordingly. This requires this PR.
> * Beat approach: Clip boundaries are expressed in beats, as well as all its public methods.
> 
> Pros	Cons
> **Publisher**	* Proven to work
> * likely fastest delivery	* need to maintain up-to-date values
> * rounding errors might build up (each tempo change implies a `*=` operation on those time values)
> **Beat**	* clips don't worry about tempo changes
> * consistent with UI : project tempo changes modify position in time view, not B&M view -> might have unforeseen positive side-effect	* clients must know project tempo when asking these values (may not be a bad thing)
> * compatibility
> **Explanation about the compatibility problem of the beat approach** : Old Audacity projects store boundary-related clip member variables in seconds. To convert them in beats, the project tempo must be known. This information, available somewhere in the project file, will have to be communicated to the clip. Backward compatibility (opening 3.4+ projects with older versions), if deemed important, could also be achieved by saving time values, in which case project tempo at save time will also have to be communicated.
> 
> I have a goodish feeling about the beat approach. I'll give it a try, at least until new feedback arrives.

My understanding of “beats” is that you normalize out any tempo information when modifying the clip.  Then only current tempo is needed.

I think that may work better.

Maybe this PR should be closed and you will propose another that really stores new information in clips so I can better understand the whole.

saintmatthieu:
> My understanding of “beats” is that you normalize out any tempo information when modifying the clip. Then only current tempo is needed.

Modifying the project tempo also counts as clip modifications (its boundaries).

saintmatthieu:
Closing this PR for now makes sense until new developments.

saintmatthieu:
It's quite thrilling ... hard to formulate simply and convincingly, even to myself. I think key to understand the problem is indeed that tempo changes, and all time boundaries expressed in seconds are not valid anymore. What remains constant are _beats_, seconds don't. Imagine:
1. Hey clip, set start time to 60s.
2. I'm going to change project tempo now ...
3. Hey clip, what's your start time again ?
Clip cannot tell, unless it knows by how much the project tempo has changed between. It is _not_ sufficient for the client to know tempo in step 3 : the answer in 3. will be different if all three following examples:
A) Tempo was 60bpm in 1 and is now 120bpm in 3 => answer is 30s
B) Tempo was 120 and is now 60 => answer is 120s
C) Tempo was 120 and is now 240 => answer is 30s.

I've said this already earlier: the publisher approach makes sure that the tempo is known at any time, which enables giving the correct answer without changing the `WaveClip` boundary-related methods - and there's many of them:
* GetSequenceStartTime
* SetSequenceStartTime
* GetPlayStartTime
* SetPlayStartTime
* GetPlayEndTime
* SetPlayEndTime
* SetTrimLeft
* GetTrimLeft
* SetTrimRight
* GetTrimRight
* TrimLeft
* TrimRight
* TrimLeftTo
* TrimRightTo
* Offset
* WithinPlayRegion
* BeforePlayStartTime
* AfterPlayStartTime

I don't think I can give a better impression of what any alternative to the proposal in this PR would look like in terms of size. I, personally, would still prefer to have clips store their boundaries in beats. It might hurt now, but could bring good after.

Paul-Licameli:
We must choose between the pain of redundant data stored in each clip, with the problems of invalidation.

Or the pain of changing all of those functions to accept "Wave Clip Context", but then, boundaries become computed.

I still feel inclined toward the second pain.  Once it is done, "Wave clip Context" can also be enlarged to carry other information, and we won't need other publishers and listeners.

Paul-Licameli:
> We must choose between the pain of redundant data stored in each clip, with the problems of invalidation.
> 
> Or the pain of changing all of those functions to accept "Wave Clip Context", but then, boundaries become computed.
> 
> I still feel inclined toward the second pain. Once it is done, "Wave clip Context" can also be enlarged to carry other information, and we won't need other publishers and listeners.

Maybe the first will look like the lesser short-term pain, after all, but my intuition is that the second would be better for longer term maintainability.

saintmatthieu:
The context we need is `ProjectTimeSignature`, an attachment to the audacity project in `lib-numeric-format`. That dependency graph node and `lib-wave-track`, where this context need is seen for the first time, have no inter-dependency, and @Paul-Licameli is reluctant in establishing one.

DavidBailes:
> I think key to understand the problem is indeed that _one tempo change is all it takes_ to invalidate clip boundaries. _Beats_ remain constant, seconds don't.

At least some DAWs seem to have options for the timebase. eg:
https://steinberg.help/cubase_pro_artist/v9/en/cubase_nuendo/topics/track_handling/track_handling_defining_the_track_time_base_t.html
https://prorec.com/change-tempo-in-reaper-without-stretching-timebase/

Is the plan for Audacity not to have such options?

crsib:
As I understand, the primary concern of Paul is that we cannot store any references to AudacityProject in the clips due to copy/paste implementation details. This concern is very convincing :-)

Paul-Licameli:
> As I understand, the primary concern of Paul is that we cannot store any references to AudacityProject in the clips due to copy/paste implementation details. This concern is very convincing :-)

That was a concern in @saintmatthieu 's prototype efforts.  It's not my concern now, with this or the alternative proposals.

A context passed whenever some computed result is needed allows clip not to store a reference to the context object.ac

But as @saintmatthieu said I don't like accumulating unnecessary dependencies.

A context, which is not AudacityProject& or any object within it, can itself be computed as needed upstairs in src/, where wide dependencies are acceptable, and passed by value as a struct.

A problem is getting that context into all the right places must disturb much code.  But the alternative of pushing redundant copies of project tempo information into each clip whenever there is a change -- seems questionable too, increasing the problems of keeping project data consistent.

saintmatthieu:
>  the primary concern of Paul is that we cannot store any references to AudacityProject in the clips due to copy/paste implementation details

There is the possibility of passing the correct `ProjectTimeSignature` as argument to clip copy constructors, the way it already is done with `SampleBlockFactory`.

For the sake of completeness I'll add this option to the table.

crsib:
> A context, which is not AudacityProject& or any object within it, can itself be computed as needed upstairs in src/, where wide dependencies are acceptable, and passed by value as a struct.

I am strongly opposed to this solution. It will make Audacity fragile and hard to extend, it immediately makes a NEW non-UI feature in Audacity not available to Qt. 

The dependence of `WaveClip` (or rather of any `Track`) on tempo is explicitly specified in product design. It can be made implicit in code, indeed, but this sounds like an enormous source of errors more than anything else.

Previously, when the need for such context occurred, you pushed hard against it and I agreed with you. I don't honestly understand how is this case different.

crsib:
> You mean WaveClip storing boundaries (mTrimLeft, mSequenceOffset, etc.) as beats rather than seconds, implying tempo must be specified for these to become convertible to seconds and sample indices.

No. Why? I'm talking about introducing some external generic context with "all we need to calculate the boundaries, drawing, and everything". Like `TrackPanelDrawingContext`, which is so generic that it even has a `void*` "user data" field! Does it remove the dependency on tempo? It does not.

> Could you give a concrete example?

Playback will be lifted from the `src` tree, and drawing won't. Where should we fill this "externalized" context? What should we do when the context extends? And we anticipate that it will extend. Otherwise, we could just pass "tempo" value. But we have an example with effects already where we cannot have beats display because initially, we pass only the sample rate instead of a more general context.

Paul-Licameli:
> > A context, which is not AudacityProject& or any object within it, can itself be computed as needed upstairs in src/, where wide dependencies are acceptable, and passed by value as a struct.
> 
> I am strongly opposed to this solution. It will make Audacity fragile and hard to extend, it immediately makes a NEW non-UI feature in Audacity not available to Qt.
> 

This objection is not strong.  It may only mean that the place where the context is built is not yet extracted into a higher level library.  The library extraction effort is paused to get all this other work done.

I would like to avoid lib-wave-track depending on more things.  What if I concede that?  Then AudacityProject& is enough context.

> The dependence of `WaveClip` (or rather of any `Track`) on tempo is explicitly specified in product design.

Design specifies desired program behavior.  It does not specify any such implementation choice.

> It can be made implicit in code, indeed, but this sounds like an enormous source of errors more than anything else.
> 
> Previously, when the need for such context occurred, you pushed hard against it and I agreed with you. I don't honestly understand how is this case different.

Remind me what you refer to.

Anyway if context is not used, and a “push” of tempo info into each clip really must be done, then a roundabout way of doing it as in the closed PR will be needed.

But I can’t approve this PR without seeing and understanding more of the whole, and whether the alternative really makes more sense.

crsib:
> Remind me what you refer to.

I refer to #4414

> I would like to avoid lib-wave-track depending on more things

It explicitly depends on `lib-project-rate-interface` already. I understand your concern about `lib-numeric-format`, but I can easily extract ProjectTimeSignature into another library. I don't expect that it will take more than 30 minutes.

> But I can’t approve this PR without seeing

Nor do I. We are very aligned on this.

> It does not specify any such implementation choice.

Behavior specifies an explicit dependency on tempo. It can be hidden by introducing a transfer object of some kind (call it Context), but it does not remove this dependency. Failing to correctly specify the tempo will result in an error.

saintmatthieu:
> I'm talking about introducing some external generic context

Oh, ok. My concern for now is limited to tempo context. I still am not 100% sure that this is also something you see a problem with. You mention real-time effects, but I'm not too concerned that these need project tempo information.

crsib:
> but I'm not too concerned that these need project tempo information.

They do, regardless of the time stretch. We cannot allow using bars/beats for the duration in generators because they cannot access the time signature and tempo. There are plans, currently on hold, to fix that.

Paul-Licameli:
> > but I'm not too concerned that these need project tempo information.
> 
> They do, regardless of the time stretch. We cannot allow using bars/beats for the duration in generators because they cannot access the time signature and tempo. There are plans, currently on hold, to fix that.

Do you refer to my attempts to define EffectContext?

The recently reviewed RAII object for temporary effect output tracks was spun off from that other effort to remove state from EffectBase.

crsib:
> Do you refer to my attempts to define EffectContext?

Exactly!

Paul-Licameli:
> > the primary concern of Paul is that we cannot store any references to AudacityProject in the clips due to copy/paste implementation details
> 
> There is the possibility of passing the correct `ProjectTimeSignature` as argument to clip copy constructors, the way it already is done with `SampleBlockFactory`.
> 
> For the sake of completeness I'll add this option to the table.

Maybe that will after all look like the least difficult thing, if more dependency from lib-wave-track is allowed.

But consider what I asked elsewhere about maybe under specified behavior of other kinds of tracks.

Paul-Licameli:
> not too concerned

Are you two thinking of the same thing?  I think Matthieu refers to the playback, not the ui, and for that we have the stretching decorator encapsulating the time stretch calculations.

crsib:
> But consider what I asked elsewhere about maybe under specified behavior of other kinds of tracks.

Absolutely, let's raise it at the meeting today!

Paul-Licameli:
> > But consider what I asked elsewhere about maybe under specified behavior of other kinds of tracks.
> 
> Absolutely, let's raise it at the meeting today!

I will be climbing a mountain now that there’s finally some good weather here

saintmatthieu:
Trying out the beat approach. It turns out to be a very healthy exercise, insofar as it reveals that some methods now make little sense ...
<br/>

Time values cannot index to specific positions in tracks or clips anymore without a tempo specified. It'd be like asking when a given sample must be played out without knowing the speed of the cursor.

For clips, we still can have methods like `Get(sampleCount start, size_t length)`, since the left-trimmed part is actually a fraction of the raw data. We ignore that fraction of the raw samples, and start from there, reading unstretched, raw data.

Tracks are the arrangement of clips onto a timeline with beat units. As such, they _must_ return stretched data : it wouldn't make sense for the UI to show clip A lasting 1s, and then having it produce 2s worth of samples. `Get(sampleCount start, size_t len, ...` and the likes should therefore be banned from non-stretching-capable tracks.

So, `WaveTrack::Get` : gone. (Or `assert(false)` for now if it ever gets used.) `StretchingSequence` can stretch, so that's fine. The other is `CachingPlayableSequence` : I think this whole class should be gone. Without its  `Get` method it serves little purpose. The spectrum-display code should query the raw data on the clips directly (and stretch the display), using the `GetSampleView` method the way `CachingPlayableSequence` has been doing to ensure caching works.

And then come destructive effects...

On the one end, there's a `WideSampleSource`, calling `WaveTrack::Get`, and on the other, a `WideSampleSink`, calling `WaveTrack::Set`, with a per-track effect instance in between. I've argued already that `Get` should go. `Set` should go for the same reasons. What I think makes most sense instead is to have those source and sink classes handle clips directly too, in their order of appearance on the timeline, read _all_ samples (including those trim-hidden), process them, and write them back. Silence is ignored in between, leading to clip A possibly ringing into clip B, but ringing already happens in a not very convincing way UX-wise.

<br/>
I'm not finished yet with that approach. Let's hope there are no more hurdles for it to reveal. Paul pointed out the label-track case today, which also are arrangement of objects on a timeline now with beat units...

Paul-Licameli:
> Trying out the beat approach. It turns out to be a very healthy exercise, insofar as it reveals that some methods now make little sense ...
> 
> Time values cannot index to specific positions in tracks or clips anymore without a tempo specified. It'd be like asking when a given sample must be played out without knowing the speed of the cursor.
> 
> For clips, we still can have methods like `Get(sampleCount start, size_t length)`, since the left-trimmed part is actually a fraction of the raw data. We ignore that fraction of the raw samples, and start from there, reading unstretched, raw data.
> 
> Tracks are the arrangement of clips onto a timeline with beat units. As such, they _must_ return stretched data : it wouldn't make sense for the UI to show clip A lasting 1s, and then having it produce 2s worth of samples. `Get(sampleCount start, size_t len, ...` and the likes should therefore be banned from non-stretching-capable tracks.
> 
> So, `WaveTrack::Get` : gone. (Or `assert(false)` for now if it ever gets used.) `StretchingSequence` can stretch, so that's fine. The other is `CachingPlayableSequence` : I think this whole class should be gone. Without its `Get` method it serves little purpose. The spectrum-display code should query the raw data on the clips directly (and stretch the display), using the `GetSampleView` method the way `CachingPlayableSequence` has been doing to ensure caching works.
> 
> And then come destructive effects...
> 
> On the one end, there's a `WideSampleSource`, calling `WaveTrack::Get`, and on the other, a `WideSampleSink`, calling `WaveTrack::Set`, with a per-track effect instance in between. I've argued already that `Get` should go. `Set` should go for the same reasons. What I think makes most sense instead is to have those source and sink classes handle clips directly too, in their order of appearance on the timeline, read _all_ samples (including those trim-hidden), process them, and write them back. Silence is ignored in between, leading to clip A possibly ringing into clip B, but ringing already happens in a not very convincing way UX-wise.
> 
> 
> I'm not finished yet with that experiment. Let's hope there are no more hurdles for it to reveal. Paul pointed out the label-track case today, which also are arrangement of objects on a timeline now with beat units, though...

Eliminating WaveTrack::Get and WaveTrack::Set, and making Mixer aware of clip boundaries, are radical ideas that I dont like.  I need persuading.

Paul-Licameli:
What are you assuming about conservation of clip structure when rendering or applying a destructive effect?  What was specified by the design team?

Mix and render has never preserved it.  Existing destructive effects do not do it consistently.

If that change is not your idea alone but the design teams, I am very unhappy to learn about it now.

I am thinking of implications and I don’t like them.

But WaveTrackSink is not even used for all destructive effects, only those that inherit PerTrackEffect.

I oppose changes for rendering too.  Rendering for export should be oblivious to clips, but just take streams of samples to mix into an audio file, which doesn’t have clips, and that means keeping the simple abstraction of sources.

Mixing and rendering to a track should not behave differently. It should give you the as-if-exported sound but still in a track.  Clip boundaries are lost.

Paul-Licameli:
I have a growing suspicion that we have fundamentally erred in assuming the time stretch should apply upstream of stacked effects.  It seems to be creating too many problems.  Doing it downstream might cut through some problems, but raise others.

Paul-Licameli:
> I have a growing suspicion that we have fundamentally erred in assuming the time stretch should apply upstream of stacked effects. It seems to be creating too many problems. Doing it downstream might cut through some problems, but raise others.

Paul-Licameli:
> I have a growing suspicion that we have fundamentally erred in assuming ALL time stretch should apply upstream of stacked effects. It seems to be creating too many problems. Doing it downstream might cut through some problems, but raise o

Paul-Licameli:
> I have a growing suspicion that we have fundamentally erred in assuming all time stretch should apply upstream of stacked effects. It seems to be creating too many problems. Doing tempo stretch downstream, for drawing and playback and rendering, might cut through some problems. The stretch determined by clip handles could be a separate upstream stretch.
> 
> Yes, drawing of clips may need to avoid WaveTrack::Get, fetching straight from clips and taking the handle drag into account.
> 
> But now I think tempo should be a separate thing, also in drawing context, available in drawing routines of all track types.
> 
> Mixing for export should also apply the tempo stretch as the last processing stage. Mixing and rendering would omit that, so that the rendered track, as played, 
> 
> Time stretch for tempo would need to be done in AudioIO.
> 
> (Do we also want dynamic variation of tempo during playback?)
> 
> The problems of Get and Set in PerTrackEffect may be unimportant because the sources and sinks are really sequential, not random access.
> 
> WaveTrack::Set needs to exist still to implement the pencil tool. You zoom in until you see samples (though now some stretch is applied in drawing them), and you edit them directly.
> 
> Tempo stretch seems to me to want to be a high level project wide thing that is applied in the right contexts, and can be kept out of WaveClip or any other track subtype.

saintmatthieu:
> What are you assuming about conservation of clip structure when rendering or applying a destructive effect? What was specified by the design team?

Behavior relating to clip preservation would change in neither mix-and-rendering, export nor destructive effects. `WideSequenceSource` and `WideSequenceSink` would now deal with clips directly (or `ClipInterface`, or whichever few methods they need from clips) and not with a `WideSampleSequence`. Their implementation wouldn't need to know where clips are on the UI representation, but could just process them one after the other, ignoring silence in between. This process would change the _contents_ of the clips, but not affect their _boundaries_ on tracks.

saintmatthieu:
> I have a growing suspicion that we have fundamentally erred in assuming all time stretch should apply upstream of stacked effects.

In the case of destructive effects, no stretching comes in to play.

crsib:
> This process would change the contents of the clips, but not affect their boundaries on tracks.

mix-and-render changes the clip boundaries by design (effectively by creating a single clip, dropping attached effects etc), I don't think that it is wise to change this behavior. Probably I don't understand you correctly. FWTIW playback, rendering, and export could be viewed as the same process, but with different sinks, and for mix&render the sink is just a track. 

> In the case of destructive effects, no stretching comes in to play.

I am a bit confused. I don't think that there are commutative properties in effect application. Wouldn't it make a destructive application result to be very different in some cases?

saintmatthieu:
> mix-and-render changes the clip boundaries by design

By destructive effects I've been meaning selecting a track, choosing an effect and applying it. This preserves clip boundaries now, and it still would with my suggestion.

saintmatthieu:
> I am a bit confused. I don't think that there are commutative properties in effect application. Wouldn't it make a destructive application result to be very different in some cases?

Say you have a clip and stretch it by a ratio of 2. Now apply an effect. There are two possibilities:
1. Render an un-stretched clip : apply the effect a it is now, on the raw data. The result is still a clip with stretch ratio metadata 2.
2. Render a stretched clip : apply the effect and stretch (or the other way round). The result is a clip with stretch ratio metadata 1.

I have been suggesting 1. It is more conservative : stretch a clip by 2, apply gain, unstretch : it's as though you had never stretched. Also, the sources and sinks used wouldn't need to have time-stretchers injected.

Does that un-confuse you?

Paul-Licameli:
> > What are you assuming about conservation of clip structure when rendering or applying a destructive effect? What was specified by the design team?
> 
> Behavior relating to clip preservation would change in neither mix-and-rendering, export nor destructive effects. `WideSequenceSource` and `WideSequenceSink` would now deal with clips directly (or `ClipInterface`, or whichever few methods they need from clips) and not with a `WideSampleSequence`. Their implementation wouldn't need to know where clips are on the UI representation, but could just process them one after the other, ignoring silence in between. This process would change the _contents_ of the clips, but not affect their _boundaries_ on tracks.

Design team’s requirements or your ideas?

I suspect a lot of surprising consequences are not well thought out, such as the overlapping of clips, and restructuring every destructive effect to work this way is a bad surprise implying a lot of work no one budgeted.

Paul-Licameli:
I think we need to separate clip stretching from project tempo, and do one, only one, of them well in this release.

I have a growing alarm about overreach.

I think tempo is the more valued thing, but my efforts have all been to support clip stretching.

I think downstream application of tempo stretch is the right thing for various reasons, but the alternative was not given proper early consideration.

Paul-Licameli:
And doing it downstream means messing with the data communicated between threads in AudioUO, and that is the large problem zu should be figuring out.  (And fixing the outstanding latency problems with it.)

saintmatthieu:
> Design team’s requirements or your ideas?

I don't see how this could be Design's idea. The behavior doesn't change, whether with stretched clips or not.

Paul-Licameli:
So it’s not a requirement, it’s your own idea, but you are incorrect that separately applied effects on each clip would give equivalent results for all effects.

What does Reverb do it you select a clip and some empty space after it?

saintmatthieu:
This:
image

saintmatthieu:
... but that with delay:
image
That's new to me. I'm not too concerned that the source now has to know how much silence there is between each clip, but it also makes me realize that the parameters of timing-based effects (such as this one, a delay) must account for the stretch ratio. That'd need to be done on each timing-related effect one by one, would only (?) be a parameter change, but it may have other consequences. The alternative, indeed, is to do have sink and source apply time stretching - but first! Not after the effect was applied, or the effect's timing is messed up just as much. Couldn't that work with a `StretchingSequence` ?

Paul-Licameli:
> This:
> image

Try it again, with a silent clip touching the other clip on the right.

Paul-Licameli:
But I don’t like the ideas I’m hearing.

Im increasingly thinking the entire project needs a rethink.

My refactorings for wide wave clips and tracks are not wasted work, and the new caching was not either, but I have doubts now whether the stretching decorator is right.

Downstream application of time stretch seems better to me.  Why wasn’t it considered?  Maybe because doing it upstream was easier to figure out for a convincing prototype.  Doing it down may need other less obvious work but the alternative should be considered.

saintmatthieu:
> Try it again, with a silent clip touching the other clip on the right.

The reverb of A leaks into B. So would it with the clip-by-clip approach I propose, as well as with a `StretchingSequence` used by the source.

crsib:
I'm not sure that a follow the terminology, but I think that the stretching should happen before any other effect. Probably I'm wrong, @Tantacrul @LWinterberg can you please help with this?

saintmatthieu:
> I think that the stretching should happen before any other effect

I'd say so too. Timing-related effects would need to anticipate stretching to produce the wanted result. Take slap-back echo, for example. Apply it first, and stretch after, and you've stretched the slap-back time. Or the effect would have to anticipate stretching, sounds ugly.

saintmatthieu:
> I am a bit confused. I don't think that there are commutative properties in effect application. Wouldn't it make a destructive application result to be very different in some cases?

@crsib you were one step ahead of me. Later in this conversation I was writing
> The alternative, indeed, is to do have the source apply time stretching - not the sink! Or the effect's timing is messed up just as much.

Sounds like a rule of thumb.

Tantacrul:
I can't imagine applying time stretching after effects. As @crsib and @saintmatthieu say, it seems sensible for it to be first in the chain.

saintmatthieu:
> I am a bit confused. I don't think that there are commutative properties in effect application. Wouldn't it make a destructive application result to be very different in some cases?

@crsib you were one step ahead of me. Later in this conversation I was writing
> The alternative, indeed, is to do have the source apply time stretching - not the sink! Or the effect's timing is messed up just as much.

Sounds like a rule of thumb.

saintmatthieu:
To be complete, possible would also be to time-stretch downstream of `WaveTrack`, with a time-stretcher which changes its ratio on the fly. The effects would still come after, and we wouldn't have had the joint-stereo problem.
I still feel better on a clip-based stretching instance, though, and have the guarantee that timing at the start of a clip doesn't depend on the stretcher's output until then.

Paul-Licameli:
> I can't imagine applying time stretching after effects. As @crsib and @saintmatthieu say, it seems sensible for it to be first in the chain.

One argument for applying tempo stretch last:

If you export, time stretch is applied in the resulting audio file.

If you mix and render, you should get a track that plays back exactly as the exported audio.  That means:  we should precompute everything except the tempo stretch.

This may not be possible if the stretch applies first in export.

Paul-Licameli:
> > Try it again, with a silent clip touching the other clip on the right.
> 
> The reverb of A leaks into B. So would it with the clip-by-clip approach I propose, as well as with a `StretchingSequence` used by the source.

Then I do not understand the proposal.

crsib:
> If you export, time stretch is applied in the resulting audio file.

Is it what happens now? Then I would call it bug and fix it.

Paul-Licameli:
> > If you export, time stretch is applied in the resulting audio file.
> 
> Is it what happens now? Then I would call it bug and fix it.

I mean I assume it would happen and be a feature.  Whatever you play in Audacity should be what
goes into the export.  We do that for the stacked effects, and would do it too for stretch.

Now honestly let me counter argue myself:  If tempo stretch applies before other stacked effects, then a solution is that mix-and-render also applies the inverse of project tempo stretch as the last stage of transformation before re-storing the samples.

Then playback inserts tempo stretching upstream, which cancels this.

Paul-Licameli:
> > > If you export, time stretch is applied in the resulting audio file.
> > 
> > 
> > Is it what happens now? Then I would call it bug and fix it.
> 
> I mean I assume it would happen and be a feature. Whatever you play in Audacity should be what goes into the export. We do that for the stacked effects, and would do it too for stretch.
> 
> Now honestly let me counter argue myself: If tempo stretch applies before other stacked effects, then a solution is that mix-and-render also applies the inverse of project tempo stretch as the last stage of transformation before re-storing the samples.
> 
> Then playback inserts tempo stretching upstream, which cancels this.

But the counter counter is this:  Are we really sure that applying stretch by ratio r, then again by 1/r, is really an identity transformation in this implementation?

crsib:
I'm not sure I follow. Mix-and-render, similar to playback, similar to export, applies time stretch first. However the resulting track (and it's single clip) is not time stretched. Why do we need to do the reverse transform?

Yes, it will become very complicated when we allow to set the tempo curve, but right now this is not the case.

Paul-Licameli:
> I'm not sure I follow. Mix-and-render, similar to playback, similar to export, applies time stretch first. However the resulting track (and it's single clip) is not time stretched. Why do we need to do the reverse transform?
> 
> Yes, it will become very complicated when we allow to set the tempo curve, but right now this is not the case.

Isn’t the project tempo stretch applied to all tracks?  Then if rendered tracks, being in the same project, have had the tempo stretched baked into the data, the playback will insert the tempo stretch again unless somehow the rendered track cancels it.

Paul-Licameli:
On the other other hand… A solution can be that every clip stores what the project tempo was at the time it was created.  The stretch ratio depends on that and the current project tempo.

Or was that the idea all along?  I was foolish then.

Paul-Licameli:
Whatever we do decide, stretching before or after effects — I want to know that designers considered the alternatives and really have a strong preference, or are decidedly neutral about it and will leave it to development to choose an implementation.

It seems it is not so:  that Martin is only now answering the question, and Matthieu did not consider alternatives or is acting on assumptions about which is better, that I’m not persuaded of.

If design is neutral, then I dislike Matthieu’s proposals as introducing unnecessary complexities we might avoid.

If design is not neutral, and at least some of these complexities are unavoidable, very well, we must accept them and make things work.

crsib:
I think Martin has strongly asserted that time stretching should be applied first. I don't think that it makes sense to do it other way around. I feel that this oversight is similar to one with effect settings window modality. It was assumed to be obvious so it never was explicitly stated. 

For sample tracks, the amount of stretch is a property of a clip. Within a track clips
can have different stretching coefficients.

Paul-Licameli:
> I think Martin has strongly asserted that time stretching should be applied first. I don't think that it makes sense to do it other way around. I feel that this oversight is similar to one with effect settings window modality. It was assumed to be obvious so it never was explicitly stated.
> 
> For sample tracks, the amount of stretch is a property of a clip. Within a track clips can have different stretching coefficients.

Maybe it’s obvious to others.  But I’d like Martin to speak for himself.  I may not understand the use cases.

What I do know is that if project tempo can be applied after all else, then it can cut through a lot of implementation difficulty.  (Ignore what I said about render and export: I now think  that can work with either choice.)

This is not the same problem as stretching each clip by a different adjusted ratio.

And I want to ask again, whether we should commit to doing only one of these in this release — either tempo adjustment (which was the purpose stated in this PR title) or clip adjustment (which is not) — in case we can’t solve all the difficulties of both on time.

Tantacrul:
To avoid the risk of committing to a course of action without fully agreeing (because I, for one might miss something), I'd suggest we have a little chat early next week.

When talking about real-time effects, it's not very valuable to apply TS at the end of the effects chain. This is a production thing.

If someone here is discussing destructive effects, then I worry there's some use case I've failed to consider.

Paul-Licameli:
> To avoid the risk of committing to a course of action without fully agreeing (because I, for one might miss something), I'd suggest we have a little chat early next week.
> 
> When talking about real-time effects, it's not very valuable to apply TS at the end of the effects chain. This is a production thing.
> 
> If someone here is discussing destructive effects, then I worry there's some use case (or technical weirdness in Audacity) I've failed to consider.

So, destructive effects were not a topic in those design meetings yet that I skipped?

I want so say:  identify known problems with timing dependent destructive effects, to fix in next release.

There is worrying scope creep getting discussed here, here dissipating our limited efforts.  Get all the other fundamentals done.  Leave destructives alone to behave however they will for this release.

Tantacrul:
Let's chat. The last thing I want to do is accidentally give the impression that some large chunk of work now needs to be done.

saintmatthieu:
Time-stretching a clip by a factor of 2 and then again by a factor of 2 yields a result different from time-stretching that clip once by a factor of 4. Likewise, @Paul-Licameli, since you asked: stretching by 2 and then by 0.5 doesn't restore the original. Quality suffers from each pass. Our time-stretcher may be excellent, it's not perfect : there still is some amount of widening and smearing.

Therefore, the approach we're undertaking is the best from a quality point of view. Each clip's stretch ratio is the product of its individual stretch ratio (modifiable with the handles) and of the ratio `originalTempo / projectTempo`, allowing only just one time-stretching pass.

@Paul-Licameli this PR is only just about bringing the project tempo information to the clips. Their own, "handle" stretch ratio is easy to get - that's the ticket we planned for Vitaly.

saintmatthieu:
What is `originalTempo`? @Paul-Licameli you guessed right : it's the project tempo when the clip is dropped. It is on the roadmap (at least informally) to improve this by reading tempo metadata or detecting is by analyzing the signal.

Paul-Licameli:
> Time-stretching a clip by a factor of 2 and then again by a factor of 2 yields a result different from time-stretching that clip once by a factor of 4. Likewise, @Paul-Licameli, since you asked: stretching by 2 and then by 0.5 doesn't restore the original. Quality suffers from each pass. Our time-stretcher may be excellent, it's not perfect : there still is some amount of widening and smearing.
> 
> Therefore, the approach we're undertaking is the best from a quality point of view. Each clip's stretch ratio is the product of its individual stretch ratio (modifiable with the handles) and of the ratio `originalTempo / projectTempo`, allowing only just one time-stretching pass.
> 
> @Paul-Licameli this PR is only just about bringing the project tempo information to the clips. Their own, "handle" stretch ratio is easy to get - that's the ticket we planned for Vitaly.

Yes, I didn’t raise the question of repeated stretching causing degraded quality but I thought of it.

Stretching for project tempo and for handles should be adjacent in the stack of processing, whether upstream or down, and combined into one.

I was thinking about how to implement it all downstream, very differently from the stretching decorator.

But if that really is a non-starter from a design standpoint , okay forget it.

Reopen this pull request, under a new number so we can restart the conversation thread.  Put more in it if you can — enough that I see what new data a WaveClip really stores and updates.

Paul-Licameli:
About the problems of destructive effects:

I emphasize again, do not attempt a solution as you outlined it.  I think too many individual built in effects will need too much rewriting.  It will distract from more fundamental problems of this feature that are still not moving through review and test.  It will likely conflict with many of the other rewrites I’m trying to do to support joint stereo.

But I did have an idea for an alternative solution that I want to explain…

Paul-Licameli:
Whenever an effect or other editing operation must paste into a WaveTrack, map the given time to the correct clip.  Then first transform that clip with time stretch, and reset its stored stretch information to unit stretch ratio.  Then proceed as in old code.

I think this way would be more local in code modifications of class WaveTrack, will not disturb its interface, and will leave much other code undisturbed.

Do similar with Set() except for the pencil tool.

Issue #4848, opened 2023-06-28
Audacity wrongly complaining about stereo track with different sample rates


saintmatthieu:
### Steps to reproduce

1. create two mono empty tracks
2. pan one track hard left, the other hard right
3. mix and render to a new track
4. save, close and re-open project

### Expected behavior

the project re-opens

### Actual behavior

An error modal pops up, stating that "This project cannot be opened because it contains a stereo track with different sample rates."

### Audacity Version

other

### Operating system

Windows

### Additional context

The issue first appears in [this build](https://github.com/audacity/audacity/actions/runs/5379916314).

saintmatthieu:
I have a similar problem opening [this project](https://1drv.ms/u/s!ArATSZ9esLBGxizp6yeMBa1CAB26?e=HN2oWh), which was created just by duplicating a mono track. It may be related.

dozzzzer:
Tested

PR #4855, opened 2023-06-28
Fix stereo track persistency

Paul-Licameli:
Resolves: #4808
Resolves: #4848

Recommended:


saintmatthieu:
This seems to fix https://github.com/audacity/audacity/issues/4848 ; was this the intention?

saintmatthieu (approved):

Issue #4860, opened 2023-06-29
Error when pressing Record for the second time with WASAPI set as a host


dozzzzer:
### Steps to reproduce

1. select Host: WASAPI
2. Press Record
3. Stop recording
4. Press Record again
5. Observe

### Expected behavior

No errors, recording continues

### Actual behavior

![image](https://github.com/audacity/audacity/assets/70063970/6d3e4530-5d18-4230-8881-ca1fcda4b5e9)

### Audacity Version

current master / alpha / nightly build

### Operating system

Other

### Additional context

A regression introduced by https://github.com/audacity/audacity/commit/35a0869c240631d235f05c0063f340ba3d1feee8

petersampsonaudacity:
Testing on W10 with latest 3.4.0 alpha:  audacity-win-3.4.0-alpha-20230627+86de8e9-x64-msvc2022

I cannot reproduce this.

I tested with all 3 Recording devices offered to me with WASAPI as the host:
![image](https://github.com/audacity/audacity/assets/50205984/4ee88ad5-7b54-41e9-8ff8-9cd499bac7fa)

saintmatthieu:
@petersampsonaudacity it could be due to not having overdub enabled. I found this played a role while debugging the issue.

petersampsonaudacity:
@saintmatthieu  interesting, I can't test now as my pc is packed away and I'm off to Egypt very shortly.

I would have had overdub on, as it is on by default.

Personally I have never understood why that is so. I campaigned some years ago to have it turned off by default, but I was outvoted.  I still think it should be off by default as clearly not evevery users wants or needs overdub (in 20 years of Audacity use I have never used it, except for QA testing).

saintmatthieu:
If not overdub ... then it could be that your recording and playout devices have equal sample rate. You could also try changing one of these if you wish. But if I were in Egypt I'd rather go visit the pyramids :D

crsib:
I think Gonzalo has made an almost-finished PR to address this issue https://github.com/audacity/audacity/pull/4353

saintmatthieu:
Solved with https://github.com/audacity/audacity/pull/4861

petersampsonaudacity:
> But if I were in Egypt I'd rather go visit the pyramids :D

A nice idea, BUT...  Hurghada on the Red Sea coast is a long way from Cairo - and a dive boat on the Red sea makes the trip even harder.

But I have seen the pyramids twice in my life already - and mighty impressive they are  :-))

PR #4861, opened 2023-06-29
Hot fix for broken realtime effects

### LINKED ISSUE #4859: Non-op realtime effects (incl. the built-ins)
### Bug description

None of the real-time effects families are working, including the built-in ones.

### Steps to reproduce

1. Have some audio clip
2. Select some effects in the stack
3. Start playback

### Expected behavior

The effects changing the sound correspondingly

### Actual behavior

No changes to the sound, as if they were bypassed.

### Audacity Version

### Operating system

Other

### Additional context

A regression introduced by https://github.com/audacity/audacity/commit/35a0869c240631d235f05c0063f340ba3d1feee8

saintmatthieu:
Resolves: #4859 

Realtime effects work by AudioIO getting the realtime-effect list attachment onto its playback sequences.

Problem : those are now injected, attachment-less `CachingPlayableSequence`s, which decorate the `WaveTrack`.

Quick fix : we give `WideSampleSequence` a method to return it's decorated version, which can then be used by AudioIO.

Longer term fix (maybe) : this is another incarnation of the problem we had with `GetMute/GetSolo`, also implemented as `WaveTrack` attachments. The solution was to find the right place for an abstract API, which internally is implemented by `WaveTrack` through query of its attachment. Now it's an abstraction for `RealtimeEffectList` that is needed, at least the part of it needed by AudioIO to apply real-time effects.

Recommended:

### DIFF
--- a/libraries/lib-mixer/WideSampleSequence.cpp
+++ b/libraries/lib-mixer/WideSampleSequence.cpp
@@ -21,3 +21,16 @@ double WideSampleSequence::LongSamplesToTime(sampleCount pos) const
 {
    return pos.as_double() / GetRate();
 }
+
+const WideSampleSequence& WideSampleSequence::GetDecorated() const
+{
+   const WideSampleSequence* innermost = this;
+   while (const auto newP = innermost->DoGetDecorated())
+      innermost = newP;
+   return *innermost;
+}
+
+const WideSampleSequence* WideSampleSequence::DoGetDecorated() const
+{
+   return nullptr;
+}
--- a/libraries/lib-mixer/WideSampleSequence.h
+++ b/libraries/lib-mixer/WideSampleSequence.h
@@ -15,7 +15,7 @@ Paul Licameli split from SampleFrame.h
 #include "SampleCount.h"
 #include "SampleFormat.h"
 
-struct WideSampleSequence;
+class WideSampleSequence;
 
 //! Hosting of objects attached by higher level code
 using SequenceAttachments = ClientData::Site<
@@ -33,6 +33,8 @@ public:
    using Attachments = SequenceAttachments;
 
    virtual ~WideSampleSequence();
+
+   const WideSampleSequence& GetDecorated() const;
 
    //! A constant property
    /*!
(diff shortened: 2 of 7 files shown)


crsib (approved):
I can approve it as a hotfix, but I am really confused with the approach, where decorator masks vital parts of the decorated objects. I can see it as a source of a lot of bugs in the future.

petersampsonaudacity:
Testing on W10 with @sahilsekr42 's branch build:  audacity-win-3.4.0-alpha-20230629+0feda0a-x64

I confirm that the realtime effects now work again with this branch build.

But I uncovered a further associated regression when adding a further realtime effect while the audio is playing - I will add this as a new issue.

saintmatthieu:
Follow-up problems described in https://github.com/audacity/audacity/issues/4864 and https://github.com/audacity/audacity/issues/4865, as well as the broken-recording problem https://github.com/audacity/audacity/issues/4860 have a similar root cause. Since this is a small PR, I choose to fix them all here.

crsib (commented):
How long would it take to do a proper fix, that doesn't expose `GetDecorated` in all the places around the code?

saintmatthieu:
I haven't come up with a good solution. If I could find a way of getting compilation errors wherever `WideSampleSequence` pointers are compared that would count as a better fix.

Alternatively letting `WaveTrack` own the sample views for caching, instead of using a `CachingPlayableSequence` for that, that wouldn't be too bad. The stretching decorator, though, is more complex, and it's nice to keep that logic separate.

Or don't inject `CachingPlayableSequence` _in place of_ `WaveTrack` but _besides_.

saintmatthieu:
Added @Paul-Licameli to this review to contribute to this discussion.

Paul-Licameli:
I'm watching this PR and I have no strong objections.

I don't like disturbing the WideSampleSequence interface, but something like this is really needed.

I may have ideas about making it more elegant, but I'll propose my own PR after all the bugs are found and fixed.

saintmatthieu:
@dozzzzer ready for merge ?

saintmatthieu:
@petersampsonaudacity thank you!

Issue #4847, opened 2023-06-27
3.3.3 FTBFS with gettext-0.22


jlocash2:
When building with gettext-0.22 2 the be and fa locales fail to build.
The relavent errors:

```
[ 45%] Generating ../Release/locale/be/LC_MESSAGES/audacity.mo
/tmp/build/audacity-sources-3.3.3/locale/be.po:4472: 'msgstr' is not a valid C format string, unlike 'msgid'. Reason: In the directive number 1, the argument size specifier is invalid.
/usr/bin/msgfmt: found 1 fatal error
make[2]: *** [locale/CMakeFiles/locale.dir/build.make:140: Release/locale/be/LC_MESSAGES/audacity.mo] Error 1
```

```
[ 45%] Generating ../Release/locale/fa/LC_MESSAGES/audacity.mo
/tmp/build/audacity-sources-3.3.3/locale/fa.po:4416: 'msgstr' is not a valid C format string, unlike 'msgid'. Reason: In the directive number 1, the argument size specifier is invalid.
/usr/bin/msgfmt: found 1 fatal error
make[2]: *** [locale/CMakeFiles/locale.dir/build.make:215: Release/locale/fa/LC_MESSAGES/audacity.mo] Error 1

```

crsib:
@dozzzzer please check; this problem now affects CI

dozzzzer:
Tested in PR #4866:

- generators, destructive effects
- cloud uploads
- recording, playback
- split, copy/paste, trimming clips
- sorting tracks, renaming tracks, resizing tracks
- beats & measures ruler mode
- merging mono tracks into a stereo track
- mixing-and-rendering, import/export

PR #4866, opened 2023-06-30
Fix translations (be, fa) for gettext 0.22 (#4847)

SoapGentoo:
Bug: https://bugs.gentoo.org/909201

Resolves: #4847

gettext 0.22 is more pedantic with validating the destination string and ensuring that it represents a valid format string.

Recommended:


crsib (approved):

SoapGentoo:
@crsib I went ahead and replaced all occurrences of `%hs`, since they're invalid under C99

crsib:
Thank you for your contribution!

PR #4867, opened 2023-07-03
Have clips know project tempo

### LINKED ISSUE #4849: Have clips know project tempo

saintmatthieu:
Resolves: #4849 

Propose a publisher of project tempo changes. `Track` has an abstract method for `WaveTrack`, `LabelTrack`, `NoteTrack` and `TimeTrack` to implement.
Add the implementation in `WaveTrack`. `WaveClipList` is proposed as wrapper of `WaveTrack`'s `mClipList` to make sure that newly-added clips get the most up-to-date project tempo information.

This PR is a fresh start for https://github.com/audacity/audacity/pull/4853.

Recommended:

Things to test:
- [x] Project tempo change now affects all tracks except the audio tracks. For non-audio tracks, changing the project tempo should result in the track's horizontal scaling staying in sync with the beats-and-measure ruler. 
- [x] Audio tracks should still stay in sync with the time ruler
- [x] Change project tempo for all track types, and make sure that this is correct.
- [x] Copy/paste between the projects
- [x] Undo/Redo
- [x] Save, close and reopen projects should work

... and everything else you deem worthwhile testing of course.

### DIFF
--- a/libraries/lib-mixer/Envelope.cpp
+++ b/libraries/lib-mixer/Envelope.cpp
@@ -29,6 +29,7 @@ a draggable point type.
 
 #include "Envelope.h"
 
+#include <float.h>
 #include <math.h>
 
 #include <wx/wxcrtvararg.h>
@@ -481,7 +482,7 @@ void Envelope::PasteEnvelope( double t0, const Envelope *e, double sampleDur )
       return;
    }
 
-   // Make t0 relative to the offset of the envelope we are pasting into, 
+   // Make t0 relative to the offset of the envelope we are pasting into,
    // and trim it to the domain of this
    t0 = std::min( mTrackLen, std::max( 0.0, t0 - mOffset ) );
 
@@ -637,7 +638,7 @@ std::pair< int, int > Envelope::ExpandRegion
    }
 
    mTrackLen += tlen;
-   
+
    // Preserve the right-side limit.
    if ( index < range.second )
       // There was a control point already.
@@ -822,6 +823,14 @@ void Envelope::RescaleTimes( double newLength )
          point.SetT( point.GetT() * ratio );
    }
    mTrackLen = newLength;
+}
+
+void Envelope::RescaleTimesBy(double ratio)
+{
+   for (auto& point : mEnv)
+      point.SetT(point.GetT() * ratio);
+   if (mTrackLen != DBL_MAX)
+      mTrackLen *= ratio;
 }
 
 // Accessors
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -80,7 +80,8 @@ Track::Track()
    mOffset = 0.0;
 }
 
-Track::Track(const Track &orig, ProtectedCreationArg&&)
+Track::Track(const Track& orig, ProtectedCreationArg&&)
+    : mProjectTempo { orig.mProjectTempo }
 {
    mIndex = 0;
    mOffset = orig.mOffset;
@@ -233,12 +234,12 @@ void Track::DoSetLinkType(LinkType linkType, bool completeList)
 
    if (oldType == LinkType::None) {
       // Becoming linked
-   
+
       // First ensure there is no partner
       if (auto partner = GetLinkedTrack())
          partner->mpGroupData.reset();
       assert(!GetLinkedTrack());
-   
+
       // Change the link type
       MakeGroupData().mLinkType = linkType;
 
@@ -766,11 +767,11 @@ void TrackList::Clear(bool sendEvent)
    for ( auto pTrack: *this )
    {
       pTrack->SetOwner({}, {});
-      
+
       if (sendEvent)
          DeletionEvent(pTrack->shared_from_this(), false);
    }
-   
+
    for ( auto pTrack: mPendingUpdates )
    {
       pTrack->SetOwner({}, {});
@@ -1355,6 +1356,12 @@ bool ChannelAttachmentsBase::HandleXMLAttribute(
    [&](auto &pAttachment) {
       return pAttachment && pAttachment->HandleXMLAttribute(attr, valueView);
    });
+}
+
+void Track::OnProjectTempoChange(double newTempo)
+{
+   DoOnProjectTempoChange(mProjectTempo, newTempo);
+   mProjectTempo = newTempo;
 }
 
 // Undo/redo handling of selection changes
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -84,7 +84,7 @@ template<typename T>
     been added to a TrackList, or (directly or transitively) copied from such.
     (A track added by TrackList::RegisterPendingNewTrack() that is not yet applied is not
     considered added.)
- 
+
     TrackIds are assigned uniquely across projects. */
 class TrackId
 {
@@ -494,7 +494,7 @@ private:
    Track* GetLinkedTrack() const;
    //! Returns true for leaders of multichannel groups
    bool HasLinkedTrack() const noexcept;
-   
+
    //! Retrieve mNode with debug checks
    TrackNodePointer GetNode() const;
    //! Update mNode when Track is added to TrackList, or removed from it
@@ -537,6 +537,9 @@ public:
 
    void Offset(double t) { SetOffset(GetOffset() + t); }
    virtual void SetOffset (double o) { mOffset = o; }
+
+   // method to set project tempo on track
+   void OnProjectTempoChange(double newTempo);
 
    // Create a NEW track and modify this track
    // Return non-NULL or else throw
@@ -648,6 +651,12 @@ public:
 
    // Return true iff the attribute is recognized.
    bool HandleCommonXMLAttribute(const std::string_view& attr, const XMLAttributeValueView& valueView);
+
+private:
+   virtual void DoOnProjectTempoChange(
+      const std::optional<double>& oldTempo, double newTempo) = 0;
+
+   std::optional<double> mProjectTempo;
 };
 
 ENUMERATE_TRACK_TYPE(Track);
@@ -766,7 +775,7 @@ public:
 
 //! Encapsulate the checked down-casting of track pointers
 /*! Eliminates possibility of error -- and not quietly casting away const
- 
+
 Typical usage:
 ```
 if (auto wt = track_cast<const WaveTrack*>(track)) { ... }
@@ -806,9 +815,9 @@ template < typename TrackType > struct TrackIterRange;
 //! Iterator over only members of a TrackList of the specified subtype, optionally filtered by a predicate; past-end value dereferenceable, to nullptr
 /*! Does not suffer invalidation when an underlying std::list iterator is deleted, provided that is not
     equal to its current position or to the beginning or end iterator.
- 
+
     The filtering predicate is tested only when the iterator is constructed or advanced.
- 
+
     @tparam TrackType Track or a subclass, maybe const-qualified
  */
 template <
@@ -1396,10 +1405,10 @@ public:
    template<typename TrackKind>
       TrackKind *Add( const std::shared_ptr< TrackKind > &t )
          { return static_cast< TrackKind* >( DoAdd( t ) ); }
-   
+
    //! Removes linkage if track belongs to a group
    void UnlinkChannels(Track& track);
-   /** \brief Converts channels to a multichannel track. 
+   /** \brief Converts channels to a multichannel track.
    * @param first and the following must be in this list. Tracks should
    * not be a part of another group (not linked)
    * @param nChannels number of channels, for now only 2 channels supported
@@ -1490,7 +1499,7 @@ private:
 
    Track *GetPrev(Track * t, bool linked = false) const;
    Track *GetNext(Track * t, bool linked = false) const;
-   
+
    template < typename TrackType >
       TrackIter< TrackType >
          MakeTrackIterator( TrackNodePointer iter ) const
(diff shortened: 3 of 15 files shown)


libraries/lib-numeric-formats/ProjectTempoListenerInterface.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include <memory>
+#include <optional>
+
+class NUMERIC_FORMATS_API ProjectTempoListenerInterface
+{
+public:
+   virtual ~ProjectTempoListenerInterface();
+   virtual void OnProjectTempoChange(
+      const std::optional<double>& oldTempo, double newTempo) = 0;
```
Paul-Licameli, line 11: As I said before:  this entire class could be eliminated, and I stead you can write

`using ProjectTempoListener = std::function<…>;`

libraries/lib-numeric-formats/ProjectTimeSignature.cpp
```diff
@@ -49,12 +49,13 @@ void ProjectTimeSignature::SetTempo(double tempo)
 {
    if (mTempo != tempo)
    {
+      const auto oldTempo = tempo;
```
Paul-Licameli, line 52: You mean `oldTempo = Tempo`?

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -312,6 +312,12 @@ void WaveTrack::SetOffset(double o)
    mOffset = o;
 }
 
+void WaveTrack::OnProjectTempoChange(
+   const std::optional<double>& oldTempo, double newTempo)
+{
+   mClipList.OnProjectTempoChange(oldTempo, newTempo);
```
Paul-Licameli, line 318: The extra I direction of WaveClipList serves no purpose.

src/LabelTrack.cpp
```diff
@@ -174,6 +174,11 @@ void LabelTrack::SetOffset(double dOffset)
       labelStruct.selectedRegion.move(dOffset);
 }
 
+void LabelTrack::OnProjectTempoChange(const std::optional<double>& oldTempo, double newTempo)
+{
+   // todo
```
Paul-Licameli, line 179: Supply this before my approval.  I think it’s very easy.

src/NoteTrack.cpp
```diff
@@ -213,6 +213,11 @@ double NoteTrack::GetEndTime() const
    return GetStartTime() + GetSeq().get_real_dur();
 }
 
+void NoteTrack::OnProjectTempoChange(
+   const std::optional<double>& oldTempo, double newTempo)
+{
+}
```
Paul-Licameli, line 219: Not in principle difficult or either, but I won’t demand it

src/ProjectManager.h
```diff
@@ -125,6 +127,8 @@ class AUDACITY_DLL_API ProjectManager final
 
    // non-static data members
    AudacityProject &mProject;
+   ProjectTempoListener mTempoListener;
+   ProjectTempoPublisher mTempoPublisher;
```
Paul-Licameli, line 131: Again I don’t see why the publisher must be a member of ProjectManager.

I would just make it an independent attached object on the Project’s ClientData::Site.

And then it really needs only a .cpp file, no .h file, because it does all its magic by registering its factory at startup.

libraries/lib-numeric-formats/ProjectTempoListenerInterface.h
Paul-Licameli: … if in fact there remains any need to define this interface, but I doubt even that

src/ProjectTempoListener.cpp
```diff
@@ -0,0 +1,27 @@
...
+         }) }
+{
+}
+
+void ProjectTempoListener::OnProjectTempoChange(
+   const std::optional<double>& oldTempo, double newTempo)
+{
+   for (auto track : mTrackList)
+      if (auto waveTrack = dynamic_cast<WaveTrack*>(track))
+         waveTrack->OnProjectTempoChange(oldTempo, newTempo);
```
Paul-Licameli, line 26: Again:  I think you can eliminate this class, and even eliminate the std::function alias, and make all of it implementation detail of the Publisher.

And new criticism:  you need to iterate all tracks now, not just Wave tracks.

Paul-Licameli:
First round of review is done.  Not approved yet.

src/ProjectTempoListener.cpp
saintmatthieu: > And new criticism: you need to iterate all tracks now, not just Wave tracks.

Silly me.

saintmatthieu: > And new criticism: you need to iterate all tracks now, not just Wave tracks.

Silly me.

src/ProjectManager.h
saintmatthieu: Looks like it works. Coming up.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: You know it gets annoying when you're (probably) right for the 10th time in a row?

saintmatthieu:
I've noticed a problem.

1. I create a label track, add a label at time t=1.
2. I half the project tempo, the label goes to t=2, as expected.
3. I undo with ctrl+Z : now the label is at time t=0.5.

Haven't verified this, but I guess this is because the label track is copied for undo purposes. When undoing, it replaces the track modified in the last step. That would fit:

1. I create a label track, ...
2. I half the project tempo, ... => and "undo-list" label track is created beforehand, with label at t=1.
3. I undo ... => the track in the undo list replaces the project's track, with label at t=1, but then the time-signature restorer calls OnProjectTempoChange. So the original label position is restored, but then time-stretched.

Don't know the solution yet.

saintmatthieu:
Looks like I solved it, but it'll be worthwhile reviewing with care.

saintmatthieu:
> Looks like I solved it, but it'll be worthwhile reviewing with care.

I am more confident now.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: Now I remember why this is needed.

If a clip is cut from project A and pasted in B, it needs to know immediately what the tempo of B is to be painted and played back correctly. The clip copy constructor cannot rely on that information from the other clip.

When do we know for sure when a clip becomes visible in a project? I can only think of this moment, when the clip is pushed into a track's clip list. This class is here to make sure that this doesn't happen without a tempo update.

But it doesn't have to be part of this very PR, we'll have time to debate over this.

Paul-Licameli: `WaveClipList` only added a `std::optional<double>` data member, which could just as easily become part of `WaveTrack.`

Paul-Licameli:
Is there now only one commit?

src/NoteTrack.cpp
```diff
@@ -213,6 +213,21 @@ double NoteTrack::GetEndTime() const
    return GetStartTime() + GetSeq().get_real_dur();
 }
 
+void NoteTrack::DoOnProjectTempoChange(
+   const std::optional<double>& oldTempo, double newTempo)
+{
```
Paul-Licameli, line 218: Pleased to see you figured this out.  I had forgotten the details of using the allegro library.

So you have completely implemented this new virtual for all except the WaveTracks!

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: No, if you call `WaveClipList::PushBack`, it sets the project tempo on the clip.

```
void WaveClipList::PushBack(WaveClipHolder clip)
{
   if (mProjectTempo.has_value())
      clip->SetProjectTempo(*mProjectTempo);
   mClips.push_back(std::move(clip));
}
```

libraries/lib-track/Track.cpp
```diff
@@ -44,8 +44,9 @@ Track::Track()
    mOffset = 0.0;
 }
 
-Track::Track(const Track &orig, ProtectedCreationArg&&)
-: vrulerSize( orig.vrulerSize )
+Track::Track(const Track& orig, ProtectedCreationArg&&)
+    : vrulerSize { orig.vrulerSize }
+    , mProjectTempo { orig.mProjectTempo }
```
Paul-Licameli, line 49: Expect a small conflict if you rebase this onto my Clip-views branch.  I eliminate `Track::vrulerSize`.

saintmatthieu:
> Is there now only one commit?

Yes.

src/ProjectTempoListener.cpp
```diff
@@ -0,0 +1,42 @@
+#include "ProjectTempoListener.h"
```
Paul-Licameli, line 1: As I said, this header file isn't needed.  Just put all in the .cpp file.

There are many examples already of .cpp files without companion .h files, when registration at startup time is enough to cause the code to be visited.

For instance in src/export

src/ProjectTempoListener.h
```diff
@@ -0,0 +1,18 @@
...
+class ProjectTempoListener final : public ClientData::Base
+{
+public:
+   ProjectTempoListener(AudacityProject& project, TrackList& trackList);
+   void OnProjectTempoChange(double newTempo);
+
+private:
+   TrackList& mTrackList;
+   Observer::Subscription mTrackListSubstription;
+   Observer::Subscription mProjectTimeSignatureSubscription;
```
Paul-Licameli, line 17: Compilation failed for me becuase "Observer.h" was not included

libraries/lib-mixer/Envelope.cpp
```diff
@@ -824,6 +824,14 @@ void Envelope::RescaleTimes( double newLength )
    mTrackLen = newLength;
 }
 
+void Envelope::RescaleTimesBy(double ratio)
+{
+   for (auto& point : mEnv)
+      point.SetT(point.GetT() * ratio);
+   if (mTrackLen != DBL_MAX)
```
Paul-Licameli, line 831: Compilation failed for me because "float.h" was not included

src/ProjectTempoListener.cpp
```diff
@@ -0,0 +1,42 @@
...
+            if (event.mType == TrackListEvent::ADDITION)
+            {
+               const auto tempo = ProjectTimeSignature::Get(project).GetTempo();
+               if (const auto track = event.mpTrack.lock())
+                  track->OnProjectTempoChange(tempo);
+            }
+         }) }
+{
+   auto& timeSignature = ProjectTimeSignature::Get(project);
+   OnProjectTempoChange(timeSignature.GetTempo());
```
Paul-Licameli, line 30: I think this step of construction is not needed.  There will not yet be any tracks.

src/NoteTrack.cpp
saintmatthieu: Yes - which leaves Audacity in an inconsistent state : change project tempo and all tracks shrink or expand except for the most important one :D

libraries/lib-track/Track.h
```diff
@@ -519,13 +522,19 @@ class TRACK_API Track /* not final */
 
    // Return true iff the attribute is recognized.
    bool HandleCommonXMLAttribute(const std::string_view& attr, const XMLAttributeValueView& valueView);
+
+private:
+   virtual void DoOnProjectTempoChange(
+      const std::optional<double>& oldTempo, double newTempo) = 0;
+
+   std::optional<double> mProjectTempo;
```
Paul-Licameli, line 530: The previous solution remembered old project tempo in ProjectTimeSignature.  Now you store last known tempo in each track.

I think I understand what justifies this.  It's not always duplicated everywhere.  A track could be duplicated into an undo state or in the clipboard.  It remembers what project tempo was at the time the duplicate track was created.  We don't want to normalize it to some tempo value, because we want to delay the time when we must compute the time stretch, which is not perfectly invertible, and so, lossy.

src/NoteTrack.cpp
Paul-Licameli: Good to start testing.  I think copy and paste will work fine for MIDI and label tracks.  The TrackList listener was all you needed.  (TimeTrack does not copy and paste.)

Paul-Licameli:
Now that ProjectTempoListener.h is gone, you might also put all of that class inside the anonymous namespace, making the symbol tables a tiny bit smaller and link time a tiny bit less.

Remembering this in similar cases might have some cumulative effect.  Don't ask me to quantify.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: Still, you might simply change every `push_back` in WaveTrack to use instead a new private member function of WaveTrack that does this.

libraries/lib-track/Track.h
saintmatthieu: This, and also to reduce cognitive load when trying to figure out what happens when tracks in the undo stack miss out on project tempo updates and later get inserted into the list. Now it's easier.

Paul-Licameli (approved):

Paul-Licameli:
It's good enough to merge now.  Compilation is fixed.

Suggestions only to use the anonymous namespace, and skip the needless construction step.

Update the PR or issue description with what QA should test.

The expectation is that three of the four track types update themselves correctly for changes of project tempo, and also that undo and redo work -- and also, cut-copy-paste, even between projects.

Paul-Licameli:
Persistency too should be on the QA checklist

dozzzzer:
@Paul-Licameli could you please rebase this PR on the more recent `master` (containing the fix for that error of having mismatched sample rates when opening a project) so that I could make sure that persistency works?

Paul-Licameli:
What is expected to work, is that note, label, and time tracks correctly adjust how they are drawn as project tempo changes.

The work for wave tracks isn't yet here.

PR #4870, opened 2023-07-04
Set project tempo on clip when added to track

### LINKED ISSUE #4869: Set project tempo on clip when added to track
A small PR, following up on #4867. Clips have to know the project tempo as soon as they become visible in the UI, which is consistent with when a clip is added to a track's clip list. This is what we implement here.

saintmatthieu:
Resolves: #4869 

A small PR for a private method in `WaveTrack`, `InsertClip`.
This method inserts clips in order -> various places where the list was sorted are now obsolete.
The method also maintains project tempo information on a `WaveTrack`'s clips.

Recommended:

Things to test:
Changes are rather limited, and so could be testing. Checked should nevertheless at least be
- [x] Creating clips on a track in random places should not break
- [x] Reordering them by means of copy-paste
- [x] Export and mix-and-rendering after the previous two steps.

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -233,7 +233,7 @@ WaveTrack::WaveTrack(const WaveTrack &orig, ProtectedCreationArg &&a)
 {
    mLegacyProjectFileOffset = 0;
    for (const auto &clip : orig.mClips)
-      mClips.push_back(std::make_shared<WaveClip>(*clip, mpFactory, true));
+      InsertClip(std::make_shared<WaveClip>(*clip, mpFactory, true));
 }
 
 size_t WaveTrack::GetWidth() const
@@ -718,7 +718,7 @@ Track::Holder WaveTrack::Copy(double t0, double t1, bool forClipboard) const
          // Whole clip is in copy region
          //wxPrintf("copy: clip %i is in copy region\n", (int)clip);
 
-         newTrack->mClips.push_back(
+         newTrack->InsertClip(
             std::make_shared<WaveClip>(*clip, mpFactory, !forClipboard));
          WaveClip *const newClip = newTrack->mClips.back().get();
          newClip->Offset(-t0);
@@ -736,7 +736,7 @@ Track::Holder WaveTrack::Copy(double t0, double t1, bool forClipboard) const
          if (newClip->GetPlayStartTime() < 0)
             newClip->SetPlayStartTime(0);
 
-         newTrack->mClips.push_back(std::move(newClip)); // transfer ownership
+         newTrack->InsertClip(std::move(newClip)); // transfer ownership
       }
    }
 
@@ -755,7 +755,7 @@ Track::Holder WaveTrack::Copy(double t0, double t1, bool forClipboard) const
       placeholder->SetIsPlaceholder(true);
       placeholder->InsertSilence(0, (t1 - t0) - newTrack->GetEndTime());
       placeholder->Offset(newTrack->GetEndTime());
-      newTrack->mClips.push_back(std::move(placeholder)); // transfer ownership
+      newTrack->InsertClip(std::move(placeholder)); // transfer ownership
    }
 
    return result;
@@ -1164,7 +1164,7 @@ bool WaveTrack::AddClip(const std::shared_ptr<WaveClip> &clip)
 
    // Uncomment the following line after we correct the problem of zero-length clips
    //if (CanInsertClip(clip))
-      mClips.push_back(clip); // transfer ownership
+   InsertClip(clip); // transfer ownership
 
    return true;
 }
@@ -1306,7 +1306,7 @@ void WaveTrack::HandleClear(double t0, double t1,
    }
 
    for (auto &clip: clipsToAdd)
-      mClips.push_back(std::move(clip)); // transfer ownership
+      InsertClip(std::move(clip)); // transfer ownership
 }
 
 void WaveTrack::SyncLockAdjust(double oldT1, double newT1)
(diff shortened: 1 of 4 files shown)


saintmatthieu:
@Paul-Licameli the lightweight version you proposed to the project tempo sync problem.
I guess I caught all places where `mClips.push_back` was used, but still don't feel so good about it. The `WaveClipList` diff was bigger, but I felt safe.
To get the best of both worlds - safety and small diff - we'd need an `std::vector<WaveClipHolder>` decorator. I have never tried it before, though, and suspect that to be a fair piece of work.

Paul-Licameli:
The vector of clips is still private to WaveTrack so there is still only one source file to review to be sure that the change is complete.

I didn’t see the case for reuse of a separate class 
WaveClipList.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -612,11 +613,14 @@ class WAVE_TRACK_API WaveTrack final : public WritableSampleTrack
...
    bool RateConsistencyCheck() const;
 
+   void PushBack(WaveClipHolder clip);
+
    SampleBlockFactoryPtr mpFactory;
 
    wxCriticalSection mFlushCriticalSection;
    wxCriticalSection mAppendCriticalSection;
    double mLegacyProjectFileOffset;
+   std::optional<double> mProjectTempo;
```
Paul-Licameli, line 623: This duplicates the member in base class Track.  Perhaps define a const accessor function instead

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1522,6 +1523,13 @@ bool WaveTrack::RateConsistencyCheck() const
       });
 }
 
+void WaveTrack::PushBack(WaveClipHolder clip)
```
Paul-Licameli, line 1526: I’d like a different name for this function, like `InsertClip`

I may want to strengthen the internal invariant, so that clips are in fact maintained in sorted sequence.

The function name should not suggest something about the position of the clip in the list.

Paul-Licameli (approved):

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Good idea.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: Good idea.

Paul-Licameli:
But there are unit test failures

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
```diff
@@ -38,14 +38,8 @@ AudioSegmentFactory::CreateAudioSegmentSequence(
...
 {
-   ClipConstHolders sortedClips { mClips };
-   std::sort(
-      sortedClips.begin(), sortedClips.end(),
-      [](const ClipConstHolder& a, const ClipConstHolder& b) {
-         return a->GetPlayStartTime() < b->GetPlayStartTime();
-      });
    std::vector<std::shared_ptr<AudioSegment>> segments;
-   for (const auto& clip : sortedClips)
+   for (const auto& clip : mClips)
```
Paul-Licameli, line 42: Don’t add this commits!  I said I may strengthen the ordering guarantee in future.  That work is not done yet.

Paul-Licameli (changes requested):

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1522,6 +1522,18 @@ bool WaveTrack::RateConsistencyCheck() const
...
+void WaveTrack::InsertClip(WaveClipHolder clip)
+{
+   const auto& tempo = GetProjectTempo();
+   if (tempo.has_value())
+      clip->OnProjectTempoChange(std::nullopt, *tempo);
+   auto it = mClips.begin();
+   while (it != mClips.end() &&
+          (*it)->GetPlayStartTime() < clip->GetPlayStartTime())
+      ++it;
+   mClips.insert(it, std::move(clip));
```
Paul-Licameli, line 1534: Don’t do this.  Just push back as in what I reviewed before.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2289,11 +2301,6 @@ ChannelSampleView WaveTrack::GetOneSampleView(
          intersectingClips.push_back(clip);
    if (intersectingClips.empty())
       return { AudioSegmentSampleView(length) };
-   std::sort(
-      intersectingClips.begin(), intersectingClips.end(),
-      [](const auto& a, const auto& b) {
-         return a->GetPlayStartTime() < b->GetPlayStartTime();
-      });
```
Paul-Licameli, line 2262: Don’t remove this.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -613,6 +614,9 @@ class WAVE_TRACK_API WaveTrack final : public WritableSampleTrack
    //! Whether all clips have a common rate
    bool RateConsistencyCheck() const;
 
+   //! Inserts in order, beside doing other maintenance work
```
Paul-Licameli, line 617: Don’t over promise in the comment.

saintmatthieu:
@Paul-Licameli I'd like to be given a reason why it is wrong to keep the clips in sorted order now already. It may not be strictly necessary for this PR, but I guess it will never be and I can't see how it hurts.

Paul-Licameli:
Don’t complicate this pull request.

Your attempt was incomplete anyway because you didn’t sort tracks when deserializing a saved project, and I’m not sure what else is needed to maintain the invariant in case clips are modified.  Reasoning about the completeness is more review burden on me that I don’t want just now.

I see build failures because unit tests don’t pass.  Figure this out first.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2289,11 +2297,6 @@ ChannelSampleView WaveTrack::GetOneSampleView(
          intersectingClips.push_back(clip);
    if (intersectingClips.empty())
       return { AudioSegmentSampleView(length) };
-   std::sort(
```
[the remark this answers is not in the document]
saintmatthieu: Ah crap. Thanks.

Paul-Licameli (approved):

PR #4876, opened 2023-07-05
Lib channel

Paul-Licameli:

Depends on
- #4825
- #4844
- #4797

Split lib-channel from lib-track.

Define interfaces to iterate over a Track (or ChannelGroup, its new base) by
significant time intervals, alternatively to the iteration over Channels.

Define abstract ChannelInterval as the intersection of the information of a Channel
and a ChannelGroupInterval;  and Channel and ChannelGroupInterval each define
an iterator over those.

Thus these are somewhat like rows, columns, and cells of a matrix, and convenient
striding by either.

QA:
- [x] Known behavior changes to ignore:  In case of misalignments, right channel clip boundaries are ignored for snapping.
- [x] Briefly test display & interaction with the four kinds of tracks; and mono and stereo
- [x] Painting of track (not clip) names (a preference turns them on)
- [x] Resizing of track height and channel height
- [x] Time Shift Left and TimeShiftRight (under Extra > Edit), applied to selections within each of the four kinds of tracks (bind them to keystrokes to make repeated trail easy; these are important for accessibility)
- [x] Punch and roll recording
- [x] Dragging clips (left, right, and between tracks)
- [x] Undo and redo of channel height and minimization changes

Recommended:


Paul-Licameli:
Rebased onto recent master after merging two dependencies.

Paul-Licameli:
Rebased again.  All the dependencies are now merged

Paul-Licameli:
Rebased onto #4887 which fixes a recent regression and changes Track.*

Paul-Licameli:
More self-review of the time-shifting stuff to my own satisfaction now.

Rebased onto recent master.

Not dependent on bug fix #4885, which still has other blocking problems still to be figured out.

crsib (approved):

Paul-Licameli:
Rebased onto more recent master, resolving unimportant conflict in comments

dozzzzer:
Tested

PR #4883, opened 2023-07-07
Fix shifting clip to left of 0 mark breaking spectrum display

### LINKED ISSUE #4882: Shifting clip to left of 0 mark breaks spectrum display
### Bug description

_No response_

### Steps to reproduce

1. Have a track with audio.
2. Switch to spectrum display.
3. Move clip left to the zero mark.

### Expected behavior

Spectrum display works.

### Actual behavior

Spectrum display turns black.

### Audacity Version

### Operating system

Windows

### Additional context

Was introduced by recent caching-refactoring PR https://github.com/audacity/audacity/pull/4771

saintmatthieu:
Resolves: #4882

Fix in recent caching refactoring code, where the calculation of number of samples for readout wrongly compared absolute with relative times.

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -2287,21 +2287,21 @@ WaveTrack::GetOneSampleView(sampleCount start, size_t length) const
       const auto clipStartTime = clip->GetPlayStartTime();
       if (t0 < clipStartTime) {
          const auto numSamples = TimeToLongSamples(clipStartTime - t0);
-         segments.push_back(AudioSegmentSampleView(numSamples));
+         segments.push_back(AudioSegmentSampleView{numSamples});
          t0 = clipStartTime;
+         length -= numSamples.as_size_t();
       }
       const auto clipT0 = t0 - clipStartTime;
       const auto clipS0 = TimeToLongSamples(clipT0);
-      // By substitution,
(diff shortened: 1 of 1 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2268,21 +2268,21 @@ WaveTrack::GetOneSampleView(sampleCount start, size_t length) const
...
       const auto clipT0 = t0 - clipStartTime;
       const auto clipS0 = TimeToLongSamples(clipT0);
-      // By substitution,
-      //    (len - length) / rate = t0 - clipT0,
-      // and since
-      //    t0 - clipT0 <= 0,
-      // then `length >= len`. `as_size_t` won't narrow.
-      const auto len = TimeToLongSamples(t1 - clipT0).as_size_t();
+      const auto len =
+          std::min((clip->GetPlaySamplesCount() - clipS0).as_size_t(), length);
```
Paul-Licameli, line 2278: You might use `limitSampleBufferSize`

Paul-Licameli: In fact that would be more correct.  The narrowing to size_t here might sometimes make a small number.  Better to widen `length` before taking `min`

Paul-Licameli (changes requested):
Make the suggested change to avoid narrowing.

Paul-Licameli (approved):

PR #4890, opened 2023-07-10
Exporting a single track selection sets filename to track name

sdev23:
n/a

perhaps it's just me, but i was expecting the filename to be the track name when exporting a single track selection. this isn't a complaint, just to be clear. i just want to offer some code and see what happens.

i'm guessing `TrackList::Selected::empty()` will not return `true` if `Export::mNumSelected` is greater than zero, which is why i didn't check. that being said, i don't know if there could be a null `Track` in the container, so i did check for that.

Recommended:


sdev23:
thanks @LWinterberg

Issue #4892, opened 2023-07-10
Undo/redo clip trims separately


dozzzzer:
### Description

Currently, when Undoing/Redoing several clip trim operations, Audacity only cares about whether from the left or from the right edge of the clip trim has been made. So, if several clips are trimmed from the same side one by one, Audacity would undo the whole sequence of trim operations at once:

https://github.com/audacity/audacity/assets/70063970/b5ab9da0-dc0a-45fd-b906-decc892f5c13

### STR

1. Have a few clips on the same audio track
2. Trim every one of them by dragging the right edge to the left
3. Edit > Undo or Ctrl (Cmd) + Z
4. Observe

### Actual Behavior

All clips are reversed to their previous (untrimmed) state

### Expected Behavior

Only the last trim action gets undone; 
Undo/Redo trim clip is performed for each individual clip separately.

LWinterberg:
this sounds more like a bug, rather than a feature request

dozzzzer:
Tested

PR #4912, opened 2023-07-14
Don't consolidate undo items for clip trim handle

Paul-Licameli:
Resolves: #4892

Recommended:


crsib (commented):
But is it the desired behavior? I think that they should be consolidated per clip. (@dozzzzer @LWinterberg)

LWinterberg:
I don't think they should be consolidated per clip. If you take two steps to trim a clip, undoing can well show these two steps.

dozzzzer:
> But is it the desired behavior?

Yes, the fix looks good to me

crsib (approved):

PR #4916, opened 2023-07-14
Fix ImportMP3_MAD.cpp and ImportQT.cpp include directives: fixes 'file not found' errors

mehw:
Hello,

This fixes `file not found` errors about header files included in Audacity.

Tested on Gentoo GNU/Linux.

### DIFF
--- a/src/import/ImportMP3_MAD.cpp
+++ b/src/import/ImportMP3_MAD.cpp
@@ -62,10 +62,10 @@ static Importer::RegisteredUnusableImportPlugin registered
 #include <wx/file.h>
 
 #include "Prefs.h"
-#include "../Tags.h"
+#include "Tags.h"
 #include "WaveTrack.h"
-#include "../widgets/AudacityMessageBox.h"
-#include "../widgets/ProgressDialog.h"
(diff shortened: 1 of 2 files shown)


crsib:
Most likely there is something spooky about your setup. Audacity no longer builds with this importers, they are removed all together in the soon-to-be merged Import/Export rewrite.

mehw:
Thanks, good to know. I saw that the same fix has been applied in:
- commit 3cfc1482d9476c953855744fc7e8605d4884b926 (`WaveTrack.h`)
- commit 6b7849b30af768275bbee223074f74f731a6885f (`Tags.h`)
- commit 61decb5bbf3bf66519c44ba95125d7ab6adff0c3 (`AudacityMessageBox.h`, `ProgressDialog.h`)

Without, Audacity doesn't compile in my instance of Gentoo GNU/Linux when enabling the still present libmad support (it fails on `ImportMP3_MAD.cpp` about some of the mentioned headers, the `#include "WaveTrack.h"` was already fixed).

The fix to `ImportQT.cpp` in this PR is just an extra to correct the include directives there too.

PR #4917, opened 2023-07-14
Add missing Findlibmad.cmake file to resolve the system installed libmad

mehw:
Hello,

This allows compiling against the system installed `libmad`, that otherwise cannot be found.

Tested on Gentoo GNU/Linux with libmad 0.15.1b installed:
```
audacity_use_libmpg123=off
audacity_use_libmad=system
```

The `Findlibmad.cmake` file is an edited copy of `Findlibmp3lame.cmake`.

### DIFF
--- a/cmake-proxies/cmake-modules/Findlibmad.cmake
+++ b/cmake-proxies/cmake-modules/Findlibmad.cmake
@@ -1,0 +1,35 @@
+#[[
+A module to look for libmad
+]]
+
+if( NOT libmad_FOUND )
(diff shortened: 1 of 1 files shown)


crsib:
libmad support is soon to be dropped.

mehw:
Ok. I read your answer in https://github.com/audacity/audacity/pull/4916#issuecomment-1636717939, thanks.

I'll keep an eye on PR https://github.com/audacity/audacity/pull/4384, which will remove support for libmad.

PR #4938, opened 2023-07-20
fix #4410 (mURIDMap -> gURIDMap)

### LINKED ISSUE #4410: LV2 unmap feature causes plugin exit
Setting a URID from a given URI results in a valid integer.  However passing this integer along with the correct handle causes a segmentation violation.

LWinterberg:
Resolves: #4410

PRified the fix provided in the issue by @jfroyall

Recommended:


LWinterberg:
(Given the issue this fixes doesn't have a STR, I'm not sure if this really is meaningfully QA-able. This PR probably stands and falls with just the code review determining whether this actually is a typo)

crsib (approved):

PR #4945, opened 2023-07-23
Rewrite more effects

Paul-Licameli:
Resolves: #4943

Depends on
- #4937

Rewrite more iterations over tracks without Any() or Selected() in src/effects.

Still not done removing Selected() in TwoPassSimpleMono.cpp which is used by Compressor.
That will wait until a later PR.

Fix a minor problem with AutoDuck and a missed internationalization in FindClipping (strings in generated labels).

- [x] Amplify
- [x] Auto Duck
- [x] Click Removal
- [x] Equaliazation
- [x] Find Clipping
- [x] Repair
- [x] Compressor
- [x] Zoom-to-fit behavior when generating the first track in an empty project
- [x] Reverse (test with label tracks; multiple clips and silences between; selection boundaries inside clips or inside silences)
- [x] Normalize
- [x] Loudness (with and without the process-independently checkbox)

- [x] I signed [CLA](https://www.audacityteam.org/cla/)
- [x] The title of the pull request describes an issue it addresses
- [x] If changes are extensive, then there is a sequence of easily reviewable commits
- [x] Each commit's message describes its purpose and effects
- [x] There are no behavior changes unnecessary for the stated purpose of the PR

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes of behavior

PR #4961, opened 2023-07-27
Track start and end times

Paul-Licameli:

Depends on:
- #4945

Recommended:


Paul-Licameli:
I have found that this problem can be demonstrated with shorter tracks, and it may influence other effects when sample rate is not default.

Even creating a stereo track, setting its rate to 48k, and generating noise makes wrong results.

Paul-Licameli:
Rebased without changes onto the updated #4945

crsib (approved):

crsib:
Have you pushed, though?

Paul-Licameli:
> Have you pushed, though?

Sorry, yes now

crsib (approved):
I'm not sure, that this PR will be merged before there exporter PR is complete

Paul-Licameli:
> I'm not sure, that this PR will be merged before there exporter PR is complete

Exactly, I want this one based on that.  That other one may still change more, but I have done the known work of conflict resolution.

My next PR in the queue #4974 needs a rebase onto this one and other conflict resolutions with the import/export branch.

dozzzzer:
@Paul-Licameli found issues:

- **Align tracks > Align Together** produces incorrect result: 

https://github.com/audacity/audacity/assets/70063970/9ae37057-4fe7-4eef-8f53-9dc90a843594

Expected result:

https://github.com/audacity/audacity/assets/70063970/450504fe-1617-4d79-b385-d8c8688cef35

- crash when mixing stereo down to mono

Paul-Licameli:
A two-line fix for the problem in alignment.

crsib (approved):

libraries/lib-channel/Channel.cpp
```diff
@@ -64,3 +67,25 @@ size_t Channel::GetChannelIndex() const
...
+         return std::min(acc, pInterval->Start()); });
+}
+
+double ChannelGroup::GetEndTime() const
+{
+   auto range = Intervals();
+   if (range.empty())
+      return 0;
+   return std::accumulate(range.first, range.second,
+      std::numeric_limits<double>::min(),
```
vsverchinsky, line 88: Should it be simply `.0`? Or if time is allowed to be negative should it be `std::numeric_limits<double>::lowest()`?
> For floating-point types with denormalization, min() returns the minimum positive normalized value

chinakov:
@Paul-Licameli - issue with Align Tracks has been fixed. Is it true that the crash on mixdown to mono also got fixed? I couldn't reproduce it - but I am not sure if it's because it got fixed.

libraries/lib-channel/Channel.cpp
Paul-Licameli: Thanks!  You are right.  I remembered this in other commits of this pull request but forgot it here.

chinakov:
Acceptance criteria fulfilled, and checklist tested. This is ready for merge, thank you!

PR #4974, opened 2023-07-31
Last track iteration rewrites

Paul-Licameli:

Depends on:
- #4961

This very hopefully titled pull request is not complete but review may begin on the early commits.

When all is done, the last remaining uses of TrackList::Any, Selected, Find, and direct range-for
outside of class Track itself will be gone.

Then might follow renaming of Leaders, SelectedLeaders, and FindLeader back to Any, Selected, Find.

But after that still remains a large task:  Review all the uses of TrackList::Channels (many of them
recently introduced), and make them disappear except inside classes Track and WaveTrack.

QA:
- [x] Persistency of projects with stereo tracks
- [x] Ctrl+A, C, X, V keys, or corresponding menu commands, when there is an edit cursor in a clip title
- [x] Very old code in "Compare Audio" macro command was changed (so it really works on mono and stereo tracks) but maybe nobody cares
- [x] Append-record, and do Ctrl+B to add labels, and make some display changes (like track height), while recording; then stop and test Undo and Redo (All of the recording should be sequenced in undo history after the other changes)
- [x] Change Speed effect on a stereo track
- [x] A Nyquist generator
- [x] Find Clipping done with and without a label track called "Clipping" already present
- [x] Import mono, stereo, (and examples with more than two channels if you can find them??) using FFmpeg, FLAC, MPG 123 (How?  I don't know, ask Dmitry), OGG, WavPack; check correct muting when a solo track was already in the project; check numbers added to track names, in case more than one new track is made by one import
- [x] Import Raw making a stereo track (I'm not sure how you do this, but data-bending can be fun)
- [x] Recording dropout detection into stereo track (there is an alpha item in Tools to enable simulation of droupout errors)
- [x] FFmpeg import example with stream delay (how?)
- [x] Generate Silence into a selection in a stereo track
- [x] Compressor in stereo track
- [x] High quality Change Pitch on stereo
- [x] Save mono and stereo tracks in Audacity 2.4.2, then open the project in 3.4 (this crashes in master! but not in 3.3)
- [x] Cross-paste copy-paste with hidden data on stereo tracks, and long enough to cause progress dialogs, with appropriate scales; try with all the different "Pasted Audio" choices in the Tracks Behaviors preferences
- [x] Change format of a long mono or stereo track, with appropriate progress indicator
- [x] Opening of projects with mono and stereo tracks that were saved in 3.3
- [x] Compressor in a stereo track, again

Recommended:

**Noticed issues:**

- [x] Generate Silence into a selection in a stereo track

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -1551,9 +1551,10 @@ void AudioIO::StopStream()
                auto &start = interval.first;
                auto duration = interval.second;
                for (auto &sequence : mCaptureSequences) {
-                  GuardedCall([&] {
-                     sequence->InsertSilence(start, duration);
-                  });
+                  if (sequence->IsLeader())
+                     GuardedCall([&] {
+                        sequence->InsertSilence(start, duration);
+                     });
                }
             }
             if (pScope)
--- a/libraries/lib-effects/EffectBase.cpp
+++ b/libraries/lib-effects/EffectBase.cpp
@@ -230,7 +230,7 @@ const AudacityProject *EffectBase::FindProject() const
 
 void EffectBase::CountWaveTracks()
 {
-   const auto range = mTracks->SelectedLeaders<const WaveTrack>();
+   const auto range = mTracks->Selected<const WaveTrack>();
    mNumTracks = range.sum(&WaveTrack::NChannels);
    mNumGroups = range.size();
 }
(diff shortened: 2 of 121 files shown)


Paul-Licameli:
Two more commits, leaving only one more call to `Selected` to clean up, in TwoPassSimpleMono.cpp.

But that remaining one is still not easy.

Paul-Licameli:
Rebased unchanged onto #4961 which was rebased onto master after the merge of import/export refactoring

Paul-Licameli:
Rebased, unchanged onto updated #4961

crsib (approved):

chinakov:
If a User chooses to generate 'Silence' in any selection in a Stereo Track, the Right channel won't be silenced, even though it's selected. 
Silence not applied:
Screenshot 2023-08-08 at 14 43 35

Silence applied:
Screenshot 2023-08-08 at 14 44 02

crsib (approved):

Issue #4960, opened 2023-07-26
Windows: New export dialog: initial focus is not the first control


DavidBailes:
### Steps to reproduce

using the commit: https://github.com/audacity/audacity/commit/d3d98a434a0cf9511526842b70799234c0cdbba4

1. Open the Export audio dialog
2. Which control is the initial focus?

### Expected behavior

The first control is the initial focus (unless there are very good reasons otherwise), which is the file name text box.

### Actual behavior

One of the channels radio buttons is the initial focus.
For users of screen readers, and probably others, this is not at all helpful.

### Audacity Version
New Export branch build

### Operating system

Windows

chinakov:
Just a note: It works as expected on macOS builds of Audacity: the initial focus there is in the file name text box. 
<img width="2152" alt="Screenshot 2023-07-27 at 18 49 44" src="https://github.com/audacity/audacity/assets/139755141/e032abaa-ecac-4d02-a19f-b69821576b93">

DavidBailes:
> Just a note: It works as expected on macOS builds of Audacity: the initial focus there is in the file name text box.

and also works as expected on XUbuntu 22.04:
![image](https://github.com/audacity/audacity/assets/11887021/f8291e0e-4c1a-495e-81c6-88ea145d2fb7)

petersampsonaudacity:
@DavidBailes @chinakov 
Testing on macOS 12.6.7 Monterey with @vsverchinsky 's latest branch build

This works properly on Mac with the initial focus on the **File Name** field

**BUT** I can't figure out how to change the Channels setting
and **WORSE** tabbing gets stuck circling round inin the Channels section

petersampsonaudacity:
Testing on W10 (and Mac macOS 12.6.7) with @DavidBailes 's latest branch build:  audacity-win-3.4.0-alpha-20230803+43a29d1-x64

1) this now works properly on Windows with this branch build with the initial dialog entry **Filename** having the focus on calling up the Export Audio dialog.

2) On Mac **Filename** remains the initial focus so remains working properly _(as was that case in earlier 3.4.0 alphas)_

chinakov:
@DavidBailes - it seems that the focus issues has been fixed. However, @petersampsonaudacity is correct - macOS now has unwanted behaviour - tabbing no longer works properly. 

On Windows a User can tab through every single element in the Export Audio dialog window. 
On macOS, a User can only tab through the File Name and Folder fields, nothing else. 

[...]

Issue #1561, opened 2021-08-30
Preferences > Shortcuts: wrong focus in Shortcuts pref


petersampsonaudacity:
**Describe the bug**
On Windows when launching preferences the focus is on the "View by" radio buttons and not the presences selector.
This particularly affects Non-mouse users, VI users.

This was originally logged on Bugzilla as P3 2691
https://bugzilla.audacityteam.org/show_bug.cgi?id=2691

**To Reproduce**
1. Open preferences
2. Select the Shortcuts category.
3. Press the OK button.
4. Reopen preferences.
5. Observe: one of the "View by" radio buttons is the focus. 

**Expected behavior**
The "Shortcuts" item in the categories tree should be the focus.

**Additional information (please complete the following information):**
 - OS: Windows 10
 - Version: Audacity 3.0.4 and earlier

**Additional context**
David Bailes wrote on Bugzilla:
>When the preferences dialog opens, the initial focus should always be an item in the categories tree. This is not the case - see steps to reproduce.
This bug was not present in Audacity 2.4.1, but is present by the time of the commit 1f366ee0730f752da.
It looks like it may have been caused by the switch to using wxWidgets 3.1.3 - no idea why.

And subsequently he wrote:
>This appears to be a wxWidgets bug.
Found this commit, which is included in wxWidgets 3.1.4, which is in the same area:
https://github.com/wxWidgets/wxWidgets/commit/68d7cb50822451b393f34680417d85c2354783cb
I've build Audacity with wxWidgets 3.1.4, and this fixes the problem.

dozzzzer:
Not reproducible anymore: now, the first item in the tree - Audio Settings - is in focus when opening Preferences.

LWinterberg:
Hello! We just [switched development to Audacity 4](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444). We've redone, or plan on redoing, various things from scratch, including the entire UI, and the part this bug is about. As such, I'm closing this bug as it will be fixed there and we won't be fixing it for the final releases of Audacity 3.

PR #4989, opened 2023-08-03
Windows: New export dialog: initial focus is not the first control

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/4960

Problem:
This appears to be caused by a bug in wxWidgets 3.1.3, and only affects Windows. Issue #1561 is probably caused by the same bug.

Fix:
For Windows only, explicitly set the initial focus.

Recommended:

### DIFF
--- a/src/export/ExportAudioDialog.cpp
+++ b/src/export/ExportAudioDialog.cpp
@@ -218,6 +218,19 @@ ExportAudioDialog::ExportAudioDialog(wxWindow* parent,
 
 ExportAudioDialog::~ExportAudioDialog() = default;
 
+// Fix for issue #4960, which only affects Windows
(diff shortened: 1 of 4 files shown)


vsverchinsky (approved):

petersampsonaudacity:
Testing on W10 (and Mac macOS 12.6.7) with @DavidBailes 's latest branch build:  audacity-win-3.4.0-alpha-20230803+43a29d1-x64

1) this now works properly on Windows with this branch build with the initial dialog entry **Filename** having the focus on calling up the Export Audio dialog.

2) On Mac **Filename** remains the initial focus so remains working properly _(as was that case in earlier 3.4.0 alphas)_

Issue #4931, opened 2023-07-19
Beats & Measures Gridlines


crsib:
### Gridlines for Beats & Measures
Even though Users are able to snap clips and selection to the new beats & measures timeline format, the alignment process can still be a bit finicky.

Our new gridlines design seeks to do two things:
- Extend the ruler hatch marks to the entirety of the track canvas to improve visibility of the ruler.
- Act as a visual cue to help Users differentiate between Minutes & seconds and Beats & meaures.

![Image](https://github.com/audacity/designtasks/assets/111350316/f777474d-0932-46ca-84d9-69c46fb82c57)

As the User zooms in and out of Audacity, the timeline ruler adjusts accordingly. As the gridlines are essentially an extension of the timeline ruler hatch-marks, they will also respond to zoom levels. Gridline formats have been specified for typical zoom-levels in the spec provided below.

[Link to specification -> ](https://zpl.io/YYQegye)

chinakov:
Tested, thank you!

chinakov:
#5039

chinakov:
Seems like this issue is also reproduced on Master, which means that this branch is also affected. I am closing this task, please merge, and fix will be merged after.

PR #4994, opened 2023-08-04
Musical gridlines

crsib:
Resolves: #4931

Recommended:


src/widgets/Ruler.cpp
```diff
@@ -92,7 +92,8 @@ Ruler::Ruler(const RulerUpdater &updater, const RulerFormat &format)
 
 Ruler::~Ruler()
 {
-   Invalidate();  // frees up our arrays
```
vsverchinsky, line 95: It has become redundant indeed since raw pointers are eliminated

vsverchinsky (approved):

PR #5004, opened 2023-08-07
Application settings refactoring

vsverchinsky:
This PR removes explicit dependency on `wxFileConfig` that sometimes require `wxApp` instance. That should ease ongoing replaforming.

Recommended:


libraries/lib-preferences/BasicSettings.h
```diff
@@ -0,0 +1,226 @@
...
+#pragma once
+
+#include <memory>
+#include <wx/string.h>
+#include <wx/arrstr.h>
+#include "GlobalVariable.h"
+
+namespace audacity
+{
+   class PREFERENCES_API BasicSettings
```
crsib, line 20: Some Doxy comments?

libraries/lib-preferences/BasicSettings.h
```diff
@@ -0,0 +1,226 @@
...
+
+         GroupScope(BasicSettings& settings) : mSettings(settings) { }
+
+      public:
+         GroupScope(const GroupScope&) = delete;
+         GroupScope(GroupScope&&) = delete;
+         GroupScope& operator=(const GroupScope&) = delete;
+         GroupScope& operator=(GroupScope&&) = delete;
+
+         void Reset() noexcept
```
crsib, line 37: Move to CPP?

libraries/lib-preferences/BasicSettings.h
```diff
@@ -0,0 +1,226 @@
...
+#include "GlobalVariable.h"
+
+namespace audacity
+{
+   class PREFERENCES_API BasicSettings
+   {
+   public:
+      class GroupScope final
+      {
+         friend class BasicSettings;
```
crsib, line 25: No need, nested class is always a friend

libraries/lib-wx-init/SettingsWX.cpp
```diff
@@ -0,0 +1,226 @@
...
+   : mIsConfigOwner {true}
+{
+   mConfig = std::make_unique<wxFileConfig>(wxEmptyString, wxEmptyString, filepath);
+   mGroupStack.push_back("/");
+}
+
+SettingsWX::~SettingsWX()
+{
+   mConfig->Flush();
+   if(!mIsConfigOwner)
```
crsib, line 55: This scares me a lot, as I don't see how it can be safe. I think this is the case where shared_ptr shines.

libraries/lib-wx-init/SettingsWX.cpp
```diff
@@ -0,0 +1,226 @@
...
+bool SettingsWX::Write(const wxString& key, const wxString& value)
+{
+   return mConfig->Write(MakePath(key), value);
+}
+
+bool SettingsWX::Flush() noexcept
+{
+   try
+   {
+      return mConfig->Flush();
```
crsib, line 210: Probably it's worth commenting that complex Flush implementation is hidden in mConfig now.

libraries/lib-wx-init/SettingsWX.cpp
```diff
@@ -0,0 +1,226 @@
...
+   return mConfig->HasGroup(MakePath(key));
+}
+
+bool SettingsWX::Remove(const wxString& key)
+{
+   if(key.empty())
+   {
+      {
+         const auto children = GetChildGroups();
+         for(auto& group : children)
```
crsib, line 117:
```c++
for(auto& group : GetChildGroups())
   ...
```
would simplify the code a lot, removing unnecessary nesting

libraries/lib-wx-init/SettingsWX.h
```diff
@@ -0,0 +1,57 @@
...
+   void DoBeginGroup(const wxString& prefix) override;
+   void DoEndGroup() noexcept override;
+
+public:
+   ///@brief Constructs default application settings
+   SettingsWX();
+   SettingsWX(std::unique_ptr<wxConfigBase> config);
+   ///@brief Constructs BasicSettings object to access BasicSettings stored in the file,
+   ///using platform-specific format
+   SettingsWX(const wxString& filepath);
```
crsib, line 26: Mark `explicit` please

libraries/lib-wx-init/SettingsWX.h
```diff
@@ -0,0 +1,57 @@
...
+   std::unique_ptr<wxConfigBase> mConfig;
+   const bool mIsConfigOwner;
+protected:
+   void DoBeginGroup(const wxString& prefix) override;
+   void DoEndGroup() noexcept override;
+
+public:
+   ///@brief Constructs default application settings
+   SettingsWX();
+   SettingsWX(std::unique_ptr<wxConfigBase> config);
```
crsib, line 23: `unique_ptr` makes a lot of sense here, as you express that SettingsWX takes the config ownership from the caller. However, I think `mConfig` should be a shared pointer. Luckily, shared_ptr has a constructor that consumes `unique_ptr`.

Also, I vote for marking the constructor `explicit`.

crsib (changes requested):
There is some feedback I would like to see addressed

libraries/lib-wx-init/SettingsWX.h
vsverchinsky: Not sure if I understand that, but I attempted to change ownership policy in a new separate commit

crsib (approved):

dozzzzer:
Tested

PR #5008, opened 2023-08-07
Channel iteration rewrites

Paul-Licameli:

Depends on:
- #4974 
- #4987 
- #5060

Begin to eliminate the uses of TrackList::Channels outside Track.cpp and WaveTrack.cpp,
including all except one (Repeat) in src/effects.

Instead use WaveTrack::Channels which return objects of the new WaveChannel class,
which does not inherit Track but has a sufficient interface for many purposes, including
getting and setting and appending of sample data, and queries of minimum, maximum,
and RMS values, which are used by effects and analyzers and display.

QA: test stereo and mono cases
- [x] Bug since commit 7d179623: changing track rate shifts track incorrectly
- [x] Recording & playback
- [x] Split stereo (track context menu); correctly panned resulting tracks
- [x] Chirp, tone, silence generation
- [x] EQ
- [x] Change Speed
- [x] Noise Reduction
- [x] Paulstretch
- [x] Change Pitch (high and low quality)
- [x] Sliding time scale (ditto)
- [x] Amplify
- [x] Auto Duck
- [x] Click Removal
- [x] Contrast
- [x] Find Clipping
- [x] Loudness
- [x] Normalize
- [x] Repair
- [x] Truncate Silence
- [x] Compressor
- [x] Any one Vamp plug-in (they make label tracks; see that track naming doesn't change)
- [x] Some Nyquist effects and generators, stereo and mono
- [x] Mixer board sliders and output meters
- [x] Our friend Stereo To Mono, again; cases of multiple selected tracks, some of them already mono, which should be unchanged, and each stereo track correctly mixed down
- [x] Dragging of clips from one track to another
- [x] Export, split-by-tracks
- [x] Compare Audio macro command
- [x] Set rate in the track context menu (not resampling from the Track pull-down menu)
- [x] Keystrokes to edit clip names (including Ctrl+A, C, V, X)
- [x] Correct naming of clips by Import
- [x] Correct naming of multiple new clips by the Repeat effect

Recommended:

**Noticed Issues:**

- [x] Setting a different sample rate in the track context menu results in misaligned audio clips in stereo tracks.
- [x] Splitting a Stereo track to Mono via the Track Context menu results in the last channel being 'unnamed'.

### DIFF
--- a/src/effects/nyquist/Nyquist.cpp
+++ b/src/effects/nyquist/Nyquist.cpp
@@ -683,7 +683,7 @@ struct NyquistEffect::NyxContext {
 
    WaveTrack *mCurChannelGroup{};
    WaveChannel       *mCurTrack[2]{};
-   sampleCount       mCurStart[2]{};
+   sampleCount       mCurStart{};
 
    unsigned          mCurNumChannels{}; //!< Not used in the callbacks
 
@@ -952,7 +952,6 @@ bool NyquistEffect::Process(EffectInstance &, EffectSettings &settings)
                mCurNumChannels = 2;
 
                mCurTrack[1] = (* ++ channels.first).get();
-               mCurStart[1] = mCurTrack[1]->TimeToLongSamples(mT0);
             }
 
             // Check whether we're in the same group as the last selected track
@@ -960,9 +959,9 @@ bool NyquistEffect::Process(EffectInstance &, EffectSettings &settings)
             mFirstInGroup = !gtLast || (gtLast != gt);
             gtLast = gt;
 
-            mCurStart[0] = mCurChannelGroup->TimeToLongSamples(mT0);
+            mCurStart = mCurChannelGroup->TimeToLongSamples(mT0);
             auto end = mCurChannelGroup->TimeToLongSamples(mT1);
-            mCurLen = end - mCurStart[0];
+            mCurLen = end - mCurStart;
 
             wxASSERT(mCurLen <= NYQ_MAX_LEN);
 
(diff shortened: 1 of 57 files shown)


Paul-Licameli:
Rebased unchanged onto master, now that the two dependencies are merged.

src/effects/nyquist/Nyquist.cpp
```diff
@@ -943,15 +943,15 @@ bool NyquistEffect::Process(EffectInstance &, EffectSettings &settings)
...
+            if (auto channels = mCurChannelGroup->Channels()
                ; channels.size() > 1
             ) {
                // TODO: more-than-two-channels
                // Pay attention to consistency of mNumSelectedChannels
                // with the running tally made by this loop!
                mCurNumChannels = 2;
 
-               mCurTrack[1] = * ++ channels.first;
+               mCurTrack[1] = &** ++ channels.first;
```
crsib, line 954: Completely unreadable. Please at least use parenthesis to clarify the meaning. 

Original code was bad, but this is just esoteric.

crsib (approved):
I don't want to block this PR, but I sincerely hope that there are plans to remove code like `(*mOutputTracks->Any<WaveTrack>().begin())->Channels().begin()` repeated all over the place

Paul-Licameli:
> mOutputTracks

Yes there is.

Some stuff is rewritten so that the result of `WideEmptyCopy` and other functions is a single object.

Those objects can't yet be wide wave tracks.  Transitionally, they are track lists.  This isn't the first commit to introduce this sort of thing.

When wide wave tracks are real, the temporary track list objects won't be needed.

src/effects/nyquist/Nyquist.cpp
Paul-Licameli: I'll make it `(* ++ channels.first).get()` and similar wherever I wrote `&**`

Paul-Licameli:
Rebased to recent master with the suggested style changes.

Paul-Licameli:
Rebased onto recent master, and a conflict resolved in StereoToMono.cpp

chinakov:
Error message while 'getting noise profile' in Noise Reduction effect. 
Screenshot 2023-08-16 at 14 51 59

vsverchinsky (approved):

vsverchinsky (approved):

Paul-Licameli:
Rebased again without changes onto recent master

Paul-Licameli:
Rebased unchanged onto recent master

Paul-Licameli:
> I've found a new issue that I couldn't reproduce on release 3.3.3 and latest master. If the User uses the 'Split Stereo to Mono' in the tracks context menu, the tracks are split, but the last (right) channel no longer has it's original audio clip name.
> 
> Here is how it should be **(Master, 3.3.3)** after splitting a stereo track to mono track via 'Split Stereo to Mono': Screenshot 2023-08-21 at 17 32 32
> 
> Here is how it **actually** looks in this branch: Screenshot 2023-08-21 at 17 33 32

I don't observe this for newly created tracks.  I think it will only happen for imported tracks, or for clips that are made by the Repeat effect.  I can fix it by just reverting one of the commits.

Paul-Licameli:
Rebased onto #5060, permuted the assertion fix commit to first place, and reverted one of the other commits

chinakov:
Checklist is complete, all issues fixed.

Issue #4850, opened 2023-06-28
Audio clip stretching


saintmatthieu:
As opposed to now, changing project tempo must modify the clip lengths in the time view but not the beats-and-measure view.
Also, the stretch handles (todo refer to issue) should affect stretch ratio and hence clip length (in both views).
The painting of the waveform and spectrum must look convincing.

PR #5040, opened 2023-08-15
Time stretching 1 of 6 allow outward drag for stretch handles

saintmatthieu:
Resolves: #4850 

QA:
- [x] Test smart clip dragging (hidden data at ends)
- [x] Test that trim handles turn into stretch handles when pressing ALT

Small intentional behavior change:
- [x] dragging of smart clip boundaries allows 2 samples minimum, not 1 as before

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -614,12 +614,12 @@ void WaveTrack::SetWaveColorIndex(int colorIndex)
    }
 }
 
-sampleCount WaveTrack::GetPlaySamplesCount() const
+sampleCount WaveTrack::GetVisibleSampleCount() const
 {
     sampleCount result{ 0 };
 
     for (const auto& clip : mClips)
-        result += clip->GetPlaySamplesCount();
+        result += clip->GetVisibleSampleCount();
 
     return result;
 }
@@ -2348,7 +2348,7 @@ bool WaveTrack::GetOne(
       {
          // Clip sample region and Get/Put sample region overlap
          auto samplesToCopy =
-            std::min( start+len - clipStart, clip->GetPlaySamplesCount() );
+            std::min( start+len - clipStart, clip->GetVisibleSampleCount() );
          auto startDelta = clipStart - start;
          decltype(startDelta) inclipDelta = 0;
          if (startDelta < 0)
@@ -2447,7 +2447,7 @@ ChannelSampleView WaveTrack::GetOneSampleView(
       const auto clipT0 = t0 - clipStartTime;
       const auto clipS0 = TimeToLongSamples(clipT0);
       const auto len =
-         limitSampleBufferSize(length, clip->GetPlaySamplesCount() - clipS0);
+         limitSampleBufferSize(length, clip->GetVisibleSampleCount() - clipS0);
       auto newSegment = clip->GetSampleView(0u, clipS0, len);
       t0 += newSegment.GetSampleCount().as_double() / GetRate();
       segments.push_back(std::move(newSegment));
@@ -2471,7 +2471,7 @@ void WaveTrack::Set(constSamplePtr buffer, sampleFormat format,
       {
          // Clip sample region and Get/Put sample region overlap
          auto samplesToCopy =
-            std::min( start+len - clipStart, clip->GetPlaySamplesCount() );
+            std::min( start+len - clipStart, clip->GetVisibleSampleCount() );
          auto startDelta = clipStart - start;
          decltype(startDelta) inclipDelta = 0;
          if (startDelta < 0)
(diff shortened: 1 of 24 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2596,10 +2596,62 @@ void WaveTrack::GetEnvelopeValues(
...
+                   1e-9 ?
+                neighbour :
+                nullptr;
+}
+
+WaveClip*
+WaveTrack::GetAdjacentClip(const WaveClip& clip, PlaybackDirection direction)
+{
+   return const_cast<WaveClip*>(
+      const_cast<const WaveTrack&>(*this).GetAdjacentClip(clip, direction));
```
Paul-Licameli, line 2644: Can use `as_const` once more

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -398,8 +399,36 @@ class WAVE_TRACK_API WaveTrack final
...
+   const WaveClip* GetAdjacentClip(
+      const WaveClip& clip, PlaybackDirection searchDirection) const;
+   /*!
+    * @copydoc GetAdjacentClip(const WaveClip&, PlaybackDirection) const
+    */
+   WaveClip*
+   GetAdjacentClip(const WaveClip& clip, PlaybackDirection searchDirection);
+
+   WaveClipConstHolders GetOverlappingClips(double t0, double t1) const;
+   WaveClipHolders GetOverlappingClips(double t0, double t1);
```
Paul-Licameli, line 430: Declared but not defined in any of the six PRs.

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -75,19 +75,59 @@ namespace {
...
+   bool isStretchMode)
+{
+   if (!adjustingLeftBorder)
+      // A clip of one sample is invisible - who needs that ? Leave at least 2.
+      return clip.GetPlayStartTime() + 2.0 / clip.GetRate();
+
+   const auto prevClip = track.GetNextClip(clip, PlaybackDirection::backward);
+   if (isStretchMode)
+      return prevClip ? prevClip->GetPlayEndTime() :
+                        std::numeric_limits<double>::min();
```
Paul-Licameli, line 103: Beware the difference of `min` and `lowest` for floating-point types!

Paul-Licameli (changes requested):

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -75,19 +75,59 @@ namespace {
...
 WaveClipAdjustBorderHandle::AdjustPolicy::~AdjustPolicy() = default;
 
+namespace {
+double GetLeftAdjustLimit(
+   const WaveClip& clip, const WaveTrack& track, bool adjustingLeftBorder,
+   bool isStretchMode)
+{
+   if (!adjustingLeftBorder)
+      // A clip of one sample is invisible - who needs that ? Leave at least 2.
+      return clip.GetPlayStartTime() + 2.0 / clip.GetRate();
```
Paul-Licameli, line 98: Was this little change from a 1 sample minimum to 2 asked for by anyone?

Are you sure it always works as intended?  Maybe somewhere there is some roundoff of the inexact sample time to a discrete sample count that goes off by one.

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Ok, then it's not used. I'll remove. Thanks.

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
saintmatthieu: No, it wasn't, but in current release, if you trim so far that the visible part is only one sample and you release the trim handle, you cannot trim back (unless you Undo.) Let's ask @LWinterberg.

I think it should always work. The clip's start time is always displayed on a sample position.

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
saintmatthieu: Great catch, thank you!

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
Paul-Licameli: Please elaborate the comment I added at the top for the QA checklist.

Paul-Licameli (approved):

dozzzzer:
Tested

PR #5041, opened 2023-08-15
Time stretching 2 of 6 make audio track stretching effective

### LINKED ISSUE #4850: Audio clip stretching

saintmatthieu:
Resolves: #4850

This PR enables the time-stretching of audio clips!
Changing the project tempo _will_ play the clips slower or faster. Clips can also individually be stretched using the stretch handle (like trimming, but with alt depressed).

Depends on
- #5040
- #5095

Recommended:

## QA:
This is a big change, with many consequences, please plan some time to play around and do all sorts of crazy things with it. I'd be surprised if you don't find bugs, other than those underneath that will get fixed in follow-up PRs:
* drawing tools won't work on stretched clips
* copy/paste/join actions involving stretched clips
* applying effects on stretched clips.
* generating sounds on stretched clips

Checklist:
- [x] No regression of the fix for this bug https://bugzilla.audacityteam.org/show_bug.cgi?id=2288
- [x] Playback works in all directions (also backward by scrubbing)
- [x] Various spectrum displays (standard, autocorrelation, reassignment, ...)
- [x] Reassignment algorithm of spectrograms especially -- stretch clip while in that view, see that the display updates properly.  Use an example where the difference between Reassignment and the default algorithm is noticeable.
- [x] Playback of stretched clips
- [x] Scrubbing including backward and forwards, and at ends of stretched clips
- [x] Correct playback with an envelope in a stretched clip (edit the envelope before you stretch)
- [x] Contrast tool:
  - Make two clips on a track with different stretch
  - Select all, note RMS values
  - Mix and render to new track
  - Repeat, observe the same RMS
 - [x] Playback stops correctly when it is started from a position other than 0.
 - [x] Persistency of stretches in the saved project file
 - [x] Rendering and exporting of stretches
 - [x] Rebase onto #5095 and confirm that the problem does not recur
 - [x] Change project tempo with a stereo track present, then split it to mono, observe correct stretch in new track
 - [x] Undo and Redo of changes to project tempo with all types of tracks present

## Issues  that need to be addressed within this PR:

- [x] Deadlock when using "At Zero Crossings"
- [x] Stretched clips are quieter
- [x] The right channel of a stereo track seems to ignore the envelope points
- [x] Beat Finder fails when analyzing a stretched (shrunk) clip
- [x] Beat Finder may cause deadlock
- [x] Insignificant lag when playing a stretched clip not from the zero mark
- [x] Deadlock when playing extremely stretched-out clips
- [x] Poor processing of synthetic tones resulting in high peaks (TBD)
- [x] Lag when playing a clip shrunk to extreme minimum


libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1159,27 +1206,52 @@ void WaveClip::ShiftBy(double delta) noexcept
...
 }
 
-// Bug 2288 allowed overlapping clips.
-// This was a classic fencepost error.
-// We are within the clip if start < t <= end.
-// Note that BeforeClip and AfterClip must be consistent
-// with this definition.
+bool WaveClip::SplitsPlayRegion(double t) const
+{
+   return GetPlayStartTime() < t && t < GetPlayEndTime();
```
Paul-Licameli, line 1275: Should this comparison use discrete sample counts?

libraries/lib-wave-track/WaveClip.h
```diff
@@ -211,12 +216,47 @@ class WAVE_TRACK_API WaveClip final :
...
+   /*!
+    * @brief  t < [
+    */
+   bool BeforePlayRegion(double t) const;
+   /*!
+    * @brief  ) <= t
+    */
+   bool AfterPlayRegion(double t) const;
+   /*!
+    * @brief t0 and t1 both ∈ [...)
```
Paul-Licameli, line 245: Non-ascii mathematical notation in comments, nice

libraries/lib-wave-track/WaveClip.h
```diff
@@ -471,10 +512,18 @@ class WAVE_TRACK_API WaveClip final :
       bool committed{ false };
    };
 
+   // Real-time durations, i.e., stretching the clip modifies these.
```
Paul-Licameli, line 520: See uses of `@name` with `@{` and `@}` elsewhere in the source tree, so that Doxygen can associate a comment with multiple class members.

libraries/lib-wave-track/WaveClip.h
```diff
@@ -471,10 +512,18 @@ class WAVE_TRACK_API WaveClip final :
...
    double mTrimLeft{ 0 };
    double mTrimRight{ 0 };
 
+   // Used in GetStretchRatio which computes the factor, by which the sample
+   // interval is multiplied, to get a realtime duration.
+   double mClipStretchRatio = 1.;
+   std::optional<double> mRawAudioTempo;
+   std::optional<double> mProjectTempo;
+
+   // Sample rate of the raw audio, i.e., before stretching.
```
Paul-Licameli, line 531: Here you can simply use `//!` for a Doxygen comment on one member.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2187,6 +2186,12 @@ double WaveTrack::GetEndTime() const
    return ChannelGroup::GetEndTime();
 }
 
+double WaveTrack::SnapToSample(double t) const
+{
+   const auto sampleRate = GetRate();
+   return std::round(t * sampleRate) / sampleRate;
```
Paul-Licameli, line 2192: I think this function should use `TimeToLongSamples`

Paul-Licameli: No, just delete the function!  It is defined but not used anywhere in these six PRs

libraries/lib-wave-track/WaveClip.h
```diff
@@ -444,21 +486,20 @@ class WAVE_TRACK_API WaveClip final :
...
    SampleFormats GetSampleFormats() const;
    const SampleBlockFactoryPtr &GetFactory();
+   double SnapToTrackSample(double time) const noexcept;
 
    /// This name is consistent with WaveTrack::Clear. It performs a "Cut"
    /// operation (but without putting the cut audio to the clipboard)
    void ClearSequence(double t0, double t1);
 
    //! Restores state when an update loop over mSequences fails midway
+   // todo(mhodgkinson) I don't find any use. Throw away?
```
Paul-Licameli: Delete this comment please

libraries/lib-wave-track/WaveClip.h
saintmatthieu: Isn't it too much to have private members show up in Doxygen?

libraries/lib-wave-track/WaveClip.h
saintmatthieu: It's deleted.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: It is used, in `WaveTrack::ApplyStretchRatio`, in PR 5. I'll remove it from this PR, and rebasing will force me to re-introduce it. Then I'll use `TimeToLongSamples`.

libraries/lib-wave-track/WaveClip.h
Paul-Licameli: All might help someone needing to understand the implementation, not just interface.

There is the notion of “representation invariant” for instance — invariant conditions involving private member variables.  Might as well follow conventions with `@invariant` There too

Paul-Licameli:
I just bisected a bug to the commit "Make waveform display work again", but don't fully understand it yet.

- empty project
- Make stereo track
- select some range of time later than 0
- Generate > noise

Observe:  the channels appear misaligned.

Paul-Licameli:
That appears to be only a display problem.  Real playback is correct.

saintmatthieu:
> I just bisected a bug to the commit "Make waveform display work again", but don't fully understand it yet.
> 
> * empty project
> * Make stereo track
> * select some range of time later than 0
> * Generate > noise
> 
> Observe: the channels appear misaligned.

@Paul-Licameli I cannot reproduce it. If you could figure it out that'd be great.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -573,7 +573,7 @@ bool WaveClipSpectrumCache::GetSpectrogram(const WaveClip &clip,
    // purposely offset the display 1/2 sample to the left (as compared
    // to waveform display) to properly center response of the FFT
    fillWhere(mSpecCache->where, numPixels, 0.5, correction,
```
Paul-Licameli: I think I wrote this elsewhere, bug again:  `0.5` should change to `true` at the commit "Make waveform display work again" to match the change of argument type.

src/tracks/playabletrack/wavetrack/ui/WaveClipUtilities.cpp
```diff
@@ -29,7 +30,7 @@ void findCorrection(const std::vector<sampleCount> &oldWhere, size_t oldLen,
    const double denom = oldWhereLast - oldWhere0;
 
    // What sample would go in where[0] with no correction?
-   const double guessWhere0 = t0 * rate;
+   const double guessWhere0 = t0 * sampleRate / stretchRatio;
```
Paul-Licameli, line 33: near the end of this function `findCorrection`, clean up two things please:  use `std::clamp` and `assert`

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -456,7 +456,7 @@ void DrawIndividualSamples(TrackPanelDrawingContext &context, size_t channel,
    const auto &zoomInfo = *artist->pZoomInfo;
 
    const double toffset = clip->GetPlayStartTime();
-   double rate = clip->GetRate();
+   const double rate = clip->GetRate() / clip->GetStretchRatio();
```
Paul-Licameli, line 459: I see this change in `DrawIndividualSamples`.  I wonder whether it's complete.  I see another use of `clip->GetRate()` that is not corrected for stretch later.  But it's only fetching one value from the envelope and passing a time-step value for successive samples that doesn't matter and could be 0.

Paul-Licameli: The domain of the Envelope (viewed as a curve or function) is always adjusted to be real time, correct?  So `time` given to `GetValue` is right?

Did you test display of individual samples with an envelope too?

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -687,6 +687,7 @@ void DrawClipWaveform(TrackPanelDrawingContext &context, size_t channel,
    const double &t1 = params.t1;
    const double &averagePixelsPerSample = params.averagePixelsPerSample;
    const double &rate = params.rate;
+   const double &stretchRatio = 1.0;
```
Paul-Licameli: As I wrote on the closed PR, the commit "Make waveform display work again" doesn't completely do what it says.  Instead the next commit "Spectrum display..." completes the work for waveform display too.

So properly, these two commits should be squashed into one (plus the fixup I pushed).  Maybe that commit would be too large and you might find a nice way to separate small steps like changes of arguments of functions or construction of `ClipParameters`

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.h
```diff
@@ -39,36 +39,35 @@ class AUDACITY_DLL_API SpecCache {
...
-      (const SpectrogramSettings &settings,
-       const WideSampleSequence &sequence,
-       const int xx, sampleCount numSamples,
-       double offset, double rate, double pixelsPerSecond,
-       int lowerBoundX, int upperBoundX,
-       const std::vector<float> &gainFactors,
-       float* __restrict scratch,
-       float* __restrict out) const;
+   // This assumes that all clips have equal stretch ratio, i.e., no stretch
+   // handle is used, and all clips were added at a given project tempo.
```
Paul-Licameli, line 48: Is this comment in commit "Spectrum Display..." correct?

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -394,6 +477,21 @@ bool WaveClip::HandleXMLTag(const std::string_view& tag, const AttributesList &a
                return false;
             SetTrimRight(dblValue);
          }
+         else if (attr == "rawAudioTempo")
+         {
+            if (!value.TryGet(dblValue))
+               return false;
+            if (dblValue == 0)
+               mRawAudioTempo.reset();
```
Paul-Licameli, line 485: If 0 is an excluded value, maybe the member doesn't need to be `std::optional` but instead you would test the value wherever it is used.

libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -42,10 +42,9 @@ sampleCount GetLastReadSample(
 sampleCount
 GetTotalNumSamplesToProduce(const ClipInterface& clip, double durationToDiscard)
 {
-   return sampleCount { (clip.GetVisibleSampleCount().as_double() -
-                         durationToDiscard * clip.GetRate()) *
-                           clip.GetStretchRatio() +
-                        .5 };
+   return sampleCount { clip.GetVisibleSampleCount().as_double() *
+                           clip.GetStretchRatio() -
+                        durationToDiscard * clip.GetRate() + .5 };
```
Paul-Licameli, line 47: I spent a surprising amount of time studying this small commit to convince myself it was correct.  Rereading the whole class led me to push one more commit for an possible undiscovered problem.  I think backwards scrubbing with stretch applied might be wrong.

Paul-Licameli: I found this small commit difficult to understand because `durationToDiscard` is multiplied by the clip's rate only in one place, but then also divided by stretch ratio in another place.

I think the members `mLastReadSample` and `mTotalNumSamplesToProduce` should be commented.

`mLastReadSample` counts raw samples but `mTotalNumSamplesToProduce` counts stretched.

Also, what should have made the bug "obvious" in earlier reviews -- is that the clip's rate should always be divided by the stretch ratio, never multiplied by it.

saintmatthieu: I considered some time ago already providing a class only for the implementation of `TimeAndPitchSource` and let `ClipSegment` use it. It could have avoided this confusion. Now `mLastReadSample` is a member of the source, and represents a sample index before stretching. `mTotalNumSamplesToProduce` is a member of `ClipSegment` and represents a sample count after stretching.

saintmatthieu: Now I did it. The commit "review comments: extract TimeAndPitchSource from ClipSegment ..." only just does that - no change in logic.

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: One advantage of `optional`: the compiler will tell you if you are not de-referencing when you should. A bare double won't warn against unchecked uses.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.h
saintmatthieu: It's outdated, stemming from the time when spectrum display did not account for stretching. Thanks.

libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -94,17 +94,16 @@ void ClipSegment::Pull(float* const* buffers, size_t samplesPerChannel)
       constexpr auto mayThrow = false;
       const auto start =
          forward ? mLastReadSample : mLastReadSample - numSamplesToRead;
-      const auto offset = forward ? 0u : samplesPerChannel - numSamplesToRead;
```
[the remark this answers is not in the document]
Paul-Licameli: I tried to demonstrate backwards scrubbing going wrong without this fix, but I could not.

Maybe `samplesPerChannel` is never greater than `remainingSamplesInClip` because of other tests elsewhere, but the class `ClipSegment` should not depend on that pattern of usage.

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
saintmatthieu: Right. I did not a good job with my interactive rebase for this part. I may mistakenly have attempted to re-order and squash the "WaveChannelView clean-up ..." into "Make waveform display work again", but actually it should have been squashed into "Spectrum display ..." (and then squashing that into "Make waveform display work again" as you suggested). I don't mind doing it, but is now a good time, since you've reviewed this already?

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
saintmatthieu: > The domain of the Envelope (viewed as a curve or function) is always updated with ration changes to be real time, correct?

Correct. The x-position of the samples is scaled by the stretch ratio because the audio samples are stored in stretched state. The envelope is, though.

> Did you test display of individual samples with an envelope too?

I did, yes, and that works to the extent I tested it. Not so confident about the corner cases, though.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -71,16 +71,14 @@ void ComputeSpectrogramGainFactors
...
-   (int dirty_, double pixelsPerSecond,
-    const SpectrogramSettings &settings, double rate) const
+bool SpecCache::Matches(
+   int dirty_, double samplesPerPixel,
+   const SpectrogramSettings& settings) const
 {
    // Make a tolerant comparison of the pps values in this wise:
    // accumulated difference of times over the number of pixels is less than
    // a sample period.
-   const double tstep = 1.0 / pixelsPerSecond;
```
Paul-Licameli, line 81: Change “pps ” in the comment

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -71,16 +71,14 @@ void ComputeSpectrogramGainFactors
...
-   const bool ppsMatch =
-      (fabs(tstep - 1.0 / pps) * len < (1.0 / rate));
+   const bool ppsMatch = (fabs(samplesPerPixel - spp) * len < 1.0);
```
Paul-Licameli, line 81: And the variable name

libraries/lib-stretching-sequence/AudioSegmentSampleView.cpp
```diff
@@ -56,3 +56,16 @@ size_t AudioSegmentSampleView::DoCopy(float* buffer, size_t bufferSize) const
    }
    return written;
 }
+
+void FillBufferFromTrackBlockSequence(
+   const ChannelSampleView& track, float* buffer, size_t bufferSize)
+{
+   size_t written = 0;
+   for (const auto& segment : track)
```
Paul-Licameli, line 64: This moved function is now more general than it needs to be.  The first argument is always a size one vector at the sole place where you call it.  Maybe you feel like cleaning that up?

src/tracks/playabletrack/wavetrack/ui/SpectrumView.cpp
```diff
@@ -201,8 +200,8 @@ void SpectrumView::DoSetMinimized( bool minimized )
    if( bHalfWave && minimized)
    {
       // It is all right to set the top of scale to a huge number,
-      // not knowing the track rate here -- because when retrieving the
-      // value, then we pass in a sample rate and clamp it above to the
+      // not knowing the track sampleRate here -- because when retrieving the
+      // value, then we pass in a sample sampleRate and clamp it above to the
```
Paul-Licameli, line 204: Your search-and-replace made the comment read strangely

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
```diff
@@ -1524,72 +1524,68 @@ std::shared_ptr<ChannelVRulerControls> WaveChannelView::DoGetVRulerControls()
...
+}
 
-      if(outShowSamples)
-         // adjustment so that the last circular point doesn't appear
-         // to be hanging off the end
-         return  pixelsOffset / (outAveragePPS * rate); // pixels / ( pixels / second ) = seconds
-      return .0;
-   }
+bool ShowIndividualSamples(
+   const wxRect& viewRect, const ZoomInfo& zoomInfo, int sampleRate,
```
Paul-Licameli, line 1553: First two arguments are unused

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
```diff
@@ -1683,35 +1664,34 @@ ClipParameters::ClipParameters
...
+   const auto pixelsPerSecond = GetPixelsPerSecond(viewRect, zoomInfo);
+   const auto showIndividualSamples = ShowIndividualSamples(
+      viewRect, zoomInfo, clip.GetRate(), clip.GetStretchRatio(),
+      pixelsPerSecond);
+   const auto clipEndingAdjustemt =
+      CalculateAdjustmentForZoomLevel(pixelsPerSecond, showIndividualSamples);
+   if (outShowSamples != nullptr)
+      *outShowSamples = showIndividualSamples;
+   constexpr auto edgeLeft =
+      static_cast<ZoomInfo::int64>(std::numeric_limits<int>::min());
```
Paul-Licameli, line 1676: Remember `min` is not `lowest` when the type is floating! -- but that's not the case here

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
```diff
@@ -1683,35 +1664,34 @@ ClipParameters::ClipParameters
...
-           static_cast<int>(left), 
-           viewRect.y, 
-           std::max(1, static_cast<int>(right - left)), 
-           viewRect.height
-        );
+   const auto pixelsPerSecond = GetPixelsPerSecond(viewRect, zoomInfo);
+   const auto showIndividualSamples = ShowIndividualSamples(
+      viewRect, zoomInfo, clip.GetRate(), clip.GetStretchRatio(),
+      pixelsPerSecond);
+   const auto clipEndingAdjustemt =
```
Paul-Licameli, line 1671: Fix the variable name `clipEndingAdjustment`

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
```diff
@@ -1524,72 +1524,68 @@ std::shared_ptr<ChannelVRulerControls> WaveChannelView::DoGetVRulerControls()
...
+   return pixelsPerSample > 0.5;
+}
 }
 
-ClipParameters::ClipParameters
-   (bool spectrum, const SampleTrack *track, const WaveClip *clip, const wxRect &rect,
-   const SelectedRegion &selectedRegion, const ZoomInfo &zoomInfo)
+ClipParameters::ClipParameters(
+   const SampleTrack* track, const WaveClip* clip, const wxRect& rect,
+   const SelectedRegion& selectedRegion, const ZoomInfo& zoomInfo)
```
Paul-Licameli, line 1564: Arguments `track` and `selectedRegion` are not used

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
```diff
@@ -1524,72 +1524,68 @@ std::shared_ptr<ChannelVRulerControls> WaveChannelView::DoGetVRulerControls()
...
 
-   const double sps = 1. / rate;            //seconds-per-sample
+   const auto blank = GetBlankSpaceBeforePlayEndTime(*clip);
 
    // Calculate actual selection bounds so that t0 > 0 and t1 < the
    // end of the track
    t0 = std::max(tpre, .0);
-   t1 = std::min(tpost, trackLen - sps * .99) 
-      + CalculateAdjustmentForZoomLevel(rect, zoomInfo, rate, averagePixelsPerSample, showIndividualSamples);
+   t1 = std::min(tpost, clipLength - blank) +
```
Paul-Licameli, line 1586: `t1` is not used outside this constructor, so it might be a local variable

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -846,14 +840,14 @@ void DrawClipWaveform(TrackPanelDrawingContext &context, size_t channel,
...
-            CommonChannelView::GetEnvelopeValues(envelope,
-                tOffset,
+            CommonChannelView::GetEnvelopeValues(
+               *clip.GetEnvelope(), playStartTime,
 
-                 // PRL: change back to make envelope evaluate only at sample
-                 // times and then interpolate the display
-                 0, // 1.0 / rate,
+               // PRL: change back to make envelope evaluate only at sample
+               // times and then interpolate the display
```
Paul-Licameli, line 847: Below, change `rate` to `sampleRate` in commented-out code.  Sample rate NOT corrected for stretch ratio would be right because we assume the envelope already stretches its time coordinates.

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -740,19 +737,16 @@ void DrawClipWaveform(TrackPanelDrawingContext &context, size_t channel,
...
    std::vector<WavePortion> portions;
    FindWavePortions(portions, rect, zoomInfo, params);
    const unsigned nPortions = portions.size();
 
    // Require at least 1/2 pixel per sample for drawing individual samples.
-   const double threshold1 = 0.5 * rate;
+   const double threshold1 = 0.5 * sampleRate / stretchRatio;
    // Require at least 3 pixels per sample for drawing the draggable points.
-   const double threshold2 = 3 * rate;
+   const double threshold2 = 3 * sampleRate / stretchRatio;
```
Paul-Licameli, line 749: Or `6 * threshold1`

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -707,14 +704,14 @@ void DrawClipWaveform(TrackPanelDrawingContext &context, size_t channel,
...
-   CommonChannelView::GetEnvelopeValues(envelope,
-       tOffset,
+   CommonChannelView::GetEnvelopeValues(
+      *clip.GetEnvelope(), playStartTime,
 
-        // PRL: change back to make envelope evaluate only at sample times
-        // and then interpolate the display
-        0, // 1.0 / rate,
+      // PRL: change back to make envelope evaluate only at sample times
+      // and then interpolate the display
```
Paul-Licameli, line 711: Below, likewise, change commented-out code

src/tracks/playabletrack/wavetrack/ui/SpectrumView.cpp
```diff
@@ -321,19 +320,31 @@ ChooseColorSet( float bin0, float bin1, float selBinLo,
...
 
-void DrawClipSpectrum(TrackPanelDrawingContext &context,
-                                   const WideSampleSequence &sequence,
-                                   const WaveTrack* track,
-                                   const WaveClip *clip,
-                                   const wxRect &rect,
-                                   const std::shared_ptr<SpectralData> &mpSpectralData,
-                                   bool selected)
+std::pair<sampleCount, sampleCount> GetSelectedSampleIndices(
+   const ClipParameters& params, const SelectedRegion& selectedRegion,
```
Paul-Licameli, line 324: `params` is unused

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -108,18 +103,19 @@ bool SpecCache::CalculateOneSpectrum
 
    sampleCount from;
 
+   const auto offset = clip.GetSequenceStartTime();
```
Paul-Licameli, line 106: unused

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -108,18 +103,19 @@ bool SpecCache::CalculateOneSpectrum
...
+   const auto numSamples = clip.GetSequenceSamplesCount() / clip.GetWidth();
```
Paul-Licameli, line 107: Maybe the first real bug I have found:

While dividing by `GetWidth` may correct a mistake introduced earlier -- yet I look at the uses of `SpecCache::Populate` and I see one (in DrawClipSpectrum) passed the visible sample count, while the other (in this file) passed the sequence sample count.  But now both use the sequence count.

Why is this right?

Paul-Licameli (changes requested):
One doubtful thing

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -171,17 +167,12 @@ bool SpecCache::CalculateOneSpectrum
...
+            mSampleCacheHolder.emplace(clip.GetSampleView(iChannel, from, myLen));
             floats.resize(myLen);
-            auto data = floats.data();
-            constexpr auto backwards = false;
-            if (sequence.GetFloats(
-                   0u, 1u, &data,
-                   sampleCount(floor(0.5 + from.as_double() + offset * rate)),
-                   myLen, backwards, FillFormat::fillZero,
-                   // Don't throw in this drawing operation
-                   false))
```
Paul-Licameli, line 182: Notice how `GetFloats` had a `mayThrow` argument defaulting true but passed false here.

I added that years ago.  I reasoned that if reading sample blocks failed for any reason, but only in display code, it should be non-throwing, and instead just zeroes should be substituted.

But now that is lost with this transformation.

See where `SampleBlock.cpp` used to use the `mayThrow` argument.

But that was a guard around the concrete class that implemented `DoGetSamples`.

So I think `SqliteSampleBlock::GetFloatSampleView` should now add a try-catch around its new call to `DoGetSamples` and the catch block needs to add zeroes.

So that the behavior in exceptional cases does NOT change.

Paul-Licameli (changes requested):
Preserve behavior in presence of exceptions thrown from DB layer

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
saintmatthieu: At last I know what the `mayThrow` is about - my bad not for having actively searched its meaning.
`GetFloatSampleView` is used for both display and playback, though, so adding a try-catch would override the current playback, throwing behaviour.
It's not too much work to add this boolean to `GetFloatSampleView`. Otherwise we'd have to choose one behaviour for both use cases - always catching would also affect playback behaviour.
Added a commit that does this.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
saintmatthieu: Thanks for this question, it motivated me to (try and) understand this whole code better.

I think it is okay the way it is. `numSamples` will be used to know whether the time-domain input to the spectrogram needs zero-padding, if the window goes beyond that. Providing the sequence sample count, as opposed to the _visible_ sample count, allows the code to fetch samples beyond the visible boundary to draw the spectrum, which is good.

The other call used to use `GetVisibleSampleCount`, and leading to unnecessary zero-padding in case of right trimming. Note, though, that that other call only gets executed if the fisheye state is `PINNED`, but it never is, as it's hard-coded to `HIDDEN`, so I cannot test that it indeed works with these changes.

What I could test, though, is that an impulse, whether at the beginning or end of the clip, has its spectrogram displayed properly, also with left- and/or right-trimming.

src/tracks/playabletrack/wavetrack/ui/SpectrumView.cpp
saintmatthieu: ty

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
saintmatthieu: ty

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
saintmatthieu: How could I miss that ... thank you.

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
saintmatthieu: Haha, "adjustemt"

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
Paul-Licameli: I'll let your decisions stand on this, but maybe it should be reexamined later.  Maybe a clip with unusually large hidden data will cause caching to do too much work and even defeat the purpose of caching for performance.  Maybe there should be a smarter way to bound the cache size, more than the visible samples but fewer than the whole.

libraries/lib-stretching-sequence/AudioSegmentSampleView.cpp
```diff
@@ -56,16 +56,3 @@ size_t AudioSegmentSampleView::DoCopy(float* buffer, size_t bufferSize) const
...
-   const ChannelSampleView& track, float* buffer, size_t bufferSize)
-{
-   size_t written = 0;
-   for (const auto& segment : track)
-   {
-      size_t toWrite =
-         limitSampleBufferSize(bufferSize - written, segment.GetSampleCount());
-      written += segment.Copy(buffer + written, toWrite);
-   }
-   std::fill(buffer + written, buffer + bufferSize, 0.f);
```
Paul-Licameli: Now there is only a direct call to `Copy`.  The limiting of sample buffer size and the filling of extra space with zeroes don't happen.  Should they be restored?

Paul-Licameli (changes requested):
Satisfy my doubt

Paul-Licameli:
> `t1` is not used outside this constructor, so it might be a local variable

You might do this still, or not

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -277,7 +268,10 @@ bool SpecCache::CalculateOneSpectrum
                      (numRe * denomRe + numIm * denomIm) / power;
                }
 
-               int correctedX = (floor(0.5 + xx + timeCorrection * pixelsPerSecond / rate));
+               // todo(mhodgkinson) double-check, but I think here we need a raw
+               // sample index - stretchRatio not involved.
```
Paul-Licameli, line 272: You are right.  `timeCorrection` is scaled to the clip's raw sample rate, without the stretching ratio correction for real time.  We want to find the correct X coordinate for that.

So remove "todo" and explain in the comment more confidently.

Paul-Licameli: Reference #2 on this page was my guide to implement this algorithm long ago.  https://en.wikipedia.org/wiki/Reassignment_method#References

I used to be able to find freely available pdfs

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -536,7 +517,8 @@ bool WaveClipSpectrumCache::GetSpectrogram(const WaveClip &clip,
    }
 
    // Resize the cache, keep the contents unchanged.
-   mSpecCache->Grow(numPixels, settings, pixelsPerSecond, t0);
+   mSpecCache->Grow(
+      numPixels, settings, samplesPerPixel, t0 /*Grow takes absolute t0*/);
```
Paul-Licameli, line 521: Almost completing review!  The last thing I must understand is why you changed arguments in the other call of `Grow` but not this one.  Are you correcting an old error?

But I examined things and I find this added comment misleading.  `t0` is, and was before the changes, a time relative to clip start.  Not absolute.  And relative time is in fact what `Grow` was expecting.  The `t0` value is stored in `start` and ultimately used only in `GetSpectrogram` to check for a cache hit.  The `t0` in the comparison is the same `t0` used here.

Paul-Licameli: `GetSpectrogram` is called in only one place, and the `t0` value supplied there is from `ClipParameters::t0`, and that value is initialized with a duration relative to clip play start, not an absolute time.

Paul-Licameli (changes requested):
Mistake

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
saintmatthieu: You have an IEEE subscription ?? :star-struck:

saintmatthieu: Copy-pasting your explanation as comment here, tagging it with your initials.

Paul-Licameli: > You have an IEEE subscription ?? :star-struck:

Actually no

Paul-Licameli:
Approved!  Please squash all commits after 560bc7dfcb903c393e03f2580fedb33b9ea1be9b into one.

Paul-Licameli (dismissed):

Paul-Licameli:
I have combined the QA checklists from two PRs.

dozzzzer:
Found issues @saintmatthieu :

- When time-stretching from the left side (the beginning) of the clip, then dragging the new, time-stretched clip back to the zero time mark and playing the clip, it freezes for a few seconds before starting the playback. The subsequent attempts to play resulted in no stuttering. It also seems that the higher the stretch ratio is, the longer it waits before actually playing the track:

https://github.com/audacity/audacity/assets/70063970/06425b5a-636f-441c-b3dd-f273a38ea948

- When shrinking a 1-minute clip to the minimum length, playback also starts after a short pause a couple of seconds long. Also, when shrinking a 1-minute mono clip to the minimum length, duplicating it, making it into a stereo track, and hitting Play, the pause time seems to double in time
- Deadlock when placing two clips with different stretch ratios next to each other and then pressing Z on the keyboard (= at zero crossings):
https://github.com/audacity/audacity/assets/70063970/ee167601-9884-41ac-8976-0c5de569a911

- When mixing-and-rendering a time-stretched clip to a new track and then measuring the contrast (_Analyze > Contrast_) on both, the mixed-and-rendered one always appears to be a few dB quieter than the original time-stretched one:
https://github.com/audacity/audacity/assets/70063970/2119605b-d09a-414c-abe4-7d558b375b47

(to be continued)

saintmatthieu:
@dozzzzer Thanks a lot for your findings so far.

> When time-stretching from the left side ...

This is already in 3.3.3 😬 Please try the exact same on the current release, using trimming and stretching.
The reason : before trimming, you select the entire clip. During trimming, the selection stays the same. When you shift the clip left, it also shifts the selection to the left, past the zero mark. And when you press play, it plays the selecting region, i.e. starting at second -10 ...

IMO this _is_ a strange behaviour, that _could_ be acceptable if the option to see past the zero mark is checked, but probably not if it isn't, right? Hasn't anyone complained about that so far? @LWinterberg?

saintmatthieu:
@dozzzzer 
> When shrinking a 1-minute clip to the minimum length, playback also starts after a short pause a couple of seconds long. Also, when shrinking a 1-minute mono clip to the minimum length, duplicating it, making it into a stereo track, and hitting Play, the pause time seems to double in time

StaffPad's time stretcher doesn't seem to like the way we configure it and extremely low stretch ratios. It may be a simple fix. WIP.

dozzzzer:
> This is already in 3.3.3

👌 Logged as #5088

Paul-Licameli:
Perhaps I should examine the problem with envelope display.  Recently I merged to master some changes in envelopes.  Are you sure nothing was wrong already in master?

dozzzzer:
> Perhaps I should examine the problem with envelope display. Recently I merged to master some changes in envelopes. Are you sure nothing was wrong already in master?

Yes, current `master` isn't affected, i.e. envelopes of both channels of a stereo track can be edited simultaneously.

src/menus/SelectMenus.cpp
```diff
@@ -618,6 +625,37 @@ void OnZeroCrossing(const CommandContext &context)
...
+      };
+   const auto selected = tracks.Selected<const WaveTrack>();
+   if (std::any_of(
+          selected.begin(), selected.end(), [&](const WaveTrack* track) {
+             return wouldSearchStretchedClip(*track, selectedRegion.t0()) ||
+                    wouldSearchStretchedClip(*track, selectedRegion.t1());
+          }))
+   {
+      AudacityMessageBox(
+         XO("Zero-crossing search regions intersect stretched clips."),
```
saintmatthieu, line 655: @LWinterberg I need your benediction :)

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -498,6 +496,8 @@ void TimeAndPitch::feedAudio(const float* const* input_smp, int numSamples)
       if (_analysis_hop_counter >= hop_a)
       {
          _analysis_hop_counter -= hop_a;
+         d->hop_s_err += d->exact_hop_s - hop_s;
+         d->hop_a_err += d->exact_hop_a - hop_a;
```
Paul-Licameli, line 500: I don't feel competent to review this without much more time for study of the library.  I'll just ask:
- does this affect performance only or does it change the calculations?
- The reassignment of `hop_s_err` and `hop_a_err` used to happen once per function call.  Do you mean to make it happen at most once, or sometimes multiple times?
- If not multiple times, then should `hop_s` and `hop_a` be changed after these two assignments?
-

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
```diff
@@ -79,11 +72,6 @@ void StaffPadTimeAndPitch::GetSamples(float* const* output, size_t outputLen)
          }
          numOutputSamplesAvailable =
             mTimeAndPitch->getNumAvailableOutputSamples();
-         if (!firstCall)
-         {
-            break;
-         }
-         firstCall = false;
```
Paul-Licameli, line 86: Again, I'll just trust this change without digging deep

Paul-Licameli:
I re-approve this branch with the new commits, to the extent I understand it (this excludes changes inside the StaffPad library)

saintmatthieu:
> When mixing-and-rendering a time-stretched clip to a new track and then measuring the contrast (Analyze > Contrast) on both, the mixed-and-rendered one always appears to be a few dB quieter than the original time-stretched one

Time stretching will modify the samples, and so we cannot expect the exact same contrast before and after rendering.

Then the follow-up question is: should we display the value for the raw data (as now) or the stretching-rendered result ? I'm not sure that the former would be preferrable. Something similar happens with real-time effects: select e.g. Bass & Treble, reducing both bands by several dBs. The contrast-command result is that of the raw audio. If you mix and render, and run contrast again, you get a different result.

So I'd leave it as it is. @LWinterberg ?

saintmatthieu:
That's correct, modifying the phases of the components of a signal doesn't change its RMS.

Also, I shared pictures on Discord of the effect of the time stretcher on a square wave. The waveform is greatly affected, but the RMS stays the same - roughly. Looking closer you can see it oscillating. That could be one reason.

The other reason could be delay. SP's stretcher reports a delay value that should be reviewed. We discard that many samples at the beginning of the output, but I've seen, for strong stretch ratios, signs that this may be an underestimate or overestimate, depending on the stretch direction. This could lead to discarded samples, affecting the analysis result - and other more critical problems, like truncated boundaries. Seems not be noticeable at typical ratios and we may get away with it for now, though.

saintmatthieu:
@dozzzzer 
> Also noticed that when playing from the cursor immediately after stretching a clip, playback starts with a slight delay of maybe half a second as opposed to regular playback as well as all subsequent playbacks on the stretched clip that start immediately

I'm not sure how to reproduce that. A quick attempt didn't succeed. The number of tracks, sample rate, length, stretch ratios, all could play a role. Maybe make a project available for me?

saintmatthieu:
@dozzzzer 
> When importing a stereo file (wav/mp3) and adding some envelope points, visually only the left channel seems to be affected, even though in fact both left and right channel amplitudes are affected:

@Paul-Licameli provided a fix with commit 1468879. Just tested it, works fine.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
saintmatthieu: The naming of these variables isn't really descriptive. As I understand this code, `hop_a_err` is the analysis samples "credit". `exact_hop_a`, the "exact analysis hop size", can be a fractional number of samples, but of course, one can only feed discrete samples at a time. So when you need 2.5 samples to perform one FFT, as soon as you get the 3rd sample, all 3 samples are processed, but you have a credit of 0.5 samples. This goes into `hop_a_err`. Same principle for `hop_s_err` but in the IFFT direction.
There were situations (when `numSamples` wasn't the 1024 "default" or the stretch ratio is so low that `exact_hop_s < 1`) where those state variables were updated although no analysis or no synthesis happened. As a result, no samples were ever produced, and we were falling in an endless loop.

Now to your questions:
* no, negligibly
* sometimes multiple times, which is the correct thing to do: update your credit variables if and only if analysis/processing happened

@dozzzzer also tested the other extrem: stretching from 30s to 12h. I also did and these changes do not help. Looking into it.

saintmatthieu:
> Poor processing of synthetic tones, such as chirp. When stretched and then mixed and rendered, it has louder peaks than the original:

When a waveform's components' phases relationship change (which is inherent to pitch-preserving time stretching), the shape of the waveform changes, too. For synthetic, complex waveforms with distinctive shapes such as square, triangle or sawtooth waves, this is very noticeable. It comes with a change of peak values (but RMS should be preserved, at least roughly), which is partly what we observe here.

That said, we've identified with @dozzzzer and @LWinterberg a pattern, whereby compressing a clip lowers its RMS, and expanding it, increases its RMS. For typical, musical usage of time stretching this is probably small enough not to be noticeable, but an improvement here would be welcome, possibly as patch release.

The fact that louder peaks may cause clipping upon playback is something we decided not to do anything about. Clip indicators are there to make the user aware of it. Doing anything ourselves proactively would change characteristics of the waveform in likely unwanted ways. Like @Paul-Licameli said, it's the maths of it - let the user decide how to deal with it.

Paul-Licameli:
> synth

You know that there is a checkmark item in the View menu that paints the track red wherever it clips?  This is useful and I have wished it were the default.

If the stretch algorithm calculates sample values out of the [-1, 1] bounds, still those numbers are preserved if in 32 bit float format, and even persist in the .aup3 file.  It is still possible to normalize or compress or limit the result to get back a track that does not clip.

Experienced users may just have to learn this and be aware.

saintmatthieu:
Yes, Leo showed the clipping markers. Normalizing or even just lowering the gain, and, for display, zooming out vertically on the level ruler.

saintmatthieu:
@dozzzzer Beat finder problems are effect-related, and will be fixed in 5-of-6. Please ignore for now.

saintmatthieu:
@dozzzzer all your boxes are checked. Comments to some of those:

> Stretched clips are quieter

... or louder, depending on the stretch direction (squeezing -> quieter, expanding -> louder). As we agreed together with @LWinterberg , this will probably be subject for a patch release and should not block this release.

> Beat Finder fails when analyzing a stretched (shrunk) clip
> Beat Finder may cause deadlock

Please ignore for now, will be fixed with the PR that fixes destructive effects.

> Insignificant lag when playing a stretched clip not from the zero mark

Could not reproduce this.

> Lag when playing a clip shrunk to extreme minimum

Comes with the pre-processing of clips stretched to large amounts : squeezed and a lot of input has to be analyzed to output just a few samples, expanded and the synthesis will first output lots of near-zero values that must be discarded. This takes time, nothing we can do about it without delving in the StaffPad algorithm and risk breaking things while optimizing. I expect this lag for "normal" ratios not to be noticeable.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -150,6 +150,16 @@ AudioSegmentSampleView WaveClip::GetSampleView(
       start + TimeToSamples(mTrimLeft), length, mayThrow);
 }
 
+AudioSegmentSampleView WaveClip::GetSampleView(
+   size_t iChannel, double t0, double t1, bool mayThrow) const
+{
+   assert(iChannel < GetWidth());
+   const auto start = TimeToSamples(std::max(0., t0));
+   const auto length =
+      (std::min(GetNumSamples(), TimeToSamples(t1)) - start).as_size_t();
```
Paul-Licameli, line 159: What's your proof that the assertions in `as_size_t` are satisfied -- that there is no narrowing of a negative or a too-large `SampleCount` value?

src/MixerBoard.cpp
```diff
@@ -506,6 +509,29 @@ void MixerTrackCluster::UpdateForStateChange()
...
+      });
+}
+
+// Assumes that buffer has capacity >= GetNumSamplesInView(view)
+void FillBufferWithSampleView(float* buffer, const ChannelSampleView& view)
+{
+   size_t copiedSamples = 0;
+   for (const auto& segmentView : view)
+   {
+      const auto sampleCount = segmentView.GetSampleCount().as_size_t();
```
Paul-Licameli, line 528: Same question

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -150,6 +150,16 @@ AudioSegmentSampleView WaveClip::GetSampleView(
       start + TimeToSamples(mTrimLeft), length, mayThrow);
 }
 
+AudioSegmentSampleView WaveClip::GetSampleView(
```
Paul-Licameli, line 153: I see that before this commit, `AudioSegmentSampleView` was constructed in only one place outside tests, in `Sequence`, and always passing `size_t length` to the `sampleCount` argument.

If `AudioSegmentSampleView` really eliminated use of type `sampleCount` and only used `size_t`, it would move the need to use `as_size_t` to the point of construction.

There are new places where it is constructed, all of them in `WaveTrack` and added by this commit.

libraries/lib-wave-track/WaveClip.h
```diff
@@ -283,6 +283,19 @@ class WAVE_TRACK_API WaveClip final :
       size_t iChannel, sampleCount start, size_t length,
       bool mayThrow = true) const override;
 
+   /*!
+    * @brief Request interval samples within [t0, t1). `t0` and `t1` are
+    * truncated to the clip's play start and end. Stretching influences the
+    * number of samples fitting into [t0, t1), i.e., twice as few for twice as
```
Paul-Licameli, line 289: Here and elsewhere, "half as many" would read better than "twice as few"

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -445,6 +472,20 @@ class WAVE_TRACK_API WaveTrack final
       // contiguous range.
       sampleCount* pNumWithinClips = nullptr) const override;
 
+   /*!
+    * @brief Request samples within [t0, t1), not knowing in advance how
+    * many this will be.
+    *
+    * @details The stretching of intersecting intevals influences the number of
```
Paul-Licameli, line 479: "intervals"

Same spelling error in one other place

libraries/lib-wave-track/WaveClip.h
```diff
@@ -283,6 +283,19 @@ class WAVE_TRACK_API WaveClip final :
...
+    * large a stretch ratio, due to a larger spacing of raw the samples. The
```
Paul-Licameli, line 290: "the raw" x 4 places

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2403,6 +2426,58 @@ bool WaveTrack::GetOne(
...
+   assert(IsLeader());
+   ChannelGroupSampleView result;
+   for (const auto& channel : Channels()) {
+      result.push_back(channel->GetSampleView(t0, t1, mayThrow));
+   }
+   return result;
+}
+
+ChannelSampleView
+WaveChannel::GetSampleView(double t0, double t1, bool mayThrow) const
```
Paul-Licameli, line 2441: I hope that writing this commit was a good exercise in understanding the new channel and interval interfaces.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2403,6 +2426,58 @@ bool WaveTrack::GetOne(
...
+{
+   std::vector<std::shared_ptr<const WaveChannelInterval>>
+      intersectingIntervals;
+   for (const auto& interval : Intervals())
+      if (interval->Intersects(t0, t1))
+         intersectingIntervals.push_back(interval);
+   if (intersectingIntervals.empty())
+      return { AudioSegmentSampleView(
+         TimeToLongSamples(t1) - TimeToLongSamples(t0)) };
```
Paul-Licameli, line 2450: Inner pair of `{}`?

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -65,6 +68,20 @@ class WAVE_TRACK_API WaveChannelInterval final : public ChannelInterval {
 
    const WaveClip &GetClip() const { return mClip; }
    const Envelope &GetEnvelope() const { return mEnvelope; }
+   bool Intersects(double t0, double t1) const;
+   double Start() const;
+   double End() const;
```
Paul-Licameli, line 73: Just as `Start` and `End` are non-virtuals in `ChannelGroupInterval`, so should the be too in the base class `ChannelInterval`, I think.  But that can be fixed later

libraries/lib-wave-track/WaveClip.cpp
Paul-Licameli: My own, old comment in the file says
`   // Expect that the difference of t1 and t0 is the part of a track played
    // in about 1/20 second (ticks of TrackPanel timer), so this won't overflow
`
I didn't make a really rigorous proof then.  But can we do better?

Stretches aren't limited yet to "reaonable" bounds and also the calculation above is specific to default sample rate.

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: I agree, I should be consistent with myself. A sample view is typically used to store a temporary buffer, and the total amount should never exceed what `size_t` can hold. Shall I simplify this?

libraries/lib-wave-track/WaveClip.h
saintmatthieu: sure

src/MixerBoard.cpp
```diff
@@ -601,63 +627,53 @@ void MixerTrackCluster::UpdateMeter(const double t0, const double t1)
...
+   // Don't throw on read error in this drawing update routine
+   constexpr auto mayThrow = false;
+   mSampleView = pTrack->GetSampleView(t0, t1, mayThrow);
 
    // Expect that the difference of t1 and t0 is the part of a track played
    // in about 1/20 second (ticks of TrackPanel timer), so this won't overflow
-   auto nFrames = scnFrames.as_size_t();
+   const auto nFrames = GetNumSamplesInView(mSampleView[0]);
 
    Floats tempFloatsArray{ nFrames };
```
Paul-Licameli, line 639: The deleted `as_size_t` is now instead inside the function called above.

I'm not convinced there won't even be an extreme stretch ratio.

If the assertion ever fired in a debug build -- the release build might continue, with a wrapped, small value.  No harm except an inaccurate meter that put too few floats in its update queue.

But with a less extreme value, which doesn't cause the assertion, still a large size_t value might not be allocable and the release build would crash.

So how about this little bit of bullet-proofing here.

```
Floats tempFloatsArray;
try {
   tempFloatsArray.reinit(nFrames);
}
catch(const std::bad_alloc &) {
   return;
}
```

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Indeed, whenever there's the need, we do this.

src/MixerBoard.cpp
```diff
@@ -601,63 +627,53 @@ void MixerTrackCluster::UpdateMeter(const double t0, const double t1)
...
    const auto pTrack = GetWave();
-   auto startSample = (sampleCount)((pTrack->GetRate() * t0) + 0.5);
-   auto scnFrames = (sampleCount)((pTrack->GetRate() * (t1 - t0)) + 0.5);
+
+   // Don't throw on read error in this drawing update routine
+   constexpr auto mayThrow = false;
+   mSampleView = pTrack->GetSampleView(t0, t1, mayThrow);
 
    // Expect that the difference of t1 and t0 is the part of a track played
    // in about 1/20 second (ticks of TrackPanel timer), so this won't overflow
```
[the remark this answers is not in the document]
saintmatthieu: Extremely low, to be precise. Added.

src/MixerBoard.cpp
Paul-Licameli: Do this also when constructing `meterFloatsArray`.

saintmatthieu: I just tried to crash it, calling `Floats tempFloatsArray{ std::numeric_limits<size_t>::max() };`. The exception was caught by the `Publisher` code, which visits its observers within a try/catch block. This is the case here, since this scope reacts to an `OnTimer` change. So 🤷

src/MixerBoard.cpp
```diff
@@ -601,63 +627,53 @@ void MixerTrackCluster::UpdateMeter(const double t0, const double t1)
    // of the meter display, but instead consult PlaybackPolicy
 
    const auto pTrack = GetWave();
-   auto startSample = (sampleCount)((pTrack->GetRate() * t0) + 0.5);
-   auto scnFrames = (sampleCount)((pTrack->GetRate() * (t1 - t0)) + 0.5);
+
+   // Don't throw on read error in this drawing update routine
+   constexpr auto mayThrow = false;
+   mSampleView = pTrack->GetSampleView(t0, t1, mayThrow);
```
Paul-Licameli, line 633: Something very basic:  Does `mSampleView` really need a longer lifetime than this function?  I think not, it can be a local, not a member.

saintmatthieu: It does, if we want the caching to work properly.
Actually I notice one improvement: emptying this view upon start/stop so as to allow cache eviction.

libraries/lib-wave-track/WaveClip.cpp
Paul-Licameli: would be better

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -150,6 +150,16 @@ AudioSegmentSampleView WaveClip::GetSampleView(
...
+   size_t iChannel, double t0, double t1, bool mayThrow) const
+{
+   assert(iChannel < GetWidth());
+   const auto start = TimeToSamples(std::max(0., t0));
```
Paul-Licameli, line 157: I think this was wrong:  it should be `TimeToSamples(t0 - GetPlayStartTime())`

Paul-Licameli (approved):

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: No, the user is already supposed to give clip-relative times.

src/MixerBoard.cpp
saintmatthieu: But what use would this be, since the exception is already caught above ? I'm not sure we're understanding each other.

saintmatthieu:
Thanks @Paul-Licameli for contributing and pushing this forward !

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2364,6 +2364,10 @@ bool WaveTrack::GetOne(
 
       if (clipEnd > start && clipStart < start+len)
       {
+         // Yes, exact comparison
```
saintmatthieu, line 2390: An elaboration in this comment would be welcome.

PR #5042, opened 2023-08-15
Time stretching 3 of 6 repair drawing tools for stretched audio

### LINKED ISSUE #4850: Audio clip stretching

saintmatthieu:
Resolves: #4850

This PR repairs drawing operations on stretched clips.

Recommended:

## QA
- [x] editing one sample at a time by clicking with pencil,
- [x] editing many samples by dragging pencil across clips, some stretched, others not, some mono, others stereo.
- [x] same exercise with brush tool (select pencil then press alt)
- [x] trying to draw samples when individual samples are not visible (because zoomed out) issues the usual error message.
- [x] (Paul) test all in the pencil tool mode and again in multi-tool mode
- [x] (Paul) hit-testing on samples in multi tool with nontrivial envelope (this doesn’t work quite right in last release, but be sure it is not worse)
- [x] (Paul) include paintbrush cases where there is hidden smart clip data, and where clips (maybe with hidden data, maybe with different stretches) are touching or nearly touching
- [x] (Paul) I think we're also fixing an old misbehavior:  click and drag with the pencil sometimes did nothing but made a vacuous undo/redo item
- [x] (Paul) also remember (I had forgot!) that a drag with Ctrl key down adjusts only one pixel
- [x] Clicking or dragging edits nearest sample (not previous or next - nearest)
- [x] Drag-drawing on a clip's rectangle (not out-of-bound) always updates the nearest sample, not just when the pencil crosses its x-location
- [x] Works as well for stereo

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -201,6 +201,55 @@ void WaveClip::SetSamples(size_t ii,
    MarkChanged();
 }
 
+bool WaveClip::GetFloatAtTime(
+   double t, size_t iChannel, float& value, bool mayThrow) const
+{
+   if (!WithinPlayRegion(t))
+      return false;
+   const auto start = TimeToSamples(t);
+   return GetSamples(
+      iChannel, reinterpret_cast<samplePtr>(&value), floatSample, start, 1u,
+      mayThrow);
+}
+
+void WaveClip::SetFloatsFromTime(
+   double t, size_t iChannel, const float* buffer, size_t numFloats,
+   sampleFormat effectiveFormat)
+{
+   const auto maybeNegativeStart = TimeToSamples(t);
+   const auto maybeOutOfBoundEnd = maybeNegativeStart + numFloats;
+   const auto effectiveStart = std::max(sampleCount { 0 }, maybeNegativeStart);
+   const auto effectiveEnd =
+      std::min(GetVisibleSampleCount(), maybeOutOfBoundEnd);
+   if (effectiveStart >= effectiveEnd)
+      return;
+   // Cannot be greater than `numFloats` -> safe cast
+   const auto effectiveLen = (effectiveEnd - effectiveStart).as_size_t();
+   // Cannot be greater than `numFloats` -> safe cast
+   const auto numLeadingZeros =
+      (effectiveStart - maybeNegativeStart).as_size_t();
+   const auto offsetBuffer =
+      reinterpret_cast<const char*>(buffer + numLeadingZeros);
+   SetSamples(
+      iChannel, offsetBuffer, floatSample, effectiveStart, effectiveLen,
+      effectiveFormat);
+}
+
+void WaveClip::SetFloatsCenteredAroundTime(
+   double t, size_t iChannel, const float* buffer, size_t numSideSamples,
+   sampleFormat effectiveFormat)
+{
+   SetFloatsFromTime(
+      t - SamplesToTime(numSideSamples), iChannel, buffer,
+      2 * numSideSamples + 1, effectiveFormat);
+}
+
+void WaveClip::SetFloatAtTime(
+   double t, size_t iChannel, float value, sampleFormat effectiveFormat)
+{
+   SetFloatsCenteredAroundTime(t, iChannel, &value, 0u, effectiveFormat);
+}
+
 BlockArray* WaveClip::GetSequenceBlockArray(size_t ii)
 {
    assert(ii < GetWidth());
--- a/libraries/lib-wave-track/WaveClip.h
+++ b/libraries/lib-wave-track/WaveClip.h
@@ -215,7 +215,9 @@ public:
    //! Real end time of the clip, quantized to raw sample rate (track's rate)
    sampleCount GetPlayEndSample() const;
 
-   // todo comment
+   /*!
+    * Returns a number of raw samples, not accounting for stretching.
+    */
    sampleCount GetVisibleSampleCount() const override;
 
    //! Sets the play start offset in seconds from the beginning of the underlying sequence
@@ -322,6 +324,7 @@ public:
    //! Get samples from one channel
    /*!
     @param ii identifies the channel
+    @param start relative to clip play start sample
     @pre `ii < GetWidth()`
     */
    bool GetSamples(size_t ii, samplePtr buffer, sampleFormat format,
@@ -330,6 +333,7 @@ public:
    //! Get (non-interleaved) samples from all channels
    /*!
     assume as many buffers available as GetWidth()
+    @param start relative to clip play start sample
     */
    bool GetSamples(samplePtr buffers[], sampleFormat format,
                    sampleCount start, size_t len, bool mayThrow = true) const;
@@ -337,6 +341,7 @@ public:
    //! @param ii identifies the channel
    /*!
     @pre `ii < GetWidth()`
+    @param start relative to clip play start sample
     */
    void SetSamples(size_t ii, constSamplePtr buffer, sampleFormat format,
       sampleCount start, size_t len,
@@ -347,6 +352,43 @@ public:
          than the effective, then no dithering will occur.
       */
    );
+
+   /*!
+    * @param t relative to clip start sample
+    */
+   bool
+   GetFloatAtTime(double t, size_t iChannel, float& value, bool mayThrow) const;
+
+   //! Succeed with out-of-bounds requests, only changing what is in bounds.
+   //! @{
+   // clang-format off
+   /*!
+    * @brief Considers `buffer` as audio starting at `TimeToSamples(t)`
+    * (relative to clip play start time) and with equal stretch ratio. Samples
+    * at intersecting indices are then copied, leaving non-intersecting clip
+    * samples untouched. E.g.,
+    *     buffer:      [a b c d e]
+    *     clip  :            [x y z]
+    *     result:            [d e z]
+    */
+   // clang-format on
+   void SetFloatsFromTime(
+      double t, size_t iChannel, const float* buffer, size_t numSamples,
+      sampleFormat effectiveFormat);
+
+   /*!
+    * @brief Same as `SetFloatsFromTime`, but with `buffer` starting at
+    * `TimeToSamples(t0 -  SamplesToTime(numSideSamples))`.
+    * `[buffer, buffer + 2 * numSizeSamples + 1)` is assumed to be a valid span
+    * of addresses.
+    */
+   void SetFloatsCenteredAroundTime(
+      double t, size_t iChannel, const float* buffer, size_t numSideSamples,
+      sampleFormat effectiveFormat);
+
+   void SetFloatAtTime(
+      double t, size_t iChannel, float value, sampleFormat effectiveFormat);
+   //! @}
 
    Envelope* GetEnvelope() { return mEnvelope.get(); }
    const Envelope* GetEnvelope() const { return mEnvelope.get(); }
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -35,11 +35,12 @@ from the project that will own the track.
 #include <wx/debug.h>
 #include <wx/log.h>
 
+#include <algorithm>
 #include <float.h>
 #include <math.h>
-#include <algorithm>
-#include <optional>
 #include <numeric>
+#include <optional>
+#include <type_traits>
 
 #include "float_cast.h"
 
@@ -2716,6 +2717,200 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
          pTrack = *(++ *iter);
       return result;
    });
+}
+
+namespace {
+void RoundToNearestClipSample(const WaveTrack& track, double& t)
+{
+   const auto clip = track.GetClipAtTime(t);
+   if (!clip)
+      return;
+   t = clip->SamplesToTime(clip->TimeToSamples(t - clip->GetPlayStartTime())) +
+       clip->GetPlayStartTime();
+}
+}
+
+std::pair<size_t, size_t> WaveTrack::GetFloatsCenteredAroundTime(
+   double t, size_t iChannel, float* buffer, size_t numSideSamples,
+   bool mayThrow) const
+{
+   const auto numSamplesReadLeft = GetFloatsFromTime(
+      t, iChannel, buffer, numSideSamples, mayThrow, PlaybackDirection::backward);
+   const auto numSamplesReadRight = GetFloatsFromTime(
+      t, iChannel, buffer + numSideSamples, numSideSamples + 1, mayThrow,
+      PlaybackDirection::forward);
+   return { numSideSamples - numSamplesReadLeft,
+            numSideSamples + numSamplesReadRight };
+}
+
+namespace
+{
+template <typename FloatType>
+using BufferCharType = std::conditional_t<
+   std::is_const_v<std::remove_pointer_t<FloatType>>, constSamplePtr,
+   samplePtr>;
+
+template <typename BufferType> struct SampleAccessArgs
+{
+   const BufferCharType<BufferType> offsetBuffer;
+   const sampleCount start;
+   const size_t len;
+};
+
+template <typename BufferType>
+SampleAccessArgs<BufferType> GetSampleAccessArgs(
+   const WaveClip& clip, double startOrEndTime /*absolute*/, BufferType buffer,
+   size_t totalToRead, size_t alreadyRead, bool forward)
+{
+   assert(totalToRead >= alreadyRead);
+   const auto remainingToRead = totalToRead - alreadyRead;
+   const auto sampsInClip = clip.GetVisibleSampleCount();
+   const auto sampsPerSec = clip.GetRate() / clip.GetStretchRatio();
+   if (forward)
+   {
+      const auto startTime =
+         std::max(startOrEndTime - clip.GetPlayStartTime(), 0.);
+      const sampleCount startSamp { std::round(startTime * sampsPerSec) };
+      if (startSamp >= sampsInClip)
+         return { nullptr, sampleCount { 0u }, 0u };
+      const auto len =
+         limitSampleBufferSize(remainingToRead, sampsInClip - startSamp);
+      return { reinterpret_cast<BufferCharType<BufferType>>(
+                  buffer + alreadyRead),
+               startSamp, len };
+   }
+   else
+   {
+      const auto endTime = std::min(
+         startOrEndTime - clip.GetPlayStartTime(), clip.GetPlayDuration());
+      const sampleCount endSamp { std::round(endTime * sampsPerSec) };
+      const auto startSamp =
+         std::max(endSamp - remainingToRead, sampleCount { 0 });
+      // `len` cannot be greater than `remainingToRead`, itself a `size_t` ->
+      // safe cast.
+      const auto len = (endSamp - startSamp).as_size_t();
+      if (len == 0 || startSamp >= sampsInClip)
+         return { nullptr, sampleCount { 0u }, 0u };
+      const auto bufferEnd = buffer + remainingToRead;
+      return { reinterpret_cast<BufferCharType<BufferType>>(bufferEnd - len),
+               startSamp, len };
+   }
+}
+} // namespace
+
+size_t WaveTrack::GetFloatsFromTime(
+   double t, size_t iChannel, float* buffer, size_t numSamples, bool mayThrow,
+   PlaybackDirection direction) const
+{
+   RoundToNearestClipSample(*this, t);
+   auto clip = GetClipAtTime(t);
+   auto numSamplesRead = 0u;
+   const auto forward = direction == PlaybackDirection::forward;
+   while (clip)
+   {
+      const auto args = GetSampleAccessArgs(
+         *clip, t, buffer, numSamples, numSamplesRead, forward);
+      if (!clip->GetSamples(
+             iChannel, args.offsetBuffer, floatSample, args.start, args.len,
+             mayThrow))
+         return 0u;
+      numSamplesRead += args.len;
+      if (numSamplesRead >= numSamples)
+         break;
+      clip = GetAdjacentClip(*clip, direction);
+   }
+   return numSamplesRead;
+}
+
+bool WaveTrack::GetFloatAtTime(
+   double t, size_t iChannel, float& value, bool mayThrow) const
+{
+   const auto clip = GetClipAtTime(t);
+   if (!clip)
+      return false;
+   clip->GetFloatAtTime(
+      t - clip->GetPlayStartTime(), iChannel, value, mayThrow);
+   return true;
+}
+
+void WaveTrack::SetFloatsCenteredAroundTime(
+   double t, size_t iChannel, const float* buffer, size_t numSideSamples,
+   sampleFormat effectiveFormat)
+{
+   SetFloatsFromTime(
+      t, iChannel, buffer, numSideSamples, effectiveFormat,
+      PlaybackDirection::backward);
+   SetFloatsFromTime(
+      t, iChannel, buffer + numSideSamples, numSideSamples + 1, effectiveFormat,
+      PlaybackDirection::forward);
+}
+
+void WaveTrack::SetFloatsFromTime(
+   double t, size_t iChannel, const float* buffer, size_t numSamples,
+   sampleFormat effectiveFormat, PlaybackDirection direction)
+{
+   RoundToNearestClipSample(*this, t);
+   auto clip = GetClipAtTime(t);
+   auto numSamplesWritten = 0u;
+   const auto forward = direction == PlaybackDirection::forward;
+   while (clip)
+   {
+      const auto args = GetSampleAccessArgs(
+         *clip, t, buffer, numSamples, numSamplesWritten, forward);
+      if (args.len > 0u)
+      {
+         clip->SetSamples(
+            iChannel, args.offsetBuffer, floatSample, args.start, args.len,
+            effectiveFormat);
+         numSamplesWritten += args.len;
+         if (numSamplesWritten >= numSamples)
+            break;
+      }
+      clip = GetAdjacentClip(*clip, direction);
+   }
+}
+
+void WaveTrack::SetFloatAtTime(
+   double t, size_t iChannel, float value, sampleFormat effectiveFormat)
+{
+   SetFloatsCenteredAroundTime(t, iChannel, &value, 0u, effectiveFormat);
+}
+
+void WaveTrack::SetFloatsWithinTimeRange(
+   double t0, double t1, size_t iChannel,
+   const std::function<float(double sampleTime)>& producer,
+   sampleFormat effectiveFormat)
+{
+   assert(t0 <= t1);
+   const auto sortedClips = SortedClipArray();
+   if (sortedClips.empty())
+      return;
+   t0 = std::max(t0, (*sortedClips.begin())->GetPlayStartTime());
+   t1 = std::min(t1, (*sortedClips.rbegin())->GetPlayEndTime());
+   auto clip = GetClipAtTime(t0);
+   while (clip) {
+      const auto clipStartTime = clip->GetPlayStartTime();
+      const auto clipEndTime = clip->GetPlayEndTime();
+      const auto sampsPerSec = clip->GetRate() / clip->GetStretchRatio();
+      const auto roundedT0 =
+         std::round((t0 - clipStartTime) * sampsPerSec) / sampsPerSec +
+         clipStartTime;
+      const auto roundedT1 =
+         std::round((t1 - clipStartTime) * sampsPerSec + 1) / sampsPerSec +
+         clipStartTime;
+      if (clipStartTime > roundedT1)
+         break;
+      const auto tt0 = std::max(clipStartTime, roundedT0);
+      const auto tt1 = std::min(clipEndTime, roundedT1);
+      const size_t numSamples = (tt1 - tt0) * sampsPerSec + .5;
+      std::vector<float> values(numSamples);
+      for (auto i = 0u; i < numSamples; ++i)
+         values[i] = producer(tt0 + clip->SamplesToTime(i));
+      clip->SetFloatsFromTime(
+         tt0 - clipStartTime, iChannel, values.data(), numSamples,
+         effectiveFormat);
+      clip = GetNextClip(*clip, PlaybackDirection::forward);
+   }
 }
 
 bool WaveTrack::GetOne(
--- a/libraries/lib-wave-track/WaveTrack.h
+++ b/libraries/lib-wave-track/WaveTrack.h
@@ -528,6 +528,56 @@ public:
       sampleCount* pNumWithinClips = nullptr) const override;
 
    /*!
+    * @brief Gets as many samples as it can, but no more than `2 *
+    * numSideSamples + 1`, centered around `t`. Reads nothing if
+    * `GetClipAtTime(t) == nullptr`. Useful to access samples across clip
+    * boundaries, as it spreads the read to adjacent clips, i.e., not separated
+    * by silence from clip at `t`.
+    *
+    * @return The begin and end indices of the samples in the buffer where
+    * samples could actually be copied.
+    */
+   std::pair<size_t, size_t> GetFloatsCenteredAroundTime(
+      double t, size_t iChannel, float* buffer, size_t numSideSamples,
+      bool mayThrow) const;
+
+   /*!
+    * @return true if `GetClipAtTime(t) != nullptr`, false otherwise.
+    */
+   bool
+   GetFloatAtTime(double t, size_t iChannel, float& value, bool mayThrow) const;
+
+   /*!
+    * @brief Similar to GetFloatsCenteredAroundTime, but for writing. Sets as
+    * many samples as it can according to the same rules as
+    * GetFloatsCenteredAroundTime. Leaves the other samples untouched. @see
+    * GetFloatsCenteredAroundTime
+    */
+   void SetFloatsCenteredAroundTime(
+      double t, size_t iChannel, const float* buffer, size_t numSideSamples,
+      sampleFormat effectiveFormat);
+
+   /*!
+    * @brief Sets sample nearest to `t` to `value`. Silently fails if
+    * `GetClipAtTime(t) == nullptr`.
+    */
+   void SetFloatAtTime(
+      double t, size_t iChannel, float value, sampleFormat effectiveFormat);
+
+   /*!
+    * @brief Provides a means of setting clip values as a function of time.
+    * Included are closest sample to t0 up to closest sample to t1, inclusively.
+    * Given that `t0 <= t1`, always at least one sample is included.
+    * @param producer a function taking sample (absolute, not clip-relative)
+    * time and returning the desired value for the sample at that time.
+    * @pre t0 <= t1
+    */
+   void SetFloatsWithinTimeRange(
+      double t0, double t1, size_t iChannel,
+      const std::function<float(double sampleTime)>& producer,
+      sampleFormat effectiveFormat);
+
+   /*!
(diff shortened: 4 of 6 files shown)


libraries/lib-wave-track/WaveTrack.h
```diff
@@ -371,6 +372,74 @@ class WAVE_TRACK_API WaveTrack final
       */
    );
 
+   /*!
+    * @brief Gets as many samples as it can, but no more than `2 *
+    * numSideSamples + 1`, centered around `t`. Reads nothing if
+    * `GetClipAtTime(t) == nullptr`. Useful to acess samples across clip
```
Paul-Licameli, line 378: “Access”

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -371,6 +372,74 @@ class WAVE_TRACK_API WaveTrack final
...
+    * GetFloatsCenteredAroundTime. Leaves the other samples untouched. @see
+    * GetFloatsCenteredAroundTime
+    */
+   void SetFloatsCenteredAroundTime(
+      double t, size_t iChannel, const float* buffer, size_t numSideSamples,
+      sampleFormat effectiveFormat);
+
+   /*!
+    * @brief Similar to GetFloatsFromTime, but for writing. Sets as many samples
+    * as it can according to the same ruls as GetFloatsFromTime. Leaves the
```
Paul-Licameli, line 418: “Rules”

saintmatthieu:
This morning I came to remember maybe my very first experience of Audacity. It was at the National University of Ireland, Maynooth, in the computer music introductory module. I remember the lecturer saying "and you can even draw the samples with the pencil". I must have been impressed to still remember it - it was in 2005-2006. Great educational tool. @rbdannenberg @Paul-Licameli

libraries/lib-wave-track/WaveClip.h
```diff
@@ -309,6 +318,20 @@ class WAVE_TRACK_API WaveClip final :
       */
    );
 
+   bool
+   GetFloatAtTime(double t, size_t iChannel, float& value, bool mayThrow) const;
+
+   void SetFloatsFromTime(
```
Paul-Licameli, line 324: Please comment that the new setting functions will succeed with out-of-bounds requests, only changing what is in bounds.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -371,6 +372,74 @@ class WAVE_TRACK_API WaveTrack final
       */
    );
 
+   /*!
+    * @brief Gets as many samples as it can, but no more than `2 *
+    * numSideSamples + 1`, centered around `t`. Reads nothing if
+    * `GetClipAtTime(t) == nullptr`. Useful to acess samples across clip
+    * boundaries, as it spreads the read to adjacent clips, i.e., not separated
+    * by silence from clip at `t`.
```
Paul-Licameli, line 380: Is this description really what we want?

To be REALLY strictly conservative of program behavior, in the case of not time stretching, it's possible two clips were separated by a tiny gap of one or two sample intervals, but the paint brush might affect both.

And maybe nobody in the history of Audacity actually ever did this.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -371,6 +372,74 @@ class WAVE_TRACK_API WaveTrack final
...
+   /*!
+    * @brief Similar to GetFloatsCenteredAroundTime, but reads only in one
+    * direction. @see GetFloatsCenteredAroundTime
+    *
+    * @return The number of samples actually copied.
+    */
+   size_t GetFloatsFromTime(
+      double t, size_t iChannel, float* buffer, size_t numSamples,
+      bool mayThrow,
+      PlaybackDirection direction = PlaybackDirection::backward) const;
```
Paul-Licameli, line 398: `backward` seems surprising as the default value.  But you never call it with the default.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -371,6 +372,74 @@ class WAVE_TRACK_API WaveTrack final
...
+
+   /*!
+    * @brief Similar to GetFloatsFromTime, but for writing. Sets as many samples
+    * as it can according to the same ruls as GetFloatsFromTime. Leaves the
+    * other samples untouched. @see GetFloatsFromTime
+    */
+   void SetFloatsFromTime(
+      double t, size_t iChannel, const float* buffer, size_t numSamples,
+      sampleFormat effectiveFormat,
+      PlaybackDirection direction = PlaybackDirection::backward);
```
Paul-Licameli, line 424: ditto, default argument

src/tracks/playabletrack/wavetrack/ui/SampleHandle.cpp
```diff
@@ -93,8 +99,12 @@ namespace {
    {
       // Require more than 3 pixels per sample
       const auto xx = std::max<ZoomInfo::int64>(0, viewInfo.TimeToPosition(time));
+      const auto clip = wt->GetClipAtTime(time);
+      if (!clip)
+         // Don't bother the user about that with a pop-up.
+         return true;
```
Paul-Licameli, line 105: This function is used in two places, and this change, while avoiding that pop-up when you click between clips, also will change the hit test to return a handle in the case that it did not do so before.

Does clicking and dragging in the empty space between clips make any vacuous undo/redo items?

Paul-Licameli: Aha, I tried the build.  Yes it does!

So instead the hit testing function should also be changed to return null when there is no clip.

Paul-Licameli: Actually... that's an old defect, which you might fix.

I just tried released Audacity.  When the magnification does show samples, but you click and drag outside of a clip, still you do get a bogus undo/redo item.

Paul-Licameli: Actually an old problem, and I was confused about something.  `HitTest` only happens with the * tool button.  With the pencil tool button, you can click anywhere and it has always been like that.

Paul-Licameli: `SampleHandle::HitAnywhere` is also used in WaveformView.cpp in this case.

src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
```diff
@@ -169,7 +169,16 @@ UIHandle::Result CutlineHandle::Click
 
          viewInfo.selectedRegion.setTimes(cutlineStart, cutlineEnd);
       }
-      else if (mLocation.typ == WaveTrackLocation::locationMergePoint) {
+      // todo(mhodgkinson) Bluntly disabling this Join shortcut for now. There
+      // exists an issue that wants to do this properly:
+      // https://github.com/audacity/audacity/issues/5020
```
Paul-Licameli, line 174: I think that's not the right issue number

Paul-Licameli:
> This morning I came to remember maybe my very first experience of Audacity. It was at the National University of Ireland, Maynooth, in the computer music introductory module. I remember the lecturer saying "and you can even draw the samples with the pencil". I must have been impressed to still remember it - it was in 2005-2006. Great educational tool. @rbdannenberg @Paul-Licameli

Would you believe I hadn’t heard of, less used it, before December 2012?

libraries/lib-wave-track/WaveClip.h
```diff
@@ -189,9 +189,18 @@ class WAVE_TRACK_API WaveClip final :
    double GetPlayEndTime() const override;
    double GetPlayDuration() const;
 
-   // todo comment
+   /*!
+    * Accessors for visible boundaries within a track after stretching, that is
+    * (ommitting rounding):
```
Paul-Licameli, line 194: Omitting

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -455,6 +455,74 @@ class WAVE_TRACK_API WaveTrack final
...
+   void SetFloatsCenteredAroundTime(
+      double t, size_t iChannel, const float* buffer, size_t numSideSamples,
+      sampleFormat effectiveFormat);
+
+   /*!
+    * @brief Similar to GetFloatsFromTime, but for writing. Sets as many samples
+    * as it can according to the same ruls as GetFloatsFromTime. Leaves the
+    * other samples untouched. @see GetFloatsFromTime
+    */
+   void SetFloatsFromTime(
```
Paul-Licameli, line 504: You need more comments to clarify:

When direction is backward, the buffer of given samples is not reversed, and `t` rounds to the time of the clip sample corresponding to the past-the-end position.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -455,6 +455,74 @@ class WAVE_TRACK_API WaveTrack final
...
+      double t, size_t iChannel, float* buffer, size_t numSideSamples,
+      bool mayThrow) const;
+
+   /*!
+    * @brief Similar to GetFloatsCenteredAroundTime, but reads only in one
+    * direction. @see GetFloatsCenteredAroundTime
+    *
+    * @return The number of samples actually copied.
+    */
+   size_t GetFloatsFromTime(
```
Paul-Licameli, line 478: Clarify that the resulting samples are not reversed when direction is backward.

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Now that clips can be stretched, we have to choose between constant time or constant sample-count brush radius. The approach I took is the latter. But then, samples of a clip 10 seconds away should not be taken into account. So I went for the simple rule that they have to touch each other, or nothing happens.
I also reckon we won't run into trouble because of this subtle change of behaviour.

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Right, makes no particular sense - removed it (and moved this method to `private`).

src/tracks/playabletrack/wavetrack/ui/SampleHandle.cpp
saintmatthieu: > Actually... that's an old defect, which you might fix.

Each time `SampleHandle::Release` is called (when letting go of the mouse with pencil tool activated), a history item is pushed. The easiest solution to this I can think of would be to create a `bool mSampleMoved`, whose value is set to true whenever a sample is modified, checked when releasing, and then set to false again. That'd be something totally new, and not completely trivial to review.

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Moved those to `private` and made comments a bit more precise.

saintmatthieu:
@Paul-Licameli just pushed [05c4f37](https://github.com/audacity/audacity/pull/5042/commits/05c4f37a53ec2ebe42ddfbfd2b478d4cd1e6eb01), addressing all your comments.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -884,6 +931,26 @@ class WAVE_TRACK_API WaveTrack final
       bool backwards, fillFormat fill, bool mayThrow,
       sampleCount* pNumWithinClips) const;
 
+   /*!
+    * @brief Helper for GetFloatsCenteredAroundTime. If `direction ==
+    * PlaybackDirection::Backward`, fetches samples to the left of `t`, exluding
```
Paul-Licameli, line 936: "excluding"

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -197,6 +197,52 @@ void WaveClip::SetSamples(size_t ii,
...
+void WaveClip::SetFloatsFromTime(
+   double t, size_t iChannel, const float* buffer, size_t numFloats,
+   sampleFormat effectiveFormat)
+{
+   const auto maybeNegativeStart = TimeToSamples(t);
+   const auto maybeOutOfBoundEnd = maybeNegativeStart + numFloats;
+   const auto effectiveStart = std::max(sampleCount { 0 }, maybeNegativeStart);
+   const auto effectiveEnd =
+      std::min(GetVisibleSampleCount(), maybeOutOfBoundEnd);
+   const auto effectiveLen = (effectiveEnd - effectiveStart).as_size_t();
```
Paul-Licameli, line 220: As always, I see `as_size_t` and think, what's the proof of satsifaction of the assertion?

Here, "it's easy to prove that the expression is bounded by numFloats".  So just say that in a comment.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -197,6 +197,52 @@ void WaveClip::SetSamples(size_t ii,
...
+   if (effectiveLen == 0)
+      return;
+   const auto numLeadingZeros = (effectiveStart - maybeNegativeStart).as_size_t();
```
Paul-Licameli, line 223: likewise

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -197,6 +197,52 @@ void WaveClip::SetSamples(size_t ii,
...
+   const auto offsetBuffer =
+      reinterpret_cast<const char*>(buffer + numLeadingZeros);
+   SetSamples(
+      iChannel, offsetBuffer, floatSample, effectiveStart, effectiveLen,
+      effectiveFormat);
```
Paul-Licameli, line 228: Please comment in the header that the function may rewrite only a subrange of `[buffer, buffer + nFloats)`, leaving unchanged the buffer positions not corresponding to part of the clip play region.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -197,6 +197,52 @@ void WaveClip::SetSamples(size_t ii,
...
+}
+
+void WaveClip::SetFloatsCenteredAroundTime(
+   double t, size_t iChannel, const float* buffer, size_t numSideSamples,
+   sampleFormat effectiveFormat)
+{
+   SetFloatsFromTime(
+      t - SamplesToTime(numSideSamples), iChannel, buffer,
+      2 * numSideSamples + 1, effectiveFormat);
```
Paul-Licameli, line 237: Please comment that `[buffer, buffer + 2 * numSizeSamples + 1)` is assumed to be a valid span of addresses, and likewise that only a subrange may be overwritten.

In other words zero-filling is not promised for the rest.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -477,6 +477,53 @@ class WAVE_TRACK_API WaveTrack final
...
+
+   /*!
+    * @brief Provides a means of setting clip values as a function of time.
+    * @param producer a function taking sample (absolute, not clip-relative)
+    * time and returning the desired value for the sample at that time.
+    */
+   void SetFloatsWithinTimeRange(
+      double t0, double t1, size_t iChannel,
+      const std::function<float(double sampleTime)>& producer,
+      sampleFormat efffectiveFormat);
```
Paul-Licameli, line 525: "effective"

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
 }
 
+namespace {
+void RoundToNearestClipSample(const WaveTrack& track, double& t)
+{
+   const auto clip = track.GetClipAtTime(t);
+   if (!clip)
+      return;
+   t = clip->SamplesToTime(clip->TimeToSamples(t - clip->GetPlayStartTime())) +
+       clip->GetPlayStartTime();
```
Paul-Licameli, line 2349: `WaveClip::SnapToSample` analogous to the `WideSampleSequence` member function might be nice

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+struct BufferCharType<float*>
+{
+   using type = samplePtr;
+};
+
+template<>
+struct BufferCharType<const float*>
+{
+   using type = constSamplePtr;
+};
```
Paul-Licameli, line 2381: I might simplify all that with 
`#include <type_traits>`
and
`template<typename T> using BufferCharType =
   std::conditional_t<std::is_const_v<std::remove_pointer_t<T>>, constSamplePtr, samplePtr>;`

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+   double t, size_t iChannel, float* buffer, size_t numSamples, bool mayThrow,
+   PlaybackDirection direction) const
+{
+   RoundToNearestClipSample(*this, t);
+   auto clip = GetClipAtTime(t);
+   auto numSamplesRead = 0u;
+   const auto forward = direction == PlaybackDirection::forward;
+   while (clip)
+   {
+      const auto args = GetSampleAccessArgs<float*>(
```
Paul-Licameli, line 2441: I think the template argument can be omitted, nicely deduced from arguments

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+   double t, size_t iChannel, const float* buffer, size_t numSamples,
+   sampleFormat effectiveFormat, PlaybackDirection direction)
+{
+   RoundToNearestClipSample(*this, t);
+   auto clip = GetClipAtTime(t);
+   auto numSamplesWritten = 0u;
+   const auto forward = direction == PlaybackDirection::forward;
+   while (clip)
+   {
+      const auto args = GetSampleAccessArgs<const float*>(
```
Paul-Licameli, line 2488: here too

libraries/lib-wave-track/WaveClip.h
Paul-Licameli: `GetSamples` (not a new function) should also comment that `start` is relative to play start
`GetFloatAtTime` should comment that `t` is relative to play start

Paul-Licameli:
`GetSamples` (not a new function) should also comment that `start` is relative to play start
`GetFloatAtTime` should comment that `t` is relative to play start

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+{
+   const auto remainingToRead = totalToRead - alreadyRead;
+   const auto sampsInClip = clip.GetVisibleSampleCount();
+   const auto sampsPerSec = clip.GetRate() / clip.GetStretchRatio();
+   if (forward)
+   {
+      const auto startTime =
+         std::max(startOrEndTime - clip.GetPlayStartTime(), 0.);
+      const sampleCount startSamp { std::ceil(
+         std::round(startTime * sampsPerSec * 1e6) / 1e6) };
```
Paul-Licameli, line 2403: Explain the strange use of `1e6`

Paul-Licameli: Rounding to the nearest millionth of a sample?  Why?

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+      if (startSamp >= sampsInClip)
+         return { nullptr, sampleCount { 0u }, 0u };
+      const auto len =
+         limitSampleBufferSize(remainingToRead, sampsInClip - startSamp);
+      return { reinterpret_cast<typename BufferCharType<BufferType>::type>(
+                  buffer + alreadyRead),
```
Paul-Licameli, line 2409: In the case that `max` above clipped a negative value to 0,  should the buffer instead get some zeroes at `buffer + alreadyRead` (if writing), and should the first member of the return be a later position?

Maybe not but I'm trying to understand the usage

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+         startOrEndTime - clip.GetPlayStartTime(), clip.GetPlayDuration());
+      const sampleCount endSamp { std::ceil(
+         std::round(endTime * sampsPerSec * 1e6) / 1e6) };
+      const auto startSamp =
+         std::max(endSamp - remainingToRead, sampleCount { 0 });
+      const auto len = (endSamp - startSamp).as_size_t();
+      if (len == 0 || startSamp >= sampsInClip)
+         return { nullptr, sampleCount { 0u }, 0u };
+      const auto bufferEnd = buffer + remainingToRead;
+      return { reinterpret_cast<typename BufferCharType<BufferType>::type>(
```
Paul-Licameli, line 2424: Similar question whether this is the right thing when `min` above did some work

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+         *clip, t, buffer, numSamples, numSamplesRead, forward);
+      if (!clip->GetSamples(
+             iChannel, args.offsetBuffer, floatSample, args.start, args.len,
+             mayThrow))
+         return 0u;
+      numSamplesRead += args.len;
+      if (numSamplesRead == numSamples)
+         break;
+      clip = GetAdjacentClip(*clip, direction);
+   }
```
Paul-Licameli, line 2451: This loop may concatenate samples from clips of different stretch ratios into one buffer, and then we forget about the stretches and operate on the buffer.

Understand that I objected to this for Plot Spectrum (weightings of windows in the combined result might be wrong) and for Mixer Board (RMS calculations may be wrong again for mis-weighting, but I let that problem go).

But here, doing microscopic repairs with your pencil or the even less-known paintbrush tool, it is defensible.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+   {
+      const auto args = GetSampleAccessArgs<const float*>(
+         *clip, t, buffer, numSamples, numSamplesWritten, forward);
+      if (args.len > 0u)
+      {
+         clip->SetSamples(
+            iChannel, args.offsetBuffer, floatSample, args.start, args.len,
+            effectiveFormat);
+         numSamplesWritten += args.len;
+         if (numSamplesWritten == numSamples)
```
Paul-Licameli, line 2496: I think `>=` makes me surer of loop termination just looking at this function.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+   while (clip)
+   {
+      const auto args = GetSampleAccessArgs<float*>(
+         *clip, t, buffer, numSamples, numSamplesRead, forward);
+      if (!clip->GetSamples(
+             iChannel, args.offsetBuffer, floatSample, args.start, args.len,
+             mayThrow))
+         return 0u;
+      numSamplesRead += args.len;
+      if (numSamplesRead == numSamples)
```
Paul-Licameli, line 2448: Again `>=` would look better to me

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+         break;
+      clip = GetAdjacentClip(*clip, direction);
```
Paul-Licameli, line 2450: Here and elsewhere you use this or `GetNextClip`, there is a sorting of the clips each time.

Probably unimportant in practice for performance.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+      const auto tt0 = std::max(clipStartTime, roundedT0);
+      const auto tt1 = std::min(clipEndTime, roundedT1);
+      const size_t numSamples = (tt1 - tt0) * sampsPerSec + .5;
+      std::vector<float> values(numSamples);
+      for (auto i = 0u; i < numSamples; ++i)
+         values[i] = producer(tt0 + clip->SamplesToTime(i));
+      clip->SetFloatsFromTime(
+         tt0 - clipStartTime, iChannel, values.data(), numSamples,
+         effectiveFormat);
+      clip = GetNextClip(*clip, PlaybackDirection::forward);
```
Paul-Licameli, line 2534: Next, not Adjacent; and I understand why

src/tracks/playabletrack/wavetrack/ui/SampleHandle.cpp
```diff
@@ -142,7 +150,7 @@ UIHandlePtr SampleHandle::HitTest
    Envelope* env = wavetrack->GetEnvelopeAtTime(time);
    if (env)
       // Calculate sample as it would be rendered, so quantize time
-      envValue = env->GetValue( tt, 1.0 / wavetrack->GetRate() );
```
Paul-Licameli, line 145: I think the "so quantize time" part no longer makes sense

src/tracks/playabletrack/wavetrack/ui/SampleHandle.cpp
```diff
@@ -290,9 +300,8 @@ UIHandle::Result SampleHandle::Click
       }
       //Set the sample to the point of the mouse event
```
Paul-Licameli, line 301: This old comment should instead say, set a range of samples around the mouse event

src/tracks/playabletrack/wavetrack/ui/SampleHandle.cpp
```diff
@@ -345,66 +361,25 @@ UIHandle::Result SampleHandle::Drag
...
+   const int x1 =
+      event.m_controlDown ? mClickedStartPixel : viewInfo.TimeToPosition(t1);
    const float newLevel = FindSampleEditingLevel(event, viewInfo, t0);
+   const auto start = std::min(t0, t1);
+   const auto end = std::max(t0, t1);
+   // For fast pencil movements covering more than one sample between two
+   // updates, we draw a line going from v0 at t0 to v1 at t1.
+   const auto interpolator = [t0, t1, v0 = mLastDragSampleValue,
+                              v1 = newLevel](double t) {
+      const auto gradient = (v1 - v0) / (t1 - t0);
```
Paul-Licameli, line 375: Does it ever divide by 0?

Paul-Licameli:
Done with this round of review!

I await a rebase onto #5138 and conflict resolution.

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: Ok.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: It might easily be confused for `WaveTrack::SnapToSample`, though. This manipulation is safe here in this `RoundToNearestClipSample` function, I'd leave it as it is if there aren't other needs elsewhere for it.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: I can't remember. Just did a quick test with regular rounding, works too. I'll use a simple rounding, and next time should comment.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: I'm not sure I understand your question.
`alreadyRead` means `[buffer, buffer + alreadyRead)` are indices that already have been dealt with.
`startSamp` tells which first sample index of the clip matches `buffer + alreadyRead`.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2339,6 +2339,202 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
...
+   }
+   else
+   {
+      const auto endTime = std::min(
+         startOrEndTime - clip.GetPlayStartTime(), clip.GetPlayDuration());
+      const sampleCount endSamp { std::ceil(
+         std::round(endTime * sampsPerSec * 1e6) / 1e6) };
+      const auto startSamp =
+         std::max(endSamp - remainingToRead, sampleCount { 0 });
+      const auto len = (endSamp - startSamp).as_size_t();
```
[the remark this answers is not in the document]
saintmatthieu: Will comment.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: If `min` did some work, then `startOrEndTime` was past the end of the clip.
> should the first member of the return be a later position?

It may be confusing to have this buffer offsetting amidst this more complex code. All the offset is doing is moving the read-write pointer according to the number of samples already read/written. It should not be concerned with the number of samples remaining to read from the clip.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: Ok (for both)

src/tracks/playabletrack/wavetrack/ui/SampleHandle.cpp
saintmatthieu: Good question. The code before didn't check, but I cannot guarantee that my changes do not change that particular thing.
Adding a check upstream.

saintmatthieu:
@Paul-Licameli finished addressing your comments.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -198,6 +198,55 @@ void WaveClip::SetSamples(size_t ii,
...
+   const auto effectiveStart = std::max(sampleCount { 0 }, maybeNegativeStart);
+   const auto effectiveEnd =
+      std::min(GetVisibleSampleCount(), maybeOutOfBoundEnd);
+   // Cannot be greater than `numFloats` -> safe cast
+   const auto effectiveLen = (effectiveEnd - effectiveStart).as_size_t();
+   if (effectiveLen == 0)
+      return;
+   // Cannot be greater than `numFloats` -> safe cast
+   const auto numLeadingZeros =
+      (effectiveStart - maybeNegativeStart).as_size_t();
```
Paul-Licameli, line 227: Actually, your informal proof here is wrong.  What if `t` happens to be a very negative value?

So to be safe you should write
`const auto numLeadingZeros = limitSampleBufferSize(numFloats, effectiveStart - maybeNegativeStart);`

libraries/lib-wave-track/WaveClip.cpp
Paul-Licameli: See latest in the conversation

libraries/lib-wave-track/WaveClip.h
```diff
@@ -326,6 +338,43 @@ class WAVE_TRACK_API WaveClip final :
...
 
+   /*!
+    * @param t relative to clip start sample
+    */
+   bool
+   GetFloatAtTime(double t, size_t iChannel, float& value, bool mayThrow) const;
+
+   //! Succeed with out-of-bounds requests, only changing what is in bounds.
+   //! @{
+   // clan-format off
```
Paul-Licameli, line 349: You mean `clang-format` here and again below

libraries/lib-wave-track/WaveClip.h
```diff
@@ -326,6 +338,43 @@ class WAVE_TRACK_API WaveClip final :
...
+   // clan-format on
+   void SetFloatsFromTime(
+      double t, size_t iChannel, const float* buffer, size_t numSamples,
+      sampleFormat effectiveFormat);
+
+   /*!
+    * @brief Same as `SetFloatsFromTime`, but with `buffer` starting at
+    * `TimeToSamples(t0 -  SamplesToTime(numSideSamples))`.
+    * `[buffer, buffer + 2 * numSizeSamples + 1)` is assumed to be a valid span
+    * of addressed.
```
Paul-Licameli, line 368: "addresses"

Paul-Licameli:
Finished this review round

libraries/lib-wave-track/WaveClip.h
saintmatthieu: No, I meant `clan-format` XD

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: By substitution, `effectiveStart - maybeNegativeStart == std::max(sampleCount { 0 }, maybeNegativeStart) - maybeNegativeStart`,
and `maybeNegativeStart = TimeToSamples(t)`. So if `t < 0`, then `effectiveStart - maybeNegativeStart = - maybeNegativeStart`, that is, a positive value.

For `t` to be negative is actually not a corner case, here, and the code has proven to work well as far as my testing went, which is another evidence that it is correct.

Paul-Licameli (approved):
I'm still not really satisfied about the two uses of `as_size_t()`.  I'm dissatisfied now about the first, because,

`(effectiveEnd - effectiveStart)`

is bounded above by a `size_t` but there still isn't proof it is bounded below by 0 -- not negative.  You should first assign `effectiveEnd = std::max(effectiveStart, effectiveEnd);`

I am not persuaded either by the argument about the other `as_size_t()` and would rather see the `limitSampleBufferSize`.

I think then that without the other easy code changes, the function is still relying about on assumptions about its arguments so it has an incompletely stated contract.

I'd rather be deductive than argue that it works in practice so far.

But this PR was waited long enough.  I'm approving it, but with a protest.

saintmatthieu:
I was wrong about the first use: if `t` is past the end boundary of the clip, `effectiveEnd` can indeed be greater than `effectiveStart`. I couldn't reproduce this situation by trying even drag-drawing across several minuscule clips, but changing the value of `t` in the debugger showed it was possible. Thank you. Will push a safety check.

saintmatthieu:
Unit-testing would have been an easy way of demonstrating I was wrong.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -218,10 +218,9 @@ void WaveClip::SetFloatsFromTime(
    const auto effectiveStart = std::max(sampleCount { 0 }, maybeNegativeStart);
    const auto effectiveEnd =
       std::min(GetVisibleSampleCount(), maybeOutOfBoundEnd);
-   // Cannot be greater than `numFloats` -> safe cast
-   const auto effectiveLen = (effectiveEnd - effectiveStart).as_size_t();
-   if (effectiveLen == 0)
+   if (effectiveStart >= effectiveEnd)
       return;
+   const auto effectiveLen = (effectiveEnd - effectiveStart).as_size_t();
```
Paul-Licameli, line 223: Keep the comment of old line 221

Paul-Licameli:
I see and approve the latest three commits.  Correct spelling of "rectangle" in the last commit comment.

But there are conflicts with master.

saintmatthieu:
Thanks for the speedy review.

dozzzzer:
Tested.
Found issues: #5231 
Also, I've discovered that when using the Pencil tool on adjacent clips, then stretching them and mixing and rendering the track, it produces a waveform that looks different from the pre-mixed-and-rendered state:


https://github.com/audacity/audacity/assets/70063970/ef7b7985-c065-4685-aefb-114d07f5c879


https://github.com/audacity/audacity/assets/70063970/5bea899e-e5bd-4690-aa6f-540e658fea20


https://github.com/audacity/audacity/assets/70063970/b7988d7e-1cc1-494a-9439-919869604cb0


Despite this mismatch doesn't look particularly good to me, this might be insignificant and I reckon should not delay the merge.

PR #5044, opened 2023-08-15
Time stretching 5 of 6 repair effects and generators for stretched audio

### LINKED ISSUE #4850: Audio clip stretching

saintmatthieu:
Resolves: #4850

This PR repairs destructive effects for stretched clips.

Recommended:

## QA
All effects are supposed to work. Trying all of them may be too much, but a variety of types of effects should be tried (time-related effect, Nyquist, EQs and filters, noise removal, pitch shift, spectral tools, etc. Of course try them on stretched data, but also on stereo data, etc.

Note that applying an effect on a partial clip selection will create a new clip if the clip was stretched. That's wanted. The old behavior is only preserved on unstretched data.

- [ ] AutoDuck
- [ ] Change Speed
- [ ] Click Removal
- [ ] EQ
- [ ] Generate silence
- [ ] Loudness
- [ ] Noise Reduction
- [ ] Normalize
- [ ] Paulstretch
- [ ] Repair (Since this requires up to 128 samples only, the result will probably only be silence, and that's ok: it's a known limitation of the algorithm that rendering very tiny portions of audio only output silence)
- [ ] Repeat
- [ ] Reverse
- [ ] Change Tempo and Sliding Scale, high and low quality
- [ ] Stereo To Mono (previously reported by @dozzer as https://github.com/audacity/audacity/issues/5118)
- [ ] Truncate Silence
- [ ] Compressor
- [ ] Some Nyquist effects and generators
- [ ] At least one "Per Track Effect" example (any of the realtime capable, built-in or plug-in effects, but used destructively)

### DIFF
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -50,7 +50,7 @@ Track::Track(const Track& orig, ProtectedCreationArg&&)
 void Track::Init(const Track &orig)
 {
    mId = orig.mId;
-
+   mProjectTempo = orig.mProjectTempo;
    // Deep copy of any group data
    mpGroupData = orig.mpGroupData ?
       std::make_unique<ChannelGroupData>(*orig.mpGroupData) : nullptr;
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -93,11 +93,12 @@ void Track::EnsureVisible( bool modifyState )
       pList->EnsureVisibleEvent(SharedPointer(), modifyState);
 }
 
-TrackListHolder Track::Duplicate() const
+TrackListHolder Track::Duplicate(
+   std::optional<std::pair<double, double>> unstretchInterval) const
 {
    assert(IsLeader());
    // invoke "virtual constructor" to copy track object proper:
-   auto result = Clone();
+   auto result = Clone(std::move(unstretchInterval));
 
    auto iter = TrackList::Channels(*result->begin()).begin();
    const auto copyOne = [&](const Track *pChannel){
@@ -1004,7 +1005,7 @@ TrackList::RegisterPendingChangedTrack(Updater updater, Track *src)
    TrackListHolder tracks;
    std::vector<Track*> result;
    if (src) {
-      tracks = src->Clone(); // not duplicate
+      tracks = src->Clone(std::nullopt); // not duplicate
       assert(src->NChannels() == tracks->NChannels());
    }
    if (src) {
--- a/src/effects/Generator.cpp
+++ b/src/effects/Generator.cpp
@@ -30,7 +30,7 @@ bool Generator::Process(EffectInstance &, EffectSettings &settings)
 
    // Set up mOutputTracks.
    // This effect needs all for sync-lock grouping.
-   EffectOutputTracks outputs{ *mTracks, true };
+   EffectOutputTracks outputs{ *mTracks, {mT0, mT1}, true };
 
    // Iterate over the tracks
    bool bGoodResult = true;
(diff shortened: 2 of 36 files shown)


src/effects/nyquist/Nyquist.cpp
```diff
@@ -729,7 +729,7 @@ bool NyquistEffect::Process(EffectInstance &, EffectSettings &settings)
    // to operate on the selected wave tracks
    std::optional<EffectOutputTracks> oOutputs;
    if (!bOnePassTool)
-      oOutputs.emplace(*mTracks, true);
+      oOutputs.emplace(*mTracks, std::pair<double, double> { mT0, mT1 }, true);
```
Paul-Licameli, line 732: You might simplify the expression

libraries/lib-track/Track.cpp
```diff
@@ -988,7 +989,7 @@ TrackList::RegisterPendingChangedTrack(Updater updater, Track *src)
    TrackListHolder tracks;
    std::vector<Track*> result;
    if (src) {
-      tracks = src->Clone(); // not duplicate
+      tracks = src->Clone(std::nullopt); // not duplicate
```
Paul-Licameli, line 992: I think you can just write `{}` here and elsewhere that you wrote `nullopt`

Paul-Licameli:
I'm reviewing the PRs out of order.  Some of them are smaller, more fresh in my mind from recent review of the closed branch, and less problematic.

This one looks good.  I added to the QA checklist.

Paul-Licameli (approved):

src/effects/nyquist/Nyquist.cpp
Paul-Licameli: I repeat this

src/effects/Generator.cpp
```diff
@@ -71,7 +71,7 @@ bool Generator::Process(EffectInstance &, EffectSettings &settings)
                   ViewInfo::Get(*pProject).selectedRegion;
                track.ClearAndPaste(
                   selectedRegion.t0(), selectedRegion.t1(),
-                  *list, true, false, &warper);
+                  *list, true, true, &warper);
```
Paul-Licameli, line 74: Can you explain why this change is correct?

saintmatthieu: `merge` will become `mergeIfPossible` in `PasteOne`. This preserves the old behavior if possible, i.e., if `selectedRegion` spans clips of stretch ratio 1.

src/effects/nyquist/Nyquist.cpp
saintmatthieu: I haven't found how ...

src/effects/Generator.cpp
Paul-Licameli: I don't accept that explanation.

Compare behavior before and after this PR:
- Generate noise.
- Split the clip.
- Select around the split.
- Generate silence.

There is a change of behavior, as shown below.  But I thought the intention, with effects and generators (unlike with copy/paste), and no stretches, was to change no behavior.

Result, before this PR:

image

After:

image

Paul-Licameli (changes requested):

Paul-Licameli:
I think the correct solution may require a third boolean argument.  The `merge` argument in master was never passed into `PasteOne` and governed something other than what happens and the edges of the insertion.

saintmatthieu:
Things to fix on this branch:
* Auto Duck
* Generate silence over stretched clip
* Noise reduction
* Change tempo
* Change pitch
* Stereo to mono
* Nyquist noise gate (and other nyquist effects probably too)

src/effects/nyquist/Nyquist.cpp
Paul-Licameli: I was wrong!  It's `emplace` and not a direct use of the contstructor

Paul-Licameli:
I think we can agree to close the PR, as #5138 is its replacement.

PR #5043, opened 2023-08-15
Time stretching 4 of 6 repair copy paste for stretched tracks

saintmatthieu:
Resolves: #4850 
Resolves: #5273

Copying a clip of a given stretch ratio onto a clip of another stretch ratio cannot work as it has been so far. Instead of rendering the time stretching, we create new clips.

Note that this PR also changes the behavior if clips have equal ratios, since this was an ask from Product (@Tantacrul @LWinterberg ) and a low-hanging fruit for this PR.

Also note that this PR removes the feature whereby clicking the junction of two clips merges them, as a provisional step towards https://github.com/audacity/audacity/issues/2330.

- [x] I signed [CLA](https://www.audacityteam.org/cla/)
- [x] The title of the pull request describes an issue it addresses
- [x] If changes are extensive, then there is a sequence of easily reviewable commits
- [x] Each commit's message describes its purpose and effects
- [x] There are no behavior changes unnecessary for the stated purpose of the PR

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes of behavior

## QA

Please try a myriad of copy, paste and join operations, including trimmed clips, stretched clips, ...
Copy-paste operations should be tried with the "Editing a clip can move other clips" preference both on and off.
- [x] Cross-project copy-paste with difference of project tempo; observe time selection after paste, and stretching of the insertion which plays back at the new tempo
- [x] ~~Pasting a smart clip at the exact end of another clip preserves right-hand hidden data, loses (doesn't paste) left-hand~~ (The behaviour should be kept as in current release, that is, trimmed data should remain recoverable by untrimming).
**Peter 05Oct23 testing with latest 4 of 6 branch:**  The entire smart clip appears to be pasted. I had  30 second chirp and smart clipped it  by 10 seconds each side .  Then pasted it adjacent to  the right end of a 30 second tone.  in situ I could re-expand the chirp on the right.  I could expand it on the left in situ, but if I moved it rightwards away from the  tone I could re-expand it on the left too.  I absolutely think this is correct/desired behavior, the entire smart clip should be retained,
- [x] ~~Vice-versa, when pasting at the exact start of another clip~~
**Peter 05Oct23 testing with latest 4 of 6 branch:**  similarly so
- [x] Join stretched audio clips with equal stretch ratios doesn't render stretching
**Peter 05Oct23 testing with latest 4 of 6 branch:**  it still does render to 100%
- [x] Join audio clips with different stretch ratios shows progress bar
[...]

PR #4881, opened 2023-07-06
Audio Clip stretching

saintmatthieu:
Resolves: #4850

- [x] I signed [CLA](https://www.audacityteam.org/cla/)
- [x] The title of the pull request describes an issue it addresses
- [x] If changes are extensive, then there is a sequence of easily reviewable commits
- [x] Each commit's message describes its purpose and effects
- [x] There are no behavior changes unnecessary for the stated purpose of the PR

Recommended:
- [ ] Each commit compiles and runs on my machine without known undesirable changes of behavior

## Summary

This PR enables stretching audio clips. It is a significant change, as now the assumption that spacing between a track's samples is constant doesn't hold anymore. This has consequences in many places, probably a large part of which I haven't even considered.

## Testing

These changes probably require an entire regression test of the application. Note that destructive effects are likely to break. Here is a list of those known to break - please fill as you find out. Tickets will have to be created.
* Volume and Compression
  - [ ] Amplify
- [ ] Dragging left or right stretch handle of a clip is limited by other clips in the track and the opposite edge of the same clip
- [ ] No regression of the fix for: https://bugzilla.audacityteam.org/show_bug.cgi?id=2288
- [ ] Make cutlines, some of them including envelope points; stretch the containing clip; see cutlines move correctly, see cutlines and their envelopes expand correctly
- [ ] Generating inserts new clip boundaries into existing clip (this is an intentional behavior change)
- [ ] But effects applied to a selection do NOT insert new clip boundaries
* (to be continued)

We haven't identified anything else that this PR would break. Things that still should work I can think of include
- [ ] resample
- [ ] set silence
- [ ] spectrum displays in different views (FFT, auto-correlation, reassignment, etc.)
- [ ] drawing, brushing samples, including close to and beyond clip edges
- [ ] drawing in combination with clip envelope
- [ ] copy from one track into another track with different stretch ratio (the pasted audio becomes a new clip in the destination track) and across projects with different tempo.
- [ ] join clips with different tempo.
- [ ] also works with sync-lock on?
- [ ] copy/paste in all flavors: w/wo preference "Editing a clip can move other clips", sync-lock, cut lines, etc., copying entire tracks, entire clips, clip slices only, then pasting at cursor before, at start, within and at end of clips, across clip boundaries, etc.
- [ ] copy/paste, including cross-project, and undo/redo WITH change of project tempo, for ALL track types (label, time, and MIDI too), observing correct stretching

PR #5045, opened 2023-08-15
Time stretching 6 of 6 progress bars for stretching operations

### LINKED ISSUE #4850: Audio clip stretching

saintmatthieu:
Resolves: #4850

Since #5043, merging clips of different stretch ratios works again, by applying stretching to get consistent data before merging.
Since #5044, a similar principle is used to repair effects: the selected audio's time-stretching is first computed, then the effect applied.

In both cases, the time stretching can take noticeable time if the amount of data to stretch is big. Before this PR, no notification of this was given to the user, and she might wonder why Audacity is frozen until e.g. the join operation is complete.

This PR introduces progress dialogs for these situations.

QA (try these with large clips, i.e., maybe one minute long at least):
- [ ] a progress bar shows up when joining clips of contrasting stretch ratios.
- [ ] Joining of clips of equal stretch ratios should involve no stretching and hence be super fast. To stretch several clips by the same amount, easiest is to change project tempo. (Both Edit > Labeled Audio > Join and Edit > Audio Clips > Join)
- [ ] Applying an effect on a stretched clip brings a new progress bar reporting stretching progress.
- [ ] The progress bar filling-up is correctly proportioned to the number of tracks that are affected by the operation

Programmatically, the progress bars pop up whether stretching or not, but when not stretched, that should be barely noticeable.

Recommended:

### DIFF
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -15,8 +15,9 @@
 int EffectOutputTracks::nEffectsDone = 0;
 
 EffectOutputTracks::EffectOutputTracks(
-   TrackList &tracks, bool allSyncLockSelected
-)  : mTracks{ tracks }
+   TrackList& tracks, std::pair<double, double> effectTimeInterval,
+   bool allSyncLockSelected)
+    : mTracks { tracks }
 {
    // Reset map
    mIMap.clear();
@@ -31,7 +32,7 @@ EffectOutputTracks::EffectOutputTracks(
       };
 
    for (auto aTrack : trackRange) {
-      auto list = aTrack->Duplicate();
+      auto list = aTrack->Duplicate(effectTimeInterval);
       mIMap.push_back(aTrack);
       mOMap.push_back(*list->begin());
       mOutputTracks->Append(std::move(*list));
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -8,6 +8,7 @@
 
 **********************************************************************/
 #include "EffectOutputTracks.h"
+#include "BasicUI.h"
 #include "SyncLock.h"
 #include "WaveTrack.h"
 
@@ -31,11 +32,20 @@ EffectOutputTracks::EffectOutputTracks(
          : dynamic_cast<const WaveTrack*>(pTrack) && pTrack->GetSelected();
       };
 
+   auto progress = BasicUI::MakeProgress(
+      XO("Pre-processing"), XO("Rendering Time-Stretched Audio"));
+   const auto numTracks = trackRange.size();
+   auto count = 0;
+   auto reportProgress = [&](double progressFraction) {
+      const auto overallProgress = (count + progressFraction) / numTracks;
+      progress->Poll(overallProgress * 1000, 1000);
+   };
    for (auto aTrack : trackRange) {
-      auto list = aTrack->Duplicate(effectTimeInterval);
+      auto list = aTrack->Duplicate(effectTimeInterval, reportProgress);
       mIMap.push_back(aTrack);
       mOMap.push_back(*list->begin());
       mOutputTracks->Append(std::move(*list));
+      ++count;
    }
    // Invariant is established
    assert(mIMap.size() == mOutputTracks->Size());
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -44,6 +44,8 @@ using TrackListHolder = std::shared_ptr<TrackList>;
 struct UndoStackElem;
 
 using ListOfTracks = std::list< std::shared_ptr< Track > >;
+using TimeInterval = std::pair<double, double>;
+using ProgressReporter = std::function<void(double)>;
 
 //! Pairs a std::list iterator and a pointer to a list, for comparison purposes
 /*! Compare owning lists first, and only if same, then the iterators;
@@ -306,8 +308,8 @@ private:
     @post result: `NChannels() == result->NChannels()`
     */
    virtual TrackListHolder Duplicate(
-      std::optional<std::pair<double, double>> unstretchInterval =
-         std::nullopt) const;
+      std::optional<TimeInterval> unstretchInterval = std::nullopt,
+      ProgressReporter reportProgress = {}) const;
 
    //! Name is always the same for all channels of a group
    const wxString &GetName() const;
@@ -404,8 +406,9 @@ private:
     @pre `IsLeader()`
     @post result: `NChannels() == result->NChannels()`
     */
-   virtual TrackListHolder
-   Clone(std::optional<std::pair<double, double>> unstretchInterval) const = 0;
+   virtual TrackListHolder Clone(
+      std::optional<TimeInterval> unstretchInterval = {},
+      ProgressReporter reportProgress = {}) const = 0;
 
    template<typename T>
       friend std::enable_if_t< std::is_pointer_v<T>, T >
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -478,8 +478,10 @@ public:
    // Return true iff the attribute is recognized.
    bool HandleCommonXMLAttribute(const std::string_view& attr, const XMLAttributeValueView& valueView);
 
-protected:
    const std::optional<double>& GetProjectTempo() const;
+
+protected:
+   std::optional<double> mProjectTempo;
 
    /*!
     @pre `IsLeader()`
@@ -586,7 +588,7 @@ public:
     @tparam F returns a shared pointer to Attachment (or some subtype of it)
 
     @pre `f` never returns null
-   
+
     `f` may assume the precondition that the given channel index is less than
     the given track's number of channels
     */
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -305,7 +305,9 @@ private:
     @pre `IsLeader()`
     @post result: `NChannels() == result->NChannels()`
     */
-   virtual TrackListHolder Duplicate() const;
+   virtual TrackListHolder Duplicate(
+      std::optional<std::pair<double, double>> unstretchInterval =
+         std::nullopt) const;
 
    //! Name is always the same for all channels of a group
    const wxString &GetName() const;
@@ -398,10 +400,12 @@ private:
    //! Subclass responsibility implements only a part of Duplicate(), copying
    //! the track data proper (not associated data such as for groups and views)
    /*!
+    @param unstretchInterval If set, this time interval's stretching must be applied.
     @pre `IsLeader()`
     @post result: `NChannels() == result->NChannels()`
     */
-   virtual TrackListHolder Clone() const = 0;
+   virtual TrackListHolder
+   Clone(std::optional<std::pair<double, double>> unstretchInterval) const = 0;
 
    template<typename T>
       friend std::enable_if_t< std::is_pointer_v<T>, T >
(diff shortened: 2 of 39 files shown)


Paul-Licameli:
"Programmatically, the progress bars pop up whether stretching or not, but when not stretched, that should be barely noticeable."

Actually Audacity's ProgressDialog object is smart enough to construct itself but delay the initial display of itself on the screen.  You won't even see it for an instant if the operation completes in less than the minimum time.

Paul-Licameli:
I made some comments at #4881 about these recent additions.  Some are still not addressed.

libraries/lib-effects/EffectOutputTracks.cpp
```diff
@@ -31,11 +32,20 @@ EffectOutputTracks::EffectOutputTracks(
          : dynamic_cast<const WaveTrack*>(pTrack) && pTrack->GetSelected();
       };
 
+   auto progress = BasicUI::MakeProgress(
+      XO("Time Stretching"),
+      XO("First, time stretching must be rendered onto selected audio."));
```
Paul-Licameli, line 37: Ask @LWinterberg whether the exact words are OK wherever this PR uses `XO`

libraries/lib-effects/EffectOutputTracks.cpp
```diff
@@ -31,11 +32,20 @@ EffectOutputTracks::EffectOutputTracks(
...
+   auto reportProgress = [&](double progressFraction) {
+      progress->Poll(progressFraction * 1000, 1000);
+   };
    for (auto aTrack : trackRange) {
-      auto list = aTrack->Duplicate(effectTimeInterval);
+      auto list = aTrack->Duplicate(effectTimeInterval, reportProgress);
       mIMap.push_back(aTrack);
       mOMap.push_back(*list->begin());
       mOutputTracks->Append(std::move(*list));
+      ++count;
```
Paul-Licameli, line 48: `count` is not used, did you mean to use it?

libraries/lib-track/Track.h
```diff
@@ -306,8 +306,8 @@ class TRACK_API Track /* not final */
     @post result: `NChannels() == result->NChannels()`
     */
    virtual TrackListHolder Duplicate(
-      std::optional<std::pair<double, double>> unstretchInterval =
-         std::nullopt) const;
+      std::optional<std::pair<double, double>> unstretchInterval = std::nullopt,
+      std::function<void(double)> reportProgress = [](double) {}) const;
```
Paul-Licameli, line 310: As I said elsewhere, I would:
- Use type aliases like `using UnstretchInterval = std::pair<double, double>; using ProgressReporter = std::function<void(double)>;` and simplify other usages
- Default the function to '{}', not an empty lambda.  But then wherever the function is ultimately used in WaveTrack or WaveClip, test first that is not null.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1221,6 +1222,8 @@ void WaveClip::ApplyStretchRatio()
             widestSampleFormat /* computed samples need dither */
          );
       numOutSamples += numSamplesToGet;
+      reportProgress(
```
Paul-Licameli, line 1242: One null check would be here

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2822,6 +2839,19 @@ int WaveTrack::GetNumClips() const
...
 }
 
+int WaveTrack::GetNumClips(double t0, double t1) const
+{
+   const auto clips = SortedClipArray();
+   const auto firstIn =
+      std::find_if(clips.begin(), clips.end(), [t0](const auto& clip) {
+         return t0 < clip->GetPlayEndTime();
+      });
+   const auto firstOut = std::find_if(firstIn, clips.end(), [t1](const auto& clip) {
```
Paul-Licameli, line 2870: You might use `std::lower_bound` twicd instead of the linear searching `find_if`

Paul-Licameli:
I will wait for your replies before I approve.

Paul-Licameli:
Oh, but I'm not finished with review round 1!  There is a second commit.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -644,10 +645,16 @@ bool WaveClip::Paste(double t0, const WaveClip &other)
 
    Transaction transaction{ *this };
 
+   const auto otherStretchRatio = other.GetStretchRatio();
+   if (GetSequenceSamplesCount() == sampleCount{0})
+      // This is empty: use stretch ratio of other clip to avoid unnecessary stretching.
+      ApplyStretchRatio(otherStretchRatio, [](double) {});
+   const auto thisStretchRatio = GetStretchRatio();
+   const bool clipNeedsStretching =
+      fabs(thisStretchRatio - other.GetStretchRatio()) > 1e-6;
```
Paul-Licameli, line 654: Could there be a smarter choice of tolerance?  Maybe calculate whether the applied stretch times duration makes more than `1 / other.GetRate()`

src/menus/LabelMenus.cpp
```diff
@@ -591,9 +592,19 @@ void OnJoinLabels(const CommandContext &context)
    if (selectedRegion.isPoint())
       return;
 
+   const auto progress =
+      BasicUI::MakeProgress(XO("Joining"), XO("Time-stretching..."));
+   auto count = 0;
+   const auto numWaveTracks = tracks.Any<WaveTrack>().size();
+   const auto reportTrackProgress = [&](double trackProgress) {
+      const auto overallProgress = (count + trackProgress) / numWaveTracks;
```
Paul-Licameli, line 600: Post-increment count?

src/menus/EditMenus.cpp
```diff
@@ -903,9 +903,17 @@ void OnJoin(const CommandContext &context)
...
-      wt->Join(selectedRegion.t0(), selectedRegion.t1());
+   const auto progress =
+      BasicUI::MakeProgress(XO("Joining"), XO("Time-stretching..."), 0);
+   const auto selectedTracks = tracks.Selected<WaveTrack>();
+   auto count = 0;
+   for (auto wt : selectedTracks)
+      wt->Join(
+         selectedRegion.t0(), selectedRegion.t1(), [&](double trackProgress) {
+            const auto overallProgress =
+               (count + trackProgress) / selectedTracks.size();
```
Paul-Licameli, line 914: post-increment count?

libraries/lib-wave-track/PasteOverPreservingClips.cpp
```diff
@@ -50,9 +50,10 @@ ClipData CollectClipData(
    return results;
 }
 
-void PasteOverPreservingClips(const ClipData &data,
-   WaveTrack &oldTrack, sampleCount start, sampleCount len,
-   const WaveTrack &newContents)
+void PasteOverPreservingClips(
+   const ClipData& data, WaveTrack& oldTrack, sampleCount start,
+   sampleCount len, const WaveTrack& newContents,
+   std::function<void(double)> reportProgress)
```
Paul-Licameli, line 56: I mentioned in another review today that there is remapping of split lines in some of the functions but there is not-exactly-duplicated code doing it.

If that were unified, this would likely become the common utility.  But if only effects ever use this, they would always pass a no-op progress reporter anyway, for the same reasons as commented in EQ, right?

So maybe this function needs no new argument and just supplies the no-op to `Join` always.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1421,7 +1421,10 @@ void WaveTrack::BlindlyCopyTrackContents(double t0, const WaveTrack& other)
 void WaveTrack::BlindlyMergeClipFromOtherTrackInto(
    double t0, WaveClip& myClip, const WaveClip& otherTrackClip)
 {
-   myClip.Paste(t0 - myClip.GetPlayStartTime(), otherTrackClip);
+   // No need for a progress bar, since clips must have equal stretch ratios:
+   assert(
+      fabs(myClip.GetStretchRatio() - otherTrackClip.GetStretchRatio()) < 1e-6);
```
Paul-Licameli, line 1426: Another tolerance

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1183,9 +1199,9 @@ void WaveClip::ApplyStretchRatio(
       this->SetTrimRight(trimRightBeforeStretch);
       if (success)
       {
-         this->mClipStretchRatio = 1.0;
+         this->mClipStretchRatio = targetRatio;
          this->mRawAudioTempo = this->mProjectTempo;
-         assert(this->GetStretchRatio() == 1.0);
+         assert(this->GetStretchRatio() == targetRatio);
```
Paul-Licameli, line 1204: Keep the `@post` comment consistent with the assertion!

Paul-Licameli:
In general, for the second commit, as for the first, also watch out for `XO` and also pass null functions instead of empty lambdas.

That's one review round.

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: This check is meant to ignore numeric errors, which in theory could happen if you do the following: copy a clip, change project tempo multiple times, and paste the clip back. The clipboard clip will have undergone only one project tempo change, at the moment of paste, while the clip that stayed in the project, several.
Nevertheless, I just tested that, and both ratios still end up the exact same. So I'll exchange this for pure equality.

Paul-Licameli: > This check is meant to ignore numeric errors, which in theory could happen if you do the following: copy a clip, change project tempo multiple times, and paste the clip back. The clipboard clip will have undergone only one project tempo change, at the moment of paste, while the clip that stayed in the project, several. Nevertheless, I just tested that, and both ratios still end up the exact same. So I'll exchange this for pure equality.

Don't do that.  Never compare results of floating point computations for equality.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: Removed.

Paul-Licameli: > Removed.

NO!

Paul-Licameli: Use a tolerance!  But factor some common function or constant to decide what the tolerance is.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -685,23 +692,29 @@ bool WaveClip::Paste(double t0, const WaveClip &other)
...
+      if (clipNeedsStretching && !stretchBeforeResampling)
+         copy->ApplyStretchRatio(thisStretchRatio, reportProgress);
+
       if (clipNeedsNewFormat)
          // Force sample formats to match.
          copy->ConvertToSampleFormat(GetSampleFormats().Stored());
       newClip = std::move(copy);
    }
-   ApplyStretchRatio([](double) {});
-   newClip->ApplyStretchRatio([](double) {});
```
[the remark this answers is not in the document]
saintmatthieu: Performance-wise, higher stretch ratios will lead to more synthesis windows and augment the processing time, which makes the estimation more complicated than just sample-count comparison.
In terms of quality loss, stretch ratios closer to 1 may result in less loss. Our stretcher has a tendency to shift samples left or right, depending on whether you are compressing or extending the clip - I haven't tried to find out why, but that may come up in user feedback. That _could_ be a selection criterion.
However, UX-wise the most sensible I'd say is to say "whatever you paste adapts itself", which you correctly recognized as what I'm doing here.

saintmatthieu:
@Paul-Licameli your comments from round 1 are addressed.

saintmatthieu:
This PR doesn't add progress bars that aren't there in #5138 or #5043 anymore. Closing.

PR #5049, opened 2023-08-17
Update Compressor.cpp

dseo369:
included descriptions of each variable in the effect description.

Recommended:


LWinterberg:
the text you're trying to change here is this one I think:

image

I don't think it's a particularly good place to put this kind of info, as people would never think to check "presets and settings" to  find info on what each slider does. We don't have a good system for this currently in Audacity, and https://manual.audacityteam.org/man/compressor.html has this information already

Issue #5058, opened 2023-08-19
3.4.0 Spectral view of a stereo project fails for the right channel


petersampsonaudacity:
### Bug description

In 3.4.0 alpha Spectral view of a stereo project fails for the right channel - it duplicates the view of the left channel.

### Steps to reproduce

1. generate a 30 second chirp
2. add new mono track
3. generate a 30 second DTMF
4. make these into a stereo track
5. Observe: right channel chirp and left channel DTMF
6. Change the track view to spectrogram
7. Observe: both channels are displaying the left channel data
8. Change track view to multiview
9. Observe: right channel shows DTMF waveform BUT chirp spectrogram

### Expected behavior

This is 3.3.3
![Spectral 3-3-3](https://github.com/audacity/audacity/assets/50205984/d8d10e9a-d03d-4be5-83d0-7be28eb0bcec)

### Actual behavior

![Spectral test project](https://github.com/audacity/audacity/assets/50205984/1a396751-2b18-41e7-b503-7489f934f1a7)
![Spectral 3-4-0](https://github.com/audacity/audacity/assets/50205984/85d35cc5-757f-44a2-956d-0ce47900835e)
![Spectral 3-4-0 multi-view](https://github.com/audacity/audacity/assets/50205984/796486a5-bb95-4c62-86aa-4beadbcdb41e)

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

See also: 
 **Most (nearly all) effects overwrite right channel data with left channel data #5055**

Paul-Licameli:
This and #5055 bisect to the same commit at 1698b5010e24f673eaa4bf1319e4f9f3c6b3cc01

dozzzzer:
Tested

Issue #5055, opened 2023-08-18
Most (nearly all) effects overwrite right channel data with left channel data


deus-e-x:
### Bug description

Applying most (nearly all) effects to a stereo track copies the normalized left track data to the right channel, overwriting the entire original right channel data.

Repeat appears to be the sole exception.

Oddly this does **_NOT_** seem to affect **_some_** external plug-in effects - see further on in this thread ...

### Steps to reproduce

1. Import any stereo audio track
2. Navigate to the Effect menu and access the entry Normalize...
3. The entire stereo audio data gets autoselected
3. Apply an effect on the selected stereo audio data

### Expected behavior
The left and right channel audio data should be acted according to the selected parameters of the effect dialog.

### Actual behavior

Upon executing the effect, the left channel data is copied to the right channel, overwriting the entire original right channel data.

_**Audacity-win-3.3.3-x64, correct normalize-effect behavior:**_

![1](https://github.com/audacity/audacity/assets/45689656/053cdbc2-579c-42de-bbdf-f90826abedf0)

_**Audacity v3.4.0 alpha 20230817+b2626a1-x64, incorrect normalize-effect behavior:**_

![2](https://github.com/audacity/audacity/assets/45689656/e9ac3878-fc54-4a49-8405-d22045c3d428)

**_Normalize effect dialog settings, used for both versions above:_**

![3](https://github.com/audacity/audacity/assets/45689656/3cd361fa-169a-412f-9a90-e9cd0e5b1799)

### Audacity Version
3.4.0 alpha test build

### Operating system
Windows - but assume all OS

### Additional context
The **Tracks > Mix**   commands do **_NOT_** seem to be similarly affected, they work OK

petersampsonaudacity:
Testing on W10 with 3.3.3 I cannot reproduce this,

But with latest 3.4.0 alpha test build:  audacity-win-3.4.0-alpha-20230817+b2626a1-x64-msvc2022
This is readily reproducible.

I suspect that this may have something to do with the re-working of stereo tracks that @Paul-Licameli  is currently undertaking.

For my test track I created a track with a 30 second chirp in the left channel and a 30m second DTMF in the right channel so the results can be clearly see.  I tested with:
a) the setting for **Normalize stereo channels independently** set to **off** _(default setting)_
a) the setting for **Normalize stereo channels independently** set to **on**
Both worked correctly and 3.3.3 and failed with 3.4.0 alpha

[...]

Issue #5054, opened 2023-08-18
AutoDuck Effect applies unevenly to Stereo Tracks.


chinakov:
### Bug description

If the User uses AutoDuck effect in current Master, the left channel is processed correctly (the parameters set by the User in AutoDuck effect window are honoured), but the right channel is processed incorrectly (the parameters set by the User in AutoDuck effect window are not honoured).

### Steps to reproduce

1. Open Audacity
2. Import or Generate Audio in a Stereo Track
3. Create another track with some noise, for example
4. Apply AutoDuck effect to the first track created
5. Observe Left and Right channels.

### Expected behavior

Here is how the effect is applied in Audacity 3.3.3:
<img width="1617" alt="Screenshot 2023-08-18 at 14 20 59" src="https://github.com/audacity/audacity/assets/139755141/94e7561f-cd74-4bc0-9660-2427d1ad3e5b">

### Actual behavior

Here is what the effect does on Master (same project used):
<img width="1617" alt="Screenshot 2023-08-18 at 14 25 52" src="https://github.com/audacity/audacity/assets/139755141/a175f9e4-ca21-42df-89b3-a5e359438c3a">

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

petersampsonaudacity:
This is possible the same or related to:

 **Most (nearly all) effects overwrite right channel data with left channel data #5055**

chinakov:
@petersampsonaudacity - yes, this is not reproduced on Mono tracks, as shown in your example above. 

However, this is how it looks on Stereo on Paul's branch build: 
**Before:**
![image (1)](https://github.com/audacity/audacity/assets/139755141/d0df32b8-1943-40d5-8947-3921e988fac0)

**After:**
![image](https://github.com/audacity/audacity/assets/139755141/3be33128-9040-4d49-90c2-8869a3423b57)

Pay close attention to left and right channels here.

petersampsonaudacity:
> Pay close attention to left and right channels here.

@chinakov   -  I see that you have quieter ducks in the right channel

Hoever when I run a similar project on Paul's branchmine looks OK:
![image](https://github.com/audacity/audacity/assets/50205984/53f938b9-f81f-450b-9cc0-504d010ae93a)

Paul-Licameli:
> @chinakov @Paul-Licameli
> 
> This is possible the same or related to:
> 
> **Most (nearly all) effects overwrite right channel data with left channel data #5055**

[...]

PR #5060, opened 2023-08-20
Fix duplicated left channel spectrogram and most effect results...

Paul-Licameli:
... AutoDuck was affect too, but differently.  Right channel ended up with left channel's contents, with the fading applied twice.

Resolves: #5054
Resolves: #5055
Resolves: #5058

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -2285,7 +2285,8 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
    std::optional<TrackIter<const WaveTrack>> iter;
    auto pTrack = this;
    if (pOwner) {
-      iter.emplace(TrackList::Channels(this).first.advance(iChannel));
+      const auto ppLeader = TrackList::Channels(this).first;
+      iter.emplace(ppLeader.advance(IsLeader() ? iChannel : 1));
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
Testing on W10 with Paul's branch build:  audacity-win-3.4.0-alpha-20230820+9d21d5c-x64

Autoduck also appears to work fine:
image

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2285,7 +2285,8 @@ bool WaveTrack::Get(size_t iChannel, size_t nBuffers,
    std::optional<TrackIter<const WaveTrack>> iter;
    auto pTrack = this;
    if (pOwner) {
-      iter.emplace(TrackList::Channels(this).first.advance(iChannel));
+      const auto ppLeader = TrackList::Channels(this).first;
+      iter.emplace(ppLeader.advance(IsLeader() ? 0 : 1));
```
saintmatthieu, line 2289: What if `Get` is called on a leader track with `iChannel == 1` and `nBuffers == 1` ? It seems to me that this would still get the left-channel samples. My best guess is that `iter.emplace(ppLeader.advance(IsLeader() ? iChannel : 1)`.

saintmatthieu (approved):

Issue #4991, opened 2023-08-03
The pressed state of the "play" and "pause" buttons are not provided in Msaa.


paulber19:
### Bug description

Since version 3.2.2, it is no longer possible to know the pressed state of the "play", "Pause" buttons using MSAA. Only the name of the button indicates that the button is pressed, but this name is a translatable string and varies according to the language chosen by the user.
Knowing the status of these buttons is the only way to know that playback is in progress.
Until version 3.2.1 this was possible.

### Steps to reproduce

Using NVDA:
- focus playback button and press it
- - hit "nvda+f1" and look at state items.

### Expected behavior

Pressed button state must be accible by MSAA

### Actual behavior

Pressed button state is not available

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

ras

DavidBailes:
Hi Paul,
let me explain why these toggle buttons were changed.
Since the introduction of realtime effects, these toggle buttons are now also used as power toggle buttons in both the realtime effects pane, and in realtime effect dialogs, and it was important that they behaved for all the windows screen readers, Jaws, NVDA, and Narrator, something like how toggle buttons normally behave. The problems are that:
1. Narrator does not read the pressed state at all.
2. For Jaws and NVDA, the not pressed state is not explicitly read.
Because of this, I changed the toggle button so that the name of the button, its role and its state are included in the accessibility name.
Users can still read the state of the buttons, and the overall playback status is still provided in the status bar.
However, I can understand that there is now a problem robustly getting the state of of one of these buttons from a screen reader script.
One suggestion for solving this would for me to append a special character to the end of the accessibility name of the toggle button when it was in the pressed state. For example the bell character '\a' which is not read by screen readers. Then you could check the last character in the accessibility name to see whether it was pressed.
If this would be helpful, just let me know and I can put in a pull request for this. Or alternatively you may have some other suggestions for how to work around this problem.
David.

paulber19:
Hello David.
I fully understand your intention and your proposal is valid.
But what I don't understand is why indicating the "Pressed" state in the 
button name excludes positioning this state in the button state as it 
was in version 3.0.
Today, I worked around the problem by counting the word count of the 
button name, but that's not a reliable solution because it depends on 
the translation of that name.
I think it would be better to continue indicating this state as is the 
case for the "record" and "Stop" buttons and if possible, to emit an 
event when changing this state so that the readers of screen informs the 
user.
Best regards.
Paul.

[...]

PR #5066, opened 2023-08-22
Pressed state of the "play" and "pause" buttons are not provided in Msaa

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/4991

Problem:
In commit 67aad3504, the accessibility of AButtons when set to be toggle buttons was improved for all screen readers and for Narrator in particular. This involved both the role and state being included in the accessibility name. However, because the accessibility state no longer indicated the state of the toggle button, this made it difficult for screen reader scripts to robustly determine the state of the toggle button.

Fix:
Add the character '\a' to the end of the accessibility name when the toggle button is in the pressed state. Screen reader scripts can then check for this character.

Recommended:


crsib (approved):

DavidBailes:
@crsib . Thanks!

chinakov:
Tested

PR #5067, opened 2023-08-22
Qt basic UI

vsverchinsky:
Depends on:
 - #5046 

Recommended:

### DIFF
--- a/src/qt/ExtraMenu.cpp
+++ b/src/qt/ExtraMenu.cpp
@@ -1,0 +1,50 @@
+#include "ExtraMenu.h"
+
+#include "BasicUI.h"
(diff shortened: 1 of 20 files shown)


src/qt/ExtraMenu.cpp
```diff
@@ -0,0 +1,50 @@
...
+   return result;
+}
+
+void ExtraMenu::activate(int index)
+{
+   int counter { 0 };
+   ForEach([&](MenuItem& item)
+   {
+      if(counter == index)
+         item.action();
```
ksoze95, line 47: Possibly add a break after calling action?

ksoze95 (approved):

PR #5069, opened 2023-08-22
5065 wavetrack internal channel iteration rewrites

### LINKED ISSUE #5065: Easy WaveTrack-internal channel iteration rewrites (checklist)
Do easier eliminations of `TrackList::Channels` in `WaveTrack.cpp`.

In each of these functions, change the iteration so that the outer loop is over the `WaveTrack`'s Intervals, and the inner loop is inside a new member function of `Interval`, using `ForEachClip`.

Later reimplementation of those functions will be done when `WaveTrack::Interval` and `WaveClip` are ready to merge into one class.

Note too the files `WaveClipUtilities.*` introduced in #5138.  Some of the `WaveTrack` functions rewritten for this issue might become non-member functions added to those files if they are completely implementable in terms of the public interface of `WaveTrack` and `WaveTrack::Interval` and so the big file `WaveTrack.cpp` might be shrunk to only the more essential functions.

- [x] #5691
  - [x] MoveTo
  - [x] GetSequenceSamplesCount
  - [x] CountBlocks
  - [x] ConvertToSampleFormat
  - [x] Silence
  - [x] Flush (no outer loop, it applies only to one Interval)
  - [x] CloseLock
  - [x] WidestEffectiveFormat
  - [x] RemoveCutLine
  - [x] Resample
- [x] #5708
  - [x] HasHiddenData
  - [x] DiscardTrimmed
  - [x] VisitBlocks
  - [x] smartClipsExtension
  - [x] stretchedClipsExtension
- [x] #5764
  - [x] Trim
  - [x] InsertSilence
  - [x] Split

saintmatthieu:
Resolves: #5065 

Rewrite loops over channels without deprecated `TrackList::Channels`. (This may be part one of more PRs doing the same thing.)

Recommended:

QA:
- [ ] Playback of mono and stereo tracks


saintmatthieu:
@Paul-Licameli I'd like you to take a quick but early look at this PR. Commits are accumulating, and to my sense it's useful insofar as it will speed up the clean-up after we introduce wide clips, but I'd like confirmation that I'm going in the right direction. Thank you.

Paul-Licameli:
Here's my quick review of just the first three commits of this draft.

I'm not sure the first two commits are useful, or others I haven't read yet, using `EasyToRemoveCallToTrackListChannels`.

I do like the third commit, "WaveTrack::Silence rewrite".
- It eliminates one `TrackList::Channels`
- You discovered a needed extension of the `WaveTrack::Interval` interface
- A doubly nested loop in `WaveTrack::Silence` became single instead
- `WaveTrack::Interval::SetSilence` effectively hides the inner loop (unrolled)
- Looping changed from channel-major to interval-major

I don't yet see additions to `WaveChannelInterval` but maybe you will discover useful things there too, which I would like to see.

Take some time to permute the commit sequence to put the things I like first.

saintmatthieu:
I've reached a milestone in the PR: all commits up to now address uses of `TrackList::Channels` that will be trivial to remove. I don't require careful review of these commits: they just add comments explaining why they'll be trivial to remove. I don't intend to merge these changes, enough will be to know with confidence that removing all those `TrackList::Channels` calls will be easy.

Note that these trivial usages include the commits such as https://github.com/audacity/audacity/pull/5069/commits/e81291693b436054abe2e560f58c03cf37012a2d ("WaveTrack::Silence rewrite") : although these have changes (with e.g. methods added to `WaveTrack::Interval`), I learned during the process that these usages of `TrackList::Channels` were just as easy to remove from those methods as in others, i.e., roughly speaking by removing the outer loop over channels.

There remain two instance that seem to require non-trivial work: `WaveTrack::Get` and `WaveTrack::GetSampleView`. They should be addressed in upcoming commits.

Paul-Licameli:
Hey, why not cherry-pick the commit that removes `WaveTrack::GetSampleView` onto #5041, which will need some more work anyway before merge?

saintmatthieu:
That's it. The remaining used `TrackList::Channels` in `WaveTrack::Get` could be eliminated by making `Get` become a member of `WaveChannel`. Since `WaveTrack` is destined not to be a `WaveChannel` anymore, this move makes semantical sense.

Talking about semantics: this could be an argument to move some of the `WaveTrack` implementations to `WaveChannel`, such as `::Resample`. This still isn't a strong one, though, since we could just as well keep the implementation in `WaveTrack`, removing the `TrackList::Channels` iteration when tracks really are wide.

saintmatthieu:
> Hey, why not cherry-pick the commit that removes `WaveTrack::GetSampleView` onto #5041, which will need some more work anyway before merge?

Absolutely.

Paul-Licameli:
> I've reached a milestone in the PR: all commits up to now address uses of `TrackList::Channels` that will be trivial to remove. I don't require careful review of these commits: they just add comments explaining why they'll be trivial to remove. I don't intend to merge these changes, enough will be to know with confidence that removing all those `TrackList::Channels` calls will be easy.
> 
> Note that these trivial usages include the commits such as [e812916](https://github.com/audacity/audacity/commit/e81291693b436054abe2e560f58c03cf37012a2d) ("WaveTrack::Silence rewrite") : although these have changes (with e.g. methods added to `WaveTrack::Interval`), I learned during the process that these usages of `TrackList::Channels` were just as easy to remove from those methods as in others, i.e., roughly speaking by removing the outer loop over channels.
> 
> There remain two instance that seem to require non-trivial work: `WaveTrack::Get` and `WaveTrack::GetSampleView`. They should be addressed in upcoming commits.

I disagree with these opinions.

First, `GetSampleView` will just go away, as we now know, and the rewrite of `WaveTrack::Get` won't be merged, because it only duplicates something similar waiting for review in #5031.

Many of the commits are useless because they only "eliminate" `TrackList::Channels` by hiding it in a thin wrapper function.  You call these iterations "trivial to remove".  Hardly!  They are the ones that are difficult to remove, and not ready to remove now, because they are the ones that presuppose a lot of difficult data restructuring that still is in the future; and so all that was easily done now is a renaming of the function.

But the result of the branch is not useless!

The commits that lower one loop into another -- changing channel-major to interval-major iterations -- ARE the useful ones.  They are doing some useful, non-trivial code transformations that we should commit.  `WaveTrack::Silence` is one of about ten of them.

New things that need to be in the WaveTrack::Interval API were discovered.  That is not wasted work.

Also identifying a few remaining pieces of duplicate data in `WaveTrack` pairs -- color, format, origin -- which need accessors like `Track::GetProjectTempo` was no wasted work.  This could add a few more to that set of ten.

The insides of those new WaveTrack::Interval functions may be rewritten later.  I do not accept that as a reason not to define first versions of those functions now.  Incremental progress by an "outside-inward" discovery of better APIs, followed later by reimplementation of the better encapsulated details, has been my strategy for this big restructuring and I mean to continue that way.

saintmatthieu:
I still don't agree, which probably means that I'm missing something. Let's take the concrete example of `WaveTrack::Silence`.

Current code:
```cpp
   for (const auto pChannel : TrackList::Channels(this)) {
      for (const auto &clip : pChannel->mClips) {
         auto clipStart = clip->GetPlayStartSample();
         auto clipEnd = clip->GetPlayEndSample();
         if (clipEnd > start && clipStart < end) {
            auto offset = std::max(start - clipStart, sampleCount(0));
            // Clip sample region and Get/Put sample region overlap
            auto length = std::min(end, clipEnd) - (clipStart + offset);
            clip->SetSilence(offset, length);
         }
      }
   }
```
At some stage of this refactoring, there will be a PR that creates really stereo `WaveTrack`s. That PR could also make `TrackList::Channels` return a range of size 1 only. Then this code, as well as the other places I've visited in this PR, would still work, since you already made sure that `mClips` are stereo-capable.

Then, in a follow-up PR, the code above would be simplified to
```cpp
   for (const auto &clip : mClips) {
      auto clipStart = clip->GetPlayStartSample();
      auto clipEnd = clip->GetPlayEndSample();
      if (clipEnd > start && clipStart < end) {
         auto offset = std::max(start - clipStart, sampleCount(0));
         // Clip sample region and Get/Put sample region overlap
         auto length = std::min(end, clipEnd) - (clipStart + offset);
         clip->SetSilence(offset, length);
      }
   }
```
and `TrackList::Channels` would be gone.

What am I missing ?

Paul-Licameli:
See #5311 for the definition of `ForEachClip`

saintmatthieu:
#5311 doesn't match an existing PR.

saintmatthieu:
You meant #5133.

So you're suggesting
```
   for (const auto interval : Intervals()) {
      interval->ForEachClip([&](auto& clip) {
      auto clipStart = clip->GetPlayStartSample();
      auto clipEnd = clip->GetPlayEndSample();
      if (clipEnd > start && clipStart < end) {
         auto offset = std::max(start - clipStart, sampleCount(0));
         // Clip sample region and Get/Put sample region overlap
         auto length = std::min(end, clipEnd) - (clipStart + offset);
         clip->SetSilence(offset, length);
      }}
   }
```
Ok, it would be a slicker intermediate step than writing new APIs to `WaveTrack::Interval`. But at some stage you'll want to get rid of `WaveTrack::Interval::ForEachClip` too, right? So what advantage does this intermediary step have?

Paul-Licameli:
> You meant #5133.
> 
> So you're suggesting
> 
> ```
>    for (const auto interval : Intervals()) {
>       interval->ForEachClip([&](auto& clip) {
>       auto clipStart = clip->GetPlayStartSample();
>       auto clipEnd = clip->GetPlayEndSample();
>       if (clipEnd > start && clipStart < end) {
>          auto offset = std::max(start - clipStart, sampleCount(0));
>          // Clip sample region and Get/Put sample region overlap
>          auto length = std::min(end, clipEnd) - (clipStart + offset);
>          clip->SetSilence(offset, length);
>       }}
>    }
> ```
> 
> Ok, it would be a slicker intermediate step than writing new APIs to `WaveTrack::Interval`. But at some stage you'll want to get rid of `WaveTrack::Interval::ForEachClip` too, right? So what advantage does this intermediary step have?

I have suggested that `WaveTrack::Interval` and `WideClip` and `WaveClip` are three interfaces that need to be unified into one.

I think we are not ready yet to unify them.

But I think that rewriting some of these `WaveTrack` member functions now, in terms of an appropriately expanded `Interval` interface, is a small step jog progress.  We do it now, and then we won’t need to rewrite those `WaveTrack` members again.  We progress outside inward, discovering the interfaces we need, and later redoing the implementation details.

Paul-Licameli:
So please, do that work now, and it is so much less work to do later.

saintmatthieu:
I overviewed the fact that `ForEachClip` is private.
Hence a solution would be
```
void WaveTrack::Interval::SetSilence(sampleCount start, sampleCount end)
{
   ForEachClip([&](auto& clip) {
      const auto clipStart = clip.GetPlayStartSample();
      const auto clipEnd = clip.GetPlayEndSample();
      if (clipEnd > start && clipStart < end)
      {
         auto offset = std::max(start - clipStart, sampleCount(0));
         // Clip sample region and Get/Put sample region overlap
         auto length = std::min(end, clipEnd) - (clipStart + offset);
         clip.SetSilence(offset, length);
      }
   });
}
```
which, indeed, would give the `WaveTrack` code its final, post-refactoring look:
```
void WaveTrack::Silence(double t0, double t1)
{
   assert(IsLeader());
   if (t1 < t0)
      THROW_INCONSISTENCY_EXCEPTION;

   auto start = TimeToLongSamples(t0);
   auto end = TimeToLongSamples(t1);

   for (const auto interval : Intervals())
      interval->SetSilence(start, end);
}
```

But taking this step now involves taking another step later to remove the `ForEachClip`. I still see the suggestion I made [here](https://github.com/audacity/audacity/pull/5069#issuecomment-1708078967) as the most direct path. I haven't heard from you so far why it wasn't valid.

Paul-Licameli:
> I overviewed the fact that `ForEachClip` is private. Hence a solution would be
> 
> ```
> void WaveTrack::Interval::SetSilence(sampleCount start, sampleCount end)
> {
>    ForEachClip([&](auto& clip) {
>       const auto clipStart = clip.GetPlayStartSample();
>       const auto clipEnd = clip.GetPlayEndSample();
>       if (clipEnd > start && clipStart < end)
>       {
>          auto offset = std::max(start - clipStart, sampleCount(0));
>          // Clip sample region and Get/Put sample region overlap
>          auto length = std::min(end, clipEnd) - (clipStart + offset);
>          clip.SetSilence(offset, length);
>       }
>    });
> }
> ```
> 
> which, indeed, would give the `WaveTrack` code its final, post-refactoring look:
> 
> ```
> void WaveTrack::Silence(double t0, double t1)
> {
>    assert(IsLeader());
>    if (t1 < t0)
>       THROW_INCONSISTENCY_EXCEPTION;
> 
>    auto start = TimeToLongSamples(t0);
>    auto end = TimeToLongSamples(t1);
> 
>    for (const auto interval : Intervals())
>       interval->SetSilence(start, end);
> }
> ```
> 
> But taking this step now involves taking another step later to remove the `ForEachClip`. I still see the suggestion I made [here](https://github.com/audacity/audacity/pull/5069#issuecomment-1708078967) as the most direct path. I haven't heard from you so far why it wasn't valid.

I want to eliminate as many `TrackList::Channels` calls as possible now.  Please proceed as I have asked.

Once again.  Just do this boring bit of work now.

The reward is, we get the rewrites of many `WaveTrack` member functions DONE.  And we identify extensions necessary in `WaveTrack::Interval`.  Discovery of the right interfaces, proceeding top down this way, is the progress I want.

Yes, later there will be reimplementation of the lower level of detail of `Interval` and maybe merging with `WaveClip`.  The time for this is not now.  The time to do it is after we have identified the complete, sufficient interface it needs and have rewritten all the loops we can as interval-major.  The work, when the time comes, may be repetitious will mostly not be difficult.

saintmatthieu:
> The time to do it is after we have identified the complete, sufficient interface it needs and have rewritten all the loops we can as interval-major.

While I still don't see this step as a blocker to getting joint-stereo, I can conceive that interval-specific implementation is better located in `WaveTrack::Interval`. I'll leave aside my skepticism and will resume this work.

Paul-Licameli:
Writing my own instead

Issue #5080, opened 2023-08-24
Attempt record on an existing track causes Audacity to crash


dozzzzer:
### Steps to reproduce

1. have a mono / stereo clip
2. set recording channels to 1 or 2 channels respectively depending on the clip in step 1
3. Transport > Record

Update from @Paul-Licameli : also simply recording, then recording again to the end of the same track (R key not Shift+R)

### Expected behavior

Recording initiates with no crash

### Actual behavior

Crash

### Audacity Version

current master

### Operating system

All

### Additional context

Sentry log:
https://sentry.audacityteam.org/share/issue/0e72537783cd446eaa4a64b2f83ce2b8/

petersampsonaudacity:
I can reproduce this both on W10 and macOS 12.6.8 Monterey with latest 3.4.0 alpha

dozzzzer:
The PR #5082 fixes this issue
However, there's still a problem when recording in the Waveform view initially and then resuming recording in the Spectrogram view; it is logged separately as https://github.com/audacity/audacity/issues/5094

PR #5082, opened 2023-08-24
Fix crash in punch and roll recording

Paul-Licameli:
Resolves: #5080

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
@@ -988,7 +988,7 @@ void WaveformView::DoDraw(TrackPanelDrawingContext &context, size_t channel,
    channel = (track.IsLeader() ? 0 : 1);
 
    // TODO wide wave tracks -- remove this workaround
-   auto pLeader = *track.GetOwner()->Find(&track);
+   auto pLeader = *track.GetHolder()->Find(&track);
    assert(pLeader->IsLeader());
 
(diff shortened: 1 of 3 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -988,7 +988,7 @@ void WaveformView::DoDraw(TrackPanelDrawingContext &context, size_t channel,
    channel = (track.IsLeader() ? 0 : 1);
 
    // TODO wide wave tracks -- remove this workaround
-   auto pLeader = *track.GetOwner()->Find(&track);
```
saintmatthieu, line 991: Is this the line that was crashing? To my wobbly understanding returns `GetOwner()` and `GetHolder()` a pointer to the same object, only the former a the lock of a weak pointer and the latter, a raw pointer. But that wouldn't explain why this change would fix the crash.

Paul-Licameli: There is some complication during recording.  RegisterPendungChangedTrack sets up one or two tracks for which owner and holder differ.

saintmatthieu:
Trying to approve (where is that button?)

saintmatthieu (approved):

dozzzzer:
The PR partially fixes the issue #5080 
However, there's still a problem when recording in the Waveform view initially and then resuming recording in the Spectrogram view; it is logged separately as #5094

Issue #5079, opened 2023-08-24
Default quality  setting for M4A/AAC stereo export is ridiculously low at 98 kbps


petersampsonaudacity:
### Bug description
The default setting for M4A/AAC stereo export is ridiculously low at 98 kbps, this is a very poor setting for music and a not very good setting for voice.  _(BTW it was 100 kbps when I last updated the image for AAC export in the Manual)_

This is a regression as in 3.3.3 and earlier it was 160 kbps - introduced in @vsverchinsky 's work on new Export for 3.4.0

256 kbps would be a much better default as that is the quality the Apple sells its AAC files at in iTunes/Music- and most folk who are exporting to AAC are likely to be exporting music for use in an iTunes/Music library.

### Steps to reproduce
1. Clear audacity settings folder
2. launch 3.4.0 alpha
3. get some stereo audio
4. File > export Audio
5. select AAC/M4A
6. Observe offered (default) quality is 98 kbps 

### Expected behavior
256 kbps would be a much better default setting 

This is the quality at which Apple now sells music files in iTunes/Music.

### Actual behavior
![image](https://github.com/audacity/audacity/assets/50205984/08e57b7d-55b0-451c-a6b0-0842334e3d9e)

### Audacity Version
3.4.0 alpha (with new Export)

### Operating system
Windows and macOS 12.6.8 Monterey - so assume all OS

### Additional context
See also:   
**M4A(AAC) Export does not honor the user's quality settings for bit rates below 196kbps in the export dialog #4309** 

which indicates that with this default setting the user is likely to be getting 196/198 kbps as a minimum quality that we get from the AAC exporter we currently use.

And see also:
 **Native AAC support: Windows #3861 
 Native AAC support: macOS #3862**

petersampsonaudacity:
But if you use the default 160 0 you actually get 197-198

See my post from yesterday when I retested:
 **M4A(AAC) Export does not honor the user's quality settings for bit rates below 196kbps in the export dialog** #4309

I have re-run the tests today on W10 and macOS 12.6.8 Monterey with the latest 3.4.0 alpha,

The results have improved somewhat for Windows meaning that both give reasonably close results with stereo for setting of 196kbps and above:

### Actual behavior with stereo exports
**using the latest FFmpeg that @crsib recommended_**

**Windows 10**
Set rate  /  actual rate   - stereo
98 - 197  -  lowest setting
128 - 197
160 - 197
192 - 197
196 - 197
198 - 199
200 - 201
[...]

PR #5085, opened 2023-08-25
set default AAC bitrate to 256

LWinterberg:
Resolves: https://github.com/audacity/audacity/issues/5079


petersampsonaudacity:
Testing om W10 with @LWinterberg 's branch:  audacity-win-3.4.0-alpha-20230825+3264d5b-x64

This now looks fine to me Leo - many thanks for that.
image

PR #5091, opened 2023-08-25
Adjust border handle rewrites

vsverchinsky:
Depends on:
 - #5031
 
 Eliminates `TrackList::Channels` is in adjust border handle, simplifies `TrackPanel` drawing code, typos fix

Recommended:

QA: 
  Validate that there are no regressions when trim or stretch applied to mono and **aligned** stereo clips :
  - [x] Clip boundary change
  - [x] Snap points detected
  - [x] Boundary movement limits are detected correctly (no true-zero-clips, no clip overlapping)


libraries/lib-wave-track/WaveTrack.h
```diff
@@ -854,6 +854,13 @@ class WAVE_TRACK_API WaveTrack final
       const std::shared_ptr<WaveClip> mpClip1;
    };
 
+
+   std::shared_ptr<const Interval>
+   GetNextInterval(const Interval& interval, PlaybackDirection searchDirection) const;
```
Paul-Licameli, line 859: Good, just comment that it uses start times only

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -321,101 +321,6 @@ class AdjustClipBorder final : public WaveClipAdjustBorderHandle::AdjustPolicy
...
-      }
-   }
-
-   void Cancel() override
-   {
-      TrimTo(mInitialBorderPosition);
-   }
-};
-
-
```
Paul-Licameli, line 418: But was there an intention to use this?  Was the design not completely realized?

Was the intention to drag the line between touching clips, and shrink one and grow the other together?

vsverchinsky: Your guess is correct. Eventually, design team decided that we don't need that feature.

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -89,42 +76,42 @@ namespace {
...
-                std::min(
-                   clip.GetSequenceEndTime(), nextClip->GetPlayStartTime()) :
-                clip.GetSequenceEndTime();
+      return nextInterval ? nextInterval->Start() :
+                            std::numeric_limits<double>::max();
+
+   if(nextInterval)
+      return std::min(interval.GetClip(0)->GetSequenceEndTime(),
+                      nextInterval->Start());
+   return interval.GetClip(0)->GetSequenceEndTime();
```
Paul-Licameli, line 114: Later `GetClip` will not take an argument because clips will really be wide in stereo tracks.

New calls of that function with constant 0 argument are acceptable now.  I don’t want more calls with nonzero.

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -135,29 +122,30 @@ class AdjustClipBorder final : public WaveClipAdjustBorderHandle::AdjustPolicy
...
 
-   void TrimTo(double t) const
+   void TrimTo(double t)
    {
-      t = std::clamp(t, mRange.first, mRange.second);
-      for(auto& clip : mClips)
-         mAdjustHandler(*clip, t);
+      mBorderPosition = std::clamp(t, mRange.first, mRange.second);
+      for(unsigned channel = 0; channel < mInterval->NChannels(); ++channel)
+         mAdjustHandler(*mInterval->GetClip(channel), mBorderPosition);
```
Paul-Licameli, line 140: Can you rewrite this as an interval Manor, channel minor iteration, adding member functions to WaveChannelnterval, and eliminating the use of GetClip ?

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -182,14 +170,10 @@ class AdjustClipBorder final : public WaveClipAdjustBorderHandle::AdjustPolicy
                (currentTrack->GetLinkType() == Track::LinkType::Aligned && *trackList->Find(track) == currentTrack);
             for(const auto& interval : track->Intervals())
             {
-               if(isSameTrack)
+               if(isSameTrack && dynamic_cast<const WaveTrack::Interval*>(interval.get()) == &adjustedInterval)
```
Paul-Licameli, line 173: You can eliminate this ugly `dynamic_cast`, if above in the outer loop, you use `.Any<const WaveTrack>()`

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -386,85 +359,86 @@ UIHandlePtr WaveClipAdjustBorderHandle::HitAnywhere(
...
-    for (auto& clip : waveTrack->GetClips())
-    {
-        if (!WaveChannelView::ClipDetailsVisible(*clip, zoomInfo, rect))
-           continue;
-
-        auto clipRect = ClipParameters::GetClipRect(*clip.get(), zoomInfo, rect);
+   std::shared_ptr<WaveTrack::Interval> leftInterval;
+   std::shared_ptr<WaveTrack::Interval> rightInterval;
 
-        //double the hit testing area in case if clip are close to each other
```
Paul-Licameli, line 503: Maybe the comment should be kept to explain `* 2`

libraries/lib-track/Track.h
```diff
@@ -254,8 +254,6 @@ class TRACK_API Track /* not final */
    inline TrackList* GetHolder() const;
 
    LinkType GetLinkType() const noexcept;
-   //! Returns true if the leader track has link type LinkType::Aligned
-   bool IsAlignedWithLeader() const;
```
Paul-Licameli, line 258: A very pleasing deletion

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -833,9 +833,9 @@ class WAVE_TRACK_API WaveTrack final
       ~Interval() override;
 
       auto GetChannel(size_t iChannel) { return
-         WideChannelGroupInterval::GetChannel<WaveChannel>(iChannel); }
+         WideChannelGroupInterval::GetChannel<WaveChannelInterval>(iChannel); }
       auto GetChannel(size_t iChannel) const { return
-         WideChannelGroupInterval::GetChannel<const WaveChannel>(iChannel); }
+         WideChannelGroupInterval::GetChannel<const WaveChannelInterval>(iChannel); }
```
Paul-Licameli, line 838: Thank you for catching this

libraries/lib-channel/Channel.h
```diff
@@ -87,7 +87,8 @@ class CHANNEL_API WideChannelGroupInterval : public ChannelGroupInterval {
        result: `!(iChannel < NChannels()) || result`
     */
    template<typename IntervalType = ChannelInterval>
-   std::shared_ptr<IntervalType> GetChannel(size_t iChannel)
+   std::enable_if_t<std::is_base_of_v<ChannelInterval, IntervalType>, std::shared_ptr<IntervalType>>
+   GetChannel(size_t iChannel)
```
Paul-Licameli, line 91: Maybe this change isn't needed.  It rules out cases where `static_cast` below would not work but `dynamic_cast` would work.

Note too that there is a `const` overload that you did not similarly change.

Paul-Licameli (changes requested):
All the transformations look correct to me.  But eliminating the call to `GetClip` where the argument is not constant 0 is needed, to complete the intent of of my checklist item.

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
vsverchinsky: Not sure that I understand that correctly as there are no iterations over intervals in this function, and there is no need to iterate over intervals because we already have a pointer to the interval.
Having an interface that allows trimming of intervals in individual channels would be wrong thing to do, `WaveTrack::Interval` is more appropriate place for this, right?

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
vsverchinsky: That will change the behavior as we are searching for all interval boundaries, not just in wave tracks. But it could be simplified in other way - check for track and interval equality shortens to simple track pointer comparison, as we now always iterate over leader tracks and misalligned clips is not a valid case any more

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
vsverchinsky: I think it's a leftovers I forgot to remove when we disabled adjacent clip border adjustment, added a new commit that simplifies git testing

libraries/lib-channel/Channel.h
vsverchinsky: `const` version resolves to non-const, which could be SFINAE'd, but I removed this change from PR completely

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
Paul-Licameli: Right, there is only one level of looping.

What I do mean is, hide the use of `GetClip` and making the loop internal to a new member function of `Interval` would be good.

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -177,19 +164,12 @@ class AdjustClipBorder final : public WaveClipAdjustBorderHandle::AdjustPolicy
...
+               if(track == currentTrack)
                {
-                  auto waveTrackIntervalData =
-                     std::dynamic_pointer_cast<const WaveTrack::Interval>(
-                        interval);
-                  if(waveTrackIntervalData->GetClip(0).get() == adjustedClip)
-                  //exclude boundaries of the adjusted clip
-                     continue;
+                  //skip track that interval belongs to
+                  continue;
```
Paul-Licameli, line 172: Should this test be lifted out of this inner loop?

This behavior change ignores all intervals of the current track, unlike the previous version which ignored only one interval, but this makes sense:  other clips would limit the movement of the selected clip and their boundaries would not be useful snap positions.

Paul-Licameli (approved):

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -363,77 +363,39 @@ UIHandlePtr WaveClipAdjustBorderHandle::HitAnywhere(
...
+   {
+      const auto& clip = *interval->GetClip(0);
+      if(!WaveChannelView::ClipDetailsVisible(clip, zoomInfo, rect))
+         continue;
+
+      auto clipRect = ClipParameters::GetClipRect(clip, zoomInfo, rect);
+      if(px >= clipRect.GetLeft() && px - clipRect.GetLeft() <= BoundaryThreshold)
+         return makeUIHandle(interval, true);
+      if(px <= clipRect.GetRight() && clipRect.GetRight() - px <= BoundaryThreshold)
+         return makeUIHandle(interval, false);
```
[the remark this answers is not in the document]
vsverchinsky: You are right, I didn't consider such case possible

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
vsverchinsky: > Should this test be lifted out of this inner loop?

Yes, thanks!

vsverchinsky:
Rebased onto master, last commit that changes handle hit test area was removed

chinakov:
Tested, no issues found. Thank you!

Issue #5018, opened 2023-08-09
"Beep on completion of longer activities" is non-functional


SteveDaulton:
### Bug description

The "Beep on completion of longer activities" option is non-functional.

### Steps to reproduce

1. Enable the "Beep on completion of longer activities" option in Preferences > Interface
2. Perform an action that takes more than a minute. I used "Paulstretch" on a long track.

### Expected behavior

Audacity to play a beep when activities lasting longer than a minute are completed.

### Actual behavior

No beep.

### Audacity Version

current master / alpha / nightly build

### Operating system

All

### Additional context

https://manual.audacityteam.org/man/interface_preferences.html

daksh-code:
can I work on this?

PR #5100, opened 2023-08-28
Mac backwards compatibility & usability tweaks

RJVB:
- make src/AudacityApp.mm build on pre-10.14 systems which either don't have `@available()` or don't have the 10.14 SDK installed. (`@available` is a runtime operator.)
- make `ProgressDialog::Beep()` beep instead of printing an invisible failure message.
Use `NSBeep()` instead of calling a non-implemented wxWidgets method. (may fix https://github.com/audacity/audacity/issues/5018)
- Wrapper.c : change the process name (as it appears in `ps`) after forking, exit with an error if `execve()` fails and provide an env.var switch to preserve `DYLD_LIBRARY_PATH` for advanced users.
Contrary to comments in the code, setting `DYLD_LIBRARY_PATH` may also be what the user wants, and setting it doesn't necessarily interfere with loading the FFmpeg libraries just as it doesn't do that on Linux.

Committed from host : Portia.local

Recommended:

### DIFF
--- a/libraries/lib-wx-init/ProgressDialog.cpp
+++ b/libraries/lib-wx-init/ProgressDialog.cpp
@@ -52,6 +52,10 @@
 
 #include "MemoryX.h"
 
+#ifdef __WXOSX_COCOA__
+extern "C" void NSBeep(void);
+#endif
+
 // This really should be a Preferences setting
 static const unsigned char beep[] =
 {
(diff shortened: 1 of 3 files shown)


RJVB:
FWIW, I'd also like to propose to change the "pointing finger/hand" SelectHandle cursor by something where the hot point is a little more clear, like `wxCursorCross`.

That's a usability tweak too in my book but I wasn't certain it'd be a good idea to include it without bringing it up first - it's a user-visible change.

LWinterberg:
https://github.com/audacity/audacity/issues/5018 may be the relevant bug to item 2

RJVB:
> #5018 may be the relevant bug to item 2

Sounds like it and if `ProgressDialog::Beep()` is what is/was called after long operations then indeed the feature hasn't been working for a long time on Mac.

crsib (approved):

dozzzzer:
Tested:
- #5108 is fixed now
- FFmpeg is discovered
- Import/Export audio also works as expected

Issue #5102, opened 2023-08-29
Crash when joining an audio track and an empty audio track


SteveDaulton:
### Bug description

"Join tracks to stereo" crashes Audacity when upper track is empty

### Steps to reproduce

1. Add a new mono track
2. Add another mono track
3. Generate some audio into the second track
4. From the upper track dropdown menu, select "Make Stereo Track"

### Expected behavior

A stereo  track with audio in the right (lower) channel.

### Actual behavior

Crash

### Audacity Version

current master / alpha / nightly build

### Operating system

All OSs

### Additional context

Tested with [audacity-linux-3.4.0-alpha-20230828+a53fd35-x64-20.04](https://nightly.link/audacity/audacity/workflows/build/master/audacity-linux-3.4.0-alpha-20230828%2Ba53fd35-x64-20.04) on Xubuntu 22.04.

On restarting Audacity, attempting to recover the project will crash immediately. Because of this it is an **_extremely serious bug_** because if it happened in an existing project it would permanently destroy the project with no means of recovery.

Paul-Licameli:
The feature of joining of mono to stereo is in a state of flux while I'm doing a massive transformation of how stereo data is managed.

I'll figure out how not to have a crash (probably not hard) but then don't expect the resulting track always to behave as you expect in alpha.

Having tracks with clips that don't align is a feature, or misfeature, that Audacity 3.4 will retract.  I'm rewriting many other places first to use the assumption that misaligned tracks don't exist, so different right hand channel data is ignored.  Only later, I will rewrite the piece that joins mono tracks into stereo, and it will correct clip misalignment in the result.

Paul-Licameli:
I have opened a PR to fix the crash.  If you repeat the steps, the resulting stereo track will not draw the clip in the right hand channel.

Please do not include this misbehavior in the scope of this issue.  I will only fix the crash.  As I said, do not otherwise expect reasonable behavior yet when joining mono tracks with misaligned stereo clips.  That is a large task I'm planning for later in developing this release.

petersampsonaudacity:
Worse still it crashes when you attempt the offered Recovery on relaunch

Paul-Licameli:
> Worse still it crashes when you attempt the offered Recovery on relaunch

Are you referring to alpha or the branch with my proposed fix?

petersampsonaudacity:
> Are you referring to alpha or the branch with my proposed fix?

alpha master

Paul-Licameli:
[...]

Issue #4681, opened 2023-05-09
Restructuring for wide stereo clips (checklist)


Paul-Licameli:
- [x] Library reorganizations to support stretching
  - [x] #4754 (Remove lib-effects -> lib-audio-io)
  - [x] #4759 (Remove lib-audio-graph->lib-track)
  - [x] #4746 (Add lib-mixer, not using lib-track)
  - [x] #4795 (Remove lib-audio-graph->lib-track)
  - [x] #5130 (This PR is also in the next section)
- [x] De-duplication of WaveTrack data
   - [x] #2804
   - [x] #4737
   - [x] #5027
   - [x] #5101
   - [x] #5116
   - [x] #5121
   - [x] #5127 
   - [x] #5130
- [x] Rewrite outer loops over tracks (introducing new inner loops over channels); visit leaders only
  - [x] #4738
  - [x] #4886
  - [x] #4908
  - [x] #4937
  - [x] #4945
  - [x] #4961
  - [x] #4974
- [x] Rewrite inner loops over channels (remove `TrackList::Channels` outside `Track` and` WaveTrack`; hide STL track iterators)
  - [x] #4825
  - [x] #4844
  - [x] #4987
  - [x] #5008
  - [x] #5031
  - [x] #5131
  - [x] #5128
  - [x] #5091 
  - [x] #5159
  - [x] #5212
  - [x] #5180 (Includes "Wide" RecordableSequences)
  - [x] #5530
- [x] Change channel-major iterations to interval-major (except in effects, display, recording)
  - [x] #4876
  - [x] #5065 
  - [x] #5198
- [x] Re-implement representation of channels
   - [x] #6086
   - [x] #4742
   - [x] #5156
   - [x] #6081
   - [x] #4954
   - [x] #6070
   - [x] #6072
   - [x] #5460
- [x] Implement joining of channels according to design
   - [x] #5200
- [x] File compatibility break
   - [x] #5201
   - [x] #5202
   - [x] #5211
- Incidental fixes
  - #4725
  - #4730
  - #4714 
  - #4718
  - #4713 
  - #4716

Paul-Licameli:
Force-pushed once more with one commit inserted first in the sequence to fix crashes in sync-lock

Paul-Licameli:
The fourth and fifth main headings of the outline are checked off after #5965 is merged

PR #5104, opened 2023-08-29
Fix crash in drawing when right channel has more clips than left

Paul-Licameli:
Resolves: #5102

Recommended:

### DIFF
--- a/libraries/lib-channel/Channel.h
+++ b/libraries/lib-channel/Channel.h
@@ -234,13 +234,13 @@ public:
          const Channel, Channel>;
    public:
       IntervalIterator() = default;
-      IntervalIterator(ChannelType *pChannel, size_t index)
-         : mpChannel{ pChannel }, mIndex{ index }
+      IntervalIterator(ChannelType *pChannel, size_t index, size_t nIntervals)
+         : mpChannel{ pChannel }, mIndex{ index }, mNIntervals{ nIntervals }
       {}
 
       std::shared_ptr<IntervalType> operator *() const
       {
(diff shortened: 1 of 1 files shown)


libraries/lib-channel/Channel.h
```diff
@@ -261,24 +261,19 @@ class CHANNEL_API Channel
    private:
       ChannelType *mpChannel{};
```
saintmatthieu, line 262: Couldn't these be const actually ?

saintmatthieu (commented):
I understand this commit as a minimal, "preserve-from-crashing" commit that circumvents the fact that stereo tracks still can be made of inconsistent left/right clip counts, by always returning the count of the leader.
It seems to fix the situation where the right channel has most clips, but crash persists if left has most clips. I suppose this is because drawing thinks the right channel then has more clips than it actually has but still tries to draw all.
So might a better workaround be reporting the least clip count across channels ?
@Paul-Licameli you mentioned recently that disallowing joining tracks with inconsistent clip boundaries was still for relatively far away within the current refactoring effort ? Having this earlier than planned, even in a quick-and-dirty way, might still be less dirty than the situation we're trying to deal with now.

vsverchinsky (commented):
Fix doesn't seem to be complete:
изображение

STR:
1) Create mono track and single clip to it
2) Duplicate track
3) Make a copy clip in the second track
4) Call "Make Stereo Track" on the first track

Paul-Licameli:
This crash prevention isn't worth it now.

Let it remain a known issue until the right time for it in this plan #4681

Issue #5108, opened 2023-08-29
Deadlock/crash when recording in Spectrogram view


dozzzzer:
### Bug description

When in Spectrogram view, a few iterations of these two actions – click Record and after a few seconds click Stop – would cause Audacity to freeze.

### Steps to reproduce

1. Add a mono track
2. Set Recording Channels to "1"
3. Click Record button
4. After a few seconds of recording, click Stop button
5. Click Record button again
6. Repeat steps 4-5 a few times

### Expected behavior

No crash, recording continues.

### Actual behavior

Audacity freezes:

https://github.com/audacity/audacity/assets/70063970/6e8b606a-658d-4bf0-b5ca-e299dc77153d

### Audacity Version

other

### Operating system

Windows, macOS, Linux

petersampsonaudacity:
For me at STR step 6 I get a crash after a few stop-start recordings when in Spectrogram view/

I can't seem to make the same happen with Waveform view

petersampsonaudacity:
@dozzzzer 

Testing on W10 with @Paul-Licameli 's branch build for the fix for this:  audacity-win-3.4.0-alpha-20230829+22fb39b-x64

I can no longer replicate this, so it looks to be fixed in Paul's branch build.

I tried 25 repeats of stop-start record in Spectrogram view - formerly somewhere in the rage of 3-7 would cause the crash.

dozzzzer:
Tested

PR #5110, opened 2023-08-29
More complete fix for intermittent crash recording to spectrum view.

Paul-Licameli:
Resolves: #5108

Recommended:


petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's branch build for the fix for this:

I can no longer replicate this, so it looks to be fixed in Paul's branch build.

I tried 25 repeats of stop-start record in Spectrogram view - formerly somewhere in the rage of 3-7 would cause the crash.

saintmatthieu (approved):

Issue #5116, opened 2023-08-31
De-duplication of channel data: color and origin


vsverchinsky:
More elimination of duplicated track members that would need to be kept consistent across channels.

QA:
- [x] Shift a stereo track in time, delete the clip, then record more into the track
- [x] Changing WaveTrack color using the context menu; then recording a new clip which gets correct color
- [x] Persistency of wave track color choice in mono and stereo tracks (test it by making a new clip after reopening)
- [x] Mix-and-render (to new or to replacement track) keeps the color choice (again, make a new clip)
- [x] Swapping channels and splitting stereo preserve the color choice

chinakov:
Tested, thank you!

PR #5120, opened 2023-08-31
De-deplication of channel data

vsverchinsky:
Resolves: #5116 

Recommended:


libraries/lib-track/Track.cpp
```diff
@@ -1006,7 +1006,7 @@ TrackList::RegisterPendingChangedTrack(Updater updater, Track *src)
    std::vector<Track*> result;
    if (src) {
       tracks = src->Clone(); // not duplicate
-      assert(NChannels() == tracks->NChannels());
+      assert(src->NChannels() == tracks->NChannels());
```
Paul-Licameli, line 1009: The same fix is in the #5041 branch too

Issue #5127, opened 2023-09-01
When swapping (or splitting) channels of a stereo track, the effects should not be removed from the (second) stack


dozzzzer:
### Steps to reproduce

1. Have a stereo track
2. Add some effect(s) to the stack
3. Swap channels OR split the stereo track

### Expected behavior

The effects remain in the stack, or in both stacks.

### Actual behavior

The effects get removed from the stack (swapping) or from the second track's stack (splitting). When swapping channels
one more time, they re-appear.

### Audacity Version

current master / alpha / nightly build

### Operating system

Other

### Additional context

A regression against 3.3.3

dozzzzer:
Bisects to https://github.com/audacity/audacity/commit/e5bc5d5818eef068b064d3a1f240f251834e26f0

Paul-Licameli:
Bisects more exactly to 640275ef394617e6802131c3840250c41d862134 inside that PR

petersampsonaudacity:
@dozzer @Paul-Licameli 

Note too that if you have two mono tracks and only the lower one has effects in its effects stack 
if you make a stereo track from those two tracks the the resultant stereo track has no effects in the effects stack (and playing the audio confirms that the right channel plays without added effects).

BUT if then then split the stereo track you observe (and hear) the the lower track (right channel) still has its effects in its effects stack.

petersampsonaudacity:
I also confirm the Swap Stereo Channels behavior that @dozzzzer reports.

Bit also note that if you swap the channels back again, then the effects stack is restored.

dozzzzer:
The bug is still reproducible in PR #5128:

https://github.com/audacity/audacity/assets/70063970/6743b292-a812-403f-adb9-540741251d8c

dozzzzer:
Fixed in PR #5130

petersampsonaudacity:
Testing on W10 latest alpha master:  audacity-win-3.4.0-alpha-20230907+5f43627-x64-msvc2022

This issue as logged tests OK on the latest 3.4.0 master.

But there is an associated residual when making a stereo track from tracks with different effects stacks - I will log this separately

PR #5128, opened 2023-09-01
Rewrite cutline editing

Paul-Licameli:
Rewrite click to expand or delete cutlines or to merge clips, and rewrite
the painting of clip boundaries and cutlines, to remove some calls to
TrackList::Channels and lower two others into WaveTrack.cpp.

QA:  Do all this for mono and stereo tracks, and undo and redo changes:
- [x] Observe no change of display of clip boundaries and splits and cutlines in waveform and spectrum
- [x] Click clip split to merge still works (although we do intend to hide this feature later)
- [x] Left-click on a cutline to expand it
- [x] Right-click on a cutline to delete it

- [x] I signed [CLA](https://www.audacityteam.org/cla/)
- [x] The title of the pull request describes an issue it addresses
- [x] If changes are extensive, then there is a sequence of easily reviewable commits
- [x] Each commit's message describes its purpose and effects
- [x] There are no behavior changes unnecessary for the stated purpose of the PR

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes of behavior

PR #5130, opened 2023-09-03
Deduplicate effect stack

Paul-Licameli:
Resolves: #5127

Depends on
- #5128

Store the real-time effect stack in `ChannelGroup` attachments, just as
we do for name, rate, solo, etc.

`WideSampleSequence` no longer needs to be another `ClientData::Site`.

Some changes result in library dependencies, and `WaveTrack` now has one
fewer member function to explain.

Recommended:


Paul-Licameli:
Rebased onto master; no longer dependent on #5128.

This PR is meant to fix issue #5127 .  PR #5128 is no longer claimed to fix it.

Paul-Licameli:
Sorry, this DOES depend on #5128 after all!

Paul-Licameli:
Rebased unchanged onto updated #5128

crsib (approved):
I feel that `FindChannelGroup` is a bit confusing, but I can't think of a better name anyway.

Paul-Licameli:
> I feel that `FindChannelGroup` is a bit confusing, but I can't think of a better name anyway.

Elsewhere I have use the convention that “Find” functions can return null pointers (as stl containers may give you end iterators) but “Get” functions can’t fail and return references.

Any better idea?

dozzzzer:
Tested:
- #5127 fix
- Realtime effects stickiness, persistency
- Rendering with realtime effects
- Playback with RT effects
- Undo/Redo for RT effects

Issue #5121, opened 2023-08-31
Reassign Clip name, color, start time without TrackList::Channels


Paul-Licameli:
Should depend on
- #5031

See also issue #5076 

Subtasks
- `WaveTrack::Interval` has member functions to set name, color, and start time in one or two `WaveClip`s
- `SetClipCommand` rewritten as a `SetTrackBase` using interval iteration
- Remove class `SetChannelsBase`
- `EffectRepeat` sets clip names using interval iteration
- `ProjectFileManager::AddImportedTracks` can loop over leader tracks only, and intervals, to set clip names

QA:
- [x] "Set Clip" macro command to change clip names, color, start time in mono and stereo
- [x] Use Repeat effect to replicate a clip in a track; observe clip names; swap channels or split the clip; undo/redo
- [x] Import tracks, observe clip names, swap, split, undo/redo
- [x] A clip (mono or stereo) can be dragged into a track of another color and keeps its different color; swap, split, undo/redo

dozzzzer:
Tested

Paul-Licameli:
It's done, but didn't auto-close because of a bit of confusion in the merging of PRs.

Issue #5076, opened 2023-08-23
Split stero track gets old clip name assigned to the clip in the bottom track


petersampsonaudacity:
### Bug description

If you rename a clip in a stereo track and then split the stereo track to mono or stereo then the lower clip wrongly reverts to the original clip name.

_Further testing indicates that the bottom clip is **NOT** taking its name from the Track name, as I suspected might be the case._

### Steps to reproduce

1. import a stereo audio file
2. Observe clip takes its name from the audio file name
3. change the name of the clip
4. Split the stereo track _(to either mono or stereo)_
5. Observe: upper clip retains the new name assigned at step 3 - expected behavior
6. Observe: : lower clip has the original clip name from the import - steps 1-2  -  wrong behavior

### Expected behavior

the new name should be propgated to the clips in both tracks
![image](https://github.com/audacity/audacity/assets/50205984/eeef9b31-83fa-43bc-9dbf-c045c4140380)

### Actual behavior

![01](https://github.com/audacity/audacity/assets/50205984/b8f455e7-339e-451e-ab1a-55db7f4c446f)
![02](https://github.com/audacity/audacity/assets/50205984/80a8bfc9-2bf5-4eea-a4d9-643ee0f56f07)
![03](https://github.com/audacity/audacity/assets/50205984/e1b8736a-18e9-40b7-8831-68e4ac4f853e)

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

None

petersampsonaudacity:
This may be related to:

 **After merging two mono tracks in one stereo track and swapping channels, the name of the track is changing too** #2862

petersampsonaudacity:
See also:   **Clip renaming behavior of "Crossfade Clips" Nyquist effec**t #4229

chinakov:
Tested, issue has been fixed

PR #5133, opened 2023-09-04
Wave track interval extend

vsverchinsky:
Resolves: #5121, #5076

Depends on:
 - #5031

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -185,6 +185,54 @@ WaveTrack::Interval::Interval(const ChannelGroup &group,
 
 WaveTrack::Interval::~Interval() = default;
 
+void WaveTrack::Interval::SetName(const wxString& name)
+{
+   ForEachClip([&](auto& clip) { clip.SetName(name); });
+}
+
+const wxString& WaveTrack::Interval::GetName() const
+{
+   //TODO wide wave tracks:  assuming that all 'narrow' clips share common name
+   return mpClip->GetName();
+}
+
+void WaveTrack::Interval::SetColorIndex(int index)
+{
+   ForEachClip([&](auto& clip) { clip.SetColourIndex(index); });
+}
+
+int WaveTrack::Interval::GetColorIndex() const
+{
+   //TODO wide wave tracks:  assuming that all 'narrow' clips share common color index
+   return mpClip->GetColourIndex();
+}
+
+void WaveTrack::Interval::SetPlayStartTime(double time)
+{
+   ForEachClip([&](auto& clip) { clip.SetPlayStartTime(time); });
+}
+
+double WaveTrack::Interval::GetPlayStartTime() const
+{
+   //TODO wide wave tracks:  assuming that all 'narrow' clips share common beginning
+   return mpClip->GetPlayStartTime();
+}
+
+bool WaveTrack::Interval::IsPlaceholder() const
+{
+   return mpClip->GetIsPlaceholder();
+}
+
+void WaveTrack::Interval::ForEachClip(const std::function<void(WaveClip&)>& op)
+{
+   for(unsigned channel = 0,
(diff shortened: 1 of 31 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -185,6 +185,54 @@ WaveTrack::Interval::Interval(const ChannelGroup &group,
 
 WaveTrack::Interval::~Interval() = default;
 
+void WaveTrack::Interval::SetName(const wxString& name)
+{
+   ForEachClip([&](auto& clip) { clip.SetName(name); });
```
Paul-Licameli, line 190: This and other lambdas can use empty capture `[]` and that might generate slightly more efficient code, because then they convert to pointer-to-function, and the `std::function` implementation might optimize that case avoiding some heap allocation.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -185,6 +185,54 @@ WaveTrack::Interval::Interval(const ChannelGroup &group,
...
+   //TODO wide wave tracks:  assuming that all 'narrow' clips share common beginning
+   return mpClip->GetPlayStartTime();
+}
+
+bool WaveTrack::Interval::IsPlaceholder() const
+{
+   return mpClip->GetIsPlaceholder();
+}
+
+void WaveTrack::Interval::ForEachClip(const std::function<void(WaveClip&)>& op)
```
Paul-Licameli, line 226: @saintmatthieu I would like to see #5069 rebased onto this, and using this new function where possible.  The commits that would be rewritten with this function are the ones I think are important and should be merged.

src/commands/SetClipCommand.cpp
```diff
@@ -85,34 +85,34 @@ void SetClipCommand::PopulateOrExchange(ShuttleGui & S)
    S.EndMultiColumn();
 }
 
-bool SetClipCommand::ApplyInner( const CommandContext &, Track * t )
+bool SetClipCommand::Apply(const CommandContext& context)
 {
-   // if no 'At' is specified, then any clip in any selected track will be set.
-   t->TypeSwitch([&](WaveTrack &waveTrack) {
-      WaveClipPointers ptrs( waveTrack.SortedClipArray());
```
Paul-Licameli, line 92: Not sorting clips any more.  What might that change?  Nothing, I see.  No break was in the loop.  OK

src/effects/Repeat.cpp
```diff
@@ -141,25 +136,40 @@ bool EffectRepeat::Process(EffectInstance &, EffectSettings &)
...
+         auto sortedIntervals = std::vector(
+            track.Intervals().begin(),
+            track.Intervals().end()
+         );
+         auto sourceIntervals = std::vector(
+            firstTemp->Intervals().begin(),
+            firstTemp->Intervals().end()
+         );
+         std::sort(sortedIntervals.begin(), sortedIntervals.end(), compareIntervals);
+         std::sort(sourceIntervals.begin(), sourceIntervals.end(), compareIntervals);
```
Paul-Licameli, line 153: This getting of a sorted sequence of intervals looks like something that might be frequently needed and could become a utility function.  If in `lib-channel`, it might need a template type parameter for more convenience.

Do that another day...

Paul-Licameli (approved):

vsverchinsky:
Commits [ceabb64](https://github.com/audacity/audacity/pull/5133/commits/ceabb6469f59a14935ccf1111b57baeaf859acb6)...[9ab0ca7](https://github.com/audacity/audacity/pull/5133/commits/9ab0ca7ead376286b97acf79e9f4d57f2a60fdbe) merged as a part of #5031, [32df0b4](https://github.com/audacity/audacity/pull/5133/commits/32df0b44fb8329450b4d8fae2db17748d003ee83) extracted into #5165

samm9998 (commented):
#516

Issue #5113, opened 2023-08-30
Wrong display of time when punch and roll recording on an expanded clip


dozzzzer:
### Steps to reproduce

1. Generate a 10-second chirp
2. Expand it to 30 seconds
3. Place the cursor somewhere at the very end of the clip (see the screen recording)
4. Transport > Recording > Punch and roll record
5. Observe

### Expected behavior

The time display of the recorded clip has no stretch factor applied to it.

### Actual behavior

Wrong display of time on the recorded clip. 
The time of the recorded audio goes appr. 3 times faster; the playhead doesn't catch up and the gap become larger as the recording continues:

https://github.com/audacity/audacity/assets/70063970/44cec9c2-80fb-4ccd-bcf2-8085b025ccc1

### Audacity Version

[Latest master](https://github.com/audacity/audacity/actions/runs/6050706064)

### Operating system

Windows, macOS

saintmatthieu:
Strangely the bug wasn't always reproducible - here's the reason why:
In [this video](https://github.com/audacity/audacity/assets/22740106/732f8228-1d68-4bc2-bf79-9a248d3d0a0e),
1. I first place the cursor to the left sample and start the recording. In that case, a new clip is created.
2. Then, I place the cursor closer to the right sample, and no new clip is created.

This is the behaviour on current release, without stretching involved. I'm not sure what the default behaviour should be - probably appending if possible, as in 2.

With stretching, a new clip should be created if the existing clip is stretched.

PR #5134, opened 2023-09-04
Bugfix: Wrong display of time when punch and roll recording on an expanded clip

saintmatthieu:
Resolves: #5113

Fix punch-n-roll for stretched (and unstretched) clips - see bug report for more detail.

Recommended:

QA:
- [x] As in current release, a new clip is created when starting recording on a non-empty track.
- [x] Punch-n-Roll recording creates a new clip whether the cursor is in the left or right half of the space between samples ([demo](https://github.com/audacity/audacity/issues/5113#issuecomment-1705154108))
- [x] Punch-n-roll on a stretched clip doesn't have the issue reported in #5113

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -2220,6 +2220,17 @@ bool WaveTrack::CloseLock() noexcept
    return true;
 }
 
+const WaveClip* WaveTrack::GetRightmostClip() const {
+   if (mClips.empty())
+      return nullptr;
+   return std::max_element(
+             mClips.begin(), mClips.end(),
+             [](const auto& a, const auto b) {
+                return a->GetPlayEndTime() < b->GetPlayEndTime();
+             })
+      ->get();
+}
+
 ClipConstHolders WaveTrack::GetClipInterfaces() const
 {
   // We're constructing possibly wide clips here, and for this we need to have
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -37,6 +37,7 @@ Paul Licameli split from ProjectManager.cpp
 #include "TransportUtilities.h"
 #include "UndoManager.h"
 #include "ViewInfo.h"
+#include "WaveClip.h"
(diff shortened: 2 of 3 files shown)


crsib (approved):

src/ProjectAudioManager.cpp
```diff
@@ -862,7 +863,24 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
...
                // when append recording.
                const auto pending = static_cast<WaveTrack*>(newTrack);
-               if (endTime <= t0)
+               const auto lastClip = pending->GetRightmostClip();
+               // RoundedT0 to have a new clip created when punch-and-roll
+               // recording with the cursor in the second half of the space
+               // between two samples
+               // (https://github.com/audacity/audacity/issues/5113#issuecomment-1705154108)
+               const auto recordingStart =
+                  std::round(t0 * pending->GetRate()) / pending->GetRate();
```
Paul-Licameli, line 872: What if pending is null ptr?

src/ProjectAudioManager.cpp
```diff
@@ -862,7 +863,24 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
                // Less than or equal, not just less than, to ensure a clip boundary.
                // when append recording.
                const auto pending = static_cast<WaveTrack*>(newTrack);
-               if (endTime <= t0)
```
Paul-Licameli, line 865: Is endTime now unused?

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2220,6 +2220,17 @@ bool WaveTrack::CloseLock() noexcept
    return true;
 }
 
+const WaveClip* WaveTrack::GetRightmostClip() const {
+   if (mClips.empty())
+      return nullptr;
+   return std::max_element(
+             mClips.begin(), mClips.end(),
+             [](const auto& a, const auto b) {
```
saintmatthieu, line 2228: minor: `const auto& b`

dozzzzer:
Tested

Issue #5135, opened 2023-09-04
Change temp/pitch/speed effect + trimming or non-destructive tempo end of clip reveals spurious clip


DavidBailes:
### Steps to reproduce

1. Generate a chirp
2. Open change tempo effect and change speed by 10%
3. trim end of clip or change tempo by dragging end of clip

### Expected behavior

After step 3:
![image](https://github.com/audacity/audacity/assets/11887021/5eec1ff8-2253-4d9b-b646-a98de6fccfe2)

or
![image](https://github.com/audacity/audacity/assets/11887021/3339f1ed-f419-4370-b83d-b1893159b80b)

respectively

### Actual behavior

After step 3:
![image](https://github.com/audacity/audacity/assets/11887021/bf2f7fa4-effb-43ee-b537-a27f0f2a8d13)

or

![image](https://github.com/audacity/audacity/assets/11887021/59c40df6-fd0e-4ac9-818d-ad69fdb65679)

respectively

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

saintmatthieu:
Thanks for the report, @DavidBailes !
It looks like one of my currently being reviewed PRs fixes this, but I don't know how yet, so at least there'll be a need for investigation.

dozzzzer:
Fixed by #5138

Issue #5118, opened 2023-08-31
Incorrect result when mixing stretched stereo clips down to mono


dozzzzer:
### Bug description

When having multiple clips with different stretch ratios on one stereo track and mixing the track down to mono, the result is incorrect.

### Steps to reproduce

1. Have a stereo track
2. Have a few copies of the same stereo clip on that track
3. Have different stretches on different clips (see the image below as an example)
4. Tracks > Mix > Mix Stereo Down to Mono
5. Observe the result

<img width="500" alt="Screenshot 2023-08-31 at 12 25 07" src="https://github.com/audacity/audacity/assets/70063970/f1baf651-4838-4f14-9bef-0793157770b4">

### Expected behavior

Gaps are preserved, and the overall length of the resulting clip is the same as all the clips combined.
The resulting clip should look like this:

<img width="500" alt="Screenshot 2023-08-31 at 12 21 15" src="https://github.com/audacity/audacity/assets/70063970/c37621de-f022-4f70-9f0d-907b5d150e15">

And here's the pre-mixed-down-to-mono clip:
https://www.dropbox.com/scl/fi/wkejm0j4n85sydizlhpjw/before-mixing-down-to-mono.mp3?rlkey=q7ikc2u7dwb3mzmw9rulb2ksa&dl=0

### Actual behavior

The overall length is correct, but the gaps aren't preserved and the clip doesn't sound as when pre-mixed-down-to-mono

<img width="500" alt="Screenshot 2023-08-31 at 12 26 21" src="https://github.com/audacity/audacity/assets/70063970/8f0dcffe-5f75-4464-9e13-c3d2ffac5cad">

And here's the resulting mono clip: 
https://www.dropbox.com/scl/fi/wo0cqtrwrulvus3j7mivi/after-mixing-down-to-mono.mp3?rlkey=fyef8osiqqsed8f0md9osbsws&dl=0

### Audacity Version

[Other](https://github.com/audacity/audacity/actions/runs/6023850506)

### Operating system

All

saintmatthieu:
The fix for this is already in https://github.com/audacity/audacity/pull/5044/commits/af1d830b2b8015f2527df7a0bd086cb4238453f3 and figures as a checkbox there, actually. (That PR is awaiting first https://github.com/audacity/audacity/pull/5043 to be reviewed and tested.)

Paul-Licameli:
Too early to close the issue before the PR is tested and merged

saintmatthieu:
[...]

PR #5138, opened 2023-09-04
Destructive effects applied to stretched clips

Paul-Licameli:

## QA (copied from #5044)
All effects are supposed to work. Trying all of them may be too much, but a variety of types of effects should be tried (time-related effect, Nyquist, EQs and filters, noise removal, pitch shift, spectral tools, etc. Of course try them on stretched data, but also on stereo data, etc.

Note that applying an effect on a partial clip selection will create a new clip if the clip was stretched. That's wanted. The old behavior is only preserved on unstretched data.

- [x] AutoDuck
- [x] AutoDuck also works when there is stretch in the un-selected control track
- [x] Change Speed
- [x] Click Removal
- [x] EQ
- [x] Generate silence (in particular, does not create new clip boundaries at left and right edges of selection when generating over existing data and there is no stretching)
- [x] Loudness (in both modes)
- [x] Noise Reduction
- [x] Normalize
- [x] Paulstretch
- [x] Repair (Since this requires up to 128 samples only, the result will probably only be silence, and that's ok: it's a known limitation of the algorithm that rendering very tiny portions of audio only output silence)
- [x] Repeat
- [x] Reverse
- [x] Change Tempo and Sliding Scale, high and low quality
- [x] Stereo To Mono (previously reported by @dozzzzer as https://github.com/audacity/audacity/issues/5118)
- [x] Truncate Silence
- [x] Compressor
- [x] Some Nyquist effects and generators
- [x] At least one "Per Track Effect" example (any of the realtime capable, built-in or plug-in effects, but used destructively)

More for QA:
- [x] Effects applied to long enough areas, that the stretching progress indicator appears:  the indicator should fill up correctly, to 100% and then promptly vanish
- [x] Noise Reduction Analyze step, selecting stretched data and selections crossing clip boundaries with different stretch
- [x] Noise reduction analysis does not make unintended side-effects on tracks, or undo items
- [x] Test with and without the "Editing a clip can move other clips" preference (Track Behaviors)
- [x] Test all with selection of only a part of the track's duration, maybe with clip boundaries within, and different stretches
- [x] Test with silence generating a duration unequal to the selected duration
- [x] (Very minor) Click on clip boundary to merge isn't yet a hidden feature; it fails when stretch ratios differ, but makes no empty Undo item
- [x] Observe that new clip boundaries result when stretching is necessary, but if not, old behavior is preserved
- [x] Repair: stretch a mono clip by a factor of 2 and select 100 samples. You should get the error about not suitable to stretched data, not the other error that says you've selected too many samples.
- [x] #5135 is fixed by this PR
- [x] Apply a destructive effect to multiple tracks with stretch (maybe some without).  Observe just one progress bar proportioned to the number of wave tracks that have any stretch in the selection (maybe fewer than all the selected tracks).  The bar fills to completion and disappears at once, without jumping too fast over any part.
- [x] Use pencil tool _on unstretched clip(s)_ (getting this to work on stretched clips is for 3-of-6 to fix)
- [x] Effects that require sync-lock adjustment of other wave tracks that have some stretch or shrink in the selected interval
  - [x] In all of these, it should not insert splits in those tracks, and work when shortening a track, or widening a gap between clips; inserting silence into a clip is a KNOWN PROBLEM that is NOT IN THE SCOPE OF THIS PR but instead #5043 will do that
    - [x] Inserting generated silence
    - [x] Inserting generated tone, noise, chirp
    - [x] Nyquist generators, Change Speed, Paulstretch, Repeat, Change Tempo or Sliding Stretch (high and low quality)
  - [x] These two have some special case sync-lock handling and for now they WILL insert splits, though that might be reconsidered in future.
    - [x] Reverse
    - [x] Truncate Silence (Sync-locked tracks also shorten but with a cross-fade, unlike in other sync locks)

Recommended:

### DIFF
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -15,9 +15,13 @@
 int EffectOutputTracks::nEffectsDone = 0;
 
 EffectOutputTracks::EffectOutputTracks(
-   TrackList &tracks, bool allSyncLockSelected
-)  : mTracks{ tracks }
+   TrackList& tracks, std::optional<TimeInterval> effectTimeInterval,
+   bool allSyncLockSelected)
+    : mTracks { tracks }
 {
+   assert(
+      !effectTimeInterval.has_value() ||
+      effectTimeInterval->first <= effectTimeInterval->second);
    // Reset map
    mIMap.clear();
    mOMap.clear();
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -12,6 +12,7 @@
 #include "SyncLock.h"
 #include "UserException.h"
 #include "WaveTrack.h"
+#include "WaveTrackUtilities.h"
 
 // Effect application counter
 int EffectOutputTracks::nEffectsDone = 0;
@@ -51,9 +52,13 @@ EffectOutputTracks::EffectOutputTracks(
       auto progress = MakeProgress(
          XO("Pre-processing"), XO("Rendering Time-Stretched Audio"),
          ProgressShowCancel);
-      const auto waveTracks = stretchSyncLocked
+      const auto waveTracks = (stretchSyncLocked
          ? mOutputTracks->Any<WaveTrack>()
-         : mOutputTracks->Selected<WaveTrack>();
+         : mOutputTracks->Selected<WaveTrack>())
+      + [&](const WaveTrack *pTrack){
+         return WaveTrackUtilities::HasStretch(*pTrack,
+            effectTimeInterval->first, effectTimeInterval->second);
+      };
       const auto numTracks = waveTracks.size();
       auto count = 0;
       auto reportProgress = [&](double progressFraction) {
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -18,7 +18,7 @@ int EffectOutputTracks::nEffectsDone = 0;
 
 EffectOutputTracks::EffectOutputTracks(
    TrackList& tracks, std::optional<TimeInterval> effectTimeInterval,
-   bool allSyncLockSelected)
+   bool allSyncLockSelected, bool stretchSyncLocked)
     : mTracks { tracks }
 {
    assert(
@@ -51,7 +51,9 @@ EffectOutputTracks::EffectOutputTracks(
       auto progress = MakeProgress(
          XO("Pre-processing"), XO("Rendering Time-Stretched Audio"),
          ProgressShowCancel);
-      const auto waveTracks = mOutputTracks->Any<WaveTrack>();
+      const auto waveTracks = stretchSyncLocked
+         ? mOutputTracks->Any<WaveTrack>()
+         : mOutputTracks->Selected<WaveTrack>();
       const auto numTracks = waveTracks.size();
       auto count = 0;
       auto reportProgress = [&](double progressFraction) {
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -38,31 +38,35 @@ EffectOutputTracks::EffectOutputTracks(
 
    for (auto aTrack : trackRange) {
       auto list = aTrack->Duplicate();
-      if (effectTimeInterval.has_value() && effectTimeInterval->second > effectTimeInterval->first)
-      {
-         using namespace BasicUI;
-         auto progress = MakeProgress(
-            XO("Pre-processing"), XO("Rendering Time-Stretched Audio"),
-            ProgressShowCancel);
-         const auto waveTracks = list->Any<WaveTrack>();
-         const auto numTracks = waveTracks.size();
-         auto count = 0;
-         auto reportProgress = [&](double progressFraction) {
-            const auto overallProgress = (count + progressFraction) / numTracks;
-            const auto result = progress->Poll(overallProgress * 1000, 1000);
-            if (result != ProgressResult::Success)
-               throw UserException {};
-         };
-         for (const auto& track : waveTracks)
-         {
-            track->ApplyStretchRatio(effectTimeInterval, reportProgress);
-            ++count;
-         }
-      }
       mIMap.push_back(aTrack);
       mOMap.push_back(*list->begin());
       mOutputTracks->Append(std::move(*list));
    }
+
+   if (
+      effectTimeInterval.has_value() &&
+      effectTimeInterval->second > effectTimeInterval->first)
+   {
+      using namespace BasicUI;
+      auto progress = MakeProgress(
+         XO("Pre-processing"), XO("Rendering Time-Stretched Audio"),
+         ProgressShowCancel);
+      const auto waveTracks = mOutputTracks->Any<WaveTrack>();
+      const auto numTracks = waveTracks.size();
+      auto count = 0;
+      auto reportProgress = [&](double progressFraction) {
+         const auto overallProgress = (count + progressFraction) / numTracks;
+         const auto result = progress->Poll(overallProgress * 1000, 1000);
+         if (result != ProgressResult::Success)
+            throw UserException {};
+      };
+      for (const auto& track : waveTracks)
+      {
+         track->ApplyStretchRatio(effectTimeInterval, reportProgress);
+         ++count;
+      }
+   }
+
    // Invariant is established
    assert(mIMap.size() == mOutputTracks->Size());
    assert(mIMap.size() == mOMap.size());
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -8,7 +8,9 @@
 
 **********************************************************************/
 #include "EffectOutputTracks.h"
+#include "BasicUI.h"
 #include "SyncLock.h"
+#include "UserException.h"
 #include "WaveTrack.h"
 
 // Effect application counter
@@ -36,6 +38,27 @@ EffectOutputTracks::EffectOutputTracks(
 
    for (auto aTrack : trackRange) {
       auto list = aTrack->Duplicate();
+      if (effectTimeInterval.has_value() && effectTimeInterval->second > effectTimeInterval->first)
+      {
+         using namespace BasicUI;
+         auto progress = MakeProgress(
+            XO("Pre-processing"), XO("Rendering Time-Stretched Audio"),
+            ProgressShowCancel);
+         const auto waveTracks = list->Any<WaveTrack>();
+         const auto numTracks = waveTracks.size();
+         auto count = 0;
+         auto reportProgress = [&](double progressFraction) {
+            const auto overallProgress = (count + progressFraction) / numTracks;
+            const auto result = progress->Poll(overallProgress * 1000, 1000);
+            if (result != ProgressResult::Success)
+               throw UserException {};
+         };
+         for (const auto& track : waveTracks)
+         {
+            track->ApplyStretchRatio(effectTimeInterval, reportProgress);
+            ++count;
+         }
+      }
       mIMap.push_back(aTrack);
       mOMap.push_back(*list->begin());
       mOutputTracks->Append(std::move(*list));
--- a/libraries/lib-effects/EffectOutputTracks.h
+++ b/libraries/lib-effects/EffectOutputTracks.h
@@ -37,7 +37,7 @@ public:
     copies those clips are split, and new clips bounded by the interval, with
     the stretches applied, are inserted.
     @pre `!effectTimeInterval.has_value() ||
-       effectTimeInterval->first < effectTimeInterval->second`
+       effectTimeInterval->first <= effectTimeInterval->second`
     */
    EffectOutputTracks(
       TrackList& tracks, std::optional<TimeInterval> effectTimeInterval,
--- a/libraries/lib-effects/EffectOutputTracks.h
+++ b/libraries/lib-effects/EffectOutputTracks.h
@@ -14,6 +14,7 @@ class Track;
 class TrackList;
 
 #include <memory>
+#include <optional>
 #include <vector>
 
 //! Use this object to copy the input tracks to tentative outputTracks
@@ -29,7 +30,17 @@ public:
    static int nEffectsDone;
    static void IncEffectCounter() { ++nEffectsDone; }
 
-   EffectOutputTracks(TrackList &tracks,
+   using TimeInterval = std::pair<double, double>;
+   /*!
+    @param effectTimeInterval if given, and any copied tracks
+    have clips with non-unit stretch intersecting that interval, then in the
+    copies those clips are split, and new clips bounded by the interval, with
+    the stretches applied, are inserted.
+    @pre `!effectTimeInterval.has_value() ||
+       effectTimeInterval->first < effectTimeInterval->second`
+    */
+   EffectOutputTracks(
+      TrackList& tracks, std::optional<TimeInterval> effectTimeInterval,
       bool allSyncLockSelected = false);
    EffectOutputTracks(const EffectOutputTracks&) = delete;
 
--- a/libraries/lib-effects/EffectOutputTracks.h
+++ b/libraries/lib-effects/EffectOutputTracks.h
@@ -36,12 +36,18 @@ public:
     have clips with non-unit stretch intersecting that interval, then in the
     copies those clips are split, and new clips bounded by the interval, with
     the stretches applied, are inserted.
+    @param allSyncLockSelected if true, unselected tracks that are sync-locked
+    with a selected track are copied too
+    @param stretchSyncLocked if false, do not apply the stretch interval to any
+    unselected WaveTrack that is copied
+
     @pre `!effectTimeInterval.has_value() ||
        effectTimeInterval->first <= effectTimeInterval->second`
     */
    EffectOutputTracks(
       TrackList& tracks, std::optional<TimeInterval> effectTimeInterval,
-      bool allSyncLockSelected = false);
+      bool allSyncLockSelected = false,
+      bool stretchSyncLocked = false);
    EffectOutputTracks(const EffectOutputTracks&) = delete;
 
    ~EffectOutputTracks();
--- a/libraries/lib-effects/PerTrackEffect.cpp
+++ b/libraries/lib-effects/PerTrackEffect.cpp
@@ -65,7 +65,7 @@ bool PerTrackEffect::Process(
    EffectInstance &instance, EffectSettings &settings) const
 {
    auto pThis = const_cast<PerTrackEffect *>(this);
-   EffectOutputTracks outputs{ *mTracks, true };
+   EffectOutputTracks outputs{ *mTracks, {{ mT0, mT1 }}, true };
    bool bGoodResult = true;
    // mPass = 1;
    if (DoPass1()) {
--- a/libraries/lib-effects/PerTrackEffect.cpp
+++ b/libraries/lib-effects/PerTrackEffect.cpp
@@ -286,7 +286,8 @@ bool PerTrackEffect::ProcessPass(TrackList &outputs,
             genLength, sampleRate, leader, inBuffers, outBuffers);
          if (bGoodResult) {
             sink.Flush(outBuffers);
-            if (tempList) {
+            bGoodResult = sink.IsOk();
+            if (bGoodResult && tempList) {
                if (!results)
                   results = tempList;
                else
--- a/libraries/lib-effects/PerTrackEffect.cpp
+++ b/libraries/lib-effects/PerTrackEffect.cpp
@@ -31,8 +31,6 @@
 #include "WaveTrackSink.h"
 #include "WideSampleSource.h"
 
-AudioGraph::Sink::~Sink() = default;
-
 PerTrackEffect::Instance::~Instance() = default;
 
 bool PerTrackEffect::Instance::Process(EffectSettings &settings)
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -393,6 +393,9 @@ private:
    //! Subclass responsibility implements only a part of Duplicate(), copying
    //! the track data proper (not associated data such as for groups and views)
    /*!
+    @param unstretchInterval If set, this time interval's stretching must be applied.
+    @pre `!unstretchInterval.has_value() ||
+       unstretchInterval->first < unstretchInterval->second`
     @pre `IsLeader()`
     @post result: `NChannels() == result->NChannels()`
     */
@@ -581,7 +584,7 @@ public:
     @tparam F returns a shared pointer to Attachment (or some subtype of it)
 
     @pre `f` never returns null
-   
+
     `f` may assume the precondition that the given channel index is less than
     the given track's number of channels
     */
--- a/libraries/lib-wave-track/CMakeLists.txt
+++ b/libraries/lib-wave-track/CMakeLists.txt
@@ -29,6 +29,8 @@ set( SOURCES
    WaveTrack.h
    WaveTrackSink.cpp
    WaveTrackSink.h
+   WaveTrackUtilities.cpp
+   WaveTrackUtilities.h
    WideClip.cpp
    WideClip.h
 )
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -1947,7 +1947,8 @@ void WaveTrack::InsertClip(WaveClipHolder clip)
    mClips.push_back(std::move(clip));
 }
 
-void WaveTrack::ApplyStretchRatio(std::optional<TimeInterval> interval)
+void WaveTrack::ApplyStretchRatio(
+   std::optional<TimeInterval> interval, ProgressReporter reportProgress)
 {
    assert(IsLeader());
    // Assert that the interval is reasonable, but this function will be no-op
@@ -1964,11 +1965,23 @@ void WaveTrack::ApplyStretchRatio(std::optional<TimeInterval> interval)
                  GetEndTime();
    if (startTime >= endTime)
       return;
+   const auto numChannels = NChannels();
+   auto count = 0;
+   const ProgressReporter reportChannelProgress =
+      reportProgress ?
+         [&](double progress) {
+            reportProgress((count + progress) / numChannels);
+         } :
+         ProgressReporter {};
    for (const auto pChannel : TrackList::Channels(this))
-      pChannel->ApplyStretchRatioOne(startTime, endTime);
+   {
+      pChannel->ApplyStretchRatioOne(startTime, endTime, reportChannelProgress);
+      ++count;
+   }
 }
 
-void WaveTrack::ApplyStretchRatioOne(double t0, double t1)
+void WaveTrack::ApplyStretchRatioOne(
+   double t0, double t1, const ProgressReporter& reportProgress)
 {
    if (auto clipAtT0 = GetClipAtTime(t0); clipAtT0 &&
                                           clipAtT0->SplitsPlayRegion(t0) &&
@@ -1979,10 +1992,17 @@ void WaveTrack::ApplyStretchRatioOne(double t0, double t1)
                                           !clipAtT1->StretchRatioEquals(1))
       SplitAt(t1);
    auto clip = GetClipAtTime(t0);
+   const auto numClips = GetNumClips(t0, t1);
+   auto count = 0;
+   auto reportClipProgress = reportProgress ? [&](double progress) {
+      reportProgress((count + progress) / numClips);
+   }
+      : ProgressReporter{};
    while (clip && clip->GetPlayStartTime() < t1)
    {
-      clip->ApplyStretchRatio();
+      clip->ApplyStretchRatio(reportClipProgress);
       clip = GetNextClip(*clip, PlaybackDirection::forward);
+      ++count;
    }
 }
 
@@ -3123,6 +3143,22 @@ const WaveClip* WaveTrack::GetClipByIndex(int index) const
 int WaveTrack::GetNumClips() const
 {
    return mClips.size();
+}
+
+int WaveTrack::GetNumClips(double t0, double t1) const
+{
+   const auto clips = SortedClipArray();
+   // Find first position where the comparison is false
+   const auto firstIn = std::lower_bound(clips.begin(), clips.end(), t0,
+      [](const auto& clip, double t0) {
+         return clip->GetPlayEndTime() <= t0;
+      });
+   // Find first position where the comparison is false
+   const auto firstOut = std::lower_bound(firstIn, clips.end(), t1,
+      [](const auto& clip, double t1) {
+         return clip->GetPlayStartTime() < t1;
+      });
+   return std::distance(firstIn, firstOut);
 }
 
 bool WaveTrack::CanOffsetClips(
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -1715,6 +1715,8 @@ void WaveTrack::SyncLockAdjust(double oldT1, double newT1)
          for (const auto pChannel : channels) {
             auto tmp = std::make_shared<WaveTrack>(
                mpFactory, GetSampleFormat(), GetRate());
+            // tmpList exists only to fix assertion crashes in usage of tmp
+            auto tmpList = TrackList::Temporary(nullptr, tmp, nullptr);
             assert(tmp->IsLeader()); // It is not yet owned by a TrackList
             tmp->InsertSilence(0.0, duration);
             tmp->FlushOne();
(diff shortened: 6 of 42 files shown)


saintmatthieu:
I am going through the QA checklist for the original 5-of-6 PR, trying (almost) each item on both branches. I'll make separate comments for separate observations.

saintmatthieu:
Paulstretch:
1. Stretch a clip,
2. set preferences to editing a clip can move other clips,
3. select some region inside the clip and apply Paulstretch to a factor of 10.

I get such result on this branch:
image

There shouldn't be the gap. #5044 doesn't suffer this problem.

saintmatthieu:
Auto Duck: works for neither this or the original branch.

saintmatthieu:
1. Select a region on some stretched clip:
image

2. generate silence. On the original branch, the program crashes. On this branch, no crash, but unexpected result :
image

saintmatthieu:
Noise reduction:
On both branches, generating the noise profile stretch-renders the selected region and creates a clip (without creating an undo item):
image

Now I go back one step in the undo history (undoing whatever I did before generating the noise profile) and redo. Noise profile should still be in memory (?) but my undo/redo got rid of the unwanted rendered clip. Now I apply noise reduction. On this branch, I get the same spurious clip pattern as when generating silence: image

On the original branch, it works.

saintmatthieu:
Graphic EQ fails on this branch: 
image

saintmatthieu:
High-pass filter doesn't work either, yielding this spurious-clip pattern seen in generating silence or applying noise suppression. (Couldn't test on my branch, somehow the effect doesn't show up in submenu.)

saintmatthieu:
Mixing stretched stereo track to mono yields silence (on both branches)

saintmatthieu:
Truncate silence: works on this branch but not on the original branch (doesn't seem to pre-render)

saintmatthieu:
Now I know the rationale of this PR : get the destructive effects working, cherry-picking only the necessary part of 4-of-6 to make this work, i.e., `WaveClip::ApplyStretchRatio`. [Generalize use of progress bar and avoid unnecessary stretching if clip ratios match](https://github.com/audacity/audacity/pull/5045/commits/1985c7f3e179ad94994ad68f7f036081045c6fb8) introduced the `targetRatio` argument to that method, not necessary for this PR, but we ought not to forget it when repairing copy-paste later on if this PR makes it to master.

Paul-Licameli:
Rebased unchanged onto recent master

Paul-Licameli:
Fixed intermediate commit that failed to build

Paul-Licameli:
New commit fixes at least some of the problems when editing a clip can move other clips

Paul-Licameli:
> Now I know the rationale of this PR : get the destructive effects working, cherry-picking only the necessary part of 4-of-6 to make this work, i.e., `WaveClip::ApplyStretchRatio`. [Generalize use of progress bar and avoid unnecessary stretching if clip ratios match](https://github.com/audacity/audacity/pull/5045/commits/1985c7f3e179ad94994ad68f7f036081045c6fb8) introduced the `targetRatio` argument to that method, not necessary for this PR, but we ought not to forget it when repairing copy-paste later on if this PR makes it to master.

This branch does not include all of the 6th of 6.  Right, don't forget the rest of that one.

saintmatthieu:
> Paulstretch:
> 
> 1. Stretch a clip,
> 2. set preferences to editing a clip can move other clips,
> 3. select some region inside the clip and apply Paulstretch to a factor of 10.
> 
> I get such result on this branch: image
> 
> There shouldn't be the gap. #5044 doesn't suffer this problem.
> 
> If the checkbox "edit a clip can move other clips" isn't checked, this experiment works, on condition of course that the stretch factor <= 1.

This looks good now.

saintmatthieu:
> Auto Duck: works for neither this or the original branch.

@Paul-Licameli more detail:
Without stretching : image

If I stretch the clip on the lower track, though, applying auto-duck to the upper track does nothing.

saintmatthieu:
> 1. Select a region on some stretched clip:
>    image
> 2. generate silence. On the original branch, the program crashes. On this branch, no crash, but unexpected result :
>    image
> 
> This is if the "edit a clip can move other clips" is checked. Without it, in debug, `assert(target.GetTrimLeft() == 0);` fails in `WaveTrack::ClearAndPasteOne`.

Now it doesn't matter whether edit-can-move or not : the assert always fails.

saintmatthieu:
> Graphic EQ fails on this branch: image
> 
> The gap shouldn't be there. Another effect of the EQ submenu, Bass and Treble, works, though.

This works now.

saintmatthieu:
> Effects that yield the spurious-clip pattern seen in the generating silence case:
> 
> * Generating silence
> * Applying noise suppression
> * High-pass filter
> * Change tempo
> * Change pitch
> * Nyquist noise gate

These still don't work, at least silence, NS (but you're working on this one @Paul-Licameli ), high-pass and change pitch. Haven't tried the others, but I reckon they also don't work.

Paul-Licameli:
> > 1. Select a region on some stretched clip:
> >    image
> > 2. generate silence. On the original branch, the program crashes. On this branch, no crash, but unexpected result :
> >    image
> > 
> > This is if the "edit a clip can move other clips" is checked. Without it, in debug, `assert(target.GetTrimLeft() == 0);` fails in `WaveTrack::ClearAndPasteOne`.
> 
> Now it doesn't matter whether edit-can-move or not : the assert always fails.

Last evening I studied that complicated function ClearAndPasteOne.

There may be an explanation as with the bug that shifted clips too far right:  wrong inclusiveness or exclusiveness of the endpoints of a test for whether a point is in an interval, introduced when 2 of 6 was merged.

If so it would be good to understand just where this subtle problem is, not just hide it behind the 4 of 6 commit, which I think should be a logically independent fix of a different problem.

saintmatthieu:
> Now it doesn't matter whether edit-can-move or not : the assert always fails.

I think we can ignore this for now. Generating a tone with the selected region over a clip means is done as follows: create an empty clip, write the generated tone into it, clear the selected region and insert the tone in it. Since this involves copy-pasting clips onto one another, it will be repaired in the 4-of-6 PR.

Paul-Licameli:
> > Auto Duck: works for neither this or the original branch.
> 
> @Paul-Licameli more detail: Without stretching : image
> 
> If I stretch the clip on the lower track, though, applying auto-duck to the upper track does nothing.

I wonder if this is in common with the un-corrected mixing of stereo to mono.  For reasons I didn't figure out, apparently there are zeroes read from the stretched track instead of the proper values.

Paul-Licameli:
> > Effects that yield the spurious-clip pattern seen in the generating silence case:
> > 
> > * Generating silence
> > * Applying noise suppression
> > * High-pass filter
> > * Change tempo
> > * Change pitch
> > * Nyquist noise gate
> 
> These still don't work, at least silence, NS (but you're working on this one @Paul-Licameli ), high-pass and change pitch. Haven't tried the others, but I reckon they also don't work.

Have you tried again with the latest changes?

Paul-Licameli:
> Effects that yield the spurious-clip pattern seen in the generating silence case:
> 
>     * Generating silence
> 
>     * Applying noise suppression
> 
>     * High-pass filter
> 
>     * Change tempo
> 
>     * Change pitch
> 
>     * Nyquist noise gate

I'm testing every place under src/effect where `ClearAndPaste` is used.  Project tempo is 120 and 4/4 time.

I generate 4 bars of noise in a mono track, stretch out to 6, select from 2 to 4, then apply an effect.

- Change speed
- High-pass filter (Nyquist.cpp)
- Generate silence (Generator.cpp)
- Noise reduction, step 2
- Tone generator (hits PerTrackEffect)
- Change Tempo, high quality on (SBSMSEffect)
- Change Tempo, high quality off (SoundTouchEffect)

In all of these, overlapping clips result.  That's bad.  I expect the same for other Nyquist effects, Chirp, Noise, and the sliding pitch shift which visit the same functions.

But for any other effect, this PR makes an improvement over the master branch behavior.

src/effects/AutoDuck.cpp
```diff
@@ -189,6 +185,30 @@ bool EffectAutoDuck::Process(EffectInstance &, EffectSettings &)
...
 
+   TrackListHolder tempTracks;
+   auto pControlTrack = mControlTrack;
+   // If there is any stretch in the control track, substitute a temporary
+   // rendering before trying to use GetFloats
+   {
+      auto &clips = pControlTrack->GetClips();
+      const auto t0 = pControlTrack->LongSamplesToTime(start);
+      const auto t1 = pControlTrack->LongSamplesToTime(end);
+      const bool hasStretch = any_of(clips.begin(), clips.end(),
```
saintmatthieu, line 196: `WaveTrack::Duplicate` does this check for you. We should consider making it a public helper method of WaveTrack to places like these simpler.

saintmatthieu: Pardon me I thought it did: not really, or intricately.
That said, `Duplicate` won't apply stretch ratio if no clip is stretched in the interval. So behaviour would be preserved removing those checks - although it would have a negative impact on performance.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -3050,7 +3050,7 @@ int WaveTrack::GetNumClips(double t0, double t1) const
    // Find first position where the comparison is false
    const auto firstOut = std::lower_bound(firstIn, clips.end(), t1,
       [](const auto& clip, double t1) {
-         return clip->GetPlayStartTime() <= t1;
+         return clip->GetPlayStartTime() < t1;
```
saintmatthieu, line 3058: Nice !

src/effects/StereoToMono.cpp
```diff
@@ -79,7 +79,10 @@ bool EffectStereoToMono::Process(EffectInstance &, EffectSettings &)
 {
    // Do not use mWaveTracks here.  We will possibly DELETE tracks,
    // so we must use the "real" tracklist.
-   EffectOutputTracks outputs { *mTracks, { mT0, mT1 } };
+   EffectOutputTracks outputs { *mTracks,
+      // This effect ignores mT0 and mT1 but always mixes the entire tracks.
```
saintmatthieu, line 83: 👍

Paul-Licameli:
Rebased onto recent master and resolved conflicts

libraries/lib-effects/EffectOutputTracks.h
```diff
@@ -29,8 +29,9 @@ class EFFECTS_API EffectOutputTracks
    static int nEffectsDone;
    static void IncEffectCounter() { ++nEffectsDone; }
 
+   using Interval = std::pair<double, double>;
```
saintmatthieu, line 32: I've been using `TimeInterval` on my other branch, which I think is appropriate since it's a pair of doubles.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -3073,14 +3073,16 @@ int WaveTrack::GetNumClips() const
 int WaveTrack::GetNumClips(double t0, double t1) const
 {
    const auto clips = SortedClipArray();
-   const auto firstIn =
-      std::find_if(clips.begin(), clips.end(), [t0](const auto& clip) {
-         return t0 < clip->GetPlayEndTime();
+   // Find first position where the comparison is false
+   const auto firstIn = std::lower_bound(clips.begin(), clips.end(), t0,
```
saintmatthieu, line 3088: I prefer the way it was before: easier to understand (I always have to look up the subtlety between lower_bound and upper_bound) and on usage average seldom more expansive performance-wise. But sure ...

saintmatthieu (dismissed):
Please consider the comments I left before merging.

Paul-Licameli:
If you understood my last criticisms of #5139, then you will understand my dislike of the proposed complications for the analysis phase of Noise Reduction.

If the selection for analysis happens to span clips of different stretch, then there may be mis-weighting of things when combining results.

Admittedly this is very unlikely for users to do, who use noise reduction intelligently (selecting only a small sufficient duration).

But on the other hand, for just that reason, I think the pursuit of this supposed efficiency isn’t worth it either.

I would keep my simple fix:  just render the stretch if needed, only on the selection, likely to be short, and simply avoid the side effect on tracks by not committing the EffectOutputTracks object.

saintmatthieu:
Referring to my comment in the spectrum-analysis PR : https://github.com/audacity/audacity/pull/5139#discussion_r1319559253

The experiment I did there is I think more likely to happen in the noise reduction use case: the selection is likely to be at the start of a clip, where silence is but speech hasn't started, and unfortunately this is where the algorithm is in a transient phase.

The difference in spectrum characteristic is arguably not dramatic, but used as a profile it might not perform as well when used in the processing of other parts.

On the other hand, one could argue that the profile selection will probably be used to process the same clip, and since we know the RMS is unfortunately different before and after rendering the stretch, it's safer to take stretch-rendered data as profile for the processing of data that will also be stretch-rendered.

Now I've put all my arguments forth, and will let you decide. It's your baby after all.

libraries/lib-effects/EffectOutputTracks.h
Paul-Licameli: I will change the name

saintmatthieu (approved):

saintmatthieu:
Pushed another commit to fix the "Silence Audio" command.
Generate tone, stretch clip, select region and "Silence Audio" (ctrl+L)
Before fix (possible result, you may get the "Internal error" message too):
image

After fix:
image

libraries/lib-time-track/TimeTrack.cpp
```diff
@@ -222,7 +222,8 @@ void TimeTrack::Paste(double t, const Track &src)
       (void)0;// intentionally do nothing.
 }
 
-void TimeTrack::Silence(double WXUNUSED(t0), double WXUNUSED(t1))
+void TimeTrack::Silence(
+   double WXUNUSED(t0), double WXUNUSED(t1), ProgressReporter)
```
Paul-Licameli, line 226: Eliminate `WXUNUSED` when you have an excuse to disturb a function with that macro.
Just declare `double, double`

Or if you want to be really modern: https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

src/menus/EditMenus.cpp
```diff
@@ -738,8 +738,20 @@ void OnSilence(const CommandContext &context)
    auto &tracks = TrackList::Get(project);
    auto &selectedRegion = ViewInfo::Get(project).selectedRegion;
 
-   for (auto n : tracks.Selected<WaveTrack>())
-      n->Silence(selectedRegion.t0(), selectedRegion.t1());
+   const auto selectedWaveTracks = tracks.Selected<WaveTrack>();
+   auto progress = BasicUI::MakeProgress(
+      XO("Pre-processing"), XO("Rendering Time-Stretched Audio"));
```
Paul-Licameli, line 743: New strings.  @LWinterberg

saintmatthieu: FYI copy-paste of agreed-upon strings from other places where pre-rendering is needed.

dozzzzer:
Tested. Found issues:

- When clicking either **Cancel** or **Stop** in the **Rendering Time-Stretched Audio** dialog, progress hangs for a while but finishes in the background anyway, so basically neither of these two buttons is functional
- When applying "Repair" effect on a selection of max. 128 samples, it creates a spurious unwanted clip with gaps on each side:



- When applying the Noise Reduction effect (i.e. step 2), it splits the selection into a separate clip:

https://github.com/audacity/audacity/assets/70063970/8c85be70-bd2e-48c3-856c-cc059f056b10

(to be continued)

Paul-Licameli:
Rebased onto recent master and sqaushed the fixup for Noise Reduction.

src/effects/Repair.cpp
```diff
@@ -98,6 +97,13 @@ bool EffectRepair::Process(EffectInstance &, EffectSettings &)
             bGoodResult = false;
             break;
          }
+         if (WaveTrackUtilities::HasStretch(*track, repair_t0, repair_t1)) {
+            EffectUIServices::DoMessageBox(*this,
+               XO(
+"The Repair effect cannot be applied within stretched or shrunk clips") );
```
Paul-Licameli, line 103: @LWinterberg

libraries/lib-track/Track.h
```diff
@@ -399,7 +399,8 @@ class TRACK_API Track /* not final */
    //! Subclass responsibility implements only a part of Duplicate(), copying
    //! the track data proper (not associated data such as for groups and views)
    /*!
-    @param unstretchInterval If set, this time interval's stretching must be applied.
+    @param unstretchInterval If set and of positive duration, this time
```
saintmatthieu, line 402: I don't think someone should be able to call `ApplyStretchRatio` with an empty interval. I'd rather have `@pre `!unstretchInterval || unstretchInterval->second > unstretchInterval->first`. Then, in EffectOutputTracks.cpp, if the interval is empty, call `WaveTrack::Duplicate` with an `std::nullopt` (or equivalent).

saintmatthieu (changes requested):

src/effects/Repair.cpp
LWinterberg: wording sounds good

libraries/lib-track/Track.h
saintmatthieu: I realize that the `unstretchInterval` arguments of `Track::Duplicate` and `WaveTrack::ApplyStretchRatio` have opposite meanings: stretch nothing in the former and everything in the latter. Maybe we should remove that argument from `Duplicate` and `Clone` and let the callers make a call after that to `ApplyStretchRatio`.

libraries/lib-wave-track/CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ set( SOURCES
    WaveTrack.h
    WaveTrackSink.cpp
    WaveTrackSink.h
+   WaveTrackUtilities.cpp
+   WaveTrackUtilities.h
```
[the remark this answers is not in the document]
Paul-Licameli: I think smaller is often better for classes.

The function I move into that file was defined in an earlier commit without intrusion into WaveTrack.  It wasn't necessary.  The existing public interfaces were sufficient to define it.

So then, why disturb and complicate the interface of WaveTrack without need?  I'm making the function available for reuse but keeping it separate from the class.

I think we may identify more inessential member functions of WaveTrack that could also be moved out.  It's a big file doing too much.

I think it would be better to have a smaller WaveTrack class someone could study, which provides just the essential services and maintains important class invariants.

dozzzzer:
- The Repair effect is disabled for stretched/shrunk clips
- The issue with non-op progress dialog buttons is now fixed
in the most recent commit of this PR

@Paul-Licameli @saintmatthieu

src/effects/Repair.cpp
```diff
@@ -98,6 +97,13 @@ bool EffectRepair::Process(EffectInstance &, EffectSettings &)
             bGoodResult = false;
             break;
          }
+         if (WaveTrackUtilities::HasStretch(*track, repair_t0, repair_t1)) {
+            EffectUIServices::DoMessageBox(*this,
```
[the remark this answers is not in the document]
LWinterberg: semantically, the (i) message is wrong in both cases anyway as we error out and abort. The red (x) would be more appropriate here. 

I have no strong feeling whether to use the big or the small text version. The big version feels a bit Vista/7, but I also have no strong desire to see it changed at the moment.

libraries/lib-effects/EffectOutputTracks.cpp
```diff
@@ -38,7 +39,9 @@ EffectOutputTracks::EffectOutputTracks(
    auto count = 0;
    auto reportProgress = [&](double progressFraction) {
       const auto overallProgress = (count + progressFraction) / numTracks;
-      progress->Poll(overallProgress * 1000, 1000);
+      const auto result = progress->Poll(overallProgress * 1000, 1000);
```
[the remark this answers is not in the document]
Paul-Licameli: Done.

I recall there was a difference between Stop and Cancel for import and export.  I didn't look to see if that was conserved after all the refactoring in this release.

It was possible to Stop with a partially exported file instead of aborting the whole thing and deleting the file.

src/effects/Repair.cpp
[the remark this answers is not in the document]
saintmatthieu: All right, so change to red, but what's the required change for the looks-and-feel (if any) ?

libraries/lib-wave-track/CMakeLists.txt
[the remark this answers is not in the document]
saintmatthieu: No strong objection to that.

saintmatthieu (dismissed):

libraries/lib-effects/EffectOutputTracks.cpp
```diff
@@ -35,24 +36,32 @@ EffectOutputTracks::EffectOutputTracks(
...
+            const auto result = progress->Poll(overallProgress * 1000, 1000);
+            if (result != ProgressResult::Success)
+               throw UserException {};
+         };
+         for (const auto& track : waveTracks)
+         {
+            track->ApplyStretchRatio(effectTimeInterval, reportProgress);
+            ++count;
+         }
+      }
```
Paul-Licameli, line 61: This makes a progress indicator for each wave track separately.

libraries/lib-effects/EffectOutputTracks.cpp
```diff
@@ -35,24 +36,32 @@ EffectOutputTracks::EffectOutputTracks(
...
       mIMap.push_back(aTrack);
       mOMap.push_back(*list->begin());
       mOutputTracks->Append(std::move(*list));
-      ++count;
```
Paul-Licameli: Instead you could move stretching after this loop.  Count wave tracks or maybe only the tracks needing a stretch first and construct a progress indicator with that denominator.  Then another loop to do stretching.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -476,6 +474,17 @@ class WAVE_TRACK_API WaveTrack final
     */
    void Trim(double t0, double t1) /* not override */;
 
+   /*
+    * @param interval Entire track is rendered if nullopt, else only samples
+    * within [interval->first, interval->second), in which case clips are split
+    * at these boundaries before rendering - if rendering is needed.
+    *
+    * Precondition as for Clone
```
Paul-Licameli, line 482: Remove that comment line

src/effects/AutoDuck.cpp
```diff
@@ -195,12 +195,14 @@ bool EffectAutoDuck::Process(EffectInstance &, EffectSettings &)
...
       const auto t1 = pControlTrack->LongSamplesToTime(end);
       if (WaveTrackUtilities::HasStretch(*pControlTrack, t0, t1)) {
-         // TODO a progress indicator?
-         tempTracks = pControlTrack->Duplicate({ { t0, t1 } });
+         tempTracks = pControlTrack->Duplicate();
          if (tempTracks) {
             const auto pFirstTrack = *tempTracks->Any<WaveTrack>().begin();
-            if (pFirstTrack)
+            if (pFirstTrack) {
+               // TODO a progress indicator?
```
Paul-Licameli, line 202: Maybe the progress indicator for stretch can be built by a reusable utility function in WaveTrackUtils.h.  It would take another std::function just to report the varying `count` and give back another std::function.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -476,6 +474,17 @@ class WAVE_TRACK_API WaveTrack final
...
+    * @pre `!interval.has_value() || interval->first <= interval->second`
+    */
```
Paul-Licameli, line 484: When a zero duration interval is given, will there be a split there or will nothing happen?   Comment that

src/effects/Repair.cpp
```diff
@@ -76,7 +76,7 @@ bool EffectRepair::Process(EffectInstance &, EffectSettings &)
    // But for now, Cancel isn't supported without this.
    // Repair doesn't make sense for stretched clips, so don't pass a stretch
    // interval.
-   EffectOutputTracks outputs { *mTracks };
+   EffectOutputTracks outputs { *mTracks, {} };
```
Paul-Licameli, line 79: Does this stretch all or nothing?  Should be nothing

libraries/lib-effects/PerTrackEffect.cpp
```diff
@@ -31,8 +31,6 @@
 #include "WaveTrackSink.h"
 #include "WideSampleSource.h"
 
-AudioGraph::Sink::~Sink() = default;
-
```
Paul-Licameli, line 35: Tell me what warning this fixes

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2774,7 +2774,7 @@ bool WaveChannel::Set(constSamplePtr buffer, sampleFormat format,
       if (clipEnd > start && clipStart < start+len)
       {
          // Test as also in WaveTrack::GetOne()
-         if (clip->GetStretchRatio() == 1.0)
+         if (clip->GetStretchRatio() != 1.0)
```
Paul-Licameli, line 2777: Duh

libraries/lib-effects/PerTrackEffect.cpp
saintmatthieu: `warning C4273: 'AudioGraph::Sink::~Sink': inconsistent dll linkage `

src/effects/Repair.cpp
saintmatthieu: Nothing.

src/effects/AutoDuck.cpp
saintmatthieu: I'm doing something like that in 4-of-6, where rendering happens also e.g. when joining.

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: It's implied: `in which case clips are split at these boundaries before rendering - if rendering is needed` -> empty interval, no rendering needed.

libraries/lib-effects/EffectOutputTracks.cpp
saintmatthieu: Sounds better. Trying that ...

libraries/lib-wave-track/WaveTrack.h
Paul-Licameli: So I understand that means, yes there is a split?

libraries/lib-effects/PerTrackEffect.cpp
Paul-Licameli: I see, it was a duplicate definition of `~Sink` which is still defined elsewhere.

src/effects/AutoDuck.cpp
Paul-Licameli: On further thought, the new utility function could just take a reference to a counter variable instead of a function.  It would return a function.

Of course you would need to use C++14 "extended capture" [] syntax to allocate the progress indicator with `make_unique` so it would be destroyed along with the `std::function` when no longer needed.

src/effects/Loudness.cpp
```diff
@@ -432,7 +432,8 @@ bool EffectLoudness::ProcessOne(WaveChannel &track, size_t nChannels,
       else {
          if (!ProcessBufferBlock(mult))
             return false;
-         StoreBufferBlock(track, nChannels, s, blockLen);
+         if (!StoreBufferBlock(track, nChannels, s, blockLen))
```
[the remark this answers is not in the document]
Paul-Licameli: That's a very alert catch.  Thank you.

libraries/lib-effects/PerTrackEffect.cpp
```diff
@@ -67,7 +67,7 @@ bool PerTrackEffect::Process(
    EffectInstance &instance, EffectSettings &settings) const
 {
    auto pThis = const_cast<PerTrackEffect *>(this);
-   EffectOutputTracks outputs{ *mTracks, {mT0, mT1}, true };
+   EffectOutputTracks outputs{ *mTracks, {{ mT0, mT1 }}, true };
```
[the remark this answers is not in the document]
Paul-Licameli: You are right.  But you have fixed this with another commit that weakened the assertion.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2780,6 +2780,10 @@ void WaveChannel::Set(constSamplePtr buffer, sampleFormat format,
 
       if (clipEnd > start && clipStart < start+len)
       {
+         // Test as also in WaveTrack::GetOne()
+         if (clip->GetStretchRatio() == 1.0)
```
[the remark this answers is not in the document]
Paul-Licameli: Again, duh

libraries/lib-effects/PerTrackEffect.cpp
```diff
@@ -288,11 +288,14 @@ bool PerTrackEffect::ProcessPass(TrackList &outputs,
             genLength, sampleRate, leader, inBuffers, outBuffers);
          if (bGoodResult) {
             sink.Flush(outBuffers);
-            if (tempList) {
-               if (!results)
-                  results = tempList;
-               else
-                  results->Append(std::move(*tempList));
+            bGoodResult = sink.IsOk();
+            if (bGoodResult) {
```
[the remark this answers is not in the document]
Paul-Licameli: Later

src/effects/Loudness.cpp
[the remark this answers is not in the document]
saintmatthieu: Not so alert, but through testing. Other effects behave like that, I didn't comment them all.

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: No, there is no split. I didn't realize this was so unclear - will clarify the comment when I get the chance.

saintmatthieu (approved):

Paul-Licameli:
> Not so alert, but through testing. Other effects behave like that, I didn't comment them all.

Like what -- you mean other effects behave correctly when there is a stretching error?

If any behave wrongly (as Loudness did) I should know it and fix it and you should withhold approval until I do.

saintmatthieu:
Would that be a regression, though ? On master, and I suspect on 3.3.3, too, some effects may fail, modify the content partially ? and then commit the result.

Paul-Licameli:
Rebased unchanged onto recent master

Paul-Licameli:
> Would that be a regression, though ? On master, and I suspect on 3.3.3, too, some effects may fail, modify the content partially ? and then commit the result.

The failure of stretching should be propagated so that the effect fails.

Loudness was failing to do that.  Do others fail to do that?

saintmatthieu:
Pardon me, I must have recalled wrongly: looked again at your commit where you check the return value of `WaveTrack::Set`. Loudness is now the only one I see that didn't make use of `false`. Feel free to double-check.

dozzzzer:
Found issues:
- when stretching/shrinking audio and trying to silence the selection by Ctrl (Cmd) + L, it does so in a different place of the clip:

https://github.com/audacity/audacity/assets/70063970/e64505d6-fbd2-4186-9382-ebe86765ac4f

- Trim Silence effect fails:

https://github.com/audacity/audacity/assets/70063970/78ea9a9a-069b-43c9-995c-081242068925

(to be continued)

saintmatthieu:
@dozzzzer The "Silence Audio" command not working is known and fixed in that other PR, #5182.

src/effects/TruncSilence.cpp
```diff
@@ -312,9 +313,8 @@ bool EffectTruncSilence::ProcessAll()
    // This list should always be kept in order.
    RegionList silences;
 
-   if (FindSilences(silences,
-      inputTracks()->Selected<const WaveTrack>())
-   ) {
+   if (FindSilences(silences, outputs.Get().Any<const WaveTrack>()))
```
Paul-Licameli, line 316: What if output tracks contains un-selected tracks of the sync lock group?  They would be analyzed too

src/effects/TruncSilence.cpp
```diff
@@ -517,6 +517,13 @@ bool EffectTruncSilence::Analyze(RegionList& silenceList,
 {
    assert(wt.IsLeader());
    const auto rate = wt.GetRate();
+   // const auto unstretchedHolder = wt.Duplicate();
+   // const auto unstretched = *unstretchedHolder->Any<WaveTrack>().begin();
+   // const auto progress = BasicUI::MakeProgress(
+   //    XO("Pre-processing"), XO("Rendering Time-Stretched Audio"),
+   //    BasicUI::ProgressShowCancel);
+   // unstretched->ApplyStretchRatio(std::nullopt, );
+
```
Paul-Licameli, line 526: This looks like an abandoned experiment you don’t want to commit

src/effects/TruncSilence.cpp
Paul-Licameli: I checked for the other use of FindSilences but that is correctly left unchanged as it already applies to an output track.

Paul-Licameli:
Rebased onto recent master with some squashing and permuting of commits.

Added two more commits.

Paul-Licameli:
More QA checkboxes, since we discovered there are interactions with Sync Lock

Paul-Licameli:
Force-pushed once more with one commit inserted first in the sequence to fix crashes in sync-lock

dozzzzer:
Tested

PR #5139, opened 2023-09-04
Repair plot spectrum command for stretched data

### LINKED ISSUE #5119: Plot Spectrum fails on a stretched clip
### Steps to reproduce

1. Stretch a 6-second stereo clip to, say, 20 seconds
2. Analyze > Plot Spectrum
3. Observe

@Paul-Licameli Also test:
- [ ] Make a good selection and open then dialog; but then make a bad selection of a stretched clip and Replot

### Expected behavior

A corresponding spectrum image in the Frequency Analysis dialog

### Actual behavior

It doesn't plot the spectrum, i.e., it shows the blank image. 

Screenshot 2023-09-08 at 09 57 40

### Audacity Version

master

saintmatthieu:
Resolves: #5119

Recommended:

### DIFF
--- a/src/FreqWindow.cpp
+++ b/src/FreqWindow.cpp
@@ -59,19 +59,20 @@ the mouse around.
 
 #include <math.h>
 
-#include "SelectFile.h"
-#include "ShuttleGui.h"
 #include "AColor.h"
+#include "AllThemeResources.h"
+#include "BasicUI.h"
 #include "CommonCommandFlags.h"
+#include "Decibels.h"
 #include "FFT.h"
 #include "PitchName.h"
-#include "Decibels.h"
 #include "Prefs.h"
 #include "Project.h"
 #include "ProjectWindow.h"
+#include "SelectFile.h"
+#include "ShuttleGui.h"
 #include "Theme.h"
 #include "ViewInfo.h"
-#include "AllThemeResources.h"
 
 #include "FileNames.h"
 
@@ -441,7 +442,7 @@ void FrequencyPlotDialog::Populate()
    // -------------------------------------------------------------------
    // ROW 5: Spacer
    // -------------------------------------------------------------------
-   
+
    S.AddSpace(5);
 
    S.SetBorder(2);
@@ -567,7 +568,8 @@ bool FrequencyPlotDialog::Show(bool show)
       dBRange = DecibelScaleCutoff.Read();
       if(dBRange < 90.)
          dBRange = 90.;
-      GetAudio();
+      if (!GetAudio())
+         return false;
       // Don't send an event.  We need the recalc right away.
       // so that mAnalyst is valid when we paint.
       //SendRecalcEvent();
@@ -579,7 +581,7 @@ bool FrequencyPlotDialog::Show(bool show)
    return res;
 }
 
-void FrequencyPlotDialog::GetAudio()
+bool FrequencyPlotDialog::GetAudio()
 {
    mData.reset();
    mDataLen = 0;
@@ -608,19 +610,30 @@ void FrequencyPlotDialog::GetAudio()
       }
       const auto nChannels = track->NChannels();
       if (track->GetRate() != mRate) {
-         AudacityMessageBox(
-            XO(
-"To plot the spectrum, all selected tracks must be the same sample rate.") );
+         using namespace BasicUI;
+         ShowMessageBox(
+            XO("To plot the spectrum, all selected tracks must have the same sample rate."),
+            MessageBoxOptions {}.Caption(XO("Error")).IconStyle(Icon::Error));
          mData.reset();
          mDataLen = 0;
-         return;
+         return false;
       }
       Floats buffer1{ mDataLen };
       Floats buffer2{ mDataLen };
       float *const buffers[]{ buffer1.get(), buffer2.get() };
       // Don't allow throw for bad reads
-      track->GetFloats(0, nChannels, buffers, start, mDataLen,
-         false, FillFormat::fillZero, false);
+      if (!track->GetFloats(
+             0, nChannels, buffers, start, mDataLen, false,
+             FillFormat::fillZero, false))
+      {
+         using namespace BasicUI;
+         ShowMessageBox(
+            XO("Audio could not be analyzed. This may be due to a stretched clip.\nTry resetting any stretched clips, or mixing and rendering the tracks before analyzing"),
+            MessageBoxOptions {}.Caption(XO("Error")).IconStyle(Icon::Error));
+         mData.reset();
+         mDataLen = 0;
+         return false;
+      }
       for (size_t i = 0; i < mDataLen; i++)
          mData[i] = buffers[0][i];
       for (size_t iChannel = 1; iChannel < nChannels; ++iChannel) {
@@ -632,7 +645,7 @@ void FrequencyPlotDialog::GetAudio()
    }
 
    if (selcount == 0)
-      return;
+      return false;
 
    if (warning) {
       auto msg = XO(
@@ -640,6 +653,7 @@ void FrequencyPlotDialog::GetAudio()
          .Format(mDataLen / mRate);
       AudacityMessageBox( msg );
    }
+   return true;
 }
 
 void FrequencyPlotDialog::OnSize(wxSizeEvent & WXUNUSED(event))
@@ -695,7 +709,7 @@ void FrequencyPlotDialog::DrawPlot()
       }
 
       memDC.SelectObject(wxNullBitmap);
-      
+
       mFreqPlot->Refresh();
 
       Refresh();
--- a/src/FreqWindow.h
+++ b/src/FreqWindow.h
@@ -67,7 +67,7 @@ public:
 private:
    void Populate();
 
-   void GetAudio();
+   bool GetAudio();
 
    void PlotMouseEvent(wxMouseEvent & event);
    void PlotPaint(wxPaintEvent & event);


Paul-Licameli (changes requested):
Reversion of my changes to AudioSegmentSampleView is questionable

src/FreqWindow.h
```diff
@@ -137,8 +141,7 @@ class FrequencyPlotDialog final : public wxDialogWrapper,
 
 
    double mRate;
-   size_t mDataLen;
-   Floats mData;
+   std::vector<ChannelGroupSampleView> mChannelGroupSampleViews;
```
Paul-Licameli, line 144: I see a new member variable in this dialog, which is non-modal and so may live a long time between button clicks.

That makes me want to examine its uses and see if instead local variables are used, or if that's not easy, then to be sure that the vector is cleared at appropriate times when the data are no longer needed.

saintmatthieu: > see if instead local variables are used

Not easy: the audio is acquired in `GetAudio`, and read in another method: `Recalc`.

> be sure that the vector is cleared at appropriate times when the data are no longer needed

See `OnCloseButton()`.

This PR does not change the times when memory is acquired or released.

saintmatthieu:
> Reversion of my changes to AudioSegmentSampleView is questionable

I missed that commit on 2-of-6.
I looked at it again, but couldn't figure out the reason behind it ?
For the spectrum analyzer it seems simpler to me to be returned a `sampleCount` and then check whether this exceeded the maximum allowed number of samples.

Paul-Licameli:
> > Reversion of my changes to AudioSegmentSampleView is questionable
> 
> I missed that commit on 2-of-6. I looked at it again, but couldn't figure out the reason behind it ? For the spectrum analyzer it seems simpler to me to be returned a `sampleCount` and then check whether this exceeded the maximum allowed number of samples.

I wrote a longer comment but the GitHub page ate it

Paul-Licameli:
That's really annoying.  I wrote a careful long line comment and chose "Request Changes" from the top right but it lost my long comment.

src/MixerBoard.cpp
```diff
@@ -527,7 +528,7 @@ void FillBufferWithSampleView(float* buffer, const ChannelSampleView& view)
    size_t copiedSamples = 0;
    for (const auto& segmentView : view)
    {
-      const auto sampleCount = segmentView.GetSampleCount();
+      const auto sampleCount = segmentView.GetSampleCount().as_size_t();
```
Paul-Licameli, line 531: Restating my lost comment:  This line is why I don't like the change.

`as_size_t` has an assertion that the value of the sample count is really in the range of a `size_t`.  I wasn't following the `@pre` comment convention yet those years ago.  It should be a `@pre` now.

And that means, there is burden of proof of satisfaction of the precondition at the point of the call.  But I don't know what the proof is here, and I don't think there is any sufficient information to prove it.

The point of commit 48aaca450b113a971a487f768c3e941accf9395a was to shift the burden of that proof instead to the places where `AudioSegmentSampleView` is constructed.

And note too, in that commit, the places where I added `@pre stretched samples in [t0, t1) can be counted in a `size_t` ` as an informal precondition, moving the burden of proof farther up.

And in the next commit after that, which is ba920717e826d4a0d6b5d81436d90f6a2e8f69df, I made those try-catches of `bad_alloc` in case the precondition was not satisfied.

Paul-Licameli: So I don't approve of the reversal of all of that caution that I put into an earlier commit.  And I must finish my examination of this pull request to see where you use the constructor and what proof of satisfaction of the precondition is possible.

Paul-Licameli:
> > Reversion of my changes to AudioSegmentSampleView is questionable
> 
> I missed that commit on 2-of-6. I looked at it again, but couldn't figure out the reason behind it ? For the spectrum analyzer it seems simpler to me to be returned a `sampleCount` and then check whether this exceeded the maximum allowed number of samples.

Who said things should be as simple as possible, but not simpler?

This is a case of "not simpler."

src/FreqWindow.h
Paul-Licameli: I'm not persuaded by this either.

After a closer look, I think this:
- Only a local variable in `Recalc` is necessary.
- `GetAudio` can return the value instead of changing a member variable.
- `Recalc` can call `GetAudio`.  Both existing calls to `GetAudio` can be eliminated with this change.
- `DrawPlot` and `PlotPaint` only use the number `GetNumSamples(mChannelGroupSampleViews)` so instead, just store that number as a member variable each time `Recalc` is called.

src/FreqWindow.cpp
```diff
@@ -182,6 +183,36 @@ BEGIN_EVENT_TABLE(FrequencyPlotDialog, wxDialogWrapper)
    EVT_COMMAND(wxID_ANY, EVT_FREQWINDOW_RECALC, FrequencyPlotDialog::OnRecalc)
 END_EVENT_TABLE()
 
+// Permit approximately 46.60 minutes of selected samples at
```
Paul-Licameli, line 186: Is this calculation right?  I think it’s 23.3 minutes

src/FreqWindow.cpp
```diff
@@ -581,63 +611,35 @@ bool FrequencyPlotDialog::Show(bool show)
...
-         mData.reset();
-         mDataLen = 0;
+         mChannelGroupSampleViews.clear();
          return;
       }
-      Floats buffer1{ mDataLen };
-      Floats buffer2{ mDataLen };
-      float *const buffers[]{ buffer1.get(), buffer2.get() };
       // Don't allow throw for bad reads
-      track->GetFloats(0, nChannels, buffers, start, mDataLen,
```
Paul-Licameli, line 622: Do you preserve non-throwing?

src/FreqWindow.cpp
```diff
@@ -581,63 +611,35 @@ bool FrequencyPlotDialog::Show(bool show)
...
-      for (size_t i = 0; i < mDataLen; i++)
-         mData[i] = buffers[0][i];
-      for (size_t iChannel = 1; iChannel < nChannels; ++iChannel) {
-         const auto buffer = buffers[iChannel];
-         for (size_t i = 0; i < mDataLen; i++)
-            mData[i] += buffer[i];
-      }
+      constexpr auto mayThrow = false;
+      mChannelGroupSampleViews.push_back(track->GetSampleView(
+         selectedRegion.t0(), selectedRegion.t1(), mayThrow));
```
Paul-Licameli, line 635: Ok yes

src/FreqWindow.cpp
```diff
@@ -581,63 +611,35 @@ bool FrequencyPlotDialog::Show(bool show)
...
       ++selcount;
    }
 
-   if (selcount == 0)
-      return;
```
Paul-Licameli, line 635: That should not be deleted I think

src/FreqWindow.h
Paul-Licameli: I think you are correct that you make the lifetime of sample data just the same as in the previous version.

Which means, the previous version didn’t handle memory as efficiently as it might have and so we can improve on it now.

With closer study you can find that the “analyst” object has all the derived data necessary to draw the plot, and the size of that is proportional only to the analysis window size (or number of frequency bins) and independent of the length of the selection, which could be large.

src/FreqWindow.cpp
```diff
@@ -1034,9 +1062,11 @@ void FrequencyPlotDialog::Recalc()
       if (IsShown())
          blocker.emplace(this);
       wxYieldIfNeeded();
-
-      mAnalyst->Calculate(alg, windowFunc, mWindowSize, mRate,
-         mData.get(), mDataLen,
+      Floats tempFloatsArray;
+      tempFloatsArray.reinit(numSamples);
```
Paul-Licameli, line 1066: Even numbers that do fit in a size_t can be excessive and run out of memory.  There is still a bound of 64M samples and multiply that by four to get bytes.  Will there always be a quarter gig free?

So maybe a try-catch of `bad_alloc` could happen around the allocation, like what I lately did to bullet-proof the Mixer Board.

src/FreqWindow.cpp
```diff
@@ -1013,9 +1018,32 @@ void FrequencyPlotDialog::SendRecalcEvent()
...
+   for (const auto& track : tracks)
+      for (const auto& channel : track)
+         for (const auto& segment : channel)
+         {
+            const auto toRead = limitSampleBufferSize(
+               bufferSize - numWritten, segment.GetSampleCount());
+            segment.Copy(buffer + numWritten, toRead);
+            numWritten += toRead;
+            if (numWritten == bufferSize)
+               return;
```
Paul-Licameli, line 1037: This bothers me, for the same reasons that Mixer Board changes did, and why I cautioned about correct RMS calculation in Contrast.cpp.

Samples from neighboring clips with different stretches are treated as just one long sequence of samples as if they had the same real-time rate, ignoring stretches.  Is that right?

For Contrast I argued it was wrong:  rms should be calculated for clips separately, then combined into an overall RMS with the proper weighting of different clips, taking stretches into account at that point.

For Mixer Board, it was all right if only peak values are needed, but as I have also mentioned, Mixer Board also does an rms calculation that can be wrong now, but fixing it is future work that needs a totally different approach anyway to get the true post-fader levels out of the audio engine.

What wrong calculation could be happening here? …

Paul-Licameli: … answering that would take you into the guts of SpectrumAnalyst, where as I recall, values derived from many different FFT windows are combined into a single window that is later plotted.

As we have lately discussed, it is not wrong to take FFT of pre-stretched data and assign it to the same frequency bins as for the as-rendered sound that we don’t bother to calculate just yet.  However…

I think you would find there an assumption that all windows should be weighted equally in the combined result, but that could now be wrong with stretches.

What is a test case that might demonstrate this? …

Paul-Licameli:
Here is a test case:  Make neighboring clips of equal duration.  Generate a tone in one and a tone of another frequency in the other.  No stretches yet.  Select all and plot spectrum.  See two peaks.  (Not two spikes, because of the realities of spectral leakage, unless you use rectangular windows and choose frequencies as exact multiple levels of the inverse of the sample window duration.  You might also like to choose linear, not log frequency scale.) …

Paul-Licameli:
… if that is the control, the experimental would be two clips that sound the same as the control as played, but have unequal stretches.  For instance make the first clip half as long as in the control but stretch it to 2x duration, with the second clip as before.

Then plot spectrum.  What would you see?

I think you should see nearly the same plot, but I predict you will see unequal heights.  A clip stretched more will be under-weighted and another stretched less or compressed, over-.

Paul-Licameli:
So maybe the bug allows some data to be plotted but it’s bogus.  What should be done?

- ignore this as an unlikely use case?
- Detect clip stretch mismatch and display an error rather than bogus data?
- Do the hard work of more proper weighting of values?

If the last, maybe the way to do it is make a spectrum analyst for each clip separately, then write something that makes the correct weighted sums of several analyst results as one analyst object to be used later in plotting.

Paul-Licameli:
Yet another possibility:  downgrade the priority of this bug, if in fact, present master doesn’t freeze but only fails to plot anything.  Then kick the can down to a later release.

Or just detect failure of `GetFloats` and give the user an error dialog explaining the problem, instead of the quiet failure to show any data.

src/FreqWindow.h
saintmatthieu: I see three stages that matter to our decision here:
1. Selection sample read-out (`GetAudio`)
2. Calculation of spectrum (`Recalc`)
3. Display of result (`Show`)

`GetAudio` itself calls `Recalc`, but `Recalc` can also be called independently, if there is a change of analysis parameter (window size, spectrum type, ...). The user may keep the analysis window open and change the track data, which is a nice user experience. So the time-domain audio must be kept in memory for the lifetime of the window, we cannot assume that it won't change and we can re-read it.

src/FreqWindow.cpp
saintmatthieu: It is correct. Unusual is the left operand of the bit-shift being 2, it usually is 1 - which would then yield 23.3.
Changed to `1 << 27`, in case I make another push.

dozzzzer:
>  downgrade the priority of this bug

I believe P4 is already a pretty fair priority for it, but I definitely agree that it can be kicked down to the later release

src/FreqWindow.cpp
saintmatthieu: ### TL, DR:
From a result point of view, IMO it very marginally matters whether we pre-render or not, because of how rare use cases are where the limitation of either approach becomes a problem. (I assume that we don't put in the effort of weighting according to stretch ratio.) Implementation wise, I think this one is easier, isn't it? No need to make an editable copy of the track objects, just use a different, user-friendly API (`GetSampleView`).

### Detail:
Yes, the weighting is ignored here, which might yield unexpected results if the selection is made across clips of different stretch ratios. I don't expect this to happen often, and if it does, that the stretching difference is large enough to be noticeable, but in principle, yes, either accurate, preliminary stretch-rendering, or weighting would alleviate this.

I said "accurate" because, as we know already, the stretching algorithm is not perfect. I've done an experiment, which should serve as material for the spectrum analysis as well as the noise reduction problem.

1. I recorded myself saying "test" with an open mic.
2. I stretched the clip moderately, squeezing it by approx. 25%.
3. I duplicated the track for better comparison.
4. I selected a short region of background noise,
5. I rendered the stretching of that region.

The project overall:
BackgroundNoise project

The spectrum analysis of the unrendered labeled segment (upper track):
Background noise spectrum

That of the rendered segment (lower track):
Background noise spectrum - rendered

(A way of superposing the spectra would be nice, like in Adobe Audition. We don't have this yet, do we?)

For the spectrum analysis use case, I don't expect this to happen often either. It's more for the noise reduction case that this might become an issue, but I'll elaborate in that other PR.

saintmatthieu:
> … if that is the control, the experimental would be two clips that sound the same as the control as played, but have unequal stretches. For instance make the first clip half as long as in the control but stretch it to 2x duration, with the second clip as before.
> 
> Then plot spectrum. What would you see?
> 
> I think you should see nearly the same plot, but I predict you will see unequal heights. A clip stretched more will be under-weighted and another stretched less or compressed, over-.

Yes, in that case, the result would be obviously unexpected.

saintmatthieu:
Now I read all your comments till the last.

I say we merge this. It's imperfect, but at least it will work if you change project tempo and analyse over several clips. Then let's see if complains come our way.

Paul-Licameli:
> Now I read all your comments till the last.
> 
> I say we merge this. It's imperfect, but at least it will work if you change project tempo and analyse over several clips. Then let's see if complains come our way.

I do not approve this for merge.

src/FreqWindow.cpp
saintmatthieu: I don't think so. In either versions, before or after, the only thing it does is returning before the warning check that there aren't too many samples. But if `selcount == 0`, it means that there was no selected track in the project, and no sample was read, in which case, of course, no warning anyways.

Paul-Licameli:
This bug has low priority.  It is not a crash or freeze bug.

Rebasing 4th of 6 into the effect changes or refactoring tasks are far more important.

src/FreqWindow.cpp
saintmatthieu: Yep. `Recalc` then also must return a boolean, because the follow-up code that shows the result otherwise crashes. That's not hard, commit upcoming.

saintmatthieu:
I regret that. The work is now done and it does bring an improvement. Splitting hairs was more time consuming - although leading to interesting discussions.

Paul-Licameli:
> I regret that. The work is now done and it does bring an improvement. Splitting hairs was more time consuming - although leading to interesting discussions.

Reasoning about the correctness of the changes you propose is my duty.  I did it.  I'm not satisfied.

src/MixerBoard.cpp
saintmatthieu: I don't see how it was safer before. I'm looking at this:
```cpp
   // Not sure how to prove satisfaction of the invariant of GetSampleView
   mSampleView = pTrack->GetSampleView(t0, t1, mayThrow);

   // Expect that the difference of t1 and t0 is the part of a track played
   // in about 1/20 second (ticks of TrackPanel timer), so this won't overflow,
   // unless stretch ratio is extremely low.
   const auto nFrames = GetNumSamplesInView(mSampleView[0]);
```

If `AudioSegmentSampleView::mLength` is `size_t`,  overflow could happen in the first statement.
If it takes a `sampleCount`, overflow could happen in  `GetNumSamplesInView`.
Either way the overflow is caught by your try-catch.

Also, this would only happen in extreme cases : the MixerBoard refreshes every 50ms. What stretch ratio do we need to squeeze more than 2^31 samples of audio in such a short amount of time ? Possible, yet again there's the try-catch.

With `sampleCount`, the risk of overflow doesn't matter for this check in the next commit:
```cpp
   if (GetNumSamples(mChannelGroupSampleViews) > maxDataLen) {
      auto msg = XO(
"Too much audio was selected. Only the first %.1f seconds of audio will be analyzed.")
         .Format(maxDataLen / mRate);
      AudacityMessageBox( msg );
   }
```

But again, I didn't realize you had changed it, and didn't revert it properly as I left your comments untouched.
We have a disagreement about whether this PR should go in or not, let's wait until that's settled to see if I have to clean that up.

petersampsonaudacity:
Testing on W10 with @saintmatthieu 's branch build for this:  audacity-win-3.4.0-alpha-20230908+519eafe-x64

When I now try to run Plot Spectrum on a stretched clip I get this error message.
image

I find that message very opaque and unhelpful to the user.  Now I figured out that if I rendered the stretched track then I would be able to plot the spectrum - but that's _**only**_ as a result of the discussions I had with Matthieu on Thursday.  

Maybe the error message could suggest that the user renders the track and then retries.

src/FreqWindow.cpp
```diff
@@ -608,19 +610,30 @@ void FrequencyPlotDialog::GetAudio()
...
       // Don't allow throw for bad reads
-      track->GetFloats(0, nChannels, buffers, start, mDataLen,
-         false, FillFormat::fillZero, false);
+      if (!track->GetFloats(
+             0, nChannels, buffers, start, mDataLen, false,
+             FillFormat::fillZero, false))
+      {
+         using namespace BasicUI;
+         ShowMessageBox(
+            XO("Could not read for analysis. Selecting time-stretched audio is a known reason."),
```
LWinterberg, line 631: "Audio could not be analyzed. This may be due to a stretched clip. Try resetting any stretched clips, or mixing and rendering the tracks before analyzing." might be better

saintmatthieu: That makes me think that we don't have an straightforward way of unstretching clips ... Ticket ?

Paul-Licameli: You mean restring the stretch ratio exactly to 1.

src/FreqWindow.cpp
```diff
@@ -608,19 +610,30 @@ void FrequencyPlotDialog::GetAudio()
       }
       const auto nChannels = track->NChannels();
       if (track->GetRate() != mRate) {
-         AudacityMessageBox(
-            XO(
-"To plot the spectrum, all selected tracks must be the same sample rate.") );
+         using namespace BasicUI;
+         ShowMessageBox(
+            XO("To plot the spectrum, all selected tracks must be the same sample rate."),
```
Paul-Licameli, line 615: I think this other old message should be corrected to "must have"

Or maybe totally different words

@LWinterberg

Paul-Licameli (approved):

src/FreqWindow.cpp
```diff
@@ -608,19 +610,30 @@ void FrequencyPlotDialog::GetAudio()
...
       // Don't allow throw for bad reads
-      track->GetFloats(0, nChannels, buffers, start, mDataLen,
-         false, FillFormat::fillZero, false);
+      if (!track->GetFloats(
+             0, nChannels, buffers, start, mDataLen, false,
+             FillFormat::fillZero, false))
+      {
+         using namespace BasicUI;
+         ShowMessageBox(
+            XO("Audio could not be analyzed. This may be due to a stretched clip.\nTry resetting any stretched clips, or mixing and rendering the tracks before analyzing"),
```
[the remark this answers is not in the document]
Paul-Licameli: Or should I say "shrunk" to avoid other confusions!

src/FreqWindow.cpp
LWinterberg: "must have" sounds good to me.

src/FreqWindow.cpp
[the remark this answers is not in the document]
saintmatthieu: It is a problem at the basis indeed, "stretch" meaning "expand" in non-technical context. Not sure how to solve this :D

src/FreqWindow.cpp
saintmatthieu: Addressed.

src/FreqWindow.h
```diff
@@ -67,7 +67,7 @@ class FrequencyPlotDialog final : public wxDialogWrapper,
 private:
    void Populate();
 
-   void GetAudio();
+   bool GetAudio();
```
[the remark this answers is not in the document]
saintmatthieu: I suppose, yes: this PR doesn't change the execution sequence for the case where sequences have different sample rates. That's been working well and failure due to failing `GetFloats` take the same path.

saintmatthieu:
Note that this PR would benefit from the [Move a function to new file WaveTrackUtilities](https://github.com/audacity/audacity/pull/5138/commits/b066650f6effa6fedd195564f32417fccdb72191) commit in #5138 : then we'd know for sure whether it's because of stretched data or not.

dozzzzer:
Tested. On macOS, after a proper error message is displayed, a blank **Frequency Analysis** dialog is displayed:

Paul-Licameli (approved):

src/FreqWindow.cpp
```diff
@@ -1239,7 +1239,8 @@ void OnPlotSpectrum(const CommandContext &context)
 
    if( VetoDialogHook::Call( freqWindow ) )
       return;
-   freqWindow->Show(true);
+   if (!freqWindow->Show(true))
+       return;
    freqWindow->Raise();
```
Paul-Licameli, line 1244: But does this only appear to fix it?  Is the un-raised window just hiding behind the project window which you can move aside?

saintmatthieu: Good thinking, but I tested it and it seems like it really fixes it. Could @chinakov confirm ?

chinakov: Yes, the issue has been fixed.

chinakov:
Tested

PR #5165, opened 2023-09-07
Wave track affordance rewrites

vsverchinsky:
Resolves: #5076 

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
@@ -1705,7 +1705,11 @@ std::weak_ptr<WaveClip> WaveChannelView::GetSelectedClip()
 {
    if (auto affordance = std::dynamic_pointer_cast<WaveTrackAffordanceControls>(GetAffordanceControls()))
    {
-      return affordance->GetSelectedClip();
+      assert(GetChannelIndex() == 0);
+      if(auto interval = *affordance->GetSelectedInterval())
+      {
+         return interval->GetClip(0);
+      }
    }
    return {};
 }
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
@@ -12,10 +12,12 @@
 
 #include <wx/font.h>
 
+#include "Channel.h"
 #include "Observer.h"
 #include "ViewInfo.h"
-#include "../../../ui/CommonTrackPanelCell.h"
-#include "../../../ui/TextEditHelper.h"
+#include "WaveTrack.h"
+#include "tracks/ui/CommonTrackPanelCell.h"
+#include "tracks/ui/TextEditHelper.h"
 
 
 struct TrackListEvent;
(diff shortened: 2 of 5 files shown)


src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.h
```diff
@@ -36,20 +38,22 @@ class AUDACITY_DLL_API WaveTrackAffordanceControls :
     public TextEditDelegate,
     public std::enable_shared_from_this<WaveTrackAffordanceControls>
 {
-    std::weak_ptr<WaveClip> mFocusClip;
+   using IntervalIterator = ChannelGroup::IntervalIterator<WaveTrack::Interval>;
```
Paul-Licameli, line 41: Indentation

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
```diff
@@ -1705,7 +1705,10 @@ std::weak_ptr<WaveClip> WaveChannelView::GetSelectedClip()
 {
    if (auto affordance = std::dynamic_pointer_cast<WaveTrackAffordanceControls>(GetAffordanceControls()))
    {
-      return affordance->GetSelectedClip();
+      if(auto interval = *affordance->GetSelectedInterval())
+      {
+         return interval->GetClip(GetChannelIndex());
```
Paul-Licameli, line 1710: I want only constant `0` as an argument of `GetClip` and later I will remove the argument.

But the outer `if` should justify `assert(GetChannelIndex() == 0);`

So write that assertion, then change the argument of `GetClip`.

Actually `GetChannelIndex()` still always returns 0, unlike `ReallyGetChannelIndex`

The problem of unifying those two functions (keeping the name of the first, and the behavior of the second) is another box on my checklist for later.

Paul-Licameli:
I reviewed the rest of it, and have nothing more to comment about, except this:

You store `ChannelIterator` objects, and those have plain C-pointers to `GroupType` in them.  But I don't think there is anything in the contracts of these classes giving good enough lifetime guarantees against that becoming a dangling pointer.

I think nothing wrong will happen and I have no specific rewrite to insist that you do, but it bothers my mind a little.

Paul-Licameli:
So, consider this round of review complete, with just one easy request for changes.

vsverchinsky:
> I reviewed the rest of it, and have nothing more to comment about, except this:
> 
> You store `ChannelIterator` objects, and those have plain C-pointers to `GroupType` in them. But I don't think there is anything in the contracts of these classes giving good enough lifetime guarantees against that becoming a dangling pointer.
> 
> I think nothing wrong will happen and I have no specific rewrite to insist that you do, but it bothers my mind a little.

Alternatively I could use indices

Paul-Licameli:
> > I reviewed the rest of it, and have nothing more to comment about, except this:
> > You store `ChannelIterator` objects, and those have plain C-pointers to `GroupType` in them. But I don't think there is anything in the contracts of these classes giving good enough lifetime guarantees against that becoming a dangling pointer.
> > I think nothing wrong will happen and I have no specific rewrite to insist that you do, but it bothers my mind a little.
> 
> Alternatively I could use indices

Maybe, but I will approve it now.

Paul-Licameli (approved):
I approve, conditional on successful CI

chinakov:
Tested

PR #5175, opened 2023-09-09
Deterministic zero-fill whenever WaveTrack::Get[Floats] fails...

Paul-Licameli:
... This simple fix will make plot spectrum show blank instead of garbage for a selection with any stretch.

Make zero-fill on failure part of the WideSampleSequence contract too, removing a redundant defense in MixerSource.

Unproblematic un-checked uses are in:
- effects (where there will be pre-rendering of the selection)
- WideSampleSource (using new promise of WideSampleSequence; and anyway where we assume an upstream stretching sequence decorator)
- NearestZeroCrossing() (there are tests before it is called)

Problematic uses:

- [x] WaveTrack::ReverseOneClip -- won't generally work, but it happens to be called only from an effect now ; fixed by https://github.com/audacity/audacity/pull/5138

- [x] [Find clipping](https://github.com/audacity/audacity/issues/5176) (in src/effects but an analyzer, not an effect)
- [x] Punch and Roll Recording (only to find crossfade data)
- [x] [Snapping to frequency peaks](https://github.com/audacity/audacity/issues/5179) (drag of spectral box centerline, and "Next Higher/Lower Peak Frequency" menu commands; see SelectHandle)
- [x] Compare Audio macro command
- [x] SpectrumTransformer as used by experimental brush tool (ok in Noise Reduction)

QA:  Try the things enumerated above.  They may not work, but behave differently, always using zeroes instead of unpreditable data.

Issues:

- [x] Find Clipping does not generate labels in the Clipping label track on stretched clips.
- [x] Reverse audio on a stretched clip hangs Audacity.

Recommended:

### DIFF
--- a/libraries/lib-mixer/MixerSource.cpp
+++ b/libraries/lib-mixer/MixerSource.cpp
@@ -130,9 +130,11 @@ size_t MixerSource::MixVariableRates(
             constexpr auto iChannel = 0u;
             if (!mpLeader->GetFloats(
                    iChannel, nChannels, dst.data(), pos, getLen, backwards,
-                   FillFormat::fillZero, mMayThrow))
-               for (size_t iChannel = 0; iChannel < nChannels; ++iChannel)
-                  memset(dst[i], 0, sizeof(float) * getLen);
+                   FillFormat::fillZero, mMayThrow)) {
+               // Now redundant in case of failure
+               // for (size_t iChannel = 0; iChannel < nChannels; ++iChannel)
+                  // memset(dst[i], 0, sizeof(float) * getLen);
+            }
             mpLeader->GetEnvelopeValues(
                mEnvValues.data(), getLen, (pos).as_double() / sequenceRate,
                backwards);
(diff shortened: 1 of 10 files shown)


crsib (commented):
I am really surprised to see this PR. We have agreed that zeroing out is **not a fix**. I don't like that the buffer is modified on failure either.

Paul-Licameli:
> I am really surprised to see this PR. We have agreed that zeroing out is **not a fix**. I don't like that the buffer is modified on failure either.

As I explained at standup, I do not intend this to substitute for other fixes in the Plot Spectrum dialog.  But it affects other things besides plot spectrum.  It replaces failures with unspecified date with other failures with predictable zero data.  Deterministic behavior is not best but better.

Please note all the details at top about what is affected.

saintmatthieu:
Any failing use of `GetFloats` we know of that QA could use for testing?

saintmatthieu (approved):

chinakov:
Using Find Clipping (Analyze -> Find Clipping) on a stretched clip results in no labels being placed in the 'Clipping' Label Track. 

Screenshot 2023-09-25 at 14 34 03

Paul-Licameli:
Rebased onto recent master and resolved a conflict.

Now that other prior bug fixes are in the branch, I predict the freezing problems will go away.

chinakov:
Tested, thanks!

PR #5182, opened 2023-09-11
Fix silence audio command

saintmatthieu:

Bug description: https://github.com/audacity/audacity/pull/5138#issuecomment-1713591482

Recommended:

QA:
- [x] "Silence Audio" command, across tracks and clips both stretched and unstretched. New clip boundaries should appear at an edge of the selection only if the clip underneath is stretched.
- [x] `Edit > Labeled Audio > Silence Audio` also works. (Note, though, that there isn't a progress bar for this yet. It's a bit trickier to introduce there, but #5043 has a utility that we will reuse for this once the utility is merged.)

### DIFF
--- a/libraries/lib-time-track/TimeTrack.cpp
+++ b/libraries/lib-time-track/TimeTrack.cpp
@@ -222,7 +222,8 @@ void TimeTrack::Paste(double t, const Track &src)
       (void)0;// intentionally do nothing.
 }
 
-void TimeTrack::Silence(double WXUNUSED(t0), double WXUNUSED(t1))
+void TimeTrack::Silence(
+   double WXUNUSED(t0), double WXUNUSED(t1), ProgressReporter)
 {
    assert(IsLeader());
 }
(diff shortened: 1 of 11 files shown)


libraries/lib-time-track/TimeTrack.cpp
```diff
@@ -222,7 +222,8 @@ void TimeTrack::Paste(double t, const Track &src)
       (void)0;// intentionally do nothing.
 }
 
-void TimeTrack::Silence(double WXUNUSED(t0), double WXUNUSED(t1))
+void TimeTrack::Silence(
+   double WXUNUSED(t0), double WXUNUSED(t1), ProgressReporter)
```
Paul-Licameli, line 226: I repeat, you can clean up the old `WXUNUSED` macros

Just `double, double,` or you can use C++17 `[[maybe_unused]]`

Paul-Licameli:
I think you can shrink the branch so this PR does not include all of #5138.  Only the first and last commits of the sequence are needed.  That will allow it to be tester and maybe merged sooner.

saintmatthieu:
It also needs the `ProgressReporter` of "Progress bar for pre-effect stretching", which depends on "Repair effects and generators", ...
I'd rather wait than putting the extra work. This fix shouldn't be blocking anything other than the release itself, and we're not there yet.

Paul-Licameli (approved):

dozzzzer:
Tested

Issue #5180, opened 2023-09-10
A new API to join, split, and swap channels


Paul-Licameli:

dozzzzer:
@Paul-Licameli 
- importing a [5.1 .dtshd file](https://drive.google.com/file/d/1_n97s8snSUlogo1dhbIauKj6JHwuVJBZ/view?usp=drive_link) results in a crash

vsverchinsky:
> @Paul-Licameli
> 
>     * importing a [5.1 .dtshd file](https://drive.google.com/file/d/1_n97s8snSUlogo1dhbIauKj6JHwuVJBZ/view?usp=drive_link) results in a crash

@dozzzzer Is it happening in [this](https://github.com/audacity/audacity/pull/5193) branch?

dozzzzer:
> > @Paul-Licameli
> > ```
> > * importing a [5.1 .dtshd file](https://drive.google.com/file/d/1_n97s8snSUlogo1dhbIauKj6JHwuVJBZ/view?usp=drive_link) results in a crash
> > ```
> 
> @dozzzzer Is it happening in [this](https://github.com/audacity/audacity/pull/5193) branch?

No, in this branch there's no crash.

dozzzzer:
@vsverchinsky

### Cannot record to an empty stereo track
 
1. Add a stereo track
2. Set recording channels to 2
3. Click the Record button
4. Observe

Windows:
<img width=350 src=https://github.com/audacity/audacity/assets/70063970/26d77609-1397-4a19-bf7f-cf60f34175cb>

macOS:

### Resuming recording after pause results in an error or misaligned channels

1. Set recording channels to 2
2. Click the Record button
3. Click the Stop button
4. Click the Record button
5. Observe

Windows:
<img width=350 src=https://github.com/audacity/audacity/assets/70063970/26d77609-1397-4a19-bf7f-cf60f34175cb>

macOS:



### Attempt to import .ogg results in a crash

1. File > Import
7. Select .ogg file
8. Observe

PR #5193, opened 2023-09-13
Channels API extend

vsverchinsky:
Resolves: #5180

Recommended:

### DIFF
--- a/libraries/lib-effects/MixAndRender.cpp
+++ b/libraries/lib-effects/MixAndRender.cpp
@@ -95,34 +95,15 @@ TrackListHolder MixAndRender(const TrackIterRange<const WaveTrack> &trackRange,
       oneinput = true;
    // only one input track (either 1 mono or one linked stereo pair)
 
-   // EmptyCopy carries over any interesting channel group information
-   // Maybe first is stereo, or it is mono, but some other input is stereo;
-   // therefore don't use WideEmptyCopy, but make and link a right channel as
-   // needed
-   // And reset pan and gain
-   auto mixLeft =
-      first->EmptyCopy(trackFactory->GetSampleBlockFactory(), false);
+   auto result = trackFactory->Create(mono ? 1 : 2, *first);
+   auto mix = static_cast<WaveTrack*>(*result->begin());
+   mix->SetRate(rate);
+   mix->ConvertToSampleFormat(format);
+   if(!oneinput)
+      mix->SetName(newTrackName);
+   mix->MoveTo(mixStartTime);
 
-   // TODO: more-than-two-channels
-
-   decltype(mixLeft) mixRight{};
-   if (!mono)
-      mixRight = trackFactory->Create(format, rate);
-
-   auto result = TrackList::Temporary(nullptr, mixLeft, mixRight);
-
-   mixLeft->SetPan(0);
-   mixLeft->SetGain(1);
-   mixLeft->SetRate(rate);
-   mixLeft->ConvertToSampleFormat(format);
-   if (oneinput)
-      mixLeft->SetName(first->GetName()); /* set name of output track to be the same as the sole input track */
-   else
-      /* i18n-hint: noun, means a track, made by mixing other tracks */
-      mixLeft->SetName(newTrackName);
-   mixLeft->MoveTo(mixStartTime);
-
-   auto maxBlockLen = mixLeft->GetIdealBlockSize();
+   auto maxBlockLen = mix->GetIdealBlockSize();
 
    // If the caller didn't specify a time range, use the whole range in which
    // any input track had clips in it.
@@ -150,23 +131,17 @@ TrackListHolder MixAndRender(const TrackIterRange<const WaveTrack> &trackRange,
          if (blockLen == 0)
             break;
 
-         if (mono) {
-            auto buffer = mixer.GetBuffer();
-            mixLeft->Append(buffer, format, blockLen, 1, effectiveFormat);
-         }
-         else {
-            auto buffer = mixer.GetBuffer(0);
-            mixLeft->Append(buffer, format, blockLen, 1, effectiveFormat);
-            buffer = mixer.GetBuffer(1);
-            mixRight->Append(buffer, format, blockLen, 1, effectiveFormat);
+         for(auto channel : mix->Channels())
+         {
+            auto buffer = mixer.GetBuffer(channel->ReallyGetChannelIndex());
+            channel->AppendBuffer(buffer, format, blockLen, 1, effectiveFormat);
          }
 
          updateResult = pProgress->Poll(
             mixer.MixGetCurrentTime() - startTime, endTime - startTime);
       }
    }
-
-   mixLeft->Flush();
+   mix->Flush();
    if (updateResult == ProgressResult::Cancelled ||
        updateResult == ProgressResult::Failed)
       return {};
@@ -183,10 +158,7 @@ TrackListHolder MixAndRender(const TrackIterRange<const WaveTrack> &trackRange,
    wxPrintf("Elapsed time: %f sec\n", elapsedTime);
    wxPrintf("Max number of tracks to mix in real time: %f\n", maxTracks);
 #endif
-
-      for (auto pTrack : { mixLeft, mixRight })
-         if (pTrack)
-            RealtimeEffectList::Get(*pTrack).Clear();
+      RealtimeEffectList::Get(*mix).Clear();
    }
 
    return result;
--- a/libraries/lib-import-export/ImportUtils.h
+++ b/libraries/lib-import-export/ImportUtils.h
@@ -14,14 +14,18 @@
 
 #include <memory>
 #include <vector>
+
+#include "Import.h"
 #include "SampleFormat.h"
 #include "Internat.h"
+#include "Track.h"
 
 class wxString;
 
 class TrackList;
 class WaveTrackFactory;
 class WaveTrack;
+class WaveChannel;
 
 class IMPORT_EXPORT_API ImportUtils final
 {
@@ -30,15 +34,22 @@ public:
    //! Choose appropriate format, which will not be narrower than the specified one
    static sampleFormat ChooseFormat(sampleFormat effectiveFormat);
    
-   //! Build a wave track with appropriate format, which will not be narrower than the specified one
-   static std::shared_ptr<WaveTrack> NewWaveTrack( WaveTrackFactory &trackFactory,
+   //! Builds a wave track and places it into a track list.
+   //! The format will not be narrower than the specified one.
+   static TrackListHolder NewWaveTrack(WaveTrackFactory &trackFactory, unsigned nChannels,
       sampleFormat effectiveFormat, double rate);
    
    static void ShowMessageBox(const TranslatableString& message, const TranslatableString& caption = XO("Import Project"));
-   
-   using NewChannelGroup = std::vector<std::shared_ptr<WaveTrack>>;
 
-   //! Flush the given channels and group them into tracks
+   //! Iterates over channels in each wave track from the list
    static
-   std::shared_ptr<TrackList> MakeTracks(const NewChannelGroup &channels);
+   void ForEachChannel(TrackList& trackList, const std::function<void(WaveChannel&)>& op);
+
+   //! Flushes the given channels and moves them to \p outTracks
+   static
+   void FinalizeImport(TrackHolders& outTracks, const std::vector<TrackListHolder>& importedStreams);
+
+   //! Flushes the given channels and moves them to \p outTracks
+   static
+   void FinalizeImport(TrackHolders& outTracks, TrackListHolder trackList);
 };
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -681,14 +681,15 @@ TrackListHolder TrackList::ReplaceOne(Track &t, TrackList &&with)
    return result;
 }
 
-void TrackList::UnlinkChannels(Track& track)
+std::vector<Track*> TrackList::UnlinkChannels(Track& track)
 {
    auto list = track.mList.lock();
    if (list.get() == this)
    {
       auto channels = TrackList::Channels(&track);
       for (auto c : channels)
-          c->SetLinkType(Track::LinkType::None);
+         c->SetLinkType(Track::LinkType::None);
+      return { channels.begin(), channels.end() };
    }
    else
       THROW_INCONSISTENCY_EXCEPTION;
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -1026,12 +1026,6 @@ TrackList::RegisterPendingChangedTrack(Updater updater, Track *src)
    return result;
 }
 
-void TrackList::RegisterPendingNewTrack( const std::shared_ptr<Track> &pTrack )
-{
-   Add<Track>( pTrack );
-   pTrack->SetId( TrackId{} );
-}
-
 void TrackList::UpdatePendingTracks()
 {
    if (!mPendingUpdates)
@@ -1458,6 +1452,16 @@ void TrackList::Append(TrackList &&list)
       auto pTrack = *iter;
       iter = list.erase(iter);
       this->Add(pTrack);
+   }
+}
+
+void TrackList::RegisterPendingNewTracks(TrackList&& list)
+{
+   for(auto it = list.ListOfTracks::begin(); it != list.ListOfTracks::end();)
+   {
+      Add(*it);
+      (*it)->SetId({});
+      it = list.erase(it);
    }
 }
 
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -987,19 +987,16 @@ double TrackList::GetEndTime() const
       std::numeric_limits<double>::lowest(), std::max);
 }
 
-std::vector<Track*>
+Track*
 TrackList::RegisterPendingChangedTrack(Updater updater, Track *src)
 {
    // This is only done on the TrackList belonging to a project
    assert(GetOwner()); // which implies mPendingUpdates is not null
    assert(src->IsLeader());
-   TrackListHolder tracks;
-   std::vector<Track*> result;
-   if (src) {
-      tracks = src->Clone(); // not duplicate
-      assert(src->NChannels() == tracks->NChannels());
-   }
-   if (src) {
+   
+   auto tracks = src->Clone(); // not duplicate
+   assert(src->NChannels() == tracks->NChannels());
+   {
       // Share the satellites with the original, though they do not point back
       // to the pending track
       const auto channels = TrackList::Channels(src);
@@ -1008,21 +1005,18 @@ TrackList::RegisterPendingChangedTrack(Updater updater, Track *src)
          ((AttachedTrackObjects&)**iter++) = *pChannel; // shallow copy
    }
 
-   if (tracks) {
-      mUpdaters.push_back(updater);
-      auto iter = tracks->ListOfTracks::begin(),
-         end = tracks->ListOfTracks::end();
-      while (iter != end) {
-         auto pTrack = *iter;
-         iter = tracks->erase(iter);
-         mPendingUpdates->ListOfTracks::push_back(pTrack->SharedPointer());
-         auto n = mPendingUpdates->ListOfTracks::end();
-         --n;
-         pTrack->SetOwner(shared_from_this(), {n, &*mPendingUpdates});
-         result.push_back(pTrack.get());
-      }
+   const auto result = *tracks->begin();
+   mUpdaters.push_back(updater);
+   auto iter = tracks->ListOfTracks::begin(),
+      end = tracks->ListOfTracks::end();
+   while (iter != end) {
+      auto pTrack = *iter;
+      iter = tracks->erase(iter);
+      mPendingUpdates->ListOfTracks::push_back(pTrack->SharedPointer());
+      auto n = mPendingUpdates->ListOfTracks::end();
+      --n;
+      pTrack->SetOwner(shared_from_this(), {n, &*mPendingUpdates});
    }
-
    return result;
 }
 
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -541,23 +541,23 @@ auto TrackList::Find(Track *pTrack) -> TrackIter<Track>
    return iter.Filter( &Track::IsLeader );
 }
 
-bool TrackList::SwapChannels(Track &track)
+Track* TrackList::SwapChannels(Track &track)
 {
    if (!track.HasLinkedTrack())
-      return false;
+      return nullptr;
    auto pOwner = track.GetOwner();
    if (!pOwner)
-      return false;
+      return nullptr;
    auto pPartner = pOwner->GetNext(&track, false);
    if (!pPartner)
-      return false;
+      return nullptr;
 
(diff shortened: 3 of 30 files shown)


libraries/lib-track/Track.cpp
```diff
@@ -1471,6 +1465,16 @@ void TrackList::Append(TrackList &&list)
    }
 }
 
+void TrackList::RegisterPendingNewTracks(TrackList&& list)
```
Paul-Licameli, line 1468: I might say "channels" not "tracks" in this name

src/ProjectAudioManager.cpp
```diff
@@ -944,28 +942,28 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
...
-            transportSequences.captureSequences.push_back(newTrack);
+            for(auto channel : newTrack->Channels())
+            {
+               ChannelView::Get(*channel).SetMinimized(minimizeChannelView);
 
-            tempList.reset();
-            trackList.RegisterPendingNewTrack(newTrack);
+               transportSequences.captureSequences.push_back(
+                  std::static_pointer_cast<WaveTrack>(channel->GetTrack().shared_from_this())
+               );
```
Paul-Licameli, line 963: If you rebase onto my fixup commit https://discord.com/channels/@me/840616766099161088/1151545897302360105, and resolve the conflict, then you will understand that the `push_back` above should be pulled out of the loop.  `captureSequences` must contain leader tracks only, and each one only once.  Then test that recording to stereo tracks still works.

Paul-Licameli: That completes review of this one commit of the draft.

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -800,31 +802,22 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
 /// Split a stereo track (or more-than-stereo?) into two (or more) tracks...
 void WaveTrackMenuTable::SplitStereo(bool stereo)
 {
-   WaveTrack *const pTrack = static_cast<WaveTrack*>(mpData->pTrack);
-   wxASSERT(pTrack);
-   AudacityProject *const project = &mpData->project;
-
-   auto channelRange = pTrack->Channels();
+  AudacityProject *const project = &mpData->project;
```
Paul-Licameli, line 805: Indent

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -837,31 +837,21 @@ void WaveTrackMenuTable::OnSwapChannels(wxCommandEvent &)
...
+   if (auto track = TrackList::SwapChannels(*mpData->pTrack))
+   {
       if (hasFocus)
-         trackFocus.Set(partner);
+         trackFocus.Set(track);
 
       ProjectHistory::Get( *project ).PushState(
          /* i18n-hint: The string names a track  */
-         XO("Swapped Channels in '%s'").Format( pTrack->GetName() ),
+         XO("Swapped Channels in '%s'").Format( mpData->pTrack->GetName() ),
```
Paul-Licameli, line 854: That should be `track->GetName()`

Paul-Licameli: Just to make no assumptions about the continued validity of the old track pointer

libraries/lib-track/Track.cpp
```diff
@@ -721,6 +721,14 @@ bool TrackList::MakeMultiChannelTrack(Track& track, int nChannels, bool aligned)
...
 
       (*first)->SetLinkType(aligned ? Track::LinkType::Aligned : Track::LinkType::Group);
+
+      //Cleanup the group data in all channels except the first
+      for(auto it = std::next(first), last = std::next(first, nChannels);
+         it != last;
+         ++it)
+      {
+         (*it)->DestroyGroupData();
+      }
```
Paul-Licameli, line 731: Your commit comment suggest that you think this is redundant. I think you are right after all.  I overlooked other details.  But the extra insurance is harmless.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -436,6 +437,24 @@ std::shared_ptr<WaveTrack> WaveTrackFactory::Create(sampleFormat format, double
...
 
+TrackListHolder WaveTrackFactory::Create(size_t nChannels)
+{
+   return Create(nChannels, QualitySettings::SampleFormatChoice(), mRate.GetRate());
+}
+
+TrackListHolder WaveTrackFactory::Create(size_t nChannels, sampleFormat format, double rate)
+{
+   auto channels = std::vector<std::shared_ptr<Track>>{ };
+   std::generate_n(
```
Paul-Licameli, line 448: Thank you for making me look up this standard algorithm and refresh my memory of it.

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
vsverchinsky: Ah, overlooked that merge conflict, thanks!

Paul-Licameli:
Windows build errors:
`D:\a\audacity\audacity\src\import\ImportRaw.cpp(224,1): error C2440: 'initializing': cannot convert from 'int' to 'size_t &' [D:\a\audacity\audacity\.build.x32\src\Audacity.vcxproj]`
`D:\a\audacity\audacity\src\import\ImportRaw.cpp(229,1): error C2440: 'initializing': cannot convert from 'int' to 'size_t &' [D:\a\audacity\audacity\.build.x32\src\Audacity.vcxproj]`
`D:\a\audacity\audacity\src\import\ImportRaw.cpp(221,26): error C2664: 'void ImportUtils::ForEachChannel(TrackList &,const std::function<void (WaveChannel &)> &)': cannot convert argument 2 from 'ImportRaw::<lambda_cd67ecb3f9410bfd47f9bd99446097e3>' to 'const std::function<void (WaveChannel &)> &' [D:\a\audacity\audacity\.build.x32\src\Audacity.vcxproj]`

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -837,31 +837,21 @@ void WaveTrackMenuTable::OnSwapChannels(wxCommandEvent &)
 {
    // Fix assertion violation in `TrackPanel::OnEnsureVisible` by
    // dispatching any queued event
-   // TODO wide wave tracks -- remove this when there is no "leader" distinction
-   // any more
```
Paul-Licameli, line 848: Keep this TODO comment please, to remind me later to remove the line below

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -1074,6 +1074,22 @@ class WAVE_TRACK_API WaveTrackFactory final
...
+    */
+   TrackListHolder Create(size_t nChannels);
+
+   /**
+    * \brief Creates new \p nChannels tracks and places them into TrackList.
+    * If number of channels is exactly two a single stereo track is created
+    * instead.
+    * \param format 
+    * \param rate 
+    * \return
```
Paul-Licameli, line 1096: return what?

src/tracks/playabletrack/wavetrack/ui/WaveTrackMenuItems.cpp
```diff
@@ -60,22 +60,15 @@ void OnNewStereoTrack(const CommandContext &context)
...
+   (*tracks.rbegin())->SetSelected(true);
+   (*tracks.rbegin())->SetName(tracks.MakeUniqueTrackName(WaveTrack::GetDefaultAudioTrackNamePreference()));
 
    ProjectHistory::Get( project )
       .PushState(XO("Created new stereo audio track"), XO("New Track"));
 
-   TrackFocus::Get(project).Set(left.get());
-   left->EnsureVisible();
+   TrackFocus::Get(project).Set(*tracks.rbegin());
+   (*tracks.rbegin())->EnsureVisible();
```
Paul-Licameli, line 71: Very good:  we start to eliminate `MakeMultiChannelTrack` except as an implementation detail of Track.cpp.

I would declare a variable to eliminate repetition of `*tracks.rbegin()`

libraries/lib-track/Track.h
```diff
@@ -1286,6 +1286,15 @@ class TRACK_API TrackList final
    //! Remove all tracks from `list` and put them at the end of `this`
    void Append(TrackList &&list);
 
+   // Like RegisterPendingChangedTrack, but for a list of new tracks,
+   // not a replacement track.  Caller
+   // supplies the list, and there are no updates.
+   // Pending tracks will have an unassigned TrackId.
+   // Pending changed tracks WILL occur in iterations, always after actual
```
Paul-Licameli, line 1293: You copy-pasted an old mistake in the words of the comment.  Should be "new" not "changed"

src/ProjectAudioManager.cpp
```diff
@@ -944,31 +942,30 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
             // ISO standard would be nice, but ":" is unsafe for file name.
             nameSuffix.Replace(wxT(":"), wxT("-"));
 
-            if (newTrack.get() == first) {
+            if (newTrack == first) {
```
Paul-Licameli, line 945: I made a mistake here at 03c68afc55b2020b0e8086aa288090b4e5ea8ca2, neglecting the provisions for more than two channels (maybe in future).

Correct my mistake.  Change the condition back to
`if (baseTrackName.empty())`

src/ProjectAudioManager.cpp
```diff
@@ -944,31 +942,30 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
...
-                !(ProjectSettings::Get(*p).GetTracksFitVerticallyZoomed())) {
-               ChannelView::Get(*newTrack->GetChannel(0)).SetMinimized(true);
-            }
+            //create a new clip with a proper name before recording is started
+            newTrack->CreateWideClip(t0, makeNewClipName(newTrack));
 
             // Temporary until there is a "wide" overload of
             // WaveTrackFactory::Create
-            if (!(c == 1 && recordingChannels == 2))
-               transportSequences.captureSequences.push_back(newTrack);
```
Paul-Licameli: The commend before this deletion should be deleted too now

libraries/lib-effects/MixAndRender.cpp
```diff
@@ -95,34 +95,12 @@ TrackListHolder MixAndRender(const TrackIterRange<const WaveTrack> &trackRange,
       oneinput = true;
    // only one input track (either 1 mono or one linked stereo pair)
 
-   // EmptyCopy carries over any interesting channel group information
-   // Maybe first is stereo, or it is mono, but some other input is stereo;
-   // therefore don't use WideEmptyCopy, but make and link a right channel as
-   // needed
-   // And reset pan and gain
-   auto mixLeft =
-      first->EmptyCopy(trackFactory->GetSampleBlockFactory(), false);
```
Paul-Licameli, line 104: You did what I asked, removing this and instead creating a "wide" track (as a TrackList) through the factory.

But now I see that this makes an unintended change of behavior.  The call of `WaveTrack::Init()` in `EmptyCopy` doesn't happen, and so certain information of `first` is no longer copied into the new track -- things such as the color, and whatever else in future might be attached to the `ChannelGroup` client data site.

So the new `Create` function should take `first` as another argument, which could default to nullptr, and from which it copies information as in `EmptyCopy`.

Or instead of that -- I think it's nicer -- add `std::optional<size_t> oNChannels = {}` as an argument in `EmptyCopy` and call that.

libraries/lib-effects/MixAndRender.cpp
```diff
@@ -184,7 +156,7 @@ TrackListHolder MixAndRender(const TrackIterRange<const WaveTrack> &trackRange,
    wxPrintf("Max number of tracks to mix in real time: %f\n", maxTracks);
 #endif
 
-      for (auto pTrack : { mixLeft, mixRight })
+      for (auto pTrack : *result)
          if (pTrack)
             RealtimeEffectList::Get(*pTrack).Clear();
```
Paul-Licameli, line 161: This no longer needs to be a loop.  Just do this once on `mix`.

libraries/lib-effects/MixAndRender.cpp
```diff
@@ -150,23 +128,17 @@ TrackListHolder MixAndRender(const TrackIterRange<const WaveTrack> &trackRange,
...
-            mixLeft->Append(buffer, format, blockLen, 1, effectiveFormat);
-         }
-         else {
-            auto buffer = mixer.GetBuffer(0);
-            mixLeft->Append(buffer, format, blockLen, 1, effectiveFormat);
-            buffer = mixer.GetBuffer(1);
-            mixRight->Append(buffer, format, blockLen, 1, effectiveFormat);
+         for(auto channel : mix->Channels())
+         {
+            auto buffer = mixer.GetBuffer(channel->GetChannelIndex());
```
Paul-Licameli, line 133: In the present transitional state of the source tree, this needs to be `ReallyGetChannelIndex`, I think

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2117,13 +2117,18 @@ void WaveTrack::JoinOne(WaveTrack &track, double t0, double t1)
...
 }
 
+bool WaveChannel::Append(constSamplePtr buffer, sampleFormat format, size_t len)
+{
+   return GetTrack().Append(buffer, format, len, 1, widestSampleFormat);
+}
+
 /*! @excsafety{Partial}
 -- Some prefix (maybe none) of the buffer is appended,
 and no content already flushed to disk is lost. */
```
Paul-Licameli, line 2137: Duplicate these comments before the overload defined above.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1007,6 +1007,16 @@ TrackListHolder WaveTrack::WideEmptyCopy(
    return result;
 }
 
+TrackListHolder WaveTrack::MonoToStereo()
+{
+   assert(NChannels() == 1);
```
Paul-Licameli, line 1012: Duplicate this assertion in a `@pre` comment.

src/effects/nyquist/Nyquist.cpp
```diff
@@ -1695,12 +1695,8 @@ bool NyquistEffect::ProcessOne(
    if (outChannels < static_cast<int>(mCurNumChannels)) {
       // Be careful to do this before duplication
       out->Flush();
-      auto dup = (*out->Duplicate()->begin())->SharedPointer<WaveTrack>();
-      // Must destroy one temporary list before repopulating another with
-      // correct channel grouping
```
Paul-Licameli, line 1700: Don't delete the comment!

src/effects/nyquist/Nyquist.cpp
```diff
@@ -1695,12 +1695,8 @@ bool NyquistEffect::ProcessOne(
...
+      tempList = out->MonoToStereo();
       nyxContext.mOutputTracks.reset();
-      tempList = TrackList::Temporary(nullptr, out, dup);
-      assert(!dup->IsLeader());
```
Paul-Licameli, line 1703: Move the `MonoToStereo` call after the `reset`, consisntently with the comment.

Did you hit a breakpoint here in the debugger?  I think you need to do what I wrote to avoid putting the `TrackList` objects into an inconsistent state.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli, line 1012: It should also be asserted, and commented with `@pre`, that `this` has null for `GetOwner()`.

libraries/lib-import-export/ImportUtils.h
```diff
@@ -30,15 +34,22 @@ class IMPORT_EXPORT_API ImportUtils final
    //! Choose appropriate format, which will not be narrower than the specified one
    static sampleFormat ChooseFormat(sampleFormat effectiveFormat);
    
-   //! Build a wave track with appropriate format, which will not be narrower than the specified one
-   static std::shared_ptr<WaveTrack> NewWaveTrack( WaveTrackFactory &trackFactory,
+   //! Build a wave track places it into a track list.
```
Paul-Licameli, line 37: "Builds a WaveTrack and places etc..."

libraries/lib-import-export/ImportUtils.h
```diff
@@ -30,15 +34,22 @@ class IMPORT_EXPORT_API ImportUtils final
...
    static void ShowMessageBox(const TranslatableString& message, const TranslatableString& caption = XO("Import Project"));
-   
-   using NewChannelGroup = std::vector<std::shared_ptr<WaveTrack>>;
 
-   //! Flush the given channels and group them into tracks
+   //! Iterates over channels in each wave track from the list
+   static
+   void ForEachChannel(TrackList& trackList, const std::function<void(WaveChannel&)>& op);
+
+   //! Flush the given channels and moves them to \p outTracks
```
Paul-Licameli, line 48: "Flushes" and "moves", consistent verb forms

libraries/lib-import-export/ImportUtils.h
```diff
@@ -30,15 +34,22 @@ class IMPORT_EXPORT_API ImportUtils final
...
+   static
+   void FinalizeImport(TrackHolders& outTracks, const std::vector<TrackListHolder>& importedStreams);
+
+   //! Flush the given channels and moves them to \p outTracks
```
Paul-Licameli, line 52: ditto

modules/mod-ffmpeg/ImportFFmpeg.cpp
```diff
@@ -582,62 +566,68 @@ void FFmpegImportFileHandle::Stop()
...
-      // Write audio into WaveTracks
-      auto iter2 = iter->begin();
-      for (size_t chn = 0; chn < nChannels; ++iter2, ++chn)
+      unsigned chn = 0;
+      ImportUtils::ForEachChannel(*stream, [&](auto& channel)
       {
-         iter2->get()->Append(
-            reinterpret_cast<samplePtr>(data.data() + chn), sc->SampleFormat,
+         if(chn >= nChannels)
+            return;
```
Paul-Licameli, line 597: Why do you make bounds checks on this unsigned index variable only in ImportFFmpeg.cpp but not in the other place that use `ForEachChannel`?  Should there be checks in the other places?

Paul-Licameli:
Finished a round of review

libraries/lib-wave-track/WaveTrack.cpp
vsverchinsky: Then `//! @pre !GetOwner()` only is enough precondition as it implies that `NChannels() == 1`

libraries/lib-import-export/ImportUtils.h
vsverchinsky: Thanks!

modules/mod-ffmpeg/ImportFFmpeg.cpp
vsverchinsky: I don't think so, I need that to preserve existing loop exit condition.
`const auto nChannels = std::min(sc->CodecContext->GetChannels(), sc->InitialChannels);`
Which formally is not same as `mStreams.size()`, but `nChannels` was used as an upper bound in the original loop

libraries/lib-effects/MixAndRender.cpp
```diff
@@ -95,34 +95,15 @@ TrackListHolder MixAndRender(const TrackIterRange<const WaveTrack> &trackRange,
...
-   mixLeft->ConvertToSampleFormat(format);
-   if (oneinput)
-      mixLeft->SetName(first->GetName()); /* set name of output track to be the same as the sole input track */
-   else
-      /* i18n-hint: noun, means a track, made by mixing other tracks */
-      mixLeft->SetName(newTrackName);
-   mixLeft->MoveTo(mixStartTime);
-
-   auto maxBlockLen = mixLeft->GetIdealBlockSize();
+   auto result = trackFactory->Create(mono ? 1 : 2, *first);
```
Paul-Licameli, line 98: Now you are assuming that `format` and `rate`, which were formerly passed to `Create`, are the same as the format and rate of `first`.  But is this so?  Should we examine all uses of `MixAndRender`?

I was expecting instead that the new overload would take format and rate, and use the prototype only to copy other things as in `WaveTrack::Init`.

libraries/lib-import-export/ImportUtils.h
```diff
@@ -30,15 +34,22 @@ class IMPORT_EXPORT_API ImportUtils final
    //! Choose appropriate format, which will not be narrower than the specified one
    static sampleFormat ChooseFormat(sampleFormat effectiveFormat);
    
-   //! Build a wave track with appropriate format, which will not be narrower than the specified one
-   static std::shared_ptr<WaveTrack> NewWaveTrack( WaveTrackFactory &trackFactory,
+   //! Builds a wave track places it into a track list.
```
Paul-Licameli, line 37: "and places"

src/effects/nyquist/Nyquist.cpp
```diff
@@ -1695,12 +1695,10 @@ bool NyquistEffect::ProcessOne(
    if (outChannels < static_cast<int>(mCurNumChannels)) {
       // Be careful to do this before duplication
       out->Flush();
-      auto dup = (*out->Duplicate()->begin())->SharedPointer<WaveTrack>();
       // Must destroy one temporary list before repopulating another with
       // correct channel grouping
       nyxContext.mOutputTracks.reset();
-      tempList = TrackList::Temporary(nullptr, out, dup);
-      assert(!dup->IsLeader());
+      tempList = out->MonoToStereo();
```
Paul-Licameli, line 1701: In this commit's comment, "Eliminate"

Paul-Licameli:
I would also like to see a rebase that squashes by last two commits as fixups on earlier ones.

"Fix Windows Build" fixes "Eliminate `ImportUtils::MakeTracks`"

"Fix recording into stereo" fixes "Don't assume correspondence of channels and RecordableSequences"

Paul-Licameli:
All the builds are failing

vsverchinsky:
Rebased onto fixed master

Paul-Licameli:
Please rebase and reword one of my commit comments
"RecordableSequences are really wide and correspond with channels"
Should say tracks, not channels

vsverchinsky:
OGG Import crash fix & rebase

vsverchinsky:
Attempted to fix recording to stereo fix

src/ProjectAudioManager.cpp
```diff
@@ -854,34 +854,33 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
...
+            // - or the check for creating a new clip or not should be more
+            // general than that ...
+            assert(
+               !recordingStartsBeforeTrackEnd ||
+               lastClip->WithinPlayRegion(recordingStart));
+            if (!recordingStartsBeforeTrackEnd ||
+               !lastClip->StretchRatioEquals(1))
+               pending->CreateWideClip(t0, makeNewClipName(pending));
+            transportSequences.captureSequences
+               .push_back(pending->SharedPointer<WaveTrack>());
```
Paul-Licameli, line 883: Thank you for figuring out what I omitted in recording.

All of this looks like the right idea to me.  Why do you call it only an "attempt"?  Did you observe that something still doesn't work?

vsverchinsky:
Rebased

PR #5201, opened 2023-09-16
3.3.3 and older versions cannot open project files with any stretches

Paul-Licameli:

QA:
- [x] Save a project from 3.4 containing some stretched or shrunk clips, with or without some clips with hidden data.
- [x] Attempt to open in earlier 3.x and see "This project was created with a newer version of Audacity. You will need to upgrade to open it."
- [x] Save a project from 3.4 without any stretch, but with some hidden smart clip data.
- [x] Open the project successfully in 3.1 or later.  Still fail to open in earlier 3.x.

Recommended:


crsib (approved):

chinakov:
Tested

Issue #4954, opened 2023-07-25
The Detach at Silences command may incorrectly make splits where only the left channel is silent


Paul-Licameli:
### Bug description

This bug will exist after #4937 is merged

### Coding tasks
- [x] In `WaveTask::Disjoin` use the width of the track `NChannels()`
- [x] Rewrite the loop over clips as over `Intervals()` instead, adding member functions to `WaveTrack::Interval` as needed
- [x] Where samples are fetched from the clip, use `WaveChannelInterval::GetSamples` instead, in a loop over the `Channels()` of the `WaveTrack::Interval`

### Steps to reproduce

1. Make stereo track with rhythm track in the left but tone in the right
2. Select all
3. Edit > Audio Clips > Detach at Silences

### Expected behavior

No splitting of the track

### Actual behavior

The track is split at silences of the left channel, ignoring the right

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

dozzzzer:
Tested

Issue #5159, opened 2023-09-06
Rewrite WaveTrackShifter, extending WaveTrack::Interval, avoiding `GetClip`


Paul-Licameli:
See this comment in another issue: https://github.com/audacity/audacity/pull/5091#discussion_r1311846185

If you understand that:  then, find a way to rewrite these places where `WaveTrackShifter` calls`GetClip` twice:

- [x] `WaveClipShifter::Attach`
- [x] `WaveClipShifter::Detach`
- [x] `WaveClipShifter::DoHorizontalOffset`

Add member functions to `WaveTrack::Interval` as needed.

PR #5205, opened 2023-09-18
4954 disjoin fix

crsib:
Resolves: #4954 

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -248,6 +248,16 @@ double WaveTrack::Interval::GetStretchRatio() const
    return mpClip->GetStretchRatio();
 }
 
+sampleCount WaveTrack::Interval::TimeToSamples(double time) const
+{
+   return mpClip->TimeToSamples(time);
+}
+
+double WaveTrack::Interval::SamplesToTime(sampleCount s) const
+{
+   return mpClip->SamplesToTime(s);
+}
+
 bool WaveTrack::Interval::IsPlaceholder() const
 {
    return mpClip->GetIsPlaceholder();
@@ -2092,12 +2102,12 @@ void WaveTrack::Disjoin(double t0, double t1)
    std::vector<float> buffer;
    std::vector<samplePtr> buffers;
    Regions regions;
+   
+   const size_t width = NChannels();
 
-   // TODO wide wave tracks -- use width of the clip
-   const size_t width = 1;
-   for (const auto &clip : mClips) {
-      double startTime = clip->GetPlayStartTime();
-      double endTime = clip->GetPlayEndTime();
+   for (const auto &interval : Intervals()) {
+      double startTime = interval->Start();
+      double endTime = interval->End();
 
       if (endTime < t0 || startTime > t1)
          continue;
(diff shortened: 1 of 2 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -218,6 +218,23 @@ double WaveTrack::Interval::GetPlayStartTime() const
    return mpClip->GetPlayStartTime();
 }
 
+double WaveTrack::Interval::GetPlayEndTime() const
+{
+   // TODO wide wave tracks:  assuming that all 'narrow' clips share common
+   // beginning
+   return mpClip->GetPlayEndTime();
+}
```
Paul-Licameli, line 226: You don't need this function.  Instead use `End` inherited from `ChannelGroupInterval`.

vsverchinsky: `ChannelGroupInterval::mStart` and `ChannelGroupInterval::mEnd` are immutable, whereas `WaveClip::GetPlayStartTime` and `WaveClip::GetPlayEndTime` always return actual clip borders, though the difference may not be relevant in this particular PR

crsib: I've addressed it in a way @Paul-Licameli suggested, but that can be undone easily.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1992,15 +2009,19 @@ void WaveTrack::Disjoin(double t0, double t1)
...
       for (decltype(len) done = 0; done < len; done += maxAtOnce) {
          auto numSamples = limitSampleBufferSize(maxAtOnce, len - done);
 
-         clip
-            ->GetSamples(buffers.data(), floatSample, start + done, numSamples);
+         auto bufferIt = buffers.begin();
+
+         for (auto channel : interval->Channels())
+            channel->GetSamples(
+               *bufferIt++, floatSample, start + done, numSamples);
```
Paul-Licameli, line 2016: This works.

Maybe `WaveTrack::Interval` should instead implement `WideSampleSequence` and reduce that to a single call.

Not in this PR.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: Yes `mStart` and `mEnd` are immutable, but now I dislike that choice I made and may change it so that a virtual function reports start and end.

That is what bothers me when I consider #5159 which I have not approved yet but isn't actually critical for release now.

For the short-lived usage of intervals in Disjoin, it doesn't matter.

Paul-Licameli (approved):

Paul-Licameli:
I approved it.  I think you can just squash all as one commit.

libraries/lib-wave-track/WaveTrack.cpp
crsib: > That is what bothers me when I consider https://github.com/audacity/audacity/issues/5159

I think Vitaly has addressed the issue you asked. It would be great if that PR is reviewed, not thrown away.

Paul-Licameli:
There is a conflict to resolve with master, but it will be easy

crsib:
PR was rebased, and the two last commits were squashed.

Issue #5211, opened 2023-09-18
Import AUP edge cases


Paul-Licameli:
### Bug description

One loophole that may still allow the creation of projects in 3.4 with misaligned clips is the "Import AUP" feature, which opens projects saved in Audacity 2.4.2 and earlier.

### Steps to reproduce

1. Start Audacity 2.4.2
2. Generate a mono track of noise; time-shift it right some
3. Generate another mono track of noise; time-shift it farther right
4. Use "Make Stereo Track"
5. Drag the channel separator so the channels have unequal heights
6. Save Audacity project; also try "Save Lossless Copy" and "Save Compressed Copy"
8. Close and reopen projects in 2.4.2; observe (Lossless and Compressed didn't restore clip boundaries or channel heights)
9. Quit Audacity 2.4.2; start Audacity 3.4
10. Reopen the files

### Expected behavior

3.4 should not make misaligned clips

### Actual behavior

The normal project file (only; not the compressed or lossless) still has misalignment.

This results from a different code path not in the scope of issue #5202 .

Also observe that unlike in 2.4.2, the difference of channel height is not preserved.  But I recommend allowing this regression, very unlikely anyway to be seen in practice.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

chinakov:
Tested.

PR #5212, opened 2023-09-18
Remove TrackList::Channels in ImportAUP.cpp...

Paul-Licameli:
... This function only restored track properties that are now all de-duplicated and never varied independently between channels -- with one unimportant exception -- so visiting leader tracks only is all we need.

The exception is that wave channel view heights might vary independently. But experiment with 2.4.2 shows that wave channel view height differences were lost anyway when File > Save Lossless Copy or Save Compressed Copy commands were used.

Only files saved with those commands can cause this function to be visited when importing into 3.4.

QA: First follow steps in #5211.  (This PR is NOT the fix for that issue.)
Then verify that opening the compressed and lossless 2.4.2 projects still makes good 3.4 projects.

Recommended:

### DIFF
--- a/modules/mod-aup/ImportAUP.cpp
+++ b/modules/mod-aup/ImportAUP.cpp
@@ -1371,20 +1371,18 @@ bool AUPImportFileHandle::HandleImport(XMLTagHandler *&handler)
    mAttrs.erase(mAttrs.begin());
 
    for (auto pTrack: range.Filter<WaveTrack>()) {
-      for (const auto pChannel : TrackList::Channels(pTrack)) {
-         // Most of the "import" tag attributes are the same as for "wavetrack" tags,
-         // so apply them via WaveTrack::HandleXMLTag().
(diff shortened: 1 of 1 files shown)


crsib (approved):

chinakov:
Tested, the Compressed/Uncompressed AUP files from 2.4.2 open accordingly in 3, with the tracks still misaligned, but silence insterted to make both 'clips' aligned. 

Screenshot 2023-09-20 at 14 08 23


Track height is reset if the project is open in 3.4.

Paul-Licameli:
> Tested, the Compressed/Uncompressed AUP files from 2.4.2 open accordingly in 3, with the tracks still misaligned, but silence insterted to make both 'clips' aligned.
> Screenshot 2023-09-20 at 14 08 23
> 
> Track height is reset if the project is open in 3.4.

In other words, it's not real clip misalignment.  Clip boundaries agree across channels.

Issue #2263, opened 2021-12-10
Zero length clips sometimes created on copy-paste


vsverchinsky:
STR:
1. Create a clip
2. Split clip at some point
3. Select and copy the rightmost clip
4. Paste it
Observe:
Sometimes (depends on a split position), two clips are pasted: one that was selected in step 3 and a zero-length clip
![image](https://user-images.githubusercontent.com/20165230/145573735-03bc0112-c0ed-4b21-8bb0-54c689a457bc.png)

Note the bold vertical line before the "Audio Track.1".

Version: Audacity 3.1.3

==============

A more reliable way to reproduce this bug (also works in earlier versions of Audacity):
1. Generate a clip
2. Select a whole clip and copy it
3. Place cursor right after first clip with > half sample offset (but less than one sample, use snap line as a hint)
4. Paste clip
5. Reselect **pasted** clip
6. Copy and paste anywhere

Observe: bold line indicating the presence of another clip

saintmatthieu:
Isn't this actually the "click-to-merge" line referred by https://github.com/audacity/audacity/issues/2263 ? Not a clip, actually, just a clickable object as shortcut to the merge-clips command.

LWinterberg:
@saintmatthieu It certainly is the visual of the merge line, but why would it be there if there wasn't a clip (in this case: a 0-length clip) present?

saintmatthieu:
Oh, so there _is_ a truly invisible, zero-length clip, plus the merge line?

LWinterberg:
that is often the case, yes. It shows up as a single-pixel black line in some zoom levels and gets in the way of moving clips, which is highly confusing to people

saintmatthieu:
Ok. Maybe #2330 will fix this.

PR #5213, opened 2023-09-19
Remove the "click to merge clips" line

### LINKED ISSUE #2330: Remove the "Click to merge clips" line
image

... and put the option to merge clips in a context menu ("Join Clips") instead. This is because:

1. It gets in the way of selecting audio from the beginning/end of a clip when other clips are present (you can press Esc to toggle into a selection, but still)
2. It obscures the first sample in a clip if the clips are directly adjacent
3. When clips are not directly adjacent but still in a range for it to appear, it just kinda floats in empty space
4. With clips becoming more independent of the audio itself (with Smart Clips 1.0 and especially 2.0), having a one-click thing to merge clips becomes less and less useful, relative to Audacity's feature set

* See also #1497

The context menu option should function in the following way:
* If a range of audio is selected, `Right-Click > Join Clips` joins all selected clips (same as Edit > Clip Boundaries > Join)
* If no audio is selected and the cursor is on top of a Clip boundary, `Right-Click > Join Clips` joins the two adjacent clips (same thing as what clicking the line would do)

This resolves the following bugs in the process
* #1815
* #1663
* #2259
* #2905

saintmatthieu:
Resolves: #2330

Recommended:

QA:
- [x] Does this PR fix https://github.com/audacity/audacity/issues/2263 ? If so, please close.
- [x] Bold vertical line between adjacent clips is not visible anymore, whether at sample-zoom level or not.
- [x] Clicking at junction of adjacent clips doesn't merge them anymore.
- [x] `Edit > Audio Clips > Join` still works (not with stretched clips, though - will be fixed by #5043
- [x] Cutlines still work

### DIFF
--- a/src/WaveTrackLocation.cpp
+++ b/src/WaveTrackLocation.cpp
@@ -22,18 +22,8 @@ WaveTrackLocations FindWaveTrackLocations(const WaveTrack &track)
 
    // Count number of display locations
    int num = 0;
-   {
-      const WaveClip *prev = nullptr;
-      for (const auto clip : clips) {
-         num += clip->NumCutLines();
-
-         if (prev && fabs(prev->GetPlayEndTime() -
-            clip->GetPlayStartTime()) < WAVETRACK_MERGE_POINT_TOLERANCE)
-            ++num;
-
-         prev = clip;
-      }
-   }
+   for (const auto clip : clips)
+      num += clip->NumCutLines();
 
    if (num == 0)
       return locations;
@@ -44,36 +34,18 @@ WaveTrackLocations FindWaveTrackLocations(const WaveTrack &track)
    // Add all display locations to cache
    int curpos = 0;
 
-   const WaveClip *previousClip = nullptr;
    for (const auto clip: clips) {
       for (const auto &cc : clip->GetCutLines()) {
          auto cutlinePosition =
             clip->GetSequenceStartTime() + cc->GetSequenceStartTime();
          if (clip->WithinPlayRegion(cutlinePosition)) {
              // Add cut line expander point
-             locations.emplace_back(cutlinePosition,
-                WaveTrackLocation::locationCutLine);
+             locations.emplace_back(cutlinePosition);
          }
          // If cutline is skipped, we still need to count it
          // so that curpos matches num at the end
          curpos++;
       }
-
-      if (previousClip) {
-         if (fabs(previousClip->GetPlayEndTime() - clip->GetPlayStartTime())
-            < WAVETRACK_MERGE_POINT_TOLERANCE
-         ) {
-            // Add merge point
-            locations.emplace_back(previousClip->GetPlayEndTime(),
-               WaveTrackLocation::locationMergePoint,
-               track.GetClipIndex(previousClip),
-               track.GetClipIndex(clip)
-            );
-            curpos++;
-         }
-      }
-
-      previousClip = clip;
    }
 
    assert(curpos == num);
--- a/src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
@@ -187,10 +187,6 @@ UIHandle::Result CutlineHandle::Release
    switch (mOperation) {
    default:
       wxASSERT(false);
-   case Merge:
-      ProjectHistory::Get( *pProject )
-         .PushState(XO("Merged Clips"), XO("Merge"), UndoPush::CONSOLIDATE);
-      break;
    case Expand:
       ProjectHistory::Get( *pProject )
          .PushState(XO("Expanded Cut Line"), XO("Expand"));
--- a/src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
@@ -42,36 +42,16 @@ void CutlineHandle::Enter(bool, AudacityProject *)
 #endif
 }
 
-HitTestPreview CutlineHandle::HitPreview(bool cutline, bool unsafe)
+HitTestPreview CutlineHandle::HitPreview(bool unsafe)
 {
    static auto disabledCursor =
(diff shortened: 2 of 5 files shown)


vsverchinsky (commented):
Isn't this PR also supposed to
> put the option to merge clips in a context menu ("Join Clips") instead

?

saintmatthieu:
There was agreement during stand-up that this could be part of a follow-up, 3.x ticket. I think there is envy to get rid of this strange vertical line and the associated behaviour, but making it up for in the clip menu is considered not so important UX-wise.
I'll spend some time-boxed time tomorrow to experiment and see how much more work that would entail. If not much, I'll make it part of this PR. Else, I'll create a follow-up ticket.

Paul-Licameli (changes requested):
I disapprove for several reasons, to follow.

src/WaveTrackLocation.cpp
```diff
@@ -44,36 +34,18 @@ WaveTrackLocations FindWaveTrackLocations(const WaveTrack &track)
...
-            locations.emplace_back(previousClip->GetPlayEndTime(),
-               WaveTrackLocation::locationMergePoint,
-               track.GetClipIndex(previousClip),
-               track.GetClipIndex(clip)
-            );
-            curpos++;
-         }
-      }
-
-      previousClip = clip;
```
Paul-Licameli, line 76: There are changes to WaveTrackLocations and the drawing of clip edges.  The bold lines will not be drawn when magnification resolves samples — I understand that this is wanted, from the issue description.   But drawing will change at lesser magnifications too.  I don’t think this was specified.

src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
```diff
@@ -42,36 +42,16 @@ void CutlineHandle::Enter(bool, AudacityProject *)
...
-      const double tolerance = 0.5 / track.GetRate();
-      int ii = 0;
-      for (const auto loc: locations) {
-         if (loc.typ == WaveTrackLocation::locationMergePoint &&
-             fabs(time - loc.pos) < tolerance)
-            return ii;
-         ++ii;
-      }
-      return -1;
-   }
```
Paul-Licameli, line 73: If the issue is ever fully satisfied (and it is not in this PR), then the merging of clips will still be available as a context menu command, and I think you will still need what is deleted here.

src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
```diff
@@ -160,32 +140,7 @@ UIHandle::Result CutlineHandle::Click
...
-         mOperation = Expand;
-         mStartTime = viewInfo.selectedRegion.t0();
-         mEndTime = viewInfo.selectedRegion.t1();
-
-         // When user presses left button on cut line, expand the line again
-         double cutlineStart = 0, cutlineEnd = 0;
-         mpTrack->ExpandCutLine(mLocation.pos, &cutlineStart, &cutlineEnd);
-         viewInfo.selectedRegion.setTimes(cutlineStart, cutlineEnd);
-      }
-      else if (mLocation.typ == WaveTrackLocation::locationMergePoint) {
```
Paul-Licameli, line 176: This eliminates the left click action to expand cut lines, but that was not in the scope of the issue.

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
```diff
@@ -848,17 +848,7 @@ void WaveChannelSubView::DrawBoldBoundaries(
...
-            dc.SetPen( highlightLoc ? AColor::uglyPen : *wxRED_PEN );
-         }
-         else {
-#ifdef EXPERIMENTAL_DA
-            // JKC Black does not show up enough.
-            dc.SetPen(highlightLoc ? AColor::uglyPen : *wxWHITE_PEN);
-#else
-            dc.SetPen(highlightLoc ? AColor::uglyPen : *wxBLACK_PEN);
-#endif
-         }
```
Paul-Licameli, line 861: Suppress drawing of clip edges only at close magnifications.

src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
saintmatthieu: Ouch - thanks.

src/tracks/playabletrack/wavetrack/ui/WaveChannelView.cpp
saintmatthieu: I don't understand. The only remaining possibility is for `loc.typ` to equal `WaveTrackLocation::locationCutLine`, and the remaining code is what was executed in that situation.

src/tracks/playabletrack/wavetrack/ui/CutlineHandle.cpp
saintmatthieu: This is a member function of `CutlineHandle`. The functionality, when presented over the clip menu, will live someplace else. And since it is of the simple sort, I opted for not leaving dead code behind.

src/WaveTrackLocation.cpp
saintmatthieu: I thought there was an agreement that there should remain nothing of merge lines anymore. The track panel looks good without them at lower magnification. Maybe I don't understand your concern.

Paul-Licameli:
Good, the cutline actions are restored.

You can simplify CutlineHandle further by removing `Merge` from `enum Operation`.

Paul-Licameli:
Still this pull request is incomplete for addressing the issue description, and I'm not sure I understand the implications for drawing.

saintmatthieu:
Follow-up issue: #5221

Paul-Licameli:
> Follow-up issue: #5221

Then the description of the issue linked to this PR should be edited to remove mention of context menus.

Paul-Licameli (approved):
No reason to block this.  I hope you rebase and squash to one commit.  Also, I repeat:

You can simplify CutlineHandle further by removing Merge from enum Operation.

Paul-Licameli:
I see the new commit, it's ok

dozzzzer:
Tested

Issue #5200, opened 2023-09-16
"Make Stereo Track" acts just like "Mix And Render" applied to hard-panned mono tracks


Paul-Licameli:
Note:  this means that joining stereo to mono ignores pan slider settings the tracks already had.

chinakov:
Tested 'Make Stereo Track' ignores channel panning. 

Before 'Make Stereo Track':
![image](https://github.com/audacity/audacity/assets/139755141/651ae949-c9f3-427a-a03e-149a7506561b)

After 'Make Stereo Track': 
![image](https://github.com/audacity/audacity/assets/139755141/c774f0b9-df6f-4fae-92be-ee50f1c82c91)

PR #5214, opened 2023-09-19
Merge tracks to stereo rewrite

vsverchinsky:
Resolves: #5200

Recommended:

### DIFF
--- a/libraries/lib-track/Track.cpp
+++ b/libraries/lib-track/Track.cpp
@@ -560,6 +560,30 @@ bool TrackList::SwapChannels(Track &track)
    return true;
 }
 
+void TrackList::Insert(const Track* before, TrackList&& trackList)
+{
+   assert(before == nullptr || (before->IsLeader() && Find(before) != EndIterator<const Track>()));
+
+   if(before == nullptr)
+   {
+      Append(std::move(trackList));
+      return;
+   }
+
+   std::vector<Track *> arr;
+   arr.reserve(Size() + trackList.Size());
+   for (const auto track : *this) {
+      if (track == before)
+      {
+         for(const auto addedTrack : trackList)
+            arr.push_back(addedTrack);
+      }
+      arr.push_back(track);
+   }
+   Append(std::move(trackList));
+   Permute(arr);
+}
+
 void TrackList::Permute(const std::vector<Track *> &tracks)
 {
    std::vector<TrackNodePointer> permutation;
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -1173,6 +1173,11 @@ public:
 
    friend class Track;
 
+   //! @brief Inserts tracks form \p trackList starting from position where
+   //! \p before is located. If \p before is nullptr tracks are appended.
+   //! @pre `before == nullptr || (before->IsLeader() && Find(before) != EndIterator<const Track>())`
+   void Insert(const Track* before, TrackList&& trackList);
+
    /*!
     @pre `tracks` contains pointers only to leader tracks of this, and each of
     them exactly once
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
@@ -43,6 +43,8 @@ Paul Licameli split from TrackPanel.cpp
 #include <wx/frame.h>
 #include <wx/sizer.h>
 
+#include "MixAndRender.h"
+
 WaveTrackControls::~WaveTrackControls() = default;
 
 std::vector<UIHandlePtr> WaveTrackControls::HitTest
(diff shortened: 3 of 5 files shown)


libraries/lib-track/Track.h
```diff
@@ -1167,6 +1167,9 @@ class TRACK_API TrackList final
 
    friend class Track;
 
+   ///@pre (before->IsLeader() && Find(before) != end()) || before == nullptr
```
Paul-Licameli, line 1170: I would change the order of the terms of `||` and put ` ` around it so that it can really be cut and pasted unchanged as an assertion expression at the top of the function definition.

libraries/lib-track/Track.cpp
```diff
@@ -560,6 +560,29 @@ bool TrackList::SwapChannels(Track &track)
    return true;
 }
 
+void TrackList::Insert(const Track* before, TrackList&& trackList)
+{
+   if(before == nullptr)
+   {
+      Append(std::move(trackList));
+      return;
```
Paul-Licameli, line 568: The header comment could explain the behavior of this case

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -741,56 +743,99 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
...
+      RealtimeEffectList::Get(*right).GetStatesCount() != 0 ||
+      !checkAligned(*left, *right))
+   {
+      const auto answer = BasicUI::ShowMessageBox(
+         XO(
+"The tracks you are attempting to merge to stereo contain clips at\n"
+"different positions, or otherwise mismatching clips. Merging them\n"
+"will render the tracks.\n\n" 
+"This causes any realtime effects to be applied to the waveform and\n"
+"hidden data to get removed. Additionally, clips may get padded with\n"
```
Paul-Licameli, line 789: This extra checking and the message are not specified in the Issue.  Did designers specify it?

I would write "be" not "get" twice on this line.  Also, are the embedded newlines really needed, or does the text wrap?

@LWinterberg

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -741,56 +743,99 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
...
+         ++tracks.Any<const WaveTrack>().find(right)
+      },
+      Mixer::WarpOptions{ tracks.GetOwner() },
+      (*first)->GetName(),
+      &WaveTrackFactory::Get(*project),
+      //use highest sample rate
+      std::max(left->GetRate(), right->GetRate()),
+      //use widest sample format
+      std::max(left->GetSampleFormat(), right->GetSampleFormat()),
+      0.0, 0.0);
```
Paul-Licameli, line 821: You rely on `MixAndRender` using instead the minimum start and maximum end times of the range, when it is given an empty interval of times.

I had forgot that behavior and stepped in the debugger to convince myself this call works.

The header file for `MixAndRender` should comment this behavior.

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -741,56 +743,99 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
...
-           .Caption(XO("Error"))
-           .IconStyle(Icon::Error));
-      return;
-   }
-   if(pTrack->GetSampleFormat() != partner->GetSampleFormat())
+   const auto checkAligned = [](const WaveTrack& left, const WaveTrack& right)
    {
-      BasicUI::ShowMessageBox(XO(
-"Mono tracks must have the same sample format in order to be combined "
-"into a stereo track"),
```
Paul-Licameli, line 764: Why are the tests for sameness of rate and format removed?

vsverchinsky: I think it was implied that these checks are not needed any more
изображение

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -741,56 +743,99 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
...
+            {
+               return a->Start() == b->Start() &&
+                  a->End() == b->End() &&
+                  isPlainInterval(*b);
+            });
+         if(it == rightIntervals.end())
+            return false;
+      }
+      return true;
+   };
```
Paul-Licameli, line 777: This big lambda that decides whether to query before proceeding might need reuse elsewhere, when we fix opening of 3.3 files or importing of 2.x files.

It might move to the recently added WaveTrackUtilities.cpp

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
LWinterberg: I put out a message in https://www.figma.com/file/YiOF6yUQS466Fob7bD0vaW/21_WX---Modals?type=design&node-id=2-5&mode=design yes. 

I agree that be is better than get here and I think we can clarify further: 

`"This causes any realtime effects to be applied to the waveform and hidden data to be removed. Additionally, the entire track will become one large clip."`

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
LWinterberg: We're doing a destructive mix and render anyway which resamples on the way anyway, does it not?

Paul-Licameli: Yes.

The past work to disallow mismatched rates or formats was not wasted because other deduplicarion in the data structures was the important accomplishment in those PRs.

Paul-Licameli (approved):
This looks substantially correct.  Let's move it along to testing.  But please remember:
- Comment in MixAndRender.h
- Change "get" twice in the string

Paul-Licameli:
Still I would like to see the test moved to WaveTrackUtils.cpp if you need to reuse it in another commit.

This commit does not fully implement the idea in issue #5203.  To do that, something like the deleted lines would need to be restored as an alternative path to doing `MixAndRender`.

However this pull request does eliminate a call to `MakeMultiChannelTrack` which is what I want to see at this stage of refactoring.  Some other, not yet implemented method of a future wide `WaveTrack` must be invented for this future enhancement.

vsverchinsky:
Dialog text change

Paul-Licameli:
> Dialog text change

I see and reapprove

vsverchinsky:
Rebased

PR #5220, opened 2023-09-20
Applying effect on stretched clip renders only the selected region

### LINKED ISSUE #5208: Applying effect on stretched clip renders only the selected region
When applying an effect on some selection of a stretched clip, the entire clip is stretch-rendered, and then trimmed to the selection:

https://github.com/audacity/audacity/assets/22740106/3830f34b-05f9-4d67-ae66-8880d6549652

The initial motivation was to maximize quality at the beginning and end of the selection. This also has the advantage that users may find handy to access stretched hidden data.

Yet this comes with more important drawbacks:
* whether applying an effect to 1 second of 1 minute of a clip takes the same amount of time,
* the rendered data, possibly big, is saved as part of the project, most likely unnecessarily.

Make sure that quality at the selection boundaries is preserved, yet reduce the amount of stretch data input to a reasonable low and clear the data left and right beyond the selection.

saintmatthieu:
Resolves: #5208

See ticket for detailed description.

Recommended:

QA:
- [x] Small selections are stretch-rendered faster than large ones when applying effects.
- [x] The resulting new, stretch-rendered clip doesn't have hidden parts left and right.
- [x] Please also test with selections beginning/ending within 1 second of start and end of clips
- [x] (PRL) observe a change in behavior that is not just performance when only a part of a stretched clip is selected and an effect applied:  before, the middle clip still had hidden clip data that you can see if you drag it to a new track.  After, it has none.

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -1140,37 +1140,56 @@ void WaveClip::ApplyStretchRatio(const ProgressReporter& reportProgress)
    auto newSequences = GetEmptySequenceCopies();
    bool swappedOnce = false;
 
-   Finally Do { [&, trimLeftBeforeStretch = mTrimLeft,
-                 trimRightBeforeStretch = mTrimRight] {
-      // Whether successful or not, the right thing to do is to restore the
-      // original trim values.
-      this->SetTrimLeft(trimLeftBeforeStretch);
-      this->SetTrimRight(trimRightBeforeStretch);
+   Finally Do { [&, trimLeftBeforeStretch = GetTrimLeft(),
+                 trimRightBeforeStretch = GetTrimRight(),
+                 offsetBeforeStretch = GetSequenceStartTime()] {
       if (success)
       {
          this->mClipStretchRatio = 1.0;
          this->mRawAudioTempo = this->mProjectTempo;
          assert(this->GetStretchRatio() == 1.0);
       }
-      else if (swappedOnce)
-         std::swap(mSequences, newSequences);
+      else
+      {
+         this->SetTrimLeft(trimLeftBeforeStretch);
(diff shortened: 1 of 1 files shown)


libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1140,37 +1140,56 @@ void WaveClip::ApplyStretchRatio(const ProgressReporter& reportProgress)
...
-   SetTrimLeft(0);
-   SetTrimRight(0);
+   const auto originalPlayStartTime = GetPlayStartTime();
+   const auto originalPlayEndTime = GetPlayEndTime();
+
+   // Leave 1 second of raw, unstretched audio before and after visible region
+   // to ensure the algorithm is in a steady state when reaching the play
+   // boundaries.
+   TrimLeftTo(originalPlayStartTime - stretchRatio);
+   TrimRightTo(originalPlayEndTime + stretchRatio);
```
Paul-Licameli, line 1169: Is this the smartest way to leave sufficient time to find the steady state?  Deeper knowledge of the library might be needed to answer that.

Should you reserve a certain amount of time or calculcate how to allow a certain minimum number of analysis windows?

Do you really need at the right end to?  Is the algorithm influenced by some lookahead?  (As with a compressor)

Paul-Licameli (approved):

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: It's a rather quick-and-dirty approach, given. It's probably an overkill, and 1 complete FFT size worth of samples should be sufficient. But we may get the FFT size to become sample-rate dependent in the future (there's an issue for that) and a self-maintaining solution would mean communicating that value somehow. Except for extreme stretch ratios, I don't expect this overkill to have a noticeable impact on performance.

chinakov:
Tested, thank you!

Issue #5195, opened 2023-09-14
Right-click clip header pops up clip menu


saintmatthieu:
### Your idea

First step towards [that design](https://www.figma.com/file/3vAMBYovsJuTrbet34ruaZ/08_WX---Clips?type=design&node-id=1104-27663&mode=design&t=iXpPrnx6bJ9SWhWt-0). See #5192 for an overview.

### Problem to be solved

First step towards [that design](https://www.figma.com/file/3vAMBYovsJuTrbet34ruaZ/08_WX---Clips?type=design&node-id=1104-27663&mode=design&t=iXpPrnx6bJ9SWhWt-0). See #5192 for an overview.

dozzzzer:
Tested

PR #5225, opened 2023-09-21
Adds RMB click support to clip affordances

crsib:
Resolves: #5195

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveClipUtilities.h
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveClipUtilities.h
@@ -14,6 +14,8 @@
 #include <cstddef>
 #include <vector>
 
+#include "../../../ui/CommonTrackPanelCell.h"
+
 class sampleCount;
 
 AUDACITY_DLL_API
@@ -26,5 +28,7 @@ AUDACITY_DLL_API
 void fillWhere(
    std::vector<sampleCount>& where, size_t len, bool addBias, double correction,
    double t0, double sampleRate, double stretchRatio, double samplesPerPixel);
+
+std::vector<CommonTrackPanelCell::MenuItem> GetWaveClipMenuItems();
 
 #endif
--- a/src/tracks/ui/AffordanceHandle.h
+++ b/src/tracks/ui/AffordanceHandle.h
@@ -27,4 +27,8 @@ public:
 
 protected:
     virtual Result SelectAt(const TrackPanelMouseEvent& event, AudacityProject* pProject) = 0;
+
+private:
+    Result UpdateTrackSelection(
+       const TrackPanelMouseEvent& event, AudacityProject* pProject);
 };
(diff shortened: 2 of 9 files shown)


vsverchinsky (approved):

src/tracks/playabletrack/wavetrack/ui/WaveClipUtilities.h
```diff
@@ -27,4 +29,6 @@ void fillWhere(
    std::vector<sampleCount>& where, size_t len, bool addBias, double correction,
    double t0, double sampleRate, double stretchRatio, double samplesPerPixel);
 
+std::vector<CommonTrackPanelCell::MenuItem> GetWaveClipMenuItems();
```
saintmatthieu, line 32: I see that this is called from `WaveChannelSubView::GetMenuItems` when clicking the body of a clip, and from `WaveTrackAffordanceControls` when clicking the header of a clip. I wonder if there isn't a way of achieving a simpler flow where whether body or header is clicked, the same code is executed, only with an additional step toto select all clip if the header is hit ?

saintmatthieu (changes requested):
In any case, please use your .clang-format for the new code.
I think moving that new `protected` method to an anonymous namespace is worthwhile trying out, too.
As for the comment about the simpler flow, there I'm much less certain, and would like to hear your thoughts.
Otherwise I tested the behaviour. LGTM just from user experience, but it'd be interesting to check early with @LWinterberg .

src/tracks/playabletrack/wavetrack/ui/WaveClipUtilities.h
crsib: Unfortunately, they are separate, unrelated objects. I want to keep this PR as simple as possible; I don't see much point in complicating stuff here

crsib: FWTIW this code won't survive the replatforming

src/tracks/ui/AffordanceHandle.h
```diff
@@ -26,5 +26,7 @@ class AUDACITY_DLL_API AffordanceHandle : public TimeShiftHandle
     Result Release(const TrackPanelMouseEvent& event, AudacityProject* pProject, wxWindow* pParent) override;
 
 protected:
+    Result UpdateTrackSelection(const TrackPanelMouseEvent& event, AudacityProject* pProject);
```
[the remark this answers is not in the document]
crsib: It seems that it is no longer needed indeed. I used it when I was trying to get stuff to work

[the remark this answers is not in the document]
crsib: I will move it to `private` then, it still improves readability quite a bit

src/tracks/playabletrack/wavetrack/ui/WaveClipUtilities.h
saintmatthieu: ok

src/tracks/ui/AffordanceHandle.h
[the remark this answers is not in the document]
saintmatthieu: Oh, didn't notice. Please check if it could be const, too.

saintmatthieu (approved):

crsib:
Cosmetics are addressed @saintmatthieu

PR #5228, opened 2023-09-21
Fix writing to atomics in ShareAudioDialog.cpp

Paul-Licameli:

Recommended:

### DIFF
--- a/src/cloud/audiocom/ShareAudioDialog.cpp
+++ b/src/cloud/audiocom/ShareAudioDialog.cpp
@@ -119,7 +119,7 @@ public:
    
    void Cancel()
    {
-      mCancelled.store(std::memory_order_release);
+      mCancelled.store(true, std::memory_order_release);
    }
    
    ExportResult GetResult() const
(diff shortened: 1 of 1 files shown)


crsib (approved):

Paul-Licameli:
Straight to merge?

What symptoms could have been fixed?

Maybe incorrect progress bar update during upload.

crsib:
Straight to merge. I think nothing has changed, `memory_order_release` is not zero

Paul-Licameli:
> Straight to merge. I think nothing has changed, `memory_order_release` is not zero

We don't know whether `memory_order_release` might be zero in some other standard library implementation, and besides one of these is `std::atomic<double>` in which some unchanging converted integer was always stored.

Paul-Licameli:
The error would not have compiled if `std::memory_order` had been an enum class... oh well.

Use strong typing when you can to help the compiler help you!

crsib:
I think it is in c++20

Paul-Licameli:
> I think it is in c++20

Indeed

Issue #5210, opened 2023-09-18
Reverse effect produces inconsistent results when applied on multiple tracks with stretched clips


dozzzzer:
### Bug description

When selecting a region involving two stretched clips on two different tracks and applying Reverse to them several times while only changing the stretch ratio of one of them, it would also produce different results each time for the other clip too.

### Steps to reproduce

1. Generate a chirp with a square wave
2. Duplicate (Ctrl + D) it
3. Stretch/shrink the upper clip
4. Stretch/shrink the lower one by a different amount
5. Select a region inside both clips
6. Effect > Reverse
7. Select the newly created clip and copy and paste it into the new project
8. Ctrl + Z
9. Repeat steps 4 – 7
10. In the second project, Effect > Invert on one of these two clips
11. Ctrl + A > Mix and Render

### Expected behavior

Silence clip

### Actual behavior

Not a silent clip, i.e. the clips produced by the Reverse effect are different despite out of those two clips only the lower one was stretched.

https://github.com/audacity/audacity/assets/70063970/e475c7e7-24a8-47b7-82ed-d6a9f2f6a44c

### Audacity Version

[PR #5138 ](https://github.com/audacity/audacity/pull/5138)

saintmatthieu:
As @Paul-Licameli demonstrated at stand-up, mix-and-rendering of the same track several times produces each time a slightly different result. The different results are likely identical perceptually, still this surprising behaviour should at least be understood.

dozzzzer:
Fixed by #5233

PR #5233, opened 2023-09-22
Fix reads of uninitialized data in lib-time-and-pitch...

Paul-Licameli:
... These changes appear seem to make mix-and-render give repeatable results.

I don't understand why.  Possibly only some of these changes are sufficient.  I don't know which.

UPDATE:  I do understand why!  After more study of the algorithm.

Resolves: [5210](https://github.com/audacity/audacity/issues/5210)

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -166,6 +166,7 @@ void TimeAndPitch::reset()
       d->outCircularBuffer[ch].reset();
    }
    d->normalizationBuffer.reset();
+   d->last_mag.zeroOut();
    d->last_phase.zeroOut();
    d->phase_accum.zeroOut();
    _outBufferWriteOffset = 0;


saintmatthieu:
I trust that the remaining one-liner fixes the issue.
I was foolish to format the code delivered by StaffPad when I integrated it. Someone at the time (I think it was you, Paul) said he didn't like it, and I thought I eventually undid it, but apparently not.
Maybe I should clean that up. Changes we have brought so far to this library don't concern StaffPad so much (fixing Linux build and fix for extreme stretch ratios) but this one more likely does and SP might want to have this upstreamed.

saintmatthieu (approved):

Paul-Licameli:
> I trust that the remaining one-liner fixes the issue. I was foolish to format the code delivered by StaffPad when I integrated it. Someone at the time (I think it was you, Paul) said he didn't like it, and I thought I eventually undid it, but apparently not. Maybe I should clean that up. Changes we have brought so far to this library don't concern StaffPad so much (fixing Linux build and fix for extreme stretch ratios) but this one more likely does and SP might want to have this upstreamed.

I am sure they will want it.  Please communicate with them.

Issue #5202, opened 2023-09-16
If Audacity 3.4 opens a legacy project containing misaligned clips, un-merge them as split to mono


Paul-Licameli:
This needs an info box showing up upon opening a project saying: 

> "This project contained stereo tracks with non-aligned content. This feature is not supported in Audacity versions past 3.3.3. These stereo tracks have been split into mono tracks. As a result, some realtime effects may be missing. Please verify that everything works as intended before saving."

QA:
 - [x] Verify opening of a properly saved project that contains misalligned clips:
   - [x] Shows warning dialog with message above
   - [x] Stereo wave tracks that has misalligned clips is split into two mono tracks
   - [x] These tracks have empty effects stack
   - [x] Splited tracks have same rate and sample format
   - [x] Other tracks are not affected (mono wave tracks, label tracks etc.)
   - [x] Not automatically saved (so it's still possible to open it using Audacity version that was used to create project, and there are **NO** changes in channel linking or effects)
   - [x] Fixed re-saved legacy project should open up in older versions of Audacity without an issue
 - [x] Verify opening of a project that was once saved properly but has unsaved changes/was never saved at all (make changes to a project and force-quit/crash the application)
   - [x] Recovery dialog appears as usual
   - [x] Check that all items from previous list are fullfiled when auto save is discarded.
   - [x] Check that all items from previous list are fullfiled when recover option is choosen. 
 - [x] Projects saved in 3.0.x (before there were smart clips) with well aligned clips open in 3.4 without warnings

dozzzzer:
Tested

PR #5234, opened 2023-09-22
Legacy project compatatibility

vsverchinsky:
Resolves: #5202

Recommended:


libraries/lib-track/Track.h
```diff
@@ -1196,11 +1196,9 @@ class TRACK_API TrackList final
    * @param first and the following must be in this list. Tracks should
    * not be a part of another group (not linked)
    * @param nChannels number of channels, for now only 2 channels supported
-   * @param aligned if true, the link type will be set to Track::LinkType::Aligned,
-   * or Track::LinkType::Group otherwise
```
Paul-Licameli, line 1200: In the commit comment, fix spelling, "unaligned"

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -626,11 +624,8 @@ bool WaveTrack::LinkConsistencyFix(bool doFix)
       }
       else if (doFix) {
          // non-error upgrades happen here
-         auto newLinkType =
-            AreAligned(SortedClipArray(), next->SortedClipArray())
-            ? LinkType::Aligned : LinkType::Group;
-         if (newLinkType != linkType)
-            SetLinkType(newLinkType);
+         if (!AreAligned(SortedClipArray(), next->SortedClipArray()))
+            SetLinkType(LinkType::None);
```
Paul-Licameli, line 628: Again, spelling in the commit comment.

Here, I think you miss the case that an old enough legacy file, with well aligned tracks, still has the legacy value of 1 which must be upgraded to `Aligned`.  Which did happen before this change.

src/ProjectFileManager.cpp
```diff
@@ -165,20 +167,32 @@ auto ProjectFileManager::ReadProjectFile(
...
       for (auto t : tracks) {
+         const auto linkType = t->GetLinkType();
          // Note, the next function may have an important upgrading side effect,
          // and return no error; or it may find a real error and repair it, but
          // that repaired track won't be used because opening will fail.
          if (!t->LinkConsistencyFix()) {
             otherError = XO("A channel of a stereo track was missing.");
             err = true;
          }
+         if(!err && linkType != t->GetLinkType())
```
Paul-Licameli, line 186: This, then, should become the stricter test that the new link type is None and the old was not.  Otherwise the legacy case of type 1 will warn unnecessarily.

src/ProjectFileManager.cpp
```diff
@@ -165,20 +167,32 @@ auto ProjectFileManager::ReadProjectFile(
...
+         {
+            //Not an elegant way to deal with wave track linking
+            //compatibility between versions
+            if(auto waveTrack = dynamic_cast<WaveTrack*>(t))
+            {
+               RealtimeEffectList::Get(*waveTrack).Clear();
```
Paul-Licameli, line 192: I think this is incomplete.

Because `SetLinkType` with argument `None` caused duplication of the effect stack into what was the right channel before it returned.

But this erases the effect stack only on one of the resulting mono tracks.

libraries/lib-wave-track/WaveTrack.cpp
vsverchinsky: Yes, thanks

Paul-Licameli (approved):

Issue #5238, opened 2023-09-25
Menu option to render time stretched clip


crsib:
Initially, this was planned as an "Apply Destructively" button inside the dialog, but it was later decided to make it a separate menu item.

Issue #5236, opened 2023-09-25
Incorrect Undo/Redo entry when stretching clips


dozzzzer:
### Bug description

When stretching a clip, a wring Undo/Redo action state is logged

### Steps to reproduce

1. Grab some audio
2. Stretch the clip by holding Alt key
3. Check the Undo/Redo entry at the top of the Edit menu

### Expected behavior

"Undo change speed by X%" (@LWinterberg please confirm)

### Actual behavior

"Undo Trim by Xs"

### Audacity Version

current master / alpha / nightly build

### Operating system

Other

LWinterberg:
> "Undo change speed by X%" (@LWinterberg please confirm)

confirmed

chinakov:
<img width="406" alt="Screenshot 2023-09-26 at 15 01 47" src="https://github.com/audacity/audacity/assets/139755141/2de2fe24-bd36-44a1-8c83-c8f09e1d0e81">

Confirmed as fixed!

Issue #5196, opened 2023-09-14
Speed dialog: interactive clip speed textbox


saintmatthieu:
### Your idea

As in [this design](https://www.figma.com/file/3vAMBYovsJuTrbet34ruaZ/08_WX---Clips?type=design&node-id=1104-27663&mode=design&t=iXpPrnx6bJ9SWhWt-0), add a "Change Speed..." entry to the clip menu, which pops-up a "Clip Speed" window. As first step this window only has the "Clip Speed" textbox that allows textual setting of the speed.

This should stretch by changing the position of the clip's right edge.

If a clip is slowed down and collides with other following clips, trimming should occur.

### Problem to be solved

Allow users to stretch to precise percentages using a textbox.

chinakov:
> If a clip is slowed down and collides with other following clips, trimming should occur.

<img width="1501" alt="Screenshot 2023-09-26 at 13 20 50" src="https://github.com/audacity/audacity/assets/139755141/c43ae855-e523-4c1d-99b9-531edf1e1673">

Currently, if a User stretches a clip so that it collides with other following clips, no trimming occurs - User is shown an error message shown in the screenshot.

crsib:
> If a clip is slowed down and collides with other following clips, trimming should occur.

@LWinterberg was it something agreed on? This is not in spec and probably the weirdest behavior I can imagine :-)

saintmatthieu:
Sorry, I forgot about that requirement and haven't synced properly with Design.
This was the result of a discussion with @Tantacrul and I. Don't know what to do with it now, I suppose it's to be negotiated between @crsib and Design 🤷‍♂️

DavidBailes:
> Sorry, I forgot about that requirement and haven't synced properly with Design. This was the result of a discussion with @Tantacrul and I. Don't know what to do with it now, I suppose it's to be negotiated between @crsib and Design 🤷‍♂️

I don't think that trimming is very user friendly.
1. It could occur off-screen, and not be obvious visually when the user looked at the end of the clip that had been trimmed.
2. If a user reduced the speed of a clip which resulted in trimming, if they then increased the speed, there would be no adjustment to the amount of trimming - and again this could be all off-screen.
3. Keyboard users still can't un-trim a clip, so it's not at all helpful for these users.

chinakov:
I would also agree with trimming not being very user friendly, and it basically constitutes as a destructive action with no warnings. 

A warning about not enough space to 'stretch' is much more User friendly. 

[...]

PR #5239, opened 2023-09-25
5196 clip menus

crsib:
Features implemented in this PR:

Resolves: #5196
Resolves: #5238

Bugs additionally fixed:

Resolves: #2884 
Resolves: #5236

According to the agreement with QA, these issues are addressed in a single PR, not in 4 PRs, since they are closely related.

Recommended:


crsib:
@LWinterberg @DilsonsPickles quick design review won't hurt :-)

Paul-Licameli:
I only skimmed this PR.

I have mentioned that the Pan and Gain sliders show a dialog when you double-click on the slider thumb and that maybe that code could be reused, but I see this didn't happen.

Maybe they should be unified, replacing that old stuff with the new dialog, if it is in some way better?

crsib:
> I have mentioned that the Pan and Gain sliders show a dialog when you double-click on the slider thumb and that maybe that code could be reused, but I see this didn't happen

In the original design, this dialog was quite a bit more complex:

1. It had "Apply Destructively" button, which is now a menu entry instead
2. There was a check box "Adjust clip when changing project tempo", but it is not a priority now. So, probably the dialog will be extended even for 3.4, if UX testing proves that this feature is important. 

I would've done it using the dialog used now, but there was some miscommunication with the design team and I have implemented the "Apply Destructively" button before I was told it should be done differently.

LWinterberg:
Some feedback: 

* Render Clip Stretching should be greyed out when the clip isn't stretched yet
* When selecting partial or multiple stretched clips, the render option appears inoperable. In the case of multiple selected clips, I'd expect all to be rendered, in the case of a partial selection, I'd expect it to apply the stretching within the selection. As a fallback, having them greyed out when it can't apply works as well.

crsib:
> When selecting partial or multiple stretched clips, the render option appears inoperable

This is the same issue as with the "rename clip", I'm happy to discuss it after the SU

saintmatthieu:
Really looking forward to enjoying all this on master!

saintmatthieu:
If I have clips A and B one next to another, and slow-down A, I'd expect B (and all subsequent clips) to be shifted backwards if the "Editing a clip can move other clips" preference is checked, right? Currently I get the not-enough-room message.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -367,7 +367,8 @@ SelectedIntervalOfFocusedTrack(AudacityProject &project)
             return interval->Start() == viewInfo.selectedRegion.t0() &&
                interval->End() == viewInfo.selectedRegion.t1();
          });
-         if(it != ChannelGroup::IntervalIterator<WaveTrack::Interval> { })
+
+         if(it != intervals.end())
```
vsverchinsky, line 371: Thanks for catching this!

crsib:
Two minor commits were added @vsverchinsky

saintmatthieu:
Hi!
Trying to change a clip's speed to 200%, nothing happens.

saintmatthieu:
I see there's a shortcut to rename clips, I'm jealous `:P` Could we get one for speed changes, too? Not necessarily in this PR, although that'd be convenient for testing. @LWinterberg

crsib:
I think you can assign the shortcut in preferences

saintmatthieu:
I'm getting weird results: 

image

then press enter and the clip gets 600%: 
image

saintmatthieu (changes requested):
I haven't looked at the code yet but I have identified a problem while testing: https://github.com/audacity/audacity/pull/5239#issuecomment-1735004644

crsib:
I will take a look.

src/tracks/playabletrack/wavetrack/ui/ChangeClipSpeedDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+      return false;
+   }
+
+   const auto nextClip = mTrack.GetNextInterval(mTrackInterval, PlaybackDirection::forward);
+   const auto maxEndTime = nextClip != nullptr ?
+                              nextClip->Start() :
+                              std::numeric_limits<double>::infinity();
+
+   const auto expectedEndTime = mTrackInterval.End() * mOldClipSpeed / mClipSpeed;
```
saintmatthieu, line 136: I think this is the reason for the problem I found. It should be something like `interval.Begin() + (interval.End()-interval.Begin())*mOldSpeed/mSpeed;`

crsib: Yes, you are totally right

src/tracks/playabletrack/wavetrack/ui/ChangeClipSpeedDialog.h
```diff
@@ -0,0 +1,38 @@
...
+      wxWindow* parent = nullptr);
+
+   ~ChangeClipSpeedDialog() override;
+
+private:
+   void PopulateOrExchange(ShuttleGui& s);
+
+   void OnOk();
+
+   bool ApplyStretchRatio();
```
saintmatthieu, line 32: `ApplyStretchRatio` has so far meant rendering time-stretching. It took me some time to realize, though, that this is not this method's intent. Please consider something else, something like `DoChangeClipSpeed` ?

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -941,6 +941,9 @@ class WAVE_TRACK_API WaveTrack final
       void StretchLeftTo(double t);
       void StretchRightTo(double t);
 
+      void ApplyStretchRatio(const std::function<void(double)>& reportProgress);
```
saintmatthieu, line 944: Reusing the `ProgressReporter` alias would be better for consistency.

crsib: I haven't used it ... for consistency sake :-)

I was just copying the WaveClip interface, where ApplyStretchRatio is defined as

```c++
   /*!
    * @brief Renders the stretching of the clip (preserving duration).
    * @post GetStretchRatio() == 1
    */
   void ApplyStretchRatio(
      const std::function<void(double)>& reportProgress);
```

src/tracks/playabletrack/wavetrack/ui/WaveClipAdjustBorderHandle.cpp
```diff
@@ -256,7 +259,12 @@ class AdjustClipBorder final : public WaveClipAdjustBorderHandle::AdjustPolicy
       const auto dt = std::abs(mInitialBorderPosition - mBorderPosition);
       if (dt != 0)
       {
-         if (mAdjustingLeftBorder)
+         if (mIsStretchMode)
+         {
+            PushClipSpeedChangedUndoState(
```
saintmatthieu, line 264: Thank you 😬

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Oh really 😅 Well, do as you please, maybe I'll fix it up in some future PR.

saintmatthieu:
First review iteration complete. Please let me know when the speed-change not taking clip offset problem has a fix.
As for the possibility of moving clips right when expanding, I hope @LWinterberg gets back to us soon, but I won't block QA if he doesn't.

saintmatthieu (approved):

saintmatthieu:
Oh well, it's an easy fix, let's prevent disturbance during my pizza break :D

crsib:
A fix is added, rebased on the latest master unchanged otherwise

chinakov:
Just an update on how 'Render clip stretching' works with multiple stretched clips and selection within stretched clips. 

If a User has selected multiple stretched clips, and then accesses the Clip Menu - only one of the clips will be selected. 

If a User has selected a region within a clip, but not the entire clip, the User is forbidden from pressing 'Render clip stretching'.

Hope this helps, somehow.

chinakov:
@LWinterberg - please advise if this is fine, and this functionality can be considered to be behaving correctly.

chinakov:
Advised with @crsib - he has said that the acceptance criteria is fulfilled as defined. Moving to Ready to Merge. Thanks!

chinakov:
Checks for this MR has been completed, you can merge it!

Issue #5207, opened 2023-09-18
Loop region preserves beat length across tempo changes


saintmatthieu:
Currently, changing project tempo changes the loop beat duration and preserves the time duration.
Since looping is mostly a music feature, it makes most sense to preserve beat duration, e.g., if the loop spans 4 bars at tempo 50, it should still span 4 bars when changing to any other tempo.

PR #5240, opened 2023-09-25
Play region preserves beat length across tempo changes

vsverchinsky:
Resolves: #5207

Recommended:

### DIFF
--- a/src/ProjectTempoListener.cpp
+++ b/src/ProjectTempoListener.cpp
@@ -6,6 +6,8 @@
 
 #include <cassert>
 
+#include "ViewInfo.h"
+
 class ProjectTempoListener final : public ClientData::Base
 {
 public:
@@ -13,6 +15,9 @@ public:
    void OnProjectTempoChange(double newTempo);
 
 private:
+   AudacityProject& mProject;
+   ViewInfo& mViewInfo;
+   double mTempo { 0 };
    TrackList& mTrackList;
    Observer::Subscription mTrackListSubstription;
    Observer::Subscription mProjectTimeSignatureSubscription;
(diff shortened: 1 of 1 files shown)


src/ProjectTempoListener.cpp
```diff
@@ -53,4 +59,15 @@ void ProjectTempoListener::OnProjectTempoChange(double newTempo)
 {
    for (auto track : mTrackList)
       track->OnProjectTempoChange(newTempo);
+   
+   auto& viewInfo = ViewInfo::Get(mProject);
+   if(!viewInfo.playRegion.Empty() && mTempo > 0 && newTempo > 0)
```
saintmatthieu, line 64: Can `mTempo` possibly still be zero?

src/ProjectTempoListener.cpp
```diff
@@ -27,12 +31,14 @@ static const AttachedProjectObjects::RegisteredFactory key {
 
 ProjectTempoListener::ProjectTempoListener(
    AudacityProject& project, TrackList& trackList)
-    : mTrackList { trackList }
+    : mProject { project }
+    , mTempo { ProjectTimeSignature::Get(project).GetTempo() }
```
saintmatthieu, line 35: Minor: To be consistent with the way `mTrackList` is obtained, I'd have opted for adding a `mViewInfo` member variable and called `ViewInfo::Get` from the outside.

saintmatthieu (approved):
NIce, thanks!

src/ProjectTempoListener.cpp
vsverchinsky: I don't think that this is the possible case right now, but won't harm to check that. Non-positive tempo doesn't make sense, right?

src/ProjectTempoListener.cpp
vsverchinsky: Good idea

src/ProjectTempoListener.cpp
saintmatthieu: Sure, I just wanted to know if you were aware of a situation where it wasn't.

Issue #5126, opened 2023-09-01
Ramp-up at the beginning of a time-stretched clip


dozzzzer:
### Your idea

It'd be nice to further improve our time-stretching algorithm so that it doesn't cause a ramp-up when applied on an audio clip (especially in case of shrinking).

### Problem to be solved

Ramp-up, or fade-in, of approximately 35 milliseconds on a clip that has been shrunk or expanded (in the latter it's much less obvious).

### On a shrunk clip:

<img width=450 src=https://github.com/audacity/audacity/assets/70063970/6fe2db8e-b11d-470c-a043-594a7a254ede>

### On an expanded clip:

<img width=450 src=https://github.com/audacity/audacity/assets/70063970/96ba486a-7f0c-44d1-8701-b60e86e7522f>

saintmatthieu:
My guess as to what is happening here:

STFT-based algorithms such as time stretching work by splitting the audio into overlapping windows, and one-by-one transforming then to the spectrum domain, modifying them, and then transforming back to the time domain. If no modification happens, then the process is said to be transparent, that is, you get the exact same input.
However, if there _is_ modification, then after transformation back to the time domain, some of the signal in window B may leak a bit in window A and C, which is normal.
I think what we saw in this example is head leakage. Since, in our integration of the algorithm, we _do_ discard the reported delay, this fading-in should not be part of the output. So I suspect the reported delay not to be accurate.
That said, the clip in this example was shrunk by a factor of 6, which is relatively much already for standard music use cases.

dozzzzer:
Fixed by https://github.com/audacity/audacity/pull/5241

PR #5241, opened 2023-09-25
Improve delay estimation in time stretcher

### LINKED ISSUE #5206: Improve delay estimation in time-stretcher
### Your idea

Improve delay estimation of time-stretcher to reduce the risk of clicking and loss of data at beginning or end of clip.
If that turns out to be tricky, we could consider evaluating the true inserted delay based on experiment and account for this when we discard leading samples,
```cpp
   const auto latencySamples = mTimeAndPitch->getLatencySamples();
   auto numOutputSamplesToDiscard = static_cast<int>(latencySamples * mTimeRatio + 0.5); // here
```
In theory, this won't be sufficient to be safe against clip-boundary discontinuities after stretch-rendering, that's just the maths of it. Cross-fading can be considered in the future if it still turns out to be a nuisance.

### Problem to be solved

When expanding a clip, rendering of the time stretching introduces a negative delay, i.e., the result looks ahead of expected:
image

When shrinking a clip, it's the other way round:
image

Besides compromising the beat alignment of multi-track projects, this also brings the risk of truncating important audio material, especially when the delay is negative, since the beginning of music loops is often a note's onset and ugly clicks can be introduced.

### Prior art

We should ask StaffPad's dev team to see if they have the problem and how they deal with it. AFAICS, they use the return value of `TimeAndPitch::getLatencySamples()` as-is.

### Additional context

Likely will reduce the observed ramping-in reported by #5126.

saintmatthieu:
Resolves: #5206

See #5206 for a detailed description.

Please note that this PR also undoes the formatting that took place in the early integration days of the StaffPad library. This was done by copying the library again as-is and cherry-picking the few changes that followed.

Recommended:

QA:
- [x] Rendering of shrunk or expanded clips to a new track now shows better alignment. (Please test for both mono and stereo.)
- [x] This ticket should alleviate the problem reported in #5126, or maybe even fix it completely? Please close if that's the case.

Also please refer to the detailed description in #5206 - this may give you testing ideas.


crsib (approved):

Paul-Licameli:
@crsib approved but I really want to take a deeper look before this gets merged.  However @dozzzzer can surely proceed to try it out to see of the small time shift is fixed.

Paul-Licameli:
But what is NOT in scope of this is the experiment with higher order cosine windows that might reduce the problem of pre-echoes of transients.

saintmatthieu:
@Paul-Licameli it really only is the last two commits that bring something new, and they are simple. Maybe double-checking that I haven't omitted any commit for cherry-pick after the un-formatting commit, but that'd be it, really.

Paul-Licameli:
A lot of whitespace changes, just so that we don't have gratuitous differences from the upstream of the library.  The modifications we made before to it are re-applied in this commit sequence.  Some of them have my name on them.

saintmatthieu:
> But what is NOT in scope of this is the experiment with higher order cosine windows that might reduce the problem of pre-echoes of transients.

It isn't part of this PR either. That was a provisional experimental state of this branch before I opened the PR. You shouldn't see a trace of this anymore.

dozzzzer:
Rendering expanded/shrunk clips now indeed shows better alignment than previously, both for mono and stereo audio.

### Shrinking

#### Before:

before (shrunk)

#### After:

after (shrunk)

### Expansion

#### Before:

before (expanded)

#### After:

after (expanded)

Also, as far as the ramp-up issue described in #5126, this PR also fixes this, both when stretching and shrinking:

Screenshot 2023-09-26 at 16 03 05

Thanks, @saintmatthieu, great job!

PR #5243, opened 2023-09-26
Consistensy checks added to AUP importer

vsverchinsky:
Resolves: #5211
Resolves: #5304

Recommended:

QA:  Might want to wait until this is rebased onto #5260 
   - [x] Save 2.4.2 projects with misaligned stereo, mismatched channel formats, mismatched channel rates
   - [x] Open them in 3.4, seeing similar warnings and same splitting, as when opening 3.x legacy projects with similar
   - [ ] 2.4.2 projects without those problems open with no warnings

### DIFF
--- a/modules/mod-aup/ImportAUP.cpp
+++ b/modules/mod-aup/ImportAUP.cpp
@@ -318,7 +318,6 @@ void AUPImportFileHandle::Import(ImportProgressListener& progressListener,
    auto &history = ProjectHistory::Get(mProject);
    auto &tracks = TrackList::Get(mProject);
    auto &viewInfo = ViewInfo::Get(mProject);
-   auto &settings = ProjectSettings::Get(mProject);
    auto &selman = ProjectSelectionManager::Get(mProject);
 
    auto oldNumTracks = tracks.Size();
@@ -351,10 +350,12 @@ void AUPImportFileHandle::Import(ImportProgressListener& progressListener,
       progressListener.OnImportResult(ImportProgressListener::ImportResult::Error);
       return;
    }
-   else if(!mErrorMsg.empty())//i.e. warning
+   if(!mErrorMsg.empty())//i.e. warning
+   {
       ImportUtils::ShowMessageBox(mErrorMsg);
+      mErrorMsg = {};
+   }
    
-   // TODO wide wave tracks -- quit here if misaligned tracks are found.
    // (If we keep this entire source file at all)
 
    sampleCount processed = 0;
@@ -394,6 +395,33 @@ void AUPImportFileHandle::Import(ImportProgressListener& progressListener,
 
    for (auto pClip : mClips)
       pClip->UpdateEnvelopeTrackLen();
+
+   ProjectFileManager::FixTracks(
+      tracks,
+      [&](const auto& errorMessage) { SetError(errorMessage); },
+      [&](const auto& unlinkReason) { SetWarning(XO(
+//i18n-hint: Text of the message dialog that may appear on attempt
+//to import an AUP project.
+//%s will be replaced with an explanation of the actual reason of
+//project modification.
+"%s\n"
+"This feature is not supported in Audacity versions past 3.3.3.\n"
+"These stereo tracks have been split into mono tracks.\n"
+"Please verify that everything works as intended before saving.")
+                     .Format(unlinkReason));
+      }
+   );
+
+   if(mHasParseError)
+   {
+      progressListener.OnImportResult(ImportProgressListener::ImportResult::Error);
+      return;
+   }
+   if(!mErrorMsg.empty())
+   {
+      ImportUtils::ShowMessageBox(mErrorMsg);
+      mErrorMsg = {};
+   }
 
    // If the active project is "dirty", then bypass the below updates as we don't
    // want to going changing things the user may have already set up.
@@ -557,6 +585,9 @@ void AUPImportFileHandle::HandleXMLEndTag(const std::string_view& tag)
    {
       node.handler->HandleXMLEndTag(tag);
    }
+
+   if (tag == "wavetrack")
+      mWaveTrack->SetLegacyFormat(mFormat);
 
    mHandlers.pop_back();
 
--- a/src/ProjectFileManager.cpp
+++ b/src/ProjectFileManager.cpp
@@ -61,6 +61,7 @@ Paul Licameli split from AudacityProject.cpp
 #include <optional>
 
 #include "RealtimeEffectList.h"
+#include "tracks/playabletrack/wavetrack/WaveTrackUtils.h"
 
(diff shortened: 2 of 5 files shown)


modules/mod-aup/ImportAUP.cpp
```diff
@@ -557,6 +557,21 @@ void AUPImportFileHandle::HandleXMLEndTag(const std::string_view& tag)
    {
       node.handler->HandleXMLEndTag(tag);
    }
+   if(tag == "wavetrack")
```
Paul-Licameli, line 560: Commit comment:  "consistency"

modules/mod-aup/ImportAUP.cpp
```diff
@@ -557,6 +557,21 @@ void AUPImportFileHandle::HandleXMLEndTag(const std::string_view& tag)
...
+      {
+         if(auto err = first->GetErrorOpening())
+            SetError(*err);
+         if(!first->LinkConsistencyFix(true))
+            SetError(XO("A channel of a stereo track was missing."));
+
+         if(mHasParseError)
+            return;
+      }
+   }
```
Paul-Licameli, line 574: I'm convinced that this is correct to repair the track consistency, but there is also new logic in `ProjectFileManager::ReadProjectFile` to warn the user about the change, and that function is not reached.

It should not be difficult to make that reusable.

Paul-Licameli (dismissed):
Should give the user information about fixes, just as if upgrading an .aup3 project from 3.x

modules/mod-aup/ImportAUP.cpp
vsverchinsky: My guess that the main reason to show this dialog is to warn about possible effect stack loss, when you import AUP with mismatched channels nothing gets removed

Paul-Licameli: But you are also changing it in #5260 to give other warnings about mismatched rates and formats

vsverchinsky: The warning is about effect stacks that can be removed as we split channels, we don't change sample format or rate

vsverchinsky: And existing texts do not fit in that case I think

LWinterberg: the warning contains two pieces of information: 
1. your project will look different to what you last saved because we migrated some stereo tracks to mono tracks
2. realtime effect stacks may be lost. 

In the case of legacy importers, part 2 is not relevant but part 1 is, so I'd still want to see the message box.

src/tracks/playabletrack/wavetrack/WaveTrackUtils.cpp
```diff
@@ -11,10 +11,23 @@
...
+WaveTrackUtils::GetChannelUnlinkReason(WaveTrack& left, WaveTrack& right)
+{
+   if(left.GetRate() != right.GetRate())
+      //i18n-hint: explains why opened project was auto-modified 
+      return XO("This project contained stereo tracks with different sample rates per channel.");
+   if(left.GetSampleFormat() != right.GetSampleFormat())
+      //i18n-hint: explains why opened project was auto-modified  
+      return XO("This project contained stereo tracks with different sample formats in channels.");
+   //i18n-hint: explains why opened project was auto-modified 
+   return XO("This project contained stereo tracks with non-aligned content.");
```
Paul-Licameli, line 32: Strings, @LWinterberg  ?

This function causes only the first applicable message to be shown.  Possibly more that one might apply in an old project.  Maybe all applicable messages should be shown.  Or maybe we don't care about this edge case of an edge case.

src/ProjectFileManager.cpp
```diff
@@ -196,15 +197,8 @@ auto ProjectFileManager::ReadProjectFile(
...
-                     //i18n-hint: explains why opened project was auto-modified 
-                     linkTypeChangeReason = XO("This project contained stereo tracks with different sample rates per channel.");
-                  else if(left->GetSampleFormat() != right->GetSampleFormat())
-                     //i18n-hint: explains why opened project was auto-modified  
-                     linkTypeChangeReason = XO("This project contained stereo tracks with different sample formats in channels.");
-                  else
-                     //i18n-hint: explains why opened project was auto-modified 
-                     linkTypeChangeReason = XO("This project contained stereo tracks with non-aligned content.");
+                  if(!linkTypeChangeReason)
+                     linkTypeChangeReason = WaveTrackUtils::GetChannelUnlinkReason(*left, *right);
```
Paul-Licameli, line 201: However, I see that in case of opening a 3.x file, not importing .aup, the warning about possible loss of stacked effects is always given, in addition to whatever other message is reported by the function.  That is unchanged behavior compared with master.

modules/mod-aup/ImportAUP.cpp
```diff
@@ -351,10 +352,12 @@ void AUPImportFileHandle::Import(ImportProgressListener& progressListener,
...
    }
-   else if(!mErrorMsg.empty())//i.e. warning
+   if(!mErrorMsg.empty())//i.e. warning
+   {
       ImportUtils::ShowMessageBox(mErrorMsg);
+      mErrorMsg = {};
+   }
    
-   // TODO wide wave tracks -- quit here if misaligned tracks are found.
    // (If we keep this entire source file at all)
```
Paul-Licameli, line 361: Delete the second line of the comment too

modules/mod-aup/ImportAUP.cpp
```diff
@@ -395,6 +398,72 @@ void AUPImportFileHandle::Import(ImportProgressListener& progressListener,
...
+         //Wait when LinkConsistencyFix is called on the second track
+         unlinkedTrack = t;
+      }
+
+      if (const auto message = t->GetErrorOpening())
+      {
+         SetError(*message);
+         break;
+      }
+   }
```
Paul-Licameli, line 454: This looks correct but it's all duplication of a part of `ProjectFileManager::ReadProjectFile`.  Can there be just one function to do this?

Paul-Licameli:
I dismissed my request for changes, but before approval I will await an answer from @LWinterberg about a new question, and maybe @vsverchinsky will eliminate duplicated logic.

Also, correct the spelling in the commit comment please.

src/tracks/playabletrack/wavetrack/WaveTrackUtils.cpp
LWinterberg: strings look good!

vsverchinsky:
Corrected the spelling and duplicated logic extracted.
Also loop that fixes consistensy has changed a bit - now it continues to remove effect stacks from unlinked tracks even if error has been found.

Paul-Licameli (approved):
Good to see elimination of duplicated logic

PR #5250, opened 2023-09-26
5124 improve time stretching quality for high sample rates

### LINKED ISSUE #5124: Improve time-stretching quality for high sample rates
### Your idea

The StaffPad time stretcher uses a fixed FFT size (4096, with overlap 4). For the usual 44.1kHz, this is roughly 90ms.
For higher sample rates, such as 88.2 or 196.4 kHz, this duration gets inversely proportionally smaller, and lower-pitched contents do not get analyzed properly, leading to ugly garbling.

### Problem to be solved

The idea to explore is to make the FFT size dependent on the sample rate, so as to preserve this 90ms analysis duration. On could even experiment with longer windows, considering that competitor software Elastique has a latency of 150ms, which suggests an FFT size of 8096 samples at 44.1kHz. (The latency is not an issue for our current use of time-stretching, it only becomes an issue when real-time performance comes into play, which could be the case if we wanted to use the pitch-shifting capability of SP's algorithm.)

Stretch goal: Something else to consider would even be to use an overlap of 3, which would reduce the computational demand by 25% and reduce the latency by as much. If the algorithm can work with a 1st-order cosine window (such as a Hann or Blackman window) it can be proven that an overlap of 3 is sufficient to preserve transparency of an STFT. Only it's a bit more complex, since it requires zero-padding, and one would also have to make sure that the quality doesn't suffer from it.

### Prior art

_No response_

### Additional context

_No response_

saintmatthieu:
Resolves: #5124

Please refer to detailed description in #5124 

Recommended:

QA:
- [x] Get an audio file of sample rate 88.2kHz or higher, with harmonically rich and low-frequency content. (Feel free to resample an audio file of 44.1 or 48kHz, or to use [this sample](https://1drv.ms/u/s!ArATSZ9esLBGx062RpVHSYiGBs0m?e=8BgGsj)) Stretch the audio in either direction. This should sound better with this build than on current master.

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -43,6 +43,17 @@ inline float lagrange6(const float (&smp)[6], float t)
   return (c0 + c1 * t) + (c2 + c3 * t) * t2 + (c4 + c5 * t) * t2 * t2;
 }
 
+inline int getFftSize(int sampleRate)
+{
+  // 44.1kHz maps to 4096 samples (i.e., 93ms).
+  // We grow the FFT size proportionally with the sample rate to keep the
+  // window duration roughly constant, with quantization due to the
+  // power-of-two constraint.
+  // If needed some time in the future, we can decouple analysis window and
+  // FFT sizes by zero-padding, allowing for very fine-grained window duration
+  // without compromising performance.
+  return 1 << (12 + (int)std::round(std::log2(sampleRate / 44100.)));
+}
 } // namespace
 
 struct TimeAndPitch::impl
@@ -75,6 +86,11 @@ struct TimeAndPitch::impl
 
   std::vector<int> peak_index, trough_index;
 };
+
+TimeAndPitch::TimeAndPitch(int sampleRate)
+    : fftSize(getFftSize(sampleRate))
+{
+}
 
 TimeAndPitch::~TimeAndPitch()
 {
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.h
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.h
@@ -11,7 +11,7 @@ namespace staffpad {
 class TimeAndPitch
 {
 public:
-  TimeAndPitch() = default;
+  TimeAndPitch(int sampleRate);
   ~TimeAndPitch();
 
   /**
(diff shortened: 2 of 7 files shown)


saintmatthieu:
@crsib @Paul-Licameli @vsverchinsky I added three reviewers but it should be an easy PR that only one can review, whoever does it first.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -50,7 +50,14 @@ inline float lagrange6(const float (&smp)[6], float t)
...
 {
-   return 4096;
+   // 44.1kHz maps to 4096 samples (i.e., 93ms).
+   // We grow the FFT size proportionally with the sample rate to keep the
+   // window duration roughly constant, with quantization due to the
+   // power-of-two constraint.
+   // If needed some time in the future, we can decouple analysis window and
+   // FFT sizes by zero-padding, allowing for very fine-grained window duration
+   // without compromising performance.
+   return 1 << (12 + (int)std::round(std::log2(sampleRate / 44100.)));
```
crsib, line 60: `static_cast`?

crsib (approved):
Couple style suggestions, looks fine otherwise

dozzzzer:
Tested

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.h
```diff
@@ -11,7 +11,7 @@ namespace staffpad {
 class TimeAndPitch
 {
 public:
-  TimeAndPitch() = default;
+  TimeAndPitch(int sampleRate);
```
Paul-Licameli, line 14: `explicit`?

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.h
```diff
@@ -12,7 +12,7 @@ namespace staffpad
 class TIME_AND_PITCH_API TimeAndPitch
 {
 public:
-   TimeAndPitch() = default;
+   TimeAndPitch(int sampleRate);
```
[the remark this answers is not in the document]
saintmatthieu: I missed that one, sorry.

Paul-Licameli:
I reviewed too.

I question why this is important for 3.4.  This is motivated by a theoretical case that seem unlikely in practice.

It's about sound quality in case of high sample rates -- but typical human hearing just can't detect much above 20kHz.

The point of this PR is to preserve the frequency resolution, but at high rates the result is less time resolution.  Maybe you will discover transients aren't treated as well.

This could be tested also with lower sample rates.

PR #5260, opened 2023-09-27
Unlink mismatched channels in wave tracks

vsverchinsky:
Problem description: user cannot open project that has stereo tracks with mismatched sample format/rate.
Solution: unlink channels, preserving rate/format for individual channel and remove the effect stack. Changes should not be applied unless project is explicitly saved by user, user should be warned about changes.

Recommended:

QA list is similar to one in #5202, but here mismatch can be either by clip boudaries, sample rate or sample format, or any combination of them (but it will be enough to check each of them separately):
 - [x] Verify opening of a properly saved project that contains mismatches:
   - [x] Shows warning dialog with the message explaining the reason to split tracks
   - [x] Stereo wave tracks that has mismatches are split into two mono tracks
   - [x] These tracks have empty effects stack
   - [x] Splited tracks preserve their initial rate and sample format
   - [x] Other tracks are not affected (mono wave tracks, label tracks etc.)

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -702,8 +702,11 @@ bool WaveTrack::LinkConsistencyFix(bool doFix)
       }
       else if (doFix) {
          // non-error upgrades happen here
-         if (!AreAligned(SortedClipArray(), next->SortedClipArray()))
+         if (!AreAligned(SortedClipArray(), next->SortedClipArray()) ||
+             !RateConsistencyCheck() || !FormatConsistencyCheck())
+         {
             SetLinkType(LinkType::None);
+         }
          else
          {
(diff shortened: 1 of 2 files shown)


Paul-Licameli (approved):

Paul-Licameli:
Approved.  I assume this PR should be on the main board with 3.4 milestone?  I made it so.

chinakov:
@vsverchinsky has mentioned that he'd like to advise someone in regards to the splitting of tracks to L-R channels, waiting for info regarding this, and then will close this ticket.

@vsverchinsky suggests that the split stereo tracks should be sent to Left and Right channels respectively on each resulting mono channel.

LWinterberg:
Yes, the resulting mono tracks should be hard panned L/R

chinakov:
Yes, the panning works now. Thanks!

Issue #5263, opened 2023-09-27
Don't show the error message when pressing Ctrl + C with no audio selected


dozzzzer:
### Your idea

Don't show the error message when no audio is selected and when Ctrl (Cmd) + C is pressed.

### Problem to be solved

https://github.com/audacity/audacity/assets/70063970/bf33e324-fbee-4d72-ae9d-3be8b1d73c91

petersampsonaudacity:
I do note that we already do _**not**_ show an error message in respons to Ctrl+C (Cmd+C) 
when there is nothing on the Audacity clipboard.

vsverchinsky:
> I do note that we already do _**not**_ show an error message in respons to Ctrl+C (Cmd+C) when there is nothing on the Audacity clipboard.

Do you have this option checked?
![изображение](https://github.com/audacity/audacity/assets/20165230/caa4e610-1a3f-46be-a0b3-856884cd95d2)

petersampsonaudacity:
@vsverchinsky  

No I never have that option checked.

I particularly **_dislike_** that functionality - _it used to be  "on" by default and it took a lot of persuading of Gales Andrews (RIP) to get it turned the other way around_

chinakov:
No error message is shown when 'CTRL + C' is pressed with no audio selected. 
Tested, ready to merge, thanks!

PR #5277, opened 2023-09-28
Edit commands conditionally greyed out

vsverchinsky:
Resolves: #5263 

Copy does not support auto-select option any more
Cut, Copy commands are made available only when audio data or label text selected
Paste command is made available when clipboard not empty

Recommended:


Paul-Licameli (approved):
I approve, but I think the commit comment does not correctly state all of the intent?

Should it say Cut (not Copy) no longer auto-selects?

PR #5288, opened 2023-09-29
Rewrite time stretching with one fewer array and without std::abs

Paul-Licameli:

QA:  Try some time stretching and shrinking examples in master and in this branch.  Do the invert and mix test and get near-zero differences.  Use the Contrast analyzer.  I don't expect exact zero, but in one example I tried, differences were less than -140 dB.

Observe speed improvement.

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -119,7 +119,7 @@ void TimeAndPitch::setup(int numChannels, int maxBlockSize)
 
   // fft coefficient buffers
   d->spectrum.setSize(_numChannels, _numBins);
-  d->norm.setSize(_numChannels, _numBins);
+  d->norm.setSize(1, _numBins);
   d->last_norm.setSize(1, _numBins);
   d->phase.setSize(_numChannels, _numBins);
   d->last_phase.setSize(_numChannels, _numBins);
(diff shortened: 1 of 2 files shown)


libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -70,14 +70,13 @@ struct TimeAndPitch::impl
 
   SamplesReal fft_timeseries;
   SamplesComplex spectrum;
-  SamplesReal mag;
+  SamplesReal norm;
```
saintmatthieu, line 73: Hi,
What motivated this change of name? At first sight I don't like this. "Magnitude" was correct and usual in DSP.

saintmatthieu: ... is this actually a mathematical shortcut to avoid `abs` ? Would you care to explain ?

crsib: `std::abs` uses `hypotf` which was a significant bottleneck on Windows. It was rewritten in a very straightforward way using SIMD. I haven't measured it on different compilers, so I can say for sure if this change is much needed.

saintmatthieu: Did you mean you _can't_ say for sure ?

crsib: Yep, sorry. I'm just about rebasing these commits on top of my branch, but judging by the time tests take, I don't expect any significant difference.

Paul-Licameli:
`std::norm` is now used instead of `std::abs` and so I made changes of variable names throughout.

The algorithm would computes the same thing using the ideal mathematical functions, but experiment showed tiny differences below -140 dB.

Besides the polar to cartesian step, the magnitudes were used in comparisons only, to identify peaks in the spectra, and that was used.  So the norm, which varies monotonically with the magnitude, can be used just as well.

Then instead of polar to cartesian, the original cartesian coordinates are simply rotated in the complex plane by the difference between old and new phases.

Not only does this eliminate the many square root calculations in computing polar coordinates, but also, the norms of the side channel are not needed at all.

Observe too the elimination of an an unnecessary temporary array and the copies into it.

saintmatthieu:
Looks like a very smart way of avoiding the need of `abs`. There's yet one thing I'm unclear about that I'll add an inline comment for.

saintmatthieu (approved):
... nope, nothing unclear anymore, LGTM.
I would have kept the naming, and add a comment that `mag` isn't really `mag` to keep the diff smaller. But sure, there are also arguments for the renaming.
Very smart indeed.

Paul-Licameli:
This makes two things from me that I hope they take upstream

chinakov:
Tried many stretching scenarios. 

Stretched and shrunk complex audio on macOS and Windows, inverted one of the tracks and mixed and rendered to a new track. Never got a bigger difference than ~-155dB in the resulting 'silence' mix. 

Thank you, ready to merge.

PR #5289, opened 2023-09-29
Stretcher performance

crsib:
This is an attempt to improve the performance of the time stretcher on MSVC. The problem was noticed due to extreme duration required to run the time stretching test.

MSVC, especially when targeting 32 bit Windows has the following problems:

* `rintf` is not inlined. This function is approximately 2 CPU instructions. Replacing it with `int(x + 0.5f)` results in 2x performance improvement.
* Math functions are not inlined either. `void TimeAndPitch::_process_hop` first converts a large array of complex numbers (2049 were observed during the tests) to polar coordinates, manipulates the result and converts back to cartesian coordinates. This conversion turned out to be extremely slow on MSVC.

On top of that, Audacity build system only enabled SSE2 support for `SBSMS` and `soundtouch`.

As the result of this PR, the following improvement was observed on Ryzen 5800x:

* Win64: 938 seconds -> 215 seconds
* Win32: 5489 (!) seconds -> 354 seconds

This result is quite close to M1-powered MacBook Pro, where the test runs in around 200 seconds.

Checking the samples used for unit tests showed that the difference is below -90 dB.

### For QA

This PR changes how trigonometry functions are calculated: `sin`, `cos` and `atan`. The difference between the "reference" implementation and implementation is relatively small, however, the error accumulates. I have observed a difference of around -90 dB, I expect that something similar will be observed during the testing.

------------------

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -185,7 +185,8 @@ namespace {
 // wrap a phase value into -PI..PI
 inline float _unwrapPhase(float arg)
 {
-  return arg - std::rint(arg * 0.15915494309f) * 6.283185307f;
+  using namespace audio::simd;
+  return arg - rint(arg * 0.15915494309f) * 6.283185307f;
 }
 
 void _unwrapPhaseVec(float* v, int n)
--- a/cmake-proxies/cmake-modules/AudacityFunctions.cmake
+++ b/cmake-proxies/cmake-modules/AudacityFunctions.cmake
@@ -274,7 +274,12 @@ function( audacity_append_common_compiler_options var use_pch )
          $<IF:$<CONFIG:Debug>,-D_DEBUG=1,-U_DEBUG>
 
          $<$<PLATFORM_ID:Darwin>:-DUSE_AQUA_THEME>
+
+         # Enable SIMD support when available
+         ${MMX_FLAG}
+         ${SSE_FLAG}
    )
+
    # Definitions controlled by the AUDACITY_BUILD_LEVEL switch
    if( AUDACITY_BUILD_LEVEL EQUAL 0 )
       list( APPEND ${var} -DIS_ALPHA -DUSE_ALPHA_MANUAL )
@@ -400,7 +405,7 @@ function(collect_edges TARGET IMPORT_TARGETS LIBTYPE)
    endif()
 
    propagate_interesting_dependencies( ${TARGET} "${IMPORT_TARGETS}" )
- 
+
    append_node_attributes( ATTRIBUTES ${TARGET} )
 
    list( APPEND GRAPH_EDGES "\"${TARGET}\" [${ATTRIBUTES}]" )
--- a/libraries/lib-time-and-pitch/CMakeLists.txt
+++ b/libraries/lib-time-and-pitch/CMakeLists.txt
@@ -12,6 +12,7 @@ set( SOURCES
    StaffPad/FourierTransform_pffft.cpp
    StaffPad/FourierTransform_pffft.h
    StaffPad/SamplesFloat.h
+   StaffPad/SimdComplexConversions_sse2.h
    StaffPad/SimdTypes.h
    StaffPad/SimdTypes_neon.h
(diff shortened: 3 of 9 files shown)


libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -187,7 +187,7 @@ namespace {
 // wrap a phase value into -PI..PI
 inline float _unwrapPhase(float arg)
 {
-  return arg - std::rint(arg * 0.15915494309f) * 6.283185307f;
+  return arg - int(arg * 0.15915494309f + 0.5f) * 6.283185307f;
```
saintmatthieu, line 190: `assert(std::rint(x) == int(x + 0.5f)` fails for values such as `x = -0.7`, i.e., expected is -1, actual is 0.
I think the following would work:
```
  const auto x = arg * 0.15915494309f;
  const auto rounder = x < 0.f && int(-x * 2) % 2 == 1 ? -0.5f : 0.5f;
  return arg - int(arg * 0.15915494309f + rounder) * 6.283185307f;
```
I wonder, though, how this mistake didn't lead to greater errors in your output comparisons 🤔 Maybe now you'd have bit-exactness ?

saintmatthieu: To convince myself I tried the following:
```
inline float _unwrapPhase(float arg)
{
  const auto x = arg * 0.15915494309f;
  const auto rounder = x < 0.f && int(-x * 2) % 2 == 1 ? -0.5f : 0.5f;
  return arg - int(arg * 0.15915494309f + rounder) * 6.283185307f;
}

inline float _unwrapPhaseNaive(float arg)
{
  const auto x = arg * 0.15915494309f;
  const auto rounder = 0.5f;
  return arg - int(arg * 0.15915494309f + rounder) * 6.283185307f;
}
```
and then
```
  assert(_unwrapPhase(twoPi * -0.2) == twoPi * -0.2f);
  assert(_unwrapPhase(twoPi * 0.2) == twoPi * 0.2f);
  assert(_unwrapPhase(twoPi * 0.7) == twoPi * 0.7f - twoPi);
  assert(_unwrapPhase(twoPi * -0.7) == twoPi - twoPi * 0.7f);

  assert(_unwrapPhaseNaive(twoPi * -0.2) == twoPi * -0.2f);
  assert(_unwrapPhaseNaive(twoPi * 0.2) == twoPi * 0.2f);
  assert(_unwrapPhaseNaive(twoPi * 0.7) == twoPi * 0.7f - twoPi);
  assert(_unwrapPhaseNaive(twoPi * -0.7) == twoPi - twoPi * 0.7f);
```
The last assert fails.

crsib: > Maybe now you'd have bit-exactness ?

No

crsib: You raise a valid point, though. I will make sure that _unwrapPhase satisfies the post-condition. This is not the same as the exact value though, `rint` results depend on the global state of the application.

crsib: I've pushed a commit that makes `_unwrapPhase` and `_unwrapPhaseVec` to use exactly the same set of instructions

saintmatthieu: > `rint` results depend on the global state of the application

Would that explain why your comparison tests only showed a -90dB difference, i.e., on your system it was doing the rounding properly ?

crsib: I'm not sure what you mean.

saintmatthieu:
I'd give this PR the thumbs up, but I'm incompetent to review the SSE optimization part, and I understood that this was going to be expected from @vsverchinsky.

vsverchinsky (approved):

crsib:
Thank you, @vsverchinsky! Your suggestions are now implemented

vsverchinsky (approved):

chinakov:
Tested and discussed with @crsib - we've come to the conclusion that the performance is satisfactory.

PR #5290, opened 2023-09-30
change time track default values to 0.2, 2.0

LWinterberg:
just makes time tracks a little bit more versatile by default


crsib (approved):
Dunno why we need it, but it's so minor, that probably not worth discussing :-)

dozzzzer:
Tested

PR #5291, opened 2023-09-30
unsquish AudacityLogoWithName

LWinterberg:
This unsquishes the Audacity text logo, as seen in the initial splash, the welcome dialog and the about dialog. 

NB: The scaling factor here has been reset to 1. It's trivial to scale things down in each dialog to have them appear less mighty. 

(also removes some darkaudacity stuff which was unfortunate enough to be in my way)

Incidentally, the new XPM is smaller than the previous logo, so this even reduces load times a little.


petersampsonaudacity:
Testing on W10 with audacity-win-3.4.0-alpha-20230930+102c263-x64

@LWinterberg   -  Ooh that looks nicer Leo  :-))

crsib (approved):

dozzzzer:
Tested

LWinterberg:
File in question is "\win\Inno_Setup_Wizard\audacity_InnoWizardImage.bmp". I'll fix it another time

PR #5308, opened 2023-10-04
Avoid frequent allocation of little vectors of pointers in stretch

Paul-Licameli:

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
@@ -10,15 +10,12 @@ namespace
 // to be specified in the `setup` call.)
 constexpr auto maxBlockSize = 1024;
 
-std::vector<float*>
-GetOffsetBuffer(float* const* buffer, size_t numChannels, size_t offset)
+void
+GetOffsetBuffer(float **offsetBuffer,
+   float* const* buffer, size_t numChannels, size_t offset)
 {
-   std::vector<float*> offsetBuffer(numChannels);
    for (auto i = 0u; i < numChannels; ++i)
-   {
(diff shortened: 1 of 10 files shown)


saintmatthieu:
Looks fine to me, but I'd refer to @crsib to verify that the changes are worthwhile.

crsib:
If we want to upstream changes - probably we probably shouldn't merge this PR. `calloc` is frowned upon, probably StaffPad guys won't like it either.

crsib (approved):
I see no performance improvement and I do have doubts that SP guys would want this change, but I have nothing to say against this PR

Paul-Licameli:
Rebased and redrafted without `alloca`

Please review again.

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
```diff
@@ -82,9 +79,9 @@ void StaffPadTimeAndPitch::GetSamples(float* const* output, size_t outputLen)
          const auto numSamplesToGet =
             std::min({ maxBlockSize, numOutputSamplesAvailable,
                        static_cast<int>(outputLen - numOutputSamples) });
-         const auto buffer =
-            GetOffsetBuffer(output, mNumChannels, numOutputSamples);
-         mTimeAndPitch->retrieveAudio(buffer.data(), numSamplesToGet);
+         float *buffer[2]{};
```
crsib, line 82: Some sanity checks that mNumChannels <= 2 won't hurt, `assert` will do

Paul-Licameli: Done

crsib (approved):

Issue #5270, opened 2023-09-28
Time controls with bars: keyboard interactions can produce completely wrong results


DavidBailes:
### Bug description

There are problems with keyboard interactions which cause the number of digits used for the bar number to change.

In the case of using up/down arrows, in general this should not cause which digit is the focus to change. However if the number of digits reduces and the focus was on the digit that is removed, then the focus has to change - to the next digit, I guess would be sensible.

Observation: If the current number of digits for the bar number is 3 then setting a 4 digit bar number using the keyboard is a bit of a fiddle.

### Steps to reproduce

All these cases are using the selection start control, with the format set to bar:beat

Case A:
1. Set the bar number to 999.
2. Set the focus to the last digit of 999.
3. Press up arrow twice.

Case B:
1. Set the bar number to 1000.
2. Set the focus to the last digit of 1000.
3. Press down arrow twice.

Case C:
1. Set the bar number to 1000.
2. Set the focus to the first digit of 1000.
3. Type the digits 0,9,6,1.

### Expected behavior

Case A:
The bar number changes to 1001.

Case B:
The bar number changes to 998.

Case C:
The bar number changes to 961

### Actual behavior

Case A:
The bar number changes to 2000. (After the first up arrow the focus is moved to the first of the now 4 digits.)

Case B:
The bar number changes to 899. (After the first up arrow, the focus is moved to the first of the now 3 digits.)

Case C:
The bar number changes to 124, and the beat number to 2. (After typing the 0, the focus is moved to the first digit of the beats.)

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

Introduced by this commit: https://github.com/audacity/audacity/pull/5190, which allows more than 3 digits for bar numbers.

crsib:
Expected results in case B are not in line with what Audacity would do now, I've made the behavior to match the current one

DavidBailes:
> Expected results in case B are not in line with what Audacity would do now, I've made the behavior to match the current one

@crsib Thanks for having a look at this. However I'm not sure what you mean by this. I think that if the focus is on the last digit of bar 1000, and the user presses down arrow twice, then the user would expect the bar number to change to 998? Are you suggesting this shouldn't be the case?

crsib:
Oh, shoot, the last digit. Sorry, @DavidBailes. Then it is fixed as well.

dozzzzer:
Tested

Issue #5232, opened 2023-09-22
Number of digits in the Selection Start should match the one of the Selection End when bars and beats is selected


dozzzzer:
### Bug description

In the Selection toolbar, when bar:beat:tick or bar:beat format is selected, the number of digits displayed for Selection Start is always three, despite in the lower part of the Selection toolbar, Selection End, it may exceed 999 bars, adding the 4th digit in the line.

### Steps to reproduce

1. Generate 1h of audio with the tempo of 120.0 bpm
2. Ctrl + A
3. Switch the Selection toolbar format to bar:beat:tick or bar:beat
4. Observe

### Expected behavior

The number of digits of the Selection Start perfectly matches the one of the Selection End

### Actual behavior

The Selection Start is one digit less shorter than the Selection End, which makes the top and the bottom misaligned:

![image](https://github.com/audacity/audacity/assets/70063970/524dc66b-fb5b-48ab-a0b5-a13ec46d46e8)

### Audacity Version

current master / alpha / nightly build

### Operating system

Other

rohitmit98:
Hello! That's an interesting observation. Please confirm I'm correctly identifying and reproducing this issue:
![image](https://github.com/audacity/audacity/assets/51212933/9bdeb434-2195-4154-8d12-d5338be327a5)

From my understanding, it seems after '999' bars, the value wraps incorrectly. 

I'm going to try and work on this issue.

dozzzzer:
> it seems after '999' bars, the value wraps incorrectly

There's indeed an issue with the display of more than 999 bars, and it has been fixed already, although PR #5190 fixing it hasn't been merged yet.

dozzzzer:
Tested

PR #5310, opened 2023-10-05
Time Controls `bar:beats` resizing fixes

crsib:
Resolves: #5270
Resolves: #5232

This PR addresses issues introduced when the `bar:beats` formats were allowed to resize to always fit the number of bars needed.

Fixes implemented:

* Format can only "grow", simplifying the keyboard interaction. This helps to address 5270 cases A and C. Case B is intentionally kept so the behavior matches the one Audacity has now.
* In the selection bar, we always use the widest format, addressing 5232.
* Both toolbars now correctly adapt to the Time Control size.

Recommended:

### DIFF
--- a/libraries/lib-numeric-formats/tests/NumericConverterTests.cpp
+++ b/libraries/lib-numeric-formats/tests/NumericConverterTests.cpp
@@ -213,7 +213,7 @@ TEST_CASE("BeatsNumericConverterFormatter", "")
             const auto value = bar * timeSignature.GetBarDuration() +
                                beat * timeSignature.GetBeatDuration();
 
-            basicFormatter->UpdateFormatForValue(value);
+            basicFormatter->UpdateFormatForValue(value, true);
 
             const auto formattedString =
                wxString::Format("%03d bar %02d beat", bar + 1, beat + 1);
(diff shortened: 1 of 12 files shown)


libraries/lib-numeric-formats/tests/NumericConverterTests.cpp
```diff
@@ -213,7 +213,7 @@ TEST_CASE("BeatsNumericConverterFormatter", "")
             const auto value = bar * timeSignature.GetBarDuration() +
                                beat * timeSignature.GetBeatDuration();
 
-            basicFormatter->UpdateFormatForValue(value);
+            basicFormatter->UpdateFormatForValue(value, false);
```
vsverchinsky: So it never shriks? I mean all calls to that function pass `false`

src/widgets/NumericTextCtrl.cpp
```diff
@@ -687,20 +687,38 @@ void NumericTextCtrl::OnFocus(wxFocusEvent &event)
    event.Skip( false ); // PRL: not sure why, but preserving old behavior
 }
 
-void NumericTextCtrl::OnFormatUpdated()
+void NumericTextCtrl::OnFormatUpdated(bool resetFocus)
 {
    NumericConverter::OnFormatUpdated();
```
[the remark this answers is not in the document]
crsib: Great catch!

libraries/lib-numeric-formats/tests/NumericConverterTests.cpp
crsib: Never indeed

vsverchinsky (approved):

PR #5320, opened 2023-10-06
Don't compute interpolation of samples when no pitch shifting

Paul-Licameli:

This one-line change skips a calculation that is needed only when shifting pitch as well as time.

QA can confirm exactly 0 difference in before and after calculations of time stretch.

This may be worth testing for further performance improvement.  My own preliminary, rough tests suggest it might save a few percent in a macOS release build.

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -456,7 +456,7 @@ void TimeAndPitch::feedAudio(const float* const* input_smp, int numSamples)
     {
       float smp[6];
       d->inResampleInputBuffer[ch].readBlock(int_pos - 6, 6, smp);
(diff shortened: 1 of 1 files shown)


crsib:
That is pretty cool, `lagrange6` was pretty costly. I will check the result

saintmatthieu (approved):
LGTM, now let QA confirm @Paul-Licameli 's bit-exactness claim.

crsib:
This does give measurable improvement in performance

crsib (approved):

Paul-Licameli:
> This does give measurable improvement in performance

Can you share some numbers?

crsib:
I would say 10% on Windows, I will recheck later, when my PC is idle again

chinakov:
Yes, comparing Master and this branch's build via 'Invert, mix two tracks into new track' yields results of 0 on upstretching and downstretching. Tested, ready to merge!

PR #5324, opened 2023-10-06
5313 clean up all hidden data when rendering stretch

### LINKED ISSUE #5313: Render Clip Stretching produces incorrect result
### Steps to reproduce

1. download and open [this project](https://www.dropbox.com/scl/fi/8q5nxf98biy4pjonneai9/Render-Stretching-Bug.aup3?rlkey=o9arietc3qccdu7nyzeqwfagz&dl=0)
2. right-click either of the two clips' affordances and select "Render Clip Stretching"
3. observe
4. undo
5. mix and render the track
6. compare the result with the one in step 3

### Expected behavior

The results are identical;
In step 3, the result produced by the "Render Clip Stretching" doesn't create any additional envelope points on the clip

### Actual behavior

The results in steps 3 and 6 are different;
In step 3, "Render Clip Stretching" produces a wrong clip with unwanted envelopes:

https://github.com/audacity/audacity/assets/70063970/7646f17e-f901-4f5c-9510-771258faca83

### Audacity Version

current master

### Operating system

Windows, macOS

### Additional Info

Introduced by #5239

saintmatthieu:
Resolves: #5313

Please find details in #5313 

Recommended:

QA:
- [x] Bug reported in #5313 is indeed fixed,
- [x] also have cutlines on a stretched just about hidden by left- and right-trimming, and then stretch-render the clip. It should also yield normal results (previously the cutline hidden away be left trimming would creep back in to the clip).

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -1227,8 +1227,10 @@ void WaveClip::ApplyStretchRatio(const ProgressReporter& reportProgress)
    // Leave 1 second of raw, unstretched audio before and after visible region
    // to ensure the algorithm is in a steady state when reaching the play
    // boundaries.
-   TrimLeftTo(originalPlayStartTime - stretchRatio);
-   TrimRightTo(originalPlayEndTime + stretchRatio);
+   SetTrimLeft(0);
+   SetTrimRight(0);
+   ClearLeft(originalPlayStartTime - stretchRatio);
+   ClearRight(originalPlayEndTime + stretchRatio);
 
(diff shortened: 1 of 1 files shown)


saintmatthieu:
@Paul-Licameli this is based on 4-of-6, as there'd be conflicts otherwise and there are fair chances that 4-of-6 gets merged before.
Also, the second last commit ("Imperfect attempt") is just meant as fallback in case you find issues with the approach of using `ClearLeft` and `ClearRight` upfront. Please examine the overall diff first.

crsib (approved):
I understand the changes

chinakov:
Tested, ready to merge. Thanks!

PR #5328, opened 2023-10-07
One fewer circular buffer

Paul-Licameli:

Recommended:


libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -403,7 +419,11 @@ void TimeAndPitch::setTimeStretchAndPitchFactor(double timeScale, double pitchFa
 {
   assert(timeScale > 0.0);
   assert(pitchFactor > 0.0);
-  _pitchFactor = pitchFactor;
+  if (_pitchFactor != pitchFactor)
+  {
+    _pitchFactor = pitchFactor;
+    setupInputBuffers();
```
saintmatthieu, line 426: I haven't tested this library with dynamic pitch or time ratio changes, but I think `setTimeStretchAndPitchFactor` is supposed to be callable on the fly. `setupInputBuffers` seem to reset memory, which I suppose would now introduce glitches in such use case.

crsib: > I think setTimeStretchAndPitchFactor is supposed to be callable on the fly

Agree; this would be needed when we implement "tempo track"

Paul-Licameli:
It is rebased, it doesn't reallocate for change of pitch factor, and it still gives a few percent more improvement when I trye it in a release build.

saintmatthieu:
... but will this still work if you change the pitch factor at runtime ? I see that when it's not 1, the `inResampleInputBuffer` is used. Doesn't switching from using it to not using it still introduce glitches ?

Even if you can prove it doesn't, I think I'd still be less enthusiastic about this PR than e.g. the recent one-liner. This one adds significant complexity to the code.

Paul-Licameli:
> ... but will this still work if you change the pitch factor at runtime ? I see that when it's not 1, the `inResampleInputBuffer` is used. Doesn't switching from using it to not using it still introduce glitches ?
> 
> Even if you can prove it doesn't, I think I'd still be less enthusiastic about this PR than e.g. the recent one-liner. This one adds significant complexity to the code.

Good point, if ever there is a switch back to nonunit pitch ratio, then there would be a need to populate the resample buffer from the input buffer.

Yes, this adds complexity, but my own rough measurement suggests significant gains in performance.  If other measurement confirms this, then we must ask if we want that performance even at the expense of more complexity.

But the "more complexity" is actually simplicity for the present needs, when pitch ratio really is a constant 1.

Issue #5337, opened 2023-10-09
In a generator, with a screen reader running, changing the format of a duration can crash Audacity


DavidBailes:
### Bug description

With a screen reader running, in one of the the built in generators, for example chirp, changing the format of the duration can crash Audacity. More precisely, if the focused digit is in the nth field of the duration, and you change the format to one which doesn't have an nth field, Audacity crashes.

### Steps to reproduce

On Windows, with a screen reader running, eg NVDA or Narrator:
1. Open Audacity.
2. On the Generate menu, choose Chirp.
3. Tab to the duration, and press right arrow to move one digit to the right
4. Open the time control's context menu, and choose seconds.

### Expected behavior

Nothing unexpected.

### Actual behavior

Audacity crashes.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

This bug seems to have been introduced in Audacity 3.3.0.

chinakov:
Tested.

PR #5338, opened 2023-10-09
issue #5357: Change of format of a duration in a generator can cause …

DavidBailes:
…crash

Resolves: https://github.com/audacity/audacity/issues/5337

With a screen reader running, in one of the the built in generators, for example chirp, changing the format of the duration can crash Audacity.

Problem:
NumericTextCtrlAx::GetName(), uses NumericTextCtrl::GetFocusedField() to get the focused field. It then uses that as in index for a vector. However, NumericTextCtrl::UpdateAutoFocus() does not update the value of mLastField, which is returned by NumericTextCtrl::GetFocusedField(). So after a change of format, GetFocusedField() can return a value which does not exist in the new format. This can cause Audacity to crash. This happens from Audacity 3.3.0, and I don't know why it didn't in versions before this.

Possible fixes:
1. Change NumericTextCtrl::UpdateAutoFocus() so that it updates mLastField.
2. Given that NumericTextCtrl::GetFocusedField() is only used by the accessibility object, and that the focused field can be derived from NumericTextCtrl::GetFocusedDigit(), remove NumericTextCtrl::GetFocusedField() and NumericTextCtrl::mLastField.

The second fix was implemented, as it will reduce the chance of errors in this area in the future.

Recommended:

### DIFF
--- a/src/widgets/NumericTextCtrl.cpp
+++ b/src/widgets/NumericTextCtrl.cpp
@@ -156,7 +156,6 @@ NumericTextCtrl::NumericTextCtrl(
    mBackgroundBitmap{},
    mDigitFont{},
    mLabelFont{},
-   mLastField(1),
    mAutoPos(options.autoPos)
    , mType(type)
 {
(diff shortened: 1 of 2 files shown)


crsib (approved):
Awesome, thank you, David

PR #5339, opened 2023-10-09
_unwrapPhase really maps to [-PI, PI) as comment says...

Paul-Licameli:
...truncf would also do the right thing without setting rounding mode, but it doesn't compile for the vectorized function.

QA:  This makes subtle changes in the algorithm.  I'm not sure what exactly to expect yet.
- [ ] Better or worse, subjectively, in various toy and real world examples?  
- [ ] Any more or less loss of RMS in examples?
- [ ] More or less cross-platform differences in rendered results?

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <array>
 #include <cassert>
+#include <cfenv>
 #include <utility>
 
 #include "CircularSampleBuffer.h"
@@ -185,12 +186,14 @@ namespace {
 // wrap a phase value into -PI..PI
 inline float _unwrapPhase(float arg)
 {
+  assert(std::fegetround() == FE_TOWARDZERO);
   using namespace audio::simd;
   return arg - rint(arg * 0.15915494309f) * 6.283185307f;
 }
 
 void _unwrapPhaseVec(float* v, int n)
 {
+  assert(std::fegetround() == FE_TOWARDZERO);
   audio::simd::perform_parallel_simd_aligned(v, n, [](auto& a) { a = a - rint(a * 0.15915494309f) * 6.283185307f; });
 }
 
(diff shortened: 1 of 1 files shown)


Paul-Licameli:
Could these mistakes in phase have caused some destructive interference in the overlap-add of windows explaining some of the loss of RMS that we observe?

Maybe but a few experiments convince me this is not yet a complete explanation.

crsib (commented):
I'm pretty sure that there is a trunc instruction both in SSE2 (cvttps2dq) and Neon. 

Does this error matters much though? How do you measure it? `x - round(x)` is in [-0.5, 0.5], so unwrap will yield results in [-Pi, Pi]

Paul-Licameli:
> I'm pretty sure that there is a trunc instruction both in SSE2 (cvttps2dq) and Neon.
> 
> Does this error matters much though? How do you measure it? `x - round(x)` is in [-0.5, 0.5], so unwrap will yield results in [-Pi, Pi]

See my comments above.  `round(x) - trunc(x)` may be 1 or -1 meaning an error of +/- 2 * pi from `_unwrapPhase` and that can contaminate the accumulated new phase information when multiplied by `alpha` which is the stretch ratio.  When phase is unwrapped again, that extra term is not a neat multiple of 2 * pi and does not vanish.

Try it on examples too and compare results with and without this change.

Paul-Licameli:
> I'm pretty sure that there is a trunc instruction both in SSE2 (cvttps2dq) and Neon.
> 
>

Whether it can be done more efficiently with `trunc` than with rounding mode -- I leave to you.

crsib:
`x - trunc(x)` maps to [0, 1) when `x` is positive and to `(-1, 0]` when `x` is negative, so the code in this PR, which is effectively `2π *(x/2π + trunc(x/2π) ~ 2π*(a - trunc(a))` maps to (-2π, 0] when `x` is negative and to [0, 2π) otherwise. So the `_unwrapPhase` you propose does not map to the [-π, π], it maps to (-2π, 2π). At the very least that should be fixed before it makes sense to perform any checks. I am still not convinced that `x - trunc(x)` gives any better result

crsib:
On top of that, the vector version used on ARM uses `vrndnq_f32`, which ignores the rounding mode and always uses round-to-even (same as the default for `rint`), so you code will produce different results on ARM based Macs for `_unwrapPhase` and `_unwrapPhaseVec`

Paul-Licameli:
> `x - trunc(x)` maps to [0, 1) when `x` is positive and to `(-1, 0]` when `x` is negative, so the code in this PR, which is effectively `2π *(x/2π + trunc(x/2π) ~ 2π*(a - trunc(a))` maps to (-2π, 0] when `x` is negative and to [0, 2π) otherwise. So the `_unwrapPhase` you propose does not map to the [-π, π], it maps to (-2π, 2π). At the very least that should be fixed before it makes sense to perform any checks. I am still not convinced that `x - trunc(x)` gives any better result.

You are right— I wasn’t thinking clearly enough.

However — asserting that the rounding mode is to-nearest, would make explicit the assumptions.

What would you think of this PR with that substitution?

Paul-Licameli:
I think I was really grasping toward this:

Add pi.  Divide by 2 pi, rounding downward.  Multiply by 2 pi.  Subtract that from the original value.

Paul-Licameli:
Needs more thought

Issue #5292, opened 2023-09-30
When trying to paste clips, an error is thrown if there are tiny clips in the copied selection


Tantacrul:
### Bug description

See video.

(Update) I now realise that this is due to some tiny clips in my project that were included in the selection I copied. pasting works if I find and remove them. However, pasting should still work even if tiny cliplets are present.

https://github.com/audacity/audacity/assets/21022311/a74f970c-9f25-42eb-b419-5e294a76a24a

### Steps to reproduce

1. Turn on snapping and set it to 'bar'
2. Select audio across multiple clips
3. Try to paste the audio next to an existing clip

### Expected behavior

Pasting should work in this circumstance

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

petersampsonaudacity:
@Tantacrul 

Isn't part of the problem that it is almost impossible to spot such "tiny cliplets" ?

Tantacrul:
@petersampsonaudacity 

Yes, although in my video, I'm copying a region that has an exact number of beats and pasting it into an area that is the same number of beats, so the cliplets shouldn't be causing an issue.

It would be desirable (obviously) to figure out why my project has so many... but that's a separate problem. This one feels more fundamental.

Eventually, we'll be able to make this a much nicer experience by allowing for overlapping clps

petersampsonaudacity:
> It would be desirable (obviously) to figure out why my project has so many...

@Tantacrul if you find out just why Martin, I'd be intrigued to know.

I always thought such cliplets were maybe an edge-case, but I recall that we had a recent poster on the Forum who was being similarly blocked by an "invisible" cliplet.

DavidBailes:
> Isn't part of the problem that it is almost impossible to spot such "tiny cliplets" ?

Using one of the next/previous clip commands (eg Tab/Shift + Tab), you should be able to spot them. And if you temporarily turn on Narrator (Ctrl + Windows key + Enter), then narrator will read the position of the clip within the clips in the track, eg 3 of 5, which probably makes the presence of "invisible clips" even more obvious.

saintmatthieu:
Assigned @vsverchinsky - likely this is indeed due to the presence of zero-length clips.
Once he has found out how to make their creation impossible, and preventing projects from reading or writing them, it should fix this problem, as well as #5279 and #5268.

PR #5342, opened 2023-10-09
Attemp to prevent zero clips from being created

vsverchinsky:
Resolves: #5292

The issue isn't yet completely resolved - though it seem that zero clips are not a problem any more, but it's still very hard to use stretching with copy-paste approach, (especially when snapping enabled) as it often results in attempt to paste into preceding clip.

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -1437,6 +1437,11 @@ void WaveClip::TrimRightTo(double to)
    mTrimRight = std::max(mTrimRight + delta, .0);
 }
 
+double WaveClip::GetPivot() const noexcept
+{
+   return mSequenceOffset + mTrimLeft;
+}
+
 double WaveClip::GetSequenceStartTime() const noexcept
 {
     // JS: mSequenceOffset is the minimum value and it is returned; no clipping to 0
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -1373,6 +1373,11 @@ double WaveClip::GetPlayDuration() const
    return GetPlayEndTime() - GetPlayStartTime();
 }
 
+bool WaveClip::IsEmpty() const
+{
+   return GetPlayDuration() < 0.5 / GetRate();
+}
+
 sampleCount WaveClip::GetPlayStartSample() const
 {
    return sampleCount { GetPlayStartTime() * mRate + 0.5 };
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -274,6 +274,7 @@ size_t WaveClip::GetAppendBufferLen() const
 void WaveClip::OnProjectTempoChange(
    const std::optional<double>& oldTempo, double newTempo)
 {
+   const auto duration = GetPlayDuration();
    if (!mRawAudioTempo.has_value())
       // When we have tempo detection ready (either by header-file
       // read-up or signal analysis) we can use something smarter than that. In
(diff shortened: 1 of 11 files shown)


libraries/lib-wave-track/WaveClip.h
```diff
@@ -499,6 +501,7 @@ class WAVE_TRACK_API WaveClip final :
     * @return true and succeed if and only if `this->GetWidth() ==
     * other.GetWidth()` and either this is empty or `this->GetStretchRatio() ==
     * other.GetStretchRatio()`.
+    * @pre other.GetClipDuration() > 0
```
saintmatthieu, line 504: Doesn't look like there is a `WaveClip::GetClipDuration()` ...

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1366,19 +1371,23 @@ double WaveClip::GetPlayEndTime() const
...
 }
 
 double WaveClip::GetPlayDuration() const
 {
    return GetPlayEndTime() - GetPlayStartTime();
 }
 
+bool WaveClip::IsEmpty() const
+{
+   return GetPlayDuration() < 0.5 / mRate * mClipStretchRatio;
```
saintmatthieu, line 1388: Are you sure this shouldn't be `GetStretchRatio()` ? `IsEmpty()` should return the same value independently on tempo change, meanwhile `GetPlayDuration()` would.

saintmatthieu (commented):
Some comments, but still reviewing.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1427,14 +1436,12 @@ void WaveClip::TrimRight(double deltaTime)
 
 void WaveClip::TrimLeftTo(double to)
 {
-   mTrimLeft =
-      std::clamp(to, SnapToTrackSample(mSequenceOffset), GetPlayEndTime()) -
-      mSequenceOffset;
+   mTrimLeft = std::clamp(to, mSequenceOffset, GetPlayEndTime()) - mSequenceOffset;
```
saintmatthieu, line 1439: Maybe it's okay this way, but I'd round `to`. Other places in the code (`WaveClipAdjustBorderHandle`, `WaveTrack::SplitAt`) round the argument before calling the method, only `::ApplyStretchRatio` doesn't. It might cause problems and I don't see a use case for trimming to places between samples.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -661,7 +661,9 @@ void WaveClip::WriteXML(XMLWriter &xmlFile) const
 /*! @excsafety{Strong} */
 bool WaveClip::Paste(double t0, const WaveClip& other)
 {
-   if (GetWidth() != other.GetWidth())
+   assert(other.GetPlayDuration() > 0);
```
saintmatthieu, line 664: I see in `WaveTrack::ClearAndPaste`, where the code restores the hidden part of the clips, that you now discard `split.left` or `split.right` if empty. I think the result would be the same if you didn't assert here and use the check line 666, right? I'd rather that, because it might keep usage of this method simpler, e.g., you wouldn't need those changes: 
image

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1572,8 +1579,16 @@ void WaveTrack::ClearAndPasteOne(
                if (split.right)
                   // new clip was cleared left
                   attachLeft(*newClip, *split.right);
-               bool success = track.AddClip(std::move(newClip));
-               assert(success); // copied clip has same width and factory
+               assert(clip->GetPlayDuration() > 0);
```
saintmatthieu, line 1582: These three lines would be better placed immediately after the call to `ClearRight`: it took me some time to check that it had nothing to do with what's currently in between.

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -290,12 +290,14 @@ void WaveClip::OnProjectTempoChange(
 
 void WaveClip::StretchLeftTo(double to)
 {
-   const auto pet = GetPlayEndTime();
-   if (to >= pet)
+   const auto sequenceDuration = GetNumSamples().as_double() * GetStretchRatio() / mRate;
+   const auto pet = mSequenceOffset + sequenceDuration;
+   to = std::min(to, pet);
+   
+   const auto ratioChange = (pet - to) / (sequenceDuration - mTrimLeft - mTrimRight);
```
saintmatthieu, line 297: I'm having trouble with this change. Say clip looks like this:
```
0 1 2 3 4 5 6 7 // time
[   (     ) * ]
```
where `[` is sequence start time, `(` is play start time, and `*` is `to`
`ratioChange` would be `(7 - 6)/3`, does that make sense? It shouldn't even be allowed to try and stretch from the left past the play end time. I know that the border handle code prevents this, but by itself this implementation doesn't look correct.

I'm reading the commit's comment, and I understand that snapping may bring the argument `to` to be something other than the sampling period. You made me aware of this and now I can easily reproduce not nice situations, e.g. by setting the project rate to 7Hz (that's allowed) and the snapping to deciseconds. Stretching yields ugly results.

But I'm wondering if it wouldn't be simpler to for the snapping to the project sample rate period, even if the UI says otherwise. Users setting triplet snapping would I'm sure not mind a small misalignment in the display if zoomed in very close.

I don't know the snapping code, but would it be possible to make such an experiment?

saintmatthieu (commented):
Haven't looked at everything, and would like your response to my suggestion before continuing.

libraries/lib-wave-track/WaveClip.cpp
saintmatthieu: > e.g. by setting the project rate to 7Hz (that's allowed) and the snapping to deciseconds. Stretching yields ugly results.

Here's an example project, very small :
[ugly stretching.zip](https://github.com/audacity/audacity/files/12879663/ugly.stretching.zip)

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -290,12 +290,14 @@ void WaveClip::OnProjectTempoChange(
 
 void WaveClip::StretchLeftTo(double to)
 {
-   const auto pet = GetPlayEndTime();
-   if (to >= pet)
+   const auto sequenceDuration = GetNumSamples().as_double() * GetStretchRatio() / mRate;
+   const auto pet = mSequenceOffset + sequenceDuration;
```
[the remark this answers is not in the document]
vsverchinsky: > And then pet, which stood for "play end time" (sorry, I realize now that it wasn't obvious :D )
No worries, I understood that) I just forgot to add `-mTrimRight`, thanks!

libraries/lib-wave-track/WaveClip.cpp
vsverchinsky: > where [ is sequence start time, ( is play start time, and * is to
ratioChange would be (7 - 6)/3, does that make sense? It shouldn't even be allowed to try and stretch from the left past the play end time. I know that the border handle code prevents this, but by itself this implementation doesn't look correct.

Considering that there was an error in `pet` evaluation this quesition is no longer relevant, right?

vsverchinsky: > I'm reading the commit's comment, and I understand that snapping may bring the argument to to be something other than the sampling period

`GetPlayStartTime` and `GetPlayEndTime` return values that are multiples of a sample length

> Stretching yields ugly results.

Could you please be more specific?

crsib: > I don't know the snapping code, but would it be possible to make such an experiment?

All snapping should go through `ProjectSnap::SnapTime` now, so I think that would be simple

saintmatthieu: @vsverchinsky 

> Could you please be more specific?

You can try stretching the clip in the example project above, but here's a video: https://github.com/audacity/audacity/assets/22740106/7863cbc2-3351-4021-aebe-cfcdd11ea5a5

saintmatthieu: > Considering that there was an error in pet evaluation this quesition is no longer relevant, right?

Probably :)

vsverchinsky: I really do not understand that snap-to-sample strategy considering that you can easily cut-off half of a sample inside a clip if you stretch it to 50% then trim it by 1 sample.

saintmatthieu: Visually this last sample will still look like one, but when rendered, two will be generated. By cutting-off half a sample of a 50% clip you effectively remove one complete sample from the rendering.

Issue #5297, opened 2023-10-02
Unnecessary warning message when merging two identical stretched (or trimmed) mono tracks to stereo


dozzzzer:
### Steps to reproduce

1. Generate chirp
2. Stretch (shrink) it
3. Ctrl + D
4. Select "Make Stereo Track" in TCP

### Expected behavior

A stereo track combining two mono tracks

### Actual behavior

A warning dialog "Combine Mono to Stereo":

<img width=450 src=https://github.com/audacity/audacity/assets/70063970/c99368f9-f56f-4ef2-8894-799ff453d4a4>

### Audacity Version

current master / alpha / nightly build

### Operating system

All

### Additional context

Clicking "Yes" in the dialog will merge them anyway.

dozzzzer:
Bisects to https://github.com/audacity/audacity/commit/410a0515c7b612a2497f4783737c7b298f9a5c35

petersampsonaudacity:
Same happens if you trim the Clip (making a smart clip) with unmodified click&drag  at Step 2

And indeed also happens of at Step 2 you select a chunk in the middle if the chirp and use the **Trim** button in the **Edit** toolbar - or use **Edit > Remove Special > Trim**

petersampsonaudacity:
As reported by @SteveDaulton  on Discord

The same happens with an unstretched mono track if you add a clip line

1.Generate a chirp 
2. Add a split line
3. Duplicate the entire track
4. Select "Make Stereo Track" in the upper track's context  menu
5. Observe the message (even though both tracks are identical)

LWinterberg:
This is probably not a regression. In 3.3.3, if you follow the STR, or even just

1. Have a stereo track
2. Split stereo
3. Merge stereo

it creates asynchronous stereo clips. The only difference in https://github.com/audacity/audacity/commit/410a0515c7b612a2497f4783737c7b298f9a5c35 would be that the asynchronous clip now creates the message, while the async clips could be more easily overlooked

SteveDaulton:
Testing on Linux with 3.3.0
1.Generate a chirp
2. Add a split line
3. Duplicate the entire track
4. Select "Make Stereo Track" in the upper track's context menu
5. Observe that the tracks  become joined, albeit not a normal stereo track. The split from step 2 is preserved.

With Audacity 3.2 and 3.1: Same behaviour as 3.3.0

With Audacity 2.4.2:
1.Generate a chirp
2. Add a split line
3. Duplicate the entire track
[...]

PR #5345, opened 2023-10-10
Fix clip equality check in `MakeStereo` command

vsverchinsky:
Resolves: #5297 

Recommended:


src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -760,26 +760,21 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
...
             rightIntervals.begin(),
             rightIntervals.end(),
             [&](const auto& b)
             {
-               return a->Start() == b->Start() &&
-                  a->End() == b->End() &&
-                  isPlainInterval(*b);
+               //Start() and End() are always snapped to a sample grid
+               return std::fabs(a->Start() - b->Start()) < eps &&
+                  std::fabs(a->End() - b->End()) < eps &&
```
crsib, line 774: `std::abs` should be enough

src/tracks/playabletrack/wavetrack/ui/WaveTrackControls.cpp
```diff
@@ -760,26 +760,21 @@ void WaveTrackMenuTable::OnMergeStereo(wxCommandEvent &)
...
+                  a->GetTrimLeft() == b->GetTrimLeft() &&
```
crsib, line 775: I wonder if we need the same `eps` here?

crsib: Or, probably `numeric_limits<double>::epsilon() * std::max(a->GetTrimRight(), b->GetTrimRight())`

vsverchinsky: Yes, I think it could be `epsilon * max`, but not `0.5 * rate` because trim are no longer aligned with sample grid. Thanks

PR #5348, opened 2023-10-10
More time stretch hacks

Paul-Licameli:
Still more performance squeezed from time stretching.

Eliminate unnecessary array rotation with an equivalent change of phase calculation; vectorize multiplication.

On my machine in release builds, first indications are another 10% saved.

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/VectorOps.h
+++ b/libraries/lib-time-and-pitch/StaffPad/VectorOps.h
@@ -106,7 +106,7 @@ inline void calcPhases(const std::complex<float>* src, float* dst, int32_t n)
   simd_complex_conversions::perform_parallel_simd_aligned(
      src, dst, n,
      [](const __m128 rp, const __m128 ip, __m128& out)
-     { out = simd_complex_conversions::atan2_ps(ip, rp); });
+     { out = simd_complex_conversions::atan2_ps_alternating(ip, rp); });
 }
 
 inline void calcNorms(const std::complex<float>* src, float* dst, int32_t n)
@@ -128,7 +128,7 @@ inline void rotate(
 inline void calcPhases(const std::complex<float>* src, float* dst, int32_t n)
 {
   for (int32_t i = 0; i < n; i++)
-    dst[i] = std::arg(src[i]);
+    dst[i] = std::arg((i % 2) ? -src[i] : src[i]);
 }
 
 inline void calcNorms(const std::complex<float>* src, float* dst, int32_t n)
--- a/libraries/lib-time-and-pitch/StaffPad/VectorOps.h
+++ b/libraries/lib-time-and-pitch/StaffPad/VectorOps.h
@@ -71,13 +71,6 @@ inline void constantMultiplyAndAdd(const T* src, T constant, T* dst, int32_t n)
 }
 
 template <class T>
-inline void multiply(const T* src1, const T* src2, T* dst, int32_t n)
-{
-  for (int32_t i = 0; i < n; i++)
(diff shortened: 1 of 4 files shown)


crsib (changes requested):
Implementation of this code on Windows doesn't match other platforms.

Paul-Licameli:
Rebased, and the Windows implementation is complete.

But the performance advantage that I saw before on Intel Mac no longer happens after f1d08d108c8ea620ee97dc3385f35c4c0e86a003

But I don't know what will happen to performance if #5328 (suitably rewritten) is combined.

crsib (approved):

crsib:
Oh, wait, it failed to build, I think msvc doesn't like `1f` for some reason, I think `1.0f` would fix it

Paul-Licameli:
Sorry, let's try CI again.

As I said, this seems only neutral for performance now on my machine, but this may vary.  I hope you can measure it.

Paul-Licameli:
Let's try again to fix the build...

libraries/lib-time-and-pitch/StaffPad/VectorOps.h
```diff
@@ -128,7 +121,7 @@ inline void rotate(
 inline void calcPhases(const std::complex<float>* src, float* dst, int32_t n)
```
saintmatthieu, line 121: I would like to see an indication here that it's not a straightforward phase calculation of the `src` vector. I'm thinking renaming `calcPhasesAlternating`, consistently with the naming elsewhere, and also adding a comment where it is used in TimeAndPitch.cpp.

saintmatthieu: A little bit of pseudo maths in the comment explaining the mathematical equivalence would also be welcome.

saintmatthieu:
This should be tested for (near) bit exactness on different platforms and for both mono and stereo, which is a few combinations and I fear we miss something.
I would like to see the addition of a unit test for that, concretely, for some mono and some stereo file, checking in the checksum for each platform, and let the test check for equality.
Sometimes, such as in this PR, it might change, and for that case the test should also compute the RMS of the diff. We'd then check that diff manually and make sure that it is in an acceptable range, and if it's not, then do listening tests.

crsib:
From what I observed, "straightforward" unit testing won't provide good coverage for the stretcher. It is really dependent on the input and behaves like a chaotic system a lot. I think for this case the detailed description in the commit message that clarifies the math is the only way to prove that the code is equivalent.

saintmatthieu:
Chaotic? It is deterministic processing. That results are different on different architectures is clear, and that would mean a checksum for each.

saintmatthieu:
Bit-exactness on a sufficiently complex input signal (a musical mix) would increase my confidence that the changes intended for refactoring or optimization only don't introduce unintentional changes, not yours? That'd be the sole purpose of this test, with no intention to estimate quality in absolute terms, only as a statement that the quality remains equal.

Paul-Licameli:
On Intel Mac, this change makes bit-exact identical results to before, on at least one minutes long example of white noise shrunk to 75% duration.

The question is whether this saves any time in a release build.  I thought it did, but when I rebased onto  recent master, I could not reproduce the savings.  But this may vary across operating systems.

Given the ratio and duration of the input, is there dependency of the time stretching should take, on the content?  Possibly in the steps that identify peaks and propagate phases.  But I think not, in all the rest of it -- reading a track, interpolation (well skip that), forward and inverse FFT, overlap-add, writing to a track.

PR #5363, opened 2023-10-11
Don't need to copy arrays to remember last phases and norms

Paul-Licameli:

I thought this elimination of some needless data movement might improve time stretching performance yet again.

Release builds on my Intel Mac do not show much improvement.

But maybe measurement on other architectures will show otherwise.

QA:
- [ ] Again, zero difference between before and after versions of stretch computation.

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -70,13 +70,14 @@ struct TimeAndPitch::impl
 
   SamplesReal fft_timeseries;
   SamplesComplex spectrum;
-  SamplesReal norm;
-  SamplesReal phase;
-  SamplesReal last_phase;
+
+  static constexpr size_t historyLength = 2;
+  SamplesReal norm[historyLength];
+  SamplesReal phase[historyLength];
+  
   SamplesReal phase_accum;
   SamplesReal cosWindow;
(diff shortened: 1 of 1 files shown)


saintmatthieu:
These changes look ok to me. Now if the performance doesn't improve significantly, though, I think it might be better to not diverge some more. But I don't have a strong opinion there.

Only one comment: do you expect the history length to potentially change? You've looked  into the algorithm, are there approaches to using longer histories? Otherwise I'd even simplify and hard-code it to 2 everywhere. But that's minor.

Paul-Licameli:
> These changes look ok to me. Now if the performance doesn't improve significantly, though, I think it might be better to not diverge some more. But I don't have a strong opinion there.
> 
> Only one comment: do you expect the history length to potentially change? You've looked into the algorithm, are there approaches to using longer histories? Otherwise I'd even simplify and hard-code it to 2 everywhere. But that's minor.

I was speculating that perhaps a future enhanced algorithm might achieve higher quality using more than two windows of history.  I believe the paper we were referred to suggests such for getting a better, centered estimate of partial derivatives of phase with time.

I was writing it more generally than necessary, when instead my first thought was just to swap pairs of buffers.

Actually, now I might want to rewrite this in that simpler way until need really proves, to avoid conflict with another idea  that I haven't explained yet.

Paul-Licameli:
Rebased, unchanged

Issue #5364, opened 2023-10-11
Export Dialog: Variable Bitrate abbreviation should be uppercase


dozzzzer:
<img width="384" alt="Screenshot 2023-10-11 at 17 12 59" src="https://github.com/audacity/audacity/assets/70063970/a81668d5-0a1c-4e3d-9d79-dbb406f992dc">

Change "Vbr Mode" to "VBR Mode" in the Export dialog when Opus is selected as an output file format.

crsib:
This will be fixed with the new Opus exporter (fixed already, but the exporter doesn't work ... yet)

petersampsonaudacity:
Testing On W10 with @crsib 's latest branch build for this:  Sgt. Pepper's Lonely Hearts Club Band

The Opus export works properly

And the "VBR" is properly capitalized.

dozzzzer:
Tested

Issue #4709, opened 2023-05-18
Opus export crashes Audacity


mrbumpy409:
### Bug description

When I try to export audio in Opus format, Audacity crashes and leaves a 0 KB .opus file. This happens in two versions of Audacity I have tried on my Linux system: 3.3.1 compiled from source and the 3.3.2 AppImage. If it matters, the codec settings I am using are:

* Bit Rate: **128 kbps**
* Compression: **10**
* Frame Duration: **20 ms**
* Vbr Mode: **On**
* Application: **Audio**
* Cutoff: **disabled**

### Steps to reproduce

1. Open/record some audio.
2. Export Audio
3. Choose "Opus (OggOpus) Files (FFmepeg)" from the dropdown menu. I'm not sure if it's necessary to match my codec settings above.
4. Click Save and dismiss the tagging window, if necessary.

### Expected behavior

Audacity should render the Opus file.

### Actual behavior

Audacity crashes, and a 0 KB .opus file is created.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Linux

### Additional context

OS's tested: KDE neon 5.27 and Xubuntu 22.04, both based on Ubuntu 22.04.
[Audacity_dbgrpt-19251-20230518T090058.zip](https://github.com/audacity/audacity/files/11508602/Audacity_dbgrpt-19251-20230518T090058.zip)

petersampsonaudacity:
Testing on W10 with 3.3.2

@mrbumpy409  This works fine for me on Windows (using your settings)

Ryan-Miller-Audacity:
I can't seem to reproduce this issue within my Xubuntu VM. I see the export process finish without hiccup and also have the excepted .opus file which can be played back via rhythmbox or imported into Audacity without issue. Tested on 3.3.2 AppImage.

mrbumpy409:
How bizarre. I get the crash in VirtualBox using Xubuntu 22.04 and Ubuntu Studio 23.04. I haven't tried other distros yet.

mrbumpy409:
All of my Linux installs include the "ubuntu-restricted-extras" package or distro equivalent (e.g., "kubuntu-restricted-extras"). Could this be a factor, since ffmpeg is involved?

mrbumpy409:
After further testing on my two Linux PCs (desktop and laptop) both running the same version of KDE neon, I can ascertain the following:

* Desktop PC
  * The crash happens in KDE neon, native install.
[...]

Issue #443, opened 2020-03-04
Support Opus format for audio 


Ygarr:
I can open and edit opus audio file, but cannot save(export) to same format 

**Describe the solution you'd like**
Export at least as same format edited audio for .opus audio

**Describe alternatives you've considered**
"Ocenaudio" has such feature. Just choose "Save as" and choose "Ogg OPUS" then "Replace existing"

Pokechu22:
Theoretically, ffmpeg should support this, but I can't seem to get that to work (perhaps I have an old ffmpeg version); it lists libopus as a codec but doesn't like selecting the opus format, and selecting the opus format and then libopus causes an error on export (see [Bug 2166](https://bugzilla.audacityteam.org/show_bug.cgi?id=2166)).

Something's definitely strange since [the ffmpeg documentation](https://ffmpeg.org/ffmpeg-codecs.html) mentions a second opus codec (their own implementation), which I don't see listed in audacity; that probably indicates something's outdated.

finn64:
Not sure if this helps, but opus steam is normally contained in webm

petersampsonaudacity:
We (Leland Lucius that is - thanks Leland) has just added Opus Export as a direct choice in Export Audio and Export Multiple for the soon to be upcoming 2.4.0.  I have been QA testing this and writing the documentation for the Manual - and so far all looks good

JamesCrook:
Removed 2.4.0 label, as 2.4.0 does not support opus on Mac (yet).
https://bugzilla.audacityteam.org/show_bug.cgi?id=2381

MisterZeus:
> Not sure if this helps, but opus steam normally uses webm as a container

For saving files though, it should probably use the Ogg container: https://github.com/xiph/ogg

LWinterberg:
Some starting points: 

* Opus support was done previously through FFMPEG (by the looks of it) but got axed due to some problems. https://bugzilla.audacityteam.org/show_bug.cgi?id=2381 has some commits and some related issues linked which themselves may contain relevant commits. 
* https://github.com/audacity/audacity/tree/master/src/export is where Audacity's export logic lives
* According to https://ffmpeg.org/ffmpeg-codecs.html#opus, the FFMPEG opus codec is worse than [xiph's libopus](https://opus-codec.org/development/)

petersampsonaudacity:
[...]

PR #5370, opened 2023-10-12
Native Opus support

crsib:
Resolves: #443 (yes, a three digits issue!)
Resolves: #4709 (probably)
Resolves: #5364

This PR introduces built-in support for Opus files. 

Opus is a peculiar codec, always encoding 48kHz sample rate. Opus allow to store the "original" sample rate in the header, but recommends not to resample to that rate whenever possible. 

Despite that Opus encodes at fixed sample rate - it allows to feed the data at few other sample rates and allows to set the cutoff frequencies independently.

Opus allows to select from a fixed list of frame sizes. Audacity will never create larger frames, but at the end of the file smaller frames can be used.

Please note, that while the options list has a bit different values compared to `Opus (OggOpus) Files (FFmpeg)`, this is intentional and better reflects the codec setup.

* Importing uses `libopus` and `opusfile`.
* Exporting uses `libopus` and `libogg`.

Recommended:

### DIFF
--- a/cmake-proxies/cmake-modules/DependenciesList.cmake
+++ b/cmake-proxies/cmake-modules/DependenciesList.cmake
@@ -17,6 +17,7 @@ audacity_find_package(WavPack)
 audacity_find_package(Ogg OPTION_NAME libogg)
 audacity_find_package(FLAC OPTION_NAME libflac)
 audacity_find_package(Opus OPTION_NAME libopus)
+audacity_find_package(opusfile OPTION_NAME opusfile)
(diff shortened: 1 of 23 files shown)


crsib:
@LWinterberg would be nice if you can check the new strings, please

vsverchinsky (approved):

petersampsonaudacity:
Testing On W10 with @crsib 's latest branch build for this:  Sgt. Pepper's Lonely Hearts Club Band

The Opus export works properly

And the "VBR" is properly capitalized.

PR #5375, opened 2023-10-13
5170 correct stretch scaling

saintmatthieu:

Recommended:

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
@@ -4,6 +4,7 @@
 #include <array>
 #include <cassert>
 #include <utility>
+#include <fstream>
 
 #include "CircularSampleBuffer.h"
 #include "FourierTransform_pffft.h"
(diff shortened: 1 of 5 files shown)


Paul-Licameli:
Only MacOS builds succeeded

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -4,6 +4,7 @@
 #include <array>
 #include <cassert>
 #include <utility>
+#include <fstream>
```
Paul-Licameli, line 7: Why?

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -54,6 +55,18 @@ inline int getFftSize(int sampleRate)
   // without compromising performance.
   return 1 << (12 + (int)std::round(std::log2(sampleRate / 44100.)));
 }
+
+constexpr std::array<float, 2> hannCoefs { .5f, .5f };
```
Paul-Licameli, line 59: Maybe -.5 would be less surprising, then don’t alternate sign in the summation

Paul-Licameli:
Is it correct for both stretching and shrinking?

I think so for stretching because the synthesis hop is an exact fourth of the fft size.

But for shrinking it may be less.

Issue #5170, opened 2023-09-08
Stretched clip: amplify doesn't amplify enough


DavidBailes:
### Bug description

Tested with: https://github.com/audacity/audacity/commit/c9ff9faaa6ad2e348c6640c039025874fa870ac0

### Steps to reproduce

1. Create a tone with an amplitude 0.5 and a length 30s
2. Shrink the clip by about 0.5s _(using time-stretching)_
3. Select part of the clip, and apply amplify with default settings.

### Expected behavior

![image](https://github.com/audacity/audacity/assets/11887021/de24b355-8d7f-4400-a741-27b4202a66bc)

### Actual behavior

![image](https://github.com/audacity/audacity/assets/11887021/00cd34f9-0569-4582-a321-a2dec2797ca4)

If you apply amplify again, you can amplify by another ~0.5dB

### Audacity Version

other

### Operating system

All OSs

### Additional context

This issue does not occur with the Normalize effect.

petersampsonaudacity:
Might this be related to:

**Shrinking or stretching a clip reduces its loudness #5160** ?

Paul-Licameli:
Try it in branch #5138

Paul-Licameli:
> Might this be related to:
> 
> **Shrinking or stretching a clip reduces its loudness #5160** ?

Not exactly, but... (next)

Paul-Licameli:
> Try it in branch #5138

Haven't tried but I predict it won't be fixed in that branch.

Amplify uses the peaks (not the RMS) to decide the default settings:  see where it uses `GetMinMax`

And that applies to the input tracks before the rendering for stretch, even with the fixes in #5138 

Whereas, I predict Normalize will not have this problem in that branch, though it too uses `GetMinMax`.

Paul-Licameli:
It is known that if you render the stretching, RMS changes (we think it ought not to, but don't understand why)

But also, the waveform may look very different though you might not hear difference of timbre in sustained notes.  Even with the same RMS, the difference between that and the peak might change a lot, and that has to do with keeping the magnitudes of the frequency components, but not always their phase.

Example:  Try a square wave (peak == RMS); stretch it; then mix-and-render to a new track.

Amplify is still detecting the peaks of the raw data, as they are before stretching.  #5138 fixes other problems by rendering the stretch in the selected interval before computing the effect transformation.  So I think there would still be a mismatch, when computing the default amplification shown in the dialog before the real processing.

DavidBailes:
[...]

PR #5379, opened 2023-10-13
Issue 5170

Paul-Licameli:
Resolves: #5170

Recommended:


Paul-Licameli:
Rebased unchanged to recent master

saintmatthieu (commented):
Let me recap just to make sure we're on the same page first:
`EffectAmplify` during initialization analyses the audio. It has to do this on the stretch-rendered audio such that, when processing the stretch-rendered audio, it uses the correct scale factor.
This PR does this, but also avoid the duplicate stretch rendering by re-using the output tracks from the analysis phase.

What I'm not enthusiastic about is the added code complexity for optimization. Now the flow of `PerTrackEffect::Process` depends on what has happened in the `Init` method of the derived class. I've been trying to think of a bit more straightforward approach.

Could we get somehow `EffectInstance` to have a virtual `GetOutputTracks` method defaulting to the usual, and let `EffectAmplify::Instance` have it return a reference to the already stretch-rendered output tracks that it owns ? Then `DestroyOutputTracks()` wouldn't need be anymore, because the output tracks would disappear with the instance.

Paul-Licameli:
1 There should be no variation in how the output tracks are computed — only, when.  That is why I chose the interface this way.   The virtual function would be too general.

2 Effects still have the general defect that the effect object (the unique long-lived one, for each different effect) is used to store state of the computation (members like mT0 and inputTracks()).  The remedy for this will be to introduce an EffectContext class and pass it as an extra argument in many places.  I have even written draft PRs for this, awaiting the opportunity for another refactoring campaign.  The best thing will be to put the EffectOutputTracks in that.  Destruction of the EfectContext object will then destroy EffectOutputTracks with no extra call.

The time for all that cleanup is not now.

Paul-Licameli:
Rebased and changed the base branch for merge

saintmatthieu:
> 1 There should be no variation in how the output tracks are computed — only, when. That is why I chose the interface this way. The virtual function would be too general.

I'm not convinced by that, and I think it makes sense that an effect _instance_ may own a track it already has pre-processed in some way for reuse later when applying the effect. I'm not saying this shouldn't be superseeded by the effect context when the time is right, but in the meantime I see it as a possibility to spare additional complication in an already complex effect architecture.

Paul-Licameli:
> > 1 There should be no variation in how the output tracks are computed — only, when. That is why I chose the interface this way. The virtual function would be too general.
> 
> I'm not convinced by that, and I think it makes sense that an effect _instance_ may own a track it already has pre-processed in some way for reuse later when applying the effect. I'm not saying this shouldn't be superseeded by the effect context when the time is right, but in the meantime I see it as a possibility to spare additional complication in an already complex effect architecture.

I don't agree with that:  it's not the purpose of EffectInstance to own a track.  Really the only sort state an EffectInstance is meant to have are things lke settings and delay buffer memory, while it is fed a stream of samples from inside.

EffectContext is (or would be) the more logical structure to hold temporary tracks that are operated on.

The effect system isn't ideal -- it is in a state of evolution from legacy code into something better, but that refactor was itself suspended to pursue others.

saintmatthieu:
I would approve a fix for that bug without the optimization part.

Paul-Licameli:
> I would approve a fix for that bug without the optimization part.

I do not want to write a fix that computes stretch twice, wastefully.

crsib:
I think that the first commit makes an complicated system even more confusing for the sake of fixing a single effect. I like `EffectInstance::GetOutputTracks` a bit more, but again, to fix a single effect additional complexity is introduced, affecting the whole Effects framework.

However, I don't expect any new effect or let alone effect family to be implemented within the current framework, so I don't see much harm from the approach proposed in this PR. This is much better than calculating stretching twice.

I propose that we move this PR forward as is. The problems with the effects framework will be addressed in some near future with a quite high priority.

saintmatthieu:
So be it.

saintmatthieu (approved):

PR #5398, opened 2023-10-19
5390 stretch rendering always joint stereo

### LINKED ISSUE #5390: Audible deterioration when rendering clip stretching
### Bug description

As the title says, when using the context menu command "Render Clip Stretching" on non-synthetic tones, it produces an oddly sounding result, as if channels are out of phase, causing audible changes to the impact of the sound as well as its stereo image. 

It's also worth mentioning that mixing-and-rendering via `Tracks > Mix > Mix and Render` doesn't cause such an issue.

### Steps to reproduce

1. Download & open[ this project ](https://www.dropbox.com/scl/fi/8q5nxf98biy4pjonneai9/Render-Stretching-bug.aup3?rlkey=o9arietc3qccdu7nyzeqwfagz&dl=0)
2. Make sure looping is enabled and hit Play
3. By pressing the **Shift + Solo** button, compare all three tracks where the 1st is the original drum loop shrunk by 6%, the second one is the result of the `Tracks > Mix > Mix and Render` command, and the third one is the duplicate of the 1st track with the "Render Clip Stretching" context menu command applied to it (headphones are highly recommended)

@Paul-Licameli I suggest this testing method too:  Mix-and-render to a new track, then apply Invert to that track, and then play all tracks.  Listen, expecting silence.  Also watch the playback meter (maybe using the -145 dB setting in Interface preferences to see even small, inaudible non-zero signals).

### Expected behavior

All three sound equal

### Actual behavior

The 1st and the 2nd one sound identical;
The 3rd track sounds somewhat different from the previous two, introducing an altered stereo image and less impactful and "tight", "washed out" sounding kick and snare drums.

### Audacity Version

[current master / alpha / nightly build
](https://github.com/audacity/audacity/actions/runs/6546774311)

### Operating system

All

### Additional context

It seems that this behavior has been introduced since the very appearance of this command in the clip context menu, so it's not a regression.

QA (to be added to PR when there is one):
- [ ] Doesn't re-introduce #5313
- [ ] If a clip has no trimming, rendering is with RMB or with mix-and-render command yields bit-exact result
- [ ] Whether stretch-rendering a clip and then apply and effect **or** directly applying the effect yields bit-exact result
- [ ] All the above was done in mono and stereo

saintmatthieu:
Resolves: #5390

Recommended:

QA :
- [x] Doesn't re-introduce https://github.com/audacity/audacity/issues/5313
- [x] If a clip has no trimming, rendering is with RMB or with mix-and-render command yields bit-exact result
- [x] Whether stretch-rendering a clip and then applying an effect or directly applying the effect yields bit-exact result
- [x] All the above was done in mono and stereo
- [x] Test tracks with other sample formats than float
- [x] Test with a non-trivial envelope and observe the correct placement of points on the timeline
- [x] Cut lines are NOT preserved.  Whether this is agreeable to design will be decided later.

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -45,8 +45,10 @@ from the project that will own the track.
 #include "float_cast.h"
 
 #include "AudioSegmentSampleView.h"
+#include "ClipTimeAndPitchSource.h"
 #include "Envelope.h"
 #include "Sequence.h"
+#include "StaffPadTimeAndPitch.h"
 
 #include "Project.h"
 #include "ProjectRate.h"
@@ -187,7 +189,30 @@ WaveTrack::Interval::Interval(const ChannelGroup &group,
 {
 }
 
+WaveTrack::Interval::Interval(
+   const ChannelGroup& group, size_t width,
+   const SampleBlockFactoryPtr& factory, int rate, sampleFormat format)
+    : Interval(
+         group, std::make_shared<WaveClip>(1, factory, format, rate, 0),
+         width == 2 ?
+            std::make_shared<WaveClip>(1, factory, format, rate, 0) :
+            nullptr)
+{
+}
+
 WaveTrack::Interval::~Interval() = default;
+
+void WaveTrack::Interval::Append(
+   constSamplePtr buffer[], sampleFormat format, size_t len)
+{
+   for (unsigned channel = 0; channel < NChannels(); ++channel)
+      GetClip(channel)->AppendNewBlock(buffer[channel], format, len);
+}
+
+void WaveTrack::Interval::Flush()
+{
+   ForEachClip([](auto& clip) { clip.Flush(); });
+}
 
 void WaveTrack::Interval::TrimLeftTo(double t)
 {
@@ -199,6 +224,30 @@ void WaveTrack::Interval::TrimRightTo(double t)
 {
    for(unsigned channel = 0; channel < NChannels(); ++channel)
       GetClip(channel)->TrimRightTo(t);
+}
+
+void WaveTrack::Interval::SetTrimLeft(double t)
+{
+   for(unsigned channel = 0; channel < NChannels(); ++channel)
+      GetClip(channel)->SetTrimLeft(t);
+}
+
+void WaveTrack::Interval::SetTrimRight(double t)
+{
+   for(unsigned channel = 0; channel < NChannels(); ++channel)
+      GetClip(channel)->SetTrimRight(t);
+}
+
+void WaveTrack::Interval::ClearLeft(double t)
+{
+   for(unsigned channel = 0; channel < NChannels(); ++channel)
+      GetClip(channel)->ClearLeft(t);
+}
+
+void WaveTrack::Interval::ClearRight(double t)
+{
+   for(unsigned channel = 0; channel < NChannels(); ++channel)
+      GetClip(channel)->ClearRight(t);
 }
 
 void WaveTrack::Interval::StretchLeftTo(double t)
@@ -213,19 +262,95 @@ void WaveTrack::Interval::StretchRightTo(double t)
       GetClip(channel)->StretchRightTo(t);
 }
 
-void WaveTrack::Interval::ApplyStretchRatio(
-   const std::function<void(double)>& reportProgress)
+WaveTrack::IntervalHolder WaveTrack::Interval::GetStretchRenderedCopy(
+   const std::function<void(double)>& reportProgress, const ChannelGroup& group,
+   const SampleBlockFactoryPtr& factory, sampleFormat format)
 {
-   const auto channelsCount = NChannels();
+   const auto dst = std::make_shared<Interval>(
+      group, NChannels(), factory, mpClip->GetRate(), format);
 
-   // It is safe to assume that the track has at least one channel
-   assert(channelsCount > 0);
+   if (GetStretchRatio() == 1)
+      return dst;
 
-   for (unsigned channel = 0; channel < channelsCount; ++channel)
-      GetClip(channel)->ApplyStretchRatio(
-         [&reportProgress, channel, channelsCount](double progress) {
-            reportProgress((channel + progress) / channelsCount);
-         });
+   const auto originalPlayStartTime = GetPlayStartTime();
+   const auto originalPlayEndTime = GetPlayEndTime();
+   const auto stretchRatio = GetStretchRatio();
+
+   auto success = false;
+   Finally Do { [&] {
+      if (!success)
+      {
+         TrimLeftTo(originalPlayStartTime);
+         TrimRightTo(originalPlayEndTime);
+      }
+   } };
+
+   // Leave 1 second of raw, unstretched audio before and after visible region
+   // to give the algorithm a chance to be in a steady state when reaching the
+   // play boundaries.
+   const auto tmpPlayStartTime =
+      std::max(GetSequenceStartTime(), originalPlayStartTime - stretchRatio);
+   const auto tmpPlayEndTime =
+      std::min(GetSequenceEndTime(), originalPlayEndTime + stretchRatio);
+   TrimLeftTo(tmpPlayStartTime);
+   TrimRightTo(tmpPlayEndTime);
+
+   WideClip wideClip { mpClip, mpClip1 };
+
+   constexpr auto sourceDurationToDiscard = 0.;
+   constexpr auto blockSize = 1024;
+   const auto numChannels = NChannels();
+   ClipTimeAndPitchSource stretcherSource { wideClip, sourceDurationToDiscard,
+                                            PlaybackDirection::forward };
+   TimeAndPitchInterface::Parameters params;
+   params.timeRatio = stretchRatio;
+   StaffPadTimeAndPitch stretcher { mpClip->GetRate(), numChannels,
+                                    stretcherSource, std::move(params) };
+
+   // Post-rendering sample counts, i.e., stretched units
+   const auto totalNumOutSamples =
+      sampleCount { wideClip.GetVisibleSampleCount().as_double() *
+                    stretchRatio };
+
+   sampleCount numOutSamples { 0 };
+   AudioContainer container(blockSize, numChannels);
+
+   while (numOutSamples < totalNumOutSamples)
+   {
+      const auto numSamplesToGet =
+         limitSampleBufferSize(blockSize, totalNumOutSamples - numOutSamples);
+      stretcher.GetSamples(container.Get(), numSamplesToGet);
+      constSamplePtr data[2];
+      data[0] = reinterpret_cast<constSamplePtr>(container.Get()[0]);
+      if (NChannels() == 2)
+         data[1] = reinterpret_cast<constSamplePtr>(container.Get()[1]);
+      dst->Append(data, floatSample, numSamplesToGet);
+      numOutSamples += numSamplesToGet;
+      if (reportProgress)
+         reportProgress(
+            numOutSamples.as_double() / totalNumOutSamples.as_double());
+   }
+   dst->Flush();
+
+   // Now we're all like `this` except unstretched. We can clear leading and
+   // trailing, stretching transient parts.
+   dst->SetPlayStartTime(tmpPlayStartTime);
+   dst->ClearLeft(originalPlayStartTime);
+   dst->ClearRight(originalPlayEndTime);
+
+   // We don't preserve cutlines but the relevant part of the envelope.
+   Envelope dstEnvelope = GetEnvelope();
+   const auto samplePeriod = 1. / mpClip->GetRate();
+   dstEnvelope.CollapseRegion(
+      originalPlayEndTime, GetSequenceEndTime() + samplePeriod, samplePeriod);
+   dstEnvelope.CollapseRegion(0, originalPlayStartTime, samplePeriod);
+   dstEnvelope.SetOffset(originalPlayStartTime);
+   dst->SetEnvelope(dstEnvelope);
+
+   success = true;
+
+   assert(dst->GetStretchRatio() == 1);
+   return dst;
 }
 
 bool WaveTrack::Interval::StretchRatioEquals(double value) const
@@ -285,6 +410,11 @@ bool WaveTrack::Interval::IntersectsPlayRegion(double t0, double t1) const
    return mpClip->IntersectsPlayRegion(t0, t1);
 }
 
+bool WaveTrack::Interval::WithinPlayRegion(double t) const
+{
+   return mpClip->WithinPlayRegion(t);
+}
+
 double WaveTrack::Interval::GetStretchRatio() const
 {
    //TODO wide wave tracks:  assuming that all 'narrow' clips share common stretch ratio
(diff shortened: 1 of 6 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -213,19 +261,96 @@ void WaveTrack::Interval::StretchRightTo(double t)
...
+         reportProgress(
+            numOutSamples.as_double() / totalNumOutSamples.as_double());
+   }
+   dst->Flush();
+
+   // Ok to obliterates cutlines when stretch-rendering, but envelope must be
+   // preserved.
+   dst->SetPlayStartTime(tmpPlayStartTime);
+   Envelope dstEnvelope = GetEnvelope();
+   dstEnvelope.CollapseRegion(0, tmpPlayStartTime, 1 / mpClip->GetRate());
```
saintmatthieu, line 344: @Paul-Licameli this is actually not right, which is why this PR is still a draft. I couldn't figure it out before the end of the day unfortunately. Maybe you can spot what I'm doing wrong.

saintmatthieu: Fixed.

saintmatthieu:
Some comments about this PR that may help the understanding:
* `WaveClip` still being mono, it cannot be stretch-rendered independently from it's other channel. `WaveTrack::Interval` is now the place where this is done.
* The `StaffPadTimeAndPitch` processor needs a source and a sink. The sink sample size being different than the source, we cannot have the same `Interval` being at once source and sink. Hence the approach consisting of creating an empty copy, using it as sink, and afterwards copying over the envelope.

saintmatthieu:
Cutlines are not preserved. At the moment we have a (tiny?) bug: enable cutlines, stretch a clip, cut a portion of it, render stretching: the red line is still visible, and clicking it still makes it disappear, but does not expand anything. Now at least the red line isn't present anymore. I suppose it's okay to postpone the implementation of this to after this PR if Product think they're worthwhile preserving. If yes, the cutlines would probably have to be stretch-rendered, too ...

saintmatthieu:
About cutlines: @dozzzzer just logged an issue: #5401

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -613,7 +613,9 @@ void WaveTrackAffordanceControls::StartEditSelectedClipSpeed(
 void WaveTrackAffordanceControls::OnRenderClipStretching(
    AudacityProject& project)
 {
-   auto [track, it] = SelectedIntervalOfFocusedTrack(project);
+   const auto p = SelectedIntervalOfFocusedTrack(project);
+   auto track = p.first;
+   auto it = p.second;
```
Paul-Licameli, line 618: The compile on macOS doesn't let you capture a structured binding name directly in a lambda.
Another fix is to change the lambda capture below from `[&]` to `[track = track, interval = interval]`

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -201,6 +225,30 @@ void WaveTrack::Interval::TrimRightTo(double t)
       GetClip(channel)->TrimRightTo(t);
 }
 
+void WaveTrack::Interval::SetTrimLeft(double t)
+{
+   for(unsigned channel = 0; channel < NChannels(); ++channel)
+      GetClip(channel)->SetTrimLeft(t);
```
Paul-Licameli, line 231: `ForEachClip` might be used more consistently in these new functions, but that's not important.

Also I would redefine `ForEachClip` as an inline function, deducing the type of the argument as a template parameter, avoiding the construction of a `std::function` object.  But this is not important.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -318,6 +460,18 @@ bool WaveTrack::Interval::IsPlaceholder() const
...
+const Envelope& WaveTrack::Interval::GetEnvelope() const
+{
+   return *mpClip->GetEnvelope();
+}
+
+void WaveTrack::Interval::SetEnvelope(const Envelope& envelope)
+{
+   ForEachClip([&](auto& clip) {
+      clip.SetEnvelope(std::make_unique<Envelope>(envelope));
+   });
```
Paul-Licameli, line 472: Actually this is needed only for clip 0.  See the commits merged at 48154efe6d8ce8e02c570a665a31caab39786cdc.  Right channel envelopes are just ignored.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -213,19 +261,93 @@ void WaveTrack::Interval::StretchRightTo(double t)
       GetClip(channel)->StretchRightTo(t);
 }
 
-void WaveTrack::Interval::ApplyStretchRatio(
-   const std::function<void(double)>& reportProgress)
+WaveTrack::IntervalHolder WaveTrack::Interval::GetStretchRenderedCopy(
+   const std::function<void(double)>& reportProgress, const ChannelGroup& group,
+   const SampleBlockFactoryPtr& factory)
```
Paul-Licameli, line 266: This replacement for `WaveClip::ApplyStretchRatio` loses a few good things:
- Early quick exit when ratio is already 1
- pre- and post-conditions, stated in header comments, and asserted in the function

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -213,19 +261,93 @@ void WaveTrack::Interval::StretchRightTo(double t)
...
+      stretcher.GetSamples(container.Get(), numSamplesToGet);
+      constSamplePtr data[2];
+      data[0] = reinterpret_cast<constSamplePtr>(container.Get()[0]);
+      if (NChannels() == 2)
+         data[1] = reinterpret_cast<constSamplePtr>(container.Get()[1]);
+      // TODO I don't think so, but double-checking: the `excsafety{Strong}`
+      // comment of `Sequence::AppendNewBlock` doesn't mean the method can't
+      // throw, right? If it didn't throw, then we could get rid of the Finally
+      // and restore the trimming of `this` after this loop before the call to
+      // `Flush`.
```
Paul-Licameli, line 331: Answering:  Exception safety was carefully documented for those Sequence operations precisely because exceptions are possble (otherwise it would have been `excsafety{No-Fail}`).  It is a promise that in case of exceptions, the Sequence object is restored to its prior state.

A known problem with wide wave clips (such as we have them, with the WaveTrack::Interval facade) is that strong exception safety for the individual sequences does not imply strong exception safety of the wide clip as a whole when appending or flushing.  Maybe there is success for one channel but failure for the other, and then there is no reversion of the first channel to its previous state.

I did not attempt to solve this unlikely problem for 3.4 and neither should you.  Just maybe it will happen if you record and run out of space on the drive.  There is exception handling in this case to save as much of the recording as possible rather than throw away the entire recording.  This hole in strong exception safety might mean it's one remaining way to end up with mis-aligned clips.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -213,19 +261,93 @@ void WaveTrack::Interval::StretchRightTo(double t)
...
+         reportProgress(
+            numOutSamples.as_double() / totalNumOutSamples.as_double());
+   }
+   dst->Flush();
+
+   // Now we're all like `this` except unstretched. We can clear leading and
+   // trailing, stretching transient parts.
+   dst->SetPlayStartTime(tmpPlayStartTime);
+   dst->ClearLeft(originalPlayStartTime);
+   dst->ClearRight(originalPlayEndTime);
```
Paul-Licameli, line 344: The step `mRawAudioTempo = *mProjectTempo;` is lost in this version.  Should that be fixed?

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -213,19 +261,93 @@ void WaveTrack::Interval::StretchRightTo(double t)
...
+
+   // We don't preserve cutlines but the envelope.
```
Paul-Licameli, line 346: Cutlines?  Why not?

They might be preserved un-stretched.  Maybe nobody cares.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -213,19 +261,93 @@ void WaveTrack::Interval::StretchRightTo(double t)
...
+      // `Flush`.
+      dst->Append(data, numSamplesToGet);
+      numOutSamples += numSamplesToGet;
+      if (reportProgress)
+         reportProgress(
+            numOutSamples.as_double() / totalNumOutSamples.as_double());
+   }
+   dst->Flush();
+
+   // Now we're all like `this` except unstretched. We can clear leading and
```
Paul-Licameli, line 340: Not completely like `this`.  I noticed that the new constructor of `Interval` always uses `floatSample` format.  I don't see where you change format to match the track's when inserting the new interval.

Paul-Licameli (commented):
Tracks with non-float samples should be a test case

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -187,8 +189,30 @@ WaveTrack::Interval::Interval(const ChannelGroup &group,
...
+            nullptr)
+{
+}
+
 WaveTrack::Interval::~Interval() = default;
 
+void WaveTrack::Interval::Append(constSamplePtr buffer[], size_t len)
+{
+   for (unsigned channel = 0; channel < NChannels(); ++channel)
+      GetClip(channel)->AppendNewBlock(buffer[channel], floatSample, len);
```
Paul-Licameli, line 208: More about formats:  either the argument to this function should be `float *const *` or else the format should be an argument, not assumed to be `floatSample`.  The latter solution has the appeal of more generality easily implemented, even if not yet needed.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -3930,6 +4071,58 @@ bool WaveTrack::MergeOneClipPair(int clipidx1, int clipidx2)
...
+      });
+
+   // If we reach this point it means that no error was thrown - we can replace
+   // the source with the destination intervals.
+   for (auto i = 0; i < srcIntervals.size(); ++i)
+      ReplaceInterval(srcIntervals[i], dstIntervals[i]);
+}
+
+void WaveTrack::InsertInterval(const IntervalHolder& interval)
+{
```
Paul-Licameli, line 4094: The assertions in this and the next two functions should be stated in header comments as preconditions.

Paul-Licameli:
This completes a round of review.  I think the sample formats must surely be fixed, and also preserving the project tempo in the clip may be desirable.

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: `dst`, as a freshly-created empty clip, doesn't have project tempo set yet. It will be when inserted with `WaveTrack::InsertInterval`.

Paul-Licameli: I see.  Consider this point resolved.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: This is a design change.  Make sure designers agree.

Expansion of cutlines is done by WaveClip::Paste in WaveClip::ExpandCutLine.  Paste fails if ratios don't match.

So I think the complete solution would be correct remapping of cutlines along the timeline at the time stretch is applied, but then computing the stretch of the cutline could be delayed until just before the Paste.

Paul-Licameli:
Envelopes:  is it correct just to copy the envelope into the replacement clips, or do they have to be stretched too, as is done in WaveClip::OnProjectTempoChange?

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: Absolutely - I've been overseeing this argument. Addressing ...

Paul-Licameli: "Overseeing" or "overlooking"?  (Not the same...)

saintmatthieu:
> Envelopes: is it correct just to copy the envelope into the replacement clips, or do they have to be stretched too, as is done in WaveClip::OnProjectTempoChange?

Stretch-rendering should not affect the timing of envelopes, and envelope points, like `mTrimLeft` and the likes, are updated on each shrink or expand operation. So we're good 👍

libraries/lib-wave-track/WaveTrack.cpp
saintmatthieu: > This is a design change. Make sure designers agree.

@dozzzzer logged a follow-up ticket that still awaits Martin's (or Leo's, but he's on vacation) input.

Paul-Licameli:
> > Envelopes: is it correct just to copy the envelope into the replacement clips, or do they have to be stretched too, as is done in WaveClip::OnProjectTempoChange?
> 
> Stretch-rendering should not affect the timing of envelopes, and envelope points, like `mTrimLeft` and the likes, are updated on each shrink or expand operation. So we're good 👍

Well I'm adding a QA checkbox anyway

saintmatthieu:
> Tracks with non-float samples should be a test case

I see you added a QA item, thanks.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
saintmatthieu: Trying ... 🤞

Paul-Licameli (approved):
One remaining thing I wanted to see was an assertion of a postcondition.  I added it during your day off, in the interests of keeping this card moving through QA.

saintmatthieu:
> One remaining thing I wanted to see was an assertion of a postcondition. I added it during your day off, in the interests of keeping this card moving through QA.

I forgot to do that, thank you.

saintmatthieu:
Just pushed a fix for [this issue](https://github.com/audacity/audacity/assets/22740106/6ac6c650-3f0c-4cab-9f67-7e6e5a350c21) found by @dozzzzer (thank you!)

dozzzzer:
@saintmatthieu @Paul-Licameli 

- when rendering the clip that has trimmed envelope points and then dragging that clip, the trimmed envelope points re-appear:

https://github.com/audacity/audacity/assets/70063970/5e61ed5e-ba0c-45df-88c3-74e45dd55a19

saintmatthieu:
> @saintmatthieu @Paul-Licameli
> 
> * when rendering the clip that has trimmed envelope points and then dragging that clip, the trimmed envelope points re-appear:
> 
>  Screen.Recording.2023-10-27.at.11.58.55.video-converter.com.mp4

This video is linked in my previous comment.

dozzzzer:
@saintmatthieu @Paul-Licameli 

1. add some envelope points at the beginning of the clip
2. shrink/stretch the clip
3. trim away the envelope points
4. right-click and select "render clip stretching"
5. try to add more envelope points on the resulting clip

Actual result: only one envelope point can be added, at the very beginning of the clip


https://github.com/audacity/audacity/assets/70063970/dc9edbac-7241-47ec-b275-6c8c2a8a6c95

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -339,11 +339,12 @@ WaveTrack::IntervalHolder WaveTrack::Interval::GetStretchRenderedCopy(
    dst->ClearRight(originalPlayEndTime);
 
    // We don't preserve cutlines but the relevant part of the envelope.
-   Envelope dstEnvelope = dst->GetEnvelope();
+   Envelope dstEnvelope = GetEnvelope();
    const auto samplePeriod = 1. / mpClip->GetRate();
    dstEnvelope.CollapseRegion(0, originalPlayStartTime, samplePeriod);
    dstEnvelope.CollapseRegion(
       originalPlayEndTime, GetSequenceEndTime() + samplePeriod, samplePeriod);
```
Paul-Licameli, line 346: I think you may need to swap the two lines above to get correct results in all cases.  Because the function move envelope points after the collapsed part leftward.

saintmatthieu: I wouldn't expect that, but to be on the safe side I'll push that change.

Paul-Licameli:
Reapproved

Paul-Licameli:
Reapproved

dozzzzer:
Tested

PR #5408, opened 2023-10-22
Add a set of patches to better support OpenBSD and FreeBSD

brad0:
```
- When dlopening libraries on OpenBSD we need to drop the major version
- OpenBSD does not have sys/timeb.h. Note this has been removed from POSIX.1-2008
- Correct the mistake with the FreeBSD symbol, it should be __FreeBSD__
```

Recommended:


brad0:
I can't sign the CLA. The URL does not work.

LWinterberg:
> I can't sign the CLA. The URL does not work.

thanks for that, it should be fixed now. And thanks for the patch!

brad0:
OK, signed the CLA.

crsib (approved):

brad0:
Can this be back ported to the 3.4 branch?

PR #5415, opened 2023-10-23
Extract lib menus

Paul-Licameli:
Resolves: #5437

The difficult dependency cycle of CommandManager and Menus is now broken!

A toolkit-neutral library for declarative descriptions of menu trees.

Also some reduced dependencies in CommandManager.cpp, but it is
not yet toolkit neutral too because it depends on ProjectWindows.h, wx/app.h, wx/menu.h

QA:
- [x] Try "Get Info" macro command with each choice of "Format"
- [x] Verify pull-down menu tree is unchanged; including the optionally shown Extra menus
- [x] Verify menu shortcuts are unchanged (use the Export... button of Shortcuts preferences to compare builds)
- [x] File > Open Recent works as expected
- [x] Edit > Undo and Redo menu item texts update correctly after editing actions
- [x] Hover tooltips of the toolbar buttons show correct shortcut keys

Recommended:


Paul-Licameli:
Rebased onto master after merging its dependency

Let the review begin!

crsib (approved):

chinakov:
Tested

Issue #5402, opened 2023-10-20
Disable "inteligent" export type


SteveDaulton:
### Your idea

In 3.4 beta, if the project contains multiple tracks or any labels, "Export Multiple" is automatically selected.
I'm proposing that  this behaviour becomes optional.

### Problem to be solved

Almost all of my projects have either more than one track, or one or more labels. In almost all cases I want to export a mix of all non-muted tracks. I suspect that this is a common workflow used by many thousands of Audacity users.

With the new "intelligent" export type selection, I must remember to reset the export type to "Entire Project" every time I export, or the export will be wrong.

There's also a potential "gotcha":
Labels are frequently used for labelling parts of a project. The new behaviour has the potential for unwary users to lose work. For example:
1. Import a file
2. Edit the project and during the process add a region label to mark a region of interest
3. "Ctrl + Shift +  E"
4. Check the destination and file name
5. Click  "Export" -> "OK"
6. Close  the project without saving.
Unless you were paying close attention in step 4, only the labeled region will be exported and all of your editing work is lost.

### Prior art

All previous versions of Audacity used Export Multiple only when explicitly selected by the user.

### Additional context

The new behaviour is likely to be a nice convenience to a subset of Audacity users. Without a large scale poll it is impossible to know how big or small that subset is, but I suspect that it is a relatively small minority of users. Because of it's usefulness I agree that the new feature should be retained, but for the (possibly millions of ) users that have no use for the feature, it would be useful to be able to turn it off.

DavidBailes:
> In 3.4 beta, if the project contains multiple tracks or any labels, "Export Multiple" is automatically selected.
> I'm proposing that this behaviour becomes optional.

@SteveDaulton . Thanks for highlighting this behaviour. As far as I can see, "Export multiple" is automatically selected only if the project contains one or more labels, and not if it just contains multiple audio tracks.

As Steve has already said, I think that this behaviour is lightly to cause users to loose their work. Label tracks have many uses, not just the one of specifying how audio is split into a number of files. So a user could easily not notice that the Export range has been silently changed to "multiple files". This is particularly the case for users of screen readers.

@LWinterberg @Tantacrul . Please could this issue be reviewed before 3.4.0 is released?

petersampsonaudacity:
I totallt support @SteveDaulton 's and @DavidBailes 's points

SteveDaulton:
[...]

PR #5447, opened 2023-10-30
Always default export to "Entire Project"

crsib:
Resolves: #5402 

Recommended:

### DIFF
--- a/src/export/ExportAudioDialog.cpp
+++ b/src/export/ExportAudioDialog.cpp
@@ -182,6 +182,7 @@ ExportAudioDialog::ExportAudioDialog(wxWindow* parent,
    const auto labelTracks = tracks.Any<LabelTrack>();
    const auto hasLabels = !labelTracks.empty() &&
       (*labelTracks.begin())->GetNumLabels() > 0;
+   const auto hasMultipleWaveTracks = tracks.Any<WaveTrack>().size() > 1;
 
    if(ExportUtils::FindExportWaveTracks(tracks, true).empty() ||
       ViewInfo::Get(mProject).selectedRegion.isPoint())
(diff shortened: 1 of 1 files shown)


DavidBailes:
@crib. This fix doesn't always default the export range to the entire project. The default export range is now the last used setting.

crsib:
This is a kind of behavior that a lot of places in Audacity has. This can be changed easily, I'm not sure that it should.

In the very same dialog every other setting remembers the value used during the last export.

@LWinterberg please advise

vsverchinsky (approved):

petersampsonaudacity:
Testing on W10 with @crsib 's branch build:  audacity-win-3.4.0-alpha-20231030+c5fb3b0-x64

This works on first use if you have
a) multiple tracks
b) a non-empty label track
c) a selection
Then regardless if the above the "Entire Project" is properly offered 

BUT on next use it still remembers the previously used export range 0 an sometime inappropriately so - try these steps

1. purge the Audacity settings folder
2. get some audio - single track
3. add some labels
4. File > Export Audio
5. Observe: "Entire Project" is the default range offered
6. set the rang to "Multiple Files"
7. Export button to export the labeled ranges
8. Delete the Label Track
9. File > Export Audio
10. Observe: "Multiple Files" is still the default range offered (even though there a no labels and only a single track)

It's a bit smarter with Export - but not totally smart
1. purge the Audacity settings folder
2. get some audio - single track
3. select a portion of the track
4. File > Export Audio
5. Observe: "Entire Project" is the default range offered
6. set the range to be "Current selection"
7. Export button to export the selection
8. make a different selection
9. File > Export Audio
10. Observe: "Current selection" is the default range offered (remembered from last use)
11. Export button to export the selection
12. make a point "selection" - i.e. set a cursor position
13.  File Export >Audio
14. Observe: "Entire Project" is offered as the default range

LWinterberg:
I think an important bit for usability is whether the user is aware of this option. If you've never used export multiple before and suddenly find your files getting chopped up after placing labels, it's super confusing. However, with this PR, you need to have used the export multiple option before, and it's likely that you have used it within the same session. Getting your files chopped up in this case may be a bit annoying, but chances are you know exactly what happened and also know exactly how to fix it.

vsverchinsky (approved):

petersampsonaudacity:
Testing on W10 with @crsib 's latest updated branch build for this:  audacity-win-3.4.0-alpha-20231030+6cc55d0-x64

This now looks **_much_** better to me.

Yes it now remembers that previously used range setting but now only applies that if and when it is appropriate - if it is not appropriate then "Entire Project" is offered.

@LWinterberg @crsib @vsverchinsky This looks good to go to me

PR #5458, opened 2023-11-01
Bugfix: silence audio does not delete clips

### LINKED ISSUE #5456: Silence audio deletes the clip
### Bug description

_No response_

### Steps to reproduce

1. Create an audio track
2. Select some of the the audio
3. Edit>  Remove special > Silence audio

### Expected behavior

Selected audio is silence

### Actual behavior

All clips which intersect the time selection are deleted.

### Audacity Version

latest beta

### Operating system

Windows

### Additional context

Regression on 3.3.3

saintmatthieu:
Resolves: #5456

When stretch-rendering an interval (stereo clip) that didn't have any stretch, we mistakenly replaced the clips with empty clips.
Now not only do we check that we don't stretch-render intervals that are already unstretched, but we also return proper clip copies if `GetStretchRenderedCopy` is called on an unstretched interval.

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -266,11 +266,11 @@ WaveTrack::IntervalHolder WaveTrack::Interval::GetStretchRenderedCopy(
    const std::function<void(double)>& reportProgress, const ChannelGroup& group,
    const SampleBlockFactoryPtr& factory, sampleFormat format)
 {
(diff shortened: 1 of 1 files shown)


crsib (approved):
Well, that was quick

petersampsonaudacity:
Testing on W10 with @saintmatthieu 's branch build for 3.4.0 Beta:  audacity-win-3.4.0-alpha-20231101+1b8cac1-x64

This appear to properly fix this issue.

PR #5498, opened 2023-11-05
move join shortcut to standard set

LWinterberg:
this is probably fair given that the line clicking got axed


petersampsonaudacity:
Testing on W10 with @LWinterberg 's branch build:  audacity-win-3.4.1-alpha-20231105+4ab47c2-x64

Looks to be fixed Leo, with Ctr;+J for Join Clips in the Standard set of shortcuts.

chinakov:
Yeah, can confirm that 'CMD + J' command joins clips as well. @LWinterberg

Issue #5476, opened 2023-11-03
Failed to build on arm64/armhf


PandaJim:
### Bug description

The new 3.4.0 release built successfully in Ubuntu 22.04/23.10 on X86_64, but failed in arm64 and armhf.

There's no apparent error output, except `HAVE_MMX` and `HAVE_SSE/SSE2` test fail:
```
cd obj-aarch64-linux-gnu && cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DFETCHCONTENT_FULLY_DISCONNECTED=ON -DCMAKE_INSTALL_RUNSTATEDIR=/run "-GUnix Makefiles" -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/aarch64-linux-gnu -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_LIBDIR=lib -Daudacity_conan_enabled=Off -Daudacity_has_networking=Off -Daudacity_has_vst3=Off -Daudacity_lib_preference=system -Daudacity_use_ffmpeg=loaded -DwxBUILD_TOOLKIT=gtk3 -DAUDACITY_BUILD_LEVEL=2 ..
-- The C compiler identification is GNU 12.3.0
-- The CXX compiler identification is GNU 12.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
...
...
-- Looking for __atomic_fetch_add_4 in atomic
-- Looking for __atomic_fetch_add_4 in atomic - found
-- Performing Test HAVE_MMX
-- Performing Test HAVE_MMX - Failed
-- Performing Test HAVE_SSE
-- Performing Test HAVE_SSE - Failed
-- Performing Test HAVE_SSE2
-- Performing Test HAVE_SSE2 - Failed
-- Looking for 4 include files float.h, ..., string.h
-- Looking for 4 include files float.h, ..., string.h - found
...
...
...
[...]

PR #5503, opened 2023-11-06
Fixes the GCC build when targeting ARM CPUs with NEON support

crsib:
Resolves: #5476 

Recommended:


vsverchinsky (approved):

crsib:
This goes straight to merge:

1. It only affects Linux on ARM
2. I've checked it on RPi 4
3. QA doesn't have Linux on ARM ATM

Issue #5506, opened 2023-11-06
Rewrite WaveTrack::Join with interval-major iteration


Paul-Licameli:
Part of wide stereo clip cleanup

PR #5507, opened 2023-11-06
Rewrite join

Paul-Licameli:
Resolves: #5506

QA:
- [x] Edit > Clip > Join
- [x] Edit > Labels > Join

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -2386,7 +2386,7 @@ void WaveTrack::ApplyStretchRatio(
        !clipAtT1->StretchRatioEquals(1))
       Split(endTime, endTime);
 
-   std::vector<IntervalHolder> srcIntervals;
+   IntervalHolders srcIntervals;
    auto clip = GetIntervalAtTime(startTime);
    while (clip && clip->GetPlayStartTime() < endTime)
    {
@@ -2577,7 +2577,7 @@ void WaveTrack::Join(
    assert(IsLeader());
    // Merge all WaveClips overlapping selection into one
    const auto intervals = Intervals();
-   std::vector<IntervalHolder> intervalsToJoin;
+   IntervalHolders intervalsToJoin;
    for (auto interval : intervals)
       if (interval->IntersectsPlayRegion(t0, t1))
          intervalsToJoin.push_back(interval);
@@ -4106,10 +4106,10 @@ bool WaveTrack::MergeOneClipPair(int clipidx1, int clipidx2)
 }
 
 void WaveTrack::ApplyStretchRatioOnIntervals(
-   const std::vector<IntervalHolder>& srcIntervals,
+   const IntervalHolders& srcIntervals,
    const ProgressReporter& reportProgress)
 {
-   std::vector<IntervalHolder> dstIntervals;
+   IntervalHolders dstIntervals;
    dstIntervals.reserve(srcIntervals.size());
    std::transform(
       srcIntervals.begin(), srcIntervals.end(),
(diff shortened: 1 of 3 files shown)


Paul-Licameli:
Draft still -- violating assertions when I record.  Wait for a fix ...

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1860,6 +1867,61 @@ namespace
...
+   for (const auto end = list.end(); it != end; ++it) {
+      if (it->get() == &clip)
+         break;
+      ++distance;
+   }
+   if (pDistance)
+      *pDistance = distance;
+   WaveClipHolders::const_iterator it2{};
+   if (pList2)
+      it2 = pList2->begin() + distance;
```
saintmatthieu, line 1889: This assumes that the insertion order of clips in each of the left- and right channel tracks is kept the same. What makes you confident that it is? Did you consider an approach comparing metadata (e.g. clip offset and length?)

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1860,6 +1867,61 @@ namespace
...
+   return { it, it2 };
+}
+
+void WaveTrack::RemoveWideClip(IterPair pair)
+{
+   mClips.erase(pair.first);
+   const auto channels = TrackList::Channels(this);
+   if (channels.size() > 1)
+      (*channels.rbegin())->mClips.erase(pair.second);
```
saintmatthieu, line 1921: This assumes that `pair.second` really is an iterator obtained from the same leader-follower instance. Safer would be to modify the integration of the parts such that `RemoveWideClip(*clip->GetClip(0));` can be written (instead of current `RemoveWideClip(FindWideClip(*clip->GetClip(0)));`).

saintmatthieu (approved):
Looks good to me. I left some suggestions that I'd like to be considered to increase safety and remove code duplication.

Paul-Licameli:
I will not apologize for making the simplifying assumption that clips do have the same insertion order in corresponding tracks.

Other than adding a vector bounds check, I do not want to overcomplicate the transitional code we must write for this refactoring.  If you take on any of the other "difficult" channel iteration rewrites, you should do the same.

We already did the work in version 3.4 to ensure this condition.

See #5202, #5211, #5200.

Paul-Licameli:
But making a refactoring branch may be a good idea.

saintmatthieu:
Ok. Please consider the other comments nevertheless.

saintmatthieu (approved):

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1860,6 +1867,61 @@ namespace
...
+   }
+   if (pDistance)
+      *pDistance = distance;
+   WaveClipHolders::const_iterator it2{};
+   if (pList2)
+      it2 = pList2->begin() + distance;
+   return { it, it2 };
+}
+
+auto WaveTrack::FindWideClip(const WaveClip &clip, int *pDistance)
```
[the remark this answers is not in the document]
Paul-Licameli: Will do

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: Both new functions are `private` so there is a bounded amount of code to examine to ensure the correctness.  I wouldn't worry about it.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: I'll do a bounds check

Paul-Licameli:
Rebased, and changed the base branch, and inserted two FIXUPs answering the review.

PR #5510, opened 2023-11-06
5483 crossfade track bugs

### LINKED ISSUE #5483: Crossfade Track bugs
### Bug description

_No response_

### Steps to reproduce

Testing with either a couple of mono of tracks or couple of stereo tracks.
For example mono tracks:
image

Case A:
Two mono tracks, no time stretching.

Case B:
Two mono tracks, time stretching.

Case C:
Two stereo tracks.

Case D:
stereo tracks with time stretching

### Expected behavior

Case A:
The same as the actual behaviour. No problem here.

Case B:
Something like the actual behaviour of case A:

Case C:
Something happens.

Case D:
Something happens.

### Actual behavior

Case A:
image

Case B:
image

Case C:
The effect has no effect.
image

Case D:
The effect has no effect.
image

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

_No response_

saintmatthieu:
Resolves: #5483

When applying an effect to a track, the selection first gets stretch-rendered, which may introduce new splits at the beginning and end of the selection.
The problem is: the split happens at times rounded to the sample period, meanwhile the selection is not rounded.
In the particular case of the crossfade, the nyquist script checks that, within the selection, there is exactly one split. Depending on the rounding, the new stretch-rendering splits may fall into the selection boundaries, yielding the error "Invalid selection. More than 2 audio clips selected."

Proposed solution: modify the selection start and end values passed to the script by rounding them to the current track rate.

Recommended:

QA: Note that, to test this, make sure you don't have the snap option checked.

### DIFF
--- a/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/libraries/lib-effects/EffectOutputTracks.cpp
@@ -103,6 +103,15 @@ Track *EffectOutputTracks::AddToOutputTracks(TrackList &&list)
    return result;
 }
 
+const Track* EffectOutputTracks::GetMatchingInput(const Track& outTrack) const
+{
+   const auto it = std::find(mOMap.begin(), mOMap.end(), &outTrack);
+   if (it == mOMap.end())
+      return nullptr;
+   const auto index = it - mOMap.begin();
+   return mIMap[index];
+}
+
 // Replace tracks with successfully processed mOutputTracks copies.
 // Else clear and delete mOutputTracks copies.
 void EffectOutputTracks::Commit()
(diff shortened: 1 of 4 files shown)


crsib (dismissed):

saintmatthieu:
@crsib sorry, I need another round of review. I noticed that the crossfade track command was still broken for stretched clips. This is because the preliminary stretch-rendering creates new clip boundaries at the selection, which confused the fade-in/fade-out decision logic in `crossfadetrack.ny`.

Paul-Licameli (approved):
All looks correct to me for fixing the bug.

`'inclips` is now a new property of `*track*`.  Maybe a different Lisp symbol name for that could be clearer, such as `'pre-stretched-clips` (note, Lisp convention uses - not _ to separate words).

@SteveDaulton will want to know about this and document it in Nyquist reference.

Issue #5389, opened 2023-10-18
Transporting Audio Clips to Other Stereo Tracks Splits Tracks to Mono


chinakov:
@Paul-Licameli  UPDATE: please verify this bug remains fixed in the branch #5389

### Bug description

If the User has multiple tracks within a project, and then drags one of the clips within another track, it will sporadically split the tracks to mono, with no panning being applied.

#### Scenario 1 (apparently, easier to reproduce, thx @petersampsonaudacity )

1. launch 3.4.0
2. get some stereo audio
3. select a section and split the clip
4. add two stereo tracks
5. drag the middle clip down to the lower stereo track

#### Scenario 2

1. Open Audacity
2. Create new Project
3. Create Stereo Track 1 
4. Import or Generate Audio within Stereo Track 1 
5. Create Stereo Track 2
6. Import or Generate Audio within Stereo Track 2
7. Grab the audio clip within Stereo Track 2 and 'drag and drop' it into Stereo Track 1 
8. Do it a bunch of times, and observe different results	

### Expected behavior

No stereo tracks are split when clips are re-arranged from different tracks.	

#### Scenario 1

The originating track and the middle stereo track (empty one) get split to mono

<img width=600 src=https://github.com/audacity/audacity/assets/70063970/2ca98a18-332c-406d-9ad0-02314b155595>

#### Scenario 2

The tracks are split sporadically if the User re-arranges clips to different Stereo tracks. 

https://github.com/audacity/audacity/assets/139755141/be1f9500-892f-4543-a73c-52317737bbe5

### Audacity Version

latest beta

### Operating system

Windows

Paul-Licameli:
Bisects to 5fe2ce1f8e141d18805db8e194fa07441a115e26

dozzzzer:
Tested

petersampsonaudacity:
@Paul-Licameli @crsib 

I note that the build for this fix for 3.4.1 has partially failed.  
https://github.com/audacity/audacity/actions/runs/6783530050

JohnColket:
Forum user Davidcalder64 reports this issue is still present in 3.4.1: https://forum.audacityteam.org/t/3-4-bug-why-is-audacity-splitting-stereo-tracks-into-separate-tracks-for-each-channel-when-i-move-other-files-in-other-tracks/88714/6

petersampsonaudacity:
Yes my understanding is that the fix is coming in the impending 3.4.2 maintenance release.

Peter.

[...]

PR #5512, opened 2023-11-06
Fix unpredictable splitting of stereo tracks to mono...

Paul-Licameli:
... when dragging clips from track to track.

WaveTrack::mLegacyFormat was uninitialized, which explains the intermittency.

Resolves: #5389

Recommended:


crsib (approved):

vsverchinsky:
Alternatively `LinkConsistencyFix` could be removed from `WaveTrackShifter` since we remove support for unaligned tracks

petersampsonaudacity:
@Paul-Licameli @crsib

I note that the build for this fix for 3.4.1 has partially failed.
https://github.com/audacity/audacity/actions/runs/6783530050

crsib:
I'm rerunning the build now

PR #5524, opened 2023-11-07
new-clip-on-paste behavior pref, false by default

### LINKED ISSUE #5521: Create preference for paste-as-clip vs paste-into-clip
Located in: Preferences > Track Behaviors > Behaviors

Option name: "Always paste audio as new clips"

Option either restores pre-3.4 behavior or keeps 3.4 behavior. The old behavior is the default

saintmatthieu:
Resolves: #5521

Recommended:


src/prefs/TracksBehaviorsPrefs.cpp
```diff
@@ -101,9 +101,9 @@ void TracksBehaviorsPrefs::PopulateOrExchange(ShuttleGui & S)
       S.TieCheckBox(XXO("Enable &dragging selection edges"),
                     {wxT("/GUI/AdjustSelectionEdges"),
                      true});
-      S
-         .TieCheckBox(XXO("Editing a clip can &move other clips"),
+      S.TieCheckBox(XXO("Editing a clip can &move other clips"),
             EditClipsCanMove);
+      S.TieCheckBox(XXO("Ask Leo"), { wxT("/GUI/AskLeo"), false });
```
saintmatthieu, line 106: @LWinterberg awaiting exact wording.

Paul-Licameli:
This draft looks right so far, of course needing real strings

saintmatthieu:
There is still one tiny behaviour difference, which I think is rather a relief, though (but it also could be that my judgment is biased by laziness since it'd also mean more code changes and more risk of introducing new bugs):
Screenshot 2023-11-07 at 18 10 18

3.3.3: paste clipboard data onto selection, on either track, and the not-enough-room error pops up.
3.4.2 (with new setting off): paste clipboard data onto selection on
* upper track: the rightmost clip moves right.
* lower track: the not-enough-room error pops up.

So it'd be more lenient. @LWinterberg what do you think?

src/prefs/TracksBehaviorsPrefs.cpp
LWinterberg: "Always paste audio as new clips"

LWinterberg:
LGTM!

Paul-Licameli (approved):

saintmatthieu:
... cherry-picked onto 3.4.2.

Issue #5585, opened 2023-11-11
Channel-major rewrite of WaveTrack::PasteInto, WaveTrack::PasteWaveTrackAtSameTempo


Paul-Licameli:

chinakov:
Moving back to In Progress, as generating effects do not actually generate audio clips, instead just showing the appropriate selection on the timeline.

chinakov:
Found 2 issues, commented on the attached MR.

chinakov:
Tested and done! Moving to Ready to Merge.

Paul-Licameli:
Bugs found by @dozzzzer blocking the merge of #5742 were actually introduced with the merge of #5586.

I failed to make clear (or myself, understand) that I intended to change only the case of pasting to a point.  Pasting over an interval follows other paths.

The checklist of #5586 is now reproduced above.

PR #5586, opened 2023-11-11
Rewrite WaveTrack pasting as interval-major

Paul-Licameli:
Resolves: #5585

QA:  all the clip edge cases, mono and stereo...
- [x] Pasted-into track becomes selected
- [x] Cross-project paste, with and without project tempo differences
- [x] Paste with and without stretch ratio differences
- [x] With and without the new merging preference
- [x] Pasting into a clip, within one sample interval of time 0
- [x] With and without the "Edit clips can move" preference; test the prohibition
- [x] Moving of multiple other clips rightward to make room for the paste
- [x] New clips get correct names, and also when splitting to mono after the paste

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -3867,18 +3867,6 @@ const WaveClip* WaveTrack::GetClipAtTime(double time) const
       clips.rbegin(), clips.rend(), [&](const WaveClip* const& clip) {
          return clip->WithinPlayRegion(time);
       });
-
-   // When two clips are immediately next to each other, the GetPlayEndTime() of the first clip
-   // and the GetPlayStartTime() of the second clip may not be exactly equal due to rounding errors.
-   // If "time" is the end time of the first of two such clips, and the end time is slightly
-   // less than the start time of the second clip, then the first rather than the
-   // second clip is found by the above code. So correct this.
-   if (p != clips.rend() && p != clips.rbegin() &&
-      time == (*p)->GetPlayEndTime() &&
-      (*p)->SharesBoundaryWithNextClip(*(p-1))) {
-      p--;
-   }
-
    return p != clips.rend() ? *p : nullptr;
 }
 
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -1349,7 +1349,7 @@ bool WaveTrack::IsEmpty(double t0, double t1) const
       return true;
 
    //wxPrintf("Searching for overlap in %.6f...%.6f\n", t0, t1);
-   for (const auto &clip : mClips)
+   for (const auto &&clip : Intervals())
    {
       if (clip->IntersectsPlayRegion(t0, t1)) {
          //wxPrintf("Overlapping clip: %.6f...%.6f\n",
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -380,6 +380,11 @@ bool WaveTrack::Interval::StretchRatioEquals(double value) const
    return true;
 }
 
+bool WaveTrack::Interval::HasEqualStretchRatio(const Interval& other) const
+{
+   return StretchRatioEquals(other.GetStretchRatio());
+}
+
 /** Insert silence at the end, and causes the envelope to ramp
     linearly to the given value */
 void WaveTrack::Interval::AppendSilence(double len, double envelopeValue)
@@ -520,6 +525,11 @@ bool WaveTrack::Interval::WithinPlayRegion(double t) const
    return mpClip->WithinPlayRegion(t);
 }
 
+bool WaveTrack::Interval::SplitsPlayRegion(double t) const
+{
+   return mpClip->SplitsPlayRegion(t);
+}
+
 double WaveTrack::Interval::GetStretchRatio() const
 {
    //TODO wide wave tracks:  assuming that all 'narrow' clips share common stretch ratio
@@ -566,6 +576,11 @@ double WaveTrack::Interval::GetTrimRight() const
 bool WaveTrack::Interval::IsPlaceholder() const
 {
    return mpClip->GetIsPlaceholder();
+}
+
+void WaveTrack::Interval::MarkChanged()
+{
+   ForEachClip([](auto &clip){ clip.MarkChanged(); });
 }
 
 const Envelope& WaveTrack::Interval::GetEnvelope() const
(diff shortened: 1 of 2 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -480,6 +500,17 @@ bool WaveTrack::Interval::IsPlaceholder() const
...
 }
 
+void WaveTrack::Interval::MarkChanged()
+{
+   ForEachClip([](auto &clip){ clip.MarkChanged(); });
+}
+
+void WaveTrack::Interval::Resample(int rate, BasicUI::ProgressDialog *progress)
+{
+   // TODO:  correct the denominator of the progress, when it's not null
```
saintmatthieu, line 510: Such progress is also used in `WaveTrack::Resample` in a way that doesn't seem correct to me (not taking the number of clips into account and the progression from one clip to the next). I was tempted to add a QA item for that but I expect that our testers won't be happy and yet this wouldn't be a regression.

saintmatthieu (approved):

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2496,6 +2517,180 @@ void WaveTrack::InsertSilence(double t, double len)
    }
 }
 
+void WaveTrack::PasteOne(
```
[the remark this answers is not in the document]
Paul-Licameli: Correct.

I will remove more uses of `PasteOne` in two more PRs, keeping each PR not too large, not too slow to review, or to test.  Know that they are already in preparation.

Don't insist that I do all this sequentially.  I may want to have other PRs changing other parts of this big file also in flight concurrently through the project board.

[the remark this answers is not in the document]
Paul-Licameli: See #5602

Paul-Licameli:
Rebased, after merging in the lastest hot-fix changes to WaveTrack.cpp, to be sure of no conflicts or need for additional changes -- there were none.

saintmatthieu (approved):

Paul-Licameli:
@saintmatthieu please review again.

I will squash it into fewer commits.

Importantly I had to figure out a silly mistake that put duplicate shared pointers to the same clip into the same track, fixed by one deleted line.

This PR is important to unblock other PRs dependent on it.  I hope it moves soon back to testing.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2234,7 +2234,7 @@ void WaveTrack::PasteWaveTrackAtSameTempo(
 
     //wxPrintf("paste: we have at least one clip\n");
 
-    const auto clipAtT0 = track.GetIntervalAtTime(t0);
+    const auto clipAtT0 = track.GetWideClipAtTime(t0);
```
saintmatthieu, line 2237: I wonder why `GetIntervalAtTime` isn't sufficient.
In fact, looking at the `GetWideClipAtTime` implementation, I see logic that might have been relevant before the 2-of-6 PR, but which now seems futile to me. More about this in a follow-up comment closer to that implementation.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -3979,6 +3979,29 @@ const WaveClip* WaveTrack::GetClipAtTime(double time) const
...
+   // When two clips are immediately next to each other, the GetPlayEndTime() of the first clip
+   // and the GetPlayStartTime() of the second clip may not be exactly equal due to rounding errors.
+   // If "time" is the end time of the first of two such clips, and the end time is slightly
+   // less than the start time of the second clip, then the first rather than the
+   // second clip is found by the above code. So correct this.
+   if (p != clips.rend() && p != clips.rbegin() &&
+      time == (*p)->GetPlayEndTime() &&
+      (*p)->GetClip(0)->SharesBoundaryWithNextClip((*(p-1))->GetClip(0).get()))
+   {
+      --p;
```
saintmatthieu, line 3999: Can this ever be executed?
Say `clipB->WithinPlayRegion(t)` returns true, it then becomes `p`. Then further down you test that `t == p->GetPlayEndTime()`, which can never be true, since `p->WithinPlayRegion(t) == true` implies that `t < p->GetPlayEndTime()` (or there is a bug).

saintmatthieu (commented):
Review iteration complete.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: I want to make only a conservative refactoring that changes the iteration to interval major.

I did not spend my time reasoning through whether the old extra logic in `GetClipAtTime` is still important.  I assumed it was.

If you think this should simplify, and can prove it, and can also understand why @DavidBailes put that extra logic there, and can convince him too that it is no longer needed -- then I suggest you make a PR for that purpose.

But I believe that should not block approval of this more urgent cleanup.

Paul-Licameli: > Can this ever be executed? Say `clipB->WithinPlayRegion(t)` returns true, it then becomes `p`. Then further down you test that `t == p->GetPlayEndTime()`, which can never be true, since `p->WithinPlayRegion(t) == true` implies that `t < p->GetPlayEndTime()` (or there is a bug).

Did you notice that the iteration through the clips was in backward order?

saintmatthieu: https://github.com/audacity/audacity/pull/5586#discussion_r1404086453 intended to prove that already, and to my sense did. From what I read you haven't examined if that extra old logic was useful or not, and yet you introduced a new method replicating this. If `GetIntervalAtTime` didn't exist before I wouldn't mind so much, but it does, and introducing an alternative and even replacing existing usages suggests there's something wrong with it.

saintmatthieu: > Did you notice that the iteration through the clips was in backward order?

I don't see how this impacts the reasoning in https://github.com/audacity/audacity/pull/5586#discussion_r1404086453.

DavidBailes: > Can this ever be executed? Say `clipB->WithinPlayRegion(t)` returns true, it then becomes `p`. Then further down you test that `t == p->GetPlayEndTime()`, which can never be true, since `p->WithinPlayRegion(t) == true` implies that `t < p->GetPlayEndTime()` (or there is a bug).

@saintmatthieu 
The last time I changed this this function the initial lines were:
```
const auto clips = SortedClipArray();
   auto p = std::find_if(clips.rbegin(), clips.rend(), [&] (WaveClip* const& clip) {
      return time >= clip->GetStartTime() && time <= clip->GetEndTime(); });
```
Since then, it has been changed so that if a time is equal to the time of the end of the clip it is not considered to be in the clip.
So as you say, the lines that correct for rounding errors are no longer necessary.

Paul-Licameli: > > Can this ever be executed? Say `clipB->WithinPlayRegion(t)` returns true, it then becomes `p`. Then further down you test that `t == p->GetPlayEndTime()`, which can never be true, since `p->WithinPlayRegion(t) == true` implies that `t < p->GetPlayEndTime()` (or there is a bug).
> 
> @saintmatthieu The last time I changed this this function the initial lines were:
> 
> ```
> const auto clips = SortedClipArray();
>    auto p = std::find_if(clips.rbegin(), clips.rend(), [&] (WaveClip* const& clip) {
>       return time >= clip->GetStartTime() && time <= clip->GetEndTime(); });
> ```
> 
> Since then, it has been changed so that if a time is equal to the time of the end of the clip it is not considered to be in the clip. So as you say, the lines that correct for rounding errors are no longer necessary. (And the times of the start and end of clips are now rounded to multiples of samples.)

You made changes several years ago, and I don't remember the motivating use case to test.  Do you?  Was it clip movement with the keys?

DavidBailes: Yes, it was clip movement using keystrokes. It was to ensure if a user moved to the end of a clip which was immediately adjacent to a following clip, that the keystrokes to move clips would move the second clip and not the first. The definition of  WaveClip::WithinPlayRegion() ensures that this will happen.

chinakov:
Found one issue so far: when a User tries to paste a smart clip, either on:

- same track, 
- new track
- cross-project with the same or different tempo

the smart clip no longer contains the hidden data. 

This issue persists on both mono and stereo tracks.

saintmatthieu (approved):
No bug detected but zero to two simplifications.

chinakov:
Tested, all issues fixed. Thanks!

Issue #5198, opened 2023-09-15
Difficult WaveTrack-internal channel iteration rewrites (checklist)


Paul-Licameli:
Iterations over `TrackList::Channels` in `WaveTrack.cpp` that are more difficult to transform from channel-major to interval-major iterations, often because the set of intervals is changing.

- [x] #5834
- [x] #5601
- [x] #5585
- [x] Clone
- [x] Copy
- [x] #5742
- [x] #5751
- [x] #5506
- [x] #5624
- [x] DoGet
- [x] #5554
- [x] #5633
- [x] #5652
- [x] #5693

Paul-Licameli:
More uses of `TrackList::Channels` in `WaveTrack.cpp` might be rewritten using outer loops over intervals and lowering loops over channels into new member functions of `WaveTrack::Interval`, as described in #5065.

The principle is the same but the code transformations may be more involved.

- [ ] Clear
- [ ] ClearAndAddCutline
- [ ] ClearAndPaste
- [ ] SplitDelete
- [ ] ...

Paul-Licameli:
This sub-list of check boxes of the big stereo clip refactoring is now all done after #5965, which included all the checkboxes not yet broken out into their own issues.

PR #5620, opened 2023-11-15
Move handleclear to interval

ycarmon:
Resolves: https://github.com/audacity/audacity/issues/5198

Recommended:


Paul-Licameli:
Thank you for contributing.  Unfortunately you misunderstand the problem to be solved.  I can't use this PR.

PR #5642, opened 2023-11-17
Two modules independent of audacity

Paul-Licameli:

Two of the import-export modules can break their dependency on the Audacity executable, with only two .cpp files moved from Audacity to a new, "red" library, lib-preference-pages.

If these ideas are pursued, that library might turn "green." https://github.com/audacity/audacity/discussions/5628

Recommended:


Paul-Licameli:
Please give a quick look, guys.  Like it?  Nominate it for the 3.5 board.

modules/mod-mp3/ExportMP3.cpp
```diff
@@ -1702,9 +1701,6 @@ bool MP3ExportProcessor::Initialize(AudacityProject& project,
    context.channels = channels;
 
    int rate = lrint(sampleRate);
-#ifndef DISABLE_DYNAMIC_LOADING_LAME
```
crsib, line 1705: It is really odd that this code is still here, provided that Audacity links against lame directly for quite a few years. I think that we need to revise ExportMP3 code.

Paul-Licameli: That could be separately addressed of course.  What do you think of this PR otherwise?

crsib (approved):
Provided that this PR is quite simple I think we should take it

Paul-Licameli:
I showed you what's possible in the modules dependency graph with a little work.

I moved only the minimal necessary files into lib-preference-pages for this.

If more pages might be easily moved into the library, perhaps they should, but then the library will serve miscellaneous purposes, grouping rather unrelated preferences together.

Paul-Licameli:
This PR was approved but forgotten.  I rebased it and fixed conflicts.  Please reconsider.

grliszas14 (approved):

dozzzzer:
Tested

PR #5644, opened 2023-11-17
Rewrite reverse as interval-major

Paul-Licameli:
Resolves: #5633

QA: Test the Reverse effect, with mono and stereo tracks; use visually dissimilar channels for stereo
- [x] Selection is inside a clip
- [x] Selection overlaps the end of a clip
- [x] Selection contains one or more clips
- [x] Selection overlaps the start of a clip

Note the last 3 are not mutually exclusive cases.

Recommended:


libraries/lib-wave-track/WaveTrack.h
```diff
@@ -468,7 +484,13 @@ class WAVE_TRACK_API WaveTrack final
    /*!
     @pre `IsLeader()`
     */
-   void Split(double t0, double t1) /* not override */;
+   void Split(double t0, double t1);
+
+   /*!
+    @pre `IsLeader()`
+    */
+   std::pair<IntervalHolder, IntervalHolder> SplitAt(double t);
```
saintmatthieu, line 492: I see the first returned `IntervalHolder` is never used. Also, I expect that the first `IntervalHolder`, if not null, points at existing clips? I find this would give even more strength to a simpler API that returns just one holder, to a _newly created object_, if any.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -4201,198 +4233,18 @@ void WaveTrack::Resample(int rate, BasicUI::ProgressDialog *progress)
...
-      )) {
-         rc = false;
-         break;
-      }
+   bool result = true;
+   size_t ii = 0;
+   for (const auto &pChannel : Channels()) {
+      const auto buffer = buffers[ii++];
+      assert(buffer); // precondition
+      result = pChannel->SetFloats(buffer, start, len, effectiveFormat)
```
saintmatthieu, line 4244: I think you meant to use an `&=` (not `|=`), or this method will always return true.

saintmatthieu (approved):
One mistake to address and one wish, but besides that LGTM.

libraries/lib-wave-track/WaveTrack.h
Paul-Licameli: No.  This API makes more sense to me, even if there is not yet any use for the first returned clip.  A split point defines two neighboring clips.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: OK
With just that change, I assume your tacit reapproval

Issue #5652, opened 2023-11-19
Remove WaveTrack::SetWaveColorIndex


Paul-Licameli:
Rather than rewriting this one as interval-major, better to remove completely from lib-wave-track, any mention of extraneous display related things.

Some surprisingly roundabout solutions are needed with attached objects on WaveTrack and WaveClip, but they may prove useful in future to attach other extra information to tracks and clips non-intrusively.

chinakov:
There is one last issue persistent in this MR, putting the task back to 'In Progress'. 

Thanks! 
![image](https://github.com/audacity/audacity/assets/139755141/4a1b28f7-17a5-402f-bff0-4db7b4054e44)

Issue #5530, opened 2023-11-08
Complete ChannelView rewrites (checklist)


Paul-Licameli:
A step in wide wave clip refactoring.  Some goals:

- Eliminate some uses of `TrackList::Channels` in display and mouse interaction.
- Use Channels, not Tracks.
- Remove the `ReallyGetChannelIndex` function introduced at 2fcbf3a05542520e621f722e5c956a943234829b.  Instead, `GetChannelIndex` will be fixed to give the "real" answer always.

Sub-issues:
- [x] #5772
- [x] #5774
- [x] #5776
- [x] #5778

PR #5653, opened 2023-11-19
Wave color nonsense

Paul-Licameli:
Resolves: #5652
Depends on #5828

QA:
- [x] Changing of wave track color with the track's menu
- [x] Mixing-and-rendering keeps color setting of the first mixed track;  then splitting to mono channels keeps same in both
- [ ] Recording or generating into a track uses the wave track color
- [x] Dragging clips from track to track preserves clip color, even if one track then has multiple colors
- [x] Mix-and-render names the new track as before (either "Mix" + number or keeping name of the sole track)
- [x] "Get Info" macro command for clips reports correct colors
- [x] "Set Clip" macro command, for colors
- [x] "Set Track Visuals" macro command, for colors
- [x] Cut/Copy then Paste of tracks (in or cross-project) preserves colors
- [x] Pasting or generating sound into an existing clip does not change its color
- [x] When a clip lengthens for sync lock with another track, its color does not change
- [x] Splitting a clip colors both as the original
- [x] Undo and Redo of all of the above
- [x] Waveform, Spectrogram panning and zooming are not slower
- [x] Persistency of wave track color settings, and of clip color
- [x] Opening in 3.4, a project that was saved in 3.5, works with correct colors; and oppositely

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -33,8 +33,16 @@
 #include <omp.h>
 #endif
 
-WaveClipListener::~WaveClipListener()
+WaveClipListener::~WaveClipListener() = default;
+
+void WaveClipListener::WriteXMLAttributes(XMLWriter &) const
 {
+}
+
+bool WaveClipListener::HandleXMLAttribute(
+   const std::string_view &, const XMLAttributeValueView &)
+{
+   return false;
 }
 
 WaveClip::WaveClip(size_t width,
@@ -70,6 +78,11 @@ WaveClip::WaveClip(
    mTrimRight = orig.mTrimRight;
    mRate = orig.mRate;
    mColourIndex = orig.mColourIndex;
+
+   // Deep copy of attachments
+   Attachments &attachments = *this;
+   attachments = orig;
+
    mSequences.reserve(orig.GetWidth());
    for (auto &pSequence : orig.mSequences)
       mSequences.push_back(
@@ -119,7 +132,12 @@ WaveClip::WaveClip(
       mTrimRight = orig.mTrimRight;
 
    mRate = orig.mRate;
+
    mColourIndex = orig.mColourIndex;
+
+   // Deep copy of attachments
+   Attachments &attachments = *this;
+   attachments = orig;
 
    mIsPlaceholder = orig.GetIsPlaceholder();
 
@@ -369,7 +387,7 @@ constSamplePtr WaveClip::GetAppendBuffer(size_t ii) const
 
 void WaveClip::MarkChanged() // NOFAIL-GUARANTEE
 {
-   Caches::ForEach( std::mem_fn( &WaveClipListener::MarkChanged ) );
+   Attachments::ForEach(std::mem_fn(&WaveClipListener::MarkChanged));
 }
 
 std::pair<float, float> WaveClip::GetMinMax(size_t ii,
@@ -571,6 +589,11 @@ bool WaveClip::HandleXMLTag(const std::string_view& tag, const AttributesList &a
                return false;
             SetColourIndex(longValue);
          }
+         else if (Attachments::FindIf(
+            [&](WaveClipListener &listener){
+               return listener.HandleXMLAttribute(attr, value); }
+         ))
+            ;
       }
       return true;
    }
@@ -638,6 +661,9 @@ void WaveClip::WriteXML(XMLWriter &xmlFile) const
    xmlFile.WriteAttr(wxT("clipStretchRatio"), mClipStretchRatio, 8);
    xmlFile.WriteAttr(wxT("name"), mName);
    xmlFile.WriteAttr(wxT("colorindex"), mColourIndex );
+   Attachments::ForEach([&](const WaveClipListener &listener){
+      listener.WriteXMLAttributes(xmlFile);
+   });
 
    for (auto &pSequence : mSequences)
       pSequence->WriteXML(xmlFile);
@@ -1198,7 +1224,7 @@ void WaveClip::Resample(int rate, BasicUI::ProgressDialog *progress)
       mSequences = move(newSequences);
       mRate = rate;
       Flush();
-      Caches::ForEach( std::mem_fn( &WaveClipListener::Invalidate ) );
+      Attachments::ForEach( std::mem_fn( &WaveClipListener::Invalidate ) );
       MarkChanged();
    }
 }
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -47,11 +47,10 @@ bool WaveClipListener::HandleXMLAttribute(
 
 WaveClip::WaveClip(size_t width,
    const SampleBlockFactoryPtr &factory,
-   sampleFormat format, int rate, int colourIndex)
+   sampleFormat format, int rate)
 {
    assert(width > 0);
    mRate = rate;
-   mColourIndex = colourIndex;
    mSequences.resize(width);
    for (auto &pSequence : mSequences)
       pSequence = std::make_unique<Sequence>(factory,
@@ -77,7 +76,6 @@ WaveClip::WaveClip(
    mTrimLeft = orig.mTrimLeft;
    mTrimRight = orig.mTrimRight;
    mRate = orig.mRate;
-   mColourIndex = orig.mColourIndex;
 
    // Deep copy of attachments
    Attachments &attachments = *this;
@@ -132,8 +130,6 @@ WaveClip::WaveClip(
       mTrimRight = orig.mTrimRight;
 
    mRate = orig.mRate;
-
-   mColourIndex = orig.mColourIndex;
 
    // Deep copy of attachments
    Attachments &attachments = *this;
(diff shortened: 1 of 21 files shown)


libraries/lib-wave-track/WaveTrack.h
```diff
@@ -1369,6 +1371,17 @@ class WAVE_TRACK_API WaveTrackFactory final
...
 };
 
+//! TrackAttachments that implement this are notified of changes in the list
+//! of clips
+class WAVE_TRACK_API ClipListListener
+{
+public:
+   virtual ~ClipListListener();
+   //! A clip was added to the list, during deserialization or after
+   virtual void OnNewClip(WaveClip &clip, bool deserializing);
```
vsverchinsky, line 1370: `deserialization` flag feels like extra parameter

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
```diff
@@ -1194,7 +1199,48 @@ WaveformAppearance::WaveformAppearance(WaveTrack &track)
 
 WaveformAppearance::~WaveformAppearance() = default;
 
+void WaveformAppearance::OnNewClip(WaveClip &clip, bool)
+{
+   WaveClipWaveformCache::Get(clip).SetColorIndex(mColorIndex);
```
vsverchinsky, line 1204: My guess is that there will be an extra call to `SetColorIndex` during deserialization: one from `OnNewClip`, and second in `HandleXMLAttribute`, right?

libraries/lib-wave-track/WaveClip.h
```diff
@@ -89,8 +87,10 @@ class WaveDisplay
    }
 };
 
-struct WAVE_TRACK_API WaveClipListener
-{
+struct WAVE_TRACK_API WaveClipListener;
+CRTP_BASE(WaveClipListenerBase, struct,
```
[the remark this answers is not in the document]
Paul-Licameli: As linked in CRTPBase.h, it's recommended here:

https://learn.microsoft.com/en-us/cpp/cpp/general-rules-and-limitations?view=msvc-170

I use the same in `extract-lib-note-track` which has passed code review but is not yet merged into master.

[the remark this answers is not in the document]
Paul-Licameli: See the bottom of that page

libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -33,6 +33,8 @@
 #include <omp.h>
 #endif
 
+template<> ClientData::Cloneable<WaveClipListener>::~Cloneable() = default;
```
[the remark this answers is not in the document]
Paul-Licameli: I see the same in the logs for the build of `extract-lib-note-track`.  I think I should instead give an inline '{}' body to the dtor in the class definition and let the linker de-duplicate it.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -1369,6 +1371,17 @@ class WAVE_TRACK_API WaveTrackFactory final
    SampleBlockFactoryPtr mpFactory;
 };
 
+//! TrackAttachments that implement this are notified of changes in the list
+//! of clips
+class WAVE_TRACK_API ClipListListener
```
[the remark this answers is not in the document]
Paul-Licameli: How would I create the observers and the subscriptions if WaveTrack is the publisher?

Each new WaveTrack must inform its own listener, which has its own separate state (the wave track color), of the event of addition of clips.

So the listener must be created too with each WaveTrack -- and that's where `ClientData` is useful, if also I use `BuildAll` instead of creation on demand.  It allows hooking of a constructor of a class.

And I need `BuildAll` so that the new attached object is present from the beginning, before any clips are created or deserialized.

But you are right that I could then also use Observer instead of this class.

There is the problem of getting deep-copy right, but that is what `Reparent` is for.

libraries/lib-wave-track/WaveTrack.h
Paul-Licameli: I'm not sure of the complete set of messages that will be useful in future.  I could easily use Observer and an enumeration now for creation, deserialization, deletion, ...

src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
Paul-Licameli: Yes, does it matter much?

It would matter in case a file serialized wave color without clip color, but then the clip colors would still be initialized.

I don't think there are any such legacy files, though.

Paul-Licameli:
Rebased to recent `audacity/wave-clip-refactoring` and resolved an easy conflict

libraries/lib-wave-track/WaveClip.h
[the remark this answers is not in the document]
crsib: I think we should revise the usage of CRTPBase in a separate PR. I have opened a task for that #5704

chinakov:
If a User has a non-default color for a Stereo track, and then:
1. Swaps Stereo Channel - the audio track reverts back to default color

chinakov:
If a User has a **non-default** color for a Stereo track, and then:
1. Swaps Stereo Channels - the audio track reverts back to default color (blue)
2. Splits Stereo Track - the last audio track reverts back to default color (blue)
3. Splts track to Mono - the last audio track reverts back to default color (blue)

This behaviour has been changed from the last release version. 

If a User has non-default colors for 2 Mono tracks, the behaviours for joining and mixing and rendering are not changed from the last release version.

chinakov:
> Recording or generating into a track uses the wave track color
Generating indeed does use the wave track color, but recording into said track still preserves the default track color (blue). 
Pictured: Recording preserve default blue color, generated audio preserves custom color defined by the User.
image

chinakov:
Another issue: if a User opens a project with differently coloured clips within one track, one of the channels will be reverted back to default color. 

How it was saved in your Branch Build:
image


How it was opened in Current Release Version:
image

Paul-Licameli:
> If a User has a **non-default** color for a Stereo track, and then:
> 
>     1. Swaps Stereo Channels - the audio track reverts back to default color (blue)
> 
>     2. Splits Stereo Track - the last audio track reverts back to default color (blue)
> 
>     3. Splts track to Mono - the last audio track reverts back to default color (blue)
> 
> 
> This behaviour has been changed from the last release version.
> 
> If a User has non-default colors for 2 Mono tracks, the behaviours for joining and mixing and rendering are not changed from the last release version.

Considering this problem, I now realize that this issue must depend first on fixing #5530

src/tracks/playabletrack/wavetrack/ui/WaveformCache.h
```diff
@@ -38,6 +36,23 @@ struct WaveClipWaveformCache final : WaveClipListener
    /** Getting high-level data for screen display */
    bool GetWaveDisplay(const WaveChannelInterval &clip,
       WaveDisplay &display, double t0, double pixelsPerSecond);
+
+   // Write color
+   void WriteXMLAttributes(XMLWriter &writer) const override;
```
crsib, line 41: Serializable cache is not a usual thing, TBH

libraries/lib-wave-track/WaveClip.h
```diff
@@ -95,6 +95,13 @@ struct WaveClipListener : WaveClipListenerBase {
    virtual ~WaveClipListener() = 0;
    virtual void MarkChanged() = 0;
    virtual void Invalidate() = 0;
+
+   // Default implementation does nothing
+   virtual void WriteXMLAttributes(XMLWriter &writer) const;
```
crsib, line 100: I think this violates SRP and in a bad way. I missed it somehow, during the first pass of the review, but it caught my eye when I saw color index serialization in the wave cache.

crsib (commented):
I am really confused by this PR, I will wait before the parent branch is merged and this PR is rebased before giving it a new round of review.

Reviewing it commit-by-commit makes it hard to understand if there is a simple way to improve it and I am seriously concerned by some of the decisions made in it. 

I cannot approve it as it is.

Paul-Licameli:
As you understood, "Cache" is now a misnomer and the next commit renames them as "Attachments" to WaveClip.  The are attached data on the clip getting certain notifications of events in the clip's lifetime.

Some attachment on WaveClip removes the irrelevant mention of color from lib-wave-track which shouldn't be there.

I did reuse existing WaveClipWaveformCache to define the override of WriteXMLAttrubutes.  If you don't like it, it's easily done to make another class just to hold the color with its own registered factory.  I didn't see that as necessary.

I don't like it that clips are now mentioning stretches and pitch shifts (also serialized data) and would put them in WaveClip attachments too.

Paul-Licameli:
Force pushed with changes that separate the attachments for colors, from the other attachments that really perform only caching.

crsib:
Looks like you forgot to push.

Paul-Licameli:
Rebased without change to updated wave-clip-refactoring branch

libraries/lib-wave-track/WaveClip.h
crsib: I have a strong opinion that the fix is worse than the problem. I understand that it is pointless to argue.

crsib (approved):
This is an incredibly complex PR addressing a minor problem. I don't like some decisions made here, but I don't want to block your branch either.

Paul-Licameli:
Rebased without changes

Paul-Licameli:
One new commit for review

Paul-Licameli:
Rebased onto updated dependency branch

crsib (approved):

Paul-Licameli:
Rather than simply rewriting SetWaveColorIndex as interval-major, better to remove completely from lib-wave-track, any mention of extraneous display related things.

Some surprisingly roundabout solutions are needed with attached objects on WaveTrack and WaveClip, but they may prove useful in future to attach other extra information to tracks and clips non-intrusively.

Issue #5603, opened 2023-11-13
Allow different stretching algorithms to be used during export


petersampsonaudacity:
TL;DR of below discussion:

* the SP stretcher correlates both sides of the stereo signal to make sure transients don't cancel out
* In cases of highly uncorrelated stereo, the signal may appear to wander from one channel to the other
* Other time stretching algorithms may not have this issue. However, SBSMS in particular is orders of magnitude too slow for realtime and thus would need to be selectable in the export window instead.

### Bug description
Clip Speed Rendering of a stretched clip destroys a stereo track's stereo image.  

This happens either if you speed render a stretched clip explicitly with the Render Clip Stretching context menu command or if you apply an effect to a stretched clip (as that will auto render prior to applying the effect).

This effectively loses a user's data as it is not readily possible to restore the stereo image later

### Steps to reproduce
1. import a known stereo track with clear, extreme, stereo imaging (I use track 1 of Sgt Pepper)
2. Time Stretch the clip shorter to 125% speed
3. Observe: Stereo image remains visually and aurally intact - **_UPDATE looks Ok visually BUT the stereo image is already corrupted/destroyed at this stage_**
4. use **Render Clip Stretching** from the clip-handle context menu (or from the Edit > Audio Clips submenu)
5. Observe: the waveforms in both channels are largely similar - the stereo image looks to be destroyed
6. Press Play
7. Observe the stereo sound stage has disappeared with the audio sounding centered _(destroying George Martin's careful work in the Abbey Road studios)_

### Expected behavior
The stereo image should not be altered by rendering a stretched clip

### Actual behavior
**Step 1**
![image](https://github.com/audacity/audacity/assets/50205984/65cff9f0-d6e5-49d2-91bd-59993d731461)

**Step 3**
![image](https://github.com/audacity/audacity/assets/50205984/59948a17-b69a-41b5-a24a-b6a9be60ec5f)

**Step 5**
![image](https://github.com/audacity/audacity/assets/50205984/7b0b2815-1e2e-4fdf-8b81-3a2fe54fd3fd)

### Audacity Version
3.4.1 and latest 3.4.2 beta (in fact all 3.4)

### Operating system
Windows 10 - but assume all OS

### Additional context
Originally discovered in #5588 @saintmatthieu  suggested that this should be an enhancement but I _**strongly**_ believe that is is a bug as is loses users' stereo data which will be hard or impossible (in some cases) to recover later.

petersampsonaudacity:
[...]

PR #5654, opened 2023-11-20
Use mid channel to find peaks but unwrap phases in left and right

Paul-Licameli:
Resolves: #5603

Recommended:


LWinterberg:
this build can invert phase polarity during processing

waveform:
image
after render:
image

Paul-Licameli:
Curious... maybe there is something "obvious"

Paul-Licameli:
Can you share the example file?

Do you observe any difference if you vary the amount of initial hidden waveform?

PR #5660, opened 2023-11-20
 fix cmake options for VST3SDK >= 3.7.8 

christian-heusel:
As these options were renamed in the meantime they caused the explictly
disabled parts of VST3SDK to be built again. We therefore adapt the build options of VST3SDK so they are disabled again.

https://github.com/steinbergmedia/vst3sdk/commit/0041ef2c879c3c54c03d33cdc11a97eaebfb5752

Resolves: -

Recommended:

This patch was already deployed to the archlinux testing repos: https://gitlab.archlinux.org/archlinux/packaging/packages/audacity/-/commit/66bf9779bd5253cfd1a91ce8dc15d3cf89ce1f83


christian-heusel:
I think the referenced commit also drops `SMTG_ADD_MYPLUGINS_SRC_PATH`, I can also check if we can drop that if wished! 😄

crsib (approved):

crsib:
Straight to merge, only affects the Arch system package.

PR #5668, opened 2023-11-22
Remove mentions of libmad

cjmayo:
libmad use was dropped by:
d4b2ac46c ("Drop outdated MAD imported", 2023-05-11)

---

Includes update to About dialog.

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes of behavior - I've used the About dialog and the man page - not the packaging scripts or Doxygen.


lib-src/audacity-patches.txt
```diff
@@ -24,28 +24,6 @@ Patches: mac-asm-fixes.patch. Fixes build on OS X Intel 10.4 by creating and
             gcc (4.0)...not need after we upgrade past Xcode 3.2.6.
 Upstream version: 1.3.1
 
-libid3tag
```
crsib, line 27: This lib is still used I believe

crsib: Hmm, you are right, sorry for confusion

crsib (approved):

PR #5669, opened 2023-11-22
Fix build with -Daudacity_use_lv2=off

cjmayo:
libraries/lib-lv2/LV2Utils.h:18:10: fatal error: lilv/lilv.h: No such file or directory
   18 | #include "lilv/lilv.h" // for lilv_free
      |          ^~~~~~~~~~~~~
compilation terminated.

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes of behavior - I've successfully built without lv2 installed, and checked that without lv2 -Daudacity_use_lv2=system still fails on checking for lv2. I hope the CI here will show it builds with lv2 installed and selected.

### DIFF
--- a/libraries/CMakeLists.txt
+++ b/libraries/CMakeLists.txt
@@ -50,7 +50,6 @@ set( LIBRARIES
    lib-vst3
    lib-snapping
    lib-vst
-   lib-lv2
    lib-ladspa
    lib-audio-unit
    lib-playable-track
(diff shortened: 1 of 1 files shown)


crsib:
> -Daudacity_use_lv2=system

Which system are you on? I think LV2 is properly found on Arch at least.

crsib (approved):

Issue #5666, opened 2023-11-22
When scrolling left to zero, beat & measures grid gets messed up


dozzzzer:
### Steps to reproduce

1. enable Beats & Measures ruler mode
2. enable "scrolling left to zero"
3. add an audio track
4. start scrolling horizontally from left to zero

### Expected behavior

The grid is consistent

### Actual behavior

The grid gets messed up:

<img width=450 src=https://github.com/audacity/audacity/assets/70063970/7632101d-cbab-4f72-9ac3-b9ef35e5de07>

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

Issue #4440, opened 2023-03-22
An issue with a stray "1" marker in negative time when using Beats and Bars view in the Timeline and "0.0" when using Minutes and Seconds view


Paul-Licameli:
There is also an issue with a stray "**1**" marker in negative time when using **Beats and Bars** view in the Timeline.

This always appears to be consistently 2.5 cm  _(on my machine)_  to the left of zero, independent of the zoom level.

![image](https://user-images.githubusercontent.com/50205984/225243073-2bd1bd2c-f720-48ab-b258-efa1714e8735.png)

_Originally posted by @petersampsonaudacity in https://github.com/audacity/audacity/issues/4412#issuecomment-1469518574_

### UPDATE
You can also observe this when using Play with Pinned playhead.

petersampsonaudacity:
You get a similar stray marker when you use the default Minutes and Seconds format for the Timeline - but with this you get a stray "0.0" marker c. 2.5 cm to the left of T=0  (and that distance seems to be independent of the zoom level.

![image](https://user-images.githubusercontent.com/50205984/227228228-a5f5c453-7515-4bb5-8aef-08d648778e46.png)

petersampsonaudacity:
This is still the case in the latest alpha:  audacity-win-3.3.0-alpha-20230329+8715a71-x64-msvc2022

petersampsonaudacity:
This remains the case in the latest 3.3.0 Beta and 3.4.0 alpha

petersampsonaudacity:
And remains the case on W10 today with the latest Beta, the one which inhibits the numbers display in negative time.

Issue #4341, opened 2023-02-21
Tracks Behaviors prefefences dialog window is not long enough


petersampsonaudacity:
### Bug description
The new preferences introduced in 3.3.0 for controlling the behaviors on copying clips to new projects overflow the size of the window for Tracks Behaviors preferences,  The window needs to be a little longer

### Steps to reproduce
1. launch 3.3.0 alpha
2. Edit > Preferences > Tracks Behaviors
3. Observe: text at the bottom of the parameters pane is trimmed and obscured

### Expected behavior
**Sufficient room to show the full set of parameters**
![image](https://user-images.githubusercontent.com/50205984/220298904-8fb3e45e-68ab-4171-93a3-374d1a72a918.png)

### Actual behavior
**Insufficient room to show the full set of parameters**
![image](https://user-images.githubusercontent.com/50205984/220300188-9cbcc5a5-340b-4948-9d75-560635ada2b7.png)

**Worse on Mac - a whole parameter is missing/obscured**
![image](https://user-images.githubusercontent.com/50205984/220304143-9055112a-eee7-48d1-a063-2e4041f5a742.png)

### Audacity Version
3.3.0 alpha:  audacity-win-3.3.0-alpha-20230220+55584ef-x64-msvc2022(2)

### Operating system
Windows 10  and macOS 12.6.3 Monterey - but assume all OS

### Additional context
See also:   
**Clip copy dialog #4033** 
**Add prompt dialog when copying clips between projects #4004**
**Performance and file size issues when copying and pasting a small portion of a very large clip #3820**

LWinterberg:
> Worse on Mac - a whole parameter is missing/obscured

But it's not missing - there is a scrollbar. People understand what scrollbars are.

petersampsonaudacity:
> > Worse on Mac - a whole parameter is missing/obscured
> 
> But it's not missing - there is a scrollbar. People understand what scrollbars are.

Some of them do, but not all.

Is there a problem with making the dialog a  little longer in height?

Issue #4185, opened 2023-01-16
Audacity erroneously plays selected audio that is in negative time


petersampsonaudacity:
### Bug description
If you have audio in negative time and then use Ctrl+A to select all then the audio starts playing from the beginning of the selected audio in negative time and not from T=0.

And note that there is no indication on the Time toolbar that the audio being played is in negative time - the counter only kicks in when the playhead finally reaches T=0 - nor is there any indication in the Selection Toolbar that a selection exists in negative time (I will log these later as separate issues).

![image](https://user-images.githubusercontent.com/50205984/212664528-a5cc4c0d-6ad7-470f-87c4-7cfb60e7a8e2.png)

This is a regression on 3.0.5, introduced in 3.1.0 

### Steps to reproduce
1. get a 3 minute song
2. use the clip-handle drag-bar to drag half of it into negative time
3. Observe: only the second half of the audio is visible (default settings don't show audio in negative time)
4. Ctrl+A to select all
5. press Play
6. Observe: audio starts playing from the beginning of the three minute song
7. Observe: Time toolbar does not operate until the audio reaches the visible part at T=0

### Expected behavior
Audio in negative time should not be played.

This was true in 3.0.5 and earlier even if the user had turned on **Enable scrolling left of zero** in **Tracks Behaviors** preferences and the audio in negative time was displayed on-screen

### Actual behavior
Selected audio in negative time is played - and with no indication in the Time toolbar

### Audacity Version
3.2.3 - but present since 3.1.0

### Operating system
Windows 10, but assume all OS

### Additional context
Originally reported by Forum user NOOBtools:
https://forum.audacityteam.org/viewtopic.php?f=46&t=128756&p=462464

petersampsonaudacity:
@LWinterberg @dozz

Testing on W10 with latest 3.5.- master alpha for 3.5.0

shows that this is _**NOT**_ fixed for 3.5.0

STRs
1. Record this sentence: "This is the beginning of the audio"
2. Time shift it so that half of the track is before zero
3. Ctrl+A to select all.
4. Press Play
5. Observe; the whole recording is played, including the before zero part

petersampsonaudacity:
See also #4187

petersampsonaudacity:
And see also

 **The Selection and Time Toolbars do not show a selection that is in negative time #4186**

LWinterberg:
[...]

PR #5675, opened 2023-11-24
Hide some lesser known preferences

LWinterberg:
Resolves: https://github.com/audacity/audacity/issues/5714
Resolves: https://github.com/audacity/audacity/issues/4341

This PR removes or hides some preferences which I expect to not be very relevant these days. 

* Removes the "no solo" button case and removes supporting code. 
* Hides "Enable dragging selection edges". I don't expect there to be a major use case where you *don't* want to be able to adjust an existing selection and only make new ones.
* Hides "Advanced Vertical Zooming". See #5685 for a new implementation
* Hides "Enable scrolling left of zero". This Resolves: https://github.com/audacity/audacity/issues/4185 Resolves: https://github.com/audacity/audacity/issues/3511 Resolves: https://github.com/audacity/audacity/issues/4440 Resolves: https://github.com/audacity/audacity/issues/5666
* Hides scrub toolbar and scrub bar. Resolves: https://github.com/audacity/audacity/issues/5683 
* removes timeline tooltip option. Resolves: https://github.com/audacity/audacity/issues/2880
* Removes leftover warnings for mixing down to mono/stereo which don't do anything anymore

### DIFF
--- a/src/menus/ViewMenus.cpp
+++ b/src/menus/ViewMenus.cpp
@@ -317,7 +317,6 @@ auto ViewMenu()
    Menu( wxT("View"), XXO("&View"),
       Section( "Basic",
          Menu( wxT("Zoom"), XXO("&Zoom"),
-            Section( "",
                Command( wxT("ZoomIn"), XXO("Zoom &In"), OnZoomIn,
                   ZoomInAvailableFlag(), wxT("Ctrl+1") ),
                Command( wxT("ZoomNormal"), XXO("Zoom &Normal"), OnZoomNormal,
@@ -328,12 +327,6 @@ auto ViewMenu()
                   TimeSelectedFlag(), wxT("Ctrl+E") ),
                Command( wxT("ZoomToggle"), XXO("Zoom &Toggle"), OnZoomToggle,
                   TracksExistFlag(), wxT("Shift+Z") )
-            ),
-            Section( "",
-               Command( wxT("AdvancedVZoom"), XXO("Advanced &Vertical Zooming"),
-                  OnAdvancedVZoom, AlwaysEnabledFlag,
-                  Options{}.CheckTest( wxT("/GUI/VerticalZooming"), false ) )
-            )
          ),
 
          Menu( wxT("TrackSize"), XXO("T&rack Size"),
--- a/src/prefs/TracksBehaviorsPrefs.cpp
+++ b/src/prefs/TracksBehaviorsPrefs.cpp
@@ -98,9 +98,6 @@ void TracksBehaviorsPrefs::PopulateOrExchange(ShuttleGui & S)
       S.TieCheckBox(XXO("Enable cut &lines"),
                     {wxT("/GUI/EnableCutLines"),
                      false});
-      S.TieCheckBox(XXO("Enable &dragging selection edges"),
-                    {wxT("/GUI/AdjustSelectionEdges"),
-                     true});
       S.TieCheckBox(XXO("Editing a clip can &move other clips"),
             EditClipsCanMove);
       S.TieCheckBox(
@@ -110,17 +107,10 @@ void TracksBehaviorsPrefs::PopulateOrExchange(ShuttleGui & S)
                     {wxT("/GUI/CircularTrackNavigation"),
                      false});
       S.TieCheckBox(XXO("&Type to create a label"),
-                    {wxT("/GUI/TypeToCreateLabel"),
+                     {wxT("/GUI/TypeToCreateLabel"),
                      false});
       S.TieCheckBox(XXO("Use dialog for the &name of a new label"),
                     {wxT("/GUI/DialogForNameNewLabel"),
-                     false});
-#ifdef EXPERIMENTAL_SCROLLING_LIMITS
-      S.TieCheckBox(XXO("Enable scrolling left of &zero"),
-                    ScrollingPreference);
-#endif
-      S.TieCheckBox(XXO("Advanced &vertical zooming"),
-                    {wxT("/GUI/VerticalZooming"),
                      false});
 
       S.AddSpace(10);
--- a/libraries/lib-playable-track/PlayableTrack.cpp
+++ b/libraries/lib-playable-track/PlayableTrack.cpp
@@ -189,9 +189,9 @@ EnumSetting<SoloBehavior> TracksBehaviorsSolo{
    wxT("/GUI/Solo"),
    {
       ByColumns,
-      { XO("Multi-track"), XO("Simple"),  XO("None") },
-      { wxT("Multi"),     wxT("Simple"), wxT("None") }
+      { XO("Multi-track"), XO("Simple")},
+      { wxT("Multi"),     wxT("Simple")}
    },
-   0, // "Simple"
-   { SoloBehaviorMulti, SoloBehaviorSimple, SoloBehaviorNone },
+   0, // "Multi-track"
+   { SoloBehaviorMulti, SoloBehaviorSimple },
 };
--- a/libraries/lib-playable-track/PlayableTrack.h
+++ b/libraries/lib-playable-track/PlayableTrack.h
@@ -68,8 +68,7 @@ ENUMERATE_TRACK_TYPE(PlayableTrack);
 
 enum SoloBehavior {
    SoloBehaviorSimple,
-   SoloBehaviorMulti,
-   SoloBehaviorNone,
+   SoloBehaviorMulti
 };
 
 extern PLAYABLE_TRACK_API EnumSetting<SoloBehavior> TracksBehaviorsSolo;
--- a/modules/mod-cl/ExportCL.cpp
+++ b/modules/mod-cl/ExportCL.cpp
@@ -462,7 +462,6 @@ public:
 
    std::unique_ptr<ExportProcessor> CreateProcessor(int format) const override;
 
-   bool CheckFileName(wxFileName &filename, int format) const override;
 };
 
 int ExportCL::GetFormatCount() const
@@ -855,21 +854,6 @@ std::vector<char> CLExportProcessor::GetMetaChunk(const Tags *tags)
 #endif
 
    return buffer;
-}
-
-bool ExportCL::CheckFileName(wxFileName &filename, int WXUNUSED(format)) const
-{
-   ExtendPath ep;
-
-   if (filename.GetExt().empty()) {
-      if (ShowWarningDialog(NULL,
-                            wxT("MissingExtension"),
-                            XO("You've specified a file name without an extension. Are you sure?"),
-                            true) == wxID_CANCEL) {
-         return false;
-      }
-   }
-   return true;
 }
 
 static ExportPluginRegistry::RegisteredPlugin sRegisteredPlugin{ "CommandLine",
--- a/src/AdornedRulerPanel.cpp
+++ b/src/AdornedRulerPanel.cpp
@@ -741,8 +741,7 @@ public:
       // May come here when recording is in progress, so hit tests are turned
       // off.
       TranslatableString tooltip;
-      if (mParent->mTimelineToolTip)
-         tooltip = XO("Timeline actions disabled during recording");
+      tooltip = XO("Timeline actions disabled during recording");
 
       static wxCursor cursor{ wxCURSOR_DEFAULT };
       return {
@@ -1328,7 +1327,6 @@ AdornedRulerPanel::AdornedRulerPanel(AudacityProject* project,
 
    mIsRecording = false;
 
-   mTimelineToolTip = !!gPrefs->Read(wxT("/QuickPlay/ToolTips"), 1L);
    mPlayRegionDragsSelection = (gPrefs->Read(wxT("/QuickPlay/DragSelection"), 0L) == 1)? true : false; 
 
 #if wxUSE_TOOLTIPS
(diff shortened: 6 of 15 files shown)


Paul-Licameli:
Coincidentally preferences are a part of the program getting my attention now as a next frontier of refactoring to libraries.

Our drafts may conflict.

petersampsonaudacity:
>Hides "Enable scrolling left of zero". 

But note that audio can still be shifted left of zero.

In fact one can shift it to be completely left of zero such that it now becomes irretrievable!

petersampsonaudacity:
I have set P1s in the alpha manual for these potential changes

SteveDaulton:
> * Hides "Advanced Vertical Zooming". If there's some use of this, it can stay until ultimately being replaced by a zoom which keeps 0 at center (or at bottom, in case of half-waves)

So will "Advanced Vertical Zooming" be "always enabled", or "always disabled"?

(I use "Advanced Vertical Zooming" very frequently, especially in spectrogram view, and would consider it a serious usability regression if it were permanently disabled.)

SteveDaulton:
> * "Type to create a label"

If I recall correctly, about 4 years ago, this feature was accidentally broken during refactoring, and there was an outcry from users that the loss of typing to create labels severely impacted their workflow. The feature is important for workflows such as transcribing.

petersampsonaudacity:
>* Hides "Enable scrolling left of zero"

You could, while doing this also make it impossible to shift audio left of zero either
a) just block it happening
b) trim the clip from the left as a smart clip as it gets "pushed past zero"

See this discussion thread:
**Why do we have negative time in Audacity - do we need it? #4188**

petersampsonaudacity:
> So will "Advanced Vertical Zooming" be "always enabled", or "always disabled"?

@SteveDaulton  testing shows that it retains the value that was set in earlier Audacity usage.

If you set it "on" in 3.4.2 and then upgrade to @LWinterberg 's branch build then int will remain "on" and available in that branch build.

petersampsonaudacity:
@LWinterberg   while you're at it Leo, in Interface Preferences

a) I think **Show Timeline tooltips** is redundant these days - it doesn't seem to do any thing.
The Manual says of it
> **Show Timeline Tooltips:** Timeline tooltips are visible when hovering over the Timeline. These tooltips are enabled by default. They are useful for indicating whether Timeline Quick-Play is enabled. Experienced users may find that the Timeline tooltips get in the way, so this option allows these tooltips to be disabled. 

But Timeline Quickplay is always on these days AFAICT - and I only see **Click and drag to define a looping region** when I hover over the Timeline - and that is regardless of the setting for this pref

-------------------------------------------------------

b) **Show Scrub Ruler:** The [Scrub Ruler](https://alphamanual.audacityteam.org/man/Scrubbing_and_Seeking#scrubruler) can be displayed or hidden using this setting.

This is a very obscure bit of Audacity functionality - and only one of three ways to access Audacity's scrubbing.  We could probably "remove" or "hide" it as part of this prefs trimming exercise.

@Paul-Licameli may disagree

petersampsonaudacity:
@LWinterberg 

And in **Warnings** preferences isn't **Missing file name extension during export** redundant now?

As I believe that in 4.4.2 you now always automatically add the appropriate extension if none is supplied?

LWinterberg:
NB, a reason this is a draft and most settings are merely hidden is for this to gauge interest and have it be easily reversible in the case that it's not . I wanted to do some more surveys about these next week, 

@petersampsonaudacity thanks for the pointers to the other things, I'll add them into this later

petersampsonaudacity:
> NB, a reason this is a draft and most settings are merely hidden is for this to gauge interest and have it be easily reversible in the case that it's not . I wanted to do some more surveys about these next week,

@LWinterberg I had guessed that to be the case  (from the "hidden")

_But note the persistence of these hidden settings from their use in earlier versions of Audacity before upgrading._

> ... the other things, I'll add them into this later

Thanks Leo.

-----------------------------------------------------------------------------------

BTW I throughly approve of

a) removing the "**None**" setting for the Solo button

b) removing **Enable dragging selection edges**  -  I can see no use case for having that turned off

With regard to **Enable scrolling left of zero** I think it is dangerous to hide/remove that while the user can still push audio into negative time.

I can see both **Advanced Vertical Zooming** and **Type to create a label** potentially being contentious.  
_Don't forget the current furore over the removal of a command/shortcut for **Export Selected Audio**._

SteveDaulton:
> But Timeline Quickplay is always on these days AFAICT - and I only see **Click and drag to define a looping region** when I hover over the Timeline - and that is regardless of the setting for this pref
> 
> _@SteveDaulton you are (or were) the Timeline king - does this make sense to you?_

Very much past tense ;-)

The "Show Timeline Tooltips" setting in Preferences is broken. (Tested on Linux with 3.4.2 AppImage)

petersampsonaudacity:
> The "Show Timeline Tooltips" setting in Preferences is broken. (Tested on Linux with 3.4.2 AppImage)

I seem to recall that it was totally changed when new looping was introduced.

Paul-Licameli:
> > * Hides "Advanced Vertical Zooming". If there's some use of this, it can stay until ultimately being replaced by a zoom which keeps 0 at center (or at bottom, in case of half-waves)
> 
> So will "Advanced Vertical Zooming" be "always enabled", or "always disabled"?
> 
> (I use "Advanced Vertical Zooming" very frequently, especially in spectrogram view, and would consider it a serious usability regression if it were permanently disabled.)

I agree with Steve that if this is removed from Preferences, I would rather default it on not off.

There is a checkbox to turn in on also in one of the View submenus, but that is less obvious.

Some very useful scrolling and zooming of the vertical rulers of waveform and spectrogram views are governed by this.

It used to be just the default until James decided otherwise.  Which I never really liked, as a user.

Paul-Licameli:
> @LWinterberg while you're at it Leo, in **Interface** Preferences
> 
> a) I think **Show Timeline tooltips** is redundant these days - it doesn't seem to do any thing. The Manual says of it
> 
> > **Show Timeline Tooltips:** Timeline tooltips are visible when hovering over the Timeline. These tooltips are enabled by default. They are useful for indicating whether Timeline Quick-Play is enabled. Experienced users may find that the Timeline tooltips get in the way, so this option allows these tooltips to be disabled.
> 
> But Timeline Quickplay is always on these days AFAICT - and I only see **Click and drag to define a looping region** when I hover over the Timeline - and that is regardless of the setting for this pref
> 
> _@SteveDaulton you are (or were) the Timeline king - does this make sense to you?_
> 
> b) **Show Scrub Ruler:** The [Scrub Ruler](https://alphamanual.audacityteam.org/man/Scrubbing_and_Seeking#scrubruler) can be displayed or hidden using this setting.
> 
> This is a very obscure bit of Audacity functionality - and only one of three ways to access Audacity's scrubbing. We could probably "remove" or "hide" it as part of this prefs trimming exercise _(note that it eats valuable screen real-estate)_.
> 
> @Paul-Licameli may disagree
> 
> I have added a P1 to the Manual in case either of both of this is "removed" or "hidden"

As for scrubbing, I’m proud of making it work in the audio engine, but I’m not attached to the UI design that never pleased everyone.

Actually Scrubbing made code dependency tangles that I regret now…

LWinterberg:
@Paul-Licameli @SteveDaulton I see a good use case for advanced vertical scrolling in spectrogram view. I do not see a good use case for it in the waveform. In what situation do you not want 0 at the center in a waveform, or at the bottom in a half-wave?

Paul-Licameli:
> @Paul-Licameli @SteveDaulton I see a good use case for advanced vertical scrolling in spectrogram view. I do not see a good use case for it in the waveform. In what situation do you not want 0 at the center in a waveform, or at the bottom in a half-wave?

If that is your opinion, which I  might agree with. and you want to realize that in 3.5, it will need more work.

LWinterberg:
As it turns out, type-to-create-label is scoped to work only when a label track is in focus. Given that's the case, I have absolutely no trouble with just having that functionality, without any setting for turning it off.

DavidBailes:
> As it turns out, type-to-create-label is scoped to work only when a label track is in focus. Given that's the case, I have absolutely no trouble with just having that functionality, without any setting for turning it off.

But for users who don't wish to make use of the type to create a label functionality, it's much more convenient for them to have an option to turn it off so that they can use all the single character and shift + single character shortcuts when a label track is the focus.

DavidBailes:
> Do these users exist?

Users of screen readers?

> but I don't see any user complaints for the reverse - where they want to use a shortcut but get annoyed that a label appears instead.

The option has been off by default since 2.3.2, over 4 years ago. The default was changed in this commit: f4aa75f4fc7b3e5c

LWinterberg:
> >Do these users exist?

> Users of screen readers?

I was unable to find complaints about this in the audacity4blind archive. Maybe freelist's search function let me down here though. The discussion in audacity-quality that lead to this change also appears mostly academical.

That said, type to create labels does feel awkward together with the "use dialogs for labels" setting and I'm tempted to tie it to that instead.

petersampsonaudacity:
> type to create labels does feel awkward together with the "use dialogs for labels" setting and I'm tempted to tie it to that instead.

@LWinterberg  -  so how do you see that working Leo?

LWinterberg:
If you use the "use dialogs" setting, you have to summon the dialog via ctrl+B. If you don't have the setting active, then type to create labels is active.

LWinterberg:
After further consideration, I'll just not touch the type to create labels preference for the time being. A "transscription workspace" may have some merit down the line.

Paul-Licameli:
> shift

Is it planned to disable the shifting of clips left of zero?

But maybe that would impair your ability to make small latency corrections.  Maybe @SteveDaulton would support that opinion?

If we simply disable the scrolling of the view left of zero, are we certain that really resolves all the issues linked above that relate to negative time?  I don't know how the change of scrolling behavior would fix a bug in playback.  But I have not yet reviewed changes.

src/menus/ViewMenus.cpp
```diff
@@ -328,12 +327,6 @@ auto ViewMenu()
                   TimeSelectedFlag(), wxT("Ctrl+E") ),
                Command( wxT("ZoomToggle"), XXO("Zoom &Toggle"), OnZoomToggle,
                   TracksExistFlag(), wxT("Shift+Z") )
-            ),
-            Section( "",
-               Command( wxT("AdvancedVZoom"), XXO("Advanced &Vertical Zooming"),
-                  OnAdvancedVZoom, AlwaysEnabledFlag,
-                  Options{}.CheckTest( wxT("/GUI/VerticalZooming"), false ) )
-            )
```
Paul-Licameli, line 336: This leaves the function `OnAdvancedVZoom` unused, so delete it too

Paul-Licameli: Did you scan the code for all occurrences of "/GUI/VerticalZooming"?

Maybe I should review #5685 first.

So far this looks like a change I could not approve, because of incompleteness.  The code is still governed by the behavior of a preference that you can't change any more in the user interface.  That behavior would be "stuck" one way or another depending on your audacity.cfg file.

By the way -- the long-distance coincidence of repeated string literals for the preference pathname  is a thing I have long disliked.  So instead there are the classes in `Prefs.h` -- `BoolSetting`, etc. -- which ought to eliminate all of these repetitions some day as a code cleanup.  There would be a lot of volume of work there.

Maybe you can figure out examples and volunteer to do some of those cleanups if you like.  That's for preferences we don't remove.

petersampsonaudacity:
> But maybe that would impair your ability to make small latency corrections

@Paul-Licameli 

In this discussion thread: **Why do we have negative time in Audacity - do we need it? #4188**

@LWinterberg wrote:
>I see two potential solutions here: 
>
>1. Remove negative time altogether, and accept that trimming now is a 2-step operation when it used to be one.
>2. Remove negative time altogether but attempting to drag a clip before 0 automatically trims the beginning (after a bit of resistance).

Of those two I would tend to favor option 2

And note that that would similarly allow for the "small latency corrections"  that Paul raises as a use-case - and in a non-destructive way **_BUT_** without the necessity for audio in negative time _(which will now be forever invisible to the user form 3.5.0 onwards)_.

src/prefs/TracksBehaviorsPrefs.cpp
```diff
@@ -110,18 +107,11 @@ void TracksBehaviorsPrefs::PopulateOrExchange(ShuttleGui & S)
...
       S.TieCheckBox(XXO("Use dialog for the &name of a new label"),
                     {wxT("/GUI/DialogForNameNewLabel"),
                      false});
-#ifdef EXPERIMENTAL_SCROLLING_LIMITS
-      S.TieCheckBox(XXO("Enable scrolling left of &zero"),
-                    ScrollingPreference);
-#endif
-      S.TieCheckBox(XXO("Advanced &vertical zooming"),
-                    {wxT("/GUI/VerticalZooming"),
-                     false});
```
Paul-Licameli, line 124: Maybe you should remove one preference at a time in each commit, and do each really completely.

"/GUI/AdjustSelectionEdges" is used in `ViewInfo::UpdatePrefs` so there is the same objection that it's possible to have behavior frozen one or another way depending on the config file.  Admittedly it's unlikely anyone would do this.  Still it leaves the code simplification incomplete.

So you should also eliminate the member variable `ViewInfo::bAdjustSelectionEdges` and simplify by assuming it was true wherever it was used before.

Paul-Licameli (changes requested):
Sufficient comments now to let you do some redrafting work.

In the interests of unblocking QA if I can, I will turn attention to your other pull requests before completing this review.

PR #5681, opened 2023-11-24
Fix bug in Noise Gate

SteveDaulton:
Typo in peak-db() causing incorrect level measurement for mono tracks.

Resolves: Incorrect level measurement when analyzing mono track noise.

*Fixes typo*

Recommended:


crsib (approved):

chinakov:
Tested via creating 2 tracks with clicks and noise, both mono and stereo, then applied the effect. The application of the effect yielded similiar results, so it seems that it's fixed. Thank you!

chinakov:
@SteveDaulton - thank you, I checked this as well, I guess I misunderstood. Tested.

PR #5685, opened 2023-11-25
Symmetric vertical zoom

LWinterberg:
Resolves: #5715

This is a companion to #5675.

It

* Always disables "advanced vertical zoom" for waveforms
* Always enables "advanced vertical zoom" for spectrograms and MIDI tracks
* uses a somewhat more appropriate cursor to drag selections in spectrograms
* uses simpler mouse gestures (click/shift-click to zoom in/out replaced with ctrl+scroll, shift+rightclick to reset replaced with simple click)
* Always zooms on the center in waveforms

It's still possible to create offset waveforms (half-wave zoom preset or shift+scroll), but zooming resets it. It'd be a nice-to-have if half-wave was a toggle, in which case zooming would keep 0 at the bottom. I don't think either of those things are blockers though.

### DIFF
--- a/src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
+++ b/src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
@@ -182,8 +182,6 @@ protected:
 // Note that these can be with or without spectrum view which
 // adds a constant.
 //const int kZoom1to1 = 1;
-//const int kZoomTimes2 = 2;
-//const int kZoomDiv2 = 3;
 //const int kZoomHalfWave = 4;
 //const int kZoomInByDrag = 5;
       kZoomIn = 6,
@@ -198,8 +196,6 @@ protected:
    void OnZoom( int iZoomCode );
 // void OnZoomFitVertical(wxCommandEvent&){ OnZoom( kZoom1to1 );};
    void OnZoomReset(wxCommandEvent&){ OnZoom( kZoomReset );};
-// void OnZoomDiv2Vertical(wxCommandEvent&){ OnZoom( kZoomDiv2 );};
-// void OnZoomTimes2Vertical(wxCommandEvent&){ OnZoom( kZoomTimes2 );};
 // void OnZoomHalfWave(wxCommandEvent&){ OnZoom( kZoomHalfWave );};
    void OnZoomInVertical(wxCommandEvent&){ OnZoom( kZoomIn );};
    void OnZoomOutVertical(wxCommandEvent&){ OnZoom( kZoomOut );};
--- a/src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
+++ b/src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
@@ -42,9 +42,7 @@ namespace
    bool IsDragZooming(int zoomStart, int zoomEnd)
    {
       const int DragThreshold = 3;// Anything over 3 pixels is a drag, else a click.
-      bool bVZoom;
-      gPrefs->Read(wxT("/GUI/VerticalZooming"), &bVZoom, false);
-      return bVZoom && (abs(zoomEnd - zoomStart) > DragThreshold);
+      return abs(zoomEnd - zoomStart) > DragThreshold;
    }
 
 }
@@ -72,17 +70,12 @@ HitTestPreview NoteTrackVZoomHandle::HitPreview(const wxMouseState &state)
       ::MakeCursor(wxCURSOR_MAGNIFIER, ZoomOutCursorXpm, 19, 15);
    static  wxCursor arrowCursor{ wxCURSOR_ARROW };
 
-   bool bVZoom;
-   gPrefs->Read(wxT("/GUI/VerticalZooming"), &bVZoom, false);
-   bVZoom &= !state.RightIsDown();
-   const auto message = bVZoom ? 
-      XO("Click to vertically zoom in. Shift-click to zoom out. Drag to specify a zoom region.") :
-      XO("Right-click for menu.");
+   const auto message = 
+      XO("Click to vertically zoom in. Shift-click to zoom out. Drag to specify a zoom region.");
 
    return {
       message,
-      bVZoom ? (state.ShiftDown() ? &*zoomOutCursor : &*zoomInCursor) : &arrowCursor
-      // , message
+      state.ShiftDown() ? &*zoomOutCursor : &*zoomInCursor
    };
 }
 
@@ -256,8 +249,7 @@ void NoteTrackVRulerMenuTable::OnZoom( int iZoomCode ){
 BEGIN_POPUP_MENU(NoteTrackVRulerMenuTable)
 
    // Accelerators only if zooming enabled.
-   bool bVZoom;
-   gPrefs->Read(wxT("/GUI/VerticalZooming"), &bVZoom, false);
+   bool bVZoom = true;
 
    BeginSection( "Zoom" );
       BeginSection( "Basic" );
@@ -320,10 +312,7 @@ UIHandle::Result NoteTrackVZoomHandle::Release
       return data.result;
    }
 
-   bool bVZoom;
-   gPrefs->Read(wxT("/GUI/VerticalZooming"), &bVZoom, false);
-   bVZoom &= event.GetId() != kCaptureLostEventId;
-   if( !bVZoom )
+   if( event.GetId() == kCaptureLostEventId )
       return RefreshAll;
 
    NoteTrackDisplayData data{ *pTrack, evt.rect };
--- a/src/tracks/playabletrack/wavetrack/ui/SpectrumVZoomHandle.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/SpectrumVZoomHandle.cpp
@@ -297,21 +297,10 @@ EndSection();
 
 
 BeginSection( "Zoom" );
-   // Accelerators only if zooming enabled.
-   bool bVZoom;
-   gPrefs->Read(wxT("/GUI/VerticalZooming"), &bVZoom, false);
-
-   AppendItem( "Reset", OnZoomResetID,         XXO("Zoom Reset"),
-              POPUP_MENU_FN( OnZoomReset ) );
-   AppendItem( "Fit", OnZoomFitVerticalID,
-      MakeLabel( XXO("Zoom to Fit"), bVZoom, XXO("Shift-Right-Click") ),
-      POPUP_MENU_FN( OnZoomFitVertical ) );
-   AppendItem( "In", OnZoomInVerticalID,
-      MakeLabel( XXO("Zoom In"), bVZoom, XXO("Left-Click/Left-Drag") ),
-      POPUP_MENU_FN( OnZoomInVertical ) );
-   AppendItem( "Out", OnZoomOutVerticalID,
-      MakeLabel( XXO("Zoom Out"), bVZoom, XXO("Shift-Left-Click") ),
-      POPUP_MENU_FN( OnZoomOutVertical ) );
+   AppendItem( "Reset", OnZoomResetID, XXO("Zoom Reset"),POPUP_MENU_FN( OnZoomReset ) );
+   AppendItem( "Fit", OnZoomFitVerticalID, XXO("Zoom to Fit"), POPUP_MENU_FN( OnZoomFitVertical ) );
+   AppendItem( "In", OnZoomInVerticalID, XXO("Zoom In"),POPUP_MENU_FN( OnZoomInVertical ) );
+   AppendItem( "Out", OnZoomOutVerticalID, XXO("Zoom Out"), POPUP_MENU_FN( OnZoomOutVertical ) );
 EndSection();
 
 END_POPUP_MENU()
--- a/src/tracks/playabletrack/wavetrack/ui/SpectrumVZoomHandle.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/SpectrumVZoomHandle.cpp
@@ -60,13 +60,13 @@ UIHandle::Result SpectrumVZoomHandle::Drag
    auto pTrack = TrackList::Get( *pProject ).Lock(mpTrack);
    if (!pTrack)
       return Cancelled;
-   return WaveChannelVZoomHandle::DoDrag(evt, pProject, mZoomStart, mZoomEnd);
+   return WaveChannelVZoomHandle::DoDrag(evt, pProject, mZoomStart, mZoomEnd, true);
 }
 
 HitTestPreview SpectrumVZoomHandle::Preview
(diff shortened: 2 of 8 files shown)


SteveDaulton:
Here we can see that there is  slight clipping of the negative going waveform in the right channel:

First Track000

SteveDaulton:
What's an appropriate level to set the Limiter threshold?

First Track001


Looking more closely at the peaks, I think I may need a bit of compression before applying the limiter.

First Track002

SteveDaulton:
Fixing the vertical scale might be useful.

First Track003

petersampsonaudacity:
I think the big plus-sign cursor for advanced vertical zooming in Spectrogram view looks most odd, cheap-and-tatty rather than carefully crafted.  Maybe it's just too big?

> Fixing the vertical scale might be useful.

In Spectrogram view it can shrinl to a very thin bar with no reticules or numbers at high V-Zoom levels:
image

petersampsonaudacity:
I did some testing on old versions - and it appears that Advanced Vertical zooming was present from at least 1.2.6 onwards, so Dominic implemented that very early on.  And in those days it was only waveforms, no Spectrogram views back then.

a) The cursor back then was the little magnifying glass with the plus sign in it (which changed to a magnifying glass with a minus sign when you you used Shift modified left click to reverse the vertical zoom.

b) you could specify the vertical range to be zoomed with the click and drag in the Vertical Scale 

c) There was no right-click context menu

d) But the vertical zooming with a left mouse button click in the vertical scale could be used many times - and it centred on where the cursor was in the Vertical Scale.  In contrast the simple non-advanced context menu zoom in on the waveform seems to be limited to just one zoom in in this branch.

And I think that @DavidBailes is absolutely right in that it was only "Advanced" vertical zooming for many years and many versions of Audacity.  And is was on;y later when James introduced the new "Simple" vertical zooming accessed from the Vertical Scale context menu that the old logstanding verical zoom acquired the label "Advanced"

### TL;DR
**I would would _strongly_ like to see Advanced vertical zooming retained for Waveforms as well as Spectrograms.rongly like to see Advanced vertical zooming retained for Waveforms as well as Spectrograms.**

LWinterberg:
>>  In contrast the simple non-advanced context menu zoom in on the waveform seems to be limited to just one zoom in in this branch
> That's a bug in my branch, I'll fix it. use ctrl+scroll in the meantime.
actually, I can't reproduce it. right-click > zoom in happily zooms in again and again for me.

SteveDaulton:
Rather that removing functionality that close to a quarter of responders to your own survey say that they use, maybe it would be better to redesign it in a way that is more discoverable and less prone to accidental clicking. For example, some audio apps use explicit zoom / pan controls on the track that operate in a similar way to scroll bars and are designed to look like zoom / pan controls.

By the way, testing https://github.com/LWinterberg/audacity/actions/runs/7011324514 on Linux, because the "Advanced Vertical Zooming" option is still available in the menu but non-function in any of the track waveform views, I predict that if Audacity is released like this, there will be a lot of people thinking that it is yet another thing that is broken.

Personally, I think that the way that "Advanced Vertical Zooming" was implemented (up to and including 3.4.2) is sub-optimal for the reasons that you describe. I also think that Advanced Vertical Zooming is a very valuable feature for advanced audio editing that must be retained. In my opinion, this pull request should be closed and an "Enhancement" issue drafted for a better implementation of current functionality.

LWinterberg:
> because the "Advanced Vertical Zooming" option is still available in the menu but non-function in any of the track waveform views, I predict that if Audacity is released like this, there will be a lot of people thinking that it is yet another thing that is broken.

https://github.com/audacity/audacity/pull/5675 is the companion PR which removes the preferences.

LWinterberg:
The zoom got unbroken now. Using ctrl+scroll and shift+scroll, you can do all the advanced zooming you want now, plus keep the benefit of it centering around 0 if you never offset the view.

SteveDaulton:
I don't need to zoom asymmetrically on the waveform very frequently, but as in the examples above, I do sometimes need to do so. Will I still be able to do that when I need to?

LWinterberg:
> I don't need to zoom asymmetrically on the waveform very frequently, but as in the examples above, I do sometimes need to do so. Will I still be able to do that when I need to?

Definitely yes. Ctrl+scroll zooms in and out, Shift+scroll moves the wave up and down.

SteveDaulton:
> Ctrl+scroll zooms in and out, Shift+scroll moves the wave up and down.

FWIW, I find that a lot more cumbersome than just click and drag, especially with the dB scale where there are three different mouse-wheel actions.

LWinterberg:
> In the Vertical Scale context menu in Spectrogram view what is the difference between Zoom to Fit and Reset Zoom ?

One goes to half the sample rate of the track, one goes to "normal" zoom (I think 20k?). There's a noticable difference between them if you go into the >48k sample rates, but for 44.1k it's very similar.

Paul-Licameli:
> > In the Vertical Scale context menu in Spectrogram view what is the difference between Zoom to Fit and Reset Zoom ?
> 
> One goes to half the sample rate of the track,

Also known as the Nyquist frequency, beyond which, any higher frequency content you might have sampled is aliased.

(Study rising square wave chirps, contrasting Square with Square, no alias.  The overtones of the perfect square wave are "bouncing off the ceiling".  But Square, no alias only sums lower terms of the Fourier series, making a wave not exactly square.)

image


image

Here you see the "Gibbs Effect" at the start of the second chirp -- a certain mathematically unavoidable overshooting that happens when you truncate the Fourier series.

image

> one goes to "normal" zoom (I think 20k?). There's a noticable difference between them if you go into the >48k sample rates, but for 44.1k it's very similar.

Paul-Licameli:
> > In the Vertical Scale context menu in Spectrogram view what is the difference between Zoom to Fit and Reset Zoom ?
> 
> One goes to half the sample rate of the track, one goes to "normal" zoom (I think 20k?). There's a noticable difference between them if you go into the >48k sample rates, but for 44.1k it's very similar.

Oh yeah -- I forgot this commit 3142060 where I changed default Spectrogram preferences according to Martin's wishes.  But for long before that, maximum frequency preference defaulted to 8000 Hz not 20000.

Zoom reset readjusts the top of the scale to whatever that preference is (but strangely, not the bottom -- which will be 0 for linear or 1 for logarithmic).

Zoom to fit goes up to Nyquist.

src/tracks/playabletrack/wavetrack/ui/SpectrumVZoomHandle.cpp
```diff
@@ -297,21 +297,10 @@ EndSection();
...
-   AppendItem( "In", OnZoomInVerticalID,
-      MakeLabel( XXO("Zoom In"), bVZoom, XXO("Left-Click/Left-Drag") ),
-      POPUP_MENU_FN( OnZoomInVertical ) );
-   AppendItem( "Out", OnZoomOutVerticalID,
-      MakeLabel( XXO("Zoom Out"), bVZoom, XXO("Shift-Left-Click") ),
-      POPUP_MENU_FN( OnZoomOutVertical ) );
+   AppendItem( "Reset", OnZoomResetID, XXO("Zoom Reset"),POPUP_MENU_FN( OnZoomReset ) );
+   AppendItem( "Fit", OnZoomFitVerticalID, XXO("Zoom to Fit"), POPUP_MENU_FN( OnZoomFitVertical ) );
+   AppendItem( "In", OnZoomInVerticalID, XXO("Zoom In"),POPUP_MENU_FN( OnZoomInVertical ) );
+   AppendItem( "Out", OnZoomOutVerticalID, XXO("Zoom Out"), POPUP_MENU_FN( OnZoomOutVertical ) );
```
Paul-Licameli, line 295: Looking at the second of five commits so far.  This change is correct for the purpose stated in the commit comment.  But I would ask, how will the mouse actions be discoverable?  The read-only Mouse preferences pages too used to serve as a reference but that is gone.  I will review the rest.

Paul-Licameli: I now say, this commit is correct, but not complete.  Find all `"/GUI/VerticalZooming"` and see that the MIDI track, too, has a vertical ruler influenced by the same preference.

src/tracks/playabletrack/wavetrack/ui/SpectrumVZoomHandle.cpp
```diff
@@ -133,8 +133,8 @@ void SpectrumVZoomHandle::DoZoom(
    const bool spectrumLinear =
       (SpectrogramSettings::Get(*pTrack).scaleType == SpectrogramSettings::stLinear);
 
-
-   bool bDragZoom = WaveChannelVZoomHandle::IsDragZooming(zoomStart, zoomEnd);
+   bool hasDragZoom = true;
```
Paul-Licameli, line 136: You might qualify this little variable as `const` or to be even more modern, `constexpr`.  That's a minor bit of style.  I skip it sometimes when the scope of a variable is small, but it can be helpful if ever you have something at the top of a large function or block.

src/tracks/playabletrack/wavetrack/ui/WaveChannelVZoomHandle.cpp
```diff
@@ -125,34 +117,25 @@ UIHandle::Result WaveChannelVZoomHandle::DoRelease(
...
-      if( bVZoom ) {
-         if( shiftDown )
-            zoomStart = zoomEnd;
-         doZoom(pProject, pTrack,
-            shiftDown
-               ? (rightUp ? kZoom1to1 : kZoomOut)
-               : kZoomIn,
-            rect, zoomStart, zoomEnd, !shiftDown);
-      }
+       bool notLost = event.GetId() != kCaptureLostEventId;
```
Paul-Licameli, line 121: Indentation.  The convention is three spaces in most of our source files.  You might change preferences in your source code editor.

Paul-Licameli (changes requested):
I'm done with this round of review.

Something must be done about the MIDI track vertical ruler too, for completeness, and to remove all uses of the "/GUI/VerticalZooming" preference.

I would also rather see this pull request also change the Preferences page to eliminate the checkbox, and also remove the item in ViewMenus (including the callback function).

Keep this separate from #5675.  Let that one address other preferences that are less complicated.

Paul-Licameli:
I'm pleased to see you start figuring out more substantial C++ changes, Leo.

About the mouse wheel actions:  there is Shift+wheel, and Ctrl+wheel (or pinch and spread on a touch pad), but did you observe that the latter can also be Shift-modified too?  That causes, for log dB view only, another change:  variation of the bottom of the dB scale.

Consider whether you like that, or want to rebind it, or mention it in tool tips.

I'm also reminded that I find the fineness of mouse wheel control for these things a bit unsatisfactory.

src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
```diff
@@ -72,17 +70,12 @@ HitTestPreview NoteTrackVZoomHandle::HitPreview(const wxMouseState &state)
...
    static  wxCursor arrowCursor{ wxCURSOR_ARROW };
 
-   bool bVZoom;
-   gPrefs->Read(wxT("/GUI/VerticalZooming"), &bVZoom, false);
-   bVZoom &= !state.RightIsDown();
-   const auto message = bVZoom ? 
-      XO("Click to vertically zoom in. Shift-click to zoom out. Drag to specify a zoom region.") :
-      XO("Right-click for menu.");
+   const auto message = 
+      XO("Click to vertically zoom in. Shift-click to zoom out. Drag to specify a zoom region.");
```
Paul-Licameli, line 74: This change is not quite equivalent to the previous version.

"Right-click for menu." was a message that appeared in the status bar, when you right click, and have not yet released the button.  This is lost.

But I suppose you observed this, and the answer is that it doesn't make sense as a discoverability aid if you really have to do that right click before you see the message suggesting right click!

And you are just assuming right click for context menu is a widespread convention we don't need to mention.

Yes?

Paul-Licameli:
> This dispels my doubt about discoverability. Rely on the tooltips.

Actually it's not tooltips, it's status bar messages, which are easier to overlook, but still it is something.

src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
```diff
@@ -256,8 +249,7 @@ void NoteTrackVRulerMenuTable::OnZoom( int iZoomCode ){
 BEGIN_POPUP_MENU(NoteTrackVRulerMenuTable)
 
    // Accelerators only if zooming enabled.
-   bool bVZoom;
-   gPrefs->Read(wxT("/GUI/VerticalZooming"), &bVZoom, false);
+   bool bVZoom = true;
```
Paul-Licameli, line 252: I think you didn't complete your intended work:  this context menu items still mention the mouse button shortcuts, but you removed those for waveform and spectrum rulers.

Paul-Licameli (changes requested):
Did you forget to remove context menu shortcuts for MIDI tracks?

src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
petersampsonaudacity: > And you are just assuming right click for context menu is a widespread convention we don't need to mention.

@Paul-Licameli  that is a _**very**_ widespread convention an many apps and OSs

petersampsonaudacity:
Testing on W10 with @LWinterberg 's latest branch:  audacity-win-3.5.0-alpha-20231130+9817f6c-x64

I am somewhat surprised ti find that while the Vertical click&drag to define a vertical zoom region - we still have it retained in the Note (MIDI) tracks.

I'm still finding the ability to zoom to a VS click&drag vertical region a loss

Yes you can achieve the same or similar efffect with Ctrl and Shift modified mouse-wheel actions - but I find those a _**lot**_ more fiddly - with more clicking _(if Gale Andrews were still alive he'd be complaining about his RSI)_.

src/tracks/playabletrack/notetrack/ui/NoteTrackVZoomHandle.cpp
LWinterberg: indeed

Paul-Licameli (approved):

SteveDaulton:
I'll rephrase:
This is very poor. You have removed _a certain kind of interaction_ that even your own survey indicates is regularly used.

torikamal:
Please bring this back to waveform view! I use this functionality every single time I edit dialog, and it has broken my flow. I can no longer zoom in to see where mouth sounds are to remove them effectively.
I apologize for being dramatic, but I actually will need to find a new DAW if this doesn't come back in. It's totally fine with me if it's an advanced, buried option. I just need the functionality which has been present for years restored.

Note that I've been an Audacity evangelist for years, and when I was teaching audio, introduced all my students to Audacity as it has simply been the fastest audio editor around.

Please restore the functionality!

Issue #5691, opened 2023-11-27
Easy channel iteration rewrites, part 1


Paul-Licameli:
An arbitrary-looking subset of the subtasks in #5065, which for various reasons are unblocked now without sequential dependencies on other tasks.

Certain of the other "easy" ones may need to wait on "hard" ones first.

This list might easily grow or shrink or be subdivided if that aids faster review and testing.

  - MoveTo
  - GetSequenceSamplesCount
  - CountBlocks
  - ConvertToSampleFormat
  - Silence
  - Flush (no outer loop, it applies only to one Interval)
  - CloseLock
  - WidestEffectiveFormat
  - RemoveCutLine
  - Resample

chinakov:
Tested, thank you Paul!

PR #5692, opened 2023-11-27
Some easy channel iteration rewrites

Paul-Licameli:
Resolves: #5691

QA: Try with mono and stereo tracks:
- [x] Dragging clips
- [x] Changing sample format on a long enough track -- observe progress indicator filling up
- [x] Unchanged size estimation in the dialog that performs cross-document copy-paste with hidden samples
- [x] Filling of progress dialog for cross-document copy-paste with hidden samples
- [x] Silence command (Ctrl+L -- not the generator)
- [x] Recording; nothing lost at the end
- [x] Save, reopen project with wave data; nothing lost
- [x] Correct non-application of dither, when you import 16 or 24 bit audio to default float, and export again.  (To test, import both files, invert one, mix, and then use Contrast to verify all is really silence)
- [x] Delete an unexpanded cutline (right click)
- [x] Tracks > Resample

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -402,6 +402,14 @@ size_t WaveTrack::Interval::CountBlocks() const
    return result;
 }
 
+void WaveTrack::Interval::ConvertToSampleFormat(sampleFormat format,
+   const std::function<void(size_t)> & progressReport)
+{
+   // TODO fix progress denominator?
+   ForEachClip([&](auto& clip) {
+      clip.ConvertToSampleFormat(format, progressReport); });
+}
+
 void WaveTrack::Interval::SetName(const wxString& name)
 {
    ForEachClip([&](auto& clip) { clip.SetName(name); });
@@ -1240,10 +1248,8 @@ void WaveTrack::ConvertToSampleFormat(sampleFormat format,
    const std::function<void(size_t)> & progressReport)
 {
    assert(IsLeader());
-   for (const auto pChannel : TrackList::Channels(this)) {
-      for (const auto& clip : pChannel->mClips)
-         clip->ConvertToSampleFormat(format, progressReport);
-   }
+   for (const auto &&pClip : Intervals())
+      pClip->ConvertToSampleFormat(format, progressReport);
    WaveTrackData::Get(*this).SetSampleFormat(format);
 }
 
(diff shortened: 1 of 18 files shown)


libraries/lib-wave-track/WaveTrack.h
```diff
@@ -1058,6 +1058,10 @@ class WAVE_TRACK_API WaveTrack final
       void ConvertToSampleFormat(sampleFormat format,
          const std::function<void(size_t)> & progressReport = {});
 
+      //! Silences the 'length' amount of samples starting from 'offset'
```
saintmatthieu, line 1054: Seems to me like `start` is actually treated as an absolute time in this method's body.

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1035,6 +1035,25 @@ size_t WaveTrack::NIntervals() const
    return mClips.size();
 }
 
+auto WaveTrack::GetWideClip(size_t iInterval) -> IntervalHolder
```
saintmatthieu, line 1062: Looks like this method is intended for calls on leaders ; if so, an assert would clarify that.

saintmatthieu (approved):
One method comment that will need to be correct, I think, and the rest looks fine. Thanks for doing this ingrateful work.

libraries/lib-wave-track/WaveTrack.h
Paul-Licameli:
```
void WaveClip::SetSilence(sampleCount offset, sampleCount length)
{
   const auto start = TimeToSamples(mTrimLeft) + offset;
   Transaction transaction{ *this };
   for (auto &pSequence : mSequences)
      pSequence->SetSilence(start, length);
   transaction.Commit();
   MarkChanged();
}
```

I don't think so?

`start` is a count of samples given to the `Sequence` object, which doesn't know the timeline.  It doesn't even know the rate.  `WaveClip` knows its rate and offset.

Paul-Licameli:
Rebased to the updated base branch, and added the suggested assertions and preconditions.

Paul-Licameli:
Fixed-up the commits to fix the Windows link; no other change

Paul-Licameli:
> One method comment that will need to be corrected, I think, and the rest looks fine. Thanks for doing this ingrateful work.

"Thankless" is the word to describe this tedious work.

But it is not that, if you are not "ungrateful" for it.

Gratitude is what our recent American holiday is about.  To God first, but to our friends.

Je vous remercie.

libraries/lib-wave-track/WaveTrack.h
saintmatthieu: Right - I was confusing `WaveTrack::Silence` and `WaveTrack::Interval::SetSilence`: the former takes an absolute value - which makes sense - and makes it relative to clip play start.

Issue #5670, opened 2023-11-23
Compressor produces corrupted audio when blank space is selected ahead of the audio


JamzTyson:
### Bug description

The output from Audacity's compressor effect may be weirdly wrong.Tested on Linux.

### Steps to reproduce

1. Download this small project: https://www.dropbox.com/scl/fi/0otz4p63lemw3ugxp5fgh/compressor-bug.aup3?rlkey=5e9f32z5xvq8jui6rp233vepc&dl=0
2. Launch Audacity 3.4.2 and open the project
3. Apply Audacity's compressor effect with default settings.

### Simple steps to reproduce
1. generate a 30 second tone starting at 15 seconds
2. silence it from 7 to 30 seconds
3. select from 5 seconds to the end of the clip
4. Effect >Compressor
5. Observe:  the result is corrupt.

**Step 3**
<img width=450 src=https://github.com/audacity/audacity/assets/50205984/fec917b9-fb44-4823-b121-2f03da08c9b2>

**Step 5**
<img width=450 src=https://github.com/audacity/audacity/assets/50205984/b3f2c338-7ce5-4993-9293-3d04f55c275c>

### Expected behavior

<img width="450" alt="Screenshot 2023-11-23 at 13 03 00" src="https://github.com/audacity/audacity/assets/70063970/bba0e6b6-f36f-46b3-b808-38613ac45025">

### Actual behavior

<img width=450 src=https://github.com/audacity/audacity/assets/151654652/54bb3b3f-6669-471a-8593-9bda4a022b74>

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Linux, Windows

petersampsonaudacity:
Testing on W10 with 3.4.2

I cannot replicate this with a very similar looking project.  What I get after applying Compressor with default settings is:

![image](https://github.com/audacity/audacity/assets/50205984/fe178546-e4a6-4762-8657-3d5095359622)

I have not tested yet with  @JamzTyson 's project - but the inference is that there ,may be something odd about that project.

petersampsonaudacity:
Testing on W10 with 3.4.2

Using @JamzTyson 's project I not that unlike mine above the audio starts at c.14 seconds and the selection includes not just the audio but also blank space from c. 2-14 seconds.
![image](https://github.com/audacity/audacity/assets/50205984/011a22c4-eeb8-4b1d-83de-1edddfe03e52)

When I apply Compressot that that selection ten I get the same  result as @JamzTyson 
[...]

Issue #3903, opened 2022-11-08
Compressor processes wrong audio selection in second clip on second track


JohnColket:
### Bug description

Forum user LeoFlash726 has reported the compressor is chopping out pieces of audio here: https://forum.audacityteam.org/viewtopic.php?p=458591#p458591 and he has supplied screeshots here: https://forum.audacityteam.org/viewtopic.php?p=458591#p458591

As the screenshots are a little tricky to follow I have tried to illustrate the issue with slightly easier steps to follow below.

### Steps to reproduce

1) Start Audacity
2) Tools > Reset Configuration
3) Transport > Recording > Record
4) Make an approximately 7-second recording of yourself reciting the counting numbers from one to as far as you get comfortably in seven seconds.
5) Track Control Panel > Split Stereo to Mono
6) "x" delete the second track.
7) Ctrl-A, Effect > Amplify and compression > Amplify the track so you can see it.
8) Tracks > Add New > Mono Track
9) On the first track, click near the one-second mark, then Edit > Audio Clips > Split (Ctrl-I)
10) Drag the second clip directly down to the second track, then roughly align the left edge of the clip with the 3-second mark.
11) Ctrl-A, Transport > Playing > Play/Stop (so you know what it sounds like: 1, silence, 2,3,4,..10 etc.).
12) Effect > Volume and Compression > Compressor > Apply
13) Ctrl-A, Transport > Playing > Play/Stop
14) Hear (1, silence, 7,8,9,10,7,8,9,10)

15) Note that the last 3 seconds of audio remains unprocessed.
16) Curiously, if you go back to step 10 and align the left edge of the clip with the 2-second mark, then the last two seconds of audio remain unprocessed.

### Expected behavior

(You should hear: 1, silence, 2,3,4,5,6,7,8,9..)

### Actual behavior

(You hear 1, silence, 7,8,9,10,7,8,9,10)

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

This seems to be a regression on 3.0.5 that somehow crept into 3.1.0 and beyond (currently 3.2.1)

This issue may be related to https://github.com/audacity/audacity/issues/1687

JohnColket:
Thank you. I ran my most recent test on the Audacity 3.2.2-beta-20221129 nightly build to confirm the problem is still preset, and yes, the problem is still present.

[...]

PR #5698, opened 2023-11-27
Compressor: fix for when there is blank space before audio

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/5670
Resolves: https://github.com/audacity/audacity/issues/3903

Problem:
When the time selection includes blank space before the audio, the results of the compressor effect are incorrect.

Relevant previous commits:
618106f1
0f830b44
The latter commit is a fix for the former. In the fix, a period of silence, mT0, is inserted into the WorkTracks, so that in ProcessOne() samples are appended starting at the same time as the start of processed audio in the original track. However, in ProcessPass(), the start of processing is given by std::max(trackStart, mT0), not mT0.

Fix:
Insert a period of silence based on std::max(trackStart, mT0). In addition, because after the addition of the silence the WorkTracks start from 0, the value of trackStart isn't calculated using WorkTracks, so avoiding unnecessary processing. trackEnd is handled similarly, just for consistency, although this isn't necessary.

Recommended:

### DIFF
--- a/src/effects/TwoPassSimpleMono.cpp
+++ b/src/effects/TwoPassSimpleMono.cpp
@@ -37,8 +37,6 @@ bool EffectTwoPassSimpleMono::Process(
    }
    for (const auto pNewTrack : mWorkTracks->Any<WaveTrack>()) {
       pNewTrack->ConvertToSampleFormat(floatSample);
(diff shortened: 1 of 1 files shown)


DavidBailes:
Needs fixing

PR #5707, opened 2023-11-28
unbundle EQ XML to TXT converter

LWinterberg:
same as https://github.com/audacity/audacity/pull/4832 - the reason this converter exists is a switch from the EQ preset format between 2.3 and 2.4. That's a few years ago now. We'll continue hosting the converter on the plugins page, but it no longer is necessary to bundle it with the executable.


crsib (approved):

petersampsonaudacity:
I have added a P1 to the Manual for this

petersampsonaudacity:
Testing on W10 with @LWinterberg 's branch build:  audacity-win-3.5.0-alpha-20231128+4b18679-x64


**EQ XML to TXT converter** is no longer in the Tools menu in this branch build:
image

Paul-Licameli (approved):

Issue #5718, opened 2023-11-28
Remove Screenshot tools


LWinterberg:

petersampsonaudacity:
Testing on W10 with @LWinterberg 's branch build for this:  

This looks to be fixed in Leo's branch.

There is no **Screenshot** entry in the Tools menu
![image](https://github.com/audacity/audacity/assets/50205984/cad7e4d2-9f4e-4a2b-874c-6e8cc576b7ce)

I have added P1s to the Manual for this for when this gets pulled into master for 3.5.0

Issue #2335, opened 2021-12-22
Screenshot tool tracking bug


LWinterberg:
- #2336
- [x] #2306 
- [ ] #1608
- [x] #1611
- [ ] #1378 
- [x] #1554

petersampsonaudacity:
> * Should the screenshot tool be deprecated? #2336

I would vote for deprecating it. 

In all the images I #captured for the manual over the years (and there were very many) I never used the screenshot tool.  Rtaher I used the Windows Print Screen button and MS Paint to perform any trims or edits.

Mac too has perfectly good screen capture tools - albeit accessed by shortcuts and not a simple "discoverable" button on the keyboard.

If the screen shot tool were to be retained it would a lot of improvement, particularly in the GUI - and for my money that would represent a poor ROI.

Penikov:
I don't think we should rid of the Screenshot tool. There is a tool called UI Vision. It could be integrated into Jenkins CI. 
We can set up Jenkins as a separate CI for automated tests. UI vision can capture the whole screen, but it works slow comprising a small image of the part of  UI. 
Better if we just point UI Vision to a folder with a set of reference images. The screenshot tool provides that. Apart from that, UI Vision works more precisely with small pictures than with full-screen images.

Yes, honestly, I do not use that tool in my daily testing. But if we talk about the coverage, developing Continuous Testing processes - that tool will be very useful.

LWinterberg:
general thinking for now is to keep the screenshot tool 2 (with the dropdown) around for QA purposes into 4.x, but to not really put in any effort into the many-buttons screenshot tool. in each case, it's not going to be touched any time soon.

PR #5719, opened 2023-11-28
No more screenshots

LWinterberg:
Resolves: #5718

Also removes some scripts relying on the screenshot tool, and VetoDialogHook which - as far as I can tell - only was useful for screenshots.


petersampsonaudacity:
Testing on W10 with @LWinterberg 's branch build for this:  

This looks to be fixed in Leo's branch.

There is no **Screenshot** entry in the Tools menu
image

I have added P1s to the Manual for this for when this gets pulled into master for 3.5.0

scripts/piped-work/get_gui_structure.py
```diff
@@ -65,7 +65,6 @@ def getStructure() :
     #do( 'Help: CommandName=Help' )
     #do( 'Help: CommandName=SetPreference' )
     #do( 'SetPreference: PrefName=GUI/Theme PrefValue=light' )
-    #do( 'Screenshot: CaptureMode=menus' )
```
Paul-Licameli, line 68: Wow, I never once in my life looked in this subdirectory of the source tree before!

Did you verify that things still run?

What remains in here can serve as examples for how to use the script pipe.

Paul-Licameli: I'm trying to run the old Perl script.  What I didn't know is that first you must start Audacity and have an audio track, not an empty project.  I get frequent error dialogs, and now the new export dialog messes up the macros by pausing for user input.

So anyway making this all run again seems out of scope for now.  But tell @dozzzzer and @chinakov to study these examples if they haven't.  Maybe writing Python programs driving the script pipe for regression tests would be easier than writing macros using Audacity's own user interface.

Paul-Licameli: I see `pipe_test.py` too, and there is still mention of screenshot in a comment in it.

src/CMakeLists.txt
```diff
@@ -891,7 +891,6 @@ list( APPEND SOURCES
       widgets/TimeFormat.h
       widgets/UnwritableLocationErrorDialog.cpp
       widgets/UnwritableLocationErrorDialog.h
-      widgets/VetoDialogHook.h
```
Paul-Licameli, line 898: You are correct that `VetoDialogHook.h` existed only to enable screenshots.  This commit looks complete to me, for the purpose of removing it.

petersampsonaudacity:
> and it was used to generate pictures in the online Audacity manual.

_**NOT**_ by me it wasn't - and after Gale Andrews died I was the major contributor and editor of the manual.

It was never that reliable a tool - I used my PC screenshot button and edited in Paint.

>But I would ask first why it's important to remove, rather than letting this harmless thing remain in Audacity 3.

Because basically it's just useless clutter that burdens the Tools menu

Paul-Licameli:
> > and it was used to generate pictures in the online Audacity manual.
> 
> _**NOT**_ by me it wasn't - and after Gale Andrews died I was the major contributor and editor of the manual _(I don't know about Steve, Bill or Ed)_
> 
> I found that it was never that reliable a tool - I used my PC screenshot button and edited in Paint.
> 
> > But I would ask first why it's important to remove, rather than letting this harmless thing remain in Audacity 3.
> 
> Because basically it's just useless clutter that burdens the Tools menu
> 
> **Big** +1 from me for its removal

In which case, I'll formally approve the PR.

Paul-Licameli (approved):
Reminder to fix just one comment in Python.

petersampsonaudacity:
It's also worth removing becase its UI is so rubbishy - it lets Audacity down.

See:
image

Many years ago there was an ENH to improve/upgrade it - but no-one ever picked that up, so it just languished ...

Paul-Licameli:
> > But I would ask first why it's important to remove, rather than letting this harmless thing remain in Audacity 3.
> 
> It has a number of bugs associated with it (see #2335), is not maintained, won't make the jump to 4, and crucially, is not used. Peter doesn't use it, QA doesn't want it, and users have easier to use options these days (Win+Shift+S on Windows, Cmd+Shift+4 on Mac). If there was some internal use for it, I'd leave the short-format dialog and the functionality intact, and just remove the button-rich dialog, but there doesn't appear to be one.

I already approved!

I’m convinced!

Thank you for yet more reasons!

While I default to conservative behaviorally neutral refactorings, and don’t make redesign my job, this thing has also been a mildly annoying maintenance burden sometimes.  Bye bye.

Issue #5742, opened 2023-12-05
Rewrite WaveTrack::Clear, ClearAndAddCutline, SplitDelete


Paul-Licameli:
Interval-major rewrites of the functions named in this issue.

QA: test for mono and stereo tracks:
- [x] Making cutlines, expanding and deleting them
- [x] Change Speed, Change Tempo, Sliding stretch, where the selected time includes gaps between clips (high and low quality)
- [x] Edit > Remove Special > Split Delete (Ctrl + Alt + K)
- [x] Edit > Remove Special > Split Cut (Ctrl + Alt + X)
- [x] Edit > Remove Special > Trim (Ctrl + T)
- [x] Edit > Clips > Split New (Ctrl + Alt + I)
- [x] Edit > Clips > Disjoin (Ctrl + Alt + J)
- [x] Apply Equalization, Truncate Silence, Noise Reduction
- [x] Generating inserted Silence of zero duration
- [x] Tracks > Stereo to Mono
- [x] Shortening of a sync-locked wave track
- [x] Include time selections that completely surround a clip
- [x] Include time selections that overlap a clip partly at the left or right end, or are inside a clip; both for split-delete and for simple deletion
- [x] With "Always paste audio as new clips" off, paste into a clip, verify no splits are made 
- [x] With "Always paste audio as new clips" on, paste into a clip, verify splits are made, and also if you delete the middle clip, the left and right clips retain hidden data
- [x] Repeat the last, in the case of two clips with a gap between them, and the selection is from inside one clip to inside the other, surrounding the gap

dozzzzer:
@Paul-Licameli 
- with "Always paste audio as new clips" ON, attempting to paste audio from one stereo track to another always causes a crash, regardless of the "Editing a clip can move other clips" setting:

https://github.com/audacity/audacity/assets/70063970/97183f40-4a67-4172-bc22-5b3f2cdc56a1

dozzzzer:
@Paul-Licameli an unintended change of behavior is introduced by this PR:
- with "Always paste audio as new clips" ON and "Editing clips can move other clips" OFF, copy a selection from one audio track and paste it to the other. Currently, the selection is pasted as a separate clip. Previously, it would yield "not enough room" error message.

#### Current master

https://github.com/audacity/audacity/assets/70063970/172f5e6f-2db2-4665-a585-5cbc573c6625


#### This PR


[...]

PR #5748, opened 2023-12-06
Rewrite clear and others

Paul-Licameli:
Resolves: #5742 

Depends on
- #5763

Interval-major rewrites of functions using WaveTrack::HandleClear, other than ClearAndPaste

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -2270,14 +2270,6 @@ void WaveTrack::HandleClear(
    // Only now, change the contents of this track
    // use No-fail-guarantee for the rest
 
-   const auto moveClipsLeft = !split && GetEditClipsCanMove();
-   if (moveClipsLeft)
-      // Clip is "behind" the region -- offset it unless we're splitting
-      // or we're using the "don't move other clips" mode
-      for (const auto& clip : mClips)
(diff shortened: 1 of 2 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -272,14 +272,14 @@ void WaveTrack::Interval::SetTrimRight(double t)
       GetClip(channel)->SetTrimRight(t);
 }
 
-void WaveClip::Interval::TrimLeft(double deltaTime)
+void WaveTrack::Interval::TrimLeft(double deltaTime)
```
saintmatthieu, line 275: ... oh, there we go. You might want to squash these with the previous commit, which probably won't compile.

saintmatthieu (approved):
Clean-up looks good and sequence of commit made the review almost entertaining :D

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: Yes

Paul-Licameli:
Rebased unchanged just to clean up the commit history

saintmatthieu (approved):

Paul-Licameli:
Rebased without change onto the updated base branch

Issue #5757, opened 2023-12-07
Music import dialog: When it opens, it's not automatically read by some screen readers.


DavidBailes:
### Bug description

At least on my computer, this bug occurs when using Narrator or NVDA, but not when using Jaws.

### Steps to reproduce

On Windows, open either the Narrator or NVDA screen reader.
1. Open audacity.
2. Import a track with xxxbpm in the filename. A music import dialog opens.

### Expected behavior

The screen reader automatically reads the music import dialog.

### Actual behavior

The screen reader reads the name of the track - the screen reader thinks the track is the focus.
Here's a screen shot when using the narrator screen reader - the blue outline around the track shows where the screen reader thinks the focus is:
![image](https://github.com/audacity/audacity/assets/11887021/a7034671-3635-47f1-aaf3-1a80ad10fe4b)

After the dialog opens, if you then press any standard navigation keystrokes then when a new control in the dialog becomes the focus, the screen reader then reads that control correctly.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

DavidBailes:
@LWinterberg . I've submitted a PR for this, please can this issue be added to the 3.5 milestone.

saintmatthieu:
@DavidBailes, @ianalarina assigned me, but I think it makes more sense for you to be the assignee, what do you think?

DavidBailes:
> @DavidBailes, @ianalarina assigned me, but I think it makes more sense for you to be the assignee, what do you think?

I've already submitted a PR to fix this.

dozzzzer:
Tested

PR #5770, opened 2023-12-11
Music import dialog: screen reader problem

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/5757

When the Music import dialog opens, it's not automatically read by NVDA or Narrator, but it is by Jaws. Both NVDA and Narrator still think that the focus is the imported track.

Problem:
In TrackPanelAx::GetState(), the state of the list or a list item is given as focussed without checking whether it is actually focussed. Looks like this has always been the case, but hasn't been a problem up to now.

Fix:
Check whether focused: For the list item, explicitly check, and the for list, let wxWidgets do this using a standard accessible object.

QA:
This fix needs to be checked with NVDA and with Narrator.

Recommended:

### DIFF
--- a/src/TrackPanelAx.cpp
+++ b/src/TrackPanelAx.cpp
@@ -453,7 +453,7 @@ wxAccStatus TrackPanelAx::GetState( int childId, long* state )
       *state = wxACC_STATE_SYSTEM_FOCUSABLE | wxACC_STATE_SYSTEM_SELECTABLE;
       if (t)
       {
-         if (t == pFocus->PeekFocus())
+         if (t == pFocus->PeekFocus() && GetWindow() == wxWindow::FindFocus())
          {
             *state |= wxACC_STATE_SYSTEM_FOCUSED;
          }
@@ -466,7 +466,8 @@ wxAccStatus TrackPanelAx::GetState( int childId, long* state )
    }
    else     // childId == wxACC_SELF
    {
-      *state = wxACC_STATE_SYSTEM_FOCUSABLE + wxACC_STATE_SYSTEM_FOCUSED;
+      // let wxWidgets use a standard accessible object for the state
+      return wxACC_NOT_IMPLEMENTED;
    }
 #endif
 


src/TrackPanelAx.cpp
```diff
@@ -466,7 +466,8 @@ wxAccStatus TrackPanelAx::GetState( int childId, long* state )
    }
    else     // childId == wxACC_SELF
    {
-      *state = wxACC_STATE_SYSTEM_FOCUSABLE + wxACC_STATE_SYSTEM_FOCUSED;
+      // let wxWidgets use a standard accessible object for the state
+      return wxACC_NOT_IMPLEMENTED;
```
saintmatthieu, line 470: I'm not familiar with that code, and also I cannot hit a breakpoint here that'd give me more context. Would I need a narrator for that, I suppose? @DavidBailes if you're okay with it let's just proceed with this one to QA, provided you've added a QA checklist in the description.

DavidBailes: Yes, TrackPanelAx::GetState() only gets called if a screen reader or one or the accessibility testing tools are running. The relevant code for how wxACC_NOT_IMPLEMENTED is handled is pasted below.
Note sure a QA checklist is needed - it's pretty obvious whether on opening, the dialog is automatically read by Narrator and NVDA.
I'm happy for it to go straight to QA.

```
STDMETHODIMP wxIAccessible::get_accState ( VARIANT varID, VARIANT* pVarState)
{
    wxLogTrace(wxT("access"), wxT("get_accState"));
    wxASSERT( ( m_pAccessible != NULL ) || ( m_bQuiescing == true ) );
    if (!m_pAccessible)
        return E_FAIL;

    if (varID.vt != VT_I4 || varID.lVal < 0)
    {
        wxLogTrace(wxT("access"), wxT("Invalid arg for get_accState"));
        return E_INVALIDARG;
    }

    long wxstate = 0;

    wxAccStatus status = m_pAccessible->GetState(varID.lVal, & wxstate);
    if (status == wxACC_FAIL)
        return E_FAIL;
    if (status == wxACC_INVALID_ARG)
        return E_INVALIDARG;

    if (status == wxACC_NOT_IMPLEMENTED)
    {
        // Try to use child object directly.
        if (varID.lVal > 0)
        {
            IAccessible* childAccessible = GetChildAccessible(varID.lVal);
            if (childAccessible)
            {
                varID.lVal = 0;
                HRESULT hResult = childAccessible->get_accState(varID, pVarState);
                childAccessible->Release();
                return hResult;
            }
            else if (m_pAccessible->GetIAccessibleStd())
                return m_pAccessible->GetIAccessibleStd()->get_accState(varID, pVarState);
        }
        else if (m_pAccessible->GetIAccessibleStd())
            return m_pAccessible->GetIAccessibleStd()->get_accState (varID, pVarState);
    }
    else
    {
        LONG state = wxConvertToWindowsState(wxstate);
        pVarState->lVal = state;
        pVarState->vt = VT_I4;
        return S_OK;
    }
    return E_NOTIMPL;
}
```

saintmatthieu: Ok.

saintmatthieu (approved):

src/TrackPanelAx.cpp
DavidBailes: > provided you've added a QA checklist in the description

I've added a note for QA to test with both Narrator and NVDA.

Issue #5772, opened 2023-12-11
Remove distinction between `GetTrack` and `ReallyGetTrack` in `WaveChannel`


Paul-Licameli:
Depends on
- #5806

QA: Mono and stereo...
- [x] Recording
- [x] Mix-and-render
- [x] Import (any file format)
- [x] A generator (any)
- [x] A destructive effect (any)
- [x] Run benchmark tool
- [x] Mix Stereo to Mono
- [x] Low quality change pitch, change tempo effects
- [x] Correct refresh of waveform and spectrum display after track resampling, Tracks > Align commands, dragging clips from track to track
- [x] Dark blue and light blue waveform areas are computed the same
- [x] Save project in 3.3 or earlier with a stereo track with misaligned clips and non-default sample rate; open it in this branch and verify the correct split and correct rates in the tracks and clips
- [x] Loudness effect on stereo tracks really processes both channels
- [x] Cut, copy, paste, select all (Ctrl+A) of text in focused clip name

crsib:
There is one case in PR that should be fixed, but I don't expect that the fix would change the behavior of the program. I would say it is save to start testing.

Paul-Licameli:
> @Paul-Licameli
> 
>     * Generating onto a stereo track using any Nyquist generator causes a crash

That bug already exists in the refactoring branch.  It escaped our detection earlier.

But I have figured out the fix and will add it to this branch.

Paul-Licameli:
In fact that crash bisects to commit 09ca4cdec66eded3dadd297dd0449f218f08a397 in #5692

Paul-Licameli:
I made this issue now dependent on another new issue, to fix the preexisting crash bug in Nyquist generators.

dozzzzer:
Tested

PR #5773, opened 2023-12-11
Really get track

Paul-Licameli:
Resolves: #5772

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -3195,7 +3195,7 @@ bool WaveChannel::AppendBuffer(constSamplePtr buffer, sampleFormat format,
    size_t len, unsigned stride, sampleFormat effectiveFormat)
 {
    const size_t iChannel = ReallyGetChannelIndex();
-   return ReallyGetTrack()
+   return GetTrack()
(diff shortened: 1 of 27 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -3439,13 +3472,12 @@ std::pair<float, float> WaveChannel::GetMinMax(
    if (t0 == t1)
       return results;
 
-   for (const auto &clip: GetTrack().mClips)
+   for (const auto &&clip: Intervals())
```
crsib, line 3475: Constant rvalue references are highly unusual. Why they are needed here?

Paul-Licameli: I will strike const

crsib: Why the rvalue reference is needed here?

Paul-Licameli: Const was a mistake.

&& isn’t necessary but it is a reminder that the interval iterator dereferences to a value, not a reference as in the old version.

Interval iterators dereference to a shared pointer to a temporary structure.

Paul-Licameli:
Rebased onto the new dependency, and also fixed that rvalue references in loops.

PR #5779, opened 2023-12-12
Attack of the fifty foot jellyfish!!!

Paul-Licameli:
Oh no!  That four-year-old laboratory experiment that was long left for dead has REVIVED!

(Especially for you at your request to see it again, @petersampsonaudacity)

Recommended:


Paul-Licameli:
Updated -- I did some important work on the problem of per-track metering.

Try it out again, with multiple, different stereo tracks, and observe that the goniometers display the correct per-track output, not the master output.

Also vary the gain and pan sliders during play and observe how the meters respond.

The work on internals to make this happen is serious work that should be merged to master, even if the display code is not.

Issue #3781, opened 2022-10-07
The user is asked to reset preferences even if they selected "Reset Preferences" in the install wizard


dozzzzer:
### Steps to reproduce

1. Download the [3.2.1 installer](https://github.com/audacity/audacity/releases/tag/Audacity-3.2.1)
2. Run the installer
3. Check the checkbox "Reset Preferences"
4. Finish the installation and launch Audacity
5. Observe

### Expected behavior

No **Reset Preferences** confirmation dialog

### Actual behavior

<img width="670" alt="image" src="https://user-images.githubusercontent.com/70063970/194604223-4fd0e6fc-f18c-476c-99fe-0509d8293299.png">

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

dozzzzer:
Apparently, this confirmation dialog was added intentionally, though it seems redundant to me. 
WDYT, @Tantacrul ?

Tantacrul:
I HATE this dialog. It utterly confused me the other day, when I came into contact with it for the first time. It is also really confusing.

I initially thought it was asking 'have you reset your preferences already?'

Agree with @dozzzzer!

petersampsonaudacity:
I'm minded to agree with @dozzzzer amd @Tantacrul 

But I would probably opt for removing the option in the installation process - and retaining the offer on initial launch.

dozzzzer:
Still the case in 3.4.2

dozzzzer:
Tested

PR #5781, opened 2023-12-13
remove dialog when resetting prefs via installer

LWinterberg:
Resolves: https://github.com/audacity/audacity/issues/3781

This probably isn't testable with the PR build, but should be in the next beta. Unless @crsib has an idea on how to make it happen earlier.

### DIFF
--- a/src/AudacityApp.cpp
+++ b/src/AudacityApp.cpp
@@ -247,18 +247,8 @@ void PopulatePreferences()
    // User requested that the preferences be completely reset
    if (resetPrefs)
    {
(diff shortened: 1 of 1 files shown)


crsib:
You can run the job manually with the beta/release type

LWinterberg:
aha! In that case, https://github.com/LWinterberg/audacity/actions/runs/7194667312 would be the build to test this with (I think - I'll try when it finishes)

Paul-Licameli (approved):

PR #5800, opened 2023-12-19
Use custom selectionleft/right cursors

LWinterberg:
Resolves: #5801


src/tracks/ui/SelectHandle.cpp
```diff
@@ -316,11 +318,11 @@ namespace
          break;
       case SBLeft:
          tip = XO("Click and drag to move left selection boundary.");
-         pCursor = &adjustLeftSelectionCursor;
+         pCursor = &*adjustLeftSelectionCursor;
```
Paul-Licameli, line 321: Alternatively `pCursor = adjustLeftSelectionCursor.get()`

`adjustRightSelectionCursor` now has type `std::unique_ptr<wxCursor>`

DilsonsPickles (approved):
Looks good to me, goodbye white hands!

Issue #5580, opened 2023-11-11
Timer Record automatic export defaults to mono when project is set to record in stereo


petersampsonaudacity:
### Bug description
Timer Record automatic export defaults to mono when project even if the user has their project set to record in stereo, which is the Audacity default setting.

This is a regression on 3.3.3 introduced in 3.4.0

### Steps to reproduce
1. in Audio Setup ensure Audacity is set to stereo
2. Transport > Recording > Timer record
3. click in the **Enable Automatic Export** box
4. click on the Automatic Export **Select** button
5. Observe:  the export is set to default to mono

### Expected behavior
When the project is set to 2-channel stereo then the default Automatic Export in Timer Record should be stereo

When the project is set to 1-channel mono then the default Automatic Export in Timer Record should be mono

### Actual behavior
![image](https://github.com/audacity/audacity/assets/50205984/3d8af594-6bf8-461e-b7c4-a0e03b81234b)

### Audacity Version
3.4.1 and 3.4.2 latest Beta

### Operating system
Windows 10 - but assume all OS

### Additional context
Originally reported on the Forum:  
https://forum.audacityteam.org/t/save-to-mp3-defaults-monoral/89484

petersampsonaudacity:
Fortunately the workaround is simple.

_**BUT**_ it does require the user to be observant and take action ELSE they may lose valuable stereo data if they mistakenly export the file as mono and then delete the project.

fory-horio:
FYI, I just exported to mp3 this morning thinking that it was "stereo", but was not. I had to redo all of the work, wasted all most 2 hours.

petersampsonaudacity:
@LWinterberg  @dozzzzer 

note that this is particularly egregious if you just accept the default folder offered in the automatic export dilalog and just press **OK** to effect the setting up of the export.
![image](https://github.com/audacity/audacity/assets/50205984/7ca318d6-d5e6-438b-881a-5ffa484db33d)


It is **ONLY** if you press the **Select** button that you get the dialog showing that the export will be mono
![image](https://github.com/audacity/audacity/assets/50205984/7757b7ce-6a7d-4904-818d-0f60b82299c1)

_On that basis I believe that this deserves a higher rating than P4_

DavidBailes:
@petersampsonaudacity . I've submitted a fix for this: https://github.com/audacity/audacity/pull/5802

petersampsonaudacity:
[...]

Issue #4952, opened 2023-07-25
Timer Record can fail to Export


petersampsonaudacity:
### Bug description

Using defaults in Timer Record and choosing Automatic Export
![image](https://github.com/audacity/audacity/assets/50205984/888b9ed8-9ca0-475c-a6ce-02d2e505ffb2)

 The Export on completion fails:
![image](https://github.com/audacity/audacity/assets/50205984/ed5bd996-0490-47ae-84d2-86566981efc1)

_I assume that it is because although the folder location is set there is no default filename set (see STRs below)_

### Steps to reproduce

1. clear Audacity settings folder
2. Launch Audacity (Vitaly's branch build for new Export)
3. Transport > Recording > Timer Record
4. set the timer for 5 seconds recording (immediate)
5. Check "on"  the **Automatic Export**
6. Observe:  A default location is now provided for the export   _(not the case in earlier Audacity versions)_
7. Press OK in the Timer Record dialog
8. Observe: 5 seconds recorded
9. Observe: The Export fails (see above)

If instead at step 5 you 
5.1 press the Select button by the Automatic Export
5.2 Observe: folder location and filename untitled.wav are populated
5.3 make no changes just press the OK button
Then the Export (of untitled.wav) proceeds correctly

### Expected behavior
Either the TR should not be allowed to be activated if the file location is not set

or

a default filename of untitled .wav should be used

### Actual behavior
Export fails

### Audacity Version
@vsverchinsky 's latest brabch build for new Export

### Operating system
Windows 10 - but assume all OS

### Additional context
None

petersampsonaudacity:
This is a regression on 3.3.3 as with that version (and earlier) if you leave the Export filed unmodified the Timed Recording is blocked when you press the OK button in the TR dialog:
![image](https://github.com/audacity/audacity/assets/50205984/1ee3f2e9-8765-4478-97a4-c09a46451e2f)

petersampsonaudacity:
@LWinterberg @dozzzzer @crsib 

Testing on W10 with @DavidBailes ' branch:  audacity-win-3.5.0-alpha-20231219+2e17667-x64

As Suggested by David in PR #5911  This issue also appears to be fixed in David's branch

dozzzzer:
Tested

PR #5802, opened 2023-12-19
Timer Record: incorrect default number of channels.

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/5580
May also at least help with: https://github.com/audacity/audacity/issues/4952

Timer Record automatic export defaults to mono when project is set to record in stereo

Problem:
In the TimerRecordDialog class, m_iAutoExportChannels is initialized to 0; In the TimerRecordDialog, if you enable automatic export, but don't press the Select button to open the Export Audio dialog, then this leads to the export failing. If you do press the Select button and open the Export Audio dialog, then initially channels is set to mono, even if you're going to record in stereo. (There's a a call to TimerRecordExportDialog::Bind with number of channels equal to 0, then a call to ExportFilePanel::Init, with number of channels 0. When this parameter is zero this Init function gets the correct number of channels from the tracks to be exported, but in this context, there are as yet no tracks.)

Fix:
Set m_iAutoExportChannels to the number of recording channels.

Recommended:

### DIFF
--- a/src/TimerRecordDialog.cpp
+++ b/src/TimerRecordDialog.cpp
@@ -214,6 +214,7 @@ TimerRecordDialog::TimerRecordDialog(
    }
 
    m_iAutoExportSampleRate = ProjectRate::Get(mProject).GetRate();
+   m_iAutoExportChannels = AudioIORecordChannels.Read();
 
    ShuttleGui S(this, eIsCreating);
    this->PopulateOrExchange(S);


petersampsonaudacity:
Testing on W10 with @DavidBailes ' branch for this:  audacity-win-3.5.0-alpha-20231219+2e17667-x64

This now works properly exporting with the correct number of channels the user has set to record to before setting up th Timer Recording.

Works properly either if the user examines the deeper dialog:
image

Or just uses the **OK** button without drilling down with the **Select**
image

petersampsonaudacity:
Testing on W10 with @DavidBailes ' branch for this:  audacity-win-3.5.0-alpha-20231219+2e17667-x64

This now works properly in David's branch, exporting with the correct number of channels the user has set to record to before setting up the Timer Recording.

Works properly either if the user examines the deeper dialog:
image

Or just uses the **OK** button without drilling down with the **Select**
image

Thanks for that Fix @DavidBailes

petersampsonaudacity:
@LWinterberg @dozzzzer @crsib 

Testing on W10 with @DavidBailes ' branch for this:  audacity-win-3.5.0-alpha-20231219+2e17667-x64

This now works properly in David's branch, exporting with the correct number of channels the user has set to record to before setting up the Timer Recording.

Works properly either if the user examines the deeper dialog:
image

Or just uses the **OK** button without drilling down with the **Select**
image

Thanks for that Fix @DavidBailes

petersampsonaudacity:
As David suggests the PR also appears to properly fix #4952

crsib (approved):

petersampsonaudacity:
@crsib Dmitry, now that you have approved this, any chance we can get this pulled into master please?

Issue #5637, opened 2023-11-17
Update/changelog window may exceed screen bounds and then be impossible to dismiss


jamshark70:
### Bug description

Note in the screenshot that the bottom of the updates window (where the buttons are) is offscreen.

![audacity-bad-update-window](https://github.com/audacity/audacity/assets/318301/8a1460f7-f8fe-4cb7-8e4e-e16d7530332e)

At this point, tab-selection of buttons seems inoperative (I tried tab-space, tab-space a few times, and was unsuccessful in dismissing the window). Also the changelog window does not allow resizing (?). Had I been able to resize it, then I wouldn't have had to log an issue -- just resize, "OK" and done. Buuuuut... no dice.

I had no choice but to force-kill Audacity and use other software in this class session.

### Steps to reproduce

Launch the software when updates are pending. (However the problem doesn't reproduce unless the changelog is rather long, and your screen resolution settings aren't very deep, e.g. 900 pixels high.)

### Expected behavior

Windows opened by the software should always be navigable: completely visible onscreen, or if too large, at least possible to resize and bring onscreen.

### Actual behavior

Un-dismissable and un-resizable modal window covering Audacity's workspace, rendering the software unusable until updating.

### Audacity Version

other

### Operating system

Linux

### Additional context

Audacity 3.3.3. I *will* update soon but didn't have time to do that when in the middle of a classroom demo.

dozzzzer:
I was able to reproduce the issue on Fedora 39.

PR #5811, opened 2023-12-21
limit update dialog size

LWinterberg:
Resolves: #5637

same approach as https://github.com/audacity/audacity/pull/3609

### DIFF
--- a/src/update/UpdatePopupDialog.cpp
+++ b/src/update/UpdatePopupDialog.cpp
@@ -16,6 +16,7 @@
 #include <wx/debug.h>
 #include <wx/sstream.h>
 #include <wx/txtstrm.h>
+#include <wx/display.h>
 
 enum { DontShowID = wxID_HIGHEST + 1 };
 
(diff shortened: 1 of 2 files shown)


src/update/UpdatePopupDialog.cpp
```diff
@@ -64,6 +65,10 @@ UpdatePopupDialog::UpdatePopupDialog (wxWindow* parent, const VersionPatch& vers
     Layout();
     Fit();
     Center();
+
+    wxSize sz = GetSize();
+    SetSizeHints(sz.x, std::min(sz.y, 600),
+        wxDefaultCoord, wxDisplay().GetGeometry().GetHeight() - 100);
```
Paul-Licameli, line 71: Maybe instead
```
const auto maxHeight = wxDisplay().GetGeometry().GetHeight() - 100;
const auto minHeight = std::min({ sz.y, 600, maxHeight });
SetSizeHints(sz.x, minHeight, wxDefaultCoord, maxHeight);
```

just to be sure your minimum height is not more than the maximum

vsverchinsky (approved):

PR #5812, opened 2023-12-21
Don't split source stereo track, dragging clip into another track...

Paul-Licameli:
... the bug occurred only for tracks deserialized from a project file, and only the first time (not again, after undo and repetition of the drag)

Resolves: #5389 

Recommended:


Paul-Licameli:
Should this fix be considered for another 3.4 patch?

crsib:
Let's discuss this at triaging ( @dozzzzer )

crsib (approved):

PR #5818, opened 2023-12-26
`lib-wave-track-paint`

vsverchinsky:
Extracts platform-indepentend parts of a waveform image generation implemented in Dmitry's branch into a new `lib-wave-track-paint`, but it not yet used by-default. Insead one will have to enable it by uncommenting `EXPERIMENTAL_WAVEFORM_PAINT` flag in Experimental.cmake. 
The major difference is that this branch uses `wxDC` to paint the generated waveform image, thus changes in the existing code are minimal and performance benefits are still noticeable.
Before:
before
After:
изображение

Recommended:


crsib:
Have you addressed the issue with zoom levels where min and max have the same sign?

vsverchinsky:
> Have you addressed the issue with zoom levels where min and max have the same sign?

I did an attempt to fix that but not sure that it was correct:
изображение
изображение

Paul-Licameli:
Linux build is failing.

Is this only about waveform painting? If so, rename it all as Lin-waveform-paint.

Or maybe lib-waveform-geometry if the library will also handle any hit testing.  The painting and hit testing for anything painted on the screen should be kept together.  Painting maps some thing to pixel x and y coordinates, while hit testing must know the inverse of the same mapping.

crsib:
> The painting and hit testing for anything painted on the screen should be kept together

I strongly disagree. These topics are irrelevant most of the time, and definitely irrelevant here.

crsib:
> If so, rename it all as Lin-waveform-paint.

`lib-wave-track-paint` is a green library providing a way to build visuals for wave tracks. Right now, only waveform painting has been migrated to this library. Neither Vitaly nor I ever migrated the "green" part of the spectrogram painting. This can be done, however.

vsverchinsky:
> Linux build is failing.

Trying to fix that...

> Is this only about waveform painting? If so, rename it all as Lin-waveform-paint.

Right now it has only waveform image generation part. As you mentioned it could provide geomentry data that will be useful for hit testing and in that case `lib-waveform-geometry` could be an option too.

> Or maybe lib-waveform-geometry if the library will also handle any hit testing

It could indeed provide geometry information for hit testing purpose, as hit testing itself usually refers to platform or framework specific code, I think.

crsib:
> It could indeed provide geometry information for hit testing purpose, as hit testing itself usually refers to platform or framework specific code, I think.

I strongly disagree. This is a lower-level library, that is only responsible for helping the WaveformView to **paint** given the ZoomInfo, selection, and screen bounds. It is not responsible for interactions and it should be kept that way.

Paul-Licameli:
> > It could indeed provide geometry information for hit testing purpose, as hit testing itself usually refers to platform or framework specific code, I think.
> 
> I strongly disagree. This is a lower-level library, that is only responsible for helping the WaveformView to **paint** given the ZoomInfo, selection, and screen bounds. It is not responsible for interactions and it should be kept that way.

No, don't make it responsible completely for interactions.  However the association between pixel coordinates and things in the data model ought to be defined in one place.  A mapping of things to pixels required for painting, and its inverse mapping, should be defined together and kept consistent.

So maybe what I want to argue is that lib-waveform-geometry would be separated from a painting library and do only such coordinate transformations.  lib-waveform-painting could depend on it, and hit testing would also depend on it, but neither of those two would depend on the other.

My opinion about spectrogram is that it is a completely different association of things with pixels, at least in the y coordinate, and would have its own separate libraries.

crsib:
> No, don't make it responsible completely for interactions. However the association between pixel coordinates and things in the data model ought to be defined in one place

ZoomInfo is still used to achieve this purpose.

> My opinion about spectrogram is that it is a completely different association of things with pixels, at least in the y coordinate, and would have its own separate libraries.

This library provides the way to cache imagery, required to draw WaveTrack. Spectrograms are no different in this aspect. The way how the specific bitmap is generated can be lifted away from the library, resulting in 3 libraries: caching, waveform bitmap generation, and spectrogram image generation.

Paul-Licameli:
> > No, don't make it responsible completely for interactions. However the association between pixel coordinates and things in the data model ought to be defined in one place
> 
> ZoomInfo is still used to achieve this purpose.

Yes, the association x-coordinate with time is common to spectrum and waveform views, and is already in lib-screen-geometry which does not even depend  on wxBase.

The two kinds of views define very different meanings for y-coordinates however and would have completely different paint procedures, in the detailed use of that caching and blitting layer.

Really I must go beyond my simple reaction to the pull request title and really review it to see what's really there.  I'm still waking up.

libraries/lib-graphics/graphics/Point.h
```diff
@@ -0,0 +1,192 @@
...
+#include <cmath>
+#include <type_traits>
+#include <numeric>
+
+namespace graphics
+{
+
+//! A point in 2D space.
+template <typename DataType>
+struct PointType final
```
Paul-Licameli, line 22: I would make a type distinction between points and vectors.  Then type checking will not allow meaningless operations.

Two vectors can be added or subtracted to make a vector.  A vector can be scaled by a number.

Addition of a point to a point or scaling of a point does not make sense.  Subtracting a point from a point does make sense, but the result is not a point, but a vector.

libraries/lib-graphics/graphics/Point.h
```diff
@@ -0,0 +1,192 @@
...
+      return *this;
+   }
+
+   PointType& operator/=(PointType rhs) noexcept
+   {
+      x /= rhs.x;
+      y /= rhs.y;
+
+      return *this;
+   }
```
Paul-Licameli, line 57: Why are the two operations above useful?  They are not usual operations you want with vectors.

libraries/lib-graphics/graphics/Point.h
```diff
@@ -0,0 +1,192 @@
...
+
+   template <typename ScaleType> PointType& operator*=(ScaleType scale) noexcept
+   {
+      x = static_cast<DataType>(x * scale);
+      y = static_cast<DataType>(y * scale);
+
+      return *this;
+   }
+
+   template <typename ScaleType> PointType& operator/=(ScaleType scale) noexcept
```
Paul-Licameli, line 67: Comment `@pre scale != 0`

libraries/lib-graphics/graphics/Point.h
crsib: So essentially, the points are meaningless; vectors can be used all over the place :-)

libraries/lib-graphics/graphics/Point.h
```diff
@@ -0,0 +1,192 @@
...
+{
+   return { lhs.x * rhs.x, lhs.y * rhs.y };
+}
+
+template <typename DataType>
+PointType<DataType>
+operator/(PointType<DataType> lhs, PointType<DataType> rhs) noexcept
+{
+   return { lhs.x / rhs.x, lhs.y / rhs.y };
+}
```
Paul-Licameli, line 134: Again, two questionable, non-member operations

libraries/lib-graphics/graphics/Point.h
```diff
@@ -0,0 +1,192 @@
...
+   return { static_cast<DataType>(lhs.x / rhs),
+            static_cast<DataType>(lhs.y / rhs) };
+}
+
+//! Returns the dot (inner) product between the vectors, defined by points.
+template <typename DataType>
+auto DotProduct(PointType<DataType> lhs, PointType<DataType> rhs) noexcept
+{
+   return lhs.x * rhs.x + lhs.y * rhs.y;
+}
```
Paul-Licameli, line 162: A vector, not point operation

libraries/lib-graphics/graphics/Point.h
crsib: Welcome to the world of graphics :-) Probably this operation is not used though

libraries/lib-graphics/graphics/Point.h
crsib: This depends on ScaleType

libraries/lib-graphics/graphics/Point.h
```diff
@@ -0,0 +1,192 @@
...
+{
+   return std::sqrt(DotProduct(lhs, lhs));
+}
+
+//! Returns the distance between the two points
+template <typename DataType>
+auto Distance(PointType<DataType> lhs, PointType<DataType> rhs) noexcept
+{
+   return std::sqrt(DotProduct(lhs, rhs));
+}
```
Paul-Licameli, line 176: That's wrong.  Distance between points is norm of the difference.

libraries/lib-graphics/graphics/Point.h
```diff
@@ -0,0 +1,192 @@
...
+{
+   return lhs.x * rhs.x + lhs.y * rhs.y;
+}
+
+//! Returns the length of a vector matching the point.
+template <typename DataType>
+auto Norm(PointType<DataType> lhs) noexcept
+{
+   return std::sqrt(DotProduct(lhs, lhs));
+}
```
Paul-Licameli, line 169: Norm is a function of a vector.

libraries/lib-graphics/graphics/Point.h
crsib: This one is an unfortunate mistake

libraries/lib-graphics/CMakeLists.txt
```diff
@@ -5,6 +5,12 @@ A library responsible for custom rendering in Audacity
 set( SOURCES
    FrameStatistics.cpp
    FrameStatistics.h
+
+   graphics/Color.cpp
+   graphics/Color.h
+   graphics/Point.h
+   graphics/Rect.h
+   graphics/Size.h
```
Paul-Licameli, line 13: I would put Point, Rect, and Size in lib-screen-geometry which can be just about the geometry with no mention of color.  That's a leaf library not even using lib-utility.

Yes, lib-screen-geometry still has extraneous constants about the particular Audacity 3 layout, as we discussed before.   Let me figure out how to clean that up.

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+template<typename DataType>
+struct AxisAlignedBoundingBoxType
+{
+   //! Top left corner of the box
+   PointType<DataType> topLeft { std::numeric_limits<DataType>::max(),
+                                 std::numeric_limits<DataType>::max() };
+
+   //! Bottom right corner of the box
+   PointType<DataType> bottomRight { std::numeric_limits<DataType>::min(),
+                                     std::numeric_limits<DataType>::min() };
```
Paul-Licameli, line 108: `lowest` or `min`?

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+
+   //! Return true if the box is not empty.
+   bool IsValid() const noexcept
+   {
+      return topLeft.x <= bottomRight.x && topLeft.y <= bottomRight.y;
```
Paul-Licameli, line 113: So the convention is that x coordinates increase rightwards, y coordinates downwards.  It's the usual graphics convention, but worth explicit comment near the top of the class definition.

libraries/lib-graphics/CMakeLists.txt
crsib: The "problem" with this specific PR is that it is two years old and predates many of the libraries. I haven't re-reviewed much myself, but most, if not all, of the lib-graphics are likely not strictly needed for the purposes of the pull request. 

I am not sure that we want it for 3.5 or for Audacity 3 at all; this was never discussed @ianalarina. Please note that the PR is not even on the board.

libraries/lib-graphics/graphics/Rect.h
Paul-Licameli: Are the bottom and right coordinates included or excluded?

libraries/lib-graphics/graphics/Rect.h
Paul-Licameli: A box could be valid but empty, if bottom and right are excluded.

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+   explicit AxisAlignedBoundingBoxType(RectType<DataType> rect)
+       : topLeft(rect.origin)
+       , bottomRight(rect.origin + PointType<DataType>{ rect.size.width, rect.size.height })
+   {
+   }
+
+   //! Constructs the box from the specified points.
+   AxisAlignedBoundingBoxType(PointType<DataType> p1, PointType<DataType> p2)
+       : topLeft { std::min(p1.x, p2.x), std::min(p1.y, p2.y) }
+       , bottomRight { std::max(p1.x, p2.x), std::max(p1.y, p2.y) }
```
Paul-Licameli, line 135: Inclusive, or exclusive of p2?

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+      return result;
+   }
+
+   //! Expands the box to include the specified rectangle.
+   AxisAlignedBoundingBoxType& Expand(const RectType<DataType>& rect) noexcept
+   {
+      topLeft.x = std::min(topLeft.x, rect.origin.x);
+      topLeft.y = std::min(topLeft.y, rect.origin.y);
+      bottomRight.x = std::max(bottomRight.x, rect.origin.x + rect.size.width);
+      bottomRight.y = std::max(bottomRight.y, rect.origin.y + rect.size.height);
```
Paul-Licameli, line 169: bottomRight describes a point EXCLUDED from the given rectangle.

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+   AxisAlignedBoundingBoxType(PointType<DataType> p1, PointType<DataType> p2)
+       : topLeft { std::min(p1.x, p2.x), std::min(p1.y, p2.y) }
+       , bottomRight { std::max(p1.x, p2.x), std::max(p1.y, p2.y) }
+   {
+   }
+
+   AxisAlignedBoundingBoxType(const AxisAlignedBoundingBoxType &) = default;
+   AxisAlignedBoundingBoxType(AxisAlignedBoundingBoxType &&) = default;
+   AxisAlignedBoundingBoxType &operator=(const AxisAlignedBoundingBoxType &) = default;
+   AxisAlignedBoundingBoxType &operator=(AxisAlignedBoundingBoxType &&) = default;
```
Paul-Licameli, line 142: There are only numbers in this class, so I think you don't need moves

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+      topLeft.x = std::min(topLeft.x, point.x);
+      topLeft.y = std::min(topLeft.y, point.y);
+      bottomRight.x = std::max(bottomRight.x, point.x);
+      bottomRight.y = std::max(bottomRight.y, point.y);
+
+      return *this;
+   }
+
+   //! Expands the copy of the box to include the specified rectangle.
+   AxisAlignedBoundingBoxType Expanded(PointType<DataType> point) const noexcept
```
Paul-Licameli, line 156: I might name it with a noun, `Expansion`

libraries/lib-graphics/graphics/Rect.h
Paul-Licameli: There is an assumption that the rectangle is valid -- not negative area

libraries/lib-graphics/graphics/Rect.h
crsib: I have a strong opinion that function and methods must be verbs. I will push hard towards it.

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+   const PointType<DataType> p12 = { firstRect.origin.x + firstRect.size.width,
+                                     firstRect.origin.y +
+                                        firstRect.size.height };
+   const PointType<DataType> p21 = secondRect.origin;
+   const PointType<DataType> p22 = {
+      secondRect.origin.x + secondRect.size.width,
+      secondRect.origin.y + secondRect.size.height
+   };
+
+   if (p12.x < p21.x || p11.x > p22.x || p11.y > p22.y || p12.y < p21.y)
```
Paul-Licameli, line 62: I think `>=` and `<=` are correct.  `p12` and `p22` are excluded points of the rectangles.

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+      return {};
+
+   const PointType<DataType> p31 = { std::max(p11.x, p21.x),
+                                     std::max(p11.y, p21.y) };
+
+   const PointType<DataType> p32 = { std::min(p12.x, p22.x),
+                                     std::min(p12.y, p22.y) };
+
+   return { p31, SizeType<DataType> { p32.x - p31.x, p32.y - p32.y - p31.y } };
```
Paul-Licameli, line 71: This calculation is correct, but making the change above may return a different rectangle when the resulting area is zero.

libraries/lib-graphics/graphics/Rect.h
Paul-Licameli: Also if you use {} for zero-area rectangles, canonicalizing origin to zero -- then should this do nothing for such rectangles?  Or is that the caller's problem to check?

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+      topLeft.x = std::min(topLeft.x, rect.origin.x);
+      topLeft.y = std::min(topLeft.y, rect.origin.y);
+      bottomRight.x = std::max(bottomRight.x, rect.origin.x + rect.size.width);
+      bottomRight.y = std::max(bottomRight.y, rect.origin.y + rect.size.height);
+
+      return *this;
+   }
+
+   //! Expands the copy of the box to include the specified rectangle.
+   AxisAlignedBoundingBoxType Expanded(const RectType<DataType>& rect) const noexcept
```
Paul-Licameli, line 175: Or `Expansion`

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+      topLeft.x = std::min(topLeft.x, other.topLeft.x);
+      topLeft.y = std::min(topLeft.y, other.topLeft.y);
+      bottomRight.x = std::max(bottomRight.x, other.bottomRight.x);
+      bottomRight.y = std::max(bottomRight.y, other.bottomRight.y);
+
+      return *this;
+   }
+
+   //! Expands the copy of the box to include the specified box.
+   AxisAlignedBoundingBoxType Expanded(const AxisAlignedBoundingBoxType other) const noexcept
```
Paul-Licameli, line 194: ditto

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+      AxisAlignedBoundingBoxType result = *this;
+      result.Expand(other);
+      return result;
+   }
+
+   //! Returns true if the box contains the specified point.
+   bool Contains(PointType<DataType> point) const noexcept
+   {
+      return (point.x >= topLeft.x) && (point.x <= bottomRight.x) &&
+             (point.y >= topLeft.y) && (point.y <= bottomRight.y);
```
Paul-Licameli, line 205: Ah, here it is:  bottom and right are indeed meant to be included.  That is not like top + height or bottom + width of a rectangle.

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+   }
+
+   //! Returns true if the box contains the specified rectangle.
+   bool Contains(const RectType<DataType>& rect) const noexcept
+   {
+      return (rect.origin.x >= topLeft.x) &&
+             (rect.origin.x + rect.size.width <= bottomRight.x) &&
+             (rect.origin.y >= topLeft.y) &&
+             (rect.origin.y + rect.size.height <= bottomRight.y);
```
Paul-Licameli, line 214: Wrongly returns false when the true bottom or right edge of the rectangle coincides with the included boundary.

Fix with `- 1` twice

crsib:
I would like the review to pause, for now, @Paul-Licameli 

1. It was not agreed yet that we need it for 3.5 or for Audacity 3 at all.
2. Most of the comments are related to disliking names so far. This brings an enormous cognitive load when reading the comments.
3. Likely the part you are looking at now can be omitted.
4. The next commit is enourmous.  Unless we agree first on some basic ideas I expect the review to span hundreds of comments again. I would like to avoid that. 
5. I don't remember the details of the code, so I can't respond to the comments.

If Martin and Yana would approve moving this PR forward, we will clean it up with Vitaly first and discuss the ideas on the dev call first.

libraries/lib-graphics/graphics/Rect.h
Paul-Licameli: It is possible that a box contains both r1 and r2, but does not contain `Intersect(r1, r2)` if they are disjoint, because you canonicalize the empty rectangle to all zeroes.  That seems weird, not corresponding to set-theoretic identities.

Maybe the function should always return true for empty rectangles, regardless where the top left corner is.

libraries/lib-graphics/graphics/Size.h
```diff
@@ -0,0 +1,175 @@
...
+{
+namespace details
+{
+template <typename DataType>
+constexpr DataType GetPositiveSizeValue(DataType value) noexcept
+{
+   if constexpr (std::is_unsigned_v<DataType>)
+      return value;
+   else
+      return value >= DataType {} ? value : -value;
```
Paul-Licameli, line 26: I think `value >= DataType{}` would optimize away anyway for unsigned types

libraries/lib-graphics/graphics/Rect.h
```diff
@@ -0,0 +1,267 @@
...
+{
+   return { point_cast<To>(rect.origin), size_cast<To>(rect.size) };
+}
+
+//! Alias for rectangle with float data type.
+using Rect = RectType<float>;
+
+//! Axis aligned bounding box.
+template<typename DataType>
+struct AxisAlignedBoundingBoxType
```
Paul-Licameli, line 100: I would call it `AxisAlignedBoundingBox`

libraries/lib-graphics/graphics/Size.h
```diff
@@ -0,0 +1,175 @@
...
+}
+} // namespace details
+
+//! A class that represents size in 2D space.
+template <typename DataType>
+struct SizeType final
```
Paul-Licameli, line 32: Maybe `Size` could turn into the 2d vector type, distinguished from a point.  A point and a diagonal vector determine a rectangle.

Paul-Licameli:
> I would like the review to pause, for now, @Paul-Licameli
> 
>     1. It was not agreed yet that we need it for 3.5 or for Audacity 3 at all.
> 
>     2. Most of the comments are related to disliking names. This brings an enormous cognitive load when reading the comments.
> 
>     3. Likely the part you are looking at now can be omitted.
> 
>     4. The next commit is enourmous.  Unless we agree first on some basic ideas I expect the review to span hundreds of comments again. I would like to avoid that.
> 
>     5. I don't remember the details of the code, so I can't respond to the comments.
> 
> 
> If Martin and Yana would approve moving this PR forward, we will clean it up with Vitaly first and discuss the ideas on the dev call first.

All right, I will stop.  But please note:  there is more than just naming to object to.  I think there are real off-by-one errors in the bounding box class.

crsib:
>  I think there are real off-by-one errors in the bounding box class.

I agree, there are errors for sure.

Paul-Licameli:
> > I think there are real off-by-one errors in the bounding box class.
> 
> I agree, there are errors for sure.

I'm NOT reviewing, just noting other thoughts while compiling something else.

There IS a need to distinguish point and vector.  Surely Rectangle stores a point for top left, and BoundingBox for top left and bottom right.

I didn't review Size completely, but it too has some strangeness (like multiplication of two sizes, why?).  Maybe Size should become the vector class and a Rectangle can be defined by a point and a diagonal vector.

But should rectangles be "oriented" (with possibly negative area) or just have non-negative area always? -- Then maybe Rectangle should instead always canonicalize itself to have a diagonal with non-negative components, and the corner as the real top-left.

crsib:
There is a common terminology in UI toolkits, where Point has some vector-like properties. You can observe similar properties with QPoint and wxPoint. There is absolutely no need to distinguish points and vectors unless the ultimate goal is to teach students about doing dimensions in C++. I don't recall the concept of "point" in math though. They are pretty much the synonyms, used interchangeably.

Usually rendering engines call such structures "vectors" and UI toolkit "points". I think that "Point" is used in UI for the last 40 years or so. We can rename it to vector, just to save time on this code review, but I'd rather not. 

I am hard against having both points and vectors. It will create lots of mess in the code for very little and strictly academic (?) reason. We are writing software, not a research paper. It must be fast, readable, and maintainble.

Both vector and points may have component by component division, even if it makes little sense from the mathematical perspective. They are not multiplicative inverses, but from the programming point of view they can be useful. For example , when you need to represent a different scale on different axises. In fact, this operation is implemented in hardware both for CPUs and GPUs.

Defining a rectangle is quite complex topic indeed. That is why there are two different definitions in this library alone. However, the way how rectangle is defined is never about defining the most correct one from the math point of view. It's about convince and performance.

I don't think that we need AABB for the sake of this PR. They were used in code that was further 10x faster than this. That's it, 100x faster than Audacity is now.

However, I think that Point (with vector properties, sorry), Size, and Rect as (Point, Size) could be useful overall. I agree, that there is a better place for them now than lib-graphics. They surely need better unit test coverage. I would love to have adequate Color class too. wxColour is quite heavy and not too useful.

To summarize, I don't like renaming points to vectors, but I don't care much. I will stand hard against adding vectors on top of the points or defining rectangles using diagonals.

Paul-Licameli:
The standard library distinguishes time points from durations in chrono.  It does not affect the performance of compiled code but it makes distinctions between variables that allow programming errors to be caught at compile time and increases confidence in correctness of code that does compile.

What I suggest is simply analogous to that.

Paul-Licameli:
Affine space is a different mathematical notion from vector space.  One has points with no distinguished origin.  The other describes displacements in an affine space.  https://en.m.wikipedia.org/wiki/Affine_space

crsib:
To be able to describe a point in affine space you need to have a basis. When you have a basis vectors and points are pretty much the synonyms.

I don't see how std::chrono and Points as defined in UI are related. Chrono attaches units to the durations and time_points have implementation defined basis, unknown to the user in principle. Both concerns are critical to chrono. At the same time chrono is very verbose and rarely used in extensive computations.

In graphics, we always have a well defined basis for the screen space. We don't need to attach units to the points (this is not true for typography or for physics, but we are not doing a physics library). The vectors and points are the synonyms in this case. There is absolutely no need to separate between them except making the point (no pun intended). I won't change my opinion. In case the discussion about this topic will take too much time, this PR will be dropped and QPoint (with vector properties!) will be used for the needs of the Qt version of the Audacity.

Paul-Licameli:
You asked my opinion in review and I gave it.  I would design these classes differently.

Points and vectors may have a common representation but they do not have the same algebra.  A static type distinction between them can catch errors at compile time and also make the signatures of functions more informative.

Just as we distinguish different types of Identifier rather than pass naked strings around, or use enumerations — similarly there are number pairs used for different purposes and I would distinguish them.

A point is a position in a geometric space.  A vector is not, but is a displacement from point to point.  Scaling a vector is meaningful.  Scaling a point is not.

crsib:
And I have strong reasons to disagree with you.

> they do not have the same algebra

I'm unsure how to read this, but I will stress again: we are not doing academic papers, and we are not implementing an abstract concept.

> Just as we distinguish different types of Identifier 

We don't. There are 2 cases in the code where TaggedIdentifier is used, and URLString is a clearly a bad example of one. It is not used in the code anymore; I have forgotten to remove it. Otherwise, we just use Identifier. It gives no better type-safety over the raw strings. We have exactly one type-safe identifier: NormalizedKeyString. Is it an Identifier at all? I don't know, but that feels a bit wrong.

>  or use enumerations

We rarely use them, preferring Identifiers, which are essentially strings and are not type-safe.

Gosh, we have places where `any` is used. We are constantly dependent on a global state. We have so many places that need attention right now, so why waste time and mental resources on discussing points vs vectors, irrelevant to the use case they are designed for?

> Scaling a point is not.

This statement is wrong when we are talking about GUI of any kind. At least, since the first release of Android :-)

---

I honestly feel that we are in the "snapping" PR space again. However, that PR was critical because it was fixing the functionality required for release, which was implemented wrong. This one is not. I have no time, or energy, or will to discuss this topic.

libraries/lib-utility/MemoryX.h
```diff
@@ -400,4 +401,40 @@ constexpr IntType SwapIntBytes(IntType value) noexcept
...
+         const auto result = numerator / denominator;
+
+         if (result < 0 || result * denominator == numerator)
+            return result;
+         else
+            return result + 1;
+      }
+   }
+}
+
```
Paul-Licameli, line 439: I don’t like adding to MemoryX.h.  These might be in a new small header in lib math

src/Experimental.cmake
```diff
@@ -106,6 +106,10 @@ set( EXPERIMENTAL_OPTIONS_LIST
    # PRL disabled 31 Aug 2021 because, at least on Mac, it misfires the
    # preference dialog whenever any keystroke shortcuts with Shift+ are used
    # EASY_CHANGE_KEY_BINDINGS
+
+   # DV 26 December 2023
+   # Use lib-wave-track-paint for in-memory waveform image generation
```
Paul-Licameli, line 111: I just merged a PR commit-experimental-branches and proposed another, maintain-experimental-branches.  I want to eliminate Experimental.cmake.

See the second PR.  If you are not certain to commit to using the experiment, instead define a bool constant in namespace Experimental.  Use if constexpr not ifdef to control new code.  That makes the compiler still check that it can build, when it does not build.

crsib: If this PR passes QA, I think the proper solution would be to remove the old code altogether.

libraries/lib-wave-track-paint/GraphicsDataCache.cpp
```diff
@@ -0,0 +1,434 @@
...
+      return;
+
+   const int64_t left  = zoomInfo.TimeToPosition(t0);
+   const int64_t right = zoomInfo.TimeToPosition(t1) + 1;
+
+   mLRUHelper.reserve(cacheRangeLength);
+
+   for (auto it = ppsMatchRange.first; it != ppsMatchRange.second; ++it)
+   {
+      if (it->Key.FirstSample < left)
```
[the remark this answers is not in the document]
crsib: An excellent point! Let me just kill this method LOL

libraries/lib-utility/MemoryX.h
crsib: Sure

libraries/lib-wave-track-paint/GraphicsDataCache.cpp
```diff
@@ -0,0 +1,434 @@
...
+      GetPPSMatchRange(mLookup, pixelsPerSecond, mSampleRate);
+
+   for (int64_t itemIndex = 0; itemIndex < cacheItemsCount; ++itemIndex)
+   {
+      const int64_t column = cacheLeftColumn + itemIndex * CacheElementWidth;
+
+      const int64_t firstSample =
+         static_cast<int64_t>(column * samplesPerPixel);
+
+      const auto it = std::find_if(
```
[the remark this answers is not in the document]
crsib: I have never seen this function as a hot spot.

Generally, with the current implementation, the cache can hold up to `width * 4 / 256`, which limits the cache size to under 64 items on the 4k display. I think the difference between linear and logarithmic time is negligible in this case.

vsverchinsky (commented):
All good

grliszas14 (approved):

Issue #5117, opened 2023-08-31
Accurate mixer board (or other per-track) meter output; metering intermediate effect stages too


Paul-Licameli:
### Your idea

In the mixer board, or whatever other separate per-track meters Audacity will have, define what exactly should the levels correspond to, and solve the non-trivial implementation problems.

### Problem to be solved

Should the per-track meter really show the contribution of the associated track to the output -- including influence of gain, pan, envelope, stretch, and stacked effects -- just before those samples are mixed with all other tracks?

Present, old Mixer Board code assumed it can map from an interval of track times just played to a portion of a WaveTrack, and recreate the calculation.  But it for many versions, it only included the gain slider -- so showing possibly wrong levels even then -- and since then Audacity's playback is further complicated by stretch and effects.

Better that a queue of intermediate data should be tapped from the sample streams flowing through the audio engine and communicated in a properly race-free way to the main thread which paints the meters.

Good enough code to implement such lock-free queues already exists in MeterPanel.cpp but is used only for the meters in the toolbars.

Rearrangement of some audio engine code, and perhaps elaboration of the interfaces in lib-audio-graph, might make it easier to connect intermediate stages of sample processing to the meters, and also even allow the monitoring of intermediate stages of effect stack calculation (for such purposes as detecting clipping).

Paul-Licameli:
@LWinterberg what do you think?  Incidentally to other code reviews, I was reminded of this problem, and it might be a nice project for next release.

Even if Mixer Board will be mostly thrown away for Qt migration, still this problem that needs solving is independent of the UI toolkit.

Paul-Licameli:
@DilsonsPickles

Paul-Licameli:
> Playback meters should display what you hear, including all effects and gain sliders and whatever ("post-fader").

Or more precisely, what you hear when the track is solo?

Making Mixer Board really work this way would be a worthwhile development even if Mixer Board as presently designed looks obsolete.

Paul-Licameli:
The only really correct meters are now those in the toolbar.

Separate track data, before the mix of tracks to output, is not yet properly communicated to the main thread for painting mixer board meters.

LWinterberg:
> Or more precisely, what you hear when the track is solo?

yes.

PR #5830, opened 2023-12-30
Per track meter internals

Paul-Licameli:

Depends on
- #6442

Some of this originated in #5779 

Put into AudioIO.cpp all the support needed for per-track meters displaying correct peak and RMS data
(post all adjustments for the effect stack, pan, gain, envelope).

But also other cleanups of AudioIO.  Move the experimental automated input level adjustment out, so the Meter
interface needs fewer virtual functions.

Also clean up the publication of events from AudioIO and correct some non-thread-safe message callbacks.

To do:  MixerBoard should use this.  It doesn't show correct data now in all cases.  See #5117 

Recommended:


src/widgets/MeterPanel.cpp
```diff
@@ -1126,8 +1126,8 @@ void MeterPanel::OnMeterUpdate(wxTimerEvent & WXUNUSED(event))
 
    if (numChanges > 0) {
       #ifdef EXPERIMENTAL_AUTOMATED_INPUT_LEVEL_ADJUSTMENT
-         if (gAudioIO->AILAIsActive() && mIsInput && !discarded) {
-            gAudioIO->AILAProcess(maxPeak);
+         if (AudioIO::Get()->AILAIsActive() && mIsInput && !discarded) {
```
crsib: It seems that the only place `EXPERIMENTAL_AUTOMATED_INPUT_LEVEL_ADJUSTMENT` should be used in such approach is `AILAIsActive`.

src/widgets/MeterPanel.cpp
```diff
@@ -1126,8 +1126,8 @@ void MeterPanel::OnMeterUpdate(wxTimerEvent & WXUNUSED(event))
...
+            AudioIO::Get()->AILAProcess(maxPeak);
             putchar('\n');
```
crsib: But this code is disturbing. Why is it here? 

This transformation is not quite conservative anyway, so probably this line should be removed?

It seems that AILA logs extensively, probably way to extensively. Is it ready for use at all?

src/widgets/MeterPanel.cpp
```diff
@@ -830,23 +823,16 @@ void MeterPanel::SetStyle(Style newStyle)
...
-      int inputSource;
-
+   if (mSlider) {
       Refresh();
-
       auto gAudioIO = AudioIO::Get();
-      gAudioIO->GetMixer(&inputSource, &inputVolume, &outputVolume);
-
+      auto settings = gAudioIO->GetMixer();
+      auto &[inputSource, inputVolume, outputVolume] = settings;
```
crsib, line 715: I think the code would be much more readable without this line.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2376,7 +2377,9 @@ void AudioIO::AILAProcess(double maxPeak) {
 
          putchar('\n');
          mAILAMax = pInputMeter ? ToLinearIfDB(mAILAMax, pInputMeter->GetDBRange()) : 0.0;
-         double iv = (double) Px_GetInputVolume(mPortMixer);
+         auto settings = audioIO.GetMixer();
+         auto &[_, inputVolume, __] = settings;
```
crsib: Again, why?

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2401,14 +2404,16 @@ void AudioIO::AILAProcess(double maxPeak) {
...
                auto msg = XO(
-"Automated Recording Level Adjustment decreased the volume to %f.").Format( vol );
+"Automated Recording Level Adjustment decreased the volume to %f.")
+                  .Format(inputVolume);
                ProjectStatus::Get( *proj ).Set(msg);
                changetype = 1;
-               wxPrintf("\tnew vol:%f\n", vol);
-               float check = Px_GetInputVolume(mPortMixer);
+               wxPrintf("\tnew vol:%f\n", inputVolume);
+               auto [_, check, __] = audioIO.GetMixer();
```
crsib: This one is the weirdest case of structured binding use so far.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2348,14 +2348,18 @@ bool AudioIO::AILAIsActive() {
...
-   wxPrintf("START TIME %f\n\n", mAILAAbsolutStartTime);
+   mAILAAbsoluteStartTime = AudioIO::Get()->GetClockTime();
+   wxPrintf("START TIME %f\n\n", mAILAAbsoluteStartTime);
 }
 
 double AudioIO::AILAGetLastDecisionTime() {
    return mAILAAnalysisEndTime;
 }
 
+double AudioIO::GetClockTime() const {
```
crsib, line 2287: `GetStreamTime` returns stream time specific to Audacity, `GetClockTime` - actual stream time. Quite confusing, and I don't understand why is it called "clock" time.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -224,8 +224,12 @@ bool AudioIO::ValidateDeviceNames(const wxString &play, const wxString &rec)
    return pInfo != nullptr && rInfo != nullptr && pInfo->hostApi == rInfo->hostApi;
 }
 
+static constexpr auto kTimerInterval = std::chrono::milliseconds{ 50 };
+
 AudioIO::AudioIO()
 {
+   wxTimer::Start(kTimerInterval.count());
```
crsib, line 218: There is no safety around what kind of duration the kTimerInterval will be. Explicit `duration_cast<milliseconds>` would help.

Paul-Licameli:
I should have put the big red Experimental label on this bit of holiday hackathon.  I didn't ask review yet, but thank you.

Paul-Licameli:
@saintmatthieu @vsverchinsky 

Rebased onto latest master.

There is much cleanup of AudioIO.cpp that I prepared.  It's still not all of the cleanup.

Correct inter-thread communication of the stream of most recently played samples is necessary for a per-track meter display that is not meaningless garbage.

PR #5843, opened 2024-01-05
Benchmarking and visualization test

Paul-Licameli:
This is unusual... This is a place for me to review the regression test, so all my comments aren't lost.

Recommended:


libraries/lib-music-information-retrieval/tests/MirTestUtils.cpp
```diff
@@ -0,0 +1,83 @@
+#include "MirTestUtils.h"
```
Paul-Licameli, line 1: banner

libraries/lib-music-information-retrieval/tests/MirTestUtils.h
```diff
@@ -0,0 +1,143 @@
...
+#include <numeric>
+#include <string>
+#include <vector>
+
+namespace MIR
+{
+
+// Some tests, such as benchmarking and visualization, are not meant to be run
+// on CI. This variable is used to disable them.
+static constexpr auto runLocally = false;
```
Paul-Licameli, line 18: So I understand you change this to true, run locally, and then decide just two magic number thresholds to put into the application, in `MusicInformationRetrieval.h`.  The very last thing in `GetMeterUsingTatumQuantizationFit.cpp` is use of those thresholds.

Maybe that header file should have a comment about how the numbers are experimentally derived.

libraries/lib-music-information-retrieval/tests/MirTestUtils.h
```diff
@@ -0,0 +1,143 @@
...
+   const auto it = std::upper_bound(
+      falsePositiveRates.begin(), falsePositiveRates.end(),
+      allowedFalsePositiveRate);
+
+   if (it == falsePositiveRates.end())
+      // All breakpoints satify the constraint. Return the least score.
+      return { auc, results.back().score };
+   else if (it == falsePositiveRates.begin())
+      // No breakpoint satisfies the constraint. Return the greatest score.
+      return { auc, results.front().score };
```
Paul-Licameli, line 90: I think this branch should return greatest score plus some epsilon.

libraries/lib-music-information-retrieval/tests/MirTestUtils.h
```diff
@@ -0,0 +1,143 @@
...
+   double remainder;
+};
+
+OctaveError GetOctaveError(double expected, double actual);
+
+// A simple hashing that add the sum of all samples to the given hash.
+// Reproducible benchmarking must use the same input. We use this to make sure
+// that it does.
+template <int bufferSize = 1024>
+void UpdateHash(const MirAudioReader& source, float& hash)
```
Paul-Licameli, line 145: Maybe write it in functional style, taking a value and returning a new value.

I lift eyebrows to see a floating point type for a "hash" value.  I looked for uses of this function.  You are not really using it in a standard unordered container, which would require some quantization of the value to `size_t`.  I think "checksum" better describes the use of the value.

libraries/lib-music-information-retrieval/tests/TatumQuantizationFitBenchmarkingOutput/summary.txt
```diff
@@ -0,0 +1,5 @@
+AUC: 0.9093477144324603
+B&M Threshold: 0.8423423487195607
+M&S Threshold: 0.9160608375297137
+Octave error RMS: 0.1024786300174427
+Audio file hash: -1840.31005859375
```
Paul-Licameli, line 5: That's strange -- I thought the checksum would be a summation of samples of audio files, which would tend to average toward 0.

libraries/lib-music-information-retrieval/tests/TatumQuantizationFitBenchmarking.cpp
```diff
@@ -0,0 +1,220 @@
...
+   // 4. the hash of the audio files used.
+   //
+   // The AUC can only be used for comparison if the hash doesn't change. At the
+   // time of writing, the benchmarking can only conveniently be run on the
+   // author's machine (Windows), because the files used are not
+   // redistributable. Setting up a redistributable dataset that can be used by
+   // other developers is currently being worked on.
+
+   // We only observe the results for the most lenient classifier. The other,
+   // stricter classifier will yield the same results, only with less false
```
Paul-Licameli, line 121: fewer -- sorry to be pedantic

libraries/lib-music-information-retrieval/tests/MirTestUtils.cpp
```diff
@@ -0,0 +1,83 @@
...
+   for (int i = 0; i < width; ++i)
+      if (i < progress)
+         std::cout << "=";
+      else
+         std::cout << " ";
+   std::cout << "] " << std::setw(3) << percent << "%\r";
+   std::cout.flush();
+}
+
+OctaveError GetOctaveError(double expected, double actual)
```
Paul-Licameli, line 67: I or another maintainer would need help understanding this uncommented function.

I think I understand, from examining usage, that expected and actual really are frequencies, but not of audible tones -- rather the low frequencies of beats per minute.

"expected" might perhaps be called "nominal" as the bpm parsed from a filename.  "actual" might perhaps be called "measured" as by your analyses.

The function finds a discrepancy between them on a logarithmic scale.  (Really any logarithmic base could do as well as 2 for the purpose of ranking.)

Except that it allows displacing one of them on the scale by the listed five factors, whichever minimizes error.  I don't quite get that part.

How should I interpret it when the chosen factor is not 1?

I look through your csv files for such examples.  They are a minority of the positive population.  Only one with 1/3, three with 1/2, one with 3.  2 is more frequent.

libraries/lib-music-information-retrieval/tests/TatumQuantizationFitBenchmarking.cpp
```diff
@@ -0,0 +1,220 @@
...
+               << rocInfo.threshold << "\n"
+               << "Octave error RMS: " << octaveErrorStd << "\n"
+               << "Audio file hash: " << hash << "\n";
+
+   constexpr auto previousAuc = 0.9093477144324603;
+
+   // If this changed, then some non-refactoring code change happened. If
+   // `rocInfo.areaUnderCurve > previousAuc`, then there's probably no argument
+   // about the change. On the contrary, though, the change is either an
+   // inadverted bug, and if it is deliberate, should be well justified.
```
Paul-Licameli, line 217: inadvertent

libraries/lib-music-information-retrieval/tests/TatumQuantizationFitBenchmarking.cpp
```diff
@@ -0,0 +1,220 @@
...
+   REQUIRE(rocInfo.areaUnderCurve == previousAuc);
```
Paul-Licameli, line 218: I think you might not get a precise match of numbers if you run the tests an a different platform or CPU

libraries/lib-music-information-retrieval/tests/TatumQuantizationFitBenchmarking.cpp
```diff
@@ -0,0 +1,220 @@
...
+   std::ofstream summaryFile {
+      std::string(CMAKE_CURRENT_SOURCE_DIR) +
+      "/GetMusicalMeterFromSignalBenchmarkingOutput/summary.txt"
+   };
+
+   summaryFile << std::setprecision(std::numeric_limits<double>::digits10 + 1)
+               << "AUC: " << rocInfo.areaUnderCurve << "\n"
+               << "Strict Threshold (Minutes-and-Seconds): " << strictThreshold
+               << "\n"
+               << "Lenient Threshold (Beats-and-Measures): "
```
Paul-Licameli, line 207: I notice the checked-in summary.txt was not generated by this version of the .cpp file

Paul-Licameli:
Please note the suggested extra commits to fix minor things.

I'm done studying "benchmarking" today and I may try to understand "visualization" another day.

libraries/lib-music-information-retrieval/tests/MirTestUtils.cpp
Paul-Licameli: I think I should understand this allowance of factors as the "common denominator" thing?  You have explained to me how the tatums that the analysis detects might really be more or fewer than the quarter note beats, by various whole or fractional multipliers.

Paul-Licameli: ... and yet `possibleTatumsPerBeat` in `GetMeterTatumUsingQuantizationFit.cpp` lists nine possibilities.

So 4, 6, 1/4, and 1/6 are the even more outlying possibilities, of which, you don't yet have examples in your dataset?

saintmatthieu: The best explanation you can get is probably in the paper I got this from, [Music Tempo Estimation: Are We Done Yet?](https://storage.googleapis.com/jnl-up-j-tismir-files/journals/1/articles/43/submission/proof/43-1-1337-1-10-20200824.pdf). I'll make sure to at least add a comment that refers to this tomorrow.

saintmatthieu: Done.

libraries/lib-music-information-retrieval/tests/TatumQuantizationFitVisualization/visualize_debug_output.py
```diff
@@ -0,0 +1,39 @@
...
+import matplotlib.pyplot as plt
+import math
+
+plt.figure(1)
+# Plot the ODF, using odfSr to convert from samples to seconds
+plt.plot([i/odfSr for i in range(len(odf))], odf)
+# Plot the peaks
+plt.plot([i/odfSr for i in odf_peak_indices], [odf[i] for i in odf_peak_indices], 'ro')
+# Convert tatum rate from tatums per minute to tatums per ODF sample:
+tatumRate = tatumRate / 60.0
```
Paul-Licameli, line 13: Should you then divide by `odfSr` ?  This doesn't agree with the comment.

libraries/lib-music-information-retrieval/tests/TatumQuantizationFitVisualization/visualize_debug_output.py
```diff
@@ -0,0 +1,39 @@
...
+
+# Plot the tatums as vertical lines, offset by `lag` samples
+numTatums = int(math.ceil(len(odf) / odfSr * tatumRate))
```
Paul-Licameli, line 16: But here, if the comment is right, `tatumRate` is tatums / second.  So the previous comment is what's wrong.

saintmatthieu: Addressed.

Issue #5849, opened 2024-01-08
Wrong undo/redo when import silently changes tempo or time signature


Paul-Licameli:
### Bug description

The problem was introduced with the merge of #5570 

It was discovered in code review of #5791 which is not yet merged.  It is therefore not a blocker for that merge.  After that merge, the possible steps to reproduce are more various.

#### Scenario 1
1. Make a new project
2. In Import/Export preferences, change "When Audacity detects music..." to "Switch view..."  i.e. the don't-ask-me option
3. Change tempo and time signatures to non-default, such as 110 and 3/4 using the toolbar
4. Import a clip that has different tempo and time signature, such as 120 and 4/4
5. Undo the import, so you see an empty track panel
6. Redo
7. Observe changes in the toolbar 

#### Scenario 2
1. Use Beats & Bars ruler
2. Have a track already in the project
3. Import a clip that will auto-stretch and whose tempo is different from that of the project
4. Observe stretch in title bar of that clip
5. Undo, redo
6. Observe title bar again and duration

#### Scenario 1
Redo brings back the imported clip, but also 120 and 4/4 in the toolbar

#### Scenario 2
Redo brings back the clip and re-auto-stretches it

#### Scenario 1
Redo does not change the toolbar

#### Scenario 2
Redo brings back the clip but does not re-auto-stretches it

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

Note that if instead the preference is to ask each time, there are separate undo/redo items for the import and then for the signature change.  But the total effect of undoing and redoing both steps is correct.

dozzzzer:
Tested

PR #5850, opened 2024-01-09
Fix undo/redo when import quietly stretches clip or changes tempo/t.s.

Paul-Licameli:
Resolves: #5849

Recommended:

### DIFF
--- a/src/ProjectFileManager.cpp
+++ b/src/ProjectFileManager.cpp
@@ -1399,11 +1399,13 @@ void ReactOnMusicFileImport(
       assert(clips.size() == 1);
       const auto clip = *clips.begin();
 
+      auto &ph = ProjectHistory::Get(*pProj);
       if (!isFirstWaveTrack && isBeatsAndMeasures)
       {
          clip->SetRawAudioTempo(syncInfo.rawAudioTempo);
          clip->TrimQuarternotesFromRight(syncInfo.excessDurationInQuarternotes);
          clip->StretchBy(syncInfo.stretchMinimizingPowOfTwo);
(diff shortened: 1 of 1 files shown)


src/ProjectFileManager.cpp
```diff
@@ -1375,9 +1375,11 @@ void ReactOnMusicFileImport(
 
    if (!isFirstWaveTrack && isBeatsAndMeasures)
    {
+      auto &ph = ProjectHistory::Get(project);
       clip->SetRawAudioTempo(syncInfo.rawAudioTempo);
       clip->TrimQuarternotesFromRight(syncInfo.excessDurationInQuarternotes);
       clip->StretchBy(syncInfo.stretchMinimizingPowOfTwo);
+      ph.ModifyState(true);
```
saintmatthieu, line 1382: I, like you, would have thought this change is necessary now that I see it, but I just tried and undo/redo restores the stretch correctly _even without this change_ ...
I think this is because these lines are executed synchronously to file import, and later there is `   AddImportedTracks(fileName, std::move(newTracks));` which in turns pushes a new state in the history.

saintmatthieu (commented):
Unnecessary call to `ModifyState` in the synchronous branch should be removed. (The bug description can then also be simplified.)

src/ProjectFileManager.cpp
Paul-Licameli: It was unnecessary in the commit I based this on at first.  But it will be necessary in latest master, which delays those lines into the `CallAfter`.

saintmatthieu (approved):

PR #5856, opened 2024-01-10
5795 tempo detection benchmarking improvement

### LINKED ISSUE #5795: Tempo detection benchmarking improvement
A unit-test serves me to control that tweaks or refactoring work does not reduce the quality of the tempo detection.

In short, it works by running the algorithm on all wav files found in a subdirectory, comparing the detected tempo with the BPM info found in the title, and gathering all those results to yield a performance score as well as the thresholds needed in the production code.

There is the major limitation in the current implementation that the files used are not available online, essentially because they are not redistributable.

An interesting resource to consider is freesound.org. The files are numerous, redistributable and some are correctly BPM-annotated. The site even has a developer API enabling the query of files using certain criteria from the command line.

Doing this work now is desirable, because the next steps are fine tuning, where more files than currently used are desirable.

saintmatthieu:
Resolves: #5795

Please refer to 5795 for more detail.
**Note that this PR does not modify production code** (only marginally, see commit comment) **and hence does not require QA.**

Recommended:

### DIFF
--- a/tests/AudioFileIO.cpp
+++ b/tests/AudioFileIO.cpp
@@ -1,0 +1,21 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  AudioFileIO.cpp
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#include "AudioFileIO.h"
+#include "Mp3FileReader.h"
+#include "WavFileIO.h"
+
+bool AudioFileIO::Read(
+   const std::string& path, std::vector<std::vector<float>>& audio,
+   AudioFileInfo& info, const std::optional<std::chrono::seconds>& upTo)
+{
+   return WavFileIO::Read(path, audio, info, upTo) ||
+          Mp3FileReader::Read(path, audio, info);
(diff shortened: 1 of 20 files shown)


tests/AudioFileIO.cpp
```diff
@@ -0,0 +1,11 @@
+#include "AudioFileIO.h"
```
Paul-Licameli, line 1: banner?

tests/AudioFileIO.h
```diff
@@ -0,0 +1,30 @@
...
+#include <vector>
+
+struct AudioFileIO
+{
+   static bool Read(
+      const std::string& path, std::vector<std::vector<float>>& audio,
+      AudioFileInfo& info,
+      const std::optional<std::chrono::seconds>& upTo = std::nullopt);
+
+   static bool Write(
```
Paul-Licameli, line 27: Declared but undefined, at first commit at least

tests/AudioFileInfo.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
Paul-Licameli, line 1: banner

Paul-Licameli (approved):

PR #5859, opened 2024-01-11
Tempo detection: cache miss minimization

### LINKED ISSUE #5854: Tempo detection speed improvement: reduce number of cache misses
The implementation already uses the caching mechanism that was introduced for time stretching, but the algorithm evaluates the hypothesis that the clip is a loop and hence accesses its audio circularly. Readout of the beginning and end of the clips jumps back and forth, leading to cache misses.

saintmatthieu:
Resolves: #5854 

I tested this by adding logging to the `SqliteSampleBlock`. I imported an 8-second mono file, which needs two sqlite blocks. Without this change, there were 11 sqlite queries, against only just 3 with.

Recommended:

QA:
- [x] Smoke-testing the tempo detection feature looks fine. (Most noticeable speed improvement would be on a weak, weak machine for clips between 6 and 10 seconds. Maybe even this may not be noticeable, but the code change was easy.)

### DIFF
--- a/src/ClipMirAudioReader.cpp
+++ b/src/ClipMirAudioReader.cpp
@@ -44,8 +44,10 @@ void ClipMirAudioReader::AddChannel(
    size_t iChannel, float* buffer, sampleCount start, size_t len) const
 {
    constexpr auto mayThrow = false;
-   auto& cache =
-      const_cast<std::optional<AudioSegmentSampleView>&>(mCache[iChannel]);
-   cache.emplace(mClip.GetSampleView(iChannel, start, len, mayThrow));
+   const auto iCache = mUseFirst[iChannel] ? 0 : 1;
+   auto& cache = mCache[iChannel][iCache];
+   auto view = mClip.GetSampleView(iChannel, start, len, mayThrow);
+   cache.emplace(std::move(view));
    cache->AddTo(buffer, len);
+   mUseFirst[iChannel] = !mUseFirst[iChannel];
 }
--- a/src/ClipMirAudioReader.h
+++ b/src/ClipMirAudioReader.h
@@ -38,5 +38,10 @@ private:
       size_t iChannel, float* buffer, sampleCount start, size_t len) const;
 
    const ClipInterface& mClip;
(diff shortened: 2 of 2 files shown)


src/ClipMirAudioReader.cpp
```diff
@@ -32,20 +32,23 @@ void ClipMirAudioReader::ReadFloats(
    float* buffer, long long where, size_t numFrames) const
 {
    std::fill(buffer, buffer + numFrames, 0.f);
-   AddChannel(0, buffer, where, numFrames);
+   auto& mutableMe = const_cast<ClipMirAudioReader&>(*this);
+   mutableMe.AddChannel(0, buffer, where, numFrames);
    if (mClip.GetWidth() == 1)
       return;
-   AddChannel(1, buffer, where, numFrames);
+   mutableMe.AddChannel(1, buffer, where, numFrames);
```
Paul-Licameli, line 39: So, I understand that staggered use of a pair of sample views doubles the number of sample blocks that keep caches, and avoids destruction of the earlier blocks too soon.  

The commit comment misleads?  The problem was not only where the fetching wraps around from end to start, but at other hops too?

It is appropriate to use mutable members instead of const-casting when the purpose is caching.

Paul-Licameli (approved):

src/ClipMirAudioReader.cpp
saintmatthieu: The comment is correct at least in its intention - which isn't saying much. I suppose there's a lot of context to give to fully understand why this is needed, which would probably be overwhelming in a comment. And maybe I'm lazy, too.

Making the members mutable and `AddChannel` const.

saintmatthieu:
> Tested. Importing a 4-bar musical loop before this PR: `Operation 'Music Information Retrieval' took 0.141000 seconds.`
> 
> This PR: `Operation 'Music Information Retrieval' took 0.079000 seconds.`
> 
> Which means about a 78% increase in speed when importing loops.

I never checked that log window before, good to know!
How long is that file of yours ? I expect the improvement to have proportionally most impact for files between 4 and 8 seconds. After that, it should converge with longer files.

Issue #5658, opened 2023-11-20
No longer possible to reset the Maximum Peak and Clipping indicators in the meters


petersampsonaudacity:
### Bug description
It is no longer possible to reset the Maximum Peak and Clipping indicators in the meters.
![image](https://github.com/audacity/audacity/assets/50205984/85d10747-9b21-4ab8-a55d-161581401b60)

This is a regression on 3.1.3 introduced in 3.2.0 when the Meter toolbars and the Mixer toolbar were consolidated (with the recording and playback levels set on a slider in the meter toolbars).   In those earlier versions you could left-click in an active meter to reset it (now left-clicking changes the level setting - just as it used to in the Mixer Toolbar).

_The only workaround is to stop monitoring and restart it - or stop Recording or Playback if they are in progress and restart them._

### Steps to reproduce
1.  right click in the Recording Meter to get the context menu
2. select **Enable Silent Monitoring**
3. make noise to activate meter monitoring
4. try to reset the Maximum Peak (or Clipping indicator if that presented)
5. Observe: you can't

### Expected behavior
An ability to reset either of the meters while they are active.

I would suggest adding an entry for **Reset Meter** (or suchlike) to context menus of both of the meter toolbars.

Example:
![image](https://github.com/audacity/audacity/assets/50205984/ca582120-2015-4f1f-a19d-5f5689e7d46c)

### Actual behavior
No ability to reset the Maximum Peak level or Clipping indicator in either of the meters while they are active

### Audacity Version
3.4.2 (and 3.5.0 alpha) all versions 3.2.0 onwards

### Operating system
Windows 10 - but assume all OS

### Additional context
Originally reported on the Forum:
https://forum.audacityteam.org/t/no-way-to-reset-meter-toolbar-mp-indicator/90446

I also fell foul of this myself recently when setting levels for an important impending recording I was planning to make from the radio.

petersampsonaudacity:
Note that the Manual (3.2.0) still says you can left click to reset the meters.  I have corrected that in the 3.4.3/3.5.0 alpha Manual.
![image](https://github.com/audacity/audacity/assets/50205984/ca4f20b5-2ff2-450d-a84b-9f2f18fa0637)

unhandyandy:
I also find this regression irritating.

Eniki:
I also strongly support getting this regression fixed.

The-Real-Rustla:
[...]

PR #5861, opened 2024-01-12
Per track meters

Paul-Licameli:

Depends on
- #6442
- #5830 

Recommended:


petersampsonaudacity:
@Paul-Licameli 

I just tested on W10 your branch:  audacity-win-3.5.0-alpha-20240112+5d1bafd-x64

The per track meters seem to work propery.

One thing to note: just like the main meters these per track meters can on;y be reset by stopping and restarting:

See: https://github.com/audacity/audacity/issues/5658
 **No longer possible to reset the Maximum Peak and Clipping indicators in the meters #5658**

petersampsonaudacity:
@Paul-Licameli 

The per track meters do not work while recording@
image

petersampsonaudacity:
Testing on W10 with @Paul-Licameli 's latest branch:

1) The per track meters do not work while recording still.

2) When working with just a single track (which is a use case for many of Audacity's users, myself included) then the per=track meter is just duplicated clutter.  Which begs the question: is there, or will there be, an option a preference maybe, to turnn them off and hide them?
image

Issue #5848, opened 2024-01-08
Crossfade Clips fails when Audacity is set to Deutsch (German)


petersampsonaudacity:
### Bug description
Crossfade Clips fails, with an incorrect error message, when Audacity is set to Deutsch (German).

The user gets an error message that they have more than two clips selected which is _**NOT**_ the case

Works fine when Audacity is in English

### Steps to reproduce
1. Ensure Audacity is set to English
2. Generate a 30 second chirp
3. move cursor to 10 secs and Ctrl+I to create a clip
4. select from 5 to 15 seconds
5. Effect > Fading Crossfade Clips
6. Observe: the crossfade works
7. Undo
8. Set Audacity to Deutsch in Interface prefs
9. select from 5 to 15 seconds (should have remained selected anyway)
10. Effekt > Überblendungen > Clips überblenden
11. **Observe:** the crossfade FAILS - with an error messag telling the user (incorrectly) that they have more than two clips selected.

Reset Audacity to English and it works fine again.

### Expected behavior
It should work properly in German as it does in English:

**Before**
![image](https://github.com/audacity/audacity/assets/50205984/50c58bce-c7b2-490e-b2ef-ef1c95cfe022)

**After**
![image](https://github.com/audacity/audacity/assets/50205984/bdfeb963-181a-47e9-890a-231a8cbf1468)

### Actual behavior
**Auf Deutsch:**
![image](https://github.com/audacity/audacity/assets/50205984/ff3ae521-e2b0-4c7b-b4a3-8ccd337a3417)

And clearly there are ONLY two clips in the project anyway

### Audacity Version
3.4.2 and latest 3.5.0 alpha

### Operating system
Windows 10 - but assume all OS

### Additional context
Works properly if Audacity is set to Français (French)

I do not know yet if this is a regression or not.

_Originally reported to me offline by email from a German user (Heiner Meyer) whom I have been helping._

petersampsonaudacity:
I have asked Heiner to try the workaround of temporarily setting Audacity to work in English on his Deutsch local machine and see if the command works for him in English.

petersampsonaudacity:
Testing shows that this is a regression on 3.2.5 introduced in 3.3.0

SteveDaulton:
Looks like a translation issue with this line of the plug-in:

`(setf err2 (format nil (_ "Error.~%Invalid selection.~%Empty space at start/ end of the selection.")))`

LWinterberg:
the translation reads: 
```
[...]

PR #5865, opened 2024-01-15
some strings!

LWinterberg:
* Resolves: #5848 
* Removes ancient versioning system remnants (arch)
* Rename "close track" to "delete track" (the tooltip for hovering over the x on a track)
* update `man audacity` from 1.3.5 (probably will only go into effect on linux repositories?)
* Move K.Soze to contributors


petersampsonaudacity:
Testing on W10 with @LWinterberg 's branch:  audacity-win-3.5.0-alpha-20240115+de3c348-x64

1) #5848 now works properly in Leo's branch build when the Audacity language is set to Deutsch.

2) And the hovertext tooltip when hovering over the X in the TCP now says **Delete Track (Ctrl+C)**

3)  **K. Soze** is moved to **Contributors**

saintmatthieu (approved):

PR #5866, opened 2024-01-16
Track Panel and Waveform painting

vsverchinsky:
Resolves: #5424, #5425 

Recommended:


libraries/lib-graphics/CMakeLists.txt
```diff
@@ -5,6 +5,12 @@ A library responsible for custom rendering in Audacity
 set( SOURCES
    FrameStatistics.cpp
    FrameStatistics.h
+
+   graphics/Color.cpp
+   graphics/Color.h
```
crsib, line 10: I would like to revisit at some point and check, which types are actually used. As Paul mentioned, we need some fixes and unit tests for Rect classes.

crsib: Judging by the next commit we need Color, Rect, Point and Size, but we don't need most of the functions except for basic math

libraries/lib-wave-track/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ set( SOURCES
    Sequence.h
    WaveClip.cpp
    WaveClip.h
+   WaveformScale.cpp
```
crsib, line 26: Should we just merge the master in Qt first?

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1905,6 +1906,7 @@ std::shared_ptr<WaveClip> WaveTrack::RemoveAndReturnClip(WaveClip* clip)
    if (it != mClips.end()) {
       auto result = std::move(*it); // Array stops owning the clip, before we shrink it
       mClips.erase(it);
+      Observer::Publisher<WaveTrackClipEvent>::Publish({});
```
crsib, line 1909: Probably this should be wrapped into some kind `NotifyClipsChanged` function or whatnot

PR #5867, opened 2024-01-16
5794 tempo detection speed improvement

### LINKED ISSUE #5794: Tempo detection speed improvement
The analysis takes time. The feature would be less irritating if it were much faster.
Currently, it looks like the vast majority of the time is spent on the time-frequency analysis. The time-frequency precision used is that of a Phase Vocoder, but for our purpose something much more rough is probably sufficient.
Ideas:
1. Downsample the signal before analysis. With a bit of luck this could make the detection twice as fast with no or negligible loss of quality.
2. Use a filter bank instead of an STFT to have control over the frequency resolution. Currently, the STFT settings are such that the frequency resolution is of 40Hz. For our purpose this is probably far too much.
3. Find approximations for the log and sqrt methods maybe would help ?

saintmatthieu:
Resolves: #5794 

This PR intends to speed up the tempo detection **without modifying its quality**. Guarantee that quality is not modified is given by the benchmarking, re-run after each optimization and checking that none of the dataset's 140 files was evaluated differently.

Processing time measurement was added to the benchmarking. On a Windows laptop, the algorithm is about 4 times faster after these optimizations, and on a Macbook, about 3 times. Commit-by-commit improvement is detailed in the commit messages.

Recommended:

QA:
That the quality did not change noticeably was verified with the benchmarking tests. Please nevertheless smoke-test the build, importing a few rhythmic and non-rhythmic files, preferably that you've already used before. If you have long files with high sample rate that have brought the progress bar in focus before this change, please try them again now: the speed should be noticeably faster.


cmake-proxies/CMakeLists.txt
```diff
@@ -18,3 +18,4 @@ endif ()
 addlib( libsbsms           sbsms       SBSMS       NO    YES   "sbsms >= 2.2.0" )
 addlib( soundtouch         soundtouch  SOUNDTOUCH  NO    YES   "soundtouch >= 1.7.1" )
 addlib( twolame            twolame     LIBTWOLAME  NO    YES   "twolame >= 0.3.13" )
+addlib( pffft              pffft       PFFFT       YES   NO    "" )
```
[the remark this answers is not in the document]
crsib: I would put pffft in `lib-math/3dpart` and build it along with lib-math. Similar to what we do with the staffpad library now. I don't think that this library is available in any distro.

libraries/lib-music-information-retrieval/DecimatingMirAudioReader.h
```diff
@@ -0,0 +1,44 @@
...
+ * an issue. In fact, for onset detection, it may even be beneficial, since it
+ * preserves a trace of the highest frequency components by folding them down
+ * below the nyquist. Thus we can decimate the audio signal to a certain extent.
+ * This is fast and easy to implement, meanwhile reducing dramatically the
+ * amount of data and operations.
+ */
+class DecimatingMirAudioReader : public MirAudioReader
+{
+public:
+   DecimatingMirAudioReader(const MirAudioReader& reader);
```
Paul-Licameli, line 32: `explicit` maybe

Paul-Licameli: This is an example of a decorator-pattern class

libraries/lib-music-information-retrieval/MirDsp.cpp
```diff
@@ -25,6 +25,21 @@ namespace MIR
...
+   union
+   {
+      float val;
+      int32_t x;
+   } u = { x };
+   auto log_2 = (float)(((u.x >> 23) & 255) - 128);
+   u.x &= ~(255 << 23);
+   u.x += 127 << 23;
+   log_2 += ((-0.3358287811f) * u.val + 2.0f) * u.val - 0.65871759316667f;
+   return log_2;
```
Paul-Licameli, line 40: I wrote elsewhere how I almost understand where the magic numbers come from, and how the method might be generalized to higher or lower degree polynomial approximations.  We might experiment with the tradeoff of computation speed and accuracy of the log2 approximation.  But then the greatest gain is likely just from avoiding the transcendental function.

libraries/lib-music-information-retrieval/PowerSpectrumGetter.h
```diff
@@ -0,0 +1,45 @@
...
+/*!
+ * @brief Much faster that FFT.h's `PowerSpectrum`, at least in Short-Time
+ * Fourier Transform-like situations, where many power spectra of the same size
+ * are needed. Currently only power spectrum, but may be generalized to other
+ * uses.
+ */
+class PowerSpectrumGetter
+{
+public:
+   PowerSpectrumGetter(int fftSize);
```
Paul-Licameli, line 28: `explicit` ?

libraries/lib-music-information-retrieval/PowerSpectrumGetter.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+
+PowerSpectrumGetter::~PowerSpectrumGetter()
+{
+   pffft_destroy_setup(mSetup);
+}
+
+void PowerSpectrumGetter::operator()(float* buffer, float* output)
+{
+   pffft_transform_ordered(mSetup, buffer, buffer, mWork.data(), PFFFT_FORWARD);
```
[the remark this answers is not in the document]
Paul-Licameli: "This is
     similar to calling pffft_transform and then pffft_zreorder."

Meaning `pffft_transform_ordered`.

Comments for `pffft_zreorder`:

"input and output should not alias."

But you violate that.

This makes me doubt correctness.

libraries/lib-music-information-retrieval/MirDsp.cpp
```diff
@@ -74,13 +75,29 @@ std::vector<float> GetNormalizedCircularAutocorr(std::vector<float> x)
...
+   x[1] *= x[1];
+   for (auto n = 2; n < N; n += 2)
+   {
+      x[n] = x[n] * x[n] + x[n + 1] * x[n + 1];
+      x[n + 1] = 0.f;
+   }
+
+   pffft_transform_ordered(
+      setup, x.data(), x.data(), work.data(), PFFFT_BACKWARD);
+   pffft_destroy_setup(setup);
```
Paul-Licameli, line 110: I might put this line in a Finally

libraries/lib-music-information-retrieval/MirDsp.cpp
```diff
@@ -74,13 +75,29 @@ std::vector<float> GetNormalizedCircularAutocorr(std::vector<float> x)
...
-   PowerSpectrum(N, x.data(), x.data());
-   // We need the entire power spectrum for the auto-correlation, not only the
-   // left-hand side.
-   std::copy(x.begin() + 1, x.begin() + N / 2 - 1, x.rbegin());
-   InverseRealFFT(N, x.data(), nullptr, x.data());
-   // For efficiency, only keep the positive half of this symmetric signal.
+   PFFFT_Setup* setup = pffft_new_setup(N, PFFFT_REAL);
+   std::vector<float> work(N);
+   pffft_transform_ordered(
+      setup, x.data(), x.data(), work.data(), PFFFT_FORWARD);
```
Paul-Licameli, line 96: Again, aliasing of input and output

libraries/lib-music-information-retrieval/MirDsp.cpp
```diff
@@ -74,13 +75,29 @@ std::vector<float> GetNormalizedCircularAutocorr(std::vector<float> x)
       return x;
    const auto N = x.size();
    assert(IsPowOfTwo(N));
-   PowerSpectrum(N, x.data(), x.data());
-   // We need the entire power spectrum for the auto-correlation, not only the
-   // left-hand side.
-   std::copy(x.begin() + 1, x.begin() + N / 2 - 1, x.rbegin());
```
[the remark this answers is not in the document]
Paul-Licameli: The bin one before Nyquist was not mirrored!

Paul-Licameli, line 80: I notice now that the old version had an off-by-one mistake that escaped earlier review!  The bin one before Nyquist was not correctly mirrored.

Paul-Licameli:
My review is complete.  Some comments could be improved.  I have doubts about correctness, because of aliasing of input and output buffers, contrary to what the commens in pffft.h recommend.

libraries/lib-music-information-retrieval/MirDsp.cpp
saintmatthieu: Indeed ...

libraries/lib-music-information-retrieval/PowerSpectrumGetter.cpp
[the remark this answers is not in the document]
saintmatthieu: > Did you try this? Does it improve performance?

No, but `mWork` is allocated only once for the entire STFT, so I doubt it has noticeable impact. Also, my optimizations were guided by visual studio's profiler, and this also told me that STFT wasn't the performance bottleneck after these optimizations anymore.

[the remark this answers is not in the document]
saintmatthieu: > This makes me doubt correctness.

I don't quite understand the logical order of what gets you to this conclusion.

In any case, did you also see that `pffft_transform_ordered`'s comment expressly says "input and output may alias." ?

libraries/lib-music-information-retrieval/MirDsp.cpp
saintmatthieu: I think so too. I must admit I have understood little of your demonstration, especially the bit-shift part, maybe you can explain that to me more at length in an upcoming conversation. Besides me not understanding it looked correct, judging by your results and also some I ran on my side.

cmake-proxies/CMakeLists.txt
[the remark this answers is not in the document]
saintmatthieu: That's similar to my first approach, but I thought, yet another copy of PFFFT maybe we could spare us. Or did you mean have StaffPad depend on `lib-math` ?

saintmatthieu:
@Paul-Licameli comments addressed.

libraries/lib-music-information-retrieval/PowerSpectrumGetter.cpp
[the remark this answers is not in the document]
Paul-Licameli: You are right.  The warning against aliasing applies only to `pffft_zreorder` called directly.

Paul-Licameli (approved):
My doubts about correctness were answered.

I approve.

I also leave this suggestion:  use `pffft_transform` not `pffft_transform_ordered` and see if there is some more measurable improvement.

You can do that because none of the calculations of the novelty depend on bin number.  Comparing corresponding binds, however permuted, you will compute the same thing (up to small non-associativities in floating point addition).

I believe the un-reordered output will still put DC at subscript 0 and Nyquist at 1.

saintmatthieu:
> cript 0 and Nyquist at 1.

That's an interesting idea.
The input to the novelty measure is power, though, not the complex output of PFFFT. I have to know where the real and imaginary parts of a given bin are to combine them.

Paul-Licameli:
> > cript 0 and Nyquist at 1.
> 
> That's an interesting idea. The input to the novelty measure is power, though, not the complex output of PFFFT. I have to know where the real and imaginary parts of a given bin are to combine them.

I believe you really get a sequence of complex numbers, except that the first two floats are two real numbers.  So you can really add the squares of successive float pairs after subscript 1.

Paul-Licameli:
how FFT delivers permuted answers: https://en.wikipedia.org/wiki/Bit-reversal_permutation

That's as applied to bin numbers 1, ..., N / 2 - 1 but as complete complex numbers.

saintmatthieu:
Aha, you think the internal layout is something standard? I guess you're right, but the PFFFT API doesn't give an indication of it. Maybe someone exchanges the PFFFT library some day and it changes?
Also, after these optimizations, the STFT isn't the bottleneck anymore (at least on Windows), so further improvements here would be marginal.

Issue #5771, opened 2023-12-11
Compilation error: Apparent missing file


evertvorster:
### Bug description

When compiling the latest master of Audacity as of 2023-12-11, a compilation error pops up:
```
[ 13%] Linking CXX shared library ../../Release/lib/audacity/lib-utility.so
/mnt/int/system/aur/audacity-local-git/src/audacity/libraries/lib-music-information-retrieval/tests/MusicInformationRetrievalTests.cpp:2:10: fatal error: catch2/catch.hpp: No such file or directory
    2 | #include <catch2/catch.hpp>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

It would appear that catch.hpp is missing, according to the error message. 

### Steps to reproduce

Compile Audacity from git master.

### Expected behavior

A clean compile.

### Actual behavior

The error described in the description.

### Audacity Version

current master / alpha / nightly build

### Operating system

Linux

### Additional context

About my system:
```
Operating System: Arch Linux 
KDE Plasma Version: 5.27.10
KDE Frameworks Version: 5.112.0
Qt Version: 5.15.11
Kernel Version: 6.6.6-arch1-1 (64-bit)
Graphics Platform: Wayland
Processors: 32 × 13th Gen Intel® Core™ i9-13900HX
Memory: 62.5 GiB of RAM
Graphics Processor: Mesa Intel® Graphics
Manufacturer: TUXEDO
Product Name: PD5x_7xSNC_SND_SNE
System Version: Not Applicable
```

Version of catch2 installed: 3.4.0, from Arch binary repos, second release.

crsib:
Arch build is quite broken right now, I had no time to check so far

evertvorster:
Just checked, and Audacity from the latest master still does not build.

evertvorster:
I am still unable to compile the sources from current master on Arch Linux. Is this related to something missing in Arch? Is anyone able to build current master on other flavors of Linux?

crsib:
Arch support is still not fixed. We build Linux on CI continuously, AFAIK wx 3.2 build is broken.

Unfortunately I have no resources right now to fix the Arch build and I am alone working on that.

Generally I wouldn't recommend using master ATM, it's rather unstable

evertvorster:
Building on Arch just follows the build instructions listed on the Audacity site. 

Would it help if I can pinpoint the commit that broke compilation?

evertvorster:
[...]

PR #5870, opened 2024-01-17
allow building the tests without using conan

christian-heusel:
Resolves: https://github.com/audacity/audacity/discussions/5841

This change allows to build the tests even though conan is not used.

- [ ] ~~If changes are extensive, then there is a sequence of easily reviewable commits~~

Recommended:

---

Related to https://github.com/audacity/audacity/issues/5771
cc @matoro since they initially wrote the patch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,15 +228,12 @@ cmd_option( ${_OPT}has_url_schemes_support
    "Build custom URL schemes support into Audacity"
    Off)
 
-include( CMakeDependentOption )
-
-cmake_dependent_option(
+cmd_option(
    ${_OPT}has_tests
    "Enables automated testing support"
-   On
-   "${_OPT}conan_enabled"
-   Off
-)
+   On)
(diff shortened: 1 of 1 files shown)


crsib:
I expect this pull request to break the `Build Audacity Linux Packages` job. It is not run for PRs, only for branches, so this may be overlooked here.

matoro:
> I expect this pull request to break the `Build Audacity Linux Packages` job. It is not run for PRs, only for branches, so this may be overlooked here.

This should be fixed by explicitly adding `-Daudacity_conan_enabled=on` to the job definition right?

crsib:
> This should be fixed by explicitly adding -Daudacity_conan_enabled=on to the job definition right?

`-Daudacity_conan_enabled=off` would be a fix, but not in the job, rather in the scripts it invokes

LWinterberg:
Audacity 4 won't use conan, so there no longer is a need for this

Issue #5862, opened 2024-01-12
WaveTrack::LinkConsistencyFix modifies the AUP3 state, but doesn't force the project table to update


crsib:
### Bug description

This issue was observed with a project reported corrupted here: https://forum.audacityteam.org/t/corrupt-or-otherwise-broken-audacity-project-recovery/64162/126

As it appears, in the case of this project `WaveTrack::LinkConsistencyFix` removes some zero-length clips it effectively removes the corresponding blocks from the `sampleblocks` table. However, the project is not marked as modified, nor is it (auto)saved. The project will get corrupted after Audacity is closed without the manual saving of the project.

This happens because Audacity checks for missing blocks before it checks for empty clips.

### Steps to reproduce

1. Make yourself a project with zero-length clip
2. Open it with 3.4.2 or latest master
3. Close Audacity
4. Observe that project can no longer be loaded. 

### Expected behavior

Project is not corrupted

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

DanPinGF:
I've seen that the bug affects older projects regardless of length, and it's so bad it needs immediate fixing. Many of my Audacity projects are kept in case they need editing.

Paul-Licameli:
I still have not found exact enough steps to reproduce, but after some examination of the code, and over-complicated fixes, I found a surprisingly simple alternative that may work.

chinakov:
Tested. 

Managed to reproduce the issue in question on 3.4.2, and verified that the issue has been fixed.

Thank you.

Paul-Licameli:
I request that QA re-test this in recent master, because of some unusual moves involved in merging it with wave-clip-refactoring.

I doubt have serious doubts, but the implementation of this fix was disturbed again a little.

dozzzzer:
@chinakov ^

chinakov:
Will do!

DanPinGF:
'Fixed'? I don't see anything fixed in 3.4.2, as this is the latest stable build. The bug really concerns me because whenever I open an older Audacity project, it becomes corrupt without manually saving it and I have to permanently delete it, losing access to the project for good.

LWinterberg:
@DanPinGF Fixes from last week don't magically appear in a version released 3 months ago, indeed. It is fixed (pending verification from QA) in the current master (aka nightly or alpha), and will be present in the next release.

chinakov:
Re-tested, behaviour has not changed. Thanks!

PR #5873, opened 2024-01-18
Bump reference counts to blocks of newly loaded file earlier

Paul-Licameli:
Resolves: #5862

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -800,13 +800,14 @@ WaveTrack::WaveTrack( const SampleBlockFactoryPtr &pFactory,
    DoSetRate(static_cast<int>(rate));
 }
 
-WaveTrack::WaveTrack(const WaveTrack &orig, ProtectedCreationArg &&a)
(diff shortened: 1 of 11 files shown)


crsib (dismissed):
That's really elegant 😻

libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2359,17 +2359,12 @@ bool WaveTrack::FormatConsistencyCheck() const
       });
 }
 
-bool WaveTrack::InsertClip(WaveClipHolder clip)
+void WaveTrack::InsertClip(WaveClipHolder clip)
 {
-   if(!clip->GetIsPlaceholder() && clip->IsEmpty())
```
vsverchinsky, line 2364: I think that the explanation of this line presense is that we've found out that zero clips could have been created during editing the project.

Paul-Licameli:
Please review it again.  The second commit is very different now and makes more conservative changes.

vsverchinsky (approved):

Paul-Licameli:
Fix the corrupting of projects that were saved with empty clips with smart data

Issue #5427, opened 2023-10-25
Project Cloud Sharing


crsib:

chinakov:
Tested on macOS, Windows 11, Linux 22.04 Ubuntu, no issues noticed.

PR #5885, opened 2024-01-23
Make ExportMP3 configurable

crsib:
Part of #5427 

Allows MP3 exporter to be configured using JSON

Recommended:


crsib:
I would say that this PR is untestable at this point

vsverchinsky (approved):

PR #5887, opened 2024-01-24
Adds ability to run code after the app initialization

crsib:
Part of #5427

This PR is untestable

Recommended:

### DIFF
--- a/libraries/lib-utility/AppEvents.h
+++ b/libraries/lib-utility/AppEvents.h
@@ -1,0 +1,28 @@
+/*
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ * SPDX-FileName: AppEvents.h
+ * SPDX-FileContributor: Dmitry Vedenko
+ */
+
+#pragma once
+
+#include <functional>
+
+namespace AppEvents
+{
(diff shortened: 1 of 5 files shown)


libraries/lib-utility/AppEvents.h
```diff
@@ -0,0 +1,28 @@
...
+namespace AppEvents
+{
+/*! Register callback to be called when application is initialized.
+ *  If application is already initialized, callback will be called immediately.
+ *  @param callback Callback to be called when application is initialized.
+ *  @pre `!!calback`
+ */
+UTILITY_API void OnAppInitialized(std::function<void()> callback);
+
+class UTILITY_API AppEventsProvider /* not final */
```
vsverchinsky, line 20: My guess is that `AppEventsProvider` exists to protect parts of events API from unwanted use? If so I'd suggest also to make a statement to the class comment that the number of instances should be limited to one (assertion in ctor wouldn't harm).
`AppEvents::AppEventsProviders` could be shortened to `AppEvents::Provider`, sorry :)

vsverchinsky (approved):

vsverchinsky (approved):

Issue #5884, opened 2024-01-23
Export dialog exceeds monitor width


dozzzzer:
### Bug description

Originally posted here: https://forum.audacityteam.org/t/export-dialog-exceeds-monitor-width/96531

### Steps to reproduce

1. Grab some audio, then Ctrl + Shift + E
2. Select "Format: (external program)"
3. Type in a long string into "Command" text box
4. Proceed to complete export
5. Navigate to the Export Audio dialog again
6. Observe the width of the dialog

### Expected behavior

The maximum width of the Export Audio dialog should be limited by the width of the screen or remain fixed at all times.

### Actual behavior

The Export Audio width depends on the length of the previously entered string into the Command text box which may extend the dialog beyond the screen edge:

<img width="450" alt="image" src="https://github.com/audacity/audacity/assets/70063970/c69776d6-a4b6-4f3c-bb7e-2122e3f61ea1">

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

Issue #5515, opened 2023-11-06
Open the Export Audio dialog in center of screen


GermanKiwi:
### Your idea

On Windows, the new Export Audio dialog, introduced in 3.4.0, always opens near the top-left corner of the screen.

If you move it to another location, then close it and reopen it, it opens again near the top-left corner.

This is annoying when you're using a very large screen (eg. 32").

### Problem to be solved

Please either:

A) Have this dialog always open in the center of the screen, just as the Preferences dialog does;

or

B) Have this dialog remember its last location, so it always reopens wherever the user last positioned it.

petersampsonaudacity:
Testing on W10 with 3.4.0 and the latest Beta build I have for the upcoming 3.4.1

I cannot reproduces this - the export dialog opens over my Audacity window:
![image](https://github.com/audacity/audacity/assets/50205984/701438d2-9ab2-4da1-9ca8-054a9b9daf63)

dozzzzer:
Tested

PR #5889, opened 2024-01-24
export dialog: center and max width

LWinterberg:
Resolves: #5515
Resolves: #5884

hey look it's the same PR again but for a different dialog and limiting the exciting direction of X this time instead of Y. 

It's intentional that the thing is not centered for the updates; the window should just extend downwards instead of shifting position to stay centered.

### DIFF
--- a/modules/mod-cl/ExportCL.cpp
+++ b/modules/mod-cl/ExportCL.cpp
@@ -217,6 +217,7 @@ public:
                   mLastCommand = event.GetString();
                });
                mLastCommand = mCommandBox->GetValue();
+               mCommandBox->SetMaxSize(wxSize(50,400));
 
                S.AddButton(XXO("Browse..."), wxALIGN_CENTER_VERTICAL)
                   ->Bind(wxEVT_BUTTON, &ExportOptionsCLEditor::OnBrowse, this);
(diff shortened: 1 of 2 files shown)


LWinterberg:
Actually, looking at it in practice I think I just want to set a simple max size for this thing and not allow it to stretch all across the screen.

vsverchinsky:
> Actually, looking at it in practice I think I just want to set a simple max size for this thing and not allow it to stretch all across the screen.

@LWinterberg 
Export editor UI is also present in Timer Record dialog.
In that particular case limiting the maximum width of the dialog may not look good with some translations, I think. If so, maybe the constraints could be set on `wxComboBox` that is created in `ExportOptionsCLEditor`?

vsverchinsky (approved):

GermanKiwi:
Hi @LWinterberg, does this patch mean the Export dialog will now always be centered in the middle of the Audacity window - just as the Preferences dialog already is?

The reason this is important for me, is because I always keep Audacity maximised on my screen, and I have a large 32-inch 4K monitor, and it's harder to interact with the Export dialog when it always opens up in the top-left corner of the screen, rather than in the center, directly in front of my eyes. :)

LWinterberg:
> Hi @LWinterberg, does this patch mean the Export dialog will now always be centered in the middle of the Audacity window - just as the Preferences dialog already is?

That indeed would be the case yes.

Let me know if you see other dialogs which don't open up in the center - the fix for these is quite easy.

Issue #5894, opened 2024-01-25
Timeline options button improvements


LWinterberg:
in scope for 3.5:

- [x] The timeline options button currently assumes the appearance of the playhead and has 4 states. It really only needs 1. 
- [x] The timeline options button is hard to recognize as an options button and needs a different icon.
- [x] The "pin" metaphor for the pinned playhead is not terribly helpful. Let's remove it.
- [x] Some options inside the timeline options button are hard to understand ("update display while playing"), so some strings can be clarified

nice-to-haves eventually: 
- (code: There seems to be a large amount of scaffolding to allow for multiple buttons in there, which is unused. This may be something which can be cleaned up)

petersampsonaudacity:
Note that the Timeline options can also be accessed from the context menu you get from right-clicking anywhere in the Timeline (including right-clicking on the Options button) .

And this is the same menu, so maybe the options button isn't need at all and the user can just rely on context menu right-click - a gesture that is available in many other places in Audacity (and other apps) so users should be fairly used to that.

petersampsonaudacity:
I seem to recall that in much older versions of Audacity that green button in the Time was _**just**_ used to pin/unpin the playhead (which is why it looks like a playhead icon)

petersampsonaudacity:
I have added P1s in the manual to track this in the documentation

petersampsonaudacity:
I have updated the Manual for this change  - with the exception of one page that I'll tackle over the weekend

PR #5895, opened 2024-01-25
Timeline options button improvements

LWinterberg:
Resolves: #5894


src/AdornedRulerPanel.cpp
```diff
@@ -1413,44 +1413,16 @@ void AdornedRulerPanel::ReCreateButtons()
...
       bmpRecoloredUpSmall, bmpRecoloredDownSmall, 
       bmpRecoloredUpHiliteSmall, bmpRecoloredHiliteSmall, 
-      //bmpUnpinnedPlayHead, bmpUnpinnedPlayHead, bmpUnpinnedPlayHead,
       bmpPlayPointer, bmpPlayPointer, bmpPlayPointer,
-      size);
+      OnTogglePinnedStateID, position, true, size
+   );
+
+   position.x += size.GetWidth();
+   mButtons[iButton++] = button;
```
Paul-Licameli, line 1425: Commit #2 eliminates the pin images.

Great!  You know I implemented the needed drawing code for the pinned play years ago, but -- the silly pin button image was NOT my idea and I never liked it.  Let it just be a triangle.

However -- what you also lose here, is the alternation of the color of the triangle between green for play, and red for record.

Are you sure you want that too?

src/tracks/ui/PlayIndicatorOverlay.cpp
```diff
@@ -62,10 +62,8 @@ namespace {
 std::pair< wxPoint, wxBitmap > GetIndicatorBitmap( AudacityProject &project,
    wxCoord xx, bool playing)
 {
-   bool pinned = Scrubber::Get( project ).IsTransportingPinned();
-   wxBitmap & bmp = theTheme.Bitmap( pinned ?
-      (playing ? bmpPlayPointerPinned : bmpRecordPointerPinned) :
-      (playing ? bmpPlayPointer : bmpRecordPointer)
+   wxBitmap & bmp = theTheme.Bitmap(
+      playing ? bmpPlayPointer : bmpRecordPointer
```
Paul-Licameli, line 66: I see commit number 3 still colors the triangle on the timeline either red or green, even if the button is always green.

src/AdornedRulerPanel.cpp
```diff
@@ -1417,7 +1417,7 @@ void AdornedRulerPanel::ReCreateButtons()
       this,
       bmpRecoloredUpSmall, bmpRecoloredDownSmall, 
       bmpRecoloredUpHiliteSmall, bmpRecoloredHiliteSmall, 
-      bmpPlayPointer, bmpPlayPointer, bmpPlayPointer,
+      bmpCogwheel, bmpCogwheel, bmpCogwheel,
```
Paul-Licameli, line 1420: Now, I have reviewed to the end, and I understand the button is an unchanging gearwheel.

OK, I have no objections to anything.

Paul-Licameli (approved):
I never liked the stupid pin image.  Bye!

PR #5899, opened 2024-01-28
Fivefold faster tempo detection and spectrograms too

Paul-Licameli:

You read that right, fivefold.

At least ... on one computer, with the averages of a few before-and-after trials, with 6 cores and hyperthreading.

Results may vary with hardware but have great promise to scale as the processing power gets cheaper.

In all the recent performance improvements of tempo detection performance, the "obvious" thing wasn't done --
which is, to use multiple cores to exploit the "embarrassing" parallelism inherent in the algorithm.

This PR adds some template classes and functions in new lib-concurrency -- the hard part -- to make transformations of
such data-parallel computations comparatively easy to do.

Tempo detection is the first application -- unit tests pass, when `runLocally = true` in MirTestUtils.cpp, and `timeMeasurement.txt` then contains much smaller numbers.

The computation of spectrograms for display is the example of a second, even easier application, in which one simply 
supplies an iterator range and a (copyable) lambda, and the speedup follows.  Try using the largest spectrogram window 
size (327678) and compare before-and-after zooming-in and -out, scrolling, and playing with the pinned head.  (But sorry, 
no parallelism yet if you use the Reassignment algorithm.)

What are other possible easy applications?  I think there are some in Sequence.cpp such as converting sample format, and
finding min/max/rms.  Others?

Recommended:


libraries/lib-concurrency/ParallelProgress.h
```diff
@@ -0,0 +1,55 @@
...
+      if (abandoned.load(std::memory_order_relaxed))
+         throw detail::AbortException{};
+      const auto old_total = total.fetch_add(inc, std::memory_order_acq_rel);
+      return old_total + inc;
+   }
+
+   //! Cause shut-down of tasks
+   void abandon() { abandoned.store(true, std::memory_order_relaxed); }
+private:
+   std::atomic<increment_type> total{ Increment{} };
```
crsib, line 49: This, however, introduces **false sharing**, which is a problem in this case. `abandoned` is mostly read, and rarely modified (never, ideally). `total` is modified every call to `operator ()`, effectively invalidating `abandoned` for every other CPU.

libraries/lib-concurrency/ParallelProgress.h
```diff
@@ -0,0 +1,55 @@
...
+
+//! A callback supplied to run methods of a task so it can "yield" while it
+//! reports progress, also detecting total progress
+/*!
+ Throws an exception when a thread team is shutting down.  Do not handle this
+ exception.
+
+ @tparam Increment an Integral type
+ */
+template<typename Increment> class Progress {
```
crsib, line 27: This looks like a cancellation pattern but with progress tracking. I don't think that the progress tracking is accurate or needed here.

At the same time, I think that using an exception and, specifically, an exception from the `details` namespace reduces the scope where this class is useful.

libraries/lib-concurrency/ParallelProgress.h
```diff
@@ -0,0 +1,55 @@
...
+ 
+ Paul Licameli
+ 
+ **********************************************************************/
+#ifndef __AUDACITY_PARALLEL_PROGRESS__
+#define __AUDACITY_PARALLEL_PROGRESS__
+
+#include <atomic>
+
+namespace Parallel {
```
crsib, line 15: What is the justification of the capital P here?

libraries/lib-concurrency/ParallelProgress.h
```diff
@@ -0,0 +1,55 @@
+/**********************************************************************
```
crsib, line 1: No license identifier. Previously, Audacity ignored it in most files, but that is not correct.

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+namespace Parallel {
+
+//! Compute the result of a task that admits data-parallel decomposition of
+//! a domain described by one iterator range
+/*!
+ The simple subdivision assumes that the time spent in Task::run is proportional
+ to the size of the iterator range.
+
+ Copies of Task for non-empty, disjoint sub-ranges may be created and each of
+ them run on its own thread.  The calling thread always runs the original task
```
crsib, line 35: Why? This limits the use of this class.

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+ exactly once for each task made by the split constructor (absent exceptions),
+ and only after `run` is called on `this` and on `task`
+
+ @tparam Increment defaults to the difference type of the Splittable iterators
+
+ @param task If `(std::distance(task.begin(), task.end()))` is zero or less, no
+ other threads are recruited and execution is serial.
+
+ @param n_threads maximum number of threads to use in the computation, including
+ the calling thread; supply 1 to do all work seriallyx
```
crsib, line 57: Typo in `seriallyx`

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+template<
+   typename Task,
+   typename Increment
+      = typename std::iterator_traits<typename Task::iterator_type>
+         ::difference_type
+>
+void OneDimensionalReduce(Task &task,
+   unsigned n_threads = std::thread::hardware_concurrency());
+
+enum ReportChoice{ NoReport, Report };
```
crsib, line 78: Good!

libraries/lib-concurrency/ParallelProgress.h
```diff
@@ -0,0 +1,55 @@
...
+
+   //! Report progress; poll total progress; possibly non-returning
+   /*!
+    This may throw an exception internally to stop a thread when it has been
+    detected that other threads in the team have stopped.
+
+    TODO C++20: make Progress an awaitable that can suspend instead of throwing,
+    and let tasks be coroutines
+    */
+   increment_type operator()(increment_type inc = {}) {
```
crsib, line 39: What is the justification for making progress callable? I would not object if the users below were allowed to receive a callable instead of progress. However, Progress is used only to stop the workers if the exception occurs.

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+   {
+      std::for_each(this->begin(), this->end(), [this, &progress](auto &&value){
+         if constexpr (doReport) {
+            auto total =
+               progress((*this)(std::forward<decltype(value)>(value)));
+            if (is_first)
+               this->report_progress(total);
+         }
+         else {
+            (*this)(std::forward<decltype(value)>(value));
```
crsib, line 113: I find such code to be very poorly readable.

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+ForEach(Iterator, Iterator, Callable) -> ForEach<Iterator, Callable>;
+
+template<typename Task, typename Increment>
+void OneDimensionalReduce(Task &task, unsigned n_threads)
+{
+   assert(n_threads >= 1);
+   // TODO: This function might be modified instead to allocate tasks to threads
+   // from a pool, and the Task class might define other policies for
+   // subdivision of its range
+   n_threads = std::max(1u, n_threads);
```
crsib, line 181: I like that the assertion is not the only check, but do we need the `n_threads > 0` at all?

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+      const auto remainder = count % n_threads;
+      Distance distance{},
+         fraction{};
+      for (size_t ii = 1; ii < n_threads; ++ii) {
+         // Don't evaluate (ii * count) / n_threads because there might be
+         // overflow
+         auto newDistance = distance + step;
+         fraction += remainder;
+         if (fraction >= n_threads)
+            fraction -= n_threads, ++newDistance;
```
crsib, line 204: I am very concerned about this code. It must be rewritten without the comma.

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+      auto pt = Packaged{ std::cref(runner) };
+      auto future = pt.get_future();
+      channels.emplace_back(move(pt), move(future));
+   }
+
+   // Now it's safe to run the new tasks concurrently, for all indices more than
+   // 0
+   size_t index = 0;
+   for (auto &task : tasks) {
+      std::thread{ ref(channels[index].first), std::ref(task), index + 1 }
```
crsib, line 296: Contrary to the previous place, I see very little reason to split this into multiple lines instead of using ++index right in place.

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+         .detach();
```
crsib, line 297: I can't say I like this, but I understand that this is safe.

libraries/lib-concurrency/ParallelReduce.h
```diff
@@ -0,0 +1,307 @@
...
+// Deduction guide
+template<typename Iterator, typename Callable>
+ForEach(Iterator, Iterator, Callable) -> ForEach<Iterator, Callable>;
+
+template<typename Task, typename Increment>
+void OneDimensionalReduce(Task &task, unsigned n_threads)
+{
+   assert(n_threads >= 1);
+   // TODO: This function might be modified instead to allocate tasks to threads
+   // from a pool, and the Task class might define other policies for
```
crsib, line 179: Not having thread pools limits the usability of this function quite significantly.

crsib (commented):
Judging by the code style you are pulling in some of your older work. 

I have only looked on `lib-concurrency`. I think there few changes needed, but they are rather minor.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -373,43 +374,40 @@ void SpecCache::Populate(
...
+            scratch = std::vector<float>(scratchSize)
+         ](int xx) mutable {
+            float* buffer = &scratch[0];
+            CalculateOneSpectrum(sampleCacheHolder,
+               settings, clip, xx, pixelsPerSecond, lowerBoundX, upperBoundX,
+               gainFactors, buffer, &freq[0]);
+         };
+         using namespace Parallel;
+         ForEach task{ first, last, calcOne };
+         OneDimensionalReduce(task);
```
crsib, line 391: If I understand correctly, this function may put CPU to 100% load during the playback. There must be more control over the thread priorities and count here.

Paul-Licameli:
> Judging by the code style you are pulling in some of your older work.
> 
> I have only looked on `lib-concurrency`. I think there few changes needed, but they are rather minor.

Two old files were simply moved into lib-concurrency.  The rest of the work is very recent.  But it abstracts out for reuse the pattern that was applied in this experiment a year ago. https://github.com/audacity/audacity/pull/4141

libraries/lib-concurrency/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+#[[
+Synchronization objects and parallel algorithms
+]]#
+
+set( SOURCES
+   MessageBuffer.h
```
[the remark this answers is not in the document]
Paul-Licameli: It was first written and used in scrubbing to communicate to the thread that fetches audio in response to mouse movements.

That file is just moved without change and does not need review.

libraries/lib-concurrency/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+   spinlock.h
```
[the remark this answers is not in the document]
Paul-Licameli: That file too is just moved unchanged and not used in the new files.

libraries/lib-concurrency/ParallelProgress.h
```diff
@@ -0,0 +1,55 @@
...
+      if (abandoned.load(std::memory_order_relaxed))
+         throw detail::AbortException{};
```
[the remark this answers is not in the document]
Paul-Licameli: The alternative is to complicate how tasks are written, requiring them to test and exit.

It is simpler instead to let them use a possibly non-returning function, which also, in the future C++20 version perhaps, would instead become a coroutine suspension point.  Then there could be the possibility of non-resumption or migration of a task to another thread in a more sophisticated task scheduler.

libraries/lib-concurrency/ParallelProgress.h
Paul-Licameli: Very good point.  And this is why I wrote the `NonInterfering` utility class template, which should more to lib-concurrency and be used here.

libraries/lib-concurrency/ParallelProgress.h
Paul-Licameli: None really.

We don't have very well defined conventions.

We have already namespace Observer

libraries/lib-concurrency/ParallelReduce.h
Paul-Licameli: At least when the supplied argument is 1, you know that the task just runs serially on the calling thread, so you can easily confirm first that the algorithm is serially correct before debugging it with multiple threads.

In the first application of this algorithm for tempo detection, the calling thread is the main thread and is the only one that may update the progress dialog.

libraries/lib-concurrency/ParallelReduce.h
Paul-Licameli: It's the unfortunate correct way to forward an argument perfectly.

libraries/lib-concurrency/ParallelReduce.h
Paul-Licameli: Wrong -- `index` is used twice in arguments of a constructor and order of evaluation is not specified.  I must increment only in another statement.

libraries/lib-concurrency/ParallelReduce.h
Paul-Licameli: Synchronizing on the get of a future of a packaged_task that runs on the thread should be equivalent to joining the thread.

libraries/lib-concurrency/ParallelProgress.h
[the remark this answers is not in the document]
crsib: > The alternative is to complicate how tasks are written

I can have a task where I need to handle complications explicitly. I do have such use cases now, but not using this class, obviously. 

I don't see how handling the result complicates anything.

libraries/lib-concurrency/ParallelReduce.h
Paul-Licameli: I didn't want to distract myself yet with writing a thread pool too.  I wanted to get to a demonstration of some real performance wins and also ease of use of the new utility in a few places.

Do we already have a usable thread pool in a third party library/

Paul-Licameli:
> > At least ... on one computer, with the averages of a few before-and-after trials, with 6 cores and hyperthreading.
> 
> You have 12 threads and just a 5x speedup. The good question is why, but I probably need to examine the rest of the PR.

Speedup won't be ideal.  There should be some serial component.  Or less than fully parallel, as in is the final merging step in which not all threads are doing work and there is inter-thread communication.

You pointed out some cache ping-pong from false sharing.

There is thread creation and destruction overhead because I didn't bother yet to invent or reuse a thread pool.

There might be contention for a shared resource like the main memory bus or I/O.

The estimated speedup was from @saintmatthieu 's regression test for tempo detection.  In fact that test pulls entire files into big memory buffers first, then used to go sequentially through the memory, but now that can be fetched in a random-access way.

The real-world tempo detection first creates a WaveClip object, and then this parallelized version through the `MirAudioReader` subclass will fetch sample blocks out of it, requiring database access.

Issue #5728, opened 2023-11-30
Availability of right-click "Change Speed" determined by mysterious logic


SteveDaulton:
### Bug description

The "Change Speed" option in an audio track's right click menu may be unexpectedly available / unavailable (greyed out /not greyed out).

### Steps to reproduce

1. Launch Audacity
2. Generate a 30 second tone
3. Click on an empty part of the tracks window
4. Generate a 10 second tone (creates a new, selected track)
5. Right click on the 10 second track
6. Observe that "Change Speed" is greyed out
7. Right click again on the track (the menu closes and the selection remains)
8. Right click again on the track
9. Observe that "Change Speed" is now available
10. Right click again on the track to close the menu
11. Right click on the 30 second track
12. Observe that "Change Speed" is available (not greyed out)
13. Select "Change Speed"
14. Observe that the menu closes without opening the Change Speed dialog window

### Expected behavior

Step 6: "Change Speed" available.
Step 12: "Change Speed" greyed out

### Actual behavior

Step 6: "Change Speed" greyed out.
Step 12: "Change Speed" available

### Audacity Version

3.4.2

### Operating system

All OSs

petersampsonaudacity:
Hmmm...

I notice that the image I have in the manual gas it grayed-out even though a selection is present:
![image](https://github.com/audacity/audacity/assets/50205984/f157132c-edb1-4c9f-b18a-54511bb2f299)

petersampsonaudacity:
> While we're at it, #3101 goes in the same vein - when a clip is right-clicked, it the relevant options for the clip could be accessible even if it's not selected

**Big** +1 from me

LWinterberg:
Hello! We just [switched development to Audacity 4](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444). We've redone, or plan on redoing, various things from scratch, including the entire UI, and the part this bug is about. As such, I'm closing this bug as it will be fixed there and we won't be fixing it for the final releases of Audacity 3.

PR #5900, opened 2024-01-29
5891 non destructive pitch shifting

### LINKED ISSUE #5863: Changing clip speed during playback possible with "change speed" menu
### Bug description

Changing the amount of time stretching of clips during playback is not supported and is supposed to be disabled. However, one can still do this with the change-speed menu

### Steps to reproduce

1. Create a clip
2. Start playback
3. Right-click the clip
4. select the "change speed" option

### Expected behavior

Changing the clip's speed should be disabled.

### Actual behavior

It is possible. The clip is visually stretched, but the audio continues as though it weren't.

### Audacity Version

other

### Operating system

Windows

### Additional context

Discovered on alpha, but it most likely is present in 3.4.2 already.

### LINKED ISSUE #5891: Non-destructive pitch shifting
Now that we've done the hard work integrating StaffPad's time-and-pitch library for non-destructive time stretching, non-destructive pitch shifting is a low-hanging fruit.

Figma design: https://www.figma.com/file/6kBToiSyv40MOWaNzkxCdj/32_WX---Pitch-%26-Onset-snapping?type=design&node-id=30-761&mode=design&t=NXQu7CHTFSEalCZ6-0

Not every single aspect of this design have to be implemented with this ticket, leftovers can be addressed separately. At least this ticket should make the feature usable.

saintmatthieu:
Resolves: #5891, #5863

Non-destructive pitch shifting

Recommended:

## QA
Please be aware of #5728, which this PR does _not_ fix.
- [x] resolves #5863
- [x] "Change Speed" dialog is now called "Pitch And Speed"
- [x] "Pitch And Speed" dialog is accessible like before (Select clip + right-click, right-click on clip header, `Edit > Audio Clips > Pitch and Speed...`)
- [x] "Pitch And Speed" is now also accessible by "Alt+P" when clip is fully selected
- [x] "Alt+Up" / "Alt+Down" change clip pitch by +/- 1 semitone
- [x] Pitch shift can be changed also during playback
- [x] Using extreme pitch shift and speed combinations does not freeze or crash Audacity (there may be pauses in audio, but normal audio should be recoverable by coming back to more reasonable settings)
- [x] Pitch-shift by cents is available in dialog
- [x] Pitch-shift and speed changes are drawn sensible in the clip header, or not drawn at all when space doesn't suffice.
- [x] Also with language set to e.g. Arabic or Hebrew
- [x] Quitting the dialog by ESC or Cancel restores previous pitch
- [x] Changing pitch adds an entry in history and is undoable
- [x] Changing pitch makes project dirty
- [x] Pitch changes persist after closing and re-opening project
- [x] Changing pitch, saving project and re-opening with earlier Audacity (e.g. 3.4.2) works fine (no pitch shifting)
- [x] Opening older project works fine

For mix-and-rendering, copy/paste or merging clips, the same rules apply for pitch-shifted and time-stretched clips. Here are some of the items to verify I can remember:

- [x] Mix-And-Render renders pitch shift
- [x] Export audio renders pitch shift
- [x] Merging clips with different pitch shift renders them
- [x] Pasting clip of pitch shift _x_ onto other clip with pitch shift _y_ doesn't merge them


Paul-Licameli (changes requested):
When you add images, you should also update the list of image file names in `lib-theme-resources/CMakeLists.txt`, so that the build will correctly reconfigure when you checkout a commit with a changed image list.

libraries/lib-wave-track/WideClip.cpp
```diff
@@ -56,3 +56,17 @@ double WideClip::GetStretchRatio() const
...
+{
+   return mChannels[0u]->GetSemitoneShift();
+}
+
+void WideClip::SetPitchShiftChangePublisher(
+   std::weak_ptr<PitchShiftChangePublisher> publisher)
+{
+   // On purpose set the publisher on the left channel only. This is not a clip
+   // property that is saved to disk, and else we'll get two callbacks for the
+   // same event.
```
Paul-Licameli, line 70: What happens if you swap channels?

libraries/lib-wave-track/WaveClip.h
```diff
@@ -611,6 +615,9 @@ class WAVE_TRACK_API WaveClip final :
    double mTrimRight { 0 };
    //! @}
 
+   std::weak_ptr<PitchShiftChangePublisher> mPitchShiftChangePublisher;
+   double mSemitoneShift { 0 };
```
Paul-Licameli, line 619: This commit is strange.

As I said I don't like it that lib-wave-track mentions extraneous things like quarter notes and stretch ratios, and now this is one more thing.  I might figure out a restructuring in my branch.

Why can't a track or clip simply publish whenever it changes, and have subscribers attached to it?  Why must you first set a publishing interface into it?

src/TrackArt.cpp
```diff
@@ -332,30 +372,40 @@ bool TrackArt::DrawClipTitle(
 
 bool TrackArt::DrawAudioClipTitle(
```
Paul-Licameli, line 373: I think this function should be moved into `WaveTrackAffordanceControls.cpp` which is the sole user of it.

src/TrackArt.cpp
```diff
@@ -322,6 +323,45 @@ std::optional<ClipTitle> DoDrawAudioTitle(
...
+//! Returns the width of the drawn text and icon.
+int DrawPitchOrSpeedIconIfItFits(
+   wxDC& dc, const wxRect& titleRect, const wxString& clipTitle,
+   const wxBitmap& icon, const wxString& text, int right, int availableWidth)
+{
+   const auto textWidth = dc.GetTextExtent(text).GetWidth();
+
+   const auto iconHeight = icon.GetHeight();
+   const auto iconWidth = textWidth == 0 ? 0 : icon.GetWidth();
+   const auto rectWidth = iconWidth + dc.GetTextExtent(text).GetWidth();
```
Paul-Licameli, line 350: Repeated expression could be `textWidth`

src/TrackArt.cpp
```diff
@@ -322,6 +323,45 @@ std::optional<ClipTitle> DoDrawAudioTitle(
...
+   const auto iconTop =
+      titleRect.GetTop() + (titleRect.GetHeight() - iconHeight) / 2;
+   const auto textTop = titleRect.GetTop();
+   const auto height = titleRect.GetHeight();
+
+   if (rectWidth == 0 || rectWidth > availableWidth)
+      return 0;
+
+   dc.DrawText(text, right - textWidth, textTop);
+   dc.DrawBitmap(icon, right - textWidth - iconWidth, iconTop);
```
Paul-Licameli, line 361: Should you always draw that bitmap, or not in case you reassigned 0 to iconWidth?

Paul-Licameli: Oh I see the early return did that, but I found it a bit non obvious

src/tracks/playabletrack/wavetrack/ui/ChangeClipSpeedDialog.cpp
```diff
@@ -11,106 +11,237 @@
...
+{
+public:
+   ScopedVerticalLay(ShuttleGui& s, int iProp = 1)
+       : ScopedSizer<void, int>(
+            s, &ShuttleGui::StartVerticalLay, &ShuttleGui::EndVerticalLay,
+            std::move(iProp))
+   {
+   }
+};
+} // namespace
```
Paul-Licameli, line 81: Those classes look nice, but I'd rather you just follow conventional use of ShuttleGui for bracketing things.

libraries/lib-wave-track/WideClip.cpp
saintmatthieu: This method is called when preparing tracks for playback. During playback, swapping channels isn't possible. After playback, the publisher will have expired, so swapping channels wouldn't have consequence.

src/tracks/playabletrack/wavetrack/ui/ChangeClipSpeedDialog.cpp
saintmatthieu: It helped me when I had to change things around, and I think it would also have helped me when I looked at this source file the first time and was trying to figure out how it all worked.

src/tracks/playabletrack/wavetrack/ui/ChangeClipSpeedDialog.cpp
```diff
@@ -11,106 +11,237 @@
...
-            .NameSuffix(Verbatim("%"))
-            .TieNumericTextBox({}, mClipSpeed, 14, true)
-            ->Bind(wxEVT_TEXT_ENTER, [this](auto&) { OnOk(); });
+         ScopedHorizontalLay h { s, wxLeft };
+         s.AddPrompt(XO("Clip Pitch"));
+      }
+      {
+         ScopedHorizontalLay h { s, wxLeft };
+         constexpr auto minSemis = -24;
+         constexpr auto maxSemis = 60;
```
Paul-Licameli, line 140: Two octaves down, five octaves up -- might comment somewhere that these are the bounds.  Maybe more prominently in the header file.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
```diff
@@ -648,6 +653,18 @@ void WaveTrackAffordanceControls::OnRenderClipStretching(
    SelectInterval(project, *interval);
 }
 
+void WaveTrackAffordanceControls::OnPitchShift(AudacityProject& project, bool up)
+{
+   const auto [track, it] = SelectedIntervalOfFocusedTrack(project);
+   if (track != FindTrack().get())
+      return;
+   auto interval = *it;
+   interval->SetSemitoneShift(interval->GetSemitoneShift() + (up ? 1 : -1));
```
Paul-Licameli, line 662: This does not impose the limits on pitch shift that the dialog does.

Paul-Licameli: It's a little strange that this is a member function of a UI object (representing a certain rectangle of the screen).  Can't it simply be a non-member function?

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
```diff
@@ -38,10 +38,11 @@ AudioSegmentFactory::CreateAudioSegmentSequence(
 std::vector<std::shared_ptr<AudioSegment>>
 AudioSegmentFactory::CreateAudioSegmentSequenceForward(double t0) const
 {
-   ClipConstHolders sortedClips { mClips };
+   auto sortedClips = mClips;
```
Paul-Licameli, line 43: I really dislike this entire commit -- adding mutability.

Let it stand, but I may explain the alternative with another suggested commit.

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.h
```diff
@@ -15,7 +15,8 @@ class TIME_AND_PITCH_API StaffPadTimeAndPitch final :
 
 private:
    void BootStretcher();
-   const std::unique_ptr<staffpad::TimeAndPitch> mTimeAndPitch;
```
Paul-Licameli, line 18: Why remove `const`?  That's not necessary and having it here is useful information to a maintainer.

libraries/lib-wave-track/WaveClip.h
saintmatthieu: Looks good to me - see "Use Observer framework" commit

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
```diff
@@ -56,6 +70,7 @@ void StaffPadTimeAndPitch::GetSamples(float* const* output, size_t outputLen)
       // Pass-through
       return mAudioSource.Pull(output, outputLen);
 
+   std::lock_guard<std::mutex> lock(mTimeAndPitchMutex);
```
Paul-Licameli, line 61: Oh, now I see at this commit why the member becomes non-const.

Consider using the utility `AtomicUniquePointer` instead of a mutex.

src/tracks/playabletrack/wavetrack/ui/ChangeClipSpeedDialog.cpp
saintmatthieu: Adding a comment.

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
saintmatthieu: Looks indeed like it should.

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
```diff
@@ -58,15 +60,22 @@ AudioSegmentFactory::CreateAudioSegmentSequenceForward(double t0) const
       }
       else if (clip->GetPlayEndTime() <= t0)
          continue;
+      const auto propagator = std::make_shared<PitchShiftChangePropagator>();
+      clip->SetPitchShiftChangePublisher(propagator);
+      auto subscriber = std::bind(
+         &PitchShiftChangePropagator::Subscribe, propagator,
+         std::placeholders::_1);
```
Paul-Licameli: Or just `std::mem_fn`

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
```diff
@@ -89,9 +98,28 @@ AudioSegmentFactory::CreateAudioSegmentSequenceBackward(double t0) const
...
       t0 = clip->GetPlayStartTime();
    }
    return segments;
 }
+
+void AudioSegmentFactory::PitchShiftChangePropagator::Publish(double semitones)
+{
+   if (mCb)
+      mCb(semitones);
+}
```
Paul-Licameli: It seems you are reinventing the Publisher utility and I don't know why.

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
```diff
@@ -28,15 +29,16 @@ AudioSegmentFactory::AudioSegmentFactory(
 
 std::vector<std::shared_ptr<AudioSegment>>
 AudioSegmentFactory::CreateAudioSegmentSequence(
-   double playbackStartTime, PlaybackDirection direction) const
+   double playbackStartTime, PlaybackDirection direction)
 {
+   mPropagators.clear();
```
Paul-Licameli: I think `mPropagators` never gets size more than 1, so why a vector?

src/tracks/playabletrack/wavetrack/ui/WaveTrackAffordanceControls.cpp
saintmatthieu: ... there is the call to `FindTrack`, though, which is a member method. I haven't understood yet under what conditions it may cause an early return.
Making that method const, at least, is a no-brainer.

saintmatthieu: Ok, found something better. Coming up...

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
saintmatthieu: I don't think an atomic would be helpful here, since it's not only just the reallocation of the object that must be synchronized. Remember that comment in SP:
```
  /**
    Set independent time stretch and pitch factors (synchronously to processing thread).
    ...
  */
  void setTimeStretchAndPitchFactor(double timeStretch, double pitchFactor);
  ```

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
saintmatthieu: It's outdated now, through the use of `Observer`.

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
saintmatthieu: You may ignore most of a9752ac3a7431db60b6742de5ae9a5b5281ea104 and 250725b5876d32aa78c9114af86f0dff36d0fe24, as I did in a recent commit put in place almost exactly what you described. See, too, that this change is outdated.

Paul-Licameli: The constructor of `ClipSegment` isn't passed a function that does subscription.  It is just passed a reference to `Observer::Publisher<PitchRatio>`.  That reference can be stored in `StaffPadTimeAndPitch::Parameters`.  Then `StaffPadTimeAndPitch` in its constructor simply subscribes to the publisher, passing in that lambda, and holding the `Subscription` object which is a sort of smart, weak pointer representing the connection between publisher and subscriber.

Paul-Licameli: There are multiple threads involved here, and that means some caution in the use of `Observer`.  Creating and destroying Subscriptions, or destroying the Publisher before the Subscription that it gave out, is safe so long as it is all in the main thread.  Publishing from other threads will be safe so long as the callback does not itself cause Subscriptions to be created or destroyed.  I think that applies.

Paul-Licameli: Where is Publish called then?

Paul-Licameli: I looked through this confusing stuff again, and now I see, ultimately it's in `WaveClip::SetSemitoneShift`

So then, that can simply call the inherited `Publish`

Paul-Licameli:
I see messages crossed about Observer.

You made simplifications, not exactly as I suggested, and I hope you agree it's all more understandable now.

Can you answer my questions about thread safety?  The subscriptions may be created and destroyed only in the main thread, and that includes destruction of either the publisher or the subscriber before the other.

Publication on another thread is safe if that is so.

Paul-Licameli:
I would move `CentShiftChange` into `ClipSegment.h` and then `WaveClip` could publish that type and simply pass a function to `Subscribe` without the extra lambda adaptor.

It's minor though.

Paul-Licameli:
All right, I see no real blocker.

I dislike the now increasing involvement of pitch and stretch notions in lib-wave-track, where they should be extraneous, but that restructuring is for later.

Assuming too that the thread safety of Observer is correct, and that you will squash it into fewer commits -- I can approve it

Paul-Licameli (approved):

libraries/lib-stretching-sequence/AudioSegmentFactory.cpp
saintmatthieu: > Creating and destroying Subscriptions, or destroying the Publisher before the Subscription that it gave out, is safe so long as it is all in the main thread. Publishing from other threads will be safe so long as the callback does not itself cause Subscriptions to be created or destroyed.

Incidentally, here it's the other way round: subscription creation and destruction is done on the audio thread ; publication on the main thread.

Just took a look at some of the Observer implementation, couldn't see thread unsafety (but I haven't strained myself too much).
Also tried a worst-case scenario, with very fast subscription creation and destructions, concurrently with very fast publications. Couldn't get the program to behave incorrectly, which is more evidence that I can take your word for it.

Paul-Licameli: That “inverse” pattern should be safe too.

Is there no way to lift the making and destroying of subscriptions to the main thread?

Shared and weak pointers manage the shared state in a thread safe way — but provided each thread uses different pointer objects.

It is possible to have races in the use of the identical pointer object in two threads.  Don’t do that.

Observer uses weak pointers.

Paul-Licameli: The thread safety issue should be commented!

dozzzzer:
@saintmatthieu 

1. When opening up the Pitch & Speed dialog and entering a new value, a strange "-24" pops up in the text box at random. Furthermore, after the text box is cleared and the new value is entered, it gets increased by 1:

https://github.com/audacity/audacity/assets/70063970/eb05ca0d-ba3d-4ae0-8d1e-3765b852582d

Also, I've observed that every time I put the cursor in the "cents" text box and press Backspace, the semitones value decreases by 1:

https://github.com/audacity/audacity/assets/70063970/013bcdcd-f4f0-4cd9-9f51-daefad08b2ad



2. Not sure if anything at all should be done about it, but during testing, I've observed somewhat deteriorating quality at higher values. It becomes more obvious from around 20+ semitones up. At this point, the sound becomes quite unusable while the competitor's algorithm yields a much cleaner result. Tested on one-shots, drum loops, bass loops and vocals. Here's an example of such deterioration: https://www.dropbox.com/scl/fo/71moh408cvrlz34pmrafr/h?rlkey=kq1xllldkag4e71hswmk5t4zg&dl=0 This isn't too critical since it manifests itself at higher (although not extreme) values but still smth worth to be aware of.

(TBC)

saintmatthieu:
> 1. When opening up the Pitch & Speed dialog and entering a new value, a strange "-24" pops up in the text box at random. Furthermore, after the text box is cleared and the new value is entered, it gets increased by 1

Tweaks pushed, hope they improve UX.

2.
About 2., I allowed pitch shifting to such extreme ratios because the weird resulting sounds might be interesting to the users at an educational and experimental level, and maybe useful in the creation of weird atmospheres. When we have tempo curves, I can imagine that one might want to slowly drop or raise the pitch to something increasingly unrecognizable.

Ideally of course, it'd pitch up and down more realistically, but the amount of work involved in improving this is completely unknown.

I'm confident that the technology will be very useful as it is. One potential area of improvement is formant preservation: when shifting voice, one quickly gets into a Chipmunk effect. I just created an issue for this: https://github.com/audacity/audacity/issues/5902. Maybe this will address some of the limitations you've noticed ? (Could you please share your examples using Google Drive instead?)

dozzzzer:
@saintmatthieu 

1. It seems that the upper/lower boundaries aren't in place when pitch-shifting using Alt (Option) + ↑/↓. After reaching a value of a couple of hundred, an attempt to play back causes Audacity to freeze.
2. On macOS, quitting the Pitch & Speed dialog by ESC doesn't restore the previous pitch value and submits the entered value instead; clicking Cancel, though, doesn't work as it should either: it resets the current pitch value to zero.

saintmatthieu:
> I'm confident that the technology will be very useful as it is. One potential area of improvement is formant preservation: when shifting voice, one quickly gets into a Chipmunk effect. I just created an issue for this: https://github.com/audacity/audacity/issues/5902. Maybe this will address some of the limitations you've noticed ? (Could you please share your examples using Google Drive instead?)

@dozzzzer I omitted the problem of aliasing and imaging. This may also explain the lower quality you noticed. I created another ticket for that: https://github.com/audacity/audacity/issues/5904

petersampsonaudacity:
Testing on W10 with @saintmatthieu 's latest branch for this:  audacity-win-3.5.0-alpha-20240202+81dc5f9-x64-msvc2022

Pitch and Speed  changing appears to work while Audacity is playing - but there are some issues of note:

1) when Audacity is playing and you hover the cursor over the Clip-handle drag bar the cursor icon that you get is a **white no-entry**  icon - to me this implies that no action is available but clicking on it still brings up the drag-bar context menu, including the **Pitch and Speed** command.
This is **_not_** introduced as part of this change - the same occurs in 3.4.2

2) when Audacity is playing the **Clip Speed %** in the **Pitch and Speed** dialog is grayed-out and unavailable.  
This is _**not**_ the case when the command is invoked while Audacity is not playing.
image

3) The **Pitch and Speed** context menu :command has a much reduced UI compared to the **Effect > Pitch and Tempo > Change Speed and Pitch** menu command:
image

4) I was really expecting the **Change Speed and Pitch** effect to be a further Audacity command that would now be available in the Real-time effects stack - and not just a clip-handle context menu.

5) In the clip-handle context menu command dialog for **Pitch and Speed** - you appear to be able to change pitch and speed independently.  
This appears to contrast with what happens with the **Effect > Pitch and Tempo > Change Speed and Pitch** menu command.

petersampsonaudacity:
@saintmatthieu  @LWinterberg 

Perhaps most importantly:

6) this PR states that the dynamic pitch change in the  clip being played is "non-destructive"
BUT testing shows that the pitch change is actually applied destructively (the change goes on the Undo/History stack).

petersampsonaudacity:
@saintmatthieu @LWinterberg 

7)  the dialog seems un- necessarily wide

image

petersampsonaudacity:
8) in the old clip-handle context menu the command was **Change Speed** - in the new menu it is just **Pitch and Speed**.

Surely it would be beyter to add back the "Change" and make the new command be **Change Pitch and Speed**
a) for clarity of purpose  ("does what it says on the tin")
b) to differentiate from the **Render Pitch and Speed** just beneath it

image

petersampsonaudacity:
I have updated the Manual for the changes as they currently are - but retained P1s just in case further changes are effected.

Issue #5858, opened 2024-01-11
Windows: Audacity crashes, when making a stereo track using keyboard


Giovani93:
### Bug description

<details>
Hello.
I have a problem with Audacity. When I make stereo tracks using make stereo track in track menu, Audacity crashes.
Also, there are some mistakes in this report. I use really Windows 10 (64 bit), not Windows NT.
Please, could You look on this report, if It is possible? Or what can It be? Something with My notebook?
Thank You.
Marco Oros
</details>

### Steps to reproduce

1. create two mono tracks
2. Open the first track's menu by pressing the Application key or Shift + M
3. Choose Make stereo track, using either the mouse or the keyboard. 

### Expected behavior

A stereo track is made

### Actual behavior

A crash

[Audacity crash report.txt](https://github.com/audacity/audacity/files/13899552/Audacity.crash.report.txt)

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Not reproducible on macOS and Linux

Giovani93:
I've tried to reinstall Audacity, but with same result.
So, what should I do? I don't want to change My computer due to one software.

DavidBailes:
> I cannot reproduce this on W10 with 3.4.2 - works fine for me.

@petersampsonaudacity . Whether or not there is a crash seems to depend on whether the first track's menu was opened using   the keyboard or the mouse:
1. Open the first track's menu by pressing the Application key or Shift + M.
2. Choose Make stereo track, using either the mouse or the keyboard. The combine mono to stereo message box opens.
3. Press the Yes button, using either the mouse or the keyboard.
Audacity crashes.

If step 1 is replaced by Open the track's menu using the mouse, then Audacity does not crash.

dozzzzer:
Thanks, @DavidBailes ! Great observation

dozzzzer:
@Giovani93 FYI as for the crash reports failing to send, there's an issue logged https://github.com/audacity/audacity/issues/4154

petersampsonaudacity:
@DavidBailes 

 good catch (I always use the mouse, so I overlooked the Shift+M to get the menu of the focused track)

The additional fascinating thing is that if you accept the offered recovery on relaunch then the project is restored with the track as stereo.

1. Import a stereo track
2. Split the stereo track
3. Observe: two single channel tracks
4. Ensure focus is on top track
5. Shift+M to get the TCP menu
6. choose Make Stereo Track
7. Observe: Crash
[...]

PR #5911, opened 2024-02-03
Fix: Windows: Audacity crashes, when making a stereo track using keyboard

JoeS209:
Resolves: #5858

Resolve TrackPanelCell lifetime issues via usage of shared_ptr vs. raw pointer.  I'm no expert in shared_ptr but this seems to work.  "Works on my machine..."  Open to critiques, suggestions.

Recommended:

### DIFF
--- a/src/CellularPanel.cpp
+++ b/src/CellularPanel.cpp
@@ -188,7 +188,7 @@ bool CellularPanel::CancelDragging( bool escaping )
       auto pClickedCell = state.mpClickedCell.lock();
       if (pClickedCell)
          ProcessUIHandleResult(
-            pClickedCell, {},
+            pClickedCell.get(), {},
             refreshResult | state.mMouseOverUpdateFlags );
       state.mpClickedCell.reset();
       state.mUIHandle.reset(), handle.reset(), ClearTargets();
@@ -319,7 +319,7 @@ void CellularPanel::HandleMotion
          // Re-draw any highlighting
          if (oldCell) {
             ProcessUIHandleResult(
-               oldCell, oldCell, updateFlags);
+               oldCell.get(), oldCell.get(), updateFlags);
          }
       }
 
@@ -419,7 +419,7 @@ void CellularPanel::HandleMotion
       UpdateStatusMessage( {} );
 
    if (newCell)
-      ProcessUIHandleResult(newCell, newCell, refreshCode);
+      ProcessUIHandleResult(newCell.get(), newCell.get(), refreshCode);
 }
 
 void CellularPanel::Leave()
@@ -547,7 +547,7 @@ void CellularPanel::HandleWheelRotation( TrackPanelMouseEvent &tpmEvent )
    unsigned result =
       pCell->HandleWheelRotation( tpmEvent, GetProject() );
    ProcessUIHandleResult(
-      pCell, pCell, result);
+      pCell.get(), pCell.get(), result);
 }
 
 void CellularPanel::OnCaptureKey(wxCommandEvent & event)
@@ -564,7 +564,7 @@ void CellularPanel::OnCaptureKey(wxCommandEvent & event)
    if (t) {
       const unsigned refreshResult =
          t->CaptureKey(*kevent, *mViewInfo, this, GetProject());
-      ProcessUIHandleResult(t, t, refreshResult);
+      ProcessUIHandleResult(t.get(), t.get(), refreshResult);
       event.Skip(kevent->GetSkipped());
    }
 
@@ -621,7 +621,7 @@ void CellularPanel::OnKeyDown(wxKeyEvent & event)
    if (t) {
       const unsigned refreshResult =
          t->KeyDown(event, *mViewInfo, this, GetProject());
-      ProcessUIHandleResult(t, t, refreshResult);
+      ProcessUIHandleResult(t.get(), t.get(), refreshResult);
    }
    else
       event.Skip();
@@ -644,7 +644,7 @@ void CellularPanel::OnChar(wxKeyEvent & event)
    if (t) {
       const unsigned refreshResult =
          t->Char(event, *mViewInfo, this, GetProject());
-      ProcessUIHandleResult(t, t, refreshResult);
+      ProcessUIHandleResult(t.get(), t.get(), refreshResult);
    }
    else
       event.Skip();
@@ -676,7 +676,7 @@ void CellularPanel::OnKeyUp(wxKeyEvent & event)
    if (t) {
       const unsigned refreshResult =
          t->KeyUp(event, *mViewInfo, this, GetProject());
-      ProcessUIHandleResult(t, t, refreshResult);
+      ProcessUIHandleResult(t.get(), t.get(), refreshResult);
       return;
    }
 
@@ -791,7 +791,7 @@ try
          const UIHandle::Result refreshResult =
             handle->Drag( tpmEvent, GetProject() );
          ProcessUIHandleResult
-            (pClickedCell, pCell, refreshResult);
+            (pClickedCell.get(), pCell.get(), refreshResult);
          state.mMouseOverUpdateFlags |= refreshResult;
          if (refreshResult & RefreshCode::Cancelled) {
             // Drag decided to abort itself
@@ -813,7 +813,7 @@ try
          UIHandle::Result refreshResult =
             handle->Release( tpmEvent, GetProject(), this );
          ProcessUIHandleResult
-            (pClickedCell, pCell,
+            (pClickedCell.get(), pCell.get(),
              refreshResult | moreFlags);
          state.mUIHandle.reset(), handle.reset(), ClearTargets();
          state.mpClickedCell.reset();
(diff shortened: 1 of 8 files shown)


petersampsonaudacity:
@dozzzzer @DavidBailes 

Testing on W10 with @JoeS209 's latest branch build:  audacity-win-3.5.0-alpha-20240204+b11feb8-x64

This now seems to work properly in Joe's branch with no crash and the stereo track is properly created.

I tested entirely using the keyboard with no mouse
1. Shift+M in the top track to get the menu
2. scroll down in menu with down arrow to get the "Make Stereo Track" command
3. Press enter
4. Observe: stereo track is made _(and no crash occurs)_

src/CellularPanel.cpp
```diff
@@ -962,29 +961,30 @@ void CellularPanel::HandleClick( const TrackPanelMouseEvent &tpmEvent )
...
       state.mMouseOverUpdateFlags |= refreshResult;
    }
 }
 
 void CellularPanel::DoContextMenu( TrackPanelCell *pCell )
 {
-   if( !pCell ) {
-      pCell = GetFocusedCell();
-      if( !pCell )
+   std::shared_ptr<TrackPanelCell>pCellShared(pCell);
```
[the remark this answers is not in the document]
JoeS209: Please see the second commit in the pull request.  This was resolved by not passing a raw pointer to DoContextMenu, and instead passing a shared pointer.  I realized this myself after the first commit.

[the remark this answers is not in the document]
Paul-Licameli: I agree with @vsverchinsky  -- as originally written, this hazardous code could lead to a double-free error corrupting the memory manager.  But, indeed, you must have discovered this by yourself and fixed it.

It is rare to construct a `shared_ptr` directly from a raw pointer -- instead one more often uses `std::make_shared` to combine construction of an object with the making of the first shared pointer to it.  All other shared (or weak) pointers must then be copies of previously existing shared pointers (or lockings of previously existing weak pointers).

src/CellularPanel.cpp
```diff
@@ -962,29 +961,30 @@ void CellularPanel::HandleClick( const TrackPanelMouseEvent &tpmEvent )
...
       }
       ProcessUIHandleResult(
-         pCell.get(), pCell.get(), refreshResult);
+         pCell, pCell, refreshResult);
       state.mMouseOverUpdateFlags |= refreshResult;
    }
 }
 
-void CellularPanel::DoContextMenu( TrackPanelCell *pCell )
+void CellularPanel::DoContextMenu( const std::shared_ptr<TrackPanelCell> &pCell )
```
Paul-Licameli, line 969: I would simplify this function:  let its `pCell` argument be a non-reference -- which implies a copying of a shared pointer when it is called.  So the explicit copying when you make `pCellLocal` is instead implicit.  And then you can simply reassign `pCell` as needed.  Fewer lines of code would change.

src/CellularPanel.cpp
```diff
@@ -962,29 +961,30 @@ void CellularPanel::HandleClick( const TrackPanelMouseEvent &tpmEvent )
...
 {
-   if( !pCell ) {
-      pCell = GetFocusedCell();
-      if( !pCell )
+   std::shared_ptr<TrackPanelCell>pCellLocal = pCell;
```
Paul-Licameli, line 971: So this variable initialization might disappear -- but even if this were needed, I would also use `auto` to simplify the declaration, avoiding the the verbosity of the long type name.  Let the compiler deduce it instead.

src/CellularPanel.cpp
JoeS209: Thanks Paul.  I like this suggestion.  I'll make the change, retest, and commit the changes into the pull request.  Stay tuned.

src/TrackPanel.cpp
```diff
@@ -1019,8 +1020,9 @@ void TrackPanel::UpdateVRulerSize()
 
 void TrackPanel::OnTrackMenu(Track *t)
 {
+   std::shared_ptr<TrackPanelCell> ptr1(nullptr);
    CellularPanel::DoContextMenu(
-      t ? &ChannelView::Get(*t->GetChannel(0)) : nullptr);
+      t ? ChannelView::Get(*t->GetChannel(0)).GetShared() : ptr1);
```
Paul-Licameli, line 1025: You can simplify this:  you don't need `ptr1` but can just keep `nullptr` in the ternary expression.

Furthermore:  Don't define the function `GetShared` but just use `shared_from_this` directly.

src/CellularPanel.cpp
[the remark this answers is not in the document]
JoeS209: Yes, I realized the error not long after I checked in the first commit.  I then fixed it on my own and added the second commit to the pull request.

src/TrackPanel.cpp
Paul-Licameli: As I understand it, the real heart of the bug fix is here:  the shared pointer constructed from the view object extends its lifetime for the duration of the call that pops up the context menu.

src/CellularPanel.cpp
JoeS209: In general I don't like "auto" because it makes the code less comprehensible, for me anyway.  But like you also suggest, this initialization should disappear after I make the other change.

src/TrackPanel.cpp
Paul-Licameli: ... So that, inside `DoContextMenu`, `pCellLocal` (or just `pCell`) is still guaranteed to point to a live object.

Paul-Licameli:
Thank you for your fix!  The first commit alone is wrong, but the two together look correct and sufficient to fix the crash.

I suggest simplifications.  Please make them, and I will approve it.

src/TrackPanel.cpp
JoeS209: I don't have the code in front of me at the moment but essentially yes.  The crash was occurring across the UI function near the end, because the object that the raw pointer was pointing to disappeared.

Paul-Licameli:
Wait now, Joe -- I had an afterthought that could simplify the entire fix further.

Don't change `ProcessUIHandleResult` at all.

Rather, in `CellularPanel::DoContextMenu`, assign `pCell` to a local `weak_ptr<TrackPanelCell>`, and then use `pCell.reset()` after `DoContextMenu`.

Say, `wCell` is that new local.

Then, pass `wCell.lock().get()` not `pCell` to `ProcessUIHandleResult`

This should be enough to fix the problem, with less disturbance of other code.

Do you understand why?

JoeS209:
Thanks, Paul.  Yes, I think I understand what you are saying here.  I had originally was looking for a solution like that, but I became fond of getting away from passing raw pointers around.  But I'll implement your suggestion if you like.

JoeS209:
And I assume that we still pass a shared_ptr to `CellularPanel::DoContextMenu`.

Paul-Licameli:
Yes, change the signature of `DoContextMenu`.

Remember this about `weak_ptr` : it allows some code to safely point at an object, while not controlling its lifetime.

So if the cell object wants to die after handling the context menu, let it die, but then safely detect whether it died.

`TrackPanel::ProcessUIHandleResult` was written to deal correctly will null pointers passed into it, but not with junk pointers to destroyed objects.

What I suggest is enough to replace the unsafe junk pointers with safe null pointers, while also not changing the lifetime of the cell objects.

And therefore, it's a more conservative fix, not delaying the destruction of the cells compared to the previous master.

JoeS209:
OK, will do.  Note that I had changed `ProcessUIHandleResult` to take shared pointers largely due to this comment that was in `CellularPanel::DoContextMenu`, before my changes:

```
   // To do: use safer shared_ptr to pCell
   ProcessUIHandleResult(pCell, pCell, refreshResult);
```

Paul-Licameli:
> OK, will do. Note that I had changed `ProcessUIHandleResult` to take shared pointers largely due to this comment that was in `CellularPanel::DoContextMenu`, before my changes:
> 
> ```
>    // To do: use safer shared_ptr to pCell
>    ProcessUIHandleResult(pCell, pCell, refreshResult);
> ```

Did I write that?  Oh, yeah, in 2018 at commit f6609e5.

But for present purposes, indeed, you can write less code to achieve the result we need.

JoeS209:
OK, changed made, tested locally, and committed.  Checks currently in progress.  Please let me know if there are any additional changes needed.

Paul-Licameli (approved):

Paul-Licameli:
Now it's a good, simple fix.

The three commits should be squashed down to one, after this is also approved by our QA.

petersampsonaudacity:
Tested again on W10 with @JoeS209 's latest branch:  audacity-win-3.5.0-alpha-20240206+d0c721a-x64

Still works fine with the revised "good, simple fix"

JoeS209:
> Now it's a good, simple fix.
> 
> The three commits should be squashed down to one, after this is also approved by our QA.

Thanks, Paul.  I really appreciate the feedback and suggestions.  About the "squashing" of the commits, is there something that I need to do for that, or does that happen automatically when the changes are pulled into the master?

crsib:
We will squash before the merge, no worries

PR #5916, opened 2024-02-05
Bugfix: do not use project-global object for loop metadata

### LINKED ISSUE #5914: Loop tempo metadata persists through imports
### Bug description

The tempo metadata of an imported loop, when present, remains in memory. Upon any subsequent file import, whether music or not, that metadata will be interpreted as being that of those new files, too.

### Steps to reproduce

1. Extract these files: 
[persistent-metadata-bug.zip](https://github.com/audacity/audacity/files/14167477/persistent-metadata-bug.zip)
2. Import `square.wav`. Tempo isn't detected.
3. Delete the track. Import `tagged-70.wav`. 70BPM is detected.
4. Remove the track.
5. Change project tempo to something other than 70.
6. Import again `square.wav`.

### Expected behavior

Tempo isn't detected.

### Actual behavior

 It gets detected as 70BPM.

### Audacity Version

### Operating system

Windows, macOS

saintmatthieu:
Resolves: #5914 

Recommended:


src/ProjectFileManager.cpp
```diff
@@ -1529,12 +1524,10 @@ bool ProjectFileManager::Import(
...
-      bool success = Importer::Get().Import(project, fileName,
-                                            &importProgress,
-                                            &WaveTrackFactory::Get( project ),
-                                            newTracks,
-                                            newTags.get(),
-                                            errorMessage);
+      std::optional<LibFileFormats::AcidizerTags> acidTags;
+      bool success = Importer::Get().Import(
+         project, fileName, &importProgress, &WaveTrackFactory::Get(project),
+         newTracks, newTags.get(), acidTags, errorMessage);
```
vsverchinsky, line 1530: `newTags` could be a local object, that will simplify things. the acid tags could be then extracted like this:
```
if(newTags.HasTag(TAG_ACID))
   ...
```
I wonder if it's a desired behaviour at all when all other tags get overwritten? But it is not related to the scope of this PR

modules/mod-pcm/ImportPCM.cpp
```diff
@@ -439,10 +438,8 @@ void PCMImportFileHandle::Import(ImportProgressListener &progressListener,
    // To begin with, only trust the Muse Hub, with whom we collaborate and can
    // ensure they comply. In the future we may extend this list.
    const std::vector<std::string> trustedDistributors { "Muse Hub" };
-   if (
-      const auto acidTags =
-         LibImportExport::GetAcidizerTags(*mFile, trustedDistributors))
-      tags->SetTag(TAG_ACID, LibImportExport::AcidizerTagsToString(*acidTags));
+   if (const auto acidTags = LibImportExport::GetAcidizerTags(*mFile, trustedDistributors))
+      outAcidTags.emplace(*acidTags);
```
vsverchinsky, line 442: So this is the change: instead of being written into project-owned tags structure, ACID tag is exclusively written into new parameter.
But now ACID tag never gets written into project tags? Is that desired change?

vsverchinsky (approved):
Approved, though a question about project-owned tags needs to be clarified. 

This PR introduces many code style changes that make it harder to read and understand.

src/ProjectFileManager.cpp
saintmatthieu: Not sure I understand: 
```
if(newTags.HasTag(TAG_ACID))
   acidTag = StringToAcidizerTags(newTags.GetTag(TAG_ACID))
```
is what was there previously. But then, `newTags` gets committed.
The loop metadata tags are clip-dependent, though, so re-using those of clip A for the import of clip B is wrong.
The other tags are
```
TAG_TITLE
TAG_ARTIST
TAG_ALBUM
TAG_TRACK
TAG_YEAR
TAG_GENRE
TAG_COMMENTS
TAG_SOFTWARE
TAG_COPYRIGHT
```
and to me it is ok that they are project-wide if they are used when exporting?

modules/mod-pcm/ImportPCM.cpp
saintmatthieu: Yes, it was wrong to create `TAG_ACID` and use it for clip-dependent metadata.
> now ACID tag never gets written into project tags?

They simply don't exist anymore.

src/ProjectFileManager.cpp
vsverchinsky: On export tags are copied from the project tags, so answerd is yes.
What I meant was that by making newTags a local structure we could avoid unnecessary (and there no control on the user side over that) tag overwrites, currently it's a bit complicated:
```
auto cleanup = finally([&]{
   if ( !committed )
      Tags::Set( project, oldTags );
});
auto newTags = oldTags->Duplicate();
Tags::Set( project, newTags );
```

saintmatthieu: Isn't this to satisfy this comment:
```
   // do the actual import, creating whatever tracks are necessary with
   // the WaveTrackFactory and calling the progress callback every iteration
   // through the importing loop
   // The given Tags and AcidizerTags structures may also be modified.
   // In case of errors or exceptions, it is not necessary to leave outTracks,
   // tags or acidTags unmodified. If resulting outTracks is not empty, then
   // each member of it must be a nonempty vector.
   virtual void Import(
      ImportProgressListener& progressListener, WaveTrackFactory* trackFactory,
      TrackHolders& outTracks, Tags* tags,
      std::optional<LibFileFormats::AcidizerTags>& acidTags) = 0;
      ```
in ImportPlugin.h ?

Issue #2408, opened 2022-01-11
Proposal DC Offset removal


petersampsonaudacity:
**Describe the bug**
DC offset removal is only bundled in the Normalize effect. This leads to DC offset not being done at all, or being done too late as a final editing step in Normalize.

DC on the signal should be removed early in the editing process but it is often overlooked by users as it is "hidden" in an effect normally used late in the editing process. 

This was originally logged as a Proposal in the Audacity Wiki:
https://wiki.audacityteam.org/wiki/Proposal_DC_Offset_removal

**To Reproduce**
1. Look for a command in the menus for DC Offset Removal
2. Observe: you won't find one

**Expected behavior**
1. Create a new command for **Remove DC offset**
2. arguable as to whether it should be in the Effect menu of the Tools menu
3. remove the DC offset option from the Normalize effect

**Screenshots**
![image](https://user-images.githubusercontent.com/50205984/148985909-f7380559-f813-4017-a87c-e1a96e4dcb86.png)

![image](https://user-images.githubusercontent.com/50205984/148986465-f4d3f10f-78d1-4f05-87ab-817f867a96dd.png)

**Additional information (please complete the following information):**
 - OS: all
 - Version:  all

**Additional context**
Could be automatically invoked immediately post-capture, or post import, if Audacity detects significant DC (to nudge the user into optionally dealing with it).

werame:
Can you clarify what you mean by the first two algorithms "dynamic bias adjust" and "absolute offset"?

SuperCollider only does the 3rd one IIRC, as LeakDC.

Some useful discussion regarding algorithm here: https://scsynth.org/t/why-leakdc-and-not-a-standard-hpf-for-dc-removal/4780/3

SteveDaulton:
> Can you clarify what you mean by the first two algorithms "dynamic bias adjust" and "absolute offset"?

That came from some experiments that I tried over a decade ago. 

If I recall correctly, "dynamic bias adjust" measured the amount DC bias over the first part of the selection, and applied compensation. It would then repeat the process on blocks of data, updating the bias correction as it went - in effect, a very low frequency high-pass filter, but with the advantage that it looked ahead to determine the required bias correction, so  no click at the start. 

- If the DC offset is constant, then "dynamic bias adjust" is identical to "DC offset correction" in Audacity's Normalize effect. 
- If the amount of DC offset drifts over time, then "dynamic bias adjust" performs better than "DC offset correction".


[...]

PR #5917, opened 2024-02-05
Add RemoveDCOffset effect

PatrickKallenbach:
Resolves: https://github.com/audacity/audacity/issues/2408

The original issue was requesting a DC Offset tool similar to what the Normalize effect can do before normalization. In my university class, we were assigned to find and contribute to an open source project, so I thought I should contribute to this project which I have used myself in the past. The new effect is modified from the existing Normalize effect to automatically remove the offset, bypassing the menu similar to the Invert effect. I am certain improvements could be made upon the modified files, perhaps adding more features, or removing steps that were required for the Normalize effect but not for DC offset removal. I also only added the effects to the src/effects folder, but I do not know if there is something else that needs to know of the updated effects, since it does not currently categorize with the other related effects.

Recommended:

### DIFF
--- a/RemoveDCOffset.cpp
+++ b/RemoveDCOffset.cpp
@@ -1,0 +1,355 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  RemoveDCOffset.h
+
+  Patrick Kallenbach
+
+  Adapted from Normalize
+  Dominic Mazzoni
+  Vaughan Johnson (Preview)
+
+*******************************************************************//**
+
+\class EffectRemoveDCOffset
+\brief An Effect to remove DC offset from a selection, adapted from Normalize.cpp
+
+*//*******************************************************************/
+#include "RemoveDCOffset.h"
+#include "EffectEditor.h"
+#include "EffectOutputTracks.h"
+#include "LoadEffects.h"
+
+#include <math.h>
+
+#include <wx/checkbox.h>
+#include <wx/stattext.h>
+#include <wx/valgen.h>
+
+#include "Prefs.h"
+#include "../ProjectFileManager.h"
+#include "ShuttleGui.h"
+#include "WaveTrack.h"
+#include "../widgets/valnum.h"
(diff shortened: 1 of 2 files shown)


Paul-Licameli:
This is not a complete code review, but in fact -- the .h file is not needed at all.   You can just define the class locally in the .cpp file.  More than that -- I would also wrap the whole thing in an anonymous namespace, and save a little bit of space in the symbol tables that the linker has to cope with.

Audacity wasn't originally written this way back in the late twentieth century, but nowadays, static registration objects do a lot of magic to push things into registries (of effects, macro commands, whatever) allowing scattered definitions of details, and making it unnecessary for those details to expose a header file.

LWinterberg:
> I do not know if there is something else that needs to know of the updated effects, since it does not currently categorize with the other related effects.

That's in resources/EffectsMenuDefaults.xml, probably in the "noise removal and repair" category.

RemoveDCOffset.cpp
```diff
@@ -0,0 +1,355 @@
...
+            msg = topMsg +
+               // TODO: more-than-two-channels-message
+               XO("Processing first track of stereo pair: %s")
+                  .Format(trackName);
+
+         // Use multiplier in the second, processing loop over channels
+         auto pOffset = offsets.begin();
+         auto pExtent = extents.begin();
+         for (const auto channel : channels) {
+            const auto extent = oneChannel ? *pExtent++: maxExtent;
```
saintmatthieu, line 183: Hey Patrick, thanks for your contribution!
It looks like you don't use `extent`, in the end.

Paul-Licameli:
Anonymous namespace just means this -- add the line
`namespace {`
after any #include directives and forward declarations of incomplete types, then at the bottom of the file,
'}'

petersampsonaudacity:
@LWinterberg   

If/when this PR gets pulled for this new effect, are you planniing to remove the DC offset removal from the Normalize effect or just leave it in there?

LWinterberg:

Issue #5906, opened 2024-02-01
Occasional crash in tempo estimation


saintmatthieu:
### Bug description

Importing that audio file sometimes crashes Audacity: [AudacitySpectral.zip](https://github.com/audacity/audacity/files/14126902/AudacitySpectral.zip)

### Steps to reproduce

1. On empty project, import that file

### Expected behavior

File opens

### Actual behavior

Audacity crashes

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

Crash occurs in PFFFT, when used by the recently introduced tempo-estimation algorithm.
Seems intermittent (??)

dozzzzer:
Tested

saintmatthieu:
Here too, @Paul-Licameli you may merge.

PR #5918, opened 2024-02-06
5906 fix: pass aligned vectors to pffft

saintmatthieu:
Resolves: #5906

Recommended:

### DIFF
--- a/libraries/lib-music-information-retrieval/PowerSpectrumGetter.cpp
+++ b/libraries/lib-music-information-retrieval/PowerSpectrumGetter.cpp
@@ -17,22 +17,34 @@ namespace MIR
 PowerSpectrumGetter::PowerSpectrumGetter(int fftSize)
     : mFftSize { fftSize }
     , mSetup { pffft_new_setup(fftSize, PFFFT_REAL) }
-    , mWork(fftSize)
+    , mData { reinterpret_cast<float*>(
+         pffft_aligned_malloc(fftSize * sizeof(float))) }
+    , mWork { reinterpret_cast<float*>(
+         pffft_aligned_malloc(fftSize * sizeof(float))) }
 {
+   std::fill(mData, mData + fftSize, 0.f);
+   std::fill(mWork, mWork + fftSize, 0.f);
 }
 
 PowerSpectrumGetter::~PowerSpectrumGetter()
 {
    pffft_destroy_setup(mSetup);
+   pffft_aligned_free(mData);
+   pffft_aligned_free(mWork);
 }
 
-void PowerSpectrumGetter::operator()(float* buffer, float* output)
+float* PowerSpectrumGetter::GetInputPtr()
 {
-   pffft_transform_ordered(mSetup, buffer, buffer, mWork.data(), PFFFT_FORWARD);
-   output[0] = buffer[0] * buffer[0];
+   return mData;
+}
+
+void PowerSpectrumGetter::Process(float* output)
+{
+   pffft_transform_ordered(mSetup, mData, mData, mWork, PFFFT_FORWARD);
+   output[0] = mData[0] * mData[0];
    for (auto i = 1; i < mFftSize / 2; ++i)
       output[i] =
-         buffer[i * 2] * buffer[i * 2] + buffer[i * 2 + 1] * buffer[i * 2 + 1];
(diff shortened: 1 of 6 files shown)


libraries/lib-music-information-retrieval/PowerSpectrumGetter.h
```diff
@@ -29,17 +29,22 @@ class PowerSpectrumGetter
...
     */
-   void operator()(float* buffer, float* output);
+   void Process(float* output);
 
 private:
    const int mFftSize;
    PFFFT_Setup* mSetup;
-   std::vector<float> mWork;
+   float* mData;
+   float* mWork;
```
Paul-Licameli, line 48: Now that the class has dumb pointers instead of a vector member, it requires the explicit free in the destructor.

So it should also `delete` the copy and assignment operators, to prevent accidental implied copies that would lead to double-free errors.

That's the easy way.  Or for future generality you could define a copy instead, doing the work of separate allocation -- in fact I did need that for this class, in my experimental branch to parallelize it.

And you could define a move constructor and assignment for completeness, or not -- it's not necessary yet.

Paul-Licameli: In fact, it was already true because of the `mSetup` member, that copy and assignment should have been `delete`d even before this.  I overlooked it in earlier review.

libraries/lib-music-information-retrieval/PowerSpectrumGetter.cpp
```diff
@@ -17,22 +17,34 @@ namespace MIR
 PowerSpectrumGetter::PowerSpectrumGetter(int fftSize)
     : mFftSize { fftSize }
     , mSetup { pffft_new_setup(fftSize, PFFFT_REAL) }
-    , mWork(fftSize)
+    , mData { reinterpret_cast<float*>(
+         pffft_aligned_malloc(fftSize * sizeof(float))) }
+    , mWork { reinterpret_cast<float*>(
+         pffft_aligned_malloc(fftSize * sizeof(float))) }
```
Paul-Licameli, line 23: So I understand that this change in the construction was necessary to avoid the crash -- the allocation in the `vector<float>` construction was not guaranteed to be aligned.

But you do not make similar changes in the unit tests.

Could the unit tests have intermittent crashes too?

Paul-Licameli:
It seems to me, there ought to be a simpler way to do this, by providing a non-default allocator as the second template argument of `vector<float, ...>` rather than defaulting it. and that allocator would guarantee alignment.

Surprisingly I can't find a convenient allocator already defined in the C++ standard.

Paul-Licameli:
Maybe the alternative would not be hard to write
https://en.cppreference.com/w/cpp/memory/allocator/allocate
Just define your own allocator inheriting `std::allocator<float>` but redefining `allocate` to use the aligned malloc and free that the pffft library wants.

Maybe that alternative makes the whole thing easier to write.

But where there was `std::vector<float>` before, there would have to be `std::vector<float, PffftAllocator>` which would be a distinct type.

Do that or not, but still I want to see the copy of the spectrum getter deleted, before I approve this.

libraries/lib-music-information-retrieval/PowerSpectrumGetter.cpp
saintmatthieu: PFFFT isn't exposed by the library, so no changes are needed in the UTs.
I haven't experience a crash in the unit tests. It appears to happen seldom. So far I've only been able to reproduce it intermittently with the wav file attached to the ticket.

saintmatthieu: When running the benchmarking test in debug, I actually fail the assert in pffft that my data is aligned. I normally run it in release, I guess that's why I haven't noticed it earlier.

saintmatthieu:
@Paul-Licameli I very much liked your idea of the custom allocator, and I tried it hard. But for some reason, my custom `allocate` and `deallocate` methods never get called, only those of the default vector allocator. If you'd like to take a look, here is a commit where I tried this: https://github.com/audacity/audacity/commit/f5b140d598cee6f9d9a6e86229d40a595e764888
So will keep it the way I have it now, just deleting copy and move constructors.

crsib:
This PR should be closed, ain't it?

PR #5919, opened 2024-02-06
Account for pitch ratio when evaluating time-and-pitch delay

### LINKED ISSUE #5905: Improve delay compensation of pitch shifting
### Your idea

https://github.com/audacity/audacity/issues/5206 improved the delay estimation when using time stretching. The problem is back with pitch shifting. Something seemingly was omitted.

### Problem to be solved

Time-alignment of clips is important for synchronization and also to reduce the risk of discontinuities when looping a pitch-shifted clip.

### Prior art

_No response_

### Additional context

_No response_

saintmatthieu:
Resolves: #5905 

Recommended:

QA:
- [x] pitch-shifting and time-stretching clips doesn't introduce delay, whether positive or negative. (If it is not obvious aurally, you may render the clip to observe the delay.)

### DIFF
--- a/libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
+++ b/libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
@@ -46,6 +46,7 @@ StaffPadTimeAndPitch::StaffPadTimeAndPitch(
     , mSampleRate(sampleRate)
     , mNumChannels(numChannels)
     , mTimeRatio(parameters.timeRatio.value_or(1.))
+    , mPitchRatio(parameters.pitchRatio.value_or(1.))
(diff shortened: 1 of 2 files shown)


saintmatthieu:
3885e7c33e9d76683905ca852c7b1b2ef2119263 seems to fix well the problem:
image

Tracks 1 and 4 show an impulse centered at 1s.
Tracks 2 and 3 show octave-up-shifted rendered tracks before and after this change.
Tracks 5 and 6 show octave-down-shifted rendered tracks before and after this change.
In both directions, the impulse response is nicely centered after the fix.
(That the single impulse has pre- and post-echoes is a known characteristic of the algorithm.)

crsib (approved):

Issue #5693, opened 2023-11-27
Rewrite WaveTrack::DoOnProjectTempoChange


Paul-Licameli:
May depend on
- #5652

chinakov:
Tested. Thanks!

PR #5921, opened 2024-02-06
Stretch ratio nonsense

Paul-Licameli:
Resolves: #5693

WaveTrack::DoOnProjectTempoChange becomes interval-major.

More than that:  there is library restructuring to remove all mention of "tempo" in lib-track.

Recommended:


saintmatthieu (approved):
This PR also includes the colour-index change, but it's not mentioned in the description. (I reviewed that part as well.)
QA checklist will be long, will add what I can think of.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -199,10 +199,20 @@ class WAVE_TRACK_API WaveChannelInterval final
    const size_t miChannel;
 };
 
+struct WaveTrackMessage {
```
saintmatthieu, line 200: I don't expect this to convey more than just the insertion of a new clip, so I think something like `WaveTrackNewClipMessage` would be more informative.

libraries/lib-wave-track/WaveTrack.h
```diff
@@ -1249,7 +1278,7 @@ class WAVE_TRACK_API WaveTrack final
     @param backup whether the duplication is for backup purposes while opening
     a project, instead of other editing operations
     */
-   bool InsertClip(WaveClipHolder clip, bool backup = false);
+   bool InsertClip(WaveClipHolder clip, bool newClip, bool backup = false);
```
saintmatthieu, line 1281: IIUC, this is to know whether the clip's colour should be overwritten, and this depends on whether it's the copy of a clip from another track or a freshly created clip.
If I'm right, I'd much prefer it to be called `fromOtherTrack` or similar. It denotes precisely the intention, and I don't really know what "new clip" means anyways.

libraries/lib-wave-track/WaveTrack.h
Paul-Licameli: Ignore this commit

Paul-Licameli:
Rebased without changes

Paul-Licameli:
Rebased again

saintmatthieu (approved):

Issue #5932, opened 2024-02-09
Move the Timeline sub menu from the Extra menu to the View sub menu


DavidBailes:
### Your idea

As in title. See discussion: https://github.com/audacity/audacity/discussions/5931

### Problem to be solved

So the the Timeline menu is easier to Find

DavidBailes:
@LWinterberg . Please could this be added to the 3.5 release.

petersampsonaudacity:
Tested on W10 with @DavidBailes  latest branch build for this  audacity-win-3.5.0-alpha-20240212+3f4bc06-x64

All seems good:

![image](https://github.com/audacity/audacity/assets/50205984/88497273-ee2b-40ac-9a9e-2d4b494cf791)

![image](https://github.com/audacity/audacity/assets/50205984/4e5645d4-8024-4d67-9162-017d8a0857e2)

dozzzzer:
Tested

petersampsonaudacity:
I've updated the 3.5.0 alpha manual for this - a LOT of menu images (Extra menu commands) had to be changed.

PR #5943, opened 2024-02-12
Move the Timeline sub menu

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/5932

Problem:
Timeline sub menu isn't easy to find in the extra menu, which isn't shown by default, and the timeline view has increased importance in the upcoming 3.5. release.

Fix:
Move the timeline sub menu to the View menu.
The only available access key for Timeline on the View menu was the letter l. Given that is not a desirable access key due to the shortness of the underline indicating the access key, the access key "e" was chosen, and the access key of another item changed.

Recommended:

### DIFF
--- a/src/menus/TimelineMenus.cpp
+++ b/src/menus/TimelineMenus.cpp
@@ -48,7 +48,7 @@ using namespace MenuRegistry;
 auto ExtraSelectionMenu()
 {
    static auto menu = std::shared_ptr{ Menu(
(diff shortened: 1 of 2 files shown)


LWinterberg (approved):

petersampsonaudacity:
Tested on W10 with @DavidBailes  latest branch build for this.

All seems good:

image

image

PR #5950, opened 2024-02-13
Fix misaligned pffft

Paul-Licameli:
Resolves: #5906

Alternative fix for misalignment of buffers passed to pffft uses a non-default
allocator in std::vector.

Also PffftFloats is a pointer, always to the start of aligned span of floats (but agnostic
about the memory management).

Also moves FFT related files out of lib-math.

Recommended:


Paul-Licameli:
There was some discussion in another channel.  @crsib give an opinion.

@saintmatthieu doesn't trust this because the allocation and destruction functions weren't hit in the debugger.  But what I wrote was not what was compiled in his build.

For unknown reasons my code compiles in CI for all Windows build but did not compile on Matt's machine.

crsib (approved):
I don't see anything wrong with this code and I don't want to be picky about the names

saintmatthieu:
No I do get the allocators hit.
I can't build in `Debug`, though, compilation errors. @crsib

saintmatthieu:
> Now I do get the allocators hit. I can't build in `Debug`, though, compilation errors, have to use `RelWithDebugInfo`. Is the debugging experience really as good as in `Debug` ? @crsib

Paul fixed this.

Paul-Licameli:
I think a redefined allocator and a a neater solution, but there is more to it than a redefined deleted for unique_ptr.  I’m less familiar with this technique and muffed my first attempt a little!

libraries/lib-music-information-retrieval/PowerSpectrumGetter.h
```diff
@@ -50,9 +51,154 @@ template<typename T> struct PffftAllocator
    }
 };
 
+//! An unsigned number of floats that can span two addresses that are well
+//! aligned for pffft
+struct PffftAlignedCount
+{
+   // see source for pffft_aligned_malloc
+   static constexpr size_t ByteAlignment = 64;
```
crsib: Is this constant valid, though? The library aligns memory regions to 64 bytes, but this is not required by the instruction sets it uses internally (SSE2/Neon).

Probably, it doesn't harm, so I won't block this PR for this reason.

crsib (approved):

Paul-Licameli:
> Is this constant valid, though? The library aligns memory regions to 64 bytes, but this is not required by the instruction sets it uses internally (SSE2/Neon).
> 
> Probably, it doesn't harm, so I won't block this PR for this reason.

Yes, that is the weak point, but at least the magic number 64 is in only once place, reproducing a detail hidden in the pffft sources.

Paul-Licameli:
Alternative fix for misalignment of buffers passed to pffft uses a non-default
allocator in std::vector.

Also PffftFloats is a pointer, always to the start of aligned span of floats (but agnostic
about the memory management).

Also moves FFT related files out of lib-math.

Issue #5639, opened 2023-11-17
Crash when locating FFmpeg library during export


dozzzzer:
### Bug description

When selecting "Custom FFmpeg export" in the Export dialog, if Audacity hasn't yet detected the FFmpeg location and asks for locating it, after doing so it crashes.

### Steps to reproduce

1. Have FFmpeg installed
2. Purge audacity.cfg
3. Launch Audacity
4. Grab some audio
5. Ctrl + Shift + E
6. Select Format: Custom FFmpeg Export
7. Observe: Locate FFmpeg dialog pops up
8. Click the Browse button and select the `avformat-59.dll`
9. Click Open
10. Click OK in the Locate FFmpeg dialog

### Expected behavior

No crash; Locate FFmpeg dialog gets closed.

### Actual behavior

A crash

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

JoeS209:
Adding a null check for mFFmpeg in GetSampleRateList() in ExportFFmpeg.cpp fixes the immediate problem of the crash.  The actual export then fails with an error until Audacity is exited and restarted.  Here is the code snippet with the addition of the null check:

```
         if (mFFmpeg) {
            mAVCodec = mFFmpeg->CreateEncoder(codecId.c_str());
         }
```

dozzzzer:
Tested

PR #5953, opened 2024-02-14
Fix crash when locating FFmpeg library during export

JoeS209:
Resolves: #5639

1) Add null check to GetSampleRateList().
2) Ensure ffmpeg is loaded in FFmpegExporter class.

Recommended:


JoeS209:
Anyone?

LWinterberg (approved):
looks good to me. A quick test has it functions now, though I'll leave more thorough testing to QA.

Issue #5947, opened 2024-02-13
Export Multiple: After successful export, list of exported files is focused, instead of the action (OK) button


HighPriest:
### Steps to reproduce

1. Create a project which can make use of export multiple. E.g. Multiple tracks.
2. CTRL+SHIFT+E (Export Audio)
3. ENTER (with export multiple enabled)
4. ENTER... ENTER... ENTER... Nothing happens because the files list dialog is selected

### Expected behavior

I don't have to tab in the confirmation dialog to confirm it.

### Actual behavior

![image](https://github.com/audacity/audacity/assets/58490197/03d96308-7917-48ae-b46a-18fad500933e)

Mind the `|` cursor sign in the file list. OK button can be focused by pressing tab.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

petersampsonaudacity:
Testing shows that this is a regression on 1.3.11, introduced in the 1.3.12 release, released on 01Apr2010

_So we have had this for a very long time and this is the first time anybody has noticed this or logged it as an issue._

petersampsonaudacity:
Testing on W10 with @DavidBailes  latest branch for this fix:  audacity-win-3.5.0-alpha-20240214+eb85171-x64

This looks fixed to me 
a) after the multiple file export the focus is on the OK button
b) pressing the Enter button dismisses the dialog

![image](https://github.com/audacity/audacity/assets/50205984/9e984775-7ecc-426d-bdc3-60317452706f)

DavidBailes:
> This looks fixed to me
> a) after the multiple file export the focus is on the OK button
> b) pressing the Enter button dismisses the dialog

@petersampsonaudacity . Just to clarify:
a) after the the confirmation dialog opens, the focus is still on the text box (you can use up/down arrows). The blue outline around the OK button indicates that it's the default button.
b) pressing Enter now presses the default button (at least from the user's view of things).

dozzzzer:
Tested. Fixed for Windows as described.
On macOS, though, the existing behavior remains unchanged. A separate issue has been created: https://github.com/audacity/audacity/issues/5958

@DavidBailes

DavidBailes:
> Tested. Fixed for Windows as described. On macOS, though, the existing behavior remains unchanged. A separate issue has been created: #5958
> 
@dozzzzer .Despite the title of the original bug, I think that problem was that with the text box as the focus, pressing Enter did not press the default button, ie the OK button.
[...]

PR #5957, opened 2024-02-14
Export multiple confirmation dialog: make Enter close dialog

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/5947

Problem:
In this dialog, a read-only text box showing the files exported is the initial focus. A user would expect that pressing Enter would close the dialog without the need to first tab to the OK button, but this is not the case.

Fix:
Use the wxEVT_TEXT_ENTER event, to close the dialog when the text box is the focus, and the user presses Enter.

Recommended:


petersampsonaudacity:
Testing on W10 with @DavidBailes  latest branch for this fix:  audacity-win-3.5.0-alpha-20240214+eb85171-x64

This looks fixed to me 
a) after the multiple file export the focus is on the OK button
b) pressing the Enter button dismisses the dialog

crsib (approved):

Issue #5880, opened 2024-01-22
Use PFFFT for spectrogram calculation


saintmatthieu:
### Your idea

Working on #5794, I could see that PFFFT might be 10 times faster than our custom `RealFFTf` function. At least it was a game changer in that PR.
Maybe using it for spectrum display isn't too difficult and would make it smoother.

### Problem to be solved

Sometimes FFT spectrum display feels slow.

PR #5960, opened 2024-02-15
Use pffft in spectrogram

Paul-Licameli:
Resolves: #5880

Depends on
- #5950

Serial performance improvement in spectrogram drawing, using pffft

Recommended:

### DIFF
--- a/libraries/lib-fft/PffftTransformer.cpp
+++ b/libraries/lib-fft/PffftTransformer.cpp
@@ -6,6 +6,7 @@
   PffftTransformer.cpp
 
   Matthieu Hodgkinson
+  Paul Licameli
 
 **********************************************************************/
 #include "PffftTransformer.h"
--- a/libraries/lib-fft/PffftTransformer.cpp
+++ b/libraries/lib-fft/PffftTransformer.cpp
@@ -48,11 +48,13 @@ PffftTransformer::PffftTransformer(size_t N)
 
 void PffftTransformer::Reset()
 {
+   N = 0;
    reset();
 }
 
 void PffftTransformer::Reset(size_t N)
 {
+   this->N = N;
    reset(pffft_new_setup(N, PFFFT_REAL));
 }
 
--- a/libraries/lib-fft/PffftTransformer.cpp
+++ b/libraries/lib-fft/PffftTransformer.cpp
@@ -11,6 +11,7 @@
 **********************************************************************/
 #include "PffftTransformer.h"
 
+#include <algorithm>
 #include <cassert>
 #include <pffft.h>
 
--- a/libraries/lib-fft/PffftTransformer.cpp
+++ b/libraries/lib-fft/PffftTransformer.cpp
@@ -13,6 +13,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cstring>
 #include <pffft.h>
 
 void PffftSetupDeleter::Pffft_destroy_setup(PFFFT_Setup *p)
@@ -40,24 +41,58 @@ PffftConstFloats PffftFloatVector::aligned(PffftAlignedCount c) const
    return PffftConstFloats{ data() + c };
 }
 
+size_t PffftTransformer::MinSize()
+{
+   return pffft_min_fft_size(PFFFT_REAL);
+}
+
+static constexpr bool IsPowerOfTwo(size_t N)
+{
+   return (N > 0) && ((N & (N - 1)) == 0);
+}
+
+bool PffftTransformer::IsAllowedSize(size_t N)
+{
+   if (N < MinSize())
+      return IsPowerOfTwo(N);
+   while (0 == (N % 2))
+      N /= 2;
+   while (0 == (N % 3))
+      N /= 3;
+   while (0 == (N % 5))
+      N /= 5;
+   return (N == 1);
+}
+
+PffftAlignedCount PffftTransformer::PaddedCount(size_t fftLen)
+{
+   // The padding and the minimum size constraint happen to be the same
+   // in the present pffft implementation, but don't mention that in the
+   // header or rely on that elsewhere.
+   return PffftAlignedCount{ std::max<size_t>(
+      MinSize(), PffftAlignedCount{ fftLen }) };
+}
+
 PffftTransformer::PffftTransformer(size_t N)
    : std::unique_ptr<PFFFT_Setup, PffftSetupDeleter>{
-      pffft_new_setup(N, PFFFT_REAL)
+      pffft_new_setup(std::max(N, MinSize()), PFFFT_REAL)
    }
-   , N{ N }
+   , requestedSize{ N }
 {
+   assert(IsAllowedSize(N));
 }
 
 void PffftTransformer::Reset()
 {
-   N = 0;
+   requestedSize = 0;
    reset();
 }
 
 void PffftTransformer::Reset(size_t N)
 {
-   this->N = N;
-   reset(pffft_new_setup(N, PFFFT_REAL));
+   assert(IsAllowedSize(N));
+   this->requestedSize = N;
+   reset(pffft_new_setup(std::max(N, MinSize()), PFFFT_REAL));
 }
 
 void PffftTransformer::TransformOrdered(PffftConstFloats input,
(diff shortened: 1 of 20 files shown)


src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -22,33 +22,33 @@
...
-      const int index = hFFT->BitReversed[i];
+   for (size_t index = 2; index < len; index += 2) {
       const float re = buffer[index], im = buffer[index + 1];
       power = re * re + im * im;
       if(power <= 0)
-         out[i] = -160.0;
+         out[index / 2] = -160.0;
       else
-         out[i] = 10.0*log10f(power);
+         out[index / 2] = 10.0 * log10f(power);
```
saintmatthieu, line 51: Hey, maybe we could replace this log10f with an optimized log2. We saw in the tempo-estimation optimization PR that it sped up the STFT a great deal. (We probably should not blindly reuse the `GetLog2` of `MirDsp.cpp`, though, because that approximation is only accurate to two decimal places, as you correctly found out, @Paul-Licameli ).

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -369,8 +373,10 @@ void SpecCache::Populate(
    const size_t fftLen = windowSizeSetting * zeroPaddingFactorSetting;
    const auto nBins = settings.NBins();
 
+   // One work area for pffft, one work area for the outputs, two extra
+   // output areas to compute reassignment
    const size_t scratchSize =
-      reassignment ? 3u * PffftAlignedCount{ fftLen } : fftLen;
+      (reassignment ? 4u : 2u) * PffftAlignedCount{ fftLen };
```
saintmatthieu, line 379: We're here in `Populate`, and within that function it's difficult to find out why there's a need for more than just `fftLen`. One has to dig `CalculateOneSpectrum` to find out about that dependency. And in fact, it doesn't seem that the standard spectrogram view requires `scratch`.

saintmatthieu (commented):
I think it'd be a relatively high-impact change for our users, but unfortunately it is a lot to consider, all the more that there are several algorithms to review and test.

src/effects/EqualizationFilter.cpp
```diff
@@ -156,25 +157,23 @@ void EqualizationFilter::Filter(size_t len,
...
+      im = newValue.imag();
    }
    // Fs/2 component is purely real
-   scratch[1] = buffer[1] * mFilterFuncR[len/2];
+   buffer[1] *= mFilterFuncR[len/2];
 
-   // Inverse FFT and normalization
-   InverseRealFFTf(scratch, hFFT.get());
-   ReorderToTime(hFFT.get(), scratch, buffer);
+   transformer.InverseTransformOrdered(buf, buf, scr, true);
```
saintmatthieu, line 178: anonymous boolean

saintmatthieu (commented):
This is an important improvement and the type safety is good, but I'd like this ambiguity in the `PffftFloatVector` API to be resolved, at least by making that `aligned` overload private.

saintmatthieu (changes requested):
Indeed, @Paul-Licameli, using a spectrogram length of 8 in the settings causes an assertion in PFFFT.
Reading more carefully `pffft.h`:
```
   - supports only transforms for inputs of length N of the form
   N=(2^a)*(3^b)*(5^c), a >= 5, b >=0, c >= 0 (32, 48, 64, 96, 128,
   144, 160, etc are all acceptable lengths). Performance is best for
   128<=N<=8192.
```` 
This would imply that the least size is 32. (Maybe by luck, but I just tried a size of 16 and that worked.)

@LWinterberg do you have an idea of the amount of dissatisfaction raising the least spectrum size to 32 would cause? I never came across a need for such small window sizes in audio applications. Maybe analysis of non-audio waves sampled at extremely low rates ?

PS: I wasn't aware that non-power-of-two sizes were also possible - interesting!

**Comment by LWinterberg** [comment]
> @LWinterberg do you have an idea of the amount of dissatisfaction raising the least spectrum size to 32 would cause? I never came across a need for such small window sizes in audio applications. Maybe analysis of non-audio waves sampled at extremely low rates ?

After I didn't find a a spectrogram in promising places like [Multi-Beam Spatio-Spectral Beamforming Receiver for Wideband Phased Arrays](https://www.researchgate.net/publication/260502662_Multi-Beam_Spatio-Spectral_Beamforming_Receiver_for_Wideband_Phased_Arrays), [Pipeline FFT architectures optimized for FPGAs](https://www.researchgate.net/publication/26850326_Pipeline_FFT_architectures_optimized_for_FPGAs) and [Goldstein phase filtering on InSAR data](https://forum.step.esa.int/t/parameters-of-goldstein-phase-filtering/6423) (it's an interferogram), I checked Sonic Visualiser and found:

image

If even they don't need anything below 32, it's probably fair to say we don't either.

**Comment by Paul-Licameli** [comment]
> Indeed, @Paul-Licameli, using a spectrogram length of 8 in the settings causes an assertion in PFFFT. Reading more carefully `pffft.h`:
> 
> ```
>    - supports only transforms for inputs of length N of the form
>    N=(2^a)*(3^b)*(5^c), a >= 5, b >=0, c >= 0 (32, 48, 64, 96, 128,
>    144, 160, etc are all acceptable lengths). Performance is best for
>    128<=N<=8192.
> ```
> 
> This would imply that the least size is 32. (Maybe by luck, but I just tried a size of 16 and that worked.)
> 
> @LWinterberg do you have an idea of the amount of dissatisfaction raising the least spectrum size to 32 would cause? I never came across a need for such small window sizes in audio applications. Maybe analysis of non-audio waves sampled at extremely low rates ?
> 
> PS: I wasn't aware that non-power-of-two sizes were also possible - interesting!

Interesting, they allow other prime factors than 2.  See Lanczos Lemma. 

The divide and conquer step of fft.

File: src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp (hunk 1 above)
Comment (Paul-Licameli) [reply src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp:50]: No.

Don’t use approximation for speed. Plot the best spectral data we can calculate.

File: src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp (hunk 2 above)
Comment (Paul-Licameli) [reply src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp:378]: I think the comment before this line is sufficient explanation.

File: libraries/lib-fft/PffftTransformer.cpp [hunk 4]
```diff
@@ -3,12 +3,12 @@
 
   Audacity: A Digital Audio Editor
 
-  PowerSpectrumGetter.cpp
```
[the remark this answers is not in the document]
Paul-Licameli: I think the power spectrum is a closely enough related utility that it can live in the same file.

libraries/lib-fft/PowerSpectrumGetter.h
```diff
@@ -57,26 +57,26 @@ struct PffftAlignedCount
...
-   PffftAlignedCount(const PffftAlignedCount &) = default;
+   constexpr PffftAlignedCount(const PffftAlignedCount &) = default;
    PffftAlignedCount &operator =(const PffftAlignedCount &) = default;
 
    //! @invariant `result: result % FloatAlignment == 0`
-   operator size_t() const { return value; }
+   operator size_t() const noexcept { return value; }
 
 private:
    size_t value{};
```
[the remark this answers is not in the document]
Paul-Licameli: I was explicit in comments above the class:

```
 Usual hazards of pointer invalidation or out-of-bounds subscripting still
 apply.

 Size and capacity are NOT constrained to be aligned.
```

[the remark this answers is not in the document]
Paul-Licameli: I disagree that the one-argument `aligned` should be made private.

There may be future need.

src/effects/EqualizationFilter.cpp
Paul-Licameli: One easily jumps to the function definition to know what the boolean is, if the IDE browser is good enough.

libraries/lib-fft/PffftTransformer.cpp
```diff
@@ -6,6 +6,7 @@
   PffftTransformer.cpp
 
   Matthieu Hodgkinson
+  Paul Licameli
```
saintmatthieu, line 9: Funny character

libraries/lib-fft/PffftTransformer.h
```diff
@@ -68,7 +68,7 @@ struct PffftAlignedCount
    PffftAlignedCount &operator =(const PffftAlignedCount &) = default;
 
    //! @invariant `result: result % FloatAlignment == 0`
-   operator size_t() const noexcept { return value; }
+   constexpr operator size_t() const noexcept { return value; }
```
saintmatthieu, line 71: `value` could/should still be `const`.

saintmatthieu (changes requested):
Unsatisfied to see argued suggestions rejected for no apparent good reason, but the remaining time for completion of this PR doesn't allow further argument. Please raise the least allowed frame size to 32 and I'll approve.

libraries/lib-fft/PffftTransformer.cpp
```diff
@@ -67,22 +102,42 @@ void PffftTransformer::TransformOrdered(PffftConstFloats input,
...
+      memset(pInput, '\0', (N - requestedSize) * sizeof(float));
+   }
    pffft_transform_ordered(get(), input.get(), output.get(), work.get(),
       PFFFT_FORWARD);
+   if (requestedSize < N) {
+      // Now decimate
+      auto pOutput = output.get();
+      const auto factor = N / requestedSize;
+      for (size_t ii = 1, nn = requestedSize; ii < nn; ++ii)
+         pOutput[ii] = pOutput[ii * factor];
```
saintmatthieu, line 122: You overlooked that `output` here has layout `[DC, Nyq, real1, imag1, ...]`.
But besides that I think it should work.

Issue #4247, opened 2023-01-28
Memory leaks while recording on Wayland


kuba-orlik:
### Bug description

When recording in Audacity in Gnome on Wayland, while the waveform is zoomed in to a scale where around 10s of recording fit horizontally on the screen, the memory usage goes drastically up. Zooming out the waveform or minimizing the Audacity window stops the growth of memory usage, but the already used up memory is released only after closing Audacity. The same issue does not occur if audacity is ran as `env GDK_BACKEND=x11 audacity` under the same kernel and Audacity version.

### Steps to reproduce

1. Open Audacity in Wayland and your preferred resource usage monitor (like gnome-monitor or `watch free -h`
2. Start recording audio
3. Wait around 10 minutes and observe memory usage

### Expected behavior

Memory usage should remain more or less constant

### Actual behavior

The memory usage of Audacity is ramping up. Eventually I ran out of RAM and the PC froze. The growth pauses when Audacity is minimized or when the timeline view is zoomed out.

### Audacity Version

stable version (from a Linux repository)

### Operating system

Linux

### Additional context

The drop of RAM usage after closing audacity that's been leaking memory:

![image](https://user-images.githubusercontent.com/2697916/215258590-123e82e0-c429-473b-93ba-b5d887a27b47.png)
![image](https://user-images.githubusercontent.com/2697916/215258599-1705a731-439d-408d-9cea-7d9d8be59ae2.png)

The buff/cache metric is the one that keeps growing. The facts that minimizing the window pauses the memory usage growth and that running it in xorg fixes the issue seem to point to the GUI as the culprit

dvzrv:
Hi! This has been broken for months apparently. There is a downstream ticket with Arch Linux as well at https://bugs.archlinux.org/task/77335

It would be really great if this could be addressed. Is this still a problem with 3.3.1?
If so, can the env please be added to the XDG desktop integration file until you have this figured out? Thanks! :)

crsib:
Wayland is not an option for wx based Audacity anyway. Attempt to fix that resulted in 25k lines, so I don't think that we will fix any issues in Audacity 3.x timeline.

dvzrv:
> Wayland is not an option for wx based Audacity anyway.

Okay. Can you please disable it in the XDG desktop integration then as suggested in the original ticket? That would help managing expectations and lead to fewer bug reports I think.

crsib:
I'm not quite sure what you mean

dvzrv:
[...]

Issue #459, opened 2020-03-18
Timeline time tracker not updating with playback [GNOME Wayland]


docquantum:
**Describe the bug**
Timeline vertical bar tracker does not update position automatically and requires me to click the waveform to update it's visual location. The audio still plays as expected.

**To Reproduce**
1. Open a audio file or record one.
2. Play the file and watch the vertical line that follows the playback position.
3. It will not update, clicking anywhere on the waveform as it plays back will update the tracker position.

**Expected behavior**
The playback tracker will update automatically as the audio is played back.

**Screenshots**
![animation](https://user-images.githubusercontent.com/3020563/77002306-4aef6500-6929-11ea-9618-015e99b2b015.gif)

**Additional information (please complete the following information):**
 - Arch Linux 5.5.9-1-clear using GNOME Shell 3.36.0 in Wayland (Also existed in older version of GNOME 3 going back to 3.32 at least [in Wayland].)
 - Audacity 2.3.3
 - ALSA

**Additional context**
This does not occur when running GNOME on Xorg.

imsedgar:
It also occurs in Fedora 31 with Gnome on Wayland with the following packages installed:

audacity-freeworld-2.3.3-1.fc31.x86_64
gnome-shell-3.34.5-1.fc31.x86_64
gnome-session-wayland-session-3.34.2-2.fc31.x86_64
libwayland-client-1.17.0-2.fc31.x86_64

There is a downstream bug report on RPMFusion: https://bugzilla.rpmfusion.org/show_bug.cgi?id=5551

lysiasthefox:
I can confirm the missing playback cursor under Wayland (GNOME Shell 3.34.5 - Fedora 31), using Xorg/X11 mitigates the issue.

Audacity 2.3.3
Linux XXXXX 5.6.16-200.fc31.x86_64

SteveDaulton:
Audacity currently requires GTK2 and Wayland is not yet supported.
Building with GTK3  or running on Wayland are both likely to cause problems.

lllucius:
What Steve says is still true...GTK2 is still preferred for now.  But, you might want to try the recently released 2.4.2.  You might find it works a tad bit better with GTK3 and Wayland.  Let us know how it goes if you try it.

aaronjohnsabu1999:
This happened to me using GNOME on Ubuntu 20.04.1 LTS

[...]

PR #5977, opened 2024-02-17
Resolve memory leak during Wayland recording using Xwayland #4247

tkna91:
Add GDK_BACKEND=x11 to Exec env

Resolves: https://github.com/audacity/audacity/issues/4247

Add GDK_BACKEND=x11 to Exec env
Resolve memory leak during Wayland recording using Xwayland 

Recommended:

### DIFF
--- a/src/audacity.desktop.in
+++ b/src/audacity.desktop.in
@@ -61,7 +61,7 @@ Categories=AudioVideo;Audio;AudioVideoEditing;
 
 Keywords=sound;music editing;voice channel;frequency;modulation;audio trim;clipping;noise reduction;multi track audio editor;edit;mixing;WAV;AIFF;FLAC;MP2;MP3;
 
-Exec=env UBUNTU_MENUPROXY=0 @AUDACITY_NAME@ %F
+Exec=env GDK_BACKEND=x11 UBUNTU_MENUPROXY=0 @AUDACITY_NAME@ %F
 StartupNotify=false
 Terminal=false
 MimeType=@MIMETYPES@;


LWinterberg:
Do we know that the x11 environment/xwayland is always available? If systems without exist, I believe this change would break Audacity on them. 

I see in the [docs](https://developer-old.gnome.org/gtk4/stable/gtk-running.html#id-1.9.4.2.13) that there'd potentially an easy fix to that: 

> This environment variable can contain a comma-separated list of backend names, which are tried in order. The list may also contain a *, which means: try all remaining backends.

so `GDK_BACKEND=x11,*` would probably make this safe everywhere.

tkna91:
Sorry, I was including a link to the gtk3 documentation.
If the next Audacity will use gtk4, it would be better to include a link to the following documentation
https://docs.gtk.org/gtk4/running.html#gdk_backend

LWinterberg (approved):

LWinterberg:
ah, actually, it doesn't work. Audacity uses wxwidgets, and for compatibility reasons we build with GTK2. I see [the * thing goes back at least 10 years](https://gitlab.gnome.org/GNOME/gtk/-/commit/632ee007a46dfb869f8b2f2598c54f9d19f06cdb), but that's not enough to reach GTK2 apparently. Please do an interactive rebase and drop the commit with my suggestion, then force-push.

Anyway, this means we actually have to answer the question: Do we know that the x11 environment/xwayland is always available for anything that uses the .desktop file? I will research that.

LWinterberg:
Okay, so: 

* RedHat is dropping the xorg server, but explicitly keeping the x11 environment (xwayland) around. 
* BSD seems to still run xorg most of the time.
* Online forks of Audacity may or may not use broadway, but given that online services only need the maintainer to update this (and not every individual end user), it'd be fine.
* Gnome 40+ has an xwayland-on-demand thing which would launch xwayland only when an app requests it (which we'd do here)
* Outside of a few arch users, I have not seen any attempts to ditch xwayland by anyone.

Given that (and given that Mac and Windows don't care if you break the .desktop file), I think it's safe to assume that x11 is present in some capacity.

imciner2:
Fedora has already been carrying a patch adding this for at least the last 3 or so years: https://src.fedoraproject.org/rpms/audacity/blob/rawhide/f/gdk_x11_backend.patch

tkna91:
@LWinterberg That is, `GDK_BACKEND=x11,*` is not good and `GDK_BACKEND=x11` is good. right?
But I am having a hard time understanding why.
Are you saying that the `x11,*` values are not supported in GTK 3.0-3.9.x and may cause malfunctions?

- https://developer-old.gnome.org/gtk2/2.24/gtk-running.html
- https://docs.gtk.org/gtk3/running.html#environment-variables
- https://docs.gtk.org/gtk4/running.html#gdk_backend

tkna91:
@LWinterberg No, no. You simply want me to organize my commits in the form `GDK_BACKEND=x11,*`?

LWinterberg:
> @LWinterberg That is, GDK_BACKEND=x11,* is not good and GDK_BACKEND=x11 is good. right?

correct. Simply drop my suggested change, you had it right the first time. 

> But I am having a hard time understanding why.

It seems like old GTK does an exact string comparison, while new GTK does a substring comparison.

> Fedora has already been carrying a patch adding this for at least the last 3 or so years

That's good to know. I imagine you're not aware of anyone running into problems because x11 is missing either then, correct? @imciner2

imciner2:
> That's good to know. I imagine you're not aware of anyone running into problems because x11 is missing either then, correct? @imciner2

No, we've never had any problems reported by any users when using this. We first introduced this to fix https://github.com/audacity/audacity/issues/459, actually.

PR #5979, opened 2024-02-19
Use aligned vector in GetNormalizedCircularAutocorr, too.

### LINKED ISSUE #5906: Occasional crash in tempo estimation

saintmatthieu:
Resolves: #5906 

A use of PFFFT was forgotten in #5950

Recommended:

### DIFF
--- a/libraries/lib-music-information-retrieval/MirDsp.cpp
+++ b/libraries/lib-music-information-retrieval/MirDsp.cpp
@@ -84,14 +84,16 @@ std::vector<float> GetMovingAverage(const std::vector<float>& x, double hopRate)
 }
 } // namespace
 
-std::vector<float> GetNormalizedCircularAutocorr(std::vector<float> x)
+std::vector<float>
+GetNormalizedCircularAutocorr(const std::vector<float>& ux /* unaligned x*/)
 {
-   if (std::all_of(x.begin(), x.end(), [](float x) { return x == 0.f; }))
-      return x;
-   const auto N = x.size();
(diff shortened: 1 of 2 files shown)


libraries/lib-music-information-retrieval/MirDsp.cpp
```diff
@@ -84,14 +84,16 @@ std::vector<float> GetMovingAverage(const std::vector<float>& x, double hopRate)
 }
 } // namespace
 
-std::vector<float> GetNormalizedCircularAutocorr(std::vector<float> x)
+std::vector<float>
+GetNormalizedCircularAutocorr(std::vector<float> ux /* unaligned x*/)
```
crsib, line 88: I think, but I might be wrong, that const ref should be used now in this place. `x` is no longer used in place, and the ownership of `x` is not transferred on the main branch

crsib (approved):
I have just one, quite minor concern. It's not blocking for sure

saintmatthieu:
> I have just one, quite minor concern. It's not blocking for sure

Addressed. Squashed and hard-pushed.

Issue #3865, opened 2022-10-27
Add search option in Plugin Manager


dozzzzer:
### Your idea

Adding an option to search by the plugin's name in the Plugin Manager dialog.

Adding an option to specify an additional path to scan Plugins

<img width="700" alt="image" src="https://user-images.githubusercontent.com/70063970/198307851-3cd5ab1b-98e0-4599-a7ca-55aae0db399b.png">

### Problem to be solved

Will make the navigation in the list of plugins easier, especially when having dozens or even hundreds of them.

petersampsonaudacity:
@dozzzzer  you can already type the first letter or first few letters of the plugin you want to access in the plugin manager

dozzzzer:
Indeed. What a discovery! Thanks for pointing out, @petersampsonaudacity

petersampsonaudacity:
> @dozzzzer you can already type the first letter or first few letters of the plugin you want to access in the plugin manager

Hopefully one should be able to type the middle or end part of the plugin and not just be restricted to searching based on the initial characters.

petersampsonaudacity:
I added Tip divs to the Plugin Manager and the Macro Manager pages in the alpha manual for these selection Easter-Eggs
a) https://alphamanual.audacityteam.org/man/Manage_Effects,_Generators_and_Analyzers#Enabling_and_disabling
b) https://alphamanual.audacityteam.org/man/Manage_Macros#insert

DilsonsPickles:
> @LWinterberg @jessjwilliamson @DilsonsPickles guys, feel free to work out the UI/UX

@dozzzzer do we currently have any search bars being used in WX?

dozzzzer:
@DilsonsPickles No, I don't think so

DavidBailes:
> do we currently have any search bars being used in WX?

I'm not sure I fully understand the question, but in the keyboard category of preferences, there's a search box.

DilsonsPickles:
![PluginSearch_00](https://user-images.githubusercontent.com/111350316/199263062-a38c0e75-a3c6-44f8-8cf5-51577340d594.png)
![PluginSearch_01](https://user-images.githubusercontent.com/111350316/199263158-d5d6c0f0-89df-4663-9e83-05cfce5623f3.png)
![PluginSearch_02](https://user-images.githubusercontent.com/111350316/199263183-7d257f0a-cb92-4cc2-837f-946e36575fed.png)
[...]

PR #5980, opened 2024-02-19
Plugin manager window update

vsverchinsky:
Resolves: #3865
Resolves: #6031

Recommended:

QA: 
 - [x] Group toggling with mouse click does not work as in spec, though it works via the spacebar. 
 - [ ] Rescanning works as before.
 - [ ] Behaviour when rescanning finds an incompatible plugin is the same as before.

### DIFF
--- a/src/PluginDataModel.h
+++ b/src/PluginDataModel.h
@@ -1,0 +1,82 @@
+/*!*********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  @file PluginDataModel.h
+
+  @author Vitaly Sverchinsky
+
+**********************************************************************/
+#pragma once
+
+#include <wx/dataview.h>
+
+#include "PluginDescriptor.h"
+
+class PluginDescriptor;
+
+///!brief A plugins list model that can be attached to wxDataViewCtrl
+class PluginDataModel final : public wxDataViewModel
+{
+   static const void* RowIdRoot;
+
+   std::vector<std::pair<PluginDescriptor*, bool>> mPluginStateModel;
+   std::vector<void*> mIndexFilterMap;
+
+   wxString mFilterType;
+   wxString mFilterExpr;
+   int mFilterState{-1};
+   int mFilterCategory{-1};
+
+public:
+
+   enum Column
+   {
+      ColumnName,
+      ColumnType,
+      ColumnPath,
+      ColumnState,
+      ColumnNum
+   };
+
+   PluginDataModel(int filterEffectCategory = -1,
+                   int filterState = -1,
+                   const wxString& filterType = {},
+                   const wxString& filterExpr = {});
+
+   wxDataViewItem GetItemForRow(uint32_t row) const;
+   bool GetRowForItem(const wxDataViewItem& item, uint32_t& row) const;
+   uint32_t GetRowCount() const;
+
+   bool IsEnabled(const wxDataViewItem&, unsigned) const override;
+   bool HasContainerColumns(const wxDataViewItem& item) const override;
+
+   void SetFilterType(const wxString& type);
+   wxString GetFilterType() const;
+   void SetFilterState(int state);
+   int GetFilterState() const;
+   void SetFilterExpr(const wxString& expr);
+   wxString GetFilterExpr() const;
+   void SetFilterCategory(int category);
+   int GetFilterCategory() const;
+
+   int Compare(const wxDataViewItem& item1, const wxDataViewItem& item2, unsigned column, bool ascending) const override;
+
+   void ApplyChanges(const std::function<bool(int, int, const wxString&)>& progressUpdateFn,
+                      const std::function<void(const TranslatableString&)>& errorFn);
+
+   unsigned GetColumnCount() const override;
+   wxString GetColumnType(unsigned col) const override;
+   void GetValue(wxVariant& variant, const wxDataViewItem& item, unsigned col) const override;
+   bool SetValue(const wxVariant& variant, const wxDataViewItem& item, unsigned col) override;
+   wxDataViewItem GetParent(const wxDataViewItem& item) const override;
+   bool IsContainer(const wxDataViewItem& item) const override;
+   unsigned GetChildren(const wxDataViewItem& item, wxDataViewItemArray& children) const override;
+
+   PluginDescriptor* GetPlugin(wxDataViewItem item);
+
+   bool IsListModel() const override { return false; }
+private:
+   void UpdateFilter();
+   bool IsFilterEmpty() const;
+};
--- a/src/PluginDataViewCtrl.cpp
+++ b/src/PluginDataViewCtrl.cpp
@@ -1,0 +1,662 @@
+#include "PluginDataViewCtrl.h"
+
+#include <wx/renderer.h>
+
+#include "PluginDataModel.h"
+
+BEGIN_EVENT_TABLE(PluginDataViewCtrl, wxDataViewCtrl)
+   EVT_CHAR(PluginDataViewCtrl::OnCharEvent)
+END_EVENT_TABLE()
+
+#if defined(wxHAS_GENERIC_DATAVIEWCTRL) || defined(__WXGTK__)
+class PluginDataViewRenderer : public wxDataViewCustomRenderer
+{
+public:
+   using wxDataViewCustomRenderer::wxDataViewCustomRenderer;
+
+   bool ActivateCell(const wxRect& cell,
+                     wxDataViewModel* model,
+                     const wxDataViewItem& item,
+                     unsigned col,
+                     const wxMouseEvent* mouseEvent) final
+   {
+      if(mouseEvent == nullptr)
+      {
+         wxCHECK(GetView(), false);
+
+         wxVariant value;
+         model->GetValue(value, item, PluginDataModel::ColumnState);
+         value = !value.GetBool();
+
+         wxDataViewItemArray sel;
+         GetView()->GetSelections(sel);
+         if(!sel.empty())
+         {
+            for(const auto& other : sel)
+               model->ChangeValue(value, other, PluginDataModel::ColumnState);
+         }
+         else
+            model->ChangeValue(value, item, PluginDataModel::ColumnState);
+
+   #if wxUSE_ACCESSIBILITY
+         if(const auto ctrl = dynamic_cast<PluginDataViewCtrl*>(GetView()))
+         {
+            wxAccessible::NotifyEvent(
+               wxACC_EVENT_OBJECT_NAMECHANGE,
+               ctrl,
+               wxOBJID_CLIENT,
+               ctrl->GetRowByItem(item) + 1);
+         }
+   #endif
+         return true;
+      }
+      return OnCellClicked(cell, model, item, col, mouseEvent);
+   }
+
+protected:
+   virtual bool OnCellClicked(const wxRect& cell,
+                       wxDataViewModel* model,
+                       const wxDataViewItem& item,
+                       unsigned col,
+                       const wxMouseEvent* mouseEvent)
+   {
+      return false;
+   }
+};
+
+class PluginDataViewTextRenderer final : public PluginDataViewRenderer
+{
+   wxString mText;
+public:
+
+   PluginDataViewTextRenderer()
+      : PluginDataViewRenderer("string", wxDATAVIEW_CELL_ACTIVATABLE)
+   {
+      
+   }
+
+   bool SetValue(const wxVariant& value) override
+   {
+      mText = value.GetString();
+      return true;
+   }
+
+   bool GetValue(wxVariant&) const override
+   {
+      return false;
+   }
+
+   bool Render(wxRect cell, wxDC* dc, int state) override
+   {
+      if(!mText.empty())
+         RenderText(mText, 0, cell, dc, state);
+      return true;
+   }
+
+   wxSize GetSize() const override
+   {
+      if(!mText.empty())
+         return GetTextExtent(mText);
+      return GetView()->FromDIP(wxSize(wxDVC_DEFAULT_RENDERER_SIZE,
+                                          wxDVC_DEFAULT_RENDERER_SIZE));
+   }
+   
+#if wxUSE_ACCESSIBILITY
+   wxString GetAccessibleDescription() const override
+   {
+      return mText;
+   }
+#endif
+};
+
+// Activated cell also toggles all other cells that are currently selected
+class PluginDataViewStateRenderer final : public PluginDataViewRenderer
+{
+   wxVariant mValue;
+public:
+
+   PluginDataViewStateRenderer()
+      : PluginDataViewRenderer("bool", wxDATAVIEW_CELL_ACTIVATABLE)
+   { }
+
+   bool SetValue(const wxVariant& value) override
+   {
+      mValue = value;
+      return true;
+   }
+   bool GetValue(wxVariant&) const override
+   {
+      return false;
+   }
+   bool Render(wxRect cell, wxDC* dc, int state) override
+   {
+      if(mValue.IsNull())
+         return false;
+      //Duplicated from wxDataViewToggleRenderer
+      int flags = 0;
+      if (mValue.GetBool())
+         flags |= wxCONTROL_CHECKED;
+      if (GetMode() != wxDATAVIEW_CELL_ACTIVATABLE ||
+         !(GetOwner()->GetOwner()->IsEnabled() && GetEnabled()))
+      {
+         flags |= wxCONTROL_DISABLED;
+      }
+      wxSize size = cell.GetSize();
+      size.IncTo(GetSize());
+      cell.SetSize(size);
+
+      wxRendererNative& renderer = wxRendererNative::Get();
+      wxWindow* const win = GetOwner()->GetOwner();
+      renderer.DrawCheckBox(win, *dc, cell, flags);
+      return true;
+   }
+
+   wxSize GetSize() const override
+   {
+      return wxRendererNative::Get().GetCheckBoxSize(GetView());
+   }
+   
+#if wxUSE_ACCESSIBILITY
+   wxString GetAccessibleDescription() const override
+   {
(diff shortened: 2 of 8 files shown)


petersampsonaudacity:
@vsverchinsky 

You have a small typo Vitaly - it should say Di**s**abled

image

vsverchinsky:
Last branch update aims to fix compilation and keyboard navigation issues on platforms other than win. Remaining accessibility problems are yet to be fixed

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch build:   audacity-win-3.5.0-alpha-20240227+067c5e1-x64

The list of EGATS is still NOT in strict alphabetical order - rather it has alphabetically ordered subsections.

a) This makes it harder to use.

b) It is not really necessary as the user can filter by Type to see the subsections.

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch build: audacity-win-3.5.0-alpha-20240227+95f0559-x64

That looks better to me now with proper alphabetic ordering.

image

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch build for this:  audacity-win-3.5.0-alpha-20240229+0df2b93-x64

I see we now have an additional filter for Category (type of EGAT)

The cute thing is that the default setting you get for Category depends on the way you get to the Plugin Manager, so coming for Effect>Plugin Manager will default to Category Effect and similar for the other EGAT category.

src/PluginRegistrationDialog.cpp
```diff
@@ -918,151 +327,59 @@ void PluginRegistrationDialog::OnRescan(wxCommandEvent& WXUNUSED(evt))
...
-      if (!failedPlugins.empty())
+      if(self)
       {
-         auto dialog = safenew IncompatiblePluginsDialog(this, wxID_ANY, ScanType::Manual, failedPlugins);
-         dialog->ShowModal();
+         self->ReloadModel();
+         if (!failedPlugins.empty())
+         {
+            auto dialog = safenew IncompatiblePluginsDialog(self.get(), wxID_ANY, ScanType::Manual, failedPlugins);
+            dialog->ShowModal();
```
saintmatthieu, line 336: I noticed (and think I remember someone else mentioning this in the ticket, maybe Peter) that once this modal shows, the visible plugin items become greyed out:
image

I don't know why or how to fix this, but it probably would be around those lines.

saintmatthieu (dismissed):
Not done reviewing, but I'd like to get answers about `wxHAS_GENERIC_DATAVIEWCTRL` before completing it, as a lot of code depends on it.
Otherwise, requesting changes because of the items-greyed-out little bug.

src/PluginsDataViewCtrl.h
```diff
@@ -0,0 +1,39 @@
...
+
+class PluginsDataViewCtrl final : public wxDataViewCtrl
+{
+   friend class PluginsDataViewCtrlAx;
+   friend class PluginDataViewRenderer;
+
+   void Init();
+public:
+   template<typename... Args>
+   PluginsDataViewCtrl(Args&&... args)
```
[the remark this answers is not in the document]
vsverchinsky: Yes 😅

src/PluginsDataViewCtrl.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+#include <wx/renderer.h>
+
+#include "PluginDataModel.h"
+
+BEGIN_EVENT_TABLE(PluginsDataViewCtrl, wxDataViewCtrl)
+   EVT_CHAR(PluginsDataViewCtrl::OnCharEvent)
+END_EVENT_TABLE()
+
+#if defined(wxHAS_GENERIC_DATAVIEWCTRL) || defined(__WXGTK__)
```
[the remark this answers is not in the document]
vsverchinsky: > Ideally this #if would be false, so that e.g.

Ideally, yes. But `wxWidgets` controls are not always have same behavior across platforms. There are three different underlying types depending on target platform: generic on Windows (custom implementation that doesn't use high level OSs controls), gtk tree view on Linux-based platforms, and `NSOutlineView` on MacOS. They are all react differently on user input, and not exactly as we want it to. 
`wxDataView*Renderer` implementations are also platform-dependent, and class hierarchies they belong to are also different depending on platform.

> Are your custom implementations inferior

No doubt 🤣 

>  If not, which not just keeping them and remove the #if ?

I think I answered that question?

src/PluginsDataViewCtrl.cpp
```diff
@@ -0,0 +1,649 @@
...
+public:
+   PluginDataViewTextRenderer()
+      : wxDataViewTextRenderer(GetDefaultType(), wxDATAVIEW_CELL_INERT, wxDVR_DEFAULT_ALIGNMENT)
+   { }
+};
+
+#endif
+
+#if defined(wxHAS_GENERIC_DATAVIEWCTRL) && wxUSE_ACCESSIBILITY
+class PluginsDataViewCtrlAx final : public wxAccessible
```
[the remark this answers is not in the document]
vsverchinsky: https://github.com/audacity/audacity/issues/3865#issuecomment-1956381348

src/PluginRegistrationDialog.cpp
```diff
@@ -430,452 +96,195 @@ void PluginRegistrationDialog::Populate()
...
 
-         mEffects = S.Id(ID_List)
-            .Style(wxSUNKEN_BORDER | wxLC_REPORT | wxLC_HRULES | wxLC_VRULES )
-            .ConnectRoot(wxEVT_KEY_DOWN,
-                      &PluginRegistrationDialog::OnListChar)
-            .AddListControlReportMode({ XO("Name"), XO("State"), XO("Path") });
-#if wxUSE_ACCESSIBILITY
-         mEffects->SetAccessible(mAx = safenew CheckListAx(mEffects));
-#endif
+         if(S.GetMode() == eIsCreating)
```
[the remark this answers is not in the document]
vsverchinsky: Yes, I'll remove that `if` statement

src/PluginsDataViewCtrl.cpp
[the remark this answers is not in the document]
saintmatthieu: ACK

src/PluginRegistrationDialog.cpp
petersampsonaudacity: > I noticed (and think I remember someone else mentioning this in the ticket, maybe Peter) that once this modal shows, the visible plugin items become greyed out

@saintmatthieu 

Yes I recall that was me - but:
a) I can't find where I reported it,
b) I can't seem to be able to replicate it on the latest branch build for this.

vsverchinsky: @petersampsonaudacity As Matthieu accurately noticed, it seem that rows become grayed out when "Incompatible Plugins" dialog appear. In latest update I provided a hotfix which works for me.
Also, search when typing should now loop over the table.

src/PluginDataViewCtrl.h
```diff
@@ -2,22 +2,20 @@
 
 #include <wx/dataview.h>
 
-class PluginsDataViewCtrl final : public wxDataViewCtrl
+class PluginDataViewCtrl final : public wxDataViewCtrl
```
saintmatthieu, line 5: 🥲

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch:  audacity-win-3.5.0-alpha-20240306+c6c5a1c-x64

a) The cycling based on initial letter now seems good

b) if you are lower in the list that the first letter you type it will go back and find the EGATs starting with that letter

-------------------------------------------------------------------------------------------------

But I observe a small oddity with the alphabetic ordering of the list.

A couple of my VST/VST3 plugins are out of order at the bottom of the list.  This does not appear to be a VST/VST3 problem per se  as I have other VST/VST3 plugins further up the list which are in the correct alpahbetic order.

image

src/PluginDataViewCtrl.cpp
```diff
@@ -0,0 +1,662 @@
...
+   EVT_CHAR(PluginDataViewCtrl::OnCharEvent)
+END_EVENT_TABLE()
+
+#if defined(wxHAS_GENERIC_DATAVIEWCTRL) || defined(__WXGTK__)
+class PluginDataViewRenderer : public wxDataViewCustomRenderer
+{
+public:
+   using wxDataViewCustomRenderer::wxDataViewCustomRenderer;
+
+   bool ActivateCell(const wxRect& cell,
```
saintmatthieu, line 17: `override`

src/PluginDataModel.h
```diff
@@ -0,0 +1,80 @@
...
+      ColumnState,
+      ColumnNum
+   };
+
+   PluginDataModel(int filterEffectCategory = -1,
+                   int filterState = -1,
+                   const wxString& filterType = {},
+                   const wxString& filterExpr = {});
+
+   wxDataViewItem GetItemForRow(uint32_t row) const;
```
saintmatthieu, line 47: Either this should be an `override` or it is not used.
Please review this for other methods as well.
That will help the review, and making as many of these `private` as possible on top of that would help even more.

saintmatthieu (commented):
I can say that the code style and the much I've understood of the code looks ok, but I've only made a superficial review of this PR. For a more confident review I'd have had to invest a lot more time learning about those many classes, inheritances and flows. Thankfully, it is being thoroughly eyeballed and tested by @petersampsonaudacity and @DavidBailes, which I think can catch bugs better than my review. If it looks good to both these gentlemen I therefore suggest that it moves on to our internal QA.
Okay with that, @crsib and @ianalarina ?

DavidBailes:
@vsverchinsky . Having C as the access key for Category has created a duplicate, as Cancel also uses C. Suggest you change the access key of Category to a.
Also as noted above by @petersampsonaudacity , the alphabetic ordering should not be case sensitive.

petersampsonaudacity:
> this dialog is going to be the closest thing to perfection we have in this app

I'm certainly liking it a lot :-)

saintmatthieu (approved):

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch for this:  audacity-win-3.5.0-alpha-20240311+dcc70d2-x64

The alphabetic ordering now properly ignores case:
image

src/PluginRegistrationDialog.cpp
```diff
@@ -430,452 +92,194 @@ void PluginRegistrationDialog::Populate()
...
 void PluginRegistrationDialog::PopulateOrExchange(ShuttleGui &S)
 {
-   S.StartVerticalLay(true);
+   constexpr int Margin = 12;
+
+   static const TranslatableStrings hiddenProviders = {
+      XO("Builtin Commands")
+   };
+   static const std::unordered_map<TranslatableString, TranslatableString> nameSubstitutions = {
+      { XO("Builtin Effects"), XO("Audacity Effects") }
```
LWinterberg, line 101: One small thing: Currently, sorting by "Generator", the "Type" menu saying "Audacity Effects"/"Nyquist Effects"/... is confusing, as you really are seeing something that's not an effect. Can you rename them to the following instead?

* Audio Unit 
* LADSPA 
* LV2 
* VST 
* VST3 
* Nyquist 
* Native

(where native is what currently is "Audacity Effects")

dozzzzer:
@vsverchinsky 

These two issues can be reliably reproduced:

#### 1. On Windows, when clicking into a checkbox to enable a plugin/effect, it wrongly clicks in a different row:

https://github.com/audacity/audacity/assets/70063970/cee9135a-2053-4c66-a497-aa5e50696537

#### 2. On macOS, when double-clicking on an empty row in the Plugin Manager, Audacity crashes.

These issues occur seemingly at random and only on macOS, but still, it isn't too hard to come across them since they exhibit themselves frequently. These issues are:

#### 3. Audacity crashes after having disabled some built-in effect(s), then switching the Category to Generator and trying to rescan.

##### Example 1:

https://github.com/audacity/audacity/assets/70063970/8413ff05-79c4-4d7d-840e-cde3d54232f2

##### Example 2:

https://github.com/audacity/audacity/assets/70063970/11e90f59-580b-420d-8b5d-35f187e24c1c

#### 4. Audacity may also freeze sometimes with or without the "Audacity Support Data" dialog popping up when performing pretty much the same set of actions, i.e.:

1. Effect > Plugin Manager
2. Disable one-two effects
3. Click Rescan
4. Click Cancel to abort scanning
5. Click OK to dismiss the dialog
6. Select Category: Generator
7. Click Rescan

##### Example 1

https://github.com/audacity/audacity/assets/70063970/432aecc0-c6b5-41d9-8635-36aabab21467

##### Example 2

https://github.com/audacity/audacity/assets/70063970/b34d122f-68bb-4bb1-9f21-ac3c3a37f31c

#### 5. And lastly, much more rarely Audacity may freeze when canceling scanning. For this one, I don't have a screen recording.

petersampsonaudacity:
@dozzzzer 

> 1. On Windows, when clicking into a checkbox to enable a plugin/effect, it wrongly clicks in a different row:

I cannot reproduce this on W10

_Well, I can if I have run older versions and get multiple copies of the same EGAT - buut not if I clear the Audacity settings folder before launching the latest branch build for this._

petersampsonaudacity:
>3. Audacity crashes after having disabled some built-in effect(s), then switching the Category to Generator and trying to rescan.

@dozzzzer 

I cannot reproduce this on W10

petersampsonaudacity:
@dozzzzer @vsverchinsky 

>4. Audacity may also freeze sometimes with or without the "Audacity Support Data" dialog popping up when performing pretty much the same set of actions

>5. And lastly, much more rarely Audacity may freeze when canceling scanning. For this one, I don't have a screen recording.

I can never get to the **Cancel** button on the **Rescan** dialog anything like quickly enough - so I find myself unable to test this.

vsverchinsky:
@dozzzzer 
> 1. On Windows, when clicking into a checkbox to enable a plugin/effect, it wrongly clicks in a different row:

Nice catch! When I was debugging this issue I've found out that to reproduce this you'll need to have multiple rows with equal values in the column and enable sorting for this column. "Type" is a best option.

vsverchinsky:
@dozzzzer It seem that all 2-5 crashes/freezes may have same origin. At least I could reproduce any of 3-5 after I caught and fixed second issue.

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch build for this:  audacity-win-3.5.0-alpha-20240314+cfb3b3a-x64

The list of **Types** has had "effects" trimmed from them as per @LWinterberg 's request
And "Native Audacity" is there for Audacity supplied EGATS (thanks for that)

image

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch build for this:  audacity-win-3.5.0-alpha-20240314+cfb3b3a-x64

A few posts ago in this thread @LWinterberg requested that in the Type field we should lose the term "effect" from the type listings as not all EGATs are effects.  And this is now done as reported above.

**BUT** in the EGAT lists the non-effect EGATs show as "effects" in their path listing:

image

petersampsonaudacity:
Testing on W10 with @vsverchinsky 's latest branch build for this:  audacity-win-3.5.0-alpha-20240320+4f4b2e0-x64

@LWinterberg @dozzzzer @Tantacrul 

From a user and usability perspective all seems fine but with one small oddity/niggle - and that is the behavior of the space-bar shortcut with multiple selections (basically it acts as a toggle).

a) If you select a single EGAT in the displayed list and press the space-bar it toggles the enabled/disabled status of that selected EGAT.

b) If you make a selection of multiple EGATs in the displayed list (with Shift or Ctrl modified selection) and press the space-bar it toggles the entire selection based on the status of the **last** EGAT that you select- rather than toggling the status of each individual EGAT list based on its current status (and it is topmost not the first one you select).
Is this intended and desired behavior ?   
Is it what a user would expect ?  I would have thought that most users would expect a toggle of existing status for all selected EGATs _(my "Mystery-Shopper" thought so too when I ran this past her)_.

-----------------------------------------------------------------------

STR
1. Effect > Plugin Manager
2. set Type filter to native Audacity
3. Observe:  Classic Filters is the only disabled entry
4. Click on Classic Filters to select it
5. Ctrl+Click on Auto Duck and Bass&Treble to select those too
6. Press the space-bar: 
7. Observe all three become un-selected i.e. Classic Filters does _**not**_ toggle

src/PluginRegistrationDialog.cpp
```diff
@@ -108,22 +105,19 @@ void PluginRegistrationDialog::PopulateOrExchange(ShuttleGui &S)
...
       {
-         auto it = nameSubstitutions.find(symbol.Msgid());
-         if(it != nameSubstitutions.end())
+         auto it = extraProviders.find(provider->GetSymbol().Msgid());
+         if(it != extraProviders.end())
             pluginProviderNames.push_back(it->second);
          else
-            pluginProviderNames.push_back(symbol.Msgid());
-         mPluginProviderIDs.push_back(PluginManager::GetID(provider.get()));
+            continue;
```
saintmatthieu, line 118: With this change it is not guaranteed anymore that `pluginProviderNames` and `mPluginProviderIds` have equal size anymore, is it? Then it is used in
```cpp
void PluginRegistrationDialog::OnTypeFilterValueChanged(wxCommandEvent& evt)
{
   const auto index = evt.GetInt();
   if(index >= 0 && index < mPluginProviderIDs.size())
      mPluginsModel->SetFilterType(mPluginProviderIDs[index]);
}
```
and I haven't yet understood what the consequences could be.

saintmatthieu (commented):
Review the commits since the last I reviewed, which was "Loop the search over the table while typing", until the last commit at present, "Don't show hidden effects". ~~I just have that one thing that might not be correct, I hope you can clarify.~~ All good.

src/PluginRegistrationDialog.cpp
saintmatthieu: All good, I didn't realize that `continue` also skipped the call to `mPluginProviderIDs.push_back(PluginManager::GetID(provider.get()));`.

PR #6068, opened 2024-02-29
6001 formant preservation

### LINKED ISSUE #6001: Formant preservation
### Your idea

Use a formant preservation technique to improve the pitch shift quality, targeting vocals. Which one and where to integrate it is open.

### Problem to be solved

Currently, pitch shifting a human voice by only a few semitones makes it sound unrealistic.
This is because the resonances of the vocal cavity are also pitched up and down, virtually reducing or enlarging the vocal cavity dimensions and making for unrealistic characters (chipmunk or Darth Vador).
Some commercial pitch shift products offer formant preservation, e.g. Ableton's clip transposition, when using the "Warp" option and the "Pro" algorithm, yields currently a convincing result.
image

### Prior art

_No response_

### Additional context

_No response_

saintmatthieu:
Resolves: #6001 

Recommended:

QA:
- [x] Pitch and Speed dialog now has an "Optimize for Voice" checkbox.
- [x] Checking it noticeably (hopefully for the best when the input is voice) affects the audio result when pitch shifting up- and down.
- [x] Checking it also affects time stretching, although less noticeably, and hopefully positively when the input is voice.
- [x] Doesn't hurt the ears when used on silence, synthetic tones or other unusual signals (if you can think of any).
- [x] The checkbox state is persistent across project saving and closing.
- [x] The checkbox can be checked/unchecked during playback. (A click may be heard when doing so, but that's acceptable.)

### DIFF
--- a/libraries/lib-stretching-sequence/ClipSegment.h
+++ b/libraries/lib-stretching-sequence/ClipSegment.h
@@ -14,8 +14,8 @@
 #include "ClipTimeAndPitchSource.h"
 #include "Observer.h"
 #include "PlaybackDirection.h"
-
 #include <memory>
+#include <mutex>
 
 class ClipInterface;
 class TimeAndPitchInterface;
@@ -41,8 +41,11 @@ private:
    const sampleCount mTotalNumSamplesToProduce;
    sampleCount mTotalNumSamplesProduced = 0;
    ClipTimeAndPitchSource mSource;
-   // Careful that this guy is constructed last, as its ctor refers to *this.
+   std::mutex mStretcherMutex;
+   // Careful that this guy is constructed after `mSource`, which it refers to
+   // in its ctor.
    // todo(mhodgkinson) make this safe.
    std::unique_ptr<TimeAndPitchInterface> mStretcher;
    Observer::Subscription mOnSemitoneShiftChangeSubscription;
+   Observer::Subscription mOnFormantPreservationChangeSubscription;
 };
--- a/libraries/lib-stretching-sequence/StretchingSequence.cpp
+++ b/libraries/lib-stretching-sequence/StretchingSequence.cpp
@@ -17,9 +17,9 @@
 
 namespace
 {
-void
-GetOffsetBuffer(float **offsetBuffer,
-   float* const* buffer, size_t numChannels, size_t offset)
+void GetOffsetBuffer(
+   float** offsetBuffer, float* const* buffer, size_t numChannels,
+   size_t offset)
 {
    for (auto i = 0u; i < numChannels; ++i)
       offsetBuffer[i] = buffer[i] + offset;
@@ -44,7 +44,7 @@ void StretchingSequence::ResetCursor(double t, PlaybackDirection direction)
 }
 
 bool StretchingSequence::GetNext(
-   float *const buffers[], size_t numChannels, size_t numSamples)
+   float* const buffers[], size_t numChannels, size_t numSamples)
 {
    if (!mExpectedStart.has_value())
       ResetCursor(0., PlaybackDirection::forward);
@@ -55,9 +55,9 @@ bool StretchingSequence::GetNext(
       const auto& segment = *mActiveAudioSegmentIt;
       // More-than-stereo isn't supported
       assert(mSequence.NChannels() <= 2);
-      float *offsetBuffers[2]{};
-      GetOffsetBuffer(offsetBuffers,
-         buffers, mSequence.NChannels(), numProcessedSamples);
+      float* offsetBuffers[2] {};
+      GetOffsetBuffer(
+         offsetBuffers, buffers, mSequence.NChannels(), numProcessedSamples);
       numProcessedSamples += segment->GetFloats(
          offsetBuffers,
          numSamples - numProcessedSamples); // No need to reverse, we feed the
@@ -71,7 +71,7 @@ bool StretchingSequence::GetNext(
    {
       // More-than-stereo isn't supported
       assert(mSequence.NChannels() <= 2);
-      float *offsetBuffers[2]{};
+      float* offsetBuffers[2] {};
       GetOffsetBuffer(
          offsetBuffers, buffers, mSequence.NChannels(), numProcessedSamples);
       for (auto i = 0u; i < mSequence.NChannels(); ++i)
@@ -94,16 +94,16 @@ float StretchingSequence::GetChannelGain(int channel) const
    return mSequence.GetChannelGain(channel);
 }
 
-bool StretchingSequence::DoGet(size_t iChannel, size_t nBuffers,
-   const samplePtr buffers[], sampleFormat format, sampleCount start,
-   size_t len, bool backwards, fillFormat fill,
-   bool mayThrow, sampleCount* pNumWithinClips) const
+bool StretchingSequence::DoGet(
+   size_t iChannel, size_t nBuffers, const samplePtr buffers[],
+   sampleFormat format, sampleCount start, size_t len, bool backwards,
+   fillFormat fill, bool mayThrow, sampleCount* pNumWithinClips) const
 {
    return const_cast<StretchingSequence&>(*this).MutableGet(
       iChannel, nBuffers, buffers, format, start, len, backwards);
 }
 
-const ChannelGroup *StretchingSequence::FindChannelGroup() const
+const ChannelGroup* StretchingSequence::FindChannelGroup() const
 {
    return mSequence.FindChannelGroup();
 }
@@ -183,14 +183,15 @@ bool StretchingSequence::MutableGet(
          t,
          backwards ? PlaybackDirection::backward : PlaybackDirection::forward);
    }
-   return GetNext(reinterpret_cast<float *const *>(buffers), nBuffers, len);
+   return GetNext(reinterpret_cast<float* const*>(buffers), nBuffers, len);
 }
 
 std::shared_ptr<StretchingSequence> StretchingSequence::Create(
    const PlayableSequence& sequence, const ClipHolders& clips)
 {
+   const int sampleRate = sequence.GetRate();
    return std::make_shared<StretchingSequence>(
-      sequence, sequence.GetRate(), sequence.NChannels(),
+      sequence, sampleRate, sequence.NChannels(),
       std::make_unique<AudioSegmentFactory>(
-         sequence.GetRate(), sequence.NChannels(), clips));
+         sampleRate, sequence.NChannels(), clips));
 }
--- a/libraries/lib-time-and-pitch/CMakeLists.txt
+++ b/libraries/lib-time-and-pitch/CMakeLists.txt
@@ -37,6 +37,7 @@ set( SOURCES
 set( LIBRARIES
 PUBLIC
    lib-files-interface
+   lib-utility-interface
    pffft
 )
 audacity_library( lib-time-and-pitch "${SOURCES}" "${LIBRARIES}"
--- a/libraries/lib-time-and-pitch/CMakeLists.txt
+++ b/libraries/lib-time-and-pitch/CMakeLists.txt
@@ -4,8 +4,6 @@ Provides time-stretching and pitch-shifting functionality.
 Based on the more versatile time-stretching from StaffPad, but tailored for the needs of Audacity.
 ]]
 
-add_definitions( -DCMAKE_CURRENT_SOURCE_DIR="${CMAKE_CURRENT_SOURCE_DIR}" )
-
 set( SOURCES
    StaffPad/CircularSampleBuffer.h
    StaffPad/FourierTransform_pffft.cpp
@@ -22,6 +20,8 @@ set( SOURCES
    StaffPad/VectorOps.h
    AudioContainer.cpp
    AudioContainer.h
+   DummyFormantShifterLogger.cpp
+   DummyFormantShifterLogger.h
    FormantShifter.cpp
    FormantShifter.h
    FormantShifterLogger.cpp
@@ -29,11 +29,14 @@ set( SOURCES
    FormantShifterLoggerInterface.h
    StaffPadTimeAndPitch.cpp
    StaffPadTimeAndPitch.h
+   TimeAndPitchExperimentalSettings.cpp
+   TimeAndPitchExperimentalSettings.h
    TimeAndPitchInterface.cpp
    TimeAndPitchInterface.h
 )
 set( LIBRARIES
 PUBLIC
+   lib-files-interface
    pffft
 )
 audacity_library( lib-time-and-pitch "${SOURCES}" "${LIBRARIES}"
--- a/libraries/lib-time-and-pitch/CMakeLists.txt
+++ b/libraries/lib-time-and-pitch/CMakeLists.txt
@@ -4,6 +4,8 @@ Provides time-stretching and pitch-shifting functionality.
 Based on the more versatile time-stretching from StaffPad, but tailored for the needs of Audacity.
 ]]
 
+add_definitions( -DCMAKE_CURRENT_SOURCE_DIR="${CMAKE_CURRENT_SOURCE_DIR}" )
+
 set( SOURCES
    StaffPad/CircularSampleBuffer.h
    StaffPad/FourierTransform_pffft.cpp
@@ -20,6 +22,11 @@ set( SOURCES
    StaffPad/VectorOps.h
    AudioContainer.cpp
    AudioContainer.h
+   FormantShifter.cpp
+   FormantShifter.h
+   FormantShifterLogger.cpp
+   FormantShifterLogger.h
+   FormantShifterLoggerInterface.h
    StaffPadTimeAndPitch.cpp
    StaffPadTimeAndPitch.h
    TimeAndPitchInterface.cpp
--- a/libraries/lib-time-and-pitch/FormantShifter.cpp
+++ b/libraries/lib-time-and-pitch/FormantShifter.cpp
@@ -10,6 +10,7 @@
 **********************************************************************/
 #include "FormantShifter.h"
 #include "FormantShifterLoggerInterface.h"
+#include "MathApprox.h"
 #include <algorithm>
 #include <cassert>
 #include <fstream>
@@ -32,23 +33,6 @@ constexpr auto MapToPositiveHalfIndex(int index, int fullSize)
 
 static_assert(MapToPositiveHalfIndex(-7, 4) == 1);
 static_assert(MapToPositiveHalfIndex(9, 4) == 1);
-
-constexpr float GetLog2(float x)
-{
-   // https://stackoverflow.com/a/28730362
-   // Precise up to two decimal places. If given unnormalized power, this
-   // shouldn't be a problem.
-   union
-   {
-      float val;
-      int32_t x;
-   } u = { x };
-   auto log_2 = (float)(((u.x >> 23) & 255) - 128);
-   u.x &= ~(255 << 23);
-   u.x += 127 << 23;
-   log_2 += ((-0.3358287811f) * u.val + 2.0f) * u.val - 0.65871759316667f;
-   return log_2;
-}
 
 // `x` has length `fftSize/2+1`.
 // Returns the last bin that wasn't zeroed.
@@ -111,9 +95,9 @@ void FormantShifter::Process(
    // Take the log of the normalized magnitude. (This assumes that
    // the window averages to 1.)
    std::complex<float>* pEnv = mEnvelope.getPtr(0);
-   const float normalizer = GetLog2(fftSize);
+   const float normalizer = FastLog2(fftSize);
    std::transform(powSpec, powSpec + numBins, pEnv, [&](float power) {
-      return .5f * GetLog2(power) - normalizer;
+      return .5f * FastLog2(power) - normalizer;
    });
 
    // Get the cosine transform of the log magnitude, aka the cepstrum.
--- a/libraries/lib-time-and-pitch/FormantShifter.cpp
+++ b/libraries/lib-time-and-pitch/FormantShifter.cpp
@@ -1,0 +1,188 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  FormantShifter.cpp
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#include "FormantShifter.h"
+#include "FormantShifterLoggerInterface.h"
+#include <algorithm>
+#include <cassert>
+#include <fstream>
+
+namespace
+{
+constexpr auto MapToPositiveHalfIndex(int index, int fullSize)
+{
+   const auto inputIsValid = fullSize > 0 && fullSize % 2 == 0;
+   if (!inputIsValid)
+      return 0;
+   if (index >= 0)
+      index = index % fullSize;
+   else
+      index = fullSize - (-index % fullSize);
+   if (index > fullSize / 2)
+      index = fullSize - index;
+   return index;
+}
+
+static_assert(MapToPositiveHalfIndex(-7, 4) == 1);
+static_assert(MapToPositiveHalfIndex(9, 4) == 1);
+
+constexpr float GetLog2(float x)
+{
+   // https://stackoverflow.com/a/28730362
+   // Precise up to two decimal places. If given unnormalized power, this
+   // shouldn't be a problem.
+   union
+   {
+      float val;
+      int32_t x;
+   } u = { x };
+   auto log_2 = (float)(((u.x >> 23) & 255) - 128);
+   u.x &= ~(255 << 23);
+   u.x += 127 << 23;
+   log_2 += ((-0.3358287811f) * u.val + 2.0f) * u.val - 0.65871759316667f;
+   return log_2;
+}
+
+// `x` has length `fftSize/2+1`.
+// Returns the last bin that wasn't zeroed.
+size_t resampleFreqDomain(float* x, size_t fftSize, double factor)
+{
+   const auto size = fftSize / 2 + 1;
+   const auto end = std::min(size, size_t(size * factor));
+   std::vector<float> tmp(end);
+   for (size_t i = 0; i < end; ++i)
+   {
+      const int int_pos = i / factor;
+      const float frac_pos = 1.f * i / factor - int_pos;
+      const auto k = MapToPositiveHalfIndex(int_pos, fftSize);
+      const auto l = MapToPositiveHalfIndex(int_pos + 1, fftSize);
+      tmp[i] = (1 - frac_pos) * x[k] + frac_pos * x[l];
+   }
+   std::copy(tmp.begin(), tmp.begin() + end, x);
+   if (end < size)
+      std::fill(x + end, x + size, 0.f);
+   return end;
+}
+} // namespace
+
+FormantShifter::FormantShifter(
+   int sampleRate, double cutoffQuefrency,
+   FormantShifterLoggerInterface& logger)
+    : cutoffQuefrency { cutoffQuefrency }
+    , mSampleRate { sampleRate }
+    , mLogger { logger }
+{
+}
+
+void FormantShifter::Reset(size_t fftSize)
+{
+   mFft = std::make_unique<staffpad::audio::FourierTransform>(fftSize);
+   const auto numBins = fftSize / 2 + 1;
+   mEnvelope.setSize(1, numBins);
+   mCepstrum.setSize(1, fftSize);
+   mEnvelopeReal.resize(numBins);
+   mWeights.resize(numBins);
+}
+
+void FormantShifter::Reset()
+{
+   mFft.reset();
+}
+
+void FormantShifter::Process(
+   const float* powSpec, std::complex<float>* spec, double factor)
+{
+   assert(factor > 0);
+   if (factor <= 0 || cutoffQuefrency == 0 || !mFft)
+      return;
+
+   const auto fftSize = mFft->getSize();
+   const auto numBins = fftSize / 2 + 1;
+
+   mLogger.Log(fftSize, "fftSize");
+
+   // Take the log of the normalized magnitude. (This assumes that
+   // the window averages to 1.)
+   std::complex<float>* pEnv = mEnvelope.getPtr(0);
+   const float normalizer = GetLog2(fftSize);
+   std::transform(powSpec, powSpec + numBins, pEnv, [&](float power) {
+      return .5f * GetLog2(power) - normalizer;
+   });
+
+   // Get the cosine transform of the log magnitude, aka the cepstrum.
+   mFft->inverseReal(mEnvelope, mCepstrum);
+   auto pCepst = mCepstrum.getPtr(0);
+   mLogger.Log(pCepst, fftSize, "cepstrum");
+
+   // "Lifter" the cepstrum.
+   const auto binCutoff = int(mSampleRate * cutoffQuefrency * factor);
+   if (binCutoff < fftSize / 2)
+      std::fill(pCepst + binCutoff + 1, pCepst + fftSize - binCutoff, 0.f);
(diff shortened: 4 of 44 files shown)


Paul-Licameli:
You need a Windows build fix

saintmatthieu:
> You need a Windows build fix

I have a segmentation fault on the StaffPadTimeAndPitch unit-test, which I can't reproduce locally :/

saintmatthieu:
Found what the problem was - memory alignment. Pushed a fix commit with an explanation. Thanks unit tests 🙏

libraries/lib-time-and-pitch/FormantShifter.cpp
```diff
@@ -0,0 +1,183 @@
...
+   const auto numBins = fftSize / 2 + 1;
+
+   mLogger.Log(fftSize, "fftSize");
+
+   // Take the log of the normalized magnitude. (This assumes that
+   // the window averages to 1.)
+   std::complex<float>* pEnv = mEnvelope.getPtr(0);
+   const float normalizer = GetLog2(fftSize);
+   std::transform(powSpec, powSpec + numBins, pEnv, [&](float power) {
+      // Why -12 ?
```
patrickvi2, line 114: Why?

libraries/lib-time-and-pitch/FormantShifter.cpp
```diff
@@ -0,0 +1,183 @@
...
+   mWeights.resize(numBins);
+}
+
+void FormantShifter::Reset()
+{
+   mFft.reset();
+}
+
+void FormantShifter::Process(
+   const float* powSpec, std::complex<float>* spec, double factor)
```
patrickvi2, line 99: As you use `factor` for calculating the bounds of your iterators, it's maybe a good idea to check `factor` is in expected range.

libraries/lib-time-and-pitch/FormantShifter.cpp
```diff
@@ -0,0 +1,183 @@
...
+      [fftSize = fftSize](const std::complex<float>& env) {
+         return std::exp2(env.real() / fftSize);
+      });
+   mLogger.Log(mEnvelopeReal.data(), numBins, "envelope");
+
+   // Get the weights, which are the ratio of the desired envelope to the
+   // current envelope (which has the effect of downsampling).
+   std::transform(
+      mEnvelopeReal.begin(), mEnvelopeReal.end(), mWeights.begin(),
+      [](float env) { return std::isnormal(env) ? 1 / env : 0.f; });
```
patrickvi2, line 144: Shouldn't it be `1.f / env`?

patrickvi2, line 144: As `env` can still be close to infinity small, the resulting weight might get a huge number introducing noise and artifacts. I could imagine, that either limiting `1/env` or the weight would result in less audible artifacts.

patrickvi2 (approved):

libraries/lib-time-and-pitch/FormantShifter.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <algorithm>
+#include <fstream>
+
+namespace
+{
+constexpr auto MapToPositiveHalfIndex(int index, int fullSize)
+{
+   const auto inputIsValid = fullSize > 0 && fullSize % 2 == 0;
+   if (!inputIsValid)
+      return 0;
```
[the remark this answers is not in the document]
saintmatthieu: I didn't know `expected`, that's interesting, thank you. Unfortunately we still C++17, and even our CI toolchain for MacOS doesn't support the full C++17 std lib, e.g. `<filesystem>`. I don't know the exact version numbers, but the oldest MacOS version we release Audacity for cannot deal with the most recent C++ versions.
Also, we are relatively lenient in terms of safety: we document functions with `@pre` conditions, assert these preconditions, but generally don't otherwise deal with error handling. You won't find this pattern often, where the return value is checked for error. Seems to work for us 🤷 On the other hand, unit-testing is encouraged, and I have been lazy doing this time...

libraries/lib-time-and-pitch/FormantShifter.cpp
```diff
@@ -0,0 +1,183 @@
...
+   u.x &= ~(255 << 23);
+   u.x += 127 << 23;
+   log_2 += ((-0.3358287811f) * u.val + 2.0f) * u.val - 0.65871759316667f;
+   return log_2;
+}
+
+// `x` has length `fftSize/2+1`.
+// Returns the last bin that wasn't zeroed.
+size_t ResampleFreqDomain(float* x, size_t fftSize, double factor)
+{
```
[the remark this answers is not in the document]
saintmatthieu: I will assert at the beginning of `FormantShifter::Process` like I should have, given the pre-condition stated in that methods documentation. Concerning this particular method, given its narrow scope, it usually is tolerated without.

libraries/lib-time-and-pitch/FormantShifter.cpp
saintmatthieu: A remnant of me copying chunks from the [stftPitchShift repo](https://github.com/jurihock/stftPitchShift/blob/48042d72a43287676543950691d227732a46759b/cpp/StftPitchShift/Cepster.h#L48) to get started.
It's not a great protection, because, as you commented someplace else, the weights can still end up being huge. I expect this can happen if for some reason the input signal has a frequency region of near-zero values.
I'm about to push a protection that I think makes sense: the purpose of formant preservation is to boost or attenuate spectral regions by some reasonable amount of decibels. This commit will limit the amplification to +20dB, which is plenty for our purpose and then avoid crazy amplification.

libraries/lib-time-and-pitch/FormantShifter.cpp
saintmatthieu: Since `env` has type `float`, I expect `1` to be typecast to float, too. But sure, it doesn't hurt adding the `.f`.

saintmatthieu: Agreed, limiting the weights it will be.

saintmatthieu:
@patrickvi2 thanks a lot for your comments ❤️ I just pushed commits to address them, feel free to re-review.

libraries/lib-utility/MathApprox.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include <cstdint>
+
+/*!
+ * @brief Approximates the base-2 logarithm of a float to two decimal places,
+ * adapted from https://stackoverflow.com/a/28730362
+ *
+ * @details Accuracy claim is demonstated in MathApproxTest.cpp, and coincides
+ * with @Paul-Licameli's analysis (copied from conversation):
+ * "If x == 1.0f, then log2 before the += correction actually becomes -1 --
```
saintmatthieu, line 21: @Paul-Licameli would you be so kind to review that me pasting your Discord message here is relevant?

saintmatthieu:
@vsverchinsky: @patrickvi2 has reviewed the DSP part of this PR. Would you be so kind to review the rest?

libraries/lib-utility/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set( SOURCES
    GlobalVariable.h
    IteratorX.cpp
    IteratorX.h
+   MathApprox.h
```
crsib, line 35: I am confused. Why `lib-utility` instead of `lib-math`?

saintmatthieu: `lib-math` depends on `lib-soxr`. I'm reluctant to let the time-and-pitch depend on this mainly because it compiles PFFFT functions which are also found in `lib-fft`, another dependency of time-and-pitch.

libraries/lib-time-and-pitch/FormantShifterLogVisualization.py
```diff
@@ -0,0 +1,62 @@
...
+#
+# 1. Create overrideLogTime.txt in lib-time-and-pitch's root directory,
+# 2. Enter a time in seconds, e.g. (0.5)
+# 3. Run your Audacity build and play a track with pitch shifting and formant preservation enabled.
+#
+# You will hear a gentle noise burst at the specified time, which indicates precisely the place where the snapshot was taken.
+# Now you can run this script from lib-time-and-pitch's root directory.
+
+import matplotlib.pyplot as plt
+import FormantShifterLog as FSL
```
crsib, line 11: I think this should be moved into a separate directory, and with `requirements.txt`

libraries/lib-time-and-pitch/FormantShifterLoggerInterface.h
```diff
@@ -0,0 +1,40 @@
...
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#pragma once
+
+#include <complex>
+#include <functional>
+
+class FormantShifterLoggerInterface
```
crsib, line 18: Do you plan to have different implementation?

libraries/lib-time-and-pitch/FormantShifterLogger.h
```diff
@@ -0,0 +1,51 @@
...
+      const override;
+   /*!
+    * @brief If not already, disables the logging and marks the spectrum with an
+    * audible event to make clear where in the signal the logging took place.
+    * (Of course not for use in production :D)
+    */
+   void ProcessFinished(std::complex<float>* spectrum, size_t fftSize) override;
+
+public:
+   static std::optional<double> GetCutoffQuefrencyOverride();
```
crsib, line 39: Why is this a part of a logger (which is a logger otherwise)?

libraries/lib-time-and-pitch/FormantShifterLogger.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+**********************************************************************/
+#include "FormantShifterLogger.h"
+#include <algorithm>
+#include <cmath>
+#include <vector>
+
+namespace
+{
+const std::string logDir { CMAKE_CURRENT_SOURCE_DIR };
```
crsib, line 18: I believe you only use it for tests. It's way too intrusive, TBH

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.cpp
```diff
@@ -19,38 +20,72 @@ void GetOffsetBuffer(
       offsetBuffer[i] = buffer[i] + offset;
 }
 
+int GetFftSize(int sampleRate, bool formantPreservationOn)
+{
+   if (const auto fftSize = FormantShifterLogger::GetFftSizeOverride())
```
crsib, line 25: You use `FormantShifterLogger` directly here. This happens because `GetFftSizeOverride` must not be a part of it.

I think that you should have an empty logger for production and use FormantShifterLogger only for tests. 

Configurability must not be a part of logger.

libraries/lib-time-and-pitch/StaffPadTimeAndPitch.h
```diff
@@ -14,16 +15,20 @@ class TIME_AND_PITCH_API StaffPadTimeAndPitch final :
...
+   void OnFormantPreservationChange(bool preserve) override;
 
 private:
    void BootStretcher();
    bool IllState() const;
+   void ResetStretcher();
+
+   const int mSampleRate;
+   TimeAndPitchInterface::Parameters mParameters;
+   FormantShifterLogger mFormantShifterLogger;
```
crsib, line 27: You do not utilize the interface again. Was there any need for it then?

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -376,9 +376,10 @@ void TimeAndPitch::_process_hop(int hop_a, int hop_s)
        // is what one typically gets at those frequencies.
        const int imagingBeginBin = fftSize / 2 * _pitchFactor + 1;
        const auto n = _numBins - imagingBeginBin;
-       vo::rotate(
```
crsib: We spent a lot of time and effort to optimize it, it is weird to see it go.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -44,12 +44,15 @@ inline float lagrange6(const float (&smp)[6], float t)
   return (c0 + c1 * t) + (c2 + c3 * t) * t2 + (c4 + c5 * t) * t2 * t2;
 }
 
-std::vector<float> generateRandomPhaseVector(size_t size)
+std::vector<std::complex<float>> generateRandomPhaseVector(size_t size)
```
crsib, line 47: Paul has introduced the class for aligned vectors, hasn't he?

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
crsib: You go from SIMD to scalar algorithm, and I don't see any good reason for that

libraries/lib-utility/CMakeLists.txt
crsib: And why is it a problem? It doesn't introduce a circular dependency.

libraries/lib-utility/MathApprox.h
```diff
@@ -0,0 +1,45 @@
...
+ * bits, but one more was subtracted. > u.x by that point has been normalized to
+ * the range [1.0f, 2.0f), by changing the exponent, multiplying it by an
+ * integer power of 2. > So the correction must add 1 -- so that log2(1) comes
+ * out (approximately) zero -- but then, add a quadratic polynomial in u.x,
+ * a*x*x + b*x + c, where a is -0.3358287811f, b is 2, c is -1.65871759316667f
+ * (the magic number as written, minus 1) > Evaluate that at x = 1, and you
+ * don't get exactly 0 !  You get a little off.  It's correct to only two
+ * decimal places.  Curious why they chose it that way."
+ */
+constexpr float GetLog2(float x)
```
crsib, line 32: And probably, it should be called FastLog2 or whatever.

libraries/lib-utility/MapToPositiveHalfIndex.h
```diff
@@ -0,0 +1,31 @@
...
+   if (index >= 0)
+      index = index % fullSize;
+   else
+      index = fullSize - (-index % fullSize);
+   if (index > fullSize / 2)
+      index = fullSize - index;
+   return index;
+}
+
+static_assert(MapToPositiveHalfIndex(-3, 4) == 1);
```
crsib, line 24: I don't like it, but I don't have a strong opinion, either

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -43,16 +44,17 @@ inline float lagrange6(const float (&smp)[6], float t)
...
 {
-  // 44.1kHz maps to 4096 samples (i.e., 93ms).
-  // We grow the FFT size proportionally with the sample rate to keep the
-  // window duration roughly constant, with quantization due to the
-  // power-of-two constraint.
-  // If needed some time in the future, we can decouple analysis window and
-  // FFT sizes by zero-padding, allowing for very fine-grained window duration
-  // without compromising performance.
-  return 1 << (12 + (int)std::round(std::log2(sampleRate / 44100.)));
+  // keep the processing deterministic
```
crsib, line 49: I don't understand this point. Why do we need it deterministic? How much do we care about this determinism? 

Provided that `srand` modifies the global state I wonder if something like https://en.cppreference.com/w/cpp/numeric/random/linear_congruential_engine should be used instead.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
```diff
@@ -354,6 +360,34 @@ void TimeAndPitch::_process_hop(int hop_a, int hop_s)
...
+       // spectrum, which, if not lowpass-filtered by the resampler, may yield
+       // an aliasing-like quality if what is mirrored is a harmonic series. A
+       // simple thing to do against that is to scramble the phase values there.
+       // This way we preserve natural energy fluctuations, but the artifact
+       // isn't noticeable as much anymore, because much more noise-like, which
+       // is what one typically gets at those frequencies.
+       const int imagingBeginBin = fftSize / 2 * _pitchFactor + 1;
+       const auto n = _numBins - imagingBeginBin;
+       auto pSpec = d->spectrum.getPtr(0);
+       std::transform(
```
crsib, line 380: I understand that this is a new code, but I will stress that it is located in a known hot point

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.h
```diff
@@ -3,17 +3,26 @@
 //
 #pragma once
 
+#include <complex>
+#include <functional>
 #include <memory>
 #include <vector>
+#include <random>
```
crsib, line 10: Why this include is in the header?

crsib (changes requested):
I have two issues with this PR

* FormantShifterLogger is responsible for handling configuration.
  * You define it using and interface too, and I don't see why. The only place it is used - it is used directly.
* I think we should keep process_hop SIMD ready. I expect that the fix will be very simple.

There are some minor comments as well, but I will not block the PR because of them.

libraries/lib-utility/CMakeLists.txt
saintmatthieu: Not a circular dependency, but how can you be confident that it's the `lib-fft` symbols that will be used?

libraries/lib-time-and-pitch/FormantShifterLogVisualization.py
saintmatthieu: What is `requirements.txt` ?

libraries/lib-time-and-pitch/FormantShifterLoggerInterface.h
saintmatthieu: No. It ended up being a way for me to expose only relevant methods to `FormantShifter` and the rest (`NewSamplesComing` and override getters) to whoever integrates logger and shifter together (`StaffPadTimeAndPitch`).

libraries/lib-time-and-pitch/FormantShifterLogVisualization.py
crsib: Your script has non-trivial dependencies. `requirements.txt` is a standard Python to define such dependencies.

libraries/lib-time-and-pitch/FormantShifterLogger.h
saintmatthieu: Because it's often related by usage: these allow overriding otherwise unexposed parameters, the effect of which one might not only need to hear but also to visualize, which then really implies the logger.

libraries/lib-time-and-pitch/FormantShifterLogger.cpp
saintmatthieu: It's convenient. What problem do you see with it?

libraries/lib-time-and-pitch/FormantShifterLogger.h
crsib: I rest my case; such configuration must not be a part of the logger.

libraries/lib-time-and-pitch/FormantShifterLogger.cpp
crsib: Provided that you need a custom config in Audacity itself, why not use the "Setting"?

crsib: > It's convenient. What problem do you see with it?

It is a hack that hardcodes some random path into the build for no reason.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
saintmatthieu: This removal is only the removal of an earlier commit in this very PR ; the older code is left untouched.
Bad things happen if I use `vo::rotate` because it expects aligned float pointers, and `pSpec + imagingBeginBin` isn't necessarily aligned.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
saintmatthieu: Not necessary here, due to [this](https://github.com/audacity/audacity/pull/6068#discussion_r1521902915).

libraries/lib-utility/MapToPositiveHalfIndex.h
saintmatthieu: static-asserting in the header file?

crsib: Writing unit tests in a header file.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
saintmatthieu: Sounds good, I'll try.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
saintmatthieu: Yes, but I cannot think of a more optimized alternative, and there are extenuating circumstances:
* only gets executed for clips that are pitch-shifted downwards,
* for a number of bins that is at most a quarter of the FFT size, when down-shifted by an octave, and less for lesser shift amounts.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.h
saintmatthieu: That's a remnant of me using something more sophisticated. Can be removed I think.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
crsib: Consider me convinced:-)

saintmatthieu:
>   * You define it using and interface too, and I don't see why. The only place it is used - it is used directly.

Not strictly speaking: `FormantShifter` only sees the abstract API.
My experimentations lead me to this solution, mixing up both logging and configuration responsibilities. I admit that it's not perfect, but I thought it could be tolerated and further clean-up work be spared. I don't really mind addressing this, if you see it as a must.

libraries/lib-time-and-pitch/FormantShifterLogger.cpp
saintmatthieu: > Provided that you need a custom config in Audacity itself, why not use the "Setting"?

You mean as defined in `Prefs.h`? I'd still need a write mechanism.

crsib: > I'd still need a write mechanism.

Can you explain please?

saintmatthieu: Sure: if e.g. a method like `std::optional<int> GetFftSizeOverride()` finds `overrideFftSizeExponent.txt` in `logDir` that has a valid int value, it will use this rather than the production value. I can hear the result change on the fly by editing that file during playback. Even if I used the prefs, I'd need this functionality, that's not exposed to the user. (And in this regard it's actually not bad that `logDir` has some random string in the production build, as it makes it near impossible for the user to inadvertently have that file at that path and get a non-production parameter.)

crsib: Well, you've just made it worse. During the playback, do you randomly try to open a file? I wouldn't call that a good idea at all. Open/read/close are syscalls. They are not cheap, and they can cause havoc in some scenarios. Do we want such an approach to support debug scenarios tied strictly to one machine? I say we don't.

> And in this regard it's actually not bad that logDir has some random string in the production build

Linux package maintainers will disagree. You make the build result depend on the file system structure of the computer used to build it. 

---

I'm very much confused by this approach. Probably, it doesn't cause any noticeable difference for you, but I am very much against it.

Please create a separate config for the pitch-shifter. If you want to stick to reading files for now, please make access to the files conditionally compiled. I also prefer you use an executable/data path as a root for those files.

In the future, if we want to expose these settings to users, it would be a very quick and easy operation.

libraries/lib-time-and-pitch/StaffPad/TimeAndPitch.cpp
saintmatthieu: > I don't understand this point. Why do we need it deterministic? How much do we care about this determinism?

For A/B comparison before and after some algorithmic tweaks, for example. I might want to make sure that my modifications do affect the signal, so I render it once before and once after. If the process weren't deterministic I couldn't tell.

libraries/lib-time-and-pitch/FormantShifterLogger.cpp
saintmatthieu: > You make the build result depend on the file system structure of the computer used to build it.

I don't understand how does that matters: it's intended for development and functional only for local builds. But I don't mind using an app-data directory. I was lazy but it has advantages.

> Open/read/close are syscalls. They are not cheap, and they can cause havoc in some scenarios.

I didn't have this in mind. I want to reduce conditional compilation to a maximum to avoid code becoming outdated, but if it's only just around the file opening that's completely fine.

crsib: >  it's intended for development and functional only for local builds

You push it to production builds, and that matters.

> I want to reduce conditional compilation to a maximum

You push the code only needed for the feature tuning into the production build. Expensive code.

libraries/lib-stretching-sequence/StretchingSequence.cpp
```diff
@@ -183,14 +183,15 @@ bool StretchingSequence::MutableGet(
...
 
 std::shared_ptr<StretchingSequence> StretchingSequence::Create(
    const PlayableSequence& sequence, const ClipHolders& clips)
 {
+   const int sampleRate = sequence.GetRate();
    return std::make_shared<StretchingSequence>(
-      sequence, sequence.GetRate(), sequence.NChannels(),
+      sequence, sampleRate, sequence.NChannels(),
       std::make_unique<AudioSegmentFactory>(
-         sequence.GetRate(), sequence.NChannels(), clips));
```
Paul-Licameli, line 195: What compilation warning was fixed by the first commit?

saintmatthieu: loss of precision warning (size_t vs int or the other way round)

libraries/lib-time-and-pitch/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ Provides time-stretching and pitch-shifting functionality.
 Based on the more versatile time-stretching from StaffPad, but tailored for the needs of Audacity.
 ]]
 
+add_definitions( -DCMAKE_CURRENT_SOURCE_DIR="${CMAKE_CURRENT_SOURCE_DIR}" )
```
Paul-Licameli: This page recommends `add_compile_definitions` instead https://cmake.org/cmake/help/latest/command/add_definitions.html

You use this macro in `FormantShifterLogger.cpp`.  I would choose a different name for the macro, like THIS_DIRECTORY, just so we aren't confusing compile-time preprocessor symbols with Cmake configure-time variables.

This commit fails to compile.

```
/Users/paullicameli/Github/audacity/libraries/lib-time-and-pitch/FormantShifter.cpp:104:31: error: no member named 'getSize' in 'staffpad::audio::FourierTransform'
   const auto fftSize = mFft->getSize();
                        ~~~~~~^
```

libraries/lib-time-and-pitch/FormantShifterLogger.h
```diff
@@ -0,0 +1,51 @@
...
+#include <optional>
+
+class FormantShifterLogger : public FormantShifterLoggerInterface
+{
+   // FormantShifterLoggerInterface, intended to be called from FormantShifter
+public:
+   void Log(int value, const char* name) const override;
+   void Log(const float* samples, size_t size, const char* name) const override;
+   void Log(
+      const std::complex<float>* samples, size_t size, const char* name,
```
Paul-Licameli, line 33: Perhaps a `ComplexFloat` alias will be convenient

libraries/lib-time-and-pitch/FormantShifterLogger.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+void FormantShifterLogger::Log(
+   const float* samples, size_t size, const char* name) const
+{
+   if (!mOfs)
+      // Keep it lightweight if we're not logging.
+      return;
+   PrintPythonVector(
+      *mOfs, std::vector<float> { samples, samples + size }, name);
```
Paul-Licameli, line 74: This unnecessary temporary vector construction makes me wish we had `std::span` instead.

Maybe instead let `PrintPythonVector` take a range of iterators as two arguments.

libraries/lib-time-and-pitch/TimeAndPitchExperimentalSettings.cpp
```diff
@@ -9,22 +9,36 @@
...
+
+// #define TIME_AND_PITCH_TUNING
 
 namespace
 {
-const std::string logDir { CMAKE_CURRENT_SOURCE_DIR };
+
+std::string GetLogDir()
+{
+   const char* appDataDir = getenv("APPDATA");
```
crsib: This code only works on Windows

saintmatthieu: That's all right, I only need it there, and if someone needs it elsewhere, she can generalize it.

libraries/lib-time-and-pitch/FormantShifterLoggerInterface.h
saintmatthieu: Update: after some clean-up there now is an alternative, dummy implementation that's injected for production.

crsib (approved):
I don't see any pressing issues left

crsib:
Please note, that tests has failed, @saintmatthieu

dozzzzer:
@saintmatthieu 

The discovered bugs are macOS only:
1. While playing, an attempt to change the clip's pitch by entering any non-zero semitones/cents value causes Audacity to freeze;
2. While playing, if semitones/cents value is other than zero and formant preservation is ON, an attempt to disable formant preservation causes a freeze as well;
3. While playing, if the Pitch and Speed dialog is opened, an attempt to close the dialog causes a freeze.

saintmatthieu:
@dozzzzer thank you, I didn't get thread safety right and for some reason it still worked on Windows.
@crsib the last commit addresses this.

libraries/lib-stretching-sequence/ClipSegment.h
```diff
@@ -41,7 +41,9 @@ class STRETCHING_SEQUENCE_API ClipSegment final : public AudioSegment
    const sampleCount mTotalNumSamplesToProduce;
    sampleCount mTotalNumSamplesProduced = 0;
    ClipTimeAndPitchSource mSource;
-   // Careful that this guy is constructed last, as its ctor refers to *this.
+   std::mutex mStretcherMutex;
```
crsib, line 44: It seems that you use the mutex for making parts of the TimeAndPitchInterface to be thread safe. I don't understand, why this mutex is not in the TimeAndPitchInterface implementation in this case.

saintmatthieu: In `ClipSegment.cpp`, you can see that we add two callbacks for pitch shift change and formant preservation toggling. These callbacks are issued from the main thread, meanwhile the processing is on the audio thread.
Since we have the guarantee that the `StaffPadTimeAndPitch` ctor is completely executed before these callbacks are registered, it would be safe to internalize the mutex, but the safety is more straightforward to implement and for reviewers to verify this way.
The drawback with this approach is that we lock the mutex at frame rate also if there is no time stretching. I still haven't got to the bottom of whether this is a bad thing, and I was thinking maybe you could enlighten me. In our case there'd only be contention when the user changes pitch during playback.

crsib (approved):
I don't see blocking issues with this PR

Issue #6070, opened 2024-02-29
"Wide" `Track` functions, transitionally modified to return or accept `TrackLists`, again simply use `Track`


Paul-Licameli:

chinakov:
@Paul-Licameli - this is ready for merge. A huge commit with no issues noticed, and thanks for the expansive QA list. Thank you!

Paul-Licameli:
> @Paul-Licameli - this is ready for merge. A huge commit with no issues noticed, and thanks for the expansive QA list. Thank you!

🎉

PR #6071, opened 2024-02-29
Simplifications

Paul-Licameli:
Resolves: #6070
Resolves: #6072

Depends on
- #5965

Recommended:

QA
- [x] Wouldn't hurt to re-test every box of #5965 
- [x] Mix and render menu commands
- [x] Lengthening sync-lock adjustment of wave tracks
- [x] Loudness effect works correctly with mono tracks alone, and with stereo
- [x] Import Raw, with more than two channels, works as before
- [x] Change Speed effect
- [x] Effect preview play
- [x] Equalization
- [x] Generate silence
- [x] Noise Reduction
- [x] Paulstretch
- [x] Repeat effect
- [x] Change Pitch (high and low quality), Change Tempo (ditto), Sliding time/pitch shift effects
- [x] Compressor
- [x] At least one Nyquist effect and Nyquist generator
- [x] Cross-document copy/paste with hidden smart clip data
- [x] Edit > Duplicate
- [x] Split Cut
- [x] Split New
- [x] Edit > Labeled Audio menu commands
- [x] FIndClipping done once; then, done again on another track, re-using the same label track named "Clipping"
- [x] Import MIDI
- [x] Import Ffmpeg
- [x] Import FLAC
- [x] Import MP3
- [x] Import OGG
- [x] Import Opus
- [x] Import PCM
- [x] Import WavPack
- [x] Shift + Click in track control panel to extend the range of selected tracks (either upward or down)
- [x] Track focus changes (up and down arrow keys moving the yellow border)
- [x] Mix and render examples of stereo with nontrivial envelope
- [x] Cross-project copy and paste, where project tempi differ, do the correct stretch on insertion
- [x] Ctrl- and unmodified-clicks on track control panels change track selection correctly
- [x] Clicks of icon buttons in mixer board (with or without shift, or ctrl) too
- [x] Click of right channel of partly visible track at top of viewport, correctly scrolls it into view
- [x] Updates of vertical ruler width for changes of track height or waveform/spectrogram scale type behave as before
- [x] Mute and solo buttons work as before
- [x] Edit > Audio Clips > Split
- [x] Click in right channel to delete or expand CutLine


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1133,6 +1133,13 @@ std::shared_ptr<WaveTrack> WaveTrackFactory::Create(sampleFormat format, double
...
 TrackListHolder WaveTrackFactory::Create(size_t nChannels)
+{
+   assert(nChannels > 0);
+   assert(nChannels <= 2);
+   return Create(nChannels, QualitySettings::SampleFormatChoice(), mRate.GetRate());
+}
+
+TrackListHolder WaveTrackFactory::CreateMany(size_t nChannels)
 {
    return Create(nChannels, QualitySettings::SampleFormatChoice(), mRate.GetRate());
```
saintmatthieu: Shouldn't this be calling `CreateMany` ?

src/menus/TrackMenus.cpp
```diff
@@ -74,7 +74,7 @@ void DoMixAndRender(AudacityProject &project, bool toNewTrack)
       }
 
       // Add new tracks
-      const bool stereo = newTracks->NChannels() > 1;
+      const bool stereo = (*newTracks->begin())->NChannels() > 1;
```
saintmatthieu: So now `newTracks.size() == 1` ?

libraries/lib-note-track/NoteTrack.cpp
```diff
@@ -173,7 +173,7 @@ Alg_seq &NoteTrack::GetSeq() const
    return *mSeq;
 }
 
-TrackListHolder NoteTrack::Clone(bool) const
+Track::Holder NoteTrack::Clone(bool) const
```
saintmatthieu, line 176: Nice!

saintmatthieu (commented):
Reviewed commits up to 43228cedb1ecbe62db4e99a4a973e059e342112b exclusive.
Best PR *ever* !!

Paul-Licameli:
> Reviewed commits up to [43228ce](https://github.com/audacity/audacity/commit/43228cedb1ecbe62db4e99a4a973e059e342112b) exclusive. Best PR _ever_ !!

I have other candidates for that honor over all the years..

Do you know the term "scaffolding?"

What I did in 3.4 was put some scaffolding in while figuring out a big transformation.  At last the building is sound enough to take the scaffolding down again.

libraries/lib-wave-track/WaveTrack.cpp
Paul-Licameli: Wait to read a later commit that makes the distinction in return types.

src/menus/TrackMenus.cpp
Paul-Licameli: Yes

In fact, I now think this commit needs to be in the previous PR for correctness of all stages of the transformation.

So I squashed it into a68333aad87a7dd417aa8ad73e5c8a8c3f8b5bbc

Paul-Licameli:
Commit "Track::Duplicate and Clone return just a track, not a list" introduced a crashing bug, for instance when generating noise.  The fix was that `DuplicateWithOtherTempo` must remove the second, output argument, which was no longer getting initialized.  You may see this change in my next force-push of this branch.

saintmatthieu (approved):

Paul-Licameli:
Rebased unchanged onto updated dependency

Paul-Licameli:
Rebased unchanged onto updated wave-clip-refactoring branch

Issue #5792, opened 2023-12-17
Audacity crashes when pasting over a clip in 192k/24-bit


mapisilafi:
### Bug description

My workflow involves copy-pasting from a multi-hour recording to individual tracks. I usually select-copy from the original and paste to another Audacity track on another Audacity window. If I directly paste on top of selected audio, the Audacity crashes and exits completely. If I delete the content of the track first, and paste on it only after deletion, it works fine. Happened several times. Sending crash report fails. 

<details>Here's the beginning of the crash dump:
Operating system: Windows NT
                  10.0.22621 2861
CPU: amd64
     family 6 model 158 stepping 13
     12 CPUs

GPU: UNKNOWN

Crash reason:  EXCEPTION_ACCESS_VIOLATION_READ
Crash address: 0x0
Process uptime: 676 seconds</details>

### Steps to reproduce

1. Open a large, multi-hour recording (192k/24b)
2. Select a few minutes of that recording
3. Copy using ctrl+c
4. ctrl+n for a new file
5. ctrl+v -> "only selected audio"
6. switch to the original project, copy another few minute snippet of the audio
7. switch back to the other Audacity window with a short audio clip (which is currently selected)
8. ctrl-v to paste over selected audio

### Expected behavior

Paste over should replace selected audio on the track.

### Actual behavior

Crash

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Not reproducible on macOS 14.2 as well as Fedora 39

dozzzzer:
I couldn't reproduce this on my Windows 11 machine, neither with a mono nor with a stereo track. @mapisilafi could you possibly share the screen recording depicting this issue?

mapisilafi:
Sure, here u are:
[audacity crash when pasting over existing clip.zip](https://github.com/audacity/audacity/files/13707153/audacity.crash.when.pasting.over.existing.clip.zip)

mapisilafi:
Btw: I previously opted out for ability to trim the copied snippet afterwards.

mapisilafi:
This is 100% reproducible on my computer.

dozzzzer:
Alright, I was able to reproduce the issue now. It will most likely get fixed in version 3.5 @mapisilafi

PR #6080, opened 2024-03-01
Attempt to fix crash when copy-pasting between projects

vsverchinsky:
Resolves: #5792

The bug seem to happen because `WaveChannel` intervals counting and interval access may happen on diffrenet objects due to inderection that may happen in `Channel::ReallyDoGetChannelGroup` when interval iterator dereferencing tries to access interval via `Channel::GetInterval`

NB: to reproduce bug you'll need to completely replace existing clip in the track with the pasted one

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
@@ -977,6 +977,14 @@ void WaveformView::DoDraw(TrackPanelDrawingContext &context, size_t channel,
    auto pLeader = *track.GetHolder()->Find(&track);
    assert(pLeader->IsLeader());
 
+   if(&track != pLeader && track.NIntervals() != pLeader->NIntervals())
+      //We don't have a guarantee that left and right channels have same number
+      //of intervals. Sometimes pasting a large amount of data may trigger
+      //UI event loop updates. When that happens it can well be that one
+      //channel has finished pasting while the other hasn't.
+      //TODO: The check would become unnecessary once wave-clip-refactoring is complete. 
+      return;
+
    for (const auto pInterval :
       static_cast<const WaveTrack*>(pLeader)->GetChannel(channel)->Intervals()
    ) {


libraries/lib-channel/Channel.h
```diff
@@ -607,7 +607,7 @@ class CHANNEL_API ChannelGroup
 
 inline size_t Channel::NIntervals() const
 {
-   return GetChannelGroup().NIntervals();
+   return ReallyDoGetChannelGroup().NIntervals();
```
saintmatthieu, line 610: What is it that I'm missing ...
Before this change:
```cpp
const ChannelGroup &Channel::GetChannelGroup() const
{
   assert(FindChannelIndex() >= 0);
   return DoGetChannelGroup();
}
```
After this change:
```cpp
ChannelGroup &Channel::ReallyDoGetChannelGroup() const
{
   return DoGetChannelGroup();
}
```
So what difference does it make?
I haven't been able to reproduce that bug. I suppose the clips don't actually need to be hours long? Mine were just a few seconds.<

vsverchinsky: > What is it that I'm missing ...
`ReallyDoGetChannelGroup` is a virtual function 😎 and it's overwritten in `WaveChannel` :
```
ChannelGroup &WaveTrack::ReallyDoGetChannelGroup() const
{
   const Track *pTrack = this;
   if (const auto pOwner = GetHolder())
      pTrack = *pOwner->Find(this);
   const ChannelGroup &group = *pTrack;
   return const_cast<ChannelGroup&>(group);
}
```
GetHolder()->Find(this) may return track that is in fact different from this and that contains different amount of intervals.

> I haven't been able to reproduce that bug. I suppose the clips don't actually need to be hours long? Mine were just a few seconds.

For me crash occured on repaint event. So probably the length of the pasted clip is a important to reproduces this bug, because when you paste long clip UI thread isn't blocked and repaint events may slip into copying process.

saintmatthieu: Ok, so there _is_ a need for the track to be stereo for the crash to happen, right? Paste happens, but the process is slow and repainting happens before the left channel is finished copied. At this stage, left channel has size 1, and right channel, 0, but the painting thinks both have size 1. Could that be the problem? I came across that problem in the recent past.

saintmatthieu: Actually what I said doesn't make sense. If asking `NIntervals()` on the right channel of that yet incomplete stereo track, the painting would have got 0, which would be the truth ... Now it would say `1`, but that wouldn't be the truth anymore. So I wonder if fixing this place wouldn't break another one.
Do you know exactly which line crashed? I'll try to reproduce.

saintmatthieu: Caught it: 
image

vsverchinsky: Hm, I think you are right and the hotfix you are suggesting is correct and mine is not. I didn't considered that the problem comes from invariant violation. Obviously track could be inspected while being modified.

Paul-Licameli: I see Matt's long comment.  I expect this crash is not reproducible in the wave-clip-refactoring branch because of changes from channel-major to interval-major in pasting.

If this hot fix is merged to master, and master is then merged to wave-clip-refactoring to keep it current, the fix should then be reverted in that branch.

saintmatthieu (approved):

Issue #5156, opened 2023-09-05
Appending buffers to alternate channels, with exception safety and preservation of clip alignment


Paul-Licameli:
Task breakdown:
- [x] Weaken the stated invariant on `WaveClip::mSequences`: the vector of sequences have non-increasing lengths (right channel is not longer than left, assuming two channels)
   - [x] change `CheckInvariants` to match that
   - [x] remove comment in `GetNumSamples`
- [x] Do not assume equal append buffer lengths either
   - [x] So rewrite `WaveClip::GetAppendBufferLen` and functions that use it
   - [x] And  `WaveClipWaveformCache::GetWaveDisplay` needs the maximum length to paint both channels correctly 

QA:
- [x] Drawing of stereo tracks (wave or spectrogram) during recording always shows channels as having same length
- [x] Recording to stereo tracks, including tests of exception handling when external drive runs out of storage
- [x] One of these: Change Speed, Paulstretch, Compressor, Mix Stereo To Mono, Change Tempo, Sliding Stretch, EQ, Noise Reduction
- [x] Tone, Chirp, or Noise
- [x] A Nyquist effect
- [x] Import, any file format

chinakov:
@Paul-Licameli - quick clarification, I see that the task breakdown checklist items are not checked. Does this not block the QA progress? Just making sure.

Paul-Licameli:
> @Paul-Licameli - quick clarification, I see that the task breakdown checklist items are not checked. Does this not block the QA progress? Just making sure.

Those were notes to myself months ago and for possible delgation to another developer.  You can ignore them.

chinakov:
@Paul-Licameli - thank you!

PR #6089, opened 2024-03-03
Wave clip invariants

Paul-Licameli:
Resolves: #5156

Depends on:
- #6071

WaveClips are not yet ever stereo, but this is a preparation.

Weaken the WaveClip invariant, not to require equal lengths of all sequences.

That stronger condition is called the "strong" invariant.  It is required for certain mutating
operations that insert or delete in the sequences.  There is a method to repair the
WaveClip when this is not so by appending silence to the shorter sequence.

The strong condition will be violated often, as appending to stereo tracks is still
done in many places in a channel-major fashion, alternating channels.

In most cases this temporary misalignment is soon repaired.  In generators (including
Nyquist), new mono tracks are zipped, and the zipping function refuses to do it if
it would violate the strong condition.

The one possible gap is in exception handling during recording, which attempts to
save as much as possible of the recording.  There is an attempt now to repair the
misalignment in that contingency too.

Recommended:

QA:
- [x] Record
- [x] Pencil tool
- [x] Generators, and some destructive effects
- [x] cut,  with and without the making of cut lines
- [x] paste, including pasting exactly at the start or end of a clip
- [x] Lengthening sync Lock adjustment
- [x] Silence command (not generator) or the Silence edit toolbar button
- [x] Recording interrupted by exhaustion of drive space loses at most one block (about 5.9 seconds with default sample format and rate)

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -475,17 +475,25 @@ void WaveClip::UpdateEnvelopeTrackLen()
 std::shared_ptr<SampleBlock>
 WaveClip::AppendNewBlock(constSamplePtr buffer, sampleFormat format, size_t len)
 {
+   return mSequences[0]->AppendNewBlock( buffer, format, len );
+}
+
+/*! @excsafety{Strong} */
+std::shared_ptr<SampleBlock>
+WaveClip::AppendLegacyNewBlock(constSamplePtr buffer, sampleFormat format, size_t len)
+{
    // This is a special use function for legacy files only and this assertion
-   // does not need to be relaxed
+   // does not need to be relaxed.  The clip is in a still unzipped track.
    assert(GetWidth() == 1);
    return mSequences[0]->AppendNewBlock( buffer, format, len );
 }
 
 /*! @excsafety{Strong} */
-void WaveClip::AppendSharedBlock(const std::shared_ptr<SampleBlock> &pBlock)
+void WaveClip::AppendLegacySharedBlock(
+   const std::shared_ptr<SampleBlock> &pBlock)
 {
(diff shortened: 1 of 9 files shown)


libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -719,8 +727,19 @@ void WaveClip::WriteXML(size_t ii, XMLWriter &xmlFile) const
...
+bool WaveClip::Paste(double t0, const WaveClip& o)
+{
+   const WaveClip *pOther = &o;
+   WaveClipHolder dup;
+   if(!o.StrongInvariant()) {
+      assert(false); // precondition not honored
+      // But try to repair it and continue in release
+      auto dup = std::make_shared<WaveClip>(o, o.GetFactory(), true);
+      dup->RepairChannels();
+      pOther = dup.get();
```
saintmatthieu, line 739: Looks like a bug: `dup` gets destroyed when this scope closes, making `pOther` unusable.

saintmatthieu (changes requested):
`bool WaveClip::Paste(double t0, const WaveClip& o)` seems to have a bug.

libraries/lib-wave-track/WaveClip.cpp
Paul-Licameli: Sharp eyes

Paul-Licameli:
Rebased unchanged onto updated dependency

Paul-Licameli:
Rebased unchanged onto updated dependency

Paul-Licameli:
Rebased unchanged onto updated base branch

dwlegg:
This may be relevant:
https://bugzilla.rpmfusion.org/show_bug.cgi?id=7056

ndokos:
Also, issue #9120 may be relevant.

Issue #4999, opened 2023-08-06
Import of audio file now (significantly?) slower


deus-e-x:
### Bug description

Hello dear Audacity team,

this is actually not a bug report but a heads-up to check whether a regression has crept in (i.e. a "just-in-case" report of sorts).

This is a comparison of the below-listed nightly builds:
audacity-win-3.4.0-alpha-20230804+ca2098e-x64-msvc2022
audacity-win-3.4.0-alpha-20230731+c8fda52-x64-msvc2022

Importing a FLAC audio file, 248 MB in size, takes about one second longer with the latest nightly build from 2023-08-04 compared to the nightly build from 2023-07-31 and earlier nightlies.

On the contrary, the export of mentioned 248 MB FLAC file is about half a second faster with the new audio export dialog in the current nightly from 2023-08-04.

Thank you for your ongoing support and further development of Audacity, guys.

### Steps to reproduce

1. Import a FLAC audio file of reasonable file size (e.g. my example is 248 MB) using the open dialog (Ctrl+O).
2. Measure the time it takes to import once you clicked/pressed the open button to execute the loading process. 
3. Repeat several times for reproducible results.

### Expected behavior

The time it takes to import the above-mentioned audio file (FLAC, 248 MB) would be at the same speed as in previous builds if that is possible with regards to the current restructuring of certain processes of Audacity.

### Actual behavior

As described above importing an example audio file (FLAC, 248 MB) takes ~1 second longer compared to older recent nightly builds.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

SteveDaulton:
I don't see a significant difference when testing with a 100 MB FLAC file (both versions take about 3 seconds), but with a large (75 MB) MP3 file, the latest alpha version is about 38% slower than Audacity 3.1. (Tested on Xubuntu  22.04, i7 + SSD)

dozzzzer:
In my case (M1 Pro, macOS 13.0) the import of a 2h MP3 file is **significantly** slower: ~7.5s in 3.3.3 against ~17.7s in the latest `master`.

DavidBailes:
> In my case (M1 Pro, macOS 13.0) the import of a 2h MP3 file is **significantly** slower: ~7.5s in 3.3.3 against ~17.7s in the latest `master`.

Windows 11, importing a 2h mp3 is roughly twice as slow in latest master vs 3.3.3 (roughly 14s vs 7s)
This isn't going to get fixed for 3.4?

[...]

PR #6107, opened 2024-03-06
Import slowdown fix

vsverchinsky:
Resolves: #4999 

There were two major reasons for import slowdown: 
1) Channel iterator dereference operation that was called repeatedly for each chunk/block in sequence adds a noticeable overhead to the import. As `TrackList` doesn't change during import channels could be enumerated in advance and stored in local variable.
2) Overhead added by data preservation performed by `Transaction` object in `WaveClip`. This is solved by introducing `AppendUnsafe` that omits this kind of safety, which not seem to be useful in case of import because data is appended to a temporary object that will be destroyed in case if `Sequence` throws an exception.

Recommended:

### DIFF
--- a/libraries/lib-import-export/ImportUtils.cpp
+++ b/libraries/lib-import-export/ImportUtils.cpp
@@ -63,13 +63,11 @@ void ImportUtils::FinalizeImport(TrackHolders& outTracks, TrackListHolder trackL
    outTracks.push_back(std::move(trackList));
 }
 
-void ImportUtils::ForEachChannel(TrackList& trackList, const std::function<void(WaveChannel&)>& op)
+std::vector<std::shared_ptr<WaveChannel>> ImportUtils::GetAllChannels(TrackList& trackList)
 {
-   for(auto track : trackList.Any<WaveTrack>())
-   {
+   std::vector<std::shared_ptr<WaveChannel>> channels;
+   for(const auto track : trackList.Any<WaveTrack>())
       for(auto channel : track->Channels())
-      {
-         op(*channel);
-      }
-   }
+         channels.push_back(std::move(channel));
+   return channels;
 }
--- a/libraries/lib-import-export/ImportUtils.h
+++ b/libraries/lib-import-export/ImportUtils.h
@@ -41,9 +41,8 @@ public:
    
    static void ShowMessageBox(const TranslatableString& message, const TranslatableString& caption = XO("Import Project"));
 
-   //! Iterates over channels in each wave track from the list
    static
-   void ForEachChannel(TrackList& trackList, const std::function<void(WaveChannel&)>& op);
+   std::vector<std::shared_ptr<WaveChannel>> GetAllChannels(TrackList& trackList);
 
    //! Flushes the given channels and moves them to \p outTracks
    static
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -536,6 +536,22 @@ bool WaveClip::Append(constSamplePtr buffers[], sampleFormat format,
    return appended;
 }
 
+bool WaveClip::AppendUnsafe(constSamplePtr buffers[], sampleFormat format, size_t len, unsigned stride,
+   sampleFormat effectiveFormat)
+{
+   size_t ii = 0;
+   bool appended = false;
+   for (auto &pSequence : mSequences)
+      appended =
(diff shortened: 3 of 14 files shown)


Paul-Licameli:
A quick look at this suggests to me that it will be difficult to merge it with the wave-clip-refactoring branch, which includes simplifications that eliminate the use of TrackList in import.  I also see another use of `TrackList::Channels` but I'm trying to eliminate almost all uses of that.

What is the explanation for slow import?  How does this pull request fix it?

libraries/lib-import-export/ImportUtils.cpp
```diff
@@ -63,13 +63,11 @@ void ImportUtils::FinalizeImport(TrackHolders& outTracks, TrackListHolder trackL
...
 {
-   for(auto track : trackList.Any<WaveTrack>())
-   {
+   std::vector<std::shared_ptr<WaveChannel>> channels;
+   for(const auto track : trackList.Any<WaveTrack>())
       for(auto channel : track->Channels())
-      {
-         op(*channel);
-      }
-   }
```
Paul-Licameli, line 74: Was the overhead of constructing and destroying `std::function` objects really significant? -- then maybe an alternative is that you make this an inline function in the header file, and make the type of `op` a deduced function template argument.

vsverchinsky: > Was the overhead of constructing and destroying std::function objects really significant?

No, the overhead was largely due to interator dereferencing operation:
изображение

vsverchinsky:
> What is the explanation for slow import? How does this pull request fix it?

Explained this in PR description

Paul-Licameli:
I think a different fix could be developed that enlarges the scope of the transaction over many Append calls, and it may also be applied in destructive effects.

Paul-Licameli:
The correct thing in the refactoring branch may be to remove Transaction from Append.

See the latest PR in QA.  I distinguish strong and weak invariants.  Append no longer needs to conserve the strong invariant of equal length sequences.

Wave channel iteration will be simpler too.

Paul-Licameli:
Do you get acceptable improvement with ApoendUnsafe alone?

vsverchinsky:
> I think a different fix could be developed that enlarges the scope of the transaction over many Append calls, and it may also be applied in destructive effects.

I did think similar concept too, but considered it to be too risky to attempt it now

> Do you get acceptable improvement with ApoendUnsafe alone?

Not to me to decide, but both have significant impact on import performance. Approx 30% of losses covered by `ForEachChannel` change.

Paul-Licameli:
> > I think a different fix could be developed that enlarges the scope of the transaction over many Append calls, and it may also be applied in destructive effects.
> 
> I did think of similar concept too, but considered it to be too risky to attempt it now
> 
> > Do you get acceptable improvement with ApoendUnsafe alone?
> 
> Not to me to decide, but both have significant impact on import performance. Approx 30% of losses covered by `ForEachChannel` change.

I am surprised if the iterator really explains the second.  Are you sure it was not std::function overhead?  The observation that this rewrite improves performance does not distinguish them.  Trying my alternative with a template function could do that.

petersampsonaudacity:
I just tested on W10 with 3.4.2 versus @vsverchinsky 's branch build for this: audacity-win-3.5.0-alpha-20240306+52fb883-x64

I imported a 2 hour stereo MP3 266MB

It took slightly longer on the new branch build than on 3.4.2 - 1m20 versus 1m06

Paul-Licameli:
> > > I think a different fix could be developed that enlarges the scope of the transaction over many Append calls, and it may also be applied in destructive effects.
> > 
> > 
> > I did think of similar concept too, but considered it to be too risky to attempt it now
> > > Do you get acceptable improvement with ApoendUnsafe alone?
> > 
> > 
> > Not to me to decide, but both have significant impact on import performance. Approx 30% of losses covered by `ForEachChannel` change.
> 
> I am surprised if the iterator really explains the second. Are you sure it was not std::function overhead? The observation that this rewrite improves performance does not distinguish them. Trying my alternative with a template function could do that.

@vsverchinsky maybe a foolish question -- surely though, your performance data were from a release build?

Maybe debug builds have a lot of expensive assertions that would distort the measurement.

vsverchinsky:
> Maybe debug builds have a lot of expensive assertions that would distort the measurement.

No, I tried it with release builds.

vsverchinsky:
Closed in favor of https://github.com/audacity/audacity/pull/6094

Issue #5924, opened 2024-02-07
Labels: Export as subtitle formats


LWinterberg:
### Your idea

Labels should be exportable as subtitle formats, such as VTT or subrip (SRT)

### Problem to be solved

The current label export format looks like this: 

```
105.000000	106.200000	 text1
106.200000	109.160000	 text2
```

This is a perfectly fine format, but it isn't supported by anything else. This means that if you're using the transcription plugin, or otherwise used Audacity to subtitle things, you can't easily use it elsewhere. 

YouTube supports a wide range of subtitle formats, the easiest ones probably being [WebVTT](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API): 

```
0:01:45.000 --> 0:01:46.200
text1

0:01:46.200 --> 0:01:49.160
text2
```

The conversion between those two formats isn't entirely trivial to the end user (especially the sss.µµµµµµ to hh:mm:ss.ttt conversion) so being able to export in this format would make things a lot more viable and fit well into subtitle workflows. 

### Prior art

This has an ancient PR in https://github.com/audacity/audacity/pull/394

leodom:
Hello, thanks for the work on this labels export !

I am using version 3.4.2, how can I install/use the export label plugin ?

LWinterberg:
It's not a plugin, it's part of the upcoming 3.5 release which will release later this month.

PR #394, opened 2019-11-24
Add support for labels in the SubRip and WebVTT formats

Pokechu22:
Allows exporting labels in the [SubRip](https://en.wikipedia.org/wiki/SubRip) and [WebVTT](https://en.wikipedia.org/wiki/WebVTT) formats (and importing labels in SubRip format as well, but not WebVTT).  This is a resubmission of #383 that has been cleaned up a ton.  There's [a wiki article on the use for this](https://wiki.audacityteam.org/wiki/Movie_subtitles_%28*.SRT%29).

Improvements from before:
* Split into 3 commits to make it easier to review (mainly because of indentation changes)
* The new formats are controlled by `EXPERIMENTAL_SUBRIP_LABEL_FORMATS` (and I've made sure that everything compiles with it enabled and with it disabled)
* Timestamps are parsed using `wxDateTime`, based on what @kempniu did in #170.
* Deciding which format to use now happens in a single method (`LabelTrack::FormatForFileName`) so only one place needs to be modified, instead of 4 locations in 2 files.
* I've fixed all of the references to "WebVVT" and replaced them with "WebVTT" (that took me an embarrassingly long time to notice).  "VTT" stands for "Video Text Tracks".

There are a few known issues:
* WebVTT files cannot be imported.  The format is much more complicated ([see the spec](https://w3c.github.io/webvtt/)) and supports content that does not directly translate into labels (such as comments which are not displayed and do not have a timestamp, and regions that indicate positioning of cues).
* Both SubRip and WebVTT support multiple lines in a cue, but labels do not currently support multiple lines (and the current txt format doesn't have the ability to do it).  When importing SubRip files, I simply join the lines with spaces. (#170 used `|` based on [MicroDVD](https://en.wikipedia.org/wiki/MicroDVD), which I have no experience with.)
* Exporting multiple tracks at the same time leads to an invalid file &mdash; it effectively concatenates what happens if you export each individually.  The identifier (which is just a counter here) is reset to `1` for each file, and the `WEBVTT` header is repeated.  Although both of those are fixable, WebVTT also requires that the start time "must be greater than or equal to the start time offsets of all previous cues in the file" ([§4.1, "WebVTT cue timings"](https://w3c.github.io/webvtt/#webvtt-cue-timings)).  Resolving that requires merging the two files in sorted order and that would require a more substantial rework of how label track I/O works (which would be good eventually, as the labels editor would also be sorted by timestamp instead of track and then timestamp).

PR #6139, opened 2024-03-14
Add support for labels in the SubRip and WebVTT formats

lbordowitz:
Resolves: #5924

This is a rebase of https://github.com/audacity/audacity/pull/394 . The original author is @Pokechu22 and the entirety of this code change should be credited as such.

- [x] I signed [CLA](https://www.audacityteam.org/cla/) - Pokechu22 did, too

Recommended:

### DIFF
--- a/src/LabelTrack.cpp
+++ b/src/LabelTrack.cpp
@@ -46,6 +46,10 @@ for drawing different aspects of the label and its text box.
 #include "TimeWarper.h"
 #include "AudacityMessageBox.h"
 
+#ifdef EXPERIMENTAL_SUBRIP_LABEL_FORMATS
+const FileNames::FileType LabelTrack::SubripFiles{ XO("SubRip text file"), { wxT("srt") }, true };
+#endif
+
 LabelTrack::Interval::~Interval() = default;
 
 std::shared_ptr<ChannelInterval>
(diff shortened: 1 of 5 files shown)


src/LabelTrack.cpp
```diff
@@ -715,6 +731,13 @@ LabelFormat LabelTrack::FormatForFileName(const wxString & fileName)
 /// Import labels, handling files with or without end-times.
 void LabelTrack::Import(wxTextFile & in, LabelFormat format)
 {
+#ifdef EXPERIMENTAL_SUBRIP_LABEL_FORMATS
+   if (format == LabelFormat::WEBVTT) {
+      ::AudacityMessageBox( XO("Importing WebVTT files is not currently supported.") );
```
LWinterberg, line 736: This would be something to improve in a followup PR. I [see](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API) that webVTT does some fancy stuff, so the importer would need to be told to strip away everything that we can't deal with, eg. 

```

WEBVTT

STYLE
::cue {
  background-image: linear-gradient(to bottom, dimgray, lightgray);
  color: papayawhip;
}

00:00:00.000 --> 00:00:04.000 position:10%,line-left align:left size:35%
Where did <b>he</b> go?
```

would need to be converted to

```
00:00:00.000 --> 00:00:04.000 
Where did <b>he</b> go?
```

and from there converted to a label. 

The tags (`<b>, <c>, <i>, <u>`, etc) inside the text can stay as they may be things you'd want to set or edit in label form, too - for example to add emphasis to certain words when a character does so too. The style would need to be discarded for the time being. 

A second followup would be to make a new exporter for the labels in a similar vein to the Export Audio exporter, which then could have custom fields for styles and such in case of vtt.

src/LabelTrack.cpp
```diff
@@ -422,85 +428,171 @@ ImportExportPrefs::RegisteredControls reg{ wxT("LabelStyle"), AddControls };
...
          throw BadFormatException{};
 
-      token = toker.GetNextToken();
-      double f1;
-      if (!Internat::CompatibleToDouble(token, &f1))
+      wxString timestamp = file.GetLine(index++);
+      // Parsing data of the form 'HH:MM:SS,sss --> HH:MM:SS,sss'
+      // Assume that the line is in exactly that format, with no extra whitespace
+      // and less than 24 hours.
+      if (timestamp.length() != 29)
```
LWinterberg, line 534: While I have never seen a vtt or srt file do anything other than this, I'm a bit uncomfortable with such an assumption. I'd prefer something more robust here.

LWinterberg (approved):
This looks good to me overall. There's some more improvements which may be something for the future. As-is, it adds a valuable feature with the label export in a pretty straight-forward manner, which in my book is good for a merge..

PR #6153, opened 2024-03-21
5799: Catch failing AudioUnit UI loading

### LINKED ISSUE #5799: (macOS) Crash when launching AUMultibandCompressor
### Bug description

Attempting to use AUMultibandCompressor results in crash every time

### Steps to reproduce

1 - open audio file
2- select audio 
3 - open AUMultibandCompressor

### Expected behavior

AUMultibandCompressor should appear, instead app crashes

### Actual behavior

crash

### Audacity Version

### Operating system

macOS

### Additional context

_No response_

saintmatthieu:
Resolves: #5799 5799

Recommended:

### DIFF
--- a/src/effects/audiounits/AUControl.mm
+++ b/src/effects/audiounits/AUControl.mm
@@ -291,10 +291,14 @@ void AUControl::CreateCocoa()
             if (auto factoryClass = [bundle classNamed:viewClass])
                // Create an instance of the class
                if (id factoryInst = [[[factoryClass alloc] init] autorelease])
-                  // Create the view, suggesting a reasonable size
-                  if ((mView = [factoryInst uiViewForAudioUnit:mUnit
-                                                 withSize:NSSize{800, 600}]))
+                 // Create the view, suggesting a reasonable size
+                 try {
+                   if ((mView =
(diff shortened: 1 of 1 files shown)


src/effects/audiounits/AUControl.mm
```diff
@@ -291,10 +291,14 @@ - (void)cocoaViewResized:(NSNotification *)notification
             if (auto factoryClass = [bundle classNamed:viewClass])
                // Create an instance of the class
                if (id factoryInst = [[[factoryClass alloc] init] autorelease])
-                  // Create the view, suggesting a reasonable size
-                  if ((mView = [factoryInst uiViewForAudioUnit:mUnit
-                                                 withSize:NSSize{800, 600}]))
+                 // Create the view, suggesting a reasonable size
+                 try {
+                   if ((mView =
```
crsib, line 296: I don't understand why this works. I assume, that you try to catch Obj-C exception, and the syntax is different https://stackoverflow.com/questions/4648952/objective-c-exceptions

saintmatthieu: I guess because objective-c++ exceptions are derived from c++ exceptions.

crsib (approved):

Issue #6159, opened 2024-03-22
Add methods to sync with external programs


LWinterberg:
### Your idea

Add a (or better yet: several) methods to communicate from within Audacity where it's currently at. This can be
* a scripting solution (GetInfo)
* Industry standard data, such as
    *  SMPTE, 
    * LTC, 
    * MTC

### Problem to be solved

Until we end up adding a video track natively, this would fill the gap to sync up external players. And even afterwards, this serves a purpose for professional use cases, for example to keep audio gear in sync with Audacity, where Audacity would act as a clock giver.

SteveDaulton:
Also: https://github.com/audacity/audacity/issues/5038
(which gives you  a "pror art")

Issue #5418, opened 2023-10-24
Add an entry in the menu to generate LTC wave files


ycollet:
### Your idea

LTC signals are used to synchronize audio / video:
https://github.com/x42/ltc-tools

With a LTC wave file, you can send this signal to xjadeo and have an audio track from audacity synchronized with a video played by xjadeo. xjadeo is used by ardour to manage video.

Repository of xjadeo:
https://github.com/x42/xjadeo

### Problem to be solved

Synchronization of audio file with a video.

### Prior art

Ardour is using xjadeo and sync video via MTC (Midi Time Code) signals.

jonata:
Can you explain more about this? Do you want a LTC Wave Generator inside Audacity, or you want to control/communicate with XJadeo somehow?

ycollet:
For me, the kickest way to be able to control the position in a movie is to have a LTC track generated via audacity.
So, a LTC wave generator inside audacity is my first choice :)

jonata:
Yes, that's what I want too. But I think the way we can do it is the same as Ardour, using MTC, as Audacity already have some portmidi architecture. Maybe I need to research more, but as fas as I understand, using a wave LTC would require sending the audio signal (this specific timecode wave) to somewhere. In the case of XJadeo, it should be send using Jack, but I'm not sure if that's available in all platforms. I will look into MTC.

On a related note, I made a pull request (https://github.com/audacity/audacity/pull/6152) to enable Audacity send the current position when requested. In this way, we could control an external player getting this information from Audacity. Not the best solution, though.

jonata:
Good news! Yes, MTC works:

[Gravação de tela de 2024-03-22 00-00-17.webm](https://github.com/audacity/audacity/assets/908065/0ea17c1f-8780-464f-9d22-524613eb306b)

Now I'll figure out how to integrate it in Audacity properly.

ycollet:
Excellent ! Thanks a lot !

devmaiter:
Hi, I'd like to work on this if it's still open

Issue #5038, opened 2023-08-15
Support for "Jack Transport"


SteveDaulton:
### Your idea

This feature request will mostly benefit Linux users (because Jack Audio System is  rarely used on Windows or MacOS).

Jack Transport provides the ability to synchronise multiple Jack clients on Linux, similar to Steinberg's [ReWire](https://steinberg.help/cubase_artist/v11/en/cubase_nuendo/topics/rewire/rewire_c.html), but open source and works with many multimedia applications.

An example setup might be to run Hydrogen drum machine, synchronised with Audacity, so that you can start playing or recording in Audacity from an arbitrary point in the project, and Hydrogen drum machine will automatically synchronise to the same position.

Another example might be to synchronise video playback in XJADEO with audio transport in Audacity.

Another example might be to synchronise Reaper or Ardour with Audacity.

Another example might be all of the above at the same time!

### Problem to be solved

While it is possible to synchronise many multi-media apps on Linux using Jack Audio System, this functionality is missing from Audacity.

### Prior art

Ardour, Blender, ecasound, Epichord, Freecycle, fsthost, Hydrogen, JAMin, jpmidi, xjadeo, and more.

### Additional context

See also https://github.com/audacity/audacity/issues/2167

LWinterberg:
A cross-platform alternative for this would be Ableton Link: https://github.com/Ableton/link

trebmuh:
So is [JACK](https://jackaudio.org/).

luzpaz:
> A cross-platform alternative for this would be Ableton Link: https://github.com/Ableton/link

Here's an attempt over at mixxxdj https://github.com/mixxxdj/mixxx/pull/3783

Issue #2167, opened 2021-11-19
Ability for Audacity to import and show video to facilitate dubbing the audio track


petersampsonaudacity:
**Is your feature request related to a problem? Please describe.**
There is no ability with Audacity to import and show videos, this would be useful to be able to dub them. 

**Describe the solution you'd like**
Functionality in Audacity to import and show videos to facilitate dubbing the audio track

**Describe alternatives you've considered**
None

**Additional context**
This post is "nudged" by this recent post on the Forum:
https://forum.audacityteam.org/viewtopic.php?f=20&t=121667&p=439781

But I note that it is a long-standing Feature Request in the Audacity Wiki (one of the most popular with 47 votes).
_**Update:** now 49 votes - added a Forum user's plus mine_
https://wiki.audacityteam.org/wiki/Feature_Requests#newformats
Where it states:
>**Import/play video for synchronized soundtrack editing** (47 votes 09Dec18) 
>No need for Audacity to edit the video. Alternatively, Audacity could send/receive 
>SMPTE or MTC timecode data e.g. to VLC or mplayer.

SteveDaulton:
**Describe alternatives you've considered**

For Linux users, "Jack Transport" support would allow the use of XJadeo (http://xjadeo.sourceforge.net/main.html)
Jack Transport would also allow many other benefits for synchronizing with other apps.

Penikov:
@petersampsonaudacity Hmm! Nice Idea! I think we can ask @Tantacrul for that feature in future releases.

Tantacrul:
I definitely think this is a BIG win for Audacity. We've already done some preliminary design work on it. I don't quite know where it will fit in our backlog though. But certainly on our radar, for sure.

petersampsonaudacity:
> I definitely think this is a BIG win for Audacity. We've already done some preliminary design work on it.

Excellent  👍

LWinterberg:
duplicate of https://github.com/audacity/audacity/issues/1023

PR #6164, opened 2024-03-24
Initial implementation of MTC

jonata:
Resolves: #6159, #5418, #37, #2167, #1023, helps #5038

This enables Audacity to send MTC through PortMidi. Of course, I'm not sure if the TimeToolBar is the best place to put it, but for the user perspective, it make sense that it would be "sent" by the time toolbar. This adds a checkbox on a context menu and also adds an option on Playback panel on Preferences. We would need to find a way to expose the video framerate selection for the user, though. Another thing to consider is that it's selecting the "Default" device to send. Maybe we should expose an option for the user to choose..?

To test, open the audio of a video in Audacity and open XJadeo. In XJadeo, select the Sync mode to MTC (Portmidi) and enable the option on Audacity. If video is not sinchronized, that's probably because the video framerate is hardcoded to 29.97. Launching XJadeo with the option "-M 1" should work around this.

Of couse, PLEASE, give me feedback on the code as I'm a Python developer, not C++. Thanks!!

Recommended:

### DIFF
--- a/src/prefs/PlaybackPrefs.cpp
+++ b/src/prefs/PlaybackPrefs.cpp
@@ -152,6 +152,7 @@ void PlaybackPrefs::PopulateOrExchange(ShuttleGui & S)
          S.TieCheckBox(XXO("Always scrub un&pinned"),
             {UnpinnedScrubbingPreferenceKey(),
              UnpinnedScrubbingPreferenceDefault()});
+         S.TieCheckBox(XXO("Send MTC (MIDI Time Code)"), {"/AudioIO/SendMTC", true});
       }
       S.EndVerticalLay();
    }
--- a/src/toolbars/TimeToolBar.cpp
+++ b/src/toolbars/TimeToolBar.cpp
@@ -15,6 +15,7 @@
 
 // For compilers that support precompilation, includes "wx/wx.h".
 #include <wx/wxprec.h>
+#include <wx/log.h>
 
 #include <wx/setup.h> // for wxUSE_* macros
 
--- a/src/toolbars/TimeToolBar.cpp
+++ b/src/toolbars/TimeToolBar.cpp
@@ -41,7 +41,8 @@ IMPLEMENT_CLASS(TimeToolBar, ToolBar);
 // the Jaws screen reader script for Audacity.
 enum {
    TimeBarFirstID = 2800,
-   AudioPositionID
+   AudioPositionID,
+   SendMTCID
 };
 
 BEGIN_EVENT_TABLE(TimeToolBar, ToolBar)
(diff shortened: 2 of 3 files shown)


LWinterberg:
image
Right-clicking the bulk of the time toolbar opens the time format picker; even knowing it would be there I needed a while to figure out that I'd need to right-click the place where the time display is not. I don't think it's viable to have a right-click menu in there. 

image
I have no strong opinion on where in the preferences this option goes; though given that it a) probably also should be sent during recording and b) it is called *MIDI* time code, I'd say it probably has a better home inside the "MIDI devices" section of the preferences (which may just get renamed to just "MIDI" as well). 

Code-wise, I'm not sure if this code wants to be part of the Time toolbar, maybe it should be its own little `lib-mtc` library instead. @crsib or @vsverchinsky may have stronger opinions on that.

jonata:
Thanks for the response. I agree with you. Just explaining more about my decision, but I agree with you:
- For MTC, we would need to select what is the video FPS in some way. That's why I put a dropdown menu on Time Toolbar. And as you mentioned, it's naturally hidden because in practice, not all users would need to use this feature. But the user would need to be able to select the FPS for each video he wants to sync. And I don't think a good idea to put this option only on Preferences window, as the user would need to select the FPS every time he opens a project. But maybe a dropdown menu on main menubar?
- For the settings option, as I mentioned, although it is a **MIDI** protocol, this feature is not related to "music" (as the user would expect by looking into the MIDI panel). 
- For the Toolbar placement of the code, that's mainly because I'm not familiar with the code and C++ in general. I just thought that it would be easier for the user to select there, as I noted, the FPS for the loaded video.

Thanks!!!

LWinterberg:
We could structure the menu similar to the snapping options right next-door and put things into submenus instead:

image

This design assumes you still want your #316 in still, and further assumes that a 30fps timecode would be added. When switching between different formats, the MTC would need to *slightly* lie (eg sending NTSC drop frames even though non-drop frames are selected) but overall it'd keep things in sync.

src/prefs/PlaybackPrefs.cpp
```diff
@@ -152,6 +152,7 @@ void PlaybackPrefs::PopulateOrExchange(ShuttleGui & S)
          S.TieCheckBox(XXO("Always scrub un&pinned"),
             {UnpinnedScrubbingPreferenceKey(),
              UnpinnedScrubbingPreferenceDefault()});
+         S.TieCheckBox(XXO("Send MTC (MIDI Time Code)"), {"/AudioIO/SendMTC", true});
```
Paul-Licameli, line 155: I would make a named BoolSetting object in this file, declared in a header, to avoid repetition of the string literal.  See other examples

src/toolbars/TimeToolBar.h
```diff
@@ -64,6 +66,20 @@ class TimeToolBar final : public ToolBar
...
+   void CreateDropdownMenu();
+   void OnCheckboxItemSelected(wxCommandEvent &event);
+   void OnRightClick(wxMouseEvent &event);
+
+   PortMidiStream *portMidiStream;
+   PmError portMidiError;
+   bool isPortMidiInitialized = false;
+   void InitializePortMidi();
+   void TerminatePortMidi();
+   void UpdatePortMidiAccordingToPreferences();
```
Paul-Licameli, line 80: This function isn’t reached when the state of the checkbox is changed in PlaybackPrefs.

You could call it from the Commit method of the preference page.

Again, move the functions that do all the real business to a file whose job is not GUI.  Neither the toolbar nor preference page should depend on the other.

src/toolbars/TimeToolBar.cpp
```diff
@@ -386,6 +463,31 @@ void TimeToolBar::OnIdle(wxIdleEvent &evt)
...
+   int minutes = remainingSeconds / 60;
+   int seconds = remainingSeconds % 60;
+   int frames = static_cast<int>((audioTime - static_cast<int>(audioTime)) * 29.97); // 29.97 hardcoded... TODO
+
+   // Sending MIDI Time Code (MTC) message (Full Frame: 0xF0 0x7F 0x7F 0x01 0x01 hours minutes seconds frames 0xF7)
+   unsigned char message[] = {0xF0, 0x7F, 0x7F, 0x01, 0x01, static_cast<unsigned char>(hoursWithFPS), 
+                              static_cast<unsigned char>(minutes), static_cast<unsigned char>(seconds), 
+                              static_cast<unsigned char>(frames), 0xF7};
+   
+   Pm_WriteSysEx(portMidiStream, 0, message);
```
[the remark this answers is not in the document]
jonata: Where could it trigger the MTC send function?

LWinterberg:

Issue #407, opened 2020-01-03
Get selection time via pipe / NyquistPrompt command


buggymcbugfix:
I'm trying to get the start and end times of the current selection for scripting purposes but I have only managed to _set_ the time via `SelectTime:`.

I am looking for some command `GetTime:` which returns the information that is displayed in the audio position toolbar:
<img width="694" alt="Screenshot 2020-01-03 at 02 12 35" src="https://user-images.githubusercontent.com/17603372/71703795-8ec4da80-2dce-11ea-918b-8764bb2a9d61.png">

buggymcbugfix:
I am aware that Nyquist can get this information, but I have not managed to funnel this information out of Audacity via the pipe.

SteveDaulton:
I don't think it's possible to get the selection times "down the pipe", but
also, I've not yet seen a case where that is really needed.
I'd suggest that you write into the Audacity forum and describe what you
are trying to do - the forum is the main Audacity help channel.
https://forum.audacityteam.org/viewforum.php?f=69

Steve

PR #6165, opened 2024-03-24
Implemented Selection on GetInfo

jonata:
Resolves: #407

Added selection information (Start and End) when user send "GetInfo: Type=Selection" to pipe.

Recommended:

### DIFF
--- a/src/commands/GetInfoCommand.cpp
+++ b/src/commands/GetInfoCommand.cpp
@@ -17,6 +17,7 @@ This class now lists
 - Clips
 - Labels
 - Boxes
+- Selection
 
 *//*******************************************************************/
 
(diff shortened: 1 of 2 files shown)


LWinterberg (approved):

SteveDaulton:
This does not do what it says it does. An "Audio Selection" is two dimensional, having both a temporal range, and a track list. "GetInfo: Selection" only returns the temporal range.

Possible solutions:

* Rename the command "GetInfo: SelectionTimes" (or similar)
* Extend the returned value to return both the times and a list of tracks.

Also, it does not appear to work with the Nyquist "get-audio-info" command, thus requiring Nyquist users to implement their own parsing of strings in the form: `("\n{ "Start":8.70968, "End":15.1092 }" . T)`.

LWinterberg:
> "GetInfo: Selection" only returns the temporal range.

And GetInfo: Tracks returns selectedness, so if you need it you can figure it out. I think this implementation is sufficient for syncing Audacity with external tools which likely aren't interested in the internal track selection to begin with.

SteveDaulton:
> which likely aren't interested in the internal track selection to begin with.

That's a very bold assumption.

> Also, it does not appear to work with the Nyquist "get-audio-info" command

Fixed by this pull request: https://github.com/audacity/audacity/pull/6347

Issue #6170, opened 2024-03-25
Pitch and speed dialog: when open it should be reachable from the main window by pressing Alt + F6


DavidBailes:
### Bug description

With the full default shortcuts enabled, pressing Alt + F6 or Shift Alt + F6 cycles you round the main window and any open modeless dialogs. However, the Pitch and speed dialog is not included in this cycle.

### Steps to reproduce

With the full default shortcuts enabled:
1. Select a clip
2. Open the Pitch and speed dialog.
3. Press Alt + F6. The main window becomes the focus.
4. Press Alt + F6.

### Expected behavior

The Pitch and speed dialog becomes the focus.

### Actual behavior

The main window remains the focus.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

saintmatthieu:
Will be addressed with https://github.com/audacity/audacity/issues/6168 and corresponding PR - thanks @DavidBailes !

PR #6171, opened 2024-03-25
Pitch and speed dialog: make it reachable via the Alt + F6 shortcut

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6170

Problem:
With the full default shortcuts enabled, pressing Alt + F6 or Shift Alt + F6 cycles you round the main window and any open modeless dialogs. However, the Pitch and speed dialog is not included in this cycle. For this to be the case the dialog's parent should be the project window. (See in NavigationMenus.cpp, OnPrevWindow() and OnNextWindow().) This isn't the case for the Pitch and speed dialog, the parent is nullptr.

Fix:
Change the parent of the Pitch and speed dialog to be the project window.

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/PitchAndSpeedDialog.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/PitchAndSpeedDialog.cpp
@@ -156,8 +156,8 @@ void PitchAndSpeedDialog::Destroy(AudacityProject& project)
 
 PitchAndSpeedDialog::PitchAndSpeedDialog(AudacityProject& project)
     : wxDialogWrapper(
-         nullptr, wxID_ANY, XO("Pitch and Speed"), wxDefaultPosition,
-         { 480, 250 }, wxDEFAULT_DIALOG_STYLE)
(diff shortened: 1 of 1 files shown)


src/tracks/playabletrack/wavetrack/ui/PitchAndSpeedDialog.cpp
```diff
@@ -156,8 +156,8 @@ void PitchAndSpeedDialog::Destroy(AudacityProject& project)
 
 PitchAndSpeedDialog::PitchAndSpeedDialog(AudacityProject& project)
     : wxDialogWrapper(
-         nullptr, wxID_ANY, XO("Pitch and Speed"), wxDefaultPosition,
-         { 480, 250 }, wxDEFAULT_DIALOG_STYLE)
+         FindProjectFrame(&project), wxID_ANY, XO("Pitch and Speed"),
```
saintmatthieu, line 159: I thought I did try that, but it led to some crash, can't remember the exact circumstances. Anyways, that seems to work now and I don't see by looking at the code why it shouldn't.

DavidBailes:
This fix has been included in https://github.com/audacity/audacity/pull/6174

PR #6174, opened 2024-03-25
6168 pitch and speed focus fix

### LINKED ISSUE #6168: Pitch and Speed dialog stays hidden when accessed via overflow button
### Bug description

Pitch and Speed dialog stays behind main window when accessed via overflow button if it has already been opened and is in the background.

### Steps to reproduce

1. Invoke the dialog
2. Click somewhere else - the dialog is now hidden
3. Invoke the dialog using the overflow button

### Expected behavior

The dialog is brought to the fore.

### Actual behavior

The dialog stays hidden.

### Audacity Version

### Operating system

macOS

### Additional context

_No response_

saintmatthieu:
Resolves: #6168, #6170

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/PitchAndSpeedDialog.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/PitchAndSpeedDialog.cpp
@@ -163,7 +163,7 @@ void PitchAndSpeedDialog::Destroy(AudacityProject& project)
 
 PitchAndSpeedDialog::PitchAndSpeedDialog(AudacityProject& project)
     : wxDialogWrapper(
-         nullptr, wxID_ANY, XO("Pitch and Speed"), wxDefaultPosition,
+         FindProjectFrame(&project), wxID_ANY, XO("Pitch and Speed"), wxDefaultPosition,
          { 480, 250 }, wxDEFAULT_DIALOG_STYLE)
     , mProject { project }
     , mProjectCloseSubscription { ProjectWindow::Get(mProject).Subscribe(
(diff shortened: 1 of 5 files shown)


saintmatthieu:
@DavidBailes this actually depends on your fix, so to reduce overhead I included it in this PR.

crsib (commented):
I don't quite understand what is goal of this PR, except for parenting.

I think what you a looking for is `wxDialog::Raise`, but probably I am wrong.

DavidBailes:
> I think what you a looking for is `wxDialog::Raise`, but probably I am wrong.

OnNextWindow() and OnPrevWindow() both use Raise(). There's also a mac/Linux only SetFocus().
If you haven't already tried using Raise() might be worth doing so to avoid the hide/show workaround.

saintmatthieu:
Yes, looks good! Coming up ...

src/tracks/playabletrack/wavetrack/ui/PitchAndSpeedDialog.cpp
```diff
@@ -269,13 +269,8 @@ PitchAndSpeedDialog& PitchAndSpeedDialog::SetFocus(
...
    if (item)
       item->SetFocus();
-#ifndef _WIN32
-   // On macOS, just calling `SetFocus` isn't enough, but on Windows, changing
-   // visibility to false and then back to true is ugly, so don't do this there.
-   // TODO: Linux ?
-   Show(false);
-#endif
-   Show(true);
+   wxDialog::Show(true);
```
crsib, line 272: But why do we still need both Hide/Show cycle and Raise? I expected that Raise would be sufficient alone

saintmatthieu: Hide/Show cycle? We are only calling `Show(true)`. If that's what you meant: maybe, but does it hurt?

crsib: Hmm, I misread the diff, sorry

crsib (approved):

PR #6176, opened 2024-03-25
6158 buzz when toggling formant preservation

### LINKED ISSUE #6158: When enabling/disabling formant preservation during playback, at the end of the clip a short buzzing sound is played
### Bug description

https://github.com/audacity/audacity/assets/70063970/e76879bb-6f97-4487-a13b-f3aaf217ed6a

### Steps to reproduce

1. Grab a clip
2. Play it
3. (optional) Set a looping region to be at the exact length of the clip
4. Bring up the Pitch and Speed dialog
5. Enter some non-zero value into Semitones
6. While the clip is playing, enable/disable "Optimize for Voice" a few times
7. Pay attention to the very end of the clip

### Expected behavior

No unwanted artifacts 

### Actual behavior

At the very end of the clip, a short but audible "buzzing" sound appears. It seems that the more times you enable/disable "Optimize for voice", the longer this sound is.

### Audacity Version

### Operating system

Windows, macOS

### Additional context

_No response_

saintmatthieu:
Resolves: #6158 

The buzz was due to the readout of an uninitialized buffer, which occurred when the stretcher has to reboot during playback. This easy fix is to zero it out instead, which makes the imperfection less audible. A proper fix would be to resynchronize the time stretcher's sample provider, but the effort is probably not worth it: allowing formant preservation toggling during playback is just a convenience for the user to hear with and without with just one click, but is by no means intended for actual music production.

Working on this ticket also uncovered a potential crash, due to a race condition between stretcher initialization on the audio thread and formant-preservation callbacks from the main thread. One could crash rather easily by selecting a very short loop of a few hundred ms, playback and toggling fast formant preservation on and off. A fix is proposed.

Recommended:

### DIFF
--- a/libraries/lib-stretching-sequence/ClipSegment.cpp
+++ b/libraries/lib-stretching-sequence/ClipSegment.cpp
@@ -65,6 +65,12 @@ ClipSegment::ClipSegment(
 {
 }
 
+ClipSegment::~ClipSegment()
+{
+   mOnSemitoneShiftChangeSubscription.Reset();
+   mOnFormantPreservationChangeSubscription.Reset();
+}
+
 size_t ClipSegment::GetFloats(float* const* buffers, size_t numSamples)
 {
    // Check if formant preservation of pitch shift needs to be updated.
--- a/libraries/lib-stretching-sequence/ClipSegment.cpp
+++ b/libraries/lib-stretching-sequence/ClipSegment.cpp
@@ -12,7 +12,6 @@
 #include "ClipInterface.h"
 #include "SampleFormat.h"
 #include "StaffPadTimeAndPitch.h"
-
 #include <cassert>
 #include <cmath>
 #include <functional>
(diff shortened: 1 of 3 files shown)


libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -65,6 +65,12 @@ ClipSegment::ClipSegment(
 {
 }
 
+ClipSegment::~ClipSegment()
+{
+   mOnSemitoneShiftChangeSubscription.Reset();
```
crsib, line 70: Why?

crsib (commented):
I don't understand second commit. I doesn't seem that it should have any visible behavior change. 

Observers are not thread safe indeed.

I do approve the last commit

libraries/lib-stretching-sequence/ClipSegment.cpp
saintmatthieu: To get the publisher to know a wee bit earlier that the subscription isn't valid anymore, hopefully before this object and its members gets destroyed ... but that reveals my lack of confidence that risking
```
[this](int cents) {
            mCentShift = cents;
            mUpdateCentShift = true;
         }
```
to be called after `this` is destroyed` (and hence `mCentShift` (`bool`) and `mUpdateCentShift` (`atomic<bool>`)) is safe. I couldn't crash Audacity anymore when stress-testing this on Windows and Mac, but I don't know the theoretical justification, if it does exist. Input welcome.

crsib: That depends on the data layout, but ok

libraries/lib-stretching-sequence/ClipSegment.cpp
```diff
@@ -44,28 +43,54 @@ ClipSegment::ClipSegment(
...
 {
-   std::lock_guard<std::mutex> lock(mStretcherMutex);
+   // Check if formant preservation of pitch shift needs to be updated.
+   // This approach is not immune to a race condition, but it is unlikely and
+   // not critical, as it would only affect one playback pass, during which the
+   // user could easily correct the mistake if needed. On the other hand, we
+   // cannot trust that the observer subscriptions do not get called after
+   // destruction of this object, so better not do anything too sophisticated
+   // there.
+   if (mUpdateFormantPreservation)
```
crsib, line 83: This can be done as `if(mUpdateFormantPreservation.exchange(false)`, but I don't care much

crsib (approved):

libraries/lib-stretching-sequence/ClipSegment.cpp
saintmatthieu: Thanks, I didn't know that.

Issue #6127, opened 2024-03-12
Cannot build release beta Windows


SteveALee:
### Bug description

I'm having errors building 3.5.0 beta x64 Release for Windows

I first found the problem in my [ASIO build script](https://gist.github.com/SteveALee/da24c2be633340b8791066dd98eb5d0b?permalink_comment_id=4430980) which builds x64 release only

Is it buildable? How is the release beta exe built?

### Steps to reproduce

1. get source with git label Audacity-3.5.0.beta (I assume this is correct even though audacity splash screen indicates a label with the date suffixed)
3. run cmake-gui,, config ,generate, open VS
4. Select Release and build solution

### Expected behavior

build with no errors

### Actual behavior

```
Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	LNK2001	unresolved external symbol "__declspec(dllimport) const MenuRegistry::MenuItem::`vftable'{for `MenuRegistry::ItemProperties'}" (__imp_??_7MenuItem@MenuRegistry@@6BItemProperties@1@@)	Audacity	C:\projects\build\src\TranscriptionToolBar.obj	1		
Error	LNK2001	unresolved external symbol "__declspec(dllimport) const MenuRegistry::MenuItem::`vftable'{for `MenuRegistry::ItemProperties'}" (__imp_??_7MenuItem@MenuRegistry@@6BItemProperties@1@@)	Audacity	C:\projects\build\src\SpectrumView.obj	1		
Error	LNK2001	unresolved external symbol "__declspec(dllimport) const MenuRegistry::MenuItem::`vftable'{for `MenuRegistry::ItemProperties'}" (__imp_??_7MenuItem@MenuRegistry@@6BItemProperties@1@@)	Audacity	C:\projects\build\src\Scrubbing.obj	1		
Error	LNK2001	unresolved external symbol "__declspec(dllimport) const MenuRegistry::MenuItem::`vftable'{for `MenuRegistry::ItemProperties'}" (__imp_??_7MenuItem@MenuRegistry@@6BItemProperties@1@@)	Audacity	C:\projects\build\src\ViewMenus.obj	1		
Error	LNK2001	unresolved external symbol "__declspec(dllimport) const MenuRegistry::MenuItem::`vftable'{for `MenuRegistry::ItemProperties'}" (__imp_??_7MenuItem@MenuRegistry@@6BItemProperties@1@@)	Audacity	C:\projects\build\src\DeviceToolBar.obj	1		
[...]

PR #6179, opened 2024-03-27
Fixes the Release builds on MSVC

crsib:
It seems there is a bug in the compiler or linker, that optimizes out ItemProperties destructor from the binary, along with the VTable

Resolves: #6127 

Recommended:


saintmatthieu (approved):
❤️

Issue #5782, opened 2023-12-13
export cannot recursively create new folders


LWinterberg:
> I’ve noticed that since the 3.4 update (I believe), the export audio dialogue box now doesn’t allow me to create a new folder in the “Folder” path if the path seems to be more than 1 folder deeper than the current folder.

> For example, if the path is set to C:\export
> I can create C:\export\export
> But I cannot create C:\export\export\a

> In this case, Audacity has the error of “Unable to create destination folder”

Original report from https://forum.audacityteam.org/t/bug-or-error-when-trying-to-make-new-folder-when-exporting-audio/92959/ 

The workaround here is to use the browse dialog instead to create the folders.

dozzzzer:
Tested

PR #6185, opened 2024-03-28
exporting audio now creates directories recursively

andIsome:
Resolves: https://github.com/audacity/audacity/issues/5782

exporting audio now creates directories recursively

Recommended:


LWinterberg (approved):
excellent find!

Issue #6186, opened 2024-03-29
Subtitle handling improvements


LWinterberg:
With #6139 now merged (thanks @Pokechu22 @lbordowitz!), there's some additional polishing the feature could use: 

- [ ] Support import of WebVTT files. This requires cleaning up the imported file of all the additional stuff vtt files can have (styling, comments, etc) before converting them to labels
- [ ] Add an "all supported" option for Import Labels. Currently, you get to choose between txt (default) and srt, but there is no "all supported files". 
- [ ] Add a position field to the Label Editor (currently, there's a Low Frequency and High Frequency field which can optionally be exported - adding a positioning field here for subtitles doesn't seem unwarranted)
- [ ] Add a new exporter with relevant options for vtt, eg a textbox to input styling (probably together with https://github.com/audacity/audacity/issues/4891 ) 

----

Also worth mentioning:

<details><summary>Known issues from #394 </summary>

There are a few known issues:
* WebVTT files cannot be imported.  The format is much more complicated ([see the spec](https://w3c.github.io/webvtt/)) and supports content that does not directly translate into labels (such as comments which are not displayed and do not have a timestamp, and regions that indicate positioning of cues).
* Both SubRip and WebVTT support multiple lines in a cue, but labels do not currently support multiple lines (and the current txt format doesn't have the ability to do it).  When importing SubRip files, I simply join the lines with spaces. (#170 used `|` based on [MicroDVD](https://en.wikipedia.org/wiki/MicroDVD), which I have no experience with.)
* Exporting multiple tracks at the same time leads to an invalid file &mdash; it effectively concatenates what happens if you export each individually.  The identifier (which is just a counter here) is reset to `1` for each file, and the `WEBVTT` header is repeated.  Although both of those are fixable, WebVTT also requires that the start time "must be greater than or equal to the start time offsets of all previous cues in the file" ([§4.1, "WebVTT cue timings"](https://w3c.github.io/webvtt/#webvtt-cue-timings)).  Resolving that requires merging the two files in sorted order and that would require a more substantial rework of how label track I/O works (which would be good eventually, as the labels editor would also be sorted by timestamp instead of track and then timestamp).

</details>

* https://github.com/audacity/audacity/pull/819 from @JohnColket with a similar new import format idea

leodom:
Hello, I am using version 3.4.2, how can I use the export label plugin ?

LWinterberg:
[...]

PR #6201, opened 2024-04-02
Commit experimental branches

Paul-Licameli:

Cleanup conditional compilation directives for all the EXPERIMENTAL flags
that have been turned on in releases.

Recommended:


src/AdornedRulerPanel.cpp
```diff
@@ -1065,7 +1065,6 @@ std::vector<UIHandlePtr> AdornedRulerPanel::QPCell::HitTest(
    std::vector<UIHandlePtr> results;
    auto xx = state.state.m_x;
 
-#ifdef EXPERIMENTAL_DRAGGABLE_PLAY_HEAD
```
LWinterberg, line 1068: I'm not sure this function does anything at the moment, it seems to me like instead of dragging the playhead around, I instead do things with loop regions.

src/Experimental.cmake
```diff
@@ -106,9 +106,6 @@ set( EXPERIMENTAL_OPTIONS_LIST
    # PRL disabled 31 Aug 2021 because, at least on Mac, it misfires the
    # preference dialog whenever any keystroke shortcuts with Shift+ are used
    # EASY_CHANGE_KEY_BINDINGS
-
-   # Poke 13 Nov 2019
-   SUBRIP_LABEL_FORMATS
```
LWinterberg, line 172: This one actually only got merged this friday; somehow this feature had fallen through the cracks all these years. It is somewhat unfinished (see https://github.com/audacity/audacity/issues/6186) still and possibly the only flag I've seen so far which actually warrants an EXPERIMENTAL flag - but then, I did decide that in the current form it is useful and have no intention on disabling it if anything goes wrong. Removing this flag is fine.

LWinterberg (approved):
found some minor things, but overall it looks good to me!

src/effects/NoiseReduction.cpp
```diff
@@ -74,7 +74,7 @@
 
 // SPECTRAL_SELECTION not to affect this effect for now, as there might be no indication that it does.
 // [Discussed and agreed for v2.1 by Steve, Paul, Bill].
-#undef EXPERIMENTAL_SPECTRAL_EDITING
+#undef SPECTRAL_EDIT_NOISE_REDUCTION
```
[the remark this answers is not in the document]
Paul-Licameli: This “obvious” idea from 2014 was vetoed by the old team and I forgot it.  But it remains latent.

You don’t yet see part 2 of this contribution.  I also want to rewrite the un-enabled experimentals with C++17 `if constexpr` instead of old `#ifdef` and that would include this idea.  That rewrite would let the compiler check those branches too so that they remain buildable while other parts of Audacity evolve.

I’m waiting first for acceptance of the easier rewrites in this PR.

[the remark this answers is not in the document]
Paul-Licameli: And so the file Experimental.cmake will be eliminated, and the switching-on of other experiments will also not require recompiling everything as now.

src/prefs/MidiIOPrefs.h
```diff
@@ -13,8 +13,6 @@ class wxChoice;
 class wxTextCtrl;
 class ShuttleGui;
 
-#ifdef EXPERIMENTAL_MIDI_OUT
-
 #ifndef __AUDACITY_MIDI_IO_PREFS__
 #define __AUDACITY_MIDI_IO_PREFS__
```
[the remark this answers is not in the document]
Paul-Licameli: That’s an old fashioned C/C++ “include guard” that allows a header file to be included twice but fully parsed only once.  It’s still my habit to write these.  `#pragma once` is an alternative, not quite standard, but recognized by the compilers that matter for our builds.

src/AdornedRulerPanel.cpp
Paul-Licameli: Can’t you click precisely on the down pointing green (or red) triangle and move it, when the head is in pinned state?

Paul-Licameli: I implemented the pinned head, which at first was always centered, then in a later release let you drag it off center, thinking that it’s more useful during recording to have it farther right.

Also little known perhaps — you can double click it to recenter.

Paul-Licameli:
Thank you for review.  I may open the other PR then, which is is not as simple.

src/AdornedRulerPanel.cpp
LWinterberg: oh, in pinned state. Yes, that still works.

Paul-Licameli: Good

Issue #6032, opened 2024-02-23
When shrunk to almost absolute minimum, audio clip has no longer stretch ratio indication in its title bar


dozzzzer:
### Steps to reproduce

1. Have an audio clip
2. Ctrl (Cmd) + F
3. While holding Alt (Option) key, drag the right edge of the clip all the way to T=0
4. Ctrl (Cmd) + F
5. Observe: no speed indicator in the title
6. Try to expand the clip while holding Alt (Option)
7. Observe

### Expected behavior

The clip's speed indicator remains visible at all times.

### Actual behavior

The speed indicator won't re-appear even after the clip has been expanded.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

LWinterberg:
The issue appears to be that the indicator can't display numbers higher than 99999% and instead disappears after.

LWinterberg:
Hello! We just [switched development to Audacity 4](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444). We've redone, or plan on redoing, various things from scratch, including the entire UI, and the part this bug is about. As such, I'm closing this bug as it will be fixed there and we won't be fixing it for the final releases of Audacity 3.

PR #6202, opened 2024-04-02
Speed indicator now remains visible

pedrofpclima:
Resolves: #6032 

When shrunk to near absolute minimum, the audio clip's speed indicator would disappear. This is not the case anymore, as the lowest stretch ratio is now set to 100 / (double)99999. Such change ensures that the clip's speed never exceeds 99999%.

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -361,31 +361,43 @@ double WaveClip::GetStretchRatio() const
          1.0;
    double stretchRatio = mClipStretchRatio * dstSrcRatio;
 
-   auto projectTempo = mProjectTempo;
-   auto rawAudioTempo = mRawAudioTempo;
+   double lowestStretchRatio = 100 / (double)99999;
+
+   if (stretchRatio < lowestStretchRatio) {
+      stretchRatio = lowestStretchRatio;
+   }
+
+   return stretchRatio;
+}
+
+double WaveClip::GetStretchRatio()
+{
+   const auto dstSrcRatio =
+      mProjectTempo.has_value() && mRawAudioTempo.has_value() ?
+      *mRawAudioTempo / *mProjectTempo :
+      1.0;
+   double stretchRatio = mClipStretchRatio * dstSrcRatio;
 
    double lowestStretchRatio = 100 / (double)99999;
 
    if (stretchRatio < lowestStretchRatio) {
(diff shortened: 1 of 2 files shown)


LWinterberg:
I have two comments on this: 

Firstly, this creates a situation where a clip never can be made shorter than that. My gut feeling says that's fine - we're running into the issue that clips which are too short really struggle during playback anyway. I'd still hear @saintmatthieu's thought on the matter though. 

Secondly, this fix causes a bug @pedrofpclima 

1. Using Alt+drag on the edge, drag the clip to 99999%
2. Still holding alt+drag, continue dragging the clip shorter
3. Let go of the clip
4. Using alt+drag on the edge again, try to expand the clip/slow it down

In this case, dragging is unresponsive by as many pixels as you overshot in step 2. I suspect that you need to clamp another value somewhere else as well.

pedrofpclima:
> I have two comments on this:
> 
> Firstly, this creates a situation where a clip never can be made shorter than that. My gut feeling says that's fine - we're running into the issue that clips which are too short really struggle during playback anyway. I'd still hear @saintmatthieu's thought on the matter though.
> 
> Secondly, this fix causes a bug @pedrofpclima
> 
> 1. Using Alt+drag on the edge, drag the clip to 99999%
> 2. Still holding alt+drag, continue dragging the clip shorter
> 3. Let go of the clip
> 4. Using alt+drag on the edge again, try to expand the clip/slow it down
> 
> In this case, dragging is unresponsive by as many pixels as you overshot in step 2. I suspect that you need to clamp another value somewhere else as well.

Thank you for your feedback.

Regarding your first comment, you're right. Setting a minimum limit may restrict the clip from being made shorter. And yet I feel like this is the natural thing to do, since short clips tend to struggle during playback.

As for the second issue you mentioned, I apologize for the oversight. I've identified the bug you described and have already fixed it. It was caused by not adjusting the tempos accordingly.

I'm sorry for the delay in addressing this, I've been quite busy with university work recently. I'll make sure to be more prompt in the future.

Please let me know if there are any other concerns.

LWinterberg:
Hello! We just [switched master to 4.0.0](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444) which makes this PR incompatible and/or obsolete as we're switching UI framework and reorganizing large parts of the codebase. This bug should be fixed in Audacity 4.

Issue #5966, opened 2024-02-16
Reset Config moves Share Audio toolbar to the next row


dozzzzer:
### Steps to reproduce

1. Launch Audacity
2. Tools > Reset config
3. Observe

### Expected behavior

Share Audio toolbar remains next to the Audio Setup toolbar

### Actual behavior

Share Audio toolbar has been moved to the next row:

<img width="450" alt="image" src="https://github.com/audacity/audacity/assets/70063970/dac759b6-4358-421d-ae32-d3059cbb823a">

### Audacity Version

[other](https://nightly.link/crsib/audacity/workflows/build/au_project_sync)

### Operating system

Windows, macOS

petersampsonaudacity:
Testing on W10 with latest 3.5.0 alpha test build I cannot replicate this

crsib:
Bugs with Cloud label are not present in master or any opened PR so far. And fingers crossed - the all will be fixed before that :-)

petersampsonaudacity:
> Bugs with Cloud label

But there is no Cloud label for this bug ...

dozzzzer:
Now there is!

petersampsonaudacity:
Testing on w10 with the latest Beta test build:  audacity-win-3.5.0-beta-20240301+ed81a44-x64-msvc2022

This looks to be fixed for the 3.5.0 Beta

chinakov:
This is done, thank you!

PR #6212, opened 2024-04-03
Audacity does not depend on lib-cloud-audiocom

Paul-Licameli:

Recommended:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1355,12 +1355,6 @@ if( ${_OPT}has_url_schemes_support )
    )
 endif()
 
-if( ${_OPT}has_audiocom_upload )
-   list( APPEND AUDACITY_LIBRARIES
-      lib-cloud-audiocom-interface
-   )
-endif()
-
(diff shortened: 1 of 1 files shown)


crsib:
ATM Audacity must depend on lib-cloud-audiocom. It should probably be reworked in a different way, but not right now. 

You can see the usage of HAS_AUDIOCOM_UPLOAD in ToolManager.cpp

Paul-Licameli:
`DefaultConfigTable` wants to become a scattered registry but that work isn't done

crsib:
`HAS_AUDIOCOM_UPLOAD` changes the visible behavior.

Paul-Licameli:
> `HAS_AUDIOCOM_UPLOAD` changes the visible behavior.

That does not prove that this change is incorrect.  But I'll drop it.

crsib:
This change results in a bug: https://github.com/audacity/audacity/issues/5966

Paul-Licameli:
> This change results in a bug: #5966

Ah, now I see it.

Still this could be fixed with some other lines of Cmake language.

Issue #6188, opened 2024-03-29
File menu: all items should have access keys


DavidBailes:
### Steps to reproduce

Press Alt + F to open the File menu.

### Expected behavior

All the items on the menu have access keys

### Actual behavior

These items are missing access keys:
Open from cloud
Save to cloud
Update Cloud audio preview
Export other
Share audio

### Audacity Version

latest beta

### Operating system

Windows

PR #6227, opened 2024-04-05
File menu: fix missing access keys

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6188

Fixed 5 missing access keys.

Recommended:


LWinterberg (approved):

DavidBailes:
@LWinterberg Was just wondering whether this change would cause any problems with the translations.

DavidBailes:
> Was just wondering whether this change would cause any problems with the translations.

@LWinterberg @crsib It does cause problems. Is there a work around for this?

Issue #6253, opened 2024-04-10
Import raw offset broken above 2147483648 bytes


namazso:
### Bug description

When importing raw data, attempting to set a byte offset over 2147483648 bytes will import with no offset applied

### Steps to reproduce

1. File > Import > Raw Data...
2. Select a big enough file
3. Enter a "Start offset:" of at least 2147483648

### Expected behavior

Import happens from the 2147483648th byte

### Actual behavior

Import happens from the beginning

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

The window resizing is also broken, so bigger offsets cannot be seen:

![image](https://github.com/audacity/audacity/assets/8676443/ef03fcf4-f77c-412f-9f71-541ccd833329)

chinakov:
Tested, thanks

PR #6254, opened 2024-04-10
support larger offsets for import raw

LWinterberg:
Resolves: #6253 probably. Can't test because I need to reconfigure my local build environment.

### DIFF
--- a/src/import/ImportRaw.cpp
+++ b/src/import/ImportRaw.cpp
@@ -71,7 +71,7 @@ class ImportRawDialog final : public wxDialogWrapper {
    // Make static to preserve value for next raw import
    static int mEncoding;
    static unsigned mChannels;
-   static int mOffset;
+   static long long mOffset;
    static double mRate;
    static double mPercent;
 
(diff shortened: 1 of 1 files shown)


LWinterberg:
oh wow, apparently I can't use `int64_t` because `long` is defined differently on Linux compared to macOS and Windows, and wxString::toLong gets confused on one platform or the other. I guess this means then that you'll be able to put in a larger number on Linux compared to the other platforms then in case you need to have an offset greater than 8 petabytes  ¯\\\_(ツ)_/¯

src/import/ImportRaw.cpp
```diff
@@ -95,7 +95,7 @@ class ImportRawDialog final : public wxDialogWrapper {
 // Initial value for Import Raw dialog
 int ImportRawDialog::mEncoding = SF_FORMAT_RAW | SF_ENDIAN_CPU | SF_FORMAT_PCM_16;
 unsigned ImportRawDialog::mChannels = 1;
-int ImportRawDialog::mOffset = 0;
+long ImportRawDialog::mOffset = 0;
```
namazso, line 98: `long` on Windows is 32 bits, so this wouldn't fix the issue

namazso:
You probably want to just use `long long`, which is 64 bits on both platforms, and there exists a wxWidgets function for it.

LWinterberg:
thanks!

namazso (approved):
I tested the CI artifact for Windows and it works as expected

grliszas14 (approved):

namazso:
image
Seems like the previous values are filled back incorrectly

Issue #6265, opened 2024-04-12
Make track header wider


LWinterberg:
Currently, the track header (TCP) is uncomfortably narrow, to the point where View > Show track names as overlay needed to be made a thing because the thing isn't wide enough to show the entire track name. 

A wider track header decreases the need for that as more info can be shown at once.

petersampsonaudacity:
I tested this on W10 with @LWinterberg 's branch for this.

The TCP is indeed a little wider.

----------------------------------------------------------------------------------------------

What would be really nice would be if the user could expand/contract the TCP by clicking&dragging on its right border.

_I seem to recall user request for this in the past, in the old Forum._

PR #6266, opened 2024-04-12
make track header wider

LWinterberg:
Resolves:  #6265

Still to investigate: The font size got larger in the process somehow. Not sure why. Not sure I dislike it either.

### DIFF
--- a/libraries/lib-screen-geometry/ZoomInfo.h
+++ b/libraries/lib-screen-geometry/ZoomInfo.h
@@ -27,7 +27,7 @@ enum : int {
    kLeftMargin = kLeftInset + kBorderThickness,
    kRightMargin = kRightInset + kShadowThickness + kBorderThickness,
 
-   kTrackInfoWidth = 100 - kLeftMargin,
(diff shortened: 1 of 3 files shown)


grliszas14:
Take a look here:
https://github.com/audacity/audacity/blob/master/src/TrackInfo.cpp#L53-L63
kTrackInfoWidth is bigger -> so allowableWidth is bigger  -> so the loop that reduces the font size is run fewer times. So that the font size remains smaller while keeping larger kTrackInfoWidth you'd need to make allowableWidth independent of the kTrackInfoWidth. Up to you though

grliszas14 (approved):

LWinterberg:
OK, I understand this function now. It looks for "Stereo, 999999 Hz" as that probably is the longest text string and if that doesn't fit, it makes the font smaller. That unfortunately, it kinda works around the wrong issue. In 3.5:

image
image

In this build:

image
image

I think the text downscaling is both insufficient for what it was trying to do, and also probably unnecessary now. I think I'll remove it.

Issue #6276, opened 2024-04-13
Share Audio toolbar is in wrong position when resetting configuration


Paul-Licameli:
### Steps to reproduce

View > Toolbars > Reset

### Expected behavior

Share Audio is in the proper place as in release 3.5

### Actual behavior

Share Audio toolbar is alone in a new row of toolbars

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

### Additional context

The error was in b730d50b6d0aeaa50941f433c951e296592d34fa

dozzzzer:
The bug is no longer reproducible

dozzzzer:
Disregard my previous comment. The issue does exist in master which #6277 fixes.

PR #6277, opened 2024-04-13
Fix position of Share Audio toolbar when resetting

Paul-Licameli:
Resolves: #6276

Recommended:


crsib (approved):

LWinterberg:
Just to verify: This is for master/3.6 and not 3.5, correct?

crsib:
This is for 3.6

LWinterberg (approved):

PR #6280, opened 2024-04-15
ctrl+v can import files from clipboard

### LINKED ISSUE #6028: Ctrl+V with audio file on clipboard should import it
If I copy an audio file on my computer's clipboard, ctrl+V on Audacity should import it.

saintmatthieu:
Resolves: #6028

Recommended:

QA:
Basically, copy-pasting from the clipboard behaves the exact same as drag-dropping. In particular,
- [x] files are sorted alphabetically
- [x] there is one history entry per imported file, even if several were copied and pasted in one go (@LWinterberg shouldn't this be changed?)
- [x] Not effective during playback
- [x] A mixture of file types can be imported, e.g. audio and midi.
- [x] The bottom-most track gets focus and is fully selected.

Besides:
- [x] Copy a track, then copy a wav file, paste: the wav file is pasted. Vice versa. (It is important that this is tested on all Windows, Mac and Linux, as implementation interacts with the OS clipboard).
- [x] If import of one of the files fails, then entire operation is aborted.

### DIFF
--- a/src/DropTarget.cpp
+++ b/src/DropTarget.cpp
@@ -17,7 +17,6 @@
 #include "Project.h"
 #include "ProjectFileManager.h"
 #include "TrackPanel.h"
-#include "Viewport.h"
 
 #if wxUSE_DRAG_AND_DROP
 class FileObject final : public wxFileDataObject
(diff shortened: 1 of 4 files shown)


LWinterberg:
> there is one history entry per imported file, even if several were copied and pasted in one go (@LWinterberg shouldn't this be changed?

Yes, consolidating these actions makes sense to me. However, if the files get imported one by one, those actions should not be consolidated.

crsib (approved):

saintmatthieu:
> > there is one history entry per imported file, even if several were copied and pasted in one go (@LWinterberg shouldn't this be changed?
> 
> Yes, consolidating these actions makes sense to me. However, if the files get imported one by one, those actions should not be consolidated.

Leaving it as is since it's beyond the scope of this PR and it turns out to be non trivial.

LWinterberg (approved):

Issue #6269, opened 2024-04-12
Remove track name overlay


LWinterberg:
The track name overlay is redundant with the wider track headers and can be removed. It also is buggy. 

Original bug: 

![image](https://github.com/audacity/audacity/assets/87814144/29b2dd38-d16a-4a8f-a5a7-353e42ac8e48)

With View -> Show Track Name as overlay

Fwiw: I wonder if this option even is still necessary if we make the header a bit wider - https://github.com/audacity/audacity/issues/6265 - and considering that the clip names use the track name by default anyway.

LWinterberg:
> Fwiw: I wonder if this option even is still necessary if we make the header a bit wider...

After further consideration and [a quick poll](https://www.youtube.com/channel/UCHUCowIfMiQd_IeLDaftV_g/community?lb=UgkxpeADvJD5tVt3vCIK6YA4euC8d5I3dFqc), I think the answer is a pretty clean "no". The expectation pretty strongly is that the name should be in the header, with a good chunk having a mental model of continuous track (ie not even considering that clips are their own entities), for which "above the waveform" (so the clip title) also makes some sense. 

For people who are familiar with clips and events, the placement of the bubble would indicate that it'd be a clip property, not a track one. 

I now think the wider header solves a superset of problems the overlay was trying to solve, while causing considerably less confusion and not needing an option.

PR #6287, opened 2024-04-16
Remove track name overlay option

LWinterberg:
Resolves: #6269

### DIFF
--- a/src/TrackArtist.cpp
+++ b/src/TrackArtist.cpp
@@ -129,8 +129,6 @@ void TrackArtist::UpdateSelectedPrefs( int id )
 {
    if( id == ShowClippingPrefsID())
       mShowClipping = gPrefs->Read(wxT("/GUI/ShowClipping"), mShowClipping);
-   if( id == ShowTrackNameInWaveformPrefsID())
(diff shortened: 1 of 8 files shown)


saintmatthieu (approved):

QuantumGood:
1. This is a major truncation of Audacity functionality and usefulness for those that use Audacity's feature that allows track name to be exported as filename—a typical filename with project name, date and user name (e.g. "IRS.Efile_1-5-2024_Kevin.Elin") was formerly readable as a track name.
2. Truncating from dozens of characters down to ten (numeric characters) is nowhere near enough for using track names for version and project management. Around 99% of my track names are now not viewable. I have dozens of projects with track names that follow client filename standards, all now unreadable without manually checking eack. Also, knowing overlay was available, we adopted other standards for version and project tracking that use track names longer than ten, also zero of which are now viewable.  
3. I understand you can do what you want without considering user feedback, but removal without checking seems counterproductive. I am returning to working in Audacity after a few months off, or would have commented sooner.
image
4. I understand that inexperienced users do nearly zero version and project management, but experienced users have to take version and project management into account on most projects.

PR #6306, opened 2024-04-18
Realtime limiter

### LINKED ISSUE #6305: Real-time-capable limiter
### Your idea

A limiter is an essential component in a mastering chain to reduce severeness of clipping.
Add a limiter to Audacity's built-in real-time effects. If appropriate, design it so that it can also be used as compressor.
Stretch goal: make it configurable for extra smoothness and classical music, [in a Chris's compressor style](https://manual.audacityteam.org/man/chris_s_dynamic_compressor.html).

### Problem to be solved

Also, Audacity lacks real-time limiter and compressor, although these are essential to mastering.
Stretch goal would also resolve #2397.

### Prior art

_No response_

### Additional context

_No response_

saintmatthieu:
Resolves: #6305

This PR wraps [Daniel Rudrich's SimpleCompressor](https://github.com/DanielRudrich/SimpleCompressor) in two new real-time capable built-in effects, namely "Real-time Compressor" and "Real-time Limiter".

The limiter is the same as the compressor but with a reduced set of parameters, ratio and attack time being hard-coded to infinity and 0ms.

The legacy, offline compressor files were renamed, but for now stays found in the same place in the menu (`Volume and Compression > Compressor` and `Volume and Compression > Limiter`). Removal from default list is planned in #6319. At that stage the new versions should also find their final place in the menu.

UI currently only consists of sliders. Real-time visual indicators are essential for this type of effect, but their introduction can be postponed to a follow-up PR.

Recommended:

QA:
- [ ] Changing parameters in real time works
- [ ] Latency compensation works (use look-ahead for this to be non-trivial)
- [ ] Effect preview works
Channel routing tests:
- [ ] Mono track is processed properly
- [ ] Stereo track is processed properly (use a clip with contrasting left/right content)

### DIFF
--- a/libraries/lib-effects/CMakeLists.txt
+++ b/libraries/lib-effects/CMakeLists.txt
@@ -19,6 +19,13 @@ set( SOURCES
    PerTrackEffect.h
    StatefulEffectBase.cpp
    StatefulEffectBase.h
+
+   processors/CompressorProcessor.cpp
+   processors/CompressorProcessor.h
+   processors/SimpleCompressor/GainReductionComputer.cpp
+   processors/SimpleCompressor/GainReductionComputer.h
+   processors/SimpleCompressor/LookAheadGainReduction.cpp
+   processors/SimpleCompressor/LookAheadGainReduction.h
 )
 set( LIBRARIES
    lib-command-parameters-interface
--- a/libraries/lib-effects/tests/CompressorProcessorTests.cpp
+++ b/libraries/lib-effects/tests/CompressorProcessorTests.cpp
@@ -1,0 +1,47 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  CompressorProcessorTests.cpp
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#pragma once
+
+#include "processors/CompressorProcessor.h"
+#include <catch2/catch.hpp>
+
+TEST_CASE("CompressorProcessor", "smoke test")
+{
+   CompressorSettings settings;
+   settings.lookaheadMs = 5;
+   CompressorProcessor sut { settings };
+   constexpr auto sampleRate = 44100;
+   constexpr auto numChannels = 2;
+   constexpr auto blockSize = 44100;
+   constexpr auto signalSize = 2 * blockSize;
+   sut.Init(sampleRate, numChannels, blockSize);
+   std::vector<std::vector<float>> buffer(numChannels);
+   std::vector<float*> pointers(numChannels);
+   for (auto i = 0; i < numChannels; ++i)
+   {
+      auto& in = buffer[i];
+      in.resize(signalSize);
+      std::fill(in.begin(), in.begin() + signalSize / 2, 0.0f);
+      std::fill(in.begin() + signalSize / 2, in.end(), 1.0f);
+   }
+   auto progress = 0;
+   while (progress < signalSize)
+   {
+      const auto remaining = signalSize - progress;
+      const auto toProcess = std::min(remaining, blockSize);
+      std::transform(
+         buffer.begin(), buffer.end(), pointers.begin(),
+         [progress](std::vector<float>& v) { return v.data() + progress; });
+      const auto p = pointers.data();
+      sut.Process(p, p, toProcess);
+      progress += toProcess;
+   }
+}
--- a/src/effects/Compressor.cpp
+++ b/src/effects/Compressor.cpp
@@ -1,0 +1,96 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  Compressor.cpp
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+
+#include "Compressor.h"
+#include "CompressorEditor.h"
+#include "CompressorInstance.h"
+#include "LoadEffects.h"
+#include "ShuttleAutomation.h"
+
+const ComponentInterfaceSymbol EffectCompressor::Symbol { XO(
+   "Real-time Compressor") };
+
+namespace
+{
+BuiltinEffectsModule::Registration<EffectCompressor> reg;
+} // namespace
+
+const EffectParameterMethods& EffectCompressor::Parameters() const
+{
+   static CapturedParameters<
+      EffectCompressor, CompressorEditor::thresholdDb, CompressorEditor::kneeDb,
+      CompressorEditor::lookaheadMs, CompressorEditor::attackMs,
+      CompressorEditor::releaseMs, CompressorEditor::ratio,
+      CompressorEditor::makeUpDb>
+      parameters;
+   return parameters;
+}
+
+std::shared_ptr<EffectInstance> EffectCompressor::MakeInstance() const
+{
+   return std::make_shared<CompressorInstance>(*this);
+}
+
+bool EffectCompressor::CheckWhetherSkipEffect(
+   const EffectSettings& settings) const
+{
+   // TODO
+   return Effect::CheckWhetherSkipEffect(settings);
+}
+
+EffectCompressor::EffectCompressor()
+{
+   SetLinearEffectFlag(true);
+}
+
+EffectCompressor::~EffectCompressor()
+{
+}
+
+ComponentInterfaceSymbol EffectCompressor::GetSymbol() const
+{
+   return Symbol;
+}
+
+TranslatableString EffectCompressor::GetDescription() const
+{
+   return XO("A compressor");
+}
+
+ManualPageID EffectCompressor::ManualPage() const
+{
+   // TODO
+   return L"";
+}
+
+EffectType EffectCompressor::GetType() const
+{
+   return EffectTypeProcess;
+}
+
+auto EffectCompressor::RealtimeSupport() const -> RealtimeSince
+{
+   return RealtimeSince::After_3_1;
+}
+
+// Effect implementation
+
+std::unique_ptr<EffectEditor> EffectCompressor::MakeEditor(
+   ShuttleGui& S, EffectInstance&, EffectSettingsAccess& access,
+   const EffectOutputs*) const
+{
+   auto& settings = access.Get();
+   auto& myEffSettings = GetSettings(settings);
+   auto result =
+      std::make_unique<CompressorEditor>(*this, access, myEffSettings);
+   result->PopulateOrExchange(S);
+   return result;
+}
--- a/src/effects/Compressor.cpp
+++ b/src/effects/Compressor.cpp
@@ -78,7 +78,7 @@ EffectType EffectCompressor::GetType() const
 
 auto EffectCompressor::RealtimeSupport() const -> RealtimeSince
 {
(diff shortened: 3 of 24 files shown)


src/effects/Compressor.cpp
```diff
@@ -1,737 +1,96 @@
...
 {
-   Parameters().Reset(*this);
-
-   mThreshold = 0.25;
-   mNoiseFloor = 0.01;
-   mCompression = 0.5;
-   mFollowLen = 0;
-
-   SetLinearEffectFlag(false);
+   SetLinearEffectFlag(true);
```
saintmatthieu, line 51: Oops, mistake ...

src/effects/Compressor.cpp
```diff
@@ -1,737 +1,96 @@
...
-
 ComponentInterfaceSymbol EffectCompressor::GetSymbol() const
 {
    return Symbol;
 }
 
 TranslatableString EffectCompressor::GetDescription() const
 {
-   return XO("Compresses the dynamic range of audio");
+   return XO("A compressor");
```
saintmatthieu, line 65: All right, I overlooked these implementations ... WIP.

src/effects/Compressor.cpp
```diff
@@ -1,737 +1,96 @@
...
-      // With RMS-based compression don't change values below mThreshold - 'downward' compression
-      out = value * pow(mThreshold/env, mCompression);
-   }
-
-   // Retain the maximum value for use in the normalization pass
-   if(mMax < fabs(out))
-      mMax = fabs(out);
-
-   return out;
+   return RealtimeSince::After_3_1;
```
saintmatthieu, line 81: I'm not sure here. Eventually it will replace the offline processor, "stealing" its name (currently "Real-time Compressor"), so maybe it should stay `After_3_1` and not be `Always` ? On the other hand, I don't see any different consequence in returning `After_3_1` or `Always`.

crsib: If you reuse the old ID, you should introduce After_3_6 (3_7?)

This is a way Paul has introduced to distinguish between RT/non-RT versions of built in effects

src/effects/DynamicRangeProcessorEditor.h
```diff
@@ -0,0 +1,155 @@
...
+public:
+   DynamicRangeProcessorEditor(
+      const EffectUIServices& services, EffectSettingsAccess& access,
+      const SettingType& settings)
+       : EffectEditor { services, access }
+       , mSettings { settings }
+   {
+   }
+
+   virtual ~DynamicRangeProcessorEditor() = default;
```
saintmatthieu, line 26: Not needed.

src/effects/Limiter.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+bool EffectLimiter::CheckWhetherSkipEffect(const EffectSettings& settings) const
+{
+   // TODO
+   return Effect::CheckWhetherSkipEffect(settings);
+}
+
+EffectLimiter::EffectLimiter()
+{
+   SetLinearEffectFlag(true);
```
saintmatthieu, line 49: Also wrong - coming up.

src/effects/LimiterEditor.h
```diff
@@ -0,0 +1,58 @@
...
+       : DynamicRangeProcessorEditor<LimiterEditor, LimiterSettings> {
+          services, access, settings
+       }
+   {
+   }
+   virtual ~LimiterEditor() = default;
+
+   void PopulateOrExchange(ShuttleGui& S);
+
+   static constexpr Parameter thresholdDb { &LimiterSettings::thresholdDb,
```
saintmatthieu, line 24: These defaults should be reviewed.

libraries/lib-viewport/Viewport.cpp
```diff
@@ -422,7 +422,8 @@ void Viewport::DoScroll()
 
    auto width = viewInfo.GetTracksUsableWidth();
    const auto zoom = viewInfo.GetZoom();
-   viewInfo.hpos = std::clamp(sbarH / zoom, lowerBound, total - width / zoom);
+   viewInfo.hpos =
```
crsib, line 425: I think I fixed it on my branch too, hence the conflict

libraries/lib-effects/tests/CompressorProcessorTests.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#pragma once
+
+#include "processors/CompressorProcessor.h"
+#include <catch2/catch.hpp>
+
+TEST_CASE("CompressorProcessor", "smoke test")
```
crsib, line 16: It's not quite clear what it tests TBH

src/effects/CompressorInstance.cpp
```diff
@@ -0,0 +1,134 @@
...
+      // so `Reinit()` should not reallocate memory.
+      slave.mCompressor->Reinit();
+   return true;
+}
+
+bool CompressorInstance::RealtimeAddProcessor(
+   EffectSettings& settings, EffectOutputs*, unsigned numChannels,
+   float sampleRate)
+{
+   mSlaves.emplace_back(std::move(mProcessor));
```
crsib, line 66: This looks ... odd. I think there should be something in line with `mSlaves.push_back(CompressorInstance { mProcessor })`

crsib: See `EffectReverb::Instance::RealtimeAddProcessor`

libraries/lib-effects/CMakeLists.txt
```diff
@@ -19,6 +19,13 @@ set( SOURCES
    PerTrackEffect.h
    StatefulEffectBase.cpp
    StatefulEffectBase.h
+
+   processors/CompressorProcessor.cpp
+   processors/CompressorProcessor.h
+   processors/SimpleCompressor/GainReductionComputer.cpp
+   processors/SimpleCompressor/GainReductionComputer.h
+   processors/SimpleCompressor/LookAheadGainReduction.cpp
+   processors/SimpleCompressor/LookAheadGainReduction.h
```
Paul-Licameli, line 28: Why do files relating only to one particular effect belong in lib-effects?  I don't agree with the file placement.

src/effects/CompressorInstance.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+   const LimiterSettings* pSettings = settings.cast<LimiterSettings>();
+   if (!pSettings)
+      pSettings = settings.cast<CompressorSettings>();
+   return pSettings->lookaheadMs * sampleRate / 1000;
+}
+
+unsigned CompressorInstance::GetAudioOutCount() const
+{
+   return 2;
```
crsib, line 128: Should this be configurable?

src/effects/Limiter.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+ComponentInterfaceSymbol EffectLimiter::GetSymbol() const
+{
+   return Symbol;
+}
+
+TranslatableString EffectLimiter::GetDescription() const
+{
+   return XO("A limiter");
```
crsib: @LWinterberg WDYT?

src/effects/Limiter.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+EffectType EffectLimiter::GetType() const
+{
+   return EffectTypeProcess;
+}
+
+auto EffectLimiter::RealtimeSupport() const -> RealtimeSince
+{
+   return RealtimeSince::After_3_1;
```
crsib, line 76: Provided that this effect is new, I think `RealtimeSince::Always` makes much more sense

crsib: The same I think applies to the compressor, as you have changed the ID

libraries/lib-effects/CMakeLists.txt
saintmatthieu: Because then they can be reused by another target, namely a unit test.

src/effects/Limiter.cpp
saintmatthieu: Let's make sure we discuss the string I pushed in a follow-up commit.

src/effects/Limiter.cpp
saintmatthieu: Yes.

src/effects/CompressorInstance.cpp
saintmatthieu: Isn't this a capability? I expect the `RealtimeEffectState` class to create the correct number of slaves, provide the correct input, etc., which, AFAICS, it does, too.

libraries/lib-effects/tests/CompressorProcessorTests.cpp
saintmatthieu: It doesn't test much indeed. I've been using it a little bit as debug helper. Maybe it will come in handy when I start tweaking the algorithm, which I consider doing after the UI follow-up work.

libraries/lib-viewport/Viewport.cpp
saintmatthieu: Rebased.

libraries/lib-effects/CMakeLists.txt
Paul-Licameli: It is extraneous to lib-effects.  I would rather give it its own library if you must have I it tests.  Ideally mid-effect-classes could be separated as in my pull request and made independent of the Audacity executable and given its own unit tests, but that presupposes much dependency breaking that is not yet figured out.

saintmatthieu: I would also like to have framework-agnostic processors have their own library, easy to re-use by other projects.
We're only at the beginning of the release cycle, the code around the compressor will not rest until 3.7, I'll have time to relocate.

src/effects/LimiterEditor.h
saintmatthieu: Done. They may be refined until the next release.

DavidBailes:
@saintmatthieu . In LimiterEditor::PopulateOrExchange, it would be better if the accessibility names of the sliders were just the names of the properties, rather than also including the units, eg, just Threshold, Knee width, etc. For these sliders, screen readers read a percentage value between 0 and 100. If units are included in the accessibility name, eg db, then the user may mistakenly assume that the percentage value is the value in db.

saintmatthieu:
@DavidBailes thanks for pointing this out.
On the other hand, the visual indication of the unit would be gone, and I haven't found an easy way of placing it somewhere else without affecting my current layout too much.
I propose to keep an eye on this until the first version is merged and create a ticket then.

saintmatthieu:
Closing in favour of #6395

DavidBailes:
> @DavidBailes thanks for pointing this out. On the other hand, the visual indication of the unit would be gone, and I haven't found an easy way of placing it somewhere else without affecting my current layout too much. I propose to keep an eye on this until the first version is merged and create a ticket then.

I think the code I was referring to may have changed, so I'll have another look when the first version is merged, as you suggested.

DavidBailes:
> @DavidBailes thanks for pointing this out. On the other hand, the visual indication of the unit would be gone, and I haven't found an easy way of placing it somewhere else without affecting my current layout too much. I propose to keep an eye on this until the first version is merged and create a ticket then.

@saintmatthieu . Now that it is merged, I've created an issue for this: https://github.com/audacity/audacity/issues/6475

Issue #6251, opened 2024-04-10
Open from cloud dialog: you can arrow past the start and end of the list


DavidBailes:
### Steps to reproduce

1. Open the Open from cloud dialog.

Case A:
2. Select the first item in the list of projects.
3. Press up arrow.

Case B:
2. Select the last item in the list of projects.
3. Press down arrow.

### Expected behavior

Case A and B. The selection is not changed.

### Actual behavior

Case A and B. No project is selected.

### Audacity Version

other

### Operating system

Windows

### Additional context

version: release-3.5.0

dozzzzer:
Tested

PR #6309, opened 2024-04-18
Open from cloud dialog: fix up/down arrow bug in list

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6251

Problem:
You can arrow past the start and end of the list.

Fix:
Don't pass on to the wxGrid for normal processing the down arrow and up arrow keydown events when they are going to cause this problem.

Recommended:


DavidBailes:
I guess this may be too late for 3.5.0.

crsib (approved):

Issue #6328, opened 2024-04-22
[FreeBSD] Build failed: libraries/lib-basic-ui/BasicUI.cpp:44:27: error: use of undeclared identifier 'environ'


VVD:
### Bug description

Build error last release 3.5.0 on FreeBSD 13.3-p1 amd64.

### Steps to reproduce

1. Build from sources

### Expected behavior

1. Build without errors

### Actual behavior

Build failed with error:
```
FAILED: libraries/lib-basic-ui/CMakeFiles/lib-basic-ui.dir/BasicUI.cpp.o
[...]

Issue #6327, opened 2024-04-22
[FreeBSD] Build failed: ld: error: relocation R_X86_64_32 cannot be used against local symbol; recompile with -fPIC: defined in cmake-proxies/pffft/libpffft.a(pffft.c.o)


VVD:
### Bug description

Build error last release 3.5.0 on FreeBSD 13.3-p1 amd64.

### Steps to reproduce

1. Build from sources

### Expected behavior

1. Build without errors

### Actual behavior

Build failed with error:
```
FAILED: Release/lib/audacity/lib-time-and-pitch.so
[...]

Issue #6326, opened 2024-04-22
[FreeBSD] Build failed: lib-src/libnyquist/nyquist/xlisp/path.c:204:9: error: unknown type name 'intptr_t'


VVD:
### Bug description

Build error last release 3.5.0 on FreeBSD 13.3-p1 amd64.

### Steps to reproduce

1. Build from sources

### Expected behavior

1. Build without errors

### Actual behavior

Build failed with error:
```
FAILED: lib-src/libnyquist/CMakeFiles/libnyquist.dir/nyquist/xlisp/path.c.o
[...]

PR #6331, opened 2024-04-23
FreeBSD compilation fixes

crsib:
Based on patches provided in #6326, #6327, #6328, #6329

Resolves: #6326
Resolves: #6327
Resolves: #6328
Resolves: #6329

Recommended:


VVD:
Don't see this commit in main. It is in release-3.5.1 only.

LWinterberg:
The release branch will get merged into master after it gets released.

PR #6333, opened 2024-04-23
leaked progress dialog window fix

### LINKED ISSUE #6321: "Progress" window stuck after applying macro to files
### Bug description

audacity

### Steps to reproduce

Tools -> Macro Manager
Apply Macro to: Files...

Select some files. The macro is applied properly and all the expected output is produced, but then it gets stuck on that black "Progress" window that cannot be closed. Nor can the "Manage Macros" window behind it.

There is no console output.

### Expected behavior

See above. This did not happen with 3.4.2, so it's something introduced by 3.5.0.

### Actual behavior

_No response_

### Audacity Version

stable version (from a Linux repository)

### Operating system

Linux

### Additional context

_No response_

saintmatthieu:
Resolves: #6321

All instances of `SqliteSampleBlockFactory` stored their progress-window instances onto the same singleton. They did so using the `Scope` mechanism, which works by temporarily swapping its state with that of the singleton.

When two objects use this mechanism in an interleaved way, things go wrong. Say `s` is our singleton, `a` our factory A, and `b`, our factory `B`:
```
s = "S"; a = "A"; b = "B";
swap(a, s); // s == "A", a == "S"
swap(b, s); // s == "B", b == "A"
swap(a, s); a.clear(); // s == "S", a == ""
swap(b, s); b.clear() // s == "A", b == ""
```
"A" is still alive. In this case, it's a modal dialog, which is in a weird, half-dead state, blocking everything else.

I don't see the need for a singleton in this particular case: `SqliteSampleBlock` instances have a reference to their factory, and can just ask them for their callback directly.

Recommended:

QA:
Please try macro-processing heavy files and/or lots of them at once, and see if you observe something weird with the progress bar.

### DIFF
--- a/libraries/lib-project-file-io/SqliteSampleBlock.cpp
+++ b/libraries/lib-project-file-io/SqliteSampleBlock.cpp
@@ -167,6 +167,11 @@ public:
    SampleBlockPtr DoCreateFromId(
       sampleFormat srcformat, SampleBlockID id) override;
 
+   SampleBlock::DeletionCallback GetSampleBlockDeletionCallback() const
+   {
+      return mSampleBlockDeletionCallback;
+   }
+
 private:
    void OnBeginPurge(size_t begin, size_t end);
    void OnEndPurge();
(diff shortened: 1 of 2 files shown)


crsib (approved):
Magnifique!

Issue #6340, opened 2024-04-24
When recording to a new track, the track area doesn't automatically scroll up to show the new track in full.


dozzzzer:
### Steps to reproduce

1. Have enough tracks to occupy the entire window's height
2. Shift + R to record to a new track
3. Observe

### Expected behavior

<img width="500" alt="image" src="https://github.com/audacity/audacity/assets/70063970/620227c2-f353-4904-b506-3bfaa7a9c0ad">

### Actual behavior

The window doesn't scroll up to put the new track entirely in the window:

<img width="500" alt="image" src="https://github.com/audacity/audacity/assets/70063970/f2f38f88-cc9e-41e0-89d2-e3de28e2c4dd">

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows, macOS, Linux

Paul-Licameli:
Bisected exactly to 2418ec4286e075a9342f626ae35ba92779236dc7

LWinterberg:
> Should it be in 3.5.2?

We'll have a pretty rapid 3.6 coming up, so there may not be a need for a 3.5.2. We'll discuss it on friday.

petersampsonaudacity:
But the workarounds are pretty simple

a) use the vertical scroll bar to bring the track in progrees of being recorded into full view, or

b) just use the keyboard down-arrow

dozzzzer:
I agree, but some might say this is a deal breaker for VO professionals who have their workflow perfectly tuned.

Paul-Licameli:
No 3.5 patch for this?  Only a merge to master?

dozzzzer:
> No 3.5 patch for this? Only a merge to master?

Correct, it will be shipped with 3.6

PR #6345, opened 2024-04-25
Issue 6340: wrong vertical scroll recording to new track...

Paul-Licameli:
... Solution is to delay the scrolling until after TrackListEvent subscriptions fire and update the cumulative ChannelView heights.

Resolves: #6340

Recommended:

### DIFF
--- a/src/ProjectAudioManager.cpp
+++ b/src/ProjectAudioManager.cpp
@@ -952,8 +952,15 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
          trackList.RegisterPendingNewTracks(std::move(*newTracks));
          // Bug 1548.  First of new tracks needs the focus.
          TrackFocus::Get(project).Set(first);
-         if (!trackList.empty())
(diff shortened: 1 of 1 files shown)


src/ProjectAudioManager.cpp
```diff
@@ -952,8 +952,15 @@ bool ProjectAudioManager::DoRecord(AudacityProject &project,
          trackList.RegisterPendingNewTracks(std::move(*newTracks));
          // Bug 1548.  First of new tracks needs the focus.
          TrackFocus::Get(project).Set(first);
-         if (!trackList.empty())
-            Viewport::Get(project).ShowTrack(**trackList.rbegin());
+         if (!trackList.empty()) {
+            BasicUI::CallAfter([pProject = project.weak_from_this()]{
+               if (!pProject.expired()) {
```
crsib, line 957: if (auto project = pProject.lock())
   ...

crsib (approved):
I would prefer to see `if` written differently, I don't have to strong opinion in this case.

LWinterberg (approved):

PR #6354, opened 2024-04-27
Fix building on OpenBSD

brad0:
OpenBSD does not have dlinfo().

Recommended:


shoogle:
Does OpenBSD use AppImage? If not, you could disable compilation of this entire file in [src/CMakeLists.txt](https://github.com/audacity/audacity/blob/ae0ab087588d5c54c1231ecdf42f4eead5aa11be/src/CMakeLists.txt#L1250).

brad0:
> Does OpenBSD use AppImage? If not, you could disable compilation of this entire file in [src/CMakeLists.txt](https://github.com/audacity/audacity/blob/ae0ab087588d5c54c1231ecdf42f4eead5aa11be/src/CMakeLists.txt#L1250).

Ah, I was wondering what this bit of code was about. AppImage is for Linux and does not apply to *BSD.

LWinterberg (approved):

Issue #6346, opened 2024-04-25
Master channel implementation


ianalarina:
### Your idea

Please, implement master channel according to the designs

chinakov:
Issues found, moved back to In Progress, and commented on the relevant MR.

PR #6359, opened 2024-04-29
Master effects

vsverchinsky:
Resolves: #6346

Known issues:
 - UI: Master effects section does not fit automatically to the contents
 - No micro fades applied to solo/gain change
 - Latency increased for track controls in real-time (delay between the moment you press TC button and the moment you hear changes is audiable)

Recommended:

QA:
- [x] Several projects can be open at once, each with master channel effects, and behavior is as expected.
- [x] Real-time effects can be added and removed during playback - works for both track and master channel.
- [x] Real-time effect UI indicates track name or "Master", depending.
- [x] Playback of MIDI file in otherwise empty project works (i.e., MIDI messages are sent, and only silence is played back).
- [x] Mute / Solo during playback still works (there may be a click audible but that shouldn't be a regression)
- [x] Playback of several tracks, some with effects with latency and others without, within a loop, yield expected result.
- [x] Master track effects are applied when exporting but not rendering (in place or to new track) or stereo-to-mono
- [x] Export does delay compensation
- [x] Export works for files of 10 seconds or more
- [x] Master effects are persistent across closing/re-opening a project
- [x] Master effect stacks in several projects open at once do not interfere with each other (e.g. Reverb in project A isn't applied to tracks in project B).
- [x] Backward compatibility: a project with master effects can be opened in previous Audacity version, only without the master effects.
- [x] Scrubbing

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -1225,28 +1225,47 @@ bool AudioIO::AllocateBuffers(
             // Adjust mPlaybackRingBufferSecs correspondingly
             mPlaybackRingBufferSecs = PlaybackPolicy::Duration { playbackBufferSize / mRate };
 
-            const size_t totalWidth = std::accumulate(
-               mPlaybackSequences.begin(), mPlaybackSequences.end(), 0,
-               [](size_t acc, const auto &pSequence){
-                  return acc + pSequence->NChannels(); });
-
-            // mPlaybackBuffers buffers correspond many-to-one with
-            // mPlaybackSequences
-            // Except, always make at least one playback buffer, in case of
-            // MIDI playback without any audio
             mPlaybackBuffers.resize(0);
-            mPlaybackBuffers.resize(
-               std::max<size_t>(1, totalWidth));
-            // Number of scratch buffers depends on device playback channels
-            if (mNumPlaybackChannels > 0) {
-               mScratchBuffers.resize(mNumPlaybackChannels * 2 + 1);
-               mScratchPointers.clear();
-               for (auto &buffer : mScratchBuffers) {
-                  buffer.Allocate(playbackBufferSize, floatSample);
-                  mScratchPointers.push_back(
-                     reinterpret_cast<float*>(buffer.ptr()));
+            mProcessingBuffers.resize(0);
+            mMasterBuffers.resize(0);
+
+            // Always make at least one playback buffer, in case of
+            // MIDI playback without any audio
+            if(mPlaybackSequences.empty())
+               mPlaybackBuffers.resize(1);
+            else
+            {
+               mPlaybackBuffers.resize(mNumPlaybackChannels);
+               mProcessingBuffers.resize(std::accumulate(
+                  mPlaybackSequences.begin(),
+                  mPlaybackSequences.end(),
+                  0, [](int n, auto& seq) { return n + seq->NChannels(); }
+               ));
+               for(auto& buffer : mProcessingBuffers)
+                  buffer.reserve(playbackBufferSize);
+               
+               mMasterBuffers.resize(mNumPlaybackChannels);
+               for(auto& buffer : mMasterBuffers)
+                  buffer.reserve(playbackBufferSize);
+
+               // Number of scratch buffers depends on device playback channels
+               if (mNumPlaybackChannels > 0) {
+                  mScratchBuffers.resize(mNumPlaybackChannels * 2 + 1);
+                  mScratchPointers.clear();
+                  for (auto &buffer : mScratchBuffers) {
+                     buffer.Allocate(playbackBufferSize, floatSample);
+                     mScratchPointers.push_back(
+                        reinterpret_cast<float*>(buffer.ptr()));
+                  }
                }
             }
+
+            std::generate(
+               mPlaybackBuffers.begin(),
+               mPlaybackBuffers.end(),
+               [=]{ return std::make_unique<RingBuffer>(floatSample, playbackBufferSize); }
+            );
+
             mPlaybackMixers.clear();
 
             const auto &warpOptions =
@@ -1264,25 +1283,11 @@ bool AudioIO::AllocateBuffers(
             mPlaybackQueueMinimum = mPlaybackSamplesToCopy *
                ((mPlaybackQueueMinimum + mPlaybackSamplesToCopy - 1) / mPlaybackSamplesToCopy);
 
-            if (mPlaybackSequences.empty())
-               // Make at least one playback buffer
-               mPlaybackBuffers[0] =
-                  std::make_unique<RingBuffer>(floatSample, playbackBufferSize);
-
-            mOldChannelGains.resize(mPlaybackSequences.size());
-            size_t iBuffer = 0;
+            // Bug 1763 - We must fade in from zero to avoid a click on starting.
+            mOldPlaybackGain = 0.0f;
             for (unsigned int i = 0; i < mPlaybackSequences.size(); i++) {
                const auto &pSequence = mPlaybackSequences[i];
-               // Bug 1763 - We must fade in from zero to avoid a click on starting.
-               mOldChannelGains[i][0] = 0.0;
-               mOldChannelGains[i][1] = 0.0;
-
-               for (size_t jj = 0, nChannels = pSequence->NChannels();
-                  jj < nChannels; ++jj
-               )
-                  mPlaybackBuffers[iBuffer++] = std::make_unique<RingBuffer>(
-                     floatSample, playbackBufferSize);
-
+               
                // By the precondition of StartStream which is sole caller of
                // this function:
                assert(pSequence->FindChannelGroup());
@@ -1307,7 +1312,7 @@ bool AudioIO::AllocateBuffers(
                Mixer::Inputs mixSequences;
                mixSequences.push_back(Mixer::Input{ pSequence });
                mPlaybackMixers.emplace_back(std::make_unique<Mixer>(
-                  move(mixSequences),
+                  std::move(mixSequences),
                   // Don't throw for read errors, just play silence:
                   false,
                   warpOptions, startTime, endTime, pSequence->NChannels(),
@@ -1979,6 +1984,8 @@ void AudioIO::FillPlayBuffers()
    }
 }
 
+#define stackAllocate(T, count) static_cast<T*>(alloca(count * sizeof(T)))
+
 bool AudioIO::ProcessPlaybackSlices(
    std::optional<RealtimeEffects::ProcessingScope> &pScope, size_t available)
 {
@@ -1992,6 +1999,13 @@ bool AudioIO::ProcessPlaybackSlices(
    // user interface.
    bool done = false;
    bool progress = false;
+
+   // remember initial processing buffer offsets
+   // they may be different depending on latencies
+   const auto processingBufferOffsets = stackAllocate(size_t, mProcessingBuffers.size());
+   for(unsigned n = 0; n < mProcessingBuffers.size(); ++n)
+      processingBufferOffsets[n] = mProcessingBuffers[n].size();
+
    do {
       const auto slice =
          policy.GetPlaybackSlice(mPlaybackSchedule, available);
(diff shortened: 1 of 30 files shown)


dozzzzer:
@vsverchinsky when playing a mono clip, only the left channel is playing audio:

image

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2021,173 @@ bool AudioIO::ProcessPlaybackSlices(
...
+            {
+               auto& buffer = mProcessingBuffers[bufferIndex + i];
+               buffer.erase(buffer.begin(), buffer.begin() + discardable);
             }
          }
+         bufferIndex += seq->NChannels();
+      }
+   }
+
+   auto samplesAvailable = std::min_element(
```
saintmatthieu, line 2114: This crashes when playing a MIDI file on an empty project, because `mPlaybackSequences` and, consequently, `mProcessingBuffers`, are then empty.

saintmatthieu (dismissed):
Haven't finished yet the review, but stumbled on a potential crash.

libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: Maybe a solution: the size of `mMasterBuffers` equals the number of playback channels, maybe it could be used instead.
Also, AFAICS these buffers always have the same size, so a simple `mMasterBuffers[0].size()` would be more expressive and sufficient.

saintmatthieu: Or `mPlaybackBuffers`.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -148,6 +148,7 @@ struct AudioIoCallback::TransportState {
                assert(false);
                continue;
             }
+            //TODO: initialize with sequence own channels count and sampleRate
```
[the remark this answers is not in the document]
vsverchinsky: The comment isn't specifically about master channel. 
When we allocate two mono effects to one stereo track there is at least one problem with UI : you can't see the output of the second channel. 
I think configuring effects to process samples in native track format and rate could have benefit of reducing CPU load in case if playback sample rate is higher compared to playback rate. But that may be conceptually wrong.

saintmatthieu (commented):
Finished reviewing the first commit (the audio part) ; nothing major besides that crash, but hopefully this is an easy fix.
Please remove the dead code ; I don't see why it should be kept, and it's noisy and sometimes confusing.

libraries/lib-audio-io/AudioIO.cpp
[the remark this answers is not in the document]
saintmatthieu: Regarding the CPU, I don't know, it looks like there is no negotiation at the moment: the framework dictates the sampling rate to the effect instances, which I guess is much simpler. Also, I think resampling is inevitable, so I'm not sure we want to try and minimize it.
My point is that this TODO should probably be tempered with some doubt, e.g., "consider initializing (...) because (...)"

libraries/lib-realtime-effects/RealtimeEffectManager.h
```diff
@@ -166,26 +171,21 @@ class REALTIME_EFFECTS_API RealtimeEffectManager final :
    // using StateVisitor =
       // std::function<void(RealtimeEffectState &state, bool listIsActive)> ;
 
-   //! Visit the per-project states first, then states for group
+   //! Visit states for group
```
Paul-Licameli, line 174: " or for the master when group is null"

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -1225,28 +1225,47 @@ bool AudioIO::AllocateBuffers(
             // Adjust mPlaybackRingBufferSecs correspondingly
             mPlaybackRingBufferSecs = PlaybackPolicy::Duration { playbackBufferSize / mRate };
 
-            const size_t totalWidth = std::accumulate(
-               mPlaybackSequences.begin(), mPlaybackSequences.end(), 0,
-               [](size_t acc, const auto &pSequence){
-                  return acc + pSequence->NChannels(); });
```
Paul-Licameli, line 1231: Why is it correct to delete this?

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2023,208 @@ bool AudioIO::ProcessPlaybackSlices(
          // warping
          if (frames > 0) {
             size_t produced = 0;
+            // Check for asynchronous user changes in mute, solo status
```
Paul-Licameli, line 2026: Comment "mPlaybackMixers correspond one-to-one with mPlaybackSequences" above this line is no longer correct

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2023,208 @@ bool AudioIO::ProcessPlaybackSlices(
...
-               const auto put = mPlaybackBuffers[iBuffer++]->Put(
-                  warpedSamples, floatSample, produced, frames - produced);
-               // wxASSERT(put == frames);
-               // but we can't assert in this thread
-               wxUnusedVar(put);
+
+               auto& buffer = mProcessingBuffers[iBuffer++];
+               const auto offset = buffer.size();
+               //there could be leftovers from the previous pass, don't discard them
+               buffer.resize(buffer.size() + frames, 0);
```
Paul-Licameli, line 2040: I would add a comment that this `resize` is expected not to allocate memory in the worker thread, because sufficient size should have been reserved in `AllocateBuffers`.  Perhaps, test capacity before and after and `assert` no change.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2023,208 @@ bool AudioIO::ProcessPlaybackSlices(
...
+                     reinterpret_cast<const float*>(warpedSamples),
+                     produced,
+                     buffer.data() + offset);
+                  std::fill_n(
+                     buffer.data() + offset + produced,
+                     frames - produced,
+                     .0f);
+               }
+               else
+                  //TODO: mixer->Process() may be wasting CPU cycles
```
Paul-Licameli, line 2055: I don't think the mixer can be properly eliminated in the case that it does resampling of a track whose rate does not match the device output rate, or there is a time track.  There will be some state in the resampler dependent on previous input.

However, note that the mixer stands as a producer in relation to this function which is its consumer.  Perhaps, if it really matters, another thread could mix in advance.

But all that effort at overlapping computation may be unnecessary because this thread really spends most of its time idle anyway, going faster than the ultimate real-time consumption of input.

Or then again maybe it might in future because we want to have ever more capacity for complicated calculations like more effects, more stretching, more pitch shifting.

Or again, maybe you were thinking of power consumption, not performance.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2023,208 @@ bool AudioIO::ProcessPlaybackSlices(
...
+                  //TODO: fade out smoothly
+                  std::fill_n(buffer.data() + offset, frames, 0);
```
Paul-Licameli, line 2057: I notice that as part of this PR you have moved the logic to silence tracks (which might be driven by unpredictable presses of the mute buttons by the user) from the low-latency thread to this high-throughput thread.

Was that for simplification of the former, so that it does less work?  Or some other reason?

Does this comment mean you have not yet restored the microfading feature?  (I have not yet read enough to answer that for myself)

I think I like this change because as I recollect, the microfading had a dependency on the buffer size given to `audacityAudioCallback` which can depend on operating system or larency settings.  That should not be:  the duration of the fade in or out should be independent of that.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2095,63 +2106,86 @@ bool AudioIO::ProcessPlaybackSlices(
...
-      std::fill_n(buffer.begin(), samplesAvailable, 0);
+   //introduced latency may be too high...
+   //but we don't discard what already was written
+   if(samplesAvailable == 0)
+      return progress;
+
+   //Prepare master buffers.
+   //These buffers may also contain leftovers from the previous pass
+   //Remember initial offset for each of them
+   const auto masterOffsets = stackAllocate(int, mMasterBuffers.size());
```
saintmatthieu, line 2141: I have doubts about the correctness of the handling of latency for the master channel.
The `discardable` value returned by a real-time effect is the number of leading samples that can be discarded.
For the track effects, this is indeed what is done:
```cpp
         for(int i = 0; i < seq->NChannels(); ++i)
         {
            auto& buffer = mProcessingBuffers[bufferIndex + i];
            buffer.erase(buffer.begin(), buffer.begin() + discardable);
         }
```
For the master track, however, no leading samples are discarded, and then
```cpp
      unsigned bufferIndex = 0;
      for(auto& buffer : mMasterBuffers)
      {
         mPlaybackBuffers[bufferIndex++]->Put(
            reinterpret_cast<constSamplePtr>(buffer.data()),
            floatSample,
            samplesAvailable,
            0
         );
      }
```
which means trailing samples are discarded instead. It was already that way in the previous commit but I missed it.

If I'm correct, the consequence of this mistake would be that, if the master channel has an effect that adds delay, an artefact due to zeroed audio would be heard at the beginning of playback.

saintmatthieu: I also don't see the necessity of `masterOffsets`. I don't think there can be left-overs from one `ProcessPlaybackSlices` to the next, as it would mean this method would produce more audio samples than asked for, and I don't see it to be the case.

saintmatthieu (commented):
Not requesting changes because I'm not certain of my analysis, but I think there is another bug, already present in the first commit but that I missed before.

libraries/lib-audio-io/AudioIO.h
```diff
@@ -271,17 +255,24 @@ class AUDIO_IO_API AudioIoCallback /* not final */
...
    // Old gain is used in playback in linearly interpolating
    // the gain.
-   std::vector<OldChannelGains> mOldChannelGains;
+   float mOldPlaybackGain;
    // Temporary buffers, each as large as the playback buffers
    std::vector<SampleBuffer> mScratchBuffers;
    std::vector<float *> mScratchPointers; //!< pointing into mScratchBuffers
 
    std::vector<std::unique_ptr<Mixer>> mPlaybackMixers;
+   std::unique_ptr<Mixer> mMasterMixer;
```
Paul-Licameli, line 275: Unused

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2652,19 +2719,21 @@ bool AudioIoCallback::FillOutputBuffers(
...
+   // do it here instead (but not if looping or scrubbing)
+   // PRL:  Also consume from the single playback ring buffer
+   if (numPlaybackSequences == 0) {
+      mMaxFramesOutput = mPlaybackBuffers[0]->Discard(toGet);
+      CallbackCheckCompletion(mCallbackReturn, 0);
+      mLastPlaybackTimeMillis = ::wxGetUTCTimeMillis();
+      return false;
+   }
+
    // The drop and dropQuickly booleans are so named for historical reasons.
```
Paul-Licameli, line 2737: Comment refers to variables that no longer exist

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2023,208 @@ bool AudioIO::ProcessPlaybackSlices(
...
+         {
+            auto& buffer = mProcessingBuffers[bufferIndex + i];
+            buffer.erase(buffer.begin(), buffer.begin() + discardable);
+         }
+
+         bufferIndex += seq->NChannels();
+      }
+   }
+
+   //samples at the beginning of could have been discarded
```
Paul-Licameli, line 2124: delete "of"

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2023,208 @@ bool AudioIO::ProcessPlaybackSlices(
...
+   auto samplesAvailable = std::min_element(
+      mProcessingBuffers.begin(),
+      mProcessingBuffers.end(),
+      [](auto& first, auto& second) { return first.size() < second.size(); }
+   )->size();
+
+   //introduced latency may be too high...
+   //but we don't discard what already was written
+   if(samplesAvailable == 0)
+      return progress;
```
Paul-Licameli, line 2136: I am reminded that the handling of latency is still not satisfactory.  I had ideas for fixing it that I have not found time to implement.

libraries/lib-audio-io/AudioIO.cpp
dozzzzer: The problem does exist indeed. In the attached video, you can hear an audible crackle at the beginning of the playback with a non-zero latency effect (TDR Limiter) added to the stack:

https://github.com/audacity/audacity/assets/70063970/742dc31e-0188-4751-b6c6-833cd5279f0d

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2015,104 +2023,208 @@ bool AudioIO::ProcessPlaybackSlices(
...
+         masterOffsets[bufferIndex++] = buffer.size();
+         buffer.resize(buffer.size() + samplesAvailable, 0);
+      }
+   }
+
+   {
+      unsigned bufferIndex = 0;
+      for(const auto& seq : mPlaybackSequences)
+      {
+         //TODO: apply micro-fades
```
Paul-Licameli, line 2155: Is that really TODO?  I think you have conserved the micro-fade logic on the consumer side in `FillOutputBuffers`

libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: Thanks @dozzzzer .

saintmatthieu (changes requested):
https://github.com/audacity/audacity/pull/6359#discussion_r1596816243

libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: IIUC that one handles the changes in the volume slider, which are global. This is for the track-specific mute/unmute.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2682,19 +2719,21 @@ bool AudioIoCallback::FillOutputBuffers(
...
-   // And these are larger structures....
-   for (unsigned int c = 0; c < numPlaybackChannels; c++)
-      tempBufs[c] = stackAllocate(float, framesPerBuffer);
-   // ------ End of MEMORY ALLOCATION ---------------
-
    // Choose a common size to take from all ring buffers
    const auto toGet =
       std::min<size_t>(framesPerBuffer, GetCommonlyReadyPlayback());
 
+   // Poke: If there are no playback sequences, then the earlier check
```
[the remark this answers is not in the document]
Paul-Licameli: The comment refers to what now happens inside `CallbackCheckCompletion`.  Amend the comment to make that more explicit.  And it's now "below" not "above".

Paul-Licameli:
I have completed a review of changes in AudioIO.cpp.

If my contribution of per track meters is also wanted, there is a lot of needed work to make these compatible.

I may volunteer my own very different rewrite of AudioIO.cpp to support a master channel in the right way.

I understand that:
- The sequence of transformations on individual tracks includes time warping (for time track), muting (of individual tracks), effects (stacked on single tracks), panning, in that order.
- Then you mix tracks and apply the master stack, on the PRODUCER side of the RingBuffer.  (As for individual stacks.  We must avoid who knows what computations in opaque code, on the low-latency consumer side.)
- Micro-fading is still applied on the consumer side.
- There are only one or two RingBuffers corresponding to the device output channels.

The old version, without a master stack, differed in these ways:
- There was a RingBuffer for each WaveChannel.
- Applying mute and pan, and mixing of the RingBuffer outputs, was done on the consumer side.

Questions might arise about subtle differences of results, moving the muting before the effect processing.

More later...

libraries/lib-audio-io/AudioIO.cpp
Paul-Licameli: Doing latencies really right is a known bit of technical debt.  I have had ideas.  I didn't find time to elaborate and share them yet.

libraries/lib-audio-io/AudioIO.cpp
vsverchinsky: I though that mixer could be "advanced" without doing any actual work

libraries/lib-audio-io/AudioIO.cpp
vsverchinsky: > Was that for simplification of the former, so that it does less work? Or some other reason?

The reason is that we can't do master effects processing on low-latency thread and that processing should happen after all the tracks have been mixed into playback channels.

Paul-Licameli:
More:

The existing ring buffers should not be removed.  Ring buffers for the master mix should be added.  The consumer thread should take equal samples out of all of them.  Only the samples from the master mix are stored in the output buffer of the audio io callback.

This allows future development to send the buffers of pre-mixed per-track data inter-thread, back to the master thread for drawing.

 See `SendVuInputMeterData` and `SendVuOutputMeterData` which now do such inter-thread communication:  but I would need to send other buffers too.

 My thinking is this:  the consumer thread needs to have access to intermediate states of calculation of the mix samples, corresponding in time to the ultimate output.  Those intermediates should be sent to other code that performs graphical display, keeping as closely synchronized as possible with the real-time playback.  That synchronization will not be perfect, given the reality of frame refresh rates.

Still, the latency of the RingBuffer itself should be accounted for in that drawing update.  Sending the per-track information to the main thread more directly from the producer thread would make less perfect synchronization of the display.

Therefore, more data should be travelling through ring buffers from the high-throughput but variable-latency producer thread, to the more regularly updating consumer thread.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2095,63 +2106,86 @@ bool AudioIO::ProcessPlaybackSlices(
...
    auto samplesAvailable = std::min_element(
       mProcessingBuffers.begin(),
       mProcessingBuffers.end(),
       [](auto& first, auto& second) { return first.size() < second.size(); }
    )->size();
 
-   for(auto& buffer : mMasterBuffers)
-      std::fill_n(buffer.begin(), samplesAvailable, 0);
+   //introduced latency may be too high...
+   //but we don't discard what already was written
```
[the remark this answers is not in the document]
vsverchinsky: I think the answer to your question is in `FillPlayBuffers`. If for some reason `samplesAvailable` will be zeroed, then there will be more processing passes until `GetNeeded() == 0`. Otherwise if I `return false` as you suggest then extra buffer exchange step will be initiated for same frame, which I think is less preferrable.
 May be we should test how it works now with latency values that are much larger than our regular buffer sizes

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2036,8 +2037,15 @@ bool AudioIO::ProcessPlaybackSlices(
 
                auto& buffer = mProcessingBuffers[iBuffer++];
                const auto offset = buffer.size();
+               const auto toConsume = std::min(
```
saintmatthieu, line 2040: What are these new lines fixing ? It looks like we are now ensuring that no new memory is allocated in that function. This comes at the risk, though, that we provide less than `frames` samples, if for some reason the capacity of the buffer is insufficient. Wouldn't that be worse ?

I wonder if this is because of the initial playback click that we hear in @dozzzzer 's video above. I suspect this not to be due to buffer allocation, but simply because he is playing a sine wave without any sort of smoothing envelope, and it's then completely normal to hear this click.

saintmatthieu:
@Paul-Licameli good point about reverb.
I confirm that muting a track may not instantly mute the sound but any effect with a tail will continue ringing.
It may sound better but I don't think this is the wanted behaviour: muting is muting. What if the effect is a long echo ?
@LWinterberg ?

saintmatthieu:
About panning before or after, yes, if the effect is non LTI, this would make a difference. If there is one place for panning and gain, I think it should be where it is a the moment, after the effect stack. I guess other DAWs do it that way, too (or may have also a "pre-gain"), and if a non LTI effect such as a compressor comes in, it will come with its own pre-processing gain parameter.
A mid/side encoding/decoding operation is a linear operation, though, so in that particular case it should not make a difference.

libraries/lib-audio-io/AudioIO.cpp
vsverchinsky: >  It looks like we are now ensuring that no new memory is allocated in that function. 

Yes

> This comes at the risk, though, that we provide less than frames samples

Memory allocations are not better. The disadvantage of this code is that it's not capable to handle very large latencies (usually > 250ms).

> I wonder if this change is motivated by the initial playback click that we hear in @dozzzzer 's video above

No, I did that in response to Paul's comment above

saintmatthieu (commented):
Following up on https://github.com/audacity/audacity/pull/6359#discussion_r1600928970:
Vitaly and I had a conversation and he helped me understand the bigger picture. This seems like a good change to me.

Paul-Licameli:
I disapprove.

please wait to see my alternative that will be compatible with oerctrack meters.

expext it today.

saintmatthieu:
Approving for now. My purpose is not to override @Paul-Licameli's disapproval, because maybe his alternative makes this very complex logic easier ? I'll leave up to @vsverchinsky to decide but I don't want to block QA anymore.

saintmatthieu (approved):

Paul-Licameli:
Let it merge then:  but my PR will begin with reversion of all the changes in AudioIO.cpp (not of other UI changes) and then a more careful stepwise rewrite.

QA will likely need to re-test everything involving sound quality, but not the UI changes.

chinakov:
Found isssues:
1. When a User adds MT-A Plugin by Mercurial to the Master Effects, and uses the built-in bypass feature, the audio played back is panned 100% to left, instead of stereo playback. 
2. When a User adds Kilohearts Distortion plugin to the Master Effects, and playbacks some audio, full scale noise can be heard for a second after playback has been initialized.
3. If a User has added any Master Effects, and enabled them, the effects are not applied to files on file export (I am not sure what to expect here, should Master Effects render when they are enabled?)

LWinterberg:
>  (I am not sure what to expect here, should Master Effects render when they are enabled?)

Yes!

Paul-Licameli:
> Let it merge then: but my PR will begin with reversion of all the changes in AudioIO.cpp (not of other UI changes) and then a more careful stepwise rewrite.
> 
> QA will likely need to re-test everything involving sound quality, but not the UI changes.

My changes are more complicated than expected.

However, I might be fixing long standing problems with effect latencies.  Search for issues with “latency”

DavidBailes:
@vsverchinsky .  Accessibility issue, logged here: https://github.com/audacity/audacity/issues/6456

DavidBailes:
@vsverchinsky . Another accessibility issue, logged here: https://github.com/audacity/audacity/issues/6457

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -1243,7 +1243,7 @@ bool AudioIO::AllocateBuffers(
                ));
                for(auto& buffer : mProcessingBuffers)
                   buffer.reserve(playbackBufferSize);
-
```
vsverchinsky: @saintmatthieu Please avoid unnecessary formatting changes if you don't change the line itself - it complicates the search through commit history and review process

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2241,7 +2241,10 @@ bool AudioIO::ProcessPlaybackSlices(
          );
       }
    }
-   
+
+   for(auto& buffer : mMasterBuffers)
```
vsverchinsky, line 2245: There is a `finally` block already that does cleanup job

vsverchinsky:
Completed review of @saintmatthieu commits :+1:

saintmatthieu (approved):

PR #6363, opened 2024-04-29
Play Meter remains just right of AudioSetup when maximizing window

Paul-Licameli:

QA:
Test this in builds configured with and without the Audio.com upload

Recommended:


LWinterberg (approved):

Paul-Licameli:
I should have said, just right of AudioSetup or of Share Audio, depending on the build configuration.  And also assuming default preferences.

chinakov:
Tested, thanks!

PR #6377, opened 2024-05-02
New Audacity 4 Begin.

igorkorsukov:
We (me, @Eism and @grliszas14 ) are Muse employees.
Some time ago we started working on Audacity 4.
The main difference from previous projects to transfer Audacity to Qt is that in this project we are not just transferring the Audacity UI to Qt(Qml), but are starting to use a lot of developments from MuseScore.
To do this, we separate the common part in the Muse Framework, which will be used by both Audacity and MuseScore. This work has just begun; there is still a long way to go.

The main goal is to reuse as much as possible what has already been done in order to be able to quickly implement a new interface and be able to focus on the main functionality.

We also plan to combine development approaches, architecture, project structure, etc. So that the Muse Developers and the community can effectively develop both of our projects.

At this stage, AU4 will be developed as a separate project, a separate application in the `au4` directory.

I believe and hope that all our ideas will be realized and we all will get many benefits from this collaboration :)


Eism (approved):

LWinterberg:
did you mean to target this to master? not  the qt branch?

LWinterberg:
Right. I'd suggest a new branch then, maybe called `au4`.

igorkorsukov:
> Right. I'd suggest a new branch then, maybe called `au4`.

We decided that a separate branch was not needed, it would not provide any benefits and would only create unnecessary problems and costs.

Working as a master, we have the following advantages:
* If it is necessary to change AU3 for AU4, these changes will be immediately integrated.
* We will immediately see how changes in AU3 affect AU4

We have experience of moving from MuseScore 3 to MuseScore 4, we also moved there without a branch, using a separate directory - it was successful.

vsverchinsky (approved):

Issue #6357, opened 2024-04-28
Audacity swap16/32/64 functions conflict with OpenBSD swap macros


brad0:
OpenBSD has some swapping macros that originate from the endian.h / sys/endian.h header.

```
     uint64_t
     swap64(uint64_t val64);

     uint32_t
     swap32(uint32_t val32);

     uint16_t
     swap16(uint16_t val16);
```

These currently conflict with functions within Audacity. We have a bit of a hackish patch locally to undefine
the macros, but pretty much every project I've had fixed results in the internal functions being renamed. I would
file a PR but I am not sure what the preference is or naming scheme. So I am opening a bug report to spur
discussion.

```
[1/1105] /home/ports/pobj/audacity-3.5.1/bin/c++ -DCMTSTUFF -DEXPERIMENTAL_CRASH_REPORT -DEXPERIMENTAL_DRAGGABLE_PLAY_HEAD -DEXPERIMENTAL_FULL_WASAPI -DEXPERIMENTAL_HALF_WAVE -DEXPERIMENTAL_KEY_VIEW -DEXPERIMENTAL_MIDI_OUT -DEXPERIMENTAL_MODULE_PREFS -DEXPERIMENTAL_
NOISE_REDUCTION -DEXPERIMENTAL_NOTETRACK_OVERLAY -DEXPERIMENTAL_NYQUIST_SPLIT_CONTROL -DEXPERIMENTAL_PUNCH_AND_ROLL -DEXPERIMENTAL_SCIENCE_FILTERS -DEXPERIMENTAL_SCRUBBING_SCROLL_WHEEL -DEXPERIMENTAL_SCRUBBING_SUPPORT -DEXPERIMENTAL_SPECTRAL_EDITING -DEXPERIMENTAL_S
UBRIP_LABEL_FORMATS -DEXPERIMENTAL_THEMING -DEXPERIMENTAL_ZOOM_TOGGLE_BUTTON -DEXT -DUSE_NYQUIST=1 -I/home/ports/pobj/audacity-3.5.1/audacity-Audacity-3.5.1/lib-src/libnyquist/nyquist/cmt -I/home/ports/pobj/audacity-3.5.1/audacity-Audacity-3.5.1/lib-src/libnyquist/n
[...]

PR #6384, opened 2024-05-03
Rename swap functions to not conflict with OpenBSD's swap macros

brad0:
Resolves: https://github.com/audacity/audacity/issues/6357

Rename swap functions to not conflict with OpenBSD's swap macros

Recommended:


LWinterberg (approved):

brad0:
Ping.

igorkorsukov:
Maybe better to add something like this:
Stk.h
```
#ifdef swap16
#undef swap16
#endif
```

vsverchinsky:
Existing code-style suggests using lower camel-case naming convention, name like `byteSwap*` or similar will fit better

brad0:
@vsverchinsky Ping.

vsverchinsky:
@brad0 Thank you!

PR #6412, opened 2024-05-10
[AU4] Playback toolbar. Base for playback level component

Eism:

Recommended:

### DIFF
--- a/au4/src/au3wrap/CMakeLists.txt
+++ b/au4/src/au3wrap/CMakeLists.txt
@@ -26,13 +26,11 @@ set(MODULE_SRC
     ${CMAKE_CURRENT_LIST_DIR}/au3wrapmodule.h
 
     ${CMAKE_CURRENT_LIST_DIR}/wxtypes_convert.h
-    ${CMAKE_CURRENT_LIST_DIR}/audacity3project.cpp
-    ${CMAKE_CURRENT_LIST_DIR}/audacity3project.h
-    ${CMAKE_CURRENT_LIST_DIR}/audacity3playback.cpp
-    ${CMAKE_CURRENT_LIST_DIR}/audacity3playback.h
+    ${CMAKE_CURRENT_LIST_DIR}/au3project.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/au3project.h
     ${CMAKE_CURRENT_LIST_DIR}/iau3wavepainter.h
-    ${CMAKE_CURRENT_LIST_DIR}/iaudacity3playback.h
-    ${CMAKE_CURRENT_LIST_DIR}/iaudacity3audiooutput.h
+    ${CMAKE_CURRENT_LIST_DIR}/iau3playback.h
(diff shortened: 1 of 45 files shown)


au4/src/au3wrap/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ set(MODULE_SRC
     ${CMAKE_CURRENT_LIST_DIR}/audacity3playback.cpp
     ${CMAKE_CURRENT_LIST_DIR}/audacity3playback.h
```
igorkorsukov, line 32: please move to internal

au4/src/au3wrap/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ set(MODULE_SRC
...
     ${CMAKE_CURRENT_LIST_DIR}/iau3wavepainter.h
+    ${CMAKE_CURRENT_LIST_DIR}/iaudacity3playback.h
+    ${CMAKE_CURRENT_LIST_DIR}/iaudacity3audiooutput.h
```
igorkorsukov, line 35: maybe better name "iau3playback.h", "iau3audiooutput.h" ?

au4/src/au3wrap/iaudacity3audiooutput.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "playback/audiotypes.h"
+
+namespace au::au3 {
+class IAudacity3AudioOutput
+{
+public:
+    virtual ~IAudacity3AudioOutput() = default;
+
+    virtual muse::async::Promise<float> playbackVolume() const = 0;
```
igorkorsukov, line 21: Hmm, is Promise really appropriate here?
Typically, a promise is used in cases where you call some kind of asynchronous action, and you need to get the result of this action - successful, not successful

au4/src/au3wrap/iaudacity3audiooutput.h
```diff
@@ -0,0 +1,31 @@
...
+    virtual void setPlaybackVolume(float volume) = 0;
+    virtual muse::async::Channel<float> playbackVolumeChanged() const = 0;
+
+    virtual muse::async::Promise<au::audio::AudioSignalChanges> playbackSignalChanges() const = 0;
```
igorkorsukov, line 25: Hmm, is Promise really appropriate here?
Typically, a promise is used in cases where you call some kind of asynchronous action, and you need to get the result of this action - successful, not successful

au4/src/au3wrap/iaudacity3audiooutput.h
igorkorsukov: ahh, I get it, this is an asynchronous retrieval of the value... ok.

Eism: Actually, the idea is to make the interfaces of au3 classes and similar classes from the framework the same, so that it is easier to replace later

igorkorsukov (approved):

igorkorsukov (approved):

Issue #6365, opened 2024-04-30
Redesign "what's new" Dialog


LWinterberg:
figma: https://www.figma.com/file/u8nXqqmvQvTR1w0rT2TrPr/35_WX---Episode-IV%3A-A-New-Look?type=design&node-id=3217-22166&mode=design

dozzzzer:
Tested

PR #6437, opened 2024-05-14
Updated "What's new" dialog

vsverchinsky:
Resolves: #6365 

Needs correct links and images.

Recommended:


src/SplashDialog.cpp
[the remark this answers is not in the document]
vsverchinsky: When compiling out of sources you need to explicitly define `-DSHOW_WHATS_NEW_SECTION=Yes` cmake variable to enable this new dialog. CI is configured to define that variable for us, but it's not the case when compiling out of sources. Linux distros maintainers likely compile Audacity with old-style `SplashDialog` without images and links to YouTube.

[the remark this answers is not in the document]
LWinterberg: Right. Let's set SHOW_WHATS_NEW_SECTION=Yes in the source by default then, and let maintainers explicitly opt out of it if they don't want it. And if they do opt out, let's just not show any startup screen.

src/WhatsNewDialog.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+#include "../images/WhatsNewBtn.jpeg.h"
+
+namespace
+{
+
+const char* WhatsNewURL = "https://audacityteam.org/3.5.0-video";
+const char* PluginsURL = "";
+const char* MuseHubURL = "";
+
```
LWinterberg, line 40: MusehubURL and PluginsURL are identical. rename instances of PluginsURL to MuseHubURL and set MuseHubURL "https://www.musehub.com"

src/WhatsNewDialog.cpp
```diff
@@ -0,0 +1,214 @@
...
+}
+}
+
+BEGIN_EVENT_TABLE(WhatsNewDialog, wxDialogWrapper)
+   EVT_BUTTON(wxID_OK, WhatsNewDialog::OnOK)
+END_EVENT_TABLE()
+
+WhatsNewDialog::WhatsNewDialog(wxWindow* parent, wxWindowID id)
+   : wxDialogWrapper(parent, id, XO("Welcome to Audacity!"),
+      wxDefaultPosition, wxSize(906,  580))
```
LWinterberg, line 113: this is a little mighty I found. Unless there's rendering differences between platforms, I found that a window height of 480 worked well with a 20px border.
 
before: 
image

after: 
image

src/WhatsNewDialog.cpp
```diff
@@ -0,0 +1,214 @@
...
+void WhatsNewDialog::Populate(ShuttleGui& S)
+{
+   bool showSplashScreen;
+   gPrefs->Read(wxT("/GUI/ShowSplashScreen"), &showSplashScreen, true );
+
+   FSHelper helper;
+
+   S.StartHorizontalLay(wxEXPAND);
+   {
+      S.SetBorder(40);
```
LWinterberg, line 140: (this border I set to 20)

LWinterberg (approved):
still awaiting final approval for the strings from martin

LWinterberg:
Oh also, while we wait for the strings: similar to #6446 we need to hide the musehub part of this on Linux because it's not available there. 

Might look something like this:

image

src/WhatsNewDialog.cpp
vsverchinsky: This is how it looks on my machine with 40 px border :/
изображение

vsverchinsky: Ah, this is because of dpi scaling ...

LWinterberg:
Strings approved

vsverchinsky:
@LWinterberg can you please check how it looks now for you?

LWinterberg:
windows @ 100%
image

@ 150%

image

@ 225%
image

LWinterberg (approved):

vsverchinsky:
Squashed and rebased

Issue #3651, opened 2022-09-20
Playback can freeze with high latency plugins


Ryan-Miller-Audacity:
### Bug description

With some high latency plugins, playback will freeze and become inoperable after performing the latency testing steps. 

So far this has been confirmed with Z-Noise from Waves

https://user-images.githubusercontent.com/88568755/191254992-652cc11d-d9ae-439c-ba88-a2477acfb0ba.mp4

![waves-x-z-noise-latency-chart](https://user-images.githubusercontent.com/88568755/191255022-670ac650-fba9-4066-93ab-10f6c63fe94f.png)

Waves lists their latency values for their plugins here: https://www.waves.com/support/tech-specs/plugin-latency

### Steps to reproduce

1. Go to www.waves.com and register an account and get a demo license of the "Z-Noise" plugin via get demo button on that plugin's webpage: https://www.waves.com/plugins/z-noise
2. Download Waves plugin manager
3. Use Waves plugin manager to install/activate your demo license for Z-Noise
4. Open Audacity
5. Generate a rhythm track
6. Add the Z noise plugin to that track's realtime effect stack
7. Go to Tracks -> Mix -> Mix and Render to a new track
8. On the new track apply the "Invert" effect via the destructive effect dropdown
9. Click play
10. Note that playback is frozen (the play marker doesn't move) the only indicator that playback is active is the depressed play button.

### Expected behavior

Playback should not freeze and function appropriately.

### Actual behavior

Playback freezes. 

### Audacity Version

latest beta

### Operating system

Windows, macOS, Linux

crsib:
I think STR can be simplified, no mix and render is needed, although it shows that mix and render works

DavidBailes:
I can reproduce the problem using the free voxengo latency delay plugin, using default settings:
https://www.voxengo.com/product/latencydelay/

saintmatthieu:
@dozzzzer I think this is fixed now, since @vsverchinsky implemented the master channel. Could you please verify ?
Note that high-latency is still isn't properly handled, and there is another issue for that: #6519

dozzzzer:
The original issue cannot be reproduced anymore: I experience no playhead stuttering or playback freeze with the Invert applied to the second track. However, playhead still stutters when a plugin with latency (e.g. Voxengo Latency Delay) is added to the Master channel too, and even when it's disabled, stuttering persists.


[...]

PR #6442, opened 2024-05-16
Master channel

Paul-Licameli:
Resolves: #6346
Resolves: #6507
Resolves: #6519
Resolves: #3651
Resolves: #6536

Alternative implementation of changes in AudioIO.cpp that will be compatible with per-track meters.

Also fixing:

- error in handling effect latencies
- duration of microfades (for changes of pan, mute, solo, track gain, or master gain) is independent of platform or buffer preferences

QA:
- playing of MIDI tracks whether alone or with simultaneous audio
- Applying effects like Reverb (with varying stereo depth) to mono tracks
- Changing of mute, solo, pan, gain, and master gain (on the output meter) during play

Recommended:


Paul-Licameli:
Updated the draft in progress today.  It is still incomplete.

vsverchinsky (approved):
Looks good to me

vsverchinsky:
Fails on assertion when I try to apply effect with latency on master channel in `AudioIO.cpp` in debug build:
1) Generate tone 
2) Apply Voxengo and set non zero latency 
3) Press play

Issue #6427, opened 2024-05-11
Incorrect font scaling in track's control panel


SteveDaulton:
### Bug description

The text in the the control block on the left end of the track is scaled incorrectly.

### Steps to reproduce

1. Add an audio track.

### Expected behavior

![good](https://github.com/audacity/audacity/assets/11681097/d78b618e-734d-44f5-8771-6870b2054334)

Audacity  3.3.3

### Actual behavior

![font-scaling](https://github.com/audacity/audacity/assets/11681097/cf295ea5-c1b1-4191-9df4-2bb862bad63d)

Audacity 3.6.0 alpha Commit Id: a55100 

### Audacity Version

current master / alpha / nightly build

### Operating system

Linux

### Additional context

Both screenshots taken with Audacity running on the same system.

petersampsonaudacity:
I cannot reproduce this on W11 with 3.5.1 or latest master alpha build for 3.6.0

![image](https://github.com/audacity/audacity/assets/50205984/d8162e5f-5528-4778-956c-cf0f8fe20b4f)

SteveDaulton:
I suspect that this may be a Linux only issue, or perhaps  an XFCE only issue, but  as can be seen from the screenshots, the problem was not present in 3.3.3.

LWinterberg:
This is due to the font scaling you have enabled to make Audacity usable on your HiDPI display, Steve. It also makes the "Snap" checkbox and clip header larger than it should be, and it doesn't seem to enlarge the scale everywhere evenly, because your ruler marks are smaller than they should be

DavidBailes:
I've reopened the issue, because with the Windows display scaling set to 125%, the track panel looks awful.

Issue #4788, opened 2023-06-12
High DPI support


JinZhuXing:
### Your idea

I think current audacity program (for windows) does not support High DPI.
I am using 4K monitor, and using 200% scale.
But the audacity does not support high DPI.
Please make it possible.
Thank you.

### Problem to be solved

I think it will be possible.
I have no idea for it, but I think it will be possible.

dozzzzer:
We're currently switching the UI framework from wxwidgets to Qt and aren't looking to improve wx' HiDPI support as it'll be obsoleted soon anyway.

PR #6447, opened 2024-05-17
Fix HiDPI text rendering on Windows

lalitshankarchowdhury:
Resolves: #4788 (partially)
Resolves: #6427 on windows

This PR fixes rendering of text on HiDPI monitors on Windows.

## Before
Screenshot 2024-05-17 143322

## After
Screenshot 2024-05-17 143432

The issue is fixed by letting GDI automatically scale the text according to the current DPI: [source](https://blogs.windows.com/windowsdeveloper/2017/05/19/improving-high-dpi-experience-gdi-based-desktop-apps/).

What's not fixed:

- Drawing bitmaps according to scale

Recommended:

### DIFF
--- a/win/Audacity.exe.manifest
+++ b/win/Audacity.exe.manifest
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='UTF-8' standalone='yes'?>
-<assembly xmlns='urn:schemas-microsoft-com:asm.v1' manifestVersion='1.0'>
+<assembly xmlns='urn:schemas-microsoft-com:asm.v1' manifestVersion='1.0' xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
   <assemblyIdentity
       version="2.3.0.0"
       processorArchitecture="x86"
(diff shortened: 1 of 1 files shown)


LWinterberg (approved):
awesome find!

vsverchinsky:
@LWinterberg should this ticket be also added to the board?
https://github.com/audacity/audacity/issues/4788

LWinterberg:
Yeah, testing it a bit more extensively, the biggest issue with this is that the custom cursor images don't get scaled correctly. But then, they didn't before either. This is a straight improvement.

lalitshankarchowdhury:
> Yeah, testing it a bit more extensively, the biggest issue with this is that the custom cursor images don't get scaled correctly. But then, they didn't before either. This is a straight improvement.

So, GDI can only help scale text according to DPI, and custom bitmaps need to be scaled manually.

LWinterberg:
Heya @lalitshankarchowdhury 

Unfortunately we'll have to revert this change, as it introduces a number of minor and major issues on setups different to mine which make this nonfeasible to ship at the moment.

lalitshankarchowdhury:
No worries! Can you reference those issues so I can check if there's a solution for that?

LWinterberg:
https://github.com/audacity/audacity/issues/6466 are the issues we had found with it

lalitshankarchowdhury:
@LWinterberg what's the current status on this issue?

Issue #6376, opened 2024-05-02
Support clipboard promise when importing through copy/paste


saintmatthieu:
Follow-up on #6280

When copying a file from the Muse Hub, it may copy something that wasn't downloaded yet. The hub writes to the clipboard the place where the file will be once download is complete, but in the meantime the user might paste and there'll be nothing yet to be imported.

DavidBailes:
@saintmatthieu . Are you referring to copying one of the sounds from the Muse Hub. If so, how do you copy it? Is it possible to do this using the keyboard?

saintmatthieu:
Hi David,
I cannot say how the Muse Hub (or the newer version that'll soon be rolled out) can be keyboard-navigated, but this is Hub-specific. Do you have a concern?

DavidBailes:
> Hi David, I cannot say how the Muse Hub (or the newer version that'll soon be rolled out) can be keyboard-navigated, but this is Hub-specific. Do you have a concern?

Using the current version of the Muse Hub, the only way that I know of importing one of the sounds into Audacity is using drag and drop. Users of screen readers can't drag and drop items between different windows, and I don't think there is the ability to copy a sound to the clipboard.

So in the description above which mentions copying a file from the muse hub, I thought that there may be a way of copying a file in Muse hub to the clipboard using the keyboard, which then could be pasted into Audacity using the new functionality that you have provided. This would allow a way that users of screen readers could use to get sounds from the hub into Audacity, which I don't think is currently possible.

saintmatthieu:
Using ctrl-c to copy a file from the Muse Hub is going to be part of their upcoming roll-out. From what I read it is indeed meant as an accessibility improvement.

DavidBailes:
> Using ctrl-c to copy a file from the Muse Hub is going to be part of their upcoming roll-out. From what I read it is indeed meant as an accessibility improvement.

Thanks - that's good news.

dozzzzer:
Reopened: the dialog's title should be changed from "Audacity" to "Importing..." @grliszas14 @LWinterberg 

<img width="441" alt="image" src="https://github.com/audacity/audacity/assets/70063970/d5f68e84-65c0-412f-871c-5534091621a2">

chinakov:
I am putting this to Ready to Merge, as I am still unable to rectify one issue that I assume is only on my side. The issue is only connected to files that are not available locally, but available in iCloud.

PR #6458, opened 2024-05-20
Support macOS promise file handling

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6376

Getting file handle of file that hasn't been downloaded yet (in the middle of download) is a blocking call. If file is big or internet connection is slow this call freezes whole app. 
Moved blocking call to a separate thread and allowed user to cancel import (if so, user goes back to usual flow, thread is going to finish it's job but the result is ignored, thread cleans up after itself)

Recommended:

### DIFF
--- a/libraries/lib-basic-ui/BasicUI.h
+++ b/libraries/lib-basic-ui/BasicUI.h
@@ -144,6 +144,13 @@ enum ProgressDialogOptions : unsigned {
    ProgressConfirmStopOrCancel = (1 << 3),
 };
 
+enum GenericProgressDialogStyle : int {
+   ProgressCanAbort            = (1 << 0),
+   ProgressAppModal            = (1 << 1),
+   ProgressShowElapsedTime     = (1 << 2),
+   ProgressSmooth              = (1 << 3),
+};
+
 enum class ProgressResult : unsigned
 {
    Cancelled = 0, //<! User says that whatever is happening is undesirable and shouldn't have happened at all
@@ -210,7 +217,8 @@ public:
    virtual std::unique_ptr<GenericProgressDialog>
    DoMakeGenericProgress(const WindowPlacement &placement,
       const TranslatableString &title,
-      const TranslatableString &message) = 0;
+      const TranslatableString &message,
+      int style) = 0;
    virtual int DoMultiDialog(const TranslatableString &message,
       const TranslatableString &title,
       const TranslatableStrings &buttons,
@@ -311,10 +319,10 @@ inline std::unique_ptr<ProgressDialog> MakeProgress(
  */
 inline std::unique_ptr<GenericProgressDialog> MakeGenericProgress(
    const WindowPlacement &placement,
-   const TranslatableString &title, const TranslatableString &message)
+   const TranslatableString &title, const TranslatableString &message, int style = (ProgressAppModal | ProgressShowElapsedTime | ProgressSmooth))
 {
    if (auto p = Get())
-      return p->DoMakeGenericProgress(placement, title, message);
+      return p->DoMakeGenericProgress(placement, title, message, style);
    else
       return nullptr;
 }
--- a/libraries/lib-import-export/Import.cpp
+++ b/libraries/lib-import-export/Import.cpp
@@ -47,6 +47,7 @@ ImportFileHandle
 #include "Prefs.h"
 
 #include "ImportProgressListener.h"
+#include "BasicUI.h"
 
 namespace {
 
@@ -629,6 +630,28 @@ bool Importer::Import(
    {
       // Try to open the file with this plugin (probe it)
       wxLogMessage(wxT("Opening with %s"),plugin->GetPluginStringID());
+
+#ifdef __WXMAC__
+      struct stat s;
+      memset(&s, 0, sizeof(struct stat));
+      auto err = stat(fName.data(), &s);
+      if(err != 0 || (S_ISREG(s.st_mode) && s.st_blocks == 0))
+      {
+         int dialogStyle = BasicUI::ProgressCanAbort | BasicUI::ProgressAppModal | BasicUI::ProgressShowElapsedTime | BasicUI::ProgressSmooth;
+         auto dialog = BasicUI::MakeGenericProgress({}, XO("Importing files"), XO("Importing %s...").Format(fName.AfterLast(wxFileName::GetPathSeparator())), dialogStyle);
+         while(err != 0 || s.st_blocks == 0)
+         {
+            BasicUI::Yield();
+            if(dialog->Pulse() != BasicUI::ProgressResult::Success)
+            {
+               errorMessage = XO("Operation was canceled by user");
+               return false;
+            }
+            memset(&s, 0, sizeof(struct stat));
+            err = stat(fName.data(), &s);
+         }
+      }
+#endif
       auto inFile = plugin->Open(fName, pProj);
       if ( (inFile != NULL) && (inFile->GetStreamCount() > 0) )
       {
--- a/libraries/lib-wx-init/wxWidgetsBasicUI.cpp
+++ b/libraries/lib-wx-init/wxWidgetsBasicUI.cpp
@@ -182,12 +182,13 @@ namespace {
 struct MyGenericProgress : wxGenericProgressDialog, GenericProgressDialog {
    MyGenericProgress(const TranslatableString &title,
       const TranslatableString &message,
-      wxWindow *parent = nullptr)
+      wxWindow *parent = nullptr,
+      int style = ProgressAppModal | ProgressShowElapsedTime | ProgressSmooth)
       : wxGenericProgressDialog{
          title.Translation(), message.Translation(),
          300000,     // range
          parent,
-         wxPD_APP_MODAL | wxPD_ELAPSED_TIME | wxPD_SMOOTH
+         style
       }
    {}
    ~MyGenericProgress() override = default;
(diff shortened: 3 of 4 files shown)


libraries/lib-import-export/ImportProgressListener.h
```diff
@@ -26,6 +28,12 @@ class IMPORT_EXPORT_API ImportProgressListener
    };
    
    virtual ~ImportProgressListener();
+
+   /// Set of functions called by importer before getting file handle as it may take a while to process
+   /// when dealing with file promise on macOS
```
vsverchinsky, line 33: I think doxygen-style comment will be recognized as comment to the first function only, not a whole group. Should be single-line C-style comment `//`

src/ImportInfoDialog.h
```diff
@@ -0,0 +1,25 @@
...
+                          const wxSize& size = wxDefaultSize,
+                          long style = wxDEFAULT_DIALOG_STYLE | wxRESIZE_BORDER);
+    
+    bool IsCancelled();
+    
+
+private:
+    void OnCancel(wxCommandEvent& event);
+
+    wxStaticText* m_statusText;
```
vsverchinsky, line 19: Please don't mix code-style in a single header, we use camel-case prefixed with `m` for member variable names (except for POD structures)

src/ImportInfoDialog.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
vsverchinsky, line 1: File header comment is missing

src/ImportInfoDialog.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <wx/wx.h>
+#include "wxPanelWrapper.h"
+
+class ImportInfoDialog : public wxDialogWrapper
```
vsverchinsky, line 6: `final`

src/ProjectFileManager.cpp
```diff
@@ -1306,6 +1308,26 @@ class ImportProgress final
 
    }
 
+   void OnImportStarted(const wxString &filename) override
+   {
+       mImportInfoDialog = new ImportInfoDialog(filename, NULL, -1, XO("Importing files"));
```
vsverchinsky, line 1313: This dialog will appear on every import (even for tiny files), won't it be too intrusive?

libraries/lib-import-export/Import.cpp
```diff
@@ -629,7 +656,44 @@ bool Importer::Import(
    {
       // Try to open the file with this plugin (probe it)
       wxLogMessage(wxT("Opening with %s"),plugin->GetPluginStringID());
+
+#ifdef __WXMAC__
+      std::shared_ptr<std::promise<std::unique_ptr<ImportFileHandle>>> promise = std::make_shared<std::promise<std::unique_ptr<ImportFileHandle>>>();
+      std::future<std::unique_ptr<ImportFileHandle>> future = promise->get_future();
+      std::thread([promise, plugin, fName, pProj]() mutable {
+         try {
+            auto result = plugin->Open(fName, pProj);
```
vsverchinsky, line 665: Did you make sure that all calls are thread-safe?

src/ImportInfoDialog.cpp
```diff
@@ -0,0 +1,39 @@
...
+BEGIN_EVENT_TABLE( ImportInfoDialog, wxDialogWrapper )
+   EVT_BUTTON( wxID_CANCEL, ImportInfoDialog::OnCancel )
+END_EVENT_TABLE()
+
+ImportInfoDialog::ImportInfoDialog(const FilePath& filename, wxWindow* parent, wxWindowID id, const TranslatableString &title, const wxPoint &position, const wxSize& size, long style ):
+      wxDialogWrapper( parent, id, title, position, size, style )
+{
+    wxBoxSizer* sizer = new wxBoxSizer(wxVERTICAL);
+
+    m_statusText = new wxStaticText(this, wxID_ANY, wxString::Format(wxT("Importing %s..."), filename.AfterLast(wxFileName::GetPathSeparator())));
```
vsverchinsky, line 12: Use either `XO` or `_` macro for translatable strings, otherwise `gettext` will not see your string definition

src/ImportInfoDialog.cpp
```diff
@@ -0,0 +1,39 @@
...
+    SetSizerAndFit(sizer);
+    
+    int parentWidth = -1, parentHeight=-1;
+    if( GetParent() )  GetParent()->GetSize( &parentWidth, &parentHeight );
+    if (parentWidth > 400)
+       CenterOnParent();
+    else
+       CenterOnScreen();
+
+    Bind(wxEVT_BUTTON, &ImportInfoDialog::OnCancel, this, wxID_CANCEL);
```
vsverchinsky, line 27: Handler is already bound to the event via event table you've defined above, one of them should be removed

src/ProjectFileManager.cpp
```diff
@@ -1306,6 +1308,26 @@ class ImportProgress final
...
+       mImportInfoDialog->Show();
+       wxYield();
+   }
+    
+   bool OnCheckImportCancelled() override
```
vsverchinsky, line 1318: `Is...`?

vsverchinsky: Can't be called after `OnImportStartCleanup`

src/ProjectFileManager.cpp
```diff
@@ -1306,6 +1308,26 @@ class ImportProgress final
...
+   {
+        return mImportInfoDialog->IsCancelled();
+   }
+    
+   void OnImportStartCleanup() override
```
vsverchinsky, line 1323: Doesn't need to be a part of interface. Please consider using `ProjectWindows` or assigning this dialog directly to a project frame via `GetProjectFrame`

libraries/lib-import-export/Import.cpp
grliszas14: Tried it, unfortunately fopen returns false-positive true and after that when you call plugin->Open UI freezes anyway.
I think this is the only option we have. The most vulnerable scenario is when user cancels importing because thread is working until it finishes it's job and by that time we're out of scope so I've used shared_ptr to take care of promise object to be still available (avoiding potential segfault that way)

src/ProjectFileManager.cpp
grliszas14: I don't think there is a way to determine if file is already readable or in the middle of download so I guess it needs to stay that way.
There's other option though: there is a progress dialog displayed right after obtaining file handle - instead of showing ImportInfo dialog and then ProgressDialog, we could display only ProgressDialog but sooner - but this needs some refactoring because ProgressDialog expects file handle in its constructor.
Right now it looks like ImportInfo dialog "evolves" into ProgressDialog

libraries/lib-import-export/ImportProgressListener.h
grliszas14: fixed

src/ImportInfoDialog.h
grliszas14: fixed

src/ImportInfoDialog.h
grliszas14: fixed

src/ImportInfoDialog.cpp
grliszas14: fixed

src/ImportInfoDialog.cpp
grliszas14: fixed

src/ProjectFileManager.cpp
grliszas14: assigned dialog to project frame but left function and renamed it (still need to call EndModal)

vsverchinsky (approved):

libraries/lib-basic-ui/BasicUI.h
```diff
@@ -311,10 +312,10 @@ inline std::unique_ptr<ProgressDialog> MakeProgress(
  */
 inline std::unique_ptr<GenericProgressDialog> MakeGenericProgress(
    const WindowPlacement &placement,
-   const TranslatableString &title, const TranslatableString &message)
+   const TranslatableString &title, const TranslatableString &message, int style = 0x0002 | 0x0008 | 0x0020)
```
vsverchinsky, line 315: One last thing: `lib-basic-ui` is a toolkit-neutral library, better define `enum` and map it to a framework-specific value in the library that implements `BasicUI` interface, which is `lib-wx-init`

vsverchinsky (approved):

libraries/lib-import-export/Import.cpp
```diff
@@ -629,6 +630,28 @@ bool Importer::Import(
...
       wxLogMessage(wxT("Opening with %s"),plugin->GetPluginStringID());
+
+#ifdef __WXMAC__
+      struct stat s;
+      memset(&s, 0, sizeof(struct stat));
+      auto err = stat(fName.data(), &s);
+      if(err != 0 || (S_ISREG(s.st_mode) && s.st_blocks == 0))
+      {
+         int dialogStyle = BasicUI::ProgressCanAbort | BasicUI::ProgressAppModal | BasicUI::ProgressShowElapsedTime | BasicUI::ProgressSmooth;
+         auto dialog = BasicUI::MakeGenericProgress({}, XO("Audacity"), XO("This may take several seconds"), dialogStyle);
```
LWinterberg, line 641: This should be `XO("Importing files"), XO("Importing %s...")` where %s is the file that currently is being imported

LWinterberg (changes requested):

LWinterberg (approved):

vsverchinsky:
@grliszas14 for a single-commit PR's please use "rebase and merge", that makes history look much clearer. Ideally before merging a branch with multiple commits rebase your branch on top of recent master first, wait checks are complete and then do "merge commit"

Issue #2392, opened 2022-01-08
Rationalizing where new tracks are created to aid usability and consistency


petersampsonaudacity:
**Describe the bug**
Where new tracks are created has mostly been out of the control of the user. Mostly these are placed at the end (bottom) of the project, which may not be ideal particularly for users with large multi-track projects.

Many Feature Requests have been received from users who would like to control where new tracks are placed in their project.

**To Reproduce**
1. Tracks > Add New > ...
2. Edit > Clip Boundaries > Split New
3. Transport > Recording > ...
4. Generate > ...
5. Analyze > ...
6. Observe: the new track is always created at the bottom of the project.

**Expected behavior**
This was fixed a while back for 
a) Mix and Render where the new track replaces the last selected track 
b) Mix and Render to New Track where the new track goes below the last selected track
So now the behavior is inconsistent across Audacity

Since this was changed For Mix&Render to be based on selection (and not on focus) it makes sense for consistency, that other commands that create new tracks should also position the new track: 
1) based on selection - i.e. normally below last selected track, 
2) IF no selection is present THEN at the bottom of the project. 

### These commands should create the new track after the last (bottom-most) selected track
In these cases the commands act on the selection and not focus, so the new tracks will be relative to the selection.
* Tracks > Add New > Mono Track
* Tracks > Add New > Stereo Track
*Tracks > Add New > Label Track
** also applies when a label track is created as a result of adding a label when no Label Track exists
* Edit > Duplicate
* Edit > Clip Boundaries > Split New
* Transport > Recording > Record New Track
* Transport > Recording > Record}} - Append record when there are insufficient track channels selected
In cases where there is no track selected then the new track should be added at the bottom of the track.
Note that the selected track need not be an audio track, it could be a label track ''(where the command does not depend upon having audio selected)''.

### These commands should create the new track after the currently focused track
The behaviour would thus be the same for:
* Generate > ... when there is no selection
* Analyze > ... when no label track is selected
but would NOT apply to actions that require a track selection. 

### This command always goes at the top of the Project
* Tracks > Add New > Time Track

**Screenshots**
None

**Additional information (please complete the following information):**
 - OS: all
 - Version: Audacity 3.1.3 and earlier

**Additional context**
Flagged for @Tantacrul as a UX issue

petersampsonaudacity:
@LWinterberg 

this issue still gets a fair few mentions on the Forum from users

petersampsonaudacity:
[...]

PR #6469, opened 2024-05-22
Add option to set where new tracks are placed

jonata:
Resolves: #2392, [Forum discussion](https://forum.audacityteam.org/t/why-are-new-tracks-always-added-to-the-end-you-should-be-able-to-add-them-after-the-selected-track/52910/23)

When the user duplicates a selection or simply creates a new track, it is always added to the end (bottom) of the project. With this PR, we can set the default behavior/placement when new tracks are added.

Recommended:

### DIFF
--- a/libraries/lib-track/Track.h
+++ b/libraries/lib-track/Track.h
@@ -1117,6 +1117,9 @@ public:
    //! Remove and return the first track
    Track::Holder DetachFirst();
 
+   Track *GetPrev(Track &, bool linked = false) const;
+   Track *GetNext(Track &, bool linked = false) const;
+
 private:
    using ListOfTracks::size;
 
@@ -1148,9 +1151,6 @@ private:
       auto e = const_cast<TrackList*>(this)->getEnd();
       return { { b, b, e, pred }, { b, e, e, pred } };
    }
-
-   Track *GetPrev(Track &, bool linked = false) const;
-   Track *GetNext(Track &, bool linked = false) const;
 
    template < typename TrackType >
       TrackIter< TrackType >
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackMenuItems.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackMenuItems.cpp
@@ -88,7 +88,6 @@ void OnNewStereoTrack(const CommandContext &context)
       tracks.Add(track);
    }
 
-   // auto &newTrack = **tracks.rbegin();
    track->SetSelected(true);
    track->SetName(tracks.MakeUniqueTrackName(WaveTrack::GetDefaultAudioTrackNamePreference()));
 
--- a/src/tracks/playabletrack/wavetrack/ui/WaveTrackMenuItems.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveTrackMenuItems.cpp
@@ -36,7 +36,20 @@ void OnNewWaveTrack(const CommandContext &context)
 
    auto track = trackFactory.Create(defaultFormat, rate);
    track->SetName(tracks.MakeUniqueTrackName(WaveTrack::GetDefaultAudioTrackNamePreference()));
-   tracks.Add(track);
+
+   wxString newTrackPlacement = gPrefs->Read(wxT("/Tracks/NewTrackPlacement"), wxT("BottomOfProject"));
+
+   if (newTrackPlacement == wxT("TopOfProject")) {
+      tracks.Insert(*tracks.begin(), track);
+   } else if (newTrackPlacement == wxT("AboveCurrentTrack")) {
+      tracks.Insert(*tracks.Selected().begin(), track);
+   } else if (newTrackPlacement == wxT("BelowCurrentTrack")) {
+      auto last_selection = *tracks.Selected().rbegin();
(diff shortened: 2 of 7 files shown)


LWinterberg:
Please move the preference setting into the Tracks page, as it'd have a buddy in there:

image

jonata:
Sure thing. Do you already have this name feature implemented? Want me to implement in this PR?

LWinterberg:
The name feature currently lives in the "display" section on the tracks page, you'll just need to move it downstairs.

jonata:
Yes, 🤦 .

petersampsonaudacity:
Testing on W10 with @jonata 's branch build:   audacity-win-3.6.0-alpha-20240522+81bb202-x64

@LWinterberg 

Whilst the placement settings work properly for adding new stereo and mono tracks and for duplications - when adding a Label track this is always added to the bottom of the project, regardless of the settings.

Users who are advanced users of Sync-lock will want their label tracks in various places in their project as label tracks are used to define sync-locked track groups - so adding or duplicating a label track should also follow the placement setting the user makes in prefs.

_It is OK that adding a Time Track always goes to the top of the project - that has always been the case, it was that way by design at the outset when they were first introduced._

jonata:
Hey @petersampsonaudacity , thanks for the feedback. So, do you suggest that we can have two settings, one for "wave" tracks and another for "labels" track? I'm asking as someone who doesn't uses label tracks so often, so maybe a different setting for different type of tracks would be useful...?

petersampsonaudacity:
> Hey @petersampsonaudacity , thanks for the feedback. So, do you suggest that we can have two settings, one for "wave" tracks and another for "labels" track? I'm asking as someone who doesn't use label tracks so often, so maybe a different setting for different type of tracks would be useful...?

Hi @jonata  - You'd have to discuss that with @LWinterberg really. it would be his call for if we had an extra setting for Label tracks _(I'm just a part-time pro-bono external tester, and documenter, these days, not a Muse employee)._

But note that further testing shows that Split New and Record do _**NOT**_ observe the placement setting for new audio tracks either - and I think they should,  Same for the Analyze tools that produce label tracks as their output.

petersampsonaudacity:
> Do you already have this name feature implemented? Want me to implement in this PR?

AFAICT that seem to be properly working already @jonata

jonata:
> But note that further testing shows that Split New and Record do _**NOT**_ observe the placement setting for new audio tracks either - and I think they should,
> 
> Same for the Analyze tools that produce label tracks as their output.

I agree. We should define in which actions this would be useful.

> AFAICT that seem to be properly working already jonata

It is, my fault not noticing it :)

src/tracks/playabletrack/wavetrack/ui/WaveTrackMenuItems.cpp
```diff
@@ -59,16 +72,31 @@ void OnNewStereoTrack(const CommandContext &context)
...
+      tracks.Insert(*tracks.begin(), track);
+   } else if (newTrackPlacement == wxT("AboveCurrentTrack")) {
+      auto focusedTrack = TrackFocus::Get(project).Get();
+      tracks.Insert(focusedTrack, track);
+   } else if (newTrackPlacement == wxT("BelowCurrentTrack")) {
+      auto focusedTrack = TrackFocus::Get(project).Get();
+      tracks.Insert(tracks.GetNext(*focusedTrack), track);
+   } else {
+      tracks.Add(track);
+   }
```
grliszas14, line 89: Similar block of code repeats 3 times - it can be one function used across these 3 places

libraries/lib-track/Track.h
```diff
@@ -1117,6 +1117,9 @@ class TRACK_API TrackList final
    //! Remove and return the first track
    Track::Holder DetachFirst();
 
+   Track *GetPrev(Track &, bool linked = false) const;
+   Track *GetNext(Track &, bool linked = false) const;
```
grliszas14, line 1121: that looks like an unnecessary change

jonata: It's needed, yes.

src/tracks/playabletrack/wavetrack/ui/WaveTrackMenuItems.cpp
jonata: @grliszas14 , do you have a suggestion where this function could be placed? I was thinking on `WaveTrackUtilities`, or even on `Tracks.cpp` but here I was not able to import `TrackFocus`.

petersampsonaudacity:
> I think that when creating new tracks, the focused track should be used, and for Duplicate, the selected track or tracks should be used.

@DavidBailes 
I am somewhat minded to agree

a) In the simple use case the user has a single track selected which gets the focus because of its selection - so focus and selection are the same

b) it gets more complex in the use case with multiple tracks selected when it is likely the the last track selected will get the focus

But my big concerns are:
1) most users overlook Focus and have no idea what it is
2) the Focus border color in the new Light theme is quite hard to spot - see #6702

LWinterberg:
Ah, sorry, I dropped the ball on this one. Same as the others - we just [switched master to 4.0.0](https://forum.audacityteam.org/t/master-has-been-switched-to-4-0-0-development/120444) which makes this PR incompatible. Audacity 4 will rework track importing and placing rules (specifically, when dragging and dropping it'll be placed at your mouse cursor).

PR #6472, opened 2024-05-22
[AU4] Timeline ruler

grliszas14:


au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/Timeline.qml
```diff
@@ -42,13 +42,14 @@ Rectangle {
     }
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    StyledTextLabel {
+    TimelineRuler {
+        id: ruler
+        offset: timelineContext.offset
+        zoom: timelineContext.zoom
```
igorkorsukov, line 48: It may be better to pass the entire context as a property, this is done in other places, this can be useful when more properties from it are needed (for example, a selected loop)

au4/src/projectscene/view/clipsview/timeformat.h
```diff
@@ -0,0 +1,25 @@
...
+
+  Audacity: A Digital Audio Editor
+
+**********************************************************************/
+
+#pragma once
+
+#include "rulerformat.h"
+
+class TimeFormat final : public RulerFormat
```
igorkorsukov, line 11: At the moment, the reasons why цу need to have a base class with virtual methods and an inheritor are not clear. I believe this is not necessary.

au4/src/projectscene/view/clipsview/timeformat.h
```diff
@@ -0,0 +1,25 @@
...
+{
+public:
+    static const TimeFormat& Instance();
```
igorkorsukov, line 14: In AU4 we don't use singletons

au4/src/projectscene/view/clipsview/timeformat.h
```diff
@@ -0,0 +1,25 @@
...
+    ~TimeFormat() override;
+
+    void SetTickSizes(
+        double units, double& major, double& minor, double& minorMinor, int& mDigits) const override;
+
+    void SetLabelString(
+        QString& s, double d, double units, double minor, int mDigits, TickType tickType) const override;
```
igorkorsukov, line 21: I think it’s possible in one line... (in AU3, in my opinion, the code width is too small, I think this is a legacy of many years ago, when there were small monitors with low resolution)

au4/src/projectscene/view/clipsview/timeformat.cpp
```diff
@@ -0,0 +1,237 @@
...
+const TimeFormat &TimeFormat::Instance()
+{
+    static TimeFormat instance;
+    return instance;
+}
+
+void TimeFormat::SetTickSizes(
+    double zoom, double& major, double& minor, double& minorMinor,
+    int& mDigits
+    ) const
```
igorkorsukov, line 18: It's best to avoid in-out arguments unless clearly necessary.
For example, here I think it’s better to do this:
```
TickInfo TimeFormat::tickInfo(double zoom) const 
{
   ...
}
```

au4/src/projectscene/view/clipsview/timeformat.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    // NOTE: 24 is minimal px size between minorMinor ticks
+    // 0.000005 is distance [s] between minorMinor ticks
+    // same goes for other time intervals
+    if (zoom > (24 / 0.000005)) { // 0.000005s ticks
+        minor = 0.00002;
+        major = 0.0001;
+        minorMinor = 0.000005;
+        mDigits = 6;
+    }
```
igorkorsukov, line 28: missing return?

au4/src/projectscene/view/clipsview/timeformat.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    minor = 24 * 3600 + 1; // HACK to not to draw minor ticks labels
+    major = 7 * 24 * 3600.0;
+    minorMinor = 24 * 3600.0;
+    return;
+}
+
+void TimeFormat::SetLabelString(
+    QString& s, double d, double, double minor, int mDigits, TickType tickType
+    ) const
```
igorkorsukov, line 179: better 
```
QString TimeFormat::label(const TickInfo& info, TickType tickType) const
```

au4/src/projectscene/view/clipsview/timeformat.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "rulerformat.h"
+
+class TimeFormat final : public RulerFormat
+{
+public:
+    static const TimeFormat& Instance();
+    ~TimeFormat() override;
+
+    void SetTickSizes(
```
igorkorsukov, line 17: In AU4 we use `camelCase` for method names

au4/src/projectscene/view/clipsview/timeformat.h
igorkorsukov: And I think these methods can be made static for convenience

au4/src/projectscene/view/clipsview/timeformat.cpp
```diff
@@ -0,0 +1,237 @@
...
+        }
+        else if (minor >= 60.0) {
+            int minutes = (int)(d / 60.0 + 0.5);
+            QString m;
+            if (minutes >= 60)
+            {
+                m = QTime(minutes / 60, minutes % 60).toString("HH:mm:ss");
+            }
+            else
+            {
```
igorkorsukov, line 203: see https://github.com/musescore/MuseScore/wiki/CodeGuidelines

au4/src/projectscene/view/clipsview/timeformat.cpp
```diff
@@ -0,0 +1,237 @@
...
+                s = QString("%1").arg(static_cast<int>(floor(d + 0.5)));
+            }
+            else {
+                s = QString::number(d, 'f', mDigits);
+            }
+        }
+    }
+}
+
+TimeFormat::~TimeFormat() = default;
```
igorkorsukov, line 237: not necessary for us

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+    : QQuickPaintedItem(parent)
+{
+}
+
+void TimelineRuler::paint(QPainter* painter)
+{
+    assert(painter != nullptr);
+    if (painter == nullptr) {
+        return;
+    }
```
igorkorsukov, line 29: here the painter cannot be null, there is no mandatory check

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    const auto w = width();
+    const auto h = height();
```
igorkorsukov, line 32: Please do not use auto unless necessary (for example, if the type is too complex)

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+    QVector<QLineF> ticks;
+    TickLabels labels;
+
+    // determine current time interval and prepare ticks
+    TimeFormat format;
+    format.SetTickSizes(m_zoom, tickInfo.major, tickInfo.minor, tickInfo.minorMinor, tickInfo.digits);
+    prepareTickData(ticks, labels, tickInfo, format, w, h);
+
+    // begin painting
+    painter->save();
```
igorkorsukov, line 44: There’s no need to call saves and restore here.

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    if (m_tickColor.isValid()) {
+        drawTicks(painter, ticks);
+    }
+
+    painter->restore();
+}
+
+void TimelineRuler::prepareTickData(QVector<QLineF>& ticks, TickLabels& labels, TickInfo& tickInfo, TimeFormat& format, double w, double h)
```
igorkorsukov, line 59: there are a lot of objects being passed by non-constant reference, are they all changed by this method?

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+    m_textFont.setPixelSize(FONT_SIZE);
+    QPen pen = painter->pen();
+
+    for (qsizetype i = 0; i < labels.count(); i++) {
+        m_textFont.setBold(std::get<2>(labels[i]) == RulerFormat::TickType::MAJOR);
+
+        pen.setColor(m_textColor);
+
+        painter->setFont(m_textFont);
+        painter->setPen(pen);
```
igorkorsukov, line 117: pen (color) does not change, so there is no need to set it in the loop

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+void TimelineRuler::drawTicks(QPainter* painter, const QVector<QLineF>& ticks)
+{
+    QPen pen = painter->pen();
+    for (auto i = 0; i < ticks.count(); ++i) {
+        m_tickColor.setAlphaF(i % 10 == 0 ? TICK_ALPHA_MAJOR : TICK_ALPHA_MINOR);
+        pen.setColor(m_tickColor);
+
+        painter->setPen(pen);
```
igorkorsukov, line 129: pen (color) does not change, so there is no need to set it in the loop

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+void TimelineRuler::SetTextFont(const QFont& font)
+{
+    if (m_textFont != font) {
+        m_textFont = font;
+        update();
+        emit textFontChanged();
+    }
+}
+
+QColor TimelineRuler::TextColor() const
```
igorkorsukov, line 176: In AU4 we use `camelCase` for method names

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,207 @@
...
+        m_tickColor = color;
+        update();
+        emit tickColorChanged();
+    }
+}
+
+void TimelineRuler::componentComplete()
+{
+    QQuickItem::componentComplete();
+}
```
igorkorsukov, line 207: not need

au4/src/projectscene/view/clipsview/timelineruler.h
```diff
@@ -0,0 +1,73 @@
...
+#include <QtGui/QColor>
+#include <QtGui/QFont>
+#include <QtQuick/QQuickPaintedItem>
+
+#include "rulerformat.h"
+#include "timeformat.h"
+
+class QPainter;
+
+using TickLabelInfo = std::tuple<int, QString, RulerFormat::TickType>;
```
igorkorsukov, line 12: It’s better to use a structure with named names, it’s much more convenient and clearer than `std::get<0>`

au4/src/projectscene/view/clipsview/timelineruler.h
```diff
@@ -0,0 +1,73 @@
...
+private:
+    void prepareTickData(QVector<QLineF>& ticks, TickLabels& labels, TickInfo& tickInfo, TimeFormat& format, double w, double h);
+    void drawLabels(QPainter* painter, const TickLabels& labels, double w, double h);
+    void drawTicks(QPainter* painter, const QVector<QLineF>& ticks);
+
+    int m_offset { 0 };
+    double m_zoom{ 0 };
+    QFont m_textFont{};
+    QColor m_textColor{};
+    QColor m_tickColor{};
```
igorkorsukov, line 72: there is no need to explicitly call initialization here

au4/src/projectscene/view/clipsview/timelineruler.h
```diff
@@ -0,0 +1,73 @@
...
+protected:
+    void componentComplete() override;
+
+private:
+    void prepareTickData(QVector<QLineF>& ticks, TickLabels& labels, TickInfo& tickInfo, TimeFormat& format, double w, double h);
+    void drawLabels(QPainter* painter, const TickLabels& labels, double w, double h);
+    void drawTicks(QPainter* painter, const QVector<QLineF>& ticks);
+
+    int m_offset { 0 };
+    double m_zoom{ 0 };
```
igorkorsukov, line 69: in AU4 we prefer to use `=` for initialization, and only use `{}` where necessary
```
double m_zoom = 0.0;
```

au4/src/projectscene/view/clipsview/timeformat.h
grliszas14: It makes sense if you know that there can be different types of rulers in Audacity. I implemented timeline but you can also have "Beats and measures" ruler - which needs to have the same methods implemented. I can remove it right now if you want but we're gonna need it later anyway

au4/src/projectscene/view/clipsview/timelineruler.h
grliszas14: Isn't it a good habit to use {} initialization? I mean it has advantage of additional check - it saves you from conversions that lose information like:
int b{8.1} -> will produce compilation error
I know it's rather rare case but :P

au4/src/projectscene/view/clipsview/timeformat.h
igorkorsukov: Yes, I assumed that there might be something like that, but then the architecture must be appropriate. 
that is, then it turns out that there should be something like this:
* Single implementation of timeline for different things
* Must be type timeline 
* There must be a factory that created a formatter for a given type.

(These are all assumptions, I didn't look at how it was done in AU3 :) )

Yes, everything may be like this... but it may not be like that :) Therefore, I suggest not complicating things until it is clear why it is needed. Because in reality it may be a little different.

au4/src/projectscene/view/clipsview/timelineruler.h
igorkorsukov: There may be some benefits to using `{}`, but the big value is readability, and easy to read code means consistent code. It's common practice in the muse framework to use `=`, and there's no good reason to change hundreds of thousands of lines of code to do otherwise :)

More philosophical thoughts, if we use `{}` to initialize class members, then it is consistent to use `{}`  to initialize local variables, but in practice this is not the case, in one place it’s one thing, in another place it’s another :)

Another point, I often see `double v {}`, but this only raises questions about what the default value is, and how much time is spent on thinking depends on the developer’s experience :) In the `double v = 0.0;` option, for developers of any experience the same minimum time of awareness.

i.e., pros and cons of both approaches can be found

If it is important for you to use `{}`, for technical reasons, or you are used to it and think that it is more correct, then I do not insist on using `=`. 
I see that AU3 uses `{}`, so it could be a big cognitive problem to switch to `=` for no obvious reason (other than code consistency)

grliszas14: It's not that important for me, I'll switch to `=`, that's not a problem :) It's just the first time I see a different opinion about that, but it's always good to get a different point of view - that's why I asked. Good point with `double v{}`, I didn't think about that

au4/src/projectscene/view/clipsview/timeformat.h
grliszas14: Sure, then I'm going to remove base class for now

au4/src/projectscene/view/clipsview/timeformat.cpp
```diff
@@ -0,0 +1,213 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+**********************************************************************/
+
+#include "timeformat.h"
+
+TickInfo TimeFormat::tickInfo(
+    double zoom)
```
igorkorsukov, line 10: I think it can be one line

grliszas14: fixed

au4/src/projectscene/view/clipsview/timelineruler.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+void TimelineRuler::drawTicks(QPainter* painter, const Ticks& ticks)
+{
+    QPen pen = painter->pen();
+
+    for (auto i = 0; i < ticks.count(); ++i) {
+        QColor tickColor = ticks[i].tickType == TickType::MAJOR
+                           ? QColor(uiconfiguration()->currentTheme().values.value(muse::ui::STROKE_COLOR).toString())
+                           : QColor(uiconfiguration()->currentTheme().values.value(muse::ui::FONT_PRIMARY_COLOR).toString());
```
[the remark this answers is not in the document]
grliszas14: fixed

Issue #4361, opened 2023-02-26
Ability to search in tools


metemaddar:
### Your idea

A search bar like the dash in Ubuntu, that helps to find and filter tools by keyword. The keyword can search in tags and name of tools.
The results can also search in tips and helps. These results can be categorized in (tools/tips/etc..). I think this can also search in some communities which leads to more powerful work results.

Like dash which can also search in tags of applications:
![Screenshot from 2023-02-26 18-20-14](https://user-images.githubusercontent.com/3726265/221417892-547fd15b-e81e-41db-b6d5-dbb323c453b7.png)

### Problem to be solved

The main problem: It is hard to find by only reading in some situations.
When you are using audacity to perform edits on sound, you maybe in imagination mode in mind. You need some tool, but you don't know how to do that. Or you can not remember the exact name of the tools. But you may have some ideas in your mind which can fit to some or one of the tools or helps.

LWinterberg:
![image](https://user-images.githubusercontent.com/87814144/221843712-c0969493-fe0b-4026-89a0-70e96d0fedec.png)
sort of like this? :)

metemaddar:
Yes. Exactly

Issue #4420, opened 2023-03-16
Macros can launch windows they're not supposed to


LWinterberg:
![image](https://user-images.githubusercontent.com/87814144/225693294-658c8ed5-bf1b-4a41-aeee-2138afd2388c.png)

I changed some Items with an ellipsis at the end to exclude them, or reworded them altogether (eg 622b1b2c427af9774816d9ebf89ebb706ffb8c03 ). This broke the blacklist for which items should be available as macros. 

Paul says the fix looks as thus, in the example of HelpMenu.cpp:

At top of function
```cpp
using namespace MenuTable;
BaseItemSharedPtr HelpMenu()
{
   using Options = CommandManager::Options;
```
Then later
```cpp
         Command( wxT("About"), XXO("&About Audacity"), OnAbout,
            AlwaysEnabledFlag,
            Options{}.AllowInMacros(false) )
```

Paul-Licameli:
As I mentioned to you @LWinterberg : the same fix should also be applied to every command in the menus, that isn't a plug-in, with an English label containing "..." .

And then this brittle piece of code in CommandManager.cpp should eliminate the test of the string characters:

```
   bool excludeFromMacros =
      (options.allowInMacros == 0) ||
      ((options.allowInMacros == -1) && label.MSGID().GET().Contains("..."));
```

petersampsonaudacity:
I thought there was a discussion a while back wher it was said that all the ellipses should be removed from menu commands ?

PR #6497, opened 2024-05-26
Implemented search functionality: #4361

andIsome:
Resolves: [*(direct link to the issue)*](https://github.com/audacity/audacity/issues/4361)

Introduced a new ToolBar: SearchToolBar which gives queries suggestions from `CommandManager`. Upon hitting enter it executes the corresponding command using `HandleTextualCommand`. 

A Preview: 
image

Recommended:

### DIFF
--- a/libraries/lib-menus/MenuRegistry.cpp
+++ b/libraries/lib-menus/MenuRegistry.cpp
@@ -221,7 +221,7 @@ void MenuRegistry::Visit(Visitor<Traits> &visitor, AudacityProject &project)
 //"ShowMeterTB,"
 "ShowMixerTB,"
 "ShowEditTB,ShowTranscriptionTB,ShowScrubbingTB,ShowDeviceTB,ShowSelectionTB,"
-"ShowSpectralSelectionTB") },
+"ShowSpectralSelectionTB,ShowSearchTB") },
          {wxT("/Tracks/Add/Add"), wxT(
 "NewMonoTrack,NewStereoTrack,NewLabelTrack,NewTimeTrack")},
          {wxT("/Optional/Extra/Part2/Scriptables1"), wxT(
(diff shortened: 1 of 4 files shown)


LWinterberg:
This is really cool!

It was you who asked how to override the UI  from listening to keyboard inputs, right? Can't find where you posted that, so I'm answering here: A good starting point for that might be [`LabelTrackView::IsGoodLabelFirstKey`](https://github.com/audacity/audacity/blob/master/src/tracks/labeltrack/ui/LabelTrackView.cpp#L1303) and the following functions, as labels face the same trouble (though they of course live in a different context): 

Some more issues I noticed: 

- [ ] (Fix the key capturing, see above)
- [ ] When clicking on one of the suggested options, it filled out the textbox and required another Enter press. Clicking on an option should execute it immediately. 
- [ ] For settings-menu items (such as the items in View > Toolbars), there is no indication that they're settings. They really need to show the checkbox if they have one, as otherwise it's completely unexpected that some part of the UI suddenly disappears (which can be the search toolbar itself!). Alternatively, filtering settings out of the result and only leaving actual actions would be acceptable, too. The filtering infrastructure would be in place, but still needs to be set for all the commands we *don't* want (see https://github.com/audacity/audacity/issues/4420) which is something we haven't gotten around to yet. It's arguable whether blacklist for macros should be the same as the blacklist for search - if it isn't, the filtering could be done analogously though.
- [ ] accessibility - once inside the textbox, it's impossible to get out of it and towards other toolbars using Tab. (Ctrl+F6 still toggles between top/bottom toolbars and the project itself)
- [ ] When clicking into the textbox, it sometimes was reluctant to show a cursor, or accept any input. I'm not entirely sure what the steps to reproduce are (and it's really annoying to test as long as half the key inputs get eaten by something else)

I'll put this PR back into draft status until at least the first 4 issues in the checklist are fixed, but I'm already really liking where this is going.

LWinterberg:
@andIsome @pedrofpclima are you looking to finish this feature?

PR #6500, opened 2024-05-27
bugfix: scroll-to-track after import is complete

### LINKED ISSUE #6361: View scrolls to first track on importing audio file
### Bug description

Bug reported here: https://forum.audacityteam.org/t/adding-files-but-sending-me-to-the-top/105025

> when I add new files to my project, the new Audacity is sending me back to the start (top) of the project instead of keeping me where the new files are. It’s not a huge problem as I can use the bar on the right to scroll down to the bottom again - it’s just a bit annoying, especially when I’m dealing with over 40 files.
It didn’t do that before I updated to the new version last week.

### Steps to reproduce

Import multiple files one at a time. Scroll down to the bottom after each import.

### Expected behavior

The imported file to be visible at the bottom of the tracks window.

### Actual behavior

The track window scrolls to the top after each import.

### Audacity Version

### Operating system

Linux

### Additional context

Regression against Audacity 3.3.0 (I have only tested this on 3.3 and 3.5).

saintmatthieu:
Resolves: #6361

Recommended:

### DIFF
--- a/src/DropTarget.cpp
+++ b/src/DropTarget.cpp
@@ -156,9 +156,8 @@ public:
    {
       // Experiment shows that this function can be reached while there is no
       // catch block above in wxWidgets.  So stop all exceptions here.
-      return GuardedCall<bool>([&] {
-         return ProjectFileManager::Get(*mProject).ImportAndArrange(filenames);
-      });
+      return GuardedCall<bool>(
+         [&] { return ProjectFileManager::Get(*mProject).Import(filenames); });
    }
 
 private:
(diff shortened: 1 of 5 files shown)


LWinterberg (approved):

saintmatthieu:
@LWinterberg the first commit fixed the problem for drag-and-drop, but not when importing via the menu.
To fix this, the last commit makes it such that all public `Import` methods of the `ProjectFileManager` end up scrolling to the last imported track.
I guess this is desired, but maybe you can think of a case when importing should _not_ result in auto-scrolling ?

LWinterberg (approved):
I cannot think of any case where you wouldn't want the track you just added to not be visible.

Issue #6448, opened 2024-05-17
Audacity fails to find modules when installed in a directory with a dollar sign in its path


dozzzzer:
### Steps to reproduce

1. install Audacity to a non-default path containing a dollar sign (e.g. `D:\$Root\$Apps\Audacity\3.5.1\Audacity.exe`)
2. run Audacity
3. check Preferences > Modules

### Expected behavior

All modules show up

### Actual behavior

No modules are discovered by Audacity

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

chinakov:
Just tested, the issue is still present. I have tried to delete Audacity.cfg and create a Portable Settings folder, to no avail to rectify the issue. Bringing back to In Progress.

dozzzzer:
Tested

PR #6502, opened 2024-05-27
bug 6448: Do not use wxWidget env-variable expand flag when adding to `audacityPathList`

saintmatthieu:
Resolves: #6448 

`wxFileName::Normalize()` by default uses (among others) the `wxPATH_NORM_ENV_VARS`.
On Windows, this environment-variable path expansion option has a [bug](https://github.com/wxWidgets/wxWidgets/issues/19214), causing #6448. (Note that the way this was resolved by wxWidgets was by discouraging the use of that option, rather than fixing the problem itself.)

We are using `Normalize` as Audacity collects paths where to look for modules. Looking at the code ending at the place where we make this call to `Normalize`, it seems to me that environment variable expansion is not needed, and can hence be disabled (see diff).

We make other calls to `Normalize` in other contexts, which I reviewed. There seems not to be a need to disable this option there, but I would like to have the opinions of the reviewers.

Recommended:

### DIFF
--- a/libraries/lib-files/FileNames.cpp
+++ b/libraries/lib-files/FileNames.cpp
@@ -172,7 +172,7 @@ bool FileNames::HardLinkFile( const FilePath& file1, const FilePath& file2 )
 #ifdef __WXMSW__
 
    // Fix forced ASCII conversions and wrong argument order - MJB - 29/01/2019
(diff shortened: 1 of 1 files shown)


saintmatthieu:
Closing this PR because mistakenly not from a fork. Opened https://github.com/audacity/audacity/pull/6503 instead.

PR #6503, opened 2024-05-27
bug 6448: Do not use wxWidget env-variable expand flag when adding to `audacityPathList`

### LINKED ISSUE #6448: Audacity fails to find modules when installed in a directory with a dollar sign in its path

saintmatthieu:
Resolves: #6448 

`wxFileName::Normalize()` by default uses (among others) the `wxPATH_NORM_ENV_VARS`.
On Windows, this environment-variable path expansion option has a [bug](https://github.com/wxWidgets/wxWidgets/issues/19214), causing #6448. (Note that the way this was resolved by wxWidgets was by discouraging the use of that option, rather than fixing the problem itself.)

We are using `Normalize` as Audacity collects paths where to look for modules. Looking at the code ending at the place where we make this call to `Normalize`, it seems to me that environment variable expansion is not needed, and can hence be disabled (see diff).

We make other calls to `Normalize` in other contexts, which I reviewed. There seems not to be a need to disable this option there, but I would like to have the opinions of the reviewers.

Recommended:

### DIFF
--- a/libraries/lib-files/FileNames.cpp
+++ b/libraries/lib-files/FileNames.cpp
@@ -172,7 +172,7 @@ bool FileNames::HardLinkFile( const FilePath& file1, const FilePath& file2 )
 #ifdef __WXMSW__
 
    // Fix forced ASCII conversions and wrong argument order - MJB - 29/01/2019
-   //return ::CreateHardLinkA( file1.c_str(), file2.c_str(), NULL );  
+   //return ::CreateHardLinkA( file1.c_str(), file2.c_str(), NULL );
    return ( 0 != ::CreateHardLink( file2, file1, NULL ) );
 
 #else
@@ -504,7 +504,7 @@ wxFileNameWrapper FileNames::DefaultToDocumentsFolder(const wxString &preference
 
    // MJB: Bug 1899 & Bug 2007.  Only create directory if the result is the default path
    bool bIsDefaultPath = result == defaultPath;
-   if( !bIsDefaultPath ) 
+   if( !bIsDefaultPath )
    {
       // IF the prefs directory doesn't exist - (Deleted by our user perhaps?)
       //    or exists as a file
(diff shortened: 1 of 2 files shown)


LWinterberg (approved):

libraries/lib-files/FileNames.cpp
```diff
@@ -626,7 +626,12 @@ void FileNames::AddUniquePathToPathList(const FilePath &pathArg,
                                           FilePaths &pathList)
 {
    wxFileNameWrapper pathNorm { pathArg };
-   pathNorm.Normalize();
+   // https://github.com/audacity/audacity/issues/6448 :
+   // Do not seek to expand environment variables here: it is not needed, and
+   // wxWidgets has an issue doing so - See
+   // https://github.com/wxWidgets/wxWidgets/issues/19214
+   const auto flags = wxPATH_NORM_ALL & ~wxPATH_NORM_ENV_VARS;
```
vsverchinsky, line 633: One thing that bothers me a bit is that `AddUniquePathToPathList` isn't unique anymore

vsverchinsky (approved):

libraries/lib-files/FileNames.cpp
saintmatthieu: You mean it was unique before this change and now it isn't ? I don't understand.

saintmatthieu:
@LWinterberg @vsverchinsky the first commit fixed the scanning of the module paths, but not the readout from `audacity.cfg`, which had the same problem again. The second commit should address that.

libraries/lib-files/FileNames.cpp
vsverchinsky: No, I just didn't think well enough about it, please ignore that comment

LWinterberg (approved):

vsverchinsky (approved):

Issue #6456, opened 2024-05-20
Realtime effects accessibility: there is no indication for screen reader users which controls are for master effects


DavidBailes:
### Steps to reproduce

With the NVDA screen reader running:
1. Press E to open the realtime effects pane.
2. Tab through the controls in the pane.

### Expected behavior

There should be an indication which controls are for master  and which are for the focused track

### Actual behavior

There is no indication which controls are for master and which are for the focused track.

### Audacity Version

other

### Operating system

Windows

### Additional context

Audacity version: https://github.com/audacity/audacity/commit/e5267d86470f8f74c040527f994adf51502124a1

DavidBailes:
@dozzzzer . I think this is a serious accessibility bug. Please could you assign an priority to it?

DavidBailes:
@LWinterberg . Please can this be added to 3.6, rather than 3.6.1. This issue makes this it difficult for screen readers to use this new feature.

PR #6514, opened 2024-05-28
Realtime effects accessibility

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6456

Problem:
As you tab through the controls in the realtime effects pane, there is no indication which controls are for the focused track and which for master.

Fix:
1. Include the header windows in the Tab order.

2. Set the accessibility name of the header window for master effects.

Recommended:


DavidBailes:
@vsverchinsky . Updated the PR so that the only change in the tab order is that the headers are included.

vsverchinsky (approved):

Issue #6477, opened 2024-05-23
Master effects: capitalize "effects" in the name


dozzzzer:
<img width="369" alt="image" src="https://github.com/audacity/audacity/assets/70063970/d2ffddc6-647e-431e-81fd-00fd9a5806c2">

petersampsonaudacity:
Perhaps also in these places too:

![image](https://github.com/audacity/audacity/assets/50205984/17f6dcb3-b1da-4814-b4b0-84520ebf68ae)

LWinterberg:
I'd say no - title case for a title makes some sense, title case for sentences and buttons isn't necessary.

PR #6517, opened 2024-05-28
String tweaks

LWinterberg:
Resolves: #6477

Also simplifies the cloud sync success dialog by removing a button. (expected behavior: done opens the project online, X only closes the dialog)

Also clarifies the cloud save option in the save dialog more. 

Also removes the "Save project is for project files not audio files" warning, instead it's now included in the save dialog directly.

Also renames Extra -> Tracks -> Close track... to Delete track...

Also moves Dmitry to emeritus.


Paul-Licameli:
Should you also move Dmitry to emeritus in AboutDialog.cpp?

saintmatthieu (approved):

PR #6533, opened 2024-05-29
[AU4] restruct of playback

igorkorsukov:


igorkorsukov:
> Explain please?

Synchronized with the structure in MuseScore.
Our goal is to make applications similar with MuseScore, so as not to waste time on what has already been implemented and to be able to focus on the unique functions of Audacity.

Paul-Licameli:
> > Explain please?
> 
> Synchronized with the structure in MuseScore. Our goal is to make applications similar with MuseScore, so as not to waste time on what has already been implemented and to be able to focus on the unique functions of Audacity.

How deep is this work?  Will it rewrite the inter-thread communications that happen now in AudioIO.cpp?

igorkorsukov:
> How deep is this work? Will it rewrite the inter-thread communications that happen now in AudioIO.cpp?

Not now, at the moment these are only wrappers for AU3 (AudioIO)

au4/src/playback/internal/playbackcontroller.cpp
```diff
@@ -179,78 +175,62 @@ void PlaybackController::togglePlay()
...
-    msecs_t endMsecs = playbackEndMsecs();
-    msecs_t newPosition = !args.empty() ? args.arg<msecs_t>(0) : 0;
-    newPosition = std::clamp(newPosition, startMsecs, endMsecs);
+    // msecs_t startMsecs = playbackStartMsecs();
+    // msecs_t endMsecs = playbackEndMsecs();
+    // msecs_t newPosition = !args.empty() ? args.arg<msecs_t>(0) : 0;
+    // newPosition = std::clamp(newPosition, startMsecs, endMsecs);
 
-    seek(newPosition);
+    // seek(newPosition);
```
Eism, line 201: can you please restore this?

igorkorsukov:
> “Not now” suggests that you mean “later.”
> 
> Are you sure you won’t discover things AudioIO.cpp does that MuseScore does not and have to rethink this?

We have not yet explored this issue, but we plan to study it in the future, and then we will see.

au4/src/playback/internal/playbackcontroller.cpp
igorkorsukov: The method is called rewindToStart, but we accept an optional position, should we change the name of the method or remove the optional position? In what cases can a position be put?

Eism: let's remove the optional position

igorkorsukov: fixed

Eism (approved):

PR #6537, opened 2024-05-30
6508 limiter graph input output envelopes

### LINKED ISSUE #6508: Limiter graph: draw waveform envelope
### Your idea

Add waveform envelope to existing target and actual gain reduction

### Problem to be solved

May increase intelligibility of the graph and make it even prettier ;)

### Prior art

_No response_

### Additional context

_No response_

saintmatthieu:
Resolves: #6508

In addition to #6508, this PR introduces several smaller improvements.
To give reviewers an idea, the result at the time of writing is something like this:
image

Recommended:

### DIFF
--- a/src/effects/DynamicRangeProcessorHistoryPanel.cpp
+++ b/src/effects/DynamicRangeProcessorHistoryPanel.cpp
@@ -88,10 +88,6 @@ DynamicRangeProcessorHistoryPanel::DynamicRangeProcessorHistoryPanel(
 
 namespace
 {
-constexpr auto followLineWidth =
-   DynamicRangeProcessorPanel::transferFunctionLineWidth + 2;
-constexpr auto topMargin = (followLineWidth + 1) / 2;
-
 double GetDisplayPixel(float elapsedSincePacket, int panelWidth)
 {
    const auto secondsPerPixel =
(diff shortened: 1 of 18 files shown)


src/effects/DynamicRangeProcessorHistoryPanel.cpp
```diff
@@ -26,7 +26,10 @@ namespace
 {
 constexpr auto defaultHeight = 100;
 constexpr auto timerId = 7000;
-constexpr auto timerPeriodMs = 1000 / 50;
+// Of course we aren't really targetting 200fps, but when specifying 50fps, we
+// rather get 30fps, with outliers at 20. Measurements (Windows) showed that,
+// when specifying 200, we get around 60fps on average, with outlier around 40.
+constexpr auto timerPeriodMs = 1000 / 200;
```
saintmatthieu, line 32: Here some image for comparison: 
Elapsed-time-comparison

saintmatthieu: The x-axis is elapsed times between two ticks.

vsverchinsky: `wxTimer` isn't precise mechanism indeed, alternatively you can try set 1ms update period and call `wxWindow::Refresh`(without `wxWindow::Update`) from timer update function making an early return from it until next `OnPaint`. I don't think it will yield better result, but then you don't have to explain numbers.
Btw, I checked Auburn Couture plugin and it's a bit jittery too.
👍

vsverchinsky (approved):

src/effects/DynamicRangeProcessorHistoryPanel.cpp
saintmatthieu: Interesting idea. I just tried it, but indeed, no improvement.

src/effects/DynamicRangeProcessorTransferFunctionPanel.cpp
```diff
@@ -61,19 +77,14 @@ void DynamicRangeProcessorTransferFunctionPanel::OnPaint(wxPaintEvent& evt)
    wxPaintDC dc(this);
    dc.Clear();
 
-   dc.SetBrush(*wxWHITE_BRUSH);
+   dc.SetBrush(DynamicRangeProcessorPanel::backgroundColor);
```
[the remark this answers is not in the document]
saintmatthieu: No, at least not yet. It doesn't offer much choice, and the colors are being revisited by @jessjwilliamson at the moment. We can do this later if it get prioritized.

Issue #5626, opened 2023-11-15
Quality slider won't move when exporting Ogg Vorbis


dozzzzer:
### Steps to reproduce

1. grab some audio
2. Ctrl + Shift + E
3. Select Format: Ogg Vorbis
4. Look at the Quality option slider

### Expected behavior

The quality of the output file can be set using the slider

### Actual behavior

The quality slider cannot be moved:

![image](https://github.com/audacity/audacity/assets/70063970/16f14129-6b5b-4b24-b5ea-b2e10d9a01d4)

### Audacity Version

current release-3.4.2 branch

### Operating system

Ubuntu 22.04

### Additional context

No such issue with the Quality slider when exporting Opus

WangleLine:
![Audacity_YfpaNqPppL](https://github.com/audacity/audacity/assets/34343336/97415cae-0324-489c-8900-e02a7d99d3b5)
Having a similar issue, where the quality slider is unusually short in width

dozzzzer:
Tested

![image](https://github.com/audacity/audacity/assets/70063970/316b8583-1166-46b1-a1ad-eb0236a9e360)

PR #6547, opened 2024-05-30
Fixes the unusually small quality slider in export audio panel when selecting ogg vorbis

nalin243:
Resolves: *#5626*

*Manually set the quality slider size for any sliders that are added to the audio options area in export audio panel so we have a consistent slider size always.*

oggvorbis

Recommended:


LWinterberg (approved):

LWinterberg:
(your change looks good, but it seems like artifactory.audacityteam.org is down atm)

nalin243:
No problem.

LWinterberg:
Thanks for the patch!

Issue #6617, opened 2024-06-06
Compressor and Limiter accessibility: check boxes for parameters shown in graph


DavidBailes:
### Bug description

When the compressor and limiter are used as realtime effects, there are a number of check boxes immediately above a graph which determine which parameters are shown in the graph:

![image](https://github.com/audacity/audacity/assets/11887021/a1327cfe-b78e-4025-9b77-f4a495a4d8bc)

For sighted users, the purpose of these check boxes is reasonably obvious.
However, for users of screen readers, the names of the check boxes, eg input, output, without their visual context, do not make their purpose at all obvious - do they affect the output of the effect?

### Steps to reproduce

1. With a screen reader running, open the compressor or limiter as a realtime effect.
2. Tab to the check boxes above the graph.

### Expected behavior

The purpose of the check boxes above the graph should be clear.

### Actual behavior

The purpose of the check boxes above the graph is not clear given their names

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

DavidBailes:
@saintmatthieu . I could fix this by making the accessibility names of the check boxes "show input", "show output" etc.
But what's happening about translations?

petersampsonaudacity:
@DavidBailes   AFAICT these checkboxes only affect the display, the graph,  They determine which elements are shown.

But that graph is not particularly useful for VI users I'd have thought ...

DavidBailes:
@saintmatthieu . Created a PR for this. But it does include new strings, and don't know whether or not this is OK at the moment.

DavidBailes:
@LWinterberg . Please could this PR be included in 3.6.0 ? I appreciate that the strings "show overshoot", "show undershoot", etc are not currently translated, but neither are "overshoot", "undershoot", etc.

saintmatthieu:
@DavidBailes we've postponed the release by quite some time, and as a result decided we'll need another round for translations. New strings are therefore OK.

chinakov:
Tested, both plugins have 'Show Overshoot' and 'Show Undershoot' voiced lines for the checkboxes. Thanks!

PR #6619, opened 2024-06-06
Compressor and Limiter accessibility: check boxes

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6617

Problem:
When the compressor and limiter are used as realtime effects, there are a number of check boxes immediately above a graph which determine which parameters are shown in the graph. For sighted users, the purpose of these check boxes is reasonably obvious. However, for users of screen readers, the names of the check boxes, eg input, output, without their visual context, do not make their purpose at all obvious.

Fix:
Change the accessibility names of the check boxes to "Show input", "Show output", etc.

Recommended:


saintmatthieu (approved):
Thanks @DavidBailes for your contribution!

Issue #6564, opened 2024-05-31
Swapping or splitting of channels does not update display


Paul-Licameli:
### Steps to reproduce

1. Create a stereo track with visibly dissimilar channels (such as by generating a silent mono track, another with sound, and joining into stereo)
2. Swap channels
3. Play
4. Undo
5. Redo

### Expected behavior

At steps 2 and 5, expect the display to change
At step 3, the expected sound in the expected speaker

### Actual behavior

Step 3 and 5 DO behave as expected, but not so step 2

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

Paul-Licameli:
This bisects to the bracket 0ffa73b9f44acc339e12a350edadd96590dc4b17..3de5915b00e22aa5b6f5d7cbafd91c9493cd10cb which is this PR from @vsverchinsky : https://github.com/audacity/audacity/pull/5818

petersampsonaudacity:
Testing on W10 with latert 3.6.0 alpha master

At Step 2 I expect to see the channels swapped - BUT this is not the case.

However playing the audio at step 2 shows the channels are indeed swapped

petersampsonaudacity:
Also if after step 2 
![image](https://github.com/audacity/audacity/assets/50205984/5fea860a-ba40-4b4f-b443-a356c7eb82b6)


2.1  split stereo track
2.2 Observe the two tracks are both the same - the top track
![image](https://github.com/audacity/audacity/assets/50205984/0bff5d4b-c4e7-42e7-a304-b4f562b638da)


BUT playing the audio shows that the clicktrack is still present.

petersampsonaudacity:
These behaviors are a regression on 3.5.1

dozzzzer:
Tested

Paul-Licameli:
Did you test both swapping and splitting of channels?

dozzzzer:
Yes, when either swapping or splitting channels, it works just fine.
However, I've just discovered that if you first swap channels, and then split (using either of the two methods), Audacity freezes (macOS) or crashes (Windows).

chinakov:
The issue described by @dozzzzer is fixed.

PR #6636, opened 2024-06-07
Swap channel display caches on channel swap

vsverchinsky:
Resolves: #6564 

Recommended:

### DIFF
--- a/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
+++ b/src/tracks/playabletrack/wavetrack/ui/WaveformView.cpp
@@ -289,8 +289,9 @@ public:
 
    void SwapChannels() override
    {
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
Testing on W10 with @vsverchinsky 's branch build for this:  audacity-win-3.6.0-alpha-20240607+ef0fbc5-x64

This now appears to be properly fixed.

Paul-Licameli:
What about splitting channels?

petersampsonaudacity:
> What about splitting channels?

When i say "properly fixed" - I mean properly for _**all**_ errant behaviors reported in the thread - not just the STRs

So, yes splitting also works fine again with Vitaly's fix

saintmatthieu (approved):

PR #6657, opened 2024-06-11
6589 limiter graph pause interaction fix

### LINKED ISSUE #6589: New Compressor and Limiter RT graph can become blank after using Pause
### Bug description
Testing on W10 with the new Compressor and Limiter in the alpha test builds for 3.6.0

In the realtime graphs in both of these new/replacement effects if you press Pause to temporarily stop playback the graph continues scrolling but becomes blank from when the pause started.

a) If you press Pause  to restart playback while at least some of the graph is shown in the window then the graph will continue. However ...

b) If you press Pause again when all the active graph has scrolled off the screen with a totally blank graph - then the graphing display does not restart even though playback is active.  Nor does the graph restart if you change any of the realtime parameters.

### Steps to reproduce
1. get some audio
2. add Compressor (or Limiter) to the RT effects stack
3. Click on it in the RT stack to open the dialog
4. check on the Show graph
5. Press play
6. Observe the compression graph
7. Press pause
8. Observe: graph continues scrolling but blank beyond pause point
9. While some graph still shows press Pause to restart playback
10. Observe: graph continues where it left off (no gap)
11. Press Pause 
12. wait till all the graph scrolls of screen - leaving the graph blank
13. Press pause to restart the audio playback
14. Observe: graph does _**NOT**_ restart

### Expected behavior
Graph should always show when playback is arrive

### Actual behavior
Graph does not always show when playback is active

### Audacity Version

### Operating system
Windows 10

### Additional context
None

saintmatthieu:
Resolves: #6589

In addition to the strictly necessary changes to fix the problem reported in #6589, 115e46fa5f12101330ae6b1ff23e506201792796 and 4756b37334080f6adc642b3237ece04c1ec6f665 also address some apparent mistake:
`ProcessingScope` was always calling `RealtimeEffectManager::ProcessStart(bool suspended)` and `ProcessEnd(bool suspended)`  with `false`, while dead code nearby suggested that the correct value should be `true` when playback was paused.
Rectifying this has the (positive?) consequence that `RealtimeSuspend()` and `RealtimeResume()` is called on `EffectInstance` implementations also when pausing or un-pausing playback. Testing should therefore cover existing implementations (see QA list).

Recommended:

### Limiter-specific:
- [ ] Interactions of play/pause/bypass behave as expected on the graph
- [ ] If the editor is opened during playback/pause/bypass state, further play/pause/bypass interactions behave as expected.
### General
- [ ] Anything unusual when pausing/resuming playback with an effect from any of the following, please report:
* AU
* Ladspa
* LV2
* VST
* VST3

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -1701,6 +1701,7 @@ void AudioIO::SetPaused(bool state)
    }
 
    mPaused.store(state, std::memory_order_relaxed);
+   Publish({ mOwningProject.lock().get(), AudioIOEvent::PAUSE, state });
 }
 
 double AudioIO::GetBestRate(bool capturing, bool playing, double sampleRate)
--- a/libraries/lib-audio-io/AudioIO.h
+++ b/libraries/lib-audio-io/AudioIO.h
@@ -62,6 +62,7 @@ struct AudioIOEvent {
       PLAYBACK,
       CAPTURE,
       MONITOR,
+      PAUSE,
    } type;
    bool on;
 };
--- a/libraries/lib-audio-io/AudioIO.h
+++ b/libraries/lib-audio-io/AudioIO.h
@@ -169,7 +169,7 @@ public:
    std::shared_ptr< AudioIOListener > GetListener() const
       { return mListener.lock(); }
    void SetListener( const std::shared_ptr< AudioIOListener > &listener);
-   
+
    // Part of the callback
    int CallbackDoSeek();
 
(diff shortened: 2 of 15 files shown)


petersampsonaudacity:
Testing on W10 with @saintmatthieu 's latest branch build for this fix:  audacity-win-3.6.0-alpha-20240611+a470e1c-x64

This now appears to work fine:

a) if you press Pause during playback in either of the two effects then the graph now stops and doesn't continue scrolling blank.  And when you press pause to restart playback then the graph restarts.

b) also works if you press pause and then switch from RT Compressor to RT Limiter and vice versa.

Looks to be functionally fixed to me.

grliszas14 (approved):

Issue #6450, opened 2024-05-17
A series of error messages pops up when one of the values exceeds the upper boundary 


dozzzzer:
### Steps to reproduce

1. Open up a built-in effect either in the realtime stack or via the main Effect menu (e.g. Limiter)
2. Type a value that exceeds the upper boundary for that text box (e.g. +2 dB in "Make-up target")
3. Try to move any other slider
4. observe

### Expected behavior

TBD @LWinterberg ?

### Actual behavior

A series of 3 to 6 identical error messages (depending on an effect) pops up immediately:

https://github.com/audacity/audacity/assets/70063970/9c6bb5f6-513c-4b41-94de-0dc630a366ea

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

LWinterberg:
added expected behavior

nalin243:
@chinakov If I do what you asked, then it becomes impossible to even input an invalid value. The code to do so was already there but somebody chose not to do it this way for some reason. If this is fine then I can open a pull request and the maintainers can see if what is happening here might break something. 



https://github.com/audacity/audacity/assets/29351098/1be396c6-13d5-4b17-94a3-524d13e60ca4

chinakov:
@nalin243 - alright, no problem, then. I will verify this task as done. Maybe there is a reason why the code was not used indeed.

PR #6696, opened 2024-06-14
Fixed excessive validation messages due to invalid value

nalin243:
Resolves: #6450 

*I added extra code in the DoValidateNumber function for both IntegerValidatorBase and FloatingPointValidatorBase which sets the value to the closest valid value.*

Recommended:


saintmatthieu (approved):

chinakov:
Hi, @nalin243 - it seems that the issue described in the updated Expected behaviour is fixed, and the task can be closed. However, I wanted to ask if it would be possible to try to fix one more thing regarding the boundary of the user input in the effects? 

https://github.com/audacity/audacity/assets/139755141/1c64bf51-0883-4508-bf74-e1af48667cf5


Basically, if the User doesn't touch the sliders after adjusting one of the values outside of its assigned boundary, but inputs new values via numerical input, the value is not automatically adjusted. I attached a video.

nalin243:
@chinakov I will look into it and try.

PR #6726, opened 2024-06-20
[AU4] Audio output preferences

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6594

### DIFF
--- a/au4/src/appshell/qml/Preferences/internal/AudioApiSection.qml
+++ b/au4/src/appshell/qml/Preferences/internal/AudioApiSection.qml
@@ -22,24 +22,31 @@
 import QtQuick 2.15
 
 import Muse.UiComponents 1.0
+import Audacity.Playback 1.0
 
 BaseSection {
     id: root
 
-    title: qsTrc("appshell/preferences", "Audio")
+    title: qsTrc("appshell/preferences", "Audio settings")
 
     property int currentAudioApiIndex: -1
     property var audioApiList: null
 
     signal currentAudioApiIndexChangeRequested(int newIndex)
 
-    /*
-     * TODO: https://github.com/musescore/MuseScore/issues/9807
+    PlaybackStateModel {
+        id: playbackState
+    }
+
+    // /*
+    //  * TODO: https://github.com/musescore/MuseScore/issues/9807
     ComboBoxWithTitle {
         id: apiComboBox
 
         title: qsTrc("appshell/preferences", "Audio API:")
         columnWidth: root.columnWidth
+
+        enabled: !(playbackState.isPaused() || playbackState.isPlaying())
 
         currentIndex: root.currentAudioApiIndex
         model: root.audioApiList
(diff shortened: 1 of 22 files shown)


au4/src/au3wrap/internal/au3audiodevicesmanager.cpp
```diff
@@ -20,6 +21,89 @@ void Au3AudioDevicesManager::init()
     initInputChannels();
 }
 
+QStringList Au3AudioDevicesManager::getAudioOutputDevices() const
+{
+    QStringList outputDevices;
```
igorkorsukov, line 26: please don't use Qt here (use muse::String...) 
we try not to use Qt where it can be omitted :) i.e. Qt is used mainly only for view

au4/src/au3wrap/internal/au3audiodevicesmanager.h
```diff
@@ -9,13 +9,26 @@
...
     void init();
 
+    QStringList getAudioOutputDevices() const override;
+    QString currentAudioOutputDevice() const override;
+    void setAudioOutputDevice(const QString& deviceName) override;
+    muse::async::Notification audioOutputDeviceChanged() const override;
+
+    void handleDeviceChange() override;
+
+    QStringList getHosts() const override;
```
igorkorsukov, line 27: I believe the term `audio api` is more understandable than `host` in this case.

igorkorsukov: but maybe I'm wrong :)

grliszas14: Just followed au3 style at first, but audioApi sounds more understandable for me as well, changed :+1:

au4/src/au3wrap/internal/au3audiodevicesmanager.h
```diff
@@ -6,16 +6,30 @@
...
     void init();
 
+    muse::StringList audioOutputDevices() const override;
+    muse::String currentAudioOutputDevice() const override;
+    void setAudioOutputDevice(const muse::String& deviceName) override;
+    muse::async::Notification audioOutputDeviceChanged() const override;
+
+    void handleDeviceChange() override;
+
+    muse::StringList audioApi() const override;
```
igorkorsukov, line 28: I think better to name `audioApiList`

au4/src/appshell/qml/Preferences/internal/AudioApiSection.qml
```diff
@@ -22,25 +22,32 @@
...
+        id: playbackState
+    }
+
+    // /*
+    //  * TODO: https://github.com/musescore/MuseScore/issues/9807
     ComboBoxWithTitle {
         id: apiComboBox
 
-        title: qsTrc("appshell/preferences", "Audio API:")
+        title: qsTrc("appshell/preferences", "Host")
```
igorkorsukov, line 46: I think better return "Audio API:"

au4/src/au3wrap/internal/au3audiodevicesmanager.h
```diff
@@ -6,16 +6,30 @@
...
+    int currentAudioApiIndex() const override;
+    void setAudioApi(int audioApiIndex) override;
```
igorkorsukov, line 30: I know that it's done similar to MuseScore... But it looks so-so... It would be more appropriate to have something like:
```
muse::String currentAudioApi() const override;
void setAudioApi(const muse::String& audioApi) override;
```

igorkorsukov (approved):

igorkorsukov (approved):

PR #6745, opened 2024-06-24
Undo stack is cleared before project is compacted

### LINKED ISSUE #6453: Label track deleted after Tempo (or Time Signature)change and closing project (without saving)
### Bug description

Make a tempo change in the time signature toolbar to a project with a label track.
Close the project without saving.
The label track is deleted.

The result is even worse than in issue #6166, where the label track is only not in sync.

### Steps to reproduce

1. Generate Track
2. Add Label Track
3. Save Project - Save to computer
4. Close and reopen the file
5. Change Tempo (or Time Signature) in the Time Signature Toolbar
6. Close file - without saving
7. Reopen file

### Expected behavior

Closing a project does not change the project file at all.

### Actual behavior

After Step 4:

DTMF_after save

After Step 5 - tempo changed:

DMTF_after tempo change

After Step 6 - reopened project file which was previously only closed without saving:

DMTF_after tempo change _close_reopen

### Audacity Version

### Operating system

Windows

### Additional context

Tested with Audacity 3.6.0-alpha-20240514  (audacity-win-3.6.0-alpha-20240514+3af928c-x64-msvc2022.zip).

saintmatthieu:
Resolves: #6453

This particular PR fixes actually [this problem](https://github.com/audacity/audacity/issues/6453#issuecomment-2186662391), but same ticket is reused.

Recommended:

QA:
- [ ] [this problem](https://github.com/audacity/audacity/issues/6453#issuecomment-2186662391) is fixed

### DIFF
--- a/src/ProjectManager.cpp
+++ b/src/ProjectManager.cpp
@@ -530,17 +530,21 @@ void ProjectManager::OnCloseWindow(wxCloseEvent & event)
    // TODO: Is there a Mac issue here??
    // SetMenuBar(NULL);
 
-   // Set (or not) the bypass flag to indicate that deletes that would happen during
-   // the UndoManager::ClearStates() below are not necessary.
-   projectFileIO.SetBypass();
-
    // This can reduce reference counts of sample blocks in the project's
-   // tracks. Do this before `CompactProjectOnClose` so that changes after the
-   // last save don't get written to the project file.
+   // tracks. No need to have `SetBypass()` called yet because the `tracks`
+   // object still holds strong references to the tracks.
+   // Do this before `CompactProjectOnClose()`, though, so that changes that are
+   // not track-specific (like project tempo) after the last save don't get
+   // written to the project file.
    UndoManager::Get(project).ClearStates();
 
    // Compact the project.
    projectFileManager.CompactProjectOnClose();
+
+   // Set (or not) the bypass flag to indicate that deletes that would happen
+   // during tracks.Clear() below are not necessary. Must be called after
+   // `CompactProjectOnClose()`.
+   projectFileIO.SetBypass();
 
    // Delete all the tracks to free up memory
    tracks.Clear();
--- a/src/ProjectManager.cpp
+++ b/src/ProjectManager.cpp
@@ -530,21 +530,20 @@ void ProjectManager::OnCloseWindow(wxCloseEvent & event)
    // TODO: Is there a Mac issue here??
    // SetMenuBar(NULL);
 
-   // Compact the project.
-   projectFileManager.CompactProjectOnClose();
-
    // Set (or not) the bypass flag to indicate that deletes that would happen during
    // the UndoManager::ClearStates() below are not necessary.
    projectFileIO.SetBypass();
 
-   {
-      // This can reduce reference counts of sample blocks in the project's
-      // tracks.
-      UndoManager::Get( project ).ClearStates();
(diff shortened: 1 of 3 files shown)


petersampsonaudacity:
Testing on W10 with @saintmatthieu 's latest branch build for this:  audacity-win-3.6.0-alpha-20240624+7a627c1-x64

### Step 7
This mow looks better, correct in fact to me.
image
What I am now seeing is the project in the state in which it was last saved at step 3
a) the label track is not lost
b) the audio and labels are in sync
c) The track is at the 120 Tempo that it was at Step 4 (last save) and not the 180 Tempo that I changed it **_temporarily_** to at Step 5

So this looks fixed to me in Matthieu's latest branch build,

saintmatthieu:
@LWinterberg, following up on @dozzzzer 's concern, could you please check that the latest build is also giving the expected behavior?

src/ProjectManager.cpp
```diff
@@ -530,21 +530,20 @@ void ProjectManager::OnCloseWindow(wxCloseEvent & event)
...
+   // This can reduce reference counts of sample blocks in the project's
+   // tracks. Do this before `CompactProjectOnClose` so that changes after the
+   // last save don't get written to the project file.
+   UndoManager::Get(project).ClearStates();
 
-      // Delete all the tracks to free up memory
-      tracks.Clear();
-   }
+   // Compact the project.
+   projectFileManager.CompactProjectOnClose();
```
vsverchinsky, line 543: In `CompactProjectOnClose` consider `ProjectFileIO::WasCompacted` returns `false`, then it would never execute `ProjectFileIO::UpdateSaved` because undo stack was cleared. I don't know if that was a workaround for some problem, but now it becomes a dead `if`-branch.
Alternatively I could propose to fix `ProjecFileIO::Compact`:
```      
if(!ShouldCompact(tracks))
{
   mWasCompacted = true;
   ...
```
though `ProjectFileIO::SetBypass()` might need changes too.

saintmatthieu: Thanks for this comment, I admit I hadn't examined possible consequences of my change exhaustively.

That dead branch updates the "document blob" (does this include the database?), discarding the unsaved tracks, but doing so it also saves the tempo and time signature changes.

If you change something else not track-related, like the snapping precision in project A, next time you open project B it will have that precision. Something similar happens for tempo, but that's not the desired behaviour, and I guess it's the first thing that's not track-related that must be project-specific.

If that theory is correct, then it'd explain why up to now `projectFileIO.UpdateSaved( mLastSavedTracks.get() );` was ok until time signature was introduced. Follow-up question is whether the change of this commit would have been a correct alternative all this time and this now dead branch can be deleted.

saintmatthieu: In any case, I realize that this commit alone can't be completely correct. `ProjectFileIO::SetBypass()` depends on the execution of `ProjectFileIO::Compact()` (called from `ProjectFileManager::CompactProjectOnClose()`). Just pushed something to fix this.

vsverchinsky (approved):
https://bugzilla.audacityteam.org/show_bug.cgi?id=2529
Worth checking for regression
@dozzzzer @chinakov

saintmatthieu:
Wow, well done doing this research.

saintmatthieu:
I merged this PR on 3.6.2 with a force-push, can be closed.

PR #6750, opened 2024-06-25
Limiter and Compressor presets

### LINKED ISSUE #6320: Compressor and limiter factory presets
Research and add presets to the compressor and limiter effects introduced in https://github.com/audacity/audacity/issues/6305.

saintmatthieu:
Resolves: #6320

Recommended:

QA:
- [ ] Compressor and limiter have factory presets specified in [compressor presets.zip](https://github.com/user-attachments/files/15973161/compressor.presets.zip)
- [ ] Limiter has factory presets specified in [limiter presets.zip](https://github.com/user-attachments/files/15973160/limiter.presets.zip)

### DIFF
--- a/libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
+++ b/libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
@@ -1,0 +1,125 @@
+/*  SPDX-License-Identifier: GPL-2.0-or-later */
+/*!********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  DynamicRangeProcessorUtils.cpp
+
+  Matthieu Hodgkinson
+
+**********************************************************************/
+#include "DynamicRangeProcessorUtils.h"
+#include <optional>
+#include <regex>
+#include <stdexcept>
+#include <unordered_map>
+
+namespace DynamicRangeProcessorUtils
+{
+namespace
+{
+
+template <typename Struct> struct SettingDescription
+{
+   double Struct::*const mem {};
+   const char* const key {};
+};
+
+static const std::vector<SettingDescription<CompressorSettings>>
+   compressorSettingDescriptions {
+      { &CompressorSettings::thresholdDb, "thresholdDb" },
+      { &CompressorSettings::makeupGainDb, "makeupGainDb" },
+      { &CompressorSettings::kneeWidthDb, "kneeWidthDb" },
+      { &CompressorSettings::compressionRatio, "compressionRatio" },
+      { &CompressorSettings::lookaheadMs, "lookaheadMs" },
+      { &CompressorSettings::attackMs, "attackMs" },
+      { &CompressorSettings::releaseMs, "releaseMs" },
+      { &CompressorSettings::showInput, "showInput" },
+      { &CompressorSettings::showOutput, "showOutput" },
+      { &CompressorSettings::showActual, "showActual" },
+      { &CompressorSettings::showTarget, "showTarget" },
+   };
+
+const std::vector<SettingDescription<LimiterSettings>>
+   limiterSettingDescriptions {
+      { &LimiterSettings::thresholdDb, "thresholdDb" },
+      { &LimiterSettings::makeupTargetDb, "makeupTargetDb" },
+      { &LimiterSettings::kneeWidthDb, "kneeWidthDb" },
+      { &LimiterSettings::lookaheadMs, "lookaheadMs" },
+      { &LimiterSettings::releaseMs, "releaseMs" },
+      { &LimiterSettings::showInput, "showInput" },
+      { &LimiterSettings::showOutput, "showOutput" },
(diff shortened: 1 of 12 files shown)


libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
```diff
@@ -0,0 +1,125 @@
...
+   };
+
+template <typename Struct>
+std::optional<Struct> Deserialize(
+   const std::string& str,
+   const std::vector<SettingDescription<Struct>>& settings)
+{
+   Struct settingStruct;
+
+   const std::regex pattern(R"((\w+)=\"(-?\d+\.?\d*)\")");
```
grliszas14, line 63: This regex won't recognise settings like:
`1.23e-4`
or
`123.` (without digit after the decimal)
is that ok? just curious

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
```diff
@@ -0,0 +1,125 @@
...
+   if (std::any_of(
+          settings.begin(), settings.end(), [&values](const auto& setting) {
+             return values.find(setting.key) == values.end();
+          }))
+      return {};
+
+   for (const auto& setting : settings)
+      try
+      {
+         settingStruct.*(setting.mem) = std::stod(values.at(setting.key));
```
grliszas14, line 86: std::stod can throw out_of_range exception which is not handled

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
```diff
@@ -0,0 +1,125 @@
...
+      }
+      catch (const std::invalid_argument&)
+      {
+         return {};
```
grliszas14, line 90: This function returns empty optional if any setting is missing - is that desired? I mean maybe we want to handle partial settings and filling missing ones with default values?

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
```diff
@@ -0,0 +1,125 @@
...
+namespace
+{
+
+template <typename Struct> struct SettingDescription
+{
+   double Struct::*const mem {};
+   const char* const key {};
+};
+
+static const std::vector<SettingDescription<CompressorSettings>>
```
grliszas14, line 28: nitpick: could be a std::array if the size of container is known in advance

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "DynamicRangeProcessorTypes.h"
+#include "Internat.h"
+#include "TranslatableString.h"
+
+namespace DynamicRangeProcessorUtils
+{
+template <typename Struct> struct Preset
+{
+   const TranslatableString name; // Not sure this needs to be translat
```
grliszas14, line 21: typo: "translated" ?

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
saintmatthieu: Right. It's acceptable insofar as we have unit tests that check that all presets are parsed successfully, so if we introduce one with scientific notation, tests will tell us, and it can be introduced then.
Thank you, though, and thanks unit tests.

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
saintmatthieu: Should be caught by the unit tests, but you're right, I'll generalize.

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
saintmatthieu: The typical way of producing presets is to use the editor to do so: then all settings are saved, so partial presets shouldn't be a thing. If for some reason it becomes one we can make the procedure more granular.

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.cpp
saintmatthieu: Yeah, but then you have to count your number of presets and update that size - more human work `:P`. (What you suggest would have been fine too, of course.)

libraries/lib-dynamic-range-processor/DynamicRangeProcessorUtils.h
saintmatthieu: I wanted to get rid of that comment: now I know it should be translatable.

grliszas14 (approved):

Issue #6625, opened 2024-06-06
[AU4] Clips actions: delete


igorkorsukov:
### Description

called from clip context menu

PR #6767, opened 2024-06-26
[AU4] Clip actions: delete

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6625

### DIFF
--- a/au4/src/au3wrap/internal/processinginteraction.cpp
+++ b/au4/src/au3wrap/internal/processinginteraction.cpp
@@ -60,3 +60,50 @@ bool ProcessingInteraction::changeClipTitle(const processing::ClipKey& clipKey, 
 
     return true;
 }
+
+bool ProcessingInteraction::removeClip(const processing::ClipKey& clipKey)
+{
+    WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), TrackId(clipKey.trackId));
+    IF_ASSERT_FAILED(waveTrack) {
+        return false;
+    }
+
+    std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.index);
+    IF_ASSERT_FAILED(clip) {
+        return false;
+    }
+
+    clip->Clear(clip->Start(), clip->End());
+
+    processing::ProcessingProjectPtr prj = globalContext()->currentProcessingProject();
+    prj->onClipRemoved(DomConverter::clip(waveTrack, clip.get(), clipKey.index));
+
+    return true;
+}
+
+bool ProcessingInteraction::removeClipData(const processing::ClipKey& clipKey, double begin, double end)
+{
+    WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), TrackId(clipKey.trackId));
+    IF_ASSERT_FAILED(waveTrack) {
+        return false;
+    }
+
+    std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.index);
+    IF_ASSERT_FAILED(clip) {
+        return false;
+    }
+
+    processing::secs_t initialClipStart = clip->Start();
+    processing::secs_t initialClipEnd = clip->End();
+
+    clip->Clear(begin, end);
+
+    processing::ProcessingProjectPtr prj = globalContext()->currentProcessingProject();
+    if (begin <= initialClipStart && end >= initialClipEnd) {
+        prj->onClipRemoved(DomConverter::clip(waveTrack, clip.get(), clipKey.index));
+    } else {
+        prj->onClipChanged(DomConverter::clip(waveTrack, clip.get(), clipKey.index));
+    }
+
+    return true;
+}
--- a/au4/src/au3wrap/internal/processinginteraction.h
+++ b/au4/src/au3wrap/internal/processinginteraction.h
@@ -16,6 +16,8 @@ public:
 
     bool changeClipStartTime(const processing::ClipKey& clipKey, double sec) override;
     bool changeClipTitle(const processing::ClipKey& clipKey, const muse::String& newTitle) override;
+    bool removeClip(const processing::ClipKey& clipKey) override;
+    bool removeClipData(const processing::ClipKey& clipKey, double begin, double end) override;
 
 private:
     AudacityProject& projectRef() const;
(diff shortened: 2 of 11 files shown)


au4/src/au3wrap/internal/processinginteraction.cpp
```diff
@@ -60,3 +60,23 @@ bool ProcessingInteraction::changeClipTitle(const processing::ClipKey& clipKey,
...
+
+    std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.index);
+    IF_ASSERT_FAILED(clip) {
+        return false;
+    }
+
+    clip->Clear(begin, end);
+
+    processing::ProcessingProjectPtr prj = globalContext()->currentProcessingProject();
+    prj->onClipChanged(DomConverter::clip(waveTrack, clip.get(), clipKey.index));
```
Eism, line 79: I think it would be more correct to use onClipRemoved here to send a notification about the removed element in clipList

au4/src/au3wrap/internal/processinginteraction.h
```diff
@@ -16,6 +16,7 @@ class ProcessingInteraction : public processing::IProcessingInteraction
 
     bool changeClipStartTime(const processing::ClipKey& clipKey, double sec) override;
     bool changeClipTitle(const processing::ClipKey& clipKey, const muse::String& newTitle) override;
+    bool removeClip(const processing::ClipKey& clipKey, double begin, double end) override;
```
igorkorsukov, line 19: Non-obvious signature. I think it's better to have two methods, something like:
* bool removeClip(const processing::ClipKey& clipKey);
* bool removeClipData(const processing::ClipKey& clipKey, double begin, double end);

au4/src/processing/internal/processingactionscontroller.cpp
```diff
@@ -23,6 +31,51 @@ void ProcessingActionsController::notifyActionCheckedChanged(const ActionCode& a
...
+{
+    NOT_IMPLEMENTED;
+}
+
+void ProcessingActionsController::clipCopy()
+{
+    NOT_IMPLEMENTED;
+}
+
+void ProcessingActionsController::clipDelete()
```
igorkorsukov, line 44: I think it's better to have two methods: 
1.
```
void ProcessingActionsController::clipDelete(const ActionData& args)
{
       IF_ASSERT_FAILED(args.count() > 0) {
               return;
       }

      processing::ClipKey clipKey = args.arg<processing::ClipKey >(0);
      ...
}
```
see https://github.com/igorkorsukov/audacity/blob/master/au4/src/projectscene/view/clipsview/clipcontextmenumodel.cpp#L14

2. void ProcessingActionsController::clipDeleteSelected()

au4/src/processing/internal/processingactionscontroller.cpp
```diff
@@ -23,6 +31,51 @@ void ProcessingActionsController::notifyActionCheckedChanged(const ActionCode& a
...
+    auto selectedTracks = selectionController()->dataSelectedOnTracks();
+    auto selectedStartTime = selectionController()->dataSelectedStartTime();
+    auto selectedEndTime = selectionController()->dataSelectedEndTime();
+    auto tracks = project->processingProject()->trackList();
+    auto selectedClipKey = selectionController()->selectedClip();
+
+    //! TODO AU4: improve for deleting multiple selected clips
+    // remove single clip when selected via header click
+    if (selectedClipKey.index != nidx) {
+        auto clip = project->processingProject()->clipList(selectedClipKey.trackId)[selectedClipKey.index];
```
igorkorsukov, line 56: It would be better to add a method like this: Clip clip(const ClipKey& key) const;
(and it is better to use `at` method to access read-only list element)

au4/src/au3wrap/internal/processinginteraction.cpp
```diff
@@ -60,3 +60,43 @@ bool ProcessingInteraction::changeClipTitle(const processing::ClipKey& clipKey,
...
+
+    std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.index);
+    IF_ASSERT_FAILED(clip) {
+        return false;
+    }
+
+    clip->Clear(begin, end);
+
+    processing::ProcessingProjectPtr prj = globalContext()->currentProcessingProject();
+    prj->onClipRemoved(DomConverter::clip(waveTrack, clip.get(), clipKey.index));
```
igorkorsukov, line 99: here we need to find out if the start and end match the start and end of the clip, then we need to notify about the clip being removed, if they do not match (the clip is not removed), then notify about the clip being changed

au4/src/au3wrap/internal/processinginteraction.cpp
```diff
@@ -60,3 +60,50 @@ bool ProcessingInteraction::changeClipTitle(const processing::ClipKey& clipKey,
...
+        return false;
+    }
+
+    std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.index);
+    IF_ASSERT_FAILED(clip) {
+        return false;
+    }
+
+    auto initialClipStart = clip->Start();
+    auto initialClipEnd = clip->End();
```
igorkorsukov, line 97: please do not use auto for simple types
The point here is that the compiler will easily calculate the type, but for reading a code this is unnecessary complexity, we need to know what the function returns... for reading and understanding the code, we need to see what types we are dealing with. For example, in this case, seeing that the type is double, then further in the code it is clear that there may be a problem, doubles are not safe to compare with each other (<= and =>)

au4/src/au3wrap/internal/processinginteraction.cpp
```diff
@@ -60,3 +60,50 @@ bool ProcessingInteraction::changeClipTitle(const processing::ClipKey& clipKey,
...
+
+    clip->Clear(begin, end);
+
+    processing::ProcessingProjectPtr prj = globalContext()->currentProcessingProject();
+    if (begin <= initialClipStart && end >= initialClipEnd) {
```
igorkorsukov, line 102: doubles are not safe to compare with each other 
I suggest two options:
1. Use the `processing::secs_t` type, it is not just a double, but a wrapper over the double, for correct work with it
```
    processing::secs_t initialClipStart = clip->Start();
    processing::secs_t initialClipEnd = clip->End();

...
```

2.  Use functions from `#include "global/realfn.h"`   RealIsEqualOrMore, RealIsEqualOrLess

igorkorsukov (approved):

igorkorsukov (approved):

Issue #6753, opened 2024-06-25
Compressor and limiter presets from Vladislav Isaev


saintmatthieu:
Follow-up ticket to #6750, which introduces some mechanism for easy and safe introduction of new compressor and limiter presets.

@dozzzzer and I agreed that there should be a dedicated PR for each contributor preset submission.

PR #6792, opened 2024-07-01
Presets

LWinterberg:
Resolves: https://github.com/audacity/audacity/issues/6753
Supersedes: https://github.com/audacity/audacity/pull/6754

All presets are in! 
QA: Effects touched are reverb, compressor and limiter


petersampsonaudacity:
Testing on W10 with @LWinterberg 's branch test build for this:  

There is now a whole raft of Factory presets for the new 3.6 compressor:  audacity-win-3.6.0-alpha-20240702+6338b3b-x64
image
These can be selected and appear to work properly as intended - both in destructive mode and RT mode

And a smaller set is available for the new Limiter:
image


---------------------------------------------------------------------------------

Will there be user documentation available (in the manual) on the behavior of these presets?

grliszas14 (approved):

Issue #6801, opened 2024-07-02
Usability of various sliders, including the step size being too small


DavidBailes:
### Bug description

Sliders affected are : playback/recording level sliders, play-at-speed slider, pan and gain sliders, velocity slider.
For example, for playback/recording sliders, left/right arrow keys used to change the value by 1%. This is now .1%, which is not at all useful.
In addition, for the play-at-speed slider:
a) The problem that, using the keyboard, it's not easy to get back to a speed of exactly 1, has come back. See: https://github.com/audacity/audacity/issues/4619
b) The commands Extra > Play at speed > Increase/Decrease playback speed, result in very small changes of speed.

### Steps to reproduce

Case A:
1. Move the focus to the playback level slider.
2. Try changing the level with the left/right arrow keys.

Case B:
1. Move the focus to the play at speed slider.
2. If necessary set its value to 1.000.
3. Keep pressing page up till you get to the maximum value, 3.000
4. Try to set its value to 1.000 again by pressing page down a few times.

Case C:
1. Set the play at speed slider to 1.000.
2. Extra > Play at speed > Increase playback speed.

### Expected behavior

Case A: the level changes by 1% for each left/right arrow key press.

Case B: you can set its value back to 1.000

Case C: the playback speed has been changed by a "reasonable amount"

### Actual behavior

Case A: you have to press the keys many times for a change of 1%

Case B: you can't.

Case C: The playback speed is still 1.000

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

### Additional context

These bugs were introduced by the commit: https://github.com/audacity/audacity/pull/6378

So a regression wrt 3.5.1

DavidBailes:
@dozzzzer . I've submitted a PR to fix these issues: https://github.com/audacity/audacity/pull/6804

Please can this issue be added to 3.6.0?

DavidBailes:
@LWinterberg . I've submitted a PR to fix these issues: https://github.com/audacity/audacity/pull/6804

Please can this issue be added to 3.6.0?

Issue #6704, opened 2024-06-17
Playback speed display on top of already open playback editing box is annoying


nalin243:
### Your idea

No need to have the playback value showing box pop up if the user has double clicked on the playback slider and already has the playback speed box open. 

### Problem to be solved

When we double click on the playback slider a small box opens that allows us to manually input the playback value. On moving the slider in that box another box opens up on top of the already open box showing us the live playback changes which seems a little annoying and redundant. 

### Additional context

[Screencast_20240617_162941.webm](https://github.com/audacity/audacity/assets/29351098/20e8809d-b847-4859-8d08-df6ae87364e2)

SteveDaulton:
> When we double click on the playback slider

Are you referring to the [Play at Speed](https://manual.audacityteam.org/man/play_at_speed_toolbar.html) slider?
If so, that is so that you can set the playback  speed.

nalin243:
@SteveDaulton Yes that is what I was talking about. Did you watch the attached video? I am talking about the additional small box that shows up on top of the already open playback speed editor. We can already see the playback speed changing in the text box, what is the point of another box showing the same thing appearing on top of it?

LWinterberg:
What's happening here is that the tooltip is getting its own window for non-modal dialogs. This appears to be a linux-specific bug, or perhaps even a DE/WM-specific bug.

SteveDaulton:
I don't think it's a  bug Leo. I believe it is an accessibility feature.

Sighted users can drag the slider to the required speed, and the speed is indicated in the tooltip.

Screen reader users enter a double click on the slider widget, that opens a dialog window (as shown in [nalin243](https://github.com/nalin243)'s video), so that they can enter the required playback speed.

LWinterberg:
The tooltip is supposed to look like this: 
![image](https://github.com/audacity/audacity/assets/87814144/13957a25-c0cc-4349-b2b4-8cb1e20c98ea)

It's not supposed to have a separate window titled <2> as shown in OP: 

![image](https://github.com/audacity/audacity/assets/87814144/4e4ea2c7-ac67-4856-869b-fe7a584f5043)

SteveDaulton:
> It's not supposed to have a separate window titled <2> as shown in OP:

Yes, that is a bug. I mistakenly thought that nalin243 was complaining about the pop-up window on double-click.
FWIW, it looks like this on my machine (Ubuntu + XFCE):

[...]

PR #6804, opened 2024-07-03
Usability of various sliders for keyboard users

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6801

A number of bugs were introduced by a PR for changing the play-at-speed accuracy back to 3 decimal places (commit: 37330d6). These bugs are all fixed by reverting some of the unnecessary changes in that PR.

A
Problem: the step size for keyboard users of ASliders is too small (For example the recording and playback level sliders). This is fixed by reverting the constant 100.0 to 10.0 in LWSlider::Increase() and LWSLider::Decrease().

B
Problem: For the play-at-speed slider, using the keyboard, it's not easy to get back to a speed of exactly 1.000. This was caused by the change of the minimum of the speed slider being changed from 0.01 to 0.001, and the dependence of LWSlider::Increase() and LWSlider::Decrease() on that value. So the minimum value of the speed slider was reverted to 0.01 (which I think is quite slow enough).

C
Problem: The commands Extra > Play at speed > Increase/Decrease playback speed, result in very small changes of speed. Fixed by reverting the constant 0.01 to 0.1 in OnPlaySpeedInc() and OnPlaySpeedDec()

Recommended:

### DIFF
--- a/src/toolbars/TranscriptionToolBar.cpp
+++ b/src/toolbars/TranscriptionToolBar.cpp
@@ -1126,7 +1126,7 @@ void OnPlaySpeedInc(const CommandContext &context)
    auto tb = &TranscriptionToolBar::Get( project );
 
    if (tb) {
-      tb->AdjustPlaySpeed(0.01f);
+      tb->AdjustPlaySpeed(0.1f);
    }
 }
 
@@ -1136,7 +1136,7 @@ void OnPlaySpeedDec(const CommandContext &context)
    auto tb = &TranscriptionToolBar::Get( project );
 
    if (tb) {
-      tb->AdjustPlaySpeed(-0.01f);
+      tb->AdjustPlaySpeed(-0.1f);
    }
 }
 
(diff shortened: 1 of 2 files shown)


LWinterberg (approved):

nalin243:
@DavidBailes It is not possible for me to move the slider using keyboard in Linux. I did not even know this was a thing that was possible for play at speed slider. Could you try it out in a linux machine? This may be a DE specific issue like in #6704

DavidBailes:
> @DavidBailes It is not possible for me to move the slider using keyboard in Linux. I did not even know this was a thing that was possible for play at speed slider. Could you try it out in a linux machine? This may be a DE specific issue like in #6704

@nalin243 . I can confirm that it's not possible to use the play-at-speed slider using the keyboard on xUbuntu 22.04. A workaround for keyboard users is Extra > Play-at-Speed > Adjust playback speed, and then use the text box in that dialog.

Issue #6813, opened 2024-07-03
View menu: duplicate access key


DavidBailes:
### Bug description

"Track size" and "Show RMS waveform" both have the access key R.

### Steps to reproduce

Open the View menu using the keyboard so that access keys are shown in the menu.

### Expected behavior

No duplicate access keys

### Actual behavior

"Track size" and "Show RMS waveform" both have the access key R.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

PR #6814, opened 2024-07-04
View menu: duplicate access key.

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6813

Changed the access key of "Show RMS waveform" so that there isn't a duplicate access key.

Recommended:


LWinterberg:
please rebase onto master, 3.6 already is in string freeze

DavidBailes:
Will submit a PR based on master

PR #6817, opened 2024-07-04
[AU4] Record. Show recorded clips

Eism:


au4/src/au3wrap/internal/au3record.cpp
```diff
@@ -31,6 +38,57 @@ using namespace au::au3;
...
+namespace au::au3 {
+class RecordingListener : public AudioIOListener
+{
+public:
+    RecordingListener()
+    {
+        m_timer.setSingleShot(true);
+        m_timer.setInterval(100);
+
+        QObject::connect(&m_timer, &QTimer::timeout, [this]() {
```
igorkorsukov, line 50: I think better to use muse::Timer (don't use Qt here)

igorkorsukov (approved):

igorkorsukov (approved):

PR #6824, opened 2024-07-05
Update Italian translations for v3.6

mlocati:

Recommended:

### DIFF
--- a/locale/it.po
+++ b/locale/it.po
@@ -16,7 +16,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: audacity 3.6.0\n"
 "Report-Msgid-Bugs-To: audacity-translation@lists.sourceforge.net\n"
-"POT-Creation-Date: 2024-07-02 21:05+0300\n"
+"POT-Creation-Date: 2024-07-05 09:02+0200\n"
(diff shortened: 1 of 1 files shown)


LWinterberg (approved):

mlocati:
PS: I maintain the Italian translations [on Transifex](https://app.transifex.com/klyok/audacity).

The only problem is that the .po files downloaded from Transifex has many (useless) changes when compared to the .po files (re)generated by the `locale/update_po_files.sh` script.

On Posix systems you can avoid these useless changes by running `update_po_files.sh`.

On Windows that's not possible, even with WSL: since `.gitattributes` forces checking out the source .c files with the Windows line endings, the gettext tools don't work (they don't accept strings containing `\r\n`).

So, in order to avoid these useless changes I created [this tiny project](https://github.com/mlocati/audacity-fix-po).

Here's the process I've used to generate this PR:

1. I checked the Audacity repository locally
2. I downloaded the .po file from Transifex to the `locale` directory
3. I committed the changed file
4. I "normalized" the file by running
   ```sh
   docker run --rm -it -v C:\path\to\audacity:/app ghcr.io/mlocati/audacity-fix-po it
   ```

Issue #6781, opened 2024-06-27
[AU4] Fix timelineruller Arithmetic exception


igorkorsukov:
### Description

Signal name : | SIGFPE
Signal meaning : | Arithmetic exception

occurs at very high zoom

![Image](https://github.com/audacity/audacity/assets/3818029/01756615-8fc3-4c5e-974a-e283c0fb4b22)

PR #6840, opened 2024-07-09
[AU4] Apply time signatures to timeline ruler

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6837
Resolves: https://github.com/audacity/audacity/issues/6781

### DIFF
--- a/au4/src/projectscene/view/timeline/timelineruler.cpp
+++ b/au4/src/projectscene/view/timeline/timelineruler.cpp
@@ -112,21 +112,20 @@ Ticks TimelineRuler::prepareTickData(const IntervalInfo& timeInterval, double w,
 
         QString tickLabel = m_formatter->label(value, timeInterval, tickType, m_context);
         int labelsCount = 0;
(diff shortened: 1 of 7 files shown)


au4/src/projectscene/view/timeline/timelineruler.cpp
```diff
@@ -101,31 +101,31 @@ Ticks TimelineRuler::prepareTickData(const IntervalInfo& timeInterval, double w,
     {
         // determine tick type
         TickType tickType;
-        if (tickNumber % static_cast<int>(timeInterval.major / timeInterval.minorMinor) == 0) {
+        double eps = 1.0e-5f;
```
Eism, line 104: better use std::numeric_limits<double>::epsilon()

grliszas14: Yup, I thought that too but I tried that and it's to small... so I'm using the same value as AU3 does

Eism (approved):

PR #6852, opened 2024-07-11
disable RMS by default

LWinterberg:


kryksyh (approved):

dozzzzer:
The RMS in the waveform is now off by default. 
However, if it's ON and there's a clip in the project, when I reset the config, the option goes back to OFF, but the waveform is not repainted and still shows the RMS. 

Is it desired? @LWinterberg @Tantacrul

LWinterberg:
It's not desired, but I'm fairly sure it's also not introduced by this PR.

Issue #6674, opened 2024-06-12
Only the first channel of files with more than 2 channels is shown


AbsenceJam:
### Bug description

[audacity-macOS-3.6.0-alpha-20240612+c5051c4-arm64](https://nightly.link/audacity/audacity/workflows/build/master/audacity-macOS-3.6.0-alpha-20240612%2Bc5051c4-arm64)
macOS Sonoma 14.5
ffmpeg 6 *or* 7 from Homebrew

When opening any of these (all I've tested so far):
DD 3.0, DD 4.0, DD 5.1, DD+ 5.1, FLAC 5.1, DTS-HD MA 5.1
Only the first channel is shown. DD 2.0 is okay.

Apologies if this is some sort of settings change!

### Steps to reproduce

- Clean install of Audacity
- Make sure ffmpeg is working
- Open one of the aforementioned formats

### Expected behavior

Each channel should be a seperate track as in 3.5.1.

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS, Windows, Linux

### Additional context

![3 5 1 - DD 3 0](https://github.com/audacity/audacity/assets/24455272/86394383-36a2-4659-b59b-c4ae2de68a8f)
![3 5 1 - DD 5 1](https://github.com/audacity/audacity/assets/24455272/0bd97519-d436-4e1f-8094-efe5642b7501)
![3 5 1 - FLAC 5 1](https://github.com/audacity/audacity/assets/24455272/b3b3d728-1b18-44cf-b054-f0e104580536)
![3 6 0 - DD 3 0](https://github.com/audacity/audacity/assets/24455272/74505795-4246-4cdb-9023-5b8bb15b3809)
![3 6 0 - DD 5 1](https://github.com/audacity/audacity/assets/24455272/889f38fb-9611-462e-92ae-bf377321310f)
![3 6 0 - FLAC 5 1](https://github.com/audacity/audacity/assets/24455272/b6fe4897-8ae3-4649-8c64-272fe1d1fef3)

dozzzzer:
I've managed to reproduce this on macOS 14.5 and Windows 11.

dozzzzer:
Bisects to https://github.com/audacity/audacity/commit/983dc1b7c7fd5008fd40420fefe6b63d68cd6a77 @Paul-Licameli

dozzzzer:
[...]

PR #6853, opened 2024-07-11
Fix incorrect multichannel audio import (5.1, 7.1 etc)

kryksyh:
Resolves: [#6674](https://github.com/audacity/audacity/issues/6674)

Recommended:


libraries/lib-import-export/ImportUtils.h
```diff
@@ -34,12 +34,18 @@ class IMPORT_EXPORT_API ImportUtils final
    //! Choose appropriate format, which will not be narrower than the specified one
    static sampleFormat ChooseFormat(sampleFormat effectiveFormat);
    
-   //! Builds a wave track and places it into a track list.
+   //! Builds a wave track
```
vsverchinsky, line 37: 👍

modules/import-export/mod-pcm/ImportPCM.cpp
```diff
@@ -289,15 +289,15 @@ void PCMImportFileHandle::Import(
 
    wxASSERT(mFile.get());
 
-   auto track = ImportUtils::NewWaveTrack(
+   auto trackList = ImportUtils::NewWaveTracks(
```
vsverchinsky, line 292: `ImportRAW.cpp` directly invokes factory's `CreateMany`, would be nice to keep it same in both.

modules/import-export/mod-pcm/ImportPCM.cpp
```diff
@@ -386,7 +389,10 @@ void PCMImportFileHandle::Import(
       return;
    }
 
-   ImportUtils::FinalizeImport(outTracks, *track);
```
vsverchinsky, line 389: `ImportUtils::FinalizeImport(TrackHolders&, TrackList &&)`

vsverchinsky (approved):
Approved with few comments.
@LWinterberg should it be merged into `release-3.6.0`?

modules/import-export/mod-pcm/ImportPCM.cpp
```diff
@@ -353,27 +353,30 @@ void PCMImportFileHandle::Import(
 
          if (block) {
             unsigned c = 0;
-            ImportUtils::ForEachChannel(*track, [&](auto& channel)
```
vsverchinsky, line 356: There is a ImportUtils::ForEachChannel(TrackList&, const std::function<void(WaveChannel&)>&) overload

LWinterberg:
> should it be merged into release-3.6.0?

I wouldn't mind if it was, the question is if @dozzzzer @chinakov still have bandwidth to test it before release

dozzzzer:
Yeah, let's do that.

vsverchinsky (approved):
Can you please squash it and merge it with Reabase and merge after QA completes?

LWinterberg:
Does the QA result block merge? I think improving the situation for FLAC (even if it doesn't address the ffmpeg formats) would be worthwhile

Issue #6664, opened 2024-06-11
[AU4] Add a warning that the project is not saved and offer to save


igorkorsukov:
### Description

Same as MuseScore

PR #6854, opened 2024-07-11
[AU4] Ask for project save on close

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6664

### DIFF
--- a/au4/src/au3wrap/CMakeLists.txt
+++ b/au4/src/au3wrap/CMakeLists.txt
@@ -31,6 +31,8 @@ set(MODULE_SRC
 
     ${CMAKE_CURRENT_LIST_DIR}/mocks/au3settingsmock.cpp
     ${CMAKE_CURRENT_LIST_DIR}/mocks/au3settingsmock.h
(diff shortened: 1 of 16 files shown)


libraries/lib-project-file-io/DBConnection.cpp
```diff
@@ -268,7 +268,9 @@ bool DBConnection::Close()
       {
          using namespace std::chrono;
          std::this_thread::sleep_for(50ms);
-         pd->Pulse();
+         if (pd != nullptr) {
+             pd->Pulse();
```
vsverchinsky, line 272: There is a `wxASSERT(pd)` few lines above. 
Why not make a stub implementation of `BasicUI`? There are lots of similar places where it's used

grliszas14: Good idea, implemented empty QtBasicUI stub, no AU3 dependecies in current changes, please have a look

igorkorsukov (approved):

Issue #6800, opened 2024-07-02
[AU4] Snapping tool


DilsonsPickles:
This spec details the full behaviour of the snapping tool. Some behaviours described may fall outside of the scope of "Main toolbar".

[Figma Link](https://www.figma.com/design/i2A4pyP2E6VE2JiBFRIysK/Main-toolbar?node-id=1795-4)

PR #6861, opened 2024-07-12
[AU4] Playback toolbar. Snapping

Eism:
Resolves: #6800

### DIFF
--- a/au4/src/playback/CMakeLists.txt
+++ b/au4/src/playback/CMakeLists.txt
@@ -68,7 +68,8 @@ set(MODULE_SRC
 
     )
 
-set(LINK_LIB
+set(MODULE_LINK
+    projectscene
     record
 )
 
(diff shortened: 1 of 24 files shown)


au4/src/playback/CMakeLists.txt
```diff
@@ -68,7 +68,8 @@ set(MODULE_SRC
 
     )
 
-set(LINK_LIB
+set(MODULE_LINK
+    projectscene
```
igorkorsukov, line 72: This raises doubts, it seems that playback should not depend on the projectscene.
Options:
* Move the playtoolbar to the projectscene (favorite)
* Move this component to a common place, maybe something like commonscene, or even to uicomponents.

I'm not calling for this to be fixed right now, but it's technical debt. If it's not fixed, there will be a cyclical dependency.

igorkorsukov (approved):

au4/src/playback/CMakeLists.txt
Eism: https://github.com/audacity/audacity/issues/6872

Issue #6536, opened 2024-05-29
Playhead jumping back and forth when both the track and the master channel have an effect with latency


dozzzzer:
### Steps to reproduce

1. grab some audio and start playback
2. add [Voxengo Latency Delay](https://www.voxengo.com/product/latencydelay/) to the track's effect stack
3. observe: playhead moves forward as expected
4. add Voxengo Latency Delay to the Master channel
5. observe

### Expected behavior

Playhead keeps moving forward down the timeline smoothly

### Actual behavior

Playhead starts to "jump" back and forth as it moves down the timeline; disabling the Master channel's effect doesn't help.

https://github.com/audacity/audacity/assets/70063970/e45cd18d-106b-4f85-b1b1-95781253618c

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

dozzzzer:
Tested. Works like a charm!

Issue #6844, opened 2024-07-10
playback cursor can jump back and forth when using the compressor effect


DavidBailes:
### Bug description

If the compressor effect is added to either a track or master, and the lookahead is in the range roughly 60-130ms when playback is started, the playhead jumps back and forth during playback.

### Steps to reproduce

1. Add the compressor effect to master.
2. Choose the Beefy master preset.
3. Start playback

### Expected behavior

playback cursor moves smoothly

### Actual behavior

playback cursor jumps back and forth

### Audacity Version

latest beta

### Operating system

Windows

### Additional context

See also: https://github.com/audacity/audacity/issues/6536

PR #6865, opened 2024-07-12
Attempt to fix jumping playhead

vsverchinsky:
Resolves: #6536, #6844

With large latency values `TimeQueue` buffer values could have become overwritten. The issue is temporary fixed by buffer reallocation in case if overwrite is detected.

QA:
Please also check scrubbing and seek.

Recommended:


kryksyh (approved):
lgtm

Issue #4763, opened 2023-06-02
Plain gui of MuseFX effects: for some parameters, not all values are available.


DavidBailes:
### Bug description

I've only checked the Reverb, Pitchfix, and De-Ess plugins, and all had similar problems.

### Steps to reproduce

1. Install the Muse PitchFix effect.
2. Disable the graphical interface.
3. Try changing the values of the 4 parameters: preset, amount, scale, and quality.

### Expected behavior

You should be able to set the value of each parameter to any of the values available in the standard graphical user interface.

### Actual behavior

Preset, scale, and quality could each only be set to two different values - far fewer than in the graphical user interface.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Checked the MuseFX  pitchfix and reverb plugins in Reaper, and in the plugin's plain user interface, consisting of sliders, all the values of the parameters are available.

DavidBailes:
In void VST3ParametersWindow::OnParameterValueChanged(const wxCommandEvent& evt),
in the line:
const auto normalizedValue = control->GetNormalizedValue(*mEditController);
When using the pitchfix plugin, and changing the value of any of Preset, scale, or quality
the values of normalizedValue are not in the range 0.0 to 1.0, but have the values 0.0, 1.0, 2.0, 3.0 etc

DavidBailes:
@LWinterberg . Given that "get more effects" in Audacity 3.6. is going to point to [musehub.com](https://www.musehub.com/), rather than https://plugins.audacityteam.org/,  I think that it's a reasonable expectation that the MuseFX effects are usable by users of screen readers, which requires this bug to be fixed.

DavidBailes:
Some other effects also have similar issues for their discrete parameters. For example, Tokyo Dawn labs TDR Kotenlnikov: the parameters SC HP Slope, Mode, and GR Disp Scale.
However other effects don't have these issue with discrete parameters. For example iZotope Vinyl: Warp model, year, and RPM.
Reaper detects the problem, and works around it - see, eg: https://forums.cockos.com/showpost.php?p=2194665&postcount=3

DavidBailes:
@LWinterberg . I've submitted a PR to fix this. Please can this be added to 3.6.1, if there is such a release, as the MuseFX effects are currently unusable by keyboard users, and the "get more effects" links in Audacity take the user to musehub, which provides these effects?

dozzzzer:
[...]

PR #6876, opened 2024-07-15
VST3 Effects: fix problem with discrete parameters in plain UI.

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/4763

For some VST3 effects, for discrete parameters in the plain UI for keyboard users, it's not possible to set all the possible values.

It appears that the problem is that some plugins don't provide the correct implementation of the functions for converting between plain and normalized values.

Fix:
Detect when invalid functions for these conversions are provided, and when this is the case, use the formula use the formula given by Steinberg in their documentation. (Form posts on Reaper's forums, it appears that Reaper does something similar to this.)

Resolves: https://github.com/audacity/audacity/issues/4763

Recommended:

### DIFF
--- a/src/effects/VST3/VST3ParametersWindow.cpp
+++ b/src/effects/VST3/VST3ParametersWindow.cpp
@@ -157,6 +157,8 @@ namespace
    {
       const wxString mTitle;
       const wxString mUnits;
+      const int mStepCount{};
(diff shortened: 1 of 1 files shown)


DavidBailes:
Closing, and will issue another PR based on release-3.6.1

Issue #3679, opened 2022-09-23
Labeled audio operations no longer work on selected audio tracks in 3.x


JC3:
### Bug description

When any audio tracks are selected, all labeled audio operations (e.g. delete, split cut, etc.) become a no-op (although they're still added to the undo stack, oddly). Additionally, even when *all* audio tracks are selected, the operations do not work.

The only way to perform any labeled audio operations is to select the label track, ensure no audio tracks are selected, and then apply the operation. The action will be performed on *every* audio track.

This is new behavior present in at least 3.1.3 and 3.2.0 (did not test earlier 3.x), but not in 2.4.2 or prior. 

In 2.4.2 and prior, if specific audio tracks were selected, labeled audio operations were applied only to the selected audio tracks.

As a consequence, it is no longer possible to perform labeled audio operations on specific individual audio tracks. It's either all or none.

Also, possibly related, sometimes the labeled audio menu items are disabled even when a label track is selected, although I'm unable to reproduce that consistently.

### Steps to reproduce

1. In a project with multiple audio tracks and a label track, select the label track and any (or all) audio tracks.
2. Edit -> Labeled Audio -> Pick anything.

### Expected behavior

Operation is applied on the selected audio tracks, using the selected label track.

From 2.4.2:

![242opt](https://user-images.githubusercontent.com/5651184/192015532-30d6010a-fce4-4041-8a2d-2cc04ca4f50a.gif)

### Actual behavior

Nothing happens unless no audio tracks are selected.

From 3.2.0:

https://user-images.githubusercontent.com/5651184/192015616-7afce3fe-da5a-46d4-8eec-948111e11349.gif

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows, macOS

JC3:
Note: In my opinion this one is a slightly higher than medium priority; since this is an operation I perform a lot, I'm actually just using 2.4.2 when I need to do a lot of label operations on projects with many tracks for now.

petersampsonaudacity:
Testing on W10 shows that this is a regression on 2.4.2, introduced in 3.0.0

_Given that 3.0.0 was released 18 months ago and this is the first report of this I suspect that not many folk use this._

JC3:
> Testing on W10 shows that this is a regression on 2.4.2, introduced in 3.0.0
> 
> _Given that 3.0.0 was released 18 months ago and this is the first report of this I suspect that not many folk use this._

[...]

PR #6883, opened 2024-07-16
Fix for labeled audio operations don't work in certain cases

kryksyh:
Resolves: [#3679](https://github.com/audacity/audacity/issues/3679)

The intended behavior is:
- When SyncLock is enabled, all sync-locked tracks are edited regardless of their selection state if their respective Label track is selected.
- When SyncLock is disabled and playable tracks are selected, only the selected playable tracks are edited.
- When SyncLock is disabled, and no playable tracks are selected, all tracks are edited.

Here is a sample project to play with: [labels.zip](https://github.com/user-attachments/files/16253354/labels.zip)

Recommended:

### DIFF
--- a/src/menus/LabelMenus.cpp
+++ b/src/menus/LabelMenus.cpp
@@ -626,7 +626,12 @@ void OnJoinLabels(const CommandContext &context)
 
    auto editfunc = [&](Track& track, double t0, double t1, ProgressReporter reportProgress) {
       track.TypeSwitch(
-         [&](WaveTrack& t) { t.Join(t0, t1, std::move(reportProgress)); });
+         [&](WaveTrack& t) {
+            // Extend Join Region by one sample, to ensure
+            // that Split-Join operation on a given label is reversible and consistent
+            const auto sampleTime = 1.0 / t.GetRate();
+            t.Join(t0 - sampleTime, t1 + sampleTime, std::move(reportProgress));
+         });
    };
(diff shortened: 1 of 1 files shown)


src/menus/LabelMenus.cpp
```diff
@@ -237,9 +249,25 @@ void EditClipboardByLabel(AudacityProject &project,
    if (regions.empty())
       return;
 
-   const bool notLocked = (!SyncLockState::Get(project).IsSyncLocked() &&
-                           (tracks.Selected<PlayableTrack>()).empty());
+   const bool syncLocked = SyncLockState::Get(project).IsSyncLocked();
+   const bool playableTrackSelected = !tracks.Selected<PlayableTrack>().empty();
 
+   const auto tracksToEdit = tracks.Any<Track>() + [&](const auto pTrack) {
+      // All sync-locked tracks should be edited,
```
vsverchinsky, line 256: Predicate could be extracted into predicate builder function to avoid code duplication in the same file

vsverchinsky (approved):
Approved with a single suggestion to eliminate code duplication

src/menus/LabelMenus.cpp
kryksyh: makes sense, done

vsverchinsky (approved):

vsverchinsky (approved):

vsverchinsky (approved):

PR #6899, opened 2024-07-17
Added uniqueIds to WaveClips

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6896

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -35,6 +35,8 @@
 
 const char *WaveClip::WaveClip_tag = "waveclip";
 
+std::atomic<size_t> WaveClip::sNextId = 0;
+
 WaveClipListener::~WaveClipListener() = default;
 
 void WaveClipListener::WriteXMLAttributes(XMLWriter &) const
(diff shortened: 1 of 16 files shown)


libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1737,6 +1747,11 @@ double WaveClip::SnapToTrackSample(double t) const noexcept
    return std::round(t * mRate) / mRate;
 }
 
+size_t WaveClip::GetNextId()
+{
+    return sNextId++;
```
igorkorsukov, line 1752: Maybe we should make 0 - not a valid id (we can't do that with indexes, but we can with ids).
And then you need to change it to `return ++sNextId;`

libraries/lib-wave-track/WaveClip.h
```diff
@@ -976,6 +980,8 @@ class WAVE_TRACK_API WaveClip final :
    bool mIsPlaceholder { false };
 
    wxString mName;
+   size_t mUniqueId = static_cast<size_t>(-1);
```
igorkorsukov, line 983: see https://github.com/audacity/audacity/pull/6899/files#r1680808776

igorkorsukov (approved):

igorkorsukov:
reimplemented by https://github.com/audacity/audacity/pull/7123

Issue #6892, opened 2024-07-17
UI Artifacts on Button Corners in Dark Theme


CennoxX:
### Bug description

There are white points at the corners of some buttons in the Audacity UI. These artifacts can be seen in the attached screenshot. This issue affects the visual consistency and appearance of the interface.

### Steps to reproduce

1. open Audacity
2. activate dark mode

### Expected behavior

visual consistent appearance without white artifacts

### Actual behavior

![artifacts](https://github.com/user-attachments/assets/5197cb06-2d1e-4c37-9c8b-a2f3cdc1d3f3)

### Audacity Version

Audacity 3.6.0 (from audacityteam.org/download)

### Operating system

Windows 10

dozzzzer:
Tested. I no longer see the original issue:

<img width="600" alt="image" src="https://github.com/user-attachments/assets/82395cd8-8b18-4287-8654-6990449e3a9e">

PR #6914, opened 2024-07-18
Fix button artifacts on the main toolbar and effects panel

kryksyh:
Resolves: https://github.com/audacity/audacity/issues/6892

### Main toolbar
image

### Main toolbar
image
image

### Effects panel
image

image

Recommended:


src/widgets/AButton.cpp
```diff
@@ -173,6 +174,7 @@ void AButton::SetFrameMid(int mid)
 void AButton::Init(wxWindow* parent, wxWindowID id, const wxPoint& pos, const wxSize& size, bool toggle)
 {
    SetBackgroundStyle(wxBG_STYLE_PAINT);
+   SetBackgroundColour(theTheme.Colour(clrMedium));
```
vsverchinsky, line 177: There are few places that may require backgroud colour override as they use different one: `RealtimeEffectsPanel.cpp` (opened via Effects button on track panel, then add effect) and `EffectUI.cpp`(open effect ui by clicking on the button with effect name)
Did you check them?

vsverchinsky (approved):

src/widgets/AButton.cpp
kryksyh: >Did you check them?
I have not, I need to fix those, thanks!
image

kryksyh: @vsverchinsky  Should look better now, please check

vsverchinsky: изображение
Click an effect name to open effect UI. I get this under Windows platform

kryksyh: @vsverchinsky right, missed that one with the dark theme enabled. Should be fixed now.

Why do effect dialogs not use the selected theme though?

vsverchinsky: 👍 
Don't know for sure, but probably it was something that was missed during redesign or there might have been some problems with that

vsverchinsky (approved):

dozzzzer:
Tested. I no longer see the original issue:

image

CennoxX:
I also tested the fix and can confirm the issue is solved.

Issue #6751, opened 2024-06-25
TCP UI Breaks when dragging waveform splitter


chinakov:
### Bug description

If the User creates a stereo track and collapses it, then choosing to adjust the waveform via the waveform splitter, the TCP track menu UI will break.

### Steps to reproduce

1. Create Stereo Track
2. Add some audio
3. Collapse the track
4. Try to drag the waveform splitter
5. Observe result	

### Expected behavior

User should be able to use the waveform splitter to change the view of a stereo tracks' waveform

### Actual behavior

https://github.com/audacity/audacity/assets/139755141/0efa85f4-f0c5-4e05-b40d-27a50e01259e

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS, Linux

petersampsonaudacity:
Testing on W10 shows that this is a regression on 3.5.1

petersampsonaudacity:
Testing on W10 with @saintmatthieu 's branch build for this.

This now appears to work properly in Matthieu's branch build - attempting to drag the waveform splitter:
a) does not change the collapsed waveform - _this is expected behavior_
b) the collapsed TCP remains properly displayed

chinakov:
Tested, thanks!

PR #6925, opened 2024-07-19
Fix for TCP UI Breaks when dragging waveform splitter

saintmatthieu:
Resolves: #6751 

I think the perfect fix would be to fail the hit test when the track is minimized and the cursor is above the left/right channel horizontal line, but I couldn't figure out how to make this work, and I think this simple fix does the trick UX-wise.

Recommended:


petersampsonaudacity:
Testing on W10 with @saintmatthieu 's branch build for this.

This now appears to work properly in Matthieu's branch build - attempting to drag the waveform splitter:
a) does not change the collapsed waveform - _this is expected behavior_
b) the collapsed TCP remains properly displayed

vsverchinsky (approved):

Issue #6929, opened 2024-07-20
Imported audio is now auto-selected, thereby destroying any prior selection that the user had


petersampsonaudacity:
### Bug description
In 3.6.0 and 3.6.1 test builds - when you import an audio file audacity now selects that imported audio - and of course in so doing this replaces and thus destroys any selection that the user may have carefully made.

This is a regression on 3.5.1 and all earlier Audacity versions.

Reported by two users on the Forum:
https://forum.audacityteam.org/t/audacity-automatically-does-select-all-after-importing-file/112215
https://forum.audacityteam.org/t/audacity-3-6-is-out-now/111894/42

### Steps to reproduce
1. Import an audio file
2. Observe: the track is auto-selected
3. make a small selection on the track
4. Import a second audio track
5. Observe: this new track is selected, thus removing the selection made at step 3

### Expected behavior
Imported audio should **_not_** be auto-selected

### Actual behavior
Imported audio is auto-selected - removing any prior extant selection

### Audacity Version
3.6.0 and latest 3.6.1 test build

### Operating system
Windows 10 - but assume all OS

### Additional context
None

DavidBailes:
@LWinterberg . I've created a PR to fix this, please can this be added to 3.6.1?

DavidBailes:
> @LWinterberg . I've created a PR to fix this, please can this be added to 3.6.1?

I'm too late!

dozzzzer:
3.6.2 it is then :)

dozzzzer:
@DavidBailes feel free to open a PR once again.

petersampsonaudacity:
Testing on GitHub with @DavidBailes 's test branch build for this:  audacity-win-3.6.1-alpha-20240722+2594629-x64

a) importing a file does not get that audio audio selected

b) **BUT** - when you import a file the time selection that was originally present gets transferred to the newly imported track:
![image](https://github.com/user-attachments/assets/b197bb29-ff9a-4404-8753-507048b0b32b)
And applying an effect show that the selection is indeed in the new track and this is **not** just a visual/display error

petersampsonaudacity:
David Bailes wrote in the PR thread:
> This is the same behaviour as 3.5.1 and previous versions - the last imported track becomes selected. This PR does not change that, it only removes the change in the time selection.

Yes, I just tested that.

BUT that looks erroneous to me.  

@LWinterberg @dozzzzer 
[...]

PR #6933, opened 2024-07-22
Importing audio files: do not change time selection

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6929

Problem:
In 3.6.0, when importing audio files, the time selection is set to cover the audio in the last imported file. This is a change wrt previous versions, can means that users loose any previously set time selection. It was introduced by the commit: 4532f561db

Fix:
Remove the setting of the time selection when importing audio files.

Recommended:


DavidBailes:
3.6.1 has already been released.

Issue #1725, opened 2021-09-23
New audio in project to be selected automatically


SteveDaulton:
**Is your feature request related to a problem? Please describe.**
Audacity is inconsistent in selecting (or not) newly added audio.
1. When running a macro on "files", the imported track is automatically selected.
2. When manually importing a file, the imported track is not selected.
3. When generating audio, the new audio is selected.
4. When recording audio, the new audio is not selected

**Describe the solution you'd like**
New audio in a project to be automatically selected (consistently).

**Describe alternatives you've considered**
Living with the current mishmash.

**Additional context**
We receive complaints on the forum about effects throwing an error when the user has forgotten to select the track.

Previously (years ago), Audacity had a default behaviour of selecting the entire project if an effect was run without a selection, but this was highly problematic as in many cases the user did not want to apply the action to the entire project. In the worst case, a user might panic because they accidentally pressed the Del key and the entire project "disappeared".

I believe the main source of complaints comes from the situation where new audio has been added (typically from importing) and the user intuitively expects to be able to immediately apply an effect to it, but they can't because it is not selected.

petersampsonaudacity:
I do not support this issue as it stands.

1) Recording
>4. When recording audio, the new audio is not selected

This then becomes an inconsistency

2) Mix &Render
for consistency, Mix&Render to New Track would need to atuto-select the new track after the mix.
_Note that this is not the case currently, rather the selection remains with the tracks selected for the mix._

Accordingly my counter-proposal is that we **remove** the auto-selection of 
>3. When generating audio, the new audio is selected.

This is the least work - and provides great consistency.

petersampsonaudacity:
@SteveDaulton 's point1
>1. When running a macro on "files", the imported track is automatically selected.

Kind of makes sense as otherwise the user has to add a Select command to the Macro - but perhaps that is no bad thing - and it's not exactly arduous or tricky to do when you are writing a Macro.

So for absolute consistency in the handling of new audio I would also suggest that Audacity does **NOT** auto-select  imported audio when running a Macro on files.

SteveDaulton:
> 1. Recording
> This then becomes an inconsistency

No, I'm suggesting that newly added audio should always (consistently) be selected. It should be selected whether it was added by importing, or generating, or recording, or any other way.

> 2. Mix &Render
Currently we have an inconsistency here too.
After "Mix and Render", the new track is selected, but after "Mix and Render to New  Track" it isn't.

[...]

PR #6934, opened 2024-07-22
Importing audio files: do not change time selection

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/6929

Problem:
In 3.6.0, when importing audio files, the time selection is set to cover the audio in the last imported file. This is a change wrt previous versions, can means that users loose any previously set time selection. It was introduced by the commit: 4532f561db

Fix:
Remove the setting of the time selection when importing audio files.

Recommended:

### DIFF
--- a/src/ProjectFileManager.cpp
+++ b/src/ProjectFileManager.cpp
@@ -1356,7 +1356,7 @@ bool ProjectFileManager::Import(wxArrayString fileNames, bool addToHistory)
           addToHistory))
       return false;
    // Last track in the project is the one that was just added. Use it for
-   // focus, selection, etc.
+   // focus, etc.
    Track* lastTrack = nullptr;
    const auto range = TrackList::Get(mProject).Any<Track>();
    assert(!range.empty());
@@ -1373,8 +1373,6 @@ bool ProjectFileManager::Import(wxArrayString fileNames, bool addToHistory)
       viewPort.ZoomFitHorizontally();
       viewPort.ShowTrack(*track);
       viewPort.HandleResize(); // Adjust scrollers for NEW track sizes.
-      ViewInfo::Get(*project).selectedRegion.setTimes(
-         track->GetStartTime(), track->GetEndTime());
    });
    return true;
 }


saintmatthieu:
Thank you @DavidBailes for jumping on this!
I'd like to have @LWinterberg 's opinion, though. I think the intention really was to put focus and selection on the imported material.
In 3.5, the behavior was like below:
image
image
The usefulness of keeping the selection interval untouched but applying it to the newly imported track is dubious to me.

DavidBailes:
> The usefulness of keeping the selection interval untouched but applying it to the newly imported track is dubious to me.

A more common case would probably be that initially, there is no time selection, just a cursor position. Why should importing a track loose that cursor position, and instead have a time selection covering all the audio in the imported track? If a user wants to select all the audio in the import track they can easily do so, rather than it being done automatically for them, whether or not they want it.

LWinterberg:
> Why should importing a track loose that cursor position, and instead have a time selection covering all the audio in the imported track?

Because we have that horrible feature bug of giving people error messages when they have no audio selected and go to applying any effect. Automatically selecting audio would be a way around it. (automatically doing a clip selection instead would be better, but we don't have that concept yet). See also #1725

DavidBailes:
> > Why should importing a track loose that cursor position, and instead have a time selection covering all the audio in the imported track?
> 
> Because we have that horrible feature bug of giving people error messages when they have no audio selected and go to applying any effect. Automatically selecting audio would be a way around it. (automatically doing a clip selection instead would be better, but we don't have that concept yet).

But as you've already pointed out in the issue you linked to (see below), the alternative to automatically selecting audio when it's imported/recorded/etc, is to have a way of intelligently selecting audio is none is selected. This has the advantage that it doesn't mess up users who know what they are doing, see eg the forum post which raised the issue: https://forum.audacityteam.org/t/audacity-3-6-is-out-now/111894/42?u=david_bailes

> See also #1725

In that issue, here's your last comment:
<quote>
I feel like the solution here would be to

1. always apply effects to the clip the edit cursor currently is in, as opposed to the entire project, when nothing is selected, and
2. not select anything upon creating audio
</quote>

and I think that's roughly the the right way to go. I don't think that the right way to go is to select all created audio, eg recording, import, etc.

petersampsonaudacity:
Testing on GitHub with @DavidBailes 's test branch build for this:  audacity-win-3.6.1-alpha-20240722+2594629-x64

a) importing a file does not get that audio audio selected

b) **BUT** - when you import a file the time selection that was originally present gets transferred to the newly imported track:
image
And applying an effect show that the selection is indeed in the new track and this is **not** just a visual/display error

DavidBailes:
> b) **BUT** - when you import a file the time selection that was originally present gets transferred to the newly imported track

This is the same behaviour as 3.5.1 and previous versions - the last imported track becomes selected. This PR does not change that, it only removes the change in the time selection.

petersampsonaudacity:
> This is the same behaviour as 3.5.1 and previous versions - the last imported track becomes selected. This PR does not change that, it only removes the change in the time selection.

Yes, I just tested that.

BUT that looks erroneous to me.  

@LWinterberg @dozzzzer 
Do we need a new issue for the time selection being moved to the newly imported tack?  I suspect that we do,

LWinterberg:
I'd like to await a bit more feedback before committing to reverting this change. The previous behavior doesn't make much sense to me, the current one does (even though it breaks some folks' workflows).

DavidBailes:
> I'd like to await a bit more feedback before committing to reverting this change. The previous behavior doesn't make much sense to me, the current one does (even though it breaks some folks' workflows).

@LWinterberg . Concerning breaking this previous workflow. You set the cursor position based on the audio in the existing tracks, you import some audio, and then time shift the audio in the new track it so that it starts at that cursor position. This workflow is possible in 3.5.1, and in all(?) previous versions.
In Version 4.0, then I presume that this workflow will be possible, because it will be possible to select the clip "properly" and so have the audio selected, AND leave the cursor position unchanged.
So why break this workflow in 3.6.x, only for it to be OK again in 4.0?

LWinterberg (approved):

PR #6955, opened 2024-07-23
Fix out of bound access in waveform cache

kryksyh:
Resolves: on a certain combination of wxWidgets versions and OS, waveform painting was corrupting the memory and resulted in a white screen.

Recommended:

### DIFF
--- a/libraries/lib-wave-track-paint/GraphicsDataCache.cpp
+++ b/libraries/lib-wave-track-paint/GraphicsDataCache.cpp
@@ -134,7 +134,7 @@ GraphicsDataCacheBase::PerformBaseLookup(
    const int64_t width = right - left;
 
    const int64_t cacheLeft       = left / CacheElementWidth;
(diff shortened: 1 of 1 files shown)


vsverchinsky (approved):

saintmatthieu (approved):

wchouser3:
this patch seems to have fixed the problem. I wonder how long before we see it show up in Arch default repo

ChandigarIT:
No sign of 3.6.2 when I try to upgrade

root@****:~# apt install audacity
audacity is already the newest version (3.6.1+dfsg-1)

kryksyh:
@kryksyh Unfortunately, 3.6.2 is not released yet, but it will be pretty soon.

Antony-Edward-Stark:
thanks for fixing white screen issue. Thank you...
I was ust saved by your timely release.

Issue #6736, opened 2024-06-23
Time Track Range setting would be much better as a single dialog


petersampsonaudacity:
### Bug description
In the **Range** setting for a Time track (accessed via the tracks Track Control Panel context menu, or its Vertical Scale context menu)  the user is presented with not one but two similar dialogs presented one after the other with very similar input.  One for **Lower speed limit**, the other for **Upper speed limit**.

It would be much better if these were consolidated  into a single dialog, as is the case elsewhere in Audacity, giving consistency.

### Steps to reproduce
1. Tracks > Add New > Time Track
2. right click on the track name
3. in the context menu select range
4. Observe: two similar dialogs, one after the other

### Expected behavior
![image](https://github.com/audacity/audacity/assets/50205984/acb3fc8a-f715-4635-96d3-e7552949a4e1)

Apart from any other consideration it would be jolly useful to see both limits together.

### Actual behavior
![image](https://github.com/audacity/audacity/assets/50205984/6a50eca2-e1ac-4f9f-8810-885f8569e968)

followed by

![image](https://github.com/audacity/audacity/assets/50205984/05b8704c-2b49-41e3-bd7c-dc1e003e63fb)

### Audacity Version
3.5.1 (and earlier Audacity versions) - and the latest alpha master test build for 3.6.0

### Operating system
Windows 10, but assume all OS

### Additional context
A single dialog would not only be neater and more consistent - it would involve fewer clicks and thus a boon for RSI sufferers.

nalin243:
@petersampsonaudacity 
@LWinterberg 

It seems the reason it was done this way is because the minimum value of the upper limit must always be one greater than the current lower limit.

![dialog](https://github.com/user-attachments/assets/43b37ddf-ec62-41b8-a9ef-e728c14ad8e6)

I was able to create a dialog just fine, but now if the lower limit is greater than the upper limit then it just silently fails. Should there be an error message that pops up and if so what should be it's contents?

petersampsonaudacity:
@nalin243 
well what is does now with the two dialogs is that if you try to set a value for  the upper limit that is less than the lower limit, then it silently changes the upper limit to the lower limit plus 1, with no error message or warning shown to the user.

But yes, I would have thought that if you try to set an invalid limit  then there should be an error message and the action should be blocked.  

[...]

PR #6956, opened 2024-07-24
Merged lower limit and upper limit dialog boxes in time track controls

nalin243:
Resolves: #6736

- Used wxDialogWrapper the Shuttle gui system to create a new dialog box with both the lower speed limit spinctrl and upper speed limit spinctrl in one place.
- Used AudacityMessageBox to show an error message and block the user from setting invalid limits.

Recommended:

### DIFF
--- a/src/tracks/timetrack/ui/TimeTrackControls.cpp
+++ b/src/tracks/timetrack/ui/TimeTrackControls.cpp
@@ -115,18 +115,19 @@ void TimeTrackMenuTable::OnSetTimeTrackRange(wxCommandEvent & /*event*/)
    if (dlg.ShowModal() == wxID_CANCEL)
       return;
 
-   lower = scLower->GetValue();
-   upper = scUpper->GetValue();
-
-   if(lower >= upper) {
+   while(scLower->GetValue() >= scUpper->GetValue()) {
       AudacityMessageBox(
          XO("Upper Speed Limit must be greater than the Lower Speed Limit"),
          XO("Invalid Limits"),
          wxOK | wxICON_ERROR,
          mpData->pParent);
 
-      return;
+      if (dlg.ShowModal() == wxID_CANCEL)
+         return;
    }
+
+   lower = scLower->GetValue();
+   upper = scUpper->GetValue();
 
    if (lower >= TimeTrackControls::kRangeMin &&
        upper <= TimeTrackControls::kRangeMax) {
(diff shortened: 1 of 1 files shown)


nalin243:
@dozzzzer 

Done.

kryksyh:
@ nalin243 looks good to me.
I'm wondering if allowing the user to correct the input if they selected the incorrect range makes sense? 
For example, we could show the dialog again so the user can correct the input, instead of forcing them to use context menu once more.

kryksyh (approved):
Approved.

But from a user experience perspective, I would prefer that the dialog does not close in case of incorrect input, but instead prompts the user to enter the data again. 

@LWinterberg , what do you think?

nalin243:
@kryksyh 

I had thought of doing something like that but as the dialog is just two clicks away I didn't. Should I make a commit now or wait until @LWinterberg responds?

petersampsonaudacity:
Tested on W10 with @nalin243 's latest test branch for this:  audacity-win-3.6.2-alpha-20240724+cef9925-x64

a)  The dialog box now has normal Audacity buttons/controls:
image


b) The error message is much improved:
image

petersampsonaudacity:
> But from a user experience perspective, I would prefer that the dialog does not close in case of incorrect input, but instead prompts the user to enter the data again.

I am minded to agree with this - but I am not the decision-maker ...

nalin243:
@kryksyh

I am not sure how to fit a validator in here. Also like @petersampsonaudacity suggested, the error should only be shown when the dialog is submitted via 'Ok'. What I can do is something like this. 

[Screencast_20240724_165357.webm](https://github.com/user-attachments/assets/34fc9392-1915-4c3a-9d97-0bf8be5e96c6)

If this is fine then I can commit.

nalin243:
@kryksyh @petersampsonaudacity 

> And note that you may also try to change the lower limit to be higher than the upper limit (as well as the other way around). However that may be seen as correct behavior as the user may want to make both limits much higher (or lower. So any error message should probably not be triggered by changing either of the limits - but should be triggered by the user pressing the OK button when invalid limits are in place in the input fields.

If the user presses Ok button then the dialog will close. So in this case the dialog will close the error message will pop up and then once the user presses Ok button for error message the dialog will reappear. 

[Screencast_20240724_165357.webm](https://github.com/user-attachments/assets/ce767113-91ca-4a2a-b083-d8f53857f6f3)

Is this fine?

kryksyh:
>  Screencast_20240724_165357.webm 
> Is this fine?
@nalin243 looks good to me

PR #6969, opened 2024-07-25
Envelope display fix: pass absolute time value when querying envelope.

### LINKED ISSUE #6927: Incorrect waveform appearance when Envelope applied to clip offset from zero
### Bug description

When an Envelope is applied to a clip that starts after zero, the affect on the waveform is incorrectly offset.

### Steps to reproduce

1. Generate a 10 second tone
2. Apply a non-destructive fade-in to the first 2 seconds using the Envelope  tool.
3. Slide the clip to the right so that it starts at time = 5 seconds

### Expected behavior

The fade is expected to remain in the same place relative to the clip.

### Actual behavior

The fade appears to have stretched to the right, even though the Envelope control points are still in the correct places relative to the clip.

Image copied from bug report on the forum:

67de7fdf763258eb8e582db34478f8f830646966

### Audacity Version

3.6.0 / Audacity 3.7.0-alpha

### Operating system

Linux, Windows, MacSO

### Additional context

Originally reported on the forum: https://forum.audacityteam.org/t/envelope-help-what-am-i-doing-wrong/112247
Confirmed on Linux.

This is a serious bug that also affects existing projects created in previous versions.

saintmatthieu:
Resolves: #6927

Recommended:

QA:
- [x] #6927 is fixed
- [x] Also test that stretching clips still shows and plays the correct envelope

### DIFF
--- a/libraries/lib-wave-track-paint/waveform/WaveBitmapCache.cpp
+++ b/libraries/lib-wave-track-paint/waveform/WaveBitmapCache.cpp
@@ -155,7 +155,8 @@ struct WaveBitmapCache::LookupHelper final
       {
          envelope->GetValues(
             EnvelopeValues.data(), static_cast<int>(EnvelopeValues.size()),
-            key.FirstSample / cache->GetScaledSampleRate(),
+            key.FirstSample / cache->GetScaledSampleRate() +
+               envelope->GetOffset(),
             1.0 / key.PixelsPerSecond);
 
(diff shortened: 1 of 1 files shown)


kryksyh (approved):

petersampsonaudacity:
Testing on W10 with @saintmatthieu 's test branch build for this.

This appears to work properly again in Matthieu's branch build.

Issue #6572, opened 2024-06-03
[AU4] Track Panel. Make VolumePressureMeter resizable


igorkorsukov:
### Description

As the track height increases, the VolumePressureMeter should also increases.

Issue #6768, opened 2024-06-26
[AU4] Track - Resizing tracks


DilsonsPickles:
Allowing Users to resize tracks

[Figma Link](https://www.figma.com/design/k5rS9tnZl60nYmhJhvI2vf/Track-headers?node-id=280-46695)

dozzzzer:
@kryksyh 

1. When hovering over the track panel's edges, the cursor doesn't change to a proper type:
AU3: 
<img width="231" alt="image" src="https://github.com/user-attachments/assets/74072101-45de-4015-a09b-9260a9da9732">

2. When the track's height exceeds the window's height, no vertical scroll bar appears.

https://github.com/user-attachments/assets/6338893e-bdf8-4c0e-ae25-1b17ef03f300

3. When in the collapsed state, the track panel should have the Mute and Solo buttons over here:

<img width="401" alt="image" src="https://github.com/user-attachments/assets/8a76a465-c521-4b0f-8f67-cdac25add350">

4. When the track is collapsed, observe how the title bars of some clips stick and some don't as I hover over them. I assume the correct behavior here would be to show the title bar as long as the cursor hovers over the clip and to hide it as soon as it moves away, am I right @DilsonsPickles ? Or do we need to show title bar on each clip at all times?

https://github.com/user-attachments/assets/ece08881-32ea-4374-86d4-7fbadb4eb3c9

kryksyh:
@dozzzzer @chinakov 

1 & 3 should be fixed in the latest build

2. Can't reproduce, does this always happen to you?

4. Is not related to this PR, I would rather not to mix changes for separate modules in a single PR. Could you create a separate tasks for it?

chinakov:
1. When hovering

chinakov:
Also, regarding the number 4 on the list - in the latest build of the attached PR the issue is not reproducible for me. It seems that it's behaving correctly. 

https://github.com/user-attachments/assets/66944944-740c-42c4-80f9-55fb8a9a6d9f

kryksyh:
@chinakov this may be present only on macos, it does happen to me

dozzzzer:
@kryksyh Item no. 1 is still reproducible in the latest MacOS build:

Here I demonstrate it (observe how the cursor is flickering and for a fraction of a second it does change to the double-headed arrow but then reverts to the pointer one):

https://github.com/user-attachments/assets/4c369814-faf8-46c8-9550-ed4b413b0f68

[...]

PR #6984, opened 2024-07-29
[AU4] Make track panel volume indicator responsive

kryksyh:
Resolves: [#6572](https://github.com/audacity/audacity/issues/6572)

- Make the indicator and the ruler responsive to the track panel height change
- Adjust styling according to the design spec

Recommended:

### DIFF
--- a/au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
+++ b/au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
@@ -61,7 +61,7 @@ ListItemBlank {
 
         ColumnLayout {
             Layout.margins: 12
-
+            Layout.alignment: Qt.AlignTop
             spacing: 12
 
             RowLayout {
@@ -174,9 +174,11 @@ ListItemBlank {
         SeparatorLine {}
 
         Row {
-            Layout.fillHeight: true
+            Layout.alignment: Qt.AlignTop
             Layout.preferredWidth: childrenRect.width
-            Layout.margins: 8
+            Layout.margins: 5
+            Layout.bottomMargin: 20
+            Layout.preferredHeight: root.height
 
             spacing: 2
 
--- a/au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/VolumePressureMeter.qml
+++ b/au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/VolumePressureMeter.qml
@@ -9,16 +9,23 @@ import Muse.Ui 1.0
 Canvas {
     id: root
 
+    enum Style {
+        Solid,
+        Gradient
+    }
+
     property real currentVolumePressure: -60.0
     property real minDisplayedVolumePressure: -60.0
     property real maxDisplayedVolumePressure: 0.0
 
     property bool showRuler: false
+    property int style: VolumePressureMeter.Style.Solid
+    property color meterColor: "#7689E6" // TODO: Use the track color
 
     property bool isClipping: currentVolumePressure >= maxDisplayedVolumePressure
 
     width: root.showRuler ? prv.indicatorWidth + 20 : prv.indicatorWidth
-    height: prv.indicatorHeight + (prv.overloadHeight * 2)
+    height: parent.height
 
     QtObject {
         id: prv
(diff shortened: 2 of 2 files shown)


au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
```diff
@@ -174,9 +174,11 @@ ListItemBlank {
         SeparatorLine {}
 
         Row {
-            Layout.fillHeight: true
+            Layout.alignment: Qt.AlignTop
             Layout.preferredWidth: childrenRect.width
-            Layout.margins: 8
+            Layout.margins: 5
+            Layout.bottomMargin: 20
+            height: root.height
```
Eism, line 181: it is the same as `Layout.fillHeight: true`, isn't it?

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/VolumePressureMeter.qml
```diff
@@ -9,29 +9,36 @@ import Muse.Ui 1.0
 Canvas {
     id: root
 
+    enum Style {
+        Solid,
+        Gradient
```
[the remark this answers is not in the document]
kryksyh: I reached out @DilsonsPickles and he suggested to keep it for now

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/VolumePressureMeter.qml
```diff
@@ -60,43 +67,117 @@ Canvas {
...
+    }
 
-            } else {
-                ctx.fillStyle = prv.longStrokeColor
-                ctx.fillRect(currentStrokeVPos,
-                             originHPos,
-                             prv.longStrokeHeight,
-                             prv.longStrokeWidth)
+    // Draws a rectangle rounded at the top, bottom or both
+    function drawRoundedRect(ctx, fillStyle, x, y, width, height, radius, roundedEdge) {
```
grliszas14, line 97: Sounds like a function we'd like to re-use somewhere probably, helper qml module would be nice maybe? But that's just something to think of, not necessarily now

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/VolumePressureMeter.qml
```diff
@@ -117,30 +196,45 @@ Canvas {
...
+        const meterHeight = prv.heightPerUnit * (volumePressure - root.minDisplayedVolumePressure)
 
-        ctx.fillStyle = prv.gradient
-        ctx.fillRect(prv.overloadHeight, 0, prv.divisionPixels * (prv.fullValueRangeLength - Math.abs(root.currentVolumePressure)), prv.indicatorWidth)
+        if (meterHeight > 0) {
+            // Drawing the volume pressure
+            drawRoundedRect(ctx, getMeterFillStyle(ctx),
+                            0, root.height - 10 - meterHeight,
+                            prv.indicatorWidth, meterHeight,
+                            /* radus */ 2, /* rounded edge */ "bottom")
```
grliszas14, line 219: nitpick: typo

grliszas14 (approved):

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/VolumePressureMeter.qml
kryksyh: I would do that if you could guide me on where to put it

Eism: Usually we have an approach that if it is used in more than two places, only then we move it to a separate place, so we need to wait
separate place - we have `UiComponents/Utils.js` for such tasks

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
Eism: Okay, I hope this will be fixed within #6768 and we won't need the clip at all
By the way, please replace `height` with `Layout.preferedHeight` - if the object is inside the layout, then the anchors and sizes should be set via the Layout

kryksyh: Done

Eism (approved):

PR #6992, opened 2024-07-30
6982 restore legacy compressor

### LINKED ISSUE #6982: Create new Legacy section for effects
### Your idea

Add a Legacy effects category containing 

* Classic Filters
* Legacy Compressor
* Legacy Limiter

### Problem to be solved

Currently, we remove obsolete effects, or ship them around disabled by default in the case of Classic Filters. in the case of Nyquist effects, this isn't catastrophic as they are just plugins, however, in the case of builtins, there's not really a way to use any effect we discontinue.

### Prior art

_No response_

### Additional context

_No response_

saintmatthieu:
Resolves: #6982 

Recommended:

QA:
- [ ] Effect menu has a "Legacy" section containing "Legacy Compressor", "Legacy Limiter" and "Classic Filters"


saintmatthieu:
@LWinterberg [322f210](https://github.com/audacity/audacity/pull/6992/commits/322f21065ac5978a9274550d1fafb6977dc691eb) relies on the fact that when launching a freshly installed Audacity version, `AppData\Roaming\audacity\pluginregistry.cfg` is reset. I haven't found the mechanism in the code that proves this, but a quick experiment has shown this to be the case. If you could confirm that'd be great. Otherwise it'd mean that users would only see the classic filters if they've enabled it manually or they've somehow reset their config.

LWinterberg (approved):

Issue #6924, opened 2024-07-19
"Failed to decompress the Cloud project block" when opening a Cloud project in Audacity


dozzzzer:
### Steps to reproduce

1. Generate chirp, time-stretch it
2. Save the project to the Cloud
3. Make a selection inside the clip
4.  `Ctrl (Cmd) + L` to silence selected audio
5. `Ctrl (Cmd) + S` to save the new version
6. Close the project
7. Delete the project from the local temp directory `CloudProjects`
8. In Audacity, `File > Open from Cloud`
9. Select the Cloud project that has just been created

### Expected behavior

The project is loaded

### Actual behavior

MacOS/Windows: 
<img width="420" alt="image" src="https://github.com/user-attachments/assets/e87b7064-d14a-4e50-b617-e59d89a14577">

Linux: Audacity crashes. So does MacOS build sometimes btw.

https://github.com/user-attachments/assets/66d2d96a-6e9e-4a2d-85ad-e7883d655a6c

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows, macOS, Linux

### Additional context

If the clip has no silence clip inside, the bug won't be reproduced.

dozzzzer:
Tested

PR #6993, opened 2024-07-30
Silenced blocks Cloud fix

vsverchinsky:
Resolves: #6924

QA: attempt to open project that was created in previous version should also succeed

Recommended:

### DIFF
--- a/libraries/lib-cloud-audiocom/sync/WavPackCompressor.cpp
+++ b/libraries/lib-cloud-audiocom/sync/WavPackCompressor.cpp
@@ -198,7 +198,10 @@ struct Importer final
    bool IsValid() const
    {
       return Context != nullptr && SamplesCount > 0 &&
-             Format != undefinedSample && BlockId >= 0;
+             Format != undefinedSample &&
+             //The negative BlockId means silenced blocks and it's absolute value encodes it's length
+             //Older projects used to store them on the cloud as regular blocks.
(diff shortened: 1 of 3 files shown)


kryksyh (approved):

libraries/lib-cloud-audiocom/sync/WavPackCompressor.cpp
```diff
@@ -198,7 +198,11 @@ struct Importer final
    bool IsValid() const
    {
       return Context != nullptr && SamplesCount > 0 &&
-             Format != undefinedSample && BlockId >= 0;
+             Format != undefinedSample &&
+             //The negative BlockId means silenced blocks and it's absolute value encodes it's length
+             //Older projects used to store them on the cloud as regular blocks.
+             ((BlockId >= 0 && SamplesCount > 0) ||
```
saintmatthieu, line 204: `SamplesCount > 0` is already checked line 200.

saintmatthieu (approved):

libraries/lib-cloud-audiocom/sync/WavPackCompressor.cpp
vsverchinsky: Thanks!

vsverchinsky:
Removed `SamplesCount` double check

Issue #7008, opened 2024-08-01
Replacing the word "insane" in the Quality drop-down menu in the Export Audio dialog with another word


carotteatomique:
### Your idea

The word "insane" has ableist roots, and even if it is (unfortunately) commonly used in English to create an exaggeration, it's still offensive to many disabled and neurodivergent people, because its ableist connotations are still there.

Replacing this word in the drop-down menu would be a very easy way to not make disabled and neurodivergent people uncomfortable while they use Audacity. This is also a good opportunity to express creativity by picking a word of exaggeration that's more fun!

Some personal suggestions: preposterous, unbelievable, outlandish, ludicrous, unconscionable…

All of these words get across the same idea that the audio quality is incredibly good, without using ableist language!

### Problem to be solved

![The Audacity Export Audio window, the drop-down menu for quality has the options Medium, Standard, Extreme, and Insane](https://github.com/user-attachments/assets/658dd4c2-2215-4fdd-8a0e-427746244e7d)

### Additional context

Please read this if you think this is a non-issue: [BBC article about the impacts of ableist language](https://www.bbc.com/worklife/article/20210330-the-harmful-ableist-language-you-unknowingly-use)

LWinterberg:
Yeah, this is an artifact of early 2000s LAME preset naming. What needs to get across here is that

* Standard is transparent for most content
* Extreme (LAME -V0) is the highest VBR encoding and transparent for all content
* Insane (LAME -B320) maxes out the bits MP3 is allowed to use for no quality gain compared to Extreme

I wonder if Wasteful would be a good replacement for Insane here.

petersampsonaudacity:
Curiously we have had the existing "Insane" and "Extreme" terminology in the MP3 Export dialog since version 1.3.3, which was released on 18May2007.

And no-one has ever mentioned this before ...

@LWinterberg wrote:
> I wonder if Wasteful would be a good replacement for Insane here.

I am minded to agree with Leo here - but I would suggest that in place of Insane we use **Profligate** or **Excessive**, with a minor preference for the former (my "Mystery-Shopper likes "Profligate" too)


**Use Case**
My son, _who is a bit of a purist_, insisted on using 320 when he was building his iTunes library and was surprised to find his iPod filled up well before he had completed the library!

LWinterberg:
I like Excessive. It's the more common word, so non-native English speakers (including probably all of our translators) are more likely to have a good idea what we mean with that.

[...]

PR #7009, opened 2024-08-01
changed mp3 quality UI to friendlier wording

BigDataStructure:
Resolves: *https://github.com/audacity/audacity/issues/7008*

First ever commit to an open-source project so apologies in advance.
This issue seemed pretty simple so I tried my hand at it.

Chose these words specifically because of FL Studio's naming conventions.

- Changed the Insane and Extreme strings in `ExportMP3.cpp`  to Highest and High, respectively
- Left the backend PRESET_INSANE untouched
- Did the same to every occurrence inside language files
- Did NOT change the words in other languages

Recommended:

### DIFF
--- a/au4/CMakeLists.txt
+++ b/au4/CMakeLists.txt
@@ -24,7 +24,7 @@ include(FetchContent)
 FetchContent_Declare(
   muse_framework
   GIT_REPOSITORY https://github.com/musescore/framework_tmp.git
-  GIT_TAG        u0722
+  GIT_TAG        u0801_2
 )
 
 FetchContent_GetProperties(muse_framework)
--- a/au4/src/app/app.cpp
+++ b/au4/src/app/app.cpp
@@ -131,8 +131,8 @@ int App::run(int argc, char** argv)
     }
 
     QCoreApplication::setApplicationName(appName);
-    QCoreApplication::setOrganizationName("MuseScore");
-    QCoreApplication::setOrganizationDomain("musescore.org");
+    QCoreApplication::setOrganizationName("Audacity");
+    QCoreApplication::setOrganizationDomain("audacityteam.org");
     // QCoreApplication::setApplicationVersion(QString::fromStdString(MUVersion::fullVersion().toStdString()));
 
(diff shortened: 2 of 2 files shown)


modules/import-export/mod-mp3/ExportMP3.cpp
```diff
@@ -181,17 +181,15 @@ static const TranslatableStrings varModeNames {
 };
 */
 static const TranslatableStrings setRateNames {
-   /* i18n-hint: Slightly humorous - as in use an insane precision with MP3.*/
-   XO("Insane, 320 kbps"),
-   XO("Extreme, 220-260 kbps"),
+   XO("Highest, 320 kbps"),
+   XO("High, 220-260 kbps"),
```
LWinterberg, line 185: I disapprove of calling these "highest" and "high". MP3 becomes transparent (=a listener cannot reliably differentiate between the source and the encoded content) at around 192 kbps, depending on content of course. So for most content, "Standard" actually is the highest quality while giving you the smallest file sizes, and everything above that is just wasteful.

LWinterberg (changes requested):

petersampsonaudacity:
Curiously we have had the existing "Insane" and "Extreme" terminology in the MP3 Export dialog since version 1.3.3, which was released on 18May2007.

And no-one has ever mentioned this before ...

petersampsonaudacity:
@BigDataStructure 

In the Issue that spawned this PR #7008  a discussion ensued on revised wording.

The outcome was tha @LWinterberg  (responsible for Audacity design) favors **Excessive** in place of **Insane**, but **Extreme** should remain as is.  

And I am strongly minded to agree with him.

LWinterberg:
yeah. Also, just as a note: When doing these changes, it's best to just update the source string in the file itself (in this case, ExportMP3.cpp). That way, the translations will get flagged as in need of updating the next time the .po files get regenerated. The way you did it here would cause all translations to still use the "Insane" translations.

BigDataStructure:
Not sure if I took the most accurate and correct steps, but I force-pushed the audacity branch onto my fork. 

New request: #7014

Thanks again.

PR #722, opened 2021-01-15
2582 enh repeat last process

JohnColket:
# Pull Requests

If you are submitting a pull request, please read https://wiki.audacityteam.org/wiki/GitHub_Pull_Requests

PR #7017, opened 2024-08-02
6954 plugin manager and do effect method usable in au4

### LINKED ISSUE #6954: Make PluginManager and DoEffect usable in AU4
### Description

Make the AU3 effect backend reusable in AU3. This does not cover the UI work, but should facilitate it, as the PluginManager would now provide a list of plugin (effects, command, etc) descriptors according to their availability.

saintmatthieu:
Resolves: #6954 

The scope of this PR is to verify that the Audacity 3 effect framework can be re-used in Audacity 4 for destructive effects (probably paving the way for realtime effects, too).

The `PluginManager` is central in that regard, as it takes care of built-in effects, plugins (VST and such), generators, analyzers and macro commands (and possibly more), detecting them and making all relevant information accessible through its API.

On top of that, the `DoEffect` function encapsulates the Audacity 3 application logic resulting from a user applying a destructive effect on some selection: querying the data from the wave track selections, invoking the UI if needed, driving the effect whose ID was given as input, writing back the result, remembering what effect was last used, etc.

This PR also provides the AU4-specific implementation for a dummy compressor, demonstrating how a developer may get started importing an Audacity 3 effect to Audacity 4.

Commit detail :
* [Non-wxWidget-based standard paths (windows and mac only for now)](https://github.com/audacity/audacity/pull/7017/commits/fe0510fc51eb3a667027863790fe5cf181d2f261) : The plugin manager currently relies on a wxWidgets app to run to find paths such as the executable path. This commit adds platform-dependent code that does this without the the need for a wxWidgets app.
* [DoEffect reusable by Au4](https://github.com/audacity/audacity/pull/7017/commits/d05e7f745945086d20f24e9efa3c205ff8d77988) : `DoEffect` had some other dependencies on wxWidgets, sometimes indirect ones (on the `EffectManager` or the `AudacityCommand`) which had to be abstracted away :
  * The `EffectManager` contains some of the application logic of what happens when Audacity users apply an effect,
  * The `AudacityCommand`s are an abstraction of the well-known macro commands, which allow a user to execute with one click a series of operations (e.g. select all tracks, apply noise reduction and export to a wav file all at once). This abstraction allows the plugin manager to manage these commands like other effect types.
* [CompressorInstance part of lib-audacity](https://github.com/audacity/audacity/pull/7017/commits/e23800663c5f9688addcc65ae542fcdc27ba3274) : an example of the code that Au4 needs to compile for a given effect to be found by the PluginManager. Now in Au4, the plugin manager can find at least one effect, and we can use it to process audio. Such work will have to be done one a per-effect basis.
* [Au3WrapModule initializes PluginManager](https://github.com/audacity/audacity/pull/7017/commits/243542c54ec1c2da7d84051152fd11c72b2e26c9) : does just what it says
* [au4: Selection change updates ViewInfo and TrackList selection](https://github.com/audacity/audacity/pull/7017/commits/5c0756a38c9d8c39d63a00dad2b68ab166535cc8) : without this, `DoEffect` thinks no audio is selected and then nothing happens.

What this PR does not do is connecting the Audacity 4 effects menu to usage of the `DoEffect` function. Another PoC branch based on this one exists with a [hack commit](https://github.com/audacity/audacity/commit/234dbc6f14d6cc5905f69e9bcb5f7475d0ab14af) that does this quick-and-dirty to verify that things work as expected.

Recommended:

### QA list:
This is a refactoring PR, an no behavior whatsoever should have changed.
The changes are simple (mostly renaming and moving code around) but rather widespread ; while testing please don't hesitate to choose test cases that seem related to the list below, even if only vaguely.
- [x] Please all test cases you have on Macros
- [x] Applying generators, analyzers, tools or destructive effects still works.
- [x] When successfully applying an effect, tool, analyzer or generator, the effect menu displays correctly "Repeat <...>"
- [x] Editing a clip's name still works


au4/src/au3wrap/au3wrapmodule.cpp
```diff
@@ -54,8 +56,12 @@ void Au3WrapModule::onInit(const muse::IApplication::RunMode&)
     m_wxLog = new WxLogWrap();
     wxLog::SetActiveTarget(m_wxLog);
 
-    std::unique_ptr<Au3SettingsMock> auset = std::make_unique<Au3SettingsMock>();
-    InitPreferences(std::move(auset));
+    InitPreferences(
+       std::make_unique<Au3Settings>(FileNames::Configuration().ToStdString()));
```
igorkorsukov, line 60: We talked about this a little bit. I believe that an application should have one settings file. This code makes it so that there are two settings files, one that is created by the muse framework (it stores, for example, theme settings) and the other is created here.

libraries/lib-audacity/EffectManager.h
```diff
@@ -64,23 +73,28 @@ class AUDACITY_DLL_API EffectManager
...
+   // public methods
+   //
+   // Used by the outside program to register the list of effects and retrieve
+   // them by index number, usually when the user selects one from a menu.
+   //
 public:
    EffectManager();
    virtual ~EffectManager();
 
+   using DialogInvoker = std::function<bool(
```
saintmatthieu, line 93: Obsolete - will remove.

au4/src/au3wrap/au3wrapmodule.cpp
saintmatthieu: After SU a few days ago we had a follow-up discussion with @Eism and @vpereverzev and IIUC there was a consensus to trying to use the Au3 library as a black box with its own persistency mechanism and making sure that this library exposes the necessary read/write from it.
But I can conceive that one mechanism for all settings for may make things better on the middle term. Only I haven't tried it and seen how the PluginManager reacts to it.
I guess we can experiment with multiple setting files for now and maybe change in the future? The code I had to write for that isn't much, throwing it away if necessary wouldn't be a big loss. I wouldn't block this PR for that, though.

igorkorsukov: I only have concerns about the main config file (I have no concerns about the plugin config files).
When we move on, and the au3 code disappears completely or almost disappears, the user will still have 2 main config files - this will look very strange, and some effort will need to be made to ensure backward compatibility. 
There will also be problems directly when rewriting the code. For example, now the selected device can store au3, then we rewrite this functionality to the new code (without wx), but so that the user does not loss anything, we will still need to store (or at least read) from the au3 settings file. 

In fact, I have already implemented an adapter to the muse::settings, since we seemed to have discussed this yesterday.
https://github.com/audacity/audacity/pull/7019

saintmatthieu: I see : indeed, changing the setting implementation of the general config is even beyond the scope of this PR. Sorry that I mixed things up, will revert that line as soon as I can verify that my PoC still runs without it. (Having troubles launch the app at the moment.)

libraries/CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ set( LIBRARIES
    lib-sqlite-helpers
    lib-preference-pages
    lib-dynamic-range-processor
-   lib-audacity
+   lib-audacity-application-logic
```
vsverchinsky, line 75: I like previous name more :)

vsverchinsky (commented):
One suggestion. I looked into `EffectsManager` implemenation and to me it seem that actually there are two interfaces combined: one for effects and the second one for commands. Attempt to extract `IAudacityCommand` results in massive changes that are unlikely useful for Audacity 4 progress (but if so, what I suggest does not prevent us from future refactroing).
Alternative solution is to define abstract command interface inside EffectManager:
```
class EffectsManager {
public:

class BasicCommandsInterface
{
public:

   virtual bool HasCommand(const PluginID& id) = 0;
   virtual bool Do(const PluginID& id) = 0;
   virtual wxString GetURL(const PluginID& id) = 0;
   virtual wxString GetDescription(const PluginID& id) = 0;
   ...
};

struct PREFERENCES_API CommandsInterface final : GlobalHook<CommandsInterface, BasicCommandsInterface&()> { };

...
};

...
//Inject commands interface
static EffectsManager::CommandsInterface::Scope audacityCommandsInterfaceScope {
   []{ 
       static AudacityWxCommandsInterface commandsInterface;
       return commandsInterface;
      }
};

```

I expect that that would require very few trivial changes in `EffectsManager`

saintmatthieu:
> One suggestion. I looked into `EffectsManager` implemenation and to me it seem that actually there are two interfaces combined: one for effects and the second one for commands. Attempt to extract `IAudacityCommand` results in massive changes that are unlikely useful for Audacity 4 progress (but if so, what I suggest does not prevent us from future refactroing). Alternative solution is to define abstract command interface inside EffectManager:
> 
> ```
> class EffectsManager {
> public:
> 
> class BasicCommandsInterface
> {
> public:
> 
>    virtual bool HasCommand(const PluginID& id) = 0;
>    virtual bool Do(const PluginID& id) = 0;
>    virtual wxString GetURL(const PluginID& id) = 0;
>    virtual wxString GetDescription(const PluginID& id) = 0;
>    ...
> };
> 
> struct PREFERENCES_API CommandsInterface final : GlobalHook<CommandsInterface, BasicCommandsInterface&()> { };
> 
> ...
> };
> 
> ...
> //Inject commands interface
> static EffectsManager::CommandsInterface::Scope audacityCommandsInterfaceScope {
>    []{ 
>        static AudacityWxCommandsInterface commandsInterface;
>        return commandsInterface;
>       }
> };
> ```
> 
> I expect that that would require very few trivial changes in `EffectsManager`

I don't understand:
`BasicCommandsInterface` is the equivalent to my propose `IAudacityCommand`, right ?
Then you propose this global `AudacityWxCommandsInterface` ; but there are many implementations of `AudacityCommand` and these aren't instantiated as singletons.

libraries/lib-files/CMakeLists.txt
```diff
@@ -8,6 +8,14 @@ for temporary projects.
 Also the global logger which can save to a file.
 ]]
 
+if(WIN32)
```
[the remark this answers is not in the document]
saintmatthieu: Awesome ! How come I didn't think of that.

libraries/lib-audacity/GetMessageBoxCb.h
```diff
@@ -0,0 +1,10 @@
+// TODO header
+#pragma once
+
+#include "LibAudacityTypes.h"
+
+namespace LibAudacity
+{
+// To be implemented by the application.
+MessageBoxCb GetMessageBoxCb();
```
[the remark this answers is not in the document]
saintmatthieu: Hm. Not anymore. Good catch.

libraries/lib-audacity/EffectManager.cpp
```diff
@@ -853,22 +564,22 @@ EffectAndDefaultSettings &EffectManager::DoGetEffect(const PluginID & ID)
...
-      else {
-         if ( !dynamic_cast<AudacityCommand *>(component) )
-            AudacityMessageBox(
-               XO(
-"Attempting to initialize the following effect failed:\n\n%s\n\nMore information may be available in 'Help > Diagnostics > Show Log'")
-                  .Format( GetCommandName(ID) ),
+      else
+      {
+         if (!dynamic_cast<IAudacityCommand*>(component) && mMessageBoxCb)
+            mMessageBoxCb(
```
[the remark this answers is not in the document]
saintmatthieu: I think so too. I was being extra conservative. (`mMessageBoxCb` was also meant to be injected with `GetMessageBoxCb()`, but I lost that injection somewhere along the way. Anyway.)

libraries/lib-audacity/LibAudacityApi.cpp
```diff
@@ -0,0 +1,219 @@
...
+   EffectType type = plug->GetEffectType();
+
+   // Make sure there's no activity since the effect is about to be applied
+   // to the project's tracks.  Mainly for Apply during RTP, but also used
+   // for batch commands
+   if (flags & EffectManager::kConfigured)
+   {
+      stopPlaybackCb();
+      // Don't Select All if repeating Generator Effect
+      if (!(flags & EffectManager::kConfigured))
```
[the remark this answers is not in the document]
saintmatthieu: I didn't pay attention. Was added in [this PR](https://github.com/audacity/audacity/pull/722). I can't see how this could be executed either, though. Remove ? Then we'd have one callback injection less.

libraries/lib-files/CMakeLists.txt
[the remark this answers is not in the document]
saintmatthieu: Awesome ! How come I didn't think of it. Also maybe Qt has such helpers, too, when it gets needed in AU4.
I'll give it a try.

vsverchinsky:
> BasicCommandsInterface is the equivalent to my proposed IAudacityCommand, right ?

No. With that approach no changes to `AudacityCommand` is required. There will be only few changes in `EffectsManager` interface that remove dependency on ui framework, and command related logic should be moved into `AudacityWxCommandsInterface`. Example:
```cpp
bool EffectManager::DoAudacityCommand(const PluginID & ID,
                             const CommandContext &context,
                             bool shouldPrompt /* = true */)

{
   this->SetSkipStateFlag(false);
   //Below is the old implementation that makes EffectManager depend 
   //in AudacityCommand (or IAudacityCommand):
   /*AudacityCommand *command = GetAudacityCommand(ID);
   if (!command)
   {
      return false;
   }
   bool res = command->DoAudacityCommand(parent, context, shouldPrompt);
   return res;*/
   
   //With BasicCommandInterface that dependency is removed. All command-related logic should
   //be extracted into AudacityWxCommandsInterface 
   return CommandsInterface::Call().Do(context, ID, shouldPrompt);
}
```

saintmatthieu:
I pushed a better fix for that problem.
`DoEffect` is in fact only concerned with the application of plugins of type `PluginTypeEffect`. Review of the places where it is used confirms this. Hence, plugins of type `PluginTypeAudacityCommand` should have no mention in `DoEffect` whatsoever.
You were the first to observe that `EffectManager` was concerned with plugins of both type `Effect` and `AudacityCommand`. I the commits I just pushed, I extract the methods that look for both effects and commands in another class, `EffectAndCommandManager`, which can go in `src`.
Now there's no mention of commands in `lib-audacity-application-logic` anymore.

vsverchinsky (approved):

PR #7023, opened 2024-08-03
Update German Translation (3.6.2)

Gootector:
Update German Translation (3.6.2)

Greetings,
Gootector

Recommended:

### DIFF
--- a/locale/de.po
+++ b/locale/de.po
@@ -13,10 +13,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: audacity 3.6.0\n"
+"Project-Id-Version: audacity 3.6.2\n"
 "Report-Msgid-Bugs-To: audacity-translation@lists.sourceforge.net\n"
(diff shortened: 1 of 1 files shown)


LWinterberg (approved):
Fwiw: I'm unsure if Attack/Release are translated at all in German usually. I see that https://www.thomann.de/de/onlineexpert_page_kompressoren_die_dynamischen_parameter.html for example just uses the english terms, as does https://de.wikipedia.org/wiki/Kompressor_(Signalverarbeitung)

Gootector:
Hi, I'll check it soon. When deadline?

Gootector:
@LWinterberg Hi, I updated the translations. In Polish, I found "more correct" translations for Attack/Release terms, but there are Attack/Decay terms also, so... so I think that these terms are very similar. In German, I reverted to English. "The compressor's terms" are very specific...

Gootector:
@LWinterberg Please merge Polish & German. I finished the translations.

Issue #6991, opened 2024-07-30
[AU4] Generate thumbnails


grliszas14:
Generate PNG images with recent project thumbnails for home screen preview

Issue #6996, opened 2024-07-30
[AU4] Store thumbnails


grliszas14:
We need to store thumbnails for recent projects somewhere.

How thumbnails work in Musescore:
Musescore project file (*.mscz) is a zip file which contains a few things in it:
* notation project itself I guess (*.mscx)
* style file (*.mss)
* some audio/view settings (*.json)
* Thumbnails directory containing PNG files
* meta data with zip info
The way it works:
1. when you save a project, a thumbnail Pixmap is created and saved into Thumbnails directory
2. when you open Musescore, the file is just loaded from that dir

Two possible ways:
* Change the project format to a container (instead of a database) - this is a big task
* Store the Thumbnails in the database, serialize as a part of bxml project table.

QA: The thumbnail is created when AUP3 file is opened and converted to AUP4

kryksyh:
@luapmartin since the thumbnail is quite small, we can store it as a part of the project doc in serialized form, see ProjectSerializer.cpp/h

Tantacrul:
If this is relatively easy to do, I'm keeping it in Beta. However, if it causes big issues, we might delay. Seems to me the best time to do it is when releasing the new file format though, right?

ianalarina:
@embarc-gabriel take a look at this one in scope of "Save the preview image in the project file" for cloud projects

embarc-gabriel:
Estimate: (XS) 2 day

dozzzzer:
@embarc-gabriel 

Thumbnail-wise, it looks good. However, there's a regression in this PR: once the project is saved and Audacity is closed, the temp files (see the screenshot) are not removed.

<img width="415" height="262" alt="Image" src="https://github.com/user-attachments/assets/208b0d1a-9fab-49bd-91a3-8db1e26bb6de" />

embarc-gabriel:
@dozzzzer reallyyyyyy good catch! Thanks a lot!
Fixed!

dozzzzer:
Tested

PR #7032, opened 2024-08-05
[AU4] Fix project properties constructor

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6991

### DIFF
--- a/au4/src/project/view/projectpropertiesmodel.cpp
+++ b/au4/src/project/view/projectpropertiesmodel.cpp
@@ -31,7 +31,20 @@ using namespace au::project;
 ProjectPropertiesModel::ProjectPropertiesModel(QObject* parent)
     : QAbstractListModel(parent)
 {
-    m_project = context()->currentProject();
+}
+
+void ProjectPropertiesModel::init()
+{
+    m_project = globalContext()->currentProject();
+
+    globalContext()->currentProjectChanged().onNotify(this, [this]() {
+        m_project = globalContext()->currentProject();
+    });
+
+    if (!m_project) {
+        return;
+    }
+
     // if (project) {
     //     m_projectMetaInfo = project->metaInfo();
     // }
@@ -39,6 +52,8 @@ ProjectPropertiesModel::ProjectPropertiesModel(QObject* parent)
     m_project->captureThumbnailRequested().onNotify(this, [this]() {
         captureThumbnail();
     });
+
+    load();
 }
 
 void ProjectPropertiesModel::load()
--- a/au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -63,6 +63,7 @@ Rectangle {
         timeline.init()
         playCursorController.init()
         tracksViewState.init()
+        project.init()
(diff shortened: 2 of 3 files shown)


au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -63,6 +63,7 @@ Rectangle {
         timeline.init()
         playCursorController.init()
         tracksViewState.init()
+        project.load()
```
igorkorsukov, line 66: It's better to do the opposite here, call init, in which to call load.
init need call it once.
Load can be called several times for some reason.

au4/src/project/view/projectpropertiesmodel.cpp
```diff
@@ -36,13 +36,22 @@ ProjectPropertiesModel::ProjectPropertiesModel(QObject* parent)
     //     m_projectMetaInfo = project->metaInfo();
     // }
 
+}
+
+void ProjectPropertiesModel::init()
+{
+    if (!m_project) {
```
igorkorsukov, line 43: `m_project = context()->currentProject();` - it's better to call it in init.
Also we should subscribe to project changes in global context

au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
grliszas14: fixed

au4/src/project/view/projectpropertiesmodel.cpp
grliszas14: fixed

au4/src/project/view/projectpropertiesmodel.cpp
```diff
@@ -31,14 +31,29 @@ using namespace au::project;
...
+        return;
+    }
+
     // if (project) {
     //     m_projectMetaInfo = project->metaInfo();
     // }
 
+    globalContext()->currentProjectChanged().onNotify(this, [this]() {
+        m_project = globalContext()->currentProject();
+    });
```
igorkorsukov, line 50: Should be before 
```
if (!m_project) {
        **return**;
}
```

grliszas14: fixed

au4/src/project/view/projectpropertiesmodel.cpp
```diff
@@ -31,14 +31,29 @@ using namespace au::project;
...
+
+    if (!m_project) {
+        return;
+    }
+
     // if (project) {
     //     m_projectMetaInfo = project->metaInfo();
     // }
```
igorkorsukov, line 50: I lost the point, where do we get the properties now?
And when changing the project in the global context, we need to get new properties and call the load.

grliszas14: Right now we're not using the properties yet, see the commented metaInfo() function.
I've used the properties model to retrieve path for temporary thumbnail image in QML, we'll need to update this model later when we're going to have some project preferences UI to change meta info (and the thumbnail image path will change too as we implement this one: https://github.com/audacity/audacity/issues/6996 )

igorkorsukov (approved):

Issue #6552, opened 2024-05-31
[AU4] Fix play cursor shape


igorkorsukov:
### Description

![Image](https://github.com/audacity/audacity/assets/3818029/9971b5bc-cac5-48be-aa52-4e0e48327147)

DilsonsPickles:
See [this ticket](https://github.com/audacity/audacity/issues/6770) for Figma Link to detailed view of playhead cursor.

PR #7037, opened 2024-08-05
[AU4] Playhead cursor improvements

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6770
Resolves: https://github.com/audacity/audacity/issues/6552

### DIFF
--- a/au4/src/app/configs/data/shortcuts.xml
+++ b/au4/src/app/configs/data/shortcuts.xml
@@ -633,11 +633,11 @@
         <seq>End</seq>
     </SC>
     <SC>
-        <key>cursor-left</key>
+        <key>play-position-decrease</key>
         <seq>Left</seq>
     </SC>
     <SC>
-        <key>cursor-right</key>
+        <key>play-position-increase</key>
         <seq>Right</seq>
     </SC>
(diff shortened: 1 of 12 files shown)


au4/src/projectscene/iplaycursorcontroller.h
```diff
@@ -0,0 +1,19 @@
...
+{
+    INTERFACE_ID(processing::IPlayCursorController)
+
+public:
+    virtual ~IPlayCursorController() = default;
+
+    virtual void setTimelineContext(TimelineContext* context) = 0;
+    virtual void moveCursor(int pixels) = 0;
+    virtual void seekToX(double x) = 0;
+    virtual void updatePositionX(double x) = 0;
```
igorkorsukov, line 17: I believe we need to do things a little differently: 
* IPlayCursorController - this is the main controller for the play cursor and it should only operate on time (there should be no x coordinates, pixels, etc.)
* PlayCursorViewController - This is an adapter for view, and its main task is to convert coordinates, pixels into time and back, and pass this time to the main playcursor controller.

igorkorsukov (approved):

Issue #6748, opened 2024-06-25
[AU4] Track panel header - Pan knob


DilsonsPickles:
Pan knob behaviour

[Figma Link](https://www.figma.com/design/k5rS9tnZl60nYmhJhvI2vf/Track-headers?node-id=377-74245&t=nz14mBwwMmTHXTt2-4)

PR #7038, opened 2024-08-05
[AU4] Add dynamic tooltip to the balance knob

kryksyh:
Resolves: [#6748](https://github.com/audacity/audacity/issues/6748)

* added a knob with tooltip  as a separate element
* slightly changed KnobControl to handle hover and drag states properly, which is required for the tooltip display

Recommended:

### DIFF
--- a/au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/BalanceKnob.qml
+++ b/au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/BalanceKnob.qml
@@ -1,0 +1,30 @@
+/*
+* Audacity: A Digital Audio Editor
+*/
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+import Muse.Ui 1.0
+import Muse.UiComponents 1.0
+
+KnobControl {
+    id: root
+    isBalanceKnob: true
+
+    from: -100
+    to: 100
+    stepSize: 1
+
+    BalanceTooltip {
+        id: tooltip
+        value: root.value
+    }
+
+    onMouseEntered: {
+        tooltip.show()
+    }
+
+    onMouseExited: {
+        tooltip.hide()
+    }
+}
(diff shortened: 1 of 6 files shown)


au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/BalanceKnob.qml
```diff
@@ -0,0 +1,62 @@
...
+import QtQuick.Controls 2.15
+
+import Muse.Ui 1.0
+import Muse.UiComponents 1.0
+
+KnobControl {
+    id: root
+    isBalanceKnob: true
+
+    StyledPopupView {
```
igorkorsukov, line 14: It looks like this tooltip would be better made as a separate component, it will probably be needed in other places

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/BalanceKnob.qml
```diff
@@ -0,0 +1,62 @@
...
+                    let direction = value < 0 ? 'L' : value > 0 ? 'R' : '';
+                    return `${Math.abs(value)}${direction}`;
+                }
+            }
+        }
+    }
+
+    Timer {
+        id: popupTimer
+        interval: 200
```
igorkorsukov, line 49: This value should be taken from the settings.
It needs to be improved, add something like a delay tooltip to the Theme property. Take it from the uiconfiguration (need add method)

kryksyh: Done

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/BalanceKnob.qml
kryksyh: Done

igorkorsukov (approved):

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/BalanceTooltip.qml
```diff
@@ -0,0 +1,72 @@
...
+
+    Item {
+        anchors.fill: parent
+        id: content
+        Text {
+            anchors.fill: parent
+            text: "Pan:"
+        }
+
+        Text {
```
Eism, line 30: please use our components - StyledTextLabel

Issue #6644, opened 2024-06-09
In the File Menu Close and Exit would be better co-located


petersampsonaudacity:
### Your idea

The Audacity commands for **Close** and **Exit** perform somewhat related (but more than subtly different) functions and thus it might make sense to co-locate them in the File menu.

**As it is now:**
![image](https://github.com/audacity/audacity/assets/50205984/5dd122cc-522b-49cc-bfbe-4d095dc5300c)

**Proposed rearrangement:**
![image](https://github.com/audacity/audacity/assets/50205984/e6b20c04-38a3-430f-a1df-0501029d39e2)

@DavidBailes  suggests later in this thread:
> So maybe:
> **Close Project**
> **Exit**

_I support that suggestion as it would provide greater clarity of purpose of the **Close** command._

### Problem to be solved

A recent posting in a thread on GitHub #1464 (9th post) shows that users can be confused as to the differences between Close and Exit - and indeed may even overlook the Close and only use Exit (when they don't really intend to close all open projects).

So it may make sense to co-locate them in the menu to encourage "discoverabilty" of the differences between the two.  

It would also the two commands to be co-located on the manual page for the File menu.

### Prior art

See above.

### Additional context

User @jpgreen3 wrote in #1464 

>F. Exiting Audacity via 'file' - 'exit' and the upper - right corner X produces different results. Using the 'X' only closes the running instance of audacity that is the focus at the moment, which seems to me the correct behavior - though I'm not sure closing this way invokes the 'compacting' routine when appropriate. The 'file' - 'exit' mode causes all currently running instances of Audacity to close, which to me is an unexpected and undesirable action. Both modes of exiting the program should close only the current instance, invoking the 'compacting' routine when appropriate, and prompting the user to save the project - or not - before close if unsaved edits exist.

----------------------------------------------------------------------------------

@LWinterberg I will add this to your Cleanup super-bug

jpgreen3:
This is a great idea.  Out of habit, my eye always goes to the bottom of the ‘File’ menu looking for ‘Exit’.  

To make the distinction even clearer – though this is decidedly non-standard – the menu items could read: 

Close This

Exit All

 

Or something similar. 

John Green

 

From: Peter Sampson ***@***.***> 
Sent: Sunday, June 9, 2024 3:33 AM
To: audacity/audacity ***@***.***>
Cc: jpgreen3 ***@***.***>; Mention ***@***.***>
[...]

PR #7041, opened 2024-08-06
Changed quit and close file menu item's names

nalin243:
Resolves: *#6644 partially*

Recommended:

### DIFF
--- a/src/menus/FileMenus.cpp
+++ b/src/menus/FileMenus.cpp
@@ -529,7 +529,7 @@ auto FileMenu()
          //   AudioIONotBusyFlag(), wxT("Shift+A") )
       ),
 
-      Section( "Close", Command( wxT("Close"), XXO("&Close"), OnClose,
+      Section( "Close", Command( wxT("Close"), XXO("&Close Project"), OnClose,
             AudioIONotBusyFlag(), wxT("Ctrl+W") )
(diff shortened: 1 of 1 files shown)


src/menus/FileMenus.cpp
```diff
@@ -558,7 +558,7 @@ auto FileMenu()
          // pull it out
          // and put it in the Audacity menu for us based on its ID.
          /* i18n-hint: (verb) It's item on a menu. */
-         Command( wxT("Exit"), XXO("E&xit"), OnExit,
+         Command( wxT("Exit"), XXO("Quit Audacity"), OnExit,
```
LWinterberg, line 561: Please keep the accelerator key (letter after &). The way you test them (on Windows and Linux) is by pressing alt, at which point the accelerators get underlined:

image

You can make any character the accelerator as long as it doesn't overlap with any of the other ones shown in the same menu.

LWinterberg (changes requested):

src/menus/FileMenus.cpp
nalin243: Should the accelerator key in this case be Q?

LWinterberg: that looks about right.

PR #7054, opened 2024-08-06
[AU4] Effects. Concept

Eism:


Eism:
The solution was partially merged within separate PRs

Issue #7002, opened 2024-07-31
Export with a master effect can lead to silence in exported file


DavidBailes:
### Bug description

See related bug: https://github.com/audacity/audacity/issues/7001

Probably only occurs for some effects, as in the case of the related bug, but haven't tested.

### Steps to reproduce

1. Create a mono audio track
2. Add the MuseFX reverb effect to master effects.
3. Export the audio.
4. Import the file that was just exported

### Expected behavior

The audio in the imported file is as expected.

### Actual behavior

The audio in the imported file consists of silence.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

vsverchinsky:
Same with mono track with per track effects applied and exported using channel mapping (one to many)

dozzzzer:
Tested

Issue #7001, opened 2024-07-31
Export multiple with a master effect can lead to silence in exported files.


DavidBailes:
### Bug description

This only seems to effect export multiple for mono tracks.
The problem only occurs for some effects.
Effects that I tested where the problem occurs: the MuseFx effects: reverb, pitchfix and de-ess, and Tokyo dawn labs: TDR Kotelnikov. Reported on forum: Graillon 2.
Effects that I tested where the problem does not occurs: builtin reverb, and builtin bass and treble.

### Steps to reproduce

1. Create two audio tracks, and make sure they have different names.
2. Add the MuseFX reverb effect to the master effects.
3. Export multiple.
4. Import the two files that were just exported.

### Expected behavior

The audio in  the imported files should be as expected

### Actual behavior

The audio in the imported files is silence.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Reported on forum: https://forum.audacityteam.org/t/master-effects-making-exports-silent/112780

dozzzzer:
Tested

PR #7055, opened 2024-08-07
Export fixes

vsverchinsky:
Resolves: #7002, #7001, #7059

Channel maps disabled when exporting with master effects on.

QA: Check stereo to mono and mono to stereo export. Additionally please check for regression: Mix and Render and Stereo to Mono

Recommended:

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -2110,7 +2110,7 @@ bool AudioIO::ProcessPlaybackSlices(
 
          if(len > 0)
          {
-            for(unsigned i = 0; i < seq->NChannels(); ++i)
+            for(unsigned i = 0, cnt = std::min(seq->NChannels(), mNumPlaybackChannels); i < cnt; ++i)
                pointers[i] = mProcessingBuffers[bufferIndex + i].data() + offset;
 
             for(unsigned i = seq->NChannels(); i < mNumPlaybackChannels; ++i)
@@ -2178,7 +2178,7 @@ bool AudioIO::ProcessPlaybackSlices(
          const auto numChannels = seq->NChannels();
          if(numChannels > 1)
          {
-            for(unsigned n = 0; n < seq->NChannels(); ++n)
+            for(unsigned n = 0, cnt = std::min(numChannels, mNumPlaybackChannels); n < cnt; ++n)
             {
                const auto gain = seq->GetChannelGain(n);
                for(unsigned i = 0; i < samplesAvailable; ++i)
@@ -2240,9 +2240,6 @@ bool AudioIO::ProcessPlaybackSlices(
          );
       }
    }
-
-   for(auto& buffer : mMasterBuffers)
-      buffer.clear();
 
    return progress;
 }
--- a/libraries/lib-mixer/Mix.cpp
+++ b/libraries/lib-mixer/Mix.cpp
@@ -25,6 +25,8 @@
 #include "float_cast.h"
 #include <numeric>
 
+#include "DownmixSource.h"
+
 namespace {
 template<typename T, typename F> std::vector<T>
 initVector(size_t dim1, const F &f)
@@ -45,6 +47,8 @@ initVector(size_t dim1, size_t dim2)
 
 namespace
 {
+
+
 void ConsiderStages(const Mixer::Stages& stages, size_t& blockSize)
 {
    for (const auto& stage : stages)
@@ -78,6 +82,11 @@ size_t FindBufferSize(
 
    return blockSize;
 }
+
+auto NeedsDitherPred(const MixerOptions::StageSpecification &spec) {
+   return spec.mpFirstInstance && spec.mpFirstInstance->NeedsDither();
+};
+
 } // namespace
 
 Mixer::Mixer(
@@ -98,14 +107,6 @@ Mixer::Mixer(
     , mTimesAndSpeed { std::make_shared<TimesAndSpeed>(TimesAndSpeed {
          startTime, stopTime, warpOptions.initialSpeed, startTime }) }
 
-    // PRL:  Bug2536: see other comments below for the last, padding argument
-    // TODO: more-than-two-channels
-    // Issue 3565 workaround:  allocate one extra buffer when applying a
-    // GVerb effect stage.  It is simply discarded
-    // See also issue 3854, when the number of out channels expected by the
-    // plug-in is yet larger
-    , mFloatBuffers { 3, mBufferSize, 1, 1 }
-
     // non-interleaved
     , mTemp { mNumChannels, mBufferSize, 1, 1 }
     , mBuffer { initVector<SampleBuffer>(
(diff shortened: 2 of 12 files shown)


libraries/lib-mixer/Mix.cpp
```diff
@@ -78,6 +96,214 @@ size_t FindBufferSize(
...
+   void FindChannelFlags(unsigned char* channelFlags, size_t numChannels, size_t) override
+   {
+      for(size_t i = 0; i < numChannels; ++i)
+         channelFlags[i] = true;
+   }
+   bool CanMakeMono() const override { return true; }
+};
+
+//! Combines multiple audio graph sources into a single source
+class DownmixStage final : public AudioGraph::Source
```
saintmatthieu, line 185: I like the initiative of moving some of the complexity behind appropriate abstractions. Before this gets merged, though, I'd like to see these new classes get their own files. Then functions like `MixBuffers` would also be relocated to `DownmixStage.cpp`, and maybe others, too.

libraries/lib-mixer/Mix.cpp
```diff
@@ -150,45 +363,76 @@ Mixer::Mixer(
...
       for (const auto &stage : input.stages)
          if (
             auto& pNewDownstream =
-               RegisterEffectStage(*pDownstream, stage, outRate))
+               RegisterEffectStage(*pDownstream, sequence->NChannels(), stage, outRate))
          {
             pDownstream = pNewDownstream.get();
          }
-      mDecoratedSources.emplace_back(Source{ source, *pDownstream });
+      mixdownSources.emplace_back(
```
saintmatthieu, line 391: Looking at this I first wondered why a "sequence source" might be the last node of an track's realtime effect stack. Now I see that, in fact, it exposes the processed audio, but uses `*sequence` to get information about channels and gains.
I think it'd be less confusing if it were called `ProcessedSequenceDownmixSource` and it didn't take a `WideSampleSequence` reference but only the metadata it needs about it, to make it really clear that it's not going to pull data from that reference.

libraries/lib-mixer/Mix.cpp
```diff
@@ -78,6 +96,214 @@ size_t FindBufferSize(
...
+      if (mApplyGain == Mixer::ApplyGain::Discard)
+         std::fill(gains, gains + mNumChannels, 1.0f);
+
+      size_t maxOut = 0;
+      for (auto c = 0;c < data.Channels(); ++c)
+         data.ClearBuffer(c, maxToProcess);
+
+      for (auto& downmixSource : mDownmixSources)
+      {
+         auto oResult = downmixSource->GetDownstream().Acquire(mFloatBuffers, maxToProcess);
```
saintmatthieu, line 241: I was also confused by that. We have a source, from which we'd normally like to acquire data, but to do this we have to ask its _downstream_ to provide us with data. Ruined me poor brain.
I think what might have brought you to this naming is this:
```cpp
      AudioGraph::Source *pDownstream = &source;
      for (const auto &stage : input.stages)
         if (
            auto& pNewDownstream =
               RegisterEffectStage(*pDownstream, sequence->NChannels(), stage, outRate))
         {
            pDownstream = pNewDownstream.get();
         }
      mixdownSources.emplace_back(
         std::make_unique<SequenceDownmixSource>(
            *pDownstream,
            *sequence,
            pMixerSpec ? &pMixerSpec->mMap[i] : nullptr
         )
      );
```
In that case, `pDownstream` is actually an upstream. It's just called that way because, at the end of this loop, it's supposed to point to the last source of the graph, or its "downstream". But from the point of view of a `DownmixSource`, it's an upstream.

I would "simply" remove the `GetDonwstream()` method and give the `DownmixSource` the `Acquire` , `Release` and `Remaining` methods.

saintmatthieu (commented):
I think it's going in the right direction, but I've asked a few clarifications. It's a tough PR, though, so in the next round of review I may give more attention to other things.

libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2110,7 +2110,7 @@ bool AudioIO::ProcessPlaybackSlices(
 
          if(len > 0)
          {
-            for(unsigned i = 0; i < seq->NChannels(); ++i)
+            for(unsigned i = 0, cnt = std::min(seq->NChannels(), mNumPlaybackChannels); i < cnt; ++i)
```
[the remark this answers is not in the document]
vsverchinsky: > Does it then crash?

I'd say it should, but didn't check myself

> AFAICS, if we were doing things properly further down the call stack and mixed left and right rather than only keeping the left channel

```cpp
for(unsigned n = 0, cnt = std::min(numChannels, mNumPlaybackChannels); n < cnt; ++n)
{
   const auto gain = seq->GetChannelGain(n);
   for(unsigned i = 0; i < samplesAvailable; ++i)
      mMasterBuffers[n][i] += mProcessingBuffers[bufferIndex + n][i] * gain;
}
```

`Keep-only-left-channel` logic implemented in per-track effects, when source track is mono and playback is stereo, that could be clearly observed with spatial effects

libraries/lib-audio-graph/AudioGraphSource.h
```diff
@@ -28,6 +28,8 @@ class AUDIO_GRAPH_API Source {
 public:
    using Buffers = AudioGraph::Buffers;
 
+   Source() = default;
```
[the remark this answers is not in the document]
vsverchinsky: No, this is a leftover, I tried to undestand where does compiler tries to use a copy-constructor when it wasn't expected to do so. Thanks!

libraries/lib-mixer/Mix.cpp
```diff
@@ -78,6 +96,214 @@ size_t FindBufferSize(
...
+   AudioGraph::Source& mDownstream;
+   const WideSampleSequence& mSequence;
+   //! many-to-one mixing of channels
+   //! Pointer into array of arrays
+   const ArrayOf<bool> * mpMap {};
+public:
+
+   SequenceDownmixSource(AudioGraph::Source& downstream,
+                         const WideSampleSequence& sequence,
+                         //! Null or else must have a lifetime enclosing this objects's
```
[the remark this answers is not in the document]
vsverchinsky: I wish I wrote this but this is a copy-paste 🥲

libraries/lib-mixer/Mix.cpp
```diff
@@ -150,45 +363,76 @@ Mixer::Mixer(
...
+      mMixDownStage = std::make_unique<DownmixStage>(
+         std::move(mixdownSources), 2, mBufferSize, ApplyGain::MapChannels
+      );
+
+      AudioGraph::Source* pDownstream = mMixDownStage.get();
       for (const auto& stage : *mMasterEffects)
          if (
             auto& pNewDownstream =
-               RegisterEffectStage(*pDownstream, stage, outRate))
+               RegisterEffectStage(*pDownstream, 2, stage, outRate))
```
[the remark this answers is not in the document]
vsverchinsky: No good enough explanation. When I wrote this I though that the best thing I could do is to make exported audio to be exactly same as one would hear when hit play. But there could be only one playback channel depending on device configuration. 
I agree now that what you suggest is better option.

libraries/lib-mixer/Mix.cpp
vsverchinsky: Probably the confusion comes from the fact that `DownmixSource` isn't in same hierachy with `AudioSource`
`SequenceDownmixSource`. I'll add more comments that should clarify that and extract it into a sepatate source file as you suggested

libraries/lib-mixer/Mix.cpp
vsverchinsky: > I would "simply" remove the GetDonwstream() method and give the DownmixSource the Acquire , Release and Remaining methods.

I'd preferred not to do that because that will simply bloat the interface.

vsverchinsky: > It's just called that way because, at the end of this loop, it's supposed to point to the last source of the graph, or its "downstream

And it is a downstream of individual processing sequence from the point of view of `DownmixStage`.

saintmatthieu (approved):

Issue #7048, opened 2024-08-06
Audio Clip loses Data after Recording


chinakov:
### Bug description

If the User decides to record a stereo audio clip, then splitting the resulting audio clip, and recording again, after stopping the recording the resulting and previous audio clips become empty.

### Steps to reproduce

1. Launch Audacity
2. Make sure you're recording in Stereo, and record some audio
3. Split the resulting Stereo track into Mono tracks, or just split them.
4. Place playhead after the clip, press record again.
5. Stop the recording
6. Observe result

### Expected behavior

The resulting clips are empty.

### Actual behavior

The resulting clips shouldn't be empty.

https://github.com/user-attachments/assets/345bba58-3665-4343-a6f9-49f023e55ee0

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

dozzzzer:
Cannot reproduce on Windows 11 / Audacity 3.6.1:

https://github.com/user-attachments/assets/d3995aa4-2323-4f37-83dd-5eadf0f406d3

chinakov:
Confirmed to be introduced together with the mentioned issue in https://github.com/audacity/audacity/commit/aaac5fd452549dfb6791b5d58d61fe97e04ebd75 

Thank you, @dozzzzer

chinakov:
Tested, thanks

Issue #7050, opened 2024-08-06
Right channel's data gets duplicated onto the left channel when recording in stereo on a split stereo track


dozzzzer:
### Steps to reproduce

1. Generate chirp
2. Generate tone
3. Select both tracks and make a stereo track
4. Split stereo track
5. Make sure recording channels is set to '2'
6. Make sure the stereo input signal has distinct Left and Right channels that are different from one another
7. Put the cursor after the clips and click Record
8. Click Stop
9. Observe

### Expected behavior

As soon as the recording begins, the left and right channels' data remains untouched.
As soon as the recording is stopped, the data from the left and right channels is recorded in two separate clips. The two existing clips also remain unchanged.

### Actual behavior

As soon as the recording begins, the data of the left channel (chirp) gets duplicated onto the right channel (bottom mono track).
As soon as the recording is stopped, the recorded as well as existing audio data from the right channel is duplicated onto the left channel (the top mono track).

https://github.com/user-attachments/assets/b117150d-1190-446e-b9f1-6dbd71e589a5

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

### Additional context

Probably the same issue as https://github.com/audacity/audacity/issues/7048

dozzzzer:
Bisects to https://github.com/audacity/audacity/commit/983dc1b7c7fd5008fd40420fefe6b63d68cd6a77

dozzzzer:
More exact bisection: https://github.com/audacity/audacity/commit/aaac5fd452549dfb6791b5d58d61fe97e04ebd75

dozzzzer:
Tested

petersampsonaudacity:
@dozzzzer @kryksyh 

This looks to be fixed in that latest 3.6.2 Beta.

I think this isue could be closed.

Issue #6886, opened 2024-07-16
Splitting a stereo track and then recording in mono isn't displayed correctly


dozzzzer:
### Steps to reproduce

1. Generate a stereo chirp
2. Split the track to mono
3. Make sure the no. of channels is set to 1
4. Place the cursor just after the clip above and click Record
5. Observe

### Expected behavior

A new clip is being recorded only on the top track.

### Actual behavior

Both the top and bottom tracks appear to be recording to new clips. However, when the Stop button is clicked, the recorded audio is saved to the top track only.

https://github.com/user-attachments/assets/df4c6a1c-82a8-44fc-af1f-d49c131176a4

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

petersampsonaudacity:
This **only** happens after splitting stereo.

If instead you create two mono tracks in place of steps 1 and 2 at Step 5 you only get recording in the top track.

kryksyh:
@dozzzzer could you test this one using the build from #7060? I believe it should also be fixed.

dozzzzer:
This is indeed fixed by #7060
![image](https://github.com/user-attachments/assets/f7dc1017-edc8-4dff-ba89-1717ca640ab9)

petersampsonaudacity:
@dozzzzer 

This looks ti be fixed in that latest 3.6.2 Beta.

I think this isue could ne closed.

PR #7060, opened 2024-08-07
Fix "Split stereo track" operation

kryksyh:
Due to regression, after the stereo-to-mono split, both new tracks were assigned the same ID, which should be unique to all tracks. This was causing weird behavior, including data loss and crashes.

Resolves: #7048, #7050, #6886

Recommended:
- [x] Each commit compiles and runs on my machine without known undesirable changes in behavior


saintmatthieu (approved):
Congrats !

vsverchinsky (approved):

Issue #7042, opened 2024-08-06
[AU4] Fix audio data selection behaviour


DilsonsPickles:
#### Description

When the User has an audio data selection on the canvas, selecting a clip should clear the audio data selection.

PR #7080, opened 2024-08-12
[AU4] Fix audio data selection behaviour

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7042


au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
```diff
@@ -123,6 +124,7 @@ Item {
 
                 onRequestSelected: {
                     clipsModel.selectClip(clipItem.key)
+                    clipSelectedRequested()
```
igorkorsukov, line 127: Prefer this: id.funс() (signal) 
So it is clear what comes from where and the interpreter will find the method faster (no need to go through all the methods of all the ancestors)

```
root.clipSelectedRequested()
```

igorkorsukov (approved):

PR #7088, opened 2024-08-13
It is recommended to use emplace_back instead of push_back.

monster29000:

Recommended:

### DIFF
--- a/au4/src/app/app.cpp
+++ b/au4/src/app/app.cpp
@@ -60,7 +60,7 @@ App::App()
 
 void App::addModule(modularity::IModuleSetup* module)
 {
-    m_modules.push_back(module);
+    m_modules.emplace_back(module);
 }
 
 int App::run(int argc, char** argv)
--- a/au4/src/app/main.cpp
+++ b/au4/src/app/main.cpp
@@ -145,7 +145,7 @@ int main(int argc, char** argv)
 
     std::vector<char*> argvUTF8;
     for (auto& arg : argvUTF8Q) {
-        argvUTF8.push_back(arg.data());
+        argvUTF8.emplace_back(arg.data());
     }
 
     // Don't use the arguments passed to main(), because they're in the local ANSI code page.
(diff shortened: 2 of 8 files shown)


au4/src/app/app.cpp
```diff
@@ -60,7 +60,7 @@ App::App()
 
 void App::addModule(modularity::IModuleSetup* module)
 {
-    m_modules.push_back(module);
+    m_modules.emplace_back(module);
```
igorkorsukov, line 63: The use of emplace_back does not matter here, because only a pointer is inserted, and not some structure by which memory is allocated and copied.

au4/src/app/main.cpp
```diff
@@ -145,7 +145,7 @@ int main(int argc, char** argv)
 
     std::vector<char*> argvUTF8;
     for (auto& arg : argvUTF8Q) {
-        argvUTF8.push_back(arg.data());
+        argvUTF8.emplace_back(arg.data());
```
igorkorsukov, line 148: The use of emplace_back does not matter here, because only a pointer is inserted (thanks Qt Implicit Sharing), and not some structure by which memory is allocated and copied.

au4/src/appshell/internal/applicationuiactions.cpp
```diff
@@ -64,17 +64,17 @@ const UiActionList ApplicationUiActions::m_actions = {
     UiAction("about-musescore",
              au::context::UiCtxAny,
              au::context::CTX_ANY,
-             TranslatableString("action", "&About MuseScore…")
+             TranslatableString("action", "&About MuseScore�?")
```
igorkorsukov, line 67: What are these changes for? And they don't match the commit description

au4/src/appshell/view/preferences/playbackpreferencesmodel.cpp
```diff
@@ -100,7 +100,7 @@ QStringList PlaybackPreferencesModel::audioApiList() const
 {
     QStringList result;
     for (const std::string& api: audioDevicesProvider()->audioApiList()) {
-        result.push_back(QString::fromStdString(api));
+        result.emplace_back(QString::fromStdString(api));
```
igorkorsukov, line 103: Are you sure it matters for QString and std strings?

PR #7092, opened 2024-08-13
7091 UI agnostic amplify

### LINKED ISSUE #7091: UI-agnostic built-in effects: Amplify
### Description

UI-agnostic built-in effects: Amplify

saintmatthieu:
Resolves: #7091

Recommended:


saintmatthieu:
@vsverchinsky this PR depends on #6954, hence the many commits. At the time of writing, the last 4 commits only are specific to this branch.

Eism:
I liked the previous approach better - we had one class related to the amplify effect, now we have two classes in two places.
I also don't understand how to implement a qml view for this effect.
I would suggest looking at my solution in [this PR](https://github.com/audacity/audacity/pull/7054) - move Amplify to au4 code in the effects module. We will have one class and one place, we can also replace wxWidgets with qml in this class and there will be a few changes.

saintmatthieu:
Hey @Eism,

I see in your PR (which looks like an RFC (Request For Comments) at the moment?) that we have a fundamentally different approach. You opt for a copy-paste approach while I try to make AU3 code reusable. I'm doing so because what I think should be reused is _not only_ a processing method, but also (using `EffectAmplify` as example) :
* the application logic of the effect: e.g. `EffectAmplify::Init()` inspects the contents of the audio track selection, to know which amount to apply to achieve a desired target peak amplitude value.
* the integration in the broader Audacity application logic: `EffectAmplify` inherits from e.g. `PerTrackEffect::Process`, which (among other things)
  * creates temporary track objects from the selection,
  *  renders time-stretching if any
  * the effect processor method is provided the audio from the track clips
  * throws away the result of the processing if there was a failure, and commits it otherwise.

For this reason I urge to reuse as much as possible the effect framework, at least for the destructive effects. (This is not a use case for the MuseScore audio engine anyway.)

> I also don't understand how to implement a qml view for this effect.

If the generic approach we must use for Nyquist and Vamp effects is also adopted for built-in effects, then on the middle term it could look like in [my response to @vpereverzev's message](https://mu--se.slack.com/archives/C07D86RJY73/p1723634346437349?thread_ts=1722344540.779879&cid=C07D86RJY73). If you need something now, you should write a new class that inherits from `AmplifyBase` ; the members you need for read/write are declared as `protected`. I demonstrated that on my PoC, please take a look at [this video](https://mu--se.slack.com/archives/C06RX437GJV/p1722943713942859) at 1:33 : 
image
`Au3Amplify` can read and change `mRatio` and other parameters.

saintmatthieu:
@Eism actually, our concerns might be compatible.

In your PR, you've introduced `au3::effects::EffectAmplify`, which mirrors AU3's `EffectAmplify`.
`au3::effects::EffectAmplify` could well copy-paste the body of the original, wxWidget-dependent `EffectAmplify::PopulateOrExchange` method and wx with Qt widgets like you intend to do. Only you'd need `au3::effects::EffectAmplify` to also inherit from the `AmplifyBase` class I'm introducing here, so you can access the members you need for the UI (`AmplifyBase::mRatio`, `AmplifyBase::mClip`, etc.)

saintmatthieu:
You've been asking about that `Show()` method.

The `Show` method is only on my PoC branch. I introduced an `IEffectDialog` abstract API, inherited from `Au3Amplify` (my au4 version of your `au3::effects::EffectAmplify`, so that it can be called from `DoEffect` independently on the implementation (whether amplify, compressor, etc.):
```cpp
namespace au::au3
{
class IEffectDialog
{
public:
    virtual ~IEffectDialog() = default;
    virtual bool Show() = 0;
};
} // namespace au::au3
```
then `class Au3Amplify final : public AmplifyBase, public IEffectDialog`, which allows it to get called in
```cpp
auto showEffectHostInterfaceCb = [](Effect& effect, ...)
      {
         ...
         auto dialog =
            dynamic_cast<au::au3::IEffectDialog*>(&effect);
         if (!dialog)
            return false;
         if (!dialog->Show())
            return false;
         ...
         return true;
      };
...
AudacityApplicationLogic::DoEffect(..., std::move(showEffectHostInterfaceCb));
```
But this `IEffectDialog` is only PoC code. Maybe instead you will use
`class au3::effects::EffectAmplify : public AmplifyBase, public AbstractEffectModel`, the callback would dynamic-cast `effect` to an `AbstractEffectModel` and call whatever method shows the UI.

vsverchinsky (approved):

Issue #7093, opened 2024-08-13
Audacity maxes out on a CPU core when recording in a minimized state


dozzzzer:
### Steps to reproduce

1. Start Audacity
2. Minimize it
3. Observe the load on the CPU

### Expected behavior

CPU load remains the same as in the maximized (default) state of the Audacity window.

### Actual behavior

One CPU core will max out at around 100% when Audacity is minimized.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

DavidBailes:
See also: https://github.com/audacity/audacity/issues/4781

dozzzzer:
Tested! CPU usage doesn't increase anymore when Audacity is minimized.

Issue #4781, opened 2023-06-09
High CPU usage when audacity is minimized (Xorg+gtk3).


cs-cat:
### Bug description

When recording audio with Audacity, minimizing the window leads to increased CPU usage for both Audacity and Xorg. This issue appears to be specific to wxWidgets-gtk3, which is the default option in Arch Linux. However, I was unable to reproduce the problem using the AppImage provided in the releases, as it uses wxWidgets-gtk2.

### Steps to reproduce

1. launch audacity 3.3.2 built with wxWidgets-gtk3.
2. start recording audio.
3. minimize the window and observe CPU usage of audacity and Xorg.

### Expected behavior

The CPU usage of audacity is as low as when audacity is in the foreground (~25%).

### Actual behavior

CPU usage of audacity increases to ~80% and CPU usage of Xorg increases to ~30%

### Audacity Version

stable version (from a Linux repository)

### Operating system

Linux

### Additional context

After executing `perf record`, it has been noted that `wxApp::DoIdle` consumes a substantial amount of CPU time. Additionally, functions such as `CellularPanel::HandleCursorForPresentMouseState` contribute to the high CPU usage of Xorg.
The stack trace is below:
```
         - 30.44% wxApp::DoIdle                                                                                                        
            - 15.83% wxAppBase::ProcessIdle                                                                                            
               - 15.67% wxFrame::SendIdleEvents                                                                                        
                  - 15.51% wxWindowBase::SendIdleEvents                                                                                
                     - 14.12% wxWindowBase::SendIdleEvents                                                                             
                        - 13.09% wxWindowBase::SendIdleEvents                                                                          
                           + 10.22% wxWindowBase::SendIdleEvents                                                                       
                           + 1.28% wxEvtHandler::SafelyProcessEvent                                                                    
                             0.54% wxWindowBase::UpdateWindowUI                                                                        
                     + 0.67% wxFrameBase::OnInternalIdle                                                                               
            - 14.54% 0x7ffff6caac1b                                                                                                    
               - wxEvtHandler::ProcessPendingEvents                                                                                    
                  - 14.49% wxEvtHandler::ProcessEvent                                                                                  
                     - 14.48% wxEvtHandler::ProcessEventLocally                                                                        
                        - 14.47% wxEvtHandler::TryHereOnly                                                                             
                           - 14.35% CellularPanel::HandleCursorForPresentMouseState                                                    
                              + 10.81% wxGetMouseState                                                                                 
[...]

PR #7098, opened 2024-08-14
Prevent the ruler panel to call Refresh if the app is minimized

kryksyh:
Resolves: #7093, #4781

If the window is minimized, the ruler panel tries to repaint itself infinitely.
The reason is that the horizontal position is updated if auto-scrolling is enabled (`mLastDrawnH != viewInfo.hpos`), but the OnPaint method is not called because the window is not visible.  As a result, "bool changed" is always true, resulting in high CPU usage.

Recommended:


vsverchinsky (approved):
👍

PR #7099, opened 2024-08-14
[AU4] Effects scanning and meta reading. VST

Eism:


au4/src/au3wrap/CMakeLists.txt
```diff
@@ -377,24 +402,101 @@ set(AU3_SRC
...
 
+if(WIN32)
+   set(PlatformCompatibilityOsSpecific PlatformCompatibilityWindows )
+elseif(APPLE)
+   set(PlatformCompatibilityOsSpecific PlatformCompatibilityMacOS )
+else()
+   set(PlatformCompatibilityOsSpecific PlatformCompatibilityLinux )
+endif()
+
+set(MODULE_SRC ${MODULE_SRC} ${AU3_LIBRARIES}/lib-files/${PlatformCompatibilityOsSpecific}.cpp)
```
igorkorsukov, line 465: Just out of interest, I'm not calling for correction, why not just like this?
```
if(WIN32)
   set(MODULE_SRC ${MODULE_SRC} ${AU3_LIBRARIES}/lib-files/PlatformCompatibilityWindows.cpp)
elseif(APPLE)
   set(MODULE_SRC ${MODULE_SRC} ${AU3_LIBRARIES}/lib-files/PlatformCompatibilityMacOS.cpp)
else()
   set(MODULE_SRC ${MODULE_SRC} ${AU3_LIBRARIES}/lib-files/PlatformCompatibilityLinux.cpp)
endif()
```

If there are more platform-dependent files, it is easy to add them, but in the original solution we will have to create more and more new variables

au4/src/au3wrap/CMakeLists.txt
```diff
@@ -377,24 +402,101 @@ set(AU3_SRC
...
+    ${AU3_LIBRARIES}/lib-vst3/VST3Utils.h
+    ${AU3_LIBRARIES}/lib-vst3/VST3Wrapper.cpp
+    ${AU3_LIBRARIES}/lib-vst3/VST3Wrapper.h
+    ${AU3_LIBRARIES}/lib-vst3/internal/ConnectionProxy.cpp
+    ${AU3_LIBRARIES}/lib-vst3/internal/ConnectionProxy.h
+    ${AU3_LIBRARIES}/lib-vst3/memorystream.cpp
+    ${AU3_LIBRARIES}/lib-vst3/memorystream.h
 )
 
+if(WIN32)
```
igorkorsukov, line 457: better to use our platform defines from `include(GetPlatformInfo)`

au4/src/au3wrap/CMakeLists.txt
```diff
@@ -377,24 +402,101 @@ set(AU3_SRC
...
+    ${AU3_LIBRARIES}/lib-vst3/VST3Instance.cpp
+    ${AU3_LIBRARIES}/lib-vst3/VST3Instance.h
+    ${AU3_LIBRARIES}/lib-vst3/VST3Utils.cpp
+    ${AU3_LIBRARIES}/lib-vst3/VST3Utils.h
+    ${AU3_LIBRARIES}/lib-vst3/VST3Wrapper.cpp
+    ${AU3_LIBRARIES}/lib-vst3/VST3Wrapper.h
+    ${AU3_LIBRARIES}/lib-vst3/internal/ConnectionProxy.cpp
+    ${AU3_LIBRARIES}/lib-vst3/internal/ConnectionProxy.h
+    ${AU3_LIBRARIES}/lib-vst3/memorystream.cpp
+    ${AU3_LIBRARIES}/lib-vst3/memorystream.h
```
igorkorsukov, line 454: Question for the future, have you checked if there are any hard dependencies with some code from au3? Or can these sources be used independently by add them to some separate module?

au4/src/effects/effectsmodule.cpp
```diff
@@ -65,6 +86,13 @@ void EffectsModule::onInit(const muse::IApplication::RunMode& mode)
         return;
     }
 
+    ModuleManager::Get().Initialize();
```
igorkorsukov, line 89: What is ModuleManager?

libraries/lib-files/PlatformCompatibilityMacOS.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "PlatformCompatibility.h"
+#include <CoreServices/CoreServices.h>
+#include <limits.h>
+#include <mach-o/dyld.h>
+#include <pwd.h>
+#include <string>
+#include <sys/types.h>
+#include <unistd.h>
+
+std::string PlatformCompatibility::GetUserDataDir()
```
igorkorsukov, line 27: For us, these classes that return paths can become a problem.
The application as a whole should have one source of truth for returning standart paths.
In AU4, this is IGlobalConfiguration (https://github.com/musescore/framework_tmp/blob/main/framework/global/iglobalconfiguration.h) (internally, this is QStandardPaths)
If we use these classes, there will be two sources of paths and the paths can be different for the same entity, this will become a problem.
It would be good to do it like this:
* In AU3, add an interface that returns paths.
* Add the ability to register an implementation of this interface.
* In AU4, we register our implementation, which receives paths from IGlobalConfiguration.
* In AU3, our own implementation should be substituted, if needed.

saintmatthieu: I agree, there should be one source of truth. If you already have this functionality implemented in AU4, that's the one that should be reused, of course.
This code is a bit outdated. @vsverchinsky suggested a better approach which I applied to my PR: see [this commit](https://github.com/audacity/audacity/pull/7017/commits/17d74067d6794f81ee12cd2ab4483e4bde1704e1). In the AU3 source tree, we add `PlatformCompatibilityWX.cpp`, which provides definitions for `PlatformCompatibility::GetUserDataDir()` and the rest. I had in mind that the au4 sources could have `PlatformCompatibilityQt.cpp`, which would invoke `GlobalConfiguration::userDataPath()` and other appropriate methods?

igorkorsukov: > I had in mind that the au4 sources could have PlatformCompatibilityQt.cpp, which would invoke GlobalConfiguration::userDataPath() and other appropriate methods?

Just please let's make sure that there is no Qt in the name :) 
The following options are possible:
* PlatformCompatibilityAU4
* PlatformCompatibilityMuse

igorkorsukov (approved):

igorkorsukov (approved):

PR #7101, opened 2024-08-15
7097 gain is actually volume

### LINKED ISSUE #7097: Track volume slider: change tooltip from "gain" to "volume"
The "gain" slider of tracks controls amplification of the waveform _after_ the track's effect stack. In that regard, it actually is a "volume" slider.

saintmatthieu:
Resolves: #7097

* Change UI strings,
* Where the distinction is important, rename methods, variables and comments.

Recommended:

QA:
- [x] Mixer Board slider (including tooltip if any) shows "volume" instead of "gain",
- [x] Changing a track's volume through track or mixer-board slider adds a history item that mentions "volume" and not "gain".
- [x] Track slider tooltip when hovering and when dragging all say "volume"
- [x] Track volume changes in a previous audacity version are preserved when opening the project with this build


libraries/lib-audio-devices/AudioIOBase.h
```diff
@@ -233,7 +233,7 @@ class AUDIO_DEVICES_API AudioIOBase /* not final */
     * methods use whatever device is specified by the preferences.  If a
     * stream *is* running, naturally they manipulate the mixer associated
     * with that stream.  If no mixer is available, output is emulated and
-    * input is stuck at 1.0f (a gain is applied to output samples).
+    * input is stuck at 1.0f (a volume gain is applied to output samples).
```
kryksyh, line 236: is `a volume gain` is intentional here?

libraries/lib-audio-io/AudioIO.h
```diff
@@ -500,7 +500,7 @@ class AUDIO_IO_API AudioIO final
     * methods use whatever device is specified by the preferences.  If a
     * stream *is* running, naturally they manipulate the mixer associated
     * with that stream.  If no mixer is available, output is emulated and
-    * input is stuck at 1.0f (a gain is applied to output samples).
+    * input is stuck at 1.0f (a volume gain is applied to output samples).
```
kryksyh, line 503: and here too `a volume gain` , not sure if this is a typo?

kryksyh (approved):

libraries/lib-audio-devices/AudioIOBase.h
saintmatthieu: Yeah, kind of. Could have been "(volume is applied", but that's also acceptable I think.

Issue #7076, opened 2024-08-09
When adding a realtime effect, the plugin should get opened immediately


LWinterberg:

petersampsonaudacity:
Testing on W11 with @nalin243 's branch test build for this:  audacity-win-3.7.0-alpha-20240818+f038455-x64

This appears to work as intended with the realtime effect dialogbeing opened when it is added to the effect stack.

Adding a second and further realtime effects also opens the dialogs for those, leaving the existing ones open.

dozzzzer:
Tested

PR #7110, opened 2024-08-18
Adding a realtime effect makes the plugin ui open immediately

nalin243:
Resolves:  #7076

Used the RealtimeEffectState to open effect UI as soon as user adds an effect or changes to a new effect. 

Recommended:

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -56,6 +56,22 @@
 
 namespace
 {
+
+   void ToggleUi(
+      AudacityProject& project, RealtimeEffectState& state, wxString sourceName)
+   {
+      const auto ID = state.GetID();
+      const auto effectPlugin = EffectManager::Get().GetEffect(ID);
+
+      if (effectPlugin == nullptr)
+         return;
+
+      auto& effectStateUI = RealtimeEffectStateUI::Get(state);
+
+      effectStateUI.SetTargetName(sourceName);
+      effectStateUI.Toggle(project);
+   }
+
    using namespace MenuRegistry;
    class RealtimeEffectsMenuVisitor final : public Visitor<Traits> {
       wxMenu& mMenu;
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
Testing on W11 with @nalin243 's branch test build for this:  audacity-win-3.7.0-alpha-20240818+f038455-x64

This appears to work as intended with the realtime effect dialogbeing opened when it is added to the effect stack.

Adding a second and further realtime effects also opens the dialogs for those leaving the existing ones open

saintmatthieu:
Looks good, only now we have 
```cpp
         if(effectPlugin == nullptr)
         {
            ///TODO: effect is not available
            return;
         }

         if(!mProject)
            return;
         auto& effectStateUI = RealtimeEffectStateUI::Get(*mEffectState);

         effectStateUI.SetTargetName(mDelegate->GetSourceName());
         effectStateUI.Toggle(*mProject);
```
not once, but three times. Please encapsulate this in a method, called something like `ToggleUi`.

LWinterberg:
Can you make the change matt requested? We're looking to release on the 22nd btw.

nalin243:
@LWinterberg @saintmatthieu 

If you could check and tell me if something needs to be changed that would be appreciated. This works fine but I am not sure if this is the correct way to go about it. 

```
void ToggleUi(wxWeakRef<AudacityProject> project, const std::shared_ptr<RealtimeEffectState> &pState,wxString sourceName)
{
   const auto ID = pState->GetID();
   const auto effectPlugin = EffectManager::Get().GetEffect(ID);

   if(effectPlugin == nullptr)
      return;
               
   auto& effectStateUI = RealtimeEffectStateUI::Get(*pState);

   effectStateUI.SetTargetName(sourceName);
   effectStateUI.Toggle(*project);
}
```

Also where should I put this function? Currently this is just at the top of the file outside any namespace.

saintmatthieu:
Hello @nalin243 , sorry for the late reply.
That looks good to me, feel free to proceed and we might have this in our next patch release :)

saintmatthieu:
@nalin243 I just adopted your suggestion, updated the PR with it and re-targeted to the `audacity3` branch. @kryksyh FYI

kryksyh (approved):

nalin243:
@saintmatthieu Ok, I have been busy lately and couldn't follow up on this.

PR #7123, opened 2024-08-20
[AU4] replaced clip index to id

igorkorsukov:


au4/src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -113,6 +113,16 @@ ClipListItem* ClipsListModel::itemByKey(const trackedit::ClipKey& k) const
     return nullptr;
 }
 
+int ClipsListModel::indexByKey(const trackedit::ClipKey& k) const
+{
+    for (int i = 0; i < m_clipList.size(); ++i) {
+        if (m_clipList.at(i)->clip().key == k) {
+            i;
```
grliszas14, line 120: missing return statement?

igorkorsukov: oops

au4/src/au3wrap/internal/domau3types.h
```diff
@@ -0,0 +1,15 @@
...
+#pragma once
+
+#include "libraries/lib-wave-track/WaveClip.h"
+
+namespace au::au3 {
+struct WaveClipID {
+    uint64_t id = -1;
+
+    WaveClipID(const WaveClip* au3clip)
+        : id(reinterpret_cast<uint64_t>(au3clip)) {}
```
kryksyh, line 13: Since we are using pointers as id, it might make sense to use `using WaveClipID::type = uintptr_t` here for clarity, though it is effectively the same as uint64_t on most platforms.

au4/src/au3wrap/internal/domconverter.cpp
```diff
@@ -22,12 +23,12 @@ au::trackedit::Clip DomConverter::clip(const WaveTrack* waveTrack, const WaveCli
...
     au::trackedit::Clip clip;
     clip.key.trackId = trackId(waveTrack->GetId());
-    clip.key.index = index;
+    clip.key.clipId = WaveClipID(au3clip).id;
     clip.title = wxToSting(au3clip->GetName());
     clip.startTime = au3clip->GetPlayStartTime();
     clip.endTime = au3clip->GetPlayEndTime();
 
-    size_t colorIdx = clip.key.trackId + clip.key.index;
+    size_t colorIdx = clip.key.trackId + clip.key.clipId;
```
kryksyh, line 31: Here, clip ID will have aligned values, so the result might not be as expected.

au4/src/au3wrap/internal/domau3types.h
igorkorsukov: I thought about it, but I don't want to reveal the information that it is a pointer, i.e. that this pointer is a hack and just a current implementation detail

grliszas14 (approved):

au4/src/au3wrap/internal/domconverter.cpp
```diff
@@ -11,23 +12,22 @@ au::trackedit::TrackId DomConverter::trackId(const TrackId& au3trackId)
...
     clip.endTime = au3clip->GetPlayEndTime();
 
-    size_t colorIdx = clip.key.trackId + clip.key.index;
+    //! TODO AU4 Just for tests
+    static std::vector<muse::draw::Color> colors = {
+        muse::draw::Color::fromString("#677CE4"),
+        muse::draw::Color::fromString("#2C79B0"),
+        muse::draw::Color::fromString("#2D9696")
+    };
+    size_t colorIdx = std::llabs(clip.key.trackId + clip.key.clipId);
```
kryksyh, line 30: This would work as long as color count is not a multiple of 4, otherwise all clips would have the same color. But as I understand this is a temporary solution, so I believe this is fine for now.

https://godbolt.org/z/6Pz5jhWdb

PR #7126, opened 2024-08-20
Crash fix on cloud download failure

vsverchinsky:
QA: attempt to sync with remote project when there is no local copy could crash the app when something goes wrong. Should be more or less reproducible with remote project created in 32 bit 3.6.1 and clean `CloudProjects` directory

Recommended:

### DIFF
--- a/libraries/lib-cloud-audiocom/sync/RemoteProjectSnapshot.cpp
+++ b/libraries/lib-cloud-audiocom/sync/RemoteProjectSnapshot.cpp
@@ -381,8 +381,12 @@ void RemoteProjectSnapshot::DownloadBlob(
    }
 
    response->setRequestFinishedCallback(
-      [this, onSuccess = std::move(onSuccess), retries, response](auto)
+      [this, self = weak_from_this(), onSuccess = std::move(onSuccess), retries, response](auto)
       {
+         auto strong = self.lock();
+         if(!strong)
+            return;
+
          mDownloadedBytes.fetch_add(
             response->getBytesAvailable(), std::memory_order_acq_rel);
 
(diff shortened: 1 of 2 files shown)


libraries/lib-cloud-audiocom/sync/RemoteProjectSnapshot.cpp
```diff
@@ -381,8 +381,12 @@ void RemoteProjectSnapshot::DownloadBlob(
    }
 
    response->setRequestFinishedCallback(
-      [this, onSuccess = std::move(onSuccess), retries, response](auto)
+      [this, self = weak_from_this(), onSuccess = std::move(onSuccess), retries, response](auto)
       {
+         auto strong = self.lock();
+         if(!strong)
+            return;
```
kryksyh, line 388: I may be missing something, but shouldn't we call OnFailure here?

kryksyh: I see. Thanks for the explanation.

kryksyh (approved):

dozzzzer:
While testing this PR, I can still reproduce the "Sync Failed" error message after trying to retrieve the project that was previously uploaded using this exact build and then removed from the local CloudProjects directory. However, Audacity doesn't crash after the error message. Is this expected? @vsverchinsky

dozzzzer:
Please ignore my previous comment. In this PR's build I cannot reproduce the sync failed problem anymore.

vsverchinsky:
@dozzzzer Attempt to download project that was corrupted by 32bit app (there should be no local copy) should fail with the message you mentioned. But before this patch it could also crash. It crashes for me in debug version almost in 100% cases. Btw did you try it on Mac or Windows?

dozzzzer:
I've tried both and it only gives me that error message and never crashes.

Issue #7134, opened 2024-08-21
Hide the splash screen after the main window is visible


kryksyh:
### Your idea

Hide the splash screen only after the main window is visible.

### Problem to be solved

Currently, the splash screen is shown briefly, and there is a significant time gap between the splash screen and the showing of the main window

chinakov:
This is great! This has been annoying me for some time, glad to see this fixed.

chinakov:
Tested.

PR #7135, opened 2024-08-21
Hide and fadeout the splash screen only after the main window is visible

kryksyh:
Resolves: #7134

* Wait until the main window is visible before hiding the splash screen
* Fade-out splash screen instead of simply hiding it

To QA: please test this PR on all available platforms, as it may behave differently depending on the system

Recommended:


LWinterberg (approved):

kryksyh:
Need to fix tests

Issue #7124, opened 2024-08-20
"editing can move other clips" doesn't work with multiple tracks selected


LWinterberg:
### Bug description

https://github.com/user-attachments/assets/b3c3aa07-a6b6-4ec1-8994-907429c1e9f4

### Steps to reproduce

1. Have multiple tracks with content
2. Enable editing clips can move other clips pref
3. copy a clip
4. make a selection in both tracks to paste to
5. paste

### Expected behavior

Clips on both tracks are moved to make room and keep in sync

### Actual behavior

Only tracks to which content actually got pasted get moved

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

Windows

### Additional context

Interestingly, there's two previous behaviors:

* 3.0 uses per-track ripple, ie only selected tracks get rippled
* 3.5 uses global ripple, ie all tracks get rippled (same as sync lock)

PR #7146, opened 2024-08-22
Fix `WaveTrack` copy to clipboard

vsverchinsky:
Resolves: #7124 

Paste behaviour change: when pasting all clips and labels to the right from the beginning of selection intrval should move accordingly, if sync lock disabled and "...clips can move" option is enabled

Recommended:

QA: Please check operations that involve track copying with two types of selections - one that ends inside wave clip, second one that has ends with emptiness (it may also be an empty wave interval):
 - [x] Copy/Paste
 - [x] Cut/Paste
 - [x] Split cut/Paste
 - [x] Effect preview (!)
 - [x] Effect "Repeat"

### DIFF
--- a/libraries/lib-channel/Channel.h
+++ b/libraries/lib-channel/Channel.h
@@ -295,6 +295,9 @@ public:
    //! Change start time by given duration
    void ShiftBy(double t) { MoveTo(GetStartTime() + t); }
 
+   //! Shift all intervals that starts after `t0` by `delta` seconds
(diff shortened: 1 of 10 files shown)


saintmatthieu (approved):

chinakov:
Tested, thanks!

vsverchinsky:
@LWinterberg Could you please check add tell if it now behaves as expected, if so I might change implementation a little bit to exclude possible unintended changes

LWinterberg (approved):
looks good now!

PR #7148, opened 2024-08-23
`MenuItemEnabler `s have an optional post-command action

### LINKED ISSUE #2617: Playback should play indefinitely

saintmatthieu:
Resolves: #2617 (Not completely...)

A follow-up PR will enable this new functionality on the `AudioIONotBusyFlag`, which will allow us to choose whether actions that are currently blocked by playback should just automatically stop playback or stop _and_ resume once the command is executed. This, however, will require a big involvement from our designers and may take a while. In the meantime, review of the functionality underneath may begin.

Recommended:

QA: No behavioral change must be observed. In particular,
- [ ] All recording-related commands should still be disabled during playback. Also check that interactions between projects (e.g. playback ongoing in project A ; can I start recording in project B ?) remains unchanged.

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -1750,11 +1750,11 @@ double AudioIO::GetBestRate(bool capturing, bool playing, double sampleRate)
    return supportedRate;
 }
 
-double AudioIO::GetStreamTime()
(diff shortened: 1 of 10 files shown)


saintmatthieu:
Postponed till 4.x ; @kryksyh thank you for your review.

PR #7150, opened 2024-08-23
2617 perpetual playback

### LINKED ISSUE #2617: Playback should play indefinitely

saintmatthieu:
Resolves: #2617

Depends on #7148 and other PR items listed in #2617.

This PR let's audio playback continue after the cursor has reached the end or the stop button is pressed. Only exercising commands that require playback to stop will stop the stream (and maybe resume automatically afterwards).

Recommended:


saintmatthieu:
Postponed till 4.x

saintmatthieu:
Postponed till 4.x

PR #7157, opened 2024-08-26
2617 non blocking playback

### LINKED ISSUE #2617: Playback should play indefinitely

saintmatthieu:
Resolves: #2617

This is the fun part: this PR will make available ideally all user interactions that were so far unavailable during playback. These actions will either just stop playback, or stop and resume once the interaction is complete.
Whether or not playback should resume automatically is for may interactions yet to be decided.

Most of these interactions are menu interactions. Yet some of them are UI interactions, such as stretching, trimming, dragging, etc. All menu interactions are affected. This PR may try to cover all UI interactions too.

### Interactions to consider
@LWinterberg @chinakov @saintmatthieu  below we intend to list all commands (or groups of commands) that are currently blocked by audio playback. Please help complete it and, for each list, use either
* 🟢 to remove the blocking constraint altogether (e.g., "plot spectrum" might not need playback to be stopped at all)
* 🟡 to automatically pause playback while the command is executed and resume afterwards
* 🔴 to automatically stop playback

The coloring should be guided by UX rather than technical feasibility. If feasibility isn't achievable (e.g. "Plot Spectrum" actually requires audio playback to be paused) the color will be downgraded.

(Tip: look in the repo for occurrences of `AudioIONotBusyFlag()` to get probably most of the concerned commands.)

#### Modals
For some modals, it's obvious that playback should stop (like "Open..."), but for others, less. Let's list here modals for which we may want playback to resume automatically once closed.
* Effects 🟡 

#### Misc
* External links (e.g. "Quick Help...", "Get more effects...") 🟢 (Already like this)
* Add track (audio, note, label or time) 🟢 (If safe)
* "Change Playback Device" 🟡 (?)
* "Plot Spectrum" 🟢  

To be continued ...

Recommended:

### DIFF
--- a/libraries/lib-audio-io/AudioIO.cpp
+++ b/libraries/lib-audio-io/AudioIO.cpp
@@ -1750,11 +1750,11 @@ double AudioIO::GetBestRate(bool capturing, bool playing, double sampleRate)
    return supportedRate;
 }
 
-double AudioIO::GetStreamTime()
+double AudioIO::GetStreamTime(bool ignorePlaybackState) const
 {
    // Sequence time readout for the main thread
 
-   if( !IsStreamActive() )
+   if (!ignorePlaybackState && !IsStreamActive())
       return BAD_STREAM_TIME;
 
    return mPlaybackSchedule.GetSequenceTime();
--- a/libraries/lib-audio-io/AudioIO.h
+++ b/libraries/lib-audio-io/AudioIO.h
@@ -565,7 +565,7 @@ public:
     * too. So the returned time should be always between
     * t0 and t1
     */
-   double GetStreamTime();
+   double GetStreamTime(bool ignorePlaybackState = false) const;
 
    static void AudioThread(std::atomic<bool> &finish);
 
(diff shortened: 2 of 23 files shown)


petersampsonaudacity:
Testing on W10 with @saintmatthieu 's test branch build for this:  audacity-win-3.7.0-alpha-20240826+b063896-x64

* External links (e.g. "Quick Help...", "Get more effects...") 🟢 (Already like this)
This is still OK AFAICT (not tested them alll) 

* Add track (audio, note, label or time) 🟢 (If safe)
This works without interrupting play - except for a brief fractional second interrupt
a) You can add a Label track and create labels (but you could always add labels while playing or recording
b) you can add a Time track, but you can't use it until playback stops
c) you can add audio tracks
d) You cannot delete any of these added tracks using the "X" in the TCP
e) you can however use Ctrt+X to delete the tracks

* "Change Playback Device" 🟡 (?)
The Audio Setup button is not active/available during 

* "Plot Spectrum" 🟢  
a) Plot Spectrum works fine during Playback
b) Other analyzers work, they interrupt Playback with the analyzer dialog which must be dealt with/dismissed before playback continues at the pause point

===========================================================

One oddity I have experienced, with Audacity playing in the background I occasionally get interrupts in Playback as I am tying this note.  sometimes it just stops and sometimes it goes back to the beginning of the track - does not seem to be readily reproducible, but it happened several time in the course of writing here.

saintmatthieu:
Postponed till 4.x

PR #7159, opened 2024-08-26
[AU4] changed implementation of show effect

igorkorsukov:


au4/src/effects/builtin/general/GeneralEffectView.qml
```diff
@@ -0,0 +1,33 @@
+import QtQuick
+
+import Muse.UiComponents
+
+import Audacity.Effects
+
+Rectangle {
+
+    property string title: "Amplify"
```
saintmatthieu, line 9: I suppose this property will also have to be read from `viewModel` ?

au4/src/effects/builtin/general/generalviewmodel.cpp
```diff
@@ -0,0 +1,30 @@
...
+void GeneralViewModel::init()
+{
+    IF_ASSERT_FAILED(!instanceId().isEmpty()) {
+        return;
+    }
+
+    //! NOTE Temporary solution
+    m_effect = reinterpret_cast<Effect*>(instanceId().toLongLong());
+
+    LOGDA() << au3::wxToSting(m_effect->GetSymbol().Internal());
```
saintmatthieu: Are we planning to fix this `wxToSting` typo sometime or is it already too late? 😄

saintmatthieu (approved):

au4/src/effects/effectstypes.h
```diff
@@ -1,6 +1,9 @@
 /*
 * Audacity: A Digital Audio Editor
 */
+/*
```
Eism, line 4: no need

au4/src/effects/builtin/general/generalviewmodel.h
Eism, line 1: Can you please explain why we need another class and why we cannot use AbstractEffectModel only?

au4/src/effects/builtin/general/GeneralEffectView.qml
igorkorsukov: yes, that's probably how it will be

au4/src/effects/builtin/general/generalviewmodel.cpp
igorkorsukov: I will fix in a next PR, it's never too late :D

au4/src/effects/builtin/general/generalviewmodel.h
igorkorsukov: These are two different classes with different goals.
We now have two types of views (view models) - specific and general. 
Specific - have a specific UI and may have specific behavior of the view model. They manipulate each parameter in their own way.
General provides a generic UI and manipulates the parameter list. 

Abstract view model is what is common between specific and general. 

```
                         / Specific1 Model (specific view and work with params)
    AbstractEffectModel -  SpecificN Model
                         \ General Model (just list of params)
```

Eism (approved):

PR #7179, opened 2024-08-27
[AU4] Added notify about tracks changes

igorkorsukov:


au4/src/trackedit/internal/au3/au3trackeditproject.cpp
```diff
@@ -67,9 +84,54 @@ muse::async::NotifyList<au::trackedit::Track> Au3TrackeditProject::trackList() c
...
+    auto track = e.mpTrack.lock();
+    if (track) {
+        trackId = DomConverter::trackId(track->GetId());
+    }
+    LOGD() << "trackId: " << trackId << ", type: " << eventTypeToString(e);
+
+    switch (e.mType) {
+    case TrackListEvent::DELETION: {
+        if (e.mExtra == 1) {
+            m_impl->trackReplacing = true;
```
kryksyh, line 111: Don't we need to reset this flag somewhere?

kryksyh (approved):

au4/src/trackedit/internal/au3/au3trackeditproject.cpp
igorkorsukov: you are right! I had a reset, but when I discovered that there was no track in the delete and started to rework it, I lost it

PR #7181, opened 2024-08-27
7169 Reusable AU3 effects: P1 effects

### LINKED ISSUE #7169: Reusable AU3 effects: P1 effects
### Description

Bass & Treble reusable

saintmatthieu:
Resolves: #7169

A rather large PR, but where no new functionality is written, only things are moved around.

Recommended:


libraries/lib-utility/CMakeLists.txt
```diff
@@ -58,6 +60,7 @@ set( SOURCES
    Variant.h
 )
 set( LIBRARIES
+   lib-strings-interface
```
vsverchinsky, line 63: That creates a transitive dependency on `wxBase` for `lib-utility` and comment in CMake file states: "No dependencies on wxWidgets". In that case maybe `lib-math` could be a good place for it.

vsverchinsky (commented):
Added two suggestions. Compilation errors should be fixed.

libraries/lib-utility/CMakeLists.txt
saintmatthieu: PitchName.h/cpp does math things and string things. I thought a bit of math was more tolerable in a string library than the other way round, and the dependency on wxBase seems harmless to me, isn't it? If it isn't, please let me know why, otherwise I'd just remove that part of the comment.

saintmatthieu: Oh well: compilation of PitchName.cpp failed on mac because it doesn't find Internat.h (part of lib-strings). Couldn't find out why but doing as you suggest fixes the build. So be it.

vsverchinsky: The main point is to separate the sheep from the goats: `lib-utility` is more about "domainless" tools, the `lib-math` is domain aware.

vsverchinsky (approved):

Issue #7158, opened 2024-08-26
[AU4] Add new track


grliszas14:
Implement add new track functionality. There is already a flyout when pressing "+ Add new track" btn, make it work so it adds new empty tracks.

PR #7188, opened 2024-08-28
[AU4] Add new empty mono/stereo track

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7158

### DIFF
--- a/au4/src/projectscene/view/clipsview/trackslistclipsmodel.cpp
+++ b/au4/src/projectscene/view/clipsview/trackslistclipsmodel.cpp
@@ -28,11 +28,25 @@ void TracksListClipsModel::load()
 
     m_trackList = prj->trackList();
 
-    //! TODO Subscribe on tracks changed
-
     m_dataSelectedTracks = selectionController()->dataSelectedOnTracks();
     selectionController()->dataSelectedOnTracksChanged().onReceive(this, [this](const std::vector<trackedit::TrackId>& tracks) {
         setDataSelectedTracks(tracks);
+    });
+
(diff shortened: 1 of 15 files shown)


au4/src/projectscene/view/clipsview/trackslistclipsmodel.cpp
```diff
@@ -28,13 +30,27 @@ void TracksListClipsModel::load()
...
     m_trackList = prj->trackList();
 
-    //! TODO Subscribe on tracks changed
-
     m_dataSelectedTracks = selectionController()->dataSelectedOnTracks();
     selectionController()->dataSelectedOnTracksChanged().onReceive(this, [this](const std::vector<trackedit::TrackId>& tracks) {
         setDataSelectedTracks(tracks);
     });
 
+    m_trackList.onItemAdded(m_tracksNotifyReceiver.get(), [this](const trackedit::Track& track) {
```
igorkorsukov, line 38: Why did you decide to use a separate object instead of just passing this?

grliszas14: oh, I did it the same way as it is done in TracksListModel -> should I fix that in both?

igorkorsukov: I asked Elnur... in general, he took this solution from MuseScore, it made sense there, but here and in TracksListModel it doesn't make sense :) You can just pass this :)

grliszas14: fixed

PR #7197, opened 2024-08-28
[AU4] removed EffectInstancesRegister

igorkorsukov:


saintmatthieu (changes requested):
Provisionally blocking this PR for merge because I think it might be going in the wrong direction. Need to clarify first.

saintmatthieu (approved):
Thanks for this effort in clarifying the steps of `DoEffect`.
I cannot vow this is going to yield identical behavior to the previous version (ignoring the TODO's) but we can fine tune this if needed in due time.

PR #7203, opened 2024-08-29
[AU4] Restruct do effect logic. Step 1

igorkorsukov:


saintmatthieu (approved):
Thanks for this effort in clarifying the steps of `DoEffect`.
I cannot vow this is going to yield identical behavior to the previous version (ignoring the TODO's) but we can fine tune this if needed in due time.
Please address commented spelling mistake before merging.

Issue #7196, opened 2024-08-28
Stereo tracks are split into mono when opening a project


dozzzzer:
### Bug description

Originally posted on the forum: https://forum.audacityteam.org/t/error-about-non-aligned-content-version-3-6-1-windows-11-64-bit/115283

### Steps to reproduce

1. Create a new project, add some stereo tracks, edit them
2. Save and close the project.
3. Open the same project again (it might be required to repeat this step a few times)

### Expected behavior

All stereo tracks are opened as stereo tracks; no warning message either.

### Actual behavior

Stereo tracks have become mono.
The User is given the error message:

<img width="600" alt="image" src="https://github.com/user-attachments/assets/c1bf0f43-d6c9-4fb2-b874-93aa7bdef74f">

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, macOS

### Additional context

The bug was introduced in version 3.6.0

dozzzzer:
Bisects to https://github.com/audacity/audacity/commit/9ad724b81ff1af951c52c35657b8489e87a06377

dozzzzer:
Tested

PR #7205, opened 2024-08-29
Fix stereo track loading

kryksyh:
Resolves: [#7196](https://github.com/audacity/audacity/issues/7196)

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -653,8 +653,11 @@ bool WaveTrack::LinkConsistencyFix(const bool doFix)
             next = *TrackList::Channels(this).first.advance(1);
          SetRate(mLegacyRate);
          mLegacyRate = 0;
-         if (next)
+         if (next && next->mLegacyRate > 0)
+         {
+            next->SetRate(next->mLegacyRate);
             next->mLegacyRate = 0;
+         }
          if (mLegacyFormat != undefinedSample)
             WaveTrackData::Get(*this).SetSampleFormat(mLegacyFormat);
          if (next && next->mLegacyFormat != undefinedSample)


kryksyh:
@saintmatthieu there few other places where we didn't consider the stretch factor.
Please check if it also needs fixes (in another pr).

saintmatthieu:
I'm looking at [this way of reproducing](https://github.com/audacity/audacity/issues/7196#issuecomment-2316956776). After the track is split, the clips of each channels should have the exact same stretch ratio, start and end times. Then it is saved. Then it is reopened. What explains the difference we now see in `AreAligned` ?

kryksyh:
> I'm looking at [this way of reproducing](https://github.com/audacity/audacity/issues/7196#issuecomment-2316956776). After the track is split, the clips of each channels should have the exact same stretch ratio, start and end times. Then it is saved. Then it is reopened. What explains the difference we now see in `AreAligned` ?

what is happening is that we are calling SetRate only of the left channel, this leads to slight differences between left and right.

https://github.com/audacity/audacity/blob/e3b721d6730f0671a019391a4b94e764c221b660/libraries/lib-wave-track/WaveTrack.cpp#L656

saintmatthieu:
... and this redoes the clip boundary calculations, I see...
I'm not so fond of the proposed fix because it loosens the equality check because of a numeric precision error we introduce before exporting to XML.
Given your analysis, isn't calling SetRate on the left channel as well not an option? (With a comment linking to this issue.) Or alternatively (but maybe more risky because hard to guarantee it won't have a stupid impact on some other stupid code) doing smth like
```cpp
void WaveTrack::SetRate(double newRate)
{
   if (mRate == newRate) return;
   ...
}
```
?

kryksyh:
@saintmatthieu 
> I'm not so fond of the proposed fix 

Well, I don't believe that comparing floats to be equal here should be allowed. Semantically, what we are doing here is comparing if clips are aligned with sample precision, so the code should reflect that. Otherwise, if you want to compare times bitwise, let's do that, but rename the method to HaveIdenticalIntervals or something.

> calling SetRate on the left channel as well not an option
It makes sense. I checked it, and it works as expected. I can do that, but the first point still stands

saintmatthieu:
> what we are doing here is comparing if clips are aligned with sample precision

`GetPlayStartTime()` and the end-time version of it do round the result to samples all right, but `GetSequenceStartTime()` not. I suppose it's because of the latter that the alignment check fails?
I think we're safer with a bitwise comparison. If the clips get merged into a stereo track, some assumptions of equality of sequence start and end times may be made later.

saintmatthieu (approved):

PR #7215, opened 2024-08-30
Removed assert on no input device

igorkorsukov:


libraries/lib-audio-devices/AudioIOBase.cpp
```diff
@@ -802,7 +802,7 @@ int AudioIOBase::getRecordDevIndex(const wxString &devNameArg)
       // JKC: This ASSERT will happen if you run with no config file
```
kryksyh, line 802: Could you also update comments to reflect changes?

kryksyh (approved):

libraries/lib-audio-devices/AudioIOBase.cpp
igorkorsukov: fixed

PR #7243, opened 2024-09-03
7177 general effect parameter accessor

### LINKED ISSUE #7177: General effect description and parameter accessor interface
### Description

On the AU3 side, we need a C++ interface class offering read/write access to the effect's parameter list.
On the AU4 side, the `GeneralViewModel.h/cpp`  and the `GeneralEffectView.qml` will then use this API to generate a default UI for effects.
This UI will apply to built-in effects, Nyquist effects and Vamp analyzers.
Built-in effects with special UI needs will have their specific `MySpecialEffectViewModel` and QML files.

saintmatthieu:
Resolves: #7177

Recommended:

### DIFF
--- a/libraries/lib-effects/CMakeLists.txt
+++ b/libraries/lib-effects/CMakeLists.txt
@@ -11,6 +11,10 @@ set( SOURCES
    EffectOutputTracks.h
    EffectPlugin.cpp
    EffectPlugin.h
+   EffectWidget.cpp
+   EffectWidget.h
+   EffectWidgetList.cpp
+   EffectWidgetList.h
    LoadEffects.cpp
    LoadEffects.h
    MixAndRender.cpp
@@ -23,6 +27,8 @@ set( SOURCES
    StatefulEffect.h
    StatefulPerTrackEffect.cpp
    StatefulPerTrackEffect.h
+   WidgetListStatefulEffect.cpp
+   WidgetListStatefulEffect.h
 )
 set( LIBRARIES
    lib-command-parameters-interface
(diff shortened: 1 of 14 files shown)


libraries/lib-effects/CMakeLists.txt
```diff
@@ -23,6 +23,12 @@ set( SOURCES
    StatefulEffect.h
    StatefulPerTrackEffect.cpp
    StatefulPerTrackEffect.h
+   EffectWidgetList.cpp
```
saintmatthieu, line 26: reorder

libraries/lib-effects/EffectWidget.h
```diff
@@ -0,0 +1,156 @@
...
+};
+
+template <> struct WidgetValue<WidgetType::ChooseFile>
+{
+   const std::string path;
+};
+
+template <> struct WidgetValue<WidgetType::DblText>
+{
+   const double min;
```
saintmatthieu, line 65: No min/max

libraries/lib-effects/EffectWidget.h
```diff
@@ -0,0 +1,156 @@
...
+   const double max;
+   const double value;
+   const std::optional<int> numSteps;
+};
+
+template <> struct WidgetValue<WidgetType::IntText>
+{
+   const int min;
```
saintmatthieu, line 80: No min/max

libraries/lib-effects/EffectWidgetList.h
```diff
@@ -0,0 +1,53 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  EffectWidgetList.h
+
+  Matthieu Hodgkinson
+
+  Generic interface to describe an effect, list its parameters and
```
saintmatthieu, line 9: Comment somewhat outdated, will refresh.

saintmatthieu:
Change of plan: we are not doing a generic effect view anymore.

PR #7267, opened 2024-09-06
[AU4] Trimming improvements

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7258


au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
```diff
@@ -116,12 +116,32 @@ Item {
                     clipsModel.moveClip(clipItem.key, deltaX, completed)
                 }
 
-                onClipLeftTrimmed: function(deltaX) {
-                    clipsModel.trimLeftClip(clipItem.key, deltaX)
+                onClipLeftTrimmed: function(deltaX, posOnCanvas) {
```
Eism, line 119: I would suggest moving this code into the model (`trimLeftClip` method)

au4/src/projectscene/view/timeline/timelinecontext.cpp
```diff
@@ -13,6 +13,10 @@ static constexpr double ZOOM_MIN = 0.001;
 static constexpr double ZOOM_MAX = 6000000.0;
 static constexpr int PIXELSSTEPSFACTOR = 5;
 
+//! TODO Make better name
+constexpr double INSURE_VISIBLE_GAP_PX(16);
```
Eism, line 17: from the name it looks like this value will be used only in the insureVisible method - I suggest moving these constants to the method and then maybe it will be easier to choose a name

au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipHandles.qml
```diff
@@ -346,6 +362,10 @@ Item {
         }
     }
 
+    Component.onCompleted: {
+        canvas = findParentWithObjectName(this, "clipsView")
```
Eism, line 366: Let's also abandon searching for a parent by name and pass it directly through all the children?
I don't like the search approach because it's hard to control. Someone will delete or just rename the name and everything will break

Eism (approved):

PR #7270, opened 2024-09-06
[AU4] Clip joining

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7255

### DIFF
--- a/au4/src/trackedit/internal/trackeditactionscontroller.cpp
+++ b/au4/src/trackedit/internal/trackeditactionscontroller.cpp
@@ -12,6 +12,7 @@ using namespace muse::actions;
 static const ActionCode COPY_CODE("copy");
 static const ActionCode DELETE_CODE("delete");
 static const ActionCode SPLIT_CODE("split");
+static const ActionCode JOIN_CODE("join");
 
 static const ActionCode CLIP_CUT_CODE("clip-cut");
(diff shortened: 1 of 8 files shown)


au4/src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -237,6 +251,28 @@ void TrackeditActionsController::trackSplitAt(const ActionData& args)
...
+
+    secs_t begin = args.arg<secs_t>(1);
+    secs_t end = args.arg<secs_t>(2);
+    secs_t duration = end - begin;
+
+    for (const auto &trackId : tracksIds ) {
+        trackeditInteraction()->mergeSelectedOnTrack(trackId, begin, end);
+    }
+
+    pushProjectHistoryJoinState(begin, duration);
```
igorkorsukov, line 273: Should be inside trackeditInteraction()->mergeSelectedOnTrack  
The reason is that the action controller is called only for processing of actions... 
But a service method can be called in other cases, in some scenario, from some API extensions, etc.
In this case, all the logic associated with this operation must be executed...
If we ever need to not modify the history in some case, then it is better to do this explicitly, in the service method add a flag "do not change to history"

igorkorsukov (approved):

PR #7275, opened 2024-09-08
Changed quit and close file menu item's names and colocated them to bottom

nalin243:
Resolves: #6644 

Recommended:

### DIFF
--- a/src/menus/FileMenus.cpp
+++ b/src/menus/FileMenus.cpp
@@ -529,7 +529,7 @@ auto FileMenu()
          //   AudioIONotBusyFlag(), wxT("Shift+A") )
       ),
 
-      Section( "Close", Command( wxT("Close"), XXO("&Close"), OnClose,
+      Section( "Close", Command( wxT("Close"), XXO("&Close Project"), OnClose,
             AudioIONotBusyFlag(), wxT("Ctrl+W") )
       ),
 
@@ -558,7 +558,7 @@ auto FileMenu()
          // pull it out
          // and put it in the Audacity menu for us based on its ID.
          /* i18n-hint: (verb) It's item on a menu. */
-         Command( wxT("Exit"), XXO("E&xit"), OnExit,
+         Command( wxT("Exit"), XXO("Quit Audacity"), OnExit,
             AlwaysEnabledFlag, wxT("Ctrl+Q") )
       )
    ) };
--- a/src/menus/FileMenus.cpp
+++ b/src/menus/FileMenus.cpp
@@ -529,10 +529,6 @@ auto FileMenu()
          //   AudioIONotBusyFlag(), wxT("Shift+A") )
       ),
 
-      Section( "Close", Command( wxT("Close"), XXO("&Close Project"), OnClose,
-            AudioIONotBusyFlag(), wxT("Ctrl+W") )
-      ),
-
       Section( "Import-Export",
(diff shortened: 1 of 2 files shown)


kryksyh:
@nalin243 please rebase on the latest master, to get the ci build working.

nalin243:
@kryksyh 

Should I let this be or close and open a new request?

kryksyh:
> @kryksyh
> 
> Should I let this be or close and open a new request?

you can keep this pr 👍

petersampsonaudacity:
Testing on W10 with @nalin243 's latest branch build for this:  audacity-win-3.7.0-alpha-20240908+575662c-x64

image

a)The new wording is implemented properly
b) the two commands are now co-located at the bottom of the File menu

But I'm failing to understand why we have (or need) a separator line between the two commands as they are both very closely related in functionality.

src/menus/FileMenus.cpp
```diff
@@ -553,12 +549,16 @@ auto FileMenu()
          )
       ),
 
+      Section( "Close", Command( wxT("Close"), XXO("&Close Project"), OnClose,
```
LWinterberg, line 552: you don't need the `Section("Close"` bit anymore, you can instead put the `Command(...)` thing into the Exit section.

src/menus/FileMenus.cpp
```diff
@@ -553,12 +549,16 @@ auto FileMenu()
...
+            AudioIONotBusyFlag(), wxT("Ctrl+W") )
+      ),
+
       Section( "Exit",
          // On the Mac, the Exit item doesn't actually go here...wxMac will
          // pull it out
          // and put it in the Audacity menu for us based on its ID.
          /* i18n-hint: (verb) It's item on a menu. */
```
LWinterberg, line 560: That said, watch the comments. If the entire section is relocated to the App menu on macOS, you'll need to do some platform specific code via

```cpp

#ifdef __WXMAC__
  macos code
#else 
  windows/linux code
#endif
```

DavidBailes:
> But I'm failing to understand why we have (or need) a separator line between the two commands as they are both very closely related in functionality.

@petersampsonaudacity 
Standard practice in editors appears to have such commands in separate sections. (Can you find one that doesn't?) Just because two commands both close something, doesn't mean they should be in the same section. Closing a project (perhaps one of many) is quite different from closing the program completely.

src/menus/FileMenus.cpp
nalin243: I am not really familiar with mac but what I am understanding is that if I put both 'Quit Audacity' and 'Close Project' in the same section then to make sure that mac doesn't put 'Close Project' in the app menu as well, they would need to be in different sections. Won't that result in the section dividing line still being present in mac then?

nalin243:
@DavidBailes @LWinterberg 

Should I keep it in the same section or let it be then?

DavidBailes:
> @DavidBailes @LWinterberg
> 
> Should I keep it in the same section or let it be then?

It's up to @LWinterberg , not me.

LWinterberg (approved):
let's leave it like this - the additional separator also makes it slightly more difficult to misclick quit instead of close.

petersampsonaudacity:
I assume this will be merged for 3.7. 0 master and not the impending 3.6.4 ?

kryksyh:
@nalin243 could you rebase your PR on the latest master please?

PR #7277, opened 2024-09-08
Fix CI automated tests

kryksyh:
In continuation of #7215. Replace Warning with Debug, as Warning produces a modal dialog on windows, which not useful for the user, and may interfere with CI tests

Recommended:


LWinterberg (approved):
hooray!

LWinterberg:
also, this probably doesn't need QA

PR #7279, opened 2024-09-09
[AU4] Fix crash on recording

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7272

### DIFF
--- a/au4/src/projectscene/view/clipsview/au3/au3wavepainter.cpp
+++ b/au4/src/projectscene/view/clipsview/au3/au3wavepainter.cpp
@@ -14,6 +14,7 @@
 #include "Envelope.h"
 #include "FrameStatistics.h"
 #include "WaveformScale.h"
+#include "au3wrap/internal/domau3types.h"
 #include "graphics/Color.h"
 
 #include "waveform/WaveBitmapCache.h"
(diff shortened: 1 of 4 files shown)


au4/src/projectscene/view/clipsview/au3/au3wavepainter.cpp
```diff
@@ -616,22 +617,40 @@ void Au3WavePainter::paint(QPainter& painter, const trackedit::ClipKey& clipKey,
...
     }
 
-    std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.clipId);
+    // Iterate through both tracks simultaneously to find matching clip
+    // in the original track and get its counterpart in the pending track.
+    std::optional<uint64_t> clipIdToUpdate;
+    auto origIter = origWaveTrack->Intervals().begin();
+    auto pendingIter = pendingWaveTrack->Intervals().begin();
+
+    for (; origIter != origWaveTrack->Intervals().end() && pendingIter != pendingWaveTrack->Intervals().end(); ++origIter, ++pendingIter) {
```
igorkorsukov, line 637: I think we need to put this logic somewhere so that it can be reused (it's already been used twice in this PR)
Something like this
```
ClipId DomAccessor::findMatchedClip(const WaveTrack* track, const WaveTrack* originTrack, const ClipId& originClipId) const 
{
...
}
```

I would do the implementation like this
```
size_t idx = DomAccessor::findClipIndexById(const WaveTrack* track,   const ClipId& clipId);  
if (idx == muse::nidx) {
   // not found 
   return ClipId();
}

ClipId id = DomAccessor::findClipIdByIndex(const WaveTrack* track,  size_t clipIndex); 

```

findClipIndexById, findClipIdByIndex - need add implementation 

but this is just advice, I don't insist, it seems clearer to me, and the findClipIndexById and findClipIdByIndex methods may still be useful to us

au4/src/record/internal/au3/au3record.cpp
```diff
@@ -241,21 +241,46 @@ void Au3Record::init()
...
+            return;
+        }
+
+        // Iterate through both tracks simultaneously to find matching clip
+        // in the original track and get its counterpart in the pending track.
+        std::optional<uint64_t> clipIdToUpdate;
+        auto origIter = origWaveTrack->Intervals().begin();
+        auto pendingIter = pendingWaveTrack->Intervals().begin();
+
+        for (; origIter != origWaveTrack->Intervals().end() && pendingIter != pendingWaveTrack->Intervals().end(); ++origIter, ++pendingIter) {
```
igorkorsukov, line 260: see https://github.com/audacity/audacity/pull/7279/files#r1756330158

au4/src/record/internal/au3/au3record.cpp
```diff
@@ -241,21 +241,46 @@ void Au3Record::init()
...
+                break;
+            }
+        }
+
+        if (!clipIdToUpdate) {
             return;
         }
 
-        std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.clipId);
+        std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(pendingWaveTrack, clipIdToUpdate.value());
```
igorkorsukov, line 271: let's rename it to pendingClip

au4/src/record/internal/au3/au3record.cpp
```diff
@@ -241,21 +241,46 @@ void Au3Record::init()
...
 
         trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
-        prj->onClipChanged(DomConverter::clip(waveTrack, clip.get()));
+
+        // IMPORTANT: getting pending clip and assigning it to the clip from original track
+        // to make that possible we need to assign orignal clip's ID to the pending clip
+        // so onClipChanged accepts it
+        auto pendingClipWithFakeId = DomConverter::clip(pendingWaveTrack, clip.get());
+        pendingClipWithFakeId.key.clipId = clipKey.clipId;
+        prj->onClipChanged(pendingClipWithFakeId);
```
igorkorsukov, line 283: The idea is clear... but I don't like what's happening here (
I need to think about it...
Let's keep it like this for now

PR #7282, opened 2024-09-09
7180 extract UI agnostic part from nyquist effect

### LINKED ISSUE #7180: Effect migration: reusable Nyquist
### Description

Effect migration: reusable Nyquist

saintmatthieu:
Resolves: #7180

Recommended:

QA:
- [ ] Complete regression test of Nyquist (effect, analyze, command prompt, via macros, maybe even with a faulty .ny script, etc.)


libraries/lib-nyquist-effects/NyquistBase.cpp
```diff
@@ -2744,19 +2744,7 @@ void NyquistBase::OSCallback()
...
-   //       However, on the Mac, the spinning cursor appears during longer
-   //       Nyquist processing and that may cause the user to think Audacity
-   //       has crashed or hung.  In addition, yielding or not on the Mac
-   //       doesn't seem to make much of a difference in execution time.
-   //
-   //       So, yielding on the Mac only...
-#if defined(__WXMAC__)
-   wxYieldIfNeeded();
-#endif
+   YieldIfNeededHook::Call();
```
vsverchinsky, line 2747: Wouldn't `virtual void NyquistBase::OSYield() = 0` be simpler solution in that case?

saintmatthieu: I thought about that, but it feels even more wrong to use a virtual member function, meanwhile the call to yield in fact just needs a stateless function.
I tried something else, similarly to what we did for `PlatformCompatibility`:
``` 
lib-nyquist-effects/NyquistRoutines.h
src/effects/nyquist/NyquistRoutines.cpp
```
with `void YieldIfNeeded()` declared in the DLL and defined in the executable. But that fails on Windows, because building a DLL actually implies linkage, and then the definition isn't found.
It worked for `PlatformCompatibility` because there I could place `PlatformCompatibilityWX.cpp` in the DLL itself, since `wxStandardPaths` is in wxBase, which doesn't turn a library pink.
If this inspires an alternative solution to you, I'd gladly hear it.

vsverchinsky (approved):

saintmatthieu:
@vsverchinsky I understand a little bit better now why fbb6d31 fixes my issue.
The definition of my hook
```cpp
static NyquistBase::GetEffectHook::Scope getEffectHookScope {
   [](const PluginID& path) {
      return std::make_unique<NyquistEffect>(path);
   }
};
```
calls the static method `Assign` on my ` GlobalHook<GetEffectHook, std::unique_ptr<NyquistBase>(const wxString& pluginId)>` template instantiation.
The compiler generated the template, and without the dllexport directive, got my `Scope` to invoke the `Assign` definition of the executable. However, `NyquistBase` and `LoadNyquist` being defined and linked in a DLL, this `Assign` never reached the GlobalHook instance of the DLL.
So basically, without the directive, there exist two definitions of the same type. (In the commit message, the "default impl of `GlobalVariable`" I mentioned doesn't have a reality.)
Does that make sense?

vsverchinsky:
> @vsverchinsky I understand a little bit better now why [fbb6d31](https://github.com/audacity/audacity/commit/fbb6d317602bafc8cfcbbed7c3e10e17cb4ef002) fixes my issue. The definition of my hook
> 
> ```c++
> static NyquistBase::GetEffectHook::Scope getEffectHookScope {
>    [](const PluginID& path) {
>       return std::make_unique<NyquistEffect>(path);
>    }
> };
> ```
> 
> in `src/effects/nyquist/Nyquist.cpp` calls the static method `Assign` on my ` GlobalHook<GetEffectHook, std::unique_ptr<NyquistBase>(const wxString& pluginId)>` template instantiation. The compiler generated the template, and without the dllexport directive, got my `Scope` to invoke the `Assign` definition found in the executable. However, `NyquistBase` and `LoadNyquist` being defined and linked in a DLL, this `Assign` never reached the GlobalHook instance of the DLL. So basically, without the directive, there exist two definitions of the same type. (In the commit message, the "default impl of `GlobalVariable`" I mentioned doesn't have a reality.) Does that make sense?

Yes

vsverchinsky (approved):

Issue #7131, opened 2024-08-21
3.6.2 crash on exit


petersampsonaudacity:
### Bug description
I get a crash when I exit without saving when I have three or more tracks imported.

Testing shows that this is a regression on 3.6.1

### Steps to reproduce (Updated with the steps by @DavidBailes)

1. Add a new stereo track
2. Generate 30s of silence (duration doesn't seem to matter)
3. Quit Audacity

### Expected behavior
No crash

### Actual behavior
![image](https://github.com/user-attachments/assets/3227d79b-534c-4aaf-963a-22a2ef135a36)

### Audacity Version
3.6.2

### Operating system
Windows 10, 11

petersampsonaudacity:
Further testing shows that this bug was present in the 3.6.2 Beta build immediately before the build with the fix for #6929 so is not caused by the fix for that bug,

dozzzzer:
I couldn't reproduce this on my Win10 machine using [the latest 3.6.2 build](https://github.com/audacity/audacity/actions/runs/10471907862).

petersampsonaudacity:
I can still reproduce this on my w10 machine with the very latest 3.6.2 Beta

kryksyh:
@petersampsonaudacity could you attach crashdump file for this crash? It should be in `C:\Users\%USERNAME%\AppData\Local\audacity\crashreports\reports` .

kryksyh:
@petersampsonaudacity unfortunately it doesn't make it clear what's happening. Does it happen for you all the time you repeat the steps? Is it possible for you to record your desktop?

petersampsonaudacity:
@kryksyh 

Yes seems to happen every time

> Is it possible for you to record your desktop?

How do I do that?

petersampsonaudacity:
> Is it possible for you to record your desktop?

@kryksyh 

I can record the app but not the desktop - and only the primary window of the app gets recorded so I don;t even see the File>Exit dialog recorded, nor the crash report.

petersampsonaudacity:
@kryksyh 

OK so here is a desktop recording taken with my camera (had-held, so a bit wobbly)

Actually it's too big to post here, so I put it in my Dropbox - here is a link to the file:
https://www.dropbox.com/s/vwm9exfr8etb9q9/MVI_7817.MP4?dl=0

kryksyh:
@petersampsonaudacity thank you! Out of curiosity, does it happen with a different file format or only with m4a?

[...]

PR #7310, opened 2024-09-10
Improve ProjectWindow lifetime issues with scroll bars, etc.

JoeS209:
Resolves: #7131

Add check for window being deleted in addition to the existing NULL check, before calling related scroll bar functions, etc.

Recommended:


kryksyh (approved):

petersampsonaudacity:
Testing on W10 with @JoeS209 's latest branch build for this:  audacity-win-3.7.0-alpha-20240910+9ae3cfd-x64

I now get **NO** crash reports on editing without saving in any of the three STR use cases in #7131

1. I tested the latest use case I reported yesterday following a Forum report.  Change the temp directory to my large onboard HD D: drive, import a WAV file, Amplify it and then Exit without saving.

2. Bill's simple use case: generate silence and then exit without saving.  Tested both with temp files on the default C: drive and then switched to the HD D: drive.

3. My original use case: import 3 audio files and exit without saving. Tested both with temp files on the default C: drive and then switched to the HD D: drive.

4. I repeated test 1. but without switching the temp files to D:, leaving them on the C: default drive.

All worked fine with no crash dialogs on exit - and no recovery offered on restart and no crash reports in the appdata folder.

**Looks to be fixed to me.**

Thanks Joe for finding the time to work on this so quickly.

chinakov:
Yes, @kryksyh - this issue seems to be fixed within this PR.

Issue #6862, opened 2024-07-12
[AU4] Undo/redo


grliszas14:
Implement undo/redo

grliszas14:
QA bulletpoints:
* check if undo/redo generally works (from within "Edit" menu and shortcuts ctrl+z/ctrl+shift+z)
* so far available actions to test should be copy/paste/split/join (maybe something additional)
* edit menu caption doesn't change yet (it always stays "Undo", "Redo", without changing to "Undo delete" etc.) - task for that created
* there's no history window yet - task created

grliszas14:
@dozzzzer Issue fixed, please have a look

grliszas14:
@dozzzzer both issues fixed

dozzzzer:
@grliszas14 The two crashes are now fixed.
However, there's one more thing: if you open a project with clips,  perform an undoable action (e.g. Split or Copy-Paste) and then Undo, the clip color changes:

https://github.com/user-attachments/assets/b1ed1690-79e9-47a3-b099-8e3dc816d16d

The subsequent attempts to perform an action and Undo it will not change the color unless the project is re-opened.

ianalarina:
Individual clip colouring will be implemented in separate task once designs are ready https://github.com/audacity/audacity/issues/7262

PR #7322, opened 2024-09-13
[AU4] Undo/redo functionality

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6862

### DIFF
--- a/au4/src/project/internal/audacityproject.h
+++ b/au4/src/project/internal/audacityproject.h
@@ -7,6 +7,7 @@
 #include "au3wrap/iau3project.h"
 #include "io/ifilesystem.h"
 #include "projectscene/iprojectviewstatecreator.h"
+#include "trackedit/iprojecthistory.h"
 #include "trackedit/itrackeditproject.h"
 
 namespace au::au3 {
@@ -43,6 +44,7 @@ class Audacity4Project : public IAudacityProject
     muse::Inject<muse::io::IFileSystem> fileSystem;
     muse::Inject<projectscene::IProjectViewStateCreator> viewStateCreator;
     muse::Inject<context::IGlobalContext> globalContext;
+    muse::Inject<au::trackedit::IProjectHistory> projectHistory;
 
 public:
     Audacity4Project();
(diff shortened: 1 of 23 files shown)


au4/src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -628,3 +629,35 @@ void Au3Interaction::pushProjectHistoryJoinState(secs_t start, secs_t duration)
 
     trackeditProject->pushHistoryState(ss.str(), "Join");
 }
+
+void Au3Interaction::undo()
+{
+    auto trackeditProject = globalContext()->currentProject()->trackeditProject();
+    if (!trackeditProject->undoAvailable()) {
+        interactive()->error(muse::trc("undo", "Undo"), std::string("Undo not available"));
```
Eism, line 637: I like how it's done [in MuseScore](https://github.com/musescore/MuseScore/blob/master/src/notation/internal/notationactioncontroller.cpp#L224) - the user can't call undo if there's nothing to undo and therefore this dialog is not needed in MuseScore

Eism: we can fix this in the next PR

au4/src/trackedit/internal/au3/au3trackeditproject.h
```diff
@@ -32,6 +35,11 @@ class Au3TrackeditProject : public ITrackeditProject
 
     secs_t totalTime() const override;
 
+    bool undoAvailable() override;
```
Eism, line 38: It seems that it is better to move this to a separate class and get it from Audacity4Project
it's a different object than simple operations on tracks/clips

au4/src/project/internal/audacityproject.h
```diff
@@ -77,6 +77,8 @@ class Audacity4Project : public IAudacityProject
 
     uintptr_t au3ProjectPtr() const override;
 
+    au::project::IProjectHistoryPtr projectHistory() const override;
```
igorkorsukov, line 80: I think it would be better to take it out of Audacity4Project and make it like other similar services (TrackeditInteraction, SelectionController etc.).

grliszas14: fixed, please have a look

PR #7325, opened 2024-09-13
[AU4] Fixed tempo setting for tracks

Eism:


igorkorsukov (approved):

grliszas14:
What about setting tempo when creating new empty mono/stereo track?
https://github.com/audacity/audacity/blob/master/au4/src/trackedit/internal/au3/au3interaction.cpp#L569
https://github.com/audacity/audacity/blob/master/au4/src/trackedit/internal/au3/au3interaction.cpp#L592

grliszas14 (approved):

Issue #7186, opened 2024-08-28
Close icon for the spectrogram in Multi-View is misaligned


dozzzzer:
Enable Multi-View in TCP and then hover over the upper left corner of the spectrogram.

<img width="520" alt="Screenshot 2024-08-28 at 10 42 49" src="https://github.com/user-attachments/assets/964ab6d0-0065-40a3-8cbd-f5b77694bb21">

dozzzzer:
Tested

PR #7332, opened 2024-09-15
Fix Multi-View close button alignment

Kurtsley:
Resolves: #7186

The Blit function was using the incorrect y coordinate value, leading to the strange positioning. I took the liberty of centering the X and drawing the bevel first.

Recommended:


Kurtsley:
Closing because I used the wrong branch. Rookie mistake.

Issue #6760, opened 2024-06-26
[AU4] Track header - Renaming a track


DilsonsPickles:
[Figma Link](https://www.figma.com/design/k5rS9tnZl60nYmhJhvI2vf/Track-headers?node-id=274-34795)

dozzzzer:
@kryksyh 

1. Upon entering a new name, if you click away or press Esc, the new name should be saved. Currently, it reverts to the previous name.


https://github.com/user-attachments/assets/db7d8eba-11d3-4203-a63c-3e6e29c6ed5d


2. When text is selected, pressing `↑`, `↓` as well as `Shift + ↑`, `Shift + ↓` doesn't work as expected: it simply quits the Edit mode instead.

DilsonsPickles:
@dozzzzer I tested the build this morning and would like ESC to cancel the operation. All other actions however should confirm the name change. Agreed, we are missing the selection/cursor placement shortcuts.

kryksyh:
@dozzzzer I've created a separate task regarding arrow keys

dozzzzer:
Tested

PR #7366, opened 2024-09-19
[AU4] Allow the user to rename tracks

kryksyh:
Resolves: [6760](https://github.com/audacity/audacity/issues/6760)

Recommended:

### DIFF
--- a/au4/src/au3wrap/internal/domaccessor.cpp
+++ b/au4/src/au3wrap/internal/domaccessor.cpp
@@ -8,7 +8,7 @@
 using namespace au::au3;
 using namespace muse;
 
-WaveTrack* DomAccessor::findWaveTrack(AudacityProject& prj, const TrackId& au3trackId)
+Track* DomAccessor::findTrack(AudacityProject& prj, const TrackId& au3trackId)
 {
     Track* track = nullptr;
     TrackList& tracks = TrackList::Get(prj);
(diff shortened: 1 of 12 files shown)


igorkorsukov (approved):

au4/src/au3wrap/internal/domaccessor.cpp
```diff
@@ -8,7 +8,7 @@
 using namespace au::au3;
 using namespace muse;
 
-WaveTrack* DomAccessor::findWaveTrack(AudacityProject& prj, const TrackId& au3trackId)
+Track *DomAccessor::findTrack(AudacityProject &prj, const TrackId &au3trackId)
```
Eism, line 11: please fix code style

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/EditableLabel.qml
```diff
@@ -0,0 +1,98 @@
...
+import QtQuick.Controls
+import QtQuick.Layouts
+
+import Muse.Ui
+import Muse.UiComponents
+
+Item {
+    id: root
+
+    Layout.fillWidth: true
```
Eism, line 11: this approach is bad because we can use the component not inside the Layout, then these settings are not needed and in some cases can interfere
we should use this kind of settings directly in the places where we use the component

au4/src/projectscene/qml/Audacity/ProjectScene/trackspanel/EditableLabel.qml
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    Loader {
+        id: editLoader
+
+        anchors.fill: title
+
+        property bool isEditState: false
+
+        sourceComponent: editLoader.isEditState ? textEditComp : null
```
Eism, line 47: I would suggest to put StyledTextLabel inside Loader and use it instead of null. Then it should be like this: sourceComponent: editLoader.isEditState ? textEditComp : textLabelComp

PR #7368, opened 2024-09-19
7367 repeat last effect

### LINKED ISSUE #7367: Repeat last effect
### Description

* After having successfully applied an effect (at the moment, only Amplify is available), the "Repeat last effect" menu entry should be effective.
* Ctrl+R should be a shortcut to that command.

If technical solutions are readily available,
* The menu item should be greyed out when there is no last effect available.
* The menu item should mention the name of the effect to be repeated, if any.

saintmatthieu:
Resolves: #7367 

Recommended:

QA:
- [ ] The "Repeat last effect" menu item is initially greyed out.
- [ ] After having successfully applied an effect (at the moment, only Amplify is available), it isn't greyed out anymore
- [ ] Ctrl+R should be a shortcut to that command.

### DIFF
--- a/au4/src/effects/effects_base/internal/effectexecutionscenario.cpp
+++ b/au4/src/effects/effects_base/internal/effectexecutionscenario.cpp
@@ -30,8 +30,22 @@ static const int UNDEFINED_FREQUENCY = -1;
 
 muse::Ret EffectExecutionScenario::performEffect(const EffectId& effectId)
 {
-    AudacityProject& project = *reinterpret_cast<::AudacityProject*>(globalContext()->currentProject()->au3ProjectPtr());
+    AudacityProject& project = projectRef();
     return doPerformEffect(project, effectId, 0);
+}
+
+AudacityProject& EffectExecutionScenario::projectRef()
+{
+    return *reinterpret_cast<::AudacityProject*>(globalContext()->currentProject()->au3ProjectPtr());
+}
+
+muse::Ret EffectExecutionScenario::repeatLastProcessor()
+{
+    IF_ASSERT_FAILED(m_lastProcessorId) {
+        return muse::make_ret(Ret::Code::UnknownError);
+    }
+    AudacityProject& project = projectRef();
+    return doPerformEffect(project, *m_lastProcessorId, EffectManager::kConfigured);
 }
 
 muse::Ret EffectExecutionScenario::doPerformEffect(AudacityProject& project, const EffectId& effectId, unsigned flags)
@@ -68,13 +82,6 @@ muse::Ret EffectExecutionScenario::doPerformEffect(AudacityProject& project, con
         }
 
         isSelection = selectedRegion.t1() > selectedRegion.t0();
-        if (flags & EffectManager::kConfigured) {
-            // Don't Select All if repeating Generator Effect
-            if (!(flags & EffectManager::kConfigured)) {
-                //! TODO
-                // DO selectAllIfNone;
-            }
-        }
 
         //! TODO Should we do something if there is no selection and the effect is not a generator? Maybe add a check... or automatically select all...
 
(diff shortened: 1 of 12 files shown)


au4/src/effects/effects_base/internal/effectexecutionscenario.cpp
```diff
@@ -34,6 +34,13 @@ muse::Ret EffectExecutionScenario::performEffect(const EffectId& effectId)
     return doPerformEffect(project, effectId, 0);
 }
 
+muse::Ret EffectExecutionScenario::repeatEffect(const EffectId& effectId)
+{
+    AudacityProject& project = *reinterpret_cast<::AudacityProject*>(
+        globalContext()->currentProject()->au3ProjectPtr());
```
igorkorsukov, line 40: If the project is required several times in a class, it is better to add a helper method to get it, something like this
```
AudacityProject& EffectExecutionScenario::projectRef() const
{
    AudacityProject* project = reinterpret_cast<AudacityProject*>(globalContext()->currentProject()->au3ProjectPtr());
    return *project;
}
```

au4/src/effects/effects_base/internal/effectexecutionscenario.cpp
```diff
@@ -281,48 +281,18 @@ muse::Ret EffectExecutionScenario::doPerformEffect(AudacityProject& project, con
...
-                    commandManager.mRepeatToolFlags =EffectManager::kRepeatNyquistPrompt; // Nyquist Prompt is not configured
-                }
-                break;
-            }
+        //! NOTE Step 8.2 - remember a successful effect
+        if (!(flags & EffectManager::kDontRepeatLast) && effect->GetType() == EffectTypeProcess) {
+            m_processorWasApplied.send(effectId);
         }
 
         //! TODO There should probably be a saving of commandManager here, but it is not explicitly there.
```
igorkorsukov, line 289: we can delete it, now I understand what was happening here..

au4/src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -12,6 +12,17 @@ static const muse::actions::ActionCode EFFECT_OPEN_CODE = "effect-open";
 void EffectsActionsController::init()
 {
     dispatcher()->reg(this, EFFECT_OPEN_CODE, this, &EffectsActionsController::doEffect);
+    dispatcher()->reg(this, "repeat-last-effect", this, &EffectsActionsController::repeatLastEffect);
+    effectExecutionScenario()->processorWasApplied().onReceive(this, [this](const EffectId& processorId) {
+        if (m_lastProcessorId == processorId) {
+            return;
```
igorkorsukov, line 18: It is better not to store m_lastProcessorId in the action controller (or any other data, state)
The ideal controller is stateless, it's just a dispatcher that determines what needs to be called. 

On the other hand, this data or state may be needed in other places (in another controller, in some scenario, in some place, in some API, etc.), where this controller will not be appropriate to use.

For the same reasons, m_lastProcessorId should not be stored in a effectExecutionScenario, because it could be one of many other scripts.

I think the best place to store m_lastProcessorId is the effects provider, but it is the service that we cannot (shouldn't) do without.

The controller can receive lastProcessorId directly from the effects provider or via effectExecutionScenario, it doesn't matter.

au4/src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -22,5 +33,27 @@ void EffectsActionsController::doEffect(const muse::actions::ActionData& args)
...
+    if (code == "repeat-last-effect") {
+        return m_lastProcessorId.has_value();
+    } else {
+        return true;
+    }
+}
+
+muse::async::Notification EffectsActionsController::lastEffectIsNowAvailable() const
+{
+    return m_lastEffectIsNowAvailable;
```
igorkorsukov, line 58: There is no need to add this notification to the controller, for the same reasons described here https://github.com/audacity/audacity/pull/7368/files#r1768242195

au4/src/effects/effects_base/internal/effectsuiactions.cpp
```diff
@@ -17,17 +17,35 @@ const UiActionList EffectsUiActions::m_actions = {
...
+EffectsUiActions::EffectsUiActions(std::shared_ptr<EffectsActionsController> controller)
+    : m_controller{controller}
+{
+}
+
+void EffectsUiActions::init()
+{
+    m_controller->lastEffectIsNowAvailable().onNotify(this, [this]() {
+        m_actionEnabledChanged.send({ "repeat-last-effect" });
+    });
```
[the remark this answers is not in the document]
saintmatthieu: That's good, makes things simpler, too. Just pushed.

au4/src/effects/effects_base/internal/effectsactionscontroller.cpp
saintmatthieu: Yeah, I don't know ... Bear in mind that the signal to the action register should only be sent if:
1. effect type is processor,
2. the effect is applied successfully,
3. the caller says it's ok to enable the "repeat-last-effect" ; e.g. when the effect is used in the context of a macro, it shouldn't make this menu available, i.e. see where `EffectUI::DoEffect` is called in `BatchCommands.cpp`, with the option `EffectManager::kDontRepeatLast`,
4. the signal wasn't issued before (maybe this constraint is only a soft one)

Given these, I couldn't come to a solution that feels better, although I've tried. If you see this as a blocker, would you please check out my branch and give it a try? You may push if you find something satisfactory.

igorkorsukov: What I don't like about the current version is that the "repeat last effect" logic is distributed across several services, and no one has full control over it. i.e. it would be great to define who exactly is responsible for this logic and clearly make it responsible. 

I see at least the following options: 

1. The effects provider is responsible, this means: 
* It "knows" about the very concept of "repeat the last effect"
* It understands the flag kDontRepeatLast
* it knows that only a successfully completed effect like a processor can be repeated

accordingly all this logic should be in the effects provider and its interface should be something like this: 
```
    virtual bool isRepeatLastEffectAvailable() const;
    virtual muse::async::Notification isRepeatLastEffectAvailableChanged() const = 0;
    virtual muse::Ret repeatLastEffect() = 0;
```

2. The effects provider doesn't know anything about the concept of "repeat last effect" and its logic
And then we need to determine who knows who is responsible for this logic. Probably it is EffectExecutionScenario 

It means: 
* We can give the effects provider knowledge of the "last effect executed", i.e. it doesn't matter what type, just the last executed one... and a success flag.
* But since we now see that it is necessary to repeat effects performed only through EffectExecutionScenario, then we can place all the logic only in EffectExecutionScenario. And treat the effect provider only as an executor. Meaning that the effect provider can be used in other scenarios, such as batch processing (something like BatchProcessingScenario ), where there is no need to repeat the effect.

Then there must be something like this: 
```
class IEffectExecutionScenario 
{
...
    virtual bool isRepeatLastEffectAvailable() const;
    virtual muse::async::Notification isRepeatLastEffectAvailableChanged() const = 0;
    virtual muse::Ret repeatLastEffect() = 0;
```

```
muse::Ret EffectExecutionScenario::doPerformEffect(AudacityProject& project, const EffectId& effectId, unsigned flags)
{
...
if (!(flags & EffectManager::kDontRepeatLast) && effect->GetType() == EffectTypeProcess) {
            m_lastProcessorId = effectId;
            m_isRepeatLastEffectAvailableChanged.notify();
        }
}
...
}

bool EffectExecutionScenario::isRepeatLastEffectAvailable()
{
        return m_lastProcessorId.has_value();
}

muse::Ret EffectExecutionScenario::repeatEffect() 
{
       if (!isRepeatLastEffectAvailable()) {
              return muse::make_ret(Ret::Code::NotSupported); 
       }
       ...
       return doPerformEffect(*m_lastProcessorId);
}

```


In both cases, EffectsActionsController does not store anything and does not determine anything itself, it simply knows who can determine the availability of actions, who to call.

saintmatthieu: I do prefer having this in `EffectExectutionScenario`, too, because, yes, it would centralize the whole logic.

au4/src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -22,5 +26,24 @@ void EffectsActionsController::doEffect(const muse::actions::ActionData& args)
 
     muse::String effectId = args.arg<muse::String>(0);
 
-    effectExecutionScenarion()->performEffect(effectId);
+    effectExecutionScenario()->performEffect(effectId);
+}
+
+void EffectsActionsController::repeatLastEffect(const muse::actions::ActionData&)
```
igorkorsukov, line 32: By the way, if the action has no arguments, then you don’t have to specify them

igorkorsukov (approved):

au4/src/effects/effects_base/internal/effectsactionscontroller.cpp
saintmatthieu: Of course ...

Issue #6449, opened 2024-05-17
Play at Speed dialog goes under the bottom edge of the Audacity window


dozzzzer:
1. Double-click the Play-at-Speed slider

Actual result:

![image](https://github.com/audacity/audacity/assets/70063970/cb1c8062-b004-4fe6-85e1-c4b174718433)

Expected result:

![image](https://github.com/audacity/audacity/assets/70063970/bd464220-5d4b-471d-a1dd-f2c0b14b20e7)

petersampsonaudacity:
Testing on W10 with @Kurtsley 's branch build: audacity-win-3.7.0-alpha-20240920+74967ad-x64

The Play-at-Speed dialog no longer obtrudes below the Audacity window which _strictly_ addresses the issue logged by @dozzzzer in the title of this issue  - and this is true for full screen Audacity window and reduced size Audacity window,

However it does obtrude beyond the right of the Audacity window which causes the dialog to be cropped on the right if the reduced size widow is close to the right of the screen.  
![image](https://github.com/user-attachments/assets/72df5886-4171-48e1-a625-cb528ce2778f)

Or in full screen mode if the user moves many toolbars from the upper tooldock to the lower tooldock
![image](https://github.com/user-attachments/assets/b94dbcfc-4d66-4211-b8ed-26680cabcc1c)

PR #7374, opened 2024-09-20
Fix play at speed dialog appearing below window

Kurtsley:
Resolves: #6449 

Dialogs now open centered vertically on the mouse instead of the top left corner.

Recommended:


petersampsonaudacity:
Testing on W10 with @Kurtsley 's branch build: audacity-win-3.7.0-alpha-20240920+74967ad-x64

Th Play-at-Speed dialog no longer obtrudes below the Audacity window which strictly addresses the issue logged by @dozzzzer in #6449  - and this is true for full screen Audacity window and reduced size Audacity window,

However it does obtrude byond the rirht og the 


image

Kurtsley:
Yes, this isn't a perfect fix. The top can also slightly clip the top of the window if you open the recording level dialog.

I would also guess the right side issue already existed and wasn't reported.

I was having trouble getting a reference to the main audacity window from the current doShowDialog function. I would like to leave as much code as possible untouched as I'm not familiar with how the program functions. I guess this is a "better than it was but not permanently fixed" type of fix.

petersampsonaudacity:
TL;DR
For me this is a **_good_** fix _(from a UX  point of view, I can't comment on the code fix)_

It deserves to make it into the app.

===============================================================

> I would also guess the right side issue already existed and wasn't reported.

@Kurtsley  then you would guess absolutely right.  This for example is 2.4.2
image

Note that it doesn't overflow the bottom of the window as in earlier versions of Audacity the Play-at-Speed toolbar was located in the upper tooldock area - Muse relocated it to the lower tooldock several releases ago.

For me its not really a major issue at all as the user can click&drag the dialog window to a better on-screen location.

>Would it be ridiculous to open the dialog boxes in the center of the main window every time?

For myy money I would much prefer to see dialogs remain located close to the thing they relate too and not be in the center of the main window. 

But that is a design decision for the designers and UX folk on the Muse team.

kryksyh:
@Kurtsley thank you for your submission. Could I ask you to improve your PR to address Peter's comment? You can use `wxDisplay` to check current Desktop boundaries, with something like this:
```
wxRect screenRect = wxDisplay(wxDisplay::GetFromPoint(pos)).GetClientArea();
if (pos.x + dlgSize.GetWidth() > screenRect.GetRight()) {
    pos.x = screenRect.GetRight() - dlgSize.GetWidth();
}
if (pos.x < screenRect.GetLeft()) {
    pos.x = screenRect.GetLeft();
}
// and so on
```
Fill free to ask any questions

Kurtsley:
Thank you @kryksyh for pointing me in the right direction. I had to use the topLevelWindow instead of the display for cases where the window isn't maximized.

Since the dialog opens from the top left corner, the top and left checks should never happen. I left them in case this functionality changes in the future.

petersampsonaudacity:
Testing on W10 with @Kurtsley 's latest branch build:  audacity-win-3.7.0-alpha-20240921+840cec6-x64

The dialog is now completely contained with in Audacity Window.

But at default settings the dialog is set hard right and bottom of the window which works but looks a little "forced" to me.  I suspect that it may look better if it were offset a little within the window.

t@kryksyh @dozzzzer  _is this something we should be worried about, or are you happy with this current fix?_

image

petersampsonaudacity:
Note too that other dialogs also can overflow the Audacity Window (with changed window sizes or moved toolbars).  This is the time format setting for the Selection toolbar with a narrowed Audacity window.

image

kryksyh:
@petersampsonaudacity I agree that it looks a bit off. 
@Kurtsley was there any specific reason you decided to use the main window instead of the whole desktop as the reference?

Kurtsley:
@kryksyh
@petersampsonaudacity 
I assumed that all dialog boxes should be kept within the audacity window, but now I see that none of the comments actually mention that. I'll get a final opinion here, then change the PR accordingly.

So for cases like this:
image
When the window isn't maximized, is it okay for the dialogs to leave the window?

Now that I look at it again, I do think it looks fine as long as it doesn't leave the screen.
I can also add a small amount of offset to the dialog if you want to keep it within the audacity window.

petersampsonaudacity:
> When the window isn't maximized, is it okay for the dialogs to leave the window?

In most cases (and certainly default settings on my laptop) that is normally fine - the problem occurs if the Audacity Wwindow is too close to the right edge of the screen so that the the dialog gets truncated on the right:

kryksyh:
@petersampsonaudacity @Kurtsley looks good to me on macOS
image
image

kryksyh (approved):

petersampsonaudacity:
Testing on W10 with @Kurtsley 's latest branch build for this:  audacity-win-3.7.0-alpha-20240922+f2b090e-x64

This now looks good to me.  I tried:
1. full screen window
2. non-full screen window located close to right edge, bottom edge, close to both
3. also tested with the other slider toolbar, the meter toolbars.

@kryksyh  I agree that this looks good to go.

Issue #7028, opened 2024-08-04
Clicking in The Timeline while recording stops the recording


petersampsonaudacity:
### Bug description
Clicking in The Timeline while recording stops the recording - and instead invokes Timeline Quick Play, if you click within the time range of the recording - otherwise it just stops.

This is a regression on 3.0.5 and earlier, introduced in Audacity 3.1.0

### Steps to reproduce
1. Press Record
2. click in the Timeline
3. Observe:recording stops

### Expected behavior
The _**ONLY**_ transport commands/buttons/gestures that should be allowed and enabled while recording are **Stop** and **Pause**

### Actual behavior
Clicking in The Timeline while recording stops the recording 

### Audacity Version
3.6.1 and latest Beta test build for 3.6.1 - and all versions 3.1.0 and onwards

### Operating system
Windows 11 and Mac - but suspect all OS

### Additional context
Originally reported on the Forum by user odododododod

See: https://forum.audacityteam.org/t/audacity-stops-recording-if-you-click-on-timeline/113813

_We have had this regression since 28Oct21 since 3.1.0 was released - i.e for nearly three years_

petersampsonaudacity:
Testing on W10 with @Kurtsley 's branch build for this:  audacity-win-3.7.0-alpha-20240922+e7ffcb0-x64

This regression looks to be properly fixed.

Clicking in the Timeline while recording no longer stops the recording.

petersampsonaudacity:
Testing with @Kurtsley 's latest branch build: audacity-win-3.7.0-alpha-20240922+e7ffcb0-x64

This is now **NOT** fixed again, Timeline Quickplay stops at the end of the selection (see Steps1-5)

1. Generate a 30 second chirp
2.  make a selection from 10-12 seconds
3. click in the timeline at 5 seconds
4. **Observe, playback stops at 12 seconds (the end of the selection)**
5. create a loop from 15-17 seconds
6. Click in the Timeline at 5 seconds
7. Observe: playback ignores the selection and plays on till 17 seconds, then loops 15-17 seconds
8. Stop looping Playback
9. Turn looping off (with the Looping button in Transport toolbar)
10. Click in the Timeline at 5 seconds
11. **Observe, playback stops at 12 seconds (the end of the selection)**

It appears that looping trumps selection - but note that this interaction with loping is not new, not as a result of this fix, it's the same in 3.6.4 and earlier Audacity versions.

Issue #2181, opened 2021-11-23
Timeline Quick-Play stops at end of selection


SteveDaulton:
**Describe the bug**
Clicking on the Timeline should start playback at the clicked time position and continue to the end of the project unless interrupted by the user. This does not happen when there is a selection - playback stops at the end of the selection.
However, if the cursor position is after the clicked position but without a selected region, then playback continues past the cursor position.

**To Reproduce**
1. Open a project with some audio
2. Make a short audio selection somewhere near the middle
3. Click on the Timeline a short distance before the start of the selection
4. Observe that playback stops unexpectedly at the end of the audio selection.

**Expected behavior**
Clicking on the Timeline should start playback at the clicked time position and continue to the end of the project unless interrupted by the user. 

**Additional information (please complete the following information):**
 - OS: Xubuntu 20.04
 - Version Audacity 3.1.3 alpha AppImage

**Additional context**
This is something like the 20th bug created while "improving" Timeline Quick Play. I find it quite disheartening to see my work repeatedly butchered.

petersampsonaudacity:
This is a regression on 3.0.5 and earlier Audacity versions

Confirmed on W10 with 3.1.2 and 3.1.3 Beta-1

petersampsonaudacity:
Testing on W10 with latest 3.2.0 alpha master audacity-win-3.2.0-alpha-20220323+6c64dc4-x64-msvc2022

This no longer appears to be  the case.

@LWinterberg @Penikov  - I suspect this can be closed

LWinterberg:
I can still reproduce it in both 3.1.3 and the latest master on windows 11

petersampsonaudacity:
> I can still reproduce it in both 3.1.3 and the latest master on windows 11

@LWinterberg  Leo, oddly I now cannot repeat what I found earlier (two weeks ago) and this now fails for me with 3.1.3 and the latest 3.2.0 alpha master  audacity-win-3.2.0-alpha-20220405+793cd28-x64-msvc2022

SteveDaulton:
> I can still reproduce it in both 3.1.3 and the latest master on windows 11
Same here on Linux.

tboat:
Hi team, does anyone have an idea when this is scheduled to be fixed? Still experiencing it on 3.4.2 and Windows 11. 

The ability to be able to rapidly compare/listen to the audio of the selection I'm editing with the audio around it is so useful. Personally haven't been able to upgrade for sometime now because of this bug. 

Thanks

[...]

PR #7386, opened 2024-09-22
Fix timeline quick play while recording

Kurtsley:
Resolves: #7028

Disables timeline quick play during active recording.

Recommended:

### DIFF
--- a/src/AdornedRulerPanel.cpp
+++ b/src/AdornedRulerPanel.cpp
@@ -292,7 +292,7 @@ public:
       using namespace RefreshCode;
       auto result = CommonRulerHandle::Release(event, pProject, pParent);
 
-      if (mClicked == Button::Left && !mDragged)
+      if (mClicked == Button::Left && !mDragged && !mParent->mIsRecording)
          StartPlay(*pProject, event.event);
 
       if (!mDragged || 0 != (result & Cancelled))


petersampsonaudacity:
Testing on W10 with @Kurtsley 's branch build for this:  audacity-win-3.7.0-alpha-20240922+e7ffcb0-x64

This regression looks to be properly fixed.

Clicking in the Timeline while recording no longer stops the recording.

kryksyh (approved):

kryksyh:
@Kurtsley thank you for the fix!
@dozzzzer @chinakov  changes are straightforward and I don't expect any side effects, but still to be sure, could you test if the behavior not changed for any other actions related to interactions with the timeline?

petersampsonaudacity:
@kryksyh 

Dmitry, re the Timeline we still have this regression on 3.0.5 that @SteveDaulton logged almost three years ago:

 **Timeline Quick-Play stops at end of selection** #2181

petersampsonaudacity:
Testing with @Kurtsley 's latest branch build: audacity-win-3.7.0-alpha-20240922+e7ffcb0-x64

This is now **NOT** fixed again, Timeline Quickplay stops at the end of the selection (see Steps1-5)

1. Generate a 30 second chirp
2.  make a selection from 10-12 seconds
3. click in the timeline at 5 seconds
4. **Observe, playback stops at 12 seconds (the end of the selection)**
5. create a loop from 15-17 seconds
6. Click in the Timeline at 5 seconds
7. Observe: playback ignores the selection and plays on till 17 seconds, then loops 15-17 seconds
8. Stop looping Playback
9. Turn looping off (with the Looping button in Transport toolbar)
10. Click in the Timeline at 5 seconds
11. **Observe, playback stops at 12 seconds (the end of the selection)**

It appears that looping trumps selection (more accurately active looping trumps selection) in this regard - but note that this interaction with loping is not new, not as a result of this fix, it's the same in 3.6.4 and earlier Audacity versions.

kryksyh:
@petersampsonaudacity I believe that this PR is only to fix #7028, and it looks fixed to me, tested on linux and macOS.
Fix for #2181 needs to be submitted as a separate PR as I anticipate it would be more involved than a one-liner.

petersampsonaudacity:
> I believe that this PR is only to fix #7028, and it looks fixed to me, tested on linux and macOS.

@kryksyh   OOPS yes sorry - and yes this fix for #7028 looks good to me 

_I'll transfer this latest interesting STR to #2181 - and then tidy up this thread._

PR #7391, opened 2024-09-23
7390 au4 amplify slider bug

### LINKED ISSUE #7390: Amplify bug
### Description

When opening the UI, the slider is at position 0 but the value shows whatever default gain was computed to reach 0dB after amplification.
Resultingly, once moving the slider, the amplification value jumps from that default value to 0 (+/- the amount you moved it left/right).

saintmatthieu:
Resolves: #7390

Recommended:

QA:
- [ ] Initial position of slider matches needed amplification to reach 0dB output amplitude,
- [ ] Slider behaves normally when moved left/right,
- [ ] Slider responds to textual update of the amplification value.

### DIFF
--- a/au4/src/effects/builtin/amplify/AmplifyView.qml
+++ b/au4/src/effects/builtin/amplify/AmplifyView.qml
@@ -18,10 +18,12 @@ EffectBase {
 
     AmplifyViewModel {
         id: amplify
+        onAmpChanged: slider.value = amp
     }
 
     Component.onCompleted: {
         amplify.init()
+        slider.value = amplify.amp
     }
 
     Column {
@@ -59,9 +61,9 @@ EffectBase {
         }
 
         StyledSlider {
+            id: slider
             width: parent.width
 
-            value: amplify.amp
             to: amplify.ampMax
             from: amplify.ampMin
             stepSize: 0.1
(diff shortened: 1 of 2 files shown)


au4/src/effects/builtin/amplify/AmplifyView.qml
```diff
@@ -22,6 +22,7 @@ EffectBase {
 
     Component.onCompleted: {
         amplify.init()
+        slider.value = amplify.amp
```
igorkorsukov, line 25: In Qml, such assignment breaks the property binding
https://doc.qt.io/qt-6/qtqml-syntax-propertybinding.html#creating-property-bindings-from-javascript 

```
StyledSlider {
            value: amplify.amp
```

i.e. this bing of properties will stop working, it will be broken, we need to keep this in mind.

so we need to clearly define what we are doing:
* We initialize the initial value of the slider 
* if we move the slider, then it will simply show its current value - here it is ok
* if we change the value in another way, outside the slider, for example through the input field, then the slider will not react (the binding is broken), we need to explicitly update the slider value

Another option
* We have property binding, i.e. no matter how the property changes, the slider will react
* We need to notify about the property change at the beginning so that the initial value is read 


Now the second option is implemented. But there is some problem with the slider (

So, we can do the first option, but then we need to add an explicit update of the slider value when the amp changes.

```
    AmplifyViewModel {
        id: amplify

        onAmpChanged: {
            slider.value = amplify.amp
        }
    }
```

and remove the `value: amplify.amp` that it was not misleading (the binding is broken, this binding will not work)

igorkorsukov (approved):

PR #7395, opened 2024-09-23
[AU4] added update title of repeat last effect

igorkorsukov:


au4/src/effects/effects_base/internal/effectsuiactions.cpp
```diff
@@ -31,11 +36,37 @@ EffectsUiActions::EffectsUiActions(std::shared_ptr<EffectsActionsController> con
 {
 }
 
+void EffectsUiActions::init()
+{
+    effectExecutionScenario()->lastProcessorIsNowAvailable().onNotify(this, [this]() {
```
saintmatthieu, line 41: This notification is issued at most once, so that when another effect is later applied, the menu item's title won't get updated anymore, staying stuck at e.g. "Repeat Amplify".
I have copied the missing piece and improvements from this branch onto my original branch, I suggest we resume from there. Thanks for your contribution!

PR #7408, opened 2024-09-25
normalize line endings to LF

LWinterberg:
This allows locale/update_po_files.sh to be ran on WSL even if development happens on Windows. It may break compatibility with windows text editors which don't do LF, but I don't know any off the top of my head (Notepad++ and VSCode handle it just fine anyway)

Let's hope CI is kind to it too


kryksyh (approved):

LWinterberg:
CI apparently doesn't care about this at all, so nothing to test for QA

Issue #7326, opened 2024-09-13
[AU4] Fix crash when pasting stereo clips to mono tracks


Eism:
### Description

Need to process error and show message
<img width="302" alt="Screenshot 2024-09-13 at 1 59 55 PM" src="https://github.com/user-attachments/assets/f5375447-68dc-4d0c-89eb-f04367dc270f">

grliszas14:
QA bulletpoints:
* check if pasting mono clip into stereo track is possible
* check if pasting stereo clip into mono track results in error dialog
* I did some code-reworking here so it may be advised to play with copy-paste a bit more to see if there is no regression
* check if this one is fixed too: https://github.com/audacity/audacity/issues/7378

dozzzzer:
Tested

PR #7440, opened 2024-10-01
[AU4] Mono/stereo clips paste improvements

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7326

### DIFF
--- a/au4/src/trackedit/internal/au3/au3interaction.cpp
+++ b/au4/src/trackedit/internal/au3/au3interaction.cpp
@@ -19,6 +19,7 @@
 #include "trackedit/dom/track.h"
 
 #include "log.h"
+#include "trackediterrors.h"
 #include "translation.h"
 
 using namespace au::trackedit;
@@ -30,7 +31,7 @@ AudacityProject& Au3Interaction::projectRef() const
     return *project;
 }
 
-bool Au3Interaction::pasteIntoNewTrack()
+muse::Ret Au3Interaction::pasteIntoNewTrack()
 {
     auto& project = projectRef();
     auto& tracks = ::TrackList::Get(project);
@@ -51,7 +52,7 @@ bool Au3Interaction::pasteIntoNewTrack()
     }
     selectionController()->setSelectedTrack(DomConverter::trackId(pFirstNewTrack->GetId()));
 
-    return true;
+    return muse::make_ok();
 }
 
 ::Track::Holder Au3Interaction::createNewTrackAndPaste(std::shared_ptr<::Track> track, ::TrackList& list, secs_t begin)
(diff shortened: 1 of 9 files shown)


au4/src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -109,6 +109,13 @@ bool Au3Interaction::canPasteClips(const std::vector<TrackId>& dstTracksIds, sec
             //! TODO AU4: show dialog
             return false;
         }
+
+        if (dstWaveTrack->NChannels() == 1 && data[i].track.get()->NChannels() == 2) {
+            interactive()->error(muse::trc("message", "Message"), std::string(
+                "The content you are trying to paste will span across more tracks than you"
+                " currently have available. Add more tracks and try again."));
+            return false;
```
igorkorsukov, line 117: It is better not to use interactivity in services without extreme necessity.
It is better to return Ret, and somewhere higher, at the ViewModel level, show a dialog to the user...
Because this method (like other service methods) can be used in different scenarios, some of which may not involve a dialog...
Or simply, if we want to write a unit test - this dialog will be a problem

au4/src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -109,6 +109,13 @@ bool Au3Interaction::canPasteClips(const std::vector<TrackId>& dstTracksIds, sec
             //! TODO AU4: show dialog
```
igorkorsukov, line 109: There is LOGDA above, please replace it with LOGD (LOGDA should be used only for current debugging of current work)

au4/src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -328,7 +335,15 @@ bool Au3Interaction::pasteFromClipboard(secs_t begin, TrackId destinationTrackId
...
 
-        dstWaveTrack->Paste(begin, *clipboard()->trackData()[i].track);
+        if (clipboard()->trackData()[i].track.get()->NChannels() == 1 && dstWaveTrack->NChannels() == 2) {
+            // When the source is mono, may paste its only channel
+            // repeatedly into a stereo track
+            const auto pastedTrack = std::static_pointer_cast<WaveTrack>(clipboard()->trackData()[i].track);
+            pastedTrack->MonoToStereo();
+            dstWaveTrack->Paste(begin, *pastedTrack);
+        } else {
+            dstWaveTrack->Paste(begin, *clipboard()->trackData()[i].track);
```
igorkorsukov, line 345: Please try to use the `at` method to get the element to read, not the `[]` operator. The `at` method definitely returns only a constant reference, we clearly see that the element will not be changed, and there is a check for access boundaries.

And in this case I think it would be better to add a method something like `trackData(size_t i) const`

au4/src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -338,7 +339,10 @@ void TrackeditActionsController::paste()
     TrackId selectedTrackId = selectionController()->selectedTrack();
 
     if (!tracks.empty() && selectedStartTime >= 0) {
-        trackeditInteraction()->pasteFromClipboard(selectedStartTime, selectedTrackId);
+        auto ret = trackeditInteraction()->pasteFromClipboard(selectedStartTime, selectedTrackId);
+        if (ret.code() == static_cast<int>(trackedit::Err::StereoClipIntoMonoTrack)) {
```
igorkorsukov, line 343: Are we deliberately not showing other types of errors?

grliszas14: I did it on purpose at first but I checked AU3 behaviour now and I guess we should show them all, will fix

igorkorsukov (approved):

Issue #7293, opened 2024-09-09
Pasting into a new track moves clips in other tracks


billw58:
### Steps to reproduce

1. Start Audacity 3.6.2
2. Tracks Behaviors > [X] Editing a clip can move other clips
3. Generate something
4. Cut it into a few clips
5. Select a range and Copy
6. New Audio Track
7. Click somewhere in the new track
8. Paste

### Expected behavior

The clip is pasted into the new track and the audio in the first track does not move.

### Actual behavior

Clips in the first track move.

### Audacity Version

latest stable version (from audacityteam.org/download)

### Operating system

macOS (Peter: and Windows)

### Additional context

Version 3.6.1 behaves properly. To see this happen you need to cut the first track into clips and there needs to be clips (in the first track) to the right of the insert point.

DavidBailes:
> To see this happen you need to cut the first track into clips and there needs to be clips (in the first track) to the right of the insert point.

Alternatively, just create one clip in the first track, and move it to the right. Position the edit cursor in the second track before the start of the clip in the first track.

LWinterberg:
yes, this setting essentially is a global ripple (as opposed to per-track ripple) now. Where did you run into this issue, bill?

DavidBailes:
> yes, this setting essentially is a global ripple (as opposed to per-track ripple) now.

The setting is a global ripple for pasting, but per track for all other operations, such a delete, cut, generate a clip.
Very confusing for users.

billw58:
> Where did you run into this issue, bill?
Actually I was testing for something else. I was surprised when it happened because Audacity never did this before.
If this is a deliberate change, I think it's a mistake.
What uses cases were you thinking of when you made this change?
I'm sure I could come up with workflows or use cases where the user want clips to move when editing in a track, but doesn't want clips in other tracks to move when editing in a different track.

billw58:
Is this the fix for Issue #7146 ?
If I understand correctly (looking at your video), you want tracks to ripple where there is a selection in multiple tracks even though the paste happens in one track. But what about the case where there is a selection in only one track?

DavidBailes:
> Testing on W10 shows that this is a regression on 3.6.1 ad all earlier versions of Audacity - and thus unexpected behavior for the experienced user
> 
> _@DavidBailes and that is why it is currently undocumented. If it remains as a new ripple "feature" rather than a "bug" then it should be documented. But for my money this is a true bug, a regression bug._

@petersampsonaudacity .
[...]

PR #7448, opened 2024-10-02
Revert global ripple for paste

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/7293

Problem:
When the preference "Track Behaviors: Editing a clip can move other clips" is set, the commit aa12de91 changed the behaviour of paste from per track ripple to global ripple. Issues with this change:
1. A user no longer has the option of per track ripple for pasting.
2. Before the change, a user could already paste with gobal ripple by enabling synchronized tracks.
3. Paste became inconsistent with all the other actions which could move other clips, which remained per track ripple.

Fix:
Revert commit aa12de9184ced9be0d6c7667bc547ad939cd8340.

Recommended:

### DIFF
--- a/libraries/lib-channel/Channel.h
+++ b/libraries/lib-channel/Channel.h
@@ -295,9 +295,6 @@ public:
    //! Change start time by given duration
    void ShiftBy(double t) { MoveTo(GetStartTime() + t); }
 
-   //! Shift all intervals that starts after `t0` by `delta` seconds
-   virtual void ShiftBy(double t0, double delta) = 0;
-
    //! Change start time to given time point
    virtual void MoveTo(double o) = 0;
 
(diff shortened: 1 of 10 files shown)


petersampsonaudacity:
Testing on W10 with @DavidBailes 's test branch build for this:  audacity-win-3.7.0-alpha-20241002+ea208cf-x64

This appears to have properly restored to the 3.6.1 and earlier behavior, before this regression occurred.

1. With "Editing a clip can move other clips" turned off (default) with insufficient space fore the paste, it is blocked with an appropriate error message.

2. With "Editing a clip can move other clips" turned on and sync-locking turned OFF the a past ripple only occurs in the track that is being pasted into.

3. With "Editing a clip can move other clips" turned on and sync-locking turned ON then a past ripple occurs in all tacks in the  Sync-Locked Track Group (SLTG) that the track that is being pasted into belongs.  No other SLTG in the project experiences the ripple.

DavidBailes:
Note that this PR will cause the reintroduction of the issue: https://github.com/audacity/audacity/issues/7124

DavidBailes:
Will submit another PR.

Issue #7357, opened 2024-09-18
[AU4] Clip can vanish on trimming to minimum width multiple times


grliszas14:
Try zooming in once you've set the clip to it's minimum and trimming it down again. After 2-3 times the clip will vanish again

chinakov:
Works perfectly until you get into increments of 0.000020 seconds, which I think is as good as it gets. Thank you!

PR #7449, opened 2024-10-02
[AU4] Fix disappearing clips on trim

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7357


au4/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
```diff
@@ -67,7 +67,7 @@ Item {
                 property QtObject clipItem: model.item
 
                 height: parent.height
-                width: clipItem.width
+                width: clipItem.width > 3 ? clipItem.width : 3
```
igorkorsukov, line 70: better to use `Math.max(3, clipItem.width)`

igorkorsukov (approved):

PR #7457, opened 2024-10-03
Revert global ripple for paste

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/7293

Problem:
In commit aa12de9, which was part of a PR to fix issue #7124, all other tracks, rather than all other selected tracks, were potentially moved, thus introducing a global ripple.

Fix:
Move only other selected tracks.

Recommended:

### DIFF
--- a/src/menus/EditMenus.cpp
+++ b/src/menus/EditMenus.cpp
@@ -705,10 +705,10 @@ void OnPaste(const CommandContext &context)
       {
          //Special case when pasting without sync lock and
          //"...move other clips" option is 
-         //Also shift all intervals in all other tracks that
+         //Also shift all intervals in all other selected tracks that
          //starts after t0
          const auto offset = srcTracks->GetEndTime() - (t1 - t0);
-         for(auto track : tracks.Any<Track>())
+         for(auto track : tracks.Selected<Track>())
          {
             const auto it = std::find_if(correspondence.begin(), correspondence.end(),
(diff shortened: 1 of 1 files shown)


petersampsonaudacity:
Testing on W10 with @DavidBailes 's revised test branch build for this:  audacity-win-3.7.0-alpha-20241003+1c7e683-x64

This also appears to have properly restored to the 3.6.1 and earlier behavior, before this regression occurred.

1. With "**Editing a clip can move other clips**" turned **off** (default) with insufficient space fore the paste, it is blocked with an appropriate error message.

2. With "**Editing a clip can move other clips**" turned **on** and **sync-locking** turned **OFF** the a paste ripple only occurs in the track that is being pasted into.

3. With "**Editing a clip can move other clips**" turned **on** and **sync-locking** turned **ON** then a paste ripple occurs in all tracks in the  Sync-Locked Track Group (SLTG) that the track that is being pasted into belongs.  No other SLTG in the project experiences the ripple.  This is what sync-locking is for.

AND

4. This does **_NOT_** reintroduce issue #7124

grliszas14 (approved):

ismithbwr:
I'm running 3.6.4 and the global ripple on paste is still here.
"Editing a clip can move other clips" turned on and sync-locking turned off still results in other unselected tracks having their clips move when a paste is made.

PR #7482, opened 2024-10-08
7369 au4 tone generator

### LINKED ISSUE #7369: [AU4] Tone generator
### Description

Implement the UI and interactions of the Tone generator in AU4. Should behave like AU3's.

### LINKED ISSUE #7509: [AU4] Selection over clips remains active if you click in a clip or anywhere in the project
### Description

https://github.com/user-attachments/assets/af6f800f-5a30-4e90-901e-60a222b97649

saintmatthieu:
Resolves: #7369
May resolve #7509 

Recommended:

### QA:
Note that the tone generator is still to be found under **Effects > Built-in > Tone**. This must be addressed in a follow-up PR.

The UI is still unbalanced ; for now, please ignore as long as the functionality is here.

An effort was made to make it on-par with AU3's tone generator. Please use your AU3 tone-generator regression test plan and report missing bits.

Some things to make sure of (not exhaustive):
- [x] Generate on empty project creates new track
- [x] When no track is selected, generating also creates a new track.
- [x] Generate over existing clip replaces selection with tone.
- [x] If there isn't a time selection when invoking the effect, the duration format is `hh::mm::ss + milliseconds`.
- [x] If there is a time selection when invoking the effect, the duration format is `hh:mm:ss + samples`,
- [x] Generating when there is not enough room shows the expected error dialog,

### DIFF
--- a/au3/libraries/lib-exceptions/AudacityException.h
+++ b/au3/libraries/lib-exceptions/AudacityException.h
@@ -76,6 +76,7 @@ protected:
 
    MessageBoxException( const MessageBoxException& );
 
+public:
    //! %Format the error message for this exception.
    virtual TranslatableString ErrorMessage() const = 0;
    virtual wxString ErrorHelpUrl() const { return helpUrl; };
@@ -220,7 +221,7 @@ noexcept(
    catch ( AudacityException &e ) {
    #ifndef UNCAUGHT_EXCEPTIONS_UNAVAILABLE
       const auto uncaughtExceptionsCount = std::uncaught_exceptions();
-   #endif  
+   #endif
       auto end = finally( [&]()
       noexcept(noexcept(
          std::function<void(AudacityException*)>{
@@ -230,9 +231,9 @@ noexcept(
          // other exception object.
       #ifdef UNCAUGHT_EXCEPTIONS_UNAVAILABLE
          if (!std::uncaught_exception()) {
-      #else         
+      #else
          if (uncaughtExceptionsCount >= std::uncaught_exceptions()) {
-      #endif      
+      #endif
             auto pException = std::current_exception(); // This points to e
             AudacityException::EnqueueAction(
                pException, std::move(delayedHandler));
--- a/au3/libraries/lib-track/Track.cpp
+++ b/au3/libraries/lib-track/Track.cpp
@@ -439,9 +439,12 @@ void TrackList::DeletionEvent(std::weak_ptr<Track> node, bool duringReplace)
       { TrackListEvent::DELETION, std::move(node), duringReplace ? 1 : 0 });
 }
 
-void TrackList::AdditionEvent(TrackNodePointer node)
+void TrackList::AdditionEvent(TrackNodePointer node, EventPublicationSynchrony synchrony)
 {
-   QueueEvent({ TrackListEvent::ADDITION, *node });
+   if (synchrony == EventPublicationSynchrony::Synchronous)
+      Publish({TrackListEvent::ADDITION, *node});
+   else
+      QueueEvent({TrackListEvent::ADDITION, *node});
 }
 
 void TrackList::ResizingEvent(TrackNodePointer node)
@@ -482,7 +485,7 @@ void TrackList::Insert(const Track* before,
 
    if(before == nullptr)
    {
-      Add(pSrc, assignIds);
+      Add(pSrc, assignIds ? DoAssignId::Yes : DoAssignId::No);
       return;
    }
 
@@ -493,7 +496,7 @@ void TrackList::Insert(const Track* before,
          arr.push_back(pSrc.get());
       arr.push_back(track);
    }
-   Add(pSrc, assignIds);
+   Add(pSrc, assignIds ? DoAssignId::Yes : DoAssignId::No);
    Permute(arr);
 }
 
@@ -533,11 +536,13 @@ Track *TrackList::DoAddToHead(const std::shared_ptr<Track> &t)
    pTrack->SetOwner(shared_from_this(), n);
    pTrack->SetId( TrackId{ ++sCounter } );
    RecalcPositions(n);
-   AdditionEvent(n);
+   AdditionEvent(n, EventPublicationSynchrony::Asynchronous);
    return front().get();
 }
 
-Track *TrackList::DoAdd(const std::shared_ptr<Track> &t, bool assignIds)
+Track* TrackList::DoAdd(
+   const std::shared_ptr<Track>& t, DoAssignId assignId,
+   EventPublicationSynchrony synchrony)
 {
    if (!ListOfTracks::empty()) {
       auto &pLast = *ListOfTracks::rbegin();
@@ -550,10 +555,10 @@ Track *TrackList::DoAdd(const std::shared_ptr<Track> &t, bool assignIds)
    auto n = getPrev( getEnd() );
 
    t->SetOwner(shared_from_this(), n);
-   if (mAssignsIds && assignIds)
+   if (mAssignsIds && assignId == DoAssignId::Yes)
       t->SetId(TrackId{ ++sCounter });
    RecalcPositions(n);
-   AdditionEvent(n);
+   AdditionEvent(n, synchrony);
    return back().get();
 }
 
@@ -577,7 +582,7 @@ Track::Holder TrackList::ReplaceOne(Track &t, TrackList &&with)
    pTrack->SetId(save->GetId());
    RecalcPositions(node);
    DeletionEvent(save, true);
-   AdditionEvent(node);
+   AdditionEvent(node, EventPublicationSynchrony::Asynchronous);
    return save;
 }
 
@@ -879,7 +884,7 @@ void TrackList::Append(TrackList &&list, bool assignIds)
    while (iter != end) {
       auto pTrack = *iter;
       iter = list.erase(iter);
-      this->Add(pTrack, assignIds);
+      this->Add(pTrack, assignIds ? DoAssignId::Yes : DoAssignId::No);
    }
 }
 
--- a/au3/libraries/lib-track/Track.h
+++ b/au3/libraries/lib-track/Track.h
@@ -865,6 +865,12 @@ class TRACK_API TrackList final
    void clear() = delete;
 
  public:
+   // Whether TrackList assigns an ID to the track or not.
+   enum class DoAssignId { Yes, No };
+   // Whether the resulting `TrackListEvent::ADDITION` should be called
+   // synchronously or deferred.
+   enum class EventPublicationSynchrony { Synchronous, Asynchronous };
+
    static TrackList &Get( AudacityProject &project );
    static const TrackList &Get( const AudacityProject &project );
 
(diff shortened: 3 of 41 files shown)


au4/src/au3wrap/internal/au3project.cpp
```diff
@@ -50,6 +51,17 @@ std::shared_ptr<Au3Project> Au3Project::create()
 
 bool Au3Project::load(const muse::io::path_t& filePath)
 {
+    TrackList& tracks = TrackList::Get(m_data->projectRef());
+    mTrackListSubstription = tracks.Subscribe([this](const TrackListEvent& event)
+    {
+        if (event.mType == TrackListEvent::ADDITION) {
+            const auto tempo = ProjectTimeSignature::Get(m_data->projectRef()).GetTempo();
+            if (const auto track = event.mpTrack.lock()) {
+                DoProjectTempoChange(*track, tempo);
```
igorkorsukov, line 60: We already do it here https://github.com/audacity/audacity/blob/master/au4/src/trackedit/internal/au3/au3trackeditproject.cpp#L103

au4/src/au3wrap/internal/au3project.h
```diff
@@ -6,13 +6,17 @@
...
+#include "modularity/ioc.h"
+#include "context/iglobalcontext.h"
+#include "libraries/lib-utility/Observer.h"
 
 namespace au::au3 {
 struct Au3ProjectData;
 class Au3Project : public IAu3Project
 {
-public:
+    INJECT(au::context::IGlobalContext, globalContext);
```
igorkorsukov, line 17: seems, this is not need

au4/src/effects/builtin/tonegen/ToneView.qml
```diff
@@ -1,7 +1,130 @@
...
+            StyledTextLabel {
+                anchors.verticalCenter: parent.verticalCenter
+                text: qsTrc("effects/tone", "Waveform:")
+            }
+
+            ComboBox {
+                anchors.verticalCenter: parent.verticalCenter
+                width: 80
+
+                model: ["Sine", "Square", "Sawtooth", "Square, no alias", "Triangle"]
```
igorkorsukov, line 48: It is better to return this list (model) from the view model and somehow coordinate it with the enumerator.

au4/src/effects/builtin/tonegen/ToneView.qml
```diff
@@ -1,7 +1,130 @@
...
+
+        Row {
+            anchors.horizontalCenter: parent.horizontalCenter
+
+            spacing: 4
+
+            StyledTextLabel {
+                anchors.verticalCenter: parent.verticalCenter
+
+                text: qsTrc("effects/tone", "Frequency (Hz):")
```
igorkorsukov, line 65: I think there is enough context for the translation here, just `effects`, there is no need to make the context too granular

au4/src/effects/builtin/tonegen/toneeffect.h
```diff
@@ -5,14 +5,38 @@
...
 public:
     ToneEffect();
 
+    enum class Waveform
+    {
+        sine,
+        square,
+        sawtooth,
+        squareNoAlias,
+        triangle,
```
igorkorsukov, line 22: We usually have the name of enumerators in the style of CamelCase

au4/src/effects/effects_base/internal/effectexecutionscenario.cpp
```diff
@@ -59,7 +57,10 @@ muse::Ret EffectExecutionScenario::doPerformEffect(AudacityProject& project, con
     EffectManager& em = EffectManager::Get();
     Effect* effect = nullptr;
 
-    NotifyingSelectedRegion& selectedRegion = ViewInfo::Get(project).selectedRegion;
+    auto selectionCtrl = selectionController();
```
igorkorsukov, line 60: just in case, the `selectionController()` call is quite fast, i.e. it does not need to be optimized (not saved in a variable), i.e. using a variable can only make sense if you need to reduce the length of the code or make it more readable.
`selectionController()` implementation: 
```
    const std::shared_ptr<I>& get() const
    {
        if (!m_i) {
            const std::lock_guard<std::mutex> lock(StaticMutex::mutex);
            if (!m_i) {
                static std::string_view module = "";
                m_i = _ioc(iocContext())->template resolve<I>(module);
            }
        }
        return m_i;
    }

    const std::shared_ptr<I>& operator()() const
    {
        return get();
    }
```

i.e. the second and subsequent times, it is simply a return of a reference to a class member

au4/src/effects/builtin/tonegen/ToneView.qml
```diff
@@ -1,7 +1,129 @@
...
+import Muse.UiComponents
+import Audacity.Effects
+import Audacity.Playback
+import "../common"
 
-Rectangle {
+EffectBase {
 
-    color: "#009090"
+    property string title: "Tone" // review
```
igorkorsukov, line 10: I think we can return it from the effect, through its view model... But I don't see that there is a translatable name for the user. So for now we can do it like this:
```
property string title: qsTrc("effects, "Tone")
```
and then we'll see how to improve it

au4/src/effects/builtin/tonegen/toneeffect.h
```diff
@@ -4,15 +4,37 @@
...
     static const ComponentInterfaceSymbol Symbol;
 
     ComponentInterfaceSymbol GetSymbol() const override;
+
+    Waveform waveform() const;
+    void setWaveform(Waveform waveform);
+    double amplitude() const;
+    void setAmplitude(double amplitude);
+    double frequency() const;
+    void setFrequency(double frequency);
```
igorkorsukov, line 35: we need to decide how important it is to return the boundary values ​​(min, max) here, i.e. who is responsible for them, the effect itself (it says I work within these limits) or this knowledge does not belong to anyone and is indicated where needed, for example in UI controls. 
(about amplitude and frequency)

au4/src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -157,7 +162,12 @@ muse::Ret EffectsProvider::performEffect(AudacityProject& project, Effect* effec
             auto vr = valueRestorer(effect->mProgress, progress.get());
 
             assert(pInstanceEx); // null check above
-            returnVal = pInstanceEx->Process(settings);
+            try {
+                returnVal = pInstanceEx->Process(settings);
+            } catch (const ::AudacityException& e) {
+                // TODO: display message box using info from `e`.
+                returnVal = false;
```
igorkorsukov, line 169: We need to add error codes (and text for them) to the effects module 
see https://github.com/musescore/MuseScore/wiki/Error-handling
or https://github.com/audacity/audacity/blob/master/au4/src/trackedit/trackediterrors.h 

and return `Ret` with the error code and text for the user, in the scenario show the error message.

au4/src/effects/builtin/tonegen/ToneView.qml
saintmatthieu: Didn't really know what that was for 😅 Thx for the context ;)

au4/src/effects/effects_base/internal/effectexecutionscenario.cpp
saintmatthieu: I figured.
In this case, it was indeed for readability.
You may see other places where I do this not for readability, but if I have to debug a lot, I then don't have to step in and out of that getter you've pasted.

au4/src/effects/builtin/tonegen/ToneView.qml
saintmatthieu: Thanks, I missed that!

au4/src/effects/builtin/tonegen/toneeffect.h
saintmatthieu: I don't understand this comment, to begin with, why you mention min and max.

au4/src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: Will look into this - after lunch :D

au4/src/effects/builtin/tonegen/toneeffect.h
igorkorsukov: This effect can accept amplitude within certain limits, minimum - 0.0, maximum - 1.0 (not 2.0, not 20000.0 ...)
As I understand it, the frequency also has limits?
So we need to decide who is responsible for these limits.

For example, in Amplify these limits are set in the effect itself.
`AmplifyBase.h`
```
   static constexpr EffectParameter Ratio {
      &AmplifyBase::mRatio, L"Ratio", 0.9f, 0.003162f, 316.227766f, 1.0f
   };
```

```
Param<db_t> AmplifyEffect::amp() const
{
    return make_param<db_t>(muse::linear_to_db(mRatio), muse::linear_to_db(Ratio.min), muse::linear_to_db(Ratio.max));
}
```

au4/src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: Read the doc, I know what you meant now.
But here it's important that we show an error dialog. (Could be, for example, that we try to generate 10s but the next clip is 5s away only ; since we don't support overlapping clips (yet), this branch would be entered. It's not unlikely at all.
I see
```cpp
void Au3BasicUI::DoShowErrorDialog(const BasicUI::WindowPlacement& placement, const TranslatableString& dlogTitle,
                                   const TranslatableString& message, const ManualPageID& helpPage,
                                   const BasicUI::ErrorDialogOptions& options)
{
    Q_UNUSED(placement);
    Q_UNUSED(dlogTitle);
    Q_UNUSED(message);
    Q_UNUSED(helpPage);
    Q_UNUSED(options);
}
```
Is there a plan already to get this implemented?

DavidBailes: Relevant background to the reporting of errors is Error identification in WCAG standards:
https://www.w3.org/WAI/WCAG22/Understanding/error-identification

This applies to Software as well as web pages.

igorkorsukov: We do not use user interaction from logic (displaying messages, progress, etc.). The application is constantly improving, new ideas may come and, accordingly, logic functions, like bricks, can be used in completely different cases that do not imply those dialogs that should be shown now.
For example:
1. Some logic function may become part of a larger scenario, and the user needs to be informed not that this function failed, but that this larger scenario failed.
2. We can try one function, if it fails, try another one... without telling the user.
3. For some reason, we can execute some logic function many times in a row, but we don’t want to show the user a dozen dialogs...
4. Or we want to avoid what has already happened in effects - when executing an effect, two dialog progresses can now be shown, the second one on top of the first one.
and etc.
For this reasons, the usual approach is to return an error code (possibly with text), for this we use the Ret type. And somewhere above, at the level of the view model, controller or scenario, formed and display a message to the user. (edited) 

But if we can't avoid this in the AU3 code now, we'll have to implement this function and live with it for some time.

> Could be, for example, that we try to generate 10s but the next clip is 5s away only 
This is the task of forming information about the error, and the one who calls it can show it... if necessary.

igorkorsukov (approved):

au4/src/au3wrap/internal/au3project.cpp
Eism, line 60: Is it not the same as what we do in `Au3TrackeditProject::onTrackListEvent`?

saintmatthieu:
@igorkorsukov just force-pushed after rebase, together with a few commits that being fixing the issues reported by @dozzzzer.
Some of these are the manifestation of selection bugs (the generator querying selected time and tracks to make its decision where to generate). It's not ideal that I try to solve them here, or at least not until https://github.com/audacity/audacity/pull/7515 is merged.
So let's pause this work until then. I'll let you know when review can be resumed.

Issue #7498, opened 2024-10-09
update what's new dialog


LWinterberg:
and preload it with strings we may want to shuffle through in patch releases

chinakov:
Tested, Linux modal window is different to the one on Windows and macOS, but it should be fixed when the '-alpha' string will be removed upon recieving an RC. Something to keep in mind for regression.

PR #7499, opened 2024-10-09
update whatsnewdialog

LWinterberg:
Resolves: https://github.com/audacity/audacity/issues/7498

Reviewer: Verify that the new strings (even the #if 0'd ones) are in the translation catalogue

QA: Verify the what's new dialog - particularly the sizing (should be slightly larger than before), and that all links still work


src/WhatsNewDialog.cpp
```diff
@@ -127,11 +165,8 @@ END_EVENT_TABLE()
 WhatsNewDialog::WhatsNewDialog(wxWindow* parent, wxWindowID id)
    : wxDialogWrapper(parent, id, XO("Welcome to Audacity!"))
 {
-#if defined(__WXOSX__)
-   SetSize(WindowWidth, 400);
-#else
-   SetSize(FromDIP(wxSize(WindowWidth, 430)));
-#endif
+
+   SetSize(FromDIP(wxSize(WindowWidth, WindowHeight)));
```
LWinterberg, line 169: I just noticed that I have perhaps a logic error here, given that I set the pixel size separately above. Please verify the sizing on macoS

kryksyh: Yep, as it is right now, the window is too short, and scrollbars are shown:

image

using 460px looks good

image

src/WhatsNewDialog.cpp
```diff
@@ -39,13 +39,21 @@ namespace
...
 #else
-   constexpr auto WindowWidth = 400;
+   constexpr auto WindowWidth = 440;
+#endif
+
+#if defined(__WXOSX__)
+   constexpr auto WindowHeight = 400;
+#else
+   constexpr auto WindowHeight = 460;
 #endif
```
kryksyh, line 57:
```suggestion
constexpr auto WindowHeight = 460;
```

kryksyh: it is also a bit too short on windows

image

kryksyh: Nevermind, it fits okay, when the version string does not have `alpha` suffix. I believe that adjusting the dialog height automatically depending on the textbox size is not a trivial task with wxWidgets.

kryksyh (approved):

Issue #6885, opened 2024-07-16
[AU4] Selection - Track selection


DilsonsPickles:
What causes a track to be selected.

[Figma link](https://www.figma.com/design/oQEvoOqeKHQlqabmFmM0Mp/Selection?node-id=7-29582)

dozzzzer:
Tested. Issues #7631, #7630, #7629, #7628 are created.

PR #7515, opened 2024-10-11
[AU4] Consecutive and non-consecutive track selecting

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6885


au4/src/projectscene/view/clipsview/trackslistclipsmodel.cpp
```diff
@@ -31,14 +31,14 @@ void TracksListClipsModel::load()
...
-    m_selectedTrack = selectionController()->selectedTrack();
+    m_selectedTracks = selectionController()->selectedTracks();
 
     selectionController()->dataSelectedOnTracksChanged().onReceive(this, [this](const std::vector<trackedit::TrackId>& tracks) {
         setDataSelectedTracks(tracks);
     });
 
-    selectionController()->trackSelected().onReceive(this, [this](const trackedit::TrackId& trackId) {
-        setSelectedTrack(trackId);
+    selectionController()->tracksSelected().onReceive(this, [this](const std::vector<trackedit::TrackId>& tracksIds) {
```
igorkorsukov, line 40: let's add  `using TrackIdList = std::vector<TrackId>;`

grliszas14: fixed

src/trackedit/internal/au3/au3selectioncontroller.h
```diff
@@ -69,13 +69,13 @@ class Au3SelectionController : public ISelectionController
...
     // track selection
-    Val<trackedit::TrackId> m_selectedTrack;
+    Val<TrackIdList> m_selectedTracks;
 
     // clip selection
     Val<trackedit::ClipKey> m_selectedClip;
 
     // data selection
-    Val<std::vector<trackedit::TrackId>> m_selectedTrackIds;
+    Val<TrackIdList> m_dataSelectedTrackIds;
```
saintmatthieu, line 78: As discussed 1-1: there is no plan for discontinuous time selection (@LWinterberg please confirm), hence the time selection can entirely be determined by m_selectedStartTime and m_selectedEndTime.

saintmatthieu: In Audacity 3, some logic like the generators determine where to generate based on this time selection and the (possibly discontinuous) set of selected tracks.
In Audacity 4, with discontinuous clip selection, it seems to me like there is going to be an additional, prevailing logic: generators will first check if there are selected clips (`m_selectedClip` will by then be `m_selectedClips`) and use this as placeholder, else it will fall back on the Au3 behavior.

src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -526,9 +555,9 @@ TrackItem* TracksListModel::modelIndexToItem(const QModelIndex& index) const
     return m_trackList.at(index.row());
 }
 
-void TracksListModel::onSelectedTrack(trackedit::TrackId trackId)
+void TracksListModel::onSelectedTracks(const TrackIdList &tracksIds)
 {
     for (TrackItem* item : m_trackList) {
-        item->setIsSelected(item->trackId() == trackId);
+        item->setIsSelected(std::find(tracksIds.begin(), tracksIds.end(), item->trackId()) != tracksIds.end());
```
igorkorsukov, line 561: see https://github.com/audacity/audacity/pull/7515/files#r1801055990

igorkorsukov (approved):

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -868,7 +879,7 @@ void Au3Interaction::undo()
     int selectedIndex = std::distance(trackIdList.begin(),
                                       std::find(trackIdList.begin(),
                                                 trackIdList.end(),
-                                                selectionController()->selectedTrack()));
+                                                selectionController()->selectedTracks().at(0)));
```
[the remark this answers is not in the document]
grliszas14: I will create a new task for it, this case needs some extra care - it's not working as intended anyway (and never was based on your example)

src/projectscene/view/clipsview/selectionviewcontroller.h
```diff
@@ -53,9 +53,9 @@ class SelectionViewController : public QObject
 private:
 
     IProjectViewStatePtr viewState() const;
-    std::vector<trackedit::TrackId> trackIdList() const;
+    trackedit::TrackIdList trackIdList() const;
 
-    std::vector<trackedit::TrackId> determinateTracks(double y1, double y2) const;
+    trackedit::TrackIdList determinateTracks(double y1, double y2) const;
 
     TimelineContext* m_context = nullptr;
```
saintmatthieu, line 61: It seems to me that all `selectionActive`-related members (signal, getter, setter, state variable) are not used externally at all, as a consequence of which it should be removed to simplify and clarify the logic even more. Or is there a reason to keep it I'm missing?

src/projectscene/view/clipsview/selectionviewcontroller.cpp
```diff
@@ -24,7 +24,7 @@ void SelectionViewController::onPressed(double x, double y)
 
     if (!selectionController()->isDataSelected()) {
         TrackIdList tracks = determinateTracks(m_startPoint.y(), y);
```
saintmatthieu, line 26: This confused me for a while, I thought we might really pass a range, but in fact, looking at line 22 shows that `m_startPoint.y()` just equals `y` at this stage.
At least please replace `m_startPoint.y()` with just `y` to help the slow-minded of my kind understand your code better. You could even have a `determinateTrack(double y)` that'd internally just call `determinateTracks(y, y)`, but that'd be a stretch.

src/projectscene/view/clipsview/selectionviewcontroller.cpp
```diff
@@ -24,7 +24,7 @@ void SelectionViewController::onPressed(double x, double y)
...
-        selectionController()->setDataSelectedOnTracks(tracks, true);
+        selectionController()->setSelectedTracks(tracks, true);
     } else {
         setSelectionActive(false);
         selectionController()->resetDataSelection();
```
saintmatthieu, line 30: BTW, you have a `::resetDataSelection` method in this class that does this, which you could probably get rid of.

src/projectscene/view/clipsview/selectionviewcontroller.cpp
```diff
@@ -24,7 +24,7 @@ void SelectionViewController::onPressed(double x, double y)
 
     if (!selectionController()->isDataSelected()) {
         TrackIdList tracks = determinateTracks(m_startPoint.y(), y);
-        selectionController()->setDataSelectedOnTracks(tracks, true);
+        selectionController()->setSelectedTracks(tracks, true);
```
saintmatthieu, line 27: I think this method might be simplified to
```
void SelectionViewController::onPressed(double x, double y)
{
    m_selectionStarted = true;
    m_startPoint = QPointF(x, y);
    emit selectionStarted();

    selectionController()->resetDataSelection();
    TrackIdList tracks = determinateTracks(m_startPoint.y(), y);
    selectionController()->setSelectedTracks(tracks, true);
}
```
which would also make it behave more like AU3. Say I have two tracks selected, and I click on one of them to drag my selection. In Audacity 3, this will deselect the non-clicked track, while here it won't.
Or was this a new spec I'm not aware of?

src/trackedit/internal/au3/au3selectioncontroller.cpp
```diff
@@ -101,42 +95,7 @@ bool Au3SelectionController::isDataSelected() const
 
 bool Au3SelectionController::isDataSelectedOnTrack(TrackId trackId) const
```
saintmatthieu, line 96: About `isDataSelected()` : I'm not sure what it is supposed to tell its clients? For sure it's only about time (it doesn't consider whether a track is selected or not). It would return true if start time were 10 and end time were 1.
Start and end time setters are part of the public interface, and I don't see sanity checks to make sure that end time is never set to something less than start time, so it could happen. I'm not saying that these checks should be added now, but if this isn't intended to be possible and `isDataSelected` intends in fact to say `timeSelectionIsNotEmpty`, then I would like it to be renamed as such. It is used in many places in the code, and this could reveal bugs.

Likewise for `isDataSelectedOnTrack`: what it tells now is only whether the track of the given ID is selected or not, so I would like this see renamed to `isTrackSelected`.

saintmatthieu: I forgot: the implementation of `isDataSelected` should also add the `&&` that end time is strictly larger than start time (if negative selection spans are indeed not intended).

src/projectscene/view/clipsview/selectionviewcontroller.h
grliszas14: hmmm, it's actually used in TracksClipsView.qml

`property bool isNeedSelectionCursor: !selectionController.selectionActive && mainMouseArea.mouseOnTracks`

src/projectscene/view/clipsview/selectionviewcontroller.cpp
grliszas14: it's used on qml side so I can't remove it but it will be good to use it here instead of these two lines, good catch :D

src/trackedit/internal/au3/au3selectioncontroller.cpp
```diff
@@ -81,62 +84,18 @@ void Au3SelectionController::resetDataSelection()
...
-    m_dataSelectedTrackIds.set(TrackIdList(), true);
     m_selectedStartTime.set(-1.0, true);
     m_selectedEndTime.set(-1.0, true);
 }
 
-bool Au3SelectionController::isDataSelected() const
+bool Au3SelectionController::timeSelectionIsNotEmpty() const
 {
-    return muse::RealIsEqualOrMore(m_selectedStartTime.val, 0.0) && m_selectedEndTime.val > 0.0;
+    return muse::RealIsEqualOrMore(m_selectedStartTime.val, 0.0) && m_selectedEndTime.val > 0.0 && muse::RealIsEqualOrMore(m_selectedEndTime.val, m_selectedStartTime.val);
```
saintmatthieu, line 93: I think that last check must `m_selectedEndTime.wav > m_selectedStartTime.val`, shouldn't it? i.e. if they are equal, this method should return false - I think.

src/projectscene/view/clipsview/trackslistclipsmodel.cpp
```diff
@@ -106,7 +110,7 @@ QVariant TracksListClipsModel::data(const QModelIndex& index, int role) const
     case TrackIdRole:
         return QVariant::fromValue(track.id);
     case IsDataSelectedRole: {
-        return muse::contains(m_dataSelectedTracks, track.id);
+        return muse::contains(m_selectedTracks, track.id) && selectionController()->timeSelectionIsNotEmpty();
```
saintmatthieu, line 113: That makes a lot of sense now.
I'm wondering more and more if `m_selectedTracks` couldn't disappear in favor of calls to `selectionController()->selectedTracks()`, to avoid this app state being maintained in two places. Or does something speak against it?

src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -32,8 +32,16 @@ TracksListModel::TracksListModel(QObject* parent)
         onSelectedTracks(tracksIds);
     });
 
-    selectionController()->dataSelectedOnTracksSelected().onReceive(this, [this](trackedit::TrackIdList tracksIds) {
-        if (!tracksIds.empty()) {
+    selectionController()->dataSelectedStartTimeChanged().onReceive(this, [this](trackedit::secs_t begin) {
+        Q_UNUSED(begin);
+        if (selectionController()->timeSelectionIsNotEmpty()) {
+            m_audioDataSelected = true;
```
saintmatthieu, line 38: I looked and it seems to me that this `m_audioDataSelected` variable could disappear and query `selectionController()->timeSelectionIsNotEmpty()` where its value is read.

src/trackedit/internal/au3/au3selectioncontroller.cpp
```diff
@@ -81,62 +84,18 @@ void Au3SelectionController::resetDataSelection()
...
-Au3SelectionController::dataSelectedOnTracksChanged() const
-{
-    return m_dataSelectedTrackIds.changed;
-}
-
-muse::async::Channel<TrackIdList>
-Au3SelectionController::dataSelectedOnTracksSelected() const
-{
-    return m_dataSelectedTrackIds.selected;
+    return muse::contains(m_selectedTracks.val, trackId) && timeSelectionIsNotEmpty();
```
saintmatthieu, line 98: Nice.

src/trackedit/internal/au3/au3selectioncontroller.cpp
saintmatthieu: I'm more confident now: `doGlobalCut()`, `doGlobalDelete()` and other such should check this, and be no-opts if the selection is just a point selection.

src/projectscene/view/trackspanel/trackslistmodel.cpp
grliszas14: I'd like to get rid of it too but for now it's "not possible", meaning: it will break sync of track headers and tracks in the clips view. I will address it anyway as I believe Sergey will find out what's wrong with it but I need to extend muse framework first. As this may take some additional time until this thing gets merged into musescore I suggest we merge it for now as is. Added appropriate comment in the code though

src/trackedit/internal/au3/au3selectioncontroller.cpp
```diff
@@ -81,62 +84,18 @@ void Au3SelectionController::resetDataSelection()
...
-    m_dataSelectedTrackIds.set(TrackIdList(), true);
     m_selectedStartTime.set(-1.0, true);
     m_selectedEndTime.set(-1.0, true);
 }
 
-bool Au3SelectionController::isDataSelected() const
+bool Au3SelectionController::timeSelectionIsNotEmpty() const
 {
-    return muse::RealIsEqualOrMore(m_selectedStartTime.val, 0.0) && m_selectedEndTime.val > 0.0;
+    return muse::RealIsEqualOrMore(m_selectedStartTime.val, 0.0) && m_selectedEndTime.val > 0.0 && !muse::RealIsEqualOrLess(m_selectedEndTime.val, m_selectedStartTime.val);
```
saintmatthieu, line 93: Smart 😎

src/projectscene/view/clipsview/trackslistclipsmodel.cpp
```diff
@@ -30,15 +30,20 @@ void TracksListClipsModel::load()
 
     m_trackList = prj->trackList();
 
-    m_dataSelectedTracks = selectionController()->dataSelectedOnTracks();
-    m_selectedTracks = selectionController()->selectedTracks();
+    selectionController()->tracksSelected().onReceive(this, [this](const trackedit::TrackIdList& tracksIds) {
+        Q_UNUSED(tracksIds);
+        emit dataChanged(index(0), index(m_trackList.size() - 1), { IsTrackSelectedRole });
```
saintmatthieu, line 35: If `m_strackList.size() == 0`, `index(m_trackList.size() - 1)` will wrap around to an enormous number, won't it? It that's intended, please add a comment, otherwise typecast to index before subtracting.

saintmatthieu (approved):
One last easy thing to please double-check before presenting to QA.

src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -390,6 +390,7 @@ bool ClipsListModel::trimRightClip(const ClipKey& key, bool completed)
 void ClipsListModel::selectClip(const ClipKey& key)
 {
     selectionController()->setSelectedClip(key.key);
+    selectionController()->setSelectedTracks(TrackIdList({key.key.trackId}));
```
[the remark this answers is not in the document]
igorkorsukov: I think this logic should not be in a view model, but in the controller itself (in the setSelectedClip method). i.e. if this method is called from another place, the logic should be exactly the same

src/trackedit/internal/au3/au3selectioncontroller.cpp
igorkorsukov: looks like it's time to write unit tests for this controller :)

igorkorsukov (approved):

PR #7525, opened 2024-10-14
[AU4] Added effect preview function as is. Step 1

igorkorsukov:


au4/src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
```diff
@@ -58,5 +49,27 @@ StyledDialogView {
         Component.onCompleted: {
             bbox.buttonById(ButtonBoxModel.Apply).enabled = false
         }
+
```
saintmatthieu, line 52: There was a request from Product to have the `Apply` button being the default one and with blue highlight, like in the save-on-close dialog. Would you please add this?

au4/src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -182,3 +187,241 @@ muse::Ret EffectsProvider::performEffect(AudacityProject& project, Effect* effec
 
     return success ? muse::make_ok() : muse::make_ret(Ret::Code::UnknownError);
 }
+
```
saintmatthieu, line 190: I can assume that this is an unmodified copy-paste (except for formatting), right? Hence I won't really review its content.

saintmatthieu (approved):

au4/src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: Right!

saintmatthieu:
I just notice the `Au3Player::makeTransportTracks` implementation. Please consider re-using.

igorkorsukov:
> I just notice the `Au3Player::makeTransportTracks` implementation. Please consider re-using.

Yes, thank you, I will do this in the second step, during the refactoring stage

PR #7537, opened 2024-10-15
[AU4] added effect preview function as is

igorkorsukov:
restore https://github.com/audacity/audacity/pull/7525

### DIFF
--- a/src/effects/effects_base/ieffectexecutionscenario.h
+++ b/src/effects/effects_base/ieffectexecutionscenario.h
@@ -24,5 +24,7 @@ public:
     virtual muse::async::Notification lastProcessorIsNowAvailable() const = 0;
     virtual muse::async::Channel<EffectId> lastProcessorIdChanged() const = 0;
     virtual muse::Ret repeatLastProcessor() = 0;
+
+    virtual muse::Ret previewEffect(const EffectInstanceId& effectInstanceId) = 0;
 };
 }
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -8,11 +8,16 @@
 #include "libraries/lib-effects/Effect.h"
 #include "libraries/lib-components/EffectInterface.h"
 #include "libraries/lib-effects/EffectManager.h"
+#include "libraries/lib-effects/MixAndRender.h"
 #include "libraries/lib-wave-track/WaveTrack.h"
 #include "libraries/lib-transactions/TransactionScope.h"
 
 #include "libraries/lib-module-manager/PluginManager.h" // for NYQUIST_PROMPT_ID
 #include "libraries/lib-basic-ui/BasicUI.h"
+
+#include "libraries/lib-audio-io/AudioIO.h"
+#include "libraries/lib-stretching-sequence/StretchingSequence.h"
+#include "libraries/lib-audio-io/ProjectAudioIO.h"
 
 #include "log.h"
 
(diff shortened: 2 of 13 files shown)


src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -182,3 +187,241 @@ muse::Ret EffectsProvider::performEffect(AudacityProject& project, Effect* effec
...
+            XO("Preparing preview"),
+            ProgressShowStop
+            ); // Have only "Stop" button.
+        auto vr = valueRestorer(mProgress, progress.get());
+
+        auto vr2 = valueRestorer(mIsPreview, true);
+
+        access.ModifySettings([&](EffectSettings& settings){
+            // Preview of non-realtime effect
+            auto pInstance
```
GrigorasGabriel36, line 368: If dynamic_pointer_cast fails, pInstance will be nullptr, leading to undefined behavior if pInstance->Init() is called on a null pointer. Always check whether the cast was successful before using the pointer.

src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -182,3 +187,241 @@ muse::Ret EffectsProvider::performEffect(AudacityProject& project, Effect* effec
...
+}
+
+static void EffectPreview(EffectBase& effect,
+                          EffectSettingsAccess& access, std::function<void()> updateUI, bool dryOnly)
+{
+    auto cleanup0 = effect.BeginPreview(access.Get());
+
+    // These are temporary state in the Effect object that are meant be moved to
+    // a new class EffectContext
+    const auto numTracks = effect.mNumTracks;
```
GrigorasGabriel36, line 227: Many of these variables, such as mProgress, are defined but not used in the scope of the EffectPreview function or used once. Consider removing or using these variables to avoid unused-variable warnings.

src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -182,3 +187,241 @@ muse::Ret EffectsProvider::performEffect(AudacityProject& project, Effect* effec
...
+    const auto FocusDialog = BasicUI::FindFocus();
+    assert(FocusDialog); // postcondition
+
+    double previewDuration;
+    bool isNyquist = effect.GetFamily() == NYQUISTEFFECTS_FAMILY;
+    bool isGenerator = effect.GetType() == EffectTypeGenerate;
+
+    // Mix a few seconds of audio from all of the tracks
+    double previewLen;
+    gPrefs->Read(wxT("/AudioIO/EffectsPreviewLen"), &previewLen, 6.0);
```
GrigorasGabriel36, line 258: The value 6.0 appears to be a magic number

GrigorasGabriel36 (dismissed):

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: This is "Added effect preview function as is" - i.e. it is just a copy of the code from AU3 as is. 
The next step will be refactoring.
Done in several steps to see what will change

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: https://github.com/audacity/audacity/pull/7537#discussion_r1808779273

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: https://github.com/audacity/audacity/pull/7537#discussion_r1808779273

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: > If dynamic_pointer_cast fails, pInstance will be nullptr, leading to undefined behavior if pInstance->Init() is called on a null pointer. Always check whether the cast was successful before using the pointer.

There is a check in this case

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
```diff
@@ -58,5 +49,28 @@ StyledDialogView {
...
+            buttonRole: ButtonBoxModel.AcceptRole
+            buttonId: ButtonBoxModel.Cancel
+            onClicked: root.reject()
+        }
+
+        FlatButton {
+            text: qsTrc("global", "Apply")
+            buttonRole: ButtonBoxModel.DestructiveRole
+            buttonId: ButtonBoxModel.Apply
+            accentButton: true
```
igorkorsukov, line 72: >There was a request from Product to have the Apply button being the default one and with blue highlight, like in the save-on-close dialog. Would you please add this? 

added

src/effects/effects_base/ieffectexecutionscenario.h
```diff
@@ -24,5 +24,7 @@ class IEffectExecutionScenario : MODULE_EXPORT_INTERFACE
     virtual muse::async::Notification lastProcessorIsNowAvailable() const = 0;
     virtual muse::async::Channel<EffectId> lastProcessorIdChanged() const = 0;
     virtual muse::Ret repeatLastProcessor() = 0;
+
+    virtual muse::Ret previewEffect(const EffectInstanceId& effectInstanceId, EffectSettings& settings) = 0;
```
saintmatthieu, line 28: `settings` arg can't be const?

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: I prefer like Igor that these methods are guaranteed "left untouched" (with formatting tolerated), which means that their review can be skipped and time saved.

saintmatthieu (approved):

src/effects/effects_base/ieffectexecutionscenario.h
igorkorsukov: I would like to, but I don't possible now 

```
muse::Ret EffectsProvider::previewEffect(AudacityProject&, Effect* effect, EffectSettings& settings)
{
    std::shared_ptr<SimpleEffectSettingsAccess> pAccess = std::make_shared<SimpleEffectSettingsAccess>(settings);
    EffectPreview(*effect, *pAccess.get(), nullptr, false);
```

SimpleEffectSettingsAccess - there is no constructor accepting `const EffectSettings&`

Issue #6762, opened 2024-06-26
[AU4] Track - Delete track


DilsonsPickles:
[Figma Link](https://www.figma.com/design/k5rS9tnZl60nYmhJhvI2vf/Track-headers?node-id=280-42436)

dozzzzer:
Tested

PR #7561, opened 2024-10-16
Implement track deletion and a relevant minor refactoring

kryksyh:
Resolves: [#6762](https://github.com/audacity/audacity/issues/6762)

Allow the user to delete track

Recommended:

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TracksPanel.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TracksPanel.qml
@@ -98,16 +98,12 @@ Item {
                     view.interactive = true
                 }
 
-                onDuplicateRequested: {
-                    tracksModel.duplicateTrack(model.index)
-                }
-
-                onDeleteRequested: {
-                    tracksModel.deleteTrack(model.index)
+                onSelectionRequested: {
+                    tracksModel.selectRow(model.index)
                 }
 
                 onOpenEffectsRequested: {
-                        tracksModel.openEffectsForTrack(model.index)
+                    tracksModel.openEffectsForTrack(model.index)
                 }
             }
         }
--- a/src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
+++ b/src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
@@ -1,0 +1,54 @@
+/*
+* Audacity: A Digital Audio Editor
+*/
+#include "trackcontextmenumodel.h"
+
+using namespace au::projectscene;
+using namespace muse::uicomponents;
+using namespace muse::actions;
+
+void TrackContextMenuModel::load()
+{
+    AbstractMenuModel::load();
+
+    auto makeItemWithArg = [this](const ActionCode& actionCode) {
+        MenuItem* item = makeMenuItem(actionCode);
+        item->setArgs(ActionData::make_arg1<trackedit::TrackId>(m_trackId));
+        return item;
+    };
+
+    MenuItemList items {
+        makeItemWithArg("track-rename"),
+        makeItemWithArg("track-duplicate"),
+        makeItemWithArg("track-delete"),
+        makeSeparator(),
+        makeItemWithArg("track-make-stereo"),
+        makeItemWithArg("track-swap-stereo"),
+        makeItemWithArg("track-split-stereo"),
+    };
+
+    setItems(items);
+}
+
+au::trackedit::TrackId TrackContextMenuModel::trackId() const
+{
+    return m_trackId;
+}
+
+void TrackContextMenuModel::setTrackId(const trackedit::TrackId& newTrackId)
+{
+    if (m_trackId == newTrackId) {
+        return;
+    }
+    m_trackId = newTrackId;
+    emit trackIdChanged();
+}
+
+void TrackContextMenuModel::handleMenuItem(const QString& itemId)
+{
+    if (itemId == "track-rename") {
+        emit trackRenameRequested();
+    } else {
+        AbstractMenuModel::handleMenuItem(itemId);
+    }
+}
--- a/src/projectscene/view/trackspanel/trackslistmodel.cpp
+++ b/src/projectscene/view/trackspanel/trackslistmodel.cpp
@@ -142,21 +142,15 @@ void TracksListModel::addTrack(TrackTypes::Type type)
     }
 }
 
-void TracksListModel::duplicateTrack(int index)
+void TracksListModel::duplicateTrack(int row)
 {
-    UNUSED(index);
+    UNUSED(row);
     NOT_IMPLEMENTED;
 }
 
-void TracksListModel::deleteTrack(int index)
+void TracksListModel::openEffectsForTrack(int row)
 {
-    UNUSED(index);
-    NOT_IMPLEMENTED;
-}
-
-void TracksListModel::openEffectsForTrack(int index)
-{
-    UNUSED(index);
+    UNUSED(row);
     NOT_IMPLEMENTED;
 }
 
--- a/src/projectscene/view/trackspanel/trackslistmodel.h
+++ b/src/projectscene/view/trackspanel/trackslistmodel.h
@@ -42,9 +42,8 @@ public:
     Q_INVOKABLE void load();
 
     Q_INVOKABLE void addTrack(TrackTypes::Type type);
-    Q_INVOKABLE void duplicateTrack(int index);
-    Q_INVOKABLE void deleteTrack(int index);
-    Q_INVOKABLE void openEffectsForTrack(int index);
+    Q_INVOKABLE void duplicateTrack(int row);
+    Q_INVOKABLE void openEffectsForTrack(int row);
 
     Q_INVOKABLE QItemSelectionModel* selectionModel() const;
     Q_INVOKABLE void selectRow(int rowIndex);
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -76,7 +76,7 @@ Au3Track::Holder Au3Interaction::createNewTrackAndPaste(std::shared_ptr<Au3Track
 }
 
 TrackIdList Au3Interaction::determineDestinationTracksIds(const std::vector<Track>& tracks,
-                                                                                  TrackId destinationTrackId, size_t tracksNum) const
+                                                          TrackId destinationTrackId, size_t tracksNum) const
 {
     TrackIdList tracksIds;
     bool addingEnabled = false;
@@ -535,7 +535,7 @@ bool Au3Interaction::removeClipData(const trackedit::ClipKey& clipKey, secs_t be
     return true;
 }
 
-bool Au3Interaction::splitTracksAt(const TrackIdList &tracksIds, secs_t pivot)
+bool Au3Interaction::splitTracksAt(const TrackIdList& tracksIds, secs_t pivot)
 {
     for (const auto& trackId : tracksIds) {
         Au3WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
@@ -622,7 +622,7 @@ bool Au3Interaction::splitDeleteSelectedOnTrack(const TrackId trackId, secs_t be
     return true;
 }
 
-bool Au3Interaction::mergeSelectedOnTracks(const TrackIdList &tracksIds, secs_t begin, secs_t end)
+bool Au3Interaction::mergeSelectedOnTracks(const TrackIdList& tracksIds, secs_t begin, secs_t end)
 {
     secs_t duration = end - begin;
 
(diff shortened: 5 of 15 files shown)


src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -532,3 +535,33 @@ void TracksListModel::onSelectedTrack(trackedit::TrackId trackId)
         item->setIsSelected(item->trackId() == trackId);
     }
 }
+
+void TracksListModel::onTrackRenameAction(const muse::actions::ActionData& args)
```
kryksyh, line 539: @Eism i don't like the code duplication here, how can I propagate the action from TrackeditActionsController to QML item?

src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -124,28 +126,29 @@ void TracksListModel::addTrack(TrackTypes::Type type)
...
+    UNUSED(row);
     NOT_IMPLEMENTED;
 }
 
-void TracksListModel::deleteTrack(int index)
+void TracksListModel::trackDelete(int row)
 {
-    UNUSED(index);
-    NOT_IMPLEMENTED;
+    TrackItem* trackItem = modelIndexToItem(index(row));
```
Eism, line 144: please add nullptr check here

src/projectscene/view/trackspanel/trackslistmodel.h
```diff
@@ -41,11 +42,11 @@ class TracksListModel : public QAbstractListModel, public muse::async::Asyncable
 
     Q_INVOKABLE void load();
 
-    Q_INVOKABLE void addTrack(TrackTypes::Type type);
-    Q_INVOKABLE void addTracks(TrackTypes::Type type, int quantity);
-    Q_INVOKABLE void duplicateTrack(int index);
-    Q_INVOKABLE void deleteTrack(int index);
-    Q_INVOKABLE void openEffectsForTrack(int index);
+    Q_INVOKABLE void trackAdd(TrackTypes::Type type);
```
Eism, line 45: Previous version was better: verb + noun

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -831,6 +831,16 @@ void Au3Interaction::newLabelTrack()
     NOT_IMPLEMENTED;
 }
 
+void Au3Interaction::trackDelete(const TrackId trackId)
+{
+    auto& project = projectRef();
+    auto& tracks = ::TrackList::Get(project);
+    ::Track* track = DomAccessor::findTrack(project, ::TrackId(trackId));
```
Eism, line 838: also please add nullptr check

src/projectscene/view/trackspanel/trackslistmodel.h
kryksyh: I can't agree with that.

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -831,6 +831,16 @@ void Au3Interaction::newLabelTrack()
...
+    trackedit::ITrackeditProjectPtr trackEdit = globalContext()->currentTrackeditProject();
+    trackEdit->onTrackRemoved(DomConverter::track(track));
```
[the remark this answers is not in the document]
kryksyh: What if a subscriber tries to access data associated with the track directly or indirectly, say the name of the deleted track? If the track is deleted its data will be unavailable.

src/projectscene/view/trackspanel/trackslistmodel.cpp
Eism: Registering such actions is used in cases where we know for sure that these actions will be used in several places and usually they do not depend on the UI, i.e. in the case of track-rename I would assume that it accepts trackId and newName - the data is prepared and only needs to be applied.

I am not sure that there is an action for renaming a track, so I would suggest not registering such an action and simply handling the rename call - call titleEdit in qml. 
So probably we need a previous solution with `var operations` for MenuButton

src/projectscene/view/trackspanel/trackslistmodel.h
kryksyh: But, i'm okay to stick to either style, but we should enforce it. Right now there is a total mess in [itrackeditinteraction](https://github.com/audacity/audacity/blob/71a065085879ac3f265d66574f796f0729972247/src/trackedit/itrackeditinteraction.h#L34)

src/trackedit/internal/au3/au3interaction.cpp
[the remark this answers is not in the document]
kryksyh: in Qt they solved the problem by emitting two events `xxxAboutToBeRemoved` and `xxxRemoved` which we do not have in AU3 yet.
So I would rather have this event before the data is gone.

src/trackedit/internal/trackedituiactions.cpp
```diff
@@ -115,7 +115,44 @@ const UiActionList TrackeditUiActions::m_actions = {
              TranslatableString("action", "Duplicate"),
              TranslatableString("action", "Duplicate")
              ),
-    // clip
+    // track actions
+    UiAction("track-rename",
```
Eism, line 119: no longer needed

kryksyh: @Eism UiAction is needed to populate the menu; please correct me if I'm wrong.

src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
```diff
@@ -0,0 +1,51 @@
...
+    setItems(items);
+}
+
+au::trackedit::TrackId TrackContextMenuModel::trackId() const
+{
+    return m_trackId;
+}
+
+void TrackContextMenuModel::setTrackId(const trackedit::TrackId& newTrackId)
+{
```
grliszas14, line 39: I would add

```c++
if (m_trackId == newTrackId) {
    return;
}

```

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TracksPanel.qml
```diff
@@ -97,16 +97,8 @@ Item {
...
-                    tracksModel.duplicateTrack(model.index)
-                }
-
-                onDeleteRequested: {
-                    tracksModel.deleteTrack(model.index)
-                }
-
                 onOpenEffectsRequested: {
-                        tracksModel.openEffectsForTrack(model.index)
+                    tracksModel.openEffectForTrack(model.index)
```
grliszas14, line 102: shouldn't it be Effects? (looking at Q_INVOKABLE function in trackslistmodel)
```qml
tracksModel.openEffectsForTrack(model.index)
```

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -141,7 +143,8 @@ void TrackeditActionsController::doGlobalSplitCut()
         secs_t selectedEndTime = selectionController()->dataSelectedEndTime();
 
         dispatcher()->dispatch(SPLIT_CUT_SELECTED,
-                           ActionData::make_arg3<std::vector<TrackId>, secs_t, secs_t>(selectedTracks, selectedStartTime, selectedEndTime));
+                               ActionData::make_arg3<std::vector<TrackId>, secs_t, secs_t>(selectedTracks, selectedStartTime,
```
grliszas14, line 146: ups, looks like my formatter stopped working 😬

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TracksPanel.qml
kryksyh: oops, typo, thanks

src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
kryksyh: sure

grliszas14 (approved):

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -540,6 +545,26 @@ void TrackeditActionsController::newLabelTrack()
...
+        trackId = selectionController()->selectedTrack();
+    }
+
+    if (trackId == -1) {
+        return;
+    }
+
+    trackeditInteraction()->deleteTrack(trackId);
+
+    projectHistory()->pushHistoryState("Delete track", "Delete track");
```
igorkorsukov, line 565: This should be called in trackeditInteraction()->deleteTrack.
The point is that if trackeditInteraction()->deleteTrack is called from somewhere else.. the logic associated with the history should also be executed, i.e. it is part of the logic of the trackeditInteraction()->deleteTrack method

grliszas14: Actually I started to think about that in terms of implementing clip overlapping: when you move a clip over another clip in the end you end up with doing multiple actions at once:
 1. Clip move
2. Clip trim or clip delete
Both of these will add state to the undo-stack while I think it should be considered as one action
What do you think?

igorkorsukov: There is also this option: 

```
void moveClip(.., bool isModifyHistory = true)
{
    //  do move clip
    ...
    
    if (isModifyHistory) {
        pushHistoryState(...);
    }
}

void overlapClip(.., bool isModifyHistory = true)
{
   moveClip(.., false /*isModifyHistory*/);
   trimClip(.., false /*isModifyHistory*/));
   
   if (isModifyHistory) {
       pushHistoryState(...);
   }
}

```
but the first option with private methods without side effects is better

src/projectscene/view/trackspanel/trackslistmodel.h
```diff
@@ -92,6 +91,7 @@ private slots:
 
     void onProjectChanged();
     void onSelectedTrack(trackedit::TrackId trackId);
+    void onTrackRenameAction(const muse::actions::ActionData& args);
```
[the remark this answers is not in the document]
kryksyh: Yes, it was a leftover, removed.

igorkorsukov (approved):

src/trackedit/internal/trackedituiactions.cpp
Eism: in the case of track-rename, we have to create MenuItem manually and add it to the list

src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
```diff
@@ -0,0 +1,59 @@
...
+    AbstractMenuModel::load();
+
+    auto makeItemWithArg = [this](const ActionCode& actionCode) {
+        MenuItem* item = makeMenuItem(actionCode);
+        item->setArgs(ActionData::make_arg1<trackedit::TrackId>(m_trackId));
+        return item;
+    };
+
+    MenuItem* rename = new MenuItem(this);
+    rename->setTitle({ "Rename track", "Rename" });
```
igorkorsukov, line 21: Here you need to explicitly specify TranslatableString
Otherwise the utility for generating the file for translation will not process these lines (there is just a parser by keywords, TranslatableString is a keyword)

src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
```diff
@@ -0,0 +1,59 @@
...
+    rename->setId("track-rename");
+    rename->setState({ .enabled = true });
```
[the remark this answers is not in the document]
kryksyh: Here is the discussion with @Eism 
https://github.com/audacity/audacity/pull/7561#discussion_r1804252221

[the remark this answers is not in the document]
kryksyh: Basically because the rename is only to be executed from the context menu

src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
kryksyh: I see, thanks

src/trackedit/internal/trackedituiactions.cpp
igorkorsukov: @Eism @kryksyh to understand why this item is created manually and not from an action?

src/projectscene/view/trackspanel/trackcontextmenumodel.cpp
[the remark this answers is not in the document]
igorkorsukov: I don't think that's the reason, most of the action is from one place now... but that's changing..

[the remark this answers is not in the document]
igorkorsukov: For example, I really like the control that is sometimes used on websites, when you hover the mouse over the action button - and the actions open up in a fan, usually there are 3-5 actions, the most popular ones... :)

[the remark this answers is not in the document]
kryksyh: Are you talking about [pie-menu](https://en.wikipedia.org/wiki/Pie_menu)? That actually would be awesome to have as optional feature.

grliszas14 (approved):

igorkorsukov (approved):

PR #7574, opened 2024-10-17
7573 crash on close due to undoable unsaved changes

### LINKED ISSUE #7573: A crash when trying to undo/redo adding and reordering of the effect on the master
### Bug description

_No response_

### Steps to reproduce

1. Generate a rhythm track
2. Add a built-in effect to the master (e.g. Bass and treble)
3. Turn OFF and then turn ON the entire master effects chain by clicking the blue power button
4. `Edit > Undo` x2
5. Observe: the rhythm track clip is gone
6. `Edit > Redo`
7. Observe: the rhythm track clip is back
8. Click Play
9. Click Stop
10. Ctrl (Cmd) + W -> "No"

### Expected behavior

No crash; the project is closed.

### Actual behavior

A crash:

https://github.com/user-attachments/assets/2509bed6-c704-451a-8915-6c46267da781

### Audacity Version

Audacity 3.7.x

### Operating system

macOS

### Additional context

_No response_

saintmatthieu:
Resolves: #7573 

`UndoManager::MarkUnsaved()` resets its `int state` member to -1. `MarkUnsaved()` is used when toggling realtime effects on or off. On the other hand, 99f1314b16bcb74d0c52a3135bf7bde541f3b72f introduced a call to
```
ProjectHistory::Get(project).SetStateTo(undoManager.GetSavedState(), ...);
``` 
but `GetSavedState()` was now returning -1 meanwhile `SetStateTo` was expecting an `unsigned int` ... Hence the crash.

`MarkUnsaved()` resetting `state` to -1 makes the undo manager loose track of the last saved state. Simply making `SetStateTo` robust against calls with `-1` is sufficient to fix the crash but leaves the possibility that the user changes something (e.g. project tempo), toggles a realtime effect, says "No, I don't want to save" when closing, and yet, when reopening her project, sees that the change was saved nevertheless.

 , thereby loosing track of particular unsaved actions. When closing and saying "No, don't save changes", the state of the project can't be reverted to the last saved change anymore because `GetSavedState()` returns -1 and is hence unusable.

The proposed solution gets `MarkUnsaved` not to reset `state` to -1. Reviewers should verify that this doesn't have unintended side effects.

Recommended:

### DIFF
--- a/src/BatchCommands.cpp
+++ b/src/BatchCommands.cpp
@@ -698,7 +698,6 @@ bool MacroCommands::ApplyMacro(
             undoManager.Undo(
                [&]( const UndoStackElem &elem ){
                   history.PopState( elem.state ); } );
-            undoManager.AbandonRedo();
          });
       }
    });
--- a/libraries/lib-project-history/ProjectHistory.cpp
+++ b/libraries/lib-project-history/ProjectHistory.cpp
@@ -117,8 +117,12 @@ void ProjectHistory::PopState(const UndoState &state, bool doAutosave)
          pExtension->RestoreUndoRedoState(project);
 }
 
-void ProjectHistory::SetStateTo(unsigned int n, bool doAutosave)
+void ProjectHistory::SetStateTo(int n, bool doAutosave)
 {
+   assert(n >= 0);
+   if (n < 0)
+      return;
+
    auto &project = mProject;
    auto &undoManager = UndoManager::Get( project );
 
--- a/libraries/lib-project-history/ProjectHistory.h
+++ b/libraries/lib-project-history/ProjectHistory.h
@@ -43,7 +43,8 @@ public:
    ~ProjectHistory() override;
 
    void InitialState();
-   void SetStateTo(unsigned int n, bool doAutosave = true);
+   //! @pre n >= 0
+   void SetStateTo(int n, bool doAutosave = true);
    bool UndoAvailable() const;
    bool RedoAvailable() const;
    void PushState(
--- a/libraries/lib-project-history/UndoManager.h
+++ b/libraries/lib-project-history/UndoManager.h
@@ -171,7 +171,6 @@ class PROJECT_HISTORY_API UndoManager final
    void RenameState( int state,
       const TranslatableString &longDescription,
       const TranslatableString &shortDescription);
-   void AbandonRedo();
    void ClearStates();
    void RemoveStates(
       size_t begin, //!< inclusive start of range
@@ -215,6 +214,7 @@ class PROJECT_HISTORY_API UndoManager final
 
    void EnqueueMessage(UndoRedoMessage message);
    void RemoveStateAt(int n);
+   void AbandonRedo();
 
    AudacityProject &mProject;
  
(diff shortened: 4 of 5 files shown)


saintmatthieu:
Provided that this fix works as intended, the changes in the `UndoManager` and `ProjectHistory` will have to be ported to AU4. @ianalarina FYI

src/BatchCommands.cpp
```diff
@@ -698,7 +698,6 @@ bool MacroCommands::ApplyMacro(
             undoManager.Undo(
                [&]( const UndoStackElem &elem ){
                   history.PopState( elem.state ); } );
-            undoManager.AbandonRedo();
```
saintmatthieu, line 701: This shouldn't cause a problem: when reaching this stage, the macro has already pushed a new state, which means `AbandonRedo()` has already been called. At this stage, calling it again is trivial: there is nothing to undo already.

libraries/lib-project-history/UndoManager.cpp
```diff
@@ -254,9 +247,8 @@ void UndoManager::PushState(const TranslatableString &longDescription,
 
 void UndoManager::AbandonRedo()
 {
-   if (saved > current) {
-      saved = -1;
-   }
+   if (mSaved.GetValue() > current)
+      mSaved.SetValue(-1);
```
kryksyh, line 251: Does it make sense to add 'Reset' method?

libraries/lib-project-history/UndoManager.cpp
```diff
@@ -228,11 +226,6 @@ void UndoManager::PushState(const TranslatableString &longDescription,
        lastAction.Translation() == longDescription.Translation() &&
        mayConsolidate ) {
       ModifyState();
-      // MB: If the "saved" state was modified by ModifyState, reset
-      //  it so that UnsavedChanges returns true.
-      if (current == saved) {
-         saved = -1;
-      }
       return;
```
kryksyh, line 229: I believe we need a call to MarkUnsaved() here

libraries/lib-project-history/UndoManager.cpp
```diff
@@ -141,8 +139,8 @@ void UndoManager::RemoveStates(size_t begin, size_t end)
 
       if (current > begin)
         --current;
-      if (saved > static_cast<int>(begin))
-        --saved;
+      if (mSaved.GetValue() > static_cast<int>(begin))
```
kryksyh, line 142: cast is no longer needed

libraries/lib-project-history/UndoManager.cpp
saintmatthieu: I think you're right. The comment mislead me (`ModifyState()` in fact never changed `saved`), yet is a new state is pushed, of course there is something to save.

libraries/lib-project-history/UndoManager.cpp
saintmatthieu: Why not?

saintmatthieu: Right. May address on rebase.

kryksyh (approved):

saintmatthieu:
For now I reverted the changes that did not just strictly fixed the reported crash.
I haven't found a fix that doesn't cause some sort of regression so far.

PR #7588, opened 2024-10-21
Add video and musehub buttons to the WhatsNew dialog

kryksyh:

Recommended:


src/WhatsNewDialog.cpp
```diff
@@ -100,7 +105,7 @@ wxString MakeWhatsNewText()
       << wxT(R"(<p><a href=")") << WhatsNewURL << wxT(R"(">)")
       // Bug: (Windows) specified width and height should match exactly to the size of the image
       << wxT(R"(<img src="memory:whats_new_btn.jpeg" width="352" height="198" /><br></a></p>)")
-      << wxT("<h3>") << XO("What's new in Audacity %s").Format(AUDACITY_VERSION_STRING) << wxT("</h3>")
+      << wxT("<h3>") << XO("What's new in Audacity %s").Format("3.7.0") << wxT("</h3>")
```
LWinterberg, line 108: This looks like a leftover from testing

LWinterberg (approved):

src/WhatsNewDialog.cpp
kryksyh: Yep, kind of. I wanted a CI build for all platforms with a short version string

PR #7590, opened 2024-10-21
6940 vscode setup

saintmatthieu:
Resolves: #6940

 Help developers who want to use Visual Studio or VSCode to easily get setup.

Recommended:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 3
+indent_size = 4
 charset = "utf-8"
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,22 @@
+{
+   "configurations": [
+      {
+         "name": "Win32",
+         "includePath": [
+            "${workspaceFolder}/**",
+            "${QTDIR}/include/**"
+         ],
+         "defines": [
+            "_DEBUG",
+            "UNICODE",
+            "_UNICODE"
+         ],
+         "compilerPath": "cl.exe",
+         "cStandard": "c17",
+         "cppStandard": "c++17",
+         "intelliSenseMode": "msvc-x64",
+         "configurationProvider": "ms-vscode.cmake-tools"
+      }
+   ],
+   "version": 4
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,7 @@
+{
+    "recommendations": [
+        "ms-vscode.cmake-tools",
+        "ms-vscode.cpptools",
+        "zachflower.uncrustify" // for code formatting
+    ],
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,13 @@
+{
+    "configurations": [
+        {
+            "name": "run audacity",
+            "type": "cppvsdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/build/install/bin/audacity.exe",
+            "preLaunchTask": "CMake: install",
+            "stopAtEntry": false,
+            "cwd": "${workspaceFolder}",
+        }
+    ]
+}
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -66,6 +66,9 @@ TODO: generalize to other OSes
 
 Note: the default generator is Ninja. Have it installed, or change the "cmake.generator" value in `.vscode/settings.json` to something else (eg `Visual Studio 16 2019`).
 
+#### Open the workspace
+One way of doing this is by executing Ctrl+Shift+P, and choosing "Open Workspace from File". Navigate to the root of the repository and select the `.vscode/audacity.code-workspace` file. Else, you can open the folder, then the workspace file, and click the "Open Workspace" button.
+
 #### Install recommended extensions
 When opening the repository in VSCode, you will be prompted to install recommended extensions. You can do it then. If you missed it, you can install them later by executing Ctrl+Shift+P, typing `Extensions: Show Recommended Extensions`, and installing the recommended extensions.
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -39,11 +39,15 @@ Git, CMake, Ninja, Package manager, Compiler and Qt should all be added to the P
 
 ## Compiling
 
-Using QtCreator to edit and compile the code is generally preferred as, perhaps unsurprisingly, knows most things about Qt.
+### With QtCreator
+
+Using QtCreator to edit and compile the will provide the best intellisense and debugging support when interacting with QML. Nevertheless, debugging on Windows is slow, and if you mostly interact with the C++ code, you might want to other options, such as those listed below.
 
 To compile, just open CMakeLists.txt with QtCreator, configure the project with the auto-detected Qt 6.2.4 kit, and hit Build.
 
-Otherwise, standard cmake building applies: 
+### From the command line
+
+Standard cmake building applies: 
 
 ```
 # inside the audacity source:
@@ -51,3 +55,27 @@ cmake -S . -B build/ [options]  # configure (first build only)
 cmake --build build/            # build (every build)
 cmake --install build/          # install (every successful build)
 ```
+
+### With Visual Studio (Windows only)
+
+Double-click the `generate_sln.bat` script in the root of the repository. This will generate a Visual Studio solution in the `build` directory and build the `install` target. Open the generated solution (./build/audacity.sln) in Visual Studio and press F5 to run Audacity.
+
+### With VSCode (Windows only)
+
+TODO: generalize to other OSes
+
+Note: the default generator is Ninja. Have it installed, or change the "cmake.generator" value in `.vscode/settings.json` to something else (eg `Visual Studio 16 2019`).
+
+#### Install recommended extensions
+When opening the repository in VSCode, you will be prompted to install recommended extensions. You can do it then. If you missed it, you can install them later by executing Ctrl+Shift+P, typing `Extensions: Show Recommended Extensions`, and installing the recommended extensions.
+
+You should only have to do this once.
+
+##### Configure and build install target
+
+Execute Ctrl+Shift+P and choose "CMake: Configure".
+
+#### Build and run audacity
+
+Just **press F5**. It will build and install everything the first time, but afterwards it should be very fast, especially if you're using Ninja `:)`.
+
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -273,6 +273,20 @@ if (OS_IS_WIN)
               USE_SOURCE_PERMISSIONS
           )
 
+    if (MSVC AND DEVELOPER)
+        set_target_properties(${EXECUTABLE_NAME} PROPERTIES
+            VS_DEBUGGER_COMMAND "${CMAKE_INSTALL_PREFIX}/bin/${EXECUTABLE_NAME}"
+            VS_DEBUGGER_WORKING_DIRECTORY "${CMAKE_INSTALL_PREFIX}/bin"
+        )
+    endif()
+
+    if (DEVELOPER)
+        add_custom_command(TARGET ${EXECUTABLE_NAME} POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_PREFIX}/bin
+            COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:${EXECUTABLE_NAME}> ${CMAKE_INSTALL_PREFIX}/bin
+        )
+    endif()
+
 ###########################################
 # Linux
 ###########################################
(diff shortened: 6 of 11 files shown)


.editorconfig
```diff
@@ -4,7 +4,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 3
+indent_size = 4
```
LWinterberg, line 7: good lord finally

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,23 @@
+{
+   "configurations": [
+      {
+         "name": "Win32",
```
igorkorsukov, line 4: It should probably just be `Win`

.vscode/settings.json
```diff
@@ -0,0 +1,30 @@
+{
+    "cmake.generator": "Ninja",
+    "cmake.configureArgs": [
+        "-DCMAKE_INSTALL_PREFIX=${workspaceFolder}/build/install",
```
kryksyh, line 4: while we are at it, would you mind adding CMakeUserPresets.json?

 see https://github.com/audacity/audacity-private-docs/blob/main/build/CMake%20Presets.md for AU3
 
 Or we could do that within another PR

developer.bat
```diff
@@ -0,0 +1,17 @@
+IF NOT EXIST build mkdir build
+
+cd build
+
+cmake -DCMAKE_INSTALL_PREFIX=install -G "Visual Studio 16 2019" -DDEVELOPER=TRUE ..
+
+cmake --build . --config Debug --target install
```
igorkorsukov, line 7: I think there is no need to perform the building and installation itself, it is enough to just configure the project to generate .sln

developer.bat
```diff
@@ -0,0 +1,17 @@
+IF NOT EXIST build mkdir build
+
+cd build
+
+cmake -DCMAKE_INSTALL_PREFIX=install -G "Visual Studio 16 2019" -DDEVELOPER=TRUE ..
```
kryksyh, line 5: Should we hardcode VS version? I'm using 2022

src/app/CMakeLists.txt
```diff
@@ -273,6 +273,20 @@ if (OS_IS_WIN)
               USE_SOURCE_PERMISSIONS
           )
 
+    if (MSVC AND DEVELOPER)
```
igorkorsukov, line 276: DEVELOPER - I think we need to think of another name, it’s not clear what it is and where it comes from 
I think we need to check the name of the cmake generator 
Or something like this: VS_CMAKE_GENERATOR

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,23 @@
...
+         "includePath": [
+            "${workspaceFolder}/**",
+            "${QTDIR}/include/**"
+         ],
+         "defines": [
+            "_DEBUG",
+            "UNICODE",
+            "_UNICODE"
+         ],
+         "windowsSdkVersion": "10.0.22621.0",
```
kryksyh, line 14: Do we need it hardcoded here? I believe that it can be detected automatically.

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,23 @@
...
+         "compilerPath": "cl.exe",
+         "cStandard": "c17",
+         "cppStandard": "c++17",
+         "intelliSenseMode": "msvc-x64",
+         "configurationProvider": "ms-vscode.cmake-tools"
+      }
+   ],
+   "version": 4
+}
```
kryksyh, line 23: Some files are missing newlines

src/app/CMakeLists.txt
igorkorsukov: I see that this is also used for VSCode, then we still need to think about a good name.

.vscode/c_cpp_properties.json
kryksyh: does vscode not obey "insert_final_newline" for .editorconfig?

.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+    "recommendations": [
+        "ms-vscode.cmake-tools",
```
kryksyh, line 3: Shouldn't we add "SynthWave '84" them with glow enabled to the list of recommendations? (jk)

.vscode/settings.json
```diff
@@ -0,0 +1,30 @@
...
+        "-DDEVELOPER=TRUE"
+        // "-DAU_MODULE_EFFECTS_VST=ON"
+    ],
+    "cmake.environment": {
+        "QTDIR": "C:/Qt/6.2.0/msvc2019_64"
+    },
+    "qt-official.qmlls.customExePath": "C:/Qt/Tools/QtDesignStudio/qt6_design_studio_reduced_version/bin/qmlls.exe",
```
igorkorsukov, line 11: better to use QTDIR

developer.bat
```diff
@@ -0,0 +1,17 @@
+IF NOT EXIST build mkdir build
```
igorkorsukov, line 1: developer.bat - I think it would be better to call it something like `make_sln.bat` or `generate_sln.bat`

.vscode/c_cpp_properties.json
igorkorsukov: Found that this is an internal name in VSСode

.vscode/c_cpp_properties.json
igorkorsukov: As far as I can see, this is (c_cpp_properties.json) an automatically created file, it is better to add it to gitignore.

.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
+{
+    "configurations": [
+        {
+            "name": "run audacity",
+            "type": "cppvsdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/build/install/bin/audacity.exe",
+            "preLaunchTask": "build audacity",
```
igorkorsukov, line 8: Probably there should be not a build here, but an install.
@kryksyh  suggested that we can just do `CMake: Install`, and not add a task file.

.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
...
+            "cwd": "${workspaceFolder}",
+            "visualizerFile": "${workspaceFolder}/qt6.natvis",
+            "sourceFileMap": {
+                "C:/Users/qt/work/qt": "${QTDIR}/../Src"
+            },
+            "windows": {
+                "environment": [
+                    {
+                        "name": "PATH",
+                        "value": "${env:PATH};${command:qt-cpp.kitDirectory}${/}bin"
```
igorkorsukov, line 19: It is better to simply write in the instructions (and when trying to configure it, it is indicated) that us need to add Qt to the PATH

.vscode/settings.json
saintmatthieu: You see, I missed that. Will consider it for this PR.

developer.bat
saintmatthieu: The proposed approach in this PR is to add an audacity-target post-build step to copy the executable to the install directory. If you just build audacity without building install first, it will copy it, it will be alone in its `install/bin`.

developer.bat
saintmatthieu: Can we ensure a VS generator without specifying the version? I'd gladly remove this flag, but if I do, I don't know what ensures that it will pick a visual studio generator.

saintmatthieu: The only solution I've found to that is a bit complicated, but maybe worth using:
```bat
REM Detect the highest available Visual Studio version
for /f "tokens=*" %%i in ('"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -products * -requires Microsoft.Component.MSBuild -property installationVersion') do set VS_VERSION=%%i

REM The `VS_VERSION` will be something like 17.11.35312.102 ; keep only the major version
set VS_VERSION=%VS_VERSION:~0,2%

if %VS_VERSION%==16 (
    set GENERATOR="Visual Studio 16 2019"
) else if %VS_VERSION%==17 (
    set GENERATOR="Visual Studio 17 2022"
) else (
    echo No supported Visual Studio version found.
    exit /b 1
)

echo Using %GENERATOR%
```

src/app/CMakeLists.txt
saintmatthieu: I just realize that the `MSVC` check is sufficient for this part of the configuration (configuring the debugging command correctly). This then leaves the `DEVELOPER` flag to adding the post-build step to copy the executable to the install directory, which we can still discuss if we want this at all.

.vscode/c_cpp_properties.json
saintmatthieu: > Do we need it hardcoded here? I believe that it can be detected automatically.

Probably.

saintmatthieu: > As far as I can see, this is (c_cpp_properties.json) an automatically created file, it is better to add it to gitignore.

The `includePath`, adding the QTDIR, isn't default and it's something I enjoy.
Once the file is created, I'm not sure it gets modified. I suggest we give it a try and re-assess if we see that it does and it becomes a nuisance.

.vscode/c_cpp_properties.json
saintmatthieu: We have this:
```
[*.json]
insert_final_newline = unset
```
Are the missing newlines an issue, though?

.vscode/extensions.json
saintmatthieu: We should recommend the uncrustify extension, though. Will add.

src/app/CMakeLists.txt
igorkorsukov: If we decide to keep the copy step, let's name the flag something like: `OPTIMIZED_DEPLOY`

.vscode/extensions.json
saintmatthieu: > Shouldn't we add "SynthWave '84" theme with glow enabled to the list of recommendations? (jk)

I don't know. Does it work well with Doki-Theme?

.vscode/settings.json
saintmatthieu: Of course.

.vscode/c_cpp_properties.json
igorkorsukov: I believe we can add Qt to the include path in our cmake files.
Although it's strange that I didn't need it in my example...
Maybe because I have Qt in the PATH

.vscode/settings.json
```diff
@@ -0,0 +1,30 @@
+{
+    "cmake.generator": "Ninja",
+    "cmake.configureArgs": [
+        "-DCMAKE_INSTALL_PREFIX=${workspaceFolder}/build/install",
+        "-DDEVELOPER=TRUE"
+        // "-DAU_MODULE_EFFECTS_VST=ON"
+    ],
+    "cmake.environment": {
+        "QTDIR": "C:/Qt/6.2.0/msvc2019_64"
```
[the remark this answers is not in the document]
saintmatthieu: In fact, I think we can get rid of this entry altogether. I expect cmake will query the QTDIR env variable by default, and we do expect contributors to have this set.

developer.bat
saintmatthieu: I agree. It's nostalgia: my previous company had it named that way. I also like the idea of a developer 🦇

saintmatthieu: generate_sln sounds good.

.vscode/launch.json
saintmatthieu: > Probably there should be not a build here, but an install
This works because of the post-build task in the cmakelists that gets the executable copied to the install dir.
Having `install` as pre-launch task has for consequence that each time you launch after having modified the source code, the install target gets executed and, even if it's not the first time, just the check that everything (except audacity.exe) is up-to-date takes at least a couple of seconds.
On the other hand, for fast onboarding, what you suggest would be easier: the user wouldn't have to build the install target manually once.
But we can propose two launch tasks: the one you suggest and make it the default one, and for power users, the present one.

saintmatthieu: > even if it's not the first time, just the check that everything (except audacity.exe) is up-to-date takes at least a couple of seconds

I'm confused. Now it's super fast for me even with install. Ok for your proposal, then.

src/app/CMakeLists.txt
saintmatthieu: Sorry, no, it still is needed for Visual Studio, which doesn't have pre-launch tasks AFAIK.

saintmatthieu: ... but then it's not so much a matter of optimizing deployment anymore. `COPY_TO_INSTALL` ? Doesn't say everything but it gives a good clue.

saintmatthieu: ... or maybe we can just add that post-build step just if `MSVC`. Won't hurt CI anyway, as it uses Ninja. Then no need for that extra flag.

BUILDING.md
```diff
@@ -39,15 +39,43 @@ Git, CMake, Ninja, Package manager, Compiler and Qt should all be added to the P
...
 ```
 # inside the audacity source:
 cmake -S . -B build/ [options]  # configure (first build only)
 cmake --build build/            # build (every build)
 cmake --install build/          # install (every successful build)
 ```
+
+### With Visual Studio (Windows only)
+
+Double-click the `developer.bat` script in the root of the repository. This will generate a Visual Studio solution in the `build` directory and build the `install` target. Open the generated solution (./build/audacity.sln) in Visual Studio and press F5 to run Audacity.
```
igorkorsukov, line 61: `developer.bat` -> `generate_sln.bat`

igorkorsukov (approved):

developer.bat
kryksyh: I see, we can give it a shot. One thing though, it should be `%programfiles(x86)%` instead of `C:\Program Files (x86)`

.vscode/c_cpp_properties.json
kryksyh: > Are the missing newlines an issue, though?

It's not that important, it's just the way it's done historically. And without a newline, some command line tools can misbehave. For example if you `cat file` the prompt will be printed on the same line as the last line from the file

image

src/app/CMakeLists.txt
igorkorsukov: `COPY_TO_INSTALL` - acceptable for me

.vscode/extensions.json
kryksyh: Unfortunately not :(

kryksyh (approved):

Issue #7589, opened 2024-10-21
Implement base view for Reverb effect


igorkorsukov:
### Description

Perhaps the design will change, this is just a port of the view from au3, it is necessary in order to implement the work with presets

dozzzzer:
@igorkorsukov 

- There is no input value validation, meaning you can enter any positive or negative value and apply the effect, which will cause Audacity to crash.
- The increment for effect parameters is 0.5%, as opposed to 1% in the AU3 implementation.

<img width="155" alt="image" src="https://github.com/user-attachments/assets/d15bd9d0-301c-4fdd-ad6f-071398203e7f">

- When no selection is made, an attempt to open up the Reverb should bring up the following error message:

<img width="450" alt="image" src="https://github.com/user-attachments/assets/276822d6-aff3-4442-b495-22ebb1985c60">

dozzzzer:
@igorkorsukov 
- Make each text box a bit wider so that the fractional part isn't overlapped by the up/down controls anymore
<img width="450" alt="image" src="https://github.com/user-attachments/assets/3a0b15b3-96ae-4d4d-953e-862776ae1971">
- Decreasing the values using the down arrow is possible, but increasing the one by clicking the up arrow isn't:

https://github.com/user-attachments/assets/ac94d0d9-7afa-4932-902b-469b80dac2fe

- Change the error message copy when no audio is selected to
`Select the audio you want the Reverb to use and try again`

- Sometimes it throws an error when I try to apply the Reverb to a selection:

https://github.com/user-attachments/assets/72c54a22-9d6c-4077-b34c-7e76a2d25c64

igorkorsukov:
* Make each text box a bit wider so that the fractional part isn't overlapped by the up/down controls anymore - fixed 
* Decreasing the values using the down arrow is possible, but increasing the one by clicking the up arrow isn't  -fixed 
* Change the error message copy when no audio is selected to - changed 
* Sometimes it throws an error when I try to apply the Reverb to a selection - couldn't reproduce, let's watch until we get details

dozzzzer:
@igorkorsukov 

I'm still getting an error when I try to apply the effect to a selection:

<img width="400" alt="image" src="https://github.com/user-attachments/assets/c9d88500-d49b-49fc-a988-7143ab1d468c">

Please try to make a selection like on the screenshot and set every Reverb's parameter to a maximum. You can use my project if you wish: 
[...]

PR #7599, opened 2024-10-22
Added view for Reverb effect 

igorkorsukov:
Resolves: https://github.com/audacity/audacity/issues/7589

Perhaps the design will change, this is just a port of the view from au3, it is necessary in order to implement the work with presets 

ReverbEffect coped from AU3 almost as is


src/effects/builtin/reverb/ReverbView.qml
```diff
@@ -1,20 +1,104 @@
...
 
+        height: 40
+        anchors.left: parent.left
+        anchors.right: parent.right
+        anchors.margins: 8
+
+        StyledTextLabel {
+            id: label
+            anchors.left: parent.left
+            anchors.top: parent.top
```
saintmatthieu, line 43: Would just `anchors.verticalCenter: parent.verticalCenter` work, like in `IncrementalPropertyControl` ?

src/effects/builtin/reverb/ReverbView.qml
```diff
@@ -1,20 +1,104 @@
...
+            value: param.value
+            onMoved: param.value = slider.value
+        }
+    }
+
+    Column {
+        anchors.top: parent.top
+        anchors.left: parent.left
+        anchors.right: parent.right
+        anchors.bottom: wetOnly.top
```
saintmatthieu, line 77: It looks strange not to let the `CheckBox` specify `anchors.top: theColumn.bottom`. Would that have worked, too? (You don't have to change if it does, it's more out of curiosity.)

src/effects/builtin/reverb/ReverbView.qml
```diff
@@ -1,20 +1,104 @@
...
+            horizontalAlignment: Text.AlignLeft
+            width: 120
+        }
+
+        IncrementalPropertyControl {
+            id: numField
+            anchors.left: label.right
+            anchors.verticalCenter: parent.verticalCenter
+            width: 60
+            currentValue: param.value
```
saintmatthieu, line 55: I'm missing something: `param` is the `ParamItem` instance, but I don't see it declare a member named `value`.

src/effects/builtin/reverb/reverbviewmodel.h
```diff
@@ -1,9 +1,40 @@
...
+
+signals:
+    void paramsListChanged();
+    void wetOnlyChanged();
+
+private:
+
+    ReverbEffect* effect() const;
+    const ReverbSettings& settings() const;
+    ReverbSettings& mutSettings();
```
saintmatthieu, line 33: What does `mut` stand for?

saintmatthieu (approved):
LGTM - awesome!

src/effects/builtin/reverb/ReverbView.qml
igorkorsukov: Here's the logic: 
* this checkbox should be stuck to the bottom, not to the parameters (i think)
* to stick to a column, it must have geometry, i.e. a height is set. For a column, this is simple, but for example in a similar case for ListView with scrolling, when the number of elements can be different and even more than the height of the dialog, then this will all be a problem

src/effects/builtin/reverb/ReverbView.qml
igorkorsukov: here https://github.com/audacity/audacity/pull/7599/files#diff-d51d9585bf2291f98ec88a8b021e86c9cff150b62f06ef29c8a169804ac6e953R31

src/effects/builtin/reverb/reverbviewmodel.h
igorkorsukov: mut - mutable 

technically we can name both methods the same, we will be allowed to override the name.
But I usually prefer to name the constant method and the method for changing differently, so that it is clear where the value is simply read, and where it is intended to be changed

src/effects/builtin/reverb/ReverbView.qml
saintmatthieu: Aha!

src/effects/builtin/reverb/ReverbView.qml
saintmatthieu: Oy. Must have forgotten to ctrl+f...

src/effects/builtin/reverb/reverbviewmodel.h
saintmatthieu: I agree with that. I'd have preferred "mutableSettings" (It got me to wonder if it were related to muting), but it's not so bad.

PR #7617, opened 2024-10-25
Added id to Clip

igorkorsukov:


Eism (approved):

au3/libraries/lib-track/Track.h
```diff
@@ -92,8 +92,11 @@ class TrackId
    bool operator <  (const TrackId &other) const
    { return mValue <  other.mValue; }
 
+   inline operator int64_t() const { return mValue; }
+   inline int64_t raw() const { return mValue; }
+
 private:
-   long mValue;
+   int64_t mValue = -1;
```
GrigorasGabriel36, line 99: Stop using -1 as a “magic” value to represent an uninitialized state. This can be confusing and easily misinterpreted as a valid ID. Instead, use std::optional<int64_t>, which explicitly shows whether the ID is set.

au3/libraries/lib-track/Track.h
```diff
@@ -78,8 +78,8 @@ template<typename T>
 class TrackId
 {
 public:
-   TrackId() : mValue(-1) {}
-   explicit TrackId (long value) : mValue(value) {}
+    TrackId() = default;
+   explicit TrackId (int64_t value) : mValue(value) {}
 
    bool operator == (const TrackId &other) const
```
GrigorasGabriel36, line 84: Add the missing comparison operators. Defining only == and < leaves TrackId with incomplete comparisons.

au3/libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -220,10 +220,14 @@ double WaveClipChannel::GetStretchRatio() const
    return GetClip().GetStretchRatio();
 }
 
+
+static int64_t s_lastClipId = 0;
+
 WaveClip::WaveClip(size_t width,
```
GrigorasGabriel36, line 226: WaveClip::WaveClip(size_t width, const SampleBlockFactoryPtr &factory, sampleFormat format, int rate) 
    : mId(++s_lastClipId), mRate(rate), mSequences(width) {
    assert(width > 0);
}
In the constructor, initialize member variables directly in the initializer list rather than in the body. For instance, mId, mRate, and mSequences.resize(width) should ideally be in the initializer list:

src/au3wrap/internal/domaccessor.cpp
```diff
@@ -27,10 +26,10 @@ Au3WaveTrack* DomAccessor::findWaveTrack(Au3Project& prj, const Au3TrackId& au3t
     return dynamic_cast<Au3WaveTrack*>(findTrack(prj, au3trackId));
 }
 
-std::shared_ptr<Au3WaveClip> DomAccessor::findWaveClip(Au3WaveTrack* track, uint64_t au3ClipId)
+std::shared_ptr<Au3WaveClip> DomAccessor::findWaveClip(Au3WaveTrack* track, int64_t au3ClipId)
 {
     for (const std::shared_ptr<Au3WaveClip>& interval : track->Intervals()) {
```
GrigorasGabriel36, line 31: for (const auto& interval : track->Intervals()) {

GrigorasGabriel36 (changes requested):

au3/libraries/lib-track/Track.h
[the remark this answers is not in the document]
igorkorsukov: This is an AU3 class, it is designed as designed.
We do not refactor or improve AU3 classes.
We only make the minimum necessary changes to apply them to AU4

au3/libraries/lib-track/Track.h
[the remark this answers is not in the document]
igorkorsukov: https://github.com/audacity/audacity/pull/7617#discussion_r1816354011

igorkorsukov: Tell us more here, usually they don't use optinal for ID because they need performance, very often they search by ID, for example... What is your experience using optinal in such cases?

au3/libraries/lib-wave-track/WaveClip.cpp
igorkorsukov: made in accordance with the current AU3 style

src/au3wrap/internal/domaccessor.cpp
igorkorsukov: why do you think it's better to use auto? it's not clear what type we're dealing with when we review or read the code

kryksyh (approved):

au3/libraries/lib-track/Track.h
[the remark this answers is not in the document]
GrigorasGabriel36: So who and how do we decide what is considered minimum?

grliszas14:
I like that and I know that we wanted to do this thing from a long time but we simply couldn't, but:
I'd like to merge that PR only after a good testing from QA - I just opened this build quickly and immediately noticed that I can't copy clips to any place except an empty space. There are a lot of things that can get broken here (and we won't notice it) and also it's highly probable that it conflicts with most of our PR's that await for QA (so when we resolve conflicts we won't know where to quickly look for the bugs it those are found)

au3/libraries/lib-track/Track.h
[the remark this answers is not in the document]
igorkorsukov: the author of the changes is obvious. He knows what he needs to change minimally in AU3 to achieve his goal. If he didn't do something, then it's probably not necessary to achieve the goal.

Issue #7620, opened 2024-10-25
`Preferences > Modules`: individual module states do not stick


dozzzzer:
### Steps to reproduce

1. In Appimage build, go to `Preferences > Modules`
2. Set "Disabled" for any module that is enabled by default, e.g. `mod-ffmpeg`
3. Click OK to confirm
4. Restart Audacity
5. Go to `Preferences > Modules` and check the displayed state of the module

### Expected behavior

The previously set state "Disabled" sticks upon restart of Audacity.

### Actual behavior

The previously set state has been reset to "Enabled".

### Audacity Version

Audacity 3.7.x

### Operating system

All

### Additional context

The issue isn't reproducible when running a locally built version of Audacity.

kryksyh:
@dozzzzer the issue is relevant to all systems if the installation is moved. For example, if you change the name of 'Audacity.app' on mac modules will be autoenabled.

kryksyh:
@dozzzzer @LWinterberg do you mind merging this to 3.7.0?

dozzzzer:
I do not

chinakov:
Tested, modules are now disabled for good, even if the portable installation is moved to a different path post disabling the modules.

PR #7622, opened 2024-10-27
Fix modules were automatically enabled if portable installation was m…

kryksyh:
Resolves: #7620

Fix the issue when the modules are automatically enabled when audacity is used in portable mode and the installation is moved. 

Do not check for path changes; a datetime check should be enough. 

Recommended:


saintmatthieu:
> a datetime check should be enough

I completely agree.

saintmatthieu (approved):

LWinterberg (approved):
and rebased onto 3.7.1

Issue #6759, opened 2024-06-26
[AU4] Track - Moving tracks


DilsonsPickles:
The ability to move tracks up and down in the list of tracks.

[Figma Link](https://www.figma.com/design/k5rS9tnZl60nYmhJhvI2vf/Track-headers?node-id=272-19038)

chinakov:
Tested, thanks!

Re-ording with mouse dragging will be implemented in a separate task: https://github.com/audacity/audacity/issues/7656

PR #7634, opened 2024-10-29
Implement track reordeing

kryksyh:
Resolves: #[6759](https://github.com/audacity/audacity/issues/6759)

Recommended:

### DIFF
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -673,6 +673,8 @@ bool Au3Interaction::duplicateClip(const ClipKey& clipKey)
     }
 
     pushProjectHistoryDuplicateState();
+
+    return true;
 }
 
 bool Au3Interaction::clipSplitCut(const ClipKey& clipKey)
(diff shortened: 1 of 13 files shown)


src/trackedit/internal/au3/au3interaction.h
```diff
@@ -63,11 +63,23 @@ class Au3Interaction : public ITrackeditInteraction
...
+    void moveTracksDown(const TrackIdList& trackIds) override;
+    void moveTracksToTop(const TrackIdList& trackIds) override;
+    void moveTracksToBottom(const TrackIdList& trackIds) override;
+    void moveTracksTo(const TrackIdList& trackIds, int to) override;
 
     void undo() override;
     void redo() override;
 
 private:
+    enum class TrackMoveDir {
```
Eism, line 76: It's better not to abbreviate but to use Direction (confused with the directory)

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -891,6 +893,145 @@ void Au3Interaction::duplicateTracks(const TrackIdList& trackIds)
...
+        }
+    } else {
+        while (pos != to && tracks.CanMoveUp(*au3Track)) {
+            pos--;
+            tracks.MoveUp(*au3Track);
+        }
+    }
+
+    if (pos != to) {
+        LOGW() << "Can't move track to " << to;
```
Eism, line 966: It would be useful to know the current position of the track (pos) too

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -891,6 +893,145 @@ void Au3Interaction::duplicateTracks(const TrackIdList& trackIds)
...
+    }
+
+    auto track = DomConverter::track(au3Track);
+
+    trackedit::ITrackeditProjectPtr trackEdit = globalContext()->currentTrackeditProject();
+    trackEdit->notifyAboutTrackMoved(track, pos);
+}
+
+void Au3Interaction::moveTrack(const TrackId trackId, TrackMoveDir dir)
```
Eism, line 975: direction :)

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -891,6 +893,145 @@ void Au3Interaction::duplicateTracks(const TrackIdList& trackIds)
...
+        while (tracks.CanMoveDown(*au3Track)) {
+            to++;
+            tracks.MoveDown(*au3Track);
+        }
+        break;
+    default:
+        return;
+    }
+
+    if (from == to) {
```
Eism, line 1024: I don't understand this condition, and also &dir
Could you please explain?

src/trackedit/itrackeditinteraction.h
```diff
@@ -56,6 +56,11 @@ class ITrackeditInteraction : MODULE_EXPORT_INTERFACE
     virtual void newLabelTrack() = 0;
     virtual void deleteTracks(const TrackIdList& trackIds) = 0;
     virtual void duplicateTracks(const TrackIdList& trackIds) = 0;
+    virtual void moveTracksUp(const TrackIdList& trackIds) = 0;
```
Eism, line 59: or add only two methods like
```
virtual void moveTracks(const TrackIdList& trackIds, TrackMoveDir dir) =0;
virtual void moveTracks(const TrackIdList& trackIds, int pos) =0;
```

src/trackedit/internal/au3/au3interaction.h
kryksyh: Agree

src/trackedit/internal/au3/au3interaction.cpp
kryksyh: Perhaps it is a bad naming on my side. `from` is our starting position, and `to` is resulting position. If we finished at the same position where we started that would mean we failed to move anywhere :)

Will try to come up with better name for those variables

src/trackedit/internal/au3/au3interaction.cpp
kryksyh: absolutely

src/trackedit/internal/au3/au3interaction.cpp
kryksyh: @Eism is it more understandable now? :)

Eism: yes, thanks

src/trackedit/itrackeditinteraction.h
Eism: I'm afraid that if someone else from the community decides to add a new type, they will do it without switching to a parameter and we cannot require them to refactor, this is roughly how open source works

Eism (approved):

Eism (approved):

PR #7635, opened 2024-10-29
Fix inconsistent track selection

kryksyh:

Recommended:

### DIFF
--- a/src/projectscene/view/trackspanel/trackslistmodel.cpp
+++ b/src/projectscene/view/trackspanel/trackslistmodel.cpp
@@ -22,6 +22,13 @@ TracksListModel::TracksListModel(QObject* parent)
         setItemsSelected(deselected.indexes(), false);
         setItemsSelected(selected.indexes(), true);
 
+        std::vector<TrackId> selectedTrackIds;
+        for (auto index : m_selectionModel->selectedIndexes()) {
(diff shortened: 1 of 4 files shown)


src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -172,17 +172,19 @@ QItemSelectionModel* TracksListModel::selectionModel() const
...
+
     if (m_audioDataSelected) {
-        selectionController()->resetSelectedTracks();
+        selectionController()->resetDataSelection();
         clearSelection();
         m_audioDataSelected = false;
     }
 
-    selectionController()->resetDataSelection();
-    QModelIndex modelIndex = index(rowIndex);
```
grliszas14, line 184: These two lines:
```c++
QModelIndex modelIndex = index(rowIndex);
m_selectionModel->select(modelIndex);
``` 
need to stay, otherwise you'll not be able to select multiple tracks using \"ctrl/shift + track header click\"
(because ItemMultiSelectionModel is not updated)

grliszas14 (changes requested):

src/projectscene/view/trackspanel/trackslistmodel.cpp
kryksyh: Fixed

grliszas14 (approved):

Issue #7287, opened 2024-09-09
[AU4] Playback start/stop by pressing the Space bar on the keyboard


dozzzzer:
### Description

Same as in AU3, pressing the Space bar on the keyboard should start and stop the playback of the project.

PR #7639, opened 2024-10-30
Playback cursor seek fix

Eism:
Also resolves: [#7287](https://github.com/audacity/audacity/issues/7287)

### DIFF
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -145,7 +145,7 @@ Channel<uint32_t> PlaybackController::midiTickPlayed() const
 
 float PlaybackController::playbackPositionInSeconds() const
 {
-    return secondsFromMilliseconds(m_currentPlaybackTimeMsecs);
+    return secondsFromMilliseconds(m_currentPlaybackTime);
 }
 
 muse::async::Channel<TrackId> PlaybackController::trackAdded() const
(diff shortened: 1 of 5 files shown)


src/playback/internal/playbackcontroller.cpp
```diff
@@ -232,7 +232,8 @@ void PlaybackController::onSeekAction(const muse::actions::ActionData& args)
         return;
     }
 
-    double secs = args.arg<double>(0);
+    muse::secs_t secs = args.arg<double>(0);
+    m_lastPlaybackTimeSeekMsecs = secs;
```
igorkorsukov, line 236: m_lastPlaybackTimeSeekMsecs -> m_lastPlaybackSeekTime

src/playback/internal/playbackcontroller.h
```diff
@@ -125,7 +125,8 @@ class PlaybackController : public IPlaybackController, public muse::actions::Act
     muse::async::Channel<muse::actions::ActionCode> m_actionCheckedChanged;
 
     muse::async::Notification m_currentSequenceIdChanged;
-    playback::msecs_t m_currentPlaybackTimeMsecs = 0;
+    muse::secs_t m_currentPlaybackTimeMsecs = 0;
+    muse::secs_t m_lastPlaybackTimeSeekMsecs = 0;
```
igorkorsukov, line 129: m_currentPlaybackTimeMsecs  -> m_currentPlaybackTime  
m_lastPlaybackTimeSeekMsecs -> m_lastPlaybackSeekTime

`M` - exactly what needs to be removed (in seconds, not milliseconds)
but I think `Sec` can also be removed, because now everything should be in `sec` by default, and the data type shows in what units of measurement.

igorkorsukov (approved):

Issue #7335, opened 2024-09-16
[AU4] Trim/stretch clip edge hover


DilsonsPickles:
#### Description

When the User hovers over the edge of a clip, their cursor will change to the trim or stretch cursor (depending on where they hover) and on click and drag, they'll be able trim/stretch the clip.

[Figma link](https://www.figma.com/design/01XLIICmy2AG6t1ieyTp66/Clips?node-id=1550-212062&t=ZyyRXdfGbAIRE63A-4)

DilsonsPickles:
Kicked back as we will be following AU3 logic. Only a single hover area and stretch applied on keyboard modifier

DilsonsPickles:
Updated to reflect AU3 behaviour.

grliszas14:
Created separate follow-up task to fix issue found by @chinakov https://github.com/audacity/audacity/issues/7657
In the video you can see overlap trimming (but overlap trimming is not merged yet, it's a separate PR https://github.com/audacity/audacity/issues/6999). There is a chance that after merging both "overlap trimming" PR and this one - it will be fixed automatically.

PR #7642, opened 2024-10-31
Trim on edge hover

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7335


src/app/CMakeLists.txt
```diff
@@ -104,7 +104,9 @@ endif()
 ###########################################
 # Resources
 ###########################################
-qt_add_resources(APP_RCC_SOURCES app.qrc)
+qt_add_resources(APP_RCC_SOURCES
+    app.qrc
+    ${PROJECT_SOURCE_DIR}/images/resources.qrc)
```
Eism, line 109: hmm, can we move it to `projectscene` module?

grliszas14: fixed 👍

src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -301,6 +301,30 @@ bool ClipsListModel::changeClipTitle(const ClipKey& key, const QString& newTitle
     return ok;
 }
 
+QVariant ClipsListModel::getSortedNeighbor(const ClipKey& key, int offset) const
+{
+    QList<ClipListItem*> sortedList = m_clipList;
+
+    std::sort(sortedList.begin(), sortedList.end(), [](ClipListItem* a, ClipListItem* b) {
```
igorkorsukov, line 308: I think we can do sort once in the `reload` method.

grliszas14: moved sorting to update() function - that's the place where m_clipList (clips that are visible on a screen) get updated

src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -301,6 +305,26 @@ bool ClipsListModel::changeClipTitle(const ClipKey& key, const QString& newTitle
     return ok;
 }
 
+QVariant ClipsListModel::getSortedNeighbor(const ClipKey& key, int offset) const
+{
+    QList<ClipListItem*> sortedList = m_clipList;
```
igorkorsukov, line 310: now you don't need to make a copy

src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -301,6 +305,26 @@ bool ClipsListModel::changeClipTitle(const ClipKey& key, const QString& newTitle
     return ok;
 }
 
+QVariant ClipsListModel::getSortedNeighbor(const ClipKey& key, int offset) const
```
igorkorsukov, line 308: I think now we can just call it `neighbor`, or I think it's better to add two methods - the `next` and the `prev`

grliszas14: added next() and prev() methods + left neighbor() as a private member

Eism (approved):

PR #7654, opened 2024-11-04
Play+Setcursor to Space, Play to Ctrl+Space

LWinterberg:
This swaps the shortcuts for play/stop and play/stop and set cursor.


petersampsonaudacity:
Tested on W10 with @LWinterberg 's test branch build:   audacity-win-3.7.1-alpha-20241104+bfb7202-x64

Works as intended.

kryksyh (approved):

LWinterberg:
After some more testing, I'm closing this for the purposes of 3.7.1: The "set cursor" part here affects region selections too, which is consistent with the previous logic. In Audacity4 we'll do something better, where space will play/"pause" unless you've got a region selected, in which case it'll stick to just the region. We won't be making this change in Audacity 3 anymore though.

Issue #7516, opened 2024-10-11
[AU4] Non-consecutive audio-data selection


grliszas14:
https://www.figma.com/design/oQEvoOqeKHQlqabmFmM0Mp/10---Selection?node-id=16-52334&node-type=frame&t=M26XuNHPSREHSOuF-0

dozzzzer:
When I make a non-consecutive audio data selection on tracks 1 and 3, the marquee selection appears, which is correct. But then if I try to extend the selection by holding the Ctrl (Cmd) key and clicking in the desired spot, on the first click the non-consecutive selection I made previously will disappear, however, if I click once again while still holding the Ctrl key, the selection will appear. Observe how I demonstrate this behavior by clicking in the blank space of a track while holding the Ctrl key:

https://github.com/user-attachments/assets/dad41527-656e-4247-a3c9-8a0ba7cfb2f8

@DilsonsPickles Does this look right to you?

DilsonsPickles:
The second click you make (when the selection appears again) is what I'd expect to see on the first click.

grliszas14:
Fixed the behaviour

PR #7655, opened 2024-11-05
Clipsview track selection

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7516

### DIFF
--- a/src/projectscene/view/clipsview/clipslistmodel.cpp
+++ b/src/projectscene/view/clipsview/clipslistmodel.cpp
@@ -44,6 +44,9 @@ void ClipsListModel::init()
         Q_UNUSED(time);
         updateItemsMetrics();
     });
+    selectionController()->tracksSelected().onReceive(this, [this](const TrackIdList&) {
+        updateItemsMetrics();
+    });
 
     reload();
 }
--- a/src/projectscene/view/clipsview/selectionviewcontroller.cpp
+++ b/src/projectscene/view/clipsview/selectionviewcontroller.cpp
@@ -18,13 +18,43 @@ SelectionViewController::SelectionViewController(QObject* parent)
 
 void SelectionViewController::onPressed(double x, double y)
 {
-    m_selectionStarted = true;
-    m_startPoint = QPointF(x, y);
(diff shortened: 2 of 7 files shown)


src/projectscene/view/clipsview/selectionviewcontroller.cpp
```diff
@@ -33,11 +80,23 @@ void SelectionViewController::onPositionChanged(double x, double y)
         return;
     }
 
+    Qt::KeyboardModifiers modifiers = QApplication::keyboardModifiers();
+
+    //! NOTE: always treat simultaneously pressed Ctrl and Shift as Ctrl
```
Eism, line 85: it looks like we can move it to a separate function and reuse

kryksyh (commented):
Not sure what really causes this, but if the user shift+clicks on a track twice without moving on X, it resets playhead to 0

kryksyh (commented):
Another minor _glitch_ , when the user ctrl+presses it resets the data selection, until the mouse is moved or released. Probably need to

```
    selectionController()->setDataSelectedStartTime(m_context->positionToTime(x1, true /*withSnap*/), false);
    selectionController()->setDataSelectedEndTime(m_context->positionToTime(x2, true /*withSnap*/), false);
```
    
 in onPrssed

src/projectscene/view/clipsview/selectionviewcontroller.cpp
grliszas14: fixed 👍

Eism (approved):

src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -44,6 +44,9 @@ void ClipsListModel::init()
         Q_UNUSED(time);
         updateItemsMetrics();
     });
+    selectionController()->tracksSelected().onReceive(this, [this](const TrackIdList&) {
+        updateItemsMetrics();
```
Eism, line 48: I think here it is worth updating only those parameters that depend on the selection
am I right?

kryksyh (approved):

Issue #7455, opened 2024-10-03
[AU4] Fix pasting audio selection into empty space


DilsonsPickles:
### Description

Select audio over multiple tracks, paste into last track - app crashes.

dozzzzer:
@grliszas14 

When pasting into the empty space underneath the tracks, then undoing and pasting again, Audacity crashes:

https://github.com/user-attachments/assets/8404d924-3322-4b70-ac5c-f46a53f0e212

Also a very similar and perhaps related scenario: when pasting into the second track, over another clip, it does the job at first, but if you undo and try to paste again, it also crashes Audacity:

https://github.com/user-attachments/assets/3a36566d-091d-40d5-b404-050012852e10

dozzzzer:
@grliszas14 Just discovered one more way to crash Audacity in this PR's build:

- Have an audio clip, copy some of its data
- Click in the empty space underneath the track and paste
- Undo
- Redo

This crashes Audacity every time I try it.

dozzzzer:
Tested

PR #7671, opened 2024-11-06
Fixed redoing paste into empty space

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7455

### DIFF
--- a/src/projectscene/view/trackspanel/trackslistmodel.cpp
+++ b/src/projectscene/view/trackspanel/trackslistmodel.cpp
@@ -76,6 +76,8 @@ void TracksListModel::load()
         m_trackList.push_back(buildTrackItem(track));
     }
 
+    onSelectedTracks(selectionController()->selectedTracks());
+
     prj->tracksChanged().onReceive(this, [this](std::vector<au::trackedit::Track> tracks) {
         onTracksChanged(tracks);
     });
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -1394,30 +1394,12 @@ void Au3Interaction::undo()
 
     auto trackeditProject = globalContext()->currentProject()->trackeditProject();
 
-    // Find the index of the currently selected track in the old list
-    auto trackIdList = trackeditProject->trackIdList();
-    auto selectedTracks = selectionController()->selectedTracks();
-    int selectedIndex = -1;
-    if (!selectedTracks.empty()) {
-        selectedIndex = std::distance(trackIdList.begin(),
-                                      std::find(trackIdList.begin(),
-                                                trackIdList.end(),
-                                                selectedTracks.at(0)));
-    }
     projectHistory()->undo();
 
     // Undo removes all tracks from current state and
     // inserts tracks from the previous state so we need
     // to reload whole model
     trackeditProject->reload();
-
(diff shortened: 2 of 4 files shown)


src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1394,30 +1394,21 @@ void Au3Interaction::undo()
...
-    } else {
-        selectionController()->resetSelectedTracks();
+    // Select tracks that were previously selected
+    auto& project = projectRef();
+    auto& tracks = Au3TrackList::Get(project);
+    TrackIdList selectedTracks;
+    for (const auto & selectedTrack : tracks.Selected()) {
+        selectedTracks.push_back(DomConverter::trackId(selectedTrack->GetId()));
     }
+    selectionController()->setSelectedTracks(selectedTracks);
```
kryksyh, line 1411: @grliszas14 here you sync selectionController state with its internal data model (tracks.Selected), `Au3SelectionController::setSelectedTracks` is doing exactly that it sets each track to its current `selected` state, which is probably not what we want to do.

 I'm not 100% sure, but i probably would make more sense if we update selectionController state when tracklist sends SELECTION_CHANGE

src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -92,6 +92,8 @@ void TracksListModel::load()
         m_trackList.push_back(buildTrackItem(track));
     }
 
+    onSelectedTracks(selectionController()->selectedTracks());
```
kryksyh, line 95: Why receiving this event is not enough?
```
selectionController()->tracksSelected().onReceive(this, [this](trackedit::TrackIdList tracksIds) {
        onSelectedTracks(tracksIds);
    });
 ```

File: src/trackedit/internal/au3/au3interaction.cpp (hunk 1 above)
Comment (grliszas14) [reply src/trackedit/internal/au3/au3interaction.cpp:1411]: yes, it sets each track to its current selected state but it's additionally updating its own state (m_selectedTracks), yes, it would make more sense to respond to SELECTION_CHANGE if we got SELECTION_CHANGE.
undo/redo replaces whole model (which means getting multiple TrackListEvent::DELETION/ADDITION) - the problem I see here is that every DELETION event has trackId=-1 -> so I can't really respond to that because I don't know which track to delete for example

File: src/projectscene/view/trackspanel/trackslistmodel.cpp (hunk 2 above)
Comment (grliszas14) [reply src/projectscene/view/trackspanel/trackslistmodel.cpp:95]: because undo/redo reloads whole model and does it asynchronously
```c++
muse::async::Async::call(this, [this]() {
        load();
    });
``` 

so the order looks like this:
1. call reload()
2. setTracksSelected() -> emit signal that tracks changed
3. signal is handled where it's connected
4. reload is done here (it's too late to catch and handle selectedChanged signal)

so I set selectedTracks manually here

src/trackedit/internal/au3/au3interaction.cpp
kryksyh: Right, that's unfortunate. So, how about hook selectionController to TrackListEvent::DELETION/ADDITION to update its state?

As I understand there are other situations where tracklist gets changed besides undo/redo, and it will also require us to update selectionController state, right?

kryksyh (approved):

Issue #6774, opened 2024-06-27
[AU4] Playback. BPM applying to ClipsView


Eism:
### Description

When changing BPM, the ruler changes, but the changes are not applied to ClipsView

grliszas14:
I did this along with enable-stretchnig, re-assigning

dozzzzer:
Tested

Issue #6630, opened 2024-06-06
[AU4] Clip action: enable-stretching


igorkorsukov:
### Description

called from clip context menu

grliszas14:
QA bulletpoints:
* Stretch-to-match-tempo remembered for every clip separately
* Stretch-to-match-tempo saved into project file so it is remembered on next project open
* Enabling Stretch-to-match-tempo resets the tempo of the clip immediately to match project’s tempo
* Tempo change should affect clips view
* Trimming should be applied if any clips are going to overlap

dozzzzer:
Tested

PR #7715, opened 2024-11-13
Enable stretching

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/6630
Resolves: https://github.com/audacity/audacity/issues/6774

### DIFF
--- a/au3/libraries/lib-wave-track/TimeStretching.cpp
+++ b/au3/libraries/lib-wave-track/TimeStretching.cpp
@@ -15,10 +15,40 @@
 #include "TempoChange.h"
 #include "UserException.h"
 #include "WaveClip.h"
+#include "realfn.h"
 #include <algorithm>
 
 const TranslatableString TimeStretching::defaultStretchRenderingTitle =
    XO("Pre-processing");
+
+namespace {
+void trimRightOverlapping(WaveTrack &track) {
+    std::list<std::shared_ptr<WaveClip> > clips = { track.Intervals().begin(), track.Intervals().end() };
+    if (clips.size() <= 1) {
+        return;
+    }
+
+    clips.sort([](const std::shared_ptr<WaveClip>& c1, const std::shared_ptr<WaveClip>& c2) {
+        return c1->GetPlayStartTime() < c2->GetPlayStartTime();
+    });
+
+    auto prev_it = clips.begin();
+    for (auto it = std::next(clips.begin()); it != clips.end(); ++it) {
+        auto& previousClip = *prev_it;
+        auto& currentClip = *it;
+
+        if (!muse::RealIsEqualOrLess(currentClip->GetPlayStartTime(), previousClip->GetPlayStartTime())
+            && muse::RealIsEqualOrLess(currentClip->GetPlayStartTime(), previousClip->GetPlayEndTime())
+            && muse::RealIsEqualOrMore(currentClip->GetPlayEndTime(), previousClip->GetPlayEndTime())) {
+
+            double overlap = (previousClip->GetPlayEndTime() - currentClip->GetPlayStartTime());
+            previousClip->TrimRight(overlap);
+        }
+
+        prev_it = it;
+    }
+}
+}
 
 bool TimeStretching::HasPitchOrSpeed(
    const WaveTrack &track, double t0, double t1)
@@ -66,5 +96,7 @@ DEFINE_ATTACHED_VIRTUAL_OVERRIDE(OnWaveTrackProjectTempoChange) {
    {
       for (const auto pClip : track.Intervals())
          pClip->OnProjectTempoChange(oldTempo, newTempo);
+
+      trimRightOverlapping(track);
    };
 }
--- a/au3/libraries/lib-wave-track/WaveClip.cpp
+++ b/au3/libraries/lib-wave-track/WaveClip.cpp
@@ -1637,6 +1637,11 @@ void WaveClip::SetRawAudioTempo(double tempo)
    mRawAudioTempo = tempo;
 }
 
+void WaveClip::SetClipTempo(double tempo)
+{
+    mClipTempo = tempo;
+}
+
 bool WaveClip::SetCentShift(int cents)
 {
    if (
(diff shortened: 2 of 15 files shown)


au3/libraries/lib-wave-track/TimeStretching.cpp
```diff
@@ -15,11 +15,41 @@
...
 #include "UserException.h"
 #include "WaveClip.h"
+#include "realfn.h"
 #include <algorithm>
 
 const TranslatableString TimeStretching::defaultStretchRenderingTitle =
    XO("Pre-processing");
 
+namespace {
+void trimRightOverlapping(WaveTrack &track) {
```
kryksyh, line 25: there is a function `GetClipsIntersecting` that returns a list of clips in a given region that might help here

au3/libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -547,23 +550,33 @@ size_t WaveClip::GreatestAppendBufferLen() const
 void WaveClip::OnProjectTempoChange(
    const std::optional<double>& oldTempo, double newTempo)
 {
+   const auto ratioChange = *oldTempo / newTempo;
```
kryksyh, line 553: there some places where the function is called with std::nullopt

au3/libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -547,23 +550,33 @@ size_t WaveClip::GreatestAppendBufferLen() const
...
+
    if (!mRawAudioTempo.has_value())
       // When we have tempo detection ready (either by header-file
       // read-up or signal analysis) we can use something smarter than that. In
       // the meantime, use the tempo of the project when the clip is created as
       // source tempo.
       mRawAudioTempo = oldTempo.value_or(newTempo);
 
+   if (!mStretchEnabled && !muse::is_equal(ratioChange, 0.0)) {
```
kryksyh, line 562: can you explain this one please `muse::is_equal(ratioChange, 0.0)` ?

au3/libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -547,23 +550,33 @@ size_t WaveClip::GreatestAppendBufferLen() const
...
       mSequenceOffset *= ratioChange;
       mTrimLeft *= ratioChange;
       mTrimRight *= ratioChange;
       StretchCutLines(ratioChange);
       mEnvelope->RescaleTimesBy(ratioChange);
    }
-   mProjectTempo = newTempo;
+
+   if (mStretchEnabled) {
+      mProjectTempo = newTempo;
```
kryksyh, line 577: Should we rename this member since it would not contain project temp anymore?

au3/libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1002,6 +1029,12 @@ bool WaveClip::HandleXMLTag(const std::string_view& tag, const AttributesList &a
                return false;
             mClipStretchRatio = dblValue;
          }
+         else if (attr == ClipStretchEnabled_attr)
+         {
+            if (!value.TryGet(dblValue))
```
kryksyh, line 1034: should be deserialized as a bool

au3/libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -671,6 +684,19 @@ bool WaveClip::HasPitchOrSpeed() const
...
+{
+   return mStretchEnabled;
+}
+
+void WaveClip::SetStretchEnabled(bool enabled)
+{
+   if (mStretchEnabled == enabled) {
+      return;
+   }
+   mStretchEnabled = enabled;
```
kryksyh, line 697: I'm a bit concerned about that. Consider we have stretch enabled, then set non-1 stretch ratio, and then disable stretch. 
From the api standpoint the clip will have some stretching applied, but have stretch disabled. This will be a bit ambiguous.
Also there is another function called `HasPitchOrSpeed`

kryksyh: Going through the PR one more time, it seems to make sense to call the new member `mStretchSyncEnabled`, `mProjectStretchSyncEnabled` or something like that. What do you think?

grliszas14: Sounds better to me too 👍

au3/libraries/lib-wave-track/WaveClip.cpp
grliszas14: oh... I've handled that in a wrong way... when std::nullopt is passed as oldTempo then ratioChange is 0.0... I'll fix that

au3/libraries/lib-wave-track/WaveClip.cpp
kryksyh: Yeah, probably. On another topic, we reset the current clip tempo to the current project tempo once this guy is set to true

au3/libraries/lib-wave-track/TimeStretching.cpp
grliszas14: Yes, but `GetClipsIntersecting` checks if there are clips within given t0<->t1 boundaries and in this case we need to check if any pair of clips overlap each other and calculate by how much. I just simply follow the same rules as the other overlapping functions

au3/libraries/lib-wave-track/WaveClip.cpp
grliszas14: yes, exactly, my bad, fixed

au3/libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -562,14 +558,21 @@ void WaveClip::OnProjectTempoChange(
 
    if (oldTempo.has_value())
    {
-      const auto ratioChange = *oldTempo / newTempo;
+      const auto ratioChange = oldTempo.value() / newTempo;
```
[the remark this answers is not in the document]
kryksyh: nevermind, sorry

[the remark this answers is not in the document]
kryksyh: your original code was perfectly fine, i missed that there is a check two lines above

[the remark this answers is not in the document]
grliszas14: yeah but there's check `if (oldTempo.has_value())` right before access so we should be covered here, right?

au3/libraries/lib-wave-track/WaveClip.cpp
grliszas14: Fixed resetting to current project tempo + added saving clipTempo as well

dozzzzer:
@grliszas14 Please change the option's name to "Stretch with tempo changes"

Issue #7648, opened 2024-11-01
Implementation task. When a user makes a selection, the play / stop behaviour matches AU3


Eism:
### Description

At the moment the player ignores the user's selection range. 
We should match behaviour found in Audacity 3.

https://github.com/user-attachments/assets/3e4cebc9-1556-470e-b387-474b49d9569e

Desired behaviour for Play (**Spacebar**)
- When there is a selection made, and the player has stopped, pressing play (or pressing spacebar), should play the audio from the beginning to the end of the selected audio and then stop.
- If the play function is active and the user makes a selection, once the player has stopped, the cursor should move back to the beginning of the audio selection.

Desired behaviour for Play/Return cursor (**Shift+Spacebar**)
- **shift+spacebar** should ignore audio selections. When pressing **shift+spacebar**, the cursor should play from it's current position and not stop at the end of the audio selection. If the user presses **shift+spacebar** again, then the cursor should behave as it does right now - returning the cursor back to wherever the user initially played from.

DilsonsPickles:
@Eism do you require a spec to implement this behaviour? I don't want to spend time mocking up this behaviour in Figma if we can use AU3 as a reference. 

@Tantacrul I'm fine with this behaviour and would be hesitant to change it given that it's playback + selection (two of the core interaction areas)

Tantacrul:
I can provide specific instructions when needed @Eism @DilsonsPickles. I agree, we don't need a spec for this.

chinakov:
This is implemented, thanks.

PR #7721, opened 2024-11-14
Playback. When a user makes a selection, the play / stop behaviour matches AU3

Eism:
Resolves: #7648

### DIFF
--- a/src/playback/internal/au3/au3player.cpp
+++ b/src/playback/internal/au3/au3player.cpp
@@ -77,8 +77,6 @@ void Au3Player::play()
 
     double t0 = selectedRegion.t0();
     double t1 = selectedRegion.t1();
-    // SelectedRegion guarantees t0 <= t1, so we need another boolean argument
-    // to indicate backwards play.
 
     if (backwards) {
         std::swap(t0, t1);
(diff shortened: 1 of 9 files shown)


Tantacrul:
@Eism I made a comment on Slack about the behaviour when a user double-clicks on a clip. We can discuss in the morning. 

I'll repeat here:

---

For regular wave selections, it works  as I'd expect.
- When I make a wave selection and press play, it plays through the selection and then stops. Great!

However, there's one issue
1. I don't think the play position should be affected when a user normally selects a full clip by clicking on its header - or if the user selects a group of clips either. Playback should not be affected by that.
2. 1. BUT, if you double-click on a waveform, which is something people do in AU3, in that specific circumstance, when the user pressed Play, it should play for the full selection and then stop.
3. 1. However, if point 2 is highly awkward, then we might consider going back to the idea that when you double-click on a waveform, it simply extends the wave selection to the ends of the clip, just like AU3 (see below), and we ditch the idea of double-clicking to select the full clip.

I'm curious whether either of these routes is difficult. I'm not sure which option I'd choose. 
@LWinterberg any thoughts?

src/playback/internal/au3/au3player.cpp
```diff
@@ -108,38 +106,15 @@ void Au3Player::play()
...
-                t0 = tracks.GetStartTime();
-            } else if (t0 > tracks.GetEndTime()) {
-                t0 = tracks.GetEndTime();
-            }
+        // move t0 to valid range
+        if (t0 < 0) {
+            t0 = tracks.GetStartTime();
+        } else if (t0 > tracks.GetEndTime()) {
+            t0 = tracks.GetEndTime();
         }
```
igorkorsukov, line 116: better to use `std::clame`

src/playback/internal/au3/au3player.cpp
```diff
@@ -319,6 +295,31 @@ muse::async::Channel<PlaybackStatus> Au3Player::playbackStatusChanged() const
...
+}
+
+void Au3Player::setPlaybackRegion(const PlaybackRegion& region)
+{
+    Au3Project& project = projectRef();
+
+    auto& playRegion = ViewInfo::Get(project).playRegion;
+    playRegion.SetStart(region.start);
+
+    if (region.start == region.end) {
```
igorkorsukov, line 313: unsafe fraction comparison

igorkorsukov (approved):

Issue #7580, opened 2024-10-18
Effects - Generate Noise 


ianalarina:
### Description

Implement Generate Noise in Audacity 4

chinakov:
Tested - noise generator is implemented.

PR #7738, opened 2024-11-18
Add Noise generator

kryksyh:
Resolves: [#7580](https://github.com/audacity/audacity/issues/7580)

Recommended:


src/effects/builtin/noisegen/noiseviewmodel.cpp
```diff
@@ -0,0 +1,90 @@
...
+    static const QStringList names {
+        tr("White", "not a color, but 'white noise' having a uniform spectrum"),
+        tr("Pink", "not a color, but 'pink noise' having a spectrum with more power in low frequencies"),
+        tr("Brownian", "a kind of noise spectrum also known as 'red' or 'brown'"),
+    };
+
+    return names;
+}
+
+QVariantMap NoiseViewModel::types() const
```
igorkorsukov, line 21: I recommend returning a list of items here, something like that
```
QVariantList NoiseViewModel::types() const
{
       QVariantList items; 
       if (items.empty()) {
               QVariantMap white; 
               white["text"] = tr("White", "not a color, but 'white noise' having a uniform spectrum");
               white["value"] = (int)NoiseEffect::Type::White;
               items << white;

               ...
       }
       return items;
}      
```

And then this list of items can be specified in the combo box as a model
```
ComboBox {
            id: typeSelector
            model: noise.types()
            textRole: "text"
            valueRole: "value"
}
```

src/effects/builtin/noisegen/noiseviewmodel.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return effect_->amplitude();
+}
+
+void NoiseViewModel::setAmplitude(double newAmplitude)
+{
+    NoiseEffect* const effect_ = effect();
+    IF_ASSERT_FAILED(effect_) {
+        return;
+    }
+    effect_->setAmplitude(newAmplitude);
```
igorkorsukov, line 56: I recommend doing the implementation as in reverb

src/effects/builtin/noisegen/noiseviewmodel.cpp
kryksyh: Thanks! That what I was looking for!

src/effects/builtin/noisegen/noiseviewmodel.cpp
kryksyh: will do

src/effects/builtin/noisegen/noiseviewmodel.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "log.h"
+
+using namespace au::effects;
+
+QStringList NoiseViewModel::typeNames() const
+{
+    static const QStringList names {
+        tr("White", "not a color, but 'white noise' having a uniform spectrum"),
+        tr("Pink", "not a color, but 'pink noise' having a spectrum with more power in low frequencies"),
+        tr("Brownian", "a kind of noise spectrum also known as 'red' or 'brown'"),
```
igorkorsukov, line 15: please use `muse::trc` functions (with context)

src/effects/builtin/noisegen/noisegenerator.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+size_t NoiseGenerator::Instance::ProcessBlock(EffectSettings& settings, const float* const* inBlock, float* const* outBlock,
+                                              size_t blockLen)
+{
+    return InstanceProcess(settings, inBlock, outBlock, blockLen);
+}
+
+bool NoiseGenerator::Instance::InstanceInit(EffectSettings& settings, double sampleRate, ChannelNames chanMap)
+{
+    (void)settings;
```
igorkorsukov, line 84: in "log.h" we have macro `UNUSED`

src/effects/builtin/noisegen/noiseviewmodel.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "log.h"
+
+using namespace au::effects;
+
+QVariantList NoiseViewModel::types() const
+{
+    static QVariantList items;
+    if (items.empty()) {
+        items.append(QVariantMap {
+            { "text", QString::fromStdString(muse::trc("White", "not a color, but 'white noise' having a uniform spectrum")) },
```
igorkorsukov, line 16: muse::trc("White", "not a color, but 'white noise' having a uniform spectrum") 
-> 
muse::trc("effects", "White", "not a color, but 'white noise' having a uniform spectrum")

src/effects/builtin/noisegen/noisegenerator.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+size_t NoiseGenerator::Instance::InstanceProcess(
+    EffectSettings& settings, const float* const*, float* const* outbuf, size_t size)
+{
+    auto& ns = GetSettings(settings);
+
+    float* buffer = outbuf[0];
+
+    float white;
+    float amplitude;
```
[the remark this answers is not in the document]
kryksyh: That's just a code from AU3 and those variables are initialized before the use. There is no valid value at the time of declaration.

src/effects/builtin/noisegen/noiseviewmodel.cpp
```diff
@@ -0,0 +1,76 @@
...
+    return settings().type;
+}
+
+void NoiseViewModel::setType(NoiseSettings::Type type)
+{
+    mutSettings().type = type;
+    emit typeChanged();
+}
+
+NoiseGenerator* NoiseViewModel::effect() const
```
igorkorsukov, line 55: It seems this method is no longer needed.

src/effects/builtin/noisegen/noiseviewmodel.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace au::effects {
+class NoiseGenerator;
+
+class NoiseViewModel : public GeneratorEffectModel
+{
+    Q_OBJECT
+
+    Q_PROPERTY(double amplitude READ amplitude WRITE setAmplitude NOTIFY amplitudeChanged)
+    Q_PROPERTY(NoiseSettings::Type type READ type WRITE setType NOTIFY typeChanged)
```
igorkorsukov, line 18: hmm, does this work?
to use a type in Qml, it needs to be registered in the meta system.

kryksyh: I believe that you need to register an enum if you want to expose literals. In this case it just decays to int. What should i do, register an enum or replace property type to int?

igorkorsukov: I think in this case it is easier to use int.
no need to use enums from the domain model in view (or we need to make two enums and convert between them), so as not to relation the domain model with the implementation of view.

src/effects/builtin/noisegen/noiseviewmodel.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "global/translation.h"
+
+using namespace au::effects;
+
+QVariantList NoiseViewModel::types() const
+{
+    static QVariantList items;
+    if (items.empty()) {
+        items.append(QVariantMap {
+            { "text", QString::fromStdString(muse::trc("effects/noise", "White",
```
igorkorsukov, line 15: we can use `qtrc` in this case 
see https://github.com/musescore/framework_tmp/blob/main/framework/global/translation.h

src/effects/builtin/noisegen/noisegenerator.h
saintmatthieu, line 1: I'd very much have liked to see one first commit where the code is just moved without changes, or just formatting, and another where the changes happened. (There must have been some, since the effect became stateless during this migration, right?)

src/effects/builtin/noisegen/noisegenerator.h
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    static constexpr Type typeDefault = Type::White;
+    static constexpr double amplitudeDefault = 0.8;
+
+    Type type { typeDefault };
+    double amplitude { amplitudeDefault };
+};
+
+class BUILTIN_EFFECTS_API NoiseGenerator : public GeneratorEffect, public EffectWithSettings <NoiseSettings, PerTrackEffect>
```
saintmatthieu, line 25: minor: the `BUILTIN_EFFECTS_API` macros and alike aren't necessary in AU4, where libraries are linked statically.

saintmatthieu (commented):
Unfortunately I do not have time now to finish reviewing this interesting PR.
If you could in the meantime give a sense of the changes needed to make this generator stateless that could come in handy when I resume, thanks.

src/effects/builtin/noisegen/noisegenerator.h
kryksyh: @saintmatthieu you know, to make the generator accessible by AU4 I did not change the original source in any way. I just made the simplest wrapper around NoiseBase

```
struct EffectSettings;

namespace au::effects {
class NoiseEffect : public ::NoiseBase, public GeneratorEffect
{
public:
    NoiseEffect();

    enum class Type {
        White,
        Pink,
        Brownian,
    };

    static const ComponentInterfaceSymbol Symbol;

    ComponentInterfaceSymbol GetSymbol() const override;

    Type type()   const;
    void setType(Type type);
    double amplitude() const;
    void setAmplitude(double amplitude);

private:
    void doInit() override;
};
}
```

saintmatthieu (approved):
I'm happy with the backend side of this PR. I see that @igorkorsukov is thoroughly reviewing the front-end, too, I'll leave that part of the review to him. Please await his approval before moving on.

src/effects/builtin/noisegen/noisegenerator.h
kryksyh: > Next PR you do this just please add this intermediary step in your commit list.

I believe this wouldn't be practical, as the second approach is not based on the first one. 

> NoiseGenerator essentially reused NoiseBase, but with public GeneratorEffect, and inheriting EffectWithSettings <NoiseSettings, PerTrackEffect> instead of StatefulPerTrackEffect, making it stateless

that's correct. `Process` functions were moved to the Instance class, and member variables were replaced with settings. And that's it.

igorkorsukov (approved):

Issue #7668, opened 2024-11-06
It's possible to extend a clip beyond the zero time mark when un-trimming


dozzzzer:
### Steps to reproduce

1. Trim a clip
2. Move it close enough to the zero time mark so that the left trim handle is visible
3. Grab the left trim handle and move it leftwards as far as you can
4. Observe

### Expected behavior

As soon as it reaches the zero time mark, it cannot go any further.

### Actual behavior

It can go past the zero time mark. Upon clicking the clip, its left edge gets aligned with the zero.

https://github.com/user-attachments/assets/0b6eeaa1-f4f9-4e41-aaf1-56e59cab9558

### Audacity Version

Audacity 4 alpha

### Operating system

Windows, macOS, Linux

dozzzzer:
@DilsonsPickles Design is needed for this one

DilsonsPickles:
Surely the solution here is just to prohibit the User from trimming a clip beyond 0?

PR #7754, opened 2024-11-20
Prevent untrimming left beyond zero

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7668


src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -440,6 +440,10 @@ bool ClipsListModel::trimLeftClip(const ClipKey& key, bool completed)
         return false;
     }
 
+    if (muse::RealIsEqualOrLess(newStartTime, 0.0)) {
+        newStartTime = 0.0;
```
Eism, line 444: I think it would be useful to move this adjustment above the check with `item->clip().startTime`, and I also recommend using std::max

Eism (approved):

PR #7785, opened 2024-11-25
7647 effect panel part 2

### LINKED ISSUE #7647: Effect panel (part 2)
### Description

See how much of the effect panel design implementation can fit in a reasonable pull request. (This description will be made more concrete _a posteriori_ `:)` )

saintmatthieu:
Resolves: #7647 

Recommended:

## QA:
Here is the [figma spec](https://www.figma.com/design/w2kAh1gKKV5bu7htY7158W/08---Effects?node-id=1-4&node-type=canvas&t=rgnwM1ZZEvSANEgM-0). Not every aspect of this spec is implemented yet. Here is a list of what is known _not_ to be yet implemented. It is meant to be exhaustive, so if you find something not implemented that's not in this list, please raise it.
* master effects aren't there yet
* power buttons are yet ineffective
* the effect panel can yet only be opened via the track header
* the effect panel can yet only be closed via the effect panel header
* the "Get more effects" option is not there yet
* the UI window of effects isn't yet summonable, neither when adding an effect nor when clicking on the button with the effect's name (basically you're stuck with the effect's default parameters)
* at the moment, the "No effect" behaviour is the same as in AU3, (i.e., there isn't the "Delete effect" option yet)
* reordering of effects isn't yet implemented
* there is no way of scrolling the effect list when it gets taller than the available space

Aside this, everything should be there.
Things to test not part of the spec:
- [x] effects can be added, removed or replaced during playback

Known issues that can be addressed later:
* changing a track's name isn't reflected in the effect panel automatically

### DIFF
--- a/src/effects/effects_base/internal/effectsactionscontroller.cpp
+++ b/src/effects/effects_base/internal/effectsactionscontroller.cpp
@@ -2,6 +2,7 @@
 * Audacity: A Digital Audio Editor
 */
 #include "effectsactionscontroller.h"
+#include "effects/effects_base/effectstypes.h"
 
 #include "log.h"
 
@@ -13,6 +14,10 @@ void EffectsActionsController::init()
 {
     dispatcher()->reg(this, EFFECT_OPEN_CODE, this, &EffectsActionsController::doEffect);
     dispatcher()->reg(this, "repeat-last-effect", this, &EffectsActionsController::repeatLastEffect);
+    dispatcher()->reg(this, "realtimeeffect-add", this, &EffectsActionsController::addRealtimeEffect);
+    dispatcher()->reg(this, "realtimeeffect-remove", this, &EffectsActionsController::removeRealtimeEffect);
+    dispatcher()->reg(this, "realtimeeffect-replace", this, &EffectsActionsController::replaceRealtimeEffect);
+
     effectExecutionScenario()->lastProcessorIsNowAvailable().onNotify(this, [this] {
         m_canReceiveActionsChanged.send({ "repeat-last-effect" });
     });
@@ -31,6 +36,59 @@ void EffectsActionsController::doEffect(const muse::actions::ActionData& args)
 void EffectsActionsController::repeatLastEffect()
 {
     effectExecutionScenario()->repeatLastProcessor();
+}
+
+void EffectsActionsController::addRealtimeEffect(const muse::actions::ActionData& args)
+{
+    IF_ASSERT_FAILED(args.count() == 2) {
+        return;
+    }
+
+    const auto project = globalContext()->currentProject();
+    IF_ASSERT_FAILED(project) {
+        // Command issued without an open project ?..
+        return;
+    }
+
+    const auto effectId = args.arg<EffectId>(0);
+    const auto trackId = args.arg<TrackId>(1);
+    realtimeEffectService()->addRealtimeEffect(*project, trackId, effectId);
+}
+
+void EffectsActionsController::removeRealtimeEffect(const muse::actions::ActionData& args)
+{
+    IF_ASSERT_FAILED(args.count() == 2) {
+        return;
+    }
+
+    const auto project = globalContext()->currentProject();
+    IF_ASSERT_FAILED(project) {
+        // Command issued without an open project ?..
+        return;
+    }
+
+    const auto trackId = args.arg<TrackId>(0);
+    const auto effectStateId = args.arg<EffectStateId>(1);
+    realtimeEffectService()->removeRealtimeEffect(*project, trackId, effectStateId);
+}
+
+void EffectsActionsController::replaceRealtimeEffect(const muse::actions::ActionData& args)
+{
+    IF_ASSERT_FAILED(args.count() == 3) {
+        return;
+    }
+
+    const auto project = globalContext()->currentProject();
+    IF_ASSERT_FAILED(project) {
+        // Command issued without an open project ?..
+        return;
+    }
+
+    const auto trackId = args.arg<TrackId>(0);
+    const auto srcIndex = args.arg<int>(1);
+    const auto dstEffectId = args.arg<EffectId>(2);
+
+    realtimeEffectService()->replaceRealtimeEffect(*project, trackId, srcIndex, dstEffectId);
 }
 
 bool EffectsActionsController::canReceiveAction(const muse::actions::ActionCode& code) const
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -30,6 +30,13 @@ using namespace au::effects;
 
 static const char16_t* VIEWER_URI = u"audacity://effects/viewer?type=%1&instanceId=%2";
 
+void EffectsProvider::init()
+{
+    builtinEffectsRepository()->effectMetaListUpdated().onNotify(this, [this] {
+        reloadEffects();
+    });
+}
+
 bool EffectsProvider::isVstSupported() const
 {
     return vstEffectsRepository() ? true : false;
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -112,6 +112,15 @@ EffectMeta EffectsProvider::meta(const muse::String& effectId) const
     return EffectMeta();
 }
 
+std::string EffectsProvider::effectName(const std::string& effectId) const
+{
+    const auto desc = PluginManager::Get().GetPlugin(effectId);
+    if (!desc) {
+        return "";
+    }
+    return desc->GetSymbol().Msgid().Translation().ToStdString();
+}
+
 muse::Ret EffectsProvider::showEffect(const muse::String& type, const EffectInstanceId& instanceId)
 {
     LOGD() << "try open effect: " << type << ", instanceId: " << instanceId;
--- a/src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
@@ -1,0 +1,134 @@
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+
+import Muse.Ui
+import Muse.UiComponents
+
+ListItemBlank {
+    id: root
+
+    property var item: null
+    property var availableEffects: null
+    property var handleMenuItemWithState: null
+
+    height: 24
+    clip: false // should be true?
+    background.color: "transparent"
+    hoverHitColor: "transparent"
+
+    RowLayout {
+        anchors.fill: parent
+        spacing: 4
+
+        FlatButton {
+            id: gripButton
+
+            Layout.preferredWidth: 14
+            Layout.preferredHeight: 16
+            Layout.alignment: Qt.AlignVCenter
+            backgroundRadius: 0
+
+            visible: true
+            transparent: true
+            hoverHitColor: ui.theme.buttonColor
+            mouseArea.cursorShape: Qt.SizeAllCursor
+
+            // do not accept buttons as FlatButton's mouseArea will override
+            // DockTitleBar mouseArea and effect will not be draggable
+            mouseArea.acceptedButtons: Qt.NoButton
+
+            contentItem: StyledIconLabel {
+                iconCode: IconCode.TOOLBAR_GRIP
+            }
+
+        }
+
+        FlatButton {
+            id: powerButton
+
+            // TODO: make that from TrackEffectsSection reusable
+            Layout.margins: 0
+            Layout.alignment: Qt.AlignVCenter
+            Layout.preferredWidth: root.height
+            Layout.minimumHeight: root.height
+            Layout.maximumHeight: root.height
+            height: width
+
+            icon: IconCode.BYPASS
+            iconFont: ui.theme.toolbarIconsFont
+
+            accentButton: true
+
+            onClicked: {
+                accentButton = !accentButton
+            }
+        }
+
+        // Wrappinng a FlatButton in a Rectangle because of two bad reasons:
+        // * I don't find a `border` property for `FlatButton`
+        // * Somehow, the button's color is a bit darkened it direct child of the RowLayout
+        Rectangle {
+            id: effectNameRect
+
+            border.color: ui.theme.strokeColor
+            border.width: 1
+            radius: effectNameButton.backgroundRadius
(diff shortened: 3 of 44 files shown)


src/au3audio/CMakeLists.txt
```diff
@@ -28,6 +28,6 @@ include(${CMAKE_CURRENT_LIST_DIR}/../au3wrap/au3defs.cmake)
 
 set(MODULE_INCLUDE ${AU3_INCLUDE})
 set(MODULE_DEF ${AU3_DEF})
-set(MODULE_LINK au3wrap)
+set(MODULE_LINK au3wrap trackedit)
```
igorkorsukov, line 31: Why was it necessary to link this module?
This looks unexpected, it is alarming that the audio module depends on the track editing module

src/au3audio/audioenginetypes.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+#pragma once
+
+#include "global/types/string.h"
+#include <optional>
+
+namespace au::audio {
+using RtEffectState = void*;
+
+using EffectId = muse::String;
```
igorkorsukov, line 12: I think better to just include `effects/effects_base/effectstypes.h`

src/au3audio/internal/audioengine.cpp
```diff
@@ -3,18 +3,46 @@
...
 
+namespace {
+std::string getEffectName(const std::string& effectId)
+{
+    const auto desc = PluginManager::Get().GetPlugin(effectId);
+    IF_ASSERT_FAILED(desc) {
+        return "";
+    }
+    return desc->GetSymbol().Msgid().Translation().ToStdString();
+}
```
igorkorsukov, line 33: It is better to add the required method to the `IEffectsProvider` and use it through injection

src/au3audio/audioenginetypes.h
```diff
@@ -0,0 +1,41 @@
...
+enum class RtEffectListEventType
+{
+    Added,
+    Removed,
+    Replaced,
+};
+
+using TrackId = long;
+
+struct RtEffectListItem {
```
Eism, line 23: better not to use the abbreviation, at first I didn't understand what rt is

src/au3audio/internal/audioengine.cpp
```diff
@@ -3,18 +3,46 @@
...
+
+std::string getEffectName(const RealtimeEffectState& state)
+{
+    return getEffectName(state.GetID().ToStdString());
+}
+
+std::string getTrackName(const au::au3::Au3Project& project, au::audio::TrackId trackId)
+{
+    return au::au3::DomConverter::track(au::au3::DomAccessor::findTrack(project, au::au3::Au3TrackId { trackId })).title.toStdString();
+}
```
igorkorsukov, line 43: It is better to add the required method to the `ITrackeditProject` and use it by 
```
globalContext()->currentTrackeditProject()->....
```

src/au3audio/audioenginetypes.h
Eism: I am also confused by the Item in the audio module. Item is about UI, as for me, the audio module should not operate with such concepts

src/au3audio/internal/audioengine.cpp
```diff
@@ -47,6 +75,99 @@ static ProjectAudioIO::DefaultOptions::Scope s_defaultOptionsScope {
...
+            return;
+        }
+        auto au3Project = reinterpret_cast<au::au3::Au3Project*>(project->au3ProjectPtr());
+        m_tracklistSubscription = TrackList::Get(*au3Project).Subscribe([this, w = weak_from_this()](const TrackListEvent& e) {
+            const auto lifeguard = w.lock();
+            if (lifeguard) {
+                onTrackListEvent(e);
+            }
+        });
+    });
```
igorkorsukov, line 91: I believe there is a better place for all this. The audio engine is a low-level utility service, it is unexpected to see it accessing the global context, working with tracks, subscribing to track events...  

I think that all this added code needs to be moved to the `effects_base` module, adding a new interface, something like `IRealtimeEffects` and its implementation.

src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -38,6 +43,59 @@ void EffectsActionsController::repeatLastEffect()
...
+
+void EffectsActionsController::replaceRealtimeEffect(const muse::actions::ActionData& args)
+{
+    IF_ASSERT_FAILED(args.count() == 3) {
+        return;
+    }
+
+    const auto project = globalContext()->currentProject();
+    IF_ASSERT_FAILED(project) {
+        // Command issued without an open project ?..
```
Eism, line 88: these comments are useless, you set for them `UiCtxProjectOpened` context, so this won't happen and the likelihood that someone will encounter this is low (I am not sure that someone will change their context)

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
```diff
@@ -0,0 +1,134 @@
...
+
+ListItemBlank {
+    id: root
+
+    property var item: null
+    property var availableEffects: null
+    property var handleMenuItemWithState: null
+
+    height: 24
+    clip: false // should be true?
```
Eism, line 16: the default value is false, so you don't need to set it here
we use true in rare cases, and this is not your case

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
```diff
@@ -0,0 +1,134 @@
...
+    background.color: "transparent"
```
Eism, line 17: you can use `normalColor`

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
```diff
@@ -0,0 +1,134 @@
...
+
+            accentButton: true
+
+            onClicked: {
+                accentButton = !accentButton
+            }
+        }
+
+        // Wrappinng a FlatButton in a Rectangle because of two bad reasons:
+        // * I don't find a `border` property for `FlatButton`
```
Eism, line 69: If you need a border for FlatButton you should override backgroundItem

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
```diff
@@ -0,0 +1,134 @@
...
+            id: chooseEffectRect
+
+            border.color: ui.theme.strokeColor
+            border.width: 1
+            radius: effectNameButton.backgroundRadius
+            Layout.fillHeight: true
+            Layout.preferredWidth: height
+
+            FlatButton {
+                id: chooseEffectDropdown
```
Eism, line 114: you can use MenuButton for it

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectList.qml
```diff
@@ -0,0 +1,45 @@
...
+    id: root
+
+    property alias preferredHeight: trackEffectList.height
+
+    visible: trackEffectList.count > 0
+
+    RtEffectListModel {
+        id: trackEffectListModel
+        trackId: view.itemAtIndex(effectsPanel.selectedTrackIndex).item.trackId
+        Component.onCompleted: {
```
Eism, line 23: better to use Component.onCompleted for root item

src/au3audio/audioenginetypes.h
saintmatthieu: I already have a `RealtimeEffectListItem.qml` :) Wanted to avoid the two getting confused.
I agree that "list item" does sound UI-like indeed. However, the sequence of realtime effects we set on the audio engine (or in the effect module, debate is ongoing) needs a name. What do you suggest ? "Sequence" instead of "List" would do. "Entry" instead of "Item", too. Pick your combination :)

src/au3audio/internal/audioengine.cpp
saintmatthieu: Ok.

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectList.qml
saintmatthieu: Why ?

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
saintmatthieu: Awesome, thanks !

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
saintmatthieu: Thanks for the suggestion, I looked into this.
The default for FlatButton is that Component of ID `defaultBackgroundComponent`. It does everything I want my button to do, except for the border thing (it uses `ui.theme.borderWidth`, which equals 0, though). I haven't found a readily solution whereby I can reuse it without duplicating it or changing the framework to expose that property in the FlatButton. That'd be the right way, but is maybe there's another correct way ?

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
saintmatthieu: ... which is the default, so I can remove this line. Thanks !

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
Eism: I will suggest duplicating, overriding backgroundItem is done exactly for such cases, so as not to create a lot of properties

saintmatthieu: Yeah, but then I have a button whose only difference to the rest of the UI is that it has a non-zero border. If I duplicate it and `MenuButton` changes on the framework, my button will not get automatically updated.
So there seems not to be a way I overlooked. Will go for the duplication now but will look into making a PR in the framework.

saintmatthieu: I just gave it a try: it's not enough to just duplicate `defaultBackgroundComponent`, I'd have to duplicate the entire `FlatButton`.
So nevermind, for now, it'll be the rectangle wrapper.

>  overriding backgroundItem is done exactly for such cases, so as not to create a lot of properties

And how about
FlatButton.qml:
```qml
FocusScope {
    property Component backgroundItem: defaultBackgroundComponent
    property var backgroundComponentLoaderItem: backgroundLoader.item // pity backgroundItem was not called `backgroundComponent` ...
    
    Loader {
        id: backgroundLoader
        sourceComponent: root.backgroundItem
    }
```
and then
```qml
  FlatButton {
      Component.onCompleted: {
          backgroundComponentLoaderItem.border.width = 1
      }
  }
```
This only exposes one more property and gives the client control over the properties of the default item. (Only a pity that I have to use `Component.onCompleted`)

src/au3audio/CMakeLists.txt
```diff
@@ -28,6 +28,6 @@ include(${CMAKE_CURRENT_LIST_DIR}/../au3wrap/au3defs.cmake)
 
 set(MODULE_INCLUDE ${AU3_INCLUDE})
 set(MODULE_DEF ${AU3_DEF})
-set(MODULE_LINK au3wrap)
+set(MODULE_LINK au3wrap trackedit effects_base)
```
igorkorsukov, line 31: just in case, if we use data types from another module that are described only in the `.h` file, or interfaces, then we do not need to link these modules.
In the current implementation of AU4, it is usually necessary to link only au3wrap and global (linked automatically)

igorkorsukov (approved):

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
saintmatthieu: Like I said,
> I just gave it a try: it's not enough to just duplicate defaultBackgroundComponent, I'd have to duplicate the entire FlatButton.

src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -109,6 +116,15 @@ EffectMeta EffectsProvider::meta(const muse::String& effectId) const
     return EffectMeta();
 }
 
+std::string EffectsProvider::effectName(const std::string& effectId) const
+{
+    const auto desc = PluginManager::Get().GetPlugin(effectId);
```
Eism, line 121: can you get it from EffectMeta?

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
Eism: show how you tried to duplicate

src/au3audio/audioenginetypes.h
```diff
@@ -0,0 +1,26 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+#pragma once
+
+#include "global/types/string.h"
+#include <optional>
+
+namespace au::audio {
+using EffectState = void*;
```
Eism, line 10: Magic structure
Can we describe it? create our structure with important fields

src/projectscene/qml/Audacity/ProjectScene/trackspanel/RealtimeEffectListItem.qml
saintmatthieu: I threw the experiment away. You can see for yourself that defaultBackgroundComponent has dependencies on FlatButton.

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: Also.

src/au3audio/audioenginetypes.h
saintmatthieu: Was in fact just used as identifier. Just using `uintptr_t` is better indeed.

Eism (approved):

Issue #7714, opened 2024-11-13
If you zoom in past a certain point in Minute and Second mode, the timeline values become messed up


dozzzzer:
https://github.com/user-attachments/assets/95bd61e0-4179-4ad8-98ea-ec79b1cedfaa

dozzzzer:
Tested

Issue #7709, opened 2024-11-13
Timelineruler/gridlines calculation bug


grliszas14:
see how ruler values changes on the video:

https://github.com/user-attachments/assets/9d3adeeb-11c3-4d78-b5fa-2fc85ac3cfdc

grliszas14:
QA bulletpoints:
This can be tested along with https://github.com/audacity/audacity/issues/7714
* timeline ruler labels should not overlap each other on any zoom level
* labels/ticks should not shift left/right by one tick width at any time
* pay attention to the first gridline on the left, it should be better handled than in current master (should appear sooner when you scroll horizontally)
* label on the left edge you gracefully wrap underneath left edge of clips view instead of disappearing as soon as it touches an edge (when you scroll horizontally)

dozzzzer:
Tested

PR #7805, opened 2024-11-27
Timeline ruler improvements

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7709
Resolves: https://github.com/audacity/audacity/issues/7714

### DIFF
--- a/src/projectscene/view/timeline/timeformat.cpp
+++ b/src/projectscene/view/timeline/timeformat.cpp
@@ -8,166 +8,57 @@
 
 using namespace au::projectscene;
 
+const std::vector<std::pair<double, IntervalInfo>> zoomToIntervalInfo = {
+    // { min pixel width per minorMinor tick, {major, minor, minorMinor, digits} }
+    { 32 / 0.000005, { 0.0001, 0.00002, 0.000005, 6 } },            // 0.000005s ticks
+    { 28 / 0.00001, { 0.0005, 0.00005, 0.00001, 6 } },              // 0.00001s ticks
+    { 32 / 0.00005, { 0.001, 0.0002, 0.00005, 5 } },                // 0.00005s ticks
+    { 28 / 0.0001, { 0.005, 0.0005, 0.0001, 4 } },                  // 0.0001s ticks
+    { 28 / 0.0005, { 0.005, 0.0025, 0.0005, 4 } },                  // 0.0005s ticks
+    { 32 / 0.001, { 0.05, 0.005, 0.001, 3 } },                      // 0.001s ticks
+    { 38 / 0.005, { 0.05, 0.01, 0.005, 3 } },                       // 0.005s ticks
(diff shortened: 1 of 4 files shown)


src/projectscene/view/timeline/timeformat.cpp
```diff
@@ -12,72 +12,86 @@ IntervalInfo TimeFormat::intervalInfo(TimelineContext* context)
 {
     IntervalInfo timeInterval;
     double zoom = context->zoom();
-    // NOTE: 24 is minimal px size between minorMinor ticks
+    // NOTE: 32 is minimal px size between minorMinor ticks
     // 0.000005 is distance [s] between minorMinor ticks
     // same goes for other time intervals
-    if (zoom > (24 / 0.000005)) { // 0.000005s ticks
+    if (zoom > (32 / 0.000005)) { // 0.000005s ticks
```
kryksyh, line 18: @grliszas14 what do you think about converting this to lookup table?

example totally not written by llm
```
struct IntervalInfo {
    double minor;
    double major;
    double minorMinor;
    int digits;
};

const std::vector<std::pair<double, IntervalInfo>> zoomToIntervalInfo = {
    {32 / 0.000005, {0.00002, 0.0001, 0.000005, 6}},
    {28 / 0.00001, {0.00005, 0.0005, 0.00001, 6}},
    // ... other entries ...
};

IntervalInfo TimeFormat::intervalInfo(TimelineContext* context) {
    double zoom = context->zoom();

    // Find the appropriate interval info using binary search or linear search
    // based on the zoom level.
    auto it = std::lower_bound(zoomToIntervalInfo.begin(), zoomToIntervalInfo.end(), zoom,
                              [](const auto& pair) { return pair.first > zoom; });

    if (it == zoomToIntervalInfo.end()) {
        // Handle the case where zoom is larger than the largest zoom value in the table
        return zoomToIntervalInfo.back().second;
    } else {
        return it->second;
    }
}
```

grliszas14: Sounds good for minutes&seconds, changed it. I'm not so sure if to apply it as well to beats&measures, I don't think we're gonna gain any clarity there

src/projectscene/view/timeline/timeformat.cpp
```diff
@@ -8,166 +8,53 @@
...
+
+    if (it != zoomToIntervalInfo.end()) {
+        return it->second;
     }
 
-    timeInterval.minor = 24 * 3600 + 1; // HACK to not to draw minor ticks labels
-    timeInterval.major = 7 * 24 * 3600.0;
-    timeInterval.minorMinor = 24 * 3600.0;
-    return timeInterval;
+    return { 24 * 3600 + 1, 7 * 24 * 3600.0, 24 * 3600.0, 0 };
```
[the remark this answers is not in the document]
grliszas14: Done, but at the bottom of the list with 0 value 👍

kryksyh (approved):

Issue #7757, opened 2024-11-20
"Eating" a clip by un-trimming another clip misbehaves


dozzzzer:
### Steps to reproduce

1. Put two clips next to each other
2. Select the entire left-hand clip
3. Trim the right-hand one and move it closer 
4. Start un-trimming the right-hand clip so that it starts overlapping the left one

### Expected behavior

No flickering and the right-hand clip goes on top of the left-hand one

### Actual behavior

The clip flickers and the right-hand clip sometimes goes underneath, sometimes on top of the left-hand clip.

https://github.com/user-attachments/assets/a88fddaf-8c1b-40da-b25a-446740107283

### Audacity Version

Audacity 4 alpha

### Operating system

Windows, macOS, Linux

grliszas14:
It may be a duplicate of: https://github.com/audacity/audacity/issues/7657

grliszas14:
QA bulletpoints:
Both issues mentioned in the description should be fixed:
* no flickering when using hover trim handles
* currently edited clip should be on top

* additionally improved trim mouse cursor - it should stay being trim cursor no matter where you point your mouse at as long as you keep your mouse btn pressed
* fixed crash on deleting single selected clip (it was re-introduced somewhere along working with BPM changes)

chinakov:
Thanks, this is fixed.

PR #7810, opened 2024-11-28
Trim hover handles improvements

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7757

### DIFF
--- a/src/trackedit/internal/trackeditactionscontroller.cpp
+++ b/src/trackedit/internal/trackeditactionscontroller.cpp
@@ -336,6 +336,9 @@ void TrackeditActionsController::clipDelete(const ActionData& args)
         return;
     }
 
+    //! TODO AU4: improve for multiclip selection
+    selectionController()->resetSelectedClips();
+
     trackeditInteraction()->removeClip(clipKey);
 }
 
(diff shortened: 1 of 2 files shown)


src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -336,6 +336,9 @@ void TrackeditActionsController::clipDelete(const ActionData& args)
         return;
     }
 
+    //! TODO AU4: improve for multiclip selection
+    selectionController()->resetSelectedClips();
```
Eism, line 340: let's do this inside selectionController, I see we're already updating the state for the tracks there, see `Au3SelectionController::init`

grliszas14: I'd opt for leaving it as is, because:
1. we're using selectionController this way in a few more functions here - and it seems ok to me btw, it's easily readable/understandable that some action is clearing the selection
2. In this particular function we're getting clipKey as an argument -> we'll be able to easily modify multi-selection by simply removing this clipKey from selectedClips vector. If we move it to `Au3SelectionController::init` -> we'll  rely on `trackChanged()` notification -> hence we'll need to loop over every clip to find the one that was removed
3. Not every `trackChanged` should result in resetting selection - so that may cause writing a lot of additional logic in `init()` func

wdyt?

Eism (approved):

src/trackedit/internal/trackeditactionscontroller.cpp
Eism: ok

Issue #7761, opened 2024-11-20
[Effects] Working with presets


igorkorsukov:
### Description

Make functionality for working with presets using presets for reverb as an example.

chinakov:
Tested:
- Default presets are the same as they are in AU3
- New presets can be created, loaded, deleted, and are saved after re-launch
- Export/Import of presets is functional, and is backwards compatible with Audacity 3

Issues noticed:
- If the User decides to save a preset with a name that already exists, the preset is not overwritten, and the User is not informed about this. In Audacity 3, the User will be shown a 'Preset already exists' dialog window that allows the User to replace the old preset.

chinakov:
Passing this task, create a separate task for the issue noticed: https://github.com/audacity/audacity/issues/7869

PR #7823, opened 2024-12-02
Added effect presetsmenu, actions and presets controller. Step 2

igorkorsukov:
Resolves: part of #7761

### DIFF
--- a/src/effects/builtin/reverb/ReverbView.qml
+++ b/src/effects/builtin/reverb/ReverbView.qml
@@ -24,7 +24,6 @@ EffectBase {
     }
 
     Component.onCompleted: {
-        console.log("root.instanceId: " + root.instanceId)
         reverb.init()
     }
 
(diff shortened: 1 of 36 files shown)


src/effects/builtin/reverb/ReverbView.qml
```diff
@@ -8,24 +8,23 @@ import "../common"
...
-    }
+    model: reverb
 
     ReverbViewModel {
         id: reverb
+        instanceId: root.instanceId
     }
 
     Component.onCompleted: {
+        console.log("root.instanceId: " + root.instanceId)
```
kryksyh, line 27: is it a debugging leftover?

kryksyh (approved):

src/effects/builtin/reverb/ReverbView.qml
igorkorsukov: removed, thank you!

Issue #7801, opened 2024-11-26
Implement applying an effect to the selected clips


dozzzzer:
### Description

Implement the application of an effect to the selected clips, including the selection of non-consecutive clips.

QA: also test the ability to run a preview on a clip selected by double-clicking.

saintmatthieu:
#7799 should probably be fixed first: how IDs of tracks and clips behave when they are replaced should be clarified first - see [this comment](https://github.com/audacity/audacity/issues/7799#issuecomment-2528137834)

PR #7835, opened 2024-12-04
7801 effect on discontinuous selection

saintmatthieu:
Resolves: #7801

For a list of requirements of this feature (which may provide a rationale to the chosen changes), please refer to the QA list below.

There are three commits to this PR:
1. An AU3 capability extension whereby which clips should be read from or written to can be specified. (@kryksyh this one is for you)
2. A change in AU4 which changes the behavior of `selectedClipStartTime()` and `selectedClipEndTime()`: now it returns start of earliest selected clips / end of last-ending selected clip. This change should not affect the places where this method was used before, but @grliszas14 please double-check.
3. A change in the effect provider API and implementation to communicate the list of selected clips. @igorkorsukov I could also have injected the selection controller into the effects provider, not sure which approach is generally preferred, please let me know.

Recommended:

QA:
- [ ] Applying effects with time selection (as opposed to clip selection) works as before
- [ ] Whether selecting contiguous clips or using a time selection over the same clips should yield the same result
- [ ] In the case of discontinuous selection, clips _between_ should _not influence the effect_. It should behave as though there were no clips (i.e., silence) in between. For instance, in the situation of clip sequence A, B, C, with A and C being selected, B should not _leak_ into C (even if the effect has a long tail, e.g. a strong reverb). (Please approach me for clarification if needed.)
- [ ] Effect preview should work like before
- [ ] After selecting clips and applying an effect, clips are deselected and a time selection appears instead. Please check that this is not a regression. (Should be addressed at some stage, but not in this PR if it isn't below.)
- [ ] Works also when selected clips are on different tracks

### DIFF
--- a/au3/libraries/lib-builtin-effects/Generator.cpp
+++ b/au3/libraries/lib-builtin-effects/Generator.cpp
@@ -23,7 +23,7 @@
 
 #include "TimeWarper.h"
 
-bool Generator::Process(EffectInstance &, EffectSettings &settings)
+bool Generator::Process(EffectInstance &, EffectSettings &settings, std::vector<int64_t> /* whichClips */)
 {
    const auto duration = settings.extra.GetDuration();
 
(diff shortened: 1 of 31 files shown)


saintmatthieu:
Rework adopting @igorkorsukov 's suggestion ; closing PR for now.

Issue #7819, opened 2024-11-29
Show a warning message when a user tries to save a project created in an older version of Audacity


dozzzzer:
### Problem to be solved

At the moment, Audacity "silently" saves the project created in an older version of Audacity without the user being aware of it, which, due to our lack of backwards compatibility, makes it impossible for the user to open the file with an older version of Audacity than the one used to save the project.

### Your idea

To show the warning message that would inform the user that saving such a project would make it incompatible with the older versions or, in addition to that, let the user save the project file as a copy.

@LWinterberg @Tantacrul please clarify what should be done here.

### Prior art

<img width="446" alt="image" src="https://github.com/user-attachments/assets/98d07c8b-3946-4c5d-935f-6e06ac43dd69">

PR #7847, opened 2024-12-05
Notify the user if the project version will be updated

kryksyh:
Resolves: #7819 

Recommended:

### DIFF
--- a/libraries/lib-project-file-io/ProjectFileIO.cpp
+++ b/libraries/lib-project-file-io/ProjectFileIO.cpp
@@ -785,6 +785,15 @@ bool ProjectFileIO::CheckVersion()
       );
       return false;
    }
+   else if (version < BaseProjectFormatVersion) {
+      using namespace BasicUI;
+      wxString currentVersionStr = wxString::Format("%u.%u", BaseProjectFormatVersion.Major, BaseProjectFormatVersion.Minor);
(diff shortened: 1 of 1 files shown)


saintmatthieu (approved):

libraries/lib-project-file-io/ProjectFileIO.cpp
```diff
@@ -785,6 +785,20 @@ bool ProjectFileIO::CheckVersion()
...
+      bool updateVersion = (MessageBoxResult::Yes == ShowMessageBox(
+         XO("This project was created using an older Audacity version. "
+            "Once saved, the project can only be opened with Audacity version %s or newer.\n\n"
+            "Do you want to continue? This action cannot be undone.").Format(currentVersionStr),
+         MessageBoxOptions{}
+            .Caption(XO("Update Audacity project version?"))
+            .ButtonStyle(Button::YesNo)));
+      if ( !updateVersion ) {
+         return false;
+      }
```
LWinterberg, line 800: i think we should skip this bit: 

* Yes/no has the issue of always requiring quite a bit of brain energy to parse. 
* I don't think "this action cannot be undone" is true - if you close the project without saving, you won't be updating the project version, will you?

I think we can just keep the upper part with an OK dialog box.

kryksyh: Updated

PR #7871, opened 2024-12-09
EffectOutputTracks copies clip IDs when duplicating

### LINKED ISSUE #7799: A project cannot be opened once an effect has been applied to its audio
### Bug description

_No response_

### Steps to reproduce

1. Open a project created in AU3 containing audio data
2. Make a selection and apply Reverb to it
3. Ctrl + W
4. "Don't save"
5. Try to reopen the project

### Expected behavior

The project can be reopened.

### Actual behavior

The project cannot be reopened. Clicking the project thumbnail and selecting the project via Ctrl + O don't do anything.

### Audacity Version

Audacity 4 alpha

### Operating system

Other

### Additional context

_No response_

saintmatthieu:
Resolves: #7799

Addresses the case of EffectOutputTracks, where, when duplicating the original track, the clip IDs should match the original.
Problem discussed internally [here](https://mu--se.slack.com/archives/C06RX437GJV/p1731925325477879).

Recommended:

### DIFF
--- a/au3/libraries/lib-effects/EffectOutputTracks.cpp
+++ b/au3/libraries/lib-effects/EffectOutputTracks.cpp
@@ -11,6 +11,7 @@
 #include "BasicUI.h"
 #include "SyncLock.h"
 #include "UserException.h"
+#include "WaveClip.h"
 #include "WaveTrack.h"
 #include "TimeStretching.h"
 
@@ -41,6 +42,19 @@ EffectOutputTracks::EffectOutputTracks(
 
    for (auto aTrack : trackRange) {
       auto pTrack = aTrack->Duplicate();
+      if (auto aWaveTrack = dynamic_cast<WaveTrack*>(aTrack))
+      {
+         auto pWaveTrack = dynamic_cast<WaveTrack*>(pTrack.get());
+         for (auto i = 0; i < aWaveTrack->GetNumClips(); ++i)
+         {
+            auto aClip = aWaveTrack->GetClip(i);
+            auto pClip = pWaveTrack->GetClip(i);
+            if (aClip && pClip)
+               pClip->SetId(aClip->GetId());
+            else
+               assert(false);
+         }
+      }
       mIMap.push_back(aTrack);
       mOMap.push_back(pTrack.get());
       mOutputTracks->Add(pTrack);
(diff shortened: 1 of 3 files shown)


au3/libraries/lib-effects/EffectOutputTracks.cpp
```diff
@@ -41,6 +42,19 @@ EffectOutputTracks::EffectOutputTracks(
 
    for (auto aTrack : trackRange) {
       auto pTrack = aTrack->Duplicate();
+      if (auto aWaveTrack = dynamic_cast<WaveTrack*>(aTrack))
```
kryksyh, line 45: this makes me think that there are other places where duplicated track should also duplicate clip ids. I'm not opposed to have this merged, but what do you think about introduce this as an optional behavior to `Duplicate`?

saintmatthieu: You find that this PR solves this ID problem only in the context of effects but that preserving clip IDs might be desired in other situations, too?
Where `Duplicate()` is called (I changed the name to something silly and fixed compilation to make sure I haven't missed any):
* `EffectOutputTracks`: IDs must be copied
* `ProjectCloudExtension::OnSyncStarted()` : I think it doesn't matter?
* `PendingTracks`: outdated ;)
* Undo/Redo: at the moment new IDs are given, and it seems to work (I don't know why)
* In `WaveTrack::ClearAndPaste`: new IDs are given, which is correct
* `Au3Interaction::duplicateTracks`: new IDs are given, correct.

So for now at least it would seem that we're good?

Yes, in the future a situation may arise where we need to duplicate with equal IDs. I'm not sure how to change the `Duplicate` API, though, as this is a `Track` API, and other track types may not have clips (e.g. label tracks).

kryksyh: > Undo/Redo: at the moment new IDs are given, and it seems to work (I don't know why)

Now I wonder why applying an effect breaks the project while undo doesn't. If I understand correctly, undo does the same thing as an effect does:  it creates a new track and replaces the original track, including all of its clips.

kryksyh (approved):

au3/libraries/lib-effects/EffectOutputTracks.cpp
saintmatthieu: All things considered, I'll move this to QA already: having this merged would unblock the other ticket I'm working on.

saintmatthieu: @kryksyh https://github.com/audacity/audacity/pull/7880

PR #7873, opened 2024-12-10
Use deep copy for realtime effect states when cloning a ReatimeEffectList

### LINKED ISSUE #4169: The realtime effect settings of a duplicated track are not independent

embarc-gabriel:
Resolves: #4169 

By cloning a RealtimeEffectList a shallow copy of the effect state is created.
In the current implementation if the user copy and paste a track the effect list state for both tracks will point to the same memory location.

This PR modifies the code to use a deep copy instead.
We should also add the effect settings change to the history to avoid just losing the modification after undoing some other action.

Recommended:

### DIFF
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -495,8 +495,17 @@ namespace
                mEffectState->SetActive(mEnableButton->IsDown());
                if (mProject)
                {
-                  ProjectHistory::Get(*mProject).ModifyState(false);
-                  UndoManager::Get(*mProject).MarkUnsaved();
+                  auto const effectName{GetEffectName(*mEffectState)};
+                  ProjectHistory::Get(*mProject).PushState(
+                     /*! i18n-hint: undo history record
+                     first parameter - realtime effect name
+                     */
+                     XO("Change settings for effect %s").Format(effectName),
+                     /*! i18n-hint: undo history record
+                     first parameter - realtime effect name
+                     */
+                     XO("Change effect %s").Format(effectName),
+                     UndoPush::CONSOLIDATE);
                }
             }
          });
--- a/src/RealtimeEffectPanel.cpp
+++ b/src/RealtimeEffectPanel.cpp
@@ -500,11 +500,11 @@ namespace
                      /*! i18n-hint: undo history record
                      first parameter - realtime effect name
                      */
-                     XO("Change settings for effect %s").Format(effectName),
+                     XO("Change settings for realtime effect %s on %s").Format(effectName, mDelegate->GetSourceName()),
                      /*! i18n-hint: undo history record
                      first parameter - realtime effect name
                      */
-                     XO("Change effect %s").Format(effectName),
+                     XO("Change realtime effect %s on %s").Format(effectName, mDelegate->GetSourceName()),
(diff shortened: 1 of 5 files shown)


src/RealtimeEffectPanel.cpp
```diff
@@ -495,8 +495,17 @@ namespace
...
                if (mProject)
                {
-                  ProjectHistory::Get(*mProject).ModifyState(false);
-                  UndoManager::Get(*mProject).MarkUnsaved();
+                  auto const effectName{GetEffectName(*mEffectState)};
+                  ProjectHistory::Get(*mProject).PushState(
+                     /*! i18n-hint: undo history record
+                     first parameter - realtime effect name
+                     */
+                     XO("Change settings for effect %s").Format(effectName),
```
saintmatthieu, line 503: I think we have to add the name of the track it belongs to. If we don't,
1. I tweak my reverb effect on track 1,
2. I tweak my compressor effect on track 2,
3. I undo

would undo my tweaking of the reverb effect of track 1, too.
@LWinterberg please comment.

saintmatthieu, line 503: Should we mention "realtime effect" instead of just "effect", to be consistent with
image ?

LWinterberg: I mean, it should undo the last effect, shouldn't it? I don't understand why the name of the track matters here. Does the code as-is effect all instances of an effect? so if:

1.  I tweak the reverb on track 1
2.  I tweak the reverb on track 2
3.  I tweak a second reverb on track 2
4.  I undo

would all changes across all tracks and effects be undone?

saintmatthieu: Yes.
Why track name matters: this is what the history mechanism bases its logic to consolidate actions. If you change a reverb setting on track one and tell the history "Change reverb param", and then change a reverb setting on track 2 and tell the history "Change reverb param", it'll get consolidated.
So without the track name, it's as you described. With the track name:
* tweak effect on track 1: state is modified (meaning if I close the project, I'm asked if I want to save).
* tweak effect with same name on track 1 again: nothing new happens.
* tweak effect with same name on track 2: new history item is created. *

which I think is better, but please confirm.

saintmatthieu: Another note: a new history item will only get added when I close an effect's UI. Testing this on @embarc-gabriel 's branch, not sure it's the same on master. Seems strange to me. Maybe we should polish this, too, before totally freezing audacity 3.

LWinterberg: sounds good to me

embarc-gabriel: Done.

saintmatthieu (approved):

PR #7891, opened 2024-12-12
7769 generator b&m format fixes

### LINKED ISSUE #7769: Generator Time Length Values are Corrupted
### Bug description

If the User enters values in 'bars:beats' mode for the length of a tone for the tone generator, then later changes the mode of the length (time format), the value will become corrupted. Also, no changes are reflected in bar:beat modes. 

### Steps to reproduce

1. Launch Audacity
2. Create Mono/Stereo Track
3. Open Tone Generator
4. Switch to bar:beat or bar:beat:tick
5. Input any value
6. Observe no values being registered
7. Switch duration mode to any other than 'bar:beat:tick' or 'bar:beat'
8. Observe

### Expected behavior

The User should be able to enter values in bar:beat modes. 

### Actual behavior

The User is not be able to enter values in bar:beat modes. 

https://github.com/user-attachments/assets/9dc2c739-9594-47f7-8e02-223cda01f9af

### Audacity Version

Audacity 4 alpha

### Operating system

Windows

### Additional context

This is reproducible on the original AU4 tone generator PR: https://github.com/audacity/audacity/actions/runs/11720272206

### LINKED ISSUE #7859: Tone generator: duration issues
1. In tone generator, the default duration value is `00h00m00s`
2. The default duration format is `hh:mm:ss` when there's a point selection
3. Digits cannot be entered into the duration text box. Entering from the keyboard simply iterates from one digit to another. DTMF generator has the same problem.

Expected:
1. the default duration value of `00h00m30s`
2. the default duration format `hh:mm:ss+milliseconds` when there's a point selection
3. digits are allowed to be entered

image

saintmatthieu:
Resolves: #7769 
Resolves: #7859

Recommended:

QA:
- [ ] Also test that the equivalent widget in the playback toolbar still works properly.

### DIFF
--- a/src/effects/builtin/common/generatoreffect.h
+++ b/src/effects/builtin/common/generatoreffect.h
@@ -3,17 +3,27 @@
  */
 #pragma once
 
+#include "context/iglobalcontext.h"
+#include "playback/iplayback.h"
+#include "modularity/ioc.h"
+
 struct EffectSettings;
 
 namespace au::effects {
-class GeneratorEffect
+class GeneratorEffect : public muse::Injectable
 {
+    muse::Inject<au::context::IGlobalContext> globalContext;
+    muse::Inject<au::playback::IPlayback> playback;
+
 public:
-    GeneratorEffect(const double& projectRate, const double& t0, double& t1);
+    GeneratorEffect(const double& t0, double& t1);
 
     //! `settings` remain valid for the lifetime of the effect
     void init(EffectSettings* settings);
     double sampleRate() const;
+    double tempo() const;
+    int upperTimeSignature() const;
+    int lowerTimeSignature() const;
     double duration() const;
     void setDuration(double newDuration);
     QString durationFormat() const;
@@ -22,7 +32,6 @@ public:
 private:
     virtual void doInit() {}
 
-    const double& m_projectRate;
     const double& m_t0;
     double& m_t1;
     EffectSettings* m_settings = nullptr;
--- a/src/effects/builtin/common/generatoreffectmodel.h
+++ b/src/effects/builtin/common/generatoreffectmodel.h
@@ -14,18 +14,27 @@ class GeneratorEffectModel : public AbstractEffectModel
     Q_OBJECT
     Q_PROPERTY(double duration READ duration WRITE setDuration NOTIFY durationChanged)
     Q_PROPERTY(QString durationFormat READ durationFormat WRITE setDurationFormat NOTIFY durationFormatChanged)
-    Q_PROPERTY(double sampleRate READ sampleRate NOTIFY sampleRateChanged)
+    Q_PROPERTY(double sampleRate READ sampleRate NOTIFY sampleRateChanged FINAL)
+    Q_PROPERTY(double tempo READ tempo NOTIFY tempoChanged FINAL)
+    Q_PROPERTY(int upperTimeSignature READ upperTimeSignature NOTIFY upperTimeSignatureChanged FINAL)
+    Q_PROPERTY(int lowerTimeSignature READ lowerTimeSignature NOTIFY lowerTimeSignatureChanged FINAL)
 
 public:
 
     double duration() const;
     void setDuration(double newDuration);
     double sampleRate() const;
+    double tempo() const;
+    int upperTimeSignature() const;
+    int lowerTimeSignature() const;
     QString durationFormat() const;
     void setDurationFormat(const QString& newDurationFormat);
 
 signals:
     void sampleRateChanged();
+    void tempoChanged();
+    void upperTimeSignatureChanged();
+    void lowerTimeSignatureChanged();
     void durationChanged();
     void durationFormatChanged();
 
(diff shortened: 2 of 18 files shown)


src/playback/view/toolbars/components/timecodemodel.h
```diff
@@ -16,10 +16,10 @@ class TimecodeModel : public QAbstractListModel
 
     Q_PROPERTY(double value READ value WRITE setValue NOTIFY valueChanged FINAL)
 
-    Q_PROPERTY(double sampleRate READ sampleRate WRITE setSampleRate NOTIFY sampleRateChanged FINAL)
-    Q_PROPERTY(double tempo READ tempo WRITE setTempo NOTIFY tempoChanged FINAL)
-    Q_PROPERTY(int upperTimeSignature READ upperTimeSignature WRITE setUpperTimeSignature NOTIFY upperTimeSignatureChanged FINAL)
-    Q_PROPERTY(int lowerTimeSignature READ lowerTimeSignature WRITE setLowerTimeSignature NOTIFY lowerTimeSignatureChanged FINAL)
+    Q_PROPERTY(double sampleRate READ sampleRate WRITE setSampleRate FINAL)
```
saintmatthieu, line 19: I reckon that these notifications aren't necessary and even confusing, since in fact this class won't change these values from the inside.
I was also wondering: is there a need for the `READ sampleRate` ? I tried without and it seemed to work (but since this seems harmless I left them).

src/playback/qml/Audacity/Playback/components/Timecode.qml
```diff
@@ -15,10 +15,10 @@ RowLayout {
 
     property alias value: timecodeModel.value
 
-    property alias sampleRate: timecodeModel.sampleRate
-    property alias tempo: timecodeModel.tempo
-    property alias upperTimeSignature: timecodeModel.upperTimeSignature
-    property alias lowerTimeSignature: timecodeModel.lowerTimeSignature
+    property real sampleRate
```
Eism, line 18: hmm, is this necessary?

src/effects/builtin/common/generatoreffectmodel.h
```diff
@@ -14,23 +14,33 @@ class GeneratorEffectModel : public AbstractEffectModel
...
+    void tempoChanged();
+    void upperTimeSignatureChanged();
+    void lowerTimeSignatureChanged();
     void durationChanged();
     void durationFormatChanged();
 
 private:
-    void doReload() override;
+    void doReload() final override;
+    virtual void doEmitSignals() = 0;
```
Eism, line 43: Could you explain why we need another method?

src/effects/builtin/common/generatoreffect.h
```diff
@@ -3,17 +3,27 @@
  */
 #pragma once
 
+#include "context/iglobalcontext.h"
+#include "playback/iplayback.h"
+#include "modularity/ioc.h"
```
Eism, line 8: please move it to the top

src/playback/view/toolbars/components/timecodemodel.cpp
```diff
@@ -238,20 +238,9 @@ void TimecodeModel::updateValueString()
     constexpr auto toNearest = true;
     QString newValueString = m_formatter->valueToString(m_value, toNearest).valueString;
 
-    if (newValueString.size() != m_valueString.size()) {
```
Eism, line 241: the component works as expected on the main toolbar? please check it

src/playback/qml/Audacity/Playback/components/Timecode.qml
saintmatthieu: Indeed they are not. Thanks!

src/effects/builtin/common/generatoreffect.h
saintmatthieu: Why ?

src/playback/view/toolbars/components/timecodemodel.cpp
saintmatthieu: It does now. On your latest master, select the `hh:mm:ss + milliseconds` format, and then change to `bar:beat:tick`.

src/effects/builtin/common/generatoreffectmodel.h
Eism: what if we want to add some logic to doReload that is different from sending signals?
there is a high probability that we will add this, then we will revert your change?

src/playback/view/toolbars/components/timecodemodel.cpp
Eism: for qa

src/effects/builtin/common/generatoreffectmodel.h
saintmatthieu: I don't understand your concern. If `GeneratorEffectModel::doReload()` needs to do yet something else, then it does something else, and still call `doEmitSignals()`.

src/effects/builtin/common/generatoreffect.h
saintmatthieu: But certainly, a header file should not rely on other header files being included before itself, right ?
And order of importance: I don't know what that is and I don't want to spend time thinking about it when I include.

Eism: it's more about code cleanliness

Eism (approved):

Issue #7528, opened 2024-10-14
Effects -  Generate Chirp


ianalarina:
### Description

Implement Chirp generator in AU4

PR #7919, opened 2024-12-17
Chirp generator

kryksyh:
Resolves: [#7528](https://github.com/audacity/audacity/issues/7528)

Recommended:

### DIFF
--- a/src/effects/builtin/tonegen/ChirpView.qml
+++ b/src/effects/builtin/tonegen/ChirpView.qml
@@ -1,7 +1,162 @@
-import QtQuick 2.15
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import Muse.UiComponents
+import Audacity.Effects
+import Audacity.Playback
+import "../common"
 
-Rectangle {
+EffectBase {
 
-    color: "#909090"
+    id: root
 
+    property string title: qsTrc("effects", "Chirp")
+    property alias isApplyAllowed: chirp.isApplyAllowed
+
+    width: 370
+    height: 230
+
+    model: chirp
+
+    ToneViewModel {
+        id: chirp
+        instanceId: root.instanceId
+    }
+
+    Component.onCompleted: {
+        chirp.init()
+    }
+
+    GridLayout {
+        columns: 3
+        rows: 5
+        columnSpacing: 4
+        rowSpacing: 16
+
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.fill: parent
+        anchors.margins: 10
+
+        // First row
+        StyledTextLabel {
+            text: qsTrc("effects/chirp", "Waveform:")
+        }
+
+        ComboBox {
+            Layout.columnSpan: 2
+
+            model: chirp.waveforms
+            currentIndex: chirp.waveform
+
+            onActivated: function (index) {
+                chirp.waveform = index
+            }
+        }
+
+        // Second row
+        StyledTextLabel {
+            text: qsTrc("effects/chirp", "Frequency (Hz):")
+        }
+
+        TextInputField {
+            Layout.preferredWidth: 80
+
+            currentText: chirp.frequencyStart
+
+            validator: DoubleValidator {
+                bottom: 1
+            }
+
+            onTextEdited: function (newTextValue) {
+                chirp.frequencyStart = parseFloat(newTextValue)
+            }
+        }
+
+        TextInputField {
+            Layout.preferredWidth: 80
+            currentText: chirp.frequencyEnd
+
+            validator: DoubleValidator {
+                bottom: 1
+            }
+
+            onTextEdited: function (newTextValue) {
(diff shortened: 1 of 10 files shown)


src/effects/builtin/tonegen/ToneView.qml
```diff
@@ -66,15 +60,16 @@ EffectBase {
...
-            width: 80
-            currentText: formatNumber(tone.frequency, 6)
+            Layout.preferredWidth: 80
+
+            currentText: tone.frequencyStart
 
-            validator: DoubleInputValidator {
-                decimal: 6
+            validator: DoubleValidator {
+                bottom: 1
```
saintmatthieu, line 68: Where do you get this `DoubleValidator` from ? What does the `bottom` property determine ?

src/effects/builtin/tonegen/toneeffect.cpp
```diff
@@ -12,6 +12,10 @@ ToneEffect::ToneEffect()
     : ToneGenBase(false), GeneratorEffect(mT0, mT1)
 {}
 
+ToneEffect::ToneEffect(bool isChirp)
```
saintmatthieu, line 15: Why not simplify using just one ctor with bool (or enum) ?

src/effects/builtin/tonegen/ToneView.qml
kryksyh: This is a standard one https://doc.qt.io/qt-6/qml-qtquick-doublevalidator.html, top and bottom define a valid range. The difference with DoubleInputValidator is that the latter adds padding zeroes we don't need.

src/effects/builtin/tonegen/toneeffect.cpp
```diff
@@ -65,24 +71,77 @@ void ToneEffect::setWaveform(Waveform waveform)
...
+
+double ToneEffect::amplitudeEnd() const
+{
+    return mAmplitude1;
+}
+
+void ToneEffect::setAmplitudeEnd(double amplitude)
 {
-    mAmplitude0 = mAmplitude1 = amplitude;
+    mAmplitude1 = amplitude;
```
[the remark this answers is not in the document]
kryksyh: sure, why not

src/effects/builtin/tonegen/toneeffect.cpp
kryksyh: It is instantiated by `BuiltinEffectsModule::Registration` which does not allow us to pass arguments to ctor.

kryksyh: We could use a templated class instead, though not sure if it make the code prettier.

src/effects/builtin/tonegen/chirpeffect.cpp
```diff
@@ -8,7 +8,7 @@ using namespace au::effects;
 const ComponentInterfaceSymbol ChirpEffect::Symbol{ XO("Chirp") };
 
 ChirpEffect::ChirpEffect()
-    : ToneGenBase(true)
+    : ToneEffect(true)
```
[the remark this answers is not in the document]
kryksyh: Absolutely

src/effects/builtin/tonegen/ChirpView.qml
```diff
@@ -1,7 +1,167 @@
...
+    model: chirp
+
+    ToneViewModel {
+        id: chirp
+        instanceId: root.instanceId
+    }
+
+    Component.onCompleted: {
+        chirp.init()
+        timecode.currentFormatStr = chirp.durationFormat
```
saintmatthieu, line 28: I removed this line from the other views, I guess you can here, too.

src/effects/builtin/tonegen/ChirpView.qml
```diff
@@ -1,7 +1,167 @@
...
+            tempo: chirp.tempo
+            upperTimeSignature: chirp.upperTimeSignature
+            lowerTimeSignature: chirp.lowerTimeSignature
+            enabled: true
+
+            onValueChanged: {
+                chirp.duration = timecode.value
+            }
+
+            onCurrentFormatChanged: {
```
saintmatthieu, line 162: You should also be able to remove this.

src/effects/builtin/tonegen/toneviewmodel.cpp
```diff
@@ -39,53 +42,105 @@ QList<QString> ToneViewModel::waveforms() const
...
+    const ToneEffect* const te = effect();
+    if (!te) {
+        return 0.0;
+    }
+    return te->frequencyStart();
+}
+
+void ToneViewModel::prop_setFrequencyStart(double newFrequency)
+{
+    const auto wasAllowed = isApplyAllowed();
```
saintmatthieu, line 113: Shouldn't you add
```cpp
    if (!m_inited) {
        return;
    }
```
here too? (Explanation in [this comment](https://github.com/audacity/audacity/pull/7927#issue-2745281322))

saintmatthieu (approved):

PR #7920, opened 2024-12-17
7843 prevent editing during recording

### LINKED ISSUE #7843: Dissallow the User to interact with clip in Recording
### Bug description

If the User starts to record a new audio clip, the User is allowed to change the position of the clip. This can generate a crash, or loss of audio clips if the recorded audio clip is moved 'onto' those clips. 

https://github.com/user-attachments/assets/b01412d4-d420-42bd-9b40-5d79cac6881b

### Steps to reproduce

1. Launch Audacity
2. Create a Mono or Stereo track
3. Start recording into the track
4. Try to move the recorded clip
5. Observe result

### Expected behavior

The User should not be allowed to change position of any recorded clip. 

### Actual behavior

The User should is allowed to change position of any recorded clip. 

### Audacity Version

Audacity 4 alpha

### Operating system

Windows

### Additional context

_No response_

saintmatthieu:
Resolves: #7843

Recommended:

### DIFF
--- a/src/trackedit/internal/trackeditactionscontroller.cpp
+++ b/src/trackedit/internal/trackeditactionscontroller.cpp
@@ -435,9 +435,9 @@ void TrackeditActionsController::multiClipCopy()
     trackeditInteraction()->clearClipboard();
 
     secs_t offset = 0.0;
-    std::optional<secs_t> mostLeftClipStartTime = trackeditInteraction()->getMostLeftClipStartTime(selectionController()->selectedClips());
-    if (mostLeftClipStartTime.has_value()) {
-        offset = -mostLeftClipStartTime.value();
+    std::optional<secs_t> leftmostClipStartTime = trackeditInteraction()->getLeftmostClipStartTime(selectionController()->selectedClips());
+    if (leftmostClipStartTime.has_value()) {
+        offset = -leftmostClipStartTime.value();
     }
 
     for (const auto& track : tracks) {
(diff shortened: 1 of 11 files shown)


src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -49,6 +49,20 @@ static const ActionCode REDO("redo");
...
+    CUT_CODE,
+    // etc.
+};
+
+namespace {
+bool isDisallowedDuringRecording(const ActionCode& actionCode)
+{
+    return std::find(actionsDisabledDuringRecording.begin(), actionsDisabledDuringRecording.end(),
+                     actionCode) != actionsDisabledDuringRecording.end();
+}
```
igorkorsukov, line 63: for information only
```
#include "global/containers.h"

muse::contains(actionsDisabledDuringRecording, actionCode);
```

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -753,6 +773,8 @@ bool TrackeditActionsController::canReceiveAction(const ActionCode& actionCode)
 {
     if (globalContext()->currentProject() == nullptr) {
         return false;
+    } else if (recordController()->isRecording() && isDisallowedDuringRecording(actionCode)) {
```
igorkorsukov, line 776: just a suggestion, maybe we should add `isRecording` to the global context?  
It might just be a little more convenient, fewer dependencies, for example less mocking and configuring when trying to write a unit test

saintmatthieu: Makes sense.

igorkorsukov (approved):

PR #7932, opened 2024-12-18
7885 selection fixes for effect on clip selection

### LINKED ISSUE #7885: Effect on multiple clips messes up the clip selection
### Bug description

Effect on multiple clips messes up the clip selection

### Steps to reproduce

1. select multiple clips (shift+click)
2. apply an effect (e.g. fade in)

### Expected behavior

Effect is applied and clip selection remains intact.

### Actual behavior

Only one of the clips remains selected.

### Audacity Version

Audacity 4 alpha

### Operating system

Windows, macOS, Linux

### Additional context

`ClipsListModel::onSelectedClips` isn't implemented properly: will in fact only select one clip.
I suspect this to be the case because programmatically selecting several clips at once was never needed yet.
Discovered while developing and testing [effects on multiple clips](https://github.com/audacity/audacity/issues/7801).

saintmatthieu:
Resolves: #7885

Recommended:

QA:
Selection should now be preserved whether the effect is applied on
- [ ] a time selection
- [ ] a single-selected clip
- [ ] multiple selected clips

### DIFF
--- a/src/effects/effects_base/internal/effectexecutionscenario.cpp
+++ b/src/effects/effects_base/internal/effectexecutionscenario.cpp
@@ -83,7 +83,8 @@ muse::Ret EffectExecutionScenario::doPerformEffect(au3::Au3Project& project, con
     secs_t t1;
     bool isSelection = false;
 
-    const auto numSelectedClips = selectionController()->selectedClips().size();
+    const trackedit::ClipKeyList selectedClips = selectionController()->selectedClips();
+    const auto numSelectedClips = selectedClips.size();
 
     {
         //! NOTE Step 1.2 - get effect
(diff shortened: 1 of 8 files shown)


src/effects/effects_base/internal/effectexecutionscenario.cpp
```diff
@@ -243,9 +244,14 @@ muse::Ret EffectExecutionScenario::doPerformEffect(au3::Au3Project& project, con
     //! NOTE Step 6 - perform effect
     //! ============================================================================
     // common things used below
-    const Ret success = numSelectedClips > 1
-                        ? performEffectOnEachSelectedClip(project, *effect, pInstanceEx, *settings)
-                        : effectsProvider()->performEffect(project, effect, pInstanceEx, *settings);
+    Ret success;
+    if (numSelectedClips == 0) {
+        success = effectsProvider()->performEffect(project, effect, pInstanceEx, *settings);
```
igorkorsukov, line 249: I think it would be better to add a function with an appropriate name, or a comment that we are trying to apply the effect to the selected range. 
(otherwise, if not in context, it is difficult to understand what is happening here)

igorkorsukov (approved):

Issue #7899, opened 2024-12-13
Audacity Crash when Effect applied with No Project Open 


chinakov:
### Bug description

If the User has no project open, the User can 'apply' and effect via the navbar menu (Effect -> Built-in). If the User decides to apply any effect, Audacity 4 will crash. 

https://github.com/user-attachments/assets/ec6ae99f-f33c-41eb-ae6f-b49ef9108f19

### Steps to reproduce

1. Launch Audacity
2. Go to 'Effect -> Built-in'
3. Press any effect
4. Observe crash

### Expected behavior

Audacity should not allow the User to apply an effect if no project has been opened. 

### Actual behavior

Audacity allows the User to apply and effect if no project has been opened. 

### Audacity Version

Audacity 4 alpha

### Operating system

Windows

dozzzzer:
Tested

PR #7934, opened 2024-12-18
[effects] make effect open action as uri

igorkorsukov:
Resolves: https://github.com/audacity/audacity/issues/7899

### DIFF
--- a/src/effects/effects_base/internal/effectsuiactions.cpp
+++ b/src/effects/effects_base/internal/effectsuiactions.cpp
@@ -3,7 +3,6 @@
  */
 #include "effectsuiactions.h"
 
-#include "ui/view/iconcodes.h"
 #include "context/uicontext.h"
 #include "context/shortcutcontext.h"
 #include "types/translatablestring.h"
@@ -17,11 +16,7 @@ using namespace muse::actions;
 static const TranslatableString REPEAT_LAST_EFFECT_DEF_TITLE("action", "Repeat last effect");
 static const TranslatableString REPEAT_LAST_EFFECT_TITLE("action", "Repeat %1");
 
-UiActionList EffectsUiActions::m_actions = {
-    UiAction("effect-open",
-             au::context::UiCtxAny,
-             au::context::CTX_ANY
-             ),
+static UiActionList STATIC_ACTIONS = {
     UiAction("repeat-last-effect",
              au::context::UiCtxAny,
              au::context::CTX_ANY,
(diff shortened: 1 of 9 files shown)


src/effects/effects_base/internal/effectsuiactions.cpp
```diff
@@ -95,6 +90,32 @@ void EffectsUiActions::init()
...
+
+    auto makeActions = [this](const EffectMetaList& effects) {
+        m_actions.clear();
+        m_actions.reserve(effects.size() + STATIC_ACTIONS.size());
+
+        for (const EffectMeta& e : effects) {
+            UiAction action;
+            action.code = makeEffectOpenAction(e.id).toString();
+            action.uiCtx = context::UiCtxProjectOpened;
+            action.scCtx = context::CTX_PROJECT_FOCUSED;
```
saintmatthieu, line 102: Why do `uiCtx` and `scCtx` differ ? If `action.uiCtx = context::UiCtxProjectOpened;`, I'd have expected `action.scCtx = context::CTX_PROJECT_OPENED;`

saintmatthieu (approved):

src/effects/effects_base/internal/effectsuiactions.cpp
igorkorsukov: This is a simple and complex topic at the same time :) 

With the Ui context, everything is quite simple - it determines when an action is available depending on which page is open. For example, we might be on the home screen, or we might be in the project screen (with a project loaded)... the current UI context tells us where we are now. 

Shortcut context is a little different. It tells when we can accept the shortcut of this action.
The whole problem is that the number of shortcuts, key combinations, is very limited. Therefore, a situation arises when one and the same shortcut must trigger different actions, depending on where exactly the user is.

Let's assume that we have some panel on the project page, for example a mixer or something else. And there are some actions... and we can assign a shortcut to them... and this shortcut will match the shortcut assigned to the effect, what should we do with this?
Options:
1. Do not allow assigning the same shortcut to different actions - but there are too few keyboard shortcuts..
2. We can determinate where exactly the user is now and, accordingly, what exactly he expects when this shortcut is triggered. If the user clicks on clips and triggers a shortcut to which the effect is assigned - we need to show the form for working with the effect. If the user clicks on the mixer and triggers the same shortcut, then we need to call the action associated with the mixer (for example)

At the same time, the actions themselves can be available regardless of which panel the user is currently on (they can simply be sent in some other way, for example by clicking on the UI control)

saintmatthieu: Ok, I stupidly imagined `uiCtx` and `scCtx` were duplication of a same concept. Thanks for the clarification, will be useful soon enough :D

PR #7943, opened 2024-12-19
Wait to save thumbnail before close the project

### LINKED ISSUE #7874: Project thumbnail isn't generated upon closing the project and saving
### Steps to reproduce

1. Open a project
2. Make some changes to it (e.g. trim a clip)
3. Ctrl + W
4. Choose "Save"
5. Review the thumbnail of the project

### Expected behavior

The thumbnail is updated and reflects the current state of the project.

### Actual behavior

The thumbnail isn't updated

### Audacity Version

Audacity 4 alpha

### Operating system

Windows, macOS, Linux

### Additional context

_No response_

embarc-gabriel:
Resolves: #7874

A callback is called once the thumbnail is ready to be saved on the disk.
Once this actions takes some time, it was being executed after navigation was performed and the project closed.

The current solution implement the thumbnailCreator class to address this issue.

Recommended:


src/project/iaudacityproject.h
```diff
@@ -39,7 +39,8 @@ class IAudacityProject
     virtual void setNeedAutoSave(bool val) = 0;
     virtual muse::async::Notification needSaveChanged() const { return muse::async::Notification(); }
     virtual muse::Ret save(const muse::io::path_t& path = muse::io::path_t(), SaveMode saveMode = SaveMode::Save) = 0;
-    virtual muse::async::Notification captureThumbnailRequested() const = 0;
+    virtual muse::async::Channel<std::string> captureThumbnailRequested() const = 0;
+    virtual void onThumbnailCreated(bool success) = 0;
```
igorkorsukov, line 43: I suggest removing these methods from the project. In ProjectPropertiesModel, use the creator directly.
There is such logic here, the public interface for creating a thumb is the IThumbnailCreator. The ProjectPropertiesModel model is its implementation...
In the future, we can change the implementation, for example, to create Thumbnail, use not the current Qml scene grab, but independent drawing or something like that.

src/project/internal/audacityproject.cpp
```diff
@@ -266,20 +272,28 @@ Ret Audacity4Project::doSave(const muse::io::path_t& savePath, bool generateBack
...
+    if (!ret) {
+        return make_ret(Ret::Code::UnknownError);
     }
-    return muse::make_ret(muse::Ret::Code::UnknownError);
+
+    if (createThumbnail) {
+        std::string thumbnailPath = savePath.toStdString();
+        thumbnailPath.replace(thumbnailPath.size() - 4, 4, "png");
+
+        Ret ret = thumbnailCreator()->createThumbnail(thumbnailPath);
```
igorkorsukov, line 289: There are two things that bother me here: 
1. What if a path without an extension is passed here, or with some other extension, something like `file.au.backup`... I think it's better to use `muse::io::FileInfo` for path parsing 
2. the fact that the thumbnail is saved in PNG format is something only the creator “knows”, i.e. we either reveal the knowledge here, or we tell the creator to save it in PNG, and not in any other way.

src/project/internal/thumbnailcreator.cpp
```diff
@@ -0,0 +1,29 @@
+#include "thumbnailcreator.h"
+
+using namespace au::project;
+
+void ThumbnailCreator::onThumbnailCreated(bool success)
+{
+    m_thumbnailCreated.send(success);
+}
+
+muse::async::Channel<std::string> ThumbnailCreator::captureThumbnailRequested() const
```
igorkorsukov, line 10: better to use `muse::io::path_t` to indicate what is file path

src/project/internal/thumbnailcreator.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    return m_createThumbnailRequested;
+}
+
+muse::Ret ThumbnailCreator::createThumbnail(const std::string& path)
+{
+    muse::Ret ret;
+    QEventLoop loop;
+    m_thumbnailCreated.resetOnReceive(this);
```
igorkorsukov, line 19: I think it's better to move it after `loop.exec()`.
There may be a potential problem here, in `onReceive` we pass a lambda in which we save a reference to a local  variable (loop). If someone for some reason calls `onThumbnailCreated`, and `onReceive` is triggered, then there may be a crash. Therefore, it is safer for us to simply delete this callback after it is triggered.

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -29,12 +29,13 @@ Rectangle {
     ProjectPropertiesModel {
         id: project
 
-        onCaptureThumbnail: function captureThumbnail() {
+        onCaptureThumbnail: function captureThumbnail(thumbnailUrl) {
             // hide playCursor for the time grabbing image
             playCursor.visible = false
             content.grabToImage(function(result) {
                 playCursor.visible = true
-                result.saveToFile(project.thumbnailUrl)
```
igorkorsukov, line 37: better to remove `thumbnailUrl` property from ProjectPropertiesModel

embarc-gabriel: Done.

src/project/internal/thumbnailcreator.cpp
embarc-gabriel: Good catch. Done.

src/project/internal/thumbnailcreator.cpp
embarc-gabriel: Done.

src/project/internal/audacityproject.cpp
embarc-gabriel: Ok. I have created a helper method and it is way better now.

src/project/iaudacityproject.h
embarc-gabriel: Yes! I have refactor it.

src/project/internal/thumbnailcreator.h
```diff
@@ -0,0 +1,25 @@
...
+class ThumbnailCreator : public IThumbnailCreator, public muse::async::Asyncable
+{
+public:
+    ThumbnailCreator() = default;
+    void onThumbnailCreated(bool success) override;
+    muse::Ret createThumbnail(const muse::io::path_t& path) override;
+    muse::async::Channel<muse::io::path_t> captureThumbnailRequested() const override;
+
+private:
+    muse::io::path_t thumbnailPath(const muse::io::path_t& path);
```
igorkorsukov, line 19: it can be marked as `const` method

src/project/ithumbnailcreator.h
```diff
@@ -0,0 +1,21 @@
...
+
+class IThumbnailCreator : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(au::project::ithumbnailcreator)
+
+public:
+    virtual ~IThumbnailCreator() = default;
+    virtual void onThumbnailCreated(bool success) = 0;
+    virtual muse::Ret createThumbnail(const muse::io::path_t& path) = 0;
+    virtual muse::async::Channel<muse::io::path_t> captureThumbnailRequested() const = 0;
```
igorkorsukov, line 18: here are the details of the implementation revealed... I think it's not worth complicating it and solving it, I just suggest adding a comment and regrouping it somehow like this 

```
    virtual muse::Ret createThumbnail(const muse::io::path_t& path) = 0;

    // for internal use 
    virtual muse::async::Channel<muse::io::path_t> captureThumbnailRequested() const = 0;
    virtual void onThumbnailCreated(bool success) = 0;
```

src/project/projectmodule.cpp
```diff
@@ -80,6 +82,7 @@ void ProjectModule::registerExports()
     ioc()->registerExport<IRecentFilesController>(moduleName(), m_recentFilesController);
     ioc()->registerExport<IOpenSaveProjectScenario>(moduleName(), new OpenSaveProjectScenario());
     ioc()->registerExport<IProjectFilesController>(moduleName(), m_actionsController);
+    ioc()->registerExport<IThumbnailCreator>(moduleName(), m_thumbnailCreator);
```
igorkorsukov, line 85: just for information, if the object is not used in other methods (for example, we don't need to do init in onInit), then we don't have to create a class member (this is not a problem))
```
ioc()->registerExport<IThumbnailCreator>(moduleName(), new ThumbnailCreator());
```

src/project/ithumbnailcreator.h
embarc-gabriel: Done.

src/project/internal/thumbnailcreator.h
embarc-gabriel: Sure! I don´t know why I forgot this. Thanks.

src/project/ithumbnailcreator.h
igorkorsukov: mistake in grouping :)

embarc-gabriel: Done!

embarc-gabriel: Sorry for that.

igorkorsukov (approved):

Issue #7901, opened 2024-12-13
Clicking in the empty space on the right-hand side of the project will cause it to move forward


dozzzzer:
![2024-12-13_14-57-26 (1)](https://github.com/user-attachments/assets/f959ebfd-5606-41e7-bd3e-c1c738f3a699)

grliszas14:
QA: this introduces more changes than fixing issue mentioned in the description:
* improved auto-scroll over the left/right edge of the app. It is applied to clip moving (single and multi-clip), trimming (both handles) and selection
* improved selection itself, if you try (in master build) to select a range close to the view edge, the frame moves once (on mouse btn release) and selection moves along with it not respecting previously selected start time. It shouldn't happen anymore

dozzzzer:
@grliszas14 Windows: During recording, the clip doesn't appear to move, but the timeline of the project is scrolled, causing complete havoc 😛

https://github.com/user-attachments/assets/4f89ee31-49dd-4b3e-a37c-7f5b6d61a518

chinakov:
Tested:
- User is now disallowed to scroll the project during the recording if a clip has been clicked'n'dragged. (@dozzzzer reported issue)
- The auto-scroll via dragging or moving clips is now better
- Improved selection

PR #7950, opened 2024-12-19
Horizontal scrolling and selection improvements

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7901

### DIFF
--- a/src/projectscene/view/clipsview/clipslistmodel.h
+++ b/src/projectscene/view/clipsview/clipslistmodel.h
@@ -113,8 +113,6 @@ private:
     int indexByKey(const trackedit::ClipKey& k) const;
     QVariant neighbor(const ClipKey& key, int offset) const;
 
-    double autoScrollView(double newTime);
-
     Qt::KeyboardModifiers keyboardModifiers() const;
 
     TimelineContext* m_context = nullptr;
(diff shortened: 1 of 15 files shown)


src/projectscene/view/clipsview/clipslistmodel.h
```diff
@@ -127,5 +125,8 @@ private slots:
     //! Offset between mouse click position on clip's header and clip's start and end time
     double m_clipEditStartTimeOffset = -1.0;
     double m_clipEditEndTimeOffset = -1.0;
+
+    bool m_receiveAutoScrollNotifications = false;
+    QMetaObject::Connection m_autoScrollConnection;
```
igorkorsukov, line 130: can we remove `m_receiveAutoScrollNotifications` and replace it with connection check?
```
if (m_autoScrollConnection) {
...
```

src/projectscene/view/timeline/timelinecontext.cpp
```diff
@@ -11,14 +11,41 @@
...
+}
+
 TimelineContext::TimelineContext(QObject* parent)
     : QObject(parent)
 {
     m_snapTimeFormatter = std::make_shared<SnapTimeFormatter>();
+
+    m_scrollTimer = new QTimer();
+    m_scrollTimer->setInterval(16); // scroll at ~60 FPS
+    connect(m_scrollTimer, &QTimer::timeout, [this](){ autoScrollView(m_autoScrollStep); });
```
igorkorsukov, line 48: I think it's better to use the timer by value (not as a pointer), or you need to add its deletion and initialization to null.

igorkorsukov (approved):

Issue #7948, opened 2024-12-19
If an entered value is out of range or if a value is cleared, Apply button should be disabled


dozzzzer:
- If a text box is cleared, the Apply button should be disabled

<img width="596" alt="image" src="https://github.com/user-attachments/assets/745cd614-ca22-43f3-b4b0-d596f4f59557" />

- If an entered value is out of range, the Apply button should be disabled

<img width="596" alt="image" src="https://github.com/user-attachments/assets/e8e4a4e7-67ab-44d4-a823-91ef9a93aed3" />

_Originally posted by @dozzzzer in https://github.com/audacity/audacity/issues/7528#issuecomment-2551387704_

dozzzzer:
Tested

PR #7954, opened 2024-12-20
Fix apply allowed conditions in generators

kryksyh:
Resolves: [#7948](https://github.com/audacity/audacity/issues/7948)

Recommended:


src/effects/builtin/dtmfgen/dtmfgenerator.h
```diff
@@ -10,6 +10,10 @@ struct DtmfSettings
     static constexpr char DefaultSequence[] = "audacity";
     static constexpr double DefaultDutyCycle = 55.0;
     static constexpr double DefaultAmplitude = 0.8;
+    static constexpr double AmplitudeMin = 0.0;
```
saintmatthieu, line 13: In DtmfBase.h, `Amplitude` has min 0.001. Not sure what's better, please check with @LWinterberg.

saintmatthieu (approved):

Issue #7672, opened 2024-11-07
Noise reduction missing "Residue" option


LWinterberg:
![image](https://github.com/user-attachments/assets/9573c8e6-84bd-497a-95c8-ffca53de8033)

Here should be both "Reduce" (default) and Residue. 

Originally reported on https://forum.audacityteam.org/t/residue-option-missing-in-noise-reduction-effect/122116/

Tekwizz:
So I suppose I'd need to use 3.6.3 to still have the feature?

NanoExplorer:
I think the problem is as follows:
in 1304871 a lot of logic was moved from src/effects/NoiseReduction.cpp to libraries/lib-builtin-effects/NoiseReductionBase.cpp but the one line that stuck out to me was https://github.com/audacity/audacity/blob/1304871db447dca1777c7f63a52602ec828c96e6/libraries/lib-builtin-effects/NoiseReductionBase.cpp#L54
Now that define is absent from src/effects/NoiseReduction.cpp even though it is used in both files! for example:
https://github.com/audacity/audacity/blob/1304871db447dca1777c7f63a52602ec828c96e6/src/effects/NoiseReduction.cpp#L174-L176

NanoExplorer:
I made an example commit here which fixes the issue: 353a72830ac50d4692acf7e2744d3c8f60b2831f, but I don't think it meets the contributor guidelines since the commit message doesn't describe the purpose and effects, instead it just describes the changes I made.

dozzzzer:
Tested

PR #7969, opened 2024-12-28
Fix bug #7672 by moving define statement to header so that subclasses…

NanoExplorer:
Resolves: #7672

Move the define statement for RESIDUE_CHOICE (and related comments) to the header file NoiseReductionBase.h so that it gets included into both NoiseReductionBase.cpp and src/effects/NoiseReduction.cpp

Recommended:


kryksyh (approved):
Thanks @NanoExplorer !

PR #7972, opened 2024-12-30
Avoid text change due binding while editing

### LINKED ISSUE #7925: Deleting a fractional number in an effect's text box by pressing Backspace doesn't work correctly
1. Put the cursor at the very end
2. Start pressing Backspace and observe how the number is changing

embarc-gabriel:
Resolves: #7925

Remove the binding while user is editing the text.
This avoids a text change to trigger a call to toFixed and a change in the text itself due the binding.

Recommended:

### DIFF
--- a/src/effects/builtin/amplify/AmplifyView.qml
+++ b/src/effects/builtin/amplify/AmplifyView.qml
@@ -45,11 +45,19 @@ EffectBase {
                 text: qsTrc("effects/amplify", "Amplification (dB):")
             }
 
+            Binding {
+                target: ampInput
+                property: "currentText"
+                value: +amplify.amp.toFixed(4)
+                when: !ampInput.focus
+                restoreMode: Binding.RestoreNone
+            }
+
             TextInputField {
+                id: ampInput
+
                 anchors.verticalCenter: parent.verticalCenter
                 width: 80
-
-                currentText: amplify.amp.toFixed(4)
 
                 validator: DoubleInputValidator {
                     top: amplify.ampMax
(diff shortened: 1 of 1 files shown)


src/effects/builtin/amplify/AmplifyView.qml
```diff
@@ -45,12 +45,20 @@ EffectBase {
                 text: qsTrc("effects/amplify", "Amplification (dB):")
             }
 
+            Binding {
```
grliszas14, line 48: I'd add some comment explaining why we need this magic - I'm sure I will not remember that the next time I look at this code :)

grliszas14 (approved):

src/effects/builtin/amplify/AmplifyView.qml
embarc-gabriel: Done! Thank you!

galih1992 (approved):

kryksyh:
@embarc-gabriel do you think that we can use a solution similar to that pr: https://github.com/audacity/audacity/pull/7954/files

I believe that it would be a more straightforward solution. I.e.:

```qml
            TextInputField {
                anchors.verticalCenter: parent.verticalCenter
                width: 80

                currentText: amplify.amp.toFixed(4)

                validator: DoubleValidator {
                    top: amplify.ampMax
                    bottom: amplify.ampMin
                }

                inputField.onDisplayTextChanged: function () {
                     amplify.amplitude = parseFloat(inputField.displayText)
                }
            }
```

igorkorsukov (approved):

PR #7976, opened 2025-01-02
Use clang-format instead of uncrustify

kryksyh:
Use clang-format instead of uncrustify

Recommended:


.clang-format
```diff
@@ -0,0 +1,245 @@
+---
+BasedOnStyle: Google
+AccessModifierOffset: -4
+AlignAfterOpenBracket: AlwaysBreak
```
grliszas14, line 4: I’d add 
`AlignAfterOpenBracket: BlockIndent`
To move closing parenthesis to the new line, matching indentation with the first line

grliszas14 (approved):

kryksyh:
Will reopen if it would be needed

PR #7978, opened 2025-01-02
Distinguish timepoint and duration on beats formatter

### LINKED ISSUE #7928: Generators: a zero duration in a non-musical format translates into `01bar01beat`
1. Open up a generator
2. Enter 00h00m00s into the duration text box
3. Switch the duration format to "bar:beat:tick"
4. Observe

Act.: `01bar01beat01tick`
Exp.: `00bar00beat00tick`

embarc-gabriel:
Resolves: #7928 

While converting from some numerical format to beats format we should distinguish a time point and a duration.
The string 00:00:00 should be converted either to 01 bar 01 beat 01 tick if we consider it a point in time or 00 bar 00 beat 00 tick if we consider it the difference between two points.

Recommended:


src/playback/tests/timecodemodel_tests.cpp
```diff
@@ -82,10 +82,10 @@ TEST_F(TimecodeModelTests, Parse_Beats)
     m_model->setLowerTimeSignature(4);
 
     std::vector < std::pair<double, QString> > expectedValues = {
-        { 0.0, "001 bar 01 beat" },
-        { 0.6, "001 bar 02 beat" },
-        { 1.0, "001 bar 03 beat" },
-        { 1.6, "002 bar 01 beat" },
+        { 0.0, "000 bar 00 beat" },
```
grliszas14, line 85: I think it would be nice to leave this test with expected values as before (but setting TimePoint mode explicitly) and writing additional test checking Duration mode

embarc-gabriel: Yes! Sure! You are right! I should increase the test coverage, not the opposite.

embarc-gabriel: As far as I could see. we only use the Duration variant on this TimeCode widget and I don´t want to inject the formatter mode if this is not really needed.
But to keep the test coverage I have added a file to test the BeatsFormatter itself.

saintmatthieu: Hey @embarc-gabriel !
> As far as I could see. we only use the Duration variant on this TimeCode widget

Is it not the opposite? Isn't the instance in `PlaybackToolBar.qml` responsible for the display of this guy?

image

src/playback/view/toolbars/components/beatsformatter.h
```diff
@@ -5,11 +5,16 @@
...
+    TimePoint,  // 1-base indexing 00:00:00s -> 1 bar 1 beat 1
+    Duration    // 0-base indexing 00:00:00s -> 0 bar 0 beat 0
+};
+
 namespace au::playback {
 class BeatsFormatter : public TimecodeFormatter
 {
 public:
-    BeatsFormatter(const QString& formatStr, int fracPart);
+    BeatsFormatter(const QString& formatStr, int fracPart, BeatsFormatterMode mode = BeatsFormatterMode::TimePoint);
```
saintmatthieu, line 19: Personally I wouldn't make such a critical argument optional with default, to make sure the user of this class gives it some thinking.

src/playback/view/toolbars/components/beatsformatter.h
```diff
@@ -27,7 +32,7 @@ class BeatsFormatter : public TimecodeFormatter
     void updateFields(size_t barsDigits);
 
     std::array<double, 3> m_fieldLengths;
-    int m_fieldValueOffset = 1;
     int m_fracPart = 0;
+    BeatsFormatterMode m_mode;
```
saintmatthieu, line 38: Cannot be `const` ?

saintmatthieu: Actually, why not keeping `m_fieldValueOffset`, (making it const and) setting it in the ctor to 0 or 1 depending on the `BeatsFormatterMode` ? That would reduce the diff and make the code more condensed.

grliszas14:
I took a look at the build, in this scenario widget should say 1 bar 1 beat 1 tick I believe
timecode

saintmatthieu:
https://github.com/audacity/audacity/pull/7978#issuecomment-2569028352

Yes, I think the `TimecodeModel` class must allow the `BeatsFormatterMode` to be settable, probably as a Q_PROPERTY. Then the playback-toolbar instance should set it to time-point, and the instances in the different generators should set it to duration.

Then I guess @embarc-gabriel in the unit test, you could probably keep the existing case (which assumes time-point) after setting time-point, and add your new one after having set it to duration.

src/playback/tests/timecodemodel_tests.cpp
embarc-gabriel: Yes! I tottaly missed these ones.
I focused on the ones we use on the effeects.
I don´t know how I haven´t seen it,
The modification will be a little bit tricker than I thought because I will need to inject this information depending on whether we are on the bar or on the effect.
I will work on this.

src/playback/view/toolbars/components/beatsformatter.h
embarc-gabriel: I don´t know.
I always think it is better to be explicit.
An int can be anything. We can put 2, 3, 10 or whatever. That way we create many valid inputs that in the end shouldn´t exist.
Using the enum we make these options impossible to be represented.

embarc-gabriel:
> [#7978 (comment)](https://github.com/audacity/audacity/pull/7978#issuecomment-2569028352)
> 
> Yes, I think the `TimecodeModel` class must allow the `BeatsFormatterMode` to be settable, probably as a Q_PROPERTY. Then the playback-toolbar instance should set it to time-point, and the instances in the different generators should set it to duration.
> 
> Then I guess @embarc-gabriel in the unit test, you could probably keep the existing case (which assumes time-point) after setting time-point, and add your new one after having set it to duration.

Yes! I will rework this.
Thx.

src/playback/playbackmodule.cpp
```diff
@@ -67,6 +68,7 @@ void PlaybackModule::registerResources()
 
 void PlaybackModule::registerUiTypes()
 {
+    qmlRegisterUncreatableType<TimecodeModeSelector>("Audacity.Playback", 1, 0, "TimecodeModeSelector", "TimecodeModeSelector is an simple enum");
```
saintmatthieu, line 71: "a simple enum"

src/playback/tests/timecodemodel_tests.cpp
```diff
@@ -6,113 +6,100 @@
...
+    m_model.setValue(param.inputValue);
+    EXPECT_EQ(param.expectedOutput, m_model.valueString());
+}
 
+INSTANTIATE_TEST_SUITE_P(
+    AllTimecodeModelTests,
+    TimecodeModelParameterizedTests,
+    ::testing::Values(
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::HHMMSS, TimecodeMode::TimePoint, 0.0, "00 h 00 m 00 s" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::HHMMSS, TimecodeMode::TimePoint, 0.6, "00 h 00 m 01 s" },
```
saintmatthieu, line 48: Interesting. Shouldn't the display stay at 0 until the value has reached 1 ?

src/playback/tests/timecodemodel_tests.cpp
```diff
@@ -6,113 +6,100 @@
...
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::HHMMSS, TimecodeMode::TimePoint, 61.0, "00 h 01 m 01 s" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::HHMMSS, TimecodeMode::TimePoint, 3600.0, "01 h 00 m 00 s" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::HHMMSS, TimecodeMode::TimePoint, 3601.0, "01 h 00 m 01 s" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeat, TimecodeMode::TimePoint, 0.0, "001 bar 01 beat" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeat, TimecodeMode::TimePoint, 0.6, "001 bar 02 beat" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeat, TimecodeMode::TimePoint, 1.0, "001 bar 03 beat" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeat, TimecodeMode::TimePoint, 1.6, "002 bar 01 beat" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeat, TimecodeMode::Duration, 0.0, "000 bar 00 beat" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeat, TimecodeMode::Duration, 0.6, "000 bar 01 beat" },
```
saintmatthieu, line 60: Here it's like I'd expect it.

src/playback/tests/timecodemodel_tests.cpp
```diff
@@ -6,113 +6,100 @@
...
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::TimePoint, 0.0, "001 bar 01 beat 01" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::TimePoint, 0.6, "001 bar 02 beat 01" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::TimePoint, 1.0, "001 bar 03 beat 01" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::TimePoint, 1.9, "002 bar 01 beat 04" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::Duration, -1.0, "--- bar -- beat --" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::Duration, 0.0, "000 bar 00 beat 00" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::Duration, 0.6, "000 bar 01 beat 00" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::Duration, 1.0, "000 bar 02 beat 00" },
+        TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::Duration, 1.9, "001 bar 00 beat 03" }
```
saintmatthieu, line 72: We're at 120bpm, so a beat is 0.5 seconds, so in 1.9 seconds, there are 3 beats (a complete 3/4 bar, that's ok) and the remained is 0.4 seconds.
One tick is a fourth of a beat, so 0.125 seconds. 0.4/0.125 = 3.2. So I think it's correct.
But it's hard to figure out. Maybe encoding `double beatDur = 60/tempo; double tickDur = beatDur / 4;` and then 
```cpp
TimecodeModelTestParam{ TimecodeModel::ViewFormatType::BarBeatTick, TimecodeMode::Duration, 3*beatDur + 3.2*tickDur, "001 bar 00 beat 03" }
```
would be helpful for the reader ?

saintmatthieu (approved):
LGTM to be moved forward to QA, but I'd like the unit tests (which look great aside from this) made easier for the reader.

src/playback/tests/timecodemodel_tests.cpp
embarc-gabriel: Here I have just kept the same tests as before once I was not focused on the NumericFormatter.
I don´t know exactly the reason for this but I do know this is very explicit on the code that we need to round it to the nearest integer.

`void TimecodeModel::updateValueString()
{
    constexpr auto toNearest = true;`

src/playback/tests/timecodemodel_tests.cpp
embarc-gabriel: Yes! Very good point.
I have improved the readability for these tests.
I hope it is better now.

src/playback/tests/timecodemodel_tests.cpp
saintmatthieu: ... and it indeed is that way in Audacity 3, too:
image
@LWinterberg is this correct ?

saintmatthieu: I do find this rounding all the more confusing that, for musical durations, neither of bar, beat or tick is rounded:
image

saintmatthieu: ... but if no one has noticed so far it's also probably that it's not dramatic. Let's move on with this PR and create a follow-up ticket if there's a consensus that it should be changed.

saintmatthieu (approved):

src/playback/view/toolbars/components/timecodemodel.cpp
```diff
@@ -122,6 +122,29 @@ QHash<int, QByteArray> TimecodeModel::roleNames() const
...
+
+void TimecodeModel::setMode(TimecodeMode mode)
+{
+    if (m_mode == mode) {
+        return;
+    }
+
+    m_mode = mode;
+
+    reloadFormatter();
```
grliszas14, line 143: I'd move those two:
`reloadFormatter();
updateValueString();`
outside of this function - as its name suggests, it should only setMode and nothing else.
I think it will be better to make a connection within the constructor to respond to that change, sth like:
`connect(this, &TimecodeModel::valueChanged, [](){ reloadFormatter(); updateValueString();}`

src/projectscene/qml/Audacity/ProjectScene/common/InsertSilence.qml
```diff
@@ -42,6 +42,7 @@ StyledDialogView {
             Layout.fillHeight: false
 
             value: silence.duration
+            mode: TimecodeMode.Duration
```
grliszas14, line 45: I think it's registered as 
`TimecodeModeSelector` not the `TimecodeMode` right? If it works anyway because of typedef or sth it would be nice to keep it consistent in all qml files

src/effects/builtin/noisegen/NoiseView.qml
```diff
@@ -69,6 +69,7 @@ EffectBase {
             Layout.fillHeight: false
 
             value: noise.duration
+            mode: TimecodeModeSelectors.Duration
```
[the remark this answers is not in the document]
embarc-gabriel: Thx. Nice catch.

src/projectscene/qml/Audacity/ProjectScene/common/InsertSilence.qml
embarc-gabriel: Done.

src/playback/view/toolbars/components/timecodemodel.cpp
embarc-gabriel: I would say I agree to your point here.
But if you take a look on the code, this pattern is used everywhere on this file. (setTempo, setSampleRate, setUpperSignature and so on). I have used this pattern to keep the consistency, once I think it is important as you correctly stated on the other comment.
So IMO there are two options: we refactor everything on this file or keep it as it is. Or maybe leave as it is an open other ticket to refactor it IDK.
What do you think @grliszas14 ?

grliszas14: aaah, indeed it is, let's leave it as is then

grliszas14 (approved):

src/playback/tests/timecodemodel_tests.cpp
embarc-gabriel: > I do find this rounding all the more confusing that, for musical durations, neither of bar, beat or tick is rounded: image

I totally agree with you. The behavior varies depending on the formatter used.

Issue #7809, opened 2024-11-27
AU3: Allow modules to stick when minor version mismatches


LWinterberg:
This is generally only true during 3.7.x as we know there'll only be minor bugfixes for this series.

kryksyh:
@LWinterberg could you elaborate more on what this is about?

LWinterberg:
at the moment, there needs to be a new build of third party modules due to ModuleManager.cpp:114. This makes it particularly annoying for openVINO users. We should be able to make it easier on them by allowing patches (3.7.x) to just continue using the module.

kryksyh:
@dozzzzer please check if the OpenVINO module built for 3.7.1 works with the 3.7.2 build

dozzzzer:
@kryksyh It doesn't, Openvino plugins aren't available in this build.

PR #7982, opened 2025-01-03
Treat modules with the same maj.min version as compatible

kryksyh:
Resolves: https://github.com/audacity/audacity/issues/7809

Recommended:

### DIFF
--- a/libraries/lib-module-manager/ModuleManager.cpp
+++ b/libraries/lib-module-manager/ModuleManager.cpp
@@ -11,7 +11,7 @@
 *******************************************************************//*!
 
 \file ModuleManager.cpp
-\brief Based on LoadLadspa, this code loads pluggable Audacity 
+\brief Based on LoadLadspa, this code loads pluggable Audacity
 extension modules.  It also has the code to
 invoke a function returning a replacement window,
 i.e. an alternative to the usual interface, for Audacity.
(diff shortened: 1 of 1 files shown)


LWinterberg (approved):

libraries/lib-module-manager/ModuleManager.cpp
```diff
@@ -67,6 +68,21 @@ static BasicUI::MessageBoxResult DoMessageBox(const TranslatableString &msg)
...
 }
 
+// Module's Major.Minor version should match the current Audacity build
+static bool IsVersionCompatible(const wxString &moduleVersion)
+{
+   wxArrayString parts1 = wxStringTokenize(AUDACITY_VERSION_STRING, ".");
+   wxArrayString parts2 = wxStringTokenize(moduleVersion, ".");
+
+   if (parts1.size() < 2 || parts2.size() < 2) {
+      wxLogError("Invalid version format. Audacity version: %s, module version: %s", AUDACITY_VERSION_STRING, moduleVersion);
```
saintmatthieu, line 78: How about `assert(false)` ? Wouldn't that be a grave bug that needs attention?

saintmatthieu (approved):

libraries/lib-module-manager/ModuleManager.cpp
kryksyh: Agree

Issue #7737, opened 2024-11-18
Clip groups


DilsonsPickles:
### Description

Allow Users to create groups of clips. Required to remove sync-lock feature

[Figma link](https://www.figma.com/design/01XLIICmy2AG6t1ieyTp66/02---Clips?node-id=2420-2)

ianalarina:
hey @DilsonsPickles can we push it up in your priorities list? this is a next step for Grzegorz

dozzzzer:
@grliszas14 

- When clips are grouped, the "Group clips" option in the Edit and the clip context menu should be greyed since it shouldn't be allowed to group the already grouped clips

<img width="635" alt="image" src="https://github.com/user-attachments/assets/cb535b3b-7ed5-40b1-9314-cf979b69e13d" />

dozzzzer:
@grliszas14 

- Right-click in the clip's header doesn't cause the context menu to appear anymore

<img width="768" alt="image" src="https://github.com/user-attachments/assets/70dc6181-7f54-4633-8f00-01f4adda637a" />

dozzzzer:
Tested

PR #7985, opened 2025-01-03
Clip groups

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7737


src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1655,6 +1655,77 @@ bool Au3Interaction::toggleStretchToMatchProjectTempo(const ClipKey& clipKey)
...
+int64_t Au3Interaction::determineGroupId(const ClipKeyList &clipKeyList) const
+{
+    //! NOTE: check if any clip already belongs to a group
+    for (const auto& selectedClip : clipKeyList) {
+        if (clipGroupId(selectedClip) != -1) {
+            return clipGroupId(selectedClip);
+        }
+    }
+
+    //! NOTE: none of the clips is grouped, find unique id for them
```
kryksyh, line 1697: I would use the same approach as with track and clip ids, with a simple static counter, but your solution is fine too.

src/trackedit/internal/trackeditinteraction.h
```diff
@@ -71,6 +71,11 @@ class TrackeditInteraction : public ITrackeditInteraction, public muse::Injectab
 
     bool toggleStretchToMatchProjectTempo(const ClipKey& clipKey) override;
 
+    int64_t clipGroupId(const trackedit::ClipKey& clipKey) const override;
+    void setClipGroupId(const trackedit::ClipKey& clipKey, int64_t id) override;
+    int64_t determineGroupId(const ClipKeyList& clipKeyList) const override;
+    ClipKeyList clipsGroup(int64_t id) const override;
```
kryksyh, line 77: Can we name it `clipsInGroup`? Current name makes me think that it should return a group id or something.

src/trackedit/internal/au3/au3trackeditproject.cpp
```diff
@@ -50,6 +50,32 @@ Au3TrackeditProject::~Au3TrackeditProject()
...
 }
 
+std::vector<int64_t> Au3TrackeditProject::groupsIdsList() const
+{
+    std::vector<int64_t> groupsList;
+
+    for (const auto& trackId : trackIdList()) {
+        Au3WaveTrack* waveTrack = DomAccessor::findWaveTrack(*m_impl->prj, Au3TrackId(trackId));
+        IF_ASSERT_FAILED(waveTrack) {
+            return std::vector<int64_t>();
```
kryksyh, line 60: this is totally a matter of taste, but why not use just `{}` instead of `std::vector<int64_t>()` ?

au3/libraries/lib-wave-track/WaveClip.h
```diff
@@ -1049,6 +1052,8 @@ class WAVE_TRACK_API WaveClip final :
    bool mIsPlaceholder { false };
 
    wxString mName;
+
+   int64_t mGroupId = -1;
```
kryksyh, line 1056: I guess it is also a matter of taste, but I suggest using `0` as "No group / Default state" value and unsigned type. Conditions would look cleaner, and default initialized values would have a "correct" value.

kryksyh (approved):

src/trackedit/internal/au3/au3interaction.cpp
grliszas14: The reasoning behind this is that group IDs are saved in the project file, grouping and ungrouping several times would increment those numbers every time project is opened and some grouping operations are done - this way we can re-use IDs that are no longer in use.
It's unlikely that somebody will reach max_int and we set again the same ID but you never know, better safe than sorry :)

kryksyh: Yeah, that's what I thought. On the other hand, if we are using a 64-bit counter, it will overflow after 584942417 years of incrementing every millisecond. 😃

src/trackedit/internal/au3/au3trackeditproject.cpp
grliszas14: changed as well 👍

au3/libraries/lib-wave-track/WaveClip.h
grliszas14: I have no strong opinion on that, I can change that to whatever we like, what do others think about it?
@igorkorsukov @Eism @saintmatthieu @embarc-gabriel

embarc-gabriel: I will give my two cents. Probably it does not worth it if we have to change too many things.
I usually think it is better to use a std::option<uint64_t> because the intention is clearer. We have an id or none.

Eism: for me -1 is good, the same as for `mId` field

PR #7988, opened 2025-01-06
7938 summonable realtime effect dialogs

### LINKED ISSUE #7938: Make real-time effect dialogs summonable
### Description

At the moment, one can add real-time effects but cannot open a dialog to adjust their parameters.

saintmatthieu:
Resolves: #7938

Clicking on a realtime-effect's name in the effect panel now opens the UI.

Recommended:

QA:
Note that:
* power button not yet functional
* the dialogs are yet modal, will implement them as floating window in a follow-up PR.

In particular, please test
- [x] Realtime effect parameters can be changed during playback with audible effect.
- [x] Regression testing of the effects, both realtime and destructive (some refactoring was necessary)

### DIFF
--- a/src/effects/builtin/common/abstracteffectmodel.cpp
+++ b/src/effects/builtin/common/abstracteffectmodel.cpp
@@ -46,7 +46,7 @@ std::shared_ptr<au::effects::EffectInstance> AbstractEffectModel::instance() con
     return instancesRegister()->instanceById(id);
 }
 
-EffectSettings* AbstractEffectModel::settings() const
+const EffectSettings* AbstractEffectModel::settings() const
 {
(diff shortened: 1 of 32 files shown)


kryksyh (approved):

chinakov:
Tested:
- Ability of the User to change 'Reverb' realtime effect during playback via user input or via presets.
- Regression test of all built-in real time and destructive effects. 

Couldn't be tested:
- **Repair** 'destructive effect', as the sample view and the sample draw feature is still missing, disallowing for proper testing of this particular effect.
- Ability of the User to stop playback during the modifying of a real time effect, as this will be implemented in the coming task. 

Thank you!

Issue #7700, opened 2024-11-12
Deletion of a selection of audio data along with the empty space should be non-destructive


dozzzzer:
### Problem to be solved

When making a selection over the clip's audio data and space next to it, and then deleting it, the remaining clip cannot be expanded using the trim handles:

https://github.com/user-attachments/assets/288fe184-de63-45b5-9871-eb598d3721c9

### Your idea

Deleting a clip's data along with the blank space should allow the resulting clip to be restored to its original length; in other words, such an operation should be non-destructive, preserving the clip's data.

kryksyh:
Reverting this in 3.7.3 due to user complaints until the feature is more thoroughly thought out.

PR #8000, opened 2025-01-08
Clear the track in a non desctrutive way

embarc-gabriel:
Resolves: #7700

This changes clear the track in a non desctrutive way.
We can restore the content expading the track for delete and cut operation.
Apart from that,  if we cut and past the track in another position we can also recovery the missing data.

Recommended:


embarc-gabriel:
@dozzzzer this will allow to recover the track for delete and cut operations. I think this is expected, right?
But it will also allow you to recover the data after cut and paste a track. Is this an reasonable side effect of this change?

saintmatthieu (approved):

PR #8010, opened 2025-01-10
Fix importing AUP projects with custom named project data directory

Giermann:
Importing older AUP projects will fail, when the project data directory is not the default name "<projectname>_data". While reading the tag "project", the attribute "projname" is correctly read and stored in mProjDir, but when reading the data files this variable was not used.

Recommended:


kryksyh (approved):

kryksyh:
@Giermann thanks!

PR #8025, opened 2025-01-13
7980 closing while recording

### LINKED ISSUE #7980: Closing a project during recording causes the corruption of that project
1. Open a project
2. Click Record button
3. Ctrl (Cmd) + W
4. Try to open the same project again
5. Observe: the project cannot be opened anymore

https://github.com/user-attachments/assets/1b479cd1-1db8-48f0-8dae-6c2ab7c222f7

saintmatthieu:
Resolves: #7980

Actions leading to a loss of recording should be disabled during recording.

Recommended:

QA:
- [ ] Actions that can lead to losing an ongoing recording are disabled until the recording is finished. (Please raise those which still aren't.)
- [ ] On a saved project: recording, stopping recording and closing without saving does not corrupt the project.

### DIFF
--- a/src/appshell/internal/applicationactioncontroller.cpp
+++ b/src/appshell/internal/applicationactioncontroller.cpp
@@ -63,6 +63,16 @@ void ApplicationActionController::init()
     dispatcher()->reg(this, "revert-factory", this, &ApplicationActionController::revertToFactorySettings);
 }
 
+const std::vector<muse::actions::ActionCode>& ApplicationActionController::prohibitedActionsWhileRecording() const
+{
+    static const std::vector<ActionCode> PROHIBITED_WHILE_RECORDING {
+        "quit",
+        "restart",
+    };
+
+    return PROHIBITED_WHILE_RECORDING;
+}
+
 void ApplicationActionController::onDragEnterEvent(QDragEnterEvent* event)
 {
     onDragMoveEvent(event);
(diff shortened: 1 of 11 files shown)


src/project/internal/audacityproject.cpp
```diff
@@ -97,9 +97,12 @@ void Audacity4Project::close()
     history->undoUnsaved();
     history->clearUnsaved();
     // Do not save a project that has never explicitly been saved.
-    //if (m_au3Project->hasSavedVersion()) {
-    //    save();
-    //}
+    if (m_au3Project->hasSavedVersion()) {
```
saintmatthieu, line 100: @Eism FYI this was a mistakenly committed change - commenting this code back in shouldn't cause a regression (but solves a nasty one).

src/appshell/internal/applicationactioncontroller.cpp
```diff
@@ -63,6 +63,16 @@ void ApplicationActionController::init()
     dispatcher()->reg(this, "revert-factory", this, &ApplicationActionController::revertToFactorySettings);
 }
 
+const std::vector<muse::actions::ActionCode>& ApplicationActionController::prohibitedActionsWhileRecording() const
+{
+    static const std::vector<ActionCode> PROHIBITED_WHILE_RECORDING {
```
Eism, line 68: You can declare this vector at the top of this file (above all methods) without wrapping it in a method

src/project/internal/audacityproject.cpp
Eism: one thing I don't understand is that we first cancel all the changes (undo and clear the history) and then try to save

src/appshell/internal/applicationactioncontroller.cpp
saintmatthieu: This method is used in `applicationuiactions.cpp`

src/project/internal/audacityproject.cpp
saintmatthieu: Why rewinding and explicitly saving is needed although the changes since last save haven't been saved, I cannot answer on top of my head. Database / auto-save management is a complex part of Audacity 3, I each time have to study the thing again from scratch before touching it.

Eism (approved):

PR #8051, opened 2025-01-15
Duplicate clip and multi-clip

### LINKED ISSUE #8002: Duplicate Shortcut does not include Audio Clips
### Bug description

If the User duplicates any given track via the shortcut (Ctrl+D), the audio clips that are located within that track won't be duplicated alongside the track. 

If the User uses the duplicate button in track context menu, the behaviour remains correct. 

### Steps to reproduce

1. Launch Audacity
2. Create a new track, generate some audio into the track
3. Select the Track by clicking on the track header
4. Press 'Ctrl+D' (Duplicate)
5. Observe Result

### Expected behavior

The track should be duplicated with the audio clips intact if the User uses the shortcut. 

### Actual behavior

Only the track is copied if the shortcut is used. 

https://github.com/user-attachments/assets/ac11cdbe-af8b-4527-b632-83e9fbc9802c

### Audacity Version

Audacity 4 alpha

### Operating system

Windows

### Additional context

_No response_

embarc-gabriel:
Resolves: #8002 

Implement multi-clip duplication.

Ps: I have also removed a warning on a function missing a return value.

- [] There are no behavior changes unnecessary for the stated purpose of the PR

Recommended:

### DIFF
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -1247,27 +1247,39 @@ bool Au3Interaction::duplicateClip(const ClipKey& clipKey)
 
 bool Au3Interaction::duplicateClips(const ClipKeyList& clipKeyList)
 {
-    //Get ordered list of trackIds so we can duplicate in the same order
-    std::set<TrackId> trackIds;
-    std::transform(clipKeyList.begin(), clipKeyList.end(), std::inserter(trackIds, trackIds.begin()), [](const ClipKey& clipKey) {
-        return clipKey.trackId;
-    });
-
-    if (trackIds.empty()) {
+    trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+    if (!prj) {
         return false;
     }
 
-    std::vector<Au3WaveTrack*> waveTracks;
-    std::transform(trackIds.begin(), trackIds.end(), std::back_inserter(waveTracks), [this](TrackId trackId) {
-        return DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
+    //Get list of tracks in order by the UI position
+    trackedit::TrackIdList tracks = prj->trackIdList();
+
+    //Get a set of the tracks we want copy content from
+    std::set<TrackId> selectedTracks;
+    std::transform(clipKeyList.begin(), clipKeyList.end(), std::inserter(selectedTracks, selectedTracks.begin()), [](const ClipKey& clipKey) {
+        return clipKey.trackId;
     });
+    if (selectedTracks.empty()) {
+        return false;
+    }
+
+    //Get only the selected tracks but keeping the UI order
+    std::vector<Au3WaveTrack*> waveTracks;
+    for (auto trackId : tracks) {
+        if (selectedTracks.find(trackId) != selectedTracks.end()) {
+            auto waveTrack = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
+            if (waveTrack) {
+                waveTracks.push_back(waveTrack);
+            }
+        }
+    }
 
     auto& projectTracks = Au3TrackList::Get(projectRef());
-    trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+    auto& trackFactory = WaveTrackFactory::Get(projectRef());
+    auto& pSampleBlockFactory = trackFactory.GetSampleBlockFactory();
 
     for (const auto& track : waveTracks) {
-        auto& trackFactory = WaveTrackFactory::Get(projectRef());
-        auto& pSampleBlockFactory = trackFactory.GetSampleBlockFactory();
         auto newTrack = track->EmptyCopy(pSampleBlockFactory);
 
         std::vector<ClipKey> clipsToDuplicate;
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -1242,22 +1242,50 @@ bool Au3Interaction::duplicateSelectedOnTracks(const TrackIdList& tracksIds, sec
 
 bool Au3Interaction::duplicateClip(const ClipKey& clipKey)
 {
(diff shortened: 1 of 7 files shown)


src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -344,24 +344,18 @@ void TrackeditActionsController::redo()
...
-        if (selectionController()->selectedClips().empty()) {
-            return;
-        }
-        if (selectionController()->selectedClips().size() == 1) {
-            ClipKey selectedClipKey = selectionController()->selectedClips().at(0);
-            if (!selectedClipKey.isValid()) {
-                return;
-            }
-            dispatcher()->dispatch(DUPLICATE_CLIP, ActionData::make_arg1<ClipKey>(selectedClipKey));
+        const auto selectedClips = selectionController()->selectedClips();
```
grliszas14, line 350: You can have a case where user selects track1 and track2 via clicking on the track header on the left panel (no clips are selected in that case and no range selection is made as well): we'd like to copy both tracks with all of their clips in that case, that's not covered here
Am I right @LWinterberg @chinakov ?

embarc-gabriel: @grliszas14 I have updated the code to address this situation.

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -135,7 +135,7 @@ void TrackeditActionsController::init()
     dispatcher()->reg(this, UNDO, this, &TrackeditActionsController::undo);
     dispatcher()->reg(this, REDO, this, &TrackeditActionsController::redo);
     dispatcher()->reg(this, DUPLICATE_SELECTED, this, &TrackeditActionsController::duplicateSelected);
```
grliszas14, line 137: I know that this action was named like this before but I think it's a good time to rename it, I started renaming with copy/cut/delete actions, would be nice to follow this convention:
```c++
static const ActionCode CLIP_COPY_CODE("clip-copy");
static const ActionCode MULTI_CLIP_COPY_CODE("multi-clip-copy");
static const ActionCode RANGE_SELECTION_COPY_CODE("clip-copy-selected");
``` 
that way it will be much clearer what it does

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1242,22 +1242,50 @@ bool Au3Interaction::duplicateSelectedOnTracks(const TrackIdList& tracksIds, sec
...
+        return DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
+    });
+
+    auto& projectTracks = Au3TrackList::Get(projectRef());
+    trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
 
-    bool ok = duplicateSelectedOnTrack(clipKey.trackId, begin, end);
-    if (!ok) {
-        return false;
+    for (const auto& track : waveTracks) {
```
grliszas14, line 1268: not likely but `track` can be nullptr (DomAccessor::findWaveTrack can return one), we should add sth like

```c++
if (!track) {
    continue;
}
```

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1242,22 +1242,50 @@ bool Au3Interaction::duplicateSelectedOnTracks(const TrackIdList& tracksIds, sec
...
+    return duplicateClips({ clipKey });
+}
 
-    std::shared_ptr<Au3WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.clipId);
-    IF_ASSERT_FAILED(clip) {
+bool Au3Interaction::duplicateClips(const ClipKeyList& clipKeyList)
+{
+    //Get ordered list of trackIds so we can duplicate in the same order
+    std::set<TrackId> trackIds;
+    std::transform(clipKeyList.begin(), clipKeyList.end(), std::inserter(trackIds, trackIds.begin()), [](const ClipKey& clipKey) {
```
grliszas14, line 1252: This code will stop working once you try to change order of tracks in the UI.
Tracks order in UI may look like: (0,1,5,3,4,2), if user selects track 5 and 3, this code will sort it again resulting in duplicating tracks in the wrong order (3, 5) instead of (5, 3).

This one should give you trackIds properly ordered, you just need to filter not needed tracks from it
```c++
    trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
    if (!prj) {
        return {};
    }
    
    trackedit::TrackIdList tracks = prj->trackIdList();
```

embarc-gabriel: Huummm ... good catch! I really haven't thought about this possibility. Thanks.

src/trackedit/internal/trackeditactionscontroller.cpp
embarc-gabriel: Done. In fact, I have modified only the name for the actions that were somehow used in this PR. Is that ok?

src/trackedit/internal/au3/au3interaction.cpp
embarc-gabriel: Fixed.

src/trackedit/internal/au3/au3interaction.cpp
embarc-gabriel: Fixed. I have tested it as well, and it seems fine.

src/trackedit/internal/trackeditactionscontroller.cpp
embarc-gabriel: I will mark it as resolved but feel free  to not agree with the solution.

grliszas14: Oh, I made myself not clear :)
I meant to do sth like:

```c++
instead of this:
static const ActionCode DUPLICATE_SELECTED_CODE("duplicate-selected"); <--- it's ambiguous, we don't know what is selected: clips? tracks? range selection?

we can make this:
static const ActionCode DUPLICATE_RANGE_SELECTION_CODE("duplicate-range-selection");
```

embarc-gabriel: hahahah sorry for not understanding the idea.
This is fixed now.

grliszas14 (approved):

Issue #8039, opened 2025-01-14
Set minimum project length


LWinterberg:
### Description

For the purposes of zooming, the project length is 2x the content length. This is slightly weird for short projects, so we should set a sensible minimum length

chinakov:
Tested: 
- The max zoom-out length is set to 4 minutes.
- If the User drags clips to warrant more space within the project, upon deleting the clip the project length reverts back to 4 minutes.

PR #8053, opened 2025-01-15
Set minimum timeline zoom range to 4 minutes

LWinterberg:
Resolves: #8039


src/projectscene/view/timeline/timelinecontext.cpp
```diff
@@ -562,7 +562,12 @@ void TimelineContext::setZoom(double zoom, double mouseX)
     double timeRange = m_frameEndTime - m_frameStartTime;
     double mouseTime = m_frameStartTime + (mouseX / m_frameWidth) * timeRange;
 
-    double totalTimeRange = trackEditProject()->totalTime() * 2.0;
+    //we limit zoom to the total time range *2, but at least 4 minutes
+    double totalTimeRange=4*60.0;
```
grliszas14, line 566: you can do it in one line like this I think:

double totalTimeRange = std::max(trackEditProject()->totalTime() * 2.0, 4 * 60.0);

grliszas14 (approved):

PR #8061, opened 2025-01-17
Fix paste on start of clip

### LINKED ISSUE #7410: Audio pasted in wrong time position when pasting to the start of the same clip
### Bug description

When copying part of a clip and pasting to the beginning of the same clip, the audio is incorrectly pasted at its original position with the clip appended to the end, rather than being pasted at the start position of the clip.

### Steps to reproduce

1. Generate a 10 second Chirp
2. Copy from 4 to 6 seconds
3. Press the rewind button or "J" to move the cursor to the start of the clip
4. Paste

### Expected behavior

The copied audio to be pasted at time = 0, extending to time = 2, with the full clip appended to the end.

### Actual behavior

The audio is pasted at time = 4, extending to time = 6, with the full clip appended to the end.

### Audacity Version

### Operating system

Windows, Linux

### Additional context

Originally reported here: https://forum.audacityteam.org/t/strange-behavior-when-inserting-audacity-3-6-4/118816/1

embarc-gabriel:
Resolves: #7410 

The paste operation on the begging of the track can be handle as a paste on the middle but with t0 = 0.

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -1139,11 +1139,13 @@ bool WaveClip::Paste(double t0, const WaveClip& o)
    ClearSequenceFinisher finisher;
 
    //seems like edge cases cannot happen, see WaveTrack::PasteWaveTrack
+   double leftTrimDueToPaste = 0.0;
    auto &factory = GetFactory();
    if (t0 == GetPlayStartTime())
    {
       finisher = ClearSequence(GetSequenceStartTime(), t0);
       SetTrimLeft(other.GetTrimLeft());
+      leftTrimDueToPaste = other.GetTrimLeft();
 
       auto copy = std::make_shared<WaveClip>(other, factory, true);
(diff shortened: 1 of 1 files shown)


libraries/lib-wave-track/WaveClip.cpp
```diff
@@ -1140,17 +1140,7 @@ bool WaveClip::Paste(double t0, const WaveClip& o)
 
    //seems like edge cases cannot happen, see WaveTrack::PasteWaveTrack
    auto &factory = GetFactory();
-   if (t0 == GetPlayStartTime())
```
saintmatthieu, line 1143: Like I said in private, I expect this change not to allow untrimming after pasting as shown in [this video](https://github.com/user-attachments/assets/72a597ac-82b9-45bf-939a-f0a56a1196bb). Please try the patch I gave you.

embarc-gabriel: The patch worked flawlessly.

saintmatthieu (approved):

Issue #3922, opened 2022-11-12
Help menu to have an entry to link to GitBook Audacity Support


petersampsonaudacity:
### Your idea

Add a command to the Audacity Help menu to directly access the Audacity Support GitBook:

![image](https://user-images.githubusercontent.com/50205984/201473852-466e8b26-7048-4ce8-9ddf-6afb8543a1a1.png)

### Problem to be solved

The app's **Help** menu has links to the manual - but **_no_** link to the Audacity Support Gitbook.

Given that it is planned for the GitBook to eventually supersede the Manual, especially for tutorials and how-to type instructions it would seem to make sense to have a link to the GitBook direct from the app.

Already the FFmpeg installation instructions have migrated from the Manual to the GitBook - see:
 FFmpeg Download link from Libraries prefs should link to the GitBook page #3920 

This would improve the visibility of the Audacity Support GitBook.

### Prior art

Not applicable

### Additional context

Flagging @Tantacrul, @DilsonsPickles  and @LWinterberg as a user interface/design issue

petersampsonaudacity:
BTW, while we are looking at this menu I would suggest that the commands in this menu do not really need the ellipsis "**...**"

petersampsonaudacity:
In a similar vein, it would make eminent sense to add such a link to the Welcome to Audacity doormat dialog as an aid to visibility and accessibility of the GitBook:
![image](https://user-images.githubusercontent.com/50205984/201482499-1c5112ca-0fd9-43c0-9db1-ed48cf2da53e.png)

DavidBailes:
Please can this get fixed.

petersampsonaudacity:
> ...  the manual, which is no longer up to date.

Actually the Manual is pretty much up to date - I have been seeing to that.  

The one key thing that is not in the manual is the Effects Stack and real-time effects, for that the Manual has a link (or links) to the relevant page in the GitBook (I saw no sense in replicating that).

Today or tomorrow I need to update it for the new cut/copy/paste toolbar which is now in the 3.2.3 Beta (as of late yesterday).

But yes I agree with @DavidBailes  that it would be good to get this fixed, ideally for 3.2.3, - it certainly looks like @LWinterberg at least has a fix for the launch welcome-mat.

DavidBailes:
> Actually the Manual is pretty much up to date - I have been seeing to that.

But quite a few examples where it isn't. See, for example, https://alphamanual.audacityteam.org/man/Effect_Menu:_VST

[...]

PR #8068, opened 2025-01-18
Add audacity support gitbook link to help menu

nalin243:
Resolves: *#3922*

Recommended:


petersampsonaudacity:
Tested on W10 with @nalin243 's test branch build.

This works fine:
image

And clicking that link properly takes you to the Muse support site,

libraries/lib-wx-init/HelpSystem.cpp
```diff
@@ -47,6 +47,8 @@ const wxString HelpSystem::HelpHostname = wxT("manual.audacityteam.org");
 const wxString HelpSystem::HelpServerHomeDir = wxT("/");
 const wxString HelpSystem::HelpServerManDir = wxT("/man/");
 
+const wxString HelpSystem::AudacitySupportLink = wxT("gitbook.com/open/library/audacity/audacity");
```
LWinterberg, line 50: this here should be support.audacityteam.org - the link you used is gitbooks' overview site :)

PR #8071, opened 2025-01-20
Fix freeze during the project download

kryksyh:
When the download thread was scheduled before shared_ptr to RemoteProjectSnapshot was created, lambda failed to acquire a weak_ptr to self, resulting in the request not being processed.

Now dispatch threads after the shared_ptr is initialized.

Recommended:

### DIFF
--- a/libraries/lib-cloud-audiocom/sync/RemoteProjectSnapshot.h
+++ b/libraries/lib-cloud-audiocom/sync/RemoteProjectSnapshot.h
@@ -114,6 +114,7 @@ private:
 
    void ReportProgress();
 
+   void StartSync();
    bool InProgress() const;
    void RequestsThread();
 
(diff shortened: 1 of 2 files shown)


embarc-gabriel (approved):

libraries/lib-cloud-audiocom/sync/RemoteProjectSnapshot.h
```diff
@@ -71,6 +71,8 @@ class RemoteProjectSnapshot final
       ProjectInfo projectInfo, SnapshotInfo snapshotInfo, std::string path,
       RemoteProjectSnapshotStateCallback callback, bool downloadDetached);
 
+   void StartSync();
```
saintmatthieu, line 74: This extra method for documentation purpose or is there another reason?

saintmatthieu (approved):

libraries/lib-cloud-audiocom/sync/RemoteProjectSnapshot.h
kryksyh: I added it to make the code more understandable to indicate that the actual sync starts within mRequestsThread. 

I should made `StartSync` private, though; fixed that.

dozzzzer:
Tested. No freezes are observed when syncing is being resumed.

PR #8076, opened 2025-01-21
Fix paste at the start of moved clips with left trim

### LINKED ISSUE #7410: Audio pasted in wrong time position when pasting to the start of the same clip

embarc-gabriel:
Resolves: #7410

This should fix all erros presented on the ticket while pasting at the start of a clip.
This mainly includes left trimmed clips and moved clips.

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveClip.cpp
+++ b/libraries/lib-wave-track/WaveClip.cpp
@@ -1139,13 +1139,16 @@ bool WaveClip::Paste(double t0, const WaveClip& o)
    ClearSequenceFinisher finisher;
 
    //seems like edge cases cannot happen, see WaveTrack::PasteWaveTrack
-   double leftTrimDueToPaste = 0.0;
+   double pastePositionShift = 0.0;
+   double startPositionShift = 0.0;
    auto &factory = GetFactory();
    if (t0 == GetPlayStartTime())
    {
+      pastePositionShift = t0;
+      startPositionShift = other.GetSequenceStartTime() + GetTrimLeft();
+
       finisher = ClearSequence(GetSequenceStartTime(), t0);
       SetTrimLeft(other.GetTrimLeft());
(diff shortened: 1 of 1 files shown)


saintmatthieu (approved):
Just by looking at the code I haven't managed in the last 15mn to tell if this is right or not.

`startPositionShift` influences the offset, and so long this is successfully tested for different start times / trim durations for both source and destination clips, we should be able to get some confidence that the change is correct. Similarly for `pastePositionShift`.

At least one thing is evident: this change (and that of the previous commit) won't change behaviour for anything but pasting at the start of a clip, so it's safe enough that QA concentrates on pasting at the beginning.

The main reason why this is so hard to understand is how much convoluted the logic is (not from your changes, although they add to this complexity), e.g. with `finisher` being given some boundary values before this clip data is changed as input to an operation to be done after the clip data is changed.

If this gets us into more trouble, I suggest that we make a dedicated function for each of these three cases: paste at the beginning, paste in the middle, paste at the end, and nevermind if there's a bit of code we can't reuse and must duplicate.

PR #8096, opened 2025-01-26
Update appimagetool

Drsheppard01:
an update to the appimagetool version that will revert to building only one appimage for all systems (probably except musl-based, since ld-linux.so is still required)


LWinterberg:
Did you mean to target this to master? In Master, Audacity 4 development happens, the au3/ folder is merely kept as a place for legacy code. If you want to see this in the next release, please rebase and target release-3.7.2

kryksyh:
@Drsheppard01 thanks for the heads-up. 
Your changes were merged as part of #8137.

Issue #8103, opened 2025-01-27
What's new dialog: can't press the "continue for cloud storage" button using the keyboard.


DavidBailes:
### Bug description

You can't tab to either:
1. The control containing the text "sign up to Audacity's...". This only affects the users of screen readers.
2. The "Continue for cloud storage" button. This affects all keyboard users.

### Steps to reproduce

Note that because of the separate issue that there is no visual indication of focus for any of the buttons in the dialog, then it's recommended to run either the Narrator screen reader, or the accessibility insights for Windows app for these steps to reproduce.
1. Arrange for the What's new/Welcome to Audacity to open.
2. Try to Tab to the "Continue for cloud storage" button.

### Expected behavior

You can Tab to the button

### Actual behavior

You can't tab to the button.

### Audacity Version

Audacity 3.7.2

### Operating system

Windows

PR #8106, opened 2025-01-27
What's new window: fix controls not being in tab order.

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/8103

Problem: The "text" control and the button related to cloud storage are not in the tab order.

Fix: Change InfoWindow to inherit from wxPanel rather than wxWindow, so that the focus of its children is automatically taken care of.

Recommended:

### DIFF
--- a/src/WhatsNewDialog.cpp
+++ b/src/WhatsNewDialog.cpp
@@ -214,11 +214,11 @@ const wxSize infoWindowSize(746, 176);
 #else
 const wxSize infoWindowSize(1050, 250);
 #endif
-class InfoWindow : public wxWindow
+class InfoWindow : public wxPanel
 {
 public:
     InfoWindow(wxWindow* parent)
-        : wxWindow(parent, wxID_ANY, wxDefaultPosition, infoWindowSize)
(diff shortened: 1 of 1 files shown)


kryksyh (approved):

kryksyh:
@DavidBailes, currently, the user can't select any button with the keyboard on MacOS. Do you know if this is a known issue?

DavidBailes:
> @DavidBailes, currently, the user can't select any button in WhatsNew dialog with the keyboard on MacOS. Do you know if this is a known issue?

As far as I'm aware it's not a known issue, but there are a number of keyboard issues when using MacOS - mainly due to wxWidget bugs. I don't have a Mac myself, so  can't immediately test, but my guess it that Audacity currently isn't very usable on MacOS using a keyboard.

A couple of general points. For keyboard navigation on MacOS, then Full keyboard access should be on (see: https://support.apple.com/en-gb/guide/mac-help/mchlc06d1059/mac), and sometimes it's necessary to press Ctrl+Tab rather that just Tab to move the focus.

PR #8108, opened 2025-01-27
Master effects

### LINKED ISSUE #7957: Master effects
### Description

Add master effect ([design](https://www.figma.com/design/w2kAh1gKKV5bu7htY7158W/08---Effects?node-id=2001-1024&t=wdnc9sSxLKFIdHra-0))

This ticket will complete https://github.com/audacity/audacity/issues/7341

#### IMPORTANT
Commit c1197ac18e ("Master-effect stack actions are undo- and redoable") was still not cherry-picked.

saintmatthieu:
Resolves: #7957

Recommended:

### QA:

Please note that a color inconsistency with the design specification (normal state of the power buttons being brighter in the implementation) was deliberately left as such, for the reason given in [this comment](https://github.com/audacity/audacity/pull/8108#discussion_r1931698409).

- [x] Undo/Redo
- [x] Power buttons
- [x] Effect panel toggling works as before
- [x] "E" shortcut also closes the panel (previously could only open it)
- [x] Add / Remove / Replace effects on track and master effects works

### DIFF
--- a/src/effects/effects_base/qml/Audacity/Effects/BypassEffectButton.qml
+++ b/src/effects/effects_base/qml/Audacity/Effects/BypassEffectButton.qml
@@ -1,0 +1,16 @@
+import QtQuick
+
+import Muse.Ui
+import Muse.UiComponents
+
+FlatButton {
+    id: root
+
+    property bool isMasterEffect: false
+
+    height: width
+    icon: IconCode.BYPASS
+    iconColor: accentButton ? "white" : ui.theme.fontPrimaryColor
+    iconFont: ui.theme.toolbarIconsFont
+    accentColor: isMasterEffect ? "black" : ui.theme.accentColor
+}
--- a/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectList.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectList.qml
@@ -13,9 +13,10 @@ import Audacity.ProjectScene
 Rectangle {
     id: root
     implicitHeight: trackEffectList.contentHeight
+    readonly property int topMargin: 8
+    property alias isMasterTrack: trackEffectListModel.isMasterTrack
     property alias trackName: trackEffectListModel.trackName
     property alias trackEffectsActive: trackEffectListModel.trackEffectsActive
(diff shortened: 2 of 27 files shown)


src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectList.qml
```diff
@@ -13,9 +13,10 @@ import Audacity.ProjectScene
 Rectangle {
```
kryksyh, line 13: Could we add `policy: ScrollBar.AlwaysOn` to `id: scrollbar` item? Without it gets really confusing about how many effects I have in stack:

image

vs

image

kryksyh (approved):

src/effects/effects_base/qml/Audacity/Effects/BypassEffectButton.qml
```diff
@@ -0,0 +1,16 @@
...
+FlatButton {
+    id: root
+
+    property bool isMasterEffect: false
+
+    height: width
+    icon: IconCode.BYPASS
+    iconColor: accentButton ? "white" : ui.theme.fontPrimaryColor
+    iconFont: ui.theme.toolbarIconsFont
+    accentColor: isMasterEffect ? "black" : ui.theme.accentColor
```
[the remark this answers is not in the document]
saintmatthieu: Thanks for finding out where the discrepancy between the specs and my implementation lies 👍 

> Would it make sense to introduce audacity-specific color definitions?

This would mean allowing the two apps to diverge more in terms of look-and-feel, it should be discussed with the designers.

In particular for buttons, there are three opacities:
* normal state: fairly opaque,
* hover state: feebly opaque,
* pressed state: completely opaque.

I guess it was on purpose that pressed and normal states differ, but the spec for this task, indeed, would requires equal opacity for both. I would like to raise this during the next demo meeting.

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectList.qml
saintmatthieu: I agree with you. I'll make a push for this (if designers aren't happy I'll change it back later).

Issue #8018, opened 2025-01-10
Clicking Skip to start/Skip to end should cancel the selection


dozzzzer:
### Steps to reproduce

1. Have a clip/region of a clip selected
2. Click "Skip to start" or "skip to end"
3. Observe

### Expected behavior

Previously made selection disappears

### Actual behavior

Previously made selection remains

https://github.com/user-attachments/assets/e4966c71-b83a-4c78-8b7a-2b750dc57ba5

### Audacity Version

Audacity 4 alpha

### Operating system

Windows, macOS, Linux

PR #8116, opened 2025-01-28
Rewind cancels the selection

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8018

### DIFF
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -261,10 +261,7 @@ void PlaybackController::rewindToStart()
 
     seek(0.0);
 
-    IF_ASSERT_FAILED(player()) {
-        return;
-    }
-    player()->rewind();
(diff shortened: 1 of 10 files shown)


src/playback/internal/playbackcontroller.cpp
```diff
@@ -273,6 +273,7 @@ void PlaybackController::rewindToEnd()
     stop();
 
     seek(totalPlayTime());
+    player()->playbackRewound().notify();
```
Eism, line 276: Ah, please don't use this method. I have already pointed out [here](https://github.com/audacity/audacity/pull/7482#discussion_r1832687169) that this method should not be in the player

Can we call resetDataSelection from Au3SelectionController directly here?

grliszas14: fixed 👍

Eism (approved):

Issue #8043, opened 2025-01-14
Implement "Get Effects" button


ianalarina:
### Description

Implement the ability to get effects from MuseHub within Audacity

[Figma link](https://www.figma.com/design/4Yr6wm2gF3YYPSUYlj7IaS/38_WX---MuseHub_FX_Storefront?node-id=30-761&p=f&t=BjiBQaBwDbhgju1n-0)

dozzzzer:
@kryksyh 
- Crash when clicking "Get Effects" when the Internet connection is cut off.

dozzzzer:
@kryksyh I cannot reproduce the crash anymore, but the design of the Share Audio / Get Effects button is corrupted: 

![Image](https://github.com/user-attachments/assets/8c7b8e60-08c8-4b88-b23a-93f578c17322)

kryksyh:
@dozzzzer yes this happened because you have opened two different builds of 3.7.2 with "Get effects" button and without it. This is expected and would not happen for users running release build.

dozzzzer:
Tested

PR #8123, opened 2025-01-29
Add get effects dialog

kryksyh:
Resolves: [#8043](https://github.com/audacity/audacity/issues/8043)

Adds "Get Effects" dialog.

Recommended:


grliszas14 (approved):

libraries/lib-musehub/MuseHubService.cpp
```diff
@@ -0,0 +1,188 @@
...
+   }
+)";
+
+static std::optional<EffectInfo> parseEffect(const rapidjson::Value& effectObj) {
+   if (!effectObj.HasMember("product") ||
+       !effectObj.HasMember("iconImageUrl") ||
+       !effectObj["product"].HasMember("code") ||
+       !effectObj["product"].HasMember("title") ||
+       !effectObj["product"].HasMember("subtitle")) {
+      assert(true);
```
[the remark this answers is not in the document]
kryksyh: obviously 😄 
thanks!

libraries/lib-theme/AColor.cpp
```diff
@@ -130,22 +130,31 @@ namespace
       dc.StretchBlit(x0, y2, xw0, yh2, &memDC, u0, v2, uw0, vh2, wxCOPY, true);
       dc.StretchBlit(x2, y2, xw2, yh2, &memDC, u2, v2, uw2, vh2, wxCOPY, true);
 
-      if(xw1 > 0 && yh1 > 0)//Draw middle
+      //Draw top/bottom mid section
+      if(xw1 > 0)
```
saintmatthieu, line 134: Double-checking: it mentions top and bottom, but the checked variable is an `x` variable.

libraries/lib-theme/AColor.cpp
```diff
@@ -130,21 +130,21 @@ namespace
...
-      //Draw top/bottom mid section
+      //Draw top/bottom sides mid section
       if(xw1 > 0)
       {
          dc.StretchBlit(x1, y0, xw1, yh0, &memDC, u1, v0, uw1, vh0, wxCOPY, true);
          dc.StretchBlit(x1, y2, xw1, yh2, &memDC, u1, v2, uw1, vh2, wxCOPY, true);
       }
 
-      //Draw left/right mid section
+      //Draw left/right sides mid section
```
saintmatthieu, line 140: Double-checking again: this second commit, "Fix drawing narrow toolbar buttons were missing mid-sections", actually only changes three comments.

libraries/lib-wx-wrappers/GradientButton.cpp
```diff
@@ -8,11 +8,15 @@
 
 #include "GradientButton.h"
 
-const wxColour darkBlue(68, 137, 247);
-const wxColour startColorPressed(59, 129, 231);
-const wxColour startColorNormal(75, 145, 247);
-const wxColour endColorPressed(38, 106, 214);
-const wxColour endColorNormal(54, 122, 246);
+static const wxColor darkBlue(68, 137, 247);
```
saintmatthieu, line 11: About this, I've been wondering repeated:
* wouldn't it be better to have them in an anonymous namespace to avoid the risk of duplicated symbols ?
* does it actually make a difference whether `static` or not ?

saintmatthieu (commented):
Finished with the graphics part, now moving on to the serious-stuff commit :D

libraries/lib-theme/AColor.cpp
kryksyh: thanks for noticing, it was meant to be squashed with the previous one, but got through.

libraries/lib-theme/AColor.cpp
kryksyh: yes it draws the middle section at the top and at the bottom of the button. This condition check if the middle section is present.
```


│        ┌─────┬─────┬─────┐ 
│        │     │     │     │    
│ yh0    │     │     │     │    
│        ├─────┼─────┼─────┤
│        │     │     │     │    
│ yh1    │     │     │     │   
│        ├─────┼─────┼─────┤
│        │     │     │     │    
│ yh2    │     │     │     │   
│        └─────┴─────┴─────┘
         ◄────►◄────►◄────►
          xw0    xw1    xw2
```

libraries/lib-wx-wrappers/GradientButton.cpp
kryksyh: In this particular case, an anonymous namespace would have the same effect as static. I can change to a namespace if you like, I guess it is a more C++ way of doing things.
I simply prefer static it because it is makes the intent more obvious, you don't have to scroll up to find out if there is a namespace.

saintmatthieu (approved):

Issue #7851, opened 2024-12-05
Playhead Cursor doesn't Follow Recording


chinakov:
### Bug description

If the User initiates a recording of a clip, the playhead cursor won't follow along the recording. 

### Steps to reproduce

1. Launch Audacity
2. Create a Mono or Stereo track
3. Start recording 
4. Observe result

### Expected behavior

The playhead cursor follows along with the recording.	

### Actual behavior

The playhead cursor remains at the original place. 

https://github.com/user-attachments/assets/10579e4e-76c7-45a8-b5f7-7bdf0ba1f186

### Audacity Version

Audacity 4 alpha

### Operating system

Windows

chinakov:
The cursor now follows playback.

PR #8127, opened 2025-01-29
Made playcursor follow recording

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7851

### DIFF
--- a/src/record/internal/au3/au3record.cpp
+++ b/src/record/internal/au3/au3record.cpp
@@ -188,6 +188,8 @@ void Au3Record::init()
 
     audioEngine()->commitRequested().onNotify(this, [this]() {
         commitRecording();
+
+        dispatcher()->dispatch("playback-seek", muse::actions::ActionData::make_arg1<double>(globalContext()->recordPosition()));
     });
 
     audioEngine()->finished().onNotify(this, [this]() {
@@ -352,6 +354,16 @@ muse::Ret Au3Record::stop()
 IAudioInputPtr Au3Record::audioInput() const
 {
     return m_audioInput;
+}
+
+muse::async::Channel<muse::secs_t> Au3Record::recordPositionChanged() const
+{
+    return m_recordPosition.ch;
+}
+
+secs_t Au3Record::recordPosition() const
+{
+    return m_recordPosition.val;
 }
 
 Au3Project& Au3Record::projectRef() const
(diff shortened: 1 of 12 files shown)


dozzzzer:
@grliszas14 With the playhead at T=0, the actual recorded clip's start is beyond that point.


https://github.com/user-attachments/assets/bf4de3c7-ddcc-4df9-bd81-0634ed49da71

src/record/internal/au3/au3record.cpp
```diff
@@ -188,6 +188,8 @@ void Au3Record::init()
 
     audioEngine()->commitRequested().onNotify(this, [this]() {
         commitRecording();
+
+        dispatcher()->dispatch("playback-seek", muse::actions::ActionData::make_arg1<double>(globalContext()->recordPosition()));
```
Eism, line 192: I think you can send the position through m_recordPosition.ch, right?

src/record/internal/au3/au3record.h
```diff
@@ -24,6 +26,8 @@ class Au3Record : public IRecord, public muse::async::Asyncable
 {
     muse::Inject<au::context::IGlobalContext> globalContext;
     muse::Inject<au::audio::IAudioEngine> audioEngine;
+    muse::Inject<IRecordController> recordController;
```
Eism, line 29: it looks like you don't use recordController

src/record/internal/recordcontroller.h
```diff
@@ -52,6 +55,7 @@ class RecordController : public IRecordController, public muse::actions::Actiona
     void stop();
 
     void setCurrentRecordStatus(RecordStatus status);
+    void setRecordPosition(muse::secs_t time);
```
Eism, line 58: also you don't use this method

src/record/internal/au3/au3record.h
```diff
@@ -4,11 +4,13 @@
...
 
+#include "actions/iactionsdispatcher.h"
 #include "global/async/asyncable.h"
 
-#include "modularity/ioc.h"
-#include "context/iglobalcontext.h"
 #include "au3audio/iaudioengine.h"
+#include "context/iglobalcontext.h"
+#include "irecordcontroller.h"
+#include "modularity/ioc.h"
```
Eism, line 13: please move it back to the top of the group
the order should be as follows: 
ioc - the main include without which other includes will not work
includes from the framework
includes from the other audacity modules
includes from the current module

Eism (approved):

PR #8157, opened 2025-02-04
Fix crash on opening mixer with accessibility voiceover enabled

kryksyh:
The mixer's meter does not have a slider, do not try to access it.

Recommended:


embarc-gabriel (approved):

dozzzzer:
Tested

Issue #7262, opened 2024-09-05
[AU4] Clips - Ability to change individual clip colour


ianalarina:
### Description

The ability to change an individual clip colour

[Figma link](https://www.figma.com/design/01XLIICmy2AG6t1ieyTp66/02---Clips?node-id=963-9280)

dozzzzer:
@grliszas14 Changing the clip colour doesn't make the project dirty and isn't undoable/redoable.

grliszas14:
QA note:
* making project dirty and redo/undo will be addressed in separate issue
* there is a bug in master which causes unresponsive clip context menu if it is accessed from the bottom, separate task here: https://github.com/audacity/audacity/issues/8186

dozzzzer:
Tested

PR #8158, opened 2025-02-04
Ability to change individual clip colour

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7262

### DIFF
--- a/src/projectscene/internal/projectsceneconfiguration.cpp
+++ b/src/projectscene/internal/projectsceneconfiguration.cpp
@@ -113,3 +113,20 @@ void ProjectSceneConfiguration::setIsEffectsPanelVisible(bool visible)
     }
     m_effectsPanelVisible.set(visible);
 }
+
+const std::vector<std::pair<std::string, std::string> >& ProjectSceneConfiguration::clipColors() const
+{
+    static std::vector<std::pair<std::string /*name*/, std::string /*color*/> > colors = {
+        { "Blue", "#66A3FF" },
+        { "Violet", "#9996FC" },
+        { "Magenta", "#DA8CCC" },
+        { "Red", "#F08080" },
+        { "Orange", "#FF9E65" },
+        { "Yellow", "#E8C050" },
+        { "Green", "#74BE59" },
+        { "Turquoise", "#34B494" },
+        { "Cyan", "#48BECF" }
+    };
+
+    return colors;
+}
--- a/src/projectscene/internal/projectsceneuiactions.cpp
+++ b/src/projectscene/internal/projectsceneuiactions.cpp
@@ -15,7 +15,7 @@ using namespace muse;
 using namespace muse::ui;
 using namespace muse::actions;
 
-const UiActionList ProjectSceneUiActions::m_actions = {
+static UiActionList STATIC_ACTIONS = {
     UiAction("automation",
              au::context::UiCtxProjectOpened,
              au::context::CTX_PROJECT_OPENED,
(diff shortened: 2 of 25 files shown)


src/projectscene/internal/projectsceneuiactions.cpp
```diff
@@ -145,7 +145,98 @@ const UiActionList ProjectSceneUiActions::m_actions = {
...
+             Checkable::Yes
+             ),
+    UiAction("color-auto",
+             au::context::UiCtxAny,
+             au::context::CTX_ANY,
+             TranslatableString("action", "Follow track color"),
+             TranslatableString("action", "Follow track color"),
+             Checkable::Yes
+             ),
+    UiAction("color1",
```
igorkorsukov, line 158: here we definitely need to clarify what exactly is "color1", something like that "change-clip-color1" 

but I would suggest doing it differently
* Add only two action: 
- `change-clipcolor-auto`
- `change-clipcolor` 
* When forming a menu, from the configuration get a list of available colors
* Generate menu items where the action is "change color" and the argument is "color value"

MenuItem has a method setArgs, for example see for example ClipContextMenuModel::load


Another option (I think better):
* Add only two action: 
- `action://trackedit/clip/change-color-auto`
- `action://trackedit/clip/change-color`
* When forming a menu, from the configuration get a list of available colors
* Generate menu items where the action is uri like:  `action://trackedit/clip/change-color?color=#123456` 

see as example `AppMenuModel::makeEffectsItems` 
```
MenuItem* item = makeMenuItem(effects::makeEffectOpenAction(meta.id).toString());
```

src/projectscene/internal/projectsceneuiactions.cpp
```diff
@@ -170,10 +178,31 @@ const UiActionList ProjectSceneUiActions::m_actions = {
...
 }
 
 void ProjectSceneUiActions::init()
 {
+    const auto& colors = configuration()->clipColors();
+    m_actions.clear();
+    m_actions.reserve(colors.size() + STATIC_ACTIONS.size());
+
+    for (const auto& color : colors) {
+        UiAction action;
```
igorkorsukov, line 191: previously it was necessary to register all actions with parameters... Now (if you update with the master) we can make only one static action `action://trackedit/clip/change-color`, as usual, like others

src/projectscene/internal/projectsceneconfiguration.cpp
```diff
@@ -113,3 +117,25 @@ void ProjectSceneConfiguration::setIsEffectsPanelVisible(bool visible)
...
+        { COLORS[0], "Blue" },
+        { COLORS[1], "Violet" },
+        { COLORS[2], "Magenta" },
+        { COLORS[3], "Red" },
+        { COLORS[4], "Orange" },
+        { COLORS[5], "Yellow" },
+        { COLORS[6], "Green" },
+        { COLORS[7], "Turquoise" },
+        { COLORS[8], "Cyan" }
+    };
```
igorkorsukov, line 133: I think it's better to make one method that returns colors and keep the color names and their values ​​next to each other, something like this: 

```


const std::vector<std::pair<std::string/*name*/, std::string/*color*/>>& ProjectSceneConfiguration::clipColors() const
{
    static  std::vector<std::pair<std::string/*name*/, std::string/*color*/>> colors = {
        { "Blue", "#66A3FF"},
        { "Violet", "#9996FC"},
        ...
    };

    return colors;
}

We don't have to worry about searching, with such a small number of elements it doesn't matter whether it's a map or a linear search in a vector
```

igorkorsukov (approved):

PR #8167, opened 2025-02-05
Added preview and apply VST effect

igorkorsukov:
QA:
- [x] Non-VST destructive effects still work
- [x] Realtime effects also still work


src/effects/vst/view/vstviewmodel.h
```diff
@@ -0,0 +1,43 @@
...
+    muse::Inject<IEffectInstancesRegister> instancesRegister;
+    muse::Inject<IEffectExecutionScenario> executionScenario;
+
+public:
+    VstViewModel() = default;
+
+    int instanceId() const;
+    void setInstanceId(int newInstanceId);
+
+    Q_INVOKABLE void onApply();
```
saintmatthieu, line 30: Do these correspond to the `Apply` and `Preview` buttons ? The fact that one has the `on` prefix and the other not suggests that they are different things.

igorkorsukov: Yes, a little different: 
* preview method - exactly performs preview
* onApply method - does not apply the effect, but it must be executed when applying (before applying). This method only updates the settings.

src/effects/vst/view/vstviewmodel.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void VstViewModel::onApply()
+{
+    updateSettings();
+}
+
+void VstViewModel::preview()
+{
+    updateSettings();
```
saintmatthieu, line 50: Why is this update necessary ? Won't this change the default settings, even if the user cancels after previewing ?

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
```diff
@@ -51,34 +51,34 @@ StyledDialogView {
             bbox.buttonById(ButtonBoxModel.Apply).enabled = false
         }
 
-        FlatButton {
```
saintmatthieu, line 54: Won't this break non-VST destructive effects ?

igorkorsukov: This is another dialog (another file)
Before this PR, VST view was technically a different implementation.
Now VST viewer is more like the implementation of other built-in effects. But the implementation itself is separate.
In principle, we can combine them, make them like other effects... But I have some doubts about this, maybe it is better to intentionally keep them separate, but similar.

igorkorsukov: oops, I was testing something and forgot to uncomment, thanks!

igorkorsukov:
> What is the vision about `option(AU_MODULE_EFFECTS_VST "Build audacity vst module" OFF)` ? Will it eventually default to ON ? Now is too early ?

I think it will always be disabled and will be enabled only for builds on CI (or manually for development).
The VST module requires the VST SDK, I don't want to force other developers to download it and configure it :)

src/effects/vst/view/vstviewmodel.cpp
saintmatthieu: > Here, roughly the same thing happens as with built-in effects.

Indeed. I didn't realize this. This differs from AU3, where cancelling preview also discards setting changes. I'll notify QA and we'll discuss this in triaging.

saintmatthieu:
> I think it will always be disabled and will be enabled only for builds on CI (or manually for development). The VST module requires the VST SDK, I don't want to force other developers to download it and configure it :)

In principle, I agree.
Not for the scope of this PR, but in AU3, conan downloads a pre-compiled SDK. If AU3 does it, I guess that much is allowed, @LWinterberg ? Did the MuseScore team ever considered doing the same ?

saintmatthieu (approved):
Looks good. Added QA items for regression checks.

igorkorsukov:
> Did the MuseScore team ever considered doing the same ?

I remembered, we almost did it
https://github.com/musescore/MuseScore/pull/26031

chinakov:
@igorkorsukov - just tested both destructive effects and realtime effects, no regressions have been noticed. Thanks!

saintmatthieu:
> I remembered, we almost did it

@igorkorsukov one advantage be that devs would notice locally already if they broke some of the vst-module code (which they'd otherwise only notice at CI stage) ?

Issue #8153, opened 2025-02-03
Stopping recording with Bluetooth headphones as the input device causes Audacity to freeze


dozzzzer:
### Steps to reproduce

1. Connect your BT headphones 
2. Launch Audacity and make sure the input device is set to the BT headphones
3. Create a new project
4. Start recording
5. Stop recording
6. Observe

### Expected behavior

No freeze, the recording is stopped.

### Actual behavior

Audacity freezes.

### Audacity Version

master

### Operating system

OS independent

IliasBergstromMuse:
I've reproduced this on MacOS - to occur it requires that both input and output are the same bluetooth device for it to happen.

AU3Player::stop() goes into an infinite loop, waiting for bool AudioIOBase::IsBusy() to return false - but it never does, because the AudioIOBase::mStreamToken flag is never set to 0.

I'll carry on seeing where this flag should normally be set, and why in this case it is not.

IliasBergstromMuse:
I've fixed it locally. I will clean up the code and do a PR in due time.

PR #8192, opened 2025-02-10
8153 stopping recording with bt as input causes freeze

IliasBergstromMuse:
Resolves: #8153

I reproduced the issue on MacOS Sonoma 14.7.3 - to occur it requires that both input and output are the same bluetooth device.

AU3Player::stop() went into an infinite loop, waiting for bool AudioIOBase::IsBusy() to return false - but it never does, because the AudioIOBase::mStreamToken flag is never set to 0.

The reason is what appears to be a race-condition: mPortStreamV19 is, by the time stop() is called, already NULL, and so the method returns - failing to also set mStreamToken to NULL. I address that by resetting mStreamToken at the point it returns. I cannot see that there would be any undesired side effects, and it fixes the problem.

For the changes made, I have written additional annotation in-line on the code ("Files changed" tab).

### On my fix:
Importantly, in terms of program structure, I don't think setting the flag where I do is particularly good code. But I suppose you agree it's consistent with how the AudioIO class is structured currently, and, seeing as it's part of the legacy AU3 engine, I imagine it is set for deprecation.

If not, I would have argued for refactoring the logic to make it less error-prone, more readable, and with a more clearly defined "state machine".

### On unit-test:
For the same reason, I have not unit-tested this change. I don't see any unit-tests for this entire class, where I would normally have made my addition.

### Question to reviewers: 
I have a preference to make small local "scout" cleanup changes in the files I touch, if I see obvious points, either with formatting, or C++ annotation. I've made a commit with such changes in this PR, to ask you, is that OK for Audacity? Some teams prefer to have such changes in separate PR's, others are fine with them being in the same PR, if the changes are separated by distinct commits, and the PR is still easy to read, as is the case here.

I could have made many more (formatting is very inconsistent, NULL is used instead of nullptr, etc), but this is legacy AU3 code so I won't spend the effort - not to mention it would have made this PR so messy that making a separate one would definitely have been required.

Recommended:

### DIFF
--- a/au3/libraries/lib-audio-devices/AudioIOBase.h
+++ b/au3/libraries/lib-audio-devices/AudioIOBase.h
@@ -307,7 +307,7 @@ protected:
    std::atomic<bool>   mPaused{ false };
 
    /*! Read by worker threads but unchanging during playback */
-   int                 mStreamToken{ 0 };
+   std::atomic<int>    mStreamToken{ 0 };
 
    /// Audio playback rate in samples per second
    /*! Read by worker threads but unchanging during playback */
--- a/au3/libraries/lib-audio-io/AudioIO.cpp
+++ b/au3/libraries/lib-audio-io/AudioIO.cpp
@@ -1419,8 +1419,10 @@ void AudioIO::StopStream()
       mRecordingSchedule.mCrossfadeData.clear(); // free arrays
    } );
 
-   if( mPortStreamV19 == NULL )
+   if( mPortStreamV19 == NULL ) {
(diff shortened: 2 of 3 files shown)


IliasBergstromMuse:
While my fix addresses the stopping after recording, I did not test for the case where first transport is already playing, before pressing record.

In that case, a similar freeze still appears, (which may or may not be the same code location as the one fixed).
Hence why I put this story back into Draft mode.

kryksyh (approved):

saintmatthieu (approved):

PR #8214, opened 2025-02-12
Added auto download vst3 sdk

igorkorsukov:


src/au3wrap/CMakeLists.txt
```diff
@@ -547,40 +547,33 @@ if(AU_MODULE_EFFECTS_VST)
         ${AU3_LIBRARIES}/lib-vst3/VST3Wrapper.h
         ${AU3_LIBRARIES}/lib-vst3/internal/ConnectionProxy.cpp
         ${AU3_LIBRARIES}/lib-vst3/internal/ConnectionProxy.h
-        ${AU3_LIBRARIES}/lib-vst3/memorystream.cpp
-        ${AU3_LIBRARIES}/lib-vst3/memorystream.h
+        # ${AU3_LIBRARIES}/lib-vst3/memorystream.cpp
```
saintmatthieu, line 550: Why not deleting these lines?

src/au3wrap/vst3/CMakeLists.txt
```diff
@@ -3,11 +3,31 @@
...
+        file(DOWNLOAD ${remote_url}/vst3sdk.cmake ${local_path}/vst3sdk.cmake
+            HTTPHEADER "Cache-Control: no-cache"
+        )
+    endif()
 
-set(VST3_SDK_PATH ${AU_MODULE_VST_VST3_SDK_PATH})
+    include(${local_path}/vst3sdk.cmake)
+
+    # func from ${name}.cmake)
+    cmake_language(CALL vst3sdk_Populate ${remote_url} ${local_path} "source" "" "")
```
saintmatthieu, line 26: Why not `vst3std_Populate(...` ?

saintmatthieu (approved):

src/au3wrap/CMakeLists.txt
igorkorsukov: deleted (checked if they are needed for other OS - they are not needed)

src/au3wrap/vst3/CMakeLists.txt
igorkorsukov: hmm, I made this system quite a long time ago and I don't remember why... (taken from here https://github.com/audacity/audacity/blob/master/buildscripts/cmake/SetupDependencies.cmake#L45)
Perhaps because this function doesn't exist until the file is downloaded, maybe there were problems with this...
Although I'm checking it now, it works, I'll investigate this point somehow, thanks for the question.

PR #8237, opened 2025-02-16
Fix issue CVE-2018-17096: Replace assert with runtime exception

tabudz:
Resolves: Possible vulnerability in SoundTouch library

The function BPMDetect() in this repository is nearly identical to BPMDetect() from SoundTouch.
The original function was patched due to a vulnerability identified in https://gitlab.com/soundtouch/soundtouch/-/commit/a1c400eb2cff849c0e5f9d6916d69ffea3ad2c85.
The same issue exists in this repository's function but remains unpatched.
This PR applies the same patch as the one in SoundTouch to eliminate the vulnerability.

References
CVE: [CVE-2018-17096](https://nvd.nist.gov/vuln/detail/CVE-2018-17096)
Original Fix: [Original Fix](https://gitlab.com/soundtouch/soundtouch/-/commit/a1c400eb2cff849c0e5f9d6916d69ffea3ad2c85)

Recommended:

### DIFF
--- a/au3/lib-src/soundtouch/source/SoundTouch/BPMDetect.cpp
+++ b/au3/lib-src/soundtouch/source/SoundTouch/BPMDetect.cpp
@@ -130,8 +130,11 @@ BPMDetect::BPMDetect(int numChannels, int aSampleRate)
 
     // choose decimation factor so that result is approx. 1000 Hz
     decimateBy = sampleRate / 1000;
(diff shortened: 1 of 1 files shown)


LWinterberg (approved):
I don't believe we use soundtouch's BPMdetect / detectBPM functions at all, so the vulnerability probably doesn't extend into Audacity. Thanks for the patch anyway!

Issue #7581, opened 2024-10-18
Add loading VST3 Effects


ianalarina:
### Description

Investigate and implement ability to use VST3 Effects in Audacity 4.

chinakov:
@igorkorsukov - https://github.com/audacity/audacity/issues/8262

PR #8256, opened 2025-02-17
Fixed VST view

igorkorsukov:
This PR and others complete this task
https://github.com/audacity/audacity/issues/7581


saintmatthieu (approved):

chinakov:
Notes:
- If the User saves a preset for a built-in destructive effect, the effect won't be closed upon saving a preset. However, for VST3 destructive effects, the effect window is moved to the background. If the User also loads a saved preset for VST3 effect, the window will also be moved to the background in window order, making it frustrating to operate. This happens only on macOS, Windows does not exhibit this behaviour. 


https://github.com/user-attachments/assets/c6bf30d0-143e-4850-8f20-23c58e67b2c5

Issue #8110, opened 2025-01-27
Not possible to distinguish increased audio of waveform/audio track in "Envelope Tool" (and black stripe when zoomed in)


matogolf:
### Bug description

Hey there,

I was using one of the older versions on Windows 11 (not sure which one exactly but at least 1,5 year old version) and I updated to the 3.7.1 version today.

I noticed that in all themes, when I use the Envelope Tool, I just can't see the difference if the track loudness was enhanced or not. I remember in the old version the waveform had slightly different colors from 0 to +-0.5 and a little bit different color from +-0.5 to +-1 on default. Now if waveform/loudness is increased with a simple drag of the mouse, I can see that at the very edge of the track the line is dotted, which indicates loudness is higher than visible, but there is no color difference between "dragged" and "not dragged" waveform.

Thanks

### Steps to reproduce

1. open any audio
2. select Envelope Took - F2 on windows
3. click on waveform and drag some point up

### Expected behavior

dotted line AND different colors are clearly indicating areas where points were dragged up - beyond default view from 0 to +-1

### Actual behavior

only dotted line is indicating areas where points were dragged up - beyond default view from 0 to +-1

### Audacity Version

Audacity 3.7.1

### Operating system

Windows

### Additional context

The grey line is just not visible. old versions had this line separating different colors, and it was clearly visible that area was dragged up... Thanks
<img width="389" alt="Image" src="https://github.com/user-attachments/assets/68ccf1c4-61d7-4449-9dd9-7edf1f835a9d" />

Might be related https://github.com/audacity/audacity/issues/6860

petersampsonaudacity:
Testing on W10 with 3.7.1 usinf Dark theme

I cannot reproduce this, I still see the pale line:

![Image](https://github.com/user-attachments/assets/6ad67ed6-b0fd-4a60-a121-76c7f9a137fc)

Peter

dozzzzer:
I was able to reproduce it:

<img width="450" alt="Image" src="https://github.com/user-attachments/assets/58ac1a2a-5cce-40d3-97d9-62949b6e6da3" />

LWinterberg:
Yeah, it seems like that got messed up in the last round of track redesign. I don't think it's particularly important to fix though - For Audacity4, we have a better solution, and a workaround is available (in the form of zooming out on the vertical axis)

dozzzzer:
Zoom in past a certain level and it starts to look particularly ugly, with a black stripe running across the clip:

[...]

PR #8257, opened 2025-02-17
Draw envelope area in the clip view

kryksyh:
Resolves: [#8110](https://github.com/audacity/audacity/issues/8110)

Recommended:

QA:
- [x] works with show-clipping ON (extra menus must be set in prefs, then it's under View)
- [x] rendering works for both selected and unselected samples
- [x] test with stretching
- [x] once envelope was drawn, select a tool other than the envelope tool to see if display is also correct

### DIFF
--- a/libraries/lib-wave-track-paint/waveform/WaveBitmapCache.cpp
+++ b/libraries/lib-wave-track-paint/waveform/WaveBitmapCache.cpp
@@ -24,10 +24,10 @@
 #include "WaveClip.h"
 
 // The worst case scenario is:
-// blank -> background -> min -> rms -> max -> backgroud -> blank
-// So we have 7 stops
+// blank > background > envelope top + envelope bottom > sample > rms > zero line
+// So we have 7 bands
 
-constexpr size_t ColorFunctionStops = 7;
+constexpr size_t ColorFunctionBands = 7;
 
 struct Triplet final
 {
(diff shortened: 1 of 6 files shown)


libraries/lib-wave-track-paint/waveform/WaveBitmapCache.cpp
```diff
@@ -176,25 +190,27 @@ struct WaveBitmapCache::LookupHelper final
          inputData = EnvRemappedColumns.data();
       }
 
-      const bool hasTopBlankArea = cache->mPaintParamters.Max > 1.0;
-      const auto globalMaxRow = GetRowFromValue(cache->mPaintParamters.Max);
-      const auto globalMinRow = GetRowFromValue(cache->mPaintParamters.Min) + 1;
+      const bool hasTopBlankArea = cache->mPaintParameters.Max > 1.0;
```
saintmatthieu, line 193: This isn't used anymore. What situation was it meant for, I wonder.

kryksyh: good catch, I'll remove it

> What situation was it meant for, I wonder.

I guess that the idea was to draw "Blank" area when zoomed out
This is audacity 3.3
image

But in 3.6-3.7 it just causes an assert if all features are enabled.

chinakov:
- Show clipping with Envelope tool works:
Screenshot 2025-02-18 at 18 05 51
- 'Enveloped' clips can be stretched, and rendered while retaining the envelope.
Screenshot 2025-02-18 at 18 13 15
- Once envelope has been used and drawn, the User can use other tools with no display or functional issues
https://github.com/user-attachments/assets/f84e89c7-bac4-4b53-a0b7-2b75a8d61a11
- rendering works for both selected and unselected samples
https://github.com/user-attachments/assets/c22d9fd2-aa79-49c5-b96b-7103a14f298f

Issue #8234, opened 2025-02-15
In the label editor changing a label's Track assignment requires the user to click OK twice


petersampsonaudacity:
### Bug description
In that Label Editor, if a user changes a label's Track assignment then they need to click the OK button not once but twice in order for the change to be effected.

This is in stark contrast to any other changes in the Label Editor which are effected immediately with a single press of the OK button.

This does NOT appear t o be a regression

### Steps to reproduce
1. Generate b a 30 second chirp
2. Add 3 labels: A, B and C
3. Add a new label track
4. Add a Label D in that new label track
5. Edit > Labels > Label Editor
6. Click on the Track entry for Label A
7. Select the second label track from the dropdown
8. Press the OK in the main dialog
9. Observe: nothing happens
10. Press OK in the main dialog again
11. Observe: Label A is only now moved to the second track

### Expected behavior
The label move should be effected with a single press of the OK button

### Actual behavior
The label move requires two presses of the OK button

### Audacity Version
Audacity 3.7.1 and the latest 3.7.2 Beta test build

### Operating system
Windows 10

### Additional context
Originally reported as a necro-post at the end of this Forum thread:
https://forum.audacityteam.org/t/labels-editor-okay-okay-suggestion-for-manual/54166/7

I have added an advice div to the 3.7.2 manual

Kurtsley:
Since the top half of this function is a duplicate of the Validate function, it could be removed. This would make the OK button close the dialog and apply the changes immediately.

https://github.com/audacity/audacity/blob/09985d81c5cb2aa74c888f56621aab20cd3167bb/src/LabelDialog.cpp#L922-L937

If I want to apply this fix, which branch should I be working under?

petersampsonaudacity:
Testing on W10 with @Kurtsley 's test branch build for this.

This now works fine - at Step 8 in the STRs the change of label track for Label A is effected - NO second press required

LWinterberg:
also cherrypicked into release-3.7.2

PR #8259, opened 2025-02-17
Fix label editor not closing on OK when editing

Kurtsley:
Resolves: #8234 

The OK button in the label editor now closes the dialog and applies the changes even if a track is being edited. This behavior is now in line with the rest of the boxes in the label editor. Clicking OK after making a change will close the dialog and apply the change immediately.

Recommended:


kryksyh (approved):
Thanks, @Kurtsley !

petersampsonaudacity:
Testing on W10 with @Kurtsley 's test branch build for this.

This now works fine - at Step 8 in the STRs for the bug report the change of label track for Label A is effected - NO second press required.  The Label editor closes correctly on the first press of the OK button.

PR #8260, opened 2025-02-17
Update get effects dialog

kryksyh:
update the query

Recommended:


libraries/lib-musehub/MuseHubService.cpp
```diff
@@ -35,28 +35,44 @@ static const std::string musehubEffectUtmMediumPrefix = "utm_medium=";
...
+                              code
+                              title
+                              subtitle
+                              category
+                           }
                         }
                      }
                   }
                }
+               ...on ProductPageSectionRegular {
```
Eism, line 57: need space after ... as in other places

Eism (approved):

PR #8261, opened 2025-02-17
Fix track drawing

kryksyh:
- Fix track not being fully drawn after another track is deleted
- Fix track drawing incorrectly on high zoom levels with high precision trackpad

Recommended:

### DIFF
--- a/libraries/lib-wave-track-paint/GraphicsDataCache.cpp
+++ b/libraries/lib-wave-track-paint/GraphicsDataCache.cpp
@@ -32,9 +32,7 @@ bool IsSameSample(double sampleRate, double t0, double t1) noexcept
 
 bool IsSamePPS(double sampleRate, double lhs, double rhs)
 {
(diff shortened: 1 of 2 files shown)


libraries/lib-wave-track-paint/GraphicsDataCache.cpp
```diff
@@ -32,9 +32,7 @@ bool IsSameSample(double sampleRate, double t0, double t1) noexcept
 
 bool IsSamePPS(double sampleRate, double lhs, double rhs)
 {
-   return std::abs(1.0 / lhs - 1.0 / rhs) *
-             GraphicsDataCacheBase::CacheElementWidth <
-          (1.0 / sampleRate);
+   return lhs == rhs;
```
saintmatthieu, line 35: Looks like `IsSamePPS` can be gotten rid of.

saintmatthieu (approved):

dozzzzer:
Tested!

PR #8266, opened 2025-02-18
Tooltips demo

LWinterberg:
I've mentioned to @petersampsonaudacity and @saintmatthieu that I'd like to move over from a reference manual to more reference-style tooltips. This PR is mainly to show the general thinking behind it: We're using a system that is overall quite "tooltipable" and allows for a title/description pattern - so perfect to add definitions to effect parameters and such. Visible changes in this build are tooltips on the transport buttons and inside the reverb effect. You'll need to hover over the label (rather than the slider/textbox) to see it for now though. 

For the tooltips, I tried to follow these rules: 
* The title should avoid repetition in the description. This is to avoid situations where it just says "Rewind to start: Rewinds to start"
* The description should use more beginner-friendly terms. 
* Only the positive case needs describing. So for on/off toggles, it's enough for the tooltip to say "enables xyz", without a need to specify a "click again to turn off". Similarly, for 0-100% sliders, only the 100% needs describing.

For the reverb, I pulled in and condensed the descriptions in the manual into single sentences. It's worth noting that our tooltip system keeps tooltips to reasonable sizes (so they don't turn into a massively long horizontal line with text), so longer paragraphs could be brought over too to explain more complicated subjects.


LWinterberg:
We'll likely design them in a way where they don't become a nuisance. In this example, the bubble is somewhat annoying as it covers up the text below it, but this could be fixed by making the tooltip appear on the left of the item, rather than below it. For other contexts, other solutions may work better, too - hard to say without knowing the exact looks and placements of all dialogs. I'd consider a preference for this a last-resort option, when there are too many places with tooltips which don't have a good method of getting out of the way.

petersampsonaudacity:
Typo:
image

BTW you have no tooltip for the "Wet Only"

petersampsonaudacity:
> It's worth noting that our tooltip system keeps tooltips to reasonable sizes (so they don't turn into a massively long horizontal line with text), so longer paragraphs could be brought over too to explain more complicated subjects.

Is it technically possible/feasible that hovering over a tooltip itself could then yield access to the "longer paragraphs" if they exist in that case.

LWinterberg:
I think expandable tooltips may be doable, though I don't know if we need them yet. If I copy the longest section from the reverb manual page, it still fits quite nicely. 

image

petersampsonaudacity:
> If I copy the longest section from the reverb manual page, it still fits quite nicely.

That longer tooltip looks absolutely fine to me.

Personally, _and I'm no way an expert or even user of reverb_, but I was thinking the short ones were a little too terse and not really explaining properly what the particular parameter controlled/affected.

DavidBailes:
Initial thoughts on accessibility.

For sighted keyboard users:
A tooltip has to be displayed when the control gains focus, as well as on hover, so that it can be read by keyboard users.
This is covered by the following WCAG level 2A guideline: https://www.w3.org/WAI/WCAG22/Understanding/keyboard
In addition, the following WCAG level AA guideline, https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html, specifies additional behaviour.

For users of screen readers:
The following info is for Windows screen readers. I don't currently know how the screen readers on macOS and Linux handle tooltips like these, but it may be similar.
The screen reader gets the content of a tooltip from a property of the accessible object of the control. There may be a choice of what property is used for this, but the description property is used by a couple of programs I looked at.
Having the description read out all the time when using Audacity would make things very noisy for screen reader users. So an option in Audacity as to whether the description was provided, which could easily be turned on and off, would probably be essential. By default Jaws does not read the description property, so the user would have to configure Jaws so that it was read when using Audacity.

Issue #8278, opened 2025-02-18
VST3 Instrument Plugins can be used Destructively


chinakov:
### Bug description

If the User has VST3 Instruments installed, Audacity 4 will show them in list of destructive list. 

### Steps to reproduce

1. Install any VST3 instrument
2. Launch Audacity
3. Create a new track
4. Go to Effect -> VST
5. Observe VST instrument
6. Open the VST intstrument
7. Observe result

### Expected behavior

The User should not be able to use VST instruments.

### Actual behavior

The User should is able to use VST instruments.

### Audacity Version

fce695e

### Operating system

OS independent

dozzzzer:
Tested

PR #8284, opened 2025-02-19
filtering vst plugins, only fx

igorkorsukov:
Resolves: https://github.com/audacity/audacity/issues/8278 

For correct operation, it is necessary to remove file: `userAppDataPath/known_audio_plugins.json`

### DIFF
--- a/src/effects/vst/internal/vst3pluginsmetareader.cpp
+++ b/src/effects/vst/internal/vst3pluginsmetareader.cpp
@@ -80,13 +80,22 @@ RetVal<AudioResourceMetaList> Vst3PluginsMetaReader::readMeta(const io::path_t& 
     }
 
     AudioResourceMetaList metaList;
-    for (const PluginDescriptor& descriptor : descriptors) {
+    for (const PluginDescriptor& desc : descriptors) {
+        //! NOTE At the moment AU only supports Fx and Fx|Generator,
+        //! see VST3EffectBase::GetType
+        muse::String type;
+        if (desc.GetEffectType() == EffectType::EffectTypeProcess || desc.GetEffectType() == EffectType::EffectTypeGenerate) {
+            type = u"Fx";
+        } else {
+            type = u"None";
+        }
+
         muse::audio::AudioResourceMeta meta;
-        meta.id = descriptor.GetID();
+        meta.id = desc.GetID();
         meta.type = muse::audio::AudioResourceType::VstPlugin;
-        // meta.attributes.emplace(muse::audio::CATEGORIES_ATTRIBUTE, String::fromStdString(descriptor.));
-        meta.vendor = descriptor.GetVendor();
-        // meta.hasNativeEditorSupport = hasNativeEditorSupport();
+        meta.attributes.emplace(muse::audio::CATEGORIES_ATTRIBUTE, type);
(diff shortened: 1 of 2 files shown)


src/effects/vst/internal/vst3pluginsmetareader.cpp
```diff
@@ -81,12 +81,16 @@ RetVal<AudioResourceMetaList> Vst3PluginsMetaReader::readMeta(const io::path_t&
 
     AudioResourceMetaList metaList;
     for (const PluginDescriptor& descriptor : descriptors) {
+        //! NOTE At the moment AU only supports Fx and Fx|Generator,
```
saintmatthieu, line 84: So there are VSTs that behave like our Audacity generators, i.e., they don't get MIDI input, only duration and plugin-specific parameters, to generate audio ? Never realized.

src/effects/vst/internal/vst3pluginsmetareader.cpp
```diff
@@ -81,12 +81,16 @@ RetVal<AudioResourceMetaList> Vst3PluginsMetaReader::readMeta(const io::path_t&
...
+        //! see VST3EffectBase::GetType
+        muse::String type = descriptor.GetEffectType() == EffectType::EffectTypeNone ? u"None" : u"Fx";
```
saintmatthieu, line 86: I see your comment, yet it'd feel safer to check for `descriptor.GetEffectType() == EffectType::EffectTypeProcess || descriptor.GetEffectType() == EffectType::EffectTypeGenerate`

src/effects/vst/internal/vst3pluginsmetareader.cpp
```diff
@@ -81,12 +81,16 @@ RetVal<AudioResourceMetaList> Vst3PluginsMetaReader::readMeta(const io::path_t&
...
+
         muse::audio::AudioResourceMeta meta;
         meta.id = descriptor.GetID();
         meta.type = muse::audio::AudioResourceType::VstPlugin;
-        // meta.attributes.emplace(muse::audio::CATEGORIES_ATTRIBUTE, String::fromStdString(descriptor.));
+        meta.attributes.emplace(muse::audio::CATEGORIES_ATTRIBUTE, type);
         meta.vendor = descriptor.GetVendor();
-        // meta.hasNativeEditorSupport = hasNativeEditorSupport();
+        meta.hasNativeEditorSupport = true;
```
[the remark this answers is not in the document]
igorkorsukov: This flag determines whether there is a view for the given audio resource (in a general sense)

For example:
* Previously, the muse framework did not support VST view in Linux, so this flag was set to false for Linux.
* There are some audio resources in MU that simply do not have a view, such as a built-in synthesizer. But others do, such as VST synthesizers.

src/effects/vst/internal/vst3pluginsmetareader.cpp
igorkorsukov: it looks like that

src/effects/vst/internal/vst3pluginsmetareader.cpp
igorkorsukov: fixed

Issue #7869, opened 2024-12-09
Implement 'Preset already exists' Dialog


chinakov:
### Description

If the User saves a preset with a name that already exists, Audacity 4 does not warn the User that a preset with such name already exists. 

Dialog window needs to state that such preset already exists, and it should allow the User to replace, keep, or cancel the preset creation. 

This feature exists in Audacity 3:

![image](https://github.com/user-attachments/assets/c9b8f7e6-8357-4d03-9ef7-ee47673c8e0c)

chinakov:
The dialog now exists and is shown to the User, however it would be nice if you could remove the capital 'R' in 'Replace'. 

@igorkorsukov - I suppose 'Audacity Development' header can't be changed at this point as well? 

<img width="324" alt="Image" src="https://github.com/user-attachments/assets/9428af3d-cc44-43f7-99e0-052531d9092b" />

PR #8285, opened 2025-02-19
Added check on preset already exists

igorkorsukov:
Resolves: https://github.com/audacity/audacity/issues/7869

### DIFF
--- a/src/effects/effects_base/internal/effectpresetsprovider.h
+++ b/src/effects/effects_base/internal/effectpresetsprovider.h
@@ -26,6 +26,7 @@ public:
     muse::async::Channel<EffectId> userPresetsChanged() const override;
 
     muse::Ret applyPreset(const EffectInstanceId& effectInstanceId, const PresetId& presetId) override;
+    bool hasUserPresetWithName(const EffectId& effectId, const std::string& presetName) const override;
     muse::Ret saveCurrentAsPreset(const EffectInstanceId& effectInstanceId, const std::string& presetName) override;
     muse::Ret deletePreset(const EffectId& effectId, const PresetId& presetId) override;
     muse::Ret importPreset(const EffectInstanceId& effectInstanceId, const muse::io::path_t& filePath) override;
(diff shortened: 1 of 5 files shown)


src/effects/effects_base/internal/effectpresetsscenario.cpp
```diff
@@ -41,6 +41,21 @@ void EffectPresetsScenario::saveCurrentAsPreset(const EffectInstanceId& effectIn
...
     }
 
+    EffectId effectId = instancesRegister()->effectIdByInstanceId(effectInstanceId);
+    bool alreadyExists = presetsProvider()->isHasUserPresetWithName(effectId, name);
+    if (alreadyExists) {
+        IInteractive::Result res = interactive()->question(
+            muse::trc("effects", "Save Preset"),
+            muse::mtrc("effects", "Preset \"%1\" already exists, Replace?")
+            .arg(String::fromStdString(name)).toStdString(),
+            { IInteractive::Button::No, IInteractive::Button::Yes });
```
saintmatthieu, line 51: I also can't think of a difference between the "No" and the "Cancel" button, but let's double-check with @LWinterberg

saintmatthieu, line 51: Is "No" going be the leftmost button ?

src/effects/effects_base/internal/effectpresetsprovider.h
```diff
@@ -26,6 +26,7 @@ class EffectPresetsProvider : public IEffectPresetsProvider
     muse::async::Channel<EffectId> userPresetsChanged() const override;
 
     muse::Ret applyPreset(const EffectInstanceId& effectInstanceId, const PresetId& presetId) override;
+    bool isHasUserPresetWithName(const EffectId& effectId, const std::string& presetName) const override;
```
saintmatthieu, line 29: `hasUserPresetWithName` would be correct grammatically and in my opinion sufficient.

src/effects/effects_base/internal/effectpresetsscenario.cpp
igorkorsukov: I didn't think about it.. here it would be more appropriate to cancel, I'll fix it

igorkorsukov: changed to Cancel

src/effects/effects_base/internal/effectpresetsprovider.h
igorkorsukov: fixed

Issue #7382, opened 2024-09-20
Discarding undo/redo history when closing project takes a long time to complete


Dan33185:
### Bug description

I am getting a pop up when I close a project in Audacity that says " Discarding undo/redo history". I’m not sure what it is, but it’s taking sometimes up to 20 minutes to complete before I can close Audacity. Seems to be more common with 24-Bit Files, and with versions newer than 3.5.1.

### Steps to reproduce

1. Open a .WAV file
2. Make any kind of edits
3. Export audio
4. Close the project
5. Close Audacity

A window then pops up saying "Discarding undo/redo history" with an ETA of when it will finish, which starts out at 30 seconds or so and grows as it proceeds. Average wait time is 15-20 minutes, I usually just end up force it closed through the task manager.

### Expected behavior

Audacity should close in a timely manner (a few seconds)

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows

brettcl:
I have a very similar experience since recent update (ver 3.6.3 or perhaps earlier?) but I am not using 24 bit files. Instead, I notice it when I have I have long recordings of a few hours or longer -- the longer the recording, the longer it takes to discard undo/redo history when quitting. 

When I quit and indicate NOT to save changes, it can take up to several minutes to complete the "Discarding undo/redo history". At no point have I saved the aup3 file, so there should be no need to save the undo/redo history.

It seems the correct thing to do would be to simply (and quickly) delete the working file and exit ASAP -- as it always used to do.

Not 100% sure but it doesn't seem to have anything to do with the number of edits, only the length of discarded material. 

Workflow details: I am recording audio at either 44.1kHz and 48 kHz often for several hours at a time. I set Audacity to Record with USB input to digitize old audio cassettes, so only first 60-90 minutes have audio. When I come back -- often hours later -- I stop, trim the file, edit, export to WAV/MP3, and then quit.

Thanks!

petersampsonaudacity:
> It seems the correct thing to do would be to simply (and quickly) delete the working file and exit ASAP

That seems eminently sensible to me too.

dozzzzer:
I tried importing a 1-minute stereo WAV file (24-bit) and then followed the STRs but to no avail. Also worth mentioning that I've used WAV 24-bit when exporting in step 3, if that matters.

saintmatthieu:
Hello @Dan33185, thanks for your report, and @brettcl, too, for contributing.

[...]

Issue #7312, opened 2024-09-11
Database sometimes not being compacted


SteveDaulton:
### Bug description

Some saved projects may be very much bigger than the data size. As an example, in  one test a project "AUP3" file was 2.4 GB although it contained less than 0.5 MB of data.

### Steps to reproduce

- Create a new project, and save to the disk
- Generate a 10m chirp
- Save the project and observe the file size, which should be ~100mb
- Delete the clip and save the project
- Close Audacity and observe the file size

### Expected behavior

Automatic compaction to reduce wasted disk space.

### Actual behavior

Some projects may be many GB larger than necessary.

### Audacity Version

current master / alpha / nightly build

### Operating system

Windows, Linux

### Additional context

Originally reported here: https://forum.audacityteam.org/t/superbig-project-files-after-audacity-update/117363
I have also reproduced the issue on Linux.

When conducting my tests, Audacity quit normally, so this does not appear to be directly related to the Windows crash on exit bug.

petersampsonaudacity:
It could be useful to re-instate the Compact Project command.

kryksyh:
@SteveDaulton @petersampsonaudacity @dozzzzer steps to reproduce were updated

notsmoothsteve:
> > It could be useful to re-instate the Compact Project command.
> 
> @petersampsonaudacity As a user I would expect the app to do that automatically. However, I agree that there should be a failsafe to force it manually. I've created another issue #7313 to take care of that.

I agree with both - it should do it automatically (as it used to), but it could also be useful to have a way to force it manually (with a warning, if applicable, about possibly unwanted side effects like losing undo history).  Often while working on a project, my project file ends up being a few gigabytes, the vast majority of which would be undo data, and even in the absence of the bug that's causing Audacity to fail to compact the project on close, it wouldn't be a bad thing to have an option for shrinking the file other than closing and reopening it.

chinakov:
Tested, fixed. Thanks!

PR #8296, opened 2025-02-20
8067 toggling effect makes project dirty

### LINKED ISSUE #8067: Turning an effect on and off doesn't make the project dirty
### Bug description

_No response_

### Steps to reproduce

1. Create a new project 
2. Add a track
3. Add an effect to it
4. Save it 
5. Turn the effect off
6. Try to close it

### Expected behavior

The user is asked if they want to save the project.

### Actual behavior

The project is closed right away.

### Audacity Version

Audacity 4 alpha

### Operating system

macOS, Windows, Linux

### Additional context

Discovered while testing https://github.com/audacity/audacity/pull/7997

saintmatthieu:
Resolves: #8067 

The fix of 8067 for track effects was easy, only needed the project to be made dirty.

For master effects, however, this was more complicated, and also is a defect on AU3.
The reason is that track effects are attachments of wave tracks, whose management when opening, making the project dirty and saving, was carefully implemented. The track attachments benefit from this maintenance. The maintenance of the project effects needed its own implementation.

Also, this ticket addresses an issue we have on master: create an empty project, save it, generate a 1h tone, close without saving: the project size is still several hundred MBs. (see #7312 wrt AU3). This must be implemented with care, though, that it doesn't lead to another issue, see #7382. The logic seen is similar to what can be found in `./au3/src/ProjectManager.cpp`.
This is addressed in this PR because solving this also helps fixing the original purpose of this ticket or "toggling effect makes project dirty".

Another issue this solves (although there wasn't a ticket for it) is that, when closing a project with unsaved changes, the solution so far was to rewind the history, which sometimes was also visible in the UI and looking a bit silly.

Recommended:

QA:
- [x] Project is compacted on close: create new project, save, generate 1h tone, close without saving: the project file should be about 20KB, not several megs. (See #7312 for more information. Was fixed for AU3 but not implemented in AU4.)
- [x] When creating a new project or opening an empty project, the master effect power button is enabled.
- [x] Toggling the power button of either master or track effects makes the project dirty
- [x] When closing and re-opening project (including re-starting application), power button states is correct whether saving changes or not before closing

### DIFF
--- a/au3/libraries/lib-realtime-effects/SavedMasterEffectList.cpp
+++ b/au3/libraries/lib-realtime-effects/SavedMasterEffectList.cpp
@@ -1,0 +1,42 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+#include "SavedMasterEffectList.h"
+#include "Project.h"
+
+SavedMasterEffectList::SavedMasterEffectList(AudacityProject& project)
+    : m_project(project)
+{
+    UpdateCopy();
+}
+
+void SavedMasterEffectList::UpdateCopy()
+{
+    mLastSavedList = std::make_unique<RealtimeEffectList>(RealtimeEffectList::Get(m_project));
+}
+
+static const AudacityProject::AttachedObjects::RegisteredFactory key2{
+    [](AudacityProject& project)
+    {
+        return std::make_shared<SavedMasterEffectList>(project);
+    } };
+
+SavedMasterEffectList& SavedMasterEffectList::Get(AudacityProject& project)
+{
+    return project.AttachedObjects::Get<SavedMasterEffectList>(key2);
+}
+
+const SavedMasterEffectList& SavedMasterEffectList::Get(const AudacityProject& project)
+{
+    return Get(const_cast<AudacityProject&>(project));
+}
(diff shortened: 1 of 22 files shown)


au3/libraries/lib-realtime-effects/SavedMasterEffectList.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+    UpdateCopy();
+}
+
+void SavedMasterEffectList::UpdateCopy()
+{
+    mLastSavedList = std::make_unique<RealtimeEffectList>(RealtimeEffectList::Get(m_project));
+}
+
+static const AudacityProject::AttachedObjects::RegisteredFactory key2{
```
kryksyh, line 18: now that it is in a separate file it could be renamed to `key`

kryksyh (approved):

igorkorsukov (approved):

dozzzzer:
@saintmatthieu Turning individual effects on and off doesn't make the project dirty while doing so on an entire stack does.

kryksyh (approved):

Issue #8146, opened 2025-01-31
Wrong selection cursor behavior


dozzzzer:
### Steps to reproduce

1. Open up a project with a track and clips on it
2. Try to make a selection
3. Try to make a selection while holding the cursor at the clip title bar level

### Expected behavior

- The selection cursor will always remain the same type until the LMB is released
- After the first selection is made, the following selections will turn the cursor to a regular pointer type
- When selecting over clips with the cursor at the clip title bar level, the cursor always remains visible

![Image](https://github.com/user-attachments/assets/96034ee2-4a63-434f-9d65-2affe64ce78b)

### Actual behavior

- The selection cursor turns into a sideways arrow as soon as you stop moving it and move slightly in the opposite direction (MacOS only)
- When making the 2nd and the subsequent selections, the cursor turns into a regular pointer type
- When selecting over clips with the cursor at the clip title bar level, the cursor disappears (MacOS only)

![Image](https://github.com/user-attachments/assets/c7ee33d7-7874-4e96-9f87-88e0754008a9)

![Image](https://github.com/user-attachments/assets/2c23e151-c63e-4a96-82f4-747931eba9f5)

### Audacity Version

master

### Operating system

OS independent

dozzzzer:
Works like a charm now! 

https://github.com/user-attachments/assets/abdd8d9f-8f60-4413-8c8b-87c37c0d72be

PR #8303, opened 2025-02-20
8222 smooth tool

### LINKED ISSUE #8222: Sample View - Smooth tool
### Description

Implement Smooth tool in Sample view mode

embarc-gabriel:
Resolves: #8222

Implements the ability to smooth samples whenever on stem view and ALT button is pressed.
One should also be able to smooth samples on different clips and tracks without releasing the ALT button.

Recommended:


src/appshell/view/navigableappmenumodel.cpp
```diff
@@ -314,7 +316,11 @@ bool NavigableAppMenuModel::processEventForAppMenu(QEvent* event)
         } else {
             if (m_needActivateHighlight) {
                 saveMUNavigationSystemState();
+                if (!m_moveFocusToApppMenyuOnAltRelease) {
```
Eism, line 319: do we need saveMUNavigationSystemState if we don't move focus to app menu?

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
```diff
@@ -15,6 +15,11 @@ static constexpr auto X_MIN_DISTANCE = 5;
 static constexpr auto Y_MIN_DISTANCE = 5;
 static constexpr auto BASELINE_HIT_AREA_SIZE = 20;
 
+static constexpr auto SMOOTHING_KERNEL_RADIUS = 3;
+static constexpr auto SMOOTHING_BRUSH_RADIUS = 5;
```
Eism, line 19: let's explicitly specify the variable type instead of auto

src/projectscene/view/clipsview/waveview.cpp
```diff
@@ -198,12 +201,23 @@ bool WaveView::isNearSample() const
 
 void WaveView::setIsNearSample(bool isNearSample)
 {
-    if (m_isNearSample == isNearSample) {
+    m_isNearSample = isNearSample;
```
Eism, line 204: hmm, could you explain why you removed if check for this variable? it was useful

src/projectscene/view/common/customcursor.h
```diff
@@ -14,6 +14,8 @@ class CustomCursor : public QObject
     Q_PROPERTY(int size READ size WRITE setSize NOTIFY sizeChanged FINAL)
 
 public:
+    static QString s_source;
```
Eism, line 17: Could you explain why it became static?

src/appshell/view/navigableappmenumodel.cpp
```diff
@@ -55,6 +55,8 @@ QSet<int> possibleKeys(const QChar& keySymbol)
 NavigableAppMenuModel::NavigableAppMenuModel(QObject* parent)
     : AppMenuModel(parent)
 {
+    //! NOTE: We do not move focus to the app menu when the Alt key is released so we can use it easily on other components
+    m_moveFocusToApppMenyuOnAltRelease = false;
```
saintmatthieu, line 59: It's initialized to true, set to false in the ctor, and then nothing happens to this member. Why?

src/projectscene/view/common/customcursor.h
saintmatthieu: And why public ?

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
```diff
@@ -15,6 +15,11 @@ static constexpr auto X_MIN_DISTANCE = 5;
 static constexpr auto Y_MIN_DISTANCE = 5;
 static constexpr auto BASELINE_HIT_AREA_SIZE = 20;
 
+static constexpr auto SMOOTHING_KERNEL_RADIUS = 3;
```
saintmatthieu, line 18: Minor: since these are only used in the smoothing function (right?) I'd move them to the said function.

src/projectscene/view/clipsview/waveview.cpp
```diff
@@ -256,3 +270,23 @@ void WaveView::setLastClickPos(const unsigned lastX, const unsigned lastY, const
         m_currentChannel.value(),
         globalContext()->currentProject(), m_clipKey.key, lastPosition, currentPosition, params);
 }
+
+void WaveView::smoothLastClickPos(unsigned int x, const unsigned int y)
+{
+    if (wavepainterutils::getPlotType(globalContext()->currentProject(), m_clipKey.key,
```
saintmatthieu, line 276: `if (!m_isStemPlot)` doesn't work?

src/projectscene/view/clipsview/waveview.cpp
```diff
@@ -256,3 +270,23 @@ void WaveView::setLastClickPos(const unsigned lastX, const unsigned lastY, const
...
+                                      m_context->zoom()) != IWavePainter::PlotType::Stem) {
+        return;
+    }
+
+    const auto currentPosition = QPoint(x, y);
+    const auto params = getWavePainterParams();
+
+    if (!m_currentChannel.has_value()) {
+        m_currentChannel = samplespainterutils::isNearSample(globalContext()->currentProject(), m_clipKey.key, currentPosition, params);
```
saintmatthieu, line 285: I was about to ask the same question (about using `m_isNearSample` directly) but I realize that this function doesn't return a bool, as its name suggests.
Looking at the impl, it seems that something like `hitChannelIndex` would be more helpful.

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
```diff
@@ -81,6 +81,8 @@ Rectangle {
     property bool hover: hoverArea.containsMouse || headerDragArea.containsMouse
     property bool headerHovered: headerDragArea.containsMouse
     property var lastSample: undefined
+    property bool altPressed: false
```
saintmatthieu, line 84: I don't know here, but I'll learn: since this member is not written within this document, would it work to make it `readonly` ?

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
```diff
@@ -81,6 +81,8 @@ Rectangle {
...
+    property alias isStemPlot: waveView.isStemPlot
```
saintmatthieu, line 85: `isStemPlot` is always used to check whether the brush is on. Looks like `property bool isBrush: waveView.isStemPlot && altPressed` instead of this alias would simplify a bit.

src/appshell/view/navigableappmenumodel.cpp
Eism: so that there is less difference with MuseScore, since we duplicated the solution, but we don’t want them to be too different, it easier to maintain

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
embarc-gabriel: Nice! I personally prefer this approach, but usually I am the only one on the team. Maybe we should have some kind of c++ guidelines or something like that.

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
embarc-gabriel: Same for the last comment. I prefer this way as well but usually people don't agree with me on this. I will change it. Thank you!

src/projectscene/view/clipsview/waveview.cpp
embarc-gabriel: Yeah! You are right on this one!

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
embarc-gabriel: It is modified by the outside component. I am not completly sure how this works on qml, but I will figure this out.

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
embarc-gabriel: I have tried something like this but many weird things started to happen. I think isBrush was not emitting onchange information or something like that. But I will try it again to understand what happened.

embarc-gabriel: ```Invalid property assignment: "altPressed" is a read-only property``
It complains once this can be modified by the outer component.

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
embarc-gabriel: **Invalid property assignment: "altPressed" is a read-only property**
Readonly fails because I try to modify it from the parent component.

saintmatthieu: Taking note of that, thanks for verifying!

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
embarc-gabriel: Done!

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
```diff
@@ -375,4 +381,105 @@ void setLastClickPos(const unsigned int currentChannel, std::shared_ptr<au::proj
...
+
+    auto waveMetrics = wavepainterutils::getWaveMetrics(project, clipKey, params);
+    const ZoomInfo zoomInfo { waveMetrics.fromTime, waveMetrics.zoom };
+
+    const auto time = zoomInfo.PositionToTime(currentPosition.x());
+    const auto sampleOffset = waveClip->TimeToSamples(time);
+    const auto adjustedTime = waveClip->SamplesToTime(sampleOffset);
+
+    //Get the region of samples around the selected point
+    size_t sampleRegionSize = 1 + 2 * (SMOOTHING_KERNEL_RADIUS + SMOOTHING_BRUSH_RADIUS);
```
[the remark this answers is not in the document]
embarc-gabriel: Hummm this is weird because for me it seems 1 sample for the central point and one radius for both directions and I think maybe creating defines for this would make it even harder to read.
I have improved the comment and moved it to a constexpr value once this can be computed on compile time.
Please, tell me if this seems reasonable.

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
embarc-gabriel: Done!

src/projectscene/view/clipsview/waveview.cpp
embarc-gabriel: Done!

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
embarc-gabriel: It is working. I don´t know what happened before. Thank you for this. It is indeed much better.

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
IliasBergstromMuse: There is this document:
https://github.com/musescore/MuseScore/wiki/CodeGuidelines

Which indeed doesn't mention the use of auto. It would be good to update to specify what we do. For example I see in the codebase that auto is not always used consistently, also in non-static code. Maybe there is a guideline that's followed and could be added for both non-static and static.

src/projectscene/view/clipsview/waveview.cpp
IliasBergstromMuse: Indeed it ensured that change notification is emitted only when the value is flipped.

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
embarc-gabriel: Yeah! I think this document has many format related topics but not so many other guidelines.
Of course I have my feelings about how things should be like "avoiding over using auto". But I don´t want to waste energy and I don't have strong opinions on many of these topics. Usually it is a matter of taste and balance. Anyway I can live using any of this approaches but the main point IMO is to keep consistency over the codebase. For this we should define what that means.

src/projectscene/view/clipsview/waveview.cpp
embarc-gabriel: Yeah! I have added this later on. Fixed.

src/projectscene/view/common/customcursor.h
embarc-gabriel: Private now!
@Eism @saintmatthieu @IliasBergstromMuse 
I have commented on the daily and I have added a TODO so I can handle this right way after finishing the sample view topic.
The way we are handling the cursor is not correct. We have a custom cursor inside the ClipItem so multiple clips can potentially modify the global cursor value and keep the state that is not global but local.
For example, we can modify the cursor to Pencil on clip 01 and then modify it to arrow on clip 02. After returning to 01 even if it should be pencil again the component won´t modify it once on the local copy m_source is still pencil.
Apart from that, we should handle the alt key for the brush tool and this should work not only between clips but also between tracks.
I had to force isNearSample to be emitted multiple time so the custom cursor is reevaluated to workaround that as well.
I am not saying the actual version is correct. It is not! But it workaround this until I can properly fix it by moving this customcursor to TracksClipsView component or other better solution.
I would like to know your thoughts on this.
This could be also related to this one: [#8146](https://github.com/audacity/audacity/issues/8146)

src/projectscene/view/clipsview/waveview.cpp
embarc-gabriel: I had explained that on the comment above.

src/appshell/view/navigableappmenumodel.cpp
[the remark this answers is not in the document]
embarc-gabriel: Done!

src/projectscene/view/common/customcursor.h
embarc-gabriel: I am handling the cursor from the outer component now.
No more modifications on this file.

src/projectscene/view/clipsview/waveview.cpp
embarc-gabriel: I put it back and modified the code to handle the cursor from the outer components. So this modification is not needed anymore.

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
```diff
@@ -106,15 +106,19 @@ Rectangle {
     }
 
     function mousePressAndHold(x, y) {
-        waveView.setLastClickPos(x, y - header.height, x, y - header.height)
+        root.altPressed
```
kryksyh, line 109: Now if the user clicks with hold (mouse down, wait 1s, mouse release) two events are sent: first press and hold, second - click, and smooth action is performed twice.
Do we need press and hold event at all?

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
```diff
@@ -371,8 +380,121 @@ void setLastClickPos(const unsigned int currentChannel, std::shared_ptr<au::proj
...
+        return;
+    }
+
     WaveChannelUtilities::SetFloatsFromTime(*waveChannel, startTime + waveClip->GetPlayStartTime(), samples.data(),
                                             samples.size(), narrowestSampleFormat,
                                             PlaybackDirection::forward);
 }
+
+void smoothLastClickPos(const unsigned int currentChannel, std::shared_ptr<au::project::IAudacityProject> project,
+                        const trackedit::ClipKey& clipKey, const QPoint& currentPosition, const IWavePainter::Params& params)
```
IliasBergstromMuse, line 394: Maybe you planned on doing this anyway once you're out of "draft" mode, but if not I'd like to have suggested it.

This method I think would benefit from being split up into smaller ones. 
As it is a free method you cannot make the sub-methods private class members, but you can then instead include them in in an anonymous namespace at the top of the file.

Then you can separate concerns a bit, and instead of signposting with comments it can be with method names.

embarc-gabriel: Yeah! You are right on this take!
I have added so many thing to handle the corner cases that it is now a method wiht many concepts.
I will change it.

embarc-gabriel: Ahhh I was talking about the other method ... anyway.
This applies to this method as well. :)

embarc-gabriel: Done!

src/projectscene/view/clipsview/au3/samplespainterutils.cpp
```diff
@@ -15,6 +15,91 @@ static constexpr auto X_MIN_DISTANCE = 5;
...
+{
+    // Now that the NEW sample levels are determined, go through each and mix it appropriately
+    // with the original point, according to a 2-part linear function whose center has probability
+    // SMOOTHING_PROPORTION_MAX and extends out SMOOTHING_BRUSH_RADIUS, at which the probability is
+    // SMOOTHING_PROPORTION_MIN.  _MIN and _MAX specify how much of the smoothed curve make it through.
+    float prob;
+    for (auto jj = -SMOOTHING_BRUSH_RADIUS; jj <= SMOOTHING_BRUSH_RADIUS; ++jj) {
+        prob
+            =SMOOTHING_PROPORTION_MAX
+              - (float)abs(jj) / SMOOTHING_BRUSH_RADIUS
```
IliasBergstromMuse, line 59: Nitpick: avoid old-school style casting, use static/dynamic/reinterpret casting as appropriate instead.

embarc-gabriel: Fixed. I have copied this from au3 and miss this.
Thank you!

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
embarc-gabriel: Good catch.
Fixed.

kryksyh (approved):

PR #8304, opened 2025-02-21
8170 unit testing au3interaction

### LINKED ISSUE #8170: Start Au3Interaction tests
Create a few tests for Au3Interaction. Create and load sample test .aup3 project.
Right now loading project needs to have audio devices initialized so some work is needed to decouple some of these dependencies.

IliasBergstromMuse:
Resolves: [8170](https://github.com/audacity/audacity/issues/8170#issue-2832924399)

This is unfinished, and only contains a short amount of work before we decided to put the above story back to the backlog. If we ever pick it back up, this can be revisited, addressing the many obvious issues of this WIP code.

Currently the introduced unit-test fails with a crash - just that it is further ahead than the last attempt at this task by colleagues.

If we pick this back up, this is what needs doing next:
I got as far as the project actually loading, and hitting the TrackColor constructor,
where it crashes because projectSceneConfiguration() is not set.
I haven't yet looked up how that is injected, when setting up mocks like this, to TrackColor.
Is it because I haven't instantiated an ioc()? Should I be, in unit-tests?

Recommended:


igorkorsukov:
> If we pick this back up, this is what needs doing next:
I got as far as the project actually loading, and hitting the TrackColor constructor,
where it crashes because projectSceneConfiguration() is not set.
I haven't yet looked up how that is injected, when setting up mocks like this, to TrackColor.
Is it because I haven't instantiated an ioc()? Should I be, in unit-tests? 

If we are testing not individual classes, but a part of the system as a whole, then we will probably need to use IoC in tests. There is nothing wrong with that, it can also be used to substitute mocks (register mocks instead of real services, where necessary)

This can be done in the test class itself. Or we can use an environment, see for example https://github.com/musescore/MuseScore/blob/master/src/engraving/tests/environment.cpp

PR #8308, opened 2025-02-21
TracksListModel doesn't modify track selection

### LINKED ISSUE #8088: Destructive Effect can't be applied after Undoing
### Bug description

If a User has undone any action, the User won't be able to apply a destructive effect to an audio clip. 

The user will be able to apply the effect if he applies the destructive effect twice after undoing. 

Here are some examples: 

https://github.com/user-attachments/assets/601d1317-2d3d-48b0-a01b-dd680f558c62

https://github.com/user-attachments/assets/5794409e-60d5-49af-9e22-9f740037dbd4

### Steps to reproduce

1. Launch Audacity 4
2. Create a stereo or mono track
3. Generate a clip into said track
4. Apply any destructive effect, like Fade In
5. Observe effect being applied
6. Undo 
7. Apply the Fade In effect once again
8. Observe result

### Expected behavior

The user should be able to apply an effect after undo

### Actual behavior

The User can't apply an effect after undoing

### Audacity Version

c461d7d

### Operating system

OS independent

### Additional context

_No response_

saintmatthieu:
Resolves: #8088

`TracksListModel::deleteItems()`, called when undo/redo happens, calls `m_selectionModel->clear();`, which has for side effect to clear the selection, which in itself is a problem, but also it interfers with destructive effect workflows, which depend on selection.

Removing this call fixes the selection and effect issues, and I could not notice that it caused other sorts of undesired side effects.

Recommended:

QA:
- [x] fixes #8088
- [x] Undo/Redo does not reset the selection like it used to

### DIFF
--- a/src/projectscene/view/clipsview/clipslistmodel.cpp
+++ b/src/projectscene/view/clipsview/clipslistmodel.cpp
@@ -746,6 +746,8 @@ void ClipsListModel::selectClip(const ClipKey& key)
 {
     Qt::KeyboardModifiers modifiers = keyboardModifiers();
 
+    constexpr auto complete = true;
+    constexpr auto modifyState = true;
     const auto clipGroupId = trackeditInteraction()->clipGroupId(key.key);
     if (clipGroupId != -1) {
         //! NOTE: clip belongs to a group, select the whole group
@@ -754,7 +756,7 @@ void ClipsListModel::selectClip(const ClipKey& key)
                 selectionController()->addSelectedClip(key);
             }
         } else {
-            selectionController()->setSelectedClips(trackeditInteraction()->clipsInGroup(clipGroupId));
+            selectionController()->setSelectedClips(trackeditInteraction()->clipsInGroup(clipGroupId), complete, modifyState);
         }
     } else {
         if (modifiers.testFlag(Qt::ShiftModifier)) {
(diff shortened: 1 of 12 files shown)


Eism (approved):

grliszas14:
@saintmatthieu there is a small side effect which we'll need to cover:
1. create a few tracks
2. do some undoable actions
3. undo
4. try to do multiple track selection with shift modifier -> on the first try it works just like you'd hold ctrl instead of shift (selects one track instead of multiple ones)

That probably happens because m_selectionModel is in a wrong state after undo.
But it can be handled within separate task if you ask me

saintmatthieu:
Thanks @grliszas14 .
In AU3, selection is implemented as an `UndoStateExtension`. Would such approach be a good solution ? Looks like this could bring the logic of detecting the changes in selection on undo/redo in a dedicated place and nowhere else.

src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -760,6 +760,8 @@ void ClipsListModel::selectClip(const ClipKey& key)
             selectionController()->setSelectedClips(ClipKeyList({ key.key }));
         }
     }
+
+    projectHistory()->modifyState();
```
Eism, line 764: ah, it should be in the selection service, a model shouldn't do this

saintmatthieu: Will address.

grliszas14 (approved):
lgtm, tested it briefly and it seems that the issue is fixed

the codestyle is failing btw

chinakov:
In the scope of this task, the clip selection after undo has been fixed. Created a new task for time selection here: https://github.com/audacity/audacity/issues/8384 

FYI @saintmatthieu

Issue #7877, opened 2024-12-10
Remove effect preview progress dialog


dozzzzer:
### Description

There should be no preview progress dialog when clicking the Preview button inside the effect dialog.

While preview is in progress, the button label should change from "Preview" to "Stop preview"

chinakov:
- After preview has been initialized, the User can't quit the effect or generator dialog via native close button, only 'close'. 

https://github.com/user-attachments/assets/0b675087-314b-44ea-9f48-5dbd76d68af8

- There is a noticable delay when Preview has been pressed, and the playback of the preview has been initialized. 

https://github.com/user-attachments/assets/76805095-f9f0-4898-997b-104cf41a3f49

No other issues have been noticed so far.

dozzzzer:
Tested

PR #8311, opened 2025-02-21
Removed effect preview progress dialog

igorkorsukov:
Resolves: https://github.com/audacity/audacity/issues/7877

### DIFF
--- a/src/effects/effects_base/ieffectexecutionscenario.h
+++ b/src/effects/effects_base/ieffectexecutionscenario.h
@@ -26,5 +26,6 @@ public:
     virtual muse::Ret repeatLastProcessor() = 0;
 
     virtual muse::Ret previewEffect(const EffectInstanceId& effectInstanceId, EffectSettings& settings) = 0;
+    virtual void stopPreviewEffect(const EffectInstanceId& effectInstanceId = -1) = 0;
 };
 }
--- a/src/effects/effects_base/ieffectsprovider.h
+++ b/src/effects/effects_base/ieffectsprovider.h
@@ -6,6 +6,7 @@
 
 #include "global/types/ret.h"
 #include "global/async/notification.h"
+#include "global/progress.h"
 
 #include "modularity/imoduleinterface.h"
 
@@ -44,6 +45,7 @@ public:
     virtual muse::Ret performEffect(au3::Au3Project& project, Effect* effect, std::shared_ptr<EffectInstance> effectInstance,
                                     EffectSettings& settings) = 0;
 
-    virtual muse::Ret previewEffect(au3::Au3Project& project, Effect* effect, EffectSettings& settings) = 0;
+    virtual muse::Ret previewEffect(au3::Au3Project& project, Effect* effect, EffectSettings& settings,
+                                    muse::ProgressPtr playProgress = nullptr) = 0;
 };
 }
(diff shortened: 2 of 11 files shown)


src/effects/effects_base/ieffectexecutionscenario.h
```diff
@@ -26,5 +26,6 @@ class IEffectExecutionScenario : MODULE_EXPORT_INTERFACE
     virtual muse::Ret repeatLastProcessor() = 0;
 
     virtual muse::Ret previewEffect(const EffectInstanceId& effectInstanceId, EffectSettings& settings) = 0;
+    virtual void stopPreviewEffect(const EffectInstanceId& effectInstanceId = -1) = 0;
```
saintmatthieu, line 29: Similarly here: why optional ?

src/effects/effects_base/ieffectsprovider.h
```diff
@@ -44,6 +45,7 @@ class IEffectsProvider : MODULE_EXPORT_INTERFACE
     virtual muse::Ret performEffect(au3::Au3Project& project, Effect* effect, std::shared_ptr<EffectInstance> effectInstance,
                                     EffectSettings& settings) = 0;
 
-    virtual muse::Ret previewEffect(au3::Au3Project& project, Effect* effect, EffectSettings& settings) = 0;
+    virtual muse::Ret previewEffect(au3::Au3Project& project, Effect* effect, EffectSettings& settings,
```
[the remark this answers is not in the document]
igorkorsukov: Usually when developing, I design a method not based on its current use, but as I would like to see it.
In this case, I want to show that progress does not necessarily have to be putted, the preview will work (it will play the entire selected segment). More strictly, this would be well shown using optional, but for me, `T* param = nullptr` is also generally accepted for such purposes.

src/effects/effects_base/ieffectexecutionscenario.h
igorkorsukov: I had this idea here: 
* At the moment we can only preview one effect instance at a time, so the effect instance itself does not need to be specified. 
* In the future, I admit that we can preview several instances of effects at the same time. I don't know why we would need this, but the services themselves allow us to do this. 
* Accordingly, we should also be able to make a stop for each instance separately. 
* And if the instance is not specified, then this means making a stop for everyone. 

This might not be the best design, and it would be better to make two methods, stop for the instance and stop for all... If that were needed. 
And perhaps even small provisions for the future can be confusing if they are not needed now... I need to think about it.

Anyway, I changed the design, removed this method. Now the cancellation is controlled by the effect instance model itself.

src/effects/effects_base/ieffectsprovider.h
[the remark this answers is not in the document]
saintmatthieu: I agree with nullptr.
Works for me.

saintmatthieu (approved):
I don't see that the playhead is reset to its original position after the preview, but that's an improvement which doesn't question the current PR.
Otherwise LGTM.

src/effects/effects_base/ieffectexecutionscenario.h
saintmatthieu: Thanks for the detailed explanation, as always :)

Issue #8095, opened 2025-01-24
When pasting multiple clip groups, they're pasted as individual ungrouped clips


dozzzzer:
### Steps to reproduce

1. Have two clip groups
2. Select them while holding Shift
3. Copy and paste

### Expected behavior

If the User copies multiple selected clip groups and then pastes them, they will paste as two separate clip groups - reflecting the original grouping of the copied groups.

### Actual behavior

 The copied-and-pasted clip groups are pasted as individual ungrouped clips.

https://github.com/user-attachments/assets/0d496bcb-bc2f-457a-b373-6ec0712f1378

### Audacity Version

master

### Operating system

OS independent

chinakov:
This is fixed, thanks!

PR #8312, opened 2025-02-21
Copy-pasting several groups now preserves the grouping

IliasBergstromMuse:
Resolves: When pasting multiple clip groups, they're pasted as individual ungrouped clips [#8095](https://github.com/audacity/audacity/issues/8095)

The existing code assumed that all clips were in a single group.
I've now refactored it to ensure that for each group in the source data, 
a new one is created in the pasted data.

Recommended:


src/trackedit/internal/au3/au3trackeditclipboard.cpp
```diff
@@ -90,3 +74,68 @@ bool Au3TrackeditClipboard::isMultiSelectionCopy() const
 {
     return m_isMultiSelectionCopy;
 }
+
+std::set<int64_t> Au3TrackeditClipboard::getGroupIDs(std::vector<TrackData>& tracksData)
+{
+    std::set<int64_t> groupIds;
+
+    for (auto& data : tracksData) {
+        auto waveTrack = dynamic_cast<au3::Au3WaveTrack*>(data.track.get());
```
IliasBergstromMuse, line 83: To connect to a conversation on a different PR about the auto keyword: here I think it should be auto - the type is in the cast statement, duplicating it is noisy repetition, it doesn't convey any new information.

src/trackedit/itrackeditclipboard.h
```diff
@@ -25,7 +25,7 @@ class ITrackeditClipboard : MODULE_EXPORT_INTERFACE
     virtual ~ITrackeditClipboard() = default;
 
     virtual std::vector<TrackData> trackDataSource() const = 0;
-    virtual std::vector<TrackData> trackDataCopy(int64_t newGroupId = -1) const = 0;
```
IliasBergstromMuse, line 28: I removed the groupID argument, since there can be more than one groups in the track data, the new IDs are instead generated inside the method.

src/trackedit/itrackeditinteraction.h
```diff
@@ -97,7 +97,6 @@ class ITrackeditInteraction : MODULE_EXPORT_INTERFACE
     virtual void setClipGroupId(const trackedit::ClipKey& clipKey, int64_t id) = 0;
     virtual void groupClips(const trackedit::ClipKeyList& clipKeyList) = 0;
     virtual void ungroupClips(const trackedit::ClipKeyList& clipKeyList) = 0;
-    virtual int64_t determineGroupId(const ClipKeyList& clipKeyList) const = 0;
```
IliasBergstromMuse, line 100: Also this method only makes sense for a new group, and is now only used inside au3interaction's groupClips(). So I removed it from the public interface.

src/trackedit/internal/au3/au3interaction.h
```diff
@@ -99,12 +99,13 @@ class Au3Interaction : public ITrackeditInteraction
     void setClipGroupId(const trackedit::ClipKey& clipKey, int64_t id) override;
     void groupClips(const trackedit::ClipKeyList& clipKeyList) override;
     void ungroupClips(const trackedit::ClipKeyList& clipKeyList) override;
-    int64_t determineGroupId(const ClipKeyList& clipKeyList) const override;
```
IliasBergstromMuse, line 102: This method only makes sense for a new group - it was misused before. Hence it now being renamed and private below.

src/au3wrap/internal/domaccessor.cpp
```diff
@@ -36,6 +36,7 @@ const Au3Track* DomAccessor::findTrackByIndex(const Au3Project& prj, size_t inde
             track = t;
             break;
         }
+        i++;
```
IliasBergstromMuse, line 39: I'm certain this was a bug so I fixed it (it would only ever match the first track since i was never incremented)

src/au3wrap/internal/domaccessor.h
```diff
@@ -31,7 +31,7 @@ class DomAccessor
 
     static size_t findClipIndexById(const Au3WaveTrack* track, const trackedit::ClipId& clipId);
 
-    static trackedit::ClipId findClipIdByIndex(const Au3WaveTrack* track,  size_t clipIndex);
+    static trackedit::ClipId findClipIdByIndex(const Au3WaveTrack* track, size_t clipIndex);
```
IliasBergstromMuse, line 34: And while at it I removed some stray whitespace.

grliszas14:
> Currently, as a group is deleted, the groupID is just unused, and then when a new group is created, it will be +1 from the last allocated one

If you pay closer look then you'll notice that it does increment ID by one but it always starts from 0 (not from the last allocated one), which means it will assign first unused ID the is ;)

`
    int64_t newGroupId = 0;
    while (muse::contains(groupsList, newGroupId)) {
        newGroupId++;
    }
`

IliasBergstromMuse:
> > Currently, as a group is deleted, the groupID is just unused, and then when a new group is created, it will be +1 from the last allocated one
> 
> If you pay closer look then you'll notice that it does increment ID by one but it always starts from 0 (not from the last allocated one), which means it will assign first unused ID there is ;)
> 
> ```c++
>     int64_t newGroupId = 0;
>     while (muse::contains(groupsList, newGroupId)) {
>         newGroupId++;
>     }
> ```

You're very right! I'll remove that bit from the description to keep it simple(r) - thanks!

src/trackedit/internal/au3/au3trackeditclipboard.cpp
```diff
@@ -90,3 +75,68 @@ bool Au3TrackeditClipboard::isMultiSelectionCopy() const
 {
     return m_isMultiSelectionCopy;
 }
+
+std::set<int64_t> Au3TrackeditClipboard::getGroupIDs(std::vector<TrackData>& tracksData)
```
grliszas14, line 79: tracksData can be marked const here

```suggestion
std::set<int64_t> Au3TrackeditClipboard::getGroupIDs(const std::vector<TrackData>& tracksData)
```

src/trackedit/internal/au3/au3trackeditclipboard.cpp
```diff
@@ -90,3 +75,68 @@ bool Au3TrackeditClipboard::isMultiSelectionCopy() const
...
+{
+    std::set<int64_t> groupIds;
+
+    for (auto& data : tracksData) {
```
grliszas14, line 83: data can be marked const as well
```suggestion
    for (const auto& data : tracksData) {
```

src/trackedit/tests/CMakeLists.txt
```diff
@@ -11,13 +11,14 @@ set(MODULE_TEST_SRC
...
 
     ${CMAKE_CURRENT_LIST_DIR}/au3interaction_tests.cpp
-    )
+    ${CMAKE_CURRENT_LIST_DIR}/au3trackeditclipboard_tests.cpp
+)
 
 set(MODULE_TEST_LINK
-    trackedit
-    au3wrap
+        trackedit
```
grliszas14, line 18: nitpick: minor formatting issue

src/trackedit/tests/au3trackeditclipboard_tests.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+using namespace au;
+using namespace au::au3;
+using namespace au::trackedit;
+
+namespace au::trackedit {
+class au3TrackEditClipboardTests : public ::testing::Test
+{
+public:
+    void SetUp() override
```
IliasBergstromMuse, line 27: A note on these:

Some 30 lines of this code are copied from Elnur's recent work on setting up Au3Interaction for unit-testing.
Currently I considered extracting these to avoid duplication but they are so few I figured I could just as well wait.
With that said, if se see further repetition, we should consider making a shared utility class / method collection for setting up Audacity projects for testing, that could also include methods like buildTrackData() and similar.

src/trackedit/internal/au3/au3trackeditclipboard.cpp
```diff
@@ -90,3 +75,68 @@ bool Au3TrackeditClipboard::isMultiSelectionCopy() const
...
+    auto groupsList = prj->groupsIdsList();
+    int64_t newGroupId = 0;
+
+    for (auto id : groupIDs) {
+        while (muse::contains(groupsList, newGroupId)) {
+            newGroupId++;
+        }
+        newGroupIds.push_back(newGroupId);
+        newGroupId++;
+    }
```
Eism, line 112: we need to reduce code duplication: I see you are using creating new group Id here and in `Au3Interaction::determineNewGroupId`

src/trackedit/internal/au3/au3trackeditclipboard.cpp
```diff
@@ -90,3 +75,68 @@ bool Au3TrackeditClipboard::isMultiSelectionCopy() const
...
+
+    return newGroupIds;
+}
+
+void Au3TrackeditClipboard::updateTracksDataWithIDs(const std::vector<TrackData>& tracksData,
+                                                    const std::set<int64_t>& groupIDs,
+                                                    const std::vector<int64_t>& newGroupIDs)
+{
+    assert(groupIDs.size() == newGroupIDs.size());
```
Eism, line 121: please use our assert from log.h: IF_ASSERT_FAILED

src/trackedit/internal/au3/au3trackeditclipboard.cpp
IliasBergstromMuse: You're right that they're a bit too similar - I should be able make this generalisable into one method indeed.

src/trackedit/internal/au3/au3trackeditclipboard.cpp
Eism: here yes, but for example, the line below and the line above are not clear what kind of object it is, for me it isn't easy to read the code :)

src/trackedit/internal/au3/au3trackeditclipboard.cpp
```diff
@@ -3,50 +3,35 @@
 */
 
 #include "au3trackeditclipboard.h"
+
+#include "Track.h"
+#include "containers.h"
 #include "WaveClip.h"
 #include "WaveTrack.h"
```
Eism, line 10: please follow our codestyle about includes https://github.com/musescore/MuseScore/wiki/CodeGuidelines#header-and-implementation

src/trackedit/internal/au3/au3trackeditclipboard.h
```diff
@@ -3,18 +3,19 @@
...
-
-#include "trackedittypes.h"
+#include "modularity/ioc.h"
+#include "context/iglobalcontext.h"
 
 namespace au::trackedit {
 class Au3TrackeditClipboard : public ITrackeditClipboard
 {
 public:
+    muse::Inject<au::context::IGlobalContext> globalContext;
```
Eism, line 15: it should be private, please move above the public

src/trackedit/tests/au3trackeditclipboard_tests.cpp
```diff
@@ -0,0 +1,200 @@
...
+            for (const auto& clipKey : selectedTrackClips) {
+                std::shared_ptr<Au3WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.clipId);
+                clipboardTrack->InsertInterval(waveTrack->CopyClip(*clip, true), false);
+            }
+
+            TrackData trackData { clipboardTrack, ClipKey() };  // (Dummy ClipKey)
+            trackDataList.push_back(trackData);
+
+            m_au3TrackEditClipboard->addTrackData(trackData);
+            m_au3TrackEditClipboard->setMultiSelectionCopy(true);
```
Eism, line 109: you can move it outside the loop

IliasBergstromMuse: You are right - i added it there mainly to replicate what happens in the actual code, where it's called after each addTrackData (in Au3Interaction::copyNonContinuousTrackDataIntoClipboard)
But it doesn't really test anything given it's a setter, so there's little point replicating the duplication.

src/trackedit/internal/au3/au3trackeditclipboard.cpp
IliasBergstromMuse: I'm more than happy to adapt to the "Muse" style! 

I will change these, and we could formulate an addition to the wiki at some point? I would do a PR but as far as I know I'm not part of that repo.

So that we avoid style discussion in PR's, and can just point to the wiki/uncrustifier.

Something like this (just a sketch):

On the use of auto:
* When the type is repeated in the line, e.g. after a cast, use the auto keyword to avoid repetition
* When the type is long and hard to read - e.g. due to long nesting and templatisation: If it's recurrent, create a descriptive alias with "using" keyword. If it only happens once, use auto.
* Otherwise write out the type

src/trackedit/internal/au3/au3trackeditclipboard.cpp
IliasBergstromMuse: I did a test to see if I can avoid duplication and pushed it (the latest commit) It's not bad, but it's more code, I'm not sure I love it :)

grliszas14:
lgtm - the only thing that makes me raise an eyebrow is that `createNewGroupID` is in the AudacityProject class instead of TrackeditProject where everything else concerning grouping is (might be easier to find later that way)

but I won't protest if it suits the others :P

IliasBergstromMuse:
> lgtm - the only thing that makes me raise an eyebrow is that createNewGroupID is in the AudacityProject class instead of 
> TrackeditProject where everything else concerning grouping is (might be easier to find later that way)

Thank you, you are right about that, it does make more sense there as it's the same module, I've moved it.

grliszas14 (approved):

Issue #8292, opened 2025-02-20
Add unit-test that loads an .aup3 project file and attempts to instantiate project


IliasBergstromMuse:
This test will currently fail, because au3interaction requires audio devices to be instantiated and available.

PR #8321, opened 2025-02-21
Au3interaction unit test example

Eism:
Resolves: #8292


src/au3wrap/internal/trackcolor.cpp
```diff
@@ -21,6 +21,10 @@ TrackColor& TrackColor::Get(const Track* track)
 TrackColor::TrackColor(Track& track)
     : mTrack{track.shared_from_this()}
 {
+    if (!projectSceneConfiguration()) {
+        return;
+    }
```
igorkorsukov, line 26: I think better add mock, like this https://github.com/musescore/MuseScore/blob/master/src/engraving/tests/environment.cpp 

in the future there will be more and more places like this

igorkorsukov (approved):

Issue #8258, opened 2025-02-17
Get effects dialog: accessibility: there is no default button.


DavidBailes:
### Steps to reproduce

1. Open the get effects dialog. The initial focus is the first item in the list/tree of pages.
2. Look to see if there is a visual indication that one of the commit buttons is the default button. (In this case there is one one such button).
3. Press Enter.

### Expected behavior

A user would expect the OK button to be the default button, and for this to be visually indicated.
So pressing Enter would close the dialog.

### Actual behavior

There is no visual indication of a default button.
Pressing Enter does not close the dialog.

### Audacity Version

Audacity 3.7.2

### Operating system

Windows

PR #8334, opened 2025-02-25
Get effects dialog: set a default button for the dialog

DavidBailes:
Resolves: https://github.com/audacity/audacity/issues/8258

Problem:
No button is set as the default button:

Fix:
1. Set the OK button to be the default button.
2. Change the OK button to be a wxButton rather than a Gradient button. This was done because a Gradient button does not currently give any visual indication when it is a default button, and it not obvious that this could easily be fixed.
3. Handle the wxEVT_TREE_KEY_DOWN event so that when the tree is the focus, pressing the Enter key closes the dialog. Similar code to this is used in the Preferences dialog.

Recommended:


kryksyh (approved):

kryksyh:
@DavidBailes, huge thanks for fixing all those issues I introduced!

PR #8338, opened 2025-02-25
7722 mute and solo implementation

### LINKED ISSUE #7722: Changing the track's panning, volume or mute/solo states should make the project dirty
### Description

Changing the track's panning, volume or mute/solo states should make the project dirty.

saintmatthieu:
Resolves: #7722

Recommended:

QA:
- [x] changing a track's controls (volume, balance, mute, solo) makes the project dirty
- [x] the state of track controls is restored correctly when closing and re-opening a project
- [x] Mute & Solo is now implemented, please test feature thoroughly
- [x] Double-clicking on volume or balance resets to middle
- [x] Follows [Figma spec](https://www.figma.com/design/k5rS9tnZl60nYmhJhvI2vf/Track-headers?node-id=263-33437&t=kXBP8PLYRLD6twvK-4)

### DIFF
--- a/src/playback/internal/au3/au3trackplaybackcontrol.cpp
+++ b/src/playback/internal/au3/au3trackplaybackcontrol.cpp
@@ -2,6 +2,7 @@
 
 #include "WaveTrack.h"
 #include "au3wrap/internal/domaccessor.h"
+#include "global/types/translatablestring.h"
 
 using namespace au::playback;
 using namespace au::au3;
@@ -12,7 +13,7 @@ Au3Project& Au3TrackPlaybackControl::projectRef() const
     return *project;
 }
 
-volume_dbfs_t Au3TrackPlaybackControl::volume(long trackId)
+volume_dbfs_t Au3TrackPlaybackControl::volume(long trackId) const
 {
     Au3WaveTrack* track = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
     IF_ASSERT_FAILED(track) {
@@ -22,7 +23,7 @@ volume_dbfs_t Au3TrackPlaybackControl::volume(long trackId)
     return LINEAR_TO_DB(track->GetVolume());
 }
 
-void Au3TrackPlaybackControl::setVolume(long trackId, volume_dbfs_t vol)
+void Au3TrackPlaybackControl::setVolume(long trackId, volume_dbfs_t vol, bool completed)
 {
     Au3WaveTrack* track = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
     IF_ASSERT_FAILED(track) {
@@ -30,10 +31,16 @@ void Au3TrackPlaybackControl::setVolume(long trackId, volume_dbfs_t vol)
     }
 
     track->SetVolume(vol > -60 ? DB_TO_LINEAR(vol) : 0);
+
+    if (completed) {
+        projectHistory()->pushHistoryState(muse::trc("playback", "Moved volume slider"),
+                                           muse::trc("playback", "Volume"),
+                                           trackedit::UndoPushType::CONSOLIDATE);
+    }
     return;
 }
 
-balance_t Au3TrackPlaybackControl::balance(long trackId)
+balance_t Au3TrackPlaybackControl::balance(long trackId) const
 {
     Au3WaveTrack* track = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
     IF_ASSERT_FAILED(track) {
@@ -43,7 +50,7 @@ balance_t Au3TrackPlaybackControl::balance(long trackId)
     return track->GetPan();
 }
 
-void Au3TrackPlaybackControl::setBalance(long trackId, balance_t balance)
+void Au3TrackPlaybackControl::setBalance(long trackId, au::audio::balance_t balance, bool completed)
 {
     Au3WaveTrack* track = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
     IF_ASSERT_FAILED(track) {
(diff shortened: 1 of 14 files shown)


src/projectscene/qml/Audacity/ProjectScene/trackspanel/audio/BalanceKnob.qml
```diff
@@ -31,4 +33,8 @@ KnobControl {
     onMouseExited: {
         tooltip.hide(true)
     }
+
+    mouseArea.onReleased: {
```
grliszas14, line 37: please use onMouseReleased { (unless it didn't work for some reason?)

src/playback/internal/au3/au3trackplaybackcontrol.cpp
```diff
@@ -51,5 +51,44 @@ void Au3TrackPlaybackControl::setBalance(long trackId, balance_t balance)
     }
 
     track->SetPan(balance);
-    return;
+}
+
+void Au3TrackPlaybackControl::setSolo(long trackId, bool solo)
```
grliszas14, line 56: all of the arguments in newly added functions can be made const
additionally getter functions can be marked const

saintmatthieu: > all of the arguments in newly added functions can be made const

You mean `setSolo(const long trackId, ...)` ? Where is this coming from ?

> additionally getter functions can be marked const

I agree. I guess they're not because they internally use non-const ´projectRef()`, but there should be a const version made of it. Will work on it.

saintmatthieu: Mmh, yes, maybe. It's one step more restrictive than in-body const declarations, though, since the API forces the implementation not to modify these primitives. It does so with `const` methods, too, I give you that, but for the scope of an entire method it's worth the risk IMO.

grliszas14 (approved):

src/playback/internal/au3/au3trackplaybackcontrol.cpp
Eism: I never use const for primitives 👀

saintmatthieu: @Eism I'm curious to know why.

src/projectscene/view/trackspanel/trackitem.cpp
```diff
@@ -217,6 +220,13 @@ void TrackItem::setVolumeLevel(float volumeLevel)
     emit outputParamsChanged(m_outParams);
 }
 
+void TrackItem::commitVolumeLevel()
+{
+    projectHistory()->pushHistoryState(muse::TranslatableString("playback", "Moved volume slider").translated().toStdString(),
```
Eism, line 225: again, the model shouldn't push something to history, it should be in service
you can use Au3Interaction approach with completed
 ```
   if (completed) {
        projectHistory()->pushHistoryState(...);
    }
```

and call setVolumeLevel with completed=true

File: src/projectscene/view/trackspanel/trackitem.cpp [hunk 4]
```diff
@@ -230,12 +240,20 @@ void TrackItem::setBalance(int balance)
...
 void TrackItem::setSolo(bool solo)
 {
     if (m_outParams.solo == solo) {
         return;
     }
 
+    trackPlaybackControl()->setSolo(trackId(), solo);
     m_outParams.solo = solo;
 
     // project::IProjectSoloMuteState::SoloMuteState soloMuteState;
```
Eism, line 259: if we have implemented solo/mute state, then I suggest deleting all commented code about solo/mute

src/projectscene/view/trackspanel/trackitem.cpp
Eism: we can call setVolumeLevel in qml with completed=true in release

Eism: or no, anyway we need to repeat the approach Au3Interaction

saintmatthieu: I'm lost ; could you summarize your recommendation ?

Eism: see ClipsListModel::trimLeftClip

src/projectscene/view/trackspanel/trackitem.h
```diff
@@ -30,7 +30,6 @@ class TrackItem : public QObject, public muse::async::Asyncable
     Q_PROPERTY(float leftChannelPressure READ leftChannelPressure NOTIFY leftChannelPressureChanged)
     Q_PROPERTY(float rightChannelPressure READ rightChannelPressure NOTIFY rightChannelPressureChanged)
 
-    Q_PROPERTY(float volumeLevel READ volumeLevel WRITE setVolumeLevel NOTIFY volumeLevelChanged)
```
Eism, line 32: we should keep it because of bindings in qml, but remove the seting value thought property and add another method for changing the value from qml with completed (as it is now, but need to rename setVolumeLevel)

saintmatthieu: I imagined that QML properties could be bound to `Q_INVOKABLE`s, but wrongly indeed - thanks !

src/projectscene/view/trackspanel/trackitem.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "playback/playbacktypes.h"
 
+#include "global/types/translatablestring.h"
```
saintmatthieu, line 11: not needed

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
```diff
@@ -75,7 +76,7 @@ Rectangle {
     radius: 4
     color: clipSelected ? "white" : clipColor
     border.color: "#000000"
-    opacity: root.moveActive && clipSelected ? 0.5 : 1.0
+    opacity: root.moveActive && clipSelected ? 0.5 : isAudible ? 1.0 : 0.3
```
grliszas14, line 79: is 0.3 opacity by design? just curious because if we move audible track then it becomes more transparent, but in case of inaudible clip it gets other way round (less transparent from its initial state)

src/projectscene/view/clipsview/trackslistclipsmodel.cpp
```diff
@@ -165,6 +169,18 @@ QVariant TracksListClipsModel::data(const QModelIndex& index, int role) const
     case IsMultiSelectionActiveRole: {
         return selectionController()->selectedClips().size() > 1;
     }
+    case IsTrackAudibleRole: {
+        if (trackPlaybackControl()->muted(track.id)) {
+            assert(!trackPlaybackControl()->solo(track.id));
```
grliszas14, line 174: I think the general rule of thumb is to use IF_ASSERT_FAIL from log.h instead of this, I've seen Elnur mentioning it somewhere

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
saintmatthieu: The 0.3 is part of the design, yes. I don't think it specifies what it should be when inaudible and dragged, but I assume the current values are fine.

grliszas14 (approved):

PR #8340, opened 2025-02-26
Qt debugging on Windows vscode

saintmatthieu:
Qt debugging for VSCode users on Windows.

Recommended:


kryksyh (approved):
would need a follow up PR fixing hardcoded paths, as the current solution is only usable on one specific PC

embarc-gabriel (approved):

kryksyh (approved):

Issue #7328, opened 2024-09-13
[AU4] Improve project state reloading on undo/redo


grliszas14:
Currently when doing undo/redo every track with every clips gets reloaded. This is visible in UI as very quick blink.
Improve that so it's invisible for the user (maybe update tracks instead of full reload)

IliasBergstromMuse:
There's two related PR's.

The main functionality:
https://github.com/audacity/audacity/pull/8388
And some "boy-scout" cleanup: 
https://github.com/audacity/audacity/pull/8387

chinakov:
Tested, everything seems to be in order, thanks!

PR #8344, opened 2025-02-26
Boy-scouting on clips and tracks list models, and minor optimisations to avoid copying in lambda arguments.

IliasBergstromMuse:
Relates to (but does not resolve): [[AU4] Improve project state reloading on undo/redo #7328
](https://github.com/audacity/audacity/issues/7328)

I did NOT address the issue in the story - rather I did some cleanup and minor optimisation. Below is my understanding of WHY there isn't much to do:

After reading the code, my conclusion is that this is AU3 legacy that we have to live with for now.

The reason is that the AU3 UndoManager does not follow the (in my view) best practice of the “Command” design pattern (often pairing well together with the “Composite” structure for the model). If that had been done, each “state” stored, could granularly contain only single changes. And when those are undone, notifications could be triggered to relevant “observers”, meaning that only affected elements (in UI or otherwise) would need updating.

But as it is now, the entire state, of all tracks, is stored for each action. And, restored, for each Undo/Redo. So we don’t know what has changed. And thus need to refresh everything.

Concrete example:
1. A clip is dragged 
1.1 The state is stored for each UndoStageExtension
1.1.1 One of these is also the TrackListRestorer - which stores copies of all tracks, not only the modified one(!).
2. Undo is called
2.1 RestoreUndoRedoState is called on each UndoStateExtension
2.1.1 Including TrackListRestorer, which will replace the state of all the tracks.

This is an unsustainable sledgehammer everything-and-the-kitchen-sink approach to undo-redo. 

I’ve looked at ClipsListModel, TracksListClipsModel, and TracksListModel, to see if there’s anything to optimise. The methods invoked in each - necessarily - are those for re-setting the entire view, since the UndoManager doesn’t notify of detailed changes, there’s nothing else to do.

I’ve made some very minor optimisations, but really I don’t see what can be done to vastly improve the performance other than re-implementing the change storage and notification. 

Recommended:

### DIFF
--- a/src/projectscene/view/clipsview/clipslistmodel.h
+++ b/src/projectscene/view/clipsview/clipslistmodel.h
@@ -39,15 +39,13 @@ class ClipsListModel : public QAbstractListModel, public muse::async::Asyncable,
     muse::Inject<trackedit::ISelectionController> selectionController;
 
 public:
-    ClipsListModel(QObject* parent = nullptr);
-    ~ClipsListModel();
+    explicit ClipsListModel(QObject* parent = nullptr);
+    ~ClipsListModel() override;
 
     TimelineContext* timelineContext() const;
     void setTimelineContext(TimelineContext* newContext);
     QVariant trackId() const;
     void setTrackId(const QVariant& newTrackId);
-    int selectedClipIdx() const;
-    void setSelectedClipIdx(int newSelectedClipIdx);
     bool isStereo() const;
 
     Q_INVOKABLE void init();
@@ -82,7 +80,7 @@ public:
     QHash<int, QByteArray> roleNames() const override;
     QVariant data(const QModelIndex& index, int role) const override;
 
-    int cacheBufferPx() const;
+    static int cacheBufferPx();
 
 signals:
     void trackIdChanged();
--- a/src/projectscene/view/clipsview/trackslistclipsmodel.cpp
+++ b/src/projectscene/view/clipsview/trackslistclipsmodel.cpp
@@ -35,13 +35,17 @@ void TracksListClipsModel::load()
         if (m_trackList.empty()) {
             return;
         }
-        emit dataChanged(index(0), index(m_trackList.size() - 1), { IsTrackSelectedRole });
-        emit dataChanged(index(0), index(m_trackList.size() - 1), { IsDataSelectedRole });
+
+        const int lastIndex = static_cast<int>(m_trackList.size()) - 1;
+        emit dataChanged(index(0), index(lastIndex), { IsTrackSelectedRole });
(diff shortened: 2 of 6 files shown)


src/projectscene/view/clipsview/clipslistmodel.h
```diff
@@ -39,15 +39,13 @@ class ClipsListModel : public QAbstractListModel, public muse::async::Asyncable,
...
-    ~ClipsListModel();
+    explicit ClipsListModel(QObject* parent = nullptr);
+    ~ClipsListModel() override;
 
     TimelineContext* timelineContext() const;
     void setTimelineContext(TimelineContext* newContext);
     QVariant trackId() const;
     void setTrackId(const QVariant& newTrackId);
-    int selectedClipIdx() const;
-    void setSelectedClipIdx(int newSelectedClipIdx);
```
IliasBergstromMuse, line 50: These were never defined (or used)

src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -69,21 +69,21 @@ void TracksListModel::load()
...
     for (const Track& track : tracks) {
         m_trackList.push_back(buildTrackItem(track));
     }
 
     onSelectedTracks(selectionController()->selectedTracks());
 
-    prj->tracksChanged().onReceive(this, [this](std::vector<au::trackedit::Track> tracks) {
-        onTracksChanged(tracks);
+    prj->tracksChanged().onReceive(this, [this](const std::vector<au::trackedit::Track>& tr) {
+        onTracksChanged(tr);
```
IliasBergstromMuse, line 86: Minor optimisation #1: tracks lambda argument is now by reference.

src/projectscene/view/clipsview/trackslistclipsmodel.cpp
```diff
@@ -35,40 +35,49 @@ void TracksListClipsModel::load()
...
             return;
         }
-        emit dataChanged(index(0), index(m_trackList.size() - 1), { IsDataSelectedRole });
+
+        const int lastIndex = static_cast<int>(m_trackList.size()) - 1;
+        emit dataChanged(index(0), index(lastIndex), { IsDataSelectedRole });
     });
 
-    prj->tracksChanged().onReceive(this, [this](const std::vector<au::trackedit::Track> tracks) {
+    prj->tracksChanged().onReceive(this, [this](const std::vector<au::trackedit::Track>& tracks) {
```
IliasBergstromMuse, line 71: Passing by reference instead of copying. 

This the only other optimisation - the rest is "boy-scouting" throughout.

IliasBergstromMuse:
Putting this back to draft, there's 2 things I should change - and, I'm building on this for the redraw optimisation. I will request a review again only when I'm done.

Issue #2893, opened 2022-05-03
Add ripple-editing


lukebrowell:
Figma: https://www.figma.com/design/gxreIBz2y3fYzwRo0nv9SP/Tools?node-id=280-5652

<details><summary>original report</summary>

I love Audacity's new in-place trimming and clipping behaviour; but, sometimes I want to split clips in three and then delete the middle segment in such a way that the third segment is aligned to the end of the first segment. This would be particularly useful for rough cuts when selecting within a clip and pressing delete is a little too fiddly and imprecise.

**Describe the solution you'd like**
I would like an 'alt-delete' behaviour which 'ripple deletes' the selected clip so that all of the other clips starting points in the current track are moved back by the length of the clip I have removed (similar to the behaviour when the EditClipsCanMove preference is set, but just temporarily for productivity reasons on a keyboard binding).

Example:

#### Step 1 - the original clip

``` 
|                         Clip A                        |
```

#### Step 2 - the clip split into three

```
|   part 1     |     part 2  |      part 3      |
```

#### Step 3 - alt-delete the middle segment causing part 3 to 'ripple-delete' alongside part 1

```
|   part 1     |      part 3      |
```

**Describe alternatives you've considered**
I've tried using markers - but this is much too slow and cumbersome.
I've tried dragging and selecting within a clip but this too is imprecise and human-error prone.
I’ve discovered the EditClipsCanMove preference (thanks @petersampsonaudacity), but it’s several clicks to turn this on and off while editing.

**Additional context**
I believe that the feature that I am suggesting would be complimentary to the current user experience and would not impinge upon the current user interface paradigm. Since this feature is mainly used for rough-cutting a single recording, it should not ripple affect clips on non-linked tracks (e.g. tracks which aren't part of the multi-track recording with the same timecode like stereo pairs). 

As an additional feature, being able to select the blank space between clips on a track and delete it to ripple the subsequent sub-clips forward would also be very helpful - but my guess is that this would require more work to implement than the 'alt-delete' feature I have proposed here.

</details>

petersampsonaudacity:
Why do you need to make the spit into 3 clips.

Just select what would be the range of the middle clip and press Delete.

petersampsonaudacity:
Or if you do want to use clips - just turn on "**Editing a clip can move other clips**" in **Tracks Behaviors** preferences and then delete the middle clip.

lukebrowell:
[...]

PR #8347, opened 2025-02-27
Add ripple-editing #2893

kryksyh:
Resolves: #2893

- Add ripple-editing actions
- Remove ClipsCanMove setting
- Remove SyncLock

Recommended:


au3/libraries/lib-wave-track/WaveTrackUtilities.cpp
```diff
@@ -328,20 +326,19 @@ void WaveTrackUtilities::ExpandCutLine(WaveTrack& track,
     });
     if (pClip != pEnd) {
         auto&& clip = *pClip;
-        if (!editClipCanMove) {
```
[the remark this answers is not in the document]
kryksyh: That were my thoughts. But you know, we could always throw the code away later, lets keep the logic of moving clips for now just in case.

I added `moveClips` as an (yet unused) argument.

src/trackedit/internal/trackeditactionscontroller.h
```diff
@@ -49,25 +49,36 @@ class TrackeditActionsController : public ITrackeditActionsController, public mu
...
     void doGlobalJoin();
     void doGlobalDuplicate();
 
+    void doGlobalCutLeaveGap();
+    void doGlobalCutPerClipRipple();
+    void doGlobalCutPerTrackRipple();
+    void doGlobalCutAllTracksRipple();
+
+    void pasteReplace();
+    void pasteInsert();
```
saintmatthieu, line 63: Not used

grliszas14:
```c++
dispatcher()->reg(this, PASTE_INSERT_CODE, this, &TrackeditActionsController::pasteInsert);
``` 
it's in the trackeditactionscontroller.cpp - diff is very big and doesn't load by default, maybe that's why you can't find it?

saintmatthieu: It's because I'm too silly - thanks :D

au3/libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -1376,13 +1385,15 @@ void WaveTrack::ClearAndPasteAtSameTempo(
     const auto tolerance = 2.0 / track.GetRate();
 
     // This is not a split-cut operation.
+    constexpr auto addCutLines = false;
```
saintmatthieu, line 1388: You could also consider removing all the cutline functionality as preliminary commit. We won't need it (just confirm with @LWinterberg before, but :D)

au3/libraries/lib-effects/EffectOutputTracks.cpp
```diff
@@ -35,9 +33,7 @@ EffectOutputTracks::EffectOutputTracks(
 
     auto trackRange = mTracks.Any()
```
saintmatthieu, line 34: You could now simplify to `mTracks.Selected<WaveTrack>()`

au3/libraries/lib-effects/PerTrackEffect.cpp
```diff
@@ -330,9 +329,6 @@ bool PerTrackEffect::ProcessPass(TrackList& outputs,
     };
     const auto defaultTrackVisitor
```
saintmatthieu, line 330: You could now probably use `outputs.Any<WaveTrack>().VisitWhile` and delete this guy.

au3/libraries/lib-label-track/LabelTrack.h
```diff
@@ -135,13 +135,12 @@ class LABEL_TRACK_API LabelTrack final : public UniqueChannelTrack<>, public Obs
     XMLTagHandler* HandleXMLChild(const std::string_view& tag) override;
     void WriteXML(XMLWriter& xmlFile) const override;
 
-    Track::Holder Cut(double t0, double t1) override;
+    Track::Holder Cut(double t0, double t1, bool moveClips) override;
```
saintmatthieu, line 138: Interesting to realize that ripple editing may some day extend to label tracks, too.

saintmatthieu (commented):
First round of review complete.

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -112,12 +131,25 @@ void TrackeditActionsController::init()
     dispatcher()->reg(this, COPY_CODE, this, &TrackeditActionsController::doGlobalCopy);
     dispatcher()->reg(this, CUT_CODE, this, &TrackeditActionsController::doGlobalCut);
```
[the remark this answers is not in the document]
kryksyh: Yes, I was torn about this as well. On one hand, it's yet another level of redirection that doesn’t do any useful work. On the other hand, this way we separate the abstract "Cut" from its specific implementation with chosen side effects.

 I don’t have a strong preference. @grliszas14 what's your thoughts on this?

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -226,61 +257,98 @@ void TrackeditActionsController::doGlobalCopy()
...
-            return;
-        }
-    }
+    dispatcher()->dispatch(MULTI_CLIP_CUT_CODE, moveClips);
 }
 
-void TrackeditActionsController::doGlobalSplitDelete()
+void TrackeditActionsController::doGlobalDelete()
+{
+    dispatcher()->dispatch(CUT_LEAVE_GAP_CODE);
```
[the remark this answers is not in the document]
kryksyh: Thanks, this is a _cut-paste_ error

src/app/configs/data/shortcuts.xml
```diff
@@ -227,12 +227,8 @@
...
     <SC>
-        <key>cut</key>
-        <seq>Ctrl+X</seq>
-    </SC>
-    <SC>
-        <key>delete</key>
-        <seq>Ctrl+K</seq>
+        <key>cut-per-track-ripple</key>
+        <seq>Ctrl+Shift+X</seq>
     </SC>
```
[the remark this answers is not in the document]
kryksyh: Design spec specs those as "unassigned" by default. That's why I added temprorary context menus items.

au3/libraries/lib-wave-track/WaveTrackUtilities.cpp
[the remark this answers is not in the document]
kryksyh: Thanks for the link to the issue. I haven't looked into it yet. If we remove any cut-lines functionality, would we be able to convert an old project that contains cut-lines?

If there is additional work to it, then it should be implemented and tested separately from the ripple editing.

au3/libraries/lib-label-track/LabelTrack.h
kryksyh: Yes, once LabelTracks are implemented, we will need to revisit this part.

kryksyh: This would be very useful combined with auto-transcription, say you have a podcast recording with a separate track for each participant, then you can ripple-trim all tracks according to label tracks.

src/app/configs/data/shortcuts.xml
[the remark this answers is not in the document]
saintmatthieu: Perfect.

src/trackedit/internal/trackeditactionscontroller.cpp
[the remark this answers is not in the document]
grliszas14: I would leave it as it is. I think the separation that @kryksyh is talking about it what we actually want to have.
Right now it seems to be reduntant but if we're going to introduce ripple editing preferences (should cut leave a gap or not), `doGlobalCut` will be a perfect place to redirect call to some other function based on your preference

au3/libraries/lib-wave-track/WaveTrackUtilities.cpp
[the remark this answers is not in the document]
saintmatthieu: We probably wouldn't ? Don't know. Either way, say we leave it like that and release AU4, opening a project with no UI to restore cut lines isn't ideal either. That wouldn't be as bad as not opening a project at all (if more sneaky), though, so you're probably right and have a dedicated ticket.

src/trackedit/internal/trackeditactionscontroller.cpp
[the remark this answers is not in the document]
saintmatthieu: Mh. If you're sure about this... If not I'd rather simplify now and, if needed later, making it more complex again would be easy.

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -226,61 +257,98 @@ void TrackeditActionsController::doGlobalCopy()
 }
 
 void TrackeditActionsController::doGlobalCut()
+{
+    dispatcher()->dispatch(CUT_LEAVE_GAP_CODE);
```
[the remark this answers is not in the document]
Eism: @kryksyh 
The main idea of ​​separating operations into global and local was that we have several places where the user can call the operation:
- via the context menu for the object. We know exactly the parameters with which we call. then we dispatch a specific action with specific parameters, for example clip-cut with clipKey
- via the shortcut, then we don't know what context we are in and call the doGlobalSmth method. This method should only determine the context and call the action for this context.

in this case doGlobal isn't working correctly, it should not just call another action

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -226,61 +257,98 @@ void TrackeditActionsController::doGlobalCopy()
...
+        secs_t selectedStartTime = selectionController()->dataSelectedStartTime();
+        secs_t selectedEndTime = selectionController()->dataSelectedEndTime();
+
+        dispatcher()->dispatch(SPLIT_CUT_SELECTED,
+                               ActionData::make_arg3<TrackIdList, secs_t, secs_t>(selectedTracks, selectedStartTime, selectedEndTime));
         return;
     }
 
     if (selectionController()->selectedClips().empty()) {
+        interactive()->error(muse::trc("trackedit", "No audio selected"),
```
Eism, line 277: if we take into account that the doGlobal method determine the context, then I think the error should be at the end of the method

src/app/configs/data/shortcuts.xml
```diff
@@ -243,16 +239,20 @@
         <seq>Ctrl+V</seq>
     </SC>
     <SC>
-        <key>duplicate</key>
-        <seq>Ctrl+D</seq>
+        <key>paste-insert</key>
```
grliszas14, line 242: Please have a look how paste-insert works for multiple selected clips, is it desired behaviour?

basic paste works like this:
* paste multiple clips -> these are pasted separately, one by one, so no gap between copied clips overrides existing clips
* paste range selection -> this pastes track as is, taking into account gaps between clips, overriding existing clips with silence if there are any

paste-insert for both multiclips selection and range selection behaves the same - it overrides clips with silence

src/app/configs/data/shortcuts.xml
```diff
@@ -243,16 +239,20 @@
...
+        <seq>Ctrl+Shift+V</seq>
+    </SC>
+    <SC>
+        <key>paste-insert-all-tracks-ripple</key>
```
grliszas14, line 246: This one is not implemented (but it is in the design)

src/app/configs/data/shortcuts.xml
grliszas14: nevermind, confirmed with Leo that this is desired behaviour

src/appshell/view/appmenumodel.cpp
```diff
@@ -203,7 +203,9 @@ MenuItem* AppMenuModel::makeEditMenu()
         makeMenuItem("group-clips"),
         makeMenuItem("ungroup-clips"),
         makeSeparator(),
-        makeMenuItem("split-cut"),
+        makeMenuItem("cut-per-clip-ripple"),
+        makeMenuItem("cut-per-track-ripple"),
+        makeMenuItem("cut-all-tracks-ripple"),
         makeMenuItem("duplicate"),
         makeMenuItem("insert"),
         makeMenuItem("split-delete"),
```
grliszas14, line 211: I believe that this action should be changed cause currently if you expand Edit menu, both "split-delete" and "delete" actions call the same "Delete and leave gap" behaviour. One of them should be "Delete and close gap"

src/trackedit/internal/trackeditactionscontroller.cpp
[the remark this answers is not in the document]
kryksyh: As @grliszas14 mentioned once we will have preferences for default cut and delete behavior, this wouldn't be a "just call another action"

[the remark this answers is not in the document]
LWinterberg: Hang on! We've decided to simplify this to not have prefs, so the default "cut" is just the current "cut and leave gap". We'll instead have a dialog thingy informing the user that ripple editing exists, which they then either can learn the shortcut for, or go to the shortcut prefs and rebind it.

LWinterberg:
FWIW: In the context menus, we just need the normal "Cut" (which leaves the gap) and "Cut and close gap (per track)". The other ones are intentionally more hidden.

src/app/configs/data/shortcuts.xml
kryksyh: fixed

src/appshell/view/appmenumodel.cpp
kryksyh: fixed

src/trackedit/internal/trackeditactionscontroller.cpp
kryksyh: @Eism I don't really understand your point here. Could you explain it a bit?

src/trackedit/internal/trackeditactionscontroller.cpp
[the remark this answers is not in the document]
kryksyh: Now doGlobalX performs the default X action without forwarding

au3/libraries/lib-wave-track/WaveTrackUtilities.cpp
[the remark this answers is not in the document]
kryksyh: > Similarly for sync-lock: I see some remnants of that in the AU3 libraries. Have you considered doing a more thorough clean-up ?

Can you point me to synclock leftovers? I can't find any references left in the project.

src/appshell/view/appmenumodel.cpp
```diff
@@ -203,10 +203,9 @@ MenuItem* AppMenuModel::makeEditMenu()
         makeMenuItem("group-clips"),
         makeMenuItem("ungroup-clips"),
         makeSeparator(),
-        makeMenuItem("split-cut"),
```
grliszas14, line 206: Is it intended that `makeMenuItem("cut-per-clip-ripple")` - is only available through clipcontextmenu and not from edit menu?

grliszas14: `delete-all-tracks-ripple` - is well hidden as well, you can access it only with a shortcut

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -228,59 +252,79 @@ void TrackeditActionsController::doGlobalCopy()
 void TrackeditActionsController::doGlobalCut()
 {
     if (selectionController()->timeSelectionIsNotEmpty()) {
-        dispatcher()->dispatch(RANGE_SELECTION_CUT_CODE);
+        auto selectedTracks = selectionController()->selectedTracks();
+        secs_t selectedStartTime = selectionController()->dataSelectedStartTime();
+        secs_t selectedEndTime = selectionController()->dataSelectedEndTime();
+
+        dispatcher()->dispatch(SPLIT_CUT_SELECTED,
```
grliszas14, line 259: I started renaming those actions so it's clearer what it does, but there's still a few to cover, could you please rename it to: "RANGE_SELECTION_SPLIT_CUT" ?

src/appshell/view/appmenumodel.cpp
kryksyh: @LWinterberg should we add "Remove Special" submenu similar to AU3?

grliszas14 (approved):

au3/libraries/lib-wave-track/WaveTrackUtilities.cpp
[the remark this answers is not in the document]
saintmatthieu: Sorry for the late reply.
Just checked out your branch and looked for SyncLock excluding `./au3/src`. I get 12 hits, mostly in lib-builtin-effects, but also lib-track-selection.

saintmatthieu (approved):
I hope this won't conflict too much with my stereo-to-mono PR 😬

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1873,13 +1891,35 @@ void Au3Interaction::doInsertSilence(const TrackIdList& trackIds, secs_t begin,
...
-            waveTrack->Clear(begin, end);
+            constexpr bool moveClips = false;
+            waveTrack->Clear(begin, end, moveClips);
         }
 
         prj->notifyAboutTrackChanged(DomConverter::track(waveTrack));
     }
 }
 
+void Au3Interaction::insertBlankSpace(const TrackIdList& trackIds, secs_t begin, secs_t duration)
```
IliasBergstromMuse, line 1902: Late to the party, but it's worth asking about unit-testing methods that introduce some none-trivial logic, like maybe this.

And to separate the logic from the au3interaction class, like Matt did in the mono/stereo drag PR - since it's getting very large and has a bit too many responsibilities it seems.

LWinterberg (approved):
This feels good to me! Thank you very much!!

Issue #8115, opened 2025-01-28
Delete Shortcut doesn't Delete Tracks


chinakov:
### Bug description

The User can't select and delete a track via 'Delete' key

### Steps to reproduce

1. Launch Audacity
2. Create a Track
3. Add an audio clip within the track
4. Select the Track
5. Press 'Delete' key

### Expected behavior

The Track is deleted

### Actual behavior

The Track is not deleted 

https://github.com/user-attachments/assets/dee65307-da83-443b-8e90-78a0667b9a5a

### Audacity Version

d8d5420

### Operating system

OS independent

grliszas14:
Hey,
It's not like you do something wrong, it looks like a bug to me. If you open the app and open any project, you can edit clips and "delete" action is triggered if you press "Delete" or "Backspace" keys. Once you click anything on trackpanel, recognizing those keys stops working - it doesn't work if you go back to the clipsview either.
Might be an issue with focusing/changing context, hard to tell without debugging.

dozzzzer:
Tested

PR #8351, opened 2025-02-28
Fix: #8115 Delete Shortcut doesn't Delete Tracks

GovindaMadhava:
Resolves: https://github.com/audacity/audacity/issues/8115

Before: (In the bug demo video) In a single clip track, when you select the clip and press delete, only the clip would vanish via "multi-clip-delete" action while the empty track would not be deleted.

Now: When you select a clip and press delete, the entire track is successfully deleted (provided track contains a single clip) via the "track-delete" action. Unlike before, the empty track doesn't remain.

Not just the clip, if you click on any non-track panel on the left including any empty/non-clip space in the track and press delete key, the entire track is successfully deleted.

What remains unsolved?? Keyboard triggered shortcuts are buggy once you click on the left side track panel. The delete key (and sometimes Ctrl+S too!) are not even recorded in the debug log for some reason.

Recommended:


kryksyh:
@GovindaMadhava Thank you for your contribution!

However, I’m not sure I understand the purpose of the changes in this PR. #8115 is related to Audacity4, while the code you modified is in the au3 directory, which contains legacy Audacity3 source code. This directory is not used in the Audacity4 build, so modifying it would not affect or fix issues in Audacity4.

PR #8362, opened 2025-03-03
Fix: 8115 track delete in AU4

GovindaMadhava:
Resolves: https://github.com/audacity/audacity/issues/8115

Earlier as shown in the bug demo video, when you select the clip and press delete key, only the clip would vanish via "multi-clip-delete" action but the empty track would remain undeleted.

Now: (I have made changes only within the audacity/src dir AU4 build only) 
Selecting a clip / any area in the track and pressing delete key successfully deletes the entire track (via track-delete action).

https://github.com/user-attachments/assets/f02fa3e6-ae63-45b9-af62-d7e9fb2d4c2c

Recommended:

### DIFF
--- a/src/app/configs/data/shortcuts.xml
+++ b/src/app/configs/data/shortcuts.xml
@@ -287,6 +287,10 @@
         <seq>Del</seq>
         <seq>Backspace</seq>
     </SC>
+        <SC>
+        <key>track-delete</key>
+        <seq>Delete</seq>
+    </SC>
     <SC>
         <key>file-new</key>
         <seq>Ctrl+N</seq>


GovindaMadhava:
Thanks for your detailed response @grliszas14.
1 doubt I had was regarding the au3 dir. I was instructed to make changes within the audacity/src dir for it to affect AU4 build and not within the au3 dir.
But since trackeditactionscontroller.cpp is within au3, how should I proceed? Are there any equivalent files within src/ that I need to find?

GovindaMadhava:
Thank you for your detailed response @grliszas14 ! I tried to follow how the key press is recognised and what actions are called.

I was able to understand that almost all keys are recognised and atleast raise an UiActionsRegister error and how the Action Dispatcher, ShortcutsController call relevant actions. But some key strokes like V and B weren't even recognised in the logs.

Apart from the logic of track-delete action, I was also not able to understand the issue for the key strokes not being recorded once clicked on the track panel.

@LWinterberg @grliszas14 Could you suggest any ideas to move ahead?

I'm relatively new to open source contributions. Would you suggest me to try out a couple more P3, P4 bugs before attempting this P1 feature?

PR #8363, opened 2025-03-03
8309 crossfade track fix

### LINKED ISSUE #8309: Crossfade Tracks is broken when applied to stretched clips
### Bug description

The result is incorrect when Crossfade Tracks is applied to the stretched clips, like the ones in the screenshot below.

Image

### Steps to reproduce

1. Have two stretched mono clips on two mono tracks partially overlapping one another (see above)
2. Make a selection across one clip's end and another clip's start
3. Effect > Crossfade Tracks
4. Observe

### Expected behavior

Image

### Actual behavior

Image

### Audacity Version

Audacity 3.7.2

### Operating system

OS independent

### Additional context

_No response_

saintmatthieu:
Resolves: #8309

When applying the crossfade-tracks effect (and probably others) over a time interval [t0, t1], clip "splits" within that interval are supposed to be preserved. In other words, if within the selection there are two clips one against the other, they should not be merged.

The clip-split detection and restoration algorithm was considering splits at the edge of this interval. This causing seemingly unnecessary clip manipulation (copying, inserting, trimming, shifting, etc) which, in that particular case, isn't handled properly. The exact reason as to why it isn't handled properly is still unknown to me. It may seem logically satisfying, however, not to try and restore splits at the edges of the time selection ; the "merge" argument to `WaveTrack::ClearAndPaste` alone should be responsible for that. Also, targeted testing around clip-split preservation may convince that this particular change isn't introducing new problems ?

Recommended:

QA:
- [ ] Resolves #8309
- [ ] Applying effects preserve clip splits like they used to, whether the splits are strictly within the time selection or at the edges.

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -1298,7 +1298,7 @@ void WaveTrack::ClearAndPasteAtSameTempo(
       // we need to copy clips, trims and names, because the original ones
       // could be changed later during Clear/Paste routines
       st = roundTime(clip->GetPlayStartTime());
-      if (st >= t0 && st <= t1) {
+      if (st > t0 && st < t1) {
          auto it = get_split(st);
          if (clip->GetTrimLeft() != 0) {
             //keep only hidden left part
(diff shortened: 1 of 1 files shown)


embarc-gabriel (approved):

LWinterberg:
tested

Issue #8343, opened 2025-02-26
Implement audio settings


grliszas14:
### Description

.

chinakov:
Tested, thanks!

Issue #8273, opened 2025-02-18
Prefs -> Audio settings gets stuck selecting options sometimes


LWinterberg:
Selecting the desired microphone didn't work for 2 users. Selecting another option from the dropdown would simply stay on the old selection.

chinakov:
Tested

PR #8368, opened 2025-03-04
Audio preferences

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8343


src/au3wrap/internal/au3audiodevicesprovider.h
```diff
@@ -35,6 +41,30 @@ class Au3AudioDevicesProvider : public playback::IAudioDevicesProvider
...
+
+    double bufferLength() const override;
+    void setBufferLength(double newBufferLength) override;
+    muse::async::Notification bufferLengthChanged() const override;
+
+    double latencyCompensation() const override;
+    void setLatencyCompensation(double newLatencyCompensation) override;
+    muse::async::Notification latencyCompensationChanged() const override;
+
+    std::vector<uint64_t> defaultSampleRateList() const override;
```
Eism, line 58: Or should it be availableSampleRateList?

src/appshell/qml/Preferences/internal/CommonAudioApiConfiguration.qml
```diff
@@ -81,22 +75,20 @@ Item {
...
-        //     model: apiModel.bufferSizeList
+            currentIndex: indexOfValue(apiModel.currentInputChannels)
+            model: apiModel.inputChannelsList
 
-        //     navigation.name: "BufferSizeBox"
-        //     navigation.panel: root.navigation
-        //     navigation.row: root.navigationOrderStart + 1
+            navigation.name: "RecordingChannelsBox"
+            navigation.panel: root.navigation
+            navigation.row: root.navigationOrderStart + 1
```
Eism, line 87: Could you please fix navigation for RecordingDevice (start + 1) and update this one (start + 2)?
(and please check the navigation for others controls)

Eism (approved):

LWinterberg:
#8273 appears unfixed at this time

https://github.com/user-attachments/assets/055ca998-f0d3-4282-adda-80754df6dfea

I'll also note that the data propagation seems to not quite work: When switching between MME, I expect my recording device to grow from the 31-char limit of MME to the full names. This appears to happen for the playback devices (although they instead get truncated by the dropdown with a ...) but the recording devices stay cut off (for me this is `Analogue 1 + 2 (Focusrite USB A` which should continue `udio)` or get truncated via ...)

On a design note, we probably don't need to truncate the text at all - just let the dropdown grow wider than its parent component.

----

image

The sample format is in need of its actual names - 16 bit, 24 bit and 32 bit float. 

----

The sample rate is missing the unit (Hz) as well as an "Other" option at the bottom of the list which, when selected, summons a small dialog where you can type in your desired sample rate. Once entered, it shows up as "Other (12345 Hz)"

chinakov:
If a User activates playback, and then opens the Audio preferences, the preferences will be inactive. Once the User adjusts the playhead once again, the preferences are available. 

Take a look at the VU meter during this process, might be because the software thinks that playback hasn't been stopped on pause. The audio preferences are active if the User presses stop instead of pause. 


https://github.com/user-attachments/assets/c6c0ebdb-259b-4095-9eda-6bf6993d745d

grliszas14:
@chinakov can you please create separate issue for this? I looks like this bug exists in master branch as well (and probably affects other parts of the app)

chinakov:
@grliszas14, sure, on it.

chinakov:
https://github.com/audacity/audacity/issues/8398

grliszas14:
@LWinterberg @chinakov please re-test
all of the issues should be fixed including:
-  [#8398](https://github.com/audacity/audacity/issues/8398) - I misunderstood this one, now you'll be able to tweak audio settings while playback is paused - however I see the issue on the video: moving playcursor makes playback go from PAUSED to STOPPED state, but the UI playback model may still think it is in PAUSED state (there may be some UI elements that stay disabled because of that)
-  [#8273](https://github.com/audacity/audacity/issues/8273)

Issue #8249, opened 2025-02-17
Get Effects dialog: accessibility: the effects are not automatically scrolled to ensure that the effect which has the keyboard focus is visible.


DavidBailes:
### Steps to reproduce

Case A:
1. Initial state:

![Image](https://github.com/user-attachments/assets/a8bed479-386e-435b-843a-ff9224ddb305)

2. The user presses Tab.

Case B:

1. Initial state:

![Image](https://github.com/user-attachments/assets/773feded-76d7-469c-8e62-cd8103fc7330)

2. The user presses Shift + Tab

### Expected behavior

Case A:
The effect which now has the keyboard focus is visible:

![Image](https://github.com/user-attachments/assets/4b327719-5268-4f4a-bfd0-ac5dd33aeff9)

Case B:
The effect which now has the keyboard focus is visible (note that all the text for the effect needs to be visible, not just the button):

![Image](https://github.com/user-attachments/assets/d74d4cf8-fe1a-4ddf-84c1-2da999e95ced)

### Actual behavior

Case A:

![Image](https://github.com/user-attachments/assets/c4b91513-f76b-4198-9eb4-4e428fe88839)

Case B:

![Image](https://github.com/user-attachments/assets/9333f7bc-b7db-4a10-a415-bdd2cbb2d457)

### Audacity Version

Audacity 3.7.2

### Operating system

Windows

PR #8380, opened 2025-03-04
Scroll effect panel into view on focus

kryksyh:
Resolves: #8249

Recommended:

### DIFF
--- a/modules/sharing/mod-musehub-ui/GetEffectsDialog.cpp
+++ b/modules/sharing/mod-musehub-ui/GetEffectsDialog.cpp
@@ -300,6 +300,27 @@ void GetEffectsDialog::AddEffectsPage(const std::string& group, const std::vecto
       button->Bind(wxEVT_BUTTON, [code = elem.code](auto) {
          BasicUI::OpenInDefaultBrowser(GetEffectUrl(code));
       });
+
+      button->Bind(wxEVT_SET_FOCUS, [page, itemPanel, button](wxFocusEvent& event) {
+         const int pageTop = page->GetScreenPosition().y;
+         const int pageHeight = page->GetClientSize().y;
+         const int buttonTop = button->GetScreenPosition().y;
+
+         const int scrollUpOffset = itemPanel->GetSize().GetHeight();
+         const int scrollDownOffset = button->GetSize().GetHeight() * 2;
+
+         int y, dy;
+         page->GetViewStart(nullptr, &y);
+         page->GetScrollPixelsPerUnit(nullptr, &dy);
+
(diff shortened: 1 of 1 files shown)


modules/sharing/mod-musehub-ui/GetEffectsDialog.cpp
```diff
@@ -300,6 +300,27 @@ void GetEffectsDialog::AddEffectsPage(const std::string& group, const std::vecto
...
+
+      button->Bind(wxEVT_SET_FOCUS, [page, itemPanel, button](wxFocusEvent& event) {
+         const int pageTop = page->GetScreenPosition().y;
+         const int pageHeight = page->GetClientSize().y;
+         const int buttonTop = button->GetScreenPosition().y;
+
+         const int scrollUpOffset = itemPanel->GetSize().GetHeight();
+         const int scrollDownOffset = button->GetSize().GetHeight() * 2;
+
+         int y, dy, scrollDelta;
```
embarc-gabriel, line 312: Nit. Unused variable.

modules/sharing/mod-musehub-ui/GetEffectsDialog.cpp
```diff
@@ -300,6 +300,27 @@ void GetEffectsDialog::AddEffectsPage(const std::string& group, const std::vecto
...
+         page->GetViewStart(nullptr, &y);
+         page->GetScrollPixelsPerUnit(nullptr, &dy);
+
+         if (buttonTop < pageTop)
+            page->Scroll(0, y - (pageTop - buttonTop + scrollUpOffset) / dy);
```
embarc-gabriel, line 317: Can GetScrollPixelsPerUnit or GetViewStart to return 0 on y or dy? Should we do some logic to avoid division by zero?

kryksyh: If `pixelsPerUnit` is set to 0, this effectively disables the scrolling, setting the client size to the full height of the inner content. As a result we will never get into any of if/else blocks. 

But you are right, it is better to be extra cautious, there could be bugs in wxWidgets.

saintmatthieu (approved):

LWinterberg:
This doesn't appear testable at the moment as none of the categories are long enough to make things scroll - but I imagine the other things I've been going through may change that. Please merge with the rest of 'em (or rebase thereupon if you need me to recheck before merge)

kryksyh:
@LWinterberg No, it wouldn't change. Currently all categories do not have enough elements to fill the page.

The only way to test this is to either change the height of the dialog, or push dummy elements. For both options you would need to modify sources.

PR #8385, opened 2025-03-05
Time, clip and track selections are retained after undo

### LINKED ISSUE #8384: Time Selection is not Retained Upon Undo
### Bug description

If the User makes a time selection that spans one or more tracks, submits an action (applying destructive effect, moving clip, etc.) and then undoes that action, the selection will be switched to a  clip selection.

### Steps to reproduce

1. Create several tracks
2. Generate a chirp in all tracks 
3. Make a time selection that spans multiple tracks (not clips)
4. Apply an effect
5. Make a new selection
6. Undo
7. Observe Result

### Expected behavior

After undo, the selection remains as a time selection spanning multiple tracks. 

### Actual behavior

After undo, the selection is limited to one clip. 

### Audacity Version

07f1dcdf832d10102fc17d955b467b4731759b40

### Operating system

OS independent

### Additional context

https://github.com/user-attachments/assets/b2423f32-2ec0-4932-82fd-079799b96d19

saintmatthieu:
Resolves: #8384

Recommended:

QA:
- [x] time selection is preserved after undo/redo
- [x] clip selection is preserved after undo/redo
- [x] track selection is preserved after undo/redo

### DIFF
--- a/src/effects/effects_base/internal/realtimeeffectservice.cpp
+++ b/src/effects/effects_base/internal/realtimeeffectservice.cpp
@@ -337,6 +337,7 @@ void RealtimeEffectService::setIsActive(const RealtimeEffectStatePtr& stateId, b
     }
     stateId->SetActive(isActive);
     projectHistory()->modifyState();
+    projectHistory()->markUnsaved();
     m_isActiveChanged.send(stateId);
 }
 
@@ -357,6 +358,7 @@ void RealtimeEffectService::setTrackEffectsActive(TrackId trackId, bool active)
     if (list && list->IsActive() != active) {
         list->SetActive(active);
         projectHistory()->modifyState();
+        projectHistory()->markUnsaved();
     }
 }
 
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -287,6 +287,9 @@ Rectangle {
                 }
 
                 if (e.button === Qt.LeftButton) {
+
+                    tracksModel.startUserInteraction()
+
                     if (root.clipHeaderHovered) {
                         tracksClipsView.clipStartEditRequested(hoveredClipKey)
                     } else {
@@ -340,6 +343,8 @@ Rectangle {
 
                     clipsSelection.visible = false
                 }
+
+                tracksModel.endUserInteraction()
             }
 
             onClicked: e => {
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -347,6 +347,11 @@ Rectangle {
                 tracksModel.endUserInteraction()
             }
 
+            onCanceled: e => {
+                console.log("User interaction canceled")
+                tracksModel.endUserInteraction()
+            }
+
             onClicked: e => {
                 if (e.button !== Qt.LeftButton) {
                     return
--- a/src/trackedit/internal/au3/au3projecthistory.h
+++ b/src/trackedit/internal/au3/au3projecthistory.h
@@ -30,11 +30,15 @@ public:
     void pushHistoryState(
         const std::string& longDescription, const std::string& shortDescription, UndoPushType flags) override;
     void modifyState(bool autoSave) override;
+    void markUnsaved() override;
+    void startUserInteraction() override;
+    void endUserInteraction() override;
     muse::async::Notification isUndoRedoAvailableChanged() const override;
 
 private:
     au3::Au3Project& projectRef();
 
+    bool m_interactionOngoing = false;
     muse::async::Notification m_isUndoRedoAvailableChanged;
     ::Observer::Subscription m_undoRedoMessageSubscription;
 };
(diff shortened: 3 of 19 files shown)


src/effects/effects_base/internal/realtimeeffectservice.cpp
```diff
@@ -357,6 +358,7 @@ void RealtimeEffectService::setTrackEffectsActive(TrackId trackId, bool active)
     if (list && list->IsActive() != active) {
         list->SetActive(active);
         projectHistory()->modifyState();
+        projectHistory()->markUnsaved();
```
IliasBergstromMuse, line 361: Is there ever a case where you'd want to call modifyState(), and not mark unsaved?

grliszas14: +1

there's example at trackslistmodel.cpp:155 in this very PR, the question is: should it be called without marking it unsaved

grliszas14 (commented):
lgtm in terms of selection, big plus for removing code that's not used anymore

src/effects/effects_base/internal/realtimeeffectservice.cpp
saintmatthieu: Yes: changing the selection should not make the project dirty (yet update the current history state).

src/trackedit/internal/au3/au3projecthistory.h
```diff
@@ -30,11 +30,15 @@ class Au3ProjectHistory : public IProjectHistory
     void pushHistoryState(
         const std::string& longDescription, const std::string& shortDescription, UndoPushType flags) override;
     void modifyState(bool autoSave) override;
+    void markUnsaved() override;
+    void startUserInteraction() override;
+    void endUserInteraction() override;
```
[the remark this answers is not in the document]
saintmatthieu: It's for the case where an action is begun of which it is known that it will either
1. add a new history entry or
2. just modify the state

but not which of the two. If it turns out to be 1., it ensures that `modifyState` is not called until the history push ; and if it turns out to be 2, it will modify the state.

(I realize that the call to `modifyState` in `startUserInteraction()` isn't needed, actually, even if it doesn't hurt UX. I just pushed an amend.)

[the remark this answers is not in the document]
IliasBergstromMuse: Ok, got it!

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -332,6 +335,13 @@ Rectangle {
 
                     clipsSelection.visible = false
                 }
+
+                tracksModel.endUserInteraction()
+            }
+
+            onCanceled: e => {
+                console.log("User interaction canceled")
```
grliszas14, line 343: please remove debug message from here, I think no matter which one you use (console.log() or console.debug()) both will be visible in Release build anyway, better to move it to tracksModel::endUserInteraction() on c++ side with LOGD() or sth like that

saintmatthieu: `tracksModel::endUserInteraction()` can be called from `onRelease` or here, `onCanceled`. If I look into issues I may want to see where the call originates. What's the trouble in having it visible in release builds ?

src/trackedit/iprojecthistory.h
```diff
@@ -25,6 +25,27 @@ class IProjectHistory : MODULE_EXPORT_INTERFACE
...
+     * Calling `startUserInteraction()` will protect against inadvertent state modifications in the middle of an interaction,
+     * which can introduce bugs in subsequent usage of undo.
+     * If, by the time `endUserInteraction()` is called, no new history entry was pushed, the current state will be modified automatically.
+     */
+    virtual void startUserInteraction() = 0;
+
+    /**
+     * @ref startUserInteraction()
+     */
+    virtual void endUserInteraction() = 0;
```
IliasBergstromMuse, line 47: Now that I’ve looked at the code more closely I agree that these start/end methods make sense to move. To explain why: ”[Gather together the things that change for the same reasons. Separate those things that change for different reasons](https://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html)”. In this case, ProjectHistory should not care where history originates from, and should not treat entries differently depending on any external reasons. Its job is to store and allow modification of history. So the decision on whether an entry should be stored or not, lies in the selection controller more closely. Especially since this addition relates only to selection (would this ever change?). (edited)

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
grliszas14: It just looks like a bad practice for me - the only logs I'd like to see in the Release build are warnings and errors probably - in the end we have debug builds for a reason :D 
It also has a minor performance impact - very minor in this case but...

saintmatthieu: Actually, it yields `DEBUG` level traces:
```qml
                console.log("pressed")
                console.debug("pressed")
```
```
18:15:35.749 | DEBUG | main_thread     |                 | pressed
18:15:35.750 | DEBUG | main_thread     |                 | pressed
```

saintmatthieu: ... and in principle, someone who needs to diagnose a problem and only just has a log may very well need traces beyond just warnings and errors.

grliszas14 (approved):

PR #8387, opened 2025-03-05
7328 - Boyscouting on clips and tracks list model classes, and removed unused includes.

IliasBergstromMuse:
Resolves nothing. It's a cleanup ("boyscout") PR accompanying this PR: 
https://github.com/audacity/audacity/pull/8388

Recommended:

### DIFF
--- a/src/projectscene/view/trackspanel/trackslistmodel.cpp
+++ b/src/projectscene/view/trackspanel/trackslistmodel.cpp
@@ -66,7 +66,7 @@ void TracksListModel::load()
 
     TRACEFUNC;
 
-    ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+    const ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
     if (!prj) {
         return;
     }
@@ -503,7 +503,7 @@ void TracksListModel::onProjectChanged()
 
 TrackItem* TracksListModel::buildTrackItem(const Track& track)
 {
-    TrackItem* item = new TrackItem(this);
+    auto item = new TrackItem(this);
     item->init(track);
 
     return item;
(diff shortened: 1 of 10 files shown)


igorkorsukov (approved):

src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -69,21 +69,21 @@ void TracksListModel::load()
...
 
     for (const Track& track : tracks) {
         m_trackList.push_back(buildTrackItem(track));
     }
 
     onSelectedTracks(selectionController()->selectedTracks());
 
-    prj->tracksChanged().onReceive(this, [this](std::vector<au::trackedit::Track> tracks) {
-        onTracksChanged(tracks);
+    prj->tracksChanged().onReceive(this, [this](std::vector<au::trackedit::Track> tr) {
```
saintmatthieu, line 85: I realized lately that, in situation like these, you could do `const std::vector<au::trackedit::Track>&`, even if the channel is declared as `Channel<std::vector...>`.

saintmatthieu (approved):

src/projectscene/view/trackspanel/trackslistmodel.cpp
IliasBergstromMuse: I did try using a const reference for the callback arguments - not here but on the non-boyscout branch, and I got a crash due to the object having gone out of scope.

I will definitely try what you suggest, and indeed on the receiving end there may be no repercussions - but I am wary of doing it on this branch, and that particular line, because it's beyond the scope of a quick "boyscout cleanup" that might address warnings, but won't risk having side-effects.

saintmatthieu: Indeed, wouldn't be the right place to try. Maybe on a new callback introduced by this branch, though.

saintmatthieu: I meant on the real branch.

PR #8417, opened 2025-03-10
Fix cursor shape under clip data selection

### LINKED ISSUE #8146: Wrong selection cursor behavior

embarc-gabriel:
Resolves: #8146 

Added two properties on selectioncontroller (selectionInProgress and selectionEditInProgress)
We use it to disable other mouseAreas during selection to avoid cursor change on hovering.
Apart from that, we ensure the correct cursor while editing the selection area and after a selection is done.

To QA:
- [ ]  IBeam cursor during the whole selection process.
- [ ] SizeHor cursor during selection edit
- [ ] After a selection is done the cursor should continue on IBeam

---

Recommended:

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -15,7 +15,6 @@ Rectangle {
     property bool clipHeaderHovered: false
     property var hoveredClipKey: null
     property bool tracksHovered: false
-    property bool underSelection: false
     property bool altPressed: false
     property bool ctrlPressed: false
 
@@ -97,14 +96,6 @@ Rectangle {
     SelectionViewController {
         id: selectionController
         context: timeline.context
-
-        onSelectionStarted: {
-            underSelection = true
-        }
-
-        onSelectionEnded: {
-            underSelection = false
-        }
     }
 
     Component.onCompleted: {
@@ -274,6 +265,7 @@ Rectangle {
 
             preventStealing: true
             acceptedButtons: Qt.LeftButton | Qt.RightButton
+            cursorShape: Qt.IBeamCursor
 
             hoverEnabled: true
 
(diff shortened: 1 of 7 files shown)


src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -576,15 +569,6 @@ Rectangle {
                         })
                     }
                 }
-
-                HoverHandler {
-                    property bool isNeedSelectionCursor: !selectionController.selectionActive
-                    cursorShape: isNeedSelectionCursor ? Qt.IBeamCursor : Qt.ArrowCursor
-
-                    onHoveredChanged: {
-                        root.tracksHovered = hovered
```
grliszas14, line 585: `tracksHovered` is now always false. It was used to distinguish if mouse was hovering over tracks or empty space.
In both cases proper context menus should be called when right-clicked. 
Right now `canvasContextMenu` will be always called. The other context menu for tracks is not yet implemented so I don't mind leaving it as is but I'm wondering if this HoverHandler is messing with cursors? If yes, we'll need to think of other way to detect if mouse is hovering over tracks

grliszas14 (approved):

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
embarc-gabriel: Yes! It was messing cursors.
The problem is the HoverHandler has this cursorShape parameter. If hover on this area the cursor changes its shape. But it is not desired sometimes, for example when I am on the smooth tool moving from one clip to the other.
For sure, I can make it visible / invisible for example and maybe it could work.
But I have tried to reduce the number of MouseAreas because it was a little bit too much IMO.
I understand your point and you are right but I personally think it is better to remove now and add it back when needed on the existing mouse areas.

Issue #8421, opened 2025-03-10
Audio deleted when joining clips


LynnEditor:
### Bug description

When I join two audio clips together, the audio in the first clip is deleted or silenced. It looks like this is a previous issue that has returned with version 3.7.2. (see issue #2119: Data loss on joining clips)

### Steps to reproduce

1. Select two audio clips
2. Join them (CMD + J)

### Expected behavior

The two clips are joined and the audio remains preserved in both

### Actual behavior

The audio in the first clip is deleted. The audio in the second clip is preserved. 

### Audacity Version

Audacity 3.7.2

### Operating system

macOS

LWinterberg:
Cannot reproduce on Windows

DavidBailes:
Windows 11.
Steps to reproduce a problem with joining two clips (regression wrt 3.7.1):

1. Generate some audio, and split into two clips, eg:

![Image](https://github.com/user-attachments/assets/cfe7fbec-d0ca-4e26-af65-053904009de5)

2. Move the clips:

![Image](https://github.com/user-attachments/assets/5dbed85d-14a9-423f-88c6-699db7721628)

3. Make a time selection:

![Image](https://github.com/user-attachments/assets/9112dd71-a772-4615-8c79-082c4e1962fb)

4. Join:

![Image](https://github.com/user-attachments/assets/a5c094b6-c09f-4f4d-adbc-7cce2b180ba3)

DavidBailes:
And here's another example:

1. Generate some audio, and split into 3 clips, eg:

![Image](https://github.com/user-attachments/assets/b7c351d2-72cd-4f90-b293-be8c8e8a393c)

2. Delete the first clip, and move the the first of the remaining clips to the left:

![Image](https://github.com/user-attachments/assets/f68d7cc8-771d-43de-a276-65db9b5b604d)

3. Make a time selection:

![Image](https://github.com/user-attachments/assets/586e3b8c-d141-42dc-bcf3-2b44dde2f8b5)

4. Join:

![Image](https://github.com/user-attachments/assets/e18411df-065a-44b7-873e-13596b85875a)

LWinterberg:
Ah, move to the side first. Yes, that's it.

chinakov:
Tested some scenarios, all is well, thanks! Tested.

chinakov:
Re-tested in scope of https://github.com/audacity/audacity/pull/8423/checks PR.

rodrigo-l-alves:
[...]

PR #8423, opened 2025-03-10
Fix clip being incorrectly shifted after join/paste/apply effect operations (#8421)

kryksyh:
Fix clip being incorrectly shifted after join/paste/apply effect operations

Resolves: #8421 

Recommended:

### DIFF
--- a/libraries/lib-wave-track/WaveTrack.cpp
+++ b/libraries/lib-wave-track/WaveTrack.cpp
@@ -1860,7 +1860,7 @@ void WaveTrack::PasteWaveTrackAtSameTempo(
                // This branch only gets executed in `singleClipMode` - we've
                // already made sure that stretch ratios are equal, satisfying
                // `WaveClip::Paste`'s precondition.
-               assert(insideClip->GetStretchRatio() == pClip->GetStretchRatio());
+               assert(insideClip->HasEqualPitchAndSpeed(*pClip));
                // This too should follow from the assertion of the same number
                // of channels in the tracks, near the top
                assert(insideClip->NChannels() == pClip->NChannels());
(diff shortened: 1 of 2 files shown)


libraries/lib-wave-track/WaveTrack.cpp
```diff
@@ -2225,6 +2225,7 @@ void WaveTrack::Join(
       }
 
       // wxPrintf("Pasting at %.6f\n", t);
+      clip->SetSequenceStartTime(0);
```
saintmatthieu, line 2228: So this is now necessary because of the modifications we recently did in WaveClip::Paste.
There are other places in WaveTrack.cpp where WaveClip::Paste is used, e.g. WaveTrack::MergeClips. Do these need be reviewed, too?

LWinterberg:
I tried various combinations with joining, cutting and pasting clips between one as well as different tracks, and they do seem to stay in place now.

saintmatthieu:
This PR fixes the reported bug, but looking at the change I noticed that `WaveTrack::MergeClips` might also be vulnerable to that.
Indeed: perform the steps of [this comment](https://github.com/audacity/audacity/issues/8421#issuecomment-2711471363), but as last step apply an EQ filter
image (5)
image (6)

saintmatthieu:
> This isn't looking good.

I wouldn't be so dramatic: I expect all these to result in a call to `WaveTrack::MergeClips`, so a fix there may well fix all these.

DavidBailes:
> > This isn't looking good.
> 
> I wouldn't be so dramatic: I expect all these to result in a call to `WaveTrack::MergeClips`, so a fix there may well fix all these.

I meant that it isn't looking good for users who have updated to 3.7.2. I personally wouldn't advise anyone to update to 3.7.2

LWinterberg:
we have paused the update notification, fwiw.

saintmatthieu (approved):

PR #8431, opened 2025-03-12
Implement disjoin command (split clips at silences)

### LINKED ISSUE #8404: Implement Split Clips on Silences
### Description

Implement Detach at silences command

embarc-gabriel:
Resolves: #8404 

Implement split clips on silence command

The command can be triggered by the shortcut Ctrl + Alt + j or on the menu (Edit -> Clip -> Split clips at silences)
The operation will also be added to the history.

Recommended:

---

QA

- [x] Trigger the command using the menu
- [x] Trigger the command using the shortcut
- [x] Time range
- [x] Single clip
- [x] Multi clip

### DIFF
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -1502,6 +1502,25 @@ NeedsDownmixing Au3Interaction::moveSelectedClipsUpOrDown(int offset)
     return needsDownmixing;
 }
 
+bool Au3Interaction::splitClipsAtSilences(const TrackIdList& tracksIds, secs_t begin, secs_t end)
+{
+    for (const auto& trackId : tracksIds) {
+        Au3WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
+        IF_ASSERT_FAILED(waveTrack) {
+            continue;
+        }
+
+        waveTrack->Disjoin(begin, end);
+
+        trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+        prj->notifyAboutTrackChanged(DomConverter::track(waveTrack));
+    }
+
+    projectHistory()->pushHistoryState("Split clips at silence", "Split at silence");
+
+    return true;
+}
+
 bool Au3Interaction::mergeSelectedOnTrack(const TrackId trackId, secs_t begin, secs_t end)
 {
     Au3WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -1502,7 +1502,7 @@ NeedsDownmixing Au3Interaction::moveSelectedClipsUpOrDown(int offset)
     return needsDownmixing;
 }
 
-bool Au3Interaction::splitClipsAtSilences(const TrackIdList& tracksIds, secs_t begin, secs_t end)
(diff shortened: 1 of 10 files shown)


embarc-gabriel:
@grliszas14 @IliasBergstromMuse 
https://github.com/user-attachments/assets/f04b0dea-34cd-4955-bb5c-a7f4b4c12020

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -338,6 +343,16 @@ void TrackeditActionsController::doGlobalJoin()
                            ActionData::make_arg3<TrackIdList, secs_t, secs_t>(selectedTracks, selectedStartTime, selectedEndTime));
 }
 
+void TrackeditActionsController::doGlobalDisjoin()
```
grliszas14, line 346: `doGlobalSomething()` functions only determine what specialized action should be called ( should it be called for single clip / multiple clips / range selection) and then call dispatcher.
If disjoin makes sense for range selection only then shortcut and UI actions should be calling SPLIT_CLIPS_AT_SILENCES directly and you can safely get rid of doGlobalDisjoin()

grliszas14: we discussed that with Gabriel - doGlobalDisjoin will stay and it's going to handle other cases as well single clip /multi clip disjoin

grliszas14 (approved):

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1361,6 +1361,49 @@ bool Au3Interaction::splitTracksAt(const TrackIdList& tracksIds, secs_t pivot)
...
+            continue;
+        }
+
+        waveTrack->Disjoin(begin, end);
+
+        trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+        prj->notifyAboutTrackChanged(DomConverter::track(waveTrack));
+    }
+
+    projectHistory()->pushHistoryState("Split clips at silence", "Split at silence");
```
IliasBergstromMuse, line 1378: This is a point throughout the codebase more than only in this PR - but as I see it here too I ask:

Usually I never have "raw" strings in a codebase like this - they are always in constexprs, either at the top of the file in an anonymous namespace if they're only used locally, or in a shared namespace if they're used in more .h/.cpp files.

Is there a reason you don't do that in Audacity?

IliasBergstromMuse (approved):
This PR's work reads well and works as advertised!
Normally I'd have insisted that your added methods are unit-tested, but really all they do is call other methods. You could unit-test those instead if you like, where it makes sense.

As for moving the strings out of the methods into constexprs, do it if there's no reason not to I'd say! I know it's pervasive in the codebase, but we'd have to start somewhere...

src/trackedit/internal/au3/au3interaction.cpp
embarc-gabriel: I usually prefer the constexprs closer to where it is used instead of the top of the file for example, once this will be on my editor "working memory". But this is just my personal opinion.
Apart from that, I have always tried to keep consistency within the codebase. I mean, I don´t like the idea of similar things being coded in different ways. I think we should keep he code as much uniform as possible so we reduce the mental overhead and only focus on the logic for the current task.
In this particular case the whole file has the same pattern so I just kept it.
Another example, I prefer to map, filter, reduce operations over for loops. But this is not how the codebase is now, so I keep doing the later.
IMO we should add more information on the code guidelines. It is painful at first but later on we will just not think to much about this because it will became our nature.

IliasBergstromMuse: An interesting discussion indeed! That I think I should move to slack since it doesn't apply only to this PR, see you on that point over there :)

Issue #8369, opened 2025-03-04
implement playback/recording prefs


grliszas14:
### Description

https://www.figma.com/design/Gt4wJqWpinqjk6c1qE125O/19---Preferences?node-id=2210-4729&m=dev

grliszas14:
QA bulletpoints:
* "move cursor along the timeline during playback" - you can set the preference but it doesn't have an effect as the functionality isn't implemented yet 
* you can select solo behavior and playback quality
* soloing tracks while pressing shift key does the opposite of solo behavior setting
* there's a change of behaviour compared to AU3 when soloing tracks, see the design

dozzzzer:
Tested

PR #8435, opened 2025-03-13
Playback / recording preferences

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8369

### DIFF
--- a/src/appshell/qml/Preferences/internal/CursorSection.qml
+++ b/src/appshell/qml/Preferences/internal/CursorSection.qml
@@ -1,0 +1,59 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+import QtQuick 2.15
+
+import Muse.Ui 1.0
+import Muse.UiComponents 1.0
+import Audacity.Preferences
+
+BaseSection {
+    id: root
+
+    title: qsTrc("appshell/preferences", "Move cursor along the timeline during playback")
+    spacing: 16
+
+    property var playbackPreferencesModel: null
+
+    navigation.direction: NavigationPanel.Both
+
+    IncrementalPropertyControlWithTitle {
+        title: qsTrc("appshell/preferences", "Short skip")
+
+        currentValue: playbackPreferencesModel.shortSkip
+
+        columnWidth: root.columnWidth
+        controlWidth: 120
+        spacing: root.columnSpacing
+
+        measureUnitsSymbol: qsTrc("global", " seconds")
+
+        navigation.name: "ShortSkipControl"
+        navigation.panel: root.navigation
+        navigation.row: 0
+
+        onValueEdited: function(newValue) {
+            playbackPreferencesModel.setShortSkip(newValue)
+        }
+    }
+
+    IncrementalPropertyControlWithTitle {
+        title: qsTrc("appshell/preferences", "Long skip")
+
+        currentValue: playbackPreferencesModel.longSkip
+
+        columnWidth: root.columnWidth
+        controlWidth: 120
+        spacing: root.columnSpacing
+
+        measureUnitsSymbol: qsTrc("global", " seconds")
+
+        navigation.name: "LongSkipControl"
+        navigation.panel: root.navigation
+        navigation.row: 1
+
+        onValueEdited: function(newValue) {
+            playbackPreferencesModel.setLongSkip(newValue)
+        }
+    }
+}
--- a/src/appshell/view/preferences/playbackpreferencesmodel.cpp
+++ b/src/appshell/view/preferences/playbackpreferencesmodel.cpp
@@ -1,185 +1,116 @@
 /*
- * SPDX-License-Identifier: GPL-3.0-only
- * MuseScore-CLA-applies
- *
- * MuseScore
- * Music Composition & Notation
- *
- * Copyright (C) 2021 MuseScore BVBA and others
- *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License version 3 as
- * published by the Free Software Foundation.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * Audacity: A Digital Audio Editor
  */
-
-#include "playbackpreferencesmodel.h"
 
 #include "log.h"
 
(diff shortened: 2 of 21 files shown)


src/playback/internal/playbackconfiguration.cpp
```diff
@@ -1,12 +1,153 @@
 /*
 * Audacity: A Digital Audio Editor
 */
+
+#include "settings.h"
+
+#include "Resample.cpp" // !!
```
grliszas14, line 7: This is a mystery to me, I'm trying to find out why I need to include .cpp file instead of .h file to make it work

IliasBergstromMuse: Indeed that should never be needed, what error do you get with the .h file?

IliasBergstromMuse: I just tested, it's solved easily enough by adding these:

#include "Prefs.h"
#include "TranslatableString.h"

:)

grliszas14: Thanks! I'm definitely not in shape today 😅

src/appshell/qml/Preferences/internal/CursorSection.qml
```diff
@@ -0,0 +1,86 @@
...
+    PlaybackStateModel {
+        id: playbackState
+    }
+
+    IncrementalPropertyControlWithTitle {
+        title: qsTrc("appshell/preferences", "Short skip")
+
+        currentValue: playbackModel.shortSkip
+
+        enabled: !playbackState.isPlaying()
```
saintmatthieu, line 48: Here as well: should probably not only be `!isPlaying`, but also `!isPaused`.
Looks like something should be done about this, as this particular state is needed in several places.

src/appshell/qml/Preferences/internal/CursorSection.qml
```diff
@@ -0,0 +1,86 @@
...
+import Audacity.Playback 1.0
+import Audacity.Preferences
+
+BaseSection {
+    id: root
+
+    title: qsTrc("appshell/preferences", "Move cursor along the timeline during playback")
+    spacing: 16
+
+    property var playbackModel: null
```
saintmatthieu, line 35: I'd rather the full name `playbackPreferenceModel` ; this name confused me.

src/playback/iplaybackconfiguration.h
```diff
@@ -16,5 +19,27 @@ class IPlaybackConfiguration : MODULE_EXPORT_INTERFACE
     virtual ~IPlaybackConfiguration() = default;
 
     virtual muse::draw::Color playColor() const = 0;
+
+    virtual std::vector<std::string> playbackQualityList() const = 0;
```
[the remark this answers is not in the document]
grliszas14: Yes, that's exactly the reason. I thought about moving it closer to the UI too but it's all in AU3 code and we keep it separate to AU4 code - which means we don't want it in the model or qml code. We will re-visit this code anyway when we have transifex setup ready, so it will be good opportunity to change that then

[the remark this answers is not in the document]
saintmatthieu: > we keep it separate to AU4 code

Sure, but for that we could have an AU4 enum defined in `playbacktypes.h` and convert AU3 to AU4 enum values in playbackconfiguration.cpp

src/appshell/view/preferences/playbackpreferencesmodel.h
```diff
@@ -27,86 +9,62 @@
 #include "modularity/ioc.h"
 #include "async/asyncable.h"
 #include "audio/iaudioconfiguration.h"
-// #include "midi/imidiconfiguration.h"
-// #include "midi/imidioutport.h"
-// #include "midi/imidiinport.h"
 #include "playback/iaudiodevicesprovider.h"
 #include "playback/iplaybackconfiguration.h"
 
+using namespace au::playback;
```
saintmatthieu, line 15: `using namespace` in a header ??

src/playback/internal/au3/au3trackplaybackcontrol.cpp
```diff
@@ -73,10 +74,36 @@ void Au3TrackPlaybackControl::setSolo(long trackId, bool solo)
...
+            currentSoloBehavior = TracksBehaviors::SoloBehavior::SoloBehaviorMulti;
+        }
+    }
+
+    if (currentSoloBehavior == TracksBehaviors::SoloBehavior::SoloBehaviorMulti) {
+        track->SetSolo(solo);
+    } else {
+        for (auto playable : tracks.Any<PlayableTrack>()) {
+            if (playable->GetId() == trackId) {
+                playable->SetSolo(!wasSolo);
```
saintmatthieu, line 99: You don't need `wasSolo` anymore and could just use `solo`.

saintmatthieu (approved):
LGTM, except for that `using namespace` in a header, but solving this should be a no-brainer.

src/appshell/view/preferences/playbackpreferencesmodel.cpp
```diff
@@ -26,160 +8,89 @@
...
+{
+    playbackConfiguration()->setSoloBehavior(behavior);
+}
+
+double PlaybackPreferencesModel::shortSkip() const
+{
+    return playbackConfiguration()->shortSkip();
+}
+
+void PlaybackPreferencesModel::shortSkipSelected(double seconds)
```
Eism, line 83: please rename to setSmth

src/playback/internal/playbackconfiguration.cpp
```diff
@@ -1,12 +1,129 @@
...
+
+muse::async::Notification PlaybackConfiguration::soloBehaviorChanged() const
+{
+    return m_soloBehaviorChanged;
+}
+
+au::trackedit::secs_t PlaybackConfiguration::shortSkip() const
+{
+    double shortSkip;
+    gPrefs->Read(wxT("/AudioIO/SeekShortPeriod"), &shortSkip, 5.0);
```
Eism, line 96: We shouldn't use gPrefs in our configurations, instead of it you should write it to our settings
We registered Au3CommonSettings and if au3 code want to read this pref we will provide the value from our settings

src/playback/internal/playbackconfiguration.cpp
```diff
@@ -1,12 +1,129 @@
...
 */
+
+#include "settings.h"
+
+#include "Prefs.h"
+#include "TranslatableString.h"
+#include "Resample.h"
+#include "Dither.h"
+
 #include "playbackconfiguration.h"
```
Eism, line 12: please correct the order of includes

src/playback/internal/playbackconfiguration.cpp
Eism: same for FastMethodSetting

PR #8461, opened 2025-03-18
8183 vst3 realtime support

### LINKED ISSUE #8183: VST3 realtime support
### Description

The support of VST effects in the context of destructive effects was merged.
This is not yet functional with realtime effects, though.

One of the main "challenges" of this task is to get the `RealtimeEffectViewerDialog.qml` the possibility of having a VST viewer injected, or VST and built-in effect viewers must be unified. (Those are the options I see at the moment.)

Another problem: I crash AU4 if I selected e.g. MuseHub's Chorus as realtime effect. (Not sure if it's realtime-effect-specific.)

Also, a one-liner is needed to make VSTs recognized as realtime-capable effects:
Image
But this one-liner must only be merged after VSTs work as realtime effects, too ; better no realtime VSTs than broken VSTs.

saintmatthieu:
Resolves: #8183

Recommended:

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -63,6 +63,8 @@
 #endif
 #ifdef AU_MODULE_EFFECTS_VST
 #include "effects/vst/vsteffectsmodule.h"
+#else
+#include "stubs/vst/vsteffectsstubmodule.h"
 #endif
 #include "effects/nyquist/nyquisteffectsmodule.h"
 #include "importexport/import/importermodule.h"
@@ -160,6 +162,8 @@ int main(int argc, char** argv)
 #endif
 #ifdef AU_MODULE_EFFECTS_VST
     app.addModule(new au::effects::VstEffectsModule());
+#else
+    app.addModule(new au::effects::VstEffectsStubModule());
 #endif
     app.addModule(new au::effects::NyquistEffectsModule());
     app.addModule(new au::importexport::ImporterModule());
--- a/src/effects/effects_base/view/effectviewloader.cpp
+++ b/src/effects/effects_base/view/effectviewloader.cpp
@@ -20,10 +20,14 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 #include "effectviewloader.h"
+#include "effectstypes.h"
 
 #include <QQmlEngine>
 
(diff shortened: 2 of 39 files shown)


src/effects/effects_base/view/effectviewloader.cpp
```diff
@@ -33,9 +37,20 @@ EffectViewLoader::EffectViewLoader(QObject* parent)
     : QObject(parent)
 {}
 
-void EffectViewLoader::load(const QString& type, const QString& instanceId, const QString& effectState, QObject* itemParent)
+void EffectViewLoader::load(const QString& instanceId, QObject* itemParent)
 {
-    LOGD() << "type: " << type << ", instanceId: " << instanceId;
+    LOGD() << "instanceId: " << instanceId;
+
+    // TODO: could instancesRegister have a `typeByInstanceId` method?
```
IliasBergstromMuse, line 44: Minor point: Generally I only merge TODO's if they correspond clearly to a ticket which directly links to them, otherwise they're easily forgotten.

src/effects/vst3-realtime-support.md
IliasBergstromMuse, line 1: A suggestion - it is common that .md docs on design and architecture among other things, are organized in /docs in the repo. Maybe move this to /docs, and we can start collecting all things like this there?

IliasBergstromMuse:
Thanks also for explaining this!

It reads well and the design is sound. Might be worth reducing the .qml duplication for the two different effect dialog types since they're largely identical - as we discussed, in a future PR, not necessarily this one!

IliasBergstromMuse (approved):

src/app/main.cpp
```diff
@@ -160,6 +162,8 @@ int main(int argc, char** argv)
 #endif
 #ifdef AU_MODULE_EFFECTS_VST
     app.addModule(new au::effects::VstEffectsModule());
+#else
+    app.addModule(new au::effects::VstEffectsStubModule());
```
Eism, line 166: usually, we use the same name for stubs, this will simplify the code a little

saintmatthieu: Didn't think of this. From your experience, doesn't it make code navigation sometimes tricky, though ?

PR #8478, opened 2025-03-21
Add a readme for src/

LWinterberg:
Just to get this started


IliasBergstromMuse:
A great initiative!

There's a standard "structure" used in source repos, where there's no README expected in the /src folder (or /include etc). 
It's either in the root or a /docs folder for e.g. documenting the codebase, so I'd suggest moving it there!

A good example is this https://github.com/grpc/grpc/ - another this: https://github.com/elk-audio/sushi

The only exception that I've seen is if there's modules included as sub-items, which can vary (and may even be git submodules), so they have their own README's.

You could call it DEVELOPMENT.md too and we could flesh it out and link to it from the main README.md

PR #8479, opened 2025-03-21
Tooltips

LWinterberg:
Rebased version of #8266

This adds explanatory strings to various (but not all) items in the app. I don't know when I'll get around continuing this work but I think what I have may be mergeable and built upon at a later date.


src/project/internal/projectuiactions.cpp
```diff
@@ -476,12 +476,6 @@ const UiActionList ProjectUiActions::m_actions = {
              TranslatableString("action", "Plugin manager"),
              TranslatableString("action", "Plugin manager")
              ),
-    UiAction("generate-omitted",
```
saintmatthieu, line 479: `AppMenuModel::makeGenerateMenu()` still constructs a menu item with that action code ; should it be removed?

src/project/internal/projectuiactions.cpp
```diff
@@ -540,12 +510,6 @@ const UiActionList ProjectUiActions::m_actions = {
              TranslatableString("action", "Plot spectrum"),
              TranslatableString("action", "Plot spectrum")
              ),
-    UiAction("analyzer-omitted",
```
saintmatthieu, line 543: Same here.
The corresponding menu's behaviour is a bit jank ; maybe that's why ?
image

src/project/internal/projectuiactions.cpp
```diff
@@ -18,118 +18,118 @@ const UiActionList ProjectUiActions::m_actions = {
     UiAction("file-new",
              au::context::UiCtxAny,
              au::context::CTX_ANY,
-             TranslatableString("action", "&New…"),
-             TranslatableString("action", "New…")
+             TranslatableString("action", "&New project"),
+             TranslatableString("action", "Open a new window to create a new project")
```
saintmatthieu, line 22: Not that I think it hurts to have this description set here, but does it ever manifest as a tooltip ? These descriptions in the `playbackuiactions.cpp` actions do, but they have UI element counterparts (e.g. the play button), which I guess is why it works for them.

src/effects/builtin/reverb/ReverbView.qml
```diff
@@ -39,6 +42,18 @@ EffectBase {
         anchors.right: parent.right
         anchors.margins: 8
 
+        MouseArea {
+            anchors.fill: parent
+            hoverEnabled: true
+            onContainsMouseChanged: {
+                if (containsMouse) {
```
saintmatthieu, line 49: Do we want to delay the summoning of the tooltips ? Right now they appear as soon as the mouse pointer enters their area, it feels a bit intrusive.

saintmatthieu (commented):
Looks good, except I think that you should make sure that there should not remain menu items with action codes that don't exist anymore. I suspect it to be the reason for erratic behavior of the menus in question (I even got a crash, in debug).

saintmatthieu:
Codestyle is breaking. I see the reverb view model cpp file where empty spaces were changed to tabs, which is probably why.
Let me know if I can help you setup Uncrustify.

src/effects/builtin/reverb/ReverbView.qml
```diff
@@ -97,13 +113,14 @@ EffectBase {
 
     CheckBox {
         id: wetOnly
+        text: qsTrc("effects", "Wet Only")
```
Eism, line 116: please revert this change

Issue #8436, opened 2025-03-13
implement editing prefs


grliszas14:
### Description

Design: https://www.figma.com/design/Gt4wJqWpinqjk6c1qE125O/19---Preferences?node-id=2210-5174&m=dev

QA:
- [x] Introduced „Apply effects to all audio when no selection is made” - this should select the entire project's audio
- [x] Introduced „Asymmetric stereo heights”, if "workspace dependent" setting is chosen, the list of workspaces appear. The list is dynamic and adjusts to the current workspaces list.
- [x] Introduced „Always paste audio as a new clip” setting - but this is just a setting that is not connected yet to the actual paste function (so you can test if it saves its state only)

There's one scenario in which you can crash the app (but its root is in the framework so it will be handled separately):
1. Add new workspace
2. go to preferences, enable asymmetric stereo heights for that new workspace, click OK
3. change stereo height for clip
4. go to -> edit workspaces -> remove current workspace
5. current workspace should probably jump back to "Classic", just click OK -> crash

dozzzzer:
@grliszas14 

- If 'Select all audio' option is enabled, when generating stuff without making a selection in a project with audio, all audio becomes selected, which makes the default `Duration` of the generator pre-filled with the total length of the audio, which is wrong. 'Select all audio' should only kick in for effects, methinks.

Master:

https://github.com/user-attachments/assets/b81b2af1-969f-4b52-92f1-43cd9ca7bbb6

This PR:

https://github.com/user-attachments/assets/a4c5d3de-1f48-4579-a4c1-2b43f8d68c72

PR #8480, opened 2025-03-21
Edit preferences

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8436


src/appshell/qml/Preferences/EditPreferencesPage.qml
```diff
@@ -22,16 +25,45 @@ PreferencesPage {
...
+            navigation.order: root.navigationOrderStart + 1
+        }
+
+        SeparatorLine { }
+
+        PastingBehaviorSection {
+            editPreferencesModel: editPreferencesModel
+
+            navigation.section: root.navigationSection
+            navigation.order: root.navigationOrderStart + 3 // skip one as there's navi for workspaces
```
Eism, line 50: add navigationOrderEnd (alias to last navigation control) to AsymmetricStereoHeightsSection and reuse it here

src/appshell/qml/Preferences/internal/AsymmetricStereoHeightsSection.qml
```diff
@@ -0,0 +1,128 @@
...
+
+                checked: editPreferencesModel.stereoHeightsPref == AsymmetricStereoHeights.WORKSPACE_DEPENDENT
+                text: qsTrc("appshell/preferences", "Depending on workspace")
+
+                navigation.name: "WorkspaceDependentRadioBtn"
+                navigation.panel: root.navigation
+                navigation.row: 1
+
+                onToggled: {
+                    editPreferencesModel.setStereoHeightsPref(AsymmetricStereoHeights.WORKSPACE_DEPENDENT)
```
Eism, line 97: please remove this empty line

src/appshell/view/preferences/editpreferencesmodel.h
```diff
@@ -1,29 +1,66 @@
...
+ * Audacity: A Digital Audio Editor
+ */
 #pragma once
 
-#include "trackedit/itrackeditconfiguration.h"
+#include <QObject>
+
 #include "modularity/ioc.h"
+
 #include "async/asyncable.h"
```
Eism, line 10: Wrong headers order

src/projectscene/internal/projectsceneconfiguration.cpp
```diff
@@ -156,3 +170,87 @@ muse::async::Channel<ClipStyles::Style> ProjectSceneConfiguration::clipStyleChan
...
+    return m_asymmetricStereoHeightsChanged;
+}
+
+std::vector<std::string> ProjectSceneConfiguration::asymmetricStereoHeightsWorkspaces() const
+{
+    // workspaces that have asymmetricStereoHeights enabled are stored as a string in muse::settings
+    // "Classic|Music|Advanced audio editing"
+    std::vector<std::string> result;
+    std::string combinedString = muse::settings()->value(ASYMMETRIC_STEREO_HEIGHTS_WORKSPACES).toString();
+    std::stringstream ss(combinedString);
```
Eism, line 200: you can reuse split from stringutils

Eism: and join too

src/projectscene/internal/projectsceneconfiguration.cpp
```diff
@@ -156,3 +170,87 @@ muse::async::Channel<ClipStyles::Style> ProjectSceneConfiguration::clipStyleChan
...
+muse::async::Notification ProjectSceneConfiguration::asymmetricStereoHeightsWorkspacesChanged() const
+{
+    return m_asymmetricStereoHeightsWorkspacesChanged;
+}
+
+void ProjectSceneConfiguration::asymmetricStereoHeightWorkspacesCleanUp()
+{
+    // cleanup after user removes workspace
+    std::vector<std::string> asymmetricWorkspaces = asymmetricStereoHeightsWorkspaces();
+    auto workspaces = workspaceManager()->workspaces();
```
Eism, line 239: Please don't use managers/services/etc in configurations
I think we can fill the list of workspaces directly in the model, yes?

src/effects/effects_base/internal/effectexecutionscenario.cpp
```diff
@@ -109,6 +109,15 @@ muse::Ret EffectExecutionScenario::doPerformEffect(au3::Au3Project& project, con
         }
 
         isSelection = t1 > t0;
+
+        if (!isSelection && trackeditConfiguration()->applyEffectToAllAudio()) {
```
Eism, line 113: It looks like applyEffectToAllAudio should be in effectsConfiguration

src/appshell/view/preferences/editpreferencesmodel.h
grliszas14: Can you please specify what order is right? If the current one is wrong I think that order presented here: https://github.com/musescore/muse_framework/wiki/CodeGuidelines is ambiguous

src/appshell/qml/Preferences/EditPreferencesPage.qml
```diff
@@ -22,16 +25,45 @@ PreferencesPage {
...
+        SeparatorLine { }
+
         MonoStereoConversionSection {
             id: monoStereoConversionSection
+
             askBeforeConverting: editPreferencesModel.askBeforeConvertingToMonoOrStereo
+
             navigation.section: root.navigationSection
-            navigation.order: root.navigationOrderStart
+            navigation.order: root.navigationOrderStart + 4
```
Eism, line 62: A proposal that will simplify the solution and we will not have to think about what number to write here and will also work if we decide to add another control to the section above:
It is better to use a chain for navigation - each next section uses the navigation of the previous section.

You can add navigationOrderEnd to BaseSection and reuse it wherever necessary

src/appshell/view/preferences/editpreferencesmodel.h
Eism: What's wrong with your order:
```
#include <QObject> // good

#include "modularity/ioc.h" // wrong, it is "ioc include" should be after "framework includes"

#include "async/asyncable.h" // wrong, it is "framework includes" should be after qt/system includes
#include "workspace/iworkspacemanager.h" // wrong, it is "framework includes" that can't work without ioc include, so it should be in ioc group

#include "effects/effects_base/ieffectsconfiguration.h" // wrong, it should be in ioc group
#include "projectscene/iprojectsceneconfiguration.h" // wrong, it should be in ioc group
#include "trackedit/itrackeditconfiguration.h" // wrong, it should be in ioc group
```

The correct order that matches the codestyle is:
```
#include <QObject>

#include "async/asyncable.h"

#include "modularity/ioc.h"
#include "workspace/iworkspacemanager.h"
#include "effects/effects_base/ieffectsconfiguration.h"
#include "projectscene/iprojectsceneconfiguration.h"
#include "trackedit/itrackeditconfiguration.h"
```

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
```diff
@@ -5,6 +5,7 @@ import Muse.Ui
 import Muse.UiComponents
 import Muse.GraphicalEffects
 
+import Audacity.Preferences
```
Eism, line 8: There is a dependency on the appshell module in projectscene, this is not allowed. appshell knows about all modules, but modules should not know about appshell.

If you need some setting from the preferences model, put it in the projectscene(?) configurator and reuse it in some clip model.

Eism (approved):

kryksyh:
@grliszas14 currently `projectscene/asymmetricStereoHeightsWorkspaces` is saved as an app setting, with a list of workspaces that should have it enabled. I have two concerns with this:

- is it translation-aware? Meaning that "Classic" or "Music" workspace would be localized, and changing the app language could effectively "reset" this setting
- Currently, it does handle the case when the user renames a custom workspace:
  - Create a workspace `test1`
  - Enable `asymmetricStereoHeightsWorkspaces` for this workspace
  - Rename the workspace to `test2`
  - Toggle the workspace
  - `asymmetricStereoHeightsWorkspaces` is no longer set for the workspace

grliszas14:
@kryksyh 

> is it translation-aware
Not sure, I think not (@Eism can you confirm that?). Those default workspaces are saved as a files (Music.mws, Classic.mws) so even if it is translated I don't believe it's done anywhere else than just before displaying it in the UI

> Currently, it doesn't handle the case when the user renames a custom workspace
We can fix that via adding workspaceChanged() handler, that's the only thing that will work at this point because:
* workspaces don't have IDs yet (just indexes) so first we'd need to do some work on framework side
* we can't save it as a workspace preference (I've considered it) because you can only modify currently opened workspace and we need to be able to modify every possible workspace from within settings even if you're in the home screen when no workspace is active

Anyway I'd create a follow-up ticket on that if you don't mind?

kryksyh:
> Anyway I'd create a follow-up ticket on that if you don't mind?

sure, we just have to keep it in mind

kryksyh:
> Those default workspaces are saved as a files (Music.mws, Classic.mws) so even if it is translated I don't believe it's done anywhere else than just before displaying it in the UI

I haven't looked into workspace implementation. I didn't realize that workspaces do not have names, and we use filenames instead. This would also imply that we could not use certain characters in the workspace name

image

grliszas14:
> This would also imply that we could not use certain characters in the workspace name

Yup, that's why I used a forbidden character as a separator in this setting

kryksyh:
There is also validation not working as expected, if you try to create `\ / TEST` workspace, it will silently fail
```
11:06:27.798 | ERROR | main_thread     | ZipWriter::ZipWriter | failed open file: /Users/dmitry/Library/Application Support/Audacity/Audacity4Development/workspaces/\ / TEST.mws
11:06:27.798 | ERROR | main_thread     | WorkspaceFile::save | failed write files to zip
11:06:27.815 | WARN  | main_thread     | WorkspaceManager::setupCurrentWorkspace | failed get workspace: \ / TEST, will use Classic
```

Issue #8472, opened 2025-03-20
AU4 occasionally crashes as ClipsListModel::moveSelectedClips (and others) are invoked after project deletion


IliasBergstromMuse:
AU4 occasionally crashes as ClipsListModel::moveSelectedClips (and others) are invoked after project deletion.

The suggested solution:

Cause/Events: 
when closing, the geometry changes, this forces the frame time to be recalculated, and changing the frame time leads to other changes:

Solution:
Unsubscribing from the change when the page is closed.
The page should probably have onClose, there we need to call some method timeContext onClose, set the "closing" flag there and check in the onResizeFrameWidth method, if the resize occurs after closing, then do nothing.

Alternatively:

```
void DockWindow::loadPage(const QString& uri, const QVariantMap& params)
{
...
    bool isFirstOpening = (m_currentPage == nullptr);

    if (!isFirstOpening) {
        savePageState(m_currentPage->objectName());
        clearRegistry();
        m_currentPage->setVisible(false);
        m_currentPage->deinit();
    }
...
```
i.e. visible = false is set, then we can do it like this
```
Timeline.qml
    onWidthChanged: {
        if (root.visible) {
            timelineContext.onResizeFrameWidth(root.width)
        }
    }

```
add check on visible

```
bool ClipsListModel::moveSelectedClips(const ClipKey& key, bool completed)
{
    ClipListItem* item = itemByKey(key.key);
    if (!item) {
        return false;
    }

--> auto vs = globalContext()->currentProject()->viewState();
    if (!vs) {
        return false;
    }
```

Stacktrace:

```
au::projectscene::ClipsListModel::moveSelectedClips(const au::projectscene::ClipKey &, bool) clipslistmodel.cpp:550
$_0::operator()() const clipslistmodel.cpp:565
au::projectscene::TimelineContext::updateFrameTime() timelinecontext.cpp:482
au::projectscene::TimelineContext::onResizeFrameWidth(double) timelinecontext.cpp:287
KDDockWidgets::QWidgetAdapter::geometryChange(const QRectF &, const QRectF &) QWidgetAdapter_quick.cpp:249
KDDockWidgets::QWidgetAdapter::geometryChange(const QRectF &, const QRectF &) QWidgetAdapter_quick.cpp:249
KDDockWidgets::QWidgetAdapter::geometryChange(const QRectF &, const QRectF &) QWidgetAdapter_quick.cpp:249
[...]

PR #8504, opened 2025-03-24
8472 avoid invisible timeline refresh and optimise history refresh

IliasBergstromMuse:
Resolves: https://github.com/audacity/audacity/issues/8472

This SHOULD resolve an occasional, hard to reproduce crash I used to see when closing audacity. Hard to test, but at least it shouldn't have any side effects.

I also piggy-backed an undo-redo to index commit which uses redraw optimisation here, both changes together are too simple to warrant dedicated PR's, they're one small commit each.

Recommended:

- [ ] Use Audacity track and clip editing, do some undo/redo, and exit the program. It should no longer crash on close.
- [ ] Now undo/redo index jumping in history panel should properly use change detection, and so there should be less flashing.

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/Timeline.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/Timeline.qml
@@ -14,9 +14,11 @@ Rectangle {
     color: ui.theme.backgroundSecondaryColor
 
     //! NOTE This element must be the same width as the track wave visible area.
-    //! If this is different, than appropriate changes must be made.
+    //! If this is different, then appropriate changes must be made.
     onWidthChanged: {
-        timelineContext.onResizeFrameWidth(root.width)
+        if (root.visible) {
+            timelineContext.onResizeFrameWidth(root.width)
+        }
     }
 
     function init() {
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -2331,14 +2331,19 @@ bool Au3Interaction::canRedo()
 
 bool Au3Interaction::undoRedoToIndex(size_t index)
 {
+    if (projectHistory()->currentStateIndex() == index) {
+        return false;
+    }
+
     auto trackeditProject = globalContext()->currentProject()->trackeditProject();
+
+    const TracksAndClips before = trackeditProject->buildTracksAndClips();
 
     projectHistory()->undoRedoToIndex(index);
 
-    // Redo removes all tracks from current state and
-    // inserts tracks from the previous state so we need
-    // to reload whole model
-    trackeditProject->reload();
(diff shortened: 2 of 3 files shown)


src/projectscene/qml/Audacity/ProjectScene/clipsview/Timeline.qml
```diff
@@ -14,9 +14,11 @@ Rectangle {
     color: ui.theme.backgroundSecondaryColor
 
     //! NOTE This element must be the same width as the track wave visible area.
-    //! If this is different, than appropriate changes must be made.
+    //! If this is different, then appropriate changes must be made.
     onWidthChanged: {
-        timelineContext.onResizeFrameWidth(root.width)
+        if (root.visible) {
+            timelineContext.onResizeFrameWidth(root.width)
+        }
```
IliasBergstromMuse, line 21: Now the notification is only triggered if it's visible.

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -2331,14 +2331,19 @@ bool Au3Interaction::canRedo()
 
 bool Au3Interaction::undoRedoToIndex(size_t index)
```
IliasBergstromMuse, line 2332: And this now uses the undo/redo change detection.

src/trackedit/internal/au3/au3projecthistory.cpp
```diff
@@ -109,10 +109,6 @@ void Au3ProjectHistory::markUnsaved()
 
 void Au3ProjectHistory::undoRedoToIndex(size_t index)
 {
-    if (currentStateIndex() == index) {
-        return;
-    }
-
```
IliasBergstromMuse, line 115: I move the check out. This method was not used anywhere else anyway, and in the calling context above, the check is actually used to avoid unnecessary refresh.

Eism: actually, we can call this method not from Au3Interaction, but from another place, and then we will send an extra notification about the change in history

IliasBergstromMuse: True, if it's called from elsewhere (which it isn't right now). I didn't see any harm when reading where that was subscribed to but I'm happy to reintroduce it.

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -2331,14 +2331,19 @@ bool Au3Interaction::canRedo()
...
 {
+    if (projectHistory()->currentStateIndex() == index) {
+        return false;
+    }
```
IliasBergstromMuse, line 2336: This duplicates a check in project history - but that's better than storing "before", altering undoRedoToIndex to return a "bool false" if it actually did an action, and only then deciding to return.

grliszas14 (approved):

PR #8508, opened 2025-03-24
Add unit tests for merge and split operations

### LINKED ISSUE #8488: Unit tests for merge / split operations
### Description

Add unit tests for merge / split operations

embarc-gabriel:
Resolves: #8488 

Add unit tests for merge split operations.
I have added a helper class to create tracks and clips easily.

Recommended:


src/trackedit/tests/au3interaction_tests.cpp
```diff
@@ -91,4 +300,160 @@ TEST_F(Au3InteractionTests, ChangeClipColor)
...
+
+    //! [THEN] The number of intervals is 1
+    ASSERT_EQ(track->NIntervals(), 1) << "Precondition failed: The number of intervals is not 1";
+
+    //! [WHEN] Disjoin outside the clip bondaries
+    m_au3Interaction->splitRangeSelectionAtSilences(
+        { track->GetId() }, TRACK1_CLIP_END, TRACK1_CLIP_END + 1.0);
+
+    //! [THEN] The number of intervals is still one
+    ASSERT_EQ(track->NIntervals(), 1) << "The number of intervals after the split range operation is not 1";
```
kryksyh, line 341: I would also check that the length was not changed, in case the clip was trimmed

src/trackedit/tests/au3interaction_tests.cpp
```diff
@@ -40,18 +170,79 @@ class Au3InteractionTests : public ::testing::Test
...
 
         initTestProject();
+        createSampleData();
     }
 
     void initTestProject()
     {
         m_au3ProjectAccessor = std::make_shared<au3::Au3ProjectAccessor>();
-        const muse::io::path_t TEST_PROJECT_PATH = muse::String::fromUtf8(trackedit_tests_DATA_ROOT) + "/data/test.aup3";
+        const muse::io::path_t TEST_PROJECT_PATH = muse::String::fromUtf8(trackedit_tests_DATA_ROOT) + "/data/empty.aup3";
```
kryksyh, line 179: @Eism @IliasBergstromMuse 
Sorry if we already had this discussion. Why do we need to provide the test data? Wouldn't it be nicer to create the project at runtime?

src/trackedit/tests/au3interaction_tests.cpp
embarc-gabriel: Once this is outside the clip boundaries the operation should only be ignored.
But I think you are right. We should do some sanity check because we have already seen some operation resulting in crazy outputs.
I will add it.

src/trackedit/tests/au3interaction_tests.cpp
```diff
@@ -21,15 +25,141 @@ using namespace au;
...
+constexpr static double TRACK4_CLIP_DURATION = TRACK4_SILENCE_DURATION + TRACK4_FIRST_SEGMENT_DURATION;
+constexpr static double TRACK4_CLIP_END = TRACK4_CLIP_START + TRACK4_CLIP_DURATION;
+
+constexpr static double TRACK5_CLIP_START = 0.0;
+constexpr static double TRACK5_FIRST_SEGMENT_DURATION = 10 * SAMPLE_INTERVAL;
+constexpr static double TRACK5_SILENCE_DURATION = 450 * SAMPLE_INTERVAL;
+constexpr static double TRACK5_CLIP_DURATION = TRACK5_FIRST_SEGMENT_DURATION + TRACK5_SILENCE_DURATION;
+constexpr static double TRACK5_CLIP_END = TRACK5_CLIP_START + TRACK5_CLIP_DURATION;
+
+static std::vector<float> createNoise(double duration, double sampleRate)
```
kryksyh, line 133: I don't have a strong opinion on that, but I would move these two to `template factory`

kryksyh (commented):
@embarc-gabriel this is awesome!

src/trackedit/tests/au3interaction_tests.cpp
embarc-gabriel: Yes! Probably this is a good idea. I will check how to do this.

IliasBergstromMuse: My initial idea was to create the project at runtime too, it's much better to not depend on files, and to be able to create the project programmatically. 
I don't remember why we did not do that, but this was also the first week I was looking at the Audacity code so I might have missed something.

One possible issue is that Audacity cannot work on a project that's fully in memory, right? it always needs to have it cached or stored, so if it's not opening a file, it will need access to it's local cache.

But yes @embarc-gabriel definitely spend the time to figure out how to make tests that don't depend on a file, for this PR or the next, it is necessary for when we start having many more tests. 

An ideal goal is to have very fast tests, that do not depend on external resources, and build and run locally on every local build too, which needs them to be super "lean".

src/trackedit/internal/au3/tracktemplatefactory.h
IliasBergstromMuse, line 1: Great that you've created these utilities!

src/trackedit/tests/au3interaction_tests.cpp
embarc-gabriel: Done!

src/trackedit/tests/au3interaction_tests.cpp
embarc-gabriel: Done! I have created the module and I have moved the functions to the template factory.

src/trackedit/internal/au3/tracktemplatefactory.h
embarc-gabriel: Thank you! I really hope this helps on adding more unit tests.

IliasBergstromMuse (approved):
Excellent!

Good that we now also have a testing utilities module - we should start populating it with some of the other files that may be used across tests, like environment.cpp -to be moved the first time it's needed I guess, not necessarily now.

The optimisation of not relying on a project file that we discussed, can be a future ticket/PR, but is still worth looking into.

src/trackedit/tests/au3interaction_tests.cpp
embarc-gabriel: I will open a new ticket on how to create the project at runtime.

Eism: I don't have a strict opinion on what is better to create in runtime or read from a file.

We need to consider all the pros and cons and choose what we need. At that time, I liked the idea of a file more because you can open it in the app, prepare test data, and visually see the test project. You can also save the result project after the test in a separate file and see if everything is done correctly (not just check the code).

However, if you find this file solution inconvenient, we may reconsider this solution.

IliasBergstromMuse: I think both are good to have!

It's very convenient, for the reasons you say. But if we need to read a file potentially 100's of times during the test execution, they risk becoming too slow for running on each local build, which is an ideal use of unit-tests. And for that use case it'd be good to be able to construct projects "in memory". If that still always requires using a physical file in the cache there's no performance improvement however, in which case really indeed the advantage is lost (@embarc-gabriel).

Eism:
As I see, AU_BUILD_TESTUTILITY_MODULE should not be enabled if MUSE_ENABLE_UNIT_TESTS is not enabled.
I also assume that testutility will become a module in the future that will contain different classes for different modules, i.e. a super module. I would suggest placing such utilities in their own modules (in the tests folder) and reusing them where needed.

embarc-gabriel:
@Eism do you mean replicate the testutility file inside the test folder for each module that needs it or place it inside a test folder a module and link this module to a second module that needs this test utility?
Apart from that, I can add this cmake condition on my next unit test PR.

src/trackedit/tests/au3interaction_tests.cpp
embarc-gabriel: As I comment to @IliasBergstromMuse the problem I see with files is that we must ensure a proper cleanup is performed after each test. Otherwise we could make it "dirty" and broke the next unit tests. I mean, this can somehow break the "independency" between tests.
But we can study other approaches in the future.
At least for now, we are increasing the coverage and this is good.

PR #8523, opened 2025-03-26
Split into new track

### LINKED ISSUE #8484: Split into new track doesn't work
1. Make a selection on a clip
2. Edit > Clip > Split into new track
3. Observe: The action has no effect

Image

embarc-gabriel:
Resolves: #8484 

This PR implements the split into new tracks functionality.
It works "just" as au3. We can split by time range selection and clip selection.
It does not split with only the playback cursor and if the time selection does not contain clip data.

Recommended:

-------
To QA:

- [x] Time range selection on a single and multiple tracks
- [x] Clip selection with a single or multiple clips
- [x] Time range selection without clip data should not create new tracks


grliszas14 (approved):

dozzzzer:
@embarc-gabriel 

- the name of the Undo/Redo entry should be "Split into new track":

image

PR #8524, opened 2025-03-26
Fix #8204: Playcursor flickers on zoom in/out while recording audio.

### LINKED ISSUE #8204: Playcursor flickers on zoom in/out
1. Start recording
2. Try to zoom in/out

Current behaviour:
Playcursor jumps to the beginning of the clip for a brief second. 

https://github.com/user-attachments/assets/239ff692-2fde-4587-9f5b-8ae05702a7fa

Expected behaviour:
Playcursor points at the end of current recording all the time

franciscofpereira:
Resolves: #8204

The onFrameTimeChanged() function was updating the play cursor position based on the playback position, even during recording. 

While recording, the playback position defaulted to the start of the audio file. This would cause the cursor to flicker when zooming in or out during recording.

I modified the onFrameTimeChanged() function to differentiate between playback and recording states. During recording, the play cursor position is now updated based on the actual recording position rather than the playback position.

Tested both in macOS and Linux.

 I have attached a before and after comparison.
 
## Before:

https://github.com/user-attachments/assets/150bc205-8d2f-44f6-9119-a52568acdf78

## After:
https://github.com/user-attachments/assets/d17ad9f2-a1cd-43eb-a232-363d6aa3dad3

Recommended:

### DIFF
--- a/src/projectscene/view/playcursor/playcursorcontroller.cpp
+++ b/src/projectscene/view/playcursor/playcursorcontroller.cpp
@@ -105,8 +105,17 @@ void PlayCursorController::updatePositionX(muse::secs_t secs)
 
 void PlayCursorController::onFrameTimeChanged()
 {
-    m_positionX = m_context->timeToPosition(playbackState()->playbackPosition());
-    emit positionXChanged();
+    double newPosition;
+    if (globalContext()->isRecording()) {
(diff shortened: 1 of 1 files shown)


embarc-gabriel (approved):

kryksyh:
@franciscofpereira, great job, thanks!

PR #8530, opened 2025-03-26
Only load needed plugins on project-open

### LINKED ISSUE #8473: Loading of VST3 from the hub can crash the app
### Description

Loading of VST3 from the hub can crash the app

### How to reproduce

1. Sign in to the hub,
2. Download and install a protected plugin, e.g. a free trial (if you don't have one such plugin already)

#### Audacity 4
3. Start AU4
4. New project
5. add plugin to master track
6. save project
7. quit AU4
8. sign out from the hub
9. **quit the hub**
10. start AU4

_CRASH_

#### MuseScore:
3. start MSS
4. New project with one staff (track)
5. add plugin to the track's effects
6. save project
7. quit MSS
8. sign out from the hub
9. **quit the hub**
10. start MSS
11. open the project

_CRASH_

#### Audacity 3:
3. Start AU3
4. New project
5. New track
6. Add the plugin as realtime-effect
7. Save project
8. Close AU3
9. Sign out from the hub,
10. **quit the hub**
11. Re-open the project

_CRASH_

12. Try to re-open Audacity. You are asked whether you want to recover the crashed projects. Say yes or now.

_CRASH_

#### Reaper
Similarly: new track, add Uberloud, save, close Reaper, sign out from hub, quit the hub, restart Reaper

_CRASH_

### Why MSS behaves better
MSS only crashes when opening a project that has a non-authorized plugin. This is because it _lazy-loads_ the plugins a project needs when opening this project. This can may project-opening time longer, but in this particular case it is preferable.

### Open questions
Is it only plugins controlled by MH that crash the app? What other plugin protection software is there (e.g. iLok)? Does this other software handle things more gracefully?

#### A: Lazy-loading
1. Like in MS, lazy-load plugins when opening project.

#### B: Disable VSTs if sign-in doesn't work
1. If signed out, ask the user to sign in
2. If user says no, disable VSTs

#### C: Disable MH VSTs if sign-in doesn't work
If we can implement detection and labeling of MH plugins:
1. If signed out and if the registry has MH plugins, ask the user to please sign in.
2. If user says no, skip loading of MH plugins and disable them

saintmatthieu:
Resolves: #8473

Partially solving the problem described in #8473 by postponing loading of effects when opening project, and only loading those needed by the project.

Recommended:

QA:
To test this, please first read the description of #8473 on how to reproduce.
With this PR, crashing should "only" occur when opening a project that contains a MH plugin while being signed out from the hub, not on app start-up anymore. Please contact me if there are unclarities.

### DIFF
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -108,6 +108,25 @@ EffectMeta EffectsProvider::meta(const EffectId& effectId) const
     return EffectMeta();
 }
 
+bool EffectsProvider::loadEffect(const EffectId& effectId) const
+{
+    const auto it = std::find_if(m_effects.begin(), m_effects.end(), [&](const EffectMeta& meta) {
+        return meta.id == effectId;
+    });
+    if (it == m_effects.end()) {
+        return false;
+    }
+    const auto isVst = it->categoryId == VST_CATEGORY_ID;
+    if (!isVst) {
+        // If an effect is not a VST and is in m_effects, then it's a built-in effect and it's loaded already.
+        return true;
+    }
+    IF_ASSERT_FAILED(vstEffectsRepository()) {
+        return false;
+    }
+    return vstEffectsRepository()->ensurePluginIsLoaded(effectId);
+}
+
 std::string EffectsProvider::effectName(const std::string& effectId) const
 {
     const auto desc = PluginManager::Get().GetPlugin(effectId);
@@ -145,6 +164,9 @@ bool EffectsProvider::supportsMultipleClipSelection(const EffectId& effectId) co
 
 Effect* EffectsProvider::effect(const EffectId& effectId) const
 {
+    if (!loadEffect(effectId)) {
+        return nullptr;
+    }
     PluginID pluginID = effectId.toStdString();
     const PluginDescriptor* plug = PluginManager::Get().GetPlugin(pluginID);
     if (!plug || !PluginManager::IsPluginAvailable(*plug)) {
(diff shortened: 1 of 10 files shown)


igorkorsukov (approved):

src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -108,6 +108,25 @@ EffectMeta EffectsProvider::meta(const EffectId& effectId) const
...
 
+bool EffectsProvider::loadEffect(const EffectId& effectId)
+{
+    const auto it = std::find_if(m_effects.begin(), m_effects.end(), [&](const EffectMeta& meta) {
+        return meta.id == effectId;
+    });
+    IF_ASSERT_FAILED(it != m_effects.end()) {
+        return false;
+    }
+    const auto isVst = it->categoryId == VST_CATEGORY_ID;
```
igorkorsukov, line 119: I think we need to think about how to do this better, maybe introduce a type (enumerator) or make this check as a function..

saintmatthieu: I totally agree. I was wondering why category was a string. Will address this on the occasion.

IliasBergstromMuse (approved):
Great! 

Nothing further to comment on the topic that hasn't been said on slack.

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: I think it's not very good to determinate a type by category. Category is mainly for order in the menu (or more generally, in the user's UI), I can imagine that we can somehow change categories, some thing like not just VST, but for example "verified VST", "popular VST" etc., and also probably categories can be custom, created by the user (if or when we make the functionality of the ability for the user to organize their plugins themselves).

saintmatthieu: Should we then add an enum to `EffectMeta`, and the effect repositories (atm IBuiltinEffectsRepository and IVstEffectsRepository) set these values when called `effectMetaList()` upon ?

igorkorsukov: yeah I think we should do something like that

saintmatthieu: Just pushed an extra commit.

igorkorsukov (approved):

Issue #8485, opened 2025-03-21
implement "always paste as a new clip"


grliszas14:
### Description

this should be enabled only for basic paste (cmd+v) and for single clip or range selection

QA:
- [x] introduced pasting into existing clip (based on preference in Editing tab)
- [x] this works only for: single clip or range selection
- [x] range selection can combine multiple tracks, but the rule is: inserting into existing clip will work only if selection on given track contains a single clip (or part of a clip)
- [x] please check tab + arrows navigation in Editing tab (if not checked already)
- [x] updated Preferences tabs (ordering and new tabs, as in the design: https://www.figma.com/design/Gt4wJqWpinqjk6c1qE125O/19---Preferences?node-id=2210-5174&m=dev

kryksyh:
> this should be enabled only for basic paste (cmd+v) and for single clip or range selection

Is this really a desired behavior? To me, it looks like there is an inconsistency between the `Paste` and `Insert` actions. I would expect that if I paste a clip into another clip it should obey the preference regardless if it is `paste` or `insert`.

chinakov:
> I would expect that if I paste a clip into another clip it should obey the preference regardless if it is paste or insert.


@Tantacrul - in this scenario, should 'insert' always paste the clip into the other clip without separating them regardless of the preference chosen by the User?

Tantacrul:
Following a discussion offline:

'Paste Insert' should essentially be thought of as 'ripple paste'. It should move clips regardless of the preference, which should only influence how regular Paste works.

In this way, a user can set up their workflow with the most amount of flexibility. @grliszas14 is already aware of this.

In addition, we agreed to add a binding for 'Paste & Preserve Sync', which is basically 'ripple paste' that affects all tracks.

PR #8531, opened 2025-03-26
Preference: paste as new / paste into existing clip

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8485

This PR is built on top of https://github.com/audacity/audacity/pull/8480, only 3 last commits are relevant

### DIFF
--- a/src/appshell/view/preferences/preferencesmodel.cpp
+++ b/src/appshell/view/preferences/preferencesmodel.cpp
@@ -152,7 +152,7 @@ void PreferencesModel::load(const QString& currentPageId)
         makeItem("general", QT_TRANSLATE_NOOP("appshell/preferences", "General"), IconCode::Code::SETTINGS_COG,
                  "Preferences/GeneralPreferencesPage.qml"),
 
-        makeItem("appearance", QT_TRANSLATE_NOOP("appshell/preferences", "Interface"), IconCode::Code::BRUSH,
+        makeItem("appearance", QT_TRANSLATE_NOOP("appshell/preferences", "Appearance"), IconCode::Code::BRUSH,
                  "Preferences/AppearancePreferencesPage.qml"),
 
         makeItem("audio-settings", QT_TRANSLATE_NOOP("appshell/preferences", "Audio settings"), IconCode::Code::AUDIO,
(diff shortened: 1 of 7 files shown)


src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1000,12 +1025,13 @@ muse::Ret Au3Interaction::pasteFromClipboard(secs_t begin, bool moveClips, bool
     }
 
     muse::Ret ok { muse::make_ok() };
+    bool pasteIntoExistingClip = !configuration()->pasteAsNewClip() && !moveAllTracks;
```
kryksyh, line 1028: I would prefer `Au3Interaction` not to depend on the configuration. It should only rely on arguments passed to the function. That way it would be easier to test and use in scripting.

kryksyh: as a quick and dirty fix we could add a simple struct with a few bools, or refactor the function (not sure if this is a good idea though)

Eism: Au3Interaction is a simple service, it is no different from our other services, so I see no limitations in using the configurator.

As for scripts, we can set up the configurator before using this service. For example, we already do this - parsing command line arguments (applyCommandLineOptions). When a user calls a program with a flag, we simply set this value in the configurator and then the service works with this data.

On the other hand, we stick to simple interfaces for this kind of services, with a minimum number of parameters, so it is easier to understand what the method does without delving into the code of this method.

kryksyh: >so it is easier to understand what the method does without delving into the code of this method.

I mean it is harder to understand what and why the method does if its behavior can change depending on the arguments and the configuration simultaneously.

src/appshell/view/preferences/preferencesmodel.cpp
```diff
@@ -167,14 +167,17 @@ void PreferencesModel::load(const QString& currentPageId)
...
+        makeItem("music", QT_TRANSLATE_NOOP("appshell/preferences", "Music"), IconCode::Code::METRONOME, ""),
 
         makeItem("cloud", QT_TRANSLATE_NOOP("appshell/preferences", "Cloud"), IconCode::Code::CLOUD, ""),
 
         makeItem("shortcuts", QT_TRANSLATE_NOOP("appshell/preferences", "Shortcuts"), IconCode::Code::SHORTCUTS,
                  "Preferences/ShortcutsPreferencesPage.qml"),
 
-        makeItem("plugin", QT_TRANSLATE_NOOP("appshell/preferences", "Plugin manager"), IconCode::Code::PLUGIN, "")
+        makeItem("advanced", QT_TRANSLATE_NOOP("appshell/preferences", "Advanced options"), IconCode::Code::SETTINGS_COG,
+                 "Preferences/ShortcutsPreferencesPage.qml"),
```
Eism, line 180: should be empty :)

src/trackedit/internal/au3/au3interaction.cpp
Eism: A classic example of a debate about simplicity of interfaces and predictability of code behavior.

Your approach is more related to functional programming, my approach is more related to object-oriented programming.

My opinion is that the second approach is more suitable for such large products as Audacity. The less code, the better. The case with scripts is a special case, we should not change all our approaches because of this limitation. Just because it can be called from scripts, we have to add a bunch of parameters, structures, check their validity, think about their default values? I think this is a complication.
I think it is easier to hide such settings inside the service and only provide externally those parameters that are really needed.

kryksyh (approved):

Issue #8486, opened 2025-03-21
For a range selection, options from the Edit > Clip menu should open on the right-click menu.


dozzzzer:
Current context menu options when invoked on a range selection:

<img width="474" alt="Image" src="https://github.com/user-attachments/assets/778ea98e-f496-4f4d-b6fa-d08f0d97ac4a" />

dozzzzer:
@LWinterberg Design is needed

DilsonsPickles:
@dozzzzer @LWinterberg Which options are we missing here?

Just list them out here and I'll create a spec.

dozzzzer:
I believe it's these four:

<img width="653" height="276" alt="Image" src="https://github.com/user-attachments/assets/1fe64e13-5551-487f-acc4-60f56f251853" />

DilsonsPickles:
Like this?

<img width="2684" height="1672" alt="Image" src="https://github.com/user-attachments/assets/d009a492-dc5c-4cfe-9cb6-7d516098b7ce" />

dozzzzer:
@DilsonsPickles yes, that's what I'd expect it to be. 
Just don't forget to change the second item from "Cut" to "Copy" :)

DilsonsPickles:
@dozzzzer fixed in Figma!

PR #8537, opened 2025-03-27
selection context menu

LWinterberg:
Resolves: #8486

Whatcha thinking @dozzzzer? @Tantacrul @DilsonsPickles 

image


Tantacrul:
Can you include a list of the changes? Are the shortcuts on the right new?

1. It does occur to me that it might make sense to include regular delete here too.
2. I think the '(per track)' addition is a bit wordy
3. That said, wouldn't it make sense to also include 'Delete & close gap (all tracks)'

dozzzzer:
@LWinterberg are we going to rename 'split into new track' to 'move selection to new track'?

PR #8538, opened 2025-03-27
Track and clips remove operations unit tests

### LINKED ISSUE #8491: Unit tests for remove operations
### Description

Add unit tests for remove operations.

embarc-gabriel:
Resolves: #8491 

- [x] Add unit test for remove track/clips operations.
- [x] Remove the testutility plugins add the helper classes to trackedit test folder.
- [x] Some small changes to remove warnings.  

Recommended:


IliasBergstromMuse:
I'm seeing a lot of warnings in the test output console for "Uninteresting mock function calls". I know they're not introduced with this PR, but since you're working on tests, how about addressing these while working on the file? I mean those in SetUp mainly.

`GMOCK WARNING:
Uninteresting mock function call - taking default action specified at:
/Users/iliasmuse/repos/audacity/src/trackedit/tests/au3interaction_tests.cpp:176:
    Function call: currentProject()
          Returns: (ptr = 0x138839418, value = 8-byte object <C8-67 56-05 01-00 00-00>)
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/docs/gmock_cook_book.md#knowing-when-to-expect for details.
`

src/trackedit/tests/au3interaction_tests.cpp
```diff
@@ -211,21 +232,38 @@ class Au3InteractionTests : public ::testing::Test
                       { TRACK5_SILENCE_DURATION, TrackTemplateFactory::createSilence }
                   } }
             });
+        m_trackThreeClipsId = factory.addTrackFromTemplate("threeClips", {
```
IliasBergstromMuse, line 235: It seems all of these tracks are created and deleted for each test, right? 
Maybe optimise a bit already now, so that each test creates the ones it needs only?

embarc-gabriel: Done!

embarc-gabriel:
> I'm seeing warnings in the test output console for "Uninteresting mock function calls". I know they're not introduced with this PR, but since you're working on tests, how about addressing these while working on the file? I mean those in SetUp mainly.
> 
> `GMOCK WARNING: Uninteresting mock function call - taking default action specified at: /Users/iliasmuse/repos/audacity/src/trackedit/tests/au3interaction_tests.cpp:176: Function call: currentProject() Returns: (ptr = 0x138839418, value = 8-byte object <C8-67 56-05 01-00 00-00>) NOTE: You can safely ignore the above warning unless this call should not happen. Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call. See https://github.com/google/googletest/blob/master/docs/gmock_cook_book.md#knowing-when-to-expect for details. `

This was also fixed. @IliasBergstromMuse

IliasBergstromMuse (approved):
Great work!

There was one more suppression to add for uninteresting calls:

m_globalContext = std::make_shared<testing::NiceMock<au::context::GlobalContextMock>>();

std::shared_ptr<testing::NiceMock<au::context::GlobalContextMock>> m_globalContext;

But you can add that in a future test PR.

src/trackedit/tests/au3interaction_tests.cpp
```diff
@@ -141,8 +170,8 @@ class Au3InteractionTests : public ::testing::Test
     {
         m_au3Interaction = std::make_shared<Au3Interaction>();
 
-        m_globalContext = std::make_shared<context::GlobalContextMock>();
-        m_projectHistory = std::make_shared<ProjectHistoryMock>();
+        m_globalContext = std::make_shared<NiceMock<context::GlobalContextMock> >();
```
IliasBergstromMuse, line 173: Ah, I see you added it, weird that I didn't get it when I pulled from your branhc.

PR #8539, opened 2025-03-27
Fix #7674: Unfocus from the clip should rename it when in Edit mode

### LINKED ISSUE #7674: [AU4] Clicking away from the clip should rename the clip when in Edit mode
### Description

If the User edits any of the clips available, and then 'unfocuses' the text input, the text input won't  be saved. The only way a User can save the text input is by pressing the 'Enter' key while in focus. 

However, the clips' name text input should be saved on any focus lost during the text input. 

https://github.com/user-attachments/assets/d2243e78-6c8c-4fba-8320-ef004622886b

NunoCMatos:
Resolves: #7674

Previously, the onFocusChanged function in ClipItem.qml only hid the titleEdit field when it lost focus, without saving changes.

The expected behavior is that when titleEdit loses focus, it should save the changes automatically.

This fix ensures that clicking away commits the new title by calling titleEdit.accepted(), allowing future validations to be handled in the onAccepted function if needed.

This is the before and after:

### **Before:**

https://github.com/user-attachments/assets/ab98f92a-c145-46db-b2e7-608af75346fd

### **After:**

https://github.com/user-attachments/assets/d4dafdda-e798-477f-a688-fa4448ac267b

Recommended:

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipItem.qml
@@ -451,6 +451,7 @@ Rectangle {
                     onFocusChanged: {
                         if (!titleEdit.focus) {
                             titleEdit.visible = false
+                            titleEdit.accepted()
                         }
                     }
                 }


grliszas14 (approved):

kryksyh:
@NunoCMatos @grliszas14 does it handle "Esc" correctly? It should not accept new name if "Esc" is pressed

grliszas14:
@kryksyh yup, Esc seems to be handled correctly
good point btw

grliszas14:
@NunoCMatos thanks!

Issue #7712, opened 2024-11-13
User can Remove Track UI Permanently until Next Launch


chinakov:
### Bug description

If the User decides to 'unstick' the track panel, and then tries to collpase it horizontally completely, the track panel disappears until next launch. 

### Steps to reproduce

1. Launch Audacity
2. Unstick the Track Panel
3. Resize the Track Panel to its smallest via horizontal handle
4. Observe Result	

### Expected behavior

There should be a limit as to how much the track panel can be collapsed. 

### Actual behavior

There is no limit to how much the track panel can be collapsed. 

### Audacity Version

Audacity 4 alpha

### Operating system

Windows

### Additional context

Video: 

https://github.com/user-attachments/assets/6ad1d730-e048-4d89-b68b-1d4ca5d98869

chinakov:
Also, the User is unable to resize the track panel once it has been undocked and re-docked.

chinakov:
The TCP flyour can't be resized as atll on the latest master. https://github.com/audacity/audacity/actions/runs/13849065281

dozzzzer:
We should also set the maximum width of the panel, too. The same goes for the History panel:

https://github.com/user-attachments/assets/e04e9b76-7ca3-417e-92f4-7fdf6678ae62

chinakov:
This is still reproducible on the latest Master: efdaf1e91b151bcc5fe37cab8c0d1539afaf569c

chinakov:
This is still reproducible for the latest Master (97567e5405b8d6a48a22ca3855f7fce3e937)

chinakov:
This is still reproducible for the latest Master (0580af526411238244e8c9661e804a7978bbb671)

PR #8564, opened 2025-03-31
Resolves : #7712

nunezzz03:
Previously, the 'Dock Panel' in ProjectPage.qml did not have a minimum height set.

The expected behavior is that when resizing the Track Panel, it should have a lower limit to prevent it from collapsing completely and disappearing.

This fix enforces a natural minimum height for the Track Panel, ensuring a better user experience.

Before:

https://github.com/user-attachments/assets/a356ab58-f124-4237-a2c8-d4d6da37ca85

After:

https://github.com/user-attachments/assets/37dfd409-8c63-4a09-aba7-f540690240c9

Recommended:


dozzzzer:
The minimum height of the Track panel is now in place and looks like this:

image

src/appshell/qml/ProjectPage/ProjectPage.qml
```diff
@@ -215,6 +215,8 @@ DockPage {
 
             property int panelWidth: root.verticalPanelDefaultWidth + (showEffectsSection ? effectsSectionWidth : 0)
 
+            minimumHeight: 83
```
grliszas14, line 218: a small nitpick: please move this line down to line 229, where other widths are set

dozzzzer:
It doesn't fix #7712, though.

Resizing the track panel to zero is still doable, but only on Windows and only when resizing the left edge of the panel, not the right one.

https://github.com/user-attachments/assets/d1cde2fb-e1aa-47ce-bf2e-f9ea796a7f53

src/appshell/qml/ProjectPage/ProjectPage.qml
nunezzz03: > a small nitpick: please move this line down to line 229, where other widths are set

I appreciate the feedback! I've adjusted the line as suggested and updated the pull request. Please let me know if you have any further comments.

grliszas14 (approved):

src/appshell/qml/ProjectPage/ProjectPage.qml
grliszas14: @nunezzz03 thanks!

It doesn't fully solve the problem as mentioned by @dozzzzer but it fixes a part of it which makes it a good contribution already, I'm gonna merge it soon

Issue #7930, opened 2024-12-17
Destructive effect modal visuals


DilsonsPickles:
### Description

Designs for the following effect modals: 
- DTMF Tones
- Reverb
- Amplify
- Chirp
- Noise
- Tone

Please note that this design spec does not include how presets work. As the first batch of effects are all destructive, it also does not include the power switch in the modal header. These will be covered in separate specs.

[Figma link](https://www.figma.com/design/w2kAh1gKKV5bu7htY7158W/08---Effects?node-id=4782-60134)

chinakov:
Passing this task to Ready to Merge, but will log subsequent issues that have been introduced in this PR.

PR #8565, opened 2025-03-31
Update effect dialogs (#7930)

kryksyh:
Resolves: #7930

**Note** To not overload this PR, the top bar "<Preset> [save] [manage]" has not been implemented yet and will be implemented in a follow-up PRs

Recommended:

### DIFF
--- a/src/effects/builtin/amplify/AmplifyView.qml
+++ b/src/effects/builtin/amplify/AmplifyView.qml
@@ -1,26 +1,25 @@
 import QtQuick
-
 import Muse.UiComponents
-
 import Audacity.Effects
 
 import "../common"
 
 EffectBase {
-
     id: root
 
-    property string title: "Amplify"
+    property string title: qsTrc("effects/amplify", "Amplify")
     property alias isApplyAllowed: amplify.isApplyAllowed
 
-    width: 300
-    height: 200
+    width: 320
+    height: 250
 
     model: amplify
 
     AmplifyViewModel {
         id: amplify
+
         instanceId: root.instanceId
+
         onAmpChanged: slider.value = amp
     }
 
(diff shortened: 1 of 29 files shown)


src/effects/builtin/tonegen/ChirpView.qml
```diff
@@ -1,154 +1,275 @@
...
-        // Fourth row        
-        StyledTextLabel {
-            text: qsTrc("effects/chirp", "Interpolation:")
-        }
-        ComboBox {
-            Layout.columnSpan: 2
+        RoundedRectangle {
+
+            Layout.fillWidth: true
+            Layout.preferredHeight: amplitudeGroup.height + 32
```
grliszas14, line 172: Looks like this "32" and the one in line 185 are connected somehow? If yes, we can make a const variable and use it in both cases

src/effects/builtin/tonegen/ToneView.qml
```diff
@@ -6,108 +6,138 @@ import Audacity.Effects
...
+
+        width: parent.width
+        height: parent.height
+
+        spacing: prv.spacing
 
-        anchors.horizontalCenter: parent.horizontalCenter
         anchors.fill: parent
-        anchors.margins: 10
+        anchors.rightMargin: 16
```
grliszas14, line 50: I don't think this is crucial here so I don't mind it but generally using layouts + anchors is a performance issue in QML (both of these combined can have conflicts or redundant calculations)

grliszas14 (approved):

src/effects/builtin/tonegen/ToneView.qml
kryksyh: thanks!

src/effects/builtin/tonegen/ChirpView.qml
kryksyh: Thanks, updated!

src/effects/builtin/amplify/AmplifyView.qml
```diff
@@ -31,76 +30,124 @@ EffectBase {
...
+                        }
+                    }
                 }
-            }
-        }
 
-        StyledSlider {
-            id: slider
-            width: parent.width
+                IncrementalPropertyControl {
```
Eism, line 71: ahh, I'm glad we replaced RealInputField with IncrementalPropertyControl
but please check https://github.com/audacity/audacity/pull/8050 on regression

src/effects/builtin/common/EffectBase.qml
```diff
@@ -43,4 +46,45 @@ Rectangle {
...
+
+        RowLayout {
+            spacing: 4
+
+            StyledDropdown {
+                Layout.fillWidth: true
+                indeterminateText: "Default preset"
+            }
+
+
```
Eism, line 63: please remove this extra empty line

Eism (approved):

src/effects/builtin/tonegen/ChirpView.qml
```diff
@@ -1,154 +1,276 @@
...
 
-            model: chirp.interpolationTypes
-            currentIndex: chirp.interpolation
+                spacing: prv.spacing
+                width: parent.width - prv.padding
+                height: implicitHeight
 
-            onActivated: function (index) {
-                chirp.interpolation = index
+                x: prv.spacing
```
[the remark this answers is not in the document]
kryksyh: Oh, struggled with this and resorted to setting x/y explicitly. Could you help me with this particular case? 

if change it with this
```
                anchors.fill: parent
                anchors.topMargin: prv.spacing
                anchors.leftMargin: prv.spacing
```
it screws the layout completely
image

src/effects/builtin/amplify/AmplifyView.qml
kryksyh: @Eism yes, 8050 is back indeed, and it should be fixed for `IncrementalPropertyControl` as well, I'll make the PR today

PR #8573, opened 2025-03-31
8101 changing effect settings make project dirty

### LINKED ISSUE #8101: Tweaking the effect's parameters doesn't make the project dirty
### Bug description

_No response_

### Steps to reproduce

1. Open the project with an effect in the realtime effect stack
2. Open the effect's UI
3. Tweak some parameters
4. Close the project (Ctrl + W)
5. Observe

### Expected behavior

The save dialog is triggered.

### Actual behavior

The save dialog isn't triggered, however, the changes to the effect's parameters are saved.

### Audacity Version

master

### Operating system

OS independent

### Additional context

_No response_

### LINKED ISSUE #8102: Reverb parameter sliders are moving jerkily during and after playback (MacOS)
1. Open a project with clips
2. Add Reverb to the stack
3. Play
4. try to move the parameter sliders of the Reverb
5. Stop
6. try to move the sliders again

Act.: they're moving jerkily 

https://github.com/user-attachments/assets/1a5c11e0-786f-4125-8a49-ce97f5ef858c

saintmatthieu:
Resolves: #8101 
Resolves: #8102

Recommended:

QA:

- [x] Realtime effect dialog power button is in sync with corresponding power button in effect panel (for built-in or VST effects, as master effect or track effect, during playback stopped/active/paused, after undo/redo)
- [x] Changing bypass button or some realtime-effect setting makes the project dirty (for both VSTs or built-in reverb)


au3/libraries/lib-realtime-effects/RealtimeEffectState.cpp
```diff
@@ -348,13 +348,7 @@ RealtimeEffectState::RealtimeEffectState(const PluginID& id)
     BuildAll();
 }
 
-RealtimeEffectState::RealtimeEffectState(const RealtimeEffectState& other)
```
IliasBergstromMuse, line 351: Is there a good reason for why it should no longer have a copy-constructor?
(you know, [rule of three at least if not five etc](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-five)).

src/effects/effects_base/internal/mastereffectundoredo.cpp
```diff
@@ -31,26 +32,71 @@ class MasterEffectListRestorer final : public UndoStateExtension
...
-        : m_targetList(std::make_unique<RealtimeEffectList>(RealtimeEffectList::Get(project)))
+        : m_items(getStateAndSettingsItems(project)),
+        m_masterEffectList{std::make_unique<RealtimeEffectList>(RealtimeEffectList::Get(project))}
     {
     }
 
     void RestoreUndoRedoState(au3::Au3Project& project) override
     {
-        auto& changed = RealtimeEffectStackChanged::Get(project).channel;
+        for ( auto& item : m_items) {
```
IliasBergstromMuse, line 46: Here if we follow the Muse guideline we'd state the type in place of auto.

src/effects/effects_base/internal/mastereffectundoredo.cpp
```diff
@@ -31,26 +32,71 @@ class MasterEffectListRestorer final : public UndoStateExtension
...
+        {
+        }
+
+        const RealtimeEffectStatePtr state;
+        const EffectSettings settings;
+    };
+
+    static void append(std::vector<StateAndSettings>& items, RealtimeEffectList& list)
+    {
+        for (auto i = 0; i < static_cast<int>(list.GetStatesCount()); ++i) {
```
IliasBergstromMuse, line 81: Why not int i?

src/effects/effects_base/internal/realtimeeffectservice.cpp
```diff
@@ -99,17 +105,14 @@ void RealtimeEffectService::registerRealtimeEffectList(TrackId trackId, Realtime
         }
     });
 
+    // Proactively load realtime effects.
+    for (auto i = 0u; i < list.GetStatesCount(); ++i) {
```
IliasBergstromMuse, line 109: And here a different style for the same loop as above?

src/effects/effects_base/internal/realtimeeffectservice.cpp
```diff
@@ -332,23 +340,26 @@ void RealtimeEffectService::moveRealtimeEffect(const RealtimeEffectStatePtr& sta
                                        "Change effect order");
 }
 
-bool RealtimeEffectService::isActive(const RealtimeEffectStatePtr& stateId) const
+bool RealtimeEffectService::isActive(const RealtimeEffectStatePtr& state) const
 {
-    if (stateId == 0) {
+    if (state == nullptr) {
```
IliasBergstromMuse, line 345: Can it ever under normal circumstances be called with a null argument?
If not, I'm conflicted about defensive programming like this. Unless there's a bug to catch? In which case I really like the IF_ASSERT_FAILED macro.

IliasBergstromMuse (approved):
Builds and works as expected!
Some minor comments only, and a question regarding the defensive programming which may or may not be addressed depending on why it's there.

src/effects/effects_base/internal/mastereffectundoredo.cpp
saintmatthieu: Oh no please, don't begin ...

src/effects/effects_base/internal/realtimeeffectservice.cpp
```diff
@@ -52,7 +58,7 @@ void RealtimeEffectService::onProjectChanged(const au::project::IAudacityProject
 
     auto au3Project = reinterpret_cast<au::au3::Au3Project*>(project->au3ProjectPtr());
 
-    setNotificationChannelForMasterEffectUndoRedo(*au3Project, m_realtimeEffectStackChanged);
+    setNotificationChannelForMasterEffectUndoRedo(*au3Project, this);
```
igorkorsukov, line 61: There might be a problem here: 
* In AU4, most services are single-instance, accessible via IoC, and stateless.
* Here we "transform" an independent service into a project-specific service 
* Moreover, we use a raw pointer, so we won’t be able to determine whether the object has been deleted or not.

I believe that this duality - is this service one instance or is there a separate instance for each project will lead to confusion and problems in the future.

src/effects/vst/view/vstviewmodel.cpp
```diff
@@ -39,7 +39,19 @@ void VstViewModel::init()
...
+
+    m_auVst3Instance->GetWrapper().ParamChangedHandler = [this](Steinberg::Vst::ParamID) {
+        projectHistory()->modifyState();
+        projectHistory()->markUnsaved();
+    };
+
     settingsToView();
+
+    // When playback is idle (see VstViewModel::event), no need for setting updates to be low-latency. Every 100ms is plenty.
+    startTimer(std::chrono::milliseconds { 100 });
```
igorkorsukov, line 54: Doesn't require fixing, but we usually use QTimer. Its use gives several advantages - the fact that there is a timer is clear, its setup, start and callback (slot or lambda) in one place is clearly visible, it is easier to manage stop, restart, etc.

src/effects/effects_base/internal/mastereffectundoredo.h
saintmatthieu, line 1: I forgot: if the reviewers are okay with this PR in principle, I intend to rename this file and its methods: now it's not about master effects anymore.

src/effects/effects_base/internal/realtimeeffectservice.cpp
saintmatthieu: That's an interesting comment, thank you. Is this going to be an issue, though?
The way AU4 is built now, there can only be one project open per process - this is at least assumed by most of the classes that, in their `init()`, will listen to `globalContext()->projectChanged()`. Consistently when we close a project, we delete the corresponding `::AudacityProject`, which in turn destroys the `RealtimeEffectStackChanged` that's hiding behind this function. Hence, there won't be two `RealtimeEffectStackChanged` instances at a time with pointers to the single-instance `RealtimeEffectService`.

I can elaborate if you like on how the constraints that got me to this solution. Note that it isn't particular to this PR, it's already on master, only here we pass `this` instead of a copy of `muse::async::Channel<TrackId>`. If there is an issue with passing `this`, we could pass `m_realtimeEffectStackChanged` and `m_effectSettingsChanged` - I just passed this for convenience and less maintenance need.

saintmatthieu: > Moreover, we use a raw pointer, so we won’t be able to determine whether the object has been deleted or not.

I intentionally passed a raw pointer, to express that the other side does not have to worry about checking validity (since the lifetime of `this` encompasses that of all projects).

src/effects/effects_base/internal/mastereffectundoredo.cpp
saintmatthieu: `const StateAndSettings&` is better indeed - just `auto&` made even me think `item` was mutable.
(I'd also have liked uncrustify to remove the space between `( auto` :/

src/effects/effects_base/internal/mastereffectundoredo.cpp
saintmatthieu: Why not ? `0` is an int literal.

IliasBergstromMuse: True, and the cast is also in the same line.

src/effects/effects_base/internal/realtimeeffectservice.cpp
saintmatthieu: Here the use case would be that a client to this service (e.g. an effect dialog) lives with the possibility of its pointer having expired (if it has a weak_pointer). In that case it seems easier programmatically to let the service centrally deal with this possibility, sparing clients the need to control their pointers.
That said, I don't have a strong opinion - I hesitated, too.

src/effects/vst/view/vstviewmodel.cpp
saintmatthieu: Works :thumbs-up:

src/effects/effects_base/internal/realtimeeffectservice.cpp
igorkorsukov: > The way AU4 is built now

This is not a question of the current state (everything will be ok now), but a question of development. What will other developers do in the future and based on what assumptions?... 
For example, could there be a situation in the future where two or more projects are loaded? Maybe this will be needed for some batch processing in console mode...

Another problem, maybe for some reason we will need to reset the state of the IoC container and recreate the services... I don't know why exactly we might want to do this, although I can name one reason - maybe we will want to replace the service implementation at runtime, i.e. unregister one implementation and register another, maybe it will be switching to an experimental version, or to an advanced version... something like that. In this case, the pointer in the RealtimeEffectStackChanged will not be valid. 

> we could pass m_realtimeEffectStackChanged and m_effectSettingsChanged - I just passed this for convenience and less maintenance need. 

I see two options for solving this design problem:  
1. Add a structure with channels as fields, maybe add some methods to it for convenience and pass it on 
like 
```
strcut RealtimeEffectStackNotifier {
         muse::async::Channel<TrackId> stackChanged;
         ... settingChanged;
}  


void setNotificationChannelForMasterEffectUndoRedo(au::au3::Au3Project& project, RealtimeEffectStackNotifier notifier)
{
...
}
```
i.e. we have some object for notification... and someone passed it and it doesn't matter who.

2. Add proxy for this task 
Like 
```
class RealtimeEffectStackNotifier {
      Inject<IRealtimeEffectService> service;

public: 

....
}

void setNotificationChannelForMasterEffectUndoRedo(au::au3::Au3Project& project, RealtimeEffectStackNotifier* notifier)
{
...
}
```
i.e. we have an object for notification, it is in the context of the project itself, but refers to one instance of the service, which is used as usual through injection. There is another additional benefit here - we can limit the interface.

igorkorsukov (approved):

PR #8578, opened 2025-04-01
EffectsViewerDialog is reused for VSTs

### LINKED ISSUE #7796: Effects - Presets button fix
### Description

For all the effects, instead of "Manage" button bottom left corner, we should have "Presets & Settings" button located upper left corner.

### LINKED ISSUE #8582: Refactoring: consolidation of built-in and VST destructive effect views

saintmatthieu:
Resolves: #8582
Resolves: #7796

Architecture before:
image

Architecture after:
image

Recommended:

QA:
This PR should only affect effects used in destructive mode.
- [ ] Built-in effects in destructive mode behave like on master
- [ ] VST effects now are wrapped in the same container as built-in effects
- [ ] 8582 is fixed
- [ ] 7796 is fixed


src/effects/effects_base/effectstypes.h
```diff
@@ -62,6 +62,8 @@ constexpr const char16_t* EFFECT_OPEN_ACTION = u"action://effects/open?effectId=
 constexpr const char16_t* REALTIME_EFFECT_ADD_ACTION = u"action://effects/realtime-add?effectId=%1";
 constexpr const char16_t* REALTIME_EFFECT_REPLACE_ACTION = u"action://effects/realtime-replace?effectId=%1";
 
+constexpr const char16_t* EFFECT_VIEWER_URI = u"audacity://effects/effect_viewer?instanceId=%1&isVst=%2";
```
igorkorsukov, line 65: It would be good to immediately keep in mind other types of effects, i.e. to use something like: 
```
// see enum class EffectFamily , use lower case 
constexpr const char16_t* EFFECT_VIEWER_URI = u"audacity://effects/effect_viewer?instanceId=%1&family=%2"
```

igorkorsukov: I don't mind keep this for further development when some other family appears

src/effects/builtin/internal/builtinviewlauncher.cpp
```diff
@@ -12,24 +12,12 @@
...
-    muse::String type = au3::wxToString(effect->GetSymbol().Internal());
-    muse::Ret ret = interactive()->open(muse::String(BUILTIN_VIEWER_URI)
-                                        .arg(type)
-                                        .arg(size_t(instanceId)).toStdString()
-                                        ).ret;
-
-    return ret;
+    muse::UriQuery uri(muse::String(EFFECT_VIEWER_URI)
+                       .arg(size_t(instanceId)).toStdString());
+    uri.addParam("isVst", muse::Val(false));
```
igorkorsukov, line 19: it's a bit inconsistent here, one parameter is set by formatting the string (arg method), the other is like setting the parameter in the URI. I think we need to choose one :)

igorkorsukov (approved):

src/effects/builtin/internal/builtinviewlauncher.cpp
saintmatthieu: Sure.

IliasBergstromMuse (approved):
Nothing to remark on, great to reduce repetition, and great to see it will be paying off already with the merge of the preset button rework!

igorkorsukov (approved):

Issue #8463, opened 2025-03-18
Add links to Audio Setup button


ianalarina:
### Description

Audio Setup button should point to audio device preferences, similar to AU3 

QA:
- [x] when audio setup button is clicked, a context menu is shown
- [x] list of available devices in the context menu always matches audio preferences
- [x] currently selected device is checkmarked
- [x] there's a "Audio preferences" entry in the context menu which takes you straight to Audio preferences
"Rescan devices" entry is not implemented yet

dozzzzer:
@grliszas14 
Audio settings selected in the Audio Setup menu, such as host, I/O device, no. of channels, aren't remembered after restart.

PR #8594, opened 2025-04-02
Audio setup button implementation

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8463

### DIFF
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -2,7 +2,7 @@
 * Audacity: A Digital Audio Editor
 */
 #include "playbackcontroller.h"
-
+#include "playbackuiactions.h"
 #include "../playbacktypes.h"
 
 using namespace muse;
@@ -26,6 +26,11 @@ static const ActionCode REPEAT_CODE("repeat");
 
 static const secs_t TIME_EPS = secs_t(1 / 1000.0);
 
+static const ActionQuery CHANGE_AUDIO_API_QUERY("action://playback/change-api");
+static const ActionQuery CHANGE_PLAYBACK_DEVICE_QUERY("action://playback/change-playback-device");
+static const ActionQuery CHANGE_RECORDING_DEVICE_QUERY("action://playback/change-recording-device");
+static const ActionQuery CHANGE_INPUT_CHANNELS_QUERY("action://playback/change-input-channels");
+
 void PlaybackController::init()
 {
     dispatcher()->reg(this, PLAY_CODE, this, &PlaybackController::togglePlay);
@@ -40,6 +45,11 @@ void PlaybackController::init()
     // dispatcher()->reg(this, LOOP_OUT_CODE, [this]() { addLoopBoundary(LoopBoundaryType::LoopOut); });
     dispatcher()->reg(this, REPEAT_CODE, this, &PlaybackController::togglePlayRepeats);
     dispatcher()->reg(this, PAN_CODE, this, &PlaybackController::toggleAutomaticallyPan);
+
+    dispatcher()->reg(this, CHANGE_AUDIO_API_QUERY, this, &PlaybackController::setAudioApi);
+    dispatcher()->reg(this, CHANGE_PLAYBACK_DEVICE_QUERY, this, &PlaybackController::setAudioOutputDevice);
+    dispatcher()->reg(this, CHANGE_RECORDING_DEVICE_QUERY, this, &PlaybackController::setAudioInputDevice);
+    dispatcher()->reg(this, CHANGE_INPUT_CHANNELS_QUERY, this, &PlaybackController::setInputChannels);
 
     globalContext()->currentProjectChanged().onNotify(this, [this]() {
         onProjectChanged();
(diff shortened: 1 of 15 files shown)


src/playback/internal/playbackcontroller.cpp
```diff
@@ -391,6 +401,54 @@ void PlaybackController::toggleLoopPlayback()
...
 }
 
+void PlaybackController::setAudioApi(const muse::actions::ActionQuery& q)
+{
+    if (!q.contains("api")) {
+        return;
+    }
+
+    int index;
+    index = q.param("api").toInt();
```
igorkorsukov, line 411: I think better 
```
int index = q.param("api").toInt();
```

src/playback/internal/playbackcontroller.cpp
```diff
@@ -391,6 +401,54 @@ void PlaybackController::toggleLoopPlayback()
     NOT_IMPLEMENTED;
 }
 
+void PlaybackController::setAudioApi(const muse::actions::ActionQuery& q)
+{
+    if (!q.contains("api")) {
+        return;
+    }
+
+    int index;
```
igorkorsukov, line 410: I think better 
```
int index = q.param("api").toInt();
```

src/playback/internal/playbackcontroller.cpp
igorkorsukov, line 411: Maybe it's better to call the argument "api_index", otherwise it seems that there should be a name or some kind of code. By the way, isn't it possible to use some kind of code instead of an index from the list?

src/playback/internal/playbackuiactions.cpp
```diff
@@ -108,6 +125,34 @@ const UiActionList PlaybackUiActions::m_mainActions = {
              TranslatableString("action", "Set playback level"),
              IconCode::Code::AUDIO
              ),
+    UiAction("action://playback/change-api",
```
igorkorsukov, line 128: A variable (constant) is declared there, but it is not used here, maybe then remove it.

src/projectscene/view/toolbars/audiosetupcontextmenumodel.cpp
```diff
@@ -0,0 +1,233 @@
...
+}
+
+void AudioSetupContextMenuModel::updateOutputDeviceCheckedState()
+{
+    auto currentOutputDevice = audioDevicesProvider()->currentAudioOutputDevice();
+    auto outputDevicesList = audioDevicesProvider()->audioOutputDevices();
+
+    for (int i = 0; i < m_outputDeviceActionCodeList.size(); ++i) {
+        MenuItem& item = findItem(QString::fromStdString(m_outputDeviceActionCodeList.at(i)));
+        ActionQuery query(m_outputDeviceActionCodeList.at(i));
```
igorkorsukov, line 179: look like not used

src/projectscene/view/toolbars/audiosetupcontextmenumodel.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+        if (outputDevicesList.at(i) == currentOutputDevice) {
+            auto state = item.state();
+            state.checked = true;
+            item.setState(state);
```
igorkorsukov, line 184: There's already a `setChecked` method there.

src/playback/internal/playbackcontroller.cpp
grliszas14: Those devices doesn't have codes but names. Names can contain spaces -> that's why I didn't want to use them as ActionQuery arguments, not sure if that can work and indexes work just fine
I'll change it to api_index etc.

src/projectscene/view/toolbars/audiosetupcontextmenumodel.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+void AudioSetupContextMenuModel::updateApiCheckedState()
+{
+    auto currentApi = audioDevicesProvider()->currentAudioApi();
+    auto audioApiList = audioDevicesProvider()->audioApiList();
+
+    for (int i = 0; i < m_apiActionCodeList.size(); ++i) {
+        MenuItem& item = findItem(QString::fromStdString(m_apiActionCodeList.at(i)));
+
+        if (audioApiList.at(i) == currentApi) {
```
igorkorsukov, line 159: I don't really like the approach of having two lists that need to be synchronized... As I see it, when we change the current API selection (and other functions) or when we change the list of available APIs (or devices, for example), the entire menu is recreated. Accordingly, there is no need for these state update functions, they are never called after creation. I think that then it's better to remove them (and these variables with a list of codes) and set a check in the items creation methods.

grliszas14: Good point, fixed it

src/projectscene/view/toolbars/audiosetupcontextmenumodel.cpp
```diff
@@ -0,0 +1,152 @@
...
+    auto makeChangeRecordingDeviceAction = [](int index) {
+        ActionQuery q("action://playback/change-recording-device");
+        q.addParam("device_index", muse::Val(index));
+        return q;
+    };
+
+    const auto& inputDevicesList = audioDevicesProvider()->audioInputDevices();
+    for (int i = 0; i < inputDevicesList.size(); ++i) {
+        MenuItem* item = makeMenuItem(makeChangeRecordingDeviceAction(i).toString(),
+                                      muse::TranslatableString("audio setup", muse::String::fromStdString(inputDevicesList.at(i))));
```
igorkorsukov, line 118: need to use `TranslatableString::untranslatable(muse::String::fromStdString(inputDevicesList.at(i))))`  instead of `muse::TranslatableString("audio setup", muse::String::fromStdString([inputDevicesList.at](http://inputdeviceslist.at/)(i))))`

igorkorsukov (approved):

Issue #8125, opened 2025-01-29
Cannot add new tracks when launching Audacity for the first time


dozzzzer:
Expected behavior: after finishing the initial configuration, instead of creating a new project, the User must be taken to the Home screen.

https://github.com/user-attachments/assets/52777ecf-65c7-4dba-9971-45d91d9bca53

chinakov:
Still reproducible on the latest Master: https://github.com/audacity/audacity/actions/runs/13919899253

chinakov:
Just one question - when the User is on the final step of the 'Getting Started' flow, the Learn tab closes immediately after pressing 'Finish'. I assume this is intentional, but I'd like to re-check with @LWinterberg and @grliszas14

chinakov:
https://github.com/user-attachments/assets/6e8742da-b380-485d-bc91-9fbfe763892a

dozzzzer:
@grliszas14 @Tantacrul @LWinterberg 
Once I complete the setup, a new project is created. After I save it and close it, I see the "Learn" tab, which isn't what I expected. Do we want to show the "Projects" tab instead?

https://github.com/user-attachments/assets/0bbcc9fd-f929-4320-8588-e5b03e95fe8a

grliszas14:
@dozzzzer for sure we do, fixed it

dozzzzer:
@grliszas14 Tested, works like a charm

PR #8598, opened 2025-04-03
Fix first launch dialog

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8125

### DIFF
--- a/src/appshell/internal/startupscenario.cpp
+++ b/src/appshell/internal/startupscenario.cpp
@@ -180,6 +180,8 @@ void StartupScenario::onStartupPageOpened(StartupModeType modeType)
 
     if (!configuration()->hasCompletedFirstLaunchSetup()) {
         interactive()->open(FIRST_LAUNCH_SETUP_URI);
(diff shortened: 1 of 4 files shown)


src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
```diff
@@ -108,7 +108,12 @@ bool FirstLaunchSetupModel::askAboutClosingEarly()
                                  buttons,
                                  int(IInteractive::Button::Cancel));
 
-    return result.standardButton() == IInteractive::Button::Cancel;
+    auto button = result.standardButton();
+    if (button == IInteractive::Button::Cancel) {
+        interactive()->open("musescore://home?section=projects");
```
Eism, line 113: The askAboutClosingEarly method only asks and returns the user's choice.
The logic of what we should do with a certain user choice should be stored in the StartupScenario where we call FIRST_LAUNCH_SETUP dialog

PR #8614, opened 2025-04-07
Add entry on history on stem plot edit

### LINKED ISSUE #8501: Stem plot editing doesn't cause an entry into undo/redo history. Nor does it set a "dirty" flag.
The new AU4 stem plot sample editing doesn't currently result in an entry into history, and cannot be undo/redone.

This is a regression over AU3.

Stem plot editing also doesn't set a "dirty" flag, so after editing, a user can exit the program and lose their work since there's no save/yes/no warning dialog.

embarc-gabriel:
Resolves: #8501

Implemented history entries for sample edits. The versioning system has been adjusted: previously, the clip version was incremented only upon new clip creation. Now, it also tracks changes to the internal data. This enhancement allows us to efficiently determine if a clip has been modified without needing to compare the samples directly.

Currently, the history does not differentiate between sample edit types, and consecutive edits are always consolidated, regardless of the time interval. A separate ticket will be created to address and improve this behavior.

Recommended:

### DIFF
--- a/au3/libraries/lib-wave-track/WaveClip.cpp
+++ b/au3/libraries/lib-wave-track/WaveClip.cpp
@@ -442,6 +442,9 @@ void WaveClip::SetSamples(size_t ii,
 {
     StrongInvariantScope scope{ *this };
     assert(ii < NChannels());
+
+    mVersion++;
+
     // use Strong-guarantee
     mSequences[ii]->SetSamples(buffer, format,
                                start + TimeToSamples(mTrimLeft), len, effectiveFormat);
--- a/au3/libraries/lib-wave-track/WaveClip.h
+++ b/au3/libraries/lib-wave-track/WaveClip.h
@@ -1016,7 +1016,7 @@ private:
     };
 
     int64_t mId = -1;
-    const int64_t mVersion = 0;
+    int64_t mVersion = 0;
 
     //! Real-time durations, i.e., stretching the clip modifies these.
     //! @{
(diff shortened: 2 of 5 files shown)


IliasBergstromMuse:
I agree about the version increment in principle - it's much better than checking for a change over all samples for sure, that's way too inefficient.
There is a difference in how I used it:
The increment in my case was on copy only, so incrementing in the copy construction.
In your case the field can also be incremented on sample editing.
Maybe it's fine to re-use the same field!
Could there be a problem, if there's a "compound" action, which includes both copy and sample editing, in which case the two are confused, meaning you'll need a new field? I'm not sure, but. I don't think so.
If yes: add another field. But probably it's not needed.

IliasBergstromMuse:
I tried using it to see how often this creates "sample edit" entries in history. In fact for any sequence of edits, it only creates one entry. This bears discussing with the UX team!
See sample editing as painting in e.g. Illustrator. 
Sometimes several sample edits should be grouped together as one action. Other times they should be grouped into several.
The criteria for this can be comples:
1 - if the sample edits resulted from the same mouse-gesture, they should be grouped together.
2 - MAYBE if they are resulting from several individual clicks, but very close together in time, they should also be grouped together. With a threshold of e.g. 5 seconds or something like that. 

The above are just examples. This is worth checking with UX to get right in my opinion.

embarc-gabriel:
@IliasBergstromMuse thank you for the feedback.
Yes! You are right about this.
I have kept the same behavior from AU3 but for sure this can be an opportunity to improve the solution.
I will ask UX team about this.

IliasBergstromMuse:
> @IliasBergstromMuse thank you for the feedback. Yes! You are right about this. I have kept the same behavior from AU3 but for sure this can be an opportunity to improve the solution. I will ask UX team about this.

Good that you mention it's the same as AU3! Because then it is more likely we leave it as it is. Still worth asking of course, but we do want to be mindful about feature creep.

IliasBergstromMuse (approved):

Issue #8094, opened 2025-01-24
When trying to move multiple selected clip groups, only one group can be moved


dozzzzer:
https://github.com/user-attachments/assets/2bb5b2bb-b128-4981-9943-ceca23a54ce9

chinakov:
Still reporducible on Master: https://github.com/audacity/audacity/actions/runs/13919899253

dozzzzer:
Tested

PR #8615, opened 2025-04-07
Fix moving multiple groups

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8094


IliasBergstromMuse (approved):
Both single and multiple group dragging works now indeed.
I don't know my QML well enough yet to understand your change though, so maybe do give the second reviewer a chance to take a look.

Eism (approved):

Issue #8505, opened 2025-03-24
Recording Device can't be selected before recording


chinakov:
### Bug description

If the User has selected a new Audio Host Engine, and then tries to choose the recording device, the User won't be able to change the recording device until the User has commited the changes and recorded some audio.

https://github.com/user-attachments/assets/0ab0d703-d992-435a-a99f-4079019cd105

### Steps to reproduce

1. Launch Audacity
2. Switch the Audio Host Mode to DS or WASAPI
3. Try to select Recording device
4. Observe result
5. Press 'OK' 
6. Record some Audio
7. Open the Preferences once again
8. Observe that the recording device can be changed

### Expected behavior

The User should be able to select the recording device immidately. 

### Actual behavior

The User is not able to select any recording device in audio settings before initiating recording. 

### Audacity Version

a81258845f153655eb1140be16a3c93768a50df5

### Operating system

Windows

chinakov:
Fixed by https://github.com/audacity/audacity/pull/8623, thanks!

Issue #8467, opened 2025-03-19
Reset preferences doesn't reset the values right away; reopening the dialog is required.


dozzzzer:
### Steps to reproduce

1. Go to Preferences
2. Change any value to non-default
3. Click Reset preferences in the bottom left
4. Observe
5. Close the dialog and reopen
6. Observe

### Expected behavior

- After reset preferences is clicked, the previously changed value is reset to default

### Actual behavior

- After reset preferences is clicked, the previously changed value remains the same
- After reopening the Preferences, the value is reset to default

### Audacity Version

master

### Operating system

OS independent

chinakov:
'Reset Preferences' button instantly resets all preferences set now, thanks! Tested.

PR #8623, opened 2025-04-08
Fix resetting preferences

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8467

This fixes: resetting preferences to default + remembering settings after app restart
All of this is done via migrating to muse::settings

QA:
* all of the settings should be reset immediately without need to re-open preferences dialog
* please check if these settings still work properly (these were modified): every preference in "audio settings" tab, temporary files location
* all of the settings should be remembered on app restart
* bonus: this may fix https://github.com/audacity/audacity/issues/8505

### DIFF
--- a/src/appshell/qml/Preferences/internal/TemporaryFilesSection.qml
+++ b/src/appshell/qml/Preferences/internal/TemporaryFilesSection.qml
@@ -31,6 +31,10 @@ BaseSection {
 
     property string temporaryPath: ""
 
+    onTemporaryPathChanged: {
+        dirPicker.path = temporaryPath
+    }
+
     signal temporaryFilesLocationChanged(string newPath)
 
     Column {
--- a/src/au3wrap/internal/au3audiodevicesprovider.cpp
+++ b/src/au3wrap/internal/au3audiodevicesprovider.cpp
@@ -613,16 +613,7 @@ std::string Au3AudioDevicesProvider::defaultOutputDevice()
         return "";
     }
 
-    int index = -1;
-    auto apiName = currentAudioApi();
-    int nHosts = Pa_GetHostApiCount();
-    for (int i = 0; i < nHosts; ++i) {
-        wxString name = wxSafeConvertMB2WX(Pa_GetHostApiInfo(i)->name);
-        if (name == apiName) {
-            index = i;
-            break;
-        }
-    }
+    int index = AudioIO::GetHostIndex(currentAudioApi());
 
     const auto currentOutputDevice = currentAudioOutputDevice();
     const auto outputDevices = audioOutputDevices();
@@ -648,16 +639,7 @@ std::string Au3AudioDevicesProvider::defaultInputDevice()
         return "";
     }
 
-    int index = -1;
-    auto apiName = currentAudioApi();
-    int nHosts = Pa_GetHostApiCount();
-    for (int i = 0; i < nHosts; ++i) {
-        wxString name = wxSafeConvertMB2WX(Pa_GetHostApiInfo(i)->name);
-        if (name == apiName) {
-            index = i;
-            break;
-        }
-    }
+    int index = AudioIO::GetHostIndex(currentAudioApi());
 
     const auto currentInputDevice = currentAudioInputDevice();
     const auto inputDevices = audioInputDevices();
--- a/src/au3wrap/internal/au3audiodevicesprovider.cpp
+++ b/src/au3wrap/internal/au3audiodevicesprovider.cpp
@@ -2,6 +2,8 @@
 * Audacity: A Digital Audio Editor
 */
 
+#include "settings.h"
+
 #include "internal/wxtypes_convert.h"
 #include "au3wrap/au3types.h"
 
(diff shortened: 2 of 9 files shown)


src/appshell/qml/Preferences/internal/TemporaryFilesSection.qml
```diff
@@ -31,6 +31,10 @@ BaseSection {
 
     property string temporaryPath: ""
 
+    onTemporaryPathChanged: {
+        dirPicker.path = temporaryPath
```
Eism, line 35: do we need this? we already have this binding in 57 line

src/au3wrap/internal/au3audiodevicesprovider.cpp
```diff
@@ -21,11 +23,141 @@
...
+
+static const muse::Settings::Key DEFAULT_SAMPLE_RATE("au3wrap", "SamplingRate/DefaultProjectSampleRate");
+static const muse::Settings::Key SAMPLE_FORMAT("au3wrap", "SamplingRate/DefaultProjectSampleFormatChoice");
+
 void Au3AudioDevicesProvider::init()
 {
     initHosts();
     initHostDevices();
+
+    std::string defaultHost = !audioApiList().empty() ? audioApiList()[0] : "";
```
Eism, line 42: better to use front

src/au3wrap/internal/au3audiodevicesprovider.cpp
```diff
@@ -21,11 +23,141 @@
...
+    muse::settings()->setDefaultValue(AUDIO_HOST, muse::Val(defaultHost));
+    muse::settings()->valueChanged(AUDIO_HOST).onReceive(nullptr, [this](const muse::Val& val) {
+        const std::vector<DeviceSourceMap>& outMaps = DeviceManager::Instance()->GetOutputDeviceMaps();
+
+        for (const auto& device : outMaps) {
+            if (device.hostString == wxString(val.toString())) {
```
Eism, line 48: please take this conversion out of the loop

src/au3wrap/internal/au3audiodevicesprovider.cpp
Eism: Also, defining the default device should be more complicated than just selecting the first device from the list. 
I see that DeviceManager has GetDefaultOutputDevice, but I don't know what hostIndex means. 
In MuseScore, for example, we have the concept of "default" deviceId, and depending on this value, we establish a connection differently.

src/au3wrap/internal/au3audiodevicesprovider.cpp
```diff
@@ -21,11 +23,141 @@
...
+    std::string defaultOutput = !audioOutputDevices().empty() ? audioOutputDevices()[0] : "";
+    muse::settings()->setDefaultValue(PLAYBACK_DEVICE, muse::Val(defaultOutput));
+    muse::settings()->valueChanged(PLAYBACK_DEVICE).onReceive(nullptr, [this](const muse::Val& val) {
+        AudioIOPlaybackDevice.Write(wxString::FromUTF8(val.toString()));
+        Au3AudioDevicesProvider::handleDeviceChange();
+
+        m_audioOutputDeviceChanged.notify();
+    });
+
+    std::string defaultInput = !audioInputDevices().empty() ? audioInputDevices()[0] : "";
```
Eism, line 68: Also, defining the default device should be more complicated than just selecting the first device from the list.
I see that DeviceManager has GetDefaultOutputDevice, but I don't know what hostIndex means.
In MuseScore, for example, we have the concept of "default" deviceId, and depending on this value, we establish a connection differently.

src/au3wrap/internal/au3audiodevicesprovider.cpp
```diff
@@ -21,11 +23,141 @@
...
+    muse::settings()->setDefaultValue(RECORDING_DEVICE, muse::Val(defaultInput));
+    muse::settings()->valueChanged(RECORDING_DEVICE).onReceive(nullptr, [this](const muse::Val& val) {
+        std::vector<std::string> inputDevices;
```
Eism, line 71: let's move this code to separate method

src/appshell/qml/Preferences/internal/TemporaryFilesSection.qml
grliszas14: yeah, but in the line 64 binding is broken so it only changes if there's a signal emitted from FilePicker -> and we need to handle preference reset from the outside.

To be honest I don't like it as well, I found a bug with this FilePicker so maybe I can log it and fix these bindings by the way in the next PR?

src/au3wrap/internal/au3audiodevicesprovider.cpp
grliszas14: yes! my mistake, I separated 2 function especially for getting default input and output devices, should be good now

src/au3wrap/internal/au3audiodevicesprovider.cpp
```diff
@@ -560,31 +624,54 @@ void Au3AudioDevicesProvider::updateInputOutputDevices()
...
+std::string Au3AudioDevicesProvider::defaultInputDevice()
+{
+    // get api index
+    if (audioApiList().size() < 1) {
+        return "";
+    }
+
+    int index = -1;
+    auto apiName = currentAudioApi();
+    int nHosts = Pa_GetHostApiCount();
```
Eism, line 653: I would suggest not to use low-level services in this class. Btw, can we use AudioIOBase::getRecordDevIndex here? As a last resort, move this code to one of the au3 classes

src/appshell/qml/Preferences/internal/TemporaryFilesSection.qml
Eism: that would be awesome!

src/au3wrap/internal/au3audiodevicesprovider.cpp
grliszas14: not really, it returns the index of recording device while we need an index of a host(api) -> different host means different lists of input/output devices. Moved to au3 class though 👍

Eism (approved):

Issue #8173, opened 2025-02-05
When clip's speed is close to 100%, there's no speed indicator


dozzzzer:
### Bug description

When the clip's speed is in the range from 99.500% to 100.499% AND isn't equal to 100.000%, the speed indicator in the clip title bar is gone.

### Steps to reproduce

1. Have a clip in the project
2. Set the clip's speed to any value within the range 99.500-100.499% but not 100%
3. Observe

### Expected behavior

The speed value is in the clip's title bar:

<img width="223" alt="Image" src="https://github.com/user-attachments/assets/875552a3-2591-481a-93ab-5e18725612cb" />

### Actual behavior

No speed value is shown in the clip title bar:

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/7ede419d-987e-49a9-957b-03a6d1e38d0a" />

### Audacity Version

master

### Operating system

OS independent

chinakov:
Still reproducable on the latest master: https://github.com/audacity/audacity/actions/runs/13890738066

dozzzzer:
Tested

dozzzzer:
@kryksyh 

Try to stretch a clip by holding Alt (Option) to e.g. 98% and then moving the cursor in the opposite direction until the yellow guideline appears, or in other words, stretching snaps to the original length. Observe: the speed indicator reads "99.9%" although, if you click it, in the Pitch and Speed dialog, the value is 100%:

https://github.com/user-attachments/assets/b6398d45-c477-4891-b395-6628de27127c

chinakov:
This is still reproducible on the latest Master: c20b94f310442745900828ac42eed7a72a67f553

chinakov:
This is still reproducible on the latest Master (0580af526411238244e8c9661e804a7978bbb671)

PR #8632, opened 2025-04-09
8173 speed indicator near 100%

MargaridaLopes12341:
Resolves: #8173 speed indicator near 100%

Update the ClipItem file to store the speed as a double, so it's possible to distinguish between the actual speed and the rounded speed percentage. Now, when the rounded apeed percentage is 100% it appears ~100%.

Recommended:


kryksyh (approved):
~~LGTM, @grliszas14 please look at this one too~~

kryksyh (changes requested):
@MargaridaLopes12341 We need to follow the behavior of Audacity3, and in case the percentage falls into (99 - 101) range, clip header should show one digit after the dot

kryksyh:
> We need to follow the behavior of Audacity3, and in case the percentage falls into (99 - 101) range, clip header should show one digit after the dot

@LWinterberg @DilsonsPickles please confirm if this is the intended, or we could proceed with displaying the speed as `~100%`

MargaridaLopes12341:
Hey! I made the changes you requested 🙂 Please have a look and let me know if anything else needs adjustment. Thanks a lot!

MargaridaLopes12341:
@kryksyh @grliszas14

kryksyh:
@MargaridaLopes12341 hey!
I don't think that your current version solves the issue. As I mention in the first comment:

> We need to follow the behavior of Audacity3, and in case the percentage falls into (99 - 101) range, clip header should show one digit after the dot

Please check this out:
AU3:
image

Your branch on AU4:
image

MargaridaLopes12341:
@kryksyh hey!
Quick question — I noticed that 100.01 is showing as 100.1, but since the first digit after the decimal is 0, I would’ve expected 100.0.
Are we rounding here on purpose, or should it just display the first decimal digit without rounding?
Also, how should something like 99.34 be displayed — 99.3 or 99.4? Just want to be sure I’m following the expected behavior 😊

kryksyh:
@MargaridaLopes12341 

> 100.01 is showing as 100.1, but since the first digit after the decimal is 0, I would’ve expected 100.0.

Yes, here we step away from mathematically correct rounding, to give the user feedback that __some__ stretching is applied.

Please follow the code and description provided for AU3: https://github.com/audacity/audacity/blob/774ef8a28ddbed342ea26bbf8c98b7a8aa378f0a/au3/src/tracks/playabletrack/wavetrack/ui/ClipPitchAndSpeedButtonHandle.cpp#L46

MargaridaLopes12341:
Hey @kryksyh! 😊
Thanks for the feedback! I've just made the changes as requested — the clip header now displays only one digit after the decimal point when the speed value falls within the 99–101 range, following Audacity3's behavior.

Please have a look when you get a chance and let me know if anything else needs tweaking. Thanks again!

kryksyh (changes requested):
@MargaridaLopes12341 you are almost there! 
Looks good to me, please remove semicolons from qml\js

MargaridaLopes12341:
@kryksyh All done :)

MargaridaLopes12341:
Hey! 🙂
I noticed the code style error is showing up in a file I haven’t changed:

Error: Code style is incorrect in these files:
    src/app/app.cpp

Just wanted to check — is there anything I should do about this?

Thanks!

MargaridaLopes12341:
@kryksyh

PR #8633, opened 2025-04-10
Realtime effect panel info text

### LINKED ISSUE #8549: Text and video link on effect panel
Image

saintmatthieu:
Resolves: #8549

Recommended:

QA:
- [ ] behaves like in AU3

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectsSection.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectsSection.qml
@@ -16,14 +16,16 @@ Rectangle {
 
     property alias isMasterTrack: effectList.isMasterTrack
 
-    enabled: effectList.trackName !== ""
     color: ui.theme.backgroundPrimaryColor
 
-    readonly property int addEffectButtonHeight: 24
-    readonly property int addEffectButtonMargin: 12
-    readonly property int headerHeight: 40
-    readonly property int itemSpacing: 12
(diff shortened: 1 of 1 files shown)


src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectsSection.qml
```diff
@@ -122,5 +127,28 @@ Rectangle {
...
             }
         }
+
+        StyledTextLabel {
+            id: infoText
+            visible: !isMasterTrack && effectList.empty
+            text: {
+                {
+                    const line1 = qsTrc("projectscene", "Realtime effects are non-destructive and can be changed at any time.")
+                    const line2 = qsTrc("projectscene", "<a href=\"https://www.audacityteam.org/realtime-video\">Watch video</a>")
```
grliszas14, line 137: Will this hyperlink ever change? If yes, we could consider making it configurable

grliszas14 (approved):

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackEffectsSection.qml
saintmatthieu: It was hard-coded in AU3, too, so that doesn't seem to have been of concern to Vitaly and his reviewer(s) at the time.
I see the way MSS does it for e.g. `IAppShellConfiguration::museScoreUrl()`, but it looks more like this is to have it reused in several places ; down the line it also is hard-coded.
Or how would configurability be of advantage?

Issue #8591, opened 2025-04-02
`Cmd+(Fn)+F2 (Ctrl+F2)` shortcut to rename a clip doesn't work


dozzzzer:
`Cmd+(Fn)+F2 (Ctrl+F2)` shortcut to rename a clip doesn't cause the clip name to enter Edit mode when the clip is selected.

dozzzzer:
@Eism renaming via the shortcut `Cmd+(Fn)+F2/Ctrl+F2` works fine. Please change it so that `Fn` isn't required anymore. See the comment from @LWinterberg above.

kryksyh:
> Please change it so that Fn isn't required anymore.

is this even possible?

dozzzzer:
Apparently, it's not :(

Issue #8194, opened 2025-02-10
Place the shortcuts next to their corresponding shortcut menu items


dozzzzer:
### Description

In AU3, in the context menu, each option has a corresponding shortcut displayed next to it:

<img width="325" alt="Image" src="https://github.com/user-attachments/assets/f0448cfe-2c25-4ee3-9dee-44d0497f3b30" />

AU4 currently:

<img width="325" alt="Image" src="https://github.com/user-attachments/assets/cd5a0aa2-21eb-4d41-a558-a330bd222bf1" />

<img width="325" alt="Image" src="https://github.com/user-attachments/assets/15fcf19d-e8d3-4b49-b8cd-584e6784e748" />

dozzzzer:
Tested

PR #8646, opened 2025-04-11
Context menus shortcuts

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8194
Resolves: https://github.com/audacity/audacity/issues/8591

QA:
* added missing shortcuts to clip context menu: split, open pitch and speed dialog and rename clip
* fixed shortcut for renaming clip
* renaming clip or opening pitch and speed dialog via shortcut works only if single clip is selected

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
@@ -231,6 +231,7 @@ Item {
             id: clipComp
 
             ClipItem {
+                id: item
 
                 context: root.context
                 canvas: root.canvas
@@ -404,6 +405,10 @@ Item {
                     target: clipItem
                     function onWaveChanged() {
                         updateWave()
+                    }
+
+                    function onTitleEditRequested() {
+                        item.editTitle()
                     }
                 }
             }
--- a/src/projectscene/view/clipsview/clipslistmodel.h
+++ b/src/projectscene/view/clipsview/clipslistmodel.h
@@ -124,6 +124,7 @@ private:
     ClipListItem* itemByKey(const trackedit::ClipKey& k) const;
     int indexByKey(const trackedit::ClipKey& k) const;
     QVariant neighbor(const ClipKey& key, int offset) const;
+    void requestClipTitleChange();
 
     trackedit::secs_t calculateTimePositionOffset(const ClipListItem* item) const;
     int calculateTrackPositionOffset(const ClipKey& key, bool completed) const;
--- a/src/trackedit/internal/trackeditactionscontroller.cpp
+++ b/src/trackedit/internal/trackeditactionscontroller.cpp
@@ -45,6 +45,7 @@ static const ActionCode CLIP_DELETE_CODE("clip-delete");
 static const ActionCode MULTI_CLIP_DELETE_CODE("multi-clip-delete");
 static const ActionCode RANGE_SELECTION_DELETE_CODE("clip-delete-selected");
 
+static const ActionCode OPEN_CLIP_AND_SPEED_CODE("clip-pitch-speed-open");
 static const ActionCode CLIP_RENDER_PITCH_AND_SPEED_CODE("clip-render-pitch-speed");
 static const ActionCode TRACK_SPLIT("track-split");
 static const ActionCode TRACK_SPLIT_AT("track-split-at");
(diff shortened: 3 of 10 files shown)


src/projectscene/view/clipsview/clipslistmodel.h
```diff
@@ -67,6 +67,7 @@ class ClipsListModel : public QAbstractListModel, public muse::async::Asyncable,
 
     Q_INVOKABLE void selectClip(const ClipKey& key);
     Q_INVOKABLE void resetSelectedClips();
+    void requestClipTitleChange(const muse::actions::ActionData& args);
```
Eism, line 70: please move it to private section

src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
```diff
@@ -406,6 +407,15 @@ Item {
                         updateWave()
                     }
                 }
+
+                Connections {
+                    target: clipsModel
+                    onRequestClipTitleEdit: function(clipId) {
+                        if (clipItem.selected) {
```
Eism, line 414: could you please explain why you need this check?

src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
```diff
@@ -406,6 +407,15 @@ Item {
...
+                            item.titleEditRequested()
```
Eism, line 415: just call root.editTitle, no need to create a signal for it

src/trackedit/internal/trackedituiactions.cpp
```diff
@@ -281,6 +281,18 @@ const UiActionList TrackeditUiActions::m_actions = {
              TranslatableString("action", "Stretch with tempo changes"),
              Checkable::Yes
              ),
+    UiAction("clip-title-edit",
+             au::context::UiCtxAny,
+             au::context::CTX_ANY,
+             TranslatableString("action", "Rename clip"),
+             TranslatableString("action", "Rename clip")
+             ),
+    UiAction("open-clip-pitch-speed",
```
Eism, line 290: I thought we name actions using the pattern `<something>-<action>`

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -45,6 +45,8 @@ static const ActionCode CLIP_DELETE_CODE("clip-delete");
 static const ActionCode MULTI_CLIP_DELETE_CODE("multi-clip-delete");
 static const ActionCode RANGE_SELECTION_DELETE_CODE("clip-delete-selected");
 
+static const ActionCode CLIP_RENAME_CODE("rename-clip");
```
Eism, line 48: what is the difference between clip-title-edit and rename-clip?
can we leave only one of them?

src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
grliszas14: That's because every clip on a given track will be notified and only one of them (the one that's selected) needs to start title edit action

src/trackedit/internal/trackeditactionscontroller.cpp
grliszas14: rename-clip is global-like function - it gets triggered and does necessary selection checks (just like globalDelete for example) and then it re-triggers a "clip-title-edit" with a clipkey argument

Eism: I'm not sure we'll have another case for renaming a clip. Let's leave just one

src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
Eism: but you are calling "clip-title-edit" only for selected clips and also filter the processing by clipKey

```
dispatcher()->dispatch("clip-title-edit", ActionData::make_arg1<trackedit::ClipKey>(selectedClips.front()));
// and 
if (itemByKey(clipKey) != nullptr) {
    emit requestClipTitleEdit(clipKey.clipId);
}
```

grliszas14: What this filtering does is: prevent emitting signal for a model that doesn't contain clip we need -> we don't have a way to directly call that QML function any other way (unless I forgot about something)

src/trackedit/internal/trackeditactionscontroller.cpp
grliszas14: Ok, I left original "rename-clip" action

src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
Eism: Ok, I got you.
Let's then add a signal to clipItem (example onWaveChanged) and call it only for the selected clip.

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -1121,6 +1123,17 @@ void TrackeditActionsController::toggleStretchClipToMatchTempo(const ActionData&
     notifyActionCheckedChanged(STRETCH_ENABLED_CODE);
 }
 
+void TrackeditActionsController::globalOpenClipPitchAndSpeed()
```
Eism, line 1126: just openClipPitchAndSpeed

Eism (approved):

Issue #8606, opened 2025-04-04
Extract ITrackeditClipboard from Au3Interaction


saintmatthieu:
### Description

See description of #8603

dozzzzer:
@saintmatthieu 
3. Trimming/stretching a clip isn't undoable:

https://github.com/user-attachments/assets/c100fc14-82da-4275-af42-e6f1deef8678

saintmatthieu:
Thanks! Addressed.

dozzzzer:
@saintmatthieu 
4. When stereo-to-mono conversion is cancelled, often times the operation may not be cancelled properly:

https://github.com/user-attachments/assets/4f792dc1-6927-4c57-90c9-db8cf494de34

dozzzzer:
5. Seems like it might be the same issue as the previous one, but still: cancelling stereo-to-mono conversion via the dialog also causes the stereo clip to be converted to mono.

https://github.com/user-attachments/assets/d45fa1c9-76b5-4623-8678-817c5dc8c785

saintmatthieu:
Thanks @dozzzzer , should be fixed now 🤞

dozzzzer:
@saintmatthieu 

6. Have three tracks: mono, stereo, mono. Each with a clip. Move the topmost clip to track 3, and release the mouse button. Now, slowly move this clip from track 3 to track 2 (stereo). Observe: this cannot be done; the clip skips track 2 and is moved to track 1 (mono) instead.

https://github.com/user-attachments/assets/501e5510-8859-492c-a4b7-885aedeaa108

chinakov:
Issue by @dozzzzer has been addressed, re-checked the QA list items, everything seems to be in order now. Thanks!

PR #8650, opened 2025-04-11
8603 breakdown of au3interaction

### LINKED ISSUE #8604: Insert ITrackAndClipOperations
See #8603 description

### LINKED ISSUE #8605: Introduce IUndoManager
### Description

See description of #8603

saintmatthieu:
Resolves: #8604
Resolves: #8605

Commences work on #8606

Recommended:

QA:
Please test for regression (on one platform should be sufficient) :
- [x] Ripple-paste (Ctrl+Shift+V)
- [x] Operations with multiple-clip selection (cut, paste, etc)
- [x] Undo/Redo of clip operations

### DIFF
--- a/src/trackedit/tests/au3interaction_tests.cpp
+++ b/src/trackedit/tests/au3interaction_tests.cpp
@@ -4,6 +4,7 @@
 #include <gtest/gtest.h>
 
 #include "../internal/au3/au3interaction.h"
+#include "../internal/au3/au3trackdata.h"
 
 #include "context/tests/mocks/globalcontextmock.h"
 #include "context/tests/mocks/playbackstatemock.h"
@@ -894,9 +895,7 @@ TEST_F(Au3InteractionTests, CopyClipIntoClipboard)
 
     //! [EXPECT] The clipboard is notified about the new track data
     const ClipKey clipKey { track->GetId(), clip->GetId() };
-    EXPECT_CALL(*m_clipboard, addTrackData(Truly([&](const TrackData& data) {
-        return data.clipKey == clipKey;
-    }))).Times(1);
+    EXPECT_CALL(*m_clipboard, addTrackData(_)).Times(1);
 
     //! [WHEN] Copy the tracks into the clipboard
     m_au3Interaction->copyClipIntoClipboard(clipKey);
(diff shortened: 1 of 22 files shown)


src/trackedit/tests/au3interaction_tests.cpp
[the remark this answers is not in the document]
saintmatthieu: Oh! Didn't mean to commit that, sorry :D
Ninja fails compiling it, it needs some `/bigobj` compilation flag of some kind to be happy, apparently. That's been my remedy so far - erase everything.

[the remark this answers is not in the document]
saintmatthieu: As for the failing unit tests, that should be fixed now.

src/trackedit/trackeditmodule.cpp
```diff
@@ -56,7 +58,10 @@ void TrackeditModule::registerExports()
 
     ioc()->registerExport<ITrackeditActionsController>(moduleName(), m_trackeditController);
     ioc()->registerExport<ITrackeditProjectCreator>(moduleName(), new Au3TrackeditProjectCreator());
-    ioc()->registerExport<ITrackeditInteraction>(moduleName(), new TrackeditInteraction(std::make_unique<Au3Interaction>()));
+    ioc()->registerExport<ITrackAndClipOperations>(moduleName(), new Au3Interaction());
+    ioc()->registerExport<ITrackeditInteraction>(moduleName(),
+                                                 new TrackeditInteraction(std::make_unique<TrackeditOperationController>(
+                                                                              std::make_unique<UndoManager>())));
```
Eism, line 64: I think UndoManager is no different from other services and we can register it in ioc instead of storing it in the controller

saintmatthieu: In this case, we don't want the `UndoManager` to be used by clients other than the `TrackeditInteraction` (except in unit-testing), because the undo/redo interactions require playback stop, which `TrackeditInteraction` handles. Exposing it as a service, even internally to the trackedit module, might suggest otherwise.

grliszas14 (changes requested):
Some technicalities still to fix:
1. Clipboard mock, here you go:
```
namespace au::trackedit {
class ClipboardMock : public ITrackeditClipboard
{
public:

    MOCK_METHOD(std::vector<ITrackDataPtr>, trackDataCopy, (), (const, override));
    MOCK_METHOD(void, clearTrackData, (), (override));
    MOCK_METHOD(bool, trackDataEmpty, (), (const, override));
    MOCK_METHOD(size_t, trackDataSize, (), (const, override));
    MOCK_METHOD(void, addTrackData, (ITrackDataPtr), (override));

    MOCK_METHOD(void, setMultiSelectionCopy, (bool newValue), (override));
    MOCK_METHOD(bool, isMultiSelectionCopy, (), (const, override));
};
}
```
2. au3interaction_tests fail

besides this it looks good, big plus for removing dummyClipKey (at this point I don't remember why I had to introduce this).

This will require a solid copy&paste testing (including ripple editing) - I spotted some wrong paste behaviour but it turned out to be existing in master already

src/trackedit/trackeditmodule.cpp
saintmatthieu: Atm, the operation controller really is just a forwarder, but in a follow-up PR, logic in the Au3Interaction that consults the `selectionController()` to decide which action to take will be uplifted here. It probably won't be anything complex, yet it might still be something we want to test, in which case injecting an abstract undo manager may come in helpful.
But I hear your argument, and don't have a strong opinion here. If we need this abstraction later it'll be very easy to re-inject. Thumbs up to do as Elnur suggest, thumbs down to keep it that way :)

grliszas14 (approved):

Issue #8649, opened 2025-04-11
Fix header click adding 'move' history entry


grliszas14:
### Description

Currently clicking on clip header may add a "move" history entry.
Clip should not be moved until user tries to move it at least by 3px

dozzzzer:
Tested

PR #8658, opened 2025-04-14
Added clip move threshold

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8649

### DIFF
--- a/src/projectscene/view/clipsview/clipslistmodel.cpp
+++ b/src/projectscene/view/clipsview/clipslistmodel.cpp
@@ -17,6 +17,7 @@ constexpr int CACHE_BUFFER_PX = 200;
 constexpr double MOVE_MAX = 100000.0;
 constexpr double MOVE_MIN = 0.0;
 constexpr double MIN_CLIP_WIDTH = 3.0;
+constexpr double MOVE_THRESHOLD = 3.0;
 
 static const muse::Uri EDIT_PITCH_AND_SPEED_URI("audacity://projectscene/editpitchandspeed");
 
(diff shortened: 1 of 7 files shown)


src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -587,10 +589,19 @@ bool ClipsListModel::moveSelectedClips(const ClipKey& key, bool completed)
         return false;
     }
 
+    bool addToHistory = completed;
     secs_t timePositionOffset = calculateTimePositionOffset(item);
     int trackPositionOffset = calculateTrackPositionOffset(key, completed);
 
-    bool clipsMovedToOtherTrack = trackeditInteraction()->moveClips(timePositionOffset, trackPositionOffset, completed);
+    secs_t positionOffsetX = timePositionOffset * m_context->zoom();
+    if (!vs->moveInitiated() && (muse::RealIsEqualOrMore(std::abs(positionOffsetX), MOVE_THRESHOLD) || trackPositionOffset != 0)) {
```
saintmatthieu, line 597: I understand this part, but it gives me an unsafe feeling: is `timePositionOffset` reset to 0 before all other uses of it? Can `addToHistory` become `false` while `completed` is `true`, in which case, what would be the result of calling `moveClips` with `addToHistory==false` if `if ((completed && m_autoScrollConnection)) {` evaluates to `true` below? The clip dragging logic is already complex.

Wouldn't it be possible to prevent mouse movement higher up, like designing a `StickyMouseArea`, which would inherit `MouseArea` but would implement a mechanism that delays the firing of `positionChanged` signals until the movement has exceeded the threshold? Then `TracksClipsView.qml` would just have to use its `MouseArea` with the `StickyMouseArea`.

saintmatthieu (approved):

PR #8666, opened 2025-04-16
Static const error in Hamming window derivative

witwald:

Recommended:


kryksyh:
@witwald nice catch, thanks!

kryksyh (approved):

witwald:
@kryksyh I was pleased to be able to help.

PR #8670, opened 2025-04-16
VST3 realtime effects: force setting commit when closing dialog

### LINKED ISSUE #8640: VST effect's parameters aren't remembered in the Realtime Effects panel
### Bug description

_No response_

### Steps to reproduce

1. Add a VST plugin to the realtime effects stack
2. Tweak some parameters
3. Close its GUI
4. Open it
5. Check the parameters

### Expected behavior

The parameters remain the same as you've set them.

### Actual behavior

The parameters got reset to default.

https://github.com/user-attachments/assets/e32ae73e-6087-4b20-b096-15f783a93c0b

### Audacity Version

master

### Operating system

OS independent

### Additional context

_No response_

saintmatthieu:
Resolves: #8640

Recommended:

### DIFF
--- a/src/effects/vst/view/vstviewmodel.cpp
+++ b/src/effects/vst/view/vstviewmodel.cpp
@@ -16,7 +16,8 @@ using namespace au::effects;
 VstViewModel::~VstViewModel()
 {
     m_settingUpdateTimer.stop();
-    QObject::disconnect(&m_settingUpdateTimer, &QTimer::timeout, this, &VstViewModel::checkSettingChangesFromUi);
+    QObject::disconnect(&m_settingUpdateTimer, &QTimer::timeout, this, &VstViewModel::checkSettingChangesFromUiWhileIdle);
+    checkSettingChangesFromUi(true);
 }
 
 void VstViewModel::init()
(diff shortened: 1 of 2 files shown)


src/effects/vst/view/vstviewmodel.cpp
```diff
@@ -56,31 +57,35 @@ void VstViewModel::init()
 
     settingsToView();
 
-    QObject::connect(&m_settingUpdateTimer, &QTimer::timeout, this, &VstViewModel::checkSettingChangesFromUi);
+    QObject::connect(&m_settingUpdateTimer, &QTimer::timeout, this, &VstViewModel::checkSettingChangesFromUiWhileIdle);
```
kryksyh, line 60: VstViewModel is a QObject, so there is no need to explicitly call for 
> QObject::connect

src/effects/vst/view/vstviewmodel.cpp
```diff
@@ -56,31 +57,35 @@ void VstViewModel::init()
...
         wrapper.FlushParameters(settings, &hasChanges);
-        if (hasChanges) {
+        if (hasChanges || forceCommitting) {
             wrapper.StoreSettings(settings);
         }
         return nullptr;
     });
 
-    if (hasChanges) {
+    if (hasChanges || forceCommitting) {
```
kryksyh, line 88: Why do we need to force committing in destructor if there are no changes?

kryksyh (approved):

src/effects/vst/view/vstviewmodel.cpp
saintmatthieu: There may or may not be changes, but if these happened during playback, for some reason I haven't investigated, `hasChanges` never gets set to `true`.
I stupidly replicated the behavior in Audacity 3, which calls `StoreSettings` on dialog close.

src/effects/vst/view/vstviewmodel.cpp
saintmatthieu: Yeah, sure. On the other hand, it is a static method, so I don't think it harms intelligibility either.

Issue #7263, opened 2024-09-05
[AU4] Clips - tabbing


ianalarina:
### Description

- Tabbing between clips
- Tabbing Inside clips

[Figma link](https://www.figma.com/design/01XLIICmy2AG6t1ieyTp66/02---Clips?node-id=521-68168)

[Prototype link](https://audacity-tabbing-prototype.netlify.app/)

grliszas14:
QA:
* verify if you can get to the clips via tabbing and then navigate through visible clips via keyboard arrows
* pressing [enter] on a focused clip should allow to navigate through clip title, 3dots menu and select button (and additionally trim/stretch handles if the clip is selected)
* press [enter] when focused on a trim handle to activate edit mode. Pressing [enter] again or [Esc] exits edit mode
* carefully test if trimming / time-stretching works both for keyboard navigation and mouse navigation (for every handle separately)
* if clip is trimmed/stretched via mouse, history entry is added on every mouse release. If clip is trimmed via keyboard arrows, history entries are consolidated into one
* if you navigate through clips down to the bottom of the view, the view should scroll automatically
* check if horizontal auto-scroll is not broken

Focus on clipsView only for now. The goal of this task is to introduce foundation for clips accessibility so we can iterate over in the upcoming tasks. The most important thing is to make sure it doesn't break any existing features

DilsonsPickles:
@chinakov - Yes the tab focus for clips definitely needs some work. I'll propose two possible designs in our next design team meeting. 

@ianalarina @grliszas14 - I noticed the spec linked to this ticket is quite old, do you want me to update it, or are you happy to work from the prototype? (I'll add a link to the prototype in the ticket description).

ianalarina:
@DilsonsPickles add both if you can, I thought you only have a prototype

PR #8692, opened 2025-04-23
Clips accessibility

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7263

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipHandles.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipHandles.qml
@@ -3,6 +3,8 @@ import QtQuick
 import Muse.Ui
 import Muse.UiComponents
 
+import Audacity.ProjectScene
+
 Item {
     id: root
 
@@ -14,6 +16,12 @@ Item {
 
     property bool debugRectsVisible: false
 
+    property NavigationPanel clipNavigationPanel: null
+    property bool leftTrimActive: false
+    property bool rightTrimActive: false
+    property bool leftStretchActive: false
+    property bool rightStretchActive: false
+
     // mouse position event is not propagated on overlapping mouse areas
     // so we are handling it manually
     signal clipHandlesMousePositionChanged(real x, real y)
@@ -23,11 +31,11 @@ Item {
     signal clipStartEditRequested()
     signal clipEndEditRequested()
 
-    signal trimLeftRequested(bool completed)
-    signal trimRightRequested(bool completed)
+    signal trimLeftRequested(bool completed, int action)
+    signal trimRightRequested(bool completed, int action)
 
-    signal stretchLeftRequested(bool completed)
-    signal stretchRightRequested(bool completed)
+    signal stretchLeftRequested(bool completed, int action)
+    signal stretchRightRequested(bool completed, int action)
 
     //! NOTE: auto-scroll for trimming is triggered from clipslistmodel
     signal stopAutoScroll()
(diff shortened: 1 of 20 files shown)


src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -468,6 +468,34 @@ int ClipsListModel::calculateTrackPositionOffset(const ClipKey& key, bool comple
...
+                                      bool completed,
+                                      const std::function<void()>& onAutoScrollFrame)
+{
+    auto vs = globalContext()->currentProject()->viewState();
+    if (!vs) {
+        return;
+    }
+
+    // do not handle auto-scroll when using key-nav
+    if (muse::RealIsEqual(vs->clipEditStartTimeOffset(), -1.0)) {
```
kryksyh, line 481: Do I understand the logic here: `clipEditStartTimeOffset` is set in `clipStartEditRequested`, and that is called only on mouse move events? Otherwise `clipEditStartTimeOffset` will have its default value?

src/projectscene/view/clipsview/clipslistmodel.cpp
```diff
@@ -649,9 +677,32 @@ bool ClipsListModel::trimLeftClip(const ClipKey& key, bool completed)
         return false;
     }
 
-    double newStartTime = m_context->mousePositionTime() - vs->clipEditStartTimeOffset();
+    UndoPushType undoType;
+    double newStartTime;
+    if (muse::RealIsEqual(vs->clipEditStartTimeOffset(), -1.0)) {
```
kryksyh, line 682: Would this mean 
"if (trimWithKeyboard)"?

src/projectscene/view/clipsview/clipslistmodel.cpp
grliszas14: Yup, exactly

src/projectscene/view/clipsview/clipslistmodel.cpp
grliszas14: Yes 👍 maybe I should put it in variable then maybe so it's quicker to grasp?

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipHandles.qml
```diff
@@ -257,6 +373,68 @@ Item {
...
+
+                onActiveChanged: {
+                    if (!active) {
+                        root.leftStretchActive = false
+                    }
+                }
+            }
+
+            Rectangle {
+                x: -9
```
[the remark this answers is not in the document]
grliszas14: I tested it by trial and method until it was aligned perfectly from both sides :P

src/projectscene/qml/Audacity/ProjectScene/clipsview/ClipHandles.qml
```diff
@@ -91,7 +153,7 @@ Item {
             }
 
             onReleased: function(e) {
-                root.trimLeftRequested(true)
+                root.trimLeftRequested(true , ClipBoundaryAction.Shrink)
```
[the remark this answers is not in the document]
grliszas14: Yeah, that is the case, it's not used when triggered via mouse, good idea with additional enum 👍

src/projectscene/view/clipsview/clipslistmodel.cpp
kryksyh: Yeah, that would be great, you can add a small function even, because currently this comparison doesn't make much sense if you don't fully realize what's going on.

kryksyh (approved):

PR #8724, opened 2025-05-02
Add navigation to destructive and realtime-effect dialogs

### LINKED ISSUE #8107: After a preset is loaded, the effect dialog cannot be dismissed by pressing Esc
### Bug description

_No response_

### Steps to reproduce

1. Select some audio
2. Effect > Reverb
3. Select a preset (e.g. Bathroom)
4. Press Esc

### Expected behavior

The Reverb dialog closes.

### Actual behavior

The Reverb dialog doesn't close.

### Audacity Version

master

### Operating system

OS independent

### Additional context

_No response_

### LINKED ISSUE #8513: Allow closing effect's dialog by pressing the Esc key
### Problem to be solved

The only way to close the effect's dialog is to click on the 'Close' button in the top left-hand corner. This is slightly annoying.

### Your idea

To allow the effect's dialog to be closed by pressing Esc.

### Prior art

_No response_

### Additional context

Already exists in Musescore 4.

### LINKED ISSUE #8521: Effect's presets & settings are inaccessible on MacOS
On MacOS, when clicking Presets & Settings in both realtime & destructive effects, it seems to open up the menu underneath the plugin's UI, which means you don't get to see it or interact with it. Windows and Linux work fine.

Image

saintmatthieu:
Resolves: #8521
Resolves: #8107
Resolves: #8513

Adding navigation solves many problems - beside paving the way for proper accessibility.

Recommended:


src/effects/effects_base/qml/Audacity/Effects/EffectPresetsBar.qml
```diff
@@ -61,6 +66,9 @@ RowLayout {
     }
 
     FlatButton {
+        navigation.panel: root.navigationPanel
```
grliszas14, line 69: Please, try to chain these whenever you can - then in any case if we need to adjust UI, add new button or other accessibility element -> we'll need to make less changes (changing two navigation elements instead of assigning all of the items with a new number)

        
```qml
id: saveBtn
navigation.order: presetSelector.navigation.order + 1
```

src/effects/effects_base/qml/Audacity/Effects/EffectPresetsBar.qml
```diff
@@ -70,6 +78,9 @@ RowLayout {
...
```
grliszas14, line 81:
```qml
id: undoBtn
navigation.order: saveBtn.navigation.order + 1
``` 

etc.

grliszas14 (commented):
Can we add navigation for the bottom row of buttons by the way? (preview, cancel, apply)

DavidBailes:
> Can we add navigation for the bottom row of buttons by the way? (preview, cancel, apply)

@saintmatthieu @LWinterberg . I don't know what the current design is for the keyboard navigation of this row of buttons at the bottom of the dialog, but this part of Microsoft's guidelines is relevant:
https://learn.microsoft.com/en-us/windows/apps/design/input/keyboard-interactions#use-multiple-tab-stops-and-arrow-keys-for-buttons

An example of this behaviour is demonstrated in the winui 3 gallery app (https://apps.microsoft.com/detail/9p3jfpwwdzrc?hl=en-US&gl=US). In that app, Dialogs & flyouts > Content dialog.

saintmatthieu:
@DavidBailes thanks for your input. I believe that @DilsonsPickles is the one in charge of this spec.

saintmatthieu:
> Can we add navigation for the bottom row of buttons by the way? (preview, cancel, apply)

I didn't manage on the first attempt and didn't investigate why it didn't work.
Correct accessibility is a stretch goal to this ticket - now it was made even clearer by @DavidBailes 's comment that a design will be required beforehand anyway.

grliszas14 (approved):

Issue #8659, opened 2025-04-14
[Plugin issue] Some plugins have their paths displayed instead of names


dozzzzer:
### Plugin

MuseFX Reverb

### Plugin Format

VST3

### Steps to reproduce

1. Ensure MuseFX Reverb is installed
2. Launch Audacity and expand the list of effects 

### Expected behavior

The plugin is listed as "Reverb" 

### Actual behavior

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/e7ab3fb9-cc75-4c3c-ba2c-94a0c89b4a19" />

### Audacity Version

current master / alpha / nightly build

### Operating system

macOS

### Architecture

Apple Silicon (M1, M2, ...)

chinakov:
I can also reproduce this on the latest Master: efdaf1e91b151bcc5fe37cab8c0d1539afaf569c

PR #8729, opened 2025-05-05
Only disambiguate effects of the same family, type, category and title

kryksyh:
Resolves: https://github.com/audacity/audacity/issues/8659

Do not disambiguate effects that fall into different submenus

Recommended:

### DIFF
--- a/src/effects/effects_base/internal/effectsuiactions.cpp
+++ b/src/effects/effects_base/internal/effectsuiactions.cpp
@@ -82,10 +82,14 @@ UiAction makeUiAction(const char16_t* uri, const EffectMeta& meta)
 // To mitigate this, we replace the title with the path of the plugin.
 void replaceIdenticalTitlesWithPaths(EffectMetaList& effects)
 {
-    std::unordered_map<muse::String, std::vector<size_t> > duplicateMap;
+    using DuplicationKey = std::tuple<EffectFamily, EffectType, muse::String, muse::String>;
+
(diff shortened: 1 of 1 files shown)


src/effects/effects_base/internal/effectsuiactions.cpp
```diff
@@ -82,10 +82,14 @@ UiAction makeUiAction(const char16_t* uri, const EffectMeta& meta)
 // To mitigate this, we replace the title with the path of the plugin.
 void replaceIdenticalTitlesWithPaths(EffectMetaList& effects)
 {
-    std::unordered_map<muse::String, std::vector<size_t> > duplicateMap;
+    using DuplicationKey = std::tuple<EffectFamily, EffectType, muse::String, muse::String>;
+
+    std::map<DuplicationKey, std::vector<size_t> > duplicateMap;
```
saintmatthieu, line 87: That doesn't work with an `unordered_map` ? Looks like some music folks can have several hundreds of plugins.

saintmatthieu (approved):

src/effects/effects_base/internal/effectsuiactions.cpp
kryksyh: For the simplicity sake I chose to just use std::map (for a hash map we need to define a hasher). 

It runs just once on startup, and the performance should be really close:

https://quick-bench.com/q/tvCbVBF0hGuCgzhaCDzwRIbGk7s

Issue #8716, opened 2025-04-30
Range selection split is undone as 2 separate point split operations


dozzzzer:
### Steps to reproduce

1. Have a clip
2. Make a selection
3. Split (Ctrl + I)
4. Undo
5. Observe

### Expected behavior

No split lines after Undo. 

### Actual behavior

Only one of the split lines is removed. Repeated Undo would remove the second split line. In other words, the Split operation on a selection is treated as a sequence of two separate point Split operations.

https://github.com/user-attachments/assets/db177e84-1647-4f21-a0c0-97962c157d5a

 

### Audacity Version

master

### Operating system

OS independent

chinakov:
This is fixed, thanks!

PR #8738, opened 2025-05-06
Fix range split being undone as 2 separate actions

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8716

### DIFF
--- a/src/trackedit/tests/au3interaction_tests.cpp
+++ b/src/trackedit/tests/au3interaction_tests.cpp
@@ -1676,7 +1676,7 @@ TEST_F(Au3InteractionTests, SplitTracksAtEmptyList)
     EXPECT_CALL(*m_trackEditProject, notifyAboutTrackAdded(_)).Times(0);
 
     //! [WHEN] Split an empty list
-    m_au3Interaction->splitTracksAt({}, 0.0);
+    m_au3Interaction->splitTracksAt({}, { 0.0 });
 }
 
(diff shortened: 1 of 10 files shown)


src/trackedit/tests/au3interaction_tests.cpp
```diff
@@ -1715,6 +1715,32 @@ TEST_F(Au3InteractionTests, SplitTracksOnClipData)
     removeTrack(trackTwoClipsId);
 }
 
+TEST_F(Au3InteractionTests, RangeSplitTracks)
```
embarc-gabriel, line 1718: Nit: To make the test's purpose clearer, perhaps create a track with only one clip.
In SplitTrackOnClipData, while two-clip tracks were used, the subsequent check focused on whether the split operation on the first clip affected the second.
In this test, on the other hand,  the second clip isn't relevant to the test's outcome.
This is a minor point, though, and I'm okay with the current implementation if it's merged. :)

embarc-gabriel (approved):

src/trackedit/tests/au3interaction_tests.cpp
grliszas14: Yeah, I just wanted to cover one more case with as little effort possible :P

embarc-gabriel: Yeah! I HIGHLY appreciate the effort to keep the coverage! :)

PR #8741, opened 2025-05-07
8706 Track meter while recording

### LINKED ISSUE #8706: Track header - Track meter visuals during recording
### Description

Implement track meter on recording

embarc-gabriel:
Resolves: #8706

Implement track meters feature on recording.

Recommended:

QA:
- [x] Please test with a stereo recording device with contrasting left/right level (maybe this can be emulated with a virtual device)


au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3061,13 +3062,47 @@ void AudioIoCallback::SendVuOutputMeterData(
...
+        }
+    }
+
+    constexpr size_t maxMainTrackChannels = 2;
+    auto mainTrackInput = stackAllocate(float, maxMainTrackChannels * frames);
+    for (size_t i = 0; i < maxMainTrackChannels * frames; ++i) {
+        mainTrackInput[i] = 0.0f;
+    }
+
+    sptr = values;
```
embarc-gabriel, line 3083: Maybe I could send track meter input values and increment the main track input in a single step but I opted out for consistency on the push method call.
Apart from that, I decided to move the peak calculation to the "channel" itself.
It can be useful once in the future we would add rms values and this will also be encapsulated and transparent to the AudioIo module.

src/au3wrap/internal/au3audiometer.cpp
```diff
@@ -5,16 +5,30 @@
...
 
-#include "au3audiooutmeter.h"
+#include "au3audiometer.h"
 
 #include "libraries/lib-utility/MemoryX.h"
 
-namespace au::playback {
-OutMeter::OutMeter()
+namespace {
+float GetAbsValue(const float* buffer, size_t frames, size_t step)
```
embarc-gabriel, line 13: The step argument is something I would improve on a refactor.
The problem is how the memory is organized on the buffers.
Sometimes we have continuous memory segment per channel on other situations the channel data is interleaved in a memory segment.
But as I said before the idea now is not over modify this file to avoid regressions.

src/record/CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/../au3wrap/au3defs.cmake)
 
 set(MODULE_INCLUDE ${AU3_INCLUDE})
 set(MODULE_DEF ${AU3_DEF})
-set(MODULE_LINK au3wrap playback)
+set(MODULE_LINK au3wrap)
```
embarc-gabriel, line 43: @saintmatthieu I have moved the meter code to au3wrap. This way I could remove the dependency I have added on the previous PR.

au3/libraries/lib-audio-devices/AudioIOBase.h
```diff
@@ -54,7 +54,7 @@ struct AudioIOStartStreamOptions
     {}
 
     std::shared_ptr<AudacityProject> pProject;
-    std::weak_ptr<Meter> captureMeter;
+    std::weak_ptr<IMeterChannel> captureMeter;
```
saintmatthieu, line 57: Is `Meter` still used somewhere? Maybe it's time to get rid of it?

au3/libraries/lib-audio-devices/IMeterChannel.h
```diff
@@ -8,8 +8,16 @@ constexpr int64_t MASTER_TRACK_ID = -2;
 class AUDIO_DEVICES_API IMeterChannel
 {
 public:
+    struct SampleData
+    {
+        const float* buffer;
```
saintmatthieu, line 13: You could even `const` the pointer itself, `const float* const`

au3/libraries/lib-audio-devices/IMeterChannel.h
```diff
@@ -8,8 +8,16 @@ constexpr int64_t MASTER_TRACK_ID = -2;
...
+        const size_t frames;
+        const size_t step;
+    };
+
     virtual ~IMeterChannel() = default;
     virtual void push(uint8_t channel, float signal, int64_t key = MASTER_TRACK_ID) = 0;
+    virtual void push(uint8_t channel, const SampleData&& sampleData, int64_t key = MASTER_TRACK_ID) = 0;
```
saintmatthieu, line 20: A const right reference? Can't remember having seen this before. Isn't `const SampleData&` sufficient?

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3061,13 +3062,47 @@ void AudioIoCallback::SendVuOutputMeterData(
...
+        const int64_t id = sequence->GetRecordableSequenceId();
+        for (size_t ch = 0; ch < nChannels; ch++) {
+            channel->push(ch, {sptr, frames, 1}, id);
+            sptr += frames;
+        }
+    }
+
+    constexpr size_t maxMainTrackChannels = 2;
+    auto mainTrackInput = stackAllocate(float, maxMainTrackChannels * frames);
+    for (size_t i = 0; i < maxMainTrackChannels * frames; ++i) {
```
saintmatthieu, line 3079: Are you sure this is as efficient as calling `std::memset` ?

saintmatthieu (commented):
Publishing my comments for now, until the treatment of the main meter in multi-track recording situation is clarified.

au3/libraries/lib-audio-devices/AudioIOBase.h
embarc-gabriel: Removed.

au3/libraries/lib-audio-devices/IMeterChannel.h
embarc-gabriel: Done!

au3/libraries/lib-audio-devices/IMeterChannel.h
embarc-gabriel: Done!

au3/libraries/lib-audio-io/AudioIO.cpp
embarc-gabriel: Done!

au3/libraries/lib-audio-devices/IMeterChannel.h
```diff
@@ -8,8 +8,16 @@ constexpr int64_t MASTER_TRACK_ID = -2;
 class AUDIO_DEVICES_API IMeterChannel
```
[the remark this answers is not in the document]
embarc-gabriel: Done!

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3061,18 +3062,50 @@ void AudioIoCallback::SendVuOutputMeterData(
...
+    auto mainTrackInput = stackAllocate(float, maxMainTrackChannels * frames);
+    std::memset(mainTrackInput, 0, maxMainTrackChannels * frames * sizeof(float));
+
+    sptr = values;
+    for (const auto sequence : mCaptureSequences)
+    {
+        size_t nChannels = std::min(sequence->NChannels(), maxMainTrackChannels);
+        for(size_t ch = 0; ch < nChannels; ++ch) {
+            for (size_t i = 0; i < frames; ++i) {
+                mainTrackInput[ch * frames + i] += sptr[i];
```
saintmatthieu, line 3087: It looks like you are treating `mainTrackInput` as deinterleaved audio. If I trace back the origin of `values`, though, this is the buffer that is passed by the PortAudio callback, which by default is interleaved. Indeed, following this buffer to the place where it goes into the recorded track, I see
image

This would explain why, when I record a stereo signal that has energy on just one channel, the meter shows both channels with the same level:
image

embarc-gabriel: This should be fixed now.

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3075,23 +3080,35 @@ void AudioIoCallback::PushInputMeterValues(const std::shared_ptr<IMeterSender>&
...
-            for (size_t i = 0; i < frames; ++i) {
-                mainTrackInput[ch * frames + i] += sptr[i];
+
+    // Update main meter
+    // If the input source has more than 2 channels it will be splitted on multiple mono sequences
+    if (mCaptureSequences.size() == 1) {
+        const auto nChannels = mCaptureSequences[0]->NChannels();
+        assert(nChannels <= 2);
+        for (size_t ch = 0; ch < nChannels; ++ch) {
+            sender->push(ch, {sptr + ch, frames, nChannels});
```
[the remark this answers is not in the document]
embarc-gabriel: @saintmatthieu 
If you take a closer look we have a step parameter that is set to nChannels and this is passed to GetAbsValue.
If I have a single channel ch = 0 and step is also 1.
If I have two channels ch = 0 and step is 2 on the first iteration and ch = 1 and step is 2 on the second iteration so it will handle the interleave.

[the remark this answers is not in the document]
embarc-gabriel: Yes! I think so! I will update the PR.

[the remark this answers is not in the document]
embarc-gabriel: Fixed.

saintmatthieu (approved):

saintmatthieu:
Great, @embarc-gabriel , let's hand this one over to QA :)

PR #8746, opened 2025-05-08
Welcome screen carousel draft

grliszas14:
Work in progress


libraries/lib-preferences/Prefs.h
```diff
@@ -702,5 +702,6 @@ struct PREFERENCES_API PreferenceInitializer {
 extern PREFERENCES_API StickySetting<BoolSetting> DefaultUpdatesCheckingFlag;
 extern PREFERENCES_API StickySetting<BoolSetting> SendAnonymousUsageInfo;
 extern PREFERENCES_API StickySetting<StringSetting> InstanceId;
+extern PREFERENCES_API StickySetting<IntSetting> IntroOrderStart;
```
kryksyh, line 705: I don't think it has to be `sticky` setting, we could safely reset it to the default on settings reset.

kryksyh: We need `StickySetting` just for a handful of settings that should stay even if the user resets all settings to default.
For example, we do not want to reset the InstanceId or show the Usage Info agreement window again.

libraries/lib-wx-wrappers/ImageCarousel.h
```diff
@@ -0,0 +1,58 @@
...
+   
+   void DrawTitle(wxDC& dc, const wxSize& size);
+   void PositionButtons(int x, int y, const wxBitmap& bmp);
+   void DrawDots(wxDC& dc, const wxSize& size);
+   
+   wxWindowID m_id;
+   std::vector<CarouselSnapshot> m_snapshots;
+   size_t m_currentIndex = 0;
+   
+   ArrowButton* m_btnLeft;
```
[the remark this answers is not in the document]
grliszas14: I would rather not, I think it's not recommended to use smart ptrs with wxWidgets (at least with UI classes), we can get some undefined behaviors but... you made me realize that I have to call m_btnLeft->Destroy() to schedule object delete instead of using raw delete 👍

libraries/lib-wx-wrappers/ArrowButton.cpp
```diff
@@ -0,0 +1,62 @@
+/**********************************************************************
+
+  Audacity: A Digital Audio Editor
+
+  ImageCarousel.cpp
```
kryksyh, line 5: I vote to remove file names from the header :)

kryksyh (approved):

Issue #8607, opened 2025-04-04
Use Interval type instead of begin and end arguments


saintmatthieu:
Maybe reuse the one from the framework.

IliasBergstromMuse:
Maybe make a "code cleanup" task with subtasks instead, for tickets like this and #8608?

PR #8749, opened 2025-05-09
8606 extract clipboard and project history from au3interaction

### LINKED ISSUE #8606: Extract ITrackeditClipboard from Au3Interaction

saintmatthieu:
Resolves: #8606

Recommended:

QA:
Please a complete regression of editing operations of clip-and-track editing: operations should behave as before (ie. if not-working was the behavior before, do not expect it to work now) with respect to result and undoability.

Here is a list I started putting together but not completed, because it's tedious and I realized you might already have a regression plan. Feel free to use it nevertheless:
- [x] Dragging clips across tracks: please retest checklist of #8346
- [x] trim left
- [x] trim right
- [x] stretch left
- [x] stretch right
- [x] delete selection (Del)
- [x] Edit > Clip > Split
- [x] Edit > Clip > Split into new track
- [x] Edit > Clip > Merge selected clips
- [x] Edit > Cut (then paste)

### DIFF
--- a/src/trackedit/internal/au3/au3interaction.cpp
+++ b/src/trackedit/internal/au3/au3interaction.cpp
@@ -922,41 +922,35 @@ bool Au3Interaction::changeClipOptimizeForVoice(const ClipKey& clipKey, bool opt
     return true;
 }
 
-bool Au3Interaction::renderClipPitchAndSpeed(const ClipKey& clipKey, std::function<void()> onComplete)
+bool Au3Interaction::renderClipPitchAndSpeed(const ClipKey& clipKey)
 {
-    muse::Concurrent::run([this, clipKey, onComplete]() {
-        m_progress->start();
+    m_progress->start();
 
-        muse::ProgressResult result;
+    muse::ProgressResult result;
 
-        DEFER {
-            m_progress->finish(result);
-        };
+    DEFER {
+        m_progress->finish(result);
+    };
 
-        WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), ::TrackId(clipKey.trackId));
-        IF_ASSERT_FAILED(waveTrack) {
-            return;
-        }
+    WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), ::TrackId(clipKey.trackId));
+    IF_ASSERT_FAILED(waveTrack) {
+        return false;
+    }
 
-        std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.clipId);
-        IF_ASSERT_FAILED(clip) {
-            return;
-        }
+    std::shared_ptr<WaveClip> clip = DomAccessor::findWaveClip(waveTrack, clipKey.clipId);
+    IF_ASSERT_FAILED(clip) {
+        return false;
+    }
 
-        auto progressCallBack = [this](double progressFraction) {
-            m_progress->progress(progressFraction * 1000, 1000, "");
-        };
+    auto progressCallBack = [this](double progressFraction) {
+        m_progress->progress(progressFraction * 1000, 1000, "");
+    };
 
-        waveTrack->ApplyPitchAndSpeed({ { clip->GetPlayStartTime(), clip->GetPlayEndTime() } }, progressCallBack);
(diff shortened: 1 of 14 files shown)


src/trackedit/internal/trackeditoperationcontroller.cpp
```diff
@@ -336,4 +543,28 @@ muse::ProgressPtr TrackeditOperationController::progress() const
...
+void TrackeditOperationController::pushProjectHistoryJoinState(secs_t start, secs_t duration)
+{
+    std::stringstream ss;
+    ss << "Joined " << duration << " seconds at " << start;
+    projectHistory()->pushHistoryState(ss.str(), "Join");
+}
+
+void TrackeditOperationController::pushProjectHistoryDuplicateState()
+{
+    projectHistory()->pushHistoryState("Duplicated", "Duplicate");
```
embarc-gabriel, line 556: IMO It lacks a little bit of consistency here. Should we wrap the pushHistoryState call on a class method or not? Maybe we can only create methods if there is some argument and there is some data manipulation before calling the pushHistoryState function.

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1242,28 +1167,21 @@ bool Au3Interaction::copyNonContinuousTrackDataIntoClipboard(const TrackId track
...
 
     auto track = waveTrack->Copy(begin, end);
-    clipboard()->addTrackData(std::make_shared<Au3TrackData>(std::move(track)));
-
-    return true;
+    return std::make_shared<Au3TrackData>(std::move(track));
 }
 
-bool Au3Interaction::removeClip(const trackedit::ClipKey& clipKey)
+bool Au3Interaction::removeClip(const trackedit::ClipKey& clipKey, secs_t& begin, secs_t& end)
```
embarc-gabriel, line 1184: Nit. I am not much into returning data using input parameters.
I personally prefer something like returning an optional with the values.
If this is none nothing to do, if this is some we retrieve the value and do something.

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1731,20 +1631,20 @@ bool Au3Interaction::duplicateSelectedOnTrack(const TrackId trackId, secs_t begi
...
 {
     Au3WaveTrack* waveTrack = DomAccessor::findWaveTrack(projectRef(), Au3TrackId(trackId));
     IF_ASSERT_FAILED(waveTrack) {
-        return false;
+        return nullptr;
     }
 
     auto track = waveTrack->SplitCut(begin, end);
-    clipboard()->addTrackData(std::make_shared<Au3TrackData>(std::move(track)));
+    auto data = std::make_shared<Au3TrackData>(std::move(track));
```
embarc-gabriel, line 1642: const?

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1944,19 +1829,10 @@ bool Au3Interaction::trimClipLeft(const ClipKey& clipKey, secs_t deltaSec, secs_
     //! NOTE: don't be tempted to early return if delta is 0.0 or by any other reason:
```
embarc-gabriel, line 1829: Interesting that the old code did not follow this advice. Maybe we have removed some bug here.

embarc-gabriel (approved):

src/trackedit/internal/trackeditoperationcontroller.cpp
saintmatthieu: Two priorities come into play, by order of importance:
1. make the code robust,
2. avoid unnecessary effort :)

Accordingly, when `pushHistoryState` is to be called with the same set of strings from different places, then I wrap it in a method, else, just call `pushHistoryState` directly.

src/trackedit/internal/au3/au3interaction.cpp
saintmatthieu: I agree. The main reason was I didn't want to return an optional tuple. But since this is a refactoring PR and we were talking about the good it would do to have a time interval class, I just pushed a commit where I introduce `TimeSpan` and use it for this method. Please let me know if this meets the requirements you raised last time.

embarc-gabriel: Much better now! Yes! Thank you! I will take a look later but probably we can close #8607 as well.

Issue #3250, opened 2022-07-18
Wrong sorting order of the entries on the plugin list


dozzzzer:
**Describe the bug**
When browsing the list of plugins, firstly the ones titled in uppercase are listed, and then the go the ones in lowercase.

**To Reproduce**
0. Make sure you have at least a few plugins having lowercase title (e.g iZotope, bx_2000 etc.)
1. Go to Effect menu and see the full list of the plugins in an alphabetical order

**Actual behavior**
At first the ones with the sentence/upper case are shown and then go the lowercase ones 

**Expected behavior**
The case of the title of the plugin should be ignored

**Screenshots**

https://user-images.githubusercontent.com/70063970/179538077-1f8b7b8f-df0c-4457-961b-ee46c6688f59.mp4

LWinterberg:
![Image](https://github.com/user-attachments/assets/4875d1c1-3d40-430e-af8e-d0ab2bd9f608)

I also can reproduce this in Audacity 4.

dozzzzer:
It's now only reproducible in AU3

PR #8752, opened 2025-05-09
Fix audacity#3250: Wrong sorting order of the entries on the plugin list

FranciscoEpifanio:
When browsing the list of plugins, firstly the ones titled in uppercase are listed, and then the ones in lowercase.

The expected behavior is for the case of the title of the plugin to be ignored.

I modified the MenuHelper.cpp file to solve this
issue with the following changes:

Implemented sortGroupItems lambda that sorts TranslatableStrings by lowercase comparison
Uses index-based sorting to maintain correlation between names, IDs and flag vectors
Preserves the original group structure while
ordering items alphabetically

Added clear section comments for sorting and group handling logic Better variable naming (sortedNames vs names)

Effects are now alphabetically ordered within their respective groups Group hierarchy (Publisher/Type/TypePublisher) remains unchanged Case-insensitive comparison ensures consistent
ordering (e.g. "Chorus" and "chorus")

Resolves: #3250

Recommended:

### DIFF
--- a/src/menus/MenuHelper.cpp
+++ b/src/menus/MenuHelper.cpp
@@ -330,23 +330,55 @@ void AddGroupedEffectMenuItems(
    PluginIDs group;
    std::vector<CommandFlag> flags;
 
+   // Sorts items within a group alphabetically (case insensitive)
+   auto sortGroupItems = [&]() 
+   {
+      if(names.size() < 2) return;
+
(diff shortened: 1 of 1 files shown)


kryksyh:
@FranciscoEpifanio I'm closing this PR for now. If you still want to work on the issue, please make the PR into `audacity3` branch.

PR #8770, opened 2025-05-13
Use file filter for both *.srt and *.txt for Import Label File dialog

coezbek:
Resolves:

Currently, 'txt' files are preselected when importing label tracks (File -> Import -> Labels...). For users who want to import SRT files, this is inconvenient. Audacity determines import format from file extension anyway, so a generic file filter can be used which finds both types of files.

This change was only made for release 3.7.3 branch and tested to compile using Github actions runner: https://github.com/coezbek/audacity/actions/runs/14995843426

image

- [x] If changes are extensive (they are not), then there is a sequence of easily reviewable commits

Recommended:


kryksyh (approved):

kryksyh:
@coezbek lgtm, thank you!

chinakov:
This is tested, thanks!

coezbek:
Thanks for merging this PR. Would you want me to submit an identical PR against AU4 as well?

PR #8793, opened 2025-05-19
 Fix crash on close

kryksyh:
Resolves: #6890

- During shutdown `ProjectHistory::Get(project).SetStateTo(...)` is called and adds an event to the event queue. 
- Then, most UI elements are destroyed, including the call to ToolManager::Destroy, yet the subscription is kept connected. 
- If the project is closed without saving, the progress bar is shown to display the checkpointing progress.
- In turn, this runs the event loop and causes the event to be dispatched
- As a result, ToolManager is trying to update the toolbar's UI after getting an event from UndoManager.

Recommended:

### DIFF
--- a/src/toolbars/ToolManager.cpp
+++ b/src/toolbars/ToolManager.cpp
@@ -503,6 +503,7 @@ void ToolManager::Destroy()
          pair.second.reset();
 
       mIndicator.reset();
+      mMenuManagerSubscription.Reset();
    }
 }
 


src/toolbars/ToolManager.cpp
```diff
@@ -503,6 +503,7 @@ void ToolManager::Destroy()
          pair.second.reset();
 
       mIndicator.reset();
+      mMenuManagerSubscription.Reset();
```
saintmatthieu, line 506: What does this addition change? The subscription is about to get destroyed anyway and its destructor calls `Reset()`.

kryksyh: Right, that was the idea I guess.

ToolManager::Destroy is called here without a call to the destructor
https://github.com/audacity/audacity/blob/a46becac8043753e99e1e4d2bbf21a10368cc9f5/src/ProjectManager.cpp#L568

saintmatthieu: Oh, I see.

saintmatthieu (approved):

Issue #8502, opened 2025-03-24
AU4 zoom/view state isn't saved - in AU3 it is


IliasBergstromMuse:
Maybe this is planned, but I didn't see it in the backlog.

In AU3, if I zoom in to see a stem plot, save, close and re-open the project, the view state is restored to show that section of the stem plot.

In AU4 the view is reset.

chinakov:
Note: The zoom/view state of the project is saved only when the project is saved. 

This means that if the User has opened an existing project, adjusted the zoom level of the project, and then closed it, the project's new zoom level would not be retained. 

Otherwise, this change works fine. If you have saved the project, the zoom level is also saved on per-project basis.

chinakov:
Note: this behaviour described above is concurrent with AU3, so this is fine. Thanks!

PR #8795, opened 2025-05-20
Save view state

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8502

### DIFF
--- a/src/projectscene/CMakeLists.txt
+++ b/src/projectscene/CMakeLists.txt
@@ -31,6 +31,9 @@ set(MODULE_SRC
     ${CMAKE_CURRENT_LIST_DIR}/internal/projectsceneuiactions.h
     ${CMAKE_CURRENT_LIST_DIR}/internal/projectsceneactionscontroller.cpp
     ${CMAKE_CURRENT_LIST_DIR}/internal/projectsceneactionscontroller.h
+    
+    ${CMAKE_CURRENT_LIST_DIR}/internal/au3/viewinfo.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/internal/au3/viewinfo.h
 
     ${CMAKE_CURRENT_LIST_DIR}/view/common/tracksviewstatemodel.cpp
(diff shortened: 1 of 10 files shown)


src/au3wrap/internal/viewinfo.cpp
```diff
@@ -0,0 +1,90 @@
+#include "viewinfo.h"
```
embarc-gabriel, line 1: LGTM but I have a more architectural question and I will take this opportunity to learn a little bit.
Shouldn´t we persist this information on ProjectViewState/ProjectSceneConfiguration) or on au3 project?
As far as I understood reading a message from Elnur we should persist it on the former instead of relying on au3 code.
I have already used myself the same solution a couple of time (TrackColor and Snap) but after reading this message I wonder if there is an easier way to do this.
@Eism what do you think about it?

Eism: The idea is that Audacity 4 classes should access ProjectViewState to manage data, which is currently in Grzegorz's solution.
At the moment, all the data that we need to save for the current project, we save in the project file. And Grzegorz's solution looks good.

But we can discuss the following point, to rename ProjectViewState to Au3ProjectViewState, since it works with data from the project. Then we will have all the classes for working with ViewState in one place. Then maybe we will not need the ViewInfo class?

src/au3wrap/CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ set(MODULE_SRC
     ${CMAKE_CURRENT_LIST_DIR}/internal/au3basicui.h
     ${CMAKE_CURRENT_LIST_DIR}/internal/projectsnap.cpp
     ${CMAKE_CURRENT_LIST_DIR}/internal/projectsnap.h
+    ${CMAKE_CURRENT_LIST_DIR}/internal/viewinfo.cpp
```
Eism, line 49: should be in projectscene/internal/au3/...

grliszas14: fixed 👍

embarc-gabriel (approved):

Eism (approved):

Issue #8725, opened 2025-05-02
Windows Signing Certificate is outdated


kryksyh:
### Bug description

Signing certificate used to sign Windows binaries is obsolete.

Need to update the build process to follow the procedure defined at: https://alice.mu.se/display/UGuide/Code+Sign+service

### Steps to reproduce

- Run release build for AU3
- No installer is created
- Binaries are not signed

### Expected behavior

- Binaries are signed

### Audacity Version

Other

### Operating system

Windows

chinakov:
All installers and .exe files are signed with a certificate dated ‎Thursday, ‎May ‎22, ‎2025 11:55:43 AM. Thanks!

PR #8802, opened 2025-05-21
Update Windows signing process

kryksyh:
Resolves: https://github.com/audacity/audacity/issues/8725

Use signing service provided by Muse

Recommended:

### DIFF
--- a/scripts/build/windows/PfxSign.ps1
+++ b/scripts/build/windows/PfxSign.ps1
@@ -1,19 +1,15 @@
-# This script uses Set-AuthenticodeSignature to sign a file
-# or exe/dll/msi files in a directory using a pfx file.
-# If PFX file is not present, script expects a base64 encode certificate
-# in WINDOWS_CERTIFICATE environment variable.
-# Password can be passed using WINDOWS_CERTIFICATE_PASSWORD environment variable.
-# The script only signs previously unsigned files
+# This script uses signing service to sign windows binaries.
+
+# The file to be signed should be uploaded to the /unsigned/ folder.
+# The signing service runs automatically on a schedule
+# The result - the signed file - is placed by the service in the same s3 bucket
+#     in the signed folder with the same name as the source file.
+# The file must be removed from the signed folder after downloading.
 
 [CmdletBinding()]
 param (
-    # A path to the certificate file. If missing, env:WINDOWS_CERTIFICATE will be used.
-    [string]$CertFile, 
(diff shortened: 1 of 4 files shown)


scripts/build/windows/PfxSign.ps1
```diff
@@ -1,108 +1,153 @@
...
+    )
 
-$cleanupPfx = $false
+    Write-Host "Waiting for signed files from S3..."
+    Write-Host "Expecting signed files for:"
+    $FilesToDownload | ForEach-Object { Write-Host " - $_" }
+
+    $signedFilesDownloaded = @()
+    $maxAttempts = 10
+    $sleepSeconds = 60
```
embarc-gabriel, line 55: How do we know this is enough time?
"The signing service runs automatically on a schedule or via a Jenkins task."
Is this periodic or is it triggered when a new file is in the bucket?

scripts/build/windows/PfxSign.ps1
```diff
@@ -1,108 +1,153 @@
...
+
+            Write-Host "Trying to download signed file: $fileName"
+            aws s3 cp $s3SignedUrl $fileSignedPath
+            if ($LASTEXITCODE -eq 0) {
+                Write-Host "Successfully downloaded signed file: $fileName"
+                $signedFilesDownloaded += $fileName
+
+                $originalBackup = "${file}_origin"
+                Write-Host "Backing up original file to $originalBackup"
+                Rename-Item -Path $file -NewName $originalBackup -Force
```
embarc-gabriel, line 79: I don´t get why we create a backup to later remove it once there is no error handling here. It seems to be we could delete it anyway and then rename the signed file.

scripts/build/windows/PfxSign.ps1
```diff
@@ -1,108 +1,153 @@
...
+
+                Write-Host "Replacing original with signed file: $file"
+                Rename-Item -Path $fileSignedPath -NewName $file -Force
+                Remove-Item -Path $originalBackup -Force
+
+                Write-Host "Removing signed file from S3: $s3SignedUrl"
+                aws s3 rm $s3SignedUrl | Out-Null
```
embarc-gabriel, line 86: Should we check for errors on these aws calls?
Maybe we could just safely rerun this on failures. IDK

embarc-gabriel (approved):

scripts/build/windows/PfxSign.ps1
kryksyh: It is purely empirical. The exact time will depend on the runner's workload and the number of concurrent jobs.

PR #8811, opened 2025-05-22
Change the order of startup windows

kryksyh:
Now "App updates & usage info" is shown first, and "What's new" second

Recommended:


grliszas14 (approved):

chinakov:
The 'Welcome Screen' is shown only after the User has accepted or declined the UUID offer, as it now should. Thanks!

Issue #8796, opened 2025-05-20
Welcome dialog: accessibility: left and right arrow buttons


DavidBailes:
### Bug description

These buttons are cannot be activated by keyboard users, and have no visual indication of focus.
For screen reader users, their role should be button, but it isn't.
The button used for these buttons is a custom button which doesn't  have the necessary functionality for accessibility. (wxButtons, AButtons, and GradientButtons all have the necessary functionality.)
I don't know why a completely new custom button was used.

### Steps to reproduce

With a windows screen reader running:
1. When the Welcome dialog opens, the left arrow button is the focus (although there is no visible indication of this).
2. Press Enter or Spacebar to press the button.

### Expected behavior

1. The screen reader reads that the focused item is a button.
2. Pressing Enter or Spacebar activates the button, and the carousel is moved back one slide.

### Actual behavior

1. The screen reader reads that the focused item is panel. (It also has no accessibility name, but I'll log that as a separate issue.
2. Pressing Spacebar does nothing, and pressing Enter closes the dialog.

### Audacity Version

Audacity3

### Operating system

Windows

chinakov:
This is done, thanks!

PR #8816, opened 2025-05-22
Fix welcome screen accessibility

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8796

### DIFF
--- a/libraries/lib-wx-wrappers/ImageCarousel.cpp
+++ b/libraries/lib-wx-wrappers/ImageCarousel.cpp
@@ -13,6 +13,7 @@ ImageCarousel::ImageCarousel(wxWindow* parent, const std::vector<CarouselSnapsho
     : m_snapshots(snapshots), m_id(winid), wxPanel(parent, winid, pos, size) {
    SetBackgroundStyle(wxBG_STYLE_PAINT);
    SetBackgroundColour(theTheme.Colour(clrMedium));
+   SetWindowStyleFlag(GetWindowStyleFlag() | wxTAB_TRAVERSAL);
 
    gPrefs->Read(wxT("/GUI/IntroOrderStart"), &m_currentIndex, 0);
    m_currentIndex = m_currentIndex % m_snapshots.size();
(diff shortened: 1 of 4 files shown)


libraries/lib-wx-wrappers/ImageCarousel.cpp
```diff
@@ -13,28 +13,33 @@ ImageCarousel::ImageCarousel(wxWindow* parent, const std::vector<CarouselSnapsho
...
+   m_btnRight = new ArrowButton(this, ArrowDirection::Right);
 
    m_btnLeft->SetClickHandler([this] { OnLeftClicked(); });
    m_btnRight->SetClickHandler([this] { OnRightClicked(); });
    m_btnMiddle->Bind(wxEVT_BUTTON, [this](wxCommandEvent&) {
       OpenURL();
    });
        
+   CallAfter([this] {
+      m_btnLeft->SetFocus();
```
kryksyh, line 40: Maybe it would be better to put focus on the "Okay" button? 
Previously, pressing "Enter" closed the dialog. Now it lists promo pages backwards.

grliszas14: Fixed 👍

DavidBailes:
> @DavidBailes can you please try this build? Arrow keys' accessibility should be fixed

@grliszas14 . Thanks, the accessibility of the arrow keys for keyboard users now looks fine.

libraries/lib-wx-wrappers/ImageCarousel.cpp
DavidBailes: > Maybe it would be better to put focus on the "Okay" button?
> Previously, pressing "Enter" closed the dialog. Now it lists promo pages backwards.

@kryksyh . No, the user would expect the first control in the dialog to be the initial focus.
Also, in practice the user is going to want to check the checkbox to not show again before pressing the OK button.
So making the initial D an access key for the "Don't show this again at startup" checkbox would be helpful for sighted keyboard users.

(On macOS, with the first  control being the focus, pressing Enter should press the default commit button.)

grliszas14: Fixed:
* initial focus is on the left arrow
* D key checks "Don't show this again at startup" checkbox

@DavidBailes thanks!

PR #8818, opened 2025-05-23
VSCode config to build and debug on Linux

saintmatthieu:
Support is still limited (natvis doesn't seem to work, also couldn't figure out the `sourceFileMap` to debug into Qt sources or how to enable QML debugging), but it's already useful to VSCode users like me.

Recommended:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -4,7 +4,7 @@
             "name": "run audacity",
             "type": "cppvsdbg",
             "request": "launch",
-            "program": "${config:cmake.installPrefix}/bin/audacity.exe",
+            "program": "${workspaceFolder}/build/Debug/install/bin/audacity.exe",
             "args": [
                 // Comment this out to run audacity in plugin registration mode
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -26,6 +26,16 @@
...
+        },
+        {
+            "name": "run audacity (Linux)",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/build/Debug/src/app/audacity",
+            "preLaunchTask": "CMake: build",
+            "stopAtEntry": false,
+            "cwd": "${workspaceFolder}/build/Debug/src/app",
+            "visualizerFile": "${workspaceFolder}/.vscode/qt6.natvis",
```
kryksyh, line 38: I haven't tested it, but according to this: https://github.com/microsoft/vscode-cpptools/issues/476#issuecomment-277827714, you also need to enable `showDisplayString` option for it to work.

kryksyh (approved):

.vscode/launch.json
saintmatthieu: Looks better all right - thanks !

Issue #8203, opened 2025-02-11
Auto clip color improvement


grliszas14:
1. Rename "Auto" action to "Default track colour"
2. Display icon with current track colour

DimRal:
I'd like to take this on. I'll rename the 'Auto' action to 'Default track colour' and update the icon to reflect the current track color. Let me know if it's okay for me to take it on.

dozzzzer:
Tested

kryksyh:
> New updates from [@Tantacrul](https://github.com/Tantacrul):
> 
> * The 'Auto' setting should be renamed to 'Same as Track Color'
> * There should be a line break between the 'Same sa Track Color' setting and the individual colors of the clip to differentiate between these options.

@dozzzzer I believe that the wording should be updated, and the line break is missing

kryksyh:
@dozzzzer pardon me, I was looking at the wrong PR

kryksyh:
[#8847](https://github.com/audacity/audacity/pull/8847) - does not update the icon color to match the current track color

![Image](https://github.com/user-attachments/assets/333f534e-a0f1-4124-99f4-31d47bf5836e)

chinakov:
This is done, thanks!

PR #8847, opened 2025-05-27
rename "auto" color to "same as track color"

LWinterberg:
Resolves: #8203

### DIFF
--- a/src/projectscene/view/clipsview/clipcontextmenumodel.cpp
+++ b/src/projectscene/view/clipsview/clipcontextmenumodel.cpp
@@ -150,7 +150,9 @@ MenuItemList ClipContextMenuModel::makeClipColourItems()
     m_colorChangeActionCodeList.clear();
     MenuItemList items;
     items <<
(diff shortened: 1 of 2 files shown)


saintmatthieu (approved):

src/projectscene/view/clipsview/clipcontextmenumodel.cpp
```diff
@@ -150,7 +150,9 @@ MenuItemList ClipContextMenuModel::makeClipColourItems()
     m_colorChangeActionCodeList.clear();
     MenuItemList items;
     items <<
-        makeMenuItem("action://trackedit/clip/change-color-auto", muse::TranslatableString("clip", muse::String::fromStdString("Auto")));
```
embarc-gabriel, line 153: AFAIK we should also show what is the track color on this menu item. Is that correct @LWinterberg ?

embarc-gabriel: @kryksyh I think this was missing on the PR right?

kryksyh: You are right, thanks, I missed that!

PR #8879, opened 2025-05-30
8860 playback meter type

### LINKED ISSUE #8860: [AU4] Playback meter - Meter type
### Description

Implement meter types for the playback meter flyout.

Image

embarc-gabriel:
- Decoupled the meter from the ruler for both vertical and horizontal meters.
- Added support for multiple meter types.
- Adjust UI

Resolves: #8860 

https://github.com/user-attachments/assets/4d87ca03-1e94-4f5e-81fc-c92cb45f96cc

Recommended:


src/playback/internal/meterstrategy/dblinearmeter.cpp
```diff
@@ -0,0 +1,76 @@
+/*
```
embarc-gabriel: We have one implementation for meter type.
Then we export this to QML so we can draw the ruler and the values on the meter.

src/playback/qml/Audacity/Playback/components/HorizontalVolumePressureMeter.qml
```diff
@@ -47,7 +49,13 @@ Canvas {
 
         readonly property color meterBackgroundColor: Utils.colorWithAlpha(ui.theme.strokeColor, 0.7)
 
-        readonly property var maxPeakMarkerColor: "#14151A" 
+        readonly property var maxPeakMarkerColor: "#14151A"
+
+        readonly property var smallSteps: root.meterModel.smallSteps
+
+        onSmallStepsChanged: {
```
embarc-gabriel, line 56: This component must be aware of changes on the model. Probably there is a better way to do this.

src/playback/qml/Audacity/Playback/components/HorizontalVolumePressureRuler.qml
```diff
@@ -0,0 +1,63 @@
+
```
embarc-gabriel, line 1: Maybe we can unify all the rulers on a single component in the future.

src/playback/qml/Audacity/Playback/components/HorizontalVolumePressureRuler.qml
```diff
@@ -0,0 +1,63 @@
...
+    property var meterModel: null
+
+    height: rightText.paintedHeight
+
+    readonly property real leftTextMargin: leftText.paintedWidth / 2
+    readonly property real rightTextMargin: rightText.paintedWidth / 2
+    readonly property real effectiveWidth: width - leftTextMargin - rightTextMargin
+
+    onEffectiveWidthChanged: {
```
embarc-gabriel, line 18: The model should be aware of changes on the ruler size.

src/playback/internal/meterstrategy/linearmeter.cpp
```diff
@@ -0,0 +1,105 @@
+/*
+* Audacity: A Digital Audio Editor
+*/
+
+#include "linearmeter.h"
+
+#include "MemoryX.h"
```
Eism: Please, don't use au3 classes in au4 non-wrappers classes 
As I see it, you need this include for `DB_TO_LINEAR`, but you can use `db_to_linear` from `global/types/ratio.h` instead.
If you need any helper functions, just add them to the appropriate Audacity 4 file. (Or you need to write wrapers)

src/playback/iplaybackmeterstrategy.h
Eism, line 1: I’d avoid introducing a new term like Strategy. It’ll only raise more questions later - what is it, when to use it, etc. In my opinion, you can simply call it IPlaybackMeter. It might even be better without an interface - for example, we could just have AbstractPlaybackMeter, LinearPlaybackMeter, etc.

src/playback/internal/playbackmetercontroller.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <memory>
+
+#include "playback/internal/meterstrategy/dblogmeter.h"
+#include "playback/internal/meterstrategy/dblinearmeter.h"
+#include "playback/internal/meterstrategy/linearmeter.h"
+
+using namespace au::playback;
+
+static std::shared_ptr<IPlaybackMeterStrategy> createMeterStrategy(PlaybackMeterType::MeterType meterType)
```
saintmatthieu: What's the meaning of `static` here ?

src/playback/internal/meterstrategy/linearmeter.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include <sstream>
+#include <iomanip>
+#include <cmath>
+#include <algorithm>
+
+using namespace au::playback;
+
+static constexpr double MIN_VOLUME_DB = 0.0;
```
saintmatthieu: Aren't these linear values, i.e., `MIN_VOLUME_LIN` and `MAX_VOLUME_LIN` ?

src/playback/internal/meterstrategy/linearmeter.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+double LinearMeter::stepToPosition(double step) const
+{
+    const double clampedValue = std::clamp(step, MIN_VOLUME_DB, MAX_VOLUME_DB);
+    return (clampedValue - MIN_VOLUME_DB) / (MAX_VOLUME_DB - MIN_VOLUME_DB);
+}
+
+double LinearMeter::sampleToPosition(double sample) const
+{
+    const double linearValue = DB_TO_LINEAR(sample);
```
saintmatthieu: Why not `return stepToPosition(DB_TO_LINEAR(sample));` ?

src/playback/internal/meterstrategy/linearmeter.cpp
```diff
@@ -0,0 +1,113 @@
...
+    return steps;
+}
+
+std::vector<double> LinearMeter::smallSteps(int meterSize) const
+{
+    if (meterSize <= 0) {
+        return {};
+    }
+
+    const auto fullStepsValue = fullSteps(meterSize);
```
saintmatthieu: or maybe `const auto fullSteps = this->fullSteps(meterSize);` ?

src/playback/internal/meterstrategy/dblinearmeter.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+double DbLinearMeter::stepToPosition(double step) const
+{
+    const double clampedValue = std::clamp(step, MIN_VOLUME_DB, MAX_VOLUME_DB);
+    return dbToValuePercentage(clampedValue);
+}
+
+double DbLinearMeter::sampleToPosition(double sample) const
+{
+    if (sample == MIN_VOLUME_DB) {
```
saintmatthieu: I'm having a hard time understanding why this precise equality check. What if `sample < MIN_VOLUME_DB` ?

src/playback/internal/meterstrategy/dblinearmeter.cpp
saintmatthieu, line 1: Very happy with this guy.

saintmatthieu (approved):
The different meter implementations look very neat to me. Only added minor comments there.
I haven't paid much attention to the QML side, seeing as @eism is on it. I noticed, though, some imperfections: the meters, overshooting the bottom of the track (screenshot on its way). Also, playing back some dual-mono signals shows a lesser volume for the right channel, but this isn't a regression of this PR. Will mention this to QA.

saintmatthieu:
Overshooting meter:
image

src/playback/internal/playbackmetercontroller.cpp
embarc-gabriel: The idea is to ensure this function is only used within this file. Even if declare the function on another cpp file it won´t link with this implementation.
I could have added this as private member of this class or even another simple object. Something like a MeterBuilder. But usually I prefer this approach.

saintmatthieu: I totally agree not to make a member function. I didn't know that `static` had this effect in this case. To achieve this, I place the functions in an anonymous namespace,
```
namespace {
void foo() { ... }
}

embarc-gabriel: You are right here.
This is the correct way of doing this nowadays.
I will change it.

src/playback/internal/meterstrategy/linearmeter.cpp
embarc-gabriel: Done!

src/playback/iplaybackmeterstrategy.h
embarc-gabriel: Ok. Changed.

src/playback/internal/playbackmetercontroller.cpp
embarc-gabriel: Fixed.

src/playback/internal/meterstrategy/linearmeter.cpp
embarc-gabriel: You are right. Much cleaner.
Fixed.

src/playback/internal/meterstrategy/linearmeter.cpp
embarc-gabriel: Fixed.

src/playback/internal/meterstrategy/linearmeter.cpp
embarc-gabriel: Fixed.

src/playback/internal/meterstrategy/dblinearmeter.cpp
embarc-gabriel: In fact this is clamped on the other function so it can be removed.

src/playback/internal/meterstrategy/dblinearmeter.cpp
embarc-gabriel: Thanks! :)

embarc-gabriel:
The overshooting was also fixed. @saintmatthieu

PR #8883, opened 2025-05-30
UI-less Lv2 support

### LINKED ISSUE #8302: Add loading LV2 plugins
### Description

Implement the ability to use LV2 plugins in AU4 destructively

saintmatthieu:
Resolves: #8302

This PR provides UI-less support for LV2 on Linux and MacOS (*), namely
* plugins are discovered and made available in the effect menus,
* they can be applied destructively or in real time
* UI-less: the viewers will be empty, so they can only be used with default values for now

(*) On Windows, at the time of writing, the DSP binary doesn't get found. This might be a difficult problem or not, but Windows support is not a priority at all anyway. That we have MacOS support is already going to be helpful for development.

Recommended:

### QA
This PR involved some refactoring of our effect-related code. Please prioritize that
- [x] the existing effect capabilities not having regressed on any platform.

Where LV2 is concerned, please test the following, exclusively on Linux (non-Linux support being there more for developer convenience) :
- [x] Can be applied destructively
- [x] When added to a track, processes the audio and can be bypassed like other effect families (please ignore that the dialog is empty)
- [x] Cancel / Preview / Apply of the destructive dialogs already work (although the UI not being apparent)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,7 @@ option(AU_BUILD_EFFECTS_TESTS "Build effects tests" ON)
 option(AU_MODULE_EFFECTS_VST "Build audacity vst module" ON)
 set(AU_MODULE_VST_VST3_SDK_PATH "" CACHE PATH "[optional] Path to VST3_SDK. SDK version >= ${VST3_SDK_VERSION} required")
 
-option(AU_MODULE_EFFECTS_LV2 "Build audacity lv2 module" OFF)
+option(AU_MODULE_EFFECTS_LV2 "Build audacity lv2 module" ON)
 set(AU_MODULE_LV2_LV2SDK_PATH "" CACHE PATH "[optional] Path to lv2 SDK. lilv version >= 0.24.26 required")
 
 option(AU_USE_FFMPEG "Build mod-ffmpeg module" ON)
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -87,6 +87,7 @@ set(LINK_LIB
     au3wrap
     context
     effects_vst
+    effects_lv2
     project
     projectscene
     audio
@@ -104,7 +105,6 @@ if (AU_MODULE_EFFECTS_VST)
 endif()
 
 if (AU_MODULE_EFFECTS_LV2)
-    list(APPEND LINK_LIB effects_lv2)
     add_definitions(-DAU_MODULE_EFFECTS_LV2)
 endif()
 
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -43,6 +43,11 @@ bool EffectsProvider::isNyquistSupported() const
     return nyquistEffectsRepository() ? true : false;
 }
 
+bool EffectsProvider::isLv2Supported() const
+{
+    return lv2EffectsRepository() ? true : false;
+}
+
 void EffectsProvider::reloadEffects()
 {
     m_effects.clear();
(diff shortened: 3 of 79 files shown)


src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -108,14 +121,21 @@ bool EffectsProvider::loadEffect(const EffectId& effectId) const
...
     }
-    IF_ASSERT_FAILED(it->family == EffectFamily::VST3) {
-        LOGE() << "unknown family: " << static_cast<int>(it->family);
-        return false;
-    }
-    IF_ASSERT_FAILED(vstEffectsRepository()) {
-        return false;
+    switch (it->family) {
+        case EffectFamily::VST3:
+            IF_ASSERT_FAILED(vstEffectsRepository()) {
```
kryksyh: `IF_ASSERT_FAILED(vstEffectsRepository())` wouldn't this and similar checks always be true, when the module is disabled it will be stubbed isn't it?

saintmatthieu: When the module is stubbed, this `vstEffectsRepository()` would return nullptr. (But if it does, then we should never go as far as executing `EffectsProvider::loadEffect(const EffectId& effectId)` with the ID of a VST effect - hence the assert).

CMakeLists.txt
```diff
@@ -117,7 +117,7 @@ option(AU_BUILD_EFFECTS_TESTS "Build effects tests" ON)
 option(AU_MODULE_EFFECTS_VST "Build audacity vst module" ON)
 set(AU_MODULE_VST_VST3_SDK_PATH "" CACHE PATH "[optional] Path to VST3_SDK. SDK version >= ${VST3_SDK_VERSION} required")
 
-option(AU_MODULE_EFFECTS_LV2 "Build audacity lv2 module" OFF)
+option(AU_MODULE_EFFECTS_LV2 "Build audacity lv2 module" ON) # TODO do we need a stub for this module, too?
```
igorkorsukov: Stub is needed if there is a call to the module services somewhere else
And it is not needed if there are no calls, for example, the module registers itself somewhere (and when off, it simply does not register)

src/app/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ set(LINK_LIB
     au3wrap
     context
     effects_vst
+    effects_lv2
```
igorkorsukov, line 90: It is necessary to keep in mind that when disabling the module there will be a linking error. So it is necessary to make a stub

src/effects/lv2/internal/ilv2plugininstance.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+#pragma once
+
+#include "global/async/notification.h"
+
+#include "vsttypes.h"
+#include "audio/audiotypes.h"
+
+namespace muse::vst {
+class IVstPluginInstance
```
igorkorsukov: looks like a mistake

src/effects/lv2/view/lv2viewloader.h
```diff
@@ -0,0 +1,32 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+#pragma once
+
+#include <QObject>
+#include <QQuickItem>
+
+namespace au::effects {
+//! TODO Move to builtin module
```
igorkorsukov: need remove

CMakeLists.txt
saintmatthieu: I added it because otherwise the effect dialogs couldn't find the LV2 view classes. (And you also point out the need for it for linkage in https://github.com/audacity/audacity/pull/8883#discussion_r2128189714)

src/effects/lv2/internal/ilv2plugininstance.h
saintmatthieu: Hm. How did that even compile ?.. Thanks.

src/app/CMakeLists.txt
saintmatthieu: Answered in https://github.com/audacity/audacity/pull/8883#discussion_r2128206616

src/effects/lv2/internal/ilv2plugininstance.h
saintmatthieu: ... the entire file is garbage (missed by the collection ;) )

src/effects/lv2/view/lv2viewloader.h
saintmatthieu: Yep

igorkorsukov (approved):

kryksyh (approved):

Issue #8699, opened 2025-04-25
After moving a mono clip to a stereo track and undoing it, generating on the stereo track produces wrong result


dozzzzer:
### Steps to reproduce

1. Create a stereo and a mono track
2. Generate a tone on the mono track
3. Move it to the stereo track
4. Undo
5. Select the stereo track
6. Generate a tone on it
7. Observe

### Expected behavior

The generated tone clip has two channels

### Actual behavior

The generated tone clip has one channel; resizing the track reveals the second channel.

https://github.com/user-attachments/assets/671eca2f-5fe3-4635-87e5-451d9628c4c9

### Audacity Version

master

### Operating system

OS independent

chinakov:
AU4-only issue.

luapmartin:
Ready to be tested, QA please:
* check if described issue is fixed
* check if changing track height and track height ratio is not broken 
* check all 3 `Asymmetric stereo heights` settings are working as expected 

Thanks.

_open question: after deleting a stereo track with a custom track height ratio should redo reconfigure the deleted track with the same custom ratio or the default one ? (AU3 does, AU4 doesn't)_

dozzzzer:
> open question: after deleting a stereo track with a custom track height ratio should redo reconfigure the deleted track with the same custom ratio or the default one ? (AU3 does, AU4 doesn't)

Yes it should. In general, unless otherwise specified, refer to the existing behaviour of AU3. @luapmartin

luapmartin:
Follow up issue https://github.com/audacity/audacity/issues/8939

PR #8905, opened 2025-06-09
audacity-8699 fix: channelHeightRatio is handled inside ClipItem and ChannelSplitter so we default initialise for stereo case

luapmartin:
Resolves: [this issue](https://github.com/audacity/audacity/issues/8699)

no more issues when dropping a mono track on a stereo track, undo is behaving as expected now

Recommended:

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
@@ -23,7 +23,7 @@ Item {
     property bool isMultiSelectionActive: false
     property bool isTrackAudible: true
     property bool isStereo: clipsModel.isStereo
-    property double channelHeightRatio: isStereo ? 0.5 : 1
+    property double channelHeightRatio: 0.5
     property bool moveActive: false
     property bool altPressed: false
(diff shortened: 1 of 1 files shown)


kryksyh:
@luapmartin I'm sure that this change fixes the observed issue, but are we sure that there are no deeper problem with undo system?

We have STR from #8699:

> 1. Create a stereo and a mono track
> 2. Generate a tone on the mono track
> 3. Move it to the stereo track
> 4. Undo
> 5. Select the stereo track
> 6. Generate a tone on it
> 7. Observe

But if we change it a bit:

1. Create a stereo and a mono track
**2. Generate a tone on the stereo track**
**3. Undo**
4. Generate a tone on the mono track
5. Move it to the stereo track
6. Undo
7. Select the stereo track
8. Generate a tone on it
9. Observe

Then the behavior would be correct, and on second undo, the stereo track will be restored with its original L/R ratio.

luapmartin:
Hi @kryksyh , thanks, so I tested with the build both scenarios the original one and this small variation and it's track heights are appearing correctly on on my machine. So I am not sure about what you would want to be done here.

luapmartin:
Follow up issue: https://github.com/audacity/audacity/issues/8939

Issue #479, opened 2020-04-04
Request: Add Recording Notes Integration


ignacib:
I sometimes use Audacity as a recorder for the projects I am working on and I think that instead of having Audacity and text editor open it would be easy to just have the two integrated into one app instead. So if you're recording a podcast or interview,  a sidebar that has notes relating to what you want to discuss or say is right there for someone to read off of would be useful and convenient. I could just have two programs running at the same time in two windows but if I have to do some adjustment in Audacity I would have to either use the now smaller window with audacity or maximize the window and risk either having to pause or losing track of what I was saying or doing.

jonata:
Can I adopt this? Can we try to implement this in Audacity 4?

PR #8906, opened 2025-06-09
Feat #479: Add Recording Notes Integration

NunoCMatos:
Implements: #479

As described in the issue, this implementation introduces a basic notepad feature, allowing users to take recording notes directly within Audacity — avoiding the need to switch between multiple windows.

We believe this is a solid starting point, with potential for future enhancements, such as a “teleprompter” mode, bullet-point checklists for tracking progress, and more.

Here’s a short video demonstrating the new feature:

https://github.com/user-attachments/assets/6888dc72-0a04-41fc-84c5-fcaaa54e1cd9

Recommended:

### DIFF
--- a/src/appshell/qml/ProjectPage/ProjectPage.qml
+++ b/src/appshell/qml/ProjectPage/ProjectPage.qml
@@ -94,6 +94,14 @@ DockPage {
         return null
     }
 
+    function openNotepad() {
+        if (notepadPanel.visible) {
+            notepadPanel.close()
+        } else {
+            notepadPanel.open()
+        }
+    }
+
     onInited: {
         Qt.callLater(pageModel.init)
     }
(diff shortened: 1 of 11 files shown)


NunoCMatos:
Hi! The codestyle check just failed and it says:
"Please run tools/codestyle/uncrustify_run_file.sh on these files and then amend your commit."
Where can I find this script? I couldn't locate it in the repository. Thanks!

NunoCMatos:
Hi! I tried running the commands suggested in the workflow output. I executed both the exact workflow commands and the ones provided in the error message. However, it seems like the files mentioned there (e.g. tools/codestyle/uncrustify_run_file.sh) don't match what I have locally - or at least the paths/files don't seem to exist or get modified.

Could you please confirm if there’s something missing or if I might be doing something wrong? Thanks!

LWinterberg:
I believe uncrustify is present in the muse_framework repository, rather than the Audacity one. 

As for the feature: You may want to see how project data is stored in the project file, and save the project notes inside the AUP3 itself. That avoids you needing to do any of the new file/save file/open file operations on actual files in the OS, and ensures that the user never loses their recording notes from a project.

NunoCMatos:
Thanks for your input! I’ll work on it and reach out with any questions if needed.

Issue #8911, opened 2025-06-11
audacity-win-3.7.4-64bit.exe does not contain a valid digital signature


oed-pelza:
### Bug description

In the past all setup files was digital signed. The latest release is missing the signature. 

The security vendor Bkav Pro detects this file as malicious, but this is the only one from virus total. looks like a false positive.

### Steps to reproduce

Download latest version of audacity-win-3.7.4-64bit.exe and take a look at the properties of the file:

![Image](https://github.com/user-attachments/assets/f6997401-aa5e-4629-9111-b874106c8d12)

the digital signature is missing.

### Expected behavior

a valid existing digital signature.

### Audacity Version

Audacity 3.7.4

### Operating system

Windows

kryksyh:
@oed-pelza thanks! Will fix this in a moment!

kryksyh:
@oed-pelza digital signatures are now in place, and measures are taken to prevent this in the future. 
Thanks for the report!

PR #8913, opened 2025-06-11
Check if the file is signed after being processed on the signing service

kryksyh:
Resolves: https://github.com/audacity/audacity/issues/8911 [not directly, but should prevent this in future]

Files are not always signed by the signing service, to be fixed on their side.

Recommended:

### DIFF
--- a/scripts/build/windows/PfxSign.ps1
+++ b/scripts/build/windows/PfxSign.ps1
@@ -122,14 +122,28 @@ function Download-SignedFiles {
             aws s3 cp $s3SignedUrl $fileSignedPath --quiet
             if ($LASTEXITCODE -eq 0) {
                 Write-Host "Successfully downloaded signed file: $s3FileName"
-                $signedFilesDownloaded += $file
-                Write-Host "Replacing original with signed file: $file"
 
-                Remove-Item -Path $file -Force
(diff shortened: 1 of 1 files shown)


LWinterberg (approved):

scripts/build/windows/PfxSign.ps1
```diff
@@ -122,14 +122,28 @@ function Download-SignedFiles {
...
-                Write-Host "Replacing original with signed file: $file"
-
-                Remove-Item -Path $file -Force
-                Rename-Item -Path $fileSignedPath -NewName $file -Force
 
-                Write-Host "Removing signed file from S3: $s3SignedUrl"
-                aws s3 rm $s3SignedUrl | Out-Null
+                # Do not trust the signing service, check the signature
+                $sig = Get-AuthenticodeSignature -FilePath $fileSignedPath
+                if ($sig.Status -eq 'Valid') {
```
embarc-gabriel, line 128: Is this ok?
Shouldn´t it be something like: $sig.Status -eq [System.Management.Automation.SignatureStatus]::Valid ?
Maybe there is a implicit conversion here, idk.

kryksyh: @embarc-gabriel yep, it works fine in powershell
image

Issue #8572, opened 2025-03-31
No error is given when you open a project file that doesn't exist anymore


dozzzzer:
### Steps to reproduce

1. Create a project, save it
2. Close it
3. Delete the project file
4. Try to open it from the Home screen

### Expected behavior

The following error message is displayed:

<img width="260" alt="Image" src="https://github.com/user-attachments/assets/50fefee8-ca1e-45d6-99d9-7335190ac9ca" />

No project is opened. The project should also be removed from the Recent projects view.

### Actual behavior

A blank project is opened. An attempt to e.g. generate smth will result in the error "Connection to project is null".

### Audacity Version

master

### Operating system

OS independent

chinakov:
AU4-only issue.

luapmartin:
Please also test MuseScore: https://github.com/musescore/MuseScore/pull/28460

chinakov:
Tested all 3 permission types (write-only, read-only, no read/write) for Windows, and the error messages are shown correctly. Deleting the file or renaming shows a different error message correctly.

chinakov:
Re-tested, all is well. Thanks!

luapmartin:
https://github.com/musescore/MuseScore/pull/28595 merged and changes made in https://github.com/audacity/audacity/pull/8922 accordingly

luapmartin:
@chinakov this one might be related too : [#8921](https://github.com/audacity/audacity/issues/8921)

bwagner:
Please [reconsider](https://github.com/audacity/audacity/issues/8402#issuecomment-3126952708) automatic deletion of inexistant projects.

Issue #8921, opened 2025-06-12
CRASH when Opening a Read Only project file


luapmartin:
### Bug description

AU3 has a popup to warn about opening **read only** project files.

![Image](https://github.com/user-attachments/assets/88135fad-956c-401c-a7e1-651cfc51eb11)

AU4 needs to handle gracefully this scenario.

Some info:

Inside the 

```
auto ProjectFileIO::LoadProject(const FilePath& fileName, bool ignoreAutosave)
-> std::optional<TentativeConnection>
```
When opening read only files
We do call to
     `success = DeleteBlocks(blockids, true);`
That do a write operation on the SQLite file with **DELETE**
then would return the error code for read only to the parent,
we are inside the 
`bool Au3ProjectAccessor::load(const muse::io::path_t& filePath)`
we should handle this SQLite error and display the right error message to the use 
we could add some new codes inside `src/project/projecterrors.h`

using `ProjectActionsController::warnProjectCannotBeOpened` to display the popup to the user

### Steps to reproduce

- create `myproject.aup3` 
- save it
- close it
- make it read only: `chmod 444 myproject.aup3`
- open it -> **Crash**

### Expected behavior

message to the user about read only files limitation

### Actual behavior

crash

### Audacity Version

AU4

### Operating system

OS independent

chinakov:
This issue has been fixed via #8572

PR #8922, opened 2025-06-12
audacity-8572 Warning message when you open a project file that doesn't exist anymore (files that don't have read access warn too)

luapmartin:
Resolves: https://github.com/audacity/audacity/issues/8572

Warning message when you open a project file that doesn't exist anymore.
Files that don't have read access warn too.

`chmod 044 myproject.aup3`
[NEW] Popup message "This file could not be opened. Please make sure that Audacity has permission to read this file."

`chmod 444 myproject.aup3`
Crash with read only file (expected see https://github.com/audacity/audacity/issues/8921) 

`chmod 644 myproject.aup3`
OK “normal” case reads fine.

**Moved file**
[NEW] Popup message "This file does not exist or cannot be accessed at the moment."

Recommended:

### DIFF
--- a/src/project/tests/audacityproject_tests.cpp
+++ b/src/project/tests/audacityproject_tests.cpp
@@ -122,12 +122,12 @@ TEST_F(Project_Audacity4ProjectTests, Load_FileDoesNotExist_ReturnsProjectFileNo
     EXPECT_FALSE(ret.success());
     EXPECT_EQ(ret.code(), static_cast<int>(Err::ProjectFileNotFound));
 
-    const auto title = ret.data("title");
-    const auto body  = ret.data("body");
+    const auto title = ret.data<std::string>("title", std::string(""));
+    const auto body  = ret.data<std::string>("body", std::string(""));
 
-    ASSERT_TRUE(title.has_value());
-    ASSERT_TRUE(body.has_value());
-    EXPECT_THAT(std::any_cast<std::string>(title), ::testing::HasSubstr("Cannot read file"));
+    ASSERT_TRUE(!title.empty());
+    ASSERT_TRUE(!body.empty());
+    EXPECT_THAT(title, ::testing::HasSubstr("Cannot read file"));
     //can't close m_currentProject->close();
 }
 
@@ -142,7 +142,7 @@ TEST_F(Project_Audacity4ProjectTests, Load_FileCannotBeOpened_ReturnsReadProtect
 
     EXPECT_FALSE(ret.success());
     EXPECT_EQ(ret.code(), static_cast<int>(Err::ProjectFileIsReadProtected));
-    EXPECT_TRUE(ret.data("body").has_value());
+    EXPECT_TRUE(!ret.data<std::string>("body", std::string("")).empty());
     //can't close m_currentProject->close();
 }
 
@@ -171,7 +171,7 @@ TEST_F(Project_Audacity4ProjectTests, Load_NonEmptyFileIsWriteProtected_ReturnsW
 
     EXPECT_FALSE(ret.success());
     EXPECT_EQ(ret.code(), static_cast<int>(Err::ProjectFileIsWriteProtected));
-    EXPECT_TRUE(ret.data("body").has_value());
+    EXPECT_TRUE(!ret.data<std::string>("body", std::string("")).empty());
     m_currentProject->close();
 }
 } // namespace au::project


src/project/internal/projectactionscontroller.cpp
```diff
@@ -622,6 +621,57 @@ void ProjectActionsController::clearRecentProjects()
...
+    case engraving::Err::FileCorrupted:
+        return askIfUserAgreesToOpenCorruptedProject(io::filename(filepath).toString(), ret.text());
+    case engraving::Err::FileCriticallyCorrupted:
+        warnProjectCriticallyCorrupted(io::filename(filepath).toString(), ret.text());
+        return false;
+    default:
+        break;
+    }
+*/
+    warnProjectCannotBeOpened(ret, filepath);
```
igorkorsukov, line 636: I believe that this is not the best place to call this method, because the checkCanIgnoreError method now does not perform one responsibility (check) and its actions do not correspond to its name. For example, if I need to call the method to check whether an error can be ignored in some other scenario, or for example in console mode (some export, for example), then I will expect that it will only do a check and nothing more (and it will also show an error message)

src/project/projecterrors.h
```diff
@@ -33,9 +34,10 @@ enum class Err {
     NoProjectError,
     NoPartsError,
     CorruptionError,
-    CorruptionUponOpenningError,
+    CorruptionUponOpeningError,
 
     FileOpenError,
+    FileNotFound,
```
igorkorsukov: Let's rename it to ProjectOpenError, ProjectNotFound to make these errors more specific. We'll do the same in MuseScore in a little while.

luapmartin: Thanks Igor, again here, was following the `MuseScore/src/engraving/engravingerrors.h` error naming pattern. but, I am ok to change those to Project base errors names. Thought the `ProjectNotFound` would maybe imply the file is found and we managed to opened it but then inside of it the project isn't found right? And concerning the `FileOpenError`, at the end, this error is about the **Read** access missing to the file (container) not the project inside of it so as you prefer.

igorkorsukov: >  Thought the ProjectNotFound would maybe imply the file is found and we managed to opened it but then inside of it the project isn't found right? 

I think this is an error code `ProjectBadFormat`

> And concerning the FileOpenError, at the end, this error is about the Read access missing to the file (container) not the project inside of it so as you prefer.

Maybe `ProjectReadFailed` would be more appropriate?
and it doesn't really matter why, it failed to open, or the file opened but couldn't be read... probably for the caller it's all "failed to read the project"

src/project/internal/projectactionscontroller.cpp
luapmartin: Thanks Igor, here I am following the same pattern as in `MuseScore/src/project/internal/projectactionscontroller.cpp` where the `checkCanIgnoreError` is called from within the `RetVal<INotationProjectPtr> ProjectActionsController::loadProject(const muse::io::path_t& filePath)`, do you think those workflows aren't 'similar' and needs different treatment?

igorkorsukov: It looks like either there was a design error or the name of the method is incorrect and does not correspond to what the method does.

igorkorsukov (approved):

src/project/internal/audacityproject.cpp
```diff
@@ -97,6 +99,28 @@ muse::Ret Audacity4Project::doLoad(const io::path_t& path, bool forceMode, const
...
+    // check for file permissions
+    {
+        io::File file(path);
+        if (!file.open(io::IODevice::ReadOnly)) {
+            LOGE() << "failed open file (Can't Read): " << path;
+            return make_ret(Err::ProjectFileIsReadProtected, path);
+        }
+        file.close();
+
+        if (!file.open(io::IODevice::WriteOnly)) {
```
kryksyh: I don't think we need to check for this particular case here. Not being able to open the file in "w" mode is one of the errors produced by ProjectFileIO.cpp:848. Therefore, we should either handle all of them or not bother.

src/project/internal/audacityproject.cpp
```diff
@@ -97,6 +99,28 @@ muse::Ret Audacity4Project::doLoad(const io::path_t& path, bool forceMode, const
     UNUSED(forceMode);
     UNUSED(format);
 
+    // check file access
```
kryksyh: In general, we do not add comments to self-explanatory code blocks, like in this case, `!io::FileInfo::exists(path)` explains what is happening better than the comment above.

src/project/internal/audacityproject.cpp
```diff
@@ -6,6 +6,8 @@
 #include "global/io/ioretcodes.h"
 
 #include "log.h"
+#include "global/io/fileinfo.h"
```
kryksyh, line 8: Please follow the include order as described in https://github.com/musescore/muse_framework/wiki/CodeGuidelines

src/project/internal/projectactionscontroller.cpp
```diff
@@ -622,6 +626,67 @@ void ProjectActionsController::clearRecentProjects()
...
 }
 
+bool ProjectActionsController::shouldRetryLoadAfterError(const Ret& ret, const muse::io::path_t& filepath)
+{
+    if (ret) {
+        return true;
+    }
+    //! TODO AU4 :
+    switch (static_cast<project::Err>(ret.code())) {
+    /*
```
kryksyh: Do we need this commented block here? It is copied from musescore. FileTooOld, TooNew or Old300Format have no relevance to audacity. The same goes for projecterrors.h

src/project/projecterrors.h
luapmartin: fixed

src/project/internal/audacityproject.cpp
luapmartin: thanks, removed

src/project/internal/projectactionscontroller.cpp
luapmartin: the idea was to provide frame for future work, removed

src/project/internal/audacityproject.cpp
luapmartin: the idea was to early detect write protected files to avoid unnecessary DB connection and error traversing handling etc. (it makes the error more robust to a possible project format change) nevertheless that would be great to have other DB error being handled too

src/project/internal/audacityproject.cpp
luapmartin: ok, thanks

src/project/tests/testtools.h
```diff
@@ -0,0 +1,147 @@
+/*
+* Audacity: A Digital Audio Editor
+*/
+
+#ifndef TESTTOOLS_H
```
kryksyh: please use #pragma once

src/project/tests/audacityproject_tests.cpp
```diff
@@ -0,0 +1,174 @@
+/*
+* SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
```
kryksyh: Please use Audacity header

src/project/tests/mocks/clipboardmock.h
```diff
@@ -0,0 +1,24 @@
+/*
```
kryksyh: Can't we reuse it from trackedit tests?

src/project/tests/environment.cpp
```diff
@@ -0,0 +1,143 @@
+/*
+* Audacity: A Digital Audio Editor
+*/
+
+#include "testing/environment.h"
```
kryksyh, line 5: I'm a bit confused, is everything in this file required just to test that we return the correct error if the file could not be opened?

src/project/tests/data/test_writeprotected.aup3
kryksyh, line 1: Do we need these files, if we change the permissions at runtime?

src/project/tests/audacityproject_tests.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    ~ScopedTestFile()
+    {
+        testtools::removeIfExists(dstPath);
+    }
+
+    const std::string& getPath() const { return dstPath; }
+
+private:
+    std::string dstPath;
```
kryksyh: please use the same style guide we use for the reset of the code (`m_` prefix for member variables)

src/project/tests/audacityproject_tests.cpp
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#include <cassert>
```
kryksyh: not needed?

src/project/tests/audacityproject_tests.cpp
```diff
@@ -0,0 +1,174 @@
...
+    {
+        m_currentProject = new project::Audacity4Project();
+        m_currentProject->trackeditProjectCreator.set(m_trackeditProjectCreator);
+        m_currentProject->viewStateCreator.set(m_projectViewStateCreator);
+        m_currentProject->clipboard.set(m_clipboard);
+    }
+
+    void TearDown() override
+    {
+        // delete m_currentProject;
```
kryksyh: why is it commented, cant we use unique_ptr here?

src/project/tests/audacityproject_tests.cpp
```diff
@@ -0,0 +1,174 @@
...
+        testtools::removeIfExists(dstPath);
+    }
+
+    const std::string& getPath() const { return dstPath; }
+
+private:
+    std::string dstPath;
+};
+
+class Project_Audacity4ProjectTests : public ::testing::Test
```
kryksyh, line 74: I guess `Project_` prefix is not needed here?

src/project/projecterrors.h
```diff
@@ -46,6 +52,59 @@ inline muse::Ret make_ret(Err e)
 {
     return muse::Ret(static_cast<int>(e));
 }
+
+inline muse::Ret make_ret(Err err, const muse::io::path_t& filePath)
+{
+    auto ret = muse::Ret(static_cast<int>(err));
```
kryksyh: please move this to cpp, in general, if the function is not a one-liner, like the one above, it should be in cpp. This makes headers easier to read.

src/project/internal/projectactionscontroller.h
```diff
@@ -1,26 +1,26 @@
...
 
-#include "iprojectfilescontroller.h"
+#include "project/iprojectfilescontroller.h"
 
 #include "actions/actionable.h"
 #include "async/asyncable.h"
 
 #include "modularity/ioc.h"
-#include "iinteractive.h"
+#include "global/iinteractive.h"
```
kryksyh, line 8: duplication

#include "global/iinteractive.h"

is already included three lines below

src/project/internal/projectactionscontroller.h
```diff
@@ -1,26 +1,26 @@
 #ifndef AU_PROJECT_PROJECTACTIONSCONTROLLER_H
 #define AU_PROJECT_PROJECTACTIONSCONTROLLER_H
 
-#include "iprojectfilescontroller.h"
+#include "project/iprojectfilescontroller.h"
```
kryksyh: please move this below to the rest of "project*" includes

src/project/internal/audacityproject.cpp
```diff
@@ -215,7 +229,7 @@ Ret Audacity4Project::canSave() const
 
     // Ret ret = m_engravingProject->checkCorrupted();
     // if (!ret) {
-    //     Err errorCode = m_engravingProject->isCorruptedUponLoading() ? Err::CorruptionUponOpenningError : Err::CorruptionError;
+    //     Err errorCode = m_engravingProject->isCorruptedUponLoading() ? Err::CorruptionUponOpeningError : Err::CorruptionError;
```
kryksyh, line 232: I believe this could be safely removed?

luapmartin: I fixed a typo but yes there are some `//TODO AU4` comments there that we can remove I guess, though  I don't know why they were placed there in the first place.
Those TODOs are present in many methods: `needSave()`, `canSave()` `load`, `displayName`, `doSave`, `markAsSaved`, `setNeedSave`.

luapmartin: should I clean them all?

kryksyh: > should I clean them all?

depending on are they relevant to AU4 or not. Verifying this could be a task in itself. Let's keep it and avoid polluting this PR further.

src/project/tests/data/test_writeprotected.aup3
luapmartin: wrongly added, thanks, nice catch

luapmartin:
latest changes related to https://github.com/musescore/MuseScore/pull/28595 are done

PR #8933, opened 2025-06-13
Show what's new dialog for new releases

kryksyh:
Resolves: #[9029](https://github.com/audacity/audacity/issues/9029)

Display "What's new" dialog whenever the application version is increased.

Recommended:


src/AudacityApp.cpp
```diff
@@ -454,6 +434,13 @@ void PopulatePreferences()
          gPrefs->DeleteEntry("/Quality/LibsoxrSampleRateConverterChoice");
    }
 
+   // Reset the splash screen for every new version
```
luapmartin, line 437:
```suggestion
   // Reset the splash screen for every new version.release.revision
```

luapmartin: minor detail

PR #8940, opened 2025-06-16
Avoid crash when no track is selected

### LINKED ISSUE #8927: Crash after deleting a track
https://github.com/user-attachments/assets/eb4f1cdd-a78c-4abf-be98-4bc1846e5688

embarc-gabriel:
Resolves: #8927 

Avoid crash when no track is selected

Recommended:


embarc-gabriel:
Probably @grliszas14 has a better understanding of the whole context here.
At least I have tested and this fixes this two nasty bugs with no side effect as far as I could see.

kryksyh (approved):

PR #8946, opened 2025-06-18
Enable Autobot module

kryksyh:
Enables the Autobot module

Recommended:


share/autobotscripts/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+
+# Scripts
+install(FILES
+      TC1.1_BasicTest.js
+      DESTINATION ${AU4_SHARE_NAME}${AU4_INSTALL_NAME}autobotscripts
+      )
+
+# Steps
+install(FILES
+      steps/Inspector.js
```
igorkorsukov: should not be needed

src/app/CMakeLists.txt
```diff
@@ -108,6 +108,11 @@ if (AU_MODULE_EFFECTS_LV2)
     add_definitions(-DAU_MODULE_EFFECTS_LV2)
 endif()
 
+if (MUSE_MODULE_AUTOBOT)
+    list(APPEND LINK_LIB muse::autobot)
+    add_definitions(-DMUSE_MODULE_AUTOBOT)
```
igorkorsukov: it shouldn't be necessary, to use define, we need to include `muse_framework_config.h`, it's there

igorkorsukov:
> ## Pull Request Overview
> Comments suppressed due to low confidence (1)

Probably Capilot is right, there is probably a copy-paste error in the navigation escape :)

chinakov:
Autobot functions correctly, thanks. I was able to create a simple test case with limited functionality.

Issue #8897, opened 2025-06-05
Implement export with FFmpeg


grliszas14:
### Description

https://www.figma.com/design/QaCzd2bM08W8VWgCLsKJ2u/32---Export?node-id=1301-7740&t=DmjJ8aRTUBC9Xhkk-0

dozzzzer:
@grliszas14 
1. When the output directory is left blank, clicking Export causes a crash;
2. When the output directory is set to a non-existing folder inside an existing folder (e.g. `users/username/Desktop/<custom_name>`), clicking Export causes a crash;
3. Exporting audio to WMA causes a crash;
4. When FFmpeg installation isn't found, an attempt to export causes a crash (Windows).

dozzzzer:
@grliszas14 

Regardless of the selected no. of channels (mono/stereo), the output file is always stereo

dozzzzer:
Tested. New issues are created: https://github.com/audacity/audacity/issues/9138, https://github.com/audacity/audacity/issues/9137

PR #8960, opened 2025-06-19
Basic ffmpeg export with preference dialog

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/8897

This PR introduces basic export functionality with FFMPEG. It adds an ExportOptionsDialog as well.
For now you can:

- export full project or selection only audio
- change export sample rate
- set mono or stereo export
- change filename/directory
- choose from a few ffmpeg formats

For now I have used AU3 code -> but as it is tightly coupled with wxWidgets UI code I had to comment out a lot of its' code. Ideally we should extract the code we need to a separate AU4 classes - but to do that I need more time.
No navigation in the dialog is added yet. Will add missing pieces as a separate tickets.

- [] If changes are extensive, then there is a sequence of easily reviewable commits

Recommended:

### DIFF
--- a/au3/modules/import-export/mod-ffmpeg/ExportFFmpeg.cpp
+++ b/au3/modules/import-export/mod-ffmpeg/ExportFFmpeg.cpp
@@ -27,10 +27,6 @@ function.
 #include <wx/app.h>
 #include <wx/log.h>
 
-// #include <wx/window.h>
-// #include <wx/button.h>
-// #include <wx/textctrl.h>
-
 #include "BasicSettings.h"
 #include "Mix.h"
 #include "libraries/lib-tags/Tags.h"
@@ -48,12 +44,9 @@ function.
 #define OSINPUT(X) OSFILENAME(X)
 #endif
 
-// #include "ShuttleGui.h"
-
 #include "libraries/lib-import-export/ExportPluginHelpers.h"
 #include "libraries/lib-import-export/PlainExportOptionsEditor.h"
 #include "FFmpegDefines.h"
-// #include "ExportOptionsUIServices.h"
 #include "libraries/lib-import-export/ExportPluginRegistry.h"
 
 #if defined(WIN32) && _MSC_VER < 1900
@@ -65,18 +58,18 @@ function.
 
 static int AdjustFormatIndex(int format)
 {
-    // int subFormat = -1;
-    // for (int i = 0; i <= FMT_OTHER; i++) {
-    //     if (ExportFFmpegOptions::fmts[i].compiledIn) {
-    //         subFormat++;
-    //     }
-    //     if (subFormat == format || i == FMT_OTHER) {
-    //         subFormat = i;
-    //         break;
-    //     }
-    // }
-    // return subFormat;
-    return format;
+    int subFormat = -1;
+    for (int i = 0; i <= FMT_OTHER; i++) {
+        if (ExportFFmpegOptions::fmts[i].compiledIn) {
+            subFormat++;
+        }
+        if (subFormat == format || i == FMT_OTHER) {
+            subFormat = i;
+            break;
+        }
+    }
+
+    return subFormat;
 }
 
 namespace {
--- a/au3/modules/import-export/mod-ffmpeg/ExportFFmpeg.cpp
+++ b/au3/modules/import-export/mod-ffmpeg/ExportFFmpeg.cpp
@@ -36,20 +36,14 @@ function.
 #include "wxFileNameWrapper.h"
 
 #include "ExportFFmpegOptions.h"
-// #include "SelectFile.h"
 #if defined(__WXMSW__)
-// Note, on Windows we don't define an OSFILENAME() to prevent accidental use.
-// See VerifyFilename() for an explanation.
-#define OSINPUT(X) VerifyFilename(X, true)
-#define OSOUTPUT(X) VerifyFilename(X, false)
+#define OSINPUT(X) ((X).mb_str() ? (char*)(const char*)(X).mb_str() : "")
 #elif defined(__WXMAC__)
 #define OSFILENAME(X) ((char*)(const char*)(X).fn_str())
 #define OSINPUT(X) OSFILENAME(X)
-#define OSOUTPUT(X) OSFILENAME(X)
 #else
 #define OSFILENAME(X) ((char*)(const char*)(X).mb_str())
 #define OSINPUT(X) OSFILENAME(X)
-#define OSOUTPUT(X) OSFILENAME(X)
 #endif
 
 // #include "ShuttleGui.h"
@@ -806,7 +800,7 @@ ExportFFmpeg::CreateOptionsEditor(int format, ExportOptionsEditor::Listener* lis
 //     case FMT_OTHER:
 //         return {};//std::make_unique<ExportOptionsFFmpegCustomEditor>(listener);
 //     }
-//     return {};
+    return {};
 }
 
 int ExportFFmpeg::GetFormatCount() const
--- a/au3/modules/import-export/mod-ffmpeg/ExportFFmpeg.cpp
+++ b/au3/modules/import-export/mod-ffmpeg/ExportFFmpeg.cpp
@@ -18,32 +18,47 @@ function.
 
 *//*******************************************************************/
 
-#include "../FFmpeg.h"
-#include "FFmpegFunctions.h"
-#include "FifoBuffer.h"
+#include "FFmpeg.h"
+#include "lib-ffmpeg-support/FFmpegFunctions.h"
+#include "lib-ffmpeg-support/FifoBuffer.h"
 
 #include <wx/app.h>
 #include <wx/log.h>
 
-#include <wx/window.h>
-#include <wx/button.h>
-#include <wx/textctrl.h>
+// #include <wx/window.h>
+// #include <wx/button.h>
+// #include <wx/textctrl.h>
 
 #include "BasicSettings.h"
 #include "Mix.h"
-#include "Tags.h"
+#include "libraries/lib-tags/Tags.h"
 #include "Track.h"
 #include "wxFileNameWrapper.h"
 
 #include "ExportFFmpegOptions.h"
-#include "SelectFile.h"
-#include "ShuttleGui.h"
+// #include "SelectFile.h"
+#if defined(__WXMSW__)
+// Note, on Windows we don't define an OSFILENAME() to prevent accidental use.
+// See VerifyFilename() for an explanation.
+#define OSINPUT(X) VerifyFilename(X, true)
+#define OSOUTPUT(X) VerifyFilename(X, false)
+#elif defined(__WXMAC__)
+#define OSFILENAME(X) ((char*)(const char*)(X).fn_str())
+#define OSINPUT(X) OSFILENAME(X)
+#define OSOUTPUT(X) OSFILENAME(X)
+#else
+#define OSFILENAME(X) ((char*)(const char*)(X).mb_str())
+#define OSINPUT(X) OSFILENAME(X)
+#define OSOUTPUT(X) OSFILENAME(X)
+#endif
 
-#include "ExportPluginHelpers.h"
-#include "PlainExportOptionsEditor.h"
+// #include "ShuttleGui.h"
+
+#include "libraries/lib-import-export/ExportPluginHelpers.h"
+#include "libraries/lib-import-export/PlainExportOptionsEditor.h"
 #include "FFmpegDefines.h"
-#include "ExportOptionsUIServices.h"
-#include "ExportPluginRegistry.h"
+// #include "ExportOptionsUIServices.h"
+#include "libraries/lib-import-export/ExportPluginRegistry.h"
 
 #if defined(WIN32) && _MSC_VER < 1900
 #define snprintf _snprintf
@@ -54,17 +69,18 @@ function.
 
 static int AdjustFormatIndex(int format)
 {
-    int subFormat = -1;
-    for (int i = 0; i <= FMT_OTHER; i++) {
-        if (ExportFFmpegOptions::fmts[i].compiledIn) {
-            subFormat++;
-        }
-        if (subFormat == format || i == FMT_OTHER) {
-            subFormat = i;
-            break;
-        }
-    }
-    return subFormat;
+    // int subFormat = -1;
+    // for (int i = 0; i <= FMT_OTHER; i++) {
+    //     if (ExportFFmpegOptions::fmts[i].compiledIn) {
+    //         subFormat++;
+    //     }
+    //     if (subFormat == format || i == FMT_OTHER) {
+    //         subFormat = i;
+    //         break;
+    //     }
+    // }
+    // return subFormat;
+    return format;
 }
 
 namespace {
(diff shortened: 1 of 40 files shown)


au3/modules/import-export/mod-ffmpeg/ExportFFmpeg.cpp
```diff
@@ -18,32 +18,47 @@ function.
...
 #include "wxFileNameWrapper.h"
 
 #include "ExportFFmpegOptions.h"
-#include "SelectFile.h"
-#include "ShuttleGui.h"
+// #include "SelectFile.h"
+#if defined(__WXMSW__)
+// Note, on Windows we don't define an OSFILENAME() to prevent accidental use.
+// See VerifyFilename() for an explanation.
+#define OSINPUT(X) VerifyFilename(X, true)
```
saintmatthieu: Compilation of this fails on Windows.
I think you can do
```cpp
#include "ExportFFmpegOptions.h"
#if defined(__WXMSW__)
#define OSINPUT(X) ((X).mb_str() ? (char*)(const char*)(X).mb_str() : "")
#elif defined(__WXMAC__)
#define OSFILENAME(X) ((char*)(const char*)(X).fn_str())
#define OSINPUT(X) OSFILENAME(X)
#else
#define OSFILENAME(X) ((char*)(const char*)(X).mb_str())
#define OSINPUT(X) OSFILENAME(X)
#endif
```
if there is no use for `OSOUTPUT` anymore.

src/importexport/export/internal/au3/au3exporter.cpp
```diff
@@ -0,0 +1,226 @@
...
+bool Au3Exporter::doExport()
+{
+    muse::io::path_t directoryPath = exportConfiguration()->directoryPath();
+    muse::io::path_t filePath = directoryPath.appendingComponent(exportConfiguration()->filename())
+                                .appendingSuffix(formatExtension(exportConfiguration()->currentFormat()));
+
+    wxFileName filename = wxString(filePath.toStdString());
+
+    Au3Project* project = reinterpret_cast<Au3Project*>(globalContext()->currentProject()->au3ProjectPtr());
+    if (!project) {
```
saintmatthieu: Shouldn't this be an assert?

src/importexport/export/iexporter.h
```diff
@@ -0,0 +1,22 @@
...
+
+namespace au::importexport {
+class IExporter : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IExporter)
+
+public:
+    virtual ~IExporter() = default;
+
+    virtual bool doExport() = 0;
```
saintmatthieu: The `do` prefix for a public interface ?

src/importexport/export/iexportconfiguration.h
```diff
@@ -0,0 +1,52 @@
...
+namespace au::importexport {
+class IExportConfiguration : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IExportConfiguration)
+
+public:
+
+    virtual ~IExportConfiguration() = default;
+
+    virtual ProcessType process() const = 0;
```
saintmatthieu: Not `processType()` ? `process` is so commonly used in audio processing for actual processing, I think this shortcut has a potential for confusion.
To avoid ambiguities even more, I suggest to rename `ProcessType` `ExportProcessType` ...

src/appshell/view/preferences/exportpreferencesmodel.h
```diff
@@ -0,0 +1,98 @@
...
+    Q_PROPERTY(QString currentProcess READ currentProcess NOTIFY currentProcessChanged)
+    Q_PROPERTY(QVariantList processList READ processList NOTIFY processListChanged)
+
+    Q_PROPERTY(QString filename READ filename NOTIFY filenameChanged)
+    Q_PROPERTY(QString fileExtension READ fileExtension NOTIFY fileExtensionChanged)
+
+    Q_PROPERTY(QString directoryPath READ directoryPath NOTIFY directoryPathChanged)
+
+    Q_PROPERTY(QString currentFormat READ currentFormat NOTIFY currentFormatChanged)
+    Q_PROPERTY(QVariantList formatList READ formatList NOTIFY formatListChanged)
```
saintmatthieu: Nouns used as adjectives (e.g. "format" in "format list" of "chip" in "fish and chip shop") usually don't take an "s", even when referring to a plural quantity (e.g. in "fish and chip shop", there's obviously more than one chip...), and I tend to stick to this rule ...
But in this case, "format" very much could be the verb "to format" in imperative form, and indeed, I've been reading this method in different places in this PR, and each time thought it were a verb.
So in this case I'd rather go for `formatsList` :D
A long comment for a minor concern, but sure, it's Friday evening ...

src/importexport/export/internal/au3/au3exporter.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    m_selectedOnly = false;
+
+    auto exportedTracks = ExportUtils::FindExportWaveTracks(TrackList::Get(*project), m_selectedOnly);
+    if (exportedTracks.empty()) {
+        //! NOTE: All selected audio is muted
+        return false;
+    }
+
+    m_mixerSpec = std::make_unique<MixerOptions::Downmix>(exportedTracks.size(), 2).get();
```
saintmatthieu: `constexpr auto maxNumberOfChannels = 2;` would be nice.

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include "exportpreferencesmodel.h"
+
+using namespace au::importexport;
+
+std::map<ProcessType, muse::TranslatableString> EXPORT_PROCESS_MAPPING {
+    { ProcessType::FULL_PROJECT_AUDIO, muse::TranslatableString("export", "Export full project audio") },
+    { ProcessType::SELECTED_AUDIO, muse::TranslatableString("export", "Export selected audio") },
+    { ProcessType::AUDIO_IN_LOOP_REGION, muse::TranslatableString("export", "Export audio in loop region") },
+    { ProcessType::TRACKS_AS_SEPARATE_AUDIO_FILES, muse::TranslatableString("export", "Export tracks as a separate audio files (Stems") },
```
saintmatthieu: `(Stems)`

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,217 @@
...
+            return rate.second;
+        }
+    }
+
+    return {};
+}
+
+QVariantList ExportPreferencesModel::exportSampleRateList()
+{
+    //! NOTE: this list should be fetched based on selected plugin
```
saintmatthieu: This sounds like a potential bug. Is there a follow-up task?

src/appshell/view/preferences/exportpreferencesmodel.h
```diff
@@ -0,0 +1,98 @@
...
+    QString currentFormat() const;
+    Q_INVOKABLE void setCurrentFormat(const QString& format);
+    QVariantList formatList() const;
+
+    ExportChannelsPref::ExportChannels exportChannels() const;
+    Q_INVOKABLE void setExportChannels(ExportChannelsPref::ExportChannels exportChannels);
+
+    QString exportSampleRate() const;
+    QVariantList exportSampleRateList();
+    Q_INVOKABLE void exportSampleRateSelected(const QString& rate);
```
saintmatthieu: Why not the usual `setExportSampleRate` ?

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    auto it = std::find_if(m_sampleRateMapping.begin(), m_sampleRateMapping.end(),
+                           [&rateName](const auto& rate) { return rateName == rate.second; });
+    if (it != m_sampleRateMapping.end()) {
+        exportConfiguration()->setExportSampleRate(it->first);
+        return;
+    }
+
+    emit exportSampleRateChanged();
```
saintmatthieu: Shouldn't this be under `exportConfiguration()->setExportSampleRate(it->first);` ?

src/appshell/qml/Preferences/ExportOptionsDialog.qml
```diff
@@ -0,0 +1,420 @@
...
+
+    title: qsTrc("export", "Export audio")
+
+    contentWidth: 546
+    contentHeight: mainColumn.implicitHeight + 2 * mainColumn.anchors.margins
+
+    property int dropdownWidth: 354
+    property int labelColumnWidth: 80
+
+    // property NavigationPanel navigation: NavigationPanel {
```
saintmatthieu: Follow-up ticket to add navigation ?

au3/modules/import-export/mod-ffmpeg/FFmpeg.cpp
```diff
@@ -21,6 +21,7 @@ License: GPL v2 or later.  See License.txt.
 #include "lib-ffmpeg-support/FFmpegFunctions.h"
 #include "ModuleConstants.h"
 #include "ImportFFmpeg.cpp"
+#include "ExportFFmpeg.cpp"
```
saintmatthieu, line 24: I'm curious, why isn't "ExportFFmpeg.h" sufficient (and likewise for import) ? It's part of the same module, isn't it?

saintmatthieu (commented):
A few comments and questions already. I'll have to look again at the changes in the AU3 export module (where lots got commented out) to understand them, but it's a start.

saintmatthieu:
I get an exception when exporting with the first codec in the list, I think, on Windows:
image
The same export format works with AU3.

src/importexport/export/iexporter.h
grliszas14: maybe `exportData()` or `performExport()` or `runExport()` then?
I can't use bare `export()` as it's reserved C++ keyword

src/appshell/qml/Preferences/ExportOptionsDialog.qml
grliszas14: Yup, there is one: https://github.com/audacity/audacity/issues/8961

src/importexport/export/internal/au3/au3exporter.cpp
grliszas14: Well, for now yes, but I'm not sure if we cannot set it to more channels later so there's no sense in making it const?

src/appshell/view/preferences/exportpreferencesmodel.cpp
grliszas14: good catch!

au3/modules/import-export/mod-ffmpeg/FFmpeg.cpp
grliszas14: There's some chain dependency, I'll check if I can easily break it down

src/appshell/view/preferences/exportpreferencesmodel.cpp
grliszas14: Fixed

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.cpp
```diff
@@ -15,33 +15,34 @@
 
 #include "ExportFFmpegOptions.h"
 
-#include <wx/listbox.h>
-#include <wx/combobox.h>
-#include <wx/stattext.h>
+// #include <wx/listbox.h>
```
kryksyh, line 18: Just curious, if you comment out the whole function body, why are there changes in the commented out code?

src/project/internal/projectactionscontroller.cpp
```diff
@@ -523,6 +525,9 @@ Ret ProjectActionsController::doOpenProject(const io::path_t& filePath)
 
     projectHistory()->init();
 
+    exportConfiguration()->setFilename("untitled");
```
kryksyh: project->displayName()?

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.cpp
grliszas14: hmmmm, github doesn't highlight those for me so not sure where to look at but the case might be that I first tried to make these functions compile before I realized that I need to comment them out entirely

src/project/internal/projectactionscontroller.cpp
grliszas14: yesss, good idea, fixed

grliszas14:
> Not sure about all that commented code, do we want to push it as is and cleanup later?

Well, that was my idea - I was a bit in a rush to push those changes a few weeks ago. I can clean it up now, but not all of it anyway as I'm not sure if some of the commented code is still going to be needed.

au3/modules/import-export/mod-ffmpeg/FFmpeg.cpp
kryksyh: I believe both could be safely removed now, `"ImportFFmpeg.cpp"` is only needed to create temporary `FFmpegImportPlugin` which seemingly does nothing, and `"ExportFFmpeg.cpp"` is no longer used

src/importexport/export/internal/exportconfiguration.cpp
```diff
@@ -0,0 +1,174 @@
...
+    muse::settings()->valueChanged(EXPORT_FORMAT).onReceive(nullptr, [this] (const muse::Val& val) {
+        m_currentFormatChanged.notify();
+    });
+
+    muse::settings()->setDefaultValue(EXPORT_CHANNELS, muse::Val(ExportChannelsPref::ExportChannels::STEREO));
+    muse::settings()->valueChanged(EXPORT_CHANNELS).onReceive(nullptr, [this] (const muse::Val& val) {
+        m_exportChannelsChanged.notify();
+    });
+
+    muse::settings()->setDefaultValue(EXPORT_SAMPLE_RATE, muse::Val("44100"));
```
kryksyh, line 42: We should fallback to project's sample rate here, or rather "same or nearest up" supported by the selected codec. 
See `ExportAudioDialog::ExportAudioDialog` in AU3

grliszas14: Fixed default sample rate (audioDevicesProvider()->defaultSampleRate() should be the one as we now have single setting for sample rate instead of having separate default and separate for project)

but I'll surprise you with this:
> I believe both could be safely removed now, "ImportFFmpeg.cpp" is only needed to create temporary FFmpegImportPlugin which seemingly does nothing, and "ExportFFmpeg.cpp" is no longer used

We actually need both :P temporary `FFmpegImportPlugin` makes the linker look at both ImportFFmpeg.cpp and ExportFFmpeg.cpp somehow, both of these files contain lines like this:
```c++
static ExportPluginRegistry::RegisteredPlugin sRegisteredPlugin{ "FFmpeg",
                                                                 []{ return std::make_unique< ExportFFmpeg >(); }
};
``` 
`static`, once created, lives all the time that app is alive so we need to force triggering this code at least once (at the beginning), and that is how import/export plugins are registered. Static initialization works in AU3 without issues, there's no need for calling FFmpegImportPlugin at all, but for some unknown for me reason it doesn't work with AU4 at that point. I tried with a bunch of different linker flags (I think Igor tried as well?) but without a success so far

kryksyh: > Fixed default sample rate (audioDevicesProvider()->defaultSampleRate() should be the one as we now have single setting for sample rate instead of having separate default and separate for project)

I believe there is a misunderstanding; the playback sample rate is not directly linked to the track's sample rate.

We can have a 192kHz sample rate for a track, and the playback set to 44.1 kHz playback sample rate. Then during playback track data will be downsampled to 44.1.

When we are exporting the project, by default we want to preserve as much data as possible, thus we need to select exporting sample rate the same as the track's sample rate.

https://github.com/user-attachments/assets/86fd0edc-29ec-4178-a02f-4adcc0581aac

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    return {};
+}
+
+QVariantList ExportPreferencesModel::exportSampleRateList()
+{
+    std::vector<int> sampleRateList = exporter()->sampleRateList();
+    if (sampleRateList.empty()) {
+        sampleRateList = DEFAULT_SAMPLE_RATE_LIST;
```
Eism: If the exporter doesn’t support sampleRate (returns an empty list), then we can’t forcibly set a specific sampleRate. Am I right? 
(The model is supposed to only receive data from the service and display it - without adding any extra logic)

src/importexport/export/internal/exportconfiguration.h
```diff
@@ -0,0 +1,65 @@
...
+
+    std::string filename() const override;
+    void setFilename(const std::string& filename) override;
+    muse::async::Notification filenameChanged() const override;
+
+    muse::io::path_t directoryPath() const override;
+    void setDirectoryPath(const muse::io::path_t& path) override;
+    muse::async::Notification directoryPathChanged() const override;
+
+    ExportChannelsPref::ExportChannels exportChannels() const override;
```
Eism: We shouldn’t mix UI with services. Here in the configurator, you’re referencing a Q_GADGET type that belongs to the UI layer. This should be kept as a plain type for internal use. Then, in the model, you can return a list of available types along with their corresponding names - so you won’t need to match types to names directly in QML.

src/importexport/export/internal/exportconfiguration.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+void ExportConfiguration::init()
+{
+    muse::settings()->setDefaultValue(EXPORT_PROCESS, muse::Val(ExportProcessType::FULL_PROJECT_AUDIO));
+    muse::settings()->valueChanged(EXPORT_PROCESS).onReceive(nullptr, [this] (const muse::Val& val) {
+        m_processChanged.notify();
+    });
+
+    //! NOTE: "untitled" filename is set on every project load
+    muse::settings()->setDefaultValue(EXPORT_FILENAME, muse::Val("untitled"));
```
Eism: As I see it, "untitled" in AU3 is translatable, so we should translate it as well. It’s better to handle the translation in the model, while the configurator should simply return an empty string.

src/importexport/export/internal/exportconfiguration.cpp
```diff
@@ -0,0 +1,174 @@
...
+    muse::settings()->valueChanged(EXPORT_FILENAME).onReceive(nullptr, [this] (const muse::Val& val) {
+        m_filenameChanged.notify();
+    });
+
+    //! NOTE: export directory is set on every project load
+    muse::settings()->valueChanged(EXPORT_DIRECTORY_PATH).onReceive(nullptr, [this] (const muse::Val& val) {
+        m_directoryPathChanged.notify();
+    });
+
+    if (!exporter()->formatsList().empty()) {
```
Eism: The configurator shouldn’t be aware of services - this kind of logic belongs in the model.

src/importexport/export/internal/exportconfiguration.h
```diff
@@ -0,0 +1,65 @@
...
+    muse::Inject<playback::IAudioDevicesProvider> audioDevicesProvider;
+    muse::Inject<au::importexport::IExporter> exporter;
+
+public:
+    ExportConfiguration() = default;
+
+    void init();
+
+    ExportProcessType processType() const override;
+    void setProcess(ExportProcessType process) override;
```
Eism: setProcessType and processTypeChanged (as for getter)

src/importexport/export/internal/exportconfiguration.h
```diff
@@ -0,0 +1,65 @@
...
+#include "../iexporter.h"
+#include "playback/iaudiodevicesprovider.h"
+
+#include "../iexportconfiguration.h"
+
+namespace au::importexport {
+class ExportConfiguration : public IExportConfiguration
+{
+    muse::Inject<muse::IGlobalConfiguration> globalConfiguration;
+    muse::Inject<playback::IAudioDevicesProvider> audioDevicesProvider;
```
Eism: you don't use it

src/project/internal/projectactionscontroller.cpp
```diff
@@ -523,6 +525,9 @@ Ret ProjectActionsController::doOpenProject(const io::path_t& filePath)
 
     projectHistory()->init();
 
+    exportConfiguration()->setFilename(project->displayName().toStdString());
```
Eism: It’s no good when the configurator changes every time a project is opened. Instead, you should retrieve the necessary data directly from the project where it’s needed.

src/project/internal/projectactionscontroller.cpp
```diff
@@ -624,7 +629,16 @@ void ProjectActionsController::clearRecentProjects()
 
 void ProjectActionsController::exportAudio()
 {
-    NOT_IMPLEMENTED;
+    IAudacityProjectPtr project = globalContext()->currentProject();
+    auto ret = interactive()->open(EXPORT_OPTIONS_URI);
+
+    ret.onResolve(this, [=](Val ret){
```
Eism: Actually, I prefer the MuseScore approach - calling a URI that handles the entire export process. In your version, you’re waiting for completion and then calling export again.

Another point: if we go with what I suggested above, we’ll need to abandon the exportAudio method in the project, since at that stage we’ll already be inside the export module and can simply call the exporter with the project as a parameter.

src/importexport/export/iexporter.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace au::importexport {
+class IExporter : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IExporter)
+
+public:
+    virtual ~IExporter() = default;
+
+    virtual muse::Ret exportData() = 0;
```
Eism: maybe just export

grliszas14: That was my initial thought but `export` is reserved C++ keyword: https://github.com/audacity/audacity/pull/8960#discussion_r2190491949

src/appshell/view/preferences/exportpreferencesmodel.cpp
grliszas14: Well, it's other way round I methinks. I you get empty supported sample rate list then codec does not impose any hard restrictions on sample rates. Take a look at M4A format for example: exporter returns empty list but it allows you to set anything you want including totally custom number (in AU3)

src/project/internal/projectactionscontroller.cpp
```diff
@@ -624,7 +626,8 @@ void ProjectActionsController::clearRecentProjects()
 
 void ProjectActionsController::exportAudio()
 {
-    NOT_IMPLEMENTED;
+    IAudacityProjectPtr project = globalContext()->currentProject();
+    auto ret = interactive()->open(EXPORT_OPTIONS_URI);
```
Eism: Since you’ve moved the export logic inside the export dialog, we can name the URI like EXPORT_URI.

src/project/internal/projectactionscontroller.cpp
```diff
@@ -624,7 +626,8 @@ void ProjectActionsController::clearRecentProjects()
 
 void ProjectActionsController::exportAudio()
 {
-    NOT_IMPLEMENTED;
+    IAudacityProjectPtr project = globalContext()->currentProject();
```
Eism: you don't use it

src/project/internal/projectactionscontroller.cpp
```diff
@@ -20,6 +20,8 @@ static const muse::Uri PROJECT_PAGE_URI("audacity://project");
 static const muse::Uri HOME_PAGE_URI("musescore://home");
 static const muse::Uri NEW_PROJECT_URI("audacity://project/new");
 
+static const muse::Uri EXPORT_OPTIONS_URI("audacity://exportoptions");
```
Eism: please rename it to `audacity://project/export`

src/project/internal/projectactionscontroller.h
```diff
@@ -34,6 +35,7 @@ class ProjectActionsController : public IProjectFilesController, public muse::ac
     muse::Inject<IOpenSaveProjectScenario> openSaveProjectScenario;
     muse::Inject<trackedit::IProjectHistory> projectHistory;
     muse::Inject<record::IRecordController> recordController;
+    muse::Inject<au::importexport::IExportConfiguration> exportConfiguration;
```
Eism: you don't use it

src/importexport/export/internal/exportconfiguration.cpp
```diff
@@ -0,0 +1,148 @@
...
+*/
+
+#include <QDir>
+
+#include "global/settings.h"
+#include "log.h"
+
+#include "types/exporttypes.h"
+
+#include "exportconfiguration.h"
```
Eism:
```suggestion
#include "exportconfiguration.h"

#include <QDir>

#include "global/settings.h"
#include "types/exporttypes.h"

#include "log.h"
```

Eism: you don't use QDir

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,307 @@
...
+    if (static_cast<int>(exportChannels()) > maxChannels) {
+        setExportChannels(ExportChannelsPref::ExportChannels(maxChannels));
+    }
+}
+
+bool ExportPreferencesModel::verifyExportPossible()
+{
+    QDir directory(directoryPath());
+    if (!directory.exists() || directoryPath().isEmpty()) {
+        interactive()->error(muse::trc("export", "Export Audio"), muse::trc("export", "Unable to create destination folder"));
```
Eism: hmm, but you’re not trying to create the folder - looks like you forgot to add makePath. And yes, please use FileSystem - filesystem operations should be mockable for tests.

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,307 @@
...
+    Q_UNUSED(format);
+    NOT_IMPLEMENTED;
+}
+
+void ExportPreferencesModel::updateCurrentSampleRate()
+{
+    int currentSampleRate = exportConfiguration()->exportSampleRate();
+    std::vector<int> sampleRateList = exporter()->sampleRateList();
+    for (const auto& rate : sampleRateList) {
+        if (rate == currentSampleRate) {
```
Eism: you can use `muse::contains`

src/appshell/qml/Preferences/ExportOptionsDialog.qml
Eism, line 1: ahh, this dialog should be placed in the export module - it doesn’t belong in Preferences. Everything under src/appshell/qml/Preferences/ relates specifically to the PreferencesDialog.

src/appshell/qml/Preferences/ExportOptionsDialog.qml
```diff
@@ -0,0 +1,399 @@
...
+                    Row {
+                        width: parent.width
+                        spacing: 10
+
+                        RoundedRadioButton {
+                            id: monoBtn
+
+                            checked: exportPreferencesModel.exportChannels == ExportChannels.MONO
+                            enabled: exportPreferencesModel.maxExportChannels > 0
+                            text: qsTrc("export", "Mono")
```
Eism: Let’s use a list from the model to generate the list of radio buttons, just like you did for formatsList.

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,307 @@
...
+#include "types/translatablestring.h"
+
+#include "importexport/export/types/exporttypes.h"
+
+#include "exportpreferencesmodel.h"
+
+using namespace au::importexport;
+
+std::map<ExportProcessType, muse::TranslatableString> EXPORT_PROCESS_MAPPING {
+    { ExportProcessType::FULL_PROJECT_AUDIO, muse::TranslatableString("export", "Export full project audio") },
```
Eism: use `muse::trc` for it

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,307 @@
...
+#include <QDir>
+
+#include "global/containers.h"
+#include "settings.h"
+#include "log.h"
+#include "types/translatablestring.h"
+
+#include "importexport/export/types/exporttypes.h"
+
+#include "exportpreferencesmodel.h"
```
Eism:
```suggestion
#include "exportpreferencesmodel.h"

#include <QDir>

#include "global/containers.h"
#include "types/exporttypes.h"

#include "translation.h"
#include "log.h"
```

src/appshell/qml/Preferences/ExportOptionsDialog.qml
```diff
@@ -0,0 +1,399 @@
...
+
+    ExportPreferencesModel {
+        id: exportPreferencesModel
+    }
+
+    ColumnLayout {
+        id: mainColumn
+
+        anchors.fill: parent
+        anchors.margins: 20
```
Eism: We follow the approach where the content is simply rendered without constraints, and the dialog determines its size based on that content. So you shouldn’t use anchors.fill. Instead, contentHeight should be set to mainColumn.implicitHeight.

Also StyledDialogView has margins property

src/appshell/view/preferences/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,307 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+void ExportPreferencesModel::exportData()
+{
+    muse::Ret result = exporter()->exportData(filename().toStdString());
+    if (!result.success() && !result.text().empty()) {
```
Eism: or just `!result` ?

src/appshell/view/preferences/exportpreferencesmodel.h
```diff
@@ -0,0 +1,110 @@
...
+    Q_PROPERTY(QString filename READ filename NOTIFY filenameChanged)
+    Q_PROPERTY(QString fileExtension READ fileExtension NOTIFY fileExtensionChanged)
+
+    Q_PROPERTY(QString directoryPath READ directoryPath NOTIFY directoryPathChanged)
+
+    Q_PROPERTY(QString currentFormat READ currentFormat NOTIFY currentFormatChanged)
+    Q_PROPERTY(QVariantList formatsList READ formatsList NOTIFY formatsListChanged)
+
+    Q_PROPERTY(ExportChannelsPref::ExportChannels exportChannels READ exportChannels NOTIFY exportChannelsChanged)
+    Q_PROPERTY(int maxExportChannels READ maxExportChannels NOTIFY maxExportChannelsChanged)
```
Eism: Here, you’re returning the count as an int, but just above, it’s ExportChannelsPref::ExportChannels. I’d assume you don’t actually need ExportChannelsPref at all.

src/appshell/view/preferences/exportpreferencesmodel.h
```diff
@@ -0,0 +1,110 @@
...
+
+#include "modularity/ioc.h"
+#include "context/iglobalcontext.h"
+#include "iinteractive.h"
+
+#include "importexport/export/iexporter.h"
+#include "importexport/export/internal/exportconfiguration.h"
+#include "playback/iaudiodevicesprovider.h"
+
+using namespace au::importexport;
```
Eism: please don't use using in headers, this can lead to linking issues.

src/importexport/export/internal/au3/au3exporter.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "MixerOptions.h"
+#include "Track.h"
+#include "WaveTrack.h"
+#include "au3wrap/au3types.h"
+#include "libraries/lib-import-export/ExportPluginRegistry.h"
+#include "libraries/lib-import-export/ExportUtils.h"
+
+#include "translation.h"
+
+#include "au3exporter.h"
```
Eism:
```suggestion
#include "au3exporter.h"

// please use `libraries/lib-...` for such includes
#include "BasicUI.h"
#include "MixerOptions.h"
#include "Track.h"
#include "WaveTrack.h"
#include "libraries/lib-import-export/ExportPluginRegistry.h"
#include "libraries/lib-import-export/ExportUtils.h"

#include "au3wrap/au3types.h"

#include "translation.h"
```

src/appshell/qml/Preferences/ExportOptionsDialog.qml
grliszas14: I will take care of that in upcoming ticket as I will have to modify export channels later, all noted here: https://github.com/audacity/audacity/issues/8963

src/appshell/view/preferences/exportpreferencesmodel.h
grliszas14: Well, it will be useful, maybe it's hard to see now but we have something like "custom mapping" which we can express as an enum in the radio button group in the ExportOptionsDialog, but under the hood custom mapping can have any number of channels

src/importexport/export/view/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,299 @@
...
+    48000,
+    88200,
+    96000,
+    176400,
+    192000,
+    352800,
+    384000
+};
+
+namespace au::importexport {
```
Eism: you don't need it

src/importexport/export/view/exportpreferencesmodel.cpp
```diff
@@ -0,0 +1,299 @@
...
+{
+    return m_filename;
+}
+
+void ExportPreferencesModel::setFilename(const QString& filename)
+{
+    if (m_filename != filename) {
+        m_filename = filename;
+        emit filenameChanged();
+    }
```
Eism:
```suggestion
    if (m_filename != filename) {
        return;
     }
     
     m_filename = filename;
     emit filenameChanged();
```

PR #8967, opened 2025-06-21
Implement Wave Colour menu

MargaridaLopes12341:
This feature adds a Wave Colour menu similar to the existing Clip Colour menu, allowing users to change the color of waveforms in the UI. Users can right-click on a clip and select the Wave Colour option to choose a new color for the waveform, improving project organization and visual clarity.

Recommended:

### DIFF
--- a/au3/libraries/lib-wave-track-paint/waveform/WaveBitmapCache.cpp
+++ b/au3/libraries/lib-wave-track-paint/waveform/WaveBitmapCache.cpp
@@ -47,6 +47,13 @@ struct Triplet final
         b = color.GetBlue();
     }
 
+    void SetWaveColor(graphics::Color color)
+    {
+        r = color.GetRed();
+        g = color.GetGreen();
+        b = color.GetBlue();
+    }
+
     Triplet(const Triplet&) = default;
     Triplet(Triplet&&) = default;
     Triplet& operator=(const Triplet&) = default;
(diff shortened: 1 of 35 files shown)


LWinterberg:
Thanks for your contribution @MargaridaLopes12341! It's great to see you've been able to get something going in Audacity 4!

For waveforms, we have some other ideas down the road, eg a 3-band waveform (as seen in DJ tools, see eg [here](https://reallychrism.substack.com/p/the-secret-language-of-waveforms)) which would clash with your idea. I also think the clip color already does plenty for project organization – I think we should try it out on real users (ie with the launch of Audacity 4) before deciding that customizable waveform colors on top of clip colors should be a thing. 

I'm closing this PR because of that, still though – thank you very much for sending it in! I hope you had fun creating it :)

Tantacrul:
I've a quick thought about this @MargaridaLopes12341 

- @LWinterberg is right that we've some more interesting ideas coming down the pipe, although I could see this idea working, but with a few alterations.
- First, the wave colours don't work in the new clip view (where each clip has a different BG). This is simply because the colours don't match at all, and they just look extremely strange in almost any combination.
- However, they do work in the 'classic' view, which matches AU3
- Next, if there was a track setting for these - which allowed a user to apply the waveform colours to a whole track rather than just per-clip, it'd have much more utility

---

If you'd be interested, my suggestion would be:

1. Only have this feature work for the 'classic' view. Users shouldn't see these option in the other view
image

2. Allow users to apply the changes to a whole track also. This would be triggered from the track menu, which can be seen here.

image


As long as these don't appear from the multi-coloured clip view, I think they have utility.  What do you think?

PR #8971, opened 2025-06-24
Space is reserved for playback toggle

### LINKED ISSUE #8751: After an effect is added to the stack, playing/pausing by pressing the Space key no longer works
### Bug description

_No response_

### Steps to reproduce

1. Load a clip
2. Add an effect to the stack
3. Press Space on the keyboard
4. Observe: playback won't start; instead, the effect in the stack has been turned off
5. Close the effect's UI
6. Press Space again
7. Observe: playback won't start either

### Expected behavior

Pressing Space should start/pause the playback, regardless of whether the effect UI is open.

### Actual behavior

Pressing Space after adding an effect no longer controls the playback, even after reopening the project. To do that, you need to restart Audacity.

### Audacity Version

master

### Operating system

OS independent

### Additional context

_No response_

saintmatthieu:
Resolves: #8751

#8751: Like in Audacity 3, the space bar is reserved for playback toggle in appropriate focus contexts.

Recommended:

QA:
- [x] Keyboard shortcut reassignment still works for play/pause, and if another key is assigned to this action, Space behaves like enter on UI elements (e.g. it toggles the power button of an effect)

### DIFF
--- a/src/context/internal/uicontextresolver.cpp
+++ b/src/context/internal/uicontextresolver.cpp
@@ -35,6 +35,7 @@ static const Uri PROJECT_PAGE_URI("audacity://project");
 //! but we don't have that yet, so binding it to
 //! area that's being focused when opening a project
 static const QString PROJECT_NAVIGATION_PANEL("MainToolBar");
+static const QString PLAYBACK_TOOLBAR_NAVIGATION_PANEL("PlaybackToolBar");
 
 void UiContextResolver::init()
 {
@@ -114,7 +115,7 @@ UiContext UiContextResolver::currentUiContext() const
         if (activePanel) {
             if (activePanel->name() == PROJECT_NAVIGATION_PANEL) {
                 return context::UiCtxProjectFocused;
-            } else if (activePanel->name() != "ToolBarView") {
+            } else if (activePanel->name() != PLAYBACK_TOOLBAR_NAVIGATION_PANEL) {
                 return context::UiCtxProjectPlayback;
             }
         }
--- a/src/context/internal/uicontextresolver.cpp
+++ b/src/context/internal/uicontextresolver.cpp
@@ -35,7 +35,6 @@ static const Uri PROJECT_PAGE_URI("audacity://project");
 //! but we don't have that yet, so binding it to
 //! area that's being focused when opening a project
 static const QString PROJECT_NAVIGATION_PANEL("MainToolBar");
-static const QString PLAYBACK_TOOLBAR_NAVIGATION_PANEL("PlaybackToolBar");
 
 void UiContextResolver::init()
 {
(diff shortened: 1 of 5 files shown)


DavidBailes:
> #8751: Like in Audacity 3, the space bar is reserved for playback toggle.

This isn't the case in Audacity 3. Whatever shortcut is assigned to the play/stop command toggles playback. It isn't hardcoded to be space bar.

For example, if I assign W to this command, then in an realtime effect dialog, W toggles playback, and space bar does not toggle playback.

saintmatthieu:
Thanks @DavidBailes , I overlooked that.
The point of this implementation is rather that, independently on what UI element has keyboard focus, Space (or whatever it is reassigned to) toggles playback. Enter can still be used e.g. to toggle a power button.

DavidBailes:
> Enter can still be used e.g. to toggle a power button.

I'm not sure I fully understand the change to shortcuts.xml, but might this affect interacting with check boxes? eg in the realtime compressor effect on the various platforms? Just raising this in case there's an issue.
Using Audacity 3 on Windows, you can interact with such a check box by:
1. Pressing Ctrl + Spacebar (with is not used as an Audacity shortcut by default).
2. Using +(plus) and -(minus) to set or unset the checkbox (ancient Windows keystrokes with are not often used).
3. If another shortcut is assigned to play/stop, then spacebar can be used.

saintmatthieu:
I am not sure if, with the new framework, we can achieve this exact specification, but we should be able to get close to it.
I have to rework this PR anyway: Space for playback is appropriate in certain contexts, but not all of them (e.g. when the preference panel is open, or the macro manager), the amount of granularity that the simplistic approach I took cannot satisfy.

saintmatthieu:
I'm glad to see that you are still active on the Qt migration, @DavidBailes, that's a real asset to the accessibility dimension of the app.

DavidBailes:
> I am not sure if, with the new framework, we can achieve this exact specification, but we should be able to get close to it.

There's no need to have exactly the same behaviour as Audacity 3, just a need to be able interact with check boxes on all platforms.

saintmatthieu:
After a call with @Tantacrul , we decided to leave the pause commands and state as is, at least for now.
This PR now reduces to having the space bar toggling playback in appropriate focus contexts (e.g. with effect panel or realtime effect dialog in focus, but not for modals like preferences or pitch-and-speed).

DavidBailes:
> This PR now reduces to having the space bar toggling playback in appropriate focus contexts

@saintmatthieu . Wouldn't it be better to have the assigned shortcut for play/stop toggling playback in appropriate focus contexts? The user may have changed the shortcut from the default of spacebar.

Or is this something that will be sorted out in the future? And I guess other playback commands should work as well.

saintmatthieu:
@DavidBailes key reassignment should already work. I added a QA list item. My description was not general enough.

DavidBailes:
> @DavidBailes key reassignment should already work. I added a QA list item. My description was not general enough.

@saintmatthieu . Thanks for the clarification. (The reason I raised this is that in Audacity3 there's an example of a shortcut that was temporarily hard coded, and it's unfortunately still there!).

Eism (approved):

chinakov:
Keyboard re-assignment works, and space can still be used as an 'enter' key for accessibility. Thanks!

Issue #8574, opened 2025-03-31
AU plugins implementation


ianalarina:
### Description

Implement the ability to scan, load, and apply AU plugins in Audacity 4

dozzzzer:
@kryksyh 
### 1. When audio is being stopped first, and then played, the realtime effect's parameters get reset:

https://github.com/user-attachments/assets/b87c4031-d304-4845-b98b-bc443d534488

### 2. When the effect's power button state is off, changes to the parameters affect playback:

https://github.com/user-attachments/assets/8ccff34b-a18f-45f7-b4ad-7a3eb04b51d1

dozzzzer:
@kryksyh Issue no.2 is fixed; issue no.1 is not, but its behavior has changed: Now, the parameter is changing to a random value rather than resetting as before, and this only occurs the first time audio is played after the effect has been initialised.

https://github.com/user-attachments/assets/94ea9673-0c14-4070-a6f5-64c80baed65f

kryksyh:
@dozzzzer thanks for catching this! The parameter was changing not to a random number, but to the second to last batch of changes. Fixed in the last build.

dozzzzer:
@kryksyh I confirm the previously reported issue is fixed.

Another issue is that Audacity freezes if you:
1. Open an existing project with an AU effect in the stack
2. Open its UI and change one of the parameters
3. Save
4. Close the project

This can be reproduced using e.g. any of Melda's bundle effects.

https://github.com/user-attachments/assets/f3925f49-faeb-421f-adf7-b6225e0aad23


However, with some other effects (like [Deegain](https://dotec-audio.com/deegain.html), for example), when you try to do this, you will have Audacity crash, not freeze:

https://github.com/user-attachments/assets/2d01e23a-23b6-473d-b557-5eb284166484

Another issue is that for some plugins, the parameter defaults aren't initialised properly when they're added.

1. Add an instance of Deegain effect to the stack
2. Set the Gain to -20
3. Save & close the project
4. Open or create a new project
5. Add Deegain to the stack
6. Observe: Gain is set to -20 rather than 0 dB.

https://github.com/user-attachments/assets/0932a2e8-5352-48ea-a233-4dea2a6d2348

kryksyh:
@dozzzzer I pushed the changes, the first issue should be fixed. I can't reproduce the second one though

dozzzzer:
On Windows, EGATs cannot be accessed. Moving back to In Progress.

[...]

PR #9000, opened 2025-06-30
AudioUnit Support

kryksyh:
Resolves: [#8574](https://github.com/audacity/audacity/issues/8574)

Adds AudioUnit support on macOs

Recommended:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,8 +23,8 @@ set(FETCHCONTENT_BASE_DIR ${CMAKE_CURRENT_LIST_DIR}/_deps)
 include(FetchContent)
 FetchContent_Declare(
   muse_framework
-  GIT_REPOSITORY https://github.com/musescore/framework_tmp.git
-  GIT_TAG        u250618_1
+  GIT_REPOSITORY https://github.com/kryksyh/framework_tmp.git
+  GIT_TAG        audiounit_support
 )
 
 FetchContent_GetProperties(muse_framework)
@@ -120,6 +120,10 @@ set(AU_MODULE_VST_VST3_SDK_PATH "" CACHE PATH "[optional] Path to VST3_SDK. SDK 
 if (OS_IS_LIN)
     option(AU_MODULE_EFFECTS_LV2 "Build audacity lv2 module" ON)
 endif(OS_IS_LIN)
+
+if( CMAKE_SYSTEM_NAME MATCHES "Darwin" )
+    option(AU_MODULE_EFFECTS_AUDIO_UNIT "Build Audacity Audio Unit module" ON)
+endif()
 
 option(AU_USE_FFMPEG "Build mod-ffmpeg module" ON)
 if (AU_USE_FFMPEG)
--- a/au3/libraries/lib-audio-unit/AudioUnitWrapper.cpp
+++ b/au3/libraries/lib-audio-unit/AudioUnitWrapper.cpp
@@ -159,7 +159,7 @@ bool AudioUnitWrapper::FetchSettings(
         if (!GetFixedSizeProperty(kAudioUnitProperty_PresentPreset, preset)) {
             // Only want factory preset number, not a user preset (<0)
             if (preset.presetNumber >= 0) {
-                settings.mPresetNumber = { preset.presetNumber }
+                settings.mPresetNumber = { preset.presetNumber };
             }
         }
     }
@@ -495,12 +495,12 @@ bool AudioUnitWrapper::SetRateAndChannels(
                 --nChannels;
                 streamFormat.mChannelsPerFrame = nChannels;
                 failed = SetProperty(kAudioUnitProperty_StreamFormat,
-                                        streamFormat, scope);
+                                     streamFormat, scope);
             } while (failed && nChannels > 0);
             if (failed) {
                 wxLogError("%ls didn't accept stream format on %s\n",
-                            // Exposing internal name only in logging
-                            identifier.wx_str(), msg);
+                           // Exposing internal name only in logging
+                           identifier.wx_str(), msg);
                 return false;
             }
         }
--- a/src/effects/audio_unit/audiouniteffectsmodule.cpp
+++ b/src/effects/audio_unit/audiouniteffectsmodule.cpp
@@ -1,0 +1,68 @@
+/*
+* Audacity: A Digital Audio Editor
+*/
+#include "audiouniteffectsmodule.h"
+
+#include "audioplugins/iaudiopluginsscannerregister.h"
+#include "audioplugins/iaudiopluginmetareaderregister.h"
+
+#include "effects/effects_base/ieffectviewlaunchregister.h"
+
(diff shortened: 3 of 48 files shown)


CMakeLists.txt
```diff
@@ -23,8 +23,8 @@ set(FETCHCONTENT_BASE_DIR ${CMAKE_CURRENT_LIST_DIR}/_deps)
 include(FetchContent)
 FetchContent_Declare(
   muse_framework
-  GIT_REPOSITORY https://github.com/musescore/framework_tmp.git
-  GIT_TAG        u250618_1
+  GIT_REPOSITORY https://github.com/kryksyh/framework_tmp.git
```
embarc-gabriel: Is this correct?
If musescore does not support audiounits this will change completely our workflow once we'll need to be aware of every framework change and port it to our repo.
It would be better to add the support on the musescore repo.

au3/libraries/lib-audio-unit/AudioUnitWrapper.cpp
```diff
@@ -22,6 +22,11 @@
 #include <wx/osx/core/private.h>
 #include <wx/log.h>
 
+AudioUnitWrapper::~AudioUnitWrapper()
+{
+    printf("Wrapper destroyed for component %p, parameters %p, self: %p\n", mComponent, mParameters.get(), this);
```
embarc-gabriel: Should we use printf here or some log function so we can suppress the message according to the context?

au3/libraries/lib-audio-unit/AudioUnitWrapper.cpp
```diff
@@ -38,6 +43,13 @@ AudioUnitWrapper::GetSettings(const EffectSettings& settings)
     return GetSettings(const_cast<EffectSettings&>(settings));
 }
 
+AudioUnitWrapper::AudioUnitWrapper(AudioComponent component, Parameters* pParameters)
+    : mComponent{component}
+    , mParameters{pParameters ? *pParameters : mOwnParameters}
+{
+    printf("Wrapper created for component %p, parameters %p, self: %p\n", mComponent, mParameters.get(), this);
```
embarc-gabriel: Same here.

src/effects/audio_unit/audiouniteffectsmodule.cpp
```diff
@@ -0,0 +1,70 @@
...
+    m_effectsRepository = std::make_shared<AudioUnitEffectsRepository>();
+
+    ioc()->registerExport<IAudioUnitEffectsRepository>(moduleName(), m_effectsRepository);
+}
+
+void au::effects::AudioUnitEffectsModule::resolveImports()
+{
+    auto scannerRegister = ioc()->resolve<muse::audioplugins::IAudioPluginsScannerRegister>(moduleName());
+
+    if (scannerRegister) {
```
embarc-gabriel, line 39: Maybe an IF_ASSERT or early returns?
Nit: Remove the space between auto variable and if or a empty line for all of them.

au3/libraries/lib-audio-unit/AudioUnitWrapper.cpp
kryksyh: That was a temporary debugging, removed

CMakeLists.txt
kryksyh: This is temporary, will be updated when changes are merged to musescore

src/effects/audio_unit/view/AudioUnitControl.mm
```diff
@@ -0,0 +1,234 @@
...
+            // Load the bundle
+            if (auto bundle = [NSBundle bundleWithPath:[bundleLoc path]]) {
+                // Load the class from the bundle
+                if (auto factoryClass = [bundle classNamed:viewClass]) {
+                    // Create an instance of the class
+                    if (id factoryInst = [[[factoryClass alloc] init] autorelease]) {
+                        // Create the view, suggesting a reasonable size
+                        try {
+                            if ((mView = [factoryInst uiViewForAudioUnit:mUnit
+                                          withSize:NSSize{ 800, 600 }])) {
```
embarc-gabriel, line 160: I have a warning here: uiViewForAudioUnit::withSize not found(return type defaults to 'id')

src/effects/audio_unit/view/AudioUnitControl.mm
```diff
@@ -0,0 +1,234 @@
...
+    if (!mView) {
+        return false;
+    }
+
+    asAUView(mAUView)->mView = asNSView(mView);
+
+    // Attach mAUView to the native window
+    WId nativeWindowId = winId();
+    NSView* parentView = reinterpret_cast<NSView*>(nativeWindowId);
+    [parentView addSubview:asNSView(mAUView)];
```
embarc-gabriel, line 122: The Ozone 11 effect crashes here. mView is not null but the content of this object indicates it is not properly initialized.

src/effects/audio_unit/view/audiounitview.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+    const auto instance = std::dynamic_pointer_cast<AudioUnitInstance>(instancesRegister()->instanceById(m_instanceId));
+
+    // TODO: When design for this setting is ready, use the user preference here
+    bool isGraphical = true;
+
+    m_auControl = new AUControl();
+    if (!m_auControl->create(instance->GetComponent(),
+                             instance->GetAudioUnit(), isGraphical)) {
+        delete m_auControl;
```
embarc-gabriel: Is it possible to use a smart pointer here?
The code seems to handle it as expected  but if we have some future change with an early return without calling delete or if the init function is called twice before denit is called we can leak memory.

src/effects/audio_unit/view/audiounitview.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+void AudioUnitView::updateViewGeometry()
+{
+    if (!m_auControl) {
+        return;
+    }
+
+    QSize requiredSize = m_auControl->size();
+    auto availableSize = window()->screen()->availableSize();
```
embarc-gabriel: nit: const

embarc-gabriel (approved):
I have done the best as I can to review it.
I have left some minor comments and apart from the git repository there is nothing that would block merging this code in my personal opinion.

src/effects/audio_unit/view/audiounitview.cpp
kryksyh: Yes we can, we still have to manually delete it in deinit(), but with smartpointer is should be better

src/effects/audio_unit/view/AudioUnitControl.mm
kryksyh: Yeah, I don't really know why do we get it, but the header file and the documentation do not agree https://developer.apple.com/documentation/audiotoolbox/aucocoauibase/1395992-uiviewforaudiounit

PR #9017, opened 2025-07-04
Add Shortcut helper script

chinakov:
Adding a new shortcut helper script to shorten the usage of api.keyboard.key function within test cases.

Recommended:

### DIFF
--- a/share/autobotscripts/steps/Shortcut.js
+++ b/share/autobotscripts/steps/Shortcut.js
@@ -1,0 +1,354 @@
+/*
+Helper functions that consolidate shortcuts. These helper functions are written in Pascal case (eg. NewProject, CutAndLeaveGap)
+ */
+
+var Navigation = require("Navigation.js")
+
+module.exports = {
+    NewProject: function()
+    {
+        //This emulates the New Project command via Ctrl+N
+        api.keyboard.key("Ctrl+N")
+    },
+    Open: function()
+    {
+        //This emulates the Open Project command via Ctrl+O
+        api.keyboard.key("Ctrl+O")
+    },
+    Close: function()
+    {
+        //This emulates the Close Project command via Ctrl+W
+        api.keyboard.key("Ctrl+W")
+    },
+    Exit: function()
+    {
+        //This emulates the Close Project command via Ctrl+Q
+        api.keyboard.key("Ctrl+W")
+    },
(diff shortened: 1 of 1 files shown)


share/autobotscripts/steps/Shortcut.js
```diff
@@ -0,0 +1,354 @@
...
+    },
+    SelectAll: function()
+    {
+        //This emulates the select all command via Ctrl+A 
+        api.keyboard.key("Ctrl+A")
+    },
+    DeselectAll: function()
+    {
+        //This emulates the Deselect All command via Ctrl+Shift+A
+        api.keyboard.key("")
```
embarc-gabriel: Is something missing here?

share/autobotscripts/steps/Shortcut.js
```diff
@@ -0,0 +1,354 @@
+/*
```
embarc-gabriel, line 1: We usually have this on top of all files
```
/*
*/
```

share/autobotscripts/steps/Shortcut.js
```diff
@@ -0,0 +1,354 @@
...
+Helper functions that consolidate shortcuts. These helper functions are written in Pascal case (eg. NewProject, CutAndLeaveGap)
+ */
+
+var Navigation = require("Navigation.js")
+
+module.exports = {
+    NewProject: function()
+    {
+        //This emulates the New Project command via Ctrl+N
```
embarc-gabriel, line 10: nit: I would prefer to remove most of these comments. I mean, we could add the context just once as you have done on the top of the file and the function name should be clear enough to give the information that is missing.

share/autobotscripts/steps/Shortcut.js
chinakov: Yes, fixed.

share/autobotscripts/steps/Shortcut.js
chinakov: Added this in most recent commit

share/autobotscripts/steps/Shortcut.js
chinakov: Alright, let me fix this. Thanks!

embarc-gabriel (approved):

PR #9066, opened 2025-07-15
8994 Meter multiple db ranges

### LINKED ISSUE #8994: Support multiple db ranges
### Description

Link to figma: https://www.figma.com/design/vpweiC0Dfy4wFgrNrvfTTD/03---Playback-meter?node-id=822-15210&t=18EjVAC0NNz3dZPz-0

embarc-gabriel:
Resolves: #8994

Recommended:

### DIFF
--- a/src/appshell/qml/Preferences/internal/MeterDbRangeSection.qml
+++ b/src/appshell/qml/Preferences/internal/MeterDbRangeSection.qml
@@ -1,0 +1,65 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * Audacity-CLA-applies
+ *
+ * Audacity
+ * A Digital Audio Editor
+ *
+ * Copyright (C) 2025 Audacity BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+import QtQuick 2.15
+
+import Muse.UiComponents 1.0
+import Audacity.Playback 1.0
+import Audacity.Preferences
+
+BaseSection {
+    id: root
+
+    title: qsTrc("appshell/preferences", "Meter dB range")
+    spacing: 16
+
+    property var model: null
+
+    PlaybackStateModel {
+        id: playbackState
+    }
+
+    ComboBoxWithTitle {
+        title: qsTrc("appshell/preferences", "dB range")
+
+        columnWidth: 320
+
+        currentIndex: root.model.meterDbRange
+        model: [
+            "-36 dB (shallow range for high-amplitude editing)",
+            "-48 dB (PCM range of 8 bit samples)", 
+            "-60 dB (PCM range of 10 bit samples)",
+            "-72 dB (PCM range of 12 bit samples)",
+            "-84 dB (PCM range of 14 bit samples)",
+            "-96 dB (PCM range of 16 bit samples)",
+            "-120 dB (approximate limit of human hearing)",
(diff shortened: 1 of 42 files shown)


src/appshell/qml/Preferences/internal/MeterDbRangeSection.qml
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    ComboBoxWithTitle {
+        title: qsTrc("appshell/preferences", "dB range")
+
+        columnWidth: 320
+
+        currentIndex: root.model.meterDbRange
+        model: [
+            "-36 dB (shallow range for high-amplitude editing)",
```
kryksyh: These lines should be translated, I also feel like this should be in the cpp model, next to the num rather than in qml

src/playback/internal/meters/dblinearmeter.cpp
```diff
@@ -4,41 +4,82 @@
...
-    } else if (dbValue > -50.0) {
-        return (15 + 0.75 * (dbValue + 40.0)) / 100.0;
-    } else if (dbValue > -60.0) {
-        return (7.5 + 0.5 * (dbValue + 50.0)) / 100.0;
-    } else {
-        return 0.0;
+        return 1.0;
     }
+
+    if (dbValue > dbRange / 3.0) {
```
kryksyh: What is going on here? 😆 
Maybe something like this would be more readable: https://godbolt.org/z/77zeejPdz

src/playback/internal/meters/dblogmeter.cpp
```diff
@@ -11,42 +11,49 @@
...
-        { LOW_RESOLUTION_METER_THRESHOLD, 6, 2 },
-        { 0, 12, 3 }
-    };
-
-    for (const auto& step : steps) {
-        if (meterSize >= step.threshold) {
-            return step;
+    const std::array<std::array<int, 2>, 7> fullStepsInc = { { { 1, 0 }, { 2, 1 }, { 3, 1 }, { 6, 2 }, { 12, 3 }, { 24, 6 }, { 48, 6 } } };
+
+    for (const auto& step : fullStepsInc) {
```
kryksyh: minor thing, you can use struct binding here:

`for (const auto& [fullStep, smallStep] : fullStepsInc) {`

src/playback/internal/meters/linearmeter.cpp
```diff
@@ -52,7 +65,7 @@ double LinearMeter::stepToPosition(double step) const
 
 double LinearMeter::sampleToPosition(double sample) const
 {
-    if ((sample < MIN_VOLUME_DB) || muse::is_equal(sample, MIN_VOLUME_DB)) {
+    if ((sample < m_dbRange) || muse::is_equal(sample, m_dbRange)) {
```
kryksyh, line 68: why not simply `sample <= m_dbRange`?

kryksyh (approved):

src/appshell/qml/Preferences/internal/MeterDbRangeSection.qml
embarc-gabriel: Fixed.

src/playback/internal/meters/dblinearmeter.cpp
embarc-gabriel: Yeah ... I thought keeping the code like this would more or less explain the reason behind it.
But this doesn't seem to be the case.
Fixed.

src/playback/internal/meters/dblogmeter.cpp
embarc-gabriel: Fixed. Much better. Thank you.

src/playback/internal/meters/linearmeter.cpp
embarc-gabriel: I think we should not do == for double values, isn't?
That's the whole point of this muse::is_equal ...
Anyway ... I'll close it but if you think this can be removed we can talk.

kryksyh: >I think we should not do == for double values, isn't?

absolutely!
At the same time, `is_equal` ~= `(sample + eps) > m_dbRange {1} && (sample - eps) < m_dbRange) {2}`

{1} is already covered by the first part of your comparison `sample < m_dbRange`, and {2} will be nicely covered in stepToPosition

I do not insist, but less code == easy to read, easy to understand :)

embarc-gabriel: Yes! Always.
I'll modify it.

Issue #9038, opened 2025-07-09
Implement mp3 export


grliszas14:
### Description

.

chinakov:
Tested all platforms, as per @grliszas14 I was asked to create a separate issue https://github.com/audacity/audacity/pull/9067 that's connected to this PR.

PR #9067, opened 2025-07-15
Mp3 exporter

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9038

### DIFF
--- a/au3/modules/import-export/RegisterExportPlugins.cpp
+++ b/au3/modules/import-export/RegisterExportPlugins.cpp
@@ -1,0 +1,24 @@
+/*
+* Audacity: A Digital Audio Editor
+*/
+
+#include "modules/import-export/mod-mp3/ExportMP3.h"
+#include "ExportFFmpeg.h"
+#include "libraries/lib-import-export/ExportPluginRegistry.h"
+
+#include "RegisterExportPlugins.h"
+
+void RegisterExportPlugins()
+{
+    static ExportPluginRegistry::RegisteredPlugin sMP3Plugin{
+        "MP3",
+        [] { return std::make_unique<ExportMP3>(); }
+    };
+
+#ifdef AU_USE_FFMPEG
+    static ExportPluginRegistry::RegisteredPlugin sFFmpegPlugin{
+        "FFmpeg",
+        [] { return std::make_unique<ExportFFmpeg>(); }
+    };
+#endif
+}
--- a/au3/modules/import-export/mod-mp3/ExportMP3.cpp
+++ b/au3/modules/import-export/mod-mp3/ExportMP3.cpp
@@ -69,23 +69,16 @@
 #include <wx/log.h>
 #include <wx/mimetype.h>
 
-#include <wx/textctrl.h>
-#include <wx/choice.h>
-
-#include <rapidjson/document.h>
-
 #include "FileNames.h"
 #include "float_cast.h"
-#include "HelpSystem.h"
 #include "Mix.h"
 #include "Prefs.h"
-#include "Tags.h"
+#include "libraries/lib-tags/Tags.h"
 #include "Track.h"
 #include "wxFileNameWrapper.h"
-#include "wxPanelWrapper.h"
 #include "Project.h"
 
-#include "Export.h"
+#include "libraries/lib-import-export/Export.h"
 #include "BasicUI.h"
 
 #include <lame/lame.h>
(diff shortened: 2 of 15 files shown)


buildscripts/cmake/SetupDependencies.cmake
```diff
@@ -25,7 +25,9 @@ else()
     set(LIB_BUILD_TYPE "debug")
 endif()
 
-set(REMOTE_ROOT_URL https://raw.githubusercontent.com/musescore/muse_deps/main)
+# temporary, WIP
+# set(REMOTE_ROOT_URL https://raw.githubusercontent.com/musescore/muse_deps/main)
```
grliszas14: Need to switch back to it when https://github.com/musescore/muse_deps/pull/2 gets merged

src/au3wrap/au3wrapmodule.cpp
```diff
@@ -66,6 +67,8 @@ void Au3WrapModule::onInit(const muse::IApplication::RunMode&)
 #ifdef AU_USE_FFMPEG
     FFmpegStartup();
 #endif
+    // call once so static variable sRegisteredPlugin gets created
+    ExportMP3();
```
kryksyh: I understand that this is not a temporary solution, so we need to fix the linking issue explicitly. 
The comment here is misleading, we do not need to create a temporary ExportMP3 object so sRegisteredPlugin created, we do not need to call any function for that matter.

The only thing we need to do is pull any symbol from `ExportMP3.cpp` to prevent the linker from dropping ExportMP3.o and all its internal symbols.

You should make this more obvious to lower WTFPM. My suggestion:
- move ExportMP3 and ExportFFMPEG declarations to ExportMP3.h, ExportFFMPEG.h 
- define all RegisteredPlugins in one translation that is linked to the executable

Alternatively, if moving ExportXXX is not desirable:
- define `void ExportXXXForceLinking() {}` in its respective Cpp file, ugly name is okay for an ugly situation
- call those functions somewhere in the main app code.

au3/modules/import-export/mod-mp3/ExportMP3.cpp
```diff
@@ -1367,18 +1198,20 @@ bool MP3Exporter::PutInfoTag(wxFFile& f, wxFileOffset off)
...
-    } else {
-        wxLogMessage(wxT("LAME registry key does not exist."));
+    HKEY hKey;
+    const char* subkey = "Software\\Lame for Audacity";
+    const char* valueName = "InstallPath";
+    char pathBuffer[MAX_PATH];
+    DWORD pathBufferSize = sizeof(pathBuffer);
+    std::string path;
+
+    if (RegOpenKeyExA(HKEY_LOCAL_MACHINE, subkey, 0, KEY_READ, &hKey) == ERROR_SUCCESS) {
```
kryksyh: Not sure about that, would xxxA function work well with non-ASCII paths?

src/au3wrap/CMakeLists.txt
```diff
@@ -727,6 +727,12 @@ if(AU_USE_FFMPEG)
     )
 endif()
 
+set(MODULE_SRC ${MODULE_SRC}
+    ${AU3_MODULES}/import-export/mod-mp3/ExportMP3.cpp
```
Eism: I think you can add it in `set(AU3_SRC` block

au3/modules/import-export/mod-mp3/ExportMP3.cpp
LWinterberg: This is legacy code from back when mp3 was patented. It can be safely removed I think; we just need to use the binary

src/au3wrap/au3wrapmodule.cpp
Eism: Btw, if this line is needed only for export and is not needed by another part of the program, then we can put it in the export module - if we disable the export module, this method will not be called. This is true for all modules

au3/modules/import-export/RegisterExportPlugins.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "registerexportplugins.h"
+
+void RegisterExportPlugins()
+{
+    static ExportPluginRegistry::RegisteredPlugin sMP3Plugin{
+        "MP3",
+        [] { return std::make_unique<ExportMP3>(); }
+    };
+
+    static ExportPluginRegistry::RegisteredPlugin sFFmpegPlugin{
```
kryksyh, line 19: #ifdef AU_USE_FFMPEG ?

au3/modules/import-export/mod-mp3/ExportMP3.cpp
grliszas14: I'm gonna leave it just in case, for the peace of mind. Changed to xxxW functions and wchar_t so it can handle unicode

Issue #8030, opened 2025-01-13
Split tool for splitting clips


LWinterberg:
### Description

Figma
https://www.figma.com/design/GhYn9zeEigAc9g9Zy0DeHB/31---Split-Tool?node-id=3-15&p=f&t=lackELTMTw2tnTgw-0

Hotkey: S

PR #9079, opened 2025-07-16
Split tool

kryksyh:
Resolves: #8030

Implement "Split tool"

Recommended:

### DIFF
--- a/src/projectscene/internal/projectviewstate.cpp
+++ b/src/projectscene/internal/projectviewstate.cpp
@@ -79,6 +79,27 @@ ProjectViewState::ProjectViewState(std::shared_ptr<au::au3::IAu3Project> project
         zoomState.tracksVerticalOffset = y;
         saveProjectZoomState(au3Project, zoomState);
     });
+
+    globalContext()->currentTrackeditProjectChanged().onNotify(this, [this](){
+        auto prj = globalContext()->currentTrackeditProject();
+        if (!prj) {
+            return;
+        }
+
+        prj->trackRemoved().onReceive(this, [this](const trackedit::Track& track) {
+            auto it = m_tracks.find(track.id);
+            if (it == m_tracks.end()) {
+                return;
+            }
+            m_totalTracksHeight.set(m_totalTracksHeight.val - it->second.height.val);
+            m_tracks.erase(it);
+        });
+    });
+}
+
+muse::ValCh<int> ProjectViewState::totalTrackHeight() const
+{
+    return m_totalTracksHeight;
 }
 
 muse::ValCh<int> ProjectViewState::tracksVerticalOffset() const
@@ -141,6 +162,8 @@ ProjectViewState::TrackData& ProjectViewState::makeTrackData(const trackedit::Tr
     TrackData d;
     d.height.val = DEFAULT_HEIGHT;
     d.collapsed.val = false;
+    m_totalTracksHeight.set(m_totalTracksHeight.val + d.height.val);
+
     return m_tracks.insert({ trackId, d }).first->second;
 }
 
@@ -166,7 +189,7 @@ muse::ValCh<bool> ProjectViewState::isTrackCollapsed(const trackedit::TrackId& t
     return d.collapsed;
 }
 
-void ProjectViewState::changeTrackHeight(const trackedit::TrackId& trackId, int deltaY)
+void ProjectViewState::changeTrackHeight(const trackedit::TrackId& trackId, int delta)
 {
     TrackData* d = nullptr;
     auto it = m_tracks.find(trackId);
@@ -176,9 +199,13 @@ void ProjectViewState::changeTrackHeight(const trackedit::TrackId& trackId, int 
         d = &makeTrackData(trackId);
     }
 
-    int newVal = std::max(d->height.val + deltaY, MIN_HEIGHT);
-    d->height.set(newVal);
-    d->collapsed.set(newVal < COLLAPSE_HEIGHT);
+    int oldHeight = d->height.val;
+    int newHeight = std::max(oldHeight + delta, MIN_HEIGHT);
+
+    d->height.set(newHeight);
+    d->collapsed.set(newHeight < COLLAPSE_HEIGHT);
+
+    m_totalTracksHeight.set(m_totalTracksHeight.val + (newHeight - oldHeight));
 }
 
 void ProjectViewState::setTrackHeight(const trackedit::TrackId& trackId, int height)
@@ -191,15 +218,19 @@ void ProjectViewState::setTrackHeight(const trackedit::TrackId& trackId, int hei
         d = &makeTrackData(trackId);
     }
 
-    d->height.set(std::max(height, MIN_HEIGHT));
+    int oldHeight = d->height.val;
+    int newHeight = std::max(height, MIN_HEIGHT);
+    d->height.set(newHeight);
     d->collapsed.set(height < COLLAPSE_HEIGHT);
+
+    m_totalTracksHeight.set(m_totalTracksHeight.val + (newHeight - oldHeight));
 }
 
 au::trackedit::TrackId ProjectViewState::trackAtPosition(double y) const
 {
     trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
     if (!prj) {
-        return -1;
+        return trackedit::INVALID_TRACK;
     }
 
     trackedit::TrackIdList tracks = prj->trackIdList();
@@ -212,12 +243,12 @@ au::trackedit::TrackId ProjectViewState::trackAtPosition(double y) const
         trackTop = trackBottom;
         trackBottom = trackTop + trackHeight(id).val;
 
-        if (trackTop <= y && trackBottom >= y) {
+        if (muse::RealIsEqualOrMore(y, trackTop) && muse::RealIsEqualOrLess(y, trackBottom)) {
             return id;
         }
     }
 
-    return -1;
+    return trackedit::INVALID_TRACK;
 }
 
 au::trackedit::TrackIdList ProjectViewState::tracksInRange(double y1, double y2) const
--- a/src/projectscene/internal/projectviewstate.h
+++ b/src/projectscene/internal/projectviewstate.h
@@ -27,8 +27,15 @@ public:
     ProjectViewState(std::shared_ptr<au::au3::IAu3Project> project);
 
     // State of elements
+    muse::ValCh<int> totalTrackHeight() const override;
     muse::ValCh<int> trackHeight(const trackedit::TrackId& trackId) const override;
     muse::ValCh<bool> isTrackCollapsed(const trackedit::TrackId& trackId) const override;
+
+    int trackVerticalPosition(const trackedit::TrackId& trackId) const override;
+    void changeTrackHeight(const trackedit::TrackId& trackId, int delta) override;
+    void setTrackHeight(const trackedit::TrackId& trackId, int height) override;
+    trackedit::TrackId trackAtPosition(double y) const override;
+    trackedit::TrackIdList tracksInRange(double y1, double y2) const override;
 
     bool isSnapEnabled() const override;
     void setIsSnapEnabled(bool enabled) override;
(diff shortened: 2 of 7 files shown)


src/projectscene/internal/projectviewstate.cpp
```diff
@@ -195,6 +195,31 @@ void ProjectViewState::setTrackHeight(const trackedit::TrackId& trackId, int hei
...
+    for (trackedit::TrackId id : tracks) {
+        trackTop = trackBottom;
+        trackBottom = trackTop + trackHeight(id).val;
+
+        if (trackTop <= y && trackBottom >= y) {
+            return id;
+        }
+    }
+
+    return {};
```
embarc-gabriel: Shouldn´t it be -1 here?

kryksyh: it should, thanks!

src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1374,12 +1374,26 @@ bool Au3Interaction::splitTracksAt(const TrackIdList& tracksIds, std::vector<sec
             return false;
         }
 
+        bool didAnySplitOccur = false;
+
         for (const auto& pivot : pivots) {
-            waveTrack->SplitAt(pivot);
+            if (waveTrack->GetIntervalAtTime(pivot)) {
+                waveTrack->SplitAt(pivot);
```
grliszas14, line 1381: After a successful split we should call
`viewState()->updateClipsBoundaries(false)` 
so points to which yellow guideline will snap gets updated
(to be honest we should update it in every possible split function, we have a few already)

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -245,8 +258,10 @@ Rectangle {
 
     CustomCursor {
         id: customCursor
-        active: (content.isIsolationMode || content.isBrush || content.isNearSample || content.leftTrimContainsMouse || content.rightTrimContainsMouse
-            || content.leftTrimPressedButtons || content.rightTrimPressedButtons)
+        active: (content.isIsolationMode || content.isNearSample
+                || content.leftTrimContainsMouse  || content.rightTrimContainsMouse
+                || content.leftTrimPressedButtons || content.rightTrimPressedButtons
+                || (root.isSplitMode && root.clipHovered) || (content.isBrush && root.clipHovered) )
```
embarc-gabriel, line 264: Here was the smooth tool behavior changed to only be enabled when hovering a clip?
If so, this is not the design spec. It should be enabled while the user is pressing ALT and also it should work across tracks.
https://www.figma.com/design/01XLIICmy2AG6t1ieyTp66/02---Clips?node-id=1148-57511&t=PI1mvk8FuBoVFmKj-0

kryksyh: I don't see the design specifying the cursor shape when we move the cursor outside of the clip, so I thought that it would make sense if we disable the brush outside of the clip, to let the user know that there is nothing to brush at the moment

kryksyh: BUT, there are two other issues, both are present in master

1. When ALT is pressed, clipHovered is set to false. This breaks the intended behavior in this PR
2. The brush tool in AU4 works differently compared to AU3. In current master, the brush tool cursor is always enabled while holding ALT, but the smoothing only executed if the cursor is hovering near the sample, which is confusing. In AU3 the user could click anywhere on the clip area and it will smooth vertically nearest samples

kryksyh: To quote the design:

> Unlike the pen tool, which will revert back to the text cursor when it leaves the samples’ hit-box, the smooth tool will remain active as long as the User continues to hold the Alt key.

I read it as the smooth tool should smooth even if the user moved the cursor slightly way from the sample stem

kryksyh: @embarc-gabriel if we agree on it, I'll fix it in this PR

embarc-gabriel: Yeah. That's ok for me. I had a slightly different understanding at that time but I think your suggestion fits better here.

src/projectscene/internal/projectviewstate.cpp
```diff
@@ -195,6 +195,31 @@ void ProjectViewState::setTrackHeight(const trackedit::TrackId& trackId, int hei
     d->collapsed.set(height < COLLAPSE_HEIGHT);
 }
 
+au::trackedit::TrackId ProjectViewState::trackAtPosition(double y) const
```
grliszas14, line 229: We already have that function: `selectionController()->determinePointedTrack(double y)`, you can re-use it

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -16,13 +16,20 @@ Rectangle {
...
+    property var hoveredTrackId: null
+    property double hoveredTrackVerticalPosition
+    property double hoveredTrackHeight
     property bool tracksHovered: false
     property bool guidelineActive: false
     property alias altPressed: tracksViewState.altPressed
     property alias ctrlPressed: tracksViewState.ctrlPressed
+    property alias isSplitMode: tracksModel.isSplitMode
+    property alias escPressed: tracksViewState.escPressed
+    property double splitGuidelinePosition
```
grliszas14: maybe we should initialize it to something like -1.0 so in case of bad luck we don't show it at 0.0 but rather off the view?

src/projectscene/view/common/trackviewstatemodel.h
```diff
@@ -0,0 +1,71 @@
...
+#include <QObject>
+
+#include "global/async/asyncable.h"
+
+#include "modularity/ioc.h"
+#include "context/iglobalcontext.h"
+#include "playback/iplaybackcontroller.h"
+#include "playback/iplaybackconfiguration.h"
+#include "playback/view/common/playbackmetermodel.h"
+#include "record/irecordcontroller.h"
```
grliszas14: I think order of includes is not right ;) 
```suggestion
#include "modularity/ioc.h"
#include "context/iglobalcontext.h"
#include "playback/iplaybackcontroller.h"
#include "playback/iplaybackconfiguration.h"
#include "record/irecordcontroller.h"

#include "playback/view/common/playbackmetermodel.h"
```

src/projectscene/view/common/tracksviewstatemodel.cpp
```diff
@@ -178,14 +133,22 @@ int TracksViewStateModel::tracksVerticalScrollPadding() const
     return m_tracksVerticalScrollPadding;
 }
 
-int TracksViewStateModel::trackHeight() const
+int TracksViewStateModel::trackHeight(trackedit::TrackId trackId)
```
grliszas14:
```suggestion
int TracksViewStateModel::trackHeight(trackedit::TrackId trackId) const
```

src/projectscene/view/common/tracksviewstatemodel.h
```diff
@@ -9,109 +9,71 @@
...
 
     Q_INVOKABLE void requestVerticalScrollLock();
     Q_INVOKABLE void requestVerticalScrollUnlock();
 
-    Q_INVOKABLE void changeTrackHeight(int deltaY);
+    Q_INVOKABLE bool snapEnabled() const;
+    Q_INVOKABLE trackedit::TrackId trackAtPosition(double x, double y) const;
 
-    Q_INVOKABLE bool snapEnabled();
+    Q_INVOKABLE int trackHeight(trackedit::TrackId trackId);
```
grliszas14:
```suggestion
    Q_INVOKABLE int trackHeight(trackedit::TrackId trackId) const;
```

src/projectscene/view/common/tracksviewstatemodel.h
```diff
@@ -9,109 +9,71 @@
...
+    Q_INVOKABLE int trackVerticalPosition(trackedit::TrackId trackId);
```
grliszas14:
```suggestion
    Q_INVOKABLE int trackVerticalPosition(trackedit::TrackId trackId) const;
```

src/projectscene/view/common/tracksviewstatemodel.cpp
```diff
@@ -178,14 +133,22 @@ int TracksViewStateModel::tracksVerticalScrollPadding() const
...
-    return m_trackHeight.val;
+    IProjectViewStatePtr vs = viewState();
+    if (vs) {
+        return vs->trackHeight(trackId).val;
+    }
+    return {};
 }
 
-bool TracksViewStateModel::isTrackCollapsed() const
+int TracksViewStateModel::trackVerticalPosition(trackedit::TrackId trackId)
```
grliszas14:
```suggestion
int TracksViewStateModel::trackVerticalPosition(trackedit::TrackId trackId) const
```

src/projectscene/internal/projectviewstate.cpp
kryksyh: It is kind of strange place to have it in selectionController, i'll do it backwards if I may, and replace determinePointedTrack with trackAtPosition :D

src/projectscene/internal/projectviewstate.cpp
```diff
@@ -195,6 +195,80 @@ void ProjectViewState::setTrackHeight(const trackedit::TrackId& trackId, int hei
...
+{
+    trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+    if (!prj) {
+        return -1;
+    }
+
```
grliszas14, line 235: Don't remember at that moment but I think there was a bug without this:

```suggestion
    if (y < 1) {
        y = 1;
    }
```

src/projectscene/internal/projectviewstate.cpp
```diff
@@ -195,6 +195,80 @@ void ProjectViewState::setTrackHeight(const trackedit::TrackId& trackId, int hei
...
+    int tracksVerticalOffset = this->tracksVerticalOffset().val;
+    int trackTop = -tracksVerticalOffset;
+    int trackBottom = trackTop;
+
+    for (trackedit::TrackId id : tracks) {
+        trackTop = trackBottom;
+        trackBottom = trackTop + trackHeight(id).val;
+
+        if (trackTop <= y && trackBottom >= y) {
```
grliszas14: better to use muse comparison operators, just in case 
```suggestion
        if (muse::RealIsEqualOrMore(y, trackTop) && muse::RealIsEqualOrLess(y, trackBottom)) {
```

src/trackedit/internal/au3/au3selectioncontroller.cpp
```diff
@@ -124,99 +124,6 @@ void Au3SelectionController::addSelectedTrack(const TrackId& trackId)
...
-    for (trackedit::TrackId trackId : tracks) {
-        trackTop = trackBottom;
-        trackBottom = trackTop + vs->trackHeight(trackId).val;
-
-        if (muse::RealIsEqualOrMore(y, trackTop) && muse::RealIsEqualOrLess(y, trackBottom)) {
-            return trackId;
-        }
-    }
-
-    if (muse::RealIsEqualOrLess(y, trackBottom)) {
```
grliszas14, line 213: I'm trying to remind myself why this was needed... Looks reduntant at first glance but maybe there was an edge case? I think it requires some bonus testing of clips vertical movement and other features that will use `trackAtPosition`

kryksyh: I can't imagine a situation where this condition would be true. It could only be true, if for the last track `y >= trackTop` is false , and `y <= trackBottom)` is true. This contradicts condition where the first `trackTop  <= 0` (tracksVerticalOffset is always positive)

What bothers me here is that this function could return 3 kind of values:
- empty optional, **meaning something weird happen**
- optional containing INVALID_TRACK, **current position is below the last track**
- optional containing a valid track id

And all these return values are handled differently. Thanks for pointing out I'll do a bit of cleanup there too

src/projectscene/internal/projectviewstate.cpp
kryksyh: I'm not sure about that. Could you try to remember the situation where it was useful? 

From the ViewState API standpoint, negative vertical position is a valid value, and clamping it to a positive value may be undesirable.

grliszas14 (approved):

Issue #9028, opened 2025-07-08
Audio.com sign-in optimisation (Core 3.7.5 task)


Tantacrul:
### Description

Figma here:
https://www.figma.com/design/dwNSb1W7YUaxgyoQIHDdSa/01_WX---Project-template--Duplicate-me---Copy---Copy-?node-id=30-761

This is a simple optimisation that uses Audacity's existing in-app sign-in flow.

UTM link for continue button: https://audio.com/tour?mtm_campaign=audacitydesktop&mtm_content=app_launch_reg

Text changes on registration screen: 

<img width="547" height="452" alt="Image" src="https://github.com/user-attachments/assets/c5776d9c-da6f-4953-b239-8ab1948ed258" />

chinakov:
Couple of notes:
- If the User has logged in, went through the flow and landed on the UTM page, then logged out and repeated the entire process, the User won't be shown the log-in page, but only the UTM page. After completion of the UTM survey, the previous user account is signed in. This happens regardless if the User is signed into Audio.com or Audacity. If the Audaicty config is reset, the flow works correctly again. 
- If Dark Mode is enabled in Ubuntu but not Audacity, the new text is hard to see. Not really a huge issue, but should be noted regardless. If Dark theme in Audacity is enabled, it looks much better. 

<img width="489" height="446" alt="Image" src="https://github.com/user-attachments/assets/27d0e621-2cd7-4d3d-84ad-f36b23d61d90" />

<img width="524" height="462" alt="Image" src="https://github.com/user-attachments/assets/debcd684-2edd-4fa9-a56d-55bfe47be0f6" />

Are these fine @Tantacrul ?

chinakov:
Edge cases are tested, and the logging out issue outlined above is not reproducible as I've mistakenly didn't log out correctly in Audacity 3. Thanks !

DavidBailes:
There's a focus bug.
On the cloud setup slide of the  welcome dialog, if you press the Continue button (either using the mouse or keyboard), then close the Create account dialog, the main window becomes the focus, rather than the focus returning to the welcome dialog.

chinakov:
@DavidBailes - is this only on one platform? I can't reproduce this on Windows.

DavidBailes:
> [@DavidBailes](https://github.com/DavidBailes) - is this only on one platform? I can't reproduce this on Windows.

I've only tested on Windows. Here's a screen shot immediately after the Create account dialog is closed.
[...]

PR #9080, opened 2025-07-17
Audio.com sign-in optimization

kryksyh:
Resolves: #9028, #9097

Open Login Dialog from the Welcome Screen

Recommended:

### DIFF
--- a/src/WhatsNewDialog.cpp
+++ b/src/WhatsNewDialog.cpp
@@ -24,6 +24,8 @@
 
 #include "HelpSystem.h"
 #include "HelpText.h"
+#include "ModuleManager.h"
+#include "PluginProvider.h"
 #include "ProjectWindows.h"
 #include "ShuttleGui.h"
 #include "Theme.h"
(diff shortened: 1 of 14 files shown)


src/WhatsNewDialog.cpp
```diff
@@ -360,44 +363,56 @@ void WhatsNewDialog::Populate(ShuttleGui& S)
    const int width = FromDIP(572);
    const int height = FromDIP(322);
 #endif
-   std::vector<CarouselSnapshot> snapshots {
-      {  XXO("Complete your Audacity cloud setup with audio.com"),
+
+   std::vector<CarouselSnapshot> snapshots;
+   snapshots.reserve(4);
```
grliszas14: Most of the time we'll add 5 elements I presume so there will be vector reallocation. It doesn't matter much with this number of elements anyway so just nitpicking :)

grliszas14 (approved):
Veeeery nicee 👍

src/WhatsNewDialog.cpp
kryksyh: Right, makes sense, thanks!

PR #9092, opened 2025-07-17
Add Test Cases and helper functions

chinakov:
This adds 2 new helper scripts - one to create tracks via 'Alt' shortcuts, and added new stuff into navigation for easier test verification.

Recommended:

### DIFF
--- a/share/autobotscripts/TC1.5_Effects_Shortcut.js
+++ b/share/autobotscripts/TC1.5_Effects_Shortcut.js
@@ -1,0 +1,42 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+
+var Home = require("steps/Home.js")
+var Shortcut = require("steps/Shortcut.js")
+
+var testCase = {
+    name: "TC1.5: Real Time Effects Shortcut",
+    description: "Tests the Real Time Effects Shortcut key (E)",
+    steps: [
+        {
+            name: "Close project (if opened) and go to home to start", func: function () {
+                api.dispatcher.dispatch("file-close")
+                Home.goToHome()
+            }
+        },
+        {
+            name: "Create new project", func: function () {
+                Home.createNewProject()
+            }
+        },
+        {
+            name: "Open Real Time Effects Panel", func: function () {
+                Shortcut.openEffects()
+            }
+        },
+        {
+            name: "Check if the Effects are enabled", func: function () {
+                var controls = api.navigation.controls("MasterEffectsSection", "Master effects bypass")
+                if (controls[0].enabled === false) {
+                    api.autobot.error("Effects Panel is not enabled.")
+                }
+            }
+        },
+    ]
+}
+
+function main() {
+    api.autobot.setInterval(1000)
+    api.autobot.runTestCase(testCase)
+}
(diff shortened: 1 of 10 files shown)


share/autobotscripts/steps/Navigation.js
```diff
@@ -65,6 +65,28 @@ module.exports = {
             api.autobot.error("navigation error: not found control: " + controlNameOrIndex)
         }
     },
+    checkControlsCount: function(sectionName, panelName, count)
```
kryksyh: I suggest this function should return true or false, and then the caller should decide whether this is an error.

There could be situations where the number of controls varies, and the caller will branch depending on it.

share/autobotscripts/steps/Navigation.js
```diff
@@ -65,6 +65,28 @@ module.exports = {
...
+    {
+        var controls = api.navigation.controls(sectionName, panelName)
+        if (controls.length !== count) {
+            api.autobot.error("Control count is not " + count + ", section: " + sectionName + ", panel: " + panelName)
+        }
```
kryksyh, line 61:
```suggestion
        var controls = api.navigation.controls(sectionName, panelName)
        return control.length == count
```

share/autobotscripts/steps/Navigation.js
kryksyh: Or, you can keep the function body as it is right now, but rename it to AssertControlCount, RequireControlCount or EnsureControlCount to make it clear that the function will stop the test execution when test has failed.

The same goes for the rest

share/autobotscripts/TC1.5_Effects_Shortcut.js
```diff
@@ -0,0 +1,38 @@
...
+        }},
+        {name: "Create new project", func: function() {
+            Home.createNewProject()
+        }},
+        {name: "Open Real Time Effects Panel", func: function() {
+            Shortcut.openEffects()
+        }},
+        {name: "Check if the Effects are enabled", func: function() {
+            var controls = api.navigation.controls("MasterEffectsSection", "Master effects bypass")
+             if (controls[0].enabled === false) {
```
kryksyh: Extra space in indentation

kryksyh: It is better to extract this to `GetMasterEffectsPanel` helper function to avoid magic `controls[0]`,

share/autobotscripts/TC1.5_Effects_Shortcut.js
```diff
@@ -0,0 +1,38 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+
+var Navigation = require("steps/Navigation.js")
+var Home = require("steps/Home.js")
+var ClipNavigation = require("steps/ClipNavigation.js")
+var Shortcut = require("steps/Shortcut.js");
+const { checkControlsCount } = require("./steps/Navigation");
```
kryksyh: Navigation is already imported, you can use `Navigation.checkControlsCount`

share/autobotscripts/steps/Navigation.js
chinakov: This function was created specifically for checking the count of a particular control group, but I will rename it to ensureControlCount. Thanks!

kryksyh (approved):

PR #9099, opened 2025-07-20
Make main application code more legible

kaixoo12:
Resolves: *No open issue*

Makes Application class more legible by splitting `App::exec()` into different helper methods

Recommended:

### DIFF
--- a/src/app/app.cpp
+++ b/src/app/app.cpp
@@ -63,11 +63,7 @@ void App::addModule(modularity::IModuleSetup* module)
     m_modules.push_back(module);
 }
 
-int App::run(QCoreApplication& app, CommandLineParser& commandLineParser)
-{
-    // ====================================================
-    // Setup global Qt application variables
-    // ====================================================
+void setupGlobalEnvironmentVariables() {
     app_init_qrc();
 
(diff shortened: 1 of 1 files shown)


igorkorsukov:
Now it has become more difficult to understand and see what is actually happening and in what order. Now the methods are not correct. For example, the loadWindowContents method, but inside it is done not only load content, but also `m->onDelayedInit()`... or `startupScenario()->runOnSplashScreen()`... How to see this...
Or there is a `setupModules` method, it may seem that the whole setup is done there, but there is only a part... and the other part is done in other places. How to find out about the code? i.e. we need to investigate it all, but now it is more difficult to do.

kaixoo12:
You're 100% right, and I take these suggestions into account. However, now the code is self-describing.

Please note, you say this as a developer that has previous experience with the entire codebase. A new contributor like me would see it with other eyes. And making the source code easier to read is especially important in open source software, where you're 100% reliant on contributors.

However, you're absolutely right that a lot of these changes aren't right. I will take another look at it. It would help if you can send a specific review with the changes you'd make. Thanks!

**TL;DR** There are still some improvements to make in the PR. However, the new contributors now gets an overall view of what the `exec` function does.

PR #9115, opened 2025-07-22
9102 normalize

### LINKED ISSUE #9102: Migrate Normalise to AU4

saintmatthieu:
Resolves: #9102

Recommended:

### DIFF
--- a/src/effects/builtin/normalize/NormalizeView.qml
+++ b/src/effects/builtin/normalize/NormalizeView.qml
@@ -1,0 +1,96 @@
+import QtQuick
+import QtQuick.Layouts
+import Muse.UiComponents
+import Audacity.Effects
+
+import "../common"
+
+EffectBase {
+    id: root
+
+    property string title: qsTrc("effects/normalize", "Normalize")
+    property bool isApplyAllowed: removeDcCheckbox.checked || normalizePeakAmplitudeCheckbox.checked
+
+    width: 400
+    implicitHeight: column.height
+
+    model: normalize
+
+    NormalizeViewModel {
+        id: normalize
+
+        instanceId: root.instanceId
+    }
+
+    Component.onCompleted: {
+        normalize.init()
+    }
+
+    Column {
+        id: column
+        topPadding: 2
+        bottomPadding: 2
+        spacing: 10
+
+        CheckBox {
+            id: removeDcCheckbox
+            text: qsTrc("effects/normalize", "Remove DC offset (center on 0.0 vertically)")
+
+            onClicked: {
+                normalize.removeDC = !checked
+            }
+            checked: normalize.removeDC
+        }
+
+        RowLayout {
+            width: root.width
+            height: 28
+
+            CheckBox {
+                id: normalizePeakAmplitudeCheckbox
+                text: qsTrc("effects/normalize", "Normalize peak amplitude to")
+                Layout.alignment: Qt.AlignVCenter | Qt.AlignLeft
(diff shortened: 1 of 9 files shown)


src/effects/builtin/normalize/NormalizeView.qml
```diff
@@ -0,0 +1,97 @@
...
+            width: root.width
+            height: 28
+
+            CheckBox {
+                id: normalizePeakAmplitudeCheckbox
+                text: qsTrc("effects/normalize", "Normalize peak amplitude to")
+                Layout.alignment: Qt.AlignVCenter | Qt.AlignLeft
+                Layout.fillWidth: true
+
+
```
kryksyh: Extra newline, and the order of fields here and below is mixed

src/effects/builtin/normalize/NormalizeView.qml
```diff
@@ -0,0 +1,97 @@
...
+                measureUnitsSymbol: qsTrc("global", "dB")
+                decimals: 2
+                step: 0.1
+                enabled: normalizePeakAmplitudeCheckbox.checked
+                Layout.alignment: Qt.AlignVCenter | Qt.AlignRight
+                Layout.preferredWidth: 100
+
+                currentValue: normalize.peakAmplitudeTarget.toFixed(decimals)
+
+                onValueEdited: function(newValue) {
```
kryksyh, line 71: Please note that #7925 (original problem) applies here too

kryksyh (approved):

src/effects/builtin/normalize/NormalizeView.qml
saintmatthieu: Fixed - thanks!

src/effects/builtin/normalize/normalizeviewmodel.cpp
```diff
@@ -0,0 +1,141 @@
...
+        return;
+    }
+
+    fx->mStereoInd = normalizeStereoChannelsIndependently;
+    emit normalizeStereoChannelsIndependentlyChanged();
+}
+
+bool NormalizeViewModel::isApplyAllowed() const
+{
+    return !normalizePeakAmplitude() || peakAmplitudeTarget() <= 0.f;
```
kryksyh: Shouldn't it be 
`
normalizePeakAmplitude() || (peakAmplitudeTarget() && peakAmplitudeTarget() <= 0.f)
`

src/effects/builtin/normalize/NormalizeView.qml
```diff
@@ -0,0 +1,94 @@
...
+                onClicked: {
+                    normalize.normalizePeakAmplitude = !checked
+                }
+                checked: normalize.normalizePeakAmplitude
+            }
+
+            IncrementalPropertyControl {
+                measureUnitsSymbol: qsTrc("global", "dB")
+                decimals: 2
+                step: 0.1
```
kryksyh, line 64: + maxValue: 0

saintmatthieu: I was replicating the behavior in AU3, where positive values are allowed, but it disabled the Apply button.
I can't think of a reason why passing a positive values is useful, if the user can't apply because of this, so I think your suggestion is good. Now it also simplifies the code, since `isAppyAllowed` is always true.

kryksyh: >since isAppyAllowed is always true.

It will be `false` if all checkboxes are off

PR #9122, opened 2025-07-23
9104 noise reduction

### LINKED ISSUE #9027: Disable the Preset Dropdown if only Default Preset is available.
### Description

If a User has opened a plugin with no presets apart from the Default Preset, the dropdown menu should be disabled. Once a new preset has been created apart from Default Preset, the dropdown should be active immideately. 

Image

### LINKED ISSUE #9104: Migrate Noise reduction to AU4

saintmatthieu:
Resolves: #9104
Resolves: #9027

Recommended:

QA:
- [ ] Results are bit-exact with AU3, no matter what parameter combination is chose. (Please talk with me if you're uncertain how to go about this.)


DavidBailes:
@saintmatthieu .
In the noise reduction effect in Audacity 3, group boxes (wxStaticBox) are used to group the two steps. Screen readers automatically read the name of the group box when focus moves to a control within a group box.
It looks like in AU4 group boxes are generally not used. Is there a  reason for this?
Is there a plan for how to make an effect such as this accessible for screen readers, including automatically reading "step1" and "step 2" when appropriate?

saintmatthieu:
@DavidBailes I just tried Narrator with both MuseScore Studio 4 and Audacity 4. It looks like MSS uses technology that reads the name of the focused panel, which I think might be similar to what you describe?
The little I tried in AU4 was the preferences panel, and indeed, I only heard the description of the control in focus - no readout of the panel the control was in. (Does this meet your experience?)
So it seems to me that the tech is there, but we have to take case in AU4 that we make use of it in due time.

DavidBailes:
> @DavidBailes I just tried Narrator with both MuseScore Studio 4 and Audacity 4. It looks like MSS uses technology that reads the name of the focused panel, which I think might be similar to what you describe?

I think that's MuseScore's work around for not using group boxes. In MSS's preferences, it doesn't work at all well with the Jaws screen reader - I don't know if there is a script to get around the problem. Also it would be better for the screen reader to read the role as group/grouping rather than panel, as this is what the user of a screen reader would expect.

> The little I tried in AU4 was the preferences panel, and indeed, I only heard the description of the control in focus - no readout of the panel the control was in. (Does this meet your experience?)

Yes.

> So it seems to me that the tech is there, but we have to take case in AU4 that we make use of it in due time.

So some time down the line, someone's going to have to try and fix this up, and they may or may not be able to get it to work for all screen readers.

I still don't know why MuseScore is not making use of group boxes. Is it that they can't style them to their satisfaction?
Why not make accessibility easier for yourselves.

@saintmatthieu . I appreciate that it might not be your decision as to whether group boxes are used, but I thought that I should at least flag up the issue.

saintmatthieu:
Thanks for your input as always, David. I'll hand this over to @Tantacrul , who is much more knowledgeable than me in that area.

Tantacrul:
OK, I think this is a great point. We'll need a 'group' concept that screen readers can focus on.

@DilsonsPickles - read the above and let's use the new Noise Removal effect in AU4 to introduce this concept for accessibility.

Thanks @DavidBailes

DilsonsPickles:
@DavidBailes - I'm on macOS, so I only have access to VoiceOver (which is apparently not the best) and can't reproduce the behaviour you've described. However, what you've outlined seems straightforward. Just a few clarifying questions, we'll use the Noise reduction as our example:

### Step 1
- Get noise profile

### Step 2
- Noise reduction (dB)
- Sensitivity
- Frequency smoothing (bands)
- Noise: 
-- Reduce
-- Residue

**Questions**
1. As the User tabs between each of the above items, is the AU3 (and desired) behaviour that the narrator always prefixes each of the controls with "Step 1" and "Step 2" respectively, or that it only announcess the group title when the User enters/re-enters that group?

2. Does the narrator read out the descriptions for each step and if so, how is that accessed via keyboard navigation?

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,227 @@
...
+                        text: qsTrc("effects/noisereduction", "Select a few seconds of isolated noise so Audacity knows what to filter out, then click Get noise profile.")
+                        font.family: ui.theme.bodyFont.family
+                        font.pixelSize: ui.theme.bodyFont.pixelSize
+                        wrapMode: Text.Wrap
+                        Layout.fillWidth: true
+                        Layout.leftMargin: 16
+                        Layout.rightMargin: 16
+                    }
+
+                    FlatButton {
```
kryksyh, line 66: I believe this button should be disabled if no selection was made before the dialog is opened, now it will select the whole audio and the results would be suboptimal

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,227 @@
...
+                        Layout.bottomMargin: 16
+
+                        value: noiseReduction.sensitivity
+                        from: 0.01
+                        to: 24
+                        isInt: false
+                    }
+
+                    StyledTextLabel {
+                        text: qsTrc("effects/noisereduction", "Freequency smoothing")
```
kryksyh: a typo Fr**ee**quency

src/effects/builtin/noisereduction/NoiseReductionSlider.qml
```diff
@@ -0,0 +1,49 @@
...
+
+    QtObject {
+        id: prv
+        property int decimals: root.isInt ? 0 : 2
+    }
+
+    spacing: 16
+
+    StyledSlider {
+        id: reductionSlider
```
kryksyh: It is used for the other two fields, so maybe you can choose a more generic name

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,227 @@
...
+import QtQuick.Layouts
+import Muse.UiComponents
+import Audacity.Effects
+
+import "../common"
+
+EffectBase {
+    id: root
+
+    property string title: qsTrc("effects/noisereduction", "Amplify")
```
kryksyh: Noise Reduction

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,227 @@
...
+                color: ui.theme.backgroundSecondaryColor
+                radius: 4
+                width: 211
+                height: getNoiseProfileBox.height
+
+                ColumnLayout {
+                    id: getNoiseProfileBox
+                    width: parent.width
+
+                    Text {
```
kryksyh: Is there a specific reason you use Text instead of StyledTextLabel?

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,227 @@
...
+                        text: qsTrc("effects/noisereduction", "Step 1")
+                        font.family: ui.theme.bodyFont.family
+                        font.pixelSize: ui.theme.bodyFont.pixelSize
+                        font.bold: true
+                        color: ui.theme.fontPrimaryColor
+                        Layout.fillWidth: true
```
kryksyh: Please use the same fields order as in other files

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,227 @@
...
+
+                        Column {
+                            id: radioButtonColumn
+                            width: parent.width
+                            spacing: 8
+
+                            RoundedRadioButton {
+                                width: parent.width
+
+                                checked: noiseReduction.reductionMode === 0
```
kryksyh, line 212: It is better to expose the enum instead of comparing to a magic number

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,227 @@
...
+                        Layout.leftMargin: 16
+                        Layout.rightMargin: 16
+                    }
+
+                    FlatButton {
+                        id: getNoiseProfileButton
+                        text: qsTrc("effects/noisereduction", "Get noise profile")
+                        onClicked: {
+                            if (noiseReduction.getNoiseProfile()) {
+                                root.dialogView.reject()
```
kryksyh: If the operation fails, there will be no feedback. For example if the selected audio data is too short

DavidBailes:
> **Questions**
> 
> 1. As the User tabs between each of the above items, is the AU3 (and desired) behaviour that the narrator always prefixes each of the controls with "Step 1" and "Step 2" respectively, or that it only announces the group title when the User enters/re-enters that group?

@DilsonsPickles . The main point I was trying to make, is that if you use a semantically appropriate control for these these two groups of controls, then the role of the control (group) will be known by the screen reader, and it will automatically read that there's a group box - how it does that may vary slightly from screen reader to screen reader.
It will also pass WCAG 4.1.2: https://www.w3.org/TR/wcag2ict-22/#name-role-value

Here's Narrator's output for AU3, just tabbing around the dialog:

Step 1 group, Get Noise Profile, button, Alt+ g, 
Step 2 group, Noise reduction (dB):, edit, 6, Alt+ n, 
Noise reduction, slider at 12, 
Sensitivity:, edit, 6.00, Alt+ s, 
Sensitivity, slider at 25, 
Frequency smoothing (bands):, edit, 6, Alt+ f, 
Frequency smoothing, slider at 50, 
Reduce, radio button, selected, Alt+ d, 
Cancel, button, 
Help, button,

All of the Windows screen readers currently read the group's name and role only when entering a group box (though older versions of Jaws used to prefix the names of the controls with the name of the group).

See: https://doc.qt.io/qt-6.8/qml-qtquick-controls-groupbox.html
An alternative might be to use a custom control, and set the appropriate accessibility properties, eg accessibility.role: grouping. Qt's accessibility stuff does seems to have a few bugs, so this may of may not be possible.

> 2. Does the narrator read out the descriptions for each step and if so, how is that accessed via keyboard navigation?

By tabbing through the controls (see transcription given above).

src/effects/builtin/noisereduction/NoiseReductionView.qml
saintmatthieu: Good point - fixed

src/effects/builtin/noisereduction/NoiseReductionView.qml
saintmatthieu: But expensive. Atm this enum is within the `src\effects\builtin\noisereduction` scope ; for QML to have it registered I'd have to expose it to the `Audacity.Effects` module (beside the other additional work). In this case I'd rather accept this shortcut.

src/effects/builtin/noisereduction/NoiseReductionView.qml
saintmatthieu: StyledTextLabel is better indeed 👍

src/effects/builtin/noisereduction/NoiseReductionView.qml
saintmatthieu: In UX terms you're probably right, but I wonder what changes I'd need for that: it's a condition specific to NR, and it'd require modifications in the effect execution scenario, I think. By the time the dialog is open, it's already too late.

src/effects/builtin/noisereduction/noisereductioneffect.h
```diff
@@ -0,0 +1,215 @@
...
+    double mNewSensitivity = sensitivityDefault; // - log10 of a probability... yeah.
+    int mFreqSmoothingBands = freqSmoothingBandsDefault;
+    int mNoiseGain = noiseGainDefault; // in dB, positive
+    int mNoiseReductionChoice = noiseReductionChoiceDefault;
+
+    // These could be parametrized in ancient times ; to reduce the diff we make them const now.
+    static constexpr int mWindowTypes = WT_DEFAULT_WINDOW_TYPES;
+    static constexpr int mWindowSizeChoice = DEFAULT_WINDOW_SIZE_CHOICE;
+    static constexpr int mStepsPerWindowChoice = DEFAULT_STEPS_PER_WINDOW_CHOICE;
+    static constexpr auto mMethod = DM_DEFAULT_METHOD;
```
kryksyh: You could change auto to int, this fixes the compilation on macos. DM_DEFAULT_METHOD type could be changed to int too.

src/effects/builtin/noisereduction/NoiseReductionView.qml
saintmatthieu: I imagined a way in the "Prevent inadvertent whole-project selection when profiling" commit.

src/effects/builtin/noisereduction/noisereductioneffect.h
```diff
@@ -142,6 +142,8 @@ class NoiseReductionEffect : public EffectWithSettings<NoiseReductionSettings, S
     // Effect implementation
 
     const EffectParameterMethods& Parameters() const override;
+    // In case the user intends to get a noise profile, selecting the whole project makes little sense.
+    bool applyEffectToAllAudio() const { return false; }
```
saintmatthieu: Ooops - override.

src/effects/builtin/noisereduction/NoiseReductionSlider.qml
```diff
@@ -0,0 +1,45 @@
...
+        anchors.verticalCenter: root.verticalCenter
+        onMoved: {
+            slider.value = prv.decimals === 0 ? Math.round(slider.value) : parseFloat(slider.value.toFixed(prv.decimals));
+        }
+    }
+
+    IncrementalPropertyControl {
+        id: textControl
+        width: (root.width - root.spacing) * 0.4
+        decimals: root.isInt ? 0 : 2
```
kryksyh, line 42: You can use `prv.decimals` here

src/effects/builtin/noisereduction/NoiseReductionSlider.qml
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    spacing: 16
+
+    StyledSlider {
+        id: slider
+        width: (root.width - root.spacing) * 0.6
+        anchors.verticalCenter: root.verticalCenter
+        onMoved: {
+            slider.value = prv.decimals === 0 ? Math.round(slider.value) : parseFloat(slider.value.toFixed(prv.decimals));
```
kryksyh: This is a bit problematic. If you assign the value to a property, you break the property binding. 
After this, the binding `model -> slider.value` is destroyed, and when the model emits propertyChanged() signal, the value will not be updated.

kryksyh: You can verify it by loading a preset. Before you change the value via the slider, the preset value is displayed properly. After you move the slider, the slider stops responding to preset changes.

saintmatthieu: Addressed

src/effects/builtin/noisereduction/NoiseReductionView.qml
kryksyh: @saintmatthieu could you address this?

src/effects/builtin/noisereduction/NoiseReductionSlider.qml
```diff
@@ -0,0 +1,53 @@
...
+        id: textControl
+        width: (root.width - root.spacing) * 0.4
+        decimals: root.isInt ? 0 : 2
+        minValue: root.from
+        maxValue: root.to
+        step: 1
+        currentValue: root.value
+        onValueEdited: function(newValue) {
+            if (root.value !== newValue) {
+                newValueRequested(newValue);
```
kryksyh: Unnecessary semicolons here and in other places

src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -0,0 +1,233 @@
...
+                        Column {
+                            id: radioButtonColumn
+                            width: parent.width
+                            spacing: 8
+
+                            RoundedRadioButton {
+                                id: audioWithNoiseRemovedRadioButton
+                                width: parent.width
+
+                                checked: noiseReduction.reductionMode == 0
```
kryksyh: Please use `===` to please the linter

src/effects/builtin/noisereduction/NoiseReductionView.qml
saintmatthieu: What is the order in other files?
TrackItem.qml:
```
        Rectangle {
            id: spacer

            color: "transparent"

            Layout.fillHeight: true
            Layout.preferredWidth: 9
        }
```
PreferencesDialog.qml:
```
            PreferencesMenu {
                id: menu

                Layout.fillHeight: true
                Layout.preferredWidth: 220

                navigation.section: root.navigationSection
                navigation.order: 1

                model: preferencesModel
            }
```

kryksyh: There is this guideline we use for audacity too: https://github.com/musescore/muse_framework/wiki/CodeGuidelines#qml

saintmatthieu: I guess it's the lacking line breaks that made you tick.
Pushed. I hope this will be everything.

kryksyh: I mean the order of fields is 

    id: someID

    // newline here
    
    anchors.someProperty: ...
    anchors.another: ...
    // Or:
    Layout.someProperty: ...

    // newline here

    someRegularProperty: ...
    anotherOne: ...

    // newline here

    navigation.someProperty: ...

    // newline here

    onSomeSignal: {
        doSomething()
    }

    onSomeSignalWithParameter: function(param) {
        soSomethingWith(param)
    }
    
    
In a few places in this file, the order is completely random, it would be nice if you could clean it up

saintmatthieu: Last attempt pushed. After that I give up.

kryksyh (approved):

kryksyh (approved):

Issue #8755, opened 2025-05-12
Audacity crashes if you zoom in enough to see samples and try to click into the waveform while playing audio


dozzzzer:
### Steps to reproduce

1. Load an audio clip
2. Start playback
3. Zoom in enough to see individual samples
4. Observe: the cursor has changed to the "pencil"
5. Click into the waveform

### Expected behavior

While playback is active, using the Draw Tool should be blocked. The cursor should be changed correspondingly.

AU3:

<img width="502" alt="Image" src="https://github.com/user-attachments/assets/2d68f4f8-e009-4cd4-af4b-2231c8289ba0" />

### Actual behavior

A crash

### Audacity Version

master

### Operating system

OS independent

chinakov:
@embarc-gabriel should use standart cursor called 'NotAllowed'

luapmartin:
To test:
- visuals of cursor during playback AND when playback is stopped at the different zoom levels (sample editing...) 
- clip sample edition is properly disabled during playback

Thanks

chinakov:
https://github.com/user-attachments/assets/9ec3a65a-85bb-4794-a9b3-344b19115f07

One minor issue - when the NotAllowed cursor is shown while zoomed in, zooming out keeps the cursor active until the User has moved the cursor away from the clip.

luapmartin:
zoom level changes are now updating the cursor.

Since the beginning of this work another work on split tool and some cursor cleanup on the brush tool have been made, please be considering those other cursors while testing. there is also a task that has been created to compile various cursors issues in order to tackle them with a bigger scope perspective. https://github.com/audacity/audacity/issues/9204 for now it's kind of a placeholder .

dozzzzer:
Tested

PR #9136, opened 2025-07-25
8755 integrate playback state management to prevent sample editing during playback (update cursor accordingly)

luapmartin:
Resolves: [8755](https://github.com/audacity/audacity/issues/8755)

- Cursor has been updated
- sample edition isn't permitted anymore during playback

Recommended:


src/playback/view/common/playbackstatemodel.h
```diff
@@ -7,18 +7,22 @@
...
 public:
     explicit PlaybackStateModel(QObject* parent = nullptr);
 
-    Q_INVOKABLE bool isPlaying();
-    Q_INVOKABLE bool isPaused();
+    Q_INVOKABLE bool isPlaying() const;
+    Q_INVOKABLE bool isPaused() const;
+
+signals:
+    void isPlayingChange();
```
kryksyh: It should be `isPlayingChanged`

luapmartin: nice catch

kryksyh (approved):

src/playback/view/common/playbackstatemodel.h
```diff
@@ -7,18 +7,22 @@
...
-
+    Q_PROPERTY(bool isPlaying READ isPlaying NOTIFY isPlayingChanged)
     muse::Inject<au::playback::IPlaybackController> controller;
 
 public:
     explicit PlaybackStateModel(QObject* parent = nullptr);
 
-    Q_INVOKABLE bool isPlaying();
-    Q_INVOKABLE bool isPaused();
+    Q_INVOKABLE bool isPlaying() const;
```
Eism: If you decided to move this into a property, then please remove Q_INVOKABLE here and replace all calls with the property (so that the model has only one way to access the data).

Also, please do the same with isPaused, since you decided to change the approach.

src/playback/view/common/playbackstatemodel.h
```diff
@@ -7,18 +7,22 @@
 
 #include "iplaybackcontroller.h"
 #include "modularity/ioc.h"
+#include "async/asyncable.h"
```
Eism:
```suggestion
#include <QObject>

#include "async/asyncable.h"

#include "modularity/ioc.h"
#include "iplaybackcontroller.h"
```

src/projectscene/view/clipsview/waveview.cpp
```diff
@@ -370,6 +370,12 @@ void WaveView::setLastClickPos(const unsigned lastX, const unsigned lastY, const
         return;
     }
 
+    // Prevent sample editing during playback
+    bool isPlaying = globalContext()->playbackState()->playbackStatus() == playback::PlaybackStatus::Running;
+    if (isPlaying) {
+        return;
+    }
```
Eism: you can move it to a separate method and reuse it

`
if (isPlaying()) {
    return;
}
`

Eism (approved):

kryksyh (approved):

PR #9169, opened 2025-07-30
9103 graphical eq

### LINKED ISSUE #9103: Migrate Graphic EQ (not Filter Curve EQ) to AU4

saintmatthieu:
Resolves: #9103

Recommended:


src/effects/builtin/graphiceq/GraphicEqBoard.qml
```diff
@@ -0,0 +1,76 @@
...
+    property var bandsModel: null
+
+    width: gridLines.width
+    height: prv.labelHeight + prv.labelBottomMargin + prv.faderHeight
+
+    QtObject {
+        id: prv
+
+        readonly property int minDdBain: -20
+        readonly property int maxDdBain: 20
```
luapmartin: I guess it's a typo for:
```suggestion
        readonly property int maxDbGain: 20
```
it's confusing to see Db instead of dB but I guess that's ok

Same for `minDbGain`

src/effects/builtin/graphiceq/GraphicEqBoard.qml
```diff
@@ -0,0 +1,76 @@
...
+Item {
+    id: board
+
+    property var bandsModel: null
+
+    width: gridLines.width
+    height: prv.labelHeight + prv.labelBottomMargin + prv.faderHeight
+
+    QtObject {
+        id: prv
```
luapmartin, line 16: `prv` what is it ? a `provider` for the eq, the bands ... views maybe we could call it `EqViewConfig` or similar

src/effects/builtin/graphiceq/GraphicEqBoard.qml
luapmartin: As discussed together, the dB range values (min and max) to control the band fader might be better living in the Model, thanks

src/effects/builtin/graphiceq/GraphicEqBoard.qml
```diff
@@ -0,0 +1,76 @@
...
+                width: fader.width
+                spacing: prv.labelBottomMargin
+
+                StyledTextLabel {
+                    width: 16
+                    height: prv.labelHeight
+                    font.pixelSize: prv.fontSize
+                    horizontalAlignment: Text.AlignHCenter
+                    verticalAlignment: Text.AlignVCenter
+                    text: model.centerFreq
```
luapmartin, line 56: As also discussed together, the center Frequency value for which a band slider is controlling the dB boost/cut might also be better living in the preset, thanks  

this would imply to update the current presets to add the frequency value alongside the dB boost/cut values

still we would need to have a way to get the default band values (might be stored in the init preset)

This change will also allow some presets to have more or less bands, we might want to make sure the view is ready for that.

src/effects/builtin/graphiceq/GraphicEqBoard.qml
luapmartin: after our discussion I understood that this is for the private member trick.

Now I wonder if we should put aside any magic number into the ViewModel and solely rely on it.

src/effects/builtin/graphiceq/GraphicEqFader.qml
```diff
@@ -18,7 +18,7 @@ Item {
         id: faderTrack
         width: 2
         height: root.height
-        color: "#C0C5CE"
+        color: ui.theme.isDark ? "#191F24" : "#CFCFD9"
```
luapmartin: should we use `ui.theme.backgroundQuarternaryColor` ?

saintmatthieu: What makes you say this? From @DilsonsPickles I only got Hex codes.

src/effects/builtin/graphiceq/GraphicEqFaderHandle.qml
```diff
@@ -0,0 +1,19 @@
...
+Rectangle {
+    width: 16
+    height: 32
+    radius: 2
+    color: ui.theme.backgroundPrimaryColor
+    border.color: ui.theme.isDark ? "#8C9196" : "#5B5B5F"
+    border.width: 1
+
+    Rectangle {
+        width: 10
```
kryksyh, line 15: It would be nice to have an id in such cases, to make it easier to understand what it is, like
```
id: notch
```

src/effects/builtin/graphiceq/GraphicEqGridLines.qml
```diff
@@ -0,0 +1,57 @@
...
+    property int lineWidth
+    property int fontSize
+    // Out
+    property int gridlineHorizontalCenter: prv.labelWidth + prv.labelLineSpacing + root.lineWidth / 2
+    width: prv.labelWidth + prv.labelLineSpacing + root.lineWidth
+
+    QtObject {
+        id: prv
+        readonly property int labelLineSpacing: 8
+        readonly property int labelWidth: 16
```
kryksyh: You probably have to use font metrics here (see VolumeTooltop.qml for an example usage)
On Mac, the bottom part is ellipsized.  
image

src/effects/builtin/graphiceq/GraphicEqGridLines.qml
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        delegate: Item {
+            width: root.lineWidth
+            height: 1
+
+            StyledTextLabel {
+                id: label
+                width: prv.labelWidth
+                y: line.y - 6
```
kryksyh, line 46: This one is better to calculate via font metrics too:
```
y: line.y - (metrics.ascent + metrics.descent) / 2
```

Or maybe drop `descent` since we only have digits here.

src/effects/builtin/graphiceq/graphiceqbandsmodel.h
```diff
@@ -0,0 +1,47 @@
...
+    GraphicEqBandsModel(QObject* parent, std::function<GraphicEq* ()> eqGetter);
+
+    static constexpr auto NUM_BANDS = 31;
+    void reload();
+
+    Q_INVOKABLE void flatten();
+    Q_INVOKABLE void invert();
+
+private:
+    enum RoleNames
```
kryksyh: a better name for the enum would be "CustomRoles", or "EqDataRoles`, etc, as those are not role names.

src/effects/builtin/graphiceq/graphiceqbandsmodel.cpp
```diff
@@ -0,0 +1,160 @@
...
+    } else {
+        const auto khz = frequency / 1000.0;
+        return QString::number(khz, 'f', (khz == static_cast<int>(khz)) ? 0 : 1) + "k";
+    }
+}
+}
+
+GraphicEqBandsModel::GraphicEqBandsModel(QObject* parent, std::function<GraphicEq* ()> eqGetter)
+    : QAbstractListModel(parent),  m_getEq(std::move(eqGetter)), mSliders([this]() -> EqualizationCurvesList* {
+    GraphicEq* const eq = m_getEq();
```
kryksyh, line 25: Is the reason for such complications (passing a lambda to get a lambda to fetch a value) that the model is created before the effect instance is accessible?

Isn't it possible to delay the initialization, and use the original approach with references?

src/effects/builtin/graphiceq/GraphicEqFaderHandle.qml
saintmatthieu: Yep.

src/effects/builtin/graphiceq/GraphicEqGridLines.qml
saintmatthieu: Cool, thanks! I noticed that problem on my compressor PR, too.

src/effects/builtin/graphiceq/GraphicEqGridLines.qml
saintmatthieu: Sounds good, and it works on Mac, but for some reason not on Windows: 
image
I've spent a bit of time trying to debug that without success.

src/effects/builtin/graphiceq/graphiceqbandsmodel.h
saintmatthieu: Sure.

src/effects/builtin/graphiceq/graphiceqbandsmodel.cpp
saintmatthieu: Yes.
Delaying creation of `mBandsModel` in  GraphicEqViewModel is something I tried, but it brings other complications.
I just tried something else, passing a reference to an EqualizationCurvesList pointer, and right now it works, but I just realize it's less safe: upon de-init, maybe the effect has expired.

src/effects/builtin/graphiceq/GraphicEqBoard.qml
saintmatthieu: I agree it'd be better, but it would come at a refactoring price that seems too big for what it brings.
There are other improvements I've noticed: if the track's sample rate is say 8kHz, you still have faders at 20kHz. It's handled in the code, but the user can only guess what might happen.
But these would be for later, if there's a demand and this legacy EQ doesn't get superseded by something more modern.

src/effects/builtin/graphiceq/GraphicEqBoard.qml
saintmatthieu: What magic numbers? Those?
```cpp
        readonly property int faderHeight: 352
        readonly property int labelHeight: 16
        readonly property int labelBottomMargin: 16
        readonly property int fontSize: 10
```
Like we discussed, I don't think that view models should care about layout, but only about getting the data needed by the views.

kryksyh (approved):

Issue #9187, opened 2025-08-04
After changing the effect's parameter and clicking Cancel, the change gets remembered


dozzzzer:
1. Open a destructive effect
2. Change one of its parameters
3. Click Cancel
4. Re-open the effect
5. Observe: the parameter change has been remembered.

chinakov:
Still reproducible on the latest Master (https://github.com/audacity/audacity/commit/129c883ee531ba7ffbbdfa568287fddb4d1d70d0)

chinakov:
This is still reproducible on the latest Master (e9ead184ef595365efaef0ae54167634b0fd8da7)

PR #9190, opened 2025-08-04
9173 destructive compressor

### LINKED ISSUE #9173: Implement destructive Compressor in AU4
### Description

[Figma link](https://www.figma.com/design/w2kAh1gKKV5bu7htY7158W/08---Effects?node-id=4782-60134&p=f&t=vO9HGElzyMSBXVrg-0)

### LINKED ISSUE #9189: Parameters in Plugins should be named in sentence case
### Bug description

All plugins within Audacity 4 should have sentence case names in their parameters.  

However, in Reverb plugin, the 'Room Size' and 'Stereo Width' parameters are in title case. 

This is only present in Reverb plugin. 

### Steps to reproduce

1. Launch AU4
2. Generate some audio
3. Open Reverb
4. Observe result

### Expected behavior

The parameters should be sentence case

### Actual behavior

The parameters are not sentence case 

### Audacity Version

e46ee35fd7338d8c6d79aee6a428794c6f594d50

### Operating system

OS independent

### Additional context

_No response_

saintmatthieu:
Resolves: #9173
Resolves: #9189

Beside resolving the above, this PR
* Introduces `AbstractEffectSettingModel`, a simple model for one single AU3 parameter. The remaining work to write new effect models is hence drastically reduced. It is an alternative to generalizing `ReverbViewModel`.
* Generalizes `ParameterKnob` and `BigParameterKnob` and reuse them in compressor and limiter views,
* Fixes a limitation (bug?) so that `ParameterKnob` and `BigParameterKnob` support decimal places (b23a9ad42dcc3b7c8cafa2964740f40a18ccc286)
* Modifies behavior of `KnobControl` so that, independently on the range and step size, it can be turned all the way by moving the mouse 200px up or down / left or right
* Makes `KnobControl` less jumpy by using using consecutive mouse position differences rather than difference to initial position
* Allows `KnobControl` to change its value exponentially rather than linearly, so that settings where a greater sensitivity is desired in lower ranges become easier to control

Note: in fact, compressor and limiter can be used in real-time mode already, only without the animation.

Recommended:

QA:
- [x] Smoke-test other built-in effects and generators (a bit of refactoring was done), in particular the reverb effect
- [x] Knobs in UIs are still functional
- [x] Knobs: shift+drag still works
- [x] Settings work (persistent on per-track level upon closing and reopening dialog and project) (please beware of https://github.com/audacity/audacity/issues/9187, not specific to this PR)
- [x] Please test bit-exactness wrt AU3 (tip: use a loud drum beat to make sure compressor and limiter kick in)
- [ ] Please remind @saintmatthieu to drop unnecessary commit [8c5a98cc21cb0b5c8f7fc31dc4aa39b6087ef161](https://github.com/audacity/audacity/pull/9190/commits/8c5a98cc21cb0b5c8f7fc31dc4aa39b6087ef161) before merging


au3/libraries/lib-components/SettingsVisitor.h
```diff
@@ -28,7 +28,7 @@ template<
     const Type def{};         //!< Default value
     const Type min{};         //!< Minimum value
     const Type max{};         //!< Maximum value
-    const Type scale{};       //!< Scaling factor, for slider control
+    const Type step{};       //!< Step
```
kryksyh: please add one extra space here :)

src/effects/builtin/compressor/CompressionCurve.qml
```diff
@@ -0,0 +1,113 @@
...
+        readonly property int step: 12
+        readonly property int tickLength: 4
+        readonly property var ticks: (function() {
+            var result = [];
+            for (var i = prv.min; i <= prv.max; i += prv.step) {
+                result.push(i);
+            }
+            return result;
+        })()
+        readonly property color lineColor: Qt.rgba(123/255, 123/255, 129/255, 1)
```
kryksyh: Please use the standard hex notation, it will be easier to find and replace when we will refactor colors

src/effects/builtin/compressor/CompressionCurve.qml
```diff
@@ -0,0 +1,113 @@
...
+            delegate: Item {
+                y: background.height * (1 - index / (prv.ticks.length - 1))
+
+                StyledTextLabel {
+                    width: prv.labelWidth
+                    height: prv.labelHeight
+                    horizontalAlignment: Text.AlignRight
+                    anchors.left: hLine.right
+                    anchors.leftMargin: prv.labelMargin
+                    y: hLine.y - 7 // So the minus sign is centered
```
kryksyh: Font metrics?

src/effects/builtin/compressor/CompressionCurvePainter.qml
```diff
@@ -0,0 +1,54 @@
...
+        // Draw area under curve
+        ctx.beginPath();
+        ctx.moveTo(points[0].x, height); // Start at bottom left
+        for (let j = 0; j < points.length; ++j) {
+            const pt = points[j];
+            ctx.lineTo(pt.x, pt.y);
+        }
+        ctx.lineTo(points[points.length - 1].x, height); // Down to bottom right
+        ctx.closePath();
+        ctx.fillStyle = "rgba(80,202,255,0.1)";
```
kryksyh: please use the standard hex notation

src/effects/builtin/compressor/CompressionCurvePainter.qml
```diff
@@ -0,0 +1,54 @@
...
+        ctx.strokeStyle = "#50caff";
+        ctx.lineWidth = 1;
+        ctx.beginPath();
+        for (let j = 0; j < points.length; ++j) {
+            const pt = points[j];
+            if (j === 0)
+                ctx.moveTo(pt.x, pt.y);
+            else
+                ctx.lineTo(pt.x, pt.y);
+        }
```
kryksyh:
```suggestion
        ctx.moveTo(points[0].x, points[0].y)
        points.slice(1).forEach(pt => ctx.lineTo(pt.x, pt.y))
```

src/effects/builtin/common/BigParameterKnob.qml
```diff
@@ -21,26 +24,50 @@ Item {
         if (parameter) {
             knob.from = parameter["min"]
             knob.to = parameter["max"]
-            knob.value = parameter["value"]
-            textEdit.measureUnitsSymbol = parameter["unit"]
+            warper.value = parameter["value"]
+            knob.stepSize = parameter["step"] || 1;
```
kryksyh: please remove semicolons here and in other files

src/effects/builtin/compressor/CompressionCurvePainter.qml
```diff
@@ -0,0 +1,54 @@
...
+
+        // Draw area under curve
+        ctx.beginPath();
+        ctx.moveTo(points[0].x, height); // Start at bottom left
+        for (let j = 0; j < points.length; ++j) {
+            const pt = points[j];
+            ctx.lineTo(pt.x, pt.y);
+        }
+        ctx.lineTo(points[points.length - 1].x, height); // Down to bottom right
+        ctx.closePath();
```
kryksyh:
```suggestion
        ctx.beginPath()
        ctx.moveTo(points[0].x, height)

        points.forEach(pt => ctx.lineTo(pt.x, pt.y))

        ctx.lineTo(points[points.length - 1].x, height)
        ctx.closePath()
```

src/effects/builtin/compressor/compressoreffect.cpp
```diff
@@ -0,0 +1,95 @@
...
+std::unique_ptr<EffectOutputs> CompressorEffect::MakeOutputs() const
+{
+    return std::make_unique<DynamicRangeProcessorDummyOutputs>();
+}
+
+bool CompressorEffect::CheckWhetherSkipEffect(
+    const EffectSettings& settings) const
+{
+    const auto& compressorSettings = GetSettings(settings);
+    return compressorSettings.compressionRatio == 1
```
kryksyh: do we need muse::is_equal here?

src/effects/builtin/compressor/CompressionCurvePainter.qml
```diff
@@ -0,0 +1,54 @@
...
+        for (let i = 0; i <= n; ++i) {
+            const db = prv.min + (prv.max - prv.min) * i / n;
+            x.push(width * (db - prv.min) / (prv.max - prv.min));
+            dbIn.push(db);
+        }
+
+        const dbOut = model.compressionCurve(dbIn);
+        const points = [];
+        for (let i = 0; i < dbIn.length; ++i) {
+            const db = dbIn[i];
```
kryksyh: db is not used, hence no need in dbIn array

src/effects/builtin/compressor/CompressionCurvePainter.qml
```diff
@@ -0,0 +1,54 @@
...
+    required property var model
+
+    onPaint: {
+        let ctx = getContext("2d");
+        ctx.clearRect(0, 0, width, height);
+
+        const dbIn = [];
+        const x = [];
+        const n = 40;
+        for (let i = 0; i <= n; ++i) {
```
kryksyh: Here you are creating a uniformily distributed points in range [-60; 0] to pass  it to model.compressionCurve. dbIn is not used elsewhere, then it is better to do all calculation on the model's side

src/effects/builtin/common/controlwarper/controlwarper.h
```diff
@@ -0,0 +1,64 @@
+/*
```
kryksyh: I would rather name it `ValueWarper`, ControlWarper feels a bit misleading

src/effects/builtin/compressor/CompressionCurve.qml
saintmatthieu: We had this conversation on the Graphic EQ PR already: https://github.com/audacity/audacity/pull/9169#discussion_r2254907465
Still haven't debugged :)

src/effects/builtin/compressor/CompressionCurvePainter.qml
saintmatthieu: I was glad that AI could spurt out QML painting code, which I had no experience with, but I haven't reviewed it since, thanks for raising these points.
I migrated the calculation of the compression curve based on min, max db values and number of steps, but kept the mapping of these dB values to x and y pixel positions.

src/effects/builtin/common/controlwarper/controlwarper.h
saintmatthieu: Could you clarify, misleading in what direction?
To me, a "value" doesn't necessarily have a min or a max, which is a condition to warping. Maybe "range warper" ?

au3/libraries/lib-components/SettingsVisitor.h
saintmatthieu: ... or leave just one space, like for the members above:
```cpp
    typename Structure, //!< Structure holding the parameter
    typename Member, //!< Often the same as Type
    typename Type, //!< Type of the given default value
    typename Value = Type //!< A type constructible from the default
    > struct EffectParameter {
    Member Structure::* const mem{}; //!< Member holding the parameter
    const wxChar* const key{}; //!< Identifier in configuration file
    ```

File: src/effects/builtin/common/BigParameterKnob.qml (hunk 6 above)
Comment (saintmatthieu) [reply src/effects/builtin/common/BigParameterKnob.qml]: I double-checked, and the QML VSCode extension actually has a formatter.
So I set it up and ran the format document column. To my dismay, it added semicolons to those statements which didn't have one rather than the other way round.
It seems to be so because the formatter follows the Qt recommendations, which, to my surprise, actually recommend ending statements with a semicolon: https://doc.qt.io/qt-6/qml-codingconventions.html#javascript-code

Did a mistaken belief creep in somewhere that there should be none?

Comment (saintmatthieu) [reply src/effects/builtin/common/BigParameterKnob.qml]: I removed them from the BigKnobParameter file since I haven't authored it and don't want to mess up someone else's work. There's many of them in the files I've added, I'll leave them at least for now. It's handy: now I have a QML formatter :D

File: src/effects/builtin/compressor/CompressionCurve.qml (hunk 3 above)
Comment (kryksyh) [reply src/effects/builtin/compressor/CompressionCurve.qml]: I tested how it looks on windows for me. With the lucky seven there is a slight offset 
image

with font metrics it's more or less the same:
image


If there is no difference, why not go with metrics to be future-proof if the font size changes?

File: src/effects/builtin/common/BigParameterKnob.qml (hunk 6 above)
Comment (kryksyh) [reply src/effects/builtin/common/BigParameterKnob.qml]: I dont think that using mixed formatting is a good thing, if you need a script that will remove trailing semicolons before commit ask me

File: src/effects/builtin/common/controlwarper/controlwarper.h (hunk 11 above)
Comment (kryksyh) [reply src/effects/builtin/common/controlwarper/controlwarper.h]: What I meant is that the class doesn't really operate on the "control" (as if UI control), but on the numeric value. 

"RangeWarper" could also work, but to me that sounds more like it would map between different ranges, but it also warps the distribution.

File: src/effects/builtin/builtineffectsmodule.cpp [hunk 12]
```diff
@@ -50,6 +51,8 @@ void BuiltinEffectsModule::registerUiTypes()
 {
     qmlRegisterUncreatableType<AbstractEffectModel>("Audacity.BuiltinEffects", 1, 0, "AbstractEffectModel", "Not creatable abstract type");
     qmlRegisterType<BuiltinEffectViewLoader>("Audacity.BuiltinEffects", 1, 0, "BuiltinEffectViewLoader");
+    qmlRegisterType<ValueWarper>("Audacity.BuiltinEffects", 1, 0, "ValueWarper");
+    qmlRegisterUncreatableType<ValueWarpingTypes>("Audacity.BuiltinEffects", 1, 0, "ValueWarpingType", "Not creatable from QML");
```
kryksyh, line 55: is it `ValueWarpingTypes` or `ValueWarpingType`?

src/effects/builtin/common/valuewarper/valuewarpertypes.h
```diff
@@ -0,0 +1,31 @@
...
+public:
+    enum class ValueWarpingType {
+        None,
+        Soft,
+        Aggressive,
+    };
+    Q_ENUM(ValueWarpingType)
+};
+
+using ValueWarpingType = ValueWarpingTypes::ValueWarpingType;
```
kryksyh, line 21: What is the purpose of this trick?

kryksyh: Is it for exposing enum class to qml?

src/effects/builtin/builtineffectsmodule.cpp
saintmatthieu: `ValueWarpingTypes` is intended as template argument, but from this class we register the `ValueWarpingType` enum so as to be able to use it in QML.

src/effects/builtin/common/valuewarper/valuewarpertypes.h
saintmatthieu: This alias isn't necessary to expose the enum to QML, it's only a C++ shorthand.

kryksyh (approved):

PR #9200, opened 2025-08-06
9106 click removal

### LINKED ISSUE #9106: Migrate Click removal to AU4

saintmatthieu:
Resolves: #9106

Recommended:

QA:
- [x] Please test Amplify for regressions. (It had a few defects already, so if you find one on this branch, please verify that it's not on master, too - this PR doesn't intend to fix them, it just needed a bit of refactoring.)
- [x] Like in AU3, selecting a very small region (less than 4096 samples) yields an error

### DIFF
--- a/src/effects/builtin/amplify/amplifyviewmodel.cpp
+++ b/src/effects/builtin/amplify/amplifyviewmodel.cpp
@@ -51,15 +51,15 @@ void AmplifyViewModel::update()
     if (!muse::is_equal(m_amp.val, amp.val)) {
         m_amp = amp;
         m_newPeak = newPeak;
-        emit ampChanged();
-        emit newPeakChanged();
+        emit ampValueChanged();
+        emit newPeakValueChanged();
     }
 
     bool isApplyAllowed = ae->isApplyAllowed();
     setIsApplyAllowed(isApplyAllowed);
 }
 
-QString AmplifyViewModel::title() const
+QString AmplifyViewModel::effectTitle() const
 {
     return muse::qtrc("effects/amplify", "Amplify");
 }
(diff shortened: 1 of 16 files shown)


luapmartin (approved):
Thanks, I think thats great to now have this SliderWithTextInput, nice

_one thing that we might want to review is the main parameter name as for instance amp could be ampValue same for newPeakValue but where that's better it's for width that could be a bit misleading so having widthValue is better in my opinion_

src/effects/builtin/clickremoval/ClickRemovalView.qml
```diff
@@ -0,0 +1,67 @@
...
+                clickRemoval.threshold = newValue
+            }
+        }
+
+        SliderWithTextInput {
+            id: widthSlider
+
+            width: parent.width
+            text: clickRemoval.widthLabel
+            value: clickRemoval.width
```
luapmartin:
```suggestion
            value: clickRemoval.widthValue
```

src/effects/builtin/clickremoval/ClickRemovalView.qml
```diff
@@ -0,0 +1,67 @@
...
+        height: implicitHeight
+        width: parent.width
+        spacing: 16
+
+        SliderWithTextInput {
+            id: thresholdSlider
+
+            width: parent.width
+            text: clickRemoval.thresholdLabel
+            value: clickRemoval.threshold
```
luapmartin:
```suggestion
            value: clickRemoval.thresholdValue
```

src/effects/builtin/amplify/amplifyviewmodel.cpp
```diff
@@ -90,6 +101,26 @@ float AmplifyViewModel::ampMax() const
...
+}
+
+double AmplifyViewModel::ampStep() const
+{
+    return 0.02;
+}
+
+QString AmplifyViewModel::newPeakLabel() const
+{
+    return muse::qtrc("effects/amplify", "New peak amplitude");
```
saintmatthieu, line 121: Like I said during a meeting, I think it's fine for e.g. title to be part of the model, since this is something that may be reused someplace else, such as in the menu, so it's probably good to bring it one step closer to the effect's implementation. Yet where this label and the one about clipping, it's less evident to me. Grey zone? Either way it's no biggy here, so I won't bother reverting. I also would like to honor the fact that you've contributed to this PR, which I appreciate :)

PR #9208, opened 2025-08-09
AbstractEffectModel instance ID is const !

saintmatthieu:
As it stands on master, `instanceId` is a property of `EffectBase.qml` that defaults to `null`.

Since in QML, child elements are instantiated first, the effect model was instantiated with `null` as instance ID, and then `EffectBase.instanceId` is initialized to a proper value. In other words, there is no guarantee in the models that instance ID isn't null, which requires sanity checks and less optimized code.

We know for a fact, though, that the instance ID exists before built-in effect models are created. (By the way, `BuiltinEffectModel` is probably a better name.) This PR proposes a way setting m_instanceId at construction time, making it const and guaranteeing availability of effect instances and settings the entire lifetime of the model and its many derived classes ...

Recommended:

QA:
This is a pure refactoring PR and no behavioral change is expected.
The base class of built-in effects and generators was modified, and the specific effects and generators were accordingly adapted. This is meant for the better, but a mistake in the process cannot be excluded.

Please smoke-test (see below) all built-in effects and generators, also in realtime mode for those which support it, on at least one platform (preferably MacOS or Linux, as I already have done my own testing on Windows). On the other platforms, smoke-testing one generator and one effect is then probably sufficient.
The smoke test should include
* opening the UI
* changing at least one setting
* applying
* re-opening the UI and making sure the settings were remembered
* changing the presets

### DIFF
--- a/src/effects/builtin/common/abstracteffectmodel.cpp
+++ b/src/effects/builtin/common/abstracteffectmodel.cpp
@@ -3,27 +3,21 @@
 */
 #include "abstracteffectmodel.h"
 
+#include "../view/builtineffectviewloader.h"
+
 #include "log.h"
 
 using namespace au::effects;
 
 AbstractEffectModel::AbstractEffectModel(QObject* parent)
-    : QObject(parent)
+    : QObject(parent), m_instanceId(BuiltinEffectViewLoader::initializationInstanceId())
 {
+    assert(m_instanceId != -1);
 }
 
 void AbstractEffectModel::init()
 {
-    if (m_inited) {
-        return;
-    }
-
-    EffectInstanceId id = this->instanceId();
-    IF_ASSERT_FAILED(id != 0) {
-        return;
-    }
-
-    instancesRegister()->settingsChanged(id).onNotify(this, [this]() {
+    instancesRegister()->settingsChanged(m_instanceId).onNotify(this, [this]() {
         doReload();
     });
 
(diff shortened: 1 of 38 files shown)


src/effects/builtin/common/abstracteffectmodel.cpp
```diff
@@ -18,12 +20,13 @@ void AbstractEffectModel::init()
         return;
     }
 
-    EffectInstanceId id = this->instanceId();
-    IF_ASSERT_FAILED(id != 0) {
+    m_instanceId = BuiltinEffectViewLoader::initializationInstanceId();
```
kryksyh: Wouldn't it be better to move this to the constructor?

src/effects/builtin/common/abstracteffectmodel.cpp
```diff
@@ -42,42 +44,35 @@ bool AbstractEffectModel::inited() const
...
-    }
-
-    return instancesRegister()->settingsAccessById(id);
+    return instancesRegister()->settingsAccessById(instanceId());
 }
 
 EffectInstanceId AbstractEffectModel::instanceId() const
 {
-    return m_instanceId.toULongLong();
+    const auto id = m_inited ? m_instanceId : BuiltinEffectViewLoader::initializationInstanceId();
```
kryksyh: If we have a guarantee that instance will be initialized at the object creation, why do we try to reinitialize it here?

src/effects/builtin/common/abstracteffectmodel.cpp
saintmatthieu: Good catch! `m_instanceId` can actually even be `const` now. Pushed.

kryksyh (approved):

chinakov:
Tested on Windows and Linux, all smoke-test scenarios have been passed. Thanks!

PR #9214, opened 2025-08-12
Noise reduction UI fixup

saintmatthieu:
Some commit addressing review comments from @kryksyh on the noise-reduction PR somehow got lost. Restoring these.

Recommended:

### DIFF
--- a/src/effects/builtin/noisereduction/NoiseReductionView.qml
+++ b/src/effects/builtin/noisereduction/NoiseReductionView.qml
@@ -36,6 +36,8 @@ EffectBase {
 
             RoundedRectangle {
                 color: ui.theme.backgroundSecondaryColor
+                border.color: ui.theme.strokeColor
+                border.width: 1
                 radius: 4
                 width: 211
                 height: getNoiseProfileBox.height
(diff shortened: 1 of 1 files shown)


src/effects/builtin/noisereduction/NoiseReductionView.qml
```diff
@@ -100,6 +104,17 @@ EffectBase {
                         font.bold: true
                     }
 
+                    StyledTextLabel {
+                        Layout.fillWidth: true
+                        Layout.leftMargin: 16
+                        Layout.rightMargin: 16
+                        Layout.bottomMargin: 16
+
+                        text: qsTrc("effects/noisereduction", "Select all of the audio you want filtered, choose how much noise you want filtered out, and then click “Apply” to reduce noise.")
```
luapmartin, line 113: this text is pure UI stuff right? as it's an helper text, so there is no need to have that closer to the ViewModel correct?

luapmartin (approved):

src/effects/builtin/noisereduction/NoiseReductionView.qml
saintmatthieu: I'd say so : this is only ever used here, and it's less boiler-plate code to have it written here directly.

PR #9240, opened 2025-08-15
8657 delete behavior prefs

### LINKED ISSUE #8593: Changing Clip Color doesn't make the project dirty
### Bug description

Changing Clip Color doesn't make the project dirty

### Steps to reproduce

1. Open Audacity
2. Create a Track
3. Generate or Import an audio clip
4. Save the Project
5. Change the clip color via clip context menu
6. Ctrl+W
7. Observe result

### Expected behavior

The project should be dirty

### Actual behavior

The project is not dirty

### Audacity Version

38887210637b4b0ef0ec81548318a6835ef7251c

### Operating system

OS independent

### Additional context

_No response_

### LINKED ISSUE #8657: Delete behaviour preferences
When a User deletes some audio as part of an audio data selection, the other canvas content can behave in numerous ways. 

Generally speaking, there are two different approaches. After deleting the selected audio we can either: 
- Close the gap
- Leave a gap

Given that deleting audio is central to the editing experience, we expose this preference during onboarding as well as in preferences.

[Figma link](https://www.figma.com/design/Gt4wJqWpinqjk6c1qE125O/19---Preferences?node-id=2597-6392)

Depends on task #8471

### LINKED ISSUE #8779: "Delete and close gap" isn't undoable
https://github.com/user-attachments/assets/f4a0b7e7-f7e0-48c6-9a07-2d128def4ceb

saintmatthieu:
Resolves: #8657
Resolves: #8779
Resolves: #8593

Recommended:

#### QA
Tip: to reset the delete behavior pref (or any other pref), go to the DevTools tab (beside the Home, Project and other such shell tabs). In the settings menu, search for deleteBehavior and set it to -1: 
image

Beside verifying that the spec is implemented:
- [ ] #8593: changing clip color makes project dirty and adds history item (should it?)


src/trackedit/internal/deletebehavioronboardingscenario.h
```diff
@@ -0,0 +1,28 @@
...
+ * Audacity: A Digital Audio Editor
+ */
+#pragma once
+
+#include "trackedit/trackedittypes.h"
+#include "trackedit/itrackeditconfiguration.h"
+
+#include "actions/iactionsdispatcher.h"
+#include "global/iinteractive.h"
+#include "modularity/ioc.h"
```
grliszas14: the order should be:
```c++
#include "modularity/ioc.h"
#include "global/iinteractive.h"
#include "actions/iactionsdispatcher.h"
#include "trackedit/itrackeditconfiguration.h"

#include "trackedit/trackedittypes.h"
```

src/trackedit/qml/Audacity/TrackEdit/StyledTextLabelWithCustomLinkActivatedSignal.qml
```diff
@@ -0,0 +1,66 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
```
grliszas14, line 3: This header should be enough:
```c++
/*
*/
```

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorOnboardingFollowupDialog.qml
```diff
@@ -0,0 +1,82 @@
...
+        id: column
+
+        width: parent.width
+
+        padding: 16
+        spacing: 7
+
+        StyledTextLabel {
+            id: titleLabel
+            width: parent.width - 32
```
grliszas14, line 42: 32 into a variable maybe? It's used twice, for the same reason

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorOnboardingFollowupDialog.qml
```diff
@@ -0,0 +1,82 @@
...
+
+    Column {
+        id: column
+
+        width: parent.width
+
+        padding: 16
+        spacing: 7
+
+        StyledTextLabel {
```
grliszas14, line 39: please group some of these properties and add spaces, it looks like not formatted (have a look at the entire file, not only this StyledTextLabel)

```qml
id: ..

width: ...
alignment: ...

others: ...
```

src/trackedit/internal/trackeditconfiguration.h
```diff
@@ -21,8 +21,18 @@ class TrackeditConfiguration : public ITrackeditConfiguration
     void setAskBeforeConvertingToMonoOrStereo(bool value) override;
     muse::async::Notification askBeforeConvertingToMonoOrStereoChanged() const override;
 
+    DeleteBehavior deleteBehavior() const override;
+    void setDeleteBehavior(DeleteBehavior value) override;
+    muse::async::Notification deleteBehaviorChanged() const override;
+
+    CloseGapBehavior closeGapBehavior() const override;
+    void setCloseGapBehavior(CloseGapBehavior value) override;
+    muse::async::Notification closeGapBehaviorChanged() const override;
```
grliszas14, line 30: configuration should contain only basic types, in this case int, conversion to DeleteBehavior should be done in the model

src/trackedit/itrackeditconfiguration.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "trackedittypes.h"
```
grliszas14, line 3: to be removed when you switch to integers

src/trackedit/internal/deletebehavioronboardingscenario.cpp
```diff
@@ -0,0 +1,56 @@
...
+    return true;
+}
+
+bool DeleteBehaviorOnboardingScenario::showOnboardingDialog() const
+{
+    muse::UriQuery uri("audacity://trackedit/delete_behavior");
+    uri.addParam("deleteBehavior", muse::Val(static_cast<int>(DeleteBehavior::LeaveGap)));
+    uri.addParam("closeGapBehavior", muse::Val(static_cast<int>(configuration()->closeGapBehavior())));
+    const muse::RetVal<muse::Val> rv = interactive()->openSync(uri);
+    if (rv.ret.code() != static_cast<int>(muse::Ret::Code::Ok)) {
```
grliszas14: have you tried simply?
```c++
if (!rv.ret.code()) {
```

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorPanel.qml
```diff
@@ -0,0 +1,115 @@
...
+                font.bold: true
+            }
+
+            RoundedRadioButton {
+                checked: root.closeGapBehavior === CloseGapBehavior.ClipRipple
+                text: qsTrc("trackedit/preferences", "The selected clip moves back to fill the gap")
+
+                navigation.name: "ClipMovesBackRadioBtn"
+                navigation.panel: root.navigation
+                // navigation.row: 3 What's that?
```
grliszas14: to be removed or uncommented with a proper value

if you set `navigation.row` for every single radio button, you'll be able to navigate through radio buttons with arrow keys

src/trackedit/qml/Audacity/TrackEdit/StyledTextLabelWithCustomLinkActivatedSignal.qml
saintmatthieu: Ah, I forgot to comment: this is most likely going to be a very provisional solution, hence I just copy-pasted from `StyledTextLabel.qml`. The only difference is the possibility of overriding the `onLinkActivated` callback.
@igorkorsukov / @eism I see two options: integrating this change in the framework or leave this file as it is. I of course would prefer the first option.

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorOnboardingFollowupDialog.qml
saintmatthieu: yup

src/trackedit/internal/trackeditconfiguration.h
saintmatthieu: It should? Why?

src/trackedit/internal/deletebehavioronboardingscenario.cpp
saintmatthieu: I haven't. It's more concise all right, but then the reader has to verify that `code()` returns an `int`. Here it's clear that it does.

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorPanel.qml
saintmatthieu: Yes yes, I have to address this TODO. Thanks for pointing it out.

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorOnboardingFollowupDialog.qml
saintmatthieu: I would if it got complex, but it's a small easy file.

src/trackedit/internal/deletebehavioronboardingscenario.h
saintmatthieu: Too many rules for me :) But you've been so kind to write it explicitly, and it compiles, so I'll apply.

src/trackedit/internal/deletebehavioronboardingscenario.cpp
kryksyh: even slicker would be 
```
if (!rv.ret.success()) {
```
or, as it has the appropriate overload, but more vague:
```
if (!rv.ret) {
```


there is no need to expose the underlying type of code()

saintmatthieu: `!rv.ret.success()` looks perfect.

src/trackedit/internal/trackeditconfiguration.h
grliszas14: it's because we shouldn't mix UI with services: DeleteBehavior or CloseGapBehavior are Q_GADGETS that belong to the UI layer

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorPanel.qml
```diff
@@ -24,6 +24,10 @@ Column {
     spacing: 16
     width: imageRow.width
 
+    Component.onCompleted: {
+        root.navigation.direction = NavigationPanel.Both
```
grliszas14, line 28: interesting place for this, why this has to be in Component.onCompleted?

DavidBailes:
@saintmatthieu . The animations fail WCAG 2.2.2: https://www.w3.org/TR/wcag2ict-22/#pause-stop-hide

(I find the continuously repeating animation very annoying.)

src/trackedit/view/deletebehaviorpanelmodel.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "trackedit/trackedittypes.h"
+
+#include <QObject>
```
grliszas14:
```suggestion
#include <QObject>

#include "trackedit/trackedittypes.h"
```

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorPanel.qml
```diff
@@ -0,0 +1,111 @@
...
+                checked: root.deleteBehavior === modelData.value
+
+                onToggled: {
+                    newDeleteBehaviorRequested(modelData.value)
+                }
+
+                navigation.name: modelData.text
+                navigation.panel: root.navigation
+                navigation.row: 0
+                navigation.column: model.index
```
grliszas14:
```suggestion
                navigation.name: modelData.text
                navigation.panel: root.navigation
                navigation.row: 0
                navigation.column: model.index
                
                onToggled: {
                    newDeleteBehaviorRequested(modelData.value)
                }
```

src/trackedit/qml/Audacity/TrackEdit/DeleteBehaviorPanel.qml
saintmatthieu: I tried, but the script fails loading, then. I guess because of initialization order? `navigation` is set from a parent, which is initialized after this component.

Tantacrul:
> (I find the continuously repeating animation very annoying.)

I don't really agree. There's no better way for a one-time animation to work. Moreover, the accessibility rule that video shouldn't repeatedly playback is more intended for websites with videos, not for gifs like this.

The gifs help reinforce what is (if only communicated via text) a tricky concept to communicate. We've done our best with text, but believe that a decent portion of users won't understand what we're asking otherwise.

Tantacrul:
Also, on MacOS, I'm not seeing this follow-up question:
image

@DilsonsPickles We also need to choose a different colour for hyperlinks. I'm pretty sure those colours don't pass minimum contrast thresholds.

I've also included a rough mockup of the clips to make them easier to see. They're more zoomed in. Have higher contrast and a stroke around the edges so they don't bleed into the UI.
image

grliszas14 (approved):

DavidBailes:
> Moreover, the accessibility rule that video shouldn't repeatedly playback is more intended for websites with videos, not for gifs like this.

This isn't the view of the section of the document I linked to (Guidance on Applying WCAG 2 to Non-Web Information and Communications Technologies). 
It's a level A criterion, which this case fails.

Tantacrul:
Yes, it's a level A criterion because they have designed the rules with a view to stopping blatant, persistent animations on websites that cannot be dismissed and distract people from their primary focus. For example endless animations on a banner advert. 

This dialogue asks you to make a binary choice and presents tiny animations to help you make that choice. Then it's gone. If it's technically breaking the letter of the law, I can live with that. A simple solution to make everyone happy is to include pause/play buttons on each animation, although this will be included in a later PR.

Issue #8891, opened 2025-06-03
Implement track focus state keyboard nav


grliszas14:
### Description

- Pressing 'up' / 'down' moves the selection state between tracks
- Holding 'shift' while holding up/down allows the user to multi-select tracks
- Pressing 'enter' should select/deselect the track
- When 'selecting' a track using our new accessibility system, the blue cursor should highlight the selected track accordingly.

From @Tantacrul:
@DilsonsPickles @LWinterberg  - we'll need to have a discussion about this on our next design call to discuss. Once we're reasonably happy, we'll also need to bring in DB to show him what we've done. I'd suggest a call with him.

[Figma link](https://www.figma.com/design/oQEvoOqeKHQlqabmFmM0Mp/10---Selection?node-id=1169-24820&t=23yBLuEUckO1gQ76-4)

DilsonsPickles:
Added Figma link into task description.

@Tantacrul - I've made the changes we discussed during our call. I'll create a separate ticket to handle the new track selection visualisation we wanted.

@grliszas14 - The spec I've attached details the AU3 behaviour that we need to implement. It also contains one additional behaviour whereby the User can hold shift whilst selecting a new track with the track focus to create a ranged track selection. I.e. selecting all tracks between Mono track 1 and 4.

Please let me know if you have any questions!

embarc-gabriel:
This ticket must also implement what is missing on #8893

embarc-gabriel:
Everything was implemented on the sub tasks

PR #9243, opened 2025-08-18
8893 input monitoring

### LINKED ISSUE #8893: Input monitoring and metering
### Description

Exposing the "Enable audible input monitoring" in the microphone level flyout and allowing Users to monitor their input levels via the track header's meters.

Image

[Figma link](https://www.figma.com/design/vpweiC0Dfy4wFgrNrvfTTD/03---Playback-meter?node-id=1103-11142)

embarc-gabriel:
Resolves: #8893 

This PR implements most of the design documented here: https://www.figma.com/design/vpweiC0Dfy4wFgrNrvfTTD/03---Playback-meter?node-id=1103-11142&p=f&t=HVoraiaXxjcbQR5O-0

The items "Moving the focus away from the selected track" and "Input monitoring with multiple track selected" were not implemented because they depend on https://github.com/audacity/audacity/issues/8891

After this PR the input will be monitored by the track meter if the track is on focus and the track number of channels matches the input device number of channels. It is equivalent to an always enabled  "silence monitoring" option on au3. 
Also the input popup adds the option "Enable input monitoring" keeping the same behavior as in au3 in this respect. This same options can also be enabled/disabled on the preference section.

Recommended:


au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -606,7 +607,7 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions& options,
         }
       #endif
 
-        if (mSoftwarePlaythrough) {
+        if (mSoftwarePlaythrough && !isMME) {
```
embarc-gabriel, line 610: Even setting playbackParameters with a reasonable suggestedLatency for MME on Windows PortAudio opens the stream with a very high latency as we can check on Pa_GetStreamInfo. This is a workaround for sure and we could open a follow up ticket.

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -650,7 +652,7 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions& options,
         captureParameters.hostApiSpecificStreamInfo = NULL;
         captureParameters.channelCount = mNumCaptureChannels;
 
-        if (mSoftwarePlaythrough) {
+        if (mSoftwarePlaythrough && !isMME) {
```
embarc-gabriel, line 655: Same as in playbackParameters.

src/projectscene/view/trackspanel/trackslistmodel.cpp
```diff
@@ -627,7 +627,10 @@ void TracksListModel::onTrackRemoved(const trackedit::Track& track)
     for (int i = 0; i < m_trackList.size(); ++i) {
         if (m_trackList.at(i)->trackId() == track.id) {
             beginRemoveRows(QModelIndex(), i, i);
-            m_trackList.erase(m_trackList.begin() + i);
+            const auto it = m_trackList.begin() + i;
```
embarc-gabriel, line 630: Memory leak I've found during tests.

src/record/internal/recordconfiguration.cpp
```diff
@@ -3,10 +3,39 @@
...
+        m_micMeteringChanged.notify();
+    });
+}
+
 draw::Color RecordConfiguration::recordColor() const
 {
     return "#EF476F";
 }
+
+bool RecordConfiguration::micMetering() const
```
embarc-gabriel: Mic metering config was added to the settings but the input monitoring is still on gPrefs.

au3/libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: I think this change doesn't entail any risk, so I don't mind keeping it this way. I wonder why it is different in AU3, though, which doesn't have this problem.

src/record/qml/Audacity/Record/RecordLevel.qml
```diff
@@ -165,10 +191,49 @@ FlatButton {
             StyledTextLabel {
                 Layout.fillWidth: true
 
-                text: qsTrc("record", "Adjust the level of your OS configured microphone")
+                text: qsTrc("record", "Note: this control is tied to your computer’s main mic volume")
                 horizontalAlignment: Text.AlignLeft
                 wrapMode: Text.WordWrap
             }
+
+            Rectangle {
```
saintmatthieu, line 216: Why not use a SeparatorLine here? It also indeed looks 2px thick, while it should only be 1.

src/record/qml/Audacity/Record/RecordLevel.qml
```diff
@@ -67,17 +73,31 @@ FlatButton {
...
         id: meterModel
+
+        meterInputSource: MeterInputSource.Record
     }
 
     StyledPopupView {
         id: popup
 
-        margins: 12
+        margins: viewModel.margins
```
saintmatthieu, line 106: Not sure why but the dialog has a bit too much space at the bottom, maybe 28 instead of 16px. Maybe there's a way of specifying left, right and top margins to 12 and leaving the bottom one to 0 ?

saintmatthieu: ... or personally, I'd have used a `Column` rather than a `ColumnLayout`: it's simpler and since this dialog certainly doesn't have to be resizable, Column would suffice.

src/record/irecordconfiguration.h
```diff
@@ -16,5 +16,9 @@ class IRecordConfiguration : MODULE_EXPORT_INTERFACE
     virtual ~IRecordConfiguration() = default;
 
     virtual muse::draw::Color recordColor() const = 0;
+
+    virtual bool micMetering() const = 0;
```
saintmatthieu: Minor: looking at the signature it's obvious that this means "is mic metering on", yet just the name "mic metering" might suggest a struct return type or smth. `micMeteringOn` ? `isMicMeteringOn` ?

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
```diff
@@ -284,10 +294,7 @@ ListItemBlank {
                 TapHandler {
```
saintmatthieu, line 294: What's a tap handler ?

src/playback/view/toolbars/playbacktoolbarlevelitem.cpp
```diff
@@ -43,6 +43,11 @@ PlaybackToolBarLevelItem::PlaybackToolBarLevelItem(const muse::ui::UiAction& act
         setAudioChannelRMS(audioChNum, meterSignal.rms.pressure);
     });
 
+    playback()->audioOutput()->playbackVolume().onResolve(this, [this](float volume) {
```
saintmatthieu, line 46: Why did this registration have to move up ?

src/playback/view/panels/playbackmeterpanelmodel.h
```diff
@@ -22,6 +23,7 @@ class PlaybackMeterPanelModel : public QObject, public muse::async::Asyncable
     muse::Inject<IPlayback> playback;
     muse::Inject<IPlaybackConfiguration> configuration;
     muse::Inject<IPlaybackController> controller;
+    muse::Inject<record::IRecord> record;
```
saintmatthieu: I don't see `record` being used in the source file.

src/playback/view/toolbars/playbacktoolbarlevelitem.cpp
saintmatthieu: (Same question about playbackmeterpanelmodel.cpp)

src/record/qml/Audacity/Record/RecordLevel.qml
```diff
@@ -67,17 +73,31 @@ FlatButton {
...
+        readonly property int margins: 12
+        readonly property int spacing: 8
+
+        readonly property int meterHeight: 50
+    }
+
     PlaybackMeterModel {
         id: meterModel
+
+        meterInputSource: MeterInputSource.Record
```
saintmatthieu: Looking at this, it took me a while to realize that `meterInputSource` was defaulting to `MeterInputSource.Playback`. How about adding a `NotSet = -1` entry to the enum, have `PlaybackMeterModel` inherit from `QQmlParserStatus` and, in the `componentComplete` implementation, assert that it was set ?

src/playback/view/common/playbackmetermodel.cpp
```diff
@@ -45,22 +46,44 @@ PlaybackMeterModel::PlaybackMeterModel(QObject* parent)
...
-        emit meterSizeChanged();
-    });
-
     configuration()->playbackMeterDbRangeChanged().onNotify(this, [this]() {
         emit meterDbRangeChanged();
         emit dbRangeChanged();
         emit positionChanged();
     });
 
     playback()->audioOutput()->playbackVolumeChanged().onReceive(this, [this](audio::volume_dbfs_t volume) {
```
saintmatthieu, line 23: I remember @Eism stating that those subscriptions should not occur in the constructor but in an `init` method (can't remember in which case it otherwise can cause problems).
I suppose that, if you let this model inherit from QQmlParserStatus, like I suggested elsewhere, you could implement the `componentCompleted` method and do those registrations there? At that stage, you'd even know if `m_meterInputSource` is playback or recording, and we could have more efficient code:

```cpp
if (m_meterInputSource == MeterInputSource::Source::Playback) {
    playback()->audioOutput()->playbackVolumeChanged().onReceive(this, [this](audio::volume_dbfs_t volume) {
        m_volume = volume;
        emit positionChanged();
    });
}
```
Since playback volume changes fast and continuously, the performance gain might not be negligible.

saintmatthieu: Maybe better, even: you could create a `MeterModel` base class, a `PlaybackMeterModel` and a `RecordingMeterModel`, and then you wouldn't need `meterInputSource`, you could register to the events each needs, and you could even eliminate the `MeterInputSource` Q_GADGET.

src/projectscene/view/trackspanel/trackitem.cpp
```diff
@@ -37,27 +37,44 @@ TrackItem::TrackItem(QObject* parent)
...
         setAudioChannelVolumePressure(channel, meterSignal.peak.pressure);
         setAudioChannelRMS(channel, meterSignal.rms.pressure);
     });
 
+    audioDevicesProvider()->inputChannelsChanged().onNotify(this, [this]() {
+        m_inputChannelsCount = audioDevicesProvider()->currentInputChannelsCount();
+    });
+    m_inputChannelsCount = audioDevicesProvider()->currentInputChannelsCount();
+
+    record()->audioInput()->recordSignalChanges().onReceive(this, [this](const audioch_t audioChNum, const audio::MeterSignal& meterSignal) {
```
saintmatthieu: I wonder if it wouldn't be worthwhile to just register for these callbacks when they are needed? I'm debugging right now: my playback is stopped, no meter is active anywhere, yet I receive this call all the time at record-buffer rate. What if we did
```cpp
    audioDevicesProvider()->inputChannelsChanged().onNotify(this, [this]() {
        const int count = audioDevicesProvider()->currentInputChannelsCount();
        m_channelsMatch = (m_trackType == trackedit::TrackType::Mono && count == 1)
            && (m_trackType == trackedit::TrackType::Stereo && count == 2);
        checkCallbacks();
    });

void TrackItem::setIsFocused(bool value) {
    // ...
    m_isFocused = value;
    emit isFocusedChanged();
    checkCallbacks();
}

void TrackItem::checkCallbacks() {
    if (m_isFocused && m_channelsMatch) {
        record()->audioInput()->recordSignalChanges().onReceive(...)
    } else {
        record()->audioInput()->recordSignalChanges().resetOnReceive(this);
    }
}
```
?
The performance gain would be proportional to the number of tracks in the project, too.

saintmatthieu (commented):
Looks good mostly. A few minor suggestions and a bigger optimization suggestion which, if the suggestion makes sense, I'd like to see addressed before merging.

src/playback/view/toolbars/playbacktoolbarlevelitem.cpp
embarc-gabriel: I just grouped similar code.

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
embarc-gabriel: It is something like a MouseArea, but only for click and tap events. As far as I know, it is supposed to be lightweight and more straightforward than using a MouseArea itself.

src/record/qml/Audacity/Record/RecordLevel.qml
embarc-gabriel: Fixed.

src/record/irecordconfiguration.h
embarc-gabriel: Fixed.

src/record/qml/Audacity/Record/RecordLevel.qml
embarc-gabriel: Fixed.

src/playback/view/panels/playbackmeterpanelmodel.h
embarc-gabriel: Fixed.

src/record/qml/Audacity/Record/RecordLevel.qml
embarc-gabriel: Removed this type as suggested in another comment.

src/playback/view/common/playbackmetermodel.cpp
embarc-gabriel: Done.

src/projectscene/view/trackspanel/trackitem.cpp
embarc-gabriel: Much better now.
In fact, I had also done the same for the RecordLevel popup. You can only listen to the record signal if the pop-up is opened.

src/projectscene/view/trackspanel/trackitem.h
```diff
@@ -147,6 +150,7 @@ public slots:
 
     bool m_isSelected = false;
     bool m_isFocused = false;
+    int m_recordStreamChannelsMatch = 0;
```
saintmatthieu: Shouldn't this be a `bool` ?

src/playback/view/common/metermodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+}
+
+MeterModel::MeterModel(QObject* parent)
+    : QObject(parent)
+{
+}
+
+void MeterModel::init()
+{
+    meterController()->playbackMeterChanged().onNotify(this, [this]() {
```
saintmatthieu, line 31: Would it be ok in this PR to rename `playbackMeterChanged` to maybe `volumeMeterChanged` ? If it's impractal, sometime soon would be good.

saintmatthieu (approved):
Thanks a lot! If you could just address this renaming thing. Feel free to move it to QA already, though.

src/playback/view/common/metermodel.cpp
embarc-gabriel: I don´t get it. There is no volume information on the meter controller
.

src/projectscene/view/trackspanel/trackitem.h
embarc-gabriel: Yes! Fixed.

Issue #9174, opened 2025-07-31
Implement custom FFmpeg export


ianalarina:
### Description

[Figma link](https://www.figma.com/design/QaCzd2bM08W8VWgCLsKJ2u/32---Export?node-id=1750-32197&t=lWQwPYOmXWdRvVVJ-0)

chinakov:
Issue noticed:
- When the Custom FFmpeg Export option is selected as the format, the 'Export' and 'Cancel' buttons become misaligned. 
Without Custom FFmpeg Export option:

<img width="532" height="522" alt="Image" src="https://github.com/user-attachments/assets/5745a2e4-7d5a-4b19-86ed-3f72eb547416" />

With Custom FFmpeg Export option:

<img width="532" height="632" alt="Image" src="https://github.com/user-attachments/assets/2ad679f9-c8ae-4e96-9327-a123e7d65007" />

chinakov:
This is done, thanks!

PR #9249, opened 2025-08-20
Custom ffmpeg dialog

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9174

Recommended:


src/importexport/export/iffmpegoptionsaccessor.h
```diff
@@ -0,0 +1,29 @@
...
+    INTERFACE_ID(IFFmpegOptionsAccessor)
+
+public:
+    virtual ~IFFmpegOptionsAccessor() = default;
+
+    virtual void init() = 0;
+
+    virtual std::vector<std::string> formatList() const = 0;
+    virtual void fetchCompatibleFormatList(const std::string& format, const std::string& codec) = 0;
+    virtual void showAllFormats() = 0;
```
saintmatthieu: Why not `fetchAllFormats` ? I was puzzled by the term "show" in an interface that has no UI purpose. (Applies to `showAllCodecs`, too.)

src/importexport/export/view/exportpreferencesmodel.h
```diff
@@ -79,9 +84,13 @@ class ExportPreferencesModel : public QObject, public muse::async::Asyncable
     QVariantList exportSampleFormatList() const;
     Q_INVOKABLE void setExportSampleFormat(const QString& format);
 
+    Q_INVOKABLE void openCustomFFmpegDialog();
     Q_INVOKABLE bool verifyExportPossible();
     Q_INVOKABLE void exportData();
 
+    // dynamic inputs
+    Q_INVOKABLE bool customFFmpegOptionsVisible();
```
saintmatthieu: Do you need the `Q_INVOKABLE` macro if this method is already declared as getter of the customFFmpegOptionsVisible property?

src/importexport/export/view/exportpreferencesmodel.cpp
```diff
@@ -297,3 +303,8 @@ void ExportPreferencesModel::exportData()
         interactive()->error(muse::trc("export", "Export error"), result.text());
     }
 }
+
+bool ExportPreferencesModel::customFFmpegOptionsVisible()
+{
+    return currentFormat() == "Custom FFmpeg Export";
```
saintmatthieu: Better would be to iterate through `ExportFFmpegOptions::fmts` to find the one with `fmtid == FFmpegExposedFormat::FMT_OTHER`.

src/importexport/export/view/customffmpegpreferencesmodel.cpp
```diff
@@ -0,0 +1,528 @@
...
+void CustomFFmpegPreferencesModel::showAllFormats()
+{
+    ffmpegOptionsAccessor()->showAllFormats();
+    emit ffmpegFormatListChanged();
+    emit ffmpegFormatIndexChanged();
+}
+
+int CustomFFmpegPreferencesModel::ffmpegCodecIndex() const
+{
+    const auto& codecs = ffmpegOptionsAccessor()->codecList();
```
saintmatthieu: `ffmpegOptionsAccessor()->codecList()` doesn't return a reference.

src/importexport/export/view/customffmpegpreferencesmodel.cpp
```diff
@@ -0,0 +1,528 @@
...
+    for (int i = 0; i < static_cast<int>(codecs.size()); ++i) {
+        if (codecs[i] == exportConfiguration()->ffmpegCodec()) {
```
saintmatthieu, line 170: Nitpick: since the `[]` operator of an std::vector takes a size_t, probably a `for (size_t i = 0; ...)` would be less likely to issue a warning. Just then static_cast the returned `i` to an int.

src/importexport/export/view/customffmpegpreferencesmodel.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+    return -1;
+}
+
+QString CustomFFmpegPreferencesModel::ffmpegFormat() const
+{
+    QString format = QString::fromStdString(exportConfiguration()->ffmpegFormat());
+    if (format.isEmpty()
+        || !muse::contains(ffmpegOptionsAccessor()->formatList(), exportConfiguration()->ffmpegFormat())) {
+        format = ffmpegFormatList().front().toString();
```
saintmatthieu: Here I'd also check that `ffmpegFormatList()`'s return isn't empty before calling `front()`: if for some reason `FetchFormatList()` wasn't called on `ExportFFmpegOptions`, then this would return an empty list, AFAICS.

src/importexport/export/view/customffmpegpreferencesmodel.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+    return -1;
+}
+
+QString CustomFFmpegPreferencesModel::ffmpegCodec() const
+{
+    QString codec = QString::fromStdString(exportConfiguration()->ffmpegCodec());
+    if (codec.isEmpty()
+        || !muse::contains(ffmpegOptionsAccessor()->codecList(), exportConfiguration()->ffmpegCodec())) {
+        codec = ffmpegFormatList().front().toString();
```
saintmatthieu: Did you mean `ffmpegCodecList()` ?

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.h
```diff
@@ -72,12 +68,37 @@ struct ExposedFormat
...
-//     ~ExportFFmpegOptions();
+    //     /// Retrieves format list from libavformat
+    void FetchFormatList();
+    std::vector<std::string> GetFormatNames() const;
+
+    //     /// Retrieves a list of formats compatible to codec
+    //     ///\param id Codec ID
+    //     ///\param selfmt format selected at the moment
+    //     ///\return index of the selfmt in NEW format list or -1 if it is not in the list
+    int FetchCompatibleFormatList(AudacityAVCodecID id, wxString* selfmt);
```
saintmatthieu: Looks like `selfmt` can be const, like it is below for `FetchCompatibleCodecList`.
Minor: I'd keep the arguments in the same order as in `FetchCompatibleCodecList`.

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.cpp
```diff
@@ -789,222 +780,205 @@ const TranslatableStrings PredictionOrderMethodNames {
...
+}
 
-//     // Get selected codec short name
-//     wxString selcdc = mCodecList->GetString(selections[0]);
+void ExportFFmpegOptions::FetchCompatibleCodecList(const std::string& format,
+                                                   const std::string& codec)
+{
+    auto fmt = mFFmpeg->GuessOutputFormat(wxString(format).ToUTF8(), NULL, NULL);
+    if (fmt == NULL) {
+        // This shouldn't really happen
```
saintmatthieu, line 877: Then `assert(false)` ?

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.cpp
```diff
@@ -789,222 +780,205 @@ const TranslatableStrings PredictionOrderMethodNames {
...
+        return;
+    }
 
-//     // Find its index
-//     int nCodec = make_iterator_range(mCodecNames).index(selcdc);
-//     if (nCodec == wxNOT_FOUND) {
-//         return;
-//     }
+    AudacityAVCodecID selcdcid = AUDACITY_AV_CODEC_ID_NONE;
```
saintmatthieu, line 882: Stands for "selected ID" ? `:D`

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.cpp
```diff
@@ -789,222 +780,205 @@ const TranslatableStrings PredictionOrderMethodNames {
...
-
-//         if (format != nullptr) {
-//             auto codec = mFFmpeg->CreateEncoder(format->GetAudioCodec());
-
-//             if (
-//                 codec != nullptr && codec->IsAudio() && mFFmpeg->av_codec_is_encoder(codec->GetWrappedValue())) {
-//                 if ((ffmpegId >= 0) && codec->GetId() == ffmpegId) {
-//                     index = mShownCodecNames.size();
-//                 }
+    std::unique_ptr<AVCodecWrapper> cdc;
```
saintmatthieu: What's the use of this variable?

src/importexport/export/internal/au3/au3ffmpegoptionsaccessor.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "modules/import-export/mod-ffmpeg/ExportFFmpegOptions.h"
+
+#include "au3ffmpegoptionsaccessor.h"
+
+using namespace au::importexport;
+
+void Au3FFmpegOptionsAccessor::init()
+{
+    m_options.FetchFormatList();
```
saintmatthieu: Couldn't these two methods get called from the ctor of `m_options` ? Then it'd be more robust (less risk that `formatList()` returns an empty vector) and you woudn't need this `init()` anymore.

saintmatthieu (commented):
A mixture of more or less important comments. (I haven't reviewed the QML part, which @kryksyh is reviewing, so please wait for both approvals before proceeding.)

The thing that we have to clarify is about this modularity. Au3's FFMPEG is a module, meaning we can build Au3 without it, and the app will otherwise function normally.
Here it is not the case: I verified that, setting `AU_USE_FFMPEG` to `OFF` yields successful compilation but the export dialog cannot be loaded anymore.

Maybe we don't need modularity for FFMPEG, in which case this PR could be simplified. If we do, however, I recommend that we put it in place now: postponing it can only make it more work later.

src/importexport/export/view/customffmpegpreferencesmodel.cpp
```diff
@@ -0,0 +1,528 @@
...
+    if (val == frameSize()) {
+        return;
+    }
+
+    exportConfiguration()->setFFmpegFrameSize(val);
+}
+
+QString CustomFFmpegPreferencesModel::pdOMethod() const
+{
+    int methodIndex = exportConfiguration()->ffmpegPredictionOrderMethod();
```
kryksyh, line 467: since it is pulled from configuration the value could be anything, it is better to clamp it known value, or reset to initial state, otherwise `.at(methodIndex)` would throw

saintmatthieu:
As discussed in our daily: we do not care that FFMPEG support can be removed from the Audacity 4 binary (the important thing for FFMPEG being that the absence of FFMPEG dynamic libraries should not get Audacity to misbehave). Hence @grliszas14 the #ifdefs can be removed. In the unlikely event that we need this kind of modularity, we'll have to do it properly then.

src/importexport/export/view/customffmpegpreferencesmodel.cpp
grliszas14: oooh, that's a good catch

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.h
grliszas14: not my idea, it's like this in AU3, but fixed anyway :)

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.cpp
grliszas14: I think it's "selected codec ID" `:D` again: it's AU3's code, I don't know if there's any benefit from changing this, at least it's in a single naming convention across all import/export files `:P`

au3/modules/import-export/mod-ffmpeg/ExportFFmpegOptions.cpp
grliszas14: leftover, removed

src/importexport/export/internal/au3/au3ffmpegoptionsaccessor.cpp
grliszas14: yes!

src/importexport/export/view/exportpreferencesmodel.cpp
grliszas14: well... it would but there's the catch: exportpreferencesmodel doesn't know a thing about AU3 code (and we don't want it to) so we cannot just include ExportFFmpegOptions and iterate throught fmts as it's only available within the ExportFFmpegOptions.cpp/h internally.

kryksyh: Also, in AU3, this is a translatable string

src/importexport/export/qml/Export/internal/FLACOptionsSection.qml
```diff
@@ -0,0 +1,269 @@
...
+
+                StyledDropdown {
+                    width: root.controlWidth
+
+                    model: ffmpegPrefModel.pdOMethodList
+
+                    currentIndex: indexOfValue(ffmpegPrefModel.pdOMethod)
+
+                    onActivated: function(index, value) {
+                        ffmpegPrefModel.setPdOMethod(value)
```
kryksyh: wouldnt be `setPdOMethodIndex` simpler to implement here?

src/importexport/export/view/exportpreferencesmodel.cpp
saintmatthieu: I see.
Could `IExporter` have a `isCustomFfmpegExportFormat(const std::string& format) const` method? `Au3Exporter` would then do the dirty job.

src/importexport/export/view/customffmpegpreferencesmodel.cpp
```diff
@@ -0,0 +1,532 @@
...
+#include "customffmpegpreferencesmodel.h"
+
+#include "translation.h"
+
+using namespace au::importexport;
+
+CustomFFmpegPreferencesModel::CustomFFmpegPreferencesModel(QObject* parent)
+    : QObject(parent)
+{
+    init();
```
saintmatthieu: Just noticing this now, sorry: @Eism once warned me that init method (with all the `onNotify` and `onReceived` subscriptions) should be called from `Component.onCompleted`, that these ctor inits caused problems in the past in MSS, can't remember which.
It's also called from the ctor in `exportpreferencesmodel.cpp` (but it was already like that).
If you want to keep it in the ctor, since I can't remember why it's bad ( :D ) I'll be okay with it, but then the `init` method doesn't need to be Q_INVOKABLE anymore and could be private.

grliszas14: yep, fixed for both models

saintmatthieu (approved):

Issue #9251, opened 2025-08-22
Audacity is slow to quit


chinakov:
### Bug description

If the Audacity is closed via Quit, the closing process takes a longer time than usual. In case of macOS, it crashes. 

### Steps to reproduce

1. Launch Audacity 4
2. Open a project
3. Close the project
4. Observe result

### Expected behavior

Audacity should close faster 

### Actual behavior

Audacity is slow to close

### Audacity Version

d05bafa3263adef40bc2a0e895bae3e8554b4daf

### Operating system

OS independent

dozzzzer:
Tested

Issue #9245, opened 2025-08-19
UI flickers when window is moved


chinakov:
### Bug description

If the User moves the window of AU4, the UI will flicker. This will also happen during resize. 

### Steps to reproduce

1. Open Audacity 
2. Resize or move window
3. Observe result

### Expected behavior

The UI should not flicker

### Actual behavior

The UI flickers 

### Audacity Version

c5b56f6eecf11570a4ab205310d6e5dc69cc61c7

### Operating system

Windows

### Additional context

https://github.com/user-attachments/assets/17502025-2b79-4ad4-850d-fe2687d7dbb7

chinakov:
This is fixed, thanks!

PR #9255, opened 2025-08-25
Qt 6.9

kryksyh:
Update Qt to 6.9

Recommended:


chinakov:
Should fix https://github.com/audacity/audacity/issues/9251 and https://github.com/audacity/audacity/issues/9245

.github/workflows/au4_build_linux.yml
```diff
@@ -90,6 +90,16 @@ jobs:
       uses: turtlebrowser/get-conan@main
     - name: Register gcc problem matcher
       run: echo "::add-matcher::.github/problem_matchers/gcc.json"
+
+    - name: Install Qt
+      uses: jurplel/install-qt-action@v4
+      with:
+        version: '6.9.1'
```
cbjeukendrup: 6.9.2 has been released this week, so you might want to start with that version immediately

src/playback/tests/CMakeLists.txt
```diff
@@ -20,6 +20,8 @@ set(MODULE_TEST_LINK
 
     muse::uicomponents
     muse::ui
+
+    Qt::Quick
```
cbjeukendrup, line 25: Is it `playback_tests` that needs to link to Qt::Quick, or is it `playback`?

(Also, at MU, everything links to `global`, and `global` publicly links to `Qt::Quick` because of PCH. So at MU, something like this is not necessary.)

src/appshell/qml/MainToolBar.qml
```diff
@@ -70,9 +70,6 @@ Item {
 
         model: toolBarModel
 
-        width: contentItem.childrenRect.width
-        height: contentItem.childrenRect.height
```
cbjeukendrup, line 74: If you ever encounter similar issues elsewhere, where just removing these lines doesn't help, see also here: https://github.com/musescore/MuseScore/pull/24326 (a community member figured it out for us)

.github/workflows/au4_build_linux.yml
kryksyh: Will try :)

src/playback/tests/CMakeLists.txt
kryksyh: That was my first thought.
`timecodemodel_tests.cpp` includes `timecode.h` which includes QQuickItem, so it is a test.

Adding all generic modules to `global` makes sense. How does it affect compile time?

src/app/main.cpp
```diff
@@ -139,6 +145,73 @@ int main(int argc, char** argv)
...
+#endif
+
+    //! NOTE: For unknown reasons, Linux scaling for 1 is defined as 1.003 in fractional scaling.
+    //!       Because of this, some elements are drawn with a shift on the score.
+    //!       Let's make a Linux hack and round values above 0.75(see RoundPreferFloor)
+#ifdef Q_OS_LINUX
+    QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor);
+#elif defined(Q_OS_WIN)
+    QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::PassThrough);
+#endif
```
cbjeukendrup: This code was removed from MU in https://github.com/musescore/MuseScore/commit/0728fc76f18e10a5b176c5a29e7c8a05cf853ec2 and https://github.com/musescore/MuseScore/commit/dc5341cd22870f323fd863498170e2580ab08cb1

.github/workflows/au4_build_macos.yml
```diff
@@ -86,6 +86,15 @@ jobs:
...
         echo "SENTRY_URL: $SENTRY_URL"
 
+    - name: Install Qt
+      uses: jurplel/install-qt-action@v4
+      with:
+        version: '6.9.2'
+        host: 'mac'
+        target: 'desktop'
+        arch: 'clang_64'
+        modules: 'qt5compat qtnetworkauth qtscxml qtshadertools qtwebsockets'
```
cbjeukendrup: You might want to check if `qtscxml` is necessary for Audacity. In MU, it's only needed because one legacy widget uses Qt StateMachine, and therefore I doubt Audacity needs it.

src/playback/tests/CMakeLists.txt
cbjeukendrup: > How does it affect compile time?

Hm, I haven't investigated that, sorry

PR #9264, opened 2025-08-26
Focused track change and selection using arrow keys.

### LINKED ISSUE #9263: Focused track change and selection using arrow keys.
### Description

Allow the user to change the focused track using the up and down keys.
Add or remove the focused track from the selected tracks list using the enter key.

embarc-gabriel:
Resolves: #9263 

This PR allows the user to change the focused track using the up and down arrow keys.
Also, the user can add or remove the track from the list of selected tracks using the arrow key.

Recommended:

### DIFF
--- a/src/app/configs/data/shortcuts.xml
+++ b/src/app/configs/data/shortcuts.xml
@@ -58,6 +58,11 @@
         <seq>Left</seq>
     </SC>
     <SC>
+        <key>track-toggle-focused-selection</key>
+        <seq>Return</seq>
+        <seq>Enter</seq>
+    </SC>
+    <SC>
         <key>nav-up</key>
         <seq>Up</seq>
     </SC>
--- a/src/app/configs/data/shortcuts.xml
+++ b/src/app/configs/data/shortcuts.xml
@@ -467,16 +467,16 @@
         <seq>Ctrl+Shift+F6</seq>
     </SC>
     <SC>
-        <key>next-frame</key>
-        <seq>Ctrl+F6</seq>
-    </SC>
-    <SC>
         <key>focus-prev-track</key>
         <seq>Up</seq>
     </SC>
     <SC>
         <key>focus-next-track</key>
         <seq>Down</seq>
+    </SC>
+    <SC>
+        <key>next-frame</key>
+        <seq>Ctrl+F6</seq>
     </SC>
     <SC>
         <key>prev-track</key>
--- a/src/app/configs/data/shortcuts.xml
+++ b/src/app/configs/data/shortcuts.xml
@@ -471,6 +471,14 @@
         <seq>Ctrl+F6</seq>
     </SC>
     <SC>
+        <key>focus-prev-track</key>
+        <seq>Up</seq>
+    </SC>
+    <SC>
+        <key>focus-next-track</key>
+        <seq>Down</seq>
+    </SC>
+    <SC>
         <key>prev-track</key>
         <seq>Up</seq>
     </SC>
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
@@ -295,8 +295,7 @@ Item {
 
                 navigation.name: Boolean(clipItem) ? clipItem.title + clipItem.index : ""
                 navigation.panel: root.navigationPanel
-                navigation.column: index
-                navigation.row: root.trackIdx
+                navigation.column: clipItem ? Math.floor(clipItem.x) : 0
                 navigation.accessible.name: Boolean(clipItem) ? clipItem.title : ""
                 navigation.onActiveChanged: {
                     if (navigation.active) {
--- a/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
@@ -63,19 +63,12 @@ ListItemBlank {
         }
     }
 
-    property NavigationPanel navigationPanel: NavigationPanel {
-        name: "Track" + root.item.title + "Panel"
-        enabled: root.enabled && root.visible
-        direction: NavigationPanel.Horizontal
-        onActiveChanged: function(active) {
-            if (active) {
-                root.forceActiveFocus()
-            }
-        }
-    }
-
     height: trackViewState.trackHeight
     opacity: dragged ? 0.5 : 1
+
+    focusBorder.anchors.leftMargin: spacer.width + 2
+    focusBorder.anchors.rightMargin: volumePressureContainer.width + separatorLine.width + 10
+    focusBorder.anchors.bottomMargin: 2
 
     background.color: (root.isSelected || hoverHandler.hovered) ?
                    ui.theme.backgroundPrimaryColor : ui.theme.backgroundSecondaryColor
@@ -271,6 +264,7 @@ ListItemBlank {
         }
 
         SeparatorLine {
+            id: separatorLine
             Layout.bottomMargin: 2 * bottomSeparator.thickness
         }
 
@@ -367,6 +361,7 @@ ListItemBlank {
                 if(!root.isSelected) {
                     root.selectionRequested(false)
                 }
+                navigation.requestActive(true)
             }
         }
 
--- a/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
@@ -67,7 +67,7 @@ ListItemBlank {
     opacity: dragged ? 0.5 : 1
 
     focusBorder.anchors.leftMargin: spacer.width + 2
-    focusBorder.anchors.rightMargin: volumePressureContainer.width + separatorLine.width + 10
+    focusBorder.anchors.rightMargin: 24 + separatorLine.width
     focusBorder.anchors.bottomMargin: 2
 
     background.color: (root.isSelected || hoverHandler.hovered) ?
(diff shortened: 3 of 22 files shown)


src/projectscene/projectscenemodule.cpp
```diff
@@ -101,6 +103,7 @@ void ProjectSceneModule::registerExports()
     ioc()->registerExport<IConnectingDotsPainter>(moduleName(), new ConnectingDotsPainter());
     ioc()->registerExport<IMinMaxRMSPainter>(moduleName(), new MinMaxRMSPainter());
     ioc()->registerExport<ISamplesPainter>(moduleName(), new SamplesPainter());
+    ioc()->registerExport<ITrackNavigationController>(moduleName(), m_trackNavigationController);
```
embarc-gabriel: I've implemented it similarly to other interfaces, but we can safely remove this registration once, as far as I understand, once this interface is not supposed to be used elsewhere. What do you think, @grliszas14 ? Or at least move the interface to a non-internal directory.

embarc-gabriel:
@grliszas14 Another question is the correct place for the track navigation controller file, or even if it should be in a separate file.

src/app/configs/data/shortcuts.xml
```diff
@@ -57,6 +57,19 @@
         <key>nav-left</key>
         <seq>Left</seq>
     </SC>
+    <SC>
+        <key>track-focus-up</key>
```
grliszas14: These are already in shortcuts.xml so maybe we can just respond to it? prev-track/next-track sounds exactly like what we do here, no?
```
    <SC>
        <key>prev-track</key>
        <seq>Up</seq>
    </SC>
    <SC>
        <key>next-track</key>
        <seq>Down</seq>
    </SC>
```

src/projectscene/internal/tracknavigationcontroller.h
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+#pragma once
+
+#include "global/modularity/ioc.h"
+#include "actions/actionable.h"
+
+#include "trackedit/iselectioncontroller.h"
+
+#include "projectscene/internal/itracknavigationcontroller.h"
```
grliszas14: Include order:

```c++
#include "actions/actionable.h"

#include "global/modularity/ioc.h"
#include "trackedit/iselectioncontroller.h"

#include "projectscene/internal/itracknavigationcontroller.h"
```

src/projectscene/internal/tracknavigationcontroller.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace au::projectscene {
+class TrackNavigationController : public ITrackNavigationController, public muse::actions::Actionable
+{
+    INJECT_STATIC(muse::actions::IActionsDispatcher, dispatcher)
```
grliszas14: why INJECT_STATIC?

src/projectscene/projectscenemodule.cpp
grliszas14: It looks good to me to be honest, I don't have any strong opinion on that

grliszas14:
> @grliszas14 Another question is the correct place for the track navigation controller file, or even if it should be in a separate file.

At first I was surprised why not to put all of that in the trackedit module as focus track is already a member of selectionController, so maybe we should put it there? Track navigation being separate file is ok to me

src/app/configs/data/shortcuts.xml
embarc-gabriel: Fixed.

src/projectscene/internal/tracknavigationcontroller.h
embarc-gabriel: Done.

src/projectscene/internal/tracknavigationcontroller.h
embarc-gabriel: Just lazily copied code. Sorry. Fixed.

src/projectscene/projectscenemodule.cpp
embarc-gabriel: So let's keep like this.

embarc-gabriel:
@grliszas14, it is weird to me that some code that does not edit the track content is on the trackedit module. But anyway, we can discuss it later. I moved it to trackedit module.

grliszas14 (approved):

src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
```diff
@@ -295,8 +295,7 @@ Item {
 
                 navigation.name: Boolean(clipItem) ? clipItem.title + clipItem.index : ""
                 navigation.panel: root.navigationPanel
-                navigation.column: index
-                navigation.row: root.trackIdx
+                navigation.column: clipItem ? Math.floor(clipItem.x) : 0
```
grliszas14, line 298: That's an interesting approach :) I bet it solves some corner cases, I wonder about the performance of that (in the navigation internals especially)

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
```diff
@@ -63,20 +63,13 @@ ListItemBlank {
...
-            }
-        }
-    }
-
     height: trackViewState.trackHeight
     opacity: dragged ? 0.5 : 1
 
+    focusBorder.anchors.leftMargin: spacer.width + 2
+    focusBorder.anchors.rightMargin: volumePressureContainer.width + separatorLine.width + 10
+    focusBorder.anchors.bottomMargin: 2
```
grliszas14: Some magic numbers. Maybe there are already variables to replace them to add some context? Maybe it's some column spacing or something?

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TracksPanel.qml
```diff
@@ -13,10 +13,11 @@ import Audacity.ProjectScene
 Item {
     id: root
 
-    property NavigationSection navigationSection: null
-    property NavigationPanel navigationPanel: view.count > 0 ? view.itemAtIndex(0).navigationPanel : null // first panel
+    property var navpanels: null
```
grliszas14: nitpick: navPanels with camelCase

src/trackedit/internal/tracknavigationcontroller.cpp
```diff
@@ -0,0 +1,143 @@
...
+    const auto panels = section->panels();
+    if (static_cast<size_t>(2 * position) >= panels.size()) {
+        return;
+    }
+
+    const auto& panel = std::find_if(panels.begin(), panels.end(), [position](const muse::ui::INavigationPanel* p) {
+        return p->index().order() == 2 * position;
+    });
+
+    const auto firstControl = (*panel)->controls().begin();
```
grliszas14: please add a check if `panel` is not panels.end() -> otherwise we may get UB few lines later

src/trackedit/internal/tracknavigationcontroller.cpp
```diff
@@ -0,0 +1,143 @@
...
+    if (!(*firstControl)) {
+        return;
+    }
+
```
grliszas14, line 92: can firstControl be empty? this may be worth checking as well

grliszas14: same for navigateDown func

src/trackedit/view/tracknavigationmodel.cpp
```diff
@@ -0,0 +1,249 @@
...
+    if (index < 0 || index >= m_trackItemPanels.size()) {
+        return;
+    }
+
+    const auto& panel = m_trackItemPanels.at(index);
+    if (!panel) {
+        return;
+    }
+
+    const auto firstControl = panel->controls().begin();
```
grliszas14, line 177: same here, good to check if `controls` is not empty

src/trackedit/view/tracknavigationmodel.cpp
```diff
@@ -0,0 +1,249 @@
...
+{
+    globalContext()->currentTrackeditProjectChanged().onNotify(this, [this]() {
+        cleanup();
+        load();
+    });
+}
+
+void TrackNavigationModel::init(muse::ui::NavigationSection* section)
+{
+    m_section = section;
```
grliszas14, line 19: That's a surprise, can you write a comment or make it more explicit like:
```
if (!section) {
    m_section = nullptr;
}
```
so we know it's on purpose

src/projectscene/qml/Audacity/ProjectScene/trackspanel/TracksPanel.qml
embarc-gabriel: Fixed.

src/trackedit/internal/tracknavigationcontroller.cpp
embarc-gabriel: Fixed.

src/trackedit/view/tracknavigationmodel.cpp
embarc-gabriel: Fixed.

src/trackedit/view/tracknavigationmodel.cpp
embarc-gabriel: I have removed the default argument. Is that what this is about? I have added the if check as well.

src/trackedit/internal/tracknavigationcontroller.cpp
embarc-gabriel: Fixed.

src/projectscene/qml/Audacity/ProjectScene/clipsview/TrackClipsItem.qml
embarc-gabriel: Yeah... In fact, it is something we need to review while working on the clip navigation. As far as I could see, this is still not perfect, as I cannot navigate to a clip that is outside the view. But anyway, this was just modified so we have minimal functionality for the track navigation. While implement the full navigation stuff we'll need to review this change anyway.

Issue #9232, opened 2025-08-12
(MacOS) AU4 PCM export to WAV doesn't produce a bit-perfect result match with AU3


dozzzzer:
### Bug description

Discovered while testing https://github.com/audacity/audacity/issues/9107

### Steps to reproduce

1. Load a clip into AU3 
2. Export as WAV, 16 bit, 44100 Hz
3. Load the same clip into AU4
4. Export as WAV, 16 bit, 44100 Hz
5. Load these two exported clips into AU3
6. Invert one of them
7. Mix and render
8. Measure RMS of the resulting track

### Expected behavior

`-inf dB`

### Actual behavior

Around `-81 dB`, which points out that the result that AU4 produces isn't a bit-perfect match with AU3.

### Audacity Version

master

### Operating system

macOS

chinakov:
Could be fixed indirectly by https://github.com/audacity/audacity/issues/9207

chinakov:
This is fixed, thanks!

Issue #9220, opened 2025-08-12
Default location for Export


chinakov:
### Description

Currently, if the User opens the AU4 export dialog with reset configuration, the folder path will be empty until the User selects a relevant folder. Instead, the default folder location should be as such: 

**C:\Users\%username%\Documents\Audacity**

Here's how it is in AU4:
<img width="520" height="535" alt="Image" src="https://github.com/user-attachments/assets/b899b97e-badb-429e-9dfb-b017db9a83b0" />

chinakov:
This is done, thanks!

Issue #9207, opened 2025-08-08
Implement export options specific to every format


grliszas14:
### Description

Every single export format has some additional options specific to it.
All of these dynamically shows up/hides in ExportDialog. Implement these

chinakov:
This is tested, relevant options are available, compared to AU3.

Issue #9205, opened 2025-08-08
Difference in AU3 and AU4 export.


chinakov:
### Bug description

If the User exports the same file via Audacity 3/4 with the same settings in .wav or .wavpack (the only lossless formates available in AU4) - the files will not be bit-perfect.

I compared AU3 output file with Reaper output file and got -0.00dB difference. 

The difference between AU3 and AU4 is around -88.1721dB, which is negligble, but it still exists for both formats. 

### Steps to reproduce

1. Generate noise in AU3
2. Export the noise via .wv or .wav file with 32-bit float and 48kHz via AU3
3. Export the noise via .wv or .wav file with 32-bit float and 48kHz via AU3
4. Open new project in AU3
5. Import both files
6. Invert one of the files
7. Mix and Render into new Track
8. Observe Result

### Expected behavior

AU3/AU4 comparision of exported files should be -00.00dB

### Actual behavior

AU3/AU4 comparision of exported files is -88.1721dB

### Audacity Version

129c883ee531ba7ffbbdfa568287fddb4d1d70d0

### Operating system

OS independent

kryksyh:
@chinakov @grliszas14 Did a quick check, it seems that AU4 always export as PCM16, even if it shows Float32 in the export dialog.

These steps gave me bit perfect export:

1. New project created in AU3
2. Add a new track, and set the track format to PCM16, 44100
3. Create tone
4. Export as wav, 16bit PCM, mono, 44100
5. Save the project
6. Open the project in AU4
7. Export as wav

<img width="748" height="415" alt="Image" src="https://github.com/user-attachments/assets/5b56457f-047c-4e26-9984-ea75f73d3c97" />

grliszas14:
Right now there's just a placeholder text "Float32" that is not connected in any way with what's exported. These options for all formats will be implemented in: https://github.com/audacity/audacity/issues/9207

grliszas14:
Could be fixed indirectly by https://github.com/audacity/audacity/issues/9207

chinakov:
This is fixed, thanks!

PR #9271, opened 2025-08-29
Dynamic export options

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9207
Resolves: https://github.com/audacity/audacity/issues/9232
Resolves: https://github.com/audacity/audacity/issues/9205
Resolves: https://github.com/audacity/audacity/issues/9220
Resolves: https://github.com/audacity/audacity/issues/9162

Recommended:

### DIFF
--- a/src/importexport/export/iexporter.h
+++ b/src/importexport/export/iexporter.h
@@ -5,6 +5,7 @@
 
 #include "modularity/imoduleinterface.h"
 
+#include "types/exporttypes.h"
 #include "types/ret.h"
 
 namespace au::importexport {
@@ -25,5 +26,10 @@ public:
 
     virtual int maxChannels() const = 0;
     virtual std::vector<int> sampleRateList() const = 0;
+    virtual int optionsCount() const = 0;
+
+    virtual std::optional<ExportOption> option(int i) const = 0;
+    virtual std::optional<OptionValue> value(int id) const = 0;
+    virtual void setValue(int id, const OptionValue&) = 0;
 };
 }
--- a/src/importexport/export/internal/au3/au3exporter.cpp
+++ b/src/importexport/export/internal/au3/au3exporter.cpp
@@ -245,19 +245,12 @@ std::string Au3Exporter::formatExtension(const std::string& format) const
 
 bool Au3Exporter::isCustomFFmpegExportFormat() const
 {
-    std::string format = exportConfiguration()->currentFormat();
-
-    for (auto [plugin, formatIndex] : ExportPluginRegistry::Get()) {
-        if (plugin->GetFormatInfo(formatIndex).description.Translation().ToStdString() == format) {
-            auto editor = plugin->CreateOptionsEditor(formatIndex, nullptr);
-            if (!editor) {
-                return false;
-            }
-            return editor->GetName() == "custom_ffmpeg";
-        }
+    OptionsEditorUPtr editor = optionsEditor();
+    if (!editor) {
+        return false;
     }
 
-    return false;
+    return editor->GetName() == "custom_ffmpeg";
 }
 
 int Au3Exporter::maxChannels() const
@@ -275,17 +268,96 @@ int Au3Exporter::maxChannels() const
 
 std::vector<int> Au3Exporter::sampleRateList() const
 {
+    OptionsEditorUPtr editor = optionsEditor();
+    if (!editor) {
+        return {};
+    }
+
+    return editor->GetSampleRateList();
+}
+
+int Au3Exporter::optionsCount() const
+{
+    OptionsEditorUPtr editor = optionsEditor();
+    if (!editor) {
+        return 0;
+    }
+
+    return editor->GetOptionsCount();
+}
+
+std::optional<au::importexport::ExportOption> Au3Exporter::option(int i) const
+{
+    OptionsEditorUPtr editor = optionsEditor();
+    if (!editor) {
+        return std::nullopt;
+    }
+
+    ::ExportOption opt;
+
+    if (editor->GetOption(i, opt)) {
+        std::string title = opt.title.Translation().ToStdString();
+        std::vector<std::string> names;
+        for (const auto& name : opt.names) {
+            names.push_back(name.Translation().ToStdString());
+        }
+
+        return ExportOption { opt.id,
+                              title,
+                              opt.flags,
+                              opt.values,
+                              names };
+    }
+
+    return std::nullopt;
+}
+
+std::optional<au::importexport::OptionValue> Au3Exporter::value(int id) const
+{
+    OptionsEditorUPtr editor = optionsEditor();
+    if (!editor) {
+        return std::nullopt;
+    }
+
+    ::ExportValue val;
+
+    if (editor->GetValue(id, val)) {
+        return val;
+    }
+
+    return std::nullopt;
+}
+
+void Au3Exporter::setValue(int id, const OptionValue& value)
+{
+    OptionsEditorUPtr editor = optionsEditor();
+    if (!editor) {
+        return;
+    }
+
+    ::ExportValue val = value;
+
+    editor->SetValue(id, val);
+    editor->Store(*gPrefs);
+}
+
+OptionsEditorUPtr Au3Exporter::optionsEditor() const
+{
     std::string format = exportConfiguration()->currentFormat();
 
     for (auto [plugin, formatIndex] : ExportPluginRegistry::Get()) {
         if (plugin->GetFormatInfo(formatIndex).description.Translation().ToStdString() == format) {
             auto editor = plugin->CreateOptionsEditor(formatIndex, nullptr);
             if (!editor) {
-                return {};
+                LOGE() << "error: failed to create options editor";
+                return nullptr;
             }
-            return editor->GetSampleRateList();
+
+            editor->Load(*gPrefs);
+
+            return editor;
         }
     }
 
-    return {};
+    return nullptr;
 }
--- a/src/importexport/export/internal/au3/au3exporter.h
+++ b/src/importexport/export/internal/au3/au3exporter.h
@@ -13,6 +13,8 @@
 #include "trackedit/iselectioncontroller.h"
 
 namespace au::importexport {
+using OptionsEditorUPtr = std::unique_ptr<ExportOptionsEditor>;
+
 class Au3Exporter : public IExporter
 {
     muse::Inject<au::context::IGlobalContext> globalContext;
@@ -32,6 +34,13 @@ public:
 
     int maxChannels() const override;
     std::vector<int> sampleRateList() const override;
+    int optionsCount() const override;
+
+    std::optional<ExportOption> option(int i) const override;
+    std::optional<OptionValue> value(int id) const override;
+    void setValue(int id, const OptionValue&) override;
+
+    OptionsEditorUPtr optionsEditor() const;
 
 private:
     double m_t0 {};
--- a/src/importexport/export/qml/Export/ExportDialog.qml
+++ b/src/importexport/export/qml/Export/ExportDialog.qml
@@ -16,17 +16,13 @@ StyledDialogView {
     title: qsTrc("export", "Export audio")
 
     contentWidth: 490
-    // TODO: switch to mainColumn.implicitHeight when
-    // dynamic content is implemented
-    // contentHeight: mainColumn.implicitHeight
-    contentHeight: baseExportDialogHeight
+    contentHeight: mainColumn.implicitHeight
 
-    margins: 20
+    margins: 10
 
     property int dropdownWidth: 364
     property int smallDropdownWidth: 364 * 0.65
     property int labelColumnWidth: 80
-    property int baseExportDialogHeight: 450
 
     // property NavigationPanel navigation: NavigationPanel {
     //     name: root.title
@@ -43,9 +39,17 @@ StyledDialogView {
         id: ffmpegPrefModel
     }
 
+    DynamicExportOptionsModel {
+        id: dynamicOptionsModel
+    }
+
     Component.onCompleted: {
+        // NOTE: position dialog so entire content is visible
+        y = -250
+
         exportPreferencesModel.init()
         ffmpegPrefModel.init()
+        dynamicOptionsModel.init()
     }
 
     ColumnLayout {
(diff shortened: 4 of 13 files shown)


src/importexport/export/internal/au3/au3exporter.cpp
```diff
@@ -289,3 +289,94 @@ std::vector<int> Au3Exporter::sampleRateList() const
 
     return {};
 }
+
+int Au3Exporter::optionsCount() const
+{
+    std::string format = exportConfiguration()->currentFormat();
```
saintmatthieu: Looks like there could be a helper, `OptionsEditorPtr optionsEditor() const` , right?

src/importexport/export/internal/au3/au3exporter.cpp
```diff
@@ -289,3 +289,94 @@ std::vector<int> Au3Exporter::sampleRateList() const
...
+}
+
+void Au3Exporter::setValue(int id, const ExportValue& value)
+{
+    std::string format = exportConfiguration()->currentFormat();
+
+    for (auto [plugin, formatIndex] : ExportPluginRegistry::Get()) {
+        if (plugin->GetFormatInfo(formatIndex).description.Translation().ToStdString() == format) {
+            auto editor = plugin->CreateOptionsEditor(formatIndex, nullptr);
+            if (!editor) {
```
saintmatthieu, line 351: Is there a scenario where `setValue` is called and yet there is no available editor? If not, I suppose that an error should be logged.

src/importexport/export/internal/au3/au3exporter.cpp
grliszas14: aye, added 👍

src/importexport/export/types/exporttypes.h
```diff
@@ -26,4 +26,45 @@ class ExportChannelsPref
...
+        TypeEnum         = 2,
+
+        ReadOnly         = 0x100,
+        Hidden           = 0x200,
+
+        Default          = 0
+    };
+
+    int id = -1;
+    std::string title;
```
saintmatthieu, line 66: Looks like `<string>` might not be transitively included by `#include <qobjectdefs.h>` ? Explicit includes (for this and vector) would be more robust.

src/importexport/export/types/exporttypes.h
```diff
@@ -26,4 +26,45 @@ class ExportChannelsPref
     };
     Q_ENUM(ExportChannels)
 };
+
+class ExportOptionType
```
saintmatthieu, line 32: These refer to the UI framework from within an application-logic source file. We had this discussion about the delete behavior (https://mu--se.slack.com/archives/C06RX437GJV/p1755682183400249 - restricted access) and there's a way of keeping this ui-agnostic here and converting to a ui-specific type in a `view` directory.
This is a bit more work, though, so I'm not sure it's worthwhile. Only if you have some spare energy and you don't find the idea too silly `:)`

src/importexport/export/types/exporttypes.h
```diff
@@ -26,4 +26,45 @@ class ExportChannelsPref
...
+
+        ReadOnly         = 0x100,
+        Hidden           = 0x200,
+
+        Default          = 0
+    };
+
+    int id = -1;
+    std::string title;
+    ExportValue defaultValue;
```
saintmatthieu: It seems that this `defaultValue` member isn't used in non-au3 code and hence could be removed.

src/importexport/export/iexporter.h
```diff
@@ -25,5 +26,10 @@ class IExporter : MODULE_EXPORT_INTERFACE
 
     virtual int maxChannels() const = 0;
     virtual std::vector<int> sampleRateList() const = 0;
+    virtual int optionsCount() const = 0;
+
+    virtual std::optional<ExportOption> option(int i) const = 0;
+    virtual std::optional<ExportValue> value(int id) const = 0;
```
saintmatthieu: Sounds a bit terse - `optionValue`, would that be correct?

src/importexport/export/internal/au3/au3exporter.h
```diff
@@ -13,6 +13,8 @@
 #include "trackedit/iselectioncontroller.h"
 
 namespace au::importexport {
+using OptionsEditorPtr = std::unique_ptr<ExportOptionsEditor>;
```
saintmatthieu: `OptionsEditorUPtr` ? I've been using it in lv2viewmodel and I don't know of another place where just `Ptr` is used for a unique pointer.

src/importexport/export/qml/Export/ExportDialog.qml
```diff
@@ -376,56 +380,73 @@ StyledDialogView {
...
-
-                Component.onCompleted: {
-                    if (visible) {
-                        root.contentHeight = root.baseExportDialogHeight + 110
-                    }
-                }
             }
 
-            RowLayout {
+            ColumnLayout {
```
saintmatthieu, line 385: Question that I have in general: for me, `Column` and `Row` are simpler to predict that their layout counterparts. Since this dialog (and the vast majority of our dialogs) isn't supposed to be resizeable, wouldn't it make it easier to everyone to just use the simpler tool ?

src/importexport/export/qml/Export/ExportDialog.qml
```diff
@@ -376,56 +380,73 @@ StyledDialogView {
...
 
-                StyledDropdown {
-                    id: encodingDropdown
+                    delegate: RowLayout {
+                        Layout.fillWidth: true
+                        Layout.preferredHeight: 30
 
-                    Layout.preferredWidth: root.smallDropdownWidth
+                        visible: !hidden
+                        enabled: !readOnly
```
saintmatthieu: I don't understand whom this `readOnly` property belongs to?

saintmatthieu: Same for `hidden`, actually.

src/importexport/export/qml/Export/ExportDialog.qml
```diff
@@ -376,56 +380,73 @@ StyledDialogView {
...
+                            }
+                        }
 
-                    onActivated: function(index, value) {
+                        Loader {
+                            id: control
+
+                            Layout.preferredWidth: root.smallDropdownWidth
+
+                            property int rIndex: index
```
saintmatthieu: What's the meaning behind the r prefix?

src/importexport/export/view/dynamicexportoptionsmodel.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+void DynamicExportOptionsModel::reloadFromEditor()
+{
+    std::vector<Row> newRows;
+    newRows.reserve(exporter()->optionsCount());
+
+    for (int i = 0; i < exporter()->optionsCount(); ++i) {
+        std::optional<au::importexport::ExportOption> option_opt = exporter()->option(i);
+        if (!option_opt.has_value()) {
+            continue;
```
saintmatthieu, line 179: Is this a use case that `exporter()->optionsCount()` says e.g. 2 and `exporter(0 or 1)` returns nullopt? If not, please assert: it would help the understanding a bit.

src/importexport/export/qml/Export/ExportDialog.qml
```diff
@@ -499,4 +520,135 @@ StyledDialogView {
...
+            onActivated: function(index, value) {
+                dynamicOptionsModel.setData(dynamicOptionsModel.index(rIndex, 0),
+                                            rValues[index],
+                                            ExportOptionType.ValueRole)
+            }
+        }
+    }
+
+    Component {
+        id: boolComp
```
saintmatthieu, line 533: I didn't find a format that would have a boolean setting. Is there really one?

src/importexport/export/qml/Export/ExportDialog.qml
```diff
@@ -499,4 +520,135 @@ StyledDialogView {
 
     }
 
+    // dynamic export controls
+
+    Component {
+        id: enumComp
+        StyledDropdown {
+            property int rIndex: -1
```
saintmatthieu: Have you considered (if possible at all) placing these properties in a QML file,
```
QtObject {
            property int rIndex: -1
            property var rValue: -1
            property var rValues: []
            property var rNames: []
            property int rMin: 0
            property int rMax: 0
}
```
so that you can reuse it in the 6 different places it's used?
This dialog is so complex, any simplification is welcome, really.

src/importexport/export/qml/Export/ExportDialog.qml
saintmatthieu, line 1: If possible I'd really like to reduce all possible fat off of this big, big source file. Things I see possible:
* Gathering those properties that are used in 6 different place in a dedicated file, as explained in an inline comment.
* Place the components defined at the bottom of the file in their own source files.
* Use the `RadioButtonGroup` with `id: channelsGroup` properly, making use of its `model` property (see `DeleteBehaviorPanel.qml` for an example)
* Use Row and Column in place of RowLayout and ColumnLayout where possible ?

src/importexport/export/qml/Export/ExportDialog.qml
```diff
@@ -43,9 +39,17 @@ StyledDialogView {
         id: ffmpegPrefModel
     }
 
+    DynamicExportOptionsModel {
+        id: dynamicOptionsModel
+    }
+
     Component.onCompleted: {
+        // NOTE: position dialog so entire content is visible
+        y = -250
```
Eism, line 48: Could you explain what problem you are solving with this?

src/importexport/export/qml/Export/ExportDialog.qml
Eism: what does the prefix r mean?

grliszas14: ayy, I just wanted to distinguish between export properties and index/value that comes from the dropdown for example (avoid override). Maybe it would be better to name it with `m` prefix? Like an internal variable or something? Or undersore?

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: In the export dialog we have static components and dynamically loaded. At the time export dialog is positioned on the screen we only know about static ones -> and it is positioned correctly in the center. But when dynamic components are loaded, the dialog grows in height but the dialog does not reposition itself again: that means a part of the dialog gets rendered below the screen and it's not visible.

This is just a hack to move window a bit more to the top so it doesn't matter which format you choose, you always see entire dialog on the screen.

Do you have any ideas how to solve that riddle without this ugly hack?

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: It belongs to the single item from dynamicOptionsModel:

```qml
Repeater {
    model: dynamicOptionsModel

    delegate: RowLayout {
        Layout.fillWidth: true
        Layout.preferredHeight: 30

        visible: !hidden
        enabled: !readOnly
``` 

You have a model and delegate: delegate is a single item from the model. If you want to know which members of delegate are accessible in QML (and by which name), go to: `DynamicExportOptionsModel::roleNames()`. If you then want to know what is returned if you call a member: take a look at its role and look at switch in `DynamicExportOptionsModel::data`

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: https://github.com/audacity/audacity/pull/9271#discussion_r2329585209

src/importexport/export/qml/Export/ExportDialog.qml
saintmatthieu: Ah, ok, so would `modelData.readOnly` also work? In which case I'd prefer that to make it explicit.

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: yes, there is one in the design - external program export has "Show output" checkbox. It's just not implemented yet

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: Cleaned it up so it was not needed anymore 👍

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: `model.readOnly` works, I changed it

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: Made it a single property, it cleaned up beautifully

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: I'm not convinced to be honest, if it's done this way it's kinda future-proof for free, if a designer decides to make that dialog a bit different size later it will automatically re-adjust. Besides that, this dialog expands/shrinks according to selected format and I don't have to worry about if all of the elements are properly aligned, with all of the paddings/margins and if they're visible.

src/importexport/export/qml/Export/ExportDialog.qml
grliszas14: 1. All of these components are so small after a cleanup that it doesn't really make sense to me to put a separate file with this for example:

```qml
Component {
        id: enumComp
        StyledDropdown {
            property var option

            model: option.names
            currentIndex: option.value

            onActivated: function(index, value) {
                dynamicOptionsModel.setData(dynamicOptionsModel.index(option.index, 0),
                                            option.values[index],
                                            ExportOptionType.ValueRole)
            }
        }
    }
``` 
left it as is

2. About RadioButtonGroup: I have it planned to do here: https://github.com/audacity/audacity/issues/8963

src/importexport/export/qml/Export/ExportDialog.qml
saintmatthieu: Fair enough.

src/importexport/export/qml/Export/ExportDialog.qml
saintmatthieu: ok

Issue #9222, opened 2025-08-12
File Picker should be used instead of the Folder Picker in Export Dialog


chinakov:
### Description

We should change the Folder picker that is currently used in the export dialog to the default file picker. 

<img width="1166" height="839" alt="Image" src="https://github.com/user-attachments/assets/78a82e14-4677-4a5a-828f-68a15de261db" />

dozzzzer:
Tested

Issue #9221, opened 2025-08-12
File Extension in the Export Dialog should be Removed


chinakov:
### Description

For simplicity sake, we should remove the appended extension within the Export Dialog. 

<img width="522" height="567" alt="Image" src="https://github.com/user-attachments/assets/957092ff-64d7-42fd-ab60-1864b1f3ff3b" />

dozzzzer:
@grliszas14 Currently, the project's file extension is populated into the file name textbox:

<img width="510" height="551" alt="Image" src="https://github.com/user-attachments/assets/d8e2286b-b66b-43f3-ae25-54474c9e48cd" />

If you then try to click "Browse" and specify a file name, it will append the extension in that case, but not even the right one:

https://github.com/user-attachments/assets/fa924fc5-1748-4261-90b0-25283a0f1d7b

On Windows, initially, the file name doesn't have the file extension appended, but if you specify the file name, it would also add one:

https://github.com/user-attachments/assets/9e879cd1-261d-4234-8571-5211b6a9c13f

chinakov:
Tested, thanks!

PR #9295, opened 2025-09-03
ExportDialog's FilePicker improvements

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9222
Resolves: https://github.com/audacity/audacity/issues/9221

Depends on: https://github.com/musescore/MuseScore/pull/29643

Recommended:

QA:


src/importexport/export/view/exportpreferencesmodel.cpp
```diff
@@ -241,27 +281,17 @@ void ExportPreferencesModel::setExportSampleRate(const QString& rateName)
...
+    dispatcher()->dispatch("open-custom-ffmpeg-options");
 }
 
-QVariantList ExportPreferencesModel::exportSampleFormatList() const
+void ExportPreferencesModel::setFilePickerPath(const QString& path)
 {
-    NOT_IMPLEMENTED;
-    return {};
-}
+    QFileInfo info(path);
```
kryksyh: It might be better to reuse framework's FileInfo wrapper

PR #9345, opened 2025-09-10
Detect unnecessary modules during plugin registration

saintmatthieu:
Looking at #9330, I realized that no one noticed that modules that were unnecessary to plugin registration had been added even to the plugin-registration path.
It is critical that plugin scanning stays as fast as possible, hence the idea of this PR as mechanism against inadvertent loading of unnecessary modules.

Recommended:

QA:

### DIFF
--- a/src/app/app.cpp
+++ b/src/app/app.cpp
@@ -58,6 +58,34 @@ void App::addModule(modularity::IModuleSetup* module)
     m_modules.push_back(module);
 }
 
+namespace {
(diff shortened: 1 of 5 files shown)


grliszas14 (approved):

src/app/app.cpp
```diff
@@ -86,6 +114,12 @@ int App::run(QCoreApplication& app, CommandLineParser& commandLineParser)
 
     const IApplication::RunMode runMode = commandLineParser.runMode();
 
+#ifdef QT_DEBUG
```
kryksyh: runtime::isDebug()

PR #9348, opened 2025-09-10
Fix vertical playback meter volume slider position

### LINKED ISSUE #9280: Playback volume slider isn't moving
### Bug description

_No response_

### Steps to reproduce

1. enable Playback meter
2. set its position to vertical
3. click and drag the slider
4. observe

### Expected behavior

The slider is moving along the vertical meter scale as you're dragging it.

### Actual behavior

The slider remains still, although the tooltip value is changing:

https://github.com/user-attachments/assets/55d43932-72f5-464a-9e3c-ccb4f92066c0

### Audacity Version

master

### Operating system

OS independent

### Additional context

_No response_

embarc-gabriel:
Resolves: #9280 

The volume information should be updated in the meter model to calculate the slider position correctly.

Recommended:

QA:

### DIFF
--- a/src/playback/view/panels/playbackmeterpanelmodel.cpp
+++ b/src/playback/view/panels/playbackmeterpanelmodel.cpp
@@ -7,6 +7,14 @@ using namespace au::playback;
 PlaybackMeterPanelModel::PlaybackMeterPanelModel(QObject* parent)
     : QObject(parent)
 {
+    m_meterModel = new PlaybackMeterModel(this);
+}
+
+void PlaybackMeterPanelModel::init()
+{
+    m_meterModel->init();
+    emit meterModelChanged();
+
     playback()->audioOutput()->playbackSignalChanges().onReceive(this,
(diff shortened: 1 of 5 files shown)


src/playback/view/panels/playbackmeterpanelmodel.cpp
```diff
@@ -7,6 +7,10 @@ using namespace au::playback;
 PlaybackMeterPanelModel::PlaybackMeterPanelModel(QObject* parent)
     : QObject(parent)
 {
+    m_meterModel = new PlaybackMeterModel(this);
+    m_meterModel->init();
```
saintmatthieu, line 15: Better add an `init` method to this class and call `m_meterModel->init()` from inside, I think.

saintmatthieu (approved):

src/playback/view/panels/playbackmeterpanelmodel.cpp
embarc-gabriel: Fixed.

PR #9356, opened 2025-09-11
9105 compressor UI

### LINKED ISSUE #9105: Migrate New Compressor (non-destructive) to AU4

saintmatthieu:
Resolves: #9105 

Recommended:

QA:

- [x] Compressor and limiter in destructive modes are unchanged
- [x] Bypass works
- [x] Pause / Stop / Resume work
- [x] Clicking the clip indicator resets it
- [x] Clicking the compression and output meters resets them
- [x] When stopping and resuming, the meters get reset but not after pause
- [x] The clip indicator is only present on the compressor (since clipping is impossible thanks to the limiter - ha!)
- [x] Please smoke-test effects of al families on each OS, to make sure that their UI appears and that playback starts
- [x] Line below presets bar spans the entire width for built-in effects used destructively (unchanged for other families or in realtime mode)
- [x] Works equally well for other project sample rates, even extreme ones such as 8kHz.
- [x] All parameters are persistent, including the checkboxes.
- [x] Defaults settings have knobs either in the middle, hard-left or hard-right. (To reset the default settings, delete your pluginsettings.cfg, whose path can be found in the Diagnostics.)


src/effects/builtin/dynamics/draft/dynamicstimeline.h
```diff
@@ -0,0 +1,116 @@
...
+    double m_duration = 5;
+    double m_samplePeriod = 0;
+    int m_prevSampleIndex = -1;
+    double m_t = 0;
+    bool m_reset = false;
+
+    QRectF m_drawerViewport;
+    std::array<SequenceData, _eSequenceCount> m_sequences;
+    std::vector<double> m_pendingXValues;
+    std::mutex m_sampleMutex; // TODO consider using a ring buffer
```
kryksyh, line 122: That was exacly my thought, "hey, the ring buffer should fit nicely here"

.github/workflows/au4_build_linux.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         host: 'linux'
         target: 'desktop'
         arch: 'linux_gcc_64'
-        modules: 'qt5compat qtnetworkauth qtshadertools qtwebsockets'
+        modules: 'qt5compat qtnetworkauth qtshadertools qtwebsockets qtgraphs'
```
kryksyh: to be removed, right?

src/effects/builtin/dynamics/draft/dynamicstimeline.h
```diff
@@ -0,0 +1,116 @@
...
+    SequenceData createSequenceData(const QColor&, DrawerType) const;
+    void resetSequences();
+
+    void componentComplete() override;
+    double timeToX(double time) const;
+    double xToTime(double x) const;
+    double dbToY(double db) const;
+    double drawerViewportX() const;
+
+    enum SequenceId {
```
kryksyh: Could you rename this to minimize false positives when searching for Sequience from au3? Maybe `CompressionDataPoint` or something like that

.github/workflows/au4_build_linux.yml
saintmatthieu: I don't know: configuration on the build agent fails because QtGraphs is not found. I thought that this would fix it, but it doesn't.

src/effects/builtin/dynamics/draft/dynamicstimeline.h
saintmatthieu: Sure, I'll find something.

src/effects/builtin/dynamics/timeline/timelinesourcemodel.h
saintmatthieu, line 1: header

src/effects/builtin/dynamics/timeline/timelinesourcemodel.cpp
saintmatthieu, line 1: header

src/effects/builtin/dynamics/timeline/test/timelinesourcetestmodel.h
saintmatthieu, line 1: header

src/effects/builtin/dynamics/timeline/test/timelinesourcetestmodel.cpp
saintmatthieu, line 1: header

src/effects/builtin/dynamics/timeline/test/meters/dynamicseffectoutputmetertestmodel.h
saintmatthieu, line 1: header

teetow:
So I spent some time thinking about the default ranges here. These suggestions are definitely conservative, but in my view this aligns with the needs of both beginners who might need some guidance, and seasoned producers who tend to compress carefully. As we discussed, the Mid value in the table below is also what the knob will reset to, unless specified otherwise.

| Parameter | Min | Mid | Max |
|------------|------|-----|-----|
|Attack |0.1 | 3 | 100 |
|Release| 10| 100 | 1000 |
|Ratio | 1 | 4 | 20 |
|Threshold| -30 | - | 0 (default) |
|Makeup| 0 (default) | - | 30  |
|Lookahead| 0 | 3 | 100 |
|Knee|0| 6| 30|

Some clarifications on Makeup gain: Since a compressor will always reduce the gain and never raise it, negative Makeup gain isn't strictly necessary. I suggest we make it behave like the Threshold knob (albeit inverted) so it'll reset to its minimal value of 0 dB rather than its midpoint.

One nice UX tweak would be to add a checkbox to enable _Auto-makeup_. If enabled, the Makeup is set to the inverse of the Threshold; -12 dB Threshold yields 12 dB Makeup. The Makeup knob would then shift to a -12 < 0 + 12 dB bipolar knob to adjust for clipping transients, and reset to its midpoint of 0 dB. If this is trivial, we could experiment with it, otherwise we skip.

.vscode/launch.json
```diff
@@ -14,7 +14,7 @@
             "environment": [
                 {
                     "name": "PATH",
-                    "value": "${env:PATH};${command:cmake.launchTargetDirectory}\\bin"
+                    "value": "${env:PATH};${command:cmake.launchTargetDirectory}\\bin;C:/git/saintmatthieu/audacity/build/audacity-debug/src/app/bin"
```
luapmartin: maybe typo?

.github/workflows/au4_build_linux.yml
luapmartin, line 1: see all the workflows files having so much in common, I am wondering if we should refactor for simpler setup? maybe in another issue

CMakeLists.txt
```diff
@@ -98,6 +98,7 @@ option(AU_BUILD_PLAYBACK_TESTS "Build playback tests" ON)
 option(AU_BUILD_PROJECT_TESTS "Build project tests" ON)
 option(AU_BUILD_PROJECTSCENE_MODULE "Build projectscene modules" ON)
 option(AU_BUILD_PROJECTSCENE_TESTS "Build projectscene tests" ON)
+option(AU_BUILD_EFFECTS_BUILTIN_TESTS "Build builtin-effect tests" ON)
```
luapmartin: I think we have to respect alphabetical order here

au3/libraries/lib-builtin-effects/CompressorInstance.cpp
```diff
@@ -184,11 +192,12 @@ size_t CompressorInstance::RealtimeProcess(
...
-        queue->Put(MeterValues {
-            compressor.GetLastFrameStats().dbGainOfMaxInputSample,
-            GetOutputDb(
-                compressor.GetLastFrameStats(), compressor.GetSettings()) });
+    if (const auto queue = slave.mCompressionGainDbQueue.lock()) {
+        queue->Put(compressor.GetLastFrameStats().dbGainOfMaxInputSample);
+    }
+
+    if (const auto queue = slave.mOutputDbQueue.lock()) {
+        queue->Put(GetOutputDb(compressor.GetLastFrameStats(), compressor.GetSettings()));
```
luapmartin, line 207: should we name those scoped `queue` to ease reading?

au3/libraries/lib-builtin-effects/CompressorInstance.cpp
```diff
@@ -212,7 +221,8 @@ void CompressorInstance::InstanceInit(
     float sampleRate)
 {
     instance.mOutputQueue = mOutputQueue;
-    instance.mCompressionValueQueue = mCompressionValueQueue;
+    instance.mCompressionGainDbQueue = mCompressionGainDbQueue;
+    instance.mOutputDbQueue = mOutputDbQueue;
```
luapmartin, line 232: maybe we could wrap queues together in case we are going to add input visualisation too etc. and we could maybe also re-use such work with other processors

au3/libraries/lib-dynamic-range-processor/DynamicRangeProcessorHistory.h
```diff
@@ -32,18 +32,36 @@ class DYNAMIC_RANGE_PROCESSOR_API DynamicRangeProcessorHistory final
...
 
 private:
     float GetPacketTime(const DynamicRangeProcessorOutputPacket& packet) const;
+    int TotalNumPackets() const;
 
     const double mSampleRate;
     bool mBeginNewSegment = true;
-    std::vector<Segment> mSegments;
+    Segments mSegments;
+    int m_numViewedPackets = 0;
```
luapmartin, line 64: should we `m_` all other members?

src/effects/builtin/CMakeLists.txt
```diff
@@ -159,3 +185,10 @@ set(MODULE_LINK au3wrap)
 set(MODULE_USE_UNITY OFF)
 
 setup_module()
+
+if (OS_IS_WIN)
+    add_subdirectory(dynamics/timeline)
+    if (AU_BUILD_EFFECTS_BUILTIN_TESTS)
+        add_subdirectory(tests)
+    endif()
+endif()
```
luapmartin: is this expected?

src/effects/builtin/common/builtineffectsettingmodel.h
```diff
@@ -15,6 +15,7 @@ class BuiltinEffectSettingModel : public BuiltinEffectModel
     Q_PROPERTY(double min READ min CONSTANT FINAL)
     Q_PROPERTY(double max READ max CONSTANT FINAL)
     Q_PROPERTY(double step READ step CONSTANT FINAL)
+    Q_PROPERTY(double defaultValue READ defaultValue CONSTANT FINAL)
```
luapmartin, line 18: reminded me of skew etc. from JUCE https://docs.juce.com/master/classNormalisableRange.html

src/effects/builtin/common/valuewarper/valuewarper.cpp
```diff
@@ -14,24 +14,62 @@ namespace {
...
 }
 
-void ValueWarper::init()
+bool ValueWarper::useWarper() const
+{
+    if (m_middle.isNull()) {
+        return false;
+    }
+    const double mid = m_middle.toDouble();
+    // Special cases to avoid, or we'll be getting nans.
```
luapmartin:
```suggestion
    // Special cases to avoid, or we'll be getting NaNs.
```

src/effects/builtin/common/valuewarper/warpingtransformer.cpp
```diff
@@ -26,42 +26,46 @@ struct Expanded {
...
+ * Now we let y(1/2) = mid, i.e., the value when in the middle, and solve the equation for C:
+ * C = 2 * log((1 - u) / u)
  */
-constexpr double softC = 2.1972245773362196;
-constexpr double aggressiveC = 3.8918202981106265;
-static const double softExpC = std::exp(softC);
-static const double aggressiveExpC = std::exp(aggressiveC);
+double solveC(double min, double mid, double max)
+{
+    const double u = Normalized{ mid, min, max }.value;
```
luapmartin: should we do something for this min mid max vs mid min max ? maybe a little struct or unify the order ?

luapmartin: mid could be named center if it helps too

luapmartin: or regroup together  min and max in "range"

src/effects/builtin/dynamics/compressor/CompressorView.qml
```diff
@@ -0,0 +1,165 @@
...
+                        model: [
+                            {
+                                id: "thresholdDb",
+                                title: qsTrc("effects/compressor", "Threshold"),
+                                unit: "dB"
+                            },
+                            {
+                                id: "compressionRatio",
+                                title: qsTrc("effects/compressor", "Ratio"),
+                                unit: ""
```
luapmartin: actually the unit is sometime a "1:" prefix but we can't prefix units right?

src/effects/builtin/dynamics/limiter/LimiterView.qml
```diff
@@ -0,0 +1,143 @@
...
+            visible: !root.usedDestructively
+
+            instanceId: limiter.instanceId
+            playState: root.playState
+            gridColor: root.gridColor
+
+            // Specific to limiter
+            dbMin: -12
+            dbStep: 3
+            duration: 2
```
luapmartin, line 51: should these be moved to the model?

src/effects/builtin/dynamics/limiter/LimiterView.qml
```diff
@@ -0,0 +1,143 @@
...
+            needsClipIndicator: false // Clipping with the limiter is impossible.
+        }
+
+        Rectangle {
+
+            color: ui.theme.backgroundSecondaryColor
+            border.color: ui.theme.strokeColor
+            width: knobRow.width
+            height: knobRow.height
+            radius: 4
```
luapmartin, line 62: is this the same knob container as for the compressor? maybe could reuse a component there

src/effects/builtin/dynamics/limiter/LimiterView.qml
```diff
@@ -0,0 +1,143 @@
...
+                        required property var modelData
+                        isVertical: true
+                        knobFirst: false
+                        radius: 16
+                        title: modelData.title
+                        unit: modelData.unit
+                        warp: true
+                        model: LimiterSettingModel {
+                            paramId: modelData.id
+                        }
```
luapmartin, line 123: we maybe could get those  title and unit from the modelData based on the id, actually we are re-using parameters between different effects like compressor and limiter and that would simplify the translation and setup too

luapmartin: maybe also accessibility and navigation I guess

src/effects/builtin/dynamics/timeline/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+qt_add_qml_module(dynamics_timeline_testapp
+    URI dynamics_timeline_testapp
+    VERSION 1.0
+    RESOURCE_PREFIX "/qt/qml/"
+    QML_FILES
+        test/main.qml
+        ClipIndicator.qml
+        DynamicsPanel.qml
+        meters/DynamicsMeter.qml
+)
```
luapmartin: <3

src/effects/builtin/dynamics/timeline/ClipIndicator.qml
```diff
@@ -0,0 +1,71 @@
...
+
+    property bool isClipping: false
+
+    signal clicked
+
+    QtObject {
+        id: prv
+
+        readonly property int radius: 2
+        property color color: root.isClipping ? "#FF1C1C" : ui.theme.backgroundSecondaryColor
```
luapmartin: hex candidate to be moved as extra color?

src/effects/builtin/dynamics/timeline/abstractdynamicseffectinstancemodel.h
```diff
@@ -0,0 +1,42 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+#pragma once
+
+#include "dynamicstimelinetypes.h"
+#include "effects/effects_base/ieffectinstancesregister.h"
+#include "global/modularity/ioc.h"
```
luapmartin: should it be global first?

src/effects/builtin/dynamics/timeline/dynamicscolors.h
```diff
@@ -0,0 +1,38 @@
...
+    static QColor backgroundColor() { return QColor("#313147"); }
+
+    static QColor timelineDataFillColor() { return QColor("#565695"); }
+    static QColor timelineDataFillColorSemiTransparent() { return QColor("#80565695"); }
+
+    static QColor timelineCompressionDbColor() { return QColor("#FFD12C"); }
+    static QColor timelineCompressionDbColorSemiTransparent() { return QColor("#80FFD12C"); }
+
+    static QColor timelineOutputDbLineColor() { return QColor("white"); }
+
```
luapmartin, line 29: do we want to use the extra color here? that's already nicer in the model, thanks

src/effects/builtin/dynamics/timeline/dynamicstimeline.h
```diff
@@ -0,0 +1,124 @@
...
+        _eSequenceCount
+    };
+
+    QSGNode* updatePaintNode(QSGNode*, UpdatePaintNodeData*) override;
+
+    double m_dbMin = -60;
+    double m_duration = 5;
+    double m_dataPointRate = 0;
+    int m_prevSampleIndex = -1;
+    double m_stopwatchTime = 0;
```
luapmartin: do we need a model for that?

src/effects/builtin/dynamics/timeline/meters/abstractdbmetermodel.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace au::effects {
+class AbstractDbMeterModel : public AbstractDynamicsEffectInstanceModel
+{
+    Q_OBJECT
+
+    Q_PROPERTY(au::effects::Stopwatch::PlayState playState READ playState WRITE setPlaystate NOTIFY playStateChanged)
+    Q_PROPERTY(double currentMax READ currentMax NOTIFY valueChanged)
+    Q_PROPERTY(double globalMax READ globalMax NOTIFY valueChanged)
+    Q_PROPERTY(double fiveSecMax READ fiveSecMax NOTIFY valueChanged)
```
luapmartin, line 20: if it's abstract maybe we would that to be temporaryMax and be able to set the time window size of that temporaryMx to 5 seconds

src/effects/builtin/dynamics/timeline/profiler.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * Audacity: A Digital Audio Editor
+ */
+#pragma once
+
+#include <chrono>
+#include <vector>
+
+namespace au::effects {
+class Profiler
```
luapmartin: a Profiler? should this be a tool somewhere to use in debug?

src/effects/builtin/dynamics/timeline/timelinesourcemodel.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+double TimelineSourceModel::latency() const
+{
+    return deliveryPeriodMs / 1000.0;
+}
+
+double TimelineSourceModel::dataPointRate() const
+{
+    return 44100.0 / 512; // TODO query from the instance
```
luapmartin: yes, please

src/effects/builtin/effects_builtin.qrc
```diff
@@ -2,18 +2,25 @@
...
-        <file>common/ParameterKnob.qml</file>
-        <file>common/BigParameterKnob.qml</file>
-        <file>common/SettingKnob.qml</file>
+        <file>qml/Audacity/BuiltinEffects/DynamicsEffectBase.qml</file>
+        <file>qml/Audacity/BuiltinEffects/BuiltinEffectBase.qml</file>
+        <file>qml/Audacity/BuiltinEffects/SliderWithTextInput.qml</file>
+        <file>qml/Audacity/BuiltinEffects/ParameterKnob.qml</file>
+        <file>qml/Audacity/BuiltinEffects/BigParameterKnob.qml</file>
+        <file>qml/Audacity/BuiltinEffects/SettingKnob.qml</file>
+
```
luapmartin, line 11: nice

src/effects/builtin/internal/builtineffectsrepository.cpp
```diff
@@ -216,19 +224,21 @@ void BuiltinEffectsRepository::updateEffectMetaList()
                     true
                     );
         } else if (symbol == CompressorEffect::Symbol) {
+            hasDynamicRangeProcessor = true;
```
luapmartin, line 229: sweat

src/effects/builtin/qml/Audacity/BuiltinEffects/BigParameterKnob.qml
```diff
@@ -9,9 +9,10 @@ Item {
 
     required property var parameter
 
+    property double defaultValue: 0
```
luapmartin, line 12: should we put default in knob or wraper or model somewhere?

src/effects/builtin/qml/Audacity/BuiltinEffects/BigParameterKnob.qml
```diff
@@ -70,6 +67,10 @@ Item {
                 warper.warpedValue = value
             }
 
+            mouseArea.onDoubleClicked: function () {
+                root.newValueRequested(root.parameter["key"], root.defaultValue)
+            }
+
```
luapmartin, line 73: <3

src/effects/builtin/qml/Audacity/BuiltinEffects/SettingKnob.qml
```diff
@@ -31,7 +31,8 @@ Item {
 
         BigParameterKnob {
             radius: root.radius
-            warpingType: root.warpingType
+            defaultValue: root.model.defaultValue
+            middle: root.warp ? root.model.defaultValue : null
```
luapmartin, line 29: is this a typo?

src/effects/builtin/tests/DynamicRangeProcessorHistory_tests.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+TEST(DynamicRangeProcessorHistoryTests, CalledBeforeAnyPush)
+{
+    DynamicRangeProcessorHistory history(sampleRate);
+    const auto view = history.GetViewOnNewPackets();
+    EXPECT_EQ(view.numPackets(), 0);
+}
+
+TEST(DynamicRangeProcessorHistoryTests, stuff) {
```
luapmartin: that's a cute test name :)

src/effects/builtin/tests/environment.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+#include "testing/environment.h"
+
+static muse::testing::SuiteEnvironment effects_builtin_se(
+{
+},
+    nullptr,
+    []() {
+}
+    );
```
luapmartin: i guess this need some love too

luapmartin (commented):
huge work, nice !

src/effects/builtin/dynamics/compressor/CompressionCurve.qml
```diff
@@ -19,9 +19,9 @@ Item {
 
     QtObject {
         id: prv
-        readonly property int min: -60
+        readonly property int min: -36
         readonly property int max: 0
-        readonly property int step: 12
+        readonly property int step: 6
```
luapmartin, line 23: should we put that in a model?

.vscode/launch.json
saintmatthieu: Oj, I didn't mean to commit that - thanks!

.github/workflows/au4_build_linux.yml
kryksyh: I find it easier to manage 3 different files, than figuring out complex conditions and the build matrix

au3/libraries/lib-builtin-effects/CompressorInstance.cpp
saintmatthieu: I don't understand.

au3/libraries/lib-builtin-effects/CompressorInstance.cpp
saintmatthieu: What do you mean by wrapping queues together?

au3/libraries/lib-dynamic-range-processor/DynamicRangeProcessorHistory.h
saintmatthieu: I think that's ok: it doesn't look unified, on the other hand it hints that this was added when integrating the effect to Au4.

au3/libraries/lib-dynamic-range-processor/DynamicRangeProcessorTypes.h
```diff
@@ -53,14 +45,17 @@ struct Unbypassed
...
+constexpr double compressorKneeWidthDbDefault = 6;
+constexpr double compressorCompressionRatioDefault = 4;
+// Setup the default makeup gain so that the transfer function yields 0 output at 0 input.
+constexpr double compressorMakeupGainDbDefault = compressorThresholdDbDefault / compressorCompressionRatioDefault
+                                                 - compressorThresholdDbDefault;
+constexpr double compressorLookaheadMsDefault = 3;
+constexpr double compressorAttackMsDefault = 3;
+constexpr double compressorReleaseMsDefault = 100;
+
+constexpr double compressorMaxLookaheadMs = 100.;
```
luapmartin, line 60: would we want to regroup compressor and limiter defaults under a struct and do we want to use custom types for db and times ? just to ensure we won’t be passing wrong types...

src/effects/builtin/CMakeLists.txt
saintmatthieu: That's one of the things I have to clean up. Keep all the test code and make it work with the framework, or discard it altogether. Haven't decided yet, but since this doesn't affect production code, I suggest we don't delay QA for this.

src/effects/builtin/dynamics/compressor/compressoreffect.h
```diff
@@ -0,0 +1,49 @@
...
+    PARAM(compressionRatio, 1, compressorCompressionRatioDefault, 20, 0.01);
+    PARAM(lookaheadMs, 0, compressorLookaheadMsDefault, compressorMaxLookaheadMs, 0.1);
+    PARAM(attackMs, 0.1, compressorAttackMsDefault, 100, 0.01);
+    PARAM(releaseMs, 10, compressorReleaseMsDefault, 1000, 0.1);
+    PARAM(showInput, 0, showInputDefault, 1, 1);
+    PARAM(showOutput, 0, showOutputDefault, 1, 1);
+    PARAM(showActual, 0, showActualDefault, 1, 1);
+    PARAM(showTarget, 0, showTargetDefault, 1, 1);
+};
+}
```
luapmartin: should we add type ?

src/effects/builtin/dynamics/compressor/CompressorView.qml
```diff
@@ -0,0 +1,171 @@
...
+                Grid {
+                    columns: 2
+                    spacing: 24
+
+                    Repeater {
+                        model: [
+                            {
+                                id: "thresholdDb",
+                                title: qsTrc("effects/compressor", "Threshold"),
+                                unit: "dB"
```
kryksyh: Units should be translatable too

src/effects/builtin/common/builtineffectsettingmodel.h
saintmatthieu: Yes, I suppose we can picture it like that, too.

au3/libraries/lib-builtin-effects/CompressorInstance.cpp
luapmartin: instance.setQueues(q1,q2,q3,q4) or something like this the idea would be to regroup them some other instance will have also queues so we could reuse some queues 
just an idea to avoid multiplying members

au3/libraries/lib-builtin-effects/CompressorInstance.cpp
luapmartin:
```cpp
if (const auto queue = slave.mCompressionGainDbQueue.lock()) {
    queue->Put(compressor.GetLastFrameStats().dbGainOfMaxInputSample);
}
```
to
```cpp
if (const auto compressionGainDbQueue = slave.mCompressionGainDbQueue.lock()) {
    compressionGainDbQueue->Put(compressor.GetLastFrameStats().dbGainOfMaxInputSample);
}
```
not really important tho

au3/libraries/lib-dynamic-range-processor/DynamicRangeProcessorTypes.h
```diff
@@ -53,14 +45,17 @@ struct Unbypassed
 
 static constexpr auto compressorMeterUpdatePeriodMs = 1000 / 30;
```
kryksyh: It is better to use a type here, because it is not clear if the integer division was intended or not, considering that the variables below are doubles.

.github/workflows/au4_build_linux.yml
luapmartin: fair point 👍

src/effects/builtin/common/valuewarper/warpingtransformer.cpp
saintmatthieu: In this case, unifying the order (i.e., `Normalized{min, mid, max}` would probably be counterproductive, because the first argument of the `Normalized` ctor is not necessarily intended to be a mid value. In fact, the ctor arguments are ordered like in `std::clamp`. For some reason, I find that the min-mid-max order makes sense for the `solveC` function.

src/effects/builtin/dynamics/timeline/meters/compressiondbmetermodel.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "libraries/lib-builtin-effects/CompressorInstance.h"
+
+namespace au::effects {
+CompressionDbMeterModel::CompressionDbMeterModel(QObject* parent)
+    : AbstractDbMeterModel{MeterValueProvider::Direction::Downwards, parent}
+{
+}
+
+float CompressionDbMeterModel::latestValue()
```
kryksyh, line 16: is it latest or lowest?

src/effects/builtin/dynamics/compressor/CompressorView.qml
luapmartin: we could have a parameter central point where we have different types with their names etc. 
audio effect have a huge tendency to be similar along effects

src/effects/builtin/dynamics/compressor/CompressorView.qml
saintmatthieu: I don't get it, could you expand a bit?

src/effects/builtin/dynamics/timeline/meters/DynamicsMeter.qml
```diff
@@ -0,0 +1,59 @@
...
+    Rectangle {
+        id: globalMaxOutline
+
+        width: root.width
+        height: upwards ? root.height - y : prv.dbToY(root.globalMax)
+        y: upwards ? prv.dbToY(root.globalMax) : 0
+
+        color: prv.almostTransparentColor
+        border.color: "transparent"
+        // border.color: root.areaColor
```
kryksyh: to remove?

src/effects/builtin/dynamics/limiter/LimiterView.qml
saintmatthieu: I don't think so: these values really only affect the view and display. Another view might want to show a longer duration, or dB step, etc.

src/effects/builtin/dynamics/limiter/LimiterView.qml
saintmatthieu: Yeah, maybe. Lemme look into this ...

saintmatthieu: ... yes there could be, but for only these few lines, the overhead of adding a new component is maybe not worthwhile.

src/effects/builtin/dynamics/limiter/LimiterView.qml
saintmatthieu: I agree, these would make sense to be in the model. Lemme see ...

src/effects/builtin/dynamics/compressor/CompressorView.qml
kryksyh: it is not neccessary, I believe that translation tools are smart enough to reuse translations if the value and context are the same

src/effects/builtin/dynamics/limiter/LimiterView.qml
saintmatthieu: Addressed.

src/effects/builtin/dynamics/timeline/profiler.h
saintmatthieu: Yes absolutely, forgot to hide it from release builds. Thanks!

saintmatthieu: ... I'll remove it altogether: I doubt it'll be needed again, and if it is it's not difficult to write.

src/effects/builtin/dynamics/compressor/CompressionCurve.qml
saintmatthieu: I don't think so: this is strongly related to visual display and loosely, if at all, related to the compressor's implementation specifics.

src/effects/builtin/dynamics/timeline/meters/abstractdbmetermodel.h
saintmatthieu: That'd require more engineering, since `MeterValueProvider` exposes `GetFiveSecMax()` rather than `GetTemporaryMax(double seconds)` and so far there has been only a requirement for one temporary maximum.

src/effects/builtin/dynamics/timeline/dynamicstimeline.h
saintmatthieu: dbMin, duration and dataPointRate are "fallback default values". But it's confusing to have these defaults, so I'll set them to 0 now and assert that they get set to some valid valued before `componentComplete()` is called.

src/effects/builtin/dynamics/timeline/timelinesourcemodel.cpp
saintmatthieu: Oy, thanks for reminding!

src/effects/builtin/internal/builtineffectsrepository.cpp
saintmatthieu: "Sweat" or "sweet" ? :D

src/effects/builtin/qml/Audacity/BuiltinEffects/BigParameterKnob.qml
saintmatthieu: I suppose we'll want our `KnobControl` to have a default value it can jump to when double-clicked on, so that other effects would benefit from it. But if I start doing this now it'll make this already monstrous PR even more monstrous. When someone files a bug about the knobs on the reverb we'll address this :innocent:

src/effects/builtin/qml/Audacity/BuiltinEffects/SettingKnob.qml
saintmatthieu: Which one?

src/effects/builtin/tests/environment.cpp
saintmatthieu: Formatting love you mean? I'd agree.

src/effects/builtin/CMakeLists.txt
saintmatthieu: I'll remove the test app now.

src/effects/builtin/tests/DynamicRangeProcessorHistory_tests.cpp
saintmatthieu: Oh yes, these ... Clean-up coming up.

au3/libraries/lib-dynamic-range-processor/DynamicRangeProcessorTypes.h
saintmatthieu: > would we want to regroup compressor and limiter defaults under a struct

This is rather old code but I do remember having put some effort in organizing all this. I'm not sure more effort would bring much more (and in fairness I'm slowly getting tired of this PR `:D`)

> do we want to use custom types for db and times ?

I'd like that, but it wouldn't be a trivial change and at this stage I just feel like crossing the final line.

src/effects/builtin/dynamics/compressor/compressoreffect.h
saintmatthieu: What type?

src/effects/builtin/dynamics/compressor/CompressorView.qml
saintmatthieu: > Units should be translatable too

Addressed, now in compressorsettingmodel.cpp

src/effects/builtin/dynamics/compressor/compressoreffect.h
luapmartin: decibel, ratio, time

src/effects/builtin/common/valuewarper/warpingtransformer.cpp
saintmatthieu: Renamed.

src/effects/builtin/qml/Audacity/BuiltinEffects/SettingKnob.qml
saintmatthieu: Clarified

src/effects/builtin/dynamics/timeline/abstractdynamicseffectinstancemodel.h
saintmatthieu: Clarified

src/effects/builtin/dynamics/compressor/compressoreffect.h
saintmatthieu: IIUC I'd have to do this for all builtin effects. Let's leave it at that at least for now, please.

src/effects/builtin/dynamics/compressor/CompressorView.qml
```diff
@@ -0,0 +1,128 @@
...
+import Muse.UiComponents
+import Audacity.Effects
+import Audacity.BuiltinEffects
+
+import "../timeline"
+
+DynamicsEffectBase {
+    id: root
+
+    property string title: qsTrc("effects/compressor", "Compressor")
```
luapmartin, line 15: `compressor.title` ?

src/effects/builtin/dynamics/compressor/compressorsettingmodel.cpp
```diff
@@ -1,37 +1,28 @@
...
+const CompressorSettingModel::LabelMap labelMap = {
+    { "thresholdDb", { muse::qtrc("effects", "Threshold"), muse::qtrc("effects", "dB") } },
+    { "makeupGainDb", { muse::qtrc("effects", "Make-up gain"), muse::qtrc("effects", "dB") } },
+    { "kneeWidthDb", { muse::qtrc("effects", "Knee width"), muse::qtrc("effects", "dB") } },
+    { "compressionRatio", { muse::qtrc("effects", "Ratio"), QString() } },
+    { "lookaheadMs", { muse::qtrc("effects", "Lookahead"), muse::qtrc("effects", "ms") } },
+    { "attackMs", { muse::qtrc("effects", "Attack"), muse::qtrc("effects", "ms") } },
+    { "releaseMs", { muse::qtrc("effects", "Release"), muse::qtrc("effects", "ms") } },
+    // We deliberately leave out the "show" parameters: these are overridden in the UI anyway.
+};
```
luapmartin, line 21: nice

src/effects/builtin/dynamics/limiter/LimiterView.qml
```diff
@@ -0,0 +1,129 @@
...
+import Audacity.Effects
+import Audacity.ProjectScene
+import Audacity.BuiltinEffects
+
+import "../timeline"
+
+DynamicsEffectBase {
+    id: root
+
+    property string title: qsTrc("effects/limiter", "Limiter")
```
luapmartin, line 16: `limiter.title` ?

luapmartin (approved):

src/effects/builtin/dynamics/compressor/CompressorView.qml
saintmatthieu: We could probably indeed get `BuiltinEffectModel` to have a virtual method to get the title, but it's a bit late at this stage in the PR - in theory QA would have to retest all builtin effects (and I'd have to put yet more time in this PR :))

src/effects/builtin/dynamics/limiter/LimiterView.qml
saintmatthieu: https://github.com/audacity/audacity/pull/9356#discussion_r2373377519

src/effects/builtin/dynamics/compressor/compressorsettingmodel.cpp
saintmatthieu: Isn't it?

PR #9367, opened 2025-09-13
Compile time warning fixes

kryksyh:
Fixes compile time warnings

Recommended:

QA:

### DIFF
--- a/buildscripts/cmake/SetupDependencies.cmake
+++ b/buildscripts/cmake/SetupDependencies.cmake
@@ -25,7 +25,7 @@ else()
     set(LIB_BUILD_TYPE "debug")
 endif()
 
-set(REMOTE_ROOT_URL https://raw.githubusercontent.com/musescore/muse_deps/main)
+set(REMOTE_ROOT_URL https://raw.githubusercontent.com/kryksyh/muse_deps/main)
 set(LOCAL_ROOT_PATH ${FETCHCONTENT_BASE_DIR})
 
 function(populate name remote_suffix)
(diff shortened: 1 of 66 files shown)


buildscripts/cmake/SetupDependencies.cmake
```diff
@@ -25,7 +25,7 @@ else()
     set(LIB_BUILD_TYPE "debug")
 endif()
 
-set(REMOTE_ROOT_URL https://raw.githubusercontent.com/musescore/muse_deps/main)
+set(REMOTE_ROOT_URL https://raw.githubusercontent.com/kryksyh/muse_deps/main)
```
luapmartin, line 28: commenting here in case this is a typo

src/appshell/internal/appshellconfiguration.cpp
```diff
@@ -172,6 +172,7 @@ bool AppShellConfiguration::isNotationNavigatorVisible() const
 
 void AppShellConfiguration::setIsNotationNavigatorVisible(bool visible) const
 {
+    UNUSED(visible);
```
luapmartin, line 175: shouldn't we prefer `[[maybe_unused]]` ?

buildscripts/cmake/SetupDependencies.cmake
kryksyh: It is temporary change since the changes to muse_deps are not merged yet

src/appshell/internal/appshellconfiguration.cpp
kryksyh: UNUSED is a framework macro that is used everywhere

PR #9417, opened 2025-09-18
wxWidgets 3.2 API compatibility

cjmayo:
I tried compiling Audacity 3.7.5 against wxWidgets 3.3.1 and it revealed these two issues.

### Add missing #include <algorithm> to WaveChannelViewConstants

For std::sort() and std::adjacent_find()

### wxTimerEvent::wxTimerEvent() not available in wxWidgets 3.2 API

default ctor creates an unusable event object and should not be used
(in fact, no code outside wxWidgets is supposed to create event objects)https://github.com/wxWidgets/wxWidgets/commit/9e2ff111a4 ("Restore, but deprecate, default ctor of wxTimerEvent", 2019-10-11)

----------

Recommended:

QA:

### DIFF
--- a/libraries/lib-wave-track/WaveChannelViewConstants.cpp
+++ b/libraries/lib-wave-track/WaveChannelViewConstants.cpp
@@ -8,6 +8,8 @@ Paul Licameli split from class WaveTrack
 
 **********************************************************************/
 
+#include <algorithm>
+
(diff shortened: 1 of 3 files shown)


kryksyh:
@dozzzzer @chinakov please check that this dialog is working as expected 
image

Issue #9041, opened 2025-07-09
Moving clips become progressively slower the more effects there are in the stack.


dozzzzer:
### Steps to reproduce

1. Have 30-40 effects in the track's effect stack
2. Have a clip on that track
3. Start moving the clip around
4. Observe

### Expected behavior

As soon as the LMB is released, the clip changes its state from being dragged to released.

### Actual behavior

There's a significant delay between the change of states when you release the LMB:

https://github.com/user-attachments/assets/489fd94b-51ae-4dcf-8788-7fed8c53bdb9

### Audacity Version

master

### Operating system

OS independent

chinakov:
This is still reproducible on the latest Master (8696ccf6c1768c39515fb057d8f12257fe4d3706)

chinakov:
This is still reproducible on the latest Master (f7034596fe9f5df9d605c52ddc4196ffee3a15b0)

dozzzzer:
1. Moving a clip within the track is quick.
2. Moving the clip from the track with lots of effects to an empty track feels a tiny bit laggy, although miles better than it used to be.

PR #9420, opened 2025-09-19
Effect menu optimization

kryksyh:
Resolves: #9041

The realtime effect button menu is rebuilt when track is recreated, and the track is recreated at every clip movement. Until this behavior is improved (i.e. detecting that the view could be reused) we need to make the menu build process is fast enough

This PR makes the menu build process roughly 4x faster, not great, but good enough for now.

Recommended:

QA:

- [x] Please test that the effects menu is not affected: real-time effects could be added, removed, replaced, etc.

### DIFF
--- a/src/effects/effects_base/effectsutils.cpp
+++ b/src/effects/effects_base/effectsutils.cpp
@@ -14,12 +14,23 @@ using namespace au::effects;
 class CiString : public String
 {
 public:
-    bool operator<(const CiString& other) const
+    CiString() = default;
+    CiString(const String& s)
+        : String(s), m_cache(toLower(s)) {}
+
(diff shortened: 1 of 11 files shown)


saintmatthieu (approved):

src/effects/effects_base/effectsutils.cpp
```diff
@@ -14,12 +14,23 @@ using namespace au::effects;
...
+    CiString(const String& s)
+        : String(s), m_cache(toLower(s)) {}
+
+    bool operator<(const CiString& other) const noexcept
+    {
+        return m_cache < other.m_cache;
+    }
+
+private:
+    std::string m_cache;
```
saintmatthieu, line 27: Could be `const`.

PR #9437, opened 2025-09-23
9435 Update meters at a fixed rate despite portaudio

### LINKED ISSUE #9435: Update meters at a fixed rate
### Description

The meter is updated every time `audacityAudioCallback` is called.
The problem is we open the stream with `paFramesPerBufferUnspecified` parameters. The number of frames and the rate at which the meter is updated can vary significantly.
This can be easily verified by enabling audio input monitoring.

embarc-gabriel:
Resolves: #9435 

Make the meter update rate independent from portaudio callback execution rate.

Recommended:

QA:

### DIFF
--- a/src/au3wrap/internal/au3audiometer.cpp
+++ b/src/au3wrap/internal/au3audiometer.cpp
@@ -12,11 +12,23 @@
 #include "libraries/lib-utility/MemoryX.h"
 
 namespace {
-constexpr double METER_SAMPLE_PERIOD = 0.10;
+constexpr double METER_SAMPLE_PERIOD = 1 / 30.0;
 constexpr double MAX_ALLOWED_TIMESTAMP_DIFF = 1.0;
+
+auto calculateCoef()
+{
+    // Calibrate so that the meter falls by -48dB in 1 second.
+    constexpr auto minus48dB = 0.003981071705534973f;
+    return 1 - powf(minus48dB, METER_SAMPLE_PERIOD);
+}
 }
 
 namespace au::au3 {
+Meter::Meter()
+    : m_smoothingCoef{calculateCoef()}
+{
+}
+
 Meter::Sample Meter::getSamplesMaxValue(const float* buffer, size_t frames, size_t step)
 {
     auto sptr = buffer;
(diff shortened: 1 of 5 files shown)


src/au3wrap/internal/au3audiometer.cpp
```diff
@@ -26,18 +31,39 @@ IMeterSender::Sample GetAbsValue(const float* buffer, size_t frames, size_t step
...
+    }
+
+    const auto value = getSamplesMaxValue(sampleData.buffer, sampleData.frames, sampleData.nChannels);
+    if (value.peak > m_maxValue[channel].peak) {
+        m_maxValue[channel] = value;
+    }
+
+    const double lastSampleTimestamp = sampleData.firstSampleTimestamp + sampleData.frames * (1 / m_sampleRate);
+    if (lastSampleTimestamp > *m_lastSampleTimestamp[channel] + METER_SAMPLE_PERIOD) {
+        // Send data every METER_SAMPLE_PERIOD seconds approximately.
```
embarc-gabriel, line 71: @saintmatthieu I know it is not the same rate, but I think it is close enough. It isn't worth it to make it even closer if the user won't be able to perceive it in my opinion.

src/au3wrap/internal/au3audiometer.h
```diff
@@ -31,7 +40,13 @@ class Meter : public IMeterSender, public muse::async::Asyncable
...
     };
 
+    Sample getSamplesMaxValue(const float* buffer, size_t frames, size_t step);
+    void push(uint8_t channel, const Sample& sample, int64_t key);
+
+    double m_sampleRate{ 44100.0 };
     std::vector<Data> m_trackData;
     std::map<int64_t, muse::async::Channel<au::audio::audioch_t, au::audio::MeterSignal> > m_channels;
+    std::map<int64_t, Sample> m_maxValue{};
+    std::map<int64_t, std::optional<double> > m_lastSampleTimestamp{};
```
saintmatthieu: It seems to me that the value type of this map could be just `double` ?

src/au3wrap/internal/au3audiometer.h
```diff
@@ -10,19 +10,28 @@
...
+    void setSampleRate(double rate) override;
+
     muse::async::Channel<au::audio::audioch_t, au::audio::MeterSignal> dataChanged(int64_t key = MASTER_TRACK_ID);
 
 private:
+    struct Sample
+    {
+        float peak;
+        float rms;
+    };
```
saintmatthieu, line 37: Minor: why not defaulting these to 0.f? It'd make it safer and also you could use the default ctor in the source file.

saintmatthieu (approved):

PR #9448, opened 2025-09-24
9132 speed up destructive effects

### LINKED ISSUE #8878: UI blocked while stretch-rendering large clips
### Bug description

Regression introduced by https://github.com/audacity/audacity/pull/8749. While the processing is ongoing, the main thread is blocked, including the pop-up of the progress dialog. The user cannot do anything until it's over, and doesn't even know when it will be over.

### Steps to reproduce

1. Generate a large chunk of audio
2. Stretch it by some amount
3. Render the stretching

### Expected behavior

A progress dialog appears that can be canceled.

### Actual behavior

The application is frozen until the process is over, of which there is no visual indication.

### Audacity Version

master

### Operating system

OS independent

### Additional context

For now, this ticket should implement a temporary solution similar to what is used for destructive effects.
A follow-up ticket to solve it properly (and more generally the problem of data races in our popup-dialog integration) is https://github.com/audacity/audacity/issues/8877.

### LINKED ISSUE #9132: Destructive effects are slow
### Bug description

I repeatedly found that destructive effects could be noticeably slower in AU4.

This morning I just made an experiment: in both 3.7.4 and a recent AU4 master build, generate 1000s of noise, apply reverb with default settings. In AU3, it takes 1 second, in AU4, it takes 2.

Next experiment, this time on local debug builds: generate 1000s of noise, apply noise reduction with default settings. It takes about 20s on AU3 and 30s on AU4. Then I disabled the progress bar on AU4, and I achieved also 20s.
This is a one-shot experiment, it might be just luck, yet it augments the likelihood that the progress bar is responsible for this additional time.

### Steps to reproduce

1. On both AU3 and AU4 (release builds), generate the same, large amount of audio (maybe 1h)
2. Select the whole audio, apply reverb with default settings.

### Expected behavior

AU4 is not worse than AU3.

### Actual behavior

AU4 is at least 50% slower.

### Audacity Version

5983c6398fb9ebe208f254ce366d37c0938207de

### Operating system

Windows

### Additional context

Made the experiment on WIndows, haven't checked on MacOS or Linux.

saintmatthieu:
Resolves: #9132
Resolves: #8878

Introduces a helper to execute a task on a separate thread while the main thread updates the progress dialog.
Two improvements:
* the effect/generator task is not blocked by the rendering of the dialog,
* unnecessary renderings of the dialog are avoided.

This should improve performance on all operations that involve the progress bar, especially for long operations.

Recommended:

QA:
- [x] Generating audio or applying an effect is about as fast as Au3 (test on at least a 1h clip)
- [x] Generating something can be canceled
- [x] Applying an effect can be canceled
- [x] Progress dialog for effects has "Applying <effect name>..." title
- [x] Progress dialog for generators has "Generating <generator name>..." title (Maybe there a ticket for this?)
- [x] Repair effect when applied to too much data (e.g. 1s) fails and the dialog explaining the reason still appears
- [x] Cancelling an effect or getting it to fail (like Repair) does not add an undo item
- [x] Rendering pitch and speed works and is cancellable
- [x] Same when changing track format
- [x] Same when resampling a track
- [x] Copy/pasting a long stereo clip on a non-empty mono track shows the dialog and is cancellable
- [x] Same with dragging that long stereo clip

### DIFF
--- a/au3/libraries/lib-effects/Effect.cpp
+++ b/au3/libraries/lib-effects/Effect.cpp
@@ -338,7 +338,7 @@ bool Effect::TrackProgress(
 {
     auto updateResult = (mProgress
                          ? mProgress->Poll((whichTrack + frac) * 1000,
-                                           (double)mNumTracks * 1000, msg)
+                                           (double)std::max(mNumTracks, 1) * 1000, msg)
                          : BasicUI::ProgressResult::Success);
     return updateResult != BasicUI::ProgressResult::Success;
 }
--- a/src/au3wrap/internal/au3basicui.cpp
+++ b/src/au3wrap/internal/au3basicui.cpp
@@ -99,11 +99,12 @@ BasicUI::MessageBoxResult Au3BasicUI::DoMessageBox(const TranslatableString& mes
 std::unique_ptr<BasicUI::ProgressDialog> Au3BasicUI::DoMakeProgress(const TranslatableString& title, const TranslatableString& message,
                                                                     unsigned int flags, const TranslatableString& remainingLabelText)
 {
-    Q_UNUSED(title);
-    Q_UNUSED(message);
     Q_UNUSED(flags);
     Q_UNUSED(remainingLabelText);
-    return std::make_unique<ProgressDialog>();
+    auto dialog = std::make_unique<ProgressDialog>();
+    dialog->SetDialogTitle(title);
+    dialog->SetMessage(message);
+    return dialog;
 }
 
 namespace {
--- a/src/au3wrap/internal/progressdialog.cpp
+++ b/src/au3wrap/internal/progressdialog.cpp
@@ -6,11 +6,17 @@
 
 #include "progressdialog.h"
 
-ProgressDialog::ProgressDialog()
+ProgressDialog::ProgressDialog(const std::string& title)
+    : m_progressTitle{title}
 {
     // Of course, the least number of increments to yield a smooth animation depends on the width of the progress bar,
     // yet 300 increments should be enough to provide a smooth animation in most cases.
     m_progress.setMaxNumIncrements(200);
+}
+
+ProgressDialog::ProgressDialog(const TranslatableString& title)
+    : ProgressDialog{title.Translation().ToStdString()}
+{
 }
 
 ProgressDialog::~ProgressDialog()
--- a/src/au3wrap/internal/progressdialog.cpp
+++ b/src/au3wrap/internal/progressdialog.cpp
@@ -24,13 +24,13 @@ void ProgressDialog::Reinit()
 
 void ProgressDialog::SetDialogTitle(const TranslatableString& title)
 {
-    Q_UNUSED(title);
+    m_progressTitle = title.Translation().ToStdString();
 }
 
 ProgressResult ProgressDialog::Poll(unsigned long long numerator, unsigned long long denominator, const TranslatableString& message)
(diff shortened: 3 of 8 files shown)


src/au3wrap/au3wraputils.cpp
```diff
@@ -0,0 +1,98 @@
...
+                    m_museProgress.finished().send(muse::make_ret(errorCode));
+                } else if (!m_museProgress.isCanceled()) {
+                    m_museProgress.finished().send(muse::make_ok());
+                }
+            } catch (::AudacityException& e) {
+                std::string message;
+                if (const auto box = dynamic_cast<::MessageBoxException*>(&e)) {
+                    message = box->ErrorMessage().Translation().ToStdString();
+                }
+                m_museProgress.finished().send(muse::make_ret(errorCode, message));
```
kryksyh: If there will be non audacity exception we will crash / loop never exit.

src/au3wrap/au3wraputils.cpp
```diff
@@ -0,0 +1,98 @@
...
+            }
+        });
+
+        loop.exec();
+        thread.join();
```
kryksyh: The name of the class is indeed a bit misleading, this progress blocks the execution flow until finished.

saintmatthieu: Mmh, yeah. "Non-blocking" refers to the refresh of the progress dialog not blocking the execution of the task. But of course the call to `execute` is synchronous.
Maybe renaming these files to `internal/au3wrapperutils.h/cpp`, and call this function `executeInBackground`, simply. It wouldn't say it all, but what it'd say would be unambiguous.

src/au3wrap/au3wraputils.cpp
saintmatthieu: I don't think it's more at risk than it was before, but sure, I suppose I can add a general catch.

kryksyh: Yeah, we do not catch all exceptions (we probably should) on the main thread either. The difference is that even if you wrap the call to `NonBlockingProgressDialog::execute` in a catch-all, the inner thread will still crash.

Since this class is a general purpose progress task runner, it should not crash on random exceptions

src/au3wrap/au3wraputils.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        std::thread thread([&] {
+            DialogImpl::Au3Progress au3Progress { m_museProgress };
+
+            try {
+                if (!task(au3Progress)) {
+                    m_museProgress.finished().send(muse::make_ret(errorCode));
+                } else if (!m_museProgress.isCanceled()) {
+                    m_museProgress.finished().send(muse::make_ok());
+                }
```
kryksyh: I'm not sure if it is possible that we skip both ifs, but for peace of mind, i'd put finished() at the end of try{}

saintmatthieu: It is, but the last branch would mean that `cancel()` was called, and this calls internally `finish()`. The `muse::Progress` API doesn't communicate this, so there's a risk that this changes in the future, but I don't feel so good about duplicating that logic and calling again `finish(make_ret(Ret::Code::Cancel))` either.

kryksyh: I see

kryksyh (approved):

Issue #9450, opened 2025-09-24
Time to open the project increased significantly


kryksyh:
### Bug description

If the project is big enough (>200 mb), it takes multiple seconds to open the project.

### Steps to reproduce

1. Create a project
2. Generate 2-3 hour long chirp
3. Save the project and exit
4. Reopen the project

### Expected behavior

The project should open instantly

### Actual behavior

It takes multiple seconds to open the project

### Audacity Version

master

### Operating system

OS independent

dozzzzer:
Tested. The time to open a 2h chirp project has decreased from 1.78s to 0.90s in my case.

PR #9454, opened 2025-09-24
Fix slow projects opening

kryksyh:
Resolves: #9450

Recommended:

QA:


src/project/internal/audacityproject.cpp
```diff
@@ -105,15 +105,6 @@ muse::Ret Audacity4Project::doLoad(const io::path_t& path, const bool forceMode,
...
-        if (!file.open(io::IODevice::ReadOnly)) {
-            LOGE() << "failed open file (Can't Read): " << path;
-            return make_ret(Err::ProjectFileIsReadProtected, path);
-        }
-        file.close();
-    }
-
     m_au3Project = au3ProjectCreator()->create();
     ret = m_au3Project->load(path);
     if (!ret) {
```
luapmartin, line 110: if I understand correctly we defer the system check for read protection to the current project container format reader here the database driver (SQLite).

src/project/tests/audacityproject_tests.cpp
```diff
@@ -141,7 +143,7 @@ TEST_F(Project_Audacity4ProjectTests, Load_FileCannotBeOpened_ReturnsReadProtect
     const muse::Ret ret = m_currentProject->load(muse::io::path_t(tempFile.getPath()), false, "");
 
     EXPECT_FALSE(ret.success());
-    EXPECT_EQ(ret.code(), static_cast<int>(Err::ProjectFileIsReadProtected));
+    EXPECT_EQ(ret.code(), SQLITE_CANTOPEN);
     EXPECT_TRUE(!ret.data<std::string>("body", std::string("")).empty());
     //can't close m_currentProject->close();
```
luapmartin, line 148: shouldn't we internally intercept those sql error to wrap them into "our" muse errors?

src/project/tests/audacityproject_tests.cpp
```diff
@@ -170,7 +172,7 @@ TEST_F(Project_Audacity4ProjectTests, Load_NonEmptyFileIsWriteProtected_ReturnsW
     const muse::Ret ret = m_currentProject->load(muse::io::path_t(tempFile.getPath()), false, "");
 
     EXPECT_FALSE(ret.success());
-    EXPECT_EQ(ret.code(), static_cast<int>(Err::ProjectFileIsWriteProtected));
+    EXPECT_EQ(ret.code(), SQLITE_READONLY);
     EXPECT_TRUE(!ret.data<std::string>("body", std::string("")).empty());
     m_currentProject->close();
```
luapmartin, line 177: shouldn't we internally intercept those sql error to wrap them into "our" muse errors?

src/project/internal/audacityproject.cpp
kryksyh: I'm not sure I undersood the quiestion

src/project/tests/audacityproject_tests.cpp
kryksyh: Yes, we can create a follow-up ticket for refactoring this. The reason I reverted some of your changes is that only `SQLITE_READONLY` was wrapped in au4 error type, and the rest of errors were shadowed with a generic DatabaseError.

src/project/internal/audacityproject.cpp
luapmartin: https://github.com/audacity/audacity/issues/9460

src/project/tests/audacityproject_tests.cpp
luapmartin: https://github.com/audacity/audacity/issues/9459

Issue #9453, opened 2025-09-24
When launching Audacity for the 1st time, no new project is created


dozzzzer:
After you launch Audacity for the 1st time and complete the onboarding, no new project is created. 
Which means:

- No audio can be recorded into a clip
- The project can be saved, but then, when you close it, it will ask to save again
- After the project is saved and closed, it will not appear in the "New & recent" tab
- After the project is saved and closed, and a new project is created, it cannot be saved (pressing `Ctrl + S` or `File > Save` has no effect) until Audacity is restarted

dozzzzer:
Tested

PR #9455, opened 2025-09-24
Project is in a buggy state after first launch setup

kryksyh:
Resolves: #9453
During the first launch setup the project view and models are created before the project is created

Correct handling of the project changed event

Recommended:

QA:

### DIFF
--- a/src/project/view/projectpropertiesmodel.cpp
+++ b/src/project/view/projectpropertiesmodel.cpp
@@ -35,23 +35,15 @@ ProjectPropertiesModel::ProjectPropertiesModel(QObject* parent)
 
 void ProjectPropertiesModel::init()
 {
-    m_project = globalContext()->currentProject();
-
     globalContext()->currentProjectChanged().onNotify(this, [this]() {
-        m_project = globalContext()->currentProject();
+        init();
     });
-
-    if (!m_project) {
-        return;
-    }
-
-    // if (project) {
-    //     m_projectMetaInfo = project->metaInfo();
-    // }
 
     thumbnailCreator()->captureThumbnailRequested().onReceive(this, [this](const muse::io::path_t& response) {
         captureThumbnail(response.toQString());
     });
+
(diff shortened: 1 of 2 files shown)


luapmartin (approved):
Hi, I am ok with those changes, though I have a little question about the place where `onNotify`, `onReceive` ... types of callback declarations should live.

I see `PlaybackToolBarTimeItem` have some in it's constructor and I kind of like that and thought it would be the _de facto_ for such types of callbacks, a standardised, expected, easy place to find them, what are your thought ?

kryksyh:
> Hi, I am ok with those changes, though I have a little question about the place where `onNotify`, `onReceive` ... types of callback declarations should live.
> 
> I see `PlaybackToolBarTimeItem` have some in it's constructor and I kind of like that and thought it would be the _de facto_ for such types of callbacks, a standardised, expected, easy place to find them, what are your thought ?

it depends on the situation. In some cases we want to do actions in synchronization with QML side, and not load anything before QML is ready (see TracksClipsView.qml:142) .  

If we set the callback in the constructor, then the loading process would start at its own time.

luapmartin:
> > Hi, I am ok with those changes, though I have a little question about the place where `onNotify`, `onReceive` ... types of callback declarations should live.
> > I see `PlaybackToolBarTimeItem` have some in it's constructor and I kind of like that and thought it would be the _de facto_ for such types of callbacks, a standardised, expected, easy place to find them, what are your thought ?
> 
> it depends on the situation. In some cases we want to do actions in synchronization with QML side, and not load anything before QML is ready (see TracksClipsView.qml:142) .
> 
> If we set the callback in the constructor, then the loading process would start at its own time.

can't QML be "stupid" and on it's initialisation "just" paint whatever is in it's View State ? 

this way callbacks update the view state whenever it's convenient 

and 

the view "just" refresh and display whatever it has to display when we actually need to boot/init/view it ?

PR #9464, opened 2025-09-25
9462 avoid start audio stream when no meter data to display

### LINKED ISSUE #9462: Avoid starting the audio stream if no metering information is shown
### Description

Now we start the audio stream if mic metering is on, but we do not take into account whether there is a visible meter to show the data.

embarc-gabriel:
Resolves: #9462

If there is no meter to show the audio data avoid starting the audio stream.

Recommended:

QA:

### DIFF
--- a/src/record/internal/au3/au3audioinput.cpp
+++ b/src/record/internal/au3/au3audioinput.cpp
@@ -30,7 +30,7 @@ Au3AudioInput::Au3AudioInput()
         }
 
         m_inputChannelsCount = audioDevicesProvider()->currentInputChannelsCount();
-        m_focusedTrackId = selectionController()->focusedTrack();
+        m_focusedTrackChannels = getFocusedTrackChannels();
 
         initMeter();
         restartMonitoring();
@@ -56,9 +56,12 @@ Au3AudioInput::Au3AudioInput()
             restartMonitoring();
         });
 
-        selectionController()->focusedTrackChanged().onReceive(this, [this](const trackedit::TrackId& trackId) {
-            m_focusedTrackId = trackId;
-            restartMonitoring();
+        selectionController()->focusedTrackChanged().onReceive(this, [this](const trackedit::TrackId&) {
+            const int focusedTrackChannels = getFocusedTrackChannels();
+            if (focusedTrackChannels != m_focusedTrackChannels) {
+                m_focusedTrackChannels = focusedTrackChannels;
+                restartMonitoring();
+            }
         });
 
         meterController()->isRecordMeterVisibleChanged().onNotify(this, [this]() {
@@ -203,6 +206,34 @@ bool Au3AudioInput::shouldRestartMonitoring() const
     return false;
 }
 
+int Au3AudioInput::getFocusedTrackChannels() const
+{
+    if (globalContext() == nullptr) {
+        return 0;
+    }
+
+    auto prj = globalContext()->currentTrackeditProject();
+    if (!prj) {
+        return 0;
+    }
+
+    const auto trackId = selectionController()->focusedTrack();
+
+    std::optional<au::trackedit::Track> track = prj->track(trackId);
+    if (!track) {
+        return 0;
+    }
+
+    int channels = 0;
+    if (track->type == au::trackedit::TrackType::Mono) {
+        channels = 1;
+    } else if (track->type == au::trackedit::TrackType::Stereo) {
+        channels = 2;
+    }
+
+    return channels;
+}
+
 void Au3AudioInput::restartMonitoring()
 {
     stopMonitoring();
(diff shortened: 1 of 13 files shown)


src/trackedit/internal/au3/au3interaction.cpp
```diff
@@ -1959,6 +1962,12 @@ bool Au3Interaction::deleteTracks(const TrackIdList& trackIds)
         trackEdit->notifyAboutTrackRemoved(track);
     }
 
+    if (muse::contains(trackIds, focusedTrack)) {
+        trackedit::ITrackeditProjectPtr trackEdit = globalContext()->currentTrackeditProject();
+        const auto trackIds = trackEdit->trackIdList();
+        selectionController()->setFocusedTrack(trackIds.empty() ? -1 : trackIds.front());
+    }
```
saintmatthieu, line 1969: Something I don't understand here: if the focused track is part of the tracks that were deleted, then shouldn't the scope of this if condition only call `setFocusedTrack(-1)` ?

src/trackedit/internal/au3/au3selectioncontroller.cpp
```diff
@@ -511,13 +511,6 @@ au::trackedit::TrackId Au3SelectionController::focusedTrack() const
...
 {
-    Au3Track* track = au3::DomAccessor::findTrack(projectRef(), ::TrackId(trackId));
-    IF_ASSERT_FAILED(track) {
-        return;
-    }
-    auto& trackFocus = TrackFocus::Get(projectRef());
-    trackFocus.SetFocus(track->shared_from_this());
-
     m_focusedTrack.set(trackId, true);
 }
```
saintmatthieu, line 515: Am I right that `TrackFocus` was in fact never used in Au4 ?

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -179,9 +194,34 @@ void Au3AudioInput::restartMonitoring()
         return;
     }
 
+    if (!audibleInputMonitoring() && configuration()->isMicMeteringOn() && !isTrackMeterMonitoring()
+        && !meterController()->isRecordMeterVisible()) {
+        return;
+    }
```
saintmatthieu, line 204: I feel stupid, but it too me a while to get my head around this.
I wrote a truth table and used it with your two if conditions, and I agree with the outcome.
Looking at the table I came with the following alternative. Somehow I can make sense of it more easily, but maybe it's just because I was the one who came up with it. Please tell me your impression:
```cpp

// If we must playback the microphone, then of course the audio device must be active.
if (audibleInputMonitoring) {
    return true;
}

// No input monitoring. If we don't need microphone meters either, then no need for active audio device.
if (!isMicMeteringOn) {
    return false;
}

// No playback of the microphone but microphone levels were requested. If either the track meter is active or the record meter is visible, we need the device.
if (isTrackMeterMonitoring || isRecordMeterVisible) {
    return true;
}

return false;
```

saintmatthieu (approved):

src/trackedit/internal/au3/au3interaction.cpp
embarc-gabriel: oh sorry. I should have removed the focused track from the TrackIds. I'll make sure to fix that.
I cannot only do `setFocusedTrack(-1)' because what I want is to ensure that if I remove the focused track, I should assign the focus to another track. Probably the first one on the list.
That's why we have the `empty` call.
Anyway, I'll improve it and maybe add some comments.

src/trackedit/internal/au3/au3selectioncontroller.cpp
embarc-gabriel: I took a look on the code and I haven't found any reference to it.
But I'll check it again.

src/record/internal/au3/au3audioinput.cpp
embarc-gabriel: Yeah .. it is way better.
It probably isn't very clear because I have added conditions one after the other, and it has somehow become incomprehensible. 
Thank you!

src/trackedit/internal/au3/au3interaction.cpp
saintmatthieu: I see, but here, shouldn't all `trackIds` have been deleted with the `tracks.Remove(*au3Track);` line 1959 ? Then `tracksIds.empty()` would still be false, but the corresponding tracks would be deleted, so how could the focus be set on a deleted track?

embarc-gabriel: Sorry, I understood the doubt now.
The trackIds variable was shadowed by the assignment on line 1967 in that code block.
I have renamed the variable to give a proper name.
On line 1968 trackIds contains the list of renaming tracks after the deletion.

src/record/internal/au3/au3audioinput.cpp
embarc-gabriel: I reviewed the code. The problem is that it was not returning a bool but just returning from the function.
That's the reason why it is way more complicated than your code.
But I have created an auxiliary function.
Now it is clear.

src/trackedit/internal/au3/au3selectioncontroller.cpp
embarc-gabriel: Hey! It is not being used.

PR #9471, opened 2025-09-25
Fix z-ordering of the timeline, play region and the playhead

kryksyh:
Fixes timeline notches being drawn on top of the playhead

Recommended:

QA:


src/projectscene/qml/Audacity/ProjectScene/clipsview/Timeline.qml
```diff
@@ -60,14 +60,17 @@ Rectangle {
...
+        context: timelineContext
+    }
+
     TimelineRuler {
         id: timelineRuler
 
         anchors.fill: parent
 
-        // Ensure the ruler is drawn on top of selection and play region
-        z: 42
```
grliszas14, line 69: :D

Issue #9163, opened 2025-07-30
Default Sample Rate for Export is 8kHz


chinakov:
### Bug description

If the User selects a new format within the export dialog, the default sample rate will be selected as 8000 Hz. This makes the User change the default sample rate to something reasonable, like 44.1k or 44k each time a User selects a new format. 

### Steps to reproduce

1. Launch AU4
2. Generate some audio
3. Export
4. Change the format of the export 
5. Observe result

### Expected behavior

The sample rate should be retained after each format change. 

### Actual behavior

The sample rate defaults to lowest value (8000hz)

### Audacity Version

0140c97876cbb811d1291ea14de985b23fb7d0b4

### Operating system

Windows

### Additional context

https://github.com/user-attachments/assets/1ca659b2-03ba-473b-a601-3d12fd29a7b0

chinakov:
This is still reproducible on the latest Master (6a2ccc6bae123089f852ee22b47c819118b94211)

chinakov:
This is still reproducible on the latest Master (3a5e9510e24579f118f4601854bae366853fd9a6)

chinakov:
The sample rate should depend on the track with the highest or the lowest sample rate track available within a project. 

This should duplicate the behaviour of AU3.

dozzzzer:
@grliszas14, there should be no default sample rate of 44.1 kHz; the default one should be equal to the highest sample rate in the project.

dozzzzer:
@grliszas14 
1. Create a new project
2. Add two tracks
3. Set the second track to be e.g., 88.2 kHz
4. File > Export audio
5. Observe: the sample rate is correctly set to 88.2 kHz, too
6. Close the dialog/press Cancel
7. Change the second track's rate to e.g. 16 kHz
8. File > Export audio
9. Observe: the sample rate is still 88.2 kHz, rather than 16 kHz

https://github.com/user-attachments/assets/6284a4cd-4611-4a15-97db-99dcf6247941

dozzzzer:
@grliszas14 
1. New project
2. Import a 48 kHz file
3. Observe: track's sample rate is set to 48 kHz accordingly
4. In the Export dialog, change the rate to 44.1 kHz > Export
5. Ctrl + W
6. New project
7. Import the same file again
8. File > Export
9. Observe: the sample rate is left from the previous export and equals 44.1 kHz, not 48 kHz

https://github.com/user-attachments/assets/5f55ad50-2591-449d-8472-a47968882a58

[...]

PR #9476, opened 2025-09-26
Improve sample rate matching algo

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9163

Recommended:

QA:

### DIFF
--- a/src/importexport/export/view/exportpreferencesmodel.cpp
+++ b/src/importexport/export/view/exportpreferencesmodel.cpp
@@ -66,8 +66,20 @@ ExportPreferencesModel::ExportPreferencesModel(QObject* parent)
     exportSampleRateList();
 }
 
+ExportPreferencesModel::~ExportPreferencesModel()
+{
+    cancel();
+    if (m_resetSampleRate) {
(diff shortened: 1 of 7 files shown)


src/importexport/export/view/exportpreferencesmodel.cpp
```diff
@@ -300,19 +300,48 @@ void ExportPreferencesModel::setFilePickerPath(const QString& path)
...
+
+    auto it = std::find(sampleRateList.begin(), sampleRateList.end(), sampleRate);
+    int index = -1;
+
+    if (it != sampleRateList.end()) {
+        // expected sampleRate is available
+        index = static_cast<int>(std::distance(sampleRateList.begin(), it));
+    } else {
+        // check for sampleRate bigger than the preferred one
+        auto upper = std::upper_bound(sampleRateList.begin(), sampleRateList.end(), sampleRate);
```
saintmatthieu, line 366: This assumes that `sampleRateList` is sorted. It expect it to be, but to be on the safe side, `std::sort` beforehand.

saintmatthieu (approved):

PR #9484, opened 2025-09-26
follow-up improving playback flow

luapmartin:
## Follow-up of https://github.com/audacity/audacity/pull/9410

FIXES:
- While the playback is playing you can now:
    - position the playhead for next playback to start (you don’t need anymore to pause or stop for that to happen like in au3)
    - Create a range selection that will be playing on the next playback start  (you don’t need anymore to pause or stop for that to happen like in au3)
        - This was hanging the playback if done during play and needed stop to make it work 
    - Selecting a clip will create a new selection and on pause or stop the playhead will jump to it’s beginning and next playback will play it
        - Before we needed to be stopped or pause for the next playback to play the new selection (and the playhead wasn’t jumping at the beginning of the selection)
    - 
- When playback reaches the end of either the project or the region the playhead will jump back at the next playback start position 

NEED to FIX:
- [x] If in a paused state selecting a clip won’t make next playback start from clip beginning
- [x] If on a selected clip playing paused and selecting another clip will crash
- [x] If during playing user create is selecting region then pause then play/resume should playback jump to new selection : au3 YES
    - [ ] If during playing a selection user click on the timeline pause play won’t jump there until selection finishes playing
    - [ ] If during playing a selection user make a selection on the timeline pause play won’t jump there until selection finishes playing
    - [ ] If during playing user click on the timeline pause play won’t jump there 
    - [ ] If during playing user make a selection on the timeline pause play won’t jump there 
- [x] If there is an existing selection 
    - [x] Cursor update on mouse click down when inside selection but on mouse up when outside selection
- [x] Rewind backward should go to 0.0 and loose selection
- [x] Rewind forward should go to end of project and loose selection
- [x] When not playing: Clicking at the right of a existing selection do not register correctly the clicks
- [x] When looping, Playback sometimes randomly stops looping at the end of a loop.

---

Recommended:

QA:

### DIFF
--- a/src/playback/internal/au3/au3player.cpp
+++ b/src/playback/internal/au3/au3player.cpp
@@ -385,7 +385,6 @@ void Au3Player::setLoopRegion(const PlaybackRegion& region)
     Au3Project& project = projectRef();
     auto& playRegion = ViewInfo::Get(project).playRegion;
 
-    playRegion.SetActive(true);
     playRegion.SetAllTimes(region.start, region.end);
 
     m_loopRegionChanged.notify();
--- a/src/playback/internal/au3/au3player.cpp
+++ b/src/playback/internal/au3/au3player.cpp
@@ -223,6 +223,7 @@ void Au3Player::seek(const muse::secs_t newPosition, bool applyIfPlaying)
     auto& playRegion = ViewInfo::Get(project).playRegion;
     if (!playRegion.Active()) {
         playRegion.SetStart(pos);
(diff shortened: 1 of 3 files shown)


luapmartin:
this PR is old and with branch on **audacity** please look at updated one with branch on my **fork** https://github.com/audacity/audacity/pull/9692

luapmartin:
this has been merged already from the branch created on my fork

PR #9485, opened 2025-09-29
Project restoration fix

### LINKED ISSUE #9461: Exception when trying to open a project after crashing Audacity
1. Open an existing project
2. Add a clip (if needed)
3. Apply an effect to the clip
4. Crash Audacity
5. Start Audacity
6. Choose "No" in the recovery dialog
7. Try to open the project
8. Observe: an exception is thrown.

Note: choosing "yes" in the recovery opens the project just fine.

Image

saintmatthieu:
Resolves: #9461

The flag `ignoreAutosave` was not set to `true` when the user decided to discard unsaved data from a crash.  This set the `ProjectFileIO` in an inconsistent state and leading `Au3ProjectAccessor::hasUnsavedData()` to mistakenly return `true` although unsaved data should have been cleaned up. I didn't go follow the error path down to the bottom because other things would likely have gone wrong anyway hadn't it been for this.

Recommended:

QA:

- [x] Bug is fixed
- [x] Crashing a project with unsaved changes and then saying "yes" to the "do you want to recover" question still works

### DIFF
--- a/src/au3wrap/iau3project.h
+++ b/src/au3wrap/iau3project.h
@@ -20,7 +20,7 @@ public:
     virtual ~IAu3Project() = default;
 
     [[nodiscard]] virtual muse::Ret open() = 0;
-    [[nodiscard]] virtual muse::Ret load(const muse::io::path_t& filePath) = 0;
+    [[nodiscard]] virtual muse::Ret load(const muse::io::path_t& filePath, bool ignoreAutosave = false) = 0;
     virtual bool save(const muse::io::path_t& fileName) = 0;
     virtual void close() = 0;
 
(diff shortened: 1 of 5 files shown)


luapmartin (approved):
nice, unit test is cherry on top 🍰

saintmatthieu:
Ah, right, didn't think of them ...
But now I look into this: the only thing I could realistically test would be the `Au3ProjectCreator`, making sure that it forwards the `ignoreAutosave` flag to its `IAu3Project`, and for this I'd have to refactor `Au3ProjectCreator` to be able to inject it an `IAu3Project` factory ... in this case a lot of work for little testing.

Issue #9468, opened 2025-09-25
Implement export for loop region


ianalarina:
### Description

[Figma link](https://www.figma.com/design/QaCzd2bM08W8VWgCLsKJ2u/32---Export?node-id=1758-78546&t=KymmhRBIPcnFySJo-0)

Export loop region + error if no loop region exists

grliszas14:
QA info: it doesn't support exporting tail of the project just yet (reverb,delay etc. after last project's clip)

dozzzzer:
Tested

PR #9489, opened 2025-09-29
Export loop region

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9468

Recommended:

QA:

### DIFF
--- a/src/importexport/export/view/exportpreferencesmodel.cpp
+++ b/src/importexport/export/view/exportpreferencesmodel.cpp
@@ -34,8 +34,8 @@ QString prepareExtensionsString(const QStringList& extensionList)
 std::map<ExportProcessType, std::string> EXPORT_PROCESS_MAPPING {
     { ExportProcessType::FULL_PROJECT_AUDIO, muse::trc("export", "Export full project audio") },
     { ExportProcessType::SELECTED_AUDIO, muse::trc("export", "Export selected audio") },
+    { ExportProcessType::AUDIO_IN_LOOP_REGION, muse::trc("export", "Export audio in loop region") },
     //! NOTE: not implemented yet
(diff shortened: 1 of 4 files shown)


src/importexport/export/view/exportpreferencesmodel.cpp
```diff
@@ -152,6 +158,20 @@ void ExportPreferencesModel::setCurrentProcess(const QString& newProcess)
         return;
     }
 
+    if (type == ExportProcessType::AUDIO_IN_LOOP_REGION && !playback()->player()->loopRegion().isValid()) {
+        interactive()->error(muse::trc("export", "No loop region"),
+                             muse::trc("export",
+                                       "Export audio in loop region requires an active loop in the project. Please go back, create a loop and try again."));
+        return;
+    }
```
luapmartin, line 166: isValid check is nice but maybe isActive too?

grliszas14: hmmmm, it's a question to designers/QA - in theory even if loop is not active, we still have loop region start and end which can be used in exporting

kryksyh: @luapmartin good catch!
I'd say that we should allow to export with inactive loop region, because the user already did an extra step to by selecting the export option.

Insted we should activate the loop region, so the export work as expected.
(and tweak the message a bit)

grliszas14: Surprisingly, export works as expected without any change, merging

luapmartin: @grliszas14 should we update that text though?

```cpp
"Export audio in loop region requires an active loop in the project.
Please go back, create a loop and try again."
```

PR #9498, opened 2025-09-29
Meters use a LockFreeQueue

### LINKED ISSUE #9475: Lock-free audio levels ...
At the moment, we send the microphone's meter values from the audio hardware thread to the main thread using the default framework channel mechanism, which, when used in multi-threaded context, involves a **mutex**.

This is susceptible of introducing non-negligible jitter in the audio IO callbacks, preventing users to use very-low latency settings without getting hick-ups. One video on the topic that was once recommended by @IliasBergstromMuse : https://www.youtube.com/watch?v=Q0vrQFyAdWI.

We should use a lock-free queue, an implementation of which exists in Au3.

saintmatthieu:
Resolves: #9475

Recommended:

QA:
- [x] No regression with respect to playback and recording meters
- [x] Performance is not worse than before when microphone meter display is on

### DIFF
--- a/au3/libraries/lib-audio-devices/IMeterSender.h
+++ b/au3/libraries/lib-audio-devices/IMeterSender.h
@@ -16,10 +16,17 @@ public:
         const double firstSampleTimestamp;
     };
 
+    struct TrackId {
+        TrackId() = default;
+        explicit TrackId(int64_t v)
+            : value(v) {}
+        int64_t value = -1;
+        bool operator<(const TrackId& other) const { return value < other.value; }
+    };
+
     virtual ~IMeterSender() = default;
-    virtual void push(uint8_t channel, const InterleavedSampleData& sampleData, int64_t key = MASTER_TRACK_ID) = 0;
-    virtual void sendAll() = 0;
-    virtual void reset() = 0;
-    virtual void reserve(size_t size) = 0;
+    virtual void push(uint8_t channel, const InterleavedSampleData& sampleData, TrackId = TrackId { MASTER_TRACK_ID }) = 0;
+    virtual void start() = 0;
+    virtual void stop() = 0;
     virtual void setSampleRate(double rate) = 0;
 };
--- a/src/au3wrap/internal/au3audiometer.cpp
+++ b/src/au3wrap/internal/au3audiometer.cpp
@@ -8,28 +8,79 @@
 #include "au3audiometer.h"
 
 #include "global/async/async.h"
+#include "global/log.h"
 
+#include "libraries/lib-utility/MathApprox.h"
 #include "libraries/lib-utility/MemoryX.h"
 
-namespace {
-constexpr double METER_SAMPLE_PERIOD = 1 / 30.0;
-constexpr double MAX_ALLOWED_TIMESTAMP_DIFF = 1.0;
+#include "global/types/ratio.h" // muse::linear_to_db
 
-auto calculateCoef()
-{
-    // Calibrate so that the meter falls by -48dB in 1 second.
-    constexpr auto minus48dB = 0.003981071705534973f;
-    return 1 - powf(minus48dB, METER_SAMPLE_PERIOD);
-}
-}
+#include <cmath> // std::ceil
 
 namespace au::au3 {
-Meter::Meter()
-    : m_smoothingCoef{calculateCoef()}
-{
+namespace {
+constexpr double updatePeriod = 1 / 30.0;
+constexpr auto decayDbPerSecond = 36.0f;
 }
 
-Meter::Sample Meter::getSamplesMaxValue(const float* buffer, size_t frames, size_t step)
+void Meter::decay(LevelState& level)
+{
+    constexpr float decayDb = decayDbPerSecond * updatePeriod;
+    if (level.hangover == 0) {
+        level.db = std::max(level.db - decayDb, leastDb);
+    } else {
+        --level.hangover;
+    }
+}
+
+void Meter::maybeBumpUp(LevelState& level, float newLinValue, int hangover)
+{
+    const auto newPeakDb = muse::linear_to_db(newLinValue);
+    if (level.db < newPeakDb) {
+        level.db = newPeakDb;
+        level.hangover = hangover;
+    }
+}
+
+Meter::Meter()
+{
+    constexpr int letRingMs = -1000 * leastDb / decayDbPerSecond;
+    static_assert(letRingMs > 0);
+    m_stopTimer.setSingleShot(true);
+    m_stopTimer.setInterval(letRingMs);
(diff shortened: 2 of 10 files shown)


src/au3wrap/internal/au3audiometer.cpp
```diff
@@ -8,28 +8,77 @@
...
+    if (level.hangover == 0) {
+        level.db = std::max(level.db - decayDb, leastDb);
+    } else {
+        --level.hangover;
+    }
 }
+
+void Meter::maybeBumpUp(LevelState& level, float newLinValue, int hangover)
+{
+    const auto newPeakDb = 20 * log10f(newLinValue);
```
embarc-gabriel: Nit. The framework includes helpers such as `linear_to_db`. We could argue this code should be decoupled from muse score but we it is already coupled due to channels. So anyway ... we can leave it like you prefer.

src/au3wrap/internal/au3audiometer.cpp
```diff
@@ -8,28 +8,77 @@
...
+        }
+
+        QueueItem item;
+        while (m_queue.Get(item)) {
+            auto& levels = m_trackData[item.trackId].channelLevels[item.channel];
+            maybeBumpUp(levels.peak, item.sample.peak, m_hangoverCount.load());
+            maybeBumpUp(levels.rms, item.sample.rms, m_hangoverCount.load());
+        }
+
+        for (auto& [_, trackData] : m_trackData) {
```
embarc-gabriel, line 75: Nit. If we have a slow sample rate and portaudio retrieve a reasonable number of samples maybe `m_queue` can be empty and we will send the same value again.
Maybe we can just early return if m_queue is empty

src/projectscene/view/trackspanel/trackitem.cpp
```diff
@@ -37,6 +37,8 @@ TrackItem::TrackItem(QObject* parent)
 
 TrackItem::~TrackItem()
 {
+    playback()->audioOutput()->playbackTrackSignalChanges(m_trackId).close();
```
embarc-gabriel: I don't think we need this because `TrackItem` is an `Asyncable` and the cleanup is handled in the base class destructor. I'll take a closer look at this anyway.

src/au3wrap/internal/au3audiometer.h
```diff
@@ -10,48 +10,72 @@
...
     {
-        float peak = 0.0f;
-        float rms = 0.0f;
+        TrackId trackId;
+        audio::audioch_t channel;
+        QueueSample sample;
     };
 
-    struct Data
+    static inline constexpr auto leastDb = -100.0f;
```
embarc-gabriel: Nit. We can remove `inline` here.

embarc-gabriel: Should it be uppercase or starts with m_?
I say that because I was looking for the definition inside the same file.

src/projectscene/view/trackspanel/trackitem.cpp
saintmatthieu: Setting a breakpoint in `Meter::dataChanged(` shows that `channel.onClose` isn't called when a track is deleted. 
But I didn't mean to commit this, I haven't put too much care in this. Will revert.

src/au3wrap/internal/au3audiometer.cpp
```diff
@@ -8,28 +8,77 @@
...
+    constexpr int letRingMs = -1000 * leastDb / decayDbPerSecond;
+    static_assert(letRingMs > 0);
+    m_stopTimer.setSingleShot(true);
+    m_stopTimer.setInterval(letRingMs);
+    m_stopTimer.callOnTimeout([this]() {
+        m_meterUpdateTimer.stop();
+        for (auto& [_, trackData] : m_trackData) {
+            trackData.channelLevels.clear();
+        }
+    });
```
embarc-gabriel, line 56: This was very clever. Very nice! :)

src/au3wrap/internal/au3audiometer.cpp
saintmatthieu: I forgot about that - I don't mind using it.

src/au3wrap/internal/au3audiometer.h
saintmatthieu: Not nit - if there's no need for it, then let's keep it simple. Thanks.

src/au3wrap/internal/au3audiometer.cpp
saintmatthieu: 😊

src/au3wrap/internal/au3audiometer.h
saintmatthieu: > I say that because I was looking for the definition inside the same file.

I don't understand that part.
I don't know if static variables need to be treated like members? I just need this here because I want it to be private and yet I need it for the default of the `float LevelState::db` member.

au3/libraries/lib-audio-devices/IMeterSender.h
```diff
@@ -16,10 +16,17 @@ class AUDIO_DEVICES_API IMeterSender
         const double firstSampleTimestamp;
     };
 
+    struct TrackId {
+        TrackId() = default;
+        explicit TrackId(int64_t v)
+            : value(v) {}
+        int64_t value = -1;
+        constexpr bool operator<(const TrackId& other) const { return value < other.value; }
```
embarc-gabriel: Nit. I don't think `constexpr` is needed here.

embarc-gabriel (approved):

src/au3wrap/internal/au3audiometer.cpp
saintmatthieu: Yes, it is often that the queue is empty. Nevertheless, we let the value decay on each iteration (except during hangover), and hence the values are different between iterations and must be sent.

src/au3wrap/internal/au3audiometer.h
embarc-gabriel: Yeah m_ is not ok but I would use uppercase. But that's fine .. no need to change.

au3/libraries/lib-audio-devices/IMeterSender.h
saintmatthieu: You're right. AI-generated, but I don't see the use either.

src/au3wrap/internal/au3audiometer.cpp
embarc-gabriel: Oh ... yeah you are right!

src/au3wrap/internal/au3audiometer.h
saintmatthieu: Ah. Sorry, just pushed :D

Issue #9509, opened 2025-09-30
Wavelet Transform revisited


kgramhans:
### Problem to be solved

When working with Spectrogram displays, the choice of window function balances the compromise between time and frequency resolutions on linear scales, ie same temporal and frequency resolution will be applied throughout the Spectrogram display

The wavelet transform offers a different compromise between time and frequency resolutions such that

- Frequency resolution is relative to analysis frequency 
- Time resolution is a constant number of periods for analysis frequency

As such, the wavelet is much closer related to analysis tools conventionally applied to the analysis of sounds

Basically, the wavelet transform may be presented completely analogous to the existing Spectrogram display

### Your idea

A minor modification to the Audacity GUI will allow for a basic introduction of wavelet transform into the software. Such changes must of course be reflected in implementation code as well

I have played around with a simple implementation which is based on a simple extension of the "Spectrograms" preferences:

<img width="399" height="136" alt="Image" src="https://github.com/user-attachments/assets/651e84fc-dd65-424f-9b90-4110838f6fc9" />

- An Algorithm choice has been added. For demo purposes, I have chosen a so-called "1/6th octave" modulated Hann window
- The "Window size" parameter is re-used to signify the maximum length of wavelet filter used to calculate the transform. As such, this parameter is not really needed, but it limits the otherwise extensive calculations implied at low but not interesting frequencies
- "Window type" and Zero padding factors are fixed at "Hann" and "1"

The necessary adaptations of the code base have been done allowing for then displaying a wavelet based spectrogram:

1. Wavelet analysis of me pronouncing the word "wavelet"

<img width="2284" height="906" alt="Image" src="https://github.com/user-attachments/assets/9fb0e5e0-b748-4b58-8c36-c307670d725e" />

2. Spectrogram with favoured frequency resolution

<img width="2291" height="914" alt="Image" src="https://github.com/user-attachments/assets/faaa01ba-0728-4a05-8463-b000e3509c17" />

3. Spectrogram with favoured time resolution

<img width="2291" height="920" alt="Image" src="https://github.com/user-attachments/assets/84c236f7-75e3-4343-ba71-a46ddf0d1f6b" />

It is not my ambition to prove any method being superior to the other. The wavelet transform offers a compromise in resolution trade-off that is different from conventional spectrogram.

### Prior art

[...]

PR #9510, opened 2025-09-30
First simple demo implementation of Spectrogram Wavelet analysis

kgramhans:
Resolves: *https://github.com/audacity/audacity/issues/9509*

Recommended:

QA:

### DIFF
--- a/libraries/lib-wave-track-settings/SpectrogramSettings.cpp
+++ b/libraries/lib-wave-track-settings/SpectrogramSettings.cpp
@@ -171,6 +171,10 @@ SpectrogramSettings::SpectrogramSettings(const SpectrogramSettings &other)
    // Do not copy these!
    , hFFT{}
    , window{}
+   , waveletsRe{}
+   , waveletsIm{}
+   , waveletSizes{}
+   , waveletMaxLength(0)
    , tWindow{}
    , dWindow{}
 {
@@ -284,6 +288,7 @@ const TranslatableStrings &SpectrogramSettings::GetAlgorithmNames()
       XO("Reassignment") ,
       /* i18n-hint: EAC abbreviates "Enhanced Autocorrelation" */
       XO("Pitch (EAC)") ,
+      XO("Wavelet (1/6 Octave Hann)") ,
    };
    return results;
 }
@@ -527,6 +532,10 @@ void SpectrogramSettings::DestroyWindows()
 {
    hFFT.reset();
    window.reset();
+   waveletsRe.reset();
+   waveletsIm.reset();
+   waveletSizes.reset();
+   waveletMaxLength = 0;
    dWindow.reset();
    tWindow.reset();
 }
@@ -585,24 +594,100 @@ namespace
       }
       for (ii = padding; ii < endOfWindow; ++ii)
          window[ii] *= scale;
-   }
+}
+
+void RecreateWavelet(size_t &size, Floats &waveletRe, Floats &waveletIm, double frequency, size_t bins)
+{
+    assert(frequency >= 0);
+    assert(frequency <= 0.5);
+    if (frequency == 0)
+    {
+        size = 0;
+        waveletRe = Floats{};
+        waveletIm = Floats{};
+        return;
+    }
+    
+    // We will then create a modulated Hann window at the given nominal frequency
+    // Window length must be chosen such that spectral power bandwidth is 1/6th of an octave
+    // This bandwidth constraint then in return governs the duration of the hann window
+    // The RMS Bandwidth of Hann window is 2* 1/T * sqrt(1/3)
+    // Formula to obtain duration is thus
+    // RMSBW = 2 * sqrt(1/3) / T
+    // T = 2 * sqrt(1/3) / RMSBW
+    // RMSBW = 1/6th octave, ie factor 2^0.16667 = 1.122 between bands
+    // T = 2 * sqrt(1/3) / (sqrt(1.122) * Fc - Fc / sqrt(1.122)) = 2 * sqrt(1/3) / (sqrt(1.122) - 1/sqrt(1.122) / Fc = 10,02 / Fc
+    const double BandsPerOctave = 6.0;
+    const double bwFactor = pow(2, 1.0/BandsPerOctave);
+    const double durationFactor = 2 * sqrt(1.0/3.0) / (sqrt(bwFactor) - sqrt(1/bwFactor));
+    const double PI = 4.0 * atan(1);
+    const double omega = 2 * PI * frequency;
+
+    double waveletDurationHalf = 0.5 * durationFactor / frequency; // float number of samples for half-length. Hann enveloped filter will be zero outside range [-waveletDurationHalf; waveletDurationHalf]
+    size = (int) (waveletDurationHalf); // truncate to integer number of samples. "size" will hold the half-length of the wavelet filter (conjugate symmetry)
+    if (size > bins)
+    {
+        // There is no point in using wavelets that exceed resolution in our frequency array (under sampling, really)
+        size = 0;
+        waveletRe = Floats{};
+        waveletIm = Floats{};
+        return;
+    }
+
+
+    waveletRe = Floats{size};
+    waveletIm = Floats{size};
+    
+    for (int i = 0; i < size; i++)
+    {
+        double relative_dt = i / waveletDurationHalf;
+        double envelope = (1 + cos(relative_dt * PI)) / (waveletDurationHalf * 2.0); // Normalize for zero gain when doing convolution
+        waveletRe[i] = envelope * cos(omega * i);
+        waveletIm[i] = envelope * sin(omega * i);
+    }
+    
+    return;
+}
 }
 
 void SpectrogramSettings::CacheWindows()
 {
-   if (hFFT == NULL || window == NULL) {
-
-      double scale;
-      auto factor = ZeroPaddingFactor();
-      const auto fftLen = WindowSize() * factor;
-      const auto padding = (WindowSize() * (factor - 1)) / 2;
-
-      hFFT = GetFFT(fftLen);
-      RecreateWindow(window, WINDOW, fftLen, padding, windowType, windowSize, scale);
-      if (algorithm == algReassignment) {
-         RecreateWindow(tWindow, TWINDOW, fftLen, padding, windowType, windowSize, scale);
-         RecreateWindow(dWindow, DWINDOW, fftLen, padding, windowType, windowSize, scale);
-      }
+   if (algorithm == algWavelet)
+   {
+       if (waveletsRe == NULL || waveletsIm == NULL)
+       {
+           size_t num = NBins();
+           waveletsRe = FloatBuffers {num};
+           waveletsIm = FloatBuffers {num};
+           waveletSizes = ArrayOf<size_t> {num};
+           waveletMaxLength = 0;
+           double delta = 0.5 / num;
+           for (int i = 0; i < num; i++)
+           {
+               RecreateWavelet(waveletSizes[i], waveletsRe[i], waveletsIm[i], i * delta, num);
+               if (waveletSizes[i] > waveletMaxLength)
+               {
+                   waveletMaxLength = waveletSizes[i];
+               }
+           }
+       }
+   }
+   else
+   {
+       if (hFFT == NULL || window == NULL) {
+           
+           double scale;
+           auto factor = ZeroPaddingFactor();
+           const auto fftLen = WindowSize() * factor;
+           const auto padding = (WindowSize() * (factor - 1)) / 2;
+           
+           hFFT = GetFFT(fftLen);
+           RecreateWindow(window, WINDOW, fftLen, padding, windowType, windowSize, scale);
+           if (algorithm == algReassignment) {
+               RecreateWindow(tWindow, TWINDOW, fftLen, padding, windowType, windowSize, scale);
+               RecreateWindow(dWindow, DWINDOW, fftLen, padding, windowType, windowSize, scale);
+           }
+       }
    }
 }
 
@@ -648,7 +733,7 @@ size_t SpectrogramSettings::GetFFTLength() const
 //#ifndef EXPERIMENTAL_ZERO_PADDED_SPECTROGRAMS
   // return windowSize;
 //#else
-   return windowSize * ((algorithm != algPitchEAC) ? zeroPaddingFactor : 1);
+   return windowSize * ((algorithm != algPitchEAC && algorithm != algWavelet) ? zeroPaddingFactor : 1);
 //#endif
 }
 
--- a/libraries/lib-wave-track-settings/SpectrogramSettings.h
+++ b/libraries/lib-wave-track-settings/SpectrogramSettings.h
@@ -74,7 +74,7 @@ public:
    static const EnumValueSymbols &GetScaleNames();
    static const EnumValueSymbols &GetColorSchemeNames();
    static const TranslatableStrings &GetAlgorithmNames();
-
+    
    //! Return either the track's independent settings or global defaults
    //! Mutative access to attachment even if the track argument is const
    static SpectrogramSettings &Get(const WaveTrack &track);
@@ -177,7 +177,7 @@ public:
       algSTFT = 0,
       algReassignment,
       algPitchEAC,
-
+      algWavelet,
       algNumAlgorithms,
    };
    Algorithm algorithm;
@@ -198,6 +198,10 @@ public:
    // Variables used for computing the spectrum
    HFFT           hFFT;
    Floats         window;
+   FloatBuffers   waveletsRe;
+   FloatBuffers   waveletsIm;
+   ArrayOf<size_t>   waveletSizes;
+   size_t         waveletMaxLength;
 
    // Two other windows for computing reassigned spectrogram
    Floats         tWindow; // Window times time parameter
--- a/src/prefs/SpectrumPrefs.cpp
+++ b/src/prefs/SpectrumPrefs.cpp
@@ -258,7 +258,7 @@ void SpectrumPrefs::PopulateOrExchange(ShuttleGui & S)
             }
          );
 
-         S.Id(ID_WINDOW_TYPE).TieChoice(XXO("Window &type:"),
+         mWindowTypeChoiceCtrl = S.Id(ID_WINDOW_TYPE).TieChoice(XXO("Window &type:"),
             mTempSettings.windowType,
             mTypeChoices);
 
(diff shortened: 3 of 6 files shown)


kgramhans:
Purpose of this PR is mainly to share insight into extent of changes required to implement proposed functionality

saintmatthieu:
Good morning @kgramhans, thank you so much for this beautiful contribution!

I am not sure that this is going to make it in Audacity 3, as it is being sunset, but we are interested in having it in Audacity 4.

Unfortunately, the spectrum view has not yet been ported to Au4, so I can't just ask you to redirect to redirect this PR to the master branch. It's a matter of months till we have, so one option would be to just let this sit for a while and reactivate it in due time.

We will discuss this in our daily today, but in the meantime I'll begin to review it, it won't be wasted time.

libraries/lib-wave-track-settings/SpectrogramSettings.h
```diff
@@ -74,6 +74,9 @@ class WAVE_TRACK_SETTINGS_API SpectrogramSettings
    static const EnumValueSymbols &GetScaleNames();
    static const EnumValueSymbols &GetColorSchemeNames();
    static const TranslatableStrings &GetAlgorithmNames();
+    
+   static constexpr double MinimumWaveletFrequency = 20.0;
```
saintmatthieu: I don't see this constant being used anywhere. (Isn't the minimum wavelet frequency determined by the window size parameter?)

libraries/lib-wave-track-settings/SpectrogramSettings.h
```diff
@@ -74,6 +74,9 @@ class WAVE_TRACK_SETTINGS_API SpectrogramSettings
...
+   static constexpr int NumberOfWaveletsPerOctave = 12;
```
saintmatthieu: Similarly for this one: it's only referred to in a comment.
Also I'd like a bit of background on this: I thought you were using a 1/6-octave transform ; wouldn't that mean 6 wavelets per octave? Or is 1/6th the bandwidth of the passband and an overlap of 2 means 12 wavelets per octave?

libraries/lib-wave-track-settings/SpectrogramSettings.h
```diff
@@ -111,7 +114,7 @@ class WAVE_TRACK_SETTINGS_API SpectrogramSettings
 
    void InvalidateCaches();
    void DestroyWindows();
-   void CacheWindows();
+   void CacheWindows(double sampleFrequency);
```
saintmatthieu: I don't think you actually need this to calculate the transform. You can use normalized frequency instead. I've made the experiment, it looks like this:
```cpp
           const double delta = 0.5 / num;
           for (int i = 0; i < num; i++)
           {
               size_t len = RecreateWavelet(waveletSizes[i], waveletsRe[i], waveletsIm[i], i * delta, num);
```
then `RecreateWavelet` becomes `size_t RecreateWavelet(size_t &size, Floats &waveletRe, Floats &waveletIm, double frequency, size_t bins)`,
and inside,
image

Then you don't need to modify the signature of `CacheWindows`, nor that of `SpecCache::Grow`. (Passing sampling rate around may be cumbersome - if we can avoid it it'd be preferable.)

libraries/lib-wave-track-settings/SpectrogramSettings.cpp
```diff
@@ -585,24 +594,105 @@ namespace
...
+       {
+           size_t num = NBins();
+           waveletsRe = FloatBuffers {num};
+           waveletsIm = FloatBuffers {num};
+           waveletSizes = ArrayOf<size_t> {num};
+           waveletMaxLength = 0;
+           double delta = double(sampleFrequency/2) / num;
+           for (int i = 0; i < num; i++)
+           {
+               size_t len = RecreateWavelet(waveletSizes[i], waveletsRe[i], waveletsIm[i], i * delta, sampleFrequency, num);
```
saintmatthieu: I saw, by examining the implementation of RecreateWavelet, that the return value is always that which is set to waveletSizes[i]. I would therefore prefer that this function returned nothing, and then below, you'd use `if (waveletSizes[i] > waveletMaxLength)`. It'd be more informative and also would mean that the implementation of RecreateWavelet wouldn't have to synchronize the value of the `size_t &size` argument with the return value.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -315,11 +329,136 @@ bool SpecCache::CalculateOneSpectrum(
...
+         mSampleCacheHolder.emplace(
+             clip.GetSampleView(from, myLen, mayThrow));
+         floats.resize(myLen);
+         mSampleCacheHolder->Copy(floats.data(), myLen);
+         float *useBuffer = floats.data();
+         memcpy(adj, useBuffer, myLen * sizeof(float));
+      }
+
+      // Done preparing samples. Now onward with calculations
+      float * middleSample = &scratch[longestWaveletLen - 1];
```
saintmatthieu: I think this could be `const float* const middleSample = ...`. It would then propagate and you'd also need
```cpp
          const float *right = middleSample;
          const float *left = middleSample;
 ```
 When the code becomes complex, using `const` is of valuable help in understanding.

File: src/prefs/SpectrumPrefs.cpp [hunk 6]
```diff
@@ -540,6 +540,23 @@ void SpectrumPrefs::OnDefaults(wxCommandEvent &)
...
+   wxChoice *const pPaddingSizeControl =
+       static_cast<wxChoice*>(wxWindow::FindWindowById(ID_PADDING_SIZE, this));
+   wxChoice *const pAlgorithmControl =
+       static_cast<wxChoice*>(wxWindow::FindWindowById(ID_ALGORITHM, this));
+    wxChoice *const pWindowControl =
+        static_cast<wxChoice*>(wxWindow::FindWindowById(ID_WINDOW_TYPE, this));
+    wxChoice *const pWindowSizeControl =
+        static_cast<wxChoice*>(wxWindow::FindWindowById(ID_WINDOW_SIZE, this));
+
+   if (pPaddingSizeControl && pAlgorithmControl && pAlgorithmControl->GetSelection() == SpectrogramSettings::algWavelet)
```
saintmatthieu: Here I would have this condition preceded by an `assert(pPaddingSizeControl && pAlgorithmControl);`, since the implementation intends it to be always true - correct?

src/prefs/SpectrumPrefs.cpp
```diff
@@ -540,6 +540,23 @@ void SpectrumPrefs::OnDefaults(wxCommandEvent &)
...
+   {
+      pPaddingSizeControl->SetSelection(0);
+      if (pWindowControl) pWindowControl->SetSelection(eWinFuncHann);
+      if (pWindowSizeControl && !mPopulating) pWindowSizeControl->SetSelection(11);
```
saintmatthieu: To document this `11` and also make it robust against changes of the window-size `enum`, I suggest adding
```cpp
static_assert(11 < SpectrogramSettings::NumWindowSizes);
static_assert(1 << SpectrogramSettings::LogMinWindowSize + 11 == 16384);
```
somewhere close.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -315,11 +329,136 @@ bool SpecCache::CalculateOneSpectrum(
...
+   const auto samplesPerPixel = sampleRate / pixelsPerSecond / stretchRatio;
+
+   // not reassignment, xx is surely within bounds.
+   wxASSERT(xx >= 0);
+   if (xx > (int)len)
+      from = sampleCount(where[len].as_double() + (xx - len) * samplesPerPixel);
+   else
+      from = where[xx];
+
+   size_t nBins = settings.NBins();
```
saintmatthieu: Please `const` here - it's a large scope.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -315,11 +329,136 @@ bool SpecCache::CalculateOneSpectrum(
...
+
+   if (from < 0 || from >= numSamples) {
+      if (xx >= 0 && xx < (int)len) {
+         // Pixel column is out of bounds of the clip!  Should not happen.
+         float *const results = &out[nBins * xx];
```
saintmatthieu: Minor: This variable could be declared and initialized at the beginning of the method, because it's used elsewhere.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -315,11 +329,136 @@ bool SpecCache::CalculateOneSpectrum(
...
+         std::fill(results, results + nBins, 0.0f);
+      }
+   }
+   else {
+      // Iterate frequencies in turn
+      // Note that as opposed to Fourier Transform, using wavelets implies a window length that varies with frequency
+      // We extract samples corresponding to max wavelet length (the first)
+      
+      // Locate longest wavelet (first one) 
+      auto longestWaveletLen = settings.waveletMaxLength;
```
saintmatthieu: `const`

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -315,11 +329,136 @@ bool SpecCache::CalculateOneSpectrum(
    return result;
 }
 
+bool SpecCache::CalculateOneWaveletSpectrum(
```
saintmatthieu, line 332: All the preparation-of-the-samples step in this method is an almost exact duplicate of the `CalculateOneSpectrum` method. This code duplication should vanish. I would recommend the following way of doing it:
* Eliminate `CalculateOneWaveletSpectrum` ; since `CalculateOneSpectrum` takes the `SpectrogramSettings` argument, you could infer that a wavelet spectrum is desired from within,
* add two private methods, `DoCalculateOneSpectrum` and `DoCalculateOneWaveletSpectrum`, and call either of these specializations from `// Done preparing samples. Now onward with calculations`.

saintmatthieu:
Food for thoughts:
Reviewing this code has been the opportunity for me to (re)discover the implementation details of our spectrum display.
My observations:
* The frequency-bins are decoupled from the pixel rows, and then a layer assigns the value of a row based on the closest frequency bin (hence the crude aspect when using a small window size without padding)
* The hop size is **coupled** to the pixel columns. Consequences:
  * For large window (or wavelet) sizes this makes the computation inefficient, since the horizontal diff must be rather flat and interpolation could save tons of calculations.
  * The spectrum may not be invertible - for it to be, there are constraints on the hop size that are dependent on the analysis window (max hop size N for rectangular windows, N/2 for Hann, for example)

I would very much like to completely decouple the pixel grid from the spectrogram grid. It would allow us to make the spectrogram invertible, paving the way to spectral editing.

libraries/lib-wave-track-settings/SpectrogramSettings.h
kgramhans: True. This is a leftover from an initial implementation attempt. I will remove it

libraries/lib-wave-track-settings/SpectrogramSettings.h
kgramhans: Equally valid point with same explanation.
The initial idea was to use 1/6th octave Hann wavelets having 1/6th octave 3dB bandwidths. Similar to using zero-padding in conventional spectrogram for interpolation, having 12 wavelets per octave would give a smoother representation of the wavelet spectrogram. But again, now that the wavelet center frequencies are determined by the resolution of a comparable DFT, this is no longer a consideration.
This is a valid point for later consideration: Which choice of analysis parameters would make the most sense to a common user? As of now, parameters are piggy-backed onto conventional Spectrogram settings

libraries/lib-wave-track-settings/SpectrogramSettings.h
kgramhans: Very good suggestion which I will of course implement. I did not really like to modify the existing function signatures

libraries/lib-wave-track-settings/SpectrogramSettings.cpp
kgramhans: Valid point
I agree that this double book-keeping just makes it more complex
Will change

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
kgramhans: I spread some const's throughout

src/prefs/SpectrumPrefs.cpp
kgramhans: Actually, I just played it safe here since I was not sure whether this would always be true. Always a little cautious when working with GUI elements...

I am a fan of assertions and introduced those instead as suggested. Seems to work flawlessly

src/prefs/SpectrumPrefs.cpp
kgramhans: I did something similar

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
kgramhans: Sure

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
kgramhans: OK. This code was structured exactly as the regular spectrogram code. For good and for bad. I strived to make it similar, but will follow recommandation

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
kgramhans: Done

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
saintmatthieu: Preferably take a look at this other comment before you do this: https://github.com/audacity/audacity/pull/9510#discussion_r2410471574

I'm not sure that other suggestion is feasible, but if it is, then the present comment would probably be less relevant.

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
kgramhans: I will carefully look into it
Going forward a more class orientated approach would be preferable with overloaded calculation methods. But I guess this is not for now...

saintmatthieu: > Going forward a more class orientated approach would be preferable

Yes, that'd be really nice, but that would imply non-trivial refactoring that I wouldn't expect at this stage.

kgramhans: Having looked into this, I see potential for optimisation. However, I find the `CalculateOneSpectrum` method quite complicated since it mixes logic related to algorithm and buffer management. I have tried without immediate success to reuse the buffering code. 😞  
I deliberately simplified and re-wrote the buffering code for `CalculateOneWaveletSpectrum`.
I would prefer to leave it as-is for now (do not to touch `CalculateOneSpectrum`) in order not to create side-effects. In my opinion,  a dedicated re-factoring issue would be a better context for such unification

I did, however, get rid of the `PopulateWavelet` method which was really not needed

Please let me hear your opinion on this @saintmatthieu. I will push the code shortly

saintmatthieu: I think this is wise. Thank you for having looked into this from closer up.

saintmatthieu:
Hi @kgramhans, thanks for having addressed these comments, I think the integration is a bit better now.
I discussed your contribution with the rest of the team on a daily earlier this week.

We will want this feature for Audacity 4 (have you seen @Tantacrul 's [video](https://www.youtube.com/watch?v=QYM3TWf_G38&t), btw?) but we haven't ported spectrograms yet. We don't want to merge it in Au3 either because it's now only in KTLO mode, and this feature would require advertisement and a little bit of design to give it the care it deserves in the spectrogram settings.

My proposal would be to put a macro around the wavelet entry of
```cpp
const TranslatableStrings &SpectrogramSettings::GetAlgorithmNames()
{
   static const TranslatableStrings results{
      // Keep in correspondence with enum SpectrogramSettings::Algorithm:
      XO("Frequencies") ,
      /* i18n-hint: the Reassignment algorithm for spectrograms */
      XO("Reassignment") ,
      /* i18n-hint: EAC abbreviates "Enhanced Autocorrelation" */
      XO("Pitch (EAC)") ,
#ifdef WAVELET
      XO("Wavelet (1/6 Octave Hann)") ,
#endif
   };
   return results;
}
```
Then the feature would not be available to production builds but you could still showcase this with local builds and the implementation would still get built, ensuring that it doesn't go stale until it's ported to Au4.

Would you be okay with that?

kgramhans:
@saintmatthieu Hi Matt, 
I am just glad that you have embraced my proposal and of course fully respect your roadmap decisions. I will add the #define to effectively disable the wavelet stuff
I watched the video and enjoyed it. It expresses what I think are very mature thoughts on the topic of maintaining old software projects that will inevitably contain loads of technical dept (some claim that technical dept is just an acronym for code you have not written yourself). Very interesting and challenging. Seems like a lot of good decision making has been done
In case you have any topics related to signal processing or algorithms (design/review/coding/testing) I could offer assistance, just reach out

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
```diff
@@ -315,6 +329,129 @@ bool SpecCache::CalculateOneSpectrum(
...
+      {
+          if (settings.waveletSizes[iBin] == 0)
+          {
+             results[iBin] = -160.0;
+             continue;
+          }
+         
+          const float *right = middleSample;
+          const float *left = middleSample;
+          // Do the convolution
```
saintmatthieu, line 416: I'm puzzle here, sorry that I'm noticing this a bit late: you mention convolution, but what is taken really is the inner product (like it should). Are you referring to the fact that this inner product in the time domain results in a convolution in the frequency domain? Unless I'm wrong, please do clarify this comment.

saintmatthieu:
Thanks for the offer, I'll keep it in mind. Right now I can't think of anything beside migrating more of the built-in effects from Au3 to Au4, and this is more standard software engineering, it does not involve real DSP. In the future other opportunities are likely to come forth, though.

I've approved, feel free to merge, but before, please check the last comment I've added. Thanks!

saintmatthieu:
(I'm not sure you can merge, actually, let me know if it doesn't wrok.)

saintmatthieu:
@kgramhans another thing occurred to me that's of concern: the wavelet transform does not leverage FFT optimizations. AFAICS, this is the only algorithm of the four that doesn't. A quick test on my side also indeed shows that it's noticeably slower than the others. This could bother our Product guys (and me included, although I'm just an engineer), because we'd be exposing a feature that doesn't work smoothly. (I wonder if this might explain why this transform hasn't been implemented yet, maybe @SteveDaulton would know.) I expect the refresh rate during playback to be very, very slow (it's already slow for the spectrogram).

I'll raise this during our daily again on Monday. I suppose we could add a "slow" mention to the algorithm's name in the dropdown, and if we still end up deciding against it, reverting the commit would be trivial. (Please squash your commits before you do the merge, thanks.)

src/tracks/playabletrack/wavetrack/ui/SpectrumCache.cpp
kgramhans: Point taken. This comment originates from my background which much more deals with time series processing where all timestamps would need to be processed - not just particular points in time. It does not make sense here

kgramhans:
@saintmatthieu so true that the Wavelet transform is a different beast when it comes to optimisations

As of now, the current implementation will calculate the inner product for each point of time with a number of wavelets that increase in length for lower frequencies. For example, with a sample frequency of 44kHz and analysis downto 20 Hz the a rough amount of calculations for each point in time on a linear frequency axis would be (Window length N, average wavelet length W)

Octave 1: 11..22kHz W*N/2 operations
Octave 2: 5.5..11 kHz 2*W*N/4 operations 
Octave 3: 2.7..5.5kHz 4*W*N/8 operations (4x length, 1/4 number of frequencies)
etc

Totalling up to 5*W*N operations for 20-20kHz (10 octaves) audio range. The choice of window length, N, is selectable in current implementation (but naturally also limits lower frequency bound)
The standard choice of N=16384 and W=30 leads to 2.4M operations 

There is a huge potential for optimisations here. In Practice, it would not be necessary to calculate more than a constant number of transform values in each octave in order to properly represent the signal 
For the chosen wavelet, this number is roughly 12, leading to

Octave 1: W* max(N/2,12) operations
Octave 2: 2*W* max(N/4,12) operations
Octave 3: 4*W* max(N/8,12) operations
etc

Totalling up to 1024*W*12 = 0.4M operations plus the expense of a simple interpolation scheme

These considerations are highly linked to the way Spectrograms are designed into Audacity: A full transform is calculated at all frequencies irrespective of what is being displayed. And the choice of Y axis also does not reflect what is being calculated

This is in general a demanding design task. Adding also the complexity of real time behaviour does not make things simpler. With respect to the latter, it would be beneficial to implement wavelet analysis as a filter bank with downsampling a data for each octave being processed. The usage of causal filters could also prove useful. Such real time implementation and adequate  interpolation schemes would according to my experience end up having twice the computational load of an FFT based algorithm like the conventional spectrogram

But underlying all this would be a proper design, first of all. And understanding the different use cases

The current proposal has not focussed on any optimisations. I think that would be premature

saintmatthieu:
Yes: at the moment, the implementation calculates a new value for each requested time/frequency coordinates, and these don't match the Wavelet transform,

Even for the regular spectrogram, the time resolution Audacity asks of one spectrum per pixel column is often unnecessarily high. Normally, for a typical hop size of 1024 samples at 48kHz, only 46 analyses per second would be required. At my current zoom level, I have 681 pixels per second... The transition to Au4 may be a good opportunity to rectify this.

Anyway, we've made good progress here for the Wavelet transform. Let's merge what we have. I probably should redirect your PR to the audacity3 branch. Thank you!

saintmatthieu:
@kgramhans looks like I've messed up: I rebased and squashed your commits into one but somehow this got rid of the amendments you made (`XO("Wavelet (1/6 Octave Hann)") ,` isn't in the `#ifdef WAVELET`, the comment still says `// Do the convolution`, etc.). Please don't erase your local changes :) and open a new PR with them. Sorry about that.

kgramhans:
Hi Matt

I will do that end of this week since I am currently away on Vacation. I
will be back!

man. 13. okt. 2025, 03.19 skrev Matthieu Hodgkinson <
***@***.***>:

kgramhans:
@saintmatthieu 
Please have a look here : https://github.com/audacity/audacity/pull/9686

PR #9511, opened 2025-09-30
Effects menu provider

saintmatthieu:
Follows-up on @kryksyh's #9420
The realtime effect button menu is rebuilt when track is recreated, and the track is recreated at every clip movement. That previous PR optimized the code that was executed to rebuild these menus.

This PR now
* delays the creation of these menus until the user opens it
* caches the results until something relevant changed in the app, like the effect-menu-grouping setting or the list of available effects

Modifying tracks now do not trigger any effect-menu rebuild at all.

Recommended:

QA:
This is a refactoring PR. Behavior is not expected to change. Effect menus are expected to work like before (including checkmarks):

- [x] the add-effect button,
- [x] the drop-down on each realtime effect item
- [x] the destructive-effect menu
- [x] the generator menu.

A performance improvement is expected for projects with many tracks and realtime effects when modifying tracks by e.g. copy/pasting or dragging.


src/effects/builtin/dynamics/timeline/compressor-graph.md
```diff
@@ -0,0 +1,30 @@
+```mermaid
```
kryksyh: This guy got here by mistake

kryksyh (approved):

chinakov:
This is done, thanks. Tested for regressions, none found.

Issue #9403, opened 2025-09-17
Recording should honor Playhead Cursor


chinakov:
### Bug description

If the User places a cursor anywhere along the project, the recording will still only begin at the end of any given clip on any given track. This shouldn't happen. 

### Steps to reproduce

1. Launch AU4
2. Create a project
3. Generate some audio clips
4. Select a track with the clips
5. Place the playhead before the end of the clip
6. Press record
7. Observe result 

### Expected behavior

The recording starts at the position of the playhead cursor.

### Actual behavior

The recording starts at the end of the clip. 

### Audacity Version

26dbdbbabaeb7ec57a08398761aa07ace76f1754

### Operating system

OS independent

### Additional context

https://github.com/user-attachments/assets/d99ca9d7-ad3f-40ea-a8a9-0ee16462c97d

luapmartin:
what should happen to the clip(s) already present on the track? after the recording start position

- will they be pushed after the currently recorded clip
  - from the moment we start recording the entire space is getting inserted there maintaining the insert position distance with next clip (on stop the distance between the recorded clip end and the beginning of the next clip will be the same as the distance from the original playhead position)
  - or only when the recorded clip end would be touching / reaching the start time of next clip the next clips (the rest of the track) are moved to be “attached” after the currently recorded clip until we stop

- will it be recorded over
  - if partially you can get it back
  - if fully it’s gone forever
  - should the new recording be a new clip that once touching would be merge/glue with the next clip on stop having one clip or a slip between two clips?

- should the recording stop when touching the next clip

Tantacrul:
@luapmartin 

We discussed this at the standup today. Basically, it's the second approach. Overwrite, rather than insert.

- We always record from the position of your cursor.
- A new recording is always a new clip. We do not join with any other clips.
- We overwrite any clips to the right. 
- If we partially overwrite a clip to the right, you can use the trim feature to get overwritten audio back.
- If we fully overwrite a clip, it's gone forever.


Regarding the partial overwriting of clips to the right and using the trim feature to get overwritten audio back: if this turns out to be a technical problem, which I imagine it wont, then we discuss our options. Probably in that case, we just accept that the user can't use the trim feature to get back overwritten audio.

grliszas14:
As we debugged with @kryksyh:
[...]

Issue #7564, opened 2024-10-16
Recording creates a track but UI does not update


grliszas14:
Steps to reproduce:
1. Create an empty project
2. Press 'record'
3. Observe

What's happening: the track is created under the hood, recording is saved but UI is not updating at this point at all. To be able to see that it's true, save project, close it and open it again -> recorded clip on a new track will appear.

dozzzzer:
I think it might be the same problem I reported in another ticket https://github.com/audacity/audacity/issues/7431#issuecomment-2416100821

chinakov:
Tested, thanks!

PR #9524, opened 2025-10-01
Bring back pending tracks

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/7564
Partially resolves: https://github.com/audacity/audacity/issues/9403

This brings back pending tracks and fixes recording start so it always match playcursor position (as long as it is behind last clip).

Recommended:

QA:


src/projectscene/view/clipsview/au3/samplespainter.cpp
kryksyh, line 156: Not strictly related to your change, but fixes the sample view during recording, currently it is absolutely empty

```suggestion
        int yZero = samplespainterutils::getWaveYPos(0.0, zoomMin, zoomMax, paddedMetrics.height, dB, true, dBRange, false);
        yZero = paddedMetrics.top + std::max(-1, std::min(static_cast<int>(paddedMetrics.height + paddedMetrics.top), yZero));

        const auto samples = samplespainterutils::getSampleData(*waveClip, index, paddedMetrics, dB, dBRange, zoomMax, zoomMin);
        if (samples.size() == 0) {
            samplespainterutils::drawCenterLine(painter, waveMetrics, params.style, yZero);
            continue;
        }

        drawSampleStalk(samples, yZero, paddedMetrics, painter, params.style, params.showClipping);
```

grliszas14: Will apply later with upcoming PR, I changed it and suddenly all builds on CI fail, need to debug it

Issue #8928, opened 2025-06-13
The yellow guidelines are still displayed even after the audio generation has been undone.


dozzzzer:
https://github.com/user-attachments/assets/adee2243-600d-4628-bcb6-5fc176a1df89

dozzzzer:
Related https://github.com/audacity/audacity/issues/8926

chinakov:
This is still reproducible on the latest Master (ed3cc385c53d7b6cdd5af500362103847a9498fc)

dozzzzer:
This is fixed for Windows and Linux. MacOS cannot be verified due to https://github.com/audacity/audacity/issues/9661

Issue #9519, opened 2025-10-01
Dragging a stereo clip onto a non-empty mono track results in multiple issues


dozzzzer:
1. Move a stereo clip onto a non-empty mono track
2. Click Yes to mixdown to mono
3. Immediately after that, try to put the cursor in a different place by clicking or clicking around the newly created clip. At some point, you should be able to see that the cursor isn't moving, clicks around the clip cause the clip to move, disappear or get trimmed.

https://github.com/user-attachments/assets/1d73cfc6-bd9f-4231-b1e4-516f30273186

chinakov:
This is reproducible on the latest Master (132cb98949d500eb776b6a6a34fb3cbcc224375e)

chinakov:
This is fixed, thanks!

Issue #9491, opened 2025-09-29
Clicking below the generated clip appears to be the same as selecting the clip


dozzzzer:
https://github.com/user-attachments/assets/3b5ad605-3a93-4a27-8626-ee00d44cad9b

chinakov:
Still present on the latest Master (132cb98949d500eb776b6a6a34fb3cbcc224375e)

chinakov:
This is fixed, thanks.

PR #9525, opened 2025-10-01
ESC or losing focus cancels dragging operations

### LINKED ISSUE #8529: Moving a mono clip onto a stereo track produces a clip with only one channel; undoing causes a crash
### Bug description

If you try to move a mono clip onto a stereo track AND interrupt it by switching to a different application, it will move the clip and paint only one – left – channel in it. Undoing such an operation would result in a crash.

### Steps to reproduce

1. Have a mono clip and a stereo track in your project
2. Grab the mono clip and move it onto the stereo track, but don't release the mouse button just yet
3. Observe: the ghost clip has only left channel
4. Switch to a different window (Cmd + Tab)
5. ...and back to Audacity (Cmd + Tab)
6. Observe: the clip was put on the stereo track with only the left channel painted
7. Try to Undo

### Expected behavior

1. The ghost clip, when you move the mono clip to the stereo track, should have both left and right channels.
2. When you switch to a different app via Cmd + Tab, the move operation should be canceled. So, after going back to Audacity, you should see your clip still on the mono track.

### Actual behavior

1. The ghost clip, when you move the mono clip to the stereo track,  has only the left channel.
2. When you switch to a different app via Cmd + Tab, the move operation is performed, i.e. the clip is moved to the stereo track.
3. Lastly, undoing this causes a crash.

https://github.com/user-attachments/assets/69bb6f66-9c69-404e-93b4-cae31a071d95

### Audacity Version

master

### Operating system

OS independent

### Additional context

_No response_

### LINKED ISSUE #8747: Clip Selection persists when Audacity loses focus
### Bug description

If the user is in the process of adjusting a clip with guidelines appear, and User changes focus of the application, the clip selection will persist until the User has moved the clip again. 

### Steps to reproduce

1. Launch Audacity 4
2. Generate or import 2 audio clips
3. Align 2 clips so that guidelines are seen
4. Change focus (Alt-Tab) 
5. Go back to Audacity
6. Observe result

### Expected behavior

The clip selection should disappear when focus is switched back to AU4. 

### Actual behavior

The clip appears selected until a clip is moved that is currently aligned. 

### Audacity Version

71092f34a39b058f19da169e9038375b370953d5

### Operating system

OS independent

### Additional context

https://github.com/user-attachments/assets/c4332a84-ab98-4002-b285-a2f3c0a26f94

saintmatthieu:
Resolves: #8747
Resolves: #8529
Resolves: probably most other tickets that are related to Ctrl+Alt while dragging some clip, if there's more of them.

Recommended:

QA:
All the following actions are canceled when either pressing escape or changing focus to another app, and if any of these actions started auto-scroll, auto scroll should automatically stop and the viewport should be replace to where it was at drag start:

- [x] Left/Right Trim/Stretch
- [x] Dragging a clip on one same track
- [x] Dragging a clip on another track
- [x] Same but for multiple clips at a time (if applicable)

Besides:
- [x] Snipping tool still works
- [ ] Doesn't re-introduce #9604, #9519, #9491 or #9631

Besides, the usual:

### DIFF
--- a/src/projectscene/internal/projectsceneuiactions.cpp
+++ b/src/projectscene/internal/projectsceneuiactions.cpp
@@ -16,6 +16,10 @@ using namespace muse::ui;
 using namespace muse::actions;
 
 static UiActionList STATIC_ACTIONS = {
+    UiAction("escape",
+             au::context::UiCtxProjectFocused,
+             muse::shortcuts::CTX_PROJECT_FOCUSED
+             ),
     UiAction("automation",
              au::context::UiCtxUnknown,
              au::context::CTX_PROJECT_OPENED,
--- a/src/projectscene/internal/projectsceneuiactions.cpp
+++ b/src/projectscene/internal/projectsceneuiactions.cpp
@@ -16,10 +16,6 @@ using namespace muse::ui;
 using namespace muse::actions;
 
 static UiActionList STATIC_ACTIONS = {
-    UiAction("escape",
-             au::context::UiCtxProjectFocused,
-             muse::shortcuts::CTX_PROJECT_FOCUSED
-             ),
     UiAction("automation",
              au::context::UiCtxUnknown,
              au::context::CTX_PROJECT_OPENED,
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -44,17 +44,25 @@ Rectangle {
 
     property int interactionState: TracksClipsView.State.Idle
 
+    MouseHelper {
+        id: mouseHelper
+    }
+
     QtObject {
         id: prv
 
         property bool playRegionActivated: false
-        property bool releasedThroughCancel: false
 
         function cancelClipDragEdit() {
+            if (mainMouseArea.pressed) {
+                // This will lead to a cancel signal on `mainMouseArea` that will call back into this function,
+                // but this time in released state.
+                mouseHelper.callUngrabMouseOnItem(mainMouseArea)
+                return
+            }
             if (root.hoveredClipKey) {
                 tracksClipsView.cancelClipDragEditRequested(root.hoveredClipKey)
             }
-            releasedThroughCancel = true
         }
     }
 
@@ -404,8 +412,6 @@ Rectangle {
                 }
 
                 if (e.button === Qt.LeftButton) {
-                    prv.releasedThroughCancel = false
-
                     if (root.clipHeaderHovered) {
                         tracksClipsView.clipStartEditRequested(hoveredClipKey)
                         root.interactionState = TracksClipsView.State.DraggingClip
@@ -449,16 +455,6 @@ Rectangle {
 
             onReleased: e => {
                 if (e.button !== Qt.LeftButton) {
-                    return
-                }
-
-                if (prv.releasedThroughCancel) {
-                    prv.releasedThroughCancel = false
-                    return
-                }
-
-                if (prv.releasedThroughCancel) {
-                    prv.releasedThroughCancel = false
                     return
                 }
 
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -48,11 +48,13 @@ Rectangle {
         id: prv
 
         property bool playRegionActivated: false
+        property bool releasedThroughCancel: false
 
         function cancelClipDragEdit() {
             if (root.hoveredClipKey) {
                 tracksClipsView.cancelClipDragEditRequested(root.hoveredClipKey)
             }
+            releasedThroughCancel = true
         }
     }
 
@@ -402,6 +404,7 @@ Rectangle {
                 }
 
                 if (e.button === Qt.LeftButton) {
+                    prv.releasedThroughCancel = false
                     tracksModel.startUserInteraction()
 
                     if (root.clipHeaderHovered) {
@@ -447,6 +450,16 @@ Rectangle {
 
             onReleased: e => {
                 if (e.button !== Qt.LeftButton) {
+                    return
+                }
+
+                if (prv.releasedThroughCancel) {
+                    prv.releasedThroughCancel = false
+                    return
+                }
+
+                if (prv.releasedThroughCancel) {
+                    prv.releasedThroughCancel = false
                     return
                 }
 
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -405,7 +405,6 @@ Rectangle {
 
                 if (e.button === Qt.LeftButton) {
                     prv.releasedThroughCancel = false
-                    tracksModel.startUserInteraction()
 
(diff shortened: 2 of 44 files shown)


src/appshell/iguifocusstate.h
```diff
@@ -0,0 +1,20 @@
...
+namespace au::appshell {
+class IGuiFocusState : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IGuiFocusState)
+
+public:
+    virtual ~IGuiFocusState() = default;
+
+    virtual bool isFocused() const = 0;
+    virtual muse::async::Channel<bool> isFocusedChanged() const = 0;
```
igorkorsukov: It's unclear from the semantics of this interface what focus is meant.
For example:
* The application as a whole is active
* The main window has focus
* The window in which the element (dialog) is located has focus

We also need to consider what behavior should be in these cases:
* The application has become inactive, and we've switched to another one – it's clear that we need to cancel editing.
* The main window has become inactive, for example, if we're in the main window, we click somewhere and a dialog opens – should we cancel editing?
* The dialog window has become inactive (another case is when we're doing something in the dialog), for example, we've switched to the main window, if the dialog is not modal – should we cancel editing in this case?

src/trackedit/internal/trackeditoperationcontroller.cpp
```diff
@@ -8,6 +8,16 @@ namespace au::trackedit {
 TrackeditOperationController::TrackeditOperationController(std::unique_ptr<IUndoManager> undoManager)
     : m_undoManager{std::move(undoManager)} {}
 
+void TrackeditOperationController::init()
+{
+    dispatcher()->reg(this, "cancel", this, &TrackeditOperationController::onCancel);
```
igorkorsukov: This is very abstract, there could be many such places, so what should we do then? Handle one action in many places? We need to think of a better solution.

src/trackedit/internal/trackeditoperationcontroller.cpp
```diff
@@ -522,6 +532,21 @@ bool TrackeditOperationController::undoRedoToIndex(size_t index)
     return m_undoManager->undoRedoToIndex(index);
 }
 
+void TrackeditOperationController::onCancel()
```
igorkorsukov: I think we need to explore this solution: 

1. At the Qml level.
Qml items has a focus property, we can subscribe to its change, we can also handle key presses, we can check if an Esc is pressed.

2. At the level of the navigation system (I think better)
Navigation items (controls, panels) have an active property that can be used. The navigation system also provides events, allowing us to check whether Esc has been pressed.

```
onNavigationEvent: function(event) {
                    if (event.type === NavigationEvent.Escape) {
                        root.cancel()
                    }
                }
```

src/appshell/iguifocusstate.h
igorkorsukov: I think if we need to add an application activity property, it's better to add it here:  framework/global/iapplication.h

and name it active, not focus, because only windows have focus, and an application can have many windows.

saintmatthieu: I'm not sure what you mean by "active" ?
Here, it's suppose to mean "The GUI application is in focus". I imagined that "GUI" by itself would be sufficient to convey this. Also, the fact that this API is in the appshell module was supposed to help disambiguating.

igorkorsukov: But what does "The GUI application is in focus in focus" mean? Is the main window in focus? Is any application window in focus? It's not for nothing that in Qt it's "active" and not "focus" :)

src/trackedit/internal/trackeditoperationcontroller.cpp
saintmatthieu: I'm not sure I understand your concern. MSS locates this in `notationactionscontroller.cpp`,
```cpp
registerAction("notation-escape", &Controller::resetState, &Controller::isNotationPage);
```
Would it be better if I did the same from `trackeditactionscontroller` ?

src/appshell/iguifocusstate.h
saintmatthieu: I understand what "in focus" means for an application, less what "active" means. I saw that Qt used "active", too, I have to understand why. To me, an application has a main window and child windows. If any of the child windows is in focus, then the application is focus too.

src/projectscene/internal/projectsceneuiactions.cpp
```diff
@@ -16,6 +16,10 @@ using namespace muse::ui;
 using namespace muse::actions;
 
 static UiActionList STATIC_ACTIONS = {
+    UiAction("escape",
+             au::context::UiCtxProjectFocused,
+             muse::shortcuts::CTX_PROJECT_FOCUSED
+             ),
```
saintmatthieu: Obsolete - will remove

src/projectscene/view/timeline/timelinecontext.cpp
```diff
@@ -381,8 +391,6 @@ void TimelineContext::onProjectChanged()
     project->timeSignatureChanged().onReceive(this, [this](const trackedit::TimeSignature&) {
         updateTimeSignature();
     });
-
-    updateTimeSignature();
```
grliszas14, line 389: is it for sure not needed?
I think if you:
1. change the project's time signature (7/4 or smth)
2. save the project
3. close the project
4. open project again

The time signature will be default 4/4 instead of 7/4

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -450,6 +453,16 @@ Rectangle {
...
 
+                if (prv.releasedThroughCancel) {
+                    prv.releasedThroughCancel = false
+                    return
+                }
+
+                if (prv.releasedThroughCancel) {
+                    prv.releasedThroughCancel = false
+                    return
+                }
```
grliszas14: doubled?

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -44,17 +44,26 @@ Rectangle {
...
-        property bool releasedThroughCancel: false
 
         function cancelClipDragEdit() {
+            if (mainMouseArea.pressed) {
+                // This will lead to a cancel signal on `mainMouseArea` that will call back into this function,
+                // but this time in released state.
+                mouseHelper.callUngrabMouseOnItem(mainMouseArea)
+                return
+            }
+            console.debug("handling cancel signal")
```
grliszas14: leftover

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -452,15 +459,7 @@ Rectangle {
...
-                if (prv.releasedThroughCancel) {
-                    prv.releasedThroughCancel = false
-                    return
-                }
-
-                if (prv.releasedThroughCancel) {
-                    prv.releasedThroughCancel = false
-                    return
-                }
+                console.debug("MouseArea released")
```
grliszas14: leftover

src/projectscene/view/timeline/timelinecontext.cpp
saintmatthieu: I don't know how I came to remove this, I can't remember having done this... Good catch, thanks!

grliszas14 (approved):

saintmatthieu:
Probably resolves #8928, but as pointed out by @dozzzzer , can't be tested on MacOS atm due to #9661

PR #9540, opened 2025-10-03
Record anywhere

grliszas14:
test


src/record/internal/au3/au3record.cpp
```diff
@@ -186,45 +211,51 @@ void Au3Record::init()
...
+        std::shared_ptr<Au3WaveClip> pendingClip = DomAccessor::findWaveClip(pendingWaveTrack, size_t(0));
         IF_ASSERT_FAILED(pendingClip) {
             return;
         }
 
-        trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+        // need to flush, otherwise samples hang in the buffer so GetNumSamples returns
+        // same number of samples until we reach the size of sequence (262,144 samples) which
+        // results in refresh rate of once per 5.94s at 44.1 kHz
+        pendingClip->Flush();
```
luapmartin: we should circumvent having to manually Flush and rely on recording callbacks to create blocks and flush when the right amount of samples has been recorded

we have to instead push the same samples from the pending clip to the recording clip and make sure those get aligned and that will provide us with cancelation of size and not weight twice the size during recording.

kryksyh: To clarify what Paul said, calling for a flush too often will lead to a very sparse project file. Each flush call creates a new row in the database, while in the normal recording process, rows are created for blocks of ~200k samples. 

This could potentially lead to a performance downgrade on large projects and lower cloud sync.

kryksyh: Actually it is not as bad as I expected, it is twice as sparse as on master
image

Still, if possible, it is better to avoid flushes.

kryksyh:
Another issue, is that the project file is twice as large as on master when recording with same settings. Running `vacuum` fixes it, so this is probably is not a direct result of this PR, but a side effect of creating and deleting temporary clips

image

src/record/internal/au3/au3record.cpp
```diff
@@ -41,23 +41,55 @@ void pasteDelta(Au3WaveClip& pendingClip,
...
+                     delta->SamplesToTime(lastCopied));
+        delta->Clear(delta->SamplesToTime(committed),
+                     delta->GetSequenceEndTime());
 
-    delta->Clear(delta->GetSequenceStartTime(),
-                 delta->SamplesToTime(lastCopied));
-    delta->Clear(delta->SamplesToTime(total),
-                 delta->GetSequenceEndTime());
+        const double t0 = origClip.GetPlayEndTime();
+        origClip.Paste(t0, *delta);
```
luapmartin: maybe we could `ClearSequence` for the `pendingClip` till the point we pasted (starting from `GetSequenceStartTime()`)

we would avoid the delta Clear of the beginning 

and we might be able to make sure the temp recording project size doesn't twice the size of the project while recording ?

src/record/internal/au3/au3record.h
```diff
@@ -12,9 +12,12 @@
 #include "actions/iactionsdispatcher.h"
 #include "trackedit/iprojecthistory.h"
 #include "playback/iplayback.h"
+#include "trackedit/itrackeditinteraction.h"
 
 #include "au3wrap/au3types.h"
 
+#include "SampleCount.h"
+
```
luapmartin: why samplecount is needed here?

src/projectscene/view/clipsview/au3/connectingdotspainter.cpp
```diff
@@ -29,26 +29,15 @@ void drawConnectingPoints(const au::projectscene::SampleData& samples, const au:
...
+    if (!track) {
         return;
     }
 
-    std::shared_ptr<Au3WaveClip> waveClip = DomAccessor::findWaveClip(track, pendingClipId);
-    IF_ASSERT_FAILED(waveClip) {
+    std::shared_ptr<WaveClip> waveClip = au::au3::DomAccessor::findWaveClip(track, clipKey.clipId);
+    if (!waveClip) {
         return;
     }
```
luapmartin, line 42: maybe the DomAccessor::findWaveClip could directly use the clipKey as we get the track from the clipKey.trackId and then pass it to the method with the clipKey.clipId maybe we could have a useful override or replacement

src/projectscene/view/clipsview/au3/minmaxrmspainter.cpp
```diff
@@ -21,24 +20,13 @@ void MinMaxRMSPainter::paint(QPainter& painter, const trackedit::ClipKey& clipKe
...
-    if (pendingClipId == -1) {
+    WaveTrack* track = au::au3::DomAccessor::findWaveTrack(*au3Project, TrackId(clipKey.trackId));
+    if (!track) {
         return;
     }
 
-    std::shared_ptr<Au3WaveClip> waveClip = DomAccessor::findWaveClip(track, pendingClipId);
-    IF_ASSERT_FAILED(waveClip) {
+    std::shared_ptr<WaveClip> waveClip = au::au3::DomAccessor::findWaveClip(track, clipKey.clipId);
+    if (!waveClip) {
```
luapmartin, line 29: same observation as in the other painter

src/projectscene/view/clipsview/au3/samplespainter.cpp
```diff
@@ -97,26 +97,15 @@ void SamplesPainter::paint(QPainter& painter, const trackedit::ClipKey& clipKey,
...
-    if (pendingClipId == -1) {
+    WaveTrack* track = au::au3::DomAccessor::findWaveTrack(*au3Project, TrackId(clipKey.trackId));
+    if (!track) {
         return;
     }
 
-    std::shared_ptr<Au3WaveClip> waveClip = DomAccessor::findWaveClip(track, pendingClipId);
-    IF_ASSERT_FAILED(waveClip) {
+    std::shared_ptr<WaveClip> waveClip = au::au3::DomAccessor::findWaveClip(track, clipKey.clipId);
+    if (!waveClip) {
```
luapmartin, line 108: here too

src/record/internal/au3/au3record.cpp
```diff
@@ -186,45 +186,71 @@ void Au3Record::init()
...
+
+            // pending track has only single clip
+            std::shared_ptr<Au3WaveClip> pendingClip = DomAccessor::findWaveClip(pendingWaveTrack, size_t(0));
+            IF_ASSERT_FAILED(pendingClip) {
+                return;
+            }
+
+            // audio thread updates the pending clip
+            // make original clip point to the pending clip's data
+            origClip->LinkToOtherSource(*pendingClip.get());
```
luapmartin, line 216: <3

src/projectscene/view/clipsview/au3/connectingdotspainter.cpp
grliszas14: Well, I'd leave that for another PR so we keep there just relevant changes. This thing is used just everywhere, I think we can make another function that accepts just clipKey (and au3Project probably too) and returns WaveClip, but it would be nice to replace it everywhere it can be replaced in one go (I bet there are places in the code we need to have those separated anyway)

src/record/internal/au3/au3record.h
grliszas14: leftover from previous pasteDelta solution :D removed

src/projectscene/view/clipsview/au3/connectingdotspainter.cpp
luapmartin: https://github.com/audacity/audacity/issues/9616

au3/libraries/lib-track/PendingTracks.cpp
```diff
@@ -55,6 +55,23 @@ void PendingTracks::RegisterPendingNewTracks(TrackList& list)
     mTracks.Append(std::move(list), false);
 }
 
+Track* PendingTracks::FindPendingTrack(const Track& track) const
+{
+    auto pTrack = &track;
+    if (!mPendingUpdates->empty()) {
```
kryksyh: I believe find_if would work fine even if the container is empty, maybe something like this would be easier to read

```
Track* PendingTracks::FindPendingTrack(const Track& track) const
{
    auto id = track.GetId();
    auto pred = [id](const auto& pTrack) {
        return pTrack->GetId() == id;
    };

    auto it = std::find_if(mPendingUpdates->begin(), mPendingUpdates->end(), pred);
    return it != mPendingUpdates->end() ? *it : nullptr;
}
```

src/record/internal/au3/au3record.cpp
kryksyh, line 216: That's very nice, this way we could make shallow copies that do not detach on edit

src/record/internal/au3/au3record.cpp
```diff
@@ -186,45 +186,71 @@ void Au3Record::init()
...
+
+            std::shared_ptr<Au3WaveClip> origClip = DomAccessor::findWaveClip(origWaveTrack, clipKey.clipId);
+            IF_ASSERT_FAILED(origClip) {
+                return;
+            }
+
+            trackedit::ITrackeditProjectPtr prj = globalContext()->currentTrackeditProject();
+
+            prj->notifyAboutClipChanged(DomConverter::clip(origWaveTrack, origClip.get()));
+            m_recordPosition.set(origClip->GetPlayEndTime());
```
luapmartin, line 242: there's one recordPosition that's shared by all recordEntry right? 

is it necessary to set it multiple times
`            m_recordPosition.set(origClip->GetPlayEndTime());`

can we be sure all `origClip->GetPlayEndTime()` will be equal for all clips?

PR #9571, opened 2025-10-05
Remove use of private Qt function in navigableappmenumodel.cpp

AJenbo:
This PR removes the use of the private Qt function QKeyMapper::possibleKeys() in navigableappmenumodel.cpp and replaces it with a public-API-compatible implementation.

Using private Qt functions can create significant issues for both developers and the community: it prevents building Audacity from public Qt headers, introduces potential [breakage when updating Qt](https://github.com/musescore/MuseScore/issues/23980), and makes it harder for contributors to build and test changes. By relying solely on public Qt APIs, we maintain compatibility, ensure a smoother developer experience, and reduce the risk of future maintenance problems.

This change preserves the original functionality while keeping the code fully compliant with publicly available Qt headers, supporting both current and future contributors.

Recommended:

QA:

### DIFF
--- a/src/appshell/view/navigableappmenumodel.cpp
+++ b/src/appshell/view/navigableappmenumodel.cpp
@@ -25,44 +25,44 @@
 #include <QWindow>
 #include <QKeyEvent>
 
-#include <private/qkeymapper_p.h>
-
 #include "log.h"
 
 using namespace au::appshell;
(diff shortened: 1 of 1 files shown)


AJenbo:
See https://github.com/musescore/MuseScore/pull/30357

PR #9589, opened 2025-10-07
Added global actions

igorkorsukov:
We have common shortcuts for certain actions, such as Copy - Ctrl+C, Paste - Ctrl+V, Undo - Ctrl+Z.
But what exactly needs to be done depends on the context of where we are.

There are at least two solutions to this problem:

1. The currently active place can accept current shortcuts and process them. The current place can be determined by the navigation system, which element is currently active. We have a single global shortcut controller; it can redirect "global" shortcuts to the active element, so that the desired action can be performed in its context.

2. The application is built on an action system, and shortcuts are just one way to dispatch an action.
If we need to perform different actions depending on the context when pressing the same shortcut, for example, Copy (Ctrl+C), then we can introduce the concept of "global" actions and dispatch them. The application controller can determine where we are and send a more specific action. It can also be further refined to a specific action.

For example:
* Ctrl+C - sent `action://copy` (global action)
* [ApplicationActionController] Is the project open and in focus? - send `action://trackedit/copy`
* [TrackeditActionsController] Is a clip selected? - send `action://trackedit/clip/copy?clipKey=...`
* [TrackeditActionsController] Is a track selected? - send `action://trackedit/track/copy?trackId=...`

This system seems more suited to the overall concept of the application, the action system.
It allows us to send the actions we need.
For example, if it's the application's global menu, then we should send a global action there.
If it's the context menu of a project object (track, clip), then we should send at least a trackedit action (or, preferably, a more specific one, depending on the context).

This PR implements the second concept.


src/appshell/internal/applicationactioncontroller.cpp
```diff
@@ -64,6 +64,16 @@ void ApplicationActionController::init()
     dispatcher()->reg(this, "audio-settings", this, &ApplicationActionController::openAudioSettingsDialog);
     dispatcher()->reg(this, "shortcuts-preferences", this, &ApplicationActionController::openShortcutsPreferencesDialog);
     dispatcher()->reg(this, "editing-preferences", this, &ApplicationActionController::openEditingPreferencesDialog);
+
+    // Global actions
+    dispatcher()->reg(this, "action://copy", this, &ApplicationActionController::doGlobalCopy);
+    dispatcher()->reg(this, "action://cut", this, &ApplicationActionController::doGlobalCut);
+    dispatcher()->reg(this, "action://paste", this, &ApplicationActionController::doGlobalPaste);
+    dispatcher()->reg(this, "action://paste-insert", this, &ApplicationActionController::doGlobalPasteInsert);
```
kryksyh: Paste-insert is a specific trackedit action, and only relevant to trackedit, there is no need for global paste-insert.

In trackedit this action is different from the `paste` action, the first one is inserting data moving everythin to the right, and the second one overwrites overlapping data

igorkorsukov: fixed

kryksyh (approved):

PR #9602, opened 2025-10-08
Add note about installing Qt Shader Tools in setup instructions

cammo1123:
Resolves: ***Not sure if needed. If I need to create an issue, please let me know.***

Found that the build fails on Windows 11 if `Qt Shader Tools` isn't installed. Added a note to the setup guide to clarify this requirement. As it deviates from the linked MuseScore Studio requirements.


kryksyh:
@cammo1123 thanks,
Shader tools are mentioned on line 15 in the doc.

Issue #8720, opened 2025-05-01
Paste behaviour preferences


DilsonsPickles:
### Description

Behaviour and shortcuts for each of the paste methods.

[Figma link](https://www.figma.com/design/Gt4wJqWpinqjk6c1qE125O/19---Preferences?node-id=2976-8837&t=OGSNqJWrH3irgnmg-4)

**Still to-do**

- [x] Redo GIFs in same style as delete behaviour
- [ ] Tidy up spec
- [x] Design review

Tantacrul:
Is this ready for development? It's an alpha task.

luapmartin:
Hi,

I started looking into this, I see that the shortcuts are already implemented and working as expected: 
- Shift+Cmd+V to "Insert and Preserve Sync"
- Cmd+V to "Paste"
- Shift+V to "Paste (Insert)"


<img width="538" height="140" alt="Image" src="https://github.com/user-attachments/assets/ff56db8c-c671-4f13-be50-2c7894f1451d" />

So I understand that we want the **paste behaviour** to be configurable in the settings, under the `Editing` section.

The "Default" should be "Paste" but named "Pasting overlaps other clips" should we change also the description for the shortcut action ?

When choosing a behaviour or another what should happen to the shortcuts?

DilsonsPickles:
**Shortcut names/descriptions**
Let's be consistent with our naming. In the spec I have:
- Paste
- Paste (pushes clips on selected track)
- Ripple paste (preserves synchronization on all tracks)

I think you're right that "Paste" should be renamed. Let's try "Paste (overlaps other clips)" so that it better matches the option in preferences.

Please update the names in shortcuts.


**Remapping shortcuts**
As for the remapping based on the selected paste preference (nice catch btw), If the User changes the default behaviour via the paste preference then the shortcut needs to update to reflect that.

luapmartin:
I pushed a first draft (I have some conflicts)

but there are some questions here for you @DilsonsPickles https://github.com/audacity/audacity/pull/9612#issue-3496295350 thanks in advance

DavidBailes:
> I think you're right that "Paste" should be renamed. Let's try "Paste (overlaps other clips)" so that it better matches the option in preferences.

@DilsonsPickles . It trims other clips, rather than overlaps other clips, doesn't it?

DilsonsPickles:
@DavidBailes It does, this is the language we've been using internally - I'll raise it at the design team meeting tonight.

[...]

Issue #9258, opened 2025-08-26
Remove Dark Theme GIFs for delete behaviour


chinakov:
### Description

We should remove the additional GIFs that get shown on delete behaviour flyout to save space. This should also affect the preferences page for this feature. 

<img width="1028" height="696" alt="Image" src="https://github.com/user-attachments/assets/eca36c6f-2dc3-4448-83d6-e4603165d538" />

luapmartin:
really trivial to be tested with https://github.com/audacity/audacity/pull/9612 PR

dozzzzer:
tested

PR #9612, opened 2025-10-08
8720 Paste behaviour preferences

luapmartin:
Resolves: #8720 Paste behaviour preferences

- [x] Should we move the Pasting Behaviour (the “Always paste  audio as new clip”) section above like in the Figma design 
- [x] Menus etc:
    - [x] Is it ok if app menu show the default paste on top and all paste options below? 
        - [x] Other option is to show the one selected as default (with the dynamic name to what it do) on top and the other (except default) available below dynamicaly too
    - [x] All 4 paste modes are available as shortcut entries in the toolbar
    - [x] Selection context menu offer you the default paste mode (re-ordered after cut/copy)
    - [x] Is there any other menu that would need the paste actions to be listed?
- [x] Please review the currently chosen paste overlap shortcut key combination
- [x] I think we only have light mode GIFs do we need DarkMode too ?

Recommended:

QA:

### DIFF
--- a/src/appshell/view/appmenumodel.cpp
+++ b/src/appshell/view/appmenumodel.cpp
@@ -237,9 +237,9 @@ MenuItem* AppMenuModel::makeEditMenu()
         makeMenuItem("duplicate"),
         // TODO: we should only show those trackedit actions when in the trackedit context,
         // this need some framework update to support contextual shortcuts and dynamic actions based on context
-        makeMenuItem("action://trackedit/paste-overlap"),
-        makeMenuItem("action://trackedit/paste-insert"),
-        makeMenuItem("action://trackedit/paste-insert-all-tracks-ripple"),
+        // makeMenuItem("action://trackedit/paste-overlap"),
+        // makeMenuItem("action://trackedit/paste-insert"),
+        // makeMenuItem("action://trackedit/paste-insert-all-tracks-ripple"),
         makeMenuItem("delete-per-track-ripple"),
         makeSeparator(),
         makeMenu(TranslatableString("appshell/menu/clip", "Clip"), makeClipItems(), "menu-clip"),


DilsonsPickles:
> Should we move the Pasting Behaviour (the “Always paste audio as new clip”) section above like in the Figma design

PasteClipAsNewClip

If you mean putting this preference in here, yes please!

>  Is it ok if app menu show the default paste on top and all paste options below? Other option is to show the one selected as default (with the dynamic name to what it do) on top and the other (except default) available below dynamicaly too

My preference would be to only show the other paste options below so we're not doubling up and adding more options into menus that we're trying to streamline. Let's discuss in design team meeting.

> All 4 paste modes are available as shortcut entries in the toolbar

The three paste modes + the new "default paste" one

> Selection context menu offer you the default paste mode (re-ordered after cut/copy)

This should just be called "Paste"

> Is there any other menu that would need the paste actions to be listed?

Edit menu, Right click menu on empty canvas space, I can't think of anywhere else!

>  I think we only have light mode GIFs do we need DarkMode too ?

We only used the light ones for the Delete preference GIFs so light ones only is fine here.

src/app/configs/data/shortcuts.xml
```diff
@@ -128,10 +128,24 @@
 <!-- end Global actions -->
 
 <!-- Trackedit actions -->
+    <SC>
+        <key>action://trackedit/paste-default</key>
+<!--  TODO we should allow shortcut for context action
+      <seq>Ctrl+V</seq>
+      -->
+    </SC>
```
igorkorsukov: I don't think this shortcut is needed.

src/appshell/view/appmenumodel.cpp
```diff
@@ -228,14 +228,18 @@ MenuItem* AppMenuModel::makeEditMenu()
         makeSeparator(),
         makeMenuItem("action://cut"),
         makeMenuItem("action://copy"),
-        makeMenuItem("action://paste"),
+        makeMenuItem("action://paste"), // TODO: when in Trackedit context this should be dynamically changed to be makeMenuItem("action://trackedit/paste-default"),
```
igorkorsukov: Yes, at the moment all these points behave like `action://trackedit/...`, so they need to be replaced with `action://trackedit/...` (my bad)

src/projectscene/view/clipsview/clipcontextmenumodel.cpp
```diff
@@ -41,9 +41,12 @@ void ClipContextMenuModel::load()
         makeItemWithArg("rename-clip"),
         makeMenu(muse::TranslatableString("clip", "Clip color"), colorItems, "colorMenu"),
         makeSeparator(),
+        // TODO: here the trackedit/ actions won't have shortcut keys associated with and wont show in the popup,
+        // we could default to use the main action like action://cut but that doesn't make sense in terms of design,
+        // the framework should evolve to support contextual shortcut actions
```
igorkorsukov: This has already been implemented, the framework will be updated soon.
https://github.com/musescore/MuseScore/pull/30385

src/trackedit/internal/trackeditactionscontroller.cpp
```diff
@@ -789,7 +792,26 @@ void TrackeditActionsController::rangeSelectionDelete(const ActionData& args)
...
+        case PasteInsertBehavior::PasteInsert:
+            pasteInsert();
+            break;
+        case PasteInsertBehavior::PasteInsertRipple:
+            pasteInsertRipple();
+            break;
+        }
+        break;
+    }
+}
```
igorkorsukov, line 812: I don't mind this implementation, but I think it's better to send specific actions, this will eliminate any duplication and simplify

```
void TrackeditActionsController::pasteDefault()
{
    switch (configuration()->pasteBehavior()) {
    case PasteBehavior::PasteOverlap:
        dispatcher()->dispatch("action://trackedit/paste-overlap");
        break;
...
```

src/appshell/view/appmenumodel.cpp
kryksyh: I'm not sure if I understood that. I believe that "copy", "paste" etc in the main menu should be generic actions, that will be dispatched to the appropriate controller depending on the current context

src/trackedit/internal/trackedituiactions.cpp
```diff
@@ -265,11 +250,34 @@ UiActionList STATIC_ACTIONS = {
              TranslatableString("action", "Enable vertical rulers"),
              TranslatableString("action", "Enable vertical rulers")
              ),
-    UiAction("paste-insert-all-tracks-ripple",
+    // TODO: the past action is handled at app level, when we enable this trackedit/paste-default action,
+    // there will be two paste actions in the shortcut menu, until the framework supports contextual actions/shortcuts
+    UiAction("action://trackedit/paste-default",
+             au::context::UiCtxAny,
+             au::context::CTX_ANY,
```
igorkorsukov: there are incorrect contexts here, there should be
au::context::UiCtxProjectOpened,
au::context::CTX_NOTATION_FOCUSED

But after updating the framework, we will need to change it to au::context::CTX_NOTATION_FOCUSED -> mu::context::CTX_DISABLED, only for `paste-default`, so that it does not appear in the list of actions for shortcuts

src/appshell/view/appmenumodel.cpp
igorkorsukov: Technically, you're right! (That's why I wrote them as general actions in the previous PR.)
But in reality, they currently behave like trackedit actions (for example, paste isn't available if there's nothing to paste to project).
We either need to improve the availability of the global action and the change request notification (it's not difficult...). Or we clearly can't define trackedit actions here now, and then we'll improve them when other actions appear.

src/trackedit/internal/trackedituiactions.cpp
igorkorsukov: Oh, all the actions there have the wrong context.
The action itself can't be sent at any time - it must be in a project opened.
The shortcut can't be sent at any time -  it must be in a project in focus.

src/trackedit/internal/trackeditactionscontroller.cpp
luapmartin: dispatching using the `TRACKEDIT_PASTE_OVERLAP_CODE` or `"action://trackedit/paste-overlap"` codes?

src/trackedit/internal/trackedituiactions.cpp
luapmartin: I updated contexts please tell me if that's correct

src/appshell/view/appmenumodel.cpp
luapmartin: at app level I will maintain the main `action://paste` and for inner menus will use the `action://trackedit/paste`

src/app/configs/data/shortcuts.xml
luapmartin: I removed the shortcut for the `trackedit/paste` as it will rely on it's now defined parent action `paste`

src/appshell/view/appmenumodel.cpp
```diff
@@ -235,7 +235,11 @@ MenuItem* AppMenuModel::makeEditMenu()
         makeMenuItem("ungroup-clips"),
         makeSeparator(),
         makeMenuItem("duplicate"),
+        // TODO: we should only show those trackedit actions when in the trackedit context,
+        // this need some framework update to support contextual shortcuts and dynamic actions based on context
```
igorkorsukov, line 239: I don't think there will be any dynamic actions here that will change depending on the context.

The following scenarios could arise:
1. The actions are specific to a particular area (for example, for trackedit as the main work area), and in other contexts, they would simply be disabled.
2. Another approach is to recreate the menu with different items when changing pages, meaning the menu changes, not the actions (but I doubt we'll do that).
3. And the third scenario is when the item isn't specific to the user, such as Copy, Paste, etc., which should work in context. I think in this case, we'll simply add the global actions action://copy, action://paste..., and some controller will decide which action to send next.

We're not doing this right now because it requires some work. It's better to wait until more actions are added.

igorkorsukov: These three actions are just specific to trackedit and will not be used for anything else (very unlikely)

src/projectscene/view/clipsview/clipcontextmenumodel.cpp
igorkorsukov: This has already been done and merged into the master; a shortcut "parent" of the action will be shown for these actions.

src/projectscene/view/clipsview/selectioncontextmenumodel.cpp
```diff
@@ -22,11 +22,14 @@ void SelectionContextMenuModel::load()
 MenuItemList SelectionContextMenuModel::makeItems()
 {
     MenuItemList items {
+        // TODO: here the trackedit/ actions won't have shortcut keys associated with and wont show in the popup,
+        // we could default to use the main action like action://cut but that doesn't make sense in terms of design,
+        // the framework should evolve to support contextual shortcut actions
```
igorkorsukov: This has already been done and merged into the master; a shortcut "parent" of the action will be shown for these actions.

src/trackedit/internal/trackeditactionscontroller.cpp
igorkorsukov: It doesn't matter to me... since named constants are used here, it's probably better to use it.

src/appshell/view/appmenumodel.cpp
luapmartin: so if I understand correctly we want to keep on adding these 3 actions and these will show as disabled even in menus where it's not contextually relevant (instead of only showing in open project context menu, and hide in home project context)

meaning that these 3 actions will show:
- enabled when something to paste when inside the context of an opened project
- disable if nothing to paste when inside the context of an opened project
- disabled when in the context of the projects home screen

is that correct?

src/trackedit/internal/trackedituiactions.cpp
igorkorsukov: for those that were added or changed - correct

src/appshell/view/appmenumodel.cpp
igorkorsukov: then improvements are needed
The problem is that we need to disable menu items if the action is not available (for example, paste if we didn't copy, or undo if there is nothing to undo).

1. A global action is enabled if at least one action it can redirect is enabled.
2. We need to subscribe to the enable notification for all actions it redirects and update its state.

src/appshell/view/appmenumodel.cpp
igorkorsukov: Yes, that's right, that's our current approach.
This is not a technical question, but a question for the designer or product manager.

We can either create a separate menu for each primary page (but this isn't usually done).
Or we can disable items (actions) that aren't applicable on that page (this is what we do).

src/appshell/view/appmenumodel.cpp
luapmartin: ok, so I understand that it's of to use global actions and that this will be fixed forward, do you have the following task? that we could link here? thanks

igorkorsukov: No, I don't have a task.
And I don't think there's any need to add anything.
I think this will simply be an evolution of the system once the next requirements are established.

It's just that now we can clearly see what actions are in the app menu. It was all the same before, just not as clearly!

luapmartin:
when testing this PR please also test

Remove Dark Theme GIFs for delete behaviour
[#9258](https://github.com/audacity/audacity/issues/9258)

and 

Paste option is missing in the context menu
[#9639](https://github.com/audacity/audacity/issues/9639)

Issue #8917, opened 2025-06-12
Switching vertical playback meter's visibility on/off can be a bit laggy on slower machines


dozzzzer:
https://github.com/user-attachments/assets/be6fd84d-ab05-40fe-9b75-8ba55eaf653c

chinakov:
This is still reproducible on the latest Master (f7034596fe9f5df9d605c52ddc4196ffee3a15b0)

luapmartin:
most probably will be fixed by: https://github.com/audacity/audacity/issues/9647

dozzzzer:
Tested; there's a bit of delay when switching the visibility of the playback meter, but I deem it acceptable.

luapmartin:
Yes, I understand why : we are opening the audio stream so no data is received until it's open, we could fake that with some UI tweaks to avoid this uncanny pause until something is displayed, let's fix that forward if user complain, thanks for testing

Issue #8636, opened 2025-04-10
When resuming playback by pressing Space, cursor (playhead) jumps in time


dozzzzer:
### Steps to reproduce

1. Play a clip and pause by pressing the Space key
2. Repeat multiple times while observing the cursor (playhead) movement

### Expected behavior

Cursor moves smoothly, with no jitter.

### Actual behavior

The cursor moves jitterily, jumping forward in time, sometimes multiple times in a row. 

https://github.com/user-attachments/assets/4a500255-1ae7-4ff7-97c6-f8474de82fda

### Audacity Version

master

### Operating system

OS independent

chinakov:
This is still reproducible on the latest Master: efdaf1e91b151bcc5fe37cab8c0d1539afaf569c

chinakov:
This is still reproducible in the latest Master build (97567e5405b8d6a48a22ca3855f7fce3e937)

chinakov:
This is still reproducible on the latest Master build (35b55c64762ee3bbecfd7103387db78b7ed5f69c)

luapmartin:
Hypothesis: the part of the hardware buffer after the pause position that isn't streamed out is discarded but the playhead is considered to have reach the end of the hardware buffer as if it was streamed out, next play will start from wrong position and be jumping ahead of time

that would explain why this behaviour is inconsistent when using a big hardware buffer and repeating the play/pause multiple time, indeed sometime we would be close to the end of the buffer hence producing only a slight drift and some other time we could be at the beginning of the hardware buffer when pausing and the drift will be almost an entire buffer

we would need to capture the pause time and diff the time spent since the last hardware buffer has been streamed and convert back to samples and offset the difference between the full hardware buffer size and the spent time to "roll back" the playhead at the correct position in time

what do you think @saintmatthieu

saintmatthieu:
@luapmartin I think your hypothesis is correct. I am not sure about your solution - I have new elements:
Like I suspected, I verified that it was indeed Introduced by #8587

I also realize that that PR introduced yet another bug where **audio gets lost when resuming after pause**:
* I have a recording of me counting
* I set the buffer size to 999ms to make it more obvious (so don't mind the delay between my pressing play and you hearing the audio)
* I press play, then pause, then play again.

With the purge introduced by that PR, when I resume after pause, you notice that some of the counts were lost:
https://github.com/user-attachments/assets/1c36825f-1741-461e-86a2-e267aac38b4f

If I remove that purge,
[...]

PR #9613, opened 2025-10-08
9441 smoother playhead

### LINKED ISSUE #8810: Cursor visual glitch when playback is active (Windows)
https://github.com/user-attachments/assets/97d80dbe-e2b7-4bbb-988d-f8de316fcb93

### LINKED ISSUE #9441: Smoother playhead
### Problem to be solved

At the moment, the position of the playhead is updated on the audio hardware thread.
The frequency of these calls depends on the audio buffer setting from the preferences. By default, we request 100ms, which means an update rate of 10fps, which is slow.

### Your idea

An alternative implementation would be:
* relocate this functionality on the audio thread (not the hardware thread)
* create a dedicated class for this, which has access to the playback policy (so that it knows if looping back)
* each time we push a block of samples for the audio thread to consume, update the playhead position to the first sample of that block
* at a desired refresh rate, extrapolate (or "guess") the intermediary playhead position

saintmatthieu:
Resolves: #9441

Recommended:

QA:
- [x] Position of cursor is well synchronized with what is played through the speakers (tip: import a rhythm track from Audacity 3)
- [x] Buffer length in Audio settings can be set to a very large value (e.g. 999ms) and the cursor is still well synchronized.
- [x] Robust when pausing/playing/pausing/... in quick successions
- [x] Also works for stereo meters
- [x] Works for different device types (MME, Wasapi, CoreAudio, etc.)

### DIFF
--- a/au3/libraries/lib-audio-devices/IMeterSender.h
+++ b/au3/libraries/lib-audio-devices/IMeterSender.h
@@ -10,10 +10,25 @@ class AUDIO_DEVICES_API IMeterSender
 public:
     struct InterleavedSampleData
     {
+        InterleavedSampleData(
+            const float* const buf,
+            size_t frm,
+            size_t nCh,
+            double firstTs,
+            std::chrono::steady_clock::time_point when_)
+            : buffer(buf)
+            , frames(frm)
+            , nChannels(nCh)
+            , firstSampleTimestamp(firstTs)
+            , when(when_)
+        {
+        }
+
         const float* const buffer;
         const size_t frames;
         const size_t nChannels;
         const double firstSampleTimestamp;
+        const std::chrono::steady_clock::time_point when;
     };
 
     struct TrackId {
@@ -24,9 +39,10 @@ public:
         bool operator<(const TrackId& other) const { return value < other.value; }
     };
 
+    using OptionalTimePoint = std::optional<std::chrono::steady_clock::time_point>;
+
     virtual ~IMeterSender() = default;
     virtual void push(uint8_t channel, const InterleavedSampleData& sampleData, TrackId = TrackId { MASTER_TRACK_ID }) = 0;
-    virtual void start() = 0;
+    virtual void start(double sampleRate) = 0;
     virtual void stop() = 0;
-    virtual void setSampleRate(double rate) = 0;
 };
--- a/au3/libraries/lib-audio-devices/IMeterSender.h
+++ b/au3/libraries/lib-audio-devices/IMeterSender.h
@@ -4,31 +4,30 @@
 #pragma once
 
 constexpr int64_t MASTER_TRACK_ID = -2;
+using TimePoint = std::chrono::steady_clock::time_point;
 
 class AUDIO_DEVICES_API IMeterSender
 {
 public:
+
     struct InterleavedSampleData
     {
         InterleavedSampleData(
             const float* const buf,
             size_t frm,
             size_t nCh,
-            double firstTs,
-            std::chrono::steady_clock::time_point when_)
+            TimePoint dacTime)
             : buffer(buf)
             , frames(frm)
             , nChannels(nCh)
-            , firstSampleTimestamp(firstTs)
-            , when(when_)
+            , dacTime(std::move(dacTime))
         {
         }
 
         const float* const buffer;
         const size_t frames;
         const size_t nChannels;
-        const double firstSampleTimestamp;
-        const std::chrono::steady_clock::time_point when;
+        const TimePoint dacTime;
     };
 
     struct TrackId {
@@ -39,10 +38,12 @@ public:
         bool operator<(const TrackId& other) const { return value < other.value; }
     };
 
-    using OptionalTimePoint = std::optional<std::chrono::steady_clock::time_point>;
+    using OptionalTimePoint = std::optional<TimePoint>;
 
     virtual ~IMeterSender() = default;
     virtual void push(uint8_t channel, const InterleavedSampleData& sampleData, TrackId = TrackId { MASTER_TRACK_ID }) = 0;
     virtual void start(double sampleRate) = 0;
     virtual void stop() = 0;
 };
+
+using IMeterSenderPtr = std::shared_ptr<IMeterSender>;
--- a/au3/libraries/lib-audio-io/AudioIO.cpp
+++ b/au3/libraries/lib-audio-io/AudioIO.cpp
@@ -2748,10 +2748,15 @@ bool AudioIoCallback::FillOutputBuffers(
         // the device. For example mono channels output to both left and right
         // output channels.
         if (len > 0) {
-            if (!mUserStartsHearingAudioTimePoint.load().has_value()) {
-                const auto remainingSamples = mHardwarePlaybackLatencyFrames > len ? mHardwarePlaybackLatencyFrames - len : 0;
-                mUserStartsHearingAudioTimePoint.store(std::chrono::steady_clock::now()
-                                                       + std::chrono::milliseconds(static_cast<int>(remainingSamples / mRate * 1000)));
+            if (n == 0) {
+                using namespace std::chrono;
+                const auto now = steady_clock::now();
+                const auto startTime = now + milliseconds(static_cast<int>(mHardwarePlaybackLatencyFrames * 1000.0 / mRate));
+                mAudioDeliveryQueue.Put({ startTime, static_cast<int>(len) });
+                if (!mUserStartsHearingAudioTimePoint.load().has_value()) {
+                    const auto remainingSamples = mHardwarePlaybackLatencyFrames > len ? mHardwarePlaybackLatencyFrames - len : 0;
+                    mUserStartsHearingAudioTimePoint.store(now + milliseconds(static_cast<int>(remainingSamples / mRate * 1000)));
+                }
             }
 
             // Output volume emulation: possibly copy meter samples, then
--- a/au3/libraries/lib-audio-io/AudioIO.cpp
+++ b/au3/libraries/lib-audio-io/AudioIO.cpp
@@ -2661,14 +2661,13 @@ bool AudioIoCallback::FillOutputBuffers(
     }
 
     // Choose a common size to take from all ring buffers
-    const auto toGet
-        =std::min<size_t>(framesPerBuffer, GetCommonlyReadyPlayback());
+    const auto currentlyAvailableFramesAcrossBuffers = std::min<size_t>(framesPerBuffer, GetCommonlyReadyPlayback());
 
     // Poke: If there are no playback sequences, then check playback
     // completion condition and do early return
     // PRL:  Also consume frbom the single playback ring buffer
     if (numPlaybackSequences == 0) {
-        mMaxFramesOutput = mPlaybackBuffers[0]->Discard(toGet);
+        mMaxFramesOutput = mPlaybackBuffers[0]->Discard(currentlyAvailableFramesAcrossBuffers);
         CallbackCheckCompletion(mCallbackReturn, 0);
         mLastPlaybackTimeMillis = ::wxGetUTCTimeMillis();
         return false;
@@ -2702,29 +2701,29 @@ bool AudioIoCallback::FillOutputBuffers(
     }
 
     for (unsigned n = 0; n < numPlaybackChannels; ++n) {
-        decltype(framesPerBuffer) len = mPlaybackBuffers[n]->Get(
+        decltype(framesPerBuffer) numberOfRetrievedFrames = mPlaybackBuffers[n]->Get(
             reinterpret_cast<samplePtr>(tempBufs[n]),
             floatSample,
-            toGet
+            currentlyAvailableFramesAcrossBuffers
             );
 
-        if (len < framesPerBuffer) {
+        if (numberOfRetrievedFrames < framesPerBuffer) {
             // This used to happen normally at the end of non-looping
             // plays, but it can also be an anomalous case where the
             // supply from SequenceBufferExchange fails to keep up with the
             // real-time demand in this thread (see bug 1932). We
             // must supply something to the sound card, so pad it with
             // zeroes and not random garbage.
-            memset((void*)&tempBufs[n][len], 0,
-                   (framesPerBuffer - len) * sizeof(float));
+            memset((void*)&tempBufs[n][numberOfRetrievedFrames], 0,
+                   (framesPerBuffer - numberOfRetrievedFrames) * sizeof(float));
         }
 
         // PRL:  More recent rewrites of SequenceBufferExchange should guarantee a
         // padding out of the ring buffers so that equal lengths are
         // available, so maxLen ought to increase from 0 only once
-        mMaxFramesOutput = std::max(mMaxFramesOutput, len);
+        mMaxFramesOutput = std::max(mMaxFramesOutput, numberOfRetrievedFrames);
 
-        len = mMaxFramesOutput;
+        numberOfRetrievedFrames = mMaxFramesOutput;
 
         // Realtime effect transformation of the sound used to happen here
         // but it is now done already on the producer side of the RingBuffer
@@ -2738,18 +2737,18 @@ bool AudioIoCallback::FillOutputBuffers(
         // Each channel in the sequences can output to more than one channel on
         // the device. For example mono channels output to both left and right
         // output channels.
-        if (len > 0) {
+        if (numberOfRetrievedFrames > 0) {
             if (n == 0) {
                 using namespace std::chrono;
                 const auto now = steady_clock::now();
-                const auto startTime = now + milliseconds(static_cast<int>(mHardwarePlaybackLatencyFrames * 1000.0 / mRate));
-                mAudioDeliveryQueue.Put({ startTime, static_cast<int>(len) });
+                const auto adcTime = now + milliseconds(static_cast<int>(mHardwarePlaybackLatencyFrames * 1000.0 / mRate));
+                mAudioDeliveryQueue.Put({ adcTime, static_cast<int>(numberOfRetrievedFrames) });
             }
 
             // Output volume emulation: possibly copy meter samples, then
             // apply volume, then copy to the output buffer
             if (outputMeterFloats != outputFloats) {
-                for ( unsigned i = 0; i < len; ++i) {
+                for ( unsigned i = 0; i < numberOfRetrievedFrames; ++i) {
                     outputMeterFloats[numPlaybackChannels * i + n]
                         +=playbackVolume * tempBufs[n][i];
                 }
@@ -2765,13 +2764,13 @@ bool AudioIoCallback::FillOutputBuffers(
             // PRL todo:  choose denominator differently, so it doesn't depend on
             // framesPerBuffer, which is influenced by the portAudio implementation in
             // opaque ways
-            const float deltaVolume = (playbackVolume - oldVolume) / len;
-            for (unsigned i = 0; i < len; i++) {
+            const float deltaVolume = (playbackVolume - oldVolume) / numberOfRetrievedFrames;
+            for (unsigned i = 0; i < numberOfRetrievedFrames; i++) {
                 outputFloats[numPlaybackChannels * i + n]
                     +=(oldVolume + deltaVolume * i) * tempBufs[n][i];
             }
         }
-        CallbackCheckCompletion(mCallbackReturn, len);
+        CallbackCheckCompletion(mCallbackReturn, numberOfRetrievedFrames);
     }
 
     mOldPlaybackVolume = playbackVolume;
--- a/au3/libraries/lib-audio-io/AudioIO.cpp
+++ b/au3/libraries/lib-audio-io/AudioIO.cpp
@@ -2742,7 +2742,7 @@ bool AudioIoCallback::FillOutputBuffers(
                 using namespace std::chrono;
                 const auto now = steady_clock::now();
                 const auto adcTime = now + milliseconds(static_cast<int>(mHardwarePlaybackLatencyFrames * 1000.0 / mRate));
-                mAudioDeliveryQueue.Put({ adcTime, static_cast<int>(numberOfRetrievedFrames) });
+                mAudioCallbackInfoQueue.Put({ adcTime, static_cast<int>(numberOfRetrievedFrames) });
             }
 
             // Output volume emulation: possibly copy meter samples, then
--- a/au3/libraries/lib-audio-io/AudioIO.cpp
+++ b/au3/libraries/lib-audio-io/AudioIO.cpp
@@ -661,8 +661,6 @@ bool AudioIO::StartPortAudioStream(const AudioIOStartStreamOptions& options,
         SetCaptureMeter(mOwningProject.lock(), options.captureMeter);
     }
 
-    UpdateMetersRate(mRate);
-
     const auto deviceInfo = usePlayback
                             ? Pa_GetDeviceInfo(playbackParameters.device)
                             : Pa_GetDeviceInfo(captureParameters.device);
@@ -865,11 +863,6 @@ void AudioIO::StartMonitoring(const AudioIOStartStreamOptions& options)
     }
 }
 
-std::optional<std::chrono::steady_clock::time_point> AudioIoCallback::UserStartsHearingAudioTimePoint() const
-{
-    return mUserStartsHearingAudioTimePoint.load();
-}
-
 void AudioIO::StopMonitoring()
 {
     if (IsMonitoring()) {
@@ -890,8 +883,6 @@ int AudioIO::StartStream(const TransportSequences& sequences,
                          double t0, double t1, double mixerLimit,
                          const AudioIOStartStreamOptions& options)
 {
-    mUserStartsHearingAudioTimePoint.store(std::nullopt);
-
     // precondition
     assert(std::all_of(
                sequences.playbackSequences.begin(), sequences.playbackSequences.end(),
@@ -2753,10 +2744,6 @@ bool AudioIoCallback::FillOutputBuffers(
                 const auto now = steady_clock::now();
                 const auto startTime = now + milliseconds(static_cast<int>(mHardwarePlaybackLatencyFrames * 1000.0 / mRate));
                 mAudioDeliveryQueue.Put({ startTime, static_cast<int>(len) });
-                if (!mUserStartsHearingAudioTimePoint.load().has_value()) {
-                    const auto remainingSamples = mHardwarePlaybackLatencyFrames > len ? mHardwarePlaybackLatencyFrames - len : 0;
-                    mUserStartsHearingAudioTimePoint.store(now + milliseconds(static_cast<int>(remainingSamples / mRate * 1000)));
-                }
             }
 
             // Output volume emulation: possibly copy meter samples, then
@@ -3016,7 +3003,7 @@ void AudioIoCallback::DoPlaythrough(
 /* Send data to recording VU meter if applicable */
 // Also computes rms
 void AudioIoCallback::SendVuInputMeterData(const float* inputSamples, unsigned long framesPerBuffer,
-                                           const PaStreamCallbackTimeInfo* timeInfo)
+                                           const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when)
 {
     if (framesPerBuffer == 0) {
         return;
@@ -3031,12 +3018,12 @@ void AudioIoCallback::SendVuInputMeterData(const float* inputSamples, unsigned l
         return;
     }
 
-    PushInputMeterValues(inputMeter, inputSamples, framesPerBuffer, timeInfo);
+    PushInputMeterValues(inputMeter, inputSamples, framesPerBuffer, timeInfo, when);
 }
 
 /* Send data to playback VU meter if applicable */
 void AudioIoCallback::SendVuOutputMeterData(const float* outputMeterFloats, unsigned long framesPerBuffer,
-                                            const PaStreamCallbackTimeInfo* timeInfo)
+                                            const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when)
 {
     if (framesPerBuffer == 0) {
         return;
@@ -3051,12 +3038,12 @@ void AudioIoCallback::SendVuOutputMeterData(const float* outputMeterFloats, unsi
         return;
     }
 
-    PushMainMeterValues(outputMeter, outputMeterFloats, mNumPlaybackChannels, framesPerBuffer, timeInfo);
-    PushTrackMeterValues(outputMeter, framesPerBuffer, timeInfo);
+    PushMainMeterValues(outputMeter, outputMeterFloats, mNumPlaybackChannels, framesPerBuffer, timeInfo, when);
+    PushTrackMeterValues(outputMeter, framesPerBuffer, timeInfo, when);
 }
 
 void AudioIoCallback::PushInputMeterValues(const std::shared_ptr<IMeterSender>& sender, const float* values, unsigned long frames,
-                                           const PaStreamCallbackTimeInfo* timeInfo)
+                                           const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when)
 {
     if (frames == 0) {
         return;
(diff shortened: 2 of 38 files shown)


au3/libraries/lib-audio-io/AudioIO.h
```diff
@@ -400,6 +402,7 @@ class AUDIO_IO_API AudioIoCallback /* not final */ : public AudioIOBase
     void PurgeAfterPause(unsigned long framesPerBuffer, float** sampleBuffer);
 
     std::atomic<bool> mPurgeIsNeeded{ false };
+    bool mProjectSamplesReachedDeviceThread { false };
```
luapmartin: don't we need atomic here?

au3/libraries/lib-audio-io/PlaybackSchedule.h
```diff
@@ -24,7 +24,9 @@ struct AudioIOStartStreamOptions;
 class BoundedEnvelope;
 using PRCrossfadeData = std::vector< std::vector < float > >;
 
-constexpr size_t TimeQueueGrainSize = 2000;
+// This can become a bottleneck to the update rate of the play cursor position.
+// Let's set this to 10ms at 48kHz. Even at 16kHz, that'd be a refresh rate of 33fps.
+constexpr size_t TimeQueueGrainSize = 480;
```
luapmartin, line 29: can't we set that dynamically based on sample rate? to aim for fixed time (same feel) whatever sample rate we are using ? 16ms would provide ~ 60fps and 10ms be higher than 90fps

au3/libraries/lib-audio-io/AudioIO.h
saintmatthieu: I'd tend to say it's not necessary, since there's only one writer and one reader and I can't see how the compiler could introduce any reordering when optimizing, but I'm not 100% sure. You're right, let's stay on the safe side.

au3/libraries/lib-audio-io/PlaybackSchedule.h
saintmatthieu: I've looked a little bit into this. Changing this guy to a variable would require some work that I think overweights the benefits. I'd rather move on quicker to something else.

luapmartin:
@saintmatthieu could you also check if this PR will influence this behaviour: https://github.com/audacity/audacity/issues/8636
it could be related with : correct position update of the playhead on stop/pause maybe we are missing an update when the steams is interrupted.

luapmartin:
I did a first pass and I still see the problems we discussed the other day about the naming.

Latency should be bufferDurationInMs or something that explain what it measure instead of what it feels like or produce to the end user in a specific use case.

also I saw lot of `when` that should be more descriptive of what they represent

and the globals included from framework is something we changed ?

luapmartin:
I will do another pass tomorrow, could you please rebase ? thanks

au3/libraries/lib-audio-devices/IMeterSender.h
```diff
@@ -10,10 +10,25 @@ class AUDIO_DEVICES_API IMeterSender
...
+            const float* const buf,
+            size_t frm,
+            size_t nCh,
+            double firstTs,
+            std::chrono::steady_clock::time_point when_)
+            : buffer(buf)
+            , frames(frm)
+            , nChannels(nCh)
+            , firstSampleTimestamp(firstTs)
+            , when(when_)
```
luapmartin: this `when` need a bit more love, a better name is due I think, let's try asking that by using the 4 other `Ws`

what is it, or what is the reason to have it here? 
what does it stores/represent?
why do we need it?
who responsible or using/needs it?

src/playback/view/common/playbackpositiontimer.h
```diff
@@ -0,0 +1,29 @@
...
+#include "framework/global/async/asyncable.h"
+#include "framework/global/modularity/ioc.h"
+
+#include <QQuickItem>
+#include <QTimer>
+
+namespace au::playback {
+class PlaybackPositionTimer : public QQuickItem, public muse::async::Asyncable
+{
+    muse::Inject<IPlayback> playback;
```
luapmartin, line 14: could you please inject and use the `PlaybackState` instead

src/playback/internal/au3/au3player.h
```diff
@@ -5,18 +5,20 @@
 
 #include "../../iplayer.h"
 
-#include "global/async/asyncable.h"
-#include "global/types/retval.h"
-#include "global/timer.h"
+#include "framework/global/async/asyncable.h"
+#include "framework/global/types/retval.h"
+#include "framework/global/timer.h"
+#include "framework/global/modularity/ioc.h"
```
luapmartin, line 11: why is there `framework/` here?

src/playback/internal/au3/au3player.cpp
```diff
@@ -4,6 +4,7 @@
 #include "au3player.h"
 
 #include "global/types/number.h"
+#include "global/defer.h"
```
luapmartin: should we have `./framework/` here?

au3/libraries/lib-audio-devices/IMeterSender.h
```diff
@@ -24,9 +39,10 @@ class AUDIO_DEVICES_API IMeterSender
         bool operator<(const TrackId& other) const { return value < other.value; }
     };
 
+    using OptionalTimePoint = std::optional<std::chrono::steady_clock::time_point>;
+
     virtual ~IMeterSender() = default;
     virtual void push(uint8_t channel, const InterleavedSampleData& sampleData, TrackId = TrackId { MASTER_TRACK_ID }) = 0;
-    virtual void start() = 0;
+    virtual void start(double sampleRate) = 0;
```
luapmartin, line 45: from API standpoint I don't think passing a sampleRate to `start` method make much sense, why not a `configure` or `setup` or like before `setSampleRate` then a `start()/stop()` method, unless I did not understood, please explain, thanks


_Additional question does those timers can update to renegotiated sample rates during their usage?_

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2744,6 +2735,13 @@ bool AudioIoCallback::FillOutputBuffers(
         // the device. For example mono channels output to both left and right
         // output channels.
         if (len > 0) {
+            if (n == 0) {
+                using namespace std::chrono;
+                const auto now = steady_clock::now();
+                const auto startTime = now + milliseconds(static_cast<int>(mHardwarePlaybackLatencyFrames * 1000.0 / mRate));
+                mAudioDeliveryQueue.Put({ startTime, static_cast<int>(len) });
+            }
+
```
luapmartin, line 2748: `startTime` might not be the best describing name here, why not `estimatedFirstBufferCallbackTime` of something like that ?

if I understand correctly each time you fill the output buffer you try estimate when the first callback will be sending the current buffer out, producing audio output to the user right?

I think there is an issue as the `toGet` might be smaller than the `framesPerBuffer` if at leat one of the buffer in the `mPlaybackBuffers` doesn't have enough frame ready (see `GetCommonlyReadyPlayback()`) so we might have more than one callback to fill the output buffer 

so maybe we should put into the `mAudioDeliveryQueue` only the number of frames that we were able to get from the current mPlaybackBuffers

luapmartin: actually I think that while at it a bit of cleanup would be in order for those two variables `toGet` and `len` aren't best names why not `currentlyAvailableFramesAcrossBuffers` and `numberOfRetrievedFrames` respectively or something that would explicitly let the reader understand what those two are about and how to interpret the logic involved with them

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -2795,7 +2793,7 @@ void AudioIoCallback::UpdateTimePosition(unsigned long framesPerBuffer)
 
     // Update the position seen by drawing code
     mPlaybackSchedule.SetSequenceTime(
-        mPlaybackSchedule.mTimeQueue.Consumer(mMaxFramesOutput, mRate));
+        mPlaybackSchedule.mTimeQueue.Consumer(framesPerBuffer, mRate));
 }
```
luapmartin, line 2801: that make more sense indeed

only problem I see is that this is triggered by the player from a Qt trigger.

I do believe that the playback position updates should be fully isolated on the playback side (AudioIO...) and each time a new buffer is "sent" outside we should accordingly notify and we would then update the visuals accordingly knowing **when** the buffer has been sent out and what was it **size** and the **sample rate** is sufficient to animate a playhead 

I am a bit confused as how this as become apparently so complex, sorry for not fully following the logic here but we might be overcomplicating things

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3001,7 +2999,7 @@ void AudioIoCallback::DoPlaythrough(
 /* Send data to recording VU meter if applicable */
 // Also computes rms
 void AudioIoCallback::SendVuInputMeterData(const float* inputSamples, unsigned long framesPerBuffer,
-                                           const PaStreamCallbackTimeInfo* timeInfo)
+                                           const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when)
 {
     if (framesPerBuffer == 0) {
```
luapmartin: please adjust the `when`, thanks

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3016,12 +3014,12 @@ void AudioIoCallback::SendVuInputMeterData(const float* inputSamples, unsigned l
...
-    PushInputMeterValues(inputMeter, inputSamples, framesPerBuffer, timeInfo);
+    PushInputMeterValues(inputMeter, inputSamples, framesPerBuffer, timeInfo, when);
 }
 
 /* Send data to playback VU meter if applicable */
 void AudioIoCallback::SendVuOutputMeterData(const float* outputMeterFloats, unsigned long framesPerBuffer,
-                                            const PaStreamCallbackTimeInfo* timeInfo)
+                                            const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when)
 {
     if (framesPerBuffer == 0) {
```
luapmartin: please adjust the `when` thanks

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3036,12 +3034,12 @@ void AudioIoCallback::SendVuOutputMeterData(const float* outputMeterFloats, unsi
...
-    PushTrackMeterValues(outputMeter, framesPerBuffer, timeInfo);
+    PushMainMeterValues(outputMeter, outputMeterFloats, mNumPlaybackChannels, framesPerBuffer, timeInfo, when);
+    PushTrackMeterValues(outputMeter, framesPerBuffer, timeInfo, when);
 }
 
 void AudioIoCallback::PushInputMeterValues(const std::shared_ptr<IMeterSender>& sender, const float* values, unsigned long frames,
-                                           const PaStreamCallbackTimeInfo* timeInfo)
+                                           const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when)
 {
     if (frames == 0) {
```
luapmartin, line 3046: please adjust the `when`, thanks

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3078,23 +3076,24 @@ void AudioIoCallback::PushInputMeterValues(const std::shared_ptr<IMeterSender>&
...
+            sender->push(ch, { mainTrackInput + ch * frames, frames, 1, timeInfo->inputBufferAdcTime, when });
         }
     }
 }
 
 void AudioIoCallback::PushMainMeterValues(const std::shared_ptr<IMeterSender>& sender, const float* values, uint8_t channels,
-                                          unsigned long frames, const PaStreamCallbackTimeInfo* timeInfo)
+                                          unsigned long frames, const PaStreamCallbackTimeInfo* timeInfo,
+                                          const std::chrono::steady_clock::time_point& when)
 {
```
luapmartin: please adjust the `when`, thanks

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -3209,6 +3208,9 @@ int AudioIoCallback::AudioCallback(
                                    : outputBuffer;
     // ----- END of MEMORY ALLOCATIONS ------------------------------------------
 
+    const std::chrono::milliseconds hardwareLatencyMs{ static_cast<int>(mHardwarePlaybackLatencyFrames * 1000.0 / mRate) };
+    const auto levelDisplayTime = std::chrono::steady_clock::now() + hardwareLatencyMs;
+
```
luapmartin: `mHardwarePlaybackLatencyFrames` is obtained from latency in seconds we should be packing the `PaStreamInfo` (in a custom struct with the sent to output timestamp) as it contain the used sample rate and PA outputLatency (and input)
then we could directly use that PA info instead of calculating and re-calculating something that PA tell us directly

luapmartin:
```cpp
            mHardwarePlaybackLatencyFrames = lrint(outputLatency * mRate);
```

we might better use the `stream->sampleRate` when doing that to have accurate size

also the `mHardwarePlaybackLatencyFrames` is really only needed in the AllocateBuffers method and there we could be using the PA info struct

au3/libraries/lib-audio-io/AudioIO.h
```diff
@@ -186,6 +187,14 @@ class AUDIO_IO_API AudioIoCallback /* not final */ : public AudioIOBase
     { return mListener.lock(); }
     void SetListener(const std::shared_ptr< AudioIOListener >& listener);
 
+    struct AudioDelivery {
+        std::chrono::steady_clock::time_point startTime;
+        int numSamples = 0;
+    };
+    using AudioDeliveryQueue = LockFreeQueue<AudioDelivery>;
+
```
luapmartin: please add the sample rate 

also I think that this is a struct that is to be used "on a callback" right?

we might want to callback keyword somewhere to translate the fact that this is a per callback payload gathering all together

AudioDelivery is not so evocative could try AudioCallbackPayload or AudioCallbackPayloadInfo or something like that

src/au3wrap/internal/au3audiometer.cpp
```diff
@@ -112,15 +121,14 @@ void Meter::push(uint8_t channel, const IMeterSender::InterleavedSampleData& sam
...
-void Meter::start()
+void Meter::start(double sampleRate)
 {
-    if (!m_meterUpdateTimer.isActive()) {
-        m_meterUpdateTimer.start();
-    }
-    m_stopTimer.stop();
+    m_meterUpdateTimer->start();
+    m_stopTimer->stop();
+    m_sampleRate = sampleRate;
```
luapmartin: maybe we should stop the stop timer first ?

src/au3wrap/internal/au3audiometer.h
```diff
@@ -70,10 +78,11 @@ class Meter : public IMeterSender, public muse::async::Asyncable
...
     std::atomic<int> m_hangoverCount = 0;
     int m_maxFramesPerPush = 0;
-    LockFreeQueue<QueueItem> m_queue{ 1024 };
+    LockFreeQueue<QueueItem> m_lockFreeQueue{ 1024 };
+    std::queue<QueueItem> m_mainThreadQueue;
     std::map<TrackId, TrackData> m_trackData;
-    QTimer m_meterUpdateTimer;
-    QTimer m_stopTimer;
+    const std::unique_ptr<ITimer> m_meterUpdateTimer;
+    const std::unique_ptr<ITimer> m_stopTimer;
```
luapmartin: stop timer could be renamed decay or let ring timer ? as we are using this to decay the signal activity

or rename them as running / stopping  (maybe with signal or timer or acitvity in the name) or something that would let reader know those are to be used in tandem as one when updating the activity and the other one only when being stopping the activity 

_actually we might need a `starting` Activity timer too to fake signal ramping up to avoid user delay when turning on audio engine see https://github.com/audacity/audacity/issues/8917#issuecomment-3431656645_

we could think as those timers as easeIn easeOut smoothing signal flow

src/auaudio/internal/itimer.h
```diff
@@ -0,0 +1,21 @@
...
+public:
+    virtual ~ITimer() = default;
+
+    virtual void setCallback(std::function<void()>) = 0;
+    virtual void setInterval(const std::chrono::milliseconds&) = 0;
+    virtual void setSingleShot(bool) = 0;
+    virtual void start() = 0;
+    virtual void stop() = 0;
+};
+}
```
luapmartin, line 21: new module `auaudio` maybe it should be `auutility` right or something where a timer tool would make sense, wdyt?

src/context/internal/playbackstate.h
```diff
@@ -3,7 +3,8 @@
 */
 #pragma once
 
-#include "global/async/asyncable.h"
+#include "framework/global/async/asyncable.h"
+#include "framework/global/modularity/ioc.h"
```
luapmartin, line 7: the more I see that the more I like the explicit `framework/` route, would be nice to have `framework/` `libraries/` for all inclues and disable the intermediate includes path for includes from framework or libraries as it's not helping IMO then that would be super clear

src/playback/internal/au3/au3player.cpp
```diff
@@ -508,6 +492,42 @@ muse::secs_t Au3Player::playbackPosition() const
...
+    if (static_cast<long long>(m_consumedSamplesSoFar) >= expectedConsumedNow) {
+        // User isn't hearing audio yet - wait some more.
+        return;
+    }
+
+    audioIO.UpdateTimePosition(expectedConsumedNow - m_consumedSamplesSoFar);
+    m_consumedSamplesSoFar = expectedConsumedNow;
+    updatePlaybackStateTimeCritical();
+}
+
```
luapmartin, line 514: not fan of the player updating the playback position I think it should be responsibility of the audio engine (or AudioIO for now)

src/playback/internal/au3/au3player.h
```diff
@@ -73,15 +76,24 @@ class Au3Player : public IPlayer, public muse::async::Asyncable
...
+
+    struct TargetPoint {
+        TargetPoint(const std::chrono::steady_clock::time_point time, const unsigned long long consumedSamples)
+            : time{time}, consumedSamples{consumedSamples} {}
+        const std::chrono::steady_clock::time_point time;
+        const unsigned long long consumedSamples;
+    };
+
+    std::optional<TargetPoint> m_currentTarget;
+    unsigned long long m_consumedSamplesSoFar = 0;
```
luapmartin, line 97: not sure we need that, guess is to interpolate but I think we are going the wrong direction

luapmartin (changes requested):
added comments, thanks

src/auaudio/internal/itimer.h
luapmartin: `autools`
`auutils`

src/playback/view/common/playbackpositiontimer.h
luapmartin: I mean getting it from the GlobalContext

au3/libraries/lib-audio-devices/IMeterSender.h
saintmatthieu: Agreed.
PortAudio has the interesting equivalent of `outputBufferDacTime`. I used simply `dacTime`, with the context it's not hard to grasp that it's the time when the DAC starts processing the first `buf` sample.

au3/libraries/lib-audio-devices/IMeterSender.h
saintmatthieu: I do find some sense: "We start a new playback session, I need to know what sample rate it uses." At least then the implementation doesn't have to verify that a `setup` or `configure` method was called between the last stop and the new start.

To the additional question: sample rate doesn't change during a playout session.

au3/libraries/lib-audio-io/AudioIO.h
```diff
@@ -220,17 +220,12 @@ class AUDIO_IO_API AudioIoCallback /* not final */ : public AudioIOBase
...
-        const std::chrono::steady_clock::time_point& when);
-    void PushMainMeterValues(const std::shared_ptr<IMeterSender>& sender, const float* values, uint8_t channels, unsigned long frames,
-                             const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when);
-    void PushTrackMeterValues(const std::shared_ptr<IMeterSender>& sender, unsigned long frames, const PaStreamCallbackTimeInfo* timeInfo,
-                              const std::chrono::steady_clock::time_point& when);
-    void PushInputMeterValues(const std::shared_ptr<IMeterSender>& sender, const float* values, unsigned long frames,
-                              const PaStreamCallbackTimeInfo* timeInfo, const std::chrono::steady_clock::time_point& when);
+    void SendVuInputMeterData(const float* inputSamples, unsigned long framesPerBuffer, const TimePoint& dacTime);
+    void SendVuOutputMeterData(const float* outputMeterFloats, unsigned long framesPerBuffer, const TimePoint& dacTime);
+    void PushMainMeterValues(const IMeterSenderPtr& sender, const float* values, uint8_t channels, unsigned long frames,
```
luapmartin: `PushMainMeterValues` maybe we want to know it's an output Meter ?

`PushMasterOutputMeterValues` or Main but I think Output would help

for TrackMeter it's not the case as those serve 2 purposes (that might change?) serve both purposes input metering or output metering do you think we would need two methods to indicate clearly which is used ? or this single method is fine?

au3/libraries/lib-audio-devices/IMeterSender.h
luapmartin: then you have to always send the sampleRate via the start method when you start/stop/start/stop... even if sampleRate hasn't changed ?

sampleRate should be sent only when first setup and when its changing so we could start/stop/start/stop ...

au3/libraries/lib-audio-devices/IMeterSender.h
luapmartin: here we store the time at which we are creating that buffer + the duration of the buffer 

we are neglecting the extra time that is spent to move things around and have an estimate of the time it will be played.

We still have the problem if the framesPerBuffer isn't met (in the `FillOutputBuffer` see `toGet` variable) and we have double callback the timings will be off

we will be sending twice to the input vu meters

au3/libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: > if I understand correctly each time you fill the output buffer you try estimate when the first callback will be sending the current buffer out, producing audio output to the user right?

Not sure what you mean with "first callback", but beside this, yes, this is my intention: when the first of these `len` samples will reach DAC. In that sense, I mean "start time" as "start time of that ADC conversion".
Yesterday I replaced `when` with `dacTime` in the `AudioDelivery` struct and I overlooked this place.

> I think there is an issue as the toGet might be smaller than the framesPerBuffer

I don't think this is a normal case: a few lines above, there is the `if (len < framesPerBuffer) {` condition, with the comment saying that it can only happen
* at the end of the playback (probably then nothing any user would notice?)
* if the audio thread is slower than the hardware thread and there are underruns. But this is the exception and there'd be audio hick-ups anyway, I don't think trying to improve the play cursor position in this case is worthwhile.

> so maybe we should put into the mAudioDeliveryQueue only the number of frames that we were able to get from the current mPlaybackBuffers

Isn't `len` exactly this? See `len = mPlaybackBuffers[n]->Get(`

> why not currentlyAvailableFramesAcrossBuffers and numberOfRetrievedFrames respectively

Good names, I'll address.

au3/libraries/lib-audio-devices/IMeterSender.h
saintmatthieu: > we store the time at which we are creating that buffer

I do neglect the time the execution of the entire hardware callback takes, yes, and I don't think we should worry about this. If we profile this callback's duration I guess it wouldn't take more than a ms or two, which is negligible for our purpose. We are also neglecting other things, like the time it takes for the acoustic wave to reach the user's ear, the time it takes for the screen to actually render the image Qt hands over to it, or other sources of latency on either the audio or video side of things. I think these approximations could only hurt the result very marginally and this is the best we can reasonably do anyway.

I would start being concerned if this PR has logical mistakes or approximations that could yield an error of more than 20ms. I think anything under that would be very hard for a user to distinguish.

> We still have the problem if the framesPerBuffer isn't met

I responded to this, but probably you didn't have this response at the time of writing.

saintmatthieu: > we will be sending twice to the input vu meters

Not sure I understand: because of that `len < framesPerBuffer` concern you had?

src/context/internal/playbackstate.h
saintmatthieu: I do like that, too, because
* no guesswork is needed,
* if we decide to have a "global", "audio", or other module which already exists in the framework, we wouldn't have to deal with path ambiguities
* if we upgrade our formatting to sort the includes alphabetically, framework includes would automatically get grouped together.

I think in this instance I pre-pended "framework" because `#include "context/..." was mixed with framework includes.

src/auaudio/internal/itimer.h
saintmatthieu: Maybe. I actually was planning a follow-up commit to extract the `IMeter` API and implementation away from `au3wrap`. Then, the timer was just an `internal` thing and didn't need to be exposed, the only API directly under `auaudio` would have been `iaudiometer`, hence the choice of `auaudio`.
"audio" is very general, on the other hand, I'd be open to something more specific.

au3/libraries/lib-audio-io/AudioIO.h
saintmatthieu: > please add the sample rate

Why, if it is guaranteed to stay the same across all deliveries of one session?

saintmatthieu: `AudioCallbackPayloadInfo` works for me.

saintmatthieu: ... although: Just `AudioCallbackInfo` would be enough, I think, and `AudioCallbackPayloadInfo` makes things become a bit lengthy: `GetAudioCallbackPayloadInfoQueue()`

src/au3wrap/internal/au3audiometer.cpp
saintmatthieu: I this this wouldn't make a difference: `QTimer::stop` is executed synchronously (no more callback), while `QTimer::start` posts an event for later.
But the narrative would be clearer, so ok.

src/au3wrap/internal/au3audiometer.h
saintmatthieu: I like `m_runningTimer` and `m_stoppingTimer` :thumbs-up:

src/playback/internal/au3/au3player.cpp
saintmatthieu: We could.

src/playback/internal/au3/au3player.cpp
saintmatthieu: I think you're right, our `audioEngine` impl should do that. Refactoring this should be quite easy, too. Yet this branch has been in the air for a while now, maybe we can find some more appropriate time for this later?

src/playback/internal/au3/au3player.h
saintmatthieu: Replied elsewhere.

src/playback/internal/au3/au3player.h
saintmatthieu: I'm surprised to read this, maybe I misunderstand. Could you elaborate?

src/playback/view/common/playbackpositiontimer.h
saintmatthieu: This hadn't occurred to me. Your suggestion got me enthusiastic, but trying I realize that I also need to call `updatePlaybackPosition` on `IPlayer` anyway, so I don't think it's going to save me code in the end.

au3/libraries/lib-audio-devices/IMeterSender.h
saintmatthieu: > sampleRate should be sent only when first setup and when its changing so we could start/stop/start/stop ...

Like I said, at least implementation with the current API doesn't have to verify that a setup or configure method was called between the last stop and the new start. And whether there was a sample-rate change or not between two playback sessions shouldn't be a concern. We do recalculate everything in AudioIO.cpp when opening a new stream: introducing checks whether it changed since the previous session would be unnecessarily cumbersome. I think this applies here too.

au3/libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: > each time a new buffer is "sent" outside we should accordingly notify and we would then update the visuals accordingly knowing when the buffer has been sent out and what was it size and the sample rate is sufficient to animate a playhead

This is the way it's currently working on master. The problem is that it conditions the visual update rate of the play cursor to the callback rate of the soundcard, making the animation everything but smooth for buffer-size settings more than 25ms (our default is 100ms at the moment).

Now a dedicated timer (obviously on the main thread) is updating the playback position at whatever rate it pleases. In turn, these updates have to call `UpdateTimePosition`, which very nicely abstracts away the problem of looping (work from @Paul-Licameli some years ago, I believe, thanks for that).

You expressed [here](https://github.com/audacity/audacity/pull/9613#discussion_r2451939738) that you didn't feel comfortable with `Au3Player` being responsible for maintaining the cursor's position. Would you feel better if the `AudioEngine` were making these calls? I suggested we postpone that refactoring, but I'm open to giving it a go now.

au3/libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: > we might better use the stream->sampleRate when doing that to have accurate size

Agreed, will address.

saintmatthieu: About the `outputLatency` : why not, will also address.

au3/libraries/lib-audio-io/AudioIO.h
saintmatthieu: `PushMainMeterValues`: funny, I first read the first line of your comment, thought about it, came up with `PushMasterOutputMeterValues`, to find out when reading the second line of your comment that it's exactly what you came up with, too. Looks like we have an agreement here `:D`

`PushTrackMeterValues`: I think that's fine for now, as this method respects the UI: tracks have meters that can be input or output depending on the playback state. But for sure, if this change at some time, we will have to reflect this change here.

au3/libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: > also the mHardwarePlaybackLatencyFrames is really only needed in the AllocateBuffers method and there we could be using the PA info struct

Calculated in `StartPortAudioStream` and then used only in `AllocateBuffers`, correct.
So IIUC you suggest that we make a copy of the PA info struct ? Would work as well, but this starts feeling like not-so-important refactoring.

saintmatthieu:
@luapmartin rebase done. AFAICS I addressed or responded to all comments so far, please let me know if you find something I missed. Thanks for the careful review so far!

au3/libraries/lib-audio-devices/IMeterSender.h
luapmartin: what about `setupAndStart(double sampleRate)` (you can choose to have inner setup(double sampleRate) (or setSampleRate(double sampleRate)) and start() methods or not) 
just making 1 step clearer the intent of this method

au3/libraries/lib-audio-io/AudioIO.cpp
luapmartin: "yes, I like UI stupid to just repaint updated states as much as possible" a kind of rule of thumb for me, let's see what we can do (we could have a call to decide if now or later (or never XD))

src/playback/internal/au3/au3player.cpp
luapmartin: maybe could you please create a Task for that extra cleanup?

au3/libraries/lib-audio-io/AudioIO.cpp
luapmartin: thanks for renaming and all other edits, now we reach the point where I could maybe express clearer my thoughts 

we have:
```cpp
using namespace std::chrono;
const auto now = steady_clock::now();
const auto adcTime = now + milliseconds(static_cast<int>(mHardwarePlaybackLatencyMs));
mAudioCallbackInfoQueue.Put({ adcTime, static_cast<int>(numberOfRetrievedFrames) });
```
what I don't like here is that the adcTime is not always when this numberOfRetrievedFrames will be sent, 

what we know for sure at that time of the callback is that the time of the callback is now() and we have sample rate X and we have numberOfRetrievedFrames, apart from that the rest is uncertain info, that's why I would prefer pack this info into a useful struct and push it to a ringbuffer and on another thread compute the timings etc. or whatever but use the thrustful data only somehow

luapmartin (approved):
let's create those escalation tickets

luapmartin:
last added commit might be fixing : https://github.com/audacity/audacity/issues/8810

Issue #9219, opened 2025-08-12
Sort the Effects in AU4 by Importance, not Alphabetical order.


chinakov:
### Description

The AU4 effects are sorted in alphabetical order, but they should be ordered by importance, just like in AU3, like so:

- Volume and Compression
- Fading
- Pitch and Tempo
- EQ and Filters
- Noise Removal and Repair
- Delay and Reverb
- Distortion and Modulation
- Special
- Spectral Tools
- Legacy

chinakov:
This is still present on the latest Master (e9ead184ef595365efaef0ae54167634b0fd8da7)

JoshBashed:
#9641 should fix this.

dozzzzer:
Tested

PR #9641, opened 2025-10-11
9219 Sort builtin effects in menu

JoshBashed:
Resolves: https://github.com/audacity/audacity/issues/9219

The AU4 effects are sorted in alphabetical order, but they should be ordered by importance, just like in AU3. Now they are.

Recommended:

QA:

Note:
I ran it through the linter and it changed another line of code unrelated.

### DIFF
--- a/src/effects/builtin/internal/builtineffectsrepository.cpp
+++ b/src/effects/builtin/internal/builtineffectsrepository.cpp
@@ -15,6 +15,9 @@
 
 #include "general/generalviewmodel.h"
 
+#include "effects/effects_base/effectstypes.h"
+#include "effects/effects_base/internal/effectsutils.h"
+
 #include "amplify/amplifyeffect.h"
 #include "amplify/amplifyviewmodel.h"
 #include "loudness/normalizeloudnesseffect.h"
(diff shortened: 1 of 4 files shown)


JoshBashed:
If I wanted to go more indepth, I could have just edited `EffectMeta` but I didn't want to modify a whole struct for one use case. Also this only runs a few times.

JoshBashed:
@dozzzzer Hi, just wanted to check in. It’s been almost two weeks on this 44-line PR, and I haven’t seen any comments yet. No rush if @grliszas14, @kryksyh , @luapmartin, or @saintmatthieu are busy, but a quick note or review would be appreciated.

src/effects/builtin/internal/builtineffectsrepository.h
```diff
@@ -13,6 +13,8 @@
 #include "libraries/lib-components/ComponentInterfaceSymbol.h"
 
 namespace au::effects {
+int categoryIdOrder(const muse::String& category);
```
saintmatthieu: It looks like this is only used in `effectsutils.cpp`, correct? The function's body isn't big either, so I would remove the declaration from any header and have the function there (preferably in an anonymous namespace).

saintmatthieu:
Hi @JoshBashed , thanks for the reminder and the contribution. It's a small and neat one, I recently repeatedly thought it'd be nice to have it done.
I added a single comment, once this is addressed you'll get my approval and the ticket can move to QA.

src/effects/builtin/internal/builtineffectsrepository.h
JoshBashed: Thanks for the review. One detail to confirm: `categoryIdOrder` currently calls `categoryIdString`, which is private to `builtineffectsrepository.cpp`. Are you suggesting I move `categoryIdOrder` into `effectsutils.cpp`? If so, that would require either exposing `categoryIdString` or duplicating that mapping. Could you clarify your preference?

saintmatthieu: Oh, I missed that.
The more sensible would be to do it the other way round, then: migrating `categoryIdString` to `effectsutils.h` and `.cpp`, and reuse it in builtineffectsrepository (no duplication). Thank you.

JoshBashed: `EffectCategoryId` is an internal type.

JoshBashed: @saintmatthieu would you like me to move that too?

saintmatthieu: That too? Right.
effectstypes.h would be a fitting home for `EffectCategoryId`.

JoshBashed: @saintmatthieu the issue has been fixed. LMK if you would like any more hcanges

Issue #9655, opened 2025-10-14
Spacebar Play/Pause Shortcut doesn't Work


chinakov:
### Bug description

Spacebar Shortcut is no longer functional.

### Steps to reproduce

1. Launch AU4
2. Create a new project
3. Generate or Import a clip
4. Press 'Spacebar' to start Playback
5. Observe result

### Expected behavior

Playback does not start. 

### Actual behavior

Playback should start. 

### Audacity Version

61485abd33e86f8988da09b6cb1e402bf052fb99

### Operating system

OS independent

### Additional context

Most probably regression introduced by 61485abd33e86f8988da09b6cb1e402bf052fb99

luapmartin:
closed #9658 as it will be handled by @igorkorsukov

enzogianvittorio:
Yes I am experimenting the very same issue and have tried to report it as a bug but it seems this report is about the same (am I right ?) is quite difficult to work editing audio in Linux with this problem, I hope this issue could be sorted out at some point.
Many thanks to all developers !

dozzzzer:
@enzogianvittorio 

What version are you on?

enzogianvittorio:
Hi, I am using Audacity 3.7.7
running on MX Linux 25.1 (Debian based distro) wiht Xfce desktop environment.

MegaGamer22:
I can still reproduce this in `4.0.0.262451524`, it seems to happen only in some sessions.

PR #9662, opened 2025-10-15
returned match ui context  playback - projectfocused logic

igorkorsukov:
Resolves: https://github.com/audacity/audacity/issues/9655

I've returned the logic for comparing the "Project-Playback" and "Project-Focus" ui contexts to as was. It doesn't match the comments and doesn't match how it should be.

### DIFF
--- a/src/context/internal/uicontextresolver.cpp
+++ b/src/context/internal/uicontextresolver.cpp
@@ -138,8 +138,9 @@ bool UiContextResolver::match(const ui::UiContext& currentCtx, const ui::UiConte
         return true;
     }
 
+    //! SEE There's a problem here, see https://github.com/audacity/audacity/pull/9662#issuecomment-3405088750
     //! NOTE If the current context is `UiCtxProjectPlayback`, then we allow `UiCtxProjectFocused` too
-    if (currentCtx == context::UiCtxProjectPlayback && actCtx == context::UiCtxProjectFocused) {
+    if (currentCtx == context::UiCtxProjectFocused && actCtx == context::UiCtxProjectPlayback) {
         return true;
     }
 


src/context/internal/uicontextresolver.cpp
```diff
@@ -139,7 +139,7 @@ bool UiContextResolver::match(const ui::UiContext& currentCtx, const ui::UiConte
     }
 
     //! NOTE If the current context is `UiCtxProjectPlayback`, then we allow `UiCtxProjectFocused` too
-    if (currentCtx == context::UiCtxProjectPlayback && actCtx == context::UiCtxProjectFocused) {
+    if (currentCtx == context::UiCtxProjectFocused && actCtx == context::UiCtxProjectPlayback) {
         return true;
     }
```
luapmartin, line 145: ok, I see what you did there, thanks, but maybe a comment about the fact that this is a temporary fix would help (because the NOTE is now out of phased with the implementation) ? 

Also, we should add a link to the proper issue that will address the root core issue we are facing with those contexts, with proper details about the why and how for anyone to start working on this issue, because as I understand we **will** fix forward that issue properly. 
What do you think?

igorkorsukov:
A little about the problem. In audio applications, it's common to activate  Play/Stop via Spacebar. But in MacOS, it's also common to activate controls via Spacebar. All other operating systems via Enter, so the problem is specific to MacOS.

What we have: one shortcut Space and two actions: navigation/trigger and playback/toggle 
 
As I understand it, we want to send playback/toggle if: we are in the project area (project focused?), if the effects dialog is open (or not just open, but also in focus?), if the VST (plugins?) dialog is open, or in focus? 
i.e. something like this: send playback/toggle if context Project Focused or EffectsFocused or PluginFocused
Else send navigation/trigger
The requirements may be slightly different, but the essence is the same.

How to solve the problem?
1. Improve UIAction so that it accepts a combination of contexts, and the context resolver can handle it.
2. Enter synthetic contexts, something like "ProjectFocused_EffectsFocused_PluginFocused" (or even "PlayContext"), BUT it can't be current. It's specified in UIAction, and a context resolver resolves it (MuseScore has one like that - "ProjectNotFocused").
3. Send one action action://activate (global) via Space and some controller processes it and decides what to send next.

This problem isn't currently addressed correctly. A new current context has been added (UiCtxProjectPlayback) that supposedly answers the question "what to do," while the current context should answer the question "where are we?" Some shaky logic has also been added to ensure the behavior is as expected. But shaky logic prevents the system from evolving; everything will break with any change.

PR #9666, opened 2025-10-15
Do not restart monitoring if playback is paused

### LINKED ISSUE #9647: Playback Cannot be Resumed
### Bug description

If the User has a track selected, and that selected track is able to be recorded to and then starts playback, then pauses, and then resumes once again, the playback won't be resumed. 

### Steps to reproduce

1. Launch AU4
2. Record a clip
3. Start playback
4. Pause playback
5. Start playback again
6. Observe result

### Expected behavior

The playback should always resume after pause. 

### Actual behavior

The playback is not resumed after pause. 

### Audacity Version

4f8bf715d6dd933619497e4527e4b298c13bf222

### Operating system

OS independent

### Additional context

https://github.com/user-attachments/assets/18be0a53-d0d8-4068-9456-35813dcfe8d1

embarc-gabriel:
Resolves: #9647 

Do not restart monitoring if playback is paused.

Recommended:

QA:

### DIFF
--- a/au3/libraries/lib-audio-io/AudioIO.cpp
+++ b/au3/libraries/lib-audio-io/AudioIO.cpp
@@ -1906,6 +1906,11 @@ void AudioIO::AudioThread(std::atomic<bool>& finish)
                 gAudioIO->mAudioThreadAcknowledge.store(Acknowledge::eStop,
                                                         std::memory_order::memory_order_release);
             }
+            else if (lastState == State::eDoNothing) {
+                //Avoid multiple stops in a row
+                gAudioIO->mAudioThreadAcknowledge.store(Acknowledge::eStop,
+                                                        std::memory_order::memory_order_release);
+            }
             lastState = State::eDoNothing;
 
             if (gAudioIO->IsMonitoring()) {
--- a/src/record/internal/au3/au3audioinput.cpp
+++ b/src/record/internal/au3/au3audioinput.cpp
@@ -232,7 +232,7 @@ int Au3AudioInput::getFocusedTrackChannels() const
 
 void Au3AudioInput::restartMonitoring()
 {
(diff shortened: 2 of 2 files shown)


luapmartin (approved):

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -1906,6 +1906,11 @@ void AudioIO::AudioThread(std::atomic<bool>& finish)
                 gAudioIO->mAudioThreadAcknowledge.store(Acknowledge::eStop,
                                                         std::memory_order::memory_order_release);
             }
+            else if (lastState == State::eDoNothing) {
+                //Avoid multiple stops in a row
+                gAudioIO->mAudioThreadAcknowledge.store(Acknowledge::eStop,
+                                                        std::memory_order::memory_order_release);
+            }
```
embarc-gabriel, line 1913: @saintmatthieu @luapmartin 
I was talking to Paul and the freeze happens because multiple places are calling the `StopStream` function in a row (including `StartStream`)
The `StopStream` executes `WaitForAudioThreadStopped` that checks `mAudioThreadAcknowledge` to be equal to eStop.
Before the first execution, `lastState` is `Monitoring` or whatever, and we set it to eDothing, and `mAudioThreadAcknowledge` is eNone.
On the subsequent execution, `lastState` is already `eDoNothing`, so `mAudioThreadAcknowledge` does not transition to `eStop`, and `WaitForAudioThreadStopped` never returns.
The above is just a quick workaround, not the final solution. Please don't hesitate to change it.
I was talking to Paul, and we probably should do a little refactoring to avoid calling so many `AudioIO` related functions in different places and make the API lean.

embarc-gabriel:
The fix is causing multiple different issues and the workaround does not work for Windows.
I'll close it here until we find a better approach.

Issue #9619, opened 2025-10-09
Add metadata dialog


grliszas14:
### Description

https://www.figma.com/design/QaCzd2bM08W8VWgCLsKJ2u/32---Export?node-id=1650-28366&m=dev

dozzzzer:
@grliszas14 
1. The value column is sortable, which, to my mind, is unnecessary. Can we remove that? Also, the "Tag" column should probably not be sortable, too.
2. The vertical line that should split the columns is missing. As a result, it currently looks more like a list of values than a table.
3. The fact that every 'tag' value is editable only serves to reinforce the impression that this is a list rather than a tag-value table.
4. When sorting is applied, if you edit one of the tags by pasting some text, it will move down by 1 line, and the value will be saved for a different tag instead:

https://github.com/user-attachments/assets/dbd3f2cc-5fdc-4070-90d4-4de4cf4c2648

5. Sorting works incorrectly; first, it sorts the strings starting with capital letters and only then the small ones:

<img width="400" height="653" alt="Image" src="https://github.com/user-attachments/assets/b24f8595-ae82-4449-85e7-46a675b212cc" />

6. "Software" field seems redundant; the names of the rest of the tags are different from the designs (Album title, Artist name, etc.)

grliszas14:
Fixed point no.6, the rest needs to be handled separately in the framework

chinakov:
Point number 6 is fixed, this is ready for merging. Thanks!

PR #9689, opened 2025-10-17
Introduce metadata dialog

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9619

### DIFF
--- a/src/importexport/export/view/metadatamodel.h
+++ b/src/importexport/export/view/metadatamodel.h
@@ -59,17 +59,14 @@ public:
     bool isMetadataEmpty(const au::project::ProjectMeta& meta) const;
 
 private:
-    inline static const std::array<std::string, 9> kStdTags =
+    inline static const std::array<std::string, 6> kStdTags =
     {
-        muse::trc("metadata", "TITLE"),
-        muse::trc("metadata", "ARTIST"),
-        muse::trc("metadata", "ALBUM"),
-        muse::trc("metadata", "TRACKNUMBER"),
-        muse::trc("metadata", "YEAR"),
-        muse::trc("metadata", "GENRE"),
-        muse::trc("metadata", "COMMENTS"),
-        muse::trc("metadata", "Software"),
-        muse::trc("metadata", "Copyright"),
+        muse::trc("metadata", "Artist name"),
+        muse::trc("metadata", "Track title"),
+        muse::trc("metadata", "Album title"),
+        muse::trc("metadata", "Track number"),
+        muse::trc("metadata", "Year"),
+        muse::trc("metadata", "Comments"),
     };
 
     project::ProjectMeta m_meta;
--- a/src/project/internal/au3/au3metadata.h
+++ b/src/project/internal/au3/au3metadata.h
@@ -28,17 +28,14 @@ public:
     project::ProjectMeta parseXml(const std::string& xml) const override;
 
 private:
-    inline static const std::array<std::string, 9> kStdTags =
+    inline static const std::array<std::string, 6> kStdTags =
     {
-        muse::trc("metadata", "TITLE"),
-        muse::trc("metadata", "ARTIST"),
-        muse::trc("metadata", "ALBUM"),
-        muse::trc("metadata", "TRACKNUMBER"),
-        muse::trc("metadata", "YEAR"),
-        muse::trc("metadata", "GENRE"),
-        muse::trc("metadata", "COMMENTS"),
-        muse::trc("metadata", "Software"),
-        muse::trc("metadata", "Copyright"),
+        muse::trc("metadata", "Artist name"),
+        muse::trc("metadata", "Track title"),
+        muse::trc("metadata", "Album title"),
+        muse::trc("metadata", "Track number"),
+        muse::trc("metadata", "Year"),
+        muse::trc("metadata", "Comments"),
     };
 };
 }
--- a/src/project/internal/au3/au3metadata.cpp
+++ b/src/project/internal/au3/au3metadata.cpp
@@ -24,8 +24,8 @@ ProjectMeta Au3Metadata::tags() const
 
     for (auto row : au3Tags.GetRange()) {
         auto tag = row.first.ToStdString();
-        if (tag == TITLE_TAG) {
-            au4Tags.title = row.second.ToStdString();
+        if (tag == TRACK_TITLE_TAG) {
+            au4Tags.trackTitle = row.second.ToStdString();
         }
 
         const std::string key = row.first.ToStdString();
--- a/src/project/types/projectmeta.h
+++ b/src/project/types/projectmeta.h
@@ -16,15 +16,12 @@ struct ProjectMeta
 {
     muse::io::path_t filePath;
 
-    std::string title;
     std::string artist;
+    std::string trackTitle;
     std::string album;
     std::string trackNumber;
     std::string year;
-    std::string genre;
     std::string comments;
-    std::string software;
-    std::string copyright;
 
     QPixmap thumbnail;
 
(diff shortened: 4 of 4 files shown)


src/importexport/export/internal/au3/au3exporter.cpp
```diff
@@ -270,6 +273,19 @@ bool Au3Exporter::isCustomFFmpegExportFormat() const
     return editor->GetName() == "custom_ffmpeg";
 }
 
+bool Au3Exporter::canMetadata() const
```
Eism: I would suggest renaming the method because it is not clear what it does.
Maybe hasMetadata

src/importexport/export/qml/Export/MetadataDialog.qml
```diff
@@ -0,0 +1,416 @@
...
+        metadataModel.load()
+    }
+
+    ColumnLayout {
+        id: mainColumn
+
+        spacing: 0
+
+        RowLayout {
+            id: topButtonsBar
```
Eism, line 47: Please move it to a separate file and don't forget about navigation

src/importexport/export/qml/Export/MetadataDialog.qml
```diff
@@ -0,0 +1,416 @@
...
+
+            Layout.fillHeight: true
+            Layout.fillWidth: true
+
+            Layout.leftMargin: 16
+            Layout.rightMargin: 12
```
Eism, line 53: should be the same as leftMargin?

src/importexport/export/qml/Export/MetadataDialog.qml
```diff
@@ -0,0 +1,416 @@
...
+
+                    // highlight right away
+                    tagView.currentIndex = metadataModel.rowCount() - 1
+                }
+            }
+        }
+
+        SeparatorLine {}
+
+        ColumnLayout {
```
Eism: Also, please move to a separate file

src/importexport/export/view/metadatamodel.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+    m_meta.additionalTags.insert(key, value);
+    const QModelIndex idx = index(row, 0);
+    emit dataChanged(idx, idx, { RoleValue });
+}
+
+QString MetadataModel::buildXml(bool autoFormat) const
+{
+    QString xml;
+    QXmlStreamWriter w(&xml);
```
Eism: can we use serialization/xmlstreamwriter.h ?

src/importexport/export/qml/Export/MetadataDialog.qml
grliszas14: not my idea, I just took the values from the design :)

src/importexport/export/qml/Export/internal/MetadataView.qml
Eism, line 1: no longer needed?

src/importexport/export/qml/Export/internal/MetadataControlPanel.qml
```diff
@@ -0,0 +1,126 @@
...
+RowLayout {
+    id: root
+
+    Layout.fillHeight: true
+    Layout.fillWidth: true
+
+    Layout.leftMargin: 16
+    Layout.rightMargin: 12
+    Layout.topMargin: 10
+    Layout.bottomMargin: 10
```
Eism: It's better to set such properties where you use this component.
1. When reading the qml file where you use it, it will be easier to understand how the element is positioned relative to other elements.
2. If we want to reuse this component, we'll need to override these properties. We won't reuse this anywhere except in the current location, this is more of a generally approach in qml.

src/importexport/export/qml/Export/internal/MetadataControlPanel.qml
```diff
@@ -0,0 +1,126 @@
...
+
+    spacing: 6
```
Eism: please move this below the properties

src/importexport/export/qml/Export/internal/MetadataControlPanel.qml
```diff
@@ -0,0 +1,126 @@
...
+
+        onClicked: {
+            metadataModel.saveTemplate()
+        }
+    }
+
+    FlatButton {
+        id: setDefaultBtn
+
+        Layout.preferredWidth: 103
```
Eism: there are so many different widths, can we not hardcode them?

src/importexport/export/qml/Export/internal/MetadataControlPanel.qml
```diff
@@ -0,0 +1,126 @@
...
+        navigation.panel: root.navigationPanel
+        navigation.column: saveBtn.navigation.column + 1
+
+        onClicked: {
+            metadataModel.setAsDefault()
+        }
+    }
+
+    SeparatorLine {
+        orientation: Qt.Vertical
```
Eism: you don't have to set this property, the component will determine it itself

src/importexport/export/qml/Export/MetadataDialog.qml
```diff
@@ -0,0 +1,129 @@
...
+
+            keyRoleName: "tag"
+            keyTitle: qsTrc("metadata", "tag")
+            valueRoleName: "value"
+            valueTitle: qsTrc("metadata", "value")
+            valueTypeRole: "roleType"
+            readOnly: false
+
+            width: root.contentWidth
+            height: 527
```
Eism: Please move this below the id. 
When reading the qml code, the first thing you're interested in is how this element is positioned, so it's important to position such properties as high as possible.

Eism: Also, It is better to set the sizes as a multiple of 4

src/importexport/export/qml/Export/MetadataDialog.qml
```diff
@@ -0,0 +1,129 @@
...
+            id: buttonBox
+
+            Layout.fillWidth: true
+
+            padding: 8
+
+            navigationPanel.section: root.navigation.section
+            navigationPanel.order: metadataView.navigationOrderStart + 2
+
+            FlatButton {
```
Eism, line 108: maybe you can use buttons property

```
buttons: [ButtonBoxModel.Close, ButtonBoxModel.Apply]
onStandardButtonClicked: function (buttonId) {
```

src/importexport/export/view/metadatamodel.h
```diff
@@ -0,0 +1,81 @@
...
+
+    QString buildXml() const;
+    project::ProjectMeta parseXml(const QString& xml) const;
+
+    bool isMetadataEmpty(const au::project::ProjectMeta& meta) const;
+
+private:
+    inline static const std::array<QString, 9> kStdTags =
+    {
+        QString("TITLE"),
```
Eism: should these strings be translatable?

src/importexport/export/qml/Export/MetadataDialog.qml
grliszas14: I left it for now as I can't make Apply button to be of accentColor this way (+ I cannot tweak the button sizes)

src/project/internal/au3/au3metadata.h
```diff
@@ -0,0 +1,44 @@
...
+    muse::Inject<au::context::IGlobalContext> globalContext;
+    muse::Inject<muse::IInteractive> interactive;
+
+public:
+    void init();
+
+    project::ProjectMeta tags() const override;
+    void setTags(project::ProjectMeta) override;
+
+    std::string buildXml(const project::ProjectMeta&) const override;
```
Eism, line 27: can be static

PR #9690, opened 2025-10-17
9647 playback cannot be resumed

luapmartin:
Resolves: #9647 

metering only enabled when in stop mode

Recommended:

QA:
- [ ] With mic metering enabled, check the input monitoring option: input monitoring should work

### DIFF
--- a/au3/libraries/lib-audio-devices/AudioIOBase.h
+++ b/au3/libraries/lib-audio-devices/AudioIOBase.h
@@ -63,7 +63,7 @@ struct AudioIOStartStreamOptions
     double preRoll{ 0.0 };
 
     bool playNonWaveTracks{ true };
-    mutable bool inputMonitoring{ false };
+    bool inputMonitoring{ false };
 
     // contents may get swapped with empty vector
     PRCrossfadeData* pCrossfadeData{};
--- a/au3/libraries/lib-audio-io/AudioIO.cpp
+++ b/au3/libraries/lib-audio-io/AudioIO.cpp
@@ -809,7 +809,14 @@ void AudioIO::StartMonitoring(const AudioIOStartStreamOptions& options)
     if (IsBusy()) {
         return;
     }
+
+    if (IsMonitoring() && options.inputMonitoring != mSoftwarePlaythrough) {
+        // muting or unmuting monitoring, so we need to restart it
+        StopMonitoring();
+    }
+
     if (IsMonitoring()) {
+        // already monitoring, nothing to do
         return;
     }
 
@@ -828,14 +835,14 @@ void AudioIO::StartMonitoring(const AudioIOStartStreamOptions& options)
     mCaptureFormat = captureFormat;
     mCaptureRate = 44100.0; // Shouldn't matter
     const bool success = StartPortAudioStream(options,
-                                        static_cast<unsigned int>(playbackChannels),
-                                        static_cast<unsigned int>(captureChannels));
+                                              static_cast<unsigned int>(playbackChannels),
+                                              static_cast<unsigned int>(captureChannels));
 
     const auto pOwningProject = mOwningProject.lock();
     if (!success) {
         using namespace BasicUI;
         const auto msg = XO("Error opening recording device.\nError code: %s")
-                   .Format(Get()->LastPaErrorString());
+                         .Format(Get()->LastPaErrorString());
         ShowErrorDialog(*ProjectFramePlacement(pOwningProject.get()),
                         XO("Error"), msg, wxT("Error_opening_sound_device"),
                         ErrorDialogOptions { ErrorDialogType::ModalErrorReport });
@@ -901,6 +908,8 @@ int AudioIO::StartStream(const TransportSequences& sequences,
         return 0;
     }
 
+    StopMonitoring();
+
     // We just want to set mStreamToken to -1 - this way avoids
     // an extremely rare but possible race condition, if two functions
     // somehow called StartStream at the same time...
@@ -916,14 +925,6 @@ int AudioIO::StartStream(const TransportSequences& sequences,
     //       playbackChannels == mNumPlaybackChannels &&
     //       captureChannels == mNumCaptureChannels &&
     //       captureFormat == mCaptureFormat) {
-
-    if (mPortStreamV19) {
-        StopStream();
-        while (mPortStreamV19) {
-            using namespace std::chrono;
-            std::this_thread::sleep_for(50ms);
-        }
-    }
 
     mSoftwarePlaythrough = options.inputMonitoring;
     mPauseRec = SoundActivatedRecord.Read();
@@ -2355,6 +2356,8 @@ void AudioIO::DrainRecordBuffers()
         // AudacityException::DelayedHandlerAction prevents redundant message
         // boxes.
         StopStream();
+        WaitWhileBusy();
+
         DefaultDelayedHandlerAction(pException);
         for (auto& pSequence: mCaptureSequences) {
             pSequence->RepairChannels();
--- a/src/au3audio/iaudioengine.h
+++ b/src/au3audio/iaudioengine.h
@@ -22,11 +22,11 @@ public:
     virtual bool isBusy() const = 0;
 
     virtual int startStream(const TransportSequences& sequences, double startTime, double endTime, double mixerEndTime, // Time at which mixer stops producing, maybe > endTime
-                            const AudioIOStartStreamOptions& options) = 0;
+                            AudacityProject& project, bool isDefaultPlayTrackPolicy, double audioStreamSampleRate) = 0;
     virtual void stopStream() = 0;
     virtual void pauseStream(bool pause) = 0;
 
-    virtual void startMonitoring(const AudioIOStartStreamOptions& options) = 0;
+    virtual void startMonitoring(AudacityProject& project) = 0;
     virtual void stopMonitoring() = 0;
 
     virtual muse::async::Notification updateRequested() const = 0;
--- a/src/record/internal/au3/au3audioinput.cpp
+++ b/src/record/internal/au3/au3audioinput.cpp
@@ -142,7 +142,7 @@ void Au3AudioInput::startAudioEngineMonitoring() const
         if (!project) {
             return;
         }
-        audioEngine()->startMonitoring(ProjectAudioIO::GetDefaultOptions(*project));
+        audioEngine()->startMonitoring(*project);
     });
 }
 
--- a/src/record/view/toolbars/playbacktoolbarrecordlevelitem.cpp
+++ b/src/record/view/toolbars/playbacktoolbarrecordlevelitem.cpp
@@ -12,10 +12,10 @@ using namespace au::audio;
 using namespace au::playback;
 
 PlaybackToolBarRecordLevelItem::PlaybackToolBarRecordLevelItem(const muse::ui::UiAction& action,
-                                                               muse::uicomponents::ToolBarItemType::Type type, QObject* parent)
+                                                               const muse::uicomponents::ToolBarItemType::Type type, QObject* parent)
     : muse::uicomponents::ToolBarItem(action, type, parent)
 {
-    record()->audioInput()->recordVolumeChanged().onReceive(this, [this](audio::volume_dbfs_t volume){
+    record()->audioInput()->recordVolumeChanged().onReceive(this, [this](const audio::volume_dbfs_t volume){
         m_level = volume;
         emit levelChanged();
     });
@@ -37,9 +37,6 @@ PlaybackToolBarRecordLevelItem::PlaybackToolBarRecordLevelItem(const muse::ui::U
     });
 
     resetAudioChannelsVolumePressure();
-    emit recordingChannelsCountChanged();
-    // TODO not sure this 'emit' is needed, should we then also emit isMicMeteringOnChanged(); ?
-    // emit isInputMonitoringOnChanged();
 }
 
 float PlaybackToolBarRecordLevelItem::level() const
@@ -47,7 +44,7 @@ float PlaybackToolBarRecordLevelItem::level() const
     return m_level;
 }
 
-void PlaybackToolBarRecordLevelItem::setLevel(float newLevel)
+void PlaybackToolBarRecordLevelItem::setLevel(const float newLevel)
 {
     if (qFuzzyCompare(m_level, newLevel)) {
         return;
@@ -66,7 +63,7 @@ float PlaybackToolBarRecordLevelItem::rightChannelPressure() const
     return m_rightChannelPressure;
 }
 
-void PlaybackToolBarRecordLevelItem::setLeftChannelPressure(float leftChannelPressure)
+void PlaybackToolBarRecordLevelItem::setLeftChannelPressure(const float leftChannelPressure)
 {
     if (qFuzzyCompare(m_leftChannelPressure, leftChannelPressure)) {
         return;
@@ -76,7 +73,7 @@ void PlaybackToolBarRecordLevelItem::setLeftChannelPressure(float leftChannelPre
     emit leftChannelPressureChanged(m_leftChannelPressure);
 }
 
-void PlaybackToolBarRecordLevelItem::setRightChannelPressure(float rightChannelPressure)
+void PlaybackToolBarRecordLevelItem::setRightChannelPressure(const float rightChannelPressure)
 {
     if (qFuzzyCompare(m_rightChannelPressure, rightChannelPressure)) {
         return;
@@ -106,7 +103,7 @@ float PlaybackToolBarRecordLevelItem::leftRecentPeak() const
     return m_leftRecentPeak;
 }
 
-void PlaybackToolBarRecordLevelItem::setLeftRecentPeak(float newLeftRecentPeak)
+void PlaybackToolBarRecordLevelItem::setLeftRecentPeak(const float newLeftRecentPeak)
 {
     if (qFuzzyCompare(m_leftRecentPeak, newLeftRecentPeak)) {
         return;
@@ -121,7 +118,7 @@ float PlaybackToolBarRecordLevelItem::leftMaxPeak() const
     return m_leftMaxPeak;
 }
 
-void PlaybackToolBarRecordLevelItem::setLeftMaxPeak(float newLeftMaxPeak)
+void PlaybackToolBarRecordLevelItem::setLeftMaxPeak(const float newLeftMaxPeak)
 {
     if (qFuzzyCompare(m_leftMaxPeak, newLeftMaxPeak)) {
         return;
@@ -136,7 +133,7 @@ float PlaybackToolBarRecordLevelItem::rightRecentPeak() const
     return m_rightRecentPeak;
 }
 
-void PlaybackToolBarRecordLevelItem::setRightRecentPeak(float newRightRecentPeak)
+void PlaybackToolBarRecordLevelItem::setRightRecentPeak(const float newRightRecentPeak)
 {
     if (qFuzzyCompare(m_rightRecentPeak, newRightRecentPeak)) {
         return;
@@ -151,7 +148,7 @@ float PlaybackToolBarRecordLevelItem::rightMaxPeak() const
     return m_rightMaxPeak;
 }
 
-void PlaybackToolBarRecordLevelItem::setRightMaxPeak(float newRightMaxPeak)
+void PlaybackToolBarRecordLevelItem::setRightMaxPeak(const float newRightMaxPeak)
 {
     if (qFuzzyCompare(m_rightMaxPeak, newRightMaxPeak)) {
         return;
@@ -171,7 +168,7 @@ bool PlaybackToolBarRecordLevelItem::isInputMonitoringOn() const
     return recordConfiguration()->isInputMonitoringOn();
 }
 
-void PlaybackToolBarRecordLevelItem::setIsInputMonitoringOn(bool enable)
+void PlaybackToolBarRecordLevelItem::setIsInputMonitoringOn(const bool enable)
 {
     recordConfiguration()->setIsInputMonitoringOn(enable);
 }
@@ -181,7 +178,7 @@ bool PlaybackToolBarRecordLevelItem::isMicMeteringOn() const
     return recordConfiguration()->isMicMeteringOn();
 }
 
-void PlaybackToolBarRecordLevelItem::setIsMicMeteringOn(bool enable)
+void PlaybackToolBarRecordLevelItem::setIsMicMeteringOn(const bool enable)
 {
     recordConfiguration()->setIsMicMeteringOn(enable);
 }
@@ -191,7 +188,7 @@ PlaybackMeterStyle::MeterStyle PlaybackToolBarRecordLevelItem::meterStyle() cons
     return playbackConfiguration()->playbackMeterStyle();
 }
 
-void PlaybackToolBarRecordLevelItem::listenMainAudioInput(bool listen)
+void PlaybackToolBarRecordLevelItem::listenMainAudioInput(const bool listen)
 {
     recordMeterController()->setRecordMeterVisible(listen);
     if (listen) {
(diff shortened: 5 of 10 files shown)


au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -286,7 +286,7 @@ AudioIO::AudioIO()
     mNumCaptureChannels = 0;
     mSilenceLevel = 0.0;
 
-    mOutputMeter.reset();
+    ResetMeters();
```
luapmartin, line 289: here I reset both, I don't think that's problematic

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -864,6 +856,13 @@ void AudioIO::StartMonitoring(const AudioIOStartStreamOptions& options)
     }
 }
 
+void AudioIO::StopMonitoring()
+{
+    if (IsMonitoring()) {
+        StopStream();
+    }
+}
+
```
luapmartin, line 883: little helper

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -33,39 +33,79 @@ Au3AudioInput::Au3AudioInput()
...
+        };
     });
 }
 
+void Au3AudioInput::updateMonitoring(const MonitoringChangeReason reason)
+{
+    switch (reason) {
+    case MonitoringChangeReason::Initialization:
+        updateMonitoring();
+        break;
```
luapmartin: helper method practical to debug and help understand the flow

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -155,43 +197,53 @@ void Au3AudioInput::startMonitoring()
...
 
-bool Au3AudioInput::shouldRestartMonitoring() const
+bool Au3AudioInput::shouldStartMonitoring() const
 {
     if (audibleInputMonitoring()) {
         return true;
     }
 
-    if (!configuration()->isMicMeteringOn()) {
+    if (!configuration()->isMicMeteringOn()) { // PAUL: this look strange
```
luapmartin: ok I should remove that

au3/libraries/lib-audio-io/AudioIO.cpp
saintmatthieu: A method that's just an if - isn't this abusive? I might prefer to read three lines of code where this method is intended to be used rather than one line that redirects me to three lines.

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -33,39 +33,79 @@ Au3AudioInput::Au3AudioInput()
...
+        // when updating the mic metering we need to either stop or restart the monitoring
+        updateMonitoring();
+        break;
+    case MonitoringChangeReason::FocusedTrackChanged:
+        updateMonitoring();
+        break;
+    case MonitoringChangeReason::RecordMeterVisibilityChanged:
+        // we don't really have to update the monitoring when the meter is shown/hidden
+        //updateMonitoring();
+        break;
```
saintmatthieu: It didn't harm, though, right?
If you still decide to remove this, `meterController()->isRecordMeterVisible()` in `Au3AudioInput::shouldStartMonitoring()` should be assumed to be `true` - the code should be simplified accordingly.

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -33,39 +33,79 @@ Au3AudioInput::Au3AudioInput()
...
+        updateMonitoring();
+        break;
+    case MonitoringChangeReason::AudibleInputMonitoring:
+        // when updating the audible input monitoring we need to either stop or restart the monitoring
+        updateMonitoring();
+        break;
+    }
+}
+
 void Au3AudioInput::initMeter()
```
saintmatthieu: minor: I like the enum, but for the sake of keeping the number of lines reasonable, please consider
```cpp
    switch (reason) {
    case MonitoringChangeReason::Initialization:
    case MonitoringChangeReason::RecordingState:
    // when the playback stops we need to restart the monitoring if mic metering is on or input monitoring is on
    case MonitoringChangeReason::PlaybackState:
    case ...:
        updateMonitoring();
        break;
    case 
```
or smth

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -33,39 +33,79 @@ Au3AudioInput::Au3AudioInput()
...
         });
 
         meterController()->isRecordMeterVisibleChanged().onNotify(this, [this]() {
-            restartMonitoring();
+            updateMonitoring(MonitoringChangeReason::RecordMeterVisibilityChanged);
         });
+
+        audibleInputMonitoringChanged = [&]() {
+            updateMonitoring(MonitoringChangeReason::AudibleInputMonitoring);
+        };
```
saintmatthieu: This pattern is confusing: a member lambda while it could just be a class method (or no method at all since its body is trivial), even without an `m_` prefix. It's the change that had me check out your PR to see what it was :)

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -134,11 +174,13 @@ void Au3AudioInput::setAudibleInputMonitoring(bool enable)
...
-
-    restartMonitoring();
+    // Is there notification being sent when adjusting Prefs?
+    // we need to find a way to notify that the monitoring state has changed
+    // we could add a local lambda callback
+    audibleInputMonitoringChanged();
 }
 
-void Au3AudioInput::startMonitoring()
+void Au3AudioInput::startMonitoring() const
```
saintmatthieu: :thumbs-up:

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -155,43 +197,53 @@ void Au3AudioInput::startMonitoring()
...
-        using namespace std::chrono;
-        while (gAudioIO->IsBusy()) {
-            std::this_thread::sleep_for(100ms);
-        }
+        // we should not have to force restart the monitoring by stopping the stream
+        // gAudioIO->StopStream();
+        // while (gAudioIO->IsBusy()) {
+        //     using namespace std::chrono;
+        //     std::this_thread::sleep_for(100ms);
+        // }
```
saintmatthieu: I agree. Why not delete this altogether?

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -155,43 +197,53 @@ void Au3AudioInput::startMonitoring()
...
+    // it can only be started when we are stopped
+    return !controller()->isRecording() && !playbackController()->isPlaying() && !playbackController()->isPaused();
 }
 
-bool Au3AudioInput::shouldRestartMonitoring() const
+bool Au3AudioInput::shouldStartMonitoring() const
 {
     if (audibleInputMonitoring()) {
         return true;
     }
```
saintmatthieu: Now that the method is called `shouldStartMonitoring`, I suppose the first thing it should check is if monitoring is already started, right? Wouldn't beginning this method with
```cpp
if (gAudioIO->IsMonitoring()) return false;
```
make more sense?

src/record/internal/au3/au3audioinput.cpp
saintmatthieu: Not sure if you mean the comment or the whole condition. In any case, I think it was okay the way it was: if both the input monitoring and mic-metering checkboxes are unchecked, then there's no reason for there to be monitoring.

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -155,43 +197,53 @@ void Au3AudioInput::startMonitoring()
...
+{
+    // monitoring can't be started if we are recording or playing/pause
+    // it can only be started when we are stopped
+    return !controller()->isRecording() && !playbackController()->isPlaying() && !playbackController()->isPaused();
 }
 
-bool Au3AudioInput::shouldRestartMonitoring() const
+bool Au3AudioInput::shouldStartMonitoring() const
 {
     if (audibleInputMonitoring()) {
```
saintmatthieu: Not your changes, but I suggest we try to make the methods that denote the UI state easily distinguishable from the backend state. Just reading `if (audibleInputMonitoring())`, one can't exclude the possibility that it means the backend has input monitoring ON without verifying.

`audibleInputMonitoringIsChecked`, maybe?

And then there are some new private methods in this class. `startMonitoring` and `stopMonitoring` are unambiguous, it's clear they refer to the backend, but `updateMonitoring` not really. At the time of writing I believe it's for the backend, so I suggest `updateBackendMonitoring` for this one.

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -155,43 +197,53 @@ void Au3AudioInput::startMonitoring()
...
+}
 
-        if (!gAudioIO->IsMonitoring()) {
-            return;
+void Au3AudioInput::updateMonitoring()
+{
+    if (canStartMonitoring()) {
+        if (shouldStartMonitoring()) {
+            startMonitoring();
+        } else {
```
saintmatthieu: Here it's puzzling. How about
```cpp
    const auto backendIsMonitoring = gAudioIO->IsMonitoring();
    if (canStartMonitoring()) {
        const bool shouldBeMonitoring = backendShouldBeMonitoring();
        if (shouldBeMonitoring && !backendIsMonitoring) {
            startMonitoring();
        } else if (!shouldBeMonitoring && backendIsMonitoring) {
            stopMonitoring();
        }
    } else if (backendIsMonitoring) {
        // To be on the safe side
        stopMonitoring();
    }
```
?

src/record/internal/au3/au3audioinput.cpp
saintmatthieu: ... or better adopt my suggestion in https://github.com/audacity/audacity/pull/9690#discussion_r2440467221; then this method is called `backendShouldBeMonitoring()` and its body makes sense as it is.

au3/libraries/lib-audio-io/AudioIO.cpp
luapmartin: the idea is to add clarity on "who'" stopping the monitoring and "documenting" what has to be done to stop the monitoring

I do strongly believe this is adding clarity at low cost (not changing much of AudioIO)

In a near future AudioIO won't be access directly as it is now, but only via the AudioEngine so those methods will be handy to track down what is going on

It took me quite some time to navigate the AudioIO and understand how the monitoring was mixed with the rest of the steaming and playback

src/record/internal/au3/au3audioinput.cpp
luapmartin: yes, I aimed for clarity and easy break point setup and tweaks, I guess this collapsed version you propose also work

src/record/internal/au3/au3audioinput.cpp
luapmartin: I could also call directly to updateMonitoring but I wanted to have some consistency for all updateMonitoring to be registered to lambdas at the same place

src/record/internal/au3/au3audioinput.cpp
luapmartin: again I wanted to provide some kind of "documentation" I don't want someone to add that back without understanding what they do, because au3 does this doesn't mean that we should do the same

the only reason to stop the monitoring from the controller point of view is if user toggles the button monitoring off

the playback is disabling the monitoring when starting in a non explicit manner I did not wanted to replicate this bad design when turning it on

au3/libraries/lib-audio-io/AudioIO.cpp
luapmartin: basically `StopStream` should be only accessed by the AudioEngine not by any class because that lead to those strange playback state hang etc.

luapmartin: or better only internally in AudioIO

then AudioEngine should communicate the intent and AudioIO Stop or Start Stream etc.

this method could be called in `void DeviceManager::Rescan()` too

luapmartin: as an example `AudioIO::StartStream` is only called in `Ret Au3Record::doRecord` it should instead be using the AudioEngine to call `startStream` like what the `Au3Player` is doing in `doPlayTracks`

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -33,39 +33,80 @@ Au3AudioInput::Au3AudioInput()
...
+            updateMonitoring(MonitoringChangeReason::AudibleInputMonitoring);
+        };
     });
 }
 
+void Au3AudioInput::updateMonitoring(const MonitoringChangeReason reason)
+{
+    switch (reason) {
+    case MonitoringChangeReason::Initialization:
+        updateMonitoring();
```
embarc-gabriel: We are always executing `updateMonitoring` here despite the reason, right?
Is there some reason to keep it like that instead of just call updatedMonitoring()?
@luapmartin

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -134,71 +175,57 @@ void Au3AudioInput::setAudibleInputMonitoring(bool enable)
...
     }
-
-    if (!configuration()->isMicMeteringOn()) {
-        return false;
-    }
-
-    if (isTrackMeterMonitoring() || meterController()->isRecordMeterVisible()) {
+    if (configuration()->isMicMeteringOn()) {
         return true;
     }
```
saintmatthieu, line 195: Didn't we agree to leave @embarc-gabriel's optimizations in? Now if mic metering is ON but the popup is closed and the track in focus doesn't match my recording device's number of channel, this method will be returning true, although there is no strict need for it.

saintmatthieu: In other words: I don't think the body of this method needs any change in this PR.

src/record/internal/au3/au3audioinput.cpp
```diff
@@ -134,71 +175,57 @@ void Au3AudioInput::setAudibleInputMonitoring(bool enable)
 {
     gPrefs->Write(wxT("/AudioIO/SWPlaythrough"), enable);
     gPrefs->Flush();
-
-    restartMonitoring();
+    // Is there notification being sent when adjusting Prefs?
+    // we need to find a way to notify that the monitoring state has changed
+    // we could add a local lambda callback
```
saintmatthieu: Is there really a need for a signal? It's not an awkward idea, but surely the code is currently so that there isn't a need for one, right? It's just that I don't really know why there should be this comment.

au3/libraries/lib-audio-io/AudioIO.cpp
```diff
@@ -815,6 +807,16 @@ void AudioIO::ResetOwningProject()
...
 {
     if (mPortStreamV19 || mStreamToken) {
+        // we can't start monitoring if
+        // port stream is already opened
+        // OR
+        // a token is in use (!=0)
+        // What happen if mStreamToken is negative?
+        return;
+    }
+    // TODO above check kind of the same as IsMonitoring() should we get rid of it?
```
saintmatthieu: `IsMonitoring` is checking the exact same thing, so yes, please.

au3/libraries/lib-audio-devices/DeviceManager.cpp
```diff
@@ -249,15 +249,7 @@ void DeviceManager::Rescan()
...
-        if (gAudioIO) {
-            if (gAudioIO->IsMonitoring()) {
-                using namespace std::chrono;
-                gAudioIO->StopStream();
-                while (gAudioIO->IsBusy()) {
-                    std::this_thread::sleep_for(100ms);
-                }
-            }
-        }
+        gAudioIO->StopMonitoring(); // TODO Inject and use IAudioEngine instead
```
saintmatthieu, line 252: I wouldn't venture trying to injecting an interface defined in Au4:
* looks like you'd have to solve an circular dependency problem,
* although at the moment in `au3audio`, `IAudioEngine` will probably have to emancipate from `au3` when its implementation is another audio engine.
* you might be opening a Pandora box: `AudioIO` is used as such in plenty of places. Where to stop?

src/record/internal/au3/au3audioinput.cpp
luapmartin: the problems lies in all other models that have some `emit` like this one:
```cpp
void RecordingPreferencesModel::setIsInputMonitoringOn(bool enabled)
{
    record()->audioInput()->setIsInputMonitoringOn(enabled);
    emit isInputMonitoringOnChanged();
}
```

this viewmodel is already changing the audioInput and should listen to a notification sent by it 

basically we should be sending using `muse::async::Notification` from audioInput 
and both be listening to the updates in `PlaybackToolBarRecordLevelItem` and `RecordingPreferencesModel` instead of emiting there local events

src/record/internal/au3/au3audioinput.cpp
luapmartin: renamed `isInputMonitoringOn`

src/record/internal/au3/au3audioinput.cpp
luapmartin: you are right, I changed it and now it's all the same I will clean that up, thanks

src/record/internal/au3/au3audioinput.cpp
saintmatthieu: Now I understand. It's hard to understand without this additional context, though. Please consider completing this comment, or just remove it, as you prefer.

src/record/internal/au3/au3audioinput.cpp
luapmartin: logic simplified with stopMonitoring when can't monitoring added

src/record/internal/au3/au3audioinput.cpp
luapmartin: I reverted + added explanation + centralised the can Monitoring inside too easier to read hopefully

src/record/internal/au3/au3audioinput.cpp
luapmartin: comment removed because setting moved to config

au3/libraries/lib-audio-devices/DeviceManager.cpp
luapmartin: there much less direct AudioIO access thanks to this PR, at some point there will be none.
let's see if we find an alternative to the injection (maybe with actions or something)

au3/libraries/lib-audio-devices/AudioIOBase.h
```diff
@@ -63,6 +63,7 @@ struct AudioIOStartStreamOptions
     double preRoll{ 0.0 };
 
     bool playNonWaveTracks{ true };
+    mutable bool inputMonitoring{ false };
```
saintmatthieu: I'm skeptical about the need for `mutable`: in `startAudioEngineMonitoring`, you do `audioEngine()->startMonitoring(ProjectAudioIO::GetDefaultOptions(*project));`. Why not
```cpp
auto options = ProjectAudioIO::GetDefaultOptions(*project);
options.inputMonitoring = true;
audioEngine()->startMonitoring(options);
```
?

src/record/view/toolbars/playbacktoolbarrecordlevelitem.cpp
```diff
@@ -32,9 +32,14 @@ PlaybackToolBarRecordLevelItem::PlaybackToolBarRecordLevelItem(const muse::ui::U
...
+    recordConfiguration()->isInputMonitoringOnChanged().onNotify(this, [this]() {
+        emit isInputMonitoringOnChanged();
+    });
+
     resetAudioChannelsVolumePressure();
     emit recordingChannelsCountChanged();
-    emit audibleInputMonitoringChanged();
+    // TODO not sure this 'emit' is needed, should we then also emit isMicMeteringOnChanged(); ?
+    // emit isInputMonitoringOnChanged();
 }
```
luapmartin, line 41: there is this open question still

src/record/view/toolbars/playbacktoolbarrecordlevelitem.cpp
```diff
@@ -32,9 +32,14 @@ PlaybackToolBarRecordLevelItem::PlaybackToolBarRecordLevelItem(const muse::ui::U
...
 
+    recordConfiguration()->isInputMonitoringOnChanged().onNotify(this, [this]() {
+        emit isInputMonitoringOnChanged();
+    });
+
     resetAudioChannelsVolumePressure();
     emit recordingChannelsCountChanged();
-    emit audibleInputMonitoringChanged();
+    // TODO not sure this 'emit' is needed, should we then also emit isMicMeteringOnChanged(); ?
+    // emit isInputMonitoringOnChanged();
```
saintmatthieu: I strongly doubt that these signals are needed (`emit recordingChannelsCountChanged()` included). Otherwise, wouldn't we have to explicitly write the ctor of all `QObject` child classes and from there emit a signal for each property?
Also, binding hasn't yet happened, and I expect all bindings to happen after all `QObject` instantiations.
I'd be okay with you removing `emit isInputMonitoringOnChanged();` and even the signal above, just please add a QA checklist item that the affects UI elements still are functional.

au3/libraries/lib-audio-devices/AudioIOBase.h
luapmartin: because that's the responsibility of the audio engine to read that config setting before calling to the AudioIO with the right options

otherwise we potentially could forget to pass the setting in the option and have several unexpected behaviour

_note that we also have to pass that setting to the StartStream method too so that's why the audioengine took the responsibility_ 

before this value was read from inside the AudioIO with the global pref system so I just moved that responsibility one level up in the "call stack" to who's in charge of calling the AudioIO methods (the Audio Engine)

maybe we should later move the `AudioIOStartStreamOptions` down one level to be managed by the AudioEngine receiving the project and doing those setups... I don't see that so clear at the moment, let's fix forward that once the AudioEngine would be the sole responsible of AudioIO

saintmatthieu: I think that you centralizing access to the input-monitoring pref in the record configuration is a good thing.
I still have a problem with the `mutable`: someone calling `AudioEngine::startStream` with the `inputMonitoring` value of the options argument to one value might not realize that it's mutable and may be contradicted by this overwrite.
Why not adding a boolean argument to `AudioIO::StartStream` ? Or if you really hold onto passing a single struct, I suggest that `IAudioEngine::startStream` takes its own `AudioEngineStartStreamOptions`, that get converted to an `AudioIOStartStreamOptions` before being passed to `AudioIO`.

src/au3audio/iaudioengine.h
```diff
@@ -22,7 +22,12 @@ class IAudioEngine : MODULE_EXPORT_INTERFACE
     virtual bool isBusy() const = 0;
 
     virtual int startStream(const TransportSequences& sequences, double startTime, double endTime, double mixerEndTime, // Time at which mixer stops producing, maybe > endTime
-                            const AudioIOStartStreamOptions& options) = 0;
+                            AudacityProject& project, bool isDefaultPlayTrackPolicy, int projectRate) = 0;
```
saintmatthieu: I'd keep the sample rate a `double`: even if I don't think there's real use for it at project level, it still can be used in DSP and then then we'd keep consistent with the type of `AudioIO::GetRate()`.

saintmatthieu: Naming: it's not correct that it's the project rate. I think that's the rate that we'll ask to PortAudio, right? I'd say `playbackRate` or `playbackSampleRate`.

saintmatthieu (approved):
Please rename the `projectRate` argument as commented or explain why it is correct this way before merging. Otherwise LGTM.

src/au3audio/iaudioengine.h
luapmartin: optionRate?

luapmartin: that's the value coming from `ProjectRate::Get(project).GetRate();` that's why I named that projectRate as it's the one the project is setup with then yes this rate is packed into the playback options (sometimes is just the rate of one track see au3record code, other time it's project rate) those options will be created by the audioengine and sent to audioIO then internally audioIO will use those options to configure the playback end eventually setup the PortAudio stream which will negotiate this rate for setting up the playback stream

PR #9692, opened 2025-10-17
9402 extra playback cleanups

luapmartin:
Resolves: 

Recommended:

QA:

please review 
- [x] playback actions (space bar, buttons, play pause stop rewind start/end...))
- [x] record actions (rec start/stop... buttons/ shortcuts)
- [x] selection (clip and time selection) + playback

pay particular attention to the auto-stop (at the end of selection or end of project ) + restarting-playback
please test after the playback stop when reaching the end of the project or the end of a selection to start again the playback
@chinakov @dozzzzer thanks

### DIFF
--- a/src/effects/effects_base/internal/effectsactionscontroller.cpp
+++ b/src/effects/effects_base/internal/effectsactionscontroller.cpp
@@ -59,18 +59,14 @@ void EffectsActionsController::onEffectTriggered(const muse::actions::ActionQuer
     IF_ASSERT_FAILED(!effectId.empty()) {
         return;
     }
-    constexpr bool shouldSeek = false;
-    constexpr bool shouldUpdatePlaybackRegion = false;
-    playbackController()->stop(shouldSeek, shouldUpdatePlaybackRegion);
+    playbackController()->stop();
 
     effectExecutionScenario()->performEffect(effectId);
 }
 
 void EffectsActionsController::repeatLastEffect()
 {
-    constexpr bool shouldSeek = false;
-    constexpr bool shouldUpdatePlaybackRegion = false;
-    playbackController()->stop(shouldSeek, shouldUpdatePlaybackRegion);
+    playbackController()->stop();
 
     effectExecutionScenario()->repeatLastProcessor();
 }
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -192,11 +192,12 @@ void PlaybackController::seek(const muse::secs_t secs, bool applyIfPlaying)
     player()->seek(secs, applyIfPlaying);
 }
 
-void PlaybackController::reset()
+void PlaybackController::stopSeekAndUpdatePlaybackRegion()
 {
-    constexpr bool shouldSeek = true;
-    constexpr bool shouldUpdatePlaybackRegion = true;
-    stop(shouldSeek, shouldUpdatePlaybackRegion);
+    stop();
+
+    seek(m_lastPlaybackSeekTime, false);
+    updatePlaybackRegion();
 }
 
 Channel<uint32_t> PlaybackController::midiTickPlayed() const
@@ -228,9 +229,7 @@ void PlaybackController::onProjectChanged()
     au::project::IAudacityProjectPtr prj = globalContext()->currentProject();
     if (prj) {
         prj->aboutCloseBegin().onNotify(this, [this]() {
-            constexpr bool shouldSeek = true;
-            constexpr bool shouldUpdatePlaybackRegion = true;
-            stop(shouldSeek, shouldUpdatePlaybackRegion);
+            stopSeekAndUpdatePlaybackRegion();
         });
 
         seek(0.0, false); // TODO: get the previous position from the project data
@@ -259,9 +258,7 @@ void PlaybackController::togglePlayAction()
     const bool isShiftPressed = application()->keyboardModifiers().testFlag(Qt::ShiftModifier);
     if (isPlaying()) {
         if (isShiftPressed) {
-            constexpr bool shouldSeek = true;
-            constexpr bool shouldUpdatePlaybackRegion = true;
-            stop(shouldSeek, shouldUpdatePlaybackRegion);
+            stopSeekAndUpdatePlaybackRegion();
         } else {
             doPause();
         }
@@ -345,9 +342,7 @@ void PlaybackController::playTracksAction(const muse::actions::ActionQuery& q)
 void PlaybackController::rewindToStartAction()
 {
     //! NOTE: In Audacity 3 we can't rewind while playing
-    constexpr bool shouldSeek = true;
-    constexpr bool shouldUpdatePlaybackRegion = true;
-    stop(shouldSeek, shouldUpdatePlaybackRegion);
+    stopSeekAndUpdatePlaybackRegion();
 
     doSeek(0.0, false);
 
@@ -359,9 +354,7 @@ void PlaybackController::rewindToEndAction()
     //! NOTE: In Audacity 3 we can't rewind while playing
     m_lastPlaybackSeekTime = totalPlayTime();
     m_lastPlaybackRegion = { totalPlayTime(), totalPlayTime() };
-    constexpr bool shouldSeek = true;
-    constexpr bool shouldUpdatePlaybackRegion = true;
-    stop(shouldSeek, shouldUpdatePlaybackRegion);
+    stopSeekAndUpdatePlaybackRegion();
 
     selectionController()->resetTimeSelection();
 }
@@ -446,37 +439,17 @@ void PlaybackController::doPause()
     player()->pause();
 }
 
-void PlaybackController::stopAction(const muse::actions::ActionQuery& q)
+void PlaybackController::stopAction()
 {
-    IF_ASSERT_FAILED(q.contains("shouldSeek")) {
-        return;
-    }
-    IF_ASSERT_FAILED(q.contains("shouldUpdatePlaybackRegion")) {
-        return;
-    }
-
-    const bool shouldSeek = q.param("shouldSeek").toBool();
-    const bool shouldUpdatePlaybackRegion = q.param("shouldUpdatePlaybackRegion").toBool();
-    stop(shouldSeek, shouldUpdatePlaybackRegion);
+    stopSeekAndUpdatePlaybackRegion();
 }
 
-void PlaybackController::stop(const bool shouldSeek, const bool shouldUpdatePlaybackRegion)
+void PlaybackController::stop()
 {
     IF_ASSERT_FAILED(player()) {
         return;
     }
-
     player()->stop();
-
-    // FIXME: those two boolean are temporary, the idea is to revisit the calls to PlaybackController::stop from
-    // everywhere and check if we want to keep the seek and the updatePlaybackRegion extra operation
-    // because ideally we should be able to call stop() without any extra operation
-    if (shouldSeek) {
-        seek(m_lastPlaybackSeekTime, false);
-    }
-    if (shouldUpdatePlaybackRegion) {
-        updatePlaybackRegion();
-    }
 }
 
 void PlaybackController::doResume()
--- a/src/playback/iplaybackcontroller.h
+++ b/src/playback/iplaybackcontroller.h
@@ -44,9 +44,8 @@ public:
     virtual bool isPaused() const = 0;
     virtual bool isStopped() const = 0;
 
-    virtual void stop(bool shouldSeek, bool shouldUpdatePlaybackRegion) = 0;
-
-    virtual void reset() = 0;
+    virtual void stop() = 0;
+    virtual void stopSeekAndUpdatePlaybackRegion() = 0;
 
     virtual muse::async::Channel<uint32_t> midiTickPlayed() const = 0;
 
--- a/src/playback/internal/playbackcontroller.h
+++ b/src/playback/internal/playbackcontroller.h
@@ -62,8 +62,8 @@ public:
     bool isPaused() const override;
     bool isStopped() const override;
 
-    void stop(bool shouldSeek, bool shouldUpdatePlaybackRegion) override;
-    void reset() override;
+    void stop() override;
+    void stopSeekAndUpdatePlaybackRegion() override;
 
     muse::async::Channel<uint32_t> midiTickPlayed() const override;
 
@@ -106,7 +106,7 @@ private:
 
     void togglePlayAction();
     void doPlay(bool ignoreSelection);
-    void stopAction(const muse::actions::ActionQuery& q);
+    void stopAction();
     void playTracksAction(const muse::actions::ActionQuery& q);
     void rewindToStartAction();
     void rewindToEndAction();
--- a/src/projectscene/view/toolbars/playbacktoolbarmodel.cpp
+++ b/src/projectscene/view/toolbars/playbacktoolbarmodel.cpp
@@ -188,13 +188,6 @@ void PlaybackToolBarModel::updateStopState()
     const ActionCode code = isRecording ? RECORD_STOP_QUERY.toString() : PLAYBACK_STOP_QUERY.toString();
     const UiAction action = uiActionsRegister()->action(code);
     item->setAction(action);
-    if (!isRecording) {
-        constexpr bool shouldSeek = true;
-        constexpr bool shouldUpdatePlaybackRegion = true;
-        const ActionData data = ActionData::make_arg2<bool, bool>(shouldSeek, shouldUpdatePlaybackRegion);
-        item->setArgs(data);
-    }
-
     const QColor iconColor = QColor(uiConfiguration()->currentTheme().values.value(muse::ui::FONT_PRIMARY_COLOR).toString());
     item->setIconColor(iconColor);
 }
--- a/src/trackedit/internal/trackeditinteraction.h
+++ b/src/trackedit/internal/trackeditinteraction.h
@@ -131,9 +131,7 @@ private:
         if (recordController()->isRecording()) {
             return make_ret(trackedit::Err::DisallowedDuringRecording);
         }
-        constexpr bool shouldSeek = false;
-        constexpr bool shouldUpdatePlaybackRegion = false;
-        playbackController()->stop(shouldSeek, shouldUpdatePlaybackRegion);
+        playbackController()->stop();
 
         return (m_interaction.get()->*method)(std::forward<Args>(args)...);
     }


src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -59,14 +59,15 @@ void EffectsActionsController::onEffectTriggered(const muse::actions::ActionQuer
         return;
     }
 
-    playback()->player()->stop();
+    dispatcher()->dispatch("playback/stop", ActionData::make_arg2<bool, bool>(false, false));
```
embarc-gabriel: nitpick. Maybe we could add some local variables just to be easier to understand the action parameters.
Something like:
```
constexpr bool shouldSeek = false;
constexpr bool shouldUpdateRegion = false
..(shouldSeek, shouldUpdateRegion) 
```

src/effects/effects_base/internal/effectsactionscontroller.h
```diff
@@ -15,7 +15,7 @@
 #include "../ieffectsprovider.h"
 #include "../ieffectpresetsscenario.h"
 #include "iinteractive.h"
-#include "playback/iplayback.h"
+#include "playback/iplaybackcontroller.h"
```
embarc-gabriel: I don´t think we need this include.

src/playback/internal/playbackcontroller.cpp
```diff
@@ -32,7 +32,7 @@ void PlaybackController::init()
 {
     dispatcher()->reg(this, PLAY_CODE, this, &PlaybackController::togglePlay);
     dispatcher()->reg(this, PAUSE_CODE, this, &PlaybackController::pause);
-    dispatcher()->reg(this, STOP_CODE, this, &PlaybackController::stop);
+    dispatcher()->reg(this, PLAYBACK_STOP_CODE, this, &PlaybackController::stopAction);
```
embarc-gabriel: Nitpick.
As we have discussed before, isn't it better to keep the consistency? In my opinion, we should either change all codes to PLAYBACK... and the handlers to somethingAction, or we can leave it as it was before.

src/playback/internal/playbackcontroller.cpp
```diff
@@ -58,20 +58,17 @@ void PlaybackController::init()
     });
 
     m_player = playback()->player();
```
embarc-gabriel, line 61: Can´t we inject the player dependency using `muse::Inject`?

src/effects/effects_base/internal/effectsactionscontroller.cpp
luapmartin: nice idea, on it

src/playback/internal/playbackcontroller.cpp
luapmartin: I agree with you I would like to do `PLAYBACK_` for all those codes and have `Action` suffix, I can do that if it's ok in the scope if this PR

src/effects/effects_base/internal/effectsactionscontroller.h
luapmartin: agreed, removed, thanks

src/playback/internal/playbackcontroller.cpp
luapmartin: I think that the playback is the one "holding" the player (and the audiooutput) and the one responsible of setting it in the global context (next line of your comment) I haven't changed that 

```cpp

namespace au::playback {
class Au3Player;
class Au3AudioOutput;
class Au3Playback : public IPlayback
{
public:
    Au3Playback() = default;

    std::shared_ptr<playback::IPlayer> player(TrackSequenceId id = -1) const override;

    std::shared_ptr<playback::IAudioOutput> audioOutput() const override;

private:
    mutable std::shared_ptr<Au3Player> m_player;
    mutable std::shared_ptr<Au3AudioOutput> m_audioOutput;
};
}
```

no one is injecting the IPlayer all injections are of the IPlayback and access the internals of it IPlayer and IAudioOutput, I tried reducing 100% of the access to the player but since I started this PR some more access to the internal of the playback (to the player) have been made I could clean them if that's desired

luapmartin: I would prefer to access the player via the Playback API (and same for the AudioOutput)

src/effects/effects_base/internal/effectsactionscontroller.cpp
luapmartin:
```cpp
constexpr bool shouldSeek = false;
constexpr bool shouldUpdatePlaybackRegion = false;
const ActionData data = ActionData::make_arg2<bool, bool>(shouldSeek, shouldUpdatePlaybackRegion);
dispatcher()->dispatch("playback/stop", data);
```

src/playback/internal/playbackcontroller.cpp
luapmartin: that means also adding the "playback/..." prefix to the actioncode texts

src/playback/internal/playbackcontroller.cpp
embarc-gabriel: Ok.

luapmartin:
this PR is containing both the 3-fixes from https://github.com/audacity/audacity/pull/9731 and some extra refactor, we could have a look at it after the fixes are merged, thanks

src/playback/internal/playbackcontroller.cpp
luapmartin: should I do that in this PR? @embarc-gabriel

embarc-gabriel: It is up to you.
If this does not take too much time, IMO it's worth it.

src/playback/internal/playbackcontroller.cpp
```diff
@@ -153,17 +153,18 @@ bool PlaybackController::isLoopActive() const
...
         secs_t clipStartTime = selectionController()->selectedClipStartTime();
         secs_t clipEndTime = selectionController()->selectedClipEndTime();
         return { clipStartTime, clipEndTime };
     }
 
+    if (selectionController()->timeSelectionIsNotEmpty()) {
+        return { selectionController()->dataSelectedStartTime(),
+                 selectionController()->dataSelectedEndTime() };
+    }
+
```
saintmatthieu, line 162: I first imagined this PR would be refactoring only, but probably it fixes some things, too, right? What would this change fix?

src/playback/internal/playbackcontroller.cpp
```diff
@@ -154,6 +155,7 @@ bool PlaybackController::isLoopActive() const
 PlaybackRegion PlaybackController::selectionPlaybackRegion() const
 {
     // clip selection have priority over time selection
+    // this?
```
saintmatthieu: ?

src/playback/internal/playbackcontroller.cpp
luapmartin: As I understood the UI and the desired behaviour, I don't think you can technically have a clip selected and make a time selection. 

In other words if you have a single clip selected it should be used to determine the selection.

Is this correct?

src/playback/internal/playbackcontroller.cpp
```diff
@@ -246,17 +248,17 @@ void PlaybackController::togglePlay()
...
-            doSeek(m_player->playbackPosition());
+            doSeek(m_player->playbackPosition(), false);
             play(true /* ignoreSelection */);
         } else {
             resume();
         }
     } else {
-        if (isPlaybackPositionOnTheEndOfProject()) {
-            doSeek(0.0);
+        if (isPlaybackPositionOnTheEndOfProject() || isPlaybackPositionOnTheEndOfPlaybackRegion()) {
```
saintmatthieu, line 278: What does this fix?

src/playback/internal/playbackcontroller.cpp
saintmatthieu: I guess that the selectionController probably ensures that it can't have non-empty time and clip selections at once, or we'd have a bug. If that's correct, which order these two checks are in don't make a difference. Maybe adding an `assert(!(selectionController()->timeSelectionIsNotEmpty() && selectionController()->selectedClips().size() > 0))` to express this, and then nevermind the order.

src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -60,14 +60,15 @@ void EffectsActionsController::onEffectTriggered(const muse::actions::ActionQuer
         return;
     }
 
-    playbackController()->stop(false, false);
+    dispatcher()->dispatch("action://playback/stop", ActionData::make_arg2<bool, bool>(false, false));
```
saintmatthieu: Why shouldn't `EffectsActionsController` call `playbackController()` ?

saintmatthieu:
I wonder if we are not mixing two things in `PlaybackController` : "controller" in the sense of actions controller, and "playback controller" in the sense of application-logic-of-playback (which decides e.g. whether to seek to the beginning of the loop region before playing or not).
@luapmartin I think we discussed that already - didn't we agree on a `PlaybackActionsController` ?

saintmatthieu:
Had a conversation with @luapmartin , this is now clarified: this state is provisional.

Result of our conversation for the record:

### Current state

Player : calls audioengine & AudioIO
Playback : contains IPlayer
PlaybackController : listens to actions and does application logic using player from playback
PlaybackState : in global context

### Desired state

IAudioEngine : thin wrapper around implementation (currently AudioIO)
Player : decides on application logic & calls audioengine
PlaybackActionsController : not injectable, only reacts to actions
PlaybackState : in global context

src/trackedit/internal/au3/au3selectioncontroller.cpp
```diff
@@ -552,9 +552,9 @@ void Au3SelectionController::resetDataSelection()
 {
     MYLOG() << "resetDataSelection";
 
-    const auto playbackPosition = playback()->player()->playbackPosition();
-    m_selectedStartTime.set(playbackPosition, true);
-    m_selectedEndTime.set(playbackPosition, true);
+    constexpr auto initialPlaybackPosition = 0; // if this isn't working inject PlaybackController and replace by playbackController()->playbackPosition();
+    m_selectedStartTime.set(initialPlaybackPosition, true);
+    m_selectedEndTime.set(initialPlaybackPosition, true);
```
saintmatthieu: `resetDataSelection()` is called from 10 external places, and `m_selectedStartTime` is refered to externally in 25 places. Is it wise not to do as your comment says?

src/playback/internal/playbackcontroller.cpp
```diff
@@ -399,6 +397,13 @@ void PlaybackController::pause()
     player()->pause();
 }
 
+void PlaybackController::stop(const muse::actions::ActionData& args)
+{
+    const bool shouldSeek = args.count() >= 1 ? args.arg<bool>(0) : false;
+    const bool shouldUpdatePlaybackRegion = args.count() >= 2 ? args.arg<bool>(1) : false;
+    stop(shouldSeek, shouldUpdatePlaybackRegion);
```
saintmatthieu: Isn't this another behavior change: before this PR, the action "stop" would also seek and update playback region, now it doesn't anymore. Or am I missing something?

src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -60,14 +60,14 @@ void EffectsActionsController::onEffectTriggered(const muse::actions::ActionQuer
...
-    dispatcher()->dispatch("action://playback/stop", ActionData::make_arg2<bool, bool>(false, false));
+    dispatcher()->dispatch("playback/stop", ActionData::make_arg2<bool, bool>(false, false));
 
     effectExecutionScenario()->performEffect(effectId);
 }
 
 void EffectsActionsController::repeatLastEffect()
 {
-    dispatcher()->dispatch("action://playback/stop", ActionData::make_arg2<bool, bool>(false, false));
+    dispatcher()->dispatch("playback/stop", ActionData::make_arg2<bool, bool>(false, false));
```
saintmatthieu: In a previous commit I asked why `EffectsActionsController` shouldn't call `playbackController()`. Now on I also realize that it's more difficult for us to find out what these two booleans are about: with `playbackController()` we could have easily navigated to the interface.

src/playback/internal/playbackcontroller.cpp
luapmartin: leftover of personal note during a debugging session thanks (I sometime use //TODO: Paul or //this ? or //here...) sorry bout that , cleaning it

src/playback/tests/playbackcontroller_tests.cpp
```diff
@@ -232,29 +232,38 @@ TEST_F(PlaybackControllerTests, TogglePlay_WithSelection)
...
+    EXPECT_CALL(*m_player, playbackPosition())
+    .WillRepeatedly(Return(secs_t(0.0)));
+
+    //! [GIVEN] There is single clip selection from 10 to 20 secs
     PlaybackRegion selectionRegion = { secs_t(10.0), secs_t(20.0) };
-    EXPECT_CALL(*m_selectionController, timeSelectionIsNotEmpty())
-    .WillOnce(Return(false));
+    //! [GIVEN] No time selection
+    // EXPECT_CALL(*m_selectionController, timeSelectionIsNotEmpty())
+    // .WillOnce(Return(false)); // this isn't called 1st anymore in PlaybackController::selectionPlaybackRegion() as we 1st check the clip selection
```
saintmatthieu, line 260: I think those changes in the test are ok (the behavior we're interested in doesn't change, only implementation detail) ; I'd remove the lines you've commented out.

src/effects/effects_base/internal/effectsactionscontroller.cpp
luapmartin: this has been addressed after some review
this current PR does that
```cpp
void EffectsActionsController::repeatLastEffect()
{
    constexpr bool shouldSeek = false;
    constexpr bool shouldUpdatePlaybackRegion = false;
    const ActionData data = ActionData::make_arg2<bool, bool>(shouldSeek, shouldUpdatePlaybackRegion);
    dispatcher()->dispatch("playback/stop", data);

    effectExecutionScenario()->repeatLastProcessor();
}
```

src/playback/internal/playbackcontroller.cpp
```diff
@@ -31,6 +32,7 @@ static const ActionQuery CHANGE_INPUT_CHANNELS_QUERY("action://playback/change-i
 void PlaybackController::init()
 {
     dispatcher()->reg(this, PLAY_CODE, this, &PlaybackController::togglePlay);
+    dispatcher()->reg(this, PLAYBACK_PLAY_TRACKS_CODE, this, &PlaybackController::playTracksAction);
```
saintmatthieu: Where does this new action come from?

luapmartin: it should be trigged by `EffectsProvider::doEffectPreview` where we call to `player->playTracks(...`

saintmatthieu: I understand now. But on the longer term, `IPlaybackController` will be split in the actions controller and `IPlayback`, the latter of which will be injectable in `EffectsProvider` and would have the `playTracks` interface, correct? I don't see there being a need for a `play-tracks` action.

src/audio/iaudioengine.h
```diff
@@ -25,10 +32,26 @@ class IAudioEngine : MODULE_EXPORT_INTERFACE
                             AudacityProject& project, bool isDefaultPlayTrackPolicy, double audioStreamSampleRate) = 0;
     virtual void stopStream() = 0;
     virtual void pauseStream(bool pause) = 0;
+    virtual void seekStream(double time) = 0;
 
     virtual void startMonitoring(AudacityProject& project) = 0;
     virtual void stopMonitoring() = 0;
 
+    virtual void setInputVolume(float newInputVolume) = 0;
+    virtual void getInputVolume(float& inputVolume) const = 0;
```
saintmatthieu: Why not `virtual float getInputVolume() const = 0` ? (Likewise below)

src/au3wrap/internal/au3audiodevicesprovider.h
```diff
@@ -5,22 +5,25 @@
 #ifndef AU_AU3WRAP_AU3AUDIODEVICESPROVIDER_H
 #define AU_AU3WRAP_AU3AUDIODEVICESPROVIDER_H
 
-#include "modularity/ioc.h"
+#include "framework/global/modularity/ioc.h"
 
 #include "context/iglobalcontext.h"
-#include "global/types/string.h"
+#include "framework/global/types/string.h"
```
saintmatthieu: Since appending "framework" reveals that "context" isn't part of the framework, grouping the framework headers together would make sense. Then I suppose that "context" should go with the other au4 headers below.

src/effects/effects_base/internal/effectsactionscontroller.cpp
```diff
@@ -59,15 +59,21 @@ void EffectsActionsController::onEffectTriggered(const muse::actions::ActionQuer
     IF_ASSERT_FAILED(!effectId.empty()) {
         return;
     }
-
-    playback()->player()->stop();
+    constexpr bool shouldSeek = false;
+    constexpr bool shouldUpdatePlaybackRegion = false;
+    const ActionData data = ActionData::make_arg2<bool, bool>(shouldSeek, shouldUpdatePlaybackRegion);
+    dispatcher()->dispatch("playback/stop", data);
```
saintmatthieu: I've asked something similar already, sorry if I overlooked your response: why not calling `playbackController()->stop()` ? It's easier to read, navigate, and more light-weight, too.

saintmatthieu: ... and safer, too:  `EffectsActionsController` wouldn't need to know the action's string, that there is a need for two boolean arguments, etc.

src/playback/iplaybackcontroller.h
```diff
@@ -25,12 +25,27 @@ class IPlaybackController : MODULE_EXPORT_INTERFACE
...
+    virtual muse::async::Notification loopRegionChanged() const = 0;
+
+    virtual bool isLoopRegionActive() const = 0;
+    virtual void setLoopRegionActive(const bool active) = 0;
+    virtual void toggleLoopPlayback() = 0;
 
     virtual bool isPaused() const = 0;
     virtual bool isStopped() const = 0;
 
+    virtual void stopAction(const muse::actions::ActionData& args) = 0;
```
saintmatthieu: `stopAction` is used only privately by `PlaybackController` ; IIRC elsewhere in this PR, e.g. in `EffectsProvider`, the "playback/stop" action is used. Instead, what about a proper `stop` virtual method with regular arguments? (I made a similar comment elsewhere.)

saintmatthieu (commented):
A lot of very good clean-ups, thank you very much!
There is at least one comment where there might be a change of behavior (where I commented "What does this fix?"), maybe another one.
There are also several comments about e.g. EffectsProvider having to call a "playback/stop" action rather than using the interface of the playback controllers.
I think that's about it.

src/playback/internal/playbackcontroller.cpp
luapmartin: I did a bunch of renamings

src/playback/internal/playbackcontroller.cpp
saintmatthieu: Discussed with @luapmartin : please add a QA checkbox for this corner case (and others you can think of that might be affected by this).

src/trackedit/internal/au3/au3selectioncontroller.cpp
saintmatthieu: @luapmartin please respond :)

src/playback/tests/playbackcontroller_tests.cpp
saintmatthieu: @luapmartin minor but you haven't responded: why not removing the commented code?

src/playback/internal/playbackcontroller.cpp
embarc-gabriel: Nice!
Sometimes the action code contains `action:://` prefix and other times it doesn´t.
Can you explain to me when this should be added?

src/trackedit/internal/au3/au3selectioncontroller.cpp
luapmartin: replaced with `const auto initialPlaybackPosition = globalContext()->playbackState()->playbackPosition();`

luapmartin:
QA: 
- please review playback actions (space bar, buttons, play pause stop rewind start/end...))
- record actions (rec start/stop... buttons/ shortcuts)
- selection (clip and time selection) + playback
pay particular attention to the auto-stop (at the end of selection or end of project ) + restarting-playback
please test after the playback stop when reaching the end of the project or the end of a selection to start again the playback 

@chinakov @dozzzzer thanks

src/playback/tests/playbackcontroller_tests.cpp
luapmartin: I wanted to explicitly document that behaviour

src/playback/internal/playbackcontroller.cpp
luapmartin: added

src/playback/internal/playbackcontroller.cpp
luapmartin: on my side I got some troubles in another issue with using the `action://` prefix so I end-up not using it, let's see what @igorkorsukov says, thanks

igorkorsukov: > Nice!
Sometimes the action code contains action::// prefix and other times it doesn´t.
Can you explain to me when this should be added?

There is now a leisurely migration from one approach to another. 
Old approach:
We have a text code (ActionCode) for the action (for example, "play", "clip-split") and parameters are passed separately (ActionData) if the action has any.  

New approach:
We replace the code with URI and the parameters can be passed as query (ActionQuery)

The new approach has many advantages, just like simply replacing the code with a URI - it adds semantics about what it is (actions) and what area it belongs to, like `action://playback/play` instead of `play`
Passing parameters via query is also convenient in many cases, for example in a dynamic menu, like a list of effects, where the action is "open effect" and the parameter is "what effect", like `action://effects/open?effectId=...`

It's better to switch to the new system quickly so there aren't two. But no one is forcing it.

igorkorsukov: > I got some troubles in another issue with using the action:// prefix 

To understand, please tell us what problems you encountered?

igorkorsukov: Why is a URI with a schema (`action://playback/play`) better than just `playback/play` (`module/action`) ? 
The schema shows what exactly it is and what to do with it. For example, there are URIs passed to the interactive, like opening a dialog, there are settings codes, and so on. The schema determines that this isn't a settings code, for example (`playback/somekey`), but an action.
The second reason is that it's convenient to work with a ready-made class (`Uri`) that can parse a URI into its parts, including queries.

luapmartin: @embarc-gabriel and @igorkorsukov I got those actions codes converted to action query and I cleaned-up naming it should be good now.

dozzzzer:
@luapmartin 

The stop via button click doesn't work:

dozzzzer:
Tested

Issue #9693, opened 2025-10-17
Text appears fuzzy when display is not scaled to 100% on Windows 11


clin1234:
My display is set to 125% scale with a resolution of 2560x1440; when launching Audacity 3.7.5, the text appears "fuzzy"

dozzzzer:
@clin1234 
Initially, I got confused by the description mentioning fuzzy font rendering on Audacity 3. Turned out, the suggested fix addresses the issue in Audacity 4. 

As for the fix, I've tried the build and it doesn't seem to make the problem go away on my end, sadly.

My setup: Win 10 / 2K monitor / 125% scaling

kryksyh:
@dozzzzer the fix is for audacity3 indeed, I rebased the PR, let me know if this fix works

dozzzzer:
Tested

PR #9694, opened 2025-10-17
Add high DPI support on Windows

clin1234:
Resolves: #9693

This adds high DPI awareness for Windows. Specifically:

* `<dpiAware>`, introduced in Windows Vista (ignored if `<apiAwareness>` is set and run on Windows 10 and later)
* `<dpiAwareness>`, a finer-grain awareness mode introduced in Windows 10

Recommended:

QA:

- [X] Autobot test cases have been run


HumbleDeer:
Once this has been built, I'll test this out as well. I found myself feeling the same way about the text and elements. It wasn't blurry, but not sharp either.

Issue #9252, opened 2025-08-23
Ffmpeg 8


LWinterberg:
### Description

https://ffmpeg.org/index.html#pr8.0 with new Sanyo LD-ADPCM, G.728 decoders

chinakov:
Tested, export works correctly.

kryksyh:
https://github.com/audacity/audacity/pull/9702 needs to be transplanted to master

PR #9702, opened 2025-10-19
FFmpeg 8.0 support

kryksyh:
Resolves: #9252

Recommended:


grliszas14 (approved):

grliszas14:
approved but still, ubuntu 20.04 fails (I don't remember if that was the case before)

kryksyh:
> approved but still, ubuntu 20.04 fails (I don't remember if that was the case before)

it is not related, apparently python 3.10 is no longer supported by `bootstrap.pypa.io`, i'll update it later

PR #9707, opened 2025-10-20
9697 Add setting to configure whether clicking on ruler should trigger playback

### LINKED ISSUE #9697: New setting for clicking the ruler to trigger playback (default: OFF)
### Description

Based on user testing, we discovered that the current "click to initiate playback" functionality in the ruler is causing some significant UX problems. Experienced users kept accidentally triggering playback when they didn't want to.

Our solution is to disable this feature by default and add a new checkbox in the context menu to enable it.

---

Below is the full design. I've appended a new page for this in our Figma project here: https://www.figma.com/design/PIavx2Q7WayFAw2FQORrTw/09---Timeline---rulers?node-id=1854-57756

Image

embarc-gabriel:
Resolves: #9697 

Add setting on the timeline context menu to configure whether clicking on the timeline should trigger playback.

Recommended:

QA:

### DIFF
--- a/src/projectscene/internal/projectsceneactionscontroller.cpp
+++ b/src/projectscene/internal/projectsceneactionscontroller.cpp
@@ -15,6 +15,7 @@ static const ActionCode CLIPPING_IN_WAVEFORM_CODE("toggle-clipping-in-waveform")
 static const ActionCode MINUTES_SECONDS_RULER("minutes-seconds-ruler");
 static const ActionCode BEATS_MEASURES_RULER("beats-measures-ruler");
 static const ActionCode CLIP_PITCH_AND_SPEED_CODE("clip-pitch-speed");
+static const ActionCode TOGGLE_PLAYBACK_ON_RULER_CLICK_ENABLED_CODE("toggle-playback-on-ruler-click-enabled");
 
 void ProjectSceneActionsController::init()
 {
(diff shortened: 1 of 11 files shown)


src/projectscene/internal/projectsceneactionscontroller.cpp
```diff
@@ -15,6 +15,7 @@ static const ActionCode CLIPPING_IN_WAVEFORM_CODE("toggle-clipping-in-waveform")
 static const ActionCode MINUTES_SECONDS_RULER("minutes-seconds-ruler");
 static const ActionCode BEATS_MEASURES_RULER("beats-measures-ruler");
 static const ActionCode CLIP_PITCH_AND_SPEED_CODE("clip-pitch-speed");
+static const ActionCode CLICK_RULER_TO_START_PLAYBACK_CODE("click-ruler-to-start-playback");
```
Eism: the name should be `toggle-...` or `ruler-toggle-...` (I don't remember which name style we chose), but `click-ruler-to-start-playback` is confusing, at first I took it exactly as a call for click on the ruler.

embarc-gabriel: @Eism I hope the naming is a better now.
Any feedback let me know.

Eism (approved):

luapmartin:
about the naming of the action why do we need the "-enabled" suffix?

we don't use it for another toggle like the `loop-region`:
```
        makeMenuItem("toggle-playback-on-ruler-click-enabled"),
        makeSeparator(),
        makeMenuItem("toggle-loop-region"),
```

could we call that `"start-playback-on-ruler-click"` 

I don't think we need `toggle` or `enabled` words

`"start-playback-on-ruler-click"` this is more explicit if true it "start playback on ruler click" if false it doesn’t 

what do you think @Eism @embarc-gabriel ? 

_NOTE: "toggle-loop-region" is fine for me_

Eism:
> "start-playback-on-ruler-click" this is more explicit if true it "start playback on ruler click" if false it doesn’t

But you are suggesting a name for the action, and according to your version, when the action is called, the developer will expect the playback to start, and not the setting to be switched.

(As I understand the issue, after the user turns on the setting, playback will start when the user clicks on the ruler)

luapmartin:
then `toggle-playback-on-ruler-click` without the `-enabled`

PR #9710, opened 2025-10-20
9577 preview dialog freeze fix

### LINKED ISSUE #7877: Remove effect preview progress dialog

### LINKED ISSUE #9577: Using Preview in Audacity effects stalls/hangs Audacity and in some cases crashes
### Bug description

Using Preview on the available effects causes Audacity to hang or crash.

Also Preview plays the whole length of the selection (cf. in Audacity 3 and earlier there was a setting in Playback Prefs to set the required preview length.

### Steps to reproduce

1. Generate 30 second Chirp
2. select it
3. invoke Graphic EQ or Reverb _(and others)_
4. Press the Preview button
5. Observe: the whole selection plays as a preview
6. Observe: Progress bar fails to work (mine sticks at 0% or 3%)
7. Observe: when playback finishes the progress dialog box remains
8. Press the Cancel button
9. Observe dialog box "Audacity4.exe is not responding" and the offer to wait or close the program
10. Click on the Wait
11. Observe: Audacity hangs

 Use Case 2 - Compressor and Limiter crash on preview
1. Generate 30 second Chirp
2. select it
3. invoke Compressor or Limiter
4. Press the Preview button
5. Observe: a brief burst of sound and then Audacity crashes

### Expected behavior

Preview should be capable of playing for a short (user specifiable) duration.

Preview should not crash or hang Audacity

### Actual behavior
**### After preview has played**
Image

**### After Pressing the Cancel button**
Image

### Audacity Version

AU4 alpha Audacity-4.0.0.252790316[3e03151]-x86_64

### Operating system

Windows

### Additional context

In AU3 and earlier we have:

Image

This is missing in AU4:

Image

### LINKED ISSUE #9698: Removing Realtime Effects is Unclear
Image

A delete option would be better than "No Effect" preferably with an icon next to it.

### LINKED ISSUE #9738: Playhead movement becomes choppy if you move your mouse when effect preview is in progress
1. Select some audio
2. Open an effect
3. Click "Preview"
4. Move your mouse during preview
5. Observe: playhead movement becomes choppy

https://github.com/user-attachments/assets/190d5d31-62ab-4ff5-b781-65d5ad799eab

saintmatthieu:
Resolves: #9577
Resolves: #7877
Resolves: #9698
Resolves: #9738

Recommended:

QA:
The following should work for all effect types: VST, LV2, AudioUnit, built-in effects and built-in generators

- [x] Built-in, VST and AudioUnit effects: UI and presets are disabled during preview
- [x] LV2 effects: modifying a control during playback interrupts preview
- [x] "Cancel", pressing Esc or closing  the window during preview close the dialog without applying the effect.
- [x] "Apply" during preview closes the dialog and applies the effect.
- [x] Only the select portion of Audio gets played back
- [x] Please smoke-test all combinations of effect dialogs with respect to destructive/realtime and effect-family (built-in, VST, AudioUnit and LV2). Only just opening one of them should suffice.
- [x] Preview region is dissociated from loop region (e.g. have a small loop region, loop enabled, select a region that's not the same as the loop region: preview should play the selected region, not the loop region)
- [x] Preview duration is limited to 1mn. (Tip: you may go in DevTools, look for "preview" and reduce the max preview duration to speed up testing this.)

Besides:

### DIFF
--- a/src/effects/effects_base/internal/abstractviewlauncher.cpp
+++ b/src/effects/effects_base/internal/abstractviewlauncher.cpp
@@ -6,7 +6,19 @@
 namespace au::effects {
 namespace {
 constexpr const char16_t* REALTIME_VIEWER_URI
-    = u"audacity://effects/realtime_viewer?instanceId=%1&effectState=%2&sync=false&modal=false&floating=true";
+    = u"audacity://effects/realtime_viewer?effectState=%1&sync=false&modal=false&floating=true";
+
+static int gInitializationInstanceId = -1;
+}
+
+muse::Ret AbstractViewLauncher::doShowEffect(int instanceId, EffectFamily family) const
+{
+    muse::UriQuery uri(EFFECT_VIEWER_URI);
+    uri.addParam("effectFamily", muse::Val(family));
+    gInitializationInstanceId = instanceId;
+    const auto ret = interactive()->openSync(uri);
+    gInitializationInstanceId = -1;
+    return ret.ret;
 }
 
 void AbstractViewLauncher::doShowRealtimeEffect(const RealtimeEffectStatePtr& state) const
@@ -19,9 +31,8 @@ void AbstractViewLauncher::doShowRealtimeEffect(const RealtimeEffectStatePtr& st
         LOGW() << "Could not get instance for " << state->GetID().ToStdString();
         return;
     }
-    const auto instanceId = instance->id();
 
-    const muse::UriQuery query{ muse::String(REALTIME_VIEWER_URI).arg(size_t(instanceId)).arg(size_t(state.get())) };
+    const muse::UriQuery query{ muse::String(REALTIME_VIEWER_URI).arg(size_t(state.get())) };
 
     // If the dialog for this specific instance is opened, just raise it.
     if (interactive()->isOpened(query).val) {
@@ -31,7 +42,9 @@ void AbstractViewLauncher::doShowRealtimeEffect(const RealtimeEffectStatePtr& st
         return;
     }
 
+    gInitializationInstanceId = instance->id();
     interactive()->open(query);
+    gInitializationInstanceId = -1;
 }
 
 void AbstractViewLauncher::hideRealtimeEffect(const RealtimeEffectStatePtr& state) const
@@ -39,17 +52,14 @@ void AbstractViewLauncher::hideRealtimeEffect(const RealtimeEffectStatePtr& stat
     IF_ASSERT_FAILED(state) {
         return;
     }
-    const auto instance = std::dynamic_pointer_cast<effects::EffectInstance>(state->GetInstance());
-    if (!instance) {
-        LOGW() << "Could not get instance for " << state->GetID().ToStdString();
-        return;
-    }
-    const auto instanceId = instance->id();
-
-    const muse::UriQuery query{ muse::String(REALTIME_VIEWER_URI).arg(size_t(instanceId)).arg(size_t(state.get())) };
-
+    const muse::UriQuery query{ muse::String(REALTIME_VIEWER_URI).arg(size_t(state.get())) };
     if (interactive()->isOpened(query).val) {
         interactive()->close(query);
     }
+}
+
+int AbstractViewLauncher::initializationInstanceId()
+{
+    return gInitializationInstanceId;
 }
 }
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -448,12 +448,12 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
     if (effect.PreviewsFullSelection()) {
         newCtx.t1 = originCtx.t1;
     } else {
-        // Limit preview time to 1mn. We need to pre-render the audio,
-        // which would take a long time and lots of memory for long selections.
+        // Limit preview time:
+        // We need to pre-render the audio, which would take a long time and lots of memory for long selections.
         // On the other hand, preview isn't typically something users would listen to for more than a few seconds.
         // (Au3 used to read `previewLen` from the `/AudioIO/EffectsPreviewLen` setting.
         // There is no plan at the moment to reintroduce it in Au4.)
-        constexpr double maxPreviewLen = 60.0;
+        const double maxPreviewLen = configuration()->previewMaxDuration();
         const double previewLen = std::min(originCtx.t1 - originCtx.t0, maxPreviewLen);
         double previewDuration = 0.0;
         if (isNyquist && isGenerator) {
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -389,8 +389,14 @@ void restoreEffectStateHack(EffectBase& effect)
 }
 }
 
-muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& settings)
+muse::Ret EffectsProvider::previewEffect(const EffectId& effectId, EffectSettings& settings)
 {
+    ::EffectBase* pEffect = this->effect(effectId);
+    if (!pEffect) {
+        return muse::make_ret(muse::Ret::Code::InternalError);
+    }
+    auto& effect = *pEffect;
+
     const bool isNyquist = effect.GetFamily() == NYQUISTEFFECTS_FAMILY;
     const bool isGenerator = effect.GetType() == EffectTypeGenerate;
 
@@ -598,9 +604,4 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
     }
 
     return muse::make_ok();
-}
-
-muse::Ret EffectsProvider::previewEffect(au3::Au3Project&, Effect* effect, EffectSettings& settings)
-{
-    return doEffectPreview(*effect, settings);
 }
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -590,7 +590,9 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
 
         while (player->isRunning()) {
             using namespace std::chrono;
-            std::this_thread::sleep_for(10ms);
+            // A fast update rate is necessary if we want a smooth play-cursor animation
+            // whether the user interacts with the dialog during preview or not.
+            std::this_thread::sleep_for(1ms);
             QCoreApplication::processEvents();
         }
     }
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -444,18 +444,26 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
     EffectContext newCtx;
 
     //! Step 3.1 - prepare time
-
-    //const bool previewFullSelection = effect.PreviewsFullSelection(); not used at the moment
-    const double previewLen = originCtx.t1 - originCtx.t0;
-    double previewDuration = 0.0;
-    if (isNyquist && isGenerator) {
-        previewDuration = effect.CalcPreviewInputLength(settings, previewLen);
+    newCtx.t0 = originCtx.t0;
+    if (effect.PreviewsFullSelection()) {
+        newCtx.t1 = originCtx.t1;
     } else {
-        previewDuration = std::min(settings.extra.GetDuration(), effect.CalcPreviewInputLength(settings, previewLen));
+        // Limit preview time to 1mn. We need to pre-render the audio,
+        // which would take a long time and lots of memory for long selections.
+        // On the other hand, preview isn't typically something users would listen to for more than a few seconds.
+        // (Au3 used to read `previewLen` from the `/AudioIO/EffectsPreviewLen` setting.
+        // There is no plan at the moment to reintroduce it in Au4.)
+        constexpr double maxPreviewLen = 60.0;
+        const double previewLen = std::min(originCtx.t1 - originCtx.t0, maxPreviewLen);
+        double previewDuration = 0.0;
+        if (isNyquist && isGenerator) {
+            previewDuration = effect.CalcPreviewInputLength(settings, previewLen);
+        } else {
+            previewDuration = std::min(settings.extra.GetDuration(), effect.CalcPreviewInputLength(settings, previewLen));
+        }
+        newCtx.t1 = originCtx.t0 + previewDuration;
     }
 
-    newCtx.t0 = originCtx.t0;
-    newCtx.t1 = originCtx.t0 + previewDuration;
     if ((newCtx.t1 > originCtx.t1) && !isGenerator) {
         newCtx.t1 = originCtx.t1;
     }
@@ -564,6 +572,16 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
         opt.selectedOnly = true;
         opt.startOffset = startOffset;
         opt.isDefaultPolicy = false;
+
+        // Setting looping to `false` ensures that the loop region won't interfere with preview.
+        // Also, looping is of course not effective during preview. Having it visually disabled
+        // makes it clear to the user.
+        const auto loopWasActive = player->isLoopRegionActive();
+        player->setLoopRegionActive(false);
+        const auto restorePlayerState = finally([&] {
+            player->setLoopRegionActive(loopWasActive);
+        });
+        player->setPlaybackRegion({ startOffset + newCtx.t0, startOffset + newCtx.t1 });
 
         muse::Ret ret = player->playTracks(*newCtx.tracks, newCtx.t0, newCtx.t1, opt);
         if (!ret) {
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -570,25 +570,10 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
             return ret;
         }
 
-        using namespace BasicUI;
-
-        // The progress dialog must be deleted before stopping the stream
-        // to allow events to flow to the app during StopStream processing.
-        // The progress dialog blocks these events.
-        {
-            auto progress = MakeProgress(effect.GetName(), XO("Previewing"), ProgressShowStop);
-            while (player->isRunning()) {
-                using namespace std::chrono;
-                std::this_thread::sleep_for(100ms);
-                muse::secs_t playPos = player->playbackPosition() - startOffset;
-                auto previewing = progress->Poll(playPos, newCtx.t1);
-
-                if (previewing != BasicUI::ProgressResult::Success || player->reachedEnd().val) {
-                    break;
-                }
-            }
-
-            player->stop();
+        while (player->isRunning()) {
+            using namespace std::chrono;
+            std::this_thread::sleep_for(10ms);
+            QCoreApplication::processEvents();
         }
     }
 
--- a/src/effects/effects_base/internal/effectsprovider.cpp
+++ b/src/effects/effects_base/internal/effectsprovider.cpp
@@ -4,7 +4,6 @@
 #include "effectsprovider.h"
 #include "effecterrors.h"
 
-#include "global/translation.h"
 #include "au3wrap/internal/domconverter.h"
 #include "au3wrap/internal/wxtypes_convert.h"
 #include "au3wrap/internal/progressdialog.h"
@@ -28,7 +27,9 @@
 #include "au3wrap/au3types.h"
 #include "playback/iplayer.h"
 
-#include "log.h"
+#include "framework/global/log.h"
+#include "framework/global/translation.h"
+#include "framework/global/async/async.h"
 
 using namespace muse;
 using namespace au::effects;
@@ -416,22 +417,7 @@ muse::Ret EffectsProvider::previewEffect(const EffectId& effectId, EffectSetting
     //! ============================================================================
     //! NOTE Step 2 - save origin context (state)
     //! ============================================================================
-    struct EffectContext {
-        double t0 = 0.0;
-        double t1 = 0.0;
-        std::shared_ptr<TrackList> tracks;
-        BasicUI::ProgressDialog* progress = nullptr;
-        bool isPreview = false;
-    };
-
     const EffectContext originCtx = { effect.mT0, effect.mT1, effect.mTracks, effect.mProgress, effect.mIsPreview };
-    auto restoreCtx = finally([&] {
-        effect.mT0 = originCtx.t0;
-        effect.mT1 = originCtx.t1;
-        effect.mTracks = originCtx.tracks;
-        effect.mProgress = originCtx.progress;
-        effect.mIsPreview = originCtx.isPreview;
-    });
 
     // restore internal effect state on return (if needed)
     auto cleanup0 = effect.BeginPreview(settings);
@@ -542,7 +528,7 @@ muse::Ret EffectsProvider::previewEffect(const EffectId& effectId, EffectSetting
             ProgressShowStop
             ); // Have only "Stop" button.
 
-        newCtx.progress = progress.get();
+        newCtx.preparingPreviewProgress = progress.get();
         newCtx.isPreview = true;
 
         // apply new context
@@ -550,7 +536,7 @@ muse::Ret EffectsProvider::previewEffect(const EffectId& effectId, EffectSetting
             effect.mT0 = newCtx.t0;
             effect.mT1 = newCtx.t1;
             effect.mTracks = newCtx.tracks;
-            effect.mProgress = newCtx.progress;
+            effect.mProgress = newCtx.preparingPreviewProgress;
             effect.mIsPreview = newCtx.isPreview;
 
             // Update track/group counts
@@ -584,22 +570,49 @@ muse::Ret EffectsProvider::previewEffect(const EffectId& effectId, EffectSetting
         // makes it clear to the user.
         const auto loopWasActive = player->isLoopRegionActive();
         player->setLoopRegionActive(false);
-        const auto restorePlayerState = finally([&] {
-            player->setLoopRegionActive(loopWasActive);
-        });
         player->setPlaybackRegion({ startOffset + newCtx.t0, startOffset + newCtx.t1 });
+
+        m_effectPreviewState.emplace(originCtx, newCtx.tracks);
+        player->playbackStatusChanged().onReceive(this, [this, effectId, loopWasActive](playback::PlaybackStatus status) {
+            if (status == playback::PlaybackStatus::Running) {
+                return;
+            }
+
+            IF_ASSERT_FAILED(status == playback::PlaybackStatus::Stopped) {
+                // Pausing should not be possible during preview.
+                // If we reset the playback context's tracks while playback is still ongoing, though,
+                // we're in for a crash, hence wait and hope that the "stop" signal will come in due time.
+                return;
+            }
+
+            // Wait for other observers of this signal to be finished, in particular the audio engine,
+            // which should stop playback synchronously. Only then we may delete the preview tracks.
+            async::Async::call(this, [this, effectId, loopWasActive] {
+                const auto player = playback()->player();
+                const auto reset = finally([this, player] {
+                    m_effectPreviewState.reset();
+                    player->playbackStatusChanged().disconnect(this);
+                });
+
+                EffectBase* effect = this->effect(effectId);
+                IF_ASSERT_FAILED(effect && m_effectPreviewState) {
+                    return;
+                }
+
+                const EffectContext& originCtx = m_effectPreviewState->originContext;
+                effect->mT0 = originCtx.t0;
+                effect->mT1 = originCtx.t1;
+                effect->mTracks = originCtx.tracks;
+                effect->mProgress = originCtx.preparingPreviewProgress;
+                effect->mIsPreview = originCtx.isPreview;
+
+                player->setLoopRegionActive(loopWasActive);
+            });
+        });
 
         muse::Ret ret = player->playTracks(*newCtx.tracks, newCtx.t0, newCtx.t1, opt);
         if (!ret) {
             return ret;
-        }
-
-        while (player->isRunning()) {
-            using namespace std::chrono;
-            // A fast update rate is necessary if we want a smooth play-cursor animation
-            // whether the user interacts with the dialog during preview or not.
-            std::this_thread::sleep_for(1ms);
-            QCoreApplication::processEvents();
         }
     }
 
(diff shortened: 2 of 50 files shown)


saintmatthieu:
@petersampsonaudacity I'm aware that this is ignoring your request to re-establish the preview duration preference. We're changing the way we're doing the preview here, so this preference might not be relevant anymore. We still have to find a defense against the user selecting 2h of audio and pressing preview - atm this would load the rendering in memory.

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
```diff
@@ -29,6 +29,31 @@ EffectStyledDialogView {
...
+            // Call later because the preview calls `QCoreApplication::processEvents()`,
+            // and we must make sure it doesn't do this after we've closed the dialog, or we'll be getting that Qt exception
+            // "Object %p destroyed while one of its QML signal handlers is in progress."
+            Qt.callLater(function () {
+                if (accept) {
+                    root.accept()
+                } else {
+                    root.reject()
+                }
+            })
```
kryksyh: Could be simplified to
Qt.callLater(() => accept ? root.accept() : root.reject())

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
```diff
@@ -29,6 +29,31 @@ EffectStyledDialogView {
...
+        }
+    }
+
+    Connections {
+        target: root.window
+        function onClosing(event) {
+            prv.closeWindow(false)
+            // Prevent immediate close so preview/cleanup can finish
+            if (event && event.accept !== undefined) {
```
kryksyh: why is this condition?

It should be `accepted`: https://doc.qt.io/qt-6/qml-qtquick-closeevent.html

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
```diff
@@ -168,12 +194,12 @@ EffectStyledDialogView {
                             minWidth: 80
                             isLeftSide: true
 
-                            text: qsTrc("effects", "Preview")
+                            text: prv.viewer.isPreviewing ? qsTrc("effects", "Stop preview") : qsTrc("effects", "Preview")
```
kryksyh: Maybe you could add translator note here, to clarify what "Preview" and "Stop preview" are in this context, like 

```
text: prv.viewer.isPreviewing
    //: Shown on a button that stops effect preview
    ? qsTrc("effects", "Stop preview")
    //: Shown on a button that starts effect preview
    : qsTrc("effects", "Preview")
```
I always forget to put them when adding strings :'(

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
```diff
@@ -201,14 +229,26 @@ EffectStyledDialogView {
...
                 }
             }
         }
     }
 
+    EffectControlsDisablingOverlay {
+        x: viewerLoader.x
+        y: viewerLoader.y
+        width: viewerLoader.width
+        height: viewerLoader.height
```
kryksyh, line 243:
```
anchors.fill: viewerLoader
```

will do the same

src/playback/internal/au3/au3player.cpp
```diff
@@ -27,6 +27,12 @@ using namespace au::au3;
 
 Au3Player::Au3Player()
 {
+    m_playbackStatus.ch.onReceive(this, [this](PlaybackStatus st) {
+        if (st == PlaybackStatus::Running) {
```
kryksyh, line 31: what does it fix?

saintmatthieu: In `AbstractEffectViewModel`,
```cpp
    player->playbackStatusChanged().onReceive(this, [this](auto) {
        emit isPreviewingChanged();
    });
```
This was received before the first `updatePlaybackState()` (called on its own timer), so the Preview button never changed to "Stop preview".

saintmatthieu: ... ah hold on, this might be obsolete, actually ...

saintmatthieu: ... but still, I think it's better to reset the state of `reachedEnd` as soon as status changes to `Running` rather than waiting on a timer to update it.

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
saintmatthieu: Yeah, thanks. Actually I don't even have to call later `root.accept()` or `root.reject()` here, only prv.view.stopPreview().

src/effects/effects_base/qml/Audacity/Effects/EffectsViewerDialog.qml
saintmatthieu: I didn't know this syntax. Absolutely, thanks!

kryksyh (approved):

src/effects/effects_base/view/effectmanagemenu.h
```diff
@@ -42,7 +42,7 @@ class EffectManageMenu : public muse::uicomponents::AbstractMenuModel
 
     void reload(const EffectId& effectId, const EffectInstanceId& instanceId);
 
-    QString m_instanceId;
+    int m_instanceId = -1;
```
kryksyh, line 45: Is there any reason not to use `EffectInstanceId` here?

src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -590,7 +590,9 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
 
         while (player->isRunning()) {
             using namespace std::chrono;
-            std::this_thread::sleep_for(10ms);
+            // A fast update rate is necessary if we want a smooth play-cursor animation
+            // whether the user interacts with the dialog during preview or not.
+            std::this_thread::sleep_for(1ms);
```
kryksyh: I don't believe that 1ms sleep is a reasonable thing to do, it will just burn the cpu without any reason. 

Also, have you measured the sleep time on windows? As I know the minimal sleep time should be around 16ms

kryksyh: https://godbolt.org/z/P1dx1qEMb

src/effects/effects_base/internal/effectsprovider.cpp
```diff
@@ -565,30 +573,27 @@ muse::Ret EffectsProvider::doEffectPreview(EffectBase& effect, EffectSettings& s
...
-                }
-            }
-
-            player->stop();
+        while (player->isRunning()) {
+            using namespace std::chrono;
+            // A fast update rate is necessary if we want a smooth play-cursor animation
+            // whether the user interacts with the dialog during preview or not.
+            std::this_thread::sleep_for(1ms);
+            QCoreApplication::processEvents();
```
igorkorsukov: Please do not use this technique, it is a hack that creates a lot of problems now and in the future.

saintmatthieu: Not sure what you mean: this technique was in place before this PR.

src/effects/effects_base/internal/abstractviewlauncher.cpp
```diff
@@ -31,25 +42,24 @@ void AbstractViewLauncher::doShowRealtimeEffect(const RealtimeEffectStatePtr& st
         return;
     }
 
+    gInitializationInstanceId = instance->id();
     interactive()->open(query);
+    gInitializationInstanceId = -1;
```
igorkorsukov, line 47: Why did you decide to pass `instanceId` through a global variable and not through the uri parameter?

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: It's a good time to remove it.

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: On MacOS it allows the user to move the window around, move the arrow over the window, etc., without impeding the playback cursor's animation.
At least in terms of visual results, it doesn't matter if sleep time is at least 16ms on Windows, because the reason I reduced this time was specifically for mac.
Do you have other concerns?

src/effects/effects_base/internal/abstractviewlauncher.cpp
saintmatthieu: This was so that the effect view models can inherit from `AbstractEffectViewModel`. `BuiltinEffectModel` now inherits from it, and had `Q_PROPERTY(EffectInstanceId instanceId READ instanceId CONSTANT FINAL)`, so it was confusing to have `AbstractEffectViewModel` with the `instanceId` property not being const, and some class inheriting from it with the same property in const version.

saintmatthieu: Now the views can access the instance ID that's being initialized at construction time, can hence be const.

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: Sounds good, but how?

src/effects/effects_base/view/effectmanagemenu.h
saintmatthieu: I think it's quite clear that effect instance IDs are simple integers. Making sure throughout the code base that all `int`s that refer to effect instance IDs are declared with this alias is tedious. If I change it here, it'd probably mean I'd have to propagate it to the entire code base if I were to be exemplary, and I think it's too much work for too little benefit (if at all).

src/effects/effects_base/internal/abstractviewlauncher.cpp
igorkorsukov: This is all strange...
We usually don't use global variables to store state, much less pass parameters. This will inevitably have side action somewhere, causing errors and creating problems for development (changing the use case, for example, if one day we need to open two effects in a row and there is an asynchronous operation somewhere inside, everything will break)

It's also strange why views and models have a constant instanceId — they show a "set" instance, meaning we need to assign it to them.

saintmatthieu: A "set" instance? I don't understand, could you show me?

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: Looking at the current code, I don't really understand the reason why it stops here? Why not just exit the function?

src/effects/effects_base/internal/abstractviewlauncher.cpp
saintmatthieu: I understand your reaction.
The `instanceId` was first made const here: https://github.com/audacity/audacity/pull/9208. The description will give you more background.
It allowed substantial simplifications of the code (and performance improvements, too).
At a higher level: we know that the instance ID is created and available before the effect views are created. I think this is something we can trust. This is the way I found of leveraging this, allowing these simplifications. If there is another I'm all ears.

igorkorsukov: What we do and call: 
* We created viewers (View+ViewModel) for different types of effects 
* Viewers can show specified effects 
* To show UI of effect, we need to pass the effect instance ID to the viewer. 
* i.e., when opening the viewer, we pass it the instance ID of the effect it should show.

From all this it turns out that the instance ID property of the viewer is an external parameter that we need to set when creating and opening the viewer, therefore it cannot be constant. 
So the fact that it became constant is just confusing. Why is it constant?

saintmatthieu: The best I can do is refer you to the PR I mentioned above and ask you to look at the simplifications this allowed.

> To show UI of effect, we need to pass the effect instance ID to the viewer.

Here is the difference with this approach: the view models have the instance ID, and everything around that needs it asks the view model.

saintmatthieu: In essence: the view models now can rely on them being able to access the effect instance at any time. This simplifies a lot of things.

igorkorsukov: > The instanceId was first made const here: https://github.com/audacity/audacity/pull/9208. The description will give you more background.

I missed... 
But I see that the essence of the viewer (view + viewmodel) has changed significantly compared to what it was before.
I believe that the essence of things should not be changed because of technical problems unless it is absolutely necessary.

How it was: 
The viewer shows what it's asked to show. We can open as many viewers as we want simultaneously, each with a given instance. 

How it became:
The viewer opens a single, current global instance. To display an instance, we specify it as global, open the viewer, and it displays this current global instance.
As I already mentioned, this will create problems in the future. As soon as we need to open multiple instances simultaneously, and there are asynchronous operations within them, the entire system will break. Viewers will not be able to open the "current" instance.

src/effects/effects_base/internal/effectsprovider.cpp
kryksyh: Yeah, why the behavior is different on macos? is it hardware or os dependend?

src/effects/effects_base/internal/abstractviewlauncher.cpp
saintmatthieu: > The viewer opens a single, current global instance. To display an instance, we specify it as global, open the viewer, and it displays this current global instance.

To be precise: the model copies the current global instance ID, which is valid and up-to-date when the view is constructed. After that, the global one is invalidated (but it doesn't matter for the model, which made a copy while it was valid), and will only be updated when the next view is created.

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: Mmh, I thought you were onto something. I just tried, but we both overlooked that this method has a local track that was created just for the preview. Returning deletes it, and preview just crashes.
Maybe there's a way around this, though: we basically may only just have to return a holder of the data needed for playback and reset it once preview is over.

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: Good question. At least it fixed it for both @dozzzzer and myself, but this is no proof.
Let me look into the idea @igorkorsukov gave me, then there wouldn't be this problem.

src/effects/effects_base/internal/abstractviewlauncher.cpp
igorkorsukov: To avoid this global state, we introduced instance ID, to pass it to the URI. This way, the instance itself could be stored in a global variable...
But not now, and certainly not in the future. No one can guarantee that the model will have time to "read and store" it before it's rewritten to be opened by the "next" one or for some other reason we don't know yet.

igorkorsukov: The saddest thing about this is that someone will then "optimize dialog opening" and make the operation asynchronous... and break the opening of effects, and they will be blamed for it... 

```
gInitializationInstanceId = instance->id();
interactive()->open(query);   <--- async operation 
gInitializationInstanceId = -1;   <---- reset instance before model creation 
```

saintmatthieu: Interesting... this is precisely what `AbstractViewLauncher::doShowRealtimeEffect` is doing. I haven't had a problem opening realtime effects - was I lucky?

src/effects/effects_base/internal/effectsprovider.cpp
igorkorsukov: Usually this problem is solved like this: 
* We create the necessary data 
* We subscribe to some operation (cancel, finished playing, etc.) and there we clear the data 
* Starting an asynchronous operation 

That is, data, temporary tracks should be deleted not when exiting the method, but when the asynchronous operation ends. To do this, we need to subscribe to its completion.

src/effects/effects_base/internal/abstractviewlauncher.cpp
saintmatthieu: `interactive()->open(query)` calls [QQmlComponent::createWithInitialProperties](https://doc.qt.io/qt-6/qqmlcomponent.html#createWithInitialProperties), which says "Create an object instance of this component and initialize its top-level properties with initialProperties.".
It doesn't say explicitly that the whole initialization of the view is done synchronously, and I don't know if there is documentation that confirms this, but by debugging, I find that it leads to (at least some) `Component.onCompleted` signals being sent, and the following is called:
```cpp
void QQmlComponentPrivate::complete(QQmlEnginePrivate *enginePriv, ConstructionState *state)
{
    if (state->isCompletePending()) {
        QQmlInstantiationInterrupt interrupt;
        state->creator()->finalize(interrupt);

        state->setCompletePending(false);

        enginePriv->inProgressCreations--;

        if (0 == enginePriv->inProgressCreations) {
            while (enginePriv->erroredBindings) {
                enginePriv->warning(enginePriv->erroredBindings->removeError());
            }
        }
    }
}
```
which does look like there is an intention to complete the initialization of the tree.

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: Interesting, lemme see what I can come up with.

src/effects/effects_base/internal/abstractviewlauncher.cpp
igorkorsukov:
```
async::Promise<Val> Interactive::open(const UriQuery& uri)
{
    return provider()->openAsync(uri);
}

Promise<Val> InteractiveProvider::openAsync(const UriQuery& q)
{
    return async::make_promise<Val>(openFunc(q), PromiseType::AsyncByBody);
}

```

Just for optimization, someone here changes `PromiseType::AsyncByBody` to `PromiseType::AsyncByPromise`, and your entire system will break. At this level, this change is unrestricted; the public contract remains the same, but the event loop where the Qml object is created will change. No one guarantees that it "must" be created in the same event loop where it was called.

igorkorsukov: Or someone in the dialog will add a Loader to load content with the property  `asynchronous: true` - for their own purposes... your system will break

src/effects/effects_base/internal/effectsprovider.cpp
saintmatthieu: @igorkorsukov changes pushed.

src/effects/effects_base/internal/abstractviewlauncher.cpp
saintmatthieu: I agree with that.
It would break as it is, yes. However we could still use a global queue of instance IDs: We can ensure that `AbstractEffectViewModel` is the only consumer. If we can assume that successive calls to `openAsync` will be executed in order (which I guess we can, correct?) then the system still holds.

saintmatthieu: One step closer to conventionality could then be to create an `IOpenEffectViewQueryService` that would manage this queue, and the service could be called from `AbstractEffectViewModel`.

saintmatthieu: > Just for optimization, someone here changes PromiseType::AsyncByBody to PromiseType::AsyncByPromise, and your entire system will break.

In any case, this remark is already relevant to current master for built-in effects when used in realtime mode: `BuiltinEffectModel` uses a similar mechanism and would break if `Interactive::open` decided to postpone creation of the dialog..

igorkorsukov: I believe that using global variables to pass parameters is evil.
It's not for nothing that every programming book mentions this at the very beginning... It might work now, but in a year it won't, and instead of developing something new, we'll have to rewrite the old thing that was supposed to just work... 

What was the original problem? Was it that pointers needed to be checked for null? But now you've solved it, isn't that enough? 

```
    template<typename T>
    const T& settings() const
    {
        const T* st = settings().cast<T>();
        IF_ASSERT_FAILED(st) {
            static T null;
            return null;
        }
        return *st;
    }
```

saintmatthieu: Another possibility: `IInteractive` has a method that guarantees it opens the dialog synchronously, but doesn't wait for it to close before returning. Something between current `open` and `openSync`.

saintmatthieu: > What was the original problem?

It was not only this. The stat of the PR I mentioned above is 38 files changed, 193 insertions(+), 622 deletions(-), a good half of which was made possible because of this.

saintmatthieu: Please give me a chance with that service suggestion of mine. Then there is not going to be a static variable anymore, it will be managed by a service like there are tens of already, and if anything it will make this conversation more focused.

kryksyh: > I believe that using global variables to pass parameters is evil.

This is absolutely true. What bothers me (still as a qml newbie) that we create an uninitialized object, that can be accessed from outside before the initialization is complete. 
Which is also error-prone even if not considered as ugly as using global vars to pass parameters indirectly.

How about using an alternative approach that kinda unifies what Matt did, and what is the more sane thing to do being passing the instance id via url parameters:

Instead of instantiating a model that reads instance id from the environment, we create it via factory that initializes it using instance id passed via uri.

Here an example on reverb view:

reverbviewmodel.h:
```
class ReverbViewModel : public BuiltinEffectModel
...
class ReverbViewModelFactory : public QObject
{
    Q_OBJECT
public:
    Q_INVOKABLE ReverbViewModel* createModel(const EffectInstanceId instanceId)
    {
        return new ReverbViewModel(instanceId, this);
    }
};
```

builtineffectsrepository.cpp:
```
            qmlRegisterSingletonType<ReverbViewModelFactory>("Audacity.Effects", 1, 0, "ReverbViewModelFactory", [](QQmlEngine*, QJSEngine*) -> QObject* {
                return new ReverbViewModelFactory;
            });
            
```

ReverbView.qml

```
    model: ReverbViewModelFactory.createModel(root.instanceId)
```

igorkorsukov: It can be done this way...
We use a simple approach: 
1. Create a object 
2. Setup properties 
3. Call object.init()

In one form or another... 
In this case, the checks are the same in any case, whether it’s a factory, global variables, or properties - everywhere the same situation regarding the correctness of the input arguments.

saintmatthieu: @kryksyh Sounds good to me! Will give it a try.

igorkorsukov: @kryksyh @saintmatthieu to understand what specific problem you want to solve?

igorkorsukov: typical approach of the view-viewmodel pattern:
* There is a view 
* There is a view model. 
* The view shows data from the viewmodel and subscribes to changes (properties binding) 
* The view model prepares data for the view and notifies about changes 

Cases: 
1. The model has no data, we loaded it (initialized it), the model loaded the data and notified the view, the view displayed the new data
2. The model has some data (for example, a list of something), we changed the data (for example, added an element), the model notified the view about the change, the view displayed the new data

In terms of functioning and pattern, these two situations are identical.

kryksyh: @igorkorsukov could elaborate on this simple approach please. 
What we are trying to solve is that when we instantiating a view, lets use ReverbView as an example, we have properpty bindings to the reverb model:

 ```            
            CheckBox {
                id: wetOnly

                text: qsTrc("effects/model", "Wet only")
                checked: model.wetOnly
```

in this case, in every model property we have to check if the model is initialized with the instance id, failing to do so, as I understand could lead to undefined results or a crash.

Can we easily solve this problem, by making sure, that the model is fully initialized before the view reads any of its properties?

Comment (igorkorsukov) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: I agree, this is not a typical situation, typically we create a presentation for a view.
 I see possible solutions: 

1. Use a dummy object. 

```
bool ReverbViewModel::wetOnly() const
{
    return settings<ReverbSettings>().mWetOnly;
}

    template<typename T>
    const T& settings() const
    {
         if (!m_inited) {
              static T null;
             return null;
         }
        const T* st = settings().cast<T>();
        IF_ASSERT_FAILED(st) {
            static T null;
            return null;
        }
        return *st;
    }
```

2. Load the view after initializing the model 

```
Loader {
     active: model.inited
     sourceComponent: Component {
         View {
         
         }
     }
}
```

The factory solution will solve the problem of accessing the instance before it's initialized... but in Qml there will be an error: can't read property of null object (since they are read before the model is initialized)

Comment (igorkorsukov) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: This 
```
model: ReverbViewModelFactory.createModel(root.instanceId)
```

should be no different from
```
Model {
   id: model
   instanceId: root.instanceId
   Component.onComplited: {
         model.init()
   }
}
```

or from this (although I don't recommend it)

```
Model::setInstanceId(.. instanceId) 
{
      m_instanceId = instanceId;
      init();
}

```

It turns out that we essentially have options: 
1. Make it so that an "empty" model can return data by default (typical approach): 
* We can create a presentation for a view (it's not convenient here)
* We can check for init in each method (it's not convenient here) 
* We can use a dummy (empty) object (it's convenient here) 

The advantage of this approach is that we can first display the view and then initialize the model.

2. We can explicitly load the view after initialization (disadvantage - may be a blank screen) 

3. We can build the correct initialization order (using a factory, etc.), essentially implicitly loading the view after initializing the model. The downside is that the order may be broken and a black screen may occur if the initialization takes a long time. 


The first option is the best :)

Comment (igorkorsukov) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: > The advantage of this approach is that we can first display the view and then initialize the model.

A common optimization technique is to first show the view, then (at the next event loop) initialize the model and update the data. Visually, this will appear faster to the user: the dialog opens (quickly), the data is updated...
If we initialize during object creation, we're deprived of this optimization opportunity. It will appear slow to the user: they press a button, but nothing happens, and only after a while does the dialog open (or, in general, the view appears).

One of the reasons why we don't use `QQmlParserStatus` and  `void componentComplete()` in C++

Comment (igorkorsukov) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: A new idea for this specific case

```
Model {
    id: realModel
}

property var emptyState: ({
      wetOnly: false
      ...
})

property var model: realModel.inited ? realModel : emptyState 

```


but the solution with dummies is the easiest

Comment (igorkorsukov) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: I thought Dummy (T) couldn't be created because not everyone has a default constructor, or because it's "expensive" to create. But it turns out it can... 

But let's assume that it's not possible, what are the options:

1. Add a helper accessor method, for example this one 

```
    template<typename T, typename Class>
    T valueOf(T Class::* member, T def) {
        if (m_inited) {
            return settings<Class>()->*member;
        }
        return def;
    }
    
    bool Model::wetOnly() const 
    {
        return valueOf(&ReverbSettings::wetOnly, false);
    }
```

2. Eliminate the typed settings structures of built-in effects and use a collection of parameters with key access instead. This would simplify a lot of things, including the view model. It would eliminate the std::any (EffectSettings) that's difficult to work with because we don't know what's in it, or even what type, at any given code place. It would also make saving and reading from a file easier. It would also make debugging easier. 


The point is the same: don't change the meaning of things because of technical problems :) 
* The viewer shows the instance assigned to it.
* It's better for a viewmodel to provide an empty, default state rather than enforce a specific initialization order. This is much more flexible, while the initialization order is either easily broken or would need to be changed.

Comment (saintmatthieu) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: @igorkorsukov I don't think what you state about
```qml
model: ReverbViewModelFactory.createModel(root.instanceId)
```
not being different from
```qml
Model {
   id: model
   instanceId: root.instanceId
   Component.onComplited: {
         model.init()
   }
}
```
is true.
With the factory approach, `root.instanceId` is guaranteed to be valid since it's a parameter the launcher passes to `IInteractive::open`.
With the init() approach, the view model is created without a valid instance ID. Every part of the model that needs a valid instance ID must deal with the fact that it might not yet be valid.

Dmitry gave me a commit with a PoC of his idea for the reverb in destructive mode. No runtime error encountered so far.

igorkorsukov: As you know, as you know...
I've described above all the advantages of being able to perform a separate "init" operation, for example, after opening a view (dialog). And how to solve the problem of accessing uninitialized data in this case.

1. Dummy (empty) element 
```
    template<typename T>
    const T& settings() const
    {
         if (!m_inited) {
              static T empty;
             return empty;
         }
        const T* st = settings().cast<T>();
        IF_ASSERT_FAILED(st) {
            static T null;
            return null;
        }
        return *st;
    }
  ```
  
  Or
  2. Accessor 
  
  ```
    template<typename T, typename Class>
    T valueOf(T Class::* member, T def) {
        if (m_inited) {
            return settings<Class>()->*member;
        }
        return def;
    }
    
    bool Model::wetOnly() const 
    {
        return valueOf(&ReverbSettings::wetOnly, false);
    }
  ```

Comment (saintmatthieu) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: It's not only about the settings. There is e.g.
* `GraphicEqViewModel::GraphicEqViewModel` and its `GraphicEqBandsModel` child: the implementation there has become much easier since the `GraphicEq` was guaranteed available at all times,
* The effects that have no settings and where the instance is the holder of the settings, like in `ToneViewModel`,
* The `EffectSettingsModelImpl` class,
* ...

I am not willing to rewrite everything, all the more that I personally am satisfied the way it is now up to the vulnerability you found wrt `IInteractive::open`, which Dmitry's suggestion solve without breaking any good practice, it would seem.

Comment (saintmatthieu) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: @kryksyh I'm completing your suggestion. No hurdle so far (I have the PoC for built-in and VST effects in realtime mode on top of the destructive/built-in use-case you already covered) but it still is a lot of changes and everything would have to be re-tested. I suggest to postpone these to a follow-up PR and trust until then that `IInteractive::open` opens synchronously :)

Comment (saintmatthieu) [reply src/effects/effects_base/internal/abstractviewlauncher.cpp:47]: @igorkorsukov 
> The factory solution will solve the problem of accessing the instance before it's initialized... but in Qml there will be an error: can't read property of null object (since they are read before the model is initialized)

I'd like to double-check this with you: in this PR we have e.g.

BuiltinEffectBase.qml:
```qml
Rectangle {
    required property BuiltinEffectModel builtinEffectModel
}
```

NoiseView.qml:
```qml
BuiltinEffectBase {
    builtinEffectModel: NoiseViewModelFactory.createModel(root, root.instanceId)
    ...
    Column {
        ComboBoxWithTitle {
            currentIndex: builtinEffectModel.types.findIndex(type => type.value === noise.type)
        }
```

From the [docs about `required` properties](https://doc.qt.io/qt-6/qtqml-syntax-objectattributes.html#required-properties):

> As the name suggests, required properties must be set when an instance of the object is created.

My understanding of this is that `NoiseViewModelFactory.createModel(root, root.instanceId)` will be evaluated as the CPP class underlying the `Rectangle` is created, i.e. before binding begins.

igorkorsukov: I don't know... :(

PR #9712, opened 2025-10-21
Create loop region only when control key is pressed

### LINKED ISSUE #9672: Improve Region for Triggering Looping Region
### Description

Similiar to https://github.com/audacity/audacity/issues/2182#issuecomment-3405465564, we should improve the region for triggering a looping region by clicking on the timeline. 

During our User tests, we have found that Users often enabled the loop region by accident while trying to adjust the playhead on the timeline. 

---

Fix: we should require a modifier key in addition to clicking and dragging to create a loop region. The ideal modifier is CTRL / Command.

Image

I've amended the 'loop' section of the 'Timeline & Rulers' section of our Figma design to reflect this.
https://www.figma.com/design/PIavx2Q7WayFAw2FQORrTw/09---Timeline---rulers?node-id=633-14546

embarc-gabriel:
Resolves: #9672 

Following the new design, only create the loop region if control key is pressed.

Recommended:

QA:

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
@@ -227,6 +227,58 @@ Rectangle {
             }
         }
 
+        MouseArea {
+            id: timelineMouseArea
+            anchors.fill: parent
+            hoverEnabled: true
+
+            property bool playRegionActivated: false
+
+            function convertToTimelinePosition(e) {
+                let position = mapToItem(timeline, Qt.point(e.x, e.y))
+                position.x = Math.max(0, position.x)
+                position.y = Math.max(0, position.y)
+                return position
+            }
+
+            onPositionChanged: function (e) {
+                let position = convertToTimelinePosition(e)
+                timeline.updateCursorPosition(position.x, position.y)
+                playRegionController.updatePosition(position.x)
+            }
+
+            onPressed: function (e) {
+                let position = convertToTimelinePosition(e)
+                if (timeline.isMajorSection(position.y)) {
+                    playRegionController.startInteraction(position.x, root.ctrlPressed)
+                    playRegionActivated = true
+                }
+            }
+
+            onReleased: function (e) {
+                let position = convertToTimelinePosition(e)
+                playRegionController.finishInteraction(position.x)
+            }
+
+            onClicked: function (e) {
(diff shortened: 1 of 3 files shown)


luapmartin:
Hi @embarc-gabriel this one will conflict with this one https://github.com/audacity/audacity/pull/9692 I guess

src/projectscene/qml/Audacity/ProjectScene/clipsview/TracksClipsView.qml
```diff
@@ -249,6 +251,16 @@ Rectangle {
                     }
                     prv.playRegionActivated = false
                 }
+
+                Connections {
+                    target: root
+
+                    function onCtrlPressedChanged() {
+                        if (!root.ctrlPressed) {
+                            playRegionController.mouseUp(prv.positionX)
```
Eism: It's really hard to understand what's going on here, but the problem is probably in the name of the mouseUp method

Eism: Maybe also because of `prv.positionX`
We usually use the prv object for private parts throughout the file, but here it's used for MouseArea. For MouseArea, it's not needed at all, we can simply use the property.

embarc-gabriel: Ok. Sometimes I feel dumb! Thank you for pointing out the MouseArea problem.
I have renamed the function name. Probably is more readable now. @Eism

embarc-gabriel: I have also removed the positionX variable

src/projectscene/view/timeline/playregioncontroller.cpp
```diff
@@ -23,7 +24,7 @@ void PlayRegionController::init()
     updateIsActive();
 }
 
-void PlayRegionController::mouseDown(double pos)
+void PlayRegionController::startInteraction(double pos, bool cntrlPressed)
```
grliszas14: little typo (just looks strange):
cntrl -> ctrl

grliszas14 (approved):

src/projectscene/view/timeline/playregioncontroller.cpp
embarc-gabriel: Fixed.

Eism (approved):

Issue #9308, opened 2025-09-04
Audacity freezes if GTK3 Qt Platform Theme is used


kryksyh:
### Bug description

Currently, using the GTK3 theme is disabled. When enabled, the app freezes at startup:

<img width="1313" height="821" alt="Image" src="https://github.com/user-attachments/assets/e578dd5a-0c6d-4a52-90be-5d6755a07e77" />

### Steps to reproduce

Start the app with `QT_QPA_PLATFORMTHEME=gtk3`

Note: currently the value of QT_QPA_PLATFORMTHEME is reset if matched 'gtk3'

### Expected behavior

Audacity should allow the user to select platform theme according to their preference

### Audacity Version

fe4ca159

### Operating system

Linux

chinakov:
Still present on latest Master (47ecf3eeabc46945542e87107355900f1fd5c326)

chinakov:
This is fixed, thanks!

Issue #9704, opened 2025-10-20
[Build/CI] [AU3-bug] [AU4-bug] Newer Linux distributions don't have libjpeg.so.8 in their repos


PlatinumLucario:
### Bug description

In many new Linux distributions, such as Fedora 42, `libjpeg.so.8` is not found or installed in the `libjpeg-turbo` or `libjpeg-turbo-devel` packages.

Best way to solve this, is to provide `libjpeg.so.8` as a fallback if `libjpeg.so.8` isn't found or include this library version as a submodule if providing `libjpeg.so.8` isn't an option.

Fedora 42 contains `libjpeg.so.62.4.0` in its libjpeg-turbo repo, but not `libjpeg.so.8`, unfortunately.

### Steps to reproduce

Fedora:
1. Install all needed dependencies in Fedora 42 as seen [here](https://github.com/audacity/audacity/blob/master/BUILDING.md).
2. Attempt to build with either the command line or using Qt Creator.

### Expected behavior

It should complete building Audacity 3 or 4 without any issues.

### Actual behavior

The build begins failing after this line
```
[ 96%] Linking CXX executable audacity
/usr/bin/ld: warning: libjpeg.so.8, needed by ../../_deps/wxwidgets/lib/libwx_gtk2u_core-3.1.so, not found (try using -rpath or -rpath-link)
```
Then
```
/usr/bin/ld: ../../_deps/wxwidgets/lib/libwx_gtk2u_core-3.1.so: undefined reference to `jpeg_start_compress@LIBJPEG_8.0'
/usr/bin/ld: ../../_deps/wxwidgets/lib/libwx_gtk2u_core-3.1.so: undefined reference to `jpeg_read_scanlines@LIBJPEG_8.0'
/usr/bin/ld: ../../_deps/wxwidgets/lib/libwx_gtk2u_core-3.1.so: undefined reference to `jpeg_set_quality@LIBJPEG_8.0'
/usr/bin/ld: ../../_deps/wxwidgets/lib/libwx_gtk2u_core-3.1.so: undefined reference to `jpeg_destroy_compress@LIBJPEG_8.0'
/usr/bin/ld: ../../_deps/wxwidgets/lib/libwx_gtk2u_core-3.1.so: undefined reference to `jpeg_finish_compress@LIBJPEG_8.0'
[...]

Issue #9309, opened 2025-09-04
The keyboard focus is not set to the file dialog


kryksyh:
### Bug description

The full description: https://github.com/musescore/MuseScore/issues/13113

The solution used in MuseScore is to force GTK3 dialog, which is causing problems in Audacity: #9308 

### Steps to reproduce

- Open the app
- Open file dialog
- Try to use keyboard arrows

### Expected behavior

The keyboard focus should be set to file open dialog

### Actual behavior

The keyboard focus is left on the main window

### Audacity Version

fe4ca159

### Operating system

Linux

chinakov:
This is still present on the latest Master (47ecf3eeabc46945542e87107355900f1fd5c326)

chinakov:
This is fixed, thanks!

Issue #9298, opened 2025-09-03
Use Wayland (with Qt 6.9.1)


saintmatthieu:
### Bug description

While upgrading to Qt 6.9.1, we disabled Wayland support because it caused crashes and @igorkorsukov reported that it caused many issues in MSS.

What we've observed:
* possible crashes (for @kryksyh (ArchLinux, NVIDIA) always, for @saintmatthieu (WSL) not)
* Submenus not appearing in correct places.

Sticking with x11 is not without drawbacks, though: (please double-check with 6.2.4: some may not be regressions)
* Windows are resizable while they shouldn't (e.g. effect windows, pitch-and-speed dialog)
* Playback toolbar buttons are transparent: 
<img width="294" height="349" alt="Image" src="https://github.com/user-attachments/assets/004662a9-359d-4c49-9a31-4d35947898a3" />
* Realtime Effect dialogs behave like modals: can't interact with the rest of the app while they are open
* 

### Steps to reproduce

Use Audacity on Linux. May have to use WSL to reproduce all of these. 

### Expected behavior

Things work

### Audacity Version

Qt 6.9.1

### Operating system

Linux

igorkorsukov:
Here is another list of problems with Wayland that we encountered. Some are closed, but these are not fixes, but due to the forced use of X11
https://github.com/musescore/MuseScore/issues?q=is%3Aissue%20state%3Aopen%20label%3AWayland

chinakov:
Closing as https://github.com/audacity/audacity/pull/9728 fixes this.

PR #9728, opened 2025-10-24
Remove dependency on GTK2

kryksyh:
Resolves: #9308, #9309, #9704
**QA:** To check if #9298 is also solved

Removes dependency on GTK2

Recommended:

QA:


src/app/main.cpp
```diff
@@ -165,10 +165,8 @@ int main(int argc, char** argv)
 #endif
 
 #ifdef Q_OS_LINUX
-    if (qEnvironmentVariable("QT_QPA_PLATFORM") != "offscreen") {
-        // At the time of writing the app hangs when started with "gtk3" theme
-        // TODO: #9308
-        qputenv("QT_QPA_PLATFORMTHEME", "qt6ct");
+    if (qEnvironmentVariable("MU_QT_QPA_PLATFORM") != "offscreen") {
```
igorkorsukov: todo for my self: rename MU_QT_QPA_PLATFORM to MUSE_QT_QPA_PLATFORM

igorkorsukov (approved):

src/app/main.cpp
kryksyh: Oh, I copied it from MU without noticing the MU_ prefix, i'll remove it

igorkorsukov: I would recommend keep the prefix or replacing it with AU_.
Some people have this setting at the system level, probably for other Qt apps. But ours aren't working well yet with wayland

PR #9729, opened 2025-10-24
update muse to u251024

igorkorsukov:


chinakov:
Found an issue on Linux (I am unsure if these issues are Linux-specific:
- Clip Color is not available when trying to paint a clip a different color
image

igorkorsukov:
> Found an issue on Linux (I am unsure if these issues are Linux-specific:
> 
> * Clip Color is not available when trying to paint a clip a different color


fixed

Issue #9411, opened 2025-09-18
Looping region only works once in a project


chinakov:
### Bug description

If the User opens the project attached, and sets the loop to be 2 bars exactly, the loop will only be played once. 

### Steps to reproduce

1. Launch AU4
2. Open attached project
3. Set looping region to 2 bars
4. Play
5. Observe result

### Expected behavior

Loop plays infinitely

### Actual behavior

The loop only plays twice

### Audacity Version

23d28e2b795b756bf9a57a01e2535380ee41de16

### Operating system

OS independent

### Additional context

[MacMicMusic.zip](https://github.com/user-attachments/files/22405845/MacMicMusic.zip)

chinakov:
Still reproducible on the latest Master (3a5e9510e24579f118f4601854bae366853fd9a6)

luapmartin:
I think this commit would fix that
https://github.com/audacity/audacity/pull/9484/commits/69da1432a9d6b5e500debb4ebc458fc8afac60a0

luapmartin:
Hi, there is one PR for 3 issues please test using https://github.com/audacity/audacity/pull/9731, thanks

dozzzzer:
tested

Issue #9360, opened 2025-09-12
During playback, when the looping region is being dragged, playhead flickers at the beginning of the looping region


dozzzzer:
https://github.com/user-attachments/assets/5b77c919-8d87-4259-a5f6-c181110199f3

luapmartin:
fixed by 81199288834a449b8a16a8e18d4b2d3839ff4afc

luapmartin:
Hi, there is one PR for 3 issues please test using https://github.com/audacity/audacity/pull/9731, thanks

dozzzzer:
Tested

Issue #9052, opened 2025-07-11
When a range selection is cancelled  during playback, the subsequent playback will still end up at the very end of that selection


dozzzzer:
1. Double-click to select the whole clip
2. Start playback
3. While the clip is playing, click into the clip to cancel the selection
4. Observe: playback stops where it should, ie at the very end of the clip
5. Hit Play again and wait for the playback to finish
6. Observe: although the selection is no longer there, it still causes the playback to stop at the end of the clip.

https://github.com/user-attachments/assets/bcfd2ca2-21f7-463f-a3f1-deced7bb9b81

chinakov:
This is reproducible on the latest Master (8696ccf6c1768c39515fb057d8f12257fe4d3706)

chinakov:
This is reproducible on the latest Master (35b55c64762ee3bbecfd7103387db78b7ed5f69c)

luapmartin:
What is happening is that **During Playback** the single click isn't "_positioning the playhead_" so it's not **updating** the range selection. 

Should we also expect the opposite scenario where during playback we select a region or a clip, would we want the playback to stop at the end of the selected region?

Tantacrul:
@luapmartin

> _Should we also expect the opposite scenario where during playback we select a region or a clip, would we want the playback to stop at the end of the selected region?_

I think we probably shouldn't do that. I can't see a really good reason for the user to want that - and it's probably more likely, they'll just end up accidentally stopping playback by selecting a clip.

luapmartin:
OK, so we could go from the selection of a region from a clip or from any region manual selection to the de-selection during playback to let the playback continue and never the other way around meaning if the playback started without any selection it should not be affected by any possible new selection while playing.

Tantacrul:
> _if the playback started without any selection it should not be affected by any possible new selection while playing._

This is a good summary. Yes.

Tantacrul:
By the way, the way I see it in the video - there are two bugs being reported.

- When deselecting during playback, the playhead still stops at the end of the selection. I don't really think this is very important to change - but I'm happy with the change, as long as it isn't v.time consuming.
- When pressing play again, the playhead STILL stops even though there was no selection before the user pressed play. **This is the main bug to be fixed.**

luapmartin:
Hi, there is one PR for 3 issues please test using https://github.com/audacity/audacity/pull/9731, thanks

dozzzzer:
[...]

PR #9731, opened 2025-10-24
playback multi fix

luapmartin:
Resolves: When a range selection is cancelled during playback, the subsequent playback will still end up at the very end of that selection
[#9052](https://github.com/audacity/audacity/issues/9052)
if you click during playback,
the selection will go away,
current playback will reach the end,
next playback will start from click position

we now update the playback region: "on stop", "when reaching end of project or selection" and when "playing"

this PR also fixes rewind to start and rewind to end 

Resolves: During playback, when the looping region is being dragged, playhead flickers at the beginning of the looping region
[#9360](https://github.com/audacity/audacity/issues/9360)

you can now move a loop and the playhead will adjust to reaching the end of the new loop position and not try to stick to the loop start

Resolves: Looping region only works once in a project
[#9411](https://github.com/audacity/audacity/issues/9411)

was happening randomly, hard to repro, could try with a super small loop and try let play for a while and check if stopped after some time (it shoudn't now)

Recommended:

QA:

### DIFF
--- a/src/playback/tests/playbackcontroller_tests.cpp
+++ b/src/playback/tests/playbackcontroller_tests.cpp
@@ -123,21 +123,46 @@ public:
 };
 
 /**
- * @brief Toggle play when stopped
+ * @brief Toggle play when stopped without selection or loop
  * @details User clicked play without any additional params
- *          Playback should be started without seek
+ *          Project has content, no selection, no loop active
+ *          Playback should start from current stopped position without seeking
  */
-TEST_F(PlaybackControllerTests, TogglePlay)
+TEST_F(PlaybackControllerTests, TogglePlay_WhenStopped)
 {
     //! [GIVEN] Playback is stopped
     ON_CALL(*m_player, playbackStatus())
     .WillByDefault(Return(PlaybackStatus::Stopped));
 
-    //! [THEN] No seek position
+    //! [GIVEN] Project has content (totalTime = 100.0)
+    //! This is set up in SetUp() via m_trackeditProject->totalTime()
+
+    //! [GIVEN] Playback position is at some position (not at the end)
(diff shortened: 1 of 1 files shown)


src/playback/internal/au3/au3player.cpp
```diff
@@ -335,14 +334,13 @@ void Au3Player::setPlaybackRegion(const PlaybackRegion& region)
...
     if (region.start == region.end) {
-        auto& tracks = Au3TrackList::Get(project);
-        playRegion.SetEnd(tracks.GetEndTime());
-    } else {
-        playRegion.SetEnd(region.end);
+        const auto& tracks = Au3TrackList::Get(project);
+        end = tracks.GetEndTime();
     }
+    playRegion.SetStart(region.start);
+    playRegion.SetEnd(end);
```
saintmatthieu, line 343: Does this modify the behavior in any way? (Haven't looked at the commit messages yet.)

luapmartin: commit message is "cleanup logic"

just a cleanup

saintmatthieu: I didn't see that, thanks. Yet while these re-writes may present things in an way that you find more sensible, I question that it's the perception of others, too. Meanwhile, every refactoring has potential to introduce bugs, will in any case require reviewers full attention to verify that the logic stayed the same, and also increase the a risk of branch conflict. So I'd be in favor of avoiding these, unless it undoubtedly is worthwhile.

luapmartin: I wanted to reduce `PlaybackRegion` API usages by introducing one double and went from 2 `SetEnd` to 1 I do think it's an improvement, I don't think there's any potential bug that can be introduced by that.

Issue #7787, opened 2024-11-26
Labels


DilsonsPickles:
### Description

Allowing Users to create and manage labels inside a specialised label track. 

The spec below covers the main functionality for labels but does not include styling or the label editor - these will be handled in a separate spec documents.

**Note to dev**
Please arrange a call with @DilsonsPickles to walkthrough this spec. We also may want to break this into smaller development tasks there's a lot of functionality covered in this document.

[Figma link](https://www.figma.com/design/HXlGKcd1TI0NXYj3ICUI4o/17---Labels?node-id=1-89)

chinakov:
This looks comprehensive, I don't see any faults within the designs. Approved, test cases are created. Most probably will be adjusted by a degree once this is developed.

chinakov:
https://github.com/audacity/audacity/issues/9831 - found one issue that @Eism wasn't able to reproduce. Besides this issue, the PR is tested.

PR #9732, opened 2025-10-24
Label tracks interaction

Eism:
Resolves: #7787

### DIFF
--- a/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
+++ b/src/projectscene/qml/Audacity/ProjectScene/trackspanel/TrackItem.qml
@@ -112,24 +112,27 @@ ListItemBlank {
         }
     }
 
-    RowLayout {
+    Item {
         anchors.fill: parent
-
-        spacing: 0
 
         Rectangle {
             id: spacer
 
-            color: "transparent"
+            anchors.left: parent.left
+            anchors.top: parent.top
+            anchors.bottom: parent.bottom
+            width: 9
 
-            Layout.fillHeight: true
-            Layout.preferredWidth: 9
+            color: "transparent"
         }
 
         ColumnLayout {
-            Layout.topMargin: 7
-            Layout.margins: 12
-            Layout.alignment: Qt.AlignTop
+            id: contentColumn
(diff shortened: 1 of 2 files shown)


src/projectscene/qml/Audacity/ProjectScene/tracksitemsview/LabelStalk.qml
```diff
@@ -7,25 +7,32 @@ Rectangle {
...
     MouseArea {
+        id: stalkArea
         anchors.fill: parent
         anchors.leftMargin: -1
         anchors.rightMargin: -1
 
         acceptedButtons: Qt.LeftButton
         hoverEnabled: true
-        cursorShape: pressed ? Qt.CloseHandCursor : Qt.OpenHandCursor
+        cursorShape: !root.isForPoint ? Qt.SizeHorCursor : Qt.SizeHorCursor
```
grliszas14: same cursor in both cases?

src/projectscene/view/trackspanel/paneltrackslistmodel.cpp
```diff
@@ -305,6 +305,9 @@ void PanelTracksListModel::endActiveDrag()
 void PanelTracksListModel::addLabelToSelection()
 {
     dispatcher()->dispatch("label-add");
+    QTimer::singleShot(100, [this](){
```
grliszas14, line 308: Can't "label-add" dispatch a "label-rename" as soon as it finishes? Timer looks like a workaround

grliszas14 (approved):

src/projectscene/view/trackspanel/paneltrackslistmodel.cpp
Eism: No, the service doesn't know about the view, so we can't know exactly when the label will appear in the view and call rename action after it

src/projectscene/qml/Audacity/ProjectScene/tracksitemsview/LabelStalk.qml
Eism: fixed

grliszas14 (approved):

Issue #9633, opened 2025-10-10
Export loop region should include project's tail


grliszas14:
### Description

If you happen to export loop region that is beyond project's end -> it should export project's tail as well (reverb,delay etc after last clip)

additionally:
https://github.com/audacity/audacity/pull/9489#pullrequestreview-3322588887

QA:
* test exporting loop region and range selection (both should have effects tail)
* test exporting track with per-track effects as well as master effects

chinakov:
To be investigated by @grliszas14

dozzzzer:
<img width="1512" height="858" alt="Image" src="https://github.com/user-attachments/assets/1c0297fb-8e2a-4d90-af2f-802d14b23127" />

@grliszas14 
- Set up your project so that there's a clip at T>0 on a mono track
- Add an empty stereo track
- Export audio as: full project audio, stereo, (the rest doesn't seem to make any difference), WAV, 16 bit, 44.1 kHz
- Check the length of the output file
- In my example, it should be 16 seconds; its actual length is 10 seconds. This doesn't exist in the current master.

dozzzzer:
Tested

PR #9733, opened 2025-10-24
Export audio with an effects tail

grliszas14:
Resolves: https://github.com/audacity/audacity/issues/9633

Recommended:

QA:

### DIFF
--- a/au3/libraries/lib-mixer/MixerSource.cpp
+++ b/au3/libraries/lib-mixer/MixerSource.cpp
@@ -365,7 +365,6 @@ std::optional<size_t> MixerSource::Acquire(Buffers& data, size_t bound)
             ZeroFill(result, target, *pFloat);
         }
     }
-    mTime += static_cast<double>(target / mRate);
 
     mLastProduced = target;
     assert(target <= bound);
--- a/au3/libraries/lib-mixer/Mix.cpp
+++ b/au3/libraries/lib-mixer/Mix.cpp
@@ -99,6 +99,7 @@ Mixer::Mixer(
     , mApplyVolume{applyVolume}
     , mHighQuality{highQuality}
     , mFormat{outFormat}
+    , mOutRate{outRate}
     , mInterleaved{outInterleaved}
     , mTimesAndSpeed{std::make_shared<TimesAndSpeed>(TimesAndSpeed {
         startTime, stopTime, warpOptions.initialSpeed, startTime })}
@@ -312,6 +313,10 @@ size_t Mixer::Process(const size_t maxToProcess)
 
     if (!maxOut) {
         return 0;
+    }
+
+    if (mOutRate > 0.0) {
+        mTime += static_cast<double>(*maxOut) / mOutRate;
     }
 
     if (backwards) {
--- a/au3/libraries/lib-mixer/Mix.h
+++ b/au3/libraries/lib-mixer/Mix.h
@@ -141,6 +141,7 @@ private:
     const bool mHighQuality;      // dithering
     const sampleFormat mFormat; // output format also influences dithering
     const bool mInterleaved;
+    const double mOutRate;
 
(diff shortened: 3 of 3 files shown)


au3/libraries/lib-import-export/ExportPluginHelpers.cpp
```diff
@@ -34,8 +34,7 @@ std::unique_ptr<Mixer> ExportPluginHelpers::CreateMixer(
             GetEffectStages(*pTrack));
     }
     auto masterEffectStages = GetMasterEffectStages(project);
-    //custom channel mapping isn't support with master effects on
-    assert(masterEffectStages.empty() || (numOutChannels <= 2 && mixerSpec == nullptr));
+
```
saintmatthieu: Why was this assert removed?

au3/libraries/lib-mixer/DownmixStage.h
```diff
@@ -33,10 +33,13 @@ class DownmixStage final : public AudioGraph::Source
     AudioGraph::Buffers mFloatBuffers;
     size_t mNumChannels;
     ApplyVolume mApplyVolume;
+    double* mCurrentTimePtr = nullptr;
+    double mStopTime = 0.0;
```
saintmatthieu: These two should be `const`.

au3/libraries/lib-mixer/DownmixStage.h
```diff
@@ -33,10 +33,13 @@ class DownmixStage final : public AudioGraph::Source
...
 
 public:
 
-    DownmixStage(std::vector<std::unique_ptr<DownmixSource> > downmixSources, size_t numChannels, size_t bufferSize, ApplyVolume applyGain);
+    DownmixStage(std::vector<std::unique_ptr<DownmixSource> > downmixSources, size_t numChannels, size_t bufferSize, ApplyVolume applyGain,
+                 double* currentTimePtr, double stopTime);
```
saintmatthieu: Here as well: `const double*`

au3/libraries/lib-import-export/ExportPluginHelpers.cpp
grliszas14: it was triggered on a mono track (and mono export) for some unknown reason

au3/libraries/lib-mixer/MixerSource.cpp
```diff
@@ -356,16 +355,47 @@ std::optional<size_t> MixerSource::Acquire(Buffers& data, size_t bound)
...
+        // decide if we should keep this source alive by outputting silence,
+        // so downstream per-track FX still get called and can ring out.
+
+        const bool pastStop
+            =backwards
+              ? (mTime <= mT1)
+              : (mTime >= mT1);
+
+        if (pastStop) {
+            // we reached export stop time
```
saintmatthieu: This comment is confusing: it refers to export, but this method is also used for normal playback.

au3/libraries/lib-mixer/MixerSource.cpp
```diff
@@ -356,16 +355,47 @@ std::optional<size_t> MixerSource::Acquire(Buffers& data, size_t bound)
...
+            mLastProduced = 0;
+            return std::nullopt;
+        }
+
+        // otherwise we're still inside the export range,
+        // so we need to feed silence to keep FX alive.
+
+        result = bound;
```
saintmatthieu: There is a confusion in the meaning of the variables here: `bound` is a max buffer size not to exceed in one `Acquire` call, not an indication of how many samples should be produced.
This change explains why, when exporting selected audio, one would get 41000*5 (`PCMExportProcessor::maxBlockLen`) samples when exporting as wav even if only e.g. 2s of audio are selected.

saintmatthieu (approved):

Issue #9108, opened 2025-07-21
Migrate Truncate Silence


ianalarina:
[Figma Link](https://www.figma.com/design/w2kAh1gKKV5bu7htY7158W/08---Effects?node-id=4782-60134&p=f&t=bb6XnFEmwmy7EdtM-0)

chinakov:
Issue found: 
If the User adjusts the Duration value via the down arrow button, the duration will be changed from 0.5s to 0.001s. If the User goes to adjust the duration once again but upwards, it will be in increments of a second. 

https://github.com/user-attachments/assets/f0fdfc94-a6b0-4f1d-babd-b88a6575761a

chinakov:
Tested, above issue is not an issue according to @luapmartin, moving to ready to merge. The effect is bit-perfect when compared to AU3.

luapmartin:
yes, just to provide precision the increment/decrement controller is having some bound values (minimum here 0.001s and maximum, and evolves step-wise) and when under/overflowing by incrementing/decrementing we round to those bound values causing this behaviour as expected from the framework perspective.

If this becomes problematic or hard to grasp for the user, maybe those button should have a different icon when about to have the rounding effect. 

Design : @DilsonsPickles : do you think we should have a different increment/decrement icon when about to over/under flow?

PR #9781, opened 2025-10-30
Migrate Truncate Silence #9108

luapmartin:
Resolves: Migrate Truncate Silence [#9108](https://github.com/audacity/audacity/issues/9108)

please check effect work as expected with all options combinations

_Please note that as with other built-in effects Navigation isn't done yet, should we address all effects Navigation together?_

Recommended:

QA:

### DIFF
--- a/src/effects/builtin/truncatesilence/truncatesilenceviewmodel.cpp
+++ b/src/effects/builtin/truncatesilence/truncatesilenceviewmodel.cpp
@@ -4,7 +4,7 @@
 #include "truncatesilenceviewmodel.h"
 #include "truncatesilenceeffect.h"
 
-#include "../common/measureunitsymbols.h"
+#include "../common/measureunits.h"
 
 #include "framework/global/log.h"
 #include "framework/global/translation.h"
--- a/src/effects/builtin/CMakeLists.txt
+++ b/src/effects/builtin/CMakeLists.txt
@@ -26,7 +26,7 @@ set(MODULE_SRC
     ${CMAKE_CURRENT_LIST_DIR}/common/generatoreffect.h
     ${CMAKE_CURRENT_LIST_DIR}/common/generatoreffectmodel.cpp
     ${CMAKE_CURRENT_LIST_DIR}/common/generatoreffectmodel.h
-    ${CMAKE_CURRENT_LIST_DIR}/common/measureunitsymbols.h
+    ${CMAKE_CURRENT_LIST_DIR}/common/measureunits.h
 
     ${CMAKE_CURRENT_LIST_DIR}/common/valuewarper/valuewarper.cpp
     ${CMAKE_CURRENT_LIST_DIR}/common/valuewarper/valuewarper.h
--- a/src/effects/builtin/common/measureunits.h
+++ b/src/effects/builtin/common/measureunits.h
@@ -32,6 +32,7 @@ namespace units {
 inline Unit seconds()
 {
     return {
+        //: Abbreviation of "seconds"
         muse::qtrc("effects/units", "s"),
         muse::qtrc("effects/units", "seconds")
     };
@@ -40,6 +41,7 @@ inline Unit seconds()
 inline Unit milliseconds()
 {
     return {
+        //: Abbreviation of "milliseconds"
         muse::qtrc("effects/units", "ms"),
         muse::qtrc("effects/units", "milliseconds")
     };
@@ -57,6 +59,7 @@ inline Unit samples()
 inline Unit hertz()
 {
     return {
+        //: Abbreviation of "hertz"
         muse::qtrc("effects/units", "Hz"),
         muse::qtrc("effects/units", "hertz")
     };
@@ -65,6 +68,7 @@ inline Unit hertz()
 inline Unit kilohertz()
 {
     return {
+        //: Abbreviation of "kilohertz"
         muse::qtrc("effects/units", "kHz"),
         muse::qtrc("effects/units", "kilohertz")
     };
@@ -74,6 +78,7 @@ inline Unit kilohertz()
 inline Unit decibels()
 {
     return {
+        //: Abbreviation of "decibels"
         muse::qtrc("effects/units", "dB"),
         muse::qtrc("effects/units", "decibels")
     };
@@ -83,6 +88,7 @@ inline Unit decibels()
 inline Unit percent()
 {
     return {
+        //: Symbol of "percent"
         muse::qtrc("effects/units", "%"),
         muse::qtrc("effects/units", "percent")
     };
@@ -101,6 +107,7 @@ inline Unit semitones()
 inline Unit beatsPerMinute()
 {
     return {
+        //: Abbreviation of "beats per minute"
         muse::qtrc("effects/units", "bpm"),
         muse::qtrc("effects/units", "beats per minute")
     };
@@ -110,6 +117,7 @@ inline Unit beatsPerMinute()
 inline Unit degrees()
 {
     return {
+        //: Symbol of "degrees"
         muse::qtrc("effects/units", "°"),
         muse::qtrc("effects/units", "degrees")
     };


au3/libraries/lib-builtin-effects/TruncSilenceBase.cpp
```diff
@@ -19,7 +19,7 @@
 #include "EffectOutputTracks.h"
 #include "Prefs.h"
 #include "Project.h"
-#include "SyncLock.h"
+// #include "SyncLock.h"
```
kryksyh: please remove sync lock references completely. 
In case the synchronization will be reintroduced later it will be implemented differently

src/effects/builtin/truncatesilence/TruncateSilenceView.qml
```diff
@@ -0,0 +1,223 @@
...
+
+                    Column {
+                        width: (parent.width - parent.spacing) / 2
+                        spacing: prv.spacingM
+
+                        StyledTextLabel {
+                            text: truncateSilence.thresholdLabel
+                        }
+
+                        TextInputField {
```
kryksyh: It is better to use IncrementalPropertyControl here, as with other effect views. And again as a general rule we should never do this: 
```
currentText: value + " " + some text
```

luapmartin: right, thanks

luapmartin: I have been fooled by the design document not having those increment controllers updating to match your suggestion and adding a comment in Figma, thanks again for spotting that

src/effects/builtin/truncatesilence/TruncateSilenceView.qml
```diff
@@ -0,0 +1,219 @@
...
+        id: prv
+
+        readonly property int spacingS: 4
+        readonly property int spacingM: 8
+        readonly property int spacingL: 16
+        readonly property int fieldWidth: 156
+        readonly property int desiredWidth: 360
+    }
+
+    width: prv.desiredWidth - (2 * prv.spacingL) // we need to remove the padding from the dialog desired width
```
grliszas14: codestyle: width should be on top together with implicitHeight, custom properties after that - in any case, refer to [code guidelines](https://github.com/musescore/muse_framework/wiki/CodeGuidelines#qml)

src/effects/builtin/truncatesilence/TruncateSilenceView.qml
```diff
@@ -0,0 +1,219 @@
...
+
+                        RoundedRadioButton {
+                            text: truncateSilence.compressActionLabel
+                            checked: truncateSilence.actionIndex === 1
+                            onToggled: {
+                                if (truncateSilence.actionIndex !== 1) {
+                                    truncateSilence.actionIndex = 1
+                                }
+                            }
+                        }
```
grliszas14: You can use RadioButtonGroup with a delegate instead of manually handling switching between those

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.h
```diff
@@ -0,0 +1,132 @@
...
+    Q_PROPERTY(QString truncateToLabel READ truncateToLabel CONSTANT FINAL)
+    Q_PROPERTY(QString truncateActionLabel READ truncateActionLabel CONSTANT FINAL)
+    Q_PROPERTY(double truncateValue READ truncateValue WRITE setTruncateValue NOTIFY truncateValueChanged FINAL)
+    Q_PROPERTY(double truncateMin READ truncateMin CONSTANT FINAL)
+    Q_PROPERTY(double truncateMax READ truncateMax CONSTANT FINAL)
+    Q_PROPERTY(double truncateStep READ truncateStep CONSTANT FINAL)
+    Q_PROPERTY(int truncateDecimals READ truncateDecimals CONSTANT FINAL)
+
+    // Compress parameter
+    Q_PROPERTY(QString compressToLabel READ compressToLabel CONSTANT FINAL)
```
grliszas14: all of the labels and mix/max'es doesn't have to be properties, it won't change dynamically (right?) - so it sounds like an overkill. We can simply mark these function with Q_INVOKABLE and then call them as functions from within QML

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.cpp
```diff
@@ -273,4 +277,20 @@ void TruncateSilenceViewModel::doReload()
     emit compressValueChanged();
     emit independentValueChanged();
 }
+
+// Internal unit helpers
```
grliszas14: this also seems like an overkill, a chain of function calls just to get a measureUnit ;) It would be way easier to read and follow if we had this directly in QML as:
```qml
measureUnitsSymbol: qsTrc("global", "dB")
``` 
and this way we also do not multiple very same translations (which are going to be used in a different context atm)

src/effects/builtin/truncatesilence/TruncateSilenceView.qml
luapmartin: will do that, nice catch

src/effects/builtin/truncatesilence/TruncateSilenceView.qml
luapmartin: on it, thanks

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.h
luapmartin: sure good idea

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.cpp
luapmartin: the idea is at some point refactor those parameters into factories that would handle them together, like packing min max step unit ... all together into a meaningful parameter abstraction and have possibility to chose between different representation using the same data structure containing the parameters data

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.h
saintmatthieu: > it won't change dynamically

This is already expressed by the `CONSTANT`. But beside that I agree, a `Q_INVOKABLE` instead would spare one line of code.

I personally still don't see what justifies that a label be part of the model and can't be written directly in the view. For me that's the overkill. I would accept it either way, but would like to hear others opinions and hope they'll get in my direction ;)

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.cpp
saintmatthieu: I agree with the principle from @luapmartin that an effect's model providing a parameter should also provide the parameter's unit. I agree with @grliszas14 though: duplicating translations for the same string is bloating the download.
How about a centralized `dbUnitSymbol` function in e.g. `builtin/common/measureunitsymbols.h` ?

luapmartin: good idea done!

src/effects/builtin/truncatesilence/TruncateSilenceView.qml
luapmartin: done (other action parameters are also conditionally created)

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.cpp
grliszas14: Yes, that's a good idea. One thing though: measure units are used not only within effects module for sure (rather globally I think), so I'd change translation context from this:
```c++
muse::qtrc("effects/units", "ms"),
``` 
to this:
```c++
muse::qtrc("global", "ms"),
``` 

Then we should use the same translation across entire app - this way we'll avoid having multiple entries of the same string (less maintenance for us and translator later).
Unless we want to use "effects/units" context across the app? But I think making it "global" is easier to remember

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.h
luapmartin: It's not part of the `Model`, but of the `ViewModel` - the layer responsible for all view related aspects like labels, formatting, etc

In the `MVVM` pattern, you have `Model`, `View`, and `ViewModel`: in my understanding the `View` should simply "ask" the `ViewModel` what to display, and the `ViewModel` provides it

I've let spacing in qml tho

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.cpp
luapmartin: will be done in https://github.com/audacity/audacity/issues/9800

src/app/configs/dark.cfg
```diff
@@ -30,5 +30,10 @@
     "button_opacity_hover": 0.5,
     "button_opacity_hit": 1.0,
 
-    "item_opacity_disabled": 0.3
+    "item_opacity_disabled": 0.3,
+
+    "spacing_s": 4,
+    "spacing_m": 8,
+    "spacing_l": 16
```
Eism: Could you explain what this is for and how this value is supposed to change between themes? (right now, I see it doesn’t change)

luapmartin: Hi, the idea is to avoid any hard coded values (magic number) for spacing (ideally also other design settings) in qml components. (And make sure that we are respecting the design system, the current designs have to be aligned on a `4px` grid)

Once we are using the theme ui spacing settings in all components it will be possible to configure various spacing according to the currently used theme. 

For example high contrast theme might require some extra spacing different than normal contrast themes to ease reading, when using bigger font...

We could also think about other aspects (radius, font weight, font family, shadow, surface...) all this has to do with design system and design tokens.

Current themes aren't redefining all their settings, most are copy/pasted, I think that's ok for now.

Eism: Overall, I agree that this could be useful in the future. 
However, I’d recommend adding this in a separate PR and applying it to at least one area right away. That way, we can collect feedback from the designers and see if it’s actually needed - since I haven’t seen any requests for this so far.

luapmartin: ok will push that separately  (see: https://github.com/audacity/audacity/issues/9803)

src/effects/builtin/truncatesilence/truncatesilenceviewmodel.h
luapmartin: this is for the "`What` to display"

_for the "`How` to display" it's a bit of a shared mission between theming, components, design system and current ui framework tech in-use_

Issue #9800, opened 2025-10-31
Introduce "common" module


luapmartin:
### Description

introduce `common` module for shared qml components and reused class

luapmartin:
should be named `au_global`

luapmartin:
I would personally prefer not having `au_` prefix as it concern `global` or `common` things, the fact that the muse framework has `global` already should not be a problem with namespaces I guess, let's see 

_the only module starting with `au` are `au3wrap` and `au3audio`_ I don't think we should have module be prefixed by `au_`

PR #9806, opened 2025-11-03
Audio meter in auaudio

saintmatthieu:
The fact that `IMeterSender` is an Au3 API has been propagating Au3 implementations and includes in Au4 modules.
Now we have an adapter that implements `IMeterSender` and whose internal object is an Au4-only `IAudioMeter`, whose implementation is Au3-agnostic.
To make this possible without duplicating types, `audiotypes.h` was also migrated from `au3audio` to `auaudio`. This change has its own commit.

Follow-up on https://github.com/audacity/audacity/pull/9613

Recommended:

QA:


luapmartin:
first glance what I found a bit strange is the module naming, maybe that would we worth it to tidy up things, 

- `auaudio/` is module `auaudio`
- `au3wrap/` is module `au3wrap`
- `au3audio/` is module `audio`

we don't have much `au` module `auaudio` is the first one, is this a new category of modules, should it be `au4audio`, should it be `audio` and should `au3audio` be module `au3audio` instead of `audio`?

saintmatthieu:
@luapmartin I first wanted to use `audio`, but unfortunately the `au3audio` module directory actually defines the `audio` module already.
Probably I should do `audio -> au3audio`, then `auaudio -> audio` wouldn't be a problem. WDYT ?

src/app/CMakeLists.txt
```diff
@@ -85,14 +85,14 @@ set(LINK_LIB
...
     context
     effects_audio_unit
     effects_lv2
     effects_vst
     project
     projectscene
-    audio
+    au3audio
     playback
     record
```
luapmartin, line 97: should we maybe re-order?

src/au3audio/au3audiomodule.h
```diff
@@ -24,7 +23,6 @@ class AudioModule : public muse::modularity::IModuleSetup
 
 private:
 
-    std::shared_ptr<AudioEngine> m_audioEngine;
-    std::shared_ptr<AudioThreadSecurer> m_audioThreadSecurer;
+    std::shared_ptr<Au3AudioEngine> m_audioEngine;
```
luapmartin, line 26: should it be
`std::shared_ptr<au::audio::IAudioEngine> m_audioEngine;` ?

src/au3wrap/internal/au3audiometerfactory.cpp
```diff
@@ -2,11 +2,9 @@
...
+#include "audio/iaudiometer.h"
 
-std::shared_ptr<au::au3::Meter> au::au3::createAudioMeter()
+std::shared_ptr<au::au3::Au3AudioMeter> au::au3::createAudioMeter()
 {
-    auto playingTimer = std::make_unique<AuQtTimer>(Qt::PreciseTimer);
-    auto stoppingTimer = std::make_unique<AuQtTimer>(Qt::VeryCoarseTimer);
-    return std::make_shared<au3::Meter>(std::move(playingTimer), std::move(stoppingTimer));
+    return std::make_shared<Au3AudioMeter>(audio::IAudioMeter::create());
 }
```
luapmartin, line 10: should we change type returned or method name ?

src/au3audio/au3audiomodule.h
saintmatthieu: No, because the `init()` method is `Au3AudioEngine`-only. In general, modules own concrete types of the objects they need to own, you can see this pattern in the other modules, too.

luapmartin:
this will conflict with https://github.com/audacity/audacity/pull/9692 
what should we do ? in which order should we proceed?

src/app/CMakeLists.txt
```diff
@@ -85,14 +85,14 @@ set(LINK_LIB
...
-    auaudio
+    audio
     context
     effects_audio_unit
     effects_lv2
     effects_vst
     project
     projectscene
-    audio
+    au3audio
```
luapmartin: please move au3audio near au3wrap

src/audio/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+    ${CMAKE_CURRENT_LIST_DIR}/audiomodule.h
+    ${CMAKE_CURRENT_LIST_DIR}/audiotypes.h
+    ${CMAKE_CURRENT_LIST_DIR}/iaudioengine.h
+    ${CMAKE_CURRENT_LIST_DIR}/iaudiometer.h
+
+    ${CMAKE_CURRENT_LIST_DIR}/internal/audiometer.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/internal/audiometer.h
+    ${CMAKE_CURRENT_LIST_DIR}/internal/auqttimer.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/internal/auqttimer.h
+    ${CMAKE_CURRENT_LIST_DIR}/internal/itimer.h
```
luapmartin, line 18: auqttimer, itimer aren't `audio` related maybe introduce a `common` or `global` or something in those line module : see https://github.com/audacity/audacity/issues/9800

luapmartin: and my personal preference would be this: https://github.com/audacity/audacity/issues/9800#issuecomment-3487436898

src/app/CMakeLists.txt
saintmatthieu: I don't care so much.

src/audio/CMakeLists.txt
saintmatthieu: I agree, but not in this PR. Within or after 9800, certainly.

luapmartin: ok here then : https://github.com/audacity/audacity/issues/9821

luapmartin (approved):

PR #9818, opened 2025-11-04
9755 realtime effects no persistent fix

### LINKED ISSUE #9755: An effect added to the stack doesn't appear to stick
### Bug description

Effects added to the stack are lost when the project is closed and reopened.

### Steps to reproduce

1. New project
2. Add a new track
3. Add an effect to the stack
4. Close and save
5. Reopen the project
6. Open the track's effect stack
7. Look for the effect in the stack

### Expected behavior

The effect is present in the track's effect stack.

### Actual behavior

No effect in the track's effect stack. Restarting Audacity and reopening this project will fix the issue.

https://github.com/user-attachments/assets/36403bf6-3b73-4708-ad3f-b2b7084a1414

### Audacity Version

master

### Operating system

OS independent

### Additional context

_No response_

saintmatthieu:
Resolves: #9755

Recommended:

QA:

### DIFF
--- a/src/projectscene/view/trackspanel/realtimeeffectlistmodel.cpp
+++ b/src/projectscene/view/trackspanel/realtimeeffectlistmodel.cpp
@@ -25,6 +25,13 @@ void RealtimeEffectListModel::onProjectChanged()
                 emit trackNameChanged();
             }
         });
+        const std::vector<au::trackedit::TrackId> trackIds = project->trackIdList();
+        for (const auto& trackId : trackIds) {
+            if (belongsWithMe(trackId)) {
+                // Project was just opened, we can use brute-force `onChanged`, it won't have an impact on scrollbar position or alike.
+                onChanged(trackId);
+            }
+        }
     } else {
         beginResetModel();
         m_trackEffectLists.clear();


src/projectscene/view/trackspanel/realtimeeffectlistmodel.cpp
```diff
@@ -25,6 +25,13 @@ void RealtimeEffectListModel::onProjectChanged()
                 emit trackNameChanged();
             }
         });
+        const std::vector<au::trackedit::TrackId> trackIds = project->trackIdList();
+        for (const auto& trackId : trackIds) {
+            if (belongsWithMe(trackId)) {
+                // Project was just opened, we can use brute-force `onChanged`, it won't have an impact on scrollbar position or alike.
+                onChanged(trackId);
+            }
+        }
```
luapmartin, line 34: this does an "init()" kind of setup right ?

it's because we have "updates()" working and we want to have the "updates()" forced

saintmatthieu: Sorry, I don't understand the rationale.

luapmartin: what does this do ?

saintmatthieu: When a project is opened, we update the effect list for all tracks in the project.

luapmartin: like an initialisation right by forcing updates 

should we have a method to wrap that forced update into init() or something like that

luapmartin (approved):
see if init method make sense

src/projectscene/view/trackspanel/realtimeeffectlistmodel.cpp
saintmatthieu: No, it's not for an `init()`  method (which here is just called `load()` - I think the two are interchangeable). For a QtObject, `init()`. is typically called in a `Component.onCompleted`, while this is to be executed each time the current project changes.

luapmartin: `load()` is fine then ok with me

saintmatthieu: 🙏

